PR #11437, opened 2025-08-27
fix(forge): populate the git submodule url from git config

wellnana:
The `Dependency::FromStr` implementation sets url to `None` when the input is treated as an alias-only dependency, eg `forge-std`, so we can try to get the url from the git config.


onbjerg (changes requested):
The test is failing

wellnana:
> The test is failing

Ah, I shouldn't change `cmd.rs` revert now, not sure why its flaky test here, but should be good this time

PR #11438, opened 2025-08-27
feat(forge verify-contract): update Sourcify integration to support API v2

### LINKED ISSUE #11065: feat: update Sourcify integration to support API v2
### Component

Forge

- [x] Foundry
- [x] Foundryup

foundryup: 1.0.1

forge verify-check

### Operating System

macOS (Apple Silicon)

### Describe the bug

I wasn't able to get a successful response for any of the already verified contracts on Sourcify using `forge verify-check`:

```bash
$ forge verify-check 0xdAC17F958D2ee523a2206206994597C13D831ec7 --verifier sourcify    
                                                                                                           
Checking verification status on mainnet
Attempting to verify on Sourcify. Pass the --etherscan-api-key <API_KEY> to verify on Etherscan, or use the --verifier flag to verify on another provider.
Contract source code is not verified
```

See the verified contract: https://repo.sourcify.dev/1/0xdAC17F958D2ee523a2206206994597C13D831ec7

Also tried with `--chain=1`

However this command worked:
```bash
$ forge verify-check 0xa5F565650890fBA1824Ee0F21EbBbF660a179934 --verifier sourcify --chain=10        
                                                                                            
Checking verification status on optimism
Attempting to verify on Sourcify. Pass the --etherscan-api-key <API_KEY> to verify on Etherscan, or use the --verifier flag to verify on another provider.
Contract successfully verified
```

This contract also works:
```bash
$ forge verify-check 0x1231DEB6f5749EF6cE6943a275A1D3E7486F4EaE --verifier sourcify                                                                                                               

Checking verification status on mainnet
Attempting to verify on Sourcify. Pass the --etherscan-api-key <API_KEY> to verify on Etherscan, or use the --verifier flag to verify on another provider.
Contract successfully verified
```

Other failing ones:
- 0xD6C2Af244DD56DB3C9D276808EdA8B5Fb040007b
- 0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D

Feels very random and I wasn't able to recognize a pattern.

If you need example verified contracts use this API: https://sourcify.dev/server/v2/contracts/1?limit=10&sort=desc

manuelwedler:
## Motivation

Closes #11065

Sourcify API v1 is deprecated and forge should be updated to use API v2.

API v2 is based on asynchronous verification jobs similar to Etherscan. Therefore, it's necessary to return the verification job ID and check the job for a successful verification.

## Solution

- `forge verify-contract` outputs the verification job id after submitting to Sourcify
- `forge verify-contract --watch` automatically polls the Sourcify API for the job status.
- `forge verify-check` takes a verification job id to follow the same pattern as Etherscan with its GUID.
- Adds support for providing the creation transaction hash to the Sourcify API via a new argument. This makes it easier to get a creation match. `forge create --verify` automatically passes the transaction hash to the verify command.
- Adds the `is_contract_verified_check` which was missing and therefore inconsistent with the Etherscan command.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/forge/src/cmd/create.rs
+++ b/crates/forge/src/cmd/create.rs
@@ -249,6 +249,7 @@ impl CreateArgs {
             guess_constructor_args: false,
             compilation_profile: Some(id.profile.to_string()),
             language: None,
+            creation_transaction_hash: Default::default(),
         };
 
         // Check config for Etherscan API Keys to avoid preflight check failing if no
@@ -439,6 +440,7 @@ impl CreateArgs {
             guess_constructor_args: false,
             compilation_profile: Some(id.profile.to_string()),
             language: None,
+            creation_transaction_hash: Some(receipt.transaction_hash),
         };
         sh_println!("Waiting for {} to detect contract deployment...", verify.verifier.verifier)?;
         verify.run().await
--- a/crates/verify/src/sourcify.rs
+++ b/crates/verify/src/sourcify.rs
@@ -37,6 +37,16 @@ impl VerificationProvider for SourcifyVerificationProvider {
         let body = self.prepare_verify_request(&args, &context).await?;
         let chain_id = args.etherscan.chain.unwrap_or_default().id();
 
+        if !args.skip_is_verified_check && self.is_contract_verified(&args).await? {
+            sh_println!(
+                "\nContract [{}] {:?} is already verified. Skipping verification.",
+                context.target_name,
+                args.address.to_string()
+            )?;
+
+            return Ok(());
+        }
+
         trace!("submitting verification request {:?}", body);
 
         let client = reqwest::Client::new();
@@ -253,6 +263,40 @@ impl SourcifyVerificationProvider {
 
         Ok(req)
     }
+
+    async fn is_contract_verified(&self, args: &VerifyArgs) -> Result<bool> {
+        let chain_id = args.etherscan.chain.unwrap_or_default().id();
+        let base_url = args.verifier.verifier_url.as_deref().unwrap_or(SOURCIFY_URL);
+        let url = format!("{}v2/contract/{}/{}", base_url, chain_id, args.address);
+
+        match reqwest::get(&url).await {
+            Ok(response) => {
+                if response.status() == 200 {
+                    let contract_response: SourcifyContractResponse =
+                        response.json().await.wrap_err("Failed to parse contract response")?;
+
+                    let creation_exact = contract_response
+                        .creation_match
+                        .as_ref()
+                        .map(|s| s == "exact_match")
+                        .unwrap_or(false);
+
+                    let runtime_exact = contract_response
+                        .runtime_match
+                        .as_ref()
+                        .map(|s| s == "exact_match")
+                        .unwrap_or(false);
+
+                    Ok(creation_exact && runtime_exact)
+                } else {
+                    Ok(false)
+                }
+            }
+            Err(error) => Err(error).wrap_err_with(|| {
+                format!("Failed to query verification status for {}", args.address)
+            }),
+        }
+    }
 }
 
 #[derive(Debug, Serialize)]
@@ -285,6 +329,10 @@ pub struct SourcifyJobResponse {
 pub struct SourcifyContractResponse {
     #[serde(rename = "match")]
     match_status: Option<String>,
+    #[serde(rename = "creationMatch")]
+    creation_match: Option<String>,
+    #[serde(rename = "runtimeMatch")]
+    runtime_match: Option<String>,
 }
 
 #[derive(Debug, Deserialize)]
--- a/crates/verify/src/sourcify.rs
+++ b/crates/verify/src/sourcify.rs
@@ -1,15 +1,19 @@
 use crate::{
     provider::{VerificationContext, VerificationProvider},
-    verify::{VerifyArgs, VerifyCheckArgs},
+    retry::RETRY_CHECK_ON_VERIFY,
+    utils::ensure_solc_build_metadata,
+    verify::{ContractLanguage, VerifyArgs, VerifyCheckArgs},
 };
-use alloy_primitives::map::HashMap;
 use async_trait::async_trait;
-use eyre::Result;
-use foundry_common::fs;
+use eyre::{Context, Result, eyre};
+use foundry_common::retry::RetryError;
+use foundry_compilers::{
+    artifacts::{Source, StandardJsonCompilerInput, vyper::VyperInput},
+    solc::SolcLanguage,
+};
 use futures::FutureExt;
-use reqwest::Url;
 use serde::{Deserialize, Serialize};
-use std::str::FromStr;
+use std::path::Path;
 
 pub static SOURCIFY_URL: &str = "https://sourcify.dev/server/";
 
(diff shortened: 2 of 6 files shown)


crates/verify/src/verify.rs
```diff
@@ -507,6 +511,25 @@ impl figment::Provider for VerifyCheckArgs {
     }
 }
 
+fn validate_tx_hash(hash: &str) -> Result<TxHash, String> {
```
DaniPopes: This is already handled by the FromStr implementation with [FromHexError](https://docs.rs/const-hex/latest/const_hex/enum.FromHexError.html), and also allows non-0x prefix, so this function is not needed

image

crates/verify/src/sourcify.rs
```diff
@@ -47,155 +64,320 @@ impl VerificationProvider for SourcifyVerificationProvider {
...
+    error: Option<SourcifyErrorResponse>,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct SourcifyContractResponse {
+    #[serde(rename = "match")]
+    match_status: Option<String>,
+    #[serde(rename = "creationMatch")]
+    creation_match: Option<String>,
+    #[serde(rename = "runtimeMatch")]
```
DaniPopes: you can use `#[serde(rename_all = "camelCase")]` on the struct itself instead of renaming each field individually

manuelwedler:
@DaniPopes thanks for the feedback. I addressed your comments and the CI was successful now.

crates/forge/src/cmd/create.rs
```diff
@@ -249,6 +249,7 @@ impl CreateArgs {
             guess_constructor_args: false,
             compilation_profile: Some(id.profile.to_string()),
             language: None,
+            creation_transaction_hash: Default::default(),
```
grandizzy: this could be ` creation_transaction_hash: None,`

crates/verify/src/sourcify.rs
```diff
@@ -25,16 +29,29 @@ impl VerificationProvider for SourcifyVerificationProvider {
...
+            )?;
+
+            return Ok(());
+        }
 
         trace!("submitting verification request {:?}", body);
 
         let client = reqwest::Client::new();
+        let base_url = args.verifier.verifier_url.as_deref().unwrap_or(SOURCIFY_URL);
+        let url = format!("{}v2/verify/{}/{}", base_url, chain_id, args.address);
```
grandizzy: this repeats several times, let's have a helper / unified way to compose URL. Also, shouldn't `base_url` always be `SOURCIFY_URL` here, or we want to be flexible to change from `https://sourcify.dev/server/`?

crates/verify/src/sourcify.rs
```diff
@@ -47,155 +64,317 @@ impl VerificationProvider for SourcifyVerificationProvider {
...
                     let response = client
-                        .post(args.verifier.verifier_url.as_deref().unwrap_or(SOURCIFY_URL))
+                        .post(&url)
                         .header("Content-Type", "application/json")
                         .body(serde_json::to_string(&body)?)
                         .send()
                         .await?;
 
                     let status = response.status();
-                    if !status.is_success() {
```
grandizzy, line 57: nit: we can reorg this as
```rust
                    return match response.status() {
                        StatusCode::CONFLICT => {}
                        StatusCode::ACCEPTED => {}
                        _ => {}
```

also, won't the new code fail if 200 (as we're only handling 409 and 202) or doesn't matter?

crates/verify/src/sourcify.rs
```diff
@@ -47,155 +64,317 @@ impl VerificationProvider for SourcifyVerificationProvider {
...
-                        "check-by-addresses?addresses={}&chainIds={}",
-                        args.id,
-                        args.etherscan.chain.unwrap_or_default().id(),
+            .run_async_until_break(|| async {
+                let response = reqwest::get(&url)
+                    .await
+                    .wrap_err("Failed to request verification status")
+                    .map_err(RetryError::Retry)?;
+
+                if response.status() == 404 {
```
grandizzy: can be `if response.status() == StatusCode::NOT_FOUND`

crates/verify/src/sourcify.rs
```diff
@@ -47,155 +64,317 @@ impl VerificationProvider for SourcifyVerificationProvider {
...
-            files.insert(import_entry, fs::read_to_string(&import)?);
-        }
+                // remove all incompatible settings
+                settings.sanitize(&context.compiler_version, SolcLanguage::Solidity);
+
+                input.settings = settings;
+
+                serde_json::to_value(&input).wrap_err("Failed to serialize standard json input")?
+            }
+            ContractLanguage::Vyper => {
```
grandizzy, line 273: :heart_eyes:

crates/verify/src/sourcify.rs
```diff
@@ -47,155 +64,317 @@ impl VerificationProvider for SourcifyVerificationProvider {
...
+            "{}:{}",
+            context
+                .target_path
+                .strip_prefix(context.project.root())
+                .unwrap_or(context.target_path.as_path())
+                .display(),
+            context.target_name
+        );
+
+        let compiler_version = if matches!(lang, ContractLanguage::Vyper) {
```
grandizzy: can we move this in match above and return both `std_json_input` and `compiler_version` or maybe the `Ok(SourcifyVerifyRequest)` directly

crates/verify/src/sourcify.rs
```diff
@@ -47,155 +64,317 @@ impl VerificationProvider for SourcifyVerificationProvider {
...
+            context
```
grandizzy: shouldn't `context.compiler_version` be enough here?

crates/verify/src/sourcify.rs
```diff
@@ -47,155 +64,317 @@ impl VerificationProvider for SourcifyVerificationProvider {
...
-                if let Some(ts) = &response.storage_timestamp {
-                    sh_println!("Contract source code already verified. Storage Timestamp: {ts}")?;
+    async fn is_contract_verified(&self, args: &VerifyArgs) -> Result<bool> {
+        let chain_id = args.etherscan.chain.unwrap_or_default().id();
+        let base_url = args.verifier.verifier_url.as_deref().unwrap_or(SOURCIFY_URL);
+        let url = format!("{}v2/contract/{}/{}", base_url, chain_id, args.address);
+
+        match reqwest::get(&url).await {
+            Ok(response) => {
+                if response.status() == 200 {
```
grandizzy: do we care about 200 only here or should we `if response.status().is_success()`?

crates/verify/src/sourcify.rs
manuelwedler: There are self-hosted Sourcify instances that should be supported. So it makes sense to leave the option to provide a custom url.

We list them on our website, see "self-hosting" on https://sourcify.dev

crates/verify/src/sourcify.rs
manuelwedler: In the way the Sourcify api is implemented, the server will never return a 200 on this API call. I think it's fine to just expect 202.

crates/verify/src/sourcify.rs
manuelwedler: `is_success()` seems better :+1:

crates/verify/src/sourcify.rs
manuelwedler: For Solidity, Sourcify requires the build metadata. So, calling `ensure_solc_build_metadata` seems necessary. 

For Vyper, I came to the conclusion that we actually should use `context.compiler_version` due to inconsistent artifact naming of Vyper releases.

manuelwedler:
@grandizzy I just addressed the feedback

grandizzy:
> @grandizzy I just addressed the feedback

@manuelwedler thank you, will check, could you please remove the unwraps to fix compilation errors, we made some updates in the way tests work

grandizzy (approved):
lgtm, thank you! @DaniPopes @zerosnacks if you want to have one more look before merge

PR #11204, opened 2025-08-04
feat(`invariants`): sample typed storage values

yash-atreya:
## Motivation

Closes #8116 

## Solution

- Adds an optional `StorageLayout` to `ContractData` and `TargetedContracts`
- Upon insertion of `state_changeset`, the `StorageLayout` is looked up to decide whether to insert it as a typed sample value OR raw value.
- This requires enabling `storageLayout` using `extra_output = ["storageLayout"]` in `foundry.toml` OR use the cli flag `--extra-output storageLayout`
- This is especially useful in cases where the function call does not return a value OR emit an event but storage changes are made. We are now able to decode these storage changes and add them to the subsequent fuzz inputs.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #11440, opened 2025-08-27
fix(coverage): contracts by artifact from linked contracts

### LINKED ISSUE #11432: Coverage is significantly reduced after upgrade to nightly version
### Component

Forge

- [x] Foundry
- [x] Foundryup

1.3.2-nightly Commit SHA: befca98d440847160a65e71597ff03943c804190

1.3.0

forge coverage --report lcov

### Operating System

Linux

### Describe the bug

Coverage of [our project](https://github.com/stabilitydao/stability-contracts/pull/364) was reduced on 80% after upgrading 1.3.2-stable => 1.3.2-nightly. The problem appears on CI that always uses nightly version but I was able to reproduce it locally also. All tests are passed successfully on both stable and nightly versions. The only visible difference is that nightly version generates lcov.info with 0 hits for most functions. My guess is that zero hits are set for all functions called in tests on forks. There are no such problems with tests on mocks. Lcov.info files generated by both foundry versions are [here](https://drive.google.com/file/d/1wbflhzy0-aJBMUOJew4xtG4O1C2OlWiX/view?usp=sharing). Would you please check?

grandizzy:
## Motivation
- closes #11432 
- regression introduced https://github.com/foundry-rs/foundry/pull/11204/files#diff-7cf15eff5ea54b62e8810f2f11d2285a2783dc88ceb867ca0618a73487f1df60R522-R525 we need to create contracts by artifact from linked contracts

## Solution
- introduce builder to create `ContractsByArtifact` from contracts iter and enrich with storage layouts (if available)

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/common/src/contracts.rs
+++ b/crates/common/src/contracts.rs
@@ -8,8 +8,9 @@ use eyre::{OptionExt, Result};
 use foundry_compilers::{
     ArtifactId, Project, ProjectCompileOutput,
     artifacts::{
-        BytecodeObject, CompactBytecode, CompactContractBytecode, CompactDeployedBytecode,
-        ConfigurableContractArtifact, ContractBytecodeSome, Offsets, StorageLayout,
+        BytecodeObject, CompactBytecode, CompactContractBytecode, CompactContractBytecodeCow,
+        CompactDeployedBytecode, ConfigurableContractArtifact, ContractBytecodeSome, Offsets,
+        StorageLayout,
     },
     utils::canonicalized,
 };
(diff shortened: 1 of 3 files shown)


grandizzy:
> lgtm!
> 
> probably a good reason to eventually work on adding coverage to the benchmarks

yeah, I think that's more about missing unit test for common coverage case, should be added now

zerosnacks (approved):

PR #11441, opened 2025-08-27
fix(anvil): make with_steps_tracing enable steps + mem/stack snapshots

MozirDmitriy:
- Align AnvilInspector::with_steps_tracing with its name and CLI/docs semantics.
- Switch from recording state diffs to enabling actual step recording and full context:
  - set_steps(true)
  - set_memory_snapshots(true)
  - set_stack_snapshots(StackSnapshotType::Full)
- Keep with_trace_printer on state diffs (unchanged).
- This matches how cheatcodes enable debug step recording and unblocks geth-style step traces.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/inspector.rs
+++ b/crates/anvil/src/eth/backend/mem/inspector.rs
@@ -7,7 +7,7 @@ use foundry_evm::{
     decode::decode_console_logs,
     inspectors::{LogCollector, TracingInspector},
     traces::{
-        CallTraceDecoder, SparsedTraceArena, TracingInspectorConfig, StackSnapshotType,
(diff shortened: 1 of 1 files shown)


crates/anvil/src/eth/backend/mem/inspector.rs
```diff
@@ -72,7 +73,12 @@ impl AnvilInspector {
 
     /// Enables steps recording for `Tracer`.
     pub fn with_steps_tracing(mut self) -> Self {
-        self.tracer = Some(TracingInspector::new(TracingInspectorConfig::all().with_state_diffs()));
```
DaniPopes, line 75: this is already `all` so this PR does nothing: https://docs.rs/revm-inspectors/latest/src/revm_inspectors/tracing/config.rs.html#83-95

Issue #10703, opened 2025-06-04
mockCall could lead to issues


lemunozm:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

forge Version: 1.2.2-nightly

### What version of Foundryup are you on?

forge Version: 1.2.2-nightly

### What command(s) is the bug in?

forge test

### Operating System

macOS (Apple Silicon)

### Describe the bug

If I do the following, it will revert:

```solidity
// SPDX-License-Identifier: BUSL-1.1
pragma solidity 0.8.28;

import "forge-std/Test.sol";

interface IExample {
    function foo(uint a) external;
}

contract MyTest is Test {
    function testFoo() public {
        IExample(address(1)).foo(42);
    }
}
```
because `address(1)` is not a contract.

Nevertheless, if I use `mockCall` like this:

```solidity
// SPDX-License-Identifier: BUSL-1.1
pragma solidity 0.8.28;

import "forge-std/Test.sol";

interface IExample {
    function foo(uint a) external;
}

contract MyTest is Test {
    function testFoo() public {
        vm.mockCall(address(1), "randomvalue", abi.encode());
        IExample(address(1)).foo(42);
    }
}
```

it will pass because it transforms `address(1)` to be a contract address, and calling any unknown method on it will not fail (EVM...).

This behavior is pretty obscure and makes tests pass when they shouldn't. I guess `mockCall` sets some contract under the hood; I think it must revert if `fallback()` is reached.

lemunozm:
Actually, this also reverts:

```solidity
// SPDX-License-Identifier: BUSL-1.1
pragma solidity 0.8.28;

import "forge-std/Test.sol";

interface IExample {
    function foo(uint a) external;
}

contract MyContract { }

contract MyTest is Test {
    function testFoo() public {
        IExample(address(new MyContract())).foo(42);
    }
}
```

So my real question is, why, if I use `mockCall()` on an address that is not a contract, the call passes, when it shouldn't?

grandizzy:
hey @lemunozm thanks for report! the reason why this happens is that we're etching a single byte onto the account in order to circumvent the `extcodesize`  check Solidity might perform, see below

https://github.com/foundry-rs/foundry/blob/dc43e2c1ab70ec111d3272572e1732008f973a06/crates/cheatcodes/src/evm/mock.rs#L211-L223

[...]

PR #11443, opened 2025-08-27
fix(cheatcodes): use make_acc_non_empty in mockCalls_1Call for consistency

aganisgash:
All other mock functions use `make_acc_non_empty` to etch a single byte onto empty accounts, which prevents Solidity's `extcodesize` check from failing when mocking calls to addresses without existing bytecode.

Makes `mockCalls_1Call` consistent with the other mock cheatcodes and ensures mocked calls work correctly in all scenarios.

### DIFF
--- a/crates/cheatcodes/src/evm/mock.rs
+++ b/crates/cheatcodes/src/evm/mock.rs
@@ -111,7 +111,7 @@ impl Cheatcode for mockCalls_0Call {
 impl Cheatcode for mockCalls_1Call {
     fn apply_stateful(&self, ccx: &mut CheatsCtxt) -> Result {
         let Self { callee, msgValue, data, returnData } = self;
-        ccx.ecx.journaled_state.load_account(*callee)?;
+        let _ = make_acc_non_empty(callee, ccx)?;
         mock_calls(ccx.state, callee, data, Some(msgValue), returnData, InstructionResult::Return);
         Ok(Default::default())
     }


crates/cheatcodes/src/evm/mock.rs
```diff
@@ -111,7 +111,7 @@ impl Cheatcode for mockCalls_0Call {
 impl Cheatcode for mockCalls_1Call {
     fn apply_stateful(&self, ccx: &mut CheatsCtxt) -> Result {
         let Self { callee, msgValue, data, returnData } = self;
-        ccx.ecx.journaled_state.load_account(*callee)?;
+        let _ = make_acc_non_empty(callee, ccx)?;
```
zerosnacks, line 116:
```suggestion
        let _ = make_acc_non_empty(callee, ccx)?;

```

zerosnacks:
This makes sense to me, I am not sure why this was treated as a special case

For context:

https://github.com/foundry-rs/foundry/blob/6e6341b8f6518f00b64c439714c91e14db2955cf/crates/cheatcodes/src/evm/mock.rs#L211-L223

aganisgash:
> This makes sense to me, I am not sure why this was treated as a special case
> 
> For context:
> 
> https://github.com/foundry-rs/foundry/blob/6e6341b8f6518f00b64c439714c91e14db2955cf/crates/cheatcodes/src/evm/mock.rs#L211-L223

Maybe just a copy-paste oversight. All the other mock functions follow the same pattern with `make_acc_non_empty`, so this one likely got missed when the code was initially written.

grandizzy:
I think we want a sig to bypass that at some point but subject to discussion https://github.com/foundry-rs/foundry/issues/10703#issuecomment-2944263238

grandizzy:
I think that's fine but could be breaking change for some projects assuming that will revert

zerosnacks:
Hi @aganisgash would you mind also making this change for `mockCall_1Call` and `mockCall_3Call`?

https://github.com/foundry-rs/foundry/blob/92f1c6462c40e8f0657b70c8a0af90ca688d8c5f/crates/cheatcodes/src/evm/mock.rs#L62

https://github.com/foundry-rs/foundry/blob/92f1c6462c40e8f0657b70c8a0af90ca688d8c5f/crates/cheatcodes/src/evm/mock.rs#L88

This aligns the behavior for all `mockCall*`

aganisgash:
@zerosnacks Sure, all fixed

Issue #11393, opened 2025-08-21
feat(forge-lint): return non-zero exit code when lint fails


cruzdanilo:
### Component

Forge

### Describe the feature you would like

forge lint is always returning a success exit code (0), even when there are issues. this makes it harder to use it on ci.

0xClandestine:
noticed this as well

srdtrk:
It would be good to be able to adjust what severity of lint leads to non-zero exit code. Some may want to return a zero exit code if the lint only finds info and low severity.

wjmelements:
> this makes it harder to use it on ci.

I am working around like this:

```sh
! (forge lint 2>&1 | grep "")
```

`grep` returns failure on empty output. `!` inverts this so it succeeds only on empty output.

PR #11445, opened 2025-08-27
feat(lint): fail on configured diagnostic level

0xrusowsky:
## Motivation
ref:
 - https://github.com/foundry-rs/foundry/issues/11393

## Solution
- introduce a new `deny` config flag
	```rs
	/// Diagnostic level (minimum) at which the process should finish with a non-zero exit.
	pub enum DenyLevel {
	    /// Always exit with zero code.
	    #[default]
    	Never,
	    /// Exit with a non-zero code if any warnings are found.
    	Warnings,
	    /// Exit with a non-zero code if any notes or warnings are found.
    	Notes,
	}
	```
- modify the `Linter` trait so that `fn lint()` returns an error based on the `deny` flag and the emitted diagnostics.
	```rs
	fn lint(&self, input: &[PathBuf], deny: DenyLevel, compiler: &mut Compiler) -> Result<()>
	```

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [ ] **Breaking changes: NOT YET**. However, introduces a user warning explaining that we decided to remove the `--deny-warnings` flag in favor of a new `--deny` flag that takes the `DenyLevel`.

## TODO:
- [ ] remove solar patch when 0.1.7 is released --> needed for https://github.com/paradigmxyz/solar/pull/447
- [ ] update docs to reflect `--deny-warnings` deprecation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -58,9 +58,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy-chains"
-version = "0.2.11"
+version = "0.2.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8e4e355f312b270bca5144af5f003e7d238037e47a818766f9107f966cbecf52"
+checksum = "f3008b4f680adca5a81fad5f6cdbb561cca0cee7e97050756c2c1f3e41d2103c"
 dependencies = [
  "alloy-primitives",
  "num_enum",
@@ -2969,7 +2969,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "117725a109d387c937a1533ce01b450cbde6b88abceea8473c4d7a85853cda3c"
 dependencies = [
  "lazy_static",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -3569,7 +3569,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.61.0",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -3828,7 +3828,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -3943,7 +3943,7 @@ checksum = "0ce92ff622d6dadf7349484f42c93271a0d49b7cc4d466a936405bacbe10aa78"
 dependencies = [
  "cfg-if",
  "rustix 1.1.2",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -5931,7 +5931,7 @@ checksum = "e04d7f318608d35d4b61ddd75cbdaee86b023ebe2bd5a66ee0915f0bf93095a9"
 dependencies = [
  "hermit-abi",
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -5985,7 +5985,7 @@ dependencies = [
  "portable-atomic",
  "portable-atomic-util",
  "serde",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -8434,7 +8434,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.4.15",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -8447,7 +8447,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.11.0",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -9679,7 +9679,7 @@ dependencies = [
  "getrandom 0.3.3",
  "once_cell",
  "rustix 1.1.2",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -9699,7 +9699,7 @@ version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2111ef44dae28680ae9752bb89409e7310ca33a8c621ebe7b106cf5c928b3ac0"
 dependencies = [
- "windows-sys 0.61.0",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -10943,7 +10943,7 @@ version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2a7b1c03c876122aa43f3020e6c3c3ee5c05081c9a00739faf7503aeba10d22"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]


crates/forge/src/cmd/lint.rs
```diff
@@ -100,6 +108,12 @@ impl LintArgs {
             None => config.lint.severity.clone(),
         };
 
+        // Override default fail_on level config with user-defined severity
```
grandizzy: can simplify `self.fail_on.unwrap_or(config.lint.fail_on)` and move below inline

crates/forge/tests/cli/lint.rs
```diff
@@ -530,6 +531,54 @@ Compiler run successful!
...
+    // [OUTPUT: WARN + NOTE]
+    prj.update_config(|config| {
+        config.lint.fail_on = foundry_config::lint::FailOn::Warning;
+        config.lint.severity = Default::default();
+    });
+    cmd.forge_fuse().arg("lint").assert_failure();
+
+    // [OUTPUT: WARN + NOTE]
+    prj.update_config(|config| {
+        config.lint.fail_on = foundry_config::lint::FailOn::Never;
```
grandizzy: isn't this the default, so we can just move first and only `cmd.forge_fuse().arg("lint").assert_success();`?

crates/lint/src/sol/mod.rs
```diff
@@ -257,6 +263,37 @@ impl<'a> Linter for SolidityLinter<'a> {
                 }
             });
         });
+
+        // Handle diagnostics and fail if necessary.
+        let s = |count| if count == 1 { "" } else { "s" };
```
grandizzy: nit: IMO would be OK to just display
```
                "aborting due to 1 linter warning(s) ",
```

```
                "1 linter warning(s) and 1 note(s)",
```

and avoid extra logic to count / format, but happy to keep it this way too

grandizzy (commented):
makes sense, left couple of minor nits

grandizzy:
@0xrusowsky should we also add tests for `forge build` with lint set to fail and make sure build pass / fail?

grandizzy (approved):
lgtm, pending others review / solar bump and update

crates/forge/src/cmd/bind_json.rs
```diff
@@ -193,7 +193,7 @@ impl BindJsonArgs {
                     let def = hir.strukt(id);
                     let source = hir.source(def.source);
 
-                    if !target_files.contains(&source.file.stable_id) {
+                    if !target_files.contains(&source.file) {
```
zerosnacks, line 196: to double check, the changes in this file seem unrelated to the PR - required because of Solar bump?

0xrusowsky: yes, exactly

these changes are necessary because of this commit: https://github.com/paradigmxyz/solar/commit/011570cdfbb317943d1aff51e9918517ef41378e

DaniPopes:
we also have --deny-warnings, do we also integrate this?

0xrusowsky:
> we also have --deny-warnings, do we also integrate this?

sure. should it fail with `Level::Note` too? or just `Level:Warning`?

DaniPopes:
warnings. could we also rename to keep it consistent? --deny warnings/notes

crates/config/src/lib.rs
```diff
@@ -1498,6 +1498,20 @@ impl Config {
...
+    pub fn get_deny_level(&self) -> DenyLevel {
+        match self.lint.deny {
+            DenyLevel::Notes => DenyLevel::Notes,
+            other => {
+                if self.deny_warnings {
+                    DenyLevel::Warnings
+                } else {
+                    other
+                }
+            }
```
onbjerg: let's deprecate `--deny-warnings` as well? just output a note saying you should be using `--deny warnings` instead

crates/forge/tests/cli/cmd.rs
```diff
@@ -1250,7 +1250,7 @@ Warning: SPDX license identifier not provided in source file. Before publishing,
 forgetest!(can_fail_compile_with_warnings, |prj, cmd| {
     prj.update_config(|config| {
         config.ignored_error_codes = vec![];
-        config.deny_warnings = false;
+        config.deny = false.into();
```
DaniPopes: I'd rather not have this impl and just spell the enum out

crates/config/src/lib.rs
```diff
@@ -664,6 +763,19 @@ impl Config {
         let mut config = figment.extract::<Self>().map_err(ExtractConfigError::new)?;
         config.profile = figment.profile().clone();
 
+        // Handle deprecated `--deny-warnings` CLI flag
+        if config.deny_warnings {
```
DaniPopes: should just add an entry to DEPRECATIONS https://github.com/foundry-rs/foundry/blob/a131c31255c3f87c9828eb43bec3d80d2b23fe62/crates/config/src/lib.rs#L569

crates/config/README.md
```diff
@@ -107,7 +107,7 @@ etherscan_api_key = "YOURETHERSCANAPIKEY"
 # additional warnings can be added using their numeric error code: ["license", 1337]
 ignored_error_codes = ["license", "code-size"]
 ignored_warnings_from = ["path_to_ignore"]
-deny_warnings = false
+deny = false
```
DaniPopes: "never", can also add a comment on this now since it's not very self explanatory

crates/config/src/lib.rs
```diff
@@ -560,13 +563,109 @@ pub struct Config {
...
+        struct DenyLevelVisitor;
+
+        impl<'de> de::Visitor<'de> for DenyLevelVisitor {
+            type Value = DenyLevel;
+
+            fn expecting(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                formatter.write_str("a boolean or a string like 'never', 'warnings', or 'notes'")
+            }
+
+            fn visit_bool<E>(self, value: bool) -> Result<Self::Value, E>
```
DaniPopes, line 597: remove this too `deny = true` doesn't make much sense on its own

0xrusowsky: i had to left this one for the backwards compatibility, otherwise it was failing to convert from `deny_warnings` to `deny`

crates/config/src/lib.rs
```diff
@@ -665,7 +665,7 @@ pub const STANDALONE_FALLBACK_SECTIONS: &[(&str, &str)] = &[("invariant", "fuzz"
 ///
 /// See [Warning::DeprecatedKey]
 pub const DEPRECATIONS: &[(&str, &str)] =
-    &[("cancun", "evm_version = Cancun"), ("deny_warnings", "deny")];
+    &[("cancun", "evm_version = Cancun"), ("deny_warnings = true", "deny = warnings")];
```
DaniPopes: please also add a test

```suggestion
    &[("cancun", "evm_version = Cancun"), ("deny_warnings", "deny = warnings")];
```

crates/cli/src/opts/build/core.rs
```diff
@@ -48,9 +48,27 @@ pub struct BuildOpts {
...
-    /// Warnings will trigger a compiler error
-    #[arg(long, help_heading = "Compiler options")]
+    /// A compiler error will be triggered at the specified diagnostic level.
+    ///
+    /// Replaces the depracated `--deny_warnings` flag.
+    /// Accepts boolean values for backwards compatibility.
+    ///
+    /// Possible values:
+    ///  - `never` (`false`): Do not treat any diagnostics as errors.
+    ///  - `warnings` (`true`): Treat warnings as errors.
```
onbjerg: let's remove the false/true? it looks like you can pass those values in as an end-user but you cant

also add `never` just for completeness

crates/cli/src/opts/build/core.rs
```diff
@@ -48,9 +48,27 @@ pub struct BuildOpts {
     #[serde(skip_serializing_if = "Vec::is_empty")]
     pub ignored_error_codes: Vec<u64>,
 
-    /// Warnings will trigger a compiler error
-    #[arg(long, help_heading = "Compiler options")]
+    /// A compiler error will be triggered at the specified diagnostic level.
+    ///
+    /// Replaces the depracated `--deny_warnings` flag.
```
onbjerg:
```suggestion
    /// Replaces the depracated `--deny-warnings` flag.
```

crates/cli/src/opts/build/core.rs
```diff
@@ -48,9 +48,27 @@ pub struct BuildOpts {
     #[serde(skip_serializing_if = "Vec::is_empty")]
     pub ignored_error_codes: Vec<u64>,
 
-    /// Warnings will trigger a compiler error
-    #[arg(long, help_heading = "Compiler options")]
+    /// A compiler error will be triggered at the specified diagnostic level.
+    ///
+    /// Replaces the deprecated `--deny-warnings` flag.
+    /// Accepts boolean values for backwards compatibility.
```
onbjerg: this is not true

```
> forge lint --deny false
error: invalid value 'false' for '--deny <LEVEL>'
  [possible values: never, warnings, notes]

For more information, try '--help'.

oliver@dagr /m/v/P/f/forge-std
 (master)
 [2]> forge lint --deny true
error: invalid value 'true' for '--deny <LEVEL>'
  [possible values: never, warnings, notes]

For more information, try '--help'.
```

onbjerg (changes requested):

crates/cli/src/opts/build/core.rs
0xrusowsky: sorry, forgot to remove the cmnt

initially was true but Dani asked me to not accept bools

onbjerg (approved):

PR #11452, opened 2025-08-28
ci: drop force_orphan

DaniPopes:
`force_orphan` makes it so that the branch always gets reset.

This job runs on-push, so every `git pull` on the repository will reset the `gh-pages` branch and pull everything from the remote, currently about ~20,000 files and ~20MiB, which is slow and useless. Let's try removing this option to see if it fixes this problem.


0xrusowsky (approved):
let's try

PR #11454, opened 2025-08-28
ci: use actions/* directly for deploying rustdocs

DaniPopes:
`peaceiris/actions-gh-pages` pushes to the `gh-pages` branch, GitHub will detect changes to that branch and then trigger a deployment by uploading the branch. This is the "legacy"/"classic" way of using GitHub Pages.

Switch to the newer way of deploying from GitHub actions with the official actions/* workflows, bypassing the need to have a `gh-pages` branch at all by uploading the artifacts to GitHub Pages directly, improving `git pull` times significantly.


yash-atreya (approved):

0xrusowsky (approved):
makes sense

zerosnacks (approved):

Issue #11456, opened 2025-08-28
Deduplicate solar Parser / Session instantiation


DaniPopes:
Re-use existing `Compiler` from `ProjectCompileOutput`.

Demo here https://github.com/foundry-rs/foundry/pull/11371/commits/e9283ef17d16fc0896dfd23b6bc1b3acac741479.

PR #11458, opened 2025-08-28
chore(forge eip712): use solar from compiler output

DaniPopes:
Towards https://github.com/foundry-rs/foundry/issues/11456.

### DIFF
--- a/crates/forge/src/cmd/eip712.rs
+++ b/crates/forge/src/cmd/eip712.rs
@@ -83,6 +83,7 @@ impl Eip712Args {
             Ok(())
         })?;
 
+        // `compiler.sess()` inside of `ProjectCompileOutput` is built with `with_buffer_emitter`.
         let diags = compiler.sess().dcx.emitted_diagnostics().unwrap();
         if compiler.sess().dcx.has_errors().is_err() {
             eyre::bail!("{diags}");
--- a/crates/forge/src/cmd/eip712.rs
+++ b/crates/forge/src/cmd/eip712.rs
@@ -1,25 +1,19 @@
 use alloy_primitives::{B256, keccak256};
 use clap::{Parser, ValueHint};
 use eyre::Result;
-use foundry_cli::{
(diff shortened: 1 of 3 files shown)


crates/forge/src/cmd/eip712.rs
```diff
@@ -97,7 +83,12 @@ impl Eip712Args {
             Ok(())
         })?;
 
-        eyre::ensure!(compiler.sess().dcx.has_errors().is_ok(), "errors occurred");
+        let diags = compiler.sess().dcx.emitted_diagnostics().unwrap();
```
0xrusowsky, line 87: maybe mention that this is safe cause it uses a human-readable emitter?

0xrusowsky (approved):
clean 🤝

0xrusowsky:
@DaniPopes looks like tests need to be updated (not a big deal), although not sure why that wasn't an issue before

DaniPopes:
We didn't have a project before. This does call solc but with the minimum work possible

0xrusowsky:
oh sure, makes sense...

0xrusowsky (approved):

crates/forge/src/cmd/eip712.rs
```diff
@@ -97,7 +83,13 @@ impl Eip712Args {
             Ok(())
         })?;
 
-        eyre::ensure!(compiler.sess().dcx.has_errors().is_ok(), "errors occurred");
+        // `compiler.sess()` inside of `ProjectCompileOutput` is built with `with_buffer_emitter`.
```
onbjerg, line 86:
```suggestion
        // SAFETY: `compiler.sess()` inside of `ProjectCompileOutput` is built with `with_buffer_emitter`.
```

i find the preamble useful for context

DaniPopes: SAFETY should only be used for `unsafe`

PR #7334, opened 2024-03-07
feat(test): only compile files needed for tests

klkvr:
## Motivation

Compile contracts requesting only abi output and filter out abis which match given filters or just start with `test`/`invariant`

This allowed to move some checks to be performed earlier ("No tests match the provided pattern")

This might be breaking for some codebases using `vm.getDeployedCode` as we won't recompile contracts which are not appearing as dependecy of test contract, however if contract being fetched through `vm.getDeployedCode` is imported from test file, then it will still be recompiled and I believe that it's a common pattern used, for example, in optimism codebase (https://github.com/ethereum-optimism/optimism/blob/53573e0ea6a807a125784cc5c7df07cbb4dbe3bc/packages/contracts-bedrock/scripts/Deployer.sol#L25)

Closes #7045 

cc @mds1

PR #11464, opened 2025-08-28
refactor: `get_sources_to_compile`

DaniPopes:
- `get_sources_to_compile` currently is always called, even when there are no filters to apply to the input. This is wasteful because it will compile all sources to ABI, filter out nothing, then recompile everything again for bytecode regardless
- we want to filter out non-matching test contracts, instead of filtering out everything and then adding back sources. This misses scripts and other stray files
- Added in https://github.com/foundry-rs/foundry/pull/7334, this changed the exit code to fail when there are no tests, which I don't think is right and doesn't match other test frameworks
- move the "no matching test" check back to `run_tests` to simplify the function and so we can also handle the case when there are no tests at all in the project


crates/forge/tests/cli/test_optimizer.rs
```diff
@@ -41,7 +41,7 @@ forgetest_init!(toggle_invalidate_cache_on_test, |prj, cmd| {
     // All files are built with optimized tests.
     cmd.args(["test"]).with_no_redact().assert_success().stdout_eq(str![[r#"
 ...
-Compiling 21 files with [..]
+Compiling 23 files with [..]
```
DaniPopes, line 44: the extra 2 files are script/Counter.s.sol and forge-std/Script.sol

onbjerg (approved):

crates/forge/tests/cli/test_optimizer.rs
klkvr: the idea for `get_sources_to_compile` was that we don't want to compile e.g scripts for `forge test`, is this no longer the case?

DaniPopes: My hunch was that in practice it ends up being slower to compile another time before to filter out unneeded files than to just run the pipeline normally; I can revisit further by rewriting to solar

PR #11469, opened 2025-08-29
fix(forge): persist corpus per test suite

### LINKED ISSUE #11453: Invariant test fails due to stale or conflicting corpus
### Component

Forge

- [x] Foundry
- [x] Foundryup

v1.3.2

v1.3.0

forge test

### Operating System

Linux

### Describe the bug

When multiple invariant test contracts share the same invariant names and use the corpus, the run for the second contract fails with weird errors related to missing selectors.

After investigating, I made the following observations:
- Bug does not happen with v1.2.3
- Bug happens with 1.3.0-1.3.2
- Bug happens only when corpus is enabled
- Bug happens only for invariants which have the same name in multiple contracts

Reproduction (make sure foundry version is 1.3.x):

```bash
git clone https://github.com/SmarDex-Ecosystem/usdn-contracts.git
cd usdn-contracts
git checkout 6596df75342daaafbe84db262db173d7aaff55aa
forge soldeer install
printf "\n[profile.repro.invariant]\ncorpus_dir = \".corpus\"\nruns = 100\n" >> foundry.toml
FOUNDRY_PROFILE=repro forge test --mc TestUsdnInvariants -vvv
FOUNDRY_PROFILE=repro forge test --mc TestWusdnInvariants -vvv
```

The first contract's run works fine, but the second fails.

Result: `[Revert] unrecognized function selector 0x74e228f0 for contract 0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f, which has no fallback function.`

I suspect this is because the corpus' hierarchy only includes the invariant function name, and two different contracts thus share the same folder. When re-loading the existing corpus in memory, some files have nothing to do with the current test being executed, which must confuse the system.

grandizzy:
## Motivation
- closes #11453 
- persist fuzz and invariant corpus per contract

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/forge/tests/it/invariant.rs
+++ b/crates/forge/tests/it/invariant.rs
@@ -1598,8 +1598,7 @@ contract Counter2Test is Test {
     }
 }
    "#,
-    )
-    .unwrap();
+    );
 
     cmd.args(["test"]).assert_success().stdout_eq(str![[r#"
 ...


yash-atreya (dismissed):
lgtm!

beeb:
Much better! I could see potential for conflict if more than one test contract shares its name with another, which is theoretically possible. But the alternative would be to like use the full relative path or hash the path somehow. Maybe for a future improvement!

Issue #11460, opened 2025-08-28
`forge lint --json` has ANSI escape sequences in messages


onbjerg:
### Component

Forge

### Have you ensured that all of these are up to date?

- [ ] Foundry
- [ ] Foundryup

### What version of Foundry are you on?

1.3.2-nightly

### What command(s) is the bug in?

forge lint --json

### Operating System

Linux

### Describe the bug

1. Clone e.g. https://github.com/ithacaxyz/account
2. Run `forge lint --json`
3. Notice that some messages have ANSI escape sequences

As an example:

```json
{
  "$message_type": "diag",
  "message": "typecasts that can truncate values should be checked",
  "code": {
    "code": "unsafe-typecast",
    "explanation": null
  },
  "level": "warning",
  "spans": [
    {
      "file_name": "/media/vault/Projects/ithacaxyz/account/test/GuardedExecutor.t.sol",
      "byte_start": 3800,
      "byte_end": 3815,
      "line_start": 94,
      "line_end": 94,
      "column_start": 30,
      "column_end": 45,
      "is_primary": true,
      "text": [
        {
          "text": "            assertEq(address(bytes20(packed)), target2);",
          "highlight_start": 30,
          "highlight_end": 45
        }
      ],
      "label": null
    }
  ],
  "children": [
    {
      "message": "Consider disabling this lint if you're certain the cast is safe:\n\n// casting to 'bytes20' is safe because [explain why]\n// forge-lint: disable-next-line(unsafe-typecast)\n\n",
      "code": null,
      "level": "note",
      "spans": [],
      "children": [],
      "rendered": null
    },
    {
      "message": "\u001b]8;;https://book.getfoundry.sh/reference/forge/forge-lint#unsafe-typecast\u001b\\https://book.getfoundry.sh/reference/forge/forge-lint#unsafe-typecast\u001b]8;;\u001b\\",
      "code": null,
      "level": "help",
      "spans": [],
      "children": [],
      "rendered": null
    }
  ],
[...]

PR #11470, opened 2025-08-29
fix(lint): avoid ANSI chars with JSON emitter

0xrusowsky:
## Motivation
ref:
- https://github.com/foundry-rs/foundry/issues/11460

## Solution

track emitter kind in `LintContext` and avoid ANSI chars when necessary

### DIFF
--- a/crates/lint/src/linter/mod.rs
+++ b/crates/lint/src/linter/mod.rs
@@ -54,6 +54,7 @@ pub trait Lint {
 pub struct LintContext<'s, 'c> {
     sess: &'s Session,
     with_description: bool,
+    with_json_emitter: bool,
     pub config: LinterConfig<'c>,
     active_lints: Vec<&'static str>,
 }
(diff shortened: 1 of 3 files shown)


0xrusowsky:
@onbjerg can u confirm that this stderr output is the one u expect?

crates/forge/tests/cli/lint.rs
```diff
@@ -530,6 +530,47 @@ Compiler run successful!
...
+            severity: vec![LintSeverity::CodeSize],
+            exclude_lints: vec![],
+            ignore: vec![],
+            lint_on_build: true,
+            ..Default::default()
+        };
+    });
+
+    // should produce clean JSON without ANSI escape sequences (for the url nor the snippets)
+    cmd.arg("lint").arg("--json").assert_success().stderr_eq(str![[r#"
```
DaniPopes: ~~assert_json_stdout~~ need to create assert_json_stderr, you can then also format the json so its readable and for future diffs

DaniPopes (approved):

PR #11472, opened 2025-08-29
fix(forge): use stipend and refunded in simulate gas

### LINKED ISSUE #10968: Transaction failure at first time, but succeed in second time with forge script.
### Component

Forge

- [x] Foundry
- [x] Foundryup

1.2.3-stable

1.1.0

_No response_

### Operating System

Linux

### Describe the bug

Actually I am running some command, but failing at first time, succeed in second time.

`https://github.com/Lay3rLabs/wavs-middleware/tree/169-script-to-deregister-operator`

I am running this command, which only passes in second time.
```
forge script script/eigenlayer/ecdsa/WavsDeregisterOperator.s.sol -vvv --rpc-url http://localhost:8545 --private-key "$OPERATOR_KEY" --broadcast
```
First output (Fail)
```
{
  "transactions": [
    {
      "hash": "0x548a129dc738aa4c89003d9448b7480eba71ede5556e6a3d21dcd2ef6732a05a",
      "transactionType": "CALL",
      "contractName": null,
      "contractAddress": "0xfe4eae5d84412b70b1f04b3f78351a654d28da25",
      "function": "deregisterOperator()",
      "arguments": [],
      "transaction": {
        "from": "0x7e8d0ec39584da6aa0d18eda9dc6a3a4870bc325",
        "to": "0xfe4eae5d84412b70b1f04b3f78351a654d28da25",
        "gas": "0x1f3ea",
        "value": "0x0",
        "input": "0x857dc190",
        "nonce": "0x6",
        "chainId": "0x4268"
      },
      "additionalContracts": [],
      "isFixedGasLimit": false
    }
  ],
  "receipts": [],
  "libraries": [],
  "pending": [],
  "returns": {},
  "timestamp": 1752026387,
  "chain": 17000,
  "commit": "c840f74"
}
```
Second output (success)
```
{
  "transactions": [
    {
      "hash": "0x6c862df83e0cd4953ea36b1616fc929ad10b248d0afad971959464da386a5c96",
      "transactionType": "CALL",
      "contractName": null,
      "contractAddress": "0xfe4eae5d84412b70b1f04b3f78351a654d28da25",
      "function": "deregisterOperator()",
      "arguments": [],
      "transaction": {
        "from": "0x7e8d0ec39584da6aa0d18eda9dc6a3a4870bc325",
        "to": "0xfe4eae5d84412b70b1f04b3f78351a654d28da25",
        "gas": "0x35df9",
        "value": "0x0",
        "input": "0x857dc190",
        "nonce": "0x7",
        "chainId": "0x4268"
      },
      "additionalContracts": [],
      "isFixedGasLimit": false
    }
  ],
  "receipts": [
    {
      "status": "0x1",
      "cumulativeGasUsed": "0x2700f",
      "logs": [
        {
          "address": "0xfe4eae5d84412b70b1f04b3f78351a654d28da25",
          "topics": [
            "0x88770dc862e47a7ed586907857eb1b75e4c5ffc8b707c7ee10eb74d6885fe594",
            "0x0000000000000000000000007e8d0ec39584da6aa0d18eda9dc6a3a4870bc325"
          ],
          "data": "0x000000000000000000000000000000000000000000000000000059acebaf51550000000000000000000000000000000000000000000000000000000000000000",
          "blockHash": "0x5ca3417f8c8e7f246c08b45e13ad0d3158a228117af214093d52056b6eb4b0d1",
          "blockNumber": "0x3f1096",
          "blockTimestamp": "0x686dcd0e",
          "transactionHash": "0x6c862df83e0cd4953ea36b1616fc929ad10b248d0afad971959464da386a5c96",
          "transactionIndex": "0x0",
          "logIndex": "0x0",
          "removed": false
        },
        {
          "address": "0xfe4eae5d84412b70b1f04b3f78351a654d28da25",
          "topics": [
            "0x86dcf86b12dfeedea74ae9300dbdaa193bcce5809369c8177ea2f4eaaa65729b"
          ],
          "data": "0x000000000000000000000000000000000000000000000000000059acebaf51550000000000000000000000000000000000000000000000000000000000000000",
          "blockHash": "0x5ca3417f8c8e7f246c08b45e13ad0d3158a228117af214093d52056b6eb4b0d1",
          "blockNumber": "0x3f1096",
          "blockTimestamp": "0x686dcd0e",
          "transactionHash": "0x6c862df83e0cd4953ea36b1616fc929ad10b248d0afad971959464da386a5c96",
          "transactionIndex": "0x0",
          "logIndex": "0x1",
          "removed": false
        },
        {
          "address": "0x055733000064333caddbc92763c58bf0192ffebf",
          "topics": [
            "0xf0952b1c65271d819d39983d2abb044b9cace59bcc4d4dd389f586ebdcb15b41",
            "0x0000000000000000000000007e8d0ec39584da6aa0d18eda9dc6a3a4870bc325",
            "0x0000000000000000000000007078df561cf252bdda2bfc987dabb476dec31b17"
          ],
          "data": "0x0000000000000000000000000000000000000000000000000000000000000000",
          "blockHash": "0x5ca3417f8c8e7f246c08b45e13ad0d3158a228117af214093d52056b6eb4b0d1",
          "blockNumber": "0x3f1096",
          "blockTimestamp": "0x686dcd0e",
          "transactionHash": "0x6c862df83e0cd4953ea36b1616fc929ad10b248d0afad971959464da386a5c96",
          "transactionIndex": "0x0",
          "logIndex": "0x2",
          "removed": false
        },
        {
          "address": "0xfe4eae5d84412b70b1f04b3f78351a654d28da25",
          "topics": [
            "0x31e0adfec71bccee37b6e83a90c2fedb17d8f1693fee863c4771e7bfe2aed580",
            "0x0000000000000000000000007e8d0ec39584da6aa0d18eda9dc6a3a4870bc325",
            "0x0000000000000000000000007078df561cf252bdda2bfc987dabb476dec31b17"
          ],
          "data": "0x",
          "blockHash": "0x5ca3417f8c8e7f246c08b45e13ad0d3158a228117af214093d52056b6eb4b0d1",
          "blockNumber": "0x3f1096",
          "blockTimestamp": "0x686dcd0e",
          "transactionHash": "0x6c862df83e0cd4953ea36b1616fc929ad10b248d0afad971959464da386a5c96",
          "transactionIndex": "0x0",
          "logIndex": "0x3",
          "removed": false
        }
      ],
      "logsBloom": "0x00000000000000000000000000000000000000000020000000000000010000000000002000080000000000000000000020000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000000000000000800000000000008000000000000000000000000000000000400000000000000000000004000000000000000000001000000000000000000000801000080000000000000000000000000000200000000000000000000000000080000100000000000000000000000004400000000000000000000000000000000000008000000000000080000000004000800000000000000000000020000000000000",
      "type": "0x2",
      "transactionHash": "0x6c862df83e0cd4953ea36b1616fc929ad10b248d0afad971959464da386a5c96",
      "transactionIndex": "0x0",
      "blockHash": "0x5ca3417f8c8e7f246c08b45e13ad0d3158a228117af214093d52056b6eb4b0d1",
      "blockNumber": "0x3f1096",
      "gasUsed": "0x2700f",
      "effectiveGasPrice": "0xf4248",
      "blobGasPrice": "0x1",
      "from": "0x7e8d0ec39584da6aa0d18eda9dc6a3a4870bc325",
      "to": "0xfe4eae5d84412b70b1f04b3f78351a654d28da25",
      "contractAddress": null
    }
  ],
  "libraries": [],
  "pending": [],
  "returns": {},
  "timestamp": 1752026394,
  "chain": 17000,
  "commit": "c840f74"
}
```

To run this, need to deploy and register operator first.
Please check README, or I can share the script from build image to registering to reproduce this error.

grandizzy:
## Motivation
- closes #10968 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/script/src/runner.rs
```diff
@@ -358,7 +358,7 @@ impl ScriptRunner {
         calldata: &Bytes,
         value: U256,
     ) -> Result<u64> {
-        let mut gas_used = res.gas_used;
+        let mut gas_used = res.gas_used + res.gas_refunded + res.stipend;
```
zerosnacks, line 361: I think I understand the reasoning why we would want to include this in the calculation but would you mind expanding on it?

grandizzy: if I correctly get this, when we simulate script we run the txes and use `res.gas_used` as gas limit if we broadcast them.

https://github.com/foundry-rs/foundry/blob/2b56727a7c663050296d93297b37b172b547a4d9/crates/script/src/runner.rs#L302-L323

However `gas_used` from call could be too low for the limit (as it doesn't include refunds and stipend). @klkvr could you pls confirm if that makes sense?

(Maybe could also be a better logic here per tx type or so)

onbjerg: does this affect simulation?

grandizzy: it is only adding gas to tx metadata / run-latest.json that is used when broadcasting. Also this comment for create txes that seems to be inline with the assumption that gas used recorded doesn't contain stipend:

https://github.com/foundry-rs/foundry/blob/56aadcb20c7edfe2c17525df1a5a7c6e5ffe7a5a/crates/script/src/runner.rs#L270-L281

onbjerg: sg

onbjerg (approved):

crates/script/src/runner.rs
grandizzy: another way (which could make more sense) is to set the lower bound of binary search

https://github.com/foundry-rs/foundry/blob/56aadcb20c7edfe2c17525df1a5a7c6e5ffe7a5a/crates/script/src/runner.rs#L367

to gas used + stipend + refunded, this also makes tx in issue pass in first try

grandizzy: made some more tests with setting different bounds for the binary search as per @klkvr request
The failing tx has `gas used: 92655`, `gas refunded: 14400` and `stipend: 21064`

| bound | highest limit | lowest limit | calculated gas |
|----------|----------|----------| ----------|
| `gas_limit = res.gas_used + res.gas_refunded + res.stipend`  | 384357     | 128119     | 136126 (OK)   |
| `lowest_gas_limit = res.gas_used + res.gas_refunded + res.stipend`  | 277965     | 128119     | 137484 (OK)   |
| `highest_gas_limit = res.gas_used + res.gas_refunded + res.stipend`  | 128119     | 92655     | 101521 (FAIL)   |
| `highest_gas_limit = (res.gas_used + res.gas_refunded + res.stipend) * 3;`  | 384357     | 92655     | 101770 (FAIL)   |

so it looks like the lowest bound here is the one important here (therefore think `lowest_gas_limit = res.gas_used + res.gas_refunded + res.stipend`)

0xrusowsky: maybe overthinking, but imo would be helpful to either document or just link this thread for future reference

mattsse: I love binary search gas estimation 👍
looks like we have this logic in 3 places, script,anvil,reth

need to think how we could unify this in alloy-evm maybe so that this can be reused and we only need to maintain this in one place -.-

mattsse, line 361: we should also do here:

```
 let optimistic_gas_limit = (gas_used + gas_refund + CALL_STIPEND_GAS) * 64 / 63;
```

I wonder why this fn can end up returning too low gas in the first place

ref

https://github.com/ethereum/go-ethereum/blob/a5a4fa7032bb248f5a7c40f4e8df2b131c4186a4/eth/gasestimator/gasestimator.go#L135

and

https://github.com/paradigmxyz/reth/blob/339f18c48f829a2f2fca440fb255995e23c7161f/crates/rpc/rpc-eth-api/src/helpers/estimate.rs#L192-L198

mattsse (commented):
smol suggestion

gas estimation 👎

crates/script/src/runner.rs
klkvr: I don't think adding `stipend` makes perfect sense here — `stipend` is basically initial gas charged for the tx (21k, access list cost etc). I believe all of those are already accounted for in `gas_used` so we'd be simply counting them twice. 

Adding refunded gas might make sense though I believe

klkvr: just tested and this indeed produces higher estimates even for simple scrips, e.g that one:
```solidity
import "forge-std/Script.sol";

contract SimpleScript is Script {
    function run() public {
        vm.broadcast();
        address(0).call("");
    }
}
``` 

will get estimated to use 58012 instead of 29005 as on latest nightly (which is also wrong btw, because the transaction obviously spends just 21000)

grandizzy: the extra gas 29005 gas for example above comes from the fact we tolerate a 10% accuracy (so gets estimated to 22312) when we search for optimal gas

https://github.com/foundry-rs/foundry/blob/91cfab48f48f60a216c814776f02393101907a74/crates/chisel/src/runner.rs#L161-L163

and then on top of it we inflate with a default multiplier of 130

https://github.com/foundry-rs/foundry/blob/91cfab48f48f60a216c814776f02393101907a74/crates/script/src/lib.rs#L130-L132
here

https://github.com/foundry-rs/foundry/blob/91cfab48f48f60a216c814776f02393101907a74/crates/script/src/transaction.rs#L166-L167

so that makes 22312 * 130 / 100 = 29005

Happy to tweak the defaults @klkvr , still looking into the original issue

grandizzy:
closing, superseded by https://github.com/foundry-rs/foundry/pull/11599 and confirmed fixes the bug. will follow up with consolidating / unifying the binary search gas estimation

PR #11479, opened 2025-08-29
fix: correct path reference in FsPathError::Copy AsRef implementation

oooLowNeoNooo:
The Copy variant has both 'from' and 'to' fields, but the AsRef implementation was incorrectly referencing 'from' instead of 'to'. 

This fix ensures that when converting a Copy error to a Path reference, we return the destination path ('to') rather than the source path ('from'), which is more logical for error reporting and path resolution.

- Change: `Self::Copy { from: path, .. }` -> `Self::Copy { to: path, .. }`
- File: crates/common/src/errors/fs.rs
- Line: 95


DaniPopes:
Doesn't really matter, it's ambiguous and either is valid

PR #11481, opened 2025-08-29
feat: switch solar_project to use cache

DaniPopes:
The call to solc when compiling involves parsing and analysis, which is the minimum work the compiler will do if nothing is requested, but this is still slow on the order of 100s of ms to a few seconds. We want to avoid this as much as possible because it's completely unused, as we're requesting nothing and we just want to run the pipeline to get the in-memory `solar` compiler from the output.

Together with updating foundry-compilers for a cache bug fix (https://github.com/foundry-rs/compilers/pull/312), this makes `solar_project` almost never call to solc.

Currently only used by `forge eip712` (https://github.com/foundry-rs/foundry/pull/11458), but will see more usage with https://github.com/foundry-rs/foundry/issues/11456.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -600,7 +600,7 @@ dependencies = [
  "alloy-rlp",
  "alloy-serde",
  "alloy-sol-types",
- "itertools 0.14.0",
+ "itertools 0.13.0",
  "serde",
  "serde_json",
  "serde_with",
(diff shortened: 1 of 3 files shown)


0xrusowsky (approved):
makes sense!

0xrusowsky:
should also be useful for the formatter

PR #11490, opened 2025-08-30
feat: add missing eip7702 support

ongyimeng:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
oh nice, these were indeed missing

this is one case where a previously existing non_exhaustive on the tx enum causes major bugs

PR #11494, opened 2025-08-31
chore(foundry-cli): use derive Default

nkysg:
`Option` default is `None`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


DaniPopes:
would prefer default calling fn new here

PR #11503, opened 2025-09-01
perf: remove `BuildArgs` from `forge config`

onbjerg:
## Motivation

Running `forge config --json` in https://github.com/ithacaxyz/account/ takes about 100-110ms per invocation. [This profile](https://share.firefox.dev/3HRZcMM) shows that most of this is due to us autodetecting remappings multiple times: 

- We end up here, which calls `_with_root` which loads remappings: https://github.com/foundry-rs/foundry/blob/91cfab48f48f60a216c814776f02393101907a74/crates/config/src/lib.rs#L1637-L1639
- We then immediately turn `Config` into a figment: https://github.com/foundry-rs/foundry/blob/91cfab48f48f60a216c814776f02393101907a74/crates/config/src/lib.rs#L2157-L2161 https://github.com/foundry-rs/foundry/blob/91cfab48f48f60a216c814776f02393101907a74/crates/config/src/lib.rs#L693 which also resolves remappings again

We probably load remappings somewhere else again.

## Solution

Removing `BuildArgs` and `EvmArgs` from `forge config` takes the time down to 25ms per invocation. This is technically a breaking change, but I don't really know if anyone actually, in the wild, ever uses `forge config <opts> > foundry.toml`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/forge/src/cmd/config.rs
+++ b/crates/forge/src/cmd/config.rs
@@ -1,15 +1,21 @@
-use super::build::BuildArgs;
-use clap::Parser;
+use std::path::PathBuf;
+
+use clap::{Parser, ValueHint};
 use eyre::Result;
 use foundry_cli::utils::LoadConfig;
-use foundry_common::{evm::EvmArgs, shell};
+use foundry_common::shell;
 use foundry_config::fix::fix_tomls;
-
-foundry_config::impl_figment_convert!(ConfigArgs, build, evm);
 
 /// CLI arguments for `forge config`.
 #[derive(Clone, Debug, Parser)]
 pub struct ConfigArgs {
+    /// The project's root path.
+    ///
+    /// By default root of the Git repository, if in one,
+    /// or the current working directory.
+    #[arg(long, value_hint = ValueHint::DirPath, value_name = "PATH")]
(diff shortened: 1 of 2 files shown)


crates/forge/src/cmd/config.rs
```diff
@@ -1,30 +1,30 @@
...
     /// Attempt to fix any configuration warnings.
     #[arg(long)]
     fix: bool,
-
-    // support nested build arguments
-    #[command(flatten)]
-    build: BuildArgs,
-
-    #[command(flatten)]
-    evm: EvmArgs,
```
mattsse, line 26: this will most certainly break existing use cases,

can we instead add an argument like `kind: Option<FigmentProviders>` and load it accordingly

onbjerg: I am aware that this is breaking, but as I mentioned in the PR description, I've never seen anyone do `forge config <args> > foundry.toml` which is why I want to have a conversation about it

DaniPopes:
this change would make it so load_config does not find all remappings so it's incorrect

you would need to do Config::load which would do the same thing as the current implementation

PR #11506, opened 2025-09-02
perf(cli): optimize did_you_mean function to avoid unnecessary string allocations

DeVikingMark:
Replace map().filter() chain with filter_map() in did_you_mean function to improve performance.

- Avoid creating temporary strings for candidates that don't meet similarity threshold
- Reduce memory allocations when processing large candidate lists
- Maintain exact same functionality while improving efficiency

This change is particularly beneficial when the function is called with many candidates, as it prevents unnecessary String allocations for low-similarity matches.

### DIFF
--- a/crates/cli/src/utils/suggestions.rs
+++ b/crates/cli/src/utils/suggestions.rs
@@ -6,7 +6,6 @@
 /// The jaro winkler similarity boosts candidates that have a common prefix, which is often the case
 /// in the event of typos. Thus, in a list of possible values like ["foo", "bar"], the value "fop"
 /// will yield `Some("foo")`, whereas "blark" would yield `None`.
-///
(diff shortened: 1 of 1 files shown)


crates/cli/src/utils/suggestions.rs
```diff
@@ -6,15 +6,24 @@
 /// The jaro winkler similarity boosts candidates that have a common prefix, which is often the case
 /// in the event of typos. Thus, in a list of possible values like ["foo", "bar"], the value "fop"
 /// will yield `Some("foo")`, whereas "blark" would yield `None`.
+///
+/// This function is optimized to avoid creating temporary strings for candidates that don't meet
```
0xrusowsky: this comment is not necessary

DaniPopes:
Thanks but this is not necessary, this function is not called where performance matters

PR #11509, opened 2025-09-02
fix: own creation initcode to avoid borrowing temporary in trace branch

radik878:
Ensure maybe_creation_code is Bytes across branches (standard, CREATE2, trace). Replace borrowed temporary with owned value; adjust comparisons and slicing to use as_ref(). Removes a potential lifetime error and aligns with the approach used in cast creation-code.


DaniPopes:
This doesn't do anything

"Removes a potential lifetime error" if it's not there then it's not an error

PR #11511, opened 2025-09-02
refactor(common): remove `ReportKind` struct

### LINKED ISSUE #11505: chore(common): remove unnecessary `ReportKind` abstraction
I propose to remove struct `ReportKind` as it is a redundant abstraction layer over `shell`.
https://github.com/foundry-rs/foundry/blob/91cfab48f48f60a216c814776f02393101907a74/crates/common/src/reports.rs#L6

It is only used in `SizeReport`, `TestSummaryReport`, and `GasReport`. Other forge report types are already using `shell` directly (e.g. storage, selectors...).

This straightforward simplification represents a nice initial step towards refactoring and unifying the display of all Forge report types (#11473 as follow-up).

If you think it's relevant, I would be happy to propose a PR 😊

mablr:
## Motivation
Close #11505

## Solution
Updated `SizeReport`, `GasReport`, and `TestSummaryReport` to eliminate `report_kind` usage to directly check shell.

Unified pattern for table style selection based on shell.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


zerosnacks (commented):
Thanks! Changes look good to me

RE: CI failures, this is related to an issue on Reth; will re-run the CI once a fix is applied

DaniPopes:
There's a deadlock with re-entering the global shell here, looking into it

Happens here:
https://github.com/foundry-rs/foundry/blob/91cfab48f48f60a216c814776f02393101907a74/crates/forge/src/cmd/test/mod.rs#L779-L783

onbjerg (approved):

PR #11516, opened 2025-09-02
feat(sol-macro-gen): add `Clone` to `extra_derives`

fd0r:
## Motivation

When create rust bindings for solidity contracts `serde::{Serialize, Deserialize}` traits are added as `extra_derives` by `foundry` and `Debug, PartialEq, Eq, Hash` by `alloy/core`.

I need for some enums coming from generated bindings to implement the `Clone` trait and the underlying structs seem to already implement it, only the enum is missing it.

## Solution

Add `Clone` to the `extra_derives` list.

## PR Checklist

- [ ] Added Tests
    - I didn't find any related tests already implemented 
- [ ] Added Documentation
    - `extra_derives` list currently does not exist in the documentation afaict
- [ ] Breaking changes
    - Non-breaking change

### DIFF
--- a/crates/sol-macro-gen/src/sol_macro_gen.rs
+++ b/crates/sol-macro-gen/src/sol_macro_gen.rs
@@ -125,7 +125,7 @@ impl MultiSolMacroGen {
                     syn::parse_quote! {
                             #[sol(rpc, alloy_sol_types = alloy::sol_types, alloy_contract =
                     alloy::contract, all_derives = true, extra_derives(serde::Serialize,
-                    serde::Deserialize))]     }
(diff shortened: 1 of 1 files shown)


fd0r:
This is a bit of a dirty fix because I patched my locally built foundry to add the `Clone` trait as `extra_derives` then built the bindings with it.

DaniPopes:
Clone is derived by default, if you have a case where this doesn't happen please open an issue in https://github.com/alloy-rs/core

fd0r:
superseded by https://github.com/alloy-rs/core/pull/1003

PR #11532, opened 2025-09-03
feat(stack): make edge coverage output size configurable

Bashmunta:
- Add edge_coverage_size to InspectorStackInner
- Add set_edge_coverage_size to InspectorStack
- Resize edge coverage hitcount vector during collect
- Replace TODO with explanatory comment


DaniPopes:
Needs support upstream. This fix is incorrect.

Issue #6706, opened 2024-01-04
feat: allow single file remappings


arthurneuron:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### What version of Foundry are you on?

forge 0.2.0 (9e3ab9b 2024-01-04T00:18:01.892563000Z)

### What command(s) is the bug in?

forge test

### Operating System

macOS (Apple Silicon)

### Describe the bug

- Correct path: `lib/openzeppelin-contracts/contracts/token/ERC721/extensions/IERC721Metadata.sol`
- Required Import: `@openzeppelin/contracts/token/ERC721/IERC721Metadata.sol`
- This is not possible, since this exception is only for one file: `@openzeppelin/contracts/token/ERC721/=lib/openzeppelin-contracts/contracts/token/ERC721/extensions/`

**remappings.txt**
- `@openzeppelin/contracts/token/ERC721/IERC721Metadata.sol=lib/openzeppelin-contracts/contracts/token/ERC721/extensions/IERC721Metadata.sol`

**Error:**
- `failed to resolve file: "/Users/foundry/Developer/foundry/packages/contracts/lib/openzeppelin-contracts/contracts/token/ERC721/extensions/IERC721Metadata.sol/": Not a directory (os error 20); check configured remappings.`

**remappings.txt**
- `@openzeppelin/contracts/token/ERC721/=lib/openzeppelin-contracts/contracts/token/ERC721/extensions/IERC721Metadata.sol`

**Error:**
- `failed to resolve file: "/Users/foundry/Developer/foundry/packages/contracts/lib/openzeppelin-contracts/contracts/token/ERC721/extensions/IERC721Metadata.sol/IERC721Metadata.sol": Not a directory (os error 20); check configured remappings.`

**remappings.txt**
- `@openzeppelin/contracts/token/ERC721/IERC721Metadata.sol=lib/openzeppelin-contracts/contracts/token/ERC721/extensions/`

**Error:**
- `failed to resolve file: "/Users/foundry/Developer/foundry/packages/contracts/lib/openzeppelin-contracts/contracts/token/ERC721/extensions/": Is a directory (os error 21); check configured remappings.`

mattsse:
could you please provide a repro for this?

mattsse:
@tash-2s I wonder if this is related to the recent path changes?

tash-2s:
I believe this is the reproduction of the issue:

```console
[...]

PR #11536, opened 2025-09-03
perf: parallelize remappings provider

DaniPopes:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/config/src/providers/remappings.rs
+++ b/crates/config/src/providers/remappings.rs
@@ -4,6 +4,7 @@ use figment::{
     value::{Dict, Map},
 };
 use foundry_compilers::artifacts::remappings::{RelativeRemapping, Remapping};
+use rayon::prelude::*;
 use std::{
     borrow::Cow,
     collections::{BTreeMap, HashSet, btree_map::Entry},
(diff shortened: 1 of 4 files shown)


crates/config/src/providers/remappings.rs
```diff
@@ -246,50 +244,67 @@ impl RemappingsProvider<'_> {
...
+        if ![Path::new("src"), Path::new("contracts"), Path::new("lib")]
+            .contains(&config.src.as_path())
+            && let Some(name) = lib.file_name().and_then(|s| s.to_str())
+        {
+            let mut r = Remapping {
+                context: None,
+                name: format!("{name}/"),
+                path: format!("{}", lib.join(&config.src).display()),
+            };
+            if !r.path.ends_with('/') {
```
grandizzy, line 278: not this PR related but wonder if we should apply here same check as in https://github.com/foundry-rs/compilers/commit/f6e4e6a441f6ff7d45c8f0ddcf8b786252433c41 that is if nested libs has single files remapping

grandizzy: nm, we're here looking remappings for deps / dirs, so doesn't apply, tested with project from https://github.com/foundry-rs/foundry/issues/6706#issuecomment-3141270852 as a dep and remappings looks good / can use single file remapping from dep in current project

PR #11537, opened 2025-09-03
feat(cheatcodes): vm.getStorageSlots

### LINKED ISSUE #3869: Cannot find the storage slot for a public string variable
### Component

Forge

- [X] Foundry
- [ ] Foundryup

forge 0.2.0 (b28119b 2022-11-23T03:06:50.152811Z)

forge test

### Operating System

Windows

### Describe the bug

For a simple storage layout given below, it's possible to find the storage slot for `owner` and `number` by running `stdstore.target(address(test)).sig("owner()").find();` or `stdstore.target(address(test)).sig("number()").find();`, but it's failing when running `stdstore.target(address(test)).sig("sentence()").find();`.
```
contract SimpleStorage {
    address public owner;
    string public sentence;
    uint256 public number;
}
```

yash-atreya:
## Motivation

Closes #3869 

## Solution

- New cheatcode: `vm.getStorageSlots(address target, string variableName) returns (uint256[] slots)`
- Leverages the storage layout to pinpoint the base slot of the variable and fetches rest of the slots according to the endoing type
- Test

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/cheatcodes/src/evm.rs
```diff
@@ -908,6 +909,77 @@ impl Cheatcode for getStateDiffJsonCall {
...
+            if let Ok(value) = ccx.ecx.journaled_state.sload(*target, slot) {
+                let value_bytes = value.data.to_be_bytes::<32>();
+                let length_byte = value_bytes[31];
+                // Check if it's a long bytes/string (LSB is 1)
+                if length_byte & 1 == 1 {
+                    // Calculate data slots for long bytes/string
+                    let length: U256 = value.data >> 1;
+                    let num_data_slots = length.to::<usize>().div_ceil(32);
+                    let data_start =
+                        U256::from_be_bytes(alloy_primitives::keccak256(B256::from(slot).0).0);
```
zerosnacks: nit, we can import keccak256 above

zerosnacks:
Are we planning to deprecate ` StdStorage` is `forge-std`? The cheatcode seems preferable.

crates/cheatcodes/src/evm.rs
yash-atreya: [bef852a](https://github.com/foundry-rs/foundry/pull/11537/commits/bef852a54019601014f40a599b89e3f33f18f17a)

yash-atreya:
> Are we planning to deprecate ` StdStorage` is `forge-std`? The cheatcode seems preferable.

This approach leverages the `StorageLayout`, whereas the `StdStorage` approach doesn't and works well for simple variables but NOT for variables that span multiple slots.

I'm okay with deprecating `StdStorage` but users would have to compile with `--extra-output storageLayout` for this to work

zerosnacks:
I see, deprecating `StdStorage` likely causes issues in existing scripting suites and forcing a rewrite for something that is working for a user is not preferable. We do need to make sure we document this new cheatcode in the book to help guide new users towards it.

grandizzy (approved):

ZeroEkkusu:
Would be nice to include storage layout flag in `foundry.toml`.

PR #11539, opened 2025-09-03
chore(config): figment provider clean up

DaniPopes:
Minor improvements I made while trying to understand this code.


crates/config/src/providers/ext.rs
```diff
@@ -299,14 +302,14 @@ impl<P: Provider> Provider for BackwardsCompatTomlProvider<P> {
                     dict.insert("solc".to_string(), v);
                 }
             }
-
-            if let Some(v) = dict.remove("odyssey") {
-                dict.insert("odyssey".to_string(), v);
-            }
```
0xrusowsky, line 305: is this a breaking change?

DaniPopes: it's a noop, it removes (k, v) to insert (k, v)

onbjerg (approved):

PR #11549, opened 2025-09-04
fix(cast): correct error message for --ends-with hex parsing

Forostovec:
The --ends-with branch used “invalid prefix hex provided” due to a copy-paste oversight. This change updates it to “invalid suffix hex provided” for accuracy and consistency with the --starts-with and --matching branches


zerosnacks (approved):
Thanks :+1: 

To reproduce

```
cast create2 --ends-with 123123! --init-code-hash 0xb0829a887022f57872a527b4db0efc5add58f21e4268758672261249ed193bcd
```

```
Error: invalid prefix hex provided

Context:
- invalid character '!' at position 6
```

should say suffix indeed

Issue #11519, opened 2025-09-02
bug(forge-lint): erc20-unchecked-transfer false positive


Philogy:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

forge Version: 1.3.2-stable Commit SHA: b0381e15d1465396aabcb398b60d2c10cc0112f2 Build Timestamp: 2025-08-29T06:56:57.874013000Z (1756450617) Build Profile: maxperf

### What version of Foundryup are you on?

foundryup: 0.3.3

### What command(s) is the bug in?

forge lint

### Operating System

macOS (Apple Silicon)

### Describe the bug

When using a library like https://github.com/Uniswap/v4-core/blob/main/src/types/Currency.sol use of the `transfer` library method on the `Currency` type is falsely recognized as an unchecked ERC20 transfer, even though it's a library method on a custom type and the inner logic is actually checking the transfer

grandizzy:
that's ref https://github.com/foundry-rs/foundry/pull/10810#discussion_r2157167940 at that time we decided to clarify in docs but maybe we could improve now @0xrusowsky wdyt?

0xrusowsky:
yes, now that we also have the HIR and we are not limited to just the AST, i think it's time to resolve the contract and see if it is an ERC20

PR #11552, opened 2025-09-04
feat(lint): impl erc20 transfer check using HIR

0xrusowsky:
ref:
- https://github.com/foundry-rs/foundry/issues/11519

TODO:
- [x] test repro

### DIFF
--- a/crates/forge/tests/it/cheats.rs
+++ b/crates/forge/tests/it/cheats.rs
@@ -25,7 +25,9 @@ async fn test_cheats_local(test_data: &ForgeTestData) {
     }
 
     if cfg!(feature = "isolate-by-default") {
-        filter = filter.exclude_contracts("(LastCallGasDefaultTest|MockFunctionTest|WithSeed)");
+        filter = filter.exclude_contracts(
+            "(LastCallGasDefaultTest|MockFunctionTest|WithSeed|StateDiff|RecordAccount)",
+        );
     }
 
     let runner = test_data.runner_with(|config| {


crates/lint/src/sol/high/unchecked_calls.rs
```diff
@@ -25,55 +26,91 @@ declare_forge_lint!(
...
+        hir::CallArgs { kind: hir::CallArgsKind::Unnamed(args), .. },
+        ..,
+    ) = &expr.kind
+    else {
+        return false;
+    };
+
+    // Determine the expected ERC20 signature from the call
+    let (expected_params, expected_returns): (&[&str], &[&str]) = match func_ident.as_str() {
+        "transferFrom" if args.len() == 3 => (&["address", "address", "uint256"], &["bool"]),
```
grandizzy, line 75: will this fail if sig is declared with `uint` instead `uint256`?

0xrusowsky: no, cause this is the string representation of the type (which is computed after solar's type resolution)

```rs
let is_type = |var_id: hir::VariableId, type_str: &str| {
        matches!(
            &hir.variable(var_id).ty.kind,
            hir::TypeKind::Elementary(ty) if ty.to_abi_str() == type_str
        )
    };
```

grandizzy (approved):
lgtm! tested uni v4 lint doesn't show warning for https://github.com/Uniswap/v4-core/blob/59d3ecf53afa9264a16bba0e38f4c5d2231f80bc/src/ProtocolFees.sol#L56 anymore  but correctly shows for https://github.com/Uniswap/v4-core/blob/59d3ecf53afa9264a16bba0e38f4c5d2231f80bc/test/utils/CurrencySettler.sol#L28-L32

zerosnacks (approved):

PR #11560, opened 2025-09-05
feat: make `find_by_deployed_code_exact` smarter

klkvr:
## Motivation

Changes `find_by_deployed_code_exact` to also ignore the metadata hash at the end of bytecode. Review with "Hide whitespace"

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/common/src/contracts.rs
+++ b/crates/common/src/contracts.rs
@@ -325,7 +325,7 @@ impl ContractsByArtifact {
                     left = right + offset.length as usize;
                 }
 
-                let partal_match = if left < code.len() {
+                let is_partial = if left < code.len() {
                     match deployed_code {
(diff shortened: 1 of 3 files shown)


onbjerg (approved):

crates/common/src/contracts.rs
```diff
@@ -241,94 +241,130 @@ impl ContractsByArtifact {
...
+                    };
+
+                    if !matched {
+                        return false;
                     }
+
+                    left = right + offset.length as usize;
+                }
+
+                let partal_match = if left < code.len() {
```
grandizzy: typo `partal_match` but probably needs renamed due to `let mut partial_match = None` above or maybe just add an early return

grandizzy (approved):
lgtm

yash-atreya (approved):

PR #11562, opened 2025-09-05
test: use proper EVM version casing in CLI tests

aganisgash:
Updates test cases to use `Cancun` instead of `cancun` for EVM version.

The config system already warns about the lowercase `cancun` key being deprecated in favor of `evm_version = Cancun`, but our CLI tests were still using the old lowercase format. This creates an inconsistency where we tell users to use proper casing in config files but our own tests don't follow the same convention.


DaniPopes:
no that is not the same thing, we deprecated the cancun field, not the lowercase evm-version = cancun

PR #11565, opened 2025-09-06
feat(coverage): analyze with solar

DaniPopes:
Closes https://github.com/foundry-rs/foundry/issues/10964.

By using Solar to analyze contracts to find items we can avoid requesting AST output from solc, saving a considerable amount of serde time.

This also fixes a few bugs, while trying to remain 100% compatible with the previous solc AST implementation. See individual commits and their descriptions for more details.

---

Default project: 
image

ithacaxyz/account: 
image


crates/evm/coverage/src/analysis.rs
```diff
@@ -79,507 +81,287 @@ impl<'a> ContractVisitor<'a> {
...
+                        // We use `stmt.span`, which includes `else_stmt.span`, since we need to
+                        // include the condition so that this can be marked as covered.
+                        // Initially implemented in https://github.com/foundry-rs/foundry/pull/3094.
+                        self.push_item_kind(
+                            CoverageItemKind::Branch {
+                                branch_id,
+                                path_id: 1,
+                                is_first_opcode: false,
+                            },
+                            stmt.span,
```
DaniPopes, line 264: this is not entirely equivalent, but the previous logic resulted in weird spans for else-if chains (left is before, right is after): 
image

this was added in https://github.com/foundry-rs/foundry/pull/3094

grandizzy (approved):
lgtm!

PR #11566, opened 2025-09-06
feat(cast): support base + offset for storage slot calc

### LINKED ISSUE #11531: add base + offset support in `cast storage`
### Component

Cast

Needing to do `base + offset` is pretty common when using non conflicting storages/diamond storage, etc. like starting base slot at `MyContract.storage` and then having a struct there like

```solidity
contract MyContract {
    bytes32 constant STORAGE_SLOT = keccak256("MyContract.storage");

    struct MyStruct {
        uint a;
        uint b;
        uint[] c;
    }

    function _getStorage() internal returns (MyStruct storage _stg) {
        bytes32 slot = STORAGE_SLOT;
        assembly { _stg.slot := slot }
    }

    function readA() external view returns (uint) {
        return _getStorage().a;
    }

    function readC() external view returns (uint[] memory) {
        return _getStorage().c;
    }
}
```
In current form if i have to read the storage i have to do something equivalent to:
```sh
$ cast storage <addr> $(python3 -c "print(hex(int('$(cast keccak 'MyContract.storage')', 16) + 2))")
```

i would rather prefer not to use python just to do the bigint math, and instead be able to do `cast storage <address> [base_slot] [offset]` assuming offset of 0 when it is not specified.

Happy to PR this but opening issue to see if it will get accepted.

### Additional context

_No response_

0xferrous:
Closes #11531

## Motivation

Needing to do base + offset is pretty common when using non conflicting storages/diamond storage, etc.

## Solution

Add arg for specifying the offset.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/cast/src/cmd/storage.rs
```diff
@@ -49,7 +49,11 @@ pub struct StorageArgs {
 
     /// The storage slot number. If not provided, it gets the full storage layout.
     #[arg(value_parser = parse_slot)]
-    slot: Option<B256>,
+    base_slot: Option<B256>,
```
grandizzy, line 52: should we maybe add an alias and keep both slot/base_slot so it won't be a breaking change?

0xferrous:
It's a positional argument, it's not a breaking change, the name only changes in the help.

-------- Original Message --------

grandizzy:
> It's a positional argument, it's not a breaking change, the name only changes in the help.
> […](#)

ah, right, missed that, thanks

crates/cast/src/cmd/storage.rs
```diff
@@ -91,14 +95,17 @@ impl StorageArgs {
...
         let address = address.resolve(&provider).await?;
 
         // Slot was provided, perform a simple RPC call
-        if let Some(slot) = slot {
+        if let Some(slot) = base_slot {
             let cast = Cast::new(provider);
-            sh_println!("{}", cast.storage(address, slot, block).await?)?;
+            sh_println!(
+                "{}",
+                cast.storage(address, (Into::<U256>::into(slot) + offset).into(), block).await?
```
grandizzy: don't think we should saturating add here? @DaniPopes wdyt?

grandizzy (approved):
lgtm, thank you! Pending one more review before merge

crates/cast/src/cmd/storage.rs
0xferrous: uh interesting, didn't know the default was ~~saturating~~ wrapping:
```rust
    println!("add: {}", U256::MAX + U256::from(1));
```
out: `add: 0`
weird considering default math in primitive rust types is checked.

zerosnacks: I think saturating here is preferred to avoid panics

grandizzy: @0xferrous mind to accommodate above then will merge. Thank you!

DaniPopes: the default for ruint is wrapping: https://github.com/recmo/uint/issues/502
the change here is to replace `+` with `.saturating_add`.

0xferrous: ah i had it confused, thanks for the clarification, saturating is not wrapped.

PR #11574, opened 2025-09-07
Fix panic on NaN/Inf values in TOML to JSON conversion

prestoalvarez:
Fixes a potential panic when converting TOML float values containing `NaN`, `+inf`, or `-inf` to JSON format.

**Problem:**
- `serde_json::Number::from_f64()` returns `None` for non-finite values
- The code used `.unwrap()` which would panic on valid TOML input
- TOML spec allows these values, but JSON doesn't support them

**Solution:**
- Replace `unwrap()` with safe pattern matching
- Convert non-finite floats to string representation (`"NaN"`, `"inf"`, `"-inf"`)
- Maintains type safety and prevents unexpected crashes

### DIFF
--- a/testdata/default/cheats/Toml.t.sol
+++ b/testdata/default/cheats/Toml.t.sol
@@ -329,6 +329,24 @@ contract ParseTomlTest is DSTest {
 
         assertEq(keccak256(abi.encode(members)), keccak256(abi.encode(data.members)));
     }
+    
+    function test_floatNaN() public {
+        bytes memory data = vm.parseToml(toml, ".nanFloat");
+        string memory decodedData = abi.decode(data, (string));
+        assertEq("NaN", decodedData);
+    }
+
+    function test_floatInf() public {
+        bytes memory data = vm.parseToml(toml, ".infFloat");
+        string memory decodedData = abi.decode(data, (string));
+        assertEq("inf", decodedData);
+    }
+
+    function test_floatNegInf() public {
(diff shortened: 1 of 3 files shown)


DaniPopes:
Thanks, please add a test case in testdata/default/cheats/Toml.t.sol

zerosnacks:
Thanks @prestoalvarez! Agreed on the need for a test case

prestoalvarez:
@DaniPopes @zerosnacks

testdata/fixtures/Toml/test.toml
```diff
@@ -48,3 +48,9 @@ id = 1
 
 [[advancedTomlPath]]
 id = 2
+
+# Floating special values for tests
+nanFloat = nan
```
DaniPopes: These need to be at the top of the file

testdata/default/cheats/Toml.t.sol
```diff
@@ -329,6 +329,24 @@ contract ParseTomlTest is DSTest {
...
+        assertEq("NaN", decodedData);
+    }
+
+    function test_floatInf() public {
+        bytes memory data = vm.parseToml(toml, ".infFloat");
+        string memory decodedData = abi.decode(data, (string));
+        assertEq("inf", decodedData);
+    }
+
+    function test_floatNegInf() public {
```
DaniPopes, line 345: please fix formatting with forge fmt

DaniPopes (approved):
Thanks!

PR #11590, opened 2025-09-09
fix(cheatcodes): lock files on reads / writes

grandizzy:
## Motivation
- some tests are writing files using `writeFile` cheatcode and then reading with `readFile` cheatcode and assert content
- encountered race condition on one project where `readFile` issued after several writes in the same file failed with 
```bash
├─ [0] VM::keyExists("<JSON file>", ".addresses.accessManager") [staticcall]
    │   └─ ← [Revert] vm.keyExists: failed parsing JSON: EOF while parsing a value at line 1 column 0
    └─ ← [Revert] vm.keyExists: failed parsing JSON: EOF while parsing a value at line 1 column 0
```
even if traces were showing write just above the read
image

- issue cannot be reproduced at will but only if running `forge test` of the specific test in a loop
- `<JSON file>` in traces above masked at
https://github.com/foundry-rs/foundry/blob/f60a1d88e5b04400232aba3056917665c49857ef/crates/evm/traces/src/decoder/mod.rs#L565

## Solution
- introduce fs utility functions to lock/unlock file on write and lock_shared/unlock on read. Adds new fs errors variant  for lock and unlock file

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cheatcodes/src/fs.rs
+++ b/crates/cheatcodes/src/fs.rs
@@ -16,7 +16,7 @@ use revm::{context::CreateScheme, interpreter::CreateInputs};
 use revm_inspectors::tracing::types::CallKind;
 use semver::Version;
 use std::{
-    io::{BufRead, BufReader, Read, Write},
+    io::{BufRead, BufReader},
     path::{Path, PathBuf},
     process::Command,
     sync::mpsc,
@@ -149,12 +149,7 @@ impl Cheatcode for readFileCall {
     fn apply(&self, state: &mut Cheatcodes) -> Result {
         let Self { path } = self;
         let path = state.config.ensure_path_allowed(path, FsAccessKind::Read)?;
-        let file = std::fs::OpenOptions::new().read(true).open(path)?;
-        file.lock_shared()?;
-        let mut contents = String::new();
-        (&file).read_to_string(&mut contents)?;
-        file.unlock()?;
-        Ok(contents.abi_encode())
+        Ok(fs::locked_read_to_string(path)?.abi_encode())
     }
 }
 
@@ -162,12 +157,7 @@ impl Cheatcode for readFileBinaryCall {
     fn apply(&self, state: &mut Cheatcodes) -> Result {
         let Self { path } = self;
         let path = state.config.ensure_path_allowed(path, FsAccessKind::Read)?;
-        let file = std::fs::OpenOptions::new().read(true).open(path)?;
-        file.lock_shared()?;
-        let mut buffer = Vec::with_capacity(file.metadata()?.len() as usize);
-        (&file).read_to_end(&mut buffer)?;
-        file.unlock()?;
-        Ok(buffer.abi_encode())
+        Ok(fs::locked_read(path)?.abi_encode())
     }
 }
 
@@ -253,10 +243,7 @@ impl Cheatcode for writeLineCall {
         state.config.ensure_not_foundry_toml(&path)?;
 
         if state.fs_commit {
-            let mut file = std::fs::OpenOptions::new().append(true).create(true).open(path)?;
-            file.lock()?;
-            writeln!(file, "{line}")?;
-            file.unlock()?;
+            fs::locked_write_line(path, line)?;
         }
 
         Ok(Default::default())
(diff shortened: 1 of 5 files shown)


0xrusowsky (dismissed):
i think this makes sense

crates/common/src/fs.rs
```diff
@@ -97,6 +97,20 @@ pub fn write(path: impl AsRef<Path>, contents: impl AsRef<[u8]>) -> Result<()> {
...
+    let path = path.as_ref();
+    let mut file = OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
+        .open(path)
+        .map_err(|err| FsPathError::write(err, path))?;
+    file.write_all(contents.as_ref())
+        .and_then(|_| file.sync_all())
+        .map_err(|err| FsPathError::write(err, path))
```
zerosnacks: nit: the dual use of `FsPathError::write` can be confusing here as the first one refers to an error related to the opening of the file and not the writing

zerosnacks:
Makes sense to me, small nit

crates/common/src/fs.rs
grandizzy: can `FsPathError::open` the first one, is that OK?

grandizzy: https://github.com/foundry-rs/foundry/pull/11590/commits/be67708781d14f4cec24f31a3d789943d27c1ff0

zerosnacks (approved):

onbjerg (commented):
is it important that we flush *all* data (incl metadata) to disk for this to work? or can we just do [`File::sync_data`](https://doc.rust-lang.org/stable/std/fs/struct.File.html#method.sync_data)?

grandizzy:
> is it important that we flush _all_ data (incl metadata) to disk for this to work? or can we just do [`File::sync_data`](https://doc.rust-lang.org/stable/std/fs/struct.File.html#method.sync_data)?

good point, yeah, I guess we can omit metadata

DaniPopes:
this makes sense however i wonder if we should lock the file instead of syncing during cheatcode execution? this is on stable now so you can just call [file.lock()](https://doc.rust-lang.org/stable/std/fs/struct.File.html#method.lock) for write, .lock_shared() for read, and .unlock() and should avoid fs

grandizzy:
> this makes sense however i wonder if we should lock the file instead of syncing during cheatcode execution? this is on stable now so you can just call [file.lock()](https://doc.rust-lang.org/stable/std/fs/struct.File.html#method.lock) for write, .lock_shared() for read, and .unlock() and should avoid fs

```bash
warning: current MSRV (Minimum Supported Rust Version) is `1.88.0` but this item is stable since `1.89.0`
   --> crates/cheatcodes/src/fs.rs:154:14
    |
154 |         file.lock_shared()?;
```

should I bump?

DaniPopes:
yes

grandizzy:
> yes

done, please recheck. ~CI failure due to new node not supporting wss, will ask to open~. thanks

crates/cheatcodes/src/fs.rs
```diff
@@ -149,15 +149,25 @@ impl Cheatcode for readFileCall {
...
     }
 }
 
 impl Cheatcode for readFileBinaryCall {
     fn apply(&self, state: &mut Cheatcodes) -> Result {
         let Self { path } = self;
         let path = state.config.ensure_path_allowed(path, FsAccessKind::Read)?;
-        Ok(fs::read(path)?.abi_encode())
+        let file = std::fs::OpenOptions::new().read(true).open(path)?;
+        file.lock_shared()?;
```
DaniPopes: should we move to `common::fs` module?

grandizzy: yep, will do, also noticed that `writeJson` is doing `let data_string = fs::read_to_string(&data_path)?;` for read which should be locked

grandizzy: changed in https://github.com/foundry-rs/foundry/pull/11590/commits/52dc3004a0de8a9217201283b7c646b4ed100ec2

DaniPopes:
please update title/description

Issue #11433, opened 2025-08-27
feat(`anvil`): add `eth_config` RPC endpoint for Osaka hardfork


zerosnacks:
### Component

Anvil

### Describe the feature you would like

Specification: https://eips.ethereum.org/EIPS/eip-7910

Reference in Reth: https://github.com/paradigmxyz/reth/pull/18071/files#diff-fafb33484dd6dedcb26f4129ebc4bb83fc369f21d0d742b50d32b4fcade02534

I think it makes sense for Anvil to adhere to this standard RPC method

Issue #11428, opened 2025-08-26
meta: make Foundry ready for Osaka hardfork


zerosnacks:
Related to: https://github.com/foundry-rs/foundry/issues/11348

Most of this is handled by Revm but we need to make sure the integration points are also ready

- [ ] https://eips.ethereum.org/EIPS/eip-7892, make sure codebase adheres to the EIPs expectation
- [ ] https://eips.ethereum.org/EIPS/eip-7825
- [ ] https://eips.ethereum.org/EIPS/eip-7918
- [ ] https://eips.ethereum.org/EIPS/eip-7951
- [ ] https://eips.ethereum.org/EIPS/eip-7910 (anvil?)

https://eips.ethereum.org/EIPS/eip-7825 can have a significant impact and we need to make sure all scenario's are covered

Ideally we have a way to run the same test suite easily against a range of EVM versions (such as the repro's and cheat suites)

PR #11591, opened 2025-09-09
feat(`anvil`): add `eth_config` rpc endpoint

zerosnacks:
## Motivation

In order to make it easier to figure out the supported precompiles and system contracts as a user I think it makes sense to add an `eth_config`-like RPC method: https://eips.ethereum.org/EIPS/eip-7910

Closes: https://github.com/foundry-rs/foundry/issues/11433

This completes the work required to make Foundry ready for Osaka.

Closes: https://github.com/foundry-rs/foundry/issues/11428

## Solution

Adds `eth_config`-like endpoint inspired by the actual `eth_config` specification with notable exceptions:

- Note: the activation timestamp is always 0 as the configuration is set at genesis (spec compliant).
- Note: the `next` and `last` fields are always `null` as this node does not participate in any forking on the network (spec compliant).

Not spec compliant:

The `fork_id` is always `0x00000000` as this node does not participate in any forking on the network. For all intended purposes you can consider this unused by Anvil.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -27,7 +27,7 @@ use crate::{
         pool::transactions::PoolTransaction,
         sign::build_typed_transaction,
     },
-    evm::celo_precompile,
+    evm::celo_precompile::{self, CELO_TRANSFER_ADDRESS},
     inject_precompiles,
     mem::{
         inspector::AnvilInspector,
@@ -41,7 +41,10 @@ use alloy_consensus::{
     proofs::{calculate_receipt_root, calculate_transaction_root},
     transaction::Recovered,
 };
-use alloy_eips::{eip1559::BaseFeeParams, eip4844::kzg_to_versioned_hash, eip7840::BlobParams};
+use alloy_eips::{
+    eip1559::BaseFeeParams, eip4844::kzg_to_versioned_hash, eip7840::BlobParams,
+    eip7910::SystemContract,
+};
 use alloy_evm::{
     Database, Evm,
     eth::EthEvmContext,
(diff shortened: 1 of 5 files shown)


crates/anvil/tests/it/fork.rs
```diff
@@ -1607,3 +1618,277 @@ async fn test_fork_get_account() {
...
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_config_with_cancun_hardfork() {
+    let (api, _handle) =
+        spawn(NodeConfig::test().with_hardfork(Some(EthereumHardfork::Cancun.into()))).await;
+
+    let config = api.config().unwrap();
+
+    let expected_blob_params = BlobParams {
```
grandizzy, line 1654: the EIP shows only
```
      "blobSchedule": {
        "baseFeeUpdateFraction": 3338477,
        "max": 6,
        "target": 3
      }
```
probably not an issue though if we expose all of them

mattsse: the others are skipped in serde

https://github.com/alloy-rs/alloy/blob/a2da504080c883f553fc2b64793bc3ad27fdf0d1/crates/eips/src/eip7840.rs#L155-L166

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -843,6 +850,56 @@ impl Backend {
...
+                celo_precompile::PRECOMPILE_ID_CELO_TRANSFER.clone().name().to_string(),
+                CELO_TRANSFER_ADDRESS,
+            );
+        }
+
+        if let Some(factory) = &self.precompile_factory {
+            for (precompile, _) in &factory.precompiles() {
+                precompiles_map.insert(precompile.id().name().to_string(), *precompile.address());
+            }
+        }
```
mattsse, line 881: we need to do something about this, we have this in a few locations and maintaining this is a bit horrible

crates/anvil/tests/it/fork.rs
grandizzy: skipped in serde as pointed out by @mattsse

crates/anvil/src/eth/backend/mem/mod.rs
zerosnacks: Agreed, will have a look in a follow-up if we can have a shared helper for this to make this more maintainable as it currently is strictly tied to the EVM creation that we don't have access to in Anvil's backend

PR #11595, opened 2025-09-09
perf: replace format!("{}", len) with more efficient to_string()

eeemmmmmm:
## Motivation

Replace inefficient `format!("{}", value)` pattern with direct `to_string()` method call for simple numeric-to-string conversions. The `format!` macro has overhead from the formatting machinery that is unnecessary when converting a single value to string.

## Solution

Changed `Ok(format!("{}", code.len()))` to `Ok(code.len().to_string())` in the `codesize` method. This micro-optimization improves performance by avoiding the format string parsing and reduces memory allocations while maintaining identical behavior.

## PR Checklist

- [ ] Added Tests - N/A (identical behavior, no functional changes)
- [ ] Added Documentation - N/A (no API changes)  
- [x] Breaking changes - No

### DIFF
--- a/crates/cast/src/lib.rs
+++ b/crates/cast/src/lib.rs
@@ -749,7 +749,7 @@ impl<P: Provider<AnyNetwork>> Cast<P> {
     pub async fn codesize(&self, who: Address, block: Option<BlockId>) -> Result<String> {
         let code =
             self.provider.get_code_at(who).block_id(block.unwrap_or_default()).await?.to_vec();
-        Ok(format!("{}", code.len()))
+        Ok(code.len().to_string())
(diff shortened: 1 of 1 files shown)


zugdev:
I think this is not true since long ago, if I recall correctly, prev to release 1.9.0
the best idea performance wise is to be consistent about choosing format or to_string

but im not a maintainer so not up to me

zugdev:
is this still true?

https://github.com/rust-lang/rust/issues/99012

DaniPopes:
sometimes, when ToString is specialized for the type. i prefer to_string regardless because i think its intent is clearer.

Issue #9752, opened 2025-01-24
bug: derive block gas limit from RPC to prevent erroneous filling of block during simulation


zerosnacks:
### Component

Forge, Anvil

### Describe the feature you would like

Related: https://github.com/foundry-rs/foundry/issues/9723 / https://github.com/foundry-rs/foundry/issues/9723#issuecomment-2612448442

It appears we currently do not estimate / update the gas limit specified by the RPC during simulation which can result in a [situation](https://github.com/foundry-rs/foundry/issues/9723) where the filler puts too many transactions into a block during simulation before rolling over to the next

The expected behavior should be that `forge script` correctly listens to the gas limit indicated by the RPC (Anvil) and Anvil should correctly set its gas limit based on the fork.

PR #11599, opened 2025-09-10
fix(forge): account gas limit per fork when script simulate

### LINKED ISSUE #11579: bug: broadcast artifact includes `isFixedGasLimit` where not expected and causes transactions to have a too high gas limit
As shown in practice during a recent deployment on some chains and under certain conditions a `isFixedGasLimit` is included in the broadcast artifact and causes gas limit to be excessive to the point where validators find it difficult to include it in the block.

Our current heuristic for determining whether we should use a fixed gas limit is flawed. An initial idea proposed by @klkvr is that we instead should check if the `CALL` opcode in preceeded by `GAS` and if it is not then it is most likely a manually specified gas limit.

grandizzy:
## Motivation
- closes #11579 #9752 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


grandizzy:
@klkvr @zerosnacks pls check and lmk if it makes sense, thank you!

crates/cheatcodes/src/inspector.rs
```diff
@@ -2292,6 +2306,27 @@ impl Cheatcodes {
             (REVERT, 0, 1, false),
         );
     }
+
+    #[cold]
+    fn record_gas_limit_opcode(&mut self, interpreter: &mut Interpreter) {
+        if interpreter.bytecode.opcode() == op::GAS {
+            self.is_fixed_gas_limit = Some(true);
```
grandizzy: can also change and set this to `Some(false)` and in `set_gas_limit_type` if none and next opcode is `CALL` then set to `Some(true)`
```rust
        if interpreter.bytecode.opcode() == op::CALL {
            if self.is_fixed_gas_limit.is_none() {
                // If GAS opcode wasn't seen then mark as fixed gas limit.
                self.is_fixed_gas_limit = Some(true);
            }
        }
```

crates/cheatcodes/src/inspector.rs
```diff
@@ -1075,6 +1084,10 @@ impl Inspector<EthEvmContext<&mut dyn DatabaseExt>> for Cheatcodes {
     fn step(&mut self, interpreter: &mut Interpreter, ecx: Ecx) {
         self.pc = interpreter.bytecode.pc();
 
+        if self.broadcast.is_some() {
+            self.record_gas_limit_opcode(interpreter);
+        }
```
klkvr, line 1095: I believe right now if opcode sequence would be `GAS ...any sequence of opcodes other than CALL and GAS.. CALL` we will still record this as non-fixed gas limit

what we could do is instead call `record_gas_limit_opcode` in `step_end` hook, and then in `step` either unset it or keep if the opcode is `CALL`

grandizzy: implemented in https://github.com/foundry-rs/foundry/pull/11599/commits/ba1215ca52bca416d882c54dd708b88bd6f1f79a but kept `GAS` followed by `CALL` check as it seems to me better logic. I updated to track `Option<gas_seen, call_seen>` and reset on `step_end` if `GAS` is not followed by `CALL`, pls check

grandizzy: confirmed it also solves https://github.com/foundry-rs/foundry/issues/10968

crates/cheatcodes/src/inspector.rs
```diff
@@ -2292,6 +2308,40 @@ impl Cheatcodes {
...
+            return;
+        }
+
+        // Record CALL opcode if GAS opcode was seen.
+        if matches!(self.is_fixed_gas_limit, Some((true, false)))
+            && interpreter.bytecode.opcode() == op::CALL
+        {
+            self.is_fixed_gas_limit = Some((true, true));
+            return;
+        }
```
klkvr: I stillf find it a bit hard to understand. isn't `step_end` only called after call has completed meaning that this check will likely happen too late?

grandizzy: I could be wrong but afaik in`step_end` we already have access to the next opcode but before call, @DaniPopes @rakita could you pls chime in?

grandizzy: @klkvr I think I got your comment wrong, so want to go through logic again
- in `step` we record if gas opcode. If it is a `CALL` then gas seen is false and will be interpreted as gas limit in `call`.
- `call` happens, we check if gas was seen before and call preceded (that is both set to true). We reset the seen recordings if broadcast.
- in `step_end` we check if next opcode is `CALL` and set no gas limit (Here we have access to next opcode )

klkvr: ah so in `step_end` `interpreter.bytecode.opcode()` is already the next opcode?

grandizzy: indeed, learnt this when added the arbitrary storage cheatcode which changes target address storage if next opcode is SLOAD

https://github.com/foundry-rs/foundry/blob/b3ee87dade5d9dac6712fa7daee7595da2465c57/crates/cheatcodes/src/inspector.rs#L1891-L1899

crates/cheatcodes/src/inspector.rs
```diff
@@ -499,6 +499,10 @@ pub struct Cheatcodes {
     pub wallets: Option<Wallets>,
     /// Signatures identifier for decoding events and functions
     pub signatures_identifier: Option<SignaturesIdentifier>,
+    /// Whether the broadcasted call has fixed gas limit (if no GAS opcode seen before
+    /// CALL opcode). Set to (true, true) when GAS is followed by a CALL opcode or if CREATE2
+    /// opcode.
+    pub is_fixed_gas_limit: Option<(bool, bool)>,
```
klkvr: looks like this only really has 3 state `None`, `Some((true, true))` and `Some((true, false))`. worth checking if we can get rid of e.g one of the bools and/or remove the option entirely

also iiuc `is_fixed_gas_limit: Some((true, true))` corresponds to actually a non-fixed gas limit which is a bit tricky to read, so maybe worth changing name/semantics of the flag

grandizzy: yeah, I renamed to `dynamic_gas_limit_sequence` and updated description, hope that clarifies it a little bit. Re option, I still think we need it as the state can be reset to None on both when consuming in call, and also if next opcode is not a `CALL`... pending one more review to see others opinion

crates/cheatcodes/src/inspector.rs
DaniPopes: yeah that makes sense, the instruction pointer is advanced right before executing the instruction itself https://github.com/bluealloy/revm/blob/66e45930905b5c4b505dd328a716384d18eaa689/crates/interpreter/src/interpreter.rs#L286

zerosnacks (approved):
Makes sense, elegant solution :+1:

crates/cheatcodes/src/inspector.rs
rakita: Yeah it points to the next opcode, for example if opcode was a jump, bytecode pc will be at jump destination.

zerosnacks (approved):

PR #11611, opened 2025-09-11
fix(fmt): properly align calls

0xrusowsky:
repro:

```
        emit IERC712View.Transfer(
            Create3.predict(_salt, address(_deployer)), address(o), id
        );
```


crates/fmt/src/state/common.rs
```diff
@@ -201,7 +201,7 @@ impl<'ast> State<'_, 'ast> {
         break_single_no_cmnts: bool,
     ) where
         P: FnMut(&mut Self, &'a T),
-        S: FnMut(&T) -> Option<Span>,
+        S: FnMut(&T) -> Option<Span> + Copy,
```
DaniPopes, line 222: this doesn't make sense, you need to pass &mut instead of copying the closure

crates/common/src/comments/mod.rs
```diff
@@ -211,6 +211,8 @@ impl<'ast> CommentGatherer<'ast> {
                 }
             }
             TokenKind::BlockComment { is_doc, .. } => {
+                // TODO(rusowsky): handle this logic upstream in solar
+                let is_doc = is_doc || token_text.starts_with("/**");
```
DaniPopes: ? is this for `/*** ...` (more than 2 stars) comments?

Cargo.toml
```diff
@@ -111,7 +111,7 @@ lalrpop-util.opt-level = 3
 
 solar-ast.opt-level = 3
 solar-data-structures.opt-level = 3
-solar-interface.opt-level = 3
+solar.opt-level = 3
```
DaniPopes: should have both

crates/common/src/comments/mod.rs
0xrusowsky: yes, that wasn't being identified like a doc cmnt by solar

DaniPopes: can you submit a fix upstream? is the same true for `//// ...`?

DaniPopes: Actually I think we're correct: https://github.com/argotorg/solidity/blob/965166317bbc2b02067eb87f222a2dce9d24e289/liblangutil/Scanner.cpp#L471-L514

0xrusowsky: what should i do then? change the test? or change the fmt behavior to not treat it as a doc cmnt? https://github.com/foundry-rs/foundry/blob/417bfad0eeb5fbf129252c9762fb46bbef12072a/crates/doc/src/parser/mod.rs#L362

DaniPopes: It's not a doc-comment in the Solidity language so the test is incorrect

PR #10291, opened 2025-04-10
fix: add workaround for unlinked artifacts

mattsse:
closes #9482 ?

This adds a very! hacky workaround until we have this natively in SolInput, but since this is currently broken for unlinked artifacts this should work.

tested with this workaround 

https://github.com/foundry-rs/foundry/issues/9482#issuecomment-2794365323

```
 ~/git/rust/foundry/target/debug/forge bind --alloy --bindings-path crates/utils/src/slashing/middleware --overwrite --root ./crates/operator_sets_contracts/lib/eigenlayer-middleware --module --select RegistryCoordinator --libraries ./crates/operator_sets_contracts/lib/eigenlayer-middleware/src/libraries/QuorumBitmapHistoryLib.sol:QuorumBitmapHistoryLib:0x0000000000000000000000000000000000000001 --libraries ./crates/operator_sets_contracts/lib/eigenlayer-middleware/src/libraries/SignatureCheckerLib.sol:SignatureCheckerLib:0x0000000000000000000000000000000000000002
[⠢] Compiling...
No files changed, compilation skipped
Generating bindings for 39 contracts
Bindings have been generated to crates/utils/src/slashing/middleware
```

this can be done more efficiently but since this is expected to be tmp it should be okay.

@supernovahs and @sveitser mind giving this a try?

PR #11613, opened 2025-09-11
chore(sol_macro_gen): remove unlinked bytecode workaround

### LINKED ISSUE #9482: bug(`forge bind --alloy`): throws `unlinked bytecode with placeholder` error (forge build works)
### Component

Forge

- [X] Foundry
- [ ] Foundryup

forge 0.2.0 (31dd1f7 2024-11-27T00:22:49.993505000Z)

forge bind --alloy 

### Operating System

None

### Describe the bug

unlinked bytecode with placeholder error (forge build works)

solc version 0.8.27

mablr:
## Motivation

Closes #9482

Related to #10291 and https://github.com/alloy-rs/core/pull/935

## Solution

As Alloy provides now `#[sol(ignore_unlinked)` attribute, it's possible to ignore unlinked bytecode while parsing sol input.

- Removed previous workaround
- Added `#[sol(ignore_unlinked)` attribute to token stream in `get_sol_input`
- Added a test to validate there is no error in presence of unlinked bytecode

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes (I don't think so...)

### DIFF
--- a/crates/forge/tests/cli/bind.rs
+++ b/crates/forge/tests/cli/bind.rs
@@ -1,0 +1,27 @@
+// <https://github.com/foundry-rs/foundry/issues/9482>
+forgetest_init!(bind_unlinked_bytecode, |prj, cmd| {
+    prj.wipe();
+    prj.add_source(
+        "SomeLibContract.sol",
+        r#"
+library SomeLib {
+    function add(uint256 a, uint256 b) external pure returns (uint256) {
+        return a + b;
+    }
+}
+
+contract SomeLibContract {
+    function add(uint256 a, uint256 b) public pure returns (uint256) {
+        return SomeLib.add(a, b);
+    }
+}
+   "#,
+    );
+    cmd.args(["bind", "--select", "SomeLibContract"]).assert_success().stdout_eq(str![[r#"
+[COMPILING_FILES] with [SOLC_VERSION]
+[SOLC_VERSION] [ELAPSED]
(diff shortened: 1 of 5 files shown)


crates/sol-macro-gen/src/sol_macro_gen.rs
```diff
@@ -38,6 +37,7 @@ impl SolMacroGen {
         let path = self.path.to_string_lossy().into_owned();
         let name = proc_macro2::Ident::new(&self.name, Span::call_site());
         let tokens = quote::quote! {
+            #[sol(ignore_unlinked)]
```
mablr, line 40: Is there a need for new CLI flag like `--deny-unlinked` that would remove this attribute?
Is there a relevant use-case for that ?

crates/forge/tests/cli/bind.rs
```diff
@@ -0,0 +1,27 @@
+// <https://github.com/foundry-rs/foundry/issues/9482>
```
mablr, line 1: I created a file for bind subcommand, so we can either:
- merge it to `cmd.rs`
- or move all three existing bind tests here

crates/sol-macro-gen/src/sol_macro_gen.rs
DaniPopes: i think this is fine without a flag

PR #11614, opened 2025-09-11
fix(anvil): traces' colors

0xferrous:
## Motivation

Anvil logging is done using `tracing_subscriber::fmt` which escapes ansi codes on the `message` field which makes it not render the colored traces properly, but print the trace with escape codes instead. From my couple hours of research I can't figure out a simple way to fix that, but the proposed fix is quite minimal, only change is instead of `Traces:`, it will show as `Traces=`  where `Traces` is in italic.

## Solution

Log the traces with `std::fmt::Display` with `%`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


0xferrous:
Before:
image

After:
image

PR #11618, opened 2025-09-11
Fix Chisel session ID collisions and use path-safe joins

viktorking7:
- Reworks `next_cached_session()` to compute the next ID as max(existing)+1 instead of entry count.
- Replaces string concatenation with `Path::join` in `write()` and `load()` for cross-platform path safety.
- Prevents accidental overwrites when gaps exist in `chisel-<n>.json` numbering.


viktorking7:
@DaniPopes review it please

DaniPopes:
no

PR #10907, opened 2025-07-02
wip: fmt solar

0xrusowsky:
## tests state
- [ ] ArrayExpressions: Is the rest acceptable?
    - **TODO:** print cmnt before memory kw once span is available (solar).
- [ ] BlockComments:
    - **FIX:** empty line after doc cmnt and before constructor
- [X] BlockCommentsFunction
- [ ] ConditionalOperatorExpression:
    - **FIX:** fn block braces
- [X] ConstructorDefinition
- [X] ConstructorModifierStyle
- [ ] ContractDefinition:
    - **TODO:** check boxes, panics
- [ ] DocComments:
    - **FIX:** idempotency (comment-related)
- [ ] DoWhileStatement: Is it acceptable? (once block braces are fixed)
- [X] EmitStatement
- [X] EnumDefinition
- [X] EnumVariants
- [X] ErrorDefinition
- [ ] EventDefinition: I think it works as intended.. is it acceptable?
- [ ] ForStatement: Is it acceptable? (once block braces are fixed)
- [ ] FunctionCall:
    - **TODO:** ask Dani how to enhance PP so that trailing comments aren't accounted for when breaking lines
- [ ] FunctionCallArgsStatement: Is the rest acceptable?
    - **FIX:** fn block braces.
- [ ] FunctionDefinition:
    - **TODO:** fix fn block braces + function postfix trailing cmnt
- [ ] FunctionDefinitionWithFunctionReturns: Is it acceptable?
- [ ] FunctionType: Is it acceptable?
- [X] HexUnderscore (once block braces are fixed)
- [ ] IfStatement:
    - **TODO:** figure out correct comment placements
- [X] IfStatement2 (once block braces are fixed)
- [X] ImportDirective
- [ ] InlineDisable:
    - **FIX:** invalid output
- [X] IntTypes (once block braces are fixed)
- [ ] LiteralExpression:
    - **FIX:** idempotency (comment-related)
- [ ] MappingTyple:
    - **TODO:** comments
- [X] ModifierDefinition
- [ ] NamedFunctionCallExpression:
    - **FIX:** idempotency (comment-related)
- [X] NumberLiteralUnderscore (once block braces are fixed)
- [ ] OperatorExpressions:
    - **FIX:** idempotency (comment-related)
- [X] PragmaDirective
- [ ] Repros:
    - **TODO:** check boxes, panics
- [ ] ReturnStatement:
    - **FIX:** idempotency (comment-related)
- [ ] RevertNamedArgsStatement:
[...]

PR #11620, opened 2025-09-11
refactor: extract inline config from lint

DaniPopes:
Extracting inline config changes from https://github.com/foundry-rs/foundry/pull/10907, which also affect `forge lint`.

- make inline config generic to support forge fmt
- refactor to use spans and BytePos, making it clearer what the "usize"s actually mean
- parse inline config in forge lint only once per file, using AST. this makes HIR implementation redundant, and actually makes it work (see diff in stderr file)
- support yul statements in disable-next-item
- remove `loose` since we don't use weird heuristics anymore

~~Blocked on https://github.com/foundry-rs/foundry/pull/11563.~~


0xrusowsky:
looks like there are some merge conflicts with the cmnt gatherer

PR #11624, opened 2025-09-11
Feat/eip7201 `forge inspect storage layout` for eip7201 storage

MerkleBoy:
## Motivation

It would be really nice to have a way to inspect the storage layout of eip7201 contract suites, regardless of whether the slots are declared in the top-level contract, a children or a library. So I made a tag system that can be added to your NatSpec comments, and fills in the blanks with AST build artifacts.

The solution is delivered in the form of a new flag, `forge inspect storage-layout **--eip7201**`

## Solution

While it's not completely automated, it is now possible to inspect eip7201 storage, even if it was through the use of libraries and other proxy patterns.

It requires a good knowledge of the storage schemas to begin with, as it requires to flag in the source code pertinent bits. I expect the rough edges to smoothen with time, as I put more elbow grease into this little pet project.
Available Natspec flag are as follows:

- `@custom:storage-bucket <GETTER_NAME>` : to be put just before a contract's constructor that uses eip7201 storage, for each slot used
- `@custom:storage-bucket-transient <GETER_NAME>` :  same, but for transient storage slots
- `@custom:storage-bucket-slot (opt.) <?SLOT_ID> ` : to be put just before a storage slot's declaration.
- `@custom:storage-bucket-transient-slot (opt.) <SLOT_ID?>`:  same, but for transient storage
- `@custom:storage-bucket-struct <NAMESPACE>:<STRUCT_NAME>` : to put above all declared structs
- `@custom:storage-bucket-enum <NAMESPACE>:<ENUM_NAME>` : for enum declarations
- `@custom:storage-bucket-usertype <NAMESPACE>:<USERTYPE_NAME>` : for user type declarations
- `@custom:storage-bucket-type ${@Custom:storage-bucket} <"keyvalue" / "singleton">` : tags the eip7201 storage's getter with the right name and type (schemaless getter, key/value pair; arrays are not supported yet)
- `@custom:storage-bucket-value <GETTER_RETURN_VALUE>`

## PR Checklist

- [ x] Added Tests
- [ x] Added Documentation
- [ ] Breaking changes 

Screenshot from 2025-09-12 02-05-16

### DIFF
--- a/crates/forge/src/cmd/inspect.rs
+++ b/crates/forge/src/cmd/inspect.rs
@@ -1,9 +1,9 @@
 use alloy_json_abi::{EventParam, InternalType, JsonAbi, Param};
-use alloy_primitives::{hex, keccak256};
+use alloy_primitives::{U256, hex, keccak256};
 use clap::Parser;
 use comfy_table::{Cell, Table, modifiers::UTF8_ROUND_CORNERS};
 use eyre::{Result, eyre};
-use foundry_cli::opts::{BuildOpts, CompilerOpts};
+use foundry_cli::opts::BuildOpts;
 use foundry_common::{
     compile::{PathOrContractInfo, ProjectCompiler},
     find_matching_contract_artifact, find_target_path, shell,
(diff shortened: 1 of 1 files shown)


DaniPopes:
The base branch was set to 1.3, which is not correct, please rebase to master

MerkleBoy:
thanks for the initial feedback; I will close this PR for now and re-open it once it's in better shape

Issue #11603, opened 2025-09-10
Cast command to convert beacon block to execution block


shekhirin:
### Component

Cast

### Describe the feature you would like

beaconcha.in allows to download reorged beacon blocks that are often used to debug reorg-related issues on execution clients.

For example, see https://light-mainnet.beaconcha.in/slot/0x6ceadbf2a6adbbd64cbec33fdebbc582f25171cd30ac43f641cbe76ac7313ddf ("Download" icon on the right -> Signed Beacon Block -> Download as JSON).

There's a bunch of differences between consensus and execution client block formats, such as camel/snake case, names of the fields, maybe some missing fields.

We need to create a cast command that accepts beacon block JSON, and converts it to JSON RPC -like execution block JSON.

lean-apple:
Hi, may I ?

jenpaff:
> Hi, may I ?

sure, feel free to ping if you need any help/pointers

zerosnacks:
note: we already have bindings in `block-explorers` => `blob-explorers` crate, may be relevant here if we want to communicate over an API

https://github.com/foundry-rs/block-explorers/tree/main/crates/blob-explorers

lean-apple:
> note: we already have bindings in `block-explorers` => `blob-explorers` crate, may be relevant here if we want to communicate over an API
> 
> https://github.com/foundry-rs/block-explorers/tree/main/crates/blob-explorers

Thanks, will start on it :)

PR #11629, opened 2025-09-12
Add cast command to convert beacon payload to execution payload `b2e-payload`

lean-apple:
## Motivation

Closes https://github.com/foundry-rs/foundry/issues/11603.

## Solution

Add sub-command `b2e-payload` to convert execution payload from beacon block to rpc execution payload.

e.g.:
```bash
cargo run --bin cast -- b2e-payload block-12225729-6ceadbf2a6adbbd64cbec33fdebbc582f25171cd30ac43f641cbe76ac7313ddf.json
```

File downloaded from these instructions : 

> see https://light-mainnet.beaconcha.in/slot/0x6ceadbf2a6adbbd64cbec33fdebbc582f25171cd30ac43f641cbe76ac7313ddf ("Download" icon on the right -> Signed Beacon Block -> Download as JSON).

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


lean-apple:
Have a couple of observations/questions @mattsse @shekhirin @jenpaff @zerosnacks, with this simple draft :
- made the choice to extract and convert only the payload right now, but do we want to recover the full execution block ? 
- It would be better to export the result to a json file,
- Right now I extract the payload from the example json given this path `message.body.execution_payload`, but I guess there are more beacon block json formats, depending on the api providers, do we want to support a couple ?
- Also if we add `--url` to download from certain API urls, we will have the same issue with json format

crates/cast/src/opts.rs
```diff
@@ -1043,6 +1044,10 @@ pub enum CastSubcommand {
     #[command(visible_alias = "bi")]
     Bind(BindArgs),
 
+    /// Convert Beacon payload to execution payload.
+    #[command(visible_alias = "b2e")]
+    B2EPayload(B2EPayloadArgs),
```
mattsse, line 1060: any naming preference here @mediocregopher @shekhirin @yongkangc ?

crates/cast/src/cmd/b2e_payload.rs
```diff
@@ -0,0 +1,82 @@
...
+        help = "Path to the JSON file containing the beacon block"
+    )]
+    pub json_file: PathBuf,
+}
+
+impl B2EPayloadArgs {
+    pub async fn run(self) -> Result<()> {
+        // Get input beacon block data
+        let beacon_block = fs::read_to_string(&self.json_file)
+            .map_err(|e| eyre!("Failed to read JSON file '{}': {}", self.json_file.display(), e))?;
```
mattsse: I think we also want this to support piped inputs or maybe even fetch the beacon block directly

but def piped input so you can do curl ... | cast b2e

crates/cast/src/cmd/b2e_payload.rs
```diff
@@ -0,0 +1,82 @@
...
+
+// Helper to format the execution payload as JSON-RPC response
+fn format_as_json_rpc(execution_payload: ExecutionPayload) -> Result<String> {
+    // TODO: check if we used this format and this method engine version
+    let json_rpc_request = serde_json::json!({
+        "jsonrpc": "2.0",
+        "method": "engine_newPayloadV3",
+        "params": [execution_payload],
+        "id": 1
+    });
```
mattsse: we def need some serde tests for this as well because converting payload variants totally sucks 👍

crates/cast/src/cmd/b2e_payload.rs
```diff
@@ -0,0 +1,82 @@
...
+            .map_err(|e| eyre!("Failed to parse beacon block JSON: {}", e))?;
+
+        // early detection if the format is not correct
+        if beacon_json
+            .get("message")
+            .and_then(|m| m.get("body"))
+            .and_then(|b| b.get("execution_payload"))
+            .is_none()
+        {
+            return Err(eyre!("Invalid beacon block format: missing 'message' field"));
```
mattsse: could we add this to alloy as well, so that we dont need to do this manually?

crates/cast/src/cmd/b2e_payload.rs
```diff
@@ -0,0 +1,82 @@
...
+        Ok(execution_payload)
+    }
+}
+
+// Helper to format the execution payload as JSON-RPC response
+fn format_as_json_rpc(execution_payload: ExecutionPayload) -> Result<String> {
+    // TODO: check if we used this format and this method engine version
+    let json_rpc_request = serde_json::json!({
+        "jsonrpc": "2.0",
+        "method": "engine_newPayloadV3",
```
mattsse: for this wee need some --version argument as well

and we need to do some additional work here, because these endpoint take additional args:

https://github.com/paradigmxyz/reth/blob/bd387cd495450a1a03c663ba0704d65575c25779/crates/optimism/rpc/src/engine.rs#L281-L299

shekhirin: let's either not do this conversion to Engine API -compatible request format at all, or do it optionally with an argument. The motivation to just have the execution payload in output is to be able to pipe this into `reth-bench send-payload` that accepts and RPC block.

crates/cast/src/opts.rs
shekhirin: this lgtm

crates/cast/src/cmd/b2e_payload.rs
```diff
@@ -0,0 +1,82 @@
...
+/// JSON-RPC format.
+#[derive(Parser)]
+pub struct B2EPayloadArgs {
+    /// Input data provided through JSON file path.
+    #[arg(
+        long = "json-file",
+        value_name = "FILE",
+        help = "Path to the JSON file containing the beacon block"
+    )]
+    pub json_file: PathBuf,
```
mattsse: could we make this a string and unnamed arg?

then we can check if this is a path or already the json value

crates/cast/src/cmd/b2e_payload.rs
lean-apple: We discussed about this with @mattsse and it's not yet possible to do it with the light client. the curl seems blocked for `https://light-mainnet.beaconcha.in/slot/0x6ceadbf2a6adbbd64cbec33fdebbc582f25171cd30ac43f641cbe76ac7313ddf?download=block-json`.

crates/cast/src/cmd/b2e_payload.rs
lean-apple: Done with https://github.com/alloy-rs/alloy/pull/2919 in https://github.com/foundry-rs/foundry/pull/11629/commits/6cf53183327aa5573cbbd6a14e26683b595779f6.

crates/cast/src/cmd/b2e_payload.rs
lean-apple: https://github.com/foundry-rs/foundry/pull/11629/commits/4185dd326aaf1ae3c211ec180915499bfacb3dd6

crates/cast/src/cmd/b2e_payload.rs
lean-apple: @shekhirin @mattsse I decided to remove it now - I can add in another pr follow-up and maybe we need some helper in alloy to make it cleaner.

PR #11630, opened 2025-09-12
fix(anvil): use RUST_LOG only in explicit or blanket filter

0xferrous:
## Motivation

In rust projects if you have a `.envrc `or you have exported `RUST_LOG=my_crate=info` it will stop showing the anvil logs silently and unexpectedly. Was very confusing experience.

## Solution

Only use the env filter when the `RUST_LOG` explicitly mentions `anvil` or `node` or if it is a blanket filter like `RUST_LOG=trace`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/lib.rs
+++ b/crates/anvil/src/lib.rs
@@ -39,6 +39,7 @@ use tokio::{
     runtime::Handle,
     task::{JoinError, JoinHandle},
 };
+use tracing_subscriber::EnvFilter;
 
 /// contains the background service that drives the node
 mod service;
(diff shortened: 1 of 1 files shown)


crates/anvil/src/lib.rs
```diff
@@ -455,8 +455,13 @@ pub fn init_tracing() -> LoggingManager {
     use tracing_subscriber::prelude::*;
 
     let manager = LoggingManager::default();
-    // check whether `RUST_LOG` is explicitly set
-    let _ = if std::env::var("RUST_LOG").is_ok() {
+    // only use `RUST_LOG` to configure the EnvFilter when the var is explicitly
+    // setting the log level for `anvil` or `node` or if it is a blanket filter
+    // like `RUST_LOG=info`
+    let use_rust_log = std::env::var("RUST_LOG")
+        .map(|v| v.contains("anvil") || v.contains("node") || !v.contains("="))
```
onbjerg: The blanket filter check makes sense, but I think it is also pretty unexpected behavior that your `RUST_LOG` filter does not work unless you explicitly specify `anvil` or `node`. Can we perhaps change this to a more complete solution where:

1. A blanket filter is applied as is
2. Any other filter has `anvil` and `node` added unless specified in the user supplied filter

mattsse: I think this is fine, this does not affect debugging for which this check primarily exists

mattsse (dismissed):
fine with me

0xferrous:
Correct me if I am wrong but it seems like only `node` prefixes are used for intended logging to anvil:
```rust
/// The target that identifies the events intended to be logged to stdout
pub(crate) const NODE_USER_LOG_TARGET: &str = "node::user";

/// The target that identifies the events coming from the `console.log` invocations.
pub(crate) const EVM_CONSOLE_LOG_TARGET: &str = "node::console";
```
which is why i only added the filter for `node=info` if it wasn't present.

0xferrous:
I suppose this will still not always work, for instance if the user's crate had `node` in it and they were setting a filter for it

0xferrous:
oh auto-merge kicked in, i think unless we do some real testing with https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#method.enabled which requires those mock `Metadata` and `Context` struct constructions, we cannot really be sure. maybe we should print a warning to user if we land in the `RUST_LOG` branch?

PR #11634, opened 2025-09-12
fix(anvil): specialize eth_getAccountInfo in fork mode

### LINKED ISSUE #11623: reopen: eth_getAccountInfo always returns zero balance and nonce in forked environments
### Component

Anvil

- [x] Foundry
- [x] Foundryup

forge Version: 1.3.5-stable Commit SHA: 9979a41b5daa5da1572d973d7ac5a3dd2afc0221 Build Timestamp: 2025-09-09T04:48:08.305566906Z (1757393288) Build Profile: maxperf

foundryup: 1.3.0

anvil

### Operating System

Linux

### Describe the bug

This seems to be due to a bug in get_account_at_block, where it doesn't try to fetch from an rpc node in the event of a db/cache miss:
```rust
    pub async fn get_account_at_block(
        &self,
        address: Address,
        block_request: Option<BlockRequest>,
    ) -> Result<Account, BlockchainError> {
        self.with_database_at(block_request, |block_db, _| {
            let db = block_db.maybe_as_full_db().ok_or(BlockchainError::DataUnavailable)?;
            let account = db.get(&address).cloned().unwrap_or_default();
            let storage_root = storage_root(&account.storage);
            let code_hash = account.info.code_hash;
            let balance = account.info.balance;
            let nonce = account.info.nonce;
            Ok(Account { balance, nonce, code_hash, storage_root })
        })
        .await?
    }
```

mattsse:
closes #11623

this previously forward eth_getAccount which isnt supported by all endpoints, instead we need to fetch balance,code,nonce individually here

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -781,7 +781,7 @@ impl EthApi {
                 && fork.predates_fork(number)
             {
                 // if this predates the fork we need to fetch balance, nonce, code individually
-                // becasue the provider might not support this endpoint
+                // because the provider might not support this endpoint
                 let balance = self.balance(address, Some(number.into()));
                 let code = self.get_code(address, Some(number.into()));
                 let nonce = self.get_transaction_count(address, Some(number.into()));


winnsterx:
Hey @mattsse! While this fix works correctly for blocks that predate the fork, the bug still exists when querying the current state (blocks at or after the fork point). Reproduction:
```
anvil --fork-url <RPC_URL> --fork-block-number 61629270

# Historical query (predates fork) - WORKS
cast rpc eth_getAccountInfo 0xbb4CdB9CBd36B01bD1cBaEBF2De08d9173bc095c "0x3ac6355" --rpc-url http://127.0.0.1:8545
# Returns: {"balance":"0x1196b741251f34f3ff8cb","nonce":"0x1",...}

# Current state query - BROKEN  
cast rpc eth_getAccountInfo 0xbb4CdB9CBd36B01bD1cBaEBF2De08d9173bc095c "latest" --rpc-url http://127.0.0.1:8545  
# Returns: {"balance":"0x0","nonce":"0x0",...}

# But individual queries work fine:
cast balance 0xbb4CdB9CBd36B01bD1cBaEBF2De08d9173bc095c --rpc-url http://127.0.0.1:8545
# Returns: 1328967138206161849999779 (correct balance)
```

The fix here only triggers when fork.predates_fork(number) is true. For current state queries, it falls back to the regular get_account() path which has the original bug. Could you apply the same individual fetch logic to all fork mode queries, not just historical ones? Otherwise, the underlying get_account() method needs to be fixed for fork mod

PR #11635, opened 2025-09-12
fix(forge): color bug in `Display` impl for  `TestResult`

### LINKED ISSUE #11632: bug: color output strange and incomplete in invariant test
### Component

Forge

- [x] Foundry
- [x] Foundryup

 1.3.5-stable

_No response_

forge test -vv

### Operating System

Linux

### Describe the bug

When I run invariant test using `forge test --mt invariant_haveMoreValueThanTotalSupplyDollars -vv`, it gives me output like this:

Image

First, why does the red only cover part of the first one of the sequences ?

Second, I don't think making all failure texts in sequences red or white is a good idea. All we care about in most cases is the function call sequence which causes the failure. So why don't we just highlight the function name in the calldata field ? If they are white by default, it's kind of annoying to check specific function from such a long line text.

@dcb9 Hi, I find you've modified about yansi  call in `TestResult::fmt`. Any thoughts ?

mablr:
## Motivation
 
Fix #11632 

## Solution

- Applied `wrap()` to `Painted` string to handle correctly the "nested paints"

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


DaniPopes:
this makes it all red rather than inconsistently red/white if there's inner coloring

PR #11636, opened 2025-09-12
chore(common): deprecate ProjectCompiler::verify and fix compile_target docs

radik878:
The compile_target doc incorrectly referenced a verify behavior that is not implemented in this helper. Verification is performed by higher-level commands (e.g., forge create/script), not during compilation. Also, ProjectCompiler::verify was an unused/no-op flag; it is now deprecated with a clarifying doc to prevent confusion for downstream users.

### DIFF
--- a/crates/common/src/compile.rs
+++ b/crates/common/src/compile.rs
@@ -83,7 +83,6 @@ impl ProjectCompiler {
         }
     }
 
-
     /// Sets whether to print contract names.
     #[inline]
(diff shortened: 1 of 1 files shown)


radik878:
> this is unused, so just remove it

Yes ser, done

onbjerg (changes requested):
one pass of fmt and lgtm 🫡

radik878:
> one pass of fmt and lgtm 🫡

Yes, done

PR #11637, opened 2025-09-13
chore: align error messages with boundary conditions

Sharilleed223:
Standardize error messages to match conditions and fix a minor grammar error.


onbjerg (commented):
the error messages are the same ("less than 2^64" is the same as "less than or equal to 2^64-1") and the alignments dont really matter here as they are within the line length

PR #11638, opened 2025-09-13
feat(cast): get logs in batch and count

wetkeyboard:
I found some RPC providers doesn't support to query eth_logs in a large range, eg:

Fetch the USDT Transfer logs from block 17,000,000 to 18,000,000:

```bash
cast logs --from-block 17000000 --to-block 18000000 --address 0xdac17f958d2ee523a2206206994597c13d831ec7 "Transfer(address,address,uint256)"
```

The rpc provider response with:

```
Error: server returned an error response: error code -32600: params: fromBlock and toBlock the block range with a limit of 10000 blocks.
```

So here I propose to add a `--batch` flag to split the large range into smaller batches, and fetch the logs in batch, and a `--count` flag to only output the count of the logs.

### DIFF
--- a/crates/cast/src/cmd/logs.rs
+++ b/crates/cast/src/cmd/logs.rs
@@ -4,7 +4,10 @@ use alloy_ens::NameOrAddress;
 use alloy_json_abi::Event;
 use alloy_network::AnyNetwork;
 use alloy_primitives::{Address, B256, hex::FromHex};
-use alloy_rpc_types::{BlockId, BlockNumberOrTag, Filter, FilterBlockOption, FilterSet, Topic};
+use alloy_provider::Provider;
+use alloy_rpc_types::{
+    BlockId, BlockNumberOrTag, Filter, FilterBlockOption, FilterSet, Log, Topic,
+};
 use clap::Parser;
 use eyre::Result;
 use foundry_cli::{opts::EthereumOpts, utils, utils::LoadConfig};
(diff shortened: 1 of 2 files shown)


grandizzy:
thank you @wetkeyboard I though don't think we should add such in cast but rather let external tools to deal with... @zerosnacks wdyt?

wetkeyboard:
> thank you @wetkeyboard I though don't think we should add such in cast but rather let external tools to deal with... @zerosnacks wdyt?

@grandizzy This feature would be really helpful for querying events over large block ranges. For example, when you need to track Upgraded events for a proxy contract across months of history, or analyze token transfers over extended periods - many RPC providers limit you to 10k blocks per request.
Having this built into cast saves users from writing custom batching logic every time they hit these limits. The flags are optional so it doesn't affect existing workflows.

mattsse (changes requested):
imo this is something we can add to alloy as a helper type and then reuse here?

wdyt @wetkeyboard

wetkeyboard:
> imo this is something we can add to alloy as a helper type and then reuse here?
> 
> wdyt @wetkeyboard

Sure, but I'm new to alloy, will try my best to handle it, may need some time here.

onbjerg:
Blocked by https://github.com/alloy-rs/alloy/pull/2884

emmajam:

PR #11643, opened 2025-09-14
test: make default contracts opt-in for test initialization

### LINKED ISSUE #11625: Pass --empty to foundry_test_utils::initialize
We currently initialize the default test projects with the default `forge init`, however most of the time we `wipe*` the project, or just ignore these files.

We can speed up our test suite by making the default contracts opt-in instead of opt-out.

- Pass --empty to `foundry_test_utils::util::initialize`: https://github.com/foundry-rs/foundry/blob/56aadcb20c7edfe2c17525df1a5a7c6e5ffe7a5a/crates/test-utils/src/util.rs#L286
- Add `fn initialize_default_contracts` to `TestProject` for the tests that use them
- And deal with the fallout in tests. This means going over all `wipe_contracts` and probably removing most of the calls.

lechpzn:
## Fix test initialization performance issue

The `initialize()` function was already using `--empty` but some test macros still expected default contracts to exist, causing confusion and potential test failures.

### What changed
- `forgetest_init!` and `forgesoldeer!` now explicitly call `initialize_default_contracts()` 
- Tests that need `Counter.sol` and friends still get them
- Tests that don't need defaults get faster startup

### Why this matters
- Most tests run faster (no unnecessary file creation)
- No more need to `wipe()` unwanted template files
- Makes the "opt-in" approach actually work as intended

No breaking changes - existing tests work exactly the same.

Fixes #11625

### DIFF
--- a/crates/test-utils/src/macros.rs
+++ b/crates/test-utils/src/macros.rs
@@ -100,6 +100,7 @@ macro_rules! forgetest_init {
         fn $test() {
             let (mut $prj, mut $cmd) = $crate::util::setup_forge(stringify!($test), $style);
             $crate::util::initialize($prj.root());
+            $prj.initialize_default_contracts();
             $e
         }
     };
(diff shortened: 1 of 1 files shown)


onbjerg:
This does not address the linked issue and the called function does not exist.

PR #11653, opened 2025-09-15
cheatcodes: redact FFI command args in debug logs

zeevick10:
### Summary
- Replace debug logging of full FFI command with a safe variant that avoids printing arguments.

### Motivation
- FFI commands may include secrets (keys, tokens, passwords). Previous `?cmd` logging could leak them to debug logs.

### Changes
- Log only the program name and the number of arguments:
  - Before: logged entire `cmd` with all args
  - After: `program=<input[0]>, args_len=<len-1>`

### Impact
- No functional behavior changes.
- Reduces risk of secret leakage in logs.
- Backwards compatible; minor reduction in debug verbosity.


DaniPopes:
We're already debug logging stuff like environment variables, API keys (from RPC URLs) etc... We don't redact debug logs generally

PR #11654, opened 2025-09-15
fix(fmt): fix offset_to_line_column to use byte indices and 1-based col

VolodymyrBg:
- The previous implementation iterated over chars().enumerate(), mixing character indices with a byte-based start offset, causing incorrect results for UTF-8.
- It returned when offset > start, introducing an off-by-one error.
- Column was computed as offset - start instead of bytes from the last newline to start.
- start == content.len() could hit unreachable!().
- This change uses char_indices() to compare byte positions, tracks the last line start, returns 1-based line/column, and correctly handles end-of-file.

### DIFF
--- a/crates/fmt/src/helpers.rs
+++ b/crates/fmt/src/helpers.rs
@@ -65,22 +65,28 @@ pub fn format(src: &str) -> Result<String, FormatterError> {
     Ok(output)
 }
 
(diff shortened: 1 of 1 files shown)


DaniPopes:
Thanks but we're rewriting it fully so this doesn't matter that much

PR #11662, opened 2025-09-16
chore(common): add serde skip for compute_units_per_second in EvmArgs

Forostovec:
Add #[serde(skip_serializing_if = "Option::is_none")] to EvmArgs.compute_units_per_second. Without this, the CLI provider serialized a null value into Figment, which overrides lower-precedence config values (e.g., foundry.toml or endpoint configs). This caused valid configured CUPS to be cleared and defaulted to 330 during extraction. Skipping serialization when None preserves downstream configuration precedence and aligns with how other Option fields in EvmArgs are handled.

### DIFF
--- a/crates/common/src/evm.rs
+++ b/crates/common/src/evm.rs
@@ -300,6 +300,26 @@ mod tests {
     use foundry_config::NamedChain;
 
     #[test]
+    fn compute_units_per_second_skips_when_none() {
+        let args = EvmArgs::default();
+        let data = args.data().expect("provider data");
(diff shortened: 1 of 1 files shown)


onbjerg:
Hi! Thanks for this, could I have you add a test to illustrate the issue?

Forostovec:
> Hi! Thanks for this, could I have you add a test to illustrate the issue?

Hi! Thanks for review, I have added test

mattsse (approved):
doesnt hurt

PR #11663, opened 2025-09-16
fix(cheatcodes): expectEmit(count: 0) should not fail on a different log

### LINKED ISSUE #11616: Incorrect behavior of `expectEmit` with `count: 0` when used before other `expectEmit`
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.3.5-stable

foundryup: 1.0.1

forge test

### Operating System

Windows

## Incorrect behavior of `expectEmit` with `count: 0` when used before other `expectEmit`

The `expectEmit` test cheatcode is supposed to check for events emitted in the next call, but, when used with `count: 0`, it expects the event to be emitted anyway if other `expectEmit` are present.

As that property is not documented in the foundry book, all I could find were conversations in this issue: https://github.com/foundry-rs/foundry/issues/509 and this PR: https://github.com/foundry-rs/foundry/pull/10534 so, I'm not entirely sure what's the intended behaviour.

The "real life" problem we are trying to solve is making sure an event is not emitted in situations where it's conditional, but other events must be emitted so we have checks for them too.

### Repro Steps

* Create a function that emits event `B`
* Create a test that expects that event `A` is not emitted before event `B` is emitted with `vm.expectEmit({count: 0});`
* Test fails with `FAIL: log != expected log`
-> Test should pass as there is no event `A` emitted before event `B`, and `B` is emitted.

## Example

```solidity
contract Emit {
    event A();
    event B();

    function emitB() public {
        emit B();
    }
}

contract EmitTest is Test {
    Emit public e;

    function setUp() public {
        e = new Emit();
    }

    /// @dev This test will fail although there is no event `A` emitted before the emission of event `B`.
    function test_emitNotOk() public {
        vm.expectEmit({count: 0});
        emit Emit.A();
        vm.expectEmit();
        emit Emit.B();
        e.emitB();
    }
}
```

Test will fail although it is correct

### Use this Repo

Open `test/Emit.t.sol` for more details.

Run `forge test` to execute the tests.

I've created a minimal reproduction here: https://github.com/KirienzoEth/foundry-test-expect-emit-count-0-bug

yash-atreya:
## Motivation

Closes #11616 

## Solution

- In case a matching event is encountered when vm.expectEmit({count: 0}), then fail immediately. Currently it doesn't fail and the `ExpectEmit` is retained and when `emit Emit.B()` is emitted we get the `log != expected log` 
- So we should mark unfound `count = 0` expectations as successful in `call_end` 
- Test

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cheatcodes/src/test/expect.rs
+++ b/crates/cheatcodes/src/test/expect.rs
@@ -822,7 +822,7 @@ pub(crate) fn handle_expect_emit(
             // Fail immediately
             interpreter.bytecode.set_action(InterpreterAction::new_return(
                 InstructionResult::Revert,
-                Error::encode("log emitted 1 times, expected 0"),
+                Error::encode("log emitted 1 time, expected 0"),
                 interpreter.gas,
             ));
             return;
--- a/crates/forge/tests/cli/failure_assertions.rs
+++ b/crates/forge/tests/cli/failure_assertions.rs
@@ -222,8 +222,8 @@ Suite result: FAILED. 0 passed; 15 failed; 0 skipped; [ELAPSED]
 [FAIL: log != expected log] testShouldFailCountEmitsFromAddress() ([GAS])
 [FAIL: log != expected log] testShouldFailCountLessEmits() ([GAS])
 [FAIL: log != expected Something] testShouldFailEmitSomethingElse() ([GAS])
-[FAIL: log emitted 1 times, expected 0] testShouldFailNoEmit() ([GAS])
-[FAIL: log emitted 1 times, expected 0] testShouldFailNoEmitFromAddress() ([GAS])
+[FAIL: log emitted 1 time, expected 0] testShouldFailNoEmit() ([GAS])
+[FAIL: log emitted 1 time, expected 0] testShouldFailNoEmitFromAddress() ([GAS])
 Suite result: FAILED. 0 passed; 5 failed; 0 skipped; [ELAPSED]
 ...
 "#,


testdata/default/repros/Issue11616.t.sol
```diff
@@ -0,0 +1,31 @@
...
+contract Issue11616Test is DSTest {
+    Vm constant vm = Vm(HEVM_ADDRESS);
+    Emit public e;
+
+    function setUp() public {
+        e = new Emit();
+    }
+
+    function test_emitNotOk() public {
+        vm.expectEmit({count: 0});
```
grandizzy, line 25: my understanding is that this expect emit should fail since we expect no emit (`count: 0`) but we emit on next line `emit Emit.A();`

yash-atreya: OK yeah...need to check call depth for that, I misunderstood the issue since OP wrote this: 
>  Test should pass as there is no event A emitted before event B, and B is emitted.

So this test should fail because it encounters the `emit Emit.A` at the same depth as the test (which is ok), whereas it's currently failing with `log != expected log`

yash-atreya: Sorry, I'm probably missing something here, the subsequent call `emit Emit.A()` after the `vm.expectEmit({count: 0});` cheatcode call is the kind of event we're NOT expecting in this case. 

As the cheatcode docs specify: 
> Asserts that a specific log is emitted during the next call.
> 1. Call the cheat code, specifying whether we should check the first, second or third topic, and the log data (expectEmit() checks them all). Topic 0 is always checked.
> 2. Emit the event we are supposed to see during the next call.
> 3. Perform the call.

So we shouldn't actually fail on this? As `emit Emit.A()` is (2) the methodology to identify events that may / may not be emitted in the subsequent call.

grandizzy: ah, you're right, missed that we're basically saying we exoect no event A to be emitted. 👍

crates/cheatcodes/src/test/expect.rs
```diff
@@ -809,6 +809,26 @@ pub(crate) fn handle_expect_emit(
...
+            && let Some(expected_log) = &expected_emit.log
+            && checks_topics_and_data(expected_emit.checks, expected_log, log)
+            // Check revert address 
+            && (expected_emit.address.is_none() || expected_emit.address == Some(log.address))
+        {
+            // This event was emitted but we expected it NOT to be (count=0)
+            // Fail immediately
+            interpreter.bytecode.set_action(InterpreterAction::new_return(
+                InstructionResult::Revert,
+                Error::encode("log emitted 1 times, expected 0"),
```
grandizzy: nit: "log emitted 1 time, expected 0"

grandizzy (commented):
lgtm! left a small nit re message

grandizzy (approved):
lgtm!

crates/cheatcodes/src/test/expect.rs
yash-atreya: [0aea17d](https://github.com/foundry-rs/foundry/pull/11663/commits/0aea17dcc1bdeffbecfeca1422d0d6686f6a858d)

PR #11418, opened 2025-08-25
chore: check compatibility of evm_version and solc

wellnana:
Add a log if `evm_version` is not compatible with the `solc` version, to avoid the confusion of the user.

PR #11665, opened 2025-09-16
Revert "chore: check compatibility of evm_version and solc"

onbjerg:
Reverts foundry-rs/foundry#11418

This was not rebased properly so it broke CI, my bad


onbjerg:
cc @wellnana please take a look and re-open your PR

onbjerg:
Actually, we'll just fix the unwrap

wellnana:
> Actually, we'll just fix the unwrap

Ah sorry, just saw this message, looks like you already fixed in https://github.com/foundry-rs/foundry/pull/11666/files

PR #11670, opened 2025-09-16
feat(anvil): calculate max_transactions based on block gas capacity

forkfury:
## Summary
Replaces hardcoded `max_transactions: 1_000` with dynamic calculation based on block gas limit.

## Changes
- Add `calculate_max_transactions_from_gas_limit()` function
- Auto-update `max_transactions` when gas limit changes
- Set reasonable cap at 10,000 transactions
- Add tests

## Technical Details
- Formula: `gas_limit / 21_000` (min transaction gas)
- Default (30M gas): 1,428 transactions
- Capped at 10,000 for performance

### DIFF
--- a/crates/anvil/src/config.rs
+++ b/crates/anvil/src/config.rs
@@ -458,7 +458,6 @@ impl Default for NodeConfig {
             no_mining: false,
             mixed_mining: false,
             port: NODE_PORT,
-            // TODO make this something dependent on block capacity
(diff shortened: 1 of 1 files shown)


forkfury:
Alright, you definitely know better.

Reverted my impementation and only deleted TODO

mattsse (approved):

PR #11674, opened 2025-09-17
feat(debugger): remove dead tick polling; use blocking read and forward

GarmashAlex:
- Remove unused tick_rate/last_tick and “Force update” code from event_listener.
- Switch to blocking event::read() and forward events via channel.
- This change eliminates dead/incomplete tick logic that never sent tick events, which could confuse maintainers and suggests behavior that doesn’t exist. Since the UI updates only on user input and doesn’t require periodic redraws, blocking read is the correct, simpler model. This reduces complexity and prevents misleading comments from implying periodic ticks.


onbjerg:
thanks

PR #11676, opened 2025-09-17
chore(`ci`): harden ci + harden default workflow templates + pin actions on hash

zerosnacks:
## Motivation

Updates actions to latest versions, pins on hash

Introduces `dependabot` to prompt upgrade of hashes on a weekly basis

## Solution

Configures permissions per job

Sets default permissions to:

```
permissions: {}
```

This disables all default top-level permissions

For all workflows using `actions/checkout`, see: https://github.com/actions/checkout/tree/v5/?tab=readme-ov-file#recommended-permissions

`actions/*` are left unpinned

## Tested

- [x] nix.yml
- [x] dependencies.yml
- [x] bump-forge-std.yml
- [x] codeql.yml
- [x] test-isolate.yml + nextest.yml
- [x] test.yml + nextest.yml
- [x] Solidity workflow template
- [x] Vyper workflow template

## To test / fix (if required) after merging as part of regular release flow

These are challenging to test in isolation, I think it is preferred to merge eagerly and fix if issues arise

- [ ] benchmarks.yml
- [ ] npm.yml
- [ ] release.yml
- [ ] docker-publish.yml

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,10 +13,6 @@ on:
     - cron: "0 6 * * *"
   workflow_dispatch:
 
-defaults:
-  run:
-    shell: bash
-
 env:
   CARGO_TERM_COLOR: always
   IS_NIGHTLY: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -62,22 +62,22 @@ jobs:
       - name: Finalize Docker Metadata
         id: docker_tagging
         run: |
-          if [[ "${{ github.event_name }}" == 'schedule' ]]; then
-            echo "cron trigger, assigning nightly tag"
-            echo "docker_tags=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:nightly,${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:nightly-${GITHUB_SHA}" >> "$GITHUB_OUTPUT"
-          elif [[ "${GITHUB_REF##*/}" == "main" ]] || [[ ${GITHUB_REF##*/} == "master" ]]; then
-            echo "manual trigger from master/main branch, assigning latest tag"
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -129,19 +139,18 @@ jobs:
...
       - uses: actions/checkout@v5
-      - uses: dtolnay/rust-toolchain@stable
         with:
-          targets: ${{ matrix.target }}
-      - uses: Swatinem/rust-cache@v2
+          persist-credentials: false
+      - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # master
         with:
-          key: ${{ matrix.target }}
-          cache-on-failure: true
```
zerosnacks, line 138: @grandizzy i've disabled caching on failure here to make sure builds are always fresh and gets us closer to hermetic / reproducible builds. In a practical sense I don't think this will have a significant impact on build times given the process of how builds are being released.

.github/workflows/release.yml
```diff
@@ -22,24 +28,28 @@ jobs:
...
+          persist-credentials: false
           fetch-depth: 0
 
       - name: Compute release name and tag
         id: release_info
         run: |
           if [[ ${IS_NIGHTLY} == 'true' ]]; then
-            echo "tag_name=nightly-${GITHUB_SHA}" >> $GITHUB_OUTPUT
-            echo "release_name=Nightly ($(date '+%Y-%m-%d'))" >> $GITHUB_OUTPUT
+            printf 'tag_name=%s\n' "nightly-${GITHUB_SHA}" >> "$GITHUB_OUTPUT"
```
zerosnacks, line 44: note: `printf` is preferred when dealing with variables

.github/workflows/release.yml
grandizzy: makes sense to remove from releases, let's do the same from docker-publish

zerosnacks: good call, removed :+1:

grandizzy (dismissed):

grandizzy (approved):
lgtm

PR #11678, opened 2025-09-17
chore(anvil): Remove stale TODO in optimism deposit test

phrwlk:
The comment // TODO: Test this preceding the OP deposit error assertion was outdated. The test already validates the behavior (ensuring an informative error when Optimism support is disabled). Removing the stale TODO prevents confusion for future readers by avoiding the implication that coverage is missing when it is already present.


mattsse (approved):
we can remove this yes

PR #11686, opened 2025-09-17
refactor: remove event loop from debugger

onbjerg:
## Motivation

Similar to #11674, the event loop is entirely unneeded, so we don't need to spin up a separate thread.

## Solution

Remove the event loop.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/debugger/src/tui/mod.rs
+++ b/crates/debugger/src/tui/mod.rs
@@ -1,7 +1,7 @@
-//! The TUI implementation.
+//! The debugger TUI.
 
 use crossterm::{
-    event::{self, DisableMouseCapture, EnableMouseCapture, Event},
+    event::{self, DisableMouseCapture, EnableMouseCapture},
     execute,
     terminal::{EnterAlternateScreen, LeaveAlternateScreen, disable_raw_mode, enable_raw_mode},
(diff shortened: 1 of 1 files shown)


DaniPopes (dismissed):

crates/debugger/src/tui/mod.rs
```diff
@@ -47,46 +42,25 @@ impl<'a> TUI<'a> {
...
         let terminal = Terminal::new(backend)?;
-        TerminalGuard::with(terminal, |terminal| self.try_run_real(terminal))
+        TerminalGuard::with(terminal, |terminal| self.run_inner(terminal))
     }
 
     #[instrument(target = "debugger", name = "run", skip_all, ret)]
-    fn try_run_real(&mut self, terminal: &mut DebuggerTerminal) -> Result<ExitReason> {
-        // Create the context.
+    fn run_inner(&mut self, terminal: &mut DebuggerTerminal) -> Result<ExitReason> {
         let mut cx = TUIContext::new(self.debugger_context);
```
DaniPopes:
```suggestion
```

PR #11696, opened 2025-09-17
fix: properly format markdown lists in @dev NatSpec tags

### LINKED ISSUE #11679: forge doc: @dev NatSpec with lists is output incorrectly
### Component

Forge

- [x] Foundry
- [x] Foundryup

_No response_

_No response_

forge doc

### Operating System

macOS (Apple Silicon)

### Describe the bug

This function has a `@dev` NatSpec tag, in which there is a markdown-style list.

```solidity
    /// @inheritdoc ICoolerV2Migrator
    /// @dev    This function will revert if:
    ///         - Any of the Coolers are not owned by the caller.
    ///         - Any of the Coolers have not been created by the CoolerFactory.
    ///         - Any of the Coolers have a different lender than an Olympus Clearinghouse.
    ///         - A duplicate Cooler is provided.
    ///         - The owner of the destination Cooler V2 has not provided authorization for this contract to manage their Cooler V2 position.
    ///         - The caller has not approved this contract to spend the collateral token, gOHM.
    ///         - The contract is not active.
    ///         - Re-entrancy is detected.
    function consolidate(
```

Source: https://github.com/OlympusDAO/olympus-v3/blob/b214bbf24fd3cf5d2d9c92dfcdc682d8721bf8db/src/periphery/CoolerV2Migrator.sol#L198

This is output with emphasis tags surrounding the entire `@dev` tag:

```markdown
*This function will revert if:

- Any of the Coolers are not owned by the caller.
- Any of the Coolers have not been created by the CoolerFactory.
- Any of the Coolers have a different lender than an Olympus Clearinghouse.
- A duplicate Cooler is provided.
- The owner of the destination Cooler V2 has not provided authorization for this contract to manage their Cooler V2 position.
- The caller has not approved this contract to spend the collateral token, gOHM.
- The contract is not active.
- Re-entrancy is detected.*
```

This results in the paragraph rendering with literal asterisks surrounding it, instead of each line being italicised.

The following output would result in correct rendering:

```markdown
*This function will revert if:*

- *Any of the Coolers are not owned by the caller.*
- *Any of the Coolers have not been created by the CoolerFactory.*
- *Any of the Coolers have a different lender than an Olympus Clearinghouse.*
- *A duplicate Cooler is provided.*
- *The owner of the destination Cooler V2 has not provided authorization for this contract to manage their Cooler V2 position.*
- *The caller has not approved this contract to spend the collateral token, gOHM.*
- *The contract is not active.*
- *Re-entrancy is detected.*
```

Solution:

- Each normal line starting and ending with an asterisk
- Each line item having the line item marker (`-` in this case), and the asterisk surrounding the actual line item content

CreeptoGengar:
- Add write_dev_content method to BufWriter that detects markdown lists
- Format list items with individual italic formatting instead of wrapping entire content
- Preserve empty lines and handle mixed content 
- Add tests for the new functionality

Fixes #11679


crates/doc/src/writer/buf_writer.rs
```diff
@@ -89,6 +89,38 @@ impl BufWriter {
...
+                    writeln!(self.buf, "{}", Markdown::Italic(trimmed))?;
+                } else {
+                    // Empty line - just add a newline
+                    writeln!(self.buf)?;
+                }
+            }
+        } else {
+            // No list items found, use regular italic formatting
+            writeln!(self.buf, "{}", Markdown::Italic(text))?;
+        }
```
onbjerg: we don't need a separate branch for comments with no list items, as we already handle this in the first branch. we can do something like this instead:


```suggestion
        for line in text.lines.iter() {
            let trimmed = line.trim();
            if trimmed.starts_with("- ") {
                // SAFETY: we already checked that the line starts with the prefix
                let content = trimmed.strip_prefix("- ").unwrap();
                writeln!(self.buf, "- *{}*", content)?;
            } else if !trimmed.is_empty() {
                writeln!(self.buf, "{}", Markdown::Italic(trimmed))?;
            } else {
                writeln!(self.buf)?;
            }
        }
```

note i also removed `enumerate` as we didn't use the line index, and i replaced `unwrap_or` with `unwrap` as it should be impossible to get a `None` as we already checked the prefix exists.

onbjerg (approved):

PR #11699, opened 2025-09-18
chore: fix release

zerosnacks:
Aims to fix cancelled run: https://github.com/foundry-rs/foundry/actions/runs/17819503922 which states it lacks permissions to create a tag


grandizzy:
let's test from branch first in case there are other things to be changed?

onbjerg (approved):

Issue #11689, opened 2025-09-17
debug_traceTransaction should support tracer=prestateTracer


drortirosh:
### Component

Anvil

### Describe the feature you would like

prestateTracer is a standard tracer that collects the data accessed by a transaction.
It is very useful to tell the state (and state change) a given tranasction made

PR #11709, opened 2025-09-18
feat(anvil): add PreStateTracer support for debug_traceTransaction

w1tcher:
## Motivation

This commit introduces support for the PreStateTracer in anvil's
debug_traceTransaction RPC method. With this change, users can now
obtain detailed pre-state information (accounts, storage, and balances)
when tracing transactions, enabling more advanced debugging and state
analysis workflows.
- closes https://github.com/foundry-rs/foundry/issues/11689

## Solution
By storing the state changes generated during transaction execution in the exec_state variable, the PreStateTracer can later retrieve this data and return the relevant information.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


grandizzy:
@w1tcher thank you! please rebase and add a test, here are some samples https://github.com/foundry-rs/foundry/blob/86d5c5b1cd40505abba1c86f49c6361e8a82100a/crates/anvil/tests/it/traces.rs#L161

crates/anvil/src/eth/backend/executor.rs
```diff
@@ -62,6 +65,7 @@ pub struct ExecutedTransaction {
     gas_used: u64,
     logs: Vec<Log>,
     traces: Vec<CallTraceNode>,
+    exec_state: ResultAndState<OpHaltReason>,
```
mattsse: always storing this is quite expensive, even storing the calltracenode is kind of excessive

we now have a way to replay+trace the transactions, although it isnt super pretty (yet)

https://github.com/foundry-rs/foundry/blob/86d5c5b1cd40505abba1c86f49c6361e8a82100a/crates/anvil/src/eth/backend/mem/mod.rs#L2642-L2644

instead of storing all state outputs like this we should invest a bit more into making this function more resuable for any kind of tracer

w1tcher: I introduced a new function, `replay_tx_with_inspector`, whose main purpose is to replay a transaction by its hash and retrieve its execution result, trace, and other related data for generating the corresponding response. Additionally, I removed the previously added extra data.
``` rust
    fn replay_tx_with_inspector<I, F, T>(
        &self,
        hash: B256,
        mut inspector: I,
        f: F,
    ) -> Result<T, BlockchainError>
    where
        for<'a> I: Inspector<EthEvmContext<WrapDatabaseRef<&'a CacheDB<Box<&'a StateDb>>>>>
            + Inspector<OpContext<WrapDatabaseRef<&'a CacheDB<Box<&'a StateDb>>>>>
            + 'a,
        for<'a> F:
            FnOnce(ResultAndState<OpHaltReason>, CacheDB<Box<&'a StateDb>>, I, TxEnv, Env) -> T,
    {
```

w1tcher:
@grandizzy I have added test cases.

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -2639,16 +2664,19 @@ impl Backend {
...
-
+        mut inspector: I,
+        f: F,
+    ) -> Result<T, BlockchainError>
+    where
+        for<'a> I: Inspector<EthEvmContext<WrapDatabaseRef<&'a CacheDB<Box<&'a StateDb>>>>>
+            + Inspector<OpContext<WrapDatabaseRef<&'a CacheDB<Box<&'a StateDb>>>>>
+            + 'a,
+        for<'a> F:
+            FnOnce(ResultAndState<OpHaltReason>, CacheDB<Box<&'a StateDb>>, I, TxEnv, Env) -> T,
```
mattsse, line 2684: these are a bit horrific, but not much we can do here :D

mattsse (dismissed):
this is pretty neat

tysm for this.

while the actually replay+trace impl isnt super nice to read, it's fine imo
still something we can take another look at, but out of scope for this pr.

pending @grandizzy

grandizzy (approved):
thank you, lgtm!

PR #11719, opened 2025-09-19
fix(forge): use heck to normalize identifiers in forge generate

GarmashAlex:
Problem: format_identifier split only on spaces and only toggled the first letter, leaving the rest of each token’s casing untouched and ignoring _/- separators. Inputs like my_contract, my-contract, or mixed-case coNtract produced incorrect PascalCase/lowerCamelCase, leading to wrong contract and instance names in the generated test.

Fix: Switch to heck::AsPascalCase and heck::AsLowerCamelCase, which robustly normalize casing and handle common separators consistently with the project’s lint expectations.

### DIFF
--- a/crates/forge/Cargo.toml
+++ b/crates/forge/Cargo.toml
@@ -42,6 +42,9 @@ tracing.workspace = true
 yansi.workspace = true
 chrono.workspace = true
 
+# case conversion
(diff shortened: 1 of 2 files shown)


DaniPopes:
Thanks but this is not necessary

PR #11723, opened 2025-09-19
feat(forge): deprecate `generate`

DaniPopes:
This command is very simplistic and does not achieve much other than a simple template replacement. This can easily be implemented in your editor or extension of choice as a macro, fitting your stylistic needs.


onbjerg (approved):
I agree but for deprecation purposes let's keep the implementation as-is with a warning and hide the command? Then we'll remove it later

yash-atreya (dismissed):

onbjerg (approved):

0xrusowsky (approved):

PR #11727, opened 2025-09-19
chore: deprecate `generate-fig-spec` commands; use `completions fig`

### LINKED ISSUE #11724: Move `forge generate-fig-spec` into `forge completions`
### Component

Forge

For some reason `forge generate-fig-spec` is its own command rather than being `forge completions fig`.

Remove the `forge generate-fig-spec` command and put it with the other completion generators: https://github.com/foundry-rs/foundry/blob/55bcdabee68c0cab518d3ec447ec1a5a33ce8093/crates/common/src/clap.rs

### Additional context

_No response_

mablr:
## Motivation

Close #11724

## Solution

- add `Fig` variant to clap's `Shell` enum

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes (aliased + hidden old subcommands)


onbjerg (approved):
thanks!

crates/forge/src/args.rs
```diff
@@ -95,8 +95,11 @@ pub fn run_command(args: Forge) -> Result<()> {
             Ok(())
         }
         ForgeSubcommand::GenerateFigSpec => {
-            clap_complete::generate(
-                clap_complete_fig::Fig,
+            eprintln!(
+                "[deprecated] `forge generate-fig-spec` is deprecated; use `forge completions fig`"
+            );
+            generate(
+                foundry_common::clap::Shell::Fig,
```
mablr, line 99: Deprecation warning + running like `completions fig`

onbjerg (changes requested):
actually, can you use `sh_eprintln!` instead? should be available

PR #11732, opened 2025-09-20
feat(forge-inspect): add ability to inspect libraries

0xferrous:
## Motivation

Its useful to be able to inspect which libraries a contract needs to dynamically link to.

## Solution

Uses the `link_references` mapping already available on the artifact.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/forge/src/cmd/inspect.rs
+++ b/crates/forge/src/cmd/inspect.rs
@@ -148,6 +148,27 @@ impl InspectArgs {
                 };
                 print_json(&standard_json)?;
             }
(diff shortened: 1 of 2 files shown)


0xferrous:
I'd like to also be able to inspect which libraries a contract internally links to. But I couldn't find a simple or any way to find that. Any insights regarding that is appreciated. Useful to determine if a contract needs to be redeployed after updating a specific internal library.

onbjerg (approved):

PR #11733, opened 2025-09-20
fix(common): use proper ESC in spinner erase-line sequence

VolodymyrBg:
The spinner used an invalid ANSI CSI introducer: \x33[2K (literal '3') instead of ESC \x1B[2K. As a result, terminals printed over shorter updates, leaving trailing characters from previous messages.
Correct sequence per ANSI/ECMA-48: ESC 0x1B (aka \x1B or \033), then [2K to erase the entire line.

before change: 
image

after: 
image


DaniPopes (approved):
Makes sense, however I don't know why it worked before

PR #11734, opened 2025-09-20
fix(script): prevent --verify without --broadcast from causing deployment not found error

### LINKED ISSUE #11726: bug(`forge script`): passing `--verify` without `--broadcast` results in: `Deployment not found for chain <chain_id>`
### Component

Forge

- [x] Foundry
- [x] Foundryup

tested v1.2.3 - forge 1.3.6-nightly (f7563cf448 2025-09-19T12:35:10.245500379Z)

foundryup: 1.3.0

forge script

### Operating System

Linux

### Describe the bug

This is apparently a longer standing edge case.

Minimal repro

```
$ forge init counter
$ forge script script/Counter.s.sol --rpc-url <RPC_URL> --verify --keystore <KEYSTORE>
```

where `--keystore` can be any wallet type

Results in

```
forge script script/Counter.s.sol --rpc-url <SEPOLIA_RPC_URL> --verify
[⠊] Compiling...
[⠆] Compiling 17 files with Solc 0.8.30
[⠰] Solc 0.8.30 finished in 248.23ms
Compiler run successful!
Error: Deployment not found for chain `11155111`.

Context:
- failed to read from "/home/zerosnacks/Projects/work/counter/broadcast/Counter.s.sol/11155111/dry-run/run-latest.json": No such file or directory (os error 2)
```

When passing `--broadcast` this works fine.

I think instead of erroring we should make `--verify` contingent on `--broadcast` being specified as `--verify` without broadcasting is, to my understanding, a meaningless statement.

DeVikingMark:
## Problem
`forge script --verify` without `--broadcast` fails with "Deployment not found for chain" error because it tries to load non-existent deployment files.

## Solution
- Remove problematic condition that attempts to resume when `--verify` is used without `--broadcast`
- Add validation to require `--broadcast` when using `--verify`
- Provide clear error message: "The --verify flag requires --broadcast to be specified"

## Changes
- `crates/script/src/lib.rs`: Fix logic and add validation
- `crates/forge/tests/cli/script.rs`: Add test case

Fixes #11726


crates/script/src/lib.rs
```diff
@@ -248,10 +248,16 @@ impl ScriptArgs {
         let create2_deployer = state.script_config.evm_opts.create2_deployer;
         let compiled = state.compile()?;
 
+        // Validate that --verify is only used with --broadcast
+        if compiled.args.verify && !compiled.args.should_broadcast() {
+            eyre::bail!(
+                "The --verify flag requires --broadcast to be specified. Verification without broadcasting is not meaningful."
+            );
+        }
```
onbjerg: you can do this with `requires` on `verify`: https://docs.rs/clap/latest/clap/struct.Arg.html#method.requires

onbjerg (changes requested):

onbjerg:
Ping @DeVikingMark, some of the tests are failing because the help output of `forge script` changed as a result of the changes in this PR :) Can you take a look so we can get this merged?

crates/forge/tests/cli/script.rs
```diff
@@ -3250,15 +3260,17 @@ Traces:
...
 ## Setting up 1 EVM.
+==========================
 Simulated On-chain Traces:
 
-  [..] → new Counter@0x5FbDB2315678afecb367f032d93F642f64180aa3
+  [96345] → new Counter@0x5FbDB2315678afecb367f032d93F642f64180aa3
     └─ ← [Return] 481 bytes of code
 
-  [..] Counter::increment()
+  [22418] Counter::increment()
```
onbjerg: please change this to dots again (the gas usage isn't relevant):)

onbjerg:
tests are still failing ptal

DeVikingMark:
@onbjerg please accept my apologies for the delayed update, but it seems right now everything runs like a well-oiled machine

PR #11744, opened 2025-09-21
Fix: Secure temporary file creation in chisel edit_session

zeevick10:
### Description

**Problem**: The `edit_session` function in `crates/chisel/src/dispatcher.rs` used a predictable temporary filename (`chisel-tmp.sol`) in the system temp directory, creating security risks:

- **Race conditions**: Multiple chisel sessions could overwrite each other's files
- **Symlink attacks**: Attackers could create symlinks to arbitrary paths (CWE-377)

**Solution**: 
- Replaced `std::env::temp_dir().join("chisel-tmp.sol")` with `tempfile::Builder`
- Added unique filename generation with `prefix("chisel-").suffix(".sol")`
- Maintained same user experience while eliminating security vulnerabilities

### DIFF
--- a/crates/chisel/Cargo.toml
+++ b/crates/chisel/Cargo.toml
@@ -26,7 +26,7 @@ foundry-compilers.workspace = true
 foundry-config.workspace = true
 foundry-evm.workspace = true
 
-tempfile = "3"
+tempfile.workspace = true
 
 solar.workspace = true
 


crates/chisel/Cargo.toml
```diff
@@ -26,6 +26,8 @@ foundry-compilers.workspace = true
 foundry-config.workspace = true
 foundry-evm.workspace = true
 
+tempfile = "3"
```
onbjerg: please use `.workspace = true`

onbjerg (changes requested):

onbjerg (approved):

PR #11750, opened 2025-09-22
fmt nits

grandizzy:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/fmt/src/state/yul.rs
```diff
@@ -196,19 +192,23 @@ impl<'ast> State<'_, 'ast> {
...
+        prefix: F,
+        block: &'ast yul::Block<'ast>,
         span: Span,
-        inline: bool,
         skip_opening_brace: bool,
-    ) {
+    ) where
+        F: FnOnce(&mut Self),
+    {
+        prefix(self);
```
DaniPopes: what are the reasons for these changes? it makes it harder to reason about when the closure is just executed first then the rest of the code is the same?

grandizzy: agree, is just playground to group and see if there's code that can be reused, not the final form

grandizzy: I gave up on that style, please check proposed changes :pray:

crates/fmt/testdata/ReprosCalls/80.fmt.sol
```diff
@@ -111,17 +118,3 @@ function returnLongBinaryOp() returns (bytes32) {
             | uint256(Nonce.unwrap(nonce))
     );
 }
-
-function new_y(
```
DaniPopes: why is this removed?

grandizzy: @0xrusowsky moved it in a different file in base branch but tests were not updated https://github.com/foundry-rs/foundry/pull/10907/files#diff-50e038a95258339134905679d8ee31a57e632c04c94386dd8362b0efd8e88db9R62

crates/doc/src/parser/item.rs
```diff
@@ -83,10 +83,8 @@ impl ParseItem {
 
     /// Set formatted code on the [ParseItem].
     pub fn with_code(mut self, source: &str, config: FormatterConfig) -> ParserResult<Self> {
-        // TODO(rusowsky): ensure that this is equivalent to the old fmt output
```
DaniPopes: this todo is valid, the previous code formatted individual items not the whole string

grandizzy: :+1: I add it back, will address in doc follow-up

crates/fmt/src/state/common.rs
```diff
@@ -630,45 +575,50 @@ impl<'ast> State<'_, 'ast> {
...
+            let should_break_after_disabled = is_disabled
                 && next_enabled
-                && let Some(next_lo) = next_lo
-                && self
-                    .peek_comment_before(next_lo)
-                    .is_none_or(|cmnt| self.inline_config.is_disabled(cmnt.span))
-            {
-                self.hardbreak_if_not_bol()
+                && next_lo
+                    .filter(|&lo| {
```
DaniPopes: is_some_and, but i feel like the previous is more readable. let chain > bunch of option methods

crates/fmt/src/state/sol.rs
```diff
@@ -282,9 +281,9 @@ impl<'ast> State<'_, 'ast> {
 
         if let Some(first) = bases.first().map(|base| base.span())
             && let Some(last) = bases.last().map(|base| base.span())
-            && self.inline_config.is_disabled(Span::new(first.lo(), last.hi()))
+            && self.inline_config.is_disabled(first.until(last))
```
DaniPopes: `to` not `until`, see doc comments on both

crates/fmt/src/state/sol.rs
```diff
@@ -2004,7 +2076,7 @@ impl<'ast> State<'_, 'ast> {
         // NOTE(rusowsky): unless we add bracket spans to solar,
```
DaniPopes: this comment is outdated, can this be fixed / what's missing? @0xrusowsky 
the `until` is correct here, but should be `to` as indicated by the comment

crates/fmt/src/state/sol.rs
```diff
@@ -2181,7 +2253,7 @@ impl<'ast> State<'_, 'ast> {
 
     fn is_inline_stmt(&self, stmt: &'ast ast::Stmt<'ast>, cond_len: usize) -> bool {
         if let ast::StmtKind::If(cond, then, els_opt) = &stmt.kind {
-            let if_span = Span::new(cond.span.lo(), then.span.hi());
+            let if_span = cond.span.until(then.span);
```
DaniPopes: same here, `to`

crates/fmt/src/state/common.rs
grandizzy: I reverted to prev in https://github.com/foundry-rs/foundry/pull/11750/commits/5cc4a15a7b3caaaf04d2cdf7731a17af2c709a17

crates/fmt/src/state/sol.rs
grandizzy: yep, https://github.com/foundry-rs/foundry/pull/11750/commits/5cc4a15a7b3caaaf04d2cdf7731a17af2c709a17

crates/fmt/src/state/sol.rs
grandizzy: changed in https://github.com/foundry-rs/foundry/pull/11750/commits/5cc4a15a7b3caaaf04d2cdf7731a17af2c709a17

crates/fmt/src/state/sol.rs
0xrusowsky: brain fart, sorry

crates/fmt/src/state/sol.rs
0xrusowsky: why is it outdated? @DaniPopes 

```rs
    /// An `if` statement with an optional `else` block: `if (expr) { ... } else { ... }`.
    If(Box<'ast, Expr<'ast>>, Box<'ast, Stmt<'ast>>, Option<Box<'ast, Stmt<'ast>>>),
```

with this info i cannot print the comments as in the original code for these 2 scenarios:
```solidity
    if /* comment10 */ (anotherLongCondition) // comment11
       /* comment12 */ {
            execute() ;
    } // comment13
```
```solidity
    if (
        step.opcode == 0x52
            && /*MSTORE*/ step.stack[0] == testContract.memPtr() // MSTORE offset
            && step.stack[1] == testContract.expectedValueInMemory() // MSTORE val
    ) {
        mstoreCalled = true;
    }
```

cause the span of the if expr in the 1st scenario is only the `anotherLongCondition` ident, and in the 2nd one the binary operator expr

so i either need to:
- `self.print_if_cond("if", cond, then.span.lo())`
- `self.print_if_cond("if", cond, cond.span.hi())`

the former is the one we do now, cause imo is more consistent with the old behavior (puts everything inside the condition's parenthesis)

the later leaves all comments out of the parenthesis, so the first scenario would work, but the other one would output something like:
```solidity
    if (
        step.opcode == 0x52
            && /*MSTORE*/ step.stack[0] == testContract.memPtr() // MSTORE offset
            && step.stack[1] == testContract.expectedValueInMemory()
    ) // MSTORE val
    {
        mstoreCalled = true;
    }
```

anyways, i think we can simply remove the comment. i only added it initially because i didn't know we were willing to not have 100% exact compatibility with the old formatter.
imo the if statement comment handling looks good, so the comment is not really necessary (unlike the `DeclMulti` comment, which could be a little bit more annoying for users, although they can simply use inline disable)

0xrusowsky: fyi, i deleted the cmnt.

the only implication is that this code:
```solidity
    if /* comment10 */ (anotherLongCondition) // comment11
       /* comment12 */ {
            execute() ;
    } // comment13
```

was previously formatted as:
```solidity
    if /* comment10 */ (
       anotherLongCondition // comment11
    ) {
       /* comment12 */
       execute();
    } // comment13
```
and now is:
```solidity
    if /* comment10 */ (
       anotherLongCondition // comment11
       /* comment12 */
    ) {
       execute();
    } // comment13
```

which is totally acceptable.
and if users want cmnt12 to go below, they simply need to put it inside the block, so that its parsed as part of the stmt

DaniPopes (approved):

crates/fmt/src/state/sol.rs
DaniPopes: that's fine

the comment was outdated wrt the inline code `.span` etc inside of the comment itself

PR #11757, opened 2025-09-22
fix: use sourcify for selector lookups

onbjerg:
## Motivation

Openchain has been transferred to Sourcify, and the original domain was not migrated.

## Solution

Use the new Sourcify API for selector lookup and import.

I've removed some duplicate tests, and some tests that no longer apply. Additionally, I've had to ignore some tests because they depend on selectors that are not present in Sourcify.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


DaniPopes:
> Error: Access is denied. (os error 5)
> Error: error decoding response body

etc.

onbjerg:
This is also blocked by the fact that there is no selector upload :+1:

kathleenmotley49-maker (approved):

crates/cast/src/args.rs
```diff
@@ -594,15 +593,13 @@ pub async fn run_command(args: CastArgs) -> Result<()> {
...
-            let signatures = stdin::unwrap_vec(signatures)?;
-            let ParsedSignatures { signatures, abis } = parse_signatures(signatures);
-            if !abis.is_empty() {
-                import_selectors(SelectorImportData::Abi(abis)).await?.describe();
-            }
-            if !signatures.is_empty() {
-                import_selectors(SelectorImportData::Raw(signatures)).await?.describe();
-            }
+        CastSubcommand::UploadSignature { .. } => {
+            sh_eprintln!(
```
DaniPopes: sh_warn

crates/cast/tests/cli/selectors.rs
```diff
@@ -1,19 +1,22 @@
...
 
 "#]]);
-});
+    }
+);
 
 casttest!(fourbyte, |_prj, cmd| {
     cmd.args(["4byte", "0xa9059cbb"]).assert_success().stdout_eq(str![[r#"
 transfer(address,uint256)
+_____$_$__$___$$$___$$___$__$$(address,uint256)
```
0xrusowsky, line 19: is this the other fn sig that produces a selector collision that u mentioned earlier?

should we add a small cmnt or is it clear enough already?

onbjerg: yes and idt a comment makes sense, we just have to live with the fact that people can find collissions and break our test suite 😢

0xrusowsky (approved):
makes sense to me!

DaniPopes (approved):

PR #11759, opened 2025-09-22
fix(linking): deep-merge link_references from creation and deployed bytecode

VolodymyrBg:
Before: Dependencies were built by shallow-merging link_references maps via extend. When both creation and deployed bytecode contained references under the same file key but with different library names, the latter overwrote the former, dropping dependencies.

After: We deep-merge the inner maps per file, unioning library names and offsets from both sources. This matches how we link and validate both bytecode objects and aligns with forge inspect ... libraries (union semantics). Prevents missing libs when a file has libraries used exclusively in constructor vs runtime.


DaniPopes:
Is there a test case we could add for this that would fail before?

@klkvr thoughts?

VolodymyrBg:
> Is there a test case we could add for this that would fail before?
> 
> @klkvr thoughts?

Yes, added a test case

crates/linking/src/lib.rs
```diff
@@ -105,14 +105,27 @@ impl<'a> Linker<'a> {
...
         let contract = self.contracts.get(target).ok_or(LinkerError::MissingTargetArtifact)?;
 
+        // Deep-merge link references from creation and deployed bytecode.
+        // This ensures we don't lose libraries when both bytecode objects reference
+        // libraries under the same source file key but with different library names.
         let mut references = BTreeMap::new();
         if let Some(bytecode) = &contract.bytecode {
-            references.extend(bytecode.link_references.clone());
+            for (file, libs) in &bytecode.link_references {
+                let entry = references.entry(file.clone()).or_insert_with(BTreeMap::new);
```
DaniPopes: or_insert_with(new) can be or_default

VolodymyrBg: > or_insert_with(new) can be or_default

corrected

crates/linking/src/lib.rs
```diff
@@ -717,6 +730,29 @@ mod tests {
         });
     }
 
+    #[test]
+    fn link_samefile_union() {
+        link_test("../../testdata/default/linking/samefile_union", |linker| {
```
klkvr: looks like you didn't commit the `linking/samefile_union` but the test is still passing somehow 🤔

crates/linking/src/lib.rs
```diff
@@ -105,14 +105,27 @@ impl<'a> Linker<'a> {
...
+        // libraries under the same source file key but with different library names.
+        let mut references: BTreeMap<String, BTreeMap<String, Vec<Offsets>>> = BTreeMap::new();
         if let Some(bytecode) = &contract.bytecode {
-            references.extend(bytecode.link_references.clone());
+            for (file, libs) in &bytecode.link_references {
+                let entry = references.entry(file.clone()).or_default();
+                for (name, offsets) in libs {
+                    entry.entry(name.clone()).or_default().extend(offsets.clone());
+                }
+            }
```
klkvr: I see, the issue is that we were not supposed to use `extend` here bc it would make second cycle's values overwrite the file keys

this does make sense

it seems that we accumulate `offsets` in this map for no reason, do you mind also getting rid of this and changing `references` to `BTreeMap<String, BTreeSet<String>>` so that it's a bit easier to reason about?

VolodymyrBg: > I see, the issue is that we were not supposed to use `extend` here bc it would make second cycle's values overwrite the file keys
> 
> this does make sense
> 
> it seems that we accumulate `offsets` in this map for no reason, do you mind also getting rid of this and changing `references` to `BTreeMap<String, BTreeSet<String>>` so that it's a bit easier to reason about?

done, also added a test if artifacts are missing and committed samefile_union

VolodymyrBg:
> needs rebase and ci fixes

[test / nextest / test unit (x86_64-pc-windows-msvc) (pull_request)](https://github.com/foundry-rs/foundry/actions/runs/17981493126/job/51148499811?pr=11759)
test / nextest / test unit (x86_64-pc-windows-msvc) (pull_request)Failing after 6m
[test / nextest / test unit (x86_64-unknown-linux-gnu) (pull_request)](https://github.com/foundry-rs/foundry/actions/runs/17981493126/job/51148499762?pr=11759)
test / nextest / test unit (x86_64-unknown-linux-gnu) (pull_request)Failing after 2m

I'm sorry but i can't fix this, i can't just test it on my computer, it's not that powerful, and I think making noise here is not the best idea..

crates/linking/src/lib.rs
```diff
@@ -557,94 +522,101 @@ mod tests {
...
+                    vec![
                         (
-                            "default/linking/duplicate/Duplicate.t.sol:A",
+                            "default/linking/duplicate/Duplicate.t.sol:A".to_string(),
                             Address::from_str("0x5a443704dd4b594b382c22a083e2bd3090a6fef3")
                                 .unwrap(),
                         ),
                         (
-                            "default/linking/duplicate/Duplicate.t.sol:C",
+                            "default/linking/duplicate/Duplicate.t.sol:C".to_string(),
```
DaniPopes: please undo these changes, you accepted the wrong side when merging master

VolodymyrBg:
> Thanks! Updated tests. I verified that the added test fails on master and passes in this PR.

Wow, thank you, too. I didn't really know how to revert my incorrect rebase so i was stuck, thank you for your help

Issue #8898, opened 2024-09-18
`forge test` doesn't utilize all available threads for fuzzing/invariants


Philogy:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### What version of Foundry are you on?

forge 0.2.0 (03ea54c 2024-09-18T16:31:08.207328652Z)

### What command(s) is the bug in?

forge test

### Operating System

None

### Describe the bug

Forge testing is multi-threaded but only on a per test level, for instance if I have 15 different fuzz runs and 224 available threads foundry will assign 15/224 threads with one test each, however some tests take much longer to run than others. Furthermore individual runs of the same test are completely independent. I'd expect different runs to get scheduled to different threads but foundry doesn't seem to do that, leaving a lot of performance on the table.

grandizzy:
good point, we're currently limited by proptest test runner but should consider changing this post 1.0 release

Philogy:
> good point, we're currently limited by proptest test runner but should consider changing this post 1.0 release

This would be an amazing feature to have, would mean a 8-224x speedup depending on the machine I run the tests on

Philogy:
Could there be a temporary workaround where foundry duplicates a given test or invariant for you by the number of available threads and then does `total_runs / N` runs on each?

zerosnacks:
You are correct, we use `T-breaking` as an internal tag to indicate it would likely require making a _breaking change_  and we should decide whether to include it before or after `1.0`. Moving away from `proptest` would be a huge change but indeed should not be experienced as breaking by the end-user.

Philogy:
I see, good to know thx

0xalpharush:
I believe implementing this is unblocked by https://github.com/foundry-rs/foundry/pull/10190

I anticipate that there is a diminishing return for adding threads and, at some point, it may be better to run multiple processes or run on multiple machines and sync the corpuses intermittently...

figtracer:
retested on current master: a 32-run invariant campaign passes with `--invariant-workers 2`. closing as completed.

PR #11769, opened 2025-09-23
feat(fuzz): `WorkerCorpus` for multiple worker threads

yash-atreya:
## Motivation

towards #8898 (#11842)

## Solution

-  Replaces `CorpusManager` with `WorkerCorpus` 
- `WorkerCorpus` is the corpus to be used by parallel worker threads.
- Each `WorkerCorpus` has an `id: u32`. The master worker has the `id = 0`.
- The `WorkerCorpus`'s share their `in_memory_corpus` amongst each other via the file system using a star-pattern where the master worker (`id = 0`) is in the center.
- The `corpus_dir` is now organized as:
```
corpus_dir/
    worker0/ - // master
        sync/
        corpus/
    worker1/
        sync/
        corpus/
    worker2/
        sync/
```
- Each non-master worker exports their corpus to `worker0/sync/`  - See `fn export` in [089f30b](https://github.com/foundry-rs/foundry/pull/11769/commits/089f30b52a2c5edb56e9928e183616e742dab272)
- The master worker distributes it's `worker0/corpus` entries (which includes entries from all workers when synced) to each workers `sync/` directory - See `fn distribute` in [d4200e4](https://github.com/foundry-rs/foundry/pull/11769/commits/d4200e4b4fe26d284c8be48cf776cc00a610485f)
- Each worker then pulls the new corpus entries from their respective `corpur_dir/workerId/sync` dir into `corpus_dir/workerId/corpus` if it leads to new coverage and updates it's `history_map` - See `fn calibrate` in [488d09d](https://github.com/foundry-rs/foundry/pull/11769/commits/488d09dd4b4ab59aee2cbc151fc660d000d0bf7c)
- In `fn calibrate` we are fetching the new corpus entries from the workers `sync/` dir and *replaying* the tx sequences to check if they lead to new coverage for this particular worker. If it does then we're updating `history_map`.
- The `pub fn sync` introduced in [e9d8d3c](https://github.com/foundry-rs/foundry/pull/11769/commits/e9d8d3ce06f85fb3be43ac773084f5a388eeda26) handles all of the above. 

Note: This PR does not address parallelizing the fuzz runs, only prepares for it. Opened for initial feedback on the approach.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/evm/src/executors/corpus.rs
+++ b/crates/evm/evm/src/executors/corpus.rs
@@ -218,7 +218,8 @@ impl WorkerCorpus {
         let mut failed_replays = 0;
 
         if id == 0 && config.corpus_dir.is_some() {
-            // Master worker loads the initial corpus if it exists
+            // Master worker loads the initial corpus, if it exists. Then, [distribute]s it to
+            // workers.
             let corpus_dir = config.corpus_dir.as_ref().unwrap();
 
             let can_replay_tx = |tx: &BasicTxDetails| -> bool {
@@ -685,7 +686,7 @@ impl WorkerCorpus {
 
     // Sync Methods
 
-    /// Exports the new corpus entries to the master workers (id = 0) sync dir.
+    /// Exports the new corpus entries to the master worker's (id = 0) sync dir.
     #[tracing::instrument(skip_all, fields(worker_id = self.id))]
     fn export(&self) -> eyre::Result<()> {
         // Early return if no new entries or corpus dir not configured
@@ -723,6 +724,7 @@ impl WorkerCorpus {
                 let file_path = corpus_dir.join(&file_name);
                 let sync_path = master_sync_dir.join(&file_name);
 
+                // TODO(dani): can we symlink and copy later instead?
                 let Ok(_) = foundry_common::fs::copy(file_path, sync_path) else {
                     debug!(target: "corpus", "failed to export corpus {}", entry.uuid);
                     continue;
@@ -941,27 +943,16 @@ impl WorkerCorpus {
         fuzzed_function: Option<&Function>,
         fuzzed_contracts: Option<&FuzzRunIdentifiedContracts>,
     ) -> eyre::Result<()> {
-        if self.id == 0 {
-            // Master worker
-            self.calibrate(executor, fuzzed_function, fuzzed_contracts)?;
-
-            self.distribute(num_workers)?;
-
-            self.new_entry_indices.clear();
-
-            trace!(target: "corpus", "master synced");
-
-            return Ok(());
+        let is_master = self.id == 0;
+        if !is_master {
+            self.export()?;
         }
-
-        self.export()?;
-
         self.calibrate(executor, fuzzed_function, fuzzed_contracts)?;
-
+        if is_master {
+            self.distribute(num_workers)?;
+        }
         self.new_entry_indices.clear();
-
         trace!(target: "corpus", "synced");
-
         Ok(())
     }
 }


crates/evm/evm/src/executors/corpus.rs
```diff
@@ -47,19 +47,19 @@ enum MutationType {
...
 /// Holds Corpus information.
 #[derive(Serialize)]
-struct CorpusEntry {
+pub(crate) struct CorpusEntry {
     // Unique corpus identifier.
-    uuid: Uuid,
+    pub(crate) uuid: Uuid,
     // Total mutations of corpus as primary source.
-    total_mutations: usize,
+    pub(crate) total_mutations: usize,
```
yash-atreya: The `pub(crate)` accessibility is a temp measure, will be removed once I remove the existing `CorpusManager` which is currently being used `InvariantExecutor`

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -128,35 +133,40 @@ impl CorpusMetrics {
...
+    /// workers. Mutation is performed on these.
+    ///
+    /// Map of corpus [`Uuid`] to [`CorpusEntry`].
+    in_memory_corpus: Arc<RwLock<Vec<CorpusEntry>>>,
+    /// Number of failed replays from persisted corpus.
+    failed_replays: Arc<AtomicUsize>,
+    /// History of binned hitcount of edges seen during fuzzing
+    history_map: Arc<RwLock<Vec<u8>>>,
+    /// Corpus metrics.
+    pub(crate) metrics: Arc<RwLock<CorpusMetrics>>,
```
0xalpharush: I would suggest making these per-worker (avoiding locks) and designating one worker the master. Then at some interval, each worker would copy its corpus "changeset" to the master i.e. what's new since the last sync. The master node then would merge new inputs from disk with its own corpus and keep all entries that contribute unique coverage in-memory.

```
# corpus dir
master/
worker1/
worker2/
```
sync creates
```
master/worker1_new/
master/worker2_new/
```

AFL ++ stores a bit map of the history_map (1-byte hitcount -> bool indicating hit)  for each input and then uses this bit map to make sure it doesn't eject an entry that is unique (this would mean [updating](https://github.com/foundry-rs/foundry/blob/5dcf01b7d32b059a2bfa3855e17af03b5cadf103/crates/evm/evm/src/executors/corpus.rs#L302-L303) `is_favored` to mean it contributes a unique bit to the bitmap).  However, I'm not exactly sure how slave corpus reloads should work... I think it can just copy from the master and perform the same calibration stage.

Regarding the sync interval: for invariant tests that are run for 256 runs during CI, I don't think it makes sense to parallelize or sync. For long-running campaigns, if there are more invariant contracts than threads available, you would have to run them round robin, and it'd make sense to sync when the task stops and minimize the corpus when it restarts. If there are less invariant contracts than threads, then you can just cleanly sync at a given interval e.g. ten minutes (or maybe a fraction of the total run-time if it's less)

I am not attached to the use of directories for sync like AFL++ and it may be possible to use some concurrency primitive like bi-directional channels in Rust. Having a mutex on the history map is not scalable, however. Feedback to my feedback welcome :)

yash-atreya: Thanks! This is valuable feedback, and a better approach over workers frequently contending for locks. looking into this.

yash-atreya: Double-clicking on the syncing bit:

You are suggesting syncing the workers' findings via the file system by writing to `corpus_dir/master`? On periodic sync intervals, all workers update their `in_memory_corpus` and `history_map` by looking up the data provided by other workers. 

This removes the locks but each worker will have some more I/O on sync, which is fine I think

0xalpharush: It is a star pattern of communication and both the master and worker would updated their `in_memory_corpus` and `history_map` after calibrating the new finds. To avoid re-calibrating, the file name can be a content hash, or timestamp/index. Then, the worker would just check if it's seen it or if the index/timestamp has increased since it last. This would require recording `last_seen_idx` from master and for every-worker in the case of the master.

Here is an example:
```
master/
    sync/
    corpus/A<content_hash | timestamp>
worker1/
     sync
    corpus/B<content_hash | timestamp>
worker2/
     sync
    corpus/C<content_hash | timestamp>
```
At a sync interval, worker1 and worker2 copy to master. Master copies to worker1 and worker2

```
master/
    sync/
         B<content_hash | timestamp>
         C<content_hash | timestamp>
    corpus/A<content_hash | timestamp>
worker1/
     sync/
           A<content_hash | timestamp>
    corpus/B<content_hash | timestamp>
worker2/
     sync/
           A<content_hash | timestamp>
    corpus/C<content_hash | timestamp>
```

If B is "new find" for master but not B, it would be added to the worker2 at the next sync interval but C would not be given to worker 1
```
worker2/
     sync/
           B<content_hash | timestamp>
    corpus/
        C<content_hash | timestamp>
```
After calibration it would import the new find it has no entry for 
```
worker2/
     sync/

    corpus/
         C<content_hash | timestamp>
         B<content_hash | timestamp>_imported
```

It takes a couple sync periods to disseminate, but avoids too much duplicate work calibrating whether an entry contributes a unique hit as the master node effectively has the "best" corpus after its sync'd and calibrated

yash-atreya: Yeah makes sense, timestamps make it a lot better. Thanks

kathleenmotley49-maker (approved):

crates/evm/evm/src/executors/corpus.rs
yash-atreya: @grandizzy I have applied @0xalpharush's suggestion to sync via fs. 

Please note that we're replaying the new corpus entries when syncing a worker to figure out if it leads to new coverage for it. If it does, we're updating the `history_map`. See [`fn calibrate`](https://github.com/foundry-rs/foundry/blob/5ddc908c9df6fed8b7fe7320c868cc5e3c8d0f21/crates/evm/evm/src/executors/corpus.rs#L791)

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -321,36 +359,55 @@ impl CorpusManager {
...
-                "persisted {} inputs for new coverage in {corpus_uuid} corpus",
-                &corpus.tx_seq.len()
+                "persisted {} inputs for new coverage in worker {} for {corpus_uuid} corpus",
+                self.id, &corpus.tx_seq.len()
             );
         }
 
+        // Track in-memory corpus changes to update MasterWorker on sync
+        let new_index = self.in_memory_corpus.len();
+        self.new_entry_indices.push(new_index);
```
0xalpharush, line 384: I think this is fine, but may result in some corpus entries not getting sync'd e.g. due to a crash or ctr+c and restart. If you persist the last sync'd timestamp, it can recover from restarts by checking if there are newer entries written before a sync could occur

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -166,127 +192,139 @@ impl CorpusManager {
...
+        let mut metrics = CorpusMetrics::default();
+        let mut failed_replays = 0;
 
-                        let (new_coverage, is_edge) =
-                            call_result.merge_edge_coverage(&mut history_map);
-                        if new_coverage {
-                            metrics.update_seen(is_edge);
-                        }
+        if id == 0 && config.corpus_dir.is_some() {
+            // Master worker loads the initial corpus if it exists
```
0xalpharush:
```suggestion
            // Master worker loads the initial corpus, if it exists. Then, [export]s to workers.
```

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -321,36 +359,55 @@ impl CorpusManager {
...
         let corpus = CorpusEntry::from_tx_seq(inputs);
         let corpus_uuid = corpus.uuid;
-
+        let timestamp = corpus.timestamp;
         // Persist to disk if corpus dir is configured.
         let write_result = if self.config.corpus_gzip {
             foundry_common::fs::write_json_gzip_file(
-                corpus_dir.join(format!("{corpus_uuid}{JSON_EXTENSION}.gz")).as_path(),
+                worker_corpus
+                    .join(format!("{corpus_uuid}-{timestamp}{JSON_EXTENSION}.gz"))
```
0xalpharush, line 361: maybe the uuid is redundant if we prefix the worker id to the timestamp to avoid collision

crates/evm/evm/src/executors/corpus.rs
0xalpharush: Actually I guess this is `distribute`

DaniPopes:
please merge master, this is still using old CI runners

kathleenmotley49-maker (approved):

crates/evm/evm/src/executors/corpus.rs
yash-atreya: Timestamps are in seconds, corpus-uuid avoids collision within the same worker

yash-atreya: Timestamps are in seconds, corpus-uuid avoids collision within the same worker

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -638,6 +687,303 @@ impl CorpusManager {
...
+                let ext = self
+                    .config
+                    .corpus_gzip
+                    .then_some(format!("{JSON_EXTENSION}.gz"))
+                    .unwrap_or(JSON_EXTENSION.to_string());
+                let file_name = format!("{}-{}{ext}", entry.uuid, entry.timestamp);
+                let file_path = corpus_dir.join(&file_name);
+                let sync_path = master_sync_dir.join(&file_name);
+
+                let Ok(_) = foundry_common::fs::copy(file_path, sync_path) else {
```
0xalpharush, line 728: I wonder if we could symlink and only copy if it is going to be deleted or moved later.

kathleenmotley49-maker (approved):

crates/evm/evm/src/executors/corpus.rs
0xalpharush: Or once it's decided to import from the sync dir, we create the hardlink

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -321,36 +359,55 @@ impl CorpusManager {
...
         if let Err(err) = write_result {
-            debug!(target: "corpus", %err, "Failed to record call sequence {:?}", &corpus.tx_seq);
+            debug!(target: "corpus", %err, "Failed to record call sequence {:?} in worker {}", &corpus.tx_seq, self.id);
         } else {
             trace!(
                 target: "corpus",
-                "persisted {} inputs for new coverage in {corpus_uuid} corpus",
-                &corpus.tx_seq.len()
+                "persisted {} inputs for new coverage in worker {} for {corpus_uuid} corpus",
+                self.id, &corpus.tx_seq.len()
```
DaniPopes: we can avoid logging worked id by using a span on the function, also the argument order is incorrect on this specific log

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -575,13 +615,12 @@ impl CorpusManager {
             let corpus = self.in_memory_corpus.get(index).unwrap();
 
             let uuid = corpus.uuid;
-            debug!(target: "corpus", "evict corpus {uuid}");
+            debug!(target: "corpus", "evict corpus {uuid} in worker {}", self.id);
```
DaniPopes: same

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -166,127 +192,139 @@ impl CorpusManager {
...
 
-                        // Commit only when running invariant / stateful tests.
-                        if fuzzed_contracts.is_some() {
-                            executor.commit(&mut call_result);
-                        }
-                    } else {
-                        failed_replays += 1;
+            let executor = executor.expect("Executor required for master worker");
+            'corpus_replay: for entry in std::fs::read_dir(corpus_dir)? {
+                let path = entry?.path();
```
onbjerg, line 237: Do we continually write to the corpus directory? This is very expensive as we not only iterate a directory and read the files, but we also (if gzip is enabled) do decompression over and over, potentially of the same file. It feels like the corpus should be default in-memory, and we only write at the end.

0xalpharush: This is just happening at start up. The entries are held in memory so long as they don't exceed a configurable limit and then flushed to disk (and compressed if it's enabled). 

Your point does still stand elsewhere. IIUC workers share compressed corpus entries so they potentially are repeatedly decompressing the same files. Moving compression to the very end would resolve this

kathleenmotley49-maker (approved):

kathleenmotley49-maker (approved):

kathleenmotley49-maker (approved):

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -630,6 +682,299 @@ impl CorpusManager {
...
+
+        let sync_dir = worker_dir.join(SYNC_DIR);
+        let corpus_dir = worker_dir.join(CORPUS_DIR);
+
+        let mut executor = executor.clone();
+        for tx_seq in self.import()? {
+            if !tx_seq.is_empty() {
+                let mut new_coverage_on_sync = false;
+                for tx in &tx_seq {
+                    if can_replay_tx(tx) {
```
0xalpharush, line 818: Checking the entire sequence prior to executing may make sense as I think currently `new_coverage_on_sync` can be set and a later tx contains an invalid selector. Ideally, we could just drop those txs and see if we can recover and retain the modified, valid corpus entry

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -630,6 +682,299 @@ impl CorpusManager {
...
+            return Ok(());
+        }
+
+        self.export()?;
+
+        self.calibrate(executor, fuzzed_function, fuzzed_contracts)?;
+
+        self.new_entry_indices.clear();
+
+        trace!(target: "corpus", "synced");
```
0xalpharush, line 955: Maybe add worker id

0xalpharush:
Overall looks pretty good to me. I left a few comments and there are some outstanding. I would recommend adding a config to turn syncing off and also a way to make it occur more/less frequently.

crates/evm/evm/src/executors/corpus.rs
DaniPopes: part of the function span

DaniPopes:
Merged into #12713.

PR #11771, opened 2025-09-23
Harden foundryup temp handling: secure mktemp and quote tar paths

viktorking7:
- **What**:  
  - Replace insecure `mktemp -d` fallback (`|| echo "."`) with explicit error handling.  
  - Quote temporary archive paths in `tar` invocations.

- **Why**:  
  - Prevents writing predictable temp files in the current directory if `mktemp` fails.  
  - Avoids path parsing issues when filenames contain spaces or glob characters.

- **Impact**:  
  - Safer installation flow; no behavior change under normal conditions.

### DIFF
--- a/foundryup/foundryup
+++ b/foundryup/foundryup
@@ -165,7 +165,7 @@ main() {
       say "checking if forge, cast, anvil, and chisel for $FOUNDRYUP_TAG version are already installed"
 
       # Create a temporary directory to store the attestation link and artifact.
-      tmp_dir="$(mktemp -d 2>/dev/null || echo ".")"
+      tmp_dir="$(mktemp -d 2>/dev/null)" || err "failed to create temp dir"
       tmp="$tmp_dir/attestation.txt"
       ensure download "$ATTESTATION_URL" "$tmp"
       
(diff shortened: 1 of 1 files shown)


viktorking7:
> what is wrong with falling back to cwd?


Falling back to the current working directory makes the temp file path predictable and potentially attacker-controlled.

- Predictable name in a writable CWD enables symlink/hardlink clobbering of arbitrary files (CWE‑377/CWE‑59), e.g. `./foundry.tar.gz` or `./attestation.txt` can point elsewhere.
- Users often run installers from repo dirs or shared/workspace dirs; CWD isn’t a trustworthy temp location.
- The safe behavior is to fail fast if `mktemp` fails, rather than writing into CWD.

If a fallback is desired, it should use a trusted temp base, not CWD, e.g.:

```bash
tmp_dir="$(mktemp -d 2>/dev/null)" || {
  base="${TMPDIR:-/tmp}"
  tmp_dir="$(mktemp -d "${base%/}/foundryup.XXXXXX")" || err "failed to create temp dir"
}
```

This avoids predictable paths in untrusted directories and keeps the change minimal.

onbjerg:
I'm not sure I really understand the attack vector here. We overwrite the target file, so if it's a symlink it will either fail or be overwritten

viktorking7:
> I'm not sure I really understand the attack vector here. We overwrite the target file, so if it's a symlink it will either fail or be overwritten

The issue isn’t about tar/unzip reading — it’s about writing predictable temp files into CWD.

- With the fallback, the temp path becomes predictable and attacker-controllable if the user runs the script in an untrusted/shared directory.
- A pre‑placed symlink in CWD turns our write into arbitrary file overwrite. `curl -o "$tmp"` will happily follow a symlink and truncate the target (CWE‑377 / CWE‑59), which is not a failure but clobbering.
- Example: `ln -s ~/.zshrc ./foundry.tar.gz` → running the installer will overwrite `~/.zshrc` before any checks.
- CWD is not a trusted temp location; failing fast (or falling back to `$TMPDIR`/`/tmp` with a randomized subdir) is the standard safe pattern.

The change simply removes the unsafe CWD fallback. Under normal conditions (when `mktemp` works) behavior is unchanged.

PR #11773, opened 2025-09-23
Ensure bin dir exists in foundryup; guard npm local registry operations

kilavvy:
### Description
- Added `mkdir -p` for `FOUNDRY_BIN_DIR` in `foundryup`:
  - Before creating symlinks (local install)
  - Before copying binaries in `use()`
- Added a safety guard to `npm/scripts/setup-local.sh`:
  - `require_local_registry` blocks publish/unpublish to non-local registries
  - Allow override via `NPM_ALLOW_NONLOCAL=1`

#### Why
- Prevents first-time install failures when `~/.foundry/bin` doesn’t exist
- Avoids accidental publish/unpublish to public registries

#### Impact
- No breaking changes; safer defaults for local workflows

### DIFF
--- a/crates/anvil/rpc/src/response.rs
+++ b/crates/anvil/rpc/src/response.rs
@@ -47,7 +47,11 @@ impl ResponseResult {
     where
         S: Serialize + 'static,
     {
-        Self::Success(serde_json::to_value(&content).unwrap())
+        // Avoid panicking if serialization fails; return an internal error instead.
+        match serde_json::to_value(&content) {
+            Ok(value) => Self::Success(value),
+            Err(_) => Self::Error(RpcError::internal_error()),
+        }
     }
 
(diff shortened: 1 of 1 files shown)


crates/anvil/rpc/src/response.rs
```diff
@@ -47,7 +47,11 @@ impl ResponseResult {
     where
         S: Serialize + 'static,
     {
-        Self::Success(serde_json::to_value(&content).unwrap())
+        // Avoid panicking if serialization fails; return an internal error instead.
+        match serde_json::to_value(&content) {
+            Ok(value) => Self::Success(value),
+            Err(_) => Self::Error(RpcError::internal_error()),
+        }
```
onbjerg, line 54: this is unrelated

DaniPopes:
Thanks but we don't accept contributions to foundryup anymore

PR #11781, opened 2025-09-24
feat: add trace transaction opts

### LINKED ISSUE #9391: Respect `TransactionOpts` in `cast call --trace`
### Component

Cast

Currently `cast call` has "Transaction options", which specifies options like gas limit, gas price, etc. However they are not used during simulation with `--trace`. This is a discussion whether they should also be used.

### Additional context

_No response_

kien6034:
Closes #9391 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/cast/src/cmd/call.rs
+++ b/crates/cast/src/cmd/call.rs
@@ -322,7 +322,6 @@ impl CallArgs {
                 env_tx.nonce = nonce;
             }
 
-            // Note: value is already set via the `value` variable extracted from tx.value above
 
             if let Some(tx_type) = tx.inner.transaction_type {
                 env_tx.tx_type = tx_type;
--- a/crates/cast/src/cmd/call.rs
+++ b/crates/cast/src/cmd/call.rs
@@ -297,6 +297,33 @@ impl CallArgs {
             let tx_kind = tx.inner.to.expect("set by builder");
             let env_tx = &mut executor.env_mut().tx;
 
+            // Set transaction options with --trace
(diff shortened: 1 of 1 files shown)


crates/cast/src/cmd/call.rs
```diff
@@ -297,6 +297,33 @@ impl CallArgs {
...
+
+            if let Some(max_fee_per_blob_gas) = tx.inner.max_fee_per_blob_gas {
+                env_tx.max_fee_per_blob_gas = max_fee_per_blob_gas;
+            }
+
+            if let Some(nonce) = tx.inner.nonce {
+                env_tx.nonce = nonce;
+            }
+
+            // Note: value is already set via the `value` variable extracted from tx.value above
```
onbjerg:
```suggestion
```

smol nit, this comment is unnecessary:)

crates/cast/src/cmd/call.rs
```diff
@@ -297,6 +297,32 @@ impl CallArgs {
...
+            }
+
+            if let Some(max_fee_per_blob_gas) = tx.inner.max_fee_per_blob_gas {
+                env_tx.max_fee_per_blob_gas = max_fee_per_blob_gas;
+            }
+
+            if let Some(nonce) = tx.inner.nonce {
+                env_tx.nonce = nonce;
+            }
+
```
onbjerg, line 324:
```suggestion
```

onbjerg (changes requested):

onbjerg (approved):
thanks!

PR #11785, opened 2025-09-24
fix: osaka hardfork tx gas limit

onbjerg:
## Motivation

The EIP-7825 check was incorrect

## Solution

Enforce the gas limit from EIP-7825

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


onbjerg:
This is incorrect, even though the method and variable have the exact same name, the semantics are different. To check if you need to enforce the gas limit, you do `gas_limit_cap.is_none()` and then you get the cap by `gas_limit_cap()` -.-

PR #11786, opened 2025-09-24
Use POSIX-compliant shell test in Dockerfile to fix Alpine build

viktorking7:
### PR Description 
- **Summary**: Replace bash-specific `[[ ... ]]` with POSIX `[ ... ]` in the `Dockerfile` to ensure compatibility with Alpine’s `/bin/sh` (BusyBox `ash`).

- **Change**:
  - `Dockerfile`: `[[ "$TARGETARCH" = "arm64" ]]` → `[ "$TARGETARCH" = "arm64" ]`

- **Why**:
  - Alpine’s default shell doesn’t support `[[ ... ]]`, which can cause build failures

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN apk add clang lld curl build-base linux-headers git \
     && chmod +x ./rustup.sh \
     && ./rustup.sh -y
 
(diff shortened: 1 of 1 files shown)


onbjerg:
This works fine, the Docker image is published every day

PR #11801, opened 2025-09-24
fix(coverage): sync implementation with forge test

DaniPopes:
Coverage was calling a low level function of forge test args, which does not setup stuff like config.fuzz.gas_report_samples, EvmArgs, etc., as well as missing options in the test runner builder etc.

This means that `forge coverage` falls out of sync gradually with `forge test`. One of the options which was thus missed, `gas_report_samples`, controls how many traces are saved from fuzz and invariant tests, which with the recent addition of tracing for backtraces means we are storing a lot of traces unnecessarily, massively increasing memory usage.

### DIFF
--- a/crates/forge/src/cmd/test/mod.rs
+++ b/crates/forge/src/cmd/test/mod.rs
@@ -39,12 +39,15 @@ use foundry_config::{
     filter::GlobMatcher,
 };
 use foundry_debugger::Debugger;
-use foundry_evm::traces::{backtrace::BacktraceBuilder, identifier::TraceIdentifiers};
+use foundry_evm::{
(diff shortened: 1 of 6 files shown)


crates/forge/src/cmd/test/mod.rs
```diff
@@ -281,9 +275,30 @@ impl TestArgs {
             .files(self.get_sources_to_compile(&config, &filter)?);
         let output = compiler.compile(&project)?;
```
DaniPopes, line 276: this doesn't sync the compilation, which is currently also mismatched due to `get_sources_to_compile` and `dynamic_test_linking`

PR #11812, opened 2025-09-25
fix(preprocessor): load all sources into SourceMap; parse only tests/scripts

radik878:
Previously, the preprocessor loaded only tests/scripts into the SourceMap, because new_source_file was gated by is_test_or_script(path). This contradicted the comment (“include all sources in the source map, parse only tests/scripts”) and could cause unnecessary disk reads for non-test imports during resolution.
This change:
- Always calls new_source_file for all sources to preload them into the SourceMap.
- Adds files to the parsing context (pcx.add_file) and preprocessed_paths only for tests/scripts.
- Refactors a nested if into a let-else + continue to satisfy Clippy without changing semantics.

Result: behavior now matches the documented intent, avoids extra disk IO, and aligns with patterns used elsewhere in the codebase.


DaniPopes:
doesn't matter and this change does nothing

PR #11818, opened 2025-09-25
Prevent secret leakage by not logging first positional CLI argument

maximevtush:
Description
- What: Remove inclusion of the first positional argument from error messages in CommandUtils::exec.
- Why: Positional args may contain secrets (e.g., keys, tokens, mnemonics, credentials in URLs). Logging them risks secret exposure.
- Scope: crates/cli/src/utils/mod.rs only; no functional behavior change.

Security
- Reduces risk of “secrets-in-logs” leakage in error paths.
- No change to stdout/stderr content; only the synthesized error header is safer.


onbjerg (changes requested):
this makes it really hard to debug and this is primarily used for ffi tool usage

Issue #11821, opened 2025-09-25
bug(fmt): subsequent runs format same code with different style


grandizzy:
### Component

Forge

### Have you ensured that all of these are up to date?

- [ ] Foundry
- [ ] Foundryup

### Operating System

None

### Describe the bug

- first fmt run
```diff
-            uint256 globalBuyAmount = Take.take(
-                state, notes, uint32(IPoolManager.take.selector), recipient, minBuyAmount
-            );
+            uint256 globalBuyAmount =
+                Take.take(state, notes, uint32(IPoolManager.take.selector), recipient, minBuyAmount);
```
- second run
```diff
-            uint256 globalBuyAmount =
-                Take.take(state, notes, uint32(IPoolManager.take.selector), recipient, minBuyAmount);
+            uint256 globalBuyAmount = Take.take(
+                state, notes, uint32(IPoolManager.take.selector), recipient, minBuyAmount
+            );

```

PR #11824, opened 2025-09-25
fix(fmt): estimate size + account for all blocks

0xrusowsky:
## Motivation

closes: 
- [x] https://github.com/foundry-rs/foundry/issues/11821

## Solution

this PR introduces 2 fixes:
- when estimating the size of a snippet, only account for line breaks that are preceeded by a line that ends with a semicolon.
- track block depth so that, when estimating the space left, we can account for all of them rather than just contracts and funcs

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/fmt/src/state/mod.rs
+++ b/crates/fmt/src/state/mod.rs
@@ -832,6 +832,13 @@ impl<'sess> State<'sess, '_> {
         self.comments.iter().take_while(|c| c.pos() < pos).find(|c| !c.style.is_blank())
     }
 
+    fn has_comment_before_with<F>(&self, pos: BytePos, f: F) -> bool
+    where
+        F: FnMut(&Comment) -> bool,
+    {
+        self.comments.iter().take_while(|c| c.pos() < pos).any(f)
+    }
+
     fn peek_comment_between<'b>(&'b self, pos_lo: BytePos, pos_hi: BytePos) -> Option<&'b Comment>
     where
         'sess: 'b,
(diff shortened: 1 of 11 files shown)


onbjerg:
does this fix 11821?

crates/fmt/src/state/mod.rs
```diff
@@ -370,16 +369,20 @@ impl State<'_, '_> {
...
+                if prev_needs_space {
+                    // Previous line ended with a bracket and config with bracket spacing.
+                    // Previous line ended with ',' a hardbreak or a space are required.
+                    // Previous line ended with ';' a hardbreak is required.
                     size += 1;
                 }
+                size += line.trim().len();
+
+                prev_needs_space = self.config.bracket_spacing
+                    && (line.ends_with('(') || line.ends_with('{'))
```
grandizzy: is that right, cannot end in "{\n" or extra space after for example?

0xrusowsky: i missed a parenthesis, i wanted to do:
`(self.config.bracket_spacing && (line.ends_with('(') || line.ends_with('{')))`

meaning that if the config tells us that we must print a space after a bracket, we will always add either a space or a line brake, so we know size will be +1

onbjerg:
if this closes 11821 please mark it as such

0xrusowsky:
confirmed that the fix works fine on 0x-settler and ithaca/account

pending @grandizzy final review after testing on other repos like euler or uni

PR #11826, opened 2025-09-25
fix: normalize path separators in git submodule URL lookup

wellnana:
Git always uses forward slashes in submodule config keys, even on Windows, so we need to normalize the path separators before looking up the URL.

This also fixed flaky test. 
Might be the reason why the test is failing during https://github.com/foundry-rs/foundry/pull/11437#pullrequestreview-3230949833 cc @onbjerg

### DIFF
--- a/crates/cli/src/utils/mod.rs
+++ b/crates/cli/src/utils/mod.rs
@@ -8,6 +8,7 @@ use foundry_common::{
 };
 use foundry_config::{Chain, Config};
 use itertools::Itertools;
+use path_slash::PathExt;
 use regex::Regex;
 use serde::de::DeserializeOwned;
 use std::{
@@ -718,13 +719,9 @@ ignore them in the `.gitignore` file."
     }
 
     /// Get the URL of a submodule from git config
-    pub fn submodule_url(self, path: impl AsRef<OsStr>) -> Result<Option<String>> {
+    pub fn submodule_url(self, path: &Path) -> Result<Option<String>> {
         self.cmd()
-            .args([
-                "config",
-                "--get",
-                &format!("submodule.{}.url", path.as_ref().to_string_lossy()),
-            ])
+            .args(["config", "--get", &format!("submodule.{}.url", path.to_slash_lossy())])
             .get_stdout_lossy()
             .map(|url| Some(url.trim().to_string()))
     }
--- a/crates/forge/src/cmd/remove.rs
+++ b/crates/forge/src/cmd/remove.rs
@@ -31,23 +31,8 @@ impl_figment_convert_basic!(RemoveArgs);
 impl RemoveArgs {
     pub fn run(self) -> Result<()> {
         let config = self.load_config()?;
-        let (root, mut paths, _) = super::update::dependencies_paths(&self.dependencies, &config)?;
-
-        // Canonicalize paths to ensure consistent path separators for git submodule config lookup
-        paths = paths
-            .into_iter()
-            .map(|path| {
-                dunce::canonicalize(root.join(&path))
-                    .ok()
-                    .and_then(|canonical| {
-                        canonical.strip_prefix(&root).ok().map(|p| p.to_path_buf())
-                    })
-                    .unwrap_or(path)
-            })
-            .collect();
-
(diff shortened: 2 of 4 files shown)


yash-atreya (changes requested):
@wellnana Can we use `dunce::canonicalize` on the `paths` returned here: https://github.com/wellnana/foundry/blob/85587aa848a5cbddf5720495583e1551e40a2f05/crates/forge/src/cmd/remove.rs#L34

crates/cli/src/utils/mod.rs
```diff
@@ -719,12 +719,10 @@ ignore them in the `.gitignore` file."
 
     /// Get the URL of a submodule from git config
     pub fn submodule_url(self, path: impl AsRef<OsStr>) -> Result<Option<String>> {
+        // Git always uses forward slashes in submodule config keys, even on Windows
+        let path_str = path.as_ref().to_string_lossy().replace('\\', "/");
```
yash-atreya: @wellnana Can we use dunce::canonicalize on the paths returned here: https://github.com/wellnana/foundry/blob/85587aa848a5cbddf5720495583e1551e40a2f05/crates/forge/src/cmd/remove.rs#L34 

Instead of replacing

yash-atreya (dismissed):

crates/cli/src/utils/mod.rs
wellnana: Sure, should be good now.  @yash-atreya

crates/forge/src/cmd/remove.rs
```diff
@@ -31,7 +31,29 @@ impl_figment_convert_basic!(RemoveArgs);
...
+        paths = paths
+            .into_iter()
+            .map(|path| {
+                dunce::canonicalize(root.join(&path))
+                    .ok()
+                    .and_then(|canonical| {
+                        canonical.strip_prefix(&root).ok().map(|p| {
+                            // Convert path separators to forward slashes for git config
+                            // compatibility
+                            PathBuf::from(p.to_string_lossy().replace('\\', "/"))
```
onbjerg: `dunce::canonicalize` should do this

crates/forge/src/cmd/remove.rs
```diff
@@ -31,7 +31,29 @@ impl_figment_convert_basic!(RemoveArgs);
...
+                        })
+                    })
+                    .unwrap_or_else(|| {
+                        // Fallback: ensure forward slashes even if canonicalization fails
+                        PathBuf::from(path.to_string_lossy().replace('\\', "/"))
```
onbjerg: in this case just leave it as is

wellnana:
Thanks for the review @onbjerg ! I've updated the code per your suggestions - removed the path separator replacement and now rely only on `dunce::canonicalize`, with the fallback keeping paths as-is.

onbjerg:
doesn't seem to work

crates/forge/src/cmd/remove.rs
```diff
@@ -31,7 +31,21 @@ impl_figment_convert_basic!(RemoveArgs);
 impl RemoveArgs {
     pub fn run(self) -> Result<()> {
         let config = self.load_config()?;
-        let (root, paths, _) = super::update::dependencies_paths(&self.dependencies, &config)?;
+        let (root, mut paths, _) = super::update::dependencies_paths(&self.dependencies, &config)?;
```
DaniPopes: shouldn't this be done inside of the function?

grandizzy:
issue here is that the command should be
```bash
> git config --get submodule.lib/forge-std.url
https://github.com/foundry-rs/forge-std
```
but on win path was `lib\forge-std` hence
```bash
git config --get submodule.lib\forge-std.url
```
returned nothing

grandizzy (approved):
lgtm, thank you, pending one more review CC @onbjerg

crates/forge/src/cmd/remove.rs
grandizzy: updated in https://github.com/foundry-rs/foundry/pull/11826/commits/613462a810729d565d8fdbd4c0f11ee2b1116550

Issue #11797, opened 2025-09-24
feat: beacon API support for blob fetching


sam-iamm:
### Component

Anvil

### Describe the feature you would like

Anvil supports anvil_getBlobsByTransactionHash and anvil_getBlobByHash to fetch blobs, but it does not support the [beacon API](https://ethereum.github.io/beacon-APIs/#/Beacon/getBlobSidecars) method to fetch blobs: `/eth/v1/beacon/blob_sidecars/{block_id}`

By supporting this beacon API method, we can test with anvil using code that expects access to a beacon API endpoint to fetch blobs.

### Additional context

This is useful for testing an L2 that sends blob transactions to Anvil and then needs a beacon API endpoint to fetch blobs and perform derivation. Our CL binary, for example, is unaware of the anvil methods and currently expects to use the beacon api to fetch blobs.

mattsse:
this endpoint is returning https://github.com/alloy-rs/alloy/blob/249310a0b503fb7db9f4e90ab44f2b3096bb2e9d/crates/rpc-types-beacon/src/sidecar.rs#L87-L87

what we can do is simply returning either returning

https://github.com/alloy-rs/alloy/blob/249310a0b503fb7db9f4e90ab44f2b3096bb2e9d/crates/eips/src/eip4844/sidecar.rs#L38-L38

for the entire block, or a flattened version of that


similar to

https://github.com/foundry-rs/foundry/blob/f3625e01bcc952522f1674d863d41a55ce671a79/crates/anvil/src/eth/api.rs#L1349-L1353

## TODO
* add endpoint like https://github.com/foundry-rs/foundry/blob/f3625e01bcc952522f1674d863d41a55ce671a79/crates/anvil/core/src/eth/mod.rs#L185-L187
* impl endpoint like anvil_getBlobsByTransactionHash

we can also add another endpoint that exactly mirrors the beacon api and use fake data when where we dont have access to

https://github.com/alloy-rs/alloy/blob/249310a0b503fb7db9f4e90ab44f2b3096bb2e9d/crates/rpc-types-beacon/src/sidecar.rs#L11-L11

mablr:
May I do it 👍

sam-iamm:
[...]

PR #11828, opened 2025-09-25
feat(anvil): add support for `anvil_getBlobSidecarsByBlockId`

mablr:
## Motivation

towards #11797

## Solution

- Added `anvil_getBlobSidecarsByBlockId` endpoint 
- Added primitive Engine REST API

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/anvil/server/src/beacon.rs
mattsse, line 1: this we can remove entirely

crates/anvil/src/eth/api.rs
```diff
@@ -1352,6 +1356,15 @@ impl EthApi {
         Ok(self.backend.get_blob_by_tx_hash(hash)?)
     }
 
+    /// Handler for RPC call: `anvil_getBlobSidecarsByBlockId`
+    pub fn anvil_get_blob_sidecars_by_block_id(
```
mattsse, line 1360: we can mention here that this returns the beacon api format and reference the doc

crates/anvil/src/server/handler.rs
mattsse, line 1: this we also dont need because the endpoints are all auto supported via the enum already

crates/anvil/src/server/mod.rs
mattsse, line 1: can undo this as well

crates/anvil/server/src/beacon.rs
```diff
@@ -0,0 +1,20 @@
...
+    routing::get,
+};
+
+use crate::{BeaconApiHandler, BeaconRequest, BeaconResponse};
+
+pub fn beacon_router<Beacon: BeaconApiHandler>(beacon: Beacon) -> Router {
+    Router::new()
+        .route("/v1/beacon/blob_sidecars/{block_id}", get(get_blob_sidecars_by_block_id::<Beacon>))
+        .with_state(beacon)
+}
```
mattsse: ah wait, we do need this because this is a regular get request -.-

mattsse (changes requested):
cool, need to think about the regular http server support a bit,

can we limit this pr to just the extra anvil endpoint that returns `BlobTransactionSidecar` (all blobs in a block)

and think about adding support for GET requests and the beacon route in a followup pr?

crates/anvil/server/src/beacon.rs
mablr: yess it was actually less trivial than I expected 😅

mablr:
no problem i'll split this PR 👌 (...tomorrow morning)

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -3281,6 +3283,27 @@ impl Backend {
         Ok(None)
     }
 
+    pub fn get_blob_sidecars_by_block_id(
+        &self,
+        block_id: BlockId,
+    ) -> Result<Option<Vec<BlobTransactionSidecar>>> {
```
mablr: We return a `Vec<BlobTransactionSidecar>`, is this what we want? or would it be better to flatten this into a single `BlobTransactionSidecar` ?

mattsse: let's do just `BlobTransactionSidecar` which is usually the format required

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -3281,6 +3283,27 @@ impl Backend {
...
+        if let Some(full_block) = self.get_full_block(block_id) {
+            Ok(Some(
+                full_block
+                    .into_transactions_iter()
+                    .map(TypedTransaction::try_from)
+                    .collect::<Result<Vec<_>, _>>()?
```
mattsse: do we even need this conversion?

mattsse (changes requested):
smol nit re response format

once merged we can start with the regular get endpoints

crates/anvil/src/eth/backend/mem/mod.rs
mablr: ok yess that's right, let's filter results instead

crates/anvil/src/eth/backend/mem/mod.rs
mablr: I think a basic `fold` will do the job as we don't have `FromIterator` impl for `BlobTransactionSidecar` (could be a follow up in alloy btw)

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -3281,6 +3283,29 @@ impl Backend {
         Ok(None)
     }
 
+    pub fn get_blob_sidecars_by_block_id(
+        &self,
+        block_id: BlockId,
+    ) -> Result<Option<BlobTransactionSidecar>> {
```
mablr, line 3289: We may remove Result<> wrapping as conversion errors are filtered ?

PR #11831, opened 2025-09-26
revert evm version incompat warning

grandizzy:
## Motivation
- reverts changes added in https://github.com/foundry-rs/foundry/commit/d22355229fb23e8bf7a32b4785856e93c5331cf9

There are two issues with the approach:
- warning appears in all commands involving project config (forge clean, etc.)
- warning appears multiple times and wrong 

E.g.
on https://github.com/pcaversaccio/createx
```
> forge clean

Warning: evm_version 'paris' may be incompatible with solc version. Consider using 'prague'
Warning: evm_version 'paris' may be incompatible with solc version. Consider using 'shanghai'
```

on https://github.com/ithacaxyz/account
```
> forge clean

Warning: evm_version 'paris' may be incompatible with solc version. Consider using 'prague'
```
even if `evm_version = "prague"` set

Suggest this to be implemented as part of `WarningProviders` https://github.com/foundry-rs/foundry/blob/954accf6d7dcecd69ac62754d5f1911281a9ae15/crates/config/src/providers/warnings.rs#L9 and make sure all cases are covered

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -2218,44 +2218,18 @@ impl Config {
     /// This normalizes the default `evm_version` if a `solc` was provided in the config.
     ///
     /// See also <https://github.com/foundry-rs/foundry/issues/7014>
-    #[expect(clippy::disallowed_macros)]
     fn normalize_defaults(&self, mut figment: Figment) -> Figment {
-        let evm_version = figment.extract_inner::<EvmVersion>("evm_version").ok().or_else(|| {
-            figment
-                .extract_inner::<String>("evm_version")
-                .ok()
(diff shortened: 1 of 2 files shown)


0xrusowsky (approved):
LGTM

maybe we want to document why we we decided to not issue the warning in case that it is revisited in the future?

this way it would be easier to avoid doing the same mistake again.

grandizzy:
> LGTM
> 
> maybe we want to document why we we decided to not issue the warning in case that it is revisited in the future?
> 
> this way it would be easier to avoid doing the same mistake again.

the unit test I added should cover

Issue #11835, opened 2025-09-26
bug(fmt): don't break bin exprs when they overflow but fit


0xrusowsky:
### Component

Forge

### Have you ensured that all of these are up to date?

- [ ] Foundry
- [ ] Foundryup

### Operating System

None

### Describe the bug

identified some edgecases when aligning complex calls in uni `v4-core` repo:
```
-                feeGrowthInside0X128 =
-                    self.feeGrowthGlobal0X128 - lower.feeGrowthOutside0X128 - upper.feeGrowthOutside0X128;
-                feeGrowthInside1X128 =
-                    self.feeGrowthGlobal1X128 - lower.feeGrowthOutside1X128 - upper.feeGrowthOutside1X128;
+                feeGrowthInside0X128 = self.feeGrowthGlobal0X128 - lower.feeGrowthOutside0X128
+                    - upper.feeGrowthOutside0X128;
+                feeGrowthInside1X128 = self.feeGrowthGlobal1X128 - lower.feeGrowthOutside1X128
+                    - upper.feeGrowthOutside1X128;
```

grandizzy:
addressed in https://github.com/foundry-rs/foundry/pull/11837

PR #11837, opened 2025-09-26
fix(fmt): don't break exprs that overflow but fit assignement

0xrusowsky:
## Motivation

closes:
- [x] https://github.com/foundry-rs/foundry/issues/11835

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.24.2"
+version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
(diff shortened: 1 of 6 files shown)


crates/fmt/src/state/mod.rs
```diff
@@ -369,20 +367,29 @@ impl State<'_, '_> {
...
+
             for line in snip.lines() {
+                let line = line.trim();
+
                 if prev_needs_space {
-                    // Previous line ended with a bracket and config with bracket spacing.
-                    // Previous line ended with ',' a hardbreak or a space are required.
-                    // Previous line ended with ';' a hardbreak is required.
                     size += 1;
+                } else if !first && let Some(c) = line.chars().next() {
```
grandizzy: suggest simplify to
```rust
    } else if !first
        && let Some(c) = line.chars().next()
        && matches!(c, '&' | '|' | '=' | '>' | '<' | '+' | '-' | '*' | '/' | '%' | '^')
    {
        // if the line starts with an operator, a space or a line break are required.
        size += 1
    }
```

Issue #11841, opened 2025-09-26
fix(fmt): properly handle CRLF on windows files


0xrusowsky:
running `forge-cmd` on windows on https://github.com/0xProject/0x-settler panics.

the panic seems to be related CRLF because of a trailing comment within a disabled line

PR #11843, opened 2025-09-26
fix(fmt): handle CRLF when needed

0xrusowsky:
closes:
- [x] https://github.com/foundry-rs/foundry/issues/11841

### DIFF
--- a/crates/fmt/src/pp/helpers.rs
+++ b/crates/fmt/src/pp/helpers.rs
@@ -1,6 +1,30 @@
 use super::{Printer, Token};
 use std::borrow::Cow;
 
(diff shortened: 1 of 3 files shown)


grandizzy:
I suggest we should leave not formatted in such case and prevent panic in offset as https://github.com/foundry-rs/foundry/pull/11843/commits/a6d1e227e29dd84f6fd985f11541906cdc51a0b3

0xrusowsky (commented):
note that this only prevents the panic, but doesn't fix the underlying issue.

the output won't be properly formatted, as there won't be a line break with an offset, as what would happen with a unix compatible input file

PR #11844, opened 2025-09-26
ffi: Decode stdout as hex only when prefixed with 0x

maradini77:
### Description
- Clarifies `ffi` stdout handling to avoid accidental hex decoding of plain text.
- Now hex decoding occurs only when output starts with `0x`/`0X`.

- Changes:
  - Decode hex only if `stdout` has a `0x`/`0X` prefix; otherwise treat as raw text.
  - Added inline comment explaining the rationale.

- Why:
  - Prevents surprising behavior where hex-like strings (e.g., "deadbeef") were silently decoded into bytes.

- Compatibility:
  - If your command intends to return hex, prefix it with `0x` (e.g., `0xdeadbeef`).

### DIFF
--- a/crates/cheatcodes/src/fs.rs
+++ b/crates/cheatcodes/src/fs.rs
@@ -648,12 +648,19 @@ fn ffi(state: &Cheatcodes, input: &[String]) -> Result<FfiResult> {
         .output()
         .map_err(|err| fmt_err!("failed to execute command {cmd:?}: {err}"))?;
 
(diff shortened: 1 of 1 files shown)


onbjerg (changes requested):
This is a breaking change, I'm not sure if it's worth it at this point

mattsse (commented):
this will mostlikely break existting code so we shouldnt do this imo

onbjerg:
Agreed. Closing this PR because it will break too much and we don't currently plan to release a breaking version

PR #11847, opened 2025-09-27
Fix redirect loop vulnerability in npm installer

viktorking7:
The `makeRequest` function in `npm/src/install.ts` was vulnerable to infinite redirect loops and potential DoS attacks. The function followed redirects recursively without any limits or cycle detection, which could cause:

- Infinite loops when registry misconfiguration creates redirect cycles
- Excessive network usage and CI timeouts
- Potential SSRF bypasses through redirect chains

**Changes:**

- Added `maxRedirects` parameter (default: 5) to limit redirect chain length
- Implemented cycle detection using a `visited` Set to track URLs
- Added proper error handling for redirect limits and loops
- Maintained existing security checks and URL resolution logic

### DIFF
--- a/npm/src/install.ts
+++ b/npm/src/install.ts
@@ -38,42 +38,49 @@ function ensureSecureUrl(urlString: string, purpose: string) {
   }
 }
 
-function makeRequest(url: string): Promise<Buffer> {
+/**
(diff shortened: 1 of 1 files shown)


DaniPopes:

PR #11848, opened 2025-09-27
fix: correct tx gas limit cap check (Osaka / EIP-7825)

sashaodessa:
### Summary
Enforce the transaction gas limit cap only when explicitly enabled.

#### What changed
- Replaced `is_none()` with `is_some()` in:
  - `crates/anvil/src/eth/backend/mem/mod.rs`
  - `crates/anvil/src/eth/backend/executor.rs`

#### Why
- The previous logic inverted the check, causing the cap to not be enforced when configured.
- Aligns validation/execution with Osaka (EIP-7825) expectations and prevents oversized txs slipping through.


sashaodessa:
@onbjerg but you approved this #11783

Issue #11834, opened 2025-09-26
bug(fmt): edgecase in complex call alignment


0xrusowsky:
### Component

Forge

### Have you ensured that all of these are up to date?

- [ ] Foundry
- [ ] Foundryup

### What command(s) is the bug in?

forge-fmt

### Operating System

None

### Describe the bug

identified some edgecases when aligning complex calls in uni `v4-core` repo:
```
-    function test_ffi_fuzz_addLiquidity_defaultPool_ReturnsCorrectLiquidityDelta(
-        IPoolManager.ModifyLiquidityParams memory paramSeed
-    ) public {
+    function test_ffi_fuzz_addLiquidity_defaultPool_ReturnsCorrectLiquidityDelta(IPoolManager
+                .ModifyLiquidityParams memory paramSeed) public {
```
```
-            CurrenciesOutOfOrderOrEqual.selector.revertWith(
-                Currency.unwrap(key.currency0), Currency.unwrap(key.currency1)
-            );
+            CurrenciesOutOfOrderOrEqual.selector
+            .revertWith(Currency.unwrap(key.currency0), Currency.unwrap(key.currency1));
```

PR #11854, opened 2025-09-28
fix(fmt): indent calls with single member that fits but breaks when args

0xrusowsky:
## Motivation

closes:
- [x] https://github.com/foundry-rs/foundry/issues/11834
   - [x] member expr in function args
   - [X] indent calls with single member that fits but breaks when args

## PR Checklist

- [X] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 4
 
 [[package]]
 name = "addr2line"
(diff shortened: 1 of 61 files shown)


grandizzy (approved):
lgtm

PR #11859, opened 2025-09-28
fix(npm): robust fallback to dist binary in forge launcher

maximevtush:
### Summary
Ensure the CLI uses the bundled dist binary when the platform package resolves but its binary file is missing, improving reliability on partially installed or corrupted environments.

### Changes
- Refactor `getBinaryPath()` to:
  - Prefer resolved platform-specific binary only if the file exists.
  - Fallback to `dist/forge` (or `forge.exe` on Windows) when the resolved path is missing.
  - Exit with a clear error only if neither binary is available.

### Why
`require.resolve(...)` can succeed while the actual file is missing (e.g., cache/corruption). Previously, the launcher didn’t use the fallback in this case, causing unnecessary failures.

### DIFF
--- a/npm/src/forge.ts
+++ b/npm/src/forge.ts
@@ -9,14 +9,22 @@ const require = NodeModule.createRequire(import.meta.url)
 const __dirname = NodePath.dirname(fileURLToPath(import.meta.url))
 
 function getBinaryPath() {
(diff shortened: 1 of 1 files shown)


DaniPopes:

PR #11861, opened 2025-09-28
fix(npm/prepublish): stage binary into packagePath/bin and handle forge.exe on Windows

kilavvy:
This PR fixes two issues in the prepublish script: (1) the binary was being staged to a relative path @foundry-rs/.../bin instead of the prepared packagePath/bin, which could place it outside the package directory and break published artifacts; (2) findForgeBinary() hardcoded 'forge' and would fail to locate forge.exe on Windows. The changes ensure the binary is correctly placed inside the package directory and use the proper executable name based on platform (forge.exe for Windows, forge otherwise). This improves cross-platform reliability of the npm packaging process without breaking existing functionality.


DaniPopes:

PR #11862, opened 2025-09-28
fix(fmt): preserve comments in tuple LHS with uninitialized vars (DeclMulti)

### LINKED ISSUE #11836: bug(fmt): unable to place comments with uninitialized vars
### Component

Forge

- [ ] Foundry
- [ ] Foundryup

_No response_

_No response_

_No response_

### Operating System

None

### Describe the bug

because uninitialized vars in `stmt::DeclMulti` are `None`, we don't have their span information.
Thus making it impossible to properly place comments.

```rs
    /// A multi-variable declaration statement: `(bool success, bytes memory value) = ...;`.
    ///
    /// Multi-assignments require an expression on the right-hand side.
    DeclMulti(Box<'ast, [Option<VariableDefinition<'ast>>]>, Box<'ast, Expr<'ast>>),
```

example from `0x-settler`:
```
-       (
-            /* poolIndex */,
-            uint256 sellAmount1,
-            uint256 buyAmount1,
-            /* poolKey1 */,
-            /* sellToken */,
-            /* buyToken */,
-            /* sellTokenBalanceBefore */,
-            uint256 buyTokenBalanceBefore1,
-            /* hashMul */,
-            /* hashMod */
-        ) = _swapPre(2, TOTAL_SUPPLY / 1_000, false, zeroForOne1);

        (
+            , /* poolIndex */
+            uint256 sellAmount1,
+            uint256 buyAmount1, /* poolKey1 */ /* sellToken */ /* buyToken */ /* sellTokenBalanceBefore */ /* hashMul */
+            /* hashMod */
+            ,,,,
+            uint256 buyTokenBalanceBefore1,,
         ) = _swapPre(2, TOTAL_SUPPLY / 1_000, false, zeroForOne1);
```

Forostovec:
Fix #11836 

Problem: In multi-variable declarations (DeclMulti), empty tuple slots are represented as None, so they lack span information. Without spans, the formatter cannot anchor comments between commas for uninitialized positions, causing comments to merge, drift, or be misplaced.

Solution: Compute synthetic per-slot spans for the tuple LHS by scanning the source between the ( of the declaration and the start of the initializer =. Split at top-level commas (skipping nested parens and comments) to derive a span for each slot, including empty ones. Pass these spans to the list/tuple printer so print_comments can correctly place comments even when the slot has no variable.

### DIFF
--- a/crates/fmt/src/state/sol.rs
+++ b/crates/fmt/src/state/sol.rs
@@ -24,6 +24,110 @@ macro_rules! get_span {
 
 /// Language-specific pretty printing: Solidity.
 impl<'ast> State<'_, 'ast> {
+    /// Compute per-slot spans for a parenthesized tuple between `pos_lo` (inclusive)
+    /// and `pos_hi` (exclusive). This enables anchoring comments for empty tuple slots.
+    fn compute_tuple_slot_spans(
+        &self,
+        pos_lo: BytePos,
+        pos_hi: BytePos,
+        expected_slots: usize,
+    ) -> Vec<Option<Span>> {
+        let mut result: Vec<Option<Span>> = Vec::with_capacity(expected_slots);
+        let Ok(snippet) = self.sm.span_to_snippet(Span::new(pos_lo, pos_hi)) else {
+            result.resize(expected_slots, None);
+            return result;
+        };
+
+        let bytes = snippet.as_bytes();
+        let len = bytes.len();
+
+        // Find opening '('
+        let mut open_idx = 0usize;
+        while open_idx < len && bytes[open_idx] != b'(' {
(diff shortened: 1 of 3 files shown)


0xrusowsky:
First of all thanks for the PR!

Despite the solution is pragmatic and it works, imo it doesn't make sense for the formatter to have to parse source code that solar has already processed.

Instead we should do a PR upstream to solar, and then use the spans (hence why i had the issue labelled as "blocked by solar" in the meta ticket).

Note that users can always disable the formatter in that code block.

With that being said though, i'll defer the decision of whether to accept the proposed fix or not to @DaniPopes and @grandizzy

PR #11868, opened 2025-09-29
fix: replace unsafe unwrap() calls with proper error handling in linking

DeVikingMark:
This commit fixes three critical panic vulnerabilities in the linking module:

1. Line 180: Replace get_bytecode_bytes().unwrap() with safe error handling
   - get_bytecode_bytes() can return None for contracts without bytecode
   - Using unwrap() could cause runtime panics in production

2. Line 210: Replace double unwrap() in link().unwrap().bytecode.unwrap()
   - Both link() and bytecode field can fail/be None
   - Double unwrap() creates two potential panic points

3. Line 231: Replace bytecode.bytes().unwrap() with safe error handling
   - bytes() method can return None when bytecode object is missing
   - Using unwrap() could cause runtime panics

Fixes potential runtime panics that could occur when:
- Contract artifacts have missing or invalid bytecode
- Library linking fails due to missing dependencies
- Bytecode objects are in an unexpected state

### DIFF
--- a/crates/linking/src/lib.rs
+++ b/crates/linking/src/lib.rs
@@ -177,7 +177,12 @@ impl<'a> Linker<'a> {
         let libs_to_deploy = libs_to_deploy
             .into_par_iter()
             .map(|(id, _)| {
-                Ok(self.link(id, &libraries)?.get_bytecode_bytes().unwrap().into_owned())
+                let linked_contract = self.link(id, &libraries)?;
+                let bytecode = linked_contract.get_bytecode_bytes()
+                    .ok_or_else(|| LinkerError::LinkingFailed { 
+                        artifact: id.source.to_string_lossy().into() 
+                    })?;
+                Ok(bytecode.into_owned())
             })
             .collect::<Result<Vec<_>, LinkerError>>()?;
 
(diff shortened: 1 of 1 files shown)


crates/linking/src/lib.rs
```diff
@@ -177,7 +177,12 @@ impl<'a> Linker<'a> {
         let libs_to_deploy = libs_to_deploy
             .into_par_iter()
             .map(|(id, _)| {
-                Ok(self.link(id, &libraries)?.get_bytecode_bytes().unwrap().into_owned())
+                let linked_contract = self.link(id, &libraries)?;
+                let bytecode = linked_contract.get_bytecode_bytes()
+                    .ok_or_else(|| LinkerError::LinkingFailed { 
+                        artifact: id.source.to_string_lossy().into() 
+                    })?;
```
onbjerg, line 184: the unwrap here should be sound, given a return of `Ok(_)` on `Self::link` implies the existence of a linked bytecode, which also means `get_bytecode_bytes()` cannot return `Err(_)`.

crates/linking/src/lib.rs
```diff
@@ -207,10 +212,17 @@ impl<'a> Linker<'a> {
...
-                (id, bytecode)
+                let linked_contract = self.link(id, &libraries)
+                    .map_err(|_| LinkerError::LinkingFailed { 
+                        artifact: id.source.to_string_lossy().into() 
+                    })?;
+                let bytecode = linked_contract.bytecode
+                    .ok_or_else(|| LinkerError::LinkingFailed { 
+                        artifact: id.source.to_string_lossy().into() 
+                    })?;
+                Ok((id, bytecode))
```
onbjerg, line 223: same here

PR #11869, opened 2025-09-29
feat(forge): no warning on external config sections

### LINKED ISSUE #5866: feat: support external sections in `foundry.toml`
### Component

Forge

There are various requests and use cases for adding support for custom sections to `foundry.toml`:
- [scopelint](https://github.com/ScopeLift/scopelint) is a tool for linting forge projects. Instead of having it's own config file, it would be nice to just let users define the settings directly in the forge config file
- In https://github.com/foundry-rs/foundry/issues/4808, there's requests to support parsing the config file so it can be extended to contain settings for [forge deploy](https://github.com/wighawag/forge-deploy) or Noir circuits

Currently, you can add a custom section to the config file, but it will trigger a warning:

```
$ forge test
warning: Unknown section [scopelint] found in foundry.toml. This notation for profiles has been deprecated and may result in the profile not being registered in future versions. Please use [profile.scopelint] instead or run `forge config --fix`.
```

Therefore, I'd suggest requiring that unknown sections are prefixed with something like `external.`, which makes it clear they are for other tools that extend foundry and we'd suppress the warning for those sections.

For example:

```toml
# Default foundry configuration
[profile.default]
optimizer = false

[profile.ci]
optimizer = true

# Custom sections for other tools
[external.scopelint]
some_flag = 1

[external.forge_deploy]
another_setting = 123
```

### Additional context

_No response_

grandizzy:
## Motivation
- closes #5866 no warning on `external.` sections

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -679,6 +679,9 @@ impl Config {
     /// TOML section for profiles
     pub const PROFILE_SECTION: &'static str = "profile";
 
+    /// External config sections, ignored from warnings.
+    pub const EXTERNAL_SECTION: &'static str = "external";
+
     /// Standalone sections in the config which get integrated into the selected profile
     pub const STANDALONE_SECTIONS: &'static [&'static str] = &[
         "rpc_endpoints",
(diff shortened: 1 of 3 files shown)


onbjerg (approved):
impl is sound, wondering if `external` is the right name, but I don't want to bikeshed this

PR #11873, opened 2025-09-29
fix(forge): same way to flatten for all commands

### LINKED ISSUE #11754: `forge verify-contract --flatten` produces different flattened code than `forge flatten` command
### Component

Forge

- [x] Foundry
- [x] Foundryup

1.3.5-stable

0.3.3

forge flatten / forge verify-contract --flatten

### Operating System

macOS (Apple Silicon)

### Describe the bug

The `forge verify-contract --flatten` command produces different flattened source code compared to the standalone `forge flatten` command. This inconsistency causes verification failures on Etherscan because the bytecode generated from the two different flattened versions doesn't match.

### Specific Differences Found:
1. SPDX License Identifier Placement:
- `forge flatten`: Places SPDX license at the very top of the file (line 1).
- `forge verify-contract --flatten`: Places SPDX license after the first file's content.
2. Interface Name Handling:
- `forge flatten`: Adds suffixes to disambiguate duplicate interface names (e.g., `OracleLike_0`, `OracleLike_1`)
- `forge verify-contract --flatten`: Creates interfaces without suffixes (e.g., `OracleLike`)

### Impact:
- Verification failures on Etherscan due to bytecode mismatch
- Inconsistent behavior between two commands that should produce identical output
- Developers cannot rely on `forge flatten` output to match what gets sent for verification

### Reproduction Steps:
1. Create a project with multiple files containing duplicate interface names
2. Run `forge flatten src/Contract.sol --output flat.sol`
3. Run `forge verify-contract --flatten --skip-is-verified-check <address> src/Contract.sol:Contract`. To capture the source, you can create a fake local verifier server.
4. Compare the two outputs - they will differ in SPDX placement and interface naming

### Expected Behavior:
Both commands should produce identical flattened source code, ensuring that local flattening matches what gets sent for verification.

### Evidence:
- File size difference
- Different checksums
- Interface naming differences cause different bytecode generation

This bug makes contract verification unreliable and forces developers to use workarounds or manual verification processes.

grandizzy:
## Motivation
- closes #11754 use consistent flatten for `forge flatten`, `forge verify-contract` and `cast src`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cast/src/lib.rs
+++ b/crates/cast/src/lib.rs
@@ -30,10 +30,10 @@ use foundry_common::{
     TransactionReceiptWithRevertReason,
     abi::{coerce_value, encode_function_args, get_event, get_func},
     compile::etherscan_project,
+    flatten,
     fmt::*,
     fs, get_pretty_tx_receipt_attr, shell,
 };
-use foundry_compilers::flatten::Flattener;
 use foundry_config::Chain;
 use foundry_evm_core::ic::decode_instructions;
 use futures::{FutureExt, StreamExt, future::Either};
@@ -2192,7 +2192,7 @@ impl SimpleCast {
         let project = etherscan_project(metadata, tmp.path())?;
         let target_path = project.find_contract_path(&metadata.contract_name)?;
 
-        let flattened = Flattener::new(project, &target_path)?.flatten();
+        let flattened = flatten(project, &target_path)?;
 
         if let Some(path) = output_path {
             fs::create_dir_all(path.parent().unwrap())?;
--- a/crates/common/src/utils.rs
+++ b/crates/common/src/utils.rs
@@ -1,9 +1,14 @@
 //! Uncategorised utilities.
 
 use alloy_primitives::{B256, Bytes, U256, hex, keccak256};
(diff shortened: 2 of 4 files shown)


DaniPopes:
in forge flatten we have a fallback for using the old flattener: https://github.com/foundry-rs/foundry/blob/1cc7cc9e54442be31ca03ceb794d9c7ad1d3d9b3/crates/forge/src/cmd/flatten.rs#L52

can we unify all the invocations to a single common util function that handles this?

grandizzy:
> in forge flatten we have a fallback for using the old flattener:
> 
> https://github.com/foundry-rs/foundry/blob/1cc7cc9e54442be31ca03ceb794d9c7ad1d3d9b3/crates/forge/src/cmd/flatten.rs#L52
> 
> can we unify all the invocations to a single common util function that handles this?

good point, added in https://github.com/foundry-rs/foundry/pull/11873/commits/d60e7b2d431cdcadd0203f1684b5d38369991151 and extended to `cast src` as well

DaniPopes (approved):

PR #11874, opened 2025-09-29
fix(fmt): account for CRLF when handling cursor

0xrusowsky:
## Motivation

closes:
- [x] https://github.com/foundry-rs/foundry/issues/11841

## Solution

rather than blindly advancing the cursor by 1, we need to check if the current character is `\r` to account for the whole CRLF sequence and advance 2 chars instead.

## PR Checklist

- [ ] Added Tests
- [X] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/fmt/src/state/sol.rs
+++ b/crates/fmt/src/state/sol.rs
@@ -155,7 +155,7 @@ impl<'ast> State<'_, 'ast> {
         self.print_comments(span.hi(), CommentConfig::default());
         self.print_trailing_comment(span.hi(), None);
         self.hardbreak_if_not_bol();
-        self.cursor.advance(if self.is_at_crlf() { 2 } else { 1 });
+        self.cursor.next_line(self.is_at_crlf());
     }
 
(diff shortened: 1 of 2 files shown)


crates/fmt/src/state/sol.rs
```diff
@@ -25,6 +25,11 @@ macro_rules! get_span {
 /// Language-specific pretty printing: Solidity.
 impl<'ast> State<'_, 'ast> {
     pub(crate) fn print_source_unit(&mut self, source_unit: &'ast ast::SourceUnit<'ast>) {
+        // Figure out if the cursor needs to check for CR (`\r`).
+        if let Some(item) = source_unit.items.first() {
+            self.check_crlf(item.span.to(source_unit.items.last().unwrap().span));
+        }
```
0xrusowsky, line 31: i believe that initially checking if the file contains `\r` to then skip checking the sourcemap each time is more efficient, but feel free to push back if u disagree

PR #11876, opened 2025-09-29
fix(fmt): always return formatted when source read from stdin

### LINKED ISSUE #11871: Formatter integration in VS Code is broken (with latest nightly)
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.3.5-stable

foundryup: 1.3.0

_No response_

### Operating System

macOS (Apple Silicon)

### Describe the bug

Latest nightly build doesn't work well with VS Code: the option "format on save" (using Forge as the formatter) will empty the file before saving it. This behavior is not present in the latest stable.

grandizzy:
## Motivation
- closes #11871
- always return formatted when source read from stdin (and no other args `check` or `raw` passed)
- vscode plugin use stdin and replace content with `forge fmt` output

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/forge/src/cmd/fmt.rs
+++ b/crates/forge/src/cmd/fmt.rs
@@ -127,12 +127,19 @@ impl FmtArgs {
                     let path = source_unit.file.name.as_real();
                     let original = source_unit.file.src.as_str();
                     let formatted = forge_fmt::format_ast(gcx, source_unit, fmt_config.clone())?;
+                    let from_stdin = path.is_none();
+
+                    // Return formatted code when read from stdin without check or raw switch.
+                    // <https://github.com/foundry-rs/foundry/issues/11871>
+                    if from_stdin && !self.check && !self.raw {
+                        return Some(Ok(formatted));
+                    }
 
                     if original == formatted {
                         return None;
                     }
 
-                    if self.check || path.is_none() {
+                    if self.check || from_stdin {
                         let summary = if self.raw {
                             formatted
                         } else {
--- a/crates/forge/tests/cli/fmt.rs
+++ b/crates/forge/tests/cli/fmt.rs
@@ -89,3 +89,12 @@ Diff in stdin:
 
 "#]]);
 });
+
+// Test that original is returned if read from stdin and no diff.
+// <https://github.com/foundry-rs/foundry/issues/11871>
+forgetest!(fmt_stdin_original, |_prj, cmd| {
+    cmd.args(["fmt", "-"]);
+
+    cmd.stdin(FORMATTED.as_bytes());
+    cmd.assert_success().stdout_eq(FORMATTED.as_bytes());
+});


crates/forge/src/cmd/fmt.rs
```diff
@@ -157,30 +157,31 @@ impl FmtArgs {
...
+                        if self.check {
+                            let name = path
+                                .strip_prefix(&config.root)
+                                .unwrap_or(path)
+                                .display()
+                                .to_string();
+                            let summary = format_diff_summary(&name, &diff);
+                            Some(Ok(summary))
+                        } else {
+                            match fs::locked_write(path, formatted) {
```
DaniPopes: why are we locking? this implies that readers also lock but i dont think that is the case

crates/forge/src/cmd/fmt.rs
```diff
@@ -157,30 +157,31 @@ impl FmtArgs {
...
-                            &TextDiff::from_lines(original.as_str(), &formatted),
-                        );
-                        Some(Ok(summary))
-                    } else {
-                        match fs::write(path, formatted) {
-                            Ok(_) => {
-                                let _ = sh_println!("Formatted {}", path.display());
-                                None
+                    let diff = TextDiff::from_lines(original.as_str(), &formatted);
+                    if diff.ratio() < 1.0 {
```
DaniPopes: is this not the same as original == formatted ?

grandizzy: tbh I have to check this more and find a way to repro first, I guess they're the same ... moving in wip

grandizzy: brain fart, pls ignore

crates/forge/src/cmd/fmt.rs
grandizzy: brain fart, pls ignore

crates/forge/src/cmd/fmt.rs
```diff
@@ -127,12 +127,15 @@ impl FmtArgs {
                     let path = source_unit.file.name.as_real();
                     let original = source_unit.file.src.as_str();
                     let formatted = forge_fmt::format_ast(gcx, source_unit, fmt_config.clone())?;
+                    let from_stdin = path.is_none();
 
                     if original == formatted {
-                        return None;
+                        // Return original code if read from stdin.
+                        // <https://github.com/foundry-rs/foundry/issues/11871>
+                        return (from_stdin && !self.check && !self.raw).then_some(Ok(formatted));
```
0xrusowsky: maybe we could enhance the cmnt with something like:
```
if user passes the `--check` or `--raw` flag they are performing
a check and only want to get an output if there is an actual diff
```

but its implicit in the flags, so no need if u don't want to

DaniPopes (approved):

Issue #4232, opened 2023-01-31
TX fails due to `PrevrandaoNotSet` on non-mainnet forks


Craigson:
### Component

Anvil

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### What version of Foundry are you on?

forge 0.2.0 (31f945c 2023-01-30T00:05:12.356073Z)

### What command(s) is the bug in?

anvil --fork-url https://moonbeam.blastapi.io/<api_key>

### Operating System

macOS (Apple Silicon)

### Describe the bug

Attempting to deploy a contract to local fork of the Moonbeam network

Run the local node:

```
anvil --fork-url https://moonbeam.blastapi.io/<api_key>
```

Deploy the contract:

```
forge script scripts/foundry/DeployGarage.s.sol:DeployGarage --rpc-url http://localhost:8545 --broadcast --legacy
```

Fails with:

```
Fork
==================
Endpoint:       https://moonbeam.blastapi.io/<api_key>
Block number:   2844720
Block hash:     0x1ee3577268a3d1f1fcb2da74d0b1cd4f56a6ccc89abe80771486551d3e9bdc2b
Chain ID:       1284
...
eth_blockNumber
eth_chainId
eth_getBlockByNumber
eth_chainId
eth_chainId
eth_sendRawTransaction

    Transaction: 0xd76504b4c53e2f0051c26f106469a0d47d1807fb3cc865aab50f064e017991cc
    Gas used: 0
    Error: failed due to PrevrandaoNotSet

    Block Number: 2844721
    Block Hash: 0x98f3806dff433e5c3d38550262854e0763980f3fa32bb0287356ff228df44964
    Block Time: "Tue, 31 Jan 2023 00:30:23 +0000"

eth_getTransactionReceipt
```

mattsse:
can you please post an `eth_getBlockByHash` response from that provider?

Craigson:
@mattsse :

```
curl --location --request POST 'localhost:8545/' \
--header 'Content-Type: application/json' \
--data-raw '{
        "jsonrpc":"2.0",
        "method":"eth_getBlockByHash",
        "params":[
                "0x84fa58a278a7fd1782796fb62f728c5cf687212a022c621ecc03532be4f64eb3",
                true
        ],
        "id":1
}'
```

with response:

```json
[...]

PR #11877, opened 2025-09-29
fix(prevrandao not set): support gnosis network

chen4903:
## Motivation

I make some tests in Gnosis network:
```solidity
    function setUp() public {
        string memory RPC_URL = vm.envString("GNOSIS_MAINNET_RPC");
        vm.createSelectFork(RPC_URL, 21_120_284 - 1);
```
Then I get that error:
```bash
[FAIL: EVM error; header validation error: `prevrandao` not set]
```
I find the relevant issues: https://github.com/foundry-rs/foundry/issues/4232

Here is the provider I use:
```bash
https://weathered-silent-panorama.xdai.quiknode.pro/xxxxxxxxxxxx
```

## Solution

Just add `Gnosis`

## PR Checklist

btw I am not sure how to run the test. Could u kindly provide a demo so that I could test it in my side? And then I could add a test case in this PR.

image

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/core/src/utils.rs
+++ b/crates/evm/core/src/utils.rs
@@ -55,7 +55,7 @@ pub fn apply_chain_and_block_specific_env_changes<N: Network>(
                 env.block.prevrandao = Some(env.block.difficulty.into());
                 return;
             }
-            Moonbeam | Moonbase | Moonriver | MoonbeamDev | Rsk | RskTestnet | Quicknode => {
+            Moonbeam | Moonbase | Moonriver | MoonbeamDev | Rsk | RskTestnet | Gnosis => {
                 if env.block.prevrandao.is_none() {
                     // <https://github.com/foundry-rs/foundry/issues/4232>
                     env.block.prevrandao = Some(B256::random());
--- a/crates/evm/core/src/utils.rs
+++ b/crates/evm/core/src/utils.rs
@@ -55,7 +55,7 @@ pub fn apply_chain_and_block_specific_env_changes<N: Network>(
                 env.block.prevrandao = Some(env.block.difficulty.into());
                 return;
             }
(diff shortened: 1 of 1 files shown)


crates/evm/core/src/utils.rs
```diff
@@ -55,7 +55,7 @@ pub fn apply_chain_and_block_specific_env_changes<N: Network>(
                 env.block.prevrandao = Some(env.block.difficulty.into());
                 return;
             }
-            Moonbeam | Moonbase | Moonriver | MoonbeamDev | Rsk | RskTestnet => {
+            Moonbeam | Moonbase | Moonriver | MoonbeamDev | Rsk | RskTestnet | Gnosis => {
```
mattsse: can we also do this for gnosis chiadao?

mattsse (changes requested):
smol nit

crates/evm/core/src/utils.rs
chen4903: I added 2: `GnosisTestnet` and `GnosisChiado`. Should we only maintain one of them?


https://github.com/foundry-rs/foundry/pull/11877/commits/45c3aa3c18d4f662992230b3ad52ca20c948be79

chen4903: ohoh I get it: https://github.com/alloy-rs/chains/blob/748003102157fbc20170e466108d49ce8edbd223/src/named.rs#L190

let me update it, one sec

chen4903: added: https://github.com/foundry-rs/foundry/pull/11877/commits/16660523413da0114fa131b4ea23618d5a84b9c6

chen4903:
squash merge. plz forgive my above messy commits

Issue #9625, opened 2025-01-05
bug: inline config errors are misleading


DaniPopes:
another thing is that `line` is documented and interpreted as `{line}:{column}:{length}`, but is in fact `{byte_start}:{byte_end}:{source_id}` in both parsers. This should be fixed in another PR

_Originally posted by @DaniPopes in https://github.com/foundry-rs/foundry/pull/9615#discussion_r1902001087_

This refers to when inline config errors are printed like `{path}:1:2:3`, the `1:2:3` should be `{line}:{column}` but is instead something else

grandizzy:
@DaniPopes is this still relevant with latest inline config fixes?

PR #11883, opened 2025-09-30
fix: inline config locations

DaniPopes:
Fixes https://github.com/foundry-rs/foundry/issues/9625.


grandizzy (approved):
lgtm, one test fix needed

Issue #11760, opened 2025-09-22
feat: lint to enforce declaring struct fields


0xClandestine:
### Component

Forge

### Describe the feature you would like

Add a lint that enforces explicit field naming when initializing structs, improving code readability and maintainability by requiring named parameters instead of positional arguments.

```solidity
struct Person {
    string name;
    uint256 age;
    address wallet;
}

function namedArgs() public {
    Person memory person = Person({
        name: "Alice",
        age: 25,
        wallet: 0x123...
    });
}

function positionalArgs() public {
    Person memory person = Person("Alice", 25, 0x123...);
}
```

0xClandestine:
May be a nice addition to #10907

grandizzy:
@0xClandestine we're looking to make this a lint, would that work? thank you

0xClandestine:
> [@0xClandestine](https://github.com/0xClandestine) we're looking to make this a lint, would that work? thank you

I think a lint could work (especially with auto-fix), however this feels more like a formatting issue since both approaches are equally valid, it's purely stylistic. Happy either way, just curious what the rationale is for `lint` over `fmt`?

A quick side note I forgot to mention: when using named fields, they don’t need to match the struct’s declaration order. However, for readability, the rule should probably enforce that order as well.

DaniPopes:
The formatter doesn't know anything about the semantics of the program. It may be a purely visual change but it does require knowing where the struct constructors are and what struct they refer to, which is out of scope for a formatter which does not need this information otherwise.

0xrusowsky:
impl guidance:

- follow the steps specified in https://github.com/foundry-rs/foundry/blob/master/docs/dev/lintrules.md#L1
- create a new struct for the lint and impl [trait](https://github.com/foundry-rs/foundry/blob/master/crates/lint/src/linter/late.rs#L8) `LateLintPass<'hir>`
- check for `ExprKind::Call` expressions where the callee is a struct type `ExprKind::Type` + `TypeKind::Custom`
- if the call uses `CallArgsKind::Unnamed` emit the lint
- use the gcx to resolve the type and create the fix suggestion using the struct fields.

the lint should have informational severity

0xClandestine:
thanks @onbjerg

mablr:
Hey @0xClandestine !
I implemented the fix suggestions for the new lint yesterday before I saw that you had closed the issue as resolved. I just published the PR, as I think it's still relevant.

[...]

PR #11892, opened 2025-09-30
feat: struct initializer lint

onbjerg:
Adds the lint in #11760, without a snippet. I'll leave the issue open for the snippet portion.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/lint/src/sol/info/named_struct_fields.rs
+++ b/crates/lint/src/sol/info/named_struct_fields.rs
@@ -1,0 +1,31 @@
+use solar::sema::hir::{CallArgs, CallArgsKind, Expr, ExprKind, ItemId, Res};
+
+use crate::{
+    linter::{LateLintPass, LintContext},
+    sol::{Severity, SolLint, info::NamedStructFields},
+};
+
+declare_forge_lint!(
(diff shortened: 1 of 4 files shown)


0xrusowsky (approved):
makes sense!

it's fine to wait for the snippet as with the solar release i'll change the api to emit suggestions 🤝

crates/lint/src/sol/info/named_struct_fields.rs
```diff
@@ -0,0 +1,34 @@
...
+        _hir: &'hir solar::sema::hir::Hir<'hir>,
+        expr: &'hir solar::sema::hir::Expr<'hir>,
+    ) {
+        if let ExprKind::Call(
+            Expr { kind: ExprKind::Ident(idents), .. },
+            CallArgs { kind: CallArgsKind::Unnamed(_args), .. },
+            _,
+        ) = &expr.kind
+        {
+            let [Res::Item(ItemId::Struct(_struct_id))] = idents else {
```
DaniPopes: you can flatten with let chains, even the individual patterns in exprcall if you want

PR #11895, opened 2025-09-30
fix: only include item's source in docs

onbjerg:
## Motivation

A change in https://github.com/foundry-rs/foundry/pull/10907 led `forge doc` to include the entire source file as the source code for each item in the contract.

## Solution

Attempt to format the file once before we parse it, then walk the parse tree, getting the correct source ranges for each item.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/doc/src/parser/item.rs
```diff
@@ -81,11 +77,10 @@ impl ParseItem {
         self
     }
 
-    /// Set formatted code on the [ParseItem].
-    pub fn with_code(mut self, source: &str, config: FormatterConfig) -> ParserResult<Self> {
-        self.code =
-            forge_fmt::format(source, config).into_result().map_err(ParserError::Formatter)?;
-        Ok(self)
+    /// Set code on [ParseItem].
```
DaniPopes: mention this should be the source code of the file

PR #11896, opened 2025-09-30
perf(anvil): optimize internal data passing in log queries

silvekkk:
Improves performance of log query operations by optimizing how data is passed between internal functions. Refactoring reduces redundant operations by passing additional context that's already available at call sites, resulting in measurable performance improvements particularly for range queries. 

Internal optimization only.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -2093,7 +2093,7 @@ impl Backend {
         hash: B256,
     ) -> Result<Vec<Log>, BlockchainError> {
         if let Some(block) = self.blockchain.get_block_by_hash(&hash) {
-            return Ok(self.mined_logs_for_block(filter, block));
+            return Ok(self.mined_logs_for_block(filter, block, hash));
         }
 
         if let Some(fork) = self.get_fork() {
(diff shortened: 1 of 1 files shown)


crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -2167,8 +2166,12 @@ impl Backend {
...
 
         for number in from..=to {
-            if let Some(block) = self.get_block(number) {
-                all_logs.extend(self.mined_logs_for_block(filter.clone(), block));
+            let storage = self.blockchain.storage.read();
+            if let Some(&block_hash) = storage.hashes.get(&number)
+                && let Some(block) = storage.blocks.get(&block_hash).cloned()
+            {
+                drop(storage);
+                all_logs.extend(self.mined_logs_for_block(filter.clone(), block, block_hash));
```
onbjerg, line 2174: idt this change makes sense, using `get_block` is fine, what is the reasoning?

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -2104,9 +2104,8 @@ impl Backend {
     }
 
     /// Returns all `Log`s mined by the node that were emitted in the `block` and match the `Filter`
-    fn mined_logs_for_block(&self, filter: Filter, block: Block) -> Vec<Log> {
+    fn mined_logs_for_block(&self, filter: Filter, block: Block, block_hash: B256) -> Vec<Log> {
```
onbjerg, line 2107: 👍

crates/anvil/src/eth/backend/mem/mod.rs
silvekkk: The change avoids redundant `hash_slow()` calls. `get_block(number)` internally does `hashes.get(&number)` + `blocks.get(&hash)` but discards the hash, forcing `mined_logs_for_block` to recompute it via expensive RLP encoding + Keccak256.

onbjerg (approved):

PR #10988, opened 2025-07-11
feat: fuzzer metrics

0xalpharush:
## Motivation
Better observability and benchmarking capabilities of fuzzer performance

closes https://github.com/foundry-rs/foundry/issues/10876

The number of failures don't really make sense until https://github.com/foundry-rs/foundry/issues/9727 is implemented

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #11902, opened 2025-10-01
fix(evm): correct favored threshold comparison

MozirDmitriy:
## Motivation

The selection criterion is inverted:
- Comment: “selected” if the proportion of new finds is above the threshold.
- Code: is_favored is set when the proportion is below the threshold (< FAVORABILITY_THRESHOLD).
- This reverses the meaning and distorts metrics.favored_items.

## Solution

- Fix incorrect favored comparison from < to > so only productive corpus entries (ratio > threshold) are marked as favored.
- Add unit tests to assert favored transitions (> threshold true, < threshold false, == threshold false with strict >) and eviction behavior (favored protected, non‑favored evicted).
- This prevents productive seeds from being evicted and aligns code with documented intent.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/evm/src/executors/corpus.rs
+++ b/crates/evm/evm/src/executors/corpus.rs
@@ -300,7 +300,7 @@ impl CorpusManager {
                     corpus.new_finds_produced += 1
                 }
                 let is_favored = (corpus.new_finds_produced as f64 / corpus.total_mutations as f64)
-                    < FAVORABILITY_THRESHOLD;
+                    > FAVORABILITY_THRESHOLD;
                 self.metrics.update_favored(is_favored, corpus.is_favored);
                 corpus.is_favored = is_favored;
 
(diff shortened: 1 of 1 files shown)


grandizzy (commented):
thank you! oh, I think you're on to something here, indeed if for example 40 new finds produced from 100 mutations that would be 0.4 while if only let's say 20 new finds from 100 mutations that will make 0.2 so indeed favored should be determined by greater than 0.3 
Will make some tests with this fix, meanwhile mind to fix the failing `favored_sets_false_and_metrics_decrement_when_ratio_lt_threshold` test you added?
CC @0xalpharush

0xalpharush:
This seems right. The refactoring I made here should have inverted it like is done in this PR. Thanks @MozirDmitriy 
https://github.com/foundry-rs/foundry/pull/10988/files#diff-6672478b3ff4d87907eb18ba56f9037395efbe25bb567b668d9b82a580d0b0c5L298-L299

grandizzy (approved):
lgtm, thank you!

PR #11903, opened 2025-10-01
Added foundryup\install.ps1 and foundryup\foundryup.ps1

tboy1337:
Added foundryup\install.ps1 and foundryup\foundryup.ps1 and updated the README.md with installation instructions for Windows and enhance usage examples for foundryup.

### DIFF
--- a/foundryup/README.md
+++ b/foundryup/README.md
@@ -6,11 +6,25 @@ Update or revert to a specific Foundry branch with ease.
 
 ## Installing
 
+### Linux / macOS / WSL
+
 ```sh
 curl -L https://foundry.paradigm.xyz | bash
 ```
 
+### Windows
+
+```powershell
+# Download and run install.ps1
+irm https://foundry.paradigm.xyz/install.ps1 | iex
+
+# Or if you have the repository locally:
+.\install.ps1
+```
+
 ## Usage
+
+### Linux / macOS / WSL
 
 To install the **nightly** version:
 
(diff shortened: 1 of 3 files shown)


onbjerg (changes requested):
We do not think maintaining a powershell version of foundryup makes sense, especially given bash is readily available on windows if you install git or use wsl.

tboy1337:
@onbjerg IMO it definitely makes sense, using inbuilt programs is always going to be better because you dont need to install or configure anything and less can go wrong. To use bash via WSL it requires you to install a distro and to use bash via git it requires you to have configured git correctly at install. You are basically making it as awkward as possible to install on Windows, maybe making it available via winget is the solution?

onbjerg:
Sorry, but this requires a team with 0 Windows installations to maintain about 1300 lines of code for Windows. We have rewriting the installer to be cross platform on the roadmap in the future, but we are not interested in maintaining a powershell port.

tboy1337:
@onbjerg how about 152 lines of batch instead? xD https://github.com/foundry-rs/foundry/pull/11918

tboy1337:
@onbjerg how about 433 lines of batch instead? xD https://github.com/foundry-rs/foundry/pull/11926

DaniPopes:
That doesn't change the fact that we don't want it

tboy1337:
@DaniPopes what? You don't want your project to actually be usable on windows, the operating system that the majority of people on the planet use? I have literally offered you a fully working solution on a plate that requires no maintenance and your too ignorant to even accept the PR and would rather your project look like a high school project on windows?

DaniPopes:
1. It already is usable on Windows with WSL / git bash, or downloading/installing the file manually
2. Nobody on the team uses Windows, any code in the repository needs maintenance, so this would be an extra burden for not much gain
3. We have no plan or motivation to maintain any more installation scripts when `foundryup` is on the roadmap to be rewritten as mentioned [here](https://github.com/foundry-rs/foundry/pull/11903#issuecomment-3356827564) and [here](https://github.com/foundry-rs/foundry/pull/11926#issuecomment-3361180596)

With this being said, and previous discussion in mind, we choose not to include this in our repository.

PR #11904, opened 2025-10-01
chore(gitignore): restrict bin ignore to repo root

Fallengirl:
Limit .gitignore to ignore only root bin (/bin) to avoid hiding nested */bin/ sources; consistent with existing patterns (e.g., npm/.gitignore); does not affect tracked files or CI.


onbjerg (changes requested):
this is not an issue

PR #11906, opened 2025-10-01
docs(Makefile): Precise nextest filters

Sharilleed223:
## Description

Refine nextest filters to avoid over-excluding tests:
- Previous filter: !test(/\b(issue|ext_integration)/) excluded any test name containing “issue”, not just repros. This risks silently skipping unrelated tests.
- Repros are generated under module path “repros::issue_…”, and ext integration tests are under the “ext_integration” binary.


onbjerg (changes requested):
the current filters match what we have in ci

PR #11917, opened 2025-10-01
feat: add Sourcify trace identifier

DaniPopes:
Rewrite crates/evm/traces/src/identifier/etherscan.rs to support multiple endpoints. Sourcify is enabled by default (if not offline).


onbjerg (approved):
lgtm, makes sense to me that we make this abstraction generic over an external fetcher and the impls of those look fine to me too

0xrusowsky (approved):

PR #11923, opened 2025-10-02
fix(anvil): correct evm_set_time units to seconds

Galoretka:
- Replace Duration::from_millis(offset) with Duration::from_secs(offset) since current_call_timestamp() and timestamp are in seconds; previous code divided by 1000, returning an incorrect value.
- Update EvmSetTime RPC docs to state the input is Unix seconds and the return value is seconds delta.
- This fixes incorrect time handling and aligns implementation, tests, and documentation.


mattsse (commented):
we want ms precision here

PR #11930, opened 2025-10-02
feat(forge): implement forge snapshot --format table stdout rendering

radik878:
- Add comfy-table output for gas snapshots when --format table is set
- Keep snapshot file format unchanged for compatibility with parsing/diff/check
- Reuse project table style (UTF8/Markdown via foundry_common::shell)

### DIFF
--- a/crates/forge/src/cmd/snapshot.rs
+++ b/crates/forge/src/cmd/snapshot.rs
@@ -2,8 +2,12 @@ use super::test;
 use crate::result::{SuiteTestResult, TestKindReport, TestOutcome};
 use alloy_primitives::{U256, map::HashMap};
 use clap::{Parser, ValueHint, builder::RangedU64ValueParser};
+use comfy_table::{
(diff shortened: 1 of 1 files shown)


crates/forge/src/cmd/snapshot.rs
```diff
@@ -111,13 +115,17 @@ impl GasSnapshotArgs {
...
+                sh_println!("\n{}", table)?;
+            }
             write_to_gas_snapshot_file(&tests, self.snap, self.format)?;
         }
         Ok(())
     }
 }
 
-// TODO implement pretty tables
+// Pretty table output is implemented via `--format table` and `build_gas_snapshot_table`.
```
onbjerg:
```suggestion
// Gas report format on stdout.
```

onbjerg (changes requested):
please provide example output in the pr as well

radik878:
> please provide example output in the pr as well

image

radik878:
> please provide example output in the pr as well

That's it, boss?

onbjerg (approved):

PR #11935, opened 2025-10-02
fix: cross-platform path construction in Chisel cache

maradini77:
**Description:**
Replace string concatenation with `PathBuf::join` for cache directory paths to ensure proper cross-platform compatibility.

**Changes:**
- Refactor `cache_dir()` to return `PathBuf` instead of `String`
- Use `PathBuf::join` for path construction in `write()`, `next_cached_session()`, and `load()`
- Maintain backward compatibility by converting to `String` only when needed

**Why:**
String concatenation with hardcoded `/` separators can cause path resolution issues on Windows and other platforms that use different path separators.

### DIFF
--- a/crates/chisel/src/session.rs
+++ b/crates/chisel/src/session.rs
@@ -80,7 +80,7 @@ impl ChiselSession {
 
         let cache_file_name = match self.id.as_ref() {
             Some(id) => {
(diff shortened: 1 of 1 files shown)


DaniPopes:
That doesn't have any effect in rust

PR #11939, opened 2025-10-03
refactor(anvil): introduce get_block_with_hash and convert_block_with_hash

silvekkk:
Introduces `get_block_with_hash` and `convert_block_with_hash` to avoid redundant hash computation.

When fetching blocks, we already have the hash (from parameters or storage lookups), but `convert_block` calls `hash_slow()` to recompute it. The new helpers pass the hash through instead, saving a lot cycles per query.


silvekkk:
commited the new refactor just now cc @DaniPopes

onbjerg (approved):

Issue #11933, opened 2025-10-02
bug(fmt): `multiline_func_header = "params_first"` seems to always break the params


0xrusowsky:
### Component

Forge

### Have you ensured that all of these are up to date?

- [ ] Foundry
- [ ] Foundryup

### Operating System

None

### Describe the bug

it seems like `multiline_func_header = "params_first"` always breaks the params (if there are more than 1)

repro:
```
---- expected: testdata/Aaaa/fmt.sol
++++ actual:   In-memory
   1    1 | // config: line_length = 100
   2    2 | // config: multiline_func_header = "params_first"
   3    3 | contract Test {
   4      -     function initialize(address, address, bytes memory) external pure override returns (address) {
        4 +     function initialize(
        5 +         address,
        6 +         address,
        7 +         bytes memory
        8 +     ) external pure override returns (address) {
   5    9 |         return address(0xdeadbeef);
   6   10 |     }
   7   11 | }
```

0xrusowsky:
this was a misunderstanding due to inconsistent variant naming.

as per the unit tests in https://github.com/foundry-rs/foundry/blob/master/crates/fmt/testdata/FunctionDefinition/params-first.fmt.sol#L6 `"params_first"` is in fact supposed to always break the fn parameters, regardless of the header length.

instead, the user should use `"params_first_multi"` so that the params only break if:
- +1 param
- fn header doesn't fit

see original discussion here:
- https://github.com/foundry-rs/foundry/pull/8762

PR #11944, opened 2025-10-03
chore(fmt): rename and alias params config varaint for clarity

0xrusowsky:
## Motivation

the naming for the the `multiline_func_header` config variant `"params_first"` is counter intuitive, as it is not aligned with the behavior of `"params_first_multi"`.

see how it created confusion for users, who opened this issue:
 - #11933 

## Solution

because of that, i renamed the config variant to `"params_always"` to better reflect that this styling option will ALWAYS break the fn header params.

the old name `"params_first"` is kept as an alias for backwards compatibility

## PR Checklist

- [ ] Added Tests
- [X] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,22 +1,26 @@
 .DS_STORE
 /target*
 /*.sol
+CLAUDE.md
+
+# Foundry artifacts
 out/
 snapshots/
-out.json
+
+# IDEs and extensions
 .idea
 .vscode
 .claude
-CLAUDE.md
+.zed
+
+# NPM
 .env
 node_modules
 dist
 bin
 _
 *.tgz
-.zed
 .vercel
 .vite
 .wrangler
-build
 *.zip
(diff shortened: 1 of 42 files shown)


0xrusowsky:
> makes sense, let's update README too [`master`/crates/fmt#configuration](https://github.com/foundry-rs/foundry/tree/master/crates/fmt#configuration) and dupe the test for `params_first` too to make sure alias does not get removed / introduce breaking change

i tested both locally to make sure. do we still want to duplicate the whole file just for that?

completely forgot about the readme, thanks for the reminder 🤝

grandizzy:
> > makes sense, let's update README too [`master`/crates/fmt#configuration](https://github.com/foundry-rs/foundry/tree/master/crates/fmt#configuration) and dupe the test for `params_first` too to make sure alias does not get removed / introduce breaking change
> 
> i tested both locally to make sure. do we still want to duplicate the whole file just for that?
> 
oh, not the whole file, but maybe just a test with inline config to make sure we don't remove the alias in future without ack the breaking change

grandizzy (approved):
lgtm

grandizzy (approved):

PR #11946, opened 2025-10-03
refactor(config): simplify fs permission retrieval logic

### LINKED ISSUE #11937: Refactor path permission checks
### Component

Other (please describe)

See this thread: https://github.com/foundry-rs/foundry/pull/10679#discussion_r2151410466

We currently have `find_permissions` and `find_all_permissions` - ideally, we just have one. The way this should work is:

- If you have write access, you also have read access
- If multiple permissions are defined for a given path, the highest privilege is granted
- We search for the permissions based on *longest path match*, that is, permissions for paths that are more concrete take precedence over permissions for broader paths

### Additional context

_No response_

mablr:
## Motivation

Close #11937 

## Solution
- merged `find_all_permissions` logic into `find_permission`
- Refactored `find_permission` to return the highest privilege permission based on the longest matching path.
- Updated tests

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes (I don't think so)

### DIFF
--- a/crates/config/src/fs_permissions.rs
+++ b/crates/config/src/fs_permissions.rs
@@ -37,7 +37,7 @@ impl FsPermissions {
     /// Caution: This should be called with normalized paths if the `allowed_paths` are also
     /// normalized.
     pub fn is_path_allowed(&self, path: &Path, kind: FsAccessKind) -> bool {
-        self.find_permission(path).map_or(false, |perm| perm.is_granted(kind))
+        self.find_permission(path).is_some_and(|perm| perm.is_granted(kind))
     }
 
     /// Returns the permission for the matching path.
@@ -176,7 +176,7 @@ impl fmt::Display for FsAccessKind {
 }
 
 /// Determines the status of file system access
-#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, PartialOrd)]
 pub enum FsAccessPermission {
     /// FS access is _not_ allowed
     #[default]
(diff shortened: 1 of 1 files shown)


crates/config/src/fs_permissions.rs
```diff
@@ -37,66 +37,54 @@ impl FsPermissions {
...
                     max_path_len = path_len;
-                    matching_permissions.clear();
-                    matching_permissions.push(perm.access);
+                    highest_permission = perm.access;
                 } else if path_len == max_path_len {
-                    matching_permissions.push(perm.access);
+                    // Same path length, keep the highest privilege
+                    if perm.access > highest_permission {
+                        highest_permission = perm.access;
+                    }
```
mablr, line 88: I think `PartialOrd` is ok to keep highest privilege.

We can also implement more advanced logic like:
```
highest_permission = match (highest_permission,  perm.access) {
    (FsAccessPermission::ReadWrite, _) |
        (FsAccessPermission::Read, FsAccessPermission::Write) |
        (FsAccessPermission::Write, FsAccessPermission::Read) => FsAccessPermission::ReadWrite,
    (FsAccessPermission::None, perm) => perm,
    (existing_perm, _) => existing_perm,
}
```
with this we don't need to consider `Write` variant as readable in `is_granted`.

grandizzy: yeah, I think better to not give read permission if only write permission is allowed - we have `ReadWrite` for that...

mablr: Cool, we are at the same page. I've just implemented first what the issue was suggesting.

I'll change that 👍

PR #11950, opened 2025-10-03
cast: remove `--` prefix from subcommand aliases to match CLI style

sashaodessa:
### Description
- Clean up visible aliases that looked like flags for subcommands.
- Updated in `crates/cast/src/opts.rs`: `format-bytes32-string`, `parse-bytes32-string`, `parse-bytes32-address`.
- Rationale: improve help/usage clarity and keep subcommand aliases consistent with CLI conventions.


onbjerg (changes requested):
this is a breaking change, it's not worth breaking cast and releasing a major version for this atm. thanks

Issue #11907, opened 2025-10-01
feat(`forge script`): add flag to disable HTTPS certificate validation for RPC


mirokuratczyk:
### Component

Forge

### Describe the feature you would like

https://github.com/foundry-rs/foundry/issues/10855, but for forge.

Currently, it is not possible to deploy contracts with `forge script` against an RPC endpoint or via an HTTP proxy that serves an untrusted certificate (e.g., self-signed, development, untrusted root CA, etc.).

I made an example change that will follow the foundry `eth_rpc_accept_invalid_certs` configuration across all requests https://github.com/mirokuratczyk/foundry/commit/80db4b02cdb3595869b3a2e44e035db712ffbe2d. But, I imagine that's antithetical to how that should be done in the project and an implementation like what was done for cast would be preferable https://github.com/foundry-rs/foundry/pull/10869.

I'm new to the project/rust and was having trouble threading insecure across `forge script` and the forked environment requests to make a proper implementation and thought it would be better to seek guidance first.

mablr:
I think we don't need to modify `RuntimeTransport` impl, or create a new config key.

Adding new `-k/--insecure` arg to `EvmArgs` with a "rename" is ok, like for :
https://github.com/foundry-rs/foundry/blob/bd0e4a729d28ff677c0aaa8ba8c1c409a2366faa/crates/cli/src/opts/evm.rs#L46-L48

straightforward actually 👍 . You might want to try to make a PR ?


It also brings back this older discussion about deduplicating args in `EvmArgs`&&`RpcOpts`: https://github.com/foundry-rs/foundry/issues/9303#issuecomment-2472901217

mirokuratczyk:
@mablr thanks for taking the time to respond to this issue.

I created a draft PR, but as you'll see there were many touch points to thread the new arg through https://github.com/foundry-rs/foundry/pull/11960.

I'm pretty sure that also won't follow the value of `eth_rpc_accept_invalid_certs` in the target foundry profile in `foundry.toml`, but figured might be enough to get the ball rolling.

edit: it does get `forge script --insecure/-k` working for me testing against a self-signed cert.

mablr:
@mirokuratczyk Your welcome!

I made a quick review, it works, but I think the refactor I mentioned is actually needed before continuing.

PR #11960, opened 2025-10-03
feat(forge): add flag to disable HTTPS certificate validation for RPC

mirokuratczyk:
## Motivation

Resolve https://github.com/foundry-rs/foundry/issues/11907#issuecomment-3358036876.

## Solution

- Add `--insecure/-k` flag to `forge script` which toggles ignoring invalid certs (e.g., self-signed)

I'm not sure this PR handles all possible codepaths as new providers are built in many different places and each needs to be configured with `.accept_invalid_certs(bool)`.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/common/src/provider/mod.rs
+++ b/crates/common/src/provider/mod.rs
@@ -70,15 +70,15 @@ pub type RetryProviderWithSigner<N = AnyNetwork> = FillProvider<
 /// ```
 #[inline]
 #[track_caller]
-pub fn get_http_provider(builder: impl AsRef<str>) -> RetryProvider {
-    try_get_http_provider(builder).unwrap()
+pub fn get_http_provider(builder: impl AsRef<str>, accept_invalid_certs: bool) -> RetryProvider {
+    try_get_http_provider(builder, accept_invalid_certs).unwrap()
 }
 
 /// Constructs a provider with a 100 millisecond interval poll if it's a localhost URL (most likely
 /// an anvil or other dev node) and with the default, or 7 second otherwise.
 #[inline]
-pub fn try_get_http_provider(builder: impl AsRef<str>) -> Result<RetryProvider> {
-    ProviderBuilder::new(builder.as_ref()).build()
+pub fn try_get_http_provider(builder: impl AsRef<str>, accept_invalid_certs: bool) -> Result<RetryProvider> {
+    ProviderBuilder::new(builder.as_ref()).accept_invalid_certs(accept_invalid_certs).build()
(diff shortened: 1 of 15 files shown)


crates/common/src/provider/mod.rs
```diff
@@ -70,15 +70,15 @@ pub type RetryProviderWithSigner<N = AnyNetwork> = FillProvider<
 /// ```
 #[inline]
 #[track_caller]
-pub fn get_http_provider(builder: impl AsRef<str>) -> RetryProvider {
-    try_get_http_provider(builder).unwrap()
+pub fn get_http_provider(builder: impl AsRef<str>, accept_invalid_certs: bool) -> RetryProvider {
```
mablr: This should not be there

crates/evm/core/src/backend/mod.rs
```diff
@@ -2044,7 +2044,7 @@ mod tests {
     async fn can_read_write_cache() {
         let Some(endpoint) = ENDPOINT else { return };
 
-        let provider = get_http_provider(endpoint);
+        let provider = get_http_provider(endpoint, false);
```
mablr, line 2047: same

mablr (commented):
Nice, but I guess it would be actually cleaner to build the provider using :

https://github.com/foundry-rs/foundry/blob/c646b8844612f287247bdd935513b6a9b7e71e82/crates/cli/src/utils/mod.rs#L102-L106

this would retrieve `eth_rpc_accept_invalid_certs` key from the config directly.

This actually confirms the need for a refactor of the `RpcOpts`/`EvmArgs` common fields. So I'll open a dedicated issue to get feedback from the project team.

crates/common/src/provider/mod.rs
mirokuratczyk: Fixed.

crates/evm/core/src/backend/mod.rs
mirokuratczyk: Fixed.

mirokuratczyk:
I pushed an update that uses `get_provider` as suggested. Seems to be working with `--insecure`.

Sounds like we'll wait for https://github.com/foundry-rs/foundry/pull/11964 to be merged first though and go from there.

mablr:
Hi @mirokuratczyk, thanks for your contribution!

Your PR was superseded by #14224 with a more systematic approach.

PR #11967, opened 2025-10-04
refactor(evm): improve cold path hint implementation

viktorking7:
## Motivation

The `cold_path()` function had conflicting attributes: `#[cold]` and `#[inline(always)]`. The `#[inline(always)]` attribute contradicts the purpose of `#[cold]` by forcing inlining of a function that should be marked as unlikely to be executed, which can hurt optimization.

## Solution

- Removed the conflicting `#[inline(always)]` attribute from `cold_path()`
- Added explicit `std::hint::cold()` call to make the cold path hint more explicit
- Updated the TODO comment to reflect the current implementation and future migration path to `std::hint::cold_path()` once stabilized


DaniPopes:
all of this is wrong

PR #11972, opened 2025-10-05
feat(anvil): add basic Beacon API support

### LINKED ISSUE #11797: feat: beacon API support for blob fetching

mablr:
## Motivation

Add basic Beacon REST API to close #11797

This is a follow-up of #11828

## Solution

- New Beacon response and error types.
- Updated the server router to handle Beacon REST API calls.
- Implemented tests for the Beacon API endpoint to validate blob sidecar retrieval.

I endeavored to design this inspiring as closely as possible to the existing JSON-RPC server implementation.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -58,9 +58,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy-chains"
-version = "0.2.13"
+version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fc7aacbb0ac0f76aaa64d1e1412f778c0574f241e4073b2a3e09c605884c9b90"
+checksum = "bf01dd83a1ca5e4807d0ca0223c9615e211ce5db0a9fd1443c2778cacf89b546"
 dependencies = [
  "alloy-primitives",
  "num_enum",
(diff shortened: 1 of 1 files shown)


crates/anvil/src/eth/beacon/response.rs
```diff
@@ -0,0 +1,75 @@
...
+    ///
+    /// For Anvil, this is always `false` since there's no real consensus layer
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub execution_optimistic: Option<bool>,
+    /// Whether the response references finalized history
+    ///
+    /// For Anvil, this is always `false` since there's no real consensus layer
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub finalized: Option<bool>,
+}
```
mattsse, line 32: we can maybe move this type to alloy beacon types?

crates/anvil/src/server/mod.rs
```diff
@@ -33,11 +34,30 @@ pub async fn serve_on(
...
+    let ws = PubSubEthRpcHandler::new(api.clone());
+
+    // JSON-RPC router
+    let rpc_router = anvil_server::http_ws_router(config, http, ws);
+
+    // Beacon REST API router
+    let beacon_router = beacon_router(api);
+
+    // Merge the routers
+    rpc_router.merge(beacon_router)
```
mattsse, line 50: this isn't invasive at all

mablr: Yep axum is beautiful 🤩

crates/anvil/src/eth/beacon/response.rs
mablr: Ok I’ll do it👍

PR #11977, opened 2025-10-05
feat(wallet): enhance keystore listing with address suffix

### LINKED ISSUE #11094: feat(`cast`): show addresses in the output of `cast wallet list`
### Component

Cast

Currently `cast wallet list` only shows the account names, such as:

```
$ cast wallet list
test1 (Local)
test2 (Local)
test3 (Local)
```

I want to see all the addresses.

### Additional context

_No response_

mablr:
## Motivation

Close #11094

## Solution

- Updated keystore file naming to include wallet address for better identification.
- Introduced `find_keystore_by_name` utility to support both old and new filename formats.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes (account naming is backward-compatible)

### DIFF
--- a/crates/cast/src/cmd/wallet/mod.rs
+++ b/crates/cast/src/cmd/wallet/mod.rs
@@ -13,7 +13,7 @@ use eyre::{Context, Result};
 use foundry_cli::{opts::RpcOpts, utils, utils::LoadConfig};
 use foundry_common::{fs, sh_println, shell};
 use foundry_config::Config;
-use foundry_wallets::{RawWalletOpts, WalletOpts, WalletSigner};
+use foundry_wallets::{RawWalletOpts, WalletOpts, WalletSigner, utils::find_keystore_by_name};
 use rand_08::thread_rng;
 use serde_json::json;
 use std::path::Path;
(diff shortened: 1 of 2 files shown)


crates/cast/src/cmd/wallet/mod.rs
```diff
@@ -337,25 +337,37 @@ impl WalletSubcommands {
...
+                            let old_path = path.join(identifier);
+                            let path_with_address = path.join(format!(
+                                "{}_{}",
+                                identifier,
+                                wallet.address().to_checksum(None)
+                            ));
+
+                            if old_path.exists() && old_path != path_with_address {
+                                std::fs::rename(&old_path, &path_with_address)?;
+                            }
```
mablr, line 352: The renaming of the keystore file is a bit dirty due to eth_keystore lib which directly creates the file before returning the wallet necessary to derive the address. 🫤

Ideally the lib should be modified, or upstreamed to foundry to solve this in an elegant way.

mablr:
Not relevant anymore as `foundry-wallets` was extracted to `foundry-core`.

PR #11980, opened 2025-10-06
fix(doc): reuse solar sema compiler

### LINKED ISSUE #11966: Last couple of weeks we have seen a litany of forge issues in our ci pipeline
### Component

Forge

- [x] Foundry
- [x] Foundryup

1.4.0-nightly

_No response_

forge lint, forge fmt --check, forge doc --build

### Operating System

Linux

### Describe the bug

3 bugs in last 2 weeks:
- `forge lint`: https://github.com/foundry-rs/foundry/issues/11668
- `forge fmt --check`: not using foundry.toml settings
- `forge doc --build`: 
```
thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow, aborting
```

These checks have been running for months without issues. Now they are failing from some recent foundry updates.

grandizzy:
## Motivation
- closes #11966 
- include only fn sig in doc (`func.loc_prototype` no body) + add `;` at the end of sig to match prev behavior
- remove extra indent of 4 spaces to fix format issue as

image
image

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/doc/src/builder.rs
+++ b/crates/doc/src/builder.rs
@@ -9,7 +9,6 @@ use foundry_config::{DocConfig, FormatterConfig, filter::expand_globs};
 use itertools::Itertools;
 use mdbook::MDBook;
 use rayon::prelude::*;
-use solar::interface::source_map::FileName;
 use std::{
     cmp::Ordering,
     fs,
@@ -136,13 +135,9 @@ impl DocBuilder {
                     let mut files = vec![];
 
                     // Read and parse source file
-                    if let Some(ast) = gcx.sources.iter().find(|source| {
-                        if let FileName::Real(src_path) = &source.file.name {
-                            return src_path.eq(path);
-                        }
-                        false
-                    }) && let Some(source) =
-                        forge_fmt::format_ast(gcx, ast, self.fmt.clone().into())
+                    if let Some((_, ast)) = gcx.get_ast_source(path)
+                        && let Some(source) =
+                            forge_fmt::format_ast(gcx, ast, self.fmt.clone().into())
                     {
                         let (mut source_unit, comments) = match solang_parser::parse(&source, i) {
                             Ok(res) => res,
--- a/crates/doc/src/parser/item.rs
+++ b/crates/doc/src/parser/item.rs
@@ -91,13 +91,7 @@ impl ParseItem {
         // Remove extra indent from source lines.
         self.code = code
             .lines()
-            .map(|line| {
-                if line.chars().take_while(|c| c.is_whitespace()).count() > 4 {
-                    line.strip_prefix("    ").unwrap_or(line)
-                } else {
-                    line
-                }
-            })
+            .map(|line| line.strip_prefix("    ").unwrap_or(line))
             .collect::<Vec<_>>()
             .join("\n");
         self


crates/doc/src/parser/item.rs
```diff
@@ -81,7 +81,20 @@ impl ParseItem {
...
+        if let ParseSource::Function(_) = self.source {
+            code.push(';');
+        }
+
+        // Remove extra indent from docs.
+        self.code = code
+            .lines()
+            .map(|line| line.strip_prefix("    ").unwrap_or(line))
+            .collect::<Vec<_>>()
+            .join("\n");
```
0xrusowsky: i think that we should make this more robust.

what i'd like to see is the following logic:
1. get the source code from beginning of line to `self.source.range().start`
2. strip the indentation of this source code (should be an empty string or just indentation) from all lines


```solidity
function topLevel() public {
	// this needs 1 indentation level, none should be removed
}

contract TopLevel {
	function innerLevel() public {
		// this has 2 ind, 1 needs to be removed
	}
}
```

grandizzy: I changed the logic to remove only if starts with > 4 whitespaces (that is to proper format the free functions as in prev version), since we're going to remove solang anyways not sure if should spend more time on this, wdyt?

0xrusowsky: i think it is fine as a temp workaround as we need to drop the solang dep anyways.
once we do that, we can do it properly

0xrusowsky (approved):

0xrusowsky:
let's wait for @DaniPopes to have a look before we merge

crates/doc/src/builder.rs
```diff
@@ -124,135 +125,153 @@ impl DocBuilder {
...
+            let documents = combined_sources
+                .par_iter()
+                .enumerate()
+                .map(|(i, (path, from_library))| {
+                    let path = *path;
+                    let from_library = *from_library;
+                    let mut files = vec![];
+
+                    // Read and parse source file
+                    if let Some(ast) = gcx.sources.iter().find(|source| {
```
DaniPopes: gcx.get_ast_source

crates/doc/src/parser/item.rs
```diff
@@ -81,7 +81,25 @@ impl ParseItem {
...
+        // Special function case, add `;` at the end of definition.
+        if let ParseSource::Function(_) = self.source {
+            code.push(';');
+        }
+
+        // Remove extra indent from source lines.
+        self.code = code
+            .lines()
+            .map(|line| {
+                if line.chars().take_while(|c| c.is_whitespace()).count() > 4 {
```
DaniPopes: I think you can just trim_start always

0xrusowsky: this is what he did initially, but that's inaccurate.

see the reasoning here: https://github.com/foundry-rs/foundry/pull/11980#discussion_r2405211333

grandizzy: indeed, however I just realized the count check should have been >= 4 so it didn't apply format properly, hence I am in favour of `trim_start` always, this affects only free functions which will be displayed as below but I don't think it's a biggie

image

grandizzy: https://github.com/foundry-rs/foundry/pull/11980/commits/e7aa1627e873b1e3c4ce42ed00f8a41bda1aaba3

crates/doc/src/builder.rs
grandizzy: https://github.com/foundry-rs/foundry/pull/11980/commits/e7aa1627e873b1e3c4ce42ed00f8a41bda1aaba3

DaniPopes (approved):

PR #11982, opened 2025-10-06
feat(lint): emit suggestions for `named_struct_fields` lint

### LINKED ISSUE #11760: feat: lint to enforce declaring struct fields

mablr:
## Motivation
Follow-up of #11892 to close #11760 (last part about fix suggestion https://github.com/foundry-rs/foundry/issues/11760#issuecomment-3333640581)

## Solution

- Updated the `named_struct_fields` lint to emit suggestion for initializing structs with named fields.
- Updated test data to reflect the new suggestion format in lint messages.

## PR Checklist

- [x] Added Tests (updated existing one)
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/lint/src/sol/info/named_struct_fields.rs
+++ b/crates/lint/src/sol/info/named_struct_fields.rs
@@ -50,7 +50,9 @@ impl<'hir> LateLintPass<'hir> for NamedStructFields {
         for (field_id, arg) in fields.iter().zip(args.iter()) {
             let field = hir.variable(*field_id);
 
-            let Some((field_name, arg_snippet)) = ctx.span_to_snippet(arg.span).zip(field.name.map(|n| n.to_string()) else {
+            let Some((arg_snippet, field_name)) =
+                ctx.span_to_snippet(arg.span).zip(field.name.map(|n| n.to_string()))
+            else {
                 // Emit without suggestion if we can't get argument snippet
                 ctx.emit(&NAMED_STRUCT_FIELDS, expr.span);
                 return;


onbjerg (changes requested):
let's for completeness sake add an example in the test where the number of arguments does not match the struct definition. lint does parsing and lowering, but does not do typeck, so this should be fine.

crates/lint/src/sol/info/named_struct_fields.rs
```diff
@@ -16,16 +16,53 @@ impl<'hir> LateLintPass<'hir> for NamedStructFields {
...
+        hir: &'hir solar::sema::hir::Hir<'hir>,
         expr: &'hir solar::sema::hir::Expr<'hir>,
     ) {
-        if let ExprKind::Call(
-            Expr { kind: ExprKind::Ident([Res::Item(ItemId::Struct(_struct_id))]), .. },
-            CallArgs { kind: CallArgsKind::Unnamed(_args), .. },
+        let ExprKind::Call(
+            callee_expr @ Expr {
+                kind: ExprKind::Ident([Res::Item(ItemId::Struct(struct_id))]), ..
+            },
```
onbjerg: instead of this bind, we can just ask for `span` as well


```suggestion
            Expr {
                kind: ExprKind::Ident([Res::Item(ItemId::Struct(struct_id))]), span, ..
            },
```

crates/lint/src/sol/info/named_struct_fields.rs
```diff
@@ -16,16 +16,53 @@ impl<'hir> LateLintPass<'hir> for NamedStructFields {
...
+
+        let strukt = hir.strukt(*struct_id);
+        let fields = &strukt.fields;
+
+        // Basic sanity conditions for a consistent auto-fix
+        if fields.len() != args.len() || fields.is_empty() {
+            // Emit without suggestion
             ctx.emit(&NAMED_STRUCT_FIELDS, expr.span);
+            return;
+        }
```
0xrusowsky, line 39: i believe this is redundant/unnecessary as otherwise solar would have complained when lowering, but feel free to add a unit test to confirm.

if i'm wrong feel free to leave it as is

crates/lint/src/sol/info/named_struct_fields.rs
```diff
@@ -16,16 +16,53 @@ impl<'hir> LateLintPass<'hir> for NamedStructFields {
...
+            let field_name = field.name.map(|n| n.to_string()).unwrap_or_else(|| "?".to_string());
+
+            let arg_snippet =
+                ctx.span_to_snippet(arg.span).unwrap_or_else(|| "/* unknown */".to_string());
+
+            field_assignments.push(format!("{field_name}: {arg_snippet}"));
         }
+
+        let struct_name =
+            ctx.span_to_snippet(callee_expr.span).unwrap_or_else(|| "StructName".to_string());
```
0xrusowsky: i'd rather emit the lint warning without the suggestion when we are unable to resolve any of these

0xrusowsky (changes requested):

crates/lint/src/sol/info/named_struct_fields.rs
onbjerg: this is not correct, solar does not do typeck at present, so no error is produced:

```
echo 'contract Foo { struct A { uint256 x; uint256 y; } function f() public { A memory baz = A(1); } }' | solar -
```

additionally, we only lower, we don't perform analysis, so even with typeck implemented, we would not emit these errors

crates/lint/src/sol/info/named_struct_fields.rs
mablr: Ok no problem

crates/lint/src/sol/info/named_struct_fields.rs
mablr: yess, that's cleaner

crates/lint/src/sol/info/named_struct_fields.rs
```diff
@@ -16,16 +16,58 @@ impl<'hir> LateLintPass<'hir> for NamedStructFields {
...
+        let mut field_assignments = Vec::new();
+        for (field_id, arg) in fields.iter().zip(args.iter()) {
+            let field = hir.variable(*field_id);
+            let field_name = field.name.map(|n| n.to_string()).unwrap_or_else(|| "?".to_string());
+
+            let Some(arg_snippet) = ctx.span_to_snippet(arg.span) else {
+                // Emit without suggestion if we can't get argument snippet
+                ctx.emit(&NAMED_STRUCT_FIELDS, expr.span);
+                return;
+            };
```
onbjerg, line 59:
```suggestion
            let field = hir.variable(*field_id);

            let Some((arg_snippet, field_name)) = ctx.span_to_snippet(arg.span).zip(field.name) else {
                // Emit without suggestion if we can't get argument snippet
                ctx.emit(&NAMED_STRUCT_FIELDS, expr.span);
                return;
            };
```

if the field name is undefined we want the fix-less diagnostic as well. we can use [`Option::zip`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.zip)

crates/lint/src/sol/info/named_struct_fields.rs
mablr: @onbjerg When adding your snippet to forge ui testdata I encounter this compiler error:

```
Error: Compiler run failed:
Error (9755): Wrong argument count for struct constructor: 1 arguments given but expected 2.
  --> onbjergSolarExampleNoTypeCheck.sol:LL:CC:
   |
10 |         A memory baz = A(1);
   |                        ^^^^

```

and the "ui" test target fails unless setting `ProjectCompiler` with `.bail(false)`.

I'm unsure at which level this error is triggered, but in the current setup I'm unable to reach this if-branch in the lint with a test case having struct with incorrect amount of args.

onbjerg: ahhh, ok, this is because we also run solc. this might not be the case in the future, so we can't rely on it for lint correctness. just skip adding the unit test I suggested.

onbjerg (approved):

0xrusowsky (approved):

PR #11984, opened 2025-10-06
fix(common): preserve contract type in try new by wrapping deployCode

phrwlk:
The dynamic test linking preprocessor failed to preserve the contract type for try-constructor statements. For code like try new C(...) returns (C c) { ... }, the transformation produced vm.deployCode(...) without wrapping, leaving the expression typed as address. This mismatched the expected type C and could fail compilation when a returns (C) clause was present.

This change unifies handling for both try and non-try constructor creations by always wrapping the transformed expression as C(payable(vm.deployCode({...}))), ensuring the resulting type remains the contract type. Added a CLI test that asserts try new C(...) returns (C c) compiles after preprocessing.

### DIFF
--- a/crates/common/src/preprocessor/deps.rs
+++ b/crates/common/src/preprocessor/deps.rs
@@ -326,7 +326,7 @@ pub(crate) fn remove_bytecode_dependencies(
                     call_args_offset,
                     value,
                     salt,
-                    try_stmt,
+                    try_stmt: _,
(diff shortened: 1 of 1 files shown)


grandizzy:
thank you! that's a valid bug but needs a little bit more complex fix, will include in https://github.com/foundry-rs/foundry/pull/12050 with the test you added

PR #11995, opened 2025-10-06
chore(test): add hints after test failed

silvekkk:
During optimized #11993, I just realized there are much more magic function to help the test, but for the lazy people who don't like to read the whole docs like me, its really hard to know this, we just wait for all the test pass 😓. These hints may give alot help.

### DIFF
--- a/crates/forge/src/result.rs
+++ b/crates/forge/src/result.rs
@@ -192,6 +192,14 @@ impl TestOutcome {
             successes.to_string().green()
         )?;
 
+        // Show helpful next steps for failed tests
+        sh_println!("\n💡 Next steps:")?;
+        sh_println!("  • Rerun only failed tests: forge test --rerun")?;
+        sh_println!("  • Debug specific test: forge test --debug --match-test <test_name>")?;
+        sh_println!("  • Watch mode with smart rerun: forge test --watch --rerun-failed")?;
+        sh_println!("  • View detailed traces: forge test -vvv --match-test <test_name>")?;
+        sh_println!("  • Run specific test: forge test --match-test <test_name>")?;
+
         // TODO: Avoid process::exit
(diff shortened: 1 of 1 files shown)


silvekkk:
> i think it's ok to highlight e.g. `forge test --rerun` but I don't think we want a 5 bullet point list when tests fail

Thanks!  only `--rerun` is more clean.

new version ⬇️
```
Encountered a total of 3 failing tests, 77 tests succeeded

💡 Tip: Run `forge test --rerun` to retry only the 3 failed tests
```

crates/forge/src/result.rs
```diff
@@ -192,6 +192,15 @@ impl TestOutcome {
             successes.to_string().green()
         )?;
 
+        // Show helpful hint for rerunning failed tests
+        let test_word = if failures == 1 { "test" } else { "tests" };
+        sh_println!(
+            "\n💡 Tip: Run {} to retry only the {} failed {}",
```
onbjerg:
```suggestion
            "\nTip: Run {} to retry only the {} failed {}",
```

onbjerg (changes requested):
please also address the test failures:)

silvekkk:
> please also address the test failures:)

Hi @onbjerg, I've updated all the test snapshots to include the tip message. Could you please help re-trigger the CI? Latest run was 1635/1636 - flaky tests passed on retry, the one failure seems to be an unrelated cast issue, but let me know if I missed something.

silvekkk:
combine two similar commits for CI trigger, finally passed!!!! should we fix this ci problem? Its kinda annoying TBH, but last time [this](https://github.com/foundry-rs/foundry/pull/11942#event-20090964410) just closed, cc @onbjerg

PR #12005, opened 2025-10-07
fix(forge): dynamic test linking should follow pranks

### LINKED ISSUE #11978: bug(forge test/config): dynamic test linking breaks test suites that call `vm.prank` before constructor calls
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.3.5-stable

foundryup: 1.3.0

forge test

### Operating System

None

### Describe the bug

# Summary

`vm.prank` can be used to simulate different deployer addresses via the `new` keyword.

Dynamic test linking substitutes `new` contract deploys with the `vm.deployCode` cheatcode. 

Since `prank` normally applies to the next external call (or contract creation), and might be combined with other cheatcodes like `vm.expectEmit` or `vm.expectRevert`, Forge ignores pranks when calling cheatcodes, including `vm.deployCode`. This leads to test failures failures and/or unexpected behavior, since the prank is not applied normally (and may be left "dangling" before `vm.prank` is called again, which results in a cheatcode error)

If this is expected behavior, it should be considered a breaking change.

# Possible fixes

- Special handling for `vm.deployCode` so it is treated like a normal external call, and does not ignore `vm.prank`  status
- add `vm.deployCode` overloads that take in a `msgSender` param and  update dynamic test linking preprocessing to delete the `vm.prank` call and fill in the appropriate `vm.deployCode` with the pranker

# Minimal repro
`Counter.sol`
```solidity
pragma solidity ^0.8.13;

contract Counter {
    uint256 public number;
    address public deployer;

    constructor() {
        deployer = msg.sender;
    }

        number = newNumber;
    }

    function increment() public {
        number++;
    }
}
```
`Counter.t.sol`
```solidity
pragma solidity ^0.8.13;

import {Counter} from "../src/Counter.sol";

contract CounterTest is Test {
    Counter public counter;

    function setUp() public {
        counter = new Counter();
        counter.setNumber(0);
    }

    function test_Increment() public {
        counter.increment();
        assertEq(counter.number(), 1);
    }

    function testFuzz_SetNumber(uint256 x) public {
        counter.setNumber(x);
        assertEq(counter.number(), x);
    }

    function test_Deployer(address addr) public {
        vm.prank(addr);
        Counter _counter = new Counter{salt: 0}();
        assertEq(_counter.deployer(), addr);
    }
}
```

`foundry.toml`
```toml
[profile.default]
src = "src"
out = "out"
libs = ["lib"]
# uncomment to reproduce failure
# dynamic_test_linking = true
```

grandizzy:
## Motivation
- fixes #11978
- use pranked address for inputs

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cheatcodes/src/fs.rs
+++ b/crates/cheatcodes/src/fs.rs
@@ -12,7 +12,10 @@ use dialoguer::{Input, Password};
 use forge_script_sequence::{BroadcastReader, TransactionWithMetadata};
 use foundry_common::fs;
 use foundry_config::fs_permissions::FsAccessKind;
-use revm::{context::CreateScheme, interpreter::CreateInputs};
+use revm::{
+    context::{CreateScheme, JournalTr},
+    interpreter::CreateInputs,
+};
 use revm_inspectors::tracing::types::CallKind;
 use semver::Version;
 use std::{
@@ -372,9 +375,16 @@ fn deploy_code(
     let scheme =
         if let Some(salt) = salt { CreateScheme::Create2 { salt } } else { CreateScheme::Create };
 
+    // If prank active at current depth, then use it as caller for create input.
+    let caller = ccx
+        .state
+        .get_prank(ccx.ecx.journaled_state.depth())
+        .map(|prank| prank.new_caller)
+        .unwrap_or(ccx.caller);
+
     let outcome = executor.exec_create(
         CreateInputs {
-            caller: ccx.caller,
+            caller,
             scheme,
             value: value.unwrap_or(U256::ZERO),
             init_code: bytecode.into(),
--- a/crates/forge/tests/cli/test_optimizer.rs
+++ b/crates/forge/tests/cli/test_optimizer.rs
@@ -1535,3 +1535,56 @@ Compiling 1 files with [..]
 
 "#]]);
 });
+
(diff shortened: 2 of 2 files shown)


0xrusowsky (dismissed):

crates/cheatcodes/src/fs.rs
```diff
@@ -372,9 +375,16 @@ fn deploy_code(
     let scheme =
         if let Some(salt) = salt { CreateScheme::Create2 { salt } } else { CreateScheme::Create };
 
+    // If prank active at current depth, then use it as caller for create input.
+    let caller = ccx
+        .state
+        .get_prank(ccx.ecx.journaled_state.depth())
+        .map(|prank| prank.new_caller)
+        .unwrap_or(ccx.caller);
```
0xrusowsky: fyi, u could also use `.map_or(ccx.caller, |prank| prank.new_caller)` to do it in one line

grandizzy: yep,  https://github.com/foundry-rs/foundry/pull/12005/commits/b35d0012830229e95dd2b21b49a6b549dcf30013 one line less :+1:

PR #12011, opened 2025-10-07
fix: improve error handling in available_senders with logging and deduplication

CreeptoGengar:
Improve error handling in WalletSigner::available_senders method

This commit addresses the issue where errors when retrieving addresses from hardware wallets (Ledger/Trezor) were silently ignored, making it difficult for users to understand why some addresses might be unavailable.

Why these changes are necessary:
1. Silent error ignoring made debugging hardware wallet issues nearly impossible
2. Users had no visibility into why certain addresses were missing from results
3. Potential duplicate addresses could confuse users when both LedgerLive and
   Legacy derivation paths returned the same address
4. The graceful degradation pattern is preserved while adding observability


crates/wallets/src/wallet_signer.rs
```diff
@@ -129,44 +131,73 @@ impl WalletSigner {
     /// - the result for Ledger signers includes addresses available for both LedgerLive and Legacy
     ///   derivation paths
     /// - for Local and AWS signers the result contains a single address
+    /// - errors when retrieving addresses are logged but do not prevent returning available addresses
     pub async fn available_senders(&self, max: usize) -> Result<Vec<Address>> {
         let mut senders = Vec::new();
+        let mut seen_addresses = HashSet::new();
```
0xrusowsky: rather than allocating to both every time, drop `seen_addresses` and turn `senders` into a `HashSet`. At the very end of the function convert it into a vector.

0xrusowsky (changes requested):
please address the feedback and ensure the CI is green by running rustfmt

CreeptoGengar:
@0xrusowsky thanks for the feedback, I fixed it and also fixed rustfmt, 

but there're other CI's failing, do they related to mine PR? for me it seems they aren't

grandizzy:
> @0xrusowsky thanks for the feedback, I fixed it and also fixed rustfmt,
> 
> but there're other CI's failing, do they related to mine PR? for me it seems they aren't

Please merge in master to fix those

Issue #9840, opened 2025-02-07
Chain dependent solc config / Multichain scripts


sakulstra:
### Component

Forge

### Describe the feature you would like

At bgd we rely quite a lot on multichain scripts, which worked fine as long as Aave was only deployed on shanghai compatible blockchains.
With the recent expansion to linea(london) and zksync(zksolc) we're reaching a point where our workarounds no longer work.

Deploying for different chains, we try to solve with `profiles`.
```
[profile.bnb]
evm_version = 'shanghai'

[profile.linea]
evm_version = 'london'
```

One issue we face here is that whenever switching the evm version a complete recompile is needed. It would be great if the cache was per evm_version, so things only have to be compiled once.

---

While the above issue is annoying, one more problematic thing is that multichain script (single script switching between networks), will always run with a single solc & evm_version. This leads to problematic situations:
- if you select `london` (with the goal to unite on the lowest accross chains), the script might fail with `not activated` when touching sth deployed on shanghai or cancun
- if you select `cancun` you will end up deploying non executable code on chains that are stuck on london
- another caveat is that when you rely on address prediction based on create2, addresses might be different then expected

Currently I don't see any solution for this problem, the workaround we currently use is to no longer use multichain scripts... which ofc is painful.
It would be great if one could configure solc per chainId and dynamically switch based on which network is currently forked. No idea how this could work, but I guess if using forge multichain this would be a must.

sakulstra:
@grandizzy I think no.

This only really helps when I want to have one evm_version for the full script, right?
But the point of multichain scripts is to do things on multiple chains in the same script.

E.g. in [this example](https://github.com/bgd-labs/aave-proposals-v3/blob/main/src/20250120_Multi_FebruaryFundingUpdate/FebruaryFundingUpdate_20250120.s.sol#L115) i'd like to switch the evm_version on every `build<Network>Payload` (as what we do under the hood is switch to that network and check if the payload was deployed via create2). <- but e.g. this will will now no longer work with linea. If i deployed the payload with london, but execute the script via shanghai the addresses will be different. If I run the script with london, things on other networks might not work.

grandizzy:
[...]

PR #12014, opened 2025-10-08
feat(cheatcodes): add setEvmVersion / getEvmVersion

grandizzy:
## Motivation
- ref #9840 should be accompanied by a forge-std pr for `Config`
- adds cheatcodes to set / get evm execution spec in tests / scripts
```Solidity
    /// Returns the execution evm version.
    #[cheatcode(group = Evm, safety = Safe)]
    function getEvmVersion() external pure returns (string memory evm);

    /// Set the execution evm version.
    #[cheatcode(group = Evm, safety = Safe)]
    function setEvmVersion(string calldata evm) external;
```

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cheatcodes/spec/src/vm.rs
+++ b/crates/cheatcodes/spec/src/vm.rs
@@ -615,7 +615,7 @@ interface Vm {
     #[cheatcode(group = Evm, safety = Safe)]
     function getEvmVersion() external pure returns (string memory evm);
 
-    /// Set the test or script execution evm version.
+    /// Set the exact test or script execution evm version, e.g. `berlin`, `cancun`.
     ///
     /// **Note:** The execution evm version is not the same as the compilation one.
     #[cheatcode(group = Evm, safety = Safe)]
--- a/crates/cheatcodes/assets/cheatcodes.json
+++ b/crates/cheatcodes/assets/cheatcodes.json
@@ -9813,7 +9813,7 @@
     {
       "func": {
         "id": "setEvmVersion",
-        "description": "Set the test or script execution evm version.\n**Note:** The execution evm version is not the same as the compilation one.",
+        "description": "Set the exact test or script execution evm version, e.g. `berlin`, `cancun`.\n**Note:** The execution evm version is not the same as the compilation one.",
         "declaration": "function setEvmVersion(string calldata evm) external;",
         "visibility": "external",
         "mutability": "",


0xrusowsky (approved):

grandizzy:
> unsure if this solves the original ticket entirely, given some evm versions also imply changes in solc and therefore the resulting artifacts. for example, earlier evm versions do not support tstore/tload. wdyt?

yep, good point, I updated comment to prevent closing the issue.  For the 2nd part I think we could extend the additional compilation profiles as in https://github.com/foundry-rs/foundry/issues/9840#issuecomment-3381060802

crates/cheatcodes/spec/src/vm.rs
```diff
@@ -609,6 +609,14 @@ interface Vm {
     #[cheatcode(group = Evm, safety = Unsafe)]
     function coolSlot(address target, bytes32 slot) external;
 
+    /// Returns the execution evm version.
```
DaniPopes: actually can we add a note on the cheatcodes docs too? that runtime/execution evm version is not the same as compilation one

DaniPopes (changes requested):

DaniPopes:
string also is not the best because you don't encode the hierarchy that we have with integer/enum, eg that cancun enables paris, london etc

crates/cheatcodes/spec/src/vm.rs
grandizzy: :+1: updated in https://github.com/foundry-rs/foundry/pull/12014/commits/0838c5421b44208623f60558227e380bdf601e88

grandizzy:
> string also is not the best because you don't encode the hierarchy that we have with integer/enum, eg that cancun enables paris, london etc

can you pls detail, I thought these are OK to return from `getEvmVersion` call https://github.com/bluealloy/revm/blob/1bf799dd51648d3c8114b1a7496133ed8bc6603f/crates/primitives/src/hardfork.rs#L98

For setting it I reused the same as config does to be consistent https://github.com/foundry-rs/foundry/blob/dae1a1224a830a737d069b05da7bf98563176427/crates/config/src/utils.rs#L286

grandizzy:
> string also is not the best because you don't encode the hierarchy that we have with integer/enum, eg that cancun enables paris, london etc

updated note to reflect that exact evm version is needed, same as in config toml https://github.com/foundry-rs/foundry/pull/12014/commits/26cc8537b52139d1600a1853624f3416b47705ae

Issue #6623, opened 2023-12-19
Invalid Enum value when fuzzing


ScreamingHawk:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### What version of Foundry are you on?

forge 0.2.0 (73fb616 2023-12-19T00:16:21.131413571Z)

### What command(s) is the bug in?

forge test

### Operating System

Windows

### Describe the bug

When fuzzing an array of enums, forge supplies a value out of range. 

Example:

```sol
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.18;

import "forge-std/Test.sol";

contract ForgeFail is Test {

    enum EnumVal {
        VAL_0,
        VAL_1,
        VAL_2
    }

    function testForgeFail(EnumVal[] memory vals) external {
      // No content. Still fails
    }
}
```

Output:

```sh
Running 1 test for test/ForgeFail.t.sol:ForgeFail
[FAIL. Reason: EvmError: Revert; counterexample: calldata=0x98a2e150000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000003 args=[[3]]] testForgeFail(uint8[]) (runs: 2, μ: 647, ~: 647)
Traces:
  [643] ForgeFail::testForgeFail([3])
    └─ ← EvmError: Revert

Test result: FAILED. 0 passed; 1 failed; 0 skipped; finished in 5.63ms

Ran 1 test suites: 0 tests passed, 1 failed, 0 skipped (1 total tests)

Failing tests:
Encountered 1 failing test in test/sample/endorser/ForgeFail.t.sol:ForgeFail
[FAIL. Reason: EvmError: Revert; counterexample: calldata=0x98a2e150000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000003 args=[[3]]] testForgeFail(uint8[]) (runs: 2, μ: 647, ~: 647)

Encountered a total of 1 failing tests, 0 tests succeeded
```

The argument `3` is outside the bounds of `EnumVal`.

Note: I am running forge on WSL.

mds1:
[...]

PR #12015, opened 2025-10-08
feat(fuzz): ast-seeded dictionary

### LINKED ISSUE #10233: Using AST to seed the fuzzer dictionary
### Component

Forge

Currently, the fuzzer fails to break this very simple test case:

```solidity
pragma solidity ^0.8.13;

import "forge-std/Test.sol";

contract Hash is Test {
    function testHash(bytes memory v) external {
        assert(keccak256(v) != keccak256("hello"));
    }
}
```

My understanding is the dictionary is seeded using PUSH values from bytecode, therefore, it will not find any constants that are "optimized away" by the compiler. For the example above, `keccak256("hello")` becomes a `PUSH32` with its `keccak256` hash in bytecode, as shown below:

```yul
PUSH32 1c8aff950685c2ed4bc3174f3472287b56d9517b9c948127319a09a7a36deac8
```

As a result, the dictionary will never have "hello".

From what I know, Echidna/Medusa seed their dictionary using outputs from slither, so they do not struggle with this case. It may be beneficial for Foundry to perform some form of AST analysis before fuzzing (in this case, it would just be collecting all literals and constants).

### Additional context

_No response_

0xrusowsky:
## Motivation

closes #10233 

## Solution

1. leverage `solar::sema::Compiler` to collect all relevant AST literals found in the sources (excluding libs and scripts) and seed the `FuzzerDictionary` with them at initialization.
2. modify the strategies to source from a new pool of values (AST literals), when available
3. modify the string strategy to not always generate random strings, but also source from the string literals pool

## Future improvements

- explore literal value folding as suggested in: https://github.com/foundry-rs/foundry/pull/12015#issuecomment-3385828190. Tracked on a separate issue: https://github.com/foundry-rs/foundry/issues/12044
- use HIR rather than AST to figure out the type of the numeric literals, so that we can minimize dict size

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [ ] Breaking changes


crates/cli/src/opts/build/utils.rs
```diff
@@ -182,7 +150,6 @@ pub fn configure_pcx_from_compile_output(
     );
 
     configure_pcx_from_solc(pcx, &config.project_paths(), &solc, true);
-
     Ok(())
 }
```
0xrusowsky, line 154: waiting on https://github.com/foundry-rs/foundry/pull/12019 to be merged. should be reviewd there rather than here, but was necessary in orther for the fuzzer unit tests to pass

crates/evm/fuzz/src/strategies/param.rs
```diff
@@ -116,20 +116,34 @@ pub fn fuzz_param_from_state(
     param: &DynSolType,
     state: &EvmFuzzState,
 ) -> BoxedStrategy<DynSolValue> {
-    // Value strategy that uses the state.
+    // Value strategy that uses the state with AST literal support with the following allocations:
+    // - default: 20% AST literals, 40% DB state, 40% runtime samples (when AST available)
```
0xrusowsky: this is totally arbitrary, and i'm open to change it for different weights.

with that being said, i want to do a follow-up PR to have dynamic weights based on the pool sizes

crates/evm/fuzz/src/strategies/param.rs
```diff
@@ -170,13 +184,32 @@ pub fn fuzz_param_from_state(
...
-                )
-            })
-            .boxed(),
+        DynSolType::String => {
+            let state_clone = state.clone();
+            (any::<prop::sample::Index>(), any::<prop::sample::Index>())
+                .prop_flat_map(move |(use_ast_index, select_index)| {
+                    let dict = state_clone.dictionary_read();
+
+                    // AST string literals available: use 30/70 allocation
```
0xrusowsky: arbitrary value, we could change it if you are opinionated.

crates/evm/fuzz/src/strategies/state.rs
```diff
@@ -52,6 +53,13 @@ impl EvmFuzzState {
         // Create fuzz dictionary and insert values from db state.
         let mut dictionary = FuzzDictionary::new(config);
         dictionary.insert_db_values(accs);
+
+        // Seed dict with AST literals if analysis is available.
+        if let Some(literals) = analysis {
+            dictionary.ast_values = Some(literals);
```
grandizzy: IMO better to keep this simpler and just insert / reuse the existing fuzz dict samples - `insert_sample_values` which stores values by type and use them during fuzz runs instead having new strategy / weights and `ast_values`. We probably need to make the samples limit configurable and bump the default value 
https://github.com/foundry-rs/foundry/blob/020d515e21feaf31ec9e665546c94123a7cb1908/crates/evm/fuzz/src/strategies/state.rs#L374-L377

0xalpharush:
Not blocking but I would recommend implementing constant folding to some degree i.e. evaluate `2 * 2 ether`, evaluate `bytes32 IMPLEMENTATION_SLOT = bytes32(uint256(keccak256('eip1967.proxy.implementation')) - 1);`, evaluate `uint(-2)`. Arguably, solar will need this and the code could live there

See https://github.com/crytic/echidna/issues/636

grandizzy:
> Not blocking but I would recommend implementing constant folding to some degree i.e. evaluate `2 * 2 ether`, perform the keccack hash of a string, evaluate `uint(-2)`.
> 
> See [crytic/echidna#636](https://github.com/crytic/echidna/issues/636)

thanks! One thing here - this means we should collect from tests too which we don't do in PR, is this correct?

0xalpharush:
AFAIK neither Echidna or slither's printer filters tests out. I think not including forge-std makes sense.

Also, I am not sure how the push/pop/log dictionary is managed currently in Foundry, but I think Echidna will always keep the constant pool around and eject the dynamically collected values after running a full sequence. For example, a user's balance that is emitted in one run may help **within** the same sequence but probably unlikely to help in a totally unrelated sequence.

grandizzy:
> AFAIK neither Echidna or slither's printer filters tests out. I think not including forge-std makes sense.
> 

:+1: @0xrusowsky let's include too

> Also, I am not sure how the push/pop/log dictionary is managed currently in Foundry, but I think Echidna will always keep the constant pool around and eject the dynamically collected values after running a full sequence. For example, a user's balance that is emitted in one run may help **within** the same sequence but probably unlikely to help in a totally unrelated sequence.

- The push / pop dictionary + db addresses / storage values are populated when test starts and used across all runs, without being evicted.
These are defined as
https://github.com/foundry-rs/foundry/blob/b823ae0605edb42778a09af962a9c5f85661b28a/crates/evm/fuzz/src/strategies/state.rs#L129-L134
and collected when dict is created
https://github.com/foundry-rs/foundry/blob/b823ae0605edb42778a09af962a9c5f85661b28a/crates/evm/fuzz/src/strategies/state.rs#L52-L54

- We also maintain a dict of so called sample values, dynamically collected from logs, return values & state changes of runs up to a limit (set rn to the test depth) - these are also reused across all runs
https://github.com/foundry-rs/foundry/blob/b823ae0605edb42778a09af962a9c5f85661b28a/crates/evm/fuzz/src/strategies/state.rs#L135-L136

- Then there are the regular values dynamically collected from runs that are not shared between runs
https://github.com/foundry-rs/foundry/blob/b823ae0605edb42778a09af962a9c5f85661b28a/crates/evm/fuzz/src/strategies/state.rs#L123-L124

Please let us know if you see any redundant data / ways to improve the dict. Thank you!

0xrusowsky:
^ note that AST literals are injected into `sample_values`

0xrusowsky:
> Not blocking but I would recommend implementing constant folding to some degree i.e. evaluate `2 * 2 ether`, evaluate `bytes32 IMPLEMENTATION_SLOT = bytes32(uint256(keccak256('eip1967.proxy.implementation')) - 1);`, evaluate `uint(-2)`. Arguably, solar will need this and the code could live there
> 
> See [crytic/echidna#636](https://github.com/crytic/echidna/issues/636)

thanks for the advise! will be tackled next on a follow-up PR:
- https://github.com/foundry-rs/foundry/issues/12044

crates/evm/evm/src/inspectors/stack.rs
```diff
@@ -209,6 +210,7 @@ impl InspectorStackBuilder {
             let mut cheatcodes = Cheatcodes::new(config);
             // Set analysis capabilities if they are provided
             if let Some(analysis) = analysis {
```
grandizzy, line 213: the `analysis` here is technically the compiler, not the analysis per se, should we process / analyze already, smth like 
```rust
if let Some(compiler) = compiler {
    let ast_analysis = AstAnalysis::new(compiler);
    cheatcodes.set_struct_defs(ast_analysis.get_struct_defs().clone());
    stack.set_ast_analysis(ast_analysis);
}
```

and we consolidate AST analysis in single place instead have parts of it in cheatcodes, parts in stack? Then in `EvmFuzzState::new` we just pass `AstAnalysis` and populate dict with `AstAnalysis` words, strings and bytes - side note, in this way we could pass to fuzzer also the enums to be used for https://github.com/foundry-rs/foundry/issues/6623 but that's different scope and complex (it affects mutations as well) as @DaniPopes pointed out

crates/evm/fuzz/src/strategies/param.rs
```diff
@@ -170,27 +170,74 @@ pub fn fuzz_param_from_state(
...
```
grandizzy: IMO this should follow the sample rules, we already have logic / bias to select them

https://github.com/foundry-rs/foundry/pull/12015/files#diff-d37d278bbc4bfc5240900ba4963f1a0f562f98808670ca692658aed9e0fdf624R128-R130

maybe we could reuse same and return `DynSolValue`s from ast analyzed String / bytes here?

crates/evm/fuzz/src/strategies/param.rs
```diff
@@ -170,27 +170,74 @@ pub fn fuzz_param_from_state(
...
                 .boxed(),
             1..=31 => value()
                 .prop_map(move |value| {
-                    // Generate a uintN in the correct range, then shift it to the range of intN
-                    // by subtracting 2^(N-1)
-                    let uint = U256::from_be_bytes(value.0) % U256::from(1).wrapping_shl(n);
-                    let max_int_plus1 = U256::from(1).wrapping_shl(n - 1);
-                    let num = I256::from_raw(uint.wrapping_sub(max_int_plus1));
+                    // Extract lower N bits
+                    let uint_n = U256::from_be_bytes(value.0) % U256::from(1).wrapping_shl(n);
```
grandizzy, line 234: good catch, need to make some more tests to see how this affects overall perf

crates/evm/fuzz/src/strategies/state.rs
```diff
@@ -426,11 +470,181 @@ impl FuzzDictionary {
...
+// -- AST LITERALS COLLECTOR ---------------------------------------------------
+
+enum LitTy {
+    Word(B256),
+    Str(String),
+    Bytes(Bytes),
+}
+
+#[derive(Clone, Default, Debug)]
+pub struct LiteralMaps {
```
grandizzy: as in comment above, would be nice to have all AST analysis consolidated and performed only once, these could be good candidates to move there. Let's add comments to the enum / structs and their members too

crates/evm/evm/src/inspectors/stack.rs
0xrusowsky: i was also thinking about where to place the `LiteralsCollector`, and i guess it could also make sense to upstream it to the inspector stack so that other inspectors could benefit from it.

however, i wouldn't eagerly perform the analysis as you suggest here, as most of the times you won't need to use the analysis capabilities (i.e. only a small subset of tests will use the cheatcodes that require struct defs).

also, i expect each consumer (inspector) to have different needs, hence why i thought having a more granular approach and implementing the actual analysis capabilities on each inspector (i.e. `crates/evm/fuzz/src/strategies/state.rs`, `crates/cheatcodes/src/inspector/analysis.rs`) would make more sense 🤔

let's see what @DaniPopes prefers and we can do what majority thinks its best? haha

crates/evm/fuzz/src/strategies/param.rs
0xrusowsky: wdym exactly?

`bias` is a randomly generated `bool` (50-50) but allocating 50% to ast seeded literals feels like a lot (before it was 0-100).

my idea was that by using `Index` we can allocate a smaller pct to AST string literals, but we are already using them (30% of the time)

crates/evm/evm/src/inspectors/stack.rs
grandizzy: should we then analyze at build time and cache the values / write them to disk, then lazy loading what's needed in different components and only when / where needed? this will also mean we don't need to analyze each time we forge test

crates/config/src/fuzz.rs
```diff
@@ -90,8 +95,10 @@ impl Default for FuzzDictionaryConfig {
             include_push_bytes: true,
             // limit this to 300MB
             max_fuzz_dictionary_addresses: (300 * 1024 * 1024) / 20,
```
DaniPopes: atp can add const MB = 1024*1024 and remove the comments

crates/evm/evm/src/inspectors/stack.rs
DaniPopes: running a couple of visitors is way cheaper than caching anything on disk

crates/evm/fuzz/src/strategies/param.rs
```diff
@@ -170,27 +170,74 @@ pub fn fuzz_param_from_state(
...
-            .prop_map(move |value| {
-                DynSolValue::String(
-                    value.as_str().unwrap().trim().trim_end_matches('\0').to_string(),
-                )
-            })
-            .boxed(),
+        DynSolType::String => {
+            let state_clone = state.clone();
+            (any::<prop::sample::Index>(), any::<prop::sample::Index>())
+                .prop_flat_map(move |(use_ast_index, select_index)| {
```
DaniPopes: you can use [bool::weighted](https://docs.rs/proptest/latest/proptest/bool/fn.weighted.html)

crates/evm/fuzz/src/strategies/param.rs
```diff
@@ -170,27 +170,74 @@ pub fn fuzz_param_from_state(
...
         DynSolType::Bool => DynSolValue::type_strategy(param).boxed(),
-        DynSolType::String => DynSolValue::type_strategy(param)
-            .prop_map(move |value| {
-                DynSolValue::String(
-                    value.as_str().unwrap().trim().trim_end_matches('\0').to_string(),
-                )
-            })
-            .boxed(),
+        DynSolType::String => {
+            let state_clone = state.clone();
```
DaniPopes: rename to state

crates/evm/fuzz/src/strategies/state.rs
```diff
@@ -52,6 +59,20 @@ impl EvmFuzzState {
         // Create fuzz dictionary and insert values from db state.
         let mut dictionary = FuzzDictionary::new(config);
         dictionary.insert_db_values(accs);
+
+        // Seed dict with AST literals if analysis is available.
+        if let Some(compiler) = analysis {
```
DaniPopes: i'd prefer making it even lazier by cloning the compiler into the dictionary, and having OnceLock on the values. for tests you can call .set on the OnceLock and otherwise `if let Some() = compiler { values.get_or_init(|| ) } else { None }`

I guess it's a bit more complex because of the rwlock but you get the point, we want to initialize only when a value is requested

crates/evm/fuzz/src/strategies/state.rs
```diff
@@ -426,11 +470,305 @@ impl FuzzDictionary {
...
+        self.string_literals = map.strings;
+        self.byte_literals = map.bytes;
+        self.sample_values = map.words;
+    }
+}
+
+// -- AST LITERALS COLLECTOR ---------------------------------------------------
+
+#[derive(Debug, Default)]
+pub struct LiteralMaps {
```
DaniPopes: new file?

crates/forge/tests/cli/test_cmd/mod.rs
```diff
@@ -4154,3 +4155,114 @@ Tip: Run `forge test --rerun` to retry only the 2 failed tests
 
 "#]]);
 });
+
+forgetest_init!(should_fuzz_literals, |prj, cmd| {
+    prj.clear_cache_dir();
```
DaniPopes: dont need to clear cache

crates/evm/fuzz/src/strategies/param.rs
0xrusowsky: [`9a05ffb` (#12015)](https://github.com/foundry-rs/foundry/pull/12015/commits/9a05ffb24ddef901d75a9e42eaabe5a0dd85db1c)

crates/evm/fuzz/src/strategies/param.rs
0xrusowsky: [`9a05ffb` (#12015)](https://github.com/foundry-rs/foundry/pull/12015/commits/9a05ffb24ddef901d75a9e42eaabe5a0dd85db1c)

crates/config/src/fuzz.rs
0xrusowsky: [`9a05ffb` (#12015)](https://github.com/foundry-rs/foundry/pull/12015/commits/9a05ffb24ddef901d75a9e42eaabe5a0dd85db1c)

crates/evm/fuzz/src/strategies/state.rs
0xrusowsky: [`5813a5a` (#12015)](https://github.com/foundry-rs/foundry/pull/12015/commits/5813a5adacdb3a5d19d6466d2e4c033116464db1)

crates/evm/fuzz/src/strategies/state.rs
0xrusowsky: addressed in [`e7d3159` (#12015)](https://github.com/foundry-rs/foundry/pull/12015/commits/e7d3159e0b78bf45685fd771bfb2cdde17e3a367)

some impl notes:
- compiler is cloned into `LiteralsDictionary`, which has `OnceLock<LiteralMaps>` (as requested)
- because of the `RwLock` limitations, i kinda used a two-stage lazy init:
  1. AST traversal is deferred until first access via `samples()`, `ast_strings()`, or `ast_bytes()`, which trigger the oncelock
  2. merging literals into `sample_values` is deferred until first runtime sample insertion via `seed_samples()`
      - before merging: `samples()` returns AST literals directly (runtime samples are empty)
      - after merging: `samples()` returns from unified `sample_values`

the trade-off is some memory overhead from keeping `LiteralsDictionary` alive vs avoiding O(n) AST traversal unless fuzzer actually needs samples

crates/evm/fuzz/src/strategies/param.rs
0xrusowsky: moved to `bool::weighted` as suggested in https://github.com/foundry-rs/foundry/pull/12015#discussion_r2426328350

DaniPopes:
need to resolve conflicts by moving tests to forge/tests/cli

crates/evm/evm/src/inspectors/stack.rs
```diff
@@ -388,6 +399,12 @@ impl InspectorStack {
         });
     }
 
+    /// Set the solar compiler instance.
+    #[inline]
+    pub fn set_analysis(&mut self, analysis: Arc<solar::sema::Compiler>) {
```
DaniPopes: can we make a new type for `Arc<Compiler>` to have more control over the type? most code doesn't care that this is using solar

0xrusowsky: decided to avoid `Compiler` as it is also a type used by foundry compilers, so i ended up going with `Analysis`, but lmk if u are opinionated and prefer something else

DaniPopes (dismissed):

crates/forge/tests/cli/test_cmd/invariant/target.rs
```diff
@@ -658,6 +651,7 @@ Tip: Run `forge test --rerun` to retry only the 4 failed tests
 "#]]);
 });
 
+// TODO(rusowsky): figure out why it is flaky
```
grandizzy: is this still a TODO or good to be removed?

crates/forge/tests/cli/test_cmd/mod.rs
```diff
@@ -4206,3 +4207,111 @@ Tip: Run `forge test --rerun` to retry only the 2 failed tests
 
 "#]]);
 });
+
+forgetest_init!(should_fuzz_literals, |prj, cmd| {
```
grandizzy: that's awesome! I assume this one should pass now quite quickly and we can include too? https://github.com/foundry-rs/foundry/blob/e98bff5a57d9f23e60b96e23627512eb5397a598/crates/forge/tests/cli/test_cmd/invariant/storage.rs#L3-L6

Since `should_fuzz_literals` test is only for stateless fuzz, would it make sense to add an invariant one as well? I am thinking in particular to replicate this one https://github.com/foundry-rs/foundry/blob/e98bff5a57d9f23e60b96e23627512eb5397a598/crates/forge/tests/cli/test_cmd/invariant/common.rs#L440 which we weren't able to break without providing the fixtures hints.

grandizzy (commented):
lgtm! left a comment re tests, please check if worth adding them. thanks!

crates/forge/tests/cli/test_cmd/invariant/target.rs
0xrusowsky: good to be removed, always passes locally

crates/forge/tests/cli/test_cmd/mod.rs
0xrusowsky: @grandizzy all passes as expected, see [`4d9ee25` (#12015)](https://github.com/foundry-rs/foundry/pull/12015/commits/4d9ee254c9851444b0813c9030242938ccd93df3)

crates/forge/tests/cli/test_cmd/invariant/common.rs
```diff
@@ -550,6 +552,92 @@ Tip: Run `forge test --rerun` to retry only the 1 failed test
 "#]]);
 });
 
+forgetest_init!(invariant_breaks_without_fixtures, |prj, cmd| {
+    prj.wipe_contracts();
+    prj.update_config(|config| {
+        config.invariant.runs = 1;
```
grandizzy, line 560: just a nit, maybe we should set a seed to make sure it doesn't become flaky

0xrusowsky: [`a4d0c1c` (#12015)](https://github.com/foundry-rs/foundry/pull/12015/commits/a4d0c1ce51e2e9d481d6fbcecdb466a4a3158987)

grandizzy (approved):

grandizzy (approved):

PR #12024, opened 2025-10-08
fix: replace Some(URL) with URL during installation output

milosdjurica:
## Motivation

While installing instead of displaying `(url: Some("https://github.com/foundry-rs/forge-std"), tag: None)` , display `(url: "https://github.com/foundry-rs/forge-std", tag: None)` . Examples can be seen in tests

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


zerosnacks:
Thanks! This looks much better

PR #12025, opened 2025-10-08
feat(cast): cache Etherscan sources under Foundry cache with temp fallback

sashass1315:
This change replaces the ephemeral temp directory used for Etherscan source retrieval with a persistent cache path at ~/.foundry/cache/etherscan/<chain>/sources/<address>. We use the standard Foundry cache location exposed by foundry_config to align with existing caching practices and to avoid repeatedly downloading and writing the same source tree. If creating the cache directory fails (e.g., permissions), we safely fall back to a temp directory. The compilation flow and behavior remain unchanged (no artifacts, same solc selection logic). This improves performance on subsequent runs, reduces unnecessary I/O, and keeps caching consistent with the project’s existing Etherscan client caching strategy.

### DIFF
--- a/crates/cast/src/cmd/storage.rs
+++ b/crates/cast/src/cmd/storage.rs
@@ -172,7 +172,7 @@ impl StorageArgs {
                 contract_root
             }
         } else {
-            tempfile::tempdir()?.path().to_path_buf()
+            tempfile::tempdir()?.keep()
         };
         let mut project = etherscan_project(metadata, &root_path)?;
         add_storage_layout_output(&mut project);


crates/cast/src/cmd/storage.rs
```diff
@@ -159,11 +159,22 @@ impl StorageArgs {
...
+            let sources_root = cache_root.join("sources");
+            let contract_root = sources_root.join(format!("{address}"));
+            if let Err(err) = std::fs::create_dir_all(&contract_root) {
+                sh_warn!("Could not create etherscan cache dir, falling back to temp: {err}")?;
+                tempfile::tempdir()?.path().to_path_buf()
+            } else {
+                contract_root
+            }
+        } else {
+            tempfile::tempdir()?.path().to_path_buf()
```
DaniPopes:
```suggestion
            tempfile::tempdir()?.keep()
```

Issue #11823, opened 2025-09-25
bug(doc): forge doc does not support overloaded functions in interfaces properly


KirienzoEth:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

1.3.5-stable

### What version of Foundryup are you on?

1.3.0

### What command(s) is the bug in?

forge doc --serve

### Operating System

Windows

### Describe the bug

`forge doc` does not support overloaded functions in interfaces properly.
The doc generated for the interface is correct, but the doc for the contract uses the NatSpecs of the first function it finds with the same name in the interface.

Interface:

<img width="782" height="429" alt="Image" src="https://github.com/user-attachments/assets/02b27fc9-922d-4fd5-a540-66ac27c4aa77" />

Contract:

<img width="786" height="402" alt="Image" src="https://github.com/user-attachments/assets/7f21b169-8325-412c-9c5c-f9f6c18b138d" />

I published an example repo here: https://github.com/KirienzoEth/foundry-forge-doc-overloaded-functions-interface-issue

I suspect it's because the name of the function is used instead of its signature when merging the comments:
https://github.com/foundry-rs/foundry/blob/v1.3.6/crates/doc/src/writer/as_doc.rs#L304

PR #12040, opened 2025-10-09
fix(forge): handle overloaded functions correctly in doc

wellnana:
closes https://github.com/foundry-rs/foundry/issues/11823, it seems like all the overloaded functions would inherit the documentation from the first function with the same name, rather than matching based on the full function signature, so the `forge doc` output would be incorrect for overloaded functions using `@inheritdoc`.

### DIFF
--- a/crates/doc/src/parser/item.rs
+++ b/crates/doc/src/parser/item.rs
@@ -1,4 +1,4 @@
-use crate::{Comments, solang_ext::SafeUnwrap};
+use crate::{Comments, helpers::function_signature, solang_ext::SafeUnwrap};
 use solang_parser::pt::{
     ContractDefinition, ContractTy, EnumDefinition, ErrorDefinition, EventDefinition,
     FunctionDefinition, StructDefinition, TypeDefinition, VariableDefinition,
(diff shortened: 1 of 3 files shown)


crates/doc/src/writer/as_doc.rs
```diff
@@ -301,8 +301,21 @@ impl Document {
         code: &str,
     ) -> Result<(), std::fmt::Error> {
         let func_name = func.name.as_ref().map_or(func.ty.to_string(), |n| n.name.to_owned());
+        let func_sign = if !func.params.is_empty() {
```
grandizzy: let's extract this in a helper fn and reuse / avoid code duplication

wellnana:
> thank you! even if we're going to replace solang with solar in `forge doc` I think it's still good to have the fix and the unit test. Left a nit, please have a look

Thanks! Fix and add test now.

crates/doc/src/parser/item.rs
```diff
@@ -169,6 +169,29 @@ impl ParseSource {
         }
     }
 
+    /// Get the signature of the source (for functions, includes parameter types)
+    pub fn signature(&self) -> String {
+        match self {
+            Self::Function(func) => {
+                let name = func.name.as_ref().map_or(func.ty.to_string(), |n| n.name.to_owned());
```
grandizzy: this logic is still duped in `fn function_signature` can we reuse?

wellnana: Sorry for missing this one, move the function_signature into the helpers package, can reuse it now

Issue #10577, opened 2025-05-21
chore: fix compilation warnings in Foundry's Solidity test suite


zerosnacks:
We have a ton of compilation warnings (unused variables, shadowing builtins, unused return values etc..) in Foundry's test suite

To start fixing them:

```
cd testdata
```

temporarily comment out `solc = "0.8.18"` in `foundry.toml`

and run `forge build --ignored-error-codes=2018`

Note we permit `Warning (2018): Function state mutability can be restricted to pure`.

You will be presented with a long list of warnings

```
Warning (1878): SPDX license identifier not provided in source file. Before publishing, consider adding a comment containing "SPDX-License-Identifier: <SPDX-License>" to each source file. Use "SPDX-License-Identifier: UNLICENSED" for non-open-source code. Please see https://spdx.org for more information.
Warning: SPDX license identifier not provided in source file. Before publishing, consider adding a comment containing "SPDX-License-Identifier: <SPDX-License>" to each source file. Use "SPDX-License-Identifier: UNLICENSED" for non-open-source code. Please see https://spdx.org for more information.
--> fixtures/SolidityGeneration/GeneratedFastLane.sol

Warning (1878): SPDX license identifier not provided in source file. Before publishing, consider adding a comment containing "SPDX-License-Identifier: <SPDX-License>" to each source file. Use "SPDX-License-Identifier: UNLICENSED" for non-open-source code. Please see https://spdx.org for more information.
Warning: SPDX license identifier not provided in source file. Before publishing, consider adding a comment containing "SPDX-License-Identifier: <SPDX-License>" to each source file. Use "SPDX-License-Identifier: UNLICENSED" for non-open-source code. Please see https://spdx.org for more information.
--> fixtures/SolidityGeneration/GeneratedGaugeController.sol

Warning (1878): SPDX license identifier not provided in source file. Before publishing, consider adding a comment containing "SPDX-License-Identifier: <SPDX-License>" to each source file. Use "SPDX-License-Identifier: UNLICENSED" for non-open-source code. Please see https://spdx.org for more information.
Warning: SPDX license identifier not provided in source file. Before publishing, consider adding a comment containing "SPDX-License-Identifier: <SPDX-License>" to each source file. Use "SPDX-License-Identifier: UNLICENSED" for non-open-source code. Please see https://spdx.org for more information.
--> fixtures/SolidityGeneration/GeneratedLiquidityGaugeV4.sol

[...]

Issue #12044, opened 2025-10-10
feat(fuzz): impl constant folding for typical expressions


0xrusowsky:
### Component

Forge

### Describe the feature you would like

as pointed in https://github.com/foundry-rs/foundry/pull/12015#issuecomment-3385828190, the set of dictionary seeds should be expanded beyond individual literals, and should incorporate folding of typical expressions.

this is feasible to implement by enhancing the `LiteralsCollector` visitor

ref: https://github.com/crytic/echidna/issues/636

PR #10744, opened 2025-06-10
chore: fix compilation warnings in Foundry's Solidity test suite

knelsondev:
## Motivation

This PR aims to close #10577 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #12049, opened 2025-10-10
test: refactor testdata/ tests to be run in `forge test`

DaniPopes:
Refactor manual integration tests to just run `forge test` on the `testdata` folder project.

We set up testdata/ to be a Foundry project, and a single test that runs `forge test` on it.

Not all current tests pass due to inability of asserting higher level things like console output, general test failure, etc. These tests are migrated to inline `forgetest!` Rust snapshot tests.

The bulk of the manual work was migrating fuzz and invariant tests that assert data structures like counterexample list into `str![[]]` tests, making sure the config is also migrated from the relevant Rust test.
The rest was mostly automated / claude-coded.

Followups:
- make testdata/ project have a more standard structure (src/, test/, better directory name)
- remove DSTest in favor of forge-std
- optimize `Config::load_with_root` in `update_config`. We ignore remappings so we shouldnt try to resolve them at all

### DIFF
--- a/crates/forge/tests/cli/test_cmd/trace.rs
+++ b/crates/forge/tests/cli/test_cmd/trace.rs
@@ -80,6 +80,7 @@ Ran 1 test suite [ELAPSED]: 1 tests passed, 0 failed, 0 skipped (1 total tests)
 "#]]);
 });
 
+#[cfg(not(feature = "isolate-by-default"))]
 forgetest_init!(trace_test, |prj, cmd| {
     prj.wipe_contracts();
     prj.add_test(


grandizzy:
this will be a game changer for CI! I'd suggest getting rid of default / Paris/ shanghai testdata / test runners stuff and using the newly added vm.setEvmVersion cheatcode to assert different outcomes in same tests instead

DaniPopes:
right, we can do that as a follow-up since the diff is going to be big already from just fixing tests and making it work in the first place

crates/forge/tests/cli/test_cmd/invariant/storage.rs
```diff
@@ -1,6 +1,13 @@
-pragma solidity >0.8.13;
+use super::*;
 
-import "ds-test/test.sol";
+forgetest_init!(
+    #[ignore = "slow"]
```
grandizzy, line 4: we should re-enable this one with AST seeded fuzz dictionary PR, should be much faster cc @0xrusowsky

crates/forge/tests/cli/test_cmd/fuzz.rs
```diff
@@ -0,0 +1,735 @@
...
+
+Tip: Run `forge test --rerun` to retry only the 1 failed test
+
+"#]]);
+});
+
+// Test that showcases PUSH collection on normal fuzzing.
+// Ignored until we collect them in a smarter way.
+forgetest_init!(
+    #[ignore]
```
grandizzy, line 396: same as for https://github.com/foundry-rs/foundry/pull/12049#discussion_r2438755939 with AST seeded fuzz dictionary we will collect values of `0x4446` and `0x5556` and use to fuzz selectors.

For the
```
        counter += numToIncrement;
        counterX2 += numToIncrement * 2;
```
case wonder if folding could help https://github.com/foundry-rs/foundry/issues/12044 cc @0xrusowsky

0xrusowsky: in our current plans we only had constant folding. In this case how would we derive a value for `numToIncrement`?

at most we could use the `+=` and the `* 2` to create smarter mutators for dict values, but not totally sure how to leverage that

crates/forge/tests/cli/test_cmd/repros.rs
```diff
@@ -1,13 +1,438 @@
-// SPDX-License-Identifier: MIT OR Apache-2.0
-pragma solidity ^0.8.18;
+//! Regression tests for specific GitHub issues
 
-import "ds-test/test.sol";
-import "cheats/Vm.sol";
+use foundry_test_utils::str;
```
grandizzy, line 3: just a note, pls lmk if I am wrong: we should still add sol files repros as before in `testdata/default/repros` but if there are repros that require check of output (fails or we want to check traces) we should add them here and assert

0xrusowsky: that's also what i understood + it aligns with the PR desc

testdata/foundry.toml
```diff
@@ -1,14 +1,60 @@
...
 test = "./"
-solc = "0.8.18"
-evm_version = "paris"
-ffi = false
-fs_permissions = [{ access = "read-write", path = "./" }]
+optimizer = true
+optimizer_runs = 200
+via_ir = false
+ignored_error_codes = [
+    1878, # SPDX license identifier not provided
```
grandizzy, line 8: should we close https://github.com/foundry-rs/foundry/issues/10577 with this PR? cc @zerosnacks 
And probably https://github.com/foundry-rs/foundry/pull/10744 not needed anymore

zerosnacks: it would be cleaner to resolve them instead of ignoring them but I can quite easily create a follow-up PR to reduce the number of ignored error codes, especially `2319` and `2519` need to be addressed

zerosnacks: #10744 has been closed, I'll update the description of #10577 after merging of this PR and assign it to me

zerosnacks (approved):
lgtm

crates/forge/tests/cli/test_cmd/repros.rs
DaniPopes: yeah that's right, if the test can be implemented using foundry directly then it should go in testdata
only if it needs higher level assertions like console output should it go to forgetest!

PR #12050, opened 2025-10-10
fix(forge): dynamic test linking for try catch with custom return

grandizzy:
## Motivation
- supersede https://github.com/foundry-rs/foundry/pull/11984
- if try catch with custom return then cast address returned by cheatcode to proper type by using helper fn

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/common/src/preprocessor/deps.rs
+++ b/crates/common/src/preprocessor/deps.rs
@@ -100,8 +100,8 @@ enum BytecodeDependencyKind {
         value: Option<String>,
         /// `salt` (if any) used when creating contract.
         salt: Option<String>,
-        /// Whether it's a try contract creation statement.
-        try_stmt: bool,
+        /// Whether it's a try contract creation statement, with custom return.
+        try_stmt: Option<bool>,
     },
 }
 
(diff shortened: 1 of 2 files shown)


crates/common/src/preprocessor/deps.rs
```diff
@@ -328,8 +351,14 @@ pub(crate) fn remove_bytecode_dependencies(
                     salt,
                     try_stmt,
                 } => {
-                    let (mut update, closing_seq) = if *try_stmt {
-                        (String::new(), "})")
+                    let (mut update, closing_seq) = if let Some(has_ret) = try_stmt {
+                        if *has_ret {
+                            // try this.addressToCounter1() returns (Counter c)
+                            try_catch_helpers.insert(name);
+                            (format!("this.addressTo{name}{id}(", id = contract_id.get()), "}))")
```
DaniPopes, line 358: should ID be dep.referenced_contract?

DaniPopes:
rest lgtm

crates/common/src/preprocessor/deps.rs
grandizzy: I added the id of the preprocessed contract (test contract) to avoid issues when test inherits another test, for example if `CounterTestA is CounterTestB` and they both have the `addressToCounter` function. Adding test contract id will make them unique and avoid collisions

PR #12056, opened 2025-10-11
fix: use blobparams for blob fee calc

mattsse:
fixes an blob fee calc issue that was causing issues in alloy ci, triggered by high blob fees

https://github.com/alloy-rs/alloy/actions/runs/18424312562/job/52503400740

we we're using outdated excess gas calc logic (pre prague).

This now integrates BlobParams into the feemanager which should also future proof this for the upcoming blob param only hardforks.

we can do some additional cleanup as followup and entirely get rid of `get_blob_base_fee_update_fraction`

### DIFF
--- a/crates/anvil/tests/it/eip4844.rs
+++ b/crates/anvil/tests/it/eip4844.rs
@@ -96,7 +96,7 @@ async fn can_send_eip4844_transaction_eth_send_transaction() {
         .with_to(bob)
         .with_blob_sidecar(sidecar.clone());
 
-    let pending_tx = provider.send_transaction(tx.into()).await.unwrap();
+    let pending_tx = provider.send_transaction(tx).await.unwrap();
     let receipt = pending_tx.get_receipt().await.unwrap();
     let tx_hash = receipt.transaction_hash;
 


crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -1149,11 +1149,15 @@ impl Backend {
...
     /// Returns the environment for the next block
+    ///
+    /// This is used for obtaining the evm environment for the next (pending) block (e.g.
+    /// transaction validation in eth_sendrawTransaction)
     fn next_env(&self) -> Env {
         let mut env = self.env.read().clone();
         // increase block number for this block
         env.evm_env.block_env.number = env.evm_env.block_env.number.saturating_add(U256::from(1));
         env.evm_env.block_env.basefee = self.base_fee();
+        env.evm_env.block_env.blob_excess_gas_and_price = self.excess_blob_gas_and_price();
```
mattsse, line 1160: this was one bug: we need to use the next blob fees for tx validation

crates/anvil/src/eth/fees.rs
```diff
@@ -158,15 +162,27 @@ impl FeeManager {
 
     /// Calculates the next block blob base fee.
     pub fn get_next_block_blob_base_fee_per_gas(&self) -> u128 {
-        alloy_eips::eip4844::calc_blob_gasprice(
```
mattsse, line 161: this was using cancun only fee calc logic that was changed slightly in prague, this now uses the generalized BlobParams api

crates/anvil/tests/it/eip4844.rs
```diff
@@ -76,6 +76,33 @@ async fn can_send_eip4844_transaction_fork() {
     let _blobs = api.anvil_get_blob_by_tx_hash(tx_hash).unwrap().unwrap();
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn can_send_eip4844_transaction_eth_send_transaction() {
```
mattsse, line 80: this is the same alloy test so this should no longer be flaky in alloy

PR #12057, opened 2025-10-11
fix: tighten BroadcastReader directory validation

rnkrtt:
Replaced the guard in `BroadcastReader::new` to bail as soon as the provided path isn’t a directory—previously regular files slipped through and crashed deeper in the reader. Before: `if !broadcast_path.exists() && !broadcast_path.is_dir()` let a file path through; after: `if !broadcast_path.is_dir()` fails fast with `bail!("broadcast dir does not exist")`. I’m highlighting to reviewers: _If someone accidentally points at a file instead of the broadcast folder, they now get a clear error up front instead of a confusing panic later._

### DIFF
--- a/crates/script-sequence/src/reader.rs
+++ b/crates/script-sequence/src/reader.rs
@@ -26,7 +26,7 @@ impl BroadcastReader {
     /// Create a new `BroadcastReader` instance.
     pub fn new(contract_name: String, chain_id: u64, broadcast_path: &Path) -> Result<Self> {
         if !broadcast_path.is_dir() {
-            bail!("broadcast dir does not exist");
+            bail!("broadcast dir does not exist or is not a directory");
         }
 
         Ok(Self {


crates/script-sequence/src/reader.rs
```diff
@@ -25,7 +25,7 @@ pub struct BroadcastReader {
 impl BroadcastReader {
     /// Create a new `BroadcastReader` instance.
     pub fn new(contract_name: String, chain_id: u64, broadcast_path: &Path) -> Result<Self> {
-        if !broadcast_path.exists() && !broadcast_path.is_dir() {
+        if !broadcast_path.is_dir() {
             bail!("broadcast dir does not exist");
```
zerosnacks: Let's update this message to reflect that:

broadcast directory does not exist or is not a directory

rnkrtt:
Message updated, thanks

PR #12062, opened 2025-10-11
Limit HTTP redirects in npm installer to prevent infinite loops

vtjl10:
- **Summary**: Cap redirect chain to 10 in `makeRequest` within `npm/src/install.ts` to avoid hangs/DoS on cyclic redirects.
- **Changes**:
  - Add `maxRedirects` parameter (default: 10).
  - Decrement on each 3xx follow; reject with “Too many redirects” when exhausted.

### DIFF
--- a/npm/src/install.ts
+++ b/npm/src/install.ts
@@ -38,7 +38,7 @@ function ensureSecureUrl(urlString: string, purpose: string) {
   }
 }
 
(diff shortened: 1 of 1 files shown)


zerosnacks:
Thanks for your PR, however I don't think this is necessary

@o-az feel free to re-open if you think this is required

Issue #12008, opened 2025-10-07
chore(lint): disable linter when versions solidity versions are not supported by solar


0xrusowsky:
## Motivation

solar has compatibility with the latest Solidity language breaking version (0.8.*), so linting projects that are older results in errors like the ones reported in https://github.com/foundry-rs/foundry/issues/11668#issuecomment-3376414455

## Solution

we should disable `lint_on_build` for older solidity version

Issue #11668, opened 2025-09-16
`forge lint` throws `Error: solar run failed` with latest nightly build


clemlak:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

1.3.5-nightly (9cd8a9511b 2025-09-16T06:02:33.471215000Z)

### What version of Foundryup are you on?

1.3.0

### What command(s) is the bug in?

forge lint

### Operating System

macOS (Apple Silicon)

### Describe the bug

Running `forge lint` with the latest nightly build throws `Error: solar run failed`. Using the latest stable version works fine though.

DaniPopes:
Can you please provide more context? Any code I can reproduce with, the full output of forge, etc.

There should be diagnostics reported for lints before this error message.

grandizzy:
bump @clemlak thank you!

MaxMustermann2:
I can reproduce this on one of my code bases. I am trying to decipher why. It does not error on a fresh folder (generated with `forge init`) using the same configuration file.

```
forge Version: 1.3.6-nightly
Commit SHA: 5dbd958d16125937bc761aece6cf3e69883ebdb3
Build Timestamp: 2025-09-17T06:03:07.571539555Z (1758088987)
Build Profile: maxperf
```

```
Warning: evm_version 'paris' may be incompatible with solc version. Consider using 'prague'
[⠊] Compiling...
No files changed, compilation skipped
Error: Lint failed: solar run failed
```

```toml
[profile.default]
src = "src"
out = "out"
libs = [
    "lib",
    "node_modules",
]
# ignore warnings about unused parameters
# and code size and init code size: the code size
# issue is not relevant because
# this repository does not make any deployments.
ignored_error_codes = [5667, 5574, 3860]

# turn off the annoying linter
[lint]
exclude_lints = ["unused-import", "unaliased-plain-import"]
```

Edit: I cannot reproduce this on a public repository just yet.

grandizzy:
@MaxMustermann2 at minimum would you be able to provide output of running with RUST_LOG=trace ? thank you

MaxMustermann2:
[log.txt](https://github.com/user-attachments/files/22407037/log.txt)

clemlak:
[...]

PR #12065, opened 2025-10-12
chore(forge): filter out unsupported solidity versions [solar]

marcvernet31:
## Motivation

As commented on this task https://github.com/foundry-rs/foundry/issues/11668#issuecomment-3376414455, Solar linter is not compatible with Solidity versions older that 0.8.0, which creates errors on execution.
- Documentation on Solar limitation: https://github.com/paradigmxyz/solar/blob/14ff19855dc49f8ef68bc3758ed5c2473e4a87b7/README.md?plain=1#L29

closes https://github.com/foundry-rs/foundry/issues/12008 

## Solution

- Added a check that disables the linter when any of the compiled files has a version inferior to 0.8.0, displaying a warning message on console.
- Added unit tests on `lint.rs`
## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/test-utils/src/rpc.rs
+++ b/crates/test-utils/src/rpc.rs
@@ -192,7 +192,6 @@ fn next_url_inner(is_ws: bool, chain: NamedChain) -> String {
         if !rpc_url.is_empty() {
             return rpc_url;
         }
-        return "https://arbitrum-one-rpc.publicnode.com".to_string();
     }
 
     let reth_works = true;


0xrusowsky:
hey @marcvernet31, thanks for the PR!

just fyi i had this assigned to me, cause its part of a bigger solar-integration workstream. This isn't a linter-specific issue, as it impacts any component of the stack that use the solar compiler.

the proposed solution in the PR works, but the scope is too narrow. What we want is to filter incompatible sources in `fn configure_pcx_from_compile_output()` so that all commands that use the solar compiler benefit from it automatically.

the issue is that solar currently errors when you try to lower without sources, so we'd need to manually handle conditionals lowering everywhere, which is annoying... because of that, i opened https://github.com/paradigmxyz/solar/pull/572 to make solar treat empty sources as a no-op.

imo we should wait for that to merge, then implement the proper fix at the higher level.
happy to collaborate on this if you want to wait for the solar change, or i can take it from  here. Let me know what you prefer!

marcvernet31:
Hey @0xrusowsky, happy to try to finish this one once the solar fix is available. Just to clarify, we want to filter out all sources with incompatible Solidity version on `configure_pcx_from_compile_output` so that they are ignored right?

0xrusowsky:
@marcvernet31 the solar PR was merged this afternoon, and i went ahead and implemented the fix at the higher level as previously mentioned [`23ec372` (#12065)](https://github.com/foundry-rs/foundry/pull/12065/commits/23ec372aead9400403e72513177798755af00c48)

this means the checks you added in `build.rs` with `fn has_non_supported_solar_solidity_version()` are now redundant, since the filtering happens upstream for all commands that use solar (not just linting).

the tests you wrote are great though! we just need to adapt them a bit and remove the redundant code from your initial commit.
imo we want to keep the `forge build` tests to assert that the cmd runs without errors nor warnings, and also run `forge lint` to assert that the "no sources" warning is logged.

could u take care of that, please? thanks a lot!

0xrusowsky:
please have a second look after the changes.

@DaniPopes friendly ping as u may be opinionated on the approach

crates/forge/tests/cli/lint.rs
```diff
@@ -778,3 +782,43 @@ fn ensure_no_privileged_lint_id() {
...
+            exclude_lints: vec![],
+            ignore: vec![],
+            lint_on_build: true,
+            ..Default::default()
+        };
+    });
+
+    // Run forge build - should SUCCEED without linting
+    cmd.arg("build").assert_success().stderr_eq(
+        "Warning: unable to lint. Solar only supports Solidity versions prior to 0.8.0\n",
```
DaniPopes: please use str! macro

crates/forge/tests/cli/lint.rs
```diff
@@ -778,3 +782,43 @@ fn ensure_no_privileged_lint_id() {
         assert_ne!(lint.id(), "all", "lint-id 'all' is reserved. Please use a different id");
     }
 }
+
+forgetest!(skips_linting_for_old_solidity_versions, |prj, cmd| {
+    prj.wipe_contracts();
+
```
DaniPopes: this does nothing
```suggestion
```

crates/cli/src/opts/build/utils.rs
```diff
@@ -59,10 +61,16 @@ pub fn configure_pcx(
         .ok_or_else(|| eyre::eyre!("no Solidity sources"))?
         .1
         .into_iter()
+        // Filter unsupported versions
+        .filter(|(v, _, _)| v >= &MIN_SUPPORTED_VERSION)
         // Always pick the latest version
         .max_by(|(v1, _, _), (v2, _, _)| v1.cmp(v2))
         .unwrap();
```
DaniPopes: this could panic now

0xrusowsky: addressed in [`80e9697` (#12065)](https://github.com/foundry-rs/foundry/pull/12065/commits/80e96970ddc455ff25af5619ac74dd98dfa62d6c)

crates/cli/src/opts/build/utils.rs
```diff
@@ -7,12 +7,14 @@ use foundry_compilers::{
...
 use foundry_config::{Config, semver::Version};
 use rayon::prelude::*;
-use solar::sema::ParsingContext;
+use solar::{interface::MIN_SOLIDITY_VERSION as MSV, sema::ParsingContext};
 use std::{
     collections::{HashSet, VecDeque},
     path::{Path, PathBuf},
 };
 
+const MIN_SUPPORTED_VERSION: Version = Version::new(MSV.0, MSV.1, MSV.2);
```
DaniPopes, line 16: can we make this a semver Version in solar instead of ints?

DaniPopes (approved):

crates/cli/src/opts/build/utils.rs
0xrusowsky: sure, let's do that since we haven't released yet and ended up patching the version

Issue #12076, opened 2025-10-13
bug(`fmt`): bad block comment formatting and formatter rules ignored


Solthodox:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

v1.4.0-rc3

### What version of Foundryup are you on?

Latest

### What command(s) is the bug in?

forge fmt

### Operating System

None

### Describe the bug

The formatting is not exactly as is used to be with former `forge fmt` as it breaks some custom commets formatting. In my case im using some of the code from [solady](https://github.com/Vectorized/solady): 

For example in `Initializable` contract:

BEFORE : 
```solidity
    /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
    /*                       CUSTOM ERRORS                        */
    /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
```

AFTER : 
```solidity
    /* ´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
    /* CUSTOM ERRORS */
    /* .•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
```

Also the commented custom formatter rules are not detected by this new formatter such as : 
`// forgefmt: disable-start` and similars

0xrusowsky:
i will look into this, most likely is cause we do some trimming for block comments. I think we need to avoid trimming when a block comment takes a single line.

regarding inline-config, i believe that u are not following the expected format. @Solthodox  can you try adding a space after the `:`? we have several unit tests in place, and `// forgefmt: disable-next-line` works as intended there:
https://github.com/foundry-rs/foundry/blob/master/crates/fmt/testdata/InlineDisable/fmt.sol#L3

grandizzy:
@Solthodox can you please confirm using `// forgefmt: disable-next-line` works for you?

Solthodox:
[...]

PR #12078, opened 2025-10-13
fix(fmt): don't normalize single-line non-doc block cmnts

0xrusowsky:
## Motivation

closes https://github.com/foundry-rs/foundry/issues/12076

## Solution

don't normalize single-line non-doc block comments

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/fmt/src/state/mod.rs
+++ b/crates/fmt/src/state/mod.rs
@@ -603,7 +603,11 @@ impl<'sess> State<'sess, '_> {
         let content = &line[prefix.len()..];
         let content = if is_doc {
             // Doc comments preserve leading whitespaces (right after the prefix) as nbps.
-            let ws_len = content.chars().take_while(|&c| c.is_whitespace()).count();
+            let ws_len = content
+                .char_indices()
+                .take_while(|(_, c)| c.is_whitespace())
+                .last()
+                .map_or(0, |(idx, c)| idx + c.len_utf8());
             let (leading_ws, rest) = content.split_at(ws_len);
             if !leading_ws.is_empty() {
                 self.word(leading_ws.to_owned());


crates/fmt/testdata/SimpleComments/wrap-comments.fmt.sol
```diff
@@ -4,6 +4,10 @@ contract SimpleComments {
     mapping(address /* asset */ => address /* router */)
         public router;
 
+    /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
+    /* FUNCTIONS */
+    /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
```
0xrusowsky, line 14: when wrapping comments we need to split the comment content into words, and separate them with soft breaking spaces.

if a user has comment wrapping enabled and they want to do custom art with their comments, they may have to disable the fmt for that block.

imo this is totally fine, as custom style conflicts with comment wrapping

crates/fmt/src/state/mod.rs
```diff
@@ -606,6 +606,9 @@ impl<'sess> State<'sess, '_> {
             }
             let prefix = post_break_prefix(prefix, rest.len());
             (prefix, rest)
+        } else if line.starts_with("/*") && !line.starts_with("/* ") {
```
grandizzy: cannot this happen on comments starting with `//` or `///` ?

0xrusowsky: afaik doc comments must have the space afterwards, otherwise they are not identified as doc comments.

personally i felt like `//*` was weird, but i can modify the patch to allow it too. ~Should i do it @grandizzy?~ addressed in [`6983489` (#12078)](https://github.com/foundry-rs/foundry/pull/12078/commits/6983489f2517c7b44777c7d4370e009d04e37ed7)

crates/fmt/testdata/SimpleComments/wrap-comments.fmt.sol
DaniPopes: does this mean that you can't have more than 1 consecutive space inside of a comment when wrapping is enabled? we can ignore this here but that doesn't sound right

0xrusowsky: since when wrapping comments we defer the breaking logic to the PP, and we wouldn't know when it would break, the current approach is to split cmnt content into words, stripping whitespaces and exchanging them for a single PP breaking space:
https://github.com/foundry-rs/foundry/blob/8e3d10e226a9c08de3918e791965062dfa77197d/crates/fmt/src/state/mod.rs#L620

i can change the approach and go char by char though if we are not happy with this approach

0xrusowsky: addressed in [`6983489` (#12078)](https://github.com/foundry-rs/foundry/pull/12078/commits/6983489f2517c7b44777c7d4370e009d04e37ed7)

crates/fmt/src/state/mod.rs
```diff
@@ -584,52 +584,79 @@ impl<'sess> State<'sess, '_> {
...
-            self.word(prefix);
-            let content = &line[prefix.len()..];
-            let (leading_ws, rest) =
-                content.split_at(content.chars().take_while(|&c| c.is_whitespace()).count());
+        self.word(prefix);
+
+        let content = &line[prefix.len()..];
+        let content = if is_doc {
+            // Doc comments preserve leading whitespaces (right after the prefix) as nbps.
+            let ws_len = content.chars().take_while(|&c| c.is_whitespace()).count();
```
DaniPopes: this is char count instead of byte count

PR #12080, opened 2025-10-13
fix(anvil): set envelope for non deposit tx in debug tracers

grandizzy:
## Motivation
- if optimisim then set enveloped tx when replay tx for debug tracers
- ref https://github.com/foundry-rs/foundry/pull/11709

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


grandizzy:
> do we need tests? or is the effort-reward ratio so low that we can skip them?

it's quite complex, was able to repro and test with relay e2e but not sure how to translate in a simpler one...

Issue #12075, opened 2025-10-13
chore(`deps`): bump revm to v30+


zerosnacks:
To bump to revm 30+:

Dependencies:

- [x] foundry-fork-db: https://github.com/foundry-rs/foundry-fork-db/pull/77
- [x] alloy-evm: bumped in https://github.com/alloy-rs/evm/pull/183/files
- [x] revm-inspectors: https://github.com/paradigmxyz/revm-inspectors/pull/367

~~Expected required changes:~~

- [ ] ~~Fix breaking changes arising from https://github.com/bluealloy/revm/pull/3086~~

Things to note:

- `skip_cold_load` when account loading, needs to be set to true
- revm 30 bump in Reth: https://github.com/paradigmxyz/reth/pull/18999
- review https://github.com/bluealloy/revm/pull/3086 together w/ Arsenii

Migration guide:

https://github.com/bluealloy/revm/blob/main/MIGRATION_GUIDE.md#v88-tag-revm-v3000

~~Blocked by:~~

~~In Revm 30, `CallInputs` currently requires setting `bytecode` and `bytecode_hash`, this will be updated in `revm` to be optional. This limits the changes required in Foundry and would fix the currently failing test related to mock function.~~

zerosnacks:
No longer blocked, to be bumped to `revm = "30.2.0"` which includes the requested API change of `CallInputs`

PR #12094, opened 2025-10-14
chore(deps): bump revm 30+, alloy-evm, revm-inspectors

zerosnacks:
## Motivation

Closes: https://github.com/foundry-rs/foundry/issues/12075

## Solution

Bumps to latest `revm`, `alloy-evm` & `revm-inspectors` including requested API changes

See inline callouts for additional context

## Breaking changes

- Serializing the `TraceArena` no longer includes the `depth` and `contractAddr` in the step trace, ref: https://github.com/paradigmxyz/revm-inspectors/pull/355. In turn the `DebugStep` struct no longer includes these fields when calling `vm.stopAndReturnDebugTraceRecording` as these values are constant throughout a single call depth. At the top-level the `depth` is now included.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Breaking changes


crates/evm/evm/src/inspectors/stack.rs
```diff
@@ -944,6 +944,7 @@ impl Inspector<EthEvmContext<&mut dyn DatabaseExt>> for InspectorStackRefMut<'_>
                     call.input.bytes(ecx).get(..4).and_then(|selector| mocks.get(selector))
                 }) {
                     call.bytecode_address = *target;
+                    call.known_bytecode = None;
```
zerosnacks, line 965: If not reset to `None` old bytecode is loaded by hash rather than from the set target the line above

`known_bytecode` is a new field introduced in revm 30

crates/cheatcodes/src/evm/record_debug_step.rs
```diff
@@ -65,25 +74,31 @@ fn recursive_flatten_call_trace<'a>(
...
+    let is_out_of_gas = matches!(
+        ctx.step.status,
+        Some(
+            InstructionResult::OutOfGas
+                | InstructionResult::MemoryOOG
+                | InstructionResult::MemoryLimitOOG
+                | InstructionResult::PrecompileOOG
+                | InstructionResult::InvalidOperandOOG
+        )
+    );
```
zerosnacks, line 92: small optimization to avoid branching

crates/cheatcodes/src/evm/record_debug_step.rs
```diff
@@ -40,13 +47,15 @@ fn recursive_flatten_call_trace<'a>(
     }
 
     let node = &arena.nodes()[node_idx];
+    let depth = node.trace.depth as u64 + 1;
+    let contract_addr = node.execution_address();
```
zerosnacks: Per https://github.com/paradigmxyz/revm-inspectors/issues/354 the `depth` and `contract` fields are no longer included and expected to be collected in the context of the steps' CallTrace.

The `+1` fixes an index mismatch causing the a test failure on `testDebugTraceCanRecordDepth`.

I believe this fix and implementation is correct but please double check this @DaniPopes 

The `+1` was inspired by a follow-up fix added here: https://github.com/paradigmxyz/revm-inspectors/pull/369 by @klkvr

zerosnacks:
Remaining test failure is Etherscan / CI related, going to mark as ready for review

zerosnacks:
~~Moved back to draft to discuss changes raised by Dragan around account loading and the `skip_cold_load` functions to circumvent BAL restrictions.~~

After discussion we decided that no change is required on our side in regards to `load_account`, `sload`, `sstore` as they are not gas calculation sensitive in Foundry and behavior of old functions remains the same as before.

Looking into panic in external test suite.

crates/cheatcodes/src/inspector.rs
```diff
@@ -834,6 +834,8 @@ impl Cheatcodes {
 
                 // At the target depth we set `msg.sender`
                 if curr_depth == prank.depth {
+                    // Ensure new caller is loaded and touched
+                    let _ = journaled_account(ecx, prank.new_caller);
```
zerosnacks, line 839: This fixes a regression introduced in `revm 30` where it now assumes the "from" account in a transfer is always loaded. We were experiencing a panic if `startPrank` was applied in the `setup` and then a value transfer happened in a test.

Ref: https://github.com/foundry-rs/foundry/blob/3554208b218ef166171bb8aeceda477e9481ee94/testdata/default/repros/Issue12075.t.sol

grandizzy: can you pls check `startBroadcast` too, see https://github.com/foundry-rs/foundry/pull/11650 and if possible add a unit test?

crates/cheatcodes/src/evm/record_debug_step.rs
DaniPopes: you should store ref to the node and the step, and retrieve depth and such later

crates/cheatcodes/src/inspector.rs
zerosnacks: added in https://github.com/foundry-rs/foundry/pull/12094/commits/cb2aae663448635dd2f7ac3d371124ad36d64059, is think this is what you are looking for

crates/cheatcodes/src/evm/record_debug_step.rs
zerosnacks: addressed in https://github.com/foundry-rs/foundry/pull/12094/commits/74d36ad2387da98fb6ffaf5dad3d6db59c8062e9

The `+1` still appears to be necessary, not exactly clear to me why.

zerosnacks:
Pending final reviews @DaniPopes & @grandizzy

crates/cheatcodes/src/inspector.rs
grandizzy: just a note that we touch account now twice, when calling cheatcode and here when we apply prank

zerosnacks: good catch, looking at the code this should effectively be a noop but still good to be aware

PR #12108, opened 2025-10-15
refactor(forge): Deduplicate special function validation logic

vtjl10:
### **Description:**

This pull request refactors the `ContractRunner` to eliminate duplicated code for validating `setUp` and `afterInvariant` functions within the `run_tests` method.

#### **Changes**

- **Extracted a new helper function:** A new private method, `validate_special_function`, has been introduced. This function now encapsulates the shared logic for:
    - Verifying that no more than one special function (like `setUp`) is defined.

- **Simplified `run_tests`:** The main test running function now calls this new helper, making the control flow cleaner and more readable.

#### **Reasoning**

This refactoring adheres to the **DRY (Don't Repeat Yourself)** principle, which significantly improves code maintainability. By centralizing the validation logic, we make it easier to modify in the future and reduce the chance of introducing bugs.

### DIFF
--- a/crates/forge/src/runner.rs
+++ b/crates/forge/src/runner.rs
@@ -66,6 +66,8 @@ pub struct ContractRunner<'a> {
     tokio_handle: &'a tokio::runtime::Handle,
     /// The span of the contract.
     span: tracing::Span,
+    /// The start time of the test run.
+    start_time: Instant,
     /// The contract-level configuration.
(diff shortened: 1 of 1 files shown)


onbjerg (changes requested):
this clones a lot, and this can be very expensive for test contracts, because they are pretty big on average. this might degrade performance quite a bit.

PR #12112, opened 2025-10-15
fix(fmt): ensure no underscores in yul number literals

### LINKED ISSUE #12111: bug(fmt): yul number literals cannot have underscores
### Component

Forge

- [ ] Foundry
- [ ] Foundryup

_No response_

_No response_

_No response_

### Operating System

None

### Describe the bug

as reported in https://github.com/foundry-rs/foundry/issues/12076#issuecomment-3406892129, this breaks compilation:
```sh
Error: Compiler run failed:
Error (4828): Invalid number literal.
  --> src/Oko.sol:20:44:
   |
20 |             let doy := add(shr(11, add(mul(62_719, mod(add(month, 9), 12)), 769)), day)
```

0xrusowsky:
## Motivation

closes #12111

## Solution

ensure no underscores in yul number literals

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -133,9 +133,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-dyn-abi"
-version = "1.4.0"
+version = "1.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a6c2905bafc2df7ccd32ca3af13f0b0d82f2e2ff9dfbeb12196c0d978d5c0deb"
+checksum = "3fdff496dd4e98a81f4861e66f7eaf5f2488971848bb42d9c892f871730245c8"
 dependencies = [
  "alloy-json-abi",
  "alloy-primitives",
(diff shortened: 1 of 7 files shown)


crates/fmt/testdata/Yul/original.sol
```diff
@@ -93,7 +93,7 @@ contract Yul {
             // empty bytes
             mstore(0xe0, 0x80)
 
-            let s2 := call(sub(gas(), 5000), pair, 0, 0x7c, 0xa4, 0, 0)
+            let s2 := call(sub(gas(), 5_000), pair, 0, 0x7c, 0xa4, 0, 0)
```
zerosnacks, line 96: do i understand correctly that this is formatted like this as a kind of regression test showing that `5_000` in Yul gets correctly transformed to `5000`?

considering `5_000` would not compile per example in https://github.com/foundry-rs/foundry/issues/12111 is it correct to "auto-fix" this for the user?

0xrusowsky: since solc doesn't allow underscores in yul decimals, i guess before this wasn't an issue cause solang/solc would panic.

but since solar allows them, i figured we'd help users get rid of them.

tldr; if they write with underscore and run:
- `forge build` -> panic (same as before)
- `forge fmt` + `forge build` -> we save them the fix

0xrusowsky: this is a nice to have though, the actual fix is ensuring we don't format with underscores when the config is set to do so. Otherwise, as shown in the reported issue, we would "break" valid code

zerosnacks: I see, makes sense

PR #12115, opened 2025-10-15
fix(config): canonicalize ignored_file_paths in canonic_at

silvekkk:
While using `ignored_warnings_from` with relative paths in my config, I noticed the warnings weren't being filtered properly. Turns out `ignored_file_paths` wasn't being canonicalized with the project root in `Config::canonic_at`, so the relative paths couldn't match correctly.

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -5453,40 +5453,6 @@ mod tests {
     }
 
     #[test]
-    fn test_ignored_file_paths_normalization() {
-        // Test that "./" prefixes are stripped from ignored_warnings_from paths
-        figment::Jail::expect_with(|jail| {
-            jail.create_file(
-                "foundry.toml",
-                r#"
-                [profile.default]
-                ignored_warnings_from = ["src/ignored.sol", "./test/ignored.sol"]
-                "#,
-            )?;
-
-            let config = Config::load().unwrap();
-
-            // Simulate the normalization in create_project
-            let normalized: Vec<PathBuf> = config
-                .ignored_file_paths
-                .iter()
-                .map(|path| path.strip_prefix("./").unwrap_or(path).to_path_buf())
-                .collect();
-
-            // Both paths should work the same way (with or without ./ prefix)
-            assert_eq!(
-                normalized,
-                vec![
-                    PathBuf::from("src/ignored.sol"),
-                    PathBuf::from("test/ignored.sol"), // ./ prefix removed
-                ]
-            );
-
-            Ok(())
-        });
-    }
-
-    #[test]
     fn test_inheritance_with_different_profiles() {
         figment::Jail::expect_with(|jail| {
             // Create base config with multiple profiles
--- a/crates/forge/tests/cli/config.rs
+++ b/crates/forge/tests/cli/config.rs
@@ -1972,3 +1972,58 @@ Warning: Found unknown `foo` config for profile `default` defined in foundry.tom
 
 "#]]);
 });
+
+forgetest_init!(test_ignored_file_paths_normalization, |prj, cmd| {
+    fn gen_contract(name: &str) -> String {
+        let fn_name = name.chars().next().unwrap().to_lowercase().to_string() + &name[1..];
+        format!(
+            r#"
+contract {name} {{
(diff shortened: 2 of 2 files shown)


0xrusowsky (approved):

silvekkk:
> looks like you change broke test `can_compile_without_warnings_ignored_file_paths`. Please check the CI output

CI now passes—canonicalize ignored_file_paths in canonic_at for consistency, then convert back to relative paths when passing to the compiler (following the same pattern as update_libs).

0xrusowsky:
@silvekkk thanks for the PR!

the fix works, but i'm wondering if the roundtrip (relative → absolute → relative) is more complex than needed... could we skip the `canonic_at()` change and just normalize directly in `create_project()` instead?
```rs
  // i'm thinking smth like this
  .ignore_paths(
      self.ignored_file_paths
          .iter()
          .map(|path| {
              if path.is_absolute() {
                  path.strip_prefix(&self.root).unwrap_or(path).to_path_buf()
              } else {
                  path.strip_prefix("./").unwrap_or(path).to_path_buf()
              }
          })
          .collect::<Vec<_>>(),
  )
```

this would normalize paths without modifying how other config paths work. I think that `ignored_file_paths` needs to stay relative for matching against compiler output, whereas other paths (src, out, etc.) are kept absolute for internal use.

thoughts?

silvekkk:
> @silvekkk thanks for the PR!
> 
> the fix works, but i'm wondering if the roundtrip (relative → absolute → relative) is more complex than needed... could we skip the `canonic_at()` change and just normalize directly in `create_project()` instead?
> 
> ```rust
>   // i'm thinking smth like this
>   .ignore_paths(
>       self.ignored_file_paths
>           .iter()
>           .map(|path| {
>               if path.is_absolute() {
>                   path.strip_prefix(&self.root).unwrap_or(path).to_path_buf()
>               } else {
>                   path.strip_prefix("./").unwrap_or(path).to_path_buf()
>               }
>           })
>           .collect::<Vec<_>>(),
>   )
> ```
> 
> this would normalize paths without modifying how other config paths work. I think that `ignored_file_paths` needs to stay relative for matching against compiler output, whereas other paths (src, out, etc.) are kept absolute for internal use.
> 
> thoughts?

@0xrusowsky Thanks I've simplified the implementation - removed the canonicalization from canonic_at() and now normalize paths directly in create_project() without the unnecessary roundtrip. you can check here https://github.com/foundry-rs/foundry/pull/12115/commits/5ab108564a31170fa68f5f8bc2de7deec14e341b

silvekkk:
> LGTM now.
> 
> @silvekkk can we add a unit test to validate what u claimed in the PR desc?

Sure added

0xrusowsky:
fyi @silvekkk to make reviewing easier and increasing the likelyhood of ur PRs being approved (basically by increasing the confidence that merging new code will work as expected) you always want to add tests + when possible e2e tests that validate the full functionality being impl/patched

0xrusowsky (approved):

silvekkk:
> fyi @silvekkk to make reviewing easier and increasing the likelyhood of ur PRs being approved (basically by increasing the confidence that merging new code will work as expected) you always want to add tests + when possible e2e tests that validate the full functionality being impl/patched

May I ask when should I add e2e test? sometime I really confused which new part should add a e2e test.

0xrusowsky:
> May I ask when should I add e2e test? sometime I really confused which new part should add a e2e test.

generally when a unit test is not enough to validate the whole functionality being changed. 
in this case, the unit test was only validating the normalization, but we want to make sure that the cmd will work as expected. 

as a rule of thumb, when there is a patch, we always want to have a repro (a test that reproduces the bug and that now is fixed). if a unit test is not enough to do that, u want to add an e2d test

silvekkk:
> > May I ask when should I add e2e test? sometime I really confused which new part should add a e2e test.
> 
> generally when a unit test is not enough to validate the whole functionality being changed. in this case, the unit test was only validating the normalization, but we want to make sure that the cmd will work as expected.
> 
> as a rule of thumb, when there is a patch, we always want to have a repro (a test that reproduces the bug and that now is fixed). if a unit test is not enough to do that, u want to add an e2d test

Thanks for the explanation ❤️

PR #12128, opened 2025-10-16
fix(evm): forward selfdestruct to inner inspectors in InspectorStackRefMut

GarmashAlex:
## Motivation

The InspectorStack was not propagating SELFDESTRUCT events to its inner inspectors. Although InspectorStack implemented selfdestruct by delegating to as_mut(), InspectorStackRefMut did not override the selfdestruct hook, so the default no-op from the Inspector trait was used. As a result, inner inspectors that implement selfdestruct (e.g., CustomPrintTracer) never received the event.

## Solution

This change adds a selfdestruct implementation to InspectorStackRefMut that forwards the hook to all inner inspectors via the existing call_inspectors! macro (tracer, printer, cheatcodes, log_collector, line_coverage, edge_coverage, fuzzer, script_execution_inspector, chisel_state, revert_diag). This aligns the stack with AnvilInspector’s behavior and with how other hooks (step, call, create, log, etc.) are already propagated.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -2,10 +2,10 @@
 
 [profile.default]
 retries = { backoff = "exponential", count = 2, delay = "5s", jitter = true }
-slow-timeout = { period = "1m", terminate-after = 3 }
+slow-timeout = { period = "30s", terminate-after = 3 }
 
 [[profile.default.overrides]]
-filter = "test(/ext_integration|can_test_forge_std/)"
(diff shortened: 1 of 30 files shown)


DaniPopes (approved):
thanks, i made it so only custom printer is delegated. this matches the other methods which only delegate when necessary

PR #12129, opened 2025-10-16
Fix GCP env var names

MamunC0der:
Motivation
wallet.rs read wrong GCP env vars (GCP_KEYRING, GCP_NAME), causing GCP KMS init to fail despite correct config.
Inconsistent naming with docs and other modules (GCP_KEY_RING, GCP_KEY_NAME).
Local var key_names was plural while holding a single value.

Solution
Use GCP_KEY_RING and GCP_KEY_NAME in wallets/src/wallet.rs.
Rename key_names → key_name in wallets/src/multi_wallet.rs.
Aligns env usage


onbjerg (approved):
this a bug so while its kinda breaking there's not really much we can do about it

PR #12144, opened 2025-10-17
fix(anvil): always disable nonce check for `eth_call`s

### LINKED ISSUE #12077: anvil: disable nonce check in `eth_call`
### Component

Anvil

Now in `build_call_env`, we would check if nonce exists, is it any switch to ignore input nonce even though it exists.
Since in some framework like wagmi, it would sometimes pass nonce in `eth_call`, and most rpc node would ignore this.

https://github.com/foundry-rs/foundry/blob/21e7ebcb5613bb578a6d8ed6b62b38dceb37def8/crates/anvil/src/eth/backend/mem/mod.rs#L1608-L1613 

### Additional context

_No response_

grandizzy:
## Motivation
- closes #12077 to be inline with https://github.com/ethereum/go-ethereum/blob/380688c636a654becc8f114438c2a5d93d2db032/core/state_transition.go#L145-L148

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -1520,7 +1520,7 @@ impl Backend {
     ///  - `disable_eip3607` is set to `true`
     ///  - `disable_base_fee` is set to `true`
     ///  - `tx_gas_limit_cap` is set to `Some(u64::MAX)` indicating no gas limit cap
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):

zerosnacks (approved):
Makes sense!

PR #12145, opened 2025-10-17
perf(evm): reduce rpc calls when instantiating a fork environment

DaniPopes:
- if no block number is passed we first get the latest block number, then fetch the block associated with it. just fetch the latest block directly.
- `gas_price` and `chain_id` are unused if passed in as an argument


grandizzy (approved):
nice!

PR #12150, opened 2025-10-17
feat(cheats): get profile artifacts metadata

0xrusowsky:
## Motivation

towards https://github.com/foundry-rs/foundry/issues/9840#issuecomment-3404827530

foundry is unable to fully support multichain deployments within a single script/run if the target chain have different EVM versions.

this is a summary of the issues/challenges that we face:
- [x] scripts run with a single `solc` & `evm_version` configuration. #12014 introduced the ability to select the evm version at runtime with `setEvmVersion()`
- [ ] the runtime evm version is not enough, we also need to deploy bytecode compiled with the corresponding target version, which requires accessing the multiple artifacts at runtime

## Solution
this PR introduces `getProfile()` cheatcodes that forwards profile metadata (artifacts path and EVM version) to the script runner, which is the last missing piece.

combined with `setEvmVersion()`, and `deployCode()` cheatcode, this enables:

1. **`getProfile()`**: get the evm version and locate the correct artifacts directory for each chain.
2. **`setEvmVersion()`**: switch to the required runtime environment
3. **`deployCode()`**: deploy the bytecode of the corresponding artifacts

### Implementation Details
```solidity
struct ProfileMetadata {
    /// The output path of the generated artifacts.
    string artifacts;
    /// The EVM version.
    string evm;
}

/// Derives the `ProfileMetadata` from the `foundry.toml` for the active profile.
function getProfile() external returns (ProfileMetadata memory);

/// Derives the `ProfileMetadata` from the `foundry.toml` for the given profile.
/// Reverts if the profile does not exist or the path cannot be resolved.
function getProfile(string calldata profile) external returns (ProfileMetadata memory);
```

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [ ] Breaking changes

## Next Steps

leverage the new cheatcodes in forge-std so that users can have a smooth and safe multi-chain experience (simply by inheriting the `Config` abstract).


zerosnacks:
Ran into a test failure, maybe a regression?

Ref: https://github.com/foundry-rs/foundry/actions/runs/18844750249/job/53782078338

```
thread 'tests::link_create2_nested' panicked at crates/linking/src/lib.rs:453:21:
    assertion `left == right` failed: incorrect library address for dependency default/linking/nested/Nested.t.sol:Lib of default/linking/nested/Nested.t.sol:LibraryConsumer
      left: 0xa201b394ef61ad72e744f23d62930b8552763ed3
     right: 0x773253227cce756e50c3993ec6366b3ec27786f9
```

zerosnacks:
Overall looks good :+1:

DaniPopes:
I think that test's addresses depend on the source code, for now you can just update the address but we should look into allowing any address

emmajam:

PR #12152, opened 2025-10-17
feat(traces): add contract context to Etherscan compilation errors

silvekkk:
When using cast run --debug to debug transactions, if one of the Etherscan contracts fails to compile, it's really hard to tell which one caused the problem. 
The compiler output is super long and scrolls past the initial "Compiling: ContractName" messages.

Adds the contract index and name directly to the error message so you can immediately know which contract is the problem.

Before(which contract? no idea):
```
Before:
Compiling: UniswapV2Router02 0x7a25...
Compiling: WETH9 0xC02a...
Compiling: UniswapV2Pair 0xB4e1...
Compiling: SomeOldContract 0x1234...
Compiling: AnotherContract 0x5678...

Error: compilation failed
Error (3853): Stack too deep...
```

After(ah, it's the 4th one - SomeOldContract):
```
Compiling [1/5]: UniswapV2Router02 0x7a25...
Compiling [2/5]: WETH9 0xC02a...
Compiling [3/5]: UniswapV2Pair 0xB4e1...
Compiling [4/5]: SomeOldContract 0x1234...
Compiling [5/5]: AnotherContract 0x5678...

Error: Failed to compile contract [4/5]: SomeOldContract at 0x1234...
Error (3853): Stack too deep...
```

### DIFF
--- a/crates/evm/traces/src/identifier/external.rs
+++ b/crates/evm/traces/src/identifier/external.rs
@@ -78,35 +78,31 @@ impl ExternalIdentifier {
         let contracts_info: Vec<_> = self
             .contracts
             .iter()
-            .filter(|(_, (_, metadata))| {
-                metadata.as_ref().is_some_and(|metadata| !metadata.is_vyper())
+            // filter out vyper files and contracts without metadata
+            .filter_map(|(addr, (_, metadata))| {
+                if let Some(metadata) = metadata.as_ref()
+                    && !metadata.is_vyper()
+                {
+                    Some((*addr, metadata))
+                } else {
+                    None
+                }
             })
-            .map(|(addr, (_, metadata))| (*addr, metadata.as_ref().unwrap().contract_name.clone()))
             .collect();
 
-        let outputs_fut = self
-            .contracts
+        let outputs_fut = contracts_info
             .iter()
-            // filter out vyper files
-            .filter(|(_, (_, metadata))| {
-                metadata.as_ref().is_some_and(|metadata| !metadata.is_vyper())
-            })
-            .map(|(address, (_, metadata))| {
-                let addr = *address;
-                let metadata = metadata.as_ref().unwrap().clone();
-                async move {
-                    sh_println!("Compiling: {} {addr}", metadata.contract_name)?;
-                    let root = tempfile::tempdir()?;
-                    let root_path = root.path();
-                    let project = etherscan_project(&metadata, root_path)?;
-                    let output = project.compile()?;
-
-                    if output.has_compiler_errors() {
-                        eyre::bail!("{output}")
-                    }
-
-                    Ok((project, output, root))
+            .map(|(addr, metadata)| async move {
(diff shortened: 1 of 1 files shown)


onbjerg (changes requested):
please check ci & can we remove the enumeration? (x/y)

silvekkk:
> please check ci & can we remove the enumeration? (x/y)

Sure, removed

onbjerg (dismissed):

silvekkk:
Is the CI failed in merge queue? looks this was removed from merge queue, let me know if I need to fix anything cc @DaniPopes @onbjerg

crates/evm/traces/src/identifier/etherscan.rs
```diff
@@ -72,20 +73,24 @@ impl EtherscanIdentifier {
...
-            .map(|(address, metadata)| async move {
-                sh_println!("Compiling: {} {address}", metadata.contract_name)?;
-                let root = tempfile::tempdir()?;
-                let root_path = root.path();
-                let project = etherscan_project(metadata, root_path)?;
-                let output = project.compile()?;
+            .map(|(address, metadata)| {
+                let contract_name = metadata.contract_name.clone();
+                let addr = *address;
+                let metadata_clone = metadata.clone();
```
0xrusowsky: i believe that it should be possible the achieve the same outcome with just the metadata clone, which then gives u access to an already cloned contract name.

also, i'd rather shadow `metadata` than naming it `metadata_clone`

0xrusowsky (changes requested):

crates/evm/traces/src/identifier/etherscan.rs
silvekkk: Make sense, removed now, plz have a check

silvekkk:
Hi @0xrusowsky rebased on latest master. 
Applied same changes to external.rs after #11917 refactoring moved the code from etherscan.rs. 
And in this new version romoved the redundant clone and rename `metadata` cc @onbjerg

0xrusowsky:
actually, since `fn etherscan_project` takes a metadata ref, you don't need to clone at all.
also since u need to pre-collect the metadata in a vec, you can save some work by directly iterating over the filtered vec.

silvekkk:
> actually, since `fn etherscan_project` takes a metadata ref, you don't need to clone at all. also since u need to pre-collect the metadata in a vec, you can save some work by directly iterating over the filtered vec.

Got it,thanks! this this new to me, nice to learn a new trick!

0xrusowsky:
> Got it,thanks! this this new to me, nice to learn a new trick!

@silvekkk sure, np! that's why we try to explain what can be improved.
thanks for taking the time to open a PR

onbjerg:
>Is the CI failed in merge queue? looks this was removed from merge queue, let me know if I need to fix anything cc @DaniPopes @onbjerg

yeah the global aws outage yesterday affected gh & depot (who we use for gh runners) so we were unable to merge anything yesterday. we might still not be able to today but i'll try regardless

PR #12162, opened 2025-10-17
ci: disable Windows on PRs

DaniPopes:
With merge queues we don't need to run all tests on Windows in CI on every PR update, since any failure will still block merging.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,6 +7,7 @@ on:
     branches:
       - master
   pull_request:
+  merge_group:
 
 concurrency:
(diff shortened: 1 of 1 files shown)


grandizzy:
I think that's OK, would that slow down on feedback to dev or should not worry?

DaniPopes:
Yeah but the assumption is that the majority of changes won't affect other OS's differently than Linux, so the faster feedback loop here outweighs the slower one where they fail because of related changes

grandizzy (approved):
makes sense, let's do it!

Issue #11932, opened 2025-10-02
flamegraphs and other features are broken by function selector collisions


moodysalem:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

forge Version: 1.3.5-stable

### What version of Foundryup are you on?

foundryup: 0.3.1

### What command(s) is the bug in?

forge test --flamegraph

### Operating System

macOS (Apple Silicon)

### Describe the bug

if two functions in the repository have the same function selector, e.g. because the zero function selector was mined for functions on 2 different contracts, the flamegraph and other tools will incorrectly label one of the functions even if it's not present on the contract where it was called

grandizzy:
thank you @moodysalem mind to provide a repro to debug and fix?

marcvernet31:
@grandizzy I think I managed to reproduce the bug. As described above, on situations where there are two contracts with functions with the same signature, `forge test` confuses them.

Using as example functions:

```bash
cast sig "transferFrom(address,address,uint256)"
# Output: 0x23b872dd

cast sig "gasprice_bit_ether(int128)"
# Output: 0x23b872dd

``` 

````solidity

pragma solidity ^0.8.3;


contract ContractA {
    event FunctionCalled(string name, bytes4 selector);
    
    // Selector: 0x23b872dd
    function transferFrom(address from, address to, uint256 value) public returns (bool) {
        emit FunctionCalled("ContractA.transferFrom", bytes4(keccak256("transferFrom(address,address,uint256)")));
        return true;
    }
}

contract ContractB {
    event FunctionCalled(string name, bytes4 selector);
    
    // DIFFERENT function but SAME selector: 0x23b872dd
    function gasprice_bit_ether(int128 value) public returns (bool) {
        emit FunctionCalled("ContractB.gasprice_bit_ether", bytes4(keccak256("gasprice_bit_ether(int128)")));
        return true;
    }
}
// Contract that calls both to demonstrate the collision
contract CollisionCaller {
    function callBoth(address a, address b) public returns (bytes4, bytes4) {
        // Both of these calls use the SAME selector
        bytes4 selectorA = bytes4(keccak256("transferFrom(address,address,uint256)"));
        bytes4 selectorB = bytes4(keccak256("gasprice_bit_ether(int128)"));
        
        // Make the calls
        (bool successA,) = a.call(abi.encodeWithSelector(selectorA, address(0), address(0), uint256(100)));
[...]

PR #12169, opened 2025-10-18
fix(forge): Avoid trace confusion on functions with shared signature

marcvernet31:
## Motivation
Closes https://github.com/foundry-rs/foundry/issues/11932

## Solution
When selecting functions based on signatures, on case of having several functions with the same signature the first one was always selected, leading to cases where the function behind the signature was confused. With the new changes, `select_contract_function` checks that the contract the signsture comes from matches, avoiding the confusion.
Added unit test for it.
## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/evm/traces/src/decoder/mod.rs
```diff
@@ -271,6 +271,33 @@ impl CallTraceDecoder {
...
+        trace: &CallTrace,
+    ) -> &'a [Function] {
+        if functions.len() <= 1 {
+            return functions;
+        }
+
+        // When there are selector collisions, try to decode the calldata with each function
+        // to determine which one is actually being called. The correct function should
+        // decode successfully while the wrong ones will fail due to parameter type mismatches.
+        if functions.len() > 1 {
```
grandizzy, line 285: nit: this is not needed since we already return early for `functions.len() <= 1` or we can keep it and remove the early return

marcvernet31: updated

grandizzy (approved):
thank you, lgtm

PR #12171, opened 2025-10-18
chore: eliminate redundant test_fail iterator clone

Bilogweb3:
Precompute the test_fail functions so we reuse the same vector for the result generation, eliminating the iterator clone while keeping behaviour identical


DaniPopes:
This is worse

PR #12178, opened 2025-10-19
fix(anvil): remove transactions before block/hash in unwind_to to prevent leaks

sashass1315:
## Motivation

Transactions from rolled-back blocks were not removed from self.transactions due to the order of operations in unwind_to.
The code first removed the hashes entry and the block from blocks, then called remove_block_transactions_by_number, which relies on hashes and blocks to still contain the entry. As a result, mined transactions of those blocks were left in the transactions index, causing memory leaks.

## Solution

- In unwind_to, remove transactions by iterating the block’s transactions while the block is still available, then remove the number→hash mapping, and finally collect the block for return.
- This makes cleanup independent of hashes/blocks state during the iteration and ensures transactions are always removed.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/anvil/src/eth/backend/mem/storage.rs
```diff
@@ -337,11 +337,19 @@ impl BlockchainStorage {
...
+                    for tx in &block.transactions {
+                        self.transactions.remove(&tx.hash());
+                    }
+                    // Now drop the number -> hash mapping
+                    self.hashes.remove(&i);
+                    removed.push(block);
+                } else {
+                    // Ensure we still drop the number -> hash mapping if the block is gone
+                    self.hashes.remove(&i);
+                }
```
onbjerg, line 347: let's make use of `remove_block_transactions_by_number` first and then remove the mappings after instead of inlining the functionality of that function

sashass1315: > let's make use of `remove_block_transactions_by_number` first and then remove the mappings after instead of inlining the functionality of that function

ok

PR #12180, opened 2025-10-19
fix(invariant): revisit sequence indices and respect shrink limit

0xalpharush:
## Motivation
@rappie reported the shrinking was finishing very quickly: https://discord.com/channels/1193917206141091910/1207683521435664404/1429482097294315572

 I found that it was only indexing through the sequence once and not revisiting the sequence until it reached the shrinking limit. You can tell the shrink limit isn't reached bc the progress bar doesn't have a chance to appear for short sequences with a high limit configured. 

## Solution
Reset the call index to attempt removal again. This is not a perfect solution bc the index could have already been removed. I can work on this in follow-up PR

## PR Checklist

Tested on https://github.com/aviggiano/shrinking-mwe/ and after this PR, the shrinking progress bar shows up as expected with this PR
```
[invariant]
runs = 5000000
depth = 100
shrink_run_limit = 1000000
show_solidity = true
show_metrics = true
corpus_dir = "foundry-corpus"
```

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


0xalpharush:
I think the tests are failing bc this regex isn't picking up when a sequence shrinks, but I'm not sure 
https://github.com/foundry-rs/foundry/blob/027ed5f46947e09e028677bd3394865665d0beae/crates/forge/tests/cli/test_cmd/invariant/mod.rs#L11-L11

DaniPopes:
you just have to update the number

For `test_cmd::invariant::target::filters` i guess the sequence isn't printed anymore?

crates/evm/evm/src/executors/invariant/shrink.rs
```diff
@@ -96,7 +48,7 @@ pub(crate) fn shrink_sequence(
 
     // Reset run count and display shrinking message.
     if let Some(progress) = progress {
-        progress.set_length(min(calls.len(), failed_case.shrink_run_limit as usize) as u64);
+        progress.set_length(failed_case.shrink_run_limit as usize as u64);
```
DaniPopes:
```suggestion
        progress.set_length(failed_case.shrink_run_limit as u64);
```

0xalpharush:
Lowering the default or just disabling in some tests may help. For UX, the counter does get shown and show it's shrinking, so the user shouldn't just think something froze. 

I don't think linearly searching through a call sequence of length N once can necessarily shrink it maximally. Sometimes you can't remove an earlier index until you've moved a later one and setting `shrink_run_limit` > N allows that. Does that make sense to you as well @grandizzy ?

grandizzy:
> Lowering the default or just disabling in some tests may help. For UX, the counter does get shown and show it's shrinking, so the user shouldn't just think something froze.

:+1: 

> 
> I don't think linearly searching through a call sequence of length N once can necessarily shrink it maximally. Sometimes you can't remove an earlier index until you've moved a later one and setting `shrink_run_limit` > N allows that. Does that make sense to you as well @grandizzy ?

Yeah, was a basic / naive implementation that was added just to make sure the test is completed as fast as possible. Makes sense, let's then merge this PR and follow up with other improvements if needed, would you agree @0xalpharush ?

PR #12193, opened 2025-10-20
fix: Simplify equal-case formatting in `fmt_change` by using `to_string()`

maximevtush:
### Description
- Replace `format!("{change}")` with `change.to_string()` in `crates/forge/src/cmd/snapshot.rs` (`fmt_change`, `Ordering::Equal` branch).
- Removes redundant formatting without changing behavior.

#### Rationale
- Improves clarity and avoids unnecessary `format!` machinery when plain `ToString` suffices.
- Aligns with idiomatic Rust and keeps logic consistent across branches.


crates/forge/src/cmd/snapshot.rs
```diff
@@ -543,7 +543,7 @@ fn fmt_change(change: i128) -> String {
     match change.cmp(&0) {
         Ordering::Less => format!("{change}").green().to_string(),
         Ordering::Equal => {
-            format!("{change}")
+            change.to_string()
         }
```
onbjerg:
```suggestion
Ordering::Equal => change.to_string(),
```

PR #12200, opened 2025-10-20
refactor: remove redundant nested function in fd_lock

maradini77:
## Summary

Removed unnecessary nested function `new_lock` that was shadowing the outer function name and creating redundant indirection.


onbjerg (changes requested):
this is a common trick in rust to reduce the bloat from monomorphization

PR #12201, opened 2025-10-20
feat: add interval support to ProviderBuilder and preserve mining interval in anvil_setRpcUrl

CreeptoGengar:
Add interval() method to ProviderBuilder to allow setting custom poll intervals, which resolves TODO comment in anvil_set_rpc_url function.


CreeptoGengar:
my apologies, the root problem seems to be much harder than I thought, thank you for your time in helping me

PR #12205, opened 2025-10-20
fix: prevent panic in format_help when categories is empty

hawkadrian:
Added safety check to prevent panic when accessing empty categories vector in format_help function. 

Commands without help headings now get grouped under a default "General" category instead of causing a crash.


DaniPopes:
this is never empty

PR #12212, opened 2025-10-21
doc(cheats): `deployCode` reverts w/ unlinked artifact libs

0xrusowsky:
## Motivation

while working on the forge-std impl that will build on top of https://github.com/foundry-rs/foundry/pull/12150 i realized i wanted to verify that when devs use `deployCode` they can't shoot themselves in the foot by deploying broken bytecode from an artifact with an unresolved placeholder. When checking the impl, i realized that this is already the case, but that we weren't explicitly documenting nor testing it.

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [ ] Breaking changes


grandizzy (approved):
👍

PR #12215, opened 2025-10-21
fix: simplify string references in argument passing

strmfos:
Clean up unnecessary string method chaining by leveraging Rust's automatic deref coercion from &String to &str. The .as_str() calls after .to_string() are redundant since Rust automatically handles this conversion in function argument contexts.


DaniPopes:
This doesn't matter at all

PR #12218, opened 2025-10-21
fix(fmt): correct peek_comment_between to scan comments within range

Galoretka:
## Motivation

The function used take_while with a range predicate starting from the first comment in the file. This stops iteration as soon as a comment before pos_lo is encountered, causing the function to almost always return None when earlier comments exist.
Lists relying on this check (e.g., commasep’s has_comments computation) could mis-detect “no comments between”, affecting line breaking and spacing decisions.

## Solution

Replace the logic with skip_while(c.pos() <= pos_lo) + take_while(c.pos() < pos_hi) and find the first non-blank. This matches the intended semantics and aligns with has_comment_between’s filtering approach.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/fmt/src/state/mod.rs
+++ b/crates/fmt/src/state/mod.rs
@@ -1190,3 +1190,47 @@ fn style_doc_comment(style: DocCommentStyle, mut cmnt: Comment) -> Comment {
         _ => cmnt,
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use solar::parse::interface::source_map::FileName;
+
+    // Repro: when there is a comment before `pos_lo`, the buggy take_while-based
+    // implementation of `peek_comment_between` fails to find a later comment
+    // that is actually within (pos_lo, pos_hi).
+    #[test]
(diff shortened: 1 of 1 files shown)


grandizzy (changes requested):
can you pls add a test with the issue this fixes? Thank you

0xrusowsky:
looks AI generated, but also a legit bug/patch.
i'll try to come up with a repro myself if they don't answer

Galoretka:
> looks AI generated, but also a legit bug/patch. i'll try to come up with a repro myself if they don't answer

I will add tests if you don't mind

0xrusowsky:
@Galoretka great, thanks! 
i think that a unit test where u mock several cmnts will probably be easier to setup than finding an actual solidity repro, but do whatever u prefer

Galoretka:
> @Galoretka great, thanks! i think that a unit test where u mock several cmnts will probably be easier to setup than finding an actual solidity repro, but do whatever u prefer

I added a test

emmajam:

PR #12220, opened 2025-10-21
fix(cast): store canonical event signatures in selector cache

Forostovec:
- Switch selector cache writing from Event.full_signature() to Event.signature() to ensure canonical event signatures (without parameter names) are stored.
- This makes cached entries parsable by Event::parse and aligns with how functions/errors are stored, fixing the failing test cast::cli selectors::error_event_decode_with_cache and reducing flaky decoding due to non-canonical event strings.

### DIFF
--- a/crates/evm/traces/src/identifier/signatures.rs
+++ b/crates/evm/traces/src/identifier/signatures.rs
@@ -118,7 +118,7 @@ impl SignaturesCache {
                 Some((SelectorKind::Error(e.selector()), e.signature()))
             }
             alloy_json_abi::AbiItem::Event(e) => {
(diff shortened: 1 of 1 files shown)


onbjerg (changes requested):
I don't really see how this reduces flakiness at all. This just changes what is displayed in traces. Can you please provide a motivation for this change?

Forostovec:
> I don't really see how this reduces flakiness at all. This just changes what is displayed in traces.

I noticed that there is a failling test 'cast::cli selectors::error_event_decode_with_cache' so i thought maybe it would be useful to fix this, but obviously my fix isn't correct

PR #12221, opened 2025-10-21
feat: add support for sei mainnet and testnet

### LINKED ISSUE #12196: feat: add support for sei testnet
### Component

Forge

- [x] Foundry
- [x] Foundryup

1.4.2-stable

1.3.0

forge script

### Operating System

Linux

### Describe the bug

```
Error: Chain 1328 not supported
```
Image

silvekkk:
closes #12196

### DIFF
--- a/crates/cheatcodes/src/config.rs
+++ b/crates/cheatcodes/src/config.rs
@@ -706,6 +706,24 @@ fn create_default_chains() -> HashMap<String, ChainData> {
         },
     );
 
+    chains.insert(
+        "sei".to_string(),
+        ChainData {
+            name: "Sei".to_string(),
+            chain_id: 1329,
+            default_rpc_url: "https://evm-rpc.sei-apis.com".to_string(),
+        },
+    );
+
+    chains.insert(
+        "sei_testnet".to_string(),
+        ChainData {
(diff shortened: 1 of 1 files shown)


onbjerg:
can you verify this solves the issue? i don't think it does; I think it needs to be added to https://github.com/alloy-rs/chains/

grandizzy:
> can you verify this solves the issue? i don't think it does; I think it needs to be added to https://github.com/alloy-rs/chains/

yep, @silvekkk please see https://github.com/alloy-rs/chains/blob/main/src/named.rs#L8-L21 and sei PR for reference https://github.com/alloy-rs/chains/pull/160 We can then bump alloy chains in foundry

silvekkk:
> > can you verify this solves the issue? i don't think it does; I think it needs to be added to https://github.com/alloy-rs/chains/
> 
> yep, @silvekkk please see https://github.com/alloy-rs/chains/blob/main/src/named.rs#L8-L21 and sei PR for reference [alloy-rs/chains#160](https://github.com/alloy-rs/chains/pull/160) We can then bump alloy chains in foundry

Got it, working on it now

grandizzy:
closing per above

PR #12226, opened 2025-10-21
refactor: remove redundant `_non_mut` suffix from method names

kilavvy:
This PR removes the redundant `_non_mut` suffix from two public methods in `CheatsConfig` to improve API consistency and readability.

- Renamed `get_chain_data_by_alias_non_mut` → `get_chain_data_by_alias`
- Renamed `get_rpc_url_non_mut` → `get_rpc_url`
- Updated internal method call to use the new name

The `_non_mut` suffix was redundant since:
- Methods already use `&self` (non-mutating by design)
- Other methods in the same struct follow the pattern without such suffixes (`rpc_endpoint`, `rpc_urls`, `set_chain_data`)

### DIFF
--- a/crates/cheatcodes/src/config.rs
+++ b/crates/cheatcodes/src/config.rs
@@ -259,7 +259,7 @@ impl CheatsConfig {
     }
 
     /// Get chain data by alias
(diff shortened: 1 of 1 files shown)


onbjerg:
These functions are basically unused, removed in https://github.com/foundry-rs/foundry/pull/12245

PR #12230, opened 2025-10-21
ci: don't run most CI on push

DaniPopes:
Merge group status checks are reported on push. I believe there is no point in re-running everything again after the merge group is committed.

We only need to run `docs` to publish rustdoc to GH Pages.


DaniPopes:
hmm

> analyze (actions)
Unable to validate code scanning workflow: MissingPushHook
analyze (actions)
1 issue was detected with this workflow: Please specify an on.push hook to analyze and see code scanning alerts from the default branch on the Security tab.

grandizzy:
maybe @zerosnacks pls chime in?

DaniPopes:
Maybe we need to update the configuration here? https://github.com/foundry-rs/foundry/security/code-scanning/tools/CodeQL/status/

grandizzy:
> Maybe we need to update the configuration here? https://github.com/foundry-rs/foundry/security/code-scanning/tools/CodeQL/status/

I think is because the on push master is removed from ci.yml and needs to be on the default branch?

This warning means your GitHub Actions workflow is not configured to run code scanning on the default branch when a new commit is pushed.
To fix this, you need to add or modify the on.push trigger in your workflow file to include the default branch (e.g., main or master) so that security alerts are analyzed and appear in the Security tab

DaniPopes:
It might be fine, let's try merging and see if the warning goes away. I don't think it matters anyway since we still run it as normal on PRs and merge queue

PR #12235, opened 2025-10-21
test(config): add test for test_failures_file path normalization

silvekkk:
Adds e2e test verifying `test_failures_file` path normalization works with custom paths and `--rerun` from subdirectories.

Related to #12071


Yasersarhadikarbasak:
ممنون بسیار عالی

PR #12262, opened 2025-10-23
fix(common): spurious detection off-by-one

MozirDmitriy:
The spurious detection logic used the pre-increment value returned by fetch_add(1) to compare against max_timedout_requests, which delayed marking the connection as spurious until one additional failure beyond the configured threshold. This off-by-one meant the client would continue attempting network calls after the maximum allowed timeouts, degrading responsiveness and violating the intended behavior described in the code comments. The comparison was corrected to use the incremented count so the flag flips exactly on the Nth failure, and an async unit test was added to ensure the spurious state is set precisely at the threshold and not earlier.


grandizzy (approved):
makes sense, thank you!

PR #12263, opened 2025-10-23
fix(benchmark): remove unnecessary line continuation characters in export statements

anim001k:
The `benchmark.sh` script contains unnecessary line continuation characters (`\`) after `export` statements. These characters can cause unexpected behavior when the script is modified, as they indicate command continuation but are followed by empty lines, making the script fragile and harder to maintain.


DaniPopes:
Doesn't matter

PR #12264, opened 2025-10-23
refactor: eliminate code duplication in transaction encoding

oooLowNeoNooo:
Replaced repetitive match patterns in transaction and receipt encoding functions with reusable macros. This reduces code duplication while maintaining the same functionality and improving maintainability for future transaction type additions.


DaniPopes:
it's not any better

PR #12266, opened 2025-10-24
refactor: replace process::exit with Result errors in ensure_ok

DeVikingMark:
Replace TODO's `std::process::exit(1)` calls with proper Result error returns in TestOutcome::ensure_ok method.


onbjerg (changes requested):
the comment is a leftover and idt we should fix this, especially given this leads to double output

DeVikingMark:
@onbjerg may I create a PR with the deletion of TODO's?

PR #12296, opened 2025-10-27
docs: clean up config docs

onbjerg:
Looking at the config struct, I noticed some fields seem to no longer be used, so I marked them as potential removal candidates. I also cleaned up some of the comments.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -231,7 +231,6 @@ pub struct Config {
     /// Glob patterns for file paths to skip when building and executing contracts.
     pub skip: Vec<GlobMatcher>,
     /// Whether to forcefully clean all project artifacts before running commands.
-    // todo(onbjerg): is this useful anymore?
     pub force: bool,
     /// The EVM version to use when building contracts.
     #[serde(with = "from_str_lowercase")]
@@ -308,7 +307,6 @@ pub struct Config {
     /// List of solidity error codes to always silence in the compiler output.
     pub ignored_error_codes: Vec<SolidityErrorCode>,
     /// List of file paths to ignore.
-    // todo(onbjerg): how is this different from `skip`?
     #[serde(rename = "ignored_warnings_from")]
     pub ignored_file_paths: Vec<PathBuf>,
     /// Diagnostic level (minimum) at which the process should finish with a non-zero exit.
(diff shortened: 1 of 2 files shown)


crates/config/src/lib.rs
```diff
@@ -429,8 +434,10 @@ pub struct Config {
     #[serde(default)]
     pub extra_output_files: Vec<ContractOutputSelection>,
     /// Whether to print the names of the compiled contracts.
+    // todo(onbjerg): this seems unused.
     pub names: bool,
     /// Whether to print the sizes of the compiled contracts.
+    // todo(onbjerg): this seems unused.
     pub sizes: bool,
```
zerosnacks, line 436: These should not be unused; should correspond with the `forge build --sizes` / `forge build --names`  / `forge build --names --sizes` flags

I am not sure if this ever worked

https://github.com/morpho-org/morpho-optimizers/blob/04143d929372e53d3f6822f62ee90df1609fc784/foundry.toml#L9-L10

crates/config/src/lib.rs
```diff
@@ -397,6 +401,7 @@ pub struct Config {
     /// If this limit is exceeded, a `MemoryLimitOOG` result is thrown.
     ///
     /// The default is 128MiB.
+    // todo(onbjerg): this seems unused.
     pub memory_limit: u64,
```
zerosnacks: revm does allow configuring this and we are passing the value into it; must not be wired correctly

DaniPopes: it is, it goes through the serialize config->deserialize into evmopts dance

crates/config/src/lib.rs
```diff
@@ -302,9 +305,10 @@ pub struct Config {
...
     #[serde(default, skip_serializing_if = "EtherscanConfigs::is_empty")]
     pub etherscan: EtherscanConfigs,
-    /// list of solidity error codes to always silence in the compiler output
+    /// List of solidity error codes to always silence in the compiler output.
     pub ignored_error_codes: Vec<SolidityErrorCode>,
-    /// list of file paths to ignore
+    /// List of file paths to ignore.
+    // todo(onbjerg): how is this different from `skip`?
     #[serde(rename = "ignored_warnings_from")]
     pub ignored_file_paths: Vec<PathBuf>,
```
DaniPopes, line 311: this is for ignoring warnings https://github.com/foundry-rs/compilers/blob/799f83ed4a1b67e39428106f51128d1c8f891f8c/crates/compilers/src/compile/output/mod.rs#L831

crates/config/src/lib.rs
DaniPopes: actually it's broken but for a different reason, we set it correctly in cfg env https://github.com/foundry-rs/foundry/pull/12299

PR #12303, opened 2025-10-27
feat(fmt): add single_line_imports option to keep single imports on one line

### LINKED ISSUE #12251: feat(fmt): add config to keep single imports in a single line
### Component

Forge

The pre v1.4.0 forge formatter did not force line wrapping on imports over the line length config. It would be nice if this could be restored via a config option.

Pre v.1.4.0, this line was not wrapped despite being 121 characters, 1 more than the default limit:
```solidity
import { ITransparentUpgradeableProxy } from "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
```

Post v1.4.3, this line now gets wrapped like so:
```solidity
import {
    ITransparentUpgradeableProxy
} from "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
```

As far as I know, imports were the only lines that didn't obey the line length wrapping. They look rather ugly when forced to be wrapped. A config option to exclude imports from wrapping would be nice. I understand that technically I am requesting a config to reenable what could be considered to be a bug, so no worries if this just gets closed. We will deal with it. But style-wise, this looks bad when its only one contract/interface/library being imported. It'd be a different story if it was multiple imports from a single file.

### Additional context

You can see the differences the new formatter has imposed on our code compared to the old one in this PR: https://github.com/omni-network/omni/pull/4449/files

silvekkk:
Closes #12251

Adds a `single_line_imports` config option (default `false`) so single imports don't get wrapped even when they exceed `line_length`.

Before (when exceeding line length):
```solidity
import {
    ITransparentUpgradeableProxy
} from "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
```

After (with `single_line_imports = true`):
```solidity
import {ITransparentUpgradeableProxy} from "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
```

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -137,7 +137,7 @@ jobs:
         run: ./.github/scripts/commit-and-read-benchmarks.sh benches "${{ github.event_name }}" "${{ github.repository }}"
 
       - name: Upload benchmark results as artifacts
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
           name: benchmark-results
           path: |
@@ -165,7 +165,7 @@ jobs:
           persist-credentials: false
 
       - name: Download benchmark results
-        uses: actions/download-artifact@v5
+        uses: actions/download-artifact@v6
         with:
           name: benchmark-results
           path: benches/
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -166,7 +166,7 @@ jobs:
         with:
           toolchain: stable
       - uses: rui314/setup-mold@725a8794d15fc7563f59595bd9556495c0564878 # v1
-      - uses: taiki-e/install-action@e43a5023a747770bfcb71ae048541a681714b951 # v2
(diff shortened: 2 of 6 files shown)


crates/fmt/src/state/sol.rs
```diff
@@ -215,24 +215,51 @@ impl<'ast> State<'_, 'ast> {
...
+                        self.print_commasep_aliases(aliases.iter());
+                    };
 
-                self.braces_break();
-                self.s.offset(-self.ind);
-                self.word("}");
-                self.end();
-                self.word(" from ");
-                self.print_ast_str_lit(path);
+                    self.braces_break();
```
0xrusowsky, line 246: have u tried simply doing:
```rs
if !use_single_line {
    self.brace_break();
}
```
on both breaks?

your proposal seems too verbose for how simple it should be

silvekkk: Yeah, simplify it just now

0xrusowsky:
@grandizzy unrelated CI fail (tried 3 times)

PR #12327, opened 2025-10-28
ci: use sccache in release workflow

DaniPopes:
We can't use GHA cache since it gets evicted every day


grandizzy:
Cc @zerosnacks

PR #12335, opened 2025-10-28
test: make forgetest_init! init an empty project

DaniPopes:
Closes https://github.com/foundry-rs/foundry/issues/11625

There are a lot more tests that want forge-std but with an empty project (`wipe_contracts`), than with a `forge init` default project since we usually dont want to run the Counter tests


mattsse (approved):
🫡

PR #12349, opened 2025-10-29
chore: simplify calldata assignment

Sharilleed223:
Removes redundant `.0` tuple accessor when assigning calldata to `CallInput::Bytes`. The `calldata` field already implements `Into` trait, making the tuple access unnecessary.


onbjerg:
https://github.com/foundry-rs/foundry/blob/master/CONTRIBUTING.md#contributions-assisted-by-ai

PR #12343, opened 2025-10-28
fix(fmt): properly calc fn header size

0xrusowsky:
## Motivation

closes #12109, replacing https://github.com/foundry-rs/foundry/pull/12110 which was flawed.

Last week, when i submitted the patch, i didn't see that 'function ' was computed in  `fn estimate_header_params_size`. The real issue was that i missed 2 characters on ' returns(..)' + the end (opening brace ' {' or semicolon ';').

Unfortunately the mistake slipped through the unit tests + as it didn't show in the diffs, the reviewers missed it too.

## Solution

fix the calculation

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [ ] Breaking changes

PR #12350, opened 2025-10-29
test(fmt): ensure fn header sizes are computed correctly

0xrusowsky:
## Motivation

after yesterday's patch when calculating fn header sizes #12343, i thought it would be better to add unit tests to make sure we cover all cases

## Solution

- add unit tests to make sure that the size of all sorts of fn headers are properly computed.
- although unlikely to break a line due to its shorter nature, i realized i always assuming the keyword was `function` but it can also be `modifier`, `constructor`, `fallback` or `receive`.
- although it is barely used, `virtual` was never handled

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/core/src/state_snapshot.rs
+++ b/crates/evm/core/src/state_snapshot.rs
@@ -60,9 +60,8 @@ impl<T> StateSnapshots<T> {
     /// Inserts the new state snapshot at the given `id`.
     ///
     ///  Does not auto-increment the next `id`.
-    pub fn insert_at(&mut self, state_snapshot: T, id: U256) -> U256 {
+    pub fn insert_at(&mut self, state_snapshot: T, id: U256) {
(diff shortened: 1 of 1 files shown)


crates/fmt/src/state/sol.rs
```diff
@@ -2554,49 +2552,66 @@ impl<'ast> State<'_, 'ast> {
...
-        let end = if has_body { 2 } else { 1 };
+        let end = if body.is_some() { 2 } else { 1 };
 
-        self.estimate_header_params_size(header) // accounts for 'function name(..)'
+        self.estimate_header_params_size(func)
             + visibility
             + mutability
             + modifiers
             + override_
+            + virtual_
```
grandizzy, line 2591: 👍

0xrusowsky:
yes imo it is good as is 🙂

PR #12353, opened 2025-10-29
ci: fix release workflow

DaniPopes:
Fixes https://github.com/foundry-rs/foundry/issues/12348.


grandizzy (approved):
🙌

grandizzy:
@DaniPopes release still fails here https://github.com/foundry-rs/foundry/actions/runs/18902827668/job/53953906907#step:8:79
do we need
```
  with:
    tool:
```
for taiki-e/cache-cargo-install-action too?

l.e. ah, actually it fails for musl releases that use cross

PR #12366, opened 2025-10-29
feat: warn on soldeer.lock revision mismatch during build

### LINKED ISSUE #12357: feat: Add warning on `forge build` with inconsistent `rev` in foundry.lock file
Add a warning/fail on `forge build` if the lock file has different `rev` than what's checked out.

silvekkk:
Adds warnings during `forge build` when lockfile state doesn't match actual dependencies:

- **soldeer.lock**: Uses `soldeer_core` APIs to check dependency integrity (missing files, failed checksums)
- **foundry.lock**: Compares git submodule revisions against lockfile entries

Also warns when `soldeer.lock` exists but fails to parse.

Closes #12357

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -3,6 +3,8 @@
 [profile.default]
 retries = { backoff = "exponential", count = 2, delay = "5s", jitter = true }
 slow-timeout = { period = "30s", terminate-after = 3 }
+# Exclude flaky tests from regular CI runs - they run in nightly test-flaky workflow
+default-filter = "not test(/flaky_/)"
 
 [[profile.default.overrides]]
 filter = "test(/ext_integration/)"
@@ -12,3 +14,10 @@ slow-timeout = { period = "5m", terminate-after = 4 }
 [[profile.default.overrides]]
 filter = "package(foundry-cheatcodes-spec)"
 retries = 0
+
+# Profile for running flaky tests (used by nightly CI and for local debugging)
+# Run with: cargo nextest run --profile flaky
+[profile.flaky]
+default-filter = "test(/flaky_/)"
+retries = { backoff = "exponential", count = 5, delay = "10s", jitter = true }
+slow-timeout = { period = "60s", terminate-after = 3 }
--- a/.github/FLAKY_TEST_FAILURE_TEMPLATE.md
+++ b/.github/FLAKY_TEST_FAILURE_TEMPLATE.md
@@ -1,0 +1,10 @@
+---
+title: "bug: flaky tests workflow failed"
+labels: P-normal, T-bug
+---
+
+The nightly flaky tests workflow has failed. This indicates external API rate limiting, RPC reliability issues, or other intermittent failures that may affect users.
+
+Check the [flaky tests workflow page]({{ env.WORKFLOW_URL }}) for details.
+
+This issue was raised by the workflow at `.github/workflows/test-flaky.yml`.
--- a/.github/workflows/test-flaky.yml
+++ b/.github/workflows/test-flaky.yml
@@ -1,13 +1,17 @@
-# Daily CI job to run tests with isolation mode enabled by default
+# Daily CI job to run flaky tests that are excluded from regular CI via nextest default-filter
 
-name: test-isolate
+name: test-flaky
 
 permissions: {}
 
 on:
   schedule:
-    - cron: "0 0 * * *" # Run daily at midnight UTC
+    - cron: "0 1 * * *" # Run daily at 1 AM UTC (offset from test-isolate)
   workflow_dispatch: # Needed so we can run it manually
+
+concurrency:
+  group: tests-${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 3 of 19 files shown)


grandizzy:
@silvekkk I think that could work too (cc @mario-eth ) but the original ticket was meant for foundry.lock and deps installed with `forge install` and updated with `forge update`

grandizzy:
updated ticket name to reflect that. thank you!

mario-eth:
cc @beeb

beeb:
Ideally, this check would be more thorough and do the same consistency check we do in soldeer install (hashing all the files in the dependencies). It would be best if it reused the logic and types from soldeer_core too.

silvekkk:
- Add SHA256 checksum verification for dependencies
- Check both checksum and git revision
- Provide clear warnings for integrity failures
- Follow soldeer's verification approach

is this what in your mind? @beeb  cc @grandizzy @mario-eth

beeb:
@silvekkk You reimplemented everything which is not great, because if we ever decide to change something about the lockfile or dependencies folder structure, your code will break.

Please add a dependency to `soldeer_core` in `forge/Cargo.toml` using `soldeer-core.workspace = true`.

You can then use the following:
- https://docs.rs/soldeer-core/0.9.0/soldeer_core/lock/fn.read_lockfile.html
- https://docs.rs/soldeer-core/0.9.0/soldeer_core/install/fn.check_dependency_integrity.html

silvekkk:
@beeb Thanks for the feedback! I'vereverted and refactored to use the soldeer_core APIs:

- Added `soldeer-core` dependency
- Using `read_lockfile()` and `check_dependency_integrity()` as you suggested
- Removed all custom implementation

crates/forge/src/cmd/build.rs
```diff
@@ -228,6 +209,50 @@ impl BuildArgs {
...
+                        sh_warn!("Dependency '{}' integrity check failed: {:?}", dep_name, status)
+                            .ok();
+                        continue;
+                    }
+                }
+                Err(e) => {
+                    sh_warn!("Dependency '{}' integrity check error: {}", dep_name, e).ok();
+                    continue;
+                }
+            }
```
beeb, line 267: You should also handle the `DependencyStatus::FailedIntegrity` variant here, as it indicates the dependency is present but the file structure or commit hash do not match the lockfile reference hash.

crates/forge/src/cmd/build.rs
```diff
@@ -228,6 +209,50 @@ impl BuildArgs {
...
+
+            // For git dependencies, check revision from source URL
```
beeb: This comment makes no sense

crates/forge/src/cmd/build.rs
```diff
@@ -228,6 +209,50 @@ impl BuildArgs {
...
+        };
+
+        let deps_dir = config.root.join("dependencies");
+        for entry in &lockfile.entries {
+            let dep_name = entry.name();
+            let dep_path = deps_dir.join(format!("{}-{}", dep_name, entry.version()));
+
+            if !dep_path.exists() {
+                continue;
+            }
```
beeb: This check is not necessary, as it's already performed by `check_dependency_integrity` which returns `DependencyStatus::Missing` if the folder is not there.

beeb (changes requested):

beeb:
Thanks for the changes. Although this all feels very LLM generated, I left some comments.

silvekkk:
> Thanks for the changes. Although this all feels very LLM generated, I left some comments.

Yeah, about 50–60% of the later commits were actually done with Claude’s help. I asked it to add comments, clean up the structure, and handle a few checks I probably overlooked, and final review.
Haha, trust me—if you’d seen my original code, you probably wouldn’t even have felt like leaving a comment. (It was that bad.) best 20$ I paid ever

silvekkk:
@grandizzy LGTY?

grandizzy:
> @grandizzy LGTY?

@silvekkk yeah, makes sense. Can we add same check for foundry.lock file too (that was the originally ticket was opened for). Thanks!

silvekkk:
> > @grandizzy LGTY?
> 
> @silvekkk yeah, makes sense. Can we add same check for foundry.lock file too (that was the originally ticket was opened for). Thanks!

Sure added `check_foundry_lock_consistenc`

crates/forge/src/cmd/build.rs
```diff
@@ -4,7 +4,7 @@ use eyre::{Context, Result};
 use forge_lint::{linter::Linter, sol::SolidityLinter};
 use foundry_cli::{
     opts::{BuildOpts, configure_pcx_from_solc, get_solar_sources_from_compile_output},
-    utils::{LoadConfig, cache_local_signatures},
+    utils::{Git, LoadConfig, cache_local_signatures},,
```
zerosnacks:
```suggestion
    utils::{Git, LoadConfig, cache_local_signatures},
```

silvekkk:
@zerosnacks  @grandizzy Hey mates!  Is this good to go?

zerosnacks (approved):

zerosnacks:
lgtm, pending @grandizzy check

PR #12367, opened 2025-10-29
fix(config): make ResolvedEtherscanConfigs::find_chain ignore unrelated errors

phrwlk:
## Motivation

The previous implementation of ResolvedEtherscanConfigs::find_chain returned the first error it encountered, even if it was unrelated to the requested chain. This contradicted the function’s documentation (“Returns the first config that matches the chain”) and created order-dependent, spurious failures when a valid config for the requested chain existed later in the map.

## Solution

This change updates the logic to:
- Return Ok(config) for the first entry that matches the requested chain.
- Otherwise remember the first error and return it only if no matching config exists.
- Return None if there is neither a match nor an error.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/config/src/etherscan.rs
+++ b/crates/config/src/etherscan.rs
@@ -138,14 +138,19 @@ impl ResolvedEtherscanConfigs {
         self,
         chain: Chain,
     ) -> Option<Result<ResolvedEtherscanConfig, EtherscanConfigError>> {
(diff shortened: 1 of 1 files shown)


DaniPopes:
There shouldn't be any errors, if there are they should be reported and program aborts . This is not correct

PR #12372, opened 2025-10-29
fix(cast): sign erc20 transactions locally before broadcast

### LINKED ISSUE #12368: `cast erc20 transfer` does not sign the tx
### Component

Cast

### Describe the bug

Using `cast erc20 transfer` the tx is not signed, instead it defers to the node, which is incorrect.

### Additional context

_No response_

0xrusowsky:
## Motivation

closes #12368

## Solution

note: i moved the `sol!` macro with ERC20 out of cast + added a new helper `fn signing_provider` so that we can keep things organized if we add more token subcommands down the line.

## PR Checklist

- [ ] Added Tests
- [X] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cast/src/cmd/erc20.rs
+++ b/crates/cast/src/cmd/erc20.rs
@@ -1,11 +1,9 @@
 use std::str::FromStr;
 
-use crate::format_uint_exp;
+use crate::{format_uint_exp, tx::signing_provider};
 use alloy_eips::BlockId;
 use alloy_ens::NameOrAddress;
-use alloy_network::EthereumWallet;
 use alloy_primitives::U256;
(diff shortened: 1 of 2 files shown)


crates/cast/src/cmd/erc20.rs
```diff
@@ -112,60 +117,79 @@ pub enum Erc20Subcommand {
 }
 
 impl Erc20Subcommand {
-    pub async fn run(self) -> eyre::Result<()> {
+    // Helper to get the rpc and be able to de-duplicate code.
+    // Acceptable to match twice as performance is not critical in cast + there are few commands.
```
DaniPopes: pls remove comment

Issue #12361, opened 2025-10-29
Forge doc is using full path instead of relative path


smol-ninja:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

1.4.3-stable

### What version of Foundryup are you on?

1.3.0

### What command(s) is the bug in?

forge doc

### Operating System

macOS (Apple Silicon)

### Problem

`forge doc` produces the following content:

```md
[ISablierFlowState](/Users/shubham/Sablier/flow/docs/src/src/interfaces/ISablierFlowState.sol/interface.ISablierFlowState.md)
```

Whereas, with `1.3.6`, it produces the following content:

```md
**Inherits:**
[ISablierFlowState](/src/interfaces/ISablierFlowState.sol/interface.ISablierFlowState.md)
```

The latest version use full path for hyperlinks.

### Expected Behaviour

It should not use full path for hyperlinks.

PR #12026, opened 2025-10-08
feat(wallets): add Turnkey signer support

rplusq:
## Summary
- Add Turnkey as a remote signer option alongside AWS KMS and GCP KMS
- Fix wallet feature flags in forge to properly gate foundry-wallets dependency

## Changes
- Added `alloy-signer-turnkey` dependency to `foundry-wallets`
- Implemented Turnkey wallet CLI args and initialization logic  
- Added `turnkey` feature flags to cast and forge
- Updated Dockerfile and Makefile to include turnkey in release builds

## Bug Fix: forge wallet features
Made `foundry-wallets` an optional dependency in `forge/Cargo.toml` and updated feature definitions.

**Issue**: The `aws-kms`, `gcp-kms`, and `turnkey` features were defined but non-functional when using `cargo run --features forge/turnkey`. The features only worked in release builds via Makefile.

**Root cause**: `foundry-wallets` was a dev-dependency, so feature flags had no effect. Additionally, it was unconditionally compiled in the default build despite being feature-gated, causing unused dependency warnings.

**Fix**: 
- Made `foundry-wallets` an optional dependency: `foundry-wallets = { workspace = true, optional = true }`
- Updated features to enable the dependency: `turnkey = ["dep:foundry-wallets", "foundry-wallets/turnkey"]`
- Added conditional import guard in `lib.rs` to only use the crate when features are enabled

This aligns forge with cast's wallet handling and resolves clippy warnings about unused dependencies.

## Testing
```bash
# Before: --turnkey flag doesn't appear
cargo run --bin forge --features forge/turnkey -- create -h | grep turnkey
# (no output)

# After: --turnkey flag works
cargo run --bin forge --features forge/turnkey -- create -h | grep turnkey  
# --turnkey  Use Turnkey

# CI checks pass
cargo check --all                                          # ✓
cargo +nightly fmt -- --check                              # ✓
cargo +nightly clippy --all --all-targets --all-features -- -D warnings  # ✓
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12373, opened 2025-10-29
fix(forge doc): use relative path instead of full path

marcvernet31:
## Motivation

Fix issue described here: https://github.com/foundry-rs/foundry/issues/12361

## Solution

Fix forge doc hyperlinks to use relative paths
- Generate relative links like /src/... instead of absolute filesystem paths
- Ensure consistent path handling across builder, document, inheritance, and writer

Added a test on `doc.rs` to ensure paths are no longer relative. 

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/doc/src/document.rs
```diff
@@ -69,7 +69,7 @@ impl Document {
     }
 
     fn try_relative_output_path(&self) -> Option<&Path> {
-        self.target_path.strip_prefix(&self.out_target_dir).ok()?.strip_prefix(DocBuilder::SRC).ok()
```
grandizzy, line 72: I think we still need this, as docs target paths are composed prefixed with SRC here:

https://github.com/foundry-rs/foundry/blob/87a024e99d57d4fccea382b325be4c33ed396df3/crates/doc/src/builder.rs#L194

https://github.com/foundry-rs/foundry/blob/87a024e99d57d4fccea382b325be4c33ed396df3/crates/doc/src/builder.rs#L219

https://github.com/foundry-rs/foundry/blob/87a024e99d57d4fccea382b325be4c33ed396df3/crates/doc/src/builder.rs#L247

so it's a doubled `src` e.g. target path is `/home/flow/docs/src/src/interfaces/IFlowNFTDescriptor.sol/interface.IFlowNFTDescriptor.md` and out target is `/home/flow/docs` hence relative path will result in doubled src. Or we could just not join src in target dirs, maybe that's better

marcvernet31: Added `strip_prefix`back there.

DaniPopes:
This doesn't pass Windows tests: https://github.com/foundry-rs/foundry/actions/runs/19144190665/job/54717330057#step:14:1252

Reverting in https://github.com/foundry-rs/foundry/pull/12373

Same as https://github.com/foundry-rs/foundry/pull/12026#issuecomment-3498449193

Please don't force merge PRs anymore

PR #12377, opened 2025-10-30
fix: synchronize spec_id and is_initialized in CowBackend

Lil-Duckling-22:
Ensures backend reinitialization when EVM specification changes between inspect calls. 
Previously, spec_id could change without triggering backend reinitialization, potentially causing compatibility issues during fuzzing with different EVM versions.


DaniPopes:
i don't think this is right

PR #12380, opened 2025-10-30
feat(cast): allow logs filtering on multiple addresses

0xferrous:
## Motivation

Enables multi address filters.

## Solution

Accept multiple addresses. Usage is as follows:
```
$ cast logs "SomeEvent()" --address emitter.eth --address 0xdead...dead --rpc-url mainnet
```

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cast/src/cmd/logs.rs
+++ b/crates/cast/src/cmd/logs.rs
@@ -27,11 +27,8 @@ pub struct LogsArgs {
     to_block: Option<BlockId>,
 
     /// The contract address to filter on.
-    #[arg(
-        long,
-        value_parser = NameOrAddress::from_str
-    )]
-    address: Option<NameOrAddress>,
+    #[arg(long, value_parser = NameOrAddress::from_str)]
+    address: Option<Vec<NameOrAddress>>,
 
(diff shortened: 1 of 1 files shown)


crates/cast/src/cmd/logs.rs
```diff
@@ -101,7 +103,7 @@ impl LogsArgs {
 fn build_filter(
     from_block: Option<BlockNumberOrTag>,
     to_block: Option<BlockNumberOrTag>,
-    address: Option<Address>,
+    address: Option<Vec<Address>>,
```
grandizzy, line 106: nit, can we rename to addresses

0xferrous: it becomes a breaking change if we do that, and also `--addresses <address> --addresses <address>` would be a bit weird

grandizzy: ah, right, missed that, think it's ok then, looking for one more opinion before merging

grandizzy (approved):

0xrusowsky (approved):
makes sense, thanks!

Issue #12095, opened 2025-10-14
feat(`cast`): add `cast erc20` with various subcommands like `balance`, `send`, etc; deprecate `--erc20` flag on `cast balance`.


onbjerg:
### Component

Cast

### Describe the feature you would like

Currently, `cast balance` has an `--erc20` flag for querying ERC20 token balances. However, this approach doesn't scale well for other ERC20 operations, and adding similar flags to other commands (like `cast send`) would create an inconsistent and ambiguous interface.

**Proposed Solution:**

Introduce a dedicated `cast erc20` command with various subcommands for common ERC20 operations:
  - `cast erc20 balance <TOKEN> <ACCOUNT>` - Query ERC20 token balance (replaces `cast balance
  --erc20`)
  - `cast erc20 transfer <TOKEN> <TO> <AMOUNT>` - Transfer tokens using `transfer(address,uint256)`
  - `cast erc20 approve <TOKEN> <SPENDER> <AMOUNT>` - Approve token spending using
  `approve(address,uint256)`
  - `cast erc20 allowance <TOKEN> <OWNER> <SPENDER>` - Query allowance using
  `allowance(address,address)`

zerosnacks:
I'm assuming `--erc20` would refer here to `transfer(address,uint256)`?

Do we also need a shorthand for approving a token: `approve(address,uint256)`?

It opens a bit of a rabbit hole

onbjerg:
yes to first q, for second q, I'm not sure, but it's a good question. we could possibly detect this and notify the user.

marcvernet31:
@onbjerg Can I pick this up?

zerosnacks:
high-level aligned on adding a `cast erc20` command that has various subcommands (`--balance`, `--send`, etc) rather than an ambigious `--erc20` on commands like `cast send`

0xrusowsky:
@marcvernet31 still want to pick it up?

marcvernet31:
@0xrusowsky yesss

0xrusowsky:
@marcvernet31 i've updated the issue description based on what was discussed above.

please also make sure to also add a warning msg to the old `--erc20` flag on `cast balance`,  prompting users to use the new cmd (we will eventually remove the deprecated flag in a future major release).

0xClandestine:
May be nice to future proof this for the addition of future ercs.

`cast erc <number> <flags>`

Though I'll admit my suggested syntax isn't great.

[...]

PR #12381, opened 2025-10-30
feat(cast): more erc20 methods

0xrusowsky:
## Motivation

#12095 follow-up

## Solution

adds the rest of the erc20 methods and the common `mint` and `burn` extensions

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cast/src/cmd/erc20.rs
+++ b/crates/cast/src/cmd/erc20.rs
@@ -27,7 +27,6 @@ sol! {
         function transfer(address to, uint256 amount) external returns (bool);
         function approve(address spender, uint256 amount) external returns (bool);
         function allowance(address owner, address spender) external view returns (uint256);
-        function transferFrom(address from, address to, uint256 amount) external returns (bool);
         function mint(address to, uint256 amount) external;
         function burn(uint256 amount) external;
     }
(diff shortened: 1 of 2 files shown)


crates/cast/src/cmd/erc20.rs
```diff
@@ -19,10 +19,17 @@ sol! {
...
         #[derive(Debug)]
+        function name() external view returns (string);
+        function symbol() external view returns (string);
+        function decimals() external view returns (uint8);
+        function totalSupply() external view returns (uint256);
         function balanceOf(address owner) external view returns (uint256);
         function transfer(address to, uint256 amount) external returns (bool);
         function approve(address spender, uint256 amount) external returns (bool);
         function allowance(address owner, address spender) external view returns (uint256);
+        function transferFrom(address from, address to, uint256 amount) external returns (bool);
```
onbjerg: idt this one make sense, iirc it would mean you had approved an eoa to pull funds from you, i've never seen that before

0xrusowsky: yeah good call, probably not worth having it

js26h2cd8s-eng: Next time it would be good

PR #12383, opened 2025-10-30
chore: remove redundant clones in block construction and bloom building

phrwlk:
This change optimizes crates/anvil/src/eth/backend/executor.rs by removing unnecessary allocations without altering behavior. It moves the transactions vector into Block::new instead of cloning since it is not used afterwards and Block::new already performs the necessary internal cloning for transaction root computation. It also updates build_logs_bloom to accept &[Log] and adjusts the call site to pass &logs, avoiding a separate Vec<Log> clone used solely for iteration. Together these changes reduce per-block and per-transaction allocations while maintaining identical semantics.


zerosnacks (approved):
Thanks, looks good to me

PR #12384, opened 2025-10-30
Remove redundant hex prefix check in parse_value_fallback

anim001k:
This PR removes a duplicate manual check for the "0x" hex prefix in `parse_value_fallback`.  
Hex string validation and normalization are now fully centralized via utility methods, making the extra `starts_with("0x")` check unnecessary and potentially divergent.


DaniPopes:
this is not unnecessary

PR #12393, opened 2025-10-31
perf(anvil): avoid unnecessary transaction vector cloning in Block::new

CreeptoGengar:
Replace vector clone with direct field access when calculating transaction root.

Instead of cloning the entire Vec<MaybeImpersonatedTransaction>, we now iterate over references and clone only the inner TypedTransaction field. 

This reduces memory usage, especially for blocks with many transactions.


mattsse (approved):
we can even get rid of the entire clone+collect

PR #12396, opened 2025-10-31
fix(anvil): preserve OP receipt fields in convert_to_anvil_receipt

### LINKED ISSUE #12345: bug(anvil): The returned receipts in fork mode for on-chain txs are missing Optimism fields
### Component

Anvil

- [x] Foundry
- [x] Foundryup

_No response_

_No response_

_No response_

### Operating System

None

### Describe the bug

When running anvil in fork mode, w/ optimism enabled:
```bash
anvil --fork-url http://archive.optimism fork-block-number 131878000 --optimism
```

and fetching a receipt
```bash
cast receipt 0x15149db6cb54d41b6d81ab2d51854817af4eced452d1439123a0680315417934
```

It misses the Optimism fields (L1 Gas, etc.).

While 
```bash
cast receipt -r http://archive.optimism 0x15149db6cb54d41b6d81ab2d51854817af4eced452d1439123a0680315417934
```

works fine.

Syzygy106:
## Motivation

I decided to help Foundry community and fix the following bug: https://github.com/foundry-rs/foundry/issues/12345
Closes #12345

## Solution

**Anvil now preserves Optimism's L1 gas fields when forking**

When running Anvil in fork mode against Optimism, receipt fields like `l1Fee`, `l1GasPrice`, and `l1GasUsed` were getting lost. These fields are Optimism-specific and don't exist in standard Ethereum receipts, so they didn't fit into Anvil's `TransactionReceipt` type.

**The problem:**
Anvil's `ReceiptResponse` was just a plain `TransactionReceipt<TypedReceiptRpc>`, which meant any extra JSON fields from the upstream node got silently dropped. Optimism stores its L1 gas data in an `other` map, but `convert_to_anvil_receipt` was throwing it away.

**The fix:**
- Changed `ReceiptResponse` to `WithOtherFields<TransactionReceipt<TypedReceiptRpc>>` so Anvil keeps unknown fields instead of discarding them
- Updated `convert_to_anvil_receipt` to preserve the `other` fields
- Added `.inner` access where needed (like in `eth/api.rs`)
- Unwrapped `WithOtherFields` when converting to Otterscan API types (to maintain API compatibility while preserving fields internally)
- Made sure locally-built receipts also use `WithOtherFields` with an empty map for consistency
- Added a test that creates a synthetic Optimism receipt and verifies all L1 fields survive the conversion

**Why it works:**
`WithOtherFields<T>` from alloy-serde does exactly what we need—it preserves unknown JSON keys without touching the core schema. Now Anvil keeps all of Optimism's extra receipt data intact. Regular Ethereum receipts without these fields still work exactly as before.

**Notes for reviewers:**

1. I didn't add documentation since this fix brings the behavior in line with what users would naturally expect. If you think docs are needed somewhere, happy to add them—just let me know where.

2. I changed the response type and updated all the places where it's unpacked. I've run the tests and builds, and everything looks good, but I'd appreciate another set of eyes on the downstream logic—just in case I missed something that might be affected by the type change.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/tests/it/optimism.rs
+++ b/crates/anvil/tests/it/optimism.rs
@@ -3,7 +3,7 @@
 use crate::utils::{http_provider, http_provider_with_signer};
 use alloy_eips::eip2718::Encodable2718;
 use alloy_network::{EthereumWallet, TransactionBuilder};
-use alloy_primitives::{Address, TxHash, TxKind, U256, b256, Bloom};
+use alloy_primitives::{Address, Bloom, TxHash, TxKind, U256, b256};
 use alloy_provider::Provider;
 use alloy_rpc_types::TransactionRequest;
 use alloy_serde::WithOtherFields;


crates/anvil/tests/it/optimism.rs
```diff
@@ -224,3 +225,55 @@ async fn test_deposit_tx_checks_sufficient_funds_after_applying_deposited_value(
...
 }
+
+#[test]
+fn preserves_op_fields_in_convert_to_anvil_receipt() {
+    fn zero_bloom_hex() -> String {
+        let mut s = String::with_capacity(2 + 512);
+        s.push_str("0x");
+        s.extend(std::iter::repeat_n('0', 512));
+        s
+    }
```
mattsse: I think we can use Bloom::Zero here instead

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -990,7 +990,7 @@ impl Backend {
                 // Keep previous `beneficiary` and `basefee` value
                 beneficiary: env.evm_env.block_env.beneficiary,
                 basefee: env.evm_env.block_env.basefee,
-                ..Default::default()
+                ..env.evm_env.block_env.clone()
```
mattsse: I think this is unrelated?

can we exclude this from this pr?

crates/anvil/core/src/eth/transaction/mod.rs
```diff
@@ -1484,51 +1508,56 @@ pub fn convert_to_anvil_receipt(receipt: AnyTransactionReceipt) -> Option<Receip
...
-                },
-                logs_bloom: receipt_with_bloom.logs_bloom,
-            }),
-            _ => return None,
+                    logs_bloom: receipt_with_bloom.logs_bloom,
+                }),
+                _ => return None,
+            },
         },
+        other,
```
mattsse, line 1560: I see, this is the fix here

crates/anvil/src/eth/backend/mem/mod.rs
Syzygy106: > I think this is unrelated?
> 
> can we exclude this from this pr?

Sure. There are no Optimism related fields. Forgot to clear

crates/anvil/tests/it/optimism.rs
Syzygy106: > I think we can use Bloom::Zero here instead

Sure, that's more suitable

mattsse (approved):
great

mattsse:
@Syzygy106 just needs cargo +nightly fmt --all

Syzygy106:
> @Syzygy106 just needs cargo +nightly fmt --all

got it

mattsse (approved):

PR #12405, opened 2025-10-31
chore(anvil): remove dead BeaconResult alias and re-export

Forostovec:
Remove unused BeaconResult<T> alias from eth/beacon/response.rs and drop its re-export from eth/beacon/mod.rs.
The alias is not used anywhere in the repository, and web searches show no external usage. Keeping it adds public API surface without benefit and can cause confusion. This aligns with existing usage patterns where only actively used result aliases (e.g., NodeResult<T>) are kept.


mattsse (approved):
yeah we dont need these

PR #12409, opened 2025-11-01
fix(cast): validate args when sig is missing in access-list

avorylli:
Add validation to prevent silent argument ignoring when function signature is not provided. Previously, arguments were ignored without any error message if sig was None, which led to confusing behavior.

Now the command fails early with a clear error message when args are provided without a signature.


DaniPopes:
This is not right, no signature or args means no data

Issue #12408, opened 2025-11-01
Sending `SIGTERM` signal to Anvil installed as a Node.js package does not terminate Anvil


guidanoli:
### Component

Anvil

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

anvil Version: 1.4.4-v1.4.4 Commit SHA: 05794498bf47257b144e2e2789a1d5bf8566be0e Build Timestamp: 2025-10-30T15:50:28.714367747Z (1761839428) Build Profile: maxperf

### What version of Foundryup are you on?

Does not apply (Anvil was installed through Node.js package `@foundry-rs/anvil`)

### What command(s) is the bug in?

anvil (through `@foundry-rs/anvil` package)

### Operating System

Linux

### Describe the bug

There are some programs that spawn Anvil, store its PID, send JSON-RPC requests to it, and finally terminates Anvil by sending a `SIGTERM` signal through its previously-stored PID. This method works when Anvil is installed through Foundryup, but not (currently) when it is installed as a Node.js package.

One such program is the [Cannon](https://github.com/usecannon/cannon) CLI, which uses Anvil to build a development network state. The `cannon build` command follows the aforementioned recipe (see [Cannon CLI source](https://github.com/usecannon/cannon/blob/v2.25.1/packages/cli/src/rpc.ts#L107)), and fails to terminate Cannon if Anvil had been installed as a Node.js package.

After some investigation, this is what I have discovered: When Cannon sends a `SIGTERM` signal to Anvil, it is not actually sending it to the Anvil binary, but rather to a Node.js script `bin.mjs`, which, in turn, spawns Anvil as a child process:

```
node bin.mjs
└── anvil
```

(Simplified paths and arguments for ease of reading.)

When a `SIGTERM` signal is sent to Node.js, which is running the `bin.mjs` script, it doesn't propagate the signal to Anvil, its child process, leaving it running in the background. You can easily reproduce this bug like so:

```sh
mkdir nodejs-anvil-bug                          # create a tmp directory
pushd nodejs-anvil-bug                          # go to such directory
pnpm init                                       # initialize a pnpm project
pnpm add -D @foundry-rs/{anvil,cast}            # install Anvil and Cast as Node.js pkgs
pnpm exec anvil &                               # run Anvil on the background
pid=$!                                          # store the process ID
while ! pnpm cast chain-id ; do sleep 1 ; done  # wait for Anvil to spin up
[...]

PR #12412, opened 2025-11-01
fix(npm): kill child process when SIGTERM or SIGINT is received

guidanoli:
## Motivation

See #12408 for context and motivation.

## Solution

Adding handlers for `SIGTERM` and `SIGINT` signals that kill the child process.
I have tested this locally both on Linux and Windows, but not on macOS.
I would appreciate some guidance on how to add tests for this bugfix.
Do I need to add documentation for this? If so, where?
I am not sure this PR introduces any breaking changes.

### DIFF
--- a/npm/src/bin.mjs
+++ b/npm/src/bin.mjs
@@ -28,7 +28,14 @@ if (!platformPackage) {
   process.exit(1)
 }
 
-NodeChildProcess.spawn(selectBinaryPath(), process.argv.slice(2), { stdio: 'inherit' })
+const child = NodeChildProcess.spawn(
+    selectBinaryPath(),
+    process.argv.slice(2),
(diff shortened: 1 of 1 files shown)


npm/src/bin.mjs
```diff
@@ -117,3 +124,10 @@ function selectBinaryPath() {
 
   return NodePath.join(__dirname, '..', 'dist', binaryName)
 }
+
+/**
+ * Kills the child process.
+ */
+function killChild() {
+    child.kill()
+}
```
mattsse, line 133: we basically do the same in the rust wrapper type for launching anvil as process and this is working fine

onbjerg:
ping @o-az

dbeal-eth (approved):

PR #12414, opened 2025-11-01
docs: add system requirements section to benchmarks README

crStiv:
### Description 

Adds a System Requirements section documenting minimum Rust and Node.js versions, disk space requirements, and expected execution times. 
This helps users understand prerequisites before running benchmarks and avoids setup issues.

Fixes markdown lint error with [:rev] format in repo specification examples.


onbjerg (changes requested):
doesn't really matter, as long as you are not outdated by a couple of years you should be fine.

PR #12421, opened 2025-11-02
Remove redundant allocation when resolving script ABI functions

Doryu001:
Replace the intermediate collect::<Vec<_>>() with single-pass iterator checks in get_method_and_calldata. Keep identical error semantics while eliminating unnecessary allocation and iteration.


DaniPopes:
Doesn't matter

Issue #12385, opened 2025-10-30
feat(cast): `block -f` to accept multiple fields


grandizzy:
### Component

Cast

### Describe the feature you would like

specify multiple block fields to list e.g. `cast block -f timestamp,number`

dipanshuhappy:
Hey would like to work on this, I am getting a hang of the repo after working on ```cast send```

mattsse:
cool, assigned @dipanshuhappy

dipanshuhappy:
@grandizzy @mattsse  two ways i can do this. 
1. I change the args with Vec
2. I parse the fields string into "," or "-".
The first one will cause changes in tests and other calls of the block function, the second one is a quick fix rather than a proper fix.

Edit: I meant Vec

mattsse:
@dipanshuhappy I think reasonable here would be a Vec for multiple args, either delimiter but also just separated by whitespace should work I think

dipanshuhappy:
> @dipanshuhappy I think reasonable here would be a Vec for multiple args, either delimiter but also just separated by whitespace should work I think

That approach would change inputs to the function and also little refactor in tests. If thats acceptable then i can go ahead with that.

mattsse:
yep that sounds fine

grandizzy:
https://github.com/foundry-rs/foundry/pull/12427

PR #12427, opened 2025-11-03
feat(cast): block -f to accept multiple fields

dipanshuhappy:
## Motivation

Resolvs #12385 

## Solution

Basically I changed it ```field``` to ```fields```, type to ```Vec<String>```. Then loops over the fields and appends the output to the final std out stream.
## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/cast/src/opts.rs
```diff
@@ -379,11 +379,11 @@ pub enum CastSubcommand {
         block: Option<BlockId>,
 
         /// If specified, only get the given field of the block.
-        #[arg(long, short)]
-        field: Option<String>,
+        #[arg(short, long,num_args = 1..,action = ArgAction::Append,value_delimiter = ',')]
+        fields: Vec<String>,
```
onbjerg: please make sure this is renamed to `--field` in `cast` using `long = "field"` so this is not breaking

dipanshuhappy: Got it, on it

dipanshuhappy: > please make sure this is renamed to `--field` in `cast` using `long = "field"` so this is not breaking

Pushed the change

grandizzy (dismissed):
Please fix failing tests, thank you

dipanshuhappy:
> Please fix failing tests, thank you

Thanks for letting me know on that, I had the assumption that I shouldn't edit existing test cases, but I have pushed some 2 edited test cases, that were failing due to whitespace.

onbjerg (approved):

PR #12429, opened 2025-11-03
fix(debugger): remove redundant stack length check in draw_buffer

CreeptoGengar:
Remove redundant `stack_len > 0` check before calling `get_buffer_accesses`.
The function already handles empty stacks safely, and the pattern without the length check is used consistently elsewhere in the codebase (lines 488-490).


DaniPopes:
dont think thats right and i would rather not change this code

PR #12430, opened 2025-11-03
refactor(cast): remove redundant clone and duplicate import

Bilogweb3:
Two small cleanups in cast:

1. In lib.rs, encode_function_args_packed was imported twice - once separately on its own line and once in the main foundry_common import block. Moved it to be with the other abi imports where it belongs.

2. In wallet list command, we were cloning self.dir just to unwrap it. Changed to as_deref() since we only need a reference - keystore_path is just used for is_empty() and canonicalize(), both of which work fine with &str. Saves an unnecessary string allocation.

### DIFF
--- a/crates/cast/src/lib.rs
+++ b/crates/cast/src/lib.rs
@@ -35,7 +35,6 @@ use foundry_common::{
 };
 use foundry_config::Chain;
 use foundry_evm::core::bytecode::InstIter;
-pub use foundry_evm::*;
 use futures::{FutureExt, StreamExt, future::Either};
 use op_alloy_consensus::OpTxEnvelope;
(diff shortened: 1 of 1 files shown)


crates/cast/src/lib.rs
```diff
@@ -49,9 +50,6 @@ use std::{
 };
 use tokio::signal::ctrl_c;
 
-use foundry_common::abi::encode_function_args_packed;
-pub use foundry_evm::*;
```
onbjerg: please don't reorder the `pub use`

Bilogweb3:
@onbjerg done, that was clippy suggestion

Bilogweb3:
strange, it’s already gone :)

PR #12434, opened 2025-11-03
fix: proper error handling for multichain broadcast parsing

hawkadrian:
Previously, parsing errors were silently ignored, causing users to receive empty results without any indication of data corruption. This change propagates errors with descriptive messages, making it consistent with regular broadcast handling. 

May be valuable because BroadcastReader is used in cheatcodes (vm.getBroadcasts, vm.getDeployments).


DaniPopes:
not backwards compatible

Issue #11112, opened 2025-07-26
Feature Request: Fully Offline Fork Testing via Pre-Loaded Block Header


wishfulcynic:
### Component

Anvil

### Describe the feature you would like

**Problem**

Even with `--fork-chain-id`, Anvil still makes outgoing RPC requests for the fork block header and any uncached storage. That makes it impossible to run fork tests in CI pipelines or air-gapped dev environments, even when you’ve already cached the chain state.

**Proposed feature**

It would be great if there were an `--offline-fork` flag that lets you hand Anvil a pre-downloaded block header (and maybe a chain config) along with the state cache. In that mode Anvil could skip the network entirely and read everything from `~/.foundry/cache/rpc/<chain>/<block>` so people could build the cache on a machine with internet, copy it into their container, and still run their fork tests.

**Use case**

While working in a locked down container, I tried to run a Base fork test using a cached block and `--fork-chain-id`. Anvil still tried to fetch the block header and missing state from https://mainnet.base.org and immediately failed with 403s because the network was blocked. If I could point Anvil at a saved block header and keep it fully offline, those tests would have worked. I imagine others with restricted setups would find this useful too.

Happy to discuss the details and contribute a PR if this sounds interesting.

grandizzy:
@wishfulcynic thank you, this would be a great feature for Anvil suggested by others too. We're tracking this Anvil offline mode with https://github.com/foundry-rs/foundry/issues/9721 so going to mark this as dupe but will use provided details when implementing it.

Issue #9721, opened 2025-01-20
feat(`anvil`): fully offline mode for `anvil` fork


exp0nge:
### Component

Anvil

### Describe the feature you would like

It does not seem like it's currently possible to do a fully offline version of an anvil fork. Even with the following startup commands, it still tries to reach out to the RPC:

```
anvil --fork-url https://sepolia.base.org --optimism --fork-block-number 20702367 --fork-chain-id 84532 --host 0.0.0.0 --load-state ~/state.json --no-mining
```

(I ensured that both state.json and storage.json exist for that fork number).

when attempting to call `eth_chainId`

I get
```
Error: failed to get fork block

Caused by:
   0: error sending request for url (https://sepolia.base.org/)
   1: error sending request for url (https://sepolia.base.org/)
   2: operation timed out
```

(the env does not have access to the internet)

onbjerg:
Please see https://github.com/foundry-rs/foundry/pull/12442#issuecomment-3486588438

TL;DR: Passing a fork URL will always talk to upstream to ask for missing state. If you do not want this, and just want to load state from disk, use `--load-state` and omit `--fork-url`. In fork mode, a lot of the anvil RPC is just a passthrough to the forked node.

exp0nge:
> Please see [#12442 (comment)](https://github.com/foundry-rs/foundry/pull/12442#issuecomment-3486588438)
> 
> TL;DR: Passing a fork URL will always talk to upstream to ask for missing state. If you do not want this, and just want to load state from disk, use `--load-state` and omit `--fork-url`. In fork mode, a lot of the anvil RPC is just a passthrough to the forked node.

```
error: the following required arguments were not provided:
  --fork-url <URL>

Usage: anvil --fork-url <URL> --fork-block-number <BLOCK> --fork-chain-id <CHAIN> --host <IP_ADDR> --load-state <PATH> --no-mining

For more information, try '--help'.
```

It's required?

onbjerg:
I see, that should not be the case. Can we mark it as not required and see if that works instead?

PR #12442, opened 2025-11-04
feat: offline fork

exp0nge:
## Motivation

* https://github.com/foundry-rs/foundry/issues/9721
* https://github.com/foundry-rs/foundry/issues/11112

The current fork functionality calls out to the defined RPC url (internet) when either something is missing from the local state DB or misc calls like (eth_chainId). 

With --load-state, this behavior doesn't change. As outlined above in the two issues, the goal of this PR is to entirely eliminate all remote RPC calls when the fork is running in `--offline` mode: if there is state cached: use that, if not, return default/0.

This allows the Anvil fork to become portable in environments where internet egress is not possible or unwanted due to security/reproducibility concerns (e.g. namely CI systems).

## Solution

* Use a new offline fork DB
* Redirect all calls to local state or default if offline mode

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


onbjerg (changes requested):
Please provide a description and a motivation for this change. Going off the original issue, you are passing a fork URL, which Anvil correctly tries to talk to. If you want to be fully offline, you should be able to just pass the chain ID, and use `--load-state`.

exp0nge:
> Please provide a description and a motivation for this change. Going off the original issue, you are passing a fork URL, which Anvil correctly tries to talk to. If you want to be fully offline, you should be able to just pass the chain ID, and use `--load-state`.

Hey, updated the description. I actually did start with that flag (re: my original issue), but the fork will always call out to the RPC for missing state.

onbjerg:
Yes, this is expected, if you pass a fork URL it will try to reach out to get missing state and things like chain ID. It should work if you remove `--fork-url` and just use `--load-state`. That is, to use it in offline mode, use state from disk (`--load-state`) and don't pass a fork URL. I'm closing this PR and the issue given this is working as intended, unless there is a bug with `--load-state`.

exp0nge:
> Yes, this is expected, if you pass a fork URL it will try to reach out to get missing state and things like chain ID. It should work if you remove `--fork-url` and just use `--load-state`. That is, to use it in offline mode, use state from disk (`--load-state`) and don't pass a fork URL. I'm closing this PR and the issue given this is working as intended, unless there is a bug with `--load-state`.

OK, the other parts to this was missing some specific support for like debug_ rpc endpoints. Should I PR that separately?

onbjerg:
Sorry, I lost track of this, yes, that'd be great:)

exp0nge:
Decomposed into separate PRs:

- #14391 — `debug_traceBlockByHash` + `debug_traceBlockByNumber` RPC endpoints
- #14392 — `--fund-accounts` CLI flag

emmajam:

PR #7864, opened 2024-05-05
chore(deps): bump watchexec to 4

DaniPopes:
Supersedes https://github.com/foundry-rs/foundry/pull/6446

~~I don't know how to handle the update args case for `forge test`, otherwise it probably works (untested)~~

Mostly copied from watchexec's own CLI, simplified because less options

PR #12443, opened 2025-11-04
chore(forge): clean up hardcoded booleans in watch::end_of_process

zeroprooff:
Found two variables in `end_of_process` that were hardcoded since the watchexec v4 migration (#7864):

```rust
let timings = true;  // always true
let quiet = false;   // always false
```

They look like they were meant to be configurable at some point but never got implemented. Simplified the logic since the conditions are always the same anyway.


DaniPopes:
doesnt matter

PR #12445, opened 2025-11-04
fix: deduplicate compiler version normalization

prestoalvarez:
### Changes:
  - Add version_without_build and normalize_compiler_version_str in `crates/verify/src/utils.rs`.
  - Use shared helpers in `utils.rs` and `etherscan/mod.rs` instead of inline parsing.

Removes duplicated logic (split by '+' and trim leading 'v'), reducing maintenance risk and keeping behavior

### DIFF
--- a/crates/verify/src/utils.rs
+++ b/crates/verify/src/utils.rs
@@ -59,16 +59,10 @@ pub struct JsonResult {
     pub message: Option<String>,
 }
 
-/// Returns the version string without build metadata part after '+'.
-/// Returns `None` if the version string is empty or contains only '+'.
-/// Example: "0.8.26+commit.abcdef" -> Some("0.8.26")
 pub(crate) fn version_without_build(version: &str) -> Option<&str> {
     version.split('+').next().filter(|s| !s.is_empty())
 }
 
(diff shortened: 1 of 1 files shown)


crates/verify/src/utils.rs
```diff
@@ -59,6 +59,14 @@ pub struct JsonResult {
     pub message: Option<String>,
 }
 
+pub fn version_without_build(version: &str) -> String {
+    version.split('+').next().unwrap_or("").to_string()
+}
+
+pub fn normalize_compiler_version_str(version: &str) -> String {
+    version_without_build(version).trim_start_matches('v').to_string()
+}
```
onbjerg: these are only used in this crate, so please do not mark them pub.

please also add a small doc comment explaining what they do so you don't need to look up the fn definition

crates/verify/src/utils.rs
```diff
@@ -59,6 +59,14 @@ pub struct JsonResult {
     pub message: Option<String>,
 }
 
+pub fn version_without_build(version: &str) -> String {
+    version.split('+').next().unwrap_or("").to_string()
```
onbjerg: this seems like it should just be an `Option<_>` since all we do is check if it's empty and then provide a better default.

PR #12461, opened 2025-11-05
chore: bump revm 31

mattsse:
this still needs some work

with recent revm change we can no longer access &mut Account and need workarounds for this

ideally helpers on `EvmInternals`

todo

figure out which ones and upstream to alloy-evm and make nicer

### DIFF
--- a/crates/anvil/src/eth/error.rs
+++ b/crates/anvil/src/eth/error.rs
@@ -165,6 +165,7 @@ where
                 OpTransactionError::HaltedDepositPostRegolith => {
                     Self::DepositTransactionUnsupported
                 }
+                OpTransactionError::MissingEnvelopedTx => Self::InvalidTransaction(err.into()),
             },
             EVMError::Header(err) => match err {
                 InvalidHeader::ExcessBlobGasNotSet => Self::ExcessBlobGasNotSet,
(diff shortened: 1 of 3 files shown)


crates/evm/networks/src/celo/transfer.rs
```diff
@@ -62,44 +62,39 @@ pub fn celo_transfer_precompile(input: PrecompileInput<'_>) -> PrecompileResult
...
-        to_account.data.info.balance += value;
+    // Check for overflow in to account
+    if to_account.data.info.balance.checked_add(value).is_none() {
+        return Err(PrecompileError::Other("Balance overflow in to account".into()));
     }
 
+    // Transfer the value between accounts
+    internals
+        .transfer(from_address, to_address, value)
+        .map_err(|e| PrecompileError::Other(format!("Failed to perform transfer: {e:?}")))?;
```
mattsse, line 96: checks out

mattsse (commented):
these lgtm now

zerosnacks (approved):
lgtm

crates/evm/networks/src/celo/transfer.rs
mablr: The inner journal impl makes the same balance checks as above but the errors are not bubbled up, this may be a nice follow-up.
https://github.com/bluealloy/revm/blob/a1fdb9d9e98f9dd14b7577edbad49c139ab53b16/crates/context/src/journal/inner.rs#L315-L320

PR #12468, opened 2025-11-05
Remove redundant testFail iteration in ContractRunner

DotFunGuy:
Collect testFail* functions once in run_tests instead of cloning the iterator, eliminating duplicate filtering work.
Preserve behaviour while reducing unnecessary allocations and iterator clones.


DaniPopes:
wrong

PR #12470, opened 2025-11-05
fix: validate conflicting --broadcast and --resume flags in script command

hawkadrian:
Add validation to prevent using --broadcast and --resume flags together in forge script command. These flags are mutually exclusive: --resume continues a previous broadcast attempt while --broadcast starts a new one.

Also fixes context selection to prioritize --resume over --broadcast, matching the execution logic where resume takes precedence.


DaniPopes:
brking chnge

PR #12479, opened 2025-11-06
chore(foundryup): update master refs to HEAD

DaniPopes:
Allows changing the default branch in the future


mattsse (approved):

zerosnacks:
Note, if we want existing users to be prompted to update we need to update

https://github.com/foundry-rs/foundry/blob/ebcd59b575e1ff8a1fecda3f1411afa3dd0e57fb/foundryup/foundryup#L4-L6

PR #12487, opened 2025-11-07
fix: remove code duplication in script-sequence

kilavvy:
Eliminates duplicate string formatting in `get_paths` and removes a redundant comment in `sig_to_file_name`.


0xrusowsky (approved):
@DaniPopes batch candidate

PR #12490, opened 2025-11-07
fix(anvil): set evm block number to input value

### LINKED ISSUE #12417: bug(anvil): Setting `--number` changes the block number returned by the RPC method `eth_blockNumber` but not the one returned by the `NUMBER` opcode
### Component

Anvil

- [x] Foundry
- [x] Foundryup

1.4.3-stable

1.3.0

anvil --number

### Operating System

macOS (Apple Silicon)

### Describe the bug

Setting `--number` changes the block number returned by the RPC method `eth_blockNumber` but not the one returned by the `NUMBER` opcode.

## How to reproduce

Run anvil with a block `--number`

```
$ anvil --number 4242
```

Query the block number using `eth_blockNumber`

```
$ cast bn
4242
```

Call a contract that accesses `block.number`

In this example, I call a contract that does
```
[00] NUMBER
[01] PUSH0
[02] MSTORE
[03] PUSH1 0x20
[05] PUSH0
[06] RETURN
```

so whatever the selector, it returns whatever it got from the opcode `NUMBER`
```
$ cast call --override-code 0x0000000000000000000000000000000000000b0b:435f5260205ff3 0x0000000000000000000000000000000000000b0b 'num() returns(uint256)'
0
```

Syzygy106:
## Motivation

Fixes #12417

When starting Anvil with the `--number` flag in non-fork mode, there was a desynchronization between the RPC layer and the EVM execution layer:

- **RPC layer** (`eth_blockNumber`): correctly returned the configured genesis block number (e.g., 4242)
- **EVM execution layer** (`NUMBER` opcode): returned 0 until the first block was mined

This caused contracts that read `block.number` to receive incorrect values immediately after Anvil startup, breaking workflows that depend on starting at a specific block number (e.g., testing upgrade scenarios, simulating mainnet state).

**Reproduction**:
```bash
# Terminal 1
anvil --number 4242

# Terminal 2
# Returns 4242 (ok)
cast bn  
# Returns 0 (should be 4242)
cast call --override-code 0x0000000000000000000000000000000000000b0b:435f5260205ff3 \
  0x0000000000000000000000000000000000000b0b 'num() returns(uint256)' 

```

## Solution

Synchronize `env.block_env.number` with `genesis.number` during `Backend::with_genesis()` for non-fork initialization. This ensures both the RPC layer (`blockchain.best_number`) and the EVM execution environment (`env.block_env.number`) start with the same block number.

**Key changes**:
1. **`crates/anvil/src/eth/backend/mem/mod.rs`**: Add synchronization during backend initialization in non-fork mode, before the `Backend` struct is constructed. This follows the same pattern used in `reset_to_in_mem()` and `revert_state_snapshot()`
2. **`crates/anvil/tests/it/anvil.rs`**: Add integration test that validates both RPC and EVM execution layers return consistent block numbers

**Why this approach**: Fork mode already handles synchronization in `setup_fork_db_config()` when fetching remote block data. This fix applies equivalent synchronization for non-fork initialization, guarded by `if fork.read().is_none()` to ensure fork mode behavior remains unchanged.

**Test validation**:
```bash
cargo test -p anvil --test it -- test_number_opcode_reflects_genesis_block_number
```

The test deploys bytecode containing the `NUMBER` opcode and verifies it returns the correct value (accounting for pending block = genesis + 1).

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


Syzygy106:
@mattsse , I feel this bug has evolutionary roots. 

Anvil originally started blocks at 0 for local testing --> fork mode --> '--number' flag.

Also that issue could be hard to detect because after mining the first block, synchronization happens automatically in `do_mine_block()` as I see.

crates/anvil/tests/it/anvil.rs
```diff
@@ -139,6 +139,30 @@ async fn test_can_use_default_genesis_block_number() {
...
+    let provider = handle.http_provider();
+
+    // RPC layer should return configured genesis number
+    let bn = provider.get_block_number().await.unwrap();
+    assert_eq!(bn, genesis_number);
+
+    // Deploy bytecode that returns block.number
+    // 0x43 (NUMBER) 0x5f (PUSH0) 0x52 (MSTORE) 0x60 0x20 (PUSH1 0x20) 0x5f (PUSH0) 0xf3 (RETURN)
+    let target = Address::random();
+    api.anvil_set_code(target, bytes!("435f5260205ff3")).await.unwrap();
```
mattsse, line 157: cool, 
I will def steal this in the future

PR #12494, opened 2025-11-07
Remove redundant cloning in script simulation setup

Doryu001:
Take ownership of `execution_result.transactions` instead of cloning the queue stop cloning RPC strings when returning from `build_runners`


DaniPopes:
no

PR #12495, opened 2025-11-07
Remove Unused yansi::Paint Import in watch Command

soniseth0:
drop the unused yansi::Paint import from foundry/crates/forge/src/cmd/watch.rs, keeping only Color eliminates the unused_imports lint triggered by the workspace configuration (verification) attempted cargo check -p forge, but cargo is unavailable on the host; no further checks run


mattsse (commented):
we need this

PR #12501, opened 2025-11-08
Remove redundant boolean comparison in invariant test fixture

Doryu001:
replace require(hello.world() == true, "false world"); with require(hello.world(), "false world"); in the CLI invariant target to remove dead code while preserving behaviour keeps the test string clearer and aligns with Solidity style


0xrusowsky:
that's irrelevant

PR #12520, opened 2025-11-10
fix: enable different gas calculation for monad testnet

QEDK:
## Motivation
Monad OPCODEs are priced differently hence we should use `eth_estimateGas` RPC method instead of the local Anvil node

## Solution
Add the testnet as a different gas calculation chain

## PR Checklist
- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


grandizzy (approved):
thanks!

stephancill:
Is it possible to use this with anvil? @grandizzy

PR #12530, opened 2025-11-11
fix: correct comment for Git::has_submodules and remove_contract

sashass1315:
The remove_contract doc suggested mutation and returning runtime bytecode, but the function only searched and returned ABI + creation bytecode + ArtifactId.
The Git::has_submodules docstring incorrectly stated that has_submodules() returns true when there are no submodules. The implementation returns true when git submodule status produces at least one line, i.e., when submodules exist.

### DIFF
--- a/crates/forge/src/cmd/create.rs
+++ b/crates/forge/src/cmd/create.rs
@@ -14,7 +14,7 @@ use eyre::{Context, Result};
 use forge_verify::{RetryArgs, VerifierArgs, VerifyArgs};
 use foundry_cli::{
     opts::{BuildOpts, EthereumOpts, EtherscanOpts, TransactionOpts},
-    utils::{self, LoadConfig, read_constructor_args_file, find_contract_artifacts},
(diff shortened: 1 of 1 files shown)


crates/cli/src/utils/cmd.rs
```diff
@@ -22,8 +22,8 @@ use std::{
 };
 use yansi::Paint;
 
-/// Given a `Project`'s output, removes the matching ABI, Bytecode and
-/// Runtime Bytecode of the given contract.
+/// Given a `Project`'s output, finds the contract by path and name and returns its
+/// ABI, creation bytecode, and `ArtifactId`.
 #[track_caller]
 pub fn remove_contract(
```
onbjerg: please also rename the function.

sashass1315: > please also rename the function.

Did it ser, thank you for your review

sashass1315:
> code needs to be formatted with rustfmt nightly otherwise lgtm

did it, sorry for inconvenience

PR #12533, opened 2025-11-11
chore(anvil): use `Block` and `Header` from Alloy

### LINKED ISSUE #12531: replace anvil block type with alloy's
### Component

Anvil

we have https://github.com/foundry-rs/foundry/blob/eb76d29cd61e3696440b43461bff713d89706385/crates/anvil/core/src/eth/block.rs#L19-L19

which is basically just alloy's block type with extra steps

I'm 90% sure we can phase this out together with 

https://github.com/foundry-rs/foundry/blob/eb76d29cd61e3696440b43461bff713d89706385/crates/anvil/core/src/eth/block.rs#L69-L69

## TODO
* remove partialheader entirely (can just use header and change the inits e.g. state root)
* replace block type

### Additional context

_No response_

mablr:
## Motivation

Close #12531

## Solution

- Use Alloy structs:
  - Replace `PartialHeader` w/ Alloy's `Header`
  - Replace `Block` w/ Alloy's `Block` and `BlockBlody`
- Declared convenience `Block` type for `alloy_consensus::Block<Transaction>`
- Replace previous `Block` constructor w/ `create_block` free function helper 
- Adjusted all uses of `Block` w/ new struct. 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/eth/backend/mem/storage.rs
+++ b/crates/anvil/src/eth/backend/mem/storage.rs
@@ -280,27 +280,18 @@ impl BlockchainStorage {
 
         // create a dummy genesis block
         let header = Header {
-            parent_hash: Default::default(),
-            ommers_hash: Default::default(),
-            beneficiary: env.evm_env.block_env.beneficiary,
-            state_root: Default::default(),
-            transactions_root: Default::default(),
-            receipts_root: Default::default(),
-            logs_bloom: Default::default(),
-            difficulty: env.evm_env.block_env.difficulty,
-            number: genesis_number,
-            gas_limit: env.evm_env.block_env.gas_limit,
-            gas_used: 0,
             timestamp,
-            extra_data: Default::default(),
-            mix_hash: Default::default(),
-            nonce: Default::default(),
             base_fee_per_gas: base_fee,
+            gas_limit: env.evm_env.block_env.gas_limit,
+            beneficiary: env.evm_env.block_env.beneficiary,
+            difficulty: env.evm_env.block_env.difficulty,
             blob_gas_used: env.evm_env.block_env.blob_excess_gas_and_price.as_ref().map(|_| 0),
             excess_blob_gas: env.evm_env.block_env.blob_excess_gas(),
+            number: genesis_number,
             parent_beacon_block_root: is_cancun.then_some(Default::default()),
             withdrawals_root: is_shanghai.then_some(EMPTY_WITHDRAWALS),
             requests_hash: is_prague.then_some(EMPTY_REQUESTS_HASH),
+            ..Default::default()
         };
         let block = create_block(header, Vec::<MaybeImpersonatedTransaction>::new());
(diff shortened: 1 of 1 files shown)


crates/anvil/core/src/eth/block.rs
```diff
@@ -1,11 +1,14 @@
...
-use alloy_rlp::{RlpDecodable, RlpEncodable};
+use alloy_consensus::{
+    BlockBody, EMPTY_OMMER_ROOT_HASH, Header, proofs::calculate_transaction_root,
+};
 
 // Type alias to optionally support impersonated transactions
 type Transaction = crate::eth::transaction::MaybeImpersonatedTransaction;
 
+/// Type alias for Ethereum Block with Anvil's transaction type
+pub type Block = alloy_consensus::Block<Transaction>;
```
mattsse, line 10: makes sense

crates/anvil/core/src/eth/block.rs
```diff
@@ -14,110 +17,28 @@ pub struct BlockInfo {
...
-            parent_beacon_block_root: value.parent_beacon_block_root,
-            requests_hash: value.requests_hash,
-            withdrawals_root: value.withdrawals_root,
-        }
-    }
+/// Helper function to create a new block with Header and Anvil transactions
+///
+/// Note: if the `impersonate-tx` feature is enabled this will also accept
+/// `MaybeImpersonatedTransaction`.
+pub fn create_block<T>(mut header: Header, transactions: impl IntoIterator<Item = T>) -> Block
```
mattsse, line 24: this is appropriate

crates/anvil/src/eth/backend/mem/storage.rs
```diff
@@ -279,21 +279,30 @@ impl BlockchainStorage {
...
+            mix_hash: Default::default(),
+            nonce: Default::default(),
+            base_fee_per_gas: base_fee,
             blob_gas_used: env.evm_env.block_env.blob_excess_gas_and_price.as_ref().map(|_| 0),
             excess_blob_gas: env.evm_env.block_env.blob_excess_gas(),
-            number: genesis_number,
             parent_beacon_block_root: is_cancun.then_some(Default::default()),
             withdrawals_root: is_shanghai.then_some(EMPTY_WITHDRAWALS),
             requests_hash: is_prague.then_some(EMPTY_REQUESTS_HASH),
-            ..Default::default()
```
mattsse, line 294: can we do this here instead of 

`receipts_root: Default::default()`

because doing ..Default::default() will have the nice side-effect that this will use the default impl of header

mablr: Okay, so I remove all fields with `Default::default()` and replace them with `..Default::default()` at the end.

mablr: yes this is waay cleaner 😅

PR #12537, opened 2025-11-12
Polkadot/Kusama/PolkadotTestnet for RPC gas estimation

albertov19:
## Motivation

Polkadot is rolling out a REVM backend, but it requires custom gas mapping so gas estimation needs to be done via RPC.

## Solution

Adds Polkadot as a named chain that needs gas estimation via RPC call. Goes with https://github.com/alloy-rs/chains/pull/222 & https://github.com/alloy-rs/chains/pull/249

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/cli/src/utils/cmd.rs
```diff
@@ -169,12 +169,13 @@ pub fn has_different_gas_calc(chain_id: u64) -> bool {
...
                     | NamedChain::Mantle
                     | NamedChain::MantleSepolia
+                    | NamedChain::Metis
                     | NamedChain::MonadTestnet
                     | NamedChain::Moonbase
                     | NamedChain::Moonbeam
                     | NamedChain::MoonbeamDev
                     | NamedChain::Moonriver
-                    | NamedChain::Metis
+                    | NamedChain::PolkadotTestnet
```
mattsse, line 135: this is released now, so you can bump in cargo.toml

.gitignore
```diff
@@ -24,3 +24,7 @@ _
 .vite
 .wrangler
 *.zip
+
+
+.github/instructions/kluster-code-verify.instructions.md
+.kluster
```
mattsse: unsure what this is, but maybe doesnt matter

albertov19: Ah, that is an MCP server that runs to verify any AI-generated code (which was not the case here). I've removed it

crates/cli/src/utils/cmd.rs
albertov19: The toml has `0.2` (no patch), which should include `0.2.18` (tested it locally). I bumped the cargo.lock file tho

grandizzy:
lgtm, @albertov19 OK to move in ready for review / merge? thanks!

albertov19:
Thanks for the quick turnaround - the REVM backend should be ready next week. I'll mark it ready for review and ping you all by then. You were too quick 😅

albertov19:
@mattsse @grandizzy  this should be good to go. Thanks!

mattsse:
@albertov19 I just published a new chains patch, can you bump this here?

albertov19:
I think test is failing because the `alloy-rs/chains` package was published 3 min ago -> https://crates.io/crates/alloy-chains/versions

The toml has 0.2 (no patch), which should include latest

mattsse (approved):

albertov19:
Hey @mattsse anything else that we need to do here? Thanks in advance 🙏

PR #12538, opened 2025-11-12
chore: bump to revm 33

zerosnacks:
## Solution

Bumps Revm to 33, currently requires patch to address bug in latest release: https://github.com/bluealloy/revm/pull/3164

Some breaking changes that now force passing in the `db`

For now I've set `was_precompile_called` to `true` for any CallOutcome that is not a revert but I am not sure if this is correct.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/forge/tests/cli/test_optimizer.rs
+++ b/crates/forge/tests/cli/test_optimizer.rs
@@ -568,7 +568,7 @@ contract Counter {
     // compiled and both tests fail.
     cmd.with_no_redact().assert_failure().stdout_eq(str![[r#"
 ...
-Compiling 2 files with [..]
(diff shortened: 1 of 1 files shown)


crates/forge/tests/cli/test_optimizer.rs
```diff
@@ -568,7 +568,7 @@ contract Counter {
     // compiled and both tests fail.
     cmd.with_no_redact().assert_failure().stdout_eq(str![[r#"
 ...
-Compiling 2 files with [..]
+Compiling 3 files with [..]
```
grandizzy, line 571: that's fine from compilers bump

PR #12541, opened 2025-11-12
fix: return empty array encoding for parse_array

Bilogweb3:
When parse_array receives no values, it now returns a properly typed DynSolValue::Array instead of a String. While both have identical ABI encoding, this fixes semantic correctness and makes the code match its comment and function name.


0xrusowsky:
can you add a test?

Bilogweb3:
@0xrusowsky empty string and empty array have identical ABI encoding, so tests pass with both old and new code.

The fix is still correct though:
- parse_array should return Array, not String
- Old comment was wrong ("Bytes" → actually returned String)
- Integration tests already cover this (Env.t.sol lines 247-344)

If you still want me to add tests I can do that.

DaniPopes:
doesn't matter, it's the same

Issue #12518, opened 2025-11-10
Add support for eth_sendrawtransactionSync in cast send


mattsse:
### Component

Cast

### Describe the feature you would like

we have

https://github.com/foundry-rs/foundry/blob/a66490ba53c53eb2f0069ff2fe0d6ced8108e418/crates/cast/src/cmd/send.rs#L36-L38

ideally we also have an option to use eth_sendrawtransactionSync instead of polling

which we could invoke here:

https://github.com/foundry-rs/foundry/blob/a66490ba53c53eb2f0069ff2fe0d6ced8108e418/crates/cast/src/cmd/send.rs#L214-L224

https://github.com/foundry-rs/foundry/blob/a66490ba53c53eb2f0069ff2fe0d6ced8108e418/crates/cast/src/lib.rs#L299-L306

@stevencartavia

stevencartavia:
gm

grandizzy:
I think we can close this one now with https://github.com/foundry-rs/foundry/pull/12546 or are there other things to implement? @stevencartavia thank you

PR #12546, opened 2025-11-12
feat: support for eth_sendrawtransactionSync in cast send

stevencartavia:
## Motivation
towards #12518

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cast/src/lib.rs
+++ b/crates/cast/src/lib.rs
@@ -332,8 +332,30 @@ impl<P: Provider<AnyNetwork>> Cast<P> {
 
     /// Sends a transaction and waits for receipt synchronously
     pub async fn send_sync(&self, tx: WithOtherFields<TransactionRequest>) -> Result<String> {
-        let receipt = self.provider.send_transaction_sync(tx).await?;
-        Ok(serde_json::to_string_pretty(&receipt)?)
+        let mut receipt: TransactionReceiptWithRevertReason =
+            self.provider.send_transaction_sync(tx).await?.into();
+
+        // Allow to fail silently
+        let _ = receipt.update_revert_reason(&self.provider).await;
+
+        self.format_receipt(receipt, None)
+    }
+
+    /// Helper method to format transaction receipts consistently
+    fn format_receipt(
+        &self,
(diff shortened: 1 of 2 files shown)


crates/cast/src/cmd/send.rs
```diff
@@ -40,6 +40,10 @@ pub struct SendTxArgs {
     #[arg(id = "async", long = "async", alias = "cast-async", env = "CAST_ASYNC")]
     cast_async: bool,
 
+    /// Wait for transaction receipt synchronously instead of polling.
```
mattsse, line 43: we should mention here that this uses `eth_sendTransactionSync` and might not be supported by all clients

crates/cast/src/cmd/send.rs
```diff
@@ -230,19 +236,27 @@ async fn cast_send<P: Provider<AnyNetwork>>(
...
 
-    if cast_async {
-        sh_println!("{tx_hash:#x}")?;
-    } else {
-        let receipt =
-            cast.receipt(format!("{tx_hash:#x}"), None, confs, Some(timeout), false).await?;
+    if sync {
+        // Send transaction and wait for receipt synchronously
+        let receipt = cast.send_sync(tx).await?;
         sh_println!("{receipt}")?;
```
mattsse, line 248: ah this isn't actually equivalent because this now prints as json whereas

https://github.com/foundry-rs/foundry/blob/975a456ef42ab506b8d343df5541a248798c5a27/crates/cast/src/lib.rs#L840-L848

does not

what we can do is, refactor this so that this becomes reusable

https://github.com/foundry-rs/foundry/blob/975a456ef42ab506b8d343df5541a248798c5a27/crates/cast/src/lib.rs#L864-L874


we also want a new simple test case like this

https://github.com/foundry-rs/foundry/blob/975a456ef42ab506b8d343df5541a248798c5a27/crates/cast/tests/cli/main.rs#L2474-L2498

mattsse (changes requested):

mattsse (approved):
nice

PR #12551, opened 2025-11-13
refactor: deduplicate setUp/afterInvariant validation in ContractRunner

letmehateu:
## Motivation

The validation logic for setUp and afterInvariant functions in ContractRunner::run_tests() was duplicated (~50 lines). This made the code harder to maintain and would require updating multiple places when adding new special functions.

## Solution

Extracted a reusable validate_special_function() helper method that handles validation for any special test function. This eliminates code duplication and makes it easy to add validation for future special functions.
The validation behavior remains functionally identical to the previous implementation.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/forge/src/runner.rs
+++ b/crates/forge/src/runner.rs
@@ -282,60 +282,76 @@ impl<'a> ContractRunner<'a> {
         FuzzFixtures::new(fixtures)
     }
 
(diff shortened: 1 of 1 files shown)


DaniPopes:
this is worse

PR #11178, opened 2025-08-01
feat(anvil): bypass EIP-7702 authorization signatures for testing

shiyasmohd:
Closes #11170

PR #12553, opened 2025-11-13
feat(anvil): handle signature impersonation for EIP-7702 authorization list

### LINKED ISSUE #11170: Allow passing signature check on authorization list
### Component

Anvil

Anvil currently supports impersonating addresses in order to bypass signing transactions by the sender

in a recent discussion, there is a suggestion to allow mocking the precompiled ecRecover in order to bypass signature checking of contracts to be able to simulate transactions without really signing any messages

We should also have a way to bypass signature checking for authorization list checks in set code transactions to support simulation of such transactions without actually signing anything

### Additional context

_No response_

mablr:
## Motivation

Follow-up for #11178 to close #11170

## Solution

- Override 7702 txs' invalid recovered authorizations with signature overrides from cheat manager
- added an integration test w/ sig impersonation

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/eth/backend/executor.rs
+++ b/crates/anvil/src/eth/backend/executor.rs
@@ -281,7 +281,9 @@ impl<DB: Db + ?Sized, V: TransactionValidator> TransactionExecutor<'_, DB, V> {
         let mut tx_env: OpTransaction<TxEnv> =
             FromRecoveredTx::from_recovered_tx(&tx.transaction.transaction, *tx.sender());
 
-        if let TypedTransaction::EIP7702(tx_7702) = &tx.transaction.transaction {
+        if let TypedTransaction::EIP7702(tx_7702) = &tx.transaction.transaction
+            && self.cheats.has_recover_overrides()
+        {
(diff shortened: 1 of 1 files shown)


crates/anvil/src/eth/backend/executor.rs
```diff
@@ -276,6 +281,33 @@ impl<DB: Db + ?Sized, V: TransactionValidator> TransactionExecutor<'_, DB, V> {
...
+                .tx()
+                .authorization_list
+                .iter()
+                .zip(tx_env.base.authorization_list)
+                .map(|(signed_auth, either_auth)| {
+                    either_auth.right_and_then(|recovered_auth| {
+                        if recovered_auth.authority().is_none()
+                            && let Ok(signature) = signed_auth.signature()
+                            && let Some(override_addr) =
+                                self.cheats.get_recover_override(&signature.as_bytes().into())
```
mattsse, line 296: I see, this makes sense

PR #12561, opened 2025-11-14
fix: use AssumeReject instead of TooManyRejects for single vm.assume reject

DeVikingMark:
Fix incorrect error type when vm.assume rejects a single input. 

Previously, the code returned TooManyRejects immediately, which prevented proper reject counting and limit checking. 

Now it correctly returns AssumeReject for a single reject, allowing the existing limit check logic to work as intended.


grandizzy (approved):
thank you, makes sense!

PR #12562, opened 2025-11-14
docs: add missing documentation for ScriptSequence methods

CreeptoGengar:
Adds documentation comments for add_receipt, add_pending, remove_pending, and fill_sensitive methods in ScriptSequence. These methods are part of the public API and are used by other modules, so they should be documented for better code maintainability.


mattsse (commented):
this isnt needed

PR #12563, opened 2025-11-14
fix(verify): mismatch between strip_prefix and strip_suffix

Bashmunta:
Replace strip_prefix("/api") with strip_suffix("/api") in EtherscanVerificationProvider::client() so that when a custom verifier API endpoint like https://host/api is provided, the browser base URL passed to with_url() becomes https://host/. This ensures address_url() and related links are constructed correctly for Blockscout and other non-Etherscan providers


grandizzy (approved):
makes sense

PR #12569, opened 2025-11-14
drop redundant clone when building calldata

Doryu001:
replace the self.input.clone() call with a borrowed slice to avoid re‑allocating calldata bytes keeps behaviour identical while saving one allocation for every REPL run


grandizzy:
the way it is should be OK

PR #12573, opened 2025-11-14
chore(common): remove unused Typed2718 import from constants.rs

soniseth0:
Removes unused `Typed2718` import from `foundry-common/src/constants.rs` to clean up dead code and eliminate compiler warnings.


mattsse (commented):
we need this

PR #11934, opened 2025-10-02
feat(anvil): wire eth_protocolVersion, eth_hashrate, eth_coinbase RPCs

sashass1315:
## Motivation

The codebase had three RPC handler methods (protocol_version, hashrate, author) that were implemented in EthApi but were effectively dead code because they were not exposed through the JSON-RPC interface. These methods existed with proper documentation indicating they were handlers for standard Ethereum RPC calls (eth_protocolVersion, eth_hashrate, eth_coinbase), but they were never wired up in the EthRequest enum or the execute() dispatcher. This meant clients expecting these standard RPC endpoints would receive "method not found" errors, reducing Anvil's compatibility with Ethereum tooling and clients.

## Solution

1. Adding missing EthRequest variants in crates/anvil/core/src/eth/mod.rs:
- EthProtocolVersion(()) for eth_protocolVersion
- EthHashrate(()) for eth_hashrate
- EthCoinbase(()) for eth_coinbase

2. Wiring the handlers in EthApi::execute() in crates/anvil/src/eth/api.rs:
- EthProtocolVersion(_) => self.protocol_version().to_rpc_result()
- EthHashrate(_) => self.hashrate().to_rpc_result()
- EthCoinbase(_) => self.author().to_rpc_result()

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #12574, opened 2025-11-15
chore: add missing tests for eth_protocolVersion, eth_hashrate, eth_c…

Sharilleed223:
Adds the missing tests for PR #11934.


mattsse (commented):
these endpoints are basically unused and can be ignored

PR #12581, opened 2025-11-16
feat(lint): refactor LinterConfig and add 3 new linting rules

milosdjurica:
- Added a new linting rule that checks for multiple contract-like items (contracts, abstract contracts, interfaces and libraries) per `.sol` file . 
- Refactored `LinterConfig`
- Added 2 new linting rules that checks for interface names and interface file names

## Motivation

Enforce best practice.
## Solution

I have some doubts regarding this, would like to hear your opinion.
1. Should the rule skip interfaces and libraries and count only contracts?
2. Should the rule show linting note for every contract-like item in the file, or only one time per file is enough?

My solution count interfaces and libraries too, and it shows linting note only once per file. Let me know if you want something to be changed :) 

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

Keep up the good work guys, cheers!


0xrusowsky:
@milosdjurica first of all thanks for the PR and the initiative, we love new contributions to forge's linter!

i think the idea overall makes sense, however imo we can improve it.

1. i'd like each contract to be flagged, not just the 2nd one
2. imo it would make sense to add a custom config for this lint, so that users can decide if they want to allow interfaces and/or libs in the same file. To do so, you'd want to modify the [LinterConfig](https://github.com/foundry-rs/foundry/blob/master/crates/config/src/lint.rs#L12) to have a dedicated config for this lint. If we do that, i think we then want to create a new intermediate struct that group all lint configs (i.e. the pre-existing `mixed_case_exceptions` + your newly introduced config)

lmk if this seems reasonable and if u have any doubts regarding impl

milosdjurica:
> @milosdjurica first of all thanks for the PR and the initiative, we love new contributions to forge's linter!
> 
> i think the idea overall makes sense, however imo we can improve it.
> 
> 1. i'd like each contract to be flagged, not just the 2nd one
> 2. imo it would make sense to add a custom config for this lint, so that users can decide if they want to allow interfaces and/or libs in the same file. To do so, you'd want to modify the [LinterConfig](https://github.com/foundry-rs/foundry/blob/master/crates/config/src/lint.rs#L12) to have a dedicated config for this lint. If we do that, i think we then want to create a new intermediate struct that group all lint configs (i.e. the pre-existing `mixed_case_exceptions` + your newly introduced config)
> 
> lmk if this seems reasonable and if u have any doubts regarding impl

Sounds great, I like it! Will look to implement it ASAP :) 
Thanks for the response!

0xrusowsky:
@milosdjurica lmk when the PR is ready for re-review!

milosdjurica:
Hi @0xrusowsky , it should be good now 👍🏻

0xrusowsky:
great, tysm! will review it today at some point

crates/lint/src/sol/info/interface_naming.rs
```diff
@@ -0,0 +1,83 @@
...
+            return;
+        }
+
+        // Check if interface name starts with 'I'
+        let name = contract.name.as_str();
+        if !name.starts_with('I') {
+            ctx.emit(&INTERFACE_NAMING, contract.name.span);
+        }
+    }
+}
```
0xrusowsky, line 55: nit, can we simplify with a single let chain? no need for early exit as there is really no branching

```rs
fn check_item_contract(&mut self, ctx: &LintContext, contract: &'ast ast::ItemContract<'ast>) {
    if ctx.is_lint_enabled(INTERFACE_NAMING.id())
        && let ast::ContractKind::Interface = contract.kind
        && !contract.name.as_str().starts_with('I')
    {
        ctx.emit(&INTERFACE_NAMING, contract.name.span);
    }
}
```

crates/lint/src/sol/info/interface_naming.rs
```diff
@@ -0,0 +1,83 @@
...
+                    _ => return, // Mixed file, skip lint
+                }
+            }
+        }
+
+        // Emit if file contains ONLY interfaces. Emit only on the first interface.
+        if let Some(span) = first_interface_span {
+            ctx.emit(&INTERFACE_FILE_NAMING, span);
+        }
+    }
```
0xrusowsky, line 45: nit, we can also simplify as a single let chain here? we should be able to do that if we add a helper to compute the file name with something like:
```rs
if let Some(file_name) = file_name(ctx, unit)
    && !file_name.starts_with('I')
     && unit.items.iter().all(|item| {
        matches!(item.kind, ast::ItemKind::Contract(c) if c.kind == ast::ContractKind::Interface)
    })
    && let Some(ast::ItemKind::Contract(c)) = unit.items.first().map(|item| &item.kind)
{
    ctx.emit(&INTERFACE_FILE_NAMING, c.name.span);
}
```
```rs
fn file_name<'ast>(ctx: &LintContext<'ast, '_>, unit: &'ast ast::SourceUnit<'ast>) -> Option<&str> {
    let first_item_span = unit.items.first()?.span;
    let file = ctx.session().source_map().lookup_source_file(first_item_span.lo());
    file.name.as_real()?.file_name()?.to_str()
}
```

crates/lint/src/sol/info/multi_contract_file.rs
```diff
@@ -0,0 +1,56 @@
...
+
+                        ast::ContractKind::Library => should_lint_libraries,
+                        ast::ContractKind::AbstractContract => should_lint_abstract,
+                        ast::ContractKind::Contract => true, // Regular contracts are always linted
+                    };
+                    should_lint.then_some(c.name.span)
+                }
+                _ => None,
+            })
+            .collect();
```
0xrusowsky: can we move the exception logic to an impl block on `LintSpecificConfig`? this way the code will be much cleaner.

maybe something like:
```rs
impl LintSpecificConfig {
    /// Checks if a given contract kind is included in the list of exceptions
    fn is_exempted(&self, contract_kind: &ast::ContractKind) -> bool {
        let exception_to_check = match contract_kind {
            ast::ContractKind::Interface => ContractException::Interface,
            ast::ContractKind::Library => ContractException::Library,
            ast::ContractKind::AbstractContract => ContractException::AbstractContract,
            // Regular contracts are always linted
            ast::ContractKind::Contract => return false,
        };

        self.multi_contract_file_exceptions.contains(&exception_to_check)
    }
}

0xrusowsky:
@milosdjurica btw, i haven't tested the my code compiles, those are just suggestions, so you may need to tweak them a bit, but u get the gist of the requested simplifications

milosdjurica:
Hi @0xrusowsky, thanks for suggestions - it's much cleaner now. I have applied them with some small changes, lmk if this is okay now

0xrusowsky (dismissed):
great work!

pending another approval just in case.
@grandizzy could u have a look too?

milosdjurica:
Hi guys, @0xrusowsky @grandizzy  - just a quick ping in case you missed the latest changes 👀

grandizzy:
> @grandizzy good to merge?

Yep, let's send it 👍

milosdjurica:
lmk if need anything else

PR #12585, opened 2025-11-17
added MONAD_SYSTEM_ADDRESS

Jayakumar2812:
## Motivation

Monad introduces system transactions executed at zero gas by the protocol. 
These appear in the block as regular transactions but do not follow basefee rules.
Currently, Foundry tools (cast, --quick, etc.) treat these as normal transactions. 
This causes:
	  •	cast to fail with Error #1: transaction validation error: gas price is less than basefee
	  •	--quick to fails when multiple system transactions exist in the same block, leading to nonce-too-high errors

## Solution

Added MONAD_SYSTEM_ADDRESS and updated transaction handling to skip Monad system transactions, similar to how L2 system txs are ignored.
This prevents validation errors caused by 0-gas system transactions in cast.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


0xrusowsky:
@Jayakumar2812 please run rustfmt to fix the CI errors

PR #12587, opened 2025-11-17
feat(invariant): assert all invariants by default (`assert_all`)

### LINKED ISSUE #9727: Add an Option to Continue Fuzzing Run on Assertion Failure
### Component

Forge

**Description:**
Currently, Foundry’s invariant tests stop execution upon encountering the first assertion failure, ending the entire fuzzing campaign.

**Feature Request:**
Add an option to allow the fuzzing campaign to continue running until all configured runs are completed, even if assertion failures are encountered.

**Rationale:**
This feature would enable continuous fuzzing in an "Echidna-style" manner, where failing assertions are collected throughout the run without interrupting the campaign. Certain assertion failures require long-running fuzzing campaigns (e.g., 4+ hours), and stopping on the first failure can hinder the discovery of additional issues.

**Expected Behavior:**
When the `continuous_run` option is enabled, the fuzzer should:

1. Continue running despite encountering assertion failures.

2. Log all failures encountered during the campaign.

3. Upon manual cancellation (e.g., CTRL+C/CMD+C) or completion of the configured runs/timeout, present all call traces for reproducing the assertion failures.

Example Configuration:
```toml
[invariant]
continuous_run = true
```

Example Output:

```plaintext
test/fuzzing/FuzzPositions.sol:FuzzPositions                                     
    ↪ invariant_GLOB_01: [9/4294967295] Runs  
    ❌ Failures: 2
```

### Additional context

_No response_

grandizzy:
closes #9727

Adds `invariant.assert_all` (default `true`, matches Echidna/Medusa). When enabled the campaign keeps running after the first broken invariant and shrinks every broken one in a single run.

What this PR adds:
- Per-invariant rendering: each broken invariant gets a `[FAIL: reason]` line + `[Sequence]` block. The invariant function name is appended to the FAIL line whenever it's needed to disambiguate — i.e. when there are multiple failures, or the single failure isn't the campaign anchor. The anchor's name is omitted in single-failure runs because the trailing `<name>() (runs: …)` summary already identifies it.
- `Suite assert_all: <broken>/<total> invariants broken` roll-up above the per-invariant blocks
- `[i/N] Shrink: <invariant>` counter on the progress bar when more than one invariant needs shrinking
- `N invariant failures persisted to <dir> — rerun to shrink` footer
- On Ctrl+C: every known broken invariant is reported and the un-shrunk sequence is persisted, so re-running targets that invariant shrinks from cache
- Warnings: `assert_all + optimization invariant` (boolean secondaries silently filtered) and `N invariant(s) skipped due to persisted failures` on re-runs
- Set `invariant.assert_all = false` to restore the legacy "stop on first broken invariant" behavior

Internal cleanup behind the new behavior: a single `invariant_fns` list per `InvariantContract` (with an `anchor_idx`), a single `invariant_failures: Vec<InvariantFailure>` on `TestResult`, and direct return of the broken invariant from the assertion path (removing the prior `first_broken_event` re-scan). The campaign **anchor** is the `--mt`-selected invariant (or the only one), shrunk and persisted at the campaign root path; secondaries share the same shape and are shrunk + persisted under their own filenames.

Parameterized `invariant_*` functions (e.g. `invariant_bar(uint256 z)`) are now rejected upstream with a per-function `TestResult::fail` entry, instead of silently leaking into `assert_all` secondaries and breaking campaigns mid-run with confusing decode errors.

@forkforkdog @rappie @GalloDaSballo appreciate some feedback and if that's what you were looking for, thank you
CC @0xalpharush

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes (default config behavior change: campaigns now report all broken invariants by default)


crates/config/src/invariant.rs
```diff
@@ -37,6 +37,8 @@ pub struct InvariantConfig {
     pub timeout: Option<u32>,
     /// Display counterexample as solidity calls.
     pub show_solidity: bool,
+    /// Continue invariant run until all invariants declared in current test suite breaks.
```
DaniPopes:
```suggestion
    /// Continue invariant run until all invariants declared in the test suite break.
```

0xalpharush:
> invariants that already have persisted failure are not picked up / asserted

Not to say this can't be done, but this isn't how Medusa/Echidna do it. The "pro" that comes to mind is you may find a smaller reproducer or a reproducer that reveals the patch was incomplete i.e. only fixed one repro but not the other.

It is possible to do deduplication with the edge map to avoid saving too many if you went this route.

grandizzy:
> > invariants that already have persisted failure are not picked up / asserted
> 
> Not to say this can't be done, but this isn't how Medusa/Echidna do it. The "pro" that comes to mind is you may find a smaller reproducer or a reproducer that reveals the patch was incomplete i.e. only fixed one repro but not the other.
> 
> It is possible to do deduplication with the edge map to avoid saving too many if you went this route.

I see, makes sense, can accommodate such

rappie:
This looks great. Some quick ideas for further improvements:
- Keep running even after all invariants a broken because there might still be unbroken assertions in handlers
- Keep track of broken handlers. Once a handler is broken once, ignore all its assertions BUT keep calling it for building interesting sequences

crates/config/src/invariant.rs
```diff
@@ -37,6 +37,8 @@ pub struct InvariantConfig {
...
+    pub continuous_run: bool,
```
onbjerg: not sure if this term was chosen for a specific reason, but i was thinking this could be `--no-fail-fast` like for cargo?

edit: i see we have `--fail-fast` on `forge test`, i guess we can't really make a `--no-fail-fast` as these would seem linked and the current default is actually `--no-fail-fast` (for the entire suite)

Kushmanmb:
Let's not play micky mouse games. Could man to man ask. That honestly more logical

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -528,7 +526,7 @@ impl<'a> InvariantExecutor<'a> {
                     "timestamp": SystemTime::now()
                         .duration_since(UNIX_EPOCH)?
                         .as_secs(),
-                    "invariant": invariant_contract.invariant_function.name,
+                    "invariant": invariant_contract.invariant_fn.name,
```
0xalpharush: for `continuous_run==true`, we should probably just do `invariantContract`. Really, I think this should be the behavior regardless of config as `continuous_run` should just mean don't stop on failure immediately

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -501,7 +489,9 @@ impl<'a> InvariantExecutor<'a> {
             corpus_manager.process_inputs(&current_run.inputs, current_run.new_coverage);
 
             // Call `afterInvariant` only if it is declared and test didn't fail already.
-            if invariant_contract.call_after_invariant && !invariant_test.has_errors() {
+            if invariant_contract.call_after_invariant
+                && !invariant_test.has_errors(invariant_contract.invariant_fn)
```
0xalpharush: I'm not sure why this doesn't check `invariant_contract.invariant_fns` for `continuous_run`. Is this purposeful

Same for below
```
failures.get_failure(invariant_contract.invariant_fn)
```

crates/evm/evm/src/executors/invariant/result.rs
```diff
@@ -130,37 +138,50 @@ pub(crate) fn assert_invariants<FEN: FoundryEvmNetwork>(
...
         // We only care about invariants which we haven't broken yet.
-        if invariant_failures.error.is_none() {
-            let case_data = FailedInvariantCaseData::new(
-                invariant_contract,
-                invariant_config,
-                targeted_contracts,
-                calldata,
-                call_result,
-                &inner_sequence,
+        if invariant_failures.has_failure(invariant) {
```
0xalpharush, line 158: Given this is mut (`invariant_failures: &mut InvariantFailures,`), couldn't we remove failing invariant_fn and avoid this lookup? Only other reason to keep it around would be to store duplicate failures, but we're not doing that here

grandizzy: yep, good point. I evaluated the change and it's a bigger refactor, OK if will do in a follow-up / cleanup?

0xalpharush: Tbh I think it is fine to just keep calling these, and we should figure out a deduplication strategy. https://github.com/foundry-rs/foundry/pull/14482#discussion_r3154539830

grandizzy: kk, will follow up on #14482 with dedup strategy

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -179,6 +180,23 @@ impl InvariantFailureMetrics {
     }
 }
 
+/// Returns the (name, reason) of the first broken invariant in declaration order. Falls back
+/// to the primary's name with an empty reason if `failures.errors` is somehow empty.
+fn first_broken_event<'a>(
```
0xalpharush: I think it is fine to collect only one and avoid this search

crates/evm/fuzz/src/invariant/mod.rs
```diff
@@ -267,7 +267,9 @@ pub struct InvariantContract<'a> {
     /// Name of the test contract.
     pub name: &'a str,
     /// Invariant function present in the test contract.
-    pub invariant_function: &'a Function,
+    pub primary_invariant_fn: &'a Function,
+    /// All invariant functions present in the test contract and their fail on revert config.
+    pub invariant_fns: Vec<(&'a Function, bool)>,
```
0xalpharush: IMO this could just be one list

crates/forge/src/result.rs
```diff
@@ -421,6 +438,26 @@ pub struct TestResult {
     /// still be successful (i.e self.success == true) when it's expected to fail.
     pub reason: Option<String>,
 
+    /// Additional broken invariants beyond the primary (the one matched by `--mt`, or the only
+    /// one if no filter was applied).
+    ///
+    /// Each entry carries the invariant's name, the failure reason, an optional counterexample,
+    /// and the path where the counterexample has been persisted for shrinking on a subsequent run.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub invariant_secondary_failures: Vec<InvariantSecondaryFailure>,
```
0xalpharush: Again, I don't think we need primary vs secondary to have different data structures

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -695,8 +698,12 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
...
             );
 
-            // Call `afterInvariant` only if it is declared and test didn't fail already.
-            if invariant_contract.call_after_invariant && !invariant_test.has_errors() {
+            // Call `afterInvariant` only if declared and the current run produced no new
+            // failure. Under `assert_all` the campaign keeps running after earlier failures,
+            // but the hook must still execute on subsequent runs.
+            if invariant_contract.call_after_invariant
+                && invariant_test.test_data.failures.errors.len() == failures_before_run
+            {
```
mablr, line 697: Is there a risk of inconsistent `afterInvariant` hook execution after a failure for subsequent runs ?

What if an invariant fails, is the `afterInvariant` hook executed for others ?

crates/evm/fuzz/src/invariant/mod.rs
mablr: +1

crates/forge/src/runner.rs
```diff
@@ -345,7 +345,9 @@ impl<'a, FEN: FoundryEvmNetwork> ContractRunner<'a, FEN> {
 
         // Invariant testing requires tracing to figure out what contracts were created.
         // We also want to disable `debug` for setup since we won't be using those traces.
-        let has_invariants = self.contract.abi.functions().any(|func| func.is_invariant_test());
+        let invariant_fns: Vec<_> =
```
figtracer, line 367: should we validate all collected invariants before building the `InvariantContract`?  

e.g.: `invariant_bar(uint256 z)` can leak as a secondary and this fails unexpectedly

crates/evm/evm/src/executors/invariant/error.rs
```diff
@@ -17,16 +18,41 @@ pub struct InvariantFailures {
     /// The latest revert reason of a run.
     pub revert_reason: Option<String>,
     /// Maps a broken invariant to its specific error.
```
stevencartavia:
```suggestion
    /// Maps each broken invariant (by function name) to its specific error.
```

crates/evm/evm/src/executors/invariant/mod.rs
grandizzy: removed in https://github.com/foundry-rs/foundry/pull/12587/changes/4bb552fd95d504f744e1944eb46d4d362e3164b4

crates/evm/fuzz/src/invariant/mod.rs
grandizzy: changed in https://github.com/foundry-rs/foundry/pull/12587/changes/4bb552fd95d504f744e1944eb46d4d362e3164b4

crates/forge/src/result.rs
grandizzy: unified in https://github.com/foundry-rs/foundry/pull/12587/changes/4bb552fd95d504f744e1944eb46d4d362e3164b4 introduced concept of anchor for the current test performed

crates/forge/src/runner.rs
grandizzy: good catch, addressed in https://github.com/foundry-rs/foundry/pull/12587/changes/0d8cc990864673b71b3e239669d051f1984d6bdf

crates/evm/evm/src/executors/invariant/mod.rs
grandizzy: yep, how this works is afterInvariant is skipped only on the specific run that fails, but will still be asserted for the rest of campaign (further runs)

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -602,40 +588,51 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
...
-                            });
-                            result::RichInvariantResults::new(false, None)
+                            invariant_test.set_error(
+                                anchor,
+                                if assertion_failure {
+                                    InvariantFuzzError::BrokenInvariant(case_data)
+                                } else {
+                                    InvariantFuzzError::Revert(case_data)
+                                },
+                            );
```
mablr, line 634: Only the anchor is marked as broken. Is this fine ?

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -602,40 +588,51 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
...
+                            }
+                            .failed_case(
+                                anchor,
+                                self.config.fail_on_revert,
+                                assertion_failure,
                                 call_result,
                                 &[],
-                            )
-                            .with_assertion_failure(assertion_failure);
+                            );
```
mablr, line 624: When `check_interval > 1`, the call loop skips invariant assertion on non-interval calls. For those skipped calls, this fast path seems to handle reverts and assertion failures independently from `can_continue`, diverging from the full-check path in three compounding ways under `assert_all=true`.

crates/evm/evm/src/executors/invariant/mod.rs
grandizzy: yep, that's fine, failures for the other invariants are shown inline, e.g.
- anchor invariant_a passes, invariant_b and invariant_c fail
```
Ran 1 test for test/Demo.t.sol:Demo
[FAIL: condition b broken] invariant_b
	[Sequence] (original: [..], shrunk: [..])
		sender=0x[..] addr=[src/Counter.sol:Counter]0x… calldata=inc() args=[]
		sender=0x[..] addr=[src/Counter.sol:Counter]0x… calldata=inc2() args=[]

[FAIL: condition c broken] invariant_c
	[Sequence] (original: [..], shrunk: [..])
		sender=0x[..] addr=[src/Counter.sol:Counter]0x… calldata=spike() args=[]

Suite assert_all: 2/3 invariants broken
2 invariant failure(s) persisted to cache/invariant/failures/Demo — rerun to shrink
 invariant_a() (runs: 50, calls: 4000, reverts: 12)
```
- anchor invariant_a fails AND invariant_b fails
```
Ran 1 test for test/Demo.t.sol:Demo
[FAIL: condition a broken] invariant_a
	[Sequence] (original: [..], shrunk: [..])
		sender=0x[..] addr=[src/Counter.sol:Counter]0x… calldata=inc2() args=[]
		sender=0x[..] addr=[src/Counter.sol:Counter]0x… calldata=spike() args=[]

[FAIL: condition b broken] invariant_b
	[Sequence] (original: [..], shrunk: [..])
		sender=0x[..] addr=[src/Counter.sol:Counter]0x… calldata=inc() args=[]

Suite assert_all: 2/3 invariants broken
2 invariant failure(s) persisted to cache/invariant/failures/Demo — rerun to shrink
 invariant_a() (runs: 5, calls: 273, reverts: 8)
```

crates/evm/evm/src/executors/invariant/mod.rs
grandizzy: good point, handler-side assertion failures are deferred to #14482 and should already handle such

mablr: oh okay I see ^^

mablr (commented):
lgtm ! 🥳

mablr (approved):

0xalpharush:
```
Suite assert_all: 2/3 invariants broken
2 invariant failure(s) persisted to cache/invariant/failures/Demo — rerun to shrink
 invariant_a() (runs: 5, calls: 273, reverts: 8)
 ```
I'm still not sure the UX is what I'd expect but we can change this later. I think a lot of it stems from overloading `test` e.g. --mt doesn't really make sense for invariant fuzzing. I would expect to see the contract's name as the suite and each method that we've determined is an assertion test + `invariant_` as "tests". I don't really get why `invariant_a` is the "anchor" and called out here to be rerun when there are multiple failures.

grandizzy:
> ```
> Suite assert_all: 2/3 invariants broken
> 2 invariant failure(s) persisted to cache/invariant/failures/Demo — rerun to shrink
>  invariant_a() (runs: 5, calls: 273, reverts: 8)
> ```
> 
> I'm still not sure the UX is what I'd expect but we can change this later. I think a lot of it stems from overloading `test` e.g. --mt doesn't really make sense for invariant fuzzing. I would expect to see the contract's name as the suite and each method that we've determined is an assertion test + `invariant_` as "tests". I don't really get why `invariant_a` is the "anchor" and called out here to be rerun when there are multiple failures.

Indeed UX is not polished yet but was under the impression we want first to get this one + https://github.com/foundry-rs/foundry/issues/14448 + https://github.com/foundry-rs/foundry/pull/14565 , then get SCFuzz baseline and then iterate on top of it (both fuzzer and UX improvements). Happy to redo this and introduce new UX here, but as long as it doesn't make PR diff huge (already big IMO)

0xalpharush (approved):

figtracer (approved):

forkforkdog:
No way you guys did it!

grandizzy:
> No way you guys did it!

it still needs https://github.com/foundry-rs/foundry/issues/14448 to be fully done, should land soon

PR #12589, opened 2025-11-17
fix(chisel): correct dynamic type detection and .length semantics

VolodymyrBg:
This change implements accurate dynamic type detection in `Type::is_dynamic()` and corrects `.length` handling in 
`Type::map_special()` to align with Solidity semantics. Previously, the dynamicness check treated arrays broadly and left a TODO that allowed fixed arrays of static elements and tuples of static elements to be considered dynamic, which is incorrect per ABI rules; the new logic recurses through tuples and fixed arrays to mark types as dynamic only when appropriate, while retaining dynamic behavior for bytes, string, and dynamic arrays. Additionally, `.length` should not be available on fixed-size bytes `(bytes1..bytes32)` since they are value types and not arrays; the condition for exposing .length was adjusted to include only dynamic types and arrays, excluding fixed bytes. This change is localized: 
`is_dynamic()` is only referenced in the `.length` gate within `map_special()`, and pop remains unchanged via `is_dynamic_array()`. The result is stricter and more correct behavior for member access resolution without altering unrelated flows.

### DIFF
--- a/crates/chisel/src/executor.rs
+++ b/crates/chisel/src/executor.rs
@@ -669,7 +669,7 @@ impl Type {
                     // Array / bytes members
                     let ty = Self::Builtin(ty);
                     match access.as_str() {
-                        "length" if ty.is_dynamic() || ty.is_array() || ty.is_fixed_bytes() => {
+                        "length" if ty.is_dynamic() || ty.is_array() => {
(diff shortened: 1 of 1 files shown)


DaniPopes:
this part is deprecated

PR #12595, opened 2025-11-17
feat(anvil): add `eth_fillTransaction` support

### LINKED ISSUE #12590: Add support for eth_fillTransaction
### Component

Anvil

same as

https://github.com/paradigmxyz/reth/blob/2a16222ea1cd473eaaf15fb2714d4ebd5a19a4b7/crates/rpc/rpc-eth-api/src/helpers/transaction.rs#L449-L456

we can duplicate 

```rust
/// Response type for `eth_fillTransaction` RPC method.
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct FillTransactionResult<T> {
    /// RLP-encoded transaction bytes
    pub raw: Bytes,
    /// Filled transaction object
    pub tx: T,
}
```

in anvil but also upstream the type to foundry as well

### Additional context

assigning @mablr

mablr:
## Motivation

Close #12590 

## Solution

- Introduced `EthFillTransaction` to handle filling transaction fields with defaults.
- Implemented `fill_transaction` method to populate missing fields like nonce, gas limit, and fees.
- Added `FillTransactionResult` struct to encapsulate the response of the new RPC method.
- Updated tests to verify the functionality of the `fill_transaction` method, ensuring it correctly fills in required fields and preserves provided values.

heavily inspired from reth impl s/o @kumaryash90

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -1422,7 +1422,10 @@ impl EthApi {
         }
 
         let typed_tx = self.build_typed_tx_request(request, nonce)?;
-        let tx = self.sign_request(&from, typed_tx)?;
+        let tx = build_typed_transaction(
+            typed_tx,
+            Signature::new(Default::default(), Default::default(), false),
+        )?;
 
         let raw = tx.encoded_2718().to_vec().into();
 


crates/anvil/core/src/eth/transaction/mod.rs
```diff
@@ -1273,6 +1273,18 @@ impl Decodable2718 for TypedReceipt {
...
+///
+/// This type represents a transaction that has been "filled" with default values
+/// for missing fields like nonce, gas limit, and fee parameters.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct FillTransactionResult<T> {
+    /// RLP-encoded transaction bytes
+    pub raw: Bytes,
+    /// Filled transaction request
+    pub tx: T,
+}
```
mattsse, line 1286: I think we can now upstream this to allo-eth-rpc

crates/anvil/src/eth/api.rs
```diff
@@ -1361,6 +1364,71 @@ impl EthApi {
...
+                request.as_mut().set_max_priority_fee_per_gas(tip);
+                tip
+            };
+            if request.as_ref().max_fee_per_gas().is_none() {
+                request.as_mut().set_max_fee_per_gas(self.gas_price() + tip);
+            }
+        }
+
+        let typed_tx = self.build_typed_tx_request(request, nonce)?;
+        let tx = self.sign_request(&from, typed_tx)?;
```
mattsse: this we can't use here, because this expects an actual signer, so this would currently fail for arbitrary senders

what we should do instead here is use the simulate signature to construct `build_typed_transaction` directly

mattsse (changes requested):
nice, smol nit

crates/anvil/src/eth/api.rs
mablr: ok I see

PR #12602, opened 2025-11-18
fix(coverage): use consistent hash method in insert_map

DeVikingMark:
## Description

Fixes inconsistent hash method usage in `LineCoverageCollector::insert_map()`. 

The method was using `hash().unwrap()` while `get_or_insert_map()` uses `get_or_calculate_hash()`, which could lead to mismatched hash values and potential panics.

## Changes

- Replace `hash().unwrap()` with `get_or_calculate_hash()` in `insert_map()` method
- Use `B256::from()` to convert the hash reference to an owned value
- Ensures consistent hash calculation across both methods
- Removes potential panic from `unwrap()` call


grandizzy:
thanks, should be good as it is

PR #12605, opened 2025-11-19
refactor: simplify gas snapshot handling in cheatcodes

Sharilleed223:
Simplify Option unwrapping patterns in gas snapshot handling:
- Replace is_some() + unwrap() with if let Some pattern.
- Remove redundant intermediate variable binding.

### DIFF
--- a/crates/cheatcodes/src/evm.rs
+++ b/crates/cheatcodes/src/evm.rs
@@ -1214,8 +1214,7 @@ fn inner_start_gas_snapshot(
     name: Option<String>,
 ) -> Result {
     // Revert if there is an active gas snapshot as we can only have one active snapshot at a time.
-    if ccx.state.gas_metering.active_gas_snapshot.is_some() {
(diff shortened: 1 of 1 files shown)


Sharilleed223:
Hi @grandizzy, did it, pls check, thanks!

crates/cheatcodes/src/evm.rs
```diff
@@ -1214,8 +1214,7 @@ fn inner_start_gas_snapshot(
     name: Option<String>,
 ) -> Result {
     // Revert if there is an active gas snapshot as we can only have one active snapshot at a time.
-    if ccx.state.gas_metering.active_gas_snapshot.is_some() {
-        let (group, name) = ccx.state.gas_metering.active_gas_snapshot.as_ref().unwrap().clone();
+    if let Some((group, name)) = ccx.state.gas_metering.active_gas_snapshot.clone() {
```
grandizzy: we can avoid clone here
```rust
    if let Some((group, name)) = &ccx.state.gas_metering.active_gas_snapshot {
        bail!("gas snapshot was already started with group: {group} and name: {name}");
    }
```

Issue #12601, opened 2025-11-18
feat(`forge script`): add `--interactive` alias for `--interactives 1`


zerosnacks:
### Component

Forge

### Describe the feature you would like

Currently we require passing `--interactives 1` rather than `--interactive` (as used in other commands). This is confusing for regular users.

dipanshuhappy:
Happy to work on this

zerosnacks:
Thanks @dipanshuhappy, assigned!

zerosnacks:
@dipanshuhappy ideally we would want also not to have a user need to pass a `--sender <address>` but rather we just derive this from their private key similar to how we do it in `forge create`'s `--interactive`.

dipanshuhappy:
> `create`'s

got it

grandizzy:
> [@dipanshuhappy](https://github.com/dipanshuhappy) ideally we would want also not to have a user need to pass a `--sender <address>` but rather we just derive this from their private key similar to how we do it in `forge create`'s `--interactive`.

@zerosnacks should we tackle https://github.com/foundry-rs/foundry/issues/10632 here too?

zerosnacks:
> [@zerosnacks](https://github.com/zerosnacks) should we tackle [#10632](https://github.com/foundry-rs/foundry/issues/10632) here too?

For simplicity let's disconnect the two as they are quite distinct but would be a great follow up!

PR #12608, opened 2025-11-19
feat!(`forge script`): add `--interactive` flag for deploying with a single keypair

zerosnacks:
## Motivation

Adds `--interactive` flag mirroring current use in `RawWalletOpts` preventing the user from having to pass `--interactives 1 --sender <SENDER_ADDRESS>`

Difficult to test due to the interactive nature but tested manually

## Solution

Closes: https://github.com/foundry-rs/foundry/issues/12601

Added @dipanshuhappy as co-author as it was originally assigned to them, took over to unblock myself

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

## Breaking changes

Uses the `-i` shorthand for `--interactive` flag mirroring behavior of `RawWalletOpts` that people are used to.

Existing users will be faced with error:

```
[⠊] Compiling...
No files changed, compilation skipped
Error: encode length mismatch: expected 0 types, got 1
```

### DIFF
--- a/crates/cast/src/cmd/wallet/list.rs
+++ b/crates/cast/src/cmd/wallet/list.rs
@@ -58,7 +58,14 @@ impl ListArgs {
             || self.all
             || (!self.ledger && !self.trezor && !self.aws && !self.gcp)
         {
-            let _ = self.list_local_senders();
+            match self.list_local_senders() {
+                Ok(()) => {}
+                Err(e) => {
+                    if !self.all {
+                        sh_err!("{}", e)?;
+                    }
+                }
+            }
         }
 
         // Create options for multi wallet - ledger, trezor and AWS


crates/wallets/src/wallet_multi/mod.rs
```diff
@@ -264,6 +262,9 @@ impl MultiWalletOpts {
         if let Some(mnemonics) = self.mnemonics()? {
             signers.extend(mnemonics);
         }
+        if self.interactive {
+            pending.push(PendingSigner::Interactive);
```
zerosnacks, line 266: We simply push this in the same `pending` queue that `interactives` uses and re-use the existing flow

crates/wallets/src/wallet_multi/mod.rs
```diff
@@ -93,15 +93,13 @@ pub struct MultiWalletOpts {
...
     ///
     /// Takes a value for the number of keys to enter.
-    #[arg(
-        long,
-        short,
-        help_heading = "Wallet options - raw",
-        default_value = "0",
-        value_name = "NUM"
-    )]
+    #[arg(long, help_heading = "Wallet options - raw", default_value = "0", value_name = "NUM")]
```
zerosnacks, line 96: Note the removal of `short`, breaking change in favor of `interactive` carrying the `short` flag

I think this is worth it for UX consistency, https://github.com/search?q=%22forge+script%22+-i+&type=code shows a few very minor examples in unmaintained repos

PR #12611, opened 2025-11-19
feat(anvil): extend Content-Type support on Beacon API

### LINKED ISSUE #12610: Add support for SSZ encoding in Anvil's Beacon API
### Component

Anvil

It would be great if Anvil’s Beacon API could conform to HTTP content type headers and serve responses in SSZ-encoded format based on the client’s content type preference.

### Additional context

See https://ethereum.github.io/beacon-APIs intro paragraph.

mablr:
## Motivation

Close #12610

## Solution

- Support "application/octet-stream" and "application/json" content types
- Implemented `must_be_ssz` helper to determine if the Accept header prefers SSZ encoding.
- Updated `handle_get_blobs` to return SSZ-encoded blobs when requested.
- Enhanced tests to verify both JSON and SSZ responses for blob sidecars.
- Added accept header assertions in tests to ensure correct content type selection.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/server/beacon_handler.rs
+++ b/crates/anvil/src/server/beacon_handler.rs
@@ -9,10 +9,49 @@ use alloy_rpc_types_beacon::{
 use axum::{
     Json,
     extract::{Path, Query, State},
+    http::HeaderMap,
     response::{IntoResponse, Response},
 };
-use hyper::StatusCode;
+use ssz::Encode;
 use std::{collections::HashMap, str::FromStr as _};
+
+/// Helper function to determine if the Accept header indicates a preference for SSZ (octet-stream)
+/// over JSON.
+pub fn must_be_ssz(headers: &HeaderMap) -> bool {
+    headers
+        .get(axum::http::header::ACCEPT)
+        .and_then(|v| v.to_str().ok())
+        .map(|accept_str| {
+            let mut octet_stream_q = 0.0;
+            let mut json_q = 0.0;
+
+            // Parse each media type in the Accept header
+            for media_type in accept_str.split(',') {
+                let media_type = media_type.trim();
+                let quality = media_type
+                    .split(';')
+                    .find_map(|param| {
+                        let param = param.trim();
+                        if let Some(q) = param.strip_prefix("q=") {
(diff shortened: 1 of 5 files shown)


crates/anvil/src/server/beacon_handler.rs
```diff
@@ -9,11 +9,50 @@ use alloy_rpc_types_beacon::{
...
+    http::HeaderMap,
     response::{IntoResponse, Response},
 };
-use hyper::StatusCode;
+use ssz::Encode;
 use std::{collections::HashMap, str::FromStr as _};
 
+/// Helper function to determine if the Accept header indicates a preference for SSZ (octet-stream)
+/// over JSON.
+pub fn must_be_ssz(headers: &HeaderMap) -> bool {
```
grandizzy, line 20: nit: shouldn't be public

crates/anvil/src/server/beacon_handler.rs
```diff
@@ -9,11 +9,50 @@ use alloy_rpc_types_beacon::{
...
+    headers
+        .get(axum::http::header::ACCEPT)
+        .and_then(|v| v.to_str().ok())
+        .map(|accept_str| {
+            let mut octet_stream_q = 0.0;
+            let mut json_q = 0.0;
+
+            // Parse each media type in the Accept header
+            for media_type in accept_str.split(',') {
```
grandizzy, line 29: should we worry if same media type present multiple times? (I think that's technically allowed) - probably not an issue as we'll consider the last one

crates/anvil/src/server/beacon_handler.rs
mablr: This was merged too fast 😅

I may fix it as a follow-up. Btw Anvil's `server` module can be split/reorganized for clarity:
- crates/anvil/src/server/mod.rs
- crates/anvil/src/server/beacon/
- crates/anvil/src/server/rpc/

I can wrap it all in one PR.

grandizzy: thank you! yeah not a prio, can include in future PR

PR #12612, opened 2025-11-19
fix(common): replace Solidity link placeholders with 20 zero bytes

radik878:
This change fixes the placeholder replacement to match Solidity’s library linking https://docs.soliditylang.org/en/latest/using-the-compiler.html#library-linking semantics by replacing each __$...$__ region with exactly 20 zero bytes (40 hex chars) instead of 40 bytes. The Solidity documentation specifies that the placeholder is a 40-character span replaced in place by a 20-byte address encoded as 40 hex characters; emitting 80 hex characters was inflating bytecode lengths and could desynchronize offsets used by source maps and PC/IC mappings. The adjacent comment was also corrected to reflect the intended 20-byte replacement, aligning the code, comments, and external specification.

### DIFF
--- a/crates/common/src/utils.rs
+++ b/crates/common/src/utils.rs
@@ -82,8 +82,8 @@ pub fn strip_bytecode_placeholders(bytecode: &BytecodeObject) -> Option<Bytes> {
     match &bytecode {
         BytecodeObject::Bytecode(bytes) => Some(bytes.clone()),
         BytecodeObject::Unlinked(s) => {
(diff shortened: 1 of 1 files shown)


onbjerg (changes requested):
on the link you provided there is nothing that backs up the motivation for the pr:

>The placeholder is a 34 character prefix of the hex encoding of the keccak256 hash of the fully qualified library name. The bytecode file will also contain lines of the form // <placeholder> -> <fq library name> at the end to help identify which libraries the placeholders represent.

PR #12615, opened 2025-11-20
feat: use different gas calc for monad mainnet

QEDK:
Use different gas calculation for Monad mainnet, also bumps the lockfile to use `alloy-chains = 0.2.20` to include the new chain. If y'all could also cut a new release for this commit, we'd really appreciate it!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -67,9 +67,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy-chains"
-version = "0.2.18"
+version = "0.2.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


grandizzy:
@QEDK alloy chains was already bumped to include monad in https://github.com/foundry-rs/foundry/pull/12613

grandizzy (approved):

QEDK:
> @QEDK alloy chains was already bumped to include monad in #12613

Ah, I meant the Foundry release so that devs can use this on stable.

grandizzy:
> > @QEDK alloy chains was already bumped to include monad in #12613
> 
> Ah, I meant the Foundry release so that devs can use this on stable.

it will be available in nightly builds when this complete https://github.com/foundry-rs/foundry/actions/runs/19524606561 Mind though it won't be part of stable as it is not cut yet

PR #12618, opened 2025-11-20
fix: handle check_sequence errors during shrinking

Bashmunta:
In shrink.rs, shrink_sequence previously swallowed Err from check_sequence via a catch-all match arm, implicitly treating executor/backend errors as valid shrink outcomes. This is incorrect because Solidity-level reverts are represented as Ok(RawCallResult) and Err typically denotes infrastructure failures (backend/inspector). The new logic explicitly restores the removed call on Err, tracks consecutive errors, and aborts shrinking after a small threshold by returning the error. This prevents false shrinking, surfaces real executor issues, and aligns with how replays and runners treat errors elsewhere in the codebase.

### DIFF
--- a/crates/evm/evm/src/executors/invariant/shrink.rs
+++ b/crates/evm/evm/src/executors/invariant/shrink.rs
@@ -61,6 +61,8 @@ pub(crate) fn shrink_sequence(
     let mut call_idx = 0;
 
     let mut shrinker = CallSequenceShrinker::new(calls.len());
+    // Track consecutive executor errors during shrinking to avoid producing misleading results.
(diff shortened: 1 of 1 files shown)


grandizzy:
this is not correct.

https://github.com/foundry-rs/foundry/blob/c8ff851e025b7faeaf2e43130ae02d2aed1957f3/crates/evm/evm/src/executors/invariant/shrink.rs#L72-L86

means that we test entire sequence that produced invariant fail and if we cannot replay entirely (that is an error occured) then we keep the original sequence, then restart checking it

PR #12622, opened 2025-11-20
chore(anvil): use `FillTransaction` from alloy

mablr:
## Motivation

Follow-up for #12595 and https://github.com/alloy-rs/alloy/pull/3210

## Solution

- Bumped alloy to 1.1.2
- Added alloy-rpc-types-eth dep
- Removed `FillTransactionResult` from alloy-core crate*

*I think this is okay (no need to re-export), as `FillTransaction` is used only in anvil crate.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -79,9 +79,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-consensus"
-version = "1.1.1"
+version = "1.1.2"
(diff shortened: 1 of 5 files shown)


grandizzy (approved):
thank you!

PR #10623, opened 2025-05-24
fix: eip7702 cheatcodes multiple auth

zjesko:
fixes #10551

Add the ability to have multiple active delegations and signed auth. Also takes cares of correctly incrementing the nonce.

```
vm.signAndAttach
vm.signAndAttach
vm.signAndAttach
performTx
```
now works correctly

PR #12627, opened 2025-11-21
feat(cast): accept multiple 7702 authorizations

0xferrous:
## Motivation

There can be multiple 7702 authorizations provided in a single transaction. So, let's support that.

## Solution

We allow at max one unsigned/address as authorization since otherwise it'd be signing multiple authorizations for delegating same account which will just make the last one execute and all the previous ones will essentially be noop. So, for multiple authorizations, the expectation is that signed authorizations will be provided as the args.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cast/src/cmd/send.rs
+++ b/crates/cast/src/cmd/send.rs
@@ -126,7 +126,7 @@ impl SendTxArgs {
         {
             // ensure we don't violate settings for transactions that can't be CREATE: 7702 and 4844
             // which require mandatory target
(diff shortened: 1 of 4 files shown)


mattsse (approved):
this lgtm and makes sense

pending @grandizzy

grandizzy (approved):
thank you, checks out. Similar with what we did for cheatcodes in https://github.com/foundry-rs/foundry/pull/10623 by switching from an option to a vec of signed auth

PR #12628, opened 2025-11-21
Collapse redundant Windows pipe path check in provider

Doryu001:
merge the nested if let/if in resolve_path into a single condition, keep the Windows named-pipe fast path while removing the clippy::collapsible_if warning


onbjerg (approved):

Doryu001:
we can do migrate

PR #12634, opened 2025-11-22
feat(cast): validate-auth

0xferrous:
## Motivation

There is no way to determine the validity of the authorizations just looking at the RPC responses with `cast tx` or `cast receipt`.

## Solution

So, we need to build a new command that manually validates the authorizations with the logic laid out in EIP-7702.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cast/src/args.rs
+++ b/crates/cast/src/args.rs
@@ -752,6 +752,9 @@ pub async fn run_command(args: CastArgs) -> Result<()> {
             let auth: SignedAuthorization = serde_json::from_str(&auth)?;
             sh_println!("{}", auth.recover_authority()?)?;
         }
+        CastSubcommand::ValidateAuth(cmd) => {
+            cmd.run().await?;
+        }
         CastSubcommand::TxPool { command } => command.run().await?,
         CastSubcommand::Erc20Token { command } => command.run().await?,
         CastSubcommand::DAEstimate(cmd) => {
--- a/crates/cast/src/cmd/validate_auth.rs
+++ b/crates/cast/src/cmd/validate_auth.rs
@@ -1,0 +1,229 @@
+use std::collections::HashMap;
+
+use alloy_consensus::Transaction;
+use alloy_eips::{BlockId, eip7702::SignedAuthorization};
+use alloy_network::{AnyNetwork, TransactionResponse};
+use alloy_primitives::{Address, B256};
+use alloy_provider::Provider;
+use eyre::OptionExt;
+use foundry_cli::{
+    opts::RpcOpts,
+    utils::{self, LoadConfig, init_progress},
+};
+use foundry_common::shell;
+
+#[derive(Debug, clap::Parser)]
+pub struct ValidateAuthArgs {
+    /// Transaction hash.
+    tx_hash: B256,
+
+    #[command(flatten)]
+    rpc: RpcOpts,
+}
+
+/// Validates an authorization and updates the running nonce map if valid.
+/// Returns (valid_chain, valid_nonce, expected_nonce, authority) if signature recovery succeeds.
+async fn validate_and_update_nonce<P: Provider<AnyNetwork>>(
+    auth: &SignedAuthorization,
+    chain_id: u64,
+    block_number: u64,
+    running_nonces: &mut HashMap<Address, u64>,
+    provider: &P,
+) -> eyre::Result<Option<(bool, bool, u64, Address)>> {
+    let valid_chain = auth.chain_id == chain_id || auth.chain_id == 0;
+
+    match auth.recover_authority() {
+        Ok(authority) => {
+            // Get expected nonce for this authority
+            let expected_nonce = if let Some(&nonce) = running_nonces.get(&authority) {
+                nonce
+            } else {
+                // Fetch nonce at block - 1 (state before this block)
+                let prev_block = BlockId::number(block_number - 1);
+                provider.get_transaction_count(authority).block_id(prev_block).await?
+            };
+
+            let valid_nonce = auth.nonce == expected_nonce;
+
+            // If authorization was valid, update running nonce
+            if valid_chain && valid_nonce {
+                running_nonces.insert(authority, expected_nonce + 1);
+            }
+
+            Ok(Some((valid_chain, valid_nonce, expected_nonce, authority)))
+        }
+        Err(_) => Ok(None),
+    }
+}
+
+impl ValidateAuthArgs {
+    /// Validates all the authorizations in an EIP-7702 transaction. It does so by validating the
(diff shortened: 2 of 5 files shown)


crates/cast/src/opts.rs
```diff
@@ -1119,6 +1119,16 @@ pub enum CastSubcommand {
     #[command(visible_aliases = &["decode-auth"])]
     RecoverAuthority { auth: String },
 
+    /// Validate EIP-7702 authorizations in a transaction and print validity status.
+    #[command(name = "validate-auth", visible_aliases = &["va", "validate-auths"])]
+    ValidateAuth {
+        /// Transaction hash.
+        tx_hash: String,
```
mattsse: let's make this a B256, then we don't need to manually parse this

crates/cast/src/args.rs
```diff
@@ -751,6 +755,147 @@ pub async fn run_command(args: CastArgs) -> Result<()> {
             let auth: SignedAuthorization = serde_json::from_str(&auth)?;
             sh_println!("{}", auth.recover_authority()?)?;
         }
+        CastSubcommand::ValidateAuth { tx_hash, rpc } => {
```
mattsse: since this logic isn't just ~2lines, we should introduce `struct ValidateAuthArgs` and move this into `ValidateAuth::run`

crates/cast/src/opts.rs
```diff
@@ -1119,6 +1119,16 @@ pub enum CastSubcommand {
     #[command(visible_aliases = &["decode-auth"])]
     RecoverAuthority { auth: String },
 
+    /// Validate EIP-7702 authorizations in a transaction and print validity status.
```
mattsse, line 1122: what is being validated here?
just the auth itself?
because for this we only need to fetch the tx itself and perform recovery on this, then this can be simplified

this info is also already included in cast tx (only displays the recovered auth if valid)
so I'm not actually sure how exactly this new command is different

mattsse (commented):
> There is no way to determine the validity of the authorizations just looking at the RPC responses with cast tx

this isn't quite accurate because: we display: `recoveredAuthority` if the auth is valid

```
authorizationList    [
        {recoveredAuthority: 0x6d88AED96570f99f17803AAadCB7d52a903A8000, signedAuthority: {"chainId":"0x1","address":"0xa845c74344fc9405b1fcf712f04668979573c1bf","nonce":"0x0","yParity":"0x0","r":"0x65f49a34f31703fec5612c7022ad9ad3411524f3265f926ff3ad9710ee3c0495","s":"0x254672cc95922e76b069ee1458155cb74d1b179ba1d26239ffb4b220816982e2"}}
]

```

crates/cast/src/opts.rs
0xferrous: I don't think the `recoveredAuthority` in itself tells you whether the authorization was valid because, the signature can be valid, but the nonce may be outdated or too ahead(there is only one acceptable nonce), the chain id can be invalid, etc. Essentially, my intention was to determine which authorizations got "executed", and which got rejected, and this is my attempt at displaying that information.

I think it shows the decoded data and recovered authority from the signature, but it doesn't show the the validity. I had to refer to etherscan for instance when I couldn't figure out why my eoa was not getting delegated 
image

crates/cast/src/cmd/validate_auth.rs
```diff
@@ -0,0 +1,159 @@
...
+};
+
+#[derive(Debug, clap::Parser)]
+pub struct ValidateAuthArgs {
+    /// Transaction hash.
+    tx_hash: B256,
+
+    #[command(flatten)]
+    rpc: RpcOpts,
+}
```
mattsse, line 22: this needs some docs, what this actually does

crates/cast/src/cmd/validate_auth.rs
```diff
@@ -0,0 +1,159 @@
...
+
+        // Extract authorization list from EIP-7702 transaction
+        let auth_list = match &*tx.inner.inner {
+            AnyTxEnvelope::Ethereum(TxEnvelope::Eip7702(signed_tx)) => {
+                signed_tx.tx().authorization_list.clone()
+            }
+            _ => {
+                eyre::bail!("transaction is not an EIP-7702 transaction");
+            }
+        };
```
mattsse: I believe we can use the `trait Transaction's` function for this

crates/cast/src/cmd/validate_auth.rs
```diff
@@ -0,0 +1,159 @@
...
+        // Fetch the block to get all transactions up to this one
+        let block = provider
+            .get_block_by_number(block_number.into())
+            .full()
+            .await?
+            .ok_or_else(|| eyre::eyre!("block not found: {}", block_number))?;
+
+        // Build a map of address -> running nonce from txs in this block up to and including
+        // our tx
+        let mut running_nonces: HashMap<Address, u64> = HashMap::new();
```
mattsse: why do we need this?

mattsse (changes requested):
still a bit unclear what this exactly validates

could you please add some additional docs

crates/cast/src/cmd/validate_auth.rs
onbjerg: i assume this is because naively getting the nonce at the block before this assumes there is only 1 transaction from the sender account in the current block, so we need to account for the fact that the account might send 1 or more txs before their eip7702 authorization which would bump the nonce.

0xferrous: yes, although i have realized my implementation is still incomplete, because all the _valid_ authorizations before the current tx in the given block can also have an authorization that can affect the running nonce of current recovered authority, which means i have to validate authorizations for all txs before current tx to get the _real_ running nonce and accurate validation.

crates/cast/src/cmd/validate_auth.rs
```diff
@@ -0,0 +1,229 @@
...
+    /// - For each valid authorization in previous transactions, also increments that authority's
+    ///   running nonce (since valid authorizations execute and increment the authority's nonce)
+    /// - If an authority is not in the running nonce map, fetches their nonce at block - 1
+    ///
+    /// Then, for each authorization in the target transaction:
+    /// - Validates nonce against the running nonce
+    /// - If the authorization is valid (both chain and nonce), increments the running nonce for
+    ///   that authority (for subsequent authorizations in the same transaction)
+    ///
+    /// For chain id validation, it checks if it is zero or the same chainid as the current chain.
```
0xferrous, line 75: Does this clarify things @mattsse ?

emmajam:

PR #12638, opened 2025-11-22
refactor(cast): consolidate cast convert

### LINKED ISSUE #12500: feat(`cast`): consolidate conversion commands
### Component

Cast

## Motivation

Cast is getting bloated with excessive conversion commands, increasing cognitive load when trying to remember/find the right one.

## Proposal

Create a `cast convert` subcommand to group all conversion-related commands, similar to how `cast wallet` groups wallet operations.

**Before:**
```bash
cast to-wei 1 ether
cast from-utf8 "hello"
```

**After:**
```bash
cast convert to-wei 1 ether
cast convert from-utf8 "hello"
```

### Additional context

Proposed (26) commands to consolidate under `cast convert`

- `from-wei`, `to-wei`, `to-unit`
- `from-fixed-point`, `to-fixed-point`
- `format-units`, `parse-units`
- `to-hex`, `to-dec`, `to-base`
- `to-int256`, `to-uint256`
- `from-utf8`, `to-utf8`, `to-ascii`, `from-bin`
- `to-bytes32`, `to-hexdata`
- `format-bytes32-string`, `parse-bytes32-string`, `parse-bytes32-address`
- `to-check-sum-address`, `concat-hex`, `pad`
- `to-rlp`, `from-rlp`

dipanshuhappy:
## Motivation
Resolves #12500 

## Solution
Create a new file ```convert.rs``` which encapsulates all the sub commands mentioned. 

## PR Checklist
NOTE: This does break existing cli commands related to conversation. Refer to the issue to know more about the cli commands that will break.
- [ ] Added Tests
- [x] Added Documentation
- [x] Breaking changes


grandizzy (changes requested):
thank you will need to fix tests to move forward

dipanshuhappy:
> thank you will need to fix tests to move forward

Updated

grandizzy (dismissed):
nice, ty, that's much cleaner. pending other reviewers as this is a breaking change and need consensus.

0xClandestine:
@dipanshuhappy appreciate the work ser

onbjerg (changes requested):
I don't think we should merge this as is as it is breaking. What I would suggest is shimming the existing commands to call `convert <...>` internally and displaying a "This command is deprecated and may be removed in the future, please use `cast convert <...>` instead" message.

I realize this ends up making the code a bit messy again, but I'd much prefer this to us pushing a breaking change for a very common set of commands in a non-major bump

dipanshuhappy:
> I don't think we should merge this as is as it is breaking. What I would suggest is shimming the existing commands to call `convert <...>` internally and displaying a "This command is deprecated and may be removed in the future, please use `cast convert <...>` instead" message.
> 
> I realize this ends up making the code a bit messy again, but I'd much prefer this to us pushing a breaking change for a very common set of commands in a non-major bump

Understood... Updating to that ASAP.

0xClandestine:
curious what the status is here?

emmajam:

PR #12406, opened 2025-11-01
refactor(anvil): use alloy envelope macro

mablr:
## Motivation

Close #12389

## Solution

- Updated TypedTransaction to use TransactionEnvelope
- Removed unused/redundant code

## PR Checklist

- [x] Added Tests (some updates due to breaking change on tx serialization)
- [ ] Added Documentation
- [x] Breaking changes (tx serialization)

PR #12639, opened 2025-11-22
refactor(anvil): use alloy's `effective_gas_price` in `Backend::mined_transaction_receipt`

mablr:
## Motivation

#12406 follow-up

Clean up the Anvil codebase by removing custom implementations that are no longer needed.

## Solution

- replace custom `effective_gas_price` impl w/ alloy's `Transaction` trait fn
 
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes (maybe ... https://github.com/foundry-rs/foundry/pull/12639#discussion_r2554255622)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -408,8 +408,8 @@ dependencies = [
  "derive_more",
  "foldhash 0.2.0",
  "getrandom 0.3.4",
- "hashbrown 0.16.0",
- "indexmap 2.12.0",
+ "hashbrown 0.16.1",
+ "indexmap 2.12.1",
  "itoa",
  "k256",
  "keccak-asm",
@@ -835,7 +835,7 @@ dependencies = [
  "alloy-sol-macro-input",
  "const-hex",
  "heck",
- "indexmap 2.12.0",
+ "indexmap 2.12.1",
  "proc-macro-error2",
  "proc-macro2",
  "quote",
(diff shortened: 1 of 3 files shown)


mattsse (dismissed):

mattsse:
I think we need to update the failing transaction_receipt test

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -3120,26 +3120,7 @@ impl Backend {
...
-                .map_or(self.base_fee() as u128, |g| g as u128)
-                .saturating_add(t.tx().tx().max_priority_fee_per_gas),
-            TypedTransaction::EIP7702(t) => block
-                .header
-                .base_fee_per_gas
-                .map_or(self.base_fee() as u128, |g| g as u128)
-                .saturating_add(t.tx().max_priority_fee_per_gas),
-            TypedTransaction::Deposit(_) => 0_u128,
-        };
+        let effective_gas_price = transaction.effective_gas_price(block.header.base_fee_per_gas);
```
mablr, line 3123: Imo using `block.header.base_fee_per_gas` as base fee arg is okay and non-breaking, despite `self.base_fee()` not being used anymore

Indeed after careful review of the blocks init/mining logic, I've not found any scenario where `block.header.base_fee_per_gas` could be `None` at this point. Which would mean that in the previous implementation we never had a fallback to the default value `self.base_fee()` (giving always base_fee_per_gas + max_priority_fee_per_gas for TXs >= eip1559).

So Alloy's impl would "more correct", I guess.

Let me know, if you think i'm wrong.

mablr: If my observation is correct, we could improve robustness by making base_fee non-optional in block header ?

mattsse (approved):

PR #12640, opened 2025-11-22
refactor(anvil): simplify `effective_reward` calculation in `FeeHistoryService`

mablr:
## Motivation
Anvil cleanup in the same vein as #12639 + smol nits

However, I created this PR separately as I’m not entirely certain that this one isn’t breaking, so it requires a more thorough review.

Imo Alloy's helper has the same logic as anvil's custom one https://github.com/alloy-rs/alloy/blob/9770720d77000d48c9d1f9979bf393f3b09b118b/crates/eips/src/eip1559/helpers.rs#L14-L30

## Solution

- Replace custom effective reward calculation with alloy's `Transaction::effective_tip_per_gas()`
- Cleanup useless `base_fee` casts, only one is necessary u64->u128, instead of u64->u128->u64
- More elegant default rewards: `vec![]` instead of heavy `.iter().map().collect()`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes (maybe...)

### DIFF
--- a/crates/anvil/src/eth/fees.rs
+++ b/crates/anvil/src/eth/fees.rs
@@ -6,13 +6,12 @@ use std::{
     task::{Context, Poll},
 };
 
-use alloy_consensus::Header;
+use alloy_consensus::{Header, Transaction};
 use alloy_eips::{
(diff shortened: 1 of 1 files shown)


mattsse (approved):
cool, yeah this is equivalent

PR #12647, opened 2025-11-24
fix: validate indexed arguments count in `cast logs`

maradini77:
When using `cast logs` with an event signature and indexed arguments, extra arguments beyond the number of indexed inputs are silently ignored. This causes the filter to be broader than expected, potentially matching unintended events, which is problematic for automated scripts and monitoring tools.

The issue occurs because `build_filter_event_sig` uses `zip()` to pair event inputs with user-provided arguments, and any excess arguments are simply dropped without notification.


DaniPopes:
this is not right

PR #12658, opened 2025-11-25
refactor: use `into_owned()` instead of `to_string()` on `Cow<str>`

0xLogicalx:
Replaces `.to_string_lossy().to_string()` with `.to_string_lossy().into_owned()` across the codebase for more idiomatic and efficient string conversion from path types.


0xrusowsky:
irrelevant

PR #12662, opened 2025-11-26
fix(cli): correct deprecation note to ProjectPathOpts::get_remappings

Fibonacci747:
The deprecation note on BuildOpts::get_remappings() referenced a non-existent type ProjectPathsArgs, which is misleading and breaks discoverability of the intended replacement. The correct, existing type is ProjectPathOpts, re-exported in 
crates/cli/src/opts/build/mod.rs, and it provides the get_remappings() method used throughout the codebase. No occurrences of ProjectPathsArgs exist elsewhere, indicating this was a typo rather than an intentional alias.

### DIFF
--- a/crates/cli/src/opts/build/core.rs
+++ b/crates/cli/src/opts/build/core.rs
@@ -174,7 +174,7 @@ impl BuildOpts {
     }
 
     /// Returns the remappings to add to the config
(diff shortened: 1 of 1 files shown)


mattsse (commented):
imo it's clear what this means

PR #12680, opened 2025-11-28
refactor(anvil): remove redundant SubscriptionId bound

justinbay:
remove the duplicated PartialEq constraint on SubscriptionId since Eq already implies it, simplify the PubSubRpcHandler API surface without affecting behavior


mattsse (commented):
this doesnt matter really

PR #12688, opened 2025-11-28
Optimize RPC data collection in RpcData::from_transactions

soniseth0:
collapse the double iteration over broadcast transactions into a single pass, keep identical behavior while avoiding repeated any + collect work


mattsse (commented):
this doesnt really optimize

PR #12692, opened 2025-11-29
feat: support cast logs query chunking

### LINKED ISSUE #12687: support cast logs query chunking
### Component

Cast

there's often a length query limit on eth_getlogs:

> Error: server returned an error response: error code -32602: query exceeds max block range 100000

> Error: server returned an error response: error code -32603: eth_getLogs range is too large, max is 1k blocks

> Error: server returned an error response: error code 35: ranges over 10000 blocks are not supported on freetier

so usually 1k or 10k

> Error: server returned an error response: error code -32602: eth_getLogs is limited to 1024 block range. Please check the parameter requirements at  https://docs.blockpi.io/documentations/api-reference

when done with 

```
cast logs --from-block <> --to-block <>
```

Ideally we can detect this and auto chunk this in based on the response, naive option would be to just junk by 10k blocks, that should work for most
but we should also make this a cli arg, like --query-size

the same we have for the matches:

> Error: server returned an error response: error code -32602: query exceeds max results 20000, retry with the range 23879634-23879696

e.g.:  `cast logs --from-block 23879634 --to-block 23889634  "Transfer(address indexed from, address indexed to, uint256 value)"`

so ideally we also figure this out from the response and adjust the size accordingly

### Additional context

_No response_

stevencartavia:
closes #12687 

tested with 
```
./target/debug/cast logs --from-block 23879634 --to-block 23889634 "Transfer(address indexed from, address indexed to, uint256 value)" --query-size 7 --rpc-url https://reth-ethereum.ithaca.xyz/rpc
```

```
./target/debug/cast logs --from-block 23879634 --to-block 23889634 "Transfer(address indexed from, address indexed to, uint256 value)" --query-size 10 --rpc-url ...
```

### DIFF
--- a/crates/cast/src/cmd/logs.rs
+++ b/crates/cast/src/cmd/logs.rs
@@ -47,7 +47,7 @@ pub struct LogsArgs {
 
     /// Number of blocks to query in each chunk when the provider has range limits.
     /// Defaults to 10000 blocks per chunk.
-    #[arg(long, default_value = "10000")]
+    #[arg(long, default_value_t = 10000)]
     query_size: u64,
 
     #[command(flatten)]
--- a/crates/cast/src/lib.rs
+++ b/crates/cast/src/lib.rs
@@ -994,19 +994,12 @@ impl<P: Provider<AnyNetwork> + Clone + Unpin> Cast<P> {
         Ok(res)
     }
 
-    fn extract_block_range(filter: &Filter) -> Result<(Option<u64>, Option<u64>)> {
+    fn extract_block_range(filter: &Filter) -> (Option<u64>, Option<u64>) {
         let FilterBlockOption::Range { from_block, to_block } = &filter.block_option else {
-            return Ok((None, None));
+            return (None, None);
         };
 
-        let extract_number = |block: Option<BlockNumberOrTag>| {
-            block.and_then(|b| match b {
-                BlockNumberOrTag::Number(n) => Some(n),
-                _ => None,
-            })
-        };
-
-        Ok((extract_number(*from_block), extract_number(*to_block)))
+        (from_block.and_then(|b| b.as_number()), to_block.and_then(|b| b.as_number()))
     }
 
     /// Retrieves logs with automatic chunking fallback.
@@ -1038,7 +1031,7 @@ impl<P: Provider<AnyNetwork> + Clone + Unpin> Cast<P> {
     where
         P: Clone + Unpin,
     {
-        let (from_block, to_block) = Self::extract_block_range(filter)?;
+        let (from_block, to_block) = Self::extract_block_range(filter);
         let (Some(from), Some(to)) = (from_block, to_block) else {
             return self.provider.get_logs(filter).await.map_err(Into::into);
         };
--- a/crates/cast/tests/cli/main.rs
+++ b/crates/cast/tests/cli/main.rs
@@ -1611,6 +1611,24 @@ casttest!(logs_sig_2, |_prj, cmd| {
     .stdout_eq(file!["../fixtures/cast_logs.stdout"]);
 });
 
+casttest!(logs_chunked_large_range, |_prj, cmd| {
+    let rpc = next_http_archive_rpc_url();
+    cmd.args([
+        "logs",
+        "--rpc-url",
+        rpc.as_str(),
+        "--from-block",
+        "18000000",
+        "--to-block",
+        "18050000",
+        "--query-size",
+        "1000",
+        "Transfer(address indexed from, address indexed to, uint256 value)",
+        "0xA0b86a33E6441d02dd8C6B2b7E5D1E3eD7F73b4b",
+    ])
+    .assert_success();
+});
+
 casttest!(mktx, |_prj, cmd| {
     cmd.args([
         "mktx",


crates/cast/src/lib.rs
```diff
@@ -981,6 +982,119 @@ impl<P: Provider<AnyNetwork>> Cast<P> {
...
+            let mut s = vec![];
+            for log in logs {
+                let pretty = log
+                    .pretty()
+                    .replacen('\n', "- ", 1) // Remove empty first line
+                    .replace('\n', "\n  "); // Indent
+                s.push(pretty);
+            }
+            s.join("\n")
+        };
```
mattsse, line 993: this is duplicated, so we can reuse this

crates/cast/src/lib.rs
```diff
@@ -981,6 +982,119 @@ impl<P: Provider<AnyNetwork>> Cast<P> {
...
+        Ok(res)
+    }
+
+    async fn get_logs_chunked(&self, filter: &Filter, chunk_size: u64) -> Result<Vec<Log>> {
+        // Try the full range first
+        match self.provider.get_logs(filter).await {
+            Ok(logs) => Ok(logs),
+            Err(e) => {
```
mattsse: this section is a bit hard to follow, could we rewrite this a bit so that it's more obvious what it does

crates/cast/src/lib.rs
```diff
@@ -981,6 +982,119 @@ impl<P: Provider<AnyNetwork>> Cast<P> {
...
+                s.push(pretty);
+            }
+            s.join("\n")
+        };
+        Ok(res)
+    }
+
+    async fn get_logs_chunked(&self, filter: &Filter, chunk_size: u64) -> Result<Vec<Log>> {
+        // Try the full range first
+        match self.provider.get_logs(filter).await {
```
mattsse: this ignores the chunksize for this inital attempt

crates/cast/src/lib.rs
```diff
@@ -981,6 +982,119 @@ impl<P: Provider<AnyNetwork>> Cast<P> {
...
+                    .replacen('\n', "- ", 1) // Remove empty first line
+                    .replace('\n', "\n  "); // Indent
+                s.push(pretty);
+            }
+            s.join("\n")
+        };
+        Ok(res)
+    }
+
+    async fn get_logs_chunked(&self, filter: &Filter, chunk_size: u64) -> Result<Vec<Log>> {
```
mattsse: I believe this entire impl is more akin to a stream, we could also further optimize this by doing more requests concurrently.

could we try a dedicated QueryStream type herps? that polls a futuresOrdered of request or similar?

mattsse (changes requested):
a few suggestions

crates/cast/src/lib.rs
stevencartavia: it's trying the full range first if possible

crates/cast/src/lib.rs
```diff
@@ -981,6 +990,129 @@ impl<P: Provider<AnyNetwork>> Cast<P> {
...
+        &self,
+        filter: &Filter,
+        chunk_size: u64,
+    ) -> Result<Vec<Log>>
+    where
+        P: Clone + Unpin,
+    {
+        let (from_block, to_block) = Self::extract_block_range(filter)?;
+        let (Some(from), Some(to)) = (from_block, to_block) else {
+            return Err(eyre::eyre!("Cannot extract valid block range from filter"));
```
mattsse: in this case we should just return self.get_logs

crates/cast/src/lib.rs
```diff
@@ -981,6 +990,129 @@ impl<P: Provider<AnyNetwork>> Cast<P> {
...
+
+        if from >= to {
+            return Ok(vec![]);
+        }
+
+        let total_blocks = to - from;
+
+        // Use very high concurrency for tiny chunks, moderate for small chunks
+        let max_concurrency = if chunk_size <= 10 {
+            std::cmp::min(200, total_blocks as usize) // Up to 200 concurrent requests for tiny chunks
```
mattsse: this might lead to rate limits, we can use `.buffered(5)` via stream::iter instead of futuresordered 

which has native support for this

crates/cast/src/lib.rs
```diff
@@ -981,6 +990,129 @@ impl<P: Provider<AnyNetwork>> Cast<P> {
...
+        let (from_block, to_block) = Self::extract_block_range(filter)?;
+        let (Some(from), Some(to)) = (from_block, to_block) else {
+            return Err(eyre::eyre!("Cannot extract valid block range from filter"));
+        };
+
+        if from >= to {
+            return Ok(vec![]);
+        }
+
+        let total_blocks = to - from;
```
mattsse: there's also `.step_by()` from iterator which we could use to compute the individual ranges like

```
for (chunk_start, chunk_end) in (start..=end)
    .step_by(chunk_size)
    .map(|s| (s, (s + chunk_size).min(end)))
```

crates/cast/src/cmd/logs.rs
```diff
@@ -45,14 +45,27 @@ pub struct LogsArgs {
     #[arg(long)]
     subscribe: bool,
 
+    /// Number of blocks to query in each chunk when the provider has range limits.
+    /// Defaults to 10000 blocks per chunk.
+    #[arg(long, default_value = "10000")]
```
mattsse: I think we can use default_value_t = 10000 here
```suggestion
    #[arg(long, default_value_t = 10000)]
```

crates/cast/src/lib.rs
```diff
@@ -981,6 +994,107 @@ impl<P: Provider<AnyNetwork>> Cast<P> {
...
+        };
+
+        let extract_number = |block: Option<BlockNumberOrTag>| {
+            block.and_then(|b| match b {
+                BlockNumberOrTag::Number(n) => Some(n),
+                _ => None,
+            })
+        };
+
+        Ok((extract_number(*from_block), extract_number(*to_block)))
```
mattsse: let's also make this a helper on filter type itself by adding a function for this on alloy?

crates/cast/src/lib.rs
```diff
@@ -981,6 +994,107 @@ impl<P: Provider<AnyNetwork>> Cast<P> {
...
+    /// First tries to fetch logs for the entire range. If that fails,
+    /// falls back to concurrent chunked requests with rate limiting.
+    async fn get_logs_chunked(&self, filter: &Filter, chunk_size: u64) -> Result<Vec<Log>>
+    where
+        P: Clone + Unpin,
+    {
+        // Try the full range first
+        if let Ok(logs) = self.provider.get_logs(filter).await {
+            return Ok(logs);
+        }
```
mattsse, line 1016: ~~we should log the error here and we also need to consider that this might have failed for other reasons to provider better error responses, e.g. rn we might fail the request and then return an error if we can exract the range in get_logs_chunked_concurrent~~

crates/cast/src/lib.rs
```diff
@@ -981,6 +994,107 @@ impl<P: Provider<AnyNetwork>> Cast<P> {
...
+    fn extract_block_range(filter: &Filter) -> Result<(Option<u64>, Option<u64>)> {
+        let FilterBlockOption::Range { from_block, to_block } = &filter.block_option else {
+            return Ok((None, None));
+        };
+
+        let extract_number = |block: Option<BlockNumberOrTag>| {
+            block.and_then(|b| match b {
+                BlockNumberOrTag::Number(n) => Some(n),
+                _ => None,
+            })
```
mattsse: I think this is just as_number()

crates/cast/src/lib.rs
```diff
@@ -981,6 +994,107 @@ impl<P: Provider<AnyNetwork>> Cast<P> {
...
+        &self,
+        filter: &Filter,
+        chunk_size: u64,
+    ) -> Result<Vec<Log>>
+    where
+        P: Clone + Unpin,
+    {
+        let (from_block, to_block) = Self::extract_block_range(filter)?;
+        let (Some(from), Some(to)) = (from_block, to_block) else {
+            return self.provider.get_logs(filter).await.map_err(Into::into);
```
mattsse, line 1036: this makes sense here

crates/cast/src/lib.rs
```diff
@@ -981,6 +994,107 @@ impl<P: Provider<AnyNetwork>> Cast<P> {
...
+    /// 5 concurrent requests. Falls back to single-block queries if chunks fail.
+    async fn get_logs_chunked_concurrent(
+        &self,
+        filter: &Filter,
+        chunk_size: u64,
+    ) -> Result<Vec<Log>>
+    where
+        P: Clone + Unpin,
+    {
+        let (from_block, to_block) = Self::extract_block_range(filter)?;
```
mattsse: I dont think this returns an error?

crates/cast/src/lib.rs
stevencartavia: done, and updated here for now

mattsse (approved):
cool, tysm!

PR #12693, opened 2025-11-29
feat(lint): lower default verbosity

### LINKED ISSUE #12635: forge lint: reduce verbosity
### Component

Forge

(Feedback from devs) forge lint is too verbose, improve by
- disable inline assembly by default
- review other rules and make it have more sane output by default

### Additional context

_No response_

dipanshuhappy:
Resolves #12635 
## Solution

I made the severity by default to be low 
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/forge/src/cmd/lint.rs
```diff
@@ -103,7 +103,11 @@ impl LintArgs {
             .with_description(true)
             .with_lints(include)
             .without_lints(exclude)
-            .with_severity(if severity.is_empty() { None } else { Some(severity) })
+            .with_severity(if severity.is_empty() {
+                Some(vec![Severity::Low])
+            } else {
+                Some(severity)
+            })
```
0xrusowsky: this is incorrect, what we should do is that when no severity is informed we exclude the `Severity::Gas` lints

additionally forge lint is also invoked via forge build, so rather than changing the mulitple entry points, it may be better to change the behavior in the lint crate

0xrusowsky (changes requested):
besides the lint-related comment that i left, why are you modifying cast?

please remove all of that

crates/forge/src/cmd/lint.rs
dipanshuhappy: Ah Got it. On it

dipanshuhappy: > this is incorrect, what we should do is that when no severity is informed we exclude the `Severity::Gas` lints
> 
> additionally forge lint is also invoked via forge build, so rather than changing the mulitple entry points, it may be better to change the behavior in the lint crate

Updated

PR #12703, opened 2025-12-01
fix(config): prefer matching chain in ResolvedEtherscanConfigs::find_chain

Galoretka:
The previous iteration returned the first error in the etherscan configs, even if a valid config for the requested chain existed later. This was inconsistent with the function’s doc (“Returns the first config that matches the chain”) and caused spurious failures when unrelated entries were misconfigured.
The new logic returns a matching Ok config if present; otherwise, it prefers UnknownChain errors for the requested chain, then falls back to the first error, else None. Added tests to enforce the intended behavior.

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -3273,6 +3273,77 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


DaniPopes:
dont think this is right

PR #12704, opened 2025-12-02
Remove redundant double .into() conversion in RPC error methods

Sharilleed223:
Changes the generic constraint from `Into<String>` to `Into<Cow<'static, str>>` and removes the double `.into()` conversion in three RPC error methods. Both `String` and `&str` implement `Into<Cow<'static, str>>` directly, so the intermediate conversion through `String` is unnecessary.


mattsse (commented):
we can keep this as is

Issue #12394, opened 2025-10-31
chore: strip `WalletOpts` and `EtherscanOpts` from cast subcommands that dont expect a signer or need etherscan api


0xferrous:
### Component

Cast

### Describe the feature you would like

The `WalletOpts` and `EtherscanOpts` are part of the the `EthereumOpts` which is used in several `CastSubcommand`

https://github.com/foundry-rs/foundry/blob/master/crates/cli/src/opts/rpc.rs#L186-L197

<img width="745" height="243" alt="Image" src="https://github.com/user-attachments/assets/4b806660-a831-4668-82cc-e8f91200e866" />

of which:
1. `logs`
    - doesn't need etherescan api
    - doesn't need signer
2. `call`
    - doesn't need etherscan api
    - doesn't need signer
    - should accept optional `--from`
3. `estimate`
    - doesn't need etherscan api
    - doesn't need signer
    - should accept optional `--from`
4. `access_list`
    - doesn't need etherscan api
    - doesn't need signer
    - should accept optional `--from`
 
Why should we eliminate them?
1. wallet options bloat the `--help` with extra 66 lines
2. `EtherescanOpts` is actually relatively small, only two fields, of which the `chain` field seems like it shouldn't belong there anyway, should be part of the `EthereumOpts` or `RpcOpts`. So, it doesn't bloat the help that much, so I don't have any strong opinions here.

Similarly, I think any such unused options bloat elsewhere(forge probably) should be removed too to make the help page more useful and easy to navigate.

PR #12705, opened 2025-12-02
feat(cast): strip WalletOpts and EtherscanOpts from cast subcommands that dont expect a signer or need etherscan api

tskoyo:
Closes https://github.com/foundry-rs/foundry/issues/12394

## Motivation
Went through cast subcommands that don't require WalletOpts and EtherscanOpts and removed EthereumOpts.
RpcOpts is the only type that is being used.

## Solution
Here's where EthereumOpts is still required:
image

## PR Checklist

- [ ] Added Tests
- [ X] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cast/src/cmd/call.rs
+++ b/crates/cast/src/cmd/call.rs
@@ -240,7 +240,6 @@ impl CallArgs {
 
         let provider = utils::get_provider(&config)?;
         let sender = SenderKind::from_wallet_opts(wallet).await?;
-
         let from = sender.address();
 
         let code = if let Some(CallSubcommands::Create {


crates/cast/src/cmd/estimate.rs
```diff
@@ -74,11 +77,16 @@ pub enum EstimateSubcommands {
...
     pub async fn run(self) -> Result<()> {
-        let Self { to, mut sig, mut args, mut tx, block, cost, eth, command } = self;
+        let Self { to, mut sig, mut args, mut tx, block, cost, from, rpc, command } = self;
 
-        let config = eth.load_config()?;
+        let config = rpc.load_config()?;
         let provider = utils::get_provider(&config)?;
-        let sender = SenderKind::from_wallet_opts(eth.wallet).await?;
+
+        let sender = if let Some(from) = from {
```
DaniPopes: idk if this is right, we can probably keep WalletOpts too since we accept it already? @mattsse

tskoyo: I think you're actually right because I just should've removed `EtherscanOpts` that was used in [EthereumOpts](https://github.com/foundry-rs/foundry/blob/master/crates/cli/src/opts/rpc.rs#L188)

tskoyo: I'll add the change!

tskoyo:
@DaniPopes , any help with the test failure? I don't see any issue when I run them locally

DaniPopes:
unfortunately it's just flaky

crates/cast/src/cmd/call.rs
```diff
@@ -226,7 +239,8 @@ impl CallArgs {
         }
 
         let provider = utils::get_provider(&config)?;
-        let sender = SenderKind::from_wallet_opts(eth.wallet).await?;
+        let sender = SenderKind::from_wallet_opts(wallet).await?;
+
```
DaniPopes:
```suggestion
```

DaniPopes (dismissed):

DaniPopes (approved):

DaniPopes:
Thanks!

PR #11842, opened 2025-09-26
perf(fuzz): parallel stateless fuzzing

yash-atreya:
## Motivation

Stacked on #11769 

## Solution

- Introduce `SharedFuzzState` which consists of global values of runs, timer, fail_fast and rejects to determine whether the fuzz test `should_continue` using the `SharedFuzzState::should_continue` 
- Fuzz workers atomically update the values in `SharedFuzzState` 
- For timed campaigns, `timeout` specified in `foundry.toml` is applied to each worker as-is to fit as many runs as possible i.e, `timeout` value is NOT divided by number of workers. 
- `fuzz.runs` config value is divided by the number of workers i.e, `worker_runs = fuzz.runs / num_worker` — reducing the time required for the campaign
- `fuzz.seed` set in config is altered per worker so that the workers are covering different inputs. For `worker0 / master` worker, the provided seed is used as is. For workers with `WorkerID > 0`, the seed is set to `keccak256(fuzz.seed || WorkerID)`
- Currently, the corpus `SYNC_INTERVAL` is naively set to `1000`, we can adjust this after some more empirical feedback 

TODO

- [x] Add `max_rejects` in `SharedFuzzState` to track total rejects across workers and fail accordingly  [40263fd](https://github.com/foundry-rs/foundry/pull/11842/commits/40263fd747d26693d903c8c2cbc5563537cfcfb2)
- [x] Set failures in `run_worker` via `SharedState::try_claim_failure` - This will intercept other workers as well and stop fuzzing entirely [e4c6060](https://github.com/foundry-rs/foundry/pull/11842/commits/e4c60608aa4f172a84900065303b6aef75ae43df)
- [x] Introduce `sync_interval` for the `WorkerCorpus` and call `WorkerCorpus::sync` [ed234d8](https://github.com/foundry-rs/foundry/pull/11842/commits/ed234d8347dd0255c9a6d08b8683af7f88a830b1)
- [x] Introduce `GlobalCorpusMetrics` and sync worker corpus metrics [d973b79](https://github.com/foundry-rs/foundry/pull/11842/commits/d973b79ceb6c8ddcf8fddc3e461c041b32a691b6)
- [x] Determine number of workers - all available cores / `--jobs` [e991826](https://github.com/foundry-rs/foundry/pull/11842/commits/e991826970dee7c561fc64fed8bc30571ba8febd)
- [x] Integrate into `fn fuzz` and run workers using `rayon::IntoParallelIterator` 
[...]

Issue #12397, opened 2025-10-31
bug(invariant): unbounded memory usage in long invariant testing runs


rappie:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

forge Version: 1.4.4-nightly Commit SHA: 65f96cf93b4fab66813f7de14ce50b66c7887eb3 Build Timestamp: 2025-10-31T06:02:26.081956256Z (1761890546) Build Profile: maxperf

### What version of Foundryup are you on?

foundryup: 1.0.0

### What command(s) is the bug in?

forge test -vv --show-progress --match-test invariant

### Operating System

Linux

### Describe the bug

When running invariant testing overnight I noticed it was using more than 100GB of memory.

You can reproduce it using the command listed above on the following repo (use this commit hash):
https://github.com/perimetersec/origin-woeth-fuzz/commit/be6e4df6969c54e769dad6ab912f5d74e8b1e3c1

<img width="625" height="244" alt="Image" src="https://github.com/user-attachments/assets/2b1cc85b-4b51-472b-ae22-5ab4a832a8b9" />
<img width="902" height="215" alt="Image" src="https://github.com/user-attachments/assets/9bff3399-6452-4621-98ee-07150f8c24cd" />
<img width="1353" height="621" alt="Image" src="https://github.com/user-attachments/assets/01ef62f7-d2e4-4337-9d85-8d2e4ce4a471" />

0xalpharush:
Can you share some additional info like how many threads and the progress/log output (use show_progress or https://getfoundry.sh/config/reference/testing#show_edge_coverage)?
```
test/forge/invariant/StaticInvariantTest.sol:StaticInvariantTest
 → invariantHealthy: [60/2000] Runs
  - cumulative edges seen: 41
  - cumulative features seen: 6
  - corpus count: 15
  - favored items: 14
```

I was going to suggest lowering `corpus_min_size`, but it looks like the default is 0 and you don't have it configured...

 Would you see if lowering `corpus_min_mutations` has impact on the memory usage? If it's not the corpus/coverage map that's using the memory, it could be `include_storage` or some other feature that collects values from observing the campagin.

rappie:
It takes a long time to reproduce this. I'm currently in the middle of an engagement and I don't have the bandwidth to do this.

I'm happy to revisit this in about one month. However, you should be able to reproduce this yourself quite easily by using my repo.

[...]

PR #12713, opened 2025-12-02
perf(forge): parallelize stateless fuzzing

DaniPopes:
ref #8898. implemented for stateless fuzzing, statefull to be done in follow-up

Continuation of #11769 + #11842. See these PRs for more details.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -3,6 +3,8 @@
 [profile.default]
 retries = { backoff = "exponential", count = 2, delay = "5s", jitter = true }
 slow-timeout = { period = "30s", terminate-after = 3 }
+# Exclude flaky tests from regular CI runs - they run in nightly test-flaky workflow
+default-filter = "not test(/flaky_/)"
 
 [[profile.default.overrides]]
 filter = "test(/ext_integration/)"
@@ -12,3 +14,10 @@ slow-timeout = { period = "5m", terminate-after = 4 }
 [[profile.default.overrides]]
 filter = "package(foundry-cheatcodes-spec)"
 retries = 0
+
+# Profile for running flaky tests (used by nightly CI and for local debugging)
+# Run with: cargo nextest run --profile flaky
+[profile.flaky]
+default-filter = "test(/flaky_/)"
+retries = { backoff = "exponential", count = 5, delay = "10s", jitter = true }
+slow-timeout = { period = "60s", terminate-after = 3 }
--- a/.github/FLAKY_TEST_FAILURE_TEMPLATE.md
+++ b/.github/FLAKY_TEST_FAILURE_TEMPLATE.md
@@ -1,0 +1,10 @@
+---
+title: "bug: flaky tests workflow failed"
+labels: P-normal, T-bug
+---
+
+The nightly flaky tests workflow has failed. This indicates external API rate limiting, RPC reliability issues, or other intermittent failures that may affect users.
+
+Check the [flaky tests workflow page]({{ env.WORKFLOW_URL }}) for details.
+
+This issue was raised by the workflow at `.github/workflows/test-flaky.yml`.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
-      - uses: crate-ci/typos@bb4666ad77b539a6b4ce4eda7ebb6de553704021 # v1
+      - uses: crate-ci/typos@65120634e79d8374d1aa2f27e54baa0c364fff5a # v1
 
   shellcheck:
     runs-on: depot-ubuntu-latest
@@ -140,7 +140,7 @@ jobs:
         with:
           toolchain: stable
       - uses: rui314/setup-mold@725a8794d15fc7563f59595bd9556495c0564878 # v1
-      - uses: taiki-e/install-action@03ef6f57d573ca4522fb02950f326083373b85bf # v2
+      - uses: taiki-e/install-action@542cebaaed782771e619bd5609d97659d109c492 # v2
         with:
           tool: cargo-hack
       - uses: mozilla-actions/sccache-action@7d986dd989559c6ecdb630a3fd2557667be217ad # v0.0.9
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -23,7 +23,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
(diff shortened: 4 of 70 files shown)


crates/evm/evm/src/executors/fuzz/mod.rs
```diff
@@ -104,282 +79,449 @@ impl FuzzedExecutor {
...
+            1
+        } else if let Some(threads) = self.config.threads {
+            threads as u32
         } else {
-            Ok(FuzzOutcome::CounterExample(CounterExampleOutcome {
-                exit_reason: call.exit_reason,
-                counterexample: (calldata, call),
-                breakpoints,
-            }))
+            rayon::current_num_threads() as u32
```
0xalpharush: It's unclear to me if this would be 0 if threads config is omitted https://github.com/foundry-rs/foundry/blob/a78c89f74f0f4dd26fca74cbb54e31a4ee90b4ac/crates/cli/src/opts/global.rs#L67-L69

If it's all available threads, I'm wondering if it may exacerbate issues like this https://github.com/foundry-rs/foundry/issues/12397. Would it make sense to take memory into account and lower the thread count accordingly or default to a lower num?

DaniPopes: Yes it would, the default is number of threads. i don't think this would have memory issues more than the current implementation, but it does exacerbate storage instead since we never prune anything

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -630,6 +733,384 @@ impl CorpusManager {
...
+
+                debug!(
+                    target: "corpus",
+                    name=%entry.name(),
+                    "moved synced corpus to corpus dir",
+                );
+
+                let corpus_entry = CorpusEntry::new_existing(tx_seq.to_vec(), entry.path.clone())?;
+                self.in_memory_corpus.push(corpus_entry);
+            }
```
0xalpharush, line 867: Should this rm corpus files that didn't bring new coverage?

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -630,6 +733,384 @@ impl CorpusManager {
...
+            debug!("imported {} new corpus entries", imports.len());
+        }
+
+        Ok(imports)
+    }
+
+    /// Syncs and calibrates the in memory corpus and updates the history_map if new coverage is
+    /// found from the corpus findings of other workers.
+    #[instrument(skip_all)]
+    fn calibrate(
```
0xalpharush, line 762: From https://github.com/foundry-rs/foundry/pull/11769#discussion_r2374142332
> AFL ++ stores a bit map of the history_map (1-byte hitcount -> bool indicating hit) for each input and then uses this bit map to make sure it doesn't eject an entry that is unique (this would mean [updating](https://github.com/foundry-rs/foundry/blob/5dcf01b7d32b059a2bfa3855e17af03b5cadf103/crates/evm/evm/src/executors/corpus.rs#L302-L303) is_favored to mean it contributes a unique bit to the bitmap).

I think here we need to add a `minimize` step that finds the smallest set of inputs that cover the intersection of all the edges. Really, a worker should `replace` the in-memory corpus element if we find a smaller sequence or it finds one sequence that is the super set of N sequences. This would apply to both while fuzzing (prior to exporting to master) and also while syncing (master has a smaller seq.)

crates/evm/evm/src/executors/fuzz/mod.rs
```diff
@@ -104,282 +188,438 @@ impl FuzzedExecutor {
...
-                .reverter
-                .is_some_and(|reverter| reverter != address && reverter != CHEATCODE_ADDRESS)
-        {
-            true
-        } else {
-            self.executor.is_raw_call_mut_success(address, &mut call, false)
-        };
+    /// Determines the number of workers to run.
+    fn num_workers(&self) -> usize {
+        rayon::current_num_threads()
```
grandizzy: should we make this a config so we could set number of workers per test? and maybe default to 1 to preserve current behavior?

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -311,46 +397,68 @@ impl CorpusManager {
...
-                corpus_dir.join(format!("{corpus_uuid}{JSON_EXTENSION}.gz")).as_path(),
-                &corpus.tx_seq,
-            )
+        let timestamp = corpus.timestamp;
+        let ext = self.file_extension();
+        let file_path = worker_corpus.join(format!("{corpus_uuid}-{timestamp}{ext}"));
+
+        // Persist to disk.
+        let should_gzip = || {
+            // TODO(dani): actually implement this. other places don't do this calc.
```
grandizzy: this optimization would be really nice to have

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -525,64 +643,49 @@ impl CorpusManager {
...
-            let corpus = self.in_memory_corpus.get(index).unwrap();
-
-            let uuid = corpus.uuid;
-            debug!(target: "corpus", "evict corpus {uuid}");
+            let corpus = &self.in_memory_corpus[index];
 
+            // TODO(dani): metadata?
+            /*
             // Flush to disk the seed metadata at the time of eviction.
+            let uuid = corpus.uuid;
```
grandizzy: let's then totally get rid of this, I think @0xalpharush mentioned not strictly required

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1,8 +1,45 @@
...
+//! └── workerN/                  # Worker N's directory
+//!     ├── corpus/               # Worker N's local corpus
+//!     │   └── ...
+//!     └── sync/                 # Worker 2's sync directory
+//!         └── ...
+//! ```
+//!
+//! ## Workflow
+//!
+//! - Each worker maintains its own local corpus with entries stored as JSON files
```
grandizzy, line 29: do you see any issue if we implement the workers sync mechanism with an in-memory database / notify channel instead files and dump on disk only the interesting corpus like we do now

crates/evm/evm/src/executors/fuzz/mod.rs
DaniPopes: I don't like threads to be a configurable value other than CLI -jX. this is an implementation detail and any weird behavior should be tackled in the implementation not with a config that disables it

crates/evm/evm/src/executors/corpus.rs
DaniPopes: added

crates/evm/evm/src/executors/fuzz/mod.rs
DaniPopes: for example adding a minimum runs per worker f90ee86f0

onbjerg (dismissed):

PR #12716, opened 2025-12-02
perf: compute literals only once

DaniPopes:
`LiteralsDictionary` contains data for the entire project, so it should be computed only once. Currently it is computed for every fuzz/invariant test.

For example, `RUST_LOG=trace forge t` on solady:

image


zerosnacks (approved):

0xrusowsky (approved):
nice, good catch!

PR #12718, opened 2025-12-03
fix: add `log` fn support alongside `log_full` fn, follow-up fix for `revm 33`

zerosnacks:
## Motivation

We should be implementing `log` alongside `log_full` not either. This allows for `vm.expectEmit` to work against logs in cases where the interpreter is not available.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1076,7 +1076,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


crates/cheatcodes/src/test/expect.rs
```diff
@@ -818,14 +822,19 @@ pub(crate) fn handle_expect_emit(
...
-                Error::encode("log emitted 1 time, expected 0"),
-                interpreter.gas,
-            ));
-            return;
+            if let Some(interpreter) = &mut interpreter {
+                // This event was emitted but we expected it NOT to be (count=0)
+                // Fail immediately
+                interpreter.bytecode.set_action(InterpreterAction::new_return(
+                    InstructionResult::Revert,
+                    Error::encode("log emitted but expected 0 times"),
```
0xrusowsky, line 830: optional nit: `log emitted but not expected` or smth like this?

DaniPopes: msg shouldnt change here just keep it the same

PR #12727, opened 2025-12-04
fix(cast): --json support for erc20 cmds

0xferrous:
## Motivation

Enable using `cast erc20` commands in scripts.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


0xferrous:
I think a more general solution where all cast commands are refactored to return a value that implements `std::fmt::Display` and `serde::Serialize` and using that to print the relevant value would be more robust.

0xferrous:
it is not, it is just a way to not have the output like `994462556577 [9.944e11]` so the consumer can parse it as bigint/int

PR #12732, opened 2025-12-04
fix(anvil): resolve incorrect historical states dump behavior

w1tcher:
## Motivation

When I use Anvil with the `--preserve-historical-states` feature, I noticed that the cached state recorded is incorrect.

For example:

```bash
anvil --fork-url <MAINNET_RPC_URL> \
    --fork-block-number 23333467 \
    --state ./state_init.json \
    --chain-id 1 \
    --preserve-historical-states

cast rpc hardhat_setBalance <ADDRESS> "0x1000000000000000000000" \
    -r http://localhost:8545

# First transaction
cast send 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2 "deposit()" \
  --value 1ether \
  --private-key <ADDRESS_KEY> \
  -r http://localhost:8545

# Second transaction
cast send 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2 "deposit()" \
  --value 1ether \
  --private-key <ADDRESS_KEY> \
  -r http://localhost:8545

cast balance 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2 \
    -B 23333468 \
    -r http://localhost:8545
# result: 2311023801249512057471331
```

After executing two transactions, querying the `WETH` balance at block `23333468` returns `2311023801249512057471331`.

At this point, I stop Anvil, restart it using the cached state, and query the `WETH` balance again:
```bash
anvil --fork-url <MAINNET_RPC_URL> \
    --fork-block-number 23333467 \
    --state ./state_init.json \
    --chain-id 1 \
    --preserve-historical-states

cast balance 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2 \
    -B 23333468 \
    -r http://localhost:8545
# result: 2311022801249512057471331
```

I found that the returned `WETH` balance is incorrect.

## Solution

After analyzing the code, I discovered that the cached data being written ultimately comes from `ForkDbStateSnapshot::read_as_state_snapshot`.

```rust
pub struct ForkDbStateSnapshot {
    pub local: CacheDB<SharedBackend>,
    pub state_snapshot: StateSnapshot,
}

impl MaybeFullDatabase for ForkDbStateSnapshot {
    fn maybe_as_full_db(&self) -> Option<&HashMap<Address, DbAccount>> {
        Some(&self.local.cache.accounts)
    }

    fn clear_into_state_snapshot(&mut self) -> StateSnapshot {
        std::mem::take(&mut self.state_snapshot)
    }

    fn read_as_state_snapshot(&self) -> StateSnapshot {
        self.state_snapshot.clone()
    }

    fn clear(&mut self) {
        std::mem::take(&mut self.state_snapshot);
        self.local.clear()
    }

    fn init_from_state_snapshot(&mut self, state_snapshot: StateSnapshot) {
        self.state_snapshot = state_snapshot;
    }
}
```

For `ForkDbStateSnapshot`, its state consists of both `local` and `state_snapshot`.
However, `read_as_state_snapshot` only persists the `state_snapshot` portion,
discarding the contents stored in `local`, which leads to the incorrect behavior.

A correct fix should persist both `local` and `state_snapshot`, and since `ForkDbStateSnapshot` ensures `local` overrides matching keys in `state_snapshot`, the merged result must respect that ordering.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1076,7 +1076,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1100,7 +1100,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -3931,7 +3931,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -6144,7 +6144,7 @@ checksum = "3640c1c38b8e4e43584d8df18be5fc6b0aa314ce6ebf51b53313d4306cca8e46"
 dependencies = [
  "hermit-abi",
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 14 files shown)


mattsse (approved):
oh I see, I believe this is the correct fix, ty!

vincemartinez627-sys (commented):
G

Issue #12702, opened 2025-12-01
feature(forge selectors): Add selectors to cache from a folder of ABIs


JuaniRios:
### Component

Forge

## What
- Add the possibility to add selectors to the foundry cache from a folder of JSON ABIs

## Why
- We had a situation at work where we wanted to decode an error from a complex contract flow, but the error could possibly be in any of multiple contracts and repos, and this decoding needed to happen in a CLI tool. So forcing the CLI user to compile all the possible repos was not a reasonable option. We ended up storing all ABIs in one place and developing a decoder tool. If this would have been possible with `cast` we would have really appreciated it.

## How
- At first I thought about extending `cast decode-error` to accept an optional `--extra-abis` flag that would add the selectors to the in-memory cache by calling `cache.extend_from_abi(json_abi)` from all the files in the folder path.
- George in the telegram group suggested instead to achieve the same result by extending `forge selectors cache` with a new flag that would add all the selectors in the ABIs to the foundry cache. This method makes more sense since it will also be usable for other tools.

## Anything Else?
I will be implementing this feature and submitting a PR soon

### Additional context

<img width="3502" height="1450" alt="Image" src="https://github.com/user-attachments/assets/d0549c3b-5bf8-4bf3-b467-90a8a33bd61d" />

PR #12742, opened 2025-12-05
feat(forge selectors): cache selectors from abis

JuaniRios:
## What
Closes https://github.com/foundry-rs/foundry/issues/12702
Enables `forge selectors cache` to populate the selector cache from a specific directory of JSON ABIs.

## Why
This addresses scenarios where error or transaction decoding is required across multiple contracts or repositories without requiring full compilation of every project. 

For example, when building CLI tools or debugging complex flows, users may need to decode data from contracts where they possess the artifacts (ABIs) but do not wish to (or cannot) compile the full source code. This feature allows `cast` and other tools to utilize a centralized folder of ABIs for resolution.

## How
- Adds a new optional flag: `--extra-abis-path <PATH>`.
- When provided, the command parses all JSON ABIs in the specified folder and adds their selectors to the cache.
- **Behavior:** This is additive. The command first loads selectors from the provided path, then proceeds with the standard behavior of caching selectors from the current Foundry project.

## Open Questions
- **Interaction with default behavior:** Currently, the command creates a union of the "extra" ABIs and the local project ABIs. Should this flag be mutually exclusive (i.e., *only* use the folder when the flag is present), or remain additive?

## PR Checklist
- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/cli/src/utils/cmd.rs
```diff
@@ -385,3 +386,70 @@ pub fn cache_local_signatures(output: &ProjectCompileOutput) -> Result<()> {
     signatures.save(&path);
     Ok(())
 }
+
+pub fn cache_signatures_from_abis(folder_path: impl AsRef<Path>) -> Result<()> {
```
grandizzy, line 391: please add a comment for this new fn

DaniPopes (approved):

PR #12743, opened 2025-12-05
Feat(cast): implemented prestate fetch in cast run by debug_trace rpc 

### LINKED ISSUE #12719: feat(cast): use prestate tracer to speed up cast run
### Component

Cast

solution for #12711

currently we replay the entire block so that we get the proper state for a txs in a block

however, if the provider supports the prestate `debug_traceTransaction` then we get all the accessed state for that transaction which we can then commit into the db before executing the transaction.

this is similar to quick mode where we can bypass this entirely:

https://github.com/foundry-rs/foundry/blob/d3167db683c001c196126c912027437a9cb0bfb9/crates/cast/src/cmd/run.rs#L228-L229

however not everyone has access to the debug_ api or maybe it's not supported by the chain's client

so we could do this optimistically if the trace request succeeds

we like need some helpers for this:

https://github.com/foundry-rs/foundry/blob/d3167db683c001c196126c912027437a9cb0bfb9/crates/evm/evm/src/executors/mod.rs#L539-L539

that operates just on the account->storage map (pre-state)

and ideally this is either opt-in or opt-out 

### Additional context

_No response_

dipanshuhappy:
## Motivation

Resolves #12719 

## Solution

Solution is summarised in the issue

## PR Checklist
I can't add next because the rpc url ```reth-ethereum.ithaca.xyz/rpc``` does not implement the debug_traceTransaction method. Let me know if there is a consistent rpc I can use then I don't mind adding a test case.

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


dipanshuhappy:
It seems like there is a test that keeps failing. I can't reproduce it locally so I am un sure on how to fix it

emmajam:

Issue #10593, opened 2025-05-22
feat: multi-verify contracts by default and always submit to Sourcify


zerosnacks:
### Component

Other (please describe)

### Describe the feature you would like

Recently I wanted to verify contracts on multiple verification providers and we can improve the developer experience around this. I would argue that developers by default would want to verify everywhere at once.

We should make it very easy default to at least verify automatically on Etherscan + Verifier Alliance in one go.
Because of the https://verifieralliance.org/ both Blockscout and Sourcify (Argot Collective) can be verified at once by verifying on Sourcify. Ideally Etherscan would join the Verifier Alliance but this has not yet happened.

From personal experience - what currently happens is that if a user runs into any issues they eventually find a way to work around it and because they have resolved their own problem there is little incentive to report it as an issue.

Not many users write their scripts to be idempotent so users can easily end up deploying contracts multiple times.

Let's discuss how we can improve the developer experience around this.

### Common user experiences that raise the question "uh... what do I do now?"

A contract being deployed but the verification failing, users are tempted to just rerun. It is unclear what a user should do now given their contract is deployed. If this was ETH mainnet they paid quite a bit for it too.

```
==========================

##### sepolia
✅  [Success] Hash: 0xa7486709fbb1bc28da16126aceebf91a14aa4caf97fe4129938003869a8e9a75
Contract Address: 0xC8320C3CC70dc234298Bb84616f108d61a8215E0
Block: 8383084
Paid: 0.00000147893596998 ETH (151395 gas * 0.009768724 gwei)

✅ Sequence #1 on sepolia | Total Paid: 0.00000147893596998 ETH (151395 gas * avg 0.009768724 gwei)
                                                                                                                                                                

==========================

ONCHAIN EXECUTION COMPLETE & SUCCESSFUL.
##
Start verification for (1) contracts
Start verifying contract `0xC8320C3CC70dc234298Bb84616f108d61a8215E0` deployed on sepolia
EVM version: cancun
Compiler version: 0.8.30

Submitting verification for [src/Counter.sol:Counter] 0xC8320C3CC70dc234298Bb84616f108d61a8215E0.
Error: Encountered an error verifying this contract:
Response: `NOTOK`
Details:
                        `Missing/Invalid API Key`
```

On Etherscan - contract verification failing for whatever reason.

```
Start verification for (1) contracts
Start verifying contract `0x3163084A4C2Fd035BCC2950B0D8aFC51a4484653` deployed on sepolia
[...]

PR #12753, opened 2025-12-07
perf(script): run contract verifications in parallel

avorylli:
Replace sequential await loop with join_all to execute all contract verifications concurrently. 
This improves performance when verifyingmultiple contracts, especially for large deployments.

The change maintains the same error handling behavior while allowing all verification futures to run in parallel instead of waiting for each one to complete before starting the next.

### DIFF
--- a/crates/script/src/progress.rs
+++ b/crates/script/src/progress.rs
@@ -1,6 +1,7 @@
 use crate::receipts::{PendingReceiptError, TxStatus, check_tx_status, format_receipt};
 use alloy_chains::Chain;
 use alloy_primitives::{
+    Address,
     B256,
     map::{B256HashMap, HashMap},
 };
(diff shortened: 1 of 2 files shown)


grandizzy (commented):
thanks, this would be a great improvement but in order to be really useful we need verification to happen with progress showing status for each contract that is verified and not fail on first failure (see `forge test --show-progress`) as an example, would you be willing adding such? Thank you!

avorylli:
@grandizzy thanks for review, did changes that you requested, is it going in the right direction?

grandizzy:
> @grandizzy thanks for review, did changes that you requested, is it going in the right direction?

@avorylli yep, thank you, that's going in the right direction! Some things to fix
- let's display details near each contract that is verified (like chain, evm version, compiler, constructor args, verifier)
- let's display progress messages inline as the last entry - rn they are displayed inline and makes progress bar flaky

smth like instead
gg

to have
```
⡀ Mail (0xE51e8ae776431114628fD12E037044A2eE1ddec6)  
    - chain: 1111
    - EVM version: prague
    - Compiler version: 0.8.30
    - Constructor args: 00000000000000000000000020c00000000000000000000000000000000005aa
    - verifier: Sourcify
    - Submitted contract for verification:
	Verification Job ID: `56ead4aa-76c3-4b04-a1ff-9cf125d2d661`
	URL:                                                                                                                                                   
⡀ SimpleMail (0xB3Cf7142C78947281EdC6d634bB265B587300D81)
    - chain: 2222
    - EVM version: osaka
    - Compiler version: 0.8.18
    - verifier: Etherscan
    - Failed to deserialize response: data did not match any variant of untagged enum ResponseData res="{\"message\":\"unsupported media type application/x-www-form-urlencoded\",\"errors\":[{\"path\":\"/verify\",\"message\":\"unsupported media type application/x-www-form-urlencoded\"}]}"
```

grandizzy:
ref part of https://github.com/foundry-rs/foundry/issues/10593 CC @zerosnacks

avorylli:
@grandizzy wow, thanks for detailed explanation,

I hope my implementation includes what you asked for

grandizzy:
@avorylli oh, yes, I tested and def the right direction, there's still a little flakiness to figure it out, will get on review it asap

emmajam:

PR #12754, opened 2025-12-07
fix: remove redundant logs clone in fuzz counterexample

Lil-Duckling-22:
## Motivation

The fuzz executor was cloning logs from counterexample outcomes before moving them into `test_data.logs`, even though the logs are never accessed from the counterexample afterwards. This unnecessary clone adds overhead, especially when processing many failed fuzz cases.

## Solution

Replace `logs.clone()` with `std::mem::take()` to move logs instead of cloning them. This matches the pattern already used in successful fuzz cases and eliminates the redundant allocation.


DaniPopes:
not the same

PR #12756, opened 2025-12-07
fix: Clamp tarball gunzip size in npm installer to block gzip bombs

anim001k:
Add bounded gunzip helper for npm installer to prevent unbounded decompression. Apply max-size check when unpacking downloaded tarball before writing binary.


zerosnacks:
Considering we control what artifacts are downloaded this fix is not required

PR #12757, opened 2025-12-08
fix(evm): avoid duplicate cache checks

Forostovec:
Adjust ForkDbStateSnapshot::get_storage to read from state_snapshot.storage instead of the local cache. This removes a redundant cache lookup in storage_ref, ensures the snapshot is honored before falling back to the backend, and aligns storage precedence with basic_ref and block_hash_ref as documented.


zerosnacks:
This is wrong, state snapshots are only occupied when in use with the cheatcode; these are not equivalent

Issue #12773, opened 2025-12-09
forge: verify on plasma should retry if contract not found


sakulstra:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

9dc83c98223df59d102dde47405f0650c97b256a

### What command(s) is the bug in?

forge script

### Operating System

None

### Describe the bug

On latest nightly (i think is not related to nightly but fukasa gas price drops), when i try to submit a txn on optimism, txns error with:
`Sequence #1 on optimism | retrying transaction Err(server returned an error response: error code -32000: insufficient gas for floor data gas cost: gas 719122, minimum needed 741890)` and similar

sakulstra:
On gnosis chain, seems like foundry also can no longer submit txns.

```
## Setting up 1 EVM.

==========================

Chain 100

Estimated gas price: 0.000000101 gwei
```

Seems to dramatically underestimate gas cost: https://gnosisscan.io/gastracker

sakulstra:
Sorry will just dump here issues i face while i face them as i need to deploy sth on various chains 😅 

On Plasma, verification failed initially because I think etherscan is not fast enough.
```
Response: `NOTOK`
Details: `Error: contract does not exist 9745 0x<sth>
```

Manual verification works, but probably would be good if foundry retried when contract does not exist (and it knows it exists, given it just deployed).

grandizzy:
> Sorry will just dump here issues i face while i face them as i need to deploy sth on various chains 😅
> 
> On Plasama, verification failed initially because I think etherscan is not fast enough.
> 
> ```
> Response: `NOTOK`
> Details: `Error: contract does not exist 9745 0x<sth>
> ```
> 
> Manual verification works, but probably would be good if foundry retried when contract does not exist (and it knows it exists, given it just deployed).

@sakulstra for this you could tweak the params
```
      --retries <RETRIES>
          Number of attempts for retrying verification
          
          [default: 5]

      --delay <DELAY>
          Optional delay to apply in between verification attempts, in seconds
          
          [default: 5]
```

sakulstra:
@grandizzy Will try next time, but honestly feels like this should be the default.
At least with this error it's quite obvious that the explorer lags behind.


Also on polygon, without any custom settings it does:
```
Submitting verification for [src/UpgradePayload.sol:UpgradePayload] 0x<sth>.
[...]

PR #12782, opened 2025-12-10
chore: reset default retries

aganisgash:
Trying to figure out the gas estimated wrong issue, but no clue now.
Anyway fix this [chore](https://github.com/foundry-rs/foundry/issues/12773#issuecomment-3631941468) firstly.

### DIFF
--- a/crates/verify/src/retry.rs
+++ b/crates/verify/src/retry.rs
@@ -16,7 +16,7 @@ pub struct RetryArgs {
     #[arg(
         long,
         value_parser = RangedU64ValueParser::<u32>::new().range(1..),
-        default_value = "5",
+        default_value = "10",
     )]
     pub retries: u32,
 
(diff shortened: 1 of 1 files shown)


sakulstra:
Checked the pr, but i'm not sure if it fixes/ addresses the problem.
5 retries with some fallback delay, as it currently is i think should be fine.

The problem was more that one one chain it does fallback / retry, while the other does not.

Plasma: no retry, the following error.
```
Response: `NOTOK`
Details: `Error: contract does not exist 9745 0x<sth>
```

Polygon: retry, the following error.
```
Submitting verification for [src/UpgradePayload.sol:UpgradePayload] 0x<sth>.
Warning: Could not detect deployment: Unable to locate ContractCode at <sth>; waiting 5 seconds before trying again (4 tries remaining)
```

It seems similar, but even the error is slightly different, so i guess it might be in different parts of the process.

grandizzy:
Ah, I see, we should update here 

https://github.com/foundry-rs/foundry/blob/b4d888b20f874f9936687cbac507871e94aa0a1c/crates/verify/src/etherscan/mod.rs#L108

grandizzy:
@aganisgash thank you for the PR, made a fix and was able to verify in https://github.com/foundry-rs/foundry/pull/12790
Re this one, going to close for now, I think this will be greatly improved as UX with verifying contracts in parallel https://github.com/foundry-rs/foundry/pull/12753

aganisgash:
> @aganisgash thank you for the PR, made a fix and was able to verify in #12790 Re this one, going to close for now, I think this will be greatly improved as UX with verifying contracts in parallel #12753

Sure, thanks for the info

PR #12785, opened 2025-12-10
fix(forge): failing tests trigger early exit without --fail-fast

0xferrous:
## Motivation

When I encountered this I was not able to figure out why I was getting undeterministic output. Took a while before I figured it was a bug. It is unexpected to see undeterministic output without `--fail-fast`.

## Solution

Only `record_exit` when `--fail-fast` was provided.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/forge/src/runner.rs
+++ b/crates/forge/src/runner.rs
@@ -402,7 +402,7 @@ impl<'a> ContractRunner<'a> {
             let interrupt = early_exit.clone();
             self.tokio_handle.spawn(async move {
                 signal::ctrl_c().await.expect("Failed to listen for Ctrl+C");
-                interrupt.record_exit();
+                interrupt.record_ctrl_c();
             });
         }
 
@@ -442,9 +442,9 @@ impl<'a> ContractRunner<'a> {
                 );
                 res.duration = start.elapsed();
 
-                // Set fail fast flag if current test failed.
+                // Record test failure for early exit (only triggers if fail-fast is enabled).
                 if res.status.is_failure() {
-                    early_exit.record_exit();
+                    early_exit.record_failure();
                 }
 
                 Some((sig, res))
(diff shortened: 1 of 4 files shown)


crates/forge/src/runner.rs
```diff
@@ -442,8 +442,8 @@ impl<'a> ContractRunner<'a> {
                 );
                 res.duration = start.elapsed();
 
-                // Set fail fast flag if current test failed.
-                if res.status.is_failure() {
+                // Set fail fast flag if current test failed and fail-fast is enabled.
+                if res.status.is_failure() && self.tcfg.fail_fast {
                     early_exit.record_exit();
```
grandizzy: @0xferrous `early_exit.record_exit();` is a noop when fail fast not enabled (`inner` is None), not sure if this fix is correct, am I'm missing something?

https://github.com/foundry-rs/foundry/blob/b4d888b20f874f9936687cbac507871e94aa0a1c/crates/evm/evm/src/executors/mod.rs#L1143-L1157

0xferrous: It makes sense when you look here https://github.com/0xferrous/foundry/blob/push-pklwyzlpvzum/crates/forge/src/multi_runner.rs#L614. Since, it's an OR.

grandizzy: ah, ok, so you were getting the weird behavior when running `--show-progress` without `--fail-fast`

0xferrous: yes, it was `| show_progress` so that it can respect ctrl-c early exit, but the side effect was it also started exiting early even when fail_fast was not true

grandizzy: thanks, makes sense. can we group the early exit logic (for fail fast and show progress) in `EarlyExit` and have `record_ctrl_c` / `record_failure` instead single `record_exit` so this will be cleaner going forward?

0xferrous:
fwiw the way this is implemented, for a very slow test say `vm.sleep(1000000000);` it will not exit `--show-progress` even with ctrl-c because that early exit condition is only checked before starting a test, it does not cancel an already executing test.

grandizzy:
> fwiw the way this is implemented, for a very slow test say `vm.sleep(1000000000);` it will not exit `--show-progress` even with ctrl-c because that early exit condition is only checked before starting a test, it does not cancel an already executing test.

yep, indeed it doesn't cancel that

Issue #12784, opened 2025-12-10
refactor(anvil): use primitives to build typed tx from tx request


mablr:
This old helper looks similar to `TransactionBuilder` trait's `build_unsigned`, but it fills silently some fields while the latest returns an error with the list of missing fields.
https://github.com/foundry-rs/foundry/blob/b4d888b20f874f9936687cbac507871e94aa0a1c/crates/anvil/core/src/eth/transaction/mod.rs#L30

So the remplacement is not "drop-in" - but I think - it is very beneficial. For now it used in 3 places:
- `‎EthApi::build_typed_tx_request`, imo the whole method should be refactored, by prefilling request and then building typed tx.
https://github.com/foundry-rs/foundry/blob/b4d888b20f874f9936687cbac507871e94aa0a1c/crates/anvil/src/eth/api.rs#L3412

- `determine_base_gas_by_kind` helper, I think we don’t need to build typed tx here as we can simply rely on `TransactionBuilder` fns to determine type, kind, and access list len.
https://github.com/foundry-rs/foundry/blob/b4d888b20f874f9936687cbac507871e94aa0a1c/crates/anvil/src/eth/api.rs#L3621

- `‎Backend::simulate‎`, we can determine proactively tx type and fill missing fields using `TransactionBuilder` trait fns, or handle `build_unsigned` error, filling the fields we can then retry. btw this may be wrap into an helper providing similar service as original `transaction_request_to_typed`.
https://github.com/foundry-rs/foundry/blob/b4d888b20f874f9936687cbac507871e94aa0a1c/crates/anvil/src/eth/backend/mem/mod.rs#L1764

PR #12787, opened 2025-12-10
refactor(anvil): simplify  `determine_base_gas_by_kind` logic

mablr:
## Motivation

Towards #12784

Get rid of useless tx req to typed tx conversion.

## Solution

-  Just inspect tx kind and authorizations list len to determine base gas.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


onbjerg (approved):
much better, thank you!

Issue #12783, opened 2025-12-10
Add helper command for trace_rawTransaction using txpool-content response


mattsse:
### Component

Cast

### Describe the feature you would like

 cast tx-pool content returns a list of transactions

it would be nice if we had a command that accepts the json of a tx as input and calls trace_rawTransaction. for this we need to parse the json input and encode it.

## TODO
* add a new Trace command to https://github.com/foundry-rs/foundry/blob/b4d888b20f874f9936687cbac507871e94aa0a1c/crates/cast/src/opts.rs#L36-L36
* should accept any input and have an argument for --raw like `cast trace '{}' --raw`

figtracer:
🫡

dipanshuhappy:
Happy to work on this . I have been contributing to cast before also

figtracer:
can be closed

PR #12788, opened 2025-12-10
feat(cast): add trace_transaction and trace_rawTransaction

figtracer:
## Motivation
adds support for #12783 

## Solution
- adds `CastSubCommand::Trace` and `run_command` implementation

we can input the tx json with any trace type (https://reth.rs/jsonrpc/trace)
```bash
cast trace '{"type":"0x2","chainId":"0x1","nonce":"0x15","gas":"0x5a3c","maxFeePerGas":"0x18701a80","maxPriorityFeePerGas":"0x4b571c0","to":"0x8dde80a55d95647e419f0c8da19a53a8b62b5109","value":"0x10fc6d9c991b48","accessList":[],"input":"0x","r":"0x521e2aa4952078661311a8323d89ee9d350b6aa4b69f812e42b586c3d12d934d","s":"0x179efd11ff1dea53336c00f009949852c645482b11a50d23bc52c4f01043cabf","yParity":"0x0","v":"0x0","hash":"0x7332384d63de1a06e4b27e667933960a67b4a2469cba0aadc1c5fe93135de9bc","blockHash":"0x8650ef8deef20fc848582ab68f33c8020db73220dea0d0bb65168cb1904b8478","blockNumber":"0x16dea3c","transactionIndex":"0x10c","from":"0x9e7c924b09427b94be16d191719c4654c6f5ea6f","gasPrice":"0xe99a42a"}' --raw (--trace/--vm-trace/--state-diff)
```

with the raw trasaction
```bash
cast trace 0x02f87101158404b571c08418701a80825a3c948dde80a55d95647e419f0c8da19a53a8b62b51098710fc6d9c991b4880c080a0521e2aa4952078661311a8323d89ee9d350b6aa4b69f812e42b586c3d12d934da0179efd11ff1dea53336c00f009949852c645482b11a50d23bc52c4f01043cabf --raw (--trace/--vm-trace/--state-diff)
```

or using the tx hash
```bash
cast trace 0x7332384d63de1a06e4b27e667933960a67b4a2469cba0aadc1c5fe93135de9bc
```

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cast/src/cmd/trace.rs
+++ b/crates/cast/src/cmd/trace.rs
@@ -9,7 +9,6 @@ use foundry_cli::{
     utils::{self, LoadConfig},
 };
 use foundry_common::stdin;
-use op_alloy_consensus::OpTxEnvelope;
 
 /// CLI arguments for `cast trace`.
 #[derive(Debug, Parser)]
@@ -55,8 +54,7 @@ impl TraceArgs {
                 hex::decode(trimmed.strip_prefix("0x").unwrap_or(trimmed))?
             } else if is_json {
                 let tx: AnyRpcTransaction = serde_json::from_str(trimmed)?;
-                let envelope = tx.try_into_either::<OpTxEnvelope>()?;
-                envelope.encoded_2718().to_vec()
+                tx.inner.inner.encoded_2718().to_vec()
             } else {
                 hex::decode(trimmed)?
             };


crates/cast/src/opts.rs
```diff
@@ -1146,6 +1146,32 @@ pub enum CastSubcommand {
...
+        #[arg(long, requires = "raw")]
+        vm_trace: bool,
+
+        /// Include state changes caused by the transaction (requires --raw).
+        #[arg(long, requires = "raw")]
+        state_diff: bool,
+
+        #[command(flatten)]
+        rpc: RpcOpts,
+    },
```
mattsse: I'd prefer having a dedicating struct type for this

mattsse: I'd prefer having a dedicating struct type for this

mattsse: I'd prefer having a dedicating struct type for this

crates/cast/src/opts.rs
```diff
@@ -1146,6 +1146,32 @@ pub enum CastSubcommand {
         #[command(subcommand)]
         command: Erc20Subcommand,
     },
+    #[command(name = "trace")]
+    Trace {
```
mattsse: I believe trace_rawTransaction also supports a blockid argument, so we should add this here as well

crates/cast/src/args.rs
```diff
@@ -756,6 +758,60 @@ pub async fn run_command(args: CastArgs) -> Result<()> {
...
+                        .collect();
+
+                if trace_types.is_empty() {
+                    eyre::bail!(
+                        "No trace type specified. Use --trace, --vm-trace, or --state-diff"
+                    );
+                }
+
+                let params = serde_json::json!([out, trace_types]);
+                Cast::new(&provider).rpc("trace_rawTransaction", params).await?
```
mattsse: we can use the TraceApiExt for provider, migh require enabling the feature

crates/cast/src/opts.rs
figtracer: yeah, on it

crates/cast/src/opts.rs
figtracer: it doesn't i think
https://reth.rs/jsonrpc/trace#trace_rawtransaction
https://openethereum.github.io/JSONRPC-trace-module#trace_rawtransaction

crates/cast/src/args.rs
figtracer: oh nice, might be better to just create a new file `trace.rs` with everything in there

PR #12804, opened 2025-12-11
refactor(anvil): use `TransactionBuilder` trait to build tx requests

### LINKED ISSUE #12784: refactor(anvil): use primitives to build typed tx from tx request

mablr:
## Motivation

Close #12784

Towards `foundry-primitives` integration in Anvil, more contributions coming soon...

## Solution

- Removed the `transaction_request_to_typed` helper and integrated its logic into the `TransactionBuilder::build_tx_request` trait impl in `foundry_primitives`.
- Updated `FoundryTransactionRequest` to handle anvil's corner cases
- Improved error handling to bubble-up missing fields in transaction requests.
- Added `TxType` -> `FoundryTxType` conversion impl

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes (All existing tests pass, but if we want to fill `max_priority_fee_per_gas` with suggested min fee, it breaks existing `test_debug_trace_transaction_pre_state_tracer()`, would be relevant to update it, and investigate if any regression)

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -3396,7 +3396,7 @@ impl EthApi {
             FoundryTypedTx::Eip4844(TxEip4844Variant::TxEip4844(_))
                 if !self.backend.skip_blob_validation(from) =>
             {
-                // If blob validation is not skiped, reject TxEip4844 variant without sidecar.
+                // If blob validation is not skipped, reject TxEip4844 variant without sidecar.
                 Err(BlockchainError::FailedToDecodeTransaction)
             }
             res => Ok(res),


crates/anvil/core/src/eth/transaction/mod.rs
```diff
@@ -1,184 +1,25 @@
...
-use op_alloy_consensus::{OpDepositReceipt, OpDepositReceiptWithBloom, TxDeposit};
+use foundry_primitives::FoundryTxEnvelope;
+use op_alloy_consensus::{OpDepositReceipt, OpDepositReceiptWithBloom};
 use revm::interpreter::InstructionResult;
 use serde::{Deserialize, Serialize};
 use std::ops::Deref;
 
-/// Converts a [TransactionRequest] into a [FoundryTypedTx].
-/// Should be removed once the call builder abstraction for providers is in place.
-pub fn transaction_request_to_typed(
```
mattsse, line 30: rip bozo

crates/primitives/src/transaction/request.rs
```diff
@@ -348,77 +336,67 @@ impl TransactionBuilder<FoundryNetwork> for FoundryTransactionRequest {
...
-        let legacy = self.gas_price().is_some();
-        let eip2930 = legacy && self.access_list().is_some();
-        let eip1559 = self.max_fee_per_gas().is_some() && self.max_priority_fee_per_gas().is_some();
-        let eip4844 = eip1559 && self.as_ref().sidecar.is_some() && self.kind().is_some();
-        let eip7702 = eip1559 && self.as_ref().authorization_list.is_some();
-
-        let deposit = self.is_deposit() && self.from().is_some() && self.kind().is_some();
+        let deposit = self.get_other_field::<B256>("sourceHash").is_some()
+            && self.get_other_field::<u128>("mint").is_some()
+            && self.get_other_field::<bool>("isSystemTx").is_some();
```
mattsse, line 341: unsure if we have a helper for this in op-alloy, but if not worth adding a meta type for these fields imo

mablr: Oh yes that’s right, there is an helper type. I’ll change that.

crates/anvil/core/src/eth/transaction/mod.rs
mablr: Just the beginning 😎

grandizzy (approved):
nice stuff, ty!

crates/primitives/src/transaction/request.rs
grandizzy: @mablr do you want to add such before merge?

grandizzy: let's do that in a followup actually, going to merge for now, thanks!

mablr: okay, the helper is ready, this is waaay cleaner, opening a PR rn.

PR #12806, opened 2025-12-11
fix(verify): handle lack of trailing slash in verifier url

### LINKED ISSUE #12802: `forge verify-contract` doesn't work when `--verifier` has a port
### Component

Forge

- [x] Foundry
- [x] Foundryup

1.5.0

1.5.0

forge create src/Mail.sol:SimpleMail --private-key $PK --rpc-url $RPC_URL --broadcast --verify --verifier sourcify --verifier-url http://localhost:3000 --constructor-args $ADDRESS

### Operating System

macOS (Apple Silicon)

### Describe the bug

when `--verifier-url` / `VERIFIER_URL` has a port number, contract verification fails with:

```
Warning: builder error; invalid port number; waiting 10 seconds before trying again (9 tries remaining)
```

onbjerg:
## Motivation

Not providing a trailing slash to `--verifier-url` when using Sourcify would give you odd errors, see e.g. #12802

## Solution

As a simple solution we can just parse the URL using `Url::parse` and use the `Display` impl of `Url`. This is ok but I had to do a somewhat ugly workaround to keep `get_base_url` infallible. A follow up would be to simply rewrite the Sourcify verifier to use more sane types.

Closes #12802

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4446,6 +4446,7 @@ dependencies = [
  "tempfile",
  "tokio",
  "tracing",
+ "url",
  "yansi",
 ]
 
--- a/crates/verify/Cargo.toml
+++ b/crates/verify/Cargo.toml
@@ -40,6 +40,7 @@ semver.workspace = true
 regex = { workspace = true, default-features = false }
 yansi.workspace = true
 itertools.workspace = true
+url.workspace = true
 
 [dev-dependencies]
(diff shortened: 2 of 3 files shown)


o-az:
thanks chief

PR #12807, opened 2025-12-12
fix: include transaction index in simulation failure error messages

aganisgash:
When simulating multiple transactions, if any transaction fails, the error message only shows "Simulated execution failed." without indicating which transaction caused the failure. Makes debugging scripts with many transactions difficult.

Old:
```
Error: Simulated execution failed.
```

Now:
```
Error: Simulated execution failed for: Transaction #3, contract: MyToken, function: transfer
```

For multiple failures:
```
Error: Simulated execution failed for: Transaction #2, contract: MyToken, function: mint; Transaction #5, contract: Staking, function: stake
```

### DIFF
--- a/crates/script/src/simulate.rs
+++ b/crates/script/src/simulate.rs
@@ -115,6 +115,12 @@ impl PreSimulationState {
 
         let mut final_txs = VecDeque::new();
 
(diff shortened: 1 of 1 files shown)


emmajam:

PR #12808, opened 2025-12-12
feat(primitives): `get_deposit_tx_parts` helper

mablr:
## Motivation

#12804 follow-up

## Solution

- Introduce `get_deposit_tx_parts` helper, that converts `OtherFields` to `DepositTransactionParts`, on failure it returns the list of missing fields

Could be moved to alloy later, and/or wrapped into `TryFrom<OtherFields>` trait impl.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/primitives/src/transaction/request.rs
+++ b/crates/primitives/src/transaction/request.rs
@@ -7,6 +7,7 @@ use alloy_rpc_types::{AccessList, TransactionInputKind, TransactionRequest};
 use alloy_serde::{OtherFields, WithOtherFields};
 use derive_more::{AsMut, AsRef, From, Into};
 use op_alloy_consensus::{DEPOSIT_TX_TYPE_ID, TxDeposit};
+use op_revm::transaction::deposit::DepositTransactionParts;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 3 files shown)


crates/primitives/src/transaction/request.rs
```diff
@@ -406,3 +384,37 @@ impl TransactionBuilder<FoundryNetwork> for FoundryTransactionRequest {
...
     }
 }
+
+/// Converts `OtherFields` to `DepositTransactionParts`, produces error with missing fields
+pub fn get_deposit_tx_parts(
+    other: &OtherFields,
+) -> Result<DepositTransactionParts, Vec<&'static str>> {
+    let mut missing = Vec::new();
+    let source_hash =
+        other.get_deserialized::<B256>("sourceHash").transpose().ok().flatten().unwrap_or_else(
```
mattsse, line 394: @mablr do you want to add a helper to alloy that is `get_deserialized_flattened`?

which basically does this? won't be the most efficient but is convenient

mablr: @mattsse do you mean a simple helper returning `-> (Option<B256>, Option<u128>, Option<bool>)` to avoid using revm’s type ?

PR #12813, opened 2025-12-14
fix(anvil): propagate gas estimation errors 

### LINKED ISSUE #12812: `eth_fillTransaction` no longer reverts
### Component

Anvil

- [x] Foundry
- [x] Foundryup

anvil Version: 1.5.0-nightly Commit SHA: c21d2692ec1869e36baf1ef064768ddcdf7a39fe Build Timestamp: 2025-12-13T06:04:26.173327000Z (1765605866) Build Profile: maxperf

_No response_

_No response_

### Operating System

None

### Describe the bug

It seems as of https://github.com/foundry-rs/foundry/commit/52b4dd1aba4f3f398f2b0cb54d5e51647b2e93df, `eth_fillTransaction` no longer reverts when making a call to a contract that is expected to revert. Instead, it returns with a very high `gas`.

Steps to reproduce:

```bash
# 1. Set up Anvil, and fork mainnnet.
anvil --fork-url https://eth.merkle.io

# 2. Call `eth_fillTransaction` using calldata on a contract that is expected to revert.
cast rpc eth_fillTransaction '{"data":"0xa0712d6800000000000000000000000000000000000000000000000000000000000001a4","from":"0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266","to":"0xFBA3912Ca04dd458c843e2EE08967fC04f3579c2"}'

# 3. It returns data, instead of reverting
> {"raw":"0x02f84e01820ec28084402f6e77840393868e94fba3912ca04dd458c843e2ee08967fc04f3579c280a4a0712d6800000000000000000000000000000000000000000000000000000000000001a4c0808080","tx":{"type":"0x2","chainId":"0x1","nonce":"0xec2","gas":"0x393868e","maxFeePerGas":"0x402f6e77","maxPriorityFeePerGas":"0x0","to":"0xfba3912ca04dd458c843e2ee08967fc04f3579c2","value":"0x0","accessList":[],"input":"0xa0712d6800000000000000000000000000000000000000000000000000000000000001a4","r":"0x0","s":"0x0","yParity":"0x0","v":"0x0","hash":"0x060ca81919a6a1672b2b92d24a6f0b708d2fbcc1e55febfbc54b14b866023ebd","blockHash":null,"blockNumber":null,"transactionIndex":null,"from":"0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266","gasPrice":"0x402f6e77"}}

# It is expected to return: 
> Error: server returned an error response: error code 3: execution reverted: Token ID is taken, data: "0x08c379a000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000011546f6b656e2049442069732074616b656e000000000000000000000000000000"
```

jxom:
Fixes #12812

`eth_fillTransaction` no longer reverts when calling a contract that is expected to revert. Instead, it was returning a transaction with the block gas limit.

The fix propagates gas estimation errors instead of silently falling back to the block gas limit

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -1350,7 +1350,7 @@ impl EthApi {
     /// Handler for ETH RPC call: `eth_fillTransaction`
     pub async fn fill_transaction(
         &self,
-        request: WithOtherFields<TransactionRequest>,
+        mut request: WithOtherFields<TransactionRequest>,
     ) -> Result<FillTransaction<AnyRpcTransaction>> {
         node_info!("eth_fillTransaction");
 
@@ -1364,6 +1364,7 @@ impl EthApi {
         } else {
             self.request_nonce(&request, from).await?.0
         };
+        
         // Prefill gas limit with estimated gas, bubble up the error if the gas estimation fails
         // This is a workaround to avoid the error being swallowed by the `build_tx_request` function
         if request.as_ref().gas_limit().is_none() {


crates/anvil/src/eth/api.rs
```diff
@@ -3349,9 +3349,7 @@ impl EthApi {
         if request.gas_limit().is_none() {
             request.set_gas_limit(
                 self.do_estimate_gas(request.as_ref().clone(), None, EvmOverrides::default())
-                    .await
-                    .map(|v| v as u64)
-                    .unwrap_or(self.backend.gas_limit()),
+                    .await? as u64,
```
mablr:
```suggestion
                    .await
                    .map(|v| v as u64)
                    .unwrap_or(self.backend.gas_limit()),
```

crates/anvil/src/eth/api.rs
mablr, line 1376:
```suggestion
        // Prefill gas limit with estimated gas, bubble up the error if the gas estimation fails
        // This is a workaround to avoid the error being swallowed by the `build_tx_request` function
        if request.as_ref().gas_limit().is_none() {
            let estimated_gas = self
                .estimate_gas(request.clone(), None, EvmOverrides::default())
                .await?;
            request.as_mut().set_gas_limit(estimated_gas.to());
        }

```

While not perfectly clean, shadowing here resolves  the problem.

mablr (commented):
Yikes, good catch!

In previous PR I refactor/simplified the tx req building logic, that was duplicating/shadowing similar logic in 4-5 different places.

I proposed a fix that makes the CI successfull again.

PR #12814, opened 2025-12-14
fix: rename tx_id to sign_id in request_signing

reallesee:
The variable was incorrectly named tx_id after copying code from request_transaction. Since this function handles signing requests, not transactions, sign_id is the correct name.


onbjerg (changes requested):
i don't think this really matters here

PR #12815, opened 2025-12-15
docs: clarify WSL solc path example

priyanshusinghchouhan:
## Motivation

When using Foundry inside a WSL environment, solc is not always resolved correctly from `$PATH`.
This can lead to confusing failures or unexpected behavior when running forge commands.

The existing documentation mentions how to configure a local solc, but does not explicitly highlight that this configuration may be required for WSL users.

This PR aims to reduce environment-related friction for WSL users by making this requirement explicit in the docs.

Related discussion: https://github.com/Cyfrin/foundry-full-course-cu/discussions/4927

## Solution
Updated the documentation to clarify that WSL users may need to explicitly set the local solc binary path in `foundry.toml`.

Added an example configuration:
```
solc = "/home/<user>/.foundry/bin/solc"
```
This helps ensure Foundry consistently uses the intended solc binary when running inside WSL.

## PR Checklist

- [ ] Added Documentation

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -1350,7 +1350,7 @@ impl EthApi {
     /// Handler for ETH RPC call: `eth_fillTransaction`
     pub async fn fill_transaction(
         &self,
-        request: WithOtherFields<TransactionRequest>,
+        mut request: WithOtherFields<TransactionRequest>,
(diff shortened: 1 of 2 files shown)


priyanshusinghchouhan:
Actually, I was facing this issue on WSL enviroment but running `foundryup` again and again didn't help so I kinda found out this solution after struggling for days. Although it's not helpful but happy to contribute.
Thanks for the clarification. Appreciated

PR #12826, opened 2025-12-15
perf(chisel): reduce SessionSource cloning in parse_fragment

hawkadrian:
## What

Optimizes `parse_fragment` and `clone_with_new_line` methods to reduce expensive `SessionSource` cloning operations from up to 9 clones down to 2 in the worst case.

## Why

The original implementation cloned `SessionSource` three times within `parse_fragment` (once per scope attempt) and `clone_with_new_line` could call `parse_fragment` up to three times, resulting in excessive memory allocations. This was especially costly when `SessionSource` contains large code strings. The optimization reuses a single cloned instance and rolls back changes using `truncate()` between attempts, maintaining the same behavior while significantly reducing allocations.


emmajam:

PR #12838, opened 2025-12-16
fix(verify): avoid building standard json input for vyper

SashaMalysehko:
When verifying contracts via Etherscan using the standard-json provider, always constructed a StandardJsonCompilerInput and sanitized solc settings even when the target language was Vyper. For Vyper we then ignored this input and built a separate VyperInput, which adds unnecessary work and allocations on larger projects. This change detects the contract language first and only builds and normalizes the StandardJsonCompilerInput in the Solidity branch, leaving the Vyper path to construct only the Vyper input JSON as before.


emmajam:

PR #12845, opened 2025-12-16
fix: add null check for function name in is_constant_getter

oooLowNeoNooo:
## Problem

`is_constant_getter` used `header.name.unwrap()` without a `None` check, risking a panic. The comment didn't clarify that `check_screaming_snake_case` returns `None` for valid SCREAMING_SNAKE_CASE names.

## Solution

Replace `unwrap()` with `header.name.is_some_and()` for safe checking. Clarify the comment to note that `check_screaming_snake_case` returns `None` when the name is already correctly formatted.


onbjerg (changes requested):
`header.name` cannot be `None` here

PR #12846, opened 2025-12-16
perf: avoid unnecessary clone in HashSet duplicate check

PivasDesant:
Check for duplicates before cloning to avoid unnecessary allocations.


DaniPopes:
that is the opposite

PR #12847, opened 2025-12-16
perf(forge): cache contract dump directory path to avoid redundant allocations

marukai67:
Eliminates redundant path construction in the `dump_sources` function by caching `contract_dump_dir` once and reusing it.


mattsse (commented):
Not accepting low-value changes.

PR #12848, opened 2025-12-16
fix(forge): fail compilation on unresolved imports

### LINKED ISSUE #9272: forge compiles despite unresolved imports
### Component

Forge

- [X] Foundry
- [X] Foundryup

forge 0.2.0 (57bb12e 2024-11-06T00:26:30.417023000Z)

forge build

### Operating System

macOS (Apple Silicon)

### Describe the bug

1. Clone [this repo](https://github.com/morpho-org/morpho-blue-bundlers/tree/1361511cf34510b8b24a62f7838e7d8c2182e3ce)
2. `forge build`

Some imports will be noted as unresolved by forge, but compilation + tests will succeed.

May be related to #9271.

subwaycookiecrunch:
Fixes #9272

### Problem
`forge build` succeeds even when imports cannot be resolved via remappings, which can lead to silent failures and broken artifacts.

### Cause
Unresolved imports were not treated as fatal errors during compilation due to recursion limits in dependency resolution.

### Solution
Updated `foundry-compilers` to `0.19.10` which correctly handles import resolution.
Treats unresolved imports as hard failures and surfaces a clear diagnostic.

### Tests
Added regression tests to ensure compilation fails on unresolved imports.


crates/forge/tests/cli/issue_9272.rs
```diff
@@ -0,0 +1,20 @@
+
+
+forgetest!(test_unresolved_import_fails, |prj, cmd| {
```
onbjerg: please put this in the `build.rs` test file and name it/format it like so: https://github.com/foundry-rs/foundry/blob/e0636b834368f79fa9302775b92057d94789d72a/crates/forge/tests/cli/test_cmd/repros.rs#L69-L121

this helps us very easily dig up context if something regresses

onbjerg (changes requested):
thanks! a suggestion

subwaycookiecrunch:
Thanks! I've moved the regression test to repros.rs and updated the stderr_eq assertion to match the requested format.

PR #12851, opened 2025-12-17
feat: include actual tx type in unknown transaction error

Sharilleed223:
Improve error message by including the actual transaction type value when encountering an unknown transaction type.


Sharilleed223:
Hi @onbjerg , fixed, pls check, thx!

PR #12853, opened 2025-12-17
feat(cast):  `--replay-system-txes` / `--sys` arg to cast run system txes

grandizzy:
## Motivation
- adds `--replay-system-txes` / `--sys` arg to replay system txes with `cast run` (default false for backwards compatibility)

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cast/src/cmd/run.rs
+++ b/crates/cast/src/cmd/run.rs
@@ -245,7 +245,9 @@ impl RunArgs {
                 };
 
                 for (index, tx) in txs.iter().enumerate() {
-                    // Replay system transactions only if running with `all` option.
+                    // Replay system transactions only if running with `sys` option.
+                    // System transactions such as on L2s don't contain any pricing info so it
+                    // could cause reverts.
                     if !self.replay_system_txes
                         && (is_known_system_sender(tx.from())
                             || tx.transaction_type() == Some(SYSTEM_TRANSACTION_TYPE))


crates/cast/src/cmd/run.rs
```diff
@@ -61,6 +61,10 @@ pub struct RunArgs {
     #[arg(long)]
     quick: bool,
 
+    /// Whether to replay all transactions (including system transactions).
+    #[arg(long, short)]
+    all: bool,
```
onbjerg: i don't think the flag name is descriptive of what it does, can we rename it to something else? this name almost suggests it should be conflicting with `--quick` but i don't think that's necessarily true

onbjerg (changes requested):
can you also add a pr description 😊

crates/cast/src/cmd/run.rs
grandizzy: agree, would `--replay-system-txes` work?

onbjerg: yeah and maybe just `--sys` for short as an alias

grandizzy: added in https://github.com/foundry-rs/foundry/pull/12853/changes/6607d51d7f8fb864bd01c6b5f4c327b63ab397b0

PR #12862, opened 2025-12-17
fix(preprocessor): use regex for precise memory/calldata removal

PivasDesant:
Use word-boundary regex instead of simple string replace to avoid breaking type names that contain "memory" or "calldata" as substrings.


DaniPopes:
no

PR #12864, opened 2025-12-17
fix(debugger): handle missing memory in buffer rendering

forkfury:
## Fix

Replaces unsafe `unwrap()` calls on `step.memory` with safe `map_or()` handling to prevent runtime panics when memory data is unavailable.

## Changes

- `draw.rs`: Use `map_or()` to return empty slice when memory is `None` in `draw_buffer()`
- `context.rs`: Use `map_or()` to return empty slice when memory is `None` in `active_buffer()`

Both functions now gracefully handle missing memory data instead of panicking, allowing the debugger to continue functioning when switching to Memory buffer on steps without memory snapshots.


DaniPopes:
Cannot be None

Issue #12508, opened 2025-11-09
feat(`forge fmt`): should give an option to place each if condition in a separate line


GalloDaSballo:
### Component

Forge

### Describe the feature you would like

I'm researching the literature on coverage being a predictor of bugs.

While coverage is not the full story, there's strong correlation between having high coverage and finding more bugs (although the correlation is moderate when adjusting for size).

Either way, we can all agree that professionally we rely on coverage as a indicator of attention and effort put in the testing suite.

Branches, (called Decisions in Condition / Decision Coverage), are a key part of coverage as each branch implies some conditional action.

Branches choice are determined by Conditions (the boolean variables evaluated in the branch).

Forge fmt, currently inlines all conditions in a branch instead of separating them.

`forge coverage summary` correctly counts statements covered and will address conditions that haven't been covered.

However, generating the `lcov` report will ignore statement coverage and will lead most developers to miss this nuance.

## Ask: Allow option to put each condition in a separate line instead of defaulting to inlining conditions

```toml
[fmt]
format_conditions = "inline" // "inline" | "multi"
```

I would also suggest defaulting to multi, however if this could lead to changes for other projects I wouldn't push for this.

## POC

Toy example of conditions that can never be hit, which are ignored by the coverage metric since the if condition is in one line.

### Code with realistic 70% coverage

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.13;

contract Counter {
    uint256 public number = 1;

    function setNumber(uint256 newNumber) public {
        if (
            newNumber % 2 == 0 || 
            newNumber % 2 == 1 || 
            newNumber != 0 || 
            newNumber != 1 || 
            newNumber != 2
        ) {
            number = newNumber;
        }
    }

    function increment() public {
        number++;
    }
}
```

### Code with deceptive 100% coverage

Running `forge fmt` leads to this, which causes lcov to report 100% coverage.

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.13;

contract Counter {
    uint256 public number = 1;

    function setNumber(uint256 newNumber) public {
        if (newNumber % 2 == 0 || newNumber % 2 == 1 || newNumber != 0 || newNumber != 1 || newNumber != 2) {
            number = newNumber;
        }
    }

    function increment() public {
        number++;
    }
}
```

### Lcov for the deceptive snippet

[...]

PR #12866, opened 2025-12-18
Fix/lcov condition coverage 12657

### LINKED ISSUE #12657: feature(forge): improve lcov condition coverage
### Component

Forge

Related: #12508 (formatter option for multi-line conditions)

Unlike the original proposed solution in the issue above, this needs to be solved not only through formatting, but by fixing the lcov analysis mechanism itself. This will solve the problem completely, rather than only through correct formatting (which can implicitly cause this problem).

Since I developed the solution for the previous issue and am familiar with the details of the current problem, I could solve this one as well.

### Additional context

_No response_

Syzygy106:
## Motivation

Closes #12657 

LCOV branch coverage produced by `forge coverage` was misleading and formatting-dependent for compound boolean conditions. In particular, short-circuit chains (`&&` / `||`) could appear as 100% branch coverage when written inline, even if most sub-conditions were never evaluated. Multi-line `if (...)` conditions could also yield confusing `BRDA` attribution (branches reported on the body line instead of the condition line). This is tracked in #12657 (rooted in the POC from #12508).

## Solution

This PR improves how `forge coverage` generates LCOV branch data so that short-circuit boolean conditions are reflected in branch coverage and the result is independent of formatting (inline vs multi-line). Previously, LCOV would typically only see the top-level construct (`if` true/false, `require` pass/fail), which could report deceptively high branch coverage even when later operands in `&&`/`||` chains were never evaluated due to short-circuiting.
To address this, we extend the coverage source analysis to treat each `&&` / `||` chain as a sequence of short-circuit decision points and emit corresponding LCOV branch items. This makes LCOV’s `BRF`/`BRH` counts represent whether each boolean operand was actually evaluated across execution paths, rather than only whether the outer statement was taken.
The same condition-branch modeling is applied consistently across `if` conditions, `require(...)` conditions, and loop conditions (`while`, `for`, and `do-while`), so condition coverage behaves the same regardless of where the boolean expression appears.
Finally, the PR fixes LCOV BRDA line attribution for multi-line `if (...)` statements by ensuring branches are reported on the condition line (while still anchoring to the correct bytecode range for execution mapping). This avoids confusing reports where branch entries appear on the body line purely due to formatting and keeps line attribution stable.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/coverage/src/analysis.rs
+++ b/crates/evm/coverage/src/analysis.rs
@@ -122,8 +122,26 @@ impl<'gcx> SourceVisitor<'gcx> {
         self.all_lines.dedup();
         let mut lines = Vec::new();
         for &line in &self.all_lines {
-            if let Some(reference_item) =
-                self.items.iter().find(|item| item.loc.lines.start == line)
+            if let Some(reference_item) = self
+                .items
+                .iter()
+                .filter(|item| item.loc.lines.start == line)
+                // Prefer statement/function anchors for line coverage; branch items can be
+                // synthetic (e.g. short-circuit condition branches) and should not skew line
+                // coverage anchoring.
+                //
+                // Within the same kind priority, pick the smallest byte-start so the line item
+                // anchors to the earliest code on that source line (typically the condition /
+                // entrypoint), and doesn't accidentally anchor to a later branch body.
+                .min_by_key(|item| {
+                    let prio = match item.kind {
+                        CoverageItemKind::Statement => 0u8,
+                        CoverageItemKind::Function { .. } => 1u8,
+                        CoverageItemKind::Branch { .. } => 2u8,
+                        CoverageItemKind::Line => 3u8,
+                    };
+                    (prio, item.loc.bytes.start)
+                })
             {
                 lines.push(CoverageItem {
                     kind: CoverageItemKind::Line,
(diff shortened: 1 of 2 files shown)


grandizzy (commented):
thank you! However it doesn't seem to fix the reported issue, for example, given the contract
```Solidity
pragma solidity ^0.8.13;

contract Counter {
    uint256 public number = 1;

        if (newNumber != 0 || newNumber != 1 || newNumber != 2 || newNumber != 3 || newNumber != 4) {
            number = newNumber;
        }
        number = newNumber;
    }

    function increment() public {
        number++;
    }
}
```

`forge coverage` reports `22.22% (2/9)` but there are only 5 branches and not 9. Also the lcov report shows duped branches (pairs of (branch 0 / branch 1) instead single (branch 0 / branch 1 / branch 2 / branch 3)) which is confusing

Image

Syzygy106:
> thank you! However it doesn't seem to fix the reported issue, for example, given the contract
> 
> ```solidity
> // SPDX-License-Identifier: UNLICENSED
> pragma solidity ^0.8.13;
> 
> contract Counter {
>     uint256 public number = 1;
> 
>     function setNumber(uint256 newNumber) public {
>         if (newNumber != 0 || newNumber != 1 || newNumber != 2 || newNumber != 3 || newNumber != 4) {
>             number = newNumber;
>         }
>         number = newNumber;
>     }
> 
>     function increment() public {
>         number++;
>     }
> }
> ```
> 
> `forge coverage` reports `22.22% (2/9)` but there are only 5 branches and not 9. Also the lcov report shows duped branches (pairs of (branch 0 / branch 1) instead single (branch 0 / branch 1 / branch 2 / branch 3)) which is confusing
> 
> Image

But LCOV BRF/BRH counts branch paths, not branch points. For an `||` chain with 5 operands, the short-circuit modeling yields 4 decision points, each with 2 outcomes -> 8 branch paths. In Foundry’s LCOV branch model (if I’ve got it right), an `if` without an `else` contributes one reported branch path (the “then” path only), so the total is BRF = 8 + 1 = 9. The “duplicated” (0/1) branches are an artifact of genhtml displaying the branch index without the block component; each 0/1 pair belongs to a different block (decision point)

emmajam:

Syzygy106:
@grandizzy , here I'm not sure. If you feel that bad -> could be dropped. But could you recheck?

PR #12867, opened 2025-12-18
chore: fix overflow in cast opts tests

grandizzy:
## Motivation
- cast clap based tests overflows during argument parsing, can be repro as `cargo test -p cast opt`
- seen in CI failures on foundry https://github.com/foundry-rs/foundry/actions/runs/20090120594/job/57635699566#step:14:2037 and tempo https://github.com/tempoxyz/tempo-foundry/actions/runs/19987440359/job/57330249157#step:16:1321
- (even with boxed enums the cast command was too large for the default test thread stack) 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cast/src/opts.rs
+++ b/crates/cast/src/opts.rs
@@ -1174,92 +1174,110 @@ mod tests {
 
     #[test]
     fn verify_cli() {
-        Cast::command().debug_assert();
+        with_test_stack(|| {
+            Cast::command().debug_assert();
+        });
     }
 
     #[test]
     fn parse_proof_slot() {
-        let args: Cast = Cast::parse_from([
-            "foundry-cli",
-            "proof",
-            "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
-            "0",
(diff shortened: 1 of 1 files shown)


onbjerg (commented):
this might be a regression in clap https://github.com/clap-rs/clap/issues/5134#issuecomment-2929766787

let's try to roll back instead?

grandizzy:
> this might be a regression in clap [clap-rs/clap#5134 (comment)](https://github.com/clap-rs/clap/issues/5134#issuecomment-2929766787)
> 
> let's try to roll back instead?

made some checks and it's actually failing with nightly rust, with stable they pass ...

onbjerg:
can we pin nightly then?

grandizzy:
> can we pin nightly then?

yeah, can try figure out which one starts to fail, CC @DaniPopes any objection against pining nightly? thanks!

DaniPopes:
you can set RUST_MIN_STACK

PR #12868, opened 2025-12-18
fix(evm): seed cheatcodes RNG from fuzzer to ensure randomness in fuzz tests

### LINKED ISSUE #12817: Lack of randomness in `vm.randomUint()`
### Component

Forge

- [x] Foundry
- [x] Foundryup

foundryup: 1.0.1

forge test

### Operating System

macOS (Apple Silicon)

### Describe the bug

I am observing lack of randomness when using `vm.randomUint()` and `vm.randomUint(a, b)`, compared to when using fuzzed input. In the example below, only `test_fuzz_fuzzedInput` fails.

Execution command: `forge t --mc RandomTest --fuzz-runs 1000000`
```
pragma solidity ^0.8.28;

import {Test} from 'forge-std/Test.sol';

contract RandomTest is Test {
  function test_fuzz_randomUintRange(uint256) public {
    uint256 randomUint = vm.randomUint(0, 3);
    if (randomUint != 1) {
      revert();
    }
  }

  function test_fuzz_randomUint(uint256) public {
    uint256 randomUint = vm.randomUint() % 4;
    if (randomUint != 1) {
      revert();
    }
  }

  function test_fuzz_fuzzedInput(uint256 randomUint) public {
    randomUint = bound(randomUint, 0, 3);
    if (randomUint != 1) {
      revert();
    }
  }
}
```

I think all tests should be failing given that the used range for `randomUint` is small, and fuzz runs are high.

subwaycookiecrunch:
## Description
Fixes #12817

This PR addresses the lack of randomness in `vm.randomUint()` when running fuzz tests. Previously, `vm.randomUint()` relied on a potentially constant or decoupled RNG state that was not updated across fuzz runs, leading to identical values in every run even when the fuzzer seed changed.

## Changes
- Modified the fuzz executor to inject a new seed into the `Cheatcodes` environment before every individual fuzz run.
- The seed is derived from the fuzzer's main `TestRunner` RNG, ensuring that the randomness is deterministic with respect to the fuzz seed but varies across runs.
- Added `rand` dependency to the EVM crate.

## Verification
- Created a reproduction test case that logs `vm.randomUint()` output.
- Verified that values now vary across fuzz runs (e.g., `0, 1, 2...`) instead of remaining constant.


.github/workflows/ci.yml
```diff
@@ -107,6 +107,8 @@ jobs:
           toolchain: nightly
           components: rustfmt
       - run: cargo fmt --all --check
+        env:
+          RUSTC_WRAPPER: ""
```
onbjerg: hmm, unrelated? 🤔

.github/workflows/ci.yml
```diff
@@ -126,6 +128,8 @@ jobs:
       - name: forge fmt
         shell: bash
         run: ./.github/scripts/format.sh --check
+        env:
+          RUSTC_WRAPPER: ""
```
onbjerg: same

.github/workflows/ci.yml
subwaycookiecrunch: Thanks for the review!

You’re right , this part looks unrelated.  
I’m still digging into the CI/YAML side and may need some time to clean this up properly.
I’ll update the PR once I’ve isolated the correct fix.

DaniPopes:
thanks, this is the right fix, but I was already working on this here #12843, i will add you as a co-author

subwaycookiecrunch:
Thanks for the clarification, and appreciate you adding me as a co-author.

Issue #12872, opened 2025-12-19
chore(`forge doc`): update mdbook to 0.5.2


zerosnacks:
### Component

Forge

### Describe the feature you would like

We are currently still using version 0.4, there have been a number of breaking changes that need to be sorted out. Particularly `mdbook` crate is now a `bin`, not a `lib` and therefore fails to build.

See: https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#05-migration-guide

Dhananjay0719:
Hi! I’d like to work on this issue if it’s still available.
Please let me know if there’s any guidance or preferred approach.

PR #12874, opened 2025-12-19
chore: update dependencies

zerosnacks:
Helps with syncing downstream forks

Updates all dependencies, minor breaking change in `toml_edit`

Opened ticket for update `mdbook`: https://github.com/foundry-rs/foundry/issues/12872 as a good first issue as it is a larger breaking change

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "1.5.0"
+version = "1.5.1"
 edition = "2024"
 rust-version = "1.89"
 authors = ["Foundry Contributors"]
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -348,7 +348,7 @@ mesc = "0.3"
 memchr = "2.7"
 num-format = "0.4"
 parking_lot = "0.12"
-proptest = "1.8.0"
+proptest = "1.9.0"
```
zerosnacks, line 351: note, we may want to leave this bump out of this PR

was released 2 months ago https://crates.io/crates/proptest/1.9.0 but may require special handling / review but tests are passing

onbjerg (approved):

Cargo.toml
zerosnacks: cc @grandizzy for visibility

PR #12876, opened 2025-12-19
feat(vm): support compilation profile selection in getCode

subwaycookiecrunch:
## Motivation
Users often define multiple compilation profiles in the configuration file to test contracts against different settings (e.g., different EVM versions, optimization levels). While `vm.getCode` currently supports selecting artifacts by Solidity version, it lacks the ability to explicitly select an artifact by its profile name. This feature enables users to retrieve the bytecode for a specific profile using the syntax `vm.getCode("Contract:Profile")`.

## Solution
I updated the cheatcode logic to handle profile names in the artifact path string.
- The logic now attempts to parse the suffix (after the `:`) as a Semantic Version.
- If version parsing fails, it treats the suffix as a profile name.
- The artifact filtering logic was updated to check against the profile when a profile name is identified.
I also fixed a small regression in an existing test file where `vm.seed` was being used incorrectly; it has been replaced with `vm.setSeed`.

## PR Checklist
- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


emmajam:

PR #12879, opened 2025-12-19
test: actually cache default build artifacts

DaniPopes:
`forge build` on an empty project doesn't build anything, not even the libraries. So clear after building. Also run `forge test` since for various reasons it might build different files, messing with the cache of `forge build`.


grandizzy (dismissed):

DaniPopes:
well nevermind i dont want to bother with this https://github.com/foundry-rs/foundry/actions/runs/20374981337/job/58551012726?pr=12879#step:14:6100

PR #12905, opened 2025-12-21
feat(traces): make TraceIdentifier async to improve performance

### LINKED ISSUE #12714: perf: printing traces is too slow
### Component

Forge

- [x] Foundry
- [x] Foundryup

1.5.0

i use foundry.nix

forge test -vvv

### Operating System

Linux

### Describe the bug

<summary>
<details>

```

              // ... processing ...
          })
  );

  The Issue:

  Performance Impact:

  The files affected:

```
</details>
</summary>

subwaycookiecrunch:
Fixes #12714.

This PR addresses the performance degradation observed when running tests with verbose tracing.

The issue was caused by synchronous blocking calls during the trace identification process. Specifically, fetching contract metadata from external sources was blocking the execution runtime, leading to significant delays when processing multiple addresses.

## Changes
- Async Identification: Updated the trace identification logic to be asynchronous, allowing for non-blocking operations.
- Concurrent Lookups: Refactored the external metadata fetching to run concurrently without blocking the main thread.
- System Updates: Updated all relevant components to support the new asynchronous identification flow.

## Performance
These changes ensure that network requests for identifying contracts are handled efficiently, resulting in faster execution times for operations that require external trace identification.


zerosnacks:
Hi @subwaycookiecrunch, thanks for your PR!

Your PR currently contains a lot of unrelated changes currently, would you mind doing a second pass and avoid any changes that are not strictly related to your proposed fix for https://github.com/foundry-rs/foundry/issues/12714

I am not sure if your current proposed solution addresses the core issue; please clarify - preferably through a before-after timing table.

emmajam:

PR #12926, opened 2025-12-24
chore: respect configured thread count in create2

tefyosL-sol:
Use `rayon::current_num_threads()` instead of `std::thread::available_parallelism()`, respecting user-configured thread count.


grandizzy:
this is not right, rayon is not yet configured

PR #12928, opened 2025-12-24
perf(verify): clone only VyperSettings instead of entire context

CreeptoGengar:
Replace context.clone().compiler_settings.vyper with  context.compiler_settings.vyper.clone() to avoid unnecessary  cloning of the entire VerificationContext. This reduces memory  allocations when creating VyperInput instances.
Applied to both etherscan and sourcify verification providers.


grandizzy:
thank you, should be fine

PR #12929, opened 2025-12-24
refactor: extract sequence save-and-reload pattern into helper method

forkfury:
### Fix
Eliminates code duplication in the broadcast sequence save-and-reload pattern.
### Changes
- Added save_and_reload_sequence() helper method
- Replaced 3 duplicate instances with calls to the helper
- Replaced unwrap() with ok_or_else() for better error handling


grandizzy:
thanks, it's fine as it is

PR #8644, opened 2024-08-10
feat(invariant): support fuzz with random msg.value

QiuhaoLi:
## Motivation

Currently, the foundry's invariant fuzz testing doesn't support txs with value > 0, which makes it unable to find sequences in some situations. For example, the Pay contract below will only set the `hacked` variable if the calls are `C() --> B() --> A()` with `2, msg.value>0.11 ether, 0`, but foundry can't generate such calls.

```solidity
contract Pay {
    uint256 private counter;
    bool public hacked; // CBA with 2,msg.value>0.11,0

    function A(uint8 x) external {
        if (counter == 2 && x == 0) hacked = true; else counter = 0;
    }
    function B() external payable {
        if (counter == 1 && msg.value > 0.11 ether) counter++; else counter = 0;
    }
    function C(uint8 x) external {
        if (counter == 0 && x == 2) counter++;
    }
}

contract PayTest is Test {
    Pay public pay;

    function setUp() public {
        pay = new Pay();
    }
    /// forge-config: default.invariant.runs = 10000
    function invariant_Pay() view external {
        assertEq(pay.hacked(), false);
    }
}
```

## Solution

When generating a tx, we set the `msg.value` as a random number (uint96) if the target function is payable. After applying this strategy, foundry can find the sequence quickly:

```bash
qiuhao@pc:~/tmp$ ./forge test
[⠊] Compiling...
[⠑] Compiling 1 files with Solc 0.8.26
[⠘] Solc 0.8.26 finished in 564.77ms
Compiler run successful!

Ran 1 test for test/Counter.t.sol:PayTest
[FAIL. Reason: invariant_Pay replay failure]
        [Sequence]
                sender=0x0000000000000000000000000000000000000190 addr=[test/Counter.t.sol:Pay]0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f calldata=C(uint8) args=[2]
                sender=0x0000000000000000000000000000000000000851 addr=[test/Counter.t.sol:Pay]0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f calldata=B() args=[] value=[79228162514264337593543950333]
                sender=0x0000000000000000000000000000000000000103 addr=[test/Counter.t.sol:Pay]0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f calldata=A(uint8) args=[0]
 invariant_Pay() (runs: 1, calls: 1, reverts: 1)
[...]

PR #12934, opened 2025-12-26
feat(corpus): add sender and value mutation in corpus fuzzing

mattvaux:
Fixes TODO: mutate sender and value in corpus fuzzing.

Only calldata was being mutated, so we missed bugs in contracts that depend on msg.sender or msg.value. Added mutation for both fields with 15% probability each.

Added value field to CallDetails with serde default for backward compatibility.

### DIFF
--- a/crates/evm/evm/src/executors/corpus.rs
+++ b/crates/evm/evm/src/executors/corpus.rs
@@ -1,7 +1,7 @@
 use crate::executors::{Executor, RawCallResult, invariant::execute_tx};
 use alloy_dyn_abi::JsonAbiExt;
 use alloy_json_abi::Function;
-use alloy_primitives::Bytes;
+use alloy_primitives::{Address, Bytes, U256};
 use eyre::eyre;
 use foundry_config::FuzzCorpusConfig;
 use foundry_evm_fuzz::{
@@ -451,8 +451,6 @@ impl CorpusManager {
                     let idx = rng.random_range(0..new_seq.len());
                     let tx = new_seq.get_mut(idx).unwrap();
                     if let (_, Some(function)) = targets.fuzzed_artifacts(tx) {
-                        // TODO add call_value to call details and mutate it as well as sender some
-                        // of the time
                         if !function.inputs.is_empty() {
                             self.abi_mutate(tx, function, test_runner, fuzz_state)?;
                         }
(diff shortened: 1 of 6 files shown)


crates/evm/evm/src/executors/corpus.rs
```diff
@@ -628,6 +626,25 @@ impl CorpusManager {
...
+            // Generate random value, biased towards smaller values for practical fuzzing
+            let value = match test_runner.rng().random_range(0..=10) {
+                0..=5 => U256::from(test_runner.rng().random_range(0..=1000)), // Small values
+                6..=8 => U256::from(test_runner.rng().random_range(0..=1_000_000_000_000_000u64)), // Medium values (up to 0.001 ETH)
+                9 => U256::from(test_runner.rng().random_range(0..=1_000_000_000_000_000_000u64)), // Large values (up to 1 ETH)
+                10 => U256::MAX, // Edge case
+                _ => unreachable!(),
+            };
+            tx.call_details.value = Some(value);
+        }
```
0xalpharush, line 646: This should use the existing U256 [mutator](https://github.com/foundry-rs/foundry/blob/fdf5732d08ce1c67aa0aaf047c3fb86614caf5ae/crates/evm/fuzz/src/strategies/param.rs#L284) or something like [this one](https://github.com/foundry-rs/foundry/pull/8644/) that considers whether func is payable and be added as a mutation [here](https://github.com/foundry-rs/foundry/blob/fdf5732d08ce1c67aa0aaf047c3fb86614caf5ae/crates/evm/evm/src/executors/corpus.rs#L376)



I believe you will also need to add the balance to the sender in the DB as in https://github.com/foundry-rs/foundry/pull/8644/

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -628,6 +626,25 @@ impl CorpusManager {
 
         tx.call_details.calldata =
             function.abi_encode_input(&prev_inputs).map_err(|e| eyre!(e.to_string()))?.into();
+
+        // Mutate sender with 15% probability
```
0xalpharush, line 630: This cannot be purely random and needs to use this [function](https://github.com/foundry-rs/foundry/blob/fdf5732d08ce1c67aa0aaf047c3fb86614caf5ae/crates/evm/fuzz/src/strategies/invariants.rs#L81). It should also be added as a mutation [here](https://github.com/foundry-rs/foundry/blob/fdf5732d08ce1c67aa0aaf047c3fb86614caf5ae/crates/evm/evm/src/executors/corpus.rs#L376)

emmajam:

PR #12942, opened 2025-12-28
feat: faster `cast run` when ran against anvil

0xferrous:
## Motivation

Much faster tracing response for `cast run` when running against remote anvil instances because it saves 100s of sequential network requests querying the contract states.

## Solution

Introduce a new rpc method in anvil `anvil_revmTrace` that returns `CallTraceArena` as response which then cast can print directly instead of doing local simulation which is very slow over network because it doesn't know which state it will need and the state gets queried sequentially as revm steps through the execution.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/core/src/eth/mod.rs
+++ b/crates/anvil/core/src/eth/mod.rs
@@ -722,6 +722,10 @@ pub enum EthRequest {
     /// Set the executor (sponsor) wallet
     #[serde(rename = "anvil_setExecutor", with = "sequence")]
     AnvilSetExecutor(String),
+
+    /// Returns a decoded CallTraceArena for a transaction
+    #[serde(rename = "anvil_revmTrace", with = "sequence")]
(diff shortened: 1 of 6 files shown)


onbjerg (changes requested):
I don't think we want this because it adds complexity in both Anvil and Cast, and to me the upsides are fairly minimal:

All we are doing here is offloading the sequential loading of remote storage from cast to Anvil. If you are not using anvil in forked mode, then most likely you are running Anvil locally in which case this will be fast anyway.

If the primary idea for this is that Anvil caches remote storage, then the gains from this are already possible today as well, as you can just run `cast run` against a locally running Anvil instance.

0xferrous:
This is mainly useful for remote anvils where the network back and forth is much more costly(slow).

PR #12944, opened 2025-12-28
fix: use network-specific BaseFeeParams for Optimism in Anvil

### LINKED ISSUE #12943: Anvil uses Ethereum base fee params for Optimism instead of Canyon params
### Component

Anvil

- [x] Foundry
- [x] Foundryup

foundryup: 1.5.0

anvil

### Operating System

macOS (Apple Silicon)

### Describe the bug

# Problem

Anvil currently hardcodes `BaseFeeParams::ethereum()` for all EIP-1559 base fee calculations, even when running in Optimism mode (`--optimism` flag). This causes incorrect base fee adjustments for Optimism networks.

# Expected Behavior

When running with `--optimism`, Anvil should use `BaseFeeParams::optimism_canyon()` which has different parameters.

# Current Behavior

Anvil always uses Ethereum's params (12.5% max change) regardless of the `--optimism` flag, resulting in base fee adjustments that don't match the actual Optimism network behavior.

haythemsellami:
## Motivation

fix: use network-specific BaseFeeParams for Optimism in Anvil.

Closes #12943.

## Solution

This PR fixes Anvil to use BaseFeeParams::optimism_canyon() when running in Optimism mode instead of always using Ethereum's parameters.

Changes:
  - crates/anvil/src/eth/fees.rs
    - Added base_fee_params field to FeeManager
    - Updated new() to accept network-specific BaseFeeParams
    - Added base_fee_params() getter
    - get_next_block_base_fee_per_gas() now uses stored params
    - Removed unused default_elasticity() function
    - Added doc comment to calculate_next_block_base_fee() clarifying it uses Ethereum params
  - crates/anvil/src/config.rs
    - Select BaseFeeParams::optimism_canyon() when is_optimism() is true
  - crates/anvil/src/eth/backend/mem/mod.rs
    - Use self.fees.base_fee_params() in block simulation instead of hardcoded Ethereum params
  - crates/anvil/tests/it/optimism.rs
    - Added test_optimism_base_fee_params() to verify correct params are used

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/config.rs
+++ b/crates/anvil/src/config.rs
@@ -16,7 +16,7 @@ use crate::{
 };
 use alloy_chains::Chain;
 use alloy_consensus::BlockHeader;
-use alloy_eips::eip7840::BlobParams;
+use alloy_eips::{eip1559::BaseFeeParams, eip7840::BlobParams};
 use alloy_evm::EvmEnv;
 use alloy_genesis::Genesis;
 use alloy_network::{AnyNetwork, TransactionResponse};
@@ -1094,6 +1094,9 @@ impl NodeConfig {
             self.networks,
         );
 
+        let base_fee_params: BaseFeeParams =
+            self.networks.base_fee_params(self.get_genesis_timestamp());
+
         let fees = FeeManager::new(
             spec_id,
             self.get_base_fee(),
@@ -1101,6 +1104,7 @@ impl NodeConfig {
             self.get_gas_price(),
             self.get_blob_excess_gas_and_price(),
             self.get_blob_params(),
+            base_fee_params,
         );
 
         let (db, fork): (Arc<TokioRwLock<Box<dyn Db>>>, Option<ClientFork>) =
--- a/crates/evm/networks/src/lib.rs
+++ b/crates/evm/networks/src/lib.rs
@@ -9,7 +9,9 @@ use alloy_chains::{
     NamedChain,
     NamedChain::{Chiado, Gnosis, Moonbase, Moonbeam, MoonbeamDev, Moonriver, Rsk, RskTestnet},
 };
+use alloy_eips::eip1559::BaseFeeParams;
 use alloy_evm::precompiles::PrecompilesMap;
+use alloy_op_hardforks::{OpChainHardforks, OpHardforks};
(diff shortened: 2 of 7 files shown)


crates/anvil/src/config.rs
```diff
@@ -1094,13 +1094,20 @@ impl NodeConfig {
             self.networks,
         );
 
+        let base_fee_params = if self.networks.is_optimism() {
```
grandizzy: I suggest to move this in `self.networks.base_fee_params()` fn

crates/anvil/src/config.rs
```diff
@@ -1094,13 +1094,20 @@ impl NodeConfig {
...
+            BaseFeeParams::optimism_canyon()
```
grandizzy: should we try figuring out the optimisim hardfork in order to apply such? (not sure it has any value though but for backwards compatibility maybe) https://github.com/alloy-rs/hardforks/blob/bef91264a05e8a26711913a247cf9d6377fe80d8/crates/op-hardforks/src/lib.rs#L59

crates/anvil/src/config.rs
haythemsellami: Updated.

crates/anvil/src/config.rs
haythemsellami: Updated.

crates/evm/networks/src/lib.rs
```diff
@@ -47,6 +49,23 @@ impl NetworkConfigs {
         self.optimism
     }
 
+    /// Returns the base fee parameters for the configured network.
+    ///
+    /// For Optimism networks, returns Canyon parameters if the Canyon hardfork is active
+    /// at the given timestamp, otherwise returns pre-Canyon parameters.
+    pub fn base_fee_params(&self, timestamp: u64) -> BaseFeeParams {
```
mattsse, line 56: this makes sense

mattsse (approved):
makes sense, ty!

mattsse:
there's perhaps one more thing which we could tackle here or in a followup pr.

for forked anvil, on opstack we should extract the basefeeparams from the header's extradata past holocene

PR #12948, opened 2025-12-29
wallet: delegate TxSigner::address() to Signer::address()

aganisgash:
No need to duplicate the address logic, just call the existing Signer implementation.

### DIFF
--- a/crates/wallets/src/signer.rs
+++ b/crates/wallets/src/signer.rs
@@ -329,7 +329,7 @@ impl TxSigner<Signature> for WalletSigner {
         &self,
         tx: &mut dyn SignableTransaction<Signature>,
     ) -> alloy_signer::Result<Signature> {
-        <Self as Signer>::sign_transaction(self, tx).await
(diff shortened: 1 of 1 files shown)


crates/wallets/src/signer.rs
onbjerg, line 333: Can we do the same here? e.g. `<Self as Signer>::sign_transaction(self, tx)`?

onbjerg (changes requested):

crates/wallets/src/signer.rs
aganisgash: Nice catch expended to sign_transaction now

aganisgash: I tried to apply the same pattern but hit a limitation:

1. `sign_transaction` only exists in `TxSigner` trait, not in `Signer` 
2. Since `WalletSigner` is an enum, we still need the delegate! macro for pattern matching

but I've updated it to explicitly specify the trait: `TxSigner::sign_transaction(inner, tx)` 
instead of the implicit `inner.sign_transaction(tx)`. 

please take a look ;)

PR #12954, opened 2025-12-30
fix(config): fail on unknown config keys

### LINKED ISSUE #12792: Incorrect key in `foundry.toml` fails silently
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.5.0-stable Commit SHA: 1c57854462289b2e71ee7654cd6666217ed86ffd Build Timestamp: 2025-11-26T09:14:24.173470686Z (1764148464) Build Profile: maxperf

foundryup: 1.5.0

forge test

### Operating System

Linux

### Describe the bug

Misconfiguration in the `foundry.toml` file fails silently, even when adding a key-value pair that is obviously wrong. This can be a security risk if the particular configuration is critical.

Reproduce: add the entry `my-key-that-doesnt-exist = 42` anywhere in the file, and notice that it won't break anything.

This is problematic when the configuration is critical. For example, if I know that a prior solidity version contains a known bug, and I want to target a newest solidity version. One can think that `version = xxx` would work (as in the [corresponding compiler restriction key](https://getfoundry.sh/config/reference/solidity-compiler#compilation_restrictionversion) `version`), and be convinced that it works given that no exception are raised, when in fact `solc_version = xxx` should be used instead.

Ideally, a key that is not recognized would crash the program, so that this error is detected early

grandizzy:
## Motivation
- closes #12792

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/config/src/bind_json.rs
+++ b/crates/config/src/bind_json.rs
@@ -4,6 +4,7 @@ use std::path::PathBuf;
 
 /// Contains the config for `forge bind-json`
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
 pub struct BindJsonConfig {
     /// Path for the generated bindings file.
     pub out: PathBuf,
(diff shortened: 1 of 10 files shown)


onbjerg (changes requested):
I think this should be a warning

grandizzy:
superseded by https://github.com/foundry-rs/foundry/pull/13154

PR #12955, opened 2025-12-30
fix(config): reject bare versions in compilation restrictions

### LINKED ISSUE #12794: Misconfiguration of `version` in compiler restriction fails silently
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.5.0-stable Commit SHA: 1c57854462289b2e71ee7654cd6666217ed86ffd Build Timestamp: 2025-11-26T09:14:24.173470686Z (1764148464) Build Profile: maxperf

foundryup: 1.5.0

forge test

### Operating System

Linux

### Describe the bug

Misconfiguring the `version` to give it a bad value in compiler restrictions fails silently. This can be critical if a solidity version has a known bug and the goal was to avoid that bug.

The documentation about the `version` key [here](https://getfoundry.sh/config/reference/solidity-compiler#compilation_restrictionversion) redirects to `solc_version` which points [here](https://getfoundry.sh/config/reference/solidity-compiler#solc_version). We see:
> Only strict versions are supported (i.e. 0.8.11 is valid, but ^0.8.0 is not).

but in fact strict version (writing `version = "0.8.11"` for example) is not supported, and would fail silently.

Reproduce: check out [this PR](https://github.com/QGarchery/forge-compiler-restriction-version/pull/1). Notice notably that the first commit of this PR goes through the CI (when it shouldn't), and that the second commit fixes the issue.

Ideally, if a badly formatted string in the version is passed the program would crash, which would help detecting the error early

tefyosL-sol:
closes #12794

Version '0.8.11' silently matches 0.8.12+ due to caret parsing, require explicit operators like '=0.8.11' to prevent security issues.

### DIFF
--- a/crates/config/src/compilation.rs
+++ b/crates/config/src/compilation.rs
@@ -7,7 +7,7 @@ use foundry_compilers::{
     solc::{Restriction, SolcRestrictions},
 };
 use semver::VersionReq;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 
 /// Keeps possible overrides for default settings which users may configure to construct additional
 /// settings profile.
@@ -69,6 +69,7 @@ pub enum RestrictionsError {
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub struct CompilationRestrictions {
     pub paths: GlobMatcher,
+    #[serde(default, deserialize_with = "deserialize_version_req")]
     pub version: Option<VersionReq>,
     pub via_ir: Option<bool>,
     pub bytecode_hash: Option<BytecodeHash>,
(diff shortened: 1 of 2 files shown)


crates/config/src/compilation.rs
```diff
@@ -85,6 +86,42 @@ pub struct CompilationRestrictions {
     pub max_evm_version: Option<EvmVersion>,
 }
 
+/// Custom deserializer for version field that rejects ambiguous bare version numbers.
```
grandizzy, line 89: nit, we can use early return to make it cleaner

```rust
fn deserialize_version_req<'de, D>(deserializer: D) -> Result<Option<VersionReq>, D::Error>
where
    D: Deserializer<'de>,
{
    let opt_string: Option<String> = Option::deserialize(deserializer)?;
    let Some(opt_string) = opt_string else {
        return Ok(None);
    };

    let version = opt_string.trim();
    // Reject bare versions like "0.8.11" that lack an operator prefix
    if version.chars().next().is_some_and(|c| c.is_ascii_digit()) {
        return Err(serde::de::Error::custom(format!(
            "Invalid version format '{opt_string}' in compilation_restrictions. \
             Bare version numbers are ambiguous and default to caret requirements (e.g. '^{version}'). \
             Use an explicit constraint such as '={version}' for an exact version or '>={version}' for a minimum version."
        )));
    }

    let req = VersionReq::parse(&opt_string).map_err(|e| {
        serde::de::Error::custom(format!(
            "Invalid version requirement '{opt_string}': {}. \
             Examples: '=0.8.11' (exact), '>=0.8.11' (minimum), '>=0.8.11 <0.9.0' (range).",
            e
        ))
    })?;

    Ok(Some(req))
}
```

tefyosL-sol:
@grandizzy, done, much nicer this way, thanks for the feedback.

PR #12957, opened 2025-12-30
feat: custom errors lint

milosdjurica:
## Motivation

Add another linting rule suggesting use of custom errors over empty reverts and usage of strings in revert/require.  Checked out from this branch/PR -> https://github.com/foundry-rs/foundry/pull/12581

## Solution

Not sure if the rule is too demanding, could make it configurable to choose which cases should be linted and which no. For example, lint only `revert()` , or lint only reverts and requires when using strings, etc...

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


milosdjurica:
Hi @onbjerg , sorry for the late response. Opened a new PR containing only changes related to custom errors lint rule. You can check it out here: https://github.com/foundry-rs/foundry/pull/13126

Closing this one in favor of that PR.

PR #12959, opened 2025-12-30
fix(evm): use timestamp-based blob base fee calculation

cakevm:
## Motivation

`m.createSelectFork` fails for a tx that is in a block, that is after BPO1, and contains a blob tx.

The provides test fails without the fix:
```
Failing tests:
Encountered 1 failing test in test/BlobFork.t.sol:BlobForkTest
[FAIL: vm.createSelectFork: backend: failed committing transaction; transaction validation error: blob gas price (149641569615) is greater than max fee per blob gas (1000000000)] test_fork_with_blob_replay() (gas: 3303)
```
## Solution

The blob fee should be determined by the timestamp. The hardfork is not correctly detected without https://github.com/alloy-rs/hardforks/pull/86

The PR is just to showcase the fix, it required a new alloy-hardforks release. Me or a maintainer could update this PR.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -291,9 +291,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-hardforks"
-version = "0.4.6"
+version = "0.4.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "33bd636d42d97bddbc8363d3f23b2e0c46db0fb82e15cacbd8e079348f4728ee"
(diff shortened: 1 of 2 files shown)


mattsse (dismissed):
cool, ty!

mattsse (dismissed):

mattsse (approved):

PR #12960, opened 2025-12-30
docs: fix misleading comment in did_you_mean function

forkfury:
The comment incorrectly stated "by least confidence" while the function
sorts by similarity in ascending order, making the last element the
most similar. Updated the comment to accurately describe the sorting
behavior.


mattsse (commented):
we can keep this as is

PR #12968, opened 2025-12-31
fix(anvil): add cancellation to blocking RPC tasks

echowandere:
Was looking into the resource usage in forking mode and noticed that `on_blocking_task` doesn't handle client disconnects - tasks just keep running even after the client is gone.

Added a `CancellationToken` that listens for when the caller drops the future and aborts the blocking task, covers all the heavy RPC methods like eth_call, eth_estimateGas, and the mining stuff.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1171,6 +1171,7 @@ dependencies = [
  "tempfile",
  "thiserror 2.0.17",
  "tokio",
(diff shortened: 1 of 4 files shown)


mattsse (commented):
this doesnt quite work unfortunately because all fork ops are non async never reaching an await point

PR #12972, opened 2026-01-02
feat: add support for constructing tempo tx

onbjerg:
We want to add support for constructing Tempo transactions in e.g. Cast. This requires support in the primitives crate, which this PR adds.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/primitives/src/transaction/request.rs
```diff
@@ -46,17 +47,62 @@ impl FoundryTransactionRequest {
         Self(inner)
     }
 
+    /// Consume self and return the inner [`WithOtherFields<TransactionRequest>`].
+    #[inline]
+    pub fn into_inner(self) -> WithOtherFields<TransactionRequest> {
+        self.0
+    }
```
0xrusowsky, line 54: just flagging that this method seems to be unused in the PR.
good to leave it for convenience of potential future consumers, though.

DaniPopes (approved):

PR #12973, opened 2026-01-02
feat(cast): add tempo tx construction support

onbjerg:
Adds support for constructing Tempo transactions in `cast send` and `cast mktx`. Note that the implementation is a bit washed because of some existing debt we should address.

On top of https://github.com/foundry-rs/foundry/pull/12972

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/cast/src/cmd/mktx.rs
```diff
@@ -125,11 +126,14 @@ impl MakeTxArgs {
             return Ok(());
         }
 
+        // Check if this is a Tempo transaction
```
DaniPopes: slop comments

onbjerg:
@grandizzy I would hold off on that until we have trace decoding in maybe?

grandizzy:
> @grandizzy I would hold off on that until we have trace decoding in maybe?

sounds good

PR #12981, opened 2026-01-04
fix(cast): Improve vanity wallet regex validation

kyu-c:
## Motivation
`cast wallet vanity` doesn't sufficiently validate the inputs.

For example,  running `cast wallet vanity --starts-with deadbeeg` doesn't cause an error and will run forever in attempt to generate an address that isn't even valid. 

## Solution

Check if the characters in regex patterns are at least valid hex characters

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


kyu-c:
@onbjerg It won't reject a regex that contains `^` and `$` as valid hex character check only happens if  a character is `is_alphanumeric`. 

```
cargo run --bin cast -- wallet vanity --starts-with '^11.*11$'
...
Address: 0x11838A2Ed82dCe4f86378990d659C7FCfF2bb611
...
```

PR #12993, opened 2026-01-05
merge master upstream in dani/parallel-fuzz

grandizzy:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/evm/evm/src/executors/fuzz/mod.rs
```diff
@@ -464,6 +464,7 @@ impl FuzzedExecutor {
         let sync_threshold = SYNC_INTERVAL + sync_offset;
         let mut runs_since_sync = sync_threshold; // Always sync at the start.
         let mut last_metrics_report = Instant::now();
+        let mut executor = self.executor.clone();
```
grandizzy, line 467: @DaniPopes any suggestion if we could accomplish https://github.com/foundry-rs/foundry/pull/12843 this without cloning the executor?

grandizzy: ping @DaniPopes any change here or should be good? ty!

DaniPopes: this is fine

PR #13000, opened 2026-01-06
feat(cast): batch send/sign/mktx in cast

### LINKED ISSUE #12987: Batch sending in `cast`
### Component

Cast

For `cast send`, `cast sign` and `cast mktx` (any others?) we should support a batch mode where you can actually construct 1 or more transactions, each with their own value (optional), input (optional) and destination address. I'm not entirely sure what good UX would be here.

For e.g. Ethereum/Optimism or other chains without native batching capability, this would just result in multiple transactions. For networks with native batching support, such as Tempo, this should result in 1 transaction.

Some care needs to be taken to ensure you can't construct "bogus" batches, e.g. transactions with the same nonce. It might be preferable to simply manage the nonce for the user in this case, e.g. if I am on nonce 3 and I send a batch of 3 transactions, they would have nonce 4, 5, 6 respectively. For Tempo this would just result in a transaction with nonce 4 (since it has native batching).

### Additional context

There is some documentation on call batching in Tempo [here](https://docs.tempo.xyz/protocol/transactions/spec-tempo-transaction#transaction-type).

stevencartavia:
## Motivation
closes #12987

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cast/src/cmd/batch_mktx.rs
+++ b/crates/cast/src/cmd/batch_mktx.rs
@@ -1,4 +1,4 @@
-use crate::tx::CastTxBuilder;
+use crate::{tx::CastTxBuilder, tx_spec::TxSpec};
 use alloy_eips::Encodable2718;
 use alloy_ens::NameOrAddress;
 use alloy_network::{EthereumWallet, TransactionBuilder};
@@ -22,7 +22,7 @@ pub struct BatchMakeTxArgs {
     ///   --tx "0x123:0.1ether"  (ETH transfer)
     ///   --tx "0x456::transfer(address,uint256):0x789,1000"  (contract call)
     ///   --tx "0xabc::0x123def"  (raw data)
-    #[arg(long = "tx", value_name = "SPEC", value_delimiter = ',')]
+    #[arg(long = "tx", value_name = "SPEC", value_delimiter = ';')]
     pub transactions: Vec<String>,
 
     #[command(flatten)]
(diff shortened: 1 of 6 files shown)


crates/cast/src/cmd/batch_send.rs
```diff
@@ -0,0 +1,244 @@
...
+}
+
+/// Parsed transaction specification
+#[derive(Debug, Clone)]
+pub struct TxSpec {
+    pub to: String,
+    pub value: Option<String>,
+    pub sig: Option<String>,
+    pub args: Vec<String>,
+}
```
mattsse: are these parsed differently?

if not we can unify them and avoid duplicating parsing

crates/cast/src/cmd/batch_mktx.rs
```diff
@@ -0,0 +1,234 @@
...
+use std::str::FromStr;
+
+/// CLI arguments for `cast batch-mktx`.
+#[derive(Debug, Parser)]
+pub struct BatchMakeTxArgs {
+    /// Transaction specifications in format: to\[:value\]\[:sig\[:args\]\]
+    ///
+    /// Examples:
+    ///   --tx "0x123:0.1ether"  (ETH transfer)
+    ///   --tx "0x456::transfer(address,uint256):0x789,1000"  (contract call)
```
mattsse, line 23: can we add some clap test for this, because I believe the signature must be wrapped in a string due to the `,`?

mattsse (changes requested):

crates/cast/src/cmd/batch_send.rs
stevencartavia: unified

crates/cast/src/cmd/batch_mktx.rs
stevencartavia: updated it to `;`

PR #13002, opened 2026-01-06
feat(cast): add tx flags to `erc20` command

### LINKED ISSUE #12991: Add transaction flags to `cast erc20` commands
### Component

Cast

Commands that build transactions in `cast erc20` (e.g. `cast erc20 transfer`) should have the regular transaction flags, such as gas price, nonce.

See [`TransactionOpts`](https://github.com/foundry-rs/foundry/blob/2417ba3bb7a55d70867247d75cc989f976f17452/crates/cli/src/opts/transaction.rs#L36)

### Additional context

_No response_

mablr:
## Motivation

Close #12991 

## Solution

- Introduced `Erc20TxOpts` struct to encapsulate transaction options such as gas limit, gas price, priority gas price, and nonce for ERC20 operations. `TransactionOpts` is not usable here as it would imply collisions/shadowing, e.g. `value` vs `amount`.
- Implemented `apply_tx_opts` function to apply transaction options to transaction requests.
- Added tests to validate the new transaction options

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,7 +84,7 @@ jobs:
           persist-credentials: false
       - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # master
         with:
-          toolchain: nightly
+          toolchain: nightly-2026-01-10
           components: clippy
       - uses: rui314/setup-mold@725a8794d15fc7563f59595bd9556495c0564878 # v1
       - uses: mozilla-actions/sccache-action@7d986dd989559c6ecdb630a3fd2557667be217ad # v0.0.9
@@ -104,7 +104,7 @@ jobs:
           persist-credentials: false
       - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # master
         with:
-          toolchain: nightly
+          toolchain: nightly-2026-01-10
           components: rustfmt
       - run: cargo fmt --all --check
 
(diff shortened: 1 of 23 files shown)


onbjerg (approved):
ty as always

onbjerg (changes requested):
ah actually, this reminds me, the tempo options will not result in a tempo tx because this still uses `AnyNetwork` and the ethereum envelopes + `TransactionRequest`. we'd need to use `FoundryNetwork`/`FoundryTransactionRequest`. there is a small workaround in `cast send` we could do instead, but up to you. eventually we do need to replace `AnyNetwork` but I recognize that this is a larger chunk of work.

in order to check it works, these are the instructions:

- `cast wallet new`
- `cast rpc tempo_fundAddress <address> --rpc-url https://rpc.testnet.tempo.xyz`
- `cast erc20 transfer 0x20c0000000000000000000000000000000000000 $(cast az) 1000 --tempo.fee-token 0x20c0000000000000000000000000000000000003`

Then check the receipt/tx. It has to have type 0x76 and a fee token of `0x20c0..03`. it's ok if the tx itself fails as long as the encoding is correct.

onbjerg:
@mablr Did you fund the address? or can you share a tx hash perhaps

mablr:
@onbjerg Yes I fund a test address following tempo doc. Here is the address 0xa70ab0448e66cD77995bfBBa5c5b64B41a85F3fd.
When I'm debugging, I see the tx built is of tempo type (0x76). I get the tx_hash but never receive the receipt (that's why I posted the request and the tx in my previous message)


But now it seems that Tempo is broken :/ (getting 502 Bad Gateway errors, maybe due to migration ?)

mablr (commented):
Ok it works now on new testnet https://explore.tempo.xyz/tx/0x042f03f5c576c9a75eaff4215bab019352e7db2a98fa43758328d70d7de60278

PR #13003, opened 2026-01-06
fix(cheatcodes): return correct 4-byte value from randomBytes4()

strmfos:
`randomBytes4()` was incorrectly using `B32` (32 bytes) instead of `FixedBytes<4>` (4 bytes), causing the function to return a 32-byte value padded with zeros instead of the expected 4-byte random value as specified in the cheatcode signature `function randomBytes4() external view returns (bytes4)`.


onbjerg:
all words in the evm are 32 bytes

PR #13019, opened 2026-01-08
chore(anvil): rm dead code

onbjerg:
## Motivation

Cleanup

## Solution

Remove dead/unused code.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -2794,15 +2794,6 @@ impl EthApi {
         Ok(blocks)
     }
 
-    /// Sets the reported block number
-    ///
-    /// Handler for ETH RPC call: `anvil_setBlock`
-    pub fn anvil_set_block(&self, block_number: u64) -> Result<()> {
-        node_info!("anvil_setBlock");
-        self.backend.set_block_number(block_number);
-        Ok(())
-    }
-
     /// Sets the backend rpc url
     ///
     /// Handler for ETH RPC call: `anvil_setRpcUrl`
--- a/crates/anvil/src/eth/fees.rs
+++ b/crates/anvil/src/eth/fees.rs
@@ -191,16 +191,6 @@ impl FeeManager {
     }
 }
 
-/// Calculate base fee for next block using Ethereum mainnet parameters.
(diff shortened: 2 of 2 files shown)


onbjerg:
which ones? there are some that we can def remove because there are equivalents in alloy (e.g. base fee calc)

DaniPopes:
idk maybe the inspect. @mattsse thoughts

crates/anvil/src/eth/fees.rs
```diff
@@ -191,16 +191,6 @@ impl FeeManager {
...
 
-/// Calculate base fee for next block using Ethereum mainnet parameters.
-///
-/// See [EIP-1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md) spec.
-///
-/// Note: This uses Ethereum's base fee params. For network-specific calculations
-/// (e.g., Optimism), use [`FeeManager::get_next_block_base_fee_per_gas`] instead.
-pub fn calculate_next_block_base_fee(gas_used: u64, gas_limit: u64, base_fee: u64) -> u64 {
-    calc_next_block_base_fee(gas_used, gas_limit, base_fee, BaseFeeParams::ethereum())
-}
```
mattsse, line 202: this we can remove for sure

mattsse (dismissed):
I'd rather do this on a case by case basis, keeping the pub fn doesnt really hurt

crates/anvil/src/eth/api.rs
```diff
@@ -2794,15 +2794,6 @@ impl EthApi {
...
     }
 
-    /// Sets the reported block number
-    ///
-    /// Handler for ETH RPC call: `anvil_setBlock`
-    pub fn anvil_set_block(&self, block_number: u64) -> Result<()> {
-        node_info!("anvil_setBlock");
-        self.backend.set_block_number(block_number);
-        Ok(())
-    }
```
onbjerg, line 2804: i think we should also remove this then if we are doing case by case, this is entirely unused. there is no `anvil_setBlock` @mattsse

PR #13027, opened 2026-01-08
refactor(script): remove redundant clone in build compilation

0xlupin:
Removes unnecessary `to_path_buf()` clone when chaining source files for compilation by using `iter::once()` instead.


DaniPopes:
doesn't compile, needs clone to work

PR #13029, opened 2026-01-09
feat: uncomment transaction replacement test

echowandere:
The FIXME referenced alloy-rs/alloy#614 which was merged back in April 2024. Since we're now on alloy 1.1.3, the receipt polling issue has been fixed for a while. 

Restored the test assertions with a small fix - removed the attempt to get receipt for the replaced transaction since it never gets mined.


DaniPopes:
doesn't compile

echowandere:
> doesn't compile

All good now, removed the attempt to get receipt for the replaced transaction since it never gets mined.

PR #13032, opened 2026-01-09
Deduplicate call input lookup in inspector stack

splinter012:
Remove double call.input.bytes(ecx) by caching input_bytes when resolving mocked functions; behavior unchanged, avoids redundant work/allocation. Scope: foundry/crates/evm/evm/src/inspectors/stack.rs.


crates/evm/evm/src/inspectors/stack.rs
```diff
@@ -963,12 +963,13 @@ impl Inspector<EthEvmContext<&mut dyn DatabaseExt>> for InspectorStackRefMut<'_>
 
         if let Some(cheatcodes) = self.cheatcodes.as_deref_mut() {
             // Handle mocked functions, replace bytecode address with mock if matched.
+            let input_bytes = call.input.bytes(ecx);
```
DaniPopes: wrong scope, move down one line

splinter012: Done

PR #13041, opened 2026-01-10
Persist environment's `Spec` on fork creations

### LINKED ISSUE #13040: bug(forge): forks incorrectly use default hardfork instead of configured evm_version
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.5.1-dev Commit SHA: e589ccfc404f878b77c1e6a5ea85023ad99482b6

foundryup: 1.5.0

forge test

### Operating System

Linux

### Describe the bug

When creating forks, [the default `Spec` (hardfork) is being used for it](https://github.com/foundry-rs/foundry/blob/e589ccfc404f878b77c1e6a5ea85023ad99482b6/crates/evm/core/src/fork/init.rs#L109) (currently `prague`).
This is surprising as I would expect it to either use the fork's actual EVM version or keep the environment's `Spec` (which is the behavior that the [section on `evm_version` configuration](https://getfoundry.sh/forge/tests/fork-testing#evm-version) suggests).

The following test fails:
```solidity
/// forge-config: default.evm_version = "cancun"
function testKeepsEvmVersion() public {
    assertEq(vm.getEvmVersion(), "cancun");
    vm.createSelectFork("anyfork", 1);
    vm.assertEq(vm.getEvmVersion(), "cancun"); // is "prague"
}
```

die-herdplatte:
Fixes #13040 by creating forks with the environment's `Spec`.

Another option would have been to try and determine the `Spec` by chain id and block timestamp with `alloy_hardforks` and `alloy_op_hardforks` but I decided against it because it would introduce unpredictability when the mappings aren't known for a chain or are lagging behind (which is opaque to the user).

This is a breaking change.

### DIFF
--- a/crates/evm/core/src/fork/multi.rs
+++ b/crates/evm/core/src/fork/multi.rs
@@ -539,7 +539,8 @@ async fn create_fork(mut fork: CreateFork) -> eyre::Result<(ForkId, CreatedFork,
     let provider = fork.evm_opts.fork_provider_with_url(&fork.url)?;
 
     // Initialise the fork environment.
-    let (env, block) = fork.evm_opts.fork_evm_env_with_provider(&fork.url, &provider).await?;
+    let (mut env, block) = fork.evm_opts.fork_evm_env_with_provider(&fork.url, &provider).await?;
+    env.evm_env.cfg_env.spec = *fork.env.evm_env.spec_id();
     fork.env = env;
     let meta = BlockchainDbMeta::new(fork.env.evm_env.block_env.clone(), fork.url.clone());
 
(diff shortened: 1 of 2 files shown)


emmajam:

die-herdplatte:
The original issue hasn't been fixed, so this should remain open

die-herdplatte:
I can't reopen @emmajam

figtracer:
@die-herdplatte can you fix conflicts

mablr (commented):
@figtracer It's going to be hard to reconcile, `Env` struct was withdrawn, and `CreateFork` doesn't hold this info anymore.

die-herdplatte:
Closing in favor of #14870

PR #13054, opened 2026-01-11
feat(cheatcodes): support both 4844/7594 formats in `attachBlob`

### LINKED ISSUE #12363: bug: `vm.attachBlob` will probably stop working
### Component

Forge

- [x] Foundry
- [ ] Foundryup

1.4.3

1.3.0

forge script

### Operating System

Linux

### Describe the bug

Since we're currently using an older version of alloy, it probably won't work on testnet like Hoodi

mablr:
## Motivation

Close #12363

## Solution

- Use `BlobTransactionSidecarVariant` in `attachBlob` cheatcode support both 4844/7594 formats

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/cheatcodes/src/script.rs
```diff
@@ -228,8 +228,12 @@ impl Cheatcode for attachBlobCall {
...
         );
         let sidecar: SidecarBuilder<SimpleCoder> = SidecarBuilder::from_slice(blob);
-        let sidecar = sidecar.build().map_err(|e| format!("{e}"))?;
-        ccx.state.active_blob_sidecar = Some(sidecar);
+        let sidecar_variant = if ccx.ecx.cfg.spec < SpecId::OSAKA {
+            sidecar.build_4844().map_err(|e| format!("{e}"))?.into()
+        } else {
+            sidecar.build_7594().map_err(|e| format!("{e}"))?.into()
+        };
+        ccx.state.active_blob_sidecar = Some(sidecar_variant);
```
mablr, line 236: I propose to select implicitly the blob format based on the spec.

Is there a real use-case for explicit blob format selection ? i.e. use 4844 one after Osaka ?

StackOverflowExcept1on (approved):

crates/cheatcodes/src/script.rs
mattsse: > Is there a real use-case for explicit blob format selection ? i.e. use 4844 one after Osaka ?

there is not, this format is deprecated on protocol level after osaka

PR #13058, opened 2026-01-12
feat(cast): add eip7594 support

### LINKED ISSUE #12449: Support 7594 in cast mktx
### Component

Cast

ref https://github.com/foundry-rs/foundry/issues/12222

## TODO
* add param to specify 7594 format, which will eventually become the default after fusaka activates

@mablr 

cc @pcaversaccio 

### Additional context

_No response_

mablr:
## Motivation

Close #12449 

## Solution

- Added support for EIP-7594 (PeerDAS) in transaction handling, allowing users to specify the new blob format.
- Added tests for both EIP-4844 and EIP-7594 blob transactions in the CLI.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes (no, since eip7594 format is not the default one yet)

### DIFF
--- a/crates/cast/src/tx.rs
+++ b/crates/cast/src/tx.rs
@@ -313,8 +313,8 @@ pub struct CastTxBuilder<P, S> {
     /// Whether the transaction should be sent as a legacy transaction.
     legacy: bool,
     blob: bool,
-    /// Whether the blob transaction should use EIP-7594 format instead of EIP-4844.
-    peerdas: bool,
+    /// Whether the blob transaction should use EIP-4844 (legacy) format instead of EIP-7594.
+    eip4844: bool,
     auth: Vec<CliAuthorizationList>,
     chain: Chain,
     etherscan_api_key: Option<String>,
@@ -375,7 +375,7 @@ impl<P: Provider<AnyNetwork>> CastTxBuilder<P, InitState> {
             tx,
             legacy,
             blob: tx_opts.blob,
-            peerdas: tx_opts.peerdas,
+            eip4844: tx_opts.eip4844,
             chain,
             etherscan_api_key,
             auth: tx_opts.auth,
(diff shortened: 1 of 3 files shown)


crates/cast/src/tx.rs
```diff
@@ -668,9 +673,15 @@ where
...
-        let sidecar = coder.build()?;
 
-        self.tx.set_blob_sidecar(sidecar);
+        if self.peerdas {
+            let sidecar = coder.build_7594()?;
+            self.tx.set_blob_sidecar_7594(sidecar);
+        } else {
+            let sidecar = coder.build()?;
+            alloy_network::TransactionBuilder4844::set_blob_sidecar(&mut self.tx, sidecar);
+        }
```
mablr: lmk if you'd rather prefer setting eip7594 format by default

mattsse: yeah let's do 7594 by default now

crates/cli/src/opts/transaction.rs
```diff
@@ -78,10 +78,18 @@ pub struct TransactionOpts {
     pub legacy: bool,
 
     /// Send a EIP-4844 blob transaction.
+    ///
+    /// Note: This is the current default blob variant. After the Fusaka hardfork,
+    /// EIP-7594 (PeerDAS) will become the default blob variant.
     #[arg(long, conflicts_with = "legacy")]
     pub blob: bool,
```
mattsse: lets do 7594 by default now

crates/cast/src/tx.rs
mablr: done🫡

mattsse (approved):

satoshinakamotovrxyz-dotcom:
mablr:feat/cast_eip7594_support

PR #13061, opened 2026-01-12
chore(common): rm unused needs_clear flag from Shell

kurahin:
The Shell implementation carried over Cargo's needs_clear flag and related logic, but nothing in Foundry ever set this flag, so the erase-line sequence was effectively dead code. This change drops the AtomicBool field and its accessors and turns maybe_err_erase_line into an explicit no-op, keeping existing behavior while simplifying the type and removing unnecessary synchronization overhead.

### DIFF
--- a/crates/common/src/io/shell.rs
+++ b/crates/common/src/io/shell.rs
@@ -12,10 +12,7 @@ use std::{
     fmt,
     io::{IsTerminal, prelude::*},
     ops::DerefMut,
-    sync::{
-        Mutex, OnceLock, PoisonError,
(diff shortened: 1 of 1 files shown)


crates/common/src/io/shell.rs
```diff
@@ -331,14 +307,7 @@ impl Shell {
     }
 
     /// Erase from cursor to end of line if needed.
-    pub fn maybe_err_erase_line(&mut self) {
-        if self.err_supports_color() && self.set_needs_clear(false) {
```
onbjerg, line 335: it's used right here in the code you deleted

PR #13072, opened 2026-01-13
refactor: remove code duplication in `locked_read_to_string`

sashaodessa:
The `locked_read_to_string` function duplicated the entire file locking and reading logic from `locked_read`, differing only in the final UTF-8 conversion step. This code duplication increases maintenance burden and risk of divergence between the two functions when changes are made to the file locking logic.

Refactor `locked_read_to_string` to delegate the file reading logic to `locked_read`, making it a thin wrapper that only handles UTF-8 conversion. This eliminates code duplication while maintaining identical behavior.


mattsse (commented):
we can keep as is

PR #13073, opened 2026-01-13
perf(verify): remove unnecessary clone in check_explorer_args

0xlupin:
Eliminates redundant allocation in check_explorer_args by consuming the owned ContractMetadata parameter instead of cloning its contents.


onbjerg (changes requested):
this is worse

PR #13081, opened 2026-01-14
chore(evm): misleading error message in traces serialization

tefyosL-sol:
Corrected an error message in `render_trace_arena_inner` where `serde_json::to_string` failure was reported as "Failed to write traces". Changed to "Failed to serialize traces" to accurately reflect that this is a serialization operation, not an I/O write operation.


tefyosL-sol:
Looks like this is ready to go but automerge is pending. cc @onbjerg

onbjerg:
yeah its flaky dw

PR #13088, opened 2026-01-14
feat(coverage): add experimental ``--instrument-source`` source-to-source instrumentation engine

derekbar90:
###  Motivation
I'm tired of being on teams/projects and having to do major rewrites just to get a coverage report as we head into audit preparedness. Traditional bytecode-based coverage in forge coverage relies heavily on the accuracy of EVM source maps. These mappings frequently break or become ambiguous when using viaIR, aggressive compiler optimizations, or very complex contract structures. 

This often results in **"stack too deep"** errors during instrumentation or inaccurate/missing coverage data for production-grade contracts. The motivation for this change is to provide a robust, alternative instrumentation path that is independent of source map accuracy, allowing users to get reliable coverage signals even for their most complex logic.

###   Solution
This PR introduces an experimental source-to-source instrumentation engine. Instead of instrumenting the bytecode and mapping back to source, this engine transforms the Solidity source code directly to track coverage hits.

### Key Technical Components:
   - AST-Based Transformation: Leverages the Solar compiler framework to perform precise AST visitation. The `Instrumenter` visitor injects `coverageHit` calls into statements, branches, and function bodies while preserving original control flow.
   - Runtime Interception: Implements `SourceCoverageCollector`, a specialized EVM Inspector that intercepts the injected cheatcode calls during execution to record hit data.
   - Unified Pipeline: Integrates the engine into the existing CoverageArgs workflow. When the ``--instrument-source`` flag is used, the project is transparently built in a isolated temporary directory (using tempfile), and results are unified into the standard `CoverageReport` format.

  PR Checklist

   - [x] Added Tests
   - [x] Added Documentation
   - [NONE] Breaking changes


onbjerg (changes requested):
Hi, while we appreciate the motivation, we do not think the correct solution is to implement a completely separate/alternative way to do coverage. We would rather that we revisit how forge coverage works now, which we have some ideas for. This PR introduces a completely separate way to perform coverage, and is a large change we'd have to maintain. for that reason I am closing this PR.

onbjerg:
Actually, I'd like some input from @grandizzy @zerosnacks  first on second thought as this is somewhat directionally the way we would want to take it in, but ideally this replaces the current mechanism instead of being a separate mode.

grandizzy:
Agree, we plan to improve coverage through solar bytecode instrumentation (on the roadmap, the first step was done by replacing solang) so make sense not to approach and maintain a foundry impl. @derekbar90 appreciate if you willing to give a hand to improve solar in this direction, thanks!

derekbar90:
I understand how you feel @grandizzy but wouldn't an interim solutions help the industry as a whole? I've gone through dozens of audits at this point and every time there is a stack to deep issue somewhere in the code. Depending on the developers skill level that could take them days. It also allows teams to know their security posture as they head into audit.

That said, I understand your reasoning. Let me review the solar codebase and see whats up. This was a good amount of work so i'll see if it's possible to port easily.

derekbar90:
I've begun the work, it should be ready sometime next week. I have to update forge to use the new solar dependency and so i'll update this PR with those changes as this change now requires two repos to be updated independently. @grandizzy

grandizzy:
@derekbar90 that's awesome, thank you!

derekbar90:
@grandizzy and team, I’ve spent time prototyping the migration to a Solar-native bytecode instrumentation backend as suggested. While I fully agree that Solar is the right long-term home for this logic, my investigation reveals that moving it today would be premature and would actually introduce more 'hacky' complexity into Forge than the current source-rewriting approach.

**Technical Findings in Solar:**

   1. **Storage Semantics Gap**: Solar’s HIR-to-Yul lowering is currently in its infancy. Specifically, it lacks a storage-slot mapping engine. When lowering Solidity to Yul, it currently treats state variables as local Yul variables rather than emitting the necessary sstore/sload logic based on the contract's storage layout. This means that while we can instrument the Yul, the resulting contracts are non-functional for any test that relies on state persistence.
   2. **Yul Compilation Bottleneck**: solc --yul is strictly limited to exactly one input file. Supporting modern, multi-file Solidity projects via a Yul pipeline would require Forge to either:
       * Implement a brittle loop that compiles Yul files individually and manually merges artifacts (breaking cross-contract dependencies).
       * Implement a complex Standard JSON generator for Yul that mocks a virtual file system.
   3. **Metadata and Tooling Loss**: By bypassing the Solidity compiler and moving to a Yul-only pipeline, we lose critical Solidity-specific metadata (Natspec, storage layouts, and developer documentation) that users and downstream tools expect in Forge artifacts.

**The Case for the Current Approach:**
The Foundry-side source instrumentation I’ve implemented uses Solar's robust AST visitor to perform surgical rewrites, but then hands the code back to solc's Solidity pipeline. This allows us to:
   * Inherit `solc`’s Maturity: We get storage management, optimized stack handling, and full metadata support for free.
   * Support Multi-file Projects: It works out-of-the-box with existing ProjectCompiler workflows.
   * Maintain Performance: It avoids the overhead of managing a custom Yul-to-EVM toolchain inside Forge.

**Path Forward**:
I suggest we treat the Foundry-side instrumentation as Phase 1. It is a 'working today' solution that provides accurate coverage without requiring us to first finish implementing a Solidity-to-EVM backend inside Solar. 

I have already refactored the PR to use Solar's library for the AST traversal, keeping the logic much cleaner than a regex-based approach. Once Solar reaches parity with solc regarding storage and multi-file Standard JSON support, we can migrate the backend with minimal changes to the Forge CLI logic.

I’d appreciate it if we could move forward with this refined Foundry implementation so we can get these coverage improvements into the hands of users.

gakonst:
ACKing your Twitter comment, I'll discuss internally and circle back as I see the value AND I want it to work w/ Solar.

derekbar90:
Agreed, sadly Solar isn't ready for this change yet. It needs to mature in a few areas and then this change could be migrated into a native feature. More discussed above.

emmajam:

PR #13093, opened 2026-01-15
feat(cheatcodes): add getRecordedLogsJson cheatcode

### LINKED ISSUE #12854: getRecordedLogs[Json]
### Component

Forge

Similar to `getStateDiffJson` it would be nice to have a simple built in way of extracting Logs, to be processed externally.

I'm not quite convinced that this is the correct way to do it.
Perhaps it would be better to just have a more generic way of running a script/test with `--json path` and writing a well defined structure including everything somewhere.

### Additional context

We heavily rely on tenderly simulations, which is fine, but also makes us quite reliant on tenderly. Whenever there is an issue, we cannot really help outselfes. Therefore for some time we've been building tooling around foundry to "generate sth similar".

What we are usually curious about is:
- the contracts touched. Are they verified? Are they upgradable? Who owns them? Is there a way they can self-destruct?
- the storage being altered.
- the logs being emitted.
- the trace.

Currently what we do is using:
-`startStateDiffRecording` -> `getStateDiffJson`
- `recordLogs` -> custom transform to json -> `writeJson`

Then we post-process this data.
It's quite imperfect as:
- stateDiff currently does not decode/trace the changes in mappings
- contracts touched we infer from stateDiff, which i think is incomplete
- logs we decode via a locally maintained library of known logs
- code we fetch from rpc / etherscan

I think the post processing itself, does not really make sense to be foundry native as obviously it's very opinionated, but wondering if there could be a more holistic `record*` that records everything in a reliable format so ppl can do more generalized post processing.

grandizzy:
## Summary

Adds a new cheatcode `getRecordedLogsJson` that returns recorded logs as a JSON string, similar to the existing `getStateDiffJson` pattern.

This allows users to easily post-process recorded logs externally without needing to manually transform the `Log[]` array to JSON.

## Motivation

Closes #12854

Users heavily rely on log analysis for simulation and auditing purposes. Currently they have to use `recordLogs` -> custom transform to JSON -> `writeJson`, which is cumbersome. This cheatcode provides a native way to get recorded logs as JSON for external processing.

## JSON Format

```json
[
  {
    "topics": ["0x...", "0x..."],
    "data": "0x...",
    "emitter": "0x..."
  }
]
```

All binary fields are hex-encoded with `0x` prefix.

## Changes

- Added `getRecordedLogsJson()` function declaration in `vm.rs`
- Added `LogJson` struct for JSON serialization
- Implemented the cheatcode following the same pattern as `getStateDiffJson`
- Added tests verifying JSON structure and values

## Testing

- `testRecordedLogsJson`: Verifies emitter address, topics (event signature + indexed args), and hex-encoded data
- `testRecordedLogsJsonEmpty`: Verifies empty logs returns `[]`

### DIFF
--- a/crates/cheatcodes/assets/cheatcodes.json
+++ b/crates/cheatcodes/assets/cheatcodes.json
@@ -6526,6 +6526,26 @@
     },
     {
       "func": {
+        "id": "getRecordedLogsJson",
+        "description": "Gets all the recorded logs, in JSON format.",
+        "declaration": "function getRecordedLogsJson() external view returns (string memory logsJson);",
+        "visibility": "external",
+        "mutability": "view",
+        "signature": "getRecordedLogsJson()",
+        "selector": "0x3b171111",
+        "selectorBytes": [
+          59,
+          23,
+          17,
+          17
+        ]
+      },
+      "group": "evm",
+      "status": "stable",
+      "safety": "safe"
+    },
+    {
+      "func": {
         "id": "getStateDiff",
         "description": "Returns state diffs from current `vm.startStateDiffRecording` session.",
         "declaration": "function getStateDiff() external view returns (string memory diff);",
--- a/crates/cheatcodes/spec/src/vm.rs
+++ b/crates/cheatcodes/spec/src/vm.rs
@@ -951,6 +951,10 @@ interface Vm {
     #[cheatcode(group = Evm, safety = Safe)]
     function getRecordedLogs() external view returns (Log[] memory logs);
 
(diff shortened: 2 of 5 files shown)


sakulstra:
Hey, I'm a bit confused - how can i use this on latest nightly? Do i need to wait for a release of forge-std, as it seems to not to be available yet.

sakulstra:
Added a pr here https://github.com/foundry-rs/forge-std/pull/809 as i think it's missing (on docs it is mentioned, but seems to not exist)

PR #13097, opened 2026-01-15
perf: add dist profile for smaller release binaries

onbjerg:
Add a new `dist` Cargo profile optimized for distribution:
- Fat LTO and codegen-units=1 for better optimization
- Strip symbols for smaller binaries
- opt-level="s" overrides for non-perf-critical dependencies

Benchmarks on Solady test suite show dist is 8% faster than release while being 45% smaller (43MB vs 78MB).

Update release workflows to use the dist profile instead of maxperf.

| Profile | Size | Time | vs dist |
|---------|------|------|---------|
| **dist** | 43MB | 1.97s ± 0.06s | baseline |
| release (this branch) | 78MB | 2.10s ± 0.07s | 7% slower |
| release (master) | 75MB | 2.10s ± 0.04s | 7% slower |

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -96,7 +96,6 @@ inherits = "profiling"
 inherits = "release"
 lto = "fat"
 codegen-units = 1
-strip = "debuginfo"
 
 # Speed up tests and dev build.
 [profile.dev.package]


Cargo.toml
```diff
@@ -92,10 +92,13 @@ strip = "none"
 [profile.bench]
 inherits = "profiling"
 
-[profile.maxperf]
+# Optimized for distribution: smaller binary size while maintaining performance.
+# Build with: cargo build --profile dist
```
DaniPopes: rm comment

Cargo.toml
```diff
@@ -92,10 +92,13 @@ strip = "none"
...
+[profile.dist]
 inherits = "release"
 lto = "fat"
 codegen-units = 1
+strip = "symbols"
```
DaniPopes: let's keep symbols, they're useful for debugging stack traces and profiling

onbjerg: this adds 6 megabytes, I think thats fine. pushed

Cargo.toml
```diff
@@ -92,10 +92,11 @@ strip = "none"
 [profile.bench]
 inherits = "profiling"
 
-[profile.maxperf]
+[profile.dist]
 inherits = "release"
 lto = "fat"
 codegen-units = 1
+strip = "debuginfo"
```
DaniPopes:
```suggestion
codegen-units = 1
```

PR #13101, opened 2026-01-15
chore(fmt): replace RingBuffer unwraps with expect

Olexandr88:
Replace unwrap() with expect() in RingBuffer to provide clearer panic messages while preserving existing invariants


DaniPopes:
not necessary

PR #13105, opened 2026-01-15
chore: bump revm and patch dependencies with staging-revm branches

rakita:
## Summary
- Bump revm and related dependencies to staging versions
- Update code to match new API changes in revm

## Changes
- Updated Cargo.toml with patched dependencies pointing to staging-revm branches
- Modified anvil backend code for API compatibility
- Updated cheatcodes inspector utilities
- Adjusted EVM core and inspector code for new revm interfaces

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,7 +148,7 @@ jobs:
       - run: cargo hack check
 
   deny:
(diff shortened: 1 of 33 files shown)


rakita:
@onbjerg revm is released only pending release are on revm-precompiles, foundry-fork-db and alloy/evm

PR #13108, opened 2026-01-16
feat(fmt): Add `namespace_import_style` config

### LINKED ISSUE #12726: `forge fmt` changes import formatting when it shouldn't
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.5.0-stable

foundryup: 1.4.0

forge fmt

### Operating System

Linux

### Describe the bug

I want to use an import statement like this:
```Solidity
import * as Owner from "../../access/Owner/Owner.sol";
```
But `forge fmt` reformats it to this:
```Solidity
import "../../access/Owner/Owner.sol" as Owner;
```

I don't want that.  What is the best way to prevent this?  Or can this be changed or configured?

quangloc99:
## Motivation

Fixes #12726.

## Solution

This PR introduces a new config in the formatter called `namespace_import_style` which allows users to specify their preferred style for namespace imports.

- `prefer_plain` for the style `import "foo" as foo;`.
- `prefer_glob` for the style `import * as foo from "foo";`.
- `preserve` to keep the existing style.

The default is `prefer_plain` to keep the existing behavior.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

I'm happy to add the documentation for this feature too.

### DIFF
--- a/crates/fmt/README.md
+++ b/crates/fmt/README.md
@@ -127,6 +127,7 @@ The formatter supports multiple configuration options defined in `foundry.toml`.
 | `ignore` | `[]` | Globs to ignore. |
 | `contract_new_lines` | `false` | Add a new line at the start and end of contract declarations. |
 | `sort_imports` | `false` | Sort import statements alphabetically in groups. A group is a set of imports separated by a newline. |
+| `namespace_import_style` | `prefer_plain` | Style for namespace imports. Options: `prefer_plain` (`import "foo" as foo;`), `prefer_glob` (`import * as foo from "foo";`), `preserve`. |
 | `pow_no_space` | `false` | Suppress spaces around the power operator (`**`). |
 | `single_line_imports` | `false` | Keep single imports on a single line, even if they exceed the line length limit. |
 


crates/config/src/fmt.rs
```diff
@@ -189,6 +191,18 @@ impl MultilineFuncHeaderStyle {
...
 }
 
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum NamespaceImportStyle {
+    /// prefer plain imports: `import * as name from "source";`
+    #[default]
+    PreferPlain,
+    /// prefer glob imports: `import * as name from "source";`
+    PreferGlob,
```
onbjerg: these two have the same doc comment

onbjerg (changes requested):
please also add correct formatter tests, you are missing an original input: https://github.com/foundry-rs/foundry/tree/master/crates/fmt#testing

there are also some other failing tests (e.g. the config test).

crates/config/src/fmt.rs
quangloc99: Sorry for this :pray: . I have fixed the comment.

quangloc99:
@onbjerg About the "original.sol" file, I actually reused the current file in the `ImportDirective/` test directory https://github.com/foundry-rs/foundry/blob/master/crates/fmt/testdata/ImportDirective/original.sol

I have also fixed the default config test. Though there are other failed test but it seems to me that it was RPC-related, so I'm not sure how to resolve that yet.

Anyway please take a look again :pray: .

PR #13112, opened 2026-01-16
feat: use Osaka as default EVM hardfork

zerosnacks:
## Summary

Updates the default EVM hardfork from Prague to Osaka, following upstream revm changes.

## Changes

- Update default `evm_version` from `Prague` to `Osaka`
- Bump default solc version to 0.8.33 for Osaka support
- Update EVM tests to use `SpecId::default()` and `OpSpecId::default()` instead of hardcoding specific hardforks
- Add `ETH_OSAKA_PRECOMPILE` (EIP-7951) and `OP_ISTHMUS_PRECOMPILE` constants for precompile verification
- Add tests verifying default spec includes expected precompiles
- Add `build_eth_evm_with_extra_precompiles_london_spec` and `build_op_evm_with_extra_precompiles_bedrock_spec` tests to verify older specs exclude newer precompiles
- Update CREATE2 address test expectations for Osaka bytecode changes
- Update config and compiler test snapshots for solc 0.8.33

## Design

Tests now assert behavior of defaults rather than hardcoding versions. This ensures foundry automatically follows upstream revm when defaults change (e.g., when OpSpecId moves to Jovian).

## Testing

- All existing tests updated for new defaults
- New tests verify precompile activation/deactivation across hardforks

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "1.5.1"
+version = "1.6.0"
 edition = "2024"
 rust-version = "1.89"
 authors = ["Foundry Contributors"]
@@ -423,7 +423,8 @@ tower-http = "0.6"
 tracing = "0.1"
 tracing-subscriber = "0.3"
 url = "2"
-vergen = { version = "8", default-features = false }
+vergen = { version = "9", default-features = false, features = ["build", "cargo"] }
+vergen-git2 = "9"
 yansi = { version = "1.0", features = ["detect-tty", "detect-env"] }
 path-slash = "0.2"
 jiff = { version = "0.2", default-features = false, features = [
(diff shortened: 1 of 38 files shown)


mattsse (approved):
gg

testdata/foundry.toml
```diff
@@ -4,15 +4,18 @@ test = "./"
 optimizer = true
 optimizer_runs = 200
 via_ir = false
+evm_version = "cancun"
```
zerosnacks: Mostly working but this is odd, source of it is Vyper not supporting Osaka yet so we need to normalize conditionally

Fixing upstream in foundry-compilers

zerosnacks: Ref: https://github.com/foundry-rs/compilers/pull/345/changes

zerosnacks: Fixed in v0.19.12

Cargo.toml
```diff
@@ -274,7 +274,7 @@ foundry-primitives = { path = "crates/primitives" }
 
 # solc & compilation utilities
 foundry-block-explorers = { version = "0.22.0", default-features = false }
-foundry-compilers = { version = "0.19.10", default-features = false, features = [
+foundry-compilers = { version = "0.19.12", default-features = false, features = [
```
zerosnacks, line 277: See: https://github.com/foundry-rs/compilers/pull/345 + https://github.com/foundry-rs/compilers/pull/347

PR #13114, opened 2026-01-16
feat(cast): add --curl flag to output equivalent curl commands

gakonst:
## Summary

Adds a `--curl` flag to cast RPC commands that outputs the equivalent curl command instead of executing the RPC request.

## Motivation

This is useful when working with systems that have curl but not cast installed (e.g., Stripe's servers). Instead of manually constructing curl commands, you can generate them directly:

```bash
$ cast bn --curl -r https://eth.example.com
curl -X POST -H 'Content-Type: application/json' --data-raw '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}' 'https://eth.example.com'
```

## Changes

- Added `--curl` flag to `RpcOpts` (available to all commands using RPC options)
- Added `foundry_cli::curl` module with `generate_curl_command` helper
- Implemented `--curl` support for:
  - `cast rpc`
  - `cast block-number` (bn)
  - `cast chain-id`
  - `cast gas-price`

The flag respects configured headers and JWT authentication, including them in the generated curl command.

## Future Work

Additional commands can be updated to support `--curl` following the same pattern.


crates/cast/src/args.rs
```diff
@@ -447,6 +484,21 @@ pub async fn run_command(args: CastArgs) -> Result<()> {
...
+                let curl_cmd = generate_curl_command(
+                    url.as_ref(),
+                    "eth_gasPrice",
+                    serde_json::json!([]),
+                    config.eth_rpc_headers.as_deref(),
+                    jwt.as_deref().map(|s| s.as_ref()),
+                );
+                sh_println!("{}", curl_cmd)?;
+                return Ok(());
+            }
```
onbjerg: adding this to every command is cumbersome and error prone. i'd suggest we wrap the alloy provider or the alloy transport instead.

https://github.com/foundry-rs/foundry/blob/f613d83e3057040643a97ec1aad0e0c909163cf5/crates/cli/src/utils/mod.rs#L104-L134

PR #13115, opened 2026-01-17
chore(anvil): handle Tempo transactions with errors instead of todo!

tefyosL-sol:
Replace `todo!()` with explicit error returns and descriptive panics for unsupported Tempo transactions in Anvil, preventing node crashes on unsupported transaction types.


tefyosL-sol:
cc @onbjerg
I replaced `todo!()` with explicit errors to prevent panics on unsupported Tempo transactions.  I've kept the TODO comments as-is since the actual implementation is still pending.

tefyosL-sol:
@onbjerg Thanks for reviewing! I kept the TODO comments in the code, just replaced  `todo!()` with errors to prevent crashes. Should I adjust something or close this PR?

PR #13126, opened 2026-01-19
feat(`forge-lint`): add custom errors rule

milosdjurica:
## Motivation

Add another linting rule suggesting use of custom errors over empty reverts and usage of strings in revert/require.

## Solution

Not sure if the rule is too demanding, could make it configurable to choose which cases should be linted and which no. For example, lint only revert() , or lint only reverts and requires when using strings, etc...

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/lint/testdata/CustomErrors.sol
+++ b/crates/lint/testdata/CustomErrors.sol
@@ -9,7 +9,7 @@ error CustomErrorWithNamedArgs(uint256 x, string message);
 
 contract CustomErrors {
     // Require examples
-    function requireWithString1(uint256 a, uint256 b) public pure {
+    function requireWithString(uint256 a, uint256 b) public pure {
         require(a > 0, "Value must be greater than zero"); //~NOTE: prefer using custom errors on revert and require calls
         require(a >= 0 && a <= 100 || b == 50, "Complex condition should be linted"); //~NOTE: prefer using custom errors on revert and require calls
     }
@@ -22,7 +22,7 @@ contract CustomErrors {
     }
 
     // Custom error examples
-    function requireWithCustomError(uint256 value) public pure {
+    function customErrors(uint256 value) public pure {
         require(value > 0, CustomError());
         require(value < 100, CustomErrorWithArg(value));
         require(value > 0);


crates/lint/src/sol/gas/custom_errors.rs
```diff
@@ -0,0 +1,72 @@
...
+        if let StmtKind::Expr(expr) = &stmt.kind
+            && let ExprKind::Call(callee, args) = &expr.kind
+            && ((is_require_call(callee) && should_lint_require(args))
+                || (is_revert_call(callee) && should_lint_revert(args)))
+        {
+            self.ctx.emit(&CUSTOM_ERRORS, expr.span);
+        }
+        self.walk_stmt(stmt)
+    }
+}
```
0xrusowsky, line 24: since `EarlyLintPass` already visits the full AST, the `CustomErrorsChecker` seems useless, as it doesn't really need to hold any state. Instead, we can simplify the lint impl by directly doing:
```rs
 impl<'ast> EarlyLintPass<'ast> for CustomErrors {
      fn check_expr(&mut self, ctx: &LintContext, expr: &'ast Expr<'ast>) {
          if let ExprKind::Call(callee, args) = &expr.kind
              && ((is_require_call(callee) && should_lint_require(args))
                  || (is_revert_call(callee) && should_lint_revert(args)))
          {
              ctx.emit(&CUSTOM_ERRORS, expr.span);
          }
      }
  }
```

crates/lint/testdata/CustomErrors.sol
```diff
@@ -0,0 +1,73 @@
...
+    function testDisableShouldNotLint() public pure {
+        // forge-lint: disable-next-line(custom-errors)
+        require(true, "This should not lint");
+    }
+
+    function testDisableShouldLint() public pure {
+        // forge-lint: disable-next-line(custom-errors)
+        revert("This should not lint");
+    }
+}
```
0xrusowsky, line 41: let's consolidate the test fns in blocks to avoid having to do the numbered suffixes.

imo we only need 4 fns to check:
- revert
- require
- custom errors
- inline disable

0xrusowsky (changes requested):
great lint!

just left some minor cmnts @milosdjurica but after that we are g2g

PR #13133, opened 2026-01-19
feat(invariant): add check_interval config for faster deep runs

grandizzy:
## Problem

Invariant tests with high depth (e.g., `depth=1000`) are significantly slower than equivalent shallow runs (e.g., `100 runs × depth=10`) because the invariant function is executed via EVM on **every single call**, taking ~86% of total time.

### Profiling Results

| Operation | Time at 500 calls | Time at 1000 calls | % of Total |
|-----------|------------------|-------------------|-----------|
| execute_tx | 105ms | 204ms | ~4% |
| commit | 1.8ms | 3.7ms | <1% |
| clone | 0.2ms | 0.4ms | <1% |
| collect_data | 267ms | 542ms | ~10% |
| **can_continue** | **2.76s** | **5.26s** | **~86%** |

## Solution

Add `check_interval` config option to control how often invariants are asserted:

| Value | Behavior |
|-------|----------|
| `0` | Only assert on the last call of each run (fastest, but may miss exact breaking call) |
| `1` (default) | Assert after every call (current behavior, most precise) |
| `N` | Assert every N calls AND always on the last call |

## Performance

With `check_interval = 10`:

| Configuration | check_interval=1 | check_interval=10 | Speedup |
|--------------|------------------|-------------------|--------|
| 1×1000 | **15.09s** | **4.17s** | **3.6x faster** |

Deep runs with `check_interval=10` are now **faster than shallow runs** while still catching invariant violations.

## ⚠️ Warning

Using `check_interval > 1` may cause **failures to be missed** if the invariant is violated and then restored between checks. To mitigate this:

1. **Use `fail_on_revert = true`** to catch unexpected reverts immediately
2. **Assert invariants as modifiers on each handler call** as a best practice—this ensures critical properties are validated on every state change regardless of `check_interval`

If a violation is detected, re-run with `check_interval = 1` to pinpoint the exact breaking call.

## Usage

```toml
[invariant]
depth = 1000
check_interval = 10  # Assert every 10 calls + always on last call
```

Or for maximum speed (only check at the end):

```toml
[invariant]
depth = 1000
check_interval = 0  # Only assert on the last call
```

## Files Changed

- `crates/config/src/invariant.rs` - Added `check_interval: u32` field (default: 1)
- `crates/evm/evm/src/executors/invariant/mod.rs` - Skip invariant checks based on interval
- `crates/evm/evm/src/executors/invariant/result.rs` - Made `RichInvariantResults::new` pub(crate)

### DIFF
--- a/crates/config/src/invariant.rs
+++ b/crates/config/src/invariant.rs
@@ -41,6 +41,14 @@ pub struct InvariantConfig {
     pub max_time_delay: Option<u32>,
     /// Maximum number of blocks elapsed between generated txs.
     pub max_block_delay: Option<u32>,
+    /// Number of calls to execute between invariant assertions.
+    ///
+    /// - `0`: Only assert on the last call of each run (fastest, but may miss exact breaking call)
+    /// - `1` (default): Assert after every call (current behavior, most precise)
+    /// - `N`: Assert every N calls AND always on the last call
+    ///
+    /// Example: `check_interval = 10` means assert after calls 10, 20, 30, ... and the last call.
+    pub check_interval: u32,
 }
 
 impl Default for InvariantConfig {
@@ -61,6 +69,7 @@ impl Default for InvariantConfig {
             show_solidity: false,
             max_time_delay: None,
             max_block_delay: None,
+            check_interval: 1,
         }
     }
 }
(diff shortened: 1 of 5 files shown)


DaniPopes (approved):

7zkm7b8gw9-web (commented):
K

wyattloveslife-create (commented):
Let's get this fixed all ave way

PR #13141, opened 2026-01-19
fix(corpus): handle empty inputs in abi_mutate

forkfury:
Add early return for functions with no inputs to prevent panic when accessing prev_inputs[0]. Synchronize arg_mutation_rounds calculation with round_arg_idx creation to ensure consistent behavior when inputs_len <= 1.


emmajam:

PR #13149, opened 2026-01-20
fix(script): set msg.sender from TURNKEY_ADDRESS when using --turnkey

### LINKED ISSUE #12916: bug(`forge script`): msg.sender is not set to `TURNKEY_ADDRESS` when using `--turnkey`
### Component

Forge

- [x] Foundry
- [x] Foundryup

foundryup: 1.3.0

forge script --turnkey

### Operating System

None

### Describe the bug

forge script currently allows using --turnkey by passing the following variables:
`TURNKEY_API_PRIVATE_KEY`
`TURNKEY_ORGANIZATION_ID`
`TURNKEY_ADDRESS`

currently, when we run `forge script --turnkey`, the `msg.sender` used in the scripts is the default one (0x1804c8AB1F12E6bbf3894d4083f33e07309d1f38)

instead `msg.sender` should automatically change to the `TURNKEY_ADDRESS`. currently to achieve this we have to forcefully also do `forge script --turnkey --sender TURNKEY_ADDRESS ` which seems redundant

gakonst:
## Summary

When using `forge script --turnkey`, the script's `msg.sender` now automatically uses the address from the `TURNKEY_ADDRESS` environment variable, eliminating the need to redundantly specify `--sender TURNKEY_ADDRESS`.

## Problem

As reported in #12916, when running `forge script --turnkey`, the `msg.sender` used in scripts defaults to the standard address (`0x1804c8AB1F12E6bbf3894d4083f33e07309d1f38`) instead of the Turnkey address. Users had to work around this by explicitly passing `--sender TURNKEY_ADDRESS`, which is redundant since the address is already provided via the environment variable.

## Solution

1. Added `turnkey_address()` method to `MultiWalletOpts` that returns the parsed `TURNKEY_ADDRESS` when the `--turnkey` flag is set
2. Updated `maybe_load_private_key()` in `ScriptArgs` to check for Turnkey address first, similar to how it handles single private keys

This ensures that when `--turnkey` is used, the sender is automatically derived from `TURNKEY_ADDRESS` without requiring the additional `--sender` flag.

## Changes

- `crates/wallets/src/wallet_multi/mod.rs`: Added `turnkey_address()` method
- `crates/script/src/lib.rs`: Updated `maybe_load_private_key()` to prioritize Turnkey address

Fixes #12916

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1077,7 +1077,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -1101,7 +1101,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


zerosnacks:
> Are the cargo updates needed for this PR?

Not needed, undid the changes

grandizzy (approved):
makes sense

zjesko:
thank you!! @gakonst

PR #13155, opened 2026-01-20
fix(cli): add `--no-proxy` to disable `reqwest` proxying to prevent crash on macOS in sandboxed environments

### LINKED ISSUE #12733: bug(forge): crash on macOS in sandboxed environment (SCDynamicStore NULL)
### Component

Forge

- [x] Foundry
- [x] Foundryup

1.5.0-stable

1.4.0

`forge test`

### Operating System

macOS (Apple Silicon)

### Describe the bug

Running `forge test` in a sandboxed environment causes a panic when Foundry attempts to read macOS system proxy configuration.

**Error message:**
```
The application panicked (crashed).
Message:  Attempted to create a NULL object.
Location: /Users/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/system-configuration-0.6.1/src/dynamic_store.rs:154

This is a bug. Consider reporting it at https://github.com/foundry-rs/foundry
```

**Relevant backtrace:**
```
 8: system_configuration::dynamic_store::SCDynamicStoreBuilder<T>::build
 9: hyper_util::client::proxy::matcher::mac::with_system
10: hyper_util::client::proxy::matcher::Builder::from_system
11: hyper_util::client::proxy::matcher::Matcher::from_system
12: reqwest::async_impl::client::ClientBuilder::build
13: foundry_common::provider::runtime_transport::RuntimeTransport::reqwest_client
14: foundry_common::selectors::OpenChainClient::new
15: foundry_evm_traces::identifier::signatures::SignaturesIdentifier::new_with
16: foundry_evm_traces::identifier::signatures::SignaturesIdentifier::from_config
17: forge::cmd::test::TestArgs::run_tests_inner
18: forge::cmd::test::TestArgs::run_tests
19: forge::cmd::test::TestArgs::compile_and_run
```

**Root cause:**

When creating the `OpenChainClient` for signature lookups, Foundry builds a `reqwest` HTTP client. On macOS, this triggers reading system proxy settings via `SCDynamicStore`. In sandboxed environments (e.g., Cursor IDE sandbox, App Sandbox, or restricted shells), the `SCDynamicStore` API returns NULL, which causes the `system-configuration` crate to panic.

**Reproduction steps:**

1. Run `forge test` in a sandboxed macOS environment (e.g., within an IDE that sandboxes terminal commands)
2. Observe the crash

**Workarounds:**

- Use `--offline` flag to disable signature lookups
- Run outside the sandbox

Note: Lower verbosity does **not** help - the crash occurs during initialization regardless of verbosity level.

## Suggested fix

The code should gracefully handle the case where `SCDynamicStore` returns NULL instead of panicking. This could be done by:

1. Catching the NULL case in `foundry_common` before building the reqwest client
2. Falling back to no proxy when system proxy detection fails
3. Or using `reqwest`'s `no_proxy()` option when in environments where system proxy detection might fail

This would make Foundry more robust when running in sandboxed/restricted environments on macOS.

gakonst:
## Summary

Add a new `--no-proxy` (alias: `--disable-proxy`) CLI flag to disable automatic proxy detection, preventing crashes on macOS in sandboxed environments.

## Problem

On macOS, when `forge test` or other commands are run in a sandboxed environment (e.g., Cursor IDE sandbox, App Sandbox, or restricted shells), Foundry crashes with:

```
The application panicked (crashed).
Message:  Attempted to create a NULL object.
Location: .../system-configuration-0.6.1/src/dynamic_store.rs:154
```

This happens because reqwest automatically reads system proxy settings via `SCDynamicStore`. In sandboxed environments, this API returns NULL, which causes the `system-configuration` crate to panic.

## Solution

Add an opt-in `--no-proxy` flag (with `--disable-proxy` alias) that disables automatic proxy detection when specified. This preserves existing behavior for users who need proxy support while giving sandboxed users an escape hatch.

**Changes:**
- `crates/config/src/lib.rs` - Add `eth_rpc_no_proxy` config option
- `crates/cli/src/opts/rpc.rs` - Add `--no-proxy` / `--disable-proxy` CLI flag
- `crates/common/src/provider/runtime_transport.rs` - Add `no_proxy` option to builder
- `crates/common/src/provider/mod.rs` - Wire up the option through `ProviderBuilder`
- `crates/forge/tests/cli/config.rs` - Update tests

## Usage

```bash
# Use the flag in sandboxed environments
forge test --no-proxy
# or
forge test --disable-proxy

# Or set in foundry.toml
[profile.default]
eth_rpc_no_proxy = true
```

## Trade-offs

When `--no-proxy` is enabled:
- Environment variables (`HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY`, `NO_PROXY`) are ignored
- macOS System Preferences proxy settings are ignored  
- Windows Registry proxy settings are ignored

This is opt-in, so existing users are unaffected.

Fixes #12733

### DIFF
--- a/crates/cli/src/opts/rpc.rs
+++ b/crates/cli/src/opts/rpc.rs
@@ -30,6 +30,14 @@ pub struct RpcOpts {
     #[arg(short = 'k', long = "insecure", default_value = "false")]
     pub accept_invalid_certs: bool,
 
+    /// Disable automatic proxy detection.
+    ///
+    /// Use this in sandboxed environments (e.g., Cursor IDE sandbox, macOS App Sandbox) where
+    /// system proxy detection causes crashes. When enabled, HTTP_PROXY/HTTPS_PROXY environment
+    /// variables and system proxy settings will be ignored.
+    #[arg(long = "no-proxy", alias = "disable-proxy", default_value = "false")]
+    pub no_proxy: bool,
+
     /// Use the Flashbots RPC URL with fast mode (<https://rpc.flashbots.net/fast>).
     ///
     /// This shares the transaction privately with all registered builders.
@@ -117,6 +125,9 @@ impl RpcOpts {
         }
         if self.accept_invalid_certs {
             dict.insert("eth_rpc_accept_invalid_certs".into(), true.into());
+        }
+        if self.no_proxy {
+            dict.insert("eth_rpc_no_proxy".into(), true.into());
         }
         dict
     }
--- a/crates/common/src/provider/mod.rs
+++ b/crates/common/src/provider/mod.rs
@@ -100,6 +100,8 @@ pub struct ProviderBuilder {
     is_local: bool,
     /// Whether to accept invalid certificates.
     accept_invalid_certs: bool,
+    /// Whether to disable automatic proxy detection.
+    no_proxy: bool,
     /// Whether to output curl commands instead of making requests.
     curl_mode: bool,
 }
@@ -152,6 +154,7 @@ impl ProviderBuilder {
             headers: vec![],
             is_local,
             accept_invalid_certs: false,
+            no_proxy: false,
             curl_mode: false,
         }
     }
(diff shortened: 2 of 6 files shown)


zerosnacks:
Hi @uri-99, my suggested approach would be to add a `--no-proxy` flag one can pass to disable system proxying on `reqwest`. In my opinion this is preferred over outright disabling as there are legitimate cases (and user expectations) where the current default setting is appropriate.

grandizzy (approved):
lgtm

PR #13156, opened 2026-01-20
feat(forge script): display contract name and function in tx output

### LINKED ISSUE #12739: Display more tx metadata in script output
### Component

Forge

we currently display:

> ##### chainid
> ✅  [Success] Hash: <>
> Block: ...
> Paid: 0.001080630...

for script results

it would be great if we can also add
* function if we have it (we should because we write it to the report)
* contract name if we have it

displayed here:

https://github.com/foundry-rs/foundry/blob/efdf0ff81807fa8dc426cbf1316503945aa5544e/crates/script/src/receipts.rs#L91-L92

but we should have access to the `ScriptSequence` here

### Additional context

_No response_

gakonst:
## Summary

When running `forge script`, the transaction output now shows:
- **Contract name** for the called contract
- **Function signature** that was called

This information is already available in the `ScriptSequence` transactions and is written to the `run-latest.json` report, so we now surface it in the terminal output as well.

## Example Output

```
##### Ethereum
✅  [Success] Hash: 0x...
Contract: MyContract
Function: initialize(address,uint256)
Block: 12345
Paid: 0.001 ETH (50000 gas * 20 gwei)
```

For JSON output (`--json`), the `contract_name` and `function` fields are now included when available.

## Changes

- Modified `format_receipt()` in `crates/script/src/receipts.rs` to accept an optional `ScriptSequence` parameter
- Look up the transaction by hash in the sequence to extract contract name and function metadata
- Updated callers in `progress.rs` to pass the sequence

## Closes

Closes #12739

## Checklist
- [x] Added Tests (existing tests cover this path)
- [x] Added Documentation (inline in commit)
- [ ] Breaking changes (none - purely additive)

### DIFF
--- a/crates/script/src/receipts.rs
+++ b/crates/script/src/receipts.rs
@@ -212,11 +212,7 @@ mod tests {
             },
         }))
         .unwrap();
-        ScriptSequence {
-            transactions: VecDeque::from([tx]),
-            chain: 1,
-            ..Default::default()
-        }
+        ScriptSequence { transactions: VecDeque::from([tx]), chain: 1, ..Default::default() }
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


grandizzy (approved):
nice!

PR #13161, opened 2026-01-20
fix(test): remove cast binary dependency from forge verify-bytecode tests

### LINKED ISSUE #8856: verify-bytecode tests depend on cast in forge crate
Can't run cargo t / cargo nextest run on -p forge because verify-bytecode tests depend on the cast binary which may not be available

gakonst:
Fixes #8856

## Problem

The verify-bytecode tests in the forge crate use `cast_fuse()` to shell out to the `cast source --flatten` command to fetch source code from Etherscan. This means `cargo test -p forge` fails if the cast binary isn't built and available in PATH.

## Solution

This PR:
- Adds a new `fetch_etherscan_source_flattened()` helper function in `foundry-test-utils` that uses the `foundry-common` and `foundry-block-explorers` libraries directly (same underlying implementation as `cast source --flatten`)
- Updates the verify-bytecode tests to use this helper instead of shelling out to cast
- Moves `foundry-block-explorers` from dev-dependencies to regular dependencies in the test-utils crate

Now `cargo test -p forge` can run verify-bytecode tests without requiring the cast binary to be built first.

## Note

The other `cast_fuse()` usages in forge tests (`script.rs` and `test_cmd/mod.rs`) are intentionally left unchanged as they test cast functionality in integration tests, which is appropriate. Only the verify-bytecode tests were using cast as a utility when forge should be self-contained.

### DIFF
--- a/crates/forge/tests/cli/verify_bytecode.rs
+++ b/crates/forge/tests/cli/verify_bytecode.rs
@@ -20,12 +20,13 @@ async fn test_verify_bytecode(
     verifier: &str,
     verifier_url: &str,
     expected_matches: (&str, &str),
+    chain: Chain,
 ) {
     let etherscan_key = next_etherscan_api_key();
     let rpc_url = next_http_archive_rpc_url();
 
     // fetch and flatten source code using the library directly
-    let source_code = fetch_etherscan_source_flattened(addr, &etherscan_key, Chain::mainnet())
+    let source_code = fetch_etherscan_source_flattened(addr, &etherscan_key, chain)
         .await
         .expect("failed to fetch source code from etherscan");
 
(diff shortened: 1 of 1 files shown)


crates/forge/tests/cli/verify_bytecode.rs
```diff
@@ -21,13 +24,10 @@ fn test_verify_bytecode(
...
 
-    // fetch and flatten source code
-    let source_code = cmd
-        .cast_fuse()
-        .args(["source", addr, "--flatten", "--etherscan-api-key", &etherscan_key])
-        .assert_success()
-        .get_output()
-        .stdout_lossy();
+    // fetch and flatten source code using the library directly
+    let source_code = fetch_etherscan_source_flattened(addr, &etherscan_key, Chain::mainnet())
```
zerosnacks: We should pass the chain in here

zerosnacks:
I think this is correct @DaniPopes but when time permits please have a look (ref: https://github.com/foundry-rs/foundry/issues/8856)

PR #13162, opened 2026-01-20
fix(config): extract test and script paths from ProjectPathsConfig in _with_root()

### LINKED ISSUE #10300: leaky interface: _with_root
### Component

Forge

- [x] Foundry
- [ ] Foundryup

commit hash: 256cc50331d8a00b86c8e1f18ca092a66e220da5

irrelevant

forge build entrypoint

### Operating System

None

### Describe the bug

I am building on top of `forge build` and I couldn't figure out why I can't change the default `src` directory of the foundry project.

It seems that here: https://github.com/foundry-rs/foundry/blob/2564718bbbddb59cd07fc3b9ffc775dff548c558/crates/config/src/lib.rs#L1625

It will be replaced with the `src` default in the foundry compilers codepath that is entered [here](https://github.com/foundry-rs/foundry/blob/2564718bbbddb59cd07fc3b9ffc775dff548c558/crates/config/src/lib.rs#L1627). 

That seems like a weird leaky abstraction between the two crates, as I would assume that `ProjectPathsConfig::builder()` should be instantiated with the defaults that are define in the Default implementation of the config, and not have two sources of truth for the default source directory of the smart contracts.

I may be missing something though, so would love to see if I should be using this in a different way. My use-case is basically to run the `BuildArgs` command (by either calling `run()` or replicating the code inside `run`, but with a different set of default directories for `src`, `,test`, etc. 

It's worth noting that it's only `src` and `scripts` that are not respected, but `test` will respect the definition in the Default implementation of the Config.

gakonst:
## Summary

Fixes #10300

## Problem

`Config::_with_root()` has an inconsistency in how it initializes path fields:

- `src` and `out` are extracted from `ProjectPathsConfig` (auto-detected by foundry-compilers)
- `test` and `script` fall through to `Self::default()` (hardcoded defaults)

This creates a leaky abstraction with two sources of truth for path defaults.

## Solution

Extract `test` and `script` from `ProjectPathsConfig` the same way `src` is extracted:

```rust
Self {
    root: paths.root,
    src: paths.sources.file_name().unwrap().into(),
    test: paths.tests.file_name().unwrap().into(),      // NEW
    script: paths.scripts.file_name().unwrap().into(),  // NEW
    out: artifacts.clone(),
    // ...
}
```

## Why this matters

This ensures **architectural consistency** - all path defaults now come from a single source (`ProjectPathsConfig`). If foundry-compilers ever adds auto-detection for test/script directories (like it does for src→contracts), Config will automatically inherit that behavior.

## Testing Note

A regression test is not included because `ProjectPathsConfig` and `Config::default()` currently use identical values for test/script (`"test"` and `"script"`), making it impossible to distinguish implementations via value comparison. The fix is self-documenting with an inline comment.

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -1853,6 +1853,8 @@ impl Config {
         // autodetect paths
         let paths = ProjectPathsConfig::builder().build_with_root::<()>(root);
         let artifacts: PathBuf = paths.artifacts.file_name().unwrap().into();
+        // All path fields (src, test, script, out) must be extracted from ProjectPathsConfig
+        // to maintain a single source of truth. See: https://github.com/foundry-rs/foundry/issues/10300
         Self {
             root: paths.root,
             src: paths.sources.file_name().unwrap().into(),
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -1856,6 +1856,8 @@ impl Config {
         Self {
             root: paths.root,
             src: paths.sources.file_name().unwrap().into(),
+            test: paths.tests.file_name().unwrap().into(),
+            script: paths.scripts.file_name().unwrap().into(),
             out: artifacts.clone(),
             libs: paths.libraries.into_iter().map(|lib| lib.file_name().unwrap().into()).collect(),
             remappings: paths


zerosnacks:
@odyslam I think this is the change you are looking for but I am having trouble testing it in isolation. Can you let me know if this resolves https://github.com/foundry-rs/foundry/issues/10300? If so we can include this in v1.6

emmajam:

PR #13166, opened 2026-01-21
fix(fmt): keep struct field access on same line after named args call

### LINKED ISSUE #12399: bug(fmt): unexpected line breaks when accessing struct fields
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.4.4-v1.4.4 Commit SHA: 05794498bf47257b144e2e2789a1d5bf8566be0e

foundryup: 1.3.0

forge fmt

### Operating System

None

### Describe the bug

Minimal repro to illustrate the case
```solidity

contract A {
    function a() external {
        bytes32 guid = _lzSend({
            _dstEid: dstEid,
            _message: message,
            _options: OptionsBuilder.newOptions().addExecutorLzReceiveOption({_gas: gasLimit, _value: 0}),
            _fee: MessagingFee({nativeFee: msg.value, lzTokenFee: 0}),
            _refundAddress: msg.sender
        }).guid;
    }

    function b() external view returns (uint256) {
        return _quote({
            _dstEid: dstEid,
            _message: message,
            _options: OptionsBuilder.newOptions().addExecutorLzReceiveOption({_gas: gasLimit, _value: 0}),
            _payInLzToken: false
        }).nativeFee;
    }
}
```

Gets formatted as
```solidity

contract A {
    function a() external {
        bytes32 guid =
            _lzSend({
            _dstEid: dstEid,
            _message: message,
            _options: OptionsBuilder.newOptions().addExecutorLzReceiveOption({_gas: gasLimit, _value: 0}),
            _fee: MessagingFee({nativeFee: msg.value, lzTokenFee: 0}),
            _refundAddress: msg.sender
        })
        .guid;
    }

    function b() external view returns (uint256) {
        return _quote({
            _dstEid: dstEid,
            _message: message,
            _options: OptionsBuilder.newOptions().addExecutorLzReceiveOption({_gas: gasLimit, _value: 0}),
            _payInLzToken: false
        })
        .nativeFee;
    }
}
```

`.structField` line breaks looks weird, and the extra line break in the case with variable assigning doesn't look right.

Using the following fmt config
```toml
[fmt]
line_length = 120
multiline_func_header = 'all'
number_underscore = 'thousands'
sort_imports = true
prefer_compact = 'none'
```

gakonst:
## Motivation

Closes #12399

When accessing a struct field after a function call with named arguments (e.g., `_lzSend({...}).guid`), the formatter was incorrectly placing the `.field` on a new line with weird indentation.

**Before:**
```solidity
bytes32 guid =
    _lzSend({
    _dstEid: dstEid,
    ...
})
.guid;  // <-- unwanted line break
```

**After:**
```solidity
bytes32 guid =
    _lzSend({
    _dstEid: dstEid,
    ...
}).guid;  // field stays on same line
```

## Solution

Added detection for calls with named arguments in the member expression handling. When the preceding expression is a call using named arguments (the `{key: value}` syntax), we skip the `zerobreak()` that was causing the unwanted line break.

This specifically targets the pattern from the issue report while preserving existing behavior for:
- Regular method chains (`obj.method().anotherMethod()`)
- Calls with positional arguments

## PR Checklist

- [x] Added Tests (`testdata/StructFieldAccess/`)
- [x] Added Documentation (inline comments)
- [ ] Breaking changes (none)

### DIFF
--- a/crates/fmt/src/state/sol.rs
+++ b/crates/fmt/src/state/sol.rs
@@ -1701,7 +1701,8 @@ impl<'ast> State<'_, 'ast> {
 
             // Determine if this chain will add its own indentation
             let chain_has_indent = is_call_chain(&child_expr.kind, true)
-                || !(no_cmnt_or_mixed || matches!(&child_expr.kind, ast::ExprKind::CallOptions(..)))
+                || !(no_cmnt_or_mixed
+                    || matches!(&child_expr.kind, ast::ExprKind::CallOptions(..)))
                 || !callee_fits_line
                 || (member_depth(0, child_expr) >= 2
                     && (!total_fits_line || member_or_args.has_comments()));


gakonst:
Hey @gakonst 👋

Could you add an extra test case for **chained calls with named args** to ensure the second chained call and final `.guid` are properly formatted?

Something like this in `original.sol`:

```solidity
// Chained calls with named args
function e() external {
    bytes32 guid = firstCall({
        param1: value1,
        param2: value2
    }).secondCall({
        arg1: val1,
        arg2: val2
    }).guid;
}
```

And the corresponding expected output in `fmt.sol`:

```solidity
// Chained calls with named args
function e() external {
    bytes32 guid = firstCall({
        param1: value1,
        param2: value2
    }).secondCall({
        arg1: val1,
        arg2: val2
    }).guid;
}
```

This would verify that when you have multiple chained calls each using named args, the fix works for all of them (not just the first call), and that the final `.guid` field access stays on the same line as the closing `})`.

PR #13169, opened 2026-01-21
fix(anvil): clean up in-memory indexes when disk cache is corrupted

echowandere:
When `disk_cache.read()` fails (file corrupted or missing), the old code just returned None but left stale entries in `on_disk_states`, `oldest_on_disk`, and `present` maps. This caused the in-memory index to drift from actual disk state.

Detect this case and clean up all three indexes, plus log a warning so it's easier to debug if it happens. make the system handle cache files go bad.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/storage.rs
+++ b/crates/anvil/src/eth/backend/mem/storage.rs
@@ -173,35 +173,41 @@ impl InMemoryBlockStates {
 
     /// Returns on-disk state for the given `hash` if present
     pub fn get_on_disk_state(&mut self, hash: &B256) -> Option<&StateDb> {
-        if let Some(state) = self.on_disk_states.get_mut(hash) {
-            match self.disk_cache.read(*hash) {
-                Some(cached) => {
+        // Check if hash exists in on_disk_states index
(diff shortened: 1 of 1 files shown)


onbjerg (changes requested):
this seems very expensive for a cache miss, have you experienced this locally and if so, can you provide a reproducible example?

echowandere:
> this seems very expensive for a cache miss, have you experienced this locally and if so, can you provide a reproducible example?

Thanks for the review! Just to clarify - this isn't the normal cache miss path.

I've added two tests to show the difference:

1. `normal_cache_miss_does_not_cleanup` - normal miss is O(1):
   - Hash not in index → `contains_key` returns false → immediate return
   - No cleanup, no iteration

2. `cleans_up_corrupted_disk_cache_index` - cleanup only on corruption:
   - Index says "state on disk" but disk_cache.read() fails
   - Only then we clean up on_disk_states and oldest_on_disk
   - This fixes the data integrity issue in #21150

The VecDeque iterations are O(n) but only run when cache files are actually corrupted/missing (rare error path), and capped at 3600 entries max. Normal operation stays fast.

emmajam:

PR #12616, opened 2025-11-20
feat(forge): warp and roll on invariant tx

grandizzy:
## Motivation
- closes https://github.com/foundry-rs/foundry/issues/12332
- introduce `max_time_delay` and `max_block_delay` invariant configs to fuzz block timestamp and block number on consecutive tx
- display in solidity counterexample as
```Solidity
		vm.warp(block.timestamp + 446165);
		vm.roll(block.number + 28222);
		vm.prank(0x000000000000000000000000000000000000024c);
		Counter(0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f).increment();
```
and in regular counterexample as
```bash
		sender=0x000000000000000000000000000000000000024c addr=[src/Counter.sol:Counter]0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f warp=446165 roll=28222 calldata=increment() args=[]
```

CC @rappie

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #13177, opened 2026-01-22
feat(fuzz): enhance corpus mutation with all-call strategy and msg.value support

grandizzy:
## Motivation

This PR adds enhancements to the corpus-based invariant fuzzer:

1. **Unified mutation strategies**: Renamed `Prefix`/`Suffix`/`Abi` to `GenPrefix`/`GenSuffix`/`GenMutate` for clarity. `GenMutate` now mutates a random number of calls (1 to all) using shuffled indices, similar to how `GenPrefix` generates a random prefix of new calls.

2. **Random `msg.value` for payable functions**: Currently, Foundry invariant fuzz testing does not support transactions with `value > 0`, which makes it unable to find certain bug sequences. This adds automatic `msg.value` generation for payable functions.

3. **`max_deal` config option**: New invariant config to fund senders before payable function calls, enabling `msg.value > 0` transactions without requiring manual `vm.deal()` in setUp.

### Mutation Strategy Unification
- Renamed mutation types: `Prefix` -> `GenPrefix`, `Suffix` -> `GenSuffix`, `Abi` -> `GenMutate`
- `GenMutate` now picks a random count (1 to all) of calls to mutate, using Fisher-Yates shuffle to select which calls
- More gradual mutation than the previous 30% all / 70% single approach

### msg.value Support (based on PR #8644)
- **Initial call generation** (`fuzz_contract_with_calldata`): For payable functions, generates random value ~15% of the time
- **Corpus mutation** (`abi_mutate`): During ABI mutation, mutates sender (15%) and `msg.value` (15% for payable functions)
- **Display support**: Shows value in both regular output (`value=X`) and Solidity output (`{value: X}`) formats

### max_deal Config Option

New `[invariant]` config option to automatically fund senders:

```toml
[invariant]
max_deal = 1000000000000000000  # 1 ETH in wei
```

**Behavior:**
- **If `max_deal` is NOT configured**: Value is generated for payable functions but falls back to 0 if sender has insufficient balance
- **If `max_deal` IS configured**: Random deal amount (0 to max_deal) is generated and applied before each payable call, enabling `msg.value > 0` transactions to succeed

**Implementation details:**
- Deal is only applied when the function is payable (has `value > 0`)
- If balance is still insufficient after deal, value falls back to 0
- Counterexamples display deal as:
  - Regular: `deal=X`
  - Solidity: `vm.deal(sender, sender.balance + X);`

### Value Generation Strategy

Biased towards smaller values to avoid sender balance issues:
- 85% chance: no value (None)
- 10% chance: small values (0-1000 wei)
- 4% chance: medium values (up to 0.001 ETH)
- 1% chance: larger values (up to 1 ETH)

## Changes

- Rename mutation types to `GenPrefix`/`GenSuffix`/`GenMutate`
- Add `deal: Option<U256>` field to `BasicTxDetails` and `BaseCounterExample`
- Add `max_deal: Option<u64>` to `InvariantConfig`
- Add `value: Option<U256>` field to `CallDetails` and `BaseCounterExample`
- Add shared value generation helpers: `fuzz_msg_value()` and `generate_msg_value()`
- Add sender mutation (15% probability) using addresses from dictionary
- Update `execute_tx` to apply deal before call and pass value to `call_raw`
- Update display formatters to show deal and value in sequence output
- Add backwards-compatible serde support for corpus files
- Add CLI tests for max_deal functionality

## Testing

Tested against the Daedaluzz maze invariant test suite where this change helped break additional invariants by ensuring consistent parameter patterns across the entire call sequence.

## Credits

Based on https://github.com/foundry-rs/foundry/pull/8644 by @QiuhaoLi

Co-authored-by: QiuhaoLi <qiuhaoli@outlook.com>
Co-authored-by: Amp <amp@ampcode.com>


grandizzy:
@0xalpharush mind to quickly check this? thank you!

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -687,7 +698,26 @@ impl WorkerCorpus {
         test_runner: &mut TestRunner,
         fuzz_state: &EvmFuzzState,
     ) -> Result<()> {
-        // let rng = test_runner.rng();
+        // Mutate sender with 15% probability using addresses from dictionary.
+        if test_runner.rng().random_ratio(15, 100) {
```
0xalpharush, line 709: Should this also check targeted/excluded senders as was done here https://github.com/foundry-rs/foundry/pull/13090?

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -560,12 +560,23 @@ impl WorkerCorpus {
...
-                    let idx = rng.random_range(0..new_seq.len());
-                    let tx = new_seq.get_mut(idx).unwrap();
-                    if let (_, Some(function)) = targets.fuzzed_artifacts(tx) {
-                        // TODO: add call_value to call details and mutate it as well as sender some
-                        // of the time.
-                        if !function.inputs.is_empty() {
+                    // 30% chance to mutate ALL calls in the sequence.
+                    // This helps break multi-constraint bugs where any call could hit the target.
+                    if rng.random_range(0..10) < 3 {
+                        for tx in &mut new_seq {
```
0xalpharush: The `Prefix` mutation is essentially this but a subset and generation instead of mutation. Maybe we should have `GenPrefix` and `GenMutate` and mutate up to every element, but not always every element

0xalpharush:
For the call value, is there any reason it can't be much larger occasionally and we set the balance in the DB, minting the ETH effecitvely?

crates/evm/evm/src/executors/corpus.rs
grandizzy: good point, will merge with prefix mutations

grandizzy:
> For the call value, is there any reason it can't be much larger occasionally and we set the balance in the DB, minting the ETH effecitvely?

I wanted to avoid minting / setting balance directly in db and let tester control that by using the deal cheatcode as could be confusing in a scenario with sender that starts with a certain balance and expected to go down during tests but ends up with a higher one . Lmk if this makes sense

0xalpharush:
> I wanted to avoid minting / setting balance directly in db and let tester control that by using the deal cheatcode as could be confusing in a scenario with sender that starts with a certain balance and expected to go down during tests but ends up with a higher one . Lmk if this makes sense

That makes sense. Maybe like how you added warp and roll (https://github.com/foundry-rs/foundry/pull/12616), we can also add a deal before txs to increase the balance. And the mutator should be able to mutate these calls as if it were an ABI encoded to the respective signature

crates/evm/evm/src/executors/corpus.rs
grandizzy: added in https://github.com/foundry-rs/foundry/pull/13177/changes/d45219737e9e2650b3fa849dde79ffd2febf46f2

crates/evm/evm/src/executors/corpus.rs
grandizzy: pushed https://github.com/foundry-rs/foundry/pull/13177/changes/40fd0450cdf9470b8c83a99b8936adfbd412f444

grandizzy:
> > I wanted to avoid minting / setting balance directly in db and let tester control that by using the deal cheatcode as could be confusing in a scenario with sender that starts with a certain balance and expected to go down during tests but ends up with a higher one . Lmk if this makes sense
> 
> That makes sense. Maybe like how you added warp and roll (#12616), we can also add a deal before txs to increase the balance. And the mutator should be able to mutate these calls as if it were an ABI encoded to the respective signature

added max_deal config and vm.deal in counterexample with https://github.com/foundry-rs/foundry/pull/13177/changes/6ce3a1e0e38b1e9d6f6210c6e53a6131362fb41f

grandizzy:
@0xalpharush this is rfr, please recheck. thank you!

crates/config/src/invariant.rs
```diff
@@ -69,6 +71,7 @@ impl Default for InvariantConfig {
             show_solidity: false,
             max_time_delay: None,
             max_block_delay: None,
+            max_deal: None,
```
0xalpharush, line 74: Not necessarily in this PR, but maybe we should provides `Some(..)` defaults for `max_time_delay`, `max_block_delay`, and `max_deal`.

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -560,13 +566,24 @@ impl WorkerCorpus {
...
-                    let tx = new_seq.get_mut(idx).unwrap();
-                    if let (_, Some(function)) = targets.fuzzed_artifacts(tx) {
-                        // TODO: add call_value to call details and mutate it as well as sender some
-                        // of the time.
-                        if !function.inputs.is_empty() {
-                            self.abi_mutate(tx, function, test_runner, fuzz_state)?;
+                    let len = new_seq.len();
+                    // Mutate a random number of calls (1 to all), similar to how GenPrefix
+                    // generates a random number of new calls.
+                    let n_to_mutate = rng.random_range(1..=len);
```
0xalpharush, line 572: It would be nice if multiple args of the same tx can be mutated. I think right now, only one per tx can mutated even if `n_to_mutate` > 1

crates/evm/evm/src/executors/corpus.rs
0xalpharush: I think the name `GenMutate` was a mistake on my part and the old name `Abi` was better. The implementation is fine though.

 What I was suggesting was to add two new mutators,  `MutatePrefix` and `MutateSuffix`, where gen means call `new_tx` and mutate means use a tx in the seq and mutate it with `abi_mutate` (optionally you can have even an identity one which clones existing txs) . Here is a [patch](https://github.com/0xalpharush/foundry/commit/1505aba6ef02d559580bfdc7dc4d6ce5d1ded270) to reduce confusion. It also fixes the repeat mutation to insert instead of splice (which overwrites) as well as adds a swap and delete mutation.

grandizzy:
@0xalpharush thank you, going to address your comments asap!

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -1124,7 +1130,24 @@ pub(crate) fn execute_tx(executor: &mut Executor, tx: &BasicTxDetails) -> Result
...
+    let value = if requested_value.is_zero() {
+        U256::ZERO
+    } else {
+        // Apply deal (increase sender balance) if specified.
+        if let Some(deal) = tx.deal {
+            let current_balance = executor.get_balance(tx.sender)?;
+            executor.set_balance(tx.sender, current_balance + deal)?;
+        }
+
+        // Only use value if sender has sufficient balance (after deal), otherwise fall back to 0.
```
0xalpharush, line 1145: Maybe just do `sender_balance.min(requested_value)` instead of sending zero?

0xalpharush:
I would prefer holding off on "Unified mutation strategies", finalizing and merging https://github.com/foundry-rs/foundry/pull/12587, and splitting out `msg.value` support into its own PR. That way, we will be able to more easily gauge changes e.g. https://github.com/0xalpharush/foundry/pull/7 with scfuzzbench.

grandizzy:
> I would prefer holding off on "Unified mutation strategies", finalizing and merging #12587, and splitting out `msg.value` support into its own PR. That way, we will be able to more easily gauge changes e.g. [0xalpharush#7](https://github.com/0xalpharush/foundry/pull/7) with scfuzzbench.



> I would prefer holding off on "Unified mutation strategies", finalizing and merging #12587, and splitting out `msg.value` support into its own PR. That way, we will be able to more easily gauge changes e.g. [0xalpharush#7](https://github.com/0xalpharush/foundry/pull/7) with scfuzzbench.

https://github.com/foundry-rs/foundry/pull/12587 is rfr, ptal

grandizzy:
> I would prefer holding off on "Unified mutation strategies", finalizing and merging #12587, and splitting out `msg.value` support into its own PR. That way, we will be able to more easily gauge changes e.g. [0xalpharush#7](https://github.com/0xalpharush/foundry/pull/7) with scfuzzbench.

made msg.value fuzzable in https://github.com/foundry-rs/foundry/pull/14565 will retire this one and then follow-up on corpus improvements

Issue #12123, opened 2025-10-16
Push v1 docker tag


pimousse1099:
### Component

Other (please describe)

### Describe the feature you would like

Currently, nightly/stable/latest and specific docker tags per release (ex: v1.4.1, v1.3.0, etc) are being pushed. However, users cannot follow a major or minor version reliably via Docker as tags like 'v1' or 'v1.4' are not published. 

Please push tags for major (v1) and minor (v1.4) releases, in addition to patch-specific tags, to allow users to easily stick to a major/minor version.

This is standard for many Docker projects and makes maintenance and upgrades safer and simpler, especially for production infrastructure.

### Additional context

Many infrastructure setups rely on semantic tags to avoid breaking changes and automate dependency upgrades. Lack of v1/v1.4 tags requires manual updates and increases the risk of running outdated or incompatible versions.

Does that makes sens to you guys or do you have another philosophy regarding docker tag management ?

zerosnacks:
Hi @pimousse1099, we've just released `v1.6.0-rc1` release candidate. When we publish `v1.6.0` (soon) it will also be tagged on `v1` and `v1.6` . Thanks for your suggestion!

PR #13182, opened 2026-01-22
feat(docker): add semver major/minor floating tags on release

zerosnacks:
Closes: https://github.com/foundry-rs/foundry/issues/12123

When publishing Docker images for version releases, this adds floating major and minor tags alongside the exact version tag.

For example, releasing `v1.2.3` now produces three tags:
- `ghcr.io/foundry-rs/foundry:v1.2.3` (exact version)
- `ghcr.io/foundry-rs/foundry:v1.2` (minor floating)
- `ghcr.io/foundry-rs/foundry:v1` (major floating)

This allows users to pin to a major or minor version and automatically receive patch updates.

## How it works

The regex `^v([0-9]+)\.([0-9]+)\.([0-9]+)$` captures semver components into `BASH_REMATCH`:

```bash
TAG="v1.2.3"
[[ "$TAG" =~ ^v([0-9]+)\.([0-9]+)\.([0-9]+)$ ]]
# BASH_REMATCH[1] = "1" (major)
# BASH_REMATCH[2] = "2" (minor)
# BASH_REMATCH[3] = "3" (patch)
```

These are then used to construct the floating tags `v1` and `v1.2`.

## Pre-release handling

Release candidates (e.g., `v1.2.3-rc1`) are intentionally excluded by the `$` anchor in the regex, which requires the string to end after the patch number. RC tags only get their exact tag, ensuring the floating `v1` and `v1.2` tags continue pointing to the latest stable release.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -65,15 +65,24 @@ jobs:
       - name: Finalize Docker Metadata
         id: docker_tagging
         run: |
+          TAG="${{ inputs.tag_name }}"
+          REGISTRY="${{ env.REGISTRY }}"
+          IMAGE="${{ env.IMAGE_NAME }}"
+
           if [[ "${{ github.event_name }}" == "schedule" ]]; then
             printf "cron trigger, assigning nightly tag\n"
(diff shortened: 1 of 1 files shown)


zerosnacks:
Quite a small thing but makes sense to me, it is in line with the `v1` tag we maintain for `foundry-toolchain` / `forge-std`.

cc @grandizzy when time permits, would you mind reviewing? I think you have most context on this.

PR #13188, opened 2026-01-23
fix(cheatcodes): fix expectRevert with internal reverts for non-contract calls

### LINKED ISSUE #11559: expectRevert bug with internal call
### Component

Forge

- [x] Foundry
- [x] Foundryup

1.3.2-stable

_No response_

forge t

### Operating System

macOS (Apple Silicon)

### Describe the bug

Both
```solidity
/// forge-config: default.allow_internal_expect_revert = true
function testA() {
    IERC20(address(0)).totalSupply();
}
```
and
```solidity
/// forge-config: default.allow_internal_expect_revert = true
function testB() {
    vm.expectRevert();
    IERC20(address(0)).totalSupply();
}
```
fail (the first one for "revert" and the second one for "call didn't revert as expected"). I guess that this is not the expected behavior (the second one should pass, considering this call really fails as address zero has no code).

edit: note that also both
```solidity
function testC() public {
    IERC20(address(0)).totalSupply();
}

function testD() public {
    vm.expectRevert();
    IERC20(address(0)).totalSupply();
}
```
fail but that's expected

gakonst:
## Summary

Fixes #11559

When `allow_internal_expect_revert = true`, calling a non-contract address (e.g., `IERC20(address(0)).totalSupply()`) with `expectRevert` would fail with "next call did not revert as expected".

## Root Cause

The issue is that calls to addresses with no code return immediately with success (`Stop`), and the `expectRevert` was being consumed by this successful call. The actual revert happens afterward when Solidity's generated code checks the return data (or extcodesize) and reverts.

The sequence was:
1. `vm.expectRevert()` sets up the expectation at depth N
2. `IERC20(address(0)).totalSupply()` makes an external call to address(0)
3. The call returns `Stop` (success with no data) because there's no code
4. `call_end` is triggered, sees `curr_depth <= expected_revert.depth`, and consumes the expectation
5. The expectation fails because the call didn't revert
6. Solidity's code then reverts (too late - expectation already consumed)

## Fix

The fix modifies the `call_end` hook to only consume the `expectRevert` when:
1. The call actually reverted, OR
2. An external call was made (`max_depth > depth`), OR  
3. `internal_expect_revert` is disabled (maintaining backward compatibility)

When `internal_expect_revert` is enabled and a call succeeds without going deeper, the expectation is preserved, allowing the subsequent internal revert (from Solidity's code validation) to satisfy it.

## Test

Added `Issue11559.t.sol` in testdata/default/repros/ with test cases.

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -1331,7 +1331,8 @@ impl Inspector<EthEvmContext<&mut dyn DatabaseExt>> for Cheatcodes {
                         // (e.g., calls to non-contract addresses that return Stop).
                         //
                         // Only process if:
-                        // 1. It's not a cheatcode call, AND
+                        // 1. It's not a cheatcode call (unless internal_expect_revert is enabled
+                        //    and the cheatcode reverted), AND
                         // 2. Either the call reverted, OR we made an external call (max_depth >
                         //    depth), OR we're at the root call (depth 0)
                         //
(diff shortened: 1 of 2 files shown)


decofe:
Closing in favor of an in-repository branch PR.

PR #13191, opened 2026-01-23
feat(forge): specify compilation profile in vm.getCode

### LINKED ISSUE #12821: feat(forge): specify compilation profile in vm.getCode
### Component

Forge

It is possible to do `vm.getCode("Contract.sol:0.8.18")` and `vm.getCode("Contract.sol:0.8.28")` to choose between multiple compilations of the same contract.

This only works if the compilations use a different solidity version.

It would be useful to generalize to any compilation restriction listed in the current config. So you could do `vm.getCode("Contract.sol:v1")` and `vm.getCode("Contract.sol:v2")` if `foundry.toml` contains

```toml
[[profile.default.additional_compiler_profiles]]
name = "v1"
via_ir = false

[[profile.default.additional_compiler_profiles]]
name = "v2"
via_ir = true
```

### Additional context

_No response_

gakonst:
## Summary

Add support for specifying a compilation profile in artifact path parsing for `vm.getCode`, `vm.getDeployedCode`, and `vm.deployCode` cheatcodes.

## Motivation

This allows users with multiple compilation profiles in `foundry.toml` to select specific artifact versions. For example, if you have:

```toml
[profile.v1]
optimizer = false

[profile.v2]
optimizer = true
optimizer_runs = 1000
```

You can now use:
- `vm.getCode("Contract.sol:v1")`
- `vm.getCode("Contract.sol:v2")`
- `vm.getCode("Contract.sol:ContractName:optimized")`

## Implementation

The feature distinguishes between semver versions (like `0.8.23`) and profile names (like `v1`, `optimized`) by:
1. First checking if the suffix contains a dot (`0.8.23`)
2. Then attempting to parse as a semver version
3. If neither, treating it as a profile name

The profile filter is applied during artifact lookup to select the correct compiled artifact.

## Supported Formats

- `path/to/contract.sol:ContractName:profile`
- `path/to/contract.sol:profile`
- `ContractName:profile`

Closes #12821


crates/cheatcodes/src/fs.rs
```diff
@@ -30,6 +30,76 @@ use std::{
...
+    let path_or_name = parts.next().unwrap();
+    if path_or_name.contains('.') {
+        file = Some(PathBuf::from(path_or_name));
+        if let Some(name_or_version_or_profile) = parts.next() {
+            if name_or_version_or_profile.contains('.')
+                || Version::parse(name_or_version_or_profile).is_ok()
+            {
+                version = Some(name_or_version_or_profile);
+            } else {
+                contract_name = Some(name_or_version_or_profile);
```
mablr, line 71: This seems to make the documented `path/to/contract.sol:profile` form unusable. For a two-part file path like `paris/cheats/GetCodeProfile.t.sol:paris`, the suffix is always stored as `contract_name`, so lookup searches for a contract named `paris` instead of filtering by profile `paris`. We should either remove that advertised format or resolving the ambiguity during artifact lookup, and add a test for the two-segment file+profile case?

grandizzy: pls check https://github.com/foundry-rs/foundry/pull/13191/changes/b06d8a77253d2e3af15d83170518e957d3cd7057

mablr (approved):

grandizzy (approved):

Issue #12683, opened 2025-11-28
Use pending for fetching nonces in scripts


LogvinovLeon:
### Component

Forge

### Have you ensured that all of these are up to date?

- [ ] Foundry
- [x] Foundryup

### What version of Foundry are you on?

forge Version: 1.5.0-dev

### What version of Foundryup are you on?

 leo@sov-dev-box:~/sdk/foundry$ 

### What command(s) is the bug in?

forge script

### Operating System

Linux

### Describe the bug

When running foundry scripts - I'm sometimes encountering nonce issues.
This happens because when the script fetches the nonce for account - it uses "latest" instead of "pending" so if there is a pending tx - we will submit the tx with a colliding nonce.

The fix is a 1-line fix here: https://github.com/foundry-rs/foundry/blob/master/crates/script/src/broadcast.rs#L58

Please let me know if you agree and I'll submit the PR.

grandizzy:
thank you, I think this makes sense though wonder if you could share a repro / the chain you experience this? CC @mattsse @klkvr could you pls chime in?

LogvinovLeon:
I was experiencing it on a non-standard high-performance EVM node implementation that had a different execution model. Instead of the mempool model or mini-blocks - it used the preferred sequencer and executed new txs immediately into a pending block. The state changes (including the nonces) were visible immediately using the `pending` tag but you needed to wait for the block time to see them using the `latest` tag.

When submitting this issue I understood that it might not be the best use of foundry resources to support such an implementation. At the same time - it seemed like a simple fix which has no downsides on all major implementations. `latest` -> `pending`. On nodes that do not mine - they're equivalent. Now I see that it complicates the implementation as it needs to be done consistently in all the places we fetch state (next_nonce, fork-db, maybe more). And also - not all nodes support pending tags.

Therefore - feel free to close this issue on your discression

PR #13192, opened 2026-01-23
fix(script): use pending block tag for nonce fetching with fallback

gakonst:
## Summary

Fixes nonce collisions when running forge scripts with pending transactions in the mempool.

## Problem

When the script fetches the nonce for an account, it uses `latest` instead of `pending`, so if there are pending transactions, we submit with a colliding nonce.

Ref: #12683

## Solution

- Use `pending` block tag by default to account for mempool transactions
- Memoize per-provider fallback to `latest` for RPCs that don't support `pending` (some L2s, older providers)
- Avoids repeated failed requests after first fallback detection

## Tradeoffs

| Aspect | Notes |
|--------|-------|
| **Correctness** | `pending` gives accurate nonces accounting for mempool |
| **Compatibility** | Graceful fallback handles RPCs without `pending` support |
| **Performance** | One extra request only on first failure per provider URL |

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4329,6 +4329,7 @@ dependencies = [
  "foundry-debugger",
  "foundry-evm",
  "foundry-linking",
+ "foundry-test-utils",
  "foundry-wallets",
  "futures",
  "indicatif",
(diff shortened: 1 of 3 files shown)


zerosnacks:
Closing per https://github.com/foundry-rs/foundry/issues/12683#issuecomment-3790777782

PR #13193, opened 2026-01-23
fix(invariant): fix optimizer shrink logic for warp/roll sequences

grandizzy:
## Summary

This PR fixes issues with the invariant optimizer shrink logic that caused the "best sequence" to not reproduce the same value when replayed.

## Feature: Invariant Optimization Mode

Added optimization mode for invariant testing (activated when invariant function returns `int256`). The fuzzer maximizes the return value instead of checking for boolean success.

Key implementation:
- `InvariantMode` detection based on ABI return type
- Tracking `optimization_best_value` and `optimization_best_sequence` in `InvariantTestData`
- `can_continue` handles optimization mode by tracking max value without failing on invariant
- `check_sequence_value` for replaying and verifying optimization sequences
- Shrinking finds shortest sequence that still produces the target maximum value

### Problem
The invariant optimizer's "best sequence" didn't reproduce the same value when replayed because:

1. **Reverted calls lost warp/roll values**: When a call reverted (in non-fail-on-revert mode), it was removed from the sequence, but its warp/roll values were still applied during fuzzing. During shrinking replay, these warps were missing.

2. **Shrunk sequence didn't accumulate warps**: When shrinking removed calls, their warp/roll values were also lost. The displayed sequence showed individual call warps instead of accumulated values.

3. **Inspector block override issue**: The inspector's `cheatcodes.block` overrides the env during interpreter initialization. When warp/roll was applied to `executor.env` but not to `cheatcodes.block`, subsequent calls saw incorrect timestamps.

### Solution

1. **Keep reverted calls in optimization mode**: In `can_continue` and the skip-invariant-check path, don't pop reverted calls when in optimization mode. This preserves warp/roll values for correct replay.

2. **Accumulate warps in shrunk output**: When building the shrunk sequence, accumulate warp/roll values from all removed calls into the next kept call. This ensures the displayed sequence can be copied directly into a test.

3. **Update both env and inspector block**: In `execute_tx`, update both `executor.env.block_env` AND `cheatcodes.block` with warp/roll values. This ensures correct behavior regardless of whether cheatcodes.block overrides the env.

4. **Fix `check_sequence_value`**: Apply warps/rolls from ALL calls in the sequence, but only execute calls that are in the kept set. This correctly reproduces the block environment during shrinking.

## Test Results

- All 46 invariant tests pass
- Optimizer shrinking correctly produces reproducible sequences
- Example: A sequence shrunk from 285 calls to 1 call with accumulated warp correctly reproduces the best value

## Example

Before fix:
```
[Best sequence] (original: 285, shrunk: 1)
    vm.warp(block.timestamp + 482030);  // Only this call's warp
    ...
invariant_opt_price_difference() (best: 2660072, ...)
```
When replayed: produces 0 (wrong)

After fix:
```
[Best sequence] (original: 285, shrunk: 1)
    vm.warp(block.timestamp + 87139295);  // Accumulated from all 285 calls
    ...
invariant_opt_price_difference() (best: 72729989..., ...)
```
When replayed: produces the exact best value (correct)

### DIFF
--- a/crates/evm/evm/src/executors/invariant/mod.rs
+++ b/crates/evm/evm/src/executors/invariant/mod.rs
@@ -6,7 +6,7 @@ use crate::{
     inspectors::Fuzzer,
 };
 use alloy_primitives::{
-    Address, Bytes, FixedBytes, Selector, U256,
+    Address, Bytes, FixedBytes, I256, Selector, U256,
     map::{AddressMap, HashMap},
 };
 use alloy_sol_types::{SolCall, sol};
@@ -56,7 +56,7 @@ mod result;
 pub use result::InvariantFuzzTestResult;
 
 mod shrink;
-pub use shrink::check_sequence;
+pub use shrink::{check_sequence, check_sequence_value};
 
 sol! {
     interface IInvariantTest {
@@ -145,6 +145,11 @@ struct InvariantTestData {
     // until the desired `depth` so we can use the evolving fuzz dictionary
     // during the run.
     branch_runner: TestRunner,
+
+    // Optimization mode state: tracks the best (maximum) value and the sequence that produced it.
+    // Only used when invariant function returns int256.
+    optimization_best_value: Option<I256>,
+    optimization_best_sequence: Vec<BasicTxDetails>,
 }
 
 /// Contains invariant test data.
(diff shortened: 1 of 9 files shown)


grandizzy:
https://github.com/foundry-rs/foundry/pull/13196

PR #13196, opened 2026-01-23
feat(invariant): add optimization mode for invariant testing

### LINKED ISSUE #12190: feat(invariant): add Optimization mode to Invariant Testing similar to Echidna
### Component

Forge

In Echidna, optimization mode is used to fuzz for the maximized return value of a given function.

Example:
```solidity
    function echidna_opt_price_difference() public view returns (int256) {
        return maxPriceDifference;
    }
```

It basically runs the fuzzer like normal, but instead of looking for broken invariants, it stores the highest value returned by the function being tested. I often use this to determine the max error due to rounding to determine if this is acceptable, but it has many uses.

### Additional context

For more info see: https://secure-contracts.com/program-analysis/echidna/advanced/optimization_mode.html

grandizzy:
## Summary

Adds optimization mode for invariant testing, similar to Echidna. This mode maximizes an `int256` return value from an `invariant*` function.

Closes #12190

## Usage

Define an invariant function returning `int256` :

```solidity
function invariant_optimize_maxRoundingError() external view returns (int256) {
    return int256(pool.totalShares()) - int256(pool.expectedShares());
}
```

- Foundry will fuzz to find the sequence that maximizes this value
- The best value and sequence are tracked and reported
- Sequence shrinking is applied to find the minimal reproducing sequence

```solidity

contract InvariantOptimizer is Test {
    int256 maxPriceDifference;

    function setUp() public {
        targetContract(address(this));
    }

    function setMaxPriceDifference(uint256 startPrice, uint256 endPrice, uint256 startTime, uint256 endTime) public {
        /// logic to set maxPriceDifference here
    }

    function invariant_opt_price_difference() public view returns (int256) {
        return maxPriceDifference;
    }
}

[PASS]
	[Best sequence] (original: 392, shrunk: 1)
		vm.warp(block.timestamp + 123408035);
		vm.prank(0x7FA9385bE102ac3EAc297483Dd6233D62b3e1496);
		InvariantOptimizer(0x7FA9385bE102ac3EAc297483Dd6233D62b3e1496).setMaxPriceDifference(115792089237316195423570985008687907853269984665640564039457584007913129639933, 115792089210356248756420345214020892766250353992003419616917011526809519390721, 2230, 3400);
 invariant_opt_price_difference() (best: 2843574705161656291476921143606979357380022738070982514754023847085364228, runs: 256, calls: 128000)
```

## Bug fix during implementation

Fixed a bug in shrink logic where `vm.warp` and `vm.roll` values were not correctly accumulated when replaying shrunk sequences. This caused the "best" sequence to be non-reproducible because time/block values depended on removed calls.

The fix:
- Keeps reverted calls in sequence during optimization to preserve warp/roll
- Accumulates warps/rolls from removed calls into kept calls during shrinking
- Updates inspector's cheatcodes.block alongside executor.env

## Tests

- `invariant_optimization_mode` - basic optimization with deterministic increment
- `invariant_optimization_negative_values` - optimization starting from negative values
- `invariant_optimization_with_warp` - warp-based optimization with shrunk sequence output (fixed seed)

### DIFF
--- a/crates/evm/evm/src/executors/invariant/shrink.rs
+++ b/crates/evm/evm/src/executors/invariant/shrink.rs
@@ -30,6 +30,50 @@ impl CallSequenceShrinker {
     fn current(&self) -> impl Iterator<Item = usize> + '_ {
         (0..self.call_sequence_len).filter(|&call_id| self.included_calls.test(call_id))
     }
+
+    /// Advance to the next call index, wrapping around to 0 at the end.
+    fn next_index(&self, call_idx: usize) -> usize {
+        if call_idx + 1 == self.call_sequence_len { 0 } else { call_idx + 1 }
+    }
+}
+
+/// Resets the progress bar for shrinking.
+fn reset_shrink_progress(config: &InvariantConfig, progress: Option<&ProgressBar>) {
+    if let Some(progress) = progress {
+        progress.set_length(config.shrink_run_limit as u64);
+        progress.reset();
+        progress.set_message(" Shrink");
+    }
+}
+
+/// Applies accumulated warp/roll to a call, returning a modified copy.
+fn apply_warp_roll(call: &BasicTxDetails, warp: U256, roll: U256) -> BasicTxDetails {
+    let mut result = call.clone();
+    if warp > U256::ZERO {
+        result.warp = Some(warp);
+    }
+    if roll > U256::ZERO {
+        result.roll = Some(roll);
+    }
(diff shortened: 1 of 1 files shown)


grandizzy:
> makes sense 👍, it would be good if this was paired with a PR in the Foundry book because for a user it is not very intuitive.

yep https://github.com/foundry-rs/book/pull/1737

PR #13199, opened 2026-01-23
fix(script): update sender nonce when using startBroadcast(privateKey) without libraries

### LINKED ISSUE #12646: Wrong nonce when setting the deployer in a script
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.3.5-stable

foundryup: 1.3.0

forge script

### Operating System

Linux

### Describe the bug

If the sender is set from within the script by passing the private key - foundry fetches the nonce for the default sender and because it doesn't change - ends up re-using the 0 nonce causing transactions to get rejected.
Nonces are fetched here: https://github.com/foundry-rs/foundry/blob/75151e9cae397d34034c1d03defc4f94b6363a3c/crates/script/src/broadcast.rs#L50
This is called in two places:
* new (with default sender): https://github.com/foundry-rs/foundry/blob/master/crates/script/src/lib.rs#L586
* update_sender (with custom sender): https://github.com/foundry-rs/foundry/blob/master/crates/script/src/lib.rs#L597
`update_sender` is only called after `maybe_new_sender` returns `Some(sender)` https://github.com/foundry-rs/foundry/blob/master/crates/script/src/execute.rs#L118
Unfortunately - maybe_new_sender only returns Some when there are libraries to predeploy. So when we have none - sender is left as default.

There is a workaround of just using --sender in CLI but would still be nice to get it fixed.

gakonst:
## Summary

When using `vm.startBroadcast(privateKey)` in a script without the `--sender` CLI flag, the sender nonce was not updated if there were no libraries to predeploy. This caused transactions to be rejected due to incorrect nonces.

## Problem

The `maybe_new_sender` function in [execute.rs](https://github.com/foundry-rs/foundry/blob/master/crates/script/src/execute.rs#L180-L209) previously only checked for a new sender when `libraries_count() > 0`:

```rust
if self.build_data.predeploy_libraries.libraries_count() > 0
    && self.args.evm.sender.is_none()
```

This meant that when a script used `vm.startBroadcast(pk)` to set a different sender, but had no libraries to predeploy, `update_sender` was never called. The `sender_nonce` in `ScriptConfig` remained set to the default sender's nonce, causing nonce mismatches during broadcast.

## Solution

This PR relaxes the condition to also detect sender changes when there are no libraries, ensuring the sender nonce is updated correctly for all broadcast scenarios.

## Testing

Added a new test case `can_deploy_script_private_key_no_lib` that:
1. Uses `vm.startBroadcast(privateKey)` without `--sender` CLI flag
2. Has no libraries to predeploy
3. Verifies correct nonce increment during broadcast

Fixes #12646

### DIFF
--- a/crates/forge/tests/cli/script.rs
+++ b/crates/forge/tests/cli/script.rs
@@ -713,6 +713,26 @@ forgetest_async!(can_deploy_script_with_lib, |prj, cmd| {
         .await;
 });
 
+// Regression test for https://github.com/foundry-rs/foundry/issues/12646
+// Tests that using vm.startBroadcast(privateKey) without --sender CLI flag
+// correctly updates the sender nonce even when there are no libraries to predeploy.
+forgetest_async!(can_deploy_script_private_key_no_lib, |prj, cmd| {
+    let (_api, handle) = spawn(NodeConfig::test()).await;
+    let mut tester = ScriptTester::new_broadcast(cmd, &handle.http_endpoint(), prj.root());
+
+    tester
+        .load_addresses(&[address!("0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC")])
+        .await
+        .add_sig("BroadcastTestNoLinking", "deployWithPrivateKeyNoLibraries()")
+        .simulate(ScriptOutcome::OkSimulation)
+        .broadcast(ScriptOutcome::OkBroadcast)
+        .assert_nonce_increment_addresses(&[(
+            address!("0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC"),
+            2,
+        )])
+        .await;
+});
+
 forgetest_async!(can_deploy_script_private_key, |prj, cmd| {
     let (_api, handle) = spawn(NodeConfig::test()).await;
     let mut tester = ScriptTester::new_broadcast(cmd, &handle.http_endpoint(), prj.root());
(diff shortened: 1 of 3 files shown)


gakonst:
Closing this PR - my initial fix was incorrect and broke existing tests. 

After deeper analysis, the issue is more complex:

**Root Cause Analysis:**
The `maybe_new_sender` function only returns a new sender when there are libraries to predeploy. When it returns `Some(new_sender)`, the code calls `update_sender` (which fetches the correct nonce from RPC) and then re-executes the script with the new sender.

When there are NO libraries to predeploy, `maybe_new_sender` returns `None`, and `update_sender` is never called. However, this doesn't directly cause the nonce issue because:

1. During script execution, broadcast transactions get their nonces from `account.info.nonce` in the EVM state (inspector.rs line 918)
2. `load_account` is called for the broadcast address before reading the nonce (inspector.rs line 895)
3. `load_account` should fetch the account from the fork RPC if forking

**Why my fix failed:**
My fix made `maybe_new_sender` return the new sender even without libraries, which triggered a re-link and re-execute cycle. This caused issues for tests that:
1. Don't have a fork URL but expect scripts to work
2. Rely on the script only executing once

**Next steps:**
A proper fix likely needs to either:
1. Separate the sender nonce update logic from the library relinking logic
2. Investigate why `load_account` might not be returning the correct nonce in certain scenarios
3. Get a reproducible test case from the issue reporter

I'll leave investigation notes on the issue.

PR #13201, opened 2026-01-23
feat(cast): add --flatten flag to cast interface

### LINKED ISSUE #9960: cast interface generates unexpected library
### Component

Forge

- [x] Foundry
- [ ] Foundryup

9bcfbaa6e6066fb9f87ce07163f5479d4b124563

_No response_

cast interface

### Operating System

None

### Describe the bug

Hey, i think there is a regression in `cast interface` (or at least an unexpected change).

I have a:
```
interface ITestContract {
  struct TestStruct {
    address asset;
  }
}

contract TestContract is ITestContract {
  function test(TestStruct memory) external {}
}
```

Now `cast interface ./src/Test.sol:TestContract` will yield:
```
pragma solidity ^0.8.4;

library ITestContract {
    struct TestStruct {
        address asset;
    }
}

interface TestContract {
    function test(ITestContract.TestStruct memory) external;
}
```

Which I think different to previous versions as i'm pretty sure that before structs were included in the generated interface. Is this change intentional, as ux wise it's a bit weird.

---

More of a question out of curiosity: Would it be possible to include the natspec on the generated interface?
In most smart contracts i'm reading i have to constantly jump between interface and contract to fully understand the code. Therefore i'm wondering if it wouldn't be a better approach to just have all the natspec on the actual contract (not the interface) and then simply generate the interface.

mattsse:
## Summary

Adds a new `--flatten` flag to `cast interface` that inlines inherited/library struct types directly into the generated interface.

## Problem

When running `cast interface` on a contract that uses structs from an inherited interface, the output generates a `library` for the struct definitions instead of keeping them in the interface (#9960).

Example of the issue:
```solidity
// Generated without flag - struct in separate library
library IBase {
    struct TestStruct { address asset; }
}
interface Contract {
    function test(IBase.TestStruct memory) external;
}
```

## Solution

Uses alloy-json-abi's `ToSolConfig::one_contract(true)` option (introduced in alloy-core 0.8.24 via alloy-rs/core#911) to consolidate all types into a single interface:

```solidity
// Generated with --flatten flag
interface Contract {
    // Types from \`IBase\`
    struct TestStruct { address asset; }
    
    function test(TestStruct memory) external;
}
```

## Changes

- Added `--flatten` CLI flag to `cast interface` command
- Pass `ToSolConfig::new().one_contract(true)` to `JsonAbi::to_sol()` when flag is set
- Added test case for the new flag

Closes #9960

### DIFF
--- a/crates/cast/tests/cli/main.rs
+++ b/crates/cast/tests/cli/main.rs
@@ -2513,13 +2513,8 @@ interface Interface {
     ]]);
 
     // With --flatten, the struct is inlined into the interface
-    cmd.cast_fuse()
-        .arg("interface")
-        .arg("--flatten")
-        .arg(&path)
-        .assert_success()
-        .stdout_eq(str![[
-            r#"// SPDX-License-Identifier: UNLICENSED
+    cmd.cast_fuse().arg("interface").arg("--flatten").arg(&path).assert_success().stdout_eq(str![
+        [r#"// SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.4;
 
 interface Interface {
@@ -2531,8 +2526,8 @@ interface Interface {
     function test(TestStruct memory param) external;
 }
 
-"#
-        ]]);
+"#]
+    ]);
 });
 
 // tests that fetches WETH interface from etherscan


crates/cast/src/cmd/interface.rs
```diff
@@ -50,13 +50,19 @@ pub struct InterfaceArgs {
     )]
     output: Option<PathBuf>,
 
+    /// If set, generate all types in a single interface, inlining any inherited or library types.
+    ///
+    /// This can fail if there are structs with the same name in different interfaces.
+    #[arg(long)]
+    all_in_one: bool,
```
zerosnacks: maybe `flatten` is more intuitive here

PR #13203, opened 2026-01-23
fix(chisel): skip external identifiers for local execution

echowandere:
Chisel was creating external identifiers unconditionally in decode_traces, causing unnecessary Etherscan/Sourcify requests even for purely local REPL sessions. This adds the same 5s delay that was reported in the original issue.

Applied the same fix as forge test - only create external identifiers when there's actually a remote chain connected. No point hitting external APIs when running locally.

inspired by https://github.com/foundry-rs/foundry/pull/13189

### DIFF
--- a/crates/chisel/src/dispatcher.rs
+++ b/crates/chisel/src/dispatcher.rs
@@ -165,10 +165,13 @@ impl ChiselDispatcher {
             )?)
             .build();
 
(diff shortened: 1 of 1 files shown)


echowandere:
Mind rerun the CI? looks weird here.

emmajam:

PR #13209, opened 2026-01-24
refactor(linking): remove unused Artifact import

marukai67:
Removes unused `Artifact` import from `foundry-linking` crate to eliminate unnecessary dependencies.


onbjerg:
this is used

PR #11675, opened 2025-09-17
[WIP] feat(forge): add precompiles crate

grandizzy:
## Motivation
- closes #11622 
- moves network precompiles in own crate, reused in forge
- todo: apply to cast as well

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #9110, opened 2024-10-14
feat(`anvil`): `wallet_` namespace + inject P256BatchDelegation + executor

yash-atreya:
## Motivation

Introduce `odyssey` features to anvil, activated using the `--odyssey`.

## Solution

- [x] `wallet_getCapabilities`
- [x] `wallet_sendTransaction`
- [x]  Inject P256BatchDelegation in backend
- [x]  Setup funded executor that sponsors the transactions
- [x] Tests

PR #13211, opened 2026-01-24
chore(anvil): clean up remaining dead code after Odyssey sunset

mablr:
## Motivation

Clean up remaining dead code left behind after Odyssey sunset.

This is follow-up for #11675

# Solution

- Removed remaining RPC endpoints, constants and methods initially added in #9110 for Odyssey support

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes (removed only dead code)

### DIFF
--- a/crates/anvil/core/src/eth/mod.rs
+++ b/crates/anvil/core/src/eth/mod.rs
@@ -706,18 +706,6 @@ pub enum EthRequest {
     /// Rollback the chain
     #[serde(rename = "anvil_rollback", with = "sequence")]
     Rollback(Option<u64>),
-
-    /// Wallet
-    #[serde(rename = "wallet_getCapabilities", with = "empty_params")]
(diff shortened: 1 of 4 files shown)


crates/anvil/core/src/eth/mod.rs
```diff
@@ -706,18 +706,6 @@ pub enum EthRequest {
     /// Rollback the chain
     #[serde(rename = "anvil_rollback", with = "sequence")]
     Rollback(Option<u64>),
-
-    /// Wallet
-    #[serde(rename = "wallet_getCapabilities", with = "empty_params")]
-    WalletGetCapabilities(()),
```
zerosnacks, line 712: I believe we would still want this per https://eips.ethereum.org/EIPS/eip-5792 

cc @jxom / @tmm

Maybe `anvil_addCapability` is also still relevant

PR #13215, opened 2026-01-25
fix(invariant): preserve warp/roll from removed calls during shrinking

### LINKED ISSUE #13214: bug(invariant): wrong simplified sequence with roll and warp
### Component

Forge

- [ ] Foundry
- [ ] Foundryup

v1.6.0-rc1

_No response_

_No response_

### Operating System

None

## Problem:

the way counterexample is shrunk when newly introduced max_block_delay and max_time_delay are used is not ideal, as the reverted calls are advancing these block properties as well but they're removed from the final sequence to shrink if optimization mode not enabled https://github.com/foundry-rs/foundry/blob/3374adb64d9e23ee8fca7e16906df4c91872df0c/crates/evm/evm/src/executors/invariant/mod.rs#L521-L524 This is exemplified in tests below, running forge test --mt invariant on

```solidity
pragma solidity ^0.8.13;

contract Roll {
    uint256 public number;

    function increment() public {
        require(block.number > 50000, "wrong block");
        number++;
    }
}

contract InvariantRoll is Test {
    Roll public roll;

    function setUp() public {
        roll = new Roll();
    }

    /// forge-config: default.fuzz.seed = "119"
    /// forge-config: default.invariant.max_block_delay = 60480
    /// forge-config: default.invariant.show_solidity = true
    function invariant_roll() public view {
        require(roll.number() == 0, "max block");
    }

    function testReproRoll() external {
        vm.roll(block.number + 26603);
        vm.prank(0x00000000000000000000000000000000000011C7);
        Roll(0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f).increment();
        this.invariant_roll();
    }
}

contract Warp {
    uint256 public number;

    function increment() public {
        require(block.timestamp > 500000, "wrong timestamp");
        number++;
    }
}

contract InvariantWarp is Test {
    Warp public warp;

    function setUp() public {
        warp = new Warp();
    }

    /// forge-config: default.fuzz.seed = "119"
    /// forge-config: default.invariant.max_time_delay = 604800
    /// forge-config: default.invariant.show_solidity = true
    function invariant_warp() public view {
        require(warp.number() == 0, "max time");
    }

    function testReproWarp() external {
        vm.warp(block.timestamp + 208043);
        vm.prank(0x0000000000000000000000000000000000001264);
        Warp(0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f).increment();
        this.invariant_warp();
    }
}
```

produces
```
[FAIL: max block]
	[Sequence] (original: 1, shrunk: 1)
		vm.roll(block.number + 26603);
		vm.prank(0xa2bd4EC1081f77bEC1c4626B6a92BBc3B8D4f24a);
		Roll(0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f).increment();
 invariant_roll() (runs: 0, calls: 0, reverts: 3)

Encountered 1 failing test in test/WarpAndRollBug.t.sol:InvariantWarp
[FAIL: max time]
	[Sequence] (original: 1, shrunk: 1)
		vm.warp(block.timestamp + 208043);
		vm.prank(0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF);
		Warp(0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f).increment();
 invariant_warp() (runs: 0, calls: 0, reverts: 3)
```

output but forge test --mt testRepro for repros generated fails with different revert message
```
Encountered 1 failing test in test/WarpAndRollBug.t.sol:InvariantRoll
[FAIL: wrong block] testReproRoll() (gas: 7221)

Encountered 1 failing test in test/WarpAndRollBug.t.sol:InvariantWarp
[FAIL: wrong timestamp] testReproWarp() (gas: 7176)
```

## Proposed solution:
- keep strategy to generate warp and roll the same https://github.com/foundry-rs/foundry/blob/3374adb64d9e23ee8fca7e16906df4c91872df0c/crates/evm/fuzz/src/strategies/invariants.rs#L106-L110 but update BasicTxDetails with the effective block and timestamp so when execute_tx happens we apply them entirely instead adding to current block and timestamp
https://github.com/foundry-rs/foundry/blob/3374adb64d9e23ee8fca7e16906df4c91872df0c/crates/evm/evm/src/executors/invariant/mod.rs#L416
- unify logic of optimizer mode with regular invariant calls so shrink_sequence_value and check_sequence_value are no longer needed
- shrinked sequence should show `vm.roll(effective_block_number)` and `vm.warp(effective_block_timestamp)` instead block.number + diff and block.timestamp + diff for solidty output; for normal output should show block.number = and block.timestamp = instead current roll = and warp = ;
- persisted failure and corpus should also retain the effective block number and timestamp at which the tx should be applied
- include example above as unit test to prevent against regressions  
- small nit, in existing invariant_warp_and_roll test checks block.number < 200000 in invariant_warp and block.timestamp < 500000 in invariant_roll whereas should be the other way around - invariant_warp check timestamp and invariant_roll checks block number;

gakonst:
## Summary

This fixes the bug where shrunk invariant counterexample sequences have incorrect warp/roll values because reverted calls' timing contributions were lost when they were removed.

Fixes #13214

### 1. Updated `check_sequence` in shrink.rs
Accumulates warp/roll from all calls (including removed ones) and applies the accumulated values to the next kept call. This matches the behavior of `check_sequence_value` in optimization mode.

### 2. Updated `shrink_sequence` in shrink.rs
Builds the final shrunk sequence with accumulated warp/roll values from removed calls, ensuring the reproduced sequence reaches the correct block state.

### 3. Fixed swapped test assertions
The `invariant_warp_and_roll` test had assertions that were swapped:
- `invariant_warp` was checking `block.number` instead of `block.timestamp`
- `invariant_roll` was checking `block.timestamp` instead of `block.number`

### 4. Added regression test
Added `invariant_shrink_preserves_warp_roll` test that verifies shrunk sequences have sufficient roll values to break invariants that depend on `block.number`.

## Test Plan

- [x] Code compiles with `cargo check -p foundry-evm`
- [ ] Run invariant tests: `cargo test invariant_warp_and_roll`
- [ ] Run regression test: `cargo test invariant_shrink_preserves_warp_roll`

/cc @grandizzy for review


grandizzy:
superseded by https://github.com/foundry-rs/foundry/pull/14218

PR #13224, opened 2026-01-26
fix(mutation): recursively symlink nested lib directories in temp workspace

gakonst:
## Problem

Projects with nested git submodules fail mutation testing because the temp workspace only symlinks top-level lib directories.

When libraries have their own dependencies in nested `lib/` folders, and remappings reference them with context-specific prefixes like:
```
lib/euler-earn:@openzeppelin=lib/euler-earn/lib/openzeppelin-contracts/
```

These paths need to exist in the mutation workspace for compilation to succeed.

## Solution

Add `symlink_nested_libs()` which recursively walks through each library and symlinks any nested `lib/` directories, ensuring deeply nested submodules are accessible in the temp workspace.

## Testing

Tested against `euler-xyz/evk-periphery` which has multiple levels of nested submodules. Before this fix, all 245 mutants failed with:
```
error="/tmp/forge_mutation_*/lib/openzeppelin-contracts/contracts/utils/math/SafeMath.sol": No such file or directory
```

## Checklist

- [x] Fixes compilation failures for projects with nested submodules
- [x] Recursively handles deeply nested lib structures
- [x] Only creates symlinks if target doesn't already exist

### DIFF
--- a/crates/forge/src/mutation/runner.rs
+++ b/crates/forge/src/mutation/runner.rs
@@ -230,7 +230,7 @@ fn symlink_nested_libs(lib_src: &Path, lib_dst: &Path) -> Result<()> {
     // Fall back to default "lib" if no config exists.
     let nested_lib_dirs: Vec<PathBuf> =
         if let Ok(config) = Config::load_with_root_and_fallback(lib_src) {
-            config.libs.clone()
+            config.libs
         } else {
             vec![PathBuf::from("lib")]
         };


zerosnacks:
cc @grandizzy additional fix for projects with recursive submodules

crates/forge/src/mutation/runner.rs
```diff
@@ -210,6 +215,54 @@ fn symlink_dir(src: &Path, dst: &Path) -> Result<()> {
...
+    };
+
+    for entry in entries.flatten() {
+        let entry_path = entry.path();
+        if !entry_path.is_dir() {
+            continue;
+        }
+
+        let entry_name = entry.file_name();
+        let nested_dst = lib_dst.join("lib").join(&entry_name);
```
grandizzy: Is this right or should we read config in case a custom lib dir is specified?

zerosnacks: Good call, we should indeed use the custom directory if specified - fixed here: https://github.com/foundry-rs/foundry/pull/13224/changes/6bf6b14e9f0c69e1663ca8934e11e9887dfd2794

PR #13225, opened 2026-01-26
fix(script): use weighted average for gas price calculation

dizer-ti:
The previous formula sum(gas_price) / count was incorrect because it didn't account for varying gas consumption per transaction. The correct formula is sum(gas_used * gas_price) / sum(gas_used) which ensures that total_gas * avg_gas_price = total_paid.


emmajam:

PR #13226, opened 2026-01-26
feat(cast): use prestate tracer to speed up cast run

### LINKED ISSUE #12719: feat(cast): use prestate tracer to speed up cast run

Himess:
## Motivation

Closes #12719

`cast run` replays every prior transaction in the block to reconstruct the pre-tx state, which is slow. When a node exposes the `debug_` namespace, the prestate tracer (`debug_traceTransaction`) returns exactly the accounts/storage the tx touches, letting us skip the replay.

## Solution

Add an opt-in `--prestate-tracer` flag to `cast run`. When set, it fetches the prestate via `debug_traceTransaction` and applies it directly (new `Executor::apply_prestate_trace`). If the call or response can't be used, it silently falls back to block replay.

Default behavior is unchanged (block replay), since most nodes don't expose `debug_`.

Builds on the approach from #12743 by @dipanshuhappy.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


onbjerg (commented):
I think the issue with this approach is that it assumes by default that `debug_` is even available. The reason `cast run` doesn't just use tracing APIs on nodes is that we assumed these were *not* available, which is the case for most nodes.

crates/cast/tests/cli/main.rs
```diff
@@ -3747,6 +3747,127 @@ Executing previous transactions from the block.
...
+    cmd.args([
+        "script",
+        "--private-key",
+        "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80",
+        "--rpc-url",
+        endpoint,
+        "--broadcast",
+        "CounterScript",
+    ])
+    .assert_success();
```
mablr: This should probably call `cmd.forge_fuse()` before invoking `script` like other tests do. Without switching to the Forge binary, the new cast tests are brittle outside full-workspace runs; because the `forge` binary could not available if running tests only on `cast` package.

mablr (changes requested):

grandizzy (approved):

mablr (approved):

Issue #10893, opened 2025-07-01
Potential regression bug: "chain ... not supported" (in v1.2.3)


0xDEnYO:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundryup are you on?

foundryup: 0.3.3

### What command(s) is the bug in?

forge script

### Operating System

macOS (Apple Silicon)

### Describe the bug

Hi,

I think there is a regression bug between v1.0.0 and v1.2.3.

We support 40+ EVM networks. 
I will share our foundry.toml below.

When I want to deploy to network Fuse with v1.2.3 I get an error 
`Error: Chain 122 not supported`

No idea where this error is originating from.

When I downgrade to v1.0.0, the same script runs fine and deploys any contract without issues. 

Our repo is public:
https://github.com/lifinance/contracts

A typical solidity deploy script:
https://github.com/lifinance/contracts/blob/main/script/deploy/facets/DeployExecutor.s.sol

Our Solidity deploy scripts are called by a bash script: 
https://github.com/lifinance/contracts/blob/main/script/deploy/deploySingleContract.sh

Please let me know if you need anything else to understand this bug, I am happy to support.

foundry.toml
````
[profile.default]
test = 'test/solidity'
solc_version = '0.8.17'
auto_detect_solc = false
evm_version = 'london'
optimizer = true
optimizer_runs = 1000000
# use a known address with a balance so ZkEVM scripts work
sender = '0x00a329c0648769a73afac7f9381e08fb43dbea73'
tx_origin = '0x00a329c0648769a73afac7f9381e08fb43dbea73'
out = 'out'
fs_permissions = [
  { access = "read", path = "./deployments/" },
  { access = "read", path = "./config/" },
  { access = "read-write", path = "./test/logs/" },
  { access = "read", path = "./zkout/" },
]
ffi = true
libs = ["node_modules", "lib"]
cache = true

[profile.zksync]
solc_version = '0.8.26'
test = "notest"
script = "script/deploy/zksync"
cache_path = "./zkcache"
out = "out/zksync"

[rpc_endpoints]
mainnet = "${ETH_NODE_URI_MAINNET}"
abstract = "${ETH_NODE_URI_ABSTRACT}"
apechain = "${ETH_NODE_URI_APECHAIN}"
arbitrum = "${ETH_NODE_URI_ARBITRUM}"
aurora = "${ETH_NODE_URI_AURORA}"
[...]

PR #13238, opened 2026-01-27
fix(config): Respect user-configured etherscan URL over chain defaults

YuyaMaruyama21D4E:
Thank you for maintaining Foundry! I use it regularly and really appreciate the work you do.

## Motivation
User-configured custom Etherscan URLs in `foundry.toml` are ignored for chains without default block explorer URLs (e.g., `NamedChain::Dev` / chain ID 1337), causing deployment failures even when a custom verification URL is explicitly configured.

**Problem**: When using Tenderly Virtual TestNet with chain ID 1337, Foundry fails with "Chain dev not supported" error despite having a custom verification URL in `foundry.toml`. This prevents developers from using Foundry's deployment and verification features with development networks and private chains.

  - Reference: [Tenderly Foundry Contract Verification](https://docs.tenderly.co/contract-verification/foundry)

**Root cause**: The `ResolvedEtherscanConfig::into_client()` method unconditionally calls `client_builder.chain(chain)?`, which:
1. Ignores the `api_url` that was already resolved by `resolve()` (either custom or default)
2. Queries `alloy-chains` again via `etherscan_urls()` to get default URLs
3. Fails for chains without default URLs

The error occurs even without `--verify` flag, as the Etherscan client is built during trace decoder construction for address identification.

**Related to #10893**, but addresses a different use case:
- **#10893**: Production chains (like Fuse) that should be added to `alloy-chains`
- **This PR**: Dev chains/private networks that should NOT be in `alloy-chains` but need custom URLs

## Solution

Respect user-configured URLs by eliminating the redundant `chain()` call in `into_client()`.

The fix modifies `ResolvedEtherscanConfig::into_client()` in `crates/config/src/etherscan.rs` to:

1. **Remove the unconditional `chain()` call**: Since `resolve()` already ensures `api_url` is populated (either custom or default), there's no need to query `alloy-chains` again
2. **Use `api_url` directly**: The `api_url` passed to `into_client()` is guaranteed to be valid (either user-configured or chain's default from `resolve()`)
3. **Browser URL fallback**: Use `api_url` for `browser_url` if not explicitly set

I'd appreciate any feedback on this approach. If there's a better way to handle this, or if I've overlooked something, I'm happy to revise.

This change:
- ✅ Respects user's explicit configuration in `foundry.toml`
- ✅ Maintains backward compatibility (`resolve()` still queries `alloy-chains` for default URLs when no custom URL is set)
- ✅ Eliminates redundant `alloy-chains` query in `into_client()` (already done in `resolve()`)
- ✅ Enables custom block explorers for chains without default URLs (e.g., chain ID 1337)

**Testing**:
- Added 2 unit tests covering chains with/without default URLs
- All 22 tests in `foundry-config` etherscan module pass
- Manually tested with Tenderly Virtual TestNet (chain ID 1337) - deployment and verification both work

**AI Assistance Disclosure**: This PR was developed with assistance from Claude (Anthropic) for codebase analysis, solution design, and documentation. Implementation was verified by the human contributor.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/config/src/etherscan.rs
+++ b/crates/config/src/etherscan.rs
@@ -485,10 +485,8 @@ mod tests {
         let config = resolved.remove("mainnet").unwrap().unwrap();
         assert_eq!(config.key, "ABCDEFG");
         let client = config.into_client().unwrap();
-        assert_eq!(
-            client.etherscan_api_url().as_str(),
-            "https://api.etherscan.io/v2/api?chainid=1"
-        );
+        // Custom URL should be used even when chain has a default URL
+        assert_eq!(client.etherscan_api_url().as_str(), "https://api.etherscan.io/api");
 
         unsafe {
             std::env::remove_var(env);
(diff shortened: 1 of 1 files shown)


grandizzy:
thank you, makes sense! @zerosnacks should we run the flaky CI with this PR to make sure verifier tests are OK? (or there aren't any that could be affected?)

zerosnacks:
@grandizzy yep makes sense, will run from https://github.com/foundry-rs/foundry/tree/zerosnacks/test-13238

zerosnacks:
Tested in https://github.com/foundry-rs/foundry/pull/13239

PR #13239, opened 2026-01-27
fix(config): respect user-configured etherscan URL over chain defaults

zerosnacks:
Follow up of https://github.com/foundry-rs/foundry/pull/13238#issuecomment-3804414941

Tested w/ `flaky-tests` and passing: https://github.com/foundry-rs/foundry/actions/runs/21394409266


zerosnacks:
cc @grandizzy

PR #13244, opened 2026-01-28
feat(anvil): support compressed state on-disk

itegulov:
Adds ability to load/dump gzip-compressed on-disk state when using anvil with `--load-state/--dump-state`. This is consistent with existing functionality in RPC (`anvil_{load,dump}State`) that already supports gzip compression.

The change is not breaking for `--load-state` as we are parsing out gzip header to identify whether this is plain JSON or gz. For `--dump-state` I decided to derive intention based on the provided file extension, i.e. `--dump-state state.json.gz` vs `--dump-state state.json`. Happy to change this to an extra config parameter or whatever you think would work better here.

## Motivation

State file (especially with preserved history) can be quite large. According to my experiments using gzip reduces the size by ~85% on average.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


zerosnacks:
Hi @itegulov thanks for your PR! Would you mind adding a few basic tests to verify the new behavior?

itegulov:
@zerosnacks I added tests but I had to make `PeriodicStateDumper` public otherwise it seems like there is no sane way to test dumping logic. Not sure how big of a deal is that; lmk if I missed something.

popzxc:
@zerosnacks can you please provide another review?

popzxc:
@mattsse maybe you can give it a look?

emmajam:

popzxc:
@emmajam still needed and valuable

PR #13247, opened 2026-01-28
perf(evm): reduce CallOutcome/CreateOutcome clones in inspector stack

zeroprooff:
Remove unnecessary clones in do_call_end and do_create_end hot paths.

Before: cloned entire CallOutcome for each inspector (up to 5x per call) plus unused return clone.
After: only clone output bytes on revert path (rare case).

Benchmark shows 1.4-3x speedup on isolated clone operations.


emmajam:

PR #13248, opened 2026-01-28
feat(primitives): introduce `NetworkWallet<FoundryNetwork>` impl for `EthereumWallet`

mablr:
## Motivation

Towards the `FoundryNetwork` transition, and simplified/seamless tx signature flow.

This is one of the first stage of this process, then the focus will be to make most of foundry components generic over `Network`.

This contribution will allow to get rid `WalletSigner`'s `NetworkWallet<FoundryNetwork>` impl once `FoundryNetwork` will be used everywhere 

## Solution

- Seamless `EthereumWallet` support for eth/op/tempo txs
- Implemented `RecommendedFillers` for `FoundryNetwork` (will be super useful later)


crates/primitives/src/transaction/request.rs
```diff
@@ -491,6 +491,24 @@ impl TransactionBuilder<FoundryNetwork> for FoundryTransactionRequest {
     }
 }
 
+impl TransactionBuilder4844 for FoundryTransactionRequest {
```
mablr, line 494: This is subject to change after the release of Alloy v2.0.

crates/primitives/src/network/wallet.rs
```diff
@@ -0,0 +1,69 @@
...
+                Ok(FoundryTxEnvelope::Eip4844(tx.into_signed(sig)))
+            }
+            FoundryTypedTx::Eip7702(mut tx) => {
+                let sig = sign_with_wallet(self, sender, &mut tx).await?;
+                Ok(FoundryTxEnvelope::Eip7702(tx.into_signed(sig)))
+            }
+            FoundryTypedTx::Deposit(tx) => {
+                // Deposit transactions don't require signing
+                Ok(FoundryTxEnvelope::Deposit(Sealed::new(tx)))
+            }
```
mablr, line 50: No error/panic here, according to the current foundry’s implementation, which is the opposite of op_alloy’s implementation. Let me know if you want to change this.

PR #13256, opened 2026-01-29
feat(forge): visualize traces in speedscope with --evm-profile

DaniPopes:
Just like `--flamegraph`/`--flamechart`, but opens a [speedscope.app](https://www.speedscope.app) window which has a more friendly UI to navigate the profile.

More info available at [speedscope#usage](https://github.com/jlfwong/speedscope#usage).
In our case, the "CPU time" is gas; `--flamechart` maps to the "Time Order" tab (default), `--flamegraph` maps to "Left heavy".

These screenshots are from a temporary deep-call fixture generated for visual verification (`DeepProfileTest::testDeepProfile`: 56 frames, max open stack depth 14). The Time Order screenshot is from the trace-order profile with 2,904 events.

### Speedscope: Time Order

Speedscope Time Order view for a deep EVM trace

### Speedscope: Left Heavy

Speedscope Left Heavy view for a deep EVM trace

### Speedscope: Sandwich

Speedscope Sandwich view for a deep EVM trace

### Existing `--flamegraph` comparison

Existing Forge flamegraph for the same deep EVM trace


emmajam:

crates/forge/src/cmd/test/mod.rs
```diff
@@ -1183,6 +1222,44 @@ impl TestArgs {
...
+                outcome.remove_first().ok_or_eyre("no tests were executed")?;
+
+            let (_, arena) = test_result
+                .traces
+                .iter_mut()
+                .find(|(kind, _)| *kind == TraceKind::Execution)
+                .unwrap();
+
+            // Decode traces.
+            let decoder = outcome.last_run_decoder.as_ref().unwrap();
```
grandizzy: Both `unwrap()`s panic on normal input:
- **No Execution trace** when `setUp()`/constructor reverts (only `Setup`/`Deployment` traces exist, but `num_filtered == 1` still passes). Repro: `forge test --mt testFoo --evm-profile` with a reverting `setUp()`.
- **`last_run_decoder` is `None`** with `--json`/`--junit` (those return early before it's set, and `--evm-profile` doesn't conflict with them). Repro: `forge test --mt testFoo --json --evm-profile`.

```suggestion
            let (_, arena) = test_result
                .traces
                .iter_mut()
                .find(|(kind, _)| *kind == TraceKind::Execution)
                .ok_or_eyre("cannot generate EVM profile: selected test has no execution trace")?;

            // Decode traces.
            let decoder = outcome
                .last_run_decoder
                .as_ref()
                .ok_or_eyre("cannot generate EVM profile: trace decoder was not initialized")?;
```

grandizzy (changes requested):
Please see finding below.

crates/evm/traces/src/speedscope/builder.rs
```diff
@@ -0,0 +1,224 @@
...
+        for name in stack.iter().skip(common_len) {
+            let frame_idx = *frame_cache
+                .entry(name.as_str())
+                .or_insert_with(|| file.add_frame(Frame::new(Cow::Owned(name.clone()))));
+            profile.open_frame(frame_idx, cumulative_gas);
+            open_stack.push(frame_idx);
+        }
+
+        // Advance cumulative gas by this frame's gas consumption.
+        cumulative_gas += gas;
```
mablr: Are we sure the evented Time Order view can be built from the folded-stack entries here? `build_entries()` returns child-subtracted aggregate self gas per stack, so any self gas spent in the parent before/between/after child calls gets emitted as one block before the first child frame. In a small repro like `child.first(); <inline parent work>; child.second();`, speedscope shows the parent work before `child.first()`, even though the call happened first.

crates/forge/src/cmd/test/mod.rs
```diff
@@ -169,6 +179,27 @@ pub struct TestArgs {
...
+    /// Implies `--decode-internal`.
+    #[arg(
+        long,
+        value_name = "FORMAT",
+        num_args = 0..=1,
+        default_missing_value = "speedscope",
+        value_enum,
+        conflicts_with_all = ["flamegraph", "flamechart"]
+    )]
+    evm_profile: Option<EvmProfileFormat>,
```
figtracer, line 195: this should be rejected under `--machine`: it prints human output and then waits on the profile server, so a selected test can hang before the final envelope. please add `--evm-profile` to `reject_machine_unsupported_flags()` and cover it in the machine-mode test.

crates/forge/src/cmd/test/mod.rs
DaniPopes: this is copy pasted from surrounding code, it doesn't matter

crates/evm/traces/src/speedscope/builder.rs
DaniPopes: ~~i don't think this is representable in the ui~~

crates/forge/src/cmd/test/mod.rs
figtracer: will be removed

crates/evm/traces/src/speedscope/builder.rs
DaniPopes: Good catch. I changed the speedscope builder to walk the trace arena order directly instead of reconstructing Time Order from folded/self-gas entries.

The small repro below is `parent work -> child.first() -> parent work -> child.second() -> parent work`.

Before, folded entries aggregate the parent self gas and place the second child immediately after the first:

Before: folded entries place the second child immediately after the first child in speedscope Time Order

After, trace-order emission keeps the second child at its actual later position:

After: trace-order emission preserves the later second child call in speedscope Time Order

mablr: Nice ! 🙌

crates/forge/src/cmd/test/mod.rs
mablr: sooon🫡

mablr (dismissed):

crates/forge/src/cmd/test/mod.rs
```diff
@@ -1472,7 +1549,7 @@ impl TestArgs {
         let config = Arc::new(config);
         let showmap = self.showmap_config();
         let runner = MultiContractRunnerBuilder::new(config.clone())
-            .set_debug(execution.should_debug)
+            .set_debug(execution.should_debug || self.evm_profile.is_some())
```
stevencartavia: `set_debug(... || evm_profile.is_some())` makes `trace_mode()` resolve to `TraceMode::Debug`. `Debug::into_config()` records all opcode steps (`record_opcodes_filter: None`) and enables memory snapshots, full stack snapshots, returndata snapshots, immediate bytes, and state diffs. For step data the builder only reads `DecodedTraceStep::InternalCall` markers and `step.gas_cost` — it doesn't read memory/stack/returndata/immediate.

In a local 60k-iteration loop test with the same reported gas (5,448,500), `--flamechart` peaked around 136 MB RSS and `--evm-profile` around 719 MB.

The builder does need all-opcode steps for accurate internal-function gas: with JUMP/JUMPDEST-only tracing, non-JUMP `gas_cost` consumed while an internal frame is open would be missed and backfilled to the enclosing EVM call instead (via `gas_used.saturating_sub(consumed)`). But the snapshots and state diff are unused.

Because `set_debug` feeds `tcfg.debug`, it also disables the setUp-trace downgrade in `runner.rs` (gated on `!self.tcfg.debug`), so `setUp` is traced at debug level even though the profile only consumes the `Execution` trace.

crates/forge/src/cmd/test/evm_profile_server.rs
```diff
@@ -0,0 +1,118 @@
...
+    let title = format!("{contract_name}::{test_name}");
+
+    let encoded_url = percent_encode(&profile_url);
+    let encoded_title = percent_encode(&title);
+    let viewer_url =
+        format!("https://www.speedscope.app/#profileURL={encoded_url}&title={encoded_title}");
+
+    sh_println!("Profile server running at http://127.0.0.1:{port}")?;
+
+    if no_open {
```
stevencartavia: With `--no-open` we skip `opener::open` and print the speedscope URL, but `serve_and_open` still unconditionally awaits `axum::serve(...)`, so the process stays alive until interrupted even though the standalone `cache/evm_profile_*.json` has already been written before the await. If `--no-open` means "write the profile but don't launch a viewer", skip starting the server — e.g. guard the `serve_and_open(...)` call with `if !self.no_open`, or `return Ok(outcome)` after writing (note `run_tests` returns `Result<TestOutcome>`, not `Result<()>`). Then update the help text from "the server is started" to say it saves the JSON and exits.

crates/evm/traces/src/folded_stack_trace.rs
```diff
@@ -7,11 +7,23 @@ use revm_inspectors::tracing::{
...
     let mut fst = EvmFoldedStackTraceBuilder::new(isolate);
-    fst.process_call_node(arena.nodes(), 0);
+    if !arena.nodes().is_empty() {
+        fst.process_call_node(arena.nodes(), 0);
+    }
     fst.build()
 }
 
+/// Builds trace entries from a call trace arena.
+pub fn build_entries(arena: &CallTraceArena, isolate: bool) -> Vec<TraceEntry> {
```
stevencartavia: `build_entries`, `EvmFoldedStackTraceBuilder::into_entries`, and `FoldedStackTraceBuilder::into_entries` appear unused in-repo; the speedscope builder walks the call trace arena directly rather than going through folded-stack entries. Deleting all three still builds (`forge`) and passes the traces tests. Remove them and revert `TraceEntry` to private type/fields; `Clone` (added in this PR) also looks unnecessary if it was only added for this. Keep the `saturating_sub` change and its test — they're independent of the dead APIs.

DaniPopes:
woops, auto merged while trying to stack prs. re-opened https://github.com/foundry-rs/foundry/pull/15394

crates/forge/src/cmd/test/mod.rs
DaniPopes: good point https://github.com/foundry-rs/foundry/pull/15393

PR #13258, opened 2026-01-29
fix(verify): respect user etherscan config URL when verifier-url not …

Noisyxl:
Fixes etherscan config URL being ignored by replacing no-op `.or(None)` with proper fallback to `etherscan_config.api_url`.


Noisyxl:
> please rebase, ty
already done there https://github.com/foundry-rs/foundry/pull/13275 

ty!

PR #13260, opened 2026-01-29
chore(config): `curl` mode as config key

### LINKED ISSUE #13252: `--curl` flag silently ignored in several Cast commands
## Problem

The `--curl` flag is available but silently ignored in many Cast subcommands. For example:
- `cast balance --curl`
- `cast block --curl` 
- `cast chain --curl`
- And many others...

While it's not always relevant to display curl command as most of the cast subcommands process the RPC responses, it feels weird to provide the `curl` flag but ignore it silently. 

## Root Cause

The `curl` field in `RpcOpts` is parsed from the CLI but never propagated to the `Config` struct. It is passed inline through helper function parameters:
https://github.com/foundry-rs/foundry/blob/2b85d1fbd3647865efdae4c0e17b994638ff722c/crates/cli/src/utils/mod.rs#L111
which means commands that didn't explicitly use this helper would silently ignore curl flag.

## Proposed solution

- Add `eth_rpc_curl: bool` field to the `Config` struct
- Update `get_provider_builder()` to read `curl_mode` directly from config instead of accepting it as a parameter
- Remove the `get_provider_with_curl()` helper entirely
- Simplify call sites to use `get_provider(&config)?` consistently

The benefits:
- All commands now respect the `--curl` flag uniformly
- Simplified provider builder helper (better maintainability)

I’ve already prepared the patch, but first please let me know if you find it relevant.

mablr:
## Motivation

Close #13252 

## Solution

- Provide `--curl` consistently in all subcommands
- Simplified helpers for provider instantiation using config
- Adjusted the `signing_provider_with_curl` function to `get_provider_with_wallet` for consistency.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes (always consider `--curl`, no more implicit ignore)

### DIFF
--- a/crates/common/src/transactions.rs
+++ b/crates/common/src/transactions.rs
@@ -161,7 +161,7 @@ pub fn get_pretty_tx_receipt_attr(
         "gasUsed" | "gas_used" => Some(receipt.receipt.gas_used.to_string()),
         "logs" => Some(receipt.receipt.inner.inner.inner.receipt.logs.as_slice().pretty()),
         "logsBloom" | "logs_bloom" => Some(receipt.receipt.inner.inner.inner.logs_bloom.pretty()),
-        "root" | "stateRoot" | "state_root " => Some(receipt.receipt.state_root().pretty()),
+        "root" | "stateRoot" | "state_root" => Some(receipt.receipt.state_root().pretty()),
         "status" | "statusCode" | "status_code" => {
             Some(receipt.receipt.inner.inner.inner.receipt.status.pretty())
         }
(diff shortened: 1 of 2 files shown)


mablr:
> * Use `CurlRpcOpts` instead of `RpcOpts` for all subcommands to which we want to provide the `--curl` flag

After looking a bit deeper, I realize this is not very clean because `RpcOpts` are often wrapped in `EthereumOpts`.

Config key approach may be actually the "less worse"...

PR #13263, opened 2026-01-29
fix(verify): respect user-configured Etherscan API URL from config

Aleksandr1732:
Fixes Etherscan client initialization to respect user-configured API URLs from `foundry.toml` config.


mattsse (commented):
Not accepting low-value changes.

PR #13265, opened 2026-01-29
chore(deps): bump alloy-core crates to 1.5.4

echowandere:
Update alloy-primitives and related alloy-core crates from 1.5.2 to 1.5.4. This brings in the asm-optimized keccak implementation from upstream


DaniPopes:
Thanks but would rather update weekly as normal

PR #13267, opened 2026-01-29
fix(anvil): return error when querying future block number in with_database_at

thunggis:
I noticed that `with_database_at` has a weird behavior - if you call something like `eth_getStorageAt` or `eth_call` with a block number that doesn't exist yet, it just returns the current state without any warning or error. So if you accidentally pass a wrong block number or there's some off-by-one issue in your code, you'd get back data that looks totally valid but is actually from a different block. That's pretty misleading and could cause some head-scratching bugs.                                                                                               
                                                                                                                                                                                                                                                                                             Turns out `ensure_block_number` already handles this correctly by returning `BlockOutOfRange`, so I just added the same check here. Now if you ask for a future block, you get a proper error instead of wrong data.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10329,9 +10329,9 @@ dependencies = [
 
 [[package]]
 name = "time"
-version = "0.3.46"
+version = "0.3.47"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


mattsse (approved):
this makes sense

thunggis:
> this makes sense

Thanks, should we merge the CI issue in https://github.com/foundry-rs/foundry/pull/13325 firstly?

PR #13271, opened 2026-01-30
feat(debugger): display actual gas usage alongside refund counter

### LINKED ISSUE #13262: Foundry trace & debugger use skewed "gas used" value with already deducted refund
### Component

Forge

- [x] Foundry
- [x] Foundryup

1.5.1-stable

1.5.0

forge test 

### Operating System

macOS (Apple Silicon)

### Describe the bug

I've been investigating foundry's gas usage metrics as I've noticed major divergences between the values that both debugger & trace share, the ones shown in the flamechart and roughly what the values should be according to my intuition. (Note this all concerns gas values in `--isolate` mode)

After investigating the root cause it seems to be that the values that the debugger & trace use come from the tracing revm-inspector `CallTraceStep.gas_used` which is computed from the "actual" gas used (`interp.gas.spent()`) with the refund preemptively subtracted out (via the `gas_used` helper function in revm-inspectors). This is unlike the execution spec where refunds are only subtracted out at the very end of the top-level transaction. This may be intentional but can be unintuitive especially for profiling because when the refund exceeds the cap you incur some "refund loss" which is slowly gained back as usage increases, it also leads to weird numbers in the trace where sub-calls can be shown as consuming more gas than the total of the parent.

As a user I'd want the trace to either only show the "actual" underlying gas used value by default in both the trace, debugger & flamegraph. Additionally if possible it'd be nice to visualize the gas refund somehow. My suggestion is:
- debugger: display actually spent gas as "gas used" and optionally a `refund_total` along gas used
- trace: display `[spent, refund]` or `[<spent>] .... (refund total: <refund_total>`
- flamegraph: add nodes for `SSTORE` operations with their gas cost being `initial_gas_used - refund + refund_loss_proportional` where `refund_loss_proportional` is the total refund amount that's lost due the cap shared proportionally among the opcodes causing refunds, for `SSTORE` operations that reverse refunds at the reversed refund to that node, this will cause certain `SSTORE` nodes in the graph to look like they cost more gas than they should but I think it'd be useful to visualize that for the sake of profiling

 Even after fixing this locally for the debugger I still cannot explain why the gas values shown in the flamegraph show a large gap with what seems to be the actual values now, even if you add the 21k intrinsic cost that `--isolate` adds after. If I find the root cause for that I will share.

Philogy:
## Motivation

Closes #13262 

## Solution
- Compute actual gas used in debugger by using gas limit + gas remaining instead of `gas_used` which prematurely deducted refund
- Added new "Gas refund" counter in debugger so that users can still see that information if they need
- Fix the `folded_stack_trace` generation to use actual gas for internal call diffs & re-add gas refunds to ensure gas values remain consistent

## PR Checklist

- [x] Added Tests (updated the existing tests)
- [x] Added Documentation (don't think it needs any)
- [x] Breaking changes (Flamechart & debugger now display pre-refund gas used)

### DIFF
--- a/crates/cast/src/lib.rs
+++ b/crates/cast/src/lib.rs
@@ -2336,7 +2336,7 @@ fn explorer_client(
     api_url: Option<String>,
     explorer_url: Option<String>,
 ) -> Result<Client> {
-    let mut builder = Client::builder().chain(chain)?;
+    let mut builder = Client::builder();
 
     let deduced = chain.etherscan_urls();
 
--- a/crates/evm/evm/src/executors/fuzz/mod.rs
+++ b/crates/evm/evm/src/executors/fuzz/mod.rs
@@ -280,7 +280,7 @@ impl FuzzedExecutor {
 
         if success {
             Ok(FuzzOutcome::Case(CaseOutcome {
-                case: FuzzCase { calldata, gas: call.gas_used, stipend: call.stipend },
+                case: FuzzCase { gas: call.gas_used, stipend: call.stipend },
                 traces: call.traces,
                 coverage: call.line_coverage,
                 breakpoints,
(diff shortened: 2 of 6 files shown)


DaniPopes:
needs some test updates and clippy

DaniPopes (approved):
thx bossman

PR #13272, opened 2026-01-30
fix(cast): remove redundant chain() call in explorer_client

tefyosL-sol:
Fixes similar issue as #13238, `explorer_client` was calling `.chain(chain)?` which fails for chains without default URLs even when custom URLs are provided via `--explorer-api-url` and `--explorer-url` flags.


DaniPopes (approved):
I see yeah .chain() calls .etherscan_urls() .with_api_url() .with_url() internally, but we override it after anyway. Thanks!

PR #13275, opened 2026-01-31
fix(verify): respect user-configured etherscan URL over chain defaults

tefyosL-sol:
Fixes similar issue as #13238, use etherscan URL from foundry.toml when  --verifier-url is not provided.  Previously only checked the command-line flag and fell back to chain defaults, skipping the config file entirely.


crates/verify/src/etherscan/mod.rs
mattsse, line 528: I see, this is now respected, this makes sense

mattsse (approved):
this makes sense, and matches a recent fix attempt I tried

lgtm, pending @grandizzy

PR #13277, opened 2026-01-31
fix(cast): include end block in chunked eth_getLogs queries

Aleksandr1732:
The chunked `eth_getLogs` implementation had an off-by-one error that excluded the final block in the requested range when fallback chunking was triggered.


onbjerg (changes requested):
please add a test, ty

emmajam:

PR #13283, opened 2026-02-01
perf(anvil): deduplicate block fetches in Otterscan search transactions

marukai67:
build_ots_search_transactions was calling block_by_number() once per transaction receipt to fetch timestamps. When multiple transactions belonged to the same block, this resulted in redundant RPC calls and unnecessary overhead.


mattsse (commented):
we can keep this as is

PR #13286, opened 2026-02-01
fix(traces): prevent verified contract metadata from being overwritten by unverified results

0xxFloki:
Fixes a bug in `ExternalIdentifier` where verified contract metadata could be incorrectly overwritten with `None` when multiple fetchers (Etherscan and Sourcify) return results in non-deterministic order.


DaniPopes:
I don't think this is right

PR #13290, opened 2026-02-02
feat(cast): add flashblocks flag that sets confirmations for transaction receipt to zero and reduces polling time

limyeechern:
## Motivation
Flashblocks-aware RPC nodes can include new transactions in pending blocks, providing transaction receipts much faster than traditional block confirmation times. Currently, `cast send` has no way to take advantage of this capability since it uses default polling intervals and waits for block confirmations, negating the latency benefits of flashblocks.

## Solution
Add a `--flashblocks` flag to `cast send` that optimizes transaction receipt polling for flashblocks-enabled environments:
- Faster polling: Reduces the poll interval from the default to 10ms, matching the rapid availability of flashblocks receipts
- Zero confirmations: Skips waiting for block confirmations since flashblocks provide immediate inclusion feedback

## Usage
`cast send --flashblocks <TO> <SIG> [ARGS]`

## Benchmark
Tested on a local blockchain with flashblocks (200ms block time). The chart below shows user-perceived latency with and without the `--flashblocks` flag:

### Methodology
Transactions are concurrently sent via the two methods below approximately every second. 
1. Vanilla
`time cast send <TO> <SIG> [ARGS]`

2. With `--flashblocks` flag
`time cast send --flashblocks <TO> <SIG> [ARGS]`

image

The `--flashblocks` flag reduces user-perceived latency by ~40–50% by polling more frequently (10ms vs default) and not waiting for block confirmations.

### DIFF
--- a/crates/cast/src/cmd/send.rs
+++ b/crates/cast/src/cmd/send.rs
@@ -124,7 +124,13 @@ impl SendTxArgs {
 
         if let Some(interval) = send_tx.poll_interval {
             provider.client().set_poll_interval(Duration::from_secs(interval))
+        } else if send_tx.eth.rpc.flashblocks {
+            // Use fast polling for flashblocks since receipts are available quickly
+            provider.client().set_poll_interval(Duration::from_millis(10))
         }
+
+        // When flashblocks is enabled set confirmations to 0
+        let tx_confirmations = if send_tx.eth.rpc.flashblocks { 0 } else { send_tx.confirmations };
 
         let builder = CastTxBuilder::new(&provider, tx, &config)
             .await?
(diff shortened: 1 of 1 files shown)


crates/cast/src/cmd/send.rs
```diff
@@ -175,7 +181,7 @@ impl SendTxArgs {
                 tx.into_inner(),
                 send_tx.cast_async,
                 send_tx.sync,
-                send_tx.confirmations,
+                tx_confirmations,
```
mattsse, line 184: most flashblocks aware endpoints should also support sync, which would avoid the massive polling here, can you do your testing with --sync as well?
imo that would be preferable to another flashblocks arg

limyeechern:
hey @mattsse, the `--sync` flag is the correct way to do this, closing this PR 

here's the test for your information, thanks for pointing it out!

image

PR #13301, opened 2026-02-03
fix(forge): add browser wallet support for forge create

### LINKED ISSUE #12958: bug(`forge create`): incorrect browser wallet signing
### Component

Forge

- [x] Foundry
- [x] Foundryup

1.5.1-stable

1.5.0

forge create src/Counter.sol:Counter --rpc-url http://127.0.0.1:8545 --broadcast --browser

### Operating System

None

### Describe the bug

Once the browser wallet is connected the following error happens:
```
Error: local usage error: Use send_transaction_via_browser for browser wallets
```

This issue is related to #12922, but I’m opening it here because it’s actually a bug, not a missing feature. It is going to be fixed as follow-up of this #12952.

MarkFizz77:
Fixes #12958                                                               

`forge create --browser` now works correctly. Previously it failed with "Use send_transaction_via_browser for browser wallets" error.  The fix detects browser wallet signers and uses `send_transaction_via_browser()` instead of wrapping them in `EthereumWallet`.

### DIFF
--- a/crates/forge/Cargo.toml
+++ b/crates/forge/Cargo.toml
@@ -50,7 +50,7 @@ forge-script.workspace = true
 forge-sol-macro-gen.workspace = true
 foundry-cli.workspace = true
 foundry-debugger.workspace = true
-foundry-wallets = { workspace = true, optional = true }
+foundry-wallets.workspace = true
 
 alloy-chains.workspace = true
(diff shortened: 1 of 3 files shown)


mablr:
Fixed in a different manner by #14394

Thanks for your effort!

PR #13321, opened 2026-02-04
feat(forge,chisel): realtime `console.log`

### LINKED ISSUE #9916: realtime console.log
### Component

Forge

waiting for tests to finish before showing any console.log is not a great experience

### Additional context

_No response_

quangloc99:
## Motivation

Resolves #9916, that is, make `console.log` messages appear during script/test/chisel execution.

## Solution

This PR follows the behavior of Rust's `cargo test` for `println!` closely. Here is what I found out about how `cargo test` handles output:

- To see `println!` output during test execution, you need to run `cargo test -- --no-capture`. So the default behavior is to _capture_ output during tests.
- Without `--no-capture`:
    - `cargo test` [calls `io::set_output_capture`](https://github.com/rust-lang/rust/blob/930ecbcdf8905c5c8549056c73fcabdd8d6e1b3d/library/test/src/lib.rs#L127-L159) to capture the output, which is a simple [`Vec<u8>` buffer](https://github.com/rust-lang/rust/blob/f60a0f1bcc5a2a6dd8eb2d2021da79b40f5fbae9/library/std/src/io/stdio.rs#L19-L26).
    - After each test, the captured output is printed only if the test fails.
- With `--no-capture`:
    - Seems like `cargo test` does not do anything to override the behavior of `println!`. This means tests can print everything out of orders since they are running in parallel. This can be confirmed with a simple test file.
    - When there is IO error, [the program just panics](https://github.com/rust-lang/rust/blob/f60a0f1bcc5a2a6dd8eb2d2021da79b40f5fbae9/library/std/src/io/stdio.rs#L1164-L1166). No further error handling is done.

Similar to `cargo test` without `--no-capture`, `console.log` messages are already captured in the `LogCollector`. This is very good as the behavior is very close to `cargo test`'s default behavior.

In this PR, I change the `LogCollector` from a struct into an enum with two variants: `Capture` and `LiveLogs`. I then update every broken parts after the change to make sure everything works as expected.
                                                                                       
By default `Capture` is used. To enable live logs, I added a new config for that, and then plug it into **every** place that `ExecutorBuilder` is used, so I'm sure that I did not miss any place.

I think the PR works as expected. Though here are some considerations that I have thought of:

- For `chisel`, old `console.log` from previous commands are still printed, because of how chisel works. I still think it is acceptable, as the same behavior can still be observed if `-v` is passed in, or the user modifies files using `vm.writeLine`.
- For `test`, since they are running in parallel, the output may be interleaved, similar to `cargo test -- --no-capture`. But if the user also pass in `--show-progress`, the progress bar will not be updated properly. A simple fix is to make `--show-progress` and `--live-logs` mutually exclusive, but I did not do that in this PR. Let me know if you think it is necessary.
- Other type of log from DSTest won't be affected by `--live-logs`. There is already a [fallback](https://github.com/foundry-rs/foundry/blob/1bd687f0ddf8ee3b4fa3ddf3e6baeec161fa77ce/crates/evm/evm/src/executors/mod.rs#L1072) to print them after the execution.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/.github/FLAKY_TEST_ISOLATE_FAILURE_TEMPLATE.md
+++ b/.github/FLAKY_TEST_ISOLATE_FAILURE_TEMPLATE.md
@@ -1,0 +1,10 @@
+---
+title: "bug: flaky tests workflow failed (isolate)"
+labels: P-normal, T-bug
+---
+
+The nightly flaky tests workflow (with isolation mode enabled) has failed. This indicates external API rate limiting, RPC reliability issues, or other intermittent failures that may affect users.
+
+Check the [flaky tests workflow page]({{ env.WORKFLOW_URL }}) for details.
+
+This issue was raised by the workflow at `.github/workflows/test-isolate.yml`.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
-      - uses: crate-ci/typos@93cbdb2d23269548cf0db0f74d0bc6a09a3f0d5c # v1
+      - uses: crate-ci/typos@78bc6fb2c0d734235d57a2d6b9de923cc325ebdd # v1
 
   shellcheck:
     runs-on: depot-ubuntu-latest
@@ -140,7 +140,7 @@ jobs:
         with:
           toolchain: stable
       - uses: rui314/setup-mold@725a8794d15fc7563f59595bd9556495c0564878 # v1
-      - uses: taiki-e/install-action@650c5ca14212efbbf3e580844b04bdccf68dac31 # v2
+      - uses: taiki-e/install-action@1e67dedb5e3c590e1c9d9272ace46ef689da250d # v2
         with:
           tool: cargo-hack
       - uses: mozilla-actions/sccache-action@7d986dd989559c6ecdb630a3fd2557667be217ad # v0.0.9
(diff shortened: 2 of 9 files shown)


quangloc99:
Hi @onbjerg. Thanks for looking into this! 

My initial approach to this PR is to do the simplest thing and to demonstrate how this would be done. I think this is what I normally would expect if I call `console.log` when writing foundry script.

The suggestion on for printing out the logging context is nice too. I can probably change the logging format to something like:

```
<contract_name>.<function>: <log message>
```

`<function>` can be `setUp`, `run`, `test`, or from the `--sig` input.

I'll help adding this or your preferred format too. I can try comming up with a way to set this automatically. Or at worst I can set this manually when a contract function is invoked from foundry.

I just want to share that it will be more complicated than the current PR. What do you think?

quangloc99:
On another note, `cargo test -- --no-capture` does not print any context either, and the `println!` will be out of order because of parallelization. It can be tested like this:

```rs
fn main() {}

#[test]
fn test1() {
    for _ in 0..100 {
        println!("test1");
    }
}

#[test]
fn test2() {
    for _ in 0..100 {
        println!("test2");
    }
}

```

I think `--no-capture` for `cargo test` is normally intended for users that want to debug their code. We can make a similar case for `forge test`.

As for `forge script`, as it is single threaded, I think no context is needed.

onbjerg:
going to merge this so we can get general feedback; thanks for your contribution:)

PR #13327, opened 2026-02-04
perf(evm): wrap Executor.backend in Arc for copy-on-write cloning

gakonst:
## Summary

Wrap `Executor.backend` in `Arc<Backend>` with copy-on-write via `Arc::make_mut()`.

During parallel fuzzing, each worker clones the `Executor`. Previously this deep-cloned the entire `Backend` (CacheDB, JournaledState, state snapshots). This change makes cloning an `Executor` a cheap pointer copy for the backend, with automatic deep-clone only when a worker actually mutates state.

## Changes

- `Executor.backend`: `Backend` → `Arc<Backend>`
- `Executor::backend_mut()` uses `Arc::make_mut()` for copy-on-write
- `Executor::new()` and `clone_with_backend()` wrap backend in `Arc`

### DIFF
--- a/crates/forge/tests/cli/config.rs
+++ b/crates/forge/tests/cli/config.rs
@@ -157,6 +157,14 @@ lint_on_build = true
 mixed_case_exceptions = [
     "ERC",
     "URI",
+    "ID",
+    "URL",
+    "API",
+    "JSON",
+    "XML",
+    "HTML",
+    "HTTP",
+    "HTTPS",
 ]
 
 [doc]
@@ -1362,7 +1370,15 @@ forgetest_init!(test_default_config, |prj, cmd| {
     "lint_on_build": true,
     "mixed_case_exceptions": [
       "ERC",
-      "URI"
+      "URI",
+      "ID",
+      "URL",
+      "API",
+      "JSON",
+      "XML",
+      "HTML",
+      "HTTP",
+      "HTTPS"
     ]
   },
   "doc": {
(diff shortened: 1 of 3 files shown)


crates/forge/tests/cli/config.rs
```diff
@@ -157,6 +157,14 @@ lint_on_build = true
...
     "ERC",
     "URI",
+    "ID",
+    "URL",
+    "API",
+    "JSON",
+    "XML",
+    "HTML",
+    "HTTP",
+    "HTTPS",
```
zerosnacks:
```suggestion
```

crates/forge/tests/cli/config.rs
```diff
@@ -1362,7 +1370,15 @@ forgetest_init!(test_default_config, |prj, cmd| {
...
-      "URI"
+      "URI",
+      "ID",
+      "URL",
+      "API",
+      "JSON",
+      "XML",
+      "HTML",
+      "HTTP",
+      "HTTPS"
```
zerosnacks, line 1381:
```suggestion
```

zerosnacks:
cc @DaniPopes I think this make sense but curious to hear what you think

DaniPopes:
make sense
what about CowBackend?

zerosnacks:
> make sense what about CowBackend?

To my understanding `CowBackend` avoids cloning per fuzz iteration within a worker, the `Arc<Backend>` avoids cloning per worker when the Executor is cloned for rayon threads

PR #13330, opened 2026-02-05
chore(anvil): remove unnecessary clone operations

MarkFizz77:
Small simplification as `.clone().iter()` is redundant when we can directly use `.iter()`, and `.as_ref().map(|x| x.clone())` can be simplified to `.as_ref().cloned()`.


MarkFizz77:
flaky

PR #13331, opened 2026-02-05
skip checksum hash in create2 mining when case-insensitive

DanielGuupta:
## Motivation
The checksum format calls keccak256 internally, but we only need it when --case-sensitive is set. 

## Solution

Switched to plain hex encoding for the default case, only use checksum when --case-sensitive is specified.
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -6870,6 +6870,156 @@ mod tests {
         });
     }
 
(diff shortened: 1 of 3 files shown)


zerosnacks (approved):
makes sense

PR #13332, opened 2026-02-05
fix(anvil): handle disk cache write failures in state eviction

thunggis:
Fix state loss when `disk_cache.write()` fails during memory eviction in `InMemoryBlockStates`.

- Make `DiskStateCache::write()` synchronous and return success status
- Restore state to memory on write failure instead of losing it
- Only track state in `on_disk_states` after confirmed successful write

### DIFF
--- a/.github/FLAKY_TEST_ISOLATE_FAILURE_TEMPLATE.md
+++ b/.github/FLAKY_TEST_ISOLATE_FAILURE_TEMPLATE.md
@@ -1,0 +1,10 @@
+---
+title: "bug: flaky tests workflow failed (isolate)"
+labels: P-normal, T-bug
+---
+
+The nightly flaky tests workflow (with isolation mode enabled) has failed. This indicates external API rate limiting, RPC reliability issues, or other intermittent failures that may affect users.
+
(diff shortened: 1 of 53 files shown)


thunggis:
@onbjerg hi there, could this be checked when available? thanks!

thunggis:
> i'm a little worried about the fallback here - i don't think it's necessarily desirable that we roll back. it's either ok that the write failed and we continue (current behavior), or this is a crit and the node stops. can you provide a rationale for the rollback?

`clear_into_state_snapshot()` destructively moves data out of memory. If the subsequent disk write fails, the state is lost from both memory and disk.

For example, with the old code:

    let state_snapshot = state.0.clear_into_state_snapshot(); // memory cleared
    self.disk_cache.write(hash, state_snapshot);               // write fails silently
    self.on_disk_states.insert(hash, state);                   // still tracked as "on disk"

Later when `get_on_disk_state()` tries to load it:

    if let Some(state) = self.on_disk_states.get_mut(hash)  // found (was inserted above)
        && let Some(cached) = self.disk_cache.read(*hash)   // None (write failed)
    {
        ...
    }
    None  // silent data loss

The rollback simply prevents this by restoring the state to memory when the disk write fails.

PR #13333, opened 2026-02-05
fix: support custom Etherscan URL in cast/forge commands

tefyosL-sol:
Multiple commands (cast storage, cast creation-code, forge clone, etc.) were using Client::new() which ignores custom Etherscan URLs in config, changed to Client::builder() pattern to respect user configuration.


emmajam:

PR #13335, opened 2026-02-05
foundryup: tempo now distributes all binaries

gakonst:
Tempo now distributes all foundry binaries (forge, cast, anvil, chisel), so the constraint limiting the tempo network to only forge and cast is no longer needed.

This removes the `BINS=(forge cast)` override that was previously applied when `--network tempo` was used.

### DIFF
--- a/foundryup/foundryup
+++ b/foundryup/foundryup
@@ -55,11 +55,6 @@ main() {
     esac; shift
   done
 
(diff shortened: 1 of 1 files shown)


gakonst:
Recreating PR from correct fork with rebased branch

PR #13342, opened 2026-02-05
feat(test): `--brutalize` to assess robustness of assembly blocks

zerosnacks:
## Summary

Adds a standalone source-level brutalizer sanitizer accessible via `forge test --brutalize`, inspired by [Solady's Brutalizer.sol](https://github.com/Vectorized/solady/blob/main/test/utils/Brutalizer.sol).

The brutalizer rewrites source files in a temporary workspace before compiling and running tests. Any test failure indicates a robustness issue in the code under test.

## What it does

- **Value brutalization** — XORs type-cast expressions with a deterministic per-site mask to dirty unused bits of sub-256-bit types (`address`, `uint8`, `bytes4`, etc.)
- **Memory brutalization** — Injects inline assembly at external function entry points to dirty scratch space (0x00-0x3f) and fill 1 KB beyond the free memory pointer with deterministic junk
- **FMP misalignment** — Misaligns the free memory pointer by a small deterministic odd offset (1-31 bytes) to catch assembly that assumes word-aligned pointers

## Changes

```
+423 / -1 across 4 files

New:
  crates/forge/src/brutalizer/mod.rs    — standalone single-pass source rewriter
  crates/forge/src/brutalizer/utils.rs  — deterministic hashing utilities (splitmix64, span_seed)

Modified:
  crates/forge/src/cmd/test/mod.rs      — --brutalize flag (conflicts_with mutate)
  crates/forge/src/lib.rs               — pub mod brutalizer
```

Zero changes to the mutation framework.

## Design decisions

- **XOR instead of OR** — matches Solady's approach, can both set and clear bits
- **Deterministic via splitmix64** — same source produces same mutations, fully reproducible
- **keccak256 chain for memory fill** — compact injected assembly with cryptographic-quality distribution
- **Type-known only** — only explicit type casts are mutated (no heuristic name matching)
- **Bool skipped** — non-canonical truthy values cannot be expressed at the Solidity source level
- **Assembly-gated memory mutations** — only functions containing inline assembly are targeted
- **External only** — `public` functions excluded from memory/FMP brutalization due to ambiguous call semantics (internal JUMP shares caller memory)
- **Not a mutator** — brutalizers measure *code quality* (robustness) while mutators measure *test quality* (detection); keeping them separate avoids conflating the two

## Testing

- `make lint` passes (fmt + clippy + typos)
- `--brutalize` and `--mutate` are mutually exclusive via clap


crates/forge/src/mutation/mutators/brutalizer_mutator.rs
```diff
@@ -0,0 +1,499 @@
...
+    format!(
+        " assembly {{ \
+        mstore(0x00, 0x{w0:016x}{w1:016x}) \
+        mstore(0x20, 0x{w2:016x}{w3:016x}) \
+        let _b_p := mload(0x40) \
+        mstore(_b_p, 0x{w0:016x}{w2:016x}) \
+        mstore(add(_b_p, 0x20), 0x{w1:016x}{w3:016x}) \
+        }} "
+    )
+}
```
zerosnacks: Note for me, this is a good beginning but this is just scratch space; needs to be expanded

zerosnacks: Fixed, now uses copy trick to dirty 1KB of memory

grandizzy (commented):
Do we have configs to include / exclude mutators? Would it make sense to have this one as an advanced one that could be turned off for shorter runs?

zerosnacks:
> Do we have configs to include / exclude mutators? Would it make sense to have this one as an advanced one that could be turned off for shorter runs?

Yes we should and this could be an advanced one; I'll propose a PR for the config

zerosnacks:
RFR but marked as blocked because I don't think it should be merged into the mutation PR

crates/forge/src/brutalizer/value.rs
```diff
@@ -0,0 +1,176 @@
...
+pub(super) fn deterministic_mask(span: Span) -> String {
+    let h = span_seed(span);
+    let mask = if h == 0 { 1 } else { h };
+    format!("0x{mask:016x}")
+}
+
+pub(super) fn brutalize_cast(ty: &Type<'_>, arg_text: &str, mask: &str) -> Option<String> {
+    match &ty.kind {
+        TypeKind::Elementary(elem_ty) => match elem_ty {
+            ElementaryType::Address(_) => Some(brutalize_address(arg_text, mask)),
```
figtracer: this drops the payable bit: `payable(a)` is rewritten through `address(...)`, so uses like `payable(a).transfer(...)` can stop compiling. please preserve `address payable` here, e.g. emit `payable(address(...))` for payable casts and add a regression test.

crates/forge/src/brutalizer/value.rs
```diff
@@ -0,0 +1,176 @@
...
+}
+
+fn brutalize_int(size: TypeSize, arg_text: &str, mask: &str) -> Option<String> {
+    let bits = size.bits_raw();
+    let actual_bits = if bits == 0 { 256 } else { bits };
+    if actual_bits >= 256 {
+        return None;
+    }
+    Some(format!(
+        "int{actual_bits}(int256(int{actual_bits}({arg_text})) | int256(uint256({mask}) << {actual_bits}))"
```
figtracer: for signed casts, `|` can leave negative values unbrutalized because `int{n}(-1)` is already sign-extended with ones above `n` bits. using `^` for the dirty region would actually flip high bits while preserving the final `int{n}` value; please cover `int16(-1)` through assembly.

crates/forge/src/brutalizer/value.rs
```diff
@@ -0,0 +1,176 @@
+use solar::ast::{ElementaryType, Span, Type, TypeKind, TypeSize};
+
+use super::span_seed;
+
+pub(super) fn deterministic_mask(span: Span) -> String {
+    let h = span_seed(span);
+    let mask = if h == 0 { 1 } else { h };
```
figtracer, line 7: this only guarantees the 64-bit mask is nonzero, not the effective dirty mask. for `uint248` or `bytes31`, the surviving low byte can be zero, so the rewrite may not dirty anything; please constrain to the unused-bit width and force that value nonzero.

crates/forge/src/workspace.rs
```diff
@@ -32,6 +33,98 @@ pub fn relative_to_root(root: &Path, path: &Path) -> PathBuf {
...
+
+    if let Some(path) = &config.fuzz.failure_persist_dir {
+        temp_config.fuzz.failure_persist_dir =
+            Some(rebase_project_path(&config.root, temp_path, path));
+    }
+    if let Some(path) = &config.invariant.failure_persist_dir {
+        temp_config.invariant.failure_persist_dir =
+            Some(rebase_project_path(&config.root, temp_path, path));
+    }
+    for permission in &mut temp_config.fs_permissions.permissions {
```
figtracer, line 93: rebasing `fs_permissions` here makes project-local fixtures point into the temp workspace, but `copy_project` never copies those paths. a read-file test with `path = "./fixtures"` can pass normally and fail under `--brutalize` (and mutation now uses this helper too); please copy existing project-local permission targets.

figtracer (dismissed):
gaps

crates/forge/src/brutalizer/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+/// Brutalize all .sol source files in a temp project directory.
+///
+/// Walks the src directory under `temp_dir`, parses each .sol file, applies all
+/// brutalizations (value XOR, memory, FMP), and writes the result back in-place.
+///
+/// Returns the number of files brutalized.
+pub fn brutalize_project(config: &Config, temp_dir: &Path) -> eyre::Result<usize> {
+    let src_rel = config.src.strip_prefix(&config.root).unwrap_or(&config.src);
```
stevencartavia, line 53: With a flat layout (`src = "."`) `brutalize_project`'s walk resolves to the workspace root and rewrites test files and copied library sources, not just code-under-test — reproduced: lib-less flat project with `A.sol` + `A.t.sol` gives `Brutalized 2 source files` (the test file is rewritten too), ~22 with forge-std present. Subtree exclusion won't fix it since `test`/`src` coincide here; skip `libs` and filter test/script files at the file level, and add a `src = "."` regression test.

crates/forge/src/brutalizer/assembly.rs
```diff
@@ -0,0 +1,190 @@
...
+    let w1 = splitmix64(s.wrapping_add(1));
+    let w2 = splitmix64(s.wrapping_add(2));
+    let w3 = splitmix64(s.wrapping_add(3));
+    let s0 = splitmix64(s.wrapping_add(4));
+    let s1 = splitmix64(s.wrapping_add(5));
+    let s2 = splitmix64(s.wrapping_add(6));
+    let s3 = splitmix64(s.wrapping_add(7));
+    format!(
+        " assembly {{ \
+        mstore(0x00, 0x{w0:016x}{w1:016x}) \
```
stevencartavia: In `generate_memory_brutalization_assembly`, `mstore(0x00, 0x{w0:016x}{w1:016x})` is a 128-bit literal, so the top 16 bytes of each scratch slot stay zero — only the low half is dirtied (the beyond-FMP fill uses full words). Reproduced: an uninitialized `shr(128, mload(0x00))` read PASSES under `--brutalize` while the low-128 read FAILS. Fix: emit four 64-bit halves per mstore (8 total).

crates/forge/tests/cli/test_cmd/brutalize.rs
```diff
@@ -0,0 +1,597 @@
...
+    cmd.args(["test", "--mc", "MemVulnTest"]);
+    cmd.assert_success();
+
+    // Brutalized test fails — memory past FMP is filled with junk
+    cmd.forge_fuse().args(["test", "--brutalize", "--mc", "MemVulnTest"]);
+    cmd.assert_failure();
+});
+
+// Catches dirty scratch space: reading 0x00 without writing first
+forgetest_init!(brutalize_catches_dirty_scratch_space, |prj, cmd| {
```
stevencartavia, line 423: No test proves the value (dirty-bits) pillar catches anything — only memory/scratch have negative tests. It does work (`bytes4 y = bytes4(x); assembly { r := y }` passes normally, fails under `--brutalize`), so please add a negative test.

crates/forge/src/cmd/test/mod.rs
```diff
@@ -891,6 +894,24 @@ pub struct TestArgs {
...
     pub mutation_via_ir: Option<bool>,
+
+    /// Enable brutalization mode.
+    ///
+    /// Catches latent bugs that normal tests miss because the EVM initializes
+    /// memory to zero and registers to clean values. Applies source-level
+    /// sanitizers before compiling:
+    ///
+    /// - Dirties unused bits in sub-256-bit type casts (address, uint8, bytes4, etc.) to catch
+    ///   assembly code that assumes clean upper bits
```
stevencartavia: Value brutalization is silently ineffective under `--via-ir`: the same `bytes4` bug fails under `--brutalize` but passes under `--brutalize --via-ir`, because the outer cast wraps the dirt and via-IR's eager cleanup removes it (memory/FMP pillar still works). Emit a `sh_warn!` when `via_ir = true` and fix the doc bullet claiming casts catch "assembly code that assumes clean upper bits" to note it's legacy-codegen-only.

crates/forge/src/workspace.rs
```diff
@@ -85,7 +178,24 @@ pub fn copy_project(config: &Config, temp_dir: &Path) -> Result<()> {
...
+        copy_extra_project_path(
+            &config.root,
+            temp_dir,
+            extra_path,
+            &handled_extra_roots,
+            "include/allow",
+        )?;
+    }
+    for permission in &config.fs_permissions.permissions {
+        if !permission.is_granted(FsAccessKind::Read) {
```
stevencartavia: Only `Read`-granted `fs_permissions` paths are copied, so `write`-permission dirs aren't created in the temp workspace and `vm.writeFile` (which doesn't create parents) fails. Reproduced: a test writing to a `write`-permission dir passes normally but fails under `--brutalize` with "No such file or directory". Also create (empty) `Write`-granted dirs in the temp workspace.

crates/forge/src/workspace.rs
```diff
@@ -85,7 +178,24 @@ pub fn copy_project(config: &Config, temp_dir: &Path) -> Result<()> {
 
     let handled_extra_roots = handled_project_roots(config)?;
     for extra_path in config.include_paths.iter().chain(config.allow_paths.iter()) {
-        copy_extra_project_path(&config.root, temp_dir, extra_path, &handled_extra_roots)?;
+        copy_extra_project_path(
```
stevencartavia, line 206: `copy_dir_recursive` (used here for src/test) skips symlinked directories, but normal test discovery follows them (`WalkDir::follow_links(true)`), so with in-repo symlinked source dirs `--brutalize` copies nothing and silently passes while testing zero code. Reproduced: project with `src/core -> ../.shared/src` and `test/suite -> ../.shared/test` → `forge test` runs 2 tests and passes, but `forge test --brutalize` prints `Brutalized 0` / `Nothing to compile`, runs 0 tests, exits 0. Copy in-root symlinked dir contents (with `ensure_within_root` + cycle guard), or at minimum fail/warn when the temp workspace has zero `.sol` files while the real project has some.

crates/forge/tests/cli/test_cmd/brutalize.rs
mablr: https://github.com/foundry-rs/foundry/pull/13342/commits/92a6244c0938e822b12609dfe845440662db42dc

crates/forge/src/cmd/test/mod.rs
mablr: https://github.com/foundry-rs/foundry/pull/13342/commits/92a6244c0938e822b12609dfe845440662db42dc

crates/forge/src/workspace.rs
mablr: https://github.com/foundry-rs/foundry/pull/13342/commits/1db39294ad7267d6f3eac45b6ac51071761d49de

crates/forge/src/workspace.rs
mablr: https://github.com/foundry-rs/foundry/pull/13342/commits/d2aaef1db34daffe65d3e2aaa112d8c0a083f7d1

crates/forge/src/workspace.rs
```diff
@@ -313,11 +472,31 @@ fn process_nested_lib_dir(
...
+    dst: &Path,
+    allowed_root: &Path,
+    visited: &mut HashSet<PathBuf>,
+) -> Result<()> {
     if !src.exists() {
         return Ok(());
     }
+    ensure_within_root(allowed_root, src, "copied directory", src)?;
+    let canonical = src.canonicalize()?;
+    if !visited.insert(canonical) {
```
stevencartavia: This per-copy `visited` set skips the *destination* when a second symlink resolves to an already-copied canonical target: the `insert` returns before `create_dir_all(dst)`, so the second alias's dir is never created. Two in-repo symlinks to the same dir within one copied tree (both under `src`, or both under `test`) → that tree's second alias is dropped. Reproduced: `src/first -> ../.shared/pkg` and `src/nested/second -> ../../.shared/pkg` → `forge test` passes but `forge test --brutalize` fails with `Source "src/nested/second/Y.sol" not found`. Also hits mutation testing (shared `copy_project`). Use a recursion-stack guard (suppress only descent for paths on the current stack) instead of a whole-copy "already seen" set.

crates/forge/src/workspace.rs
```diff
@@ -218,6 +338,45 @@ fn copy_extra_project_path(
...
+        return Ok(());
+    }
+    ensure_safe_relative_path(&rel, "fs permission", path)?;
+    ensure_within_root(root, &resolved, "fs permission", path)?;
+
+    if is_covered_by_handled_root(&rel, handled_roots) {
+        return Ok(());
+    }
+
+    if resolved.exists() && resolved.is_dir() {
```
stevencartavia: `create_project_local_permission_dir` only materializes existing *directory* permissions. For a write-only permission targeting a file whose parent exists, `copy_project` creates nothing in the temp tree; the rebased permission stays in the config but the temp run doesn't honor it, because the checked write path is normalized through the temp root while the nonexistent permission target can't be canonicalized. Reproduced: `fs_permissions = [{ access = "write", path = "logs/sub/a.txt" }]` with `logs/sub/` existing → `forge test` passes but `forge test --brutalize` rejects `vm.writeFile` as "not allowed to be accessed for write operations". (Directory write perms and existing-file `read-write` perms are fine.) Materialize the file target's existing parent, and ensure the rebased permission path normalizes consistently for nonexistent file targets.

mablr (dismissed):
No more blockers found.

crates/forge/src/workspace.rs
```diff
@@ -77,15 +172,38 @@ pub fn copy_project(config: &Config, temp_dir: &Path) -> Result<()> {
...
+            )?;
+        } else if permission.is_granted(FsAccessKind::Write) {
+            create_project_local_permission_dir(
+                &config.root,
+                temp_dir,
+                &permission.path,
+                &handled_extra_roots,
+            )?;
+        }
     }
```
grandizzy, line 231: **High** — `read-write` fs permissions targeting a not-yet-existing file are silently dropped in the temp workspace. `FsAccessPermission::ReadWrite` grants `Read`, so it takes the first branch; `copy_project_local_permission_path` is a no-op when the target doesn't exist, and the `else if` prevents the `Write` branch from running, so the parent dir is never materialized.

Repro: `fs_permissions = [{ access = "read-write", path = "./logs/sub/a.txt" }]` with `logs/sub/` existing but `a.txt` absent → `forge test` passes, `forge test --brutalize` rejects the `vm.writeFile`. Same class as the prior write-only file-parent finding, via the read-write access kind.

```diff
     for permission in &config.fs_permissions.permissions {
         if permission.is_granted(FsAccessKind::Read) {
             copy_project_local_permission_path(
                 &config.root,
                 temp_dir,
                 &permission.path,
                 &handled_extra_roots,
             )?;
-        } else if permission.is_granted(FsAccessKind::Write) {
+        }
+        if permission.is_granted(FsAccessKind::Write) {
             create_project_local_permission_dir(
                 &config.root,
                 temp_dir,
                 &permission.path,
                 &handled_extra_roots,
             )?;
         }
     }
```

crates/forge/src/workspace.rs
```diff
@@ -218,6 +337,62 @@ fn copy_extra_project_path(
...
+    if parent.as_os_str().is_empty() || is_covered_by_handled_root(parent, handled_roots) {
+        return Ok(());
+    }
+
+    fs::create_dir_all(temp_dir.join(parent))?;
+    if resolved.exists() {
+        fs::copy(&resolved, temp_dir.join(rel))?;
+    } else {
+        fs::File::create(temp_dir.join(rel))?;
+    }
```
grandizzy, line 413: **Medium** — write-only permission to a nonexistent file materializes the file itself, not just its parent. The intent was to create the target's existing parent; fabricating an empty file changes behavior for tests using `writeFile`/`removeFile`/append or existence checks — the file pre-exists under `--brutalize` but not in a normal run.

```diff
     fs::create_dir_all(temp_dir.join(parent))?;
-    if resolved.exists() {
-        fs::copy(&resolved, temp_dir.join(rel))?;
-    } else {
-        fs::File::create(temp_dir.join(rel))?;
-    }
+    // Only materialize an existing file target; never fabricate a missing file,
+    // which would change existence/append/remove semantics under --brutalize.
+    if resolved.exists() && resolved.is_file() {
+        fs::copy(&resolved, temp_dir.join(rel))?;
+    }
 
     Ok(())
```

crates/forge/src/brutalizer/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+    for path in config.libs.iter().chain([&config.script]) {
+        if path.exists() && path != &config.src {
+            let rel = path.strip_prefix(&config.root).unwrap_or(path);
+            skipped_dirs.push(temp_dir.join(rel));
+        }
+    }
+    if config.test.exists() && config.test != config.src {
+        let test_rel = config.test.strip_prefix(&config.root).unwrap_or(&config.test);
+        skipped_dirs.push(temp_dir.join(test_rel));
+    }
```
grandizzy: **Medium** — flat `src = "."` layouts brutalize dependency sources. With `src = "."`, `src_dir == temp_dir` and the recursive walk covers the whole workspace. The skip set is built from `config.libs`/`script`/`test` only — but `node_modules` is in `libs` just under `--hardhat` and Soldeer's `dependencies/` is never in `libs`. So `dependencies/**` and (non-hardhat) `node_modules/**` get rewritten, surfacing failures in third-party code and inflating the file count.

Reuse the existing `handled_project_roots` helper (it already combines `src`/`test`/`script`/`libs` and the `node_modules`/`dependencies` pair, matching exactly what `copy_project` treats as dependency roots).

```diff
@@ crates/forge/src/workspace.rs
-fn handled_project_roots(config: &Config) -> Result<Vec<PathBuf>> {
+pub(crate) fn handled_project_roots(config: &Config) -> Result<Vec<PathBuf>> {
```

```diff
@@ crates/forge/src/brutalizer/mod.rs
-    let mut skipped_dirs = Vec::new();
-    for path in config.libs.iter().chain([&config.script]) {
-        if path.exists() && path != &config.src {
-            let rel = path.strip_prefix(&config.root).unwrap_or(path);
-            skipped_dirs.push(temp_dir.join(rel));
-        }
-    }
-    if config.test.exists() && config.test != config.src {
-        let test_rel = config.test.strip_prefix(&config.root).unwrap_or(&config.test);
-        skipped_dirs.push(temp_dir.join(test_rel));
-    }
+    // Skip everything copy_project treats as a dependency/non-source root
+    // (test, script, libs, node_modules, dependencies). Matters most for flat
+    // `src = "."` layouts where the walk covers the whole workspace.
+    let src_rel_owned = src_rel.to_path_buf();
+    let skipped_dirs: Vec<PathBuf> = crate::workspace::handled_project_roots(config)?
+        .into_iter()
+        .filter(|rel| !rel.as_os_str().is_empty() && *rel != src_rel_owned)
+        .map(|rel| temp_dir.join(rel))
+        .collect();
```

(`handled_project_roots` returns root-relative paths and filters empty via `is_covered_by_handled_root`; `src` maps to empty in flat layouts and is excluded here so sources are still brutalized.)

crates/forge/src/brutalizer/assembly.rs
```diff
@@ -0,0 +1,194 @@
...
+    let insert_span = Span::new(BytePos(insert_pos), BytePos(insert_pos));
+    vec![
+        Transform::Insert {
+            offset: insert_pos as usize,
+            replacement: generate_memory_brutalization_assembly(insert_span),
+        },
+        Transform::Insert {
+            offset: insert_pos as usize,
+            replacement: generate_fmp_misalignment_assembly(insert_span),
+        },
```
grandizzy: **Low** — two inserts at the identical offset have unspecified final order. `apply_transforms` sorts descending by start/end with a non-stable `sort_by`; both keys tie, and repeated `insert_str` at one offset reverses application order. So it's undefined whether the memory-fill loop reads the original FMP or the already-misaligned one — a semantic difference in what gets brutalized.

```diff
     let insert_pos = body.span.lo().0 + 1;
     let insert_span = Span::new(BytePos(insert_pos), BytePos(insert_pos));
-    vec![
-        Transform::Insert {
-            offset: insert_pos as usize,
-            replacement: generate_memory_brutalization_assembly(insert_span),
-        },
-        Transform::Insert {
-            offset: insert_pos as usize,
-            replacement: generate_fmp_misalignment_assembly(insert_span),
-        },
-    ]
+    // Single insert with a deterministic order: dirty memory first (reading the
+    // original FMP), then misalign the FMP. Two inserts at the same offset would
+    // have unspecified final order (non-stable sort + repeated insert_str).
+    let replacement = format!(
+        "{}{}",
+        generate_memory_brutalization_assembly(insert_span),
+        generate_fmp_misalignment_assembly(insert_span),
+    );
+    vec![Transform::Insert { offset: insert_pos as usize, replacement }]
```

grandizzy (changes requested):
Please see findings inline.

mablr (approved):

grandizzy (approved):
lgtm

PR #13360, opened 2026-02-06
fix(verify): correct Sourcify API URL construction for custom chains

marukai67:
Fixes incorrect URL construction in `sourcify_api_url()` that corrupted HTTPS scheme.


grandizzy:
> I believe this fix to be correct, we did not experience the issue because the `url` crate normalizes `https:/<link>` back to `https://<link>`.
> 
> pending additional check @grandizzy

makes sense

PR #13367, opened 2026-02-06
fix(fmt): correct indentation of named args nested inside chained calls

### LINKED ISSUE #13362: forge(fmt): Incorrect indentation
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.6.0-nightly Commit SHA: 099390013af6606c37c2034413d59b4b698b307d Build Timestamp: 2026-02-06T06:16:43.361599000Z (1770358603) Build Profile: dist

foundryup: 1.5.0

forge fmt

### Operating System

macOS (Apple Silicon)

### Describe the bug

`forge fmt` results in the following formatting:

```solidity
function initialize(
    V4PoolManager self,
    address currency0,
    address currency1,
    uint24 fee,
    int24 tickSpacing,
    uint160 sqrtPriceX96
) returns (int24 tick) {
    tick = IPoolManager(V4PoolManager.unwrap(self))
        .initialize(
            PoolKey({
            currency0: Currency.wrap(currency0),
            currency1: Currency.wrap(currency1),
            fee: fee,
            tickSpacing: tickSpacing,
            hooks: IHooks(address(0))
        }),
            sqrtPriceX96
        );
}
```

expected:

```solidity
function initialize(
    V4PoolManager self,
    address currency0,
    address currency1,
    uint24 fee,
    int24 tickSpacing,
    uint160 sqrtPriceX96
) returns (int24 tick) {
    tick = IPoolManager(V4PoolManager.unwrap(self))
        .initialize(
            PoolKey({
                currency0: Currency.wrap(currency0),
                currency1: Currency.wrap(currency1),
                fee: fee,
                tickSpacing: tickSpacing,
                hooks: IHooks(address(0))
            }),
            sqrtPriceX96
        );
}
```

Himess:
Fixes #13362

## Problem

When a struct literal with named arguments is used as an argument inside a chained method call, the fields lose their indentation:

```solidity
// before (wrong)
tick = IPoolManager(V4PoolManager.unwrap(self))
    .initialize(
        PoolKey({
        currency0: Currency.wrap(currency0),
        currency1: Currency.wrap(currency1),
        fee: fee,
        tickSpacing: tickSpacing,
        hooks: IHooks(address(0))
    }),
        sqrtPriceX96
    );
```

## Fix

`print_named_args` suppressed indentation whenever **any** ancestor in the call stack was a chained call with indent — even for deeply nested struct literals. The fix adds a `!self.call_stack.is_nested()` check so indentation is only suppressed when named args are **directly** part of the chain (e.g. `flow.withdraw{value: 1}`), not when nested inside another function call.

```solidity
// after (correct)
tick = IPoolManager(V4PoolManager.unwrap(self))
    .initialize(
        PoolKey({
            currency0: Currency.wrap(currency0),
            currency1: Currency.wrap(currency1),
            fee: fee,
            tickSpacing: tickSpacing,
            hooks: IHooks(address(0))
        }),
        sqrtPriceX96
    );
```


Philogy:
Change is incorrect, it makes calls with arguments get extra indented as shown in the tests you changed. Probably good idea to add a test case for the code I shared in the issue.

Himess:
@Philogy 
I've reworked the fix: instead of checking !is_nested(), it now `uses a named_args_directly_in_chain()` method that only suppresses indent when named args belong directly to the chained call (stack ends with `[Chained, Nested])`, not when they're in a deeper nested call like `PoolKey({...})` inside `.initialize(...)`. 
Also added a test case for the code from the issue. Existing tests are back to their original expectations.

crates/fmt/src/state/mod.rs
```diff
@@ -86,10 +86,18 @@ impl CallStack {
...
+    ///
+    /// This distinguishes:
+    /// - `TransferMessage({fields}).pack()` → true  (suppress extra indent)
+    /// - `.initialize(PoolKey({fields}), …)` → false (keep normal indent)
+    pub(crate) fn named_args_directly_in_chain(&self) -> bool {
+        let len = self.stack.len();
+        len >= 2 && self.stack[len - 1].is_nested() && {
+            let parent = &self.stack[len - 2];
+            parent.is_chained() && parent.has_indent
+        }
```
0xrusowsky: i think this works, however it is a little bit difficult to reason about.
instead, could we use this modified version which reads closer to the original one (clear intent), but with the proposed fix?


```rs
/// Returns true if the direct parent chain has its own indentation.
/// Used to determine if commasep should skip its own indentation (to avoid double indent).
pub(crate) fn has_indented_parent_chain(&self) -> bool {
    matches!(
        self.stack.as_slice(),
        [.., parent, last] if last.is_nested() && parent.is_chained() && parent.has_indent
    )
}
```

Himess: Make sense ty !
done https://github.com/foundry-rs/foundry/pull/13367/changes/a0db57100ed0cce866f6bd0582bf7e3580dee8db

mablr (approved):
lgtm, retested after 1.7.0 it works fine, pending @0xrusowsky or @zerosnacks

zerosnacks (approved):
Thanks @Himess, lgtm!

figtracer (approved):

PR #13370, opened 2026-02-07
 fix(forge): substraction underflow panic in LineNumberCache::get_position when offset precedes first line

cuiweixie:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/forge/src/coverage.rs
+++ b/crates/forge/src/coverage.rs
@@ -370,9 +370,74 @@ impl LineNumberCache {
         };
         let lo = match line_offsets.binary_search(&offset) {
             Ok(lo) => lo,
-            Err(lo) => lo - 1,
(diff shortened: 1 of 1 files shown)


onbjerg (changes requested):
please provide an example where this would be able to underflow as a test

cuiweixie:
> please provide an example where this would be able to underflow as a test

Done

onbjerg:
sorry, to be clear, can you provide an example where this would be hit in a real coverage run? I don't think there is a case where the line number cache would be in this state during a real coverage run.

cuiweixie:
> sorry, to be clear, can you provide an example where this would be hit in a real coverage run? I don't think there is a case where the line number cache would be in this state during a real coverage run.

try, but failed. So close it.

PR #13379, opened 2026-02-07
fix(fuzz): sort state changeset iteration for deterministic dictionary generation 

thunggis:
Sort `StateChangeset` iteration by address and storage by slot in `FuzzDictionary::insert_new_state_values`, matching the deterministic sorting already applied in `insert_db_values` and `EvmFuzzState::new`.

**How to reproduce**:

  The underlying `HashMap` uses `foldhash::fast::RandomState` (random seed per process).
  Run the same fuzz test with a fixed seed in two separate processes:

  ```bash
  forge test --match-test testFuzz_xxx --fuzz-seed 12345
  # exit, then run again:
  forge test --match-test testFuzz_xxx --fuzz-seed 12345

  The dictionary insertion order will differ between runs. After this fix, both runs
  produce identical ordering.
  ```


DaniPopes:
doesn't matter

PR #13380, opened 2026-02-07
perf(otterscan): fix pagination overflow and eliminate N+1 block queries

Aleksandr1732:
Fixes pagination overflow and reduces redundant RPC calls in Otterscan API endpoints.


mattsse:
Closing — the pagination fix is reasonable but the N+1 block query optimization adds significant complexity for negligible benefit in a local dev tool like anvil. The existing code already handles timestamps correctly via `block_timestamp()` with a fallback, and the added HashMap/dedup/separate-futures pattern is over-engineered for this context.

PR #13381, opened 2026-02-07
fix: correct byte offset in `first_non_whitespace` for multi-byte chars

VolodymyrBg:
## Motivation

`first_non_whitespace` https://github.com/foundry-rs/foundry/blob/b523fd29ad7e1fe872005a2dacd806a3a1b73fc8/crates/common/src/comments/mod.rs#L349-L361 used the formula `i + 1 - len` to compute the byte offset of the first non-whitespace character. This is wrong when the last whitespace char before content is multi-byte (e.g. NBSP U+00A0, 2 bytes): the result points into the middle of a UTF-8 character, and the subsequent `&s[offset..]` slice in `normalize_block_comment_ws` https://github.com/foundry-rs/foundry/blob/b523fd29ad7e1fe872005a2dacd806a3a1b73fc8/crates/common/src/comments/mod.rs#L363-L374 panics.

For ASCII whitespace (spaces, tabs) the bug is masked because `len` is always 1, so `i + 1 - 1 == i`.

## Solution

Replace the manual loop with `s.char_indices().find(…).map(…)` — `i` from `char_indices()` is already the correct byte offset. Add regression tests covering both ASCII and multi-byte whitespace inputs.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/common/src/comments/mod.rs
+++ b/crates/common/src/comments/mod.rs
@@ -349,15 +349,7 @@ fn all_whitespace(s: &str, col: CharPos) -> Option<usize> {
 /// Returns `Some(k)` where `k` is the byte offset of the first non-whitespace char. Returns `k = 0`
 /// if `s` starts with a non-whitespace char. If `s` only contains whitespaces, returns `None`.
 fn first_non_whitespace(s: &str) -> Option<usize> {
-    let mut len = 0;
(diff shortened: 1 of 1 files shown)


emmajam:

PR #13387, opened 2026-02-08
refactor(anvil): deduplicate RpcError constructor logic

sashaodessa:
Three `RpcError` constructors (`invalid_params`, `internal_error_with`, `transaction_rejected`) duplicated the same struct initialization pattern, including the redundant `message.into().into()` conversion chain. This duplication makes the code harder to maintain and increases the risk of inconsistent changes when modifying `RpcError` initialization logic.


mattsse (commented):
no

PR #13389, opened 2026-02-08
perf(config): skip redundant remapping detection in _with_root

### LINKED ISSUE #10215: Remappings are auto detected twice
### Component

Forge, Other (please describe)

- [x] Foundry
- [x] Foundryup

forge 1.0.0-stable (e144b82070 2025-02-13T20:02:34.979686000Z)

N/A

_No response_

### Operating System

macOS (Apple Silicon)

### Describe the bug

Remappings are auto detected twice when loading a foundry_config::Config through load_config, in different ways.

1. Config::with_root builds a default ProjectPathsConfig which detects remappings https://github.com/foundry-rs/foundry/blob/759b7d6f6db0bb995f22ec9cce6401c9edb82980/crates/config/src/lib.rs#L1626
2. Config::to_figment uses RemappingsProvider which has different logic of auto detection https://github.com/foundry-rs/foundry/blob/34ab2354e91139c4f86bca4a340936600397968a/crates/config/src/providers/remappings.rs#L226

Both are called in load_config through Config::figment_with_root https://github.com/foundry-rs/foundry/blob/759b7d6f6db0bb995f22ec9cce6401c9edb82980/crates/config/src/lib.rs#L1598-L1600

See samply profile: https://share.firefox.dev/41SgAYV

0xMars42:
Closes #10215

`_with_root` calls `ProjectPathsConfig::builder().build_with_root()`
which scans lib directories and auto-detects remappings. These
remappings are then overwritten when `to_figment` runs the
`RemappingsProvider`, which performs the same scan with proper
precedence handling (env vars, remappings.txt, toml config,
auto-detect).

Remove the redundant extraction from `_with_root` and let
`RemappingsProvider` be the single source of truth, avoiding a
duplicate filesystem scan on every config load.

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -1863,11 +1863,6 @@ impl Config {
             src: paths.sources.file_name().unwrap().into(),
             out: artifacts.clone(),
             libs: paths.libraries.into_iter().map(|lib| lib.file_name().unwrap().into()).collect(),
-            remappings: paths
-                .remappings
-                .into_iter()
-                .map(|r| RelativeRemapping::new(r, root))
-                .collect(),
             fs_permissions: FsPermissions::new([PathPermission::read(artifacts)]),
             ..Self::default()
         }


0xMars42:
if I'm not mistaken, `_with_root` runs before any toml/env parsing, so it can't respect settings like `auto_detect_remappings = false` or custom `libs` from foundry.toml. RemappingsProvider runs after the full figment is built, so it can handle env vars, remappings.txt, nested foundry.toml scanning and dedup. The detection in `_with_root` was just a basic `Remapping::find_many()` via `build_with_root` its results were always overwritten by RemappingsProvider anyway.

onbjerg:
sgtm

PR #13390, opened 2026-02-08
perf(traces): make trace identification async

### LINKED ISSUE #12714: perf: printing traces is too slow

0xMars42:
Closes #12714

## Problem

`ExternalIdentifier::identify_addresses()` used `foundry_common::block_on()` to synchronously wait for Etherscan/Sourcify HTTP requests during trace rendering. Since every caller is already in an async context, this unnecessarily blocked tokio worker threads via `block_in_place`.

## Solution

Convert the `TraceIdentifier` trait to async using `async_trait` (already a dependency, already used in the same file for `ExternalFetcherT`), remove the `block_on` wrapper in `ExternalIdentifier`, and add `.await` at all call sites.

The only exception is `load_contracts()` in `lib.rs`, which is called from sync contexts but only uses `LocalTraceIdentifier` (pure in-memory, no I/O) — the `block_on` there resolves immediately with zero overhead.

## Changes

- `TraceIdentifier::identify_addresses()` → `async fn`
- `ExternalIdentifier`: remove `block_on()`, `.await` the stream directly
- `FetchFuture` type alias: add `+ Send` bound (required by `async_trait`)
- `CallTraceDecoder::identify()` / `identify_addresses()` → `async fn`
- All call sites (forge test, forge script, cast debug, chisel): add `.await`

### DIFF
--- a/crates/cast/src/debug.rs
+++ b/crates/cast/src/debug.rs
@@ -66,7 +66,7 @@ pub(crate) async fn handle_traces(
     let mut decoder = builder.build();
 
     for (_, trace) in result.traces.as_deref_mut().unwrap_or_default() {
-        decoder.identify(trace, &mut identifier);
+        decoder.identify(trace, &mut identifier).await;
     }
 
     if decode_internal || debug {
--- a/crates/chisel/src/dispatcher.rs
+++ b/crates/chisel/src/dispatcher.rs
@@ -171,7 +171,7 @@ impl ChiselDispatcher {
         )?;
         if !identifier.is_empty() {
             for (_, trace) in &mut result.traces {
-                decoder.identify(trace, &mut identifier);
+                decoder.identify(trace, &mut identifier).await;
             }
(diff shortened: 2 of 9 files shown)


0xMars42:
Closing this for now — I noticed #12905 addresses the same issue and I should have referenced it.

My approach is the same (async trait + remove block_on) but cleaner (single focused commit, no unrelated changes). However, as @zerosnacks pointed out on #12905, this needs a before/after timing to prove it actually helps.

I'll reopen with proper benchmarks (forge test -vvv on a real project, before and after) later today.

Issue #12985, opened 2026-01-05
Slim the README


onbjerg:
### Component

Other (please describe)

### Describe the feature you would like

Most users should go to the website for help getting started. I think a brief overview of what Foundry is, how to contribute to it, and how to get help, is all we should have in the README. In other words, the README should funnel users to the website, and otherwise be contributor focused.

Issue #12984, opened 2026-01-05
Replace DappTools benchmarks with Hardhat


onbjerg:
### Component

Other (please describe)

### Describe the feature you would like

DappTools development was halted on March 23, 2022 (https://github.com/dapphub/dapptools/commit/f46d16707b90b79e59f2cfd1530dfea906f16a63) and is now in casual maintenance mode. We should replace the benchmarks in the README with a more relevant comparison today, which would be Hardhat.

I think for completeness we can add a small link to the older DappTools benchmarks, if anything, which are available in the blog post currently linked in the README.

PR #13393, opened 2026-02-09
docs: slim readme

onbjerg:
## Motivation

The README is a bit noisy & largely duplicates stuff from the book.

## Solution

Slim the README and link out to our other resources (such as the book, contributing guidelines etc). For example, the current up-to-date benchmarks live in the book.

Closes https://github.com/foundry-rs/foundry/issues/12985 and also closes https://github.com/foundry-rs/foundry/issues/12984 since it won't be relevant anymore.

This slim might be too aggressive, so looking for feedback.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,8 @@
 &nbsp;
 
 [![Github Actions][gha-badge]][gha-url] [![Telegram Chat][tg-badge]][tg-url] [![Telegram Support][tg-support-badge]][tg-support-url]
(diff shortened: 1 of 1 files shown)


zerosnacks:
I think it makes sense to slim the README as proposed & polish the book

PR #13398, opened 2026-02-09
perf(invariant): avoid cloning state changeset in fuzz runs

thunggis:
`state_changeset` (`HashMap<Address, Account>`) isn't read after being cloned on every fuzz call, so just `std::mem::take` it out. 
In high-depth invariant runs (depth=1000) this avoids 1000 deep copies of what can be a large map. 
similar to  #13396 and #13397, but this one has more impact since the cloned data is significantly larger.


grandizzy:
thank you, looks good! Pending gh recovery for CI run

PR #13399, opened 2026-02-09
fix: handle absolute library paths in remappings provider

avorylli:
Fix incorrect handling of absolute library paths in RemappingsProvider.

Previously, absolute paths in lib_paths were incorrectly replaced with
root.join("lib") in find_nested_foundry_remappings, causing remappings
from absolute library paths to be ignored. The same issue existed in
auto_detect_remappings where absolute paths were incorrectly joined with
root.

Now absolute paths are used directly, allowing remappings to be correctly
detected from libraries specified with absolute paths as documented.


emmajam:

PR #13410, opened 2026-02-10
feat: make mutation test operators configurable

zerosnacks:
Adds a `[mutation]` section to `foundry.toml` for configuring which mutation operators are active.

All operators are enabled by default. New operators added in future versions are automatically active. Users can exclude specific operators or re-include ones that are excluded by default:

```toml
[mutation]
# Exclude additional operators beyond the defaults
exclude_operators = ["elim-delegate", "assembly"]

# Re-enable operators that are excluded by default
include_operators = []
```

Effective operator set: `all() - default_excluded - exclude_operators + include_operators`

Available operators: `assembly`, `assignment`, `binary-op`, `delete-expression`, `elim-delegate`, `require`, `unary-op`.

**Implementation**

- Adds `MutatorType` enum and `MutationConfig` in `foundry-config` (uses `strum` for `Display`, `FromStr`, `EnumIter`)
- Adds `mutation: MutationConfig` to the `Config` struct
- Updates `MutatorRegistry::from_enabled()` to construct the registry from the resolved operator list
- Threads config through `MutationHandler` → `MutantVisitor` → `MutatorRegistry`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4799,6 +4799,7 @@ dependencies = [
  "snapbox",
  "solar-compiler",
  "soldeer-core",
+ "strum",
  "tempfile",
  "thiserror 2.0.18",
  "toml",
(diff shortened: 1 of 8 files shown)


zerosnacks:
> nice! Let's add them to be configs validated / warn if unknown keys?

@grandizzy yep good call, added

PR #13412, opened 2026-02-10
fix: address mutation testing issues

zerosnacks:
Addresses issues identified during code review of the mutation testing feature.

**Panic safety**: `generate_mutated_solidity` and `restore_original_source` were dead code (the runner uses its own `apply_mutation` with checked slicing and temp directories). Removed both.

**Path safety**: `copy_project_for_mutation` validates all relative paths before joining with temp directories via `ensure_safe_relative_path` helper. User-provided `--mutate` paths are canonicalized and verified to be under the project root.

**Naming collision**: Renames the orchestrator `MutationConfig` to `MutationRunConfig` to avoid collision with `foundry_config::MutationConfig`.

**Cache collisions**: Cache filenames include a hash of the full contract path, consolidated into a single `cache_file_path` helper.

**Error propagation**: `generate_ast` propagates parse errors instead of silently producing zero mutants.

**`--mutate-contract` fix**: Uses `.filter().any()` instead of `.find().is_some_and()` so all contracts in a multi-contract file are matched.

**JSON output purity**: Suppresses informational stdout prints in JSON mode. Skips the test runner's early JSON return when `--mutate` is active so only mutation JSON is emitted.

**Cross-platform `relative_path`**: Uses `Path::components()` with explicit "/" joining instead of string-based separator matching.

### DIFF
--- a/crates/forge/src/mutation/mod.rs
+++ b/crates/forge/src/mutation/mod.rs
@@ -317,7 +317,10 @@ impl MutationHandler {
 
     // Note: we now get the build hash directly from the recent compile output (see test flow)
 
-    fn cache_filename_prefix(&self) -> String {
+    /// Returns the cache file path for the given build hash and extension.
+    /// The filename encodes a hash of the full contract path to prevent collisions
+    /// between files with the same stem in different directories.
+    fn cache_file_path(&self, hash: &str, ext: &str) -> PathBuf {
         use std::{
             collections::hash_map::DefaultHasher,
             hash::{Hash, Hasher},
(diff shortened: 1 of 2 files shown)


crates/forge/src/mutation/mod.rs
```diff
@@ -370,62 +382,71 @@ impl MutationHandler {
...
-
         let src_content = Arc::clone(&self.src);
-
+        let src_bytes = src_content.as_bytes();
         let start_pos = span.lo().0 as usize;
         let end_pos = span.hi().0 as usize;
 
-        let before = &src_content[..start_pos];
-        let after = &src_content[end_pos..];
+        if start_pos > end_pos || end_pos > src_bytes.len() {
```
grandizzy: curious when this could happen

crates/forge/src/mutation/mod.rs
```diff
@@ -317,15 +317,27 @@ impl MutationHandler {
...
     /// Persists cached mutants using build hash for cache invalidation.
     /// Writes to `cache/mutation/<hash>_<filename>.mutants`.
     pub fn persist_cached_mutants(&self, hash: &str, mutants: &[Mutant]) -> std::io::Result<()> {
         let cache_dir = self.config.root.join(&self.config.mutation_dir);
         std::fs::create_dir_all(&cache_dir)?;
 
-        let filename =
-            self.contract_to_mutate.file_stem().and_then(|s| s.to_str()).unwrap_or("unknown");
-        let cache_file = cache_dir.join(format!("{hash}_{filename}.mutants"));
+        let prefix = self.cache_filename_prefix();
```
grandizzy: worth extracting these 2 lines in a helper as it repeats several times

crates/forge/src/mutation/runner.rs
```diff
@@ -133,9 +133,21 @@ fn relative_to_root(root: &Path, path: &Path) -> PathBuf {
 /// Uses symlinks for lib directories (read-only dependencies) to avoid expensive copies.
 /// Preserves the project's directory layout (handles custom src/test paths).
 fn copy_project_for_mutation(config: &Config, temp_dir: &Path) -> Result<()> {
-    // Compute relative paths to preserve project layout
     let src_rel = relative_to_root(&config.root, &config.src);
+    if !is_safe_relative_path(&src_rel) {
```
grandizzy: let's also have a helper /closure for this validation

crates/forge/src/mutation/mod.rs
zerosnacks: as written previously this would panic on OOB but it turns out `generate_mutated_solidity` is dead code; going to follow up with a PR to clean up dead code in this PR; appears there is quite a bit that occurred through the iterations

crates/forge/src/mutation/runner.rs
zerosnacks: added in https://github.com/foundry-rs/foundry/pull/13412/changes/c78b2684eb45f858201d0de446974b36be7d2998

crates/forge/src/mutation/mod.rs
zerosnacks: added in https://github.com/foundry-rs/foundry/pull/13412/changes/c78b2684eb45f858201d0de446974b36be7d2998

crates/forge/src/mutation/mod.rs
zerosnacks: follow up here: https://github.com/foundry-rs/foundry/pull/13413

Issue #13304, opened 2026-02-03
feat: getStateDiff / getStateDiffJson decode storage for non local contracts


sakulstra:
### Component

Forge

### Describe the feature you would like

We're using foundry to analyze transactions via `getStateDiff[Json]`.

For this case it would be very useful to have state decoding also for touched contracts that are **not** part of the current codebase. Not exactly sure about how it works on tracing, but i guess it would be a matter of downloading the contracts from the explorer when running the tests.

### Additional context

`getStateDiff[Json]` already is a huge improvement to local transaction analysis, but it falls short when using foundry for testing upgrades against forks / interacting with pre-deployed existing contracts.
Therefore it would be great to have the ability (even if with a flag) to have state decoding on external contracts as well.

simon0820s:
Can i try it ?

pjdurden:
Hi @sakulstra — I'd like to work on this (saw @simon0820s asked but there's no PR yet).

Approach: extend `getStateDiff` / `getStateDiffJson` so storage slots for touched contracts that aren't part of the local project are still decoded, by fetching the ABI for those addresses from the configured explorer (reusing the existing explorer/etherscan client) and caching per-address for the run. When no ABI or explorer key is available it would fall back to the current raw-slot output so fork/upgrade tests still work offline, and the network fetch would be gated behind the existing explorer config.

Does that direction work for you?

pjdurden:
Stepping back.  I see @sakulstra already has #13417 open for it and it's under review. I missed that earlier when I commented. Deferring to that PR.

Kris0721:
Hi @sakulstra — I'd like to pick this up. I have experience with Rust systems work (state root/latency-profiling type code) and want to start contributing to foundry.

Before diving in, want to confirm scope: for non-local contracts, we don't have the storage layout from the ABI, so decoding would need to fall back to either (a) heuristic slot-type inference, or (b) an external layout source (e.g. Etherscan/Sourcify if verified). Is there a preferred approach, or should I explore both and propose one?

Will start by reading through the existing local-decode path in crates/cheatcodes to understand the current implementation before proposing changes.

figtracer:
Implemented by [#16766](https://github.com/foundry-rs/foundry/pull/16766): `--decode-external-storage` enables verified-source storage-layout decoding for non-local contracts, including proxies.

PR #13417, opened 2026-02-11
feat: decoding external contracts

sakulstra:
Disclaimer: The code was written under heavy assistance by claude code, as i'm not very familiar with the foundry codebase. I did my best to not make it slop though, and carefully reviewed everything & added tests.

## Motivation

`getStateDiff()` / `getStateDiffJson()` currently only decodes storage slot names and values for **locally compiled** contracts. When a test interacts with external contracts on a fork (e.g., WETH, Aave, Uniswap), those addresses show raw hex slots and values because `get_contract_data()` only searches `available_artifacts` from local compilation output.
As we use foundry for transaction analysis (as alternative to tenderly), for us it would be incredible useful to not have this limitation.

- https://github.com/foundry-rs/foundry/issues/13304

## Solution

When `getStateDiff` is called, `get_recorded_state_diffs`:
- first tries local artifact matching (like right now), fallback to
- lookup in-memory, fallback to
- fetching from etherscan -> compile layout -> store in memory

I'm not quite sure how heavy artifact compilation is in practice, especially with caching, to me feels like it could be neglectable and just be the default. Anyhow, to not slow down things for all users, for now i introduced a `decode_external_storage` flag, so it's opt-in.

Misc:
I moved add_storage_layout_output from storage to compile as i use it in two places now (not sure if that is considered breaking).

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

---

It seems like there are quite some issues with the state decoding even before my pr.
- there is multiple storage layouts that cause panics (fixed the ones i faced on this branch)
- `Contract SomeContract` is decoded as bytes not address (fixed on this branch)
- structs are wrongly decoded <- **I did not fix this, as i didn't know how**
```
"0x5d0024d0336b60fc0da1975415d27ca027b626e6aa8bfa90b1c7c2ba7511f4d3": {
          "previousValue": "0x0000000000000000000000000000000000000000000000000000000000000000",
          "newValue": "0x0000000000000000000000000000000000000000000000000000000000000001",
          "label": "_roles[0x8aa855a911518ecfbe5bc3088c8f3dda7badf130faaf8ace33fdc33828e18167][0x000000000000000000000000a35358159f42e11c5689c68f181a71d51bb22de3]",
          "type": "mapping(bytes32 => struct AccessControl.RoleData)",
          "offset": 0,
          "slot": "42065348994196265316569453230947134357755835792195920849695704633452950320339",
          "decoded": {
            "previousValue": "0x",
            "newValue": "0x"
          },
          "keys": [
            "0x8aa855a911518ecfbe5bc3088c8f3dda7badf130faaf8ace33fdc33828e18167",
            "0x000000000000000000000000a35358159f42e11c5689c68f181a71d51bb22de3"
          ]
        },
```

Here is the json output of the storage diff

### DIFF
--- a/crates/cheatcodes/src/config.rs
+++ b/crates/cheatcodes/src/config.rs
@@ -121,14 +121,13 @@ impl CheatsConfig {
         let chain: alloy_chains::Chain = chain_id.into();
 
         // Try to find a matching config in the etherscan configs by chain ID.
-        if let Some(config) = self.etherscan_configs.clone().resolved().find_chain(chain) {
-            if let Ok(mut resolved) = config {
-                // Override key if CLI key is set.
-                if let Some(key) = &self.etherscan_api_key {
-                    resolved.key.clone_from(key);
-                }
-                return Some(resolved);
+        if let Some(Ok(mut resolved)) = self.etherscan_configs.clone().resolved().find_chain(chain)
+        {
+            // Override key if CLI key is set.
+            if let Some(key) = &self.etherscan_api_key {
+                resolved.key.clone_from(key);
             }
+            return Some(resolved);
         }
 
         // Fallback: create from API key + chain.
--- a/crates/cheatcodes/src/config.rs
+++ b/crates/cheatcodes/src/config.rs
@@ -4,8 +4,8 @@ use alloy_primitives::{U256, map::AddressHashMap};
 use foundry_common::{ContractsByArtifact, fs::normalize_path};
 use foundry_compilers::{ArtifactId, ProjectPathsConfig, utils::canonicalize};
 use foundry_config::{
-    Config, FsPermissions, ResolvedRpcEndpoint, ResolvedRpcEndpoints, RpcEndpoint, RpcEndpointUrl,
-    cache::StorageCachingConfig, fs_permissions::FsAccessKind,
+    Config, FsPermissions, ResolvedEtherscanConfig, ResolvedRpcEndpoint, ResolvedRpcEndpoints,
+    RpcEndpoint, RpcEndpointUrl, cache::StorageCachingConfig, fs_permissions::FsAccessKind,
 };
 use foundry_evm_core::opts::EvmOpts;
 use std::{
@@ -48,6 +48,10 @@ pub struct CheatsConfig {
     /// If Some, `vm.getDeployedCode` invocations are validated to be in scope of this list.
     /// If None, no validation is performed.
     pub available_artifacts: Option<ContractsByArtifact>,
+    /// Etherscan config for fetching external contract storage layouts.
+    /// When `Some`, enables fetching verified source code from Etherscan for storage layout
+    /// decoding in `getStateDiff()`.
+    pub etherscan_config: Option<ResolvedEtherscanConfig>,
     /// Currently running artifact.
     pub running_artifact: Option<ArtifactId>,
     /// Whether to enable legacy (non-reverting) assertions.
@@ -65,6 +69,7 @@ impl CheatsConfig {
         evm_opts: EvmOpts,
         available_artifacts: Option<ContractsByArtifact>,
         running_artifact: Option<ArtifactId>,
+        etherscan_config: Option<ResolvedEtherscanConfig>,
     ) -> Self {
         let rpc_endpoints = config.rpc_endpoints.clone().resolved();
         trace!(?rpc_endpoints, "using resolved rpc endpoints");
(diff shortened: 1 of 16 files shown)


sakulstra:
testing a bit against our actual codebase there are some issues re proxies, will test a bit more before i undraft it 😅

crates/common/src/slot_identifier.rs
```diff
@@ -218,11 +218,12 @@ impl SlotInfo {
             if length_byte & 1 == 1 {
                 // Long bytes/string - populate members
                 let length: U256 = U256::from_be_bytes(base_value.0) >> 1;
-                let num_slots = length.to::<usize>().div_ceil(32).min(256);
```
sakulstra, line 221: not sure if that's the correct approach, but when testing on some contracts i got a panic at:
```
The application panicked (crashed).
Message:  Uint conversion error: Overflow(256, 8343345644458265962, 18446744073709551615)
Location: crates/common/src/slot_identifier.rs:221
```

My understanding is that the .min guard is applied to late, so the overflow already happened.

crates/common/src/slot_identifier.rs
```diff
@@ -94,7 +94,7 @@ impl SlotInfo {
 
                 if length_byte & 1 == 0 {
                     // Short string/bytes (less than 32 bytes)
-                    let length = (length_byte >> 1) as usize;
+                    let length = ((length_byte >> 1) as usize).min(31);
```
sakulstra, line 97: Here i got:
```
The application panicked (crashed).
Message:  range end index 75 out of range for slice of length 32
Location: crates/common/src/slot_identifier.rs:99
```

On some contracts, so clamping it.

crates/cheatcodes/src/config.rs
```diff
@@ -100,6 +111,34 @@ impl CheatsConfig {
         Self::new(config, evm_opts, self.available_artifacts.clone(), self.running_artifact.clone())
     }
 
+    /// Resolves the etherscan config for the given chain ID.
+    /// Returns `None` if `decode_external_storage` is disabled or no config could be resolved.
+    pub fn get_etherscan_config(&self, chain_id: u64) -> Option<ResolvedEtherscanConfig> {
```
sakulstra, line 116: perhaps one could have used `Config::get_etherscan_config_with_chain` somehow, but started to feel a bit like a feature creep.

crates/forge/tests/cli/test_cmd/mod.rs
```diff
@@ -135,24 +136,28 @@ forgetest_init!(decode_external_storage_on_fork, |prj, cmd| {
...
 
-interface IWETH {
-    function deposit() external payable;
+interface IERC20 {
+    function transfer(address to, uint256 amount) external returns (bool);
 }
 
 contract DecodeExternalStorageTest is Test {
-    // WETH on mainnet
-    address constant WETH = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
```
sakulstra: eventually weth is not supported as it solc <0.5.13, replaced with 1inch

sakulstra:
@zerosnacks any feedback on this, it's quite a pain to run against a fork 😅
Would love to have this/or some version of it upstream so it works properly in ci.

zerosnacks:
Hi @sakulstra thanks for the PR! Apologies for the delay, I have updated it to master and will review.

crates/common/src/contracts.rs
```diff
@@ -635,6 +635,163 @@ pub fn find_matching_contract_artifact(
...
+        }
+    };
+    let metadata = source.items.first()?;
+
+    let (impl_address, impl_metadata) = if metadata.proxy != 0 {
+        let implementation = metadata.implementation?;
+        sh_println!(
+            "Contract at {address} is a proxy, trying to fetch source at {implementation}..."
+        )
+        .ok()?;
```
zerosnacks, line 677: This would silently abort, skipping the decoding of proxy contracts I believe

zerosnacks:
I have a few concerns in regards to the current implementation:

- The Etherscan fetching & compilation happens synchronously inside of `getStateDiff` which can cause racing lookups and can be quite slow (especially with the rps limits of Etherscan)
- `forge test` runs in parallel which can create a race condition / corruption of `{address}.json` cache files
- I think the remappings defined by the Etherscan source are trusted without path sanization; this is important to handle.
- Failed lookups are not cached and retried on every run; this should match the behavior of selector lookups where we record failed lookups and not re-attempt without forge clean

Directionally I think it is an interesting idea and generally supportive of it :+1:

sakulstra:
> and can be quite slow (especially with the rps limits of Etherscan)

This seems to be generally a problem with foundry, no? Have you folks considered putting / allowing to put a proxy in front of explorers. It's quite sad that in 2026, etherscan api limits are still a problem for fetching immutable data.

PR #13418, opened 2026-02-11
fix(cast): clean up temp dir in `cast storage` when etherscan cache is unavailable

thunggis:
The fallback temp directory created in `cast storage` was never cleaned up. `tempdir()?.keep()` permanently persists the dir, and `tempdir()?.path().to_path_buf()` drops the `TempDir` immediately making the path invalid. Hold the `TempDir` in an outer variable so it lives for the function scope and gets cleaned up on return.

### DIFF
--- a/crates/cast/src/cmd/storage.rs
+++ b/crates/cast/src/cmd/storage.rs
@@ -156,8 +156,7 @@ impl StorageArgs {
         }
 
         // Create or reuse a persistent cache for Etherscan sources; fall back to a temp dir
-        #[allow(clippy::collection_is_never_read)]
-        let mut _temp_dir = None;
+        let mut temp_dir = None;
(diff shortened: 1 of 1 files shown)


crates/cast/src/cmd/storage.rs
```diff
@@ -156,19 +156,27 @@ impl StorageArgs {
         }
 
         // Create or reuse a persistent cache for Etherscan sources; fall back to a temp dir
+        #[allow(clippy::collection_is_never_read)]
+        let mut _temp_dir = None;
```
onbjerg: please make this an explicit drop instead.


```suggestion
        let mut temp_dir = None;
```

later:
```
drop(temp_dir);
```

thunggis: Done

PR #13430, opened 2026-02-13
fix(cli): add Etherlink Shadownet to has_different_gas_calc

### LINKED ISSUE #13425: feat(`forge script`): add Etherlink Shadownet to calculate gas by RPC due to insufficient gas estimation related to different gas accounting model
### Component

Forge

- [x] Foundry
- [x] Foundryup

1.6.0-nightly

1.6.0-nightly

forge script --broadcast

### Operating System

macOS (Apple Silicon)

### Describe the bug

# `forge script --broadcast` fails on Etherlink Shadownet due to insufficient gas estimation

## Description

Etherlink uses a non-standard gas accounting model with inclusion fees. At a gas price of 1 gwei, Etherlink applies a fixed inclusion fee of approximately 600,000 gas on top of the EVM execution gas. This means transactions require significantly more gas than standard Ethereum networks.

When running `forge script --broadcast` on Etherlink Shadownet, Foundry estimates gas usage based on EVM execution costs only (e.g., ~170k gas) but sends transactions with even lower gas limits (e.g., ~135k gas). The Etherlink sequencer rejects these transactions with errors like:
```
The provided gas limit (135817) is insufficient to cover the transaction cost of 1709178 gas.
```

## Issues

The `--gas-limit` parameter appears to be ignored: even when explicitly passing `--gas-limit 2500000`, Foundry still sends transactions with ~150k gas limits. The `--gas-estimate-multiplier` option also fails because it multiplies an already incorrect base estimate from the RPC.

## Comparison

The same scripts execute successfully on Ethereum Sepolia without any manual gas configuration, confirming the issue is specific to Etherlink's gas model.

**Note:** `forge create` and `cast send` work correctly on Etherlink, suggesting the issue is specific to `forge script --broadcast`.

gakonst:
## Summary
Add Etherlink Shadownet (chain ID 127823) to `has_different_gas_calc` so `forge script --broadcast` re-estimates gas via `eth_estimateGas` on this chain.

## Motivation
Etherlink uses a non-standard gas model with inclusion fees (~600k gas on top of EVM execution gas). Without RPC-based gas estimation, `forge script` sends transactions with gas limits based on local EVM simulation only, which the sequencer rejects.

Etherlink Mainnet (42793) and Ghostnet Testnet (128123) are already handled via `NamedChain`, but Shadownet (127823) is not yet in alloy-chains, so we match the raw chain ID.

Closes #13425

## Changes
- `crates/cli/src/utils/cmd.rs`: add chain ID 127823 check before the `NamedChain` match

### DIFF
--- a/crates/cli/src/utils/cmd.rs
+++ b/crates/cli/src/utils/cmd.rs
@@ -111,6 +111,11 @@ pub fn init_progress(len: u64, label: &str) -> indicatif::ProgressBar {
 
 /// True if the network calculates gas costs differently.
 pub fn has_different_gas_calc(chain_id: u64) -> bool {
+    // Etherlink Shadownet (not yet in alloy-chains as a NamedChain variant).
+    if chain_id == 127823 {
+        return true;
+    }
+
     if let Some(chain) = Chain::from(chain_id).named() {
         return chain.is_arbitrum()
             || chain.is_elastic()


mablr (commented):
This is already handled here:
https://github.com/foundry-rs/foundry/blob/a249f5cc35685c7d0ac5871885e06da5da623d52/crates/cli/src/utils/cmd.rs#L126

PR #13435, opened 2026-02-13
feat: log broken invariant as soon as it is found

### LINKED ISSUE #13285: Log broken invariant as soon as it is found
### Component

Forge

I am trying to benchmark Foundry invariant tests against other fuzzers, and it would be really important to have real time information of the exact time an invariant breaks.

### Additional context

N/A

CreeptoGengar:
Adds real-time logging when invariants break during fuzzing with timestamp, invariant name, reason, and sequence length for benchmarking purposes.

Closes #13285


mablr (commented):
This PR got superseeded by #14433.

Thank you for the effort 🫡

PR #13436, opened 2026-02-13
fix(anvil): variable shadowing bug in ReadyTransactions::remove_with_markers

letmehateu:
Fixes variable shadowing bug that caused incorrect transaction removal from unlock lists.


mattsse (approved):
Correct fix for a real variable shadowing bug in `remove_with_markers`. The old code shadowed `hash` (the tx being removed) with the provider's hash from `provided_markers.get(mark)`, then searched `unlocks` for the wrong value — meaning unlock-list entries were never actually cleaned up. The rename to `provider_hash`/`provider_tx` and the comparison against the outer `hash` is the right fix.

PR #13440, opened 2026-02-14
fix(cli): avoid claiming the global log backend

beeb:
## Motivation

Currently, verbose logging in `forge soldeer` (like `forge soldeer install -vvv`) does not work.

`subscriber()` calls `.init()` on the tracing subscriber, which internally installs a `LogTracer`  via `log::set_logger()`. When soldeer later tries to initialize `env_logger` for its own verbose output, `try_init()` silently fails because the log backend is already claimed.
Soldeer's log events get routed through the `LogTracer` bridge into forge's tracing `EnvFilter`, which defaults to ERROR level so no log is shown in normal operation.

## Solution

Replace `.init()` with `tracing::subscriber::set_global_default()`, which sets the tracing subscriber without installing a `LogTracer`. This leaves the `log` backend free for soldeer's `env_logger`.

This is safe because forge itself uses `tracing` macros (not `log`), and the libraries that do use `log` (hyper, rustls, mio, etc.) are already silenced via `default_directives.txt`.

This was tested locally by hand.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cli/src/utils/mod.rs
+++ b/crates/cli/src/utils/mod.rs
@@ -87,8 +87,7 @@ pub fn subscriber() {
     let registry = tracing_subscriber::Registry::default().with(env_filter());
     #[cfg(feature = "tracy")]
     let registry = registry.with(tracing_tracy::TracyLayer::default());
-    let subscriber =
-        registry.with(tracing_subscriber::fmt::layer().with_writer(std::io::stderr));
+    let subscriber = registry.with(tracing_subscriber::fmt::layer().with_writer(std::io::stderr));
     // NOTE: we use `set_global_default` instead of `init()` to avoid installing a `LogTracer`.
     // `init()` calls `tracing_log::LogTracer::init()` which claims the global `log` backend,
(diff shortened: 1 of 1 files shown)


beeb:
@onbjerg then please tell me how I can control the logging level from within soldeer?

onbjerg:
i'm not sure what you mean. libraries should not be installing subscribers: https://docs.rs/tracing/latest/tracing/#in-libraries

beeb:
I'm using env_logger, not tracing

onbjerg:
same story: https://docs.rs/log/latest/log/#in-libraries (env_logger links to log)

if this is because you need the logs in a soldeer binary, the binary itself should instantiate the logger, not the library

beeb:
Ok I need to check how I can adjust the logging level in forge before calling soldeer, because it seems forge does not honor the `-v` flag. It uses the ERROR level even with `-vvv`

beeb:
@onbjerg new PR is open https://github.com/foundry-rs/foundry/pull/13485

PR #13445, opened 2026-02-14
fix(anvil-server): validate allow_origin and no_cors consistency

PivasDesant:
Prevent allow_origin and no_cors from conflicting. Setting allow_origin now enables CORS automatically, and deserialization fails if both are set incorrectly.


mattsse (commented):
we can keep as is

PR #13446, opened 2026-02-14
fix(traces): check HTTP status before JSON parsing in Sourcify fetcher

marukai67:
Fixes incorrect error handling and removes unnecessary JSON parsing in SourcifyFetcher::fetch for 404/429 responses.


onbjerg (dismissed):

DaniPopes (changes requested):
dont need tests

PR #13455, opened 2026-02-15
feat(fuzz): add cheatcode to populate in-memory corpus

### LINKED ISSUE #12526: feat(fuzz): cheatcode to populate in-memory corpus
### Component

Forge

A straightforward way to allow experimentation without requiring modifying the fuzzer would be to add a cheatcode that takes a hash and have the fuzzer save new hashes into the queue for mutation. It would be similar to IJON (https://github.com/RUB-SysSec/ijon). Then, power uses can define how interesting a state is with a hash (as fine-or-coarse grained as you please) 

Basically permanently add an input to the in-memory corpus

### Additional context

_No response_

CreeptoGengar:
Implements `vm.addToCorpus(bytes32 hash)` cheatcode that allows users to permanently add the current fuzz input to the in-memory corpus for mutation.

This enables power users to define interesting states with a hash (as fine-or-coarse grained as desired), similar to IJON. The cheatcode marks the current input during execution, and the fuzzer adds it to the corpus unconditionally after the call completes.

**Changes:**
- Add `addToCorpus(bytes32 hash)` cheatcode definition
- Track current fuzz input in `Cheatcodes` state
- Add `add_input_unconditionally()` method to `WorkerCorpus`
- Integrate with both stateless fuzz tests and invariant tests

Closes #12526

### DIFF
--- a/crates/cheatcodes/spec/src/vm.rs
+++ b/crates/cheatcodes/spec/src/vm.rs
@@ -984,6 +984,12 @@ interface Vm {
     #[cheatcode(group = Testing, safety = Safe)]
     function assume(bool condition) external pure;
 
+    /// Permanently add the current fuzz input to the in-memory corpus for mutation.
+    /// The hash parameter allows users to define how interesting a state is (as fine-or-coarse grained as you please).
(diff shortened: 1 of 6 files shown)


emmajam:

PR #13473, opened 2026-02-17
add EIP-3860 initcode size check in txpool validation

DanielGuupta:
## Motivation
Contract creation transactions with initcode exceeding 49152 bytes were accepted into the mempool and only rejected at execution time. 

## Solution
Add the missing EIP-3860 size check in `validate_pool_transaction_for` so oversized txs are rejected early at `eth_sendRawTransaction`, consistent with how we handle other fork-aware validations (EIP-1559, EIP-4844).

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -3685,12 +3685,17 @@ impl TransactionValidator for Backend {
             }
         }
 
-        // EIP-3860 initcode size validation
-        if env.evm_env.cfg_env.spec >= SpecId::SHANGHAI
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
so I think instead of this we should add a way to make this configurable

DanielGuupta:
> so I think instead of this we should add a way to make this configurable

updated to respect `limit_contract_code_size` from CfgEnv. Now it follows the EIP-3860 formula (`2 * code_size_limit`) and is effectively skipped when `--disable-code-size-limit` is used.

PR #13476, opened 2026-02-17
fix(primitives): use proper JSON serialization for deposit tx fields

dizer-ti:
## Motivation

`From<FoundryTypedTx> for FoundryTransactionRequest` for Deposit variant used `.to_string().into()` which created JSON strings instead of typed values. This broke roundtrip conversion since `get_deposit_tx_parts()` expects `B256`, `U256`, `bool` types.

## Solution

Use `serde_json::to_value()` for Deposit fields, matching existing Tempo implementation.

## PR Checklist

- [x] Added Tests


mablr (dismissed):
Makes sense

stevencartavia (dismissed):

mablr (approved):

stevencartavia (approved):

PR #13477, opened 2026-02-17
fix: use stored hi position in DisableEnd handler

MozirDmitriy:
What was wrong: In the DisableEnd handler, the code was ignoring the stored `hi` position from the disabled_blocks entry and using `span.hi()` instead. This caused incorrect range calculation when multiple disable-start/disable-end pairs were nested.

Changes: Modified the DisableEnd case to use the stored `hi` value from the entry when creating the disabled range, ensuring the range correctly ends at the first disable-end (outermost) rather than the last one.


mablr (commented):
We can keep as is, ty.

PR #13485, opened 2026-02-18
feat(cli): allow soldeer to update the logging level

beeb:
This PR follows the discussion in #13440 

## Motivation

Currently, verbose logging in `forge soldeer` (like `forge soldeer install -vvv`) does not work.

`subscriber()` calls `.init()` on the tracing subscriber, which internally installs a `LogTracer`  via `log::set_logger()`. Soldeer's log events get routed through the `LogTracer` bridge into forge's tracing `EnvFilter`, which defaults to ERROR level so no log is shown in normal operation.

## Solution

This PR adds a `reload` layer to the `EnvFilter` so it can be changed by the `soldeer` subcommand according to `-v` flags, just before calling the `soldeer-commands` library.

Soldeer-commands were modified to avoid trying to init the log backend again as a good practice: https://github.com/mario-eth/soldeer/pull/350

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


beeb:
@onbjerg do you think this approach is acceptable?

beeb:
@onbjerg bump

mattsse (approved):
LGTM

crates/cli/src/utils/mod.rs
```diff
@@ -82,17 +86,36 @@ impl<T: AsRef<Path>> FoundryPathExt for T {
...
+
+/// Replaces the active tracing `EnvFilter` at runtime.
+///
+/// `directives` is parsed as an [`EnvFilter`] (e.g. `"info"`, `"debug,hyper=off"`).
+/// This is a no-op if [`subscriber`] has not been called yet.
+pub fn update_tracing_filter(directives: &str) {
+    let Some(handle) = FILTER_RELOAD_HANDLE.get() else {
+        return;
+    };
+    let Ok(new_filter) = directives.parse::<EnvFilter>() else {
```
mablr, line 110: Seems that `update_tracing_filter("info"/"debug"/"trace")` replaces the whole `EnvFilter`, so `forge soldeer -vv+` drops the default suppressions from `default_directives.txt` (`hyper=off`, `h2=off`, `rustls=off`, etc.). That may re-enable unrelated dependency logs that Foundry intentionally silences.

figtracer (changes requested):
Found one medium-confidence behavioral regression.

`forge soldeer -vv...` drops Foundry's default trace suppressions.

In the PR, `SoldeerArgs::run` calls:

```rust
utils::update_tracing_filter(level);
```

with just `"info"`, `"debug"`, or `"trace"` when `RUST_LOG` is unset. `update_tracing_filter` then parses that string and reloads the `EnvFilter` wholesale.

That bypasses the default directives applied by `env_filter()`:

```text
"hyper=off",
"hyper_util=off",
"h2=off",
"rustls=off",
"mio=off",
"jsonpath_lib=off",
```

So `forge soldeer -vvv` / `-vvvv` can start emitting logs from dependencies that Foundry intentionally suppresses in the initial subscriber.

Suggested fix: factor the default-directive application into a helper and apply it to both the initial filter and reloaded filters, e.g. parse the soldeer level, then append the same defaults before `handle.reload(...)`.

Relevant changed lines:

- `crates/forge/src/cmd/soldeer.rs`: new `update_tracing_filter(level)` call around lines 24-32
- `crates/cli/src/utils/mod.rs`: `update_tracing_filter` around lines 106-113 and `env_filter` defaults around lines 116-121

Notes:

- I don't think this PR newly breaks `RUST_LOG` precedence: the soldeer reload is guarded by `RUST_LOG` being absent.
- The verbosity mapping itself looks consistent with Soldeer's expected levels: `-v=warn`, `-vv=info`, `-vvv=debug`, `-vvvv=trace`.

crates/cli/src/utils/mod.rs
beeb: I think that's fine, soldeer never calls into forge code after starting to run a command. If any other consumer needs to use this mechanism, then it's on them to correctly merge the directives if needed. For soldeer I don't think we need to filter out hyper, h2, etc.

figtracer (approved):

crates/cli/src/utils/mod.rs
mablr: Okay, thank you for the clarification @beeb !

PR #13489, opened 2026-02-19
chore: update LATEST_SOLC to 0.8.34

mattsse:
Update `LATEST_SOLC` constant from 0.8.30 to 0.8.34.


zerosnacks:
CI fails, do we need to bump svm-rs crate version here?

PR #13493, opened 2026-02-19
perf(invariant): cache ABI function selectors in TargetedContract

letmehateu:
Pre-compute a HashMap<Selector, Function> once per TargetedContract instead of calling Function::selector() (which allocates a String + computes keccak256 with no caching) on every comparison in the invariant fuzz hot loop.


0xMars42:
The switch from `calldata[..4]` to `first_chunk::<4>()` in the three lookup sites is a nice side-effect here. The old code would panic if calldata happened to be shorter than 4 bytes, while this gracefully returns None/false. Probably not reachable in practice since the invariant fuzzer generates selector-based calls, but could be worth noting in the PR description as a small defensive improvement?

letmehateu:
@onbjerg sorry, can you check please my PR?

PR #13514, opened 2026-02-22
fix(fmt): correct trailing comment handling in Yul switch cases

sashass1315:
The trailing comment lookup after each switch case incorrectly used `selector.span.hi()` instead of the case's own `span.hi()`. This caused trailing comments after individual cases to be skipped or misprocessed, since the lookup was searching on the wrong source line.

### DIFF
--- a/crates/fmt/src/state/yul.rs
+++ b/crates/fmt/src/state/yul.rs
@@ -101,7 +101,7 @@ impl<'ast> State<'_, 'ast> {
                     }
                     self.print_yul_block(body, *span, false, 0);
 
-                    self.print_trailing_comment(selector.span.hi(), None);
+                    self.print_trailing_comment(span.hi(), None);
                 }
             }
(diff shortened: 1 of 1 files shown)


zerosnacks:
Hi @sashass1315 thanks for your PR, would you mind adding a test case for this?

grandizzy:
The only path where the changed line actually executes is when `print_yul_block` returns early — i.e. with a `// forgefmt: disable-line` on the case. But that codepath is non-idempotent independent of this change (formatting the result a second time garbles it into case \ncase 0  { pop(0) }), so it can't be expressed as a passing snapshot test either.

So while the change is arguably more "correct" in isolation, it doesn't fix any observable behavior, and the one scenario where it matters is blocked by a separate disable-line idempotency bug. Going to close this for now — if you want to pursue it, the more impactful fix would be to address the forgefmt: disable-line handling for Yul switch cases directly, which would give us something testable.

Thanks!

PR #13515, opened 2026-02-23
feat(cast): add `--data` flag to access-list command

VolodymyrBg:
The `cast access-list` command was missing the `--data` flag that exists in `cast call` and `cast send`, making the API inconsistent. Users had to pass raw hex calldata through the positional `sig` argument, which was undocumented and non-obvious.

Added `--data` flag to `AccessListArgs` with proper conflicts_with_all constraint, matching the pattern used in call.rs and send.rs.


VolodymyrBg:
> please add a test, ty

Sorry for delay, added

onbjerg (approved):

PR #13519, opened 2026-02-23
feat(invariant): add fail_on_assert for Solidity assert failures

### LINKED ISSUE #13322: Report assertion failures as broken invariants
### Component

Forge

When running foundry invariant tests with `fail_on_revert = false`, a broken solidity assertion (`assert(x == y)`) will be discarded from a sequence and not be reported by the fuzzer.

This is perhaps the expected behavior, but it is unintuitive.

Comparatively to other smart contract fuzzers, assertion failures will also be reported as a bug.

I propose to either change the current implementation or to allow, through a `[invariant]` specific configuration, to also report assertion failures during invariant tests. 

### Additional context

As a workaround, I have been doing the following:

```
    mapping(string => bool) private assertionFailures;

    function _isAssertion(string memory reason) internal pure returns (bool) {
        return bytes(reason).length >= 3 &&
            bytes(reason)[0] == '!' &&
            bytes(reason)[1] == '!' &&
            bytes(reason)[2] == '!';
    }

    function t(bool b, string memory reason)
        internal
        virtual
        override(FoundryAsserts, Asserts)
    {
        if(_isAssertion(reason)) {
            _recordAssertion(b, reason);
        }
        else {
            super.t(b, reason);
        }
    }

    function _recordAssertion(bool ok, string memory reason) internal {
        if (ok) {
            return;
        }

        assertionFailures[reason] = true;
    }

    function invariant_assertion_failure_WITHDRAW_DOS() public view {
        assertTrue(!assertionFailures[ASSERTION_WITHDRAW_DOS], ASSERTION_WITHDRAW_DOS);
    }

    function invariant_assertion_failure_LIQUIDATION_CALL_DOS() public view {
        assertTrue(!assertionFailures[ASSERTION_LIQUIDATION_CALL_DOS], ASSERTION_LIQUIDATION_CALL_DOS);
    }

    function invariant_assertion_failure_REPAY_DOS() public view {
        assertTrue(!assertionFailures[ASSERTION_REPAY_DOS], ASSERTION_REPAY_DOS);
    }
```

This is a hack because

1. foundry needs one `invariant_XXX` per "assertion failure"
2. if I do `assertEq` in an assertion, the call is discarded from the sequence, so `_recordAssertion` is a workaround
3. to distinguish between global invariants and local assertions, I prefix the failure reason with `!!!`, so I have e.g.  `assertTrue(success, "!!! withdraw DOS")` that will make foundry pick this assertion failure in my tests

aviggiano:
## Summary
- add `[invariant].fail_on_assert` to `InvariantConfig` (default `false`)
- fail invariant campaigns on Solidity assert failures even when `fail_on_revert = false`
  - `Panic(0x01)`
  - legacy invalid-opcode assertion behavior (`InstructionResult::InvalidFEOpcode`)
- keep behavior unchanged for:
  - `require`/generic reverts
  - non-assert panics (e.g. `Panic(0x11)`)
  - `vm.assert*` / `GLOBAL_FAIL_SLOT` semantics
- classify `fail_on_assert` failures as `BrokenInvariant` (not generic `Revert`)
- apply the same condition in replay/shrink paths so persisted failures stay reproducible

## Implementation Notes
- config model
  - `crates/config/src/invariant.rs`
- runtime decision paths
  - `crates/evm/evm/src/executors/invariant/result.rs`
  - `crates/evm/evm/src/executors/invariant/mod.rs`
- replay/shrink parity
  - `crates/evm/evm/src/executors/invariant/shrink.rs`
  - `crates/forge/src/runner.rs`
- propagated case data
  - `crates/evm/evm/src/executors/invariant/error.rs`
- config snapshots
  - `crates/forge/tests/cli/config.rs`
- regression tests
  - `crates/forge/tests/cli/test_cmd/invariant/common.rs`

## Tests
- `cargo +1.91.0 test -p foundry-evm invariant::result::tests --lib --locked`
- `cargo +1.91.0 test -p forge --test cli fail_on_assert --locked`
- `cargo +1.91.0 test -p forge --test cli invariant_ignore_assert_panic_when_flag_off --locked`
- `cargo +1.91.0 test -p forge --test cli test_default_config --locked`

Closes #13322

### DIFF
--- a/crates/config/src/invariant.rs
+++ b/crates/config/src/invariant.rs
@@ -13,6 +13,9 @@ pub struct InvariantConfig {
     pub depth: u32,
     /// Fails the invariant fuzzing if a revert occurs
     pub fail_on_revert: bool,
+    /// Fails the invariant fuzzing if a Solidity assert failure occurs (`Panic(0x01)` or legacy
+    /// invalid opcode assert behavior), even if `fail_on_revert` is `false`.
+    pub fail_on_assert: bool,
     /// Allows overriding an unsafe external call when running invariant tests. eg. reentrancy
     /// checks
     pub call_override: bool,
@@ -57,6 +60,7 @@ impl Default for InvariantConfig {
             runs: 256,
             depth: 500,
             fail_on_revert: false,
+            fail_on_assert: false,
             call_override: false,
             dictionary: FuzzDictionaryConfig { dictionary_weight: 80, ..Default::default() },
             shrink_run_limit: 5000,
--- a/crates/evm/evm/src/executors/invariant/error.rs
+++ b/crates/evm/evm/src/executors/invariant/error.rs
@@ -3,7 +3,7 @@ use crate::executors::RawCallResult;
 use alloy_primitives::{Address, Bytes};
 use foundry_config::InvariantConfig;
 use foundry_evm_core::decode::RevertDecoder;
-use foundry_evm_fuzz::{BasicTxDetails, Reason, invariant::FuzzRunIdentifiedContracts};
+use foundry_evm_fuzz::{invariant::FuzzRunIdentifiedContracts, BasicTxDetails, Reason};
 use proptest::test_runner::TestError;
 
 /// Stores information about failures and reverts of the invariant tests.
@@ -67,6 +67,8 @@ pub struct FailedInvariantCaseData {
     pub fail_on_revert: bool,
     /// Fail on Solidity assert failures, used to check sequence when shrinking.
     pub fail_on_assert: bool,
+    /// Handler function that triggered a fail-on-assert violation, when available.
+    pub failing_handler: Option<String>,
 }
 
 impl FailedInvariantCaseData {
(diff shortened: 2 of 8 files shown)


aviggiano:
Wait a sec, this is not exactly what we need

I'm testing this branch on https://github.com/Recon-Fuzz/scfuzzbench-canary/pull/2 but it is reporting `invariant_noop` as a root cause for assertion failure, when in reality it should report `counter_increment`

I'll try to fix this

aviggiano:
Still not feature parity with medusa, since we will only see assertion failures on logs. I'd also see them as "first class citizen" in metrics JSON

0xalpharush:
I rebased https://github.com/foundry-rs/foundry/pull/12587/ and got assertion failures working on this canary (with shrinking/replay working w/o changing from an assertion to invariant). I also updated the logs and trace format 

https://github.com/0xalpharush/foundry/tree/fail_on_assert

aviggiano:
hey, thanks for the contribution

This looks really good overall, with one caveat: it is picking up `assert(...)` but not Forge cheatcode assertions `vm.assert*`

So we need to think whether we'd want to support both out of the box (I think so)

0xalpharush:
It is checking this slot but maybe there's some weird interaction with `assertions_revert`. or it's not the right way..
https://github.com/foundry-rs/foundry/commit/2cf1340a8e2c974d1273c5f0e4656b3ed3c7de9e#diff-6547e74fc7387bb7d95a9612a3a04ba728664f9b0a359c598fd6bdb64b2ce118R702-R720

grandizzy:
addressed in https://github.com/foundry-rs/foundry/pull/14275

PR #7701, opened 2024-04-18
Upgrade to latest version of Alloy and port Anvil tests

zerosnacks:
## Motivation

Builds on top of #7619 
Closes #7449 

## Solution

Upgrades Foundry to latest version of Alloy

Updates the following examples to use Alloy

- [x] abi.rs
- [x] anvil.rs
- [x] anvil_api.rs
- [x] api.rs
- [x] fork.rs
- [x] ganache.rs
- [x] gas.rs
- [x] genesis.rs
- [x] geth.rs
- [x] ipc.rs
- [x] logs.rs
- [x] optimism.rs
- [x] otterscan.rs
- [x] proof.rs
- [x] pubsub.rs
- [x] revert.rs
- [x] sign.rs
- [x] state.rs
- [x] traces.rs
- [x] transaction.rs
- [x] txpool.rs
- [x] utils.rs
- [x] wsapi.rs

**Breaking changes**:

- Remove special case handling of `Polygon` and `Polygon Mumbai` EIP-1559 fees handling of GasOracle middleware

**Notes**:

- The final migration of `cast bind` and `forge bind` is planned to be handled in a standalone ticket given its possible complexity and the diff has already grown massively
- Given that the port of Alloy has largely already occurred the odd cases of failing tests are largely descriptive of current issues (and flagged as such), not new issues introduced by this ticket.

- Few tests have been migrated but fail due to issues:

1. ~https://github.com/alloy-rs/alloy/issues/389~
~Tests failing due to `test_fork_snapshotting_blocks`,  `test_fork_timestamp`, `test_fork_reset_moonbeam`, `can_call_ots_get_transaction_error`, `can_replace_transaction`.~

2. RpcError in `watch_full_pending_transactions`
```bash
ErrorPayload { code: -32602, message: "invalid type: boolean `true`, expected
// unit", data: None } originating from watch_full_pending_transactions
```
~3. Test failing: `can_stream_pending_transactions`: Pubsub `channel_size` issue in `test_sub_new_heads_fast`.~
4. Optimism `test_deposits_not_supported_if_optimism_disabled`, likely the other Optimism tests aren't working as expected
5. Test failing: `can_get_block_number_ipc` and `test_sub_new_heads_ipc` fail on Windows

PR #13540, opened 2026-02-25
fix(anvil): clear transaction pool on anvil_reset

Kinaqu:
## Motivation
After calling `anvil_reset`, the transaction pool was not cleared. 
Subsequent transactions would be submitted to the pool but never mined, 
causing `get_receipt()` to hang indefinitely.

This bug was introduced in #7701 and marked with a FIXME comment in 
`test_fork_timestamp`.

## Solution
Added `self.pool.clear()` in `anvil_reset` before delegating to 
`reset_fork` or `reset_to_in_mem`, ensuring stale pending transactions 
are removed on reset.

## PR Checklist
- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/eth/backend/executor.rs
+++ b/crates/anvil/src/eth/backend/executor.rs
@@ -18,6 +18,7 @@ use alloy_consensus::{
     proofs::calculate_receipt_root, transaction::Either,
 };
 use alloy_eips::{
+    Encodable2718,
     eip7685::EMPTY_REQUESTS_HASH,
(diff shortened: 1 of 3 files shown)


mattsse (commented):
I dont think this is right

Kinaqu:
Thanks for the review! Could you point me in the right direction? 
I'd like to understand the correct approach for fixing the hang in 
`test_fork_timestamp` after `anvil_reset`.

Issue #13501, opened 2026-02-20
feat: avoid dependency on git binary


beeb:
### Component

Forge

### Describe the feature you would like

Forge uses the `git` CLI for many operations, so does [Soldeer](https://github.com/mario-eth/soldeer) (which is used in the `forge soldeer` subcommand).

Would it be acceptable to start rewriting the operations that depend on `git` to use the [`gix`](https://docs.rs/gix/latest/gix/) library?

This would avoid potential inconsistencies when it comes to the git version installed by the user, and would avoid the overhead of spawning a new process. Also, it reduces the potential for errors when parsing the output of the `git` command.

I could start by rewriting Soldeer to use `gix` as a proof-of-concept, if the additional dependency is acceptable.

zerosnacks:
I think this makes sense but would be good if it is done following an exploration in Soldeer so we can better evaluate the impact and usefulness. Any objections @onbjerg / @DaniPopes / @grandizzy? It is also unlikely any of us have the bandwidth / can prioritize this.

mablr:
Closing as not planned. I explored this in #14452. While `gix` could replace some read-only operations, Forge would still need the Git binary for cloning, checkout, index mutation, and extensive submodule management.

A partial migration would add a second Git implementation without removing the dependency or providing enough benefit to justify the added complexity. We can reconsider if `gix` eventually supports the complete workflow.

beeb:
Came to the same conclusion while investigating for Soldeer :)

PR #13542, opened 2026-02-26
feat: replace git cli with gix for clone/fetch as POC (#13501)

benslimanh:
## Motivation

#13501. 
Spawning the `git` CLI creates overhead and depends on the user's git installation. As requested in the issue, this is a proof-of-concept to replace git operations with the Rust `gix` crate, starting with `clone` and `fetch`.

## Solution
I updated `Git::clone` and `Git::fetch` to use a "try gix first, fallback to CLI" approach. This ensures we don't break anything if `gix` fails during this testing phase.

Main changes:
- Added `git_gix.rs` to handle `clone` and `fetch` using `gix` (version 0.72.1).
- Added `worktree-mutation` and `revision` features to `gix` in the workspace `Cargo.toml`.
- If the `gix` operation fails, it logs a warning/debug and safely falls back to the old `Command::new("git")`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1244,6 +1244,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "arc-swap"
+version = "1.8.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f9f3647c145568cec02c42054e07bdf9a5a698e15b466fb2341bfc393cd24aa5"
+dependencies = [
+ "rustversion",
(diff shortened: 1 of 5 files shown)


build_err.txt
zerosnacks, line 1: avoid committing these

benslimanh: removed

PR #13547, opened 2026-02-26
fix(anvil): consistent revert data format between eth_call and eth_estimateGas

FredPhilipy:
`eth_call` always returns `"data": "0x"` on reverts with no output, while `eth_estimateGas` returns `"data": null` for the same transaction. The difference comes from `ensure_return_ok` converting `None` output to `Bytes::default()`, while the gas estimation path passes `None` through directly.

This trips up client libraries (ethers.js, viem, etc.) that branch on `error.data` type when parsing revert reasons.


mablr (approved):
Added a test.

LGTM

grandizzy (approved):

PR #13557, opened 2026-02-27
fix(forge): apply --access-list in forge create

FredPhilipy:
Reopen of #13521, which was closed due to a compilation error after merging master (borrow of partially moved value on self). Fixed by using ref binding in the access list pattern match to avoid partial move.

cc @onbjerg


onbjerg (approved):

onbjerg:
thanks. sorry for abruptly closing, the amount of PRs we get where this is the case has become unmanageable unfortunately

PR #13560, opened 2026-02-28
fix(fmt): dropped statements for single-line **nested** if statements

### LINKED ISSUE #13555: `forge fmt` silently drops statements from single-line `while` loop bodies with multiple statements
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge nightly (b0a9dd9)
Commit SHA: b0a9dd9ceda36f63e2326ce530c10e6916f4b8a2
Build Timestamp: 2025-12-22T11:41:09.812070000Z
Build Profile: maxperf

1.0.1

forge fmt

### Operating System

macOS

### Describe the bug

`forge fmt` silently drops statements from single-line `while` loop bodies containing multiple statements in a braced block, changing program semantics without any warning.

**Steps to Reproduce**

Save the following as `test_fmt.sol`:

```solidity
pragma solidity ^0.8.20;

contract Test {
    // count the number of decimal digits in n
    function digitCount(uint256 n) internal pure returns (uint256) {
        if (n == 0) return 1;
        uint256 temp = n;
        uint256 count;
        while (temp != 0) { count++; temp /= 10; }
        return count;
    }

    // reverse-fill a decimal string buffer
    function toDecimalString(uint256 n) internal pure returns (string memory) {
        if (n == 0) return "0";
        uint256 len = digitCount(n);
        bytes memory buf = new bytes(len);
        while (n != 0) { len -= 1; buf[len] = bytes1(uint8(48 + n % 10)); n /= 10; }
        return string(buf);
    }
}
```

Run:

```bash
forge fmt test_fmt.sol && cat test_fmt.sol
```

**Actual output after `forge fmt`:**

```solidity
        while (temp != 0) count++;
        temp /= 10;
        ...
        while (n != 0) len -= 1;
        buf[len] = bytes1(uint8(48 + n % 10));
        n /= 10;
```

**Expected output:**

```solidity
        while (temp != 0) {
            count++;
            temp /= 10;
        }
        ...
        while (n != 0) {
            len -= 1;
            buf[len] = bytes1(uint8(48 + n % 10));
            n /= 10;
        }
```

**Impact**

This is a silent semantic-breaking transformation — the formatted code compiles without any errors or warnings, but has completely different runtime behavior:

| Loop | Before `forge fmt` | After `forge fmt` |
|---|---|---|
| `while (temp != 0) { count++; temp /= 10; }` | ✅ correct | ❌ `temp /= 10` moved outside loop → **infinite loop → OOG** |
| `while (n != 0) { len -= 1; buf[len] = ...; n /= 10; }` | ✅ correct | ❌ `n /= 10` moved outside loop → **infinite loop → OOG** |

We discovered this in a production codebase: a contract was formatted with `forge fmt` during development, compiled cleanly, but caused out-of-gas reverts at runtime. The bug was only found during a manual security audit. `forge fmt` gave no warning of any kind.

**Root cause (hypothesis)**

The formatter appears to treat `while (...) { stmt1; stmt2; }` written on a single line as if it were a brace-less `while (...) stmt1;`. When expanding to multi-line form, only the first statement is kept inside the loop body; all subsequent statements are emitted into the enclosing scope.

The same bug likely affects `for` and `if`/`else` with multiple statements on a single line.

Relevant source: `crates/fmt/src/formatter.rs` — while-statement visitor.

### LINKED ISSUE #13675: bug: `forge fmt` dropped statements for single-line **nested** if statements
### Component

Other (please describe)

- [x] Foundry
- [x] Foundryup

1.6.0-dev

foundryup: 1.5.0

`forge fmt`

### Operating System

None

### Describe the bug

This is a follow up for the issue #13555 . I have made the fix in #13560 but I think this issue deserves its own report.

The issue #13555 shows that, previously `forge fmt` formats the following statement incorrectly.

```solidity
// before forge fmt
if (condition) { execute(); execute(); }


// after forge fmt
if (condition) execute();
execute();
```

This got fixed, but the issue remains. Now for the **nested** if statement, `forge fmt` still formats incorrectly.

```solidity
// before forge fmt
if (condition) if (condition) { execute(); execute(); }


// after forge fmt
if (condition) if (condition) execute();
execute();
```

More incorrectly-formatted tests can be seen in the commit [daf10d2](https://github.com/foundry-rs/foundry/pull/13560/commits/daf10d27ccb68d65d2b1a67910d25b0fa9261566)

Impact-wise, we rarely see nested `if` as above, but `while (cond) if (cond) { ... }` is more likely to be seen.

quangloc99:
## Motivation

Originally I attempted to fix #13555 (now fix #13675 ). After it is closes with #13566, to be frank I was still a bit skeptical with the solution. After some testing, the following case fails:

```solidity
// before forge fmt
if (condition) if (condition) { execute(); execute(); }


// after forge fmt
if (condition) if (condition) execute();
execute();
```

A more realistic case for this is `while (cond) if (cond) { ... }`.

More tests can be seen in the commit [daf10d2](https://github.com/foundry-rs/foundry/pull/13560/commits/daf10d27ccb68d65d2b1a67910d25b0fa9261566)

## Solution

The solution presented in this PR is to change `is_multiline_block_stmt` to check the deeper structure of the statement. But besides the missing checks, I think the root cause is the caching of `single_line_stmt`: this variable is updated only at the top level of when looking at the `if` statement, and only the top `if` statement is examined carefully, while the nested `if` is only consider a normal statement. So imo my fix might not be enough.

I understand that caching here is necessary to make sure the work is done once. But to fix this root cause, a more reliable way should be calculating this variable for all nodes of the tree from bottom-up instead of top-down like this.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


zerosnacks:
cc @DaniPopes / @0xrusowsky when time permits

crates/fmt/src/state/sol.rs
```diff
@@ -2558,10 +2558,34 @@ impl<'ast> State<'_, 'ast> {
...
         empty_as_multiline: bool,
     ) -> bool {
-        if let ast::StmtKind::Block(block) = &stmt.kind {
-            return self.is_multiline_block(block, empty_as_multiline);
+        match &stmt.kind {
+            ast::StmtKind::Block(block) => self.is_multiline_block(block, empty_as_multiline),
+
+            ast::StmtKind::For { body, .. } | ast::StmtKind::While(_, body) => {
+                self.is_multiline_block_stmt(body, empty_as_multiline)
+            }
```
mablr: It seems this still lets a parent `if`/`while` classify `for` bodies as inline when the `for` body itself is single-line, but `print_for_stmt` always prints the `for` body with braces. For input like `while (c) for (;;) x();`, the first format pass emits `while (c) for (;;) { ... }` and the second pass rewrites it to `while (c) { for (;;) { ... } }`, causing the formatter idempotency check to panic. Consider treating `StmtKind::For { .. }` as multiline here and adding a regression test for `while (c) for (;;) x();` / `if (c) for (;;) x();`.

crates/fmt/src/state/sol.rs
```diff
@@ -2558,10 +2558,34 @@ impl<'ast> State<'_, 'ast> {
         stmt: &'ast ast::Stmt<'ast>,
         empty_as_multiline: bool,
     ) -> bool {
-        if let ast::StmtKind::Block(block) = &stmt.kind {
-            return self.is_multiline_block(block, empty_as_multiline);
+        match &stmt.kind {
+            ast::StmtKind::Block(block) => self.is_multiline_block(block, empty_as_multiline),
```
mablr: A braced block with exactly one nested control statement can still be misclassified as inline through this `Block(block)` path because `is_multiline_block` only checks the outer block statement count/source lines. For `if (c) { if (d) { a(); b(); } }`, the first pass drops the outer braces into `if (c) if (d) { ... }`, then the second pass restores them, which panics the idempotency check.

mablr (commented):
Looks good, just 2 nits

crates/fmt/src/state/sol.rs
quangloc99: Thank you for raising this. I fixed it by manually resolve the the single statement inside the for loop body.

I thought that `for` loop will print just print a single line for a single element so I did not consider the detection for for loop :pray:

crates/fmt/src/state/sol.rs
quangloc99: Thank you for this as well. So what I have done is to further call the `is_multiline_block_stmt` on the only statement of the block. Imo this should resolve this case.

quangloc99: What is more annoying tho is that my old approach did not consider about comment inside the block. I have added a bunch of test for the nested case and updated the detection to include the comments as well.

crates/fmt/src/state/sol.rs
```diff
@@ -2237,19 +2237,19 @@ impl<'ast> State<'_, 'ast> {
...
 
             // Handle 'catch' clauses
             for (pos, ast::TryCatchClause { name, args, block, span: catch_span }) in
                 other.iter().delimited()
             {
-                let current_block_multiline = self.is_multiline_block(block, false);
+                let current_block_non_empty = !block.stmts.is_empty();
                 if !pos.is_first || !skip_ind {
-                    if prev_block_multiline && (current_block_multiline || pos.is_last) {
+                    if prev_block_non_empty && (current_block_non_empty || pos.is_last) {
```
quangloc99: This change is of `print_try_stmt`. With the latest change, the test for this kinda break.
I change this just detect if the block is non-empty, since the braces are printed either way, and based on the code and test, I guess it only intend for the case of empty catch clause.

crates/fmt/testdata/OperatorExpressions/120.fmt.sol
quangloc99, line 1: This test changed as there are comments inside the if condition, and the code now promotes _if condition_ with comments to multiline.

We could keep the old behavior by checking the nested level, but I think it is more complicated and might be more error-prone. So I decided to keep the current detection code for simplicity, imo it is an OK compromise.

crates/fmt/src/state/sol.rs
quangloc99: Never mind I reuse the `is_mlutiline_block` with the new flag `forge_single_as_multi=true`. I think the addition of the new flag come very handy.

mattsse:
Thanks for addressing the earlier nested-block and `for` cases. I think there is still one related gap around nested `while` headers with comments.

The current [`While(_, body)` arm](https://github.com/foundry-rs/foundry/blob/068c46e8fb3ea5eb854b388b88005c62feda68aa/crates/fmt/src/state/sol.rs#L2580) only looks at the body, so a parent `if` can still cache `inline = true` before the nested `while` gets to run the comment-aware check in [`is_single_line_block`](https://github.com/foundry-rs/foundry/blob/068c46e8fb3ea5eb854b388b88005c62feda68aa/crates/fmt/src/state/sol.rs#L2446-L2448).

Small repro in default `preserve` mode:

```solidity
contract C {
    function f() external {
        if (a) while /* comment */ (b) { x(); }
    }
}
```

On the first pass this formats to an inlined outer `if` with a multiline `while` condition, and on the second pass it becomes fully braced, which trips the formatter idempotency check.

A fix direction could be to make the `while` branch in `is_multiline_block_stmt` account for the `while` header itself, not just its body. For example, pass enough span/context into the helper to detect comments between the `while` start and its body before the parent caches `inline = true`, or route nested `while` through the same comment-aware single-line decision used by top-level `while`/`if`. A regression test for the repro above should catch this.

quangloc99:
Hi @mattsse . Thank you for pointing out the same comment issue for the while statement. I have added test and fix as suggested.

mattsse (approved):
lgtm

mablr (approved):
Lgtm, thanks @quangloc99.

PR #13570, opened 2026-03-01
feat(evm): add `FoundryJournalExt` trait

figtracer:
Adds `FoundryJournalExt` trait implemented on `Journal<DB: DatabaseExt, JournalEntry>`. Provides borrow splitting for generic code that accesses the journal via `ctx.journal_mut()` (returns `&mut impl JournalTr`) and needs separate `&mut dyn DatabaseExt` + `&mut JournaledState` references to call `DatabaseExt` methods.

Two methods:
- `as_db_and_inner()` — splits the journal into its database and inner state components
- `set_inner()` — replaces the journal inner state (used by sub-EVM execution to write back modified state)

### DIFF
--- a/crates/evm/core/src/backend/mod.rs
+++ b/crates/evm/core/src/backend/mod.rs
@@ -20,7 +20,7 @@ use eyre::Context;
 use foundry_common::{SYSTEM_TRANSACTION_TYPE, is_known_system_sender};
 pub use foundry_fork_db::{BlockchainDb, SharedBackend, cache::BlockchainDbMeta};
 use revm::{
-    Database, DatabaseCommit, JournalEntry,
+    Database, DatabaseCommit, Journal, JournalEntry,
     bytecode::Bytecode,
     context::JournalInner,
     context_interface::{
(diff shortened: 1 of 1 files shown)


crates/evm/core/src/backend/mod.rs
```diff
@@ -380,6 +380,38 @@ pub trait DatabaseExt: Database<Error = DatabaseError> + DatabaseCommit + Debug
...
+///
+/// Generic code accesses the journal via `ctx.journal_mut()` which returns `&mut impl JournalTr`.
+/// This trait adds the ability to split the journal into its database and inner state components,
+/// enabling direct [`DatabaseExt`] method calls with zero-copy borrow splitting.
+pub trait FoundryJournalExt: JournalExt {
+    /// Returns mutable references to the database and journal inner state.
+    ///
+    /// Enables calling [`DatabaseExt`] methods directly, e.g.:
+    /// ```ignore
+    /// let (journal, env) = ctx.journal_and_env_mut();  // FoundryContextExt
```
figtracer, line 393: `FoundryContextExt` will be added here #13605

PR #13571, opened 2026-03-01
fix(cast): add confirmation prompt before overwriting keystore files

### LINKED ISSUE #13561: Keystore overwriting issue. Severity: medium.
Keystore overwriting issue. Severity: medium.

Hi guys,

Found a nasty bug today which I suggest is fixed immediately before anyone gets into trouble. It's simple to replicate.

Create a wallet and keystore with cast wallet new /home/bm/.foundry/keystores account1, skip secret.

Go into keystores and observe the account1 details with nano.

Now let's say you accidentally input the same command again, i. e.:

cast wallet new /home/bm/.foundry/keystores account1.

What happens is the keystore is immediately overwritten without prompting first. Meaning someone could loose a whole bunch of assets here. The desired function is of course to prompt before overwriting something that could contain enormous amounts of value. 

By the way, I'm hungry and there is not enough food in my home. Feel free to tip a few tokens or eth to: 0x59343299b5Cb9AEc28590d0934e644574F050f5F if you like this feedback which i myself consider to be valuable.

#Machu Picchu

0u-Y:
## Motivation

The `cast wallet new` command silently overwrites existing keystore files without any confirmation prompt. This can lead to permanent loss of wallet access and assets if users accidentally run the same command twice.

Fixes #13561 

## Solution

Added an interactive confirmation prompt that checks all target keystore files before generation starts. User must explicitly type 'y' to proceed with overwriting. The operation cancels on any other input, leaving all files unmodified.
For multiple wallets (`--number` flag), all existing files are listed upfront, ensuring an atomic operation (all or nothing).

**Example output:**
```
The following keystore file(s) already exist:
   - test_account_1
   - test_account_2

Do you want to overwrite all 2 file(s)? [y/N]:
```

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


zerosnacks:
Applied some enhancements here: https://github.com/foundry-rs/foundry/pull/13679; added you as a co-author - thanks for your PR!

PR #13579, opened 2026-03-02
feat(evm): `Backend` generic network

mablr:
## Description

- `Backend<N>` generic over Network, leveraging foundry-fork-db that has become generic as well https://github.com/foundry-rs/foundry-fork-db/pull/103
- `TxEnv` produced thanks to this powerful bound `TxEnv: FromRecoveredTx<N::TxEnvelope>`, made possible by https://github.com/foundry-rs/foundry/pull/13728
- Add `TryAnyIntoTxEnv` trait for `AnyTxEnvelope` to `TxEnv` conversion


zerosnacks:
Small test failure, looks to be a regression in `default/repros/Issue4232.t.sol:Issue4232Test`:  

```
[FAIL: vm.createSelectFork: could not instantiate forked environment with provider moonbeam-rpc.publicnode.com; deserialization error: missing field `mixHash` at line 1 column 2970
```

mablr (commented):
When forking non-standard EVM chains (e.g. Moonbeam, Arbitrum), their block headers may lack standard Ethereum fields.

It's hacky to go back to `AnyNetwork` as default since we can't recover generically `TxEnv` from `AnyTxEnvelope`. A simple workaround is to spawn an `AnyNetwork` provider on init just to get the `evm_env`, and then continue with generic one.

crates/evm/core/src/fork/multi.rs
```diff
@@ -543,8 +543,11 @@ async fn create_fork<N: Network>(
 
     let provider = fork.evm_opts.fork_provider_with_url::<N>(&fork.url)?;
 
-    // Initialise the fork environment.
-    let (evm_env, number) = fork.evm_opts.fork_evm_env(&provider).await?;
+    // Initialize the fork environment using AnyNetwork to handle non-standard EVM chains
+    // (e.g. Moonbeam, Arbitrum) whose block headers may lack standard Ethereum fields like
+    // `mixHash`. The provider typed with N is used for ongoing block/tx fetching only.
+    let any_provider = fork.evm_opts.fork_provider_with_url::<AnyNetwork>(&fork.url)?;
+    let (evm_env, number) = fork.evm_opts.fork_evm_env(&any_provider).await?;
```
0xrusowsky: couldn't this `AnyNetwork` workaround be problematic? iiuic it only protects `fork_evm_env`, but `SharedBackend<N>` still uses the "strongly-typed" provider for all subsequent rpc calls.

shouldn't the provider type be consistent across the whole fork lifetime? why wouldn't the caller be required to pass the correct network type when calling `create_fork`?

mablr: Good point, I'm going to see if I can reduce the scope of strongly typed generic `Network`. We actually just need generic Network in `Backend` to get Evm's `ContextTr::Tx` generic recovery.

mablr: I'm going to remove this workaround by introducing a conversion helper trait. Turning this to draft for now.

PR #13591, opened 2026-03-03
refactor(fuzz): simplify sample seeding

MarkFizz77:
Replace `samples_seeded` with an optional `sample_values` map in `FuzzDictionary`, unify lazy initialization paths while preserving literal and runtime sample behavior.


DaniPopes:
not needed

PR #13597, opened 2026-03-03
chore: remove duplicate dev-dependencies

strmfos:
These deps already exist in [dependencies] with same features, no need to duplicate in [dev-dependencies]


onbjerg (commented):
are there any of these that should be dev-dep only?

strmfos:
no, checked all - they're used in production code, not just tests

PR #13599, opened 2026-03-03
refactor(evm): replace `ecx` param with `depth` in `should_use_create2_factory`

figtracer:
## Summary

- Replace the `&mut EthEvmContext<&mut dyn DatabaseExt>` parameter in `InspectorExt::should_use_create2_factory` with a plain `depth: usize`
- The method only ever called `ecx.journal().depth()` on the context — the full context reference was unnecessary
- Removes the last method-level `EthEvmContext` reference from `InspectorExt`, making all 4 methods context-agnostic

This is a prerequisite for extracting the Foundry-specific methods into a separate trait (`FoundryInspectorExt`) that has no `Inspector<CTX>` supertrait, enabling generic multi-network inspector code.

### DIFF
--- a/crates/evm/traces/src/decoder/mod.rs
+++ b/crates/evm/traces/src/decoder/mod.rs
@@ -323,7 +323,7 @@ impl CallTraceDecoder {
                 self.contracts.entry(address).or_insert(contract);
             }
 
-            if let Some(label) = label {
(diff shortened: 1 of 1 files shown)


figtracer:
this was merged with #13600

PR #13536, opened 2026-02-25
feat(cheatcodes): generic `Cheatcode`

figtracer:
## Summary

Makes the cheatcode dispatch system generic over the EVM context type (`CTX`), removing the hard dependency on `EthEvmContext`.

### What changed

- **`Cheatcode` trait** becomes `Cheatcode<CTX>` with a default `CTX = ()` for cheatcodes that don't touch the EVM context
- **`CheatsCtxt`** becomes `CheatsCtxt<'a, CTX>` — the execution context passed to each cheatcode impl
- **`vm_calls!` macro** generates monomorphized match arms that dispatch to `Cheatcode::apply_full()` calls
- **~200 cheatcode impls** updated:
  - Context-free cheatcodes (base64, string, json, toml): `impl<CTX> Cheatcode<CTX> for ...`
  - Context-accessing cheatcodes (warp, roll, prank): `impl<CTX: FoundryContextExt> Cheatcode<CTX> for ...`
  - Fork/DB cheatcodes: `impl<CTX: FoundryContextExt + ContextTr<Db: DatabaseExt, Journal: JournalExt>> Cheatcode<CTX> for ...`
  - Deploy cheatcodes: `impl<CTX: NestedEvmExt + ...> Cheatcode<CTX> for ...`

### Read access is generic, write access uses inner types

- **Reads** go through `ContextTr` trait methods: `.block().number()`, `.cfg().chain_id()`, `.tx().caller()`, `.cfg().spec().into()`
- **Writes** go through `FoundryContextExt` mutators which return the inner concrete type: `block_mut() -> &mut BlockEnv`, `tx_mut() -> &mut TxEnv`, `cfg_mut() -> &mut CfgEnv`

This means each network (Eth, Op, Tempo) provides its own `FoundryContextExt` impl that unwraps its wrapper types to the shared inner types.

### What stays concrete

The `Cheatcodes` inspector struct and its `impl Inspector<EthEvmContext<...>>` remain concrete — genericizing the Inspector is the next step and builds on top of this PR.

### Prerequisites

- #13605 — `FoundryContextExt` trait (mutable env access + borrow splitting)
- #13570 — `FoundryJournalExt` trait (DB + journal state access)
- #13606 — `NestedEvm` + `NestedEvmExt` traits (sub-EVM construction)
- #13600 — `FoundryInspectorExt` (inspector methods decoupled from `EthEvmContext`)
- #13582 — `ContextTr` env accessors migration
[...]

PR #13605, opened 2026-03-03
feat(evm): add `FoundryContextExt` trait

figtracer:
Adds `FoundryContextExt` trait alongside existing `ContextExt`. Provides mutable field access to block, tx, and cfg environments for generic code that goes through `ContextTr`.

`ContextTr` only exposes immutable references (`block()`, `tx()`, `cfg()`). Cheatcodes like `vm.warp()`, `vm.roll()`, `vm.chainId()` need to mutate these fields. `FoundryContextExt` adds `block_mut()`, `tx_mut()`, `cfg_mut()`, plus:

- `to_env()` / `apply_env()` — snapshot and restore the full environment
- `journal_and_env_mut()` — simultaneous mutable access to journal and env, solving the borrow-splitting problem where `self.journal_mut()` and `self.block_mut()` would create conflicting borrows

Blanket impl provided for `Context<BlockEnv, TxEnv, CfgEnv, DB, J, C>`.

Prerequisite for #13536 (generic `Cheatcode<CTX>`). `ContextExt` will be removed once all callers migrate.

### DIFF
--- a/crates/evm/core/src/env.rs
+++ b/crates/evm/core/src/env.rs
@@ -2,6 +2,7 @@ pub use alloy_evm::EvmEnv;
 use revm::{
     Context, Database, Journal, JournalEntry,
     context::{BlockEnv, CfgEnv, JournalInner, JournalTr, TxEnv},
+    context_interface::ContextTr,
     primitives::hardfork::SpecId,
 };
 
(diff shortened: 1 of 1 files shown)


crates/evm/core/src/env.rs
```diff
@@ -112,3 +113,64 @@ impl<DB: Database, C> ContextExt
...
+
+    /// Returns mutable references to the journal and environment simultaneously.
+    ///
+    /// This solves the borrow-splitting problem: calling `self.journal_mut()` and
+    /// `self.block_mut()` separately would both borrow `&mut self`. This method
+    /// splits the borrows at the field level in one call.
+    fn journal_and_env_mut(&mut self) -> (&mut Self::Journal, EnvMut<'_>);
+}
+
+impl<DB: Database, J: JournalTr<Database = DB>, C> FoundryContextExt
```
figtracer, line 147: https://github.com/alloy-rs/evm/pull/304 would solve having to impl this for every network

mattsse (approved):
makes sense

I think havin multiple manual impls of this isnt too bad, for now

PR #13616, opened 2026-03-04
chore(cli-markdown): remove redundant clap dev-dependency

edgarr1986:
Remove the duplicate `clap` declaration from `dev-dependencies` in `foundry-cli-markdown`, add `unused_crate_dependencies` lint at crate root to catch future dependency drift.


DaniPopes:
the is fine as is

PR #13621, opened 2026-03-04
chore(cheatcodes): use `Cfg`/`Transaction` trait methods in `Inspector` impl

figtracer:
Same pattern as #13582 and #13581. Prep for generic `Inspector<CTX>` impl.

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -55,8 +55,7 @@ use revm::{
     Inspector,
     bytecode::opcode as op,
     context::{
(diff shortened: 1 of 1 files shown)


DaniPopes:
please fmt

crates/cheatcodes/src/inspector.rs
```diff
@@ -895,7 +898,8 @@ impl Cheatcodes {
                     }
 
                     let input = TransactionInput::new(call.input.bytes(ecx));
-                    let chain_id = ecx.cfg().chain_id;
+
```
DaniPopes:
```suggestion
```

PR #13622, opened 2026-03-04
feat(cast): `estimate` generic network

mablr:
## Motivation

Seamless generic `Network` support (Tempo/Ethereum) for `cast estimate`

### DIFF
--- a/crates/cast/Cargo.toml
+++ b/crates/cast/Cargo.toml
@@ -50,7 +50,6 @@ alloy-provider = { workspace = true, features = [
 alloy-rlp.workspace = true
 alloy-rpc-types = { workspace = true, features = ["eth", "trace"] }
 alloy-rpc-types-beacon.workspace = true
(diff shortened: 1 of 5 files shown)


crates/cast/src/cmd/estimate.rs
```diff
@@ -78,10 +82,21 @@ pub enum EstimateSubcommands {
...
 impl EstimateArgs {
     pub async fn run(self) -> Result<()> {
+        if self.tx.tempo.is_tempo() {
+            self.run_generic::<TempoNetwork>().await
+        } else {
+            self.run_generic::<AnyNetwork>().await
+        }
+    }
+
+    pub async fn run_generic<N: Network>(self) -> Result<()>
```
mattsse: maybe we rename to run_with_network ?

mablr: Good idea 👍

let’s use this naming everywhere

PR #13637, opened 2026-03-05
chore(anvil): clean-up impersonated tx building

mablr:
## Motivation

Replace the ad-hoc pattern of constructing a bypass signature externally and passing it to the generic `build_typed_transaction` with a oneshot `build_impersonated` helper that owns the zeroed signature internally.

- Remove `Ethapi::impersonated_signature`
- Eliminate unnecessary `Result` wrapper
- Ensures all call sites use a consistent signature rather than constructing it independently.

... tl;dr less code 😄

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -1,6 +1,6 @@
 use super::{
     backend::mem::{BlockRequest, DatabaseRef, State},
-    sign::build_typed_transaction,
+    sign::build_impersonated,
 };
 use crate::{
     ClientFork, LoggingManager, Miner, MiningMode, StorageInfo,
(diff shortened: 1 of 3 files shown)


onbjerg (approved):

ngotchac:
This introduces a bug, as transactions from impersonated accounts are now built w/ `r == s == 0`, while both values were at 1 before. It breaks at least the go-ethereum Go client, which errors with `invalid transaction v, r, s values`.

I suggest replacing
```
Signature::new(Default::default(), Default::default(), false)
```

back to
```rust
Signature::from_scalars_and_parity(
    B256::with_last_byte(1),
    B256::with_last_byte(1),
    false,
)
```

PR #13641, opened 2026-03-05
refactor(evm): generic `Inspector` impl for `InspectorStack`

figtracer:


crates/evm/evm/src/inspectors/stack.rs
```diff
@@ -692,49 +692,50 @@ impl InspectorStackRefMut<'_> {
...
-            // need to reset the env in case it was modified via cheatcodes during execution
-            *env.cfg = evm.cfg.clone();
-            *env.block = evm.block.clone();
+                evm.transact(modified_env.tx)
+            };
 
-            *env.tx = cached_env.tx;
-            env.block.basefee = cached_env.evm_env.block_env.basefee;
+            // Restore the env, discarding any cheatcode changes from inside the nested EVM.
+            // TODO: add `to_env()` to `NestedEvm` to preserve cheatcode cfg/block changes.
```
figtracer: this is important, will push a fix for this method and for `executeTransactionCall` too since the pattern repeats there

PR #13644, opened 2026-03-05
chore(anvil): use `BlockResponse`/`HeaderResponse` traits

mablr:
## Motivation

Use `BlockResponse`/`HeaderResponse`/`BlockHeader` traits everywhere to prepare generic `Network` support

+ this simplifies `convert_to_full_block` method impl.


mattsse (approved):
nice

PR #13647, opened 2026-03-05
refactor(anvil): generalize fork types over Network

stevencartavia:
Makes `ClientFork`, `ClientForkConfig`, and `ForkedStorage` in anvil generic over `N: Network = AnyNetwork`.

- `ClientForkConfig<N>`: provider field becomes `Arc<RetryProvider<N>>` (resolves the `// TODO make provider agnostic` comment)
- `ForkedStorage<N>`: cached blocks/transactions use `N::BlockResponse`/`N::TransactionResponse`
- `ClientFork<N>`: propagates `N` through config and storage fields

Network-agnostic methods (traces, proofs, storage, balances, etc.) are in `impl<N: Network> ClientFork<N>`. Methods depending on `AnyRpcBlock`/`AnyRpcTransaction` internals remain in `impl ClientFork`.

All three types default to `AnyNetwork`, so no downstream changes needed.

### DIFF
--- a/crates/anvil/src/eth/backend/fork.rs
+++ b/crates/anvil/src/eth/backend/fork.rs
@@ -4,7 +4,7 @@ use crate::eth::{backend::db::Db, error::BlockchainError, pool::transactions::Po
 use alloy_consensus::{BlockHeader, TrieAccount};
 use alloy_eips::eip2930::AccessListResult;
 use alloy_network::{
-    AnyNetwork, AnyRpcBlock, AnyRpcTransaction, BlockResponse, Network, TransactionResponse,
+    AnyNetwork, AnyRpcBlock, BlockResponse, Network, TransactionResponse,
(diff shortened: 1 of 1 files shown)


crates/anvil/src/eth/backend/fork.rs
```diff
@@ -302,6 +205,170 @@ impl ClientFork {
...
+
+        self.database.write().await.insert_block_hash(U256::from(number), block_hash);
+
+        Ok(())
+    }
+
+    /// Sends `eth_call`
+    pub async fn call(
+        &self,
+        request: &WithOtherFields<TransactionRequest>,
```
mablr: The point of being generic over `Network` is to get rid of such concrete rpc types, and have instead `N::TransactionRequest`.

0xrusowsky (approved):
makes sense

PR #13649, opened 2026-03-06
chore(script): remove `EitherSigner` abstraction

mablr:
## Motivation

Remove `EitherSigner` abstraction, to facilitate future `Network` generic propagation to `BrowserSigner`. Related to #13648

... less code btw 😎 !


mattsse (approved):
nice

PR #13651, opened 2026-03-06
feat(cheatcodes): inversion of control for `CheatcodesExecutor`

figtracer:
## Summary
- Remove `get_inspector() -> Box<dyn InspectorExt>` from `CheatcodesExecutor`
- Instead, the executor assembles the inspector internally and owns the operation
- New methods: `with_nested_evm`, `with_fresh_nested_evm`, `transact_on_db`, `transact_from_tx_on_db`
- `InspectorExt` no longer appears on the `CheatcodesExecutor` trait boundary

**Stack**: #13650 → PR 2/3 → #next

Depends on #13650

## Test plan
- [x] `cargo check` passes (zero errors, zero warnings)
- [x] All cheatcode callsites updated mechanically

PR #13652, opened 2026-03-06
refactor(evm): replace NestedEvmExt trait with free function

figtracer:
## Summary
- Replace `NestedEvmExt` trait with `with_cloned_context` free function
- The function handles the network-agnostic work: clone env + journal, split db borrow, write back
- Callers build their own EVM inside a callback that receives `(&mut dyn DatabaseExt, Env, JournaledState)`
- Remove `new_nested_evm` from trait and `new_evm_with_existing_context` (dead code)
- Simplify `CheatsCtxExt` to just `FoundryContextExt` with journal/db bounds (one less trait)

**Stack**: #13650 → #13651 → PR 3/3

Depends on #13651

## Test plan
- [x] `cargo check` passes (zero errors, zero warnings)
- [x] All callers updated: `InspectorStackInner`, `TransparentCheatcodesExecutor`

### DIFF
--- a/crates/evm/core/src/evm.rs
+++ b/crates/evm/core/src/evm.rs
@@ -67,25 +67,6 @@ pub fn new_evm_with_inspector<'db, I: InspectorExt>(
     evm
 }
 
-pub fn new_evm_with_existing_context<'a>(
-    ctx: EthEvmContext<&'a mut dyn DatabaseExt>,
-    inspector: &'a mut dyn InspectorExt,
-) -> FoundryEvm<'a, &'a mut dyn InspectorExt> {
-    let spec = ctx.cfg.spec;
-
-    let mut evm = FoundryEvm {
-        inner: RevmEvm::new_with_inspector(
-            ctx,
-            inspector,
-            EthInstructions::default(),
-            get_precompiles(spec),
-        ),
-    };
-
-    evm.inspector().get_networks().inject_precompiles(evm.precompiles_mut());
-    evm
-}
-
 /// Get the precompiles for the given spec.
 fn get_precompiles(spec: SpecId) -> PrecompilesMap {
     PrecompilesMap::from_static(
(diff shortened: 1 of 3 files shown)


crates/evm/core/src/evm.rs
```diff
@@ -304,91 +285,63 @@ impl<I: InspectorExt> NestedEvm for FoundryEvm<'_, I> {
...
+///
+/// This trait does **not** mention any inspector type — callers bring their own.
 pub trait NestedEvmExt: FoundryContextExt {
-    /// Clones the current context (env + journal), builds a nested EVM, runs the closure,
-    /// and writes modified state back. Used by `exec_create` (`deployCode`).
+    /// Clones the current context (env + journal), passes the database, cloned env,
+    /// and cloned journal inner to the callback. The callback builds whatever EVM it
+    /// needs, runs its operations, and returns `(result, modified_env, modified_journal)`.
+    /// The trait impl writes the modified state back.
     fn with_nested_evm<R>(
```
figtracer: this should definitely be renamed since now it doesn't make any sense. something like `with_cloned_context` 

`NestedEvmExt` owns splitting and provides what we need to build a EVM, and `CheatcodesExecutor` assembles it

mablr: +1

Later it will take the form alloy-evm's `EvmFactory`

figtracer:
> Free function `with_cloned_context` looks good to me.
> 
> I'm not sure about adding another abstraction via `CheatsCtxExt` trait for the moment, as the bounds are still subject to change (eg: `FoundryContextExt` support only Ethereum EVM). Let's stabilize this first? wdyt @figtracer 

imo we should proceed with this if it makes sense atm, if anything changes we come back to it?

i believe that having this extra-bounded version of it can help to frame the problem in the future if that makes sense and we stack the more refined ergonomics on top of it

PR #13660, opened 2026-03-07
fix(cli): avoid panic on invalid remappings env and validate config path

edgarr1986:
Invalid remappings env inputs now produce a warning instead of a panic

### DIFF
--- a/crates/cli/src/opts/build/paths.rs
+++ b/crates/cli/src/opts/build/paths.rs
@@ -11,7 +11,7 @@ use foundry_config::{
     find_project_root, remappings_from_env_var,
 };
 use serde::Serialize;
-use std::{ffi::OsStr, path::PathBuf};
+use std::path::PathBuf;
 
 /// Common arguments for a project's paths.
 #[derive(Clone, Debug, Default, Serialize, Parser)]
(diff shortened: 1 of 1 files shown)


crates/cli/src/opts/build/paths.rs
```diff
@@ -81,15 +86,94 @@ impl ProjectPathOpts {
...
 }
 
+/// Parses and validates `--config-path`.
+fn parse_config_path(path: &str) -> Result<PathBuf, String> {
+    let path = PathBuf::from(path);
+    if !path.exists() {
+        return Err(format!("config-path `{}` does not exist", path.display()));
+    }
+    if path.file_name() != Some(OsStr::new(Config::FILE_NAME)) {
+        return Err("the config-path must be a path to a foundry.toml file".to_string());
```
onbjerg: we should not enforce the file name. if you specify `--config-path`, it's perfectly valid to point to e.g. `foundry.dev.toml` or `foundry.example.toml` or whatever else you want

onbjerg (dismissed):
in the future I would recommend splitting PRs more. for example, the remapping change looks good to me in this PR, but the config path validation does not. had it been split we could merge the remappings fix without blocking on the latter:)

crates/cli/src/opts/build/paths.rs
edgarr1986: @onbjerg  Updated as requested, thanks for your time!

figtracer (approved):

mablr (approved):
Nice, thank you.

PR #13662, opened 2026-03-07
feat(common): generic `TransactionMaybeSigned`

mablr:
## Motivation

Towards generic `Network` in script/cheatcodes

## Solution

- Make `TransactionMaybeSigned` generic.
- Defaults to `Ethereum` to limit the scope and avoid breaking changes for now


crates/common/src/transactions.rs
```diff
@@ -160,25 +161,28 @@ where
...
-pub enum TransactionMaybeSigned {
+pub enum TransactionMaybeSigned<N: Network = Ethereum> {
     Signed {
         #[serde(flatten)]
-        tx: TxEnvelope,
+        tx: N::TxEnvelope,
         from: Address,
     },
-    Unsigned(WithOtherFields<TransactionRequest>),
+    Unsigned(N::TransactionRequest),
```
mablr, line 170: Selecting `Ethereum` as the default network has a side effect of changing the `Unsigned` variant from `WithOtherFields<TransactionRequest>` to simply `TransactionRequest`. This is acceptable because `TransactionMaybeSigned` is only used in the context of scripts or cheatcodes (Ethereum-only).

The reason why it was `WithOtherFields<TransactionRequest>` before is due to the `estimate_gas` helper in script crate using `Provider<AnyNetwork>`.

mattsse (approved):

PR #13665, opened 2026-03-08
perf(fuzz): reuse SlotIdentifier per account

edgarr1986:
Build SlotIdentifier once per account and reuse it across storage slots, reducing repeated construction while preserving slot type identification behavior.


grandizzy (approved):
makes sense, thanks!

PR #13667, opened 2026-03-08
feat(evm): add `as_any_mut()` to `FoundryInspectorExt`

figtracer:
## Summary

Adds an `as_any_mut()` method to `FoundryInspectorExt` that returns `&mut dyn Any`, enabling type recovery via downcasting.

This is a prerequisite for `FoundryEvmFactory` — the factory receives `&mut dyn FoundryInspectorExt` (network-agnostic) and needs to recover the concrete inspector type (e.g. `InspectorStack`) to construct a network-specific EVM.

### Changes

- Added `as_any_mut(&mut self) -> &mut dyn std::any::Any` to `FoundryInspectorExt`
- Implemented on all types: `NoOpInspector`, `AccessListInspector`, `Cheatcodes`, `InspectorStack`, `LogCollector`
- `InspectorStackRefMut` uses `unimplemented!()` — its non-`'static` lifetime prevents `Any`, and it's never passed to the factory

### Why `as_any_mut` instead of `as_inspector_ext`

An earlier design had `as_inspector_ext() -> &mut dyn InspectorExt` which leaked the Eth-specific `InspectorExt` into the network-agnostic trait. The `Any` approach keeps `FoundryInspectorExt` completely network-agnostic — each factory implementation downcasts to whatever concrete type it needs.

### DIFF
--- a/crates/evm/core/src/lib.rs
+++ b/crates/evm/core/src/lib.rs
@@ -75,9 +75,8 @@ pub trait FoundryInspectorExt {
 
     /// Returns a mutable reference to the concrete type as `dyn Any`, for downcasting.
     ///
-    /// Used by [`FoundryEvmFactory`](crate::evm::FoundryEvmFactory) implementations to recover
-    /// the concrete inspector type (e.g. `InspectorStack`) from the network-agnostic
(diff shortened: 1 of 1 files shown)


figtracer:
Merged into #13668

PR #13533, opened 2026-02-24
feat(cast): generic `CastTxBuilder`

mablr:
## Motivation
- Leverage new `FoundryTransactionBuilder` trait for seamless generic `TransactionRequest` building in `CastTxBuilder`
- Provide generic `cast mktx` using generic `CastTxBuilder` by auto-selecting the `Network` based on `TransactionOpts`.
 
 ## Status
 - [x] Bumped Alloy 2.0 RC (s/o @stevencartavia)
 - [x] Ready to merge (if you don't mind tmp cargo patches until Alloy 2.0 actual release)

PR #13668, opened 2026-03-08
feat(evm): introduce `FoundryEvmFactory` 

figtracer:
## Summary
- Replace `&mut dyn InspectorExt` with an execute callback in `DatabaseExt::transact` / `transact_from_tx` — callers provide a closure that captures their concrete inspector, removing the Eth-specific `InspectorExt` from the `dyn DatabaseExt` boundary
- Introduce `FoundryEvmFactory` trait in `foundry-evm-core` — abstracts EVM construction so `Backend::inspect` / `CowBackend::inspect` delegate to the factory instead of calling `new_evm_with_inspector` directly
- Add `EthFoundryEvmFactory` in `foundry-evm` — concrete Eth implementation using `InspectorStack` and `new_evm_with_inspector`
- `Executor` stores `evm_factory: EthFoundryEvmFactory` and threads it through inspect calls

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -37,7 +37,7 @@ use foundry_evm_core::{
     backend::{DatabaseError, DatabaseExt, FoundryJournalExt, RevertDiagnostic},
     constants::{CHEATCODE_ADDRESS, HARDHAT_CONSOLE_ADDRESS, MAGIC_ASSUME},
     env::FoundryContextExt,
-    evm::{NestedEvm, NestedEvmClosure, new_evm_with_inspector, with_cloned_context},
+    evm::{NestedEvm, NestedEvmClosure, new_eth_evm_with_inspector, with_cloned_context},
 };
 use foundry_evm_traces::{
     TracingInspector, TracingInspectorConfig, identifier::SignaturesIdentifier,
@@ -173,7 +173,7 @@ impl<CTX: EthCheatCtx> CheatcodesExecutor<CTX> for TransparentCheatcodesExecutor
         f: NestedEvmClosure<'_, CTX::Block, CTX::Tx, <CTX::Cfg as Cfg>::Spec>,
     ) -> Result<(), EVMError<DatabaseError>> {
         with_cloned_context(ecx, |db, evm_env, tx_env, journal_inner| {
-            let mut evm = new_evm_with_inspector(db, evm_env, tx_env, cheats);
+            let mut evm = new_eth_evm_with_inspector(db, evm_env, tx_env, cheats);
             *evm.journal_inner_mut() = journal_inner;
             f(&mut evm)?;
             let sub_evm_env = evm.to_evm_env();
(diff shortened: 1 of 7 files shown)


crates/cheatcodes/src/inspector.rs
```diff
@@ -2000,6 +2000,10 @@ impl FoundryInspectorExt for Cheatcodes {
     fn create2_deployer(&self) -> Address {
         self.config.evm_opts.create2_deployer
     }
+
+    fn as_any_mut(&mut self) -> &mut dyn std::any::Any {
+        self
+    }
```
mablr: This `Any` approach is interesting because it offers remarkable flexibility, at the cost of safety. In contrast, the "early network dispatch" proposal in cast #13533 employs compile-time `Network` generic, which, while invasive and restrictive, provide more robustness.

Let’s thoroughly discuss the potential benefits and side effects before implementing one of these two approaches in the evm crate.

crates/evm/core/src/evm.rs
figtracer: this should also just be moved to `foundry-evm` or renamed since it doesn't serve the same purpose rn but we shouldn't do this prematurely, there're still a bunch of steps to take

figtracer: for ref #13673

crates/evm/core/src/lib.rs
```diff
@@ -72,8 +72,24 @@ pub trait FoundryInspectorExt {
     fn create2_deployer(&self) -> Address {
         DEFAULT_CREATE2_DEPLOYER
     }
+
+    /// Returns a mutable reference to the concrete type as `dyn Any`, for downcasting.
+    ///
+    /// Used to recover the concrete inspector type (e.g. `InspectorStack`) from a
+    /// `&mut dyn FoundryInspectorExt` trait object.
+    fn as_any_mut(&mut self) -> &mut dyn std::any::Any;
```
0xrusowsky: we already have an `unimplemented!` in `InspectorStackRefMut`, which makes me think that if we were to go the `Any` route, it may be better to change the return type `Option<&mut dyn Any>`. That way types that can't be downcasted opt out cleanly (we should properly doc when `None` is acceptable in the trait def)

crates/evm/evm/src/evm_factory.rs
```diff
@@ -0,0 +1,42 @@
...
+/// Ethereum-flavoured [`FoundryEvmFactory`].
+#[derive(Debug, Clone, Default)]
+pub struct EthEvmFactory;
+
+impl FoundryEvmFactory for EthEvmFactory {
+    type Inspector = InspectorStack;
+    type Spec = SpecId;
+    type Block = BlockEnv;
+    type Tx = TxEnv;
+    type HaltReason = HaltReason;
```
mablr, line 24: Should we instead take advantage of alloy-evm's `EthEvmFactory` impl ?

figtracer:
closed in favor of https://github.com/foundry-rs/foundry/pull/14003, this is very stale

PR #13672, opened 2026-03-08
feat(evm): `Block`/`Tx`/`Cfg` setters traits

mablr:
## Motivation

- Foundry requires mutable `Block`/`Tx`/`Cfg` for cheatcodes.
- We want Foundry to support multiple EVMs: Eth, Op and Tempo
- For that we need to get rid of concrete fields mutations, and use instead generic setters
- But ... revm provides only getters traits `Block`/`Transaction`/`Cfg`

## Solution

- Introduced `FoundryBlock`/`FoundryTransaction`/`FoundryCfg`, respectively extending revm's `Block`/`Transaction`/`Cfg`.
- Enforced through `FoundryContextExt: ContextTr<Block: FoundryBlock, Tx: FoundryTransaction, Cfg: FoundryCfg>`
- Provides idiomatic patterns like `ctx.cfg_mut().set_spec(spec_id)`

The extension of these traits for Tempo will follow a logic similar to #13512, with "no-op" defaults.

### DIFF
--- a/crates/evm/core/src/env.rs
+++ b/crates/evm/core/src/env.rs
@@ -94,25 +94,25 @@ impl<DB: Database, J: JournalTr<Database = DB>, C> AsEnvMut
 /// Extension of [`Block`] with mutable setters, allowing EVM-agnostic mutation of block fields.
 pub trait FoundryBlock: Block {
     /// Sets the block number.
-    fn set_number(&mut self, _number: U256);
+    fn set_number(&mut self, number: U256);
 
     /// Sets the beneficiary (coinbase) address.
-    fn set_beneficiary(&mut self, _beneficiary: Address);
+    fn set_beneficiary(&mut self, beneficiary: Address);
 
     /// Sets the block timestamp.
-    fn set_timestamp(&mut self, _timestamp: U256);
+    fn set_timestamp(&mut self, timestamp: U256);
 
     /// Sets the gas limit.
-    fn set_gas_limit(&mut self, _gas_limit: u64);
+    fn set_gas_limit(&mut self, gas_limit: u64);
 
     /// Sets the base fee per gas.
-    fn set_basefee(&mut self, _basefee: u64);
+    fn set_basefee(&mut self, basefee: u64);
 
     /// Sets the block difficulty.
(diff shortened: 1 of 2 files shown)


crates/evm/core/src/env.rs
0xrusowsky, line 1: nit: imo it is cleaner to have trait defs without the underscore prefix.
i'd only add them if we were to use them in a default impl and the params were unused

crates/evm/evm/src/inspectors/stack.rs
```diff
@@ -740,21 +740,21 @@ impl InspectorStackRefMut<'_> {
     {
         let cached_env = ecx.to_env();
 
-        ecx.block_mut().basefee = 0;
-        ecx.tx_mut().chain_id = Some(ecx.cfg().chain_id());
```
0xrusowsky, line 744: looks like this line was dropped rather than converted to `.set_chain_id()`

crates/evm/core/src/env.rs
0xrusowsky, line 338: now that the bounds require `FoundryBlock`, `FoundryTransaction`, and `FoundryCfg`, shouldn't `block_mut()` / `tx_mut()` / `cfg_mut()` return `&mut Self::Block` / `&mut Self::Tx` / `&mut Self::Cfg` rather than concrete types?

otherwise ctx with different env types won't be able to impl the trait.
i guess the same problem applies to `Env` / `EnvMut`.

figtracer: > &mut Self::Block / &mut Self::Tx / &mut Self::Cfg

yes, @mablr working on this

mablr: @0xrusowsky yes that's the next stage, but we need to put generics on `Cheatcodes` first. https://github.com/foundry-rs/foundry/blob/4a4a4033e44b35ca9eb24ba09e6f6d8c1471b500/crates/cheatcodes/src/inspector.rs#L462-L463

mablr: Mb I'll fix that. 👍

zerosnacks (approved):
makes sense! pending final check @0xrusowsky

PR #13674, opened 2026-03-08
fix(cheatcodes): return error instead of panicking when no active gas snapshot

figtracer:
## Summary

- `stopGasSnapshot()` (without group/name arguments) panics with `.unwrap()` if no snapshot was previously started via `startGasSnapshot()`
- Replace with a descriptive revert error: `"no active gas snapshot; call startGasSnapshot first"`

## Test plan

- [ ] Call `vm.stopGasSnapshot()` without a prior `vm.startGasSnapshot()` — should revert with error message instead of panicking

### DIFF
--- a/crates/cheatcodes/src/evm.rs
+++ b/crates/cheatcodes/src/evm.rs
@@ -1454,15 +1454,10 @@ fn inner_stop_gas_snapshot<CTX>(
     name: Option<String>,
 ) -> Result {
     // If group and name are not provided, use the last snapshot group and name.
-    let (group, name) = match group.zip(name) {
-        Some(pair) => pair,
-        None => ccx
-            .state
-            .gas_metering
-            .active_gas_snapshot
-            .clone()
-            .ok_or_else(|| fmt_err!("no active gas snapshot; call `startGasSnapshot` first"))?,
-    };
+    let (group, name) = group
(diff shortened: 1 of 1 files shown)


crates/cheatcodes/src/evm.rs
```diff
@@ -1454,9 +1454,15 @@ fn inner_stop_gas_snapshot<CTX>(
...
-        .unwrap_or_else(|| ccx.state.gas_metering.active_gas_snapshot.clone().unwrap());
+    let (group, name) = match group.zip(name) {
+        Some(pair) => pair,
+        None => ccx
+            .state
+            .gas_metering
+            .active_gas_snapshot
+            .clone()
+            .ok_or_else(|| fmt_err!("no active gas snapshot; call `startGasSnapshot` first"))?,
+    };
```
0xrusowsky: nit: let's keep the `or_else` pattern?
```suggestion
    let (group, name) = group
        .zip(name)
        .or_else(|| ccx.state.gas_metering.active_gas_snapshot.clone())
        .ok_or_else(|| fmt_err!("no active gas snapshot; call `startGasSnapshot` first"))?;
```

figtracer: makes sense 👍

PR #13714, opened 2026-03-11
fix(wallets): remove redundant B256 length check

dizer-ti:
Remove dead code that checks `hash.as_slice().len() != 32`.

B256 is a fixed-size type (FixedBytes<32>) - its length is guaranteed at compile time. This check can never fail and serves no purpose.


zerosnacks:
Incorrect, this is the hash returned by the JSON and it is a cheap check regardless

PR #13715, opened 2026-03-11
chore(evm): simplify `fork_evm_env` by returning only block number

mablr:
## Motivation

Simplify `fork_evm_env` by returning only block number, instead of full block response.

Less generic annotations.

### DIFF
--- a/crates/evm/core/src/fork/init.rs
+++ b/crates/evm/core/src/fork/init.rs
@@ -1,6 +1,6 @@
 use crate::{Env, EvmEnv, utils::apply_chain_and_block_specific_env_changes};
 use alloy_consensus::BlockHeader;
-use alloy_primitives::{Address, U256};
+use alloy_primitives::{Address, BlockNumber, ChainId, U256};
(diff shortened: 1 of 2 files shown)


crates/evm/core/src/fork/init.rs
```diff
@@ -20,7 +20,7 @@ pub async fn environment<N: Network, P: Provider<N>>(
     disable_block_gas_limit: bool,
     enable_tx_gas_limit: bool,
     configs: NetworkConfigs,
-) -> eyre::Result<(Env, N::BlockResponse)> {
+) -> eyre::Result<(Env, u64)> {
```
zerosnacks: for clarity lets use the BlockNumber alias from https://github.com/alloy-rs/core/blob/e6b30e4c2407cd1d2ea93e79f2768e5a4f21d266/crates/primitives/src/aliases.rs#L95-L96

mablr: done 🫡

PR #13720, opened 2026-03-11
refactor(anvil): generalize Backend over Network

stevencartavia:
Parameterize Backend<N: Network>, `Blockchain<N>`, fork types, task manager.

- Split `impl Backend` into `impl<N: Network> Backend<N>` and `impl Backend<FoundryNetwork>`
- Propagate `Backend<FoundryNetwork>` to `EthApi`, `StorageInfo`, `NodeService`, and `BlockProducer`
- Add `MiningResult<T>` type alias in service.rs

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -49,7 +49,7 @@ use alloy_evm::{
     precompiles::{DynPrecompile, Precompile, PrecompilesMap},
 };
 use alloy_network::{
-    AnyHeader, AnyRpcBlock, AnyRpcHeader, AnyRpcTransaction, AnyTxEnvelope, AnyTxType,
+    AnyHeader, AnyRpcBlock, AnyRpcHeader, AnyRpcTransaction, AnyTxEnvelope, AnyTxType, Network,
     ReceiptResponse, TransactionBuilder, UnknownTxEnvelope, UnknownTypedTransaction,
 };
 use alloy_primitives::{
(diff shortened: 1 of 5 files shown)


crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -175,8 +176,7 @@ impl<T> BlockRequest<T> {
 }
 
 /// Gives access to the [revm::Database]
-#[derive(Clone, Debug)]
-pub struct Backend {
+pub struct Backend<T = FoundryTxEnvelope> {
```
figtracer: can we propagate T only for the impl methods that need it?

i believe that this forces T into types that don't need it (not 100% sure)

ref https://github.com/foundry-rs/foundry/pull/13677

stevencartavia: makes sense, I tried it but this is needed for future PRs and it will eventually be needed I think

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -235,184 +235,47 @@ pub struct Backend {
     mining: Arc<tokio::sync::Mutex<()>>,
     /// Disable pool balance checks
     disable_pool_balance_checks: bool,
+    _network: PhantomData<N>,
```
0xrusowsky: this is not necessary as `Blockchain<N>` already uses the generic

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -1166,7 +1380,9 @@ impl Backend {
 
         evm
     }
+}
 
+impl Backend<FoundryNetwork> {
```
0xrusowsky: why do we split the `impl Backend<FoundryNetwork>` block?

stevencartavia: mb 😅

zerosnacks (approved):

7zkm7b8gw9-web (commented):
?

PR #13587, opened 2026-03-02
feat(cast): `send`+`erc20` generic Network support

mablr:
## Status

Stacked on top of https://github.com/foundry-rs/foundry/pull/13533

## Insights
- Seamless generic Network support (Tempo/Ethereum) for `cast send+erc20+mktx`
- Brings Tempo tx signing support thanks to `WalletSigner` refactor https://github.com/foundry-rs/foundry/pull/13550 https://github.com/foundry-rs/foundry/pull/13648 
- Removes dummy impls in `foundry-primitives` that were introduced to match bounds on `FoundryNetwork` (rolled out here)

PR #13721, opened 2026-03-11
refactor(cast): simplify CastTxBuilder and extend Tempo transaction support

figtracer:
Refactors `CastTxBuilder` to use a `fill` flag + `.raw()` builder method instead of separate `build`/`build_raw`/`build_unsigned_raw` methods. Moves tx option application into `TransactionOpts::apply` and `TempoOpts::apply`. Adds support for Tempo sponsored transactions (`--tempo.sponsor-sig`, `--tempo.print-sponsor-hash`), expiring nonces, key IDs, and validity windows.

Based on top of https://github.com/foundry-rs/foundry/pull/13587.


crates/cast/src/cmd/access_list.rs
```diff
@@ -72,7 +72,8 @@ impl AccessListArgs {
             .await?
             .with_code_sig_and_args(None, sig, args)
             .await?
-            .build_raw(sender)
+            .raw()
+            .build(sender)
```
mablr, line 76: This is clean!

PR #13723, opened 2026-03-11
fix(config): add symmetric serialization for FuzzDictionaryConfig usize fields

gutonosa:
The three `max_fuzz_dictionary_*` fields use `deserialize_usize_or_max` to accept "max" as `usize::MAX`, but had no custom serializer. This means `usize::MAX` gets written as a raw integer that exceeds TOML's i64 range, causing serialization to fail. Added `serialize_usize_or_max` mirroring what `GasLimit` already does — same pattern, just wasn't done here.


zerosnacks:
Merging this as it is a clear fix and not controversial, cc @onbjerg / @0xrusowsky for visibility

PR #13731, opened 2026-03-12
refactor(evm): use `EvmEnv` in `CreateFork`

mablr:
## Motivation

It's useless to carry `TxEnv` in `CreateFork`, as `CfgEnv` is the source of truth for `chain_id`. 

We can simplify `MultiFork` impl to just handle `EvmEnv`.

### DIFF
--- a/crates/evm/core/src/backend/mod.rs
+++ b/crates/evm/core/src/backend/mod.rs
@@ -1010,7 +1010,7 @@ impl DatabaseExt for Backend {
 
             // merge additional logs
             snapshot.merge(current_state);
-            let BackendStateSnapshot { db, mut journaled_state, snap_evm_env, snap_tx_env } =
+            let BackendStateSnapshot { db, mut journaled_state, snap_evm_env, snap_tx_env: _ } =
                 snapshot;
             match db {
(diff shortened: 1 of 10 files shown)


crates/evm/core/src/backend/mod.rs
```diff
@@ -1075,18 +1075,18 @@ impl DatabaseExt for Backend {
...
 
         // we still need to roll to the transaction, but we only need an empty dummy state since we
         // don't need to update the active journaled state yet
         self.roll_fork_to_transaction(
             Some(id),
             transaction,
-            &mut env.evm_env,
-            &mut env.tx,
+            &mut evm_env,
+            &mut TxEnv::default(),
```
mablr, line 1089: `TxEnv` is overwritten anyway with block transactions.

https://github.com/foundry-rs/foundry/blob/56ce07b060b4ff62944c939424248a643327f767/crates/evm/core/src/backend/mod.rs#L952-L960

https://github.com/foundry-rs/foundry/blob/56ce07b060b4ff62944c939424248a643327f767/crates/evm/core/src/backend/mod.rs#L2043

mablr: I’m going to do a follow-up to simplify this.

PR #13732, opened 2026-03-12
refactor(cast): consolidate erc20 tx opts with `TempoOpts::apply`

figtracer:
Replace standalone `apply_tx_opts` free function with `Erc20TxOpts::apply` method that delegates Tempo handling to `TempoOpts::apply`. ERC20 commands now get full Tempo support (expiring nonces, key_id, valid_before/after, sponsor sig) instead of only fee_token and sequence_key.

Also adds missing `requires = "valid_before"` clap constraint on `--tempo.expiring-nonce`.

### DIFF
--- a/crates/cli/src/opts/tempo.rs
+++ b/crates/cli/src/opts/tempo.rs
@@ -54,7 +54,7 @@ pub struct TempoOpts {
     ///
     /// Sets nonce to 0 and nonce_key to U256::MAX, enabling time-bounded transaction
     /// validity via `--tempo.valid-before` and `--tempo.valid-after`.
-    #[arg(long = "tempo.expiring-nonce")]
+    #[arg(long = "tempo.expiring-nonce", requires = "valid_before")]
     pub expiring_nonce: bool,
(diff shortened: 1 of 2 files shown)


crates/cli/src/opts/tempo.rs
```diff
@@ -54,7 +54,7 @@ pub struct TempoOpts {
     ///
     /// Sets nonce to 0 and nonce_key to U256::MAX, enabling time-bounded transaction
     /// validity via `--tempo.valid-before` and `--tempo.valid-after`.
-    #[arg(long = "tempo.expiring-nonce")]
+    #[arg(long = "tempo.expiring-nonce", requires = "valid_before")]
```
zerosnacks, line 57: For context: https://github.com/tempoxyz/tempo-foundry/blob/65bea523312400dfeefec80a212dff9f915aa049/crates/cli/src/opts/tempo.rs#L66-L67, lgtm

zerosnacks (approved):

PR #13738, opened 2026-03-12
feat(forge/fmt): Allman brackets style formatting support

### LINKED ISSUE #13334: forge(fmt): option for Allman
### Component

Other (please describe)

Currently, forge fmt requires using the K&R style for opening braces. Can you add support for the Allman style via a flag?

### Additional context

Currently `forge fmt --check` always throws:
```Diff in contracts\AssetVault.sol:
2   2    | 
3   3    | import {ERC4626} from "@openzeppelin-contracts/token/ERC20/extensions/ERC4626.sol";
4   4    | 
5        |-contract AssetVault is ERC4626
6        |-{
    5    |+contract AssetVault is ERC4626 {
7   6    |      constructor(address asset, string memory name, string memory symbol) ERC4626(asset, name, symbol) {}
8   7    | }

### PR DESCRIPTION
**Pull request by Syzygy106**
## Motivation

Closes #13334

`forge fmt` currently enforces same-line opening braces, so users cannot preserve Allman-style block formatting. Allman style could be convenient for former C devs. 

## Solution

Add a new `[fmt]` config option, `brace_style`, with support for `same_line` and `allman`.

While the original request mentioned a flag, this is implemented as a `foundry.toml` formatter option to match the existing formatter configuration model in Foundry.

The implementation keeps the default behavior unchanged and applies brace-style selection across Solidity and Yul block formatting. It only affects block braces and does not change delimiter braces such as import lists or named arguments.

This feature touches several formatter paths because brace placement is currently distributed across multiple formatting entry points. That is still reasonable for the current scope, but if the formatter style system becomes more complex, this part of the architecture should be revisited.

## How to use

In `foundry.toml`:

```toml
[fmt]
brace_style = "allman" // same_line by default
```

Then run:

```forge fmt```

or:

```forge fmt --check```

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### CODE REVIEW

**Comment by Syzygy106** [comment]
During local development on my MacBook Pro I met that problem: #13741
But this doesn't happen on CI checks.

**Comment by zerosnacks** [comment]
Hi @Syzygy106 thanks for your PR - I also thought this would have been more limited in scope; cc @DaniPopes / @0xrusowsky to evaluate the implementation and whether they see value in having this.

I am personally not a fan of the `allman` style, it is quite particular but it is a well defined style.

**Comment by Syzygy106** [comment]
> Hi @Syzygy106 thanks for your PR - I also thought this would have been more limited in scope; cc @DaniPopes / @0xrusowsky to evaluate the implementation and whether they see value in having this.
> 
> 
> 
> I am personally not a fan of the `allman` style, it is quite particular but it is a well defined style.

Oh, I'm not an Allman fan too. But I did that because the following issue was approved. I hope that wasn't a pointless effort.

But yeah. I don't like that architecture too. In future upgrades it may require full refactor. However for now it should be safe anyway.

In other way the issue author may clone and fork my solution.

**Comment by 0xInnominatus** [comment]
Hi, are there any news? I am not rust developer, but I have reviewed and the code looks good.

`In other way the issue author may clone and fork my solution.`

Unfortunately, this option is not available to me, since I have no experience at all to support the rust code. But may be I can help with PR by the other way?

**Comment by Syzygy106** [comment]
> Hi, are there any news? I am not rust developer, but I have reviewed and the code looks good.
> 
> `In other way the issue author may clone and fork my solution.`
> 
> Unfortunately, this option is not available to me, since I have no experience at all to support the rust code. But may be I can help with PR by the other way?

@zerosnacks , how're doing? What would we do?

**Comment by 0xInnominatus** [comment]
@zerosnacks hi! I am sorry for ping, but may be there are any updates?

**Comment by zerosnacks** [comment]
Hey all, sorry this slipped passed. I'll do a final pass review tomorrow and merge. Thanks!

File: crates/fmt/testdata/BraceStyle/allman.fmt.sol [hunk 1]
```diff
@@ -0,0 +1,108 @@
...
+        } while (x < 2);
+
+        for (uint256 i = 0; i < 2; i++)
+        {
+            x += i;
+        }
+
+        unchecked
+    {
+        }
```
zerosnacks, line 43: This looks like an indentation bug

crates/fmt/testdata/BraceStyle/allman.fmt.sol
```diff
@@ -0,0 +1,108 @@
...
+
+        try this.bar({value: x}) returns (
+            uint256 y
+        )
+        {
+            x = y;
+        } catch Error(string memory)
+            {
```
zerosnacks, line 51: This indentation looks incorrect

crates/fmt/src/state/sol.rs
```diff
@@ -1923,12 +1979,13 @@ impl<'ast> State<'_, 'ast> {
         _ = self.handle_span(self.cursor.span(span.lo()), false);
         if !self.handle_span(span.until(block.span), false) {
             self.cursor.advance_to(span.lo(), true);
-            self.print_word("assembly "); // 9 chars
+            self.print_word("assembly"); // 8 chars
```
zerosnacks, line 1982: This looks like a regression to me in same_line mode, it is not compensated by an additional Nbsp separator

crates/fmt/testdata/BraceStyle/allman.fmt.sol
```diff
@@ -0,0 +1,108 @@
+// config: brace_style = "allman"
+contract BraceStyle
```
zerosnacks, line 2: I would like to see additional cases for `modifier`, `constructor`, `receive`, `fallback`, `library` and `interface` keywords

crates/fmt/src/state/sol.rs
```diff
@@ -2255,8 +2329,9 @@ impl<'ast> State<'_, 'ast> {
                     CommentConfig::skip_ws().mixed_no_break().mixed_post_nbsp(),
                 );
 
-                self.print_word("catch ");
+                self.print_word("catch");
                 if !args.is_empty() {
+                    self.nbsp();
```
zerosnacks, line 2334: A similar regression here

zerosnacks (changes requested):
I have some concerns about regressions in `same_line` mode near line breaks.

In particular I would like to see test cases for:

1. Assembly with dialect near line-length limit: e.g. `assembly ("memory-safe") { ... }` inside a deeply-indented function, where the line is near `line_length`. Confirms the space before `{` doesn't disappear and wrapping matches current behavior.

2. Assembly with no dialect and no flags: plain `assembly { let x := 1 }` near line-length. Confirms no break between `assembly` and `{`.

3. Bare `catch` near line-length: `try ... catch { ... }` where the `catch {` sits near the wrapping boundary. Confirms no break between `catch` and `{`.

All three should be added to the existing `BraceStyle/fmt.sol` (same_line) fixture and verified via `forge fmt --check` round-trip.

crates/fmt/src/state/sol.rs
zerosnacks: This changes the assembly printing in a way that regresses `same_line` mode:

1. `"assembly "` (hard space) became `"assembly"` + `Nbsp`: the space is now a break-point, so the printer could wrap between `assembly` and `{` under line pressure, which was impossible before.
2. The dialect/space order was swapped: original was `dialect` then `Nbsp`, now it's `Nbsp` then `dialect`. This moves the break-point from *after* the dialect string to *before* it, changing wrapping behavior.
3. The trailing `Nbsp` before `{` was moved out of the `flags` block into a separate conditional — this works for `same_line` but the structural change makes it fragile.

Suggest keeping the original `"assembly "` + `dialect` + `Nbsp` ordering and only guarding the allman-specific space removal behind `uses_allman_braces()`.

PR #13745, opened 2026-03-13
feat(cast): add `--network` flag to `cast tx` for network-specific raw encoding

mablr:
## Motivation

Replace the `FoundryNetwork`-based `transaction_raw` workaround with proper network selection via a new `--network` / `-n` CLI flag. This moves raw tx encoding back into `Cast::transaction` dispatching to the correct provider (Ethereum, Optimism, or Tempo) based on the flag.

## Solution

- Add `NetworkVariant` enum (Ethereum/Optimism/Tempo) in foundry-cli opts
- use it w/ `--network` through CastSubcommand::Tx and provider construction
- Add test for Tempo raw tx encoding (`tx_raw_tempo`)

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/cast/src/opts.rs
+++ b/crates/cast/src/opts.rs
@@ -521,12 +521,7 @@ pub enum CastSubcommand {
         to_request: bool,
 
         /// Specify the Network for correct encoding.
-        #[arg(
-            long,
-            short,
-            num_args = 0..=1,
-            value_name = "NETWORK"
-        )]
+        #[arg(long, short, num_args = 1, value_name = "NETWORK")]
         network: Option<NetworkVariant>,
     },
 


crates/cli/src/opts/network.rs
```diff
@@ -0,0 +1,11 @@
...
+#[derive(Clone, Debug, Default, clap::ValueEnum)]
+pub enum NetworkVariant {
+    /// Ethereum (default)
+    #[default]
+    Ethereum,
+    /// Optimism / OP-stack
+    Optimism,
+    /// Tempo
+    Tempo,
+}
```
zerosnacks, line 11: Makes sense, would like to get @onbjerg's input & OK on this as we had a prior conversation about this when working on `tempo-foundry`; what is preferable is if we somehow have a central location where this is maintained; I don't have an issue with the current proposal.

Context: https://github.com/foundry-rs/foundry/blob/ce695662617a24c1ed2a19f18d2c7a01a9b52f95/crates/evm/networks/src/lib.rs#L22-L37

zerosnacks:
> for cast this makes sense, however maybe it's worth thinking about:
> 
> 1. can we autodetect the network in some cases? how would we do this? (e.g. a preliminary call to get chain id with eth transport -> use concrete provider)
> 2. for forge, it would probably make sense to have this in config
> 
> wdyt?

Agreed; let's take this into consideration for the next steps @mablr and possibly do this as a follow-up

PR #13751, opened 2026-03-13
refactor(anvil): make EthApi generic over `N: Network`

stevencartavia:
Makes `EthApi<N: Network>` generic, previously it was generic over `T = FoundryTxEnvelope`, not entirely network agnostic.

- Struct now holds `Arc<Backend<N>>`, `Arc<Pool<N::TxEnvelope>>`, `Arc<Vec<Box<dyn Signer<N>>>>`, `Miner<N::TxEnvelope>`
- Network-agnostic methods in `impl<N: Network> EthApi<N>`
- Receipt and state methods remain in `impl EthApi<FoundryNetwork>` pending next steps
- All callers updated with explicit `EthApi<FoundryNetwork>`


crates/anvil/src/eth/api.rs
```diff
@@ -1798,7 +1798,7 @@ impl EthApi {
     }
 
     /// Sends a transaction and waits for receipt
-    ///
+    // TODO: generalize return type once N::ReceiptEnvelope is constructable
```
0xrusowsky: nit: weird to have doc cmnt + non-doc cmnt + doc cmnt

stevencartavia: removed!

PR #13786, opened 2026-03-16
fix(anvil): handle zero prune history

lmorett1:
## Motivation
                                                                                                                                          
  `--prune-history 0` currently reaches the state cache as `in_memory_limit = 0`. In `InMemoryBlockStates::enforce_limits()`, that turns  `while self.present.len() >= self.in_memory_limit` into an always-true condition, which can stall Anvil instead of pruning safely. This is a real availability bug caused by an unchecked boundary value.                                                                       
                                                                                                                                          
  ## Solution                                                                                                                             
                                                                                                                                          
  Treat `0` as "do not keep state history" at the config boundary, which matches the intended pruning semantics, and clamp the storage cache limit to at least `1` as a defensive backstop for any other call path. Added tests for both behaviors: config normalization for `--prune-history 0` and storage initialization with a zero limit.                                                                       
                                                                                                                                          
  ## PR Checklist                                                                                                                         
                                                                                                                                          
  - [x] Added Tests                                                                                                                       
  - [ ] Added Documentation                                                                                                               
  - [ ] Breaking changes


zerosnacks:
Hi @lmorett1 thanks for your PR, fixes lgtm; would you mind adding a regression test for this?

lmorett1:
> Hi @lmorett1 thanks for your PR, fixes lgtm; would you mind adding a regression test for this?

Hi @zerosnacks Added a regression test for --prune-history 0 that exercises the mining path and asserts no historical state is stored.

mablr (changes requested):
Could you please fix clippy errors, and make fmt.

mablr (approved):
Just fixed fmt. Sgtm. Thanks!

figtracer (approved):
lgtm, just needs conflicts

stevencartavia (commented):
looks like this was superseded by #14307

PR #13793, opened 2026-03-17
fix(cast): reject --trace for Tempo transactions in cast call

edgarr1986:
Add an upfront validation that rejects `cast call --trace` when Tempo transaction options are set, and include unit tests to lock this unsupported mode.

### DIFF
--- a/crates/cast/src/cmd/call.rs
+++ b/crates/cast/src/cmd/call.rs
@@ -220,11 +220,22 @@ impl CallArgs {
         if self.rpc.curl {
             return self.run_curl().await;
         }
+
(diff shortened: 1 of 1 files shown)


zerosnacks:
Thanks for your suggestion, I would prefer not to introduce too many LOC that we are removing soon once we have the full EVM abstraction in place (that we rely on in `--trace`), ref: https://github.com/foundry-rs/foundry/pull/13668. It is expected that during the generalization and partial to full migration of tempo-foundry into foundry we will face areas that are not yet covered. We are tracking these TODO's. Thanks!

PR #13812, opened 2026-03-17
feature(evm): owned `Tx`/`Evm` getters and `Evm` setter for `FoundryContextExt`

mablr:
## Motivation

Introduce following `FoundryContextExt` methods:
-  owned `Tx`/`Evm` getters
- `Evm` setter

with these, we will finally be able to drop all `Env::clone_evm_and_tx()` and `Env::apply_evm_and_tx()` !

### DIFF
--- a/crates/evm/core/src/env.rs
+++ b/crates/evm/core/src/env.rs
@@ -286,7 +286,7 @@ pub trait EthCheatCtx:
         Block = BlockEnv,
         Tx = TxEnv,
         Cfg = CfgEnv,
-        Journal: FoundryJournalExt,
+        Journal: FoundryJournalExt<Self>,
         Db: DatabaseExt,
     >
 {
@@ -296,7 +296,7 @@ impl<CTX> EthCheatCtx for CTX where
             Block = BlockEnv,
             Tx = TxEnv,
             Cfg = CfgEnv,
-            Journal: FoundryJournalExt,
+            Journal: FoundryJournalExt<Self>,
             Db: DatabaseExt,
         >
 {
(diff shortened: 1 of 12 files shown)


crates/evm/core/src/env.rs
```diff
@@ -228,18 +228,37 @@ pub trait FoundryContextExt:
...
     }
-    /// Sets configuration environment
+    /// Sets configuration environment.
     fn set_cfg(&mut self, cfg: Self::Cfg) {
         *self.cfg_mut() = cfg;
     }
+    /// Sets EVM environment.
+    fn set_evm(&mut self, evm_env: EvmEnv<<Self::Cfg as FoundryCfg>::Spec, Self::Block>)
+    where
+        Self::Cfg: From<CfgEnv<<Self::Cfg as FoundryCfg>::Spec>>,
```
mablr, line 246: This bound is a trick I propose to have this trait method "provided".

crates/evm/core/src/env.rs
```diff
@@ -228,18 +228,37 @@ pub trait FoundryContextExt:
...
+        *self.block_mut() = evm_env.block_env;
+    }
+    /// Cloned transaction environment.
+    fn tx_clone(&self) -> Self::Tx {
+        self.tx().clone()
+    }
+    /// Cloned EVM environment (Cfg + Block).
+    fn evm_clone(&self) -> EvmEnv<<Self::Cfg as FoundryCfg>::Spec, Self::Block>
+    where
+        Self::Cfg: Into<CfgEnv<<Self::Cfg as FoundryCfg>::Spec>>,
```
mablr, line 258: Same here.

PR #13826, opened 2026-03-18
chore(evm): remove `Env` abstraction

mablr:
## Motivation

Finally rolling out `Env` abstraction completely 🎉 !


crates/forge/src/multi_runner.rs
```diff
@@ -294,7 +294,9 @@ pub struct TestRunnerConfig {
     /// EVM configuration.
     pub evm_opts: EvmOpts,
     /// EVM environment.
-    pub env: Env,
+    pub evm_env: EvmEnv,
+    /// EVM environment.
```
zerosnacks: nit

```suggestion
    /// Transaction environment.
```

Issue #9516, opened 2024-12-07
Add `ignored_error_codes_from` config option


adhusson:
### Component

Forge

### Describe the feature you would like

It would be useful to specify a list of `(path_pattern,code)` pairs such that the error code is ignored on matching paths. 2 cases I'm aware of:

- ignore `test` contract sizes but not `src` contract sizes
- ignore `lib` contract sizes that are used for testing or their interface but were deployed with different compiler parameters to go below the size limit, but not `src` contract sizes

mimisavage:
May I handle this issue?

chiscookeke11:
Is it okay if I tackle this?

wengDavo:
Can I handle this task?

ndavd:
This can be closed now #13841

mattsse:
Implemented in #13841: `ignored_error_codes_from` is now supported in the config and takes (path, codes) pairs.

PR #13841, opened 2026-03-18
feat: add ignored_error_codes_from config option

ndavd:
Adds `ignored_error_codes_from` config option that allows ignoring error codes from files matching specific path prefixes.

Solves https://github.com/foundry-rs/foundry/issues/9516

### Example usage

Add to `foundry.toml`:

```toml
ignored_error_codes_from = [
    ["test", ["code-size"]],
    ["lib", ["code-size", "license"]],
]
```

- [x] Depends on https://github.com/foundry-rs/compilers/pull/361
  - Have temporarily patched `foundry-compilers` dep (34ebdc5d49d18a569430f76c42111db29dc3e76f) until the PR is merged
- [x] Depends on https://github.com/foundry-rs/foundry-core release

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/forge/tests/cli/config.rs
```diff
@@ -69,6 +69,7 @@ ignored_error_codes = [
     "transfer-deprecated",
     "natspec-memory-safe-assembly-deprecated",
 ]
+ignored_error_codes_from = []
```
mattsse, line 72: could you please give an example how this would look like in the toml, I assume like `[[..],[...]]`

mattsse (commented):
I think this makes a lot of sense and would be a nice addition

crates/forge/tests/cli/config.rs
ndavd: Exactly, it would look like this:

```toml
ignored_error_codes_from = [
    ["test", ["code-size"]],            # ignore contract size only in test
    ["lib", ["code-size", "license"]],  # ignore size and license in lib
]
```

Should I add that example somewhere?

crates/verify/src/etherscan/flatten.rs
```diff
@@ -89,7 +89,7 @@ impl EtherscanFlattenedSource {
         if out.errors.iter().any(|e| e.is_error()) {
             let mut o = AggregatedCompilerOutput::<SolcCompiler>::default();
             o.extend(version, RawBuildInfo::new(&input, &out, false)?, "default", out);
-            let diags = o.diagnostics(&[], &[], Default::default());
+            let diags = o.diagnostics(&[], &[], &[], Default::default());
```
mattsse: actually, this likely would also break on tempo-foundry, if we release a patch compilers release so maybe we should just do a minor @grandizzy

mattsse (dismissed):
lgtm, pending release

crates/verify/src/etherscan/flatten.rs
grandizzy: good point, yeah better to have a minor to avoid such, hopefully we'll get rid of the double maintenance soon

Cargo.toml
```diff
@@ -536,6 +536,7 @@ snapbox = { version = "0.6", features = ["json", "regex", "term-svg"] }
 idna_adapter = "=1.1.0"
 
 [patch.crates-io]
+foundry-compilers = { git = "https://github.com/ndavd/foundry-compilers", branch = "feat/ignored_error_codes_from" }
```
mablr: This needs to be removed before merge

mablr (commented):
@ndavd fyi we moved `foundry-compilers` crate to https://github.com/foundry-rs/foundry-core/tree/main/crates/compilers

Cargo.toml
ndavd: Was just checking this. I noticed foundry-compilers is still imported from the old repository. Should I change it?

ndavd: Created a PR for this https://github.com/foundry-rs/foundry/pull/14427

mablr: They will be released soon on crates.io

mablr: @ndavd fyi [foundry-compilers/0.20.0](https://crates.io/crates/foundry-compilers/0.20.0)  was released

ndavd: Thanks. It's now up-to-date with master and I've removed the patch which is no longer needed. Should be good now.

mablr (approved):

mattsse (approved):

PR #13842, opened 2026-03-19
Fix `vm.rpc` cheatcode abi encoding for structs

e1Ru1o:
## Breaking change

Breaking change: `vm.rpc` now always returns ABI-encoded bytes. Previously, the return value's memory layout varied by type, which let some users cast results directly (e.g., `string(vm.rpc(...)))`. After this change, all return types — including strings, bools, and structs — must be decoded with `abi.decode(data, (Type))`. For structs, fields that are 20-byte hex values decode as `address` and 32-byte hex values as `bytes32`, so you can use native Solidity types in your structs instead of `bytes`.

## Motivation

`vm.rpc` doesn't work as expected for methods returning JSON objects as tuples

```solidity
// This is how the return value should be handled 
// but doing so results in a revert
function getBlockDataRevert(uint256 blockNumber) internal returns (BlockRaw memory data) {
    bytes memory returnData = vm.rpc("eth_getBlockByNumber", string.concat("[\"", vm.toString(bytes32(blockNumber)), "\", false]"));
    data = abi.decode(returnData, (Block));
}

// This is a trick to get the correct object
function getBlockData(uint256 blockNumber) internal returns (BlockRaw memory data) {
    vm.rpc("eth_getBlockByNumber", string.concat("[\"", vm.toString(bytes32(blockNumber)), "\", false]"));
    bytes memory returnData;
    
    assembly ("memory-safe") {
        let ptr := mload(0x40)

        let size := returndatasize()
        mstore(ptr, size)
        returndatacopy(add(0x20, ptr), 0x00, returndatasize())
        returnData := ptr

        mstore(0x40, add(0x20, add(size, ptr)))
    }
    data = abi.decode(returnData, (Block));
}
```

## Solution

Modify `rpc_call` returned value to ensure that all objects are properly encoded into bytes

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


zerosnacks:
Hi @e1Ru1o thanks for your PR, would you mind adding regression tests for regular string / bool? 

web3_clientVersion → abi.decode(data, (string)))

net_listening → abi.decode(data, (bool))

e1Ru1o:
@zerosnacks added 👍

zerosnacks:
@e1Ru1o small fmt issue, tests lgtm - thanks!

e1Ru1o:
Should be good now @zerosnacks

decofe:
The CI failure in `testRpcTransactionByHash` is caused by a provider-dependent field in the `LegacyTransactionResult` struct.

**Root cause:** The `blockTimestamp` field is **not part of the standard Ethereum JSON-RPC spec** for `eth_getTransactionByHash`. Some providers (like publicnode) include it, but DRPC (used by CI) does not. This causes a field count mismatch — the struct expects 17 fields but only 16 are returned — which shifts all decoded values after `blockNumber` by one position.

**Fix:** Remove `blockTimestamp` from `LegacyTransactionResult`:

```diff
     struct LegacyTransactionResult {
         bytes blockHash;
         bytes blockNumber;
-        bytes blockTimestamp;
         bytes chainId;
         bytes from;
         bytes gas;
```

The other 2 failures (`testGetCodeByNameMultiVersion`) are pre-existing/unrelated.

testdata/default/cheats/Fork2.t.sol
```diff
@@ -232,10 +232,135 @@ contract ForkTest is Test {
...
+        bytes memory data = vm.rpc("sepolia", "eth_getBlockByNumber", '["0xffffffffffffff", false]');
+        // Null responses are encoded as zero bytes32
+        assertEq(data.length, 32, "null should encode as bytes32");
+    }
+
+    // Struct matching a legacy (type 0) transaction fields sorted alphabetically.
+    struct LegacyTransactionResult {
+        bytes blockHash;
+        bytes blockNumber;
+        bytes blockTimestamp;
```
zerosnacks:
```suggestion
```

e1Ru1o:
Anything missing here?

zerosnacks:
There are a few small follow-ups and I've updated the PR description to mention the breaking change, other than that lgtm - thanks!

PR #13851, opened 2026-03-19
feat: add MPP support

zerosnacks:
## feat: add MPP support

Adds transparent HTTP 402 payment handling via the [Machine Payments Protocol](https://github.com/tempoxyz/mpp-rs) (MPP). When a Tempo wallet is configured (`tempo wallet login`), the transport automatically intercepts 402 responses, pays the challenge, and retries with the credential.

> **Stacked on [#13904](https://github.com/foundry-rs/foundry/pull/13904)** (Tempo wallet signer)

### How it works

1. When Foundry creates an HTTP transport (`connect_http()`), it wraps it in `MppHttpTransport`
2. If the RPC endpoint returns **HTTP 402 Payment Required**, the transport:
   - Parses the challenge from `WWW-Authenticate`
   - Lazily discovers the Tempo wallet key (from `~/.tempo/wallet/keys.toml`)
   - Signs the challenge and opens a payment session
   - Retries the request with the payment credential
3. If no Tempo wallet is configured, a clear error with setup instructions is shown

### Key discovery

Uses the shared types from `foundry_common::tempo` (introduced in #13904):
- `TEMPO_PRIVATE_KEY` env var — checked first (ephemeral, no disk access)
- `$TEMPO_HOME/wallet/keys.toml` — read from disk (`TEMPO_HOME` defaults to `~/.tempo`)

Primary key selection (mirrors `Keystore::primary_key()` in tempo-common):
passkey entry → first entry with inline key → first entry

### Code layout

```
crates/common/src/provider/mpp/
  mod.rs        — module definition
  keys.rs       — MPP key config discovery (uses shared tempo::KeyEntry)
  session.rs    — SessionProvider with expiring nonces
  transport.rs  — MppHttpTransport + LazySessionProvider (402 challenge handling)
```

### Dependencies

- `mpp-rs` (alloy-2.0 branch) — MPP protocol types and payment provider
- `tempo` (alloy-2.0 branch) — Tempo wallet primitives and signing
- `stevencartavia/reth` — transitive dep of tempo (reth primitives)


crates/cli/src/opts/rpc.rs
```diff
@@ -73,6 +73,7 @@ pub struct RpcOpts {
     /// Print the equivalent curl command instead of making the RPC request.
     #[arg(long)]
     pub curl: bool,
+
```
zerosnacks:
```suggestion
```

onbjerg (changes requested):
this does not support websockets. can we also get a check on what the failures look like if you are using an rpc url that returns http 402 but you don't have mpp configured? we should have a cta here to "fix" it by e.g. setting up mpp.

crates/common/src/provider/runtime_transport.rs
```diff
@@ -228,6 +235,47 @@ impl RuntimeTransport {
     /// Connects to an HTTP [alloy_transport_http::Http] transport.
     fn connect_http(&self) -> Result<InnerTransport, RuntimeTransportError> {
         let client = self.reqwest_client()?;
+
+        // Auto-discover MPP key from Tempo wallet (TEMPO_PRIVATE_KEY env or
+        // ~/.tempo/wallet/keys.toml).
+        if let Some(config) = &discover_mpp_config() {
```
zerosnacks: this early initialization is wrong, we need to lazily do this upon receiving a 402

crates/common/src/provider/mpp/session.rs
```diff
@@ -0,0 +1,300 @@
...
+const VALID_BEFORE_SECS: u64 = 25;
+
+/// Default gas limit for session open transactions.
+const SESSION_OPEN_GAS_LIMIT: u64 = 2_000_000;
+
+/// Max fee per gas (20 gwei — Tempo's fixed base fee).
+const MAX_FEE_PER_GAS: u128 = 20_000_000_000;
+
+/// Max priority fee per gas.
+const MAX_PRIORITY_FEE_PER_GAS: u128 = 20_000_000_000;
```
zerosnacks, line 49: this should not be defined here

crates/common/src/provider/mpp/transport.rs
```diff
@@ -0,0 +1,570 @@
...
+    ///
+    /// Requires a funded Tempo wallet with keychain access key configured in
+    /// `~/.tempo/wallet/keys.toml` (managed by `tempo wallet`).
+    ///
+    /// ```sh
+    /// cargo test -p foundry-common test_mpp_live_pay -- --ignored --nocapture
+    /// ```
+    #[tokio::test]
+    #[ignore = "requires network access and a funded Tempo wallet"]
+    async fn test_mpp_live_pay() {
```
zerosnacks, line 789: This actually works E2E 😄

zerosnacks:
Currently missing: persistent channel storage; tempo-wallet uses SQLite for this, we may have to roll our own custom solution. Without it on every request a new channel is opened which is costly and a bad experience.

grandizzy:
> Currently missing: persistent channel storage; tempo-wallet uses SQLite for this, we may have to roll our own custom solution. Without it on every request a new channel is opened which is costly and a bad experience.

added in https://github.com/foundry-rs/foundry/pull/13920 as JSON files

zerosnacks:
Closing in favor of https://github.com/foundry-rs/foundry/pull/13934

PR #13852, opened 2026-03-19
refactor(script): `into_maybe_signed` generic over `Network`

figtracer:
## Summary
- Make `into_maybe_signed` generic over `N: Network` using `TransactionBuilder` + `FoundryTransactionBuilder` trait methods instead of Ethereum-specific struct literals and direct field access
- Use `set_input_kind(input, TransactionInputKind::Both)` as the generic equivalent of `TransactionInput::maybe_both`
- Use `N::TxEnvelope::decode_2718` instead of concrete `TxEnvelope` for signed transactions
- Remove dead `TransactionMaybeSigned::new` and `new_signed` constructors (unused after #13849)

Follows up on #13849.

### DIFF
--- a/crates/common/src/transactions.rs
+++ b/crates/common/src/transactions.rs
@@ -1,6 +1,6 @@
 //! Wrappers for transactions.
 
-use alloy_consensus::{Transaction, transaction::SignerRecoverable};
+use alloy_consensus::Transaction;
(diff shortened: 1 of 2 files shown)


figtracer:
need to think of a better approach here

PR #13858, opened 2026-03-20
fix(traces): use max instead of min for verbosity trace mode

dizer-ti:
`min` was incorrectly lowering trace mode instead of raising it.

Before:
_ => std::cmp::min(self, Self::Steps) // Debug → Steps (wrong)

After:
_ => std::cmp::max(self, Self::Steps) // Debug stays Debug (correct)


zerosnacks:
Hi @dizer-ti thanks for your PR, I've included it in #13859 and added you as a co-author.

PR #13872, opened 2026-03-21
fix: fix tempo CLI arg constraints for valid-before/valid-after

FredPhilipy:
comments said these flags require `--tempo.expiring-nonce`, but clap didn't enforce it. you could pass `--tempo.valid-before` alone and get a transaction with a time window but no expiring nonce setup. fixed, two lines.


emmajam:

PR #13874, opened 2026-03-21
refactor(verify): make verify-bytecode network-agnostic

joohhnnn:
Make `configure_env_block` and `get_runtime_codes` generic over `N: Network` instead of hardcoding `AnyNetwork`.

Replace the `AnyTxEnvelope::Ethereum` match in `verify-bytecode` with the full `FoundryTxEnvelope` conversion path, which preserves all transaction fields and handles OP deposit transactions instead of panicking on non-Ethereum transaction types.


emmajam:

PR #13875, opened 2026-03-21
test(cli): add regression tests for Tempo CLI capabilities

FredPhilipy:
The newer Tempo flags (`--tempo.sponsor-sig`, `--tempo.print-sponsor-hash`, `--tempo.expiring-nonce`,` --tempo.valid-before/after,` `--tempo.seq`, `--tempo.key-id`) had basically zero test coverage beyond the raw block/tx queries. Added 4 unit tests for TempoOpts::apply() covering expiring nonce overrides, fee token + sequence key + key id, validity windows, and the sponsor-forces-AA logic. Also added 3 CLI integration tests for `cast mktx` — print-sponsor-hash early exit, expiring nonce success path, and the clap constraint requiring `--tempo.valid-before`. Also includes the prior fix for the missing `requires` constraint on valid-before/valid-after args.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4591,6 +4591,7 @@ dependencies = [
  "strsim",
  "strum",
  "tempfile",
+ "tempo-alloy",
  "tikv-jemallocator",
  "tokio",
  "tracing",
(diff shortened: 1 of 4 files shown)


emmajam:
hey we are cleaning up stale PRs, if you still think this is valuable please reopen thank you!

PR #13882, opened 2026-03-22
cheatcodes: mark copyStorage and setArbitraryStorage as Unsafe

ParthSinghPS:
## What this PR does
Adds `safety = Unsafe` to `copyStorage` and `setArbitraryStorage` (both overloads)
in `crates/cheatcodes/spec/src/vm.rs`.

## Why
Both cheatcodes are in the `Utilities` group, which defaults to `Safety::Safe`.
Neither had an explicit safety override, so both were silently inheriting `Safe`.

This is incorrect:
- `copyStorage` directly overwrites EVM journal state by copying entire storage from
  one address to another, the same category of mutation as `store` and `etch`,
  both of which are explicitly marked `Unsafe`.
- `setArbitraryStorage` intercepts every subsequent SLOAD on the marked address and
  returns a random value. Non-deterministic storage reads in a script context are
  exactly what `Unsafe` is meant to flag.

`interceptInitcode` in the same group has similar impact and was explicitly marked
`safety = Unsafe`, suggesting these two were simply missed.

## Changes
- `crates/cheatcodes/spec/src/vm.rs`: added `safety = Unsafe` to `copyStorage`,
  `setArbitraryStorage(address)` and `setArbitraryStorage(address, bool)`
- Updated `assets/cheatcodes.json` via `cargo cheats`

### DIFF
--- a/crates/cheatcodes/assets/cheatcodes.json
+++ b/crates/cheatcodes/assets/cheatcodes.json
@@ -3758,7 +3758,7 @@
       },
       "group": "utilities",
       "status": "stable",
-      "safety": "safe"
+      "safety": "unsafe"
     },
     {
       "func": {
@@ -9868,7 +9868,7 @@
       },
       "group": "utilities",
       "status": "stable",
-      "safety": "safe"
+      "safety": "unsafe"
     },
     {
       "func": {
@@ -9888,7 +9888,7 @@
       },
       "group": "utilities",
       "status": "stable",
-      "safety": "safe"
+      "safety": "unsafe"
     },
     {
       "func": {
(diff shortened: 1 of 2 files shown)


Dlove123:
## 🎯 Claiming this Bounty!

**Plan**:
1. Mark copyStorage/setArbitraryStorage as Unsafe
2. Update documentation
3. Submit PR

## 💰 Payment Information

**PayPal**: 979749654@qq.com
**ETH**: 0x31e323edC293B940695ff04aD1AFdb56d473351D
**SOL**: 9AGDJNKgbKD4C9QaTXBWsdF2PiFjck9849ytrGqvtEAc
**RTC**: RTCb72a1accd46b9ba9f22dbd4b5c6aad5a5831572b
**GitHub**: Dlove123

PR #13884, opened 2026-03-22
fix(anvil): guard zero prune history limit

mk0walsk:
## Motivation
                                                                                                                                         
  Passing `--prune-history 0` can propagate a zero `max_memory_history` into `InMemoryBlockStates`, where `enforce_limits` loops on  `while self.present.len() >= self.in_memory_limit`. With a zero limit, that condition never becomes false, which can leave Anvil stuck  while pruning state history.                                                                                                           
                                                                                                                                         
  ## Solution                                                                                                                            
                                                                                                                                         
  Treat `0` as an unsupported prune-history value at the config layer, and clamp in-memory cache limits to at least `1` in storage as a defensive fallback. This prevents the infinite eviction loop and adds regression tests for zero-value config parsing and cache initialization.                                                                                                                        
                                                                                                                                         
  ## PR Checklist                                                                                                                        
                                                                                                                                         
  - [x] Added Tests                                                                                                                      
  - [ ] Added Documentation                                                                                                              
  - [ ] Breaking changes


zerosnacks:
Duplicate of https://github.com/foundry-rs/foundry/pull/13786

Thanks for your suggestion

Issue #13885, opened 2026-03-23
bug(script): inflated gas price estimate on Tempo


joohhnnn:
### Component

Forge

### Have you ensured that all of these are up to date?

- [ ] Foundry
- [ ] Foundryup

### What command(s) is the bug in?

forge script

### Operating System

macOS (Apple Silicon)

### Describe the bug

`forge script` shows inflated gas price estimates on Tempo mainnet (chain 4217).

```
forge script script/Deploy.s.sol:DeployScript --rpc-url https://rpc.presto.tempo.xyz/ --private-key <key>
```

Output:

```
Estimated gas price: 41000 USDC
Estimated amount required: 8358096000. USDC
```
Actual values from RPC:

- `eth_gasPrice`: 20 gwei (20000000000)
- `eth_maxPriorityFeePerGas`: 0
- `baseFeePerGas`: 20 gwei

Expected estimate should be ~40 gwei (2 * baseFee + priority), not 41000.

Additionally, the gas price unit is displayed as "gwei" which doesn't apply to Tempo's USDC gas token.

joohhnnn:
fixed in https://github.com/alloy-rs/alloy/pull/3828 waiting for review

joohhnnn:
> Additionally, the gas price unit is displayed as "gwei" which doesn't apply to Tempo's USDC gas token.

should we handle this in someways?

zerosnacks:
Thanks @joohhnnn! I'll get some eyes on the Alloy PR.

> > Additionally, the gas price unit is displayed as "gwei" which doesn't apply to Tempo's USDC gas token.
> 
> should we handle this in someways?

The gas token of Tempo is expressed as USD w/ 6 decimals

PR #13886, opened 2026-03-23
refactor(script): dedup gas currency symbol lookup

joohhnnn:
While testing Tempo compatibility, found that `forge script` displays "ETH" as the gas currency on Tempo (chain 4217), which uses USDC as its gas token.

The fix is upstream in alloy-rs/chains#266, which adds `native_currency_symbol()` for Tempo chains. Once that lands and alloy-chains is bumped, the existing `.unwrap_or("ETH")` fallback will correctly resolve to "USDC".

This PR deduplicates the symbol lookup (repeated in simulate.rs, receipts.rs, broadcast.rs) into a shared `gas_currency_symbol()` helper to keep things clean.

Also found a separate issue — gas price estimates are inflated on Tempo (~41000 gwei vs actual 20 gwei). Filed as https://github.com/foundry-rs/foundry/issues/13885

### DIFF
--- a/crates/script/src/broadcast.rs
+++ b/crates/script/src/broadcast.rs
@@ -1,5 +1,6 @@
 use std::{cmp::Ordering, sync::Arc, time::Duration};
 
+use crate::gas_currency_symbol;
 use alloy_chains::Chain;
 use alloy_consensus::{SignableTransaction, Signed};
 use alloy_eips::{BlockId, eip2718::Encodable2718};
@@ -629,7 +630,7 @@ where
                 .and_then(|avg| format_units(avg, 9).ok())
                 .unwrap_or_else(|| "N/A".to_string());
 
-            let token_symbol = crate::gas_currency_symbol(sequence.chain);
+            let token_symbol = gas_currency_symbol(sequence.chain);
             seq_progress.inner.write().set_status(&format!(
                 "Total Paid: {} {} ({} gas * avg {} gwei)\n",
                 paid.trim_end_matches('0'),
--- a/crates/script/src/broadcast.rs
+++ b/crates/script/src/broadcast.rs
@@ -1,6 +1,6 @@
 use std::{cmp::Ordering, sync::Arc, time::Duration};
 
-use alloy_chains::{Chain, NamedChain};
+use alloy_chains::Chain;
 use alloy_consensus::{SignableTransaction, Signed};
 use alloy_eips::{BlockId, eip2718::Encodable2718};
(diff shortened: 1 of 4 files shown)


crates/script/src/broadcast.rs
```diff
@@ -550,10 +550,7 @@ where
                 .and_then(|avg| format_units(avg, 9).ok())
                 .unwrap_or_else(|| "N/A".to_string());
 
-            let token_symbol = NamedChain::try_from(sequence.chain)
-                .unwrap_or_default()
-                .native_currency_symbol()
-                .unwrap_or("ETH");
+            let token_symbol = crate::gas_currency_symbol(sequence.chain);
```
zerosnacks: nit, please import on top

zerosnacks:
small merge conflict & nit; otherwise lgtm

joohhnnn:
> small merge conflict & nit; otherwise lgtm

Fixed the conflict and nit  :)

crates/script/src/lib.rs
```diff
@@ -71,6 +71,17 @@ mod simulate;
 mod transaction;
 mod verify;
 
+/// Returns the native currency symbol for gas display on a given chain.
+///
+/// All EVM chains use the same 18-decimal internal representation for gas pricing,
+/// so only the symbol changes (e.g. "ETH" vs "USDC" on Tempo).
```
zerosnacks:
```suggestion
/// so only the symbol changes (e.g. "ETH" vs "USD" on Tempo).
```

crates/script/src/lib.rs
```diff
@@ -71,6 +71,17 @@ mod simulate;
 mod transaction;
 mod verify;
 
+/// Returns the native currency symbol for gas display on a given chain.
+///
+/// All EVM chains use the same 18-decimal internal representation for gas pricing,
```
zerosnacks, line 76: FYI: For Tempo we need to format to 6 decimals

joohhnnn: @zerosnacks Thanks for the note! Could you clarify more?

On Tempo, gas prices are still in 18-decimal attodollars internally (same as wei on Ethereum), so `format_units(gas_cost, 18)` gives the correct USD amount. The 6-decimal precision only applies to TIP-20 token balances (the actual deduction from the user's wallet).

Are you suggesting we:

1. Truncate the display to 6 decimal places (since that's the real on-chain precision), or
2. Use `format_units(value, 6)` instead of 18 (which would be incorrect — off by 10^12)?

Happy to address in this PR or a follow-up, just want to make sure I understand the ask.

figtracer:
imo we can keep this as-is

PR #13892, opened 2026-03-23
refactor: simplify percent cast

eeemmmmmm:
Use `as u32` instead of `try_into()` since value is already bounds-checked


zerosnacks:
I prefer the prior, thanks

PR #13896, opened 2026-03-23
fix(ci): adapt to `TransactionMaybeSigned::to()` return type change

decofe:
Fixes master build broken by semantic merge conflict between #13895 and #13894.

#13895 changed `TransactionMaybeSigned::to()` from `Option<TxKind>` to `Option<Address>`. #13894 merged 4 minutes later without rebasing and introduced `.is_create()` on the result — which doesn't exist on `Address`.

Since `.to()` now returns `Option<Address>`, `None` already means create — `.is_none()` is the correct check.

Prompted by: zerosnacks

### DIFF
--- a/crates/script/src/execute.rs
+++ b/crates/script/src/execute.rs
@@ -196,7 +196,7 @@ impl PreExecutionState {
                 && self.args.evm.sender.is_none()
             {
                 for tx in txs {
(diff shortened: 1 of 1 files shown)


zerosnacks:
@mablr master unexpectedly broke because, can you confirm this proposed fix looks OK?

PR #13909, opened 2026-03-24
feat: Tempo wallet access key support for cast

onbjerg:
## Summary

Add support for Tempo wallet access keys (keychain mode) in `cast send` and `cast erc20`. Access keys sign on behalf of a root wallet address using V2 keychain signatures, bypassing `EthereumWallet` (which cannot be used since the signer address differs from the `from` address).

## Example Usage

Set up a Tempo access key via `tempo wallet login`, which creates `~/.tempo/wallet/keys.toml`. Then use `--from` with the **wallet address** (not the key address):

```bash
# Send a transaction using the access key
cast send 0xRecipient "transfer(address,uint256)" 0xTo 1000000 \
  --from 0x1bbb1d0e0d7c0783c6d034a97b22730cb70b57e2 \
  --rpc-url https://rpc.tempo.xyz

# ERC20 transfer
cast erc20 transfer 0xTokenAddress 0xRecipient 100 \
  --from 0x1bbb1d0e0d7c0783c6d034a97b22730cb70b57e2 \
  --rpc-url https://rpc.tempo.xyz

# ERC20 approve
cast erc20 approve 0xTokenAddress 0xSpender 100 \
  --from 0x1bbb1d0e0d7c0783c6d034a97b22730cb70b57e2 \
  --rpc-url https://rpc.tempo.xyz
```

No `--private-key` or other wallet flags are needed — the signer is automatically resolved from `keys.toml` when `--from` matches a wallet address.

## Design

1. **EthereumWallet bypass**: keychain mode requires signing with a key whose address differs from the transaction `from`. `EthereumWallet` matches signers by address, so all keychain paths sign manually and use `send_raw_transaction`.

2. **Smart key_authorization**: instead of always including `key_authorization` from `keys.toml` (which would cause `KeyAlreadyExists` errors after first provisioning), we query the AccountKeychain precompile to check if the key is already provisioned on-chain. Only includes the authorization if the key is not yet provisioned.

3. **Tempo wallet fallback**: when no explicit wallet option is provided but `--from` is set, `WalletOpts` automatically checks `~/.tempo/wallet/keys.toml` for a matching entry.

## Testing

Tested on Tempo mainnet (chain 4217):
- Wallet: `0x1bbb1d0e0d7c0783c6d034a97b22730cb70b57e2`
- Access key: `0x350b997babd15192a44268e04ff18abe29baa54a`
- `cast send` ✅
- `cast erc20 transfer` ✅

### DIFF
--- a/crates/cast/src/cmd/erc20.rs
+++ b/crates/cast/src/cmd/erc20.rs
@@ -67,27 +67,22 @@ pub struct Erc20TxOpts {
     pub tempo: TempoOpts,
 }
 
-/// Creates a provider with wallet for signing transactions locally.
-///
-/// Also returns a [`TempoAccessKeyConfig`] if the signer was resolved from a Tempo keychain entry.
-pub(crate) async fn get_provider_with_wallet<N: Network + RecommendedFillers>(
+/// Creates a provider with a pre-resolved signer.
+pub(crate) fn build_provider_with_signer<N: Network + RecommendedFillers>(
     tx_opts: &SendTxOpts,
-) -> eyre::Result<(RetryProviderWithSigner<N>, Option<TempoAccessKeyConfig>)>
+    signer: WalletSigner,
+) -> eyre::Result<RetryProviderWithSigner<N>>
 where
     N::TxEnvelope: From<Signed<N::UnsignedTx>>,
     N::UnsignedTx: SignableTransaction<Signature>,
 {
     let config = tx_opts.eth.load_config()?;
-    let (signer, tempo_access_key) = match tx_opts.eth.wallet.maybe_signer().await? {
-        (Some(s), ak) => (s, ak),
-        (None, _) => (tx_opts.eth.wallet.signer().await?, None),
-    };
     let wallet = EthereumWallet::from(signer);
     let provider = ProviderBuilder::<N>::from_config(&config)?.build_with_wallet(wallet)?;
     if let Some(interval) = tx_opts.poll_interval {
         provider.client().set_poll_interval(Duration::from_secs(interval))
     }
-    Ok((provider, tempo_access_key))
+    Ok(provider)
 }
 
 impl Erc20TxOpts {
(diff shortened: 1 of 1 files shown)


crates/cast/src/cmd/erc20.rs
```diff
@@ -328,16 +334,38 @@ impl Erc20Subcommand {
     }
 
     pub async fn run(self) -> eyre::Result<()> {
-        if let Some(erc20) = self.erc20_opts()
-            && erc20.tempo.is_tempo()
-        {
-            self.run_generic::<TempoNetwork>().await
+        // Check for Tempo access key (keychain mode) early.
+        let tempo_access_key = self.resolve_tempo_access_key().await;
```
0xrusowsky: shouldn't we lazily compute this, only if:
```rs
if self.erc20_opts().is_some_and(|erc20| erc20.tempo.is_tempo()) || self.has_wallet_from()
```

to avoid unnecessary i/o?

crates/cast/src/cmd/erc20.rs
```diff
@@ -448,78 +476,221 @@ impl Erc20Subcommand {
...
+                    cast_send(
+                        provider,
+                        tx,
+                        send_tx.cast_async,
+                        send_tx.sync,
+                        send_tx.confirmations,
+                        timeout,
+                    )
+                    .await?
+                }
```
0xrusowsky: there is a lot of duplication across variants now, this could probably be refactored to DRY the code

crates/cast/src/cmd/erc20.rs
0xrusowsky: i also noticed that both `fn resolve_tempo_access_key()` and `fn get_provider_with_wallet()` call `maybe_signer()` internally.

this means that for non-keychain erc20 txs, the signer gets resolved twice (first result is discarded).

should we match the `send.rs` pattern and resolve once in `run()` + pass the signer to `get_provider_with_wallet()` / `run_generic()`?

crates/wallets/src/tempo.rs
```diff
@@ -0,0 +1,196 @@
...
+
+/// Cryptographic key type.
+#[derive(Clone, Copy, Default, serde::Deserialize)]
+#[serde(rename_all = "lowercase")]
+enum KeyType {
+    #[default]
+    Secp256k1,
+    P256,
+    WebAuthn,
+}
```
0xrusowsky, line 31: if we plan on reusing tempo primitives once we upstream tempo, we should maybe use camelCase? not opinionated, just flagging

https://github.com/tempoxyz/tempo/blob/9f771897e3ed43298b869134bfaffa7a054b51fe/crates/primitives/src/transaction/tempo_transaction.rs#L31-#L42

onbjerg: its a toml format constraint, this is not for the tx but for the keys.toml config

crates/cast/src/cmd/erc20.rs
onbjerg: pushed a macro, not much else we can do because of the generics

PR #13913, opened 2026-03-24
perf(evm): remove unnecessary clones in do_call_end/do_create_end

figtracer:
Both methods returned `CallOutcome/CreateOutcome` but all callers discarded the return value — the outcome is already mutated in-place via &mut.

Removes the final `outcome.clone()` and changes the early-return from `then_some(outcome.clone())` to `then(|| ())`, eliminating up to 2 clones per call/create end hook invocation.


DaniPopes:
Please fix clippy

zerosnacks (approved):

PR #13916, opened 2026-03-25
feat(forge): Tempo access key support for forge create

stevencartavia:
Adds Tempo access key (keychain mode) support to `forge create`, mirroring the pattern introduced in https://github.com/foundry-rs/foundry/pull/13909 for `cast send`.

When a Tempo access key is detected in `~/.tempo/wallet/keys.toml`, `forge create --broadcast` now routes through `run_tempo_keychain` instead of `EthereumWallet`, delegating to `FoundryTransactionBuilder::sign_with_access_key` (introduced in https://github.com/foundry-rs/foundry/pull/14120) for provisioning checks and signing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5519,7 +5519,6 @@ dependencies = [
  "alloy-dyn-abi",
  "alloy-network",
  "alloy-primitives",
- "alloy-provider",
  "alloy-rlp",
  "alloy-signer",
  "alloy-signer-aws",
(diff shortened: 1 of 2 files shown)


grandizzy (changes requested):
makes sense, ty! the bulk code within `run_tempo_keychain` is duped from `run_generic` / `deploy` can we reuse / have helpers to reduce this?

PR #13904, opened 2026-03-24
feat(wallets): resolve signer from Tempo wallet keystore

zerosnacks:
## feat(wallets): resolve signer from Tempo wallet keystore

Adds automatic signer resolution from the [Tempo CLI wallet](https://docs.tempo.xyz) keystore. When `--from` is set but no explicit wallet flag is provided, Foundry looks up the address in the Tempo keystore and uses the matching private key.

### How it works

1. Checks `TEMPO_PRIVATE_KEY` env var first (ephemeral, no disk access)
2. Reads `$TEMPO_HOME/wallet/keys.toml` (default `~/.tempo/wallet/keys.toml`)
3. Matches `--from` address against `wallet_address` or `key_address` fields
4. Validates the derived signer address equals the requested sender
5. Falls back to the existing error if no match is found

### Shared types (`foundry_common::tempo`)

Introduces a `tempo` module in `foundry-common` with shared keystore types used by both the wallet signer and the MPP transport ([#13851](https://github.com/foundry-rs/foundry/pull/13851)):

- `KeyEntry` / `KeysFile` — serde models for `keys.toml`
- `WalletType` — `local` or `passkey` (for primary key selection)
- `tempo_home()` / `tempo_keys_path()` / `read_tempo_keys_file()` — discovery helpers
- `TEMPO_PRIVATE_KEY_ENV` / `TEMPO_HOME_ENV` — env var constants

### Safety

- Read/parse errors are warnings, never hard failures
- Derived address is validated against the requested sender before returning
- Keychain-style entries (where `wallet_address` differs from `key_address`) are safely skipped
- Unknown fields are ignored for forward compatibility

### Testing

```
cargo test -p foundry-wallets -- wallet_tempo
```

9 tests covering: wallet_address match, key_address match, no match, missing key, case insensitivity, empty keystore, wallet_address/key_address mismatch, unknown fields, and full discovery chain (TEMPO_HOME, env var priority, missing inline key).

Inspired by [tempoxyz/tempo-foundry](https://github.com/tempoxyz/tempo-foundry/pull/348) and [#13851](https://github.com/foundry-rs/foundry/pull/13851) (MPP key auto-discovery).

PR #13920, opened 2026-03-25
fix(mpp): key_authorization, persistent channels, topUp, e2e tests

grandizzy:
## Changes

1. **Plumb `key_authorization` from `keys.toml`** — decode RLP hex, pass to `TempoSigningMode::Keychain`. Fixes passkey/keychain wallets.
2. **Optimistic key provisioning** — skip `key_authorization` by default; on 402, retry with it. Avoids `KeyAlreadyExists`.
3. **Bump gas limit** — 2M → 10M for WebAuthn signature verification.
4. **Persistent channel storage** — `~/.tempo/foundry/channels.json`. Reuse channels across invocations (voucher-only, zero on-chain cost).
5. **Channel topUp** — when deposit exhausted, call `topUp()` on escrow instead of opening a new channel. No stranded deposits.
6. **410 recovery** — on stale channel (server restart), clear local state and surface actionable error.
7. **Configurable deposit** — `MPP_DEPOSIT` env var overrides default (100K base units).

## Testing

`scripts/mpp-test.sh` runs e2e against moderato (requires `tempo wallet login` + funded wallet):

1. `cast block-number` via MPP — opens channel, pays deposit
2. Second `cast` call — verifies channel reuse (0 spent)
3. `forge script` — runs Solidity script via MPP
4. `forge test` with `vm.createSelectFork` — fork test via MPP
5. `anvil --fork-url` — fork node via MPP
6. `chisel --fork-url` — REPL via MPP

All steps show voucher count to confirm MPP payment.

```bash
cargo build --bin cast --bin forge --bin anvil --bin chisel
./scripts/mpp-test.sh ./target/debug
```

Unit tests: `cargo test -p foundry-common -- persist::tests`

CI: `.github/workflows/mpp-e2e.yml` (requires `TEMPO_KEYS_TOML_B64` secret)

## Related

Addresses https://github.com/foundry-rs/foundry/pull/13851#issuecomment-4117540188

### DIFF
--- a/crates/cheatcodes/src/config.rs
+++ b/crates/cheatcodes/src/config.rs
@@ -176,6 +176,9 @@ impl CheatsConfig {
     pub fn rpc_endpoint(&self, url_or_alias: &str) -> Result<ResolvedRpcEndpoint> {
         if let Some(endpoint) = self.rpc_endpoints.get(url_or_alias) {
             Ok(endpoint.clone().try_resolve())
+        } else if let Some(builtin_url) = foundry_config::builtin_rpc_url(url_or_alias) {
+            let url = RpcEndpointUrl::Url(builtin_url.to_string());
+            Ok(RpcEndpoint::new(url).resolve())
         } else {
             // check if it's a URL or a path to an existing file to an ipc socket
             if url_or_alias.starts_with("http") ||
--- a/crates/config/src/endpoints.rs
+++ b/crates/config/src/endpoints.rs
@@ -453,9 +453,28 @@ impl DerefMut for ResolvedRpcEndpoints {
     }
 }
 
+/// Returns the URL for a built-in RPC alias, if one exists.
+///
+/// Built-in aliases act as fallbacks: they are only used when the alias has **not** been
+/// defined by the user in `[rpc_endpoints]` or resolved via MESC.
+pub fn builtin_rpc_url(alias: &str) -> Option<&'static str> {
+    match alias {
+        "tempo" => Some("https://rpc.mpp.tempo.xyz"),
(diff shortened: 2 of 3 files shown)


crates/common/src/provider/mpp/transport.rs
```diff
@@ -69,16 +78,36 @@ impl LazySessionProvider {
...
+                    })?;
+                    let auth =
+                        alloy_rlp::Decodable::decode(&mut bytes.as_slice()).map_err(|e| {
+                            TransportErrorKind::custom(std::io::Error::other(format!(
+                                "invalid MPP key_authorization RLP: {e}"
+                            )))
+                        })?;
+                    Ok::<_, TransportError>(Box::new(auth))
+                })
+                .transpose()?;
```
onbjerg, line 114: think we might have this in https://github.com/foundry-rs/foundry/blob/80ae59941ada986c49e05f2a8412c3fc1f9075d0/crates/wallets/src/tempo.rs#L111-L116 ?

scripts/mpp-test.sh
onbjerg, line 1: can we add a test that the call fails without mpp as well? i.e. before log in

crates/common/src/provider/mpp/transport.rs
grandizzy: mm, yeah, thing is that this PR is stacked on top of https://github.com/foundry-rs/foundry/pull/13851 which was on top of https://github.com/foundry-rs/foundry/pull/13904 (that diverged from main / didn't have `decode_key_authorization`) that was closed with https://github.com/foundry-rs/foundry/pull/13904#issuecomment-4120157911 So ig I need to redo PR13851 on top of master and then this PR on top of PR13851...

PR #13923, opened 2026-03-25
fix(chisel): remove unused `latest` variable in next_cached_session

strmfos:
The `latest` variable was computed by comparing file modification times on each iteration but never actually used - only the count mattered.
Removed unnecessary filesystem metadata calls.


crates/chisel/src/session.rs
```diff
@@ -109,24 +109,12 @@ impl ChiselSession {
...
+        // Count existing sessions to determine the next session number
+        let mut session_num = 0;
         for entry in entries {
-            let entry = entry?;
-            if entry.metadata()?.modified()? >= latest.metadata()?.modified()? {
-                latest = entry;
-            }
-
-            // Increase session_num counter rather than cloning the iterator and using `.count`
+            let _ = entry?;
```
onbjerg: we can just do `let session_num = entries.filter(Result::is_ok).count();`

onbjerg (dismissed):

crates/chisel/src/session.rs
strmfos: fixed, thanks

mablr (approved):
lgtm

figtracer (approved):
nice

PR #13939, opened 2026-03-26
fix(lint): detect unsafe casts on function returns and ternaries

### LINKED ISSUE #13921: lint misses some unsafe casts
### Component

Forge

- [x] Foundry
- [x] Foundryup

latest

latest

forge lint

### Operating System

None

### Describe the bug

forge lint misses some unsafe casts. 

https://github.com/morpho-org/forge-lint-bug-repro

ArshLabs:
Closes #13921

## Problem

The `unsafe-typecast` lint misses unsafe casts when the source expression is a function return value, ternary, or other complex expression. `infer_source_types` only handles `Ident`, `Lit`, type-cast `Call`, `Unary`, and `Binary` -- everything else falls through `_ => None`, which makes `source_types` empty and the lint assumes the cast is safe.

## Fix

- For non-cast function calls, resolve the callee's return type from the HIR (`Function.returns`) so casts like `uint128(getUint256())` are now flagged.
- For ternary expressions, recurse on both branches (matching the existing `Binary` pattern) so casts like `uint128(cond ? x : y)` is now flagged.

## Test cases added

- `uint128(getUint256())` -- function return value narrowing cast
- `uint128(cond ? x : y)` -- ternary with uint256 branches narrowed to uint128

### DIFF
--- a/crates/lint/src/sol/med/unsafe_typecast.rs
+++ b/crates/lint/src/sol/med/unsafe_typecast.rs
@@ -48,7 +48,7 @@ fn is_unsafe_typecast_hir(
     target_type: &hir::ElementaryType,
 ) -> bool {
     let mut source_types = Vec::<ElementaryType>::new();
-    infer_source_types(Some(&mut source_types), hir, source_expr);
+    infer_source_types(&mut source_types, hir, source_expr);
 
     if source_types.is_empty() {
         return false;
@@ -59,26 +59,10 @@ fn is_unsafe_typecast_hir(
 
 /// Infers the elementary source type(s) of an expression.
 ///
-/// This function traverses an expression tree to find the original "source" types.
-/// For cast chains, it returns the ultimate source type, not intermediate cast results.
-/// For binary operations, it collects types from both sides into the `output` vector.
-///
-/// # Returns
-/// An `Option<ElementaryType>` containing the inferred type of the expression if it can be
-/// resolved to a single source (like variables, literals, or unary expressions).
-/// Returns `None` for expressions complex expressions (like binary operations).
-fn infer_source_types(
-    mut output: Option<&mut Vec<ElementaryType>>,
-    hir: &hir::Hir<'_>,
-    expr: &hir::Expr<'_>,
-) -> Option<ElementaryType> {
-    let mut track = |ty: ElementaryType| -> Option<ElementaryType> {
-        if let Some(output) = output.as_mut() {
-            output.push(ty);
-        }
-        Some(ty)
-    };
-
+/// Traverses an expression tree to find the original "source" types and pushes them
+/// into `output`. For cast chains, finds the ultimate source type, not intermediate
+/// cast results. For binary and ternary operations, collects types from all branches.
+fn infer_source_types(output: &mut Vec<ElementaryType>, hir: &hir::Hir<'_>, expr: &hir::Expr<'_>) {
     match &expr.kind {
         // A type cast call: `Type(val)`, or a regular function call.
         ExprKind::Call(call_expr, args, ..) => {
@@ -88,14 +72,12 @@ fn infer_source_types(
                 && let Some(inner) = args.exprs().next()
             {
                 // Recurse to find the original (inner-most) source type.
-                return infer_source_types(output, hir, inner);
+                infer_source_types(output, hir, inner);
+                return;
             }
 
             // For non-cast function calls, infer the return type from the function signature.
-            if let Some(elem_ty) = resolve_call_return_type(hir, call_expr) {
-                return track(elem_ty);
-            }
-            None
+            resolve_call_return_types(output, hir, call_expr);
         }
 
         // Identifiers (variables)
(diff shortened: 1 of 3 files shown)


grandizzy (dismissed):
ty, lgtm! @0xrusowsky @zerosnacks mind to have one more check?

zerosnacks (changes requested):
Hi @ArshLabs

With the help of our internal AI reviewing I've identified several findings I would like to see addressed prior to merging.

## Findings

### 1. Early Return in Overload Resolution — Medium

**File:** `crates/lint/src/sol/med/unsafe_typecast.rs#L166-L180`

`resolve_function_return` iterates through function resolutions and returns the first matching single-return elementary type. For overloaded functions, `resolutions` can contain multiple functions with different return types. Returning early on the first match can lead to false negatives if the actual invoked overload returns a type that makes the cast unsafe, but an earlier overload in the list returns a safe type.

**Recommendation:** Collect and return all possible return types from all matching overloads. Flag the cast if *any* of the possible return types are unsafe.

### 2. Missing Member Access Calls — Medium

**File:** `crates/lint/src/sol/med/unsafe_typecast.rs#L154-L163`

`resolve_call_return_type` only matches `ExprKind::Ident` for direct function calls (e.g., `foo()`). It ignores `ExprKind::Member`, used for external calls and member accesses (e.g., `contract.getUint256()`). This means unsafe casts on member function returns are silently ignored.

**Recommendation:** Add a match arm for `ExprKind::Member` to resolve the member function's return type.

### 3. Unnecessary API Complexity — Low

**File:** `crates/lint/src/sol/med/unsafe_typecast.rs#L70-L148`

`infer_source_types` accepts `mut output: Option<&mut Vec<ElementaryType>>` and returns an `Option<ElementaryType>`. The root caller always passes `Some` and ignores the return value, forcing unnecessary branching and a `track` closure.

**Recommendation:** Simplify signature to `fn infer_source_types(output: &mut Vec<ElementaryType>, hir: &hir::Hir<'_>, expr: &hir::Expr<'_>)` (returning `()`).

### 4. Incomplete Test Coverage — Low

**File:** `crates/lint/testdata/UnsafeTypecast.sol#L459-L474`

Tests cover basic happy paths but miss edge cases:

- Overloaded functions with different return types
- Member access calls (e.g., `uint128(this.getUint256())`)
- Mixed-safety ternary expressions (e.g., `uint128(cond ? uint128(x) : uint256(y))`)

Thanks!

ArshLabs:
Thanks for the thorough review @zerosnacks, pushed the fixes:

1. **Overload resolution** -- `resolve_function_returns` now iterates all resolutions and pushes every match instead of returning the first one.

2. **Member access calls** -- Added `ExprKind::Member` handling in `resolve_call_return_types`. Resolves the contract from the base expression (variable with contract type or interface cast), then finds matching functions by name via `contract_item_ids`. Follows the same pattern as `unchecked_calls.rs`.

3. **API simplification** -- `infer_source_types` now takes `&mut Vec<ElementaryType>` directly and returns `()`. Removed the `Option` wrapper and the `track` closure.

4. **Test coverage** -- Added:
   - Mixed-safety ternary (`uint128(cond ? safe : unsafe_)`)
   - Member function call via interface (`uint128(foo.getUint256())`)
   - Dropped the overloaded function test since solar doesn't compile it cleanly, but the code path is covered defensively.

ArshLabs:
@zerosnacks would appreciate a re-review when you get a chance

zerosnacks:
Hi @ArshLabs thanks for the ping, I'll get back to you soon

ArshLabs:
> Hi @ArshLabs thanks for the ping, I'll get back to you soon

no worries! take your time and lemme know if something is required from my end!

ArshLabs:
> Hi @ArshLabs thanks for the ping, I'll get back to you soon

no worries! lemme know if something is required from my end!

ArshLabs:
@zerosnacks the 3 failing CI checks are pre-existing and unrelated to this PR

crates/lint/src/sol/med/unsafe_typecast.rs
```diff
@@ -59,77 +59,143 @@ fn is_unsafe_typecast_hir(
...
             {
                 // Recurse to find the original (inner-most) source type.
-                return infer_source_types(output, hir, inner);
+                infer_source_types(output, hir, inner);
+                return;
             }
-            None
+
+            // For non-cast function calls, infer the return type from the function signature.
+            resolve_call_return_types(output, hir, call_expr);
```
figtracer: this ignores call arguments when resolving return types, so selected-safe overloads can still warn

crates/lint/src/sol/med/unsafe_typecast.rs
```diff
@@ -59,77 +59,143 @@ fn is_unsafe_typecast_hir(
...
+) {
+    match &call_expr.kind {
+        // Direct function call: `foo()`
+        ExprKind::Ident(resolutions) => {
+            resolve_function_returns(output, hir, resolutions);
+        }
+        // Member function call: `contract.foo()`
+        ExprKind::Member(contract_expr, func_ident) => {
+            // Resolve the contract from the base expression.
+            let contract_id = match &contract_expr.kind {
```
figtracer: this misses `this.getUint256()` receivers, so unsafe casts on current-contract external calls remain false negatives

figtracer (commented):
nits

grandizzy (commented):
A few more cases from the issue's "member access, index, …, globals" list that aren't yet exercised by the fixture. 

False positive — inner cast through a ternary leaks the pre-cast type
```Solidity
function innerCastInTernaryIsSafe(bool cond, uint256 x, uint128 y) internal pure {
    // `x` is narrowed to uint128 in the true branch — the outer
    // cast is safe and should NOT warn. The inner `uint128(x)` is independently
    // (and correctly) flagged.
    uint128 a = uint128(cond ? uint128(x) : y);
}
```

False negatives the issue explicitly enumerates

```Solidity
// 1. Non-call Member access — the canonical unsafe casts.
function blockTimestampNarrowing() internal view returns (uint32) {
    return uint32(block.timestamp); //~WARN
}
function msgValueNarrowing() internal view returns (uint128) {
    return uint128(msg.value); //~WARN
}
function arrayLengthNarrowing(uint256[] memory data) internal pure returns (uint32) {
    return uint32(data.length); //~WARN
}
function addressBalanceNarrowing(address who) internal view returns (uint128) {
    return uint128(who.balance); //~WARN
}

// 2. Struct field reads.
struct Position { uint256 amount; uint128 shares; }

function structFieldNarrowing(Position memory p) internal pure returns (uint128) {
    return uint128(p.amount); //~WARN
}

// 3. Library / namespace member calls.
library MathLib {
    function asUint256(uint128 x) internal pure returns (uint256) { return x; }
}
function libraryMemberCallUnsafe(uint128 x) internal pure returns (uint128) {
    return uint128(MathLib.asUint256(x)); //~WARN
}

// 4. Index expressions — arrays, mappings, indexed bytes.
function arrayIndexNarrowing(uint256[] memory data) internal pure returns (uint128) {
    return uint128(data[0]); //~WARN
}
function mappingIndexNarrowing(
    mapping(address => uint256) storage balances,
    address who
) internal view returns (uint128) {
    return uint128(balances[who]); //~WARN
}

// 5. Composition — nested Index, indexed struct, nested mapping.
function nestedIndexNarrowing(uint256[][] memory matrix) internal pure returns (uint128) {
    return uint128(matrix[0][0]); //~WARN
}
function indexedStructFieldNarrowing(Position[] memory positions)
    internal pure returns (uint128)
{
    return uint128(positions[0].amount); //~WARN
}
function nestedMappingNarrowing(
    mapping(address => mapping(uint256 => uint256)) storage balances,
    address who, uint256 id
) internal view returns (uint128) {
    return uint128(balances[who][id]); //~WARN
}

// 6. Fixed-bytes `.length` is uint8 (not uint256), so this is unsafe.
function fixedBytesLengthIsUint8(bytes32 b) internal pure returns (int8) {
    return int8(b.length); //~WARN  (uint8 -> int8 changes sign)
}

// 7. External function pointer `.selector` is bytes4.
function selectorNarrowing() external pure returns (bytes2) {
    function () external pure returns (uint256) fn;
    return bytes2(fn.selector); //~WARN
}
```

@ArshLabs pls let us know if you still have bandwidth to address new comments, otherwise we could take it over. Thank you!

crates/lint/src/sol/med/unsafe_typecast.rs
```diff
@@ -59,77 +59,143 @@ fn is_unsafe_typecast_hir(
...
+                // Find matching functions in the contract by name.
+                for item in hir.contract_item_ids(cid) {
+                    let Some(fid) = item.as_function() else { continue };
+                    let func = hir.function(fid);
+                    if func.name.is_some_and(|name| name.as_str() == func_ident.as_str())
+                        && func.returns.len() == 1
+                    {
+                        let ret_var = hir.variable(func.returns[0]);
+                        if let TypeKind::Elementary(elem_ty) = &ret_var.ty.kind {
+                            output.push(*elem_ty);
```
mablr: Resolves member-call return types by function name only, without checking call arity or argument types.

This can create false positives for overloaded methods. Example:

```solidity
interface IOverloaded {
    function value(uint256 x) external view returns (uint128);
    function value(address x) external view returns (uint256);
}

contract Repro {
    function safeOverload(IOverloaded foo) external view returns (uint128) {
        return uint128(foo.value(uint256(1)));
    }
}
```

ArshLabs:
as per the feedback provided, 

false positives fixed
- Ternary/binary branches which are themselves a cast now use the cast's result type so `uint128(cond ? uint128(x) : y)` no longer emits a spurious warning on the outer cast
- Member call overload resolution now filters by arity so `foo.value(x)` wont accidentally match a same named overload with a different parameter count

new patterns detected
- Built-in globals: `block.timestamp`, `block.number`, `msg.value`, `msg.gas`, `tx.gasprice`, `msg.sender`
- Address members: `who.balance`
- Dynamic array/bytes `.length`
- Struct field reads: `p.amount` where `p` is a struct with known field types
- Array element and mapping value types via index expressions: `data[0]`, `balances[who]`
- Library/namespace calls: `MathLib.asUint256(x)`

new test cases added for each

mablr:
@ArshLabs Nice, thank you!

Could you please resolve conflicts?

crates/lint/src/sol/med/unsafe_typecast.rs
```diff
@@ -57,114 +56,306 @@ fn is_unsafe_typecast_hir(
...
+    call_expr: &hir::Expr<'_>,
+    args_count: usize,
+) {
+    match &call_expr.kind {
+        ExprKind::Ident(resolutions) => {
+            resolve_function_returns(output, hir, resolutions);
+        }
+        ExprKind::Member(base_expr, func_ident) => {
+            let contract_id = match &base_expr.kind {
+                ExprKind::Ident(reses) => {
```
mablr, line 140: This seems to miss calls through the built-in this receiver. For example, `uint128(this.getUint256())` where `getUint256()` returns `uint256` is still not flagged, because `base_expr` is `Res::Builtin(This)` and the current match only handles variables, contract idents, and interface/contract casts.

crates/lint/src/sol/med/unsafe_typecast.rs
```diff
@@ -57,114 +56,306 @@ fn is_unsafe_typecast_hir(
...
+            };
+
+            if let Some(cid) = contract_id {
+                for item in hir.contract_item_ids(cid) {
+                    let Some(fid) = item.as_function() else { continue };
+                    let func = hir.function(fid);
+                    if func.name.is_some_and(|name| name.as_str() == func_ident.as_str())
+                        && func.parameters.len() == args_count
+                        && func.returns.len() == 1
+                    {
```
mablr, line 170: Filtering member overloads by name and arity is too broad. I verified that an interface with `value(address) returns (uint128)` and `value(uint256) returns (uint256)` emits an `unsafe-typecast` warning for the safe call `uint128(foo.value(who))`, because the `uint256` overload is also collected.

mablr (commented):
smol nits :)

crates/lint/src/sol/med/unsafe_typecast.rs
```diff
@@ -22,7 +23,6 @@ impl<'hir> LateLintPass<'hir> for UnsafeTypecast {
         hir: &'hir hir::Hir<'hir>,
         expr: &'hir hir::Expr<'hir>,
     ) {
```
mablr, line 25: Master's `LateLintPass::check_expr` now takes 5 params (adds `gcx: Gcx<'hir>`)

crates/lint/src/sol/med/unsafe_typecast.rs
```diff
@@ -41,14 +41,13 @@ impl<'hir> LateLintPass<'hir> for UnsafeTypecast {
     }
 }
 
-/// Determines if a typecast is potentially unsafe (could lose data or precision).
 fn is_unsafe_typecast_hir(
     hir: &hir::Hir<'_>,
     source_expr: &hir::Expr<'_>,
     target_type: &hir::ElementaryType,
 ) -> bool {
```
mablr, line 48: On master `hir::Hir` was replaced by `Gcx`; needs to be re-plumbed through all new helpers.

crates/lint/src/sol/med/unsafe_typecast.rs
```diff
@@ -57,114 +56,312 @@ fn is_unsafe_typecast_hir(
...
+                }
+            }
+        }
+    }
+}
+
+fn resolve_expr_type<'hir>(
+    hir: &'hir hir::Hir<'hir>,
+    expr: &hir::Expr<'_>,
+) -> Option<&'hir hir::Type<'hir>> {
```
mablr, line 209: not needed anymore on master, handled in solar.

stevencartavia (commented):
the issue that this PR aims to close was addressed by #15003

stevencartavia:
@ArshLabs ty, maybe the tests can be ported :)

ArshLabs:
> @ArshLabs ty, maybe the tests can be ported :)

alr

PR #13958, opened 2026-03-27
chore: deduplicate prefixed base detection

anim001k:
`Base::detect()` implements the same prefixed parsing flow three times for binary, octal, and hexadecimal values. This duplication makes the autodetection path harder to maintain and increases the chance of future changes drifting between branches.

### DIFF
--- a/crates/cast/src/base.rs
+++ b/crates/cast/src/base.rs
@@ -139,6 +139,8 @@ impl Base {
         }
     }
 
+    /// Detects a prefixed base by stripping the 2-char prefix and parsing the remainder.
+    /// `PosOverflow` is treated as a valid match since the digits are correct for the base.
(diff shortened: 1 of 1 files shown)


crates/cast/src/base.rs
```diff
@@ -155,6 +139,16 @@ impl Base {
         }
     }
 
+    fn detect_prefixed(s: &str, radix: u32, base: Self, label: &str) -> Result<Self> {
```
zerosnacks, line 144:
```suggestion
    /// Detects a prefixed base by stripping the 2-char prefix and parsing the remainder.
    /// `PosOverflow` is treated as a valid match since the digits are correct for the base.
    fn detect_prefixed(s: &str, radix: u32, base: Self, label: &str) -> Result<Self> {
```

PR #13940, opened 2026-03-26
refactor(anvil): extract `execute_pool_transactions`

figtracer:
## Summary

Extract the ~150-line per-tx mining loop duplicated in `do_mine_block` and `with_pending_block` into a shared `execute_pool_transactions` function.

The function is generic over `B: BlockExecutor` — callers pass any executor satisfying the trait bounds. Tx env construction is a closure parameter, removing network-specific logic from the function.

Follow-up: #13959 (replace `EitherEvm` with compile-time dispatch)

PR #13959, opened 2026-03-27
refactor(anvil): remove `new_eth_evm_with_inspector ` + some instances of `EitherEvm`

figtracer:
## Summary

Stacked on #13940.

Replace runtime `EitherEvm` dispatch with compile-time branching on `is_optimism()`. Creates concrete `EthEvm`/`OpEvm` instances directly via their respective `EvmFactory`, eliminating the `EitherEvm` wrapper.

- Add `transact_with_inspector_ref` for single-tx call sites
- Add `execute_with_block_executor` for mining/pending block paths  
- Add `build_mining_inspector` helper
- Remove `new_eth_evm_with_inspector` and `EitherEvm` usage

Prep for threading `EvmFactory` through `Backend`.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -1,6 +1,6 @@
 //! In-memory blockchain backend.
 use self::state::trie_storage;
-use super::executor::new_eth_evm_with_inspector;
+
 use crate::{
     ForkChoice, NodeConfig, PrecompileFactory,
     config::PruneStateHistoryConfig,
@@ -9,7 +9,7 @@ use crate::{
             cheats::{CheatEcrecover, CheatsManager},
             db::{AnvilCacheDB, Db, MaybeFullDatabase, SerializableState, StateDb},
             executor::{
-                AnvilBlockExecutor, PoolTxGasConfig, build_tx_env_for_pending,
+                AnvilBlockExecutor, ExecutedPoolTransactions, PoolTxGasConfig,
                 execute_pool_transactions,
             },
             fork::ClientFork,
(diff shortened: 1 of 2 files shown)


crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -1045,23 +1049,108 @@ impl<N: Network> Backend<N> {
...
+                result: result.result.map_haltreason(OpHaltReason::Base),
+                state: result.state,
+            })
+        }
+    }
+
+    /// Creates a concrete EVM + [`AnvilBlockExecutor`], runs pre-execution changes, and
+    /// executes pool transactions. Returns the execution results and drops the EVM.
+    #[allow(clippy::too_many_arguments, clippy::type_complexity)]
+    fn execute_with_block_executor<DB>(
```
mablr, line 1097: It's better than it used to be 👍

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -1045,23 +1049,108 @@ impl<N: Network> Backend<N> {
...
         DB: DatabaseRef + ?Sized,
         I: Inspector<EthEvmContext<WrapDatabaseRef<&'db DB>>>
             + Inspector<OpContext<WrapDatabaseRef<&'db DB>>>,
         WrapDatabaseRef<&'db DB>: Database<Error = DatabaseError>,
     {
-        let mut evm =
-            new_eth_evm_with_inspector(WrapDatabaseRef(db), evm_env, inspector, self.is_optimism());
-        self.inject_precompiles(evm.precompiles_mut());
-        evm
+        if self.is_optimism() {
```
grandizzy, line 1067: wonder if we could avoid checking if optimisim here, e.g. below we're doing `self.inject_precompiles` https://github.com/figtracer/foundry/blob/8b401e320345026feaa1a0d313c86b0f51127817/crates/anvil/src/eth/backend/mem/mod.rs#L1034 which checks and injects proper precompiles through network configs crate single place https://github.com/figtracer/foundry/blob/8b401e320345026feaa1a0d313c86b0f51127817/crates/anvil/src/eth/backend/mem/mod.rs#L1034
IMO ideally we shouldn't have any network check directly in anvil code

figtracer: +1, but i believe this can't get dropped until `Backend<EvmFactory>` or something similar exists

`transact_with_inspector_ref` would just know the evm by heart and we wouldn't need this i think

grandizzy: sgtm if that's the direction, thanks!

PR #13962, opened 2026-03-27
fix(create): check receipt status before reporting deployment success

### LINKED ISSUE #13954: forge create reports success on failed deployment receipts on Tempo Moderato
### Component
forge create / forge script broadcast

### What happened?
On Tempo Moderato testnet (`https://rpc.moderato.tempo.xyz`, chain id `42431`), `forge create` can exit `0` and print `Deployed to: ...` even when the on-chain receipt has `status = 0` and the deployed address has no code.

I also see `forge script` hit `OpcodeNotFound` when the script itself calls Tempo TIP-20 native precompiles, even though the same calls work:
- via direct RPC `cast call`
- via a real deployed Solidity contract calling the same token

The most dangerous part is the false-positive deploy output from `forge create`.

### Versions
- forge `1.6.0-dev`
- cast `1.6.0-dev`
- commit `5b6f3c8e64d46b9b9c421106e5281f53896733b7`

### Repro 1: false-positive create success
Command:
```bash
forge create src/PredictionMarket.sol:PredictionMarket \
  --broadcast \
  --gas-limit 20000000 \
  --rpc-url https://rpc.moderato.tempo.xyz \
  --private-key <funded-key>
```

Forge output:
```text
Deployer: 0x312e253d1C18b92112aD2276Ade650FE7E620D1a
Deployed to: 0x6b821802e2cB86d3082EA47342b6d49D3618D62d
Transaction hash: 0x5f38ff9fc3c2d29436d098c74aa5b040159da19a5c31b99bfd801dcb838ed314
```

Actual receipt:
```text
status               0 (failed)
contractAddress      0x6b821802e2cB86d3082EA47342b6d49D3618D62d
gasUsed              20000000
transactionHash      0x5f38ff9fc3c2d29436d098c74aa5b040159da19a5c31b99bfd801dcb838ed314
feeToken             0x20C0000000000000000000000000000000000000
```

And `cast code 0x6b821802e2cB86d3082EA47342b6d49D3618D62d --rpc-url https://rpc.moderato.tempo.xyz` returns `0x`.

Expected behavior:
- `forge create` should fail non-zero if the receipt status is `0`
- it should not print `Deployed to:` as if deployment succeeded when no code exists

### Repro 2: script VM vs live chain mismatch on Tempo TIP-20 precompile
Tempo native stable precompiles are at `0x20c0...`.

Direct RPC works:
```bash
cast call 0x20c0000000000000000000000000000000000000 'balanceOf(address)(uint256)' 0x312e253d1C18b92112aD2276Ade650FE7E620D1a --rpc-url https://rpc.moderato.tempo.xyz
```

A real deployed probe contract calling `balanceOf` and `approve` on the same token also works on-chain.

But a `forge script` path that calls `IERC20(pathUSD).balanceOf(...)` inside the Solidity script hits:
```text
OpcodeNotFound
```

That suggests a simulation/runtime incompatibility in the script path against this chain's native precompiles.

### Notes
Local gas report for this contract says deployment cost is about `5,216,990`, but the live create above consumed the full `20,000,000` gas cap and still failed. I am not sure whether that part belongs in Foundry or in Tempo, but the false-positive success reporting definitely seems like a Foundry bug.

decofe:
## Motivation

`forge create` prints `Deployed to:` and exits 0 even when the on-chain receipt has `status = 0` (failed). When a chain returns a `contractAddress` in the receipt for reverted creates, the existing `contract_address().is_some()` check passes — the user sees a false-positive success but `cast code` returns `0x` (no code deployed).

Closes #13954.

By contrast, `forge script` already checks receipt status via `TxStatus` in `crates/script/src/receipts.rs`.

## Solution

- Check `receipt.status()` in `send_with_receipt()` **before** extracting `contract_address()`
- Add `DeploymentFailed(TxHash)` error variant so the tx hash is surfaced in the error
- Add regression test: deploy a contract with a reverting constructor + `--gas-limit` (to bypass `eth_estimateGas` rejection), assert stderr snapshot matches expected error

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

Prompted by: zerosnacks

### DIFF
--- a/crates/forge/tests/cli/create.rs
+++ b/crates/forge/tests/cli/create.rs
@@ -499,7 +499,7 @@ Error: no bytecode found in bin object for AbstractCounter
 
 // Tests that `forge create` fails when the deployment transaction reverts
 // <https://github.com/foundry-rs/foundry/issues/13954>
-forgetest_async!(should_fail_on_reverted_deployment, |prj, cmd| {
+forgetest_async!(flaky_should_fail_on_reverted_deployment, |prj, cmd| {
     let (_api, handle) = spawn(NodeConfig::test()).await;
     let rpc = handle.http_endpoint();
     let wallet = handle.dev_wallets().next().unwrap();
@@ -528,11 +528,10 @@ contract RevertingContract {
         "--broadcast",
         "--gas-limit",
         "1000000",
-    ]);
+    ])
+    .assert_failure()
+    .stderr_eq(str![[r#"
+Error: deployment transaction failed (receipt status 0): [..]
 
-    let output = cmd.assert_failure().get_output().stdout_lossy();
-    assert!(
-        !output.contains("Deployed to:"),
-        "should not print 'Deployed to:' for a reverted deployment, stdout: {output}"
-    );
+"#]]);
 });


grandizzy (approved):
makes sense!

PR #13967, opened 2026-03-27
feat(evm): generic `EvmOpts`

mablr:
## Motivation

Since `CreateFork` does not contain `EvmEnv` anymore (#13965), we can make `EvmOpts` fully generic over evm types.

### DIFF
--- a/crates/evm/core/src/opts.rs
+++ b/crates/evm/core/src/opts.rs
@@ -129,7 +129,7 @@ impl EvmOpts {
             .build()
     }
 
-    /// Returns a tuple with [`EvmEnv`] and [`TxEnv`]
+    /// Returns a tuple with [`EvmEnv`] and `TxEnv`
     ///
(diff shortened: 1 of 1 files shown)


crates/evm/core/src/opts.rs
```diff
@@ -146,10 +152,15 @@ impl EvmOpts {
 
     /// Returns the [`EvmEnv`] (cfg + block) and [`BlockNumber`] fetched from the fork endpoint via
     /// provider
-    pub async fn fork_evm_env<N: Network, P: Provider<N>>(
+    pub async fn fork_evm_env<
+        SPEC: Into<SpecId> + Default + Copy,
+        BLOCK: FoundryBlock + Default,
+        N: Network,
+        P: Provider<N>,
+    >(
```
mattsse, line 160: maybe there's a way to simplify these bounds, with some extension trait approach somehow, but we can re-evaluate later.

like we could maybe try move all of this logic into the `FoundryNetwork` type itself or similar

PR #13975, opened 2026-03-28
feat(common): add Tempo wallet keystore types

figtracer:
## Summary

Add types for reading keys from the Tempo CLI wallet keystore (`~/.tempo/wallet/keys.toml`).

- `KeyEntry`, `KeysFile`, `WalletType`, `KeyType`, `StoredTokenLimit` structs
- `tempo_home()`, `tempo_keys_path()`, `read_tempo_keys_file()` helpers
- `decode_key_authorization()` generic RLP decode helper

Structured as a `tempo/` module directory for future additions (provider, wallet integration, etc.).

Ported from tempoxyz/tempo-foundry#357 (partial — keystore types only).


crates/common/src/tempo/keystore.rs
mattsse, line 1: we likely need to add more things, can we make this a module instead?

PR #13978, opened 2026-03-28
fix(anvil): clean up `InMemoryBlockStates` on rollback

FredPhilipy:
Added `remove_block_states` method to `InMemoryBlockStates` and call it in `rollback()` after `unwind_to()`. This prevents leaked block states in memory, on-disk tracking, and disk cache files when blocks are rolled back during chain reorganization.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/storage.rs
+++ b/crates/anvil/src/eth/backend/mem/storage.rs
@@ -728,6 +728,62 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_remove_block_states_on_rollback() {
+        let mut storage = InMemoryBlockStates::new(10, MAX_ON_DISK_HISTORY_LIMIT);
+
+        // Insert 5 states
+        let hashes: Vec<B256> = (0..5)
+            .map(|i| {
+                let hash = B256::from(U256::from(i));
+                let mut state = MemDb::default();
+                let addr = Address::from_word(hash);
+                state.insert_account(addr, AccountInfo::from_balance(U256::from(i * 100)));
+                storage.insert(hash, StateDb::new(state));
+                hash
+            })
+            .collect();
+
+        assert_eq!(storage.present.len(), 5);
+
+        // Simulate rollback: remove the last 3 blocks
+        let removed_hashes = &hashes[2..];
+        storage.remove_block_states(removed_hashes);
+
+        // Only the first 2 states should remain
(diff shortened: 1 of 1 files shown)


zerosnacks:
Hi @FredPhilipy, thanks for your PR!

Would you mind adding two test cases: one verifying only particular hashes are removed and another that removal is actually occuring?

Something like:

```rust
async fn test_remove_block_states_on_rollback() {
    let mut storage = InMemoryBlockStates::new(10, MAX_ON_DISK_HISTORY_LIMIT);

    // Insert 5 states
    let hashes: Vec<B256> = (0..5)
        .map(|i| {
            let hash = B256::from(U256::from(i));
            let mut state = MemDb::default();
            let addr = Address::from_word(hash);
            state.insert_account(addr, AccountInfo::from_balance(U256::from(i * 100)));
            storage.insert(hash, StateDb::new(state));
            hash
        })
        .collect();

    assert_eq!(storage.present.len(), 5);

    // Simulate rollback: remove the last 3 blocks
    let removed_hashes = &hashes[2..];
    storage.remove_block_states(removed_hashes);

    // Only the first 2 states should remain
    assert_eq!(storage.present.len(), 2);
    assert!(storage.get_state(&hashes[0]).is_some());
    assert!(storage.get_state(&hashes[1]).is_some());
    for h in removed_hashes {
        assert!(storage.get_state(h).is_none());
        assert!(!storage.present.contains(h));
    }
}

async fn test_remove_block_states_cleans_disk_cache() {
    // Use limit=1 to force states to disk
    let mut storage = InMemoryBlockStates::new(1, MAX_ON_DISK_HISTORY_LIMIT);

    let hash_a = B256::from(U256::from(1));
    let hash_b = B256::from(U256::from(2));

    storage.insert(hash_a, StateDb::new(MemDb::default()));
    storage.insert(hash_b, StateDb::new(MemDb::default()));

    // Wait for disk flush
    tokio::time::sleep(std::time::Duration::from_secs(1)).await;

    assert!(storage.on_disk_states.contains_key(&hash_a));

    // Remove hash_a (on disk)
    storage.remove_block_states(&[hash_a]);

    assert!(!storage.on_disk_states.contains_key(&hash_a));
    assert!(!storage.oldest_on_disk.contains(&hash_a));
    assert!(storage.get_on_disk_state(&hash_a).is_none());
}
```

FredPhilipy:
Added the test @zerosnacks

PR #13981, opened 2026-03-28
fix(traces): decode expectRevert payload before revert formatting

ArshLabs:
## Motivation

`expectRevert` trace formatting currently feeds full cheatcode calldata into `RevertDecoder`, instead of decoding the cheatcode input first and passing only the expected-revert payload. This can produce misleading trace output.

## Solution

Decode `expectRevert` inputs first in `CallTraceDecoder::decode_cheatcode_inputs`, then pass the decoded expected-revert bytes (`bytes` or `bytes4`) into `RevertDecoder`.

Added local decoder test coverage in the same module for:
- `expectRevert(bytes4)`
- `expectRevert(bytes)`
- `expectRevert()` fallback behavior

Validation performed:
- `cargo +nightly fmt -- --check`
- `cargo test -p foundry-evm-traces`
- `cargo check --all`
- `cargo +nightly clippy -p foundry-evm-traces --all-targets -- -D warnings`
- attempted `cargo test --all --all-features` (currently fails in this environment at `tracy-client-sys`, unrelated to touched code)

AI disclosure: I used AI assistance for basic code navigation, then did drafting & manually verified the implementation and test behavior.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/traces/src/decoder/mod.rs
+++ b/crates/evm/traces/src/decoder/mod.rs
@@ -513,8 +513,17 @@ impl CallTraceDecoder {
     fn decode_cheatcode_inputs(&self, func: &Function, data: &[u8]) -> Option<Vec<String>> {
         match func.name.as_str() {
             "expectRevert" => {
-                let decoded = func.abi_decode_input(&data[SELECTOR_LEN..]).ok()?;
-                let expected_revert = match decoded.first()? {
+                let decoded = match data.get(SELECTOR_LEN..) {
+                    Some(data) => func.abi_decode_input(data).ok(),
+                    None => None,
+                };
+                let Some(decoded) = decoded else {
+                    return Some(vec![self.revert_decoder.decode(data, None)]);
+                };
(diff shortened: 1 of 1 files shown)


ArshLabs:
small context on the bug this fixes: `expectRevert` trace formatting was decoding full cheatcode calldata, which can show misleading revert text. this patch decodes `expectRevert` input first and passes only the expected-revert payload (`bytes`/`bytes4`) into `RevertDecoder`. added local decoder coverage for `expectRevert(bytes4)`, `expectRevert(bytes)`, and `expectRevert()`.

any review suggestions or requested changes are very welcome.

@DaniPopes @mattsse @grandizzy @zerosnacks @onbjerg @0xrusowsky

mattsse (approved):
LGTM. Decoding the cheatcode inputs first and forwarding only the user-supplied revert payload into RevertDecoder is the right fix. Verified locally: tests, clippy, and nightly fmt all pass.

PR #13986, opened 2026-03-29
lint: add post-increment-decrement gas lint

ParthSinghPS:
## What this PR does
Adds a new `post-increment-decrement` lint rule in the `gas` category that flags
`i++` and `i--` when used in for loop update expressions, suggesting `++i` and
`--i` instead.

## Why
Post-increment (`i++`) and post-decrement (`i--`) return the value before
modification, requiring an extra copy operation under the hood. In for loop
update expressions the return value is always discarded, so the copy is
pure overhead. Pre-increment (`++i`) skips the copy entirely.

The lint is intentionally scoped to for loop update expressions only. Outside
of loops, `i++` vs `++i` is a semantic choice the developer may have made
deliberately.

## Changes
- `crates/lint/src/sol/gas/increment.rs`: new `EarlyLintPass` implementation
- `crates/lint/src/sol/gas/mod.rs`: registered the new lint
- `crates/lint/testdata/PostIncrementDecrement.sol`: test cases covering
  triggering and non-triggering patterns
- `crates/lint/testdata/PostIncrementDecrement.stderr`: blessed test output

### DIFF
--- a/crates/lint/src/sol/gas/increment.rs
+++ b/crates/lint/src/sol/gas/increment.rs
@@ -1,0 +1,23 @@
+use super::PostIncrementDecrement;
+use crate::{
+    linter::{EarlyLintPass, LintContext},
+    sol::{Severity, SolLint},
+};
+use solar::ast::{Expr, ExprKind, UnOpKind};
+
+declare_forge_lint!(
+    POST_INCREMENT_DECREMENT,
+    Severity::Gas,
+    "post-increment-decrement",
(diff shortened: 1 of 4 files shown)


zerosnacks:
Hi @ParthSinghPS thanks for your suggestion, I do not think this gas rule is that impactful and harms readability / code style pushing users in an opinionated way like this. To my knowledge the compiler is now smart enough to optimize this.

Issue #9619, opened 2025-01-03
Error intrinsic gas too low when deploying contract on Plume Devnet


live9369:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [ ] Foundryup

### What version of Foundry are you on?

0.2.0

### What command(s) is the bug in?

forge script script/Counter.s.sol:CounterScript --rpc-url $RPC --private-key $PRIVATE_KEY --broadcast

### Operating System

Linux

### Describe the bug

```
[⠊] Compiling...
No files changed, compilation skipped
Script ran successfully.

## Setting up 1 EVM.

==========================

Chain 98864

Estimated gas price: 0.020000001 gwei

Estimated total gas used for script: 548895

Estimated amount required: 0.000010977900548895 ETH

==========================

Transactions saved to: /home/live/chainopera/t_foundry/broadcast/Counter.s.sol/98864/run-latest.json

Sensitive values saved to: /home/live/chainopera/t_foundry/cache/Counter.s.sol/98864/run-latest.json

Error: Failed to send transaction

Context:
- server returned an error response: error code -32000: intrinsic gas too low
```

live9369:
@zerosnacks Hi, what needs to be done now？

zerosnacks:
> @zerosnacks Hi, what needs to be done now？

Is there a bridge available for bridging from Sepolia to Plume Devnet?

The Plume team will need to check https://github.com/alloy-rs/chains/compare/main...zerosnacks/add-plume for correctness

live9369:
> > @zerosnacks Hi, what needs to be done now？
> 
> Is there a bridge available for bridging from Sepolia to Plume Devnet?
> 
> The Plume team will need to check [alloy-rs/chains@main...zerosnacks/add-plume](https://github.com/alloy-rs/chains/compare/main...zerosnacks/add-plume) for correctness and / or submit a PR
> 
> Once it is added as a chain we can add it in the check above which will query the RPC for the gas calculation

Hi，sorry， I am not entirely sure about the available bridge from Sepolia to Plume Devnet at the moment.  I just started working with this chain today.

chingfeng2021:
there is a bridge: https://plume-testnet-m8kfz7osif-fbc24081e9bf0dcf.testnets.rollbridge.app/

sakulstra:
Any update on this? Having the exact same behavior when trying to use megaeth with foundry.

zerosnacks:
[...]

PR #13999, opened 2026-03-30
fix: add `MegaEth` and `MegaEthTestnet` to different gas calc due to `Error intrinsic gas too low`

zerosnacks:
Fixes: https://github.com/foundry-rs/foundry/issues/9619#issuecomment-4154062690


mattsse (approved):

zerosnacks:
Confirming to be working here: https://github.com/foundry-rs/foundry/issues/9619#issuecomment-4154297305

PR #14007, opened 2026-03-30
fix(evm): query `state_snapshot.storage` in `ForkDbStateSnapshot::storage_ref`

gutonosa:
`ForkDbStateSnapshot::storage_ref` was skipping `state_snapshot.storage` entirely. The lookup path went from `local.cache` straight to the remote RPC backend, missing any storage data that `SharedBackend` had previously fetched and persisted in the `BlockchainDb`. This caused stale or incorrect values to be returned when reverting to a state snapshot, since the RPC would return current on-chain state rather than the snapshotted state.

aligns `storage_ref` with `basic_ref`, which already queries `state_snapshot` as the second lookup layer. Replaced the redundant `get_storage()` helper (which only re-checked `local.cache`) with `storage_from_snapshot_or_backend()`, giving the correct lookup order: `local.cache` → `state_snapshot.storage` → `SharedBackend` / RPC.

### DIFF
--- a/crates/evm/core/src/fork/database.rs
+++ b/crates/evm/core/src/fork/database.rs
@@ -302,4 +302,28 @@ mod tests {
         assert!(loaded.is_some());
         assert_eq!(loaded.unwrap(), info);
     }
+
+    /// Verifies that `ForkDbStateSnapshot::storage_ref` reads from `state_snapshot.storage`
+    /// when the slot is missing from `local.cache.accounts`. Without this lookup the call
+    /// would fall through to the backend and return the unrelated remote value.
+    #[tokio::test(flavor = "multi_thread")]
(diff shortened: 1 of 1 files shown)


figtracer (commented):
let's add a test to verify the behavior

gutonosa:
> let's add a test to verify the behavior

sorry for holding a long time, rebase adn add the test

mablr (dismissed):
lgtm, thank you.

Pending another review + conflict resolution

figtracer (dismissed):
thanks! pls solve conflict

gutonosa:
solved @mablr @figtracer

mattsse (approved):

figtracer (approved):

PR #14012, opened 2026-03-30
feat(config): add `hardfork` field to `Config`

figtracer:
## Summary

Add `hardfork: Option<FoundryHardfork>` to `Config` for specifying the EVM hardfork in `foundry.toml`. Defaults to `None`.

Re-exports `FoundryHardfork` from `foundry-config` for convenience.

Ported from tempoxyz/tempo-foundry#152.


figtracer:
we actually won't need this

PR #14016, opened 2026-03-30
fix(evm): align local create2 setup with configured deployer

ArshLabs:
## Motivation

When `create2_deployer` is customized, local setup currently checks or deploys only the default deployer address, while CREATE2 execution uses the inspector-configured address. This can make setup and runtime target different addresses and lead to `missing CREATE2 deployer` reverts even though setup already ran.

## Solution

`deploy_create2_deployer` now resolves the deployer address from the executor inspector, so setup and runtime CREATE2 execution use the same address.

Behavior by case:
- Default deployer address: keep existing deployment path via the signed deployer transaction flow.
- Custom deployer address: if empty, install the expected deployer runtime bytecode directly at the configured address for local execution.

A regression test is added for `forge script` no-endpoint execution with `--create2-deployer` to verify custom deployer setup aligns with runtime behavior.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


zerosnacks:
Hi @ArshLabs, thanks for your fix!

This is a sensitive code path so I would like to see additional tests for it for `forge test` as the current PR only covers `forge script`. I would also like to see a test covering a forking scenario.

ArshLabs:
> Hi @ArshLabs, thanks for your fix!
> 
> This is a sensitive code path so I would like to see additional tests for it for `forge test` as the current PR only covers `forge script`. I would also like to see a test covering a forking scenario.

Added the requested forge test coverage for custom CREATE2 deployer in both local and forked paths, and pushed the update. Both new tests pass locally.

ArshLabs:
cc @DaniPopes @mattsse @grandizzy @zerosnacks @onbjerg @0xrusowsky

I merged latest master into this branch, resolved the conflict, and pushed the updated commits.

I also re-ran the custom CREATE2 deployer coverage:

-forge test (local)
-forge test (fork)
-forge script path (no endpoint)

All of these passed locally.
could one of you please take a look and approve when you have a chance?

PR #14022, opened 2026-03-31
feat(evm): add tmp `TempoFoundryEvm` wrapper

figtracer:
## Summary

Adds `TempoFoundryEvm`, `TempoFoundryHandler`, and `new_tempo_evm_with_inspector()` as Tempo counterparts to the existing Eth types. This lets Tempo use the CREATE2 factory redirect handler.

Temporary. All of this gets deleted when the next revm release includes [bluealloy/revm#3518](https://github.com/bluealloy/revm/pull/3518) and CREATE2 moves to a simple inspector.

### DIFF
--- a/crates/evm/core/src/evm.rs
+++ b/crates/evm/core/src/evm.rs
@@ -31,6 +31,12 @@ use revm::{
     },
     primitives::hardfork::SpecId,
 };
+use tempo_chainspec::hardfork::TempoHardfork;
+use tempo_evm::evm::TempoEvmFactory;
+use tempo_revm::{
+    TempoBlockEnv, TempoHaltReason, TempoInvalidTransaction, TempoTxEnv, evm::TempoContext,
+    gas_params::tempo_gas_params, handler::TempoEvmHandler,
+};
 
 /// Marker trait for [`EvmFactory`] implementations compatible with Foundry's EVM infrastructure.
 ///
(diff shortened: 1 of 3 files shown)


crates/evm/core/src/evm.rs
```diff
@@ -447,3 +452,415 @@ impl<'db, I: FoundryInspectorExt<EthEvmContext<&'db mut dyn DatabaseExt>>> Inspe
...
 }
+
+// Will be removed when the next revm release includes bluealloy/revm#3518.
+type TempoRevmEvm<'db, I> = TempoEvm<&'db mut dyn DatabaseExt, I>;
+
+/// Tempo counterpart of [`FoundryEvm`]. Wraps [`TempoEvm`] and routes execution through
+/// [`TempoFoundryHandler`] which provides CREATE2 factory redirect logic.
+pub struct TempoFoundryEvm<'db, I: FoundryInspectorExt<TempoContext<&'db mut dyn DatabaseExt>>> {
+    pub inner: TempoRevmEvm<'db, I>,
+}
```
mablr: It could be better to wrap `tempo_evm::TempoEvm` instead of raw `tempo_revm::TempoEvm` ? So that we can reuse `EvmFactory` impls (same for Eth).

The only thing is that we will need to backport https://github.com/tempoxyz/tempo/pull/3296 on tempo's alloy-2.0 branch.

wdyt?

mablr (commented):
Good start for this temporary workaround👍

crates/evm/core/src/evm.rs
figtracer: makes sense

mablr: fyi https://github.com/tempoxyz/tempo/pull/3393

PR #14023, opened 2026-03-31
common: add unit tests for SlotIdentifier

ParthSinghPS:
## What this PR does
Adds unit tests for `SlotIdentifier` in `crates/common/src/slot_identifier.rs`.

## Why
`SlotIdentifier` is a 990-line file with complex recursive logic for identifying
Solidity storage slots , handling primitive types, fixed arrays, structs, mappings,
and bytes/strings. It had zero test coverage despite being used in two critical
paths: fuzzing state generation (`crates/evm/fuzz`) and cheatcode storage decoding
(`crates/cheatcodes`).

## What's tested
- Simple primitive types (`uint256`, `address`, `bool`) are identified correctly
- Unknown slots return `None`
- Fixed array elements resolve with correct index labels
- `is_struct()` helper returns correct results for struct and non-struct labels

## Changes
- `crates/common/src/slot_identifier.rs`: added `#[cfg(test)]` module with 5 unit
  tests using inline `StorageLayout` fixtures

### DIFF
--- a/crates/common/src/slot_identifier.rs
+++ b/crates/common/src/slot_identifier.rs
@@ -988,3 +988,132 @@ fn get_array_base_indices(dyn_type: &DynSolType) -> String {
 pub fn is_struct(s: &str) -> bool {
     s.starts_with("struct ")
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn make_storage(label: &str, slot: &str, type_ref: &str) -> Storage {
+        Storage {
+            ast_id: 0,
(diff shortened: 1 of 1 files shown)


zerosnacks:
Hi @ParthSinghPS, I think we already cover this sufficiently through our existing test suite and the newly added tests only cover a surface level (not things like dynamic slot resolution for mappings, dynamic arrays, nested structs, packed slots, etc..). I'll therefore close the PR, thanks for your suggestion.

PR #14049, opened 2026-04-01
refactor: use native Solidity types for vm.rpc struct decoding

decofe:
Follow-up to #13842.

Removes the `Tuple` recursion from `convert_to_bytes` so nested `Address` and `FixedBytes` values keep their native ABI types instead of being flattened to dynamic `bytes`. This lets users decode `vm.rpc` struct results with native Solidity types (`address`, `bytes32`) instead of `bytes` + manual casting.

## Changes
- Remove `DynSolValue::Tuple` arm from `convert_to_bytes` in `fork.rs`
- Update `BlockResult` and `LegacyTransactionResult` test structs to use `address`/`bytes32` where appropriate
- Remove manual `bytes32(...)` / `address(bytes20(...))` casts from assertions

Prompted by: zerosnacks


decofe:
Superseded by new PR based on master (after #13842 was merged).

PR #14057, opened 2026-04-01
feat(anvil): add Tempo async pool validation (time bounds + fee token balance)

stevencartavia:
Add async validation checks for Tempo transactions at pool admission:

- Time bounds: reject if `valid_before` is expired/too close (< 3s buffer) or valid_after is too far in the future (> 1h)
- Fee token balance: ERC-20 `balanceOf` check on the fee payer's token balance, scaled from 18-decimal wei to 6-decimal token units
Also adds `get_fee_token_balance` helper for ERC-20 balance queries.

Follow-up: move `call()` and related methods out of the `Self: TransactionValidator` bounded impl block so `TransactionValidator` can be made generic over `N: Network`

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -58,7 +58,7 @@ use alloy_network::{
 };
 use alloy_op_evm::OpEvmFactory;
 use alloy_primitives::{
-    Address, B256, Bloom, Bytes, TxHash, TxKind, U64, U256, address, hex, keccak256, logs_bloom,
+    Address, B256, Bloom, Bytes, TxHash, TxKind, U64, U256, hex, keccak256, logs_bloom,
     map::{AddressMap, HashMap, HashSet},
 };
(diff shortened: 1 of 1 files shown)


crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -4049,7 +4087,60 @@ impl TransactionValidator<FoundryTxEnvelope> for Backend<FoundryNetwork> {
...
+                        max_allowed,
+                    }
+                    .into());
+                }
+            }
+
+            // Fee token balance check
+            let fee_payer = tempo_tx.recover_fee_payer(address).unwrap_or(address);
+            let fee_token =
+                tempo_tx.fee_token.unwrap_or(address!("20C0000000000000000000000000000000000000"));
```
zerosnacks: let's use https://github.com/tempoxyz/tempo/blob/57f8cf1ecc05b92eb54f1478f1d26494b40e6062/crates/contracts/src/precompiles/mod.rs#L27 instead

PR #14060, opened 2026-04-01
feat(anvil): use hash-based pool ordering for Tempo transactions

stevencartavia:
Use hash-based provides/requires markers instead, allowing Tempo txs to be mined independently.

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -1947,14 +1947,16 @@ impl EthApi<FoundryNetwork> {
         let priority = self.transaction_priority(&pending_transaction.transaction);
 
         // Tempo txs use a 2D nonce system — no sequential ordering by account nonce.
-        let (requires, provides) =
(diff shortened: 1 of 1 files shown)


crates/anvil/src/eth/api.rs
```diff
@@ -1943,18 +1943,21 @@ impl EthApi<FoundryNetwork> {
...
-        let pool_transaction = PoolTransaction {
-            requires,
-            provides: vec![to_marker(nonce, *pending_transaction.sender())],
-            pending_transaction,
-            priority,
-        };
+
+        // Tempo txs use a 2D nonce system — no sequential ordering by account nonce.
+        let (requires, provides) =
+            if matches!(pending_transaction.transaction.as_ref(), FoundryTxEnvelope::Tempo(_)) {
```
mattsse: I think we also need to check that the nonce key isnt zero because the 0 nonce key behaves just like other txs

PR #14062, opened 2026-04-01
refactor(evm): use `FoundryEvmFactory` in `Backend`/`CowBackend`

mablr:
## Motivation

Fully generic `Backend`/`CowBackend` over `Network` and `EvmFactory`.

Introduced `FoundryEvmFactory::transact_with_dyn_inspector` workaround to handle dyn inspector dispatch through `CowBackend`->`Backend`.

OP support is not guaranteed yet since `DatabaseExt::{transact,transact_from_tx}` need to use concrete `Context`. To provide we'll need to add `CHAIN` generic to `DatabaseExt` but it's a huge diff, let's do it later (after tempo upstream).

### DIFF
--- a/crates/evm/core/src/backend/cow.rs
+++ b/crates/evm/core/src/backend/cow.rs
@@ -2,25 +2,27 @@
 
 use super::BackendError;
 use crate::{
-    FoundryInspectorExt,
+    FoundryInspectorExt, TryAnyToTxEnv,
     backend::{
         Backend, DatabaseExt, JournaledState, LocalForkId, RevertStateSnapshotAction,
         diagnostic::RevertDiagnostic,
     },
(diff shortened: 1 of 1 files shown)


crates/evm/core/src/backend/cow.rs
```diff
@@ -42,19 +44,22 @@ use std::{borrow::Cow, collections::BTreeMap};
 /// which would add significant overhead for large fuzz sets even if the Database is not big after
 /// setup.
 #[derive(Clone, Debug)]
-pub struct CowBackend<'a> {
+pub struct CowBackend<'a, N: Network, F: FoundryEvmFactory> {
```
figtracer, line 47: very nice

figtracer (approved):

PR #14071, opened 2026-04-02
chore(evm): import `PATH_USD_ADDRESS` from `tempo-contracts`

figtracer:
## Summary

Import `PATH_USD_ADDRESS` from `tempo_contracts::precompiles` instead of redefining it.


zerosnacks:
We can do the same for the other addresses, if they don't exist as constants let's add them upstream

figtracer:
@zerosnacks should we bump lock in this PR?

zerosnacks:
@figtracer yeah that would make sense, it is required for the new imports

PR #14074, opened 2026-04-02
refactor(evm) `RawCallResult` type parameters

figtracer:
## Summary

Adds `RawCallResult<SPEC, BLOCK, TX, N: Network>` type parameters. Prep work for `Executor<F>`

### DIFF
--- a/crates/evm/evm/src/executors/mod.rs
+++ b/crates/evm/evm/src/executors/mod.rs
@@ -11,7 +11,7 @@ use crate::inspectors::{
 };
 use alloy_dyn_abi::{DynSolValue, FunctionExt, JsonAbiExt};
 use alloy_json_abi::Function;
-use alloy_network::Ethereum;
+use alloy_network::{Ethereum, Network};
 use alloy_primitives::{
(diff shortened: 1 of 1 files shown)


crates/evm/evm/src/executors/mod.rs
```diff
@@ -907,23 +907,27 @@ pub struct RawCallResult {
...
     pub chisel_state: Option<(Vec<U256>, Vec<u8>)>,
     pub reverter: Option<Address>,
 }
 
-impl Default for RawCallResult {
+impl<SPEC, BLOCK, TX, N: Network> Default for RawCallResult<SPEC, BLOCK, TX, N>
+where
+    EvmEnv<SPEC, BLOCK>: Default,
+    TX: Default,
+{
```
mablr:
```suggestion
impl<SPEC: Default, BLOCK: Default, TX: Default, N: Network> Default for RawCallResult<SPEC, BLOCK, TX, N> {
```
This works also I think.

figtracer: TIL

PR #14092, opened 2026-04-02
clippy: warn on `if_not_else` (autofix)

decofe:
Enable `clippy::if_not_else` and fix all findings.

All fixes are auto-applied via `cargo clippy --fix`, except for three cases that clippy cannot auto-fix:
- `crates/config/src/filter.rs` — flipped negation in `is_match`
- `crates/config/src/lib.rs` — flipped `!map.contains_key` to `map.contains_key` (required adding a type annotation since inference order changed)
- `crates/cast/src/cmd/keychain.rs` — flipped `!=` comparisons

Prompted by: zerosnacks


crates/config/src/lib.rs
```diff
@@ -1188,6 +1186,8 @@ impl Config {
                         );
                         map.insert(source.clone(), value);
                     }
+                } else {
+                    map.insert(source.clone(), res);
                 }
```
zerosnacks, line 1191: note: to double check

decofe: The original code was:

```rust
if !map.contains_key(source) {
    map.insert(source.clone(), res);
} else {
    let value = map.remove(source.as_path()).unwrap();
    // merge logic...
}
```

The `if_not_else` lint flagged the negation. Flipping the branches gives us `if map.contains_key(source) { remove + merge } else { insert }`, but that's a redundant `contains_key` followed by `remove`. Instead, this uses `if let Some(value) = map.remove(...)` which combines the check and removal into one operation — same semantics, one fewer map lookup, no `.unwrap()`.

The explicit type annotation on `map` (line 1154-1155) was needed because the compiler previously inferred the value type from the first `map.insert(source, res)` call which now appears in the `else` branch rather than before the `remove`.

mablr (dismissed):

PR #14112, opened 2026-04-03
feat(anvil): add Tempo genesis and basic behavior tests

stevencartavia:
Add initial Tempo test suite for `anvil --tempo` mode.

Ported:
- `test_tempo_precompiles_have_code` — precompiles and tokens have code at genesis
- `test_tip20_token_metadata` — name, symbol, decimals for all 4 fee tokens
- `test_fee_token_balances_minted_to_dev_accounts` — dev accounts have token balances
- `test_dev_accounts_have_balance` — dev accounts have genesis balance
- `test_native_value_transfer_rejected` — value > 0 rejected with error
- `test_nonce_increments` — nonce updates after TIP20 transfer

New:
- `test_zero_value_tx_succeeds` — zero-value EIP-1559 tx mines successfully
- `test_contract_deployment` — contract creation works with value=0


crates/anvil/tests/it/tempo.rs
```diff
@@ -0,0 +1,259 @@
...
+    let precompiles: &[Address] = &[
+        NONCE_PRECOMPILE_ADDRESS,
+        STABLECOIN_DEX_ADDRESS,
+        TIP20_FACTORY_ADDRESS,
+        TIP403_REGISTRY_ADDRESS,
+        TIP_FEE_MANAGER_ADDRESS,
+        VALIDATOR_CONFIG_ADDRESS,
+        VALIDATOR_CONFIG_V2_ADDRESS,
+        ACCOUNT_KEYCHAIN_ADDRESS,
+    ];
```
zerosnacks: Preferred to maintain this as a single list outside of the test suite so implementation and test suite cannot diverge

PR #14115, opened 2026-04-03
feat(script): `ScriptConfig` generic `Network`, `FoundryEvmFactory`

mablr:
## Motivation

Towards generic `forge script`

Make `ScriptConfig` generic over `Network` and `FoundryEvmFactory`

- Removed useless bound  `Spec: From<SpecId>` on `Executor` 
- Wired `Config::evm_spec_id` generic `<SPEC: FromEvmVersion>`

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -1326,7 +1326,7 @@ impl Config {
         Ok(None)
     }
 
-    /// Returns the [SpecId] derived from the configured [EvmVersion]
+    /// Returns the Spec derived from the configured [EvmVersion]
     pub fn evm_spec_id<SPEC: FromEvmVersion>(&self) -> SPEC {
         evm_spec_id(self.evm_version)
     }


crates/script/src/lib.rs
```diff
@@ -605,15 +609,20 @@ struct JsonResult<'a, N: Network> {
 }
 
 #[derive(Clone, Debug)]
-pub struct ScriptConfig {
+pub struct ScriptConfig<N: Network, F: FoundryEvmFactory> {
```
mattsse, line 612: I do wonder if we could combine these two generics somehow, because I assume we only have a 1:1 mapping?

crates/script/src/lib.rs
```diff
@@ -605,15 +609,20 @@ struct JsonResult<'a, N: Network> {
...
     pub sender_nonce: u64,
     /// Maps a rpc url to a backend
-    pub backends: HashMap<String, Backend>,
+    pub backends: HashMap<String, Backend<N, F>>,
 }
 
-impl ScriptConfig {
+impl<N: Network, F: FoundryEvmFactory> ScriptConfig<N, F>
+where
+    N::TxEnvelope: Decodable + SignerRecoverable,
```
mattsse, line 622: this decodable bound looks a bit strange, where do we use this, maybe this should actually be 2178 encoding?

mattsse (approved):
smol q re decodable, but could be done separatel, in case there's anythig to do

crates/script/src/lib.rs
mablr: Decodable is necessary is Cheatcodes:
https://github.com/foundry-rs/foundry/blob/850a87ef72de85ca7814905671c07a98ea316f7f/crates/cheatcodes/src/evm.rs#L1216-L1227

crates/script/src/lib.rs
mablr: yep now we know `Executor` bounds we can merge them.

i'll do this as follow-up👍

mattsse (approved):

crates/script/src/lib.rs
mattsse: hmm, since this is broadcastraw, maybe this should even be decode2718

PR #14116, opened 2026-04-03
chore(evm): remove useless `Spec: From<SpecId>` bounds

mablr:
## Motivation

Remove useless `Spec: From<SpecId>` bounds everywhere.


figtracer (approved):
whoopsie

mattsse (approved):

PR #14121, opened 2026-04-03
refactor(cast): generic `cast run`

figtracer:
## Motivation
Makes `cast run` generic over `FoundryEvmNetwork`.

## Dependencies
- #14119 
- #14122


figtracer:
on it

PR #14136, opened 2026-04-04
fix(forge): preserve invariant replay failure reason

### LINKED ISSUE #14134: feat(invariant): replaying a broken invariant should show the property description being broken
### Component

Forge

When you initially break in invariant it prints
```
[FAIL: Property After All failed]
```
with code
```solidity
        assertTrue(_propertyAfterAll(), "Property After All failed");
```

When replaying, it always prints
```
[FAIL: invariantLP replay failure]
```
where `invariantLP` is the invariant function that breaks

This can make it difficult to debug when multiple different invariants broke.

### Additional context

_No response_

0xWeakSheep:
## Motivation

During invariant replay, Foundry shows a generic message like:

`[FAIL: <invariant_name> replay failure]`

This hides the actual property/assert message and makes debugging harder.

Closes #14134.

## Solution

- Preserve the replayed failure reason when available.
- Use that reason in replay output.
- Keep existing generic fallback if no reason is available.

So replay now shows the real failure message (for example `never owner`) instead of only `<invariant_name> replay failure`.

## Tests

Updated invariant CLI replay-output assertions to match the new behavior.

## AI Disclosure

This PR was developed with AI assistance.

### DIFF
--- a/crates/anvil/tests/it/anvil_api.rs
+++ b/crates/anvil/tests/it/anvil_api.rs
@@ -1190,3 +1190,116 @@ async fn test_anvil_reset_fork_to_non_fork() {
     let new_block = provider.get_block(BlockId::latest()).await.unwrap().unwrap();
     assert_eq!(new_block.header.number, 1);
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn can_get_node_info_tempo_t0() {
+    let (api, handle) = spawn(NodeConfig::test_tempo()).await;
+
+    let node_info = api.anvil_node_info().await.unwrap();
+
+    let provider = handle.http_provider();
+
+    let block_number = provider.get_block_number().await.unwrap();
+    let block = provider.get_block(BlockId::from(block_number)).await.unwrap().unwrap();
+
+    let expected_node_info = NodeInfo {
+        current_block_number: 0_u64,
+        current_block_timestamp: 1,
+        current_block_hash: block.header.hash,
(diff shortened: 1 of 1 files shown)


0xWeakSheep:
Thanks for the review. I added two replay regression tests:

- `invariant_replay_preserves_fail_reason`
- `invariant_replay_preserves_custom_error_reason`

Both run the same invariant twice and assert that replay preserves the specific failure reason (`panic: assertion failed (0x01)` for `fail()` and `InvariantCustomError(111, "custom")` for custom error), rather than showing a generic replay failure message.

0xWeakSheep:
Thanks for the heads-up. I checked the CI failure and it appears unrelated to this PR’s changes.

The failing job is `anvil::it proof::test_storage_proof` with:
`failed to spawn node: Address already in use (os error 98)`.

This PR only updates invariant replay tests under `crates/forge/tests/cli/test_cmd/invariant/mod.rs`, and does not touch Anvil/proof code paths. This looks like an infra/port-conflict flaky failure. I’m re-running CI.

0xWeakSheep:
I also added one more replay regression test for an invariant-level custom error:
`invariant_replay_preserves_invariant_custom_error_reason`.

It ensures that when the invariant function itself reverts with a custom error, replay still preserves and displays the same custom error message.

grandizzy (approved):
thank you, lgtm!

PR #14145, opened 2026-04-06
feat(invariant): persist optimization best value across runs

### LINKED ISSUE #14138: feat(invariant): optimization mode should reuse corpus with previous best value
### Component

Forge

Currently it starts over every time, even with no code changes.

### Additional context

_No response_

decofe:
## Motivation

Closes #14138. Invariant optimization mode starts from scratch every run — even with no code changes, the previously found best value and its sequence are lost. With corpus persistence enabled, the fuzzer should reuse the best-known optimization state so subsequent runs can build on prior progress.

## Solution

- **Persist optimization state**: After each run that improves the best value, write `optimization_best.json` (containing `best_value` + `best_sequence`) to the corpus directory alongside existing corpus entries.
- **Load on startup**: When the master worker initializes, load the persisted optimization state if it exists, and seed the invariant test with it — but only if the current invariant is actually in optimization mode (returns `int256`).
- **Optimization-aware corpus**: Sequences that improve the optimization value are now added to the in-memory corpus and persisted to disk even if they don't discover new edge coverage. This lets the coverage-guided mutation engine build on optimization-improving sequences.
- **Correct sequence tracking**: Each run tracks the exact call sequence prefix that achieved its best value (not the full terminal sequence), so the persisted state is always reproducible.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

Prompted by: grandizzy

### DIFF
--- a/crates/evm/evm/src/executors/corpus.rs
+++ b/crates/evm/evm/src/executors/corpus.rs
@@ -344,6 +344,19 @@ impl WorkerCorpus {
                     }
                 }
             }
+
+            // Seed in-memory corpus with the persisted optimization best sequence
+            // so the mutation engine can build on it in future runs.
+            if !optimization_best_sequence.is_empty() {
+                in_memory_corpus.push(CorpusEntry::new(optimization_best_sequence.clone()));
+                metrics.corpus_count += 1;
+                debug!(
+                    target: "corpus",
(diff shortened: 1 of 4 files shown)


grandizzy:
Superseded by new PR from foundry-rs branch

PR #14146, opened 2026-04-06
chore(evm): remove tempo deadcode

mablr:
## Motivation

Remove `initialize_tempo_precompiles_and_contracts`/`warm_tempo_precompile_accounts` deadcode (revert #14082), as we will impl a more generic logic.

- The tempo precompiles initialization is already made through `TempoEvmFactory` wrapped in `FoundryEvmFactory` impl.

Todo, as follow-up:
- tip20 tokens creation should be handled at the factory level for simple and pure generic approach.
- `warm_tempo_precompile_accounts` logic can be simply incorporated to `Executor`, no-op except for `TempoEvmNetwork`


figtracer:
> warm_tempo_precompile_accounts logic can be simply incorporated to Executor, no-op except for TempoEvmNetwork

+1, also just change the naming `warm_precompile_accounts` i think it looks better overall we can infer it's only for tempo per the impls

figtracer (approved):

PR #14148, opened 2026-04-06
fix(cast): add MegaETH system sender to known system senders

decofe:
MegaETH includes a system transaction (legacy type 0, `gasPrice=0`) as the second tx in every block from `0xa887dcb9d5f39ef79272801d05abdf707cfbbd1d`. This causes `cast run` to fail with `gas price is less than basefee` when replaying prior transactions, because revm validates the zero gas price against the block's non-zero `baseFeePerGas` (1000000).

Same pattern as #10608 — adds the MegaETH system sender to `is_known_system_sender()` so these txs are skipped during replay.

Example failing tx: `cast run 0x3775b355f222b27271b06819fb357696d4c72e0d366b4bb43b8d142aa27a8525 --rpc-url https://mainnet.megaeth.com/rpc`

Prompted by: georgen


grandizzy:
will do in https://github.com/foundry-rs/foundry/pull/14149

PR #14159, opened 2026-04-07
feat(evm): tempo genesis init and fork warmup into `FoundryEvmFactory`

mablr:
## Motivation

Upstream tempo genesis init and fork warmup into `FoundryEvmFactory` tempo impl

- Add `initialize_tempo_evm()` using `StorageCtx::enter_evm` to handle both fork and non-fork modes directly through the EVM context
- Non-fork: run full genesis init (precompile sentinels, TIP20 tokens, standard contracts) via `initialize_tempo_genesis_inner()`
- Fork: warm up precompile and TIP20 addresses with sentinel bytecode to prevent RPC storms during invariant fuzzing
- Extract `initialize_tempo_genesis_inner()` from `initialize_tempo_genesis()` for reuse within `StorageCtx::enter_evm` closures
- Remove `TempoStorageProvider` (no longer needed)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2721,6 +2721,7 @@ dependencies = [
  "serde_json",
  "tempfile",
  "tempo-alloy",
+ "tempo-contracts",
  "tempo-primitives",
  "tokio",
  "tracing",
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -4399,6 +4399,15 @@ where
     }
 }
 
+/// Replaces the cached hash of a [`Signed`] transaction, preserving the inner tx and signature.
+fn rehash<T>(signed: Signed<T>, hash: B256) -> Signed<T>
+where
+    T: alloy_consensus::transaction::RlpEcdsaEncodableTx,
+{
+    let (t, sig, _) = signed.into_parts();
+    Signed::new_unchecked(t, sig, hash)
+}
+
 /// Creates a `AnyRpcTransaction` as it's expected for the `eth` RPC api from storage data
 pub fn transaction_build(
     tx_hash: Option<B256>,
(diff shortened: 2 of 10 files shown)


crates/evm/core/src/tempo.rs
```diff
@@ -231,105 +69,106 @@ pub fn initialize_tempo_genesis(
...
+        STABLECOIN_DEX_ADDRESS,
+        TIP20_FACTORY_ADDRESS,
+        TIP403_REGISTRY_ADDRESS,
+        TIP_FEE_MANAGER_ADDRESS,
+        VALIDATOR_CONFIG_ADDRESS,
+        VALIDATOR_CONFIG_V2_ADDRESS,
+        ACCOUNT_KEYCHAIN_ADDRESS,
+        SIGNATURE_VERIFIER_ADDRESS,
+        ADDRESS_REGISTRY_ADDRESS,
+    ] {
```
zerosnacks: can we rely on https://github.com/mablr/foundry/blob/f212cd2393055d45fcdf1eaa20499f4b5659bf36/crates/evm/core/src/tempo.rs#L36-L46? this makes it more difficult to miss newly introduced precompiles

basically we want a single declaration point; either in tempo or in foundry

crates/evm/core/src/evm.rs
```diff
@@ -649,6 +655,43 @@ pub struct TempoFoundryEvm<
...
+) {
+    let ctx = &mut evm.inner.inner.ctx;
+    StorageCtx::enter_evm(&mut ctx.journaled_state, &ctx.block, &ctx.cfg, &ctx.tx, || {
+        if is_forked {
+            // In fork mode, warm up precompile accounts to avoid repeated RPC fetches.
+            let mut sctx = StorageCtx;
+            let sentinel = Bytecode::new_legacy(Bytes::from_static(&[0xef]));
+            for addr in TEMPO_PRECOMPILE_ADDRESSES
+                .iter()
+                .chain(&[SIGNATURE_VERIFIER_ADDRESS, ADDRESS_REGISTRY_ADDRESS])
```
zerosnacks: I believe these are now available here, https://github.com/mablr/foundry/blob/f212cd2393055d45fcdf1eaa20499f4b5659bf36/crates/evm/core/src/tempo.rs#L30-L34

I think these can just be folded into the `TEMPO_PRECOMPILE_ADDRESSES`

crates/evm/core/src/tempo.rs
mablr: good point

mablr (commented):
@zerosnacks Updated according to your comments 👍

PR #14162, opened 2026-04-07
chore(cast): remove `src/tempo.rs`, inline helpers

figtracer:
## Motivation
Inlines Tempo helpers where they're used and removes `src/tempo.rs`.
- `is_key_provisioned` → private in `erc20.rs`
- `iso4217_warning_message` + `is_iso4217_currency` → `tip20.rs`


crates/cast/src/cmd/tempo/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+    let cast = CastTxSender::new(provider);
+    cast.print_tx_result(tx_hash, cast_async, confirmations, timeout).await
+}
+
+// ── token_send! macro ───────────────────────────────────────────────────────
+
+/// Macro for the keychain-vs-normal send pattern used by token commands.
+///
+/// The closure receives `$provider` and must return a `N::TransactionRequest`.
+macro_rules! token_send {
```
figtracer: extracted `erc20_send!` -> `token_send!`

crates/cast/src/cmd/tempo/tip20.rs
```diff
@@ -0,0 +1,175 @@
...
+                    (None, None)
+                }
+            }
+        };
+
+        // TIP-20 operations always use TempoNetwork.
+        self.run_generic::<TempoNetwork>(signer, tempo_access_key).await
+    }
+
+    async fn run_generic<N: Network + RecommendedFillers>(
```
mablr: Since it is a tempo-only command, we don't need the generic `Network` dispatch. Let's directly go with `TempoNetwork` `Provider`, it would be simpler and more concise.

crates/cast/src/cmd/tempo/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+    tx: &mut N::TransactionRequest,
+    config: Option<&TempoAccessKeyConfig>,
+) where
+    N::TransactionRequest: FoundryTransactionBuilder<N>,
+{
+    if let Some(config) = config {
+        tx.set_from(config.wallet_address);
+        tx.set_key_id(config.key_address);
+    }
+}
```
mablr: Such logic should go to `FoundryTransactionBuilder`

PR #14167, opened 2026-04-07
feat(script): add `--batch` flag for Tempo native batching

mablr:
## Motivation
Adds `--batch` flag to forge script to leverage Tempo's native call batching (type 0x76 transactions). When enabled, all vm.broadcast() calls are collected and sent as a single atomic batch transaction instead of individual transactions.

Also adds `--batch-size` flag (requires --batch) to control the number of calls per batch transaction (default: 100).

Changes:
- Add `broadcast_batch()` on `BundledState<TempoEvmNetwork>` from https://github.com/tempoxyz/tempo-foundry/pull/222 
- Refactor `run_generic_script` into `prepare_bundled` + `broadcast` dispatch so Tempo path can call `broadcast_batch()` on the concrete type

### DIFF
--- a/crates/script/src/lib.rs
+++ b/crates/script/src/lib.rs
@@ -12,7 +12,7 @@ extern crate foundry_common;
 #[macro_use]
 extern crate tracing;
 
-use crate::runner::ScriptRunner;
+use crate::{broadcast::BundledState, runner::ScriptRunner};
 use alloy_json_abi::{Function, JsonAbi};
 use alloy_network::Network;
 use alloy_primitives::{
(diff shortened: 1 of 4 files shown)


crates/script/src/lib.rs
```diff
@@ -122,10 +122,19 @@ pub struct ScriptArgs {
...
+    /// This provides atomicity (all-or-nothing execution) and gas savings.
+    #[arg(long)]
+    pub batch: bool,
+
+    /// Number of calls per Tempo batch transaction.
+    ///
+    /// When `--batch` is enabled, splits the collected calls into multiple batch
+    /// transactions of at most this many calls each.
+    #[arg(long, requires = "batch", default_value = "100")]
     pub batch_size: usize,
```
figtracer, line 138: can we just do like `batch: Option<usize>` that way we don't need 2 different flags for this, we assume that if size=0 there's no batching?

mablr: Imo, having 2 Clap flags offers a better UI.

Having `--batch <SIZE>` defaulting to 100 may be misleading.

PR #14169, opened 2026-04-07
feat(anvil): add TIP20 token operation tests

stevencartavia:
Add 6 tests for TIP20 token operations in Tempo mode: 
- test_tip20_transfer
- test_tip20_approve_and_transfer_from
- test_tip20_total_supply
- test_transfer_between_different_fee_tokens
- test_all_fee_tokens_have_correct_metadata
- test_transfer_emits_event

Also expands the IERC20 sol interface with totalSupply, allowance, approve, and transferFrom.


zerosnacks:
@figtracer makes sense - I prefer we complete the transition prior to starting any development of the next Anvil (as it is quite an undertaking and we need to do some more preliminary spec'ing)

PR #11127, opened 2025-07-28
fix(`cast`): unknown signatures are cached as an empty string

zerosnacks:
## Motivation

Closes: https://github.com/foundry-rs/foundry/issues/11125

## Solution

Make sure signatures are not returned empty, previously we were caching empty results ("") preventing refetching.

previously the query would return as follows:

```
query: [Error(0x37d01491)]
res: [[]]
len(res): 1
```

meaning `0x37d01491` got paired with `[]`, yielding empty `""` string when written to cache

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #14172, opened 2026-04-07
fix(traces): don't persist unknown signatures to disk cache

### LINKED ISSUE #11125: Unknown signatures are cached as an empty string
### Component

Cast

- [x] Foundry
- [x] Foundryup

forge Version: 1.2.3-stable

foundryup: 1.3.0

cast decode-error

### Operating System

Linux

### Describe the bug

`cast decode-error` caches unknown selectors as an empty string.

Reproduction steps:

1. Encode an error that doesn't exist in https://openchain.xyz/signatures. For example:
    ```bash
    $ cast calldata 'ErrorThatDoesntExistIHope(uint256,uint256)' 1 2
    0x37d0149300000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000002
    ```
2. Try to decode it:
    ```bash
    $ cast decode-error 0x37d0149300000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000002
    Error: No matching error signature found for selector `37d01493`
    ```
3. [Add that error](https://openchain.xyz/signatures/import) to the Signature database
4. Try to decode it again. It doesn't work (though `cast 4byte 37d01493` does).
5. Open `~/.foundry/cache/signatures`. You'll see an entry that is `"0x37d01493":""`. Remove it.
6. Try to decode the error again. This time it works.

Of course, use a different error because I already added that one while writing these reproduction steps.

skyc1e:
Fixes #11125.

When OpenChain returns no results for a selector, the signature cache stores `None` in memory. During serialization this becomes an empty string `""` in `~/.foundry/cache/signatures`. On reload, the empty string maps back to `None` and `contains_key` returns `true`, so the selector is never re-queried — even if the signature has since been added to the database.

The previous fix (#11127) stopped caching negative results entirely, but was reverted (#11140) because it caused repeated API calls for genuinely unknown selectors within a single session.

This patch takes a middle path: unknown selectors stay in the in-memory cache (preventing repeated API calls during a session) but are skipped when writing to disk. On the next run, they're re-queried.

Existing cache files with stale empty-string entries self-heal: they load as `None`, and on next save are dropped.

One file changed, one test added.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @danipopes @mattsse @grandizzy @yash-atreya @zerosnacks @onbjerg @0xrusowsky
+* @danipopes @mattsse @grandizzy @zerosnacks @onbjerg @0xrusowsky @mablr @figtracer @stevencartavia
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4691,6 +4691,8 @@ dependencies = [
  "serde",
  "serde_json",
  "tempfile",
+ "tempo-alloy",
+ "tempo-primitives",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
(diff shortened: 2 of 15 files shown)


figtracer:
pls fix CI @skyc1e

PR #14226, opened 2026-04-09
feat(invariant): show best value realtime in optimization mode

### LINKED ISSUE #14133: feat(invariant): optimization mode should show best value realtime during execution
### Component

Forge

Currently you have to CTRL+C out to see the best result. It would be nice if you can see it realtime to judge if it is worth continuing.

### Additional context

_No response_

figtracer:
Show the current best optimization value in the `--show-progress` bar during invariant campaigns, so users can judge whether to keep running.

Closes #14133


mablr (approved):

zerosnacks (approved):
Makes sense! cc @grandizzy for visibility

PR #14235, opened 2026-04-10
clippy: warn on `if_then_some_else_none` (autofix)

decofe:
Enable `clippy::if_then_some_else_none` and fix all findings.

All fixes are auto-applied via `cargo clippy --fix` and formatted with `cargo +nightly fmt`.

Replaces `if cond { Some(x) } else { None }` with `cond.then(|| x)` / `cond.then_some(x)`.

Prompted by: zerosnacks


mattsse (approved):
these are actually pretty good

mablr (approved):

zerosnacks:
> these are actually pretty good

I'm slowly implementing the majority of the lint rules of Reth as I like them a lot, the PRs are split up per lint rule so it is easy to review 😄

PR #14242, opened 2026-04-10
fix(script): Tempo direct signer + gas calc logic

mablr:
## Motivation

Fix `forge script` to handle Tempo's direct signer and specific gas calc logic.

Tested against ✅ https://github.com/tempoxyz/tempo-foundry/blob/tempo/.github/scripts/tempo-wallet.sh

### DIFF
--- a/crates/cli/src/utils/cmd.rs
+++ b/crates/cli/src/utils/cmd.rs
@@ -112,7 +112,12 @@ pub fn init_progress(len: u64, label: &str) -> indicatif::ProgressBar {
 
 /// True if the network calculates gas costs differently.
 pub fn has_different_gas_calc(chain_id: u64) -> bool {
-    if let Some(chain) = Chain::from(chain_id).named() {
+    let chain = Chain::from(chain_id);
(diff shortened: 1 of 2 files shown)


figtracer (approved):
🥳

crates/cli/src/utils/cmd.rs
```diff
@@ -112,7 +112,12 @@ pub fn init_progress(len: u64, label: &str) -> indicatif::ProgressBar {
 
 /// True if the network calculates gas costs differently.
 pub fn has_different_gas_calc(chain_id: u64) -> bool {
-    if let Some(chain) = Chain::from(chain_id).named() {
+    let chain = Chain::from(chain_id);
+    // Is either Tempo | TempoModerato | TempoTestnet | TempoDevnet
+    if chain.is_tempo() || chain.id() == 31318 {
```
zerosnacks, line 117: I'll double check if its OK to publish `TempoDevnet` in `alloy-chains`, would avoid this ugly hack

For now its OK

zerosnacks (approved):
👍 !

PR #14244, opened 2026-04-10
fix(anvil): return second offset from evm_setTime

cuiweixie:
## Summary

`evm_setTime` documents that it returns the number of **seconds** between the requested timestamp and the current time. The offset was computed in seconds, but `Duration::from_millis(offset).as_secs()` incorrectly interpreted that value as milliseconds, dividing the returned offset by 1000.

Return the offset directly.

## Test plan

- `cargo check -p anvil` (optional: full `make pr` per CONTRIBUTING.md)

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -485,7 +485,7 @@ impl<N: Network> EthApi<N> {
 
         // number of seconds between the given timestamp and the current time.
         let offset = timestamp.saturating_sub(now);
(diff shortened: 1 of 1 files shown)


cuiweixie:
```
 cast rpc evm_setTime $(($(date +%s) + 60)) --rpc-url http://localhost:8545
Returns 0 instead of 60 without this pr
```

zerosnacks:
Thanks for your PR; decided to implement an alternative solution here https://github.com/foundry-rs/foundry/pull/14292 - added you as co-author

PR #14246, opened 2026-04-10
fix(cli): restore short flag for `--rpc-url`

mablr:
## Motivation

Restore short flag for `--rpc-url`, regression from #14224.

Restored `estimate_base_da` test to use short flag.


figtracer (dismissed):

zerosnacks:
>     Message:  Command coverage: Short option names must be unique for each argument, but '-r' is in use by both 'report_file' and 'rpc_url'

Fine to remove shorthand of `report_file`

zerosnacks (approved):

figtracer (approved):

PR #14248, opened 2026-04-10
fix(forge): remember last test path in forge test --watch

cuiweixie:
## Summary

`watch_test` keeps a `last_test_files` cache so that when a non-test Solidity file changes, it can still pass `--match-path` for the most recently touched test file. That cache was never populated when a single test file changed, so the fallback stayed empty and the watcher skipped reconfiguration.

Persist the detected test path when exactly one `.t.sol` file changes.

## Test plan

- Manual: `forge test --watch`, edit a test file then a src file; expect tests to rerun with the prior test path filter.
- `cargo check -p forge`


emmajam:

PR #14250, opened 2026-04-10
fix(cast): `batch-send` handling by clearing `to`/`value` fields

mablr:
## Motivation

Fix `cast batch-send` Tempo CI tests, by clearing systematically `to`/`value` fields.

https://github.com/tempoxyz/tempo-foundry/blob/e95b51a3f845d8453622ed8ebbbb27fe8721b057/.github/scripts/tempo-check.sh#L527-L540


zerosnacks (approved):
Makes sense 👍

stevencartavia (approved):

PR #14253, opened 2026-04-10
feat(cheatcodes): external cheatcode plugin API

decofe:
Adds an extension API that lets Foundry library consumers register custom cheatcodes without forking. External handlers are dispatched when a call to the cheatcode address does not match any built-in selector.

## Motivation

Ecosystem teams want to extend Foundry with custom cheatcodes (assertions, storage snapshots, invariant checks) while reusing the fuzzer, test runner, and EVM runtime. Today the only way is to fork and modify the sealed `VmCalls` enum.

## API

Three public types in `foundry-cheatcodes`:

```rust
#[non_exhaustive]
pub enum ExternalCheatcodeOutcome {
    Unhandled,
    Return(Vec<u8>),
    Revert(Error),
}

pub trait CheatcodeHost {
    fn caller(&self) -> Address;
    fn gas_limit(&self) -> u64;
    fn load(&mut self, account: Address, slot: U256) -> Result<U256>;
    fn balance(&mut self, account: Address) -> Result<U256>;
    fn store(&mut self, account: Address, slot: U256, value: U256) -> Result<()>;
    fn set_balance(&mut self, account: Address, value: U256) -> Result<()>;
    fn set_code(&mut self, account: Address, code: Bytes) -> Result<()>;
}

pub trait ExternalCheatcode: Send + Sync + Debug + 'static {
    fn call(&self, host: &mut dyn CheatcodeHost, calldata: &[u8]) -> ExternalCheatcodeOutcome;
}
```

`CheatcodeHost` is object-safe and non-generic, so consumers never see `FoundryEvmNetwork`.

## Usage

```rust
#[derive(Debug)]
struct MyCheatcodes;

impl ExternalCheatcode for MyCheatcodes {
    fn call(&self, host: &mut dyn CheatcodeHost, calldata: &[u8]) -> ExternalCheatcodeOutcome {
        if calldata.len() < 4 { return ExternalCheatcodeOutcome::Unhandled }
        // decode, read/write state via host, return data or revert
        ExternalCheatcodeOutcome::Return(Vec::new())
    }
}

let executor = ExecutorBuilder::new()
    .inspectors(|stack| {
        stack
            .cheatcodes(cheats_config)
            .external_cheatcode(Arc::new(MyCheatcodes))
    })
    .build(env, backend);
```

## Implementation

- `CheatcodeHostImpl<FEN>` bridges `FoundryContextFor<FEN>` to the object-safe `CheatcodeHost`, routing reads through `load_account` + `sload` and writes through `sstore`/`set_code` on the journal. `set_balance` records `DealRecord` for revert rollback. `store` and `set_code` enforce precompile guards.
- Dispatch runs on the `UnknownSelector` cold path only — zero overhead for built-in cheatcodes.
- Handlers stored as `Vec<Arc<dyn ExternalCheatcode>>` for clone safety during fuzzing.
- Builder auto-enables `Cheatcodes` when external handlers are registered without `.cheatcodes()`.

## Changes

- `crates/cheatcodes/src/lib.rs` — `ExternalCheatcodeOutcome`, `CheatcodeHost`, `ExternalCheatcode` traits; `CheatsCtxt` accessor methods
- `crates/cheatcodes/src/inspector.rs` — `CheatcodeHostImpl`, registration, fallback dispatch, 7 unit tests
- `crates/evm/evm/src/inspectors/stack.rs` — `.external_cheatcode()` builder method, auto-enable logic

## Testing

```
cargo test -p foundry-cheatcodes -- inspector::tests  # 7 tests
cargo clippy -p foundry-cheatcodes -p foundry-evm -p forge -- -D warnings  # clean
```

Prompted by: zerosnacks


zerosnacks:
cc @odyslam is this the kind of API that you are looking for?

mablr (commented):
I like this lightweight design masking `FoundryEvmNetwork`.

To go further, what if `foundry-evm` is moved to `foundry-core` and published, making users able to impl `FoundryEvmNetwork` for their own primitives? Consequently, we would no longer need the `CheatcodeHost` abstraction and could provide the capability to directly interact w/ EVM internals through `FoundryContextExt`, offering the highest level possible of EVM customizability.

Both approaches can co-exist btw.
- Current's PR "low-hanging fruit" introduction to external cheatcodes
- the one suggested above as F2.0 target ?

PR #14256, opened 2026-04-10
fix(anvil): reject setNextBlockTimestamp when <= last block time

cuiweixie:
## Summary

`TimeManager::set_next_block_timestamp` documents: **Fails if it's before (or at the same time) the last timestamp.** The implementation only compared with `<`, so a timestamp **equal** to the last block was incorrectly accepted.

- Use `<=` for the check and align the error message with “must be after” semantics.
- Adjust `test_mine_blk_with_prev_timestamp` to expect failure when the requested time equals the previous block’s timestamp, and mine with `last + 1` for the success path.

## Test plan

- `cargo test -p anvil test_mine_blk_with_prev_timestamp`


emmajam:

PR #14258, opened 2026-04-10
fix(forge): `script`/`test` Tempo selection when used w/ Anvil

mablr:
## Motivation

As `forge script` was only inferring network from chain id, on anvil it would always fallback to Eth.

Now to select Tempo on `forge test` and `forge script` you can either:
- explicitly use `--tempo`, fee token will be implicitly set to default
- implicitly use `--tempo.fee-token`

This was tested with Tempo CI's Anvil section:
https://github.com/tempoxyz/tempo-foundry/blob/e95b51a3f845d8453622ed8ebbbb27fe8721b057/.github/scripts/tempo-check.sh#L730-L894 (everything pass expect `cast` expiring nonce test which is not related, fixed in #14259)


figtracer (dismissed):

stevencartavia (dismissed):

mablr (commented):
fixed `EvmArgs` serialization to be skipped when no flag provided

figtracer (approved):

stevencartavia (approved):

zerosnacks (approved):

PR #14261, opened 2026-04-11
fix(anvil): guard deposit tx parsing in Tempo mode

stevencartavia:
Skip `get_deposit_tx_parts()` in `build_call_env()` when running in Tempo mode.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -1412,8 +1412,8 @@ impl<N: Network> Backend<N> {
             evm_env.cfg_env.disable_base_fee = true;
         }
 
-        // Deposit transaction? (not supported in Tempo mode)
-        if !self.is_tempo()
+        // Deposit transaction? (only valid when op-stack deposits are active)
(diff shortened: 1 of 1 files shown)


mablr (commented):
Unsure it's necessary.

zerosnacks (approved):

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -1412,8 +1412,10 @@ impl<N: Network> Backend<N> {
             evm_env.cfg_env.disable_base_fee = true;
         }
 
-        // Deposit transaction?
-        if let Ok(deposit) = get_deposit_tx_parts(&other) {
+        // Deposit transaction? (not supported in Tempo mode)
+        if !self.is_tempo()
```
zerosnacks: Maybe better to instead use the ensure_op_deposits_active check rather than a negative inline network

zerosnacks (changes requested):

zerosnacks (approved):

mablr (approved):

PR #14264, opened 2026-04-11
evm: only classify skip reverts from cheatcode address

ArshLabs:
## Motivation

`RawCallResult::into_evm_error` classified any revert payload prefixed with `FOUNDRY::SKIP` as `EvmError::Skip`.

That means a test contract can forge raw revert bytes with that prefix and have a real failure reported as skipped.

## Solution

Gate skip classification on revert origin.

`into_evm_error` now returns `EvmError::Skip` only when:
- revert payload decodes as `SkipReason`, and
- `reverter == CHEATCODE_ADDRESS`

If the same payload comes from any other address (or no tracked reverter), execution is reported as a normal `EvmError::Execution`.

Added regression tests in `crates/evm/evm/src/executors/mod.rs` for:
- cheatcode-originated skip payload
- forged skip payload from non-cheatcode address
- skip payload with no reverter

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mablr (dismissed):
Thanks, lgtm.

figtracer (dismissed):
nice

decofe:
The fix in `into_evm_error` is correct, but there's a second `SkipReason::decode` call in `RevertDecoder::maybe_decode` (`crates/evm/core/src/decode.rs`) that isn't guarded:

```rust
pub fn maybe_decode(&self, err: &[u8], status: Option<InstructionResult>) -> Option<String> {
    if let Some(reason) = SkipReason::decode(err) {
        return Some(reason.to_string());
    }
    // ...
}
```

This matters because the fuzz aggregator in `crates/evm/evm/src/executors/fuzz/mod.rs` calls `rd.maybe_decode()` on counterexample results, and then `aggregate_results` runs `SkipReason::decode_self` on that string — setting `result.skipped = true`. So a forged skip payload in a fuzz run can still be marked as skipped via this path.

Since `into_evm_error` now handles the legitimate skip case (with the reverter guard) before ever calling `rd.decode()`, the `SkipReason::decode` in `maybe_decode` can be removed — legitimate skips never reach it.

zerosnacks (dismissed):
Thanks! Generally lgtm, minor feedback above

ArshLabs:
> Thanks! Generally lgtm, minor feedback above

re-audited the skip flow end to end and pushed a follow-up commit.

skip classification now only happens in `into_evm_error` with the cheatcode-origin guard, and `maybe_decode` no longer interprets `FOUNDRY::SKIP` as a skip marker so basically this closes the fuzz path where `maybe_decode` output could be re-decoded and marked as skipped.

added a regression test to lock this behavior in and rebased the branch on latest master.

mablr (approved):
Should be good now, thanks.

zerosnacks (approved):
lgtm - thanks!

Issue #13575, opened 2026-03-02
Add gas/s and tx/s to the invariant metrics JSON


aviggiano:
### Component

Forge

### Describe the feature you would like

I would like to see foundry's tx/s and gas/s as part of the invariant test metrics JSON

### Additional context

See https://github.com/Recon-Fuzz/scfuzzbench/issues/112

aviggiano:
Closed per https://github.com/foundry-rs/foundry/pull/14266

PR #13898, opened 2026-03-23
feat: add gas/s and tx/s throughput metrics to invariant test output

danieljrc888:
## Summary

- Add throughput metrics (`gas/s`, `tx/s`) to invariant test output for benchmarking tools like scfuzzbench
- Track cumulative gas and elapsed time during invariant campaigns
- Add `gas` field to `InvariantMetrics` for per-selector gas tracking in the metrics table
- Extend inline JSON progress logging with `total_txs`, `total_gas`, `tx_per_sec`, `gas_per_sec`
- Display `gas/s` and `tx/s` in the human-readable test report line
- Update snapshot parsing regex to handle new format

## Example output

```
Ran 2 tests for test/Vault.invariant.t.sol:VaultInvariantTest
[PASS] invariant_depositWithdrawAccounting() (runs: 20, calls: 2000, reverts: 0, gas/s: 1054698686, tx/s: 15399)

╭--------------+----------+-------+---------+----------+----------╮
| Contract     | Selector | Calls | Reverts | Discards | Gas      |
+=================================================================+
| VaultHandler | deposit  | 1041  | 0       | 0        | 65616418 |
|--------------+----------+-------+---------+----------+----------|
| VaultHandler | withdraw | 959   | 0       | 0        | 71370771 |
╰--------------+----------+-------+---------+----------+----------╯

[PASS] invariant_solvency() (runs: 20, calls: 2000, reverts: 0, gas/s: 1171893077, tx/s: 17092)

╭--------------+----------+-------+---------+----------+----------╮
| Contract     | Selector | Calls | Reverts | Discards | Gas      |
+=================================================================+
| VaultHandler | deposit  | 1041  | 0       | 0        | 65559742 |
|--------------+----------+-------+---------+----------+----------|
| VaultHandler | withdraw | 959   | 0       | 0        | 71566351 |
╰--------------+----------+-------+---------+----------+----------╯

Suite result: ok. 2 passed; 0 failed; 0 skipped; finished in 140.49ms (265.67ms CPU time)
```

Closes #13575

## Test plan

- [x] Unit tests for `TestKindReport::Invariant` display (normal, zero elapsed, optimization mode, failed corpus replays)
- [x] Unit test for metrics table with Gas column
- [x] Snapshot parsing tests updated for new format
- [x] Integration test snapshots updated across all invariant test files
- [x] `cargo build -p forge` passes
- [x] `cargo test -p forge --lib` passes (36 tests)

PR #14266, opened 2026-04-11
Addresses scfuzzbench/scfuzzbench#112.

0xKarl98:
This closes issue #13575 
cc @mattsse 

  ## Summary

  This PR ports and narrows the original approach from #13898 onto current `master`.

  The issue is that Foundry's invariant progress JSON currently exposes corpus/coverage-
  style metrics, but does not expose throughput data. As a result, downstream benchmark
  tooling such as `scfuzzbench` cannot recover Foundry `tx/s` and `gas/s` samples, so the
  corresponding throughput time-series charts remain blank for Foundry runs.

  This PR extends the existing invariant progress JSON output with campaign-level throughput
  fields:
  - `total_txs`
  - `total_gas`
  - `tx_per_sec`
  - `gas_per_sec`

  ## What This PR Changes

  ### 1. Track campaign-level throughput during invariant runs

  This PR introduces a small invariant-local throughput accumulator that records:
  - the total number of executed calls in the campaign
  - the total gas consumed by those calls

  The accumulator is updated only for non-discard calls, which keeps the semantics aligned
  with real executed work and avoids counting `vm.assume` rejections as throughput.

  ### 2. Extend the existing invariant progress JSON event

  The existing machine-readable invariant progress payload is extended to include:
  - `total_txs`
  - `total_gas`
  - `tx_per_sec`
  - `gas_per_sec`

  This keeps throughput data in the same JSON event shape already emitted for invariant
  progress, instead of introducing a separate reporting path.

  ### 3. Include cumulative and derived throughput fields

  Issue #112 specifically asks for `tx/s` and `gas/s`, so this PR includes `tx_per_sec` and
  `gas_per_sec` directly.

  It also includes `total_txs` and `total_gas` because they are the underlying cumulative
  measurements from which the rates are derived. This makes the JSON output more robust for
  downstream consumers, which can either use the reported rates directly or recompute them

  This PR adds focused tests covering:
  - the presence and correctness of the new throughput fields
  - zero-elapsed handling, which must safely report `0.0` rates instead of dividing by zero
  - preservation of the existing optional `optimization_best` field behavior

  ## How This Addresses scfuzzbench/scfuzzbench#112

  `scfuzzbench` generates `tx/s` and `gas/s` time-series charts by parsing machine-readable
  throughput samples from fuzzer logs.

  This PR makes Foundry emit those throughput values as part of its invariant progress JSON,
  so Foundry runs can produce the same throughput samples expected by the benchmark
  pipeline.

  ## Scope

  This PR is intentionally limited to invariant progress JSON output.

  It does not change:
  - the final human-readable invariant test report line
  - snapshot formatting or parsing
  - the invariant selector metrics table
  - unrelated invariant reporting or metrics behavior

  ## Test Plan

  - [x] Add focused unit tests for invariant progress JSON throughput fields
  - [x] Add zero-elapsed coverage for safe rate calculation
  - [x] `cargo test -p foundry-evm invariant_progress_json --lib`

### DIFF
--- a/crates/evm/evm/src/executors/invariant/mod.rs
+++ b/crates/evm/evm/src/executors/invariant/mod.rs
@@ -144,11 +144,21 @@ impl InvariantThroughputMetrics {
     }
 }
 
+/// Converts a cumulative campaign total into an average per-second rate.
+///
+/// Returns `0.0` during the initial zero-elapsed startup window to avoid
+/// dividing by zero while progress reporting is warming up.
 fn rate_per_sec(total: f64, elapsed: Duration) -> f64 {
     let elapsed_secs = elapsed.as_secs_f64();
     if elapsed_secs > 0.0 { total / elapsed_secs } else { 0.0 }
 }
 
+/// Builds the machine-readable invariant progress payload emitted during a
+/// campaign.
+///
+/// This keeps the existing corpus progress metrics together with cumulative and
+/// derived throughput fields so downstream benchmark tooling can consume a
+/// single JSON event shape.
 fn build_invariant_progress_json<M: Serialize>(
     timestamp_secs: u64,
     invariant_name: &str,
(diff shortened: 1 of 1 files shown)


mablr (approved):
Lgtm, thanks.

Pending another review.

zerosnacks (approved):
Lgtm 👍

PR #14273, opened 2026-04-12
fix(traces): avoid panic on invalid internal source spans

mk0walsk:
## Motivation

Malformed or preprocessed source spans can still reach `jump_out` while `--decode-internal` is decoding function inputs and outputs. Unlike `parse_function_from_loc`, this path sliced `source[start..end]` directly and could panic on an out-of-range span, aborting trace/debug output.

## Solution

Route source slicing through a shared helper that uses `checked_add` and `str::get`, and reuse it from both `jump_out` and `parse_function_from_loc`. Added a small unit test covering valid and invalid ranges.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/traces/src/debug/mod.rs
+++ b/crates/evm/traces/src/debug/mod.rs
@@ -1,8 +1,8 @@
 mod sources;
 use crate::CallTraceNode;
 use alloy_dyn_abi::{
-    DynSolType, DynSolValue, Specifier,
     parser::{Parameters, Storage},
+    DynSolType, DynSolValue, Specifier,
 };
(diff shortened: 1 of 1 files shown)


mablr (dismissed):
Thanks, lgtm. (pending another review and formatting)

mk0walsk:
@mablr Thinks! Fixed the rustfmt/clippy.

mablr (approved):
lgtm

zerosnacks (approved):
Thanks! This also looks correct to me

PR #14275, opened 2026-04-13
 feat(invariant): add fail_on_assert for assertion failures in invariant campaigns

0xKarl98:
This closes issue #13322
  cc @grandizzy @mattsse 

  ## Summary

  This PR implements the `[invariant].fail_on_assert` path discussed in #13322 and ports it onto current `master` without changing the existing default invariant semantics.

  Today, invariant campaigns with `fail_on_revert = false` discard handler-level assertion failures together with ordinary reverts. As a result, the following bug-signals can be
  silently ignored during invariant runs:

  - `Panic(0x01)` from Solidity `assert()`
  - legacy invalid-opcode assertion behavior
  - Forge `vm.assert*` failures, including the non-reverting `GLOBAL_FAIL_SLOT` path when `assertions_revert = false`

  This PR introduces a dedicated invariant-level assertion switch:

  - `[invariant].fail_on_assert = false` by default

  When enabled, assertion-like failures are reported as invariant failures even if ordinary reverts are still ignored.

  ## What This PR Changes

  ### 1. Add an invariant-specific assertion switch

  This PR adds `[invariant].fail_on_assert` to `InvariantConfig` and keeps the default value `false`.

  This follows the configuration-based route explicitly allowed by #13322 instead of changing the historical default behavior of `fail_on_revert = false`.

  ### 2. Separate assertion semantics from generic revert semantics

  This PR does not treat assertions as “just another revert”.

  Instead, invariant execution now recognizes assertion-like failures through a single invariant-local classification path covering:

  - Solidity `assert()` panics (`Panic(0x01)`)
  - legacy invalid-opcode assertions
  - `vm.assert*` reverting failures
  - `vm.assert*` non-reverting failures surfaced through `GLOBAL_FAIL_SLOT`
  - assertion console logs used by non-reverting cheatcode assertions

  This keeps ordinary handler reverts governed by `fail_on_revert`, while allowing assertion failures to be elevated independently.

  ### 3. Keep replay and shrinking aligned with the original failure type

  This PR persists whether a recorded invariant failure originated from an assertion and threads that through replay/shrinking.

  Without this, an assertion-originated counterexample can be replayed or shrunk under the wrong semantics, especially when `fail_on_revert = false`.

  ### 4. Invalidate persisted failures when assertion semantics change

  `InvariantSettings` now includes `fail_on_assert` so persisted invariant failures are ignored when replay settings no longer match the original assertion handling mode.

  This keeps replay behavior reproducible and avoids reusing stale failures across semantic changes.

  ### 5. Add focused regression coverage

  This PR adds invariant CLI coverage for:

  - Solidity `assert` reported when `fail_on_assert = true`
  - Solidity `assert` ignored when `fail_on_assert = false`
  - non-assert panics still ignored
  - ordinary `require` reverts still ignored
  - `vm.assert*` reverting failures reported when `fail_on_assert = true`
  - `vm.assert*` reverting failures ignored when `fail_on_assert = false`
  - `vm.assert*` non-reverting `GLOBAL_FAIL_SLOT` failures reported when `fail_on_assert = true`
  - `vm.assert*` non-reverting `GLOBAL_FAIL_SLOT` failures ignored when `fail_on_assert = false`
  - replay of persisted assertion-originated failures

  ## How This Addresses #13322

  #13322 asks for assertion failures during invariant campaigns to stop being silently discarded when `fail_on_revert = false`.

  This PR addresses that by introducing a narrower invariant-specific control:

  - ordinary reverts can still be ignored
  - assertion failures can still be reported as bugs

  This matches the issue’s stated requirement more closely than reusing `fail_on_revert`, which is too coarse because it treats all reverts the same.

  ## Scope

  This PR is intentionally limited to assertion handling in invariant testing.

  It does not change:

  - the default value of `fail_on_revert`
  - generic revert handling for invariant campaigns
  - unrelated invariant metrics/UI behavior
  - multi-failure aggregation or continuous-run behavior


zerosnacks:
Nice! When time permits @grandizzy please have a look

crates/evm/evm/src/executors/invariant/error.rs
```diff
@@ -81,6 +83,13 @@ impl FailedInvariantCaseData {
             .with_abis(targeted_contracts.targets.lock().values().map(|c| &c.abi))
             .with_abi(invariant_contract.abi)
             .decode(call_result.result.as_ref(), call_result.exit_reason);
+        // Non-reverting assertion failures surface through Foundry's failure flags instead of
+        // revert data. Use a stable fallback so invariant output is not blank.
+        let revert_reason = if revert_reason.is_empty() && !call_result.reverted {
+            "assertion failed".to_string()
```
grandizzy: let's move the `assertion fail` in a const and reuse

crates/evm/evm/src/executors/invariant/error.rs
```diff
@@ -97,6 +106,17 @@ impl FailedInvariantCaseData {
...
             fail_on_revert: invariant_config.fail_on_revert,
+            assertion_failure: false,
         }
     }
+
+    /// Marks this case as assertion-originated and normalizes empty decoded revert data from
+    /// non-reverting assertion paths into a stable user-facing message.
+    pub fn with_assertion_failure(mut self, assertion_failure: bool) -> Self {
+        self.assertion_failure = assertion_failure;
+        if assertion_failure && self.revert_reason == "<empty revert data>" {
```
grandizzy: let's make `<emtpy revert data>` constant here https://github.com/foundry-rs/foundry/blob/master/crates/evm/core/src/decode.rs#L122 and reuse

crates/evm/evm/src/executors/invariant/result.rs
```diff
@@ -56,6 +60,80 @@ impl<FEN: FoundryEvmNetwork> RichInvariantResults<FEN> {
...
+    }
+
+    is_assert_panic(call_result.result.as_ref())
+        || matches!(call_result.exit_reason, Some(InstructionResult::InvalidFEOpcode))
+        || is_vm_assert_revert(call_result.result.as_ref())
+        || is_cheatcode_assert_revert(call_result)
+}
+
+fn is_assert_panic(data: &[u8]) -> bool {
+    const PANIC_SELECTOR: [u8; 4] = [0x4e, 0x48, 0x7b, 0x71];
```
grandizzy: let's not declare these consts inline but reuse, e.g.
`alloy_sol_types::Panic::SELECTOR`, `alloy_sol_types::Revert::SELECTOR`, `Vm::CheatcodeError::SELECTOR`

crates/evm/evm/src/executors/invariant/result.rs
```diff
@@ -56,6 +60,80 @@ impl<FEN: FoundryEvmNetwork> RichInvariantResults<FEN> {
...
+        return false;
+    }
+
+    let panic_code = &data[4..36];
+    panic_code[..31].iter().all(|byte| *byte == 0) && panic_code[31] == 0x01
+}
+
+fn is_vm_assert_revert(data: &[u8]) -> bool {
+    const ERROR_SELECTOR: [u8; 4] = [0x08, 0xc3, 0x79, 0xa0];
+    data.starts_with(&ERROR_SELECTOR) && String::from_utf8_lossy(data).contains("assertion failed")
```
grandizzy: let's make this robust (handle ABI encoding changes)
```rust
fn is_vm_assert_revert(data: &[u8]) -> bool {
    alloy_sol_types::Revert::abi_decode(data)
        .is_ok_and(|r| r.reason.contains("assertion failed"))
}
```

crates/evm/evm/src/executors/invariant/shrink.rs
```diff
@@ -216,17 +222,29 @@ fn check_sequence_simple<FEN: FoundryEvmNetwork>(
     for call_index in sequence {
         let tx = &calls[call_index];
         let mut call_result = execute_tx(&mut executor, tx)?;
-        executor.commit(&mut call_result);
+        let should_fail_on_assert = options.fail_on_assert
+            && did_fail_on_assert(&executor, &call_result, &call_result.state_changeset);
+        if !options.fail_on_assert {
```
grandizzy: we can guard condition on if fail on assert only e.g. `if !options.fail_on_assert && did_fail_on_assert(&executor, &call_result, &call_result.state_changeset)`

crates/evm/evm/src/executors/invariant/shrink.rs
```diff
@@ -216,17 +222,29 @@ fn check_sequence_simple<FEN: FoundryEvmNetwork>(
...
-            // Candidate sequence fails test.
-            // We don't have to apply remaining calls to check sequence.
-            return Ok((false, false, call_failure_reason(call_result, options.rd)));
+        if call_result.result.as_ref() != MAGIC_ASSUME {
+            if should_fail_on_assert {
+                return Ok((false, false, assertion_failure_reason(call_result, options.rd)));
+            }
+
+            if call_result.reverted && options.fail_on_revert {
+                if options.expect_assertion_failure {
```
grandizzy, line 234: we can guard here `if options.expect_assertion_failure && !did_fail_on_assert` This way, if the shrunken sequence still triggers the same assertion failure, it correctly counts as reproducing the bug and gets kept (only unrelated invariant failures get skipped).

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -505,6 +508,20 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
                         break 'stop;
                     }
                 } else {
+                    let should_fail_on_assert = self.config.fail_on_assert
+                        && did_fail_on_assert(
+                            &current_run.executor,
+                            &call_result,
+                            &call_result.state_changeset,
+                        );
+                    if !self.config.fail_on_assert {
```
grandizzy: let's check also if any global failure here
```rust
if !self.config.fail_on_assert
    && current_run.executor.has_global_failure(&call_result.state_changeset)
{
```

crates/evm/evm/src/executors/invariant/shrink.rs
```diff
@@ -257,12 +275,26 @@ fn check_sequence_with_accumulation<FEN: FoundryEvmNetwork>(
 
         let tx_with_accumulated = apply_warp_roll(tx, accumulated_warp, accumulated_roll);
         let mut call_result = execute_tx(&mut executor, &tx_with_accumulated)?;
+        let should_fail_on_assert = options.fail_on_assert
+            && did_fail_on_assert(&executor, &call_result, &call_result.state_changeset);
+        if !options.fail_on_assert {
```
grandizzy: let's also check `executor.has_global_failure(&call_result.state_changeset)` here to avoid the `ignore_global_failure` if not needed

grandizzy (changes requested):
that looks good, ty, left some comments re impl. I wonder though if we could just drop the new config and report assertions always (this will match Echidna style too, and if you want to silent them you can switch to require with `fail_on_revert = false`)

| Echidna mode | Foundry config (proposed) | Reverts | Assertions |
| --- | --- | --- | --- |
| `testMode: property` | `fail_on_revert = false` | ignored | **reported** |
| `testMode: assertion` | `fail_on_revert = false` | ignored | **reported** |
| `testMode: dappfoundry` | `fail_on_revert = true` | reported | reported |

CC @aviggiano @0xalpharush @rappie pls chime in

rappie:
LGTM on a UX level

crates/evm/evm/src/executors/invariant/result.rs
```diff
@@ -56,6 +64,62 @@ impl<FEN: FoundryEvmNetwork> RichInvariantResults<FEN> {
...
+        || matches!(call_result.exit_reason, Some(InstructionResult::InvalidFEOpcode))
+        || is_revert_assertion_failure(call_result.result.as_ref())
+        || is_cheatcode_assert_revert(call_result)
+}
+
+fn is_assert_panic(data: &[u8]) -> bool {
+    Panic::abi_decode(data).is_ok_and(|panic| panic == PanicKind::Assert.into())
+}
+
+fn is_revert_assertion_failure(data: &[u8]) -> bool {
```
0xalpharush, line 87: What case is this for? Above it is guarded by
```
    if !call_result.reverted 
        return false;
    }
```

0xKarl98: That guard (e.g. if !call_result.reverted )  means:

  - if the call did not revert,
  - then is_assertion_failure(...) returns false immediately.
  
  And is_revert_assertion_failure handles the subset of reverted calls where the failure is encoded as Revert(string) rather than Panic(0x01).

crates/evm/evm/src/executors/invariant/error.rs
0xKarl98: Done

crates/evm/evm/src/executors/invariant/shrink.rs
```diff
@@ -284,6 +307,9 @@ fn finish_sequence_check<FEN: FoundryEvmNetwork>(
     let (invariant_result, mut success) =
         call_invariant_function(executor, test_address, calldata)?;
     if !success {
+        if options.expect_assertion_failure {
```
grandizzy: here we should only shrink / ignore the call if it expects assertion failure and if no failure, otherwise the call is discarded and will result in wrong call sequence. 
```rust
if options.expect_assertion_failure
        && !executor.has_global_failure(&invariant_result.state_changeset)
        && !did_fail_on_assert(&invariant_result, &invariant_result.state_changeset)
    {
```

crates/evm/evm/src/executors/invariant/shrink.rs
```diff
@@ -294,6 +320,9 @@ fn finish_sequence_check<FEN: FoundryEvmNetwork>(
             call_after_invariant_function(executor, test_address)?;
         success = after_invariant_success;
         if !success {
+            if options.expect_assertion_failure {
```
grandizzy: same as above, since this code is duped maybe we could extract in a closure / helper and reuse

grandizzy (dismissed):
2 more nits, then good to good imo. ty!

aviggiano:
Thanks

FYI I've tested this on https://github.com/scfuzzbench/scfuzzbench-canary and it correctly identifies 3 issues

grandizzy (approved):
thank you!

grandizzy:
@zerosnack should be good to approve / merge 🫡

mattsse (approved):
this lgtm!

PR #14276, opened 2026-04-13
fix(lint): add missing visit methods to LateLintVisitor

solanaXpeter:
## Motivation

`LateLintPass` exposes nested item/contract/function/var, modifier, and call-args hooks, but `LateLintVisitor` only dispatched a subset of them.

This left `check_modifier` and `check_call_args` unreachable, and the legacy nested-ID hooks no longer matched the current `solar::hir::Visit` API.

## Solution

Add the missing visitor dispatch for nested item/contract/function/var, modifier, and call args in `LateLintVisitor`.

Keep the original borrowed `check_nested_*` hooks as deprecated compatibility shims, and add `*_id` hooks that receive IDs by value so the late lint API stays compatible while matching the current `solar` visitor semantics.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mablr (commented):
sgtm, thanks

I just have potentially a concern about the old deprecated hooks being silently ignored.

A solution may be either:
- Dispatch both old and new hooks
- or drop simply the deprecated shims entirely ~~(breaking change, but clearer)~~

Pending another review

solanaXpeter:
@mablr Done, thanks for your review!

mablr (commented):
Hi @solanaXpeter, sorry my prev review was misleading, after looking a bit deeper, I believe the second solution is simpler. I think the hooks can be simply replaced with the corrected by-value signatures. No deprecation needed, smaller diff.

solanaXpeter:
@mablr Done. I replaced the nested hooks with the by-value signatures directly and trimmed the extra compatibility layer.

figtracer (commented):
can we add tests for this ?

solanaXpeter:
@figtracer Added test, please re-check, thanks!

mablr (approved):
sgtm

please @figtracer could you also recheck?

figtracer (approved):
think its good for now this is already an improvement

PR #14280, opened 2026-04-13
feat(anvil): support multiple fork URLs with round-robin load balancing

### LINKED ISSUE #14265: feat(anvil): support multiple fork URLs with load balancing
## Summary

When using `anvil --fork-url` with public or rate-limited RPC providers, heavy or complex workloads (e.g., automation scripts querying many distinct addresses/storage slots) frequently hit rate limits (HTTP 429), causing request timeouts and breaking downstream workflows.

**Proposal:** Allow `--fork-url` to accept multiple RPC endpoints and distribute requests across them using a load-balancing strategy (round-robin, least-latency, or weighted), with automatic failover when one endpoint returns 429 or times out.

## Motivation

- Existing mitigations (`--retries`, `--retry-backoff`, `--compute-units-per-second`) only slow down requests — they don't increase total throughput.
- `BlockchainDb` disk cache helps with repeated queries, but cold starts or workloads touching many unique addresses/slots still generate a high volume of upstream RPC calls.
- Users often have access to multiple RPC providers (e.g., Alchemy + QuickNode + public endpoints). Distributing load across them would effectively multiply the available rate limit budget.

## Proposed UX

```bash
# Multiple --fork-url flags
anvil --fork-url https://rpc1.example.com --fork-url https://rpc2.example.com --fork-url https://rpc3.example.com

# Or comma-separated
anvil --fork-url "https://rpc1.example.com,https://rpc2.example.com"
```

### foundry.toml support

```toml
[rpc_endpoints]
mainnet = { endpoints = ["https://rpc1.example.com", "https://rpc2.example.com"], retries = 5, retry_backoff = 1000 }
```

## Proposed Behavior

1. **Load balancing:** Round-robin by default across healthy endpoints.
2. **Failover:** If an endpoint returns 429 or times out, temporarily mark it unhealthy and route to the next available endpoint. Re-check after a cooldown period.
3. **Consistency:** All endpoints must be queried at the same block number (already enforced by fork block pinning).
4. **Per-endpoint config:** Each URL can optionally have its own `compute_units_per_second` / retry settings.

## Implementation Notes

The change is mostly localized to the provider layer:

- **CLI parsing** (`crates/anvil/src/cmd.rs`): Accept multiple `--fork-url` values.
- **Provider layer** (`crates/common/src/provider/`): Introduce a `LoadBalancedProvider` that wraps multiple `RetryProvider` instances and distributes requests.
- **Fork config** (`crates/anvil/src/config.rs`): Pass the provider list through `NodeConfig` → `ClientForkConfig`.
- **Transparent to upper layers:** `ClientFork` already uses `Arc<RetryProvider>` — replacing it with `Arc<LoadBalancedProvider>` (implementing the same provider trait) requires minimal changes upstream.

## Alternatives Considered

- **External load balancer / proxy (e.g., nginx, eRPC):** Works but adds operational complexity. A native solution is more ergonomic.
- **Higher `compute_units_per_second`:** Only delays the problem; doesn't increase actual throughput.
- **Paid unlimited RPC plans:** Expensive; multi-provider load balancing is more cost-effective.

decofe:
## Summary

Allow `--fork-url` to accept multiple RPC endpoints. Requests are distributed across endpoints using a custom round-robin `tower::Service` — each request goes to the next transport in rotation. On failure, the retry layer above handles retries (which will hit the next transport in the ring).

Closes https://github.com/foundry-rs/foundry/issues/14265

### CLI: multiple `--fork-url` flags
```bash
anvil --fork-url https://rpc1.example.com --fork-url https://rpc2.example.com --fork-url https://rpc3.example.com
```

### foundry.toml: `endpoints` array (backwards compatible)
```toml
[rpc_endpoints]
mainnet = { endpoints = ["https://rpc1.example.com", "https://rpc2.example.com"], retries = 5, retry_backoff = 1000 }
```

Then use the alias:
```bash
anvil --fork-url mainnet
```

Existing single-endpoint configs continue to work unchanged:
```toml
[rpc_endpoints]
mainnet = "https://rpc1.example.com"
mainnet_alt = { endpoint = "https://rpc2.example.com", retries = 5 }
```

## Changes

- `crates/config/src/endpoints.rs`: Added `extra_endpoints` to `RpcEndpoint`, support `endpoints` array in deser (backwards-compatible with existing `endpoint` key), `ResolvedRpcEndpoint::all_urls()` helper
- `crates/anvil/src/cmd.rs`: Changed `fork_url` from `Option<ForkUrl>` to `Vec<ForkUrl>`, alias resolution expands multi-endpoint configs, validates no block suffixes on secondary URLs
- `crates/anvil/src/config.rs`: Added `fork_urls: Vec<String>` to `NodeConfig`, branched `setup_fork_db_config` to build a round-robin provider when multiple URLs are present
- `crates/common/src/provider/mod.rs`: Added `RoundRobinService` (a `tower::Service` that rotates requests across multiple `RuntimeTransport`s) and `ProviderBuilder::build_fallback()` which wraps it with a retry layer
- `crates/anvil/src/eth/backend/fork.rs`: Added `fork_urls` to `ClientForkConfig`, updated `update_url` to rebuild the provider and keep URLs in sync

## Design

Uses a custom `RoundRobinService` that distributes requests sequentially across transports using `AtomicUsize` rotation. Each request goes to exactly one transport. On error, failover is handled by the `RetryBackoffLayer` above, which will naturally hit the next transport on retry since the index advances.

**Note:** There is no health scoring, endpoint quarantine, or active deprioritization of unhealthy endpoints. A flaky endpoint will continue to receive its share (~1/N) of requests. True health-based fallback (e.g. via Alloy's `FallbackService`) could be added in a follow-up.

Transparent to all upstream consumers — `ClientFork`, `SharedBackend`, `BackendHandler`, `ForkedDatabase` see the same `Arc<RetryProvider>` / `dyn Provider<N>`.

## Backwards Compatibility

- Single `--fork-url` works exactly the same
- `foundry.toml` single `endpoint` key unchanged
- `endpoints` (array) is a new optional key — cannot be combined with `endpoint`

## Known Limitations

- No health scoring or endpoint deprioritization — all endpoints receive equal traffic regardless of error rate or latency
- No startup validation that all endpoints return the same `eth_chainId` or can serve the fork block
- `anvil_setRpcUrl` resets to a single endpoint; multi-endpoint state is startup-only
- `RoundRobinService::poll_ready()` always returns `Ready` without polling inner transports

## Testing

```
cargo test -p foundry-config -- endpoints
cargo test -p anvil --lib cmd::tests
```

Prompted by: zerosnacks


crates/anvil/src/eth/backend/fork.rs
```diff
@@ -627,6 +627,8 @@ impl ClientFork {
 #[derive(Clone, Debug)]
 pub struct ClientForkConfig<N: Network = AnyNetwork> {
     pub eth_rpc_url: String,
+    /// All fork URLs when using multi-endpoint load balancing.
+    pub fork_urls: Vec<String>,
```
mablr: `eth_rpc_url` and `fork_urls[0]` seem duplicate. Is this to prevent breaking change, or could we simply remove `eth_rpc_url` and assume `fork_urls.len() > 0` is always true ?

figtracer: imo it's cleaner to keep them separate, if we collapsed it, it would add a bit of indirection

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -2078,6 +2078,7 @@ impl<N: Network> Backend<N> {
                     // we want to force the correct base fee for the next block during
                     // `setup_fork_db_config`
                     node_config.base_fee.take();
+                    node_config.fork_urls = vec![eth_rpc_url.clone()];
```
stevencartavia, line 2082: `setup_fork_db_config` clones `self.fork_urls` into `ClientForkConfig`, but in the non-fork-to-fork reset path `node_config.fork_urls` is still the default empty vec. This panics on `fork_urls[0]` in `eth_rpc_url()`.

crates/anvil/src/eth/backend/fork.rs
```diff
@@ -659,25 +662,40 @@ pub struct ClientForkConfig<N: Network = AnyNetwork> {
...
+    /// Updates the provider URLs
     ///
     /// # Errors
     ///
     /// This will fail if no new provider could be established (erroneous URL)
-    fn update_url(&mut self, url: String) -> Result<(), BlockchainError> {
-        // let interval = self.provider.get_interval();
-        self.provider = Arc::new(
-            ProviderBuilder::<N>::new(url.as_str())
+    fn update_urls(&mut self, urls: Vec<String>) -> Result<(), BlockchainError> {
```
mablr, line 677: It seems the provider is constructed without `fork_headers`. After an `anvil_reset` RPC call with new URLs, any `--fork-header` values are silently lost.

crates/anvil/src/cmd.rs
```diff
@@ -630,13 +646,35 @@ pub struct AnvilEvmArgs {
...
+                let mut endpoints = config.rpc_endpoints.clone().resolved();
+                if let Some(endpoint) = endpoints.remove(&fork_url.url) {
+                    // Alias matched — expand all URLs from the endpoint config
+                    if let Ok(urls) = endpoint.all_urls() {
+                        for (i, url) in urls.into_iter().enumerate() {
+                            resolved_urls.push(ForkUrl {
+                                url,
+                                // Only the first URL inherits the block suffix
+                                block: if i == 0 { fork_url.block } else { None },
+                            });
```
mablr: If an env-var in `extra_endpoints` is unset, the alias expansion is silently skipped ?

crates/anvil/src/eth/backend/fork.rs
```diff
@@ -659,25 +662,40 @@ pub struct ClientForkConfig<N: Network = AnyNetwork> {
 }
 
 impl<N: Network> ClientForkConfig<N> {
-    /// Updates the provider URL
+    /// Returns the primary RPC URL (first entry in `fork_urls`).
+    pub fn eth_rpc_url(&self) -> &str {
+        &self.fork_urls[0]
+    }
```
mablr, line 670: Even though logically it shouldn't happen as first index is always set with `eth_rpc_url`, there is a risk of panic here. We maybe could make this more elegant/robust?

crates/anvil/src/config.rs
```diff
@@ -1248,16 +1266,26 @@ impl NodeConfig {
...
+                .headers(self.fork_headers.clone())
+                .build_fallback(self.fork_urls.clone())
+                .wrap_err("failed to establish fallback provider to fork urls")?
+        } else {
             ProviderBuilder::new(&eth_rpc_url)
                 .timeout(self.fork_request_timeout)
                 .initial_backoff(self.fork_retry_backoff.as_millis() as u64)
                 .compute_units_per_second(self.compute_units_per_second)
                 .max_retry(self.fork_request_retries)
                 .headers(self.fork_headers.clone())
```
mablr: dedup a bit ?

mablr (commented):
Some comments, looks good overall.

mablr (dismissed):
sgtm

zerosnacks (dismissed):
Made some small bug fix changes here: https://github.com/foundry-rs/foundry/pull/14280/commits/8cd02a54449a31f6e14f92e83445f7064c09a01e and here: https://github.com/foundry-rs/foundry/pull/14393

PR lgtm

zerosnacks (approved):

stevencartavia (approved):

PR #14307, opened 2026-04-14
fix(anvil): handle zero prune history safely

mk0walsk:
## Motivation

`--prune-history 0` currently reaches the in-memory history cache as `0`. `InMemoryBlockStates::enforce_limits()` then loops on `while self.present.len() >= self.in_memory_limit`, which can hang block production when the limit is zero.

## Solution

Normalize `max_memory_history == 0` to the unsupported path in `PruneStateHistoryConfig`, and defensively clamp in-memory cache limits to at least `1` inside `InMemoryBlockStates`. Add regression tests for the config normalization and zero-limit cache guard.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/anvil/src/eth/backend/mem/storage.rs
```diff
@@ -198,7 +199,10 @@ impl InMemoryBlockStates {
 
     /// Sets the maximum number of stats we keep in memory
     pub const fn set_cache_limit(&mut self, limit: usize) {
+        let limit = if limit == 0 { 1 } else { limit };
```
mablr, line 202:
```suggestion
        let limit = limit.max(1);
```
like above

mablr (commented):
sgtm

mk0walsk:
@mablr Done, please take another look, thanks!

crates/anvil/src/eth/backend/mem/storage.rs
```diff
@@ -70,8 +70,13 @@ pub struct InMemoryBlockStates {
 }
 
 impl InMemoryBlockStates {
+    const fn normalize_in_memory_limit(limit: usize) -> usize {
+        if limit == 0 { 1 } else { limit }
+    }
```
mablr: this can be free function, or inline btw

mablr (commented):
smol pedantic nit :)

otherwise it's good.

crates/anvil/src/eth/backend/mem/storage.rs
mablr: initial `.max(1)` was good 👍

mk0walsk: Done, please re-check. The helper is gone and the clamp is now inlined.

mablr (dismissed):
sgtm, pending another review ...

crates/anvil/src/eth/backend/mem/storage.rs
```diff
@@ -203,7 +199,7 @@ impl InMemoryBlockStates {
 
     /// Sets the maximum number of stats we keep in memory
     pub const fn set_cache_limit(&mut self, limit: usize) {
-        let limit = Self::normalize_in_memory_limit(limit);
+        let limit = if limit == 0 { 1 } else { limit };
```
figtracer, line 202: can't we use max(1) here too?

mk0walsk: Sure, updated to use `max(1)` there too. Let me know if anything still looks off.Sure, updated to use `max(1)` there too. Let me know if anything still looks off.

figtracer (approved):

mablr (approved):

PR #14311, opened 2026-04-14
fix(cast): `keychain auth` legacy mode w/ forks before T3

mablr:
## Motivation

`keychain auth` legacy mode w/ forks before T3

Tested against https://github.com/foundry-rs/foundry/pull/14206:

```
TEMPO_HARDFORK="T2" PATH="$(pwd)/target/debug:$PATH" ETH_RPC_URL="https://rpc.moderato.tempo.xyz" ./.github/scripts/tempo-check.sh
```

### DIFF
--- a/crates/cast/src/cmd/keychain.rs
+++ b/crates/cast/src/cmd/keychain.rs
@@ -17,13 +17,15 @@ use foundry_common::{
     shell,
     tempo::{self, KeyType, KeysFile, WalletType, read_tempo_keys_file, tempo_keys_path},
 };
-use tempo_alloy::{TempoNetwork, provider::TempoProviderExt};
+use foundry_evm::hardfork::TempoHardfork;
+use tempo_alloy::{TempoNetwork, provider::TempoProviderExt, rpc::TempoTransactionRequest};
 use tempo_contracts::precompiles::{
     ACCOUNT_KEYCHAIN_ADDRESS, IAccountKeychain,
     IAccountKeychain::{
-        CallScope, KeyInfo, KeyRestrictions, SelectorRule, SignatureType, TokenLimit,
+        CallScope, KeyInfo, KeyRestrictions, LegacyTokenLimit, SelectorRule, SignatureType,
+        TokenLimit,
     },
-    account_keychain::authorizeKeyCall,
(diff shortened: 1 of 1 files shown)


zerosnacks (dismissed):
Makes sense 👍

crates/cast/src/cmd/keychain.rs
```diff
@@ -660,6 +695,20 @@ async fn run_remove_scope(
     send_keychain_tx(calldata, tx_opts, &send_tx).await
 }
 
+/// Returns `true` when the connected Tempo chain has the T3 hardfork active.
+async fn detect_t3<P: Provider<TempoNetwork>>(provider: &P) -> bool {
```
0xrusowsky: nit: can we make it generic over `TempoHardfork` (passed as input) and rename to `detect_hardfork`, `is_harfork_active`, or smth similar?

crates/cast/src/cmd/keychain.rs
```diff
@@ -607,8 +627,23 @@ async fn run_remaining_limit(
...
+        let calldata = IAccountKeychain::getRemainingLimitCall {
+            account: wallet_address,
+            keyId: key_address,
+            token,
+        }
+        .abi_encode();
+        let mut tx = TempoTransactionRequest::default();
+        tx.inner = tx.inner.to(ACCOUNT_KEYCHAIN_ADDRESS).input(calldata.into());
+        let result = provider.call(tx).await?;
+        U256::from_be_slice(&result)
```
0xrusowsky: i think here you don't need to manually build a `TempoTransactionRequest` and do a raw `provider.call()` because `getRemainingLimit` (the legacy endpoint) is still exposed through `IAccountKeychainInstance`.

if i'm not wrong, you should be able to:
```rs
provider.account_keychain().getRemainingLimit(wallet_address, key_address, token).call().await?
```

crates/cast/src/cmd/keychain.rs
mablr: okay i'll do it

0xrusowsky (approved):

zerosnacks (approved):

PR #14317, opened 2026-04-15
fix(anvil): sync state snapshots with reset and revert

solanaXpeter:
## Motivation

State snapshot metadata can diverge from backend state. `anvil_reset` clears chain and DB state without clearing `active_state_snapshots`, and `revert_state_snapshot` removed metadata before the DB revert succeeded. That can leave `anvil_metadata` reporting stale snapshots and make `evm_revert` depend on a partially applied rollback path.

## Solution

Clear `active_state_snapshots` in both reset paths, and only remove a snapshot entry after the DB revert succeeds. When the backing snapshot block or DB snapshot is already gone, drop the stale metadata entry and return `false` without mutating chain state. Added regression tests for failed revert, in-memory reset, and fork reset paths.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -683,6 +683,10 @@ impl<N: Network> Backend<N> {
         self.active_state_snapshots.lock().clone().into_iter().collect()
     }
 
+    fn clear_state_snapshots(&self) {
+        self.active_state_snapshots.lock().clear();
+    }
+
     /// Returns the environment for the next block
     fn next_evm_env(&self) -> EvmEnv {
         let mut evm_env = self.evm_env.read().clone();
@@ -2160,6 +2164,7 @@ impl<N: Network> Backend<N> {
                 fork.total_difficulty(),
             );
             self.states.write().clear();
+            self.clear_state_snapshots();
             self.db.write().await.clear();
 
             self.apply_genesis().await?;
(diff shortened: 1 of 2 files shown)


crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -2243,49 +2249,60 @@ impl<N: Network> Backend<N> {
...
-            let block =
-                self.block_by_hash(best_block_hash).await?.ok_or(BlockchainError::BlockNotFound)?;
+        let Some(block) = self.block_by_hash(hash).await? else {
+            self.active_state_snapshots.lock().remove(&id);
+            return Ok(false);
+        };
 
-            let reset_time = block.header.timestamp();
-            self.time.reset(reset_time);
+        if !self.db.write().await.revert_state(id, RevertStateSnapshotAction::RevertRemove) {
```
mablr, line 2261: Would we rather need to `revert_state` any case ? (even if `self.block_by_hash(hash).await?` is `None`)

solanaXpeter: @mablr Good point, I agree this branch still needs handling. I think we should drop the DB snapshot and return `false` when `block_by_hash(hash)` is `None`, since calling `revert_state` there would leave things half-reverted. Do you think that makes sense?

figtracer (commented):
i think we don't need this now tbf, pending another review

mablr (commented):
Yes I think it's not necessary to apply these changes, as Anvil's internals are going to be completely refactored soon.

 Thanks for the effort @solanaXpeter 👍

PR #14319, opened 2026-04-15
fix(config): respect custom Etherscan URL in cast/forge commands

FredPhilipy:
Several commands (`cast storage`, `cast creation-code`, `cast interface`, `forge clone`) were bypassing the user's custom Etherscan URL configuration in `foundry.toml` by using `Client::new(chain, api_key)` directly. Changed these to use `get_etherscan_config_with_chain()` + `into_client()` pattern, consistent with the fix in verify module. This enables these commands to work with private chains and custom block explorers.

### DIFF
--- a/crates/common/src/abi.rs
+++ b/crates/common/src/abi.rs
@@ -146,11 +146,7 @@ pub async fn get_func_etherscan(
     etherscan_api_url: Option<&str>,
 ) -> Result<Function> {
     let client = if let Some(api_url) = etherscan_api_url {
-        Client::builder()
-            .with_api_key(etherscan_api_key)
-            .with_api_url(api_url)?
-            .with_url(api_url)?
-            .build()?
+        Client::builder().with_api_key(etherscan_api_key).with_api_url(api_url)?.build()?
     } else {
         Client::new(chain, etherscan_api_key)?
     };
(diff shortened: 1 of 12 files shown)


crates/common/src/abi.rs
```diff
@@ -143,8 +143,17 @@ pub async fn get_func_etherscan(
...
     chain: Chain,
     etherscan_api_key: &str,
+    etherscan_api_url: Option<&str>,
 ) -> Result<Function> {
-    let client = Client::new(chain, etherscan_api_key)?;
+    let client = if let Some(api_url) = etherscan_api_url {
+        Client::builder()
+            .with_api_key(etherscan_api_key)
+            .with_api_url(api_url)?
+            .with_url(api_url)?
```
mablr: Setting both `etherscan_url` and `etherscan_api_url` w/ the same value isn't correct imo.

mablr (changes requested):
Thanks for your PR, just a smol nit.

Clippy+fmt :)

FredPhilipy:
> Thanks for your PR, just a smol nit.
> 
> Clippy+fmt :)

Thanks! fixed.

mablr (dismissed):
sgtm 👍 pending another review.

(you just need to remove `foundry-block-explorers` dep in `foundry-cli` to make clippy happy)

FredPhilipy:
> sgtm 👍 pending another review.
> 
> (you just need to remove `foundry-block-explorers` dep in `foundry-cli` to make clippy happy)

Yeah just realized it, fixed already.

mablr (approved):

mablr (approved):

figtracer (approved):

PR #14336, opened 2026-04-15
fix(script): skip `deploy_create2_deployer` in tempo mode to avoid `CreateCollision`

mablr:
## Motivation

Tempo genesis already deploys the Arachnid CREATE2 factory at the same address as the default CREATE2 deployer (0x4e59b448...). The existing code check in deploy_create2_deployer runs before the EVM is created, so it doesn't see the genesis-initialized bytecode, causing a CreateCollision when the CREATE transaction executes.

Skip the call entirely in non-fork Tempo mode since genesis handles it.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


zerosnacks (approved):
Nice! Yes this makes sense to me

stevencartavia (approved):

PR #14337, opened 2026-04-15
feat(config): `network` key in `foundry.toml`

mablr:
## Motivation

Write tempo network variant to config when initializing a Foundry project:
```
forge init -n tempo mail
```

So that subsequent test/script commands can automatically select the right network/evm without requiring `--tempo` flag.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes (i believe no, as network field is optional)

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -1920,6 +1920,7 @@ impl Config {
             out: self.out,
             libs: self.libs,
             remappings: self.remappings,
+            network: self.networks.active_network_name().map(String::from),
         }
     }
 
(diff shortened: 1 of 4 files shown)


crates/config/src/lib.rs
```diff
@@ -2766,14 +2767,20 @@ pub struct BasicConfig {
...
 impl BasicConfig {
     /// Serialize the config as a String of TOML.
     ///
     /// This serializes to a table with the name of the profile
     pub fn to_string_pretty(&self) -> Result<String, toml::ser::Error> {
-        let s = toml::to_string_pretty(self)?;
+        let mut s = toml::to_string_pretty(self)?;
+        if let Some(ref network) = self.network {
+            s.push_str(&format!("{network} = true\n"));
+        }
```
zerosnacks: Rather than serializing first and pushing a raw string can't we set the network on self and then serialize it? That would be cleaner

0xrusowsky (dismissed):
makes sense to me, i think it's acceptable

crates/config/src/lib.rs
mablr: done 👍

zerosnacks (approved):
👍!

PR #14340, opened 2026-04-15
chore: bump revm

klkvr:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cast/src/cmd/run.rs
+++ b/crates/cast/src/cmd/run.rs
@@ -207,7 +207,7 @@ impl RunArgs {
                     evm_env.cfg_env.chain_id,
                     block.header().timestamp(),
                 ) {
-                    evm_env.cfg_env.set_spec(hardfork.into());
+                    evm_env.cfg_env.set_spec_and_mainnet_gas_params(hardfork.into());
                 } else if block.header().excess_blob_gas().is_some() {
                     // TODO: add glamsterdam header field checks in the future
                     evm_version = Some(EvmVersion::Cancun);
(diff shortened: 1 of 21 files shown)


crates/cast/src/cmd/run.rs
```diff
@@ -207,7 +207,7 @@ impl RunArgs {
                     evm_env.cfg_env.chain_id,
                     block.header().timestamp(),
                 ) {
-                    evm_env.cfg_env.set_spec(hardfork.into());
+                    evm_env.cfg_env.set_spec_and_mainnet_gas_params(hardfork.into());
```
klkvr, line 210: not 100% sure about that one, it would override the gas params inside of cfg. will it work fine for tempo?

klkvr: looks like we do this in a bunch of other places so should be ok

mattsse: yeah we should no longer use set_spec

crates/evm/core/src/evm/mod.rs
```diff
@@ -258,24 +258,27 @@ pub fn with_cloned_context<CTX: FoundryContextExt>(
...
+    journal: &mut T,
+) -> Result<CallInputs, <T::Database as Database>::Error> {
     let calldata = [&salt.to_be_bytes::<32>()[..], &inputs.init_code()[..]].concat();
-    CallInputs {
+    let account = journal.load_account_with_code(deployer)?;
+    Ok(CallInputs {
         caller: inputs.caller(),
         bytecode_address: deployer,
-        known_bytecode: None,
+        known_bytecode: (account.info.code_hash, account.info.code.clone().unwrap_or_default()),
```
mattsse, line 272: is this a driveby fix?

mattsse (approved):

crates/evm/core/src/evm/mod.rs
klkvr: this is no longer optional so we need to provide the inputs on all callsites

PR #14342, opened 2026-04-16
perf(evm): avoid hot-path clones in fuzz and replay

solanaXpeter:
## Motivation

Invariant replay and fuzz execution still clone logs and line coverage in hot paths even when the source value is immediately consumed or discarded afterward. That keeps avoidable allocations on per-call paths and on the last-run log handoff.

## Solution

- Move replay logs and line coverage out of `RawCallResult` with `std::mem::take` / `Option::take` where the value is consumed once.
- Move invariant and after-invariant replay logs and traces directly into the collected output.
- Reuse owned logs in fuzz counterexample handling and last-run aggregation instead of cloning them.
- Keep the trace clones that still have two real owners (`result.traces` and counterexample/replay outputs), so the change stays ownership-only and behavior-preserving.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/evm/src/executors/fuzz/mod.rs
+++ b/crates/evm/evm/src/executors/fuzz/mod.rs
@@ -381,9 +381,8 @@ impl<FEN: FoundryEvmNetwork> FuzzedExecutor<FEN> {
             result.breakpoints = last_run_worker.breakpoints.clone();
         }
 
-        if !self.config.show_logs {
-            result.logs = workers[last_run_worker_idx].logs.clone();
-        }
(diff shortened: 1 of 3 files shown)


mattsse (commented):
unsure if all of these are side-effect free

likely not super important so I'd rather keep as is

zerosnacks:
Agreed on closing, the `take` replaces the value with `None` so whilst this exact version works as described now it makes maintenance / extending the logic much more brittle

PR #14344, opened 2026-04-16
refactor(evm): CREATE2 factory redirect via Inspector frame hooks

mablr:
## Motivation

Leverage revm 37.0.0's `Inspector::{frame_start, frame_end}` hooks to handle CREATE2-to-factory-call rewriting directly in `InspectorStack`, removing the need for `EthFoundryHandler`, `OpFoundryHandler`, and `TempoFoundryHandler`.

- Add `frame_start`/`frame_end` to `InspectorStackRefMut` with CREATE2 redirect logic (rewrite `FrameInput::Create` -> `Call`, transform result back)
- Use upstream `MainnetHandler`, `OpHandler`, `TempoEvmHandler` directly
- Remove ~600 lines of duplicated handler wrapper code

The `alloy_evm::Evm` impls become now dummy, so they will be removed in a follow-up PR to keep the current one contained.

### DIFF
--- a/crates/evm/core/src/evm/op.rs
+++ b/crates/evm/core/src/evm/op.rs
@@ -1,6 +1,6 @@
 use super::*;
 
-type OpUpstreamHandler<'db, I> = OpHandler<
+type OpEvmHandler<'db, I> = OpHandler<
     OpRevmEvm<'db, I>,
     EVMError<DatabaseError, OpTransactionError>,
     EthFrame<EthInterpreter>,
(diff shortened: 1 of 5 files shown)


figtracer (dismissed):

crates/evm/core/src/evm/op.rs
```diff
@@ -1,5 +1,11 @@
 use super::*;
 
+type OpUpstreamHandler<'db, I> = OpHandler<
```
zerosnacks: Naming nit, shall we use `OpEvmHandler`?

It being upstream is an implementation detail imo

mablr: It's going to be removed in follow-up removing custom Evm wrappers/impls

mablr: done👍

zerosnacks (approved):

figtracer (approved):

PR #14347, opened 2026-04-16
fix(mpp): fetch fresh challenge on verification-failed key provisioning retry

zerosnacks:
## Summary

Follow up fix for #14192

When an access key is refreshed (`tempo wallet refresh`) but not yet provisioned on-chain, the first payment strips `key_authorization` (assuming already provisioned). The server rejects with `verification-failed`.

Previously, the retry reused the **same consumed challenge ID**, which the server rejected again. Now we fetch a fresh 402 challenge before retrying with `key_authorization` included. The same fix is applied to `key-not-provisioned` retries.

Also extracts `pay_and_retry`, `fetch_fresh_challenge`, and `select_challenge` helpers to deduplicate the retry paths (-90 lines net).

## Tested

- Fresh key + Tempo endpoint: verification-failed → fresh challenge → key provisioned → 200 OK ✅
- Without fix: verification-failed → stale challenge reuse → failure ❌

### DIFF
--- a/crates/common/src/provider/mpp/transport.rs
+++ b/crates/common/src/provider/mpp/transport.rs
@@ -228,59 +228,11 @@ where
         // held until the retry response is fully handled.
         let _pay_guard = self.provider.lock_pay().await;
 
-        let www_auth_values: Vec<&str> = resp
-            .headers()
-            .get_all(WWW_AUTHENTICATE_HEADER)
-            .iter()
-            .filter_map(|v| v.to_str().ok())
-            .collect();
-
-        if www_auth_values.is_empty() {
-            return Err(TransportErrorKind::custom(std::io::Error::other(
-                "402 response missing WWW-Authenticate header",
-            )));
-        }
-
-        let challenges: Vec<_> = parse_www_authenticate_all(www_auth_values)
-            .into_iter()
-            .filter_map(|r| r.ok())
-            .collect();
-
-        // Try each challenge until we find one with a matching key (chain + currency)
-        // in keys.toml. This handles servers that offer multiple chains and currencies
-        // (e.g. mainnet + testnet) — we pick the first one the user has a key for.
-        let mut last_resolve_err: Option<TransportError> = None;
-        let resolved_pair = challenges.iter().find_map(|c| {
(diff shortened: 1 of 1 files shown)


grandizzy (dismissed):
makes sense, should we worry about `key-not-provisioned` become stale as well? (we could follow-up if we hit this)

zerosnacks:
> makes sense, should we worry about `key-not-provisioned` become stale as well? (we could follow-up if we hit this)

yep good call, updated PR @grandizzy, also made it more DRY

crates/common/src/provider/mpp/transport.rs
```diff
@@ -548,6 +383,122 @@ where
...
+    ) -> TransportResult<(P::Provider, mpp::protocol::core::PaymentChallenge)> {
+        let fresh_resp = self
+            .client
+            .post(self.url.clone())
+            .headers(headers.clone())
+            .header("content-type", "application/json")
+            .body(body.to_vec())
+            .send()
+            .await
+            .map_err(TransportErrorKind::custom)?;
```
mablr, line 440: `MPP_RETRY_TIMEOUT` is missing here ?

crates/common/src/provider/mpp/transport.rs
```diff
@@ -548,6 +383,122 @@ where
...
+    ) -> TransportResult<reqwest::Response> {
+        let credential = provider.pay(challenge).await.map_err(|e| {
+            TransportErrorKind::custom(std::io::Error::other(format!("MPP payment failed: {e}")))
+        })?;
+
+        let auth_header = format_authorization(&credential).map_err(|e| {
+            TransportErrorKind::custom(std::io::Error::other(format!(
+                "failed to format MPP credential: {e}"
+            )))
+        })?;
```
mablr, line 404: Should we map errors with `self.provider.rollback_pending();` here like below?

crates/common/src/provider/mpp/transport.rs
zerosnacks: yep, good find!

figtracer (approved):

mablr (approved):

PR #14349, opened 2026-04-16
fix(anvil): avoid panic on non-standard tx conversion

apetro2:
## Motivation
                                                                                                                                                                                      
  `transaction_build` used `try_into_eth().expect(...)` for the fallback path.                                                                                                        
  This can panic when a non-standard transaction reaches this branch (for example through future tx type additions or unexpected input), which creates a runtime crash risk in Anvil RPC transaction serialization.                                                                                                                                                      
                                                                                                                                                                                      
  ## Solution
                                                                                                                                                                                      
  Replace the `expect` with an explicit `match` on `try_into_eth()`:                                                                                                                  
                                                                                                                                                                                      
  - Keep existing behavior for standard Ethereum transaction envelopes.
  - For non-standard envelopes, avoid panic and return an `Unknown` envelope fallback.                                                                                                
  - Emit an error log with `tx_type` to preserve observability.                                                                                                                       

  This keeps normal behavior unchanged while removing a crash path.                                                                                                                   
                                                                                                                                                                                      
  ## PR Checklist                                                                                                                                                                     
                                                                                                                                                                                      
  - [ ] Added Tests                                                                                                                                                                   
  - [ ] Added Documentation
  - [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -4455,16 +4455,30 @@ pub fn transaction_build(
     // there's no `info` yet.
     let hash = tx_hash.unwrap_or_else(|| eth_transaction.hash());
 
-    let eth_envelope = FoundryTxEnvelope::from(eth_transaction)
-        .try_into_eth()
-        .expect("non-standard transactions are handled above");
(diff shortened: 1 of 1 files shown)


mattsse (commented):
we can keep this as is

7zkm7b8gw9-web (commented):
?

PR #14354, opened 2026-04-16
fix(config): validate `optimizer_runs` does not exceed u32::MAX

FredPhilipy:
Added validation in config loading to reject `optimizer_runs` values exceeding `u32::MAX`. The Solidity compiler expects this parameter as u32, and the doc comment at line 271 already documents this limit. Now the config fails early with a clear error instead of causing undefined behavior at compile time.


figtracer (approved):
doesn't hurt to have this check i think

mablr (approved):

zerosnacks (approved):
Reasonable: https://docs.soliditylang.org/en/latest/internals/optimizer.html#optimizer-parameter-runs

Thanks!

PR #14355, opened 2026-04-16
feat: use ChannelDb for channel persistence

stevencartavia:
- Replace JSON file persistence (`~/.tempo/foundry/channels.json`) with SQLite via `foundry_wallets::ChannelDb` (`~/.tempo/channels.db`)
- Bump `foundry-wallets` rev to `e8c8be1` (includes [foundry-core#17](https://github.com/foundry-rs/foundry-core/pull/17))
- Update CI script to query SQLite instead of grepping JSON


crates/common/src/provider/mpp/persist.rs
```diff
@@ -1,168 +1,184 @@
...
     time::{SystemTime, UNIX_EPOCH},
 };
 use tracing::{debug, warn};
 
 use crate::tempo::tempo_home;
 
-/// Relative path from Tempo home to the Foundry channels file.
-const CHANNELS_PATH: &str = "foundry/channels.json";
+/// Re-export as `PersistedChannel` for backward compatibility.
+pub type PersistedChannel = Channel;
```
mablr: We may simply rename all `PersistedChannel` occurences ?

crates/common/src/provider/mpp/persist.rs
```diff
@@ -171,73 +187,61 @@ pub fn find_channel(
...
 
     if let Some(existing) = channels.get_mut(key) {
         existing.cumulative_amount = entry.cumulative_amount.to_string();
         existing.last_used_at = now;
-        existing.status = if entry.opened { "active" } else { "closed" }.to_string();
+        existing.state = if entry.opened { "active" } else { "closed" }.to_string();
     } else {
-        channels
-            .insert(key.to_string(), PersistedChannel::from_channel_entry(entry, deposit, origin));
+        warn!(key, "upsert_channel_in_memory called for unknown channel");
```
mablr, line 200: May this have side-effect, do we always match existing channel ?

stevencartavia: The else branch is unreachable in practice, `upsert_channel_in_memory()` is only called from the voucher path `(session.rs:689)` which always operates on an already-open channel. The `warn!` makes it defensive instead of silently inserting incomplete data.

mablr: okay nice, ty for clarification

mablr (approved):
sgtm

zerosnacks (approved):

PR #14358, opened 2026-04-17
fix(config): report cache cleanup deletion failures

solanaXpeter:
## Motivation

Cache cleanup paths currently ignore all deletion errors.
That keeps missing paths best-effort, but it also hides permission and path-type failures while leaving stale cache state behind.

## Solution

Ignore only `io::ErrorKind::NotFound` for cache cleanup removals and return all other deletion errors with path context from `cleanup` and the `clean_foundry_*` helpers.
Add focused unit tests for missing-path and wrong-path-type cases.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -45,7 +45,7 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer, de};
 use std::{
     borrow::Cow,
     collections::BTreeMap,
-    fs,
+    fs, io,
     path::{Path, PathBuf},
     str::FromStr,
 };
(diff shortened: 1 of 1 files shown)


mablr (approved):
lgtm, thanks!

figtracer (approved):

decofe (changes requested):
Cache cleanup is best-effort by design — this PR makes `forge clean` strictly worse by short-circuiting on the first deletion error (remaining dirs never get cleaned).

Instead of propagating errors, log them and continue. Since `sh_warn!` is a circular dependency in the `config` crate, use the same `eprintln!` + `yansi::Paint::yellow` pattern already used elsewhere in this file (see L1213):

```rust
// Instead of helper functions + Result propagation, just warn and continue:
if let Err(err) = fs::remove_file(&self.test_failures_file) {
    if err.kind() != io::ErrorKind::NotFound {
        eprintln!(
            "{}",
            yansi::Paint::yellow(&format!(
                "Warning: failed to remove test failures file {}: {err}",
                self.test_failures_file.display()
            ))
        );
    }
}
```

Same pattern for every `remove_dir_all` call. This way all cleanup steps always run, failures are visible to the user, and `forge clean` never aborts halfway through.

The helper functions (`remove_file_if_exists`, `remove_dir_all_if_exists`) and their tests can be dropped — they're not needed with this approach.

zerosnacks:
Hi @solanaXpeter, thanks for your PR!

I've decided to implement an alternative approach here: https://github.com/foundry-rs/foundry/pull/14379 that I prefer and added you as a co-author

PR #14363, opened 2026-04-17
fix(evm): preserve CREATE2 redirect state across isolated transactions

### LINKED ISSUE #14362: bug: test-isolate workflow failed
The daily test-isolate workflow (tests with isolation mode enabled) has failed. This indicates a regression in Foundry's test isolation behavior.

Check the [test-isolate workflow page](https://github.com/foundry-rs/foundry/actions/runs/24553757747
) for details.

This issue was raised by the workflow at `.github/workflows/test-isolate.yml`.

decofe:
## Summary

Fixes [#14362](https://github.com/foundry-rs/foundry/issues/14362). Follow-up to [#14360](https://github.com/foundry-rs/foundry/pull/14360) which skipped isolation for CREATE2 factory redirects but broke broadcast nonce accounting (`nonce too low` errors).

## Root Cause

`with_inspector` does `mem::take` on the inner inspector state, moving `pending_create2_redirects` into the nested EVM. The nested `frame_end` consumes the redirect entry, so when control returns, the outer `frame_end` skips the `Call → Create` conversion.

The previous fix skipped isolation entirely for these calls, but the cheatcodes inspector relies on `transact_inner` to manage broadcast nonces when isolation is enabled.

## Fix

Save and restore `pending_create2_redirects` around `with_inspector`. Reverts the skip-isolation approach from #14360.

**Net diff vs pre-#14360 master**: +5 lines in `with_inspector`, no changes to the `call` hook.

Prompted by: zerosnacks

### DIFF
--- a/crates/evm/evm/src/inspectors/stack.rs
+++ b/crates/evm/evm/src/inspectors/stack.rs
@@ -898,12 +898,17 @@ impl<FEN: FoundryEvmNetwork> InspectorStackRefMut<'_, FEN> {
             .map(|cheats| core::mem::replace(cheats, Cheatcodes::new(cheats.config.clone())));
         let mut inner = std::mem::take(self.inner);
 
+        // Save pending CREATE2 redirects so frame_end in the nested EVM doesn't consume them.
+        // These belong to the outer EVM's frame lifecycle and must be restored after.
+        let saved_create2_redirects = std::mem::take(&mut inner.pending_create2_redirects);
+
         let out = f(InspectorStackRefMut { cheatcodes: cheatcodes.as_mut(), inner: &mut inner });
 
         if let Some(cheats) = self.cheatcodes.as_deref_mut() {
             *cheats = cheatcodes.unwrap();
         }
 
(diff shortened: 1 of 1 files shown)


crates/evm/evm/src/inspectors/stack.rs
```diff
@@ -898,12 +898,17 @@ impl<FEN: FoundryEvmNetwork> InspectorStackRefMut<'_, FEN> {
             .map(|cheats| core::mem::replace(cheats, Cheatcodes::new(cheats.config.clone())));
         let mut inner = std::mem::take(self.inner);
 
+        // Save pending CREATE2 redirects so frame_end in the nested EVM doesn't consume them.
+        // These belong to the outer EVM's frame lifecycle and must be restored after.
+        let saved_create2_redirects = std::mem::take(&mut inner.pending_create2_redirects);
```
mablr, line 903: Good catch I believe.

mablr (approved):

figtracer (approved):

PR #745, opened 2022-02-15
fix(forge): expectEmit with call

ualtinok:
fixes https://github.com/gakonst/foundry/issues/668

Notice emitted a and b values below:

Before:
```
# [PASS] testExpectEmit() (gas: 33730)
Traces:

  [146057] ContractTest::setUp()
    ├─ → new Contract@0xce71…c246
    │   └─ ← 552 bytes of code
    ├─ [0] VM::deal(0xce71065d4017f316ec606fe4422e11eb2c47c246, 100000000000000000000)
    │   └─ ← ()
    └─ ← ()
  [33730] ContractTest::testExpectEmit()
    ├─ [0] VM::expectEmit(true, false, false, true)
    │   └─ ← ()
    ├─ emit ZEvent(a: 42893283, b: 4223232, c: 0xb4c79dab8f259c7aee6e5b2aa729821864227e84)
    ├─ [31091] Contract::testemit(10)
    │   ├─ [45] ContractTest::fallback{value: 1}()
    │   │   └─ ← ()
    │   ├─ emit ZEvent(a: 1, b: 2, c: 0xb4c79dab8f259c7aee6e5b2aa729821864227e84)
    │   └─ ← ()
    └─ ← ()
```

--

After 
```
[FAIL. Reason: Expected an emit, but no logs were emitted afterward] testExpectEmit() (gas: 33737)
Traces:

  [146057] ContractTest::setUp()
    ├─ → new Contract@0xce71…c246
    │   └─ ← 552 bytes of code
    ├─ [0] VM::deal(0xce71065d4017f316ec606fe4422e11eb2c47c246, 100000000000000000000)
    │   └─ ← ()
    └─ ← ()
  [33737] ContractTest::testExpectEmit()
    ├─ [0] VM::expectEmit(true, false, false, true)
    │   └─ ← ()
    ├─ emit ZEvent(a: 42893283, b: 4223232, c: 0xb4c79dab8f259c7aee6e5b2aa729821864227e84)
    ├─ [31091] Contract::testemit(10)
    │   ├─ [45] ContractTest::fallback{value: 1}()
    │   │   └─ ← ()
    │   ├─ emit ZEvent(a: 1, b: 2, c: 0xb4c79dab8f259c7aee6e5b2aa729821864227e84)
    │   └─ ← ()
    └─ ← "Log != expected log"
```

PR #14316, opened 2026-04-15
chore: supply chain hardening

decofe:
## Summary

Supply chain hardening

## Changes

- **Pin all GH Actions to SHA** — 83 tag-only refs → SHA-pinned via `pinact run --update --min-age 7`
- **Fix template injection** in `benchmarks.yml` — `github.event.inputs.*` moved from `run:` blocks to `env:` blocks (zizmor high-severity findings: 64 → 4)
- **Replace `secrets: inherit`** in `ci.yml` with explicit secret passthrough (`ETHERSCAN_API_KEY`, `ARBITRUM_RPC`, `ETH_SEPOLIA_RPC`) to `test.yml`; removed unnecessary secret inheritance from `docs.yml`
- **Pin Docker base images to digest** — `rust:1-bookworm` and `ubuntu:22.04` in `Dockerfile`
- **Expand Dependabot** to cover `cargo` and `npm` ecosystems (was GH Actions only)
- **Add `minimumReleaseAge = 86400`** to `npm/bunfig.toml` (1-day quarantine for new packages)

## Not addressed (transitive / upstream)

- `jsonwebtoken` 9.3.1 (GHSA-h395, via alloy) — will resolve when alloy bumps to 10.3.0 on main
- `rand` 0.8.5 (RUSTSEC-2026-0097, via alloy) — alloy still on rand 0.8
- `tracing-subscriber` 0.2.25 (RUSTSEC-2025-0055, via ark-relations) — transitive
- `paste`, `derivative` — no fix available upstream
- NPM_TOKEN → OIDC migration — handled separately

PR #14364, opened 2026-04-17
ci: remove softprops/action-gh-release dependabot ignore

decofe:
The ignore rule was added in #13528 because dependabot kept bumping to v2.5.0, which had a draft→finalize race in matrix jobs causing `Too many retries` during releases.

That race was fixed in v2.5.2 ([#746](https://github.com/softprops/action-gh-release/pull/746), [#745](https://github.com/softprops/action-gh-release/pull/745)). We're already on v2.6.1, so the ignore rule is no longer needed — dependabot can resume bumping this action.

Prompted by: zerosnacks

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,8 +6,6 @@ updates:
       interval: "weekly"
     cooldown:
       default-days: 7
-    ignore:
-      - dependency-name: "softprops/action-gh-release"
 
   - package-ecosystem: "cargo"
     directory: "/"


zerosnacks:
Version tag already bumped as part of the git hash pinning change in https://github.com/foundry-rs/foundry/pull/14316 and releases have been working without issue since cc @grandizzy

zerosnacks (approved):

mablr (approved):

PR #14371, opened 2026-04-17
fix(anvil): return error for eth_getLogs with unknown blockHash instead of empty

spalladino:
Return a JSON-RPC error (-32000 "unknown block") when eth_getLogs is called with a blockHash that doesn't resolve to a known block, instead of silently returning an empty array. This aligns with how geth and other clients behave:

```
$ curl -s -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_getLogs","params":[{"blockHash":"0x0000000000000000000000000000000000000000000000000000000000000001"}],"id":1}'  $RPC_URL

{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"unknown block"}}

$ curl -s -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"web3_clientVersion","params":[],"id":1}' $RPC_URL

{"jsonrpc":"2.0","id":1,"result":"Geth/v1.17.2-stable-be4dc0c4/linux-arm64/go1.26.1"}
```

See also https://github.com/paradigmxyz/reth/issues/7371

Code written with Claude.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -2371,7 +2371,7 @@ where
             return Ok(fork.logs(&filter).await?);
         }
 
-        Ok(Vec::new())
+        Err(BlockchainError::UnknownBlock)
     }
 
     /// Returns the logs that match the filter in the given range of blocks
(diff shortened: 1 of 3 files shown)


stevencartavia (approved):

mablr (approved):

mattsse (approved):

spalladino:
Thank you all for the fast response :rocket:

PR #14373, opened 2026-04-18
fix(wip): support offline load-state without fork url

0xKarl98:
• This change fixes the `--load-state` offline restart case from #9721 while keeping `--fork-url` in the command line.

  Previously, if `--fork-url` was still present, Anvil always tried to fetch the fork block during startup. That meant an offline
  restart could still fail with `failed to get fork block`, even when the local state dump and fork cache were already available.

  This patch changes the fork bootstrap logic to prefer local data when restarting from `--load-state`:

  - if `init_state` is present
  - and the fork target is a concrete `--fork-block-number`
  - and the local fork cache exists

  Anvil restores the fork bootstrap metadata from the local cache/state instead of calling remote `get_block()` again.

  ## Test command

  1. Start a local source Anvil:

  ```bash
  ./target/debug/anvil \
    --chain-id 84532 \
    --port 9545

  2. In another terminal, mine to block 2:

  curl --noproxy '*' -s http://127.0.0.1:9545 \
    -H 'content-type: application/json' \
    -d '{"jsonrpc":"2.0","method":"anvil_mine","params":[2],"id":1}'

  3. Create a local state dump and fork cache:

  mkdir -p /tmp/anvil-9721

  4. Generate dump && cache 
  NO_PROXY=127.0.0.1,localhost no_proxy=127.0.0.1,localhost \
    ./target/debug/anvil \
    --fork-url http://127.0.0.1:9545 \
    --fork-block-number 2 \
    --fork-chain-id 84532 \
    --dump-state /tmp/anvil-9721/state.json \
    --no-mining \
    --port 8547


  5. Confirm the local fork cache exists:

  ls ~/.foundry/cache/rpc/base-sepolia/2/storage.json

  6. Restart from local state, but keep an unreachable --fork-url:
      make sure you have quit from step 4 , then run
     ./target/debug/anvil \
    --load-state /tmp/anvil-9721/state.json \
    --fork-url http://127.0.0.1:1 \
    --fork-block-number 2 \
    --fork-chain-id 84532 \
    --no-mining \
    --port 8546

  7. Verify RPC still works:

  curl --noproxy '*' -s http://127.0.0.1:8546 \
    -H 'content-type: application/json' \
    -d '{"jsonrpc":"2.0","method":"eth_chainId","params":[],"id":1}'

  curl --noproxy '*' -s http://127.0.0.1:8546 \
    -H 'content-type: application/json' \
    -d '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}'

  ### Expected result

  - eth_chainId returns 84532 (0x14a34)
  - eth_blockNumber returns 0x2
  - startup does not fail with failed to get fork block

  cc @mattsse


crates/anvil/src/config.rs
```diff
@@ -617,6 +617,7 @@ impl NodeConfig {
     /// Returns the chain ID to use
     pub fn get_chain_id(&self) -> u64 {
         self.chain_id
+            .or_else(|| self.fork_chain_id.map(|chain_id| chain_id.to()))
```
mablr, line 620: `get_chain_id()` now falls back to `fork_chain_id`, but `build_with_eth_backend()` also calls `set_chain_id()` from the same value for the offline case. 

It seems to me  the `get_chain_id()` fallback is dead path, as `chain_id` will already be `Some` by the time it matters. Not 100% sure, could you please recheck?

mablr (commented):
sgtm 👍

Just a small question

scutuatua-crypto (approved):

scutuatua-crypto (approved):

mattsse (commented):
I'm a bit sceptical of this change, because I dont see how this prevents fetching the block, which was the original request, not the chainid behaviour

0xKarl98:
Have renewed the fix after your feedback , please take a look again @mattsse @mablr

mattsse (dismissed):
I think this makes sense,

@zerosnacks this is technically breaking behaviour but I think this is better

mablr (commented):
lgtm, some CI errors to fix

exp0nge:
Can you validate balance calls and the ones below either with the fake RPC or in a Docker container with networking disabled?

```
  - eth_getAccount → fork.get_account()
  - eth_getCode → fork.get_code()
  - eth_getStorageAt → fork.storage_at()
  - eth_call → fork.call()
  - eth_estimateGas → fork.estimate_gas()
  - eth_getProof → fork.get_proof()
  - eth_createAccessList → fork.create_access_list()
  - eth_feeHistory → fork.fee_history()
 ```

I took another look at this change again to see if this fix alone does it (https://github.com/foundry-rs/foundry/pull/12442) but the above fall-throughs and a number of other issues around backend makes cache misses so you end up with calling remote again. Not sure if you are only solving to startup load state though, so maybe it's fine just for that, but doesn't solve https://github.com/foundry-rs/foundry/issues/9721

0xKarl98:
@exp0nge Currently it only solves to startup load-state ,  would need follow-up prs to close fully off-line feature imo

exp0nge:
> @exp0nge Currently it only solves to startup load-state , would need follow-up prs to close fully off-line feature imo

okay, then I'll rebase once these PRs land. I think I have the complete offline iteration/PR above.

0xKarl98:
@exp0nge that would be great , thank you !

PR #14378, opened 2026-04-20
chore(tempo): bump rev + use extension traits

decofe:
Bumps tempo crates from `6f4f5cc` to `bb08bb9` (latest `main`) and adopts the new `TempoProviderExt` and `TempoAddressExt` traits across the codebase.

- Replace local `is_hardfork_active` helper with `TempoProviderExt::is_hardfork_active()`, which queries `tempo_forkSchedule` RPC directly instead of manually fetching chain ID + block timestamp
- Use `addr.is_tip20()` instead of standalone `is_tip20_prefix()` in `TempoLabels` inspector
- Use `Address::TIP20_PREFIX` instead of hardcoded hex bytes in fee token address parsing

Prompted by: rusowsky

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1194,7 +1194,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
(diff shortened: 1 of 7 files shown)


zerosnacks (approved):
Nice! 👍

mablr (approved):
clean 👍

PR #14383, opened 2026-04-21
feat(verify): clearer error when `ETHERSCAN_API_KEY` selects etherscan

o-az:
a user hit this trying to verify a contract against a sourcify server while an unrelated `ETHERSCAN_API_KEY` was lingering in their env:

```
Error: Provided URL `https://contracts.tempo.xyz` is host only.
 Did you mean to use the API endpoint`https://contracts.tempo.xyz/api` ?
```

the message is misleading: sourcify/blockscout accept host-only urls fine. the real cause is that `ETHERSCAN_API_KEY` silently makes `forge verify-contract` default to the etherscan verifier (step 4 in `VerificationProviderType::client`), which then rejects the host-only url. nothing in the cli output tells the user etherscan was even selected.

this pr makes that path self-explanatory:

- announce the silent fallback, mirroring the existing sourcify default announce ("attempting to verify on sourcify..."):
  ```
  ETHERSCAN_API_KEY is set, defaulting to Etherscan verifier. Unset it or pass `--verifier sourcify` (or another provider) to override.
  ```
- rewrite the host-only hint to name the verifier, give two concrete fixes, and explain why etherscan was selected. the hint is scoped to the etherscan code path so sourcify/blockscout users with host-only urls don't see misleading advice.
- extract the error-wrapping into `utils::wrap_verifier_url_error` next to `is_host_only`, matching the existing convention of free helpers in `utils.rs`.

before:

```
Error: Provided URL `https://contracts.tempo.xyz` is host only.
 Did you mean to use the API endpoint`https://contracts.tempo.xyz/api` ?
```

after:

```
ETHERSCAN_API_KEY is set, defaulting to Etherscan verifier. Unset it or pass `--verifier sourcify` (or another provider) to override.
Error: Verifier `etherscan` requires an API endpoint, but `--verifier-url` is host-only: `https://contracts.tempo.xyz`.
Fixes (pick one):
- Append the API path, e.g. `--verifier-url https://contracts.tempo.xyz/api`
- Switch verifier, e.g. `--verifier sourcify` (works with host-only URLs)

Etherscan was selected because `ETHERSCAN_API_KEY` is set or `--verifier etherscan` was passed. Unset the env var or pass `--verifier sourcify` to pick a different provider.
```

### DIFF
--- a/crates/verify/src/provider.rs
+++ b/crates/verify/src/provider.rs
@@ -223,7 +223,7 @@ impl VerificationProviderType {
 
         // 4. If no `--verifier` is specified but `ETHERSCAN_API_KEY` is set, default to Etherscan.
         if has_key {
-            sh_println!(
+            sh_eprintln!(
                 "ETHERSCAN_API_KEY is set, defaulting to Etherscan verifier. \
                  Unset it or pass `--verifier sourcify` (or another provider) to override."
             )?;
--- a/crates/verify/src/provider.rs
+++ b/crates/verify/src/provider.rs
@@ -125,13 +125,6 @@ pub trait VerificationProvider {
 
     /// Checks whether the contract is verified.
     async fn check(&self, args: VerifyCheckArgs) -> Result<()>;
-
-    /// Returns the concrete verifier type this provider implements.
-    ///
-    /// Useful for logging and for tests that need to assert which provider
-    /// was selected by [`VerificationProviderType::client`] (whose return type
-    /// is a trait object and therefore erases the concrete type).
-    fn kind(&self) -> VerificationProviderType;
 }
 
 impl FromStr for VerificationProviderType {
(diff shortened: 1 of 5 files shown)


mablr (commented):
error

crates/verify/src/provider.rs
```diff
@@ -223,6 +223,10 @@ impl VerificationProviderType {
 
         // 4. If no `--verifier` is specified but `ETHERSCAN_API_KEY` is set, default to Etherscan.
         if has_key {
+            sh_println!(
+                "ETHERSCAN_API_KEY is set, defaulting to Etherscan verifier. \
+                 Unset it or pass `--verifier sourcify` (or another provider) to override."
+            )?;
```
mablr: This should be probably `sh_eprintln!` to go to stderr, and comply with `--json` flag

crates/verify/src/utils.rs
```diff
@@ -386,6 +386,39 @@ pub fn is_host_only(url: &Url) -> bool {
...
+    if is_host_only(&url) && using_etherscan {
+        return err.wrap_err(format!(
+            "Verifier `etherscan` requires an API endpoint, but `--verifier-url` is host-only: `{verifier_url}`.\n\
+             Fixes (pick one):\n\
+             - Append the API path, e.g. `--verifier-url {verifier_url}/api`\n\
+             - Switch verifier, e.g. `--verifier sourcify` (works with host-only URLs)\n\
+             \n\
+             Etherscan was selected because `ETHERSCAN_API_KEY` is set or `--verifier etherscan` \
+             was passed. Unset the env var or pass `--verifier sourcify` to pick a different \
+             provider."
```
figtracer: i think we don't need this last sentence here

zerosnacks (approved):
Thanks!

figtracer (approved):

PR #14384, opened 2026-04-21
fix(anvil): refetch full fork blocks with missing tx cache

solanaXpeter:
## Motivation

Cached fork blocks can lose one or more stored transaction payloads while still keeping the block shell. When `block_by_hash_full` or `block_by_number_full` reconstructs that stale cache entry, RPC consumers receive an incomplete full block instead of a refetched upstream response.

## Solution

Return `None` from full-block reconstruction when any cached transaction is missing, then fall back to `fetch_full_block(...)` so the fork cache is repaired from upstream data. Add an integration test that removes one cached transaction and asserts the next full-block read refetches and restores it.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/eth/backend/fork.rs
+++ b/crates/anvil/src/eth/backend/fork.rs
@@ -442,10 +442,10 @@ impl<N: Network> ClientFork<N> {
         &self,
         hash: B256,
     ) -> Result<Option<N::BlockResponse>, TransportError> {
-        if let Some(block) = self.storage_read().blocks.get(&hash).cloned() {
-            if let Some(block) = self.convert_to_full_block(block) {
-                return Ok(Some(block));
-            }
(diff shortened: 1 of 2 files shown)


figtracer (dismissed):
this is defensive but not sure how important it is here, pending another review

mattsse (dismissed):
doesnt hurt

solanaXpeter:
Hi @mattsse, thanks for your review. I noticed this was approved and later closed while I was working on the CI issue, so I wanted to check whether there was some other concern with the change.

mablr:
re-opened 👍

solanaXpeter:
@mablr @mattsse @figtracer I fixed the CI issue. Could you please re-run the CI? Thanks very much for your time!

mattsse (approved):

PR #14385, opened 2026-04-21
refactor(cast): cleanup `cast send`

0xrusowsky:
this PR cleans up cast cmds that support Tempo access keys. It mainly does it by adding 2 new helpers:
- `fn cast_send_with_access_key()`
- `CallSpec::resolve()`

## PR Checklist

- [ ] Added Tests
- [X] Added Documentation
- [ ] Breaking changes


crates/cast/src/cmd/keychain.rs
```diff
@@ -748,10 +746,7 @@ async fn send_keychain_tx(
             .wallet(wallet)
             .connect_provider(&provider);
 
-        let cast = CastTxSender::new(provider);
-        let pending_tx = cast.send(tx).await?;
-        let tx_hash = *pending_tx.inner().tx_hash();
-        cast.print_tx_result(tx_hash, send_tx.cast_async, send_tx.confirmations, timeout).await?;
+        cast_send(provider, tx, send_tx.cast_async, false, send_tx.confirmations, timeout).await?;
```
mablr: Here would it be preferable to pass `send_tx.sync` to `cast_send` instead of `false`.

As before there were a raw use of `CastTxSender`, implicitly it was always false.

mablr (commented):
clean 👌

crates/cast/src/cmd/keychain.rs
0xrusowsky: adddressed in [`0a7a924` (this PR)](https://github.com/foundry-rs/foundry/pull/14385/commits/0a7a924159cca18fd5b2e1201167f7d57c4b5d4f)

mattsse (approved):

zerosnacks (approved):

PR #14386, opened 2026-04-21
chore: harden Makefile & CI

zerosnacks:
Hardens Makefile to always use `--locked` to prevent accidental pulls of new dependency versions when the lockfile is stale. If you run into a stale lockfile run `make lock`, this runs `cargo fetch` rather than `cargo update`.


mablr (approved):
👌

PR #14388, opened 2026-04-21
fix(cheatcodes): read broadcasts with the active network

solanaXpeter:
## Motivation

The filesystem cheatcodes already run under generic `FoundryEvmNetwork`, but the broadcast and deployment readers still deserialize artifacts as `Ethereum`. That rejects valid Tempo broadcast data even though the surrounding execution path is already network-generic.

## Solution

Thread the active `FoundryEvmNetwork::Network` through `getBroadcast*`, `getDeployment*`, `getDeployments`, and `latest_broadcast` instead of hardcoding `Ethereum`. Add a unit test with a minimal Tempo broadcast artifact to verify the latest sequence is parsed through the generic path.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cheatcodes/src/fs.rs
+++ b/crates/cheatcodes/src/fs.rs
@@ -6,7 +6,7 @@ use crate::{
 };
 use alloy_dyn_abi::DynSolType;
 use alloy_json_abi::ContractObject;
-use alloy_network::{Ethereum, Network, ReceiptResponse};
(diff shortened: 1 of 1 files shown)


mablr (approved):
Good catch! Tysm!

figtracer (approved):

zerosnacks (approved):
Thanks! Good catch

PR #14389, opened 2026-04-21
feat: make asm-keccak a default feature in all binaries

DaniPopes:
Makes `asm-keccak` a default feature in all four binary crates (forge, anvil, cast, chisel), matching reth's approach where production binaries enable assembly-optimized keccak256 out-of-the-box.

Also forwards `revm/asm-keccak` where revm is a direct dependency (forge, anvil, cast) so the EVM executor also uses the fast path.

Since `asm-keccak` is now a default, it's removed from the explicit feature lists in the Makefile and CI workflows (`release.yml`, `docker-publish.yml`).

### DIFF
--- a/crates/anvil/Cargo.toml
+++ b/crates/anvil/Cargo.toml
@@ -124,8 +124,8 @@ op-alloy-rpc-types.workspace = true
 tempo-alloy.workspace = true
 
 [features]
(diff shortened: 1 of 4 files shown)


mattsse (approved):

zerosnacks (commented):
makes sense 👍

Issue #10322, opened 2025-04-17
bug(`forge test --debug`): debugger does not open if `setUp` reverts


e1Ru1o:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

forge Version: 1.0.0-stable Commit SHA: e144b82070619b6e10485c38734b4d4d45aebe04

### What version of Foundryup are you on?

foundryup: 1.0.1

### What command(s) is the bug in?

forge test --mt someTest --debug

### Operating System

macOS (Apple Silicon)

### Describe the bug

Tests where the `setUp` fail don't open the debugger. E.g.

```solidity
import {Test} from "@forge-std/Test.sol";

contract TestTest is Test {
    function setUp() public {
        revert();
    }

    function testFoo() public {}
}
```

Then `forge test --mc TestTest --debug` just fails

zerosnacks:
able to reproduce with setup:

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.13;

import {Test, console} from "forge-std/Test.sol";
import {Counter} from "../src/Counter.sol";

contract CounterTest is Test {
    Counter public counter;

    function setUp() public {
        counter = new Counter();
        counter.setNumber(0);

        revert();
    }

    function test_Increment() public {
        counter.increment();
        assertEq(counter.number(), 1);
    }

    function testFuzz_SetNumber(uint256 x) public {
        counter.setNumber(x);
        assertEq(counter.number(), x);
    }
}
```

```
forge test --debug --match-path test/Counter.t.sol --mt test_Increment
[⠊] Compiling...
[⠆] Compiling 1 files with Solc 0.8.29
[⠰] Solc 0.8.29 finished in 240.44ms
Compiler run successful!

Ran 1 test for test/Counter.t.sol:CounterTest
[FAIL: EvmError: Revert] setUp() (gas: 0)
Suite result: FAILED. 0 passed; 1 failed; 0 skipped; finished in 270.22µs (0.00ns CPU time)

Ran 1 test suite in 2.59ms (270.22µs CPU time): 0 tests passed, 1 failed, 0 skipped (1 total tests)
Error: debug arena is empty
```

PR #14390, opened 2026-04-21
fix(forge): preserve debug traces when setUp reverts

IrshadKohl51:
## Motivation
                                                                                                                                                                                      
  `forge test --debug` currently only feeds `Execution` traces into the debugger.                                                                                                     
  When `setUp()` reverts, there is no execution trace, so debugger startup fails with `Error: debug arena is empty` and `--dump` cannot be used in this failure path (see #10322).    
                                                                                                                                                                                      
  ## Solution                                                                                                                                                                         
                                                                                                                                                                                      
  - In `forge test --debug`, prefer `Execution` traces, but fall back to available `Setup`/`Deployment` traces when execution traces are empty.                                       
  - Adjust setup tracing override logic in the runner so debug tracing is preserved for debug runs (while keeping the existing optimization for non-debug runs and invariant needs).  
  - Add a regression test `test_debug_with_dump_setup_revert` to verify:                                                                                                              
    - `--debug --dump` completes when `setUp()` reverts,                                                                                                                              
    - dump file is generated,
    - `debug arena is empty` is not emitted,                                                                                                                                          
    - setup failure is still reported in output.                                                                                                                                      

  ## PR Checklist                                                                                                                                                                     
                                                                                                                                                                                      
  - [x] Added Tests                                                                                                                                                                   
  - [ ] Added Documentation                                                                                                                                                           
  - [ ] Breaking changes


mablr (dismissed):
sgtm, thanks

figtracer (commented):
should probably do the same for invariant runs too, everything else lgtm

mablr:
> should probably do the same for invariant runs too, everything else lgtm

@figtracer follow-up ? Is this g2g ?

figtracer (dismissed):

figtracer:
> > should probably do the same for invariant runs too, everything else lgtm
> 
> @figtracer follow-up ? Is this g2g ?

would like to get another approval from @zerosnacks

DaniPopes (dismissed):

mablr (dismissed):

mablr (approved):
Fixed clippy, sgtm, ty @IrshadKohl51

figtracer (approved):

PR #14392, opened 2026-04-21
feat(anvil): add --fund-accounts CLI flag

exp0nge:
## Motivation

When testing with Anvil, it's common to need specific addresses funded with custom balances on startup (e.g. for impersonation testing or reproducing on-chain scenarios). Currently this requires sending transactions after startup or using a custom genesis file.

## Solution

Add a `--fund-accounts` CLI flag that accepts `ADDRESS:AMOUNT` pairs (amount in ETH):

```
anvil --fund-accounts 0x1234...5678:1000 0xabcd...ef01:5000
```

This funds the specified addresses during node initialization, after state loading but before the node is ready.

- Add `--fund-accounts` arg to `NodeArgs` with space-delimited values
- Add `funded_accounts` field + `with_funded_accounts()` builder to `NodeConfig`
- Apply funding via `set_balance` during backend setup

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mablr (commented):
Hi, thanks for you proposition.

Just wondering why `anvil_setBalance`/`anvil_addBalance` endpoints aren't enough ?

exp0nge:
> Hi, thanks for you proposition.
> 
> Just wondering why `anvil_setBalance`/`anvil_addBalance` endpoints aren't enough ?

This would be a one-shot (startup) option. IIUC the anvil_* namespace would need to be part of the post startup bootstrapping?

mablr:
> IIUC the anvil_* namespace would need to be part of the post startup

Yes simply call `anvil_setBalance`/`anvil_addBalance` endpoints after startup.

Is there a specific use-case where this is not suitable ? If so could please you elaborate more ?

exp0nge:
> > IIUC the anvil_* namespace would need to be part of the post startup
> 
> Yes simply call `anvil_setBalance`/`anvil_addBalance` endpoints after startup.
> 
> Is there a specific use-case where this is not suitable ? If so could please you elaborate more ?

We are using this as a drop-in replacement for other dev/localnets which come with some pre-funded addresses. Similar to the genesis configurations where you can bake in your addresses. Post-boostrap would be extra setup, I can do that and close this PR if you don't think this is worth as a extra feature

mablr:
I see, it shouldn't hurt to have this.

Could you please resolve the merge conflict? Once done, I’ll trigger CI. 👍

exp0nge:
> I see, it shouldn't hurt to have this.
> 
> Could you please resolve the merge conflict? Once done, I’ll trigger CI. 👍

done

mablr (approved):

exp0nge:
@mablr would like to rebase the original offline PR, who can be the second approver for this?

figtracer (dismissed):
this could make some agentic workflows easier as well, lgtm

exp0nge:
Next steps for merge?

crates/anvil/src/config.rs
```diff
@@ -1109,6 +1112,16 @@ impl NodeConfig {
         self
     }
 
+    /// Sets accounts to fund with custom balances on startup.
+    #[must_use]
+    pub fn with_funded_accounts(
+        mut self,
+        accounts: std::collections::HashMap<Address, U256>,
```
mablr: Could you please use imports everywhere instead of fully-qualified forms ?

mablr (changes requested):
Smol pedantic nit :)

(sorry for the delay)

crates/anvil/src/config.rs
exp0nge: 02f68b93cded1e10838d048b070dadfc0c650442

mablr (approved):
lgtm, pending another approval.

(fixed fmt, and resolved conflict)

stevencartavia (approved):

PR #14394, opened 2026-04-21
feat(forge): browser wallet support for `create` subcommand

mablr:
## Motivation

Bring browser wallet support for `forge create` subcommand

Close #12958

## Solution

- add `BrowserWalletOpts` to `CreateArgs`
- Chain-based network dispatch

## Tests

```
forge create src/Counter.sol:Counter --rpc-url https://rpc.tempo.xyz --broadcast --browser --browser-disable-open
```

Explicit `fee-token` (e.g. USDC.e):
```
forge create src/Counter.sol:Counter --rpc-url https://rpc.tempo.xyz --broadcast --browser --browser-disable-open --tempo.fee-token 0x20c000000000000000000000b9537d11c60e8b50
```

PR #13879, opened 2026-03-22
fix(cast): add browser wallet support for erc20 commands

joohhnnn:
The `--browser` flag was ignored by erc20 state-changing commands (transfer, approve, mint, burn). They called `get_provider_with_wallet` which requires a local signer, so `--browser` without a private key failed with "Error accessing local wallet".

Fixed by adding browser wallet detection before the provider setup, following the same pattern as `cast send --browser` — build a plain provider, construct the tx, then send via `BrowserSigner::send_transaction_via_browser`.

Previous attempt: #13220 (stale due to Cast refactor)

Closes #13103 cc @mablr

PR #14395, opened 2026-04-21
fix(cast): add browser wallet support for erc20 commands

figtracer:
## Summary
- add a browser-wallet branch to the erc20 command flow so `cast erc20 ... --browser` uses the browser signer instead of falling through to local signer handling
- preserve the existing transaction filling behavior for browser-submitted ERC20 transactions by filling only missing fields before dispatch
- leave Tempo fee-token browser signing behavior unchanged; MetaMask/viem still rejects custom Tempo tx type `0x76`

## Notes
- Plain Tempo browser flow reaches the wallet path correctly, but MetaMask may still block submission if the connected wallet lacks gas funds on Tempo.
- Tempo fee-token browser signing remains unsupported in MetaMask because viem rejects custom envelope type `0x76`.

### DIFF
--- a/crates/cast/src/cmd/erc20.rs
+++ b/crates/cast/src/cmd/erc20.rs
@@ -3,13 +3,13 @@ use std::{str::FromStr, time::Duration};
 use crate::{
     cmd::send::{cast_send, cast_send_with_access_key},
     format_uint_exp,
-    tx::{SendTxOpts, TxParams},
+    tx::{CastTxSender, SendTxOpts, TxParams},
 };
 use alloy_consensus::{SignableTransaction, Signed};
 use alloy_eips::BlockId;
 use alloy_ens::NameOrAddress;
-use alloy_network::{Ethereum, EthereumWallet, Network};
-use alloy_primitives::U256;
+use alloy_network::{Ethereum, EthereumWallet, Network, TransactionBuilder};
+use alloy_primitives::{Address, U256};
 use alloy_provider::{Provider, fillers::RecommendedFillers};
 use alloy_signer::Signature;
 use alloy_sol_types::sol;
@@ -300,7 +300,11 @@ impl Erc20Subcommand {
         };
 
         let is_tempo = self.erc20_opts().is_some_and(|erc20| erc20.tempo.is_tempo())
-            || tempo_access_key.is_some();
+            || tempo_access_key.is_some()
+            || {
+                let config = self.rpc_opts().load_config()?;
+                get_chain(config.chain, &get_provider(&config)?).await?.is_tempo()
+            };
 
         if is_tempo {
             self.run_generic::<TempoNetwork>(signer, tempo_access_key).await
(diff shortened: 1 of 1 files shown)


crates/cast/src/cmd/erc20.rs
```diff
@@ -503,3 +522,46 @@ impl Erc20Subcommand {
         Ok(())
     }
 }
+
+/// Fills from, chain_id, nonce, fees, and gas limit on a transaction request for the browser
+/// wallet path. Mirrors the filling logic in [`CastTxBuilder::build`] but operates on a
+/// pre-built transaction request from the sol! macro rather than through the builder pipeline.
+/// Only fills fields that haven't already been set by the user.
+async fn fill_tx<N: Network, P: Provider<N>>(
```
figtracer, line 534: this mirrors `CastTxBuilder`'s behavior, could probably extract a method here not sure

mablr (commented):
For the fee-token selection, it may be relevant to have a chain-based network detection similarly to `forge create` #14394 with such logic https://github.com/foundry-rs/foundry/blob/1dc9ece97d7eacab93922bb29775d86cf2ac8007/crates/forge/src/cmd/create.rs#L398-L405

crates/cast/src/cmd/erc20.rs
```diff
@@ -356,6 +356,38 @@ impl Erc20Subcommand {
...
+                    } else {
+                        CastTxSender::new(&$provider)
+                            .print_tx_result(
+                                tx_hash,
+                                $send_tx.cast_async,
+                                $send_tx.confirmations,
+                                timeout,
+                            )
+                            .await?
+                    }
```
mablr: Do we really need this ? `provider` should be `N`-typed, in this case both branches are the same ?

figtracer: i tried the simpler approach and it failed testing for some reason might have to test this again

edit: if we don't pass tempo flags N is Ethereum here, if we use Tempo Wallet it doesn't serialize properly
this is what i was getting before `deserialization error: unknown variant 0x76` @mablr

mablr: I had similar trouble on `forge create`, it was fixed by including the chain id in the set of conditions for the early `Network` dispatch.

mablr (changes requested):
Tested locally, it doesn't work:
```
cargo +nightly run -p cast -- erc20 transfer 0x20c0000000000000000000000000000000000000 0x0000000000000000000000000000000000000001 100000 --rpc-url 'https://rpc.tempo.xyz' --browser
```

```
Details: Cannot convert null to a BigInt
```

I believe we get this viem error because at least `fee_token` is not set.

figtracer:
> Tested locally, it doesn't work:
> 
> ```
> cargo +nightly run -p cast -- erc20 transfer 0x20c0000000000000000000000000000000000000 0x0000000000000000000000000000000000000001 100000 --rpc-url 'https://rpc.tempo.xyz' --browser
> ```
> 
> ```
> Details: Cannot convert null to a BigInt
> ```
> 
> I believe we get this viem error because at least `fee_token` is not set.


yeah i got this too, assumed this should be explicit; will make it default to PathUSD

mablr (commented):
Using Tempo wallet I managed to transfer tip20 tokens with both implicit default `fee_token` and explicit flag.

Still broken w/ metamask:
```
Error: Transaction rejected: Invalid parameters were provided to the RPC method.
Double check you have provided the correct parameters.

Request Arguments:
  chain:                 Tempo Mainnet (id: 4217)
  from:                  0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
  to:                    0x20c0000000000000000000000000000000000000
  data:                  0xa0xf39Fd6e51aad88F6F4ce6aB8827279cffFb922660xf39Fd6e51aad88F6F4ce6aB8827279cffFb922660xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
  gas:                   1000000
  maxFeePerGas:          40.000000001 gwei
  maxPriorityFeePerGas:  0.000000001 gwei
  nonce:                 2

Details: Invalid transaction envelope type: "0x76". Must be one of: 0x0, 0x1, 0x2, 0x4
```

Should be fine as-is for now as long as the CI passes...

figtracer:
> Using Tempo wallet I managed to transfer tip20 tokens with both implicit default `fee_token` and explicit flag.
> 
> Still broken w/ metamask:
> 
> ```
> Error: Transaction rejected: Invalid parameters were provided to the RPC method.
> Double check you have provided the correct parameters.
> 
> Request Arguments:
>   chain:                 Tempo Mainnet (id: 4217)
>   from:                  0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
>   to:                    0x20c0000000000000000000000000000000000000
>   data:                  0xa0xf39Fd6e51aad88F6F4ce6aB8827279cffFb922660xf39Fd6e51aad88F6F4ce6aB8827279cffFb922660xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
>   gas:                   1000000
>   maxFeePerGas:          40.000000001 gwei
>   maxPriorityFeePerGas:  0.000000001 gwei
>   nonce:                 2
> 
> Details: Invalid transaction envelope type: "0x76". Must be one of: 0x0, 0x1, 0x2, 0x4
> ```
> 
> Should be fine as-is for now as long as the CI passes...

the MM problem should be done on a follow-up and we also need to decide what path to go with 

i think the solution mentioned here should do it 
https://github.com/foundry-rs/foundry/pull/13879#issuecomment-4145282796

mablr (approved):

zerosnacks (approved):

PR #14404, opened 2026-04-22
feat(common): add MPP WebSocket transport

stevencartavia:
Adds MPP payment support for `ws://` and `wss://` RPC URLs, matching the existing HTTP 402 handling.

`MppWsConnect` implements `PubSubConnect` — on connect it opens a WebSocket, probes for an MPP challenge frame (timeout fallback for non-MPP servers), performs the credential handshake, then hands the authenticated socket to alloy's `WsBackend::from_socket` ([alloy#3911](https://github.com/alloy-rs/alloy/pull/3911)). Reuses `LazySessionProvider` from the HTTP transport.


gakonst:
Ι kind of wish we could make this work on Alloy as a layer on the transport somehow- is that possible?

stevencartavia:
cc @zerosnacks @grandizzy

zerosnacks:
I prefer we merge this as proposed so we can include it in the release and we can then move this to Alloy as a follow up soon after

crates/common/src/provider/runtime_transport.rs
```diff
@@ -244,13 +246,12 @@ impl RuntimeTransport {
         Ok(InnerTransport::Http(LazyMppHttpTransport::lazy(client, self.url.clone())))
     }
 
-    /// Connects to a WS transport.
+    /// Connects to a WS transport with automatic MPP payment handling.
     async fn connect_ws(&self) -> Result<InnerTransport, RuntimeTransportError> {
-        let auth = self.jwt.as_ref().and_then(|jwt| build_auth(jwt.clone()).ok());
-        let mut ws = WsConnect::new(self.url.to_string());
-        if let Some(auth) = auth {
+        let mut ws = MppWsConnect::new(self.url.to_string());
```
grandizzy: is this safe to assume and apply for all ws connections, wonder if we could come up with a different way to determine if mpp (maybe from URL?) and have a `connect_ws_mpp` kind?

crates/common/src/provider/mpp/ws.rs
```diff
@@ -0,0 +1,484 @@
...
+    transport::{LazySessionProvider, extract_challenge_chain_and_currency},
+};
+
+type TungsteniteStream = WebSocketStream<MaybeTlsStream<tokio::net::TcpStream>>;
+
+/// Timeout for waiting on an MPP challenge frame from the server.
+///
+/// If no challenge arrives within this window after the WebSocket upgrade
+/// completes, we assume it's a plain (non-MPP) JSON-RPC WebSocket.
+const MPP_CHALLENGE_TIMEOUT: Duration = Duration::from_millis(500);
```
mablr, line 44: This seems like a big penalty for non-MPP JSON-RPC WS. (joining @grandizzy's [remark](https://github.com/foundry-rs/foundry/pull/14404#discussion_r3129651153)) 

Should be configurable ?

crates/common/src/provider/mpp/ws.rs
```diff
@@ -0,0 +1,484 @@
...
+            self.url.as_str().into_client_request().map_err(TransportErrorKind::custom)?;
+
+        if let Some(ref auth) = self.auth {
+            let mut auth_value = reqwest::header::HeaderValue::from_str(&auth.to_string())
+                .map_err(TransportErrorKind::custom)?;
+            auth_value.set_sensitive(true);
+            request.headers_mut().insert(reqwest::header::AUTHORIZATION, auth_value);
+        }
+
+        let (mut socket, _) = connect_async(request).await.map_err(TransportErrorKind::custom)?;
```
figtracer, line 265: unsure if we're missing this step before connecting directly
https://github.com/alloy-rs/alloy/blob/v2.0.1/crates/transport-ws/src/native.rs#L138-L142

crates/common/src/provider/runtime_transport.rs
mablr: I fixed this partially by checking if MPP key is available.

crates/common/src/provider/mpp/ws.rs
mablr: fixed

crates/common/src/provider/runtime_transport.rs
grandizzy: nice, there could still be cases when we try mpp even if not needed but this should narrow them. wonder if we should keep a registry of mpp services, something similar with how we do for block explorers (we could include https://github.com/foundry-rs/foundry/blob/85b3e396012792791c60acacef15d023a0eb6062/crates/config/src/endpoints.rs#L531-L532 too)

mablr: added simple url matching, should do the job.

grandizzy: thanks!

crates/common/src/provider/runtime_transport.rs
```diff
@@ -18,6 +21,17 @@ use tokio::sync::RwLock;
 use tower::Service;
 use url::Url;
 
+/// Known MPP-enabled RPC host suffixes.
+///
+/// Endpoints matching these patterns are always connected via [`MppWsConnect`],
+/// regardless of whether local MPP keys have been discovered.
+const KNOWN_MPP_HOSTS: &[&str] = &[".mpp.tempo.xyz"];
```
zerosnacks, line 28: I would like to get Quicknode to add MPP WSS support, would they still be able to with this configuration? 

It is a sensible optimization but ideally we don't block adoption by maintaining a fixed list forcing their users to use nightly.

mablr: reverting ?

zerosnacks: Decided on keeping the guards, get Tempo MPP WS support first and then iterate from there

zerosnacks (approved):

figtracer (approved):

mablr (approved):

PR #14405, opened 2026-04-22
fix(anvil-server): distinguish empty and notification batches

solanaXpeter:
## Motivation

The batch request handler currently collapses empty batches and notification-only batches into the same `None` path. That makes `[]` return the same `invalid_request` fallback as a notification-only batch, even though the former should be rejected and the latter should not produce a response at all.

## Solution

Treat empty batches as `invalid_request` inside `handle_request`, keep notification-only batches as `None`, return HTTP `204 No Content` when no response should be emitted, and avoid queueing websocket responses for notification-only batches. Add regression coverage for the HTTP handler and pubsub connection paths.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mablr (approved):
Makes sense

pending another review...

mablr:
merged master to fix RUSTSEC deny warning

mattsse (approved):

PR #14420, opened 2026-04-23
fix(evm): sync gas params when updating executor spec

solanaXpeter:
## Motivation

`Executor::set_spec_id` still updates `cfg_env.spec` directly, while the other spec-switching paths already route through `set_spec_and_mainnet_gas_params`. That leaves spec-dependent gas params and Amsterdam EIP-8037 state stale when the executor spec changes after construction.

## Solution

Route `Executor::set_spec_id` through `set_spec_and_mainnet_gas_params` and add a regression test that forces a `HOMESTEAD -> AMSTERDAM` transition to verify both the gas table and the Amsterdam flag update together.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/evm/src/executors/mod.rs
+++ b/crates/evm/evm/src/executors/mod.rs
@@ -207,9 +207,9 @@ impl<FEN: FoundryEvmNetwork> Executor<FEN> {
         self.evm_env.cfg_env.spec
     }
 
-    /// Sets the EVM spec.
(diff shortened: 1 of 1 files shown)


mablr (commented):
good catch! thanks

edit: fixed the CI errors

mablr (approved):

figtracer (approved):

PR #14423, opened 2026-04-23
fix(cheatcodes): reject nested debug trace recording

Perico-perica46:
## Motivation

`startDebugTraceRecording` currently overwrites the saved tracer config each time it is called. If recording is started twice before `stopAndReturnDebugTraceRecording`, the final restore uses the overwritten config instead of the original one, which can leak debug tracing settings across later execution.

## Solution

Reject repeated `startDebugTraceRecording` calls while a recording session is already active, and add a regression test that verifies the second start reverts while the first recording remains usable.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cheatcodes/src/evm.rs
+++ b/crates/cheatcodes/src/evm.rs
@@ -1246,6 +1246,10 @@ impl Cheatcode for startDebugTraceRecordingCall {
             return Err(Error::from("no tracer initiated, consider adding -vvv flag"));
         };
 
+        if ccx.state.record_debug_steps_info.is_some() {
+            bail!("debug trace recording was already started");
+        }
+
         let mut info = RecordDebugStepInfo {
             // will be updated later
             start_node_idx: 0,
--- a/testdata/default/cheats/RecordDebugTrace.t.sol
+++ b/testdata/default/cheats/RecordDebugTrace.t.sol
@@ -85,14 +85,16 @@ contract RecordDebugTraceTest is Test {
         for (uint256 i = 0; i < steps.length; i++) {
             Vm.DebugStep memory step = steps[i];
             if (
-                step.opcode == 0x52 /*MSTORE*/ && step.stack[0] == testContract.memPtr() // MSTORE offset
+                step.opcode == 0x52 /*MSTORE*/
+                    && step.stack[0] == testContract.memPtr() // MSTORE offset
                     && step.stack[1] == testContract.expectedValueInMemory() // MSTORE val
             ) {
                 mstoreCalled = true;
             }
 
             if (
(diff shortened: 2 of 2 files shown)


testdata/default/cheats/RecordDebugTrace.t.sol
```diff
@@ -85,16 +85,14 @@ contract RecordDebugTraceTest is Test {
         for (uint256 i = 0; i < steps.length; i++) {
             Vm.DebugStep memory step = steps[i];
             if (
-                step.opcode == 0x52 /*MSTORE*/
-                    && step.stack[0] == testContract.memPtr() // MSTORE offset
+                step.opcode == 0x52 /*MSTORE*/ && step.stack[0] == testContract.memPtr() // MSTORE offset
```
mattsse: did anything change here?

crates/cheatcodes/src/evm.rs
```diff
@@ -1246,6 +1246,10 @@ impl Cheatcode for startDebugTraceRecordingCall {
             return Err(Error::from("no tracer initiated, consider adding -vvv flag"));
         };
 
+        if ccx.state.record_debug_steps_info.is_some() {
+            bail!("debug trace recording was already started");
```
mattsse, line 1250: should we instead just return?

mattsse (changes requested):
not super familiar with this feature but I believe this makes sense

but unsure if this should be a hard error

testdata/default/cheats/RecordDebugTrace.t.sol
mablr: just fmt should be reverted i believe

Perico-perica46:
@mattsse Thanks, I had the same concern. I leaned toward a hard error since there is only one active recording slot, but I'm happy to switch this to a no-op if you'd prefer idempotent behavior.

testdata/default/cheats/RecordDebugTrace.t.sol
Perico-perica46: This is just formatting noise, i will revert it.

crates/cheatcodes/src/evm.rs
Perico-perica46: a silent return would make the second call look valid even though it didn't actually start a new recording. but happy to change it if you'd prefer that.

testdata/default/cheats/RecordDebugTrace.t.sol
Perico-perica46: @mattsse @mablr Reverted the unrelated formatting hunk.

DaniPopes (approved):

stevencartavia (approved):

mablr (approved):

mattsse (approved):

PR #14425, opened 2026-04-23
fix(anvil): separate queued txs in txpool RPC responses

mk0walsk:
## Motivation

`txpool_inspect` and `txpool_content` currently insert queued transactions into the `pending` bucket. This makes their output inconsistent with `txpool_status` and misreports nonce-gapped transactions to RPC consumers.

## Solution

Write queued transactions into the `queued` bucket for both RPC methods and add a regression test that builds a nonce gap and asserts pending/queued separation across `txpool_status`, `txpool_inspect`, and `txpool_content`.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -3229,7 +3229,7 @@ impl EthApi<FoundryNetwork> {
             entry.insert(key, convert(pending));
         }
         for queued in self.pool.pending_transactions() {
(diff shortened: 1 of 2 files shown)


figtracer (approved):
makes sense

mattsse (approved):
ah yes, this is correct, naming is terrible here

PR #14430, opened 2026-04-23
fix(anvil): fix flaky test_increase_time_by_zero test

figtracer:
## Problem

`test_increase_time_by_zero` is flaky because it calls `evm_increase_time(U256::ZERO)` and then asserts that the mined block has the **exact same** timestamp as the previous block.

However, `evm_increase_time(0)` only adds 0 to the time offset — it does not pin an exact next-block timestamp. When `compute_next_timestamp()` runs, it computes `wall_clock + offset`, so if even 1 second of wall-clock time elapses between node spawn and mining, the next timestamp is `init_timestamp + 1` and the assertion fails.

Introduced in [`4d7435e`](https://github.com/foundry-rs/foundry/commit/4d7435e64ba1d351d128be3b1a30e6d6b246696a) (PR #9160).

## Fix

Replace `evm_increase_time(U256::ZERO)` with `evm_set_next_block_timestamp(init_timestamp)`, which deterministically pins the next block's timestamp. This matches the intent of the test and is consistent with the sibling test `test_mine_blk_with_prev_timestamp` right below it.

### DIFF
--- a/crates/anvil/tests/it/anvil_api.rs
+++ b/crates/anvil/tests/it/anvil_api.rs
@@ -1003,7 +1003,6 @@ async fn test_mine_blk_with_prev_timestamp() {
 }
 
 // increase time by 0 seconds i.e next_block_timestamp = prev_block_timestamp
-// api.evm_increase_time(0).unwrap();
 #[tokio::test(flavor = "multi_thread")]
 async fn test_increase_time_by_zero() {
     let (api, handle) = spawn(NodeConfig::test()).await;
@@ -1014,7 +1013,11 @@ async fn test_increase_time_by_zero() {
     let init_number = init_blk.header.number;
     let init_timestamp = init_blk.header.timestamp;
 
-    let _ = api.evm_increase_time(U256::ZERO).await;
+    // `evm_increase_time(0)` adds zero to the offset but does not pin an exact
+    // next-block timestamp, so wall-clock drift can cause the mined block to be
+    // 1 s ahead.  Use `evm_set_next_block_timestamp` to deterministically keep
+    // the same timestamp as the previous block.
+    api.evm_set_next_block_timestamp(init_timestamp).unwrap();
 
     api.mine_one().await;
 
--- a/crates/anvil/tests/it/anvil_api.rs
+++ b/crates/anvil/tests/it/anvil_api.rs
@@ -1013,7 +1013,7 @@ async fn test_increase_time_by_zero() {
     let init_number = init_blk.header.number;
     let init_timestamp = init_blk.header.timestamp;
 
-api.evm_set_next_block_timestamp(init_timestamp).unwrap();
+    api.evm_set_next_block_timestamp(init_timestamp).unwrap();
 
     api.mine_one().await;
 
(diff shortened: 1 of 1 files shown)


crates/anvil/tests/it/anvil_api.rs
```diff
@@ -1014,7 +1013,11 @@ async fn test_increase_time_by_zero() {
     let init_number = init_blk.header.number;
     let init_timestamp = init_blk.header.timestamp;
 
-    let _ = api.evm_increase_time(U256::ZERO).await;
+    // `evm_increase_time(0)` adds zero to the offset but does not pin an exact
```
stevencartavia: this comment without the PR context wouldn't make sense, let's just keep the comment for `api.evm_set_next_block_timestamp(init_timestamp).unwrap();`

crates/anvil/tests/it/anvil_api.rs
```diff
@@ -1014,7 +1013,11 @@ async fn test_increase_time_by_zero() {
...
+    // next-block timestamp, so wall-clock drift can cause the mined block to be
+    // 1 s ahead.  Use `evm_set_next_block_timestamp` to deterministically keep
+    // the same timestamp as the previous block.
+    api.evm_set_next_block_timestamp(init_timestamp).unwrap();
```
figtracer, line 1016:
```suggestion
api.evm_set_next_block_timestamp(init_timestamp).unwrap();
```

crates/anvil/tests/it/anvil_api.rs
figtracer: i'll just get rid of it, not doing much here

crates/anvil/tests/it/anvil_api.rs
```diff
@@ -1014,7 +1013,11 @@ async fn test_increase_time_by_zero() {
     let init_number = init_blk.header.number;
     let init_timestamp = init_blk.header.timestamp;
 
-    let _ = api.evm_increase_time(U256::ZERO).await;
+    // `evm_increase_time(0)` adds zero to the offset but does not pin an exact
+    // next-block timestamp, so wall-clock drift can cause the mined block to be
+    // 1 s ahead.  Use `evm_set_next_block_timestamp` to deterministically keep
+    // the same timestamp as the previous block.
+api.evm_set_next_block_timestamp(init_timestamp).unwrap();
```
figtracer:
```suggestion
api.evm_set_next_block_timestamp(init_timestamp).unwrap();
```

stevencartavia (dismissed):

crates/anvil/tests/it/anvil_api.rs
```diff
@@ -1014,7 +1013,7 @@ async fn test_increase_time_by_zero() {
     let init_number = init_blk.header.number;
     let init_timestamp = init_blk.header.timestamp;
 
-    let _ = api.evm_increase_time(U256::ZERO).await;
+api.evm_set_next_block_timestamp(init_timestamp).unwrap();
```
figtracer:
```suggestion
    api.evm_set_next_block_timestamp(init_timestamp).unwrap();
```

stevencartavia (approved):

mattsse (approved):

PR #14431, opened 2026-04-23
feat(lint): add block-timestamp lint

stevencartavia:
Adds a low-severity lint that warns when `block.timestamp` is used in a comparison, as validators can manipulate it.

## Testing 
`cargo run -p forge -- lint crates/lint/testdata/BlockTimestamp.sol`


crates/lint/src/sol/info/block_timestamp.rs
```diff
@@ -0,0 +1,68 @@
...
+            contains_block_timestamp(lhs) || contains_block_timestamp(rhs)
+        }
+        ExprKind::Tuple(elems) => elems.iter().any(|e| {
+            if let solar::interface::SpannedOption::Some(inner) = e.as_ref() {
+                contains_block_timestamp(inner)
+            } else {
+                false
+            }
+        }),
+        ExprKind::Call(callee, _) => contains_block_timestamp(callee),
```
mablr: This seems to drop all args, `foo(block.timestamp) > 0` may not detected ?

mablr (approved):

mattsse (approved):

Issue #10552, opened 2025-05-19
bug(`forge test`): panic crash on foundry v1.1+


0xTimepunk:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

forge Version: 1.2.0-dev Commit SHA: 4107ae0f51a4e341ace1248be33b40d4256b58b5 Build Timestamp: 2025-05-19T13:08:27.408307000Z (1747660107) Build Profile: release

### What version of Foundryup are you on?

foundryup: 1.0.1

### What command(s) is the bug in?

forge test --jobs 10

### Operating System

macOS (Apple Silicon)

### Describe the bug

Whenever test or coverage is ran, there is a random crash. This crash is present in version 1.1. By suggestion of @grandizzy updated to https://github.com/foundry-rs/foundry/pull/10546 with `foundryup --pr 10546`because we thought usage of vm.cool was the problem (however there is no direct usage of it anywhere in our /src, but we can't say for sure in our libs)

We were using ffi to call apis with SURL , but this is now disabled through the codebase (mocks now used) and the bug is still present:

```
Suite result: ok. 12 passed; 0 failed; 0 skipped; finished in 4.89s (1.45ms CPU time)
The application panicked (crashed).
Message:  called `Option::unwrap()` on a `None` value
Location: /Users/timepunk/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/revm-19.7.0/src/journaled_state.rs:402

This is a bug. Consider reporting it at https://github.com/foundry-rs/foundry

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                                ⋮ 9 frames hidden ⋮                               
  10: core::panicking::panic::h49eb163a6c085f1c
      at <unknown source file>:<unknown line>
  11: core::option::unwrap_failed::h2ae15f0d32988266
      at <unknown source file>:<unknown line>
  12: <core::iter::adapters::rev::Rev<I> as core::iter::traits::iterator::Iterator>::try_fold::h2a57210fa4f26ea6
      at <unknown source file>:<unknown line>
  13: revm::handler::mainnet::execution::call_return::h4601ab5cf339bcbf
      at <unknown source file>:<unknown line>
  14: core::ops::function::Fn::call::h26670057e514e122
[...]

PR #14432, opened 2026-04-23
refactor(anvil): pass `from` explicitly into `transaction_build`

AissataF:
Currently `transaction_build` recovers `from` from the signature, but impersonated transactions use a bypass signature whose recovery returns the zero address. Four callers worked around this by post-patching with `Recovered::new_unchecked` afterwards. Making `from` an explicit `Option<Address>` parameter drops those four patches, and the Tempo path picks up the same impersonation fix along the way.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -980,7 +980,9 @@ impl<N: Network> Backend<N> {
             let info = storage.transactions.get(&hash)?.info.clone();
             let tx = block.body.transactions.get(info.transaction_index as usize)?.clone();
 
-            let tx = transaction_build(Some(hash), tx, Some(block), Some(info), base_fee);
+            let from = info.from;
+            let tx =
+                transaction_build(Some(hash), Some(from), tx, Some(block), Some(info), base_fee);
             transactions.push(tx);
         }
         Some(transactions)
@@ -1685,6 +1687,7 @@ impl<N: Network> Backend<N> {
 
         Some(transaction_build(
             Some(info.transaction_hash),
+            Some(info.from),
             tx,
             Some(&block),
             Some(info),
@@ -1723,6 +1726,7 @@ impl<N: Network> Backend<N> {
 
         Some(transaction_build(
             Some(info.transaction_hash),
+            Some(info.from),
             tx,
             Some(&block),
             Some(info),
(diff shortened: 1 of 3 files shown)


mablr (commented):
The scoping of this PR seems a bit off to me.

I see 3 logically independent changes:
- evm fork-state sync (forge script/cast run...)
- anvil's transaction_build refactor
- anvil RPC param typing (seems unrelated)

Btw I'm unsure we really need this extra sync on fork selection as I believe it's already handled downstream... could you please share a test/use-case ?

AissataF:
Sorry @mablr, kinda forgot the context after 2 weeks  :|                                 
           
What I was thinking back then: the loop at L1184-L1209 (51f6c8ee, for #10552) only updates       `present_value` for slots already in the target journal and skips `info`, looked like a real gap. So tried to fix it, and threw in some unrelated anvil stuff I noticed nearby.                           
                              
Just re-read it — yeah, doesn't really hold up. `update_fork_db` at L1237 already does the bigger merge via `merge_account_data` (info via `state.insert`, new slots via `storage.extend`). Whatever my loop writes mostly gets overwritten right after.                                                    
                              
So:
  1. Drop the fork-sync change — no repro, downstream covers it.
  2. Keep the `transaction_build` refactor (pass `from` in explicitly, removes 4
  `Recovered::new_unchecked` patches). Can retitle the PR around that.                                 
  3. New PR for the `U256` RPC params (`setIntervalMining` etc. reject hex-string params today).
                                                                                                        
WDYT?

mablr:
Yes, please retarget this PR 👍

AissataF:
> Yes, please retarget this PR 👍

All set. and new pr created too in https://github.com/foundry-rs/foundry/pull/14650

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -4621,11 +4626,19 @@ where
 /// Creates a `AnyRpcTransaction` as it's expected for the `eth` RPC api from storage data
 pub fn transaction_build(
     tx_hash: Option<B256>,
+    from: Option<Address>,
```
mablr: The placement of the `from` arg before tx itself seems odd.

mablr (commented):
Could you please a regression test?

crates/anvil/src/eth/backend/mem/mod.rs
AissataF: Oh nice, moved from after the transaction arg.

PR #14433, opened 2026-04-24
feat: log broken invariant as soon as it is found

### LINKED ISSUE #13285: Log broken invariant as soon as it is found

stevencartavia:
## Motivation
closes #13285

## Solution
- Emit a structured `"failure"` JSON event immediately when an invariant breaks, with timestamp, invariant name, target contract, and revert reason
- Add `failures` and `unique_failures` counters to periodic pulse metrics (will be non-zero once #12587 lands)
- Add `"event": "pulse"` field to periodic metrics for uniform event dispatch

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/evm/src/executors/invariant/mod.rs
+++ b/crates/evm/evm/src/executors/invariant/mod.rs
@@ -194,13 +194,17 @@ fn build_invariant_progress_json<M: Serialize>(
     failure_metrics: &InvariantFailureMetrics,
     elapsed: Duration,
 ) -> serde_json::Value {
+    let mut metrics = serde_json::to_value(corpus_metrics).unwrap_or_default();
+    if let Some(obj) = metrics.as_object_mut() {
+        obj.insert("failures".to_string(), json!(failure_metrics.failures));
+        obj.insert("unique_failures".to_string(), json!(failure_metrics.unique_failures.len()));
+    }
+
     let mut payload = json!({
         "timestamp": timestamp_secs,
         "event": "pulse",
         "invariant": invariant_name,
-        "metrics": corpus_metrics,
-        "failures": failure_metrics.failures,
-        "unique_failures": failure_metrics.unique_failures.len(),
+        "metrics": metrics,
         "total_txs": throughput.total_txs,
         "total_gas": throughput.total_gas,
         "tx_per_sec": throughput.tx_per_sec(elapsed),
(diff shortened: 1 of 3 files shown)


crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -165,12 +191,16 @@ fn build_invariant_progress_json<M: Serialize>(
...
     throughput: InvariantThroughputMetrics,
+    failure_metrics: &InvariantFailureMetrics,
     elapsed: Duration,
 ) -> serde_json::Value {
     let mut payload = json!({
         "timestamp": timestamp_secs,
+        "event": "pulse",
         "invariant": invariant_name,
         "metrics": corpus_metrics,
+        "failures": failure_metrics.failures,
```
grandizzy: let's put these in metrics as suggested in https://github.com/foundry-rs/foundry/issues/13285#issuecomment-3849903066

grandizzy (changes requested):
looks good, ty, minor nit re payload to match proposed one

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -153,6 +153,32 @@ fn rate_per_sec(total: f64, elapsed: Duration) -> f64 {
     if elapsed_secs > 0.0 { total / elapsed_secs } else { 0.0 }
 }
 
+/// Tracks invariant failure counts during a campaign.
+#[derive(Debug, Default)]
+struct InvariantFailureMetrics {
+    failures: u64,
+    unique_failures: HashSet<String>,
```
mablr, line 160: This seems to dedup by invariant name, what if two different have an invariant with the same name ?

mablr (commented):
lgtm, just a minor concern

crates/evm/evm/src/executors/invariant/mod.rs
grandizzy: there can be invariants with same name but only in different contracts (an invariant is defined by solidity fn) but since we run each invariant campaign independently per (contract, function) pair, then they cannot collide

grandizzy (approved):
lgtm!

mablr (approved):

PR #14350, opened 2026-04-16
refactor(evm): remove useless Eth/Tempo EVM wrapper structs

mablr:
## Motivation

#14344 #14346 follow-up

Remove `EthFoundryEvm` and `TempoFoundryEvm` wrapper structs that manually reimplemented the `Evm` to use former `FoundryHandler`.

Also cleans up imports across eth.rs, tempo.rs, op.rs, and mod.rs, and fixes minor API updates (`ctx() -> ctx_ref()`, `evm_clone()`).

I haven't removed yet OP's wrapper as OpContext/OpEvm are currently messed-up upstream due to `OpTx`

Less code !😎

PR #14440, opened 2026-04-24
refactor(evm): remove useless OP EVM wrapper

mablr:
## Motivation

Now `alloy-op-evm` and `op-revm` depend on `revm` >37 we can remove useless OP EVM wrapper, in the same vein as https://github.com/foundry-rs/foundry/pull/14350.


0xrusowsky (approved):

figtracer (approved):

zerosnacks (approved):
Yes, very nice!

PR #14444, opened 2026-04-24
perf(fuzz): eagerly seed AST literals and remove per-iteration OnceLock blocking

zerosnacks:
## Summary

Addresses a suspected ~21% wall-time regression on numeric-heavy fuzz suites (e.g. Solady) introduced by the ast-seeded dictionary feature (#12015). See [benchmark gist](https://gist.github.com/figtracer/9ef6187a1316b181bc16fab9789a9eff).

## Root cause analysis

`FuzzDictionary::samples()` called `OnceLock::wait()` on **every fuzz iteration**, and returned only AST literals (ignoring `sample_values`) until a write operation triggered lazy seeding. For numeric-heavy suites like Solady:

1. **Startup cliff**: The background literal-collector thread may not have finished, so `wait()` blocks the first iterations.
2. **Changed distribution**: ~20% of numeric args were pulled from AST literal buckets instead of generic state words, hitting deeper/slower EVM execution paths.
3. **Per-iteration overhead**: Even after init, `wait()` performs an atomic check on every `samples()` call.

## Fix

- **Block once** in `EvmFuzzState::new()` to merge all AST literals into `sample_values` before any fuzz iteration runs.
- `samples()` is now a plain `HashMap::get()` — no `OnceLock` access on the hot path.
- Add non-blocking `try_seed_samples()` for write-path callers (`insert_storage_value`, `insert_sample_values`, `insert_typed_cmp_value`).
- Add `LiteralsDictionary::try_get()` for non-blocking `OnceLock` access.

## Quality

No degradation — AST literals are still fully present in the sample pool. The merged `sample_values` map is actually **richer** (AST literals + runtime values) than the pre-seeded path which returned only AST literals.

## Status

**Not yet verified with benchmarks.** Based on code analysis of the hot path.

### DIFF
--- a/crates/evm/fuzz/src/strategies/literals.rs
+++ b/crates/evm/fuzz/src/strategies/literals.rs
@@ -57,9 +57,14 @@ impl LiteralsDictionary {
         Self { maps }
     }
 
-    /// Returns a reference to the `LiteralMaps`.
+    /// Returns a reference to the `LiteralMaps`, blocking until ready.
     pub fn get(&self) -> &LiteralMaps {
         self.maps.wait()
+    }
+
+    /// Returns a reference to the `LiteralMaps` if already initialized, without blocking.
+    pub fn try_get(&self) -> Option<&LiteralMaps> {
+        self.maps.get()
     }
 
     /// Test-only helper to seed the dictionary with literal values.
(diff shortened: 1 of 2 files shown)


zerosnacks:
Made no significant difference, closing

PR #14445, opened 2026-04-24
feat: immutable releases

zerosnacks:
## Summary

Switches Foundry to fully immutable releases. No more mutable Git tags (`nightly`, `stable`, `rc`) that get force-moved on each release.

### Release workflow (`release.yml`)
- Removed mutable tag triggers (`stable`, `rc`, `rc-*`) — only `v*.*.*` semver tags trigger releases
- Removed the "Update nightly release" step that overwrote the mutable `nightly` GitHub release
- Removed the "Move nightly tag" step that force-moved the `nightly` tag to HEAD
- Removed `LAST_STABLE_VERSION` env var
- Removed `cleanup` job — no more pruning of old nightlies (immutable means permanent)
- Each nightly gets its own immutable `nightly-{SHA}` release only
- Changelog `fromTag` for stable releases now explicitly resolves to the previous `v*.*.*` tag, preventing nightlies from being picked as the base

### Docker (`docker-publish.yml`)
- Simplified tagging to derive directly from `inputs.tag_name`
- Fixes `workflow_call` bug where nightlies got mis-tagged (`event_name` is `workflow_call`, not `schedule`)
- Version tags get `:v1.2.3`, `:v1.2`, `:v1` (standard Docker convention); everything else (including `nightly-{SHA}`) tagged as-is

### foundryup (bash)
- `latest` is now the default channel — uses `/releases/latest` GitHub API endpoint
- `stable` is a silent alias for `latest`
- `nightly` resolves to the latest `nightly-{SHA}` prerelease via GitHub API (`per_page=10`), excludes drafts
- Specific versions (`v1.6.0`, `nightly-{SHA}`) still work directly
- Added `fetch()` helper for silent HTTP requests (used by release resolution and version check)
- Inlined release resolution at call sites, removed `resolve_latest_release()` function
- Improved logging: fetch/resolve steps are now visible to the user
- Fixed `set -eo pipefail` compatibility in fetch pipelines
- `check_installer_up_to_date()` now uses `fetch()` helper
- Bumped installer version to 1.8.0
- Updated README with `foundryup -i latest` alternative

### Scripts
- `.github/scripts/create-tag.js` — removed `force: true` from `createRef`; now only swallows "already exists" errors (422) and rethrows unexpected failures
- `.github/scripts/move-tag.js` — deleted
- `.github/scripts/prune-prereleases.js` — deleted

## User-facing behavior

| Command | Behavior |
|---|---|
| `foundryup` | Installs latest release (resolved via `/releases/latest`) |
| `foundryup -i latest` | Same as above |
| `foundryup -i stable` | Silent alias for `latest` |
| `foundryup -i nightly` | Installs latest nightly (resolved via API) |
| `foundryup -i nightly-abc123...` | Installs that specific nightly |
| `foundryup -i v1.6.0` | Installs that exact version |

## ⚠️ Note on current `/releases/latest`

The `foundryup` default (`latest`) resolves via GitHub's `/releases/latest` endpoint, which returns the most recent non-prerelease, non-draft release. Currently that is **`v1.6.0-rc1`** — this will remain the default until `v1.7.0` is tagged as a full release. This is expected behavior: GitHub considers `v1.6.0-rc1` the latest release because it was not marked as a prerelease.

## Migration

Old mutable tags (`nightly`, `stable`) will be frozen in place — not deleted. Old foundryup (v1.7.0) will still work with stale binaries and the existing `check_installer_up_to_date()` will prompt users to run `foundryup --update`.

## Companion PR

- foundryup Rust rewrite: https://github.com/foundry-rs/foundryup/pull/54


.github/workflows/release.yml
```diff
@@ -71,7 +70,7 @@ jobs:
         uses: mikepenz/release-changelog-builder-action@bcae7115752d4ed746ff92feb666574428a79415 # v6.2
         with:
           configuration: "./.github/changelog.json"
-          fromTag: ${{ env.IS_NIGHTLY == 'true' && 'nightly' || env.STABLE_VERSION }}
+          fromTag: ${{ steps.release_info.outputs.from_tag || '' }}
```
mablr, line 77: `fromTag` for nightlies becomes empty, should be fine ?

mablr (commented):
Sgtm.

> - `stable` is a silent alias for `latest`

Does Docker's `stable` tag become stale after this ?

figtracer (dismissed):
lgtm

.github/workflows/release.yml
zerosnacks: updated, now on nightly it uses the previous nightly tag by finding it

zerosnacks:
> Sgtm.
> 
> > * `stable` is a silent alias for `latest`
> 
> Does Docker's `stable` tag become stale after this ?

The old workflow emitted `:nightly` and `:latest` but not `:stable`, I've updated the workflow so it does mutably tag `:nightly` and `:latest` for Docker for convenience again.

mablr (dismissed):

foundryup/foundryup
```diff
@@ -110,16 +110,43 @@ main() {
...
+      ') || err "failed to fetch releases from GitHub API"
+      if [ -z "$FOUNDRYUP_TAG" ]; then
+        err "could not find a latest release for ${FOUNDRYUP_REPO}"
+      fi
+      say "resolved release tag: ${FOUNDRYUP_TAG}"
+      FOUNDRYUP_VERSION="$FOUNDRYUP_TAG"
+    elif [[ "$FOUNDRYUP_VERSION" == "nightly" ]]; then
+      # Resolve to the latest nightly (prerelease) release via the GitHub API
+      say "fetching latest nightly releases from ${FOUNDRYUP_REPO}..."
+      FOUNDRYUP_TAG=$(fetch "https://api.github.com/repos/${FOUNDRYUP_REPO}/releases?per_page=10" | awk '
```
figtracer: should we increase `per_page=10` to prevent missing releases or loop though pages

zerosnacks:
```suggestion
      FOUNDRYUP_TAG=$(fetch "https://api.github.com/repos/${FOUNDRYUP_REPO}/releases" | awk '
```

figtracer (dismissed):

foundryup/foundryup
zerosnacks: without specifying a limit it seems to fetch up to 30 which seems appropriate, updated

PR #14446, opened 2026-04-24
fix(anvil): refresh fee history cache on chain rebuilds

Perico-perica46:
## Motivation

`eth_feeHistory` and `eth_maxPriorityFeePerGas` read from `fee_history_cache`. `anvil_reset` and `anvil_loadState` can replace the active chain context without emitting new block notifications, which lets cached entries from the previous context remain visible.

## Solution

Clear the fee history cache after `anvil_reset` succeeds, and rebuild it from the current local backend after `anvil_loadState` succeeds. Add a regression test that loads state with a different genesis base fee and checks fee history reflects the loaded state.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/anvil/tests/it/gas.rs
```diff
@@ -217,6 +224,36 @@ async fn test_can_use_fee_history() {
...
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_fee_history_errors_on_cache_miss() {
+    let (api, _) = spawn(NodeConfig::test()).await;
+
+    let err =
+        api.fee_history(U256::from(1), BlockNumberOrTag::Number(1), vec![]).await.unwrap_err();
+
+    assert!(matches!(err, BlockchainError::FeeHistory(FeeHistoryError::InvalidBlockRange)));
+}
```
mablr: Codifies the breaking behavior change instead of covering the actual bug

An `anvil_reset` regression test may be more beneficial 
to assess it returns fresh fee history immediately after reset?

crates/anvil/src/eth/api.rs
```diff
@@ -605,6 +610,35 @@ impl<N: Network> EthApi<N> {
...
+    {
+        self.fee_history_cache.lock().clear();
+
+        let (_, new_blocks) = unbounded();
+        let fee_history_service = FeeHistoryService::new(
+            match self.backend.spec_id() {
+                SpecId::OSAKA => BlobParams::osaka(),
+                SpecId::PRAGUE => BlobParams::prague(),
+                _ => BlobParams::cancun(),
+            },
```
mablr: Instead of hardcoding, would it be more accurate to use `self.backend.blob_params()` to stay in sync with the live `FeeHistoryService`.?

crates/anvil/src/eth/api.rs
```diff
@@ -1153,24 +1188,26 @@ impl<N: Network> EthApi<N> {
...
-                        for p in &reward_percentiles {
-                            let p = p.clamp(0.0, 100.0);
-                            let index = ((p.round() / 2f64) * 2f64) * resolution_per_percentile;
-                            let reward = block.rewards.get(index as usize).map_or(0, |r| *r);
-                            block_rewards.push(reward);
-                        }
-                        rewards.push(block_rewards);
+                let Some(block) = fee_history.get(&n) else {
+                    return Err(FeeHistoryError::InvalidBlockRange.into());
+                };
```
mablr: Changing `fee_history()` to return `FeeHistoryError::InvalidBlockRange` on any single cache miss seems to be a breaking change unrelated to the stated motivation.

mablr (commented):
Good catch, some comments...

Perico-perica46:
@mablr Updated.

crates/anvil/src/eth/api.rs
mablr: I think `rebuild_fee_history_cache` should be also called on successful revert.

Perico-perica46:
@mablr Done.

crates/anvil/tests/it/gas.rs
```diff
@@ -217,6 +217,64 @@ async fn test_can_use_fee_history() {
...
+    let before = api.fee_history(U256::from(1), BlockNumberOrTag::Number(1), vec![]).await.unwrap();
+    assert_eq!(
+        before.base_fee_per_gas.first().copied(),
+        Some(reverted_block_base_fee.to::<u128>())
+    );
+
+    assert!(api.evm_revert(snapshot_id).await.unwrap());
+
+    let after = api.fee_history(U256::from(1), BlockNumberOrTag::Number(1), vec![]).await.unwrap();
+    assert_ne!(after.base_fee_per_gas.first().copied(), Some(reverted_block_base_fee.to::<u128>()));
```
mablr: After `evm_revert(snapshot_id)` the chain is back at block 0. Querying `fee_history(1, BlockNumberOrTag::Number(1))` targets a block that no longer exists, so `first()` returns `None`, making `assert_ne!(…, Some(1000))` always pass regardless of whether the cache was refreshed. Should use `Latest` instead ?

crates/anvil/src/eth/api.rs
```diff
@@ -605,6 +609,31 @@ impl<N: Network> EthApi<N> {
...
+    {
+        self.fee_history_cache.lock().clear();
+
+        let (_, new_blocks) = unbounded();
+        let fee_history_service = FeeHistoryService::new(
+            self.backend.blob_params(),
+            new_blocks,
+            self.fee_history_cache.clone(),
+            self.storage_info(),
+        );
```
mablr: A full `FeeHistoryService` is constructed solely to call `insert_cache_entry_for_block`, with its sender half immediately dropped (channel closed before construction). It's never polled and is dropped at end of scope. Functionally correct because the live `FeeHistoryService` keeps running and handles future blocks. But bulding a struct that implements `Future` with a dead channel  just to borrow one of its methods is confusing. A cleaner path would be to make `insert_cache_entry_for_block` (or a thin wrapper around `create_cache_entry` + `insert_cache_entry`) a free function or an associated function on `FeeHistoryCache` directly?

Perico-perica46:
@mablr Done.

mablr (approved):
Sgtm, thanks! Pending another approval...

mattsse (changes requested):
I dont really like how this looks, 

a few too many changes and I'm skeptical that all of those are really needed, converting a bunch of logic into standalone functions

Perico-perica46:
@mattsse Yeah, fair. I can trim this down.

I’ll keep the behavior change and tests, but avoid pulling the FeeHistoryService internals out into standalone helpers. I’ll make the cache rebuild path more local and only keep the minimum bit needed to avoid constructing a dead FeeHistoryService.

Would that work?

PR #14447, opened 2026-04-24
feat(doc): migrate from solang to solar

mablr:
## Motivation

Replace all `solang_parser` usage in forge-doc with solar AST. Use owned purpose-built structs (ParamInfo, ContractSource, FunctionSource, etc.) extracted at walk time instead of cloning arena-allocated solar AST nodes.

- Add parser/source.rs with owned doc structs
- Rewrite Parser to walk solar AST items directly
- Parse doc comments from solar's DocComments (raw text)
- Update builder to use solar AST from gcx, remove solang re-parse
- Update writer and preprocessor layers for owned structs
- Delete solang_ext/ module (Visitor, Visitable, AstEq, etc.)
- Remove solang-parser dependency from forge-doc
- Decouple chisel from forge-doc (inline CodeLocation/SafeUnwrap)

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


DaniPopes:
would be nicer to rewrite from scratch and have an actual test harness and suite for this, but i guess at this point we can leave that for the future.

crates/doc/src/writer/as_doc.rs
```diff
@@ -315,12 +310,20 @@ impl Document {
     fn write_function(
```
stevencartavia, line 310: this duplicates the logic in `ParseSource::signature()`. Maybe a `FunctionSource::signature()` method can be added and both can share it.

stevencartavia (commented):
lgtm, just a minor nit

crates/doc/src/writer/as_doc.rs
mablr: done 🫡

stevencartavia (dismissed):

crates/doc/src/parser/mod.rs
```diff
@@ -23,206 +16,276 @@ pub use item::{ParseItem, ParseSource};
...
+                Some(
+                    ParseItem::new(ParseSource::Variable(source))
+                        .with_comments(docs)
+                        .with_code(code),
+                )
+            }
+            ast::ItemKind::Event(event) => {
+                let source = self.parse_event(event);
+                let mut code = self.extract_code(span);
+                code.push(';');
```
figtracer: Solar already does this so this will render `;;`

figtracer (approved):

zerosnacks (dismissed):
Nice!

figtracer (approved):

crates/doc/Cargo.toml
```diff
@@ -29,7 +28,6 @@ mdbook-driver = { version = "0.5", default-features = false, features = ["search
 rayon.workspace = true
 serde_json.workspace = true
 serde.workspace = true
-solang-parser.workspace = true
```
zerosnacks, line 32: I think we can then also remove the top level solang-parser @mablr 

https://github.com/foundry-rs/foundry/blob/51e9feb51795b2acb6fdb99f36a493a978e78b7f/Cargo.toml#L352

https://github.com/foundry-rs/foundry/blob/51e9feb51795b2acb6fdb99f36a493a978e78b7f/Cargo.toml#L177

zerosnacks: Or is it still required in Chisel? If so, can we also tackle that in this PR?

zerosnacks (approved):
Lets fully deprecate Solang in Chisel in a follow up, PR lgtm as is

PR #4792, opened 2023-04-21
fix: Fix self signed certificates in chain by using correct Reqwest feature

revblock:
This change fixes the bug reported in https://github.com/foundry-rs/foundry/issues/4446 by using the feature specified in Reqwest's documentation specified here https://docs.rs/reqwest/latest/reqwest/#optional-features

Motivation
This fixes the original fix from this issue https://github.com/foundry-rs/foundry/issues/2873 in this PR https://github.com/foundry-rs/foundry/pull/2939

Using the functionality of rustls-native-certs such as using the environment variable to include your certificate had no effect and caused requests to still fail.

Solution
Reqwest's documentation providers a different string for the feature than what was originally added. Built and tested locally with the SSL_CERT_FILE environment variable and it now picks it up and works.

PR #14451, opened 2026-04-24
feature: native-tls

jorgemmsilva:
adds a `native-tls` build flag. This is helpful for people that must use cast behind corporate proxies.

tried this solution, but it didn't work for me:
https://github.com/foundry-rs/foundry/pull/4792


crates/forge/Cargo.toml
```diff
@@ -126,3 +126,4 @@ aws-kms = ["foundry-wallets/aws-kms"]
 gcp-kms = ["foundry-wallets/gcp-kms"]
 turnkey = ["foundry-wallets/turnkey"]
 isolate-by-default = ["foundry-config/isolate-by-default"]
+native-tls = ["foundry-cli/native-tls"]
```
mablr: Logically, `"foundry-block-explorers/openssl"` should also be included here?

mablr (commented):
Makes sense as it’s a pattern we also have in alloy and foundry-core. @zerosnacks wdyt?

Btw some wiring is certainly missing.

mablr:
`openssl` is banned if not wrapped, use in `deny.toml`:
```
deny = [{ name = "openssl", wrappers = ["native-tls"] }]
```

mablr (dismissed):
Makes sense, pending @zerosnacks :)

zerosnacks:
https://github.com/foundry-rs/foundry/pull/14562 may also have an impact here

crates/cast/Cargo.toml
```diff
@@ -41,11 +41,11 @@ alloy-json-rpc.workspace = true
...
-    "reqwest",
-    "ws",
-    "ipc",
-    "trace-api",
-    "txpool-api",
+  "reqwest",
+  "ws",
+  "ipc",
+  "trace-api",
+  "txpool-api",
```
mablr, line 48: Unsure these fmt changes should be in this PR ?

jorgemmsilva: just my auto-formatter. I recommend setting up some CI workflow to assert the toml files formatting, if that's important.

example: https://github.com/SyndicateProtocol/syndicate-appchains/blob/295d80b68b27f3f4cb47b6b68df1db82a7678902/.github/workflows/rust-ci.yaml#L43-L49

PR #14452, opened 2026-04-24
chore(cli): make `Git` impl more robust

mablr:
## Motivation

Initially, I intended to replace the `git` command usage with the `gix` library (as suggested in https://github.com/foundry-rs/foundry/issues/13501). However, it doesn’t address all the requirements, and the `git` binary will still be necessary (refer to https://github.com/foundry-rs/foundry/commit/6bb4ea90686673879779dbce63c3b75f776b6240). This approach introduces an additional layer of abstraction without providing substantial benefits.

However I extracted here some clean-ups making `foundry_cli::utils::Git` a bit more elegant and robust:
- some dedup
- made methods spawning git command private

### DIFF
--- a/crates/forge/src/cmd/update.rs
+++ b/crates/forge/src/cmd/update.rs
@@ -126,7 +126,7 @@ impl UpdateArgs {
             let name = dep_id.name();
 
             // Fetch and checkout the latest commit from the remote branch
-            git.fetch_and_checkout_branch(path, name)?;
+            git.fetch_and_checkout_branch(&submodule_path, name)?;
 
             // Now get the updated revision after syncing with origin
             let (updated_rev, _) = git.current_rev_branch(&submodule_path)?;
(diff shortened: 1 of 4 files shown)


zerosnacks (dismissed):
Nice 👍

crates/forge/src/cmd/update.rs
```diff
@@ -126,7 +126,7 @@ impl UpdateArgs {
             let name = dep_id.name();
 
             // Fetch and checkout the latest commit from the remote branch
-            Self::fetch_and_checkout_branch(&git, &submodule_path, name)?;
+            git.fetch_and_checkout_branch(path, name)?;
```
stevencartavia: shouldn't this be `git.fetch_and_checkout_branch(&submodule_path, name)?;`?

mablr: good catch ty

DaniPopes (approved):

stevencartavia (approved):

PR #14454, opened 2026-04-24
fix(config): disable proxy detection in Etherscan client

AissataF:
`foundry-block-explorers` defaults to `reqwest::Client::default()`, which triggers system proxy lookup via `SCDynamicStore` and crashes under macOS App Sandbox / Cursor IDE. Build a `reqwest::Client` with `no_proxy()` and pass it through `with_client`, Note: `foundry-block-explorers` pins `reqwest = "0.12"`, so `foundry-config` picks up the same version to keep `Client` types compatible.

### DIFF
--- a/crates/config/Cargo.toml
+++ b/crates/config/Cargo.toml
@@ -37,6 +37,7 @@ mesc.workspace = true
 unit-prefix = "0.5.2"
 rayon.workspace = true
 regex.workspace = true
+reqwest = { version = "0.12", default-features = false }
 semver = { workspace = true, features = ["serde"] }
 serde_json.workspace = true
 serde.workspace = true
--- a/crates/config/Cargo.toml
+++ b/crates/config/Cargo.toml
@@ -37,7 +37,7 @@ mesc.workspace = true
 unit-prefix = "0.5.2"
 rayon.workspace = true
 regex.workspace = true
-reqwest = { version = "0.12", default-features = false }
+reqwest.workspace = true
 semver = { workspace = true, features = ["serde"] }
 serde_json.workspace = true
 serde.workspace = true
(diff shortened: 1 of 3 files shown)


mablr (commented):
Makes sense.

Maybe it would be cleaner to add `no_proxy` directly to `foundry_block_explorers::ClientBuilder` (in foundry-core repo) mirroring `reqwest::ClientBuilder`? (would remove `reqwest` dep here)

crates/config/Cargo.toml
```diff
@@ -37,6 +37,7 @@ mesc.workspace = true
 unit-prefix = "0.5.2"
 rayon.workspace = true
 regex.workspace = true
+reqwest = { version = "0.12", default-features = false }
```
mablr: `foundry_block_explorers`was bumped here 1ea7425858082a8c5c8baa248dd3c83a994d476c and now depends on 0.13

AissataF: wait for https://github.com/foundry-rs/foundry-core/pull/51 and fix them together.

AissataF: Should we use `.with_client` here for the first place and optimize it later when the upstream fixed?  @mablr

mablr: Let’s use `with_client()` for now i don’t think we’ll cut a release for such small change.

AissataF: rebased and apply some changes @mablr, good to go this time?

mablr (approved):
lgtm

mablr (commented):
g2g 👍 pending another approval @figtracer or @zerosnacks ;) ?

figtracer (approved):
nice

PR #14460, opened 2026-04-26
feat(lint): add missing-zero-check

stevencartavia:
Adds a `missing-zero-check` lint that flags address parameters of externally-callable functions and constructors used in state writes or value transfers `(call/delegatecall/staticcall/transfer/send)` without a prior `require/assert/if` check.

## Testing
`cargo run -p forge -- lint crates/lint/testdata/MissingZeroCheck.sol`


crates/lint/src/sol/info/missing_zero_check.rs
```diff
@@ -0,0 +1,390 @@
...
+    let inner = match &callee.kind {
+        ExprKind::Call(inner, ..) => inner,
+        _ => callee,
+    };
+    let target = if matches!(inner.kind, ExprKind::Member(..)) { inner } else { callee };
+    if let ExprKind::Member(receiver, name) = &target.kind {
+        let n = name.name;
+        if n == kw::Call
+            || n == kw::Delegatecall
+            || n == kw::Staticcall
```
mablr: `Staticcall` is read-only, unsure it makes sense to flag it.

crates/lint/src/sol/info/missing_zero_check.rs
```diff
@@ -0,0 +1,390 @@
...
+                for res in *reses {
+                    if let Res::Item(ItemId::Variable(vid)) = res
+                        && let Some(srcs) = self.taint.get(vid)
+                    {
+                        if self.guard_depth > 0 {
+                            self.guarded.extend(srcs.iter().copied());
+                        }
+                        if self.sink_depth > 0 {
+                            self.sinks.extend(srcs.iter().copied());
+                        }
```
mablr, line 326: Guard ordering is not enforced.

This seems to be a false negative:

```
function guardAfterSink(address a) external {
    owner = a; // sink: 'a' written to state
    require(a != address(0)); // guard too late, but marks 'a' as guarded anyway
} 
```

crates/lint/src/sol/info/missing_zero_check.rs
```diff
@@ -0,0 +1,389 @@
...
+        match stmt.kind {
+            StmtKind::If(cond, then, else_) => {
+                self.guard_depth += 1;
+                let _ = self.visit_expr(cond);
+                self.guard_depth -= 1;
+                let _ = self.visit_stmt(then);
+                if let Some(e) = else_ {
+                    let _ = self.visit_stmt(e);
+                }
+                return ControlFlow::Continue(());
```
mablr, line 244: Guard is not path sensitive, idk if there's a simple way to fix it, at least we should document this limitation.

This should trigger a warning, but it doesn't
```
function guardOnOneBranch(address a, bool flag) external {
    if (flag) {
        require(a != address(0)); // only guards when flag==true
    }
    owner = a; // sink reachable unchecked when flag==false
}
```

mablr (approved):
lgtm

zerosnacks (approved):

PR #14466, opened 2026-04-27
fix(anvil): honor epoch slots for safe/finalized tags

solanaXpeter:
## Motivation

Anvil exposes `--slots-in-an-epoch`, but some safe/finalized block tag paths still used a hard-coded 32-slot epoch. With a custom value, `eth_feeHistory` and block lookup could resolve different heights for the same tag.

## Solution

Route `eth_feeHistory` through the backend block-number conversion and pass the configured epoch length when resolving block hashes. Add a regression test that mines with `slots_in_an_epoch = 2` and checks safe/finalized tags for both block lookup and fee history.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mablr (dismissed):
Lgtm, thanks!

Pending another review.

figtracer (approved):
nice

mablr (approved):

PR #14467, opened 2026-04-27
chore: update latest benchmarks

mablr:
Updated benchmarks 1.5.1 (stable) vs nightly

Added AAVE v4 to the tested repos.


zerosnacks (approved):
We can swap the `nightly` out for `v1.7.0` after release

PR #14468, opened 2026-04-27
ci: split MPP e2e into its own workflow

zerosnacks:
Moves the `Run MPP e2e test` step out of `ci-tempo.yml` and into a new standalone `ci-mpp.yml` workflow.

Motivation: the MPP RPC has been returning transient HTTP 402 "Payment verification failed" errors (e.g. https://github.com/foundry-rs/foundry/actions/runs/24991029126/job/73176371269?pr=14445), which currently fails the entire Tempo CI workflow and blocks unrelated PRs.

With this split:
- `CI Tempo` (`ci-tempo.yml`) no longer depends on the MPP service.
- `CI MPP` (`ci-mpp.yml`) runs the MPP e2e test independently and can be made non-required / re-run in isolation when the upstream service is flaky.

No script or test logic is changed; `./.github/scripts/tempo-mpp.sh` is invoked from the new workflow with the same env (`TEMPO_KEYS_TOML_B64`, `MPP_API_KEY`, `MPP_DEPOSIT`).

### DIFF
--- a/.github/workflows/ci-mpp.yml
+++ b/.github/workflows/ci-mpp.yml
@@ -17,7 +17,7 @@ env:
   RUSTC_WRAPPER: "sccache"
 
 jobs:
-  mpp-e2e:
+  mpp-check:
     runs-on: depot-ubuntu-latest
     timeout-minutes: 60
     permissions:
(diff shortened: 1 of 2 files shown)


mablr (approved):
nice 👍

PR #14471, opened 2026-04-27
chore: drop Optimism stack support

figtracer:
## Summary

Draft exploration for dropping native OP-stack support after the extraction discussion in foundry-rs/foundry-core#26 and foundry-rs/foundry#14419.

- remove OP EVM/network/deposit transaction handling from Anvil, Forge, Cast, script, primitives, and formatting code
- remove OP-specific dependencies, workspace patches, and the `cast da-estimate` command
- keep regular Ethereum/Tempo paths and raw RPC transaction retrieval working through the generic stack


crates/anvil/tests/it/optimism.rs
```diff
@@ -1,348 +0,0 @@
-//! Tests for OP chain support.
```
mablr, line 1: Maybe some tests can stay (if modified)?

mablr (commented):
Nice, the change is not that big!

crates/anvil/tests/it/optimism.rs
figtracer: we can make them work and just pull them out of this `optimism.rs` file somewhere else, can probably just get rid of them for now though imo

figtracer:
we don't need this anymore since we're feature-gating it

PR #14472, opened 2026-04-27
ci: prefer direct-key credentials for MPP e2e

brendanjryan:
## Summary
- allow the MPP e2e script to run from `TEMPO_PRIVATE_KEY` without requiring `keys.toml`
- prefer a direct-key secret in `CI MPP`, while keeping `TEMPO_KEYS_TOML_B64` as a fallback during migration

## Why
The current MPP CI path depends on Tempo wallet keystore material and access-key provisioning. That is more fragile than a funded direct key because it can fail on expired access keys, missing keychain state, or chain-id mismatches in `key_authorization`.

Foundry already supports direct-key mode internally via `TEMPO_PRIVATE_KEY`; this change lets the CI script and workflow use that path directly.

### DIFF
--- a/.github/scripts/tempo-mpp.sh
+++ b/.github/scripts/tempo-mpp.sh
@@ -28,7 +28,9 @@ else
   ANVIL="anvil"
   CHISEL="chisel"
 fi
-export MPP_DEPOSIT=100000
+export MPP_DEPOSIT="${MPP_DEPOSIT:-100000}"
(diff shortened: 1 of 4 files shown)


zerosnacks:
cc @stevencartavia

stevencartavia (approved):

mablr (approved):

zerosnacks (approved):

PR #14478, opened 2026-04-28
fix(script): preserve pending broadcast indices

Perico-perica46:
## Motivation

Concurrent script broadcasts can complete transactions out of order. Recording pending hashes with a completion-order counter can associate a hash with the wrong transaction index, which makes checkpointed broadcast state unreliable.

## Solution

Compute each transaction's original sequence index from its batch position and carry that index through retries before recording the pending hash.

Add a regression test that records pending hashes from out-of-order broadcast completions and verifies they are stored on the original transaction indices.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/script/src/broadcast.rs
```diff
@@ -39,6 +39,15 @@ use itertools::Itertools;
 use tempo_alloy::{TempoNetwork, rpc::TempoTransactionRequest};
 use tempo_primitives::transaction::Call;
 
+fn pending_sequence_index(
```
mablr:
```suggestion
const fn pending_sequence_index(
```

mablr (commented):
Makes sense, just a smol nit.

mablr (commented):
@Perico-perica46 could you please resolve the merge conflict?

Perico-perica46:
@mablr Done.

mablr (dismissed):
Just made a small style fix.

Lgtm

mablr:
@figtracer ptal

crates/script/src/broadcast.rs
```diff
@@ -40,6 +40,15 @@ use itertools::Itertools;
 use tempo_alloy::{TempoNetwork, rpc::TempoTransactionRequest};
 use tempo_primitives::transaction::Call;
 
+const fn pending_sequence_index(
```
figtracer: let's not add a helper for this we just need the original tx index

Perico-perica46:
@mblr @figtracer Updated.

mablr (dismissed):
Nice.

crates/script/src/broadcast.rs
```diff
@@ -1244,6 +1245,34 @@ mod tests {
     const ACCESS_KEY_PRIVATE_KEY: &str =
         "0x59c6995e998f97a5a004497e5da3b5d2b2b66a87f064d39c44da0b6d6e4f8ff0";
 
+    #[test]
+    fn pending_hashes_keep_original_sequence_indices_when_completions_are_unordered() {
```
figtracer: as written, the test would still pass with the old completion-order counter bug

we should cover the batching/`FuturesUnordered` path so it fails if the code regresses to recording hashes in completion order

Perico-perica46:
@figtracer Done.

PR #14480, opened 2026-04-28
ci(benchmarks): split commit/push out of bench job

emmajam:
## Motivation

The `Foundry Benchmarks` workflow has been failing on the **Commit and read benchmark results** step ([example run](https://github.com/foundry-rs/foundry/actions/runs/25018563273/job/73272826625)):

```
fatal: could not read Username for 'https://github.com': No such device or address
Error: Process completed with exit code 128.
```

The bench job (`run-benchmarks`) checks out with `persist-credentials: false` (correctly so — it then executes untrusted third-party builds via `foundryup` and `forge` against external repos like `ithacaxyz/account` and `Vectorized/solady`), so the in-script `git push` had no token and exited 128.

The minimal fix would be `persist-credentials: true` on that checkout, but that permanently parks a `contents: write` GitHub token on the self-hosted runner alongside untrusted code — not a tradeoff worth one line.

## Solution

Split responsibilities cleanly between the two existing jobs:

- **`run-benchmarks`** (self-hosted `foundry-runner`): runs benches, uploads artifact, exposes `pr_comment` as a job output. Permissions reduced to `contents: read`. **No git writes.**
- **`publish-results`** (`ubuntu-latest`): downloads the artifact, commits `benches/LATEST.md`, pushes the branch, and (for `workflow_dispatch`) opens the results PR / comments on the target PR. **This is the only job with `contents: write`.**

`commit-and-read-benchmarks.sh` is split to match its new homes:

- `read-benchmark-results.sh` — pure read, emits GitHub Actions outputs.
- `commit-benchmark-results.sh` — branch creation, commit, push.

Also removes the duplicate **Push branch for manual runs** step in `publish-results` that was attempting to push a branch which had only ever existed on the bench runner.

### Why this is better long term

1. **Token isolation.** A compromised dependency in any benched repo (or in `foundryup`'s install path) can no longer reach a write-scoped token, since the bench job has none. The blast radius shrinks to the lightweight `ubuntu-latest` job that only touches the artifact.
2. **Single source of truth for git.** Both jobs were previously trying to push the results branch. Centralizing all git/PR logic in `publish-results` removes the duplication and the broken second push.
3. **Replayable.** With the bench output captured as an artifact, `publish-results` can be re-run independently without spending another 30+ minutes regenerating numbers.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mablr (dismissed):

.github/workflows/benchmarks.yml
```diff
@@ -15,55 +15,29 @@ on:
...
-
-  COVERAGE_REPOS: >-
-    ithacaxyz/account:v0.5.7,
-    aave/aave-v4:af1f0f2ba323ac6fbaaee3abf6be060c78e22d35,
-    uniswap/v4-core:46c6834698c48bc4a463a86d8420f4eb1d7f3b75,
-    sparkdotfi/spark-psm:v1.0.0
-
+  ITHACAXYZ_ACCOUNT: "ithacaxyz/account:v0.3.2"
+  VECTORIZED_SOLADY: "Vectorized/solady:v0.1.22"
+  DEFAULT_REPOS: "ithacaxyz/account:v0.3.2,Vectorized/solady:v0.1.22"
```
zerosnacks: This looks like a regression, we need to retain the changes @mablr made

zerosnacks (changes requested):

.github/workflows/benchmarks.yml
mablr: https://github.com/foundry-rs/foundry/blob/12642e358a792e3ec1a54378d350f12c0480e261/.github/workflows/benchmarks.yml#L23-L54

mablr (commented):
@emmajam I reverted unwanted changes.

mablr (approved):
Re-targeted to an upstream branch to make CI dry-run.

zerosnacks (approved):

PR #14481, opened 2026-04-28
fix: bump Tempo deps and validate fork schedule compatibility

0xrusowsky:
## Summary
- bump the pinned `tempoxyz/tempo` workspace deps to `8bd4d01d37e3cc324030baacbce2da0862d7735a`
- add a cast integration test that calls `tempo_forkSchedule` on configured Tempo RPCs and parses the response with `TempoHardfork`
- run that compatibility check in `ci-tempo` before the script-based Tempo checks

## Context
Older Foundry Tempo pins only knew hardforks through `T4`. When a node reports `active: "T5"`, `is_hardfork_active(T3)` can silently return `false`, which causes `cast keychain` to choose the legacy authorize selector on T3+ chains.

The new integration test makes that mismatch obvious by failing as soon as a configured Tempo RPC reports a fork name this Foundry pin cannot parse.

## Verification
- `cargo check -p cast -p anvil -p forge --locked`
- `TEMPO_MAINNET_RPC_URL=https://rpc.tempo.xyz TEMPO_TESTNET_RPC_URL=https://rpc.moderato.tempo.xyz cargo test --locked -p cast --test cli tempo_fork_schedule_parses_configured_rpcs -- --exact --nocapture`


0xrusowsky:
Superseded by #14485 (re-opened from an upstream branch so the `ci-tempo` job can access repository secrets — see https://github.com/foundry-rs/foundry/actions/runs/25051145675/job/73378970641?pr=14481).

PR #14482, opened 2026-04-28
feat(invariant): decouple handler-side assertions from invariant predicates

### LINKED ISSUE #14448: feat(invariant): track handler assertion failures independently during continuous runs
## Summary

Follow-up to #12587 (assert all invariants on `continuous_run`). Two improvements to handler assertion tracking during invariant campaigns:

### 1. Keep running after all invariants are broken
Currently the campaign stops when all invariants break (`can_continue` returns false). However, there may still be undiscovered assertion failures in handler functions. The campaign should continue running for its full `runs`/`timeout` to catch these.

### 2. Track broken handlers separately
Once a handler assertion fails, record it but stop re-checking its assertions on subsequent calls. The handler should still be called for sequence building (it may trigger interesting state transitions) but its known-broken assertions should be suppressed to avoid noise.

**Implementation sketch:**
- Add `broken_handlers: HashSet<(Address, Selector)>` to `InvariantTest`
- Decouple "record failure" from "stop campaign" in the main fuzz loop
- Track handler assertion failures separately in `InvariantFailures` (`handler_errors` map)
- Report handler failures alongside invariant failures in output

Related: #14437, #12587
cc @rappie (original suggestion: https://github.com/foundry-rs/foundry/pull/12587#issuecomment-3699642873)

grandizzy:
## Summary

Decouple handler-side assertion failures from invariant predicate violations so the campaign can keep running after either class of bug is discovered, dedup handler bugs at the function-entry-point level (Echidna/Medusa semantics), and surface every break live in the pulse stream. Closes #14448.

## Motivation

Today, an `assert(false)` (or `vm.assertEq` failure) inside a fuzzed handler is attributed to every live invariant and aborts the campaign on the first hit. On real targets like Aave v4 SCFuzzbench, this means a single canary in the handler kills 95+% of the time budget — upstream Foundry finds 3 bugs where Echidna/Medusa find 10–11 in the same 24 h window.

## Result on SCFuzzbench (Aave v4)

Reference baseline from the public scfuzzbench leaderboard ([run 1772801774](https://scfuzzbench.com/runs/1772801774/454f886c9668a94e8595de32219ce2b9), 4 instances per fuzzer, target `v0.5.6-recon`). All four tools count **unique public-function entry points** that triggered an assertion failure, so the comparison is apples-to-apples on dedup granularity.

| Fuzzer | Version | 1 h (per-instance median [IQR]) | 24 h cumulative |
|---|---|---|---|
| Echidna | 2.3.1 | 8 [7,8] | 11 |
| Medusa | 1.4.1 | 6 [5,6] | 10 |
| Foundry baseline | v1.6.0-rc1 (`aviggiano/foundry@master`) | 3 [3,3] (plateaus at 6 min) | 3 |
| **Foundry-on-PR** | this PR | **8** (single 1 h run, no seed, fresh corpus) | TBD |

Single 1 h run found 8 / 11 of the reachable bug universe: `assert_canary`, `iHub_mintFeeShares`, `iSpoke_withdraw`, `invariant_canary`, `invariant_shouldNotBecomeLiquidatable`, `invariant_totalBorrowedLessThanSupplied_v0/v1/v2`. First Foundry build to break `iHub_mintFeeShares`, `iSpoke_withdraw`, and the three `totalBorrowedLessThanSupplied_*` invariants on this benchmark. Missing: `iSpoke_liquidationCall`, `iSpoke_repay`, `invariant_supplySharePriceAndDrawnIndexMonotonic` (Echidna-exclusive in 24 h baseline), `invariant_totalBorrowedSharesMatchesSpokeSum`.

**Update (1 h re-run on current commit):** A 1 h fixed-seed re-run on the latest code additionally surfaced `iSpoke_supply_ASSERTION_SUPPLY_DOS` — a handler-side assertion that **neither Echidna (24 h × 4 instances) nor Medusa surface in the SCFuzzbench leaderboard**. The bug isn't in the public 11-bug "reachable" universe at all, so this PR's handler-bug machinery is the first to expose it. The same run also independently re-confirmed `iSpoke_repay`, which is in the Echidna+Medusa shared set but had been missed by upstream Foundry.

## What changes

- **Separate failure class.** Handler-side assertion bugs are recorded in a new `broken_handlers: HashMap<(Address, Selector), HandlerAssertionFailure>` map (keyed by the asserting call's `(reverter, selector)` site) and deduplicated so each unique handler bug is reported once. Same handler asserting through N distinct code paths counts as **one** bug, matching what Echidna and Medusa report (Echidna explicitly cannot tell multiple asserts in the same function apart; Medusa surfaces one bug per failing function).
- **Site-granular dedup with smaller-reproducer rule.** On `(reverter, selector)` collision the entry with the shortest `call_sequence` wins, so persisted reproducers stay minimal as the campaign progresses.
- **Dedicated report section.** Bugs render under `Suite handlers: N assertion bug(s) found` with full reason + counterexample, separate from the per-invariant `[FAIL: ...]` blocks.
- **Continuous campaign under `assert_all`.** Preflight handler bugs are recorded and the campaign keeps running for the full budget instead of aborting on the first hit.
- **Live telemetry.** Progress bar and JSON pulse events surface unique `broken_handlers` counts alongside invariant `unique_failures`. Persisted handler bugs are pre-seeded before the run loop so they appear from the first emission.
- **Persistence + replay.** Each broken handler is written to `<failure_dir>/handlers/<site_hash>.json` where `site_hash = keccak256(reverter || selector)` — independent of edge coverage and stable across runs. Reuses `InvariantPersistedFailure` with `assertion_failure: true`. At campaign start every persisted file is replayed via `check_sequence`; bugs that still reproduce are merged into `handler_errors` (shortest-wins on collision), files that no longer reproduce are deleted in place, and settings-incompatible files are left untouched.
- **Post-campaign shrinking.** Each handler bug's call sequence is shrunk by the same shrink-loop driver used for invariant counterexamples. Intermediate calls that themselves assert are rejected so shrinking can't silently promote a different finding. The renderer surfaces `(original: N, shrunk: M)` matching invariant counterexamples.

## Known limitations / TODO

- **Multi-assertion dedup within the same handler.** When a single `(reverter, selector)` function contains multiple distinct `assert(...)` / `t(...)` calls, our key collapses them all to one bug. Echidna has the same limitation by design ("Echidna cannot determine which assertion failed"). If finer attribution becomes useful, extend the dedup key with a stable per-call discriminator (revert reason hash, source location, or label string).

### DIFF
--- a/crates/evm/evm/src/executors/invariant/error.rs
+++ b/crates/evm/evm/src/executors/invariant/error.rs
@@ -1,7 +1,7 @@
 use super::InvariantContract;
 use crate::executors::RawCallResult;
 use alloy_json_abi::Function;
-use alloy_primitives::{Address, Bytes};
+use alloy_primitives::{Address, B256, Bytes, Selector, keccak256};
 use foundry_config::InvariantConfig;
 use foundry_evm_core::{
     decode::{ASSERTION_FAILED_PREFIX, EMPTY_REVERT_DATA, RevertDecoder},
@@ -11,15 +11,54 @@ use foundry_evm_fuzz::{BasicTxDetails, Reason, invariant::FuzzRunIdentifiedContr
 use proptest::test_runner::TestError;
 use std::{collections::HashMap, fmt};
 
-/// Run-scoped context bundling the references that an invariant run needs in multiple places
-/// (recording failures, attributing breaks to a specific invariant, etc.).
-///
-/// Constructed once per loop iteration and passed by reference; produces failure records via
-/// [`InvariantRunCtx::failed_case`].
+/// A handler-side assertion bug: a `require`/`assert` inside a fuzzed handler that the
+/// campaign reached. Deduped by `(reverter, selector)` site (Echidna/Medusa semantics),
+/// shortest sequence wins on collision.
+#[derive(Clone, Debug)]
+pub struct HandlerAssertionFailure {
+    /// Handler contract whose call asserted.
+    pub reverter: Address,
+    /// 4-byte selector of the failing function.
+    pub selector: Selector,
+    /// Call sequence including the failing call (post-shrink: minimal prefix).
+    pub call_sequence: Vec<BasicTxDetails>,
+    /// Pre-shrink length, for the `(original: N, shrunk: M)` renderer.
+    pub original_sequence_len: usize,
+    /// Decoded revert/assert reason.
+    pub revert_reason: String,
+    /// Stable hash of edge coverage at the asserting call (falls back to `(reverter,
+    /// selector)`). Used by the shrinker to preserve path identity, not for dedup.
+    pub edge_fingerprint: B256,
+}
+
+impl HandlerAssertionFailure {
+    /// Builds a failure from a replayed sequence whose last call asserted.
+    pub fn from_replayed_sequence(
+        call_sequence: Vec<BasicTxDetails>,
+        edge_fingerprint: B256,
+        revert_reason: String,
+    ) -> Self {
+        let last = call_sequence.last().expect("replayed sequence is non-empty");
+        let reverter = last.call_details.target;
+        let selector_bytes: [u8; 4] =
+            last.call_details.calldata.get(..4).and_then(|s| s.try_into().ok()).unwrap_or_default();
+        let original_sequence_len = call_sequence.len();
+        Self {
+            reverter,
+            selector: Selector::from(selector_bytes),
+            call_sequence,
+            original_sequence_len,
+            revert_reason,
+            edge_fingerprint,
+        }
+    }
+}
+
+/// Run-scoped references shared by failure-recording paths in an invariant run.
 pub struct InvariantRunCtx<'a> {
-    /// The invariant test contract definition.
+    /// The invariant test contract.
     pub contract: &'a InvariantContract<'a>,
-    /// Active invariant configuration (provides `shrink_run_limit`, `fail_on_revert`, ...).
+    /// Active invariant configuration.
     pub config: &'a InvariantConfig,
     /// Fuzz targets discovered for this run.
     pub targeted_contracts: &'a FuzzRunIdentifiedContracts,
@@ -28,13 +67,9 @@ pub struct InvariantRunCtx<'a> {
 }
 
 impl<'a> InvariantRunCtx<'a> {
-    /// Builds a [`FailedInvariantCaseData`] attributed to `broken_fn`.
-    ///
-    /// `fail_on_revert` is taken separately because `assert_invariants` overrides it with
-    /// the per-invariant flag, while every other call site forwards `self.config.fail_on_revert`.
-    /// `assertion_failure` is set when the failure originated from a Solidity `assert`/
-    /// `vm.assert*` path; it normalizes empty decoded revert data into a stable user-facing
-    /// message so invariant output is not blank.
+    /// Builds a [`FailedInvariantCaseData`] attributed to `broken_fn`. `fail_on_revert` is
+    /// passed in because `assert_invariants` overrides it with the per-invariant flag.
+    /// `assertion_failure=true` normalizes empty revert data so output is not blank.
     pub fn failed_case<FEN: FoundryEvmNetwork>(
         &self,
         broken_fn: &Function,
@@ -43,21 +78,7 @@ impl<'a> InvariantRunCtx<'a> {
         call_result: RawCallResult<FEN>,
         inner_sequence: &[Option<BasicTxDetails>],
     ) -> FailedInvariantCaseData {
-        // Collect abis of fuzzed and invariant contracts to decode custom error.
-        let revert_reason = RevertDecoder::new()
-            .with_abis(self.targeted_contracts.targets.lock().values().map(|c| &c.abi))
-            .with_abi(self.contract.abi)
-            .decode(call_result.result.as_ref(), call_result.exit_reason);
-        // Non-reverting assertion failures surface through Foundry's failure flags instead of
-        // revert data. Use a stable fallback so invariant output is not blank, both for the
-        // successful-call/assertion path and the explicit assertion_failure flag.
-        let needs_fallback = matches!(revert_reason.as_str(), "" | EMPTY_REVERT_DATA);
-        let revert_reason = if needs_fallback && (!call_result.reverted || assertion_failure) {
-            ASSERTION_FAILED_PREFIX.to_string()
-        } else {
-            revert_reason
-        };
-
+        let revert_reason = self.decode_revert_reason(&call_result, assertion_failure);
         let origin = broken_fn.name.as_str();
         FailedInvariantCaseData {
             test_error: TestError::Fail(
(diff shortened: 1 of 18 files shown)


crates/evm/evm/src/executors/invariant/error.rs
```diff
@@ -10,6 +10,28 @@ use foundry_evm_fuzz::{BasicTxDetails, Reason, invariant::FuzzRunIdentifiedContr
...
 
+/// Records a single handler-side assertion bug discovered during an invariant campaign.
+///
+/// Handler-side assertions (e.g. a `require`/`assert` inside a fuzzed handler that the campaign
+/// reaches with a malformed input) are bugs in their own right, but they are *not* invariant
+/// predicate violations. We record them once per `(reverter, selector)` so the campaign can keep
+/// running for the rest of the budget and surface deeper bugs without polluting the invariant
+/// `errors` map or stopping the run.
+#[derive(Clone, Debug)]
+pub struct HandlerAssertionFailure {
```
0xalpharush, line 18: could be a variant of InvariantFuzzError?

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -203,19 +203,26 @@ fn first_broken_event<'a>(
...
 ) -> serde_json::Value {
     let mut metrics = serde_json::to_value(corpus_metrics).unwrap_or_default();
     if let Some(obj) = metrics.as_object_mut() {
         obj.insert("failures".to_string(), json!(failure_metrics.failures));
         obj.insert("unique_failures".to_string(), json!(failure_metrics.unique_failures.len()));
+        // Phase E: surface unique handler-side assertion bugs in live progress so users
+        // can watch them accumulate without waiting for the campaign to finish. These are
+        // distinct from invariant predicate violations (counted by `failures` above) and
+        // are routed via `InvariantFailures::broken_handlers`.
+        obj.insert("broken_handlers".to_string(), json!(broken_handlers));
```
0xalpharush: I think this belongs combined with the above failure metrics

grandizzy: 👍 moved with https://github.com/foundry-rs/foundry/pull/14482/changes/a2d71de6b9060f6067cfac757521e30d5bb8d55b

crates/evm/evm/src/executors/invariant/result.rs
```diff
@@ -253,13 +263,72 @@ pub(crate) fn can_continue<FEN: FoundryEvmNetwork>(
             invariant_test.test_data.failures.reverts += 1;
         }
 
-        // Collect which invariants should be marked as failed due to this revert/assertion.
+        if is_assert_failure {
+            // Handler-side assertion: a unique bug attributable to the *handler call*, not to
```
0xalpharush: I had never really considered this and not sure how common it is in practice, but it's an interesting idea. I do think it's nice to be able to say all the handlers' reachable assertions fall under one test e.g. `N/M assertions tests failed`. This does make me think we should for sure keep "duplicate" crashes (https://github.com/foundry-rs/foundry/pull/12587#issuecomment-3583894749) and not stop calling these.

grandizzy: would this work, pls check https://github.com/foundry-rs/foundry/pull/14482/changes/21ac1bb02bb3456db78548a8efac9a2fac1662b4

0xalpharush: can we avoid hashing for sequences that didn't reach failure?

grandizzy: yep, added here https://github.com/foundry-rs/foundry/pull/14482/changes/80aec25ed16e4f4c28d4bd119305310d8faabb6d

crates/evm/evm/src/executors/invariant/error.rs
```diff
@@ -19,6 +112,18 @@ pub struct InvariantFailures {
...
+    /// asserting via N different code paths counts as a single bug (Echidna/Medusa
+    /// semantics). On collision, the entry with the shortest `call_sequence` wins so
+    /// persisted reproducers stay minimal.
+    ///
+    /// TODO: dedup multiple distinct `assert(...)` failures within the same
+    /// `(reverter, selector)` handler. Echidna explicitly cannot tell them apart and
+    /// our current key collapses them as well; if/when callers need finer attribution
+    /// (e.g. per-assertion-label), extend the key with a stable per-call discriminator
+    /// (revert reason hash, source location, or label string).
+    pub broken_handlers: HashMap<(Address, Selector), HandlerAssertionFailure>,
```
0xalpharush: Maybe in follow-up we merge `errors` and `broken_handlers` and make add assertion as a variant to `InvariantFuzzError`

crates/evm/evm/src/executors/invariant/shrink.rs
```diff
@@ -141,6 +185,59 @@ fn build_shrunk_sequence(
     result
 }
 
+/// Shared shrink loop driver. Tries to drop each call; `predicate` returns whether the
+/// candidate still triggers the bug. `skip_idx` pins positions that must not be dropped
```
0xalpharush: I think this just limits the ability to shrink fwiw. Did you observe it helping?

crates/evm/evm/src/executors/invariant/shrink.rs
```diff
@@ -164,45 +261,100 @@ pub(crate) fn shrink_sequence<FEN: FoundryEvmNetwork>(
...
-    for _ in 0..config.shrink_run_limit {
-        if early_exit.should_stop() {
-            break;
+    Ok(build_shrunk_sequence(calls, &shrinker, accumulate_warp_roll))
+}
+
+/// Replays `sequence` (indices into `calls`) against `executor`. When
+/// `accumulate_warp_roll` is set, warp/roll from skipped calls is folded into the next
+/// included call. `on_call` may stop early; otherwise non-reverted calls are committed.
+fn replay_sequence<FEN, T, F>(
```
0xalpharush, line 283: Aside, it would be nice to share the same code b/w replay and fuzz. Maybe not straigthtforward..

crates/evm/evm/src/executors/invariant/error.rs
```diff
@@ -10,6 +10,99 @@ use foundry_evm_fuzz::{BasicTxDetails, Reason, invariant::FuzzRunIdentifiedContr
...
+    /// this holds the minimal prefix that still triggers the anchor assertion.
+    pub call_sequence: Vec<BasicTxDetails>,
+    /// Pre-shrink length of `call_sequence`, used by the renderer's
+    /// `(original: N, shrunk: M)` output.
+    pub original_sequence_len: usize,
+    /// Decoded revert/assert reason.
+    pub revert_reason: String,
+    /// Always `true` for entries in this struct; mirrored for symmetry with
+    /// `FailedInvariantCaseData::assertion_failure`.
+    pub assertion_failure: bool,
```
mablr: Can be removed imo, by definition `HandlerAssertionFailure` type is an assertion failure.

crates/evm/evm/src/executors/invariant/result.rs
```diff
@@ -253,13 +267,72 @@ pub(crate) fn can_continue<FEN: FoundryEvmNetwork>(
...
+                invariant_test.test_data.failures.record_handler_failure(HandlerAssertionFailure {
+                    reverter: target,
+                    selector,
+                    call_sequence,
+                    original_sequence_len,
+                    revert_reason,
+                    assertion_failure: true,
+                    edge_fingerprint: fingerprint,
+                });
+            }
```
mablr: We can probably extract this to an helper as it is duplicated in [crates/evm/evm/src/executors/invariant/mod.rs#L699-L735](https://github.com/foundry-rs/foundry/blob/1da49bdc91d5654b26eb4e8cb6dcfa39cfd2ba6e/crates/evm/evm/src/executors/invariant/mod.rs#L699-L735)

crates/evm/evm/src/executors/invariant/result.rs
```diff
@@ -253,13 +267,72 @@ pub(crate) fn can_continue<FEN: FoundryEvmNetwork>(
...
+                    invariant_contract,
+                    invariant_config.shrink_run_limit,
+                    invariant_config.fail_on_revert,
+                    &invariant_test.targeted_contracts,
+                    &invariant_run.inputs,
+                    call_result,
+                    &[],
+                )
+                .with_assertion_failure(true);
+                let revert_reason = case_data.revert_reason;
```
mablr: Infer `revert_reason` directly could be more elegant.

Maybe extracting inner `FailedInvariantCaseData::new`'s `revert_reason` logic to an helper could a nice follow-up

crates/evm/evm/src/executors/invariant/result.rs
```diff
@@ -253,13 +267,72 @@ pub(crate) fn can_continue<FEN: FoundryEvmNetwork>(
...
+                invariant_test.test_data.failures.revert_reason = Some(revert_reason.clone());
```
mablr: `InvariantFailures::revert_reason` seems overwritten by handler failures.

`revert_reason` on `InvariantFailures` is the invariant's last revert reason, surfaced in the top-level header, right? Overwriting it from handler assertion paths means the primary failure header can show a handler's message instead of the actual invariant violation?

crates/evm/evm/src/executors/invariant/error.rs
grandizzy: done in https://github.com/foundry-rs/foundry/pull/14482/changes/c38ecb1f8b229169de646246eae011ebdfa7021e

crates/evm/evm/src/executors/invariant/error.rs
grandizzy: done in https://github.com/foundry-rs/foundry/pull/14482/changes/c38ecb1f8b229169de646246eae011ebdfa7021e

crates/evm/evm/src/executors/invariant/shrink.rs
grandizzy: was a leftover, removed with https://github.com/foundry-rs/foundry/pull/14482/changes/bdba1476af8007ddb87fbe559b86ddf3ab783dcf

crates/evm/evm/src/executors/invariant/error.rs
grandizzy: removed in https://github.com/foundry-rs/foundry/pull/14482/changes/cf17bd9614c2a285a38a72cd14e55e94765499ec

crates/evm/evm/src/executors/invariant/result.rs
grandizzy: extracted in https://github.com/foundry-rs/foundry/pull/14482/changes/6d1cfe481fff7237539463a0d5b4372ffdef5687

crates/evm/evm/src/executors/invariant/result.rs
grandizzy: dead code on refactor https://github.com/foundry-rs/foundry/pull/14482/changes/6d1cfe481fff7237539463a0d5b4372ffdef5687 so I just removed in same commit

crates/evm/evm/src/executors/invariant/result.rs
grandizzy: pls check https://github.com/foundry-rs/foundry/pull/14482/changes/53c4f59fd1278888dd790c7d01c5d04084559c39

mablr: nice 👍

mablr (dismissed):
sick 🫡

zerosnacks:
@0xalpharush if your time permits, would be great if you could do a final pass review on this 😃

0xKarl98:
Think we need to clear the non-reverting assertion flag before continuing the campaign.

In the assertions_revert = false path, vm.assert* does not revert. The cheatcode writes GLOBAL_FAIL_SLOT = 1 instead:
ccx.ecx.journal_mut().sstore(CHEATCODE_ADDRESS, GLOBAL_FAIL_SLOT, U256::from(1))?;

This PR correctly detects that as a handler assertion via did_fail_on_assert(...), records it through record_handler_assertion_bug(...), and keeps the campaign running. 
  
 The issue is that the fuzzed call has already been committed before we get there, so the global failure flag can remain in executor state.

  That matters because the next can_continue(...) call gates invariant checks on:
```rust
  if !call_result.reverted && handlers_succeeded() {
      assert_invariants(...)?;
  }
 ```
and handlers_succeeded() calls executor.is_success(...), which eventually checks:

 ```rust
  if self.has_global_failure(&state_changeset) {
      return false;
  }
  ```
So a stale GLOBAL_FAIL_SLOT from the previous handler assertion can make later successful calls look failed, suppress invariant checks, or affect afterInvariant.

I think the handler-assertion path should clear the global failure flag after recording the bug and before returning continues = true, probably using the existing Executor::clear_global_failure(...) helper.
  
  whyt @zerosnacks

**Comment by p5hzehxa** [comment]
```rust

```

grandizzy:
> Think we need to clear the non-reverting assertion flag before continuing the campaign.
> 
> In the assertions_revert = false path, vm.assert* does not revert. The cheatcode writes GLOBAL_FAIL_SLOT = 1 instead: ccx.ecx.journal_mut().sstore(CHEATCODE_ADDRESS, GLOBAL_FAIL_SLOT, U256::from(1))?;
> 
> This PR correctly detects that as a handler assertion via did_fail_on_assert(...), records it through record_handler_assertion_bug(...), and keeps the campaign running.
> 
> The issue is that the fuzzed call has already been committed before we get there, so the global failure flag can remain in executor state.
> 
> That matters because the next can_continue(...) call gates invariant checks on:
> 
> ```rust
>   if !call_result.reverted && handlers_succeeded() {
>       assert_invariants(...)?;
>   }
> ```
> 
> and handlers_succeeded() calls executor.is_success(...), which eventually checks:
> 
> ```rust
>  if self.has_global_failure(&state_changeset) {
>      return false;
>  }
> ```
> 
> So a stale GLOBAL_FAIL_SLOT from the previous handler assertion can make later successful calls look failed, suppress invariant checks, or affect afterInvariant.
> 
> I think the handler-assertion path should clear the global failure flag after recording the bug and before returning continues = true, probably using the existing Executor::clear_global_failure(...) helper.
> 
> whyt @zerosnacks

addressed in https://github.com/foundry-rs/foundry/pull/14482/changes/644605cd57b266e4dc61b00bb52f54d8a254b38f

0xalpharush:
re https://github.com/foundry-rs/foundry/commit/644605cd57b266e4dc61b00bb52f54d8a254b38f 
So we'd report it as only an assertion failure but it was also an invariant failure? I think this would get picked up on replay if only the assertion bug was fixed, right? IMO it is fine to keep first failure

0xalpharush (approved):

mablr (dismissed):

grandizzy:
> So we'd report it as only an assertion failure but it was also an invariant failure?

The flag isn't local to the failing sequence so once fail slot is set every subsequent `is_success()` check in the same forge test invocation returns false (across later runs and other invariants). So we don't just lose the invariant failure that overlaps the assertion bug but we silently suppress evaluation for unrelated sequences against other invariants for the rest of the campaign. Replay only re-runs the shrunken counterexample, so the suppressed independent breaks aren't recovered by fixing the assertion bug. Made gated approach in https://github.com/foundry-rs/foundry/pull/14482/changes/36aa91d61cd910096ba8348cfc04b98e2892732a and retested aave bench (same seed as before) which now surfaces 5 predicate breaks (extra `totalBorrowedLessThanSupplied_v1/_v2` are on a different invariant from the assertion-bug shrunken sequence and aren't reproducible from its replay).

zerosnacks (approved):

figtracer (approved):

PR #14486, opened 2026-04-28
fix(config): warn instead of erroring on unknown FOUNDRY_PROFILE

zerosnacks:
## Motivation

Currently, setting `FOUNDRY_PROFILE` to a profile that isn't defined in `foundry.toml` aborts the command with:

```
Error: failed to extract foundry config:
foundry config error: selected profile `ci` does not exist
```

This is annoying when an env-driven `FOUNDRY_PROFILE=ci` is set globally but a particular project only defines `[profile.default]`.

## Solution

Treat an unknown selected profile as a warning instead of a fatal error: fall back to the `default` profile and emit a new `Warning::UnknownProfile` so users are still informed.

```
FOUNDRY_PROFILE=ci forge fmt
Warning: Selected profile `ci` does not exist; falling back to the default profile.
```

The lenient behaviour for nested lib configs (`from_figment_fallback`) is preserved as-is.

## Notes

- New variant: `Warning::UnknownProfile { profile }`.
- Updated test `fails_on_unknown_profile` -> `warns_on_unknown_profile`.

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -847,12 +847,15 @@ impl Config {
         // Check if the selected profile exists.
         if config.profiles.contains(&selected_profile) {
             config.profile = selected_profile;
-        } else if strict_profile {
-            return Err(ExtractConfigError::new(Error::from(format!(
-                "selected profile `{selected_profile}` does not exist"
-            ))));
         } else {
-            // Fall back to default profile for nested lib configs.
+            // Fall back to the default profile. In strict mode (top-level loads), emit a warning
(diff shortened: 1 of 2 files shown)


mablr (approved):
Lgtm

I confirm:
- `FOUNDRY_PROFILE=ci forge build` with only `[profile.default]` defined -> warning printed, command succeeded, exit 0
- Warning goes to `stderr` (confirmed via `2>/dev/null / 2>&1 1>/dev/null`), stdout is clean, tooling won't break

figtracer (approved):

PR #14488, opened 2026-04-28
fix(anvil): respect configured genesis number when restoring `genesis_hash`

AissataF:
`BlockchainStorage::load_blocks` was hardcoding `block_number == 0` to detect the genesis block during state-snapshot restore. With `--block-number 73 --load-state ...`, no loaded block matched, so `genesis_hash` kept pointing at the dummy block from `BlockchainStorage::new()` and `Earliest` / `Safe` / `Finalized` tag lookups returned the wrong hash.                             
                                                                                             
Stored `genesis_number` on `BlockchainStorage` and compare against it in `load_blocks` instead. All existing constructors (`new`, `forked`, `empty`) wire the new field through.
                                                                                             
the meaningful diff in `load_blocks` is one line — `block_number == 0` becomes `block_number == self.genesis_number`. The follow-up commit adds a regression test that sets `genesis_number = 73`, loads a serialized block at 73, and asserts `genesis_hash` is rewritten to the loaded hash. There is also a sanity case proving that with `genesis_number = 0` and no block 0 in the snapshot, the dummy `genesis_hash` is preserved

 — i.e. the new test would have failed under the old hardcoded `== 0` logic


mablr (dismissed):
Makes sense, thanks.

mablr:
@AissataF Could you please have a look to CI errors ?

AissataF:
> @AissataF Could you please have a look to CI errors ?

cheer to Claude, should be good now

mablr (commented):
Sorry for the delay @AissataF, I think your change makes sense.

Would you mind fix the conflict on Cargo.toml, please ?

AissataF:
> Merge remote-tracking branch 'upstream/master' into fix-load-blocks-b…

Fixed

mablr (approved):
lgtm, thank you.

Pending another approval...

figtracer (approved):
lgtm

PR #14490, opened 2026-04-28
fix(anvil): respect non-zero genesis block in Otterscan APIs 

AissataF:
The three Otterscan address-history endpoints (`ots_searchTransactionsBefore`/`After`, `ots_getTransactionBySenderAndNonce`) hardcode `unwrap_or(1)` / `unwrap_or_default()` as the scan lower bound in non-fork mode. With `anvil --number 73` (or a `genesis.json` `"number": 73`) the search endpoints' `last_page` flag only flips at block `1` instead of `genesis + 1`, so paginated clients walk past the real start of history.                   

Added `Backend::genesis_number()` and use `genesis_number() + 1` as the fallback, mirroring the post-fork `f.block_number() + 1` already on the fork branch. Behavior is identical when `genesis_number == 0`. The fork arm is untouched — only the `unwrap_or*` paths changed.


figtracer (approved):
makes sense

mablr (approved):

AissataF:
hi @mablr @figtracer  anything I need to fix before merging?

mablr:
Sgtm, thanks 👍

PR #13065, opened 2026-01-12
fix(cheatcodes): make `vm.fee`, `vm.txGasPrice`, and `vm.blobhashes` work in isolation mode

onbjerg:
In isolation mode, the transaction environment is zeroed for fee-accounting purposes during synthetic transaction execution. This conflicts with cheatcodes that set these values, as the cheatcodes mutate the environment before it is zeroed.

The fix decouples the fee-accounting environment from the contract-visible environment by:

1. Adding `EnvOverrides` to store basefee, gas_price, and blob_hashes overrides
2. Adding `in_isolation_context` flag to `Cheatcodes`, set by `InspectorStack`
3. Intercepting `BASEFEE`, `GASPRICE`, and `BLOBHASH` opcodes in `step_end` to replace stack values with the overrides
4. Updating `vm.fee`, `vm.txGasPrice`, and `vm.blobhashes` to always set overrides, and only modify the actual environment when not in isolation mode

Closes #7277

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #9941, opened 2025-02-23
fix(`--isolate`): do not override the gas price

klkvr:
## Motivation

Closes https://github.com/foundry-rs/foundry/issues/7277

As per https://github.com/foundry-rs/foundry/issues/7277#issuecomment-1971521826 the fix is straightforward but might be breaking for other codebases as it now requires caller to actually pay for gas at the configured gas price, hence marking as draft for now

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #14493, opened 2026-04-28
fix(cheatcodes):`vm.{fee,txGasPrice,blobhashes}` support in `--isolate` mode

### LINKED ISSUE #7277: bug: `vm.fee` + `vm.txGasPrice` + `vm.blobHashes` cheatcodes incompatible with `--gas-report/--isolate` option
### Component

Forge

- [X] Foundry
- [X] Foundryup

forge 0.2.0 (2f432fb 2024-02-29T00:16:38.827810253Z)

forge test --gas-report

### Operating System

Windows

### Describe the bug

Example repo: https://github.com/KirienzoEth/foundry-gas-report-txgasprice-issue

The `vm.txGasPrice` cheatcode seems to be incompatible with the use of `--gas-report` in functions that use `tx.gasprice`.

The `--gas-report` option seem to be calling the contract that is tested with a new transaction (to measure the gas used I'm assuming), but as `vm.txGasPrice` only works in the current transaction (see [foundry book](https://book.getfoundry.sh/cheatcodes/tx-gas-price)), any function relying on `tx.gasprice` will fail.

mablr:
## Motivation
Close #7277. Supersedes #9941 and #13065.

Under `--isolate` / `--gas-report`, Foundry wraps each top-level test call in a synthetic inner transaction
(`InspectorStackRefMut::transact_inner`)
that zeroes `block.basefee` and `tx.gas_price` for fee accounting, and strips EIP-4844 blob fields so revm doesn't reject the inner tx. The side effect is that BASEFEE / GASPRICE / BLOBHASH read by called contracts return 0, so cheatcode mutations like `vm.fee`, `vm.txGasPrice` and `vm.blobhashes` are invisible to inner calls.

Approach: introduce opcode-level overrides in `Cheatcodes`:

* `EnvOverrides { basefee, gas_price, blob_hashes, ... }` is updated by the cheatcode handlers and applied in `step_end` by replacing the top of the interpreter stack right after BASEFEE/GASPRICE/BLOBHASH execute.
* `Cheatcodes::in_isolation_context` is toggled by `InspectorStackRefMut::transact_inner` around the inner `transact_raw` call. When `true`, the cheatcode handlers update only the override (preserving `tx.gas_price = 0` so a pranked zero-balance caller can still call without pre-funding `gas * gasPrice`, the original #7277 bug). Outside isolation they also mutate the real env to keep the historical behavior other code paths rely on.
* In `transact_inner`, an EIP-4844 cached tx is downgraded to EIP-1559 and its blob hashes cleared so revm validation accepts `gas_price = 0`; the contract-visible BLOBHASH is restored via the override.

Tests:
* `testdata/default/cheats/Fee.t.sol`: `IsolatedFeeTest` covers BASEFEE seen inside a called contract.
* `testdata/default/cheats/Blobhashes.t.sol`: `IsolatedBlobhashesTest` covers BLOBHASH seen inside a called contract.
* `testdata/default/cheats/TxGasPrice.t.sol`: new file covering GASPRICE visibility plus the pranked-zero-balance regression.


crates/cheatcodes/src/evm.rs
```diff
@@ -512,7 +512,16 @@ impl Cheatcode for feeCall {
     fn apply_stateful<FEN: FoundryEvmNetwork>(&self, ccx: &mut CheatsCtxt<'_, '_, FEN>) -> Result {
         let Self { newBasefee } = self;
         ensure!(*newBasefee <= U256::from(u64::MAX), "base fee must be less than 2^64");
-        ccx.ecx.block_mut().set_basefee(newBasefee.saturating_to());
+        let basefee: u64 = newBasefee.saturating_to();
+        // Always record the override so `BASEFEE` reads the cheatcode value
+        // even inside the synthetic isolation transaction (which zeroes
+        // `block.basefee` for fee-accounting). See `EnvOverrides`.
+        ccx.state.env_overrides.basefee = Some(basefee);
```
figtracer: this can probably get out-of-sync with with `snapshotState` / `revertToState`

mablr: good catch, fixed w/ per snapshot overrides 👍

figtracer (dismissed):

crates/cheatcodes/src/evm.rs
```diff
@@ -1338,8 +1370,13 @@ pub(super) fn get_nonce<FEN: FoundryEvmNetwork>(
 
 fn inner_snapshot_state<FEN: FoundryEvmNetwork>(ccx: &mut CheatsCtxt<'_, '_, FEN>) -> Result {
     let evm_env = ccx.ecx.evm_clone();
+    let env_overrides = ccx.state.env_overrides.clone();
     let (db, inner) = ccx.ecx.db_journal_inner_mut();
     let id = db.snapshot_state(inner, &evm_env);
+    // Capture the cheatcode-side env overrides alongside the backend
+    // snapshot so they can be rolled back in lockstep with `EvmEnv`. See
+    // `Cheatcodes::env_overrides_snapshots`.
+    ccx.state.env_overrides_snapshots.insert(id, env_overrides);
```
figtracer: we missed a gap around `vm.blobhashes` in this case:

```
vm.blobhashes(A);
uint256 id = vm.snapshotState();
vm.blobhashes(B);
vm.revertToState(id);
```

after this, the `BLOBHASH` opcode returns A but `vm.getBlobhashes()` still returns B

mablr: ebf21ae6c5594f3968ac5b2e88d1752ceb3da7f3

crates/cheatcodes/src/evm.rs
```diff
@@ -1338,8 +1376,13 @@ pub(super) fn get_nonce<FEN: FoundryEvmNetwork>(
 
 fn inner_snapshot_state<FEN: FoundryEvmNetwork>(ccx: &mut CheatsCtxt<'_, '_, FEN>) -> Result {
     let evm_env = ccx.ecx.evm_clone();
+    let env_overrides = ccx.state.env_overrides.clone();
```
figtracer: this still misses `snapshot -> B -> revert`; we have `A -> snapshot -> B -> revert` sorted out already

let's add a test for this scenario, as well as for `vm.txGasPrice()` since it's also tx env-backed might need the same fix

crates/cheatcodes/src/inspector.rs
```diff
@@ -269,6 +269,62 @@ pub struct RecordDebugStepInfo {
...
+/// leave the real env untouched. After the inner transaction returns, the
+/// outer env is restored from the cached snapshot taken before
+/// `transact_inner`, which means:
+///
+/// - the override **does** persist for subsequent `BASEFEE`, `GASPRICE` and `BLOBHASH` reads (this
+///   hook fires in `step_end` regardless of isolation),
+/// - but the real `block.basefee` / `tx.gas_price` / `tx.blob_hashes` do **not** reflect the
+///   cheatcode value, so non-opcode env consumers will not see it. In particular,
+///   `vm.getBlobhashes()` reads `ecx.tx().blob_versioned_hashes()` directly and will return the
+///   pre-isolation value.
```
figtracer: these are stale after our changes

crates/cheatcodes/src/evm.rs
```diff
@@ -1338,11 +1376,42 @@ pub(super) fn get_nonce<FEN: FoundryEvmNetwork>(
...
+            ccx.ecx.tx_mut().set_blob_hashes(hashes);
+            ccx.ecx.tx_mut().set_tx_type(EIP4844_TX_TYPE_ID);
+        }
+        None => {
+            ccx.ecx.tx_mut().set_blob_hashes(vec![]);
+        }
+        _ => {}
+    }
+}
+
```
stevencartavia, line 1444: the `None` arms unconditionally write `gas_price = 0` and `blob_hashes =[]` to the real tx, assuming "no override at snapshot time" means the ambient tx was zero. Would this break any setup with a non-zero ambient `tx.gas_price` - `--gas-price`, `gas_price` in `foundry.toml`, or fork mode?

crates/cheatcodes/src/evm.rs
```diff
@@ -1338,11 +1376,42 @@ pub(super) fn get_nonce<FEN: FoundryEvmNetwork>(
...
+    }
+    match &ccx.state.env_overrides.blob_hashes {
+        Some(h) if !ccx.state.in_isolation_context => {
+            let hashes = h.clone();
+            ccx.ecx.tx_mut().set_blob_hashes(hashes);
+            ccx.ecx.tx_mut().set_tx_type(EIP4844_TX_TYPE_ID);
+        }
+        None => {
+            ccx.ecx.tx_mut().set_blob_hashes(vec![]);
+        }
```
stevencartavia, line 1440: should this arm also restore `tx_type`?

crates/cheatcodes/src/inspector.rs
```diff
@@ -2120,6 +2266,64 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
...
+    /// `vm.txGasPrice` and `vm.blobhashes` visible to called contracts under
+    /// `--isolate` / `--gas-report`, where the inner transaction zeroes the
+    /// real fee fields for fee-accounting purposes.
+    ///
+    /// We can't read the just-executed opcode from `interpreter.bytecode.opcode()`
+    /// here because the PC has already advanced; instead `step` stashes it in
+    /// `env_overrides.pending_opcode` for us.
+    #[cold]
+    fn apply_env_overrides(&mut self, interpreter: &mut Interpreter) {
+        let Some(opcode) = self.env_overrides.pending_opcode.take() else { return };
```
stevencartavia: Since the tracer's `step_end` runs before `Cheatcodes::step_end` in `InspectorStack::step_end_inlined` https://github.com/foundry-rs/foundry/blob/master/crates/evm/evm/src/inspectors/stack.rs#L979-L997 (different lines bc of the PR), the stack rewrite here happens after the tracer has already snapshotted the pre-override value. Would `forge test -vvvv` under `--isolate` show `BASEFEE`/`GASPRICE`/`BLOBHASH` returning 0 while the contract actually observes the cheatcode-set value?

mablr: Makes sense, but I'm unable to reproduce with `forge test -vvvv --isolate`. Could you please share the repro @stevencartavia ?

stevencartavia: re-checked and I think I was mistaken here

stevencartavia (dismissed):

figtracer (dismissed):

crates/cheatcodes/src/evm.rs
```diff
@@ -1335,11 +1373,61 @@ pub(super) fn get_nonce<FEN: FoundryEvmNetwork>(
...
+        _ => {}
+    }
+    match &ccx.state.env_overrides.blob_hashes {
+        Some(h) if !ccx.state.in_isolation_context => {
+            let hashes = h.clone();
+            ccx.ecx.tx_mut().set_blob_hashes(hashes);
+            ccx.ecx.tx_mut().set_tx_type(EIP4844_TX_TYPE_ID);
+        }
+        None => {
+            ccx.ecx.tx_mut().set_blob_hashes(vec![]);
```
zerosnacks: **Correctness (medium): pre-existing `tx.blob_hashes` lost on `revertToState`.**

This branch hard-codes `set_blob_hashes(vec![])` whenever `env_overrides.blob_hashes` is `None` after restore. That's wrong when the tx already had non-empty blob hashes _before_ any `vm.blobhashes(...)` override.

Repro:
1. start from a tx env with real blob hashes (e.g. EIP-4844 fork mode)
2. `snapshotState()`
3. `vm.blobhashes(newHashes)`
4. `revertToState(id)`

Expected: original blob hashes restored. Actual: `tx.blob_hashes` is wiped to `[]`, so both `vm.getBlobhashes()` and the `BLOBHASH` opcode return zero. `pre_override_tx_type` already restores the tx type symmetrically — the hashes themselves need the same treatment.

Suggested fix:
- add `pre_override_blob_hashes: Option<Vec<B256>>` to `EnvOverrides`
- capture it in `inner_snapshot_state` when `env_overrides.blob_hashes.is_none()` (mirror of the `pre_override_gas_price` / `pre_override_tx_type` capture)
- restore it here instead of `vec![]`

testdata/default/cheats/Blobhashes.t.sol
```diff
@@ -15,3 +15,141 @@ contract BlobhashesTest is Test {
...
+}
+
+contract BlobhashRecorder {
+    mapping(uint256 => bytes32) public hash;
+
+    function record() external {
+        hash[0] = blobhash(0);
+        hash[1] = blobhash(1);
+    }
+}
```
zerosnacks, line 237: **Coverage (low): missing regression for snapshot taken with pre-existing real blob hashes.**

The new tests cover:
- override visibility under isolation
- snapshot/revert when the override existed before snapshot (`A -> snap -> B -> revert`)
- the `snap -> B -> revert` no-prior-A scenario

But none cover the case where the tx already has real blob hashes _before_ any cheatcode override, then `snapshotState() -> vm.blobhashes(B) -> revertToState(id)` is expected to bring back the original hashes. That gap is what would let the `set_blob_hashes(vec![])` issue flagged above slip through. Worth adding both the `revertToState` and `revertToStateAndDelete` variants, mirroring the gas-price coverage.

zerosnacks (commented):
Final-pass review: one correctness issue around snapshot/revert of pre-existing blob hashes plus a small coverage gap.

testdata/default/cheats/Blobhashes.t.sol
```diff
@@ -15,3 +15,195 @@ contract BlobhashesTest is Test {
...
+
+        bytes32[] memory newHashes = new bytes32[](1);
+        newHashes[0] = bytes32(uint256(0xEF01));
+        vm.blobhashes(newHashes);
+
+        vm.revertToStateAndDelete(id);
+
+        bytes32[] memory after_ = vm.getBlobhashes();
+        assertEq(after_.length, 1, "should have original 1 hash after revertAndDelete");
+        assertEq(after_[0], original[0], "hash[0] must match original");
```
zerosnacks, line 226: The new `BlobhashesPreExistingRevertToStateTest` does not actually exercise the `pre_override_blob_hashes` restore path it was added for. `vm.blobhashes(original)` here sets `env_overrides.blob_hashes = Some(original)`, so at `vm.snapshotState()` the `if env_overrides.blob_hashes.is_none()` branch in `inner_snapshot_state` is false and `pre_override_blob_hashes` is never captured. On `revertToState`, restoration goes through the `Some(h)` arm of `sync_tx_after_env_override_restore`, not the new `None` arm — i.e. the bug-fix code path is untested.

To genuinely exercise it the test needs `env_overrides.blob_hashes == None` at snapshot time with `tx.blob_hashes` already non-empty, which is not reachable from Solidity today (no cheatcode mutates `tx.blob_hashes` without also setting the override). Suggested options:
- a small Rust-level integration test in `crates/cheatcodes/` that pre-seeds `ccx.ecx.tx_mut().set_blob_hashes(...)` directly, leaves the override unset, then drives `snapshotState` → `vm.blobhashes(new)` → `revertToState` and asserts `tx.blob_hashes` / `vm.getBlobhashes()` / `tx_type` are restored to the original;
- or a fork-mode test against a real EIP-4844 tx.

testdata/default/cheats/TxGasPrice.t.sol
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
```
zerosnacks, line 1: Coverage gap for the analogous `gas_price` case raised earlier (the `None` arm restoring `pre_override_gas_price`): there is no regression for a non-zero ambient `tx.gas_price`. Unlike `blob_hashes` this one is easy to cover from Solidity — set `gas_price` in the test profile (`foundry.toml`) or via `--gas-price`, then `snapshotState` → `vm.txGasPrice(other)` → `revertToState` and assert `tx.gasprice` is restored to the ambient value, not zeroed. Worth adding both `revertToState` and `revertToStateAndDelete` variants to mirror the existing blob-hashes coverage.

testdata/default/cheats/Blobhashes.t.sol
mablr: Added rust it test in https://github.com/foundry-rs/foundry/pull/14493/commits/529c5f660390fbe401e77986ff4d9f00072a0c72

crates/cheatcodes/src/inspector.rs
```diff
@@ -572,6 +638,36 @@ pub struct Cheatcodes<FEN: FoundryEvmNetwork = EthEvmNetwork> {
...
     // Custom execution evm version.
     pub execution_evm_version: Option<SpecFor<FEN>>,
+
+    /// Opcode-level environment overrides for `BASEFEE`, `GASPRICE` and
+    /// `BLOBHASH`. Set by `vm.fee`, `vm.txGasPrice`, `vm.blobhashes` and
+    /// applied in [`Inspector::step_end`].
+    ///
+    /// Needed because in isolation mode the synthetic inner transaction
+    /// zeroes the corresponding tx/block env fields for fee-accounting.
+    pub env_overrides: EnvOverrides,
```
grandizzy: should we keep this mapped per fork so when we switch forks (using cheatcodes) to apply the proper env overrides in `step_end`? Otherwise in multi fork tests we could end up with unwanted env

mablr: Good catch! https://github.com/foundry-rs/foundry/pull/14493/commits/d5cc9a3b20087cc38db2a3d3d1d034979bafde2d added mapping and a test.

crates/cheatcodes/src/inspector.rs
```diff
@@ -1272,6 +1385,33 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>> for Cheatcode
         if self.gas_metering.recording {
             self.meter_gas_record(interpreter, ecx);
         }
+
+        // Capture the opcode for `step_end` to use, since by the time
+        // `step_end` runs the PC has already advanced past it. Also peek the
+        // BLOBHASH index now (still on top of stack before execution) so we
+        // can look up the override later.
+        let fork_id = ecx.db().active_fork_id();
```
grandizzy: these are executed on each opcode and have high impact on performance, suggest here 
```rust
         if !self.env_overrides.is_empty() {                                                                                                           
             let fork_id = ecx.db().active_fork_id(); 
             if let Some(env_overrides) = 
                 self.env_overrides.get_mut(&fork_id).filter(|o| o.is_any_set()) 
             { 
                ...
             }
         }
```

same on `step_end` @DaniPopes could you please chime in if other improvements to be done?

grandizzy (changes requested):

crates/cheatcodes/src/inspector.rs
mablr: Good catch, ty, addressed in https://github.com/foundry-rs/foundry/pull/14493/commits/719909e34e966d03fe5ede68b55245be88eb13b3

grandizzy (approved):
lgtm!

figtracer (approved):
think it's gtg 👍

Issue #14381, opened 2026-04-20
feat(lint): feature parity with slither/aderyn


0xrusowsky:
## Missing Lint Detectors in `forge lint`

this checklist compares Slither and Aderyn detectors against `forge lint` and merges equivalent detectors into single items.

> NOTE: some items are blocked until Solar outputs a CFG (control-flow graph), as they need data-flow analysis, taint tracking, or path-sensitive reasoning.

### High

- [ ] `encode-packed-collision` — `abi.encodePacked` with multiple dynamic types causes hash collisions
- [ ] `msg-value-loop` — `msg.value` reused inside a loop (double-spend)
- [ ] `delegatecall-loop` — Payable `delegatecall` inside a loop
- [ ] `rtlo` — Right-to-left-override Unicode character hiding malicious code
- [ ] `incorrect-exp` — `^` (xor) used instead of `**` (exponentiation)
- [ ] `yul-return` — `return` in assembly/Yul halts entire EVM execution, not just the function
- [ ] `suicidal` — Unprotected `selfdestruct` (no access-control modifier)
- [ ] `weak-prng` — `block.timestamp`/`blockhash` used as randomness source
- [ ] `shadowing-state` — Child contract re-declares parent's state variable
- [ ] `controlled-array-length` — Direct assignment to storage array `.length`
- [ ] `array-by-reference` — Storage array modified via memory copy has no effect
- [ ] `function-selector-collision` — Colliding 4-byte function selectors
- [ ] `name-reused` — Contract name reused across inheritance tree
- [ ] `enumerable-loop-removal` — Removing items from `EnumerableSet` inside a loop using `at` + `remove` corrupts order

### Medium

- [ ] `tx-origin` — `tx.origin` used for authorization
- [ ] `mapping-deletion` — `delete` on struct containing a mapping doesn't clear it
- [ ] `tautological-compare` — `x >= x` always true
- [ ] `tautology` — `uint >= 0` always true, `uint8 < 512` always true
- [ ] `erc20-interface` — ERC20 functions with wrong return types
- [ ] `erc721-interface` — ERC721 functions with wrong return types
- [ ] `boolean-cst` — `if(false)` or `b || true` — misused boolean constant
- [ ] `domain-separator-collision` — Function selector collides with EIP-2612 `DOMAIN_SEPARATOR()`
- [ ] `reused-constructor` — Base constructor called with args from two inheritance paths
[...]

PR #14495, opened 2026-04-28
feat(lint): add could-be-immutable lint

figtracer:
## Summary
- add a gas lint for state variables that can be declared immutable
- detect constructor-only assignments and deployment-time state initializers using Solidity HIR
- add focused lint UI fixture coverage and list the lint in forge-lint docs

tracking: https://github.com/foundry-rs/foundry/issues/14381


crates/lint/src/sol/gas/immutable.rs
```diff
@@ -0,0 +1,411 @@
...
+    fn check_contract(
+        &mut self,
+        ctx: &LintContext,
+        hir: &'hir hir::Hir<'hir>,
+        contract: &'hir hir::Contract<'hir>,
+    ) {
+        if contract.kind == ast::ContractKind::Interface {
+            return;
+        }
+        let Some(contract_id) = contract_id(hir, contract) else { return };
```
mablr: I think you can use instead `LateLintPass::check_nested_contract` then you get `contract_id` for free, then O(n) `contract_id` helper can be dropped.

crates/lint/src/sol/gas/immutable.rs
```diff
@@ -0,0 +1,411 @@
...
+                        function,
+                        &candidate_set,
+                        &mut constructor_writes,
+                        &mut runtime_writes,
+                        &mut HashSet::new(),
+                    );
+
+                    if let Some(body) = function.body {
+                        collect_state_writes(hir, body, &candidate_set, &mut constructor_writes);
+                    }
```
mablr, line 76: It seems that if a constructor calls an internal function that writes a state variable, that write is not detected. The variable isn't flagged even though it's only ever written at construction time. Imo it worths a comment explaining the limitation (and a test?).

mablr (commented):
some remarks

crates/lint/src/sol/gas/immutable.rs
figtracer: cool, will change this

figtracer:
mb for the force-push @mablr

mablr (approved):

zerosnacks (approved):

PR #14505, opened 2026-04-29
feat(cast): open Tempo wallet fund flow for MPP failures

figtracer:
## Motivation

When cast hits a paid Tempo MPP RPC with an underfunded wallet, the command currently ends at a 402 with manual instructions. This makes the no-balance path abrupt and forces users to restart the command after funding.

## Solution

- Detect Tempo MPP funding failures, including insufficient-balance responses and generic verification-failed 402s from paid RPCs.
- In interactive terminals, launch `tempo wallet fund` directly and retry the original request after the funding command succeeds.
- Pass the known funding wallet address and Tempo network to `tempo wallet fund` when available.
- Surface the requested payment token in the funding message so users know what balance is missing.
- Keep clear `tempo wallet fund --no-browser` fallback instructions for CI, remote, and headless environments.

## Notes

`tempo wallet fund` currently supports `--address`, `--network`, and `--no-browser`; it does not expose a token-specific or swap-page argument yet. This PR includes the requested token in the UX, but token-specific funding/swap routing will need wallet/CLI support.


crates/common/src/provider/mpp/transport.rs
```diff
@@ -42,6 +47,31 @@ fn default_deposit() -> u128 {
...
+            .unwrap_or_default()
+    }
+
+    const fn network(&self) -> Option<&'static str> {
+        match self.chain_id {
+            // Tempo mainnet/testnet naming as understood by the Tempo wallet CLI.
+            Some(4217) => Some("tempo"),
+            Some(42431) => Some("moderato"),
+            _ => None,
+        }
```
mablr: Let's use `ChainId` type for `FundingContext::chain_id` field, to get `is_tempo()` and `to_string()` methods.

crates/common/src/provider/mpp/transport.rs
```diff
@@ -346,36 +594,43 @@ where
...
             let needs_key_provisioning = problem_type.ends_with("/key-not-provisioned")
                 || detail.contains("access key does not exist")
                 || detail.contains("key is not provisioned");
 
-            let needs_verification_retry = (problem_type.ends_with("/verification-failed")
-                || detail.contains("verification-failed"))
-                && self.provider.is_key_provisioned();
-
-            if needs_key_provisioning || needs_verification_retry {
+            if needs_key_provisioning {
```
mablr: The old code retried key provisioning on verification-failed when is_key_provisioned(). This seems to be a regression?

crates/common/src/provider/mpp/transport.rs
```diff
@@ -569,13 +927,28 @@ pub(crate) trait ResolveProvider {
     }
     fn resolve_for(&self, opts: DiscoverOptions) -> TransportResult<Self::Provider>;
     fn set_key_provisioned(&self, _provisioned: bool) {}
+    #[allow(dead_code)]
     fn is_key_provisioned(&self) -> bool {
         true
     }
```
mablr: This leaked from the `needs_verification_retry` removal.

crates/common/src/provider/mpp/transport.rs
```diff
@@ -60,6 +90,168 @@ fn format_http_diagnostics(headers: &HeaderMap) -> String {
...
+        ctx.token_line()
+    )
+}
+
+/// Decide whether the interactive `tempo wallet fund` flow may be launched.
+///
+/// Policy (library-safe):
+/// - never run inside CI
+/// - never run unless both stdin and stderr are real terminals
+/// - `FOUNDRY_MPP_AUTO_FUND` is honored only as an *opt-out* (`0`/`false`/`off`); it must not
```
mablr: For a simple opt-out, i think we can make the semantic simpler w/ `FOUNDRY_MPP_NO_AUTO_FUND`.

mablr (changes requested):
Looks good overall, but there are some things to fix.

crates/common/src/provider/mpp/transport.rs
```diff
@@ -60,6 +86,167 @@ fn format_http_diagnostics(headers: &HeaderMap) -> String {
...
+    let mut command = "tempo wallet fund".to_string();
+    if let Some(address) = ctx.wallet_address {
+        command.push_str(&format!(" --address {address}"));
+    }
+    if let Some(network) = ctx.network() {
+        command.push_str(&format!(" --network {network}"));
+    }
+
+    let mut no_browser = command.clone();
+    no_browser.push_str(" --no-browser");
```
mablr, line 99: idk if this can be done in a safer way

mablr (approved):
lgtm

crates/common/src/provider/mpp/transport.rs
figtracer: i think we can keep as-is for now, if something pops up we'll fix it

zerosnacks (approved):
In the short term OK with merging this; longer term I would like to see this kind of Tempo wallet specific handling abstracted away from Foundry

PR #14508, opened 2026-04-29
feat(cast): introduce `vaddr` cmd for TIP-1022

mablr:
## Motivation

Bring TIP-1022 support to cast

Added these commands:
- `cast vaddr create` derive and register a new virtual address.
- `cast vaddr resolve` look up its canonical destination.
- `cast vaddr watch` tail incoming transfers to the alias for monitoring.

## PR Checklist

- [x] Added Tests (added for `create`/`resolve`, very slow due to pow -> located in Tempo CI script)
- [x] Added Documentation
- [ ] Breaking changes


crates/cast/src/cmd/vaddr/create.rs
```diff
@@ -0,0 +1,147 @@
...
+            .checked_add(i as u64)
+            .filter(|&t| t <= MAX_USER_TAG)
+            .ok_or_else(|| eyre::eyre!("tag overflow: tag + count exceeds the 6-byte user tag range (max {MAX_USER_TAG:#x})"))?;
+        let raw = tag_value.to_be_bytes();
+        let user_tag = UserTag::new(raw[2..].try_into().expect("slice is 6 bytes"));
+        let vaddr = Address::new_virtual(output.master_id, user_tag);
+        sh_println!("  tag={user_tag}  {vaddr}")?;
+    }
+
+    register(owner, output.salt, send_tx, tx_opts).await
```
figtracer, line 105: it might be worth splitting derivation from registration, so users can precompute vaddrs without needing a signer/RPC transaction

crates/cast/src/cmd/vaddr/create.rs
```diff
@@ -0,0 +1,147 @@
...
+        "Salt:              {}
+Registration hash: {}
+Master ID:         {}",
+        output.salt,
+        output.registration_hash,
+        output.master_id,
+    )?;
+
+    const MAX_USER_TAG: u64 = 0x0000_FFFF_FFFF_FFFF;
+    sh_println!("\nVirtual addresses:")?;
```
figtracer, line 89: should `--count 0` just be rejected?

crates/cast/src/cmd/vaddr/create.rs
mablr: added `--no-register` flag

figtracer (dismissed):
lg

crates/cast/src/cmd/vaddr/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+        #[arg(long, value_name = "ADDRESS")]
+        token: Address,
+
+        #[command(flatten)]
+        send_tx: SendTxOpts,
+
+        #[command(flatten)]
+        tx: TxParams,
+    },
+}
```
0xrusowsky: i don't think we need this cmd in practice

crates/cast/src/cmd/vaddr/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+        #[command(flatten)]
+        send_tx: SendTxOpts,
+
+        #[command(flatten)]
+        tx: TxParams,
+    },
+
+    /// Resolve a virtual address to its registered master and decode its components.
+    #[command(visible_alias = "r")]
+    Resolve {
```
0xrusowsky, line 66: maybe a useful addition for businesses would be a flag to check whether it has already been used or not.

while we can't ensure that an virtual address hasn't been assigned to a customer, we can check if it already appeared on a transfer event.

0xrusowsky (commented):
directionally LGTM, left some suggestions though

crates/cast/src/cmd/vaddr/mod.rs
mablr: removed

crates/cast/src/cmd/vaddr/mod.rs
mablr: @0xrusowsky In practice we can access to the latest 100_000 blocks, if tx was before it would return unused, is this fine ?

0xrusowsky: i think if we can only do 100_000 blocks not that useful, and better to leave it out, as it could confuse people and be even worse

0xrusowsky (approved):

zerosnacks (approved):

figtracer (approved):

PR #14511, opened 2026-04-29
feat(forge): add coverage attribution report

smartcontracts:
## Human content here

I've been working on some tooling that would benefit from a reverse mapping from test => lines covered by that test. Current coverage does not generate this unless you trigger coverage for each test individually which is far too slow in practice. Foundry already had the core capabilities for this, so I just added a new output mode that spits this data out. The resulting files can be VERY large, so this isn't really meant for average use, but for the specialized use-case it is quite powerful. 

Everything below is generated by Codex 5.5
---

## Motivation

`forge coverage` currently reports aggregate coverage, but it does not expose which
individual tests hit which source items. Tooling that selects targeted verification
work needs that attribution to map source changes back to the smallest useful set of
tests instead of treating every covered file as a broad bucket.

This PR adds a machine-readable attribution report while preserving the existing
aggregate coverage reports.

## Solution

Add `forge coverage --report=attribution`, which writes
`coverage-attribution.json` at the project root.

The report contains one entry per executed test with:

- suite and test name
- test status and kind
- covered source items with source path, contract, item kind, byte range, line
  range, hit count, and function or branch metadata where available

Implementation notes:

- `foundry-evm-coverage` exposes a non-mutating `hit_items_for_hit_map` helper so
  callers can inspect which coverage items a single hit map touched.
- `forge coverage` writes attribution directly from each test result's recorded
  line coverage.
- Sources excluded by `--no-match-coverage` are filtered from attribution output
  the same way they are filtered from aggregate reports.
- The attribution JSON is streamed through serde instead of first materializing the
  full report payload.
- Existing summary, LCOV, bytecode, and debug report behavior is unchanged.

## Testing

Passed:

```sh
cargo test -p forge --test cli coverage::attribution_report -- --exact
cargo +nightly fmt -- --check
cargo +nightly clippy -p forge --all-targets --all-features -- -D warnings
```

Earlier during development, broader checks were also run successfully before the
final streaming cleanup:

```sh
cargo check --all
cargo check -p forge
cargo +nightly clippy --workspace --all-targets --all-features -- -D warnings
```

## AI Assistance Disclosure

AI assistance was used for code generation, cleanup, and local validation of this
PR.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mablr (dismissed):
lgtm, thanks👍

crates/forge/src/cmd/coverage.rs
```diff
@@ -287,6 +288,17 @@ impl CoverageArgs {
             }
         }
 
+        if self.report.iter().any(|kind| matches!(kind, CoverageReportKind::Attribution)) {
+            let reporter =
+                CoverageAttributionReporter::new(project_root.join("coverage-attribution.json"));
+            reporter.report(
```
zerosnacks: Please use the `self.report_file` / `--report-file` flag so users can control the output location, falling back to `coverage-attribution.json`

zerosnacks (dismissed):
Small change requested per comment above, overall lgtm - thanks!

crates/forge/src/cmd/coverage.rs
smartcontracts: Updated, thank you!

mablr (dismissed):

mablr:
cc @figtracer @grandizzy  :)

crates/forge/src/coverage.rs
```diff
@@ -218,6 +220,257 @@ impl CoverageReporter for LcovReporter {
...
+        out.flush()?;
+
+        sh_println!("Wrote coverage attribution report.")?;
+
+        Ok(())
+    }
+}
+
+#[derive(Serialize)]
+struct AttributionReport<'a> {
```
grandizzy, line 266: pls document, short and relevant

crates/forge/src/coverage.rs
```diff
@@ -218,6 +220,257 @@ impl CoverageReporter for LcovReporter {
...
+    version: u8,
+    tests: AttributionTests<'a>,
+}
+
+#[derive(Serialize)]
+struct AttributionTest {
```
grandizzy, line 273: pls document, short and relevant

crates/forge/src/coverage.rs
```diff
@@ -218,6 +220,257 @@ impl CoverageReporter for LcovReporter {
...
+    suite: String,
+    test: String,
+    status: &'static str,
+    kind: &'static str,
+    covered: Vec<AttributionItem>,
+}
+
+#[derive(Serialize)]
+struct AttributionItem {
```
grandizzy, line 283: pls document, short and relevant

crates/forge/src/coverage.rs
```diff
@@ -218,6 +220,257 @@ impl CoverageReporter for LcovReporter {
...
+    hits: u32,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    function: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    branch_id: Option<u32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    path_id: Option<u32>,
+}
+
+struct AttributionTests<'a> {
```
grandizzy, line 305: pls document, short and relevant

crates/forge/src/coverage.rs
```diff
@@ -218,6 +220,257 @@ impl CoverageReporter for LcovReporter {
...
+        u32,
+        Option<String>,
+        Option<u32>,
+        Option<u32>,
+    );
+
+    let mut items = BTreeMap::<AttributionItemKey, AttributionItem>::new();
+    let Some(hit_maps) = result.line_coverage.as_ref() else { return Vec::new() };
+
+    for map in hit_maps.0.values() {
```
grandizzy: that's basically done already and same as https://github.com/smartcontracts/foundry/blob/ab3d19133fcb52e55a7d35240b1a806a2096467f/crates/forge/src/cmd/coverage.rs#L268-L275 can we build a BytecodeResolver in collect and then pass it here so we won't add any performance penalty? otherwise we re-run these lookups per hitmap per test

crates/forge/src/cmd/coverage.rs
```diff
@@ -290,6 +294,18 @@ impl CoverageArgs {
             }
         }
 
+        if self.report.iter().any(|kind| matches!(kind, CoverageReportKind::Attribution)) {
```
figtracer: this writes attribution before the normal reporters, so `forge coverage --report=attribution --report=lcov --report-file combined.out` first writes JSON and then LCOV overwrites the output

i think we're expecting the final result to be:
```
coverage-attribution.json
lcov.info
```

without `combined.out` in the case where we want different output report types

crates/forge/src/cmd/coverage.rs
```diff
@@ -290,6 +294,18 @@ impl CoverageArgs {
...
+            let reporter = CoverageAttributionReporter::new(
+                self.report_path(project_root, "coverage-attribution.json"),
+            );
+            reporter.report(
+                &report,
+                &outcome,
+                filter.paths().root.as_path(),
+                filter.args().coverage_pattern_inverse.as_ref(),
+            )?;
```
grandizzy: report runs before lcov/summary/debug and errors out on missing known_contracts. If attribution fails, the user loses every other report they requested. Run attribution last, or write other reports before surfacing the attribution error.

crates/forge/src/cmd/coverage.rs
grandizzy: Attribution runs here, before the aggregate report has had coverage_pattern_inverse applied via `report.retain_sources(...)` on line 312. The reporter re-implements the same filter internally (`include_source` + `strip_prefix(file_root)` in `coverage.rs`), which risks divergence.
Move this block after the `retain_sources` call (line 309–316) so attribution operates on the already-filtered report and the duplicate `include_source` helper + `coverage_pattern_inverse` parameter can be dropped entirely.

grandizzy (changes requested):
thank you, overall lgtm. left couple of suggestions, ptal + add short and relevant documentation for structs and fns newly introduced

mablr (dismissed):
LGTM

crates/forge/src/cmd/coverage.rs
```diff
@@ -393,6 +423,13 @@ impl CoverageArgs {
         // Output final reports.
         self.report(&report)?;
 
+        if self.report.iter().any(|kind| matches!(kind, CoverageReportKind::Attribution)) {
+            let reporter = CoverageAttributionReporter::new(
+                self.report_path(project_root, "coverage-attribution.json"),
+            );
+            reporter.report(&report, &outcome, &resolved_hit_maps)?;
```
stevencartavia, line 436: `--report=attribution` mis-attributes coverage when a filtered source (`--no-match-coverage` / `coverage.skip_files`) precedes retained items in `all_items` order. Attribution runs after `retain_sources()`, which shrinks `all_items` via `Vec::retain` but doesn't update the anchors; `hit_items_for_hit_map` then resolves absolute `item_id`s against the shifted Vec, so hits land on the wrong items. `source_paths` isn't pruned either, so the excluded source isn't filtered out.

Repro: contracts W(`src/sub/`), X, Y, Z hit 2/4/6/8 times. Baseline correct; with `--no-match-coverage "sub/"`, W's 2 hits appear as `test/test.sol failed hits=2` (never called). Lower-source-id exclusions make kept contracts report each other's counts.

Fix: filter attribution items before `retain_sources` with the same predicates (aggregate/lcov/summary are fine since `add_hit_map` runs before retain). Add a regression test with a filtered middle source.

crates/forge/src/coverage.rs
```diff
@@ -218,6 +220,219 @@ impl CoverageReporter for LcovReporter {
...
+                continue;
+            };
+
+            let source = source_path.display().to_string();
+            let contract = item.loc.contract_name.to_string();
+            let (kind, function, branch_id, path_id) = coverage_item_kind_fields(&item.kind);
+            let line_start = item.loc.lines.start;
+            let line_end = item.loc.lines.end.saturating_sub(1);
+            let byte_start = item.loc.bytes.start;
+            let byte_end = item.loc.bytes.end;
```
stevencartavia: `version: 1` JSON mixes range conventions: `line_end = lines.end.saturating_sub(1)` (inclusive) but `byte_end = bytes.end` (exclusive), and `line_start` is 1-based vs `byte_start` 0-based. Repro: `return x*2;` on line 4 reports `line[4..4] byte[123..133]` (byte 133 is the `;`). Undocumented — worth settling before v1 ships: emit both half-open, or document it.

crates/forge/src/cmd/coverage.rs
```diff
@@ -106,6 +107,21 @@ pub struct CoverageArgs {
...
+            (self.file_report_count() == 1).then_some(self.report_file.as_deref()).flatten();
+        root.join(report_file.unwrap_or_else(|| Path::new(default_file_name)))
+    }
+
+    fn file_report_count(&self) -> usize {
+        self.report
+            .iter()
+            .filter(|kind| {
+                matches!(kind, CoverageReportKind::Lcov | CoverageReportKind::Attribution)
+            })
```
stevencartavia: With >1 file report (lcov + attribution), `report_file` is silently ignored — `report_path()` uses it only when `file_report_count() == 1`, so both fall back to defaults with no warning. Repro: `--report=attribution --report=lcov --report-file swap.out` writes `lcov.info`/`coverage-attribution.json`, not `swap.out`; same via `foundry.toml`. Suggest a `sh_warn!` when `report_file.is_some() && file_report_count() > 1` and updating the `--report-file` help text (the `attribution_report_with_lcov_uses_default_paths` test would need updating too).

stevencartavia (dismissed):

mablr (dismissed):
LGTM, pending @grandizzy

crates/evm/coverage/src/lib.rs
```diff
@@ -147,19 +147,41 @@ impl CoverageReport {
...
     pub fn retain_sources(&mut self, mut predicate: impl FnMut(&Path) -> bool) {
-        self.analyses.retain(|version, analysis| {
-            analysis.all_items_mut().retain(|item| {
-                self.source_paths
-                    .get(&(version.clone(), item.loc.source_id))
-                    .map(|path| predicate(path))
-                    .unwrap_or(false)
-            });
-            !analysis.all_items().is_empty()
+        self.source_paths.retain(|_, path| predicate(path));
```
grandizzy, line 182: Filtering `source_paths` instead of pruning `analyses` is right for attribution (keeps the `anchor.item_id`-as-index invariant), but it regresses `DebugReporter`, which the PR lists as "unchanged." Its anchors section resolves `report.analyses.get(item_id)` directly with no `source_paths` check, so items from sources excluded by `--no-match-coverage` / `coverage.skip_files` now still print.

Fix: filter anchor items through `source_paths` in `DebugReporter`, and only print the `Anchors for ...` header when at least one anchor survives:

```rust
let Some(item) = report
    .analyses
    .get(&contract_id.version)
    .and_then(|items| items.get(anchor.item_id))
else {
    continue;
};
if !report.source_paths.contains_key(&(contract_id.version.clone(), item.loc.source_id)) {
    continue;
}
sh_println!("- {kind} {anchor}: {item}")?;
```

grandizzy (dismissed):
Please check finding inline.

grandizzy (approved):
lgtm, thanks!

mablr (approved):

PR #14512, opened 2026-04-29
fix(forge): warn when constructor args are ignored for constructor-less contracts

AissataF:
`forge create` previously dropped `--constructor-args` and `--constructor-args-path` silently when the target contract had no constructor, so users had no way to tell their values were never encoded. Emit `sh_warn!` in that branch, naming the contract, so the misuse surfaces at deploy time.


mablr (approved):
Not very important but makes sense

stevencartavia (approved):

AissataF:
greeting @stevencartavia @mablr anything I need to fix before merging?

PR #14516, opened 2026-04-30
feat(script): land entire deployment in one atomic batch transaction

stevencartavia:
Use Tempo's batch tx capability so that `forge script --batch` lands the entire deployment script as one atomic onchain operation.

CLOSES OSS-140


crates/script/src/broadcast.rs
```diff
@@ -727,8 +727,15 @@ impl BundledState<TempoEvmNetwork> {
...
-        // Collect all transactions into Call structs
-        // Tempo batch transactions support CREATE only as the first call
+        // CREATE2 deployer must exist on-chain for the rewritten CREATEs.
+        let create2_deployer = self.script_config.evm_opts.create2_deployer;
+        if provider.get_code_at(create2_deployer).await?.is_empty() {
+            bail!(
+                "CREATE2 deployer {create2_deployer:#x} is not deployed on this Tempo network; \
+                 --batch requires it. Deploy it first and retry."
+            );
+        }
```
mablr: I think this is not restrictive, scripts without CREATE should not tigger an error when there is no `create2_deployer`.

crates/script/src/broadcast.rs
```diff
@@ -862,26 +869,29 @@ impl BundledState<TempoEvmNetwork> {
...
-            let deployed_addr = sender.create(nonce);
-            sh_println!("Contract deployed at: {:#x}", deployed_addr)?;
-            Some(deployed_addr)
-        } else {
-            None
-        };
+        // Use simulation's contract_address; fall back to sender.create(nonce) for legacy path.
+        let per_tx_addresses: Vec<Option<Address>> = sequence
+            .transactions
+            .iter()
```
mablr, line 1408: `sequence.transactions` and `calls` can diverge (some txs are skipped via continue when building calls). If they go out of sync, you push the wrong number of receipts and corrupt the broadcast JSON. Let's build `per_tx_addresses` from `calls` instead, or assert the lengths match.

zerosnacks:
Directionally I like this PR but I want to give it some more time before merging and more eyes on it because silently rewriting CREATE to CREATE2 may have unintended side effects

crates/evm/evm/src/inspectors/stack.rs
```diff
@@ -1051,8 +1053,18 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>>
...
+                CreateScheme::Create2 { salt } => Some(salt),
+                // --batch: nonce makes re-runs unique, counter makes each deploy unique.
+                CreateScheme::Create => {
+                    let counter = self.inner.batch_create_counter;
+                    self.inner.batch_create_counter = counter.wrapping_add(1);
+                    let nonce = ecx.journal_mut().load_account(inputs.caller()).ok()?.info.nonce;
+                    Some(U256::from(nonce) << 64 | U256::from(counter))
+                }
+                _ => None,
+            }
```
zerosnacks: The rewrite changes observable semantics with no user signal:
- deployed address becomes `keccak(0xff || factory || salt || keccak(init))[12:]`; `vm.computeCreateAddress(sender, nonce)` and external indexers go stale.
- constructor `msg.sender` is the factory, not the EOA.
- `inputs.value()` is forwarded as `CallValue::Transfer` via `get_create2_factory_call_inputs` (`crates/evm/core/src/evm/mod.rs:210-225`); the Arachnid factory is not payable, so `new Foo{value: x}()` reverts in simulation.

Gate behind `--batch-rewrite-creates`, or emit `sh_warn!` once and add a `rewrote N CREATE(s) via 0x4e59…956c` line to the post-broadcast summary.

crates/evm/evm/src/inspectors/stack.rs
```diff
@@ -1051,8 +1053,18 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>>
...
             && self.should_use_create2_factory(ecx.journal().depth(), inputs)
+            && let Some(salt) = match inputs.scheme() {
+                CreateScheme::Create2 { salt } => Some(salt),
+                // --batch: nonce makes re-runs unique, counter makes each deploy unique.
+                CreateScheme::Create => {
+                    let counter = self.inner.batch_create_counter;
+                    self.inner.batch_create_counter = counter.wrapping_add(1);
+                    let nonce = ecx.journal_mut().load_account(inputs.caller()).ok()?.info.nonce;
+                    Some(U256::from(nonce) << 64 | U256::from(counter))
+                }
```
zerosnacks, line 1138: `frame_start` runs before REVM bumps caller nonce, and the rewritten frame is a `Call` which never bumps it. All rewrites in one batch see the same `caller.nonce`; uniqueness within a batch depends solely on `counter`, and `counter` resets per process.

Re-running before the prior batch lands → identical (nonce, counter) → identical salt → CREATE2 collision at the factory.

Mix in `chain_id` (and ideally a per-process value), and add a test asserting that two simulations launched at the same on-chain nonce produce distinct salts.

crates/cheatcodes/src/inspector.rs
```diff
@@ -2020,19 +2020,24 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>> for Cheatcode
...
+
+        if depth != target_depth {
+            return false;
+        }
+
+        match inputs.scheme() {
+            CreateScheme::Create2 { .. } => {
+                self.broadcast.is_some() || self.config.always_use_create_2_factory
+            }
+            CreateScheme::Create => self.config.batch_rewrite_creates && self.broadcast.is_some(),
```
zerosnacks, line 2260: No test covers the `CreateScheme::Create` arm. Add a predicate test: `(flag=false, broadcast=Some)`, `(flag=true, broadcast=None)`, `(flag=true, broadcast=Some, depth != target)`, `(flag=true, broadcast=Some, depth == target)`.

crates/script/src/broadcast.rs
```diff
@@ -817,8 +817,15 @@ impl BundledState<TempoEvmNetwork> {
...
+        if provider.get_code_at(create2_deployer).await?.is_empty() {
+            bail!(
+                "CREATE2 deployer {create2_deployer:#x} is not deployed on this Tempo network; \
+                 --batch requires it. Deploy it first and retry."
+            );
+        }
+
         let mut calls: Vec<Call> = Vec::new();
         let mut has_create = false;
         for (idx, tx) in sequence.transactions().enumerate() {
```
zerosnacks: With the rewrite active, every recorded tx has `tx.to() = Some(factory)`, so `has_create` stays `false` and the loop's `idx > 0` and "only one CREATE" bails below become unreachable.

Drop the dead branches, or expose a no-rewrite mode that keeps them load-bearing.

crates/script/src/broadcast.rs
```diff
@@ -817,8 +817,15 @@ impl BundledState<TempoEvmNetwork> {
...
-        // Collect all transactions into Call structs
-        // Tempo batch transactions support CREATE only as the first call
+        // CREATE2 deployer must exist on-chain for the rewritten CREATEs.
+        let create2_deployer = self.script_config.evm_opts.create2_deployer;
+        if provider.get_code_at(create2_deployer).await?.is_empty() {
+            bail!(
+                "CREATE2 deployer {create2_deployer:#x} is not deployed on this Tempo network; \
+                 --batch requires it. Deploy it first and retry."
+            );
+        }
```
zerosnacks: `get_code_at(...).is_empty()` doesn't catch a non-Arachnid contract at the factory address. Match against `DEFAULT_CREATE2_DEPLOYER_CODEHASH` like the inspector does (`crates/evm/evm/src/inspectors/stack.rs:1076`).

crates/script/src/broadcast.rs
```diff
@@ -969,26 +976,29 @@ impl BundledState<TempoEvmNetwork> {
...
+            .iter()
+            .enumerate()
+            .map(|(idx, tx)| {
+                if idx == 0 && has_create && tx.contract_address.is_none() {
+                    Some(sender.create(nonce))
+                } else {
+                    tx.contract_address
+                }
+            })
+            .collect();
```
zerosnacks, line 1414: `sender.create(nonce)` is unreachable when the rewrite is active — `tx.contract_address` is already populated by `set_call` (`crates/script/src/transaction.rs:37-46`). Replace with:

```rust
debug_assert!(!has_create, "CREATE should have been rewritten");
let per_tx_addresses: Vec<_> = sequence.transactions.iter().map(|tx| tx.contract_address).collect();
```

crates/forge/tests/cli/script.rs
```diff
@@ -3687,6 +3687,119 @@ contract ArbScript is Script {
...
+        .assert_failure()
+        .stderr_eq(str![[r#"
+Error: --batch mode is only supported on Tempo networks
+
+"#]]);
+});
+
+// Dry-run against Tempo Moderato testnet to verify CREATE→CREATE2 rewriting.
+forgetest_async!(
+    #[ignore]
```
zerosnacks, line 3796: `#[ignore]` + dependency on `rpc.moderato.tempo.xyz` means CI never runs the only assertions for `transactionType == "CREATE2"`, factory in `to`, and address uniqueness. Run the dry-run portion against anvil with Arachnid pre-deployed; keep the on-chain broadcast `#[ignore]`'d.

zerosnacks (changes requested):

crates/script/src/broadcast.rs
stevencartavia: addressed in 2fd31b9

crates/script/src/broadcast.rs
stevencartavia: addressed in 2fd31b9

crates/script/src/broadcast.rs
stevencartavia: addressed in 2fd31b9

crates/script/src/broadcast.rs
stevencartavia: addressed in 2fd31b9

crates/script/src/broadcast.rs
stevencartavia: addressed in 2fd31b9

crates/evm/evm/src/inspectors/stack.rs
stevencartavia: addressed in [c64f9db](https://github.com/foundry-rs/foundry/pull/14516/commits/c64f9db411aa88e634bffef875a1314974d78cf0)

crates/evm/evm/src/inspectors/stack.rs
stevencartavia: [808767b](https://github.com/foundry-rs/foundry/pull/14516/commits/808767b0b031fc7861ae04cc23f5104f28d827fa)

crates/cheatcodes/src/inspector.rs
stevencartavia: [7d9648f](https://github.com/foundry-rs/foundry/pull/14516/commits/7d9648ff38cf2887235fc6c4de67a15ee7889f3e)

crates/forge/tests/cli/script.rs
stevencartavia: [bf6084f](https://github.com/foundry-rs/foundry/pull/14516/commits/bf6084fd7ac21a456b57d66ef3a610c18b4d5bfb)

crates/evm/evm/src/inspectors/stack.rs
```diff
@@ -1072,9 +1079,50 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>>
         frame_input: &mut FrameInput,
     ) -> Option<FrameResult> {
         if let FrameInput::Create(inputs) = frame_input
-            && let CreateScheme::Create2 { salt } = inputs.scheme()
             && self.should_use_create2_factory(ecx.journal().depth(), inputs)
         {
+            // The Arachnid factory is not payable, so a rewritten CREATE that carries value
+            // would silently revert at the factory. Surface the limitation up front instead.
+            if matches!(inputs.scheme(), CreateScheme::Create) && !inputs.value().is_zero() {
```
zerosnacks: The value guard only fires for `CreateScheme::Create`. Explicit `CreateScheme::Create2` with non-zero value is also routed through the Arachnid factory via `get_create2_factory_call_inputs` (`crates/evm/core/src/evm/mod.rs:210-225`), which forwards `inputs.value()` as a real `CallValue::Transfer`. Since the Arachnid factory isn't payable, `new Foo{value: x}{salt: s}()` will still silently revert at the factory.

Drop the `matches!(..., CreateScheme::Create)` discriminant so the rejection covers both schemes when factory routing is active.

crates/script/src/broadcast.rs
```diff
@@ -817,32 +820,16 @@ impl BundledState<TempoEvmNetwork> {
...
-                        bail!("Only one contract creation is allowed per --batch transaction.");
-                    }
-                    has_create = true;
-                    TxKind::Create
-                }
-            };
+        for tx in sequence.transactions() {
+            let addr = tx.to().unwrap_or_else(|| {
+                unreachable!("--batch rewrites all CREATEs via the CREATE2 factory")
+            });
```
zerosnacks: `unreachable!()` is too strong here. There are still silent rewrite-skip paths upstream in `frame_start` (caller-nonce load `stack.rs:1120`, deployer-account load `stack.rs:1130`, factory-call-inputs build `stack.rs:1157-1159`). Any of those `ok()?` exits leave the original `CreateScheme::Create` frame intact, which means a recorded tx with `to == None` can reach this loop and panic the broadcaster.

Replace with `bail!()` (or a debug_assert + recoverable error) so a regression surfaces as a script error, not a panic.

crates/evm/evm/src/inspectors/stack.rs
```diff
@@ -1522,3 +1570,51 @@ impl<FEN: FoundryEvmNetwork> DerefMut for InspectorStack<FEN> {
...
+#[cfg(test)]
+mod tests {
+    use super::compute_batch_create_salt;
+
+    #[test]
+    fn distinct_salts_across_simulations_at_same_nonce() {
+        // Two "simulations" with different per-inspector seeds but identical on-chain state.
+        let a = compute_batch_create_salt(0xabcd_ef01_2345_6789, 1, 5, 0);
+        let b = compute_batch_create_salt(0x1122_3344_5566_7788, 1, 5, 0);
+        assert_ne!(a, b);
```
zerosnacks, line 1625: These tests prove `compute_batch_create_salt` is injective in its inputs, but not the failure mode I originally asked about: two *simulations* launched at the same on-chain nonce producing distinct salts/addresses. The `process_salt` is `Option<u64>` lazily initialized via `rand::random` — that's the actual collision-prevention surface and it isn't exercised here.

Add one end-to-end test that constructs two separate `InspectorStackInner` instances (or runs two dry-runs), feeds them the same caller/nonce/chain, and asserts the resulting CREATE2 addresses differ. That's what would catch a regression where someone makes `batch_rewrite_process_salt` deterministic again.

crates/script/src/broadcast.rs
```diff
@@ -1024,6 +1025,13 @@ impl BundledState<TempoEvmNetwork> {
             sh_println!("\n\n==========================")?;
             sh_println!("\nBATCH EXECUTION COMPLETE & SUCCESSFUL.")?;
             sh_println!("All {} calls executed atomically in a single transaction.", calls.len())?;
+            let rewrites = calls
+                .iter()
+                .filter(|c| matches!(c.to, TxKind::Call(a) if a == create2_deployer))
+                .count();
+            if rewrites > 0 {
+                sh_println!("Rewrote {rewrites} CREATE(s) via {create2_deployer:#x}.")?;
```
zerosnacks: This count over-reports rewrites: any batch call whose `to == create2_deployer` is counted, which includes (a) explicit `CREATE2` calls the user wrote and (b) legitimate user calls to that address. Same coarseness applies to the `needs_factory` preflight predicate at line 846-847.

Either drop the line, or track provenance explicitly (e.g., a `was_batch_rewritten: bool` on `TransactionWithMetadata` set in `frame_start` when the CREATE→CREATE2 rewrite fires) and use it for both the count and the factory preflight. The provenance flag would also let you tighten the `unreachable!()` above into a precise check.

zerosnacks (commented):
Thanks for the thorough follow-ups — most of my earlier feedback is addressed. Verified each item against the relevant commit (and ran the new unit tests locally; all 8 pass). Four small follow-ups left, see inline.

crates/evm/evm/src/inspectors/stack.rs
stevencartavia: dropped the discriminant

crates/script/src/broadcast.rs
stevencartavia: swapped for `bail!`

crates/evm/evm/src/inspectors/stack.rs
stevencartavia: added

crates/script/src/broadcast.rs
stevencartavia: dropped the line

mattsse:
Found one blocking regression:

- **P1:** [`crates/evm/evm/src/inspectors/stack.rs:1084`](https://github.com/foundry-rs/foundry/blob/646f1540ec98dfb76bad2b2eb4e5721e9fec0fc1/crates/evm/evm/src/inspectors/stack.rs#L1084) rejects any CREATE2 factory route with non-zero value before distinguishing explicit `CREATE2` from rewritten `--batch` `CREATE`. This regresses payable `new C{salt: s, value: v}()` flows that already rely on the default `0x4e59...` factory, whose runtime forwards `CALLVALUE` into `CREATE2`. `broadcast_batch` also preserves each call value at [`crates/script/src/broadcast.rs:833`](https://github.com/foundry-rs/foundry/blob/646f1540ec98dfb76bad2b2eb4e5721e9fec0fc1/crates/script/src/broadcast.rs#L833), so the guard should be removed or scoped to an actually unsupported path.

crates/script/src/broadcast.rs
```diff
@@ -969,26 +973,27 @@ impl BundledState<TempoEvmNetwork> {
...
+        for (idx, addr) in per_tx_addresses.iter().enumerate() {
+            if let Some(addr) = addr {
+                sh_println!("  call[{idx}] contract address: {addr:#x}")?;
             }
+        }
+
+        for addr in &per_tx_addresses {
+            let mut tx_receipt = receipt.clone();
+            tx_receipt.contract_address = *addr;
             sequence.receipts.push(tx_receipt);
```
mablr, line 1426: `tx.contract_address` is set to the callee for ordinary calls, so cloned receipts get a bogus deployment `address`, and `verify.rs` may try to verify normal call targets as contracts.

crates/cheatcodes/src/inspector.rs
```diff
@@ -2057,19 +2057,24 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>> for Cheatcode
 
 impl<FEN: FoundryEvmNetwork> InspectorExt for Cheatcodes<FEN> {
     fn should_use_create2_factory(&mut self, depth: usize, inputs: &CreateInputs) -> bool {
```
mablr, line 2243: It seems the value guard in stack.rs fires unconditionally for all `should_use_create2_factory` cases before distinguishing CREATE vs CREATE2.

Not 100% sure, please could you double check?

```solidity
  import "utils/Test.sol";

  contract Funded { constructor() payable {} }

  /// forge-config: default.always_use_create_2_factory = true
  contract PayableCreate2Test is Test {
      function test_payable_create2_via_factory() public {
          vm.deal(address(this), 1 ether);
          // Fails with: "--batch cannot route a deployment with non-zero value through the CREATE2 factory"
          Funded f = new Funded{salt: bytes32(uint256(1)), value: 0.1 ether}();
          assertEq(address(f).balance, 0.1 ether);
      }
  }
```

stevencartavia: fixed in [7b93782](https://github.com/foundry-rs/foundry/pull/14516/commits/7b9378299a7c790aa5b9efc73cc50601413a56c7)

crates/script/src/broadcast.rs
```diff
@@ -817,32 +821,20 @@ impl BundledState<TempoEvmNetwork> {
             }
         };
 
-        // Collect all transactions into Call structs
-        // Tempo batch transactions support CREATE only as the first call
+        let create2_deployer = self.script_config.evm_opts.create2_deployer;
         let mut calls: Vec<Call> = Vec::new();
-        let mut has_create = false;
         for (idx, tx) in sequence.transactions().enumerate() {
```
mablr: Mixed signed + unsigned txs are silently mis-batched: senders is built from `tx.is_unsigned()` only, but the calls loop iterates all `sequence.transactions()`. A pre-signed tx slips through as a plain call sent from the unsigned sender. Maybe let's add an early `bail!` if any tx in `--batch` is `Signed`?

stevencartavia: [14aaa71](https://github.com/foundry-rs/foundry/pull/14516/commits/14aaa71c7c708e2ae33ca9893f6bcaa3c2e70253)

crates/script/src/lib.rs
mablr: This seems no-op now ?

crates/script/src/broadcast.rs
mablr: `add_pending()` seems to short-circuit when pending already contains the hash, so only `transactions[0].hash` gets set and the rest stay `None`? Also pending is never cleared (no remove_pending), so the saved JSON would ship with a stale pending entry? Not sure about this logic, please could you double check?

stevencartavia: [0249db7](https://github.com/foundry-rs/foundry/pull/14516/commits/0249db7dbf5188a056af5b8c214822d82fc2a750)

crates/script/src/lib.rs
```diff
@@ -135,13 +135,6 @@ pub struct ScriptArgs {
     #[arg(long)]
     pub batch: bool,
 
-    /// Number of calls per Tempo batch transaction.
-    ///
-    /// When `--batch` is enabled, splits the collected calls into multiple batch
-    /// transactions of at most this many calls each.
-    #[arg(long, requires = "batch", default_value = "100")]
-    pub batch_size: usize,
-
```
mablr, line 145: We should carefully recheck the consequences of removing `batch_size`.  I think this previous change re-targeted it silently 33c827e2a0264da9792195564fabb7355f09532d, I can't remember why... 🤔

mablr (commented):
Overall sgtm, just raising a point to pay attention.

mablr (commented):
Addressed some concerns:
- checkpoint batch tx hash immediately after send, before receipt poll, so a crash/timeout cannot lose the in-flight hash on resume; detect stamped hashes at startup and skip re-send when resuming
- fix alignment check to compare remaining calls (not total) against transactions, and skip already-receipted entries in `per_tx_addresses`
- bail early when any batch tx carries EIP-7702 authorization lists orblob sidecars, which the batch envelope cannot carry
- remove dead has_create branches (`CREATE`s are rewritten before broadcast); replace with a defensive bug-bail
- fix misleading comment in `stack.rs` nonce is from the EVM frame caller (script contract), not the broadcast EOA
- document that cloned receipts carry whole-batch `gasUsed`

mattsse (changes requested):
Requesting changes because the batch resume path can make `--resume` unrecoverable after an interrupted send.

P1: The batch checkpoint only stamps the hash onto remaining txs and saves it ([`broadcast.rs#L1202-L1207`](https://github.com/foundry-rs/foundry/blob/92693d195f40047fff2019ab51b66259158ffbfe/crates/script/src/broadcast.rs#L1202-L1207)), then waits via a plain receipt poll ([`broadcast.rs#L1217-L1228`](https://github.com/foundry-rs/foundry/blob/92693d195f40047fff2019ab51b66259158ffbfe/crates/script/src/broadcast.rs#L1217-L1228)). Unlike the normal path, it never adds the hash to `sequence.pending` ([`broadcast.rs#L854-L857`](https://github.com/foundry-rs/foundry/blob/92693d195f40047fff2019ab51b66259158ffbfe/crates/script/src/broadcast.rs#L854-L857)). On resume, that stamped hash is treated as `pending_batch_hash` ([`broadcast.rs#L967-L970`](https://github.com/foundry-rs/foundry/blob/92693d195f40047fff2019ab51b66259158ffbfe/crates/script/src/broadcast.rs#L967-L970), [`broadcast.rs#L1163-L1169`](https://github.com/foundry-rs/foundry/blob/92693d195f40047fff2019ab51b66259158ffbfe/crates/script/src/broadcast.rs#L1163-L1169)), so if the tx was dropped, resume keeps waiting on a dead hash instead of rebroadcasting. Please checkpoint the batch through the pending machinery and ensure dropped/timeout recovery clears or ignores the stamped batch hashes so a later `--resume` can send a replacement.

P2: Resume recovery still does send-only work before checking the already-submitted hash: signer/session lookup runs first ([`broadcast.rs#L1018-L1042`](https://github.com/foundry-rs/foundry/blob/92693d195f40047fff2019ab51b66259158ffbfe/crates/script/src/broadcast.rs#L1018-L1042)), then access-key prep/gas estimation/sponsor attachment run before the `pending_batch_hash` branch ([`broadcast.rs#L1142-L1164`](https://github.com/foundry-rs/foundry/blob/92693d195f40047fff2019ab51b66259158ffbfe/crates/script/src/broadcast.rs#L1142-L1164)). Recovering an already-submitted tx should not require the original signer/sponsor or a fresh estimate. Please split the recovery path so an existing pending batch hash goes straight to receipt/pending recovery before signer resolution, authorization prep, estimation, or sponsor attachment.

figtracer (approved):

mablr (approved):

mattsse (approved):

PR #14522, opened 2026-04-30
feat(forge): add fuzz run selection

### LINKED ISSUE #13140: feat(forge): add argument to set iteration of fuzz run
### Component

Forge

In scenarios where fuzz runs are relying on `vm.random*` function instead of input arguments, it would be useful to:
1. Log the rng seed for the failed testcase -- this is useful in case seed is not set
2. Be able to pass an additional `forge` cli argument to set the fuzz run
3. Maybe save seed and fuzz run iteration in cache to be abel to quiclly reproduce it with `forge t --rerun`

This is useful in situations where fuzzing campaigns take longer, and counterexamples are only found after long periods of time. Otherwise, debugging such tests is quite difficult. For example, test below fails after a few thousands runs if seed is not set, and fails after exactly 5932 runs if seed is set to `0x1111`. POC provided runs fast, but in case testcase is complex, waiting times can be > 3-5 minutes, making debugging difficult.

```
pragma solidity ^0.8.28;

import {Test} from 'forge-std/Test.sol';

contract RunsTest is Test {
  function test_fuzz(uint256) public view {
    uint256 ratioBps = bound(vm.randomUint(), 100_00, 200_00);
    uint256 amountA = bound(vm.randomUint(), 0, 1e30);
    uint256 amountB = amountA * ratioBps / 100_00;
    uint256 effectiveRatioBps = effectiveRatioBps(amountA, amountB);
    uint256 effectiveRatioWad = effectiveRatioWad(amountA, amountB);
    assertEq(effectiveRatioBps < 111_11, effectiveRatioWad < 1.111099999999999999e18);
  }

  function effectiveRatioBps(uint256 amountA, uint256 amountB) internal pure returns (uint256) {
    return amountB * 100_00 / amountA;
  }

  function effectiveRatioWad(uint256 amountA, uint256 amountB) internal pure returns (uint256) {
    return amountB * 1e18 / amountA;
  }
}
```

### Additional context

At the moment I am overcoming this by 
- setting a rng seed
- waiting for test to revert
- read the number of runs logged by forge test
- populate (runs-1) [here](https://github.com/foundry-rs/foundry/blob/7cf2082be4b4cf98b9f872efb0334380567ef8dd/crates/evm/evm/src/executors/fuzz/mod.rs#L110)
-  use locally-built forge for debug purposes

figtracer:
## Summary
- add `--fuzz-run` / `FOUNDRY_FUZZ_RUN` / `[fuzz] run` for executing one 1-based fuzz iteration
- fast-forward generated fuzz inputs in selected-run mode and run it with one worker
- persist `fuzz_seed` and `fuzz_run` metadata on fuzz counterexamples so `--rerun` can replay `vm.random*` failures

Closes #13140
cc: @grandizzy


crates/forge/tests/cli/test_cmd/fuzz.rs
```diff
@@ -868,3 +869,110 @@ Ran 1 test suite [ELAPSED]: 0 tests passed, 1 failed, 0 skipped (1 total tests)
...
+        uint256 rand = vm.randomUint(0, 4);
+        assertTrue(rand != 0, "hit value 0");
+    }
+}
+   "#,
+    );
+
+    let assert = cmd
+        .args(["test", "--fuzz-seed", "1", "--mt", "testFuzz_randomUint_shouldFail", "-j1"])
+        .assert_failure();
```
grandizzy: let's use snapbox snapshot here instead , same below

crates/evm/fuzz/src/lib.rs
```diff
@@ -102,6 +102,12 @@ pub struct BaseCounterExample {
     /// Whether to display sequence as solidity.
     #[serde(skip)]
     pub show_solidity: bool,
+    /// Fuzz seed used to generate this counterexample.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub fuzz_seed: Option<U256>,
+    /// 1-based fuzz run that generated this counterexample.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub fuzz_run: Option<u32>,
```
grandizzy: is this good enough or should we persist fuzz run and the worker id for make it deterministic / create WorkerState with both? since runs are parallelized so run 1 as (worker, run) pair could be (1, 1), (2, 1)

grandizzy (commented):
ty! left some comments, pls check

crates/evm/fuzz/src/lib.rs
figtracer: yea we should def persist the worker id as well, otherwise we can't reproduce deterministically, nice

crates/evm/fuzz/src/lib.rs
```diff
@@ -102,6 +102,15 @@ pub struct BaseCounterExample {
     /// Whether to display sequence as solidity.
     #[serde(skip)]
     pub show_solidity: bool,
+    /// Fuzz seed used to generate this counterexample.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
grandizzy: can we reuse here the newly added `FuzzRunMetadata` ?

crates/forge/tests/cli/test_cmd/fuzz.rs
```diff
@@ -868,3 +871,141 @@ Ran 1 test suite [ELAPSED]: 0 tests passed, 1 failed, 0 skipped (1 total tests)
...
+Suite result: FAILED. 0 passed; 1 failed; 0 skipped; [ELAPSED]
+...
+Tip: Run `forge test --rerun` to retry only the 1 failed test
+
+[SEED] (use `--fuzz-seed` to reproduce)
+
+"#]];
+
+    let assert =
+        cmd.args(["test", "--mt", "testFuzz_randomUint_shouldFail", "-j1"]).assert_failure();
```
grandizzy: nit: let's use `.assert_failure().stdout_eq`

grandizzy (commented):
makes sense, only couple of nits, pls check. thanks!

grandizzy (approved):
lgtm!

mablr (approved):

PR #14528, opened 2026-04-30
feat(forge): use network = "tempo" and add rpc_endpoints in tempo template

zerosnacks:
Updates the `forge init --network tempo` template so the generated `foundry.toml` uses `network = "tempo"` instead of the legacy `tempo = true`, and ships the Tempo RPC aliases out of the box.

```toml
[profile.default]
src = "src"
out = "out"
libs = ["lib"]
network = "tempo"

[rpc_endpoints]
tempo = "https://rpc.tempo.xyz/"
moderato = "https://rpc.moderato.tempo.xyz/"
```

- `network = "tempo"` is the canonical, non-deprecated form (legacy `tempo = true` is still parsed for back-compat).
- The `tempo` and `moderato` aliases are usable immediately via `--rpc-url` / `--fork-url` after `forge init`.

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -2870,10 +2870,7 @@ impl BasicConfig {
         {
             table.insert("network".to_string(), toml::Value::String(network.clone()));
             if network == "tempo" {
-                table.insert(
-                    "eth_rpc_url".to_string(),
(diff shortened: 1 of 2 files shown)


mablr (approved):
lgtm

PR #14530, opened 2026-04-30
feat(forge): support per-test network selection via inline config

### LINKED ISSUE #14526: feat(forge): support per-test network selection via inline config
## Summary

Allow switching the EVM network family (e.g. `Ethereum` ↔ `Tempo` ↔ `Optimism`) on a per-contract or per-test basis using inline config, e.g.:

```solidity
contract MixedSuite is Test {
    /// forge-config: default.network = "tempo"
    function test_tempo_only_behavior() public { ... }

    function test_ethereum_baseline() public { ... }
}
```

This would enable a single `forge test` invocation to cover a mixed test suite that exercises both vanilla EVM and Tempo EVM (or OP) semantics.

## Current behavior

The network type is selected once, up-front, in `crates/forge/src/cmd/test/mod.rs`:

```rust
let (libraries, mut outcome) = if evm_opts.networks.is_tempo() {
    self.build_and_run_tests::<TempoEvmNetwork>(...)
} else if evm_opts.networks.is_optimism() {
    self.build_and_run_tests::<OpEvmNetwork>(...)
} else {
    self.build_and_run_tests::<EthEvmNetwork>(...)
};
```

The entire `MultiContractRunner`, `Executor`, `EvmEnvFor<FEN>`, `SpecFor<FEN>`, etc. are generic over `FoundryEvmNetwork` and monomorphized for a single concrete network for the whole run.

`InlineConfig` *can* override `networks` in `TestRunnerConfig::reconfigure_with`, and that propagates to `inspector.networks(...)`, but it only changes inspector-side behavior (e.g. precompile injection within the same family). It cannot swap `TempoEvmNetwork` ↔ `EthEvmNetwork` mid-run because those are different generic instantiations.

The only inline switch that exists today is hardfork *within the same network family* — see `config_inline_hardfork_same_network_family` in `crates/forge/tests/cli/inline_config.rs`.

## Workaround today

Run `forge test` twice with different `--network` flags and filter tests accordingly.

## Proposal

Two possible approaches:

1. **Multi-pass execution under the hood.** When inline configs request a different network family, group tests by network and run them in separate passes per family, then merge outcomes for reporting.
2. **Dynamic dispatch over `FoundryEvmNetwork`.** Refactor the runner so the network type is no longer a compile-time generic but boxed/dispatched at runtime, allowing the executor to be rebuilt per contract/function based on the merged inline config.

Approach (1) is likely the smaller change and keeps the existing monomorphized hot path; approach (2) is more invasive but more flexible.

## Use case

Mixed test suites where some contracts assume Tempo-specific precompiles/semantics and others target stock Ethereum. Currently developers must split these into separate projects or invoke `forge test` twice.

mablr:
## Motivation

Close #14526

## Solution

Introduce `TestArgs::dispatch_network` method, handle multiple networks in `InlineConfig`.

Not the most elegant, it could be improved later.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/forge/tests/cli/inline_config.rs
```diff
@@ -425,3 +425,72 @@ Ran 1 test suite [ELAPSED]: 1 tests passed, 0 failed, 0 skipped (1 total tests)
...
+            // The Tempo FeeManager precompile must NOT exist here.
+            function test_fee_manager_absent_on_ethereum() public view {
+                require(
+                    TIP_FEE_MANAGER.code.length == 0,
+                    "TipFeeManager should not exist on Ethereum"
+                );
+            }
+        }
+
+        contract TempoNetwork {
```
zerosnacks, line 457: For good measure could we also test a mixed case of a one contract with two functions, one without and one with a declaration?

zerosnacks (commented):
Nice! Overall approach lgtm

crates/forge/tests/cli/inline_config.rs
mablr: I added a mixed test case, as run per-network batch the resulting output gets splited per-network, i believe it's fine.

zerosnacks (approved):

figtracer (approved):
lg!

PR #14531, opened 2026-04-30
feat(cast): add Tempo keychain policy commands

figtracer:
## Summary

- add `cast keychain inspect` for TIP-1011 key policy state, limits, expiry, and allowed calls
- add `cast keychain policy add-call`, `set-limit`, and `remove-target` helpers around AccountKeychain calldata
- support root-account browser signing for keychain mutations and fail early when a command resolves to an access key signer
- add local/anvil T3 detection fallback plus parser and policy merge tests

### DIFF
--- a/crates/cast/src/cmd/keychain.rs
+++ b/crates/cast/src/cmd/keychain.rs
@@ -1,7 +1,9 @@
 use alloy_ens::NameOrAddress;
-use alloy_network::EthereumWallet;
+use std::time::Duration;
+
+use alloy_network::{EthereumWallet, TransactionBuilder};
 use alloy_primitives::{Address, U256, hex, keccak256};
-use alloy_provider::ProviderBuilder as AlloyProviderBuilder;
+use alloy_provider::{Provider, ProviderBuilder as AlloyProviderBuilder};
 use alloy_signer::Signer;
 use alloy_sol_types::SolCall;
 use chrono::DateTime;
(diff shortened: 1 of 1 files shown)


crates/cast/src/cmd/keychain.rs
```diff
@@ -753,6 +1198,343 @@ async fn send_keychain_tx(
...
+fn tempo_keys_path_display() -> String {
+    tempo_keys_path()
+        .map(|path| path.display().to_string())
+        .unwrap_or_else(|| "(unknown)".to_string())
+}
+
+fn add_selector_rule_to_scope(scope: &mut CallScope, rule: SelectorRule) -> bool {
+    if scope.selectorRules.is_empty() {
+        return false;
+    }
```
mablr, line 1333: This seem to silently ignore wildcard scopes, logically correct, but may worth an info/warn message ?

zerosnacks (commented):
Overall lgtm! Would like to see some additional tests

crates/cast/src/cmd/keychain.rs
```diff
@@ -695,21 +1033,90 @@ async fn run_remove_scope(
...
+
+    let calldata =
+        IAccountKeychain::setAllowedCallsCall { keyId: key_address, scopes: vec![target_scope] }
+            .abi_encode();
+    send_keychain_tx(calldata, tx_opts, &send_tx).await
+}
+
+/// `cast keychain policy set-limit` — update a spending limit amount.
+async fn run_policy_set_limit(
```
mablr, line 1093: Delegates directly to `run_update_limit` with only `key_address`, it never calls `resolve_key_metadata`, is this okay ?

crates/cast/src/cmd/keychain.rs
```diff
@@ -753,6 +1198,343 @@ async fn send_keychain_tx(
...
+                provider.raw_request::<_, AnvilNodeInfo>("anvil_nodeInfo".into(), ()).await;
+
+            if let Ok(info) = anvil_info
+                && info.network.as_deref() == Some("tempo")
+                && let Some(active_hardfork) = info.hard_fork
+            {
+                return Ok(active_hardfork.parse::<TempoHardfork>().is_ok_and(|h| h >= hardfork));
+            }
+
+            Err(err.into())
```
mablr: If `is_hardfork_active` fails for a reason other than "method not found" (e.g., a network timeout), the code still attempts anvil_nodeInfo. There's no check that the original error is a method-not-found/unsupported-method error. On a flaky connection this could mask the real error with a confusing second failure.

Overall logic here seems not adapted, and can be at least deduped.

crates/cast/src/cmd/keychain.rs
figtracer: the precompile resolves this from `msg.sender` should be okay

mablr (approved):
sgtm

zerosnacks (approved):

PR #14532, opened 2026-04-30
refactor(chisel): migrate to solar

mablr:
## Motivation

Migrate `chisel` from `solang` to `solar`.

All existing tests pass.

### DIFF
--- a/crates/chisel/src/executor.rs
+++ b/crates/chisel/src/executor.rs
@@ -16,7 +16,10 @@ use solar::{
     ast::{BinOpKind, ElementaryType, FunctionKind, LitKind, StateMutability, StrKind, UnOpKind},
     interface::Symbol,
     sema::{
-        hir::{self, Visibility},
+        hir::{
+            ContractId, Event, Expr, ExprKind, Function, ItemId, Res, StmtKind, Type as HirType,
+            TypeKind, Visibility,
+        },
         ty::{Gcx, Ty, TyKind},
     },
 };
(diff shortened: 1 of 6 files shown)


zerosnacks:
If time permits would love to get your eyes on this @DaniPopes & @grandizzy as you are most familiar with Solar & Chisel

crates/chisel/src/source.rs
```diff
@@ -251,52 +117,25 @@ impl<'gcx> GeneratedOutputRef<'_, '_, 'gcx> {
...
-                }
-            }
-            */
-            None
-        });
+        //
+        // Yul is not yet lowered to HIR (assembly statements appear as `StmtKind::Err`),
+        // so we walk the AST of the REPL source to find a top-level `return(...)` call
+        // inside any `assembly { ... }` block in `run()`.
+        let last_yul_return_span: Option<Span> = self.first_yul_return_span();
```
DaniPopes, line 125: we might not have test coverage for these assembly paths

crates/chisel/src/executor.rs
```diff
@@ -432,233 +449,285 @@ enum Type {
     Function(Box<Self>, Vec<Option<Self>>, Vec<Option<Self>>),
 
     /// (lhs, rhs)
-    Access(Box<Self>, String),
+    Access(Box<Self>, Symbol),
```
DaniPopes: we should use solar type system for this?

mablr: I'm unsure it's possible rn, Solar's typeck internally stores expression types in `FxHashMap<hir::ExprId, Ty<'gcx>>`, but this map is not exposed through `Gcx`.

crates/chisel/src/source.rs
mablr: added tests

crates/chisel/src/executor.rs
DaniPopes: right. at least you should be using solar types and expression instead of this mess of a Type+Expression. The core function of this human slop is to map an expression to its type so we can decode+pretty-print its runtime value. you can re-implement hir::Expr->solar Ty conversion locally (at least simple ones like function return, builtins) until it works upstream (type_of_expr())

mablr: Made the direct conversion from solar hir::Expr to DynSolType, is this what you meant?

DaniPopes (approved):

zerosnacks (approved):
Nice! 👍

PR #14534, opened 2026-04-30
fix(config): suppress spurious unknown-key warning for `network`

zerosnacks:
## Summary

After the recent Tempo template upgrade, projects with `network = "tempo"` in their `foundry.toml` started emitting a spurious warning, and `forge config` produced inconsistent / contradictory output for the network configuration.

```
Warning: Found unknown `network` config for profile `default` defined in foundry.toml.
```

## Root cause

The `network` field on `NetworkConfigs` had `#[serde(skip_serializing_if = "Option::is_none")]`, so it was omitted from the serialized default `Config`. The warnings provider builds its allowed-keys set from that default, so the valid `network` key wasn't recognized.

In addition, the legacy `tempo` / `optimism` boolean aliases were being serialized alongside `network`, leading to confusing output like `network = "tempo"` next to `tempo = false`.

## Fix

1. **Recognize `network` as a known key.** Replaced `skip_serializing_if` with `#[serde(default)]` on the `network` field so it always appears in the serialized default and the warnings provider treats it as known.
2. **Canonicalize on serialization.** Added a custom `Serialize` impl for `NetworkConfigs` that always emits the *resolved* network as the canonical `network = "..."` field and never emits the legacy `tempo` / `optimism` aliases. The legacy aliases remain accepted on deserialization for backward compatibility, and `tempo` / `optimism` are added to `BACKWARD_COMPATIBLE_KEYS` so they don't trigger unknown-key warnings on input.
3. **Regression tests.** Added unit tests covering both the new `network = "tempo"` form and the legacy `tempo = true` alias.

## Behavior

| Input (`foundry.toml`)           | Output (`forge config`)            |
|------------------------------------|--------------------------------------|
| `network = "tempo"`            | `network = "tempo"`              |
| `tempo = true` (legacy)          | `network = "tempo"` (canonicalized) |
| `optimism = true` (legacy)       | `network = "optimism"` (canonicalized) |
| (none set)                         | (omitted)                            |

- TOML output is unchanged for the no-network case (`toml-rs` omits `None`).
- JSON output now includes `"network": null` when unset; the `test_default_config` snapshot is updated.
- The `tempo = false` line previously emitted by default is gone (it duplicated information now carried by `network`).

## Verification

- `forge build` with `network = "tempo"` → no warning.
- `forge build` with `tempo = true` → no warning, network resolves to Tempo.
- `forge build` with no network setting → no warning.
- `cargo test -p foundry-evm-networks` → 11 passed.
- `cargo test -p foundry-config` → all passing, including new regression tests.
- `forge test_default_config` and `can_extract_config_values` snapshots pass.

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -6698,6 +6698,55 @@ mod tests {
     }
 
     #[test]
+    fn no_unknown_key_warning_for_network_field() {
+        // Regression test: `network` is a flattened `Option` field of `NetworkConfigs`. It must
+        // not trigger an unknown-key warning, regardless of whether it is set.
+        figment::Jail::expect_with(|jail| {
+            jail.create_file(
+                "foundry.toml",
+                r#"
+                [profile.default]
+                network = "tempo"
+                "#,
+            )?;
+
+            let cfg = Config::load().unwrap();
+            assert!(
+                !cfg.warnings.iter().any(
+                    |w| matches!(w, crate::Warning::UnknownKey { key, .. } if key == "network")
+                ),
+                "did not expect UnknownKey warning for `network`, got: {:?}",
+                cfg.warnings
+            );
+            Ok(())
+        });
+    }
+
+    #[test]
+    fn no_unknown_key_warning_for_legacy_tempo_alias() {
+        // Regression test: the legacy `tempo = true` alias must keep working without warnings.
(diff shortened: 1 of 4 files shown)


mablr (approved):
Nice👍, missed it while integrating `NetworkVariant` to `NetworkConfigs`

PR #14535, opened 2026-04-30
fix(anvil): check funds before transfer gas shortcut

nivanovvv:
## Motivation

`eth_estimateGas` has a fast path for simple native transfers. Because that shortcut ran before balance and value checks, an account without enough ETH could still receive the minimum transfer gas estimate.

## Solution

Run the sender affordability check before the simple-transfer shortcut, keep the gas allowance cap for priced transactions, and add a regression test for an unfunded transfer estimate.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mablr (changes requested):
You should not modify `.github/workflows/ci-tempo.yml`

nivanovvv:
> You should not modify `.github/workflows/ci-tempo.yml`

Thanks for the callout — I reverted the workflow change in commit fb596399a.                                                                                                                 This PR no longer modifies .github/workflows/ci-tempo.yml and now only includes the Anvil code/test changes.

mablr (commented):
@nivanovvv could you please have a look to broken tests ? I'm unsure this not a breaking change.

crates/anvil/src/eth/api.rs
```diff
@@ -1393,23 +1370,47 @@ impl EthApi<FoundryNetwork> {
         let is_tempo_tx = request.other.get("feeToken").is_some_and(|v| !v.is_null());
```
mattsse: This should probably also check `self.backend.is_tempo()`. Right now any non-null `feeToken` field makes us skip the ETH balance/value check, but on normal Anvil that field is otherwise ignored and the simple-transfer fast path below still runs because `!self.backend.is_tempo()`. So an unfunded normal ETH transfer with an extra `feeToken` field can still return `21000`. Can we gate this on Tempo mode too and add a regression test?

mattsse (commented):
I found two remaining edge cases where the simple-transfer shortcut can still return `21000` without an ETH funds check.

crates/anvil/src/eth/api.rs
```diff
@@ -1393,23 +1370,47 @@ impl EthApi<FoundryNetwork> {
...
         let gas_price = fees.gas_price.unwrap_or_default();
-        // If we have non-zero gas price, cap gas limit by sender balance.
-        // Skip this check for Tempo transactions which pay with fee tokens, not ETH.
-        if gas_price > 0
-            && !is_tempo_tx
-            && let Some(from) = request.from
-        {
+        // Check transfer value before any fast path, and cap gas limit by sender balance when the
+        // request has a non-zero gas price.
+        if !is_tempo_tx && let Some(from) = request.from {
```
mattsse: This still misses requests that omit `from`. `build_call_env` later defaults a missing sender to the zero address, but this pre-check skips entirely when `request.from` is `None`, so `eth_estimateGas({ to, value: 1 })` to an EOA can still take the fast path and return `21000`. Can we either normalize the sender before this check or avoid the shortcut when no sender was checked?

nivanovvv:
> I found two remaining edge cases where the simple-transfer shortcut can still return `21000` without an ETH funds check.

Addressed both points in 18f459650: feeToken no longer bypasses funds checks outside Tempo mode, and missing from now falls back to the effective caller; added regression tests for both cases.

mattsse (dismissed):
Reviewed the reordered estimate-gas flow again. The shortcut now runs after fee/value/funds checks, the missing from handling matches build_call_env, and feeToken only bypasses ETH checks in Tempo mode. Focused gas tests passed locally.

mablr (approved):
f94410b47 fixes no-`from` transfer gas estimation, avoids treating Cast’s zero-address sentinel as an explicit sender, and updates the related Anvil/Cast tests. b2c99cdf3 Adds an Anvil gas estimation regression test around missing sender behavior and gas allowance capping.

Sgtm now. Pending another review/approval.

grandizzy (approved):

PR #14536, opened 2026-04-30
anvil: unify Tempo nonce markers across send RPCs

isagi-y22:
## Motivation
                                                                                                                                                                                               
  Tempo transactions with `nonce_key != 0` currently behave inconsistently across RPC send entrypoints:                                                                                        
                                                                                                                                                                                               
  - `eth_sendRawTransaction` uses hash-based pool markers (2D nonce semantics).                                                                                                                
  - `eth_sendTransaction` and `eth_sendUnsignedTransaction` still use `(from, nonce)` sequential markers.
                                                                                                                                                                                               
  This makes parallel nonce behavior endpoint-dependent, which can cause correctness and compatibility issues for clients that switch between RPC methods.
                                                                                                                                                                                               
  ## Solution                                                                                                                                                                                  
                                                                                                                                                                                               
  Unify marker assignment logic for Tempo parallel nonce transactions across all three send paths in `crates/anvil/src/eth/api.rs`.                                                            

  - Added a shared helper: `tempo_parallel_nonce_markers(...)`.                                                                                                                                
  - Applied it in:
    - `eth_sendTransaction`                                                                                                                                                                    
    - `eth_sendRawTransaction`                                                                                                                                                                 
    - `eth_sendUnsignedTransaction`                                                                                                                                                            
  - Behavior:                                                                                                                                                                                  
    - If tx is Tempo and `nonce_key != 0`: `requires = []`, `provides = [tx_hash]`.                                                                                                            
    - Otherwise: keep existing `(from, nonce)` marker behavior unchanged.                                                                                                                      
                                                                                                                                                                                               
  This keeps legacy and non-parallel nonce behavior intact while removing endpoint-specific divergence for Tempo parallel nonce txs.
                                                                                                                                                                                               
  ## PR Checklist                                                                                                                                                                              
                                                                                                                                                                                               
  - [ ] Added Tests                                                                                                                                                                            
  - [ ] Added Documentation                                                                                                                                                                    
  - [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -2217,7 +2217,6 @@ impl EthApi<FoundryNetwork> {
         } else {
             (required_marker(nonce, on_chain_nonce, from), vec![to_marker(nonce, from)])
         };
-        debug_assert!(requires != provides);
 
         self.add_pending_transaction(pending_transaction, requires, provides)
     }
@@ -3558,6 +3557,7 @@ impl EthApi<FoundryNetwork> {
         requires: Vec<TxMarker>,
         provides: Vec<TxMarker>,
     ) -> Result<TxHash> {
+        debug_assert!(requires != provides);
         let from = *pending_transaction.sender();
         let priority = self.transaction_priority(&pending_transaction.transaction);
         let pool_transaction =
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -2210,8 +2210,13 @@ impl EthApi<FoundryNetwork> {
         // pre-validate
         self.backend.validate_pool_transaction(&pending_transaction).await?;
 
-        let requires = required_marker(nonce, on_chain_nonce, from);
-        let provides = vec![to_marker(nonce, from)];
+        let (requires, provides) = if let Some((requires, provides)) =
+            tempo_parallel_nonce_markers(&pending_transaction)
+        {
+            (requires, provides)
+        } else {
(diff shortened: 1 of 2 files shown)


mattsse (commented):
wdyt @stevencartavia

crates/anvil/src/eth/api.rs
```diff
@@ -2210,8 +2210,13 @@ impl EthApi<FoundryNetwork> {
...
-        let requires = required_marker(nonce, on_chain_nonce, from);
-        let provides = vec![to_marker(nonce, from)];
+        let (requires, provides) = if let Some((requires, provides)) =
+            tempo_parallel_nonce_markers(&pending_transaction)
+        {
+            (requires, provides)
+        } else {
+            (required_marker(nonce, on_chain_nonce, from), vec![to_marker(nonce, from)])
+        };
         debug_assert!(requires != provides);
```
mablr: Do we need `debug_assert!(requires != provides)` in all 3 places ?

mablr (commented):
Makes sense.

crates/anvil/src/eth/api.rs
stevencartavia: yes, I think that `debug_assert!(requires != provides)` should be added to `add_pending_transaction()`

stevencartavia (commented):
makes sense

stevencartavia (dismissed):

mablr (dismissed):

mattsse (dismissed):

PR #14543, opened 2026-05-01
feat(lint): project-wide passes + pragma-inconsistent

stevencartavia:
## Summary

Adds a `ProjectLintPass` framework for **cross-file** lints and ships the first one, `pragma-inconsistent`, which flags projects whose input files declare different Solidity pragma version requirements.

## Why

Per-file lints (`EarlyLintPass`/`LateLintPass`) cannot reason across translation units, so they can't detect things like "these two files pin to incompatible pragmas." The new `ProjectLintPass` runs **once after all per-file passes have completed**, with access to every input source's AST, file handle, and inline-config. That gives us a clean place to land any cross-file analysis going forward.

`pragma-inconsistent` is the motivating example. The detection iterates the explicit `input` slice, so library/submodule pragmas don't pollute the report, and emits per-occurrence diagnostics that fit Foundry's compiler-style emitter.

## Framework changes

- New trait [`ProjectLintPass`](crates/lint/src/linter/project.rs) with a single `check_project(&mut self, ctx, sources: &[ProjectSource])` method.
- New helper [`ProjectLintEmitter`](crates/lint/src/linter/project.rs) for emitting diagnostics scoped to a particular source while honouring per-source inline config (`// forge-lint: disable-*`).
- New [`ProjectSource`](crates/lint/src/linter/project.rs) struct: pre-loaded source unit (path, file handle, AST, inline config) so passes don't re-parse comments per emit.
- [`register_lints!`](crates/lint/src/sol/macros.rs) extended with a `project` pass type alongside `early`, `late`, and `both`. Generates `as_project_lint_pass` / `create_project_lint_passes` symmetrically.
- [`SolidityLinter::process_project`](crates/lint/src/sol/mod.rs) gathers enabled project passes from every severity bucket, builds the `ProjectSource` list once, and runs each pass after the per-file phase.

## The `pragma-inconsistent` lint

- Code: [`crates/lint/src/sol/info/pragma_directive.rs`](crates/lint/src/sol/info/pragma_directive.rs).
- Severity: `Info`.
- Detection: groups every top-level `pragma solidity` directive across all input sources by the rendered version-requirement string. Emits when ≥2 distinct strings exist.
- Message: per-occurrence, listing the other distinct requirements present in the project.

## Tests

**Single-file scenarios** (testdata files in `crates/lint/testdata/`, using the existing `//~NOTE:` annotation style):

| File | What it tests |
|---|---|
| `PragmaInconsistentCaretAboveExact.sol` | caret range whose lower bound is strictly below the exact version (`^0.8.0` + `0.8.18`) |
| `PragmaInconsistentCaretMatchesExact.sol` | caret range whose lower bound equals the exact version (`^0.8.20` + `0.8.20`) — looks-the-same-but-still-distinct, guards against the lint silently doing semver intersection |
| `PragmaInconsistentCaretVsTilde.sol` | `^0.8.20` vs `~0.8.20` |
| `PragmaInconsistentRangeVsExact.sol` | `>=0.8.0 <0.9.0` vs `0.8.20` (range satisfies exact, but lint is intentionally string-based) |
| `PragmaInconsistentOrVsExact.sol` | `0.8.20 \\|\\| 0.8.21` vs `0.8.20` |
| `PragmaInconsistentThreeDistinct.sol` | `>=0.8.0`, `^0.8.0`, `~0.8.0` — verifies the `others` list contains every other variant |

**Multi-file scenarios** (`forgetest!` cases in `crates/forge/tests/cli/lint.rs`, since they need genuine cross-file projects):

- Negative (must NOT warn):
  - all files use the same exact pragma (`0.8.20`)
  - all files use the same caret pragma (`^0.8.20`)
  - single file in the project
- Positive (must warn):
  - cross-file caret vs exact (the original regression case)
  - duplicates among a conflict — two identical files plus one different pragma still emits three warnings
- Mixed:
  - file without an explicit pragma uses the test-utils default (`add_raw_source` bypasses the auto-injected pragma)

Source bodies in the `forgetest!` cases use module-level `const` raw strings to keep rustfmt readable.

## Verification

- `cargo test -p forge --test cli pragma_inconsistent` — 6/6 pass
- `cargo test -p forge --test ui pragma` — all testdata snapshots pass


crates/lint/testdata/PragmaInconsistent.sol
```diff
@@ -0,0 +1,7 @@
+//@compile-flags: --only-lint pragma-inconsistent
+
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0; //~NOTE: this file uses 'pragma solidity ^0.8.0;', but other files use 0.8.18
+pragma solidity 0.8.18; //~NOTE: this file uses 'pragma solidity 0.8.18;', but other files use ^0.8.0
```
mablr: Message is a bit misleading, but i don't think it worths handling specifically this unlikely case.

crates/lint/src/sol/info/pragma_directive.rs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        // Stable order for snapshots and JSON output.
+        entries.sort_by(|a, b| {
+            sources[a.0].path.cmp(&sources[b.0].path).then(a.1.lo().cmp(&b.1.lo()))
+        });
+
+        let mut distinct: Vec<&str> = entries.iter().map(|(_, _, s)| s.as_str()).collect();
+        distinct.sort_unstable();
+        distinct.dedup();
```
mablr: You can skip the `HashSet` entirely and just do the sort+dedup first, then check `distinct.len() < 2`.

mablr (dismissed):
sgtm

zerosnacks (approved):
Lgtm 👍 , made a few improvements to exhaustively tests the various ways pragma's can be defined

mablr (approved):

PR #14547, opened 2026-05-01
fix(cheatcodes): transfer value for payable mock calls

### LINKED ISSUE #10812: bug(`cheatcodes`): unexpected behavior when using `mockCall` for payable functions
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.2.2-nightly Commit SHA: 62ce5bc62f5951ea0575c8df70b4f855beef3af1 Build Timestamp: 2025-06-01T06:01:59.278364000Z (1748757719) Build Profile: maxperf

Installed via nix

forge test

### Operating System

macOS (Apple Silicon)

### Describe the bug

If the mock call simulates a payable function, then when the address is called, the call succeeds but the balance is not transferred.

```solidity
pragma solidity ^0.8.28;

import { Test } from "forge-std/Test.sol";

contract MockCallBugTest is Test {
    function setUp() public { }

    function testMockCallBug() public {
        vm.deal(address(this), 1 ether);

        address target = makeAddr("target");
        uint256 value = 0.1 ether;
        bytes memory data = "someData";
        bytes memory resp = "mockedResponse";

        assertEq(target.balance, 0, "Initial balance should be zero");

        // Mock a call to the target address
        vm.mockCall(target, value, data, resp);

        // Make the call
        (bool success, bytes memory returnData) = target.call{value: value}(data);
        assertTrue(success, "Call should succeed");
        assertEq(returnData, resp, "Return data should match");

        // Check the target's balance after the call
        assertEq(target.balance, 0, "Target balance should NOT be zero after mock call");
        // BUG: This is a bug since the value is not transferred to the target but the call is not reverted.
        // should pass: assertEq(target.balance, value, "Target balance should NOT be zero after mock call");
    }

    function testRealCall() public {
        vm.deal(address(this), 1 ether);

        address target = address(new RealCallContract());
        uint256 value = 0.1 ether;
        bytes memory data = abi.encodeCall(RealCallContract.payableFunction, ());
        bytes memory resp = abi.encode("mockedResponse");

        assertEq(target.balance, 0, "Initial balance should be zero");

        // Make the real call
        (bool success, bytes memory returnData) = target.call{value: value}(data);
        assertTrue(success, "Call should succeed");
        assertEq(returnData, resp, "Return data should match");

        // Check the target's balance after the call
        assertEq(target.balance, value, "Target balance should be equal to the value sent");
    }
}

contract RealCallContract {
    function payableFunction() external payable returns (bytes memory) {
        return "mockedResponse";
    }
}
```

I don't know why this occurs. I have verified this behavior with slight variations:
- Making sure `target` has some code length. (`vm.etch(target, "mock")`)
- Casting `target` into a payable address

srdtrk:
## Motivation
Fixes #10812.

`vm.mockCall` can mock payable calls with a specific `msg.value`, but matched calls currently return the mocked data without applying the ETH value transfer side effect.
This means a mocked payable call can succeed while leaving the callee balance unchanged, unlike a real EVM call.

## Solution

Apply the call value transfer when a mocked call is matched, using revm's journaled transfer path.
Successful mocked calls commit the transfer, while mocked reverts roll it back so `mockCallRevert` continues to behave like a reverting callee call.
Regression tests were added for:
- `mockCall` transferring ETH to the mocked callee
- `mockCallRevert` rolling back the transfer
- `mockCalls` accumulating value transfers across repeated mocked calls

## PR Checklist
- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/cheatcodes/src/inspector.rs
```diff
@@ -879,6 +879,36 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
                 // Else, we pop the front element
                 return_data_queue.pop_front()
             } {
+                if let Some(value) = call.transfer_value() {
+                    let checkpoint = ecx.journal_mut().checkpoint();
+                    match ecx.journal_mut().transfer_loaded(
```
srdtrk, line 919: I'm using `transfer_loaded` instead of `transfer` so that I don't need to handle database access failure cases. I'm open to changing this if there is a recommended way to handle those.

crates/cheatcodes/src/inspector.rs
```diff
@@ -879,6 +879,36 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
...
+                        call.transfer_from(),
+                        call.transfer_to(),
+                        value,
+                    ) {
+                        None => {
+                            if return_data.ret_type.is_ok() {
```
srdtrk, line 925: I didn't include `return_data.ret_type.is_ok()` in the broader if let statement above (like `Some(value) = call.transfer_value() && return_data.ret_type.is_ok()` because I didn't want to skip handling the `Some(err)` case below if return type is ok. This is because transfer failures should take precedence before return type reverts.

crates/cheatcodes/src/inspector.rs
```diff
@@ -879,6 +879,36 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
...
+                                ecx.journal_mut().checkpoint_commit();
+                            } else {
+                                ecx.journal_mut().checkpoint_revert(checkpoint);
+                            }
+                        }
+                        Some(err) => {
```
srdtrk, line 931: Note that when using `mockCalls`, if the queue has more than one item, then the mocked response will be popped (line 880) even if we hit this error case. I think this is the intended behavior, but I could also see a case for handling this without popping.

crates/cheatcodes/src/inspector.rs
```diff
@@ -879,6 +879,36 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
...
+                            ecx.journal_mut().checkpoint_revert(checkpoint);
+                            return Some(CallOutcome {
+                                result: InterpreterResult {
+                                    result: err.into(),
+                                    output: Bytes::new(),
+                                    gas,
+                                },
+                                memory_offset: call.return_memory_offset.clone(),
+                                was_precompile_called: false,
```
mablr, line 940: Is `was_precompile_called` always false ?

crates/cheatcodes/src/inspector.rs
mablr: The safest would be probably to peek value, try transfer, and finally pop on success && len>1.

crates/cheatcodes/src/inspector.rs
srdtrk: It is not always false elsewhere in codebase. The successful mocked-call path still uses true. The transfer-failure branch sets it to false, matching what I've seen elsewhere in the codebase.

crates/cheatcodes/src/inspector.rs
srdtrk: Ok, I've implemented this.

mablr (dismissed):
Lgtm, @grandizzy could you please have a look?

crates/cheatcodes/src/inspector.rs
```diff
@@ -872,13 +872,43 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
                             && mock.value.is_none_or(|value| Some(value) == call.transfer_value())
                     })
                     .map(|(_, v)| v),
-            } && let Some(return_data) = if return_data_queue.len() == 1 {
+            } && let Some(return_data) = return_data_queue.front().map(|x| x.to_owned())
+            {
+                if let Some(value) = call.transfer_value() {
+                    let checkpoint = ecx.journal_mut().checkpoint();
+                    match ecx.journal_mut().transfer_loaded(
+                        call.transfer_from(),
```
stevencartavia, line 920: I think under vm.prank, the new transfer debits the wrong account, `call.transfer_from()` is read at L880 before the prank rewrite at L944 `call.caller = prank.new_caller;`, so `address(this)` is debited instead of the pranked sender.  Maybe adding a regression test is helpful? wdyt @mablr?

srdtrk: Nice catch. I verified this with a test and moved the mocked-call handling after prank rewrite. Let me know if this looks good.

stevencartavia (approved):

crates/cheatcodes/src/inspector.rs
mablr: good catch 👍

mablr (approved):

crates/cheatcodes/src/inspector.rs
```diff
@@ -932,6 +896,72 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
...
+                        call.input.bytes(ecx).get(..mock.calldata.len()) == Some(&mock.calldata[..])
+                            && mock.value.is_none_or(|value| Some(value) == call.transfer_value())
+                    })
+                    .map(|(_, v)| v),
+            } && let Some(return_data) = return_data_queue.front().map(|x| x.to_owned())
+            {
+                if let Some(value) = call.transfer_value() {
+                    let checkpoint = ecx.journal_mut().checkpoint();
+                    match ecx.journal_mut().transfer_loaded(
+                        call.transfer_from(),
```
figtracer, line 920: we still have the same problem mentioned before but for `vm.broadcast()`

srdtrk: I'm not sure I understand. Isn't vm.broadcast for calls that'd actually go on chain. Why would you ever use vm.broadcast with vm.mockCall?

figtracer: oh you're right i glanced over it, lgtm

figtracer (approved):

PR #14555, opened 2026-05-04
fix(forge): ignore ETH_RPC_URL for test forking

### LINKED ISSUE #14538: regression: setting env variable ETH_RPC_URL now runs all tests to be run in a fork
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.6.0-v1.7.0

foundryup: 1.8.0

forge test

### Operating System

macOS (Apple Silicon)

### Describe the bug

Hey,
Our CI at Maple Finance has started failing after updating from 1.5.1 to 1.7.0.
 
We have discovered the problem is that if the environmental variable ETH_RPC_URL is set then all tests executed with `forge test` are run in a fork against the URL defined in ETH_RPC_URL. 

This was fixed by rolling back to 1.5.1.

The expected outcome here is that tests are only run on a fork when this is done via the command line, in the foundry.toml or via the solidity cheatcode vm.createSelectFork or similar. 

I believe this commit caused it: https://github.com/foundry-rs/foundry/commit/22d708056dc88a6753f08b72a84029668f461024#diff-c88b64345fdc1fa9687c10666c5f4ebf4df5885371dbe0c771dc5b6fee314391R24 ( line 23-24 of crates/cli/src/opts/rpc_common.rs)

figtracer:
## Summary

- keep `ETH_RPC_URL` support for cast RPC commands
- keep the shared `--rpc-url`/`--fork-url` option on `RpcCommonOpts`, but remove the shared `ETH_RPC_URL` Clap binding so Forge only forks when CLI/config provides an RPC URL
- make Cast resolve `ETH_RPC_URL` as a command-specific fallback
- add tests for the command-specific Clap env bindings

Closes #14538

### DIFF
--- a/crates/cli/src/opts/evm.rs
+++ b/crates/cli/src/opts/evm.rs
@@ -308,6 +308,17 @@ mod tests {
     }
 
     #[test]
+    fn rpc_url_arg_does_not_read_eth_rpc_url_env() {
+        use clap::CommandFactory;
+
+        let command = EvmArgs::command();
+        let rpc_url =
+            command.get_arguments().find(|arg| arg.get_id() == "rpc_url").expect("rpc_url arg");
+
+        assert!(rpc_url.get_env().is_none());
+    }
+
+    #[test]
     fn can_parse_chain_id() {
         let args = EvmArgs {
             env: EnvArgs { chain: Some(NamedChain::Mainnet.into()), ..Default::default() },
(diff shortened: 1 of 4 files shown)


mablr (approved):
lgtm 👍

CI errors are not related to these changes

zerosnacks (approved):

PR #14558, opened 2026-05-04
refactor(script): reuse shared Tempo CLI opts

figtracer:
## Summary
- extract shared Tempo CLI args into `TempoCommonOpts` containing only `--tempo.fee-token` and `--tempo.expires`
- flatten `TempoCommonOpts` into both full `TempoOpts` and `forge script` args

### DIFF
--- a/crates/cli/src/opts/tempo.rs
+++ b/crates/cli/src/opts/tempo.rs
@@ -11,10 +11,10 @@ use std::{
 
 use crate::utils::parse_fee_token_address;
 
-/// CLI options for Tempo transactions.
+/// CLI options common to Tempo transactions across commands.
 #[derive(Clone, Debug, Default, Parser)]
 #[command(next_help_heading = "Tempo")]
-pub struct TempoOpts {
+pub struct TempoCommonOpts {
     /// Fee token address for Tempo transactions.
     ///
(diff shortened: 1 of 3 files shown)


mablr (commented):
I'm not sure it's relevant for now since most of the Tempo flags are not used by `forge script`. This would expose many dead flags.

figtracer:
> I'm not sure it's relevant for now since most of the Tempo flags are not used by `forge script`. This would expose many dead flags.

i think this is cleaner though, exposing these dead flags won't be a problem for the user with `ensure_supported_tempo_opts`

IMO, hardcoding the flags we need doesn't look right @mablr

mablr (commented):
I agree current code duplication is not nice, but I do think exposing all flags on `forge script -h` and returning error on unsupported ones is a misleading design.

What about extracting common args `fee_token`/`expires` in their own struct, and then flatten it in parent?

figtracer:
> I agree current code duplication is not nice, but I do think exposing all flags on `forge script -h` and returning error on unsupported ones is a misleading design.
> 
> What about extracting common args `fee_token`/`expires` in their own struct, and then flatten it in parent?

let's try this 👍

mablr (approved):

zerosnacks (approved):
Nice 👍

PR #14559, opened 2026-05-04
chore(cli): tidy ETH_RPC_URL handling and add forge regression test

zerosnacks:
## Summary

Follow-up to #14555 addressing review feedback.

### Changes

- **Drop redundant flashbots branch in `RpcOpts::dict`.** `self.url(None)` already returns `FLASHBOTS_URL` when `--flashbots` is set, so the explicit `if self.flashbots { dict.insert(...) }` overwrite right after was dead code (it overwrites with the same value).
- **Inline `resolve_rpc_url` back into `RpcCommonOpts::url`.** It was only called from one place and exposed an unnecessary `pub fn` surface.
- **Restore the doc comment on `RpcCommonOpts`** and add a note explaining *why* `ETH_RPC_URL` is intentionally **not** a clap env on the shared field — so future readers don't reintroduce the regression by adding it back.
- **Add a forge integration test** that runs `forge config --json` with `ETH_RPC_URL` set in the environment and asserts that `config.eth_rpc_url` stays `None`. This directly exercises the regression scenario from #14538 (the existing structural tests only assert that the clap arg has no env binding, which would still pass if someone added the env binding on a wrapper struct elsewhere).

### Verification

- `cargo test -p foundry-cli --lib` — all opts tests pass
- `cargo test -p forge --test cli config::eth_rpc_url_env_does_not_set_fork_url` — new regression test passes
- `cargo test -p forge --test cli config::` — all 49 config tests pass

### DIFF
--- a/crates/cli/src/opts/rpc.rs
+++ b/crates/cli/src/opts/rpc.rs
@@ -97,11 +97,10 @@ impl RpcOpts {
 
     pub fn dict(&self) -> Dict {
         let mut dict = self.common.dict();
+        // `self.url(None)` already accounts for `flashbots` and the `ETH_RPC_URL` env var,
+        // so a single insert here covers both.
         if let Ok(Some(url)) = self.url(None) {
             dict.insert("eth_rpc_url".into(), url.into_owned().into());
-        }
-        if self.flashbots {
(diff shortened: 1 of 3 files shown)


figtracer (approved):
nice

mablr (approved):

PR #14560, opened 2026-05-04
feat(tempo): add sponsored transaction plumbing

figtracer:
## Summary
- Adds Tempo sponsor CLI opts: `--tempo.sponsor`, `--tempo.sponsor-signer`, and `--tempo.sponsor-sig`
- Computes and prints the sponsor digest/fee token/validity window before submission
- Attaches `fee_payer_signature` for sponsored cast send/mktx/keychain, forge script, and forge create flows

NOTE: Previously, a caller could pass only the raw sponsor signature and let the node validate it. This PR makes the expected fee payer explicit so the CLI can recover the signature locally, verify it matches the intended sponsor, reject `sponsor == sender`, and print the sponsor identity before submission.

### DIFF
--- a/crates/cli/src/opts/tempo.rs
+++ b/crates/cli/src/opts/tempo.rs
@@ -171,14 +171,11 @@ impl TempoOpts {
     /// Resolves sponsor CLI options into a reusable sponsor config for transaction submission.
     pub async fn sponsor_config(&self) -> Result<Option<TempoSponsor>> {
         let Some(sponsor) = self.sponsor else {
-            if self.sponsor_signer.is_some() || self.sponsor_sig.is_some() {
-                eyre::bail!("--tempo.sponsor is required when configuring Tempo sponsorship");
-            }
             return Ok(None);
         };
 
         let signer = if let Some(spec) = &self.sponsor_signer {
-            Some(Arc::new(Box::pin(resolve_tempo_sponsor_signer(spec)).await?))
+            Some(Arc::new(resolve_tempo_sponsor_signer(spec).await?))
         } else {
             None
         };
--- a/crates/cli/src/opts/tempo.rs
+++ b/crates/cli/src/opts/tempo.rs
@@ -1,11 +1,16 @@
 use alloy_network::{Network, TransactionBuilder};
 use alloy_primitives::{Address, ruint::aliases::U256};
-use alloy_signer::Signature;
+use alloy_signer::{Signature, Signer};
 use clap::Parser;
-use foundry_common::FoundryTransactionBuilder;
+use eyre::Result;
+use foundry_common::{
+    FoundryTransactionBuilder,
+    tempo::{TempoSponsor, resolve_tempo_sponsor_signer},
+};
 use std::{
     num::NonZeroU64,
     str::FromStr,
+    sync::Arc,
     time::{SystemTime, UNIX_EPOCH},
 };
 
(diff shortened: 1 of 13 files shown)


crates/cli/src/opts/tempo.rs
```diff
@@ -130,6 +163,44 @@ impl TempoOpts {
...
+    pub const fn has_sponsor_submission(&self) -> bool {
+        self.sponsor.is_some() || self.sponsor_signer.is_some() || self.sponsor_sig.is_some()
+    }
+
+    /// Resolves sponsor CLI options into a reusable sponsor config for transaction submission.
+    pub async fn sponsor_config(&self) -> Result<Option<TempoSponsor>> {
+        let Some(sponsor) = self.sponsor else {
+            if self.sponsor_signer.is_some() || self.sponsor_sig.is_some() {
+                eyre::bail!("--tempo.sponsor is required when configuring Tempo sponsorship");
+            }
```
mablr: This is already enforced by Clap args ?

crates/script/src/lib.rs
```diff
@@ -861,18 +871,28 @@ mod tests {
...
-            args.tempo.fee_token,
+            args.tempo.common.fee_token,
             Some(address!("0x20C0000000000000000000000000000000000001"))
         );
-        assert_eq!(args.tempo.expires, Some(10));
+        assert_eq!(args.tempo.common.expires, Some(10));
     }
 
     #[test]
-    fn does_not_parse_unsupported_tempo_opts() {
```
mablr, line 871: This test was removed

crates/script/src/lib.rs
```diff
@@ -715,6 +721,8 @@ pub struct ScriptConfig<FEN: FoundryEvmNetwork> {
     pub fee_token: Option<Address>,
     /// TIP-1009 expiring-nonce valid_before unix timestamp, derived from `--tempo.expires`.
     pub expires_at: Option<u64>,
+    /// Tempo transaction options applied to broadcast transactions.
+    pub tempo: TempoOpts,
```
mablr, line 725: We should use `TempoCommonOpts` I think.

crates/script/src/broadcast.rs
```diff
@@ -137,11 +148,28 @@ where
...
+
             // Chains which use `eth_estimateGas` are being sent sequentially and require their
             // gas to be re-estimated right before broadcasting.
             if !is_fixed_gas_limit && estimate_via_rpc {
                 estimate_gas(tx, provider, estimate_multiplier).await?;
             }
+
+            if let Some(sponsor) = tempo_sponsor {
+                let from = tx.from().expect("no sender");
+                sponsor.attach_and_print::<N>(tx, from).await?;
```
mablr, line 171: This prints on each retry, is this what we want ?

crates/cli/src/opts/tempo.rs
```diff
@@ -130,6 +163,44 @@ impl TempoOpts {
...
+            return Ok(None);
+        };
+
+        let signer = if let Some(spec) = &self.sponsor_signer {
+            Some(Arc::new(Box::pin(resolve_tempo_sponsor_signer(spec)).await?))
```
mablr, line 178: I think `Box::pin` may be unnecessary here.

mablr (changes requested):
Lgtm, some nits

crates/script/src/lib.rs
figtracer: oops leaked from another branch adding it back

crates/script/src/lib.rs
figtracer: the problem with using `TempoCommonOpts` here is that it gets quite messy with the sponsor variants

crates/script/src/broadcast.rs
figtracer: for gas re-estimation i think it can be useful

crates/cli/src/opts/tempo.rs
figtracer: we get `queries overflow the depth limit!` without pinning the future

mablr: okay 😅

mablr (dismissed):
sgtm

zerosnacks (approved):
Nice! 👍

mablr (approved):

PR #14562, opened 2026-05-04
fix(cli): fix jsonwebtoken panic

pepyakin:
`cast` panicked with this message coming from jsonwebtoken:

```
Call CryptoProvider::install_default() before this point to select a provider manually, or make sure exactly one of the
'rust_crypto' and 'aws_lc_rs' features is enabled.
See the documentation of the CryptoProvider type for more information.
```

This seemingly was introduced with the bump of jsonwebtoken to 10. Now it requires you to pick one backend used by default controlled by the compile time cargo features or call `CryptoProvider::install_default()` at the beginning.

I realised that probably it would be better to just select the feature and I picked `aws_lc_rs` as it seems to be increasingly a default and we already are using the C toolchain. The dependency is used through `alloy-rpc-types-engine` and it already exposes a transitive feature which is being toggled here.


zerosnacks:
Thanks @pepyakin! Agreed on the fix

zerosnacks (approved):

figtracer (approved):

PR #14568, opened 2026-05-04
feat(doc): solar-based rewrite + vocs migration

### LINKED ISSUE #11884: Replace solang in forge doc
Last remaining task in https://github.com/foundry-rs/foundry/issues/10960

mablr:
## Summary

Rewrite solar-based `forge doc` (#14447), migrate to `vocs`.

- Generates `vocs` scafold folder with produced MDX files
- Removes `--serve` feature to avoid depending on npm, now user has to run `forge doc --watch`, and then `npm run dev` in another terminal.
- **Sidebar restructured**: within each directory group, pages are nested under type sub-groups (*Contracts*, *Abstract Contracts*, *Libraries*, *Interfaces*, …). Directories that contain only one type list items flat. Labels show just the contract name, no `type.` prefix.
- **MDX safety fixes**: `replace_inline_links` now escapes bare `<` -> `&lt;` (fixes `<email@example.com>` parsed as JSX), unresolved `{Ident}` -> `` `Ident` `` inline code, bare `{` -> `&#123;`. Unnamed return parameters rendered as `&lt;none&gt;` instead of `<none>`. Inherited natspec content (from `@inheritdoc` resolution) now runs through the same sanitizer.
- **Constants/immutables section**: `constant` and `immutable` state variables are rendered under a `## Constants` section, separate from mutable `## State Variables`.
- **CLI grouping**: serve-related flags (`--hostname`, `--port`, `--open`) moved into a dedicated `Serve options` help section so they no longer appear under `Watch options`.

Closes #11884
Closes OSS-93

## PR Checklist

- [x] Added Tests (updated)
- [x] Added Documentation
- [x] Breaking changes

Tested against vectorized/solady and ithacaxyz/account. 

image


crates/doc/src/hir_ext.rs
```diff
@@ -0,0 +1,362 @@
...
+
+/// Build the `NameToPage` map from HIR by re-deriving each item's output path.
+///
+/// This mirrors the path computation in `render::source` so links can be resolved
+/// before rendering begins.
+pub fn build_name_to_page(gcx: Gcx<'_>, root: &Path) -> NameToPage {
+    let mut map = NameToPage::new();
+
+    // Contracts.
+    for id in gcx.hir.contract_ids() {
```
DaniPopes: can use `.item_ids()`

crates/doc/src/utils.rs
DaniPopes, line 1: maybe utils.rs

crates/doc/README.md
DaniPopes, line 1: this could still be useful with a small overview and then inline it into the crate docs with include_str!

crates/doc/src/hir_ext.rs
mablr: Yes, way better, thank you.

zerosnacks (commented):
Overall super nice! Some small points of feedback

I would like to see a differentiation in style between `@notice` and `@dev`, the latter I think should be rendered in italic. 

Image

Ref: http://localhost:5173/src/utils/g/library.EnumerableSetLib

In `forge-std` as example a problem with function overrides is visible, maybe we can come up with something that makes it easier to distinguish between them.

Image

zerosnacks:
I am also seeing brittleness with Vocs where just clicking around will cause `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`. Can reliably reproduce w/ `forge-std` and clicking around in the libraries section. Will report this to Vocs.

zerosnacks:
In `forge-std`, this looks off `;;`

Screenshot 2026-05-05 at 16 10 12

zerosnacks:
From Amp's oracle review, I think these findings are broadly accurate, worth double checking:

### Blockers

1. **UDVT xrefs land on non-existent pages.** `hir_ext.rs:62` indexes top-level
   UDVTs with prefix `"udvt"`, but `render.rs:113` writes the file as
   `type.<Name>.mdx`. So `{MyType}` → `/.../udvt.MyType` while the file on
   disk is `/.../type.MyType`.

2. **`name_to_page` is built from the full HIR, not the emitted page set.**
   `builder.rs:67-81` filters via `ignore`/`include_libraries`, but
   `builder.rs:97` walks every `gcx.hir.item_ids()`. Result: `{Ident}` and
   inheritance links can resolve to library/ignored pages that are never
   written. Map should be derived from the same filtered set, or pruned after
   render.

3. **`@inheritdoc` resolves by name only.** `hir_ext.rs:155-170` returns the
   first function in the linearised base whose name matches. For real
   same-name overloads (`foo(uint256)` / `foo(address)`) the wrong one wins.
   The `can_generate_docs_for_overloaded_functions` test in
   `tests/cli/doc.rs:15-67` actually uses two different names, so this case
   is uncovered. Consider matching by signature (param type list).

4. **Render panics are swallowed and the command still returns `Ok`.**
   `builder.rs:155-178` `catch_unwind`s and logs an `error!`, but the loop
   continues and `build()` returns `Ok(())`. A user can end up with missing
   pages and a green exit. Suggest at minimum failing the build when any
   non-library file panics, or surfacing a non-zero exit + summary.

5. **Frontmatter is not YAML-escaped.** `render.rs:741-749` writes
   `description: "{truncated}"` from the first `@notice`. Quotes / newlines /
   backslashes in a notice break the YAML. The 120-char `chars().take(120)`
   truncation can also slice mid-entity (e.g. truncating in the middle of an
   `&lt;` produced by `replace_inline_links`). Either escape per YAML rules
   or use a block scalar (`description: |`).

6. **URLs use OS-native path separators.** `hir_ext.rs:345`, `vocs.rs:95`,
   `vocs.rs:102-106`, and `utils.rs:24` all build URLs via `Path::display()`,
   so on Windows you get backslashes in vocs routes and GitHub `blob/` links.
   Foundry already standardises on `path-slash`'s `.to_slash_lossy()` for
   exactly this — see e.g. `crates/common/src/preprocessor/data.rs:84`,
   `crates/forge/src/cmd/inspect.rs:478`, `crates/forge/src/cmd/bind_json.rs:318`,
   `crates/config/src/lib.rs:1115`. Suggest using the same here for any
   path that ends up in a URL or vocs config.

### Major

7. **`find_contract_id` is suffix-based** (`render.rs:909`). Two files
   sharing a tail (e.g. `src/A.sol` and `lib/foo/src/A.sol`) will resolve
   to whichever the iterator returns first. Comparing absolute paths after
   strip-prefix would be safer.

8. **Sidebar labels collide / nesting is flat.**
   - `vocs.rs:113-123`: comment says "preserving type prefix in name for
     clarity", but `name` was already stripped — `contract.Foo` and
     `interface.Foo` both display as `Foo`.
   - `vocs.rs:102-106`: nested directories are emitted as a single label
     like `src/token/extensions` instead of a true tree.
   - No collision check across categories.

9. **`forge doc --watch` only watches `config.src`** (`watch.rs:388-391`).
   Now that `--watch` is the documented preview flow (`cmd/doc.rs:87-93`),
   it should also include:
   - libraries (when `--include-libraries` is set)
   - the project `README.md` — currently picked up as the default homepage
     in `vocs.rs:213` but never re-rendered when edited
   - any `config.homepage` override
   - the deployments dir (`<root>/deployments` or the `--deployments` path)
   - foundry config / doc config
   Otherwise the live preview goes silently stale on the most common edits
   (README + deployments).

10. **No stale-page pruning** (`builder.rs:182-194`). Renames or deletions
    leave orphan MDX files — particularly painful in watch mode and for
    static deploys. Consider tracking the previous run's outputs (or
    truncating `pages_dir` for non-scaffold files) before writing.

11. **Markdown tables don't escape `|` or newlines** (`render.rs:840`).
    `@param`/`@return` text containing `|` will shift columns. The MDX
    sanitiser handles `<`/`{` but not table delimiters.

12. **Homepage comment vs impl mismatch.** `vocs.rs:28` lists
    `config.homepage -> <sources>/README.md -> <root>/README.md -> empty`
    but `vocs.rs:203-219` skips the `<sources>/README.md` step.

13. **`master` vs `main` inconsistency.** `utils.rs:22` falls back to
    `"master"` for source links; `vocs.rs:53` hardcodes `edit/main/{path}`.
    Repos using a different default branch get one of the two broken.

DaniPopes:
code side lgtm, will leave iteration on frontend to you

crates/doc/src/hir_ext.rs
```diff
@@ -0,0 +1,419 @@
...
+        returns: Vec::new(),
+    };
+    for doc in docs.iter() {
+        for item in doc.natspec.iter() {
+            let content = doc.natspec_content(item).trim().to_string();
+            if content.is_empty() {
+                continue;
+            }
+            match item.kind {
+                NatSpecKind::Notice => result.notices.push(content),
```
figtracer: can we reuse the direct NatSpec collector here (`render.rs`) otherwise we might generate some extra blank lines; tested locally

mablr: good catch ty

figtracer (dismissed):

zerosnacks:
Screenshot 2026-05-08 at 16 46 32

Found a small formatting issue with `/** */` style multiline.

```
/**
 * @dev Elliptic Curve Digital Signature Algorithm (ECDSA) operations.
 *
 * These functions can be used to verify that a message was signed by the holder
 * of the private keys of a given address.
 */
 ```

Ref: https://github.com/aave/aave-v4/blob/af1f0f2ba323ac6fbaaee3abf6be060c78e22d35/src/dependencies/openzeppelin/ECDSA.sol#L6-L11

Screenshot 2026-05-08 at 16 51 41

Same here, we should preserve whitespace / newlines

**Comment by zerosnacks** [comment]
Other earlier flagged issues have been resolved 👍 ! Looks nice

**Comment by zerosnacks** [comment]
Small other enhancement for `forge doc`; like `forge build, would be nice if it would install missing dependencies if encountered.

Screenshot 2026-05-08 at 16 57 44

Screenshot 2026-05-08 at 16 58 04

**Comment by zerosnacks** [comment]
a nice to have - w/ https://github.com/morpho-org/morpho-blue; looks like some of the links in the README are not rewritten to point to the documentation

Screenshot 2026-05-08 at 17 02 33

**Comment by zerosnacks** [comment]
Screenshot 2026-05-08 at 17 23 46

On large codebases like https://github.com/Uniswap/v4-core `forge doc` can take quite some time to run; it would be great if we could get some visual feedback for the user with details on what it is doing / progressing. I believe that we are building the codebase as a requirement for `forge doc`, this should be displayed to the user similar to:

Screenshot 2026-05-08 at 17 27 11

Ideally with following `-vvvvv` for various verbosity levels but at first any feedback is valuable here.

File: crates/doc/src/utils.rs [hunk 3]
```diff
@@ -0,0 +1,80 @@
...
+/// Build the `{repo}/blob/{commit}/{rel}` URL for a source file.
+///
+/// Returns `None` if `item_path` is not under `root` (i.e. for absolute external paths).
+/// `commit` falls back to `"HEAD"` when empty (GitHub's `blob/HEAD/...` resolves
+/// to the repository's default branch regardless of whether it is `main`,
+/// `master`, or anything else).
+///
+/// Path components are joined with `/` so the URL is well-formed on Windows.
+pub fn git_source_url(repo: &str, commit: &str, root: &Path, item_path: &Path) -> Option<String> {
+    use path_slash::PathExt;
```
stevencartavia: let's import on top

zerosnacks (commented):
Four high-severity findings inline (NatSpec param/return continuation, shallow `@inheritdoc` resolution, remaining MDX-injection paths in `replace_inline_links`, and destructive Vocs regeneration). Several medium items inline as well.

Three low-severity items not tied to a specific line:

- **Deployments rendering is nondeterministic and unescaped.** `read_deployments` in `crates/doc/src/utils.rs` iterates `fs::read_dir` in unspecified order; `write_deployments_table` in `crates/doc/src/render.rs` writes `network` straight into a markdown table cell without `escape_table_cell`. Output can churn across machines, and oddly named directories can break table formatting.
- **MDX-safety claims are not directly tested.** I don't see assertions in `crates/forge/tests/cli/doc.rs` for bare `<` -> `&lt;`, bare `{` -> `&#123;`, unresolved `{Ident}` -> inline code, sanitized inherited `@inheritdoc` content, unnamed returns -> `&lt;none&gt;`, or inline-link label safety. The areas most likely to regress have the weakest coverage.
- **`crates/doc/README.md` is out of sync.** It refers to `foundry-doc` (vs the actual `forge-doc`), describes vocs as emitting separate `sidebar.json` / `config.ts` (the code inlines the sidebar in `vocs.config.ts`), and labels `render` as a pure AST -> MDX layer though it depends on HIR lookups for inheritance / `@inheritdoc`.

Areas checked and not flagged: no obvious user-triggerable panic path, sidebar ordering is deterministic (`BTreeMap` + sort in `build_sidebar`), and `try_rewrite_target`'s `strip_prefix(root)` is a reasonable traversal guard.

crates/doc/src/hir_ext.rs
```diff
@@ -0,0 +1,471 @@
...
+                NatSpecKind::Param { name } => {
+                    result.params.push((name.as_str().to_string(), content))
+                }
+                NatSpecKind::Return { name } => {
+                    result.returns.push((name.as_str().to_string(), content))
+                }
+                _ => {}
+            }
+        }
+    }
```
zerosnacks, line 616: **High — multiline `@param` / `@return` continuation lines are misclassified as notices.**

In `collect_comments` (and the same pattern in `collect_inherited_doc` around L286-335), only the `notice`/`dev` paths handle continuation lines. A doc block like:

```solidity
/// @param x first line
///        second line
```

produces a `NatSpecKind::Notice` for the second line, which is appended to `descriptions` rather than extending the existing `@param x` entry. Result: param/return descriptions render incorrectly or get truncated, and the bug also affects content pulled in via `@inheritdoc`.

crates/doc/src/hir_ext.rs
```diff
@@ -0,0 +1,471 @@
...
+            if let Some(got) = function_param_types(gcx, fid)
+                && got.len() == want.len()
+                && got.iter().zip(want).all(|(a, b)| a == b)
+            {
+                return extract_inherited_doc(gcx, fid);
+            }
+        }
+    }
+
+    name_matches.first().copied().and_then(|fid| extract_inherited_doc(gcx, fid))
```
zerosnacks: **High — `@inheritdoc Base` only searches that base's own items, not its inherited chain.**

The doc comment claims this walks the linearized bases, but after locating the named base in `contract.linearized_bases` we only iterate `base_contract.items`. If `Base` itself inherits the member (e.g. from an interface) without redeclaring the NatSpec locally, resolution silently drops and no inherited docs are merged.

Resolution should continue walking `Base`'s own linearized bases (or fall back to a full chain search) when the member isn't found directly on `Base`.

crates/doc/src/hir_ext.rs
```diff
@@ -0,0 +1,471 @@
...
+                    let display = label.unwrap_or(lookup_name);
+                    out.push_str(&format!("[{display}]({link})"));
+                    i += end;
+                    continue;
+                }
+
+                // Unresolved {Ident}, emit as inline code to avoid MDX treating it as a
+                // JS expression.
+                out.push_str(&format!("`{lookup_name}`"));
+                i += end;
```
zerosnacks, line 697: **High — resolved inline links interpolate user-controlled text into MDX without escaping.**

When a `{Ident}` (optionally `[label]`) reference resolves, this emits `[{display}]({link})` with `display` taken from the user-provided label / identifier verbatim. Combined with the unresolved branch below, the PR's MDX-safety story still has an injection path through parsed inline-link forms — e.g. labels containing `<`, `{`, `]`, or stray brackets are not run through the new sanitizer.

The `display` (and link target) should pass through the same escape logic used elsewhere for inline MDX.

crates/doc/src/hir_ext.rs
```diff
@@ -0,0 +1,471 @@
...
+            consumed += end + 1;
+            Some(lbl)
+        } else {
+            None
+        }
+    } else {
+        None
+    };
+
+    Some((consumed, raw_ident, raw_part, label))
```
zerosnacks, line 812: **High (same root cause) — unresolved inline references emit `` `{lookup_name}` `` with the raw identifier.**

`lookup_name` is parsed out of arbitrary NatSpec text, so inputs containing backticks or other markdown/MDX-sensitive characters will break the inline-code fence or escape it entirely. Consider routing this through the sanitizer too, or asserting/sanitizing the identifier shape before wrapping it in backticks.

crates/doc/src/hir_ext.rs
```diff
@@ -0,0 +1,471 @@
...
+                let lookup_name = if let Some(pos) = lookup_name.find('-') {
+                    &lookup_name[..pos]
+                } else {
+                    lookup_name
+                };
+
+                if let Some(page) = name_to_page.get(lookup_name) {
+                    let link = page_link(page, current_page);
+                    let display = label.unwrap_or(lookup_name);
+                    out.push_str(&format!("[{display}]({link})"));
```
zerosnacks, line 688: **Medium — member references `{Type-member}` are parsed but the `member` part is discarded.**

`parse_inline_link` returns `(ident, part, label)` but the resolved-link branch ignores `_part` and links only to the page root. That's a regression vs the legacy syntax (which the comment explicitly mentions) and means existing NatSpec like `{ERC20-transfer}` now lands the reader on the page header instead of the member anchor.

crates/doc/src/hir_ext.rs
```diff
@@ -0,0 +1,471 @@
...
+            continue;
+        }
+
+        if let Some((abs, rel)) = source_paths(gcx, source, root) {
+            if !allowed_sources.contains(&abs) {
+                continue;
+            }
+            let out_dir = rel.parent().unwrap_or(Path::new("")).to_owned();
+            let page = out_dir.join(format!("{prefix}.{}.mdx", name.as_str()));
+            map.insert(name.as_str().to_string(), page);
```
zerosnacks: **Medium — cross-reference resolution keyed by bare name only.**

`NameToPage = HashMap<String, PathBuf>` plus `map.insert(name, page)` in `build_name_to_page` means duplicate top-level names across files/directories overwrite each other (last-insert wins, ordering depends on traversal). `{Foo}` references and inheritance links can silently resolve to the wrong page in repos with duplicate contract/type names — common in large Solidity trees and vendored libs.

Consider keying by a (path, name) tuple, or resolving against the contract's own import / inheritance scope first.

crates/doc/src/builder.rs
```diff
@@ -57,419 +66,256 @@ impl DocBuilder {
...
+        let root = self.root.clone();
 
-        // Collect and parse source files
-        let sources = source_files_iter(&self.sources, SOLC_EXTENSIONS)
-            .filter(|file| !ignored.contains(file))
-            .collect::<Vec<_>>();
+        let all_pages = compiler.enter_mut(|compiler| -> eyre::Result<Vec<PathBuf>> {
+            if compiler.gcx().stage() < Some(solar::config::CompilerStage::Lowering) {
+                let _ = compiler.lower_asts();
+            }
```
zerosnacks, line 124: **Medium — AST/HIR lowering errors are explicitly discarded (`let _ = compiler.lower_asts()`).**

If lowering fails, the doc build proceeds with partial HIR state, which can lead to missing pages, bad cross-resolution, or harder-to-diagnose downstream failures. At minimum, surface the error (warn or bail), or short-circuit before traversal.

crates/doc/src/builder.rs
```diff
@@ -57,419 +66,256 @@ impl DocBuilder {
...
-        };
 
-        // Merge two book configs
-        if let Some(book_path) = book_path {
-            merge_toml_table(&mut book, toml::from_str(&fs::read_to_string(book_path)?)?);
-        }
+            // Prune stale `.mdx` pages left behind by previous runs (renames,
+            // deletions). The vocs scaffold's own `index.mdx` is preserved.
+            let kept: HashSet<PathBuf> = all_rel.iter().cloned().collect();
+            prune_stale_pages(&pages_dir, &kept)?;
```
zerosnacks: **High — destructive prune.**

`prune_stale_pages` deletes **every** `.mdx` under `docs/src/pages` that isn't in the freshly generated set (except top-level `index.mdx`). Combined with `write_site_files` unconditionally overwriting `vocs.config.ts`, `package.json`, `.gitignore`, and `src/pages/index.mdx` (see `vocs.rs`), rerunning `forge doc` wipes any hand-authored guide pages or scaffold customizations a user added.

Options to consider: scope pruning to a generated subdirectory (e.g. `pages/contracts/`), or skip files marked with a sentinel / front-matter `generated: true`. At minimum, document the destructive behavior loudly in the README.

crates/doc/src/vocs.rs
```diff
@@ -0,0 +1,443 @@
...
+            commit,
+        )
+    } else {
+        homepage_content
+    };
+    let index_path = out_dir.join("src").join("pages").join("index.mdx");
+    if let Some(parent) = index_path.parent() {
+        fs::create_dir_all(parent)?;
+    }
+    fs::write(index_path, homepage_content)?;
```
zerosnacks: **High (same root issue as the prune above) — `write_site_files` unconditionally overwrites scaffold files.**

`.gitignore`, `package.json`, `vocs.config.ts`, and `src/pages/index.mdx` get re-emitted on every `forge doc` run regardless of user edits. Users who tweak vocs config, add deps to `package.json`, or customize the landing page lose those changes silently.

Consider writing these only if absent (or hashing a managed marker block and only overwriting that block).

crates/doc/src/vocs.rs
```diff
@@ -0,0 +1,443 @@
...
+
+    // Split off `#fragment` / `?query`; we'll preserve the fragment on the rewrite.
+    let (path_part, suffix) = target.find(['#', '?']).map_or((target, ""), |i| target.split_at(i));
+    if path_part.is_empty() {
+        return None;
+    }
+
+    let path = Path::new(path_part);
+    let abs = if path.is_absolute() { path.to_path_buf() } else { base_dir.join(path) };
+    let rel = normalize_path(&abs).strip_prefix(root).ok()?.to_path_buf();
```
zerosnacks: **Medium — `rewrite_homepage_links` is a naive parser.**

It searches for `"]("` and the first `)` afterwards, which breaks any link whose URL contains parentheses, titles, or images-inside-links. Also, leading-slash targets are treated as filesystem-absolute (`strip_prefix(root)`), so repo-root-style links won't be rewritten unless they happen to live under the actual FS root prefix.

A proper markdown parser (or at least a more careful URL boundary scan handling escaped `)`) would avoid corrupting otherwise-valid links.

crates/forge/src/cmd/doc.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// Path to the `hardhat-deploy` or `forge-deploy` artifact directory.
+    ///
+    /// Leave blank to use the default (`<root>/deployments`).
+    /// Omit the flag entirely to disable deployment address injection.
+    #[arg(long, value_name = "PATH", num_args(0..=1))]
+    pub deployments: Option<Option<PathBuf>>,
+
+    #[command(flatten)]
+    pub watch: WatchArgs,
+}
```
zerosnacks, line 47: **Medium — `--serve` is a hard CLI break with no shim or migration message.**

Existing automation / muscle memory invoking `forge doc --serve` now hits a raw clap parse error. Since this is called out as a breaking change in the PR description, consider either keeping a hidden/deprecated `--serve` that prints a clear migration message ("run `npm run dev` from the generated vocs dir") or wiring a custom error in the parser. The current tests in `crates/forge/tests/cli/doc.rs` don't exercise the removed flag path either.

crates/doc/src/vocs.rs
zerosnacks: This is OK

mablr (commented):
### Fully addressed

- multiline @\param/@\return continuations: tracked via `LastSection` enum in both `collect_comments` and `collect_inherited_doc`.
- @\inheritdoc **Base** inherited chain: now walks `base_contract.linearized_bases` when the member isn't on `Base` directly.
- unresolved `{Ident}` backticks: strips backticks before wrapping in inline code.
- `{Type-member}` anchors: appends #anchor and defaults display to Type.member.
- destructive prune: scoped to docs/src/pages/src/; user-authored pages outside are preserved.
- scaffold overwrites: .gitignore, package.json, vocs.config.ts, index.mdx use `write_if_absent`; sidebar split into always-regenerated vocs.sidebar.ts.
- naive homepage link parser: rewritten with paren-depth tracking and escape handling.
- `--serve` migration: hidden flag bails with migration message; covered by `serve_flag_prints_migration_message` test.
- `NameToPage` collisions
- lowering errors bailed.

### Partially addressed

- link target escaping: could improved in a follow-up to handle `)(` better ?

figtracer (dismissed):

emmajam:
*PR #14568 review findings* ⚠️ from clanker

*P1 — `--include-libraries` can write outside `docs/`*
`crates/doc/src/builder.rs:155`, `crates/doc/src/builder.rs:225`
External absolute library paths can remain absolute, and `pages_dir.join(abs_path)` discards `pages_dir`. A config like `libs = ["/opt/shared"]` can write MDX files into `/opt/shared` instead of `docs/src/pages`.

*P2 — README/homepage output goes stale*
`crates/doc/src/vocs.rs:52`, `crates/doc/src/vocs.rs:70`, `crates/forge/src/cmd/watch.rs:410`
The homepage is reread, but `index.mdx` is only written if absent. Rerunning `forge doc` or editing `README.md`/`doc.homepage` in watch mode leaves stale homepage content until users delete `docs/src/pages/index.mdx`.

*P2 — stale-page pruning only handles default `src`*
`crates/doc/src/builder.rs:256`
Pruning is hard-coded to `docs/src/pages/src`. Projects using `src = "contracts"` or generated library docs under `lib/` keep deleted/renamed pages on disk, so Vocs can still serve stale docs.

*P2 — README Markdown is copied into MDX without MDX escaping*
`crates/doc/src/vocs.rs:54`, `crates/doc/src/vocs.rs:70`
README content only gets link rewriting before becoming `index.mdx`. Common Markdown placeholders like `{FOO}` or `<TOKEN>` can be parsed as MDX expressions/JSX and break `vocs dev` / `vocs build`.

*P2 — overloaded `@inheritdoc` can still inherit wrong docs*
`crates/doc/src/hir_ext.rs:304`, `crates/doc/src/hir_ext.rs:324`, `crates/forge/tests/cli/doc.rs:17`
Overload matching compares raw source type strings, then falls back to the first same-name function. Canonical equivalents like `uint` vs `uint256` can miss the exact match and inherit docs from the wrong overload. The current overload test uses different function names, so this case is uncovered.

*P2 — invalid `doc.ignore` globs are silently ignored*
`crates/doc/src/builder.rs:91`
`expand_globs(...).unwrap_or_default()` drops invalid ignore config without warning, so users can accidentally publish files they expected to exclude.

mablr (commented):
## Fixed

- **Stale homepage:** `index.mdx` now always regenerates on every `forge doc` run (was `write_if_absent`).
- **README MDX escaping:** homepage content now has `{` → `\{` and `<` → `&lt;` escaped outside code fences before writing `index.mdx`.
- **Silent glob errors:** invalid `doc.ignore` globs now emit a warning instead of silently dropping all ignore rules.

## Not fixed

- **P1 path traversal:** not real; library paths from `source_files_iter` are relative so `pages_dir.join` stays within `pages_dir`.
- **Pruning hard-coded to `src`:** valid, but i don't think it's important
- **`@inheritdoc` overload canonical types** (`uint` vs `uint256`): valid, needs HIR type resolution, deferred.

mattsse (changes requested):
Requesting changes on these doc-generation issues before merge.

High: [`builder.rs#L158-L159`](https://github.com/foundry-rs/foundry/blob/90061bda13b2444ffd339e27667553306ce2e6ea/crates/doc/src/builder.rs#L158-L159) can leave `rel_path` absolute for external/library sources outside `root`; then [`pages_dir.join(...)`](https://github.com/foundry-rs/foundry/blob/90061bda13b2444ffd339e27667553306ce2e6ea/crates/doc/src/builder.rs#L228-L232) can write outside the docs tree. Please normalize external sources to a synthetic relative path and reject any output path with absolute/root/parent components before writing.

Medium: overload matching for `@inheritdoc` compares raw type strings at [`hir_ext.rs#L304-L324`](https://github.com/foundry-rs/foundry/blob/90061bda13b2444ffd339e27667553306ce2e6ea/crates/doc/src/hir_ext.rs#L304-L324), then falls back to the first documented name match. That can inherit docs from the wrong overload. Please match canonical ABI/HIR types, and when multiple overloads exist, avoid name-only fallback if the requested signature cannot be matched exactly.

Medium: `@inheritdoc` resolution currently only runs for functions at [`render.rs#L270-L281`](https://github.com/foundry-rs/foundry/blob/90061bda13b2444ffd339e27667553306ce2e6ea/crates/doc/src/render.rs#L270-L281), while [`collect_comments`](https://github.com/foundry-rs/foundry/blob/90061bda13b2444ffd339e27667553306ce2e6ea/crates/doc/src/render.rs#L731) drops `Inheritdoc`. Public vars/member docs can regress. Please preserve the inheritdoc marker and resolve it for variables/members too, with coverage for inherited public state vars.

Medium: stale page pruning only covers `pages/src` at [`builder.rs#L253-L264`](https://github.com/foundry-rs/foundry/blob/90061bda13b2444ffd339e27667553306ce2e6ea/crates/doc/src/builder.rs#L253-L264). Generated pages for custom source roots or libraries can remain forever after renames/deletes. Please either prune every generated subtree or keep a generated-file manifest and delete stale entries from the previous run.

Medium: README link rewriting treats leading-slash links as OS-absolute paths at [`vocs.rs#L447-L449`](https://github.com/foundry-rs/foundry/blob/90061bda13b2444ffd339e27667553306ce2e6ea/crates/doc/src/vocs.rs#L447-L449), so repo-root links like `/src/Foo.sol` fail. Also the asset fallback at [`vocs.rs#L460-L463`](https://github.com/foundry-rs/foundry/blob/90061bda13b2444ffd339e27667553306ce2e6ea/crates/doc/src/vocs.rs#L460-L463) points images to GitHub blob HTML instead of raw assets. Please resolve leading `/` against the project root, and emit raw/download URLs for image targets.

crates/doc/src/builder.rs
```diff
@@ -57,419 +66,313 @@ impl DocBuilder {
...
-        // Attempt to find the user provided book path
-        let book_path = {
-            if self.config.book.is_file() {
-                Some(self.config.book.clone())
-            } else {
-                let book_path = self.config.book.join("book.toml");
-                book_path.is_file().then_some(book_path)
+            let new_generated: HashSet<PathBuf> = all_rel.iter().cloned().collect();
+            for stale in prev_generated.difference(&new_generated) {
+                let stale_abs = pages_dir.join(stale);
```
mattsse, line 307: High: the manifest contents are trusted before deletion. If `.forge-doc-manifest` is committed or edited with an entry like `../../../src/Foo.sol`, `pages_dir.join(stale)` escapes `docs/src/pages` and this branch can delete project files outside the docs tree.

Fix suggestion: parse each manifest entry through the same safety predicate used for rendered output paths before joining it. Reject absolute paths, root/prefix components, and any `..` component; warn and skip invalid entries. Please add a regression that writes a manifest with a parent-dir entry and asserts `forge doc` leaves the outside file intact.

crates/doc/src/hir_ext.rs
```diff
@@ -0,0 +1,712 @@
...
+        }
+    }
+
+    map
+}
+
+fn source_paths(gcx: Gcx<'_>, source_id: SourceId, root: &Path) -> Option<(PathBuf, PathBuf)> {
+    let file = &gcx.hir.source(source_id).file;
+    if let FileName::Real(p) = &file.name {
+        let rel = p.strip_prefix(root).unwrap_or(p).to_owned();
```
mattsse: Medium: this reuses the real absolute path for sources outside `root`, while `builder.rs` now renders those same files under a synthetic `lib/<tail>` relative path. With `--include-libraries` and an absolute external lib, `NameToPage` can index pages under an absolute-looking path (`//...`) that is different from the actual emitted `lib/...` page.

Fix suggestion: compute the emitted relative path in one shared helper and use it both here and in `builder.rs` before rendering. That helper should return `(abs_source, emitted_rel_source)` and apply the `lib/<tail>` normalization for outside-root files. Add coverage for an absolute library path with an intra-library `{Ident}` link.

crates/doc/src/render.rs
```diff
@@ -0,0 +1,1059 @@
...
+/// Find the HIR `ContractId` for a contract by name, requiring the contract to
+/// live in the source file currently being rendered (compared via absolute path)
+/// so contracts that share a file stem across `src/` and `lib/` cannot collide.
+fn find_contract_id<'gcx>(
+    gcx: Gcx<'gcx>,
+    name: &str,
+    rel_sol_path: &Path,
+    root: &Path,
+) -> Option<hir::ContractId> {
+    let want = root.join(rel_sol_path);
```
mattsse: Medium: this assumes every rendered `rel_sol_path` can be converted back to the real source by `root.join(rel_sol_path)`, but outside-root library sources are now passed in as synthetic `lib/<tail>` paths. For those pages, `find_contract_id` returns `None`, so inheritance links and variable/function `@inheritdoc` resolution silently disappear.

Fix suggestion: pass the real absolute source path into `render::source` alongside the emitted relative path, and make `find_contract_id` compare against that absolute path. A regression should cover `--include-libraries` with an absolute lib containing an inherited contract/member doc.

crates/forge/src/cmd/watch.rs
```diff
@@ -384,11 +384,60 @@ pub async fn watch_fmt(args: FmtArgs) -> Result<()> {
...
+        }
+
+        // Homepage source: explicit override, else <root>/README.md when present.
+        if let Some(hp) = &config.doc.homepage {
+            let hp_path = if hp.is_absolute() { hp.clone() } else { root.join(hp) };
+            if hp_path.exists() {
+                paths.push(hp_path);
+            }
+        }
+        let readme = root.join("README.md");
```
mattsse, line 423: Medium: `forge doc --watch` still misses the `<sources>/README.md` homepage. `vocs::find_homepage` prefers `config.homepage`, then `<sources>/README.md`, then `<root>/README.md`, but the watcher only tracks the explicit homepage and root README. Editing `src/README.md` leaves the preview stale.

Fix suggestion: mirror `find_homepage` here: after the explicit override, watch `config.src.join("README.md")` when it exists, then root `README.md`. Add a watch path unit/regression if there is already a helper pattern for these defaults.

crates/doc/src/hir_ext.rs
```diff
@@ -0,0 +1,712 @@
...
+}
+
+/// Compute the parameter type strings of a [`ParameterList`] from the source map.
+pub fn parameter_type_strings(gcx: Gcx<'_>, params: &ParameterList<'_>) -> Vec<String> {
+    let sm = gcx.sess.source_map();
+    params
+        .vars
+        .iter()
+        .map(|v| {
+            sm.span_to_snippet(v.ty.span)
```
mattsse: Medium: overload matching still compares raw source snippets, so canonical equivalents like `uint` and `uint256` do not match. The current fallback then skips all same-name overload docs when a signature was requested, which avoids inheriting the wrong overload but still drops valid `@inheritdoc` docs.

Fix suggestion: derive canonical ABI/HIR parameter types instead of `span_to_snippet`, or normalize Solidity aliases (`uint` -> `uint256`, `int` -> `int256`, etc.) before comparing. Please add a test with `Base.foo(uint256)` documented and `Child.foo(uint)` using `@inheritdoc Base`.

mattsse (dismissed):
Requesting changes for one file-deletion safety issue and a few remaining doc-generation regressions. The fixes are scoped: validate manifest paths before deleting, share the external-source path normalization with HIR/link lookup, watch the same homepage path that Vocs reads, and use canonical parameter types for `@inheritdoc` overload matching.

crates/doc/src/hir_ext.rs
```diff
@@ -0,0 +1,726 @@
...
+
+    let base_contract = gcx.hir.contract(base_id);
+    let search_contracts: Vec<ContractId> = std::iter::once(base_id)
+        .chain(base_contract.linearized_bases.iter().copied().filter(|&id| id != base_id))
+        .collect();
+
+    for search_id in &search_contracts {
+        let search_contract = gcx.hir.contract(*search_id);
+        for &item_id in search_contract.items {
+            if let ItemId::Variable(vid) = item_id {
```
figtracer: This only searches base `ItemId::Variable`s, so `/// @inheritdoc IERC20` on a public state variable that implements an interface getter like `function totalSupply() external view returns (uint256)` still renders without the inherited notice/return docs. Consider matching same-name zero-arg functions as well when resolving variable inheritdoc, and add a regression for the common ERC20 `totalSupply` pattern.

crates/doc/src/hir_ext.rs
```diff
@@ -0,0 +1,726 @@
...
+/// (and `int` / `int256`) compare equal during overload matching.
+pub fn parameter_type_strings(gcx: Gcx<'_>, params: &ParameterList<'_>) -> Vec<String> {
+    let sm = gcx.sess.source_map();
+    params
+        .vars
+        .iter()
+        .map(|v| {
+            let raw = sm.span_to_snippet(v.ty.span).unwrap_or_default();
+            let t = raw.split_whitespace().collect::<Vec<_>>().join(" ");
+            match t.as_str() {
```
figtracer: This canonicalizes only exact `uint`/`int` tokens, so aliases inside arrays like `uint[]` vs `uint256[]` or fixed arrays still miss the overload match and drop valid inherited docs. Consider normalizing aliases inside compound type strings, or deriving canonical ABI types from HIR, with a regression for `Base.foo(uint256[] calldata)` inherited by `Child.foo(uint[] calldata)`.

figtracer (commented):
Found two remaining doc-generation regressions from the latest fixes.

crates/doc/src/hir_ext.rs
mablr: https://github.com/foundry-rs/foundry/pull/14568/commits/b6238cd0fd06ed520ea7e606c3c8d55638ff6974

crates/doc/src/hir_ext.rs
mablr: https://github.com/foundry-rs/foundry/pull/14568/commits/0d2c79cd153033ce5a99415ee0d48f5db379ef8d

figtracer (dismissed):

crates/doc/src/render.rs
```diff
@@ -0,0 +1,1080 @@
...
+            Some(n) => n.as_str().to_string(),
+            None if !is_return => unnamed_iter.next().cloned().unwrap_or_else(|| "_".to_string()),
+            None => "&lt;none&gt;".to_string(),
+        };
+        let ty = format!("`{}`", ctx.snippet(var.ty.span).trim());
+        let desc = if is_return {
+            comments.returns.iter().find(|(n, _)| n == &name).map(|(_, d)| d.as_str()).unwrap_or("")
+        } else {
+            comments.params.iter().find(|(n, _)| n == &name).map(|(_, d)| d.as_str()).unwrap_or("")
+        };
```
mattsse: This drops `@return` text for unnamed return values. Solar parses `@return The computed value` as a return tag whose "name" is `The`, but this branch renders the table row as `&lt;none&gt;` and then only looks up docs by that rendered name. A positional fallback keeps the description, matching the legacy writer's behavior for unnamed returns.

```suggestion
    for (index, var) in params.iter().enumerate() {
        let name = match var.name {
            Some(n) => n.as_str().to_string(),
            None if !is_return => unnamed_iter.next().cloned().unwrap_or_else(|| "_".to_string()),
            None => "&lt;none&gt;".to_string(),
        };
        let ty = format!("`{}`", ctx.snippet(var.ty.span).trim());
        let desc = if is_return {
            let by_name =
                comments.returns.iter().find(|(n, _)| n == &name).map(|(_, d)| d.as_str());
            if by_name.is_none() && var.name.is_none() {
                comments.returns.get(index).map(|(_, d)| d.as_str()).unwrap_or("")
            } else {
                by_name.unwrap_or("")
            }
        } else {
            comments.params.iter().find(|(n, _)| n == &name).map(|(_, d)| d.as_str()).unwrap_or("")
        };
```

crates/doc/src/hir_ext.rs
```diff
@@ -0,0 +1,778 @@
...
+        if candidate_ident.chars().all(|c| c.is_alphanumeric() || c == '-' || c == '_')
+            && !candidate_part.is_empty()
+        {
+            (candidate_ident, Some(candidate_part))
+        } else {
+            (inner, None)
+        }
+    } else {
+        (inner, None)
+    };
```
mattsse, line 794: This still misses the legacy full-signature xref form `{xref-ERC721-_safeMint-address-uint256-}` because `rfind('-')` sees the trailing dash, treats the part as empty, and the later lookup links only the contract page. Split external xrefs at the first dash after the target identifier so the member/signature segment is preserved.

```suggestion
    // inner = "[xref-]Ident[-part]"
    let (raw_ident, raw_part) = if let Some(rest) = inner.strip_prefix("xref-") {
        if let Some(dash) = rest.find('-') {
            (&inner[..("xref-".len() + dash)], Some(&rest[dash + 1..]))
        } else {
            (inner, None)
        }
    } else if let Some(dash) = inner.find('-') {
        let candidate_ident = &inner[..dash];
        let candidate_part = &inner[dash + 1..];
        if candidate_ident.chars().all(|c| c.is_alphanumeric() || c == '_')
            && !candidate_part.is_empty()
        {
            (candidate_ident, Some(candidate_part))
        } else {
            (inner, None)
        }
    } else {
        (inner, None)
    };
```

mattsse (commented):
Found a few generated-output regressions while reviewing the solar/vocs rewrite. I left inline suggestions for the concrete fixes.

crates/doc/src/hir_ext.rs
```diff
@@ -0,0 +1,778 @@
...
+                    let page = resolve_page(candidates, current_page);
+                    let mut link = page_link(page, current_page);
+                    // Append the member anchor when the pattern is `{Type-member}`.
+                    // Sanitize to ASCII alphanumerics and `_` only, Solidity identifiers
+                    // never contain other characters, so this drops any injection attempt.
+                    if let Some(member) = part {
+                        let safe_member: String = member
+                            .chars()
+                            .filter(|c| c.is_ascii_alphanumeric() || *c == '_')
+                            .collect();
```
mattsse: Once the parser preserves full xref signatures, this should use only the member name before signature args for the generated heading anchor. Otherwise `{xref-ERC721-_safeMint-address-uint256-}` points to `#_safeMintaddressuint256`, which Vocs won't generate for the `### _safeMint` heading.

```suggestion
                    if let Some(member) = part {
                        let member = member.split('-').next().unwrap_or(member);
                        let safe_member: String = member
                            .chars()
                            .filter(|c| c.is_ascii_alphanumeric() || *c == '_')
                            .collect();
```

crates/doc/src/vocs.rs
```diff
@@ -0,0 +1,609 @@
...
+            fence_marker = "~~~";
+            out.push_str(line);
+        } else {
+            // Escape `{` and bare `<` (not already `&lt;` or a known entity).
+            for ch in line.chars() {
+                match ch {
+                    '{' => out.push_str(r"\{"),
+                    '<' => out.push_str("&lt;"),
+                    c => out.push(c),
+                }
```
mattsse, line 373: This escapes inside README inline code spans, so homepage snippets like `` `forge create <Contract>` `` or `` `{OWNER}` `` render as altered code. Keep the fenced-block handling, but also bypass escaping while inside inline backtick spans.

```suggestion
            let mut inline_code_ticks = 0usize;
            let mut pending_ticks = 0usize;
            for ch in line.chars() {
                if ch == '`' {
                    pending_ticks += 1;
                    out.push(ch);
                    continue;
                }

                if pending_ticks > 0 {
                    if inline_code_ticks == 0 {
                        inline_code_ticks = pending_ticks;
                    } else if inline_code_ticks == pending_ticks {
                        inline_code_ticks = 0;
                    }
                    pending_ticks = 0;
                }

                if inline_code_ticks > 0 {
                    out.push(ch);
                } else {
                    match ch {
                        '{' => out.push_str(r"\{"),
                        '<' => out.push_str("&lt;"),
                        c => out.push(c),
                    }
                }
            }
```

crates/doc/src/render.rs
mablr: https://github.com/foundry-rs/foundry/pull/14568/commits/7aab6fffca2a8b9820a96fe7be068c7ee0179f67

crates/doc/src/hir_ext.rs
mablr: https://github.com/foundry-rs/foundry/pull/14568/commits/7aab6fffca2a8b9820a96fe7be068c7ee0179f67

crates/doc/src/hir_ext.rs
mablr: https://github.com/foundry-rs/foundry/pull/14568/commits/7aab6fffca2a8b9820a96fe7be068c7ee0179f67

crates/doc/src/vocs.rs
mablr: https://github.com/foundry-rs/foundry/pull/14568/commits/7aab6fffca2a8b9820a96fe7be068c7ee0179f67

figtracer (dismissed):

figtracer (dismissed):

crates/doc/src/hir_ext.rs
```diff
@@ -0,0 +1,808 @@
...
+                    // never contain other characters, so this drops any injection attempt.
+                    if let Some(member) = part {
+                        let member = member.split('-').next().unwrap_or(member);
+                        let safe_member: String = member
+                            .chars()
+                            .filter(|c| c.is_ascii_alphanumeric() || *c == '_')
+                            .collect();
+                        if !safe_member.is_empty() {
+                            link.push('#');
+                            link.push_str(&safe_member);
```
stevencartavia, line 678: **Mixed-case cross-reference anchors are dead links**

`replace_inline_links` uses the raw identifier as the fragment, e.g. `{ECDSA-tryRecover}` → `…#tryRecover`. But vocs's `rehype-slug` (github-slugger) lowercases heading ids, so `### tryRecover` becomes `id="tryrecover"`. Fragment matching is case-sensitive → dead link. Confirmed against real rehype-slug output: ids `['tryrecover','tryrecover-1']` vs link `#tryRecover`. Any uppercase member (`balanceOf`, `_safeMint`) is affected; `full_signature_xref_links_member_anchor` also asserts the wrong `#_safeMint`.

Fix: slugify the fragment like vocs does (lowercase), and update the test.

crates/doc/src/hir_ext.rs
```diff
@@ -0,0 +1,808 @@
...
+                };
+
+                if let Some(candidates) = name_to_page.get(lookup_name) {
+                    let page = resolve_page(candidates, current_page);
+                    let mut link = page_link(page, current_page);
+                    // Append the member anchor when the pattern is `{Type-member}`.
+                    // Sanitize to ASCII alphanumerics and `_` only, Solidity identifiers
+                    // never contain other characters, so this drops any injection attempt.
+                    if let Some(member) = part {
+                        let member = member.split('-').next().unwrap_or(member);
```
stevencartavia: **Overload xrefs can't reach anything but the first overload**

The signature suffix is captured but discarded for anchors (`member.split('-').next()`), so all overloads link to `#<name>`. But duplicate `### name` headings get slugged to `name`, `name-1`, `name-2`, so overloads after the first can't be targeted.

Fix together with the casing issue: rendering and linking should share one anchor algorithm, keeping enough signature info to hit the right overload.

crates/doc/src/vocs.rs
```diff
@@ -0,0 +1,633 @@
...
+  }
+}
+"#
+}
+
+// ── helpers ───────────────────────────────────────────────────────────────────
+
+fn json_str(s: &str) -> String {
+    // Escape double quotes and wrap.
+    format!("'{}'", s.replace('\'', "\\'"))
```
stevencartavia: **A valid `[doc] title` can produce an invalid `vocs.config.ts`**

`json_str` only escapes single quotes (`format!("'{}'", s.replace('\'', "\\'"))`). A title ending in a backslash:
```toml
[doc]
title = "Acme\\"
```
generates `title: 'Acme\',` — the escaped quote leaves an unterminated string, breaking `vocs dev`/`build`. Same path feeds sidebar names.

Fix: use `serde_json::to_string(s)` (valid TS too).

crates/doc/src/builder.rs
```diff
@@ -57,419 +66,317 @@ impl DocBuilder {
...
-                // If no path specified, use repository URL as-is.
-                repo.clone()
+                // No manifest: fall back to pruning only the legacy `src/` subtree.
+                let src_dir = pages_dir.join("src");
+                let kept: HashSet<PathBuf> = all_rel
+                    .iter()
+                    .filter_map(|p| p.strip_prefix("src").ok().map(|s| s.to_path_buf()))
+                    .collect();
+                if src_dir.exists() {
+                    prune_dir(&src_dir, &src_dir, &kept)?;
```
stevencartavia: **Fallback prune can delete user-authored pages**

When `.forge-doc-manifest` is absent, this branch runs `prune_dir` over `pages/src/` and deletes every regular `.mdx` file not generated in the current run — with no way to distinguish user files from generated ones.

Reproduced: with no manifest, a hand-written `docs/src/pages/src/overview.mdx` is silently deleted on `forge doc` (files outside `src/` survive this fallback prune). The no-manifest case happens on first run, after a manual delete, or `git clean` if the manifest is untracked (the generated `.gitignore` doesn't cover it). When the manifest exists, unlisted user pages are correctly left alone — so the loss is specific to this fallback.

Fix: when the manifest is absent, prune nothing and just write the manifest for future runs. Filename-prefix filtering is weaker since users can author files with the same prefixes; the manifest is the only reliable ownership boundary.

crates/doc/src/hir_ext.rs
mablr: f257915

crates/doc/src/hir_ext.rs
mablr: d13c66d

crates/doc/src/vocs.rs
mablr: 0b54101

crates/doc/src/builder.rs
mablr: e400648

figtracer (approved):

crates/doc/src/hir_ext.rs
```diff
@@ -0,0 +1,849 @@
...
+///
+/// Types are normalized to their canonical ABI form so that `uint` and `uint256`
+/// (and `int` / `int256`) compare equal during overload matching.
+pub fn parameter_type_strings(gcx: Gcx<'_>, params: &ParameterList<'_>) -> Vec<String> {
+    let sm = gcx.sess.source_map();
+    params
+        .vars
+        .iter()
+        .map(|v| {
+            let raw = sm.span_to_snippet(v.ty.span).unwrap_or_default();
```
grandizzy, line 472: Proposed: open a tracking issue for "derive canonical HIR/ABI parameter types for `@inheritdoc` overload matching", and land this characterization test now so the gap is recorded and self-corrects when fixed:

```rust
// Known limitation: `@inheritdoc` overload matching compares the raw *source text*
// of each parameter's type (only `uint`/`int` aliases are normalized, see
// `normalize_sol_type`). When a valid, compiling override spells a parameter type
// differently from the base declaration — here base `I.Status` vs override
// `Status` — the signatures fail to match and the inherited NatSpec is silently
// dropped. The `uint256` overload (verbatim match) inherits correctly, isolating
// the cause to type spelling rather than `@inheritdoc` resolution in general.
//
// CHARACTERIZATION test: it locks in the *current* (incomplete) behavior. The
// proper fix is to derive canonical HIR/ABI parameter types instead of source
// snippets. When that lands, the `!contains("Sets the account status")` assertion
// below will start failing — flip it to `contains(...)` at that point.
forgetest_init!(inheritdoc_overload_drops_docs_on_type_spelling, |prj, cmd| {
    prj.add_source(
        "I.sol",
        r#"
pragma solidity ^0.8.0;

interface I {
    enum Status { Inactive, Active }

    /// @notice Sets the account status.
    /// @param s the new status
    function configure(I.Status s) external;

    /// @notice Configures by raw id.
    /// @param id the raw id
    function configure(uint256 id) external;
}
"#,
    );

    prj.add_source(
        "C.sol",
        r#"
pragma solidity ^0.8.0;

import "./I.sol";

contract C is I {
    /// @inheritdoc I
    function configure(Status s) external override {}

    /// @inheritdoc I
    function configure(uint256 id) external override {}
}
"#,
    );

    cmd.args(["doc"]).assert_success();

    let content =
        std::fs::read_to_string(prj.root().join("docs/src/pages/src/contract.C.mdx")).unwrap();

    // Control: the `uint256` overload matches verbatim and inherits correctly.
    assert!(
        content.contains("Configures by raw id"),
        "uint256 overload should inherit its notice (control case), found:\n{content}"
    );

    // Known limitation (see header comment): the `Status` overload's inherited
    // notice is currently dropped because `I.Status` != `Status` as raw text.
    // FIXME(<issue>): once canonical HIR types are used, this should be
    // `assert!(content.contains("Sets the account status"), ...)`.
    assert!(
        !content.contains("Sets the account status"),
        "EXPECTED-FAIL breadcrumb: `@inheritdoc` overload type matching was improved \
         to canonicalize types — update this test to assert the notice IS inherited.\n{content}"
    );
});
```

grandizzy (approved):

PR #14569, opened 2026-05-04
fix(anvil): return access list for createAccessList reverts

mmv08:
## Motivation

`eth_createAccessList` should preserve the generated access list when simulated execution fails. The Ethereum Execution API fixture [`create-al-abi-revert.io`](https://github.com/ethereum/execution-apis/blob/8d6b784cc57047ef10e4044ec2efc1c60950dd7f/tests/eth_createAccessList/create-al-abi-revert.io) expects a successful JSON-RPC result with `accessList`, `gasUsed`, and an in-result `error: "execution reverted"`.

Anvil previously converted EVM failures into top-level JSON-RPC errors, so callers could not inspect the accessed slots from the failed simulation.

## Solution

Return a successful `AccessListResult` for EVM execution failures. Anvil now builds the access list from the traced execution, re-executes with that access list applied to compute `gasUsed`, and reports any VM failure through the result object's optional `error` field.

The `error` value is derived from revm's `SuccessOrHalt::<HaltReason>` mapping. Successful exits omit `error`, reverts report `"execution reverted"`, and halt reasons use revm's `Display` implementation.

The access-list integration test now covers an Execution API-style revert case that reads a storage slot before reverting, then asserts that `eth_createAccessList` still returns the generated slot, nonzero gas, and `"execution reverted"`.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

Tested with:

```console
cargo +nightly fmt --check
cargo +1.94.0 check -p anvil
```


crates/anvil/src/eth/api.rs
```diff
@@ -3669,6 +3675,18 @@ fn ensure_return_ok(exit: InstructionResult, out: &Option<Output>) -> Result<Byt
...
 
+/// Maps an EVM exit code to the optional `error` string reported in
+/// `eth_createAccessList` results. Mirrors geth/reth behavior of surfacing all
+/// EVM failures alongside the traced access list, while reserving
+/// `"execution reverted"` for the actual `REVERT` opcode.
+fn execution_error(exit: InstructionResult) -> Option<String> {
+    match exit {
+        return_ok!() => None,
+        InstructionResult::Revert => Some("execution reverted".to_string()),
+        reason => Some(format!("{reason:?}")),
```
mablr: If you aim parity with Geth, you may want to pay attention to the output formatting. Here, raw debug Rust enum names like `OutOfGas`, `CallTooDeep`... will be returned. I think Geth/Reth return human strings like "out of gas", "max call depth exceeded", etc.

mablr (commented):
sgtm

crates/anvil/src/eth/api.rs
mmv08: That's a good point, thanks.

If you think it makes sense, I'm happy to either include this in this pull request or address it in a follow-up pull request to keep this one small and easy to review.

I think preserving the error behavior is of higher importance than returning human strings

mablr: Let's do it here 😃

mmv08: done!

crates/anvil/src/eth/api.rs
```diff
@@ -3669,6 +3675,54 @@ fn ensure_return_ok(exit: InstructionResult, out: &Option<Output>) -> Result<Byt
...
+    execution_error_message(exit).map(str::to_owned)
+}
+
+fn execution_error_message(exit: InstructionResult) -> Option<&'static str> {
+    match exit {
+        return_ok!() => None,
+        InstructionResult::Revert => Some("execution reverted"),
+        InstructionResult::CallTooDeep => Some("max call depth exceeded"),
+        InstructionResult::OutOfFunds => Some("insufficient balance for transfer"),
+        InstructionResult::CreateInitCodeStartingEF00 => {
```
mablr: I thought there was a proper `Display` impl ? (haven't checked)

If this is not the case i think it's not clean to maintain it in Foundry and should be reverted. Sorry for the round-trip ^^

mablr (changes requested):

crates/anvil/src/eth/api.rs
mmv08: Updated to use revm’s existing error display, thanks!

mablr: noice ty

mablr (approved):
Makes sense.

figtracer (approved):
lgtm

stevencartavia (approved):

Issue #12322, opened 2025-10-28
bug(fmt): don't break var assignments when callee fits


0xrusowsky:
### Component

Forge

### Have you ensured that all of these are up to date?

- [ ] Foundry
- [ ] Foundryup

### Operating System

None

### Describe the bug

user report:
```
107      |-        vars.previousTotalDebt = getDescaledAmount(
108      |-            flow.getSnapshotDebtScaled(streamId), flow.getTokenDecimals(streamId)
109      |-        ) + vars.previousOngoingDebtScaled;
    107  |+        vars.previousTotalDebt =
    108  |+            getDescaledAmount(flow.getSnapshotDebtScaled(streamId), flow.getTokenDecimals(streamId))
    109  |+                + vars.previousOngoingDebtScaled;
```

PR #14574, opened 2026-05-04
fix(fmt): keep callee inline on binary rhs force-break

AissataF:
`print_assign_rhs` had two break paths: `force_break` always hard-broke, while the lhs-overflow path checked `get_callee_head_size` (per #12322) and kept the call inline when it fit. Merge the conditions so both paths share the callee-fit check, and only apply the `ibox(-self.ind)` de-indent when the call's args are expected to break. Updates the `VariableAssignment` test data with a new regression case and aligns `OperatorExpressions/120` and `ReprosCalls/110` snapshots, whose old hard-break output came from the same gap.


mablr (commented):
Given the test cases I'm unsure this is a positive change for readability.

AissataF:
> Given the test cases I'm unsure this is a positive change for readability.

Good catch, addressed in the latest commit. I gated the new inline path on lhs being a Call so homogeneous binary chains keep the original equals then break behavior, and reverted the ReprosCalls 110 snapshot to its master form.

PR #14576, opened 2026-05-04
fix: compact fish completions

o-az:
## Motivation

I was investigating why my shell was slow so on the things I examined were foundry tools completions. I'm on `fish` and completions size for `cast` was 1.59 MB.

I looked into the CLI impl and saw potential for substantial reduction in completions size.

Problem: completions for tools are too big.
Size reduction strategy: kept `clap_complete` as the source of truth then post-processes the generated output for `fish`, `bash`, and `zsh` to factor repeated generated blocks into helper functions.

## Solution

| Tool | Fish | Bash | Zsh |
|---|---:|---:|---:|
| cast | 1.59 MB -> 1.00 MB | 0.56 MB -> 0.40 MB | 1.14 MB -> 0.45 MB |
| forge | 0.41 MB -> 0.31 MB | 0.27 MB -> 0.23 MB | 0.34 MB -> 0.21 MB |
| anvil | 0.014 MB -> 0.012 MB | 0.014 MB -> 0.014 MB | 0.014 MB -> 0.013 MB |

I put more details in this gist as well: https://gists.sh/o-az/b937e947a8db20d2b6f634343ad27c99?noheader&nofooter

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cli/src/clap.rs
+++ b/crates/cli/src/clap.rs
@@ -40,21 +40,27 @@ impl Generator for Shell {
     fn generate(&self, cmd: &clap::Command, buf: &mut dyn std::io::Write) {
         match self {
             Self::ClapCompleteShell(ClapCompleteShell::Bash) => {
-                let mut completion = Vec::new();
-                ClapCompleteShell::Bash.generate(cmd, &mut completion);
-                let completion = compact_bash_completion(cmd.get_name(), completion);
-                buf.write_all(completion.as_bytes()).expect("failed to write completion file");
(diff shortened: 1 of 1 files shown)


zerosnacks:
Hey @o-az, thanks for your PR!

I would prefer we not introduce this logic into Foundry directly but rather would recommend to attempt to land any optimizations in https://github.com/clap-rs/clap/tree/master/clap_complete directly that Foundry and others can then rely upon.

PR #14580, opened 2026-05-05
fix(doc): preserve NatSpec blank lines in forge doc

### LINKED ISSUE #4539: feat(doc): add two new lines between @notice and @dev
### Component

Forge

`forge doc` adds only one new line between the `@notice` and the `@dev` NatSpec tags. It would be better if it instead added two lines since in Markdown paragraphs are better delineated this way. For example, Prettier auto-formats one end line to an empty space.

Take this example:

```solidity
/// @notice Transfers the admin of the contract to a new address.
///
/// Notes:
/// - Does not revert if the admin is the same.
/// - This function can potentially leave the contract without an admin, thereby removing any
/// functionality that is only available to the admin.
///
/// Requirements:
/// - The caller must be the current contract admin.
///
/// @param newAdmin The address of the new admin.
function transferAdmin(address newAdmin) external;
```

The Markdown generated by `forge doc` is the following:

```md
### transferAdmin

Transfers the admin of the contract to a new address.
Notes:
- Does not revert if the admin is the same.
- This function can potentially leave the contract without an admin, thereby removing any
functionality that is only available to the admin.
Requirements:
- The caller must be the current contract admin.
```

I would like `forge doc` to add an extra end line between "to a new address." and "Notes:".

### Additional context

_No response_

mookim-eth:
Closes #4539

## Motivation

`forge doc` currently drops empty lines inside NatSpec tag bodies, so Markdown paragraph breaks such as the blank line between a notice summary and a `Notes:` section are lost in generated documentation.

## Solution

Preserve empty doc comment lines while parsing NatSpec continuations, including empty `///` symbols that `str::lines()` would otherwise skip. Added parser coverage for both direct comment parsing and Solidity source parsing to ensure blank lines are retained before `@param` tags are separated.

## Tests

- `cargo +1.94.0 test -p forge-doc`
- `git diff --check`

## AI Assistance Disclosure

This PR was prepared with assistance from ChatGPT/Codex for issue investigation, implementation, and PR text drafting.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/doc/src/parser/comment.rs
+++ b/crates/doc/src/parser/comment.rs
@@ -173,10 +173,18 @@ impl Comments {
             }
         };
 
+        fn push_continuation(value: &mut String, line: &str) {
+            if value.is_empty() {
+                value.push_str(line);
+            } else {
+                value.push('\n');
+                value.push_str(line);
+            }
+        }
+
         for raw_line in lines {
             let raw = raw_line.as_ref();
-            // For block comments, process each line individually
-            for line in raw.lines() {
+            let mut process_line = |line: &str| {
                 let trimmed = line.trim().trim_start_matches('*').trim();
 
                 if let Some(rest) = trimmed.strip_prefix('@') {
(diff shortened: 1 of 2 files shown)


mablr (commented):
Hi @mookim-eth!

fyi i'm conducting a major refactor on `forge doc` https://github.com/foundry-rs/foundry/pull/14568.

PR #14583, opened 2026-05-05
fix(forge): allow verify-bytecode to run without external explorer dependencies

### LINKED ISSUE #13479: forge verify-bytecode command can not verify a contract without external services
### Component

Forge

- [x] Foundry
- [x] Foundryup

1.5.1-stable

0.3.3

forge verify-bytecode

### Operating System

macOS (Apple Silicon)

#### Expectation

`forge verify-bytecode` command compiles a specified contract locally, then compares locally generated bytecode with the onchain bytecode.

#### Actual behaviour

Currently, if one tries to verify a contract deployed to a blockchain, but not verified in any block scanners, the `verify-bytecode` command will fail:
- Either with `Error: Error fetching creation data from verifier-url: InvalidApiKey`
- Or, if API key is provided, `Error: Contract source code not verified: 0x...`

This is confirmed by a quick look [into the source code](https://github.com/foundry-rs/foundry/blob/94d1caab9fef5d99107f1a5e6cf7313eb23e2a4c/crates/verify/src/bytecode.rs#L177-L178): at least constructor args, contract name, compilation metadata are being unconditionally fetched from a block scanner, expecting it to be there. While I understand that it might be helpful for debugging (to e.g. suggest that some compilation settings are incorrectly set locally) or for identifying additional problems (locally compiled bytecode match external one, but verified contract _name_ is different), it's not very helpful that those additional actions can't be disabled/ignored.

#### Proposed mitigation

Ideally, make those additional checks optional (warning-style) or add new `--ignore-...` flag to skip those checks. The goal is to be able to verify bytecode without any external services, based only on a locally available configuration.

ParthSinghPS:
Fixes #13479 

## Description

This PR fixes an issue where the `forge verify-bytecode` command unconditionally required the target contract to be verified on a block explorer (via Etherscan/Blockscout API) before comparing the bytecode. This prevented users from relying solely on an RPC node to verify locally compiled bytecode against the onchain version.

With these changes, if fetching the contract's source code from the explorer fails, the command degrades gracefully:
- It skips the contract name check that relied on explorer metadata.
- It falls back to the user-provided config's `evm_version`.
- It defaults to empty constructor arguments if none were provided by the user (rather than failing outright).
- It gracefully skips the settings mismatch diagnostic messaging, since there is no explorer metadata to compare against.

## Manual Testing

1. Deployed a small contract to a testnet (e.g., Sepolia) via `forge create` without verifying.
2. Ran `forge verify-bytecode <Address> src/MyContract.sol:MyContract --rpc-url <RPC>`
3. Earlier, it would error out with `Contract source code not verified`. Now, it prints a warning and proceeds to match the bytecode using the RPC.

### DIFF
--- a/crates/verify/src/bytecode.rs
+++ b/crates/verify/src/bytecode.rs
@@ -171,17 +171,30 @@ impl VerifyBytecodeArgs {
 
         trace!(maybe_predeploy = ?maybe_predeploy);
 
-        // Get the constructor args using `source_code` endpoint.
-        let source_code = etherscan.contract_source_code(self.address).await?;
+        // Try to get the source code from the block explorer.
+        // This may fail if the contract is not verified on any block scanner.
+        let source_code = match etherscan.contract_source_code(self.address).await {
+            Ok(source_code) => Some(source_code),
+            Err(e) => {
+                if !shell::is_json() {
+                    sh_warn!(
+                        "Failed to fetch source code from block explorer: {e}. Proceeding without explorer verification."
+                    )?;
+                }
+                None
+            }
+        };
 
-        // Check if the contract name matches.
-        let name = source_code.items.first().map(|item| item.contract_name.clone());
-        if name.as_ref() != Some(&self.contract.name) {
-            eyre::bail!("Contract name mismatch");
+        // Check if the contract name matches (only if source code is available).
+        if let Some(ref source_code) = source_code {
+            let name = source_code.items.first().map(|item| item.contract_name.to_owned());
+            if name.as_ref() != Some(&self.contract.name) {
+                eyre::bail!("Contract name mismatch");
+            }
         }
 
-        // Obtain Etherscan compilation metadata.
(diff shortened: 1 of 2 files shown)


figtracer (commented):
let's add a regression test for this

ParthSinghPS:
i have added two tests named `print_result_succeeds_without_etherscan_metadata_on_match` and `print_result_succeeds_without_etherscan_metadata_on_mismatch` 

also , I was unable to run `cargo test` locally because `master` currently references a dangling git revision for `alloy-op-evm`.
this was the error I got:
```text
error: failed to load source for dependency `alloy-op-evm`
Caused by:
  unable to update https://github.com/ethereum-optimism/optimism?rev=e3b59e76588f99db17205f5601e45a5b00f0bfbb
Caused by:
  object not found - no match for id (e3b59e76588f99db17205f5601e45a5b00f0bfbb)

PR #14589, opened 2026-05-05
feat(lint): add tx-origin detector

vikions:
## Summary

- Adds a `tx-origin` medium-severity lint.
- Reports `tx.origin` when it is used in authorization-like predicates such as `require`, `assert`, `if`, `while`, and `for` conditions.
- Leaves plain reads like `return tx.origin` unreported to avoid broad false positives.
- Adds lint docs and UI test coverage.

Partially addresses #14381.

## Testing

- `cargo fmt --package forge-lint`
- `cargo fmt --check --package forge-lint`
- `git diff --check`
- `forge build --root /tmp/foundry-tx-origin`

### DIFF
--- a/crates/lint/src/sol/med/tx_origin.rs
+++ b/crates/lint/src/sol/med/tx_origin.rs
@@ -4,7 +4,7 @@ use crate::{
     sol::{Severity, SolLint},
 };
 use solar::{
-    ast::{Expr, ExprKind, Stmt, StmtKind},
+    ast::{Expr, ExprKind, IndexKind, Stmt, StmtKind},
     interface::SpannedOption,
 };
 
@@ -54,6 +54,17 @@ fn contains_tx_origin(expr: &Expr<'_>) -> bool {
     match &expr.kind {
         ExprKind::Unary(_, inner) => contains_tx_origin(inner),
         ExprKind::Binary(lhs, _, rhs) => contains_tx_origin(lhs) || contains_tx_origin(rhs),
+        ExprKind::Index(base, index) => {
+            contains_tx_origin(base)
+                || match index {
+                    IndexKind::Index(Some(index)) => contains_tx_origin(index),
+                    IndexKind::Range(start, end) => {
+                        start.as_ref().is_some_and(|start| contains_tx_origin(start))
+                            || end.as_ref().is_some_and(|end| contains_tx_origin(end))
+                    }
+                    _ => false,
+                }
+        }
         ExprKind::Tuple(elems) => elems.iter().any(|elem| {
             if let SpannedOption::Some(inner) = elem.as_ref() {
                 contains_tx_origin(inner)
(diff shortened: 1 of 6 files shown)


crates/lint/testdata/TxOrigin.sol
```diff
@@ -0,0 +1,38 @@
+//@compile-flags: --only-lint tx-origin
```
mablr, line 1: Coverage could be improved, for `StmtKind::While`, `StmtKind::For`, and `StmtKind::DoWhile`.

crates/lint/src/sol/med/tx_origin.rs
```diff
@@ -0,0 +1,125 @@
...
+        ExprKind::Ternary(cond, then_expr, else_expr) => {
+            emit_tx_origin_reads(ctx, cond);
+            emit_tx_origin_reads(ctx, then_expr);
+            emit_tx_origin_reads(ctx, else_expr);
+        }
+        ExprKind::Tuple(elems) => {
+            for elem in elems.iter() {
+                if let SpannedOption::Some(elem) = elem.as_ref() {
+                    emit_tx_origin_reads(ctx, elem);
+                }
```
mablr: `emit_tx_origin_reads` catches nearly every `ExprKind` variant, including `Delete`, `Payable`, ... These are semantically impossible inside a guard condition. Compare to:
https://github.com/foundry-rs/foundry/blob/fa5429abb300bc27cee58efb6855cf8befa7b2ab/crates/lint/src/sol/low/block_timestamp.rs#L49-L70
which only handles the expr kinds that realistically appear in comparisons.

mablr (dismissed):
Lgtm, thanks.

Pending doc update https://github.com/foundry-rs/book/pull/1816

figtracer (commented):
to achieve closer parity with Slither, we should also tackle these 
```solidity
require(allowed[tx.origin], "not allowed");
require(allowed[tx.origin] == true, "not allowed");
require(tx.origin == owner ? true : false, "not owner");
```

ref: https://github.com/crytic/slither/blob/master/slither/detectors/statements/tx_origin.py

vikions:
> to achieve closer parity with Slither, we should also tackle these
> 
> ```solidity
> require(allowed[tx.origin], "not allowed");
> require(allowed[tx.origin] == true, "not allowed");
> require(tx.origin == owner ? true : false, "not owner");
> ```
> 
> ref: https://github.com/crytic/slither/blob/master/slither/detectors/statements/tx_origin.py

Addressed in the latest commit: the lint now traverses index expressions and ternary predicates, covering the three Slither-parity cases you listed. I also added UI coverage for all three examples.

zerosnacks:
Hi @vikions, there appears to be a test failure in the PR; hard to read from the CI logs what it exactly is in a glance

mablr (commented):
@vikions this should fix:
```
cargo test --test ui -- --bless
```

vikions:
> @vikions this should fix:
> 
> ```
> cargo test --test ui -- --bless
> ```

Thanks, fixed by blessing the TxOrigin UI snapshot. `cargo test -p forge --test ui -- TxOrigin` passes locally now.

zerosnacks (approved):

figtracer (approved):

mablr (approved):

PR #14592, opened 2026-05-05
fix(anvil): classify EVM halts as transaction rejections

mmv08:
## Motivation

Anvil currently maps non-`REVERT` EVM execution halts, such as `InvalidFEOpcode`, to a JSON-RPC internal error (`-32603`). These are execution failures caused by the simulated transaction, not JSON-RPC server faults.

Other clients do not all use the same JSON-RPC code for this case, but they do distinguish these halts from `REVERT` and do not classify them as internal server errors:

- Geth handles `REVERT` through a dedicated revert error path with JSON-RPC code `3` and revert data. Other VM errors, such as invalid opcode, are returned through the generic VM error path and surface with the default JSON-RPC error code (`-32000`), without revert data.
- Reth similarly maps `REVERT` to its dedicated `RevertError` / execution-error path, while halt reasons such as invalid opcode go through `EvmHalt` and are surfaced as transaction rejections (`-32003`).

## Solution

Map `BlockchainError::EvmError(_)` to `ErrorCode::TransactionRejected` (`-32003`) instead of `InternalError` (`-32603`). This follows Reth's treatment of non-`REVERT` EVM halts as transaction/execution failures while keeping Anvil's existing `REVERT` path unchanged.

Added a regression test that deploys runtime bytecode `0xfe` and verifies both `eth_call` and `eth_estimateGas` return `-32003`, omit revert data, preserve the `InvalidFEOpcode` message, and do not label the halt as `execution reverted`.

This is marked as a breaking change because callers that explicitly match Anvil's current `-32603` JSON-RPC code for non-`REVERT` EVM halts may need to update that handling.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/anvil/src/eth/error/mod.rs
+++ b/crates/anvil/src/eth/error/mod.rs
@@ -524,9 +524,13 @@ impl<T: Serialize> ToRpcResponseResult for Result<T> {
                         err => RpcError::internal_error_with(format!("Fork Error: {err:?}")),
                     }
                 }
-                err @ BlockchainError::EvmError(_) => {
-                    RpcError::internal_error_with(err.to_string())
-                }
+                err @ BlockchainError::EvmError(_) => RpcError {
+                    // VM halts are execution failures, not JSON-RPC server faults. REVERT has a
+                    // dedicated code/data path above; other halts, such as invalid opcode, do not.
+                    code: ErrorCode::TransactionRejected,
+                    message: err.to_string().into(),
+                    data: None,
+                },
                 err @ BlockchainError::EvmOverrideError(_) => {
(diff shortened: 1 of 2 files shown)


DaniPopes (approved):

figtracer (approved):

stevencartavia (approved):

mmv08:
It seems most CI failures are related to forks. I'm not sure if it's related to me contributing outside the organisation, but I don't think it's related to my changes.

Please tag me if there is anything that needs my attention 🙏

mablr (approved):

mattsse (approved):

PR #14595, opened 2026-05-05
fix(cli): resolve Tempo expires once

figtracer:
## Summary
- add `TempoOpts::resolve_expires()` to materialize `--tempo.expires` into the expiring nonce fields once
- use the resolved value in `cast send` and `cast mktx` so the printed expiry matches the submitted transaction


stevencartavia (approved):

mablr (approved):
Nice #14594 follow-up 👍

PR #14598, opened 2026-05-05
fix(forge): bind generates bindings for all same-named contracts

gutonosa:
## Summary
            
  `forge bind` previously deduplicated artifacts using a `HashSet<String>` keyed on contract name, silently dropping all but the first occurrence whenever the same contract name appeared in different source paths (e.g.                                                                                                                                                                   
  `src/Counter.sol`, `src/a/Counter.sol`, `src/b/Counter.sol`). The resulting bindings crate was missing entries with no warning — a real correctness issue

keeps every artifact and disambiguates the *Rust* binding name only, preserving the Solidity contract name used by the `sol!` macro.

  ## Changes

  - `SolMacroGen` now carries a separate `binding_name` (Rust module/file name)
    in addition to `name` (Solidity contract identifier). All downstream code
    that names files and `pub mod` declarations switches to `binding_name`;
    `sol!` expansion keeps using `name`.
  - `forge bind` groups artifacts by contract name. Single-occurrence contracts
    keep the previous behavior (`binding_name == name`, fully backward
    compatible). Multi-occurrence contracts derive a unique `binding_name` from
    the artifact path's disambiguating parent directories
    (`out/a/Counter.sol/Counter.json` → `a_Counter`), with a numeric suffix as a
    last-resort tiebreaker.

under the help with claude opus-4.7

### DIFF
--- a/crates/forge/src/cmd/bind.rs
+++ b/crates/forge/src/cmd/bind.rs
@@ -1,4 +1,3 @@
-use alloy_primitives::map::HashSet;
 use clap::{Parser, ValueHint};
 use eyre::Result;
 use forge_sol_macro_gen::{MultiSolMacroGen, SolMacroGen};
@@ -7,6 +6,7 @@ use foundry_common::{compile::ProjectCompiler, fs::json_files};
 use foundry_config::impl_figment_convert;
 use regex::Regex;
 use std::{
+    collections::BTreeMap,
     fs,
     path::{Path, PathBuf},
 };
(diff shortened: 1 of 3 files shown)


mablr (commented):
Thanks for your contribution.

FYI this new lint will trigger a warning on reused contract names https://github.com/foundry-rs/foundry/pull/14623

So I'm unsure we need this, pending a another review to be sure.

PR #14607, opened 2026-05-06
feat(lint): add `var-read-using-this` gas lint

stevencartavia:
Adds the `var-read-using-this` gas lint, which flags `this.X(...)` calls that read state through the contract's own external interface — they cost an unnecessary STATICCALL when a direct read would do.

- Covers public state-variable getters (simple, mapping, array, struct) and any `public/external` `view/pure` function reachable on this.
- Machine-applicable fix for simple state-variable getters (`this.foo()` -> `foo`) and best-effort fix for mapping/array getters (`this.m(k)` -> `m[k]`).


zerosnacks:
Merged https://github.com/foundry-rs/book/pull/1818 in the book :)

mablr (commented):
Let's move import in `crates/lint/src/sol/gas/var_read_using_this.rs` at L100 to the top

mablr (approved):
sgtm

zerosnacks (approved):
lgtm 👍

PR #14610, opened 2026-05-06
fix(forge): `--fuzz-seed` parameter is not effective in `forge coverage`

wtdcode:
## Motivation

The `forge coverage --fuzz-seed ...` is ignored. Trivial patch.

## Solution

Check if the seed is given. If the seed is given, use the seed, otherwise fallback to previous behavior using static seeds. No break changes are introduced for this patch.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/forge/src/cmd/coverage.rs
+++ b/crates/forge/src/cmd/coverage.rs
@@ -87,8 +87,11 @@ impl CoverageArgs {
             config = self.load_config()?;
         }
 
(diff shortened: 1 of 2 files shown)


figtracer (approved):
lg, thanks

zerosnacks (approved):
Makes sense, thanks!

wtdcode:
CI failure seems unrelated.

PR #14614, opened 2026-05-06
fix(anvil): cancel full pending transaction listener

solanaXpeter:
## Motivation

`eth_subscribe` with `newPendingTransactions` and `true` spawns a task that waits on ready transaction hashes before forwarding full transactions. When the websocket subscription is dropped while no new transaction arrives, that task can stay parked and keep the pool listener alive until a later notification.

## Solution

Watch the downstream full-transaction receiver's close signal in the forwarding task, close the ready hash listener on exit, and prune closed ready listeners from the pool immediately. Added a unit test for pruning closed ready listeners and verified the existing full-pending subscription path.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -3380,13 +3380,27 @@ impl EthApi<FoundryNetwork> {
         let this = self.clone();
 
         tokio::spawn(async move {
-            while let Some(hash) = hashes.next().await {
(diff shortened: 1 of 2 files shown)


mattsse (commented):
we can keep this as is

PR #14615, opened 2026-05-06
fix(cheatcodes): enforce `expectRevert` reverter address for CREATE frames

### LINKED ISSUE #14613: `address` field in `expectRevert` is ignored for reverts inside a `create`
### Component

Forge

- [x] Foundry
- [ ] Foundryup

forge Version: 1.6.0-v1.7.0 Commit SHA: f83bad912a9dba7bf0371def1e70bb1896048356 Build Timestamp: 2026-04-28T14:07:10.568929921Z (1777385230) Build Profile: dist

not used

forge test

### Operating System

None

### Description

When a top-level CREATE's constructor reverts directly, `vm.expectRevert(address reverter, ...)` does not actually validate the reverter address — the test passes regardless of which address is supplied.

This affects every overload that takes a `reverter` argument:
- `expectRevert(address)`
- `expectRevert(bytes4, address)`
- `expectRevert(bytes, address)`
- `expectPartialRevert(bytes4, address)`
- the `count`-bearing variants

It also affects nested-CREATE chains where the reverting frame at the bottom is itself a CREATE.

### Reproduction

```solidity
pragma solidity ^0.8.18;

contract DirectReverter {
    constructor() { revert("ctor revert"); }
}

contract NestedCreateReverter {
    constructor() { new DirectReverter(); }
}

contract MethodReverter {
    function boom() external pure { revert("method revert"); }
}

contract Inner {
    function boom() external pure { revert("from inner"); }
}

contract ChildCallReverter {
    constructor(address inner) { Inner(inner).boom(); }
}

contract IssueRepro is Test {
    // BUG: passes — 0xdead is clearly not the reverter.
    function test_directCtorRevert_wrongAddress() public {
        vm.expectRevert(address(0xdead));
        new DirectReverter();
    }

    // BUG: passes — addr argument is ignored even when bytes are also given.
    function test_directCtorRevert_bytesAndAddress() public {
        vm.expectRevert(bytes("ctor revert"), address(0xdead));
        new DirectReverter();
    }

    // BUG: passes — outer CREATE wraps inner CREATE that directly reverts.
    function test_nestedCreate() public {
        vm.expectRevert(address(0xdead));
        new NestedCreateReverter();
    }

    // CONTROL: correctly fails with "Reverter != expected reverter".
    function test_callRevert_wrongAddress() public {
        MethodReverter m = new MethodReverter();
        vm.expectRevert(address(0xdead));
        m.boom();
    }

    // CONTROL: correctly fails — constructor calls into Inner which reverts.
    // call_end populates reverted_by with Inner's address before the surrounding
    // create_end runs, so the reverter check fires as expected.
    function test_ctorChildCall_wrongAddress() public {
        Inner inner = new Inner();
        vm.expectRevert(address(0xdead));
        new ChildCallReverter(address(inner));
    }

    // CONTROL: correctly passes — reverter address matches Inner's address.
    function test_ctorChildCall_correctAddress() public {
        Inner inner = new Inner();
        vm.expectRevert(address(inner));
        new ChildCallReverter(address(inner));
    }
}
```

Run:
```
forge test
```

```
Ran 6 tests for test/Counter.t.sol:IssueRepro
[PASS] test_BUG_directCtorRevert_bytesAndAddress() (gas: 36145)
[PASS] test_BUG_directCtorRevert_wrongAddress() (gas: 35750)
[PASS] test_BUG_nestedCreate() (gas: 67964)
[FAIL: Reverter != expected reverter: 0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f != 0x000000000000000000000000000000000000dEaD] test_CONTROL_callRevert_wrongAddress() (gas: 91240)
[PASS] test_CONTROL_ctorChildCall_correctAddress() (gas: 124024)
[FAIL: Reverter != expected reverter: 0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f != 0x000000000000000000000000000000000000dEaD] test_CONTROL_ctorChildCall_wrongAddress() (gas: 124041)
Suite result: FAILED. 4 passed; 2 failed; 0 skipped; finished in 978.98µs (2.14ms CPU time)
```

The three tests marked with the `BUG` comments should fail (the supplied address never matches the actual reverter). The `CONTROL` tests demonstrate that the reverter check works correctly when a CALL is involved — both for top-level CALLs and for CALLs nested inside a constructor.

### Expected behavior

`vm.expectRevert(address)` should fail whenever the supplied address doesn't match the contract whose code emitted the revert. In the case of a constructor, I think the most consistent choice is to only match the address of the contract that would be created should the constructor not revert. In the case of a `create` this address will be annoying to predict, but for `create2` it will be quite reasonable. I think the current behaviour of silently ignoring the address arg for reverts in `create` callframes is at risk of incorrectly passing a test that should fail.

### Scope

The bug only manifests when the innermost reverting frame is a CREATE — i.e., the constructor reverts directly via `revert(...)`, `require(false, ...)`, `assert(false)`, an unchecked panic, or a nested CREATE that itself reverts directly. A revert originating from a sub-CALL inside the constructor is unaffected (see `test_ctorChildCall_*` above).

Variants without a `reverter` argument (`expectRevert()`, `expectRevert(bytes)`, `expectRevert(bytes4)`) are unaffected.

mablr:
## Motivation

Close #14613

## Solution

The reverter address argument to `vm.expectRevert` was silently ignored when the innermost reverting frame was a CREATE (top-level or nested), because create_end never populated `expected_revert.reverted_by`.

Mirror call_end's logic in create_end: when the outcome reverts and a reverter address is expected, record `outcome.address` (revm guarantees this is Some(would-be address) whenever the constructor executed).

Adds positive regression tests for top-level and nested-CREATE reverts, and a negative regression test asserting wrong-reverter now fails.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -1897,10 +1897,23 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>> for Cheatcode
         }
 
         // Handle expected reverts
-        if let Some(expected_revert) = &self.expected_revert
+        if let Some(expected_revert) = &mut self.expected_revert
             && curr_depth <= expected_revert.depth
             && matches!(expected_revert.kind, ExpectedRevertKind::Default)
         {
+            // Mirror the logic in `call_end`: when an expected reverter address is set
+            // and we don't yet have one (or we're matching multiple reverts), record the
+            // would-be deployed address as the reverter. revm guarantees `outcome.address`
+            // is `Some(_)` whenever the constructor actually ran (including the revert
+            // case); it is only `None` for pre-frame rejection (depth/balance/nonce),
+            // for which a reverter address is meaningless.
+            if outcome.result.is_revert()
+                && expected_revert.reverter.is_some()
+                && (expected_revert.reverted_by.is_none() || expected_revert.count > 1)
+                && let Some(addr) = outcome.address
+            {
+                expected_revert.reverted_by = Some(addr);
+            }
             let mut expected_revert = std::mem::take(&mut self.expected_revert).unwrap();
             return match revert_handlers::handle_expect_revert(
                 false,
--- a/crates/forge/tests/cli/failure_assertions.rs
+++ b/crates/forge/tests/cli/failure_assertions.rs
@@ -71,11 +71,12 @@ Suite result: FAILED. 0 passed; 7 failed; 0 skipped; [ELAPSED]
             r#"No files changed, compilation skipped
 ...
 [FAIL: Reverter != expected reverter: [..] != 0x000000000000000000000000000000000000dEaD] testShouldFailExpectPartialRevertWrongReverterTopLevelCreate() ([GAS])
+[FAIL: Reverter != expected reverter: [..] != [..]] testShouldFailExpectRevertNestedCreateInnerAddress() ([GAS])
 [FAIL: Reverter != expected reverter: [..] != 0x000000000000000000000000000000000000dEaD] testShouldFailExpectRevertWithBytesWrongReverterTopLevelCreate() ([GAS])
 [FAIL: Reverter != expected reverter: [..] != 0x000000000000000000000000000000000000dEaD] testShouldFailExpectRevertWrongReverterNestedCreate() ([GAS])
 [FAIL: Reverter != expected reverter: [..] != 0x000000000000000000000000000000000000dEaD] testShouldFailExpectRevertWrongReverterTopLevelCreate() ([GAS])
 [FAIL: next call did not revert as expected] testShouldFailExpectRevertsNotOnImmediateNextCall() ([GAS])
-Suite result: FAILED. 0 passed; 5 failed; 0 skipped; [ELAPSED]
+Suite result: FAILED. 0 passed; 6 failed; 0 skipped; [ELAPSED]
 ...
 "#,
         );
--- a/crates/forge/tests/cli/failure_assertions.rs
+++ b/crates/forge/tests/cli/failure_assertions.rs
@@ -70,8 +70,9 @@ Suite result: FAILED. 0 passed; 7 failed; 0 skipped; [ELAPSED]
         .stdout_eq(
             r#"No files changed, compilation skipped
 ...
+[FAIL: Reverter != expected reverter: [..] != 0x000000000000000000000000000000000000dEaD] testShouldFailExpectRevertWrongReverterTopLevelCreate() ([GAS])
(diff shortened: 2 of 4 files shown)


grandizzy (dismissed):
lgtm!

zerosnacks (dismissed):
lgtm 👍

decofe (commented):
## Coverage Review: CREATE reverter address tests

The fix itself is clean — mirroring `call_end`'s `reverted_by` population into `create_end` is the right approach.

The tests cover 4 of the 7 affected `address reverter` overloads against CREATE frames:

| Overload | Covered? |
|---|---|
| `expectRevert(address)` | ✅ top-level + nested |
| `expectRevert(bytes, address)` | ✅ top-level |
| `expectPartialRevert(bytes4, address)` | ✅ top-level |
| `expectRevert(bytes4, address)` | ❌ |
| `expectRevert(bytes4, address, uint64)` | ❌ |
| `expectRevert(bytes, address, uint64)` | ❌ |
| `expectRevert(address, uint64)` | ❌ |

### Suggested additions

1. **`bytes4` selector + reverter** — `expectRevert(bytes4(keccak256("Error(string)")), expectedAddr)` against a top-level CREATE. The issue explicitly lists this overload.

2. **At least one `count`-bearing variant** — the `count > 1` branch in the new code (`expected_revert.reverted_by.is_none() || expected_revert.count > 1`) is never exercised for CREATEs. A test deploying two reverting contracts with e.g. `expectRevert(address, 2)` would confirm multi-revert counting works through `create_end`.

3. **CREATE2** — all tests use `new X()` (CREATE opcode). If `create_end` also handles CREATE2, a test with a salt-based deploy + reverter address would confirm coverage.

4. **Nested CREATE with inner-address semantics** — the nested tests verify the *outer* would-be address is the matched reverter. A negative test explicitly confirming that using the *inner* reverting contract's address as the expected reverter fails would document the intended semantics.

All untested paths funnel through the same `create_end` fix so they likely work, but since the issue explicitly called out all overloads, adding at least the `bytes4` and one `count` variant would make the regression suite airtight.

zerosnacks:
small fmt issue

grandizzy (dismissed):

zerosnacks (approved):

figtracer (approved):

d-xo:
Thanks for the quick fix! I'm not sure I quite understand the semantics of the new approach though?

In the case of a nested chain of creates, why would we take the address of the outer `create` to be the reverter and not the inner? This does not match the semantics of how reverts in call frames are handled (which always record the innermost reverting address).

PR #14617, opened 2026-05-06
cast: honor config and CLI tx gas limit toggle

serikssonn:
## Motivation
                                                                                                                                                                                               
  `cast run` had a semantic drift in how EIP-7825 tx gas limit behavior was applied.
                                                                                                                                                                                               
  The effective toggle was derived from local CLI args in `RunArgs` and then applied directly in `run_with_evm`, instead of consistently honoring the merged config/CLI result from `EvmOpts`. 
  This could make behavior depend on entrypoint-specific logic and reduce predictability when users set `enable_tx_gas_limit` in config.                                                       
                                                                                                                                                                                               
  ## Solution                                                                                                                                                                                  
                                                                                                                                                                                               
  This PR unifies tx gas limit toggle handling in `cast run`:                                                                                                                                  
                                                                                                                                                                                               
  - Resolve effective `enable_tx_gas_limit` from merged sources (`EvmOpts` + CLI).                                                                                                             
  - Centralize cap mapping logic (`enable -> None`, `disable -> Some(u64::MAX)`) in a helper.                                                                                                  
  - Apply the resolved value once when configuring `evm_env.cfg_env.tx_gas_limit_cap`.                                                                                                         
  - Align `disable_block_gas_limit` handling with merged semantics as well.                                                                                                                    
                                                                                                                                                                                               
  Also added focused unit tests for the toggle-resolution and cap-mapping helpers.                                                                                                             
                                                                                                                                                                                               
  ## PR Checklist                                                                                                                                                                              
                                                                                                                                                                                               
  - [x] Added Tests                                                                                                                                                                            
  - [ ] Added Documentation                                                                                                                                                                    
  - [ ] Breaking changes


crates/cast/src/cmd/run.rs
```diff
@@ -353,6 +352,14 @@ impl RunArgs {
...
 }
 
+fn resolve_enable_tx_gas_limit(from_config: bool, from_cli: bool) -> bool {
+    from_config || from_cli
+}
+
+fn tx_gas_limit_cap_from_toggle(enable_tx_gas_limit: bool) -> Option<u64> {
+    if enable_tx_gas_limit { None } else { Some(u64::MAX) }
+}
+
```
mablr: These helpers should be inlined, imo they obscure the code. Then their unit tests can be dropped.

mablr (commented):
Makes sense, maybe, let's introduce an higher level integration test ?

crates/cast/src/cmd/run.rs
serikssonn: Addressed — I inlined the toggle logic and removed the helper unit tests in 61b8f5255.

serikssonn:
> Makes sense, maybe, let's introduce an higher level integration test ?

Sounds good, I’ll add a higher-level integration test for this.

PR #14619, opened 2026-05-06
fix(cheatcodes): preserve reverts with `expectEmit`

### LINKED ISSUE #10463: expectEmit behaves weirdly
### Component

Forge

- [x] Foundry
- [ ] Foundryup

1.1.0-stable

0.3.3

forge t

### Operating System

None

### Describe the bug

in a call following a `expectEmit`, if a reverts happens before the event emission the test will fail with `log != expected log`. It seems way more natural to fail with `EvmError: Revert` for two reasons: nobody should be testing even emission on failing paths (events are not emitted if the call reverted), and reverts are "more important" than wrong log. Also the inconsistent behaviour if the event is emitted before the revert is weird (the event will not be emitted if the call reverts).

minimal example:
```
    event Foo();

    function revertingBefore() external {
        revert();
        emit Foo();
    }

    function revertingAfter() external {
        emit Foo();
        revert();
    }

    // this test fails with `log != expected log`
    function testBefore() public {
        vm.expectEmit();
        emit Foo();
        this.revertingBefore(); // reverting call
    }

    // this test fails with `EvmError: Revert`
    function testAfter() public {
        vm.expectEmit();
        emit Foo();
        this.revertingAfter(); // reverting call
    }
```

srdtrk:
## Motivation

Fixes #10463

`vm.expectEmit` could mask a reverting call with `log != expected log` when the call reverted before emitting the expected event.
In that case, the original EVM revert is the primary failure and should be reported as `EvmError: Revert`. This matches the behavior when the event is emitted before the revert, and avoids surfacing a secondary emit-mismatch error for a failing execution path.

## Solution

Return early from `call_end` when the call outcome is already a revert, after preserving any fork revert diagnostic. This prevents later `expect*` validation from overwriting the original EVM revert with an `expectEmit` mismatch.

Added a regression test for the reported case where `vm.expectEmit` is followed by a call that reverts before emitting.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


figtracer:
Closing this in favor of documentation updates:

https://github.com/foundry-rs/foundry/pull/14620
https://github.com/foundry-rs/book/pull/1819

srdtrk:
@figtracer, this PR is not related to that issue. Would you like me to re-open the PR?

zerosnacks:
Sorry @srdtrk there was a mix up; we will get this PR reviewed on its own - thanks for your contribution!

figtracer:
> @figtracer, this PR is not related to that issue. Would you like me to re-open the PR?

sry! thought this was related

mablr (approved):
lgtm, thanks!

zerosnacks (approved):

PR #14612, opened 2026-05-06
fix(ci): keep no-default builds free of op deps

emmajam:
## Summary
- set the leaf binary workspace deps (`anvil`, `cast`, `forge`) to `default-features = false` so dev-dep edges do not re-enable OP support in no-default builds
- move OP-only RPC test deps in `anvil` and `foundry-evm-core` behind the existing `optimism` feature
- add a CI job for `cargo build --workspace --no-default-features --locked`

## Verification
- `cargo build --workspace --no-default-features --locked`
- `cargo tree --workspace --no-default-features | rg "(^|[[:space:]])(op-|alloy-op)" || true` (no output)
- `cargo +1.94 test -p foundry-evm-core --lib --locked`
- `cargo +1.94 test -p anvil --test it --no-run --locked`
- `cargo +nightly fmt --all --check`

PR #14622, opened 2026-05-06
fix: clean --no-default-features test build

stevencartavia:
Fixes `cargo check --tests --workspace --no-default-features`:

- Gate `mod state;` in anvil's `it` test binary behind `cfg(feature = "cmd")` — its tests use `NodeConfig::with_init_state_path()`, which is itself gated on `cmd`.
- Drop redundant `::default()` on the unit-struct `OpCallDepositInfo` in the `not(optimism)` arm of `mod.rs` (clippy `default_constructed_unit_structs`).
- Move `alloy_consensus::Sealed` import in `evm/core/src/env.rs` tests into the `optimism` submodule where it's actually used.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -188,6 +188,7 @@ jobs:
       - test
       - docs
       - doctest
+      - no-default-features
       - typos
       - clippy
       - rustfmt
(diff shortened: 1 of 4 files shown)


zerosnacks:
Ah lets also add `no-default-features` here https://github.com/foundry-rs/foundry/blob/2e8ae6c606ebe37d3e5f42a38f777da2e947e3d6/.github/workflows/ci.yml#L187-L196; I had missed that in @emmajam's PR: https://github.com/foundry-rs/foundry/pull/14612

zerosnacks (approved):

PR #14623, opened 2026-05-06
feat(lint): add `name-reused`

mablr:
## Motivation

Introduce `name-reused` info lint.

Flags contract, interface, or library names that are defined in more than one source file in the project.

Associated book PR https://github.com/foundry-rs/book/pull/1820

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


zerosnacks (dismissed):

figtracer:
apparentely this lint is Truffle-specific (https://github.com/crytic/slither/issues/2390) because Truffle can overwrite/lose e.g.: the corresponding generated`.json` artifacts for these contracts

```
src/Token.sol      contract Token { source() returns 1 }
test/Token.sol     contract Token { source() returns 2 }
```

tested locally with Foundry and we don't have this problem

zerosnacks:
> apparentely this lint is Truffle-specific ([crytic/slither#2390](https://github.com/crytic/slither/issues/2390)) because Truffle can overwrite/lose e.g.: the corresponding generated`.json` artifacts for these contracts
> 
> ```
> src/Token.sol      contract Token { source() returns 1 }
> test/Token.sol     contract Token { source() returns 2 }
> ```
> 
> tested locally with Foundry and we don't have this problem

Hmm I see, I think this lint may then cause an unnecessary number of false positives for the stated benefit it provides

mablr (commented):
Given this remark this lint is not very valuable for Foundry, just an info-level style lint.

PR #14624, opened 2026-05-06
fix(anvil): load blocks and transactions atomically in `load_state`

AissataF:
## Summary      

  `load_state` took two separate `write()` locks — one for `load_blocks`, one for `load_transactions` —
   leaving a window where `storage.blocks` held the new entries but `storage.transactions` was still
  empty. Concurrent `eth_getBlockByHash` (full), `eth_getTransactionReceipt` and `eth_getBlockReceipts`
   hit that window, short-circuited via `?` in `mined_transactions_in_block` / `mined_receipts`, and
  returned `None` — making freshly loaded blocks transiently appear missing. Fix: hold a single
  `write()` guard across both calls.

## Reproduction

  Stress test: 6 reader tasks tight-loop `block_by_hash` + `block_by_hash_full` on known hashes while  
  the main task calls `load_state`. A race read = lite form `Some` but full form `None`. To make the
  window observable I temporarily inserted `std::thread::sleep(50ms)` between the two `write()` calls  
  (experiment only, not in this patch).

  | Code | Mismatched reads |
  | --- | --- |
  | Before (two `write()` locks) | **36,754** |
  | After (single `write()` guard) | **0** |

  Before fix:
  ```
  assertion `left == right` failed: observed 36754 reads where the block hash existed but the full
  block was None — load_state left a window with blocks loaded but transactions missing                
    left: 36754
   right: 0                                                                                            
  test state::load_state_atomic_blocks_and_transactions ... FAILED                                     
  ```
                                                                                                       
  After fix:          
  ```
  test state::load_state_atomic_blocks_and_transactions ... ok
  ```

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -3888,9 +3888,13 @@ impl<N: Network<ReceiptEnvelope = FoundryReceiptEnvelope>> Backend<N> {
 
     /// Apply [SerializableState] data to the backend storage.
     pub async fn load_state(&self, state: SerializableState) -> Result<bool, BlockchainError> {
-        // load the blocks and transactions into the storage
-        self.blockchain.storage.write().load_blocks(state.blocks.clone());
-        self.blockchain.storage.write().load_transactions(state.transactions.clone());
+        // load the blocks and transactions into the storage atomically so concurrent readers
+        // never observe blocks without their transactions
+        {
+            let mut storage = self.blockchain.storage.write();
+            storage.load_blocks(state.blocks.clone());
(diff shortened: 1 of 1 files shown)


mablr (approved):
Makes sense

AissataF:
> Makes sense

Thanks! @mablr out of topic, any chance that can push my PRs a little further? thanks! https://github.com/foundry-rs/foundry/issues?q=is%3Apr+is%3Aopen+author%3AAissataF

mablr:
@AissataF Yes, thank you very much for all these contributions; I'm trying to catch up on the reviews ...😅

mattsse (approved):
Reviewed: the storage lock scope change is correct and closes the observed load_state race.

PR #14625, opened 2026-05-06
fix(cheatcodes): innermost as reverter for nested CREATE chain revert

mablr:
## Motivation

#14615 follow-up, `vm.expectRevert` for nested `CREATE` chains recorded the outer would-be address as the reverter, inconsistent with single-revert `CALL` semantics (which record the innermost reverting address) https://github.com/foundry-rs/foundry/pull/14615#issuecomment-4389574212.

## Solution

Moved `create_end`'s `reverted_by` write outside the depth gate so every reverting `CREATE` frame can attempt to record its would-be deployed address. The deepest `CREATE` frame fires first and the `is_none()` lock pins it, so nested `CREATE` / `CREATE2` chains use the innermost failed deployment as the reverter. For `count > 1`, the lock is reset after each successful CREATE expectation iteration. Existing nested `CALL` count behavior is intentionally unchanged.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cheatcodes/spec/src/vm.rs
+++ b/crates/cheatcodes/spec/src/vm.rs
@@ -1165,14 +1165,26 @@ interface Vm {
     function expectRevert(bytes calldata revertData) external;
 
     /// Expects an error with any revert data on next call to reverter address.
+    ///
+    /// The `reverter` argument is matched against the address of the innermost
+    /// frame that produced the revert:
+    ///   - For a CALL: the address that was called.
+    ///   - For a CREATE / CREATE2: the would-be deployed address of the failed
+    ///     deployment (computed from the deployer + nonce, or salt + initcode).
+    ///
+    /// In a nested chain the deepest reverting frame wins, regardless of whether
+    /// the chain is composed of CALLs, CREATEs, or a mix. With `count > 1` the
+    /// same rule applies independently to each iteration.
     #[cheatcode(group = Testing, safety = Unsafe)]
     function expectRevert(address reverter) external;
 
     /// Expects an error from reverter address on next call, with any revert data.
+    /// See `expectRevert(address)` for `reverter` matching semantics.
     #[cheatcode(group = Testing, safety = Unsafe)]
     function expectRevert(bytes4 revertData, address reverter) external;
 
     /// Expects an error from reverter address on next call, that exactly matches the revert data.
+    /// See `expectRevert(address)` for `reverter` matching semantics.
     #[cheatcode(group = Testing, safety = Unsafe)]
     function expectRevert(bytes calldata revertData, address reverter) external;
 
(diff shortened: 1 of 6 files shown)


mablr (commented):
@d-xo wdyt :) ?

Maybe not the cleanest, but I think it matches the behavior you specified.

crates/cheatcodes/src/inspector.rs
```diff
@@ -1898,49 +1898,50 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>> for Cheatcode
...
+        // Handle expected reverts.
+        if let Some(expected_revert) = &mut self.expected_revert {
+            // Record the would-be deployed address as the reverter, mirroring `call_end`.
+            // Runs at every depth so the innermost CREATE wins (deepest frame fires first
+            // and the `is_none()` lock pins it). `outcome.address` is `None` only when the
+            // constructor never ran (pre-frame rejection), in which case the surrounding
+            // `call_end` records the caller.
             if outcome.result.is_revert()
                 && expected_revert.reverter.is_some()
                 && (expected_revert.reverted_by.is_none() || expected_revert.count > 1)
```
figtracer: unsure about this `count > 1` condition here

a repeated `CREATE2` nested-constructor case might report the outer address as the reverter, not the inner one

d-xo:
> @d-xo wdyt :) ?
> 
> Maybe not the cleanest, but I think it matches the behavior you specified.

not really into the code enough to have an opinion on cleanliness but the tests seem valid and match my expectations. thanks!

crates/cheatcodes/src/inspector.rs
mablr: fixed 👍

figtracer (dismissed):
lg, should match `create_end` now

zerosnacks (dismissed):

crates/cheatcodes/spec/src/vm.rs
```diff
@@ -1165,14 +1165,26 @@ interface Vm {
...
+    ///
+    /// The `reverter` argument is matched against the address of the innermost
+    /// frame that produced the revert:
+    ///   - For a CALL: the address that was called.
+    ///   - For a CREATE / CREATE2: the would-be deployed address of the failed
+    ///     deployment (computed from the deployer + nonce, or salt + initcode).
+    ///
+    /// In a nested chain the deepest reverting frame wins, regardless of whether
+    /// the chain is composed of CALLs, CREATEs, or a mix. With `count > 1` the
+    /// same rule applies independently to each iteration.
```
figtracer: i think this is wrong, `CALL` actually differs:
- `count == 1`: deepest wins
- `count > 1`: outermost wins per revert iteration

this commit https://github.com/foundry-rs/foundry/pull/14625/changes/da190d8cf541b0a99deb8db7149e21d50d772b27 removed `count > 1` condition & reset `reverted_by`, which effectively made "deepest wins per iteration" the default for `CREATE` 

if we want this to be the global default, it makes sense to do the same thing for `CALL`; otherwise, we can revert https://github.com/foundry-rs/foundry/pull/14625/changes/da190d8cf541b0a99deb8db7149e21d50d772b27 which makes both `CREATE` and `CALL` match (`1` deepest wins/`>1`outermost wins per iter). 

IMO, the former is the path to choose.

figtracer (commented):
insight from new docs

zerosnacks:
https://github.com/foundry-rs/foundry/pull/14625/commits/65039fae496ef13fb61dedd4efa613da61f9b347 would be a breaking change for users which could break their test suites; considering the core bug of https://github.com/foundry-rs/foundry/issues/14613 has been resolved I think we should not rush this refinement and take the time to fully evaluate the different options we have.

mattsse (approved):
Reviewed the current head after the follow-up docs update. The CALL-side behavior change from `65039fa` has been reverted, the remaining CREATE/CREATE2 behavior is covered by tests, and the public docs/PR description now distinguish nested CREATE count semantics from the existing nested CALL count behavior.

PR #14626, opened 2026-05-06
fix(script): treat unknown tx as dropped to avoid receipt-poll hang

stevencartavia:
## Motivation

`forge script`'s receipt watcher hangs forever polling `eth_getTransactionReceipt` when a tx is rejected at submission or dropped from the mempool (e.g. an EVM halt under anvil's new `-32003` mapping from #14592). See [CI run](https://github.com/foundry-rs/foundry/actions/runs/25446318617/job/74667254337).

## Fix

`provider.get_transaction_by_hash(hash)` returns `Result<Option<Tx>, _>`. `check_tx_status` matched `Ok(_)` for the "node knows the tx" branch, wrongly catching `Ok(None)` and feeding it into an unbounded retry loop. Match `Ok(Some(_))` explicitly and treat `Ok(None)` as `TxStatus::Dropped`.


Cargo.lock
zerosnacks, line 1: Curious why we need the Cargo bump here

stevencartavia: all the failures in the previous commit were because of that

mattsse: doesnt really make sense -.-

crates/script/src/receipts.rs
```diff
@@ -57,28 +57,27 @@ pub async fn check_tx_status<N: Network>(
...
                             // Transaction is still known to the node, retry
                             Err(RetryError::Retry(PendingReceiptError { tx_hash: hash }.into()))
                         }
-                        Err(_) => {
-                            // Transaction is not known to the node, mark it as dropped
-                            Ok(TxStatus::Dropped)
-                        }
+                        // Tx unknown to the node (rejected/dropped) or transport error: mark
+                        // dropped instead of polling forever.
+                        Ok(None) | Err(_) => Ok(TxStatus::Dropped),
```
mattsse, line 68: right, this should check Ok(None

Cargo.lock
stevencartavia: undone

mablr (approved):

zerosnacks (approved):
Makes sense 👍

PR #14627, opened 2026-05-06
fix(script): only set fee_token on library deploy txs

decofe:
## Motivation

[#14594](https://github.com/foundry-rs/foundry/pull/14594) replaced targeted `set_fee_token()` + `apply_expires()` calls in `runner.rs` with a blanket `tempo.apply()`. This applies **all** `TempoOpts` fields (nonce_key, valid_before, key_id, sponsor) to library deploy transactions, which breaks local Anvil (`--hardfork t3`) — the transactions are malformed and Anvil stops mining blocks.

`forge script --broadcast` then polls forever for receipts that never arrive. See [CI run](https://github.com/foundry-rs/foundry/actions/runs/25446318617/job/74667254337) — the hang is at `=== ANVIL LOCAL: FORGE SCRIPT BROADCAST ===`.

## Fix

Restore the pre-#14594 behaviour in `runner.rs`: only propagate `fee_token` to library deploy transaction requests. Library deploys use sequential nonces computed by the runner and should never receive expiring nonces, nonce keys, key_id, or sponsor fields — those belong on user broadcast transactions only.

The `tempo.apply()` call in `broadcast.rs` (for actual user broadcast txs) is unaffected.

## Note

[#14626](https://github.com/foundry-rs/foundry/pull/14626) separately fixes `forge script` hanging forever when a tx is dropped/rejected (a defence-in-depth improvement to `check_tx_status`). Both fixes are independently valuable — this PR fixes the root cause, #14626 fixes the symptom.

### DIFF
--- a/crates/script/src/runner.rs
+++ b/crates/script/src/runner.rs
@@ -6,7 +6,7 @@ use alloy_network::TransactionBuilder;
 use alloy_primitives::{Address, Bytes, U256};
 use eyre::Result;
 use foundry_cheatcodes::BroadcastableTransaction;
-use foundry_common::TransactionMaybeSigned;
+use foundry_common::{FoundryTransactionBuilder, TransactionMaybeSigned};
 use foundry_config::Config;
 use foundry_evm::{
     constants::CALLER,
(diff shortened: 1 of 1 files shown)


mablr (commented):
I actually think all this `TempoOpts`'s dedup attempt is flawed from the beginning https://github.com/foundry-rs/foundry/pull/14558 (see discussion and follow-ups).

Let's revert completely ? wdyt @figtracer

figtracer:
> I actually think all this `TempoOpts`'s dedup attempt is flawed from the beginning https://github.com/foundry-rs/foundry/pull/14558 (see discussion and follow-ups).
> 
> 
> 
> Let's revert completely ? wdyt @figtracer 

yeah this looked a bit more interesting at the start, maybe there's a better way to do this 

i'm not against reverting it but maybe it's easier to just refactor it properly on top at this point

would like to know @zerosnacks's perspective

zerosnacks:
I'm OK with doing a clean revert and redoing it with the CI in place when we have found an approach.

Unrelated flags showing up because of grouping is a long standing issue. Ideally we would want to only show the exact relevant flags in each place. Sometimes there are flags that are noops, misleading or worst case behave differently than one might expect.

PR #14630, opened 2026-05-06
fix(config): gate etherscan no_proxy behind --no-proxy flag

decofe:
## Motivation

[#14454](https://github.com/foundry-rs/foundry/pull/14454) unconditionally called `reqwest::ClientBuilder::no_proxy()` on every Etherscan client. This breaks users behind corporate proxies or anyone relying on `HTTP_PROXY`/`HTTPS_PROXY` environment variables.

The `--no-proxy` flag already exists (added in [#13155](https://github.com/foundry-rs/foundry/pull/13155)) and gates proxy disabling for the RPC transport layer, but the Etherscan client was not wired up to it.

## Fix

- Add `into_client_with_no_proxy(bool)` to `ResolvedEtherscanConfig` — only calls `.no_proxy()` on the reqwest builder when `true`.
- Keep `into_client()` as a convenience wrapper that defaults to `false` (proxy-enabled).
- Thread `config.eth_rpc_no_proxy` through all production call sites: `cast storage`, `cast creation-code`, `forge clone`, `cli::utils::fetch_abi_from_etherscan`, and the external trace identifier.

Users in sandboxed environments (the original issue [#12733](https://github.com/foundry-rs/foundry/issues/12733)) pass `--no-proxy` and get the crash fix. Everyone else keeps working proxy support.


crates/config/src/etherscan.rs
```diff
@@ -307,10 +321,11 @@ impl ResolvedEtherscanConfig {
             }
         }
 
-        // Disable automatic proxy detection. In sandboxed environments (e.g., Cursor IDE,
-        // macOS App Sandbox), reqwest's system proxy lookup via SCDynamicStore can crash
-        // when the API returns NULL. See: https://github.com/foundry-rs/foundry/issues/12733
```
mattsse: lets keep this comment

crates/config/src/etherscan.rs
```diff
@@ -294,6 +294,20 @@ impl ResolvedEtherscanConfig {
...
+
+    /// Same as [`Self::into_client`] but optionally disables automatic proxy detection.
+    ///
+    /// When `no_proxy` is `true`, the underlying `reqwest` client is built with
+    /// [`reqwest::ClientBuilder::no_proxy`], which prevents system proxy lookups that can
+    /// crash in sandboxed environments (e.g., Cursor IDE, macOS App Sandbox).
+    /// See: <https://github.com/foundry-rs/foundry/issues/12733>
+    pub fn into_client_with_no_proxy(
+        self,
+        no_proxy: bool,
```
mattsse, line 309: this should probably accept the clientbuilder instead and then we can set

https://github.com/foundry-rs/foundry-core/blob/640e3d2a0b4499dd1e38890995b4fa77777ab6ac/crates/explorers/block-explorers/src/lib.rs#L314

on the callsite

mattsse (changes requested):

crates/config/src/etherscan.rs
mablr: Unsure this is better, it makes all callsites more complex, and i haven't found yet any specific need for it in the codebase.

Before:
```
.into_client_with_no_proxy(config.eth_rpc_no_proxy)?;
```

After
```
.into_client_with_builder(
    Client::builder().set_no_proxy(config.eth_rpc_no_proxy),
)?;
```

mattsse (approved):

PR #14633, opened 2026-05-07
feat: recognize check_/prove_ symbolic tests

stevencartavia:
Adds a `SymbolicTest` variant to `TestFunctionKind` and recognizes `check_* `/` prove_*` (+ `Fail` variants) as symbolic tests. With no backend or own engine wired up yet, the runner skips them with a clear reason. First step toward symbolic testing in Forge.

### DIFF
--- a/crates/common/src/traits.rs
+++ b/crates/common/src/traits.rs
@@ -3,7 +3,10 @@
 use alloy_json_abi::Function;
 use alloy_primitives::Bytes;
 use alloy_sol_types::SolError;
-use std::{fmt, path::Path};
+use std::{
+    fmt::{self, Display, Formatter},
+    path::Path,
+};
 
 /// Test filter.
 pub trait TestFilter: Send + Sync {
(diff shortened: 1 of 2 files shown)


zerosnacks (commented):
Nice, I think for the time being I prefer we use a staging branch as target for the symbolic testing feature as we develop. That way we can cleanly merge at the end and do not have a partial feature that blocks a release. For reference see the mutation testing PR.

figtracer:
> I prefer we use a staging branch as target for the symbolic testing feature as we develop

+1, also i think we shouldn't start this work without having a clearer path first. 

it's still unclear to me if we're going for a Halmos approach, if we're integrating it with the fuzzer itself like this or something different. 

we have issues on Linear (e.g.: https://linear.app/tempoxyz/issue/OSS-89) that are still very exploratory so IMO until we get this figured out, nothing should be done.

i think a good step forward would be to schedule a quick meeting with George when time permits (or async) to understand if any of these were discarded and if we have better options

zerosnacks:
Agreed w/ @figtracer, marked as draft for now

zerosnacks:
created: https://github.com/foundry-rs/foundry/tree/staging/symbolic-testing, repointed target to it

grandizzy:
@stevencartavia @figtracer is this still needed with current symbolic changes?

figtracer:
> @stevencartavia @figtracer is this still needed with current symbolic changes?

i think we can close it; we already have this on the main PR

grandizzy:
Closing per above, will reopen if needed

PR #14635, opened 2026-05-07
fix(anvil): stop node tasks on handle shutdown

Perico-perica46:
## Motivation

Dropping `NodeHandle` fired the shutdown signal, but the core HTTP/WS server tasks, IPC listener, and `NodeService` did not observe that signal. That could leave listeners running after the handle was dropped.

## Solution

Thread the existing shutdown receiver into `NodeService`, HTTP/WS server tasks, and the IPC listener. Add a regression test that drops a handle and verifies the HTTP listener address can be rebound.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/lib.rs
+++ b/crates/anvil/src/lib.rs
@@ -33,7 +33,7 @@ use foundry_primitives::FoundryNetwork;
 use futures::{FutureExt, TryFutureExt};
 use parking_lot::Mutex;
 use revm::primitives::hardfork::SpecId;
-use server::try_spawn_ipc;
+use server::try_spawn_ipc_with_shutdown;
 use std::{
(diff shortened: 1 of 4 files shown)


mattsse (changes requested):
I dont think we need this

because we already have

https://github.com/foundry-rs/foundry/blob/27568f705da3be4ad3d38a7835bbd8a957501c11/crates/anvil/src/lib.rs#L293-L298

so we could simply abort them on drop

mattsse:
Closing this because it is superseded by #14636, which keeps the same motivation but uses the existing `NodeHandle` join handles to abort the node tasks on drop.

PR #14636, opened 2026-05-07
fix(anvil): abort node tasks on handle drop

mattsse:
## Motivation

Dropping `NodeHandle` already fires the shutdown signal, but the core `NodeService`, HTTP/WS server tasks, and IPC listener are also retained as join handles on the handle itself. PR #14635 identified that dropping the handle can leave listeners running; this keeps the same motivation while using the existing handle ownership path instead of threading shutdown through the server and service internals.

## Solution

Abort the stored node service, server, and IPC join handles from `NodeHandle::drop` after firing the existing shutdown signal. Make the `node_service` and `servers` fields private so task ownership remains internal to `NodeHandle`. Add a regression test that drops a handle and verifies the HTTP listener address can be rebound.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/anvil/src/lib.rs
+++ b/crates/anvil/src/lib.rs
@@ -291,9 +291,9 @@ pub struct NodeHandle {
     /// The address of the running rpc server.
     addresses: Vec<SocketAddr>,
     /// Join handle for the Node Service.
-    pub node_service: JoinHandle<Result<(), NodeError>>,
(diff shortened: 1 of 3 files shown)


mablr (approved):
noice

figtracer (approved):

PR #14638, opened 2026-05-07
ci: run test-isolate on push to master, move flaky isolate tests to test-flaky

zerosnacks:
## Why

Will allows us to catch regressions like https://github.com/foundry-rs/foundry/actions/runs/25468453538/job/74726848813#step:14:1 earlier in the day without losing velocity of shipping PRs; if failures are found it is expected that the developer follows up with a fix.

Also removed the ticket opening; master can be flaky from time to time and this would be too annoying; we are already seeing this on nightlies.

## Changes

- `test-isolate` now runs on `push` to `master` (and `workflow_dispatch`) instead of nightly.
- Removed the issue-opening job from `test-isolate.yml` (and deleted the unused `TEST_ISOLATE_FAILURE_TEMPLATE.md`); failures will surface directly on the master commit instead.
- Moved the isolated flaky tests job (and its issue-opening job using `FLAKY_TEST_ISOLATE_FAILURE_TEMPLATE.md`) from `test-isolate.yml` into `test-flaky.yml`, which still runs nightly.


mablr (approved):
Nice !

PR #14645, opened 2026-05-07
Use Monotonic Last-Worker Ordering

ml0mbardi:
## Motivation

  In parallel fuzzing, we were using `SystemTime` (millisecond wall-clock timestamps) to decide which worker ran last. That is not reliable for ordering: multiple workers can complete within the same millisecond, and wall-clock time is not strictly monotonic. As a result, traces/logs/breakpoints can be taken from the wrong worker, making failures harder to debug and less reproducible.

  ## Solution

  Replace timestamp-based ordering with a monotonic run sequence:

  - Replace `last_run_timestamp` with `last_run_seq` in worker state.
  - Update `last_run_seq` from the global atomic `total_runs` counter when a run is recorded.
  - Select the “last worker” during aggregation by highest `last_run_seq`.

  This keeps fuzz execution unchanged, but makes result aggregation deterministic and correct under concurrency.

  ## PR Checklist

  - [ ] Added Tests
  - [ ] Added Documentation
  - [ ] Breaking changes

### DIFF
--- a/crates/evm/evm/src/executors/fuzz/mod.rs
+++ b/crates/evm/evm/src/executors/fuzz/mod.rs
@@ -73,10 +73,10 @@ struct WorkerState<FEN: FoundryEvmNetwork> {
     failure: Option<TestCaseError>,
     /// Fuzz run metadata that produced the failure.
     failure_run: Option<FuzzRunMetadata>,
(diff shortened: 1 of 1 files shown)


figtracer:
we can keep this as-is

PR #14650, opened 2026-05-07
fix(anvil): accept string values for u64 numeric RPC params

AissataF:
`anvil_setIntervalMining` / `anvil_setChainId` / `anvil_setBlockTimestampInterval` / `anvil_rollback` were typed as `u64` with the `sequence` deserializer, which only accepts raw JSON numbers. Standard JSON-RPC clients commonly encode numeric params as `0x`-prefixed hex strings, so those calls could be rejected with `invalid type: string "0x...", expected u64`.

This keeps the request variants typed as `u64` and limits the change to serde. The new `u64` helpers accept JSON numbers plus hex/decimal strings for the affected params, preserve the optional rollback shapes, and reject values that do not fit in `u64`. The helper naming and tests now make the accepted bare-value and one-element sequence forms explicit.

### DIFF
--- a/crates/anvil/core/src/eth/mod.rs
+++ b/crates/anvil/core/src/eth/mod.rs
@@ -18,7 +18,8 @@ use alloy_rpc_types::{
 };
 use alloy_serde::WithOtherFields;
 use foundry_common::serde_helpers::{
-    deserialize_number, deserialize_number_opt, deserialize_number_seq, deserialize_number_seq_opt,
+    deserialize_number, deserialize_number_opt, deserialize_number_seq, deserialize_u64_seq,
+    deserialize_u64_seq_opt,
 };
 
 pub mod block;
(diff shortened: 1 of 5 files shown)


mattsse (changes requested):
I'm open to accepting both u64 and U64 here but should not introduce type changes and only be limited to deserde impl

mattsse (dismissed):
Looks good after the serde-only update. The request variants stay typed as `u64`, the helper name now reflects the direct-value-or-single-param-sequence behavior, and the added coverage locks down accepted shapes plus overflow rejection.

mattsse (approved):
Re-approving after the clippy-only follow-up. The new commit only switches the `U64` conversion to `Self` to satisfy `clippy::use_self`; the exact CI clippy command passes locally.

PR #14661, opened 2026-05-08
feat(lint): add missing-inheritance lint

stevencartavia:
New info `missing-inheritance` lint that flags contracts implementing every external function of an interface without explicitly inheriting from it.

`ProjectLintPass` now exposes gcx via `ProjectLintEmitter::gcx()` so cross-file passes can opt into semantic analysis without bloating the trait signature.

### DIFF
--- a/crates/lint/src/sol/info/missing_inheritance.rs
+++ b/crates/lint/src/sol/info/missing_inheritance.rs
@@ -1,0 +1,204 @@
+use crate::{
+    linter::{Lint, ProjectLintEmitter, ProjectLintPass, ProjectSource},
+    sol::{Severity, SolLint, info::MissingInheritance},
+};
+use solar::{
+    interface::{Span, source_map::FileName},
+    sema::{
+        Gcx,
+        hir::{ContractId, ContractKind, FunctionKind, SourceId},
+    },
+};
+use std::collections::{BTreeSet, HashMap};
+
+declare_forge_lint!(
+    MISSING_INHERITANCE,
+    Severity::Info,
+    "missing-inheritance",
+    "contract implements an interface's external API but does not explicitly inherit from it"
+);
+
+impl<'ast> ProjectLintPass<'ast> for MissingInheritance {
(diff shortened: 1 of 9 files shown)


crates/lint/src/sol/info/missing_inheritance.rs
```diff
@@ -0,0 +1,204 @@
...
+        // "analysis target", and precompute its external selector set.
+        let mut candidates: Vec<(ContractId, BTreeSet<[u8; 4]>)> = Vec::new();
+        let mut targets: Vec<ContractId> = Vec::new();
+        let mut selectors_by_contract: HashMap<ContractId, BTreeSet<[u8; 4]>> = HashMap::new();
+
+        for cid in hir.contract_ids() {
+            let contract = hir.contract(cid);
+            if contract.linearization_failed() || !input_source_idx.contains_key(&contract.source) {
+                continue;
+            }
```
mablr: Here we filter by `input_source_idx`, which only contains user-provided files.

Not 100% sure but if a contract implements for example IERC20 from an external library (OpenZeppelin, etc.) without inheriting it, the lint may be silent? Could you double check ?

crates/lint/src/sol/info/missing_inheritance.rs
```diff
@@ -0,0 +1,204 @@
...
+/// should be considered as candidate interfaces for the missing-inheritance check.
+fn is_signature_only<'gcx>(gcx: Gcx<'gcx>, cid: ContractId) -> bool {
+    let hir = &gcx.hir;
+    let contract = hir.contract(cid);
+
+    let mut has_function = false;
+    for &item_id in contract.items {
+        match item_id {
+            solar::sema::hir::ItemId::Variable(_) => return false,
+            solar::sema::hir::ItemId::Function(fid) => {
```
mablr: Let's use import :)

crates/lint/src/sol/info/missing_inheritance.rs
stevencartavia: good catch

mablr (approved):
lgtm (needs book PR to be merged)

zerosnacks (approved):

PR #14664, opened 2026-05-08
chore(deps): pull foundry-compilers fix for PEP 440 Vyper pragmas

zerosnacks:
Counterpart to https://github.com/foundry-rs/foundry-core/pull/59 (`fix(compilers/vyper): support PEP 440 pragmas`).

The released `foundry-compilers` 0.20.0 silently drops every Vyper `pragma version` constraint:

1. `RE_VYPER_VERSION` does not match `# pragma version <req>` (with a space after `#`), which is exactly what [`crates/forge/assets/vyper/CounterTemplate.vy`](https://github.com/foundry-rs/foundry/blob/master/crates/forge/assets/vyper/CounterTemplate.vy) (`# pragma version ~=0.4.3`) and [`VYPER_INTERFACE`](https://github.com/foundry-rs/foundry/blob/master/crates/forge/tests/cli/compiler.rs#L34) (`# pragma version >=0.4.0`) use.
2. When the regex matches, the parser feeds the **outer** match to `semver::VersionReq::parse` instead of the named `version` group, so even legacy pragmas are discarded.
3. Vyper >= 0.4.0 uses PEP 440 (`~=0.5.0`, `0.5.0a1`, `==0.5.0`, …), none of which parse as semver.

Pulled in via a temporary `[patch.crates-io]` so foundry CI can verify the fix before `foundry-compilers` is released. **Draft**: drop the patch and bump the version in `[workspace.dependencies]` once the next `foundry-compilers` is published.

Adds five regression tests in `crates/forge/tests/cli/compiler.rs` covering positive (`~=0.4.0`, `#@version ^0.4.0`) and negative (`==0.3.10`, `==0.5.0b1`, `==0.5.0rc1`) pragma resolution, and updates the `cast` snapshot for the language-neutral error wording introduced by the upstream commit.

### Behavioural caveat

Because the released `foundry-compilers` returns `None` for `version_req` of *every* Vyper file today, this is the first time Vyper version pragmas actually constrain compiler resolution in foundry. Projects with stale or wrong pragmas may now see a version-mismatch error where they previously compiled silently. Worth a CHANGELOG note.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -596,6 +596,16 @@ alloy-op-hardforks = { git = "https://github.com/ethereum-optimism/optimism", re
 ## foundry-fork-db
 # foundry-fork-db = { git = "https://github.com/foundry-rs/foundry-core", rev = "2f90eb86d4549fa15a8cc2d99bfc1039bc083977" }
 
+## foundry-compilers — pull in https://github.com/foundry-rs/foundry-core/pull/59
+## (fix(compilers/vyper): support PEP 440 pragmas) ahead of the next
+## foundry-compilers release. Drop this block once foundry-compilers >= 0.20.1
+## is published and bump the version in `[workspace.dependencies]`.
+foundry-compilers = { git = "https://github.com/figtracer/foundry-core", rev = "63698079e3971a83fe643c4f14b8c5cd3fbed206" }
+foundry-compilers-artifacts = { git = "https://github.com/figtracer/foundry-core", rev = "63698079e3971a83fe643c4f14b8c5cd3fbed206" }
+foundry-compilers-artifacts-solc = { git = "https://github.com/figtracer/foundry-core", rev = "63698079e3971a83fe643c4f14b8c5cd3fbed206" }
+foundry-compilers-artifacts-vyper = { git = "https://github.com/figtracer/foundry-core", rev = "63698079e3971a83fe643c4f14b8c5cd3fbed206" }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -596,6 +596,15 @@ alloy-op-hardforks = { git = "https://github.com/ethereum-optimism/optimism", re
...
 # foundry-fork-db = { git = "https://github.com/foundry-rs/foundry-core", rev = "2f90eb86d4549fa15a8cc2d99bfc1039bc083977" }
 
+## foundry-compilers — temporarily pinned to foundry-core `main` (which contains
+## https://github.com/foundry-rs/foundry-core/pull/59) until the next foundry-compilers release.
+## Drop this block and bump the version in `[workspace.dependencies]` once published.
+foundry-compilers = { git = "https://github.com/foundry-rs/foundry-core", rev = "f406192cc95bebde3b6daa74804770d4d03d85c1" }
+foundry-compilers-artifacts = { git = "https://github.com/foundry-rs/foundry-core", rev = "f406192cc95bebde3b6daa74804770d4d03d85c1" }
+foundry-compilers-artifacts-solc = { git = "https://github.com/foundry-rs/foundry-core", rev = "f406192cc95bebde3b6daa74804770d4d03d85c1" }
+foundry-compilers-artifacts-vyper = { git = "https://github.com/foundry-rs/foundry-core", rev = "f406192cc95bebde3b6daa74804770d4d03d85c1" }
+foundry-compilers-core = { git = "https://github.com/foundry-rs/foundry-core", rev = "f406192cc95bebde3b6daa74804770d4d03d85c1" }
```
zerosnacks, line 606: Replace before merging

PR #14676, opened 2026-05-09
feat: improve forge build lint-failure UX

stevencartavia:
`forge build` previously failed lint with an opaque `Error: Lint failed / * solar run failed`.

- Add `--no-lint` (alias `--skip-lint`) for one-shot bypass.
- Capture solar's actual diagnostics into the error via a buffer emitter.
- Print a 3-line notice with bypass instructions and docs link.
- Concise `wrap_err` and clearer `convert_solar_errors` fallback.

### DIFF
--- a/crates/forge/src/cmd/build.rs
+++ b/crates/forge/src/cmd/build.rs
@@ -1,6 +1,6 @@
 use super::{install, watch::WatchArgs};
 use clap::Parser;
-use eyre::{Context, Result};
+use eyre::Result;
 use forge_lint::{linter::Linter, sol::SolidityLinter};
 use foundry_cli::{
     opts::{BuildOpts, configure_pcx_from_solc, get_solar_sources_from_compile_output},
@@ -61,6 +61,14 @@ pub struct BuildArgs {
     #[serde(skip)]
     pub ignore_eip_3860: bool,
 
+    /// Skip the post-build lint step for this invocation.
+    ///
+    /// Equivalent to setting `lint_on_build = false` under `[lint]` in foundry.toml,
+    /// but only for the current command.
+    #[arg(long, visible_alias = "skip-lint")]
+    #[serde(skip)]
+    pub no_lint: bool,
+
     #[command(flatten)]
     #[serde(flatten)]
     pub build: BuildOpts,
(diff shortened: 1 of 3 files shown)


crates/forge/src/cmd/build.rs
```diff
@@ -117,9 +125,16 @@ impl BuildArgs {
...
+        if !self.no_lint
+            && config.lint.lint_on_build
+            && !output.output().errors.iter().any(|e| e.is_error())
+            && let Err(err) = self.lint(&project, &config, self.paths.as_deref(), &mut output)
+        {
+            emit_lint_failure_notice();
+            return Err(err.wrap_err(
+                "post-build lint step failed; rerun with --no-lint or set \
+                 `lint_on_build = false` under `[lint]` in foundry.toml to bypass",
+            ));
```
mablr: There may be a duplicate bypass message ?

crates/forge/src/cmd/build.rs
```diff
@@ -188,8 +203,12 @@ impl BuildArgs {
 
             // NOTE(rusowsky): Once solar can drop unsupported versions, rather than creating a new
             // compiler, we should reuse the parser from the project output.
+            //
+            // Buffer emitter so parse-phase errors surface verbatim in `convert_solar_errors`.
             let mut compiler = solar::sema::Compiler::new(
-                solar::interface::Session::builder().with_stderr_emitter().build(),
+                solar::interface::Session::builder()
+                    .with_buffer_emitter(Default::default())
+                    .build(),
```
mablr: Switching from `with_stderr_emitter()` to `with_buffer_emitter()` means parse-phase diagnostics no longer stream, they batch until after `pcx.parse()` completes. The manual flush block below partially restores this, but only for the no-error path? Any parse warnings emitted inside the `pcx.enter()` closure are silently swallowed if there's also an error. Is this trade-off intentional?

mablr (commented):
Nice, just 2 questions...

crates/forge/src/cmd/build.rs
mablr: let's use import btw :)

stevencartavia: reverted the parse-phase emitter back to `with_stderr_emitter()` and removed the manual flush block

crates/forge/src/cmd/build.rs
```diff
@@ -322,6 +334,21 @@ impl BuildArgs {
     }
 }
 
+/// Notice shown on lint-on-build failure; printed separately so it survives single-line
+/// cause-chain rendering.
+const LINT_FAILURE_NOTICE: &str = "\
+note: post-build lint failed, but compilation succeeded.
+bypass with `--no-lint` or set `lint_on_build = false` under `[lint]` in foundry.toml
+docs: https://getfoundry.sh/forge/linting#disable-linting-on-build
+";
```
zerosnacks, line 347: The CTA should be first to report the crash to foundry-rs/foundry with the debug logs attached / source code snippet

Something along the lines of "This is a bug. Consider reporting it at https://github.com/foundry-rs/foundry" that we use for general panics but more specific.

As a temporary workaround they can bypass; we don't want users permanently disabling lint_on_build.

zerosnacks: Fixed in https://github.com/foundry-rs/foundry/pull/14676/commits/fdb84d27b8930fc4991fdef96b8e5abfb893d870

mablr (dismissed):
sgtm

zerosnacks (approved):
lgtm 👍

mablr (approved):

PR #14677, opened 2026-05-09
feat(lint): Add arbitrary-send-erc20 lint

stevencartavia:
Flags `transferFrom`/`safeTransferFrom` calls whose `from` argument isn't provably `msg.sender` or `address(this)`.


crates/lint/src/sol/high/arbitrary_send_erc20.rs
```diff
@@ -0,0 +1,572 @@
...
+                    && !self.permit_covers(token, from)
+                {
+                    self.hits.push(expr.span);
+                }
+            }
+            // Reassigning a non-state local invalidates any permit referencing it; for
+            // address-typed locals it also updates safety tracking.
+            ExprKind::Assign(lhs, _, rhs) => {
+                if let Some(target) = underlying_var(lhs)
+                    && !self.hir.variable(target).kind.is_state()
```
mablr: state var permit seems not killed (possible false-negative). If token is a state-variable field reassigned between permit and sink, the `PermitRecord` persists and suppresses the warning incorrectly.

crates/lint/testdata/ArbitrarySendErc20.sol
```diff
@@ -0,0 +1,410 @@
...
+
+library SafeERC20 {
+    function safeTransferFrom(
+        IERC20 token,
+        address from,
+        address to,
+        uint256 value
+    ) internal {
+        // Library is a passthrough: warns at *its* call site, not at the library users'.
+        require(token.transferFrom(from, to, value), "SafeERC20: transferFrom failed"); //~WARN: `transferFrom` uses an arbitrary `from`; require it to equal `msg.sender` or `address(this)`
```
mablr: That's inverted? It warns at both the library body and the call site (badLibrary).

stevencartavia: docs updated

mablr (dismissed):
lg

crates/lint/testdata/ArbitrarySendErc20.sol
```diff
@@ -0,0 +1,459 @@
...
+library SafeERC20 {
+    function safeTransferFrom(
+        IERC20 token,
+        address from,
+        address to,
+        uint256 value
+    ) internal {
+        // Library passthrough: warns here AND at the user's call site (see `badLibrary`).
+        require(token.transferFrom(from, to, value), "SafeERC20: transferFrom failed"); //~WARN: `transferFrom` uses an arbitrary `from`; require it to equal `msg.sender` or `address(this)`
+    }
```
zerosnacks, line 53: Hmm.. it is quite common for libraries to pass through args like this so the false positive rate would be quite high

zerosnacks: Perhaps easiest to simple skip if `ContractKind::Library`

crates/lint/src/sol/high/arbitrary_send_erc20.rs
```diff
@@ -0,0 +1,574 @@
...
+            && f.parameters.len() == 4
+            && matches!(
+                hir.variable(f.parameters[0]).ty.kind,
+                TypeKind::Custom(ItemId::Contract(_))
+            )
+            && is_address(hir, f.parameters[1])
+            && is_address(hir, f.parameters[2])
+            && is_elementary(hir, f.parameters[3], "uint256")
+    })
+}
```
zerosnacks: I think this would not support Solady's `safeTransferFrom` where the first param is an address not a IERC20 contract form

https://github.com/Vectorized/solady/blob/7fa3d90c2d0361c2ec2f177bc8902a41dc157f90/src/utils/SafeTransferLib.sol#L249-L272

zerosnacks (dismissed):
Overall lgtm, a few notes

crates/lint/testdata/ArbitrarySendErc20.sol
stevencartavia: skipping it now

crates/lint/src/sol/high/arbitrary_send_erc20.rs
stevencartavia: updated `library_has_safe_transfer_from()`

zerosnacks (dismissed):
lgtm 👍

mablr (dismissed):

crates/lint/src/sol/high/arbitrary_send_erc20.rs
```diff
@@ -0,0 +1,584 @@
...
+            ExprKind::Call(callee, args, _) if is_require_or_assert(callee) => {
+                if let Some(cond) = args.exprs().next() {
+                    self.add_facts(cond, false);
+                }
+            }
+            ExprKind::Call(..) => {
+                if let Some(p) = match_permit_call(expr) {
+                    self.permits.insert(p);
+                } else if let Some((from, token)) = match_sink(self.hir, expr)
+                    && !self.is_safe(from)
```
figtracer, line 565: ERC-3156 flash-lender repayment pattern is a known false positive that we aren't handling here
```
transferFrom(address(receiver), address(this), amount + fee)
```

crates/lint/src/sol/high/arbitrary_send_erc20.rs
```diff
@@ -0,0 +1,584 @@
...
+                let baseline_safe = self.safe_vars.clone();
+                let baseline_permits = self.permits.clone();
+                self.add_facts(cond, false);
+                let _ = self.visit_stmt(then);
+                let then_exits = branch_always_exits(then);
+                let after_then_safe = std::mem::replace(&mut self.safe_vars, baseline_safe);
+                let after_then_permits = std::mem::replace(&mut self.permits, baseline_permits);
+
+                let (after_else_safe, after_else_permits, else_exits) = match else_ {
+                    Some(e) => {
```
figtracer: the explicit `else` branch is visited from the baseline facts, but it should inherit the negated condition. For example, 
```solidity
if (from != msg.sender) 
	revert(); 
else { 
	token.transferFrom(from, to, a); 
}
``` 
is safe but will still warn because `add_facts(cond, true)` only runs for missing `else` branches.

crates/lint/src/sol/high/arbitrary_send_erc20.rs
stevencartavia: added a pre-scan that trusts any `.onFlashLoan()` receiver for the rest of the function (matches the EIP-3156 protocol commitment). Regression added as `okFlashLender`

crates/lint/src/sol/high/arbitrary_send_erc20.rs
stevencartavia: mirrored the implicit fall-through path, regression added as `okExplicitElseGuard`

crates/lint/testdata/UncheckedTransferERC20.sol
```diff
@@ -1,4 +1,4 @@
-//@compile-flags: --severity high info
+//@compile-flags: --only-lint erc20-unchecked-transfer multi-contract-file
```
mablr: Why not only `erc20-unchecked-transfer` ?

mablr (commented):
one smol nit, otherwise sgtm

figtracer (dismissed):

crates/lint/src/sol/high/arbitrary_send_erc20.rs
```diff
@@ -0,0 +1,625 @@
...
+        if let ExprKind::Call(callee, _, _) = &expr.kind
+            && let ExprKind::Member(recv, ident) = &callee.peel_parens().kind
+            && ident.name.as_str() == "onFlashLoan"
+            && let Some(vid) = underlying_var(recv)
+        {
+            self.out.insert(vid);
+        }
+        self.walk_expr(expr)
+    }
+}
```
mablr: These seem to be false negatives:
- `onFlashLoan pre-scanned` over entire body, branch-local call globally whitelists receiver `if (flag) { receiver.onFlashLoan(...); }` then `token.transferFrom(address(receiver), ...)` outside the branch
- Reassigning an interface-typed flash-loan receiver doesn't kill its safe fact: `receiver.onFlashLoan(...); receiver = untrusted` then `token.transferFrom(address(receiver), ...)`
- onFlashLoan exemption is name-only, method named onFlashLoan silences the lint: `IFakeFlashBorrower fake; fake.onFlashLoan(...);` then `token.transferFrom(address(fake), ...)`

mablr (commented):
Tuple reassignment of existing vars seems to be missed:
`address x = msg.sender; (x, y) = (from, to);` then `token.transferFrom(x, ...)`

mablr (dismissed):
All comments seem properly addressed, sgtm.

crates/lint/src/sol/high/arbitrary_send_erc20.rs
```diff
@@ -0,0 +1,635 @@
...
+            }
+
+            _ => {}
+        }
+        self.walk_stmt(stmt)
+    }
+
+    fn visit_expr(&mut self, expr: &'hir hir::Expr<'hir>) -> ControlFlow<Self::BreakValue> {
+        match &expr.kind {
+            ExprKind::Call(callee, args, _) if is_require_or_assert(callee) => {
```
mattsse, line 549: This applies the `require`/`assert` facts before visiting the predicate, which makes the whole condition look guarded even while its subexpressions are being evaluated. That can suppress a real sink that executes before the equality check, for example `require(token.transferFrom(from, to, a) && from == msg.sender);`: the transfer runs before the guard has established anything, but `from` is already in `safe_vars` when the call is visited. The predicate needs expression-order-aware handling, or the facts should only be added after the condition expression has been walked.

crates/lint/src/sol/high/arbitrary_send_erc20.rs
```diff
@@ -0,0 +1,635 @@
...
+                if let Some(cond) = args.exprs().next() {
+                    self.add_facts(cond, false);
+                }
+            }
+            ExprKind::Call(..) => {
```
mattsse, line 558: The flash-loan exemption is broader than the ERC-3156 repayment pattern. Once an exact-signature `onFlashLoan` call is seen, the receiver is treated as safe for every later `transferFrom(address(receiver), ...)`, even if the later pull is not the repayment, uses a different recipient, or transfers an unrelated amount. For example `receiver.onFlashLoan(..., 0, 0, data); token.transferFrom(address(receiver), to, amount);` would be suppressed. This should be tied to the repayment sink shape, token, recipient, and amount relationship rather than whitelisting the receiver variable globally.

crates/lint/src/sol/high/arbitrary_send_erc20.rs
```diff
@@ -0,0 +1,635 @@
...
+fn match_sink<'hir>(
+    hir: &'hir hir::Hir<'hir>,
+    expr: &'hir hir::Expr<'hir>,
+) -> Option<(&'hir hir::Expr<'hir>, Option<hir::VariableId>)> {
+    let ExprKind::Call(callee, args, _) = &expr.kind else { return None };
+    let hir::CallArgsKind::Unnamed(args) = args.kind else { return None };
+    let ExprKind::Member(recv, ident) = &callee.peel_parens().kind else { return None };
+    let name = ident.name.as_str();
+
+    if args.len() == 3 && (name == "transferFrom" || name == "safeTransferFrom") {
```
mattsse: This misses Solady-style extension calls such as `using SafeTransferLib for address; token.safeTransferFrom(from, to, amount);`. In that form the 3-arg member receiver is an `address`, so `receiver_contract_id` returns `None` and the sink is not reported. The PR handles direct `SafeTransferLib.safeTransferFrom(token, ...)`, but this common member-call form still false-negatives.

mattsse (changes requested):
Thanks, this is close, but I think the lint still has a few false-negative paths that should be fixed before merge.

crates/lint/src/sol/high/arbitrary_send_erc20.rs
stevencartavia: addressed in [eace41d](https://github.com/foundry-rs/foundry/pull/14677/commits/eace41d240352cd5779a3dc80ad84ce8154e26f7)

crates/lint/src/sol/high/arbitrary_send_erc20.rs
stevencartavia: addressed in [ba5791c](https://github.com/foundry-rs/foundry/pull/14677/commits/ba5791cf249ab84dc9fd9cd856c1d7cf25c5bad1)

crates/lint/src/sol/high/arbitrary_send_erc20.rs
stevencartavia: addressed in [eace41d](https://github.com/foundry-rs/foundry/pull/14677/commits/eace41d240352cd5779a3dc80ad84ce8154e26f7)

crates/lint/src/sol/high/arbitrary_send_erc20.rs
```diff
@@ -0,0 +1,774 @@
...
+    }
+
+    /// `rhs == None` (unknown slot) drops the target's safe-fact.
+    fn assign_one(&mut self, lhs: &hir::Expr<'_>, rhs: Option<&hir::Expr<'_>>) {
+        let Some(target) = underlying_var(lhs) else { return };
+        self.kill_permits_for(target);
+        self.kill_repayments_for(target);
+        if self.hir.variable(target).kind.is_state() {
+            return;
+        }
```
zerosnacks, line 328: **Blocker (part 2/2).** This early return is correct in spirit (don't trust facts about mutable storage), but combined with `collect_modifier_safety` it is unsound: a state var can be inserted into `safe_vars` via a modifier arg, then reassigned here without being removed, so the stale fact survives until the sink.

Minimum defensive fix here: `self.safe_vars.remove(&target);` before the early return, so the invariant documented at L74-80 (`safe_vars` only holds locals + immutable/constant state) holds regardless of how it got inserted.

crates/lint/src/sol/high/arbitrary_send_erc20.rs
```diff
@@ -0,0 +1,774 @@
...
+    let mut a = Analyzer::new(hir);
+    for stmt in &body.stmts[..placeholder_idx] {
+        let _ = a.visit_stmt(stmt);
+    }
+    for (mp, caller) in arg_map {
+        if a.safe_vars.contains(&mp) {
+            out_safe.insert(caller);
+        }
+    }
+}
```
zerosnacks, line 738: **Blocker (part 1/2).** This block hoists any caller-arg variable into `out_safe` whenever it appears in the modifier's `safe_vars`, without checking that the caller-arg itself is a safe target. Combined with `assign_one`'s early return for state vars (L270-272), a mutable storage var can be marked safe and then reassigned to arbitrary input without the fact being cleared:

```solidity
address owner;
modifier onlySelf(address f) { require(f == msg.sender); _; }
function bad(address from, address to, uint256 a) public onlySelf(owner) {
    owner = from;
    token.transferFrom(owner, to, a); // not flagged
}
```

Fix: only hoist `caller` when `is_safe_target(caller)` is true (i.e. locals + immutable/constant state). The defensive fix in `assign_one` (clear `safe_vars` on state-var writes before the early return) is also worth adding so the invariant holds regardless.

zerosnacks (commented):
Soundness review against the stated contract: flag `transferFrom`/`safeTransferFrom` when `from` is not provably `msg.sender` or `address(this)`.

**Blocker (genuine false negative):**

- **Modifier-hoisted mutable state.** `collect_modifier_safety` can insert any caller-arg variable — including mutable state — into `safe_vars`, and `assign_one` early-returns for state-var writes without clearing the fact. Result: a state var marked safe via a modifier arg stays safe after arbitrary reassignment. Inline comments on `collect_modifier_safety` and `assign_one`.

**Scope note (not a blocker):**

- **`permit` suppression** relaxes the stated contract without proving it (no EIP-2612 signature check on the receiver, no allowance/value modeling). In practice it doesn't mask the canonical arbitrary-send pattern, so this is a precision/contract issue. See inline comment for tightening options or splitting into a separate detector.

zerosnacks (commented):
Follow-up: posting a scope note on the `permit` suppression, separately from the modifier-hoisting blocker.

crates/lint/src/sol/high/arbitrary_send_erc20.rs
```diff
@@ -0,0 +1,774 @@
...
+            ExprKind::Call(..) => {
+                // EIP-3156: a matching repayment sink consumes the recorded obligation.
+                if let Some(rep) = match_flash_loan_call(self.hir, expr) {
+                    self.repayments.insert(rep);
+                } else if let Some(p) = match_permit_call(expr) {
+                    self.permits.insert(p);
+                } else if let Some((from, token)) = match_sink(self.hir, expr)
+                    && !self.is_safe(from)
+                    && !self.permit_covers(token, from)
+                    && !self.consume_repayment(expr, from, token)
```
zerosnacks, line 567: Scope note (not a blocker). `permit_covers` relaxes the stated contract ("provably `msg.sender` or `address(this)`") without proving it: it only tracks `(token_var, owner_var)`, doesn't verify the receiver statically declares an EIP-2612 signature, and doesn't model allowance/value semantics. In practice this doesn't hide the canonical arbitrary-send pattern (which has no preceding permit to suppress on), so it's a precision/contract issue rather than a security gap. Worth either tightening to require the same receiver type for permit and transferFrom, or splitting into a separate detector as Slither does (`arbitrary-send-erc20-permit`). Documenting the suppression in `docs/arbitrary-send-erc20.md` is the minimum.

stevencartavia: [1dfd594](https://github.com/foundry-rs/foundry/pull/14677/commits/1dfd59429fa66175b18ba92bfe74ace1d4e7e895)

crates/lint/src/sol/high/arbitrary_send_erc20.rs
stevencartavia: [5990b29](https://github.com/foundry-rs/foundry/pull/14677/commits/5990b29284f966212e2f033462d3a657f58e143b)

crates/lint/src/sol/high/arbitrary_send_erc20.rs
stevencartavia: [5990b29](https://github.com/foundry-rs/foundry/pull/14677/commits/5990b29284f966212e2f033462d3a657f58e143b)

mattsse:
Found one core false negative:

- **P1:** [`crates/lint/src/sol/high/arbitrary_send_erc20.rs:507`](https://github.com/foundry-rs/foundry/blob/cfa56eb381fd0cdd429b936931095bdf803f1c81/crates/lint/src/sol/high/arbitrary_send_erc20.rs#L507) ignores every `CallArgsKind::Named`, so unsafe calls like `token.transferFrom({from: from, to: to, amount: a})` are missed. Existing lint fixtures already show named ERC20 calls are valid Solidity syntax. Normalize named args for member and library sinks, then add positive named-argument tests.

mablr (commented):
Some curated clanker suggestions, reproduced locally. Probably not all worth to be handled.

## FN-1 · `delete x` doesn't kill safe-var facts

**Location:** [`arbitrary_send_erc20.rs:427–455`](https://github.com/foundry-rs/foundry/blob/66f6bf5ecbc3c148fee6d2cbba5e98749059ce39/crates/lint/src/sol/high/arbitrary_send_erc20.rs#L427-L455)

`handle_assign` is the only write-kill path. `ExprKind::Delete` is not matched, so `delete x` after `x = msg.sender` leaves `x` in `safe_vars`.

**Repro:**
```solidity
address x = msg.sender;
delete x;
token.transferFrom(x, to, a); // no warning — FN
```

---

## FN-2 · `&&` RHS flash-loan repayment record leaks across short-circuit

**Location:** [`arbitrary_send_erc20.rs:409–425`](https://github.com/foundry-rs/foundry/blob/66f6bf5ecbc3c148fee6d2cbba5e98749059ce39/crates/lint/src/sol/high/arbitrary_send_erc20.rs#L409-L425)

The short-circuit rollback saves/restores only `safe_vars`; `permits` and `repayments` are not snapshotted around the RHS visit. A flash-loan call on the RHS of `&&` installs a repayment record even when the RHS is skipped (flag is false).

**Repro:**
```solidity
bool ok = flag && receiver.onFlashLoan(msg.sender, address(token), amount, fee, data) == MAGIC;
token.transferFrom(address(receiver), address(this), amount + fee); // no warning — FN
```

---

## FN-3 · Struct/array/mapping element receivers are not recognised

**Location:** [`arbitrary_send_erc20.rs:519–533`](https://github.com/foundry-rs/foundry/blob/66f6bf5ecbc3c148fee6d2cbba5e98749059ce39/crates/lint/src/sol/high/arbitrary_send_erc20.rs#L519-L533)

`receiver_contract_id` only resolves bare-identifier and type-cast receivers; it returns `None` for member access, index access, and mapping lookup. All three shapes silently miss the sink.

**Repro:**
```solidity
cfg.token.transferFrom(from, to, a);   // no warning — FN
tokens[0].transferFrom(from, to, a);   // no warning — FN
tokenMap[id].transferFrom(from, to, a); // no warning — FN
```

---

## FP-1 · Unrelated `using … for address` library triggers lint

**Location:** [`arbitrary_send_erc20.rs:530–533`](https://github.com/foundry-rs/foundry/blob/66f6bf5ecbc3c148fee6d2cbba5e98749059ce39/crates/lint/src/sol/high/arbitrary_send_erc20.rs#L530-L533)

The `using for address` branch checks only that the receiver is an `address` and that the call has three positional args matching `[from, to, value/amount]` — it never verifies the library is an ERC20 wrapper. Any library bound to `address` with a matching signature fires the lint.

**Repro:**
```solidity
library UnrelatedLib {
    function safeTransferFrom(address token, address from, address to, uint256 amount)
        internal pure returns (uint256) { … }
}
contract C {
    using UnrelatedLib for address;
    address thing;
    function f(address from, address to, uint256 a) public {
        thing.safeTransferFrom(from, to, a); // spurious warning — FP
    }
}
```

---

## FP-2 · `do`-while body facts don't propagate out

**Location:** [`arbitrary_send_erc20.rs:373–376`](https://github.com/foundry-rs/foundry/blob/66f6bf5ecbc3c148fee6d2cbba5e98749059ce39/crates/lint/src/sol/high/arbitrary_send_erc20.rs#L373-L376)

`StmtKind::Loop` uses `visit_isolated`, which discards all state changes from the body. A `do`-while body always executes at least once, so facts established inside should be visible after the loop. Currently they are not, producing a spurious warning.

**Repro:**
```solidity
address x = from;
do { x = msg.sender; } while (false);
token.transferFrom(x, to, a); // spurious warning — FP
```

mablr (dismissed):
Remaining FN reported by clanker are very unlikely to happen in real world, I think it's g2g as-is.

figtracer (dismissed):

mablr (dismissed):
Resolved merge conflict. Pending @mattsse :)

crates/lint/src/sol/high/arbitrary_send_erc20.rs
```diff
@@ -0,0 +1,906 @@
...
+                            .collect(),
+                    ),
+                };
+                self.safe_vars = sv;
+                self.permits = pm;
+                self.repayments = rp;
+                return ControlFlow::Continue(());
+            }
+
+            StmtKind::Loop(block, source) => {
```
figtracer, line 485: `x = from; if (flag) break; x = msg.sender;` leaves `x` unsafe after the loop, but no warning is emitted.

crates/lint/src/sol/high/arbitrary_send_erc20.rs
```diff
@@ -0,0 +1,906 @@
...
+/// SafeTransferLib). Can't go through `contract_has_function` because the first form's
+/// first param isn't elementary.
+fn library_has_safe_transfer_from(hir: &hir::Hir<'_>, cid: hir::ContractId) -> bool {
+    hir.contract_item_ids(cid).any(|item| {
+        let Some(fid) = item.as_function() else { return false };
+        let f = hir.function(fid);
+        if f.parameters.len() != 4 || f.name.is_none_or(|n| n.name.as_str() != "safeTransferFrom") {
+            return false;
+        }
+        let token_ok = matches!(
```
figtracer: this treats any contract-typed `safeTransferFrom` library first arg as ERC20-like, so `SafeERC721.safeTransferFrom(nft, from, to, id)` is falsely reported

stevencartavia: [899015d](https://github.com/foundry-rs/foundry/pull/14677/commits/899015d898f3a6174de5059736516d551b86180f)

crates/lint/src/sol/high/arbitrary_send_erc20.rs
stevencartavia: [899015d](https://github.com/foundry-rs/foundry/pull/14677/commits/899015d898f3a6174de5059736516d551b86180f)

mattsse (approved):
Looks good after the control-flow edge case fixes and focused UI coverage.

PR #14687, opened 2026-05-11
feat(forge): accept noop --no-commit for install

Perico-perica46:
## Motivation

`forge install` no longer commits by default, but scripts and AI tooling can still pass the old `--no-commit` flag and fail during CLI parsing. `forge init` already accepts the same flag as a hidden no-op for backwards compatibility.

## Solution

Add a hidden no-op `--no-commit` flag to `InstallArgs` and leave `--commit` as the only opt-in commit behavior. A focused parser test covers `forge install --no-commit` and verifies it does not enable commit mode.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/forge/src/cmd/install.rs
+++ b/crates/forge/src/cmd/install.rs
@@ -55,12 +55,20 @@ pub struct InstallArgs {
 
     #[command(flatten)]
     opts: DependencyInstallOpts,
+
+    /// Do not create a commit after installing dependencies.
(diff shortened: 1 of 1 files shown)


figtracer:
this is being added here with other additions https://github.com/foundry-rs/foundry/pull/14689, thank you!

PR #14691, opened 2026-05-11
feat(lint): add `optimism-deprecation`

mablr:
## Motivation

Add `optimism-deprecation` that flags usage of Optimism predeploy addresses removed in the Bedrock upgrade and GasPriceOracle functions that revert unconditionally post-Ecotone.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


zerosnacks:
Hmm.. not sure if this is something we would want to include as Bedrock became Optimism's mainnet over 3 years ago and launched a year before that and Ecotone is a really old hardfork: https://github.com/ethereum-optimism/optimism/blob/be742e31ae58ee21dd4431cff839a07c63845570/op-core/forks/forks.go#L13

PR #14697, opened 2026-05-11
feat(lint): add `type-based-tautology`

mablr:
## Motivation

Add `type-based-tautology` lint,  detecting comparison expressions that are always true or always false due to the numeric range of the variable's type.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/lint/src/sol/med/tautology.rs
+++ b/crates/lint/src/sol/med/tautology.rs
@@ -57,7 +57,7 @@ const fn flip(op: BinOpKind) -> BinOpKind {
         BinOpKind::Le => BinOpKind::Ge,
         BinOpKind::Gt => BinOpKind::Lt,
         BinOpKind::Ge => BinOpKind::Le,
-        other => other,
+        _ => unreachable!(),
     }
 }
 
--- a/crates/lint/src/sol/med/tautology.rs
+++ b/crates/lint/src/sol/med/tautology.rs
@@ -1,0 +1,152 @@
+use super::TypeBasedTautology;
+use crate::{
+    linter::{LateLintPass, LintContext},
+    sol::{Severity, SolLint},
+};
+use alloy_primitives::U256;
+use solar::{
+    ast::{BinOpKind, LitKind, UnOpKind},
+    sema::hir::{self, ElementaryType, ExprKind, ItemId, Res, TypeKind},
+};
+
+declare_forge_lint!(
+    TYPE_BASED_TAUTOLOGY,
+    Severity::Med,
+    "type-based-tautology",
+    "condition is always true or false based on the variable's type"
+);
+
+impl<'hir> LateLintPass<'hir> for TypeBasedTautology {
+    fn check_expr(
+        &mut self,
+        ctx: &LintContext,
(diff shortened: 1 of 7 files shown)


zerosnacks:
Overall looks good! Small notes from Amp review:

flip() catch-all arm

- replace other => other with _ => unreachable!() since only Lt|Le|Gt|Ge can reach it

Doc limitation note 

- add a line to type-based-tautology.md stating the lint only fires on local/state variable idents and explicit casts, not struct member access or function return values (or we could address this limitation)

uintLeZero test comment

- consider changing // ok to // ok, equivalent to x == 0, not a tautology to make the intentional exclusion explicitItems 1–2

crates/lint/src/sol/med/tautology.rs
```diff
@@ -0,0 +1,152 @@
...
+        other => other,
+    }
+}
+
+/// Returns true if `var <op> val` is always true or always false for every value in the
+/// type's range.
+///
+/// The constant is represented as a sign bit (`val_neg`) and a magnitude (`val_mag`), matching
+/// how solar stores negated literals (e.g. `-128` -> `Unary(Neg, Lit(128))`).
+fn is_tautology(ty: ElementaryType, val_neg: bool, val_mag: U256, op: BinOpKind) -> bool {
```
figtracer, line 78: we can probably make this more specific since we already know which case it is

crates/lint/src/sol/med/tautology.rs
```diff
@@ -0,0 +1,152 @@
...
+    fn check_expr(
+        &mut self,
+        ctx: &LintContext,
+        hir: &'hir hir::Hir<'hir>,
+        expr: &'hir hir::Expr<'hir>,
+    ) {
+        let ExprKind::Binary(left, op, right) = &expr.kind else { return };
+
+        // Only relational comparisons can produce tautologies via type bounds.
+        if !matches!(op.kind, BinOpKind::Lt | BinOpKind::Le | BinOpKind::Gt | BinOpKind::Ge) {
```
figtracer: do we need eq and ne here too?

crates/lint/src/sol/med/tautology.rs
```diff
@@ -0,0 +1,152 @@
...
+                BinOpKind::Gt | BinOpKind::Le => hi_le_val || val_lt_lo,
+                BinOpKind::Ge | BinOpKind::Lt => val_le_lo || hi_lt_val,
+                _ => false,
+            }
+        }
+        _ => false,
+    }
+}
+
+/// Extracts the elementary integer type from a variable reference or explicit cast.
```
figtracer, line 116: this can probably go into if we decide to do this https://github.com/foundry-rs/foundry/pull/14706 looks generic

figtracer (commented):
nice, some qs

crates/lint/src/sol/med/tautology.rs
mablr: Yep, let's do this as follow-up ?

zerosnacks (approved):
👍 looks good

figtracer (approved):

PR #14698, opened 2026-05-11
fix(cast): include toBlock in chunked logs ranges

gfischer7:
## Motivation
                                                                                                                                                                                                 
  `cast logs` chunked querying has block-range boundary bugs that can produce incorrect results:                                                                                                 
                                                                                                                                                                                                 
  - `fromBlock == toBlock` is currently treated as an empty range.                                                                                                                               
  - Chunking uses half-open ranges and then applies `to_block(chunk_end - 1)`, which can miss the terminal `toBlock`.
                                                                                                                                                                                                 
  Since `eth_getLogs` uses inclusive block boundaries, this can return empty or incomplete log sets.                                                                                             
                                                                                                                                                                                                 
  ## Solution                                                                                                                                                                                    
                                                                                                                                                                                                 
  Switch chunk handling to inclusive range semantics across the chunked path:

  - Return empty only when `fromBlock > toBlock`.                                                                                                                                                
  - Build chunk ranges as inclusive `[start, end]`.                                                                                                                                              
  - Query chunk filters with `.from_block(start).to_block(end)`.                                                                                                                                 
  - Use inclusive fallback iteration (`start..=end`) for per-block retries.                                                                                                                      
  - Add unit tests for:                                                                                                                                                                          
    - terminal `toBlock` inclusion                                                                                                                                                               
    - single-block range handling (`fromBlock == toBlock`)                                                                                                                                       
                                                                                                                                                                                                 
  ## PR Checklist                                                                                                                                                                                
                                                                                                                                                                                                 
  - [x] Added Tests                                                                                                                                                                              
  - [ ] Added Documentation                                                                                                                                                                      
  - [ ] Breaking changes

### DIFF
--- a/crates/cast/src/lib.rs
+++ b/crates/cast/src/lib.rs
@@ -74,6 +74,22 @@ use rlp_converter::Item;
 
 // TODO: CastContract with common contract initializers? Same for CastProviders?
 
+/// Builds inclusive chunk ranges `[start, end]` over `[from, to]`.
+fn inclusive_chunk_ranges(from: u64, to: u64, chunk_size: u64) -> Vec<(u64, u64)> {
+    if from > to {
+        return Vec::new();
+    }
+
+    let chunk_size = chunk_size.max(1);
+    (from..=to)
+        .step_by(chunk_size as usize)
+        .map(|chunk_start| {
+            let chunk_end = chunk_start.saturating_add(chunk_size - 1).min(to);
+            (chunk_start, chunk_end)
+        })
+        .collect()
+}
+
 pub struct Cast<P, N = AnyNetwork> {
     provider: P,
     _phantom: PhantomData<N>,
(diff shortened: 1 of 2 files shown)


crates/cast/src/lib.rs
```diff
@@ -74,6 +74,22 @@ use rlp_converter::Item;
...
+    let chunk_size = chunk_size.max(1);
+    (from..=to)
+        .step_by(chunk_size as usize)
+        .map(|chunk_start| {
+            let chunk_end = chunk_start.saturating_add(chunk_size - 1).min(to);
+            (chunk_start, chunk_end)
+        })
+        .collect()
+}
+
```
mablr: This can be inlined imo

gfischer7: Inlined the chunk range helper into get_logs_chunked_concurrent in 1a773bb64 and removed the standalone helper/tests.

mablr (commented):
Makes sense.

Could you please add a simple `casttest!` validating your fix e2e? (as you removed previous unit tests)

gfischer7:
> Makes sense.
> 
> Could you please add a simple `casttest!` validating your fix e2e? (as you removed previous unit tests)

dded an e2e casttest! in logs_chunked_large_range to validate terminal toBlock inclusion in chunked queries and fromBlock == toBlock behavior, and confirmed it passes with cargo test -p cast--test cli logs_chunked_large_range.

mablr (commented):
sgtm, thanks.

Pending another review.

PR #14700, opened 2026-05-11
fix npm installer HTTPS guard

### LINKED ISSUE #14699: npm installer HTTPS guard accepts insecure registry URLs
## Component

Other: npm package postinstall fallback

## Description

`npm/src/install.mjs` intends to reject non-local `http://` registry and tarball URLs unless `ALLOW_INSECURE_REGISTRY=true` is set. The rejection is currently thrown inside a broad `try` block and then swallowed by the matching `catch`, so the guard silently accepts the same insecure HTTP URLs it is meant to reject.

The affected path is the manual fallback downloader used when the platform-specific optional dependency is unavailable. Registry configuration, registry metadata, and redirects all pass through `makeRequest`, which calls `ensureSecureUrl` before selecting `NodeHttp` or `NodeHttps`.

Minimal reproduction of the current guard behavior:

```js
function ensureSecureUrl(urlString, purpose) {
  try {
    const url = new URL(urlString)
    if (url.protocol === 'http:') {
      const allowInsecure = process.env.ALLOW_INSECURE_REGISTRY === 'true'
      if (!['localhost', '127.0.0.1', '::1'].includes(url.hostname) && !allowInsecure) {
        throw new Error(`Refusing to use insecure HTTP for ${purpose}: ${urlString}`)
      }
    }
  } catch {
  }
}

ensureSecureUrl('http://example.com/@foundry-rs/forge', 'registry URL')
// accepted
```

Expected behavior:

- non-local `http://` URLs are rejected by default
- `http://localhost`, `http://127.0.0.1`, and `http://[::1]` remain valid for local registry development
- `ALLOW_INSECURE_REGISTRY=true` remains an explicit override
- invalid URL strings can still fall through to the request layer as before

## Version

Current `master` at `57755895027876db34b5bf4bcc7473cb949cd9c7`.

## Operating System

All platforms.

afurm:
## Motivation

Closes #14699.

The npm fallback installer intends to reject non-local `http://` registry and tarball URLs unless `ALLOW_INSECURE_REGISTRY=true` is set. The rejection was thrown inside the same broad `try` block that handles URL parse errors, so the `catch` swallowed the policy error and let insecure HTTP requests continue.

## Solution

- Narrow `ensureSecureUrl` so URL parse errors still fall through to the request layer, while insecure HTTP policy errors propagate.
- Keep localhost registry development and the explicit insecure override working.
- Export the helper and guard postinstall execution behind direct module invocation so the invariant can be tested without starting a download.
- Add focused `node:test` coverage for non-local HTTP rejection, localhost allowance, explicit override allowance, HTTPS allowance, and invalid URL passthrough.

Validation:

```sh
node --test npm/src/install.test.mjs
node --check npm/src/install.mjs
node --check npm/src/install.test.mjs
git diff --check
```

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/npm/src/install.mjs
+++ b/npm/src/install.mjs
@@ -6,7 +6,7 @@ import * as NodeHttp from 'node:http'
 import * as NodeHttps from 'node:https'
 import * as NodeModule from 'node:module'
 import * as NodePath from 'node:path'
-import { fileURLToPath } from 'node:url'
+import { fileURLToPath, pathToFileURL } from 'node:url'
 import * as NodeZlib from 'node:zlib'
 
 import { BINARY_NAME, colors, getRegistryUrl, PLATFORM_SPECIFIC_PACKAGE_NAME, resolveTargetTool } from './const.mjs'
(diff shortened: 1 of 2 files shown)


zerosnacks:
We've decided to deprecate the NPM publishing flow, regardless - thanks for your contribution

PR #14706, opened 2026-05-12
feat(lint): add shared analysis helpers for Solidity lints

stevencartavia:
Introduces `crates/lint/src/sol/analysis/`, a small module of reusable HIR helpers that existing/future Solidity lints can build on. Splits the helpers into two narrow modules:

- `primitives` — syntactic/semantic probes (address typing, `require`/`assert` recognition, address-call receivers, exit detection)
- `interface` — contract function-shape matching (elementary-type checks, method-call receiver contract)

Migrates existing lints (`unchecked_calls`, `missing_zero_check`, `incorrect_erc20_interface`, `incorrect_erc721_interface`) to the helpers. Each helper is a free function over borrowed HIR — no traits, no driver, no shared state — to keep the API discoverable and grep-friendly.

### Intentional behavior tightening in `missing_zero_check`

The shared `branch_always_exits` in `primitives.rs` is strictly more correct than the previous inlined version, so this PR is **not** purely a refactor. Two semantic changes (both reducing false positives):

1. **Expression-statement exits are now recognized.** Added a `StmtKind::Expr` arm that detects `revert(...)`, `require(false, ...)`, and `assert(false)` as unconditional aborts when they appear as bare expression statements. The old implementation only handled `StmtKind::Return` / `StmtKind::Revert`, so an `if` arm that aborted via `require(false)` or `assert(false)` was treated as fall-through, producing a spurious `missing-checks-for-zero-address` warning on the variable assigned in the other arm.
2. **Block-exit short-circuits over dead code.** Changed from `block.stmts.last().is_some_and(branch_always_exits)` (block exits only if its **last** statement exits) to `block.stmts.iter().any(branch_always_exits)` (block exits if **any** statement exits). Matches what Solidity itself treats as unreachable code.

`is_require_or_assert` was kept faithful to the original (no `peel_parens()` added), per review feedback.

Regression tests added in `crates/lint/testdata/MissingZeroCheck.sol`:

- `setOwnerRequireFalseInThen` — `require(false, ...)` as expression-stmt exit
- `setOwnerAssertFalseInThen` — `assert(false)` as expression-stmt exit
- `setOwnerReturnThenDeadCode` — block whose first statement is `return` followed by dead code


crates/lint/src/sol/analysis/primitives.rs
```diff
@@ -0,0 +1,144 @@
...
+    match &stmt.kind {
+        StmtKind::Return(_) | StmtKind::Revert(_) => true,
+        StmtKind::Expr(expr) => is_exit_call(expr),
+        StmtKind::Block(b) | StmtKind::UncheckedBlock(b) => {
+            b.stmts.last().is_some_and(branch_always_exits)
+        }
+        StmtKind::If(_, t, Some(e)) => branch_always_exits(t) && branch_always_exits(e),
+        _ => false,
+    }
+}
```
mablr: I think `branch_always_exits` impl may miss:

```solidity
  {
      require(false);
      x = 1; // unreachable
  }
```
while unconditionally exists

stevencartavia: `branch_always_exits` now uses `b.stmts.iter().any(branch_always_exits)` so a block exits as soon as any contained statement does (`require(false); x = 1;` is now correctly recognized).

figtracer (dismissed):
sgtm we may aggregate more on here as time goes on

crates/lint/src/sol/analysis/interface.rs
```diff
@@ -0,0 +1,32 @@
+//! Match contract function shapes (ABI signatures, receiver contract type).
+
+use solar::sema::hir::{self, ContractId, Expr, ExprKind, ItemId, Res, TypeKind, VariableId};
```
mablr, line 3: Amp seems consistently inconsistent on these imports ^^

crates/lint/src/sol/analysis/primitives.rs
```diff
@@ -0,0 +1,73 @@
...
+fn is_exit_call(expr: &Expr<'_>) -> bool {
+    let ExprKind::Call(callee, args, _) = &expr.kind else { return false };
+    if is_builtin(callee, kw::Revert) {
+        return true;
+    }
+    if is_require_or_assert(callee)
+        && let Some(first) = args.exprs().next()
+        && matches!(
+            &first.peel_parens().kind,
+            ExprKind::Lit(lit) if matches!(lit.kind, solar::ast::LitKind::Bool(false))
```
mablr: import at the top

crates/lint/src/sol/analysis/interface.rs
```diff
@@ -0,0 +1,32 @@
...
+                _ => None,
+            },
+            Res::Item(ItemId::Contract(cid)) => Some(*cid),
+            _ => None,
+        }),
+        ExprKind::Call(callee, ..) => match &callee.peel_parens().kind {
+            ExprKind::Ident(reses) => reses.iter().find_map(|r| match r {
+                Res::Item(ItemId::Contract(cid)) => Some(*cid),
+                _ => None,
+            }),
```
mablr: Slight change vs the original logic, seems more permissive (adds peel_parens + find_map over all resolutions instead of matching only the first), please double-check.

mablr:
PR description seems outdated btw

crates/lint/src/sol/analysis/interface.rs
stevencartavia: yes, annoying to miss those. Also consistently inconsistent finding new things from the first commit after several new commits right

mablr (dismissed):
sgtm for now, let's continue to expand it later 👍

crates/lint/src/sol/analysis/primitives.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/// True if `vid` is typed as `address`/`address payable`.
+pub fn is_address_type(hir: &hir::Hir<'_>, vid: VariableId) -> bool {
+    matches!(hir.variable(vid).ty.kind, TypeKind::Elementary(ElementaryType::Address(_)))
+}
+
+/// True if `callee` resolves to the builtin `require` or `assert`.
+pub fn is_require_or_assert(callee: &Expr<'_>) -> bool {
+    matches!(&callee.peel_parens().kind, ExprKind::Ident(reses)
```
zerosnacks: **Behavior change vs. previous inlined version in `missing_zero_check.rs`.**

The old version did not call `peel_parens()`, so `(require)(addr != 0, "...")` was not recognized as a guard. New behavior correctly handles it. Edge case but real — worth either reverting to match the original or calling it out as an intentional improvement.

crates/lint/src/sol/analysis/primitives.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/// True when executing `stmt` provably prevents control from continuing past
+/// it: a `return`, `revert`/`revert(...)`, `require(false, ...)`,
+/// `assert(false)`, a block containing any such statement (any subsequent
+/// statements are unreachable), or an `if` whose both arms exit.
+pub fn branch_always_exits(stmt: &Stmt<'_>) -> bool {
+    match &stmt.kind {
+        StmtKind::Return(_) | StmtKind::Revert(_) => true,
+        StmtKind::Expr(expr) => is_exit_call(expr),
```
zerosnacks, line 52: **Behavior change vs. previous inlined version in `missing_zero_check.rs`.**

This `StmtKind::Expr(...)` arm is new — the old `branch_always_exits` did not handle expression-statement exit calls like `revert();`, `require(false, ...)`, or `assert(false)`. The new behavior is more correct (these are unconditional aborts that should suppress missing-guard warnings on the other branch), but it's a real semantic tightening of `missing_zero_check` that should be either reverted or called out explicitly in the PR description with a regression test.

zerosnacks (commented):
Nice cleanup. Flagging three subtle deltas in the extracted `branch_always_exits` / `is_require_or_assert` helpers that aren't behavior-preserving vs. the inlined originals in `missing_zero_check.rs`. They all look like strict improvements (fewer false positives in `missing_zero_check`), but the PR description says "no behavior changes" — worth either reverting to the exact old shape or calling these out as intentional tightenings + adding a regression test or two.

crates/lint/src/sol/analysis/primitives.rs
```diff
@@ -0,0 +1,74 @@
...
+        StmtKind::Block(b) | StmtKind::UncheckedBlock(b) => b.stmts.iter().any(branch_always_exits),
```
zerosnacks, line 53: **Behavior change vs. previous inlined version in `missing_zero_check.rs`.**

The old version used `block.stmts.last().is_some_and(branch_always_exits)` — block exits only if its **last** statement exits.

The new `iter().any(branch_always_exits)` treats the block as exiting if **any** statement exits. Concrete diff:

```solidity
{
    revert("x");
    doSomething();   // dead code
}
```

- Old: not an exit (last stmt is `doSomething()`)
- New: is an exit (revert short-circuits)

The new behavior is more correct, but it's a real semantic change. Same recommendation — either revert to `last()` to preserve original behavior, or keep it and call it out as an intentional improvement with a regression test.

crates/lint/src/sol/analysis/primitives.rs
stevencartavia: removed `.peel_parens()`

crates/lint/src/sol/analysis/primitives.rs
stevencartavia: added 2 tests, `setOwnerRequireFalseInThen` and `setOwnerAssertFalseInThen`

crates/lint/src/sol/analysis/primitives.rs
stevencartavia: added a new test `setOwnerReturnThenDeadCode`

mablr (approved):
@zerosnacks' comments seem to be properly addressed. PR description lacks this update.

LGTM

zerosnacks (approved):
lgtm, no blockers

PR #14707, opened 2026-05-12
feat(lint): extract shared naming-convention helpers

stevencartavia:
Adds `crates/lint/src/sol/naming.rs` with `check_pascal_case`, `check_mixed_case`, `check_screaming_snake_case`, plus shared underscore-preservation. Migrates the three existing naming lints to use them. No behavior changes


mablr (approved):
Makes sense

zerosnacks (approved):

PR #14714, opened 2026-05-12
fix(anvil): use configured epoch slots for block tags

### LINKED ISSUE #14713: anvil: eth_feeHistory ignores configured epoch slots
### Component

Anvil

- [x] Foundry
- [x] Foundryup

v1.7.0

_No response_

_No response_

### Operating System

None

### Describe the bug

Anvil exposes `--slots-in-an-epoch` / `NodeConfig::with_slots_in_an_epoch`, and most block tag resolution paths use that configured value when resolving `safe` and `finalized`.

However, `eth_feeHistory` still hardcodes `32` slots when `newestBlock` is `safe` or `finalized`.

This means different RPC methods can resolve the same block tag to different block numbers when Anvil is started with a non-default epoch size.

For example, with `--slots-in-an-epoch 3` and latest block `8`:

- `eth_getBlockByNumber("safe", false)` resolves to block `5`
- `eth_feeHistory(1, "safe", [])` resolves using `latest - 32`, so it resolves from block `0`

Expected behavior: all `safe` / `finalized` block tag resolution should use the configured epoch slot value consistently.

haythemsellami:
## Summary
- resolves Anvil safe/finalized block tags with the configured slots-in-an-epoch value in fee history
- passes configured epoch slots through storage block hash lookup
- adds Anvil coverage for custom epoch slots and genesis fallback

Fixes #14713

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -1102,16 +1102,7 @@ impl<N: Network> EthApi<N> {
         node_info!("eth_feeHistory");
         // max number of blocks in the requested range
 
-        let current = self.backend.best_number();
-        let slots_in_an_epoch = 32u64;
-
-        let number = match newest_block {
-            BlockNumber::Latest | BlockNumber::Pending => current,
-            BlockNumber::Earliest => 0,
-            BlockNumber::Number(n) => n,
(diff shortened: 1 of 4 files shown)


figtracer (approved):

mablr (approved):
lgtm, thank you @haythemsellami!

Issue #10285, opened 2025-04-10
feat: enable dynamic test linking by default


grandizzy:
### Component

Forge

### Describe the feature you would like

- turn `dynamic_test_linking` introduced in https://github.com/foundry-rs/foundry/pull/10010 on by default
- review all places where `ProjectCompiler::new` is called and where it should respect the config (e.g. script broadcast will fail with config on as it needs the proper constructor instead the `vm.deployCode`)
- test combinations of `forge test` and `forge script` / other `forge` commands and make sure cache is not invalidated

naijauser:
Can I pick this up? I would like to work on it.

zerosnacks:
I got feedback from @emo-eth at Frontiers that dynamic test linking does not always work yet and if he is able to get a reproducible he will open a ticket

We therefore need to be hesitant enabling by default cc @grandizzy

emo-eth:
@zerosnacks I have a (kinda janky) e2e [Gambit mutation test runner](https://github.com/emo-eth/gambit-runner) - when I wrote it, I had to explicitly disable dynamic test linking in a separate profile for Gambit to get the test suite to pick up changes; the tests suites would all pass since it wouldn't re-compile the mutated source contracts (the runner just copies the entire dir including cache to a temp folder lol)

But on latest stable, it seems to be working correctly!

Will play around with it and let you know if that changes

PR #14717, opened 2026-05-12
fix(common): preserve script CREATE2 broadcast in preprocessor

mablr:
## Motivation

Towards #10285.

Salted `new Contract{salt:...}()` in scripts must not be rewritten by preprocessor when `dynamic-test-linking` is enabled: at broadcast depth, Foundry redirects native CREATE2 through the deterministic factory, preserving the salt `vm.deployCode` runs at a deeper call depth and bypasses that redirect, so the broadcast would record a plain CREATE and deploy at the wrong address.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/common/src/preprocessor/deps.rs
+++ b/crates/common/src/preprocessor/deps.rs
@@ -27,6 +27,7 @@ impl PreprocessorDependencies {
     pub fn new(
         gcx: Gcx<'_>,
         paths: &[PathBuf],
+        script_paths: &HashSet<PathBuf>,
         src_dir: &Path,
         root_dir: &Path,
         mocks: &mut HashSet<PathBuf>,
@@ -81,8 +82,9 @@ impl PreprocessorDependencies {
             // which used to be a mock is refactored to a non-mock implementation).
             mocks.remove(&full_path);
 
+            let is_script = script_paths.contains(path);
             let mut deps_collector =
-                BytecodeDependencyCollector::new(gcx, source.file.src.as_str(), src_dir);
+                BytecodeDependencyCollector::new(gcx, source.file.src.as_str(), src_dir, is_script);
             // Analyze current contract.
             let _ = deps_collector.walk_contract(contract);
             // Ignore empty test contracts declared in source files with other contracts.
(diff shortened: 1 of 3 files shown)


crates/common/src/preprocessor/deps.rs
```diff
@@ -81,8 +82,9 @@ impl PreprocessorDependencies {
             // which used to be a mock is refactored to a non-mock implementation).
             mocks.remove(&full_path);
 
+            let is_script = script_paths.contains(path);
```
grandizzy: I think that's in the right direction but I remember some projects weird setup where scripts were in src? something we may want to check, can we look here maybe at inherited contracts and see if base is script?

mablr: fixed in https://github.com/foundry-rs/foundry/pull/14717/commits/c8b800013d8af1ff61dba893a05131808e59f637, it's maybe a bit broad as it captures by `Script` name, but at least it's safer.

grandizzy: looks good, ty! one last nit - recommend skipping self when checking (that is if contract itself is named `Script`) by `contract.linearized_bases.iter().skip(1)`

grandizzy (approved):
ty, lgtm!

crates/common/src/preprocessor/mod.rs
```diff
@@ -60,6 +60,7 @@ impl Preprocessor<SolcCompiler> for DynamicTestLinkingPreprocessor {
             // Include all sources in the source map so as to not re-load them from disk, but only
             // parse and preprocess tests and scripts.
             let mut preprocessed_paths = vec![];
+            let mut script_paths = HashSet::new();
```
figtracer, line 63: do we need this separate set, can't we infer from `PreprocessorDependencies`?

mablr: `script_paths` is an input to PreprocessorDependencies, not derivable from it, moving the classification inside would couple it to the path resolver. I think it's okay as-is.

zerosnacks (approved):
looks good 👍

PR #14719, opened 2026-05-12
feat(lint): add `uninitialized-local`

mablr:
## Motivation

Add `uninitialized-local` lint that warns if a local variable is read before being explicitly initialized.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


zerosnacks:
Amp flagged a couple of findings to me during review that are worth addressing imo


```
Must-fix details

1. Tuple-destructuring assignment LHS

Where: [uninitialized_local.rs#L141-L151](file:///Users/zerosnacks/Projects/foundry/crates/lint/src/sol/med/uninitialized_local.rs#L141-L151)

Problem: mark_written only matches ExprKind::Ident. Solar represents (a, b) = foo(); as ExprKind::Assign(lhs, None, rhs) where lhs is ExprKind::Tuple(&[Option<&Expr>]). Each element is Some(Ident) for a written slot or None for a skipped slot like (, x) = …. Because the tuple branch is never matched, a and b stay in uninitialized and any later read is wrongly flagged.

Repro:
function f() public returns (uint256) {
    uint256 a;
    uint256 b;
    (a, b) = foo();   // writes both
    return a + b;     // currently warns on both
}

Fix: make mark_written recurse and skip empty slots:
fn mark_written(expr: &Expr<'_>, uninitialized: &mut HashSet<VariableId>) {
    match &expr.kind {
        ExprKind::Ident(reses) => {
            for res in *reses {
                if let Res::Item(ItemId::Variable(vid)) = res {
                    uninitialized.remove(vid);
                }
            }
        }
        ExprKind::Tuple(elems) => {
            for elem in *elems {
                if let Some(e) = elem {
                    mark_written(e, uninitialized);
                }
            }
        }
        _ => {}
    }
}
Add a positive test (tuple LHS writes both) and a negative test ((, x) = foo(); still leaves the unnamed slot alone).

---

2. delete x is not a write

Where: visit_expr in [uninitialized_local.rs](file:///Users/zerosnacks/Projects/foundry/crates/lint/src/sol/med/uninitialized_local.rs)

Problem: ExprKind::Delete(inner) is not matched, so it falls through to walk_expr, which visits inner as a normal Ident. That hits the ExprKind::Ident arm and records inner as a read of an uninitialized variable. But delete is an explicit assignment to the type's zero value — semantically it is a write, not a read.

Repro:
function f() public returns (uint256) {
    uint256 x;
    delete x;     // explicit write to 0
    return x;     // currently warns on `delete x` and on `return x`
}

Fix: in visit_expr, add an arm that mirrors plain Assign semantics — delete has no RHS, so just mark the LHS as written:
ExprKind::Delete(target) => {
    mark_written(target, &mut self.uninitialized);
    // still walk in case it's a complex lhs (member/index)
    let _ = self.visit_expr(target);
    return ControlFlow::Continue(());
}
Add a test asserting uint256 x; delete x; return x; produces no warning.

---

3. Lint fires on reference-type locals

Where: the DeclSingle arm in visit_stmt, [uninitialized_local.rs](file:///Users/zerosnacks/Projects/foundry/crates/lint/src/sol/med/uninitialized_local.rs#L75-L86)

Problem: the lint adds any VarKind::Statement with no initializer to uninitialized. But:
- Storage references (MyStruct storage s;) are a Solidity compile error if uninitialized — the lint can never fire and adds noise.
- Memory/calldata references (bytes memory b;, uint[] memory a;) have well-defined empty defaults; the [docs](file:///Users/zerosnacks/Projects/foundry/crates/lint/docs/uninitialized-local.md) explicitly scope the lint to value types ("address -> address(0), uint -> 0, bool -> false"), but the implementation does not.
- Mappings can't be locals at all.

This produces both false positives (reference defaults are intentional / required) and a docs/behavior mismatch.

Repro:
function f() public pure returns (uint256) {
    uint256[] memory a;     // intentional empty memory array
    return a.length;        // currently warns
}

Fix: filter on the variable's type when deciding whether to track it. Only enroll elementary value types (the ones the docs enumerate):
StmtKind::DeclSingle(vid) => {
    let v = self.hir.variable(*vid);
    let is_value_type = matches!(
        v.ty.kind,
        TypeKind::Elementary(
            ElementaryType::Address(_)
            | ElementaryType::Bool
            | ElementaryType::UInt(_)
            | ElementaryType::Int(_)
            | ElementaryType::FixedBytes(_)
        ),
    );
    if matches!(v.kind, VarKind::Statement) && v.initializer.is_none() && is_value_type {
        self.uninitialized.insert(*vid);
    }
    if let Some(init) = v.initializer {
        let _ = self.visit_expr(init);
    }
    return ControlFlow::Continue(());
}
Add negative tests for bytes memory, string memory, uint[] memory, and a user-defined struct memory to lock the scope down.
```

+

```
4. Diagnostic should point at the read site, not the declaration

Where: [uninitialized_local.rs#L33-L39](file:///Users/zerosnacks/Projects/foundry/crates/lint/src/sol/med/uninitialized_local.rs#L33-L39)

Problem: the lint emits at hir.variable(vid).span (the declaration), but the message says "read before being initialized". In tests like conditional, the actual read is many lines below the underlined declaration, forcing the user to hunt for it. Slither, clippy uninit, and rustc used_assignments all point at the use with the declaration as a secondary note. The current findings: HashSet<VariableId> also throws away the read span before check_function can use it.

Fix:

1. Store the first read span per variable:
findings: FxHashMap<VariableId, Span>,
2. Capture it in the Ident arm:
self.findings.entry(*vid).or_insert(expr.span);
3. Emit at the read, with the declaration as a note (or just at the read if no emit_with_note-style helper exists yet):
for (vid, read_span) in checker.findings {
    ctx.emit_with_note(
        &UNINITIALIZED_LOCAL,
        read_span,
        hir.variable(vid).span,
        "declared without an initializer here",
    );
}
```

crates/lint/src/sol/med/uninitialized_local.rs
```diff
@@ -0,0 +1,177 @@
...
+                for s in block.stmts {
+                    let _ = self.visit_stmt(s);
+                }
+                self.uninitialized = before;
+                return ControlFlow::Continue(());
+            }
+
+            _ => {}
+        }
+        self.walk_stmt(stmt)
```
figtracer, line 142: this probably needs a dedicated `StmtKind::Try` flow merge, similar to how branches are handled

crates/lint/src/sol/med/uninitialized_local.rs
```diff
@@ -0,0 +1,177 @@
...
+                let _ = self.visit_stmt(then);
+                let after_then = self.uninitialized.clone();
+
+                self.uninitialized = before;
+                if let Some(else_stmt) = else_ {
+                    let _ = self.visit_stmt(else_stmt);
+                }
+                let after_else = self.uninitialized.clone();
+
+                self.uninitialized = after_then.union(&after_else).copied().collect();
```
figtracer: `missing_zero_check` has a `branch_always_exits` helper pattern that may be reusable here because this merge does not account for non-fallthrough branches

crates/lint/src/sol/med/uninitialized_local.rs
```diff
@@ -0,0 +1,177 @@
...
+                return ControlFlow::Continue(());
+            }
+
+            // Loops may execute zero times, so writes inside cannot be treated as guaranteed.
+            // Visit the body to catch reads inside, but restore the uninitialized set afterwards.
+            StmtKind::Loop(block, _) => {
```
figtracer: could branch on `LoopSource::DoWhile` instead of discarding writes for all loop sources

figtracer (approved):

zerosnacks (approved):

PR #14723, opened 2026-05-12
feat(lint): add low-level-calls lint

emmajam:
## Summary
- add a high-severity `low-level-calls` lint for direct `.call`, `.delegatecall`, and `.staticcall` usage
- extract the low-level-call matcher for reuse by `unchecked-call`
- add docs, README entry, and UI coverage for checked/unchecked calls, call options, and non-low-level members

## Tests
- `cargo +1.94 test -p forge --test ui -- LowLevelCalls --bless`
- `cargo +1.94 test -p forge --test ui -- lint`
- `rustfmt +1.94 --check --edition 2024 crates/lint/src/sol/high/calls.rs crates/lint/src/sol/high/low_level_calls.rs crates/lint/src/sol/high/mod.rs crates/lint/src/sol/high/unchecked_calls.rs`

CLOSES OSS-30

### DIFF
--- a/crates/lint/src/sol/high/calls.rs
+++ b/crates/lint/src/sol/high/calls.rs
@@ -1,0 +1,25 @@
+use solar::{
+    ast::{Expr, ExprKind},
+    interface::kw,
+};
+
+/// Checks if an expression is a low-level call.
+///
+/// Detects patterns like:
+/// - `target.call(...)`
+/// - `target.delegatecall(...)`
+/// - `target.staticcall(...)`
+/// - `target.call{value: x}(...)`
+pub(super) const fn is_low_level_call(expr: &Expr<'_>) -> bool {
+    if let ExprKind::Call(call_expr, _args) = &expr.kind {
+        let callee = match &call_expr.kind {
+            ExprKind::CallOptions(inner_expr, _) => inner_expr,
+            _ => call_expr,
+        };
+
+        if let ExprKind::Member(_, member) = &callee.kind {
+            return matches!(member.name, kw::Call | kw::Delegatecall | kw::Staticcall);
+        }
+    }
+    false
+}
--- a/crates/lint/src/sol/high/low_level_calls.rs
+++ b/crates/lint/src/sol/high/low_level_calls.rs
@@ -1,0 +1,38 @@
+use super::{LowLevelCalls, calls::is_low_level_call};
+use crate::{
+    linter::{EarlyLintPass, LintContext},
+    sol::{Severity, SolLint},
+};
+use solar::ast::{Expr, ItemFunction, visit::Visit};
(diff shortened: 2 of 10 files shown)


crates/lint/src/sol/high/low_level_calls.rs
```diff
@@ -0,0 +1,38 @@
...
+use crate::{
+    linter::{EarlyLintPass, LintContext},
+    sol::{Severity, SolLint},
+};
+use solar::ast::{Expr, ItemFunction, visit::Visit};
+use std::ops::ControlFlow;
+
+declare_forge_lint!(
+    LOW_LEVEL_CALLS,
+    Severity::High,
```
mablr: I don't think `LOW_LEVEL_CALLS` should be an High

crates/lint/src/sol/high/calls.rs
```diff
@@ -0,0 +1,25 @@
...
+};
+
+/// Checks if an expression is a low-level call.
+///
+/// Detects patterns like:
+/// - `target.call(...)`
+/// - `target.delegatecall(...)`
+/// - `target.staticcall(...)`
+/// - `target.call{value: x}(...)`
+pub(super) const fn is_low_level_call(expr: &Expr<'_>) -> bool {
```
figtracer, line 13: maybe steven did this here https://github.com/foundry-rs/foundry/pull/14706 otherwise cool

crates/lint/src/sol/high/low_level_calls.rs
figtracer: prob informational i would say, it's not really a vulnerability

crates/lint/src/sol/high/low_level_calls.rs
```diff
@@ -0,0 +1,38 @@
...
+use crate::{
+    linter::{EarlyLintPass, LintContext},
+    sol::{Severity, SolLint},
+};
+use solar::ast::{Expr, ItemFunction, visit::Visit};
+use std::ops::ControlFlow;
+
+declare_forge_lint!(
+    LOW_LEVEL_CALLS,
+    Severity::Med,
```
mablr, line 11: In slither this is actually `Info` https://github.com/crytic/slither/blob/c7f761a1564b78409ce5c3b4b0fbec2d39e95368/slither/detectors/operations/low_level_calls.py#L24

@fig was right https://github.com/foundry-rs/foundry/pull/14723#discussion_r3228508225 👍

Then this file should be moved to the correct folder.

emmajam:
Replaced by #14742 from a same-repo branch so CI can access repository secrets. The replacement also addresses the review feedback by classifying `low-level-calls` as `Info` and moving the lint into the info bucket.

PR #14724, opened 2026-05-12
feat(lint): add cache-array-length lint

emmajam:
## Summary

- Add the `cache-array-length` gas lint as a late HIR pass for `for` loop conditions.
- Detect typed dynamic array and `bytes` `.length` reads in comparison expressions, including nested/indexed values and compound conditions.
- Add lint docs, README registration, and UI coverage for flagged and cached-length cases.

## Tests

- `cargo +1.94 check -p forge-lint`
- `cargo +1.94 test -p forge --test ui -- CacheArrayLength`

CLOSES OSS-85

### DIFF
--- a/crates/lint/src/sol/gas/cache_array_length.rs
+++ b/crates/lint/src/sol/gas/cache_array_length.rs
@@ -1,0 +1,197 @@
+use super::CacheArrayLength;
+use crate::{
+    linter::{LateLintPass, LintContext},
+    sol::{Severity, SolLint},
+};
+use solar::{
+    interface::{Symbol, sym},
+    sema::hir::{
+        self, BinOpKind, ElementaryType, ExprKind, ItemId, LoopSource, Res, StmtKind, TypeKind,
+        UnOpKind,
+    },
+};
+
+declare_forge_lint!(
+    CACHE_ARRAY_LENGTH,
+    Severity::Gas,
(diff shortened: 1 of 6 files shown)


crates/lint/src/sol/gas/cache_array_length.rs
```diff
@@ -0,0 +1,197 @@
...
+fn is_array_length_member(hir: &hir::Hir<'_>, expr: &hir::Expr<'_>) -> bool {
+    let ExprKind::Member(base, member) = &expr.kind else { return false };
+    member.name == sym::length && is_array_like(hir, base)
+}
+
+fn is_array_like(hir: &hir::Hir<'_>, expr: &hir::Expr<'_>) -> bool {
+    let Some(ty) = expr_type(hir, expr) else { return false };
+    match &ty.kind {
+        TypeKind::Array(array) => array.size.is_none(),
+        TypeKind::Elementary(ElementaryType::Bytes) => true,
```
mablr, line 157: Do we also need `TypeKind::Elementary(ElementaryType::String) => true,` ?

emmajam:
Replaced by #14741 from a same-repo branch so CI can access repository secrets. Also noted that `ElementaryType::String` is intentionally not handled because Solidity strings do not expose `.length` directly.

PR #14726, opened 2026-05-12
feat(lint): add `dead-code`

figtracer:
## Summary
- Add `dead-code` lint for unreachable internal/private functions.

## Counterpart
- Book PR: https://github.com/foundry-rs/book/pull/1837

### DIFF
--- a/crates/lint/src/sol/codesize/dead_code.rs
+++ b/crates/lint/src/sol/codesize/dead_code.rs
@@ -92,9 +92,9 @@ fn dead_code_analysis(hir: &Hir<'_>) -> Arc<DeadCodeAnalysis> {
     });
 
     let mut cache = ANALYSIS_CACHE.lock().unwrap();
-    // `check_nested_source` is called once per source and constructs fresh pass instances each time,
-    // so cache per-HIR analysis across those calls. Keep it bounded for long-running test processes
-    // that lint many independent HIRs in one process.
+    // `check_nested_source` is called once per source and constructs fresh pass instances each
+    // time, so cache per-HIR analysis across those calls. Keep it bounded for long-running test
+    // processes that lint many independent HIRs in one process.
     if cache.len() >= ANALYSIS_CACHE_LIMIT {
         cache.clear();
     }
--- a/crates/lint/src/sol/codesize/dead_code.rs
+++ b/crates/lint/src/sol/codesize/dead_code.rs
@@ -4,7 +4,7 @@ use crate::{
     sol::{Severity, SolLint},
 };
 use solar::{
-    ast::{ContractKind, ElementaryType, FunctionKind, LitKind, Visibility},
+    ast::{self, ContractKind, ElementaryType, FunctionKind, LitKind, Visibility},
     interface::{Symbol, data_structures::Never, source_map::FileName, sym},
     sema::hir::{
         CallArgs, ContractId, Expr, ExprKind, Function, FunctionId, Hir, ItemId, Modifier, Res,
@@ -37,7 +37,8 @@ impl<'ast> ProjectLintPass<'ast> for DeadCode {
             return;
         }
 
-        let analysis = dead_code_analysis(hir, &input_sources.excluded);
+        let using_for = collect_using_for(hir, sources, &input_sources);
+        let analysis = dead_code_analysis(hir, &input_sources.included, &using_for);
 
         for function_id in hir.function_ids() {
             let function = hir.function(function_id);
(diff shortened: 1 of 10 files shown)


crates/lint/src/sol/codesize/dead_code.rs
```diff
@@ -0,0 +1,287 @@
+use super::DeadCode;
+use crate::{
+    linter::{LateLintPass, LintContext},
+    sol::{Severity, SolLint},
+};
+use solar::{
+    ast::{ContractKind, FunctionKind, Visibility},
+    interface::sym,
+    sema::hir::{self, Visit as _},
```
mablr: For style consistency let's import structs instead of whole `hir` mod.

crates/lint/src/sol/codesize/dead_code.rs
```diff
@@ -0,0 +1,287 @@
...
+);
+
+impl<'hir> LateLintPass<'hir> for DeadCode {
+    fn check_nested_source(
+        &mut self,
+        ctx: &LintContext,
+        hir: &'hir hir::Hir<'hir>,
+        source_id: hir::SourceId,
+    ) {
+        let reachable = Reachability::compute(hir);
```
mablr: This seems redundant as `hir` is global, it makes complexity high `O(sources * fns)`.

Should be computed once and passed in or cached.

crates/lint/src/sol/codesize/dead_code.rs
```diff
@@ -0,0 +1,287 @@
...
+) -> Vec<hir::FunctionId> {
+    hir.contract(contract_id)
+        .all_functions()
+        .filter(|&function_id| {
+            let function = hir.function(function_id);
+            function.name.is_some_and(|ident| ident.name == name)
+                && function.parameters.len() == arity
+        })
+        .collect()
+}
```
mablr, line 566: name + arity is prob not enough as `foo(uint256)` and `foo(address)` could collide.

crates/lint/src/sol/codesize/dead_code.rs
```diff
@@ -0,0 +1,287 @@
...
+        }
+    }
+
+    fn resolve_member_callee(
+        &self,
+        base: &'hir hir::Expr<'hir>,
+        member: solar::interface::Symbol,
+        arity: usize,
+    ) -> Vec<hir::FunctionId> {
+        let hir::ExprKind::Ident(resolutions) = &base.kind else { return Vec::new() };
```
mablr: This seems fragile as it only handles `Ident` base expressions ?

crates/lint/testdata/DeadCode.sol
```diff
@@ -0,0 +1,89 @@
+//@compile-flags: --only-lint dead-code
```
mablr, line 1: Maybe missing test for receive/fallback calling internal helpers?

mablr (commented):
Some nits.

crates/lint/src/sol/codesize/dead_code.rs
```diff
@@ -133,14 +188,50 @@ fn matching_functions(
...
+
+fn same_type(hir: &Hir<'_>, lhs: &Type<'_>, rhs: &Type<'_>) -> bool {
+    match (&lhs.kind, &rhs.kind) {
+        (TypeKind::Elementary(lhs), TypeKind::Elementary(rhs)) => lhs == rhs,
+        (TypeKind::Array(lhs), TypeKind::Array(rhs)) => {
+            same_type(hir, &lhs.element, &rhs.element)
+                && match (lhs.size, rhs.size) {
+                    (None, None) => true,
+                    (Some(lhs), Some(rhs)) => {
+                        format!("{:?}", lhs.kind) == format!("{:?}", rhs.kind)
```
mablr: This pattern is fragile and not efficient.

crates/lint/src/sol/codesize/dead_code.rs
```diff
@@ -21,34 +28,68 @@ impl<'hir> LateLintPass<'hir> for DeadCode {
...
+}
+
+#[derive(Debug)]
+struct DeadCodeAnalysis {
+    reachable: HashSet<FunctionId>,
+    overridden: HashSet<FunctionId>,
+}
+
+static ANALYSIS_CACHE: LazyLock<Mutex<HashMap<AnalysisCacheKey, Arc<DeadCodeAnalysis>>>> =
+    LazyLock::new(|| Mutex::new(HashMap::new()));
```
mablr: missing cache eviction, is that bad ?

crates/lint/src/sol/codesize/dead_code.rs
```diff
@@ -187,61 +279,71 @@ impl<'hir> Reachability<'hir> {
...
+                .iter()
+                .filter_map(|resolution| match resolution {
+                    Res::Item(ItemId::Contract(contract_id)) => Some(*contract_id),
+                    _ => None,
+                })
+                .collect(),
+            _ => Vec::new(),
+        }
+    }
+}
```
mablr, line 787: Seems to only handles direct identifier bases; cast/typeof expressions ignored

crates/lint/src/sol/codesize/dead_code.rs
mablr: I think matching only by name+arity is still problematic

crates/lint/src/sol/codesize/dead_code.rs
```diff
@@ -0,0 +1,441 @@
...
+#[derive(Debug)]
+struct DeadCodeAnalysis {
+    reachable: HashSet<FunctionId>,
+    overridden: HashSet<FunctionId>,
+}
+
+const ANALYSIS_CACHE_LIMIT: usize = 16;
+
+static ANALYSIS_CACHE: LazyLock<Mutex<HashMap<AnalysisCacheKey, Arc<DeadCodeAnalysis>>>> =
+    LazyLock::new(|| Mutex::new(HashMap::new()));
```
mablr: Is there a way to get rid of this global cache? It would be cleaner imo.

figtracer: going to change this to a project pass instead of a late pass as it is now

figtracer:
i think we should not add this lint: too complicated to get right with no false-positives and i'm not sure we want all this complexity just for this tbf

PR #14728, opened 2026-05-12
fix(evm-core): wait for `pending_tasks` before completing `MultiForkHandler`

AissataF:
`MultiForkHandler::poll` returned `Poll::Ready(())` as soon as `handlers` was empty and `incoming` was closed, even when `pending_tasks` still held in-flight `ForkTask::Create` futures. Those tasks register their handler only after they resolve, so an early exit dropped them and left the requesting `sender` waiting forever. Added `pending_tasks.is_empty()` to the exit condition so the future stays alive until every queued fork creation has settled.

### DIFF
--- a/crates/evm/core/src/fork/multi.rs
+++ b/crates/evm/core/src/fork/multi.rs
@@ -490,7 +490,7 @@ impl<
             }
         }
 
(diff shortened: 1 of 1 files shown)


mablr (approved):
makes sense, ty

PR #14735, opened 2026-05-13
fix(anvil): reuse spec blob params for fee history

Perico-perica46:
## Motivation

`FeeHistoryService` duplicated spec-to-blob-params logic and only matched `OSAKA` and `PRAGUE`. Newer specs such as `AMSTERDAM` therefore fell back to Cancun params even though the core blob-param mapping already exposes the
Amsterdam `bpo2` schedule.

## Solution

Add a shared `get_blob_params_by_spec_id` helper, make the update-fraction helper delegate to it, and initialize fee history from the backend's current blob params. Add coverage for the Cancun, Prague, Osaka, and Amsterdam schedules.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -105,7 +105,7 @@ use foundry_evm::{
     },
     utils::{
         block_env_from_header, get_blob_base_fee_update_fraction,
-        get_blob_base_fee_update_fraction_by_spec_id,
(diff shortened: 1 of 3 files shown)


mablr (approved):
sgtm, ty

stevencartavia (approved):

mattsse (approved):

PR #14736, opened 2026-05-13
feat(lint): add locked-ether lint

stevencartavia:
Adds the `locked-ether` lint (Med severity).

Flags contracts that can receive ETH (via `payable` functions, `receive`, or payable `fallback`) but expose no path to send it out.


crates/lint/testdata/LockedEther.sol
```diff
@@ -0,0 +1,155 @@
...
+        selfdestruct(to);
+    }
+}
+
+contract OkNewWithValue {
+    function deposit() external payable {}
+
+    function spawn(uint256 amount) external {
+        new Child{value: amount}();
+    }
```
mablr, line 225: As it calls `Child` which locks a desambiguation comment would nice here.

crates/lint/src/sol/med/locked_ether.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    fn visit_expr(&mut self, expr: &'hir hir::Expr<'hir>) -> ControlFlow<Self::BreakValue> {
+        if expr_sends_ether(expr) {
+            return ControlFlow::Break(());
+        }
+
+        // Queue calls whose callee resolves statically to a `FunctionId`.
+        if let ExprKind::Call(callee, ..) = &expr.kind
+            && let ExprKind::Ident(reses) = &callee.peel_parens().kind
```
mablr: Is there a risk of false-positive here as it seems to miss member-call internal dispatch ?

```solidity
function withdraw(...) external {
    this._doSend(to, amount);
}
```

crates/lint/src/sol/med/locked_ether.rs
```diff
@@ -0,0 +1,181 @@
...
+            .collect();
+
+        while let Some(fid) = worklist.pop() {
+            if !visited.insert(fid) {
+                continue;
+            }
+            let func = hir.function(fid);
+
+            for modifier in func.modifiers {
+                for arg in modifier.args.exprs() {
```
mablr, line 87: modifiers's body should handled like fns ?

crates/lint/src/sol/med/locked_ether.rs
```diff
@@ -0,0 +1,181 @@
...
+            }
+            if matches!(member.name, kw::Delegatecall | kw::Callcode) {
+                return true;
+            }
+        }
+        ExprKind::Ident(reses)
+            if reses.iter().any(|r| matches!(r, Res::Builtin(Builtin::Selfdestruct))) =>
+        {
+            return true;
+        }
```
mablr, line 696: deprecated behavior ?

crates/lint/src/sol/med/locked_ether.rs
stevencartavia: `this.<name>` always targets a `public`/`external` function on self or a base, and we seed the worklist with `linearized_bases.all_functions()`, so it's already analyzed.

crates/lint/src/sol/med/locked_ether.rs
stevencartavia: yes, modifiers are Function items in HIR, so `all_functions()` already covers their bodies

crates/lint/src/sol/med/locked_ether.rs
stevencartavia: post-Cancun it no longer destroys, but still forwards the full balance, remains a valid exit path

crates/lint/testdata/LockedEther.sol
stevencartavia: added a comment

crates/lint/src/sol/med/locked_ether.rs
mablr: Maybe let's add regression test to lock in the behavior?

mablr (commented):
A contract whose only ETH exit is in inline assembly may be flagged.

@stevencartavia In https://github.com/foundry-rs/foundry/pull/14769 I bailed assembly to prevent false positives, wdyt?

crates/lint/src/sol/med/locked_ether.rs
stevencartavia: added :)

stevencartavia:
> A contract whose only ETH exit is in inline assembly may be flagged.
> 
> @stevencartavia In #14769 I bailed assembly to prevent false positives, wdyt?

good, adopted the same `StmtKind::Err` bail

mablr (dismissed):
sgtm (just fixed lint ui test scope)

mablr (dismissed):

crates/lint/src/sol/med/locked_ether.rs
```diff
@@ -0,0 +1,193 @@
...
+        });
+        if !has_payable_entry {
+            return;
+        }
+
+        // Walk every function in `self` and its bases. Internal/library calls resolved to a
+        // `FunctionId` are queued for transitive analysis; unresolved external calls are
+        // conservatively ignored.
+        let mut visited: HashSet<FunctionId> = HashSet::new();
+        let mut worklist: Vec<FunctionId> = contract
```
figtracer: treats unreachable internal helpers as exits, e.g.:

```solidity
receive() external payable {}
function _sweep(address payable to, uint256 amount) internal { 
	to.transfer(amount); 
}
```
no external path reaches `_sweep`, but seeding the worklist with `all_functions()` suppresses `locked-ether`

crates/lint/docs/locked-ether.md
```diff
@@ -0,0 +1,68 @@
...
+   - Any call carrying a non-zero `{value: x}` option, including
+     `addr.call{value: x}(...)` and `new C{value: x}(...)`.
+   - Low-level `addr.delegatecall(...)` / `addr.callcode(...)` (the callee runs in this contract's
+     context and can `selfdestruct`, draining the balance).
+   - The `selfdestruct(addr)` builtin.
+3. Internal and library calls whose callee statically resolves to a function are followed
+   transitively, so a withdrawal helper buried behind several internal hops is detected. External
+   calls through unresolved member access are not followed, to keep false positives down.
+
+If no ETH-sending path is found, the lint reports the contract as locked at the contract's name.
```
mablr: A bit overstating precision? Omits that constructors count as exits, dead helpers satisfy the check, and any inline assembly suppresses it.

crates/lint/src/sol/med/locked_ether.rs
```diff
@@ -0,0 +1,193 @@
...
+            && let ExprKind::Ident(reses) = &callee.peel_parens().kind
+        {
+            for res in *reses {
+                if let Res::Item(ItemId::Function(fid)) = res
+                    && !self.visited.contains(fid)
+                {
+                    self.worklist.push(*fid);
+                }
+            }
+        }
```
mablr: Still unsure about this worklist design, it seems redundant as every fns is preloaded up front.

Joining fig's comment at line 59.

crates/lint/src/sol/med/locked_ether.rs
```diff
@@ -0,0 +1,193 @@
...
+            return;
+        }
+
+        // `receive()` and payable `fallback()` are required to be `Payable`, so a single
+        // mutability check also covers them along with payable constructors and functions.
+        let has_payable_entry = contract.linearized_bases.iter().any(|&cid| {
+            hir.contract(cid)
+                .all_functions()
+                .any(|fid| hir.function(fid).state_mutability == StateMutability::Payable)
+        });
```
mablr: `state_mutability == Payable` seems to over‑approximate receive vectors; an unconditionally reverting `receive()`/`payable` function may still be flagged?

crates/lint/docs/locked-ether.md
stevencartavia: reworded

crates/lint/src/sol/med/locked_ether.rs
stevencartavia: worklist is now seeded only from externally-reachable entry points

crates/lint/src/sol/med/locked_ether.rs
stevencartavia: worklist is now seeded only from externally-reachable entry points

crates/lint/src/sol/med/locked_ether.rs
stevencartavia: `function_always_reverts` excludes payable entries that unconditionally revert — covers `revert()`, `revert("msg")`, `revert CustomError()`, `require(false)`, `assert(false)`, and modifier-induced reverts.

mablr (dismissed):
sgtm

crates/lint/src/sol/med/locked_ether.rs
```diff
@@ -0,0 +1,314 @@
...
+    if let Some(opts) = named_args
+        && opts.iter().any(|arg| arg.name.name == sym::value && !is_literal_zero(&arg.value))
+    {
+        return true;
+    }
+
+    let callee = callee.peel_parens();
+    match &callee.kind {
+        ExprKind::Member(_, member) => {
+            // Single-arg `.transfer`/`.send` to disambiguate from ERC20's 2-arg `transfer`.
```
figtracer, line 666: this can false-negative on userland one-arg `transfer`, e.g. `token.transfer(amount)`; maybe check that the receiver is address/payable

crates/lint/src/sol/med/locked_ether.rs
```diff
@@ -0,0 +1,314 @@
...
+    if func.modifiers.iter().any(|m| {
+        m.id.as_function().is_some_and(|mid| modifier_reverts_before_placeholder(hir.function(mid)))
+    }) {
+        return true;
+    }
+    func.body.is_some_and(|body| stmts_always_revert(body.stmts))
+}
+
+/// Returns `true` if the modifier body always reverts before reaching the `_` placeholder.
+fn modifier_reverts_before_placeholder(modifier: &hir::Function<'_>) -> bool {
```
figtracer: modifier reverts after `_` still make the payable entry reject eth, but this only checks before the placeholder

mattsse (commented):
Reviewed the current head (`a08c2929`). I don't think this is ready to approve yet because the lint can still miss locked contracts in a couple of reachable cases:

- `.transfer` / `.send` are matched by member name and arity only, so a one-arg userland method like `token.transfer(amount)` is treated as an ETH exit even when no ETH can leave the contract.
- `function_always_reverts` only recognizes modifier reverts before `_`; a modifier that runs `_` and then unconditionally reverts still rejects payable entry calls, but the contract is reported as locked.
- `queue_named` follows every overload with the same name for `this` / `super` / contract-name member calls. For example, `super._do(1)` can queue a dead `_do()` overload that transfers ETH and suppress the warning.

I validated the PR builds with `cargo test -p forge-lint`, and also reproduced the edge cases with the PR-built `forge lint`. The first two are already discussed inline, but the overload case looks like another concrete false negative to address or test/declare as an accepted limitation.

crates/lint/src/sol/med/locked_ether.rs
stevencartavia: the new `receiver_is_address()` checks it

crates/lint/src/sol/med/locked_ether.rs
stevencartavia: `modifier_always_reverts` now also checks the suffix after the last `_`

stevencartavia:
> Reviewed the current head (`a08c2929`). I don't think this is ready to approve yet because the lint can still miss locked contracts in a couple of reachable cases:
> 
> * `.transfer` / `.send` are matched by member name and arity only, so a one-arg userland method like `token.transfer(amount)` is treated as an ETH exit even when no ETH can leave the contract.
> * `function_always_reverts` only recognizes modifier reverts before `_`; a modifier that runs `_` and then unconditionally reverts still rejects payable entry calls, but the contract is reported as locked.
> * `queue_named` follows every overload with the same name for `this` / `super` / contract-name member calls. For example, `super._do(1)` can queue a dead `_do()` overload that transfers ETH and suppress the warning.
> 
> I validated the PR builds with `cargo test -p forge-lint`, and also reproduced the edge cases with the PR-built `forge lint`. The first two are already discussed inline, but the overload case looks like another concrete false negative to address or test/declare as an accepted limitation.

All three addressed: receiver-type gate for `.transfer`/`.send`, post-`_` revert check, and `queue_resolved` narrows overloads by arity/named-args and stops at the most-derived contract; regression tests added.

zerosnacks (commented):
Thanks @stevencartavia — nice rule and very thorough fixtures. Sharing review notes from a deeper pass; concerns are around the dispatch model and the "send leaves the contract" classifier. Inline comments below for the specific spots, plus a general note on test-coverage gaps.

**Test coverage gaps worth adding fixtures for** (general, no single line):

- Multi-level and diamond `super` dispatch (call site in a base contract).
- Same-arity overloads that differ only by parameter type (e.g. `f(address)` vs `f(uint256)`).
- Sends targeting `this` / `address(this)` / `payable(address(this))` — should still be flagged as locked.
- `.transfer` / `.send` / `{value: x}` on struct fields, mapping/array elements, function-returned addresses, and ternaries.
- `using for` and function-pointer style dispatch.

Happy to chat further on any of these.

crates/lint/src/sol/med/locked_ether.rs
```diff
@@ -0,0 +1,395 @@
...
+        &mut self,
+        receiver: &hir::Expr<'_>,
+        member: solar::interface::Ident,
+        args: &CallArgs<'_>,
+    ) {
+        let ExprKind::Ident(reses) = &receiver.peel_parens().kind else { return };
+        for res in *reses {
+            match res {
+                Res::Builtin(Builtin::Super) => {
+                    self.queue_resolved(&self.bases[1..], member.name, args);
```
zerosnacks: **Major — `super` dispatch is resolved from the wrong point in the linearization.**

`super` is relative to the contract that *defines the currently visited function/modifier*, not the most-derived contract being linted. This always slices from `self.bases[1..]`, so when the `super.<m>()` call site lives in a base contract (multi-level or diamond inheritance), the lookup starts too high in the C3 order — it can resolve to the wrong override, or to the current function itself and stop. The transitive search then misses the real ETH-sending callee.

Needs the call-site contract id (or its index in `linearized_bases`) threaded through `SendChecker` so `super` slices from the right offset.

crates/lint/src/sol/med/locked_ether.rs
```diff
@@ -0,0 +1,395 @@
...
+    }
+}
+
+/// Returns `true` if `args` can target `params` by arity, and named-arg names map to params.
+fn args_match(hir: &hir::Hir<'_>, args: &CallArgs<'_>, params: &[VariableId]) -> bool {
+    if args.len() != params.len() {
+        return false;
+    }
+    match &args.kind {
+        CallArgsKind::Unnamed(_) => true,
```
zerosnacks: **Major — overload resolution is type-blind.**

For unnamed args this accepts purely on arity; for named args it only checks that the names exist in the parameter list. That means same-arity overloads like `f(address)` and `f(uint256)` are both followed for `f(x)`, and named-arg matching across overloads doesn't ensure the same param set. Any *one* overload that happens to send ETH will then suppress the warning, even when the actually-invoked overload doesn't.

This impacts both [`SendChecker::visit_expr`](#L282-L296) (free/static `Function` resolution) and [`queue_resolved`](#L216-L239) (member-call dispatch). Probably want to use Solar's resolved callee / param types here rather than reimplementing dispatch.

crates/lint/src/sol/med/locked_ether.rs
```diff
@@ -0,0 +1,395 @@
...
+fn expr_sends_ether(hir: &hir::Hir<'_>, expr: &hir::Expr<'_>) -> bool {
+    let ExprKind::Call(callee, args, named_args) = &expr.kind else {
+        return false;
+    };
+
+    // `foo{value: x}(...)` / `new C{value: x}(...)` with `x != 0`.
+    if let Some(opts) = named_args
+        && opts.iter().any(|arg| arg.name.name == sym::value && !is_literal_zero(&arg.value))
+    {
+        return true;
```
zerosnacks: **Major — self-sends are wrongly classified as exits.**

Any non-zero `{value: x}` call is accepted immediately, regardless of receiver. So:

```solidity
this.f{value: x}();
address(this).call{value: x}("");
payable(address(this)).transfer(x);
```

all silence the lint, even though ETH never leaves the contract. Same for `.transfer` / `.send` / `selfdestruct(address(this))` — they pass [`receiver_is_address`](#L350-L369) and short-circuit.

Need an `is_self_address` predicate (`this`, `address(this)`, `payable(address(this))`, etc.) and treat those targets as non-exits.

crates/lint/src/sol/med/locked_ether.rs
```diff
@@ -0,0 +1,395 @@
...
+                kind: TypeKind::Elementary(ElementaryType::Address(_)),
+                ..
+            })
+        ),
+        ExprKind::Member(base, member) => is_address_builtin_member(base, member.name),
+        ExprKind::Ident(reses) => reses.iter().any(|res| match res {
+            Res::Item(ItemId::Variable(id)) => is_address_type(&hir.variable(*id).ty),
+            _ => false,
+        }),
+        _ => false,
```
zerosnacks, line 714: **Major — `receiver_is_address` is too narrow and misses common withdrawal patterns.**

Only literals, `payable(...)`, `address(...)` casts, direct address variables, and a couple of builtins are recognized. Missing:

- Struct fields (`vault.owner.transfer(x)`).
- Mapping/array elements (`recipients[i].send(x)`).
- Function return values (`getTreasury().transfer(x)`, `treasury().delegatecall(data)`).
- Chained member access / ternaries.
- `this` / `super` member returns typed as address.

These all produce false positives — perfectly valid withdraw paths get flagged because their receiver isn't recognized as address-typed. Solar should be able to give the expression's resolved type directly; using that instead of pattern-matching shapes would close most of these gaps in one go.

stevencartavia: [8799f60](https://github.com/foundry-rs/foundry/pull/14736/commits/8799f60a38f163d21922cec7ebdadb96d5c69a9a)

crates/lint/src/sol/med/locked_ether.rs
stevencartavia: [69f43e2](https://github.com/foundry-rs/foundry/pull/14736/commits/69f43e27e43961e415f1280306a22b748549c0f6)

crates/lint/src/sol/med/locked_ether.rs
stevencartavia: [29d8e35](https://github.com/foundry-rs/foundry/pull/14736/commits/29d8e350c6406caee6a2551b34b7e999b4bbeedd)

crates/lint/src/sol/med/locked_ether.rs
stevencartavia: [2e723b1](https://github.com/foundry-rs/foundry/pull/14736/commits/2e723b10d47885b951d4bb90d5632343e50f9ec9)

crates/lint/src/sol/med/locked_ether.rs
```diff
@@ -0,0 +1,566 @@
...
+        ExprKind::Call(callee, ..) => match &callee.peel_parens().kind {
+            // `T(x)` elementary cast.
+            ExprKind::Type(ty) => Some(ty.kind.clone()),
+            // `f(...)` — single-return function call.
+            ExprKind::Ident(reses) => reses.iter().find_map(|res| match res {
+                Res::Item(ItemId::Function(fid)) => single_return_type(hir, *fid),
+                _ => None,
+            }),
+            _ => None,
+        },
```
mablr: For this:

```solidity
contract Registry {
    address payable private _owner;
    function getOwner() external view returns (address payable) { return _owner; }
}

contract MemberCallReceiver {
    Registry registry;

    function deposit() external payable {}

    function withdraw(uint256 x) external {
        registry.getOwner().transfer(x);
    }
}
```

`MemberCallReceiver` is incorrectly flagged despite `registry.getOwner().transfer(x)` being a valid ETH exit. The `ExprKind::Call` arm in `expr_type` only handles `ExprKind::Ident` callees; member-call callees (`ExprKind::Member`) fall through to `_ => None`, so `receiver_is_address` returns false and the transfer is silently dropped.

stevencartavia: [8346901](https://github.com/foundry-rs/foundry/pull/14736/commits/8346901d1f32a99e6728b83c65a0062840bc335a)

mattsse:
Found two false-negative paths:

- **P1:** [`crates/lint/src/sol/med/locked_ether.rs:60`](https://github.com/foundry-rs/foundry/blob/2b9ad56ded2912efacc8dcfb8dce480f10772be5/crates/lint/src/sol/med/locked_ether.rs#L60) seeds the scan with all public/external functions from all linearized bases. If a child overrides `Base.withdraw()` with a no-op, the base implementation is no longer externally dispatchable, but the lint still visits it and suppresses the warning if it sends ETH. Seed only the effective dispatch surface for the concrete contract.

- **P1:** Constructors are treated as general externally reachable withdrawal paths at [`crates/lint/src/sol/med/locked_ether.rs:175`](https://github.com/foundry-rs/foundry/blob/2b9ad56ded2912efacc8dcfb8dce480f10772be5/crates/lint/src/sol/med/locked_ether.rs#L175). A constructor-time transfer can suppress the lint even when the deployed contract later has `receive()` or another payable runtime entry with no withdrawal path. Constructor funding and runtime ETH receipt need separate analysis.

crates/lint/src/sol/med/locked_ether.rs
```diff
@@ -0,0 +1,676 @@
...
+        // plus the most-derived `receive`/`fallback`. Constructors are excluded.
+        let runtime_entries = effective_runtime_dispatch_surface(hir, contract.linearized_bases);
+
+        // Inflow channels are tracked independently: a runtime payable entry vs. a
+        // payable constructor. Constructor inflows are deployer-controlled and have
+        // no runtime exit path, so they must not be conflated with runtime ETH flow.
+        let has_runtime_inflow = runtime_entries.iter().any(|&fid| {
+            let f = hir.function(fid);
+            f.state_mutability == StateMutability::Payable && !function_always_reverts(hir, f)
+        });
```
mablr, line 57: Always-reverting functions/modifiers still count as exits (false negative)

You already have function_always_reverts for inflow filtering here, but it's not applied to exit discovery (L71-L113, L95-L97, L317-L321, L527-L535).

`to.transfer(x); revert();` will silence the lint even though the ETH movement rolls back.

crates/lint/src/sol/med/locked_ether.rs
```diff
@@ -0,0 +1,676 @@
...
+        let has_ctor_inflow = contract.linearized_bases.iter().any(|&cid| {
+            hir.contract(cid).all_functions().any(|fid| {
+                let f = hir.function(fid);
+                matches!(f.kind, FunctionKind::Constructor)
+                    && f.state_mutability == StateMutability::Payable
+            })
+        });
```
mablr: Sets `has_ctor_inflow = true` if any linearized base has a payable ctor. But a non-payable derived ctor rejects deployment value regardless of base. A `Child is Base` with payable `Base()` and non-payable `Child()` gets wrongly flagged. Compute inflow from the effective ctor only?

crates/lint/src/sol/med/locked_ether.rs
stevencartavia: [54a7148](https://github.com/foundry-rs/foundry/pull/14736/commits/54a714836f8fb39e065cd8d3cbbd9eaa57f1367e)

crates/lint/src/sol/med/locked_ether.rs
stevencartavia: [54a7148](https://github.com/foundry-rs/foundry/pull/14736/commits/54a714836f8fb39e065cd8d3cbbd9eaa57f1367e)

mattsse (commented):
Left a few follow-up edge cases on the latest `locked-ether` implementation.

crates/lint/src/sol/med/locked_ether.rs
```diff
@@ -0,0 +1,680 @@
...
+        self.walk_stmt(stmt)
+    }
+
+    fn visit_expr(&mut self, expr: &'hir hir::Expr<'hir>) -> ControlFlow<Self::BreakValue> {
+        if expr_sends_ether(self.hir, expr) {
+            return ControlFlow::Break(());
+        }
+
+        // Queue calls whose callee resolves statically to a `FunctionId`.
+        if let ExprKind::Call(callee, args, _) = &expr.kind {
```
mattsse, line 593: **P1 - inherited external functions can follow an overridden-away internal exit.**

This static `Res::Item(ItemId::Function(fid))` path is anchored to the function id resolved in the function body being visited ([permalink](https://github.com/foundry-rs/foundry/blob/54a714836f8fb39e065cd8d3cbbd9eaa57f1367e/crates/lint/src/sol/med/locked_ether.rs#L526-L540)). If the runtime entry is inherited from a base and it calls an internal `virtual` hook, a leaf override can remove the exit, but the lint still queues the base hook and suppresses the warning.

Repro shape:

```solidity
abstract contract Base {
    function withdraw(address payable to) external { _exit(to); }
    function _exit(address payable to) internal virtual { to.transfer(1 wei); }
}
contract Child is Base {
    receive() external payable {}
    function _exit(address payable) internal override {}
}
```

Suggestion: when visiting an inherited function for a specific leaf contract, resolve unqualified internal calls through the effective override in that leaf's linearization instead of directly queuing the body-local `FunctionId`, or use Solar's resolved dynamic dispatch target if available. This should get a regression test where `Child` is warned.

crates/lint/src/sol/med/locked_ether.rs
```diff
@@ -0,0 +1,680 @@
...
+    let Some(body) = modifier.body else { return false };
+    let Some(first) = body.stmts.iter().position(|s| matches!(s.kind, StmtKind::Placeholder))
+    else {
+        return stmts_always_revert(body.stmts);
+    };
+    let last = body.stmts.iter().rposition(|s| matches!(s.kind, StmtKind::Placeholder)).unwrap();
+    stmts_always_revert(&body.stmts[..first]) || stmts_always_revert(&body.stmts[last + 1..])
+}
+
+fn stmts_always_revert(stmts: &[hir::Stmt<'_>]) -> bool {
```
mattsse, line 153: **P1 - `stmts_always_revert` is path-insensitive and can hide real payable inflows.**

This currently treats a block as always reverting when the last statement reverts ([permalink](https://github.com/foundry-rs/foundry/blob/54a714836f8fb39e065cd8d3cbbd9eaa57f1367e/crates/lint/src/sol/med/locked_ether.rs#L147-L160)). That misses earlier terminating non-revert paths, so a payable entry like this is skipped even though it can accept ETH:

```solidity
contract MaybeAcceptsEth {
    bool open;
    receive() external payable {
        if (open) return;
        revert();
    }
}
```

Suggestion: only classify a statement list as always reverting when every reachable path terminates in a revert. At minimum, handle `return`, `break`/`continue` where relevant, and preceding conditionals before using the final `revert` as proof. Add a regression where a conditional `return` before `revert()` still triggers `locked-ether`.

crates/lint/src/sol/med/locked_ether.rs
```diff
@@ -0,0 +1,680 @@
...
+        return false;
+    };
+    let callee = callee.peel_parens();
+
+    // `foo{value: x}(...)` / `new C{value: x}(...)` with `x != 0`. Targeting `self`
+    // keeps the ETH in this contract, so it is not an exit.
+    if let Some(opts) = named_args
+        && opts.iter().any(|arg| arg.name.name == sym::value && !is_literal_zero(&arg.value))
+    {
+        let self_call =
```
mattsse, line 653: **P2 - self-send detection misses contract/interface casts.**

The `{value: ...}` branch only treats `Member(receiver, _)` as a self-call when `is_self_address(receiver)` recognizes the receiver ([permalink](https://github.com/foundry-rs/foundry/blob/54a714836f8fb39e065cd8d3cbbd9eaa57f1367e/crates/lint/src/sol/med/locked_ether.rs#L581-L584)). `is_self_address` handles `this`, `address(this)`, and `payable(...)`, but not contract/interface casts, so this still counts as an exit even though the value remains in the same contract:

```solidity
interface ISink { function deposit() external payable; }
contract LoopsValue {
    receive() external payable {}
    function deposit() external payable {}
    function loop(uint256 x) external {
        ISink(address(this)).deposit{value: x}();
    }
}
```

Suggestion: teach `is_self_address` to unwrap contract/interface type casts whose single argument is self, or check whether the receiver expression's target address is syntactically `address(this)` after peeling casts. Add this casted-interface form next to the existing self-send regressions.

stevencartavia: [e4b3d9a](https://github.com/foundry-rs/foundry/pull/14736/commits/e4b3d9a844f05481be37609ebd5937eb8baf1db9)

crates/lint/src/sol/med/locked_ether.rs
stevencartavia: [e4b3d9a](https://github.com/foundry-rs/foundry/pull/14736/commits/e4b3d9a844f05481be37609ebd5937eb8baf1db9)

crates/lint/src/sol/med/locked_ether.rs
stevencartavia: [e4b3d9a](https://github.com/foundry-rs/foundry/pull/14736/commits/e4b3d9a844f05481be37609ebd5937eb8baf1db9)

mattsse (approved):
LGTM. I pushed the last follow-up fixes and the focused lint UI test passes.

mablr (approved):
There are still some false positives for payable constructors, but they are very unlikely to occur.

LGTM.

PR #14737, opened 2026-05-13
feat(lint): add external-function lint

stevencartavia:
Adds the `external-function` gas lint: flags `public` functions never called internally that have a reference-type `memory` parameter, suggesting they be declared `external`.


crates/lint/src/sol/gas/external_function.rs
```diff
@@ -0,0 +1,268 @@
...
+                    }
+                }
+            }
+            ExprKind::Member(base, member) => {
+                if let ExprKind::Ident(reses) = &base.peel_parens().kind
+                    && reses.iter().any(|r| matches!(r, Res::Builtin(b) if b.name() == solar::interface::sym::super_))
+                {
+                    self.idx.super_called.insert(member.name);
+                }
+            }
```
mablr, line 180: There is maybe a risk of false-positives here, as we match by name in the whole hir (including libs ?).

I think we should scope to contracts in the inheritance chain of the contract being checked.

crates/lint/src/sol/gas/external_function.rs
```diff
@@ -0,0 +1,293 @@
...
+                let p = hir.variable(pid);
+                p.ty.kind.is_reference_type() && p.data_location == Some(DataLocation::Memory)
+            });
+            if !has_memory_reference_param {
+                continue;
+            }
+
+            // External params live in (immutable) calldata, so a body that writes to a param
+            // cannot be moved to `external` without further refactoring.
+            if body_writes_to_params(hir, &body, func.parameters) {
```
figtracer: misses memory-param writes through internal helper calls, e.g.:

```solidity
function f(Item memory item) public  {
 	_mutate(item); stored = item.value; 
}

function _mutate(Item memory item) internal pure { 
	item.value = 1; 
}
```
public+memory observes the mutation; external+calldata copies into `_mutate`, so following the lint docs changes behavior

zerosnacks (approved):
lgtm 👍 , no serious findings

figtracer (approved):

PR #14739, opened 2026-05-13
feat(config): expose extended Vyper settings

figtracer:
## Summary
- expose new Vyper standard-json settings in foundry.toml config: opt_level, debug, enable_decimals, venom_experimental, and venom pass toggles + allow new optimization values like O1/O2/O3/Os through the existing optimize field

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -1845,9 +1845,14 @@ impl Config {
     /// Returns the configured [VyperSettings] that includes:
     /// - evm version
     pub fn vyper_settings(&self) -> Result<VyperSettings, SolcError> {
+        // Let `opt_level` override `optimize` so child profiles can switch away from an
+        // inherited optimization mode without sending both mutually exclusive settings to Vyper.
+        let optimize = if self.vyper.opt_level.is_some() { None } else { self.vyper.optimize };
+
         Ok(VyperSettings {
             evm_version: Some(self.evm_version),
-            optimize: self.vyper.optimize,
+            optimize,
+            opt_level: self.vyper.opt_level,
             bytecode_metadata: None,
             // TODO: We don't yet have a way to deserialize other outputs correctly, so request only
             // those for now. It should be enough to run tests and deploy contracts.
@@ -1858,7 +1863,10 @@ impl Config {
             ]),
             search_paths: None,
             experimental_codegen: self.vyper.experimental_codegen,
-            ..Default::default()
+            debug: self.vyper.debug,
+            enable_decimals: self.vyper.enable_decimals,
+            venom_experimental: self.vyper.venom_experimental,
+            venom: self.vyper.venom.clone(),
         })
     }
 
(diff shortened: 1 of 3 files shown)


crates/config/src/vyper.rs
```diff
@@ -1,18 +1,35 @@
...
+    pub opt_level: Option<VyperOptimizationLevel>,
     /// The Vyper instance to use if any.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub path: Option<PathBuf>,
     /// Optionally enables experimental Venom pipeline
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub experimental_codegen: Option<bool>,
+    /// Optionally enables experimental Venom pipeline using Vyper's alias.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub venom_experimental: Option<bool>,
```
mablr: `experimental_codegen` vs `venom_experimental` comment, potential ambiguity

crates/config/src/lib.rs
```diff
@@ -1845,9 +1845,13 @@ impl Config {
     /// Returns the configured [VyperSettings] that includes:
     /// - evm version
     pub fn vyper_settings(&self) -> Result<VyperSettings, SolcError> {
+        // Let `opt_level` override an inherited `optimize` from a parent profile.
+        let optimize = if self.vyper.opt_level.is_some() { None } else { self.vyper.optimize };
```
mablr: This silently drops an explicitly configured optimize when opt_level is present. May be problematic when optimize comes from an inherited parent profile, the user gets no feedback. Should emit a `Config` warning ?

crates/config/src/vyper.rs
```diff
@@ -1,18 +1,35 @@
...
+    /// Compile in debug mode.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub debug: Option<bool>,
+    /// Re-enable the Vyper decimal type.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub enable_decimals: Option<bool>,
+    /// Fine-grained Venom optimizer settings.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub venom: Option<VyperVenomSettings>,
```
mablr, line 34: There is no version gating for new settings. `VyperSettings::sanitize()` in foundry-compilers currently only gates `search_paths`. If a user runs with old Vyper version, the compiler may error at invocation rather than giving a helpful message, have you tested this ?

figtracer: vyper takes care of this, if we want better sanitaziation we can add it to `VyperSettings::sanitize()` (but its not urgent, older versions seem to just ignore this)

crates/config/src/lib.rs
figtracer: i think we cant reliably tell if it was inherited because of figment; the current behavior is: if `opt_level` is present, it wins and `optimize` is omitted.

mablr (approved):
sgtm

stevencartavia (approved):

zerosnacks (approved):
lgtm 👍

PR #14741, opened 2026-05-13
feat(lint): add cache-array-length lint

emmajam:
CLOSES OSS-85

## Summary
- add the `cache-array-length` gas lint as a late HIR pass for `for` loop conditions
- detect typed dynamic array and `bytes` `.length` reads in comparison expressions, including nested/indexed values and compound conditions
- add lint docs, README registration, and UI coverage for flagged and cached-length cases

## Notes
- leave out `ElementaryType::String`; Solidity strings do not expose `.length` directly
- replaces #14724 so CI runs from a same-repo branch with repository secrets available

## Validation
- `cargo +1.94 test -p forge --test ui -- CacheArrayLength`


mablr (dismissed):

crates/lint/testdata/CacheArrayLength.sol
```diff
@@ -0,0 +1,96 @@
...
+        uint256 length;
+    }
+
+    Bag internal bag;
+    Counter internal counter;
+
+    function storageArrayLength() external view returns (uint256 sum) {
+        for (uint256 i = 0; i < items.length; ++i) { //~NOTE: array length read in loop condition
+            sum += items[i];
+        }
```
zerosnacks, line 26: There is an issue w/ a false positive if the loop body mutates the array w/ arr.push / array.pop. In those cases the suggested rewrite is not semantically equivalent.

I think we should account for this for this to be mergeable.

crates/lint/docs/cache-array-length.md
```diff
@@ -0,0 +1,56 @@
+# Array length not cached
+
+**Severity**: `Gas`
+**ID**: `cache-array-length`
+
+Flags `for` loop conditions that read a dynamic array or `bytes` value's `.length` on every
```
zerosnacks, line 6: Should we also expand this to cover `while` / `do while`? Not a blocker but the lint is quite general

crates/lint/src/sol/gas/cache_array_length.rs
```diff
@@ -0,0 +1,197 @@
...
+        }
+        ExprKind::Index(base, _) => match &expr_type(hir, base)?.kind {
+            TypeKind::Array(array) => Some(&array.element),
+            TypeKind::Mapping(mapping) => Some(&mapping.value),
+            _ => None,
+        },
+        ExprKind::Member(base, member) => {
+            struct_field_type(hir, expr_type(hir, base)?, member.name)
+        }
+        _ => None,
```
zerosnacks, line 489: any `foo().length` returning a dynamic array is not covered, common example: `bytes(s).length`

crates/lint/docs/cache-array-length.md
```diff
@@ -0,0 +1,56 @@
...
+iteration instead of comparing against a cached local length.
+
+## What it does
+
+Reports comparison expressions in `for` loop conditions when either side reads `.length` from a
+dynamic array-like value, such as `i < values.length`, `values.length > i`, or compound conditions
```
zerosnacks, line 12: Worth explaining why fixed-size arrays are excluded

zerosnacks (changes requested):
Thanks! Some notes, the false positive w/ loop bodies that mutate the array is a blocker

crates/lint/src/sol/gas/cache_array_length.rs
```diff
@@ -0,0 +1,197 @@
...
+    "array length read in loop condition should be cached outside the loop"
+);
+
+impl<'hir> LateLintPass<'hir> for CacheArrayLength {
+    fn check_stmt(
+        &mut self,
+        ctx: &LintContext,
+        hir: &'hir hir::Hir<'hir>,
+        stmt: &'hir hir::Stmt<'hir>,
+    ) {
```
mattsse, line 52: This currently emits for every dynamic array length read in a `for` condition before looking at the loop body. That creates a false positive for storage arrays whose length is intentionally re-read because the body mutates the same array, e.g. `items.push()` or `items.pop()`. In those cases the suggested cached-length rewrite is not semantics-preserving, so this should either skip loops that mutate the referenced array or limit the lint to cases where the length is stable.

mattsse (commented):
One issue from my review: the lint should avoid suggesting cached lengths when the loop body can mutate the same array.

crates/lint/src/sol/gas/cache_array_length.rs
```diff
@@ -0,0 +1,390 @@
...
+
+fn same_expr(lhs: &hir::Expr<'_>, rhs: &hir::Expr<'_>) -> bool {
+    match (&lhs.peel_parens().kind, &rhs.peel_parens().kind) {
+        (ExprKind::Ident(lhs), ExprKind::Ident(rhs)) => {
+            match (resolved_variable(lhs), resolved_variable(rhs)) {
+                (Some(lhs), Some(rhs)) => lhs == rhs,
+                _ => false,
+            }
+        }
+        (ExprKind::Lit(lhs), ExprKind::Lit(rhs)) => lhs.symbol == rhs.symbol,
```
mablr: Is this `symbol`-based matching strong enough ?

crates/lint/src/sol/gas/cache_array_length.rs
```diff
@@ -0,0 +1,390 @@
...
+) -> bool {
+    match &expr.kind {
+        ExprKind::Assign(lhs, _, _) | ExprKind::Delete(lhs) => {
+            reads.iter().any(|read| same_expr(lhs, read.receiver))
+        }
+        ExprKind::Call(callee, _, _) => {
+            let ExprKind::Member(base, member) = &callee.peel_parens().kind else { return false };
+            matches!(member.name, sym::push | kw::Pop)
+                && is_array_like(hir, base)
+                && reads.iter().any(|read| same_expr(base, read.receiver))
```
mablr: It seems the mutation guard only fires when the body writes to the exact receiver expression, a loop could be marked as safe-to-cache when caching would actually change iteration count, leading to a false-positive ?

mablr (dismissed):
sgtm

mattsse (dismissed):
LGTM after the follow-up fixes for hidden mutations and loop-variant length receivers.

mablr (approved):
Reviewed Mattsse's changes and fixed clippy error.

PR #14743, opened 2026-05-13
feat(cast): `KeyAuthorization` signing via non-EOA wallets

mablr:
_Depends on https://github.com/foundry-rs/foundry-core/pull/67 and https://github.com/foundry-rs/foundry-browser-wallet/pull/67 https://github.com/foundry-rs/foundry-browser-wallet/pull/75._

## Motivation

Close OSS-168

This PR wires the Foundry side of browser/passkey-backed `KeyAuthorization` signing into `cast`.

## Changes

- Supports normal root signers and `--browser`/passkey-backed root signers.
- Rename as subcommand as `cast key-authorization` keeps `cast key-auth` as the short alias.
- Emits JSON metadata including signature hash, RLP length, signer, authorized key type, signature type, and witness.

## CLI example

Sign an authorization with the browser wallet:

```bash
cast key-authorization sign \
  0x1111111111111111111111111111111111111111 \
  --chain-id 4217 \
  --key-type secp256k1 \
  --expiry 1782647677 \
  --limit 0x20c0000000000000000000000000000000000000:10000000:1d \
  --scope 0x20c0000000000000000000000000000000000000:transfer \
  --browser \
  --json
```

Manual browser-wallet signing was also tested against `foundry-browser-wallet` running locally with `npm run dev`.


crates/cast/src/cmd/keychain.rs
```diff
@@ -2708,10 +2730,28 @@ fn run_key_auth_encode(args: KeyAuthArgs) -> Result<()> {
...
+    args: KeyAuthArgs,
+    wallet: WalletOpts,
+    browser: BrowserWalletOpts,
+) -> Result<()> {
     let authorization = args.into_authorization()?;
     let authorized_key_type = auth_signature_type_name(&authorization.key_type);
     let signature_hash = authorization.signature_hash();
+
+    if let Some(browser) = browser.run::<TempoNetwork>().await? {
+        let signer_address = browser.address();
```
stevencartavia, line 2754: The browser path makes the connected wallet the signer without checking `--from`, and the local path below doesn't check it either, so `sign --from 0xExpectedRoot` is ignored when a different account signs. The sibling `resolve_keychain_root_signer` guards this with `ensure_root_sender(signer_address, wallet.from)?` in both paths — could we do the same here? (Also, the no-signer error further down doesn't mention `--browser`.)

crates/cast/src/cmd/keychain.rs
```diff
@@ -2708,10 +2730,28 @@ fn run_key_auth_encode(args: KeyAuthArgs) -> Result<()> {
...
+async fn run_key_auth_sign(
+    args: KeyAuthArgs,
+    wallet: WalletOpts,
+    browser: BrowserWalletOpts,
+) -> Result<()> {
     let authorization = args.into_authorization()?;
     let authorized_key_type = auth_signature_type_name(&authorization.key_type);
     let signature_hash = authorization.signature_hash();
+
+    if let Some(browser) = browser.run::<TempoNetwork>().await? {
```
stevencartavia, line 2753: `BrowserSigner::new()` (invoked here via `browser.run`) writes connection status to stdout with `println!`, so those lines get mixed in with the artifact, and under `--json` they precede the JSON object. Could the status output go to stderr instead (the warning already does)? The source is in `foundry-wallets`, shared with other `--browser` commands.

mablr: https://github.com/foundry-rs/foundry-core/pull/81 -> b648a466bd852b91db579de08b214e1f63d687a1 (bump)

crates/cast/src/cmd/keychain.rs
mablr: Fixed in 4dfcc40d938c5c0d94b40e688b0ddb398324ffe0

stevencartavia (dismissed):

stevencartavia:
added the blocked label to wait for https://github.com/foundry-rs/foundry-core/pull/67

crates/cast/src/cmd/keychain.rs
```diff
@@ -2708,10 +2721,27 @@ fn run_key_auth_encode(args: KeyAuthArgs) -> Result<()> {
...
+async fn run_key_auth_sign(
+    args: KeyAuthorizationArgs,
+    wallet: WalletOpts,
+    browser: BrowserWalletOpts,
+) -> Result<()> {
     let authorization = args.into_authorization()?;
     let authorized_key_type = auth_signature_type_name(&authorization.key_type);
     let signature_hash = authorization.signature_hash();
+
+    if let Some(browser) = browser.run::<TempoNetwork>().await? {
```
figtracer, line 2753: `--browser --witness` currently enters a browser flow that cannot carry `witness`, so it fails late with a digest/mutated-payload error. please reject this combination up front until browser signing supports witness passthrough.

Cargo.toml
```diff
@@ -336,7 +336,7 @@ foundry-evm-sancov = { path = "crates/evm/sancov" }
 foundry-evm-traces = { path = "crates/evm/traces", default-features = false }
 foundry-macros = { path = "crates/macros" }
 foundry-test-utils = { path = "crates/test-utils", default-features = false }
-foundry-wallets = { git = "https://github.com/foundry-rs/foundry-core", rev = "db0373422379ab2e98a1c6ba9ed4e8b71c094340", default-features = false }
+foundry-wallets = { git = "https://github.com/foundry-rs/foundry-core", branch = "mablr/sign-authorization-keychain", default-features = false }
```
figtracer: before merge, please replace this moving branch dependency with the merged/released `foundry-core` dependency or a fixed rev/tag.

crates/cast/src/cmd/keychain.rs
```diff
@@ -303,28 +304,31 @@ pub enum KeychainSubcommand {
...
+        #[command(flatten)]
+        browser: BrowserWalletOpts,
     },
 }
 
-/// Common fields for `cast key-auth encode` and `cast key-auth sign`.
+/// Common fields for `cast key-authorization encode` and `cast key-authorization sign`.
 #[derive(Debug, Parser)]
-pub struct KeyAuthArgs {
+pub struct KeyAuthorizationArgs {
```
figtracer, line 331: nearby help text says the authorization is signed by an ethereum secp256k1 wallet, but `--browser` can return p256/webauthn root signatures. please separate authorized key type from root signature type.

Cargo.toml
mablr: fs, that was the point of https://github.com/foundry-rs/foundry/pull/14743#issuecomment-4640753389

crates/cast/src/cmd/keychain.rs
mablr: True, webapp PR was merged before T5 https://github.com/foundry-rs/foundry-browser-wallet/pull/67, I'll follow-up on this.

mablr: https://github.com/foundry-rs/foundry/pull/14743/commits/31919c9982f075208e6d57875c88d87cb49dc1d9

crates/cast/src/cmd/keychain.rs
mablr: The root signature type is determined by the configured signer.

figtracer:
One remaining pre-merge blocker: `foundry-wallets` still points at the moving `mablr/sign-authorization-keychain` branch.

Once foundry-core#67 lands, please switch this to the merged/released dependency, or at least a fixed `rev`/tag, so builds are reproducible.

0xKarl98:
cast keychain authorize --limit now parses TOKEN:AMOUNT[:PERIOD] into TokenLimit.period at crates/cast/src/cmd/keychain.rs:515, but the pre-T3 legacy path later converts those limits into LegacyTokenLimit by copying only token and amount at crates/cast/src/cmd/keychain.rs:2695. 

This means a user can request a periodic limit, have the command succeed, and silently receive a non-periodic legacy authorization. The legacy branch should reject any limit.period != 0 with a clear error that periodic limits require a T3-
capable AccountKeychain.

Suggested as following :
  ```suggestion
  if let Some(limit) = limits.iter().find(|limit| limit.period != 0) {
      eyre::bail!(
          "legacy AccountKeychain authorization does not support periodic limits; \
           remove the period from --limit {}:{}:{} or use a Tempo T3-capable chain",
          limit.token,
          limit.amount,
          limit.period
      );
  }

  let legacy_limits: Vec<LegacyTokenLimit> = limits
      .into_iter()
      .map(|l| LegacyTokenLimit { token: l.token, amount: l.amount })
      .collect();
  ```

0xKarl98:
`KeyAuthorizationArgs::into_authorization` applies `--expiry`, limits, and scopes before setting `is_admin`, so combinations like `--admin --expiry 1` currently succeed and produce `KeyAuthorization { expiry: Some(1), is_admin: true, ... }  

Tempo validation rejects admin authorizations carrying expiry, limits, or allowed call scopes, so `cast key-authorization encode/sign` can emit an artifact that is guaranteed to fail later.

Should reject these flag combinations upfront .

mablr:
@0xKarl98 [d4bd87b](https://github.com/foundry-rs/foundry/pull/14743/commits/d4bd87bbdf9199ca2f2052fea3cf726ca11d66a1) addresses your comments.

pre-T3 fix + out-scoped post-T5 `KeyAuthorization` fields support.

Cargo.toml
```diff
@@ -329,7 +329,7 @@ foundry-evm-symbolic = { path = "crates/evm/symbolic", default-features = false
...
```
grandizzy: are we still waiting for merge?

crates/cast/src/cmd/keychain.rs
```diff
@@ -2785,10 +2828,6 @@ impl KeyAuthArgs {
             authorization = authorization.with_allowed_calls(scopes);
         }
 
-        if let Some(witness) = self.witness {
```
grandizzy, line 2788: is dropping witness support this intentional?

grandizzy (commented):
pls check comments

crates/cast/src/cmd/keychain.rs
```diff
@@ -2726,13 +2759,24 @@ async fn run_key_auth_sign(args: KeyAuthArgs, wallet: WalletOpts) -> Result<()>
     let signer = signer.ok_or_else(|| {
         eyre::eyre!(
             "a persistent root signer is required to sign key authorizations; pass a signer with \
-             --private-key, --keystore, Ledger, Trezor, AWS, GCP, or Turnkey"
+             --browser, --private-key, --keystore, Ledger, Trezor, AWS, GCP, or Turnkey"
         )
     })?;
     let signer_address = signer.address();
+    ensure_root_sender(signer_address, wallet.from)?;
```
figtracer: this rejects mismatched `--from`, but the error still says this is an accountkeychain transaction even though `key-authorization sign` only signs an artifact. can we make the signer mismatch error specific to key authorizations?

Cargo.toml
mablr: pending https://github.com/foundry-rs/foundry-core/pull/67 merge

crates/cast/src/cmd/keychain.rs
mablr: Hmm overshot a bit since browser signer will not support it.

I'll restore this, but it should conflict with `--browser` imo

crates/cast/src/cmd/keychain.rs
mablr: Fixed in 986daa294. `key-authorization sign` now uses a key-authorization-specific root signer check, so mismatched `--from` reports that the key authorization artifact must be signed by the expected root account instead of referring to an AccountKeychain transaction.

crates/cast/src/cmd/keychain.rs
mablr: Restored witness support in 50f5b8699. `cast key-authorization encode/sign` accepts `--witness` again and includes it in the `KeyAuthorization` signing hash. Account/admin remain out of scope; browser signing is safe because the pinned foundry-wallets now rejects witness-bearing browser requests before prompting.

Cargo.toml
mablr: Now pinned on https://github.com/foundry-rs/foundry-core/commit/6bea29247c2fa37344337d7434cd8aaf2f479182 after https://github.com/foundry-rs/foundry-core/pull/67 merge

figtracer (commented):
left two inline comments.

crates/cast/src/cmd/keychain.rs
```diff
@@ -2713,10 +2735,27 @@ fn run_key_auth_encode(args: KeyAuthArgs) -> Result<()> {
...
+async fn run_key_auth_sign(
+    args: KeyAuthorizationArgs,
+    wallet: WalletOpts,
+    browser: BrowserWalletOpts,
+) -> Result<()> {
     let authorization = args.into_authorization()?;
     let authorized_key_type = auth_signature_type_name(&authorization.key_type);
     let signature_hash = authorization.signature_hash();
+
+    if let Some(browser) = browser.run::<TempoNetwork>().await? {
```
figtracer, line 2753: this still starts the browser flow before rejecting `--browser --witness`, so a user can wait for the 5 minute connection timeout before seeing that the combo is unsupported. can we reject this before `browser.run()`?

crates/cast/src/cmd/keychain.rs
```diff
@@ -2726,13 +2765,24 @@ async fn run_key_auth_sign(args: KeyAuthArgs, wallet: WalletOpts) -> Result<()>
...
+    signed: &SignedKeyAuthorization,
+    signature_hash: B256,
+    signer_address: Address,
+    authorized_key_type: &'static str,
+) -> Result<()> {
+    let encoded = encode_key_authorization(signed);
+    let signature_type = auth_signature_type_name(&signed.signature.signature_type());
 
     if shell::is_json() {
         let json = serde_json::json!({
```
figtracer, line 2794: the `--json` output for `sign --witness` drops the witness even though the signed artifact preserves it and the pr description lists witness as metadata. can we include the witness field here and cover it with a json-output test?

crates/cast/src/cmd/keychain.rs
mablr: f7b5eb0fac7639298ce70b0933d3397f68dd4010

crates/cast/src/cmd/keychain.rs
mablr: yep reverted a bit fast

f7b5eb0fac7639298ce70b0933d3397f68dd4010

figtracer (approved):
lgtm

grandizzy (approved):
lgtm

PR #14752, opened 2026-05-14
feat(lint): add delegatecall-loop lint

0xKarl98:
## Motivation
Adds a low-severity Solidity lint matching Slither's `delegatecall-loop` detector. The lint flags `delegatecall` usage inside loops in externally callable payable functions, where repeated delegatecalls can reuse the same `msg.value` and mutate caller storage unexpectedly.

## Solution
- Added a new `delegatecall-loop` low-severity lint.
- Detects `delegatecall` inside `for`, `while`, and `do while` loop bodies.
- Limits detection to `public payable` and `external payable` functions to match Slither's entry-point scope.
- Added docs and UI coverage for positive and negative cases

## Testing
- `cargo fmt --check`
- `cargo test -p forge-lint`
- `cargo test -p forge --test ui DelegatecallLoop`

### DIFF
--- a/crates/lint/src/linter/late.rs
+++ b/crates/lint/src/linter/late.rs
@@ -1,4 +1,7 @@
-use solar::{interface::data_structures::Never, sema::hir};
+use solar::{
+    interface::data_structures::Never,
+    sema::{Gcx, hir},
+};
 use std::ops::ControlFlow;
 
 use super::LintContext;
@@ -58,6 +61,7 @@ pub trait LateLintPass<'hir>: Send + Sync {
     fn check_function(
         &mut self,
         _ctx: &LintContext,
+        _gcx: Gcx<'hir>,
         _hir: &'hir hir::Hir<'hir>,
         _func: &'hir hir::Function<'hir>,
     ) {
@@ -110,6 +114,7 @@ pub trait LateLintPass<'hir>: Send + Sync {
 pub struct LateLintVisitor<'a, 's, 'hir> {
     ctx: &'a LintContext<'s, 'a>,
     passes: &'a mut [Box<dyn LateLintPass<'hir> + 's>],
+    gcx: Gcx<'hir>,
     hir: &'hir hir::Hir<'hir>,
 }
 
@@ -120,9 +125,10 @@ where
     pub fn new(
         ctx: &'a LintContext<'s, 'a>,
         passes: &'a mut [Box<dyn LateLintPass<'hir> + 's>],
+        gcx: Gcx<'hir>,
         hir: &'hir hir::Hir<'hir>,
     ) -> Self {
-        Self { ctx, passes, hir }
+        Self { ctx, passes, gcx, hir }
     }
 }
 
@@ -183,7 +189,7 @@ where
 
     fn visit_function(&mut self, func: &'hir hir::Function<'hir>) -> ControlFlow<Self::BreakValue> {
         for pass in self.passes.iter_mut() {
-            pass.check_function(self.ctx, self.hir, func);
+            pass.check_function(self.ctx, self.gcx, self.hir, func);
         }
         self.walk_function(func)
     }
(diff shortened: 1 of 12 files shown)


mattsse (commented):
I found one detector-scope issue that looks worth addressing before merge.

`crates/lint/src/sol/low/delegatecall_loop.rs:27` starts the visitor from the payable entry point's own AST body, and the emit path at `:57-60` only reports delegatecall expressions that are syntactically under that local loop depth. This misses the same vulnerable shape once either the loop or the delegatecall is hidden behind an internal helper, for example `external payable { for (...) _delegate(payload); }` with `_delegate` performing the `delegatecall`. Since the PR describes this as matching Slither's `delegatecall-loop`, this should either follow internal calls while preserving loop depth or add tests/docs that make the narrower same-body scope explicit.

0xKarl98:
@mattsse This is WIP and trying to make it better ,  and have considered your feedback along the way

crates/lint/src/sol/low/delegatecall_loop.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    fn visit_stmt_with_placeholder(
+        &mut self,
+        stmt: &'hir Stmt<'hir>,
+        placeholder: Option<(&'hir [hir::Modifier<'hir>], usize, hir::Block<'hir>)>,
+    ) {
+        if matches!(stmt.kind, StmtKind::Placeholder) {
+            if let Some((modifiers, index, body)) = placeholder {
+                self.visit_modifier_chain(modifiers, index, body);
+            }
```
mablr: It seems that only top-level `StmtKind::Placeholder` is substitued, we may miss `delegateCall`?

crates/lint/src/sol/low/delegatecall_loop.rs
```diff
@@ -0,0 +1,203 @@
+use super::DelegatecallLoop;
+use crate::{
+    linter::{LateLintPass, LintContext},
+    sol::{Severity, SolLint},
+};
+use solar::{
+    ast::{StateMutability, Visibility},
+    interface::{Span, kw},
+    sema::hir::{self, Expr, ExprKind, FunctionId, ItemId, Res, Stmt, StmtKind, Visit},
```
mablr: For style-consistency with the rest of lints, let's import structs instead of whole `hir` mod.

0xKarl98:
@mablr  have updated according to your feedback

crates/lint/src/sol/low/delegatecall_loop.rs
```diff
@@ -0,0 +1,491 @@
...
+) -> bool {
+    let Some(arg_ty) = expr_ty(gcx, hir, arg) else {
+        return true;
+    };
+    let param_var = hir.variable(param);
+    let param_ty = gcx.type_of_item(param.into()).with_loc_if_ref_opt(gcx, param_var.data_location);
+    arg_ty.convert_implicit_to(param_ty, gcx)
+}
+
+fn expr_ty<'gcx>(gcx: Gcx<'gcx>, hir: &Hir<'gcx>, expr: &Expr<'gcx>) -> Option<Ty<'gcx>> {
```
mablr, line 343: I recall doing something similar in Chisel. We should definitely make this in Solar when we have the time.

crates/lint/src/linter/late.rs
```diff
@@ -62,6 +65,15 @@ pub trait LateLintPass<'hir>: Send + Sync {
         _func: &'hir hir::Function<'hir>,
     ) {
     }
+    fn check_function_with_gcx(
+        &mut self,
+        ctx: &LintContext,
+        _gcx: Gcx<'hir>,
```
mablr, line 71: This seems to pollute a bit `LateLintPass` trait. I guess this a workaround due to lifetimes ?

mablr (commented):
sgtm, just 2 questions...

crates/lint/src/linter/late.rs
0xKarl98: Yes, this was intentional as a narrow compatibility workaround

mablr: Good, no problem. Let’s go if everyone is okay with that.

cc @figtracer @stevencartavia

figtracer: hmmm i guess we can go with it, if there's really no way to just use `check_function` directly

0xKarl98:
pending @mattsse

mattsse (approved):

mattsse (dismissed):
Approving this as an incremental lint improvement. The remaining misses I found are false negatives for less common shapes, not blockers for landing this:

- `(flag ? a : b).delegatecall(...)` because ternary receiver typing is not modeled.
- `super.helper(...)` from inside a loop because internal member-call following does not resolve `super`.

Both can be handled in smaller follow-ups.

PR #14753, opened 2026-05-14
feat(lint): add could-be-constant lint

stevencartavia:
Adds `could-be-constant`,  gas lint that flags state variables with a compile-time-constant inline initializer that are never written. Implemented by extending the existing `could-be-immutable` pass into a single combined analyzer. Also fixes two pre-existing bugs in `is_compile_time_constant` (multi-resolution ident, contract-cast callees) and adds support for `type(T).{min,max,interfaceId}`.


mattsse (commented):
I found one false-positive case in the new `could-be-constant` pass.

`crates/lint/src/sol/gas/immutable.rs:63-67` classifies each state variable initializer as constant or non-constant for that same variable, but the pass never collects writes that happen inside any state-variable initializer before emitting `could-be-constant` at `:121-125`. Solidity accepts `uint256 x = 0; uint256 y = (x = 1);`; this lint would still see `x` as having a constant initializer and no constructor/runtime writes, then suggest `constant`, but applying that suggestion fails with `Cannot assign to a constant variable`. The pass should collect initializer-expression writes into constructor writes, with coverage for assignment/inc/delete side effects in state initializers.

crates/lint/src/sol/gas/immutable.rs
```diff
@@ -391,7 +445,10 @@ fn is_compile_time_constant(hir: &hir::Hir<'_>, expr: &hir::Expr<'_>) -> bool {
 
 fn is_allowed_constant_call(callee: &hir::Expr<'_>) -> bool {
     match &callee.kind {
-        ExprKind::Type(_) => true,
+        // Type casts: `address(0xCAFE)`, `uint256(x)`, etc.
+        ExprKind::Type(_) | ExprKind::TypeCall(_) => true,
```
mablr: `TypeCall(_)` corresponds to `type(...)` pattern,  already handled by `ExprKind::Member(base, member)` in `is_compile_time_constant` ?

At least the comment is misleading, worths double-checking imo.

crates/lint/testdata/CouldBeConstant.sol
```diff
@@ -0,0 +1,66 @@
+//@compile-flags: --only-lint could-be-constant
```
mablr, line 1: Amp oracle suggested additional tests (some of them may be relevant):
- string/bytes with constructor-only write → no could-be-immutable
- string/bytes literal initializer → could-be-constant
- `uint y = (x = 1);` cross-initializer write blocks x from constant
- self-writing initializer uint `x = (x = 1);` — lock down whatever the intended behavior is
- contract cast positive case `IFoo foo = IFoo(address(0x1234));`
- `type(I).interfaceId` positive case

crates/lint/src/sol/gas/immutable.rs
stevencartavia: dropped `TypeCall(_)`

crates/lint/testdata/CouldBeConstant.sol
stevencartavia: already present

crates/lint/src/sol/gas/immutable.rs
```diff
@@ -349,10 +399,10 @@ fn collect_lvalue_writes(
...
     match &expr.kind {
         ExprKind::Lit(_) | ExprKind::Type(_) | ExprKind::TypeCall(_) => true,
-        ExprKind::Ident(resolutions) => resolutions.iter().all(|res| match res {
-            Res::Item(hir::ItemId::Variable(var_id)) => hir.variable(*var_id).is_constant(),
-            _ => false,
-        }),
+        ExprKind::Ident([Res::Item(hir::ItemId::Variable(var_id)), ..]) => {
+            hir.variable(*var_id).is_constant()
+        }
+        ExprKind::Ident(_) => false,
```
mablr: Do we need to change that ? It seems to regress, trailing non-constant resolutions are ignored and can falsely treat ambiguous idents as compile-time constants.

crates/lint/src/sol/gas/immutable.rs
```diff
@@ -96,9 +115,30 @@ impl<'hir> LateLintPass<'hir> for CouldBeImmutable {
...
+
+            // `could-be-immutable`: written in the constructor (body or non-constant initializer)
+            // and never at runtime; type must be immutable-compatible.
+            if !is_immutable_candidate_type(var) {
+                continue;
+            }
+            if non_constant_initializer.contains(&var_id)
+                || constructor_body_writes.contains(&var_id)
+            {
+                ctx.emit(&COULD_BE_IMMUTABLE, span);
```
mablr, line 144: Writes performed inside other state-var initializers seem merged into `constructor_body_writes`, which can wrongly upgrade a constant-initializer var to `could-be-immutable` (e.g. `uint x = 0; uint y = (x = 1);`), please double check.

It worths to add a specific test case for this.

crates/lint/src/sol/gas/immutable.rs
```diff
@@ -377,12 +427,16 @@ fn is_compile_time_constant(hir: &hir::Hir<'_>, expr: &hir::Expr<'_>) -> bool {
         ExprKind::Tuple(exprs) => {
             exprs.iter().flatten().all(|expr| is_compile_time_constant(hir, expr))
         }
+        // `type(T).min`, `type(T).max`, `type(T).interfaceId` are compile-time constants.
+        ExprKind::Member(base, member) => {
+            matches!(base.kind, ExprKind::TypeCall(_))
+                && matches!(member.as_str(), "min" | "max" | "interfaceId")
+        }
```
mablr: This is correct, but could be made tighter: `interfaceId` only for `interfaces` and `min`/`max` only for `integer`/`enum` types.

stevencartavia: nice

mablr (approved):
sgtm

zerosnacks (approved):
lgtm 👍

PR #14758, opened 2026-05-14
Support network hardforks in vm.setEvmVersion

### LINKED ISSUE #14754: vm.setEvmVersion cannot select network-specific execution hardforks
### Component

Forge

- [x] Foundry
- [x] Foundryup

_No response_

_No response_

_No response_

### Operating System

None

### Describe the bug

`vm.setEvmVersion` currently parses its input only as `foundry_compilers::artifacts::EvmVersion`.

That works for Ethereum/compiler-style names such as `istanbul`, `shanghai`, `cancun`, `prague`, or `osaka`, but it does not work for network-specific execution hardforks introduced by Foundry's generic EVM network architecture.

For example, Tempo has network-specific execution hardforks like `T3`, but `vm.setEvmVersion("T3")` cannot select that hardfork because `T3` is not a Solidity compiler `evm_version`.

This makes the cheatcode partially Ethereum-shaped even though the EVM execution stack is now generic over network-specific spec types.

### Expected behavior

When running under a non-Ethereum generic EVM network, `vm.setEvmVersion` should be able to select that network's execution hardfork/spec.

For example, under Tempo it should be possible to select a Tempo hardfork such as `T3` if the active network spec type supports parsing it.

Namespaced hardfork strings such as `tempo:T3` could also be accepted consistently with Foundry's existing hardfork namespace model, while mismatched namespaces should be rejected.

### Actual behavior

`vm.setEvmVersion` only accepts compiler `EvmVersion` strings. Network-specific execution hardfork names are rejected even when the active EVM network supports them.
As a result, generic network support is incomplete: tests cannot switch between network-specific execution hardforks through `vm.setEvmVersion`.

### Possible fix

Parse `vm.setEvmVersion` input through the active generic EVM spec type instead of only through `EvmVersion`.

A generic approach could:

1. Preserve existing compiler `EvmVersion` parsing for Ethereum-style inputs.
2. Allow the active network spec type to parse its own hardfork names.
3. Optionally parse namespaced `FoundryHardfork` values such as `tempo:T3`.
4. Reject namespaced hardforks from the wrong network family.

`vm.getEvmVersion` may also need to report the active network-specific spec name directly instead of converting through an Ethereum `SpecId`, otherwise a network hardfork can be displayed as its Ethereum foundation spec.
```

### PR DESCRIPTION
**Pull request by haythemsellami**
## Summary
- Parse vm.setEvmVersion inputs through the active generic EVM spec type, while preserving existing compiler EVM version support.
- Allow network-specific hardfork names such as Tempo T3 and namespaced forms such as tempo:T2.
- Report the active network-specific spec name from vm.getEvmVersion instead of converting through an Ethereum foundation SpecId.
- Add a Tempo regression test covering unnamespaced and namespaced hardfork selection.

Fixes #14754

### CODE REVIEW

File: crates/evm/hardforks/src/lib.rs [hunk 1]
```diff
@@ -258,6 +258,19 @@ pub fn spec_id_from_optimism_hardfork(hardfork: OpHardfork) -> OpSpecId {
...
+
+    // Parses an unnamespaced hardfork name for the active network.
+    fn from_network_hardfork(_: &str) -> Option<Self> {
+        None
+    }
+
+    // Converts a namespaced Foundry hardfork if it belongs to this spec family.
+    fn from_foundry_hardfork(hardfork: FoundryHardfork) -> Option<Self> {
+        Some(hardfork.into())
+    }
```
mablr: Makes sense, but it feels a bit odd to add this logic to `FromEvmVersion`, it would better imo to rename it or create a new separate one.

haythemsellami: Will a new `ExecutionSpec` be good naming wise?

figtracer: yeah i think so

mablr: yes lg

haythemsellami:
Updated to keep `FromEvmVersion` focused on compiler EVM-version conversion and moved the parse/display methods to a new `ExecutionSpec` trait. `evm_spec_id_from_str` now requires `ExecutionSpec`, while `evm_spec_id` and config EVM-version mapping still use `FromEvmVersion`.

mablr (approved):
lgtm

zerosnacks (approved):
Makes sense 👍

PR #14761, opened 2026-05-14
Make contract size limits configurable

### LINKED ISSUE #14760: Make contract size checks use configurable network limits
### Component

Forge

Foundry's contract size reporting and script deployment warnings still assume Ethereum's size limits in places that are now shared by all network modes.

In current `master`, `forge build --sizes` uses hardcoded limits in `crates/common/src/compile.rs`:

- runtime code: `CONTRACT_RUNTIME_SIZE_LIMIT = 24576` (EIP-170)
- initcode: `CONTRACT_INITCODE_SIZE_LIMIT = 49152` (EIP-3860)

Those constants are used for:

- failure conditions in `SizeReport::exceeds_runtime_size_limit()` and `SizeReport::exceeds_initcode_size_limit()`
- JSON size margins
- table margins and warning colors
- error messages

Separately, `forge script` deployment-size warnings fall back to `CONTRACT_MAX_SIZE` from `foundry-common` unless `code_size_limit` is explicitly set.

This is correct for Ethereum, but Foundry now has network-aware execution paths (`NetworkConfigs`, `FoundryEvmNetwork`, network-specific EVM factories). Some networks can have protocol code-size or initcode-size limits that differ from Ethereum. The execution layer can enforce those network-specific limits, but the compile/reporting layer cannot currently consume them.

Impact:

- `forge build --sizes` can fail or report negative margins for contracts that are valid on a selected non-Ethereum network.
- JSON and table size output always reports Ethereum margins, even when another network's limits should apply.
- Network integrations have to either hardcode downstream changes, disable size reporting, or ask users to ignore warnings that should be computed from the selected network.
- `forge script` has a manual `--code-size-limit` workaround for its deployment warning path, but this does not provide network defaults and does not help `forge build --sizes`.
- `--ignore-eip-3860` only bypasses the initcode failure and does not address runtime size limits, displayed margins, or network-specific defaults.

Expected behavior:

Foundry should have a generic contract-size limit model that defaults to Ethereum's current values, while allowing callers/network configuration to provide different runtime and initcode limits.

A possible shape:

- Add a small `ContractSizeLimits` type with Ethereum defaults:
  - `runtime = 24576`
  - `initcode = 49152`
- Let `ProjectCompiler` accept size limits, defaulting to Ethereum when unset.
- Make `SizeReport` use the configured limits for failures, JSON output, table margins, and colors.
- Make script deployment-size warnings use the same configured/default runtime limit instead of directly falling back to `CONTRACT_MAX_SIZE`.

This should preserve existing Ethereum behavior while giving network integrations a clean upstream extension point.

haythemsellami:
## Summary
- Add reusable runtime/initcode contract size limits for compiler size reports.
- Thread configured `code_size_limit` into `forge build --sizes`, deriving the initcode limit from the runtime limit.
- Use the shared limit model in script deployment-size warnings.

Fixes #14760


haythemsellami:
@figtracer or @mablr would appreciate a review here, let me know if this is not the right approach :)

mablr (commented):
This makes sense imo.

haythemsellami:
@figtracer @mablr Can we get this merged?

mattsse (approved):
LGTM

mablr (approved):
lgtm

sakulstra:
small issue i have with this, is that is seems to apply also for tests.

Wanted to play with amsterdam, but selecting amsterdam evm_version does not increase code-size, and adjusting code size breaks tests that exceed it 😅

PR #14763, opened 2026-05-14
Make verify-bytecode replay use EVM network generics

### LINKED ISSUE #14762: Make forge verify-bytecode replay use FoundryEvmNetwork generics
### Component

Forge

`forge verify-bytecode` currently replays deployment bytecode with `TracingExecutor<EthEvmNetwork>` directly, even though Foundry v1.7 has generalized execution through `FoundryEvmNetwork` / `FoundryEvmFactory`.

This means `verify-bytecode` cannot reuse the same network-generic execution path that Forge, Cast, Script, and Anvil are moving toward. It also makes chain-specific EVM implementations harder to support correctly in bytecode verification, because the local runtime bytecode comparison is produced by Ethereum execution regardless of the selected network family.

### Additional context

_No response_

haythemsellami:
- factors verify-bytecode execution into a network-generic internal entry point while keeping Ethereum as the default path
- makes the tracing executor, deploy replay, block env configuration, and runtime-code helpers generic over FoundryEvmNetwork for standard tx/block env networks
- uses the selected network spec type for local replay

Fixes #14762


crates/verify/src/bytecode.rs
```diff
@@ -127,7 +134,15 @@ impl figment::Provider for VerifyBytecodeArgs {
...
-    pub async fn run(mut self) -> Result<()> {
+    pub async fn run(self) -> Result<()> {
+        self.run_with_network::<EthEvmNetwork>().await
+    }
+
+    /// Run the `verify-bytecode` command with the selected EVM network implementation.
+    pub async fn run_with_network<FEN>(mut self) -> Result<()>
+    where
+        FEN: FoundryEvmNetwork<EvmFactory: FoundryEvmFactory<BlockEnv = BlockEnv, Tx = TxEnv>>,
+    {
```
mablr: The `FEN` bound should not be used like this.

`FoundryEvmNetwork` supertrait must be used as-is, it is implemented for `EthEvmNetwork`, `OpEvmNetwork` and `TempoEvmNetwork`
https://github.com/foundry-rs/foundry/blob/c843fdcec81b6b5ca77a35880bbd9a1ac9cd245a/crates/evm/core/src/evm/mod.rs#L63-L75

crates/verify/src/utils.rs
```diff
@@ -289,20 +298,29 @@ pub async fn get_tracing_executor(
...
-pub fn configure_env_block(evm_env: &mut EvmEnv, block: &AnyRpcBlock, config: NetworkConfigs) {
+pub fn configure_env_block<FEN>(
+    evm_env: &mut EvmEnvFor<FEN>,
+    block: &AnyRpcBlock,
+    config: NetworkConfigs,
+) where
+    FEN: FoundryEvmNetwork<EvmFactory: FoundryEvmFactory<BlockEnv = BlockEnv>>,
+{
     let number = evm_env.block_env.number;
     evm_env.block_env = block_env_from_header(&block.header);
```
mablr: `block_env_from_header` is generic, so constraining `FoundryEvmFactory<BlockEnv = BlockEnv>` is a footgun.

mablr (changes requested):
I'm okay with the target 👍 , but the `FoundryEvmNetwork` usage should be revised.

haythemsellami:
Updated this to use `FoundryEvmNetwork` as-is instead of constraining the associated EVM factory types. The replay path now builds a typed provider for `FEN::Network`, uses `TxEnvFor<FEN>`/`BlockEnvFor<FEN>`, calls generic `block_env_from_header`, and applies chain/block env changes through `FEN::Network`.

crates/verify/src/bytecode.rs
```diff
@@ -127,15 +134,23 @@ impl figment::Provider for VerifyBytecodeArgs {
 impl VerifyBytecodeArgs {
     /// Run the `verify-bytecode` command to verify the bytecode onchain against the locally built
     /// bytecode.
-    pub async fn run(mut self) -> Result<()> {
+    pub async fn run(self) -> Result<()> {
+        self.run_with_network::<EthEvmNetwork>().await
+    }
```
mablr, line 186: Let's introduce network dispatch with `NetworkVariant` arg and `ChainId` obtained from RPC.

mablr (commented):
Nice, much better!

crates/verify/src/bytecode.rs
mablr: similarly to what we have in cast/forge :)

haythemsellami:
Updated `verify-bytecode` to dispatch from `run()` using the same pattern as `cast da-estimate`: optional `--network/-n`, RPC chain-id inference through `AnyNetwork` when omitted, then dispatch to the typed EVM network before entering the generic replay implementation.

crates/verify/src/bytecode.rs
```diff
@@ -127,15 +144,50 @@ impl figment::Provider for VerifyBytecodeArgs {
...
+
+    /// Run the `verify-bytecode` command with the selected EVM network implementation.
+    pub async fn run_with_network<FEN>(self) -> Result<()>
+    where
+        FEN: FoundryEvmNetwork,
+    {
+        let config = self.load_config()?;
+        self.run_with_network_and_config::<FEN>(config).await
+    }
```
mablr: Do we really need to keep this now ?

haythemsellami: not really, removed it.

haythemsellami:
@mablr Lmk if this is a good state to merge.

mablr (dismissed):
Looks good to me now, @figtracer could u please double-check?

haythemsellami:
@figtracer Lmk if you have any feedback, would love to get this one merged.

figtracer:
> @figtracer Lmk if you have any feedback, would love to get this one merged.

sorry for the delay this got lost in my notifications, will take a look

crates/verify/src/bytecode.rs
```diff
@@ -522,33 +551,35 @@ impl VerifyBytecodeArgs {
...
                     input.extend_from_slice(&local_bytecode_vec);
-                    transaction.input = TransactionInput::both(Bytes::from(input));
+                    tx_env.set_data(Bytes::from(input));
 
                     // Deploy default CREATE2 deployer
                     executor.deploy_create2_deployer()?;
                 }
             } else {
-                transaction.input = TransactionInput::both(Bytes::from(local_bytecode_vec));
+                tx_env.set_data(Bytes::from(local_bytecode_vec));
```
figtracer, line 587: this can replay the original create payload instead of `local_bytecode_vec` because `set_data` only changes the inner `TxEnv`

crates/verify/src/bytecode.rs
```diff
@@ -127,15 +144,41 @@ impl figment::Provider for VerifyBytecodeArgs {
 impl VerifyBytecodeArgs {
     /// Run the `verify-bytecode` command to verify the bytecode onchain against the locally built
     /// bytecode.
-    pub async fn run(mut self) -> Result<()> {
-        // Setup
+    pub async fn run(self) -> Result<()> {
         let config = self.load_config()?;
-        let provider = utils::get_provider(&config)?;
+        let network = match self.network {
```
figtracer: this ignores `config.networks`, so `network = "tempo"` can still dispatch as Ethereum if chain-id inference does not recognize the RPC

haythemsellami:
@figtracer pushed an update addressing both comments:

1. `verify-bytecode` network dispatch now checks configured network selection before RPC chain-id inference. Resolution order is: explicit command `--network`, then `config.networks.resolved_network()`, then RPC chain-id inference. Explicit CLI selection is also propagated back into `config.networks`, and inferred non-Ethereum networks are propagated too, so the typed replay path and `config.networks` stay consistent.

2. Tempo `FoundryTransaction` setters now keep AA call payload state in sync with the inner `TxEnv`. `set_data` updates the first `aa_call.input` as well as `inner.data`, and `set_kind` / `set_value` do the same for the first AA call target/value. This prevents replay from accidentally using the original AA create payload after `verify-bytecode` replaces it with `local_bytecode_vec`.

Added focused tests for the config-network precedence and Tempo AA setter behavior. Could you please take another look?

crates/verify/src/bytecode.rs
```diff
@@ -581,3 +629,43 @@ impl VerifyBytecodeArgs {
...
+            "tempo",
+        ]);
+
+        assert_eq!(args.network, Some(NetworkVariant::Tempo));
+    }
+
+    #[test]
+    fn configured_network_uses_config_network() {
+        let mut config = Config::default();
+        config.networks = NetworkVariant::Tempo.into();
```
mablr:
```suggestion
        let config = Config {
            networks: NetworkVariant::Tempo.into(),
            ..Default::default()
        };
```

Clippy warning fix

mablr (changes requested):

mablr (dismissed):
lgtm

figtracer (dismissed):

mablr (dismissed):

figtracer (dismissed):

mablr (approved):

figtracer (approved):

PR #14764, opened 2026-05-14
feat(forge): stream json output for long-running commands

figtracer:
# Summary

- Support incremental NDJSON event output for long-running forge --json commands.
- Breaking: default `forge test --json` now emits an NDJSON event stream instead of the previous aggregate object keyed by suite name.
- Mark `forge test --json --live-logs` as incompatible because live logs write directly to stdout and can corrupt the JSON stream.
- Replace the old aggregate JSON fixtures with focused field-level assertions because the output is now an event stream, the previous fixtures no longer match the contract, and re-snapshotting NDJSON would be large/high-churn without improving coverage.


docs/dev/ndjson-output.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Non-Streaming `forge test --json` Outputs
+
+The NDJSON event stream applies to the default test execution path.
+
+The following combinations emit non-streaming output:
+
+- `forge test --gas-report --json`
+- `forge test --summary --json`
+- `forge test --list --json`
```
figtracer, line 83: non-streaming because these don't produce output incrementally, should be treated as exceptions

mablr (commented):
Sgtm overall.

I think we must be clear this is a breaking change.

We also need to strongly care about components emitting to stdout which can corrupt the JSON output.

While testing i found for instance `LogCollector::LiveLogs::push_raw_log` doesn't comply:
```
forge test --json --live-logs
```
```solidity
import "forge-std/Test.sol";

contract LiveLogTest is Test {
    function test_live() public pure {
        console.log("THIS-LINE-CORRUPTS-NDJSON");
    }
}
```
These flags may be marked as incompatible ?

mablr (approved):

stevencartavia (approved):

PR #14765, opened 2026-05-14
fix(verify): warn and fall back to Sourcify when `ETHERSCAN_API_KEY` is set on unsupported chain

### LINKED ISSUE #10774: bug(`forge script`): if `ETHERSCAN_API_KEY` is set and `--verify` is passed it will try to verify via Etherscan no matter if Etherscan supports the chain
### Component

Forge

- [x] Foundry
- [ ] Foundryup

forge 1.2.3-nightly (6fb7c5983a 2025-06-13T06:02:04.387303000Z)

foundryup: 1.1.0

forge script --verify with `ETHERSCAN_API_KEY` environment variable set

### Operating System

None

### Describe the bug

As reported in https://github.com/foundry-rs/foundry/issues/10593#issuecomment-2959175738 and https://github.com/foundry-rs/foundry/issues/10593#issuecomment-2960012944 users are facing an issue if they have an `ETHERSCAN_API_KEY` set in a scripting environment that deploys to multiple chains as there is no guarantee that the Etherscan API is available on all the target chains.

We should perform a check whether the Etherscan API is available on a chain before attempting to verify.

We should additionally make it so that command line flags always override environment variables. The more explicit the setting is the higher priority it has. This is an example of the behavior: if the user has defined a flag that specifies the `--verifier sourcify` it should ignore the `ETHERSCAN_API_KEY`.

If the user does not specify the `--verifier` / `--verifier-url` it is the correct behaviour to select the Etherscan API. We should however make sure it actually exists - if it does not we should warn the user. This ties into dry-running behavior as defined in https://github.com/foundry-rs/foundry/issues/10593#issuecomment-2955689759 - this is out of scope for this ticket.

mablr:
## Motivation

Closes #10774

When the user does not pass `--verifier` but `ETHERSCAN_API_KEY` is set, we defaulted to Etherscan for any chain, including chains with no known Etherscan API URL, which would then fail at verification time. Now we fall back to Sourcify.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/verify/src/verify.rs
```diff
@@ -281,24 +313,17 @@ impl VerifyArgs {
...
-            || (etherscan_key.as_deref().is_some_and(|k| !k.is_empty())
-                && !matches!(
-                    self.verifier.verifier,
-                    VerificationProviderType::Blockscout
-                        | VerificationProviderType::Oklink
-                        | VerificationProviderType::Custom
-                ));
-        self.verifier
-            .verifier
+        let resolved = self.verifier.resolve(etherscan_key.as_deref(), self.etherscan.chain);
```
figtracer:
```bash
forge verify-contract --etherscan-api-key dummy --chain 3658348 --skip-is-verified-check --retries 1 --delay 0 0x0000000000000000000000000000000000000001 src/Counter.sol:Counter
```

falls back to Sourcify, but the new warning is not emitted

crates/forge/src/cmd/create.rs
```diff
@@ -650,7 +650,9 @@ impl CreateArgs {
             language: None,
             creation_transaction_hash: Some(tx_hash),
         };
-        sh_println!("Waiting for {} to detect contract deployment...", verify.verifier.verifier)?;
+        let resolved_verifier =
+            verify.verifier.resolve(verify.etherscan.key.as_deref(), Some(chain));
```
figtracer: if the API key comes from foundry.toml rather than this `VerifyArgs` value, this can print `sourcify` while the later run uses Etherscan.

figtracer (dismissed):

crates/forge/src/cmd/create.rs
```diff
@@ -650,7 +650,16 @@ impl CreateArgs {
...
         };
-        sh_println!("Waiting for {} to detect contract deployment...", verify.verifier.verifier)?;
+        // Load the full config (including foundry.toml) so the key used for resolution matches
+        // what `verify.run()` will actually use, preventing a "Waiting for sourcify..." message
+        // when the run will actually use Etherscan (or vice versa for unknown chains).
+        let verify_config = verify.load_config()?;
+        let effective_key = verify_config
+            .get_etherscan_config_with_chain(Some(chain))?
+            .map(|c| c.key)
+            .or_else(|| verify_config.etherscan_api_key.clone());
```
zerosnacks, line 672: **Blocker — `forge create` preflight is now inconsistent with `VerifyArgs::run()`.**

The new fallback you added here (and in `VerifyArgs::run()`) is good, but the *preflight* further up at lines 360-361 still does:

```rust
verify.etherscan.key =
    config.get_etherscan_config_with_chain(self.chain_id())?.map(|c| c.key);
```

On an unsupported chain `get_etherscan_config_with_chain` returns `None` even when a raw key is set, so the key is dropped before validation. The combination *unsupported chain + `--verifier etherscan` + `--verifier-url <URL>` + key* — which `client()` explicitly allows (`etherscan_allows_unknown_chain_with_verifier_url`) — can therefore be rejected at preflight even though `run()` would have accepted it.

Please mirror the same `.or_else(|| config.etherscan_api_key.clone())` fallback in the preflight assignment too.

crates/script/src/broadcast.rs
```diff
@@ -733,13 +733,12 @@ impl<FEN: FoundryEvmNetwork> BundledState<FEN> {
...
-                    .config
-                    .get_etherscan_api_key(Some(sequence.chain.into()))
-                    .is_none()
-            {
+            let chain: Chain = sequence.chain.into();
+            let etherscan_key = self.script_config.config.get_etherscan_api_key(Some(chain));
+            // Use the centralized resolver so the preflight reflects the provider that will
+            // actually be used at verification time (not just the explicit CLI value).
+            let resolved = self.args.verifier.resolve(etherscan_key.as_deref(), Some(chain));
+            if resolved == VerificationProviderType::Etherscan && etherscan_key.is_none() {
```
zerosnacks: **Blocker — preflight can reject valid configs and emits the wrong error.**

`get_etherscan_api_key(Some(chain))` returns `None` on unknown chains even when:
- a raw `ETHERSCAN_API_KEY` is set, or
- `--verifier-url` makes Etherscan valid on that chain.

So for *unsupported chain + `--verifier etherscan` + `--verifier-url` + key* (which `VerificationProviderType::client` accepts), this preflight still bails with `"Missing etherscan key"`. Additionally, for explicit `--verifier etherscan` on an unsupported chain *without* URL, the user-facing error here is `"Missing etherscan key"` instead of the more accurate unknown-chain error that `client()` would have produced.

Suggested fix: reuse the same effective-key fallback as runtime (i.e. fall back to `config.etherscan_api_key.clone()` like `VerifyArgs::run()` now does) when computing `etherscan_key`, and let `client()` produce the actual error rather than re-encoding the check here.

crates/script/src/verify.rs
```diff
@@ -189,7 +189,8 @@ async fn verify_contracts<FEN: FoundryEvmNetwork>(
 
     verify.set_chain(config, sequence.chain.into());
 
-    if verify.etherscan.has_key() || verify.verifier.verifier != VerificationProviderType::Etherscan
+    if verify.etherscan.has_key()
+        || verify.verifier.effective_type() != VerificationProviderType::Etherscan
```
zerosnacks, line 194: **Blocker — `set_chain` drops the raw key, and this gate can then skip verification entirely.**

Further up at line 101, `set_chain` still does:

```rust
self.etherscan.key = config.get_etherscan_api_key(Some(chain));
```

which returns `None` for unsupported chains even when a raw `ETHERSCAN_API_KEY` is set. Combined with this gate, the case *unsupported chain + `--verifier etherscan` + `--verifier-url` + key* (which `VerifyArgs::run()` now correctly handles) plays out here as:

- `set_chain()` drops the key → `etherscan.has_key()` is `false`
- `effective_type() == Etherscan` → second clause is `false`
- condition is `false` → **verification is silently skipped**.

Please mirror the runtime fallback in `set_chain()` (i.e. `.or_else(|| config.etherscan_api_key.clone())`) so the gate observes the same key the runtime path will use. Optionally, routing this gate through the resolved provider (e.g. `verifier.resolve(...)`) instead of `effective_type()` would make it robust against future drift.

zerosnacks (dismissed):
The central resolution model in `VerifierArgs::resolve` / `VerificationProviderType::client` reads correctly and the regression for #10774 is addressed cleanly.

However, the new key-fallback was only added in `VerifyArgs::run()`. Several callers around it (forge `create` preflight, forge `script` preflight, and `VerifyBundle::set_chain` + scheduling gate) still drop the raw `ETHERSCAN_API_KEY` on unsupported chains. That makes the otherwise-valid combination *unsupported chain + `--verifier etherscan` + `--verifier-url` + key* fail at preflight or get silently skipped, depending on path.

Marked the three locations as line comments. Once those are aligned with the runtime path (and ideally covered by a couple of regression tests for unsupported-chain + verifier-url + key), this looks good to merge.

crates/script/src/broadcast.rs
mablr: https://github.com/foundry-rs/foundry/pull/14765/commits/360ba4ddd952945cdff7602740615e6e50618e13

crates/forge/src/cmd/create.rs
mablr: https://github.com/foundry-rs/foundry/pull/14765/commits/360ba4ddd952945cdff7602740615e6e50618e13

figtracer (dismissed):

crates/verify/src/verify.rs
```diff
@@ -50,13 +50,45 @@ pub struct VerifierArgs {
...
+        etherscan_key: Option<&str>,
+        chain: Option<Chain>,
+    ) -> VerificationProviderType {
+        if let Some(v) = self.verifier {
+            return v;
         }
+        let has_key = etherscan_key.is_some_and(|k| !k.is_empty());
+        if has_key {
+            let chain_supports_etherscan = chain.is_none_or(|c| c.etherscan_urls().is_some());
+            if chain_supports_etherscan || self.verifier_url.is_some() {
```
figtracer: implicit `ETHERSCAN_API_KEY` still resolves custom-Sourcify chains like Tempo to Etherscan because `etherscan_urls()` exists for their Sourcify-compatible endpoint

crates/script/src/broadcast.rs
```diff
@@ -733,13 +733,16 @@ impl<FEN: FoundryEvmNetwork> BundledState<FEN> {
...
+            let chain: Chain = sequence.chain.into();
+            let etherscan_key = self
+                .script_config
+                .config
+                .get_etherscan_api_key(Some(chain))
+                .or_else(|| self.script_config.config.etherscan_api_key.clone());
+            // Use the centralized resolver so the preflight reflects the provider that will
+            // actually be used at verification time (not just the explicit CLI value).
+            let resolved = self.args.verifier.resolve(etherscan_key.as_deref(), Some(chain));
+            if resolved == VerificationProviderType::Etherscan && etherscan_key.is_none() {
```
figtracer: preflight still allows `--verifier etherscan` on an unknown chain with a raw key but no `--verifier-url`; deployment can proceed before `VerifyArgs::run()` fails

figtracer:
@mablr there's some extra stuff to handle maybe ptal

mattsse (dismissed):
P1: Custom-Sourcify + user verifier URL still routes to Etherscan at runtime.

`VerifyArgs::run` resolves Tempo/custom-Sourcify to `Sourcify` before URL injection: [`verify.rs#L294-L303`](https://github.com/foundry-rs/foundry/blob/6e85b856136c67b64117b11669ecc39e17b2319d/crates/verify/src/verify.rs#L294-L303). But the actual provider is selected later via `resolved.client(..., had_user_verifier_url, ...)`: [`verify.rs#L338-L345`](https://github.com/foundry-rs/foundry/blob/6e85b856136c67b64117b11669ecc39e17b2319d/crates/verify/src/verify.rs#L338-L345). When the user passed `--verifier-url`, `has_url` is true, so the implicit `ETHERSCAN_API_KEY` branch skips the custom-Sourcify fallback guarded by `&& !has_url` and returns `EtherscanVerificationProvider`: [`provider.rs#L232-L248`](https://github.com/foundry-rs/foundry/blob/6e85b856136c67b64117b11669ecc39e17b2319d/crates/verify/src/provider.rs#L232-L248).

This means a Tempo/custom-Sourcify verification with `ETHERSCAN_API_KEY` set and `--verifier-url` supplied can still send Etherscan-shaped requests to the Sourcify endpoint.

Please make `client()` mirror `resolve()`: treat `chain.is_custom_sourcify()` as a hard opt-out from the implicit Etherscan branch unless `--verifier etherscan` was explicitly selected, while preserving the user-URL fallback for non-custom unknown chains. Also add coverage for provider construction/runtime selection; the new test only asserts `resolve()` and misses this path: [`verify.rs#L704-L715`](https://github.com/foundry-rs/foundry/blob/6e85b856136c67b64117b11669ecc39e17b2319d/crates/verify/src/verify.rs#L704-L715).

crates/verify/src/provider.rs
```diff
@@ -212,25 +217,48 @@ impl VerificationProviderType {
...
+        //    is supported; otherwise warn and fall through to the Sourcify default. See <https://github.com/foundry-rs/foundry/issues/10774>.
+        //    Custom-Sourcify chains (e.g. Tempo) register Sourcify-compatible URLs under
+        //    etherscan_urls() and are excluded here even when a URL is present.
         if has_key {
-            sh_eprintln!(
-                "ETHERSCAN_API_KEY is set, defaulting to Etherscan verifier. \
-                 Unset it or pass `--verifier sourcify` (or another provider) to override."
+            if let Some(chain) = chain
+                && (chain.etherscan_urls().is_none() || chain.is_custom_sourcify())
+                && !has_url
```
stevencartavia: `resolve()` and `client()` disagree here for: no `--verifier`, key set, custom-Sourcify chain (e.g. Tempo), user `--verifier-url`. `resolve()` returns `Sourcify`, this branch builds Etherscan because of the `&& !has_url` clause. Should they agree?

mablr: https://github.com/foundry-rs/foundry/pull/14765/commits/04f4810b159395216a1a0316f053fa1a2769069d

crates/verify/src/provider.rs
```diff
@@ -212,25 +217,47 @@ impl VerificationProviderType {
...
+        //    etherscan_urls() and are hard-excluded here regardless of whether a URL is present.
         if has_key {
-            sh_eprintln!(
-                "ETHERSCAN_API_KEY is set, defaulting to Etherscan verifier. \
-                 Unset it or pass `--verifier sourcify` (or another provider) to override."
+            if let Some(chain) = chain
+                && (chain.is_custom_sourcify() || chain.etherscan_urls().is_none() && !has_url)
+            {
+                sh_warn!(
+                    "ETHERSCAN_API_KEY is set but chain {chain} has no known Etherscan API URL. \
```
figtracer: This warning is misleading for custom-Sourcify chains when `--verifier-url` is already set; live Tempo verification still says to pass `--verifier-url`. Consider a separate message that says to pass an explicit `--verifier`.

crates/verify/src/provider.rs
```diff
@@ -252,19 +279,70 @@ mod tests {
...
+
+    // Regression: implicit ETHERSCAN_API_KEY + user-supplied --verifier-url on a custom-Sourcify
+    // chain must still fall back to Sourcify, not Etherscan.
+    #[test]
+    fn implicit_etherscan_on_custom_sourcify_chain_with_url_falls_back_to_sourcify() {
+        let tempo = Chain::from(4217u64);
+        // has_url=true simulates the user passing --verifier-url (or auto-injected URL)
+        let res =
+            VerificationProviderType::Sourcify.client(Some("mykey"), Some(tempo), true, false);
+        assert!(res.is_ok(), "expected fallback to Sourcify, got error");
```
figtracer: This test only asserts `Ok`, so it would pass even if `client()` still returned Etherscan. Please assert the selected provider or cover this through an observable runtime path.

crates/verify/src/provider.rs
mablr: fixed in 831b251828ddc572761fa4e279bf83512bcf844e

crates/verify/src/provider.rs
mablr: A bit clunky https://github.com/foundry-rs/foundry/pull/14765/commits/e1a570cb2348061da997b0f6327188db8e20cae5 but now fall backs are properly asserted.

figtracer (approved):

grandizzy (approved):

PR #14766, opened 2026-05-14
feat(cast): add `cast keychain doctor`

figtracer:
## Summary

Adds `cast keychain doctor`, a diagnostic subcommand that walks the Tempo access-key signing path step by step and reports exactly which step would fail in plain English and structured JSON.

Today, an access-key tx that fails can revert deep inside the precompile and leave the user guessing whether the local config is wrong, the key is missing or only pending via `key_authorization`, revoked, expired, out of budget, scoped to deny this specific call, using an invalid expiring-nonce window, or configured with a bad sponsorship setup. This removes that AA debugging tax and gives agents a stable gate.

## CLI

```sh
cast keychain doctor [KEY_ADDRESS]
  [--root-account <addr>]       # required if KEY_ADDRESS doesn't resolve via local registry
  [--to <addr>]                 # hypothetical target for TIP-1011 scope check
  [--selector <sel>]            # hex `0x12345678`, shorthand (`transfer`), or signature
  [--recipient <addr>]          # recipient list match for selector rule
  [--fee-token <addr>]          # token to check balance/limits for; defaults to PathUSD
  [--tempo.fee-token <addr>]    # also accepted, same fallback path as send/mktx
  [--tempo.expires <seconds>]   # TIP-1009 convenience window
  [--tempo.expiring-nonce --tempo.valid-before <ts> [--tempo.valid-after <ts>]]
  [--tempo.sponsor <addr> (--tempo.sponsor-signer <uri> | --tempo.sponsor-sig <sig>)]
  [--tempo.print-sponsor-hash]
  --rpc-url <url>
```

`--selector` requires `--to`; `--recipient` requires `--selector`.

## JSON contract

JSON output includes:

- `schema_version: 1`
- `context` with the resolved `root_account`, `key_address`, `chain_id`, and `fee_token`
- stable step `name`s for automation: `local_registry`, `rpc_reachability`, `chain_id_match`, `local_signing`, `key_registration`, `revocation`, `expiry`, `hardfork`, `spending_limits`, `allowed_calls`, `expiring_nonce`, `sponsorship`, `fee_token_balance`
- top-level `status`, `no_failures`, `healthy`, `warning_count`, and `failure_count`

`healthy` remains the strict green gate. `no_failures` means runnable with warnings.

## Follow-ups

- Browser-wallet `KeyAuthorization` signing capability checks once https://github.com/foundry-rs/foundry/pull/14743 merges. This PR already validates saved `key_authorization` blobs for pending, not-yet-on-chain keys.
- Full fee-payer signature recovery still needs a concrete transaction payload/digest; doctor validates config-level sponsorship mistakes now, while exact digest verification belongs in the target send/mktx path or a future doctor mode that accepts full tx intent.


crates/cast/src/cmd/keychain.rs
```diff
@@ -776,135 +846,1632 @@ async fn run_inspect(
...
+
+    // Prefer the same default selection as `key_metadata_from_entry` callers (entries with
+    // local limit metadata first), but keep all chain-matching candidates for step 3.
+    matches.sort_by_key(|entry| usize::MAX - entry.limits.len());
+    let count = matches.len();
+    let mut candidates: Vec<DoctorCandidate> =
+        matches.into_iter().map(DoctorCandidate::from_entry).collect();
+    if let Some(candidate) = explicit_candidate() {
+        candidates.push(candidate);
+    }
```
zerosnacks: **Default-key selection diverges from the send path.**

This preference cascade (entries with more local limits first, then inline-key in `select_subject_for_chain`) is not what MPP key discovery in `crates/common/src/provider/mpp/keys.rs:83-104` does (passkey-with-inline-key first, then first inline key, then first entry).

Result: `cast keychain doctor --root-account ...` can diagnose a different key than the send path would actually use, so every later step looks precise while targeting the wrong subject.

Suggest reusing the shared `discover_mpp_config` selector (or mirroring it exactly) and adding a test where two candidate keys for the same root produce divergent picks.

crates/cast/src/cmd/keychain.rs
```diff
@@ -776,135 +846,1632 @@ async fn run_inspect(
...
+/// Pure TIP-1011 matching logic. Extracted so it can be unit-tested.
+fn match_allowed_call<'a>(
+    scopes: &'a [CallScope],
+    to: Address,
+    selector: [u8; 4],
+    recipient: Option<Address>,
+) -> AllowedCallMatch<'a> {
+    let Some(scope) = scopes.iter().find(|s| s.target == to) else {
+        return AllowedCallMatch::Denied(format!("target {to} not in any allowed scope"));
+    };
```
zerosnacks: **Only the first scope per target is inspected.**

```rust
let Some(scope) = scopes.iter().find(|s| s.target == to) else { ... };
```

Duplicate-target `CallScope`s are possible from repeated `--scope`, JSON input, or an externally-produced `key_authorization`. `.find` ignores the rest and yields a false `Denied` if the matching selector/recipient lives in a later scope.

Suggested fix: aggregate selector rules across **all** scopes where `target == to`:
- target wildcard if **any** matching scope has empty `selectorRules`
- otherwise search all selector rules across matching scopes
- recipient restrictions checked across all matching rules

Tests to add: same target split across two scopes, selector only in second; same target/selector split with different recipient sets.

crates/cast/src/cmd/keychain.rs
```diff
@@ -776,135 +846,1632 @@ async fn run_inspect(
...
+    let valid_after = tempo.valid_after;
+
+    if (valid_before.is_some() || valid_after.is_some()) && !tempo.expiring_nonce {
+        return DoctorStep::warn(
+            "expiring_nonce",
+            "Expiring nonce",
+            "validity window set without --tempo.expiring-nonce",
+            "use --tempo.expiring-nonce or --tempo.expires so nonce_key is set to the expiring lane",
+        );
+    }
```
zerosnacks: **Early return skips actual window validation.**

`TempoOpts::apply` still applies `valid_before` / `valid_after` even without `--tempo.expiring-nonce` (`crates/cli/src/opts/tempo.rs:243-255`). Returning here means the checks below are never run:
- `valid_after >= valid_before`
- `valid_before <= chain_timestamp`
- 3s / 5s / 30s thresholds

Suggest keeping the warning, but continuing to validate the window and returning the worst status.

(Confirmed no underflow on `ttl = before - chain_timestamp` further down — the `before <= chain_timestamp` guard at L2192 prevents it.)

Tests to add: `valid_after >= valid_before`, `valid_before == chain_timestamp`, `ttl ∈ {3, 4, 5, 31}`, all with and without `--tempo.expiring-nonce`.

crates/cast/src/cmd/keychain.rs
```diff
@@ -776,135 +846,1632 @@ async fn run_inspect(
...
+            "spending_limits",
+            "Spending limits",
+            "key_authorization allows no token spending",
+            "refresh the access key with spending limits if the transaction spends TIP-20 tokens",
+        ),
+        Some(limits) => {
+            let detail = format_authorization_limits(limits, fee_token);
+            DoctorStep::pass("spending_limits", "Spending limits", detail)
+        }
+    }
```
zerosnacks, line 1987: **Pending `key_authorization` limit check is too optimistic.**

For non-empty `auth.limits` this always returns `Pass`, but `format_authorization_limits` (L2384-L2386) appends `"{fee_token} not listed in key_authorization limits"` when the fee token is missing.

The doctor can therefore report the access-key signing path healthy while the chosen fee token is effectively unspendable under the pending authorization.

Suggested behavior for pending auth:
- fee token missing from `auth.limits` → at least `Warn`, arguably `Fail`
- fee token present with amount `0` → `Warn`/`Fail`
- periodic/T3-only limits with hardfork unknown → `Warn`, not `Pass`

Tests to add: fee token missing; fee token present with zero amount; periodic limits on pre-T3 / hardfork-unknown.

crates/cast/src/cmd/keychain.rs
```diff
@@ -776,135 +846,1632 @@ async fn run_inspect(
...
+    if let Some(expiry) = auth.expiry
+        && expiry.get() <= now
+    {
+        return Err(DoctorStep::fail(
+            "key_registration",
+            "Key registration",
+            format!("local key_authorization expired {}", format_relative_timestamp(expiry.get())),
+            "refresh the access key to get a later key_authorization expiry",
+        ));
+    }
```
zerosnacks: **Expiry checks use wall-clock instead of chain time.**

This (and the on-chain `info.expiry <= now` at L1300, and `check_authorization_expiry` at L1699-L1723) compares expiry against `SystemTime::now()`. Enforcement is by chain timestamp, not local time, so a skewed workstation clock can produce false expired/active reports for near-boundary keys.

`check_expiring_nonce` already fetches the latest block timestamp — suggest fetching it once early in `run_doctor` and threading it through:
- on-chain key expiry
- pending-auth expiry
- expiring-nonce window checks

crates/cast/src/cmd/keychain.rs
```diff
@@ -776,135 +846,1632 @@ async fn run_inspect(
...
+    info: &KeyInfo,
+    fee_token: Address,
+    is_t3: Option<bool>,
+) -> DoctorStep
+where
+    P: Provider<TempoNetwork>,
+{
+    let Some(is_t3) = is_t3 else {
+        return DoctorStep::pass("spending_limits", "Spending limits", "skipped; hardfork unknown");
+    };
```
zerosnacks, line 1838: **"Skipped/unknown" branches are labeled `Pass`.**

Given the PR description positions the JSON as the agent-facing contract, `pass` should mean "positively checked and OK", not "we didn't evaluate this". Same pattern at:
- L1907-L1909 (`check_allowed_calls` — `"skipped; hardfork unknown"` → Pass)
- L1362-L1367 (`run_doctor` — `"skipped; sponsorship config is invalid"` → Pass for `fee_token_balance`)

Suggest promoting these to `Warn` for `schema_version: 1`. If a distinct "skipped" state is wanted, that's a `schema_version: 2` change.

zerosnacks (commented):
A handful of correctness issues worth fixing before merge are noted inline. Top-level JSON shape, `--fee-token` precedence (`fee_token.or(tempo.fee_token).unwrap_or(PATH_USD_ADDRESS)`), and the `sponsor == sender` ordering all look correct.

Summary of inline comments:
1. Default-key selection diverges from MPP send-path discovery.
2. `match_allowed_call` only inspects the first scope per target.
3. `check_expiring_nonce` short-circuits before validating the window.
4. Pending `key_authorization` limit check is too optimistic (false-green when fee token is missing).
5. Expiry checks use wall-clock instead of chain time.
6. "Skipped/unknown" branches are labeled `Pass` — misleading when JSON is the agent-facing contract.

zerosnacks (dismissed):
lgtm! oracle and manual review did not yield new findings, others were resolved

crates/cast/src/cmd/keychain.rs
```diff
@@ -822,89 +892,1705 @@ async fn run_inspect(
...
+
+    if !provisioned {
+        sh_println!("Status:         {} not provisioned", "✗".red())?;
+        return Ok(());
+    }
+
+    // Status line: active key.
+    {
+        sh_println!("Status:         {} active", "✓".green())?;
+    }
```
mablr: Why do we have braces here ?

mablr (dismissed):
lgtm, just a non-blocking style nit

crates/cast/src/cmd/keychain.rs
figtracer: oh this is leftover from a previous iteration we can get rid of them

mablr (approved):

zerosnacks (approved):

PR #14767, opened 2026-05-14
Fix cast log chunk boundaries

ml0mbardi:
## Motivation
                                                                                                                                                                                                 
  `anvil_loadState` imported accounts and storage, but it did not repopulate the execution-layer block-hash cache from imported blocks.                                                          
  That can leave `BLOCKHASH`-based reads inconsistent after a state import, even when the block data is present.                                                                                 
                                                                                                                                                                                                 
  ## Solution                                                                                                                                                                                    
                                                                                                                                                                                                 
  This change backfills block hashes during state import and keeps historical snapshots handling intact:                                                                                         
                                                                                                                                                                                                 
  - `Db::load_state` now iterates imported `blocks` and inserts `(block_number, block_hash)` into the local block-hash cache.                                                                    
  - `Backend::load_state` takes `historical_states` before moving `state` into DB loading, avoiding unnecessary cloning and preserving behavior.                                                 
  - Added a regression test (`test_load_state_backfills_block_hash_cache`) that verifies a mined block hash is still available after `dump_state -> load_state_bytes`.                           
                                                                                                                                                                                                 
  ## PR Checklist                                                                                                                                                                                
                                                                                                                                                                                                 
  - [x] Added Tests                                                                                                                                                                              
  - [ ] Added Documentation                                                                                                                                                                      
  - [ ] Breaking changes

### DIFF
--- a/crates/anvil/src/eth/backend/db.rs
+++ b/crates/anvil/src/eth/backend/db.rs
@@ -226,7 +226,9 @@ pub trait Db:
 
     /// Deserialize and add all chain data to the backend storage
     fn load_state(&mut self, state: SerializableState) -> DatabaseResult<bool> {
-        for (addr, account) in state.accounts {
+        let SerializableState { accounts, blocks, .. } = state;
+
+        for (addr, account) in accounts {
             let old_account_nonce = DatabaseRef::basic_ref(self, addr)
                 .ok()
                 .and_then(|acc| acc.map(|acc| acc.nonce))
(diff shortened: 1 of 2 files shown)


crates/anvil/src/eth/backend/db.rs
```diff
@@ -254,6 +256,15 @@ pub trait Db:
                 self.set_storage_at(addr, k, v)?;
             }
         }
+
+        // Backfill execution-layer block hash cache for imported blocks so BLOCKHASH lookups can
+        // hit local DB data after `anvil_loadState`.
+        for block in blocks {
+            let number = U256::from(block.header.number);
+            let hash = block.header.hash_slow();
```
mablr, line 264: Maybe rehashing here can be avoided ?  `load_blocks` running in `Backend::load_state` does the job ?

crates/anvil/src/eth/backend/db.rs
```diff
@@ -254,6 +256,15 @@ pub trait Db:
                 self.set_storage_at(addr, k, v)?;
             }
         }
+
+        // Backfill execution-layer block hash cache for imported blocks so BLOCKHASH lookups can
+        // hit local DB data after `anvil_loadState`.
+        for block in blocks {
+            let number = U256::from(block.header.number);
```
mablr, line 263: style, let's `block.header.number()` for consistency.

PR #14769, opened 2026-05-14
feat(lint): add `uninitialized-state`

mablr:
## Motivation

Add `uninitialized-state` lint detecting state variables that are never initialized and used before assignment.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/lint/src/sol/med/uninitialized_state_variables.rs
```diff
@@ -0,0 +1,317 @@
...
+
+        // Collect non-constant, non-immutable state variables from the whole
+        // inheritance chain (linearized_bases[0] is the contract itself).
+        let state_vars: Vec<VariableId> = contract
+            .linearized_bases
+            .iter()
+            .flat_map(|&cid| hir.contract(cid).variables())
+            .filter(|&var_id| {
+                let var = hir.variable(var_id);
+                !var.is_constant() && !var.is_immutable()
```
figtracer: a public/read-only mapping is reported here as well, we could exclude mappings or add a false-positive fixture

crates/lint/src/sol/med/uninitialized_state_variables.rs
```diff
@@ -0,0 +1,317 @@
...
+        ExprKind::Array(exprs) => {
+            for expr in *exprs {
+                collect_expr_writes_checked(expr, candidates, writes)?;
+            }
+        }
+        ExprKind::Binary(lhs, _, rhs) => {
+            collect_expr_writes_checked(lhs, candidates, writes)?;
+            collect_expr_writes_checked(rhs, candidates, writes)?;
+        }
+        ExprKind::Call(callee, args, named_args) => {
```
figtracer, line 241: misses writes through storage-reference internal/library calls, e.g.: 

`_set(data, value)` where `_set(Data storage self, ...)` mutates `self`, or `data.set(value)` via a library

crates/lint/src/sol/med/uninitialized_state_variables.rs
```diff
@@ -0,0 +1,317 @@
...
+impl<'hir> LateLintPass<'hir> for UninitializedStateVariables {
+    fn check_nested_contract(
+        &mut self,
+        ctx: &LintContext,
+        hir: &'hir Hir<'hir>,
+        contract_id: ContractId,
+    ) {
+        let contract = hir.contract(contract_id);
+
+        if contract.kind == ContractKind::Interface {
```
figtracer: still analyzes abstract contracts, should we also skip them?

figtracer (commented):
nits :D

figtracer (dismissed):

crates/lint/src/sol/med/uninitialized_state_variables.rs
```diff
@@ -0,0 +1,343 @@
...
+            if let ExprKind::Member(base, member) = &callee.kind {
+                let name = member.name;
+                if name == sym::push || name.as_str() == "pop" {
+                    collect_lvalue_writes(base, candidates, writes);
+                }
+            }
+
+            // Direct calls to internal functions that take a `storage` parameter
+            // mutate the corresponding argument in place; treat it as a write.
+            if let ExprKind::Ident(resolutions) = &callee.kind {
```
figtracer: still misses writes through storage-reference library/member calls, e.g.:

data.set(value) via `using Lib for Data`, where `set(Data storage self, ...)` mutates `self`

direct `_set(data, value)` is covered, but member/library dispatch won't hit this `ExprKind::Ident` branch

crates/lint/src/sol/med/uninitialized_state_variables.rs
```diff
@@ -0,0 +1,341 @@
...
+                collect_lvalue_writes(base, candidates, writes);
+            }
+
+            // Direct calls to internal functions that take a `storage` parameter
+            // mutate the corresponding argument in place; treat it as a write.
+            if let ExprKind::Ident(resolutions) = &callee.kind {
+                for res in *resolutions {
+                    if let Res::Item(ItemId::Function(func_id)) = res {
+                        let func = hir.function(*func_id);
+                        for (&param_id, arg_expr) in func.parameters.iter().zip(args.exprs()) {
```
stevencartavia: this pairs parameters with arguments by position, which can break for reordered named function args because `args.exprs()` returns them in source order, not parameter order — the storage write is missed and the var is falsely flagged as uninitialized.
```solidity
contract NamedArgsBug {
    struct Data { uint256 val; }
    Data public slot; // flagged

    function _set(Data storage target, uint256 v) internal { target.val = v; }
    function set(uint256 v) external { _set({v: v, target: slot}); }
    function get() public view returns (uint256) { return slot.val; }
}
```

crates/lint/docs/uninitialized-state.md
```diff
@@ -0,0 +1,73 @@
...
+modifier argument, or compiler-synthesised public getter) but never written by any of the
+above, it is flagged.
+
+**Assembly bail-out**: if any function body in the inheritance chain contains inline assembly
+(which Solar lowers to an opaque AST node), the lint skips the entire contract conservatively
+to avoid false positives from untracked storage writes.
+
+**Known limitations**:
+- *Storage aliases*: `Foo storage f = bar; f.x = 1;` is not detected as a write to `bar`.
+- *Storage-parameter calls*: calling an internal function that accepts a `storage` reference
```
stevencartavia: the lint does handle direct positional calls to internal functions with a storage parameter (StorageRefCall testdata), worth narrowing or dropping this bullet

crates/lint/src/sol/med/uninitialized_state_variables.rs
```diff
@@ -0,0 +1,341 @@
...
+        }
+        ExprKind::Tuple(exprs) => {
+            for expr in exprs.iter().flatten() {
+                collect_lvalue_writes(expr, candidates, writes);
+            }
+        }
+        ExprKind::Index(base, _)
+        | ExprKind::Slice(base, _, _)
+        | ExprKind::Member(base, _)
+        | ExprKind::Payable(base) => collect_lvalue_writes(base, candidates, writes),
```
stevencartavia: this arm causes `payable(owner).transfer(...)` to mark owner as written via the member-call receiver path above, which makes the lint miss its own documented "Bad" example in the docs

crates/lint/src/sol/med/uninitialized_state_variables.rs
```diff
@@ -0,0 +1,341 @@
...
+        if matches!(contract.kind, ContractKind::Interface | ContractKind::AbstractContract) {
+            return;
+        }
+
+        // Only analyse the most-derived contract in each hierarchy so each
+        // state variable is reported exactly once and write/read information
+        // from every level of the inheritance chain is visible.
+        if hir.contracts().any(|c| c.linearized_bases.iter().skip(1).any(|&id| id == contract_id)) {
+            return;
+        }
```
stevencartavia: this skips any contract that has a child, but a contract can be both inherited from and deployed standalone. here `Base` is silently ignored even though deploying it directly leaves `x` uninitialized.
```solidity
contract Base {
    uint256 public x; // not flagged
    function get() external view returns (uint256) { return x; }
}
contract Derived is Base { constructor() { x = 42; } }
```

crates/lint/src/sol/med/uninitialized_state_variables.rs
```diff
@@ -0,0 +1,341 @@
...
+            if let ExprKind::Member(base, _) = &callee.kind {
+                // Covers push/pop and library dispatch (`using Lib for T` with `T storage self`);
+                // can't resolve callee without Gcx. Treat the receiver as a write target to avoid
+                // false positives.
+                collect_lvalue_writes(base, candidates, writes);
+            }
+
+            // Direct calls to internal functions that take a `storage` parameter
+            // mutate the corresponding argument in place; treat it as a write.
+            if let ExprKind::Ident(resolutions) = &callee.kind {
```
stevencartavia: this carries every visible overload, and this loop walks all of them — so if any candidate has a storage param at the arg's position, the arg is marked written even when the actually-selected overload doesn't take storage?
```solidity 
contract OverloadUnion {
    struct S { uint256 v; }
    uint256 public x; // not flagged
    S internal data;

    function f(uint256) internal {}
    function f(S storage) internal {}

    function callIt() external { f(x); }
    function read() external view returns (uint256) { return x; }
}
```

crates/lint/src/sol/med/uninitialized_state_variables.rs
```diff
@@ -0,0 +1,367 @@
...
+                                    && let Some(arg) = named
+                                        .iter()
+                                        .find(|a| param.name.is_some_and(|n| n == a.name))
+                                {
+                                    collect_lvalue_writes(&arg.value, candidates, writes);
+                                }
+                            }
+                        }
+                    }
+                }
```
zerosnacks: **Suggested test: named args, out-of-order, state-var in non-storage position.**

This exposes a latent bug here. `CallArgsKind::Named(args).exprs()` iterates by **source order** (verified in `solar-sema-0.1.8/src/hir/mod.rs:1361-1368`: `Self::Named(args) => args.iter().map(|arg| &arg.value)`), but the positional `enumerate()` loop above indexes parameters by position. For a named call, the loop misaligns parameters with arguments.

The existing test at `UninitializedStateVariables.sol:307-314` doesn't catch this because the misaligned arg is a local (`v`), not a candidate state var. A test where the wrongly-marked arg is itself a candidate state var would fail:

```solidity
// ── Named args, out-of-order, exposes the named-arg storage bug ───────────
contract NamedArgMisfire {
    struct Data { uint256 val; }
    Data slot;
    uint256 public x; //~WARN: state variable is read but never written

    function _set(uint256 v, Data storage target) internal { target.val = v; }
    function callIt() external { _set({target: slot, v: x}); }
    function read() external view returns (uint256) { return x; }
}
```

Fix: gate the positional `enumerate()` block to `CallArgsKind::Unnamed`, and make the named-args branch conservative across overloads (require **all** candidate overloads agree, matching the rule in the comment at L228-231).

crates/lint/src/sol/med/uninitialized_state_variables.rs
```diff
@@ -0,0 +1,367 @@
...
+
+        let candidate_set: HashSet<VariableId> = state_vars.iter().copied().collect();
+
+        let mut written: HashSet<VariableId> = HashSet::new();
+
+        for &var_id in &state_vars {
+            if hir.variable(var_id).initializer.is_some() {
+                written.insert(var_id);
+            }
+        }
```
zerosnacks, line 71: **Suggested test: state-var initializer with side-effect that writes another state var.**

The loop marks the declaring var as written when it has an initializer, but never traverses the initializer expression for writes to *other* state vars. This is a real functional gap — Solidity initializers can call functions with side effects:

```solidity
// ── Initializer side effect writes another state var ──────────────────────
contract InitSideEffect {
    uint256 public y; // should NOT warn (written via _init())
    uint256 public x = _init();

    function _init() internal returns (uint256) {
        y = 1;
        return 2;
    }

    function getY() external view returns (uint256) { return y; }
}
```

Fix: in the write-collection pass, also call `collect_expr_writes_checked(hir, init, &candidate_set, &mut written)` on every state-var initializer across `contract.linearized_bases` — analogous to how function bodies, modifier args, and base-constructor args are already walked.

crates/lint/src/sol/med/uninitialized_state_variables.rs
```diff
@@ -0,0 +1,367 @@
...
+            collect_expr_writes_checked(hir, lhs, candidates, writes)?;
+            collect_expr_writes_checked(hir, rhs, candidates, writes)?;
+        }
+        ExprKind::Call(callee, args, named_args) => {
+            if let ExprKind::Member(base, _) = &callee.kind {
+                // Covers push/pop and library dispatch (`using Lib for T` with `T storage self`);
+                // can't resolve callee without Gcx. Treat the receiver as a write target to avoid
+                // false positives.
+                collect_lvalue_writes(base, candidates, writes);
+            }
```
zerosnacks, line 247: **Suggested test: read-only member call on an uninitialized address/interface var.**

Marking the receiver of *any* member call as written is a much broader carve-out than the docs admit. It silently suppresses warnings for common read-only patterns like `token.balanceOf(...)`, `oracle.latestAnswer()`, or `target.call(data)` — all of which would otherwise be precisely the kind of bug this lint should catch.

```solidity
// ── Read-only member call should still warn (currently suppressed) ────────
interface IOracle { function latestAnswer() external view returns (uint256); }

contract ReadOnlyMemberCall {
    IOracle public oracle; //~WARN: state variable is read but never written

    function price() external view returns (uint256) {
        return oracle.latestAnswer();
    }
}
```

Either narrow this heuristic to known mutators (`.push()`/`.pop()`), or — if intentional — document the false-negative behavior in `docs/uninitialized-state.md` so users aren't surprised. The doc currently doesn't mention this carve-out at all.

zerosnacks (commented):
Three suggested test cases that expose latent issues in the current implementation. Posted as line comments below.

crates/lint/src/sol/med/uninitialized_state_variables.rs
mablr: added doc https://github.com/foundry-rs/foundry/pull/14769/commits/b8daa64880b10bf104b5430d687a9e671e7d9f9e

mattsse:
Found a couple of false-positive cases in the storage-write tracking:

1. `crates/lint/src/sol/med/uninitialized_state_variables.rs` around line 252: storage writes through the selected overload can still be missed. Because `all_storage` requires every visible overload candidate to have `storage` at this position, a valid call like `set(slot)` with overloads `set(Data storage)` and `set(uint256)` does not mark `slot` as written, even though Solidity selects the storage overload and the helper mutates it. I reproduced this locally with a small fixture where `Data public slot` is written through `set(slot)` but still gets reported as uninitialized.

2. `crates/lint/src/sol/med/uninitialized_state_variables.rs` around line 236: storage-reference write detection only runs for bare identifier callees, so qualified inherited helper calls are missed. For example, `BaseSetter.setS(slot)` or `super.setS(slot)` are member callees; even when the inherited internal helper takes `S storage` and mutates it, the argument is not marked written and the state variable is falsely warned as uninitialized.

mattsse (commented):
Two remaining items:

1. The [`super` lookup in `collect_callee_funcs`](https://github.com/foundry-rs/foundry/blob/b0ae5eb4dac6431d1d682d44bec790704c3e1beb/crates/lint/src/sol/med/uninitialized_state_variables.rs#L335) searches the whole analyzed linearized base list, including the current contract. Combined with [`any_storage`](https://github.com/foundry-rs/foundry/blob/b0ae5eb4dac6431d1d682d44bec790704c3e1beb/crates/lint/src/sol/med/uninitialized_state_variables.rs#L378), a child-only storage overload can mark an argument as written even when `super.f(x)` actually resolves to a non-storage parent overload, suppressing a real `uninitialized-state` warning. A negative fixture with a child storage overload and parent non-storage overload should catch this.

2. The known-limitation docs still say [storage-parameter calls are not detected](https://github.com/foundry-rs/foundry/blob/b0ae5eb4dac6431d1d682d44bec790704c3e1beb/crates/lint/docs/uninitialized-state.md#L28), but the current implementation and fixtures cover bare, named, qualified, and `super` storage arguments. The limitation should be narrowed to the remaining unsupported cases, such as aliases or unresolved/member-call tradeoffs.

mablr (commented):
Fixed doc, and  `super` lookup in `collect_callee_funcs` https://github.com/foundry-rs/foundry/pull/14769/commits/8b6fb55055dafed5a48c4b6f667c72c7c536a25c.

figtracer (approved):

mattsse (approved):
LGTM. The remaining overload/super storage-reference edge cases look acceptable to handle as follow-up limitations rather than blockers for this lint.

PR #14770, opened 2026-05-14
feat(cli): add `--introspect` for machine-readable command discovery

stevencartavia:
Adds `--introspect` to forge, cast, anvil, and chisel: a pre-parse flag that emits a stable, machine-readable JSON description of the binary's command surface (`foundry:introspect@v1`).

Each command exposes a stable `command_id`, typed args, and capability metadata (output mode, side effects, stdin/output support, project requirement). Per-binary CI checks enforce that `command_id`s never collide.

`--markdown-help` rendering is unchanged; a stderr note now points agents to `--introspect`.

Spec under `docs/agents/`. Foundation for a stack adding `--machine`, exit codes, diagnostics, envelope adoption, NDJSON streaming, and anvil session records.

Part of OSS-218


docs/agents/spec.md
```diff
@@ -0,0 +1,311 @@
+# Foundry agent contract
```
mablr, line 1: This seem to overstate what's implemented in this PR.

crates/cli/src/opts/global.rs
```diff
@@ -54,14 +54,53 @@ impl GlobalArgs {
...
             std::process::exit(0);
         }
     }
 
+    /// Check if `--introspect` was passed and print the introspection document as JSON, then exit.
+    ///
+    /// Must run **before** clap parsing so required subcommands or args don't block discovery.
+    /// Takes precedence over `--markdown-help` when both are present.
+    pub fn check_introspect<C: clap::CommandFactory>() {
+        Self::check_introspect_with(C::command(), &crate::introspect::CommandRegistry::EMPTY)
```
mablr: Every binary calls check_introspect with `CommandRegistry::EMPTY`, so every emitted command ships `command_id_stable=false`, `capabilities_declared=false`, `exit_codes=[]`.  The "stable machine contract" framed in the docs isn't actually live, let's shrink doc or leave a todo ?

crates/cli/src/introspect/document.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// Arguments declared directly on this command.
+    pub args: Vec<ArgInfo>,
+    /// Subcommands of this command.
+    pub subcommands: Vec<Self>,
+    /// Capabilities reported for agent consumers.
+    pub capabilities: Capabilities,
+    /// Whether `capabilities` was authored in the registry. When `false`,
+    /// every capability field is a non-authoritative default and consumers
+    /// MUST treat side-effects, project requirement, etc. as unknown.
+    pub capabilities_declared: bool,
```
mablr, line 67: `capabilities_declared` semantics seem to be a footgun, `CommandMeta` requires a concrete `Capabilities`, and `sets capabilities_declared = meta.is_some()`. The moment anyone adds a registry entry just to pin a `command_id`, the placeholder `Capabilities::NONE` becomes "authoritative".

crates/cli/src/opts/global.rs
```diff
@@ -132,6 +171,16 @@ impl GlobalArgs {
...
 
+/// Returns whether `flag` is present in the process arguments before any `--`
+/// separator.
+///
+/// Uses `args_os` so non-UTF-8 positional values don't panic, and stops at
+/// `--` so a literal positional like `cast call ... -- --introspect` cannot
+/// accidentally trigger a pre-parse flag handler.
+fn pre_parse_flag_present(flag: &str) -> bool {
+    std::env::args_os().skip(1).take_while(|a| a != "--").any(|a| a == flag)
+}
```
mablr, line 222: Pre-parse argv scan could misfire on literal values, as cast accept hyphen-leading values?

stevencartavia: [c4023c3](https://github.com/foundry-rs/foundry/pull/14770/commits/c4023c37784b470039537090469ceeb99018ebda)

crates/cli/src/introspect/build.rs
```diff
@@ -0,0 +1,407 @@
...
+    dups
+}
+
+/// Build an [`IntrospectDocument`] for `command` overlaid with metadata from
+/// `registry`.
+pub fn build_document(command: &Command, registry: &CommandRegistry) -> IntrospectDocument {
+    let binary = build_binary_info(command);
+    let mut commands = Vec::new();
+
+    let root_path = vec![command.get_name().to_string()];
```
mattsse, line 63: This only walks `command.get_subcommands()`, so root/global args are not exposed anywhere in the introspection document. I verified `forge --introspect` for `forge build` omits global flags such as `--json`, `--color`, `--quiet`, and `-j/--threads`, which means agents cannot reconstruct the real command surface from the document. Root/global args should either be represented separately or propagated into each command.

Permalink: https://github.com/foundry-rs/foundry/blob/c4023c37784b470039537090469ceeb99018ebda/crates/cli/src/introspect/build.rs#L57-L60

crates/cli/src/opts/global.rs
```diff
@@ -56,12 +56,46 @@ impl GlobalArgs {
...
         }
     }
 
+    /// Check if `--introspect` was passed and print the introspection document as JSON, then exit.
+    ///
+    /// Must run **before** clap parsing so required subcommands or args don't block discovery.
+    /// Uses `CommandRegistry::EMPTY`; for per-command metadata, call
+    /// [`check_introspect_with`](Self::check_introspect_with) with a populated registry.
+    pub fn check_introspect<C: clap::CommandFactory>() {
+        Self::check_introspect_with(C::command(), &crate::introspect::CommandRegistry::EMPTY)
```
mattsse: `check_introspect` builds `C::command()` before checking whether `--introspect` is present, so every normal CLI invocation now constructs the full clap tree once here and then constructs it again during `parse()`. This is especially noticeable for `cast`, where the PR already needs a larger stack just to build the clap tree in tests. Please check the flag first and only build the command lazily when it is actually requested.

Permalink: https://github.com/foundry-rs/foundry/blob/c4023c37784b470039537090469ceeb99018ebda/crates/cli/src/opts/global.rs#L78-L80

crates/cli/src/introspect/registry.rs
```diff
@@ -0,0 +1,141 @@
...
+
+impl Default for CommandMeta {
+    fn default() -> Self {
+        Self::DEFAULT
+    }
+}
+
+/// A binary's command registry.
+///
+/// Implemented as a thin wrapper over a `&'static` slice of (path, metadata) pairs to keep the call
```
mattsse, line 55: The registry is designed as a static slice, but the metadata it needs to hold is not const-friendly: `Capabilities` uses owned `String` schema refs and `ExitCodeInfo` owns its strings. A real non-empty static registry cannot author `Some("...".to_string())` or exit-code descriptions without adding lazy allocation elsewhere. Consider making registry metadata use `&'static str`/`Cow<'static, str>`, or separating the borrowed registry representation from the owned serialized document.

Permalink: https://github.com/foundry-rs/foundry/blob/c4023c37784b470039537090469ceeb99018ebda/crates/cli/src/introspect/registry.rs#L55-L59

crates/cli/src/opts/global.rs
```diff
@@ -56,12 +56,46 @@ impl GlobalArgs {
     /// This must be called **before** parsing arguments, since commands with required
     /// subcommands would fail parsing before the flag is checked.
     pub fn check_markdown_help<C: clap::CommandFactory>() {
-        if std::env::args().any(|arg| arg == "--markdown-help") {
+        if pre_parse_flag_present("--markdown-help") {
```
mattsse: This changes `--markdown-help` discovery from scanning all argv to only leading top-level options. That breaks existing usages where the flag appears after a subcommand or after top-level flattened args: I confirmed `forge build --markdown-help` and `anvil --port 0 --markdown-help` now error as unexpected arguments. Since the PR says markdown rendering is unchanged, please keep the old all-argv behavior for `--markdown-help` or explicitly scope this as a breaking behavior change.

Permalink: https://github.com/foundry-rs/foundry/blob/c4023c37784b470039537090469ceeb99018ebda/crates/cli/src/opts/global.rs#L58-L70

mattsse (commented):
Review comments from local inspection and focused test runs.

crates/cli/src/introspect/build.rs
stevencartavia: [00672ca](https://github.com/foundry-rs/foundry/pull/14770/commits/00672ca15349349352e1780241be5000bad3ee58)

crates/cli/src/opts/global.rs
stevencartavia: [00672ca](https://github.com/foundry-rs/foundry/pull/14770/commits/00672ca15349349352e1780241be5000bad3ee58)

crates/cli/src/introspect/registry.rs
stevencartavia: [00672ca](https://github.com/foundry-rs/foundry/pull/14770/commits/00672ca15349349352e1780241be5000bad3ee58)

crates/cli/src/opts/global.rs
stevencartavia: [00672ca](https://github.com/foundry-rs/foundry/pull/14770/commits/00672ca15349349352e1780241be5000bad3ee58)

crates/cli/src/introspect/build.rs
```diff
@@ -0,0 +1,431 @@
...
+    for sub in command.get_subcommands() {
```
mattsse, line 64: This only walks named subcommands, so root/default invocation args are still missing from introspection. A concrete example is `anvil`: its parser has `#[command(flatten)] pub global: GlobalArgs`, `#[command(flatten)] pub node: NodeArgs`, and only `completions` as a named subcommand. With the current builder, `anvil --introspect` reports `binary.global_args` for `verbosity`, `quiet`, `json`, etc. and the `anvil.completions` command, but it omits root options accepted by the primary invocation such as `anvil --port 8555`, `--accounts`, `--fork-url`, and `--host`.

Those `NodeArgs` should not be placed in `binary.global_args`: they are top-level root args, but not clap-global args, since `anvil completions bash --port 8555` is invalid. A safer fix is to add a root/default `CommandInfo` when the root command is invokable, with args from `command.get_arguments().filter(|a| !a.is_global_set() && !is_help_or_version(a))`, and use `registry.lookup(&[])` so `anvil` can pin that default invocation to something explicit like `anvil.start` (and `chisel` to `chisel.repl`).

mattsse (commented):
I found one issue with the current introspection shape: root/default invocations are not represented, so root-only flattened args are omitted from `anvil` and `chisel`.

crates/cli/src/introspect/build.rs
stevencartavia: [082a69f](https://github.com/foundry-rs/foundry/pull/14770/commits/082a69fe17b291d3ec253f70bcad1dc06c726d5f)

mattsse (approved):
LGTM

PR #14772, opened 2026-05-15
fix(chisel): display dynamic bytes memory layout

### LINKED ISSUE #5253: chisel: ABI-encoded values incorrectly displayed when dumped
### Component

Chisel

- [X] Foundry
- [X] Foundryup

chisel 0.1.1 (f9fc929 2023-06-29T00:05:24.689819239Z)

chisel

### Operating System

Linux

### Describe the bug

When dumping an ABI-encoded value in chisel, the entire value is put into the `Contents` section, this includes the words containing the offset and length. chisel will then derive it's own offset and length from this total value. 

Compare the following example in chisel vs `forge test` output: 
image
Note that `revertStringOne`'s offset `0x20` and length `0x0f` are both in the `Contents` and the offset and length as shown are derived from the total 96 bytes while both `decoded` and `revertStringTwo` have their own offset and length where none should exist. 

This forge test displays the values as one would expect.
image

maro20066600:
## Summary
- derive dynamic bytes/string memory display from the decoded value instead of slicing tuple ABI encoding
- keep tuple-encoded output as a separate section
- add a Chisel regression test for #5253

Closes #5253

## Testing
- cargo fmt --check
- git diff --check
- cargo test -p chisel abi_encoded_bytes_memory_display --test it -- --nocapture *(blocked locally: dependency fetch for ethereum-optimism/optimism failed with `No space left on device`; disk had only ~32MiB free before cleanup)*

### DIFF
--- a/crates/chisel/src/executor.rs
+++ b/crates/chisel/src/executor.rs
@@ -312,29 +312,18 @@ fn format_token(token: DynSolValue) -> String {
         DynSolValue::Bool(b) => {
             format!("Type: {}\n└ Value: {}", "bool".red(), b.cyan())
         }
-        DynSolValue::String(_) | DynSolValue::Bytes(_) => {
-            let hex = hex::encode(token.abi_encode());
-            let s = token.as_str();
-            format!(
-                "Type: {}\n{}├ Hex (Memory):\n├─ Length ({}): {}\n├─ Contents ({}): {}\n├ Hex (Tuple Encoded):\n├─ Pointer ({}): {}\n├─ Length ({}): {}\n└─ Contents ({}): {}",
-                if s.is_some() { "string" } else { "dynamic bytes" }.red(),
-                if let Some(s) = s {
-                    format!("├ UTF-8: {}\n", s.cyan())
-                } else {
-                    String::default()
-                },
-                "[0x00:0x20]".yellow(),
-                format!("0x{}", &hex[64..128]).cyan(),
-                "[0x20:..]".yellow(),
-                format!("0x{}", &hex[128..]).cyan(),
-                "[0x00:0x20]".yellow(),
-                format!("0x{}", &hex[..64]).cyan(),
-                "[0x20:0x40]".yellow(),
-                format!("0x{}", &hex[64..128]).cyan(),
-                "[0x40:..]".yellow(),
-                format!("0x{}", &hex[128..]).cyan(),
-            )
-        }
+        DynSolValue::String(s) => format_dynamic_token(
+            "string",
+            Some(&s),
+            s.as_bytes(),
+            DynSolValue::String(s.clone()).abi_encode(),
+        ),
+        DynSolValue::Bytes(bytes) => format_dynamic_token(
+            "dynamic bytes",
+            None,
+            &bytes,
+            DynSolValue::Bytes(bytes.clone()).abi_encode(),
+        ),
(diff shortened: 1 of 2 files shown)


crates/chisel/src/executor.rs
```diff
@@ -312,29 +312,18 @@ fn format_token(token: DynSolValue) -> String {
...
+            "string",
+            Some(&s),
+            s.as_bytes(),
+            DynSolValue::String(s.clone()).abi_encode(),
+        ),
+        DynSolValue::Bytes(bytes) => format_dynamic_token(
+            "dynamic bytes",
+            None,
+            &bytes,
+            DynSolValue::Bytes(bytes.clone()).abi_encode(),
```
mablr: Prob unnecessary clones on `s` and `bytes`.

crates/chisel/src/executor.rs
```diff
@@ -372,6 +361,35 @@ fn format_token(token: DynSolValue) -> String {
...
+    format!(
+        "Type: {}\n{}├ Hex (Memory):\n├─ Length ({}): {}\n├─ Contents ({}): {}\n├ Hex (Tuple Encoded):\n├─ Pointer ({}): {}\n├─ Length ({}): {}\n└─ Contents ({}): {}",
+        kind.red(),
+        if let Some(s) = utf8 { format!("├ UTF-8: {}\n", s.cyan()) } else { String::default() },
+        "[0x00:0x20]".yellow(),
+        format!("0x{length_hex}").cyan(),
+        "[0x20:..]".yellow(),
+        format!("0x{memory_contents_hex}").cyan(),
+        "[0x00:0x20]".yellow(),
+        format!("0x{}", &tuple_encoded_hex[..64]).cyan(),
```
mablr: `tuple_encoded_hex[..64]` still hardcodes the pointer offset ?

mablr (commented):
I think it's directionnally good, but it would be nice to address the 2 remarks below and add more tests. Maybe a 32-byte-boundary case (e.g. exactly 32 bytes, or 33 bytes to verify two content words) and an empty bytes case?

maro20066600:
Thanks for the review and guidance. Your notes helped clarify the right direction here. I addressed both remarks, removed the unnecessary clones, avoided slicing the tuple-encoded hex directly, and added the requested edge-case coverage for empty bytes, exactly 32 bytes, and 33 bytes.

maro20066600:
I checked the failing `Tempo / tempo-check` run. The Chisel tests were not the failing part; the job reached `Run Tempo check on devnet` with `ETH_RPC_URL` empty, then failed at `cast client --rpc-url "$ETH_RPC_URL"` with `Connection refused`.

That happens for fork PRs because `TEMPO_DEVNET_RPC_URL` is a repository secret and is not exposed to `pull_request` runs from forks. I pushed a small CI guard so the devnet Tempo steps still run on push/same-repo PR/manual dispatch, but are skipped for fork PRs where the secret is unavailable.

The new workflow run is currently waiting for maintainer approval because this is a fork PR.

mablr:
@maro20066600 please don't touch to `.github/*` files, the Tempo CI failure is not related to your changes.

mablr:
> The new workflow run is currently waiting for maintainer approval because this is a fork PR.

This was set on purpose to mitigate supply chain attacks (some attempts lately).

maro20066600:
Got it, sorry about touching the workflow. I cleaned up the branch and removed the `.github` change from this PR.

The PR now only contains the Chisel executor/test changes.

PR #14771, opened 2026-05-15
feat(cli): agent-machine runtime — --machine, ExitCode, diagnostics, capability guard

stevencartavia:
Stacked on #14770. Adds the runtime substrate the agent contract executes against.

## What this adds

- `--machine` global flag — the agent-contract opt-in. Pre-parse detected (via a shared `pre_parse_flag_present` scanner that handles `--`, subcommand boundary, value-taking globals, and non-UTF-8 argv). Forces no color. Mutually exclusive with `--json` and `--md`.
- `ExitCode` enum matching `docs/agents/exit-codes.md` (9 variants), with a conservative `From<eyre::Report>` heuristic.
- `DiagnosticCode` newtype + per-domain `&'static str` constants colocated by domain (`cli`, `config`, `compiler`, `network`, `wallet`, `test`, `script`, `cast`, `anvil`, `chisel`).
- Structured JSON envelopes for clap parse / usage / version / help under `--machine`. Subcommand `--help` preserves its own context (e.g. `cast call --machine --help` returns cast-call help, not root). Global runtime payload schemas `foundry:cli.help@v1` / `foundry:cli.version@v1` documented in `docs/agents/spec.md` §10.
- Hardened `capability_violations` static check — per-mode schema-ref rules (envelope/stream/session), `foundry:<stem>@vN` format, and stem-equality (`result` = `command_id`, `event` = `command_id.event`, `session` = `command_id.session`). Run per-binary in CI.
- Adopts canonical exit codes for `forge test` failure (`5`) and `anvil` Ctrl-C (`8`) — both gated on `--machine`. Legacy invocations remain `exit 1` and `exit 0`.

No command emits envelope/stream/session output yet — per-command adoption lands in follow-up PRs.

## Backward compatibility

Strictly additive. `main.rs` files unchanged. `--json` semantics unchanged. Every behavior change is gated on `--machine`.

## Manual sanity

- `forge --machine --version` → success envelope, exit `0`
- `forge --machine bogus` → error envelope with `cli.usage.invalid`, exit `2`
- `forge --machine --help` → success envelope wrapping help text, exit `0`
- `forge --json --machine build` → ArgumentConflict envelope, exit `2`
- `forge --version`, `forge bogus`, `forge test`: legacy paths unchanged

Part of OSS-218

PR #14774, opened 2026-05-15
feat(cli): adopt envelope output for forge build and cast call

stevencartavia:
Stacked on #14771. Adopts envelope output for two commands as a proof of the substrate:

- `forge build` — output_mode=envelope, foundry:forge.build@v1, side_effects=fs_write, requires_project
- `cast call` — output_mode=envelope, foundry:cast.call@v1, side_effects=network

Per-binary `CommandRegistry` wired through `check_introspect_with` so `--introspect` reports real metadata. Shape-changing flags are rejected under `--machine` (`--watch`/`--names`/`--sizes` for build; `--trace`/`--debug`/`--decode-internal`/`--curl` for call) with `cli.usage.invalid` + exit `2`. `--watch` is caught before dispatch via `BuildArgs::ensure_machine_compatible()`.

Typed failure envelopes for the primary error modes:

- `forge build --machine` compile failures → `compiler.solc.error` per solc error, exit `Build (4)`.
- `cast call --machine` RPC failures → `network.rpc.error` (new code) with eyre cause chain in `details.cause_chain`, exit `Network (6)`.

`forge build --machine` also skips `install_missing_dependencies`, pre-empts the `"Nothing to compile" + exit(0)` path with an empty envelope, and skips post-build lint at v1. `CallData.raw` replaces `CallData.result` — decoder bypassed, always raw `0x` hex.

All four binary `main.rs` files get a strictly-additive `is_machine()`-guarded handler. `report_machine_error` always tags `cli.unknown` and preserves the cause chain; typed call sites emit specific codes. Legacy `stderr + exit(1)` preserved for the human path. `check_machine()` runs before `setup()` so early failures are enveloped.

Part of OSS-218

### DIFF
--- a/crates/cast/src/args.rs
+++ b/crates/cast/src/args.rs
@@ -43,7 +43,7 @@ pub fn run() -> Result<()> {
     // Pre-parse discovery flags run before `setup()` so they cannot be blocked
     // by panic-handler / tracing init failures and avoid that init's cost.
     foundry_cli::machine::check_machine();
-    foundry_cli::opts::GlobalArgs::check_introspect_with(CastArgs::command(), &REGISTRY);
+    foundry_cli::opts::GlobalArgs::check_introspect_with(CastArgs::command, &REGISTRY);
     foundry_cli::opts::GlobalArgs::check_markdown_help::<CastArgs>();
 
     setup()?;
--- a/crates/cast/src/args.rs
+++ b/crates/cast/src/args.rs
@@ -1,6 +1,7 @@
 use crate::{
     Cast, SimpleCast,
     cmd::erc20::IERC20,
+    introspect::REGISTRY,
     opts::{Cast as CastArgs, CastSubcommand, ToBaseArgs},
     traces::identifier::SignaturesIdentifier,
     tx::CastTxSender,
@@ -42,7 +43,7 @@ pub fn run() -> Result<()> {
     // Pre-parse discovery flags run before `setup()` so they cannot be blocked
     // by panic-handler / tracing init failures and avoid that init's cost.
     foundry_cli::machine::check_machine();
-    foundry_cli::opts::GlobalArgs::check_introspect::<CastArgs>();
+    foundry_cli::opts::GlobalArgs::check_introspect_with(CastArgs::command(), &REGISTRY);
     foundry_cli::opts::GlobalArgs::check_markdown_help::<CastArgs>();
 
     setup()?;
@@ -944,7 +945,7 @@ pub async fn run_command(args: CastArgs) -> Result<()> {
 mod tests {
     use super::*;
     use foundry_cli::introspect::{
-        CommandRegistry, INTROSPECT_SCHEMA_ID, IntrospectDocument, build_document,
+        CommandRegistry, INTROSPECT_SCHEMA_ID, IntrospectDocument, OutputMode, build_document,
         capability_violations, duplicate_command_ids, render_introspect_document,
     };
 
@@ -962,7 +963,7 @@ mod tests {
             .stack_size(16 * 1024 * 1024)
             .spawn(|| {
                 let cmd = CastArgs::command();
-                let doc = build_document(&cmd, &CommandRegistry::EMPTY);
+                let doc = build_document(&cmd, &REGISTRY);
                 duplicate_command_ids(&doc)
             })
             .expect("spawn worker thread")
@@ -999,12 +1000,42 @@ mod tests {
             .stack_size(16 * 1024 * 1024)
             .spawn(|| {
                 let cmd = CastArgs::command();
-                let doc = build_document(&cmd, &CommandRegistry::EMPTY);
+                let doc = build_document(&cmd, &REGISTRY);
                 capability_violations(&doc)
             })
             .expect("spawn worker thread")
             .join()
             .expect("worker thread join");
         assert!(v.is_empty(), "cast capability violations: {v:?}");
+    }
+
+    /// Every adopted command (`output_mode = Envelope`) must pin a stable
+    /// `command_id` matching its registry entry.
+    #[test]
+    fn registered_commands_pin_stable_ids() {
+        let ids = std::thread::Builder::new()
+            .stack_size(16 * 1024 * 1024)
+            .spawn(|| {
+                let cmd = <CastArgs as clap::CommandFactory>::command();
+                let doc = build_document(&cmd, &REGISTRY);
+                fn walk(c: &foundry_cli::introspect::CommandInfo) -> Vec<String> {
+                    let mut out = Vec::new();
+                    if matches!(c.capabilities.output_mode, OutputMode::Envelope) {
+                        out.push(c.command_id.clone());
+                    }
+                    for sub in &c.subcommands {
+                        out.extend(walk(sub));
+                    }
+                    out
+                }
+                doc.commands.iter().flat_map(walk).collect::<Vec<_>>()
+            })
+            .expect("spawn worker thread")
+            .join()
+            .expect("worker thread join");
+        assert!(
+            ids.iter().any(|s| s == "cast.call"),
+            "cast.call missing from envelope ids: {ids:?}"
+        );
     }
 }
--- a/crates/cast/src/cmd/call.rs
+++ b/crates/cast/src/cmd/call.rs
@@ -59,7 +59,7 @@ pub struct CallData {
 }
 
 /// Emit a typed `network.rpc.error` envelope and exit `Network (6)`.
-fn rpc_failure(err: &eyre::Report) -> Result<()> {
+fn rpc_failure(err: &eyre::Report) -> ! {
     let cause_chain: Vec<String> = err.chain().map(ToString::to_string).collect();
     let message = cause_chain.first().cloned().unwrap_or_else(|| err.to_string());
     let envelope = JsonEnvelope::error(
@@ -243,12 +243,19 @@ impl CallArgs {
     pub async fn run(self) -> Result<()> {
         // Reject flags whose stdout shape or interactivity conflicts with the envelope contract.
         if foundry_cli::is_machine() {
+            // `--interactive` and the hardware / cloud signer flags can trigger TTY prompts,
+            // hardware confirmations, or SDK auth output (e.g. AWS SSO device-auth URLs printed
+            // directly to stdout), all of which would corrupt the envelope.
             let unsupported = [
                 ("--trace", self.trace),
                 ("--debug", self.debug),
                 ("--decode-internal", self.decode_internal),
                 ("--curl", self.rpc.curl),
                 ("--interactive", self.wallet.raw.interactive),
+                ("--ledger", self.wallet.ledger),
+                ("--trezor", self.wallet.trezor),
+                ("--aws", self.wallet.aws),
+                ("--gcp", self.wallet.gcp),
             ]
             .into_iter()
             .filter_map(|(name, on)| on.then_some(name))
@@ -259,6 +266,18 @@ impl CallArgs {
                      run without `--machine` or omit those flags.",
                     unsupported.join(", ")
                 ));
+            }
+            // Unlocked keystore: prompts for the password on TTY at signer construction.
+            let has_keystore =
+                self.wallet.keystore_path.is_some() || self.wallet.keystore_account_name.is_some();
+            let has_password = self.wallet.keystore_password.is_some()
+                || self.wallet.keystore_password_file.is_some();
+            if has_keystore && !has_password {
+                foundry_cli::machine::bail_machine_usage(
+                    "`cast call --machine` requires `--password` or `--password-file` when \
+                     `--keystore` / `--account` is set; an unlocked keystore would prompt on \
+                     stdin and corrupt the envelope.",
+                );
             }
         }
 
@@ -325,11 +344,11 @@ impl CallArgs {
 
         // Provider construction (transport setup) is a network operation, so connection /
         // DNS / TCP failures here are typed `network.rpc.error` under `--machine`.
-        let machine_mode_provider = foundry_cli::is_machine();
+        let machine_mode = foundry_cli::is_machine();
         let provider =
             match ProviderBuilder::<FEN::Network>::from_config(&config).and_then(|b| b.build()) {
                 Ok(p) => p,
-                Err(err) if machine_mode_provider => return rpc_failure(&err),
+                Err(err) if machine_mode => rpc_failure(&err),
                 Err(err) => return Err(err),
             };
         let sender = SenderKind::from_wallet_opts(wallet).await?;
@@ -368,7 +387,7 @@ impl CallArgs {
         .await;
         let (tx, func) = match builder_result {
             Ok(v) => v,
-            Err(err) if machine_mode_provider => return rpc_failure(&err),
+            Err(err) if machine_mode => rpc_failure(&err),
             Err(err) => return Err(err),
         };
 
@@ -486,14 +505,13 @@ impl CallArgs {
 
         // Bypass the ABI decoder under `--machine`; the envelope carries raw hex only.
         // Only the eth_call itself maps to `network.rpc.error`; setup/local errors fall through.
-        let machine_mode = foundry_cli::is_machine();
         let decode_func = (!machine_mode).then_some(func.as_ref()).flatten();
         let response = match Cast::new(&provider)
             .call(&tx, decode_func, block, state_overrides, block_overrides)
             .await
         {
             Ok(r) => r,
-            Err(err) if machine_mode => return rpc_failure(&err),
+            Err(err) if machine_mode => rpc_failure(&err),
             Err(err) => return Err(err),
         };
 
--- a/crates/cast/src/cmd/call.rs
+++ b/crates/cast/src/cmd/call.rs
@@ -266,18 +266,6 @@ impl CallArgs {
             return self.run_curl().await;
         }
 
-        // Under `--machine`, any failure past the flag-rejection check is an RPC
-        // interaction failure for `cast call@v1`. Funnel through one typed envelope.
-        let machine_mode = foundry_cli::is_machine();
-        let result = self.run_dispatch().await;
-        match result {
-            Ok(()) => Ok(()),
-            Err(err) if machine_mode => rpc_failure(&err),
-            Err(err) => Err(err),
-        }
-    }
-
-    async fn run_dispatch(self) -> Result<()> {
         if self.tx.tempo.is_tempo() {
             return self.run_with_network::<TempoEvmNetwork>().await;
         }
@@ -478,17 +466,25 @@ impl CallArgs {
         }
 
         // Bypass the ABI decoder under `--machine`; the envelope carries raw hex only.
+        // Only the eth_call itself maps to `network.rpc.error`; setup/local errors fall through.
         let machine_mode = foundry_cli::is_machine();
         let decode_func = (!machine_mode).then_some(func.as_ref()).flatten();
-        let response = Cast::new(&provider)
+        let response = match Cast::new(&provider)
             .call(&tx, decode_func, block, state_overrides, block_overrides)
-            .await?;
+            .await
+        {
+            Ok(r) => r,
+            Err(err) if machine_mode => return rpc_failure(&err),
+            Err(err) => return Err(err),
+        };
 
-        if response == "0x"
+        // Skip the empty-code lookup under `--machine`: it exists only for the human warning.
+        if !machine_mode
+            && response == "0x"
             && let Some(contract_address) = tx.to()
         {
             let code = provider.get_code_at(contract_address).await?;
-            if !machine_mode && code.is_empty() {
+            if code.is_empty() {
                 sh_warn!("Contract code is empty")?;
             }
         }
--- a/crates/cast/src/cmd/call.rs
+++ b/crates/cast/src/cmd/call.rs
@@ -16,7 +16,8 @@ use alloy_rpc_types::{
 use clap::Parser;
 use eyre::Result;
 use foundry_cli::{
-    json::{JsonEnvelope, print_json},
+    ExitCode, diagnostic,
+    json::{JsonEnvelope, JsonMessage, print_json},
     opts::{ChainValueParser, RpcOpts, TransactionOpts},
     utils::{LoadConfig, TraceResult, parse_ether_value},
 };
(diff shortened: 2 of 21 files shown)


mattsse (commented):
Review notes from local pass; leaving the concrete issues inline.

crates/cast/src/cmd/call.rs
```diff
@@ -422,19 +477,27 @@ impl CallArgs {
...
             let code = provider.get_code_at(contract_address).await?;
-            if code.is_empty() {
+            if !machine_mode && code.is_empty() {
                 sh_warn!("Contract code is empty")?;
             }
         }
-        sh_println!("{}", response)?;
+
+        if machine_mode {
+            print_json(&JsonEnvelope::success(CallData { raw: response }))?;
```
mattsse, line 530: This envelope can be suppressed by `--quiet`: [`print_json` writes through `sh_println!`](https://github.com/foundry-rs/foundry/blob/e8fd6ad8f0f6d42cfcb02f4c15a8812dd8857acd/crates/cli/src/json.rs#L136-L137), and `--machine` does not conflict with `--quiet` ([global flag definition](https://github.com/foundry-rs/foundry/blob/e8fd6ad8f0f6d42cfcb02f4c15a8812dd8857acd/crates/cli/src/opts/global.rs#L40-L41)). That means `cast --machine --quiet call ...` can exit successfully with no stdout envelope. The same applies to the new `forge build` machine envelopes, e.g. [`BuildData` success](https://github.com/foundry-rs/foundry/blob/e8fd6ad8f0f6d42cfcb02f4c15a8812dd8857acd/crates/forge/src/cmd/build.rs#L196-L197).

Suggested fix: have machine-envelope printing bypass the shell verbosity layer, for example by writing directly to `std::io::stdout()` in a machine-specific helper, or reject/ignore `--quiet` whenever `--machine` is set.

crates/cast/src/cmd/call.rs
```diff
@@ -219,10 +241,43 @@ pub enum CallSubcommands {
...
             return self.run_curl().await;
         }
+
+        // Under `--machine`, any failure past the flag-rejection check is an RPC
+        // interaction failure for `cast call@v1`. Funnel through one typed envelope.
+        let machine_mode = foundry_cli::is_machine();
+        let result = self.run_dispatch().await;
+        match result {
+            Ok(()) => Ok(()),
+            Err(err) if machine_mode => rpc_failure(&err),
```
mattsse: This maps every `run_dispatch()` failure to `network.rpc.error` ([wrapper](https://github.com/foundry-rs/foundry/blob/e8fd6ad8f0f6d42cfcb02f4c15a8812dd8857acd/crates/cast/src/cmd/call.rs#L271-L276)), but `run_dispatch` and `run_with_network` still perform local config/input work before the RPC call, including figment extraction, state override parsing, wallet setup, and calldata/ABI argument construction ([setup path](https://github.com/foundry-rs/foundry/blob/e8fd6ad8f0f6d42cfcb02f4c15a8812dd8857acd/crates/cast/src/cmd/call.rs#L285-L365)). Invalid local input will look like a retryable network failure to agents.

Suggested fix: only wrap the actual provider/RPC operations as `network.rpc.error`, and let local validation/config errors use `cli.usage.invalid`, `config.*`, or the generic fallback until more specific typed codes exist.

crates/cast/src/cmd/call.rs
```diff
@@ -422,19 +477,27 @@ impl CallArgs {
...
-            .call(&tx, func.as_ref(), block, state_overrides, block_overrides)
+            .call(&tx, decode_func, block, state_overrides, block_overrides)
             .await?;
 
         if response == "0x"
             && let Some(contract_address) = tx.to()
         {
             let code = provider.get_code_at(contract_address).await?;
-            if code.is_empty() {
+            if !machine_mode && code.is_empty() {
```
mattsse: In machine mode this still performs `eth_getCode` after a successful `0x` response, but the only consumer is the human warning now guarded by `!machine_mode` ([lookup and warning](https://github.com/foundry-rs/foundry/blob/e8fd6ad8f0f6d42cfcb02f4c15a8812dd8857acd/crates/cast/src/cmd/call.rs#L487-L493)). A successful `eth_call` can therefore be turned into a `network.rpc.error` by an unnecessary second RPC.

Suggested fix: skip the entire empty-code lookup under `--machine`, e.g. `if !machine_mode && response == "0x" { ... }`, so the envelope reflects the primary call result only.

crates/forge/src/cmd/build.rs
```diff
@@ -123,17 +166,38 @@ impl BuildArgs {
...
+                    .filter(|e| e.is_error())
+                    .map(|e| JsonMessage::error(SOLC_ERROR, e.to_string()))
+                    .collect();
+                print_json(&JsonEnvelope::<()>::failure(errors))?;
+                std::process::exit(ExitCode::Build.to_i32());
+            }
+
+            let payload = BuildData::from_output(&output);
+            print_json(&JsonEnvelope::success(payload))?;
+            return Ok(output);
```
mattsse, line 215: Returning here skips the existing post-build lint step ([machine return](https://github.com/foundry-rs/foundry/blob/e8fd6ad8f0f6d42cfcb02f4c15a8812dd8857acd/crates/forge/src/cmd/build.rs#L183-L198), [lint block](https://github.com/foundry-rs/foundry/blob/e8fd6ad8f0f6d42cfcb02f4c15a8812dd8857acd/crates/forge/src/cmd/build.rs#L201-L209)). With `lint_on_build = true` and `deny = "warnings"`/`"notes"`, normal `forge build` can fail on lint diagnostics while `forge build --machine` reports success.

Suggested fix: preserve the command's success semantics by running lint before emitting the success envelope and represent lint failures/warnings in the machine output, or explicitly reject/disable lint-on-build configs under `--machine` rather than returning a false success.

crates/cast/src/args.rs
```diff
@@ -42,7 +43,7 @@ pub fn run() -> Result<()> {
     // Pre-parse discovery flags run before `setup()` so they cannot be blocked
     // by panic-handler / tracing init failures and avoid that init's cost.
     foundry_cli::machine::check_machine();
-    foundry_cli::opts::GlobalArgs::check_introspect::<CastArgs>();
+    foundry_cli::opts::GlobalArgs::check_introspect_with(CastArgs::command(), &REGISTRY);
```
mattsse: This eagerly builds the full clap command tree on every invocation, even when `--introspect` is absent, because `CastArgs::command()` is evaluated before `check_introspect_with` can check the flag ([cast call site](https://github.com/foundry-rs/foundry/blob/e8fd6ad8f0f6d42cfcb02f4c15a8812dd8857acd/crates/cast/src/args.rs#L42-L46)). The same pattern was added for `forge` ([forge call site](https://github.com/foundry-rs/foundry/blob/e8fd6ad8f0f6d42cfcb02f4c15a8812dd8857acd/crates/forge/src/args.rs#L14-L18)), so normal startup now pays for an extra command-tree construction.

Suggested fix: make the registry-aware helper lazy, e.g. `check_introspect_with::<CastArgs>(&REGISTRY)` or `check_introspect_with(|| CastArgs::command(), &REGISTRY)`, and construct the command only after the pre-parse `--introspect` check succeeds.

crates/cast/src/cmd/call.rs
stevencartavia: [25a657f](https://github.com/foundry-rs/foundry/pull/14774/commits/25a657f7b4dcc380d897192e2d8f15358a995f4a)

crates/cast/src/cmd/call.rs
stevencartavia: [25a657f](https://github.com/foundry-rs/foundry/pull/14774/commits/25a657f7b4dcc380d897192e2d8f15358a995f4a)

crates/cast/src/cmd/call.rs
stevencartavia: [25a657f](https://github.com/foundry-rs/foundry/pull/14774/commits/25a657f7b4dcc380d897192e2d8f15358a995f4a)

crates/forge/src/cmd/build.rs
stevencartavia: [25a657f](https://github.com/foundry-rs/foundry/pull/14774/commits/25a657f7b4dcc380d897192e2d8f15358a995f4a)

crates/cast/src/args.rs
stevencartavia: [25a657f](https://github.com/foundry-rs/foundry/pull/14774/commits/25a657f7b4dcc380d897192e2d8f15358a995f4a)

crates/forge/src/cmd/build.rs
```diff
@@ -83,15 +86,58 @@ pub struct BuildArgs {
...
+        {
+            foundry_cli::machine::bail_machine_usage(
+                "`forge build` under `--machine` does not model lint diagnostics in the v1 \
+                 envelope; configs with `lint_on_build = true` and `deny != never` would \
+                 produce a false success. Pass `--no-lint` or disable lint-on-build / set \
+                 `deny = never` in foundry.toml.",
+            );
+        }
+
         self.check_soldeer_lock_consistency(&config).await;
```
grandizzy: these 2 soldeer locks can `sh_warn!` under `--machine`, let's gate on `!machine_mode`?

crates/cast/src/cmd/call.rs
```diff
@@ -422,19 +465,35 @@ impl CallArgs {
...
         }
 
-        let response = Cast::new(&provider)
-            .call(&tx, func.as_ref(), block, state_overrides, block_overrides)
-            .await?;
+        // Bypass the ABI decoder under `--machine`; the envelope carries raw hex only.
+        // Only the eth_call itself maps to `network.rpc.error`; setup/local errors fall through.
+        let machine_mode = foundry_cli::is_machine();
+        let decode_func = (!machine_mode).then_some(func.as_ref()).flatten();
+        let response = match Cast::new(&provider)
```
grandizzy, line 509: The narrowing went too far: only errors from `Cast::call(...).await` map to `network.rpc.error`. A connection-refused / DNS / TCP failure during provider construction now exits `cli.unknown + 1 `instead of `network.rpc.error + 6`. The new `cast_call_machine_mode_setup_failure_falls_through` test bakes that wrong semantic in by using http://127.0.0.1:1 and asserting cli.unknown.

crates/forge/tests/cli/build.rs
```diff
@@ -511,6 +512,132 @@ forgetest_init!(build_no_warning_without_foundry_lock, |prj, cmd| {
 "#]]);
 });
 
+// `forge --machine build` emits a single envelope on stdout. Asserts the
```
grandizzy: should we add test to assert empty stderr too?

crates/forge/src/cmd/build.rs
```diff
@@ -123,17 +181,38 @@ impl BuildArgs {
...
-            .bail(!format_json);
+            .bail(!format_json && !machine_mode);
+        if machine_mode {
+            compiler = compiler.quiet(true);
+        }
 
         let mut output = compiler.compile(&project)?;
 
         // Cache project selectors.
         cache_local_signatures(&output)?;
```
grandizzy, line 195: nit: `cache_local_signatures` writes function/event selectors to a global cache file on disk. If that write fails then the ? short-circuits and the error bubbles up to the top-level machine handler, which tags it `cli.unknown + exit 1`.
Suggested fix: check compiler errors before caching:

```rust
let mut output = compiler.compile(&project)?;

if machine_mode && output.has_compiler_errors() {
    // emit compiler.solc.error envelope, exit Build(4)
    ...
}

cache_local_signatures(&output)?;   // now only runs on success
```

grandizzy (dismissed):
left couple of comments, pls check. thanks!

crates/forge/src/cmd/build.rs
```diff
@@ -83,15 +86,58 @@ pub struct BuildArgs {
...
         }
 
+        // Lint output isn't modeled in the v1 envelope; refuse configs that would diverge
+        // human and machine success/failure outcomes.
+        if machine_mode
+            && !self.no_lint
+            && config.lint.lint_on_build
+            && config.deny != DenyLevel::Never
+        {
+            foundry_cli::machine::bail_machine_usage(
```
mablr: This seems to mask real solc errors as `cli.usage.invalid+2` instead of `compiler.solc.error+4`. Is this ok ?

crates/forge/src/cmd/build.rs
mablr, line 420: Should be gated ?

crates/cast/src/cmd/call.rs
mablr: I think it worths to mention wallet signer setup can prompt with `--interactive` for input under `--machine` violating non-interactive property.

crates/cast/src/diagnostic.rs
```diff
@@ -0,0 +1,31 @@
+//! Stable, machine-readable diagnostic codes emitted by `cast`.
+//!
+//! See [`docs/agents/diagnostics.md`](../../../docs/agents/diagnostics.md)
+//! for the format and registry rules. Most codes re-export per-domain
+//! constants from [`foundry_cli::diagnostic`].
+
+/// Diagnostic codes for read-only RPC commands like `cast call`, `cast tx`.
+pub mod rpc {
+    pub use foundry_cli::diagnostic::network::{RPC_TIMEOUT, RPC_UNAUTHORIZED};
```
mablr: Maybe `RPC_ERROR` belongs here ?

crates/forge/src/cmd/build.rs
stevencartavia: Done, both lock checks now gated on `!machine_mode`.

crates/cast/src/cmd/call.rs
stevencartavia: Widened scope; provider build and the `CastTxBuilder` chain now also map to `network.rpc.error`. The setup-failure test was replaced with `cast_call_machine_mode_rpc_failure_emits_network_envelope`.

crates/forge/tests/cli/build.rs
stevencartavia: Added `assert!(stderr.is_empty())` in `machine_mode_emits_envelope`.

crates/forge/src/cmd/build.rs
stevencartavia: Done, compile-error envelope emission moved before `cache_local_signatures`.

crates/forge/src/cmd/build.rs
stevencartavia: Good catch, moved the lint-divergence rejection after compile-error handling so failing builds still surface as `compiler.solc.error` + exit 4.

crates/forge/src/cmd/build.rs
stevencartavia: Yes, gated by skipping the entire `check_soldeer_lock_consistency` (and `check_foundry_lock_consistency`) call under `--machine`.

crates/cast/src/cmd/call.rs
stevencartavia: Added `--interactive` to the `cast call` unsupported-flags list under `--machine`.

crates/cast/src/diagnostic.rs
stevencartavia: Done, added `RPC_ERROR` to the `cast::diagnostic::rpc` re-export and the `ALL` slice.

crates/cast/src/cmd/call.rs
```diff
@@ -279,7 +323,15 @@ impl CallArgs {
             sig = Some(data);
         }
 
-        let provider = ProviderBuilder::<FEN::Network>::from_config(&config)?.build()?;
+        // Provider construction (transport setup) is a network operation, so connection /
+        // DNS / TCP failures here are typed `network.rpc.error` under `--machine`.
+        let machine_mode_provider = foundry_cli::is_machine();
```
mablr: Below we also have `let machine_mode = foundry_cli::is_machine();`, it seems kinda duplicate ?

crates/forge/src/cmd/build.rs
```diff
@@ -123,17 +170,51 @@ impl BuildArgs {
                     .map(ContractSizeLimits::with_runtime_limit)
                     .unwrap_or_default(),
             )
-            .bail(!format_json);
+            .bail(!format_json && !machine_mode);
```
mablr, line 173: Non-compiler `Err` from `compile()` may exit 1, not 4?

crates/cli/src/machine.rs
```diff
@@ -132,6 +132,28 @@ fn exit_code_for_clap_error(err: &clap::Error) -> ExitCode {
...
+/// `cli.unknown` and preserves the eyre cause chain in `details.cause_chain`.
+/// The process exit code is the caller's responsibility.
+pub fn report_machine_error(report: &eyre::Report) {
+    let cause_chain: Vec<String> = report.chain().map(ToString::to_string).collect();
+    let message = cause_chain.first().cloned().unwrap_or_else(|| report.to_string());
+    let envelope = JsonEnvelope::error(
+        JsonMessage::error(diagnostic::cli::UNKNOWN, message)
+            .with_details(json!({ "cause_chain": cause_chain })),
+    );
+    let _ = print_json(&envelope);
```
mablr, line 154: error is dropped, is this expected ?

crates/cast/src/cmd/call.rs
```diff
@@ -46,8 +48,28 @@ use foundry_evm::{
...
 
+/// Stable payload emitted in the `cast call` envelope under `--machine`.
+#[derive(Clone, Debug, Serialize)]
+pub struct CallData {
+    /// Raw `0x`-prefixed hex return data from `eth_call`.
+    pub raw: String,
+}
+
+/// Emit a typed `network.rpc.error` envelope and exit `Network (6)`.
+fn rpc_failure(err: &eyre::Report) -> Result<()> {
```
mablr:
```suggestion
fn rpc_failure(err: &eyre::Report) -> ! {
```
as `exit` is called here.

mablr:
The unsupported-flag check for `--interactive` tests the CLI flag value, but wallet resolution may trigger interactive prompting at runtime through other paths (e.g., a fallback that sets interactive mode implicitly). The upfront check won't catch that, and a prompt mid-machine-mode execution would write to stdout and corrupt the envelope.

crates/forge/src/cmd/build.rs
stevencartavia: Intentional, `compiler.solc.error` is reserved for solc diagnostics (gated on `has_compiler_errors()`), other Errs fall through to `cli.unknown` with the cause chain preserved.

crates/cli/src/machine.rs
stevencartavia: expected, that's the envelope's own stdout write on the error path

crates/cast/src/cmd/call.rs
stevencartavia: [8f2bd2b](https://github.com/foundry-rs/foundry/pull/14774/commits/8f2bd2becab81c0d4a5de365ec691da427c0a1d3)

crates/cast/src/cmd/call.rs
stevencartavia: [8f2bd2b](https://github.com/foundry-rs/foundry/pull/14774/commits/8f2bd2becab81c0d4a5de365ec691da427c0a1d3)

mattsse (dismissed):
LGTM

mablr (dismissed):

crates/cli/src/json.rs
```diff
@@ -131,10 +132,11 @@ impl JsonMessage {
...
 /// Prints a value as compact, single-line JSON to stdout.
 ///
-/// The trailing newline makes this suitable for NDJSON streams when each call
-/// emits one self-contained JSON record.
+/// Bypasses the shell verbosity layer so `--quiet` cannot suppress structured
+/// output the caller explicitly asked for.
 pub fn print_json<T: Serialize>(value: &T) -> Result<()> {
-    sh_println!("{}", to_string(value)?)?;
+    let mut stdout = std::io::stdout().lock();
+    writeln!(stdout, "{}", to_string(value)?)?;
```
grandizzy: should we `stdout.flush()?;` after? Call sites that immediately `std::process::exit` after `print_json` (e.g. `rpc_failure` https://github.com/foundry-rs/foundry/pull/14774/changes#diff-028cfa02373a6fc881e9ed97898d187d410102b0b4dbce90cca4e50964fecc97R70 bypass normal stdlib at-exit flushing. When stdout is piped (the agent case), the `BufWriter` inside `Stdout` can drop the envelope before the process exits, and the agent sees exit code + empty stdout.

grandizzy (dismissed):
---

### 🛑 `crates/cli/src/json.rs` — `print_json` must flush before exit

```suggestion
pub fn print_json<T: Serialize>(value: &T) -> Result<()> {
    let mut stdout = std::io::stdout().lock();
    writeln!(stdout, "{}", to_string(value)?)?;
    stdout.flush()?;
    Ok(())
}
```

`bail_machine_usage`, `rpc_failure`, and the `forge build` compile-error / empty-project paths all call `std::process::exit` immediately after `print_json`. That bypasses stdlib's at-exit flush. When stdout is piped (the agent case), the envelope can be dropped and the agent sees a non-zero exit with empty stdout — the worst failure mode for a contract whose premise is "stdout is exactly one JSON value."

`std::io::Write` is already imported in this file by the PR, so `flush()` resolves directly.

---

### 🛑 `crates/forge/src/introspect.rs` + `crates/cast/src/introspect.rs` — `exit_codes: &[]` ships wrong contract

```rust
exit_codes: &[],
```

Both new registry entries leave this empty. `--introspect` is the agent's source of truth for "what exit codes does this command produce" — and we're shipping adopted commands that explicitly produce typed non-zero exits without declaring them. Once this lands, fixing it means either silently changing introspect output (breaking agents that cached it) or a versioned migration.

Required:
- `forge.build`: `[Success(0), GenericError(1), Usage(2), Build(4)]`
- `cast.call`: `[Success(0), GenericError(1), Usage(2), Network(6), Wallet(7)]`

If empty is meant to mean "inherits global defaults," that needs to be specified in the schema and asserted by a test.

---

### 🛑 `crates/cast/src/cmd/call.rs:186-197` — env-var keystore passwords wrongly rejected

```rust
let has_password = self.wallet.keystore_password.is_some()
    || self.wallet.keystore_password_file.is_some();
if has_keystore && !has_password {
    foundry_cli::machine::bail_machine_usage(
        "`cast call --machine` requires `--password` or `--password-file` ...",
    );
}
```

Foundry supports keystore password via `ETH_PASSWORD` / `ETH_KEYSTORE_PASSWORD` env vars (see `foundry_wallets::WalletOpts`). This check only inspects CLI flags, so a perfectly valid non-interactive CI flow is rejected. Once agents work around this, the workaround becomes load-bearing and the fix becomes a behavior change.

Don't duplicate wallet resolution here — expose a helper on `WalletOpts` like `has_noninteractive_keystore_password()` that the wallet layer and this check both call.

mablr (approved):
Sgtm, btw we should keep an eye on the wallet’s interactive input.

mattsse (approved):

PR #14776, opened 2026-05-15
docs(architecture): add missing punctuation between cheatcodes link and its description

abhicris:
The single bullet under `evm/` in `docs/dev/architecture.md` runs the markdown link directly into the definition that follows:

> - [cheatcodes](./cheatcodes.md) a set of solidity calls dedicated to testing which can manipulate the environment in which the execution is run

That reads as a run-on. The parent paragraph ends with `… and has additional implementation of:`, so the bullet is a definition-style entry; a colon between the term and its description matches that setup and reads cleanly.

No semantic change.


figtracer (approved):

mattsse (commented):
fine

mattsse (approved):

PR #14777, opened 2026-05-15
feat(verify): preflight check verifier credentials before broadcasting

mablr:
## Motivation

Closes OSS-206

Validate verifier credentials before deployment so users get an early, actionable error instead of deploying and then having verification fail.

Two layers of validation are now applied in `forge script` and `forge create` before any transaction is broadcasted:
1. Presence check (`VerificationProviderType::client`): fails immediately if required credentials are obviously absent.
2. Connectivity check (`VerifierArgs::check_credentials`): makes a lightweight network call to confirm credentials are actually accepted.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


zerosnacks (dismissed):
Nice! Simple fix but saves a lot of hassle

crates/verify/src/verify.rs
```diff
@@ -64,6 +65,50 @@ impl VerifierArgs {
...
+                    config,
+                )?;
+                match client.contract_abi(Address::ZERO).await {
+                    Ok(_)
+                    | Err(
+                        EtherscanError::ContractCodeNotVerified(_)
+                        | EtherscanError::ContractNotFound(_),
+                    ) => {}
+                    Err(e) => eyre::bail!("verifier credential check failed: {e}"),
+                }
```
zerosnacks: Only `Ok`, `ContractCodeNotVerified`, and `ContractNotFound` are tolerated here. Any other `EtherscanError` — most notably `RateLimitExceeded`, transient 5xx, or generic network errors — will hard-fail the preflight and block deployment even when the credentials are perfectly valid. That re-introduces exactly the failure mode this PR is trying to prevent.

Suggestion: only bail on auth-related errors (`InvalidApiKey`, explicit 401/403 from the response), and downgrade everything else to a `sh_warn!` so the deploy can proceed. Otherwise users will hit flaky preflights on every Etherscan hiccup.

crates/script/src/broadcast.rs
```diff
@@ -731,16 +730,34 @@ impl<FEN: FoundryEvmNetwork> BundledState<FEN> {
...
-            if resolved == VerificationProviderType::Etherscan && etherscan_key.is_none() {
-                eyre::bail!("Missing etherscan key for chain {}", sequence.chain);
-            }
+            // Resolve the API key the same way verification does: CLI arg first, then config.
+            let api_key = self
+                .args
+                .verifier
+                .verifier_api_key
+                .clone()
+                .or_else(|| self.script_config.config.get_etherscan_api_key(Some(chain)));
```
zerosnacks: API-key resolution order here (`--verifier-api-key` first, then `get_etherscan_api_key`) is the inverse of what `forge create` does in [`create.rs#L367-L368`](https://github.com/foundry-rs/foundry/blob/632224cc86587ee9f1aa1bd4d5b2400a950dea5e/crates/forge/src/cmd/create.rs#L367-L368), where the etherscan key wins.

Same user inputs → different probe behavior between `script` and `create`. Worth aligning on a single precedence (the script order matches `EtherscanOpts::key()` elsewhere) and factoring this into a shared helper on `VerifierArgs` so it can’t drift again.

crates/verify/src/verify.rs
zerosnacks, line 221: No timeout on the Etherscan-style probe — the Sourcify branch below uses `Duration::from_secs(10)`, but `client.contract_abi(...)` here will use whatever default the block-explorers client has (which can be effectively unbounded for stalled connections). A hanging verifier endpoint will stall `forge create`/`script` indefinitely just because `--verify` was passed.

Consider wrapping the call in `tokio::time::timeout(Duration::from_secs(10), ...)` to match the Sourcify branch.

crates/verify/src/verify.rs
```diff
@@ -64,6 +65,50 @@ impl VerifierArgs {
...
+            }
```
zerosnacks, line 222: Using `client.contract_abi(Address::ZERO)` as the probe is reasonable for Etherscan and Blockscout, but is not guaranteed across every Etherscan-compatible provider (Oklink, arbitrary `--verifier custom` backends). Off-spec responses surface as non-matching `EtherscanError` variants and, combined with the bail in L94, become hard failures even when the credentials are valid.

A more robust probe is the Etherscan `stats` / `eth_blockNumber` endpoint (or a plain `HEAD` against the API root) — it doesn’t require an existing contract address and gives a cleaner auth-vs-network signal.

crates/forge/src/cmd/create.rs
```diff
@@ -362,7 +362,17 @@ impl CreateArgs {
 
         let context = verify.resolve_context().await?;
 
-        verify.verification_provider()?.preflight_verify_check(verify, context).await?;
+        verify.verification_provider()?.preflight_verify_check(verify.clone(), context).await?;
+
+        let api_key =
+            verify.etherscan.key.as_deref().or(verify.verifier.verifier_api_key.as_deref());
+        let chain = verify.etherscan.chain.unwrap_or_default();
```
zerosnacks: small: `verify.etherscan.chain.unwrap_or_default()` silently falls back to Mainnet if `chain_id()` were ever `None` on this path, which would probe the wrong network. The chain is already guaranteed to be `Some` by the early-resolution block at the top of `run()` (and reinforced by the [`expect("chain ID not resolved")`](https://github.com/foundry-rs/foundry/blob/632224cc86587ee9f1aa1bd4d5b2400a950dea5e/crates/forge/src/cmd/create.rs#L399) used in `deploy`), so it’s safer to `expect`/propagate here than to silently default — defaulting hides a real bug if the invariant ever breaks.

zerosnacks (commented):
A few preflight-correctness concerns worth addressing before merge — focused on cases where this check could itself become the deploy-blocker it’s meant to prevent.

crates/forge/tests/cli/verify.rs
```diff
@@ -360,3 +362,103 @@ Contract [src/Counter.sol:Counter] "0x19b248616E4964f43F611b5871CE1250f360E9d3"
...
+
+    assert!(!output.status.success(), "expected command to fail");
+    let stderr = output.stderr_lossy();
+    assert!(
+        stderr.contains("Verification preflight check failed"),
+        "expected preflight error in stderr, got: {stderr}"
+    );
+    // The command exiting non-zero with the preflight error message is sufficient evidence
+    // that no transactions were sent before the failure.
+});
```
zerosnacks, line 438: Coverage here only exercises the connection-refused path. If the bail-set in [`verify.rs#L88-L95`](https://github.com/foundry-rs/foundry/blob/632224cc86587ee9f1aa1bd4d5b2400a950dea5e/crates/verify/src/verify.rs#L88-L95) is tightened (e.g. only fail on `InvalidApiKey`, treat `RateLimitExceeded`/transient errors as warnings), it would be worth adding a small `wiremock`/local-server fixture so we lock in:

1. `ContractCodeNotVerified` response → preflight passes.
2. A 401/`InvalidApiKey` response → preflight fails with the credential error.
3. A 429/`RateLimitExceeded` response → preflight does **not** block the deploy.

That would prevent the tolerated-error contract from silently regressing in the future.

crates/verify/src/verify.rs
zerosnacks: Not actionable

crates/forge/tests/cli/verify.rs
mablr: added mock verifier server in https://github.com/foundry-rs/foundry/pull/14777/commits/4762420397ef6dfd9f0018cf5828e7103a475a01

crates/verify/src/verify.rs
```diff
@@ -64,6 +65,76 @@ impl VerifierArgs {
...
+                    }
+                    Ok(
+                        Ok(_)
+                        | Err(
+                            EtherscanError::ContractCodeNotVerified(_)
+                            | EtherscanError::ContractNotFound(_),
+                        ),
+                    ) => {}
+                    Ok(Err(
+                        EtherscanError::InvalidApiKey
```
figtracer: this treats Cloudflare blocks/challenges as credential failures, so a transient CDN challenge can block `forge create`/`script` before deployment even when the API key is valid

crates/verify/src/verify.rs
```diff
@@ -64,6 +65,76 @@ impl VerifierArgs {
...
+                        | EtherscanError::InvalidApiVersion
+                        | EtherscanError::BlockedByCloudflare
+                        | EtherscanError::CloudFlareSecurityChallenge,
+                    )) => {
+                        eyre::bail!("verifier credential check failed: invalid API key or blocked")
+                    }
+                    Ok(Err(e)) => {
```
figtracer: this catch-all warning also covers non-Etherscan-shaped 401/403 responses from custom verifiers, because the explorer client reads the body without `error_for_status()`. that lets rejected credentials pass preflight;

crates/verify/src/verify.rs
```diff
@@ -64,6 +65,133 @@ impl VerifierArgs {
...
+                        | EtherscanError::CloudFlareSecurityChallenge
+                        | EtherscanError::RateLimitExceeded
+                        // Version mismatch is a client-library issue, not a bad key.
+                        | EtherscanError::InvalidApiVersion,
+                    )) => {
+                        sh_warn!("verifier credential check inconclusive, proceeding anyway")?;
+                    }
+                    Ok(Err(e)) => {
+                        sh_warn!("verifier credential check failed: {e}, proceeding anyway")?;
+                    }
```
stevencartavia: does this print raw `reqwest::Error` / `EtherscanError` on transport failures which leaks `--verifier-api-key` into stderr because the URL contains `?apikey=...` ?

crates/verify/src/verify.rs
```diff
@@ -64,6 +65,133 @@ impl VerifierArgs {
...
+                    match reqwest::Client::new()
+                        .get(url)
+                        .query(&query)
+                        .timeout(Duration::from_secs(10))
+                        .send()
+                        .await
+                    {
+                        Err(e) => {
+                            sh_warn!("verifier credential check failed: {e}, proceeding anyway")?;
+                        }
```
stevencartavia: here too

crates/forge/tests/cli/verify.rs
```diff
@@ -360,3 +364,213 @@ Contract [src/Counter.sol:Counter] "0x19b248616E4964f43F611b5871CE1250f360E9d3"
...
+    let stdout = output.stdout_lossy();
+    assert!(
+        !stdout.contains("Contract Address"),
+        "contract was deployed but preflight check should have prevented it"
+    );
+});
+
+// Tests that the preflight check does NOT block deployment when the verifier responds
+// with a rate-limit error (transient, not an auth failure).
+forgetest_async!(create_preflight_warns_on_rate_limit, |prj, cmd| {
```
stevencartavia, line 550: Nit, worth double-checking: `--verifier custom` routes to the Custom branch (only checks 401/403 + "invalid api key" body), so `"Max rate limit reached"` is silently ignored — the `EtherscanError::RateLimitExceeded` warn never fires. Either rename, or use `--verifier blockscout` and assert stderr contains `"verifier credential check inconclusive"`

mablr: done https://github.com/foundry-rs/foundry/pull/14777/changes/278a137da2a8602e4a7852111d2b326bf7ca331e

crates/verify/src/verify.rs
mablr: Yes good catch, I removed url from err output https://github.com/foundry-rs/foundry/pull/14777/changes/278a137da2a8602e4a7852111d2b326bf7ca331e

crates/verify/src/verify.rs
```diff
@@ -64,6 +65,133 @@ impl VerifierArgs {
...
+                let etherscan_opts =
+                    EtherscanOpts { key: api_key.map(str::to_owned), chain: Some(chain) };
+                let client = EtherscanVerificationProvider::default().client(
+                    &etherscan_opts,
+                    self,
+                    config,
+                )?;
+                match tokio::time::timeout(
+                    Duration::from_secs(10),
+                    client.contract_abi(Address::ZERO),
```
stevencartavia: `client.contract_abi(...)` here calls into `foundry-block-explorers`, which logs `?err` at ERROR level — `reqwest::Error` Debug includes the full URL with `?apikey=...`, and Foundry's default tracing emits ERROR to stderr without `RUST_LOG`. Repro: `forge create --verify --verifier blockscout --verifier-url http://127.0.0.1:1/api --verifier-api-key SECRET` leaks `SECRET` to stderr before deployment. Easiest fix is to drop `contract_abi` and use a manual `reqwest` probe like the Custom branch below (and file an upstream issue on `foundry-block-explorers` to strip URLs from logged errors).

crates/verify/src/verify.rs
```diff
@@ -64,6 +65,133 @@ impl VerifierArgs {
...
+                    }
+                    match reqwest::Client::new()
+                        .get(url)
+                        .query(&query)
+                        .timeout(Duration::from_secs(10))
+                        .send()
+                        .await
+                    {
+                        Err(_) => {
+                            sh_warn!("verifier credential check failed, proceeding anyway")?;
```
stevencartavia, line 230: `reqwest::Client::new().get("not-a-url").send().await` returns `Err(_)`, which gets swallowed as the generic warn. So `forge script --broadcast --verify --verifier custom --verifier-url not-a-url` warns, broadcasts, then verification fails afterward — exactly the scenario this PR is meant to prevent. (`forge create` doesn't exhibit this because the existing `preflight_verify_check` parses the URL first.) Suggest `reqwest::Url::parse(url).wrap_err(...)?` before the probe so config errors hard-fail and only true transport errors warn.

crates/verify/src/verify.rs
```diff
@@ -64,6 +65,133 @@ impl VerifierArgs {
...
+            }
+            VerificationProviderType::Sourcify => {
+                // Only probe custom URLs; the default public endpoint is assumed reachable.
+                if let Some(url) = &self.verifier_url {
+                    let status = reqwest::Client::new()
+                        .get(url)
+                        .timeout(Duration::from_secs(10))
+                        .send()
+                        .await
+                        .map_err(|e| eyre::eyre!("Sourcify URL `{url}` is not reachable: {e}"))?
```
stevencartavia: This `?` makes any Sourcify transport error block the deploy, while Etherscan/Custom warn-and-proceed for the same condition. Repro: `--verifier sourcify --verifier-url http://127.0.0.1:1/api` exits 1; same URL with `--verifier custom` just warns. Suggest aligning the policy — but parse-validate the URL first, otherwise a malformed custom Sourcify URL would silently fall back to the default Sourcify endpoint via `SourcifyVerificationProvider::get_base_url`.

crates/forge/src/cmd/create.rs
```diff
@@ -369,7 +369,16 @@ impl CreateArgs {
 
         let context = verify.resolve_context().await?;
 
-        verify.verification_provider()?.preflight_verify_check(verify, context).await?;
+        verify.verification_provider()?.preflight_verify_check(verify.clone(), context).await?;
+
+        let api_key = verify.verifier.resolve_api_key(verify.etherscan.key.as_deref());
+        let chain = verify.etherscan.chain.expect("chain ID not resolved");
```
stevencartavia: Nit: unreachable today (chain is resolved in `CreateArgs::run` L122-127), but the same exact string is used 31 lines below with `.context("chain ID not resolved")?`. Worth using `.context(...)?` here too for consistency and to avoid a panic if a future refactor breaks the invariant.

mablr: https://github.com/foundry-rs/foundry/pull/14777/commits/588a6b5b5eb324d83535bb5f39c826324a553c4d

crates/verify/src/verify.rs
mablr: https://github.com/foundry-rs/foundry/pull/14777/commits/a3354fe22a70f0f7dc98310c536a6c3424668a34

crates/verify/src/verify.rs
mablr: https://github.com/foundry-rs/foundry/pull/14777/commits/a3354fe22a70f0f7dc98310c536a6c3424668a34

crates/verify/src/verify.rs
mablr: https://github.com/foundry-rs/foundry/pull/14777/commits/a3354fe22a70f0f7dc98310c536a6c3424668a34

crates/verify/src/verify.rs
```diff
@@ -64,6 +157,106 @@ impl VerifierArgs {
...
+                    Ok(Ok(VerifierCredentialProbe::Inconclusive) | Err(_)) => {
+                        sh_warn!("verifier credential check inconclusive, proceeding anyway")?;
+                    }
+                }
+            }
+            VerificationProviderType::Custom => {
+                // Custom verifiers may return Etherscan-shaped responses (HTTP 200 with a JSON
+                // body) or standard HTTP auth errors (401/403). Check both.
+                if let Some(url) = &self.verifier_url {
+                    let url = reqwest::Url::parse(url)
```
stevencartavia: `Url::parse` accepts any scheme, so `htps://`, `ws://`, `ftp://`, even `file:///etc/passwd` slip past the new validation. Reproduced via `forge script --broadcast --verify --verifier sourcify --verifier-url gopher://example.com/api` — broadcasts, then verification retries 5x with `URL scheme is not allowed`. Same scenario F2 was meant to prevent. Suggest also asserting `url.scheme()` is `http` or `https` (applies to the Sourcify branch below too).

crates/verify/src/verify.rs
```diff
@@ -21,7 +21,100 @@ use foundry_config::{
...
+        query.push(("apikey", api_key.to_string()));
+    }
+    query
+}
+
+fn classify_verifier_credential_response(
+    status: reqwest::StatusCode,
+    body: &str,
+) -> VerifierCredentialProbe {
+    if matches!(status, reqwest::StatusCode::UNAUTHORIZED | reqwest::StatusCode::FORBIDDEN) {
```
stevencartavia: 403 short-circuits to `InvalidApiKey` before body inspection runs, so a Cloudflare 403 with `"Sorry, you have been blocked"` returns `InvalidApiKey` instead of `Inconclusive` — even though the body classifier explicitly tries to handle that string. Reproduced end-to-end with a mock 403 + Cloudflare body: preflight bails with `verifier credential check failed: invalid API key`, blocking deploy for a user with a valid key behind a transient WAF challenge. Suggest inspecting the body first, or only short-circuiting on 401 and letting 403 fall through.

crates/verify/src/verify.rs
mablr: https://github.com/foundry-rs/foundry/pull/14777/commits/7c849cd200af4d3d78592bb3c6608f34982fa6c1

crates/verify/src/verify.rs
mablr: https://github.com/foundry-rs/foundry/pull/14777/commits/7c849cd200af4d3d78592bb3c6608f34982fa6c1

stevencartavia (dismissed):

figtracer (approved):

stevencartavia (approved):

PR #14778, opened 2026-05-15
feat(lint): detect external calls in loops

figtracer:
## Summary
- add the low-severity `calls-loop` lint
- detect external calls reachable from loops, including low-level calls, ETH `send`/`transfer`, contract/interface calls, `this.foo()`, and internal helpers called from a loop
- avoid flagging library-style member calls on non-contract values
- add lint documentation and UI coverage for flagged and ignored cases


crates/lint/src/sol/low/calls_loop.rs
```diff
@@ -0,0 +1,292 @@
...
+}
+
+fn is_address_like(hir: &Hir<'_>, expr: &Expr<'_>) -> bool {
+    match &expr.peel_parens().kind {
+        ExprKind::Payable(_) => true,
+        ExprKind::Ident(reses) => reses.iter().any(|res| {
+            res.as_variable().is_some_and(|var_id| {
+                matches!(
+                    hir.variable(var_id).ty.kind,
+                    TypeKind::Elementary(ty) if ty.to_abi_str() == "address"
```
mablr: `ElementaryType::Address(_)` pattern match is cleaner imo than `to_abi_str() == "address"`

crates/lint/src/sol/low/calls_loop.rs
```diff
@@ -0,0 +1,292 @@
...
+    matches!(
+        &expr.peel_parens().kind,
+        ExprKind::Ident(reses)
+            if reses.iter().any(|res| {
+                matches!(res, Res::Builtin(builtin) if builtin.name() == sym::this)
+            })
+    )
+}
+
+fn is_contract_like(hir: &Hir<'_>, expr: &Expr<'_>) -> bool {
```
mablr: It seems that receivers derived from returned values, mappings, or struct fields are missed: `getRecipient().foo()`, `m[addr].foo()`, `s.target.foo()`

mattsse (commented):
I found two false-negative cases worth addressing before this lands:

1. `crates/lint/src/sol/low/calls_loop.rs` ignores `StmtKind::Placeholder`, so calls executed through a modifier placeholder inside a loop are missed. For example, `modifier twice() { for (...) { _; } }` applied to `function f() external twice { receiver.ping(0); }` executes the external call in a loop, but the modifier pass sees only `_` at loop depth > 0 and the function pass sees the call at loop depth 0. The reentrancy lint expands modifier placeholders; this lint likely needs similar handling.

2. `is_internal_callable` skips `pure` internal/private helpers, but pure helpers can still make `external pure` interface calls. I verified `solc 0.8.30` accepts an internal `pure` helper that calls `IReceiver.ping(...) external pure`; if that helper is called from a loop, this lint won’t descend into it. This should either descend into pure helpers too or add coverage showing a deliberately narrower behavior.

crates/lint/src/sol/low/calls_loop.rs
```diff
@@ -0,0 +1,358 @@
...
+    match &expr.peel_parens().kind {
+        ExprKind::Call(callee, _, _)
```
mattsse: Explicit interface casts still appear to be missed here. `is_contract_like` only treats cast callees shaped as `ExprKind::Type(...)`, but the existing ERC20 unchecked-transfer matcher handles interface casts as `ExprKind::Ident([Res::Item(ItemId::Contract(_))])`; `IReceiver(target).ping(i)` inside a loop would not pass this branch and will be skipped. Can we add that HIR shape plus a regression test? This is in the documented contract/interface-call scope.

crates/lint/src/sol/low/calls_loop.rs
```diff
@@ -0,0 +1,358 @@
...
+        if self.emitted.insert(expr.span) {
+            self.ctx.emit(&CALLS_LOOP, expr.span);
+        }
+    }
+}
+
+fn is_external_call(hir: &Hir<'_>, callee: &Expr<'_>) -> bool {
+    let ExprKind::Member(base, member) = &callee.peel_parens().kind else { return false };
+
+    if matches!(member.name, kw::Call | kw::Delegatecall | kw::Staticcall) {
```
mattsse: Low-level call names are emitted solely by member name here. Solidity allows library extension methods named `call`, `delegatecall`, or `staticcall` on non-address values; `box.call()` in a loop would be reported as an external low-level call even though the docs say library-style member calls on non-contract values are ignored. Please gate these names on an address-like base or add coverage for the intended behavior.

mattsse (changes requested):
I found a couple of remaining edge cases that should be addressed before this lands.

p5hzehxa:
/

p5hzehxa:
#14794 في

p5hzehxa:
Duplicate of # #14778

p5hzehxa:
#14785

crates/lint/src/sol/low/calls_loop.rs
```diff
@@ -0,0 +1,371 @@
...
+    reses.iter().filter_map(|res| match res {
+        Res::Item(ItemId::Function(func_id)) if is_internal_callable(hir.function(*func_id)) => {
+            Some(*func_id)
+        }
+        _ => None,
+    })
+}
+
+const fn is_internal_callable(func: &Function<'_>) -> bool {
+    func.kind.is_function() && matches!(func.visibility, Visibility::Internal | Visibility::Private)
```
mablr: What about public fns ?

crates/lint/src/sol/low/calls_loop.rs
```diff
@@ -0,0 +1,371 @@
...
+    hir: &'hir Hir<'hir>,
+    callee: &Expr<'hir>,
+) -> Option<&'hir hir::Type<'hir>> {
+    let ExprKind::Ident(reses) = &callee.peel_parens().kind else { return None };
+    reses.iter().find_map(|res| {
+        let Res::Item(ItemId::Function(func_id)) = res else { return None };
+        let func = hir.function(*func_id);
+        let [ret] = func.returns else { return None };
+        Some(&hir.variable(*ret).ty)
+    })
```
mablr: It seems taht return-type inference only handles direct identifier callees, so chained calls won't be flagged.

crates/lint/src/sol/low/calls_loop.rs
```diff
@@ -0,0 +1,371 @@
...
+                    self.analyze_expr(arg, loop_depth);
+                }
+
+                if loop_depth > 0 {
+                    if is_external_call(self.hir, callee) {
+                        self.emit(expr);
+                    }
+                    for func_id in resolved_internal_function_ids(self.hir, callee) {
+                        self.analyze_internal_call(func_id, loop_depth);
+                    }
```
mablr, line 176: Internal callees are re-walked on every reachable loop path with no memoization; can blow up on DAG-shaped helper graphs. Idk if we can find a not-too-cursed solution to optimize this.

mattsse (dismissed):
Thanks, the current revision addresses the blocking cases I had called out and CI is green. Approving.

Non-blocking follow-up suggestion: this lint still has a few precision/coverage edges worth tightening after merge:
- descend into internally-called `public` functions as well as `internal`/`private` helpers, since unqualified public calls execute internally and can contain external calls;
- make internal helper resolution overload-aware, at least by arity and conservatively for ambiguous same-arity overloads, so unrelated overloads are not analyzed;
- avoid treating every member call on a contract/interface-typed receiver as external when it resolves to an internal library extension method, e.g. `using Lib for IReceiver`.

mattsse (dismissed):

crates/lint/src/sol/low/calls_loop.rs
```diff
@@ -0,0 +1,430 @@
...
+        Res::Item(ItemId::Function(func_id)) if is_internal_callable(hir.function(*func_id)) => {
+            Some(*func_id)
+        }
+        _ => None,
+    })
+}
+
+const fn is_internal_callable(func: &Function<'_>) -> bool {
+    func.kind.is_function() && matches!(func.visibility, Visibility::Internal | Visibility::Private)
+}
```
zerosnacks, line 335: `is_internal_callable` only matches `Visibility::Internal | Visibility::Private`, but `public` functions are also callable internally (e.g. `helper();` from within the same contract). This causes false negatives: if a loop calls a `public` helper that performs an external call, the helper is not traversed and the external call is missed. Consider including `Visibility::Public` here, matching how `delegatecall-loop` treats current-context helpers.

crates/lint/src/sol/low/calls_loop.rs
```diff
@@ -0,0 +1,430 @@
...
+    {
+        return true;
+    }
+
+    if matches!(member.name, sym::send | sym::transfer) && is_address_like(hir, base) {
+        return true;
+    }
+
+    is_this(base) || is_contract_like(hir, base)
+}
```
zerosnacks, line 310: `is_external_call` assumes any member call whose base is contract-typed is an external call, without checking what the resolved member actually is. With `using Lib for IContract;`, an internal library call like `contract.libFunction()` gets flagged as an external call (false positive verified locally). Consider resolving the called item and checking its visibility/execution context (or reusing the `Gcx`-based member resolution already in `delegatecall_loop.rs`) instead of inferring purely from the base type.

mablr (approved):
From @mattsse's comment
> make internal helper resolution overload-aware, at least by arity and conservatively for ambiguous same-arity overloads, so unrelated overloads are not analyzed;

seems only partially addressed, `member_return_type` resolves by `name + arity` and returns `None` on ambiguity rather than selecting the semantically resolved overload, and no overload-focused regression fixture exists.

Non-blocking as Matthias said, lgtm.

PR #14779, opened 2026-05-15
feat(ci): flag external PRs modifying `.github` folder

mablr:
## Motivation

We had lately a supply chain attack attempt on the CI trying to leak repo's secrets.

This new workflow aims to automatically add label and post a warning comment on any PR from an external contributor that touches the `.github` directory.


zerosnacks (changes requested):
Please do not merge, need to double check if pull_request_target is safe in this context

PR #14780, opened 2026-05-15
Use TUI fallback for debugger in non-interactive mode

cmende:
## Motivation
                                                                                                                                                                                                 
  `foundry-tui` now supports non-interactive fallback detection, but the debugger path was still always calling interactive `run_app`.                                                           
  In CI or non-TTY environments, `--debug` could fail without a clear, actionable fallback path.                                                                                                 
                                                                                                                                                                                                 
  ## Solution
                                                                                                                                                                                                 
  This PR updates the debugger TUI entrypoint to use `run_app_if_interactive` so behavior matches the shared TUI runtime contract.
  Interactive environments are unchanged; non-interactive environments now return a clear reasoned error message and suggest `--dump <PATH>` as the fallback workflow.                           
  A unit test was added to verify the fallback message includes both the reason and the dump hint.                                                                                               
                                                                                                                                                                                                 
  ## PR Checklist                                                                                                                                                                                

  - [x] Added Tests                                                                                                                                                                              
  - [ ] Added Documentation                                                                                                                                                                      
  - [ ] Breaking changes

### DIFF
--- a/crates/debugger/src/tui/mod.rs
+++ b/crates/debugger/src/tui/mod.rs
@@ -59,7 +59,7 @@ fn non_interactive_debugger_message(reason: TuiFallbackReason) -> String {
     )
 }
 
-fn debugger_dump_hint() -> &'static str {
+const fn debugger_dump_hint() -> &'static str {
     "Pass `--dump <PATH>` to export debugger steps."
 }
 
--- a/crates/debugger/src/tui/mod.rs
+++ b/crates/debugger/src/tui/mod.rs
@@ -66,6 +66,32 @@ fn debugger_dump_hint() -> &'static str {
 #[cfg(test)]
 mod tests {
     use super::{TuiFallbackReason, debugger_dump_hint, non_interactive_debugger_message};
+    use crate::{DebugNode, Debugger};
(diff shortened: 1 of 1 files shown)


crates/debugger/src/tui/mod.rs
```diff
@@ -36,6 +36,38 @@ impl<'a> TUI<'a> {
...
+                };
+                eyre::bail!("{message}");
+            }
+        }
+    }
+}
+
+fn non_interactive_debugger_message(reason: TuiFallbackReason) -> String {
+    format!(
+        "Cannot open the debugger TUI because {}. Re-run in an interactive terminal, or pass \
```
figtracer: can we avoid hard-coding `--dump <PATH>` in the shared debugger fallback message

cmende: Good call — I split the fallback wording so the shared non-interactive debugger message no longer hard-codes --dump <PATH>, and now append the dump guidance separately at the call site.

figtracer (dismissed):

mablr (dismissed):

figtracer (approved):

mablr (approved):

PR #14782, opened 2026-05-15
fix(npm): propagate wrapper exit codes

### LINKED ISSUE #14781: npm wrapper does not propagate child process exit codes
### Component

Foundry npm packages, specifically the meta package bin wrapper in `npm/src/bin.mjs`.

### Version

Observed with the published `@foundry-rs/forge@1.7.0` npm package.

### Platform

Observed on macOS arm64 locally and in a Linux CI job using the published npm package.

### What happened

The npm package wrapper starts the platform-specific Foundry binary with `child_process.spawn(..., { stdio: "inherit" })`, but it does not handle the child process exit result.

Because the Node wrapper has no remaining work after the child exits, Node exits with status `0` even when the underlying Foundry binary exits with a non-zero status.

This can make CI jobs continue after a failed Foundry command. For example, a test step can print failing tests, then the shell still treats the command as successful and continues to the next step.

### Minimal reproduction

Install the npm package locally and run an invalid forge subcommand through the npm wrapper:

```sh
./node_modules/.bin/forge definitely-not-a-command
echo $?
```

Expected:

```text
2
```

Actual:

```text
0
```

The platform-specific binary exits correctly:

```sh
./node_modules/@foundry-rs/forge-*/bin/forge definitely-not-a-command
echo $?
```

Expected and actual:

```text
2
```

### Expected behavior

The npm wrapper should exit with the same status as the child Foundry binary. If the child exits due to a signal, the wrapper should preserve that signal behavior.

### Why this matters

Many projects pin Foundry through npm so CI can use the same toolchain as local development. If the wrapper exits successfully after a failed `forge test`, `forge build`, or other command, CI can report a false green result.

igorroncevic:
Closes #14781

## Motivation

The npm meta package wrapper currently spawns the platform-specific Foundry binary, but it does not handle the child process exit result. That means the wrapper can exit with status `0` even when the underlying `forge`, `cast`, `anvil`, or `chisel` binary exits with a non-zero status.

For commands such as `forge test`, this can make CI continue after Foundry already printed failing tests.

## Solution

This PR updates `npm/src/bin.mjs` so the wrapper:

- exits with the same code as the child binary
- exits with `1` if the child result has no code
- preserves signal behavior when the child exits due to a signal
- cleans up installed signal handlers when the child exits or fails to spawn

I kept the focused regression test because `CONTRIBUTING.md` asks code fixes to include tests.

## Tests

```sh
cd npm
bun install --frozen-lockfile
bun tsc --project tsconfig.json --noEmit
node ./scripts/test-bin-exit.mjs
```

I also verified the wrapper against a real installed `forge` binary:

```sh
forge definitely-not-a-command
echo $?
# 2

NODE_PATH=/private/tmp/foundry-real-node-path/node_modules TARGET_TOOL=forge node ./src/bin.mjs definitely-not-a-command
echo $?
# 2
```

## AI disclosure

This PR was prepared with AI assistance. I used Codex to inspect the npm wrapper, reproduce the exit-code behavior, implement the patch, and draft the test and PR text. I reviewed the resulting change and validation steps.

### DIFF
--- a/npm/src/bin.mjs
+++ b/npm/src/bin.mjs
@@ -45,6 +45,23 @@ const signalHandlers = {
 for (const [signal, handler] of Object.entries(signalHandlers))
   process.on(signal, handler)
 
+child.on('error', error => {
+  cleanupSignalHandlers()
+  console.error(error)
+  process.exit(1)
+})
+
+child.on('exit', (code, signal) => {
+  cleanupSignalHandlers()
+
+  if (signal) {
+    process.kill(process.pid, signal)
+    return
+  }
+
+  process.exit(code ?? 1)
+})
+
 /**
  * Determines which tool wrapper is executing.
  * @returns {Tool}
@@ -149,4 +166,9 @@ function forwardSignal(signal) {
 
   process.off(signal, signalHandlers[signal])
   process.kill(process.pid, signal)
+}
+
+function cleanupSignalHandlers() {
+  for (const [signal, handler] of Object.entries(signalHandlers))
+    process.off(signal, handler)
 }
--- a/npm/scripts/test-bin-exit.mjs
+++ b/npm/scripts/test-bin-exit.mjs
@@ -1,0 +1,72 @@
+#!/usr/bin/env node
+
+import * as NodeAssert from 'node:assert/strict'
+import * as NodeChildProcess from 'node:child_process'
+import * as NodeFS from 'node:fs/promises'
+import * as NodeOS from 'node:os'
+import * as NodePath from 'node:path'
+import { fileURLToPath } from 'node:url'
+
+import { BINARY_NAME, PLATFORM_SPECIFIC_PACKAGE_NAME } from '#const.mjs'
+
+const __dirname = NodePath.dirname(fileURLToPath(import.meta.url))
+const tool = 'forge'
+const binaryName = BINARY_NAME(tool)
+const maybePlatformPackage = PLATFORM_SPECIFIC_PACKAGE_NAME(tool)
+
+if (!maybePlatformPackage)
+  throw new Error(`Unsupported platform for ${tool}: ${process.platform}/${process.arch}`)
+
+const platformPackage = maybePlatformPackage
+const tempDir = await NodeFS.mkdtemp(NodePath.join(NodeOS.tmpdir(), 'foundry-npm-bin-'))
+
+try {
+  await stageFakeBinary()
+  assertExitCode(0)
+  assertExitCode(7)
+} finally {
+  await NodeFS.rm(tempDir, { recursive: true, force: true })
+}
+
+async function stageFakeBinary() {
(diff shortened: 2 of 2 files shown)


npm/src/bin.mjs
```diff
@@ -45,6 +45,23 @@ const signalHandlers = {
 for (const [signal, handler] of Object.entries(signalHandlers))
   process.on(signal, handler)
 
+child.on('error', error => {
```
kaze-cow, line 48: I was curious what the addition of the error event was meant to accomplish. according to the [node.js docs](https://nodejs.org/api/child_process.html#event-error)

>  The 'error' event is emitted whenever:
>  *  The process could not be spawned.
>  *  The process could not be killed.
>  *  Sending a message to the child process failed.
>  *  The child process was aborted via the signal option.

In general its good that the error is printed for the user in each of these cases. IDK if its possible to test that this handler is working as expected via a script?

In the case that the process could not be killed, this could cause repeated execution of the foundry command to have an unexpected side effect where the port is already taken (ex. because the other process is still running). I don't know if there is a great solution here within the script but something to consider I guess. (I have had issues with this for other personal projects)



In any case, I don't believe this change was adequately explained in the PR description or is directly relevant to the original issue leading to this PR, but would be good to list this handler as an additional change.

npm/src/bin.mjs
```diff
@@ -45,6 +45,23 @@ const signalHandlers = {
...
+  cleanupSignalHandlers()
+  console.error(error)
+  process.exit(1)
+})
+
+child.on('exit', (code, signal) => {
+  cleanupSignalHandlers()
+
+  if (signal) {
+    process.kill(process.pid, signal)
```
kaze-cow, line 58: tbh, im not sure what repeating the signal that killed the child on the host process is meant to accomplish, and in fact, it could lead to the actual exit code being lost.

kaze-cow (commented):
thanks for submitting this!

npm/src/bin.mjs
igorroncevic: This is definitely not solving the main issue of binary not exiting with a proper exit code (like the latter code snippet does), but it still solves a similar issue where the error code is not propagated as it's supposed to (specifically, binary not starting).

Agreed that it should be stated in the description more clearly!

npm/src/bin.mjs
igorroncevic: The reason for sending the same signal to the wrapper is to preserve the child process result. If the child exits with a code, we return that code. If the child was killed by a signal, there usually is no exit code to preserve - Node gives code = null and signal = ...

So this branch is trying to avoid converting a signal termination into a misleading normal exit. It makes the npm wrapper reflect what happened to the actual Foundry process.

kaze-cow: til shells like bash "mask" processes killed by signal into a exit code, so if running a child process ex. in node it receives signals instead. makes sense!

zerosnacks:
We've decided to deprecate the NPM publishing flow, regardless - thanks for your contribution

PR #14783, opened 2026-05-15
feat(evm): symbolic-assist worker

figtracer:
## What this is

A directed-mutation worker for the existing coverage-guided fuzzer. Each cycle: replay a corpus seed under a branch-trace inspector, find unseen opposite-side `JUMPI` branches whose preceding compare opcode operands are visible, propose ABI-aware scalar rewrites that would flip the branch, validate by re-running through a cloned executor (must produce a real new EVM edge), and insert accepted candidates directly into the master's in-memory corpus.

No SMT solver, no symbolic execution. Only flips branches whose compare operands surface at runtime and are reachable by rewriting a scalar ABI argument. Keccak-of-input and similar non-invertible guards are out of scope and belong in the symbolic engine work (#14796).

## Benchmark

Real-world benchmark (contracts, profiles, raw per-seed results): https://gist.github.com/figtracer/592995f52e002710205aa91f5db15018


figtracer:
Closing: after correctness fixes (stable edge hashing, faithful replay, per-entry verbatim flag), the worker no longer shows a clear advantage over the existing coverage-guided fuzzer on the real-world benchmark (https://gist.github.com/figtracer/592995f52e002710205aa91f5db15018). Symbolic work belongs in #14796.

PR #14784, opened 2026-05-15
feat(lint): add `assert-state-change`

mablr:
## Motivation

Add `assert-state-change` lint to detect state-modifying expressions inside `assert()` arguments.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/lint/src/sol/med/assert_state_change.rs
```diff
@@ -0,0 +1,160 @@
...
+                UnOpKind::PreInc | UnOpKind::PostInc | UnOpKind::PreDec | UnOpKind::PostDec
+            ) =>
+        {
+            if lvalue_is_state_var(hir, inner) {
+                return Some(expr.span);
+            }
+            find_state_change(hir, inner)
+        }
+
+        ExprKind::Call(callee, args, named_args) => {
```
mattsse, line 113: This only treats `push`/`pop` and bare identifier functions as mutating calls. External/member calls inside `assert`, e.g. `assert(this.deposit())`, `assert(payable(to).send(1 ether))`, or a token/interface call with side effects, fall through to recursion and are not reported. Since the lint is documented as catching function calls that write state, can we either detect mutating external/member calls or narrow the docs/tests to the intended scope?

mattsse (changes requested):
One remaining behavior gap before this lands.

crates/lint/src/sol/med/assert_state_change.rs
```diff
@@ -0,0 +1,204 @@
...
+        | ExprKind::Err(_) => None,
+    }
+}
+
+/// Resolves a member-call callee expression to a HIR `FunctionId` if the base is a
+/// typed contract variable or interface cast, enabling `mutates_state()` checks.
+fn resolve_member_function<'hir>(hir: &Hir<'hir>, callee: &'hir Expr<'hir>) -> Option<FunctionId> {
+    let ExprKind::Member(base, method) = &callee.peel_parens().kind else { return None };
+    let cid = contract_id_of(hir, base)?;
+    hir.contract_item_ids(cid).find_map(|item| {
```
figtracer: resolves member calls by first same-name function only

overloads can FP/FN, e.g. `check(bytes)` mutates and `check(uint256)` is view, then `assert(t.check(1))` depends on source order instead of the selected overload

is this desired?

mablr: good point

crates/lint/src/sol/med/assert_state_change.rs
```diff
@@ -0,0 +1,217 @@
...
+                return Some(expr.span);
+            }
+
+            // Known always-mutating member calls: .call(), .delegatecall(), .send(), .transfer()
+            if let ExprKind::Member(_, method) = &callee.kind {
+                let n = method.name;
+                if n == kw::Call || n == kw::Delegatecall || n == sym::send || n == sym::transfer {
+                    return Some(expr.span);
+                }
+            }
```
stevencartavia: this matches only on the method name, so it fires on any member call named `call`/`send`/`transfer`/`delegatecall` regardless of the receiver type.

crates/lint/src/sol/med/assert_state_change.rs
```diff
@@ -0,0 +1,217 @@
...
+            let fid = item.as_function()?;
+            let f = hir.function(fid);
+            (f.name.is_some_and(|n| n.name == method.name) && f.parameters.len() == arg_count)
+                .then_some(fid)
+        })
+        .collect()
+}
+
+/// Extracts the contract ID from an expression that is a contract variable or interface cast.
+fn contract_id_of<'hir>(hir: &Hir<'hir>, expr: &'hir Expr<'hir>) -> Option<ContractId> {
```
stevencartavia: `using for` extension calls aren't detected because `contract_id_of` only handles contract-typed receivers — e.g. `assert(items.bump())` where bump is a library extension writing to storage is silently missed.

crates/lint/src/sol/med/assert_state_change.rs
```diff
@@ -0,0 +1,268 @@
...
+        }
+
+        ExprKind::Call(callee, args, named_args) => {
+            // arr.push(...) / arr.pop() on a state variable are mutations
+            if let ExprKind::Member(base, method) = &callee.kind
+                && (method.name == sym::push || method.name.as_str() == "pop")
+                && lvalue_is_state_var(hir, base)
+            {
+                return Some(expr.span);
+            }
```
zerosnacks, line 124: This matches `push`/`pop` purely by method name + state-var receiver, with no overload/all-must-mutate guard like the other call branches. `push`/`pop` are very common queue/stack/registry interface method names, so this will false-positive on real code, e.g.:

```solidity
interface IQueue {
    function pop() external view returns (bool);
}
IQueue public q;
assert(q.pop()); // flagged, but it's a view interface call, not a builtin array pop
```

Suggestion: only treat `push`/`pop` as builtin mutations when the receiver type is actually a storage array / `bytes` (i.e. the types that support the builtin members), or fall through to `resolve_member_overloads` first so contract/interface receivers are excluded from this name-only path.

crates/lint/src/sol/med/assert_state_change.rs
```diff
@@ -0,0 +1,268 @@
...
+            if !candidates.is_empty()
+                && candidates.iter().all(|&fid| hir.function(fid).mutates_state())
+            {
+                return Some(expr.span);
+            }
+
+            if let ExprKind::Member(base, method) = &callee.kind
+                && lvalue_is_state_var(hir, base)
+            {
+                let lib_candidates = resolve_library_extension(hir, method.name, args.len());
```
zerosnacks: This fallback (plus `resolve_library_extension` below) matches library functions purely by **method name + arity + first-param-is-storage**, with no receiver-type check, and scans every library in the whole compilation unit. An unrelated library can poison a normal contract/interface call:

```solidity
library QueueLib {
    function add(Queue storage q, uint256 x) internal returns (bool) { /* writes storage */ }
}

contract MyThing {
    IRegistry public reg;       // IRegistry.add(uint256) is view
    function f(uint256 x) external view {
        assert(reg.add(x));     // FP: QueueLib.add matches by name+arity+storage receiver
    }
}
```

Suggested narrowing: (1) require the library's first param type to match the receiver type, (2) only run this fallback when contract/interface overload resolution returned nothing, and (3) memoize if kept — `hir.function_ids()` over the whole unit per eligible call scales poorly. If precise extension-method resolution isn't yet available from Solar, I'd lean toward dropping this fallback for now rather than shipping a broad heuristic at Med severity.

crates/lint/src/sol/med/assert_state_change.rs
```diff
@@ -0,0 +1,268 @@
...
+            }
+
+            // Low-level address calls (.call/.delegatecall/.send/.transfer) are always mutating.
+            // Only apply this name-based heuristic when the receiver does NOT resolve to a known
+            // contract/interface, those are handled precisely by `resolve_member_overloads`,
+            // avoiding false positives on interface methods that happen to share these names.
+            if let ExprKind::Member(base, method) = &callee.kind {
+                let n = method.name;
+                if (n == kw::Call || n == kw::Delegatecall || n == sym::send || n == sym::transfer)
+                    && contract_id_of(hir, base).is_none()
```
zerosnacks: The guard here uses `contract_id_of(...).is_none()` to mean "receiver is not a contract", but `contract_id_of` only recognizes two narrow syntactic forms: a plain `Ident` of contract type, and an explicit `IFace(addr)` cast. Anything else returns `None` and falls into the low-level heuristic — even when the receiver's *type* is a contract/interface.

This PR's own test fixture demonstrates the precondition for the FP class: `IRouter` defines view methods named `call`/`send`/`transfer` (testdata `AssertStateChange.sol` L97-101). The existing tests cover only the state-var-receiver form, which `contract_id_of` matches. Any other shape of contract-typed receiver hits the heuristic:

```solidity
function getRouter() internal view returns (IRouter) { ... }
assert(getRouter().call(data));        // FP: Call receiver, not Ident

struct Cfg { IRouter router; }
Cfg cfg;
assert(cfg.router.send(x));            // FP: Member chain, not Ident

assert(this.someInterfaceCall(...));   // FP: `this`, not Ident
```

The right gate is **"receiver is address-like"** (`address` / `address payable` / result of `payable(...)`), not "receiver isn't a recognizable contract Ident." Driving this from the receiver's resolved type rather than syntactic pattern would eliminate the whole class.

(Note: `tokens[i].transfer(to, amt)` on `IERC20[]` is *not* a FP — that call is genuinely mutating — but it's flagged via this fragile path instead of the precise overload-resolution path below.)

crates/lint/src/sol/med/assert_state_change.rs
```diff
@@ -0,0 +1,399 @@
...
+    callee: &'hir Expr<'hir>,
+    arg_count: usize,
+) -> Vec<FunctionId> {
+    let ExprKind::Member(base, method) = &callee.peel_parens().kind else { return vec![] };
+    let Some(cid) = contract_id_of(hir, base) else { return vec![] };
+    hir.contract_item_ids(cid)
+        .filter_map(|item| {
+            let fid = item.as_function()?;
+            let f = hir.function(fid);
+            (f.name.is_some_and(|n| n.name == method.name) && f.parameters.len() == arg_count)
```
figtracer, line 239: Matching overloads by name + arity still misses real mutating calls. If `update(uint256)` mutates and `update(address)` is `view`, `assert(overloaded.update(x))` resolves to the mutating overload but `all(mutates_state)` is false because a same-arity view overload exists. I live-tested this and got no warning.

crates/lint/src/sol/med/assert_state_change.rs
```diff
@@ -0,0 +1,399 @@
...
+        }
+        _ => false,
+    }
+}
+
+/// Returns `true` if the lvalue expression ultimately targets a storage variable.
+/// Peels through index, slice, member, and payable wrappers to find the root identifier.
+fn lvalue_is_state_var(hir: &Hir<'_>, expr: &Expr<'_>) -> bool {
+    match &expr.peel_parens().kind {
+        ExprKind::Ident([Res::Item(ItemId::Variable(id)), ..]) => {
```
figtracer, line 507: lvalue_is_state_var` misses local storage aliases. `uint256[] storage xs = items; assert((xs[0] = 1) > 0);` and `mapping(address => uint256) storage m = balances; assert((m[user] = 1) > 0);` both mutate contract state, but I live-tested them and got no warnings.

mattsse:
Found a few receiver-resolution cases that still look actionable:

1. `crates/lint/src/sol/med/assert_state_change.rs` around line 199: member-call resolution only handles receivers that are plain contract-typed identifiers or explicit interface casts. Mutating calls through common contract-typed expressions fall through silently, e.g. `tokens[i].mutate()`, `byUser[user].mutate()`, struct fields, function-returned contracts, or `this.selfMutate()`. I reproduced the first three locally and got no `assert-state-change` warning.

2. `crates/lint/src/sol/med/assert_state_change.rs` around line 346: the `Index` recursion makes indexed arrays of interfaces/contracts look like builtin dynamic arrays. For `IQueue[] queues`, `assert(queues[i].pop())` is a view interface call on the element, but `is_dynamic_array_or_bytes` recurses to the container and returns true, so the lint reports it as a state-changing builtin `pop`. I reproduced this false positive locally.

3. `crates/lint/src/sol/med/assert_state_change.rs` around line 247: the library-extension fallback scans every library function in the HIR and matches by method name, arity, storage first param, and receiver type, but not by the actual `using for` binding or selected extension. An unrelated mutating `Lib.bump(uint256[] storage)` elsewhere in the compilation unit can make `items.bump()` warn even when the selected extension is a view helper.

mablr (commented):
@mattsse I fixed 1. and 2. in https://github.com/foundry-rs/foundry/pull/14784/commits/f30b6a327a47f3e148c80592aafaa85df4d2d106, don't think 3. is not feasible with Solar as-is (no using-for bindings in HIR).

mattsse (commented):
Two remaining receiver-resolution cases:

1. [`resolve_member_overloads`](https://github.com/foundry-rs/foundry/blob/ca35d7788c3c11794b29d97376d1e7f1889c9c13/crates/lint/src/sol/med/assert_state_change.rs#L199) still only resolves mutating contract/interface member calls when the receiver is a plain contract variable, explicit interface cast, or index expression. Calls through common contract-typed expressions such as `cfg.token.transfer(...)`, `getToken().transfer(...)`, or `this.mutate()` fall through without warning even when the selected function is non-view. The same narrow receiver typing affects low-level address calls in [`is_address_like`](https://github.com/foundry-rs/foundry/blob/ca35d7788c3c11794b29d97376d1e7f1889c9c13/crates/lint/src/sol/med/assert_state_change.rs#L371), e.g. `payees[user].send(...)` for an address mapping value.

2. [`resolve_library_extension`](https://github.com/foundry-rs/foundry/blob/ca35d7788c3c11794b29d97376d1e7f1889c9c13/crates/lint/src/sol/med/assert_state_change.rs#L249) resolves `using for` extension calls by scanning every library function in the HIR by name, arity, receiver type, and mutability, but not by the actual `using` binding. An unrelated mutating `MutLib.bump(uint256[] storage)` elsewhere in the compilation unit can make `assert(items.bump())` warn even when the selected extension is a view helper from another library.

mablr (commented):
@mattsse I fixed the 1st point https://github.com/foundry-rs/foundry/pull/14784/commits/3c042965ed7292600a8e03a62070ef14028be58e, for the 2nd I think it's better make a follow-up when Solar hir will support `using for` completely .

mattsse (approved):
Reviewed the updated assert-state-change implementation and fixtures again. The receiver/overload fixes address the prior blockers; the remaining using-for exact-binding limitation is acceptable as follow-up given current Solar HIR support. Local forge-lint checks passed, and CI/CI MPP are green on the latest head.

PR #14785, opened 2026-05-15
feat(lint): add `unused-return`

mablr:
## Motivation

Add `unused-return` lint to flag if return value of an external call is not used.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


mattsse (dismissed):
Looks good from this pass.

crates/lint/src/sol/med/unused_return.rs
```diff
@@ -0,0 +1,116 @@
...
+            &hir.variable(var_id).ty.kind,
+            TypeKind::Elementary(ty) if ty.to_abi_str() == type_str
+        )
+    };
+
+    let ExprKind::Call(callee, call_args, ..) = &expr.kind else { return false };
+    let ExprKind::Member(contract_expr, func_ident) = &callee.kind else { return false };
+
+    // Arity from either positional or named args.
+    let arity = call_args.kind.len();
```
figtracer, line 42: named args are missed here

`oracle.getPrice({token: t});` has positional arity 0, so candidate lookup skips the one-param function even though the return is discarded

mablr: It is handled properly, added a test case to confirm https://github.com/foundry-rs/foundry/pull/14785/commits/d6b0d6e41af4068a8c210f953b42cd3ce5a4efbf.

figtracer (approved):

mattsse (approved):

PR #14786, opened 2026-05-15
perf(fuzz): remove locks from invariant dictionaries

0xalpharush:
## Motivation

  Invariant fuzzing currently pays lock overhead on hot dictionary and target-contract paths even though each invariant campaign runs locally within a single
  test invocation.

  ## Solution

  - Make invariant fuzz state worker-local and lock-free.
  - Replace shared dictionary locking with local mutable state for invariant campaigns.
  - Remove target-contract locking from invariant execution paths.
  - Keep inspector-collected stack words staged locally, bounded by the configured dictionary value cap, and drain them after each executed tx, including
    discarded vm.assume runs.
  - Keep stateless fuzz using forked per-worker dictionary state.

  Aave invariant benchmark, 3 reps, 120s timeout:

  - baseline 5454d35d8: 37,967.8 elapsed tx/s mean
  - this branch: 57,637.7 elapsed tx/s mean
  - improvement: +51.8% elapsed tx/s

### DIFF
--- a/crates/evm/evm/src/executors/invariant/mod.rs
+++ b/crates/evm/evm/src/executors/invariant/mod.rs
@@ -29,7 +29,7 @@ use foundry_evm_fuzz::{
         ArtifactFilters, FuzzRunIdentifiedContracts, InvariantContract, InvariantSettings,
         RandomCallGenerator, SenderFilters, TargetedContract, TargetedContracts,
     },
-    strategies::{EvmFuzzState, invariant_strat, override_call_strat},
+    strategies::{EvmFuzzState, InvariantFuzzState, invariant_strat, override_call_strat},
 };
 use foundry_evm_traces::{CallTraceArena, SparsedTraceArena};
 use indicatif::ProgressBar;
@@ -281,7 +281,7 @@ struct InvariantTestData {
 /// Contains invariant test data.
 struct InvariantTest {
     // Fuzz state of invariant test.
-    fuzz_state: EvmFuzzState,
+    fuzz_state: InvariantFuzzState,
     // Contracts fuzzed by the invariant test.
     targeted_contracts: FuzzRunIdentifiedContracts,
     // Data collected during invariant runs.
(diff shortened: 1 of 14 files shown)


mablr (commented):
Benchmark CI run https://github.com/foundry-rs/foundry/actions/runs/25944187337

(this branch mirrored to upstream)

crates/evm/fuzz/src/inspector.rs
```diff
@@ -61,7 +66,9 @@ impl Fuzzer {
     /// Collects `stack` and `memory` values into the fuzz dictionary.
     #[cold]
     fn collect_data(&mut self, interpreter: &Interpreter) {
-        self.fuzz_state.collect_values(interpreter.stack.data().iter().copied().map(Into::into));
```
grandizzy, line 64: previously `collect_values` would use `insert_value` which would dedupe first and only after enforce the`max_fuzz_dictionary_values`, while now with `remaining` we truncate before dedupe. Can we just revert to the `FuzzDictionary::insert_value` to follow same logic?

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -1002,15 +1019,23 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
...
+                .collect::<Vec<_>>();
 
             let call_generator = RandomCallGenerator::new(
                 invariant_contract.address,
                 handler_addresses,
                 self.runner.clone(),
                 override_call_strat(
-                    fuzz_state.clone(),
-                    targeted_contracts.clone(),
+                    fuzz_state.snapshot(),
```
grandizzy, line 1037: if we use snapshots, then the calls generated for reentrancy checks (when `call_override = true`) won't benefit from values collected during runs (new values within the dictionary + dynamically created target contracts)

grandizzy (commented):
left couple of comments, ptal

PR #14787, opened 2026-05-15
fix(traces): scope function lookup by called address to fix selector collisions

figtracer:
## Bug

Two contracts sharing a function selector but with different return struct types caused trace decoding to use the wrong struct/field names.

```
// Wow.items returns WowStruct{a, b}; MuchWow.items returns MuchWowStruct{c, d}
├─ [5190] Wow::items(0) [staticcall]
│   └─ ← [Return] MuchWowStruct({ c: 11, d: 22 })   // wrong struct
```

## Cause

`self.functions` is a global `HashMap<Selector, Vec<Function>>`. On collisions, `select_contract_function` only disambiguates by trying to decode the **inputs**; identical input types → first registered candidate wins → its return ABI is used to format the output.

## Fix

Add a per-address `functions_by_address: HashMap<Address, HashMap<Selector, Function>>`, populated in `collect_abi`, and consult it first in `decode_function`. Falls back to the existing collision heuristic for unknown addresses.

After:
```
├─ [5190] Wow::items(0) [staticcall]
│   └─ ← [Return] WowStruct({ a: 11, b: 22 })
```

Includes a regression test.

### DIFF
--- a/crates/evm/traces/src/decoder/mod.rs
+++ b/crates/evm/traces/src/decoder/mod.rs
@@ -149,6 +149,9 @@ pub struct CallTraceDecoder {
 
     /// All known functions.
     pub functions: HashMap<Selector, Vec<Function>>,
+    /// Per-address function index, used to disambiguate selector collisions when the called
+    /// contract is identified. Keyed by address, then selector.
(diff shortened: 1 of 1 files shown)


figtracer:
superseded by #14788

PR #14788, opened 2026-05-15
fix(traces): prefer contract ABI for decoding

DaniPopes:
Trace decoding collected every known function ABI into a selector-wide cache, so calls with the same selector and input types could be decoded with ABI metadata from another contract. That made returned struct values render with the wrong struct and field names even though the underlying bytes were correct.

This keeps functions identified for a concrete address in an address-scoped cache and prefers that cache during trace decoding before falling back to global selector lookup. The regression uses Forge test source with two contracts that return different structs from the same selector and asserts the rendered trace uses each called contract ABI.

Reference: https://t.me/foundry_rs/42434


mattsse (approved):
lgtm

PR #14775, opened 2026-05-15
feat(cli): adopt NDJSON streaming for forge test

stevencartavia:
Stacked on #14774. Ships `forge.test` as the first streaming command in the agent contract.

- `forge.test` → `output_mode=stream`, `event_schema_ref=foundry:forge.test.event@v1`, `result_schema_ref=foundry:forge.test@v1`
- NDJSON events on stdout: `test_result`, `suite_finished`, `warning`
- Terminal envelope with `{ suites, passed, failed, skipped, duration_ms }`
- Failure path: error envelope `code=test.failed`, exit `5`

Adds `StreamRecord<T>` + `print_stream_record()` to `foundry_cli::json`. Under `--machine`, the global shell is forced to `Quiet` so existing human prints no-op; `print_json` bypasses quiet only in machine mode. Stream-incompatible flags (`--watch`, `--debug`, `--flamegraph`, `--flamechart`, `--gas-report`, `--summary`, `--list`, `--junit`, `--show-progress`) are rejected with `cli.usage.invalid` + exit `2`.

Part of OSS-218

PR #14791, opened 2026-05-15
feat(script): agent-substrate streaming for forge script

stevencartavia:
Stacked on #14775.

Adopts --machine for forge script:

- `output_mode=stream`, `event_schema_ref=foundry:forge.script.event@v1`, `result_schema_ref=foundry:forge.script@v1`, `side_effects=ChainWrite`, `long_running=true`.
- NDJSON phase events (`compile`, `simulate`, `broadcast`) and structured warnings on stdout; terminal envelope summarizes the on-chain effect (mode, tx hashes, created contracts).
- Broadcast failures emit a typed `script.broadcast_failed envelope`.

Part of OSS-218

### DIFF
--- a/crates/forge/tests/cli/script.rs
+++ b/crates/forge/tests/cli/script.rs
@@ -3927,7 +3927,8 @@ contract Demo {
     assert_eq!(envelope["success"], true);
     assert_eq!(envelope["data"]["mode"], "dry_run");
     assert_eq!(envelope["data"]["tx_count"], 0);
-    assert!(envelope["data"]["tx_hashes"].as_array().unwrap().is_empty());
+    assert!(envelope["data"]["transactions"].as_array().unwrap().is_empty());
+    assert!(envelope["data"].get("broadcast").is_none(), "broadcast must be dropped: {envelope}");
 
     // Warning duality: same `(code, message)` on the stream record and in `warnings[]`.
     let stream_warning =
@@ -3984,19 +3985,27 @@ contract DeployScript is Script {
     let envelope: Value =
         serde_json::from_str(stdout.lines().rfind(|l| !l.is_empty()).unwrap()).unwrap();
     assert_eq!(envelope["success"], false);
-    // `data: null` (key present + explicitly Null); `is_null()` alone also passes when absent.
     assert_eq!(envelope.get("data"), Some(&Value::Null), "{envelope}");
-    assert_eq!(envelope["errors"][0]["code"], "script.broadcast_failed");
-    assert!(
-        !envelope["errors"][0]["details"]["cause_chain"].as_array().unwrap().is_empty(),
-        "{envelope}"
-    );
+    let err = &envelope["errors"][0];
+    assert_eq!(err["code"], "script.broadcast_failed");
+    let details = &err["details"];
+    assert!(!details["cause_chain"].as_array().unwrap().is_empty(), "{envelope}");
+    // Typed fields agents can react to without parsing prose.
+    assert!(details["kind"].is_string(), "{envelope}");
+    assert!(details["retryable"].is_boolean(), "{envelope}");
+    // Recovery context: --resume is the v1 recovery primitive.
+    let recovery = &details["recovery"];
+    assert_eq!(recovery["resume_supported"], true, "{envelope}");
+    assert!(recovery["sequence_paths"].is_array(), "{envelope}");
+    assert!(recovery["submitted_count"].is_u64(), "{envelope}");
+    // Warning duality survives the failure path.
     let warnings = envelope["warnings"].as_array().unwrap();
     assert!(warnings.iter().any(|w| w["code"] == "script.noop_target"), "{envelope}");
 });
 
 // Reject flags whose stdout would corrupt the stream or whose runtime would
-// block on user/device input. One representative per class.
+// block on user/device input. Covers each rejected flag in the v1 list.
+// `--resume` is allowed under `--machine` (it's the documented recovery primitive).
 forgetest!(machine_mode_rejects_unsupported_flags, |prj, cmd| {
     let script = prj.add_source(
         "Foo",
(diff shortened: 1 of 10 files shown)


crates/script/src/lib.rs
```diff
@@ -623,6 +714,176 @@ impl ScriptArgs {
...
+    let mut envelope = foundry_cli::json::JsonEnvelope::error(
+        foundry_cli::json::JsonMessage::error(code, message)
+            .with_details(serde_json::json!({ "cause_chain": cause_chain })),
+    );
+    envelope.warnings = drain_machine_warnings();
+    let _ = foundry_cli::json::print_json(&envelope);
+    std::process::exit(foundry_cli::ExitCode::GenericError.to_i32());
+}
+
+fn bail_broadcast_failed(e: eyre::Report) -> ! {
```
grandizzy: 🔴 **Blocker (recoverability)**

The envelope emits `data: null`, even though `broadcast.rs` may already have checkpointed submitted hashes to `broadcast/<contract>/<chain_id>/run-latest.json`. Since `--resume` is also rejected under `--machine` (see other comment), an agent observing `script.broadcast_failed` cannot distinguish "nothing was submitted" from "some txs were submitted and need recovery." A fresh retry is not idempotent: depending on nonce/pending-chain state it can fail with nonce conflicts, re-simulate against mutated state, or duplicate side effects.

Please include recovery context in the failure envelope — at minimum submitted tx hashes and/or `sequence_path`, plus whether machine-mode resume is supported. Note this likely needs a small refactor: `bail_broadcast_failed` only receives `eyre::Report`, so either make `broadcast()` return a typed error carrying the sequence, or build the failure payload before bailing.

crates/script/src/lib.rs
```diff
@@ -322,27 +339,75 @@ impl ScriptArgs {
...
+        let len = |v: &Option<Vec<String>>| v.as_ref().map_or(0, Vec::len);
+        let keystore_count =
+            len(&self.wallets.keystore_paths) + len(&self.wallets.keystore_account_names);
+        let password_count =
+            len(&self.wallets.keystore_passwords).max(len(&self.wallets.keystore_password_files));
+        let keystore_unprotected = keystore_count > 0 && password_count < keystore_count;
+
+        let unsupported: Vec<&'static str> = [
+            ("--debug", self.debug),
+            ("--resume", self.resume),
```
grandizzy: 🔴 **Blocker (paired with `bail_broadcast_failed`)**

`--resume` is the documented recovery primitive for partial broadcast failures (existing `broadcast.rs` messages reference it explicitly). Rejecting it under `--machine` leaves no machine-readable recovery path for `script.broadcast_failed`. Either allow `--resume` under `--machine`, or include `recovery: { sequence_path, resume_supported }` in the failure envelope so agents can fall back deterministically.

crates/script/src/lib.rs
```diff
@@ -322,27 +339,75 @@ impl ScriptArgs {
...
+            ("--verify", self.verify),
```
grandizzy, line 393: 🟡 **UX gap**

`forge script --broadcast --verify` is a common one-shot deployment flow. If `--verify` remains unsupported in the MVP, please document the intended machine workflow for post-broadcast verification — e.g. parse contract addresses from the envelope and call a separate verify command. Not blocking the safety path, but worth resolving before this is advertised as agent-ready.

crates/script/src/lib.rs
```diff
@@ -623,6 +714,176 @@ impl ScriptArgs {
...
+    )
+}
+
+/// Bail with a typed code, attaching the eyre cause chain and draining recorded
+/// stream warnings into `warnings[]` so the terminal envelope stays authoritative.
+fn bail_machine_failure(code: &'static str, message: String, e: eyre::Report) -> ! {
+    let cause_chain: Vec<String> = e.chain().map(ToString::to_string).collect();
+    let mut envelope = foundry_cli::json::JsonEnvelope::error(
+        foundry_cli::json::JsonMessage::error(code, message)
+            .with_details(serde_json::json!({ "cause_chain": cause_chain })),
```
grandizzy, line 785: 🟡 **Should-fix (pre-`@v1` design)**

`cause_chain: [String]` is useful for humans but not enough for agent decisions. Since additive fields are allowed within `@v1`, please reserve/add typed fields now — e.g. `kind` (`rpc.timeout` | `nonce.too_low` | `signer.failed` | `unknown`), `submitted_count`, `retryable` — so agents don't have to parse error strings to decide whether `script.broadcast_failed` is safe to retry.

crates/script/src/lib.rs
```diff
@@ -623,6 +714,176 @@ impl ScriptArgs {
...
+fn drain_machine_warnings() -> Vec<foundry_cli::json::JsonMessage> {
+    std::mem::take(&mut *machine_warnings_lock())
+        .into_iter()
+        .map(|(code, message)| foundry_cli::json::JsonMessage::warning(code, message))
+        .collect()
+}
+
+/// Terminal `forge.script@v1` payload. Documented in `docs/agents/spec.md`.
+#[derive(Serialize)]
+struct ScriptResultData {
```
grandizzy, line 889: 🟡 **Spec-design (pre-`@v1` freeze)**

Three things worth settling before freezing `@v1`:
- `broadcast: bool` is redundant with `mode ∈ {"dry_run","broadcast"}`. Drop one.
- `tx_hashes` has no per-tx status. Consider `transactions: [{hash, status, block_number?}]`.
- `created_contracts` has no creating-tx pointer. Add `tx_hash` per entry so agents can correlate which tx created which contract.

crates/forge/tests/cli/script.rs
```diff
@@ -3892,3 +3892,255 @@ forgetest_async!(script_check_contract_sizes_honors_config_limit, |prj, cmd| {
...
+    let script = prj.add_source(
+        "Foo",
+        r#"
+contract Demo {
+    function run() external {}
+}
+   "#,
+    );
+
+    for flag in ["--debug", "--interactive", "--ledger"] {
```
grandizzy: 🟡 **Should-fix (coverage)**

Only 3 of 8 rejected flags exercised. Please add `--resume`, `--verify`, `--trezor`, `--browser`, and a keystore-without-password case. Also missing: `script.missing_rpc_url` branch, and the tempo `broadcast_batch` payload path if that's intended as a supported machine-mode surface.

grandizzy (changes requested):
One composite blocker around `script.broadcast_failed` recoverability that I'd want resolved before machine-mode `forge script` is advertised as agent-safe. Plus a few should-fixes and pre-`@v1` design questions.

_(Reviewed from a downstream agent-integration perspective — concerns are about how an autonomous agent would handle the failure surface, not about Foundry-internal correctness.)_

crates/script/src/lib.rs
stevencartavia: added recovery: { sequence_paths, submitted_count, resume_supported } to script.broadcast_failed details.

crates/script/src/lib.rs
stevencartavia: removed --resume from the rejection list.

crates/script/src/lib.rs
stevencartavia: spec.md documents the post-broadcast verify workflow (created_contracts[].address → forge verify-contract).

crates/script/src/lib.rs
stevencartavia: added typed kind (rpc.timeout | nonce.too_low | signer.failed | unknown) + retryable + submitted_count fields.

crates/script/src/lib.rs
stevencartavia: dropped broadcast, replaced tx_hashes with transactions: [{hash, status, block_number?}], added tx_hash to created_contracts.

crates/forge/tests/cli/script.rs
stevencartavia: rejected-flags matrix now covers all 7 (minus the now-allowed --resume), plus script.missing_rpc_url duality test and synthetic ScriptSequenceKind::Multi unit test for the batch payload path.

grandizzy (approved):
lgtm

crates/script/src/lib.rs
```diff
@@ -303,10 +303,27 @@ impl ScriptArgs {
...
-    #[allow(clippy::large_stack_frames)]
     pub async fn run_script(self) -> Result<()> {
         trace!(target: "script", "executing script command");
 
+        self.reject_machine_unsupported_flags()?;
+        let machine_mode = foundry_cli::is_machine();
+        if machine_mode {
+            // Avoid leaking warnings from a prior in-process invocation.
+            reset_machine_warnings();
+        }
```
mablr, line 314: Seems that `--machine` still leaks a tracing warning to stderr (`etherscan config not found`), so the machine stream is not clean.

crates/script/src/lib.rs
```diff
@@ -454,20 +535,31 @@ impl ScriptArgs {
...
-        let broadcasted = bundled.wait_for_pending().await?.broadcast().await?;
+        let bundled = bundled.wait_for_pending().await?;
+        emit_phase("broadcast")?;
+        // Snapshot recovery context before `.broadcast()` consumes `bundled`.
+        let recovery = machine_mode.then(|| RecoveryInfo::from_sequence(&bundled.sequence));
+        let broadcasted = match bundled.broadcast().await {
+            Ok(b) => b,
+            Err(e) if machine_mode => bail_broadcast_failed(e, recovery.unwrap()),
+            Err(e) => return Err(e),
+        };
```
mablr, line 554: This snapshots `RecoveryInfo` before `broadcast()` mutates and checkpoint-saves tx hashes. If a tx is submitted and a later receipt wait/RPC call fails, `details.recovery.submitted_count` can still be 0 even though `run-latest.json` contains submitted hashes. Updated recovery data should be carried out of the broadcast error path?

PR #14793, opened 2026-05-16
feat(lint): add return-bomb lint

0xKarl98:
## Motivation
Gas-capped external calls can still force the caller to copy unbounded dynamic returndata after the call returns, which can cause the caller to run out of gas.

## Solution
- Adds a low-severity `return-bomb` lint that flags gas-limited low-level calls and high-level external calls that consume dynamic return data. Includes documentation and lint test coverage for low-level calls, ignored returndata, overloads, and dynamic high-level returns.

cc @mattsse  @mablr @stevencartavia  @figtracer

### DIFF
--- a/crates/lint/src/sol/low/return_bomb.rs
+++ b/crates/lint/src/sol/low/return_bomb.rs
@@ -329,6 +329,10 @@ fn member_is_builtin_address(base: &hir::Expr<'_>, member: Symbol) -> bool {
 
 /// Returns whether an expression can match the expected type, if known.
 fn expr_matches_type(hir: &hir::Hir<'_>, expr: &hir::Expr<'_>, ty: &hir::Type<'_>) -> Option<bool> {
+    if let Some(matches) = bool_expr_matches_type(expr, ty) {
+        return Some(matches);
+    }
+
     if let Some(literal) = integer_literal(expr)
         && let Some(matches) = integer_literal_matches_type(literal, ty)
     {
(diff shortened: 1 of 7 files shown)


crates/lint/src/sol/low/return_bomb.rs
```diff
@@ -0,0 +1,329 @@
...
+        match expr_matches_type(hir, arg, &hir.variable(param).ty) {
+            Some(true) => {}
+            Some(false) => return ArgMatch::No,
+            None => maybe = true,
+        }
+    }
+    if maybe { ArgMatch::Maybe } else { ArgMatch::Exact }
+}
+/// Returns the contract id for expressions known to be contract-typed.
+fn expr_contract_id(hir: &hir::Hir<'_>, expr: &hir::Expr<'_>) -> Option<hir::ContractId> {
```
mattsse, line 245: This only recognizes call expressions when they are casts/constructor-style expressions, so it misses contract/interface values returned by ordinary function calls. For example, `factory.target().fetch{gas: gasLimit}()` still makes a gas-capped high-level external call that copies `bytes` returndata, but `expr_contract_id` returns `None` for `factory.target()`, so no lint is emitted.

Suggestion: resolve the contract id from the typed expression result for call expressions as well, then keep the explicit cast/constructor handling as the special case. I would also add a UI case for a returned interface value, since this is a common chaining shape.

crates/lint/src/sol/low/return_bomb.rs
```diff
@@ -0,0 +1,329 @@
...
+}
+
+/// Returns true if two types are equivalent for overload resolution.
+fn types_match(hir: &hir::Hir<'_>, a: &hir::Type<'_>, b: &hir::Type<'_>) -> bool {
+    match (&a.kind, &b.kind) {
+        (
+            TypeKind::Elementary(ElementaryType::Address(_)),
+            TypeKind::Elementary(ElementaryType::Address(_)),
+        ) => true,
+        (TypeKind::Elementary(a), TypeKind::Elementary(b)) => a == b,
```
mattsse: Overload matching currently requires elementary argument types to be exactly equal. That drops valid Solidity calls with implicit conversions before the return type is checked. A simple case is `function fetch(uint256) external returns (bytes memory)` called with a `uint8` value: `target.fetch{gas: gasLimit}(value)` compiles and still copies dynamic returndata, but this matcher rejects the candidate because `uint8 != uint256`.

Suggestion: either use Solar's resolved/type-checked call information if available, or model Solidity's implicit conversions for argument compatibility instead of strict ABI type equality. A focused test with integer widening would catch this.

mattsse (commented):
I left two inline comments for false negatives I could reproduce with scratch UI cases. The existing baseline checks passed for me (`cargo test -p forge-lint` and `cargo test -p forge --test ui -- ReturnBomb`).

crates/lint/src/sol/low/return_bomb.rs
0xKarl98: Yeah am finding thsoe missing false negatives this weekend

mattsse (approved):
Reviewed the latest head after the overload false-positive fix. The return-bomb UI regression now covers the bool overload case, and the targeted checks passed locally:\n\n- cargo test -p forge-lint\n- cargo test -p forge --test ui -- ReturnBomb

PR #14796, opened 2026-05-17
feat(symbolic): add native symbolic testing

mattsse:
## Summary

Adds native symbolic testing to Forge behind `forge test --symbolic`.

This introduces a Foundry-owned symbolic EVM executor, Z3-backed solving, symbolic ABI input generation, bounded stateless `check*` / `prove*` tests, bounded symbolic invariant sequences, Foundry config / CLI integration, and concrete replay before counterexamples are reported.

## Preview Semantics

This is an opt-in MVP preview, not a complete formal verification engine.

Symbolic results are scoped to the currently modeled EVM semantics and configured bounds:

- `PASS`: no feasible failure was found in the explored, modeled state space.
- `FAIL` with counterexample: the solver found a model and Forge replayed it concretely before reporting it.
- `FAIL: incomplete symbolic execution`: the run hit an unsupported feature, solver limit, timeout, unknown result, or non-replaying model.

Unsupported or unsafe-to-model behavior is intended to fail closed as `Incomplete` rather than silently passing.

## User Impact

Users can write ordinary Forge-style Solidity tests and run:

```sh
forge test --symbolic

### CODE REVIEW

File: crates/evm/symbolic/src/executor/opcodes.rs [hunk 1]
```diff
@@ -0,0 +1,825 @@
...
+                            Box::new(Expr::op(
+                                ExprOp::URem,
+                                Expr::op(ExprOp::Add, a.into_expr(), b.into_expr()),
+                                n,
+                            )),
+                        )))?;
+                    }
+                }
+                Ok(StepOutcome::Continue)
+            }
```
grandizzy, line 79: `ADDMOD` does not implement EVM semantics. EVM specifies that the intermediate sum is computed in unbounded precision and then reduced mod `n`. The concrete branch uses `a.wrapping_add(b) % n` and the symbolic branch builds `(a + b) % n` over 256-bit `Expr::Add`, both of which wrap before the modulo.

False-`PASS` repro:
```solidity
function check() external pure {
    // Real EVM: addmod(2**256 - 1, 1, 3) == 1
    // Engine:   wrapping_add => 0, 0 % 3 == 0
    assert(addmod(type(uint256).max, 1, 3) == 1);
}
```
This violates the README's fail-closed `PASS` semantics.

Fix: for concrete operands, widen to `U512` (or `ruint::Uint<512, 8>`) before the modulo. For symbolic operands either emit an exact widened SMT encoding or return `SymbolicError::Unsupported("symbolic ADDMOD")` so the run reports `Incomplete`. Please add a regression that exercises `addmod(type(uint256).max, 1, 3)`.

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -0,0 +1,825 @@
...
+                            Box::new(BoolExpr::eq(n.clone(), Expr::Const(U256::ZERO))),
+                            Box::new(Expr::Const(U256::ZERO)),
+                            Box::new(Expr::op(
+                                ExprOp::URem,
+                                Expr::op(ExprOp::Mul, a.into_expr(), b.into_expr()),
+                                n,
+                            )),
+                        )))?;
+                    }
+                }
```
grandizzy, line 93: Same issue as `ADDMOD` above: `wrapping_mul(b) % n` (concrete) and 256-bit `Expr::Mul` then `URem` (symbolic) do not match EVM `MULMOD`, which uses unbounded intermediate multiplication.

False-`PASS` repro:
```solidity
function check() external pure {
    // Real EVM: mulmod(1 << 255, 2, 3) == 1 (since 2**256 mod 3 == 1)
    // Engine:   wrapping_mul => 0, 0 % 3 == 0
    assert(mulmod(1 << 255, 2, 3) == 1);
}
```
Fix: widen concrete to `U512` for the multiply; for symbolic operands, exact widened SMT or fail closed with `SymbolicError::Unsupported("symbolic MULMOD")`. Add a regression for `mulmod(1 << 255, 2, 3)`.

crates/evm/symbolic/src/runtime/precompiles.rs
```diff
@@ -0,0 +1,199 @@
...
+pub(crate) fn precompile_address(number: u8) -> Address {
+    let mut bytes = [0u8; 20];
+    bytes[PRECOMPILE_ADDRESS_LEADING_ZEROS] = number;
+    Address::from(bytes)
+}
+
+/// Returns whether `is_supported_precompile` holds.
+pub(crate) fn is_supported_precompile(address: Address) -> bool {
+    precompile_number(address).is_some()
+}
```
grandizzy, line 35: `precompile_number` hardcodes `1..=9` and `is_supported_precompile` derives from it, so any call to a spec-active precompile that isn't in this set — today notably the EIP-4844 KZG point-evaluation precompile at `0x0a` on Cancun — is classified as "not a precompile". Calls then fall through `executor/calls.rs` to the empty-`child_code` branch, which pushes `U256::from(1)` and continues, silently reporting `Safe`.

This is a false-`PASS` path on any contract that calls `0x0a`.

Suggested fixes (any of):
1. Use revm's spec-aware precompile set for the active fork; for any address that revm considers a precompile but this module doesn't model, return `SymbolicError::Unsupported(format!("unsupported precompile {address}"))` from `execute_precompile`.
2. As a minimum, add `0x0a` to the registry and immediately return `Unsupported` from `execute_precompile` until a model lands.

Also please add a CLI regression that does a `staticcall` to `0x0a` and asserts the run is reported `Incomplete`, not `PASS`.

crates/evm/symbolic/src/executor/cheatcodes.rs
```diff
@@ -0,0 +1,2533 @@
...
+            ]));
+        }
+        if selector == selector!("snapshotGasLastCall(string)")
+            || selector == selector!("snapshotGasLastCall(string,string)")
+            || selector == selector!("stopSnapshotGas()")
+            || selector == selector!("stopSnapshotGas(string)")
+            || selector == selector!("stopSnapshotGas(string,string)")
+        {
+            return Ok(CheatcodeOutcome::Continue(vec![SymWord::zero()]));
+        }
```
grandizzy: Several cheatcodes in this no-op / fabricated-return table are observable to user tests, so silently no-op'ing them can turn a real concrete failure into a symbolic `Safe`:

- `expectSafeMemory(uint64,uint64)` / `expectSafeMemoryCall(uint64,uint64)` / `stopExpectSafeMemory()` — assertion-shaped; concrete Forge fails the test when violated. Currently silently accepted.
- `lastCallGas()` returns five zero words; any test asserting on returned gas fields will pass symbolically and may fail concretely.
- `snapshotGasLastCall*` getters whose return value is consumed in an assertion.

Suggest splitting this dispatcher into two tables:
1. "safe no-op" — cheatcodes with no observable effect on pass/fail (e.g., pure metering/debug helpers whose results aren't read).
2. "unsupported" — return `SymbolicError::Unsupported(...)` until properly modeled. All of the above belong here for now.

Please add CLI regressions that read each of these values inside an assertion and verify the run reports `Incomplete`.

crates/evm/symbolic/src/runtime/expressions.rs
```diff
@@ -0,0 +1,1286 @@
...
+                let mut conditions = vec![Self::eq((**left_len).clone(), (**right_len).clone())];
+                conditions.extend(
+                    left_bytes
+                        .iter()
+                        .cloned()
+                        .zip(right_bytes.iter().cloned())
+                        .map(|(left, right)| Self::eq(left, right)),
+                );
+                Self::and(conditions)
+            }
```
grandizzy, line 1145: This rewrite makes `keccak(a) == keccak(b)` provable from `a == b` (and, by the recursive structure, the converse too via the byte-equality decomposition). That is the standard symbolic-execution "no Keccak collisions" assumption, but the README's `Preview Result Semantics` currently presents `PASS` as scoped only to modeled semantics and configured bounds — not as additionally assuming hash collision/preimage resistance.

Either:
1. Add an explicit paragraph in the README result-semantics and known-gaps sections: "`PASS` additionally assumes collision and preimage resistance of `KECCAK256` and other hash precompiles." Same call-out in the result FAQ.
2. Or drop the converse direction of the rewrite and fail closed (`Incomplete`) when a proof obligation reduces to a Keccak-injectivity fact.

Option 1 is fine for preview, but it needs to be loud.

grandizzy (changes requested):
Pre-preview soundness audit found three concrete false-`PASS` paths that I think should block any public preview/MVP framing, plus one undocumented cryptographic assumption. Engineering scaffolding is good (unsupported-opcode → `Incomplete`, solver `unknown` → `Timeout`, model validation, concrete replay, world overlay, symbolic memory bounds all reviewed cleanly) — the issues below are localized and fixable.

Blockers:
1. `ADDMOD` / `MULMOD` use wrapping intermediate arithmetic instead of widened (see `executor/opcodes.rs`).
2. Cancun precompile `0x0a` (KZG point-evaluation) is not in the precompile registry and falls through to empty-account success (`runtime/precompiles.rs` + `executor/calls.rs`).
3. Observable cheatcodes `expectSafeMemory*`, `lastCallGas`, `snapshotGasLastCall*` are silently no-op / fabricated (`executor/cheatcodes.rs`) — can convert real concrete failures into symbolic `Safe`.

Documentation gap:
4. `BoolExpr::eq` rewrites `keccak(a) == keccak(b)` to byte-equality (`runtime/expressions.rs`), encoding collision resistance as a hard fact — fine as an assumption but the README currently presents `PASS` as fail-closed. Either document or remove the converse direction.

Would also like to see a small `false_pass_prevention` CLI suite under `crates/forge/tests/cli/test_cmd/` covering each of the above so we don't regress. After these land I'm comfortable shipping `--symbolic` as an opt-in preview.

grandizzy (dismissed):
approving with the mention this is MVP and should be treated as such

mablr (dismissed):
MVP good to go imo 👍

grandizzy (dismissed):

mablr (approved):

grandizzy (approved):

PR #14798, opened 2026-05-17
fix(cast): support tempo sponsors in erc20 commands

mattsse:
## Summary
Adds Tempo sponsor handling to state-changing `cast erc20` helpers so their Tempo transaction behavior matches `cast send`.

## Details
`cast erc20` already accepts the shared Tempo transaction options, but the ERC20 send helper was not resolving sponsor configuration, printing sponsor hashes, or attaching fee-payer signatures before submission.

This change wires `sponsor_config()` into the ERC20 send macro, supports `--tempo.print-sponsor-hash`, and attaches sponsor signatures across access-key, browser, and local-signer paths. Access-key sponsor handling prepares key authorization before the sponsor digest is computed, matching the `cast send` ordering.


figtracer (approved):

mablr (approved):
lgtm

PR #14802, opened 2026-05-18
chore(workflows): remove in-repo developer convenience Nix flake

decofe:
## Summary
- remove the in-repo Nix flake and lockfile
- remove the scheduled/manual Nix workflow that updates and smoke-tests the flake

## Rationale
The current flake only provides a contributor dev shell; it does not define a Foundry package output. Foundry is already packaged in nixpkgs, so the distribution path for Nix users can live with the Nix maintainers instead of being maintained in this repository.

Removing these files drops an in-repo convenience surface while leaving the Rust/Cargo build and release paths untouched.

## Testing
- Not run; deletion-only change removing the Nix workflow and flake files.


zerosnacks:
Hi @beeb / @shaunkh, after some consideration in an ongoing effort to reduce attack / maintenance surface area relating to (writable) Github workflows I would like to propose we remove in-repo support for the maintenance of the Nix flake. To my understanding it only serves as a dev-environment convenience and is unrelated to distribution.

grandizzy (approved):

beeb (approved):

PR #14805, opened 2026-05-18
feat(ci): benchmarks for stable vs PR/branch

mablr:
## Motivation

Update `benchmarks.yml` to run by default the current branch against stable Foundry.
- Posts results as a PR comment when a corresponding open PR exists
- Always prints to stdout for visibility in workflow logs. 
- Results publication as PR is skipped by default, done only when current branch is not in the versions to bench.


zerosnacks (changes requested):
as discussed we are looking to have a single benchmarks.yml that is more configurable

mablr (commented):
@zerosnacks Merged to `benchmarks.yml`, updated the version input defaulting for convenience.

mablr:
If this manual run w/ aave succeeds, it will be mergeable: https://github.com/foundry-rs/foundry/actions/runs/26155089151/job/76932244073

mablr:
Also one with default repos set https://github.com/foundry-rs/foundry/actions/runs/26157768166

Updated Solady excluded tests (which fail), prob because fuzzing was improved recently.

zerosnacks (approved):
Nice! Lgtm

PR #14806, opened 2026-05-18
perf(fuzz): remove locks from invariant dictionaries

decofe:
Supersedes #14786, rebased on latest `master` because the original cross-fork PR could not be reopened or updated by maintainers.

## Motivation

Invariant fuzzing currently pays lock overhead on hot dictionary and target-contract paths even though each invariant campaign runs locally within a single test invocation.

## Solution

- Make invariant fuzz state worker-local and lock-free.
- Replace shared dictionary locking with local mutable state for invariant campaigns.
- Remove target-contract locking from invariant execution paths.
- Keep inspector-collected stack words staged locally, bounded by the configured dictionary value cap, and drain them after each executed tx, including discarded `vm.assume` runs.
- Keep stateless fuzz using forked per-worker dictionary state.

Aave invariant benchmark, 3 reps, 120s timeout:

- baseline `5454d35d8`: 37,967.8 elapsed tx/s mean
- this branch: 57,637.7 elapsed tx/s mean
- improvement: +51.8% elapsed tx/s


crates/evm/fuzz/src/strategies/state.rs
```diff
@@ -69,27 +74,67 @@ impl EvmFuzzState {
...
+
+    pub fn into_invariant(self) -> InvariantFuzzState {
+        InvariantFuzzState {
+            inner: Rc::new(RefCell::new((*self.inner).clone())),
+            deployed_libs: self.deployed_libs,
+        }
+    }
+
+    pub fn fork(&self) -> Self {
+        Self { inner: Arc::new((*self.inner).clone()), deployed_libs: self.deployed_libs.clone() }
```
grandizzy: this could be `Arc::clone(&self.inner)` since the stateless `EvmFuzzState` is never mutated after construction

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -609,14 +610,28 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
...
-                    if !call_result.reverted {
-                        collect_data(
-                            &invariant_test,
-                            &mut state_changeset,
-                            current_run.inputs.last().expect("checked above"),
-                            &call_result,
-                            self.config.depth,
-                        );
+                    if let Some(fuzzer) = current_run.executor.inspector_mut().fuzzer.as_mut() {
+                        if !call_result.reverted {
```
grandizzy: duplicated if let Some(fuzzer) ... else ... can be collapsed to a single `if !reverted` because the fuzzer is always installed by `prepare_test()`
```rust
if !call_result.reverted {
    let mapping_slots = current_run.executor.inspector().fuzzer.as_ref()
        .and_then(|f| f.mapping_slots.as_ref());
    collect_data(&invariant_test, &mut state_changeset, ..., mapping_slots);
}
```

crates/evm/fuzz/src/inspector.rs
```diff
@@ -61,7 +66,9 @@ impl Fuzzer {
     /// Collects `stack` and `memory` values into the fuzz dictionary.
     #[cold]
     fn collect_data(&mut self, interpreter: &Interpreter) {
-        self.fuzz_state.collect_values(interpreter.stack.data().iter().copied().map(Into::into));
+        let remaining = self.max_collected_values.saturating_sub(self.collected_values.len());
```
grandizzy, line 69: nit: the cap now applies to raw staged words per tx (pre-dedup), where previously dedup happened in the dictionary (in `collect_values` through `insert_value`). Early repeated values can exhaust the budget and starve deep-stack unique values, impact is small because the default cap is 9.83M words, which a single tx will essentially never fill

grandizzy (commented):
nice! left couple of comments, pls check

grandizzy (dismissed):

grandizzy:
@0xalpharush ptal https://github.com/foundry-rs/foundry/pull/14806/changes/21c7c7b208fcc79c8bf05cdef2a1e76baee943a2 should be gtg from my side

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -565,6 +563,9 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
                     &mut current_run.executor,
                     current_run.inputs.last().expect("checked above"),
                 )?;
+                if let Some(fuzzer) = current_run.executor.inspector_mut().fuzzer.as_mut() {
+                    invariant_test.fuzz_state.collect_values(fuzzer.drain_collected_values());
```
figtracer, line 567: are we draining properly on every stage? e.g.: preflight should probably call this as well

grandizzy: pushed https://github.com/foundry-rs/foundry/pull/14806/changes/e0978079e9325444a47449f3b29ecef8e818cc7f to address

0xalpharush (approved):

crates/evm/evm/src/executors/invariant/mod.rs
zerosnacks: nice find 👍

grandizzy (approved):

zerosnacks (approved):

mablr (commented):
Benchmarks run: https://github.com/foundry-rs/foundry/actions/runs/26165463989

figtracer (approved):

PR #14813, opened 2026-05-18
feat(chisel): Chisel dispatch by EVM network

### LINKED ISSUE #14812: Make Chisel execution dispatch through FoundryEvmNetwork instead of EthEvmNetwork
### Component

Chisel

Chisel currently builds and stores an Ethereum-only executor internally. Even when users pass a non-Ethereum network via the shared Foundry network config, Chisel still constructs `Executor<EthEvmNetwork>` and resolves the execution spec through the Ethereum path.

This means Chisel does not actually execute with the configured network's EVM factory, hardfork type, transaction environment, precompiles, or gas schedule. For chains supported through Foundry's v1.7 generic EVM architecture, `chisel --network <network> eval ...` can silently produce Ethereum execution behavior instead of the selected network behavior.

A concrete symptom is that network-specific cold access gas pricing is ignored in Chisel because the REPL execution still uses the Ethereum executor:

- `BALANCE`/`EXTCODESIZE` use Ethereum cold account access cost
- `SLOAD` uses Ethereum cold storage access cost
- warm access may appear correct only because warm access is shared across networks

Chisel should follow the same pattern as the other Foundry commands that have been moved to the generic EVM architecture:

- make the Chisel runner/execution path generic over `FEN: FoundryEvmNetwork`
- build `Executor<FEN>` / `Backend<FEN>` / `ExecutorBuilder<FEN>` through that generic path
- resolve the spec as `SpecFor<FEN>` instead of always using Ethereum `SpecId`
- dispatch at the command boundary based on `NetworkConfigs` to `EthEvmNetwork`, `OpEvmNetwork`, `TempoEvmNetwork`, or other configured network implementations

This keeps Chisel upstream-compatible and avoids adding chain-specific behavior in the REPL itself. Network-specific behavior should come from the selected `FoundryEvmNetwork` implementation, not from hardcoded conditionals inside Chisel.

### Additional context

_No response_

haythemsellami:
Closes #14812.

This makes Chisel's session/source/runner execution path generic over FoundryEvmNetwork and dispatches from the loaded NetworkConfigs at the command boundary.

It also adds a Tempo regression test that exercises chisel --network tempo against a Tempo-only precompile.


crates/chisel/src/source.rs
stevencartavia: this line 649 now needs to be `let mut config: SessionSourceConfig = SessionSourceConfig {`. the test is broken

crates/chisel/src/dispatcher.rs
```diff
@@ -292,8 +293,8 @@ impl ChiselDispatcher {
         }
 
         let new_session = match id {
-            "latest" => ChiselSession::latest(),
-            id => ChiselSession::load(id),
+            "latest" => ChiselSession::<FEN>::latest(),
+            id => ChiselSession::<FEN>::load(id),
```
figtracer, line 297: a Tempo session saved with `--network tempo` can later be loaded as `ChiselSession<EthEvmNetwork>` if the user runs `chisel load <id>` without `--network tempo`

haythemsellami: Good catch. I added a load-time network check before rebuilding the saved session. Chisel now compares
the current config network against the network persisted in the saved session config and errors if they
differ.

crates/chisel/src/dispatcher.rs
```diff
@@ -48,8 +49,8 @@ pub const CHISEL_CHAR: &str = "⚒️";
 
 /// Chisel input dispatcher
 #[derive(Debug)]
-pub struct ChiselDispatcher {
-    pub session: ChiselSession,
+pub struct ChiselDispatcher<FEN: FoundryEvmNetwork = EthEvmNetwork> {
```
mablr: I think this shouldn't default to `EthEvmNetwork`, dispatch must be explicit to avoid misrouting.

stevencartavia (approved):

stevencartavia:
Added `with_chain_id` and `infer_network_from_fork` normalization (and mirrored the result back into `config.networks`) before executor dispatch, so `--chain` and `--rpc-url` now select the correct executor family instead of silently falling through to `EthEvmNetwork`

mattsse (approved):
LGTM

PR #14814, opened 2026-05-18
feat(symbolic): add named dynamic ABI bounds

figtracer:
## Summary

Adds Halmos-style dynamic ABI length configuration on top of the native symbolic testing branch:

- `symbolic.dynamic_lengths`: named dynamic bounds keyed by ABI argument name or generated symbolic name (`calldata_0`, nested aliases, etc.)
- `symbolic.default_array_lengths`: type-specific defaults for dynamic arrays
- `symbolic.default_bytes_lengths`: type-specific defaults for `bytes` and `string`
- `@custom:halmos` translation for named `--array-lengths` plus `--default-array-lengths` / `--default-bytes-lengths`

Length resolution is backward-compatible: named bounds first, then type defaults, then legacy positional `array_lengths`, then `default_dynamic_length`.

Length sets now expand into separate symbolic calldata shapes. Nested dynamic values explore the cross-product implied by the selected outer lengths, bringing this closer to Halmos-style dynamic ABI bounds while keeping the existing `path_width` / calldata-size limits as the main controls for combinatorial growth.


figtracer:
@grandizzy can you ptal? think this can be included now as well

grandizzy:
> @grandizzy can you ptal? think this can be included now as well

@figtracer can we split the lib.rs in modules (can be abi.rs for example for this addition, but we could have calldata.rs, cheatcode.rs etc.) it's quite hard to review the 14.5k LoC file 😬 thanks!

crates/evm/symbolic/src/abi.rs
```diff
@@ -97,38 +125,45 @@ pub(super) struct SymbolicInput {
...
     /// Constructs a new instance.
     pub(super) const fn new(config: &'a SymbolicConfig) -> Self {
-        Self { config, constraints: Vec::new(), dynamic_index: 0 }
+        Self { config, constraints: Vec::new(), positional_dynamic_index: 0 }
     }
 
     /// Validates the `finish` symbolic ABI helper.
     pub(super) fn finish(&self) -> Result<(), SymbolicError> {
-        if self.dynamic_index != self.config.array_lengths.len() {
+        if self.positional_dynamic_index != self.config.array_lengths.len() {
```
grandizzy: ### Correctness: `finish()` is too strict per variant

This check requires `positional_dynamic_index == array_lengths.len()` for **every** expanded variant. With the new length-set expansion, variants legitimately consume different numbers of positional leaves, so any mix of outer length-sets + positional `array_lengths` will reject at least one variant and fail the whole build.

Repro:

```solidity
contract BatchSymbolicTest is Test {
    MyContract c = new MyContract();

    /// forge-config: default.symbolic.default_array_lengths = [1, 2]
    /// forge-config: default.symbolic.array_lengths = [4, 4]
    function check_processBatch(bytes[] calldata items) external view {
        // ...
    }
}
```

| Variant | `items.length` | positional consumed | `finish()` |
|---|---|---|---|
| A | 1 | 1 | ❌ `1 != 2` |
| B | 2 | 2 | ✅ |

The user has no `array_lengths` value that satisfies both — `[4]` fails B, `[4,4]` fails A. Error:

```
symbolic.array_lengths has 2 entries but ABI used 1 positional dynamic leaves
```

This also blocks straightforward Halmos migrations that combine `--default-array-lengths` with positional `--array-lengths`.

**Suggested fix:** drop the per-variant `finish()` and validate once after expansion that `array_lengths.len() <= max(positional_dynamic_index)` across variants. Preserves the stale-entry guarantee for single-variant configs; lets shorter variants coexist with longer ones.

crates/evm/symbolic/src/abi.rs
```diff
@@ -31,34 +40,53 @@ impl SymbolicCalldata {
...
-            .copied()
-            .map(|byte| SymWord::Concrete(U256::from(byte)))
-            .collect::<Vec<_>>();
-        bytes.extend(encode_sequence(inputs.iter().map(|input| &input.value)));
-        if bytes.len() > config.max_calldata_bytes as usize {
-            return Err(SymbolicError::Unsupported(
-                "symbolic calldata size exceeds configured max",
-            ));
-        }
+        variants
```
grandizzy, line 76: ### Performance: eager cross-product materialization

`variants_with_prefix` fully materializes a `Vec<SymbolicCalldata>` before any runtime cap can apply. With nested length-sets the variant count grows as the product of all dynamic-length set sizes — a small config like `bytes[][]` with `default_array_lengths=[1,2,3]` and `default_bytes_lengths=[0,65,1024]` builds 3 outer × (1+2+3) inner-counts × 3^inner bytes shapes before any check fires.

- `max_paths` is enforced only after the worklist is built (`executor.rs:104`, `executor.rs:283-288`).
- `max_calldata_bytes` is checked per-variant, but only *after* the variant is fully encoded.

A user can OOM during calldata construction with no guardrail.

**Suggested fix:** add an explicit variant cap (e.g. `max_paths` or a new `max_calldata_variants`) at the top of the expansion loop, short-circuiting before the cross-product blows up. Lazy/iterator-based generation would be ideal but is a larger refactor — a cap is sufficient as a first guardrail.

grandizzy:
### Correctness: filtered compile path misses symbolic tests

(Note: posting as a general comment because the line is outside the PR's diff hunks.)

`MultiContractRunner::matches_artifact` at `crates/forge/src/multi_runner.rs:103` was updated to recognize `is_symbolic_entrypoint`, but `crates/forge/src/cmd/test/mod.rs:341` still uses the free `matches_artifact(test_filter, id, abi)` helper at `multi_runner.rs:657`, which only checks classic `func.is_any_test()`:

```rust
// cmd/test/mod.rs:341
id.source.starts_with(&config.src) || matches_artifact(test_filter, id, abi)
```

Effect: running `forge test --symbolic --match-test check_foo` may drop the source file from filtered compilation entirely, so the test never reaches the runner — even though the runtime discovery (`matching_contracts` / `all_test_functions`) would have picked it up.

**Suggested fix:** centralize one symbolic-aware artifact matcher and use it in both places, or thread `symbolic.enabled` into the free function so the two paths can't drift.

grandizzy (approved):

PR #14824, opened 2026-05-19
fix(cast): consistent serialization of Uint/Ints depending on actual type

0xferrous:
## Motivation

The current implementation dynamically tries to determine if the runtime value can fit in 64 bits, but this leads to inconsistent serialization. For instance if you were decoding an `uint[]`, some of the values that fit in 64 bits will serialize as number while others serialize as string making it require special handling on the user that is consuming the json.

## Solution

Use the type information to determine the serialization. So the user will always know that specific types will always serialize to a number or a string depending on the number of bits that type uses.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes


crates/common/fmt/Cargo.toml
```diff
@@ -31,6 +31,7 @@ chrono.workspace = true
 revm.workspace = true
 yansi.workspace = true
 comfy-table.workspace = true
+proptest.workspace = true
```
zerosnacks: let's move this to [dev-dependencies]

crates/common/fmt/src/dynamic.rs
```diff
@@ -153,15 +153,20 @@ pub fn format_token_raw(value: &DynSolValue) -> String {
 pub fn serialize_value_as_json(
     value: DynSolValue,
     defs: Option<&StructDefinitions>,
+    strict: bool,
```
zerosnacks, line 160: lets add a doc comment here for what strict means

crates/common/fmt/src/dynamic.rs
```diff
@@ -318,4 +330,76 @@ mod tests {
             "0xFb6916095cA1Df60bb79ce92cE3EA74c37c5d359"
         );
     }
+
+    proptest::proptest! {
```
zerosnacks, line 357: Could you additionally add a non-proptest unit test that serializes a uint256[] containing both a small (1) and a large (> u64::MAX) value with strict: true and asserts both come out as strings? That directly encodes the bug described in the PR motivation as a regression test.

```
#[test]
fn strict_uint256_array_is_homogeneous() {
    let small = U256::from(1u64);
    let big = U256::from(1u64) << 200; // doesn't fit in u64

    let arr = DynSolValue::Array(vec![
        DynSolValue::Uint(small, 256),
        DynSolValue::Uint(big,   256),
    ]);

    let json = serialize_value_as_json(arr, None, true).unwrap();

    // Both elements must be strings because the *type* is uint256 (>64 bits),
    // regardless of whether the runtime value happens to fit in u64.
    assert_eq!(
        json,
        serde_json::json!(["1", "1606938044258990275541962092341162602522202993782792835301376"])
    );
}
```

zerosnacks (commented):
Hi @0xferrous, thanks for the PR! Apologies the other one went stale.

Some notes, overall lgtm 👍

0xferrous:
Still relevant.

figtracer (approved):
lgtm

stevencartavia (approved):

mablr (approved):

PR #14825, opened 2026-05-19
refactor(symbolic): split executor into modules

figtracer:
## Summary
- split the native symbolic executor out of the 14k-line lib.rs into focused modules


figtracer:
@grandizzy this level of granularity should be better, i'll rebase the dynamic ABI PR after wdyt

grandizzy:
> @grandizzy this level of granularity should be better, i'll rebase the dynamic ABI PR after wdyt

@figtracer looks good! let's quickly add one liner description for each fn (like address.rs has none, OK to follow-up if you think is too much in PR scope) and good to merge.

grandizzy (approved):
nice!

PR #14827, opened 2026-05-19
feat: Add Tempo session registry scaffold

0xKarl98:
On the first step of OSS-162 

### Summary 
This PR introduces the initial Tempo session registry layer, separate from the long-lived access key store.

  - adds `SessionEntry`, `SessionRecord`, and `SessionStatus`
  - persists session metadata under `~/.tempo/wallet/sessions.toml`
  - supports atomic upsert, removal, and expiry tracking
  - extracts shared TOML read/write helpers for Tempo registries
  - keeps the existing `keys.toml` access key behavior unchanged
  
cc @mattsse  @mablr

### DIFF
--- a/crates/common/src/tempo/keystore.rs
+++ b/crates/common/src/tempo/keystore.rs
@@ -6,7 +6,9 @@
 use alloy_primitives::{Address, hex};
 use alloy_rlp::Decodable;
 use serde::{Deserialize, Serialize};
-use std::{env, fs, io::Write, path::PathBuf};
+use std::{env, path::PathBuf};
+
+use super::registry::{read_toml_file, write_toml_file_atomic};
 
 /// Environment variable for an ephemeral Tempo private key.
 pub const TEMPO_PRIVATE_KEY_ENV: &str = "TEMPO_PRIVATE_KEY";
@@ -127,26 +129,7 @@ pub fn tempo_keys_path() -> Option<PathBuf> {
 /// Errors are logged as warnings.
 pub fn read_tempo_keys_file() -> Option<KeysFile> {
     let keys_path = tempo_keys_path()?;
-    if !keys_path.exists() {
-        tracing::trace!(?keys_path, "tempo keys file not found");
-        return None;
-    }
-
-    let contents = match fs::read_to_string(&keys_path) {
-        Ok(c) => c,
-        Err(e) => {
-            tracing::warn!(?keys_path, %e, "failed to read tempo keys file");
-            return None;
-        }
-    };
-
-    match toml::from_str(&contents) {
-        Ok(f) => Some(f),
-        Err(e) => {
-            tracing::warn!(?keys_path, %e, "failed to parse tempo keys file");
-            None
-        }
-    }
+    read_toml_file(&keys_path, "tempo keys")
 }
 
 /// Decodes a hex-encoded, RLP-encoded key authorization.
@@ -168,25 +151,16 @@ pub fn decode_key_authorization<T: Decodable>(hex_str: &str) -> eyre::Result<T> 
 /// temp file + rename so a crash mid-write cannot corrupt the file.
 pub(crate) fn upsert_key_entry(entry: KeyEntry) -> eyre::Result<()> {
     let path = tempo_keys_path().ok_or_else(|| eyre::eyre!("could not resolve tempo home"))?;
-    let dir = path.parent().ok_or_else(|| eyre::eyre!("invalid keys path: {}", path.display()))?;
-    fs::create_dir_all(dir)?;
-
     let mut file = read_tempo_keys_file().unwrap_or_default();
     file.keys
         .retain(|k| !(k.wallet_address == entry.wallet_address && k.chain_id == entry.chain_id));
     file.keys.push(entry);
 
-    let body = toml::to_string_pretty(&file)?;
-    let contents = format!(
-        "# Tempo wallet keys — managed by Foundry / Tempo CLI.\n# Do not edit manually.\n\n{body}"
-    );
-
-    let mut tmp = tempfile::NamedTempFile::new_in(dir)?;
-    tmp.write_all(contents.as_bytes())?;
-    tmp.flush()?;
-    tmp.persist(&path).map_err(|e| eyre::eyre!("failed to persist keys.toml: {e}"))?;
-
-    Ok(())
+    write_toml_file_atomic(
+        &path,
+        &file,
+        "# Tempo wallet keys — managed by Foundry / Tempo CLI.\n# Do not edit manually.",
+    )
 }
 
 #[cfg(test)]
--- a/crates/common/src/tempo/keystore.rs
+++ b/crates/common/src/tempo/keystore.rs
@@ -98,16 +98,6 @@ pub struct KeysFile {
     pub keys: Vec<KeyEntry>,
 }
 
-/// Process-wide mutex used by tests that mutate `TEMPO_HOME`.
-///
-/// Returns a [`tokio::sync::Mutex`] so async tests can hold it across `.await`
-/// points without tripping `clippy::await_holding_lock`.
-#[cfg(test)]
-pub(crate) fn test_env_mutex() -> &'static tokio::sync::Mutex<()> {
-    static M: std::sync::OnceLock<tokio::sync::Mutex<()>> = std::sync::OnceLock::new();
-    M.get_or_init(|| tokio::sync::Mutex::new(()))
-}
-
 /// Resolve the Tempo home directory.
 ///
 /// Uses `TEMPO_HOME` env var if set, otherwise `~/.tempo`.
@@ -129,7 +119,13 @@ pub fn tempo_keys_path() -> Option<PathBuf> {
 /// Errors are logged as warnings.
 pub fn read_tempo_keys_file() -> Option<KeysFile> {
     let keys_path = tempo_keys_path()?;
-    read_toml_file(&keys_path, "tempo keys")
+    match read_toml_file(&keys_path, "tempo keys") {
+        Ok(value) => value,
+        Err(e) => {
+            tracing::warn!(?keys_path, %e, "failed to load tempo keys file");
+            None
+        }
+    }
 }
 
 /// Decodes a hex-encoded, RLP-encoded key authorization.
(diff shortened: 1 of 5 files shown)


mattsse (dismissed):
LGTM. I reviewed the registry scaffold and keystore TOML helper refactor and found no blocking issues.

figtracer (dismissed):
makes swnse

stevencartavia (dismissed):

zerosnacks (dismissed):

crates/common/src/tempo/session.rs
```diff
@@ -0,0 +1,270 @@
...
+/// Top-level registry persisted in `wallet/sessions.toml`.
+#[derive(Debug, Default, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub struct SessionRecord {
+    #[serde(default)]
+    pub sessions: Vec<SessionEntry>,
+}
+
+impl SessionRecord {
+    /// Returns `true` if the registry has no session entries.
+    pub fn is_empty(&self) -> bool {
```
mablr:
```suggestion
    pub const fn is_empty(&self) -> bool {
```
Clippy says this should be a const.

mablr (changes requested):

mablr (dismissed):

crates/common/src/tempo/registry.rs
```diff
@@ -0,0 +1,55 @@
+use eyre::Result;
+use serde::{Serialize, de::DeserializeOwned};
+use std::{fs, io::Write, path::Path};
+
+/// Shared TOML registry helpers for Tempo local state.
+///
+/// We keep the read/parse and atomic write logic here so `keys.toml`,
+/// `sessions.toml`, and any future Tempo registry files all use the same
+/// persistence semantics instead of duplicating the same boilerplate.
+pub(crate) fn read_toml_file<T: DeserializeOwned>(path: &Path, label: &str) -> Option<T> {
```
zerosnacks: **Blocking.** `Option<T>` collapses "file missing" and "file present but unreadable/corrupt" into the same `None`. Combined with `.unwrap_or_default()` at the upsert sites in `session.rs` and `keystore.rs`, a truncated/corrupt `keys.toml` or `sessions.toml` will be replaced by a file containing only the newly inserted entry, dropping all prior state.

For auth/session state please change this to something like `Result<Option<T>>` (`Ok(None)` = missing, `Err(_)` = corrupt) and have mutating paths fail closed on corruption. The read-only convenience wrappers can still collapse `Err` → warning + `None` if you want to keep the current discovery ergonomics.

```rust
pub(crate) fn read_toml_file<T: DeserializeOwned>(path: &Path, label: &str) -> eyre::Result<Option<T>> {
    match fs::read_to_string(path) {
        Ok(contents) => Ok(Some(toml::from_str(&contents)?)),
        Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(None),
        Err(e) => Err(e).wrap_err_with(|| format!("failed to read {label} file {}", path.display())),
    }
}
```

zerosnacks (commented):
Scaffold direction is fine, some notes 👍 . 

Main concern: the new shared read helper conflates "missing" with "corrupt", and both upsert paths then `unwrap_or_default()` on top of it — so a truncated/corrupt registry would be silently overwritten with only the new entry on the next mutation. Secondary concern: `write_toml_file_atomic` gives atomic *replacement* but not crash-safe *durability* (no fsync on file or parent dir).

crates/common/src/tempo/registry.rs
```diff
@@ -0,0 +1,55 @@
...
+        path.parent().ok_or_else(|| eyre::eyre!("invalid registry path: {}", path.display()))?;
+    fs::create_dir_all(dir)?;
+
+    let body = toml::to_string_pretty(value)?;
+    let contents = if header.trim().is_empty() { body } else { format!("{header}\n\n{body}") };
+
+    let mut tmp = tempfile::NamedTempFile::new_in(dir)?;
+    tmp.write_all(contents.as_bytes())?;
+    tmp.flush()?;
+    tmp.persist(path).map_err(|e| eyre::eyre!("failed to persist {}: {e}", path.display()))?;
```
zerosnacks, line 57: `NamedTempFile::persist` gives atomic *replacement* (incl. on Windows), but the temp file's contents are never `sync_all()`'d, and the parent directory entry isn't fsynced either — so under crash/power-loss the rename may not be durable, or the file may exist with the previous/empty contents.

Reth's `atomic_write_file` is a good reference for the pattern: write → fsync file → rename → fsync parent dir:

https://github.com/paradigmxyz/reth/blob/5568b76d59aa157c9dc90ca629d18275840a2881/crates/fs-util/src/lib.rs#L317-L351

Either adopt that sequence here (cheap and correct) or document explicitly that this helper only guarantees atomic visibility, not durability — so future callers don't assume it survives a hard crash.

crates/common/src/tempo/session.rs
```diff
@@ -0,0 +1,270 @@
...
+pub fn read_session_record() -> Option<SessionRecord> {
+    let path = session_registry_path()?;
+    read_toml_file(&path, "tempo sessions")
+}
+
+/// Atomically upsert a [`SessionEntry`] into the session registry.
+pub fn upsert_session_entry(entry: SessionEntry) -> eyre::Result<()> {
+    let path =
+        session_registry_path().ok_or_else(|| eyre::eyre!("could not resolve tempo home"))?;
+    let mut record = read_session_record().unwrap_or_default();
```
zerosnacks: Same blocking issue as in `registry.rs`: `read_session_record()` returns `None` on both missing *and* corrupt, and `.unwrap_or_default()` here will then atomically persist a record containing only `entry`, silently discarding all previously stored sessions if `sessions.toml` existed but failed to parse. Resolve once `read_toml_file` distinguishes the two cases.

crates/common/src/tempo/keystore.rs
```diff
@@ -168,25 +151,16 @@ pub fn decode_key_authorization<T: Decodable>(hex_str: &str) -> eyre::Result<T>
 /// temp file + rename so a crash mid-write cannot corrupt the file.
 pub(crate) fn upsert_key_entry(entry: KeyEntry) -> eyre::Result<()> {
     let path = tempo_keys_path().ok_or_else(|| eyre::eyre!("could not resolve tempo home"))?;
-    let dir = path.parent().ok_or_else(|| eyre::eyre!("invalid keys path: {}", path.display()))?;
-    fs::create_dir_all(dir)?;
-
     let mut file = read_tempo_keys_file().unwrap_or_default();
```
zerosnacks: Same issue as `upsert_session_entry`: a corrupt `keys.toml` would be silently overwritten with a single-entry file here. The bug exists pre-PR in spirit, but this PR moves the read into the shared helper, so fixing the helper to surface parse errors will fix both call sites in one go.

crates/common/src/tempo/session.rs
```diff
@@ -0,0 +1,270 @@
...
+}
+
+/// Persisted metadata for one temporary session.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub struct SessionEntry {
+    pub session_id: B256,
+    pub root_account: Address,
+    pub chain_id: u64,
+    pub key_address: Address,
+    pub expiry: u64,
```
zerosnacks, line 74: `SessionEntry::expiry: u64` is mandatory here, while `KeyEntry::expiry: Option<u64>` is optional — fine if intentional, but please add a doc comment clarifying:

- units (Unix seconds?)
- that `0` is **not** a "never expires" sentinel (other Tempo code treats `expiry == 0` that way for keys; sessions should not inherit that meaning accidentally)
- that sessions are always bounded-lifetime

Otherwise `is_expired_at(now)` will quietly return `true` for any `entry { expiry: 0 }` row that round-trips through the registry.

crates/common/src/tempo/mod.rs
```diff
@@ -11,10 +11,13 @@ use foundry_wallets::{RawWalletOpts, WalletOpts, WalletSigner};
 use std::sync::Arc;
 
 mod keystore;
+mod registry;
+mod session;
 
 pub(crate) use auth::is_known_tempo_endpoint;
 pub use auth::{AccessKeyOutcome, EnsureAccessKeyConfig, ensure_access_key};
 pub use keystore::*;
+pub use session::*;
```
zerosnacks, line 22: `pub use session::*;` exports the entire module — including `WALLET_SESSIONS_PATH`, `session_registry_path`, `read_session_record`, `upsert_session_entry`, `remove_session_entry` — but none of these have external consumers yet in this PR. Consider narrowing to `pub(crate)` for the file-path / read / mutate helpers, and only re-export the data model (`SessionEntry`, `SessionRecord`, `SessionStatus`, scope/limit/selector types) until a concrete external use lands. Glob re-exports here mirror `keystore::*` but lock in a wider API surface than the scaffold needs.

crates/common/src/tempo/session.rs
```diff
@@ -0,0 +1,270 @@
...
+pub fn upsert_session_entry(entry: SessionEntry) -> eyre::Result<()> {
+    let path =
+        session_registry_path().ok_or_else(|| eyre::eyre!("could not resolve tempo home"))?;
+    let mut record = read_session_record().unwrap_or_default();
+    record.upsert(entry);
+
+    write_toml_file_atomic(
+        &path,
+        &record,
+        "# Tempo session registry — managed by Foundry / Tempo CLI.\n# Do not edit manually.",
```
zerosnacks: Nit: the session header literal (`"# Tempo session registry — managed by Foundry / Tempo CLI.\n# Do not edit manually."`) appears in both `upsert_session_entry` and `remove_session_entry`. Extract a `const SESSIONS_HEADER: &str = ...;` so future edits stay in sync.

crates/common/src/tempo/session.rs
```diff
@@ -0,0 +1,270 @@
...
+    Ok(removed)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::tempo::{TEMPO_HOME_ENV, test_env_mutex};
+    use std::str::FromStr;
+
+    fn with_tempo_home<F: FnOnce()>(f: F) {
```
zerosnacks: Nit: `with_tempo_home` is duplicated verbatim from `keystore.rs` tests. Consider moving it (and `test_env_mutex` access) into a shared `#[cfg(test)] mod test_utils` under `tempo/` so future tempo modules don't keep re-implementing it.

crates/common/src/tempo/mod.rs
zerosnacks: Fine as is

0xKarl98:
@zerosnacks  ptal again , have addressed your comments

mablr (approved):
All comments seem correctly addressed, sgtm.

mattsse (approved):
LGTM

Issue #7481, opened 2024-03-24
Foundry panics after executing 3 transactions on a forked state


dsalv:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### What version of Foundry are you on?

forge 0.2.0 (7922fd5 2024-02-04T20:07:05.638459000Z)

### What command(s) is the bug in?

forge test

### Operating System

macOS (Apple Silicon)

### Describe the bug

`// SPDX-License-Identifier: Unlicense
pragma solidity ^0.8.6;

import "forge-std/console.sol";
import "ds-test/test.sol";

// Cheatcodes
interface Vm {
    struct Log {
        bytes32[] topics;
        bytes data;
    }

    function createFork(string memory, uint256) external returns (uint256);

    function selectFork(uint256) external;

    function transact(bytes32) external;

    function recordLogs() external;

    function getRecordedLogs() external view returns (Log[] memory);
}

// ERC20 interface for the shitcoin
interface TokenInterface {
    function balanceOf(address) external view returns (uint256);
}

contract ForkSimulate is DSTest {
    Vm vm = Vm(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);

    function testTransact() public {

        // Enter forking mode at block: https://etherscan.io/block/{block_number}
        uint256 fork = vm.createFork("https://rpc.notadegen.com/eth", 15596643);
        vm.selectFork(fork);

        // a random transfer transaction in the block: https://etherscan.io/tx/0xaba74f25a17cf0d95d1c6d0085d6c83fb8c5e773ffd2573b99a953256f989c89
[...]

PR #14828, opened 2026-05-19
fix(cheatcodes): fix expectRevert with internal reverts for non-contract calls

### LINKED ISSUE #11559: expectRevert bug with internal call

decofe:
## Summary

Fixes #11559.

When `allow_internal_expect_revert = true`, calling a non-contract address (e.g. `IERC20(address(0)).totalSupply()`) inside `expectRevert` would fail with *"next call did not revert as expected"*.

## Root cause

Calls to addresses with no code return immediately with `Stop`. The previous `call_end` logic consumed the `expectRevert` on this success, so when Solidity's generated extcodesize/returndata check then reverted, the expectation was already gone.

## Fix

In `call_end`, only consume the `expectRevert` when one of these holds:
1. the call actually failed, or
2. an external call went deeper (`max_depth > depth`), or
3. we're at the test function's root frame (catches dangling expectations), or
4. `internal_expect_revert` is disabled (preserves prior behavior).

With `internal_expect_revert` on, a same-depth success no longer eats the expectation, so Solidity's follow-up revert can satisfy it. The new repro lives in [`testdata/default/repros/Issue11559.t.sol`](testdata/default/repros/Issue11559.t.sol).

## Note on `Issue7481.t.sol`

The dangling-at-root processing (point 3 above) surfaces a latent issue in the existing `testRevertTransact` repro: it asserts the literal string `"vm.createSelectFork: invalid rpc url: mainnet"`, which only matched when CI's `RPC_MAINNET` happened to be unset/invalid. On master the dangling expectation was silently swallowed, hiding the brittleness. This PR makes the test deterministic by using a literal invalid URL instead of the `mainnet` alias.

The original #7481 panic (looped `vm.transact` on a real fork) was not actually exercised by that test before and is still not exercised after this change.

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -1421,15 +1421,6 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>> for Cheatcode
             }
 
             let curr_depth = ecx.journal().depth();
-            let selector = call.input.bytes(ecx);
-            let selector = selector.get(..SELECTOR_LEN);
-            let process_cheatcode_revert = cheatcode_call
-                && call_failed
-                && selector.is_some_and(|selector| {
-                    selector == Vm::createSelectFork_0Call::SELECTOR
-                        || selector == Vm::createSelectFork_1Call::SELECTOR
-                        || selector == Vm::createSelectFork_2Call::SELECTOR
-                });
             if curr_depth <= expected_revert.depth {
                 let needs_processing = match expected_revert.kind {
                     ExpectedRevertKind::Default => {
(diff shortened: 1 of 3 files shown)


mablr (approved):
lgtm

stevencartavia (approved):

PR #14829, opened 2026-05-19
fix(lint): detect named ERC20 transfer args

mattsse:
## Summary

Teach the ERC20 unchecked-transfer lint to use the total call argument count instead of matching only positional call arguments. This lets it catch `transfer` and `transferFrom` calls written with Solidity named arguments.

## Details

The `unused-return` lint added in #14785 delegates ERC20 transfer calls to `erc20-unchecked-transfer`, but that lint previously ignored `CallArgsKind::Named`. The fixture now covers unchecked named-argument calls and a checked named-argument call so the behavior stays anchored.


mablr (approved):
noice

PR #14831, opened 2026-05-19
fix(lint): cover delegatecall-loop followups

mattsse:
## Summary
- infer a conservative common type for ternary expressions used as delegatecall receivers
- resolve `super.<helper>(...)` internal calls while tracking the active contract during helper and modifier traversal
- add delegatecall-loop UI coverage for conditional address receivers and inherited `super` helpers

## Context
Follow-up to #14752 and the review notes in https://github.com/foundry-rs/foundry/pull/14752#pullrequestreview-4318566533.

### DIFF
--- a/crates/lint/src/sol/low/delegatecall_loop.rs
+++ b/crates/lint/src/sol/low/delegatecall_loop.rs
@@ -5,12 +5,12 @@ use crate::{
 };
 use solar::{
     ast::{ElementaryType, LitKind, StateMutability, StrKind, TypeSize, Visibility},
-    interface::{Span, kw},
+    interface::{Span, kw, sym},
     sema::{
         Gcx, Ty,
         hir::{
-            Block, CallArgs, CallArgsKind, Expr, ExprKind, Function, FunctionId, Hir, ItemId,
-            Modifier, Res, Stmt, StmtKind, VariableId, Visit,
+            Block, CallArgs, CallArgsKind, ContractId, Expr, ExprKind, Function, FunctionId, Hir,
+            ItemId, Modifier, Res, Stmt, StmtKind, VariableId, Visit,
         },
         ty::TyKind,
     },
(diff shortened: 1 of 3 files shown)


crates/lint/src/sol/low/delegatecall_loop.rs
```diff
@@ -189,7 +200,7 @@ impl<'hir> DelegatecallLoopChecker<'_, '_, 'hir> {
         let Some(body) = func.body else { return };
 
         self.call_stack.push(func_id);
-        self.visit_modifier_chain(func.modifiers, 0, body);
+        self.visit_modifier_chain(func.modifiers, 0, body, func.contract);
```
mablr, line 205: Passing `func.contract` as `current_contract`, overwrites the dispatch contract. Any super call inside this body walks the callee's own C3 instead of the dispatch contract's C3 past the callee's index, causing wrong `super` resolution under multiple inheritance.

crates/lint/src/sol/low/delegatecall_loop.rs
```diff
@@ -407,16 +444,31 @@ fn expr_ty<'gcx>(gcx: Gcx<'gcx>, hir: &Hir<'gcx>, expr: &Expr<'gcx>) -> Option<T
                 .collect::<Option<Vec<_>>>()?;
             Some(gcx.mk_ty_tuple(gcx.mk_tys(&tys)))
         }
+        ExprKind::Ternary(cond, true_expr, false_expr) => {
+            if !expr_ty(gcx, hir, cond)?.convert_implicit_to(gcx.types.bool, gcx) {
+                return None;
+            }
```
mablr: `expr_ty` returns `None` for `Binary` (comparisons, logical ops), so it propagates `None`, silently suppressing the address check false negative. As HIR is already type-checked, I think we can drop this cond guard entirely.

mablr (approved):
lgtm

figtracer (approved):

PR #14832, opened 2026-05-19
ci: add stale PR workflow

emmajam:
## Summary

- Add a scheduled `stale-prs` workflow that runs daily and can be triggered manually.
- Mark PRs as `stale` after they have been open for 14 days.
- Close stale PRs after 7 more days with no activity, and add `stop-stale` when activity happens after the stale warning.

## Testing

- Parsed the workflow YAML with Ruby.
- Ran `bash -n` against the embedded shell script.
- Ran `shellcheck` against the embedded shell script.

### DIFF
--- a/.github/workflows/stale-prs.yml
+++ b/.github/workflows/stale-prs.yml
@@ -44,16 +44,6 @@ jobs:
             "$close_marker" \
             "Closing this pull request because it had no activity for ${CLOSE_AFTER_DAYS} days after being marked stale. Please reopen it or open a new pull request if it is still relevant."
 
-          gh label create "$STALE_LABEL" \
-            --color BFDADC \
-            --description "Pull requests pending stale closure" \
-            --force
-
-          gh label create "$STOP_STALE_LABEL" \
-            --color 0E8A16 \
-            --description "Prevent stale automation from closing this pull request" \
-            --force
-
           stale_cutoff_ts="$(date -u -d "${STALE_AFTER_DAYS} days ago" +%s)"
           close_after_seconds="$((CLOSE_AFTER_DAYS * 24 * 60 * 60))"
 
--- a/.github/workflows/stale-prs.yml
+++ b/.github/workflows/stale-prs.yml
@@ -10,7 +10,7 @@ on:
 permissions: {}
 
 jobs:
-  close-prs:
+  stale-prs:
     if: github.repository == 'foundry-rs/foundry'
     runs-on: ubuntu-latest
     permissions:
--- a/.github/workflows/stale-prs.yml
+++ b/.github/workflows/stale-prs.yml
@@ -9,7 +9,7 @@ on:
 
 concurrency:
   group: stale-prs
-  cancel-in-progress: true
+  cancel-in-progress: false
 
 jobs:
   stale-prs:
--- a/.github/workflows/stale-prs.yml
+++ b/.github/workflows/stale-prs.yml
@@ -1,0 +1,157 @@
+name: stale-prs
+
+permissions: {}
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


figtracer (dismissed):

mattsse (dismissed):
Approved per maintainer direction.

.github/workflows/stale-prs.yml
```diff
@@ -0,0 +1,157 @@
...
+permissions: {}
+
+on:
+  schedule:
+    - cron: "15 4 * * *" # Run daily at 4:15 AM UTC.
+  workflow_dispatch:
+
+concurrency:
+  group: stale-prs
+  cancel-in-progress: true
```
zerosnacks: edge case but we do not want partial runs

```suggestion
  cancel-in-progress: false
```

.github/workflows/stale-prs.yml
```diff
@@ -0,0 +1,157 @@
...
+
+          gh label create "$STALE_LABEL" \
+            --color BFDADC \
+            --description "Pull requests pending stale closure" \
+            --force
+
+          gh label create "$STOP_STALE_LABEL" \
+            --color 0E8A16 \
+            --description "Prevent stale automation from closing this pull request" \
+            --force
```
zerosnacks: I don't think this redefining of `gh label` is required, we just need to define it once and I think you did already

zerosnacks (changes requested):
I think Reth has a simpler version of this that achieves the same goal and uses https://github.com/actions/stale: https://github.com/paradigmxyz/reth/blob/main/.github/workflows/stale.yml

.github/workflows/stale-prs.yml
```diff
@@ -0,0 +1,33 @@
...
+    if: github.repository == 'foundry-rs/foundry'
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@b5d41d4e1d5dceea10e7104786b73624c18a190f # v10.2.0
+        with:
+          days-before-issue-stale: -1
+          days-before-issue-close: -1
```
emmajam, line 23: we can do issues too

emmajam: but will leave for now

zerosnacks: yes agreed

.github/workflows/stale-prs.yml
```diff
@@ -0,0 +1,33 @@
...
+
+on:
+  workflow_dispatch: {}
+  schedule:
+    - cron: "30 1 * * *"
+
+permissions: {}
+
+jobs:
+  close-prs:
```
zerosnacks:
```suggestion
  stale-prs:
```

.github/workflows/stale-prs.yml
```diff
@@ -0,0 +1,33 @@
...
+          days-before-pr-stale: 14
+          days-before-pr-close: 7
+          stale-pr-label: "stale"
+          exempt-pr-labels: "stop-stale"
+          stale-pr-message: "This pull request is stale because it has been open for 14 days with no activity. It will be closed in 7 days if there is no further activity."
+          close-pr-message: "This pull request was closed because it has been inactive for 7 days since being marked as stale."
+          labels-to-add-when-unstale: "stop-stale"
+          exempt-all-milestones: true
+          exempt-all-assignees: true
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
zerosnacks:
```suggestion
```

I think we can rely on the default here: https://github.com/marketplace/actions/close-stale-issues#repo-token

no need to pass it in

zerosnacks (approved):

figtracer (approved):

PR #14833, opened 2026-05-19
fix(anvil): keep chain-id inferred network in node config

esorense:
## Motivation
                                                                                                                                                                                                 
  `anvil --chain-id <id>` should infer the correct network family from the chain ID, but in `NodeArgs::into_node_config` the inferred network could be overwritten later by `with_networks(self.evm.networks)`.                                                                                                                                                            
  This made chain-id-based inference unreliable on the CLI path (for example, OP chain IDs).                                                                                                     
                                                                                                                                                                                                 
  ## Solution                                                                                                                                                                                    
                                                                                                                                                                                                 
  Reordered the config builder calls so `with_chain_id(self.evm.chain_id)` runs after `with_networks(self.evm.networks)`, ensuring inferred network selection is preserved unless explicitly set by flags.                                                                                                                                                                                      
  Added `chain_id_infers_optimism_network_in_node_config` to cover the exact `NodeArgs -> NodeConfig` regression path and assert `--chain-id 10` resolves to Optimism.                           
                                                                                                                                                                                                 
  ## PR Checklist                                                                                                                                                                                
                                                                                                                                                                                                 
  - [x] Added Tests                                                                                                                                                                              
  - [ ] Added Documentation                                                                                                                                                                      
  - [ ] Breaking changes


stevencartavia (approved):

mablr (approved):
Makes sense, thanks.

PR #14835, opened 2026-05-19
feat(symbolic): add solver selection plumbing

figtracer:
## Summary
- add `symbolic.solver_command`, `--symbolic-solver-command`, and `FOUNDRY_SYMBOLIC_SOLVER_COMMAND`
- add `symbolic.solver_portfolio`, `--symbolic-solver-portfolio`, and `FOUNDRY_SYMBOLIC_SOLVER_PORTFOLIO` to race multiple SMT subprocesses in parallel
- resolve known SMT-LIB solver names (`z3`, `yices`, `cvc5`, `cvc5-int`, `bitwuzla`, `bitwuzla-abs`) into solver-specific argv/model-output commands while preserving custom z3-compatible executable paths
- translate Halmos `@custom:halmos --solver` and `--solver-command` annotations into native symbolic config
- parse solver models from decimal, hex, and binary bit-vector values, validate returned models against path constraints, and document the local-command trust boundary

This keeps install/download/cache policy out of the first PR. Follow-up directions:
- evaluate jsi-style portfolio orchestration for richer cancellation/scoring once we have performance data: https://github.com/a16z/jsi
- keep exploration strategy/performance work separate; Greed DFS/prioritizer docs are useful references for matching exhaustive verification vs bug-finding search modes: https://ucsb-seclab.github.io/greed/modules/greed.exploration_techniques.dfs/

### DIFF
--- a/crates/config/src/inline/natspec.rs
+++ b/crates/config/src/inline/natspec.rs
@@ -145,10 +145,10 @@ impl NatSpec {
 }
 
 fn translate_halmos_config(args: &str, values: &mut Vec<String>) -> Result<(), String> {
-    let tokens = args.split_whitespace().collect::<Vec<_>>();
+    let tokens = split_halmos_args(args)?;
     let mut idx = 0;
     while idx < tokens.len() {
-        let token = tokens[idx];
+        let token = tokens[idx].as_str();
         if token == "--array-lengths" {
             idx += 1;
             let Some(value) = tokens.get(idx) else {
@@ -193,6 +193,22 @@ fn translate_halmos_config(args: &str, values: &mut Vec<String>) -> Result<(), S
             push_halmos_u32(values, "timeout", value, "--solver-timeout-branching")?;
         } else if let Some(value) = token.strip_prefix("--solver-timeout-assertion=") {
             push_halmos_u32(values, "timeout", value, "--solver-timeout-assertion")?;
+        } else if token == "--solver" {
+            idx += 1;
+            let Some(value) = tokens.get(idx) else {
+                return Err("missing value for --solver".to_string());
+            };
+            push_halmos_string(values, "solver", value);
+        } else if let Some(value) = token.strip_prefix("--solver=") {
+            push_halmos_string(values, "solver", value);
+        } else if token == "--solver-command" {
+            idx += 1;
+            let Some(value) = tokens.get(idx) else {
+                return Err("missing value for --solver-command".to_string());
+            };
+            push_halmos_string(values, "solver_command", value);
+        } else if let Some(value) = token.strip_prefix("--solver-command=") {
+            push_halmos_string(values, "solver_command", value);
         } else if token.starts_with("--")
             && tokens.get(idx + 1).is_some_and(|next| !next.starts_with("--"))
         {
@@ -201,6 +217,54 @@ fn translate_halmos_config(args: &str, values: &mut Vec<String>) -> Result<(), S
         idx += 1;
     }
     Ok(())
+}
+
+fn split_halmos_args(args: &str) -> Result<Vec<String>, String> {
+    let mut tokens = Vec::new();
+    let mut current = String::new();
+    let mut quote = None;
+    let mut escaped = false;
+
+    for ch in args.chars() {
+        if escaped {
+            current.push(ch);
+            escaped = false;
+            continue;
+        }
+        if ch == '\\' {
+            escaped = true;
+            continue;
+        }
+        if let Some(quote_ch) = quote {
+            if ch == quote_ch {
+                quote = None;
+            } else {
+                current.push(ch);
+            }
+            continue;
+        }
+        if matches!(ch, '"' | '\'') {
+            quote = Some(ch);
+        } else if ch.is_whitespace() {
+            if !current.is_empty() {
+                tokens.push(std::mem::take(&mut current));
+            }
+        } else {
+            current.push(ch);
+        }
+    }
+
+    if let Some(quote_ch) = quote {
+        return Err(format!("unterminated {quote_ch} quote in @custom:halmos config"));
+    }
+    if escaped {
+        current.push('\\');
+    }
+    if !current.is_empty() {
+        tokens.push(current);
+    }
+
+    Ok(tokens)
 }
 
 fn halmos_numeric_symbolic_field(flag: &str) -> Option<&'static str> {
@@ -258,6 +322,14 @@ fn push_halmos_u32(
 ) -> Result<(), String> {
     values.push(format!("default.symbolic.{field} = {}", parse_halmos_u32(value, flag)?));
     Ok(())
+}
+
+fn push_halmos_string(values: &mut Vec<String>, field: &str, value: &str) {
+    values.push(format!("default.symbolic.{field} = {}", toml_string(value)));
+}
+
+fn toml_string(value: &str) -> String {
+    format!("\"{}\"", value.replace('\\', "\\\\").replace('"', "\\\""))
 }
 
 fn parse_halmos_u32(value: &str, flag: &str) -> Result<u32, String> {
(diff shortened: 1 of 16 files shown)


figtracer:
@grandizzy can you ptal? should be a nice first step

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -100,15 +142,32 @@ impl SymbolicSolver for Z3SubprocessSolver {
         let output = self.query(constraints, true)?;
```
grandizzy, line 145: if we run multiple solvers then we should only accept a portfolio `sat` if its model parses and satisfies the
constraints; otherwise let other solvers in the portfolio win. In current impl if output is due to a malformed/unsupported/invalid model of the first solver picked up, then the whole counterexample fails even when a slower solver would have produced a usable one.

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -148,29 +210,340 @@ impl Z3SubprocessSolver {
...
+            kill_and_reap_solver_process(&mut child, stdout_reader, stderr_reader);
+            return SolverProcessOutcome::Cancelled;
+        }
+        if deadline.is_some_and(|deadline| Instant::now() >= deadline) {
+            kill_and_reap_solver_process(&mut child, stdout_reader, stderr_reader);
+            return SolverProcessOutcome::Unknown;
+        }
+        match child.try_wait() {
+            Ok(Some(status)) => break status,
+            Ok(None) => thread::sleep(Duration::from_millis(10)),
```
grandizzy: let's make this adaptive - start tight so fast queries (most branch-feasibility checks finish in <5ms) aren't padded by a 10ms polling tick and back off to 50ms so long-running queries don't burn CPU on `try_wait`.

```rust
    let mut backoff = Duration::from_micros(200);
    const MAX_BACKOFF: Duration = Duration::from_millis(50);
...
            Ok(None) => {
                thread::sleep(backoff);
                backoff = (backoff * 2).min(MAX_BACKOFF);
            }
```

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -148,29 +210,340 @@ impl Z3SubprocessSolver {
...
+        None => Ok(String::new()),
+    }
+}
+
+fn kill_and_reap_solver_process(
+    child: &mut std::process::Child,
+    stdout_reader: Option<thread::JoinHandle<Result<String, String>>>,
+    stderr_reader: Option<thread::JoinHandle<Result<String, String>>>,
+) {
+    let _ = child.kill();
```
grandizzy, line 482: this only kills the direct child so if custom `solver_command` used (e.g. wrappers, docker, interpreters) then grandchild solvers can survive timeout/cancel and leave reader threads blocked on inherited pipes. Maybe should warn when such or at min document the behavior.

crates/config/src/inline/natspec.rs
```diff
@@ -193,6 +193,22 @@ fn translate_halmos_config(args: &str, values: &mut Vec<String>) -> Result<(), S
...
             push_halmos_u32(values, "timeout", value, "--solver-timeout-assertion")?;
+        } else if token == "--solver" {
+            idx += 1;
+            let Some(value) = tokens.get(idx) else {
+                return Err("missing value for --solver".to_string());
+            };
+            push_halmos_string(values, "solver", value);
+        } else if let Some(value) = token.strip_prefix("--solver=") {
+            push_halmos_string(values, "solver", value);
+        } else if token == "--solver-command" {
```
grandizzy, line 205: here `@custom:halmos --solver-command` lets a test contract's NatSpec inject an arbitrary local argv that runs on `forge test --symbolic`. This is analogous to `--ffi` and should get a comparable runtime warning (or opt-in flag) before running symbolic tests from untrusted projects.

crates/config/src/inline/natspec.rs
```diff
@@ -203,6 +219,54 @@ fn translate_halmos_config(args: &str, values: &mut Vec<String>) -> Result<(), S
     Ok(())
 }
 
+fn split_halmos_args(args: &str) -> Result<Vec<String>, String> {
+    let mut tokens = Vec::new();
+    let mut current = String::new();
+    let mut quote = None;
+    let mut escaped = false;
+
+    for ch in args.chars() {
```
grandizzy: this looks identical with logic within solver, let's reuse https://github.com/foundry-rs/foundry/pull/14835/changes#diff-cd8fe5fa52c294e84ede075f494fb0e54fd27624bd2696a60c05dd50ad55f359R292

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -148,29 +210,340 @@ impl Z3SubprocessSolver {
...
+            } else {
+                current.push(ch);
+            }
+            continue;
+        }
+        if matches!(ch, '"' | '\'') {
+            quote = Some(ch);
+        } else if ch.is_whitespace() {
+            if !current.is_empty() {
+                parts.push(std::mem::take(&mut current));
```
grandizzy: we should push also if had quote, otherwise it drops empty quoted args (`cmd "" arg` becomes 2 args, not 3)

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -148,29 +210,340 @@ impl Z3SubprocessSolver {
...
+        }
+    }
+
+    if let Some(quote_ch) = quote {
+        return Err(format!("unterminated {quote_ch} quote in symbolic solver command"));
+    }
+    if escaped {
+        current.push('\\');
+    }
+    if !current.is_empty() {
```
grandizzy: same as above

crates/evm/symbolic/README.md
grandizzy:
```suggestion
- `SymbolicSolver` is the small internal trait used by the default SMT-LIB subprocess backend, which resolves named solvers (z3, cvc5, yices, bitwuzla, etc.) into solver-specific argv via `solver_commands_for_config`.
```

crates/evm/symbolic/src/lib.rs
grandizzy:
```suggestion
/// SMT-LIB-backed symbolic executor.
```

grandizzy (changes requested):
overall looks good, pls check comments

crates/evm/symbolic/src/lib.rs
figtracer: addressed by [`f282d20` (this PR)](https://github.com/foundry-rs/foundry/pull/14835/commits/f282d20a7444fbdc6acd3e47330c0739d0763f9e)

crates/evm/symbolic/README.md
figtracer: addressed by [`f282d20` (this PR)](https://github.com/foundry-rs/foundry/pull/14835/commits/f282d20a7444fbdc6acd3e47330c0739d0763f9e)

crates/config/src/inline/natspec.rs
figtracer: addressed by [`f282d20` (this PR)](https://github.com/foundry-rs/foundry/pull/14835/commits/f282d20a7444fbdc6acd3e47330c0739d0763f9e)

crates/evm/symbolic/src/runtime/solver.rs
figtracer: nice catch, addressed by [`f282d20` (this PR)](https://github.com/foundry-rs/foundry/pull/14835/commits/f282d20a7444fbdc6acd3e47330c0739d0763f9e)

crates/config/src/inline/natspec.rs
figtracer: addressed by [`f282d20` (this PR)](https://github.com/foundry-rs/foundry/pull/14835/commits/f282d20a7444fbdc6acd3e47330c0739d0763f9e)

currently running extensively more live tests with all the solvers, will push a new commit asap to fix this

crates/evm/symbolic/src/runtime/solver.rs
figtracer: nice, addressed by [`f282d20` (this PR)](https://github.com/foundry-rs/foundry/pull/14835/commits/f282d20a7444fbdc6acd3e47330c0739d0763f9e)

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -538,6 +498,10 @@ fn solver_output_is_decisive(output: &str) -> bool {
     matches!(first_solver_line(output), "sat" | "unsat")
 }
 
+fn solver_output_is_sat(output: &str) -> bool {
+    first_solver_line(output) == "sat"
```
figtracer, line 507: not sure how fragile this is, live tests ran okay for every solver

grandizzy (approved):
lgtm! one nit re tests - let's use snapbox snapshots instead https://github.com/foundry-rs/foundry/pull/14835/changes#diff-b88fe61c97e1bf4847da968d63e48a7817e96cd1d2c17c354b544337597bbf0fR4222-R4228

PR #14838, opened 2026-05-20
fix(cast): resolve tempo expiry in helper commands

Perico-perica46:
## Motivation

Some Tempo-specific `cast` helpers accept `--tempo.expires` but apply the shared transaction options directly when building their requests. That leaves the relative expiry to be materialized at the application point instead of resolving it once up front, unlike `cast send`, `cast mktx`, and `cast erc20`.

## Solution

Resolve `--tempo.expires` before building transactions for the batch, keychain, TIP20, and virtual-address flows, then print the materialized unix timestamp through a shared helper before submission.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cast/src/tx.rs
+++ b/crates/cast/src/tx.rs
@@ -106,13 +106,6 @@ impl TxParams {
     }
 }
 
-pub(crate) fn print_tempo_expires(expires_at: Option<u64>) -> Result<()> {
-    if let Some(ts) = expires_at {
-        sh_println!("Transaction expires at unix timestamp {ts}")?;
-    }
-    Ok(())
-}
-
 /// Different sender kinds used by [`CastTxBuilder`].
 pub enum SenderKind<'a> {
     /// An address without signer. Used for read-only calls and transactions sent through unlocked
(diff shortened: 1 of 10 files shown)


crates/cast/src/tx.rs
```diff
@@ -106,6 +106,13 @@ impl TxParams {
     }
 }
 
+pub(crate) fn print_tempo_expires(expires_at: Option<u64>) -> Result<()> {
+    if let Some(ts) = expires_at {
+        sh_println!("Transaction expires at unix timestamp {ts}")?;
+    }
+    Ok(())
+}
```
mablr: Maybe it worths to put this helper in tempo mod?

mablr (commented):
Makes sense

Perico-perica46:
@mablr Updated.

mattsse (approved):
Reviewed the Tempo expiry handling. This mirrors the existing cast send/mktx/ERC20 flow and the focused CLI test passes.

mablr:
Looks good, thank you @Perico-perica46 !

mablr (approved):

PR #14843, opened 2026-05-20
feat: print optional opcodes in forge test

### LINKED ISSUE #8198: Add printing of OPCODEs in forge tests
### Component

Forge

One feature that is missing in the tooling would be to print the opcodes that re hit when testing contracts.

For example https://github.com/zemse/hardhat-tracer has the option for you to set if you want to print SLOADs, MLOADs and other OPCODES when testing and also their gas usage. This would be beneficial of gas analysis gas optimizations. 

I know with the debugger you can see the opcodes but its hard to navigate and you can't filter only specific opcodes there

image

This maybe can be added as a verbosity level 6 or another way where you can specify which opcodes you want printed.

### Additional context

_No response_

jaimebarrancos:
Added `--opcodes` argument in forge test.
This allows opcodes to be analyzed in more detail as well as their gas costs and storage changes (if any).

AI was not used to write any code.

This is my first contribution to open source!

## Motivation

Closes #8198.

## Solution

Passed opcode lines to the renderer.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/forge/src/cmd/test/mod.rs
```diff
@@ -848,11 +853,19 @@ impl TestArgs {
                     };
 
                     if should_include {
+                        if let Some(ref opcodes_str) = self.opcodes {
+                            decoder.opcodes = opcodes_str
+                                .split(",")
+                                .map(|opcode_name| OpCode::parse(opcode_name))
```
grandizzy: if any opcode is invalid then the feature is disabled without feedback, e.g. `--opcodes SLOAD,BOGUS`, `--opcodes ""`, and `--opcodes "SLOAD, MLOAD"` (leading space) all disable the feature - suggest validate with a clap `value_parser` that errors on unknown/empty names.

crates/evm/traces/src/decoder/mod.rs
```diff
@@ -454,6 +461,24 @@ impl CallTraceDecoder {
     /// [CallTraceDecoder::decode_event] for more details.
     pub async fn populate_traces(&self, traces: &mut Vec<CallTraceNode>) {
         for node in traces {
+
+            if let Some(codes) = &self.opcodes {
```
grandizzy: `revm-inspectors` populates `step.decoded` for internal call steps (e.g. precompile calls, internal function calls when source maps are available). The current PR's loop runs `step.decoded = Some(...)` unconditionally on every matching opcode, which silently destroys that pre-existing decoded info and replaces it with a bare [gas] OPCODE line. Below a suggested fix to skip steps that already have a decoded representation.

```diff
             if let Some(codes) = &self.opcodes {
                 for step in node.trace.steps.iter_mut() {
-
+                    // Don't clobber decoded info already attached by
+                    // revm-inspectors (e.g. internal call / precompile steps).
+                    if step.decoded.is_some() {
+                        continue;
+                    }
                     for opcode in codes {
                         if step.op == *opcode {
-                            
                             let res = match &step.storage_change {
-                                Some(change) => format!("[{}] {} {} <- ({})", &step.gas_cost.to_string(), opcode, change.key, change.value),
-                                None => format!("[{}] {}", &step.gas_cost.to_string(), opcode),
+                                Some(change) => format!("[{}] {} {} <- ({})", step.gas_cost, opcode, change.key, change.value),
+                                None => format!("[{}] {}", step.gas_cost, opcode),
                             };
-
                             step.decoded = Some(Box::new(DecodedTraceStep::Line(res)));
+                            break; // first matching opcode wins; no need to keep scanning
                         }
                     }
                 }
             }
-
```

crates/forge/src/cmd/test/mod.rs
```diff
@@ -195,6 +195,11 @@ pub struct TestArgs {
     #[arg(long)]
     pub rerun: bool,
 
+    /// Specify opcodes to analyze when verbosity level is 5.
```
grandizzy: that's not enforced, maybe add an explicit `eyre::bail!` when `--opcodes` is used without `-vvvvv`

grandizzy: maybe better
```
/// Print the given opcodes inline in trace output, with their gas cost
/// and (for SLOAD/SSTORE) the storage slot and value.
///
/// Accepts a comma-separated list of opcode names, e.g.
/// `--opcodes SLOAD,MLOAD,SSTORE`. Names are case-insensitive.
/// Requires `-vvvvv` to render.
```

grandizzy (dismissed):
thank you, direction is good, left couple of comments. Please make sure clippy and fmt pass, see https://github.com/foundry-rs/foundry/blob/master/CONTRIBUTING.md#resolving-an-issue for details

crates/forge/src/cmd/test/mod.rs
```diff
@@ -217,6 +217,15 @@ pub struct TestArgs {
...
     pub rerun: bool,
 
+    /// Print the given opcodes inline in trace output, with their gas cost
+    /// and (for SLOAD/SSTORE) the storage slot and value.
+    ///
+    /// Accepts a comma-separated list of opcode names, e.g.
+    /// `--opcodes SLOAD,MLOAD,SSTORE`. Names are case-insensitive.
+    /// Requires `-vvvvv` to render.
+    #[arg(long, value_parser = parse_opcode, value_delimiter(','), num_args(1..))]
+    pub opcodes: Vec<OpCode>,
```
mablr: The help says opcode names are case-insensitive, but `parse_opcode` delegates to `OpCode::parse`, which is an exact uppercase lookup.

crates/forge/src/cmd/test/mod.rs
```diff
@@ -921,6 +930,11 @@ impl TestArgs {
                 decoder.clear_addresses();
                 decoder.labels.extend(result.labels.iter().map(|(k, v)| (*k, v.clone())));
 
+                if !self.opcodes.is_empty() && verbosity < 5 {
+                        sh_eprintln!()?;
+                        eyre::bail!("Not enough verbosity. Use -vvvvv to show opcodes.");
+                }
```
mablr: This verbosity check runs after the test outcome has already been produced, so `forge test --opcodes SLOAD` can compile/run the suite and only then fail while rendering results. Consider validating it once near the start of `run_tests_inner`, right after verbosity is known, so the CLI fails fast.

mablr (commented):
Please fix the `Cargo.lock` file + 2 small nits

crates/forge/tests/cli/test_cmd/mod.rs
```diff
@@ -3042,6 +3042,51 @@ Ran 1 test suite [ELAPSED]: 1 tests passed, 0 failed, 0 skipped (1 total tests)
 "#]]);
 });
 
+// Tests that test traces display opcodes when verbosity level is 5
+forgetest_init!(should_show_opcodes, |prj, cmd| {
+    prj.initialize_default_contracts();
+    cmd.args(["test", "--mt", "test_Increment", "-vvvvv", "--opcodes", "SLOAD,MLOAD"])
+        .assert_success()
+        .stdout_eq(str![[r#"
```
mablr: This covers the happy path, but the fragile parts of this feature are the validation paths added in this PR: invalid opcode parsing and rejecting `--opcodes` below `-vvvvv`. Maybe add CLI tests for e.g. `--opcodes BOGUS` and `--opcodes SLOAD` without `-vvvvv`, so these fail-fast behaviors do not regress?

mablr (dismissed):
Sgtm, thanks.

Pending @grandizzy :)

crates/forge/src/cmd/test/mod.rs
```diff
@@ -217,6 +217,15 @@ pub struct TestArgs {
     #[arg(long)]
     pub rerun: bool,
 
+    /// Print the given opcodes in trace output, with their gas cost
+    /// and (for SLOAD/SSTORE) the storage slot and value.
+    ///
+    /// Accepts a comma-separated list of opcode names, e.g.
+    /// `--opcodes SLOAD,MLOAD,SSTORE`. Names are in uppercase.
+    /// Requires `-vvvvv` to render.
+    #[arg(long, value_parser = parse_opcode, value_delimiter(','), num_args(1..))]
```
stevencartavia: **`--opcodes` swallows the positional path**

`--opcodes` uses `num_args(1..)`, which makes it variadic. Since `forge test` already takes a positional `path` (shortcut for `--match-path`), the flag consumes a path that comes after it:

```bash
forge test test/Counter.t.sol --opcodes SLOAD -vvvvv   # works (path first)
forge test --opcodes SLOAD test/Counter.t.sol -vvvvv   # error: invalid opcode: test/Counter.t.sol
```

Putting the path at the end is a common pattern, so this breaks normal usage, and the error blaming the path as an "invalid opcode" is confusing. It can also fail silently when the path token is a valid opcode name — `forge test --opcodes SLOAD CALL` parses `CALL` as an opcode and runs the full suite instead of erroring.

Dropping `num_args(1..)` and keeping `value_delimiter(',')` fixes it:

```rust
#[arg(long, value_parser = parse_opcode, value_delimiter(','))]
pub opcodes: Vec<OpCode>,
```

This still supports `--opcodes SLOAD,MLOAD` and repeated `--opcodes` flags, while leaving the positional path usable. I confirmed it locally — one small follow-up: without `num_args(1..)` clap renders the arg as `<OPCODES>` instead of `<OPCODES>...`, so the `invalid_opcode_name` test needs its expected string updated to `'--opcodes <OPCODES>'`. Worth adding a regression test for path-after-flag too.

crates/evm/traces/src/decoder/mod.rs
```diff
@@ -513,6 +520,28 @@ impl CallTraceDecoder {
...
+                for step in &mut node.trace.steps {
+                    if step.decoded.is_some() {
+                        continue;
+                    }
+                    for opcode in &self.opcodes {
+                        if step.op == *opcode {
+                            let res = match &step.storage_change {
+                                Some(change) => format!(
+                                    "[{}] {} {} <- ({})",
+                                    step.gas_cost, opcode, change.key, change.value
```
mablr, line 564: `U256` values should not be printed as decimals ?

crates/evm/traces/src/decoder/mod.rs
```diff
@@ -513,6 +520,28 @@ impl CallTraceDecoder {
...
         for node in traces {
+            if !self.opcodes.is_empty() {
+                for step in &mut node.trace.steps {
+                    if step.decoded.is_some() {
+                        continue;
+                    }
+                    for opcode in &self.opcodes {
+                        if step.op == *opcode {
+                            let res = match &step.storage_change {
+                                Some(change) => format!(
```
mablr, line 562: `step.storage_change` is only `Some` when the EVM journal changed that step (cold access / actual write). A warm repeated SLOAD has `storage_change == None`, so the feature falls back to bare [100] SLOAD with no slot. The promise "for SLOAD/SSTORE, print slot and value" only holds sometimes, should be documented as "when available" or fixed at the inspector level?

mablr (commented):
made cargo file fix + 2 smol comments

crates/evm/traces/src/decoder/mod.rs
jaimebarrancos: Would it be better with hexadecimals? I noticed that most things are logged with hexadecimal, when specifically is it better to use hexa vs decimal for debugging?

crates/evm/traces/src/decoder/mod.rs
jaimebarrancos: That makes sense.

Maybe unrelated but shouldn't there be only one output of a storage change? I believe this case has only one storage change but shows 2:
```
    │   ├─ [20000] SSTORE 0 <- (1)
    │   ├─  storage changes:
    │   │   @ 0: 0 → 1
```
Also maybe it would be good to make the arrows point left to right for consistency?

Thanks!

crates/evm/traces/src/decoder/mod.rs
mablr: Yep hex is better I think

crates/evm/traces/src/decoder/mod.rs
mablr: yes good point

mablr (dismissed):

mablr (dismissed):
Please fix failed tests

jaimebarrancos:
> Please fix failed tests

I believe there is a problem regarding the new dynamic linking.

In the github actions CI, the diff adds:
├─ [0] VM::deployCode("src/Counter.sol:Counter")

However, when I test locally, there is no deployCode being called.

Im using the latest changes and am using MacOS.

crates/evm/traces/src/decoder/mod.rs
```diff
@@ -513,6 +520,29 @@ impl CallTraceDecoder {
...
+                                    "[{}] {} 0x{:x} → (0x{:x})",
+                                    step.gas_cost, opcode, change.key, change.value
+                                ),
+                                None => format!("[{}] {}", step.gas_cost, opcode),
+                            };
+
+                            step.decoded = Some(Box::new(DecodedTraceStep::Line(res)));
+                            step.storage_change = None; // show only step.decoded
```
mablr: Clearing `step.storage_change` here removes the state-diff data from the trace, not just from the duplicate text rendering. `render_trace_arena_inner(..., with_storage_changes = true)` later uses those same `storage_change` entries for things like `append_tempo_channel_storage_decodes`, so `--opcodes SSTORE` can suppress the higher-level storage decode block for selected writes. Consider preserving the trace data and deduping only in the text-rendering path, or rendering opcodes from a copy.

jaimebarrancos: Done! Is it acceptable to clone it every time it runs or should I add a flag `with_opcodes` to the `render_trace_arena_inner` function?

jaimebarrancos:
@mablr bump

figtracer (dismissed):
lgtm

crates/evm/traces/src/lib.rs
```diff
@@ -208,7 +208,17 @@ pub fn render_trace_arena_inner(
         return serde_json::to_string(&arena.resolve_arena()).expect("Failed to serialize traces");
     }
 
-    let resolved = arena.resolve_arena();
+    // Remove overlapping decoded trace data
```
DaniPopes: this is extremely expensive

crates/evm/traces/src/lib.rs
```diff
@@ -218,7 +228,7 @@ pub fn render_trace_arena_inner(
     let mut rendered =
         String::from_utf8(w.into_writer()).expect("trace writer wrote invalid UTF-8");
     if with_storage_changes {
-        append_tempo_channel_storage_decodes(&mut rendered, &resolved);
+        append_tempo_channel_storage_decodes(&mut rendered, &arena.resolve_arena());
```
grandizzy: The `TraceWriter` above renders `cloned_arena` (with `storage_change` nulled out on decoded steps), but this passes the **original** `arena.resolve_arena()`. So the appended Tempo storage-decode block is computed from storage changes that were intentionally removed from the rendered trace — the two can disagree, and `--opcodes SSTORE` can still surface decodes for writes that were suppressed above. Pass the same resolved arena the writer used.

crates/evm/traces/src/lib.rs
grandizzy: +1 to @DaniPopes — this clone runs on every non-JSON render, even when `--opcodes` is unused and `with_storage_changes` is false. At `-vvvvv` steps carry stack/memory/storage data, so it can roughly double peak memory and slow normal runs. Suggest resolving once and cloning lazily via `Cow::to_mut`, only when storage rendering is on and there's actually a decoded opcode line whose `storage_change` needs suppressing.

grandizzy (dismissed):
Thanks for the contribution! Please see inline for a couple of findings.

crates/evm/traces/src/lib.rs
jaimebarrancos: But I thought `append_tempo_channel_storage_decodes` needs the original trace data, like @mablr said:
> Clearing step.storage_change here removes the state-diff data from the trace, not just from the duplicate text rendering. render_trace_arena_inner(..., with_storage_changes = true) later uses those same storage_change entries for things like append_tempo_channel_storage_decodes, so --opcodes SSTORE can suppress the higher-level storage decode block for selected writes. Consider preserving the trace data and deduping only in the text-rendering path, or rendering opcodes from a copy.

If not, my previous solution would suffice:
```
step.decoded = Some(Box::new(DecodedTraceStep::Line(res)));
step.storage_change = None; // show only step.decoded
```

In short, we want to show the tempo storage decodes whether there is or not an opcode flag, right?

mablr: IMO Tempo decodes should use the original trace data; the dedupe should probably happen only in rendering, without clearing storage_change from the trace model?

jaimebarrancos: @grandizzy

grandizzy: 👍 makes sense

mablr (dismissed):
Moved `resolved.to_mut()` inside the `if with_storage_changes` block, so the clone only happens when we actually need to mutate the rendered copy.

Should be good to go. Thank you @jaimebarrancos.

crates/evm/traces/src/lib.rs
```diff
@@ -209,18 +209,32 @@ pub fn render_trace_arena_inner(
...
 
-    let resolved = arena.resolve_arena();
+    let mut resolved = arena.resolve_arena();
+
+    let mut tempo_changes = String::new();
+    if with_storage_changes {
+        append_tempo_channel_storage_decodes(&mut tempo_changes, &resolved);
+
+        // Remove storage text that is already represented by an opcode line.
+        for node in resolved.to_mut().nodes_mut() {
```
grandizzy: **Blocking:** `to_mut()` clones the entire `CallTraceArena` for any `-vvvvv` run, since `with_storage_changes` is just `verbosity > 4` and `resolve_arena()` is normally `Cow::Borrowed`. Only mutate when an opcode line actually duplicates a storage change:

```rust
let needs_dedup = resolved.as_ref().nodes().iter().any(|node| {
    node.trace.steps.iter().any(|step| {
        step.storage_change.is_some()
            && matches!(step.decoded.as_deref(), Some(DecodedTraceStep::Line(_)))
    })
});
if needs_dedup {
    for node in resolved.to_mut().nodes_mut() {
        for step in &mut node.trace.steps {
            if step.storage_change.is_some()
                && matches!(step.decoded.as_deref(), Some(DecodedTraceStep::Line(_)))
            {
                step.storage_change = None;
            }
        }
    }
}
```

crates/forge/src/cmd/test/mod.rs
```diff
@@ -586,6 +586,15 @@ pub struct TestArgs {
...
     pub rerun: bool,
 
+    /// Print the given opcodes in trace output, with their gas
+    /// cost and the storage slot and value, if available.
+    ///
+    /// Accepts a comma-separated list of opcode names, e.g.
+    /// `--opcodes SLOAD,MLOAD,SSTORE`. Names are in uppercase.
+    /// Requires `-vvvvv` to render.
+    #[arg(long, value_parser = parse_opcode, value_delimiter(','))]
+    pub opcodes: Vec<OpCode>,
```
grandizzy, line 813: `--opcodes` is silently ignored with `--json`, `--junit`, and `--list`. Consider `conflicts_with_all = ["json", "junit", "list"]` so users get a clear error.

grandizzy (dismissed):
One blocking issue: the storage de-dup clones the whole trace arena on **every** `-vvvvv` render, even when `--opcodes` isn't used. Fix inline below. One minor follow-up noted as well.

mablr (dismissed):

crates/evm/traces/src/lib.rs
```diff
@@ -209,18 +209,42 @@ pub fn render_trace_arena_inner(
         return serde_json::to_string(&arena.resolve_arena()).expect("Failed to serialize traces");
     }
 
-    let resolved = arena.resolve_arena();
+    let mut resolved = arena.resolve_arena();
+
+    let mut tempo_changes = String::new();
+    if with_storage_changes {
+        append_tempo_channel_storage_decodes(&mut tempo_changes, &resolved);
```
stevencartavia: `render_trace_arena_inner` now calls `append_tempo_channel_storage_decodes` on an empty `tempo_changes` buffer instead of the rendered output. Because the buffer starts empty, the `if !rendered.ends_with('\n')` guard always fires, adding an extra blank line before `Decoded TIP20ChannelReserve storage:` on every `-vvvvv` Tempo trace with channel-reserve storage changes (not covered by any test).

crates/forge/src/cmd/test/mod.rs
```diff
@@ -803,6 +803,15 @@ pub struct TestArgs {
     #[arg(long)]
     pub rerun: bool,
 
+    /// Print the given opcodes in trace output, with their gas
+    /// cost and the storage slot and value, if available.
+    ///
+    /// Accepts a comma-separated list of opcode names, e.g.
+    /// `--opcodes SLOAD,MLOAD,SSTORE`. Names are in uppercase.
+    /// Requires `-vvvvv` to render.
+    #[arg(long, value_parser = parse_opcode, value_delimiter(','), conflicts_with_all = ["json", "junit", "list"])]
```
stevencartavia: `--opcodes` forces `-vvvvv`, so `decoder.opcodes` is set and stored as `last_run_decoder`, which the `--debug` path reuses to re-decode the arenas handed to the debugger. Since `--opcodes` only conflicts with `json/junit/list` (and `--debug` doesn't list it), `forge test --debug --opcodes SLOAD -vvvvv` runs and the opcode `DecodedTraceStep::Line`s leak into the debugger: the TUI renders each matched step's line as a spurious step notice, and `set_step_notice` drops precompile-call notices for steps whose `Line` is already set (e.g. `--opcodes CALL`).

Confirmed via `--debug --dump`: with `--opcodes SLOAD` the dumped `debug_arena` steps contain `"Line":"[2100] SLOAD 0x1f → (…)"`; without it, none.

stevencartavia (approved):

mablr (approved):

PR #14844, opened 2026-05-20
feat:  Group invariant campaigns by test contract

0xKarl98:
On oss-272

Make the invariant test contract the campaign and corpus boundary for compatible boolean invariants. Per-predicate and per-handler failures are still reported individually, optimization invariants keep separate corpus state, and legacy failure caches remain readable with a warning. This also preserves counterexamples in JUnit output.


crates/forge/tests/cli/test_cmd/invariant/common.rs
```diff
@@ -1937,18 +1937,21 @@ contract InvariantShrinkWithAssert is Test {
...
- invariant_with_[..]() ([..])
+[FAIL: wrong counter require] invariant_with_require
+	[SEQUENCE]
 ...
-[FAIL: wrong counter [..]][..]
-	[Sequence] (original: 2, shrunk: 2)
+Suite assert_all: 2/2 invariants broken
 ...
- invariant_with_[..]() ([..])
+Suite predicates:
```
0xalpharush: I think we should just say "Assertion Tests" and "Invariant/Property Tests"

0xalpharush: Maybe add the invariant contract name as well?

0xalpharush (approved):

crates/forge/src/runner.rs
```diff
@@ -460,13 +489,56 @@ impl<'a, FEN: FoundryEvmNetwork> ContractRunner<'a, FEN> {
...
+            .filter(|func| func.is_invariant_test() && !is_optimization_invariant(func))
+            .collect::<Vec<_>>();
+        // Contract-local boolean invariants are eligible for a shared campaign only when all
+        // of them enable `assert_all` and share the same effective inline invariant config.
+        let boolean_invariant_fns = invariant_fns
+            .iter()
+            .copied()
+            .filter(|func| !is_optimization_invariant(func))
+            .collect::<Vec<_>>();
+        let merge_invariant_suite = self.config.invariant.assert_all
```
grandizzy: Is it intentional that `--mt` (and `--nmt`) no longer isolates a single predicate inside an `assert_all` contract? With this change, `forge test --mc Foo --mt invariant_a` still executes `invariant_b` and writes its persisted-failure cache, would be nice to keep the shared per-contract corpus but gate predicate evaluation on the filter, so `--mt` works as before.

crates/forge/src/runner.rs
```diff
@@ -460,13 +489,56 @@ impl<'a, FEN: FoundryEvmNetwork> ContractRunner<'a, FEN> {
...
+            && !matched_boolean_invariant_fns.is_empty()
+            && boolean_invariant_fns.len() > 1
+            && boolean_invariant_fns.iter().all(|func| self.invariant_assert_all_enabled(func))
+            && self.invariant_suite_configs_match(&boolean_invariant_fns);
+        // Use the first matched boolean invariant as the campaign anchor when the suite merges.
+        let invariant_suite_anchor =
+            merge_invariant_suite.then(|| matched_boolean_invariant_fns.first().copied()).flatten();
```
grandizzy: Should the merged `TestResult` be keyed off whichever predicate the filter happened to match first? Right now `invariant_suite_anchor = matched_boolean_invariant_fns.first()`, so the same merged campaign appears under `invariant_a()` or `invariant_b()` depending on `--mt`, and `primary_failure_file` lands in different paths.

grandizzy (changes requested):
thanks, that goes in the right direction! I think the target should be to get rid of `assert_all` altogether as now we assume campaign by test contract, hence redundant. Would also be nice to preserve the behavior of `--mt` to give the possibility of running single test & clean output for dev cycle (nice to have, not 100% sure if still possible?)

0xKarl98:
@ptal again , have addressed your concerns

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1335,6 +1342,15 @@ fn read_corpus_dir(path: &Path) -> impl Iterator<Item = CorpusDirEntry> {
     .into_iter()
 }
 
+fn has_legacy_invariant_corpus_dirs(path: &Path) -> bool {
+    std::fs::read_dir(path).is_ok_and(|entries| {
+        entries.flatten().any(|entry| {
+            entry.path().is_dir()
```
grandizzy: nit: This flags every non-worker subdir as legacy, but with this PR optimization invariants legitimately store their corpus under `<corpus_root>/<Contract>/<invariant_name>/ `, so a sibling of the boolean campaign's `worker*` dirs. On a contract that mixes boolean and optimization invariants the boolean run will spuriously warn about "legacy" dirs that are actually current optimization data. Would it make sense to also skip subdirs that contain `optimization_best.json` or their own `worker*` subdir?

crates/forge/src/result.rs
```diff
@@ -655,67 +683,125 @@ impl fmt::Display for TestResult {
...
+    fn write_handler_failures(&self, s: &mut String, user_facing: bool, preceded: bool) {
+        if self.invariant_handler_failures.is_empty() {
+            return;
+        }
+
+        let prefix = if preceded { "\n" } else { "" };
+        writeln!(
+            s,
+            "{prefix}{}: {} assertion bug(s) found",
+            if user_facing { "Assertion Tests" } else { "Suite handlers" },
```
grandizzy:
```suggestion
            if user_facing { "Assertion Tests" } else { "Handler assertions" },
```

crates/forge/src/result.rs
```diff
@@ -655,67 +683,125 @@ impl fmt::Display for TestResult {
...
+        user_facing: bool,
+        show_header: bool,
+    ) {
+        if self.invariant_predicate_results.len() <= 1 {
+            return;
+        }
+
+        if show_header {
+            s.push('\n');
+            s.push_str(if user_facing { "Invariant/Property Tests" } else { "Suite predicates" });
```
grandizzy:
```suggestion
            s.push_str(if user_facing { "Invariant/Property Tests" } else { "Predicates" });
```

crates/forge/src/result.rs
```diff
@@ -655,67 +683,125 @@ impl fmt::Display for TestResult {
...
+            return false;
+        };
+        if total <= 1 || !is_invariant_failure {
+            return false;
+        }
+
+        writeln!(
+            s,
+            "\n{}: {}/{total} invariants broken",
+            if user_facing { "Invariant/Property Tests" } else { "Suite predicates" },
```
grandizzy:
```suggestion
            if user_facing { "Invariant/Property Tests" } else { "Predicates" },
```

grandizzy (dismissed):
looks good, only some nits, pls check

crates/evm/evm/src/executors/corpus.rs
0xKarl98: Have added  && !path.join(OPTIMIZATION_BEST_FILE).is_file() , to prevent function-level optimization corpus from being judged to be legacy

0xKarl98:
@grandizzy  Have applied your nits ,  ptal again

grandizzy:
> @grandizzy Have applied your nits , ptal again

@0xKarl98 yep, gtg from my side, ty. Let's fix ci

crates/forge/src/runner.rs
```diff
@@ -797,6 +864,7 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
         let runner = self.invariant_runner();
         let invariant_config = self.config.invariant.clone();
```
decofe, line 852: The skip preflight just above this only checks the campaign anchor (`func`). With contract-level invariant campaigns, secondary predicates can also call `vm.skip`; those are now swallowed because we return before building per-predicate results. This shows up in `test_cmd::skip_output`: `invariant_skipInvariantReason()` disappears and the suite drops from 6 skipped to 5.

Suggested fix: preflight every predicate in `invariants`, collect skipped predicates as `InvariantPredicateResult { status: TestStatus::Skipped, reason }`, render skip reasons in `write_invariant_predicate_results` (`[SKIP: reason] name`), and exclude skipped predicates from the fuzz campaign. If all predicates skip, return a skipped campaign result carrying those predicate results.

crates/forge/tests/cli/test_cmd/invariant/mod.rs
```diff
@@ -1717,36 +2031,85 @@ contract StaleSecondaryTest is Test {
...
+[FAIL: second broken] invariant_second
+	[Sequence] (original: 3, shrunk: 3)
+		sender=[..] addr=[src/Counter.sol:Counter]0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f calldata=inc() args=[]
+		sender=[..] addr=[src/Counter.sol:Counter]0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f calldata=inc() args=[]
+		sender=[..] addr=[src/Counter.sol:Counter]0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f calldata=inc() args=[]
+
+Invariant/Property Tests: 2/2 invariants broken
+[FAIL: first broken] invariant_first
+[FAIL: second broken] invariant_second
+2 invariant failure(s) persisted to /private/var/folders/p4/rl3tc43j51s311jt6pq7zq180000gn/T/[..]/cache/invariant/failures/StaleSecondaryTest — rerun to shrink
```
decofe: This hard-codes a macOS temp root, so the snapshot fails on Linux CI where the same logical path is under `/tmp/<test-dir>/cache/invariant/failures/StaleSecondaryTest`. Please redact the temp project root here, e.g. expect `[..]/cache/invariant/failures/StaleSecondaryTest — rerun to shrink` (and same below), so the assertion checks the invariant failure subpath without depending on host temp-dir layout.

0xKarl98:
Have addressed your nits , ptal again @decofe

mattsse (dismissed):
Reviewed locally. The invariant campaign grouping looks acceptable to me, and `cargo check -p forge --tests` passed.

grandizzy (dismissed):

grandizzy (approved):

mattsse (approved):

crates/forge/src/runner.rs
```diff
@@ -894,22 +960,21 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
-        // and — under `assert_all` — every other invariant that doesn't already have a
-        // compatible persisted failure. Track the anchor's index so downstream consumers can
-        // resolve the campaign-anchor without searching by name.
-        let invariant_fns: Vec<(&Function, bool)> = invariants
+        // and every other selected predicate that doesn't already have a compatible persisted
+        // failure. Track the anchor's index so downstream consumers can resolve the campaign
+        // anchor without searching by name.
+        let invariant_fns: Vec<(&Function, bool)> = live_invariants
             .into_iter()
             .filter(|(invariant_fn, _)| {
```
figtracer, line 968: This drops selected secondary predicates with compatible persisted failures from the campaign without adding them to the result. I reproduced this: first broad `--mt invariant_` persists a secondary failure, then the same broad run exits 0 and reports only the passing anchor. Persisted selected failures should still fail/report, not just warn.

crates/forge/src/runner.rs
```diff
@@ -253,6 +255,22 @@ impl<'a, FEN: FoundryEvmNetwork> ContractRunner<'a, FEN> {
         Ok(config)
     }
 
+    /// Returns true if all invariant functions share the same effective inline invariant config.
+    fn invariant_suite_configs_match(&self, funcs: &[&Function]) -> bool {
+        let Some((anchor, rest)) = funcs.split_first() else {
+            return true;
+        };
+        let anchor_config = match self.inline_config(Some(anchor)) {
```
figtracer, line 263: `invariant_suite_configs_match` only compares `config.invariant`, but merged campaigns run under the anchor’s executor config. I reproduced a secondary with `forge-config: default.evm_version = "shanghai"`: it passes alone, but fails when merged under the anchor’s Cancun config. Please include executor-affecting inline config in compatibility, or don’t merge when function-level config differs.

PR #14845, opened 2026-05-20
fix(lint): preserve variable locations in loop lints

figtracer:
## Summary
- preserve HIR variable data locations when reconstructing semantic types in calls-loop and delegatecall-loop
- add a calls-loop regression case for storage array builtins inside loops
- prevents Solar member resolution from receiving bare dynamic array reference types and panicking

Context: seen in tempoxyz/tempo CI where forge build --sizes on nightly crashed in calls-loop with `native_members: type DynArray(Elementary(Address)) should be wrapped in Ref`.

### DIFF
--- a/crates/lint/src/sol/low/calls_loop.rs
+++ b/crates/lint/src/sol/low/calls_loop.rs
@@ -4,7 +4,7 @@ use crate::{
     sol::{Severity, SolLint},
 };
 use solar::{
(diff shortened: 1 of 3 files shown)


zerosnacks (dismissed):
Makes sense

zerosnacks (approved):

mablr (approved):

grandizzy (approved):

PR #14846, opened 2026-05-20
feat(symbolic): add portfolio solver diagnostics

figtracer:
## Summary
- Share the semantic symbolic solver registry between the symbolic runtime and Forge warning logic.
- Add per-query portfolio diagnostics when `--symbolic-dump-smt` is enabled.
- Warn when a configured solver portfolio is degraded because one or more entries are unavailable, while still using the invokable entries.

## Example output

`--symbolic-dump-smt` now includes the portfolio outcome for each query:

```text
--- symbolic solver portfolio outcomes ---
yices-smt2 --bvconst-in-decimal: sat-valid in 2.033ms winner
z3 -in -smt2: cancelled in 4.430ms
```

A partially unavailable portfolio now warns but still runs with available solvers:

```text
Warning: Symbolic solver portfolio is degraded; unavailable entries: `foundry-missing-symbolic-solver -in -smt2` (failed to execute `foundry-missing-symbolic-solver`: No such file or directory (os error 2)). Available portfolio entries will still be used.
```

Ran live z3/yices/cvc5/bitwuzla portfolio smoke tests against `/private/tmp/symexec-solver-bench-pr14835`; no `sat-invalid` models, solver errors, or serial query-fingerprint contradictions were observed.

### DIFF
--- a/crates/evm/symbolic/README.md
+++ b/crates/evm/symbolic/README.md
@@ -225,7 +225,8 @@ whitespace, quotes, or backslashes are split into argv parts like
 For latency-sensitive local runs, start with a small portfolio such as
 `["yices", "z3"]`. Broader portfolios can help on solver-diverse workloads but
 use more CPU and can be slower when one fast solver already handles most
-queries.
+queries. `--symbolic-dump-smt` also prints per-query portfolio outcomes so solver
+mixes can be compared without changing execution semantics.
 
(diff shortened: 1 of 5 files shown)


grandizzy (commented):
lgtm! could use one liner doc on all newly introduced fns

grandizzy (approved):

PR #14848, opened 2026-05-20
fix(mutation): honor filters and isolation

zerosnacks:
Stacked on #14971 towards OSS-1

- Honor Forge test filters and isolation settings during mutation runs.
- Reject mutation mode with incompatible test flags instead of silently producing misleading results.
- Preserve #14971's execution-aware result/survived cache keys while layering the remaining runtime correctness fixes on top.
- Improve adaptive/resume behavior by loading survived spans before mutant generation.
- Avoid invalid/duplicate mutations for compound assignments and tighten mutator tests.
- Improve reporting/runner plumbing for skipped, timed out, and resumed mutants.


mablr (dismissed):
lgtm, pending another approval...

crates/forge/src/cmd/test/mod.rs
```diff
@@ -644,6 +677,19 @@ impl TestArgs {
                 eyre::bail!("Cannot run mutation testing with failed tests");
             }
 
+            // A green baseline that ran *zero* tests is not a useful baseline:
+            // every compileable mutant would be reported as `Alive` (no test
+            // failed, so nothing killed it), which produces a wildly
+            // misleading mutation report. Hard-error so users get an actual
+            // signal that their filter / path / setup matched nothing.
+            if outcome.tests().next().is_none() {
```
stevencartavia: **Zero-baseline guard misses all-skipped baselines**

`outcome.tests()` yields passed + failed + skipped, so a suite where `setUp()` calls `vm.skip(true)` passes this check and runs mutation anyway. The mutation runner only marks a mutant killed on test *failure* (`runner.rs`: `any(|suite| suite.failed() > 0)`), so a skipped baseline produces exactly the misleading "every compileable mutant is Alive" report this guard is trying to prevent.

Reproduced on a `forge init` template whose `setUp` calls `vm.skip(true)`:
```
Baseline: 0 passed; 0 failed; 1 skipped
forge test --mutate src/Counter.sol
→ Mutation Score: 0.0% (0/5 mutants killed); exit 0
```

Suggested fix — require at least one passing baseline (the `failed() > 0` bail just above already covers failures, so `successes()` cleanly expresses the remaining requirement):
```rust
if outcome.successes().next().is_none() {
    eyre::bail!(
        "Mutation testing requires at least one passing baseline test; the current \
         filter/path selection matched zero non-skipped tests. Loosen `--match-test` / \
         `--match-contract` / `--match-path` or check the project layout."
    );
}
```

crates/forge/src/mutation/mutators/binary_op_mutator.rs
```diff
@@ -62,8 +65,11 @@ impl Mutator for BinaryOpMutator {
             .into_iter()
             .filter(|&kind| kind != op)
             .map(|kind| {
-                // Build mutated expression: "lhs new_op rhs"
-                let mutated_expr = format!("{} {} {}", lhs_text, kind.to_str(), rhs_text);
+                let mutated_expr = if compound_assignment {
+                    format!("{} {}= {}", lhs_text, kind.to_str(), rhs_text)
```
stevencartavia, line 70: **Compound assignment generates `**=` and `>>>=`, which aren't valid Solidity**

This branch formats every `operations_others` kind as `lhs op= rhs`, but that set includes `BinOpKind::Pow` and `BinOpKind::Sar` — and Solidity has neither `**=` nor `>>>=`. Reproduced end-to-end on `number += d`:
```
[ 3/13] Testing mutant: `number += d` → `number **= d`
[10/13] Testing mutant: `number += d` → `number >>>= d`
```
Both are rejected by solc (`Expected primary expression.` / `Operator >>>= not compatible with types uint256 and uint256.`) and always end up in the `Invalid` bucket. Score isn't corrupted (Invalid is excluded from the denominator), but it costs a compile per mutant and adds predictable noise to reports.

Suggested fix — explicit allowlist of valid Solidity compound ops (don't reuse Solar's `BinOpKind::assignable()`; it includes `Sar`):
```rust
fn is_valid_compound_assignment_op(kind: BinOpKind) -> bool {
    matches!(kind,
        BinOpKind::BitOr | BinOpKind::BitXor | BinOpKind::BitAnd
        | BinOpKind::Shl | BinOpKind::Shr
        | BinOpKind::Add | BinOpKind::Sub | BinOpKind::Mul | BinOpKind::Div | BinOpKind::Rem
    )
}
```
then in the replacement map:
```rust
.filter(|&kind| kind != op)
.filter(|&kind| !compound_assignment || is_valid_compound_assignment_op(kind))
```

crates/forge/src/mutation/mod.rs
```diff
@@ -146,16 +146,33 @@ impl MutationsSummary {
         if valid_mutants == 0 { 0.0 } else { self.dead.len() as f64 / valid_mutants as f64 * 100.0 }
     }
 
-    /// Convert to JSON output format
+    /// Convert to JSON output format.
+    ///
+    /// Output is sorted deterministically: files in lexicographic order
+    /// (`BTreeMap` keys), and survived mutants within each file sorted by
+    /// `(span.lo, span.hi, mutation_text)`. Without this, parallel worker
```
stevencartavia: Nit: doc comment doesn't match the actual sort key. Says survivors are sorted by `(span.lo, span.hi, mutation_text)`, but the implementation sorts by `(line, column, original, mutant)`. JSON shape is still deterministic (it doesn't carry span fields), so this is doc-only — either update the comment to match, or switch the sort to actually key on span.

stevencartavia (commented):
three things worth a look. All reproduced end-to-end against this branch (`pr-14848` @ `79abd60e7`).

crates/forge/src/mutation/mutators/binary_op_mutator.rs
mablr: done https://github.com/foundry-rs/foundry/pull/14848/commits/fc7fce2b594cb539851b4eb40c7609725efa1deb

grandizzy (changes requested):
**Blocking — per-mutant compilation still ignores the baseline-selected file set.**

In `crates/forge/src/mutation/runner.rs`, `compile_and_test_inner` (around L612):

```rust
let compiler =
    ProjectCompiler::new().dynamic_test_linking(config.dynamic_test_linking).quiet(true);

let compile_output = compiler.compile(&config.project()?)?;
...
let filter = filter_args.clone().merge_with_config(config);
...
runner.test_collect(&filter)
```

Execution is now filter-aware, but compilation is not: each mutant compiles every test file in the temp workspace. A filtered-out test that fails to compile against the mutated source (or is otherwise uncompilable) will mark the mutant `Invalid` and diverge from the baseline filtered `forge test` — which already excludes such files via `TestArgs::get_sources_to_compile`.

### Suggested shape

Mirror what baseline `compile_and_run` does. Compute the selected set once, carry it through `MutationRunConfig`, and re-anchor it onto each temp workspace.

**1. In `TestArgs::run_tests`, before constructing `MutationRunConfig`:**

```rust
let selected_sources_relative = self
    .get_sources_to_compile(&config_for_mutation, &filter)?
    .into_iter()
    .filter_map(|p| p.strip_prefix(&config_for_mutation.root).ok().map(PathBuf::from))
    .collect::<Vec<_>>();
```

Add `selected_sources_relative: Vec<PathBuf>` to `MutationRunConfig`.

**2. Thread through** `run_mutations_parallel_with_progress` → `test_single_mutant_isolated` → `compile_and_test` next to `filter_args` (same `Arc<…>` pattern that's already there).

**3. In `compile_and_test_inner`, re-anchor onto the temp `config.root` and pass to `.files(...)`:**

```rust
let files: Vec<PathBuf> = selected_sources_relative
    .iter()
    .map(|rel| config.root.join(rel))
    .filter(|p| p.exists())
    .collect();

let compile_output = ProjectCompiler::new()
    .dynamic_test_linking(config.dynamic_test_linking)
    .quiet(true)
    .files(files)
    .compile(&config.project()?)?;
```

### Why this is the right shape

- **Single source of truth.** `get_sources_to_compile` is already the canonical filter-aware compile-set helper used by `forge test`. Reusing it guarantees baseline and mutants stay in lockstep — no second implementation to drift.
- **Computed once.** The (expensive) ABI compile inside `get_sources_to_compile` runs once on the original project, not per mutant.
- **Path-stable.** Stripping to project-relative and re-joining onto the temp root matches the existing `src_rel`/`test_rel`/`script_rel` pattern in `test_single_mutant_isolated`.
- **Tolerant of missing files.** `.filter(|p| p.exists())` mirrors how the PR already handles `fs_permissions` paths.
- **Empty-filter fast path preserved.** `get_sources_to_compile` returns the full `src + test` set when the filter is empty, so unfiltered mutation runs are unchanged.

### Regression test

Add to `crates/forge/tests/cli/test_cmd/mutation.rs`. Without the fix every mutant is `Invalid`; with the fix the broken file is excluded from per-mutant compilation just like the baseline.

```rust
forgetest_init!(mutation_honors_match_path_at_compile_time, |prj, cmd| {
    // Source with a mutable arithmetic operator → guarantees ≥1 mutant
    // (e.g. `+` → `-`, `+` → `*`) that the selected test will kill.
    prj.add_source(
        "Foo.sol",
        r#"
pragma solidity ^0.8.13;

contract Foo {
    function add(uint256 a, uint256 b) public pure returns (uint256) {
        return a + b;
    }
}
"#,
    );

    // Selected test: compiles and passes against the original, kills the
    // arithmetic mutants.
    prj.add_test(
        "FooSelected.t.sol",
        r#"
pragma solidity ^0.8.13;

import "../src/Foo.sol";

contract FooSelectedTest {
    Foo internal foo;

    function setUp() public {
        foo = new Foo();
    }

    function test_Add() public view {
        assert(foo.add(2, 3) == 5);
    }
}
"#,
    );

    // Unselected test: deliberately fails to compile (undefined symbol).
    // Baseline `forge test --match-path FooSelected.t.sol` succeeds because
    // `get_sources_to_compile` excludes this file from the compile set.
    // Mutation runs must behave the same — otherwise every mutant gets
    // misclassified as `Invalid` due to a test the user explicitly excluded.
    prj.add_test(
        "FooBroken.t.sol",
        r#"
pragma solidity ^0.8.13;

import "../src/Foo.sol";

contract FooBrokenTest {
    function test_Broken() public pure {
        // `NonExistent` is undefined → this file cannot compile.
        NonExistent x = NonExistent(0);
        x.doSomething();
    }
}
"#,
    );

    // Sanity check: baseline filtered `forge test` succeeds despite
    // FooBroken.t.sol being uncompilable. If this fails, the test scaffold
    // itself is wrong (not the mutation code under test).
    cmd.forge_fuse()
        .args(["test", "--match-path", "test/FooSelected.t.sol"])
        .assert_success();

    // Mutation run with the same path filter must also succeed and must
    // NOT mark mutants as `Invalid` due to FooBroken.t.sol failing to
    // compile against the mutated source.
    cmd.forge_fuse().args([
        "test",
        "--mutate", "src/Foo.sol",
        "--match-path", "test/FooSelected.t.sol",
        "--mutation-jobs", "1",
        "--json",
    ]);

    let out = cmd.assert_success().get_output().stdout_lossy();
    let summary = mutation_summary(&out);

    let invalid = summary["invalid"].as_u64().unwrap_or(u64::MAX);
    let killed  = summary["killed"].as_u64().unwrap_or(0);
    let alive   = summary["survived"].as_u64().unwrap_or(0);

    assert_eq!(
        invalid, 0,
        "filtered-out FooBroken.t.sol must not cause Invalid mutants; summary={summary}"
    );
    assert!(
        killed + alive >= 1,
        "expected at least one Killed/Alive mutant from arithmetic ops; summary={summary}"
    );
});
```

Worth considering as follow-up variants in the same file: `mutation_honors_match_test_at_compile_time` (use `--match-test test_Add`) and `mutation_honors_no_match_path` (invert via `--no-match-path 'test/FooBroken.t.sol'`).

grandizzy (approved):
lgtm!

PR #14849, opened 2026-05-20
feat(symbolic): summarize portfolio diagnostics

figtracer:
## Summary

- add an aggregate `--symbolic-dump-smt` portfolio summary for staged portfolios
- report query count, solver launches, rescue launches/wins, skipped launches, cancellation/error/model-invalid counts, and per-solver winner/launch/outcome totals
- keep solver execution semantics unchanged; this only makes the staged policy easier to measure
- document the summary output and add focused regression coverage for the aggregation logic

## Example

```text
--- symbolic solver portfolio summary ---
queries: 48
solver runs: 64
rescue solver runs: 0
not-started solver runs: 32
non-primary wins: 0
rescue wins: 0
cancelled after winner: 0
invalid models: 0
solver errors: 0
winner counts:
  yices-smt2 --bvconst-in-decimal: 48
launch counts:
  yices-smt2 --bvconst-in-decimal: 48
  z3 -in -smt2: 16
outcome counts:
  not-started: 32
  sat-valid: 32
  unsat: 32
```

### DIFF
--- a/crates/evm/symbolic/README.md
+++ b/crates/evm/symbolic/README.md
@@ -231,7 +231,31 @@ For latency-sensitive local runs, start with a small portfolio such as
 can still use more CPU and be slower when one fast solver already handles most
 queries. `--symbolic-dump-smt` prints each solver's configured order and launch
 delay with the per-query portfolio outcomes so solver mixes can be compared
-without changing execution semantics.
+without changing execution semantics. It also prints an aggregate portfolio
+summary at the end of the run, for example:
+
+```text
+--- symbolic solver portfolio summary ---
+queries: 48
+solver runs: 64
+rescue solver runs: 0
+not-started solver runs: 32
+non-primary wins: 0
+rescue wins: 0
+cancelled after winner: 0
+invalid models: 0
+solver errors: 0
+winner counts:
+  yices-smt2 --bvconst-in-decimal: 48
+launch counts:
+  yices-smt2 --bvconst-in-decimal: 48
+  z3 -in -smt2: 16
+outcome counts:
+  not-started: 32
+  sat-valid: 32
+  unsat: 32
+```
+
 Forge warns when a configured portfolio is degraded because one or more solver
 entries are not available, but it still uses the entries that can be invoked.
 
--- a/crates/evm/symbolic/src/runtime/solver.rs
+++ b/crates/evm/symbolic/src/runtime/solver.rs
@@ -5,18 +5,47 @@ const MAX_SOLVER_POLL_BACKOFF: Duration = Duration::from_millis(50);
 const SECOND_PORTFOLIO_SOLVER_DELAY: Duration = Duration::from_millis(100);
 const RESCUE_PORTFOLIO_SOLVER_DELAY: Duration = Duration::from_millis(500);
 
-const OUTCOME_CANCELLED: &str = "cancelled";
-const OUTCOME_ERROR: &str = "error";
-const OUTCOME_NOT_STARTED: &str = "not-started";
-const OUTCOME_SAT_AFTER_WINNER: &str = "sat-after-winner";
-const OUTCOME_SAT_INVALID: &str = "sat-invalid";
-const OUTCOME_SAT_VALID: &str = "sat-valid";
-const OUTCOME_TIMEOUT_OR_UNKNOWN: &str = "timeout-or-unknown";
-const OUTCOME_UNKNOWN: &str = "unknown";
-const OUTCOME_UNKNOWN_AFTER_WINNER: &str = "unknown-after-winner";
-const OUTCOME_UNSAT: &str = "unsat";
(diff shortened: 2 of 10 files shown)


crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -402,32 +425,137 @@ impl SolverRunSummary {
...
+        for (outcome, count) in &self.outcome_counts {
+            let _ = writeln!(stderr, "  {outcome}: {count}");
+        }
+    }
+}
+
+impl Drop for SmtLibSubprocessSolver {
+    fn drop(&mut self) {
+        if self.dump_smt {
+            self.portfolio_diagnostics.dump();
```
grandizzy: `SmtLibSubprocessSolver` is owned per `SymbolicExecutor`, and a fresh executor is created per symbolic test, so with the current Drop, a forge test run with N symbolic tests will print N "portfolio summary" blocks (one per executor drop), interleaved through the test output, instead of the single aggregate (as in README).
Suggest dropping the Drop impl, turning `PortfolioDiagnostics::dump` into `write_to(&mut impl Write)`, and adding a `merge(&Self)` so the run owner can sum per-executor diagnostics and emit once at the end of the run.

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -402,32 +425,137 @@ impl SolverRunSummary {
...
+            *self.outcome_counts.entry(summary.outcome).or_default() += 1;
+            if summary.started_after.is_some() {
+                self.solver_runs += 1;
+                *self.launch_counts.entry(summary.display.clone()).or_default() += 1;
+                if summary.index.is_some_and(|index| index >= 2) {
+                    self.rescue_runs += 1;
+                }
+            }
+
+            match summary.outcome {
```
grandizzy, line 519: let's make these outcomes constants (a `SolverOutcome` enum could be too much for this purpose)?

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -180,13 +188,15 @@ impl SmtLibSubprocessSolver {
             constraint.collect_vars(&mut vars);
         }
 
-        let commands = self.commands()?;
```
grandizzy, line 183: think we could keep it here instead repeating twice below

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -402,32 +425,137 @@ impl SolverRunSummary {
...
+    }
+
+    /// Writes the aggregate portfolio summary to stderr.
+    fn dump(&self) {
+        if self.queries == 0 {
+            return;
+        }
+
+        let mut stderr = std::io::stderr().lock();
+        let _ = writeln!(stderr, "--- symbolic solver portfolio summary ---");
```
grandizzy: let's `impl fmt::Display for PortfolioDiagnostics` and move it there

crates/evm/symbolic/README.md
```diff
@@ -231,7 +231,31 @@ For latency-sensitive local runs, start with a small portfolio such as
 can still use more CPU and be slower when one fast solver already handles most
 queries. `--symbolic-dump-smt` prints each solver's configured order and launch
 delay with the per-query portfolio outcomes so solver mixes can be compared
-without changing execution semantics.
+without changing execution semantics. It also prints an aggregate portfolio
+summary at the end of the run, for example:
+
+```text
+--- symbolic solver portfolio summary ---
```
grandizzy, line 238: smol reminder - we need follow-ups for `--show-progress` output

grandizzy (changes requested):
looks good overall, left some comment, ptal!

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -5,6 +5,19 @@ const MAX_SOLVER_POLL_BACKOFF: Duration = Duration::from_millis(50);
...
+const OUTCOME_NOT_STARTED: &str = "not-started";
+const OUTCOME_SAT_AFTER_WINNER: &str = "sat-after-winner";
+const OUTCOME_SAT_INVALID: &str = "sat-invalid";
+const OUTCOME_SAT_VALID: &str = "sat-valid";
+const OUTCOME_TIMEOUT_OR_UNKNOWN: &str = "timeout-or-unknown";
+const OUTCOME_UNKNOWN: &str = "unknown";
+const OUTCOME_UNKNOWN_AFTER_WINNER: &str = "unknown-after-winner";
+const OUTCOME_UNSAT: &str = "unsat";
+const OUTCOME_UNSAT_AFTER_WINNER: &str = "unsat-after-winner";
+const OUTCOME_UNEXPECTED: &str = "unexpected";
```
figtracer: maybe the enum would look better here

mablr: Definitely 💯

grandizzy (approved):
lgtm!

PR #14850, opened 2026-05-20
fix(forge): default Tempo tests to latest hardfork

figtracer:
## Summary
- default Tempo EVM-version fallback to the latest active Tempo hardfork instead of T0
- include Tempo in chain/timestamp hardfork detection now that the pinned dependency exposes it
- add config and forge CLI regressions for network=tempo / --network tempo

Manual gas check: a temporary 8 fresh-slot storage write test now reports 179,552 gas by default, 2,020,352 gas with --network tempo, and 179,552 gas with --network tempo plus FOUNDRY_HARDFORK=tempo:T0.

### DIFF
--- a/crates/evm/hardforks/src/lib.rs
+++ b/crates/evm/hardforks/src/lib.rs
@@ -395,7 +395,10 @@ pub fn evm_spec_id<SPEC: FromEvmVersion>(evm_version: EvmVersion) -> SPEC {
 /// Returns the latest Tempo hardfork that has an activation on a known Tempo network.
 pub fn latest_active_tempo_hardfork() -> TempoHardfork {
     // Tempo currently publishes activation timestamps through chain-aware hardfork resolution.
-    let now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap_or_default().as_secs();
+    let now = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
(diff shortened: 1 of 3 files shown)


mablr (dismissed):

crates/evm/hardforks/src/lib.rs
```diff
@@ -391,6 +389,15 @@ pub fn evm_spec_id<SPEC: FromEvmVersion>(evm_version: EvmVersion) -> SPEC {
     SPEC::from_evm_version(evm_version)
 }
 
+/// Returns the latest Tempo hardfork that has an activation on a known Tempo network.
+pub fn latest_active_tempo_hardfork() -> TempoHardfork {
+    // Tempo currently publishes activation timestamps through chain-aware hardfork resolution.
+    // Use `u64::MAX` to select the latest scheduled fork while ignoring placeholder variants.
+    TempoHardfork::from_chain_and_timestamp(4217, u64::MAX)
+        .or_else(|| TempoHardfork::from_chain_and_timestamp(42431, u64::MAX))
```
mattsse: can we use the current systemtime here, otherwise this could end up selecting unactivated hf

mattsse (changes requested):
smol suggestion

stevencartavia (dismissed):

PR #14857, opened 2026-05-21
chore: simplify negated is_some_and checks

solanaXpeter:
## Motivation

Recent clippy cleanup fixed the same `nonminimal_bool` warning in `calls_loop`. A few production code paths still use `!Option::is_some_and(...)`, which clippy reports as a non-minimal boolean expression when warnings are denied.

## Solution

Replace those checks with the equivalent `Option::is_none_or(...)` form in lint filtering, EVM backend state refresh, and verify-bytecode ignore handling.


figtracer (approved):

mablr (approved):
Not a big deal, but it's clearer, ty.

PR #14860, opened 2026-05-21
fix(forge): lower gas in out-of-gas script tests to avoid CI timeouts

mablr:
## Motivation

I noticed `assert_exit_code_error_on_out_of_gas_script` and `assert_exit_code_error_on_out_of_gas_script_with_json` sometimes timeout.

They just trigger the out of gas error with script using a `while(true)` loop.

## Solution

Lower block gas limit to make them terminate in ms.


figtracer (approved):

stevencartavia (approved):

mattsse (approved):
LGTM

PR #14861, opened 2026-05-21
fix(symbolic): defer diagnostics during progress output

figtracer:
## Summary
- capture verbose symbolic SMT dumps / per-query portfolio diagnostics while `--show-progress` is active
- render captured diagnostics with each test result after progress output is cleared
- keep live stderr diagnostics unchanged when progress is not active

### DIFF
--- a/crates/forge/src/cmd/test/mod.rs
+++ b/crates/forge/src/cmd/test/mod.rs
@@ -896,6 +896,13 @@ impl TestArgs {
                     !self.suppress_successful_traces || result.status == TestStatus::Failure;
                 if !silent {
                     sh_println!("{}", result.short_result(name))?;
+                    if let Some(diagnostics) = result
+                        .symbolic_diagnostics
+                        .take()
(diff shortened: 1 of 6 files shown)


crates/forge/src/cmd/test/mod.rs
```diff
@@ -896,6 +896,13 @@ impl TestArgs {
                     !self.suppress_successful_traces || result.status == TestStatus::Failure;
                 if !silent {
                     sh_println!("{}", result.short_result(name))?;
+                    if let Some(diagnostics) = result
+                        .symbolic_diagnostics
+                        .as_deref()
```
grandizzy: Minor: `result.symbolic_diagnostics.as_deref()` borrows but doesn't consume, so the buffered string stays in `outcome.results` for the rest of the run. The field is `#[serde(skip)]` and only read here, so swapping `as_deref()` → `take()` frees each test's buffer right after printing — relevant when `dump_smt` is on for a large suite.

grandizzy (approved):

PR #14863, opened 2026-05-21
clippy: remove remaining needless for_each calls

solanaXpeter:
## Motivation

The workspace already warns on `clippy::needless_for_each`, and recent clippy cleanup replaced side-effecting `.for_each(...)` calls with ordinary `for` loops. A few production paths still used `.for_each(...)` for side effects.

## Solution

Replace the remaining production side-effect `.for_each(...)` calls in Anvil state serialization, selector import reporting, and default script library deployment with equivalent `for` loops.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/storage.rs
+++ b/crates/anvil/src/eth/backend/mem/storage.rs
@@ -235,11 +235,11 @@ impl InMemoryBlockStates {
             .collect::<Vec<_>>();
 
         // Get on-disk state snapshots
-        self.on_disk_states.iter().for_each(|(hash, _)| {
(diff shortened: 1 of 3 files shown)


mablr (approved):
Thank you.

Weird it wasn't already flagged/enforced by clippy, have you figured out why ? @solanaXpeter

mattsse (approved):
+.

solanaXpeter:
@mablr The affected calls all predate the `needless_for_each` lint being enabled in #14301, and that cleanup was manual, so these look like leftovers that either were not matched by Clippy at the time or were simply missed in that pass.

PR #14869, opened 2026-05-21
feat(symbolic): configure exploration order

figtracer:
Adds a small `symbolic.exploration_order` config knob for pending symbolic path selection:

- `bfs` keeps today's FIFO behavior and remains the default
- `dfs` explores the newest queued path first

This only changes the local pending-path frontier order. It does not change solver portfolios, query limits, path limits, pruning, or invariant sequence-depth scheduling.

### DIFF
--- a/crates/evm/symbolic/src/executor.rs
+++ b/crates/evm/symbolic/src/executor.rs
@@ -1,5 +1,16 @@
 use super::{abi::*, runtime::*, *};
 
+/// Pops the next pending path according to the configured exploration order.
+pub(crate) fn pop_worklist<T>(
+    worklist: &mut VecDeque<T>,
+    order: SymbolicExplorationOrder,
+) -> Option<T> {
+    match order {
+        SymbolicExplorationOrder::Bfs => worklist.pop_front(),
+        SymbolicExplorationOrder::Dfs => worklist.pop_back(),
+    }
+}
+
 impl SymbolicExecutor {
     /// Creates a symbolic executor from Foundry's symbolic configuration.
     ///
@@ -134,7 +145,7 @@ impl SymbolicExecutor {
         let path_limit = self.config.path_width() as usize;
         let depth_limit = self.config.execution_depth() as usize;
 
-        while let Some(mut state) = worklist.pop_front() {
+        while let Some(mut state) = pop_worklist(&mut worklist, self.config.exploration_order) {
             if completed_paths >= path_limit {
                 return Ok(SymbolicRunResult::Incomplete {
                     kind: SymbolicStopReason::Stuck,
@@ -532,7 +543,7 @@ impl SymbolicExecutor {
         let path_limit = self.config.path_width() as usize;
         let depth_limit = self.config.execution_depth() as usize;
 
-        while let Some(mut state) = worklist.pop_front() {
+        while let Some(mut state) = pop_worklist(&mut worklist, self.config.exploration_order) {
             if *completed_paths >= path_limit {
                 return Err(SymbolicError::Unsupported("symbolic path limit exceeded"));
             }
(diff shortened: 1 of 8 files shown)


crates/evm/symbolic/src/executor.rs
```diff
@@ -134,7 +145,7 @@ impl SymbolicExecutor {
         let path_limit = self.config.path_width() as usize;
         let depth_limit = self.config.execution_depth() as usize;
 
-        while let Some(mut state) = worklist.pop_front() {
+        while let Some(mut state) = pop_worklist(&mut worklist, self.config.exploration_order) {
```
grandizzy, line 170: **Local batch scheduler 1 — `call_concrete_target` (~line 2570):**

```rust
let mut iter = parents.into_iter();
let Some(first) = iter.next() else { return Ok(StepOutcome::AssumeRejected); };
*state = first;
worklist.extend(iter);
```

Hard-FIFO selection ignores `exploration_order`. Should route through an order-aware helper, e.g.:

```rust
let Some(first) = pop_batch(&mut parents, self.config.exploration_order) else { ... };
*state = first;
spill_batch(parents, worklist, self.config.exploration_order);
```

crates/evm/symbolic/src/executor.rs
```diff
@@ -532,7 +543,7 @@ impl SymbolicExecutor {
...
```
grandizzy, line 568: **Local batch scheduler 2 — `call_symbolic_target` (~line 2808-2820):**

```rust
StepOutcome::Continue => {
    parents.push_back(branch);
    parents.extend(branch_worklist);
}
...
let Some(first) = parents.pop_front() else { ... };
*state = first;
worklist.extend(parents);
```

Two problems under `dfs`:
1. `parents.pop_front()` is hard-FIFO.
2. The nested merge flattens inner `branch_worklist` candidates in FIFO order, so even if the outer pop becomes LIFO-aware, candidate ordering inside the batch is still wrong. Both ends need the shared scheduler.

crates/evm/symbolic/src/executor.rs
```diff
@@ -3028,7 +3039,7 @@ impl SymbolicExecutor {
...
```
grandizzy, line 3062: **Local batch scheduler 3 — `create` merge (~line 3020):**

```rust
let mut iter = parents.into_iter();
let Some(first) = iter.next() else { ... };
*state = first;
worklist.extend(iter);
```

Same pattern as `call_concrete_target` — hard FIFO selection on a local sub-worklist. Should be order-aware so contract-creation sibling exploration respects `dfs`.

**Local batch scheduler 4 — `branch_accesses_cheatcode_if_needed` (~line 3338):**

```rust
let Some(first_branch) = branches.pop_front() else { ... };
*state = first_branch;
worklist.extend(branches);
```

Same hard-FIFO selection. Replace with the shared scheduler so `dfs` actually behaves LIFO across `vm.accesses(address)` cheatcode branching.

---

**Suggested helpers to add alongside `pop_worklist`:**

```rust
pub(crate) fn pop_batch<T>(
    batch: &mut VecDeque<T>,
    order: SymbolicExplorationOrder,
) -> Option<T> {
    match order {
        SymbolicExplorationOrder::Bfs => batch.pop_front(),
        SymbolicExplorationOrder::Dfs => batch.pop_back(),
    }
}

pub(crate) fn spill_batch<T>(
    batch: VecDeque<T>,
    worklist: &mut VecDeque<T>,
    order: SymbolicExplorationOrder,
) {
    match order {
        SymbolicExplorationOrder::Bfs => worklist.extend(batch),
        SymbolicExplorationOrder::Dfs => worklist.extend(batch.into_iter().rev()),
    }
}
```

Centralizing the "select current + spill rest" pattern is the smallest change that makes `dfs` mean LIFO globally rather than only at the top-level worklist.

grandizzy (commented):
**Inconsistency: local batch schedulers still force FIFO regardless of `exploration_order`.**

The top-level worklist swap (`pop_front` → `pop_back`) is sound, but four helper-local schedulers still hard-code FIFO selection. Under `dfs` they pick the *first* sibling as the current state and append the rest at the tail in source order, so siblings within a batch are not LIFO with respect to each other. Until these are addressed, `dfs` is only LIFO at the top-level worklist.

See per-line comments below.

grandizzy (approved):
👍

PR #14877, opened 2026-05-22
fix(symbolic): harden soundness fallbacks

figtracer:
## Summary

Hardens native symbolic execution fallbacks that could previously return a silently wrong answer. The goal is conservative behavior: either model the case soundly enough for replay, or report it as incomplete/unsupported instead of pretending it was proven.

## Fixed from the original soundness gist

| Gist concern | Previous behavior | This PR |
| --- | --- | --- |
| Symbolic `SLOAD` defaulting | Unresolved symbolic slots could collapse to zero/default storage and hide mapping/storage bugs. | Uses arbitrary symbolic storage by default for unresolved symbolic slots, keeps `ZeroInit` only as explicit opt-in, and concretizes constrained symbolic keys before reading backend/overlay storage. |
| Transient storage lifetime | `TSTORE`/`TLOAD` state leaked across top-level invariant sequence txs. | Clears transient storage at top-level symbolic tx boundaries. |
| `fail_on_revert = false` | Reverting invariant sequence steps dropped the whole branch. | Continues the sequence from the pre-call state when non-failing reverts are allowed. |
| Post-Cancun `SELFDESTRUCT` | Could apply pre-Cancun deletion semantics under Cancun+. | Reports EIP-6780 `SELFDESTRUCT` as incomplete/unsupported instead of modeling it incorrectly. |
| `GAS` / `gasleft()` | Returned `U256::MAX`, making low-gas branches look dead. | Returns bounded symbolic gas constrained by the executor gas limit. |
| Dynamic-offset memory epochs | Symbolic memory reads could observe stale symbolic writes after later concrete overwrites. | Dynamic reads respect overwrite epochs. |
| Top-level sequence code reads | Invariant sequence calls bypassed world overlay code from `etch`/`CREATE`/`SELFDESTRUCT`. | Uses world-overlay code for top-level sequence calls. |
| `CREATE` with symbolic runtime length | Could install max-length runtime code for symbolic-size constructor returns. | Rejects unsupported symbolic runtime lengths instead of installing incorrect code. |
| Cheatcode divergences | Signed symbolic ints, `vm.bound`, and prank/delegatecall handling diverged from concrete semantics. | Adds signed integer bounds, tightens `vm.bound` preconditions/results, and reports active prank + `DELEGATECALL` as unsupported rather than wrong. |

## Extra hardening from follow-up audit

| Extra issue | Fix |
| --- | --- |
| Constrained symbolic storage keys | If constraints force a symbolic storage key to a concrete slot, reads now use the concrete slot for backend storage and overlay write selection. |
| Oversized symbolic integer bit widths | `randomUint(bits)`, `randomInt(bits)`, `svm.createUint(bits, ...)`, and `svm.createInt(bits, ...)` reject `bits > 256`. |
| Symbolic `vm.bound` footguns | Concrete and symbolic cases now fail on invalid/out-of-range inputs; symbolic ranges are unsupported instead of underconstrained. |
| Delegate prank ambiguity | Active prank state on `DELEGATECALL` reports incomplete/unsupported until delegate-prank semantics are modeled precisely. |


mattsse (approved):

mablr (approved):
Resolved merge conflict, lgtm.

PR #14878, opened 2026-05-22
feat(common): isolate Tempo session key storage

0xKarl98:
This PR is a small follow-up to #14827 and the next storage-semantics step for OSS-162.

#14827 added the initial Tempo session registry scaffold: `wallet/sessions.toml`, session metadata, atomic upsert/removal, expiry tracking, and shared TOML helpers. 

This PR makes the key-storage side of that design explicit: temporary session key material now belongs to the session entry in `wallet/sessions.toml`, not to the persistent wallet key store in `wallet/keys.toml`.

That matters for `cast wallet session` because session keys are short-lived, narrowly scoped keys for one command run. They should be created, found, expired, and cleaned up without replacing or deleting the user's long-lived Tempo wallet access key.

### DIFF
--- a/crates/common/src/tempo/session.rs
+++ b/crates/common/src/tempo/session.rs
@@ -165,9 +165,17 @@ impl SessionRecord {
     pub fn mark_expired(&mut self, now: u64) -> usize {
         let mut updated = 0;
         for session in &mut self.sessions {
-            if session.status.is_live() && session.is_expired_at(now) {
+            let should_expire = session.status.is_live() && session.is_expired_at(now);
+            let should_clear_key =
+                session.key.is_some() && (should_expire || session.status.is_terminal());
+
+            if should_expire {
                 session.status = SessionStatus::Expired;
+            }
+            if should_clear_key {
                 session.key = None;
+            }
+            if should_expire || should_clear_key {
                 updated += 1;
             }
         }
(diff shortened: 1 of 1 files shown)


crates/common/src/tempo/session.rs
```diff
@@ -78,13 +102,27 @@ pub struct SessionEntry {
...
     }
+
+    /// Returns `true` if this session has usable local key material.
+    pub fn has_inline_key(&self) -> bool {
+        self.key.as_ref().is_some_and(SessionKeyMaterial::has_inline_key)
+    }
+
+    /// Returns `true` if this session is active, unexpired, and has key material.
+    pub fn has_live_key_at(&self, now: u64) -> bool {
+        self.status == SessionStatus::Active && !self.is_expired_at(now) && self.has_inline_key()
```
figtracer, line 124: `has_live_key_at` rejects `Pending`/`Revoking` even though `SessionStatus::is_live()` says they are live

crates/common/src/tempo/session.rs
figtracer, line 173: this only flips the status to `Expired`, but keeps `session.key` in `sessions.toml`; since they are short-lived we should clear it

crates/common/src/tempo/session.rs
0xKarl98: live is overloaded here: is_live() means “not terminal”, while this helper really means “usable for signing” ， so it rejects  Pending/Revoking as key in this status can not be used to sign

crates/common/src/tempo/session.rs
0xKarl98: Have cleared it

crates/common/src/tempo/session.rs
```diff
@@ -113,12 +151,23 @@ impl SessionRecord {
...
+
     /// Mark expired live entries as expired. Returns the number updated.
     pub fn mark_expired(&mut self, now: u64) -> usize {
         let mut updated = 0;
         for session in &mut self.sessions {
             if session.status.is_live() && session.is_expired_at(now) {
                 session.status = SessionStatus::Expired;
+                session.key = None;
                 updated += 1;
             }
```
mattsse: This only clears key material while a live entry is being transitioned to `Expired`. If an entry is already terminal (`Expired`, `Revoked`, or `Failed`) but still has `session.key`, this cleanup path returns `0` and never rewrites the file, leaving the private key in `wallet/sessions.toml`. Since this is the cleanup path for short-lived session keys, key removal should count as an update independently of the status transition.

```suggestion
            let should_expire = session.status.is_live() && session.is_expired_at(now);
            let should_clear_key = session.key.is_some() && (should_expire || session.status.is_terminal());

            if should_expire {
                session.status = SessionStatus::Expired;
            }
            if should_clear_key {
                session.key = None;
            }
            if should_expire || should_clear_key {
                updated += 1;
            }
```

mattsse (commented):
Left one inline cleanup suggestion.

0xKarl98:
@mattsse Have applied your nits

mattsse (approved):
Looks good after the follow-up cleanup.

mablr (approved):
Nice

figtracer (approved):
Lg

stevencartavia (approved):

PR #14882, opened 2026-05-23
feat: add `reentrancy-events` lint

stevencartavia:
Low-severity Solidity lint `reentrancy-events`: flags emit statements reachable after an external call on any control-flow path (matches Slither's reentrancy-events).

### DIFF
--- a/crates/lint/docs/reentrancy-events.md
+++ b/crates/lint/docs/reentrancy-events.md
@@ -15,7 +15,7 @@ For every function body, the lint performs a control-flow analysis that tracks w
 - High-level state-mutating external calls on interface or contract types (e.g. `IERC20(token).transfer(...)`). `view` and `pure` callees are not tracked.
 - Contract deployments via `new Foo(...)` (the constructor runs as an external interaction).
 
-External calls reached through internal/private/public helper functions and modifiers are tracked transitively when the helper is invoked by a bare identifier (e.g. `_helper()`). Member-form internal dispatch such as `Lib.f(...)`, `using for` syntax, and `super.f(...)` is **not** yet followed; external calls hidden behind those forms may go undetected.
+External calls reached through internal/private/public helper functions, modifiers, and `super.f(...)` base-chain dispatch are tracked transitively when the helper is invoked by a bare identifier (e.g. `_helper()`) or via `super.`. Member-form internal dispatch such as `Lib.f(...)` and `using for` syntax is **not** yet followed; external calls hidden behind those forms may go undetected.
 
 When the analysis encounters an `emit` statement reachable from a path that already executed a tracked external call, the statement is flagged.
 
--- a/crates/lint/docs/reentrancy-events.md
+++ b/crates/lint/docs/reentrancy-events.md
@@ -7,14 +7,17 @@ Flags `emit` statements that appear after an external call within the same funct
 
 ## What it does
 
-For every function body, the lint performs a control-flow analysis that tracks whether an external call has occurred on the path leading to each statement. External calls include:
+For every function body, the lint performs a control-flow analysis that tracks whether an external call has occurred on the path leading to each statement. Only calls that can plausibly affect log ordering or observable state are considered — `staticcall` and high-level `view` / `pure` external calls are excluded. Tracked calls include:
 
-- Low-level calls: `address.call(...)`, `delegatecall(...)`, `staticcall(...)` (with or without `{value: ...}` / `{gas: ...}` options).
+- Low-level calls: `address.call(...)` and `address.delegatecall(...)` (with or without `{value: ...}` / `{gas: ...}` options).
 - ETH sends: `address.transfer(...)`, `address.send(...)`.
 - `this.method(...)` self-external calls.
-- High-level external calls on interface or contract types (e.g. `IERC20(token).transfer(...)`).
+- High-level state-mutating external calls on interface or contract types (e.g. `IERC20(token).transfer(...)`). `view` and `pure` callees are not tracked.
+- Contract deployments via `new Foo(...)` (the constructor runs as an external interaction).
 
-External calls reached through internal/private/public helper functions and modifiers are also tracked transitively. When the analysis encounters an `emit` statement reachable from a path that already executed an external call, the statement is flagged.
+External calls reached through internal/private/public helper functions and modifiers are tracked transitively when the helper is invoked by a bare identifier (e.g. `_helper()`). Member-form internal dispatch such as `Lib.f(...)`, `using for` syntax, and `super.f(...)` is **not** yet followed; external calls hidden behind those forms may go undetected.
+
+When the analysis encounters an `emit` statement reachable from a path that already executed a tracked external call, the statement is flagged.
 
 ## Why is this bad?
 
--- a/crates/lint/docs/reentrancy-events.md
+++ b/crates/lint/docs/reentrancy-events.md
@@ -1,0 +1,58 @@
+# Reentrancy Events
+
+**Severity**: `Low`
+**ID**: `reentrancy-events`
+
+Flags `emit` statements that appear after an external call within the same function (or any internal helper it transitively calls). Emitting state-change events only after the external call returns can mislead off-chain consumers — including indexers, subgraphs, monitoring tools, and bridges — that rely on log ordering to reconstruct contract state.
+
+## What it does
+
+For every function body, the lint performs a control-flow analysis that tracks whether an external call has occurred on the path leading to each statement. External calls include:
+
+- Low-level calls: `address.call(...)`, `delegatecall(...)`, `staticcall(...)` (with or without `{value: ...}` / `{gas: ...}` options).
+- ETH sends: `address.transfer(...)`, `address.send(...)`.
+- `this.method(...)` self-external calls.
+- High-level external calls on interface or contract types (e.g. `IERC20(token).transfer(...)`).
+
+External calls reached through internal/private/public helper functions and modifiers are also tracked transitively. When the analysis encounters an `emit` statement reachable from a path that already executed an external call, the statement is flagged.
+
+## Why is this bad?
+
+Reentrancy and off-chain ordering both depend on event sequence:
+
+- A reentrant callee can observe (or trigger another contract to observe) events in an order that no longer reflects the final state of the calling contract.
+- Indexers, bridges, and monitoring tools that consume logs in emission order may apply state transitions incorrectly when events are not emitted alongside the writes they describe.
+
+Emitting the event **before** the external call ensures the log is anchored to the local state change, regardless of what the callee does.
+
+## Example
+
+### Bad
+
+```solidity
+contract BadCounter {
+    uint256 public counter;
+    event Counter(uint256 value);
+
+    function count(IExternal d) external {
+        counter += 1;
+        d.notify();             // external call first ...
+        emit Counter(counter);  // ... then the event (may be reordered by reentrancy)
+    }
+}
+```
+
+### Good
+
+```solidity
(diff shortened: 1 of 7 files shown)


crates/lint/src/sol/low/reentrancy_events.rs
```diff
@@ -0,0 +1,451 @@
...
+                if let Some(opts) = opts {
+                    for opt in *opts {
+                        self.analyze_expr(&opt.value, state);
+                    }
+                }
+                for arg in args.exprs() {
+                    self.analyze_expr(arg, state);
+                }
+
+                if is_external_call(self.gcx, self.hir, callee, args.len())
```
mattsse: [P2] This reuses `calls_loop::is_external_call`, but that helper is broader than what `reentrancy-events` needs: it includes `staticcall` and public/view function pointers. That means `target.read(); emit Tick();` gets reported even though a `STATICCALL` cannot emit logs or perform state-changing reentry. This lint should use a stricter predicate for calls that can actually reenter / affect event ordering, or filter out view/pure/static calls before setting `external_call_seen`.

crates/lint/src/sol/low/reentrancy_events.rs
```diff
@@ -0,0 +1,451 @@
...
+            }
+            StmtKind::Return(expr) => {
+                if let Some(expr) = expr {
+                    self.analyze_expr(expr, &mut entry);
+                }
+                Exits::return_(entry)
+            }
+            StmtKind::Break => Exits::break_(entry),
+            StmtKind::Continue => Exits::continue_(entry),
+            StmtKind::Loop(block, _) => {
```
mattsse, line 290: [P2] The single-pass loop handling misses back-edge paths. For example, `for (...) { emit Tick(); target.notify(); }` reaches the `emit` on the second iteration after the previous iteration's external call, but this pass has already visited the `emit` with a clean state and never revisits it. Since the rule is described as checking any control-flow path, loops need a small fixpoint/back-edge step or this case should be explicitly out of scope and tested as a known limitation.

crates/lint/src/sol/low/reentrancy_events.rs
```diff
@@ -0,0 +1,451 @@
...
+                {
+                    // Deploying a contract (`new Foo(args)`) executes that contract's
+                    // constructor, which is an external interaction that may emit events
+                    // or call back into this contract.
+                    state.external_call_seen = true;
+                }
+
+                // Follow internal/private/public helpers transitively so external calls in
+                // helpers also taint the caller's flow state.
+                for func_id in resolved_internal_function_ids(self.hir, callee) {
```
mattsse, line 408: [P3] The comment/docs say helpers are followed transitively, but `resolved_internal_function_ids` only resolves bare identifier calls. Member-form internal helpers such as `Lib.f(...)`, `using for`, and `super.f()` are missed; the test fixture even codifies those misses as known limitations. Please either implement member-form helper resolution here or narrow the docs/PR wording so users do not expect those calls to be covered.

mattsse (commented):
I left three review comments: two behavioral issues in the new analysis and one docs/implementation mismatch.

crates/lint/src/sol/low/reentrancy_events.rs
stevencartavia: [3fbdb09](https://github.com/foundry-rs/foundry/pull/14882/commits/3fbdb09a324a07f3dcda3e423ab6c748ca44ba16)

crates/lint/src/sol/low/reentrancy_events.rs
stevencartavia: [3fbdb09](https://github.com/foundry-rs/foundry/pull/14882/commits/3fbdb09a324a07f3dcda3e423ab6c748ca44ba16)

crates/lint/src/sol/low/reentrancy_events.rs
stevencartavia: [3fbdb09](https://github.com/foundry-rs/foundry/pull/14882/commits/3fbdb09a324a07f3dcda3e423ab6c748ca44ba16)

crates/lint/src/sol/low/calls_loop.rs
```diff
@@ -275,6 +279,49 @@ fn is_external_call<'gcx>(
...
+
+    if member.name == kw::Staticcall && is_address_like(hir, base) {
+        return false;
+    }
+
+    if matches!(member.name, sym::send | sym::transfer) && is_address_like(hir, base) {
+        return true;
+    }
+
+    if is_this(base) {
```
mattsse, line 319: [P2] This `this.method()` fast path still bypasses the mutability filter below. A self-external `public view` / `pure` call such as `this.selfRead(); emit Tick();` is compiled as a read-only external call and cannot emit logs or perform state-changing reentry, but this returns `true` before `func.state_mutability` is checked. Please resolve the member function first for `this` calls too, or only treat `this.method()` as tainting when the resolved callee is not `view`/`pure`.

mattsse (commented):
Follow-up review after the latest fixes: one remaining false-positive path for self-calls.

crates/lint/src/sol/low/calls_loop.rs
```diff
@@ -275,6 +279,49 @@ fn is_external_call<'gcx>(
...
+        return true;
+    }
+
+    if resolves_to_internal_library_extension(gcx, hir, base, *member, explicit_arg_count) {
```
figtracer, line 335: `super.foo()` can panic the linter by asking Solar for the builtin `super` type

crates/lint/src/sol/low/calls_loop.rs
```diff
@@ -275,6 +279,49 @@ fn is_external_call<'gcx>(
...
+        return false;
+    }
+
+    matches!(
```
figtracer: overloaded external methods are missed because member lookup requires a unique name

crates/lint/src/sol/low/reentrancy_events.rs
```diff
@@ -0,0 +1,490 @@
...
+            ExprKind::Slice(base, start, end) => {
+                self.analyze_expr(base, state);
+                if let Some(start) = start {
+                    self.analyze_expr(start, state);
+                }
+                if let Some(end) = end {
+                    self.analyze_expr(end, state);
+                }
+            }
+            ExprKind::Ternary(cond, then_expr, else_expr) => {
```
figtracer, line 465: `expr_aborted` leaks across ternary branches, so an aborting branch can hide a live external-call branch before an emit. Please track abort state per branch

stevencartavia: [2146752](https://github.com/foundry-rs/foundry/pull/14882/commits/21467522e3aa59bc93b088cfa4e96bfaa26f2482)

crates/lint/src/sol/low/calls_loop.rs
stevencartavia: [2146752](https://github.com/foundry-rs/foundry/pull/14882/commits/21467522e3aa59bc93b088cfa4e96bfaa26f2482)

crates/lint/src/sol/low/calls_loop.rs
stevencartavia: [2146752](https://github.com/foundry-rs/foundry/pull/14882/commits/21467522e3aa59bc93b088cfa4e96bfaa26f2482)

crates/lint/src/sol/low/calls_loop.rs
stevencartavia: [2146752](https://github.com/foundry-rs/foundry/pull/14882/commits/21467522e3aa59bc93b088cfa4e96bfaa26f2482)

crates/lint/src/sol/low/reentrancy_events.rs
```diff
@@ -0,0 +1,531 @@
...
+                for func_id in
+                    resolved_super_function_ids(self.hir, self.enclosing_contract, callee)
+                {
+                    self.analyze_internal_call(func_id, state);
+                }
+            }
+            ExprKind::Assign(lhs, _, rhs) | ExprKind::Binary(lhs, _, rhs) => {
+                self.analyze_expr(lhs, state);
+                self.analyze_expr(rhs, state);
+            }
```
mablr, line 446: `&&` / `||` modeled as eager evaluation. `Binary(lhs, _, rhs)` always analyzes both sides, false negatives (RHS aborts unconditionally taints whole stmt) and false positives (`false && ext.call()` tainting). Fix: fork like the ternary case.

crates/lint/docs/reentrancy-events.md
```diff
@@ -0,0 +1,61 @@
...
+
+For every function body, the lint performs a control-flow analysis that tracks whether an external call has occurred on the path leading to each statement. Only calls that can plausibly affect log ordering or observable state are considered — `staticcall` and high-level `view` / `pure` external calls are excluded. Tracked calls include:
+
+- Low-level calls: `address.call(...)` and `address.delegatecall(...)` (with or without `{value: ...}` / `{gas: ...}` options).
+- ETH sends: `address.transfer(...)`, `address.send(...)`.
+- `this.method(...)` self-external calls.
+- High-level state-mutating external calls on interface or contract types (e.g. `IERC20(token).transfer(...)`). `view` and `pure` callees are not tracked.
+- Contract deployments via `new Foo(...)` (the constructor runs as an external interaction).
+
+External calls reached through internal/private/public helper functions and modifiers are tracked transitively when the helper is invoked by a bare identifier (e.g. `_helper()`). Member-form internal dispatch such as `Lib.f(...)`, `using for` syntax, and `super.f(...)` is **not** yet followed; external calls hidden behind those forms may go undetected.
```
mablr: This point is stale, the impl now does follow `super` ?

crates/lint/src/sol/low/reentrancy_events.rs
```diff
@@ -0,0 +1,531 @@
...
+                // `assert(false)`) or via an inlined helper with no normal exit.
+                if is_aborting_call(expr) || self.expr_aborted {
+                    return Exits::abort();
+                }
+                Exits::fallthrough(entry)
+            }
+            StmtKind::Block(block) | StmtKind::UncheckedBlock(block) => {
+                self.analyze_block(block, placeholder, entry)
+            }
+            StmtKind::Emit(expr) => {
```
mablr, line 257: `expr_aborted` ignored in most statement kinds. It's only consumed in `DeclSingle`/`DeclMulti`/`Expr`. So `emit`, `return`, `if cond`, `try` expr, and `modifier` args all walk past an aborting subexpression as if reachable.

crates/lint/src/sol/low/reentrancy_events.rs
stevencartavia: [32d0046](https://github.com/foundry-rs/foundry/pull/14882/commits/32d00462344d91e5a0f1004011edae5234c618b8)

crates/lint/docs/reentrancy-events.md
stevencartavia: [32d0046](https://github.com/foundry-rs/foundry/pull/14882/commits/32d00462344d91e5a0f1004011edae5234c618b8)

crates/lint/src/sol/low/reentrancy_events.rs
stevencartavia: [32d0046](https://github.com/foundry-rs/foundry/pull/14882/commits/32d00462344d91e5a0f1004011edae5234c618b8)

mattsse (approved):
Approved after follow-up fix for `this`/`super` overload resolution and passing local lint UI validation.

PR #14883, opened 2026-05-23
feat(tempo): resolve live session signers

0xKarl98:
### Summary
Follow-up to #14827 and #14878 for OSS-162.

#14827 added the initial Tempo session registry scaffold: session metadata, wallet/sessions.toml, atomic upsert/removal, expiry tracking, and shared TOML registry helpers. #14878 made the storage semantics explicit by keeping temporary session key material inside session entries instead of the persistent wallet/keys.toml access-key store.

This PR adds the next resolver layer: 
it loads a live session-scoped key from the local session registry, expires stale entries before use, validates that the stored key material matches the recorded session key address, and returns the WalletSigner plus TempoAccessKeyConfig needed to use that session key for outgoing transactions.

It also decodes and validates inline key_authorization data, ensuring the authorization belongs to the stored key, chain, and root signer, and that its expiry, spending limits, and call scope match the stored session policy.

### DIFF
--- a/crates/common/src/tempo/session.rs
+++ b/crates/common/src/tempo/session.rs
@@ -268,6 +268,7 @@ pub fn resolve_live_session_signer(
     Ok(Some(ResolvedSessionSigner { session, signer, access_key }))
 }
 
+/// Ensures a session key authorization matches the stored key, chain, root signer, and policy.
 fn validate_session_key_authorization(
     session: &SessionEntry,
     key: &SessionKeyMaterial,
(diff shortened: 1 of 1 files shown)


mattsse (commented):
One note on the stored policy semantics: `KeyAuthorization` distinguishes `None` as unrestricted from `Some([])` as deny-all, but `SessionEntry` stores `limits`/`scope` as defaulted empty vecs. As written, unrestricted authorizations are rejected when the stored policy was omitted. Could we preserve that distinction, e.g. by storing these fields as `Option<Vec<_>>` or mapping empty stored fields to expected `None` if empty means unrestricted?

0xKarl98:
@mattsse  Have cleared  up session policy semantic boundaries

mattsse (approved):
LGTM

PR #14886, opened 2026-05-23
feat(lint): add `unprotected-initializer`

figtracer:
## Summary
- add a high-severity `unprotected-initializer` lint for upgradeable implementations
- flag public/external initializer-like functions that write state when the implementation constructor does not lock initializers
- add lint docs/book counterpart and UI fixture coverage

### DIFF
--- a/crates/lint/docs/unprotected-initializer.md
+++ b/crates/lint/docs/unprotected-initializer.md
@@ -10,8 +10,7 @@ This lint reports initializer-like functions that:
 - are public or external;
 - are marked with `initializer` or `reinitializer`;
 - write state directly or through an internal helper; and
-- are in an implementation whose constructor does not call `_disableInitializers()` and is not
-  itself marked `initializer`; and
+- are in an implementation whose constructor does not call `_disableInitializers()`; and
 - are in a contract with a public or external `delegatecall`, `callcode`, or `selfdestruct` path
   that is not restricted to proxy calls with `onlyProxy`.
 
--- a/crates/lint/docs/unprotected-initializer.md
+++ b/crates/lint/docs/unprotected-initializer.md
@@ -10,7 +10,7 @@ This lint reports initializer-like functions that:
 - are public or external;
 - are marked with `initializer` or `reinitializer`;
 - write state directly or through an internal helper; and
-- are in an implementation whose constructor does not call `_disableInitializers()`; and
+- are in an implementation whose constructor does not call an inherited `_disableInitializers()`; and
 - are in a contract with a public or external `delegatecall`, `callcode`, or `selfdestruct` path
   that is not restricted to proxy calls with `onlyProxy`.
 
--- a/crates/lint/docs/unprotected-initializer.md
+++ b/crates/lint/docs/unprotected-initializer.md
@@ -1,17 +1,19 @@
 # Unprotected initializer
 
 `unprotected-initializer` flags upgradeable contracts whose public or external initializer can
-still be called directly on the implementation contract.
+still be called directly on an implementation contract that exposes a destructive entry point.
 
 ## What it detects
 
 This lint reports initializer-like functions that:
 
 - are public or external;
-- are marked with `initializer` or `reinitializer`, or are named like `initialize`;
+- are marked with `initializer` or `reinitializer`;
 - write state directly or through an internal helper; and
 - are in an implementation whose constructor does not call `_disableInitializers()` and is not
-  itself marked `initializer`.
+  itself marked `initializer`; and
+- are in a contract with a public or external `delegatecall`, `callcode`, or `selfdestruct` path
+  that is not restricted to proxy calls with `onlyProxy`.
 
 ## Examples
 
@@ -21,6 +23,11 @@ contract Vault is Initializable {
 
     function initialize(address owner_) public initializer {
         owner = owner_;
+    }
+
+    function execute(address target, bytes calldata data) external {
+        (bool ok,) = target.delegatecall(data);
+        require(ok);
     }
 }
 ```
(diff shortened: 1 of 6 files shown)


mattsse (changes requested):
Thanks, this is close, but I think it needs changes before landing. I added inline suggestions for the straightforward parts.

Blocking issues:

- [`initializers_disabled_in_constructor`](https://github.com/foundry-rs/foundry/blob/1f86e0dab41ae1b1699f88f1d62d7c79bb66841d/crates/lint/src/sol/high/unprotected_initializer.rs#L76-L81) treats `constructor() initializer` as equivalent to `_disableInitializers()`, and [`check_nested_contract`](https://github.com/foundry-rs/foundry/blob/1f86e0dab41ae1b1699f88f1d62d7c79bb66841d/crates/lint/src/sol/high/unprotected_initializer.rs#L45-L46) then returns early for the whole contract. That causes a false negative for contracts with later `reinitializer(n)` functions: the constructor locks version 1, but `reinitializer(2)` can still be called directly on the implementation. Please only treat `_disableInitializers()` as a full implementation lock, or handle constructor `initializer` by initializer version, and add a fixture around [`ProtectedConstructorInitializer`](https://github.com/foundry-rs/foundry/blob/1f86e0dab41ae1b1699f88f1d62d7c79bb66841d/crates/lint/testdata/UnprotectedInitializer.sol#L115-L123) with `constructor() initializer`, a `reinitializer(2)` state write, and a public destructive entrypoint.

- [`collect_state_write_lhs_vars`](https://github.com/foundry-rs/foundry/blob/1f86e0dab41ae1b1699f88f1d62d7c79bb66841d/crates/lint/src/sol/high/unprotected_initializer.rs#L556-L581) misses writes through local storage aliases because it only reports an LHS when the final identifier resolves directly to a state variable. Patterns like `Layout storage s = _layout(); s.owner = owner_;` are common in upgradeable storage-layout code and would evade this lint. Please either track storage aliases through the initializer body/internal helpers or explicitly narrow the lint with docs and fixtures for this limitation.

CI is also red because `forge::cli::lint::ensure_lint_rule_docs` gets a 404 for https://getfoundry.sh/forge/linting/unprotected-initializer, so the book page needs to be published/fixed before merge.

crates/lint/src/sol/high/unprotected_initializer.rs
```diff
@@ -0,0 +1,624 @@
...
+        && matches!(func.visibility, Visibility::Public | Visibility::External)
+        && !matches!(func.state_mutability, StateMutability::Pure | StateMutability::View)
+        && has_initializer_modifier(hir, func)
+}
+
+fn initializers_disabled_in_constructor(hir: &hir::Hir<'_>, contract: &hir::Contract<'_>) -> bool {
+    contract.linearized_bases.iter().filter_map(|&cid| hir.contract(cid).ctor).any(|ctor_id| {
+        let ctor = hir.function(ctor_id);
+        has_modifier_named(hir, ctor, "initializer")
+            || function_calls_named(hir, ctor, contract.linearized_bases, "_disableInitializers")
```
mattsse: Permalink: https://github.com/foundry-rs/foundry/blob/1f86e0dab41ae1b1699f88f1d62d7c79bb66841d/crates/lint/src/sol/high/unprotected_initializer.rs#L76-L81

`constructor() initializer` should not be treated as a full implementation lock. It only initializes/locks the first version, so a later `reinitializer(2)` remains callable on the implementation. The safe default here is to only accept `_disableInitializers()` as the full lock.

```suggestion
        function_calls_named(hir, ctor, contract.linearized_bases, "_disableInitializers")
```

crates/lint/docs/unprotected-initializer.md
```diff
@@ -0,0 +1,61 @@
...
+
+## What it detects
+
+This lint reports initializer-like functions that:
+
+- are public or external;
+- are marked with `initializer` or `reinitializer`;
+- write state directly or through an internal helper; and
+- are in an implementation whose constructor does not call `_disableInitializers()` and is not
+  itself marked `initializer`; and
```
mattsse: Permalink: https://github.com/foundry-rs/foundry/blob/1f86e0dab41ae1b1699f88f1d62d7c79bb66841d/crates/lint/docs/unprotected-initializer.md#L13-L14

This doc currently encodes the same unsafe exemption. Unless the implementation handles initializer versions precisely, the docs should only describe `_disableInitializers()` as the full implementation lock.

```suggestion
- are in an implementation whose constructor does not call `_disableInitializers()`; and
```

mattsse (commented):
Adding inline suggestions for the request-changes review above.

mattsse (dismissed):

crates/lint/src/sol/high/unprotected_initializer.rs
```diff
@@ -0,0 +1,621 @@
...
+}
+
+fn callee_name_is(hir: &hir::Hir<'_>, callee: &hir::Expr<'_>, name: &str) -> bool {
+    match &callee.peel_parens().kind {
+        ExprKind::Ident(resolutions) => resolutions.iter().any(|res| {
+            matches!(res, Res::Item(ItemId::Function(fid)) if hir.function(*fid).name.is_some_and(|ident| ident.as_str() == name))
+        }),
+        ExprKind::Member(_, member) => member.as_str() == name,
+        _ => false,
+    }
```
mablr: This matches by method name string alone, so `SomeLib._disableInitializers()` satisfies the constructor-protection check even though it writes to an unrelated storage slot.

```solidity
library Lock { function _disableInitializers() internal pure {} }

contract Vuln is Initializable {
    constructor() { Lock._disableInitializers(); } // falsely treated as protection
    function initialize(address o) public initializer { owner = o; } // no warning but should warn
    function execute(address t, bytes calldata d) external { t.delegatecall(d); }
}
```

mablr (dismissed):

mattsse (approved):

mablr (approved):

PR #14889, opened 2026-05-23
fix(anvil): align CLI mining default with NodeConfig

### LINKED ISSUE #14867: anvil v1.7.1 CLI binary auto-mines per eth_sendTransaction by default; diverges from NodeConfig::default() and breaks batch-sim contracts
## Summary

In v1.7.1, the CLI binary and the library-embeddable `NodeConfig::default()` diverged on the auto-mining default:

- **CLI binary** (`anvil` invoked from the shell): auto-mines a fresh block for every `eth_sendTransaction` by default. To get batch-style behavior (one block per batch), the caller must pass `--no-mining` AND explicitly fire `evm_mine` after the batch.
- **In-process embedding via `NodeConfig::default()` + `anvil::spawn()`**: batches RPC-batched txs into one block correctly without any extra config.

This divergence is undocumented and breaks downstream wrappers expecting library-style behavior from the CLI binary. At [Hypernative](https://hypernative.io) we wrap the CLI binary in a Go service that depends on "batch-sim contract: all batch txs share one block." With v1.7.1's CLI default, the contract silently breaks (tx 0 lands in block N, tx 1 in block N+1, etc.).

## Reproduce

```bash
anvil --port 8545 &
# Send two txs in a JSON-RPC batch:
curl -sS http://localhost:8545 -H 'Content-Type: application/json' \
  -d '[{"jsonrpc":"2.0","id":1,"method":"eth_sendTransaction","params":[{"from":"...","to":"...","value":"0x1"}]},
       {"jsonrpc":"2.0","id":2,"method":"eth_sendTransaction","params":[{"from":"...","to":"...","value":"0x1"}]}]'
# Tx 0 lands in block N, tx 1 lands in block N+1 — each in its own block.
```

vs. the equivalent via embedding (`NodeConfig::default().spawn()` and the same RPC batch) where both end up in block N.

## Asks

- **(Preferred.)** Align the CLI default with `NodeConfig::default()` — no auto-mining unless explicitly requested via `--block-time` or similar.
- **(Fallback.)** Document the divergence prominently in the v1.7.1+ release notes and in `anvil --help` so callers know to pass `--no-mining` for batch use cases.

## Current downstream workaround

Pass `--no-mining` when spawning the CLI binary, then issue an explicit `evm_mine` RPC after each batch's `SendRawTransactionBatch` / `SendTransactionBatch`.

HusseinAdeiza:
## Summary

Fixes #14867 

Aligns the CLI binary's default mining mode with `NodeConfig::default()` by changing from instant auto-mining to manual mining (`MiningMode::None`).

## Root Cause

The divergence occurred in `crates/anvil/src/lib.rs` (lines 182-186). When neither `block_time` nor `no_mining` was set, the CLI defaulted to `MiningMode::instant()`, which mined a new block per transaction.

## Changes

- Changed default mining mode from `MiningMode::instant(max_transactions, listener)` to `MiningMode::None`
- Added clarifying comments explaining the behavior

## Testing

Manual testing confirms:
- **Before:** TX 1 → Block N, TX 2 → Block N+1 (instant mining)
- **After:** TX 1 + TX 2 queued, `evm_mine` → Both in Block N ✅

## Breaking Change

This is a **breaking change** for users who relied on auto-mining. Migration options:

1. Use `--block-time` for interval mining:
```bash
   anvil --block-time 1
```

2. Enable auto-mine via RPC:
```bash
   curl -X POST http://localhost:8545 \
     -d '{"jsonrpc":"2.0","method":"evm_setAutomine","params":[true],"id":1}'
```

3. Use `evm_mine` manually after transactions

## Checklist

- [x] Code change implemented
- [x] Manual testing completed
- [ ] CHANGELOG entry (will add after approval)

### DIFF
--- a/crates/anvil/src/lib.rs
+++ b/crates/anvil/src/lib.rs
@@ -180,9 +180,10 @@ pub async fn try_spawn(mut config: NodeConfig) -> Result<(EthApi<FoundryNetwork>
     } else if no_mining {
         MiningMode::None
     } else {
-        // get a listener for ready transactions
-        let listener = pool.add_ready_listener();
-        MiningMode::instant(max_transactions, listener)
+        // Default to manual mining (no auto-mine)
+        // This aligns CLI behavior with NodeConfig::default() expectations
+        // Users can enable auto-mining via --block-time or evm_setAutomine RPC
+        MiningMode::None
     };
 
     let miner = match &fork {


mattsse:
Closing this since we want to keep the default mining behavior as-is. #14891 addresses the concurrent transaction batching issue without changing the default automine semantics.

HusseinAdeiza:
ok good

PR #14890, opened 2026-05-24
fix(fmt): backport multi-statement block inlining fix to 1.5

### LINKED ISSUE #14888: bug(fmt): single-line `if (cond) { a; b; }` loses braces, trailing statements reparented outside the `if`
### Component

Forge

```
forge Version: 1.5.1-stable
Commit SHA: b0a9dd9ceda36f63e2326ce530c10e6916f4b8a2
Build Timestamp: 2025-12-22T11:39:01.425730780Z
```

`forge fmt`

### Describe the bug

`forge fmt` silently changes program semantics when an `if` body is written as a single-line braced block containing multiple statements:

```solidity
if (cond) { a; b; continue; }
```

After formatting, the braces are dropped and only the first statement remains under the `if` — the remaining statements are reparented to the surrounding block and execute unconditionally:

```solidity
if (cond) a;
b;
continue;
```

This is not a cosmetic change. `continue` / `break` / `return` now fire on every iteration; intervening `i += N` or similar mutations apply on every iteration regardless of the original condition. Control flow is materially broken. The default `single_line_statement_blocks = "preserve"` does not protect this case.

### Minimal repro

```solidity
pragma solidity ^0.8.24;

contract Repro {
    function decode(bytes memory src) external pure returns (uint256 count) {
        for (uint256 i = 0; i < src.length;) {
            if (src[i] == "&") { count++; i += 4; continue; }
            i++;
        }
    }

    function indexOf(bytes memory haystack, bytes memory needle) external pure returns (uint256) {
        for (uint256 i = 0; i + needle.length <= haystack.length; i++) {
            bool match_ = true;
            for (uint256 j = 0; j < needle.length; j++) {
                if (haystack[i + j] != needle[j]) { match_ = false; break; }
            }
            if (match_) return i;
        }
        return type(uint256).max;
    }
}
```

Run `forge fmt Repro.sol`. Result:

```solidity
function decode(bytes memory src) external pure returns (uint256 count) {
    for (uint256 i = 0; i < src.length;) {
        if (src[i] == "&") count++;
        i += 4;
        continue;
        i++;
    }
}

function indexOf(bytes memory haystack, bytes memory needle) external pure returns (uint256) {
    for (uint256 i = 0; i + needle.length <= haystack.length; i++) {
        bool match_ = true;
        for (uint256 j = 0; j < needle.length; j++) {
            if (haystack[i + j] != needle[j]) match_ = false;
            break;
        }
        if (match_) return i;
    }
    return type(uint256).max;
}
```

`decode` now does `i += 4; continue;` on every iteration regardless of the character, and `i++` is dead code. `indexOf` breaks out of its inner loop on the first character comparison every time.

### Expected behavior

Either preserve the braces and the inline multi-statement layout, or move each statement onto its own line while keeping all of them inside the `if`-block braces. Never silently drop the braces and reparent statements.

### Operating System

Linux (WSL2)

### Severity

High — formatter changes program semantics. The bug is silent unless tests catch the behavioral drift (in our case, an SVG-parsing helper using `_decodeXmlEntities` and `_indexOf` started returning empty / wrong results after `forge fmt`, breaking ~12 unrelated tests until we tracked the diff back to this pattern).

figtracer:
Fixes #14888

## Summary
- Backports #13566 to `release-1.5.0`, which contains the affected `v1.5.1` / `stable` tag.
- Keeps braces around braced statement blocks with multiple statements so formatter output cannot reparent trailing statements.
- Adds an `if (...) { ...; ...; continue; }` regression fixture for the issue pattern across statement-block modes.
- Refreshes compatible patch versions in `Cargo.lock` to clear current RustSec failures on the old release branch.

## Verification
- Reproduced the bad output with the installed `forge` binary before the backport.
- Verified current `master` already keeps braces for the issue repro.
- `cargo test -p forge-fmt IfStatement --test formatter`
- `cargo fmt --check` (passes; stable rustfmt prints existing warnings for unstable rustfmt.toml options).
- `cargo deny --all-features check all` (passes with existing warnings for unmatched sources/licenses and yanked `keccak`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -96,7 +96,7 @@ dependencies = [
  "either",
  "k256",
  "once_cell",
- "rand 0.8.5",
+ "rand 0.8.6",
  "secp256k1 0.30.0",
  "serde",
  "serde_json",
@@ -416,7 +416,7 @@ dependencies = [
  "paste",
  "proptest",
  "proptest-derive",
- "rand 0.9.2",
+ "rand 0.9.3",
  "ruint",
  "rustc-hash",
  "serde",
@@ -621,7 +621,7 @@ dependencies = [
  "ethereum_ssz",
  "ethereum_ssz_derive",
  "jsonwebtoken",
- "rand 0.8.5",
+ "rand 0.8.6",
  "serde",
  "strum 0.27.2",
 ]
@@ -773,7 +773,7 @@ dependencies = [
  "coins-bip39",
  "eth-keystore",
  "k256",
- "rand 0.8.5",
+ "rand 0.8.6",
  "thiserror 2.0.17",
  "zeroize",
 ]
@@ -1076,7 +1076,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -1100,7 +1100,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


DaniPopes (approved):

figtracer:
Closing this backport. The formatter behavior is already fixed on master by #13566, and we do not need a release-1.5 backport for #14888.

PR #14892, opened 2026-05-24
docs: update install curl URL to getfoundry.sh

abhicris:
### What
Change the `curl` install command in `README.md` from `https://foundry.paradigm.xyz` to `https://getfoundry.sh`.

### Why
Every other link in the README already points to `getfoundry.sh` (the installation guide, all docs links, etc.). The curl snippet was the only place still referencing the old `foundry.paradigm.xyz` domain, creating an inconsistency for anyone copying the install command.

Small, scoped fix from kcolbchain contrib-bot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Blazing fast, portable and modular toolkit for Ethereum application development,
 ## Installation
 
 ```sh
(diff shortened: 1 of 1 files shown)


stevencartavia (commented):
hmm no, this is fine as is

PR #14893, opened 2026-05-24
 feat(tempo): plumb wallet session signer

0xKarl98:
Tempo wallet sessions should remain clearly separated from explicit wallet signers. This PR adds the CLI/session opts plumbing needed for that boundary without changing command routing or signer selection yet.

It keeps the session handling in one place and documents the guardrails for the follow-up signer integration:
- `--tempo.session` and `TEMPO_SESSION_ID` can be parsed into an effective session id;
- live session resolution validates the session chain and expected sender;
- session signing is prepared to fail closed when combined with explicit wallet signer options;
- session hints do not drive Tempo routing until the signer selection paths are wired up.

### DIFF
--- a/crates/cli/src/opts/tempo/session.rs
+++ b/crates/cli/src/opts/tempo/session.rs
@@ -44,7 +44,7 @@ impl TempoOpts {
             return Ok(None);
         };
         ensure_no_explicit_wallet_signer(wallet)?;
-        resolve_session_signer(session_id, wallet.from, expected_chain_id)
+        Ok(Some(resolve_session_signer(session_id, wallet.from, expected_chain_id)?))
     }
 
     /// Resolves the configured Tempo wallet session for multi-wallet commands.
@@ -58,7 +58,7 @@ impl TempoOpts {
             return Ok(None);
         };
         ensure_no_explicit_multi_wallet_signer(wallets)?;
-        resolve_session_signer(session_id, expected_sender, expected_chain_id)
+        Ok(Some(resolve_session_signer(session_id, expected_sender, expected_chain_id)?))
     }
 }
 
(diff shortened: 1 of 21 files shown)


crates/cli/src/opts/tempo/session.rs
```diff
@@ -0,0 +1,345 @@
...
+        );
+    }
+
+    Ok(resolved)
+}
+
+/// Rejects single-wallet signer options when a Tempo session is already selected.
+///
+/// Session signing must fail closed instead of falling back to a long-lived or explicit signer.
+fn ensure_no_explicit_wallet_signer(wallet: &WalletOpts) -> Result<()> {
```
grandizzy, line 107: better to move in WalletOpts as `has_explicit_signer` fn?

crates/cli/src/opts/tempo/session.rs
```diff
@@ -0,0 +1,345 @@
...
+    }
+    Ok(())
+}
+
+/// Rejects multi-wallet signer options when a Tempo session is already selected.
+///
+/// Script and broadcast session signing must not fall back to long-lived, browser, or explicit
+/// signers.
+fn ensure_no_explicit_multi_wallet_signer(wallets: &MultiWalletOpts) -> Result<()> {
+    let has_explicit_signer = wallets.interactive
```
grandizzy: same as above, lets move in `MultiWalletOpts`?

0xKarl98:
@grandizzy  This makes Semantics boundaries more clear , have addressed your concerns pls check again

crates/cli/src/opts/tempo/session.rs
```diff
@@ -0,0 +1,357 @@
...
+pub const TEMPO_SESSION_ID_ENV: &str = "TEMPO_SESSION_ID";
+
+impl TempoOpts {
+    /// Returns the effective session id, preferring the CLI flag over `TEMPO_SESSION_ID`.
+    pub fn session_id(&self) -> Result<Option<B256>> {
+        if let Some(session) = self.session {
+            return Ok(Some(session));
+        }
+
+        let Ok(raw) = std::env::var(TEMPO_SESSION_ID_ENV) else {
```
grandizzy: this silently ignores non-Unicode `TEMPO_SESSION_ID`, suggest handling it too
```rust
let raw = match std::env::var(TEMPO_SESSION_ID_ENV) {
    Ok(raw) => raw,
    Err(std::env::VarError::NotPresent) => return Ok(None),
    Err(std::env::VarError::NotUnicode(_)) => {
        eyre::bail!("invalid {TEMPO_SESSION_ID_ENV}: value is not valid UTF-8");
    }
};
```

grandizzy (commented):
looks good, one minor nit, pls check

grandizzy (approved):
lgtm!

mablr (approved):
sgtm

PR #14894, opened 2026-05-24
fix: align browser fee caps with suggested tips

Perico-perica46:
## Motivation

Browser wallets can replace the locally estimated priority fee with the value from `eth_maxPriorityFeePerGas`. The script broadcast path already accounts for that by raising `maxFeePerGas` when the browser-suggested tip is higher than the fee-history estimate. `forge create --browser` and `cast erc20 ... --browser` still used the raw EIP-1559 estimate, which can leave the fee cap below the tip the wallet submits.

## Solution

When filling browser-submitted transactions without an explicit priority fee, compare the EIP-1559 estimate with `eth_maxPriorityFeePerGas`. If the suggested tip is higher, raise the fee cap by the same delta and use that tip before sending the request to the browser wallet.


stevencartavia (commented):
- Moved the `eth_maxPriorityFeePerGas` bump in `cast erc20` behind a new `browser: bool` flag on `fill_tx`, so non browser sponsor flows don't make the extra RPC or get a silently-raised tip.
- Added a `priority <= maxFee` check after fee filling in both `erc20.rs` and `create.rs` to reject invalid EIP-1559 pairs locally instead of sending them to the wallet/RPC.
- Placed that check outside the inner `if`, so it also catches the case where the user explicitly supplies both `--gas price` and `--priority-gas-price` with `priority > maxFee`.

stevencartavia (dismissed):

Perico-perica46:
@stevencartavia Thanks for the update and approval. LGTM!

crates/forge/src/cmd/create.rs
```diff
@@ -486,13 +486,31 @@ impl CreateArgs {
...
+                if self.tx.priority_gas_price.is_none() {
+                    deployer.tx.set_max_priority_fee_per_gas(estimate.max_priority_fee_per_gas);
+                }
+                if self.tx.gas_price.is_none() {
+                    deployer.tx.set_max_fee_per_gas(estimate.max_fee_per_gas);
+                }
             }
-            if self.tx.gas_price.is_none() {
-                deployer.tx.set_max_fee_per_gas(estimate.max_fee_per_gas);
+            if let (Some(max_fee), Some(priority)) =
```
figtracer, line 494: `forge create --gas-price 1 --priority-gas-price 2` without `--gas-limit` can hit `eth_estimateGas` with an invalid EIP-1559 pair before this local error is reached

mablr: +1

crates/forge/src/cmd/create.rs
figtracer: this still sends invalid EIP-1559 fee pairs to `eth_createAccessList` before the new local `priority <= maxFee` validation runs. 

repro: `forge create --access-list --gas-price 1 --priority-gas-price 2` and Anvil logged `eth_createAccessList` before the CLI returned the local fee error

Perico-perica46:
@figtracer Updated, please take another look, thanks.

crates/cast/src/cmd/erc20.rs
```diff
@@ -641,13 +648,31 @@ where
...
             tx.set_gas_price(provider.get_gas_price().await?);
         }
-    } else if tx.max_fee_per_gas().is_none() || tx.max_priority_fee_per_gas().is_none() {
-        let estimate = provider.estimate_eip1559_fees().await?;
-        if tx.max_fee_per_gas().is_none() {
-            tx.set_max_fee_per_gas(estimate.max_fee_per_gas);
+    } else {
+        if tx.max_fee_per_gas().is_none() || tx.max_priority_fee_per_gas().is_none() {
+            let mut estimate = provider.estimate_eip1559_fees().await?;
+            if browser
```
figtracer: `cast send --browser` still uses the shared fee filler without this tip bump, so the same browser fee-cap issue can still happen outside `cast erc20`. 

let's move this into the shared builder

a better long-term solution for all of these would be an actual robust provider filler for the browser wallet as mentioned by @mablr... but we can keep this as-is for now i think

mablr: Yep def need to revamp all these browser cheats/workarounds. Not the scope of this PR.

Perico-perica46:
@mablr @figtracer Done, please take another look, thanks.

figtracer:
can you fix conflicts pls? @Perico-perica46

mablr (dismissed):
sgtm

figtracer (dismissed):

mablr (dismissed):
sgtm

stevencartavia (dismissed):

mablr (approved):
Made a smol fix, now g2g imo

figtracer (approved):

PR #14896, opened 2026-05-25
fix(anvil): separate Tempo txpool nonce lanes

nivanovvv:
## Motivation

Tempo transactions with a nonzero `nonceKey` can use the same sender and lane nonce while still being distinct pending transactions. `txpool_inspect` and `txpool_content` currently key each sender's entries only by `nonce`, so multiple nonzero Tempo nonce lanes with the same nonce overwrite each other in the RPC response.

## Solution

Use a `nonceKey:nonce` txpool key for nonzero Tempo nonce lanes while preserving the existing nonce-only key for regular transactions and zero-key Tempo transactions.

Add a regression test that submits two pending Tempo transactions with different nonzero `nonceKey` values and the same nonce, then checks both `txpool_inspect` and `txpool_content` keep both entries.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


figtracer (dismissed):
makes sense; i think the helper makes this less intuitive but its gtg

crates/anvil/tests/it/tempo.rs
```diff
@@ -690,6 +690,72 @@ async fn test_tempo_nonzero_lane_pending_tx_does_not_advance_scalar_nonce() {
...
+    let token = IERC20::new(PATH_USD, &provider);
+    let transfer_call = token.transfer(recipient, U256::from(50_000));
+    let calldata: Bytes = transfer_call.calldata().clone();
+
+    let chain_id = provider.get_chain_id().await.unwrap();
+    let base_fee = provider.get_gas_price().await.unwrap();
+    let nonce_keys = [U256::from(100), U256::from(101)];
+
+    for nonce_key in nonce_keys {
+        let tempo_tx = TempoTransaction {
```
mablr, line 714: `TempoTransaction` derives `Default` so you can use `..Default::default()` pattern.

mablr (commented):
Doesn't hurt imo. Just a smol non-blocking nit.

nivanovvv:
@mablr Done.

mablr (approved):

figtracer (approved):

PR #14898, opened 2026-05-25
refactor(symbolic): replace `String` errors in solver config with `enum`

mablr:
## Motivation

Clean-up `String` based symbolic config errors

## Solution

Introduces `SolverConfigError` with `EmptyCommand` and `UnterminatedQuote(char)`
variants, replacing all `Result<T, String>`.

Also tightens `split_quoted_args` to return the offending `char` directly instead
of a pre-formatted string, letting callers own the message context.


figtracer (approved):

grandizzy (approved):
sgtm!

PR #14899, opened 2026-05-25
feat(cast): wrap `--json` output in shared `JsonEnvelope`

mablr:
## Motivation

Towards OSS-239, there will be follow-ups

All scalar cast subcommands now branch on
`shell::is_json()` and call `print_json_success()`, producing `{"schema_version":1,"success":true,"data":<T>,...}` instead of bare strings.

Error paths in JSON mode now emit `JsonEnvelope::failure`, so `eyre` errors are structured rather than plain text: `{"schema_version":1,"success":false,"errors":[{"level":"error",...}],...}`

### DIFF
--- a/crates/cast/src/args.rs
+++ b/crates/cast/src/args.rs
@@ -15,7 +15,10 @@ use alloy_rpc_types::{BlockId, BlockNumberOrTag::Latest};
 use clap::{CommandFactory, Parser};
 use clap_complete::generate;
 use eyre::{Result, WrapErr};
-use foundry_cli::utils::{self, LoadConfig};
+use foundry_cli::{
+    json::print_json_success,
+    utils::{self, LoadConfig},
+};
 use foundry_common::{
     abi::{get_error, get_event},
     fmt::{format_tokens, format_uint_exp, serialize_value_as_json},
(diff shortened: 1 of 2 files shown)


grandizzy (commented):
looks good, can we refactor the dupe in a single helper, is the same pattern / code repeated several times

crates/cast/src/args.rs
figtracer, line 64: we should add a CLI regression test that checks representative scalar `--json` commands all return `JsonEnvelope`; otherwise missed direct `sh_println!` arms can ship unnoticed

crates/cast/src/args.rs
```diff
@@ -848,6 +677,7 @@ pub async fn run_command(args: CastArgs) -> Result<()> {
         }
 
         // 4Byte
+        // TODO(json): multiple signature candidates, needs array envelope
```
figtracer, line 680: should we bundle these TODO(s) in the same PR

crates/cast/src/args.rs
mablr: Good idea, as OSS-239 is not closed, I'll do it as a follow-up (added as  acceptance criteria)

crates/cast/src/args.rs
mablr: I wanted to discuss envelope format first and keep PRs contained for fast iteration.

grandizzy (approved):

mattsse (approved):
LGTM

PR #14900, opened 2026-05-25
fix: stabilize symbolic CI tests

decofe:
Fixes the two CI failures observed on #14796:

- Keep invariant skip detection on the grouped invariant path so multi-predicate skip output still reports every skipped invariant.
- Make canned symbolic solver subprocesses drain stdin before emitting fixed responses, avoiding broken-pipe races in portfolio scheduler tests.

Validation:
- `cargo test -p foundry-evm-symbolic portfolio_ -- --nocapture`
- `cargo test -p foundry-evm-symbolic portfolio_scheduler_penalizes_invalid_models -- --nocapture`
- `git diff --check`

Note: `cargo test -p forge --test cli test_cmd::skip_output -- --nocapture` compiled but failed during test setup because `forge init` could not clone `forge-std` (`HTTP 400`), before reaching the snapshot assertion.

Prompted by: @grandizzy

### DIFF
--- a/crates/evm/symbolic/src/tests.rs
+++ b/crates/evm/symbolic/src/tests.rs
@@ -2168,7 +2168,8 @@ fn portfolio_scheduler_promotes_recent_sat_winner() {
             vec![
                 "/bin/sh".to_string(),
                 "-c".to_string(),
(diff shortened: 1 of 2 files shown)


figtracer (approved):
makes sense

mablr (approved):
sgtm

PR #14901, opened 2026-05-25
feat(symbolic): normalize hard arithmetic queries

figtracer:
## Summary

| Area | Change |
| --- | --- |
| ERC4626-style inflation | Normalizes checked arithmetic and division predicates so symbolic execution reaches a concrete share-zero counterexample. |
| MiniVat-like product inequality | Adds narrow bounded monotonic-product pruning plus hard-arithmetic witness candidates for wrapping product counterexamples. |
| SMT shape | Canonicalizes commutative expressions, rebuilds split byte words, and extracts boolean conditions from 0/1 word ITEs before SMT emission. |
| Fallback safety | Validates every constraint before returning a heuristic witness, and downgrades to `Incomplete` if a heuristic witness was used but no concrete replayed counterexample was found. |

This is still counterexample-finding only for hard nonlinear arithmetic. It does not certify `Safe` from heuristic witnesses.

## Live testing

| Repro | Command | Result |
| --- | --- | --- |
| ERC4626-style inflation | `./target/debug/forge test --root /private/tmp/sym-hard-live --symbolic --match-test checkDepositReturnsShares -vv` | Replayed failure with `args=[18446744073709551615, 1, 1]`, `paths=6`, `queries=33`. |
| MiniVat-like product inequality | `./target/debug/forge test --root /private/tmp/sym-hard-live --symbolic --match-test checkPositionAlwaysSafe -vv` | Replayed failure with `args=[1, 2, 1, 2]`, `paths=4`, `queries=23`. |
| Portfolio smoke | same repros with `--symbolic-solver-portfolio z3,yices-smt2,bitwuzla` | Same counterexamples; faster locally (~0.7s ERC4626, ~0.4s MiniVat). |


crates/evm/symbolic/src/executor/run.rs
```diff
@@ -439,6 +448,15 @@ impl SymbolicExecutor {
             frontier = next_frontier;
         }
 
+        if self.solver.heuristic_witnesses() > 0 {
```
grandizzy: let's use a closure / helper to avoid code dupe

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -77,6 +77,11 @@ pub(crate) trait SymbolicSolver {
     /// Takes any captured verbose diagnostics collected by this backend.
```
grandizzy, line 92: let's split solver.rs in normalize,monotonic_product,hard_arith_fallback to be able to review easier

crates/evm/symbolic/src/executor/run.rs
```diff
@@ -244,6 +244,15 @@ impl SymbolicExecutor {
             });
         }
 
+        if self.solver.heuristic_witnesses() > 0 {
```
grandizzy: `heuristic_witnesses` is a monotonic counter, not per-run

`SmtLibSubprocessSolver::heuristic_witnesses` is a usize that only ever increments. The final-result downgrade in run_inner and run_invariant_inner compares it against 0:
```rust
if self.solver.heuristic_witnesses() > 0 {
    return Ok(SymbolicRunResult::Incomplete { ... });
}
```

If the same `SymbolicExecutor` / solver instance is reused across runs (which is the common case for an executor that drives multiple tests or invariant rounds), any earlier run that used a heuristic witness will permanently force every subsequent genuinely-safe run to be reported as Incomplete.

suggest capture a per-run baseline `let heuristic_baseline = self.solver.heuristic_witnesses();` and then compare deltas `if self.solver.heuristic_witnesses() > heuristic_baseline {`

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -1257,36 +1301,660 @@ pub(crate) fn validate_solver_model_output(
...
+        return Some(udiv_zero_condition(numerator, denominator));
+    }
+    if let Expr::Ite(condition, then_expr, else_expr) = expr {
+        let then_zero = normalize_expr_eq_zero_for_solver(then_expr).unwrap_or_else(|| {
+            BoolExpr::eq(normalize_expr_for_solver((**then_expr).clone()), Expr::Const(U256::ZERO))
+        });
+        let else_zero = normalize_expr_eq_zero_for_solver(else_expr).unwrap_or_else(|| {
+            BoolExpr::eq(normalize_expr_for_solver((**else_expr).clone()), Expr::Const(U256::ZERO))
+        });
+        if then_zero.smt().contains("bvudiv") || else_zero.smt().contains("bvudiv") {
```
grandizzy: `then_*/else_*.smt().contains("bvudiv")` serializes the entire subtree to an SMT-LIB string just to detect a residual UDiv. That allocates O(size of expression) on every Ite-normalization step. Suggested fix: add structural walkers and replace both call sites. The variants below match the actual enum definitions in expressions.rs:
```diff
--- a/crates/evm/symbolic/src/runtime/solver.rs
+++ b/crates/evm/symbolic/src/runtime/solver.rs
@@ existing helpers ...
+
+/// Returns true if `expr` syntactically contains a `UDiv` operation.
+pub(crate) fn expr_contains_udiv(expr: &Expr) -> bool {
+    match expr {
+        Expr::Const(_) | Expr::Var(_) => false,
+        Expr::Keccak { len, bytes, .. } => {
+            expr_contains_udiv(len) || bytes.iter().any(expr_contains_udiv)
+        }
+        Expr::Hash { bytes, .. } => bytes.iter().any(expr_contains_udiv),
+        Expr::Not(inner) => expr_contains_udiv(inner),
+        Expr::Op(op, left, right) => {
+            matches!(op, ExprOp::UDiv)
+                || expr_contains_udiv(left)
+                || expr_contains_udiv(right)
+        }
+        Expr::Ite(cond, then_expr, else_expr) => {
+            bool_contains_udiv(cond)
+                || expr_contains_udiv(then_expr)
+                || expr_contains_udiv(else_expr)
+        }
+    }
+}
+
+/// Returns true if `expr` syntactically contains a `UDiv` operation.
+pub(crate) fn bool_contains_udiv(expr: &BoolExpr) -> bool {
+    match expr {
+        BoolExpr::Const(_) => false,
+        BoolExpr::Not(inner) => bool_contains_udiv(inner),
+        BoolExpr::And(values) => values.iter().any(bool_contains_udiv),
+        BoolExpr::Eq(left, right) | BoolExpr::Cmp(_, left, right) => {
+            expr_contains_udiv(left) || expr_contains_udiv(right)
+        }
+    }
+}

@@ -1789,7 +1789,7 @@ pub(crate) fn normalize_expr_eq_zero_for_solver(expr: &Expr) -> Option<BoolExpr>
         let else_zero = normalize_expr_eq_zero_for_solver(else_expr).unwrap_or_else(|| {
             BoolExpr::eq(normalize_expr_for_solver((**else_expr).clone()), Expr::Const(U256::ZERO))
         });
-        if then_zero.smt().contains("bvudiv") || else_zero.smt().contains("bvudiv") {
+        if bool_contains_udiv(&then_zero) || bool_contains_udiv(&else_zero) {
             return None;
         }

```

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -1257,36 +1301,660 @@ pub(crate) fn validate_solver_model_output(
...
+    if let Expr::Ite(condition, then_expr, else_expr) = expr {
+        let then_nonzero = normalize_expr_ne_zero_for_solver(then_expr).unwrap_or_else(|| {
+            BoolExpr::eq(normalize_expr_for_solver((**then_expr).clone()), Expr::Const(U256::ZERO))
+                .not()
+        });
+        let else_nonzero = normalize_expr_ne_zero_for_solver(else_expr).unwrap_or_else(|| {
+            BoolExpr::eq(normalize_expr_for_solver((**else_expr).clone()), Expr::Const(U256::ZERO))
+                .not()
+        });
+        if then_nonzero.smt().contains("bvudiv") || else_nonzero.smt().contains("bvudiv") {
```
grandizzy: same as above, can reuse helper

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -252,18 +278,35 @@ impl SymbolicSolver for SmtLibSubprocessSolver {
         )
         .entered();
         trace!(query_id = self.queries, constraint_count = constraints.len(), "solver model");
-        if constraints_prefer_fallback_first(constraints)
-            && let Some(model) = fallback_single_var_model(constraints)
-        {
-            trace!("model: fallback single-var model");
+        if let Some(model) = hard_arith_fallback_model(constraints) {
```
grandizzy: This is bounded by `HARD_ARITH_FALLBACK_MAX_ASSIGNMENTS = 50_000` eval_bool_expr passes, executed on every model() call - including queries the solver could answer in milliseconds. For any test issuing many feasibility/model queries, this is a performance regression.

Note is_sat already takes a more conservative approach: it runs the fallback only after the solver returns Err(SolverUnknown) or "unknown". model() should probably do the same.

Suggested fix — defer to the unknown path, mirroring is_sat:
```diff
-    if let Some(model) = hard_arith_fallback_model(constraints) {
-        self.heuristic_witnesses += 1;
-        trace!("model: hard arithmetic fallback model");
-        return Ok(model);
-    }
     let output = match self.query(constraints, true) {
         Ok(output) => output,
         Err(SymbolicError::SolverUnknown) => {
             if let Some(model) = hard_arith_fallback_model(constraints) {
                 self.heuristic_witnesses += 1;
                 trace!("model: hard arithmetic fallback model after solver unknown");
                 return Ok(model);
             }
             return Err(SymbolicError::SolverUnknown);
         }
         Err(err) => return Err(err),
     };
```
Alternative if the up-front fallback is intentional for specific shapes: gate it on `constraints.iter().any(expr_contains_hard_arith)` so easy queries skip it entirely.

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -223,19 +235,33 @@ impl SymbolicSolver for SmtLibSubprocessSolver {
         )
         .entered();
         trace!(query_id = self.queries, constraint_count = constraints.len(), "solver is_sat");
-        if constraints_prefer_fallback_first(constraints)
-            && fallback_single_var_model(constraints).is_some()
-        {
-            trace!("is_sat: fallback single-var model");
-            return Ok(true);
+        if product_monotonic_unsat(constraints) {
```
grandizzy: Performance consideration: constraints are normalized twice on every `is_sat` call - `product_monotonic_unsat` calls `normalize_constraints_for_solver` internally, and `query()` does the same. Each normalize clones the entire expression tree, so every is_sat walks + allocates the full constraint set twice.

Suggested to normalize once and thread the result through:
```diff
 fn is_sat(&mut self, constraints: &[BoolExpr]) -> Result<bool, SymbolicError> {
-    if product_monotonic_unsat(constraints) {
+    let smt_constraints = normalize_constraints_for_solver(constraints);
+    if product_monotonic_unsat_normalized(&smt_constraints) {
         return Ok(false);
     }
-    let output = match self.query(constraints, false) { ... };
+    let output = match self.query_normalized(&smt_constraints, false) { ... };
     ...
 }
```
Add a `_normalized` variant of `product_monotonic_unsat` and query that skips the normalization step.

grandizzy (changes requested):
left some review comments, correctness and performance. Also makes sense to split solver.rs for better readability / review. thanks!

crates/evm/symbolic/src/executor/run.rs
figtracer: oh nice catch will address this

figtracer:
Benchmarks (`hyperfine --warmup 1 --runs 3 --ignore-failure`, local debug builds, failing repros expected):

| Command | Mean [s] | Min [s] | Max [s] |
|:---|---:|---:|---:|
| before: ERC4626 inflation | 2.196 ± 0.082 | 2.118 | 2.282 |
| after: ERC4626 inflation | 2.152 ± 0.157 | 2.056 | 2.333 |
||||‎ |
| before: MiniVat product inequality | 10.410 ± 0.731 | 9.571 | 10.900 |
| after: MiniVat product inequality | 11.588 ± 3.852 | 9.267 | 16.035 |
||||‎ |
| before: DirectDivision | 5.825 ± 0.318 | 5.476 | 6.100 |
| after: DirectDivision | 5.239 ± 1.077 | 4.225 | 6.369 |

Takeaway: no obvious regression on ERC4626 / DirectDivision; MiniVat is noisy locally and slightly slower on mean, but same order of magnitude and still returns the deterministic small witness. The performance-specific review fixes should also improve cache/allocation behavior for normal solver paths: fewer full-tree normalization clones per `is_sat`, no SMT string serialization just to detect residual `UDiv`, and no local fallback work on non-hard-arith model queries.

grandizzy (approved):

PR #14853, opened 2026-05-20
feat: add configurable collision-free EVM edge coverage

0xalpharush:
## Motivation

in the OG implementation, it is possible for edge IDs to collide which can lose some signal

OSS-249

## Solution

implement and make `evm_edge_coverage_collision_free` the default

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #14902, opened 2026-05-25
feat(invariant): gas-envelope fuzzing via invariant.gas_fuzz

grandizzy:
OSS-185

## Motivation

Foundry's invariant runner currently treats `tx.gas_limit` and `tx.gasprice`
as fixed environmental constants. Bugs that only surface when an inner call
runs out of gas — swallowed-OOG state corruption, single-selector gas DoS,
half-applied multi-step writes — are unreachable from the invariant
harness today.

This PR adds a single opt-in flag, `invariant.gas_fuzz`, that turns on:

1. **Per-call `tx.gas_limit` randomization** via a razor-power-law sampler
   centered on the observed natural cost of each `(target, selector)`.
2. **Per-call `tx.gasprice` randomization** (calibration-free).
3. **Per-selector peak-gas tracking** + a colored **"Max Gas"** column in
   the metrics table, with thresholds tied to the configured block gas
   limit.

The strategy deliberately starves calls just below their observed natural
cost so user invariants about post-conditions catch the divergence when the
sub-step OOGs but the outer call proceeds.

### New config (default: `false`, no behavior change for existing tests)

```toml
[invariant]
gas_fuzz = true
```

> Setting `gas_fuzz = true` implicitly enables the metrics table display
> (independent of `show_metrics`) because the "Max Gas" column is the
> primary surface for the feature.

### New metrics column (only when `gas_fuzz = true`)

`Max Gas` is appended to the existing `Calls | Reverts | Discards` table.
Cell color is driven by the share of the block gas limit consumed by a
single call:

| Color  | Threshold                | Meaning                               |
| ------ | ------------------------ | ------------------------------------- |
| cyan   | < 25% of block gas limit | safe                                  |
| yellow | ≥ 25%                    | warning — investigate                 |
| red    | ≥ 50%                    | DoS risk — single tx eats the block   |

image

### New JSON fields (omitted when zero)

`InvariantMetrics` gains `max_gas: u64` and `max_gas_sequence:
Vec<BasicTxDetails>` so `--json` consumers can pipe peak gas + a
reproducer sequence into dashboards.

## How the sampler works

`crates/evm/fuzz/src/strategies/gas_sampler.rs`:

- `GasObservations`: campaign-scoped `(target, selector) -> max gas_used`
  tracker, `Arc<RwLock<HashMap>>`.
- `sample_gas_limit` — razor-power-law (k=4):
  - 15% natural (no override; also keeps the observation tracker honest).
  - 80% in `[50% · max, max)` biased toward `max`.
  - 5%  in `[10% · max, 30% · max)` for griefing / DoS paths.
  - Floors at 50k gas; skips selectors whose observed max is < 100k.
- `sample_gas_price` — 40% zero / 40% `[1, 1e10)` / 20% `[1e10, 1e12)` wei.

Only **natural-gas** runs (no override) feed the observation tracker, so
truncated-OOG `gas_used` doesn't poison the running max.

## Comparison with Echidna and Medusa

> Note: the genuinely new bit is the adaptive, campaign-scoped
> `(target, selector) → max gas_used` observation feeding a razor-power-law
> sampler in `[50%, 100%)` of natural cost — i.e. driving execution into
> the OOG razor on purpose. Both Echidna and Medusa expose only static
> gas-limit configuration; Echidna additionally randomizes gas price up to
> a static cap, but neither adapts gas to per-selector behaviour or
> persists per-call gas in the corpus for replay.

| Capability                                                   | Foundry (this PR)                          | Echidna                            | Medusa                                  |
| ------------------------------------------------------------ | ------------------------------------------ | ---------------------------------- | --------------------------------------- |
| Per-call `tx.gas_limit` randomization                        | ✅ adaptive razor-power-law sampler centered on observed per-(target,selector) peak | ❌ static configured tx gas limit | ❌ static configured `blockGasLimit` / `transactionGasLimit` |
| Per-call `tx.gasprice` randomization                         | ✅ 40/40/20, calibration-free, persisted in `CallDetails` for replay | ✅ random gas price up to static `maxGasprice` cap; not adaptive, not persisted per-call for replay | ❌                                       |
| Per-selector peak-gas tracking                               | ✅ "Max Gas" column (per (target,selector), campaign-scoped, color-coded) | ⚠️ historical `estimateGas` was experimental and removed in 2.2.7 | ⚠️ available in traces, not aggregated  |
| Actively drives execution to the OOG razor                   | ✅ (the point of this PR)                  | ❌ (passive observer)               | ❌ (passive observer)                    |
| Catches swallowed-OOG state-corruption bugs                  | ✅                                         | ❌ unless full tx reverts          | ❌ unless full tx reverts                |


crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -1506,9 +1581,37 @@ pub(crate) fn execute_tx<FEN: FoundryEvmNetwork>(
...
+
+    // Per-call `tx.gasprice` override under `gas_fuzz`. The cheatcodes inspector
+    // consumes this one-shot during `initialize_interp` (via `.take()`), so the
+    // executor's natural price applies to any follow-up invariant assertion call
+    // even if the sampled price is non-zero.
+    if let Some(p) = tx.call_details.gas_price
+        && let Some(cheats) = executor.inspector_mut().cheatcodes.as_mut()
+    {
+        cheats.gas_price = Some(p);
+    }
```
mablr: It runs on a cloned inspector, so the original `cheats.gas_price` is never taken; `Executor::commit` then re-arms it from `tx_env.gas_price()`. Follow-up invariant call observes the sampled handler price; on no-commit paths a stale value may leak to a later tx with `gas_price: None`?

crates/forge/src/runner.rs
```diff
@@ -1757,6 +1757,8 @@ fn base_counterexamples_to_txes(
                     target: seq.addr.unwrap_or_default(),
                     calldata: seq.calldata.clone(),
                     value: seq.value,
+                    gas_limit: None,
+                    gas_price: None,
```
mablr: Unsure gas-fuzz failure can be replayed from cache

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -338,6 +359,11 @@ impl InvariantTest {
                 invariant_metrics.discards += 1;
             } else if reverted {
                 invariant_metrics.reverts += 1;
+            } else if gas_used > invariant_metrics.max_gas {
+                invariant_metrics.max_gas = gas_used;
+                if let Some(inputs) = run_inputs {
+                    invariant_metrics.max_gas_sequence = inputs.to_vec();
+                }
```
mablr: On monotonic-gas handlers (array-growth, payout loops, the exact targets here) every successful call is a strict new max, so the full prefix is cloned each call.

crates/evm/fuzz/src/lib.rs
```diff
@@ -81,6 +81,19 @@ pub struct CallDetails {
...
+    /// at the exact gas budget that triggered the failure.
+    /// Uses `#[serde(default)]` for backwards compatibility with existing corpus files.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub gas_limit: Option<u64>,
+    /// Optional override for `tx.gasprice`. Same lifecycle as `gas_limit`: set by the
+    /// gas-envelope sampler when `invariant.gas_fuzz = true`, `None` means default,
+    /// persisted for deterministic replay of gas-price-dependent failures.
+    /// Uses `#[serde(default)]` for backwards compatibility with existing corpus files.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub gas_price: Option<u128>,
```
figtracer: persisted corpus JSON with `gas_price: u128` fails to load (`u128 is not supported`)

crates/forge/src/cmd/test/summary.rs
```diff
@@ -147,13 +148,23 @@ pub(crate) fn format_invariant_metrics_table(
         table.apply_modifier(UTF8_ROUND_CORNERS);
     }
 
-    table.set_header(vec![
+    // Show "Max Gas" only when populated (i.e. `invariant.gas_fuzz = true`).
+    let show_max_gas = test_metrics.values().any(|m| m.max_gas > 0);
```
figtracer: `Max Gas` is hidden when `gas_fuzz = true` but all calls revert/discard

0xalpharush:
Can you post results of a bench run?

crates/evm/evm/src/executors/invariant/mod.rs
decofe: Fixed in https://github.com/foundry-rs/foundry/commit/56d0bcc10a685b5692757d225f31dbd2824d9413. The sampled gas price is now cleared after committing the gas-fuzzed handler call, with a regression asserting invariant calls still observe the default gas price.

crates/forge/src/runner.rs
decofe: Fixed in https://github.com/foundry-rs/foundry/commit/99d33b04aca55be89bd26e7d0d64eaff7ff46f25. The replay envelope now persists/restores optional gas_limit and gas_price fields on invariant counterexample steps, with a persisted-replay regression.

crates/evm/evm/src/executors/invariant/mod.rs
decofe: Fixed in https://github.com/foundry-rs/foundry/commit/906d91fb620ac332f58eaadbbfd063463674c001. The hot loop now tracks run id + prefix length and materializes max_gas_sequence once during metrics finalization, avoiding repeated full-prefix clones.

crates/forge/src/cmd/test/summary.rs
decofe: Fixed in https://github.com/foundry-rs/foundry/commit/6de87077cc3af9943ebe370a198624056847f781. The metrics table now receives an explicit gas_fuzz/show_max_gas signal, and a unit regression covers gas_fuzz enabled with all max_gas values at zero.

crates/evm/fuzz/src/lib.rs
decofe: Fixed in https://github.com/foundry-rs/foundry/commit/1eab6c2ba3bfccf62f08dd9bb34010ab8cfd7de3. Persisted/corpus gas_price is now u64, sample_gas_price returns u64, and we only cast to u128 when applying the cheatcode gas price. Added a JSON serialize/deserialize regression.

grandizzy:
# Benchmark: 1h Aave v4 SCFuzzBench, PR #14902 `gas_fuzz` overhead check

Comparing throughput, coverage, and bug discovery across configurations on the [Recon-Fuzz/aave-v4-scfuzzbench](https://github.com/Recon-Fuzz/aave-v4-scfuzzbench) suite (`forge test --mc CryticToFoundry`, `FOUNDRY_INVARIANT_TIMEOUT=3600`, all runs on the same host).

## Methodology note

After #14844 merged "Group invariant campaigns by test contract", a single `forge test --mc CryticToFoundry` invocation now runs **one** merged campaign covering all 10 invariants instead of 10 independent campaigns. To recover comparable CPU effort (~10 CPU-hours/run), runs fan out **10 parallel forge processes**, each in its own scratch copy. To control for cross-seed variance, the OFF-mode comparison was run twice with two disjoint seed sets (42–51 and 52–61).

aave-v4-scfuzzbench commit: `e54577dc1`. PR head: `4661bf390`.

## Throughput & coverage

| Run | seeds | total txs | tx/s agg | edges (avg/seed) | peak RSS |
|---|---|---|---|---|---|
| **#14853 hash+depth** *(old, ref)* | — | 32.16M | 8938 per-stream | 371–409 | 4.19 GB (single proc) |
| **master `0659e1dbc`** | 42–51 | 30.92M | 8594 | ~375 | 1.35 GB/seed |
| **master `0659e1dbc`** | 52–61 | 30.56M | 8494 | ~370 | 1.35 GB/seed |
| **PR #14902 `gas_fuzz=OFF`** | 42–51 | 29.45M | 8188 | ~375 | 1.33 GB/seed |
| **PR #14902 `gas_fuzz=OFF`** | 52–61 | 28.47M | 7915 | ~370 | 1.33 GB/seed |
| **PR #14902 `gas_fuzz=ON`** | 42–51 | 30.45M | 8464 | ~442 | 1.35 GB/seed |

Paired Δ vs master on identical seeds: −4.7% (42–51), −6.8% (52–61). Average ~−5.8% throughput regression on the OFF default path.

## Bugs found (union of `[FAIL]` lines across seeds)

| Run | seeds | invariants | handlers | **total** |
|---|---|---|---|---|
| #14853 hash+depth | — | 5 | 3 | **8** |
| master | 42–51 | 4 | 4 | **8** |
| master | 52–61 | 3 | 3 | **6** |
| PR `gas_fuzz=OFF` | 42–51 | 5 | 4 | **9** |
| PR `gas_fuzz=OFF` | 52–61 | 4 | 5 | **9** |
| PR `gas_fuzz=ON` | 42–51 | 3 | 3 | **6** |

Cross-batch variance is substantial on master (8 → 6) and on OFF (9 → 9). The bug deltas should be read as *PR finds more bugs reliably across both batches in OFF mode*, not as a precise count.

## Max Gas distribution (gas_fuzz=ON, seed42 representative)

Top-10 selectors by observed Max Gas:

| Selector | Calls | Reverts | Max Gas |
|---|---|---|---|
| `iSpoke_borrow` | 153,762 | 150,269 (97.7%) | **581,631** |
| `iAaveOracle_setPrice` | 213,164 | 67,067 (31.5%) | 548,424 |
| `iSpoke_withdraw_ASSERTION_WITHDRAW_DOS` | 146,769 | 141,050 | 497,974 |
| `iSpoke_updateUserDynamicConfig` | 132,692 | 10,019 | 418,545 |
| `iSpoke_updateUserRiskPremium` | 122,980 | 9,995 | 412,155 |
| `iSpoke_setUsingAsCollateral` | 170,449 | 28,049 | 410,520 |
| `iSpoke_supply_ASSERTION_SUPPLY_DOS` | 140,840 | 78,945 | 404,025 |
| `iHub_mintFeeShares_ASSERTION_MINT_FEE_SHARES_PPS_CHANGE` | 135,250 | 123,286 | 363,248 |
| `iSpoke_repay_ASSERTION_REPAY_DOS` | 112,087 | 111,571 | 343,610 |
| `iHub_setInterestRateData` | 131,475 | 120,318 | 313,611 |

Two selectors registered **Max Gas = 0** (zero successful execution):

| Selector | Calls | Reverts |
|---|---|---|
| `iHub_updateAssetConfig` | 144,742 | 144,742 (100%) |
| `iSpoke_liquidationCall_ASSERTION_LIQUIDATION_CALL_DOS` | 158,244 | 158,244 (100%) |

Both are scfuzzbench harness artifacts, not foundry issues:
- `iSpoke_liquidationCall_*_DOS` uses a `try { ... } catch { require(false) }` pattern by design — every call enters the catch and discards.
- `iHub_updateAssetConfig` is an `asAdmin` config-mutation handler whose `InterestRateData` struct input isn't being generated in a valid shape.

The Max Gas column surfacing these is a useful side benefit of the feature.

## Findings

1. **`gas_fuzz=ON` delivers a large coverage uplift.** Edges jump to ~442 avg (vs ~375 OFF — **+18%**), aggregate throughput rises +3.4% over OFF. Per-process peak RSS unchanged.

2. **OFF-mode regression is real and reproduces on two seed sets.** Average ~−5.8% throughput vs master on the default path. An earlier revision (`9301b445b`) widened `CallDetails` by two `Option<u64>` (+32 bytes), copied into every `BasicTxDetails`. Commit `4661bf390` boxes the overrides behind `Option<Box<GasOverrides>>` (8 bytes inline, niche-optimized, allocation-free when OFF, JSON shape preserved via `serde(flatten)`). That recovered the per-tx-layout portion of the regression; a residual ~5% gap remains and is **mode-agnostic** (ON mode shows the same magnitude drop vs the pre-fix ON measurement), suggesting a campaign-scoped cost rather than per-tx. Initial bisection toward `record_metrics` was inconclusive — the residual source has not been localized.

3. **Bug-finding is *consistently* better on the PR.** OFF mode hit 9 unique bugs across both seed batches (42–51 and 52–61) vs master's 8 / 6. Both `v1` and `v2` variants of `totalBorrowedLessThanSupplied` plus all 4 handlers landed in both PR batches. *Caveat:* the PR changes RNG consumption order, so "seed 42" explores a different campaign in PR vs master — part of the bug delta may be sampling artifact rather than search-quality improvement. The trend is consistent across two seed sets, which is suggestive but not statistically tight.

4. **`gas_fuzz=ON` finds 6 unique bugs — fewer than OFF's 9.** ON misses `invariant_totalBorrowedLessThanSupplied_v1`, `_v2`, and `iSpoke_repay_ASSERTION_REPAY_DOS` — all "deep state" bugs requiring multi-call setup chains. The wider gas envelope trades depth for breadth: more code paths attempted (+18% edges), fewer deep sequences completed before OOG. This reproduces identically on the boxed-fix binary, so it's a behavior of the gas-envelope sampling itself, not the plumbing. Worth a follow-up to understand whether the new distribution is hiding these bugs or just pushing their discovery threshold past 1h.

5. **The Max Gas column delivers value beyond bug-finding.** Surfaces two handlers producing zero successful executions — a harness-quality signal invisible from `Calls/Reverts/Discards` alone.

## Conclusion

The feature delivers a substantial coverage uplift (+18% edges) and consistent bug-finding gains on the suite tested, with the cost of a ~5% throughput regression on the default OFF path. The OFF-mode per-tx layout cost has been diagnosed and fixed in `4661bf390`; a residual campaign-scoped ~5% gap remains unattributed and affects both modes equally. Recommended path: merge with the opt-in flag as drafted (no behavioral surprise for existing users), file the residual gap and the ON-mode `v1`/`v2` starvation as follow-ups.

0xalpharush:
> The feature delivers a substantial coverage uplift (+18% edges) 

Would be nice to know specifically the area of code that is covered using lcov-diff

My suggestion would to do a 2nd experiment, setting  gaslimit to 2^24 (https://eips.ethereum.org/EIPS/eip-7825) by default, and adding a mutator that does `rng.range(tx.gaslimit/2, tx.gaslimit*2).max( 2^24)`. If we can avoid calibration and having an additional config w/o perf hit, I think that'd be simpler.

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -1532,9 +1661,43 @@ pub(crate) fn execute_tx<FEN: FoundryEvmNetwork>(
     let requested_value = tx.call_details.value.unwrap_or(U256::ZERO);
     let sender_balance = executor.get_balance(tx.sender)?;
     let value = requested_value.min(sender_balance);
-    executor
-        .call_raw(tx.sender, tx.call_details.target, tx.call_details.calldata.clone(), value)
-        .map_err(|e| eyre!(format!("Could not make raw evm call: {e}")))
+
+    // Per-call gas-limit override under `gas_fuzz`. Save/restore so the executor's
+    // natural limit applies to invariant assertion calls.
+    let saved_gas_limit = tx.call_details.gas_limit().map(|g| {
```
figtracer: this gas-price override is only cleared in the main invariant loop after commit, but `execute_tx` is also used by persisted replay, shrinking, `replay_run`, corpus replay, and showmap. 

in those paths `Executor::commit` re-arms `cheats.gas_price` from `result.tx_env.gas_price()`, so a gas-fuzzed handler price can leak into the next tx or invariant assertion. 

i reproduced this by persisting a one-call state failure with `gas_price=5278982079`, changing the invariant to only assert `tx.gasprice == 0` with `runs = 0`, and seeing persisted replay still fail before a new campaign

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -584,11 +683,19 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
                 // assumes / pops below) get zero-length entries that the corpus side filters out.
                 let call_cmp_values = call_result.evm_cmp_values.take().unwrap_or_default();
                 let discarded = call_result.result.as_ref() == MAGIC_ASSUME;
-                if self.config.show_metrics {
+                if self.config.show_metrics || self.config.gas_fuzz {
```
mablr: `gas_fuzz=true` implies `show_metrics`, intentional, but should be called out in comment?

crates/evm/fuzz/src/strategies/gas_sampler.rs
```diff
@@ -0,0 +1,200 @@
...
+                    hits_edge += 1;
+                }
+            }
+        }
+        assert!(total > 0);
+        // Edge band is ~10% of max. With 40% of samples in the [50%, 110%] range
+        // (uniform), the [60%, 70%] sub-band should get ~6-7% of samples.
+        let pct = hits_edge * 100 / total;
+        assert!(pct >= 3, "edge band hit rate too low: {pct}% ({hits_edge}/{total})");
+    }
```
mablr, line 37: It seems there is a discrepancy between what's commented here and the actual distribution [L74-L77](https://github.com/foundry-rs/foundry/pull/14902/changes#diff-00704b1b489f6c587e972688dc545a2e1565c3e314504bdb1238d64ab2a70694R74-R77)


With `powi(4)` the distribution heavily front-loads near max (~32% within 1% of max, ~51% within 7%); the [60%, 70%] sub-band sits in a low-density region and hits at roughly 3-5% of sampled calls, hence the conservative >= 3% floor?

crates/evm/evm/src/executors/invariant/mod.rs
grandizzy: good catch, fixed in https://github.com/foundry-rs/foundry/pull/14902/changes/aac8d91cfa716deee39048bd052b94ef49df1429 + regression test, ptal

grandizzy:
> My suggestion would to do a 2nd experiment, setting gaslimit to 2^24 (https://eips.ethereum.org/EIPS/eip-7825) by default, and adding a mutator that does rng.range(tx.gaslimit/2, tx.gaslimit*2).max( 2^24). If we can avoid calibration and having an additional config w/o perf hit, I think that'd be simpler.

@0xalpharush pls check https://github.com/foundry-rs/foundry/pull/14902/changes/1e0858818a6cf05d8bdc6f47de7af383f151da49 and comparison below, looks much cleaner

### Benchmarks — this PR vs master

Workload: scfuzzbench, 10 parallel 1h `forge test` runs per seed (`--fuzz-seed N`), aggregated per seed batch.

| run                                                  | mode | total txs | agg tx/s | avg edges | unique bugs |
| ---------------------------------------------------- | ---- | --------: | -------: | --------: | ----------: |
| master `0659e1dbc`, seeds 42–51                      | n/a  |    30.92M |     8594 |      ~375 |           8 |
| master `0659e1dbc`, seeds 52–61                      | n/a  |    30.56M |     8494 |      ~370 |           6 |
| **this PR HEAD, seeds 42–51**                        | **ON** |  **27.98M** | **7777** |  **~372** |       **9** |
| **this PR HEAD, seeds 52–61**                        | **ON** |  **28.45M** | **7910** |  **~371** |       **6** |

`unique bugs` = cumulative invariant + handler bugs surfaced across the 10 seeds in the batch.

### Findings

1. **Bug discovery.** PR ON beats master on seeds 42–51 (9 vs 8) and matches master on seeds 52–61 (6 vs 6).
2. **Throughput.** PR ON is −9.5% on 42–51 and −6.9% on 52–61 vs master — well inside the per-seed-batch noise seen in the master rows themselves.
3. **Edge coverage.** PR ON sits in master's range (~371–372 vs ~370–375).
4. **`gas_fuzz=OFF` path.** The design adds no campaign-scoped state, no per-call heap allocation, and no extra fields on `CallDetails` beyond a single `Option<u64>`, so the OFF path is the same hot path as master and was not re-measured.

### Conclusion

A single-knob design: when `invariant.gas_fuzz=true`, each call's `tx.gas_limit` is drawn uniformly in `[2^24, 2^25)`. No per-(target, selector) state, no calibration, no `gas_price` envelope, no boxed override struct — `CallDetails` carries one extra `Option<u64> gas_limit` field. Same or better bug discovery than master, no OFF-mode regression.

crates/evm/fuzz/src/strategies/gas_sampler.rs
grandizzy: good point, no longer applicable with the new approach in https://github.com/foundry-rs/foundry/pull/14902/changes/1e0858818a6cf05d8bdc6f47de7af383f151da49

crates/evm/evm/src/executors/invariant/mod.rs
grandizzy: yep, added in https://github.com/foundry-rs/foundry/pull/14902/changes/c7a839ed34511bece7c8e3ddd70de2c9f023aa50

crates/evm/fuzz/src/lib.rs
```diff
@@ -114,6 +119,12 @@ pub struct BaseCounterExample {
     /// Ether value sent with the call.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub value: Option<U256>,
+    /// Optional transaction gas limit override.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub gas_limit: Option<u64>,
+    /// Optional transaction gas price override.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub gas_price: Option<u64>,
```
mablr: `gas_price` seems always set to `None` after 1e0858818

grandizzy: oops, leftover from cleanup, removed in https://github.com/foundry-rs/foundry/pull/14902/changes/9588925b1eeb80538af93ba3d4c8e87576c0f64d

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -1532,9 +1630,27 @@ pub(crate) fn execute_tx<FEN: FoundryEvmNetwork>(
...
-        .call_raw(tx.sender, tx.call_details.target, tx.call_details.calldata.clone(), value)
-        .map_err(|e| eyre!(format!("Could not make raw evm call: {e}")))
+
+    // Per-call gas-limit override under `gas_fuzz`. Save/restore so the executor's
+    // natural limit applies to invariant assertion calls.
+    let saved_gas_limit = tx.call_details.gas_limit.map(|g| {
+        let saved = executor.gas_limit();
+        executor.set_gas_limit(g);
+        saved
+    });
```
mablr: This seems to apply `Some(gas_limit)`, but the loop only stamps it when `gas_fuzz=true`. Since `CallDetails::gas_limit` is persisted in the corpus, an old gas-fuzzed corpus may keep applying overrides even after turning the flag off?

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -619,7 +717,6 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
                         break 'stop;
```
mablr, line 995: The timeout path was patched L621-L627, but here the `MaxAssumeRejects` exits without calling `attach_max_gas_inputs`. `finalize_metrics` then keeps `max_gas` but emits an empty `max_gas_sequence`. Prob needs to be patched as well.

crates/config/src/invariant.rs
```diff
@@ -49,6 +49,15 @@ pub struct InvariantConfig {
     ///
     /// Example: `check_interval = 10` means assert after calls 10, 20, 30, ... and the last call.
     pub check_interval: u32,
+    /// Enable gas-envelope fuzzing and per-selector gas reporting.
+    /// Default `false` (no behavior change, no new output).
+    ///
+    /// Turns on three things at once:
+    /// 1. Per-call `tx.gas_limit` / `tx.gasprice` randomization, biased toward the OOG /
```
mablr: `gasprice` was removed https://github.com/foundry-rs/foundry/commit/9588925b1eeb80538af93ba3d4c8e87576c0f64d, doc should be updated.

grandizzy: updated in https://github.com/foundry-rs/foundry/pull/14902/changes/436e2e35bc6616d113a7a7379b2fc468933f90f6

crates/evm/evm/src/executors/invariant/mod.rs
grandizzy: yep, added in https://github.com/foundry-rs/foundry/pull/14902/changes/436e2e35bc6616d113a7a7379b2fc468933f90f6

crates/evm/evm/src/executors/invariant/mod.rs
grandizzy: pls check https://github.com/foundry-rs/foundry/pull/14902/changes/a45740ee391747ebfe88d347cc6bef3b1aee7e2a

mablr (dismissed):
lgtm !

pending @0xalpharush :)

crates/evm/fuzz/src/strategies/invariants.rs
```diff
@@ -176,6 +176,6 @@ pub fn fuzz_contract_with_calldata<S: FuzzStateReader>(
 
     (calldata_strategy, value_strategy).prop_map(move |(calldata, value)| {
         trace!(input=?calldata, ?value);
-        CallDetails { target, calldata, value }
+        CallDetails { target, calldata, value, gas_limit: None }
```
mablr, line 179: `gas_fuzz` never stamps generated calls with a sampled gas limit. `invariant_strat` receives config, but here we return always `CallDetails { ..., gas_limit: None }`, and `sample_gas_limit` is only referenced by its unit test. As written, enabling `invariant.gas_fuzz` adds reporting, but does not randomize `tx.gaslimit()`. Let's thread `config.gas_fuzz` into the call-details strategy and set `Some(sample_gas_limit(...))` only when enabled?

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -479,6 +535,20 @@ impl InvariantTest {
...
 }
 
+fn finalize_metrics(metrics: Map<String, InvariantMetricState>) -> Map<String, InvariantMetrics> {
+    metrics
+        .into_iter()
+        .map(|(key, mut state)| {
+            if let Some(max_gas_run) = state.max_gas_run
+                && let Some(inputs) = max_gas_run.inputs
+            {
+                state.metrics.max_gas_sequence = inputs[..max_gas_run.prefix_len].to_vec();
```
figtracer, line 545: this panics after `drop_corpus_payloads()` clears `current_run.inputs`: `prefix_len` can still be > 0 while `inputs` is empty. the targeted gas cli tests hit this path. please attach/materialize the max-gas prefix before clearing the run payloads.

figtracer (commented):
ran `cargo +1.95.0 test -p forge --test cli test_cmd::invariant::gas -- --nocapture`; two failures below.

crates/evm/evm/src/executors/invariant/campaign.rs
```diff
@@ -716,9 +717,15 @@ mod tests {
         assert_eq!(result.last_run_inputs[0].sender, Address::repeat_byte(0x30));
 
         let transfer_metrics = result.metrics.get("transfer(address)").unwrap();
-        assert_eq!(transfer_metrics, &InvariantMetrics { calls: 4, reverts: 1, discards: 2 });
+        assert_eq!(
+            transfer_metrics,
+            &InvariantMetrics { calls: 4, reverts: 1, discards: 2, ..Default::default() }
```
figtracer, line 722: this default hides that the aggregator now drops `max_gas` and `max_gas_sequence`: `merge_metrics` only sums calls/reverts/discards. please seed distinct max-gas values here and assert the merged result keeps the largest peak and its sequence, so multi-worker `gas_fuzz` doesn't lose the new fields.

PR #14905, opened 2026-05-25
feat: prepare invariant campaigns for sharding

0xKarl98:
On first step for OSS-273 , 

Adds the minimal structure needed to treat an invariant contract as one logical campaign that can later be split across workers.

The execution path is still single-worker. This PR only introduces the campaign/worker/result boundary so future PRs can partition runs across `--jobs` workers and merge worker outputs back into one contract-level invariant result without changing user-facing behavior

cc @grandizzy @mattsse @figtracer  @mablr


crates/evm/evm/src/executors/invariant/campaign.rs
```diff
@@ -0,0 +1,320 @@
...
+                    && run.id.seed == output.plan.seed
+                    && run.id.global_run == output.plan.first_global_run + run.id.worker_run
+            }));
+            debug_assert!(output.failures.iter().all(|failure| output.plan.contains(failure.id)));
+        }
+        output.result
+    }
+}
+
+#[cfg(test)]
```
figtracer, line 100: we should gate tests with `cfg(all(test, debug_assertions))` otherwise `cargo test --release` won't compile

crates/evm/evm/src/executors/invariant/campaign.rs
```diff
@@ -0,0 +1,320 @@
...
+    spec: InvariantCampaignSpec,
+    output: Option<InvariantWorkerOutput>,
+}
+
+impl InvariantCampaignAggregator {
+    pub const fn new(spec: InvariantCampaignSpec) -> Self {
+        Self { spec, output: None }
+    }
+
+    pub fn push(&mut self, output: InvariantWorkerOutput) {
```
figtracer: in non-debug builds, a second `push` silently overwrites the first worker output; maybe we could error here unless this is desired

crates/evm/evm/src/executors/invariant/error.rs
```diff
@@ -266,7 +273,23 @@ impl InvariantFailures {
...
+        invariant: &Function,
+        failure: InvariantFuzzError,
+        run_id: Option<InvariantRunId>,
+    ) {
+        let key = FailureKey::Invariant(invariant.name.clone());
+        let prev = self.failures.insert(key.clone(), failure);
+        #[cfg(debug_assertions)]
+        if let Some(run_id) = run_id {
+            self.failure_sources.insert(key, run_id);
+        }
```
mablr: Unsure about the behavior with duplicate invariant key.

crates/evm/evm/src/executors/invariant/error.rs
```diff
@@ -303,12 +326,25 @@ impl InvariantFailures {
...
-                FailureKey::Handler(site.0, site.1),
-                InvariantFuzzError::HandlerAssertion(failure),
-            );
+            let key = FailureKey::Handler(site.0, site.1);
+            let prev =
+                self.failures.insert(key.clone(), InvariantFuzzError::HandlerAssertion(failure));
+            #[cfg(debug_assertions)]
+            if let Some(run_id) = run_id {
+                self.failure_sources.insert(key, run_id);
+            }
```
mablr: same as above

crates/evm/evm/src/executors/invariant/campaign.rs
```diff
@@ -0,0 +1,320 @@
...
+    pub worker_id: u32,
+    pub first_global_run: u32,
+    pub runs: u32,
+    pub seed: Option<U256>,
+}
+
+impl InvariantWorkerPlan {
+    pub const fn run_id(&self, worker_run: u32) -> InvariantRunId {
+        InvariantRunId::new(
+            self.first_global_run + worker_run,
```
mablr: Unlikely, but there is a possible overflow here `u32` <- `u32` + `u32`

crates/evm/evm/src/executors/invariant/campaign.rs
0xKarl98: But this is the first pr  , think there won't be second push ? 
We can do that in later PRs

figtracer: ack! we can take a look at it afterwards

crates/evm/evm/src/executors/invariant/error.rs
0xKarl98: This is intended to be last-write-wins behaviour

mablr: okay thx for clarification

0xKarl98:
@mablr @figtracer have addressed your concerns , ptal again

crates/evm/evm/src/executors/invariant/campaign.rs
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    pub fn finish(self) -> InvariantFuzzTestResult {
+        let output = self.output.expect("at least one invariant worker output");
+        debug_assert_eq!(output.plan.runs, self.spec.total_runs);
+        debug_assert_eq!(output.plan.seed, self.spec.seed);
+        #[cfg(debug_assertions)]
+        {
+            debug_assert!(output.runs.iter().enumerate().all(|(expected, run)| {
+                run.id.global_run == expected as u32
```
figtracer: this assumes every worker starts at global run 0, so future sharded workers with `first_global_run > 0` will fail debug validation

0xKarl98: Yeah we better prepare for this in this first pr

mattsse (dismissed):
LGTM

grandizzy (changes requested):
since we're here let's first cleanup the suite test outcome, e.g. when running with show-progress there's still the invariant anchor which is no longer needed, see below - there are multiple invariant fns defined in each contract (see https://github.com/grandizzy/invariant-output to test / try)

Image

Then same on output, the `invariant_stakingSolvent` should be replaced by contract name maybe? 

Image

Also, the BuggyBank contract with 4 invariants, designed with 3 passing and one failed:
- ✅ invariant_actorsRegistered
- ✅ invariant_bankHoldsNoEth
- ✅ invariant_countersMatchHandlerActivity
- ❌ invariant_ledgerConserved (fails by design)

the output is misleading and quite hard to interpret as all invariants are marked failed. It also shows the invariant table metrics several times instead only one time as in case of success.

image

crates/evm/evm/src/executors/invariant/campaign.rs
```diff
@@ -0,0 +1,237 @@
...
+
+impl InvariantWorkerOutput {
+    pub const fn new(plan: InvariantWorkerPlan, result: InvariantFuzzTestResult) -> Self {
+        Self { plan, result }
+    }
+}
+
+/// Merges worker outputs back into one logical invariant campaign result.
+///
+/// PR1 keeps the execution model single-worker while making the campaign/result boundary explicit.
```
grandizzy: nit: maybe add TODO instead specific PR1?

grandizzy (dismissed):
lgtm!

crates/evm/evm/src/executors/invariant/campaign.rs
0xKarl98: Makes sense

grandizzy (approved):

mablr (approved):
Nice

PR #14907, opened 2026-05-25
feat(lint): add missing events arithmetic lint

figtracer:
## Summary
- add low-severity `missing-events-arithmetic` lint for access-controlled arithmetic parameter updates without events
- keep the rule conservative to reduce false positives: skips fixed writes, constructors, unprotected setters, mappings/arrays, protected-only arithmetic reads, unrelated event emission, non-access modifiers, and superficial sender checks
- add lint docs and UI coverage for warning and suppression cases

### DIFF
--- a/crates/lint/src/sol/low/missing_events_arithmetic.rs
+++ b/crates/lint/src/sol/low/missing_events_arithmetic.rs
@@ -164,7 +164,7 @@ struct WriteAnalyzer<'a, 'hir> {
 }
 
 impl<'a, 'hir> WriteAnalyzer<'a, 'hir> {
-    fn new(hir: &'hir hir::Hir<'hir>, targets: &'a HashSet<VariableId>) -> Self {
+    const fn new(hir: &'hir hir::Hir<'hir>, targets: &'a HashSet<VariableId>) -> Self {
         Self { hir, targets, call_stack: Vec::new() }
     }
 
--- a/crates/lint/src/sol/low/missing_events_arithmetic.rs
+++ b/crates/lint/src/sol/low/missing_events_arithmetic.rs
@@ -8,7 +8,7 @@ use solar::{
     interface::{Span, kw, sym},
     sema::hir::{
         self, BinOpKind, ElementaryType, ExprKind, FunctionId, ItemId, Res, StmtKind, TypeKind,
-        VariableId,
+        UnOpKind, VariableId,
     },
 };
 use std::collections::{HashMap, HashSet};
@@ -44,9 +44,7 @@ impl<'hir> LateLintPass<'hir> for MissingEventsArithmetic {
 
         for func_id in contract.all_functions() {
             let func = hir.function(func_id);
-            if !is_protected_entry_point(hir, func_id, func)
-                || function_emits_event(hir, func_id, &mut HashSet::new())
-            {
+            if !is_protected_entry_point(hir, func_id, func) {
                 continue;
             }
 
@@ -139,31 +137,49 @@ struct StateWrite {
     span: Span,
 }
 
+#[derive(Clone, Default)]
+struct WriteState {
+    taint: HashMap<VariableId, HashSet<VariableId>>,
+    pending_writes: Vec<StateWrite>,
+    emitted_event: bool,
+}
+
+impl WriteState {
+    fn record_write(&mut self, write: StateWrite) {
+        if !self.emitted_event {
+            self.pending_writes.push(write);
+        }
+    }
+
+    fn record_event(&mut self) {
+        self.emitted_event = true;
+        self.pending_writes.clear();
+    }
+}
+
 struct WriteAnalyzer<'a, 'hir> {
     hir: &'hir hir::Hir<'hir>,
     targets: &'a HashSet<VariableId>,
-    taint: HashMap<VariableId, HashSet<VariableId>>,
-    writes: Vec<StateWrite>,
     call_stack: Vec<FunctionId>,
 }
 
 impl<'a, 'hir> WriteAnalyzer<'a, 'hir> {
     fn new(hir: &'hir hir::Hir<'hir>, targets: &'a HashSet<VariableId>) -> Self {
-        Self { hir, targets, taint: HashMap::new(), writes: Vec::new(), call_stack: Vec::new() }
+        Self { hir, targets, call_stack: Vec::new() }
     }
 
     fn analyze_entry_point(&mut self, func_id: FunctionId) -> Vec<StateWrite> {
+        let mut state = WriteState::default();
         let func = self.hir.function(func_id);
-        self.taint.clear();
         for &param in func.parameters {
-            self.taint.insert(param, HashSet::from([param]));
+            state.taint.insert(param, HashSet::from([param]));
         }
 
-        self.analyze_function(func_id);
-        std::mem::take(&mut self.writes)
+        self.analyze_function(func_id, &mut state);
+        state.pending_writes
     }
 
-    fn analyze_function(&mut self, func_id: FunctionId) {
+    fn analyze_function(&mut self, func_id: FunctionId, state: &mut WriteState) {
         if self.call_stack.contains(&func_id) {
             return;
         }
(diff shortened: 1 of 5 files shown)


mablr (commented):
Here are some curated false negatives found with AI (reproduced locally)

### FN1: Branch-conditional emit suppresses all diagnostics for the function

**Location:** [`L47-L49`](https://github.com/foundry-rs/foundry/blob/7c44454de700809231c47aa6788a983a44cb7f15/crates/lint/src/sol/low/missing_events_arithmetic.rs#L47-L49)

`function_emits_event` is a whole-function predicate. One `emit` anywhere causes the entire entry-point to be skipped, so a write on a branch that never emits is silently ignored.

**Repro:**
```solidity
function setConditionallyEmittedValue(uint256 newValue, bool withEvent) external onlyOwner {
    if (withEvent) {
        conditionallyEmittedValue = newValue;
        emit ConditionallyEmittedValueUpdated(newValue);
    } else {
        conditionallyEmittedValue = newValue; // ← no warning, should warn
    }
}
```

---

### FN2: Compound assignment with a literal RHS is not recorded as a state write

**Location:** [`L234-L240`](https://github.com/foundry-rs/foundry/blob/7c44454de700809231c47aa6788a983a44cb7f15/crates/lint/src/sol/low/missing_events_arithmetic.rs#L234-L240)

`WriteAnalyzer::analyze_expr` only pushes a `StateWrite` when `sources` (input-tainted values from the RHS) is non-empty. For `x += 1` the RHS is a literal — no taint — so the write is dropped. The same path means `x += y` where `y` is also untainted (e.g. a state variable read directly) is equally missed.

**Repro:**
```solidity
function incrementSelf() external onlyOwner {
    selfIncrementedValue += 1; // ← no warning, should warn
}
```

---

### FN3: Prefix/postfix `++` / `--` are never recorded as state writes

**Location:** [`L269-L272`](https://github.com/foundry-rs/foundry/blob/7c44454de700809231c47aa6788a983a44cb7f15/crates/lint/src/sol/low/missing_events_arithmetic.rs#L269-L272)

`ExprKind::Unary(_, inner)` in `WriteAnalyzer::analyze_expr` just recurses into `inner` — it never checks whether the operator is `++`/`--` and never calls `state_lhs_vars` on the operand.

**Repro:**
```solidity
function prefixIncrement() external onlyOwner {
    ++prefixIncrementedValue; // ← no warning, should warn
}
```

---

### FN4: Wrong-polarity access guard is treated as protection, suppressing event-covered writes

**Location:** [`L866-L870`](https://github.com/foundry-rs/foundry/blob/7c44454de700809231c47aa6788a983a44cb7f15/crates/lint/src/sol/low/missing_events_arithmetic.rs#L866-L870)

`stmt_has_access_guard` matches any `if (<sender-vs-state-var>) { <exits> }` pattern without checking the comparison polarity. `if (msg.sender == owner) return` exits for the owner and lets everyone else through — the opposite of access control — but it satisfies the heuristic. Once the function is classified as "protected", the presence of any `emit` causes the whole function to be skipped.

**Repro:**
```solidity
// wrongPolarityValue is used in arithmetic by wrongPolarityQuote()
function wrongPolaritySetWithEvent(uint256 newValue) external {
    if (msg.sender == owner) return; // exits for owner, NOT a guard
    wrongPolarityValue = newValue;   // ← no warning, should warn
    emit ConditionallyEmittedValueUpdated(newValue);
}
```

mablr (commented):
Clanker unleashed 😄, he are some FN/FP findings (reproduced locally). Maybe not all worth addressing.

## False Negatives (lint misses a warning it should produce)

### 1. Unrelated emit before a write silences the warning

**Code:** [`record_write` / `record_event`, L147–156](https://github.com/foundry-rs/foundry/blob/b9159f1b752a72198a9e13b37f494f14fd82c2e2/crates/lint/src/sol/low/missing_events_arithmetic.rs#L147-L156)

`emitted_event` is a sticky boolean — once set it suppresses all subsequent writes in the function, even ones that have nothing to do with the event that was emitted.

```solidity
// No warning — Touched() is unrelated, but emitted_event is now true.
function setPrice(uint256 p) external onlyOwner {
    emit Touched();
    price = p; // ← should warn
}
```

---

### 2. `return` before an `emit` still lets the emit clear the write

**Code:** [`StmtKind::Return` arm, L253–257](https://github.com/foundry-rs/foundry/blob/b9159f1b752a72198a9e13b37f494f14fd82c2e2/crates/lint/src/sol/low/missing_events_arithmetic.rs#L253-L257)

`StmtKind::Return` is not treated as a path terminator. Processing continues, so the later emit clears `pending_writes` — including the write that already happened on the exiting path.

```solidity
// No warning — emit is unreachable on the skip=true path but still clears the write.
function setPrice(uint256 p, bool skip) external onlyOwner {
    price = p;        // ← should warn on the skip=true path
    if (skip) return;
    emit PriceUpdated(p);
}
```

---

### 3. `try` success-clause emit clears catch-clause write

**Code:** [`StmtKind::Try` arm, L238–245](https://github.com/foundry-rs/foundry/blob/b9159f1b752a72198a9e13b37f494f14fd82c2e2/crates/lint/src/sol/low/missing_events_arithmetic.rs#L238-L245)

All `try` clauses share a single mutable `WriteState`. An emit in the success clause calls `record_event()` which clears `pending_writes` for the entire state, so any silent write in the `catch` clause is never recorded.

```solidity
// No warning for catch path — success-clause emit cleared shared pending_writes.
function setPrice(uint256 fb) external onlyOwner {
    try oracle.getPrice() returns (uint256 p) {
        price = p;
        emit PriceUpdated(p);
    } catch {
        price = fb; // ← should warn
    }
}
```

---

### 5. Assigns from non-parameter sources are dropped

**Code:** [`analyze_expr` Assign arm, L267–274](https://github.com/foundry-rs/foundry/blob/b9159f1b752a72198a9e13b37f494f14fd82c2e2/crates/lint/src/sol/low/missing_events_arithmetic.rs#L267-L274)

A write is only recorded when the RHS carries parameter taint (`!sources.is_empty()`) or is a compound arithmetic assign. Assigns from state variables, block globals, or external call results carry no taint and are silently dropped.

```solidity
// No warnings — neither RHS has parameter taint.
function setPriceFromStateVar() external onlyOwner {
    price = referencePrice; // ← should warn
}
function setPriceFromTimestamp() external onlyOwner {
    price = block.timestamp; // ← should warn
}
```

---

### 6. Arithmetic use via helper return value is not propagated

**Code:** [`analyze_internal_call` in ArithmeticUseAnalyzer, L537–616](https://github.com/foundry-rs/foundry/blob/b9159f1b752a72198a9e13b37f494f14fd82c2e2/crates/lint/src/sol/low/missing_events_arithmetic.rs#L537-L616)

`ArithmeticUseAnalyzer` propagates taint only through call *parameters*, not return values. If a helper reads a state variable and returns it, and the caller uses that return value in arithmetic, the state variable is never added to `arithmetic_vars`.

```solidity
function _getRate() internal view returns (uint256) { return rate; }

// rate is never flagged as arithmetic-critical — helper return value is opaque.
function rateQuote(uint256 amount) external view returns (uint256) {
    return amount * _getRate(); // rate's involvement is invisible
}
function setRate(uint256 r) external onlyOwner {
    rate = r; // ← should warn
}
```

---

## False Positives (lint warns when it should not)

### 4. Event emitted in modifier body is invisible to the write analyzer

**Code:** [`analyze_entry_point`, L171–179](https://github.com/foundry-rs/foundry/blob/b9159f1b752a72198a9e13b37f494f14fd82c2e2/crates/lint/src/sol/low/missing_events_arithmetic.rs#L171-L179)

`WriteAnalyzer` only traverses the function body, not attached modifier bodies. A modifier that emits an event after `_` will cover every write in the function, but the lint never sees that emit.

```solidity
modifier emitAfter() { _; emit PriceUpdated(price); }

// Spurious warning — modifier emits the event, but the analyzer doesn't see it.
function setPrice(uint256 p) external onlyOwner emitAfter {
    price = p; // ← false positive
}
```

---

### 7. Non-dominating access guard still classifies function as protected

**Code:** [`stmt_has_access_guard`, L929–953](https://github.com/foundry-rs/foundry/blob/b9159f1b752a72198a9e13b37f494f14fd82c2e2/crates/lint/src/sol/low/missing_events_arithmetic.rs#L929-L953)

`function_has_access_guard` returns `true` if *any* statement in the body looks like a guard, regardless of whether it dominates the write. A conditional guard that doesn't cover all callers makes the function appear protected, so a missing emit generates a spurious warning.

```solidity
// Spurious warning — guard only applies when flag is true; anyone can bypass it.
function setPrice(uint256 p) external {
    if (flag) require(msg.sender == owner);
    price = p; // ← false positive
}
```

---

### 8. `stmt_exits_or_reverts` uses "may exit" instead of "must exit"

**Code:** [`stmt_exits_or_reverts`, L973–981](https://github.com/foundry-rs/foundry/blob/b9159f1b752a72198a9e13b37f494f14fd82c2e2/crates/lint/src/sol/low/missing_events_arithmetic.rs#L973-L981)

The implementation uses `.any()` — if any branch of a nested `if` exits, the whole block is considered exiting. An inner conditional `revert` is therefore enough to pass the guard check even when unauthorized callers can reach the write via the non-reverting branch.

```solidity
error NotOwner();
// Spurious warning — unauthorized callers pass when flag is false.
function setPrice(uint256 p) external {
    if (msg.sender != owner) {
        if (flag) revert NotOwner(); // only one branch exits
    }
    price = p; // ← false positive
}
```

---

### 9. Compound `||` condition in `require` is classified as access-controlled

**Code:** [`expr_looks_like_access_check` fallback, L1057–1079](https://github.com/foundry-rs/foundry/blob/b9159f1b752a72198a9e13b37f494f14fd82c2e2/crates/lint/src/sol/low/missing_events_arithmetic.rs#L1057-L1079)

When the condition doesn't match the exact `msg.sender == owner` pattern, the code falls back to `expr_looks_like_access_check`: reads sender AND reads a state variable → authorized. A compound `||` satisfies both criteria even though anyone meeting the second operand can bypass the owner check entirely.

```solidity
// Spurious warning — amount > 0 is sufficient to call; no real owner restriction.
function setPrice(uint256 p, uint256 amount) external {
    require(msg.sender == owner || amount > 0);
    price = p; // ← false positive
}
```

---

### 10. Access-control name heuristic applied regardless of return value use

**Code:** [`expr_has_access_guard` Call arm, L995–1000](https://github.com/foundry-rs/foundry/blob/b9159f1b752a72198a9e13b37f494f14fd82c2e2/crates/lint/src/sol/low/missing_events_arithmetic.rs#L995-L1000)

If the callee name matches the access-control heuristic (`_checkOwner`, `onlyRole`, etc.), `expr_has_access_guard` immediately returns `true` — without checking whether the call's return value is used or whether the function actually reverts. A bool-returning `_checkOwner()` whose result is discarded still classifies the caller as protected.

```solidity
function _checkOwner() internal view returns (bool) { return msg.sender == owner; }

// Spurious warning — return value is discarded; any caller can proceed.
function setPrice(uint256 p) external {
    _checkOwner(); // result ignored
    price = p; // ← false positive
}
```

crates/lint/src/sol/low/missing_events_arithmetic.rs
```diff
@@ -0,0 +1,1676 @@
...
+    is_state_mutating_entry_point(func) && is_protected(hir, func_id, func)
+}
+
+fn is_unprotected_user_function(
+    hir: &hir::Hir<'_>,
+    func_id: FunctionId,
+    func: &hir::Function<'_>,
+) -> bool {
+    is_external_function(func) && !is_protected(hir, func_id, func)
+}
```
mablr: Only used once, should be inlined imo

mablr (dismissed):
lgtm

mablr (approved):

mattsse (approved):
lgtm

PR #14908, opened 2026-05-25
fix(foundryup): unlink previous `bin` to preserve `--path` symlinks

mablr:
## Description

Previously installed version symlink was preventing switching locally built version in `benchmarks.yml` workflow.

Now we ensure any conflicting symlink/file is actually replaced.

### DIFF
--- a/foundryup/foundryup
+++ b/foundryup/foundryup
@@ -3,7 +3,7 @@ set -eo pipefail
 
 # NOTE: if you make modifications to this script, please increment the version number.
 # WARNING: the SemVer pattern: major.minor.patch must be followed as we use it to determine if the script is up to date.
-FOUNDRYUP_INSTALLER_VERSION="1.8.3"
+FOUNDRYUP_INSTALLER_VERSION="1.8.4"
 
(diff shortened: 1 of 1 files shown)


foundryup/foundryup
```diff
@@ -487,7 +487,14 @@ use() {
...
     for bin in "${BINS[@]}"; do
       bin_path="$FOUNDRY_BIN_DIR/$bin"
-      cp "$FOUNDRY_VERSION_DIR/$bin" "$bin_path"
+      # Remove first: `$bin_path` may be a symlink created by a previous
+      # `foundryup --path` invocation, pointing at e.g.
+      # `<repo>/target/release/$bin`. A plain `cp src symlink` follows the
+      # symlink and overwrites the local build artifact, which corrupts the
+      # workspace and silently makes the next `--path` activation use the
+      # downloaded release binary instead of a fresh local build.
+      rm -f "$bin_path"
```
figtracer: we could also ensure the removal (it's `-f` but for edge cases doesn't hurt i think)

figtracer (approved):
tested locally, lg

stevencartavia (approved):

PR #14909, opened 2026-05-25
feat(cast): wrap lists, and json outputs in `JsonEnvelope`

mablr:
## Description

Towards OSS-239, #14899 follow-up.

Handle Cast subcommands producing list and raw JSON.

- `print_tokens` now returns Result and wraps decoded tokens as a JSON array inside the shared envelope
- Added `print_list` helper: plain mode prints one item per line, JSON mode wraps as an array inside the envelope
- Sig `--optimize` conflicts with `--json` at the clap level (progress output is incompatible with structured JSON)


figtracer (dismissed):

stevencartavia (dismissed):

mablr (commented):
Fixed tests.

figtracer (approved):

grandizzy (approved):

PR #14912, opened 2026-05-26
fix(forge): resolve Tempo expiry for creates

nivanovvv:
## Motivation

`forge create` accepts `--tempo.expires`, but leaves the relative expiry to be materialized when transaction options are applied. Other Tempo transaction commands resolve that value once and print the resulting unix timestamp before
submission.

## Solution

Resolve `--tempo.expires` in the create path before building the deployment transaction, then print the materialized timestamp when it is present.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mablr (approved):
Good catch, thanks.

figtracer (approved):

PR #14913, opened 2026-05-26
fix(fmt): preserve braces to prevent dangling-else rebinding

stevencartavia:
`forge fmt` drops braces around a single-statement nested `if`, causing the `else` to rebind to the inner `if` and silently changing semantics.

Input:
```solidity
if (a) { if (b) return 1; } else return 2;
```

Output:
```solidity
if (a) if (b) return 1;
else return 2;
```

With `a=true, b=false`: original returns `0`, formatted version returns `2`.

Fix: in `is_single_line_block`, refuse to elide braces when the `then` block wraps a single `if` (or `while`, whose body has the same risk) and the surrounding `if` has an `else`.

### DIFF
--- a/crates/fmt/src/state/sol.rs
+++ b/crates/fmt/src/state/sol.rs
@@ -2527,13 +2527,16 @@ impl<'ast> State<'_, 'ast> {
     }
 
     /// Returns true if `then` is a block whose single statement, when printed without
-    /// the surrounding braces, would leave an inner `if` exposed to capture a trailing
-    /// `else`. Conservatively covers `if` and `while`, whose bodies may themselves be
-    /// inlined without braces.
+    /// the surrounding braces, could leave an inner `if` exposed to capture a trailing
+    /// `else`. Conservatively covers `if`, `while`, and `for`, whose bodies may carry
+    /// or expose a dangling inner `if`.
     fn then_block_can_capture_trailing_else(then: &'ast ast::Stmt<'ast>) -> bool {
         if let ast::StmtKind::Block(block) = &then.kind
             && block.stmts.len() == 1
-            && matches!(block.stmts[0].kind, ast::StmtKind::If(..) | ast::StmtKind::While(..))
+            && matches!(
+                block.stmts[0].kind,
+                ast::StmtKind::If(..) | ast::StmtKind::While(..) | ast::StmtKind::For { .. }
+            )
         {
             return true;
         }
--- a/crates/fmt/src/state/sol.rs
+++ b/crates/fmt/src/state/sol.rs
@@ -2423,6 +2423,12 @@ impl<'ast> State<'_, 'ast> {
         then: &'ast ast::Stmt<'ast>,
         els_opt: Option<&'ast &'ast mut ast::Stmt<'ast>>,
     ) -> Decision {
+        // Dangling-else guard runs before the cache check so an inlined parent can't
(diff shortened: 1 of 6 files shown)


crates/fmt/src/state/sol.rs
```diff
@@ -2519,6 +2526,20 @@ impl<'ast> State<'_, 'ast> {
...
     }
 
+    /// Returns true if `then` is a block whose single statement, when printed without
+    /// the surrounding braces, would leave an inner `if` exposed to capture a trailing
+    /// `else`. Conservatively covers `if` and `while`, whose bodies may themselves be
+    /// inlined without braces.
+    fn then_block_can_capture_trailing_else(then: &'ast ast::Stmt<'ast>) -> bool {
+        if let ast::StmtKind::Block(block) = &then.kind
+            && block.stmts.len() == 1
+            && matches!(block.stmts[0].kind, ast::StmtKind::If(..) | ast::StmtKind::While(..))
```
grandizzy: we should match on for block as well here `| ast::StmtKind::For(..)` and add an example, e.g. `if (a) { for (uint i; i < n; ++i) if (b) return 1; } else return 2;`

grandizzy (commented):
looks good, pls check comment re including for loops

crates/fmt/src/state/sol.rs
stevencartavia: [71625ab](https://github.com/foundry-rs/foundry/pull/14913/commits/71625abfb182bf42f8d0b9b838fead9d8d1fba5b)

crates/fmt/src/state/sol.rs
```diff
@@ -2433,6 +2433,13 @@ impl<'ast> State<'_, 'ast> {
             return Decision { outcome: false, is_cached: false };
         }
 
+        // Keep braces when `then` is a block wrapping a single statement that could
+        // expose a trailing `if` after brace elision; otherwise the surrounding `else`
+        // would rebind to that inner `if`.
+        if els_opt.is_some() && Self::then_block_can_capture_trailing_else(then) {
```
mattsse: This guard can still be bypassed when a parent statement has already cached an inline decision: [`is_single_line_block` returns before this check](https://github.com/foundry-rs/foundry/blob/ef11dffb8c5d2af3d3d108bd76e4dfe5e1d9aaff/crates/fmt/src/state/sol.rs#L2427-L2439). A nested case like `if (x) { if (a) { if (b) return 1; } else return 2; }` can still print under the parent's cached `true`, so braces around the inner `if (a)` block may be elided and the `else` can bind to `if (b)`.

Suggested fix: run the dangling-else safety check before accepting a cached `true`, or make cached `true` conditional on `!then_block_can_capture_trailing_else(then)` when `els_opt.is_some()`. Please also add a fixture with the dangerous `if/else` nested inside an already-inlineable parent.

mattsse (commented):
Review note from local pass; leaving the concrete issue inline.

crates/fmt/src/state/sol.rs
stevencartavia: [2b05385](https://github.com/foundry-rs/foundry/pull/14913/commits/2b05385fc4432d30a0463bd0969ecea1596e11b6)

grandizzy (approved):
lgtm, thank you!

mablr (approved):

PR #14914, opened 2026-05-26
feat(lint): add `encode-packed-collision`

mablr:
## Description
Closes OSS-58

Add `encode-packed-collision` lint.

Detects `abi.encodePacked()` calls with 2+ dynamic-type arguments (`string`, `bytes`, `T[]`) which can produce hash collisions, e.g. `encodePacked("a","bc") == encodePacked("ab","c")`.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/lint/src/sol/high/encode_packed_collision.rs
+++ b/crates/lint/src/sol/high/encode_packed_collision.rs
@@ -105,8 +105,10 @@ fn call_return_type<'hir>(
             Some(&hir.variable(*ret).ty)
         }
         // Member call: token.name(), token.symbol(), etc.
-        // If multiple functions share the method name (overloads), return None to avoid
-        // false positives from picking the wrong overload.
+        // A contract may expose multiple entries for the same method name when a derived contract
+        // overrides an inherited function. In that case all candidates share the same return type,
+        // so we accept the match as long as every candidate agrees. If they disagree (genuine
+        // overloads with different return types) we bail to avoid false positives.
         ExprKind::Member(recv, method) => {
             let cid = contract_id_of(hir, recv)?;
             let matches: Vec<_> = hir
@@ -122,8 +124,13 @@ fn call_return_type<'hir>(
                     }
                 })
                 .collect();
-            let [ty] = matches.as_slice() else { return None };
-            Some(*ty)
+            let [first, rest @ ..] = matches.as_slice() else { return None };
+            // All candidates must agree on dynamic-ness; if any disagrees it's a genuine overload
+            // and we cannot determine which was called.
+            if rest.iter().any(|ty| is_dynamic_type(&ty.kind) != is_dynamic_type(&first.kind)) {
+                return None;
+            }
+            Some(*first)
         }
         // Indirect call via a function-typed value
         _ => match &expr_type(hir, callee)?.kind {
--- a/crates/lint/src/sol/high/encode_packed_collision.rs
+++ b/crates/lint/src/sol/high/encode_packed_collision.rs
@@ -1,0 +1,108 @@
+use super::EncodedPackedCollision;
+use crate::{
+    linter::{LateLintPass, LintContext},
+    sol::{Severity, SolLint},
+};
+use solar::{
+    interface::sym,
+    sema::hir::{ElementaryType, Expr, ExprKind, Hir, ItemId, Res, Type, TypeKind, VariableId},
+};
+
+declare_forge_lint!(
+    ENCODE_PACKED_COLLISION,
+    Severity::High,
+    "encode-packed-collision",
+    "`abi.encodePacked()` called with multiple dynamic type arguments; hash collisions possible"
+);
+
+impl<'hir> LateLintPass<'hir> for EncodedPackedCollision {
+    fn check_expr(&mut self, ctx: &LintContext, hir: &'hir Hir<'hir>, expr: &'hir Expr<'hir>) {
+        let ExprKind::Call(callee, args, _) = &expr.kind else { return };
+        let ExprKind::Member(base, member) = &callee.peel_parens().kind else { return };
+        if member.name != sym::encodePacked || !is_abi_builtin(base) {
+            return;
+        }
+        let dynamic_count = args.exprs().filter(|arg| is_dynamic_arg(hir, arg)).count();
+        if dynamic_count >= 2 {
+            ctx.emit(&ENCODE_PACKED_COLLISION, expr.span);
+        }
+    }
+}
+
+fn is_abi_builtin(expr: &Expr<'_>) -> bool {
+    matches!(
+        &expr.peel_parens().kind,
+        ExprKind::Ident(reses)
+            if reses.iter().any(|r| matches!(r, Res::Builtin(b) if b.name() == sym::abi))
+    )
+}
+
+fn is_dynamic_arg(hir: &Hir<'_>, expr: &Expr<'_>) -> bool {
+    let Some(ty) = expr_type(hir, expr) else { return false };
+    is_dynamic_type(&ty.kind)
+}
+
+const fn is_dynamic_type(kind: &TypeKind<'_>) -> bool {
+    match kind {
+        TypeKind::Elementary(ElementaryType::String | ElementaryType::Bytes) => true,
+        TypeKind::Array(arr) => arr.size.is_none(),
+        _ => false,
+    }
+}
+
+fn expr_type<'hir>(hir: &'hir Hir<'hir>, expr: &'hir Expr<'hir>) -> Option<&'hir Type<'hir>> {
+    match &expr.peel_parens().kind {
+        ExprKind::Ident(resolutions) => {
+            let var_id = var_resolution(resolutions)?;
+            Some(&hir.variable(var_id).ty)
+        }
+        ExprKind::Call(callee, _, _) => call_return_type(hir, callee),
+        ExprKind::Index(base, _) => match &expr_type(hir, base)?.kind {
+            TypeKind::Array(array) => Some(&array.element),
+            TypeKind::Mapping(mapping) => Some(&mapping.value),
+            _ => None,
+        },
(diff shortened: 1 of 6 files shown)


crates/lint/src/sol/high/encode_packed_collision.rs
```diff
@@ -0,0 +1,108 @@
...
+fn is_abi_builtin(expr: &Expr<'_>) -> bool {
+    matches!(
+        &expr.peel_parens().kind,
+        ExprKind::Ident(reses)
+            if reses.iter().any(|r| matches!(r, Res::Builtin(b) if b.name() == sym::abi))
+    )
+}
+
+fn is_dynamic_arg(hir: &Hir<'_>, expr: &Expr<'_>) -> bool {
+    let Some(ty) = expr_type(hir, expr) else { return false };
```
figtracer: unresolved argument types are ignored, so dynamic cases like `abi.encodePacked("a", "bc")` and `abi.encodePacked(token.name(), token.symbol())` do not warn

crates/lint/testdata/EncodePackedCollision.sol
```diff
@@ -0,0 +1,74 @@
...
+    function fixedArray(uint256[3] memory a, uint256[3] memory b) public pure returns (bytes32) {
+        return keccak256(abi.encodePacked(a, b));
+    }
+
+    // SHOULD PASS: single string arg
+    function singleString(string memory a) public pure returns (bytes32) {
+        return keccak256(abi.encodePacked(a));
+    }
+
+    // SHOULD WARN: call returning string/bytes
```
figtracer, line 62: let's add regression cases for string literals and external/member calls; both are collision-prone and currently missed

mablr: https://github.com/foundry-rs/foundry/pull/14914/commits/27402406cf06efd6b3c79efd5bca69241c6a840e

figtracer (dismissed):

crates/lint/src/sol/high/encode_packed_collision.rs
```diff
@@ -0,0 +1,144 @@
...
+    callee: &'hir Expr<'hir>,
+) -> Option<&'hir Type<'hir>> {
+    match &callee.peel_parens().kind {
+        // Type cast: bytes(x), string(x) — the result type is the cast target
+        ExprKind::Type(ty) => Some(ty),
+        // Direct function call: getString(), getBytes()
+        ExprKind::Ident(resolutions) => {
+            let fid = resolutions.iter().find_map(|r| {
+                if let Res::Item(ItemId::Function(fid)) = r { Some(*fid) } else { None }
+            })?;
```
stevencartavia: `ExprKind::Ident` picks the first function with a matching name and ignores the call's argument types. 

```solidity
contract OverloadFP {
    function f(uint256) internal pure returns (string memory) { return ""; }
    function f(bytes32) internal pure returns (uint256) { return 1; }

    // Only `s` is dynamic — should NOT warn.
    function g(bytes32 x, string memory s) public pure returns (bytes32) {
        return keccak256(abi.encodePacked(f(x), s));
    }
}
```

mablr: https://github.com/foundry-rs/foundry/pull/14914/commits/19d7dbd3b4d164bbd2309b21618a1bdeb889414d

mattsse (dismissed):
P2: Please fix member-call return inference before shipping. The current logic scans inherited contract items by method name and then bails unless there is exactly one match: [`encode_packed_collision.rs#L110-L126`](https://github.com/foundry-rs/foundry/blob/19d7dbd3b4d164bbd2309b21618a1bdeb889414d/crates/lint/src/sol/high/encode_packed_collision.rs#L110-L126).

For concrete contracts that override inherited methods, e.g. `name()` / `symbol()`, both the base function and override can match, so `abi.encodePacked(token.name(), token.symbol())` is missed even though both return `string`.

Suggested fix: resolve the actual called function using the call args / semantic function type instead of collecting by name only, or filter overridden base functions before the uniqueness check. Please also add a regression test with a concrete contract overriding inherited `name()` and `symbol()` that still warns.

figtracer (dismissed):

grandizzy:
A few things worth addressing before merge, mostly around the hand-rolled type inference in `expr_type` / `call_return_type`:

### Likely false negatives (silent miss on real bugs)

The `ExprKind::Member` arm in `expr_type` only handles struct fields and returns `None` for everything else. That causes the lint to miss several common dynamic expressions:

- `abi.encodePacked(msg.data, s)` — `msg.data` is `bytes calldata`
- `abi.encodePacked(addr.code, s)` / `address(this).code` / `type(C).creationCode|runtimeCode`
- `abi.encodePacked(abi.encode(a), abi.encode(b))` — works only with an explicit `bytes(...)` cast today (that's all the fixture covers); the bare nested call resolves as a member call on the `abi` builtin, and `contract_id_of` bails because `abi` is not a contract. Same for nested `abi.encodePacked`, `abi.encodeWithSelector`, etc.
- `abi.encodePacked(string.concat(a, b), c)` and `bytes.concat(...)` — receiver is `ExprKind::Type`
- `abi.encodePacked(new bytes(n), s)`
- `abi.encodePacked(token.name(), s)` where `name` is a `string public` state variable — `item.as_function()` skips synthesized getters if the HIR represents them as variables (worth checking)
- `using X for string` method-style calls
- Literal inside a ternary: `flag ? a : "x"` — `expr_type(Ternary)` recurses into `expr_type`, which has no `Lit` arm, so the literal branch isn't classified as dynamic
- Overload selection: `call_return_type` filters by name+arity only. If two same-arity overloads exist and the chosen one returns `string`, the bail returns `None`. The current fixture only exercises the inverse direction (selected overload non-dynamic → no warn); please add the opposite.

### Likely false positives worth refining

The literal short-circuit (`LitKind::Str → dynamic`) over-fires on safe idioms:

- `abi.encodePacked("prefix:", s)` and `abi.encodePacked(hex"dead", s)`
- `abi.encodePacked(uint256(bytes(a).length), a, uint256(bytes(b).length), b)` — the canonical safe pattern
- `abi.encodePacked(s, s)` — injective in `s`
- EIP-191 personal-sign with `uintToString(len)` before `data`

A common mitigation other linters use: only count *non-literal* dynamics, or require ≥2 non-literal dynamic args.

### Minor soundness / style notes

- `is_abi_builtin` uses `reses.iter().any(...)` — if `abi` is ever shadowed and co-resolves with the builtin, the lint still fires. Exact-match would be safer.
- `call_return_type`'s "all candidates agree on dynamic-ness, return first" comment should be explicit that the result is only meaningful for the dynamic/non-dynamic decision, not as a true type — easy footgun if reused.
- The `Slice` arm comment says "preserves the element type of the base"; it actually preserves the **base** type (correct for `bytes` slices, just misleading wording).

### Suggested additional fixtures

```solidity
// Currently MISSED — should warn
abi.encodePacked(msg.data, s);
abi.encodePacked(addr.code, s);
abi.encodePacked(abi.encode(a), abi.encode(b));      // no bytes() cast
abi.encodePacked(string.concat(a, b), c);
abi.encodePacked(bytes.concat(a, b), c);
abi.encodePacked(new bytes(10), s);
abi.encodePacked(flag ? a : "x", b);
abi.encodePacked(token.name(), s);                    // public string getter
abi.encodePacked(Strings.toString(x), Strings.toString(y)); // library

// Known FPs
abi.encodePacked("prefix:", s);
abi.encodePacked(s, s);
abi.encodePacked(uint256(bytes(a).length), a, uint256(bytes(b).length), b);
```

### Minimal fix suggestion

If solar exposes a resolved-callee or expression-type API, prefer that over the ad-hoc walker. If not, special-casing the handful of well-known Solidity builtins inside `expr_type(Member)` / `call_return_type` (`abi.encode*` → `bytes`, `msg.data` → `bytes`, `*.code` → `bytes`, `string.concat` → `string`, `bytes.concat` → `bytes`) plus a `Lit` arm in `expr_type` would close most of the gaps above with little added complexity.

mablr (commented):
@grandizzy I fixed your findings in https://github.com/foundry-rs/foundry/pull/14914/commits/8f60378357fb3a2520cf21f8486ae39d7e586a47

grandizzy:
> @grandizzy I fixed your findings in https://github.com/foundry-rs/foundry/commit/8f60378357fb3a2520cf21f8486ae39d7e586a47

@mablr - looks good, one small new FP worth addressing before merge:

### `.code` / `.creationCode` / `.runtimeCode` member match is unconditional

`is_dynamic_builtin_member` matches these member names without checking the base type, so a struct field with one of those names is wrongly treated as dynamic `bytes`:

```solidity
struct S { uint256 code; }

function f(S memory x, string memory s) public pure returns (bytes32) {
    return keccak256(abi.encodePacked(x.code, s)); // currently warns; should pass
}
```

(`msg.data` is already safe because `is_builtin_named(base, sym::msg)` requires the base to resolve to the `msg` builtin — same pattern would work here, but `expr_type`-first is simpler.)

Trivial hardening — prefer the resolved type, fall back to the builtin-member check:

```rust
ExprKind::Member(base, member) => {
    if let Some(ty) = expr_type(hir, expr) {
        is_dynamic_type(&ty.kind)
    } else {
        is_dynamic_builtin_member(base, member)
    }
}
```

Worth adding a PASS fixture for the struct-field case too.

### Minor doc nit

The "Why is this bad?" section could note that the lint flags by *type*, not by exploitability — `abi.encodePacked(s, s)` and length-prefix-safe patterns like `abi.encodePacked(uint256(bytes(a).length), a, …, b)` will warn even though they're not collidable. That's a fine conservative choice, just worth setting expectations.

Otherwise LGTM.

grandizzy (approved):
lgtm!

figtracer (approved):
lgtm

PR #14915, opened 2026-05-26
feat(lint): add `write-after-write`

mablr:
## Description

Closes OSS-84

Add `write-after-write` lint.

Detects redundant consecutive storage writes to the same variable where the first value is overwritten before being read, wasting an SSTORE.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/lint/src/sol/gas/write_after_write.rs
+++ b/crates/lint/src/sol/gas/write_after_write.rs
@@ -7,7 +7,9 @@ use solar::{
     interface::Span,
     sema::{
         Hir,
-        hir::{Block, Expr, ExprKind, Function, ItemId, Res, Stmt, StmtKind, VariableId},
+        hir::{
+            BinOpKind, Block, Expr, ExprKind, Function, ItemId, Res, Stmt, StmtKind, VariableId,
+        },
     },
 };
 use std::collections::HashMap;
@@ -170,7 +172,12 @@ fn process_expr<'hir>(
         }
         // Any function/method call can observe state through re-entrancy or view
         // calls, so conservatively treat it as reading everything pending.
-        ExprKind::Call(_, _, _) => {
+        // Walk callee and arguments first — Solidity evaluates them before the call.
+        ExprKind::Call(callee, args, _) => {
+            collect_reads(ctx, hir, callee, pending);
+            for arg in args.exprs() {
+                collect_reads(ctx, hir, arg, pending);
+            }
             pending.clear();
         }
         // For any other expression used as a statement, scan for reads.
@@ -201,6 +208,14 @@ fn collect_reads<'hir>(
             // reads it. Delegate to process_expr so the write is tracked correctly.
             process_expr(ctx, hir, expr.peel_parens(), pending);
         }
+        // Short-circuit operators: LHS always evaluates, RHS may not.
+        // Clear outer pending before RHS to avoid false-positive WAW in the conditional path.
+        ExprKind::Binary(lhs, op, rhs) if matches!(op.kind, BinOpKind::And | BinOpKind::Or) => {
+            collect_reads(ctx, hir, lhs, pending);
+            pending.clear();
+            let mut rhs_pending = HashMap::default();
+            collect_reads(ctx, hir, rhs, &mut rhs_pending);
+        }
         ExprKind::Binary(lhs, _, rhs) => {
             collect_reads(ctx, hir, lhs, pending);
             collect_reads(ctx, hir, rhs, pending);
@@ -208,13 +223,23 @@ fn collect_reads<'hir>(
         ExprKind::Unary(_, inner) | ExprKind::Payable(inner) => {
             collect_reads(ctx, hir, inner, pending);
         }
+        // Ternary arms are mutually exclusive; analyze each independently with a fresh
+        // pending to avoid false-positive WAW between branches.
         ExprKind::Ternary(cond, t, f) => {
             collect_reads(ctx, hir, cond, pending);
-            collect_reads(ctx, hir, t, pending);
-            collect_reads(ctx, hir, f, pending);
+            pending.clear();
+            let mut then_pending = HashMap::default();
+            let mut else_pending = HashMap::default();
+            collect_reads(ctx, hir, t, &mut then_pending);
+            collect_reads(ctx, hir, f, &mut else_pending);
         }
         // Any call may observe storage through re-entrancy or view semantics.
-        ExprKind::Call(_, _, _) => {
+        // Walk callee and arguments first — Solidity evaluates them before the call.
+        ExprKind::Call(callee, args, _) => {
+            collect_reads(ctx, hir, callee, pending);
+            for arg in args.exprs() {
+                collect_reads(ctx, hir, arg, pending);
+            }
             pending.clear();
         }
         ExprKind::Index(base, index) => {
--- a/crates/lint/src/sol/gas/write_after_write.rs
+++ b/crates/lint/src/sol/gas/write_after_write.rs
@@ -43,6 +43,13 @@ fn check_block<'hir>(
 ) {
     for stmt in block.stmts {
         check_stmt(ctx, hir, stmt, pending);
+        // Stop at terminal statements; subsequent code is unreachable.
+        if matches!(
+            stmt.kind,
+            StmtKind::Return(_) | StmtKind::Revert(_) | StmtKind::Break | StmtKind::Continue
+        ) {
+            break;
+        }
     }
 }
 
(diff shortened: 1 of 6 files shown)


crates/lint/src/sol/gas/write_after_write.rs
```diff
@@ -0,0 +1,261 @@
...
+        ExprKind::Ident(resolutions) => {
+            for res in *resolutions {
+                if let Res::Item(ItemId::Variable(id)) = res
+                    && hir.variable(*id).is_state_variable()
+                {
+                    pending.remove(id);
+                }
+            }
+        }
+        ExprKind::Assign(lhs, _, rhs) => {
```
figtracer: nested assignments are treated as reads, so `x = 1; uint256 z = (x = v);` emits no note

figtracer (commented):
correctness

figtracer (dismissed):

mattsse (dismissed):
Requesting changes for a few correctness gaps in `write-after-write`.

P2: `Binary` and `Ternary` are walked linearly here: [`write_after_write.rs#L204-L215`](https://github.com/foundry-rs/foundry/blob/997bdfe954968bac4baff1de08c931e26256a2b1/crates/lint/src/sol/gas/write_after_write.rs#L204-L215). This can false-positive across `&&`/`||` short-circuiting and ternary arms, and conflicts with the docs' “flat sequence” scope: [`book#L16-L20`](https://github.com/foundry-rs/book/blob/56186faf690cd0f2d9ad8c0954ebe2a5e8de443e/src/pages/forge/linting/write-after-write.mdx#L16-L20). Suggest treating conditional/short-circuit expression boundaries like branches: process condition reads, then clear or analyze each reachable arm with separate pending state.

P2: The lint only implements `check_function`: [`write_after_write.rs#L22-L33`](https://github.com/foundry-rs/foundry/blob/997bdfe954968bac4baff1de08c931e26256a2b1/crates/lint/src/sol/gas/write_after_write.rs#L22-L33), so modifier bodies are not analyzed even though placeholder handling exists: [`write_after_write.rs#L114-L117`](https://github.com/foundry-rs/foundry/blob/997bdfe954968bac4baff1de08c931e26256a2b1/crates/lint/src/sol/gas/write_after_write.rs#L114-L117). Suggest adding modifier-body traversal or removing/de-scoping placeholder logic and tests until modifiers are actually checked.

P3: Calls clear pending without first walking the callee/args: [`write_after_write.rs#L171-L174`](https://github.com/foundry-rs/foundry/blob/997bdfe954968bac4baff1de08c931e26256a2b1/crates/lint/src/sol/gas/write_after_write.rs#L171-L174), [`write_after_write.rs#L216-L218`](https://github.com/foundry-rs/foundry/blob/997bdfe954968bac4baff1de08c931e26256a2b1/crates/lint/src/sol/gas/write_after_write.rs#L216-L218). Solidity evaluates call arguments before the call, so nested overwrites in args are missed. Suggest walking callee/arguments first, then clearing pending before modeling the call itself.

crates/lint/src/sol/gas/write_after_write.rs
```diff
@@ -0,0 +1,264 @@
...
+                    pending.remove(id);
+                }
+            }
+        }
+        ExprKind::Assign(_, _, _) => {
+            // A nested assignment (e.g. `uint256 z = (x = v)`) writes to its LHS, not
+            // reads it. Delegate to process_expr so the write is tracked correctly.
+            process_expr(ctx, hir, expr.peel_parens(), pending);
+        }
+        ExprKind::Binary(lhs, _, rhs) => {
```
stevencartavia, line 272: Short-circuit `&&` / `||` — RHS is conditionally evaluated
```
x = 1;
b && ((x = 2) == 2);  // flags `x = 1`; not dead when b == false
```
(same with || when b == true)

crates/lint/src/sol/gas/write_after_write.rs
```diff
@@ -0,0 +1,264 @@
...
+            collect_reads(ctx, hir, lhs, pending);
+            collect_reads(ctx, hir, rhs, pending);
+        }
+        ExprKind::Unary(_, inner) | ExprKind::Payable(inner) => {
+            collect_reads(ctx, hir, inner, pending);
+        }
+        ExprKind::Ternary(cond, t, f) => {
```
stevencartavia, line 281: here only one arm executes, so neither write is dead 
```
b ? (x = 1) : (x = 2);  // flags `x = 1`
```

crates/lint/src/sol/gas/write_after_write.rs
mablr: addressed here https://github.com/foundry-rs/foundry/pull/14915/commits/cf8fcd83fffad8f075ad6696239eed8e56f5d37d

crates/lint/src/sol/gas/write_after_write.rs
mablr: same https://github.com/foundry-rs/foundry/pull/14915/commits/cf8fcd83fffad8f075ad6696239eed8e56f5d37d

figtracer (dismissed):

grandizzy:
A few things worth addressing before merge, mostly around the data-flow precision in `check_stmt` / `collect_reads`:

### Real false positives from nested terminators

`check_block` only stops when the *immediate* statement is `Return | Revert | Break | Continue`. Terminators inside nested blocks/branches don't propagate, so unreachable code is still treated as live and the first write gets flagged:

```solidity
{ return; }
x = 1;
x = 2;                       // unreachable, but x = 1 will be flagged

if (c) return; else revert E();
x = 1;
x = 2;                       // unreachable on both paths, but x = 1 will be flagged

while (c) {
    { break; }
    x = 1;
    x = 2;                   // unreachable after break in this iteration
}
```

Fix: thread a small `Flow { Continue, StopCurrentBlock }` through `check_stmt` / `check_block`. `Block` / `UncheckedBlock` propagate from children; `If` returns `StopCurrentBlock` only if both branches stop; `Loop` always returns `Continue` (zero-iteration possible).

### Inline assembly handling

The `match` on `StmtKind` isn't exhaustive without an `Assembly` / `Yul` arm. Either solar's pinned HIR has no such variant (in which case fine), or it does and it's silently ignored — which would be unsound, since inline assembly can `sload` / `sstore` / call / revert any storage:

```solidity
x = 1;
assembly { /* may read or overwrite x.slot */ }
x = 2;                       // currently x = 1 likely flagged — unsound
```

Safe behavior is `pending.clear()` for the assembly statement. Worth confirming against solar's actual `StmtKind` enum and adding a fixture.

### Tuple diagnostic span is misleading

```solidity
(x, y) = (1, 2);
x = v;
```

emits one warning pointing at the **whole tuple statement**, but only the `x` slot is dead — `y` may be live. `process_assignment_lhs` should pass the component span (`e.span`) instead of the outer `assign_span` so the warning points at `x` specifically.

### Likely false negatives worth fixtures (or documenting)

These are the conservative-clear trade-offs; mostly defensible, but should at least be visible as tests:

- **Branch-local dead writes**: `x = 1; if (c) { y = 2; } x = v;` — `x = 1` dead on both paths, missed (outer pending cleared on `If`).
- **Both branches assign**: `if (c) x = 1; else x = 2; x = 3;` — both branch writes dead, missed.
- **Short-circuit / ternary in RHS**: `x = 1; bool b = a && b2; x = 2;` and `x = 1; uint256 z = c ? 7 : 8; x = 2;` — `x = 1` lost because `collect_reads`'s `&&` / `||` / `Ternary` arms unconditionally `pending.clear()` even when the surrounding context doesn't read `x`.
- **Pure / view / builtin calls**: `x = 1; pureFn(); x = 2;` or any `abi.encode(...)` — all calls clear pending. Defensible for reentrancy, but loses many obviously-safe cases.
- **`do { x = 2; } while (c);`** after `x = 1` — body executes ≥ 1, but loops always wipe pending.
- **Modifier arguments**: writes in `f() m(x = 1)` live on `func.modifiers`, not `func.body`, and aren't analyzed.

### Minor soundness / style notes

- `process_assignment_lhs` for tuple LHS uses the same `assign_span` for every component, so two same-statement WAW reports point to identical spans and are hard to distinguish.
- The `Loop(block, _)` second field is `LoopSource` (init/cond/update are desugared into the loop block in HIR) — so that's fine, not a bug.
- `DeclMulti` with state-var LHS isn't a real concern: `(uint a, x) = …` shows up as a tuple `Assign`, which is handled.
- `x = f(); x = v;` where `f()` may revert is sound — on success the first is dead, on revert the second never runs.
- `emit E(x)` correctly doesn't clear pending; reads in args are removed by `collect_reads`.

### Suggested additional fixtures

```solidity
// Currently FP (or will be after fix) — unreachable code should not flag
function nestedReturn() public {
    { return; }
    x = 1;
    x = 2;
}

function bothBranchesExit(bool c) public {
    if (c) return; else revert();
    x = 1;
    x = 2;
}

function nestedBreak(bool c) public {
    while (c) {
        { break; }
        x = 1;
        x = 2;
    }
}

// Inline assembly — verify soundness
function asmReadsX() public {
    x = 1;
    assembly { let v := sload(0) }
    x = 2;
}

// UX: span should point at the `x` component, not the whole tuple
function tuplePartial(uint256 v) public {
    (x, y) = (1, 2);
    x = v;
}

// Known FNs — document as limitations
function branchMiss(bool c, uint256 v) public {
    x = 1;
    if (c) { y = 2; }
    x = v;
}
function shortCircuitMiss(bool a, bool b) public {
    x = 1;
    bool z = a && b;
    x = 2;
}
function pureCallMiss() public {
    x = 1;
    abi.encode(uint256(0));
    x = 2;
}
```

### Minimal fix suggestion

The three items above (`Flow` propagation through nested terminators, explicit assembly handling, component-level tuple spans) close the real soundness / UX issues with little added complexity. The conservative-clear false negatives can be left as-is — but adding fixtures for them makes the design intent visible to future maintainers.

mablr (commented):
@grandizzy I fixed your findings in https://github.com/foundry-rs/foundry/pull/14915/commits/06e3f723edb83da80e7cea1717984f53fee01f01.

grandizzy (approved):

figtracer (approved):

PR #14916, opened 2026-05-26
 feat(forge): add tempo session key support for scripts

0xKarl98:
This PR adds Tempo session key support to forge script, so scripts can run with a temporary, scoped access key instead of requiring a long-lived hot wallet signer. The implementation keeps the session flow narrow: it resolves the configured session only when needed, enforces that a session covers a single root sender, handles gas estimation for access-key transactions, and preserves valid --resume / multi-chain broadcasts by validating only the transactions that still need to be sent.

### DIFF
--- a/crates/script/src/broadcast.rs
+++ b/crates/script/src/broadcast.rs
@@ -264,13 +264,6 @@ where
     }
 }
 
-struct ScriptSessionSigner {
-    chain: u64,
-    root_account: Address,
-    signer: WalletSigner,
-    access_key: TempoAccessKeyConfig,
-}
-
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub(crate) struct SignerScope {
     chain: u64,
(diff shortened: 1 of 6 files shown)


crates/script/src/lib.rs
```diff
@@ -270,10 +270,22 @@ impl ScriptArgs {
         config: Config,
         mut evm_opts: EvmOpts,
     ) -> Result<PreprocessedState<FEN>> {
+        let has_session = self.tempo.session_id()?.is_some();
+        let session_sender = if has_session && !self.resume {
+            let expected_chain_id = Self::tempo_session_chain_id(&evm_opts).await;
```
figtracer: i reproduced a chain 4217 script failing here as chain 31337 unless `--chain 4217` is added. 

can we defer session chain validation for non-resume `--multi` until broadcast, like resume does?

0xKarl98: Have included --multi case

crates/script/src/lib.rs
```diff
@@ -270,10 +270,22 @@ impl ScriptArgs {
         config: Config,
         mut evm_opts: EvmOpts,
     ) -> Result<PreprocessedState<FEN>> {
+        let has_session = self.tempo.session_id()?.is_some();
+        let session_sender = if has_session && !self.resume && !self.multi {
```
figtracer: we have a similar problem here

repro: a minimal script with `vm.startBroadcast()` against local Tempo Anvil chain 4217; 

with `--multi --tempo.session`, the saved broadcast artifact recorded `from = 0x1804c8ab1f12e6bbf3894d4083f33e07309d1f38` and then failed with: 
```bash
You seem to be using Foundry's default sender.
``` 

without `--multi`, the same fixture records the session root as `from`

0xKarl98: Have addressed this

figtracer (dismissed):

crates/script/src/broadcast.rs
```diff
@@ -369,6 +439,32 @@ impl<FEN: FoundryEvmNetwork> BundledState<FEN> {
...
+                    if let Some(session_signer) = ScriptSessionSigner::resolve(
+                        &self.script_config,
+                        &self.args.wallets,
+                        expected_session_sender,
+                        chain,
+                    )? {
+                        access_keys.insert(
+                            session_signer.root_account,
+                            (session_signer.signer, session_signer.access_key),
+                        );
```
mablr: `ScriptSessionSigner::resolve` calls `session_signer_for_multi_wallet`, which calls `resolve_session_signer`, which **bails** when `session.chain_id != expected_chain_id`. So if a script has remaining unsigned txs on two chains (say 1 and 4217) and a session is configured for 4217, iterating chain 1 in the `for chain in remaining_transactions...unique()` loop propagates an `Err`; even though chain-1 txs could use regular signers. The `Ok(None)` early-return in `resolve` only fires when no session is configured at all.

crates/script/src/build.rs
```diff
@@ -20,9 +23,37 @@ use foundry_compilers::{
...
+    chains: impl IntoIterator<Item = u64>,
+) -> Result<Vec<Address>> {
+    let mut signers = script_wallets
+        .signers()
+        .map_err(|e| eyre::eyre!("Failed to get available signers: {}", e))?;
+
+    for chain in chains {
+        if let Some(session) =
+            script_config.tempo.session_signer_for_multi_wallet(wallets, expected_sender, chain)?
+            && !signers.contains(&session.access_key.wallet_address)
```
mablr: Same chain mismatch issue

crates/script/src/broadcast.rs
```diff
@@ -1034,3 +1137,171 @@ impl BundledState<TempoEvmNetwork> {
...
+    fn script_tx(from: Address) -> TransactionWithMetadata<Ethereum> {
+        TransactionWithMetadata::from_tx_request(TransactionMaybeSigned::new(TransactionRequest {
+            from: Some(from),
+            ..Default::default()
+        }))
+    }
+
+    fn remaining_addr() -> Address {
+        address!("0x2222222222222222222222222222222222222222")
+    }
```
mablr: Seems a bit dummy

mablr (commented):
Maybe a chain mismatch problem

0xKarl98:
@mablr  have addressed the chain mismatch to be chain-aware , ptal again

crates/script/src/build.rs
```diff
@@ -20,9 +29,44 @@ use foundry_compilers::{
...
+        .signers()
+        .map_err(|e| eyre::eyre!("Failed to get available signers: {}", e))?;
+    let remaining = remaining.into_iter().collect::<HashSet<_>>();
+    let mut available = remaining
+        .iter()
+        .copied()
+        .filter(|scope| signers.contains(&scope.sender()))
+        .collect::<HashSet<_>>();
+
+    if let Some(session) =
```
figtracer: this still resolves the live Tempo session even when `remaining` is empty, so `--resume --tempo.session --verify` can fail after the session expires even though there are no transactions left to sign

crates/script/src/broadcast.rs
```diff
@@ -369,6 +537,25 @@ impl<FEN: FoundryEvmNetwork> BundledState<FEN> {
...
         } else {
+            let expected_session_sender = script_session_expected_sender_if_configured(
+                &self.script_config,
+                &required_addresses,
+            )?;
+
+            // For addresses without an explicit signer, try Tempo keys.toml fallback.
+            let mut access_keys: HashMap<SignerScope, (WalletSigner, TempoAccessKeyConfig)> =
+                HashMap::default();
+            if let Some(session_signer) = ScriptSessionSigner::resolve_any_chain(
```
figtracer: same issue during broadcast: `resolve_any_chain` runs before checking whether any remaining unsigned transaction will use the session

mattsse (dismissed):
LGTM

crates/script/src/broadcast.rs
```diff
@@ -799,13 +933,19 @@ impl BundledState<TempoEvmNetwork> {
 
         let batch_signer = if self.args.unlocked {
             BatchSigner::Unlocked
+        } else if let Some(session) = self.script_config.tempo.session_signer_for_multi_wallet(
```
figtracer, line 959: live repro: a completed batch resume artifact with `transactions=1` and `receipts=1` still fails with `Tempo session ... is not active or has no live key` when run with `forge script --broadcast --resume --batch --tempo.session`

`broadcast_batch` should skip session resolution/broadcast when receipts already cover the sequence

figtracer (commented):
nit and pls fix clippy @0xKarl98, thank you!

mattsse (dismissed):

figtracer (dismissed):

figtracer (approved):

mattsse (approved):

PR #14917, opened 2026-05-26
feat(symbolic): cache satisfiability queries

figtracer:
## Summary
- add a bounded per-executor cache for normalized `is_sat` results
- cache only definitive sat/unsat answers, not model queries, unknowns, errors, or heuristic fallback results
- use a conservative structural key that ignores top-level `true` constraints and sorts top-level conjuncts

### DIFF
--- a/crates/evm/symbolic/src/consts.rs
+++ b/crates/evm/symbolic/src/consts.rs
@@ -40,6 +40,9 @@ pub(crate) const PORTFOLIO_SCHEDULER_MIN_RECENCY_WEIGHT: i64 = 1;
 pub(crate) const PORTFOLIO_SCHEDULER_SPEED_BONUS_CAP_MS: u128 = 100;
 pub(crate) const PORTFOLIO_SCHEDULER_MAX_SPEED_BONUS: i64 = 100;
 
+// Solver query cache limits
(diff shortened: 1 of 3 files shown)


figtracer:
## SAT cache benchmark

Compared PR head `fig/symbolic-sat-cache` (`26490da22`) against base `origin/mattsse/native-symbolic-testing` (`724cda50b`). Commands used debug `forge` binaries from each worktree, same temp project, compilation already warm, `hyperfine --warmup 1 --runs 7 --ignore-failure`.

| Case | Base queries | Cache queries | Query delta | Base mean | Cache mean | Wall delta |
|---|---:|---:|---:|---:|---:|---:|
| Magic constant | 3 | 3 | 0% | 74.4ms | 74.3ms | ~0% |
| ERC4626 inflation | 33 | 27 | -18.2% | 2025.5ms | 1985.2ms | -2.0% |
| Repeated branch microcase | 39 | 35 | -10.3% | 227.2ms | 210.3ms | -7.4% |

Takeaway: this PR removes repeated normalized `is_sat` subprocess work when it exists, without changing paths or witnesses. Wall-time gains are workload-dependent: no-op on single-shot/simple solver workloads, small but measurable on ERC4626, clearer on repeated branch feasibility checks. This is still intentionally conservative: it does not cache `model()`, `unknown`, errors, timeouts, or heuristic fallback results.

grandizzy (approved):
lgtm!

PR #14918, opened 2026-05-26
feat(cast): bring remaining `--json` outputs into `JsonEnvelope`

mablr:
## Description

Towards OSS-239 (almost there after `wallet` follow-up)

Move `print_scalar`, `print_list`, `print_tokens`, and `print_json_object` from local helpers in `args.rs` into `foundry_cli::json` so every cmd module can import them directly instead of inlining the if/else pattern.

Replace all remaining inline `if shell::is_json() { sh_println!(...) }` patterns across cmd modules:
- `estimate`, `find_block`, `call`, `mktx` → `print_scalar`
- `txpool` (all variants), `rpc`, `vaddr/create` → `print_json_object`
- `keychain list/inspect/doctor` → `print_json_object` / `print_json_success`
- `erc20 name/symbol/decimals` → `print_scalar`
- `erc20 balance/allowance/total_supply` → `print_json_success` (keep `format_uint_exp` for plain-mode display)
- `Block` and `Tx` in `args.rs` → parse pre-formatted JSON string and wrap
- `Selectors` in `args.rs` → new `SelectorInfo` struct, envelope-wraps array
- `AbiEncodeEvent` in `args.rs` → new `EncodedEvent{topics,data}` struct
- `interface` → wrap ABI array in envelope on stdout path

Documented explicitly unsupported commands with inline comments: `send` (N::ReceiptResponse lacks Serialize), `logs --subscribe` (streaming), `wallet` (follow-up), `upload-signature` (external API), `four-byte-calldata` (interactive).

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes (for commands that already output JSON by default)


crates/cast/src/args.rs
```diff
@@ -521,15 +534,41 @@ pub async fn run_command(args: CastArgs) -> Result<()> {
...
+                    resolved: Option<String>,
+                }
+                let infos: Vec<SelectorInfo> = functions
+                    .into_iter()
+                    .enumerate()
+                    .map(|(pos, (selector, arguments, state_mutability))| SelectorInfo {
+                        selector: selector.to_string(),
+                        arguments,
+                        state_mutability: state_mutability.to_string(),
+                        resolved: resolve.then_some(resolve_results[pos].clone()),
```
figtracer: `resolve.then_some(...)` eagerly indexes `resolve_results` even when `resolve` is false; `cast selectors --json` panics because the vector is empty

crates/cast/src/cmd/call.rs
```diff
@@ -434,7 +435,7 @@ impl CallArgs {
                 sh_warn!("Contract code is empty")?;
             }
         }
-        sh_println!("{}", response)?;
+        print_scalar(response)?;
```
figtracer: for decoded calls, `Cast::call` already returns a JSON string when `--json` is set, so `print_scalar(response)` makes envelope `data` a string containing JSON instead of the decoded array/object

crates/cast/src/cmd/interface.rs
```diff
@@ -85,27 +85,32 @@ impl InterfaceArgs {
...
         let interfaces = get_interfaces(abis, config)?;
 
-        // Print result or write to file.
-        let res = if shell::is_json() {
-            // Format as JSON.
-            interfaces.iter().map(|iface| &iface.json_abi).format("\n").to_string()
-        } else {
-            // Format as Solidity.
-            format!(
+        if let Some(loc) = output_location {
```
figtracer, line 88: this ignores `--json` when `--output` is set and now writes Solidity to the requested file; before this PR `cast --json interface ... -o abi.json` wrote JSON ABI

crates/cast/src/cmd/call.rs
mablr: I think this is what we want. I made similar changes for json outputs here https://github.com/foundry-rs/foundry/pull/14909.

crates/cast/src/cmd/interface.rs
mablr: https://github.com/foundry-rs/foundry/pull/14918/commits/cb2b053b6df4144f740cafa730edda3d59a9a264

crates/cast/src/args.rs
mablr: efd0cc116d2b27fb9428db4b00381438869b1387

mattsse (dismissed):
Requesting changes: these paths can fail or emit invalid JSON for valid `--json` usages.

P1: [`cast call --json` reparses `response` as JSON](https://github.com/foundry-rs/foundry/blob/959137ac56953bd3e5e952cc35e835c475d6fee0/crates/cast/src/cmd/call.rs#L438-L440), but [`Cast::call` returns raw hex when no output is decoded](https://github.com/foundry-rs/foundry/blob/959137ac56953bd3e5e952cc35e835c475d6fee0/crates/cast/src/lib.rs#L193-L196). Valid calls can therefore error on `serde_json::from_str`. Please avoid reparsing arbitrary display output; return structured data from `Cast::call`, or wrap raw hex as a JSON string/envelope in the command layer.

P1: [`cast block --json`](https://github.com/foundry-rs/foundry/blob/959137ac56953bd3e5e952cc35e835c475d6fee0/crates/cast/src/args.rs#L446-L448) and [`cast tx --json`](https://github.com/foundry-rs/foundry/blob/959137ac56953bd3e5e952cc35e835c475d6fee0/crates/cast/src/args.rs#L701-L703) also assume `output` is a JSON object. `--raw` or scalar field paths such as block hash / tx from can return scalar strings, which either fail JSON parsing or are not objects. Please only call `print_json_object` for object output, and use a scalar JSON envelope/string path for field selections.

P2: [`cast mktx` prints the expiration status line to stdout](https://github.com/foundry-rs/foundry/blob/959137ac56953bd3e5e952cc35e835c475d6fee0/crates/cast/src/cmd/mktx.rs#L154-L156) before later JSON/scalar output, which corrupts `--json` output. Please move this to stderr/warning output, suppress it under `shell::is_json()`, or include it inside the JSON payload.

Please add regressions for `cast call --json` with undecoded/raw output, `cast block/tx --json` scalar field paths, and `cast mktx --json --expires-at ...` producing a single valid JSON document.

mablr (commented):
Outscoped `cast call` as it was overlapping with #14774.

Fixed the rest.

crates/cast/src/cmd/vaddr/create.rs
```diff
@@ -96,18 +99,15 @@ pub(super) async fn run(
...
-            serde_json::to_string_pretty(&json!({
-                "salt": format!("{}", output.salt),
-                "registration_hash": format!("{}", output.registration_hash),
-                "master_id": format!("{}", output.master_id),
-                "virtual_addresses": virtual_addresses.iter().map(|(tag, addr)| json!({
-                    "tag": format!("{tag}"),
-                    "address": format!("{addr}"),
-                })).collect::<Vec<_>>(),
-            }))?
-        )?;
```
figtracer, line 110: This prints a success JSON envelope before registration runs, so a later registration error produces two stdout JSON documents. Defer success output until registration succeeds.

crates/cast/src/cmd/keychain.rs
```diff
@@ -732,7 +733,7 @@ fn run_list() -> Result<()> {
 
     if shell::is_json() {
         let entries: Vec<_> = keys_file.keys.iter().map(key_entry_to_json).collect();
-        sh_println!("{}", serde_json::to_string_pretty(&entries)?)?;
```
figtracer, line 759: Empty `cast --json keychain list` prints plain text instead of an empty array envelope. Move the JSON branch before this guard.

crates/cast/src/cmd/interface.rs
```diff
@@ -85,27 +85,35 @@ impl InterfaceArgs {
...
+                     pragma solidity {pragma};\n\n\
+                     {}",
+                    interfaces.iter().map(|iface| &iface.source).format("\n")
+                )
+            };
             if let Some(parent) = loc.parent() {
                 fs::create_dir_all(parent)?;
             }
             fs::write(&loc, res)?;
             sh_println!("Saved interface at {}", loc.display())?;
```
figtracer: `cast --json interface ... -o file` still writes a human status line to stdout. Send it to stderr or suppress it in JSON mode.

figtracer (commented):
Three focused JSON-output regressions from the local review pass.

crates/cast/src/cmd/interface.rs
mablr: Should be `sh_status!`, right ?

figtracer: that works yeah

mablr: For the rest of output channel contract migration let's defer it to a follow-up...

stevencartavia (commented):
### `crates/cast/src/cmd/vaddr/create.rs` — mined salt is lost on register failure

The non-`--no-register` path now calls `register(...).await?` **before** printing the salt/registration_hash/master_id/virtual_addresses block. On master those values were printed first.

Reproduced with no signer (so `register` errors):

**master**
```
STDOUT:
Salt:              0x9bf4…155f
Registration hash: 0x0000…e5fd
Master ID:         0x3a96b610
Virtual addresses:
  tag=0x000000000000  0x3A96B610…0000
STDERR:
Error: cast vaddr create requires a signer (...)
```

**PR**
```
STDOUT:  (only the "Mining…/Found salt in…" lines)
STDERR:
Error: cast vaddr create requires a signer (...)
```

The mined salt is now lost on any registration failure — the user has to mine again (which can take minutes). Suggest restoring the prior order: compute the payload once, print it, then call `register`. That also lets you collapse the duplicated JSON/plain-text blocks between the `no_register` branch and the register branch (both currently emit the same payload; `print_json_success` and `print_json_object` are equivalent inside `if shell::is_json()`).

---

### `crates/cast/src/cmd/keychain.rs` — `run_list` empty case becomes an error

`run_list` now routes the empty-keys case through `sh_err!`, which writes to stderr and prefixes `Error:`. The same situation in `run_show` still uses `sh_println!`, so the two siblings disagree.

Reproduced with `TEMPO_HOME` pointing at a `keys.toml` containing `keys = []`:

```
master:  EXIT 0  STDOUT: "No keys found in keys.toml."  STDERR: (empty)
PR:      EXIT 0  STDOUT: (empty)                        STDERR: "Error: No keys found in keys.toml."
```

An empty list is a successful result, not an error. Suggest keeping `sh_println!` here for parity with `run_show` (or change both).

---

### `crates/cast/src/cmd/interface.rs` — `--json` stdout and `--json -o` file shapes diverge

`cast interface --json` to stdout now wraps the parsed ABIs in a `JsonEnvelope`, but `cast interface --json -o <file>` still writes the raw concatenated `json_abi` strings to the file. Same data, two shapes depending on whether `-o` is set.

Reproduced with a single-ABI input:

| | stdout (`--json`) | file (`--json -o`) |
|---|---|---|
| master | `[ { ...abi... } ]` | `[ { ...abi... } ]` |
| PR | `{"schema_version":1,"success":true,"data":[[{...}]],…}` | `[ { ...abi... } ]` |

If file output is intentionally a raw artifact (not wrapped), please call that out (doc/comment + test); otherwise both destinations should share the envelope shape. Note that for multiple interfaces the file content is also several JSON objects newline-joined, which isn't one valid JSON document.

crates/cast/src/args.rs
```diff
@@ -430,8 +446,11 @@ pub async fn run_command(args: CastArgs) -> Result<()> {
                     .block(block.unwrap_or(BlockId::Number(Latest)), full, fields)
                     .await?
             };
-            // JSON: Output is already formatted by `Cast::block()`
-            sh_println!("{output}")?;
+            if shell::is_json() && !is_raw_block && !has_fields {
+                print_json_object(serde_json::from_str::<Value>(&output)?)?;
+            } else {
+                print_scalar(output)?;
```
stevencartavia: Nit, follow-up only: `cast block --json` with multiple `--field` flags wraps a newline-joined string in `data`:

```
{"…","data":"0x88e9…cb6\n1",…}
```

Consumers have to split on `\n` to recover the values. Consider a follow-up that emits `"data": ["0x88e9…cb6", "1"]` or `"data": {"hash": "…", "number": "1"}` instead.

crates/cast/src/cmd/vaddr/create.rs
```diff
@@ -124,10 +124,19 @@ Master ID:         {}",
...
+            print_json_success(payload)?;
+        }
         return Ok(());
     }
 
-    register(owner, output.salt, send_tx, tx_opts).await
+    register(owner, output.salt, send_tx, tx_opts).await?;
+
+    if shell::is_json() {
+        print_json_object(payload)?;
```
stevencartavia: should these match? 
both in `if shell::is_json() {` 
one is `print_json_success(payload)?;` and the other one is `print_json_object(payload)?;`

mablr: Yikes, I missed it during a previous clean-up, fixed here 8d603b5006c6dd1ddbec7b0c6e01f69fc1cc76f7 👍

crates/cast/src/args.rs
mablr: Ack, I’ll follow up on this.

stevencartavia (dismissed):

figtracer (dismissed):

grandizzy (dismissed):
A few things to address before merge, plus some smaller suggestions.

### Blocking

**1. `cast block`/`cast tx` with `--raw` or `--field` will double-encode JSON in `--json` mode**

In `args.rs`, only the default whole-object branch parses output to `Value` before wrapping:

```rust
if shell::is_json() && !is_raw_block && !has_fields {
    print_json_object(serde_json::from_str::<Value>(&output)?)?;
} else {
    print_scalar(output)?;
}
```

That means `cast tx --json --field authorizationList`, `cast block --json --raw`, and any field whose value is itself an object/array will emit:

```json
{ "data": "{\"some\":\"json\"}" }
```

instead of a real nested object. Suggest a small helper that tries `serde_json::from_str::<Value>` and falls back to `print_scalar` on parse error, and using it for both `Block` and `Tx`. Worth adding tests for `--raw` and an object-valued `--field`.

**2. `keychain doctor --json` drops the payload's own `schema_version`**

The envelope's `schema_version` isn't the same thing as the doctor report schema — anything keying off the inner version will silently break. Please keep `schema_version` inside `data` alongside `context`/`steps`.

**3. Build hygiene**

- `interface.rs` uses bare `Value` in `serde_json::from_str::<Value>(&iface.json_abi)` — I don't see a `use serde_json::Value;` added. Either add the import or use `serde_json::Value` consistently with the rest of the file.
- `foundry_cli::json` now imports `alloy_dyn_abi::DynSolValue`. If `foundry-cli`'s `Cargo.toml` didn't already list `alloy-dyn-abi` as a direct dep, please add it rather than relying on a transitive.

### Non-blocking suggestions

- **Naming:** `print_json_object` is also used for arrays and scalars — `print_structured` or `print_json_value` would describe it better. Easier to rename now than later.
- **`cast rpc` plain mode:** the new path always parses and pretty-prints via `print_json_object`, which changes non-JSON stdout formatting too. If that's intentional, ignore; otherwise keep the old `sh_println!("{result}")` branch for `!is_json()`.
- **`mktx --expires-at`:** moving the "Transaction expires at..." line from `sh_println!` to `sh_status!` cleans up `--json` stdout (good) but also changes plain-mode stdout. Consider branching on `is_json()` if plain-mode parity matters.
- **`interface --json`:** wrapping always in an array means single-interface output gains a level of nesting compared to before — worth an explicit callout in the PR description's breaking-change list.
- **`vaddr create --json`:** the reorder so registration runs before printing is a nice fix. Worth confirming `register(...)` itself doesn't print to stdout, and adding a JSON test for the with-register path (not just `--no-register`).
- **`SelectorInfo`:** `arguments` as a comma-separated string preserves human output but an array (`["uint256","address"]`) would be friendlier for machine consumers. Also consider `stateMutability` to match Solidity ABI casing. Non-blocking since this is a new shape and you can iterate.

### Suggested additional tests

- `cast block --json` (default + `--raw` + scalar field + object field)
- `cast tx --json --field authorizationList` (or any non-scalar field)
- `cast rpc --json` against a scalar response like `eth_blockNumber`
- `cast keychain doctor --json` asserting `data.schema_version`
- `cast interface --json` for a single ABI (locks in the array shape)
- `cast vaddr create --json` without `--no-register`

mablr (commented):
@grandizzy I addressed your comment in https://github.com/foundry-rs/foundry/pull/14918/commits/fefa3998173e1bc9641e1b295bdac1cb7c740078.

I think 2. is not necessary, I just added a test to assert json shape as recommended.

figtracer (dismissed):

crates/cast/src/cmd/keychain.rs
```diff
@@ -616,8 +617,8 @@ fn run_show(wallet_address: Address) -> Result<()> {
     }
 
     if shell::is_json() {
-        let json: Vec<_> = entries.iter().map(|e| key_entry_to_json(e)).collect();
-        sh_println!("{}", serde_json::to_string_pretty(&json)?)?;
+        let entries_json: Vec<_> = entries.iter().map(|e| key_entry_to_json(e)).collect();
```
figtracer, line 782: `cast --json keychain show <wallet>` still prints plain text when no entries match the wallet. In JSON mode this should return an empty array envelope, like `keychain list`, instead of `No keys found...`.

crates/cast/src/cmd/keychain.rs
```diff
@@ -2406,17 +2407,15 @@ fn finalize_doctor(steps: Vec<DoctorStep>, context: DoctorContext) -> Result<()>
     };
 
     if shell::is_json() {
-        let json = serde_json::json!({
-            "schema_version": 1,
+        foundry_cli::json::print_json_success(serde_json::json!({
```
figtracer, line 2587: `keychain doctor --json` still drops the report payload’s own `schema_version`; the top-level version is only the `JsonEnvelope` version. Please include `"schema_version": 1` inside `data` and update the test to assert `parsed["data"]["schema_version"]`.

figtracer (commented):
Two JSON-shape regressions from the local review pass.

crates/cast/src/cmd/keychain.rs
mablr: Already got this from clanker review, but i don't think we need this.

crates/cast/src/cmd/keychain.rs
mablr: https://github.com/foundry-rs/foundry/pull/14918/commits/afc368a04ab74888f1ecdbcf160d4bdc49df6131

figtracer (dismissed):

crates/cast/src/cmd/vaddr/create.rs
```diff
@@ -124,10 +124,19 @@ Master ID:         {}",
...
+            print_json_success(payload)?;
+        }
         return Ok(());
     }
 
-    register(owner, output.salt, send_tx, tx_opts).await
+    register(owner, output.salt, send_tx, tx_opts).await?;
+
+    if shell::is_json() {
+        print_json_success(payload)?;
```
grandizzy, line 145: this probably isn't json but we need to construct as
```diff
-    register(owner, output.salt, send_tx, tx_opts).await?;
-
-    if shell::is_json() {
-        print_json_success(payload)?;
-    }
+    let tx_hash = register(owner, output.salt, send_tx, tx_opts).await?;
+    if shell::is_json() {
+        let mut payload = payload;
+        payload["registration_tx_hash"] = json!(format!("{tx_hash:#x}"));
+        print_json_success(payload)?;
+    }
```

crates/cast/src/cmd/erc20.rs
```diff
@@ -521,9 +522,9 @@ impl Erc20Subcommand {
                     .await?;
 
                 if shell::is_json() {
-                    sh_println!("{}", serde_json::to_string(&balance.to_string())?)?
+                    print_json_success(balance.to_string())?;
                 } else {
-                    sh_println!("{balance}")?
+                    sh_println!("{}", format_uint_exp(balance))?;
```
grandizzy: any reason to change this (e.g. `1000000000000000000000` to `1000.0…`) could be breaking change ig

crates/cast/src/cmd/rpc.rs
```diff
@@ -55,12 +54,8 @@ impl RpcArgs {
...
         let provider = utils::get_provider(&config)?;
         let result = Cast::new(provider).rpc(&method, params).await?;
-        if shell::is_json() {
-            let result: serde_json::Value = serde_json::from_str(&result)?;
-            sh_println!("{}", serde_json::to_string_pretty(&result)?)?;
-        } else {
-            sh_println!("{}", result)?;
-        }
+        let result: serde_json::Value = serde_json::from_str(&result)?;
+        print_json_object(result)?;
```
grandizzy: `print_json_object` in non-JSON mode does `serde_json::to_string_pretty(...)`. So plain-mode cast rpc eth_getBlockByNumber 0x123 false now goes from the compact server response to multi-line pretty JSON. Should we keep old behavior?

grandizzy (changes requested):
looks good, couple of more comments, pls check!

crates/cast/src/cmd/erc20.rs
mablr: bad merge conflict, I'll revert

crates/cast/src/cmd/vaddr/create.rs
mablr: https://github.com/foundry-rs/foundry/pull/14918/commits/8959557ab4c18c9794e8999960d91e542ba8dc3d

crates/cast/src/cmd/erc20.rs
mablr: https://github.com/foundry-rs/foundry/pull/14918/commits/30fa82854e1869329921736e4189fba5c9ed2288

crates/cast/src/cmd/rpc.rs
mablr: Yes I over-factorized, restored old pattern here https://github.com/foundry-rs/foundry/pull/14918/commits/636e3b3704d072f90ba83511508e40a63e779a25

grandizzy (approved):
lgtm

stevencartavia (approved):

PR #14919, opened 2026-05-26
fix(forge): preserve configured EVM version across forks

figtracer:
## Motivation
Fork creation can rebuild the fork environment from chain/default hardfork information and lose the currently configured execution spec. This makes tests that set an EVM version before creating/selecting a fork observe a different EVM version after the fork is active.

## Solution
- Carry the configured EVM version through `CreateFork` from both initial fork opts and fork cheatcodes.
- Include the EVM version in fork IDs so cached/in-progress forks do not get reused across different specs.
- Re-apply the configured spec to the fork `EvmEnv` after fetching the fork block environment.
- Add a regression test that asserts `vm.createSelectFork` preserves `default.evm_version = "cancun"`.

## Testing
- `cargo check -p foundry-evm-core -p foundry-cheatcodes -p forge --tests`
- `cargo test -p forge test_fork_keeps_configured_evm_version --test cli -- --nocapture`

### DIFF
--- a/crates/cheatcodes/src/evm/fork.rs
+++ b/crates/cheatcodes/src/evm/fork.rs
@@ -10,6 +10,7 @@ use alloy_provider::Provider;
 use alloy_rpc_types::Filter;
 use alloy_sol_types::SolValue;
 use foundry_common::provider::ProviderBuilder;
+use foundry_config::ExecutionSpec;
 use foundry_evm_core::{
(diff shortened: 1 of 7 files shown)


grandizzy:
@figtracer wonder how this works with the cheatcode for changing evm, mind to add a test?

figtracer:
Closing: opened from the wrong local workstream by mistake.

PR #14923, opened 2026-05-26
fix(forge): restore EIP-170/EIP-3860 labels in size-report errors when limits are default

stevencartavia:
Follow-up to #14761. Build/size-report failure messages now include the EIP reference when the configured limits equal the Ethereum defaults, and stay generic when the user has overridden them.


grandizzy (approved):
lgtm

mablr (approved):

PR #14924, opened 2026-05-26
fix(script): honor merged `code_size_limit` config in `check_contract_sizes`

stevencartavia:
Follow-up to #14761. `forge script`'s contract-size check now reads `code_size_limit` from the merged `Config` (foundry.toml + env + CLI) instead of only the raw `--code-size-limit` CLI flag.


mablr (approved):
sgtm

mattsse (approved):
LGTM

PR #14938, opened 2026-05-27
fix(script): fall back to chain-0 keys.toml entry

mablr:
## Motivation
Fix the forge script broadcast regression introduced in #14916.

## Solution

`lookup_signer_for_chain` now accepts entries where `chain_id` was omitted (defaults to 0 via serde) when no exact `(wallet_address, chain_id)` match exists. 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/script/src/broadcast.rs
+++ b/crates/script/src/broadcast.rs
@@ -336,7 +336,11 @@ fn lookup_signer_in(from: Address, chain: u64, file: &KeysFile) -> Result<TempoL
             continue;
         }
         if entry.chain_id == chain {
-            return build_lookup(entry);
+            if entry.key.is_some() {
+                return build_lookup(entry);
+            }
+            // exact chain match but no key -> keep searching for a fallback
+            continue;
         }
         if entry.chain_id == 0 && fallback.is_none() {
             fallback = Some(entry);
--- a/crates/script/src/broadcast.rs
+++ b/crates/script/src/broadcast.rs
@@ -28,7 +28,8 @@ use foundry_common::{
     provider::{ProviderBuilder, try_get_http_provider},
     shell,
     tempo::{
-        ResolvedSessionSigner, TempoSponsor, WALLET_KEYS_PATH, decode_key_authorization, tempo_home,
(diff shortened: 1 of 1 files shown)


crates/script/src/broadcast.rs
```diff
@@ -305,33 +324,25 @@ pub(crate) fn lookup_signer_for_chain(from: Address, chain: u64) -> Result<Tempo
...
-        let key_authorization =
-            entry.key_authorization.as_deref().map(decode_key_authorization).transpose()?;
-        let config = TempoAccessKeyConfig {
-            wallet_address: entry.wallet_address,
-            key_address,
-            key_authorization,
-        };
-        return Ok(TempoLookup::Keychain(signer, Box::new(config)));
+        }
+        if entry.chain_id == chain {
```
figtracer, line 363: this now stops on the first exact-chain entry even if that entry has no inline key, because `build_lookup` returns `NotFound`; maybe we should continue the search?

crates/script/src/broadcast.rs
```diff
@@ -291,11 +292,29 @@ fn insert_session_access_key_for_remaining_transactions(
...
 
+fn build_lookup(entry: &KeyEntry) -> Result<TempoLookup> {
+    let Some(ref key) = entry.key else {
+        return Ok(TempoLookup::NotFound);
+    };
+    let signer = foundry_wallets::utils::create_private_key_signer(key)?;
+    let Some(key_address) = entry.key_address.filter(|ka| *ka != entry.wallet_address) else {
+        return Ok(TempoLookup::Direct(signer));
+    };
+    let key_authorization =
```
figtracer, line 303: now that `chain_id == 0` is treated as a wildcard, this can carry a `key_authorization` created for a different chain into the requested broadcast chain

figtracer (commented):
nits

crates/script/src/broadcast.rs
mablr: Hmm right, the other (more robust) solution was to enforce `chain_id` presence in `keys.toml`, unsure how badly breaking it is ?

figtracer (approved):

mattsse (approved):

PR #14939, opened 2026-05-27
fix(symbolic): fail closed on gasleft

figtracer:
## Summary

- return Unsupported for the EVM GAS opcode / Solidity gasleft() in symbolic execution


grandizzy (approved):
lgtm

PR #14940, opened 2026-05-27
fix(symbolic): surface Keccak heuristic in incomplete results

figtracer:
## Summary

- attribute invalid solver models involving symbolic Keccak constraints to the symbolic Keccak heuristic
- unignore the Keccak surfacing regression from #14936
- keep existing symbolic Keccak/mapping behavior, but avoid generic incomplete errors with no Keccak vocabulary


grandizzy (approved):
lgtm

PR #14942, opened 2026-05-27
feat(symbolic): allow GAS as call operand

figtracer:
## Summary
- add a first-class opaque `GAS` / `gasleft()` token instead of treating gas as a normal SMT variable or concrete max value
- allow only a raw opaque gas token as the gas operand of `CALL`, `CALLCODE`, `DELEGATECALL`, and `STATICCALL`
- reject gas-tainted values when they cross from the unmodeled gas domain into modeled non-gas behavior or proof obligations: branch conditions, assertions, assumptions, solver queries, calldata, call target/value/copy bounds, logs, and derived call gas
- add regressions for accepted raw call gas plus rejected `gasleft()` branches/assertions, derived call gas, and gas-tainted call calldata

## Why
Solidity commonly emits `GAS` immediately before external calls. The symbolic executor previously rejected `GAS` entirely, which made ordinary Solidity calls unsupported before reaching the interesting assertions in George's `symbolic-bug-suite`.

This PR keeps the executor explicitly gas-oblivious: it does not model gas consumption, OOG, EIP-150 forwarding, or gas-sensitive control flow. `GAS` now produces an opaque gas-budget token that may be consumed only in the call gas slot that this model already abstracts away. The token is not exposed to the solver as a free variable, and derived/observed gas use still reports incomplete instead of producing non-replaying counterexamples.

## Bug-suite check
Against `grandizzy/symbolic-bug-suite` with the minimal local `Test.sol` shim:

| Branch | Result |
|---|---|
| current merged `mattsse/native-symbolic-testing` | 0/16 useful results; all incomplete on `GAS/gasleft() not modeled` |
| this PR | reaches the underlying properties: targeted/serial runs produce 15/16 counterexamples; 1/16 still incomplete on `symbolic vm.deal value` (`TheDAOReentrancyTest`) |

Note: the slow empty-market/MonoX/zkLend cases are solver-heavy and can timeout in one fully parallel suite run on this machine, but targeted reruns find the same counterexamples once `GAS` no longer blocks execution.

### DIFF
--- a/crates/evm/symbolic/src/executor/calls.rs
+++ b/crates/evm/symbolic/src/executor/calls.rs
@@ -13,7 +13,11 @@ impl SymbolicExecutor {
         let pre_call_state = state.clone();
         let call_pc = state.pc.saturating_sub(1);
         let gas = state.stack.pop()?;
+        if gas.contains_gasleft() && !gas.is_raw_gasleft() {
+            return Err(SymbolicError::Unsupported("GAS/gasleft() not modeled"));
+        }
         let target = state.stack.pop()?;
+        ensure_word_not_gasleft(&target)?;
         let target_address = state.world.resolve_address(&target);
         let value = match (kind, target_address) {
             (CallKind::Call, Some(to)) if is_known_cheatcode(to) => {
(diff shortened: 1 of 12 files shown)


grandizzy (approved):
lgtm!

PR #14943, opened 2026-05-27
feat(lint): add arbitrary-send-eth lint

stevencartavia:
Flags functions that send ETH to a caller-controlled destination without restricting who can call them.

Closes OSS-22


mablr (commented):
Some clanker findings, I've been able to repro locally.

### FN-1 — Numeric address literals accepted as safe destinations

**Source:** [`arbitrary_send_eth.rs#L246`](https://github.com/foundry-rs/foundry/blob/f51083e9d/crates/lint/src/sol/high/arbitrary_send_eth.rs#L246)

The destination safety check matches `LitKind::Address(_) | LitKind::Number(_)`. Any `address(N)` expression resolves to a `LitKind::Number` node and is silently accepted as safe — the lint never warns. Only `address(0)` and checksummed address literals are genuinely safe destinations; arbitrary numeric casts are not.

```solidity
payable(address(1)).transfer(amt); // no warning emitted — should warn
```

---

### FP-1 — Trailing `return;` disqualifies a valid caller-restriction helper

**Source:** [`arbitrary_send_eth.rs#L827`](https://github.com/foundry-rs/foundry/blob/f51083e9d/crates/lint/src/sol/high/arbitrary_send_eth.rs#L827)

`function_no_arg_returns` rejects any helper body that contains a `return` statement, including a no-op trailing `return;`. A helper like `if (msg.sender != owner) revert(); return;` is semantically identical to the same function without the final statement, but the modifier is not recognised as caller-restricting and the transfer is falsely flagged.

```solidity
function _checkOwner() internal view {
    if (msg.sender != owner) revert();
    return; // ← causes the whole helper to be ignored
}
modifier onlyOwner() { _checkOwner(); _; }
function withdraw(address payable to, uint256 amt) external onlyOwner {
    to.transfer(amt); // falsely flagged
}
```

---

### FP-2 — Named return variable not recognised as a `msg.sender`-returning helper

**Source:** [`arbitrary_send_eth.rs#L800`](https://github.com/foundry-rs/foundry/blob/f51083e9d/crates/lint/src/sol/high/arbitrary_send_eth.rs#L800)

`function_no_arg_returns` only matches `StmtKind::Return(Some(e))` as the sole statement. A helper that assigns the named return variable instead of using an explicit `return` expression produces a different HIR shape and is not recognised, so the modifier is ignored and the transfer is falsely flagged.

```solidity
function _msgSender() internal view returns (address r) { r = msg.sender; }
modifier onlyOwner() { require(_msgSender() == owner); _; }
function withdraw(address payable to, uint256 amt) external onlyOwner {
    to.transfer(amt); // falsely flagged
}
```

crates/lint/src/sol/high/arbitrary_send_eth.rs
```diff
@@ -0,0 +1,1930 @@
...
+            ItemId, LoopSource, Res, StmtKind, TypeKind, Visit,
+        },
+    },
+};
+use std::{collections::HashSet, ops::ControlFlow};
+
+declare_forge_lint!(
+    ARBITRARY_SEND_ETH,
+    Severity::High,
+    "arbitrary-send-eth",
```
figtracer, line 22: this registers the new help URL before the page exists in the published book, so `lint::ensure_lint_rule_docs` fails live with `arbitrary-send-eth (...) returned HTTP 404 Not Found`. i think this needs the matching foundry-rs/book page/deploy, or the docs check needs a path that works for newly added lints before publish.

crates/lint/src/sol/high/arbitrary_send_eth.rs
```diff
@@ -0,0 +1,1930 @@
...
+            ExprKind::Lit(lit) => match &lit.kind {
+                LitKind::Address(_) => true,
+                LitKind::Number(n) => n.is_zero(),
+                _ => false,
+            },
+            ExprKind::Ident(reses) => reses.iter().any(|r| match r {
+                Res::Item(ItemId::Variable(vid)) => self.is_safe_var(*vid),
+                _ => false,
+            }),
+            ExprKind::Call(callee, args, _) if is_address_like_cast_callee(callee) => {
```
figtracer: this only unwraps address-like casts, so `payable(address(uint160(0))).transfer(amount)` still warns even though `address(0)` is treated as safe. i think the numeric-literal fix still needs to handle the common `uint160(0)` spelling.

crates/lint/src/sol/high/arbitrary_send_eth.rs
```diff
@@ -0,0 +1,1930 @@
...
+        return Some(dest);
+    }
+
+    if let Some(opts) = opts
+        && opts.iter().any(|arg| arg.name.name == sym::value && !is_literal_zero(&arg.value))
+    {
+        let callee_inner = callee.peel_parens();
+        match &callee_inner.kind {
+            ExprKind::Member(recv, _) if !is_address_self(recv) => return Some(recv),
+            ExprKind::Ident(reses)
```
figtracer, line 592: this treats every function-pointer value call as an arbitrary destination, so `function() external payable callback = this.receiveEth; callback{value: amount}();` still warns even though the destination is `this`. i think we need to track function-pointer facts or avoid flagging pointers proven to come from `this`.

crates/lint/src/sol/high/arbitrary_send_eth.rs
```diff
@@ -0,0 +1,1930 @@
...
+    }
+    let ExprKind::Ident(reses) = &callee.peel_parens().kind else { return false };
+    reses.iter().any(|r| match r {
+        Res::Item(ItemId::Function(fid)) => {
+            if stack.contains(fid) {
+                return false;
+            }
+            let f = hir.function(*fid);
+            let Some(body) = f.body else { return false };
+            if body.stmts.iter().any(stmt_contains_return) {
```
figtracer: this still rejects caller-restriction helpers that contain a trailing bare `return;`, so the previous trailing-return fix does not cover `_checkOwner()`-style modifiers. i reproduced `_checkOwner() { if (msg.sender != owner) revert(); return; }` used by `onlyOwner`, and the guarded `to.transfer(amount)` still warns.

stevencartavia: [b4642e1](https://github.com/foundry-rs/foundry/pull/14943/commits/b4642e192bfe8e7eb9156273db5bf960658b5e30)

crates/lint/src/sol/high/arbitrary_send_eth.rs
stevencartavia: [b4642e1](https://github.com/foundry-rs/foundry/pull/14943/commits/b4642e192bfe8e7eb9156273db5bf960658b5e30)

crates/lint/src/sol/high/arbitrary_send_eth.rs
stevencartavia: [b4642e1](https://github.com/foundry-rs/foundry/pull/14943/commits/b4642e192bfe8e7eb9156273db5bf960658b5e30)

crates/lint/src/sol/high/arbitrary_send_eth.rs
stevencartavia: just needs the book PR

crates/lint/src/sol/high/arbitrary_send_eth.rs
```diff
@@ -0,0 +1,2008 @@
...
+    out_safe: &mut HashSet<hir::VariableId>,
+) {
+    let ItemId::Function(fid) = invocation.id else { return };
+    let Some((modifier, prefix)) = modifier_prefix(hir, fid) else { return };
+    let arg_map: Vec<(hir::VariableId, hir::VariableId)> = invocation
+        .args
+        .exprs()
+        .enumerate()
+        .filter_map(|(i, arg)| Some((*modifier.parameters.get(i)?, underlying_var(arg)?)))
+        .collect();
```
grandizzy: ### FN — Named modifier arguments mapped positionally, can mark wrong destination safe

`collect_modifier_safety` builds `arg_map` by zipping `invocation.args.exprs()` with `modifier.parameters` by index. When the modifier is invoked with named args whose source order differs from the parameter order, the safety fact recovered from the modifier prefix is bound to the wrong call-site variable — the lint silently no-warns on a genuinely arbitrary destination.

`arg_for_param` (L1201–L1214) already handles the named-vs-positional distinction and is used elsewhere in this file; it should be reused here.

```solidity
modifier check(address payable who, address payable ignored) {
    require(who == payable(msg.sender));
    _;
}

function withdraw(address payable to, address payable other, uint256 amt)
    external
    check({ignored: to, who: other}) // guard actually constrains `other`
{
    to.transfer(amt); // not flagged — `to` is incorrectly marked safe
}
```

Suggested fix:

```suggestion
    let arg_map: Vec<(hir::VariableId, hir::VariableId)> = modifier
        .parameters
        .iter()
        .filter_map(|&mp| {
            let arg = arg_for_param(hir, modifier, mp, &invocation.args)?;
            Some((mp, underlying_var(arg)?))
        })
        .collect();
```

Worth adding a regression test in `ArbitrarySendEth.sol` for a modifier invoked with named args in non-source order.

stevencartavia: [9840bc6](https://github.com/foundry-rs/foundry/pull/14943/commits/9840bc6767a2f47b871258175d020f5f17af7275)

grandizzy (approved):
lgtm

mablr (approved):
sgtm

PR #14949, opened 2026-05-28
chore(forge): clean up invariant campaign output and progress labels

0xKarl98:
Removes remaining invariant-anchor-oriented reporting from invariant campaign output and progress tracking

Ref : 
https://github.com/foundry-rs/foundry/pull/14905#pullrequestreview-4360892941

cc @grandizzy

### DIFF
--- a/crates/evm/evm/src/executors/invariant/mod.rs
+++ b/crates/evm/evm/src/executors/invariant/mod.rs
@@ -46,8 +46,6 @@ use std::{
     time::{Duration, Instant, SystemTime, UNIX_EPOCH},
 };
 
-const INVARIANT_CAMPAIGN_DISPLAY_NAME: &str = "Invariant/Property Tests";
-
 mod error;
 pub use error::{
     FailureKey, HandlerAssertionFailure, InvariantFailures, InvariantFuzzError,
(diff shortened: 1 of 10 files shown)


grandizzy (changes requested):
thanks, looks on the right direction - let's address also the invariant metrics table printed multiple times in BuggyBank case, should really be shown only once with this approach
Also, instead `Invariant/Property Tests` we could do smth like `{ContractName} invariants`, wdyt?

grandizzy:
> thanks, looks on the right direction - let's address also the invariant metrics table printed multiple times in BuggyBank case, should really be shown only once with this approach Also, instead `Invariant/Property Tests` we could do smth like `{ContractName} invariants`, wdyt?

@0xKarl98 actually rechecked and table doesn't seem to be printed multiple times, so we should be good there. I'd only check the `{ContractName} invariants` update and maybe a small nit to show passed invariants in green even if there's a failure (that is if not too much of a hassle)

image

grandizzy (dismissed):
lgtm! let's fix test and shoud be gtg, thanks!

0xKarl98:
Have fixed , plz check it out

grandizzy (approved):
lgtm!

mablr (approved):

PR #14727, opened 2026-05-12
feat(common): output channel foundation

stevencartavia:
Foundation for enforcing stdout/stderr discipline across Foundry CLIs.

## What this adds
- Contract doc `docs/dev/output-channels.md` with a target per-command stdout table.
- `sh_status!` and `sh_progress!` macros in `foundry_common::io` (the latter tty+quiet gated).
- `Shell::captured()` test fixture + routing/quiet contract tests.
- Moves the compilation spinner and `prompt!` question text to stderr.
- CONTRIBUTING note + accurate description of the existing `disallowed-macros` lint.

## Intentional behavior changes
- TTY-mode compile spinner/progress: stdout → stderr.
- `prompt!` question: stdout → stderr (automation scraping prompts from stdout will need to switch).

## Deferred to follow-ups
- Non-TTY compile reporter still on stdout (flipping it shifts many snapshot tests).
- `sh_print!`/`sh_println!` still suppressed by `--quiet`; target is the opposite, will flip with command migrations.
- Per-command stdout migrations across forge/cast/verify/anvil/chisel/script.

Part of OSS-224

PR #14950, opened 2026-05-28
feat(forge): migrate forge remappings to output channel contract

stevencartavia:
Migrates `forge remappings --pretty` to the stdout/stderr contract from #14727.

- Prose headers (`Context:`, `Global:`) and the visual blank separator move to stderr via `sh_status!` / `sh_eprintln!`.
- Each remapping stays on stdout via `sh_println!`, with the `- ` bullet dropped.
- Default (non-pretty) mode unchanged — already compliant.

Result: `forge remappings 2>/dev/null` and `forge remappings --pretty 2>/dev/null` produce identical stdout (one remapping per line).

Snapshot test split into `.stdout_eq` / `.stderr_eq`. Contract doc row flipped `todo` → `migrated`.

Part of OSS-224

### DIFF
--- a/crates/forge/src/cmd/remappings.rs
+++ b/crates/forge/src/cmd/remappings.rs
@@ -29,17 +29,18 @@ impl RemappingArgs {
                 groups.entry(remapping.context.clone()).or_default().push(remapping);
             }
             for (group, remappings) in groups {
+                // Prose headers are diagnostics — keep stdout to one remapping per line.
                 if let Some(group) = group {
-                    sh_println!("Context: {group}")?;
+                    sh_status!("Context: {group}")?;
                 } else {
-                    sh_println!("Global:")?;
+                    sh_status!("Global:")?;
                 }
 
                 for mut remapping in remappings {
                     remapping.context = None; // avoid writing context twice
-                    sh_println!("- {remapping}")?;
+                    sh_println!("{remapping}")?;
                 }
-                sh_println!()?;
+                sh_eprintln!()?;
             }
         } else {
             for remapping in config.remappings {
--- a/crates/forge/src/cmd/remappings.rs
+++ b/crates/forge/src/cmd/remappings.rs
@@ -29,15 +29,13 @@ impl RemappingArgs {
                 groups.entry(remapping.context.clone()).or_default().push(remapping);
             }
             for (group, remappings) in groups {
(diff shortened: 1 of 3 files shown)


crates/forge/src/cmd/remappings.rs
```diff
@@ -29,17 +29,18 @@ impl RemappingArgs {
...
                 if let Some(group) = group {
-                    sh_println!("Context: {group}")?;
+                    sh_status!("Context: {group}")?;
                 } else {
-                    sh_println!("Global:")?;
+                    sh_status!("Global:")?;
                 }
 
                 for mut remapping in remappings {
                     remapping.context = None; // avoid writing context twice
```
grandizzy: This silently breaks the PR's headline invariant ("`forge remappings 2>/dev/null` and `forge remappings --pretty 2>/dev/null` produce identical stdout") whenever a remapping has a non-global context.

Default mode prints `myctx:lib/=node_modules/` on stdout. Pretty mode strips the context here, so stdout becomes just `lib/=node_modules/` — context is lost entirely for machine consumers, since `Context: foo` now lives on stderr.

Suggest dropping the mutation and printing the remapping as-is:

```rust
for remapping in remappings {
    sh_println!("{remapping}")?;
}
```

The `Context:` header on stderr then becomes a pure human aid, and stdout is genuinely identical between modes. Worth adding a multi-context fixture to the snapshot test to lock this in — the current test only exercises the `Global` branch, which is why this slipped through.

grandizzy: Stale comment: `// avoid writing context twice` no longer matches reality. The context header is now on stderr via `sh_status!`, not stdout, so nothing would be written "twice" even without the mutation. If the mutation stays, reword to something like `// header above already shows the context`; if you drop the mutation per the other comment, this line goes away entirely.

stevencartavia: [baa9439](https://github.com/foundry-rs/foundry/pull/14950/commits/baa94396ea5f069fcba88056e92fed78702c4e0d)

stevencartavia: [baa9439](https://github.com/foundry-rs/foundry/pull/14950/commits/baa94396ea5f069fcba88056e92fed78702c4e0d)

grandizzy (approved):
lgtm

mablr (approved):

PR #14951, opened 2026-05-28
feat(forge): migrate forge flatten to output channel contract

stevencartavia:
Migrates `forge flatten` to the stdout/stderr contract from #14727.

Default mode was already compliant (flattened source on stdout). With `--output <path>`, the "Flattened file written at …" status string moved from stdout to stderr via `sh_status!` — the file is the result, stdout stays empty.

Verified:
- `forge flatten foo.sol 2>/dev/null` → flattened source
- `forge flatten foo.sol -o out.sol 2>/dev/null` → empty stdout, file written

Part of OSS-224


grandizzy (approved):
lgtm

mablr (approved):

PR #14952, opened 2026-05-28
feat(symbolic): support symbolic `vm.deal` values

figtracer:
## Summary
- allow `vm.deal(address,uint256)` to store symbolic balance values instead of requiring the amount to be concrete/constrained
- re-concretize already-constrained deal values before writing balance state to preserve precision for existing cases
- reject gas-tainted deal values so opaque `GAS` cannot leak into world-state balances
- add CLI regressions for symbolic deal-funded value transfers, insufficient symbolic balances, and direct/derived gas-derived deal values failing closed

## Why
`vm.deal` is balance assignment, and the symbolic world already supports symbolic `BALANCE` reads, symbolic transfer arithmetic, and solver-backed value-transfer solvency checks. Requiring the deal amount to be concrete was overly conservative and blocked useful tests before they reached the property under test.

This keeps the model bounded: it supports symbolic balance/value-transfer semantics, but does not claim full account-existence path splitting for `amount == 0` vs `amount != 0` or gas-derived balance semantics.

## Generic coverage
This is not special-cased for a bug-suite contract. The regressions cover the general balance semantics:

- symbolic `vm.deal` balance can be read back through `BALANCE`
- symbolic balance can fund a value transfer of the same symbolic amount
- symbolic insufficient-balance transfer fails and preserves balances
- direct `gasleft()` and derived `gasleft() + 1` deal values fail closed

## Bug-suite check
Against `grandizzy/symbolic-bug-suite`, the previous remaining targeted blocker after #14942 was TheDAO stopping on `symbolic vm.deal value`. With this PR:

```text
forge test --symbolic --match-contract TheDAOReentrancyTest --match-test checkWithdrawCannotBeReentered
[FAIL: panic: assertion failed (0x01); counterexample: calldata=0xbdaea2a8... args=[36893488147419103232 [3.689e19]]]
```

So the test now reaches a real counterexample instead of an unsupported symbolic `vm.deal` value.

### DIFF
--- a/crates/evm/symbolic/src/executor/cheatcodes.rs
+++ b/crates/evm/symbolic/src/executor/cheatcodes.rs
@@ -1237,9 +1237,12 @@ impl SymbolicExecutor {
         }
         if selector == selector!("deal(address,uint256)") {
             let target = read_abi_address_or_symbolic_slot_arg(state, args_offset, 0)?;
-            let value =
-                read_abi_constrained_word_arg(state, args_offset, 1, "symbolic vm.deal value")?;
-            state.world.set_balance(target, value);
+            let value = read_abi_word_arg(&state.memory, args_offset, 1)?;
+            if value.contains_gasleft() {
+                return Err(SymbolicError::Unsupported("GAS/gasleft() not modeled"));
+            }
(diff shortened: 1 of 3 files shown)


grandizzy (approved):
lgtm!

PR #14954, opened 2026-05-28
feat(lint): add `missing-events-access-control` lint

figtracer:
## Summary

- add the low-severity `missing-events-access-control` lint
- detect protected public/external state changes to access-control state when no event is emitted
- cover ownership, guardian, role mapping, nested role-helper, and inherited-modifier cases in UI fixtures
- document and register the lint

## Notes

The detector stays conservative to avoid noisy Slither-style false positives: it requires an input-tainted write, skips unprotected setters and fixed writes, and skips functions that emit an event directly or through an internal helper.

## Validation

- `cargo fmt --package forge-lint`
- `cargo test -p forge-lint`
- `cargo test -p forge --test ui -- MissingEventsAccessControl`

### DIFF
--- a/crates/lint/src/sol/low/missing_events_access_control.rs
+++ b/crates/lint/src/sol/low/missing_events_access_control.rs
@@ -7,7 +7,7 @@ use crate::{
     },
 };
 use solar::{
-    ast::{ContractKind, StateMutability, Visibility},
+    ast::{ContractKind, DataLocation, LitKind, StateMutability, Visibility},
     interface::{Span, kw, sym},
     sema::hir::{self, ExprKind, FunctionId, ItemId, Res, StmtKind, VariableId},
 };
@@ -38,17 +38,20 @@ impl<'hir> LateLintPass<'hir> for MissingEventsAccessControl {
 
         for func_id in contract.all_functions() {
             let func = hir.function(func_id);
-            if !is_protected_entry_point(hir, func_id, func)
-                || function_emits_event(hir, func_id, &mut HashSet::new())
-            {
+            if !is_protected_entry_point(hir, func_id, func) {
                 continue;
             }
 
-            let mut analyzer = WriteAnalyzer::new(hir, &access_control_vars);
+            let guard_vars = entry_point_access_guard_vars(hir, func_id, func);
+            let mut analyzer = WriteAnalyzer::new(hir, &access_control_vars, &guard_vars);
             let writes = analyzer.analyze_entry_point(func_id);
             let mut emitted = HashSet::new();
 
             for write in writes {
+                if write.evented {
+                    continue;
+                }
+
                 if !emitted.insert(write.var_id) {
                     continue;
                 }
@@ -102,12 +105,21 @@ fn access_control_state_vars(
                     hir,
                     modifier_id,
                     &mut HashSet::new(),
+                    &mut HashSet::new(),
                     &mut out,
+                    true,
                 );
             }
         }
 
-        collect_access_control_state_vars_in_function(hir, func_id, &mut HashSet::new(), &mut out);
+        collect_access_control_state_vars_in_function(
+            hir,
+            func_id,
+            &mut HashSet::new(),
+            &mut HashSet::new(),
+            &mut out,
+            false,
+        );
     }
 
     out.retain(|var_id| {
@@ -117,11 +129,44 @@ fn access_control_state_vars(
     out
 }
 
+fn entry_point_access_guard_vars(
+    hir: &hir::Hir<'_>,
+    func_id: FunctionId,
+    func: &hir::Function<'_>,
+) -> HashSet<VariableId> {
+    let mut out = HashSet::new();
+
+    for modifier in func.modifiers {
+        if let Some(modifier_id) = modifier.id.as_function() {
+            collect_access_control_state_vars_in_function(
+                hir,
+                modifier_id,
+                &mut HashSet::new(),
+                &mut HashSet::new(),
+                &mut out,
+                true,
+            );
+        }
+    }
+
+    collect_access_control_state_vars_in_function(
+        hir,
+        func_id,
+        &mut HashSet::new(),
+        &mut HashSet::new(),
+        &mut out,
+        false,
+    );
+    out
+}
+
 fn collect_access_control_state_vars_in_function(
     hir: &hir::Hir<'_>,
     func_id: FunctionId,
     seen: &mut HashSet<FunctionId>,
+    sender_aliases: &mut HashSet<VariableId>,
     out: &mut HashSet<VariableId>,
+    stop_at_placeholder: bool,
 ) {
     if !seen.insert(func_id) {
         return;
@@ -131,7 +176,10 @@ fn collect_access_control_state_vars_in_function(
     let Some(body) = func.body else { return };
 
     for stmt in body.stmts {
-        collect_access_control_state_vars_in_stmt(hir, stmt, seen, out);
+        if stop_at_placeholder && matches!(stmt.kind, StmtKind::Placeholder) {
+            break;
+        }
+        collect_access_control_state_vars_in_stmt(hir, stmt, seen, sender_aliases, out);
     }
 }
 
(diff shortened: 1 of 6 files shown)


crates/lint/testdata/MissingEventsAccessControl.sol
```diff
@@ -0,0 +1,225 @@
+//@compile-flags: --only-lint missing-events-access-control
+
```
mablr, line 2: The fixture should add: unrelated event after write (must warn), logging modifier (must warn), event in opposite branch (must warn), `acceptOwnership/owner = msg.sender` (must warn), `require(msg.sender == computeAddress(x))` with unrelated state setter (must not warn), local-overwrite (must not warn), revert-path taint (must not warn), modifier-checks-after-_ (must not be considered protected), modifier-write, and storage-alias mapping writes.

mablr (commented):
Clanker's findings confirmed locally:

---

### FN-1 — Any `emit` in a function (or its modifiers) silences the entire function

**Code location:**
[`function_emits_event` L1172](https://github.com/foundry-rs/foundry/blob/974ed9af0cc97746cd1ac9fd04e6dfc7e1798a74/crates/lint/src/sol/low/missing_events_access_control.rs#L1172-L1192),
call site at [L41](https://github.com/foundry-rs/foundry/blob/974ed9af0cc97746cd1ac9fd04e6dfc7e1798a74/crates/lint/src/sol/low/missing_events_access_control.rs#L41).

`function_emits_event` is checked once at the entry-point level. If *any* `emit` is found anywhere
in the call graph — including in an unrelated modifier body, inside an `if`-branch that does not
cover the write, or before the write — the entire function is skipped and no warning is emitted.
The existing `setWithUnrelatedEvent` in the fixture bakes this in as an intentional "SHOULD PASS"
case when it should actually warn.

---

### FN-2 — State-variable → state-variable writes are never reported

**Code location:**
[`write_is_reportable` L674](https://github.com/foundry-rs/foundry/blob/974ed9af0cc97746cd1ac9fd04e6dfc7e1798a74/crates/lint/src/sol/low/missing_events_access_control.rs#L674-L676).

`write_is_reportable` requires `!sources.is_empty()`, where `sources` is the set of function-input
parameters whose taint reaches the write. A state variable (e.g. `pendingOwner`) carries no input
taint; therefore `owner = pendingOwner` and `pendingOwner = address(0)` are never reported, even
when they change access-control state. The canonical `acceptOwnership` pattern is silently allowed.

---

### FN-3 — `msg.sender` stored in a local variable defeats access-guard detection

**Code location:**
[`expr_looks_like_access_check` L947](https://github.com/foundry-rs/foundry/blob/974ed9af0cc97746cd1ac9fd04e6dfc7e1798a74/crates/lint/src/sol/low/missing_events_access_control.rs#L947-L950) →
[`expr_reads_sender` L1041](https://github.com/foundry-rs/foundry/blob/974ed9af0cc97746cd1ac9fd04e6dfc7e1798a74/crates/lint/src/sol/low/missing_events_access_control.rs#L1041-L1092).

`expr_reads_sender` only recognises `msg.sender` / `tx.origin` member accesses directly; it does
not follow local-variable aliases. A modifier that copies `msg.sender` into a local before the
`require` comparison is not recognised as an access guard, so writes in functions using it are
missed entirely.

---

### FP-1 — Local-variable overwrite does not clear taint; spurious write warning

**Code location:**
[`analyze_expr` L527](https://github.com/foundry-rs/foundry/blob/974ed9af0cc97746cd1ac9fd04e6dfc7e1798a74/crates/lint/src/sol/low/missing_events_access_control.rs#L527-L546),
[`set_local_taint` L666](https://github.com/foundry-rs/foundry/blob/974ed9af0cc97746cd1ac9fd04e6dfc7e1798a74/crates/lint/src/sol/low/missing_events_access_control.rs#L666-L672).

When a local is reassigned, the new taint is `old_taint ∪ rhs_taint` instead of just `rhs_taint`.
Overwriting with a constant (`next = address(0xBEEF)`) does not clear the taint that `next`
inherited from an earlier `next = newOwner`, so a subsequent `owner = next` is spuriously reported
even though `owner` will always receive the constant.

mablr (commented):
New round 😄

---

### 1 — Unrelated event with a matching argument silently suppresses the warning

**Type:** False Negative (missed warning / false silence)

**Location:**  
[`missing_events_access_control.rs` — `mark_event`](https://github.com/foundry-rs/foundry/blob/aa0b2ebf1fa5b27aa982e5fd7b001f85a7cc750f/crates/lint/src/sol/low/missing_events_access_control.rs#L914-L925)

**Description:**  
`mark_event` matches an event to a pending write by checking whether the event's taint-sources and the write's taint-sources intersect — that is, whether they share a common input parameter or `msg.sender`. This means *any* event that happens to reference the same function argument as the write will be treated as covering the access-control state change, even if the event is semantically unrelated to it.

**Reproducer (would PASS but should WARN):**
```solidity
event Logged(address val);

function setOwnerWithUnrelatedArgEvent(address newOwner) external onlyOwner {
    owner = newOwner;        // access-control write
    emit Logged(newOwner);   // completely unrelated event, same arg
}
```

---

### 2 — `delete` on an access-control mapping is not tested (coverage gap)

**Type:** Coverage gap — no test exercising the `delete` code path

**Location:**  
[`missing_events_access_control.rs` — `ExprKind::Delete` branch](https://github.com/foundry-rs/foundry/blob/aa0b2ebf1fa5b27aa982e5fd7b001f85a7cc750f/crates/lint/src/sol/low/missing_events_access_control.rs#L757-L772)

**Description:**  
The `ExprKind::Delete` branch in `analyze_expr` is present but there is no test case in `MissingEventsAccessControl.sol` that covers `delete roles[account]` or `delete nestedRoles[role][account]`. Revoking a role via `delete` without an event should be reported, but it remains untested.

**Reproducer (would WARN, but not currently tested):**
```solidity
function revokeRole(address account) external onlyOwner {
    delete roles[account]; // should warn — no event
}
```

---

### 3 — Abstract/interface modifier names outside the heuristic list are not recognized

**Type:** False Negative (missed protection detection)

**Location:**  
[`missing_events_access_control.rs` — `name_looks_like_access_control`](https://github.com/foundry-rs/foundry/blob/aa0b2ebf1fa5b27aa982e5fd7b001f85a7cc750f/crates/lint/src/sol/low/missing_events_access_control.rs#L1597-L1608)

**Description:**  
`modifier_has_access_control` falls back to the name heuristic when a modifier has no body (abstract contracts, interfaces). The current list recognizes `onlyOwner*`, `onlyRole*`, `auth`, `restricted`, and a few `check*` variants. Common names like `onlyAdmin`, `onlyGuardian`, `onlyManager` are not in the list.

When an abstract base contract declares `modifier onlyAdmin();` (no body) and a concrete subclass both applies this modifier to a write function AND uses `admin` for access control, the lint will not recognize the modifier as protective — so it will not check whether the write has an event.

**Note:** This only affects modifiers without bodies. If the modifier has a body with `require(msg.sender == admin)`, `function_has_access_guard` picks it up correctly regardless of name.

**Reproducer (would miss protection, not warn when it should):**
```solidity
abstract contract Base {
    address public admin;
    modifier onlyAdmin() virtual; // no body
}

contract Derived is Base {
    function setAdmin(address newAdmin) external onlyAdmin {
        admin = newAdmin; // should warn — but lint doesn't see onlyAdmin as protective
    }

    function adminAction() external view {
        require(msg.sender == admin);
    }
}
```

crates/lint/src/sol/low/missing_events_access_control.rs
```diff
@@ -0,0 +1,1758 @@
...
+    fn mark_event(&mut self, expr: &hir::Expr<'_>) {
+        let Some(event_id) = emitted_event_id(expr) else { return };
+        let event_sources = self.value_sources(expr);
+        if event_sources.is_empty() {
+            return;
+        }
+
+        for write in &mut self.writes {
+            if !write.evented
+                && write.sources.intersects(&event_sources)
```
mablr: `write.sources` seems always empty for direct zero-literal writes (e.g. `owner = address(0)`), so intersects is always false and these writes can never be marked evented.

mablr (commented):
last smol one

figtracer:
@mablr addressed

mablr (dismissed):
It's fine now imo.

mattsse (dismissed):

PR #14964, opened 2026-05-29
fix(anvil): use fee manager as Tempo fork coinbase

### LINKED ISSUE #14866: anvil --network tempo --fork-url: silently drops every state-changing tx; eth_getTransactionReceipt returns null forever
## Summary

When running `anvil --fork-url <tempo-mainnet> --network tempo --fork-chain-id 4217`, every state-changing tx is silently discarded during block production. `eth_getTransactionReceipt` returns `null` indefinitely; the caller has zero signal until its own deadline fires. Two distinct root causes funnel into the same silent-drop sink at [`crates/anvil/src/eth/backend/executor.rs:478-481`](https://github.com/foundry-rs/foundry/blob/v1.7.1/crates/anvil/src/eth/backend/executor.rs#L478-L481).

### Root cause A — `block.beneficiary = address(0)` walks into the `DONOTUSE` sentinel

1. **Anvil leaves `block.beneficiary` at `address(0)` by default in fork mode.** No real Tempo block has beneficiary `0x0` — there's always a real validator address.
2. **Real Tempo's `TIP_FEE_MANAGER.validator_tokens[address(0)]` is a sentinel TIP-20.** Queried live: `0x20c00000000000000000000016c6514b53947fdc`, whose `symbol()` and `name()` both return `"DONOTUSE"`. Real Tempo never reads this slot — the entry exists specifically to trap anyone who lands at `beneficiary=0x0`.
3. **TIP-20 fee logic then tries to swap user_token (PathUSD by default) → validator_token (DONOTUSE) through the FeeAMM** ([`tempo_revm/src/tip_fee_manager/mod.rs:174`](https://github.com/tempoxyz/tempo/blob/1e5a82dff207e2cbd6cecdcfb8ff3cd2e39baa69/crates/precompiles/src/tip_fee_manager/mod.rs)). The FeeAMM pool for any PathUSD↔DONOTUSE pair is empty (no real-world participant provides liquidity for a "DONOTUSE" pair).

### Root cause B — Impersonated accounts with zero PathUSD

Independently of A: any impersonated EOA whose forked-state PathUSD balance is zero (typical for fresh wallets that have never transacted on real Tempo) fails the fee-collection step with `InsufficientBalance`, *even after A is fixed*. Tempo's `TipFeeManager` pre-charges gas in PathUSD via `collect_fee_pre_tx`, not in native ETH. `anvil_setBalance` seeds native balance, which Tempo's fee logic doesn't use.

### The common sink

Both A and B end at [`crates/anvil/src/eth/backend/executor.rs:478-481`](https://github.com/foundry-rs/foundry/blob/v1.7.1/crates/anvil/src/eth/backend/executor.rs#L478-L481), which routes non-validation EVM errors to a `tracing::trace!()` log and **silently discards the tx**. The result is identical from the caller's perspective: `null` receipts forever.

## Reproduce A

```bash
anvil --fork-url <tempo-mainnet-rpc> --fork-chain-id 4217 --network tempo --auto-impersonate &
# Use an account that DOES have PathUSD on real chain — a Tempo-active Safe:
cast rpc anvil_setBalance 0xde72cf02e058128a8c0072299ff628166a6064bd 0x100000000000000000
cast rpc anvil_impersonateAccount 0xde72cf02e058128a8c0072299ff628166a6064bd
HASH=$(cast send --unlocked --from 0xde72cf02e058128a8c0072299ff628166a6064bd \
       0x20c000000000000000000000b9537d11c60e8b50 \
       "approve(address,uint256)" 0x66793f54ae00349ebd7278447a419a58166a1d1e 12 | jq -r .transactionHash)
cast rpc evm_mine
cast receipt $HASH    # → null forever (root cause A)
```

## Reproduce B (after fixing A — set anvil coinbase to a real validator first)

```bash
# Continue the same session, then change sender to a fresh EOA with 0 PathUSD on real chain:
cast rpc anvil_impersonateAccount 0x3b16428B74e45f0B9A9019369A90Ff735e1fBC38
cast rpc anvil_setBalance       0x3b16428B74e45f0B9A9019369A90Ff735e1fBC38 0x100000000000000000
HASH=$(cast send --unlocked --from 0x3b16428B74e45f0B9A9019369A90Ff735e1fBC38 \
       0xde72cf02e058128a8c0072299ff628166a6064bd \
       "approveHash(bytes32)" 0x6f17a00d98ee8336830c5f21a7928a75188b57e1c1169fc6ea0fc809acddd19d \
       | jq -r .transactionHash)
cast rpc evm_mine
cast receipt $HASH    # → null forever (root cause B)
```

Confirm B by checking real PathUSD balance:
```bash
cast call 0x20c0000000000000000000000000000000000000 \
     "balanceOf(address)(uint256)" 0x3b16428B74e45f0B9A9019369A90Ff735e1fBC38 \
     --rpc-url <tempo-mainnet-rpc>
# → 0 (fresh EOA, never transacted on Tempo)
```

## Asks (in priority order)

1. **(Fixes A.)** In `--network tempo` fork mode, default `block.beneficiary` to a real Tempo validator address at startup — concretely: read `eth_getBlockByNumber("latest").miner` from the fork URL during `try_spawn` and use it as anvil's coinbase. The default `address(0)` walks straight into the DONOTUSE sentinel; **any real validator address** sidesteps it because (a) most real validators have an unset entry, which resolves to `DEFAULT_FEE_TOKEN = PathUSD`, and (b) the ~8% of validators with explicit non-PathUSD opt-ins have real FeeAMM pool liquidity forked correctly (`TipFeeManager` pool reserves live in standard EVM storage at the precompile address — verified at [`tempo_revm crates/precompiles/src/storage/evm.rs:157`](https://github.com/tempoxyz/tempo/blob/1e5a82dff207e2cbd6cecdcfb8ff3cd2e39baa69/crates/precompiles/src/storage/evm.rs#L157)).

2. **(Fixes B.)** When `anvil_impersonateAccount` is called on a Tempo fork, automatically seed the impersonated account with enough PathUSD to cover gas — but **only when the real-chain balance is zero**, otherwise preserve the real balance. Slot derivation: `keccak256(pad32(addr) || pad32(9))` against `PATH_USD_ADDRESS = 0x20c0…0000`; slot index 9 is Tempo's `balances` mapping (OpenZeppelin ERC20 uses slot 0; Tempo's TIP-20 differs — empirically verified). 1e18 wei is a reasonable floor.

3. **(Improves observability for both A and B and future similar cases.)** Surface a structured error to `eth_sendTransaction` / `eth_call` instead of trace-logging and silently dropping at [`executor.rs:478-481`](https://github.com/foundry-rs/foundry/blob/v1.7.1/crates/anvil/src/eth/backend/executor.rs#L478-L481). The silent drop is the worst part of the UX — debugging took us three days for A and another half-day to discover B. A propagated `InsufficientLiquidity` / `InsufficientBalance` would have been five-minute investigations. Related work in progress: #14679 ("stop masking gas estimation errors in send paths") could be a natural place to land this.

(Note: an earlier draft suggested "initialize FeeAMM pool storage in fork mode too" as a heavier alternative for A. Source-level review at `tempoxyz/tempo` revision `1e5a82d` confirms that's unnecessary — the inherited pool storage at `TipFeeManager` already contains all real-chain liquidity. The lightweight beneficiary fix alone is sufficient for A.)

## Related work in the same area

- #14679 (open) — `anvil: stop masking gas estimation errors in send paths` — natural place to land ask #3.
- #14535 (open) — `fix(anvil): check funds before transfer gas shortcut` — adjacent balance-check work; ask #2 could plug into the same pattern.
- #14159 (merged) — original fork-mode tempo init. Where the "warm precompile addresses with sentinel bytecode" code lives. Author intentionally scoped fork mode narrowly; this issue extends that scope.

BROCCOLO1D:
## Summary
- Defaults Anvil Tempo forks to the Tempo fee manager address as `block.coinbase`/beneficiary when no genesis coinbase is supplied.
- Applies the same default when `anvil_reset` starts forking from a non-forked Tempo node.
- Keeps explicit genesis coinbase configuration unchanged.

## Why
Tempo mainnet maps the zero-address validator token to a DONOTUSE sentinel. In fork mode, Anvil previously inherited the default zero beneficiary, so state-changing transactions could hit Tempo fee collection failures and then appear as missing receipts indefinitely. Tempo node-side simulation already uses the fee-manager sentinel to fall back to PathUSD.

## Changes
- Centralized the Tempo fork beneficiary default in `NodeConfig`.
- Applied the default during Tempo fork startup and during `anvil_reset` when it initializes a fork from a non-forked node.
- Added regression coverage for both startup forks and reset-to-fork flows.

## Validation
- `git diff --check`
- `cargo fmt --check`
- `cargo test -p anvil --test it tempo::test_tempo`

## Scope
Small Anvil Tempo fork/default-beneficiary behavior only; does not alter non-Tempo forks, standalone Tempo nodes, or explicit genesis coinbase settings.

Closes #14866


crates/anvil/tests/it/tempo.rs
```diff
@@ -52,6 +52,14 @@ async fn test_tempo_fork_detects_hardfork_from_fork_timestamp() {
 
     let node_info = api.anvil_node_info().await.unwrap();
     assert_eq!(node_info.hard_fork, "T3");
+
+    let latest_block = handle
```
figtracer, line 214: this test does not compile because the forked node handle is bound as `_handle` on line 50 but used as `handle` here.

crates/anvil/src/config.rs
```diff
@@ -1163,6 +1164,16 @@ impl NodeConfig {
             },
         );
 
+        if self.networks.is_tempo() && !self.fork_urls.is_empty() && self.genesis.is_none() {
```
figtracer: this default only applies to startup forks, so a tempo node that starts non-forked and later forks via `anvil_reset` can still keep the zero beneficiary and hit `donotuse`.

figtracer (commented):
two findings.

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -2140,6 +2140,7 @@ impl<N: Network> Backend<N> {
                     // `setup_fork_db_config`
                     node_config.base_fee.take();
                     node_config.fork_urls = vec![eth_rpc_url.clone()];
+                    node_config.apply_tempo_fork_beneficiary_default(&mut evm_env);
```
figtracer, line 2160: this overwrites a coinbase the user explicitly set with `anvil_setCoinbase` before resetting into fork mode. i reproduced by starting `NodeConfig::test_tempo()`, calling `anvil_set_coinbase(0x1111111111111111111111111111111111111111)`, then `anvil_reset(Some(Forking { json_rpc_url: ... }))`; the mined block used the fee-manager address instead of the custom coinbase. please only apply the default when the current beneficiary is still the default zero value, or otherwise track explicit runtime coinbase changes.

crates/anvil/src/config.rs
```diff
@@ -523,6 +524,20 @@ impl Default for NodeConfig {
 }
 
 impl NodeConfig {
+    /// Applies Tempo's safe default beneficiary for forked nodes when the user
+    /// has not supplied a genesis coinbase.
+    pub(crate) fn apply_tempo_fork_beneficiary_default<N>(&self, evm_env: &mut EvmEnv<N>) {
+        if self.networks.is_tempo() && !self.fork_urls.is_empty() && self.genesis.is_none() {
```
figtracer: this treats any genesis file as an explicit coinbase. `alloy_genesis::Genesis` defaults an omitted `coinbase` field to zero, so a tempo fork with a genesis used only for alloc/timestamp still keeps the zero beneficiary and can hit the same `donotuse` path. a local repro with `Genesis::default()` mined a block with zero beneficiary. please distinguish an actually supplied genesis coinbase, or apply the fee-manager default after genesis handling when the parsed coinbase is still zero.

figtracer (commented):
two follow-up comments from local repros.

BROCCOLO1D:
Addressed the Tempo fork beneficiary follow-ups: the fee-manager default now only applies while the beneficiary is still zero, so a runtime anvil_setCoinbase value survives anvil_reset into fork mode; after genesis handling, default Genesis::default() still gets the Tempo fee-manager beneficiary. Added targeted regressions for both cases and reran cargo fmt --check, git diff --check, and the three focused Tempo tests.

crates/anvil/src/config.rs
```diff
@@ -523,6 +524,23 @@ impl Default for NodeConfig {
 }
 
 impl NodeConfig {
+    /// Applies Tempo's safe default beneficiary for forked nodes while preserving
+    /// explicit coinbase selections.
+    pub(crate) fn apply_tempo_fork_beneficiary_default<N>(&self, evm_env: &mut EvmEnv<N>) {
```
stevencartavia, line 529: This default is applied on fork startup and in `reset_fork`, but **not** in `Backend::load_state`. `load_state` overwrites the block env wholesale and runs *after* `setup()` has already applied this default, with no equivalent normalization. So loading any state with a zero `block.beneficiary` (a non-Tempo or pre-fix dump) puts a forked Tempo node back at `0x0` and reintroduces the fee-collection failure this PR fixes — via both startup `--load-state` and runtime `anvil_loadState`.

Confirmed locally; a small fix in `load_state` resolves it — re-apply the default after assigning `state.block`:

```rust
if self.is_tempo() && self.is_fork() && env.block_env.beneficiary.is_zero() {
    env.block_env.beneficiary = TIP_FEE_MANAGER_ADDRESS;
}
```

After the fix: startup `--load-state` and runtime `anvil_loadState` both yield the fee-manager beneficiary, non-fork Tempo nodes are untouched, and all existing tempo tests still pass.

BROCCOLO1D:
Addressed the load-state follow-up in cb92dff3: `Backend::load_state` now normalizes zero beneficiaries for Tempo forks after restoring the saved block env, covering both startup `--load-state` and runtime `anvil_loadState`. Added focused regressions for both paths.

Validation:
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p anvil --test it loaded_zero_beneficiary_state -- --nocapture`
- `cargo test -p anvil --test it load_state_uses_fee_manager -- --nocapture`
- `cargo test -p anvil --test it tempo::test_tempo -- --nocapture`

stevencartavia (dismissed):

mablr (commented):
I just fixed a weak write lock pattern.

Sgtm, thanks.

mablr (approved):

figtracer (approved):
lgtm

PR #14965, opened 2026-05-29
fix(tui): always restore terminal state during panic unwinding

bravesasha:
TerminalGuard::restore previously relied on the panic hook installed in setup() to restore terminal state during panic unwinding. Since panic hooks are global mutable state and may be replaced by external code (color-eyre, human-panic, test harnesses, etc.), this could leave the terminal stuck in raw mode and alternate screen after a panic.

Always run half_restore during Drop so terminal cleanup does not depend on the panic hook remaining installed. The hook remains as defense-in-depth, and the double cleanup is safe because the operations are idempotent.


crates/tui/src/lib.rs
```diff
@@ -178,6 +178,14 @@ impl<B: Backend + Write> TerminalGuard<B> {
     }
 
     fn restore(&mut self) {
+        // Always restore terminal state, even during panic unwinding.
+        // `Drop` must not rely solely on the panic hook installed in `setup()`,
+        // because external code (e.g. color-eyre, human-panic, test harnesses)
+        // may replace it after `setup()` and before a panic occurs.
+        // `disable_raw_mode` and `LeaveAlternateScreen` are idempotent,
+        // so a double call (hook + Drop) is safe.
+        Self::half_restore(self.terminal.backend_mut());
```
mablr: Comment is too verbose.

Change seems ok to me. @figtracer could you confirm it's necessary ? (seems very narrow corner case)

bravesasha: I agree

// Restore terminal state even during panic unwinding.
Self::half_restore(self.terminal.backend_mut());

The intent is simply to make terminal cleanup independent of the panic hook state

figtracer (dismissed):
i don't see a realistic path where this happens but you never know

bravesasha:
> i don't see a realistic path where this happens but you never know

Happy to shorten the comment if needed.

figtracer:
> > i don't see a realistic path where this happens but you never know
> 
> Happy to shorten the comment if needed.

yeah pls, then its gtg

mablr (approved):

figtracer (approved):

PR #14966, opened 2026-05-29
feat(tempo): add typed session authorization policy builder

0xKarl98:
Step closer to OSS-162

Add a session_policy module that prepares Tempo KeyAuthorization values from typed session policy inputs, validates signed authorizations against the stored session identity and policy, and activates session entries with generated key material. 

The validation path now compares expiry, spend limits, and call scopes in canonical form so equivalent policy orderings are accepted while unrestricted and empty policies remain distinct.

### DIFF
--- a/crates/common/src/tempo/session_policy.rs
+++ b/crates/common/src/tempo/session_policy.rs
@@ -405,8 +405,7 @@ mod tests {
         };
         let prepared = request.prepare(1_700_000_000).unwrap();
         let signature = root.sign_hash_sync(&prepared.authorization.signature_hash()).unwrap();
-        let signed =
-            prepared.authorization.clone().into_signed(PrimitiveSignature::Secp256k1(signature));
+        let signed = prepared.authorization.into_signed(PrimitiveSignature::Secp256k1(signature));
 
         let mut other_request = request;
         other_request.session_id = B256::from([0x71; 32]);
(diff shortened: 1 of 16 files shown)


crates/common/src/tempo/session_policy.rs
```diff
@@ -0,0 +1,443 @@
...
+        ensure!(!self.scope.is_empty(), "session authorization requires a call scope");
+
+        let authorization = KeyAuthorization::unrestricted(
+            self.chain_id,
+            SignatureType::Secp256k1,
+            self.key_address,
+        )
+        .with_expiry(expiry)
+        .with_limits(session_spend_limits_to_authorization(&self.spend_limits))
+        .with_allowed_calls(self.scope.clone());
```
mablr: Should this bind the authorization to `session_id` via the `witness` field? Right now the signed payload only covers chain/key/expiry/limits/scope, and validation also ignores `auth.witness`, so the same signed authorization can be reused for another local session with the same policy.

0xKarl98: Have  bind the auth to session_id via witness

figtracer (dismissed):
lgtm

mablr (dismissed):
sgtm

mattsse (approved):
lgtm

grandizzy (approved):

PR #14970, opened 2026-05-29
feat(lint): add `reentrancy-eth` lint

figtracer:
## Summary

- rename the existing `reentrancy-unlimited-gas` lint surface to `reentrancy-eth`
- keep the detector focused on uncapped ETH-transferring low-level calls where state read before the call is written after it
- ignore zero-value calls backed by literal zero or constant-zero values
- update lint docs, README entry, and UI fixtures for the new lint ID/help URL

## Testing

- `cargo fmt --check`
- `cargo test -p forge --test ui ReentrancyEth`
- `cargo test -p forge-lint sol::tests::registered_lints_have_docs`
- `git diff --check`


crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -466,6 +480,14 @@ fn is_uncapped_value_call(callee: &hir::Expr<'_>, opts: Option<&[hir::NamedArg<'
     value.is_some_and(|value| !is_zero_literal(value)) && gas.is_none_or(gas_option_forwards_all)
 }
 
+fn is_non_reentrant_modifier(hir: &hir::Hir<'_>, modifier: &hir::Modifier<'_>) -> bool {
+    matches!(
+        modifier.id,
+        ItemId::Function(func_id)
+            if hir.function(func_id).name.is_some_and(|name| name.as_str() == "nonReentrant")
+    )
```
mablr: This suppresses the lint for any modifier whose name is `nonReentrant`, regardless of what the modifier actually does. I added a fixture with `modifier nonReentrant() { _; }` around a classic read-call-write withdraw, and `cargo test --test ui` reported the expected `~WARN` was missing. Unless this is meant to trust names only, this should either inspect the guard pattern or avoid suppressing diagnostics solely by modifier name.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -466,6 +480,14 @@ fn is_uncapped_value_call(callee: &hir::Expr<'_>, opts: Option<&[hir::NamedArg<'
     value.is_some_and(|value| !is_zero_literal(value)) && gas.is_none_or(gas_option_forwards_all)
 }
```
mablr, line 471: This only treats literal `0` as a zero-value call. A constant-zero value such as `uint256 constant ZERO = 0; receiver.call{value: ZERO}("")` still gets queued and later reported as reentrant ETH transfer even though it cannot transfer ETH. I verified this with a temporary UI fixture; the diagnostic on `value: ZERO` was an unmatched warning.

mablr (commented):
Some clanker findings, reproduced locally

crates/lint/src/sol/high/reentrancy.rs
figtracer: Fixed in b6a292f.

crates/lint/src/sol/high/reentrancy.rs
figtracer: Fixed in b6a292f.

crates/lint/testdata/ReentrancyEth.stderr
```diff
@@ -0,0 +1,88 @@
...
+   ╰ help: https://getfoundry.sh/forge/linting/reentrancy-eth
+
+warning[reentrancy-eth]: uncapped ETH transfer can be reentered before `balances` is updated
+   ╭▸ ROOT/testdata/ReentrancyEth.sol:LL:CC
+   │
+LL │ …     (bool ok,) = payable(msg.sender).call{value: amount}("");
+   │                    ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+   │
+   ╰ help: https://getfoundry.sh/forge/linting/reentrancy-eth
+
```
mattsse: `git diff --check origin/master...origin/pr/14970` now fails here with `new blank line at EOF`. Suggested fix: remove this final blank line from the generated stderr fixture so the advertised `git diff --check` validation passes again.

mattsse (commented):
One small cleanup from re-checking the latest head.

mattsse (dismissed):

mattsse (approved):

PR #14971, opened 2026-05-29
fix(mutation): key cached results by execution inputs

mablr:
## Summary
Fix mutation result caching so cached outcomes are invalidated when execution inputs change, including test artifacts, config, and EVM options.

- Keep mutant-set caching scoped to inputs that affect generated mutants.
- Added an execution cache key for result and survived-span caches.
- Added unit coverage plus a CLI regression for test changes invalidating cached mutation results.

towards OSS-1


crates/forge/src/mutation/orchestrator.rs
```diff
@@ -80,6 +98,7 @@ pub async fn run_mutation_testing(
 
     // Determine which paths to mutate
     let mutate_paths = resolve_mutate_paths(&config, output, &mutation_config)?;
+    let execution_cache_key = mutation_execution_cache_key(&config, output, &evm_opts)?;
```
figtracer: The execution cache key should include the effective worker count. Adaptive span skipping is concurrency-sensitive, so `--mutation-jobs 1` can produce a different result vector than `--mutation-jobs 4`, but both currently reuse the same results/survived cache.

crates/forge/src/mutation/orchestrator.rs
```diff
@@ -237,6 +258,51 @@ pub async fn run_mutation_testing(
...
+/// and future config fields unless explicitly skipped by `Config` itself. The
+/// artifact fingerprint covers source and test build IDs, so changing a test
+/// file invalidates result caches even when the mutated source did not change.
+fn mutation_execution_cache_key(
+    config: &Config,
+    output: &ProjectCompileOutput<MultiCompiler>,
+    evm_opts: &EvmOpts,
+) -> Result<String> {
+    let artifacts = output
+        .artifact_ids()
```
figtracer, line 285: This fingerprints the outer compile output, which can be filtered, while each mutant is recompiled and tested with `EmptyTestFilter`. A test excluded from the outer output can still affect mutation execution, so changes to it may not invalidate cached results.

figtracer (commented):
Two cache correctness concerns from review.

crates/forge/src/mutation/orchestrator.rs
mablr: Addressed in 95ed58b99.

crates/forge/src/mutation/orchestrator.rs
mablr: Addressed in https://github.com/foundry-rs/foundry/commit/95ed58b99f3e0b85a9d484a8e6b5f2bcf684b646.

figtracer (approved):

crates/forge/src/mutation/mod.rs
```diff
@@ -496,3 +509,70 @@ impl MutationHandler {
...
+    }
+
+    #[test]
+    fn mutant_cache_path_ignores_execution_only_timeout() {
+        let (_temp, mut first_config) = test_config();
+        let root = first_config.root.clone();
+        let mut second_config = first_config.clone();
+
+        first_config.mutation.timeout = Some(1);
+        second_config.root = root;
```
stevencartavia: is this a no-op since `second_config` already cloned that root from `first_config`?

mablr: yes, it's a remnant, fixed 5e8e7cb2ddfd8cc6c2e0ab86a0d2107be5619141

stevencartavia (approved):

PR #14973, opened 2026-05-29
feat(forge): migrate forge config, geiger, lint to output channel contract

stevencartavia:
Bundled compliance pass across `forge config`, `forge geiger`, `forge lint`. Per docs/dev/output-channels.md.

- `forge lint` — "nothing to lint" moved to stderr via `sh_status!`.
- `forge config` — already compliant (TOML/JSON stay on stdout, no prose). Doc row flipped only.
- `forge geiger` — already compliant (delegates to lint, uses warnings). Doc row flipped only.

Part of OSS-224

### DIFF
--- a/crates/forge/src/cmd/doc/server.rs
+++ b/crates/forge/src/cmd/doc/server.rs
@@ -74,7 +74,7 @@ impl Server {
             .unwrap_or_else(|| "404.html".to_string());
 
         let serving_url = format!("http://{address}");
-        sh_println!("Serving on: {serving_url}")?;
+        sh_status!("Serving on: {serving_url}")?;
 
         let thread_handle = std::thread::spawn(move || serve(build_dir, sockaddr, &file_404));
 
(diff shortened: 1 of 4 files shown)


crates/forge/src/cmd/doc/server.rs
```diff
@@ -74,7 +74,7 @@ impl Server {
             .unwrap_or_else(|| "404.html".to_string());
 
         let serving_url = format!("http://{address}");
-        sh_println!("Serving on: {serving_url}")?;
+        sh_status!("Serving on: {serving_url}")?;
```
mablr: We don't need to update this as https://github.com/foundry-rs/foundry/pull/14568 will simply delete this file.

mablr (changes requested):

crates/forge/src/cmd/doc/server.rs
stevencartavia: reverted `forge doc`

mablr (approved):

grandizzy (approved):

PR #14978, opened 2026-05-30
feat(cast): add Tempo wallet session create and revoke commands

0xKarl98:
Adds the focused create/revoke-local layer needed to mint a temporary session key, bind it to a constrained authorization policy, validate the signed authorization, and persist it for later `--tempo.session` use.

- Add `cast wallet session create`
    - generates a temporary secp256k1 session key
    - parses expiry, call scope, and spend limits
    - builds a `SessionAuthorizationRequest`
    - signs the authorization with the root wallet
    - validates the signed authorization against the session policy
    - stores the active session entry in the local session registry

- Add `cast wallet session revoke`
    - Just removes a local session entry by session id


mattsse (approved):
LGTM

PR #14979, opened 2026-05-30
fix(cast): parse tempo scope signatures

mattsse:
## Summary
- parse Tempo policy function signatures through the shared ABI parser used by `cast sig`
- split scope selector lists only at commas outside function parameter lists
- preserve multi-recipient selector rules and reject empty selector or recipient entries

## Root Cause
Tempo policy scopes split selector specs with a raw comma split, so full function signatures like `transfer(address,uint256)` were split in the middle of their parameter list. The selector parser also hashed any string containing parentheses directly, which let malformed signatures through instead of using the normal ABI parser.

## Impact
`cast keychain` and `cast wallet session create` can now accept full function signatures in `--scope` entries, including signatures with comma-separated parameters, while keeping the existing shorthand and 4-byte selector forms.


grandizzy (approved):
makes sense

Issue #9880, opened 2025-02-13
bug(`cast`): not all selectors are decoded when running `--decode-internal`


zerosnacks:
### Component

Cast

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

forge Version: 1.0.0-nightly Commit SHA: 1bcd17c8299d7c0792a77b2d4a0cf2c531a5cab2 Build Timestamp: 2025-02-12T00:21:20.710803208Z (1739319680) Build Profile: maxperf

### What version of Foundryup are you on?

foundryup: 1.0.1

### Operating System

None

### Describe the bug

When running `cast run `0x43dbea777eebc21aee1449b0a8a71b554b36ffa1bda7ee9c64a239687454b72c --rpc-url <RPC_URL> --decode-internal -vvvvv --etherscan-api-key <ETHERSCAN_API_KEY>` it decodes the selectors partially and returns early.

When re-running the same command it will continue fetching the source maps and load the others from cache.

It takes several runs to get a clean output.

I'm presuming this is related to a race condition or we return early upon getting throttled by Etherscan.

mablr:
Closed by #14981

PR #14981, opened 2026-05-30
fix(traces): retry Cloudflare blocks instead of aborting identification

0xMars42:
## Motivation

Addresses #9880. Under load, `cast run --decode-internal` (and other tracing that
resolves verified sources) identifies contracts only partially and needs several
runs to produce a complete trace, filling in from the on-disk cache between runs.

## Solution

The trace `ExternalIdentifier` fetches contract metadata from Etherscan / Sourcify
through a rate-limit-aware stream (`ExternalFetcher`); `cast run --decode-internal`
then compiles those sources into source maps (`get_compiled_contracts`), so a
contract whose metadata never arrives is never decoded.

`RateLimitExceeded` is already handled with backoff and re-queue, but a
`BlockedByCloudflare` was routed into the `InvalidApiKey` handling (it set the
fetcher's invalid-key flag and returned `Poll::Ready(None)`, carrying the same
`// mark key as invalid` comment). That classifies a transient block as a
permanent auth failure: it ends the stream and abandons every address still queued
in that pass, and disables the fetcher for the rest of the process. Etherscan sits
behind Cloudflare, which returns this when a request burst trips its protection, so
a single transient block stops identification for every remaining contract. The
on-disk cache is why re-running fills the gaps progressively.

This treats a Cloudflare block as the transient rate limiting it is: back off and
retry the address, bounded by `MAX_CLOUDFLARE_RETRIES` (5) so a persistent block
can't hang the command in an infinite loop. After the bound is reached the address
is given up individually (yielded as `None`) instead of aborting the whole stream,
so one blocked contract no longer prevents the rest from being identified.
`InvalidApiKey` is left unchanged and still aborts, since retrying a bad key is
pointless.

## Additional context

Adds a regression test (`cloudflare_block_retries_instead_of_abandoning_queue`)
with a fetcher that returns a transient Cloudflare block on first contact with each
address and then succeeds. Before the change the stream aborts on the first block
and yields nothing; after it, every address is yielded in a single run. The test is
deterministic and does not hit the network.

### DIFF
--- a/crates/evm/traces/src/identifier/external.rs
+++ b/crates/evm/traces/src/identifier/external.rs
@@ -227,6 +227,10 @@ impl TraceIdentifier for ExternalIdentifier {
 type FetchFuture =
     Pin<Box<dyn Future<Output = (Address, Result<Option<Metadata>, EtherscanError>)>>>;
 
+/// Maximum number of times a single address is retried through a transient Cloudflare
+/// block before we give up on it. Bounded so a persistent block can't loop forever.
+const MAX_CLOUDFLARE_RETRIES: u32 = 5;
+
 /// A rate limit aware fetcher.
 ///
 /// Fetches information about multiple addresses concurrently, while respecting rate limits.
@@ -243,6 +247,8 @@ struct ExternalFetcher {
     queue: Vec<Address>,
     /// The in progress requests
     in_progress: FuturesUnordered<FetchFuture>,
+    /// Per-address retry counter for transient Cloudflare blocks.
+    attempts: HashMap<Address, u32>,
 }
 
 impl ExternalFetcher {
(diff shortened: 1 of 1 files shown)


mattsse (approved):
I think this makes sense, wdyt @mablr

crates/evm/traces/src/identifier/external.rs
```diff
@@ -227,6 +227,10 @@ impl TraceIdentifier for ExternalIdentifier {
 type FetchFuture =
     Pin<Box<dyn Future<Output = (Address, Result<Option<Metadata>, EtherscanError>)>>>;
 
+/// Maximum number of times a single address is retried through a transient Cloudflare
+/// block before we give up on it. Bounded so a persistent block can't loop forever.
+const MAX_CLOUDFLARE_RETRIES: u32 = 5;
```
mablr, line 232: Looks a bit like "magic" number, but I don't see a real need for configurability here.
G2G as-is.

mablr (approved):
Makes sense, thank you.

PR #14988, opened 2026-05-31
fix(config): accept legacy invariant assert_all key

### LINKED ISSUE #14987: `Warning: Found unknown `assert_all` config key in section `invariant` defined in foundry.toml.`
PR https://github.com/foundry-rs/foundry/pull/12587 introduced `assert_all` as config key for invariant tests. In snekmate I have the [following](https://github.com/pcaversaccio/snekmate/blob/main/foundry.toml#L25):

```toml
invariant = { runs = 256, depth = 15, assert_all = false }
```

However, any `forge` command triggers a warning now:
```sh
Warning: Found unknown `assert_all` config key in section `invariant` defined in foundry.toml.
```
Image

I use latest nightly for testing:
Image

BROCCOLO1D:
## Summary
- Allows the legacy `invariant.assert_all` config key in warning validation for both standalone and profile-scoped invariant sections.
- Adds a regression test covering existing configs that still set `assert_all`.

## Why
`invariant.assert_all` was introduced as a public config key in #12587, then later removed from the serialized invariant config when invariant campaigns became contract-level. Existing projects now see an unknown-key warning on every forge command even though the config was previously valid.

## Changes
- Adds section-local backward-compatible config keys to the warning provider.
- Registers `assert_all` as a backward-compatible `invariant` key.
- Tests that `[invariant]` and `[profile.default.invariant]` usages do not emit unknown-key warnings.

## Validation
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p foundry-config no_unknown_key_warning_for_legacy_invariant_assert_all`
- `cargo test -p foundry-config unknown_key_warning`
- `cargo check -p foundry-config`

## Scope
- Warning validation only; this does not reintroduce a behavior toggle for invariant campaigns.

Closes #14987

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -6865,6 +6865,36 @@ mod tests {
     }
 
     #[test]
+    fn no_unknown_key_warning_for_legacy_invariant_assert_all() {
+        // Regression test: `invariant.assert_all` was a public config key before invariant
+        // campaigns were grouped by contract. Existing configs should continue to load quietly.
+        figment::Jail::expect_with(|jail| {
+            jail.create_file(
+                "foundry.toml",
+                r#"
+                [invariant]
+                assert_all = false
+
+                [profile.default.invariant]
+                assert_all = true
+                "#,
+            )?;
+
+            let cfg = Config::load().unwrap();
+            assert!(
+                !cfg.warnings.iter().any(|w| matches!(
+                    w,
+                    crate::Warning::UnknownSectionKey { key, section, .. }
+                        if key == "assert_all" && section == "invariant"
+                )),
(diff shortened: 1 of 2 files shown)


mablr (commented):
Thanks for your help @BROCCOLO1D.

But I'm unsure we want `assert_all` to be accepted as "legacy" since it lived about ten days? Moreover this doesn't address the gas snapshot problem later https://github.com/foundry-rs/foundry/issues/14987#issuecomment-4586858911

BROCCOLO1D:
Closing this based on maintainer feedback: accepting assert_all as a legacy key does not look like the right direction, and the gas snapshot regression is being handled separately in #14990. Happy to revisit if a compatibility shim is still desired.

PR #14989, opened 2026-05-31
feat(evm): shard invariant campaign runs across workers

0xKarl98:
Shard invariant campaign runs into contiguous worker plans and execute eligible campaigns in parallel. Timeout-driven campaigns and `call_override` campaigns remain single-worker to preserve existing execution semantics.

Each worker owns its local EVM state, runner, fuzz state, and corpus manager. Worker outputs are merged back into one logical invariant result with deterministic policies for:

- worker run range validation
- predicate failures
- handler assertion failures
- metrics, reverts, traces, and coverage
- optimization best value and sequence
- master-worker-only failed corpus replay counts

Corpus entries selected by workers are returned to the campaign aggregator and persisted once after worker outputs are merged, along with the merged optimization best state. 

Handler assertion shrinking also runs after merge so the final result is selected at the campaign boundary.

### DIFF
--- a/crates/evm/evm/src/executors/invariant/campaign.rs
+++ b/crates/evm/evm/src/executors/invariant/campaign.rs
@@ -106,7 +106,7 @@ pub struct InvariantWorkerOutput {
 
 impl InvariantWorkerOutput {
     #[cfg(test)]
-    pub fn new(plan: InvariantWorkerPlan, result: InvariantFuzzTestResult) -> Self {
+    pub const fn new(plan: InvariantWorkerPlan, result: InvariantFuzzTestResult) -> Self {
         Self { plan, result, corpus_entries: Vec::new() }
     }
 }
--- a/crates/evm/evm/src/executors/invariant/campaign.rs
+++ b/crates/evm/evm/src/executors/invariant/campaign.rs
@@ -1,5 +1,5 @@
 use super::{InvariantFuzzError, InvariantFuzzTestResult, InvariantMetrics};
-use crate::executors::{EarlyExit, FuzzTestTimer, corpus::CampaignCorpusEntry};
+use crate::executors::{EarlyExit, corpus::CampaignCorpusEntry};
 use alloy_primitives::{Address, I256, Selector};
 use eyre::{Result, ensure};
 use foundry_evm_coverage::HitMaps;
@@ -68,28 +68,19 @@ pub struct InvariantWorkerPlan {
 pub struct InvariantCampaignState {
     total_runs: AtomicU32,
     global_early_exit: EarlyExit,
-    timer: FuzzTestTimer,
 }
 
 impl InvariantCampaignState {
-    pub fn new(timeout: Option<u32>, early_exit: EarlyExit) -> Self {
-        Self {
-            total_runs: AtomicU32::new(0),
-            global_early_exit: early_exit,
-            timer: FuzzTestTimer::new(timeout),
-        }
+    pub fn new(early_exit: EarlyExit) -> Self {
+        Self { total_runs: AtomicU32::new(0), global_early_exit: early_exit }
     }
 
     pub fn increment_runs(&self) -> u32 {
         self.total_runs.fetch_add(1, Ordering::Relaxed) + 1
     }
 
-    pub fn should_continue(&self) -> bool {
-        !(self.global_early_exit.should_stop() || self.timer.is_timed_out())
-    }
-
-    pub const fn is_timed_campaign(&self) -> bool {
-        self.timer.is_enabled()
+    pub fn should_stop(&self) -> bool {
+        self.global_early_exit.should_stop()
     }
 }
 
@@ -485,12 +476,6 @@ mod tests {
 
         let err = InvariantCampaignSpec::new(0).worker_plans(0).unwrap_err();
         assert!(err.to_string().contains("requires at least one worker"));
-    }
-
-    #[test]
-    fn campaign_state_reports_timed_campaigns() {
-        assert!(!InvariantCampaignState::new(None, EarlyExit::new(false)).is_timed_campaign());
-        assert!(InvariantCampaignState::new(Some(1), EarlyExit::new(false)).is_timed_campaign());
     }
 
     #[test]
--- a/crates/evm/evm/src/executors/invariant/campaign.rs
+++ b/crates/evm/evm/src/executors/invariant/campaign.rs
@@ -6,7 +6,7 @@ use foundry_evm_coverage::HitMaps;
 use foundry_evm_fuzz::BasicTxDetails;
 use std::{
     collections::{HashMap, HashSet},
-    sync::atomic::{AtomicU32, Ordering},
+    sync::atomic::{AtomicBool, AtomicU32, Ordering},
 };
 
 /// Immutable plan-level description for an invariant campaign.
@@ -67,12 +67,17 @@ pub struct InvariantWorkerPlan {
 /// Shared state used only to coordinate invariant worker execution.
 pub struct InvariantCampaignState {
     total_runs: AtomicU32,
+    terminal_stop: AtomicBool,
     global_early_exit: EarlyExit,
 }
 
 impl InvariantCampaignState {
     pub const fn new(early_exit: EarlyExit) -> Self {
-        Self { total_runs: AtomicU32::new(0), global_early_exit: early_exit }
+        Self {
+            total_runs: AtomicU32::new(0),
+            terminal_stop: AtomicBool::new(false),
+            global_early_exit: early_exit,
+        }
     }
 
     pub fn increment_runs(&self) -> u32 {
@@ -80,7 +85,15 @@ impl InvariantCampaignState {
     }
 
     pub fn should_stop(&self) -> bool {
-        self.global_early_exit.should_stop()
+        self.global_early_exit.should_stop() || self.terminal_stop.load(Ordering::Relaxed)
+    }
+
+    pub fn request_terminal_stop(&self) {
+        self.terminal_stop.store(true, Ordering::Relaxed);
+    }
+
+    pub const fn early_exit(&self) -> &EarlyExit {
+        &self.global_early_exit
     }
 }
 
@@ -476,6 +489,16 @@ mod tests {
 
         let err = InvariantCampaignSpec::new(0).worker_plans(0).unwrap_err();
         assert!(err.to_string().contains("requires at least one worker"));
+    }
+
+    #[test]
+    fn campaign_state_stops_after_terminal_request() {
+        let state = InvariantCampaignState::new(EarlyExit::new(false));
+        assert!(!state.should_stop());
+
+        state.request_terminal_stop();
+
+        assert!(state.should_stop());
     }
 
     #[test]
--- a/crates/evm/evm/src/executors/invariant/campaign.rs
+++ b/crates/evm/evm/src/executors/invariant/campaign.rs
@@ -70,6 +70,12 @@ pub struct InvariantWorkerOutput {
     pub result: InvariantFuzzTestResult,
 }
 
+impl InvariantWorkerOutput {
+    pub const fn new(plan: InvariantWorkerPlan, result: InvariantFuzzTestResult) -> Self {
+        Self { plan, result }
+    }
+}
+
 /// Merges worker outputs back into one logical invariant campaign result.
 ///
 /// Merge policy:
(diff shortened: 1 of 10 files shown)


crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -946,43 +1057,104 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
...
         );
-        let campaign_spec = InvariantCampaignSpec::new(planned_runs);
-        let worker_plan = campaign_spec.worker_plans(1)?.pop().expect("one worker plan requested");
-        let worker_output = InvariantWorkerOutput::new(worker_plan, worker_result);
-        let mut aggregator = InvariantCampaignAggregator::new(campaign_spec);
-        aggregator.push(worker_output);
-        aggregator.finish()
+        plan.runs = planned_runs;
+        Ok(InvariantWorkerOutput { plan, result: worker_result, corpus_entries })
+    }
```
mablr: May this turn a real invariant failure into an aggregation error when workers run in parallel ? Rewriting `plan.runs` to the actual `planned_runs` changes only the worker’s length, but later workers keep their original `first_global_run`; if worker 0 stops early after finding a failure, worker 1 still starts at the old offset and `finish_with_corpus_entries` rejects the gap as “worker outputs do not cover the logical campaign”.

0xKarl98: In the current code this does not shrink on early failure because ,  planned_runs is initialized to the assigned plan.runs and only grows, so parallel non-timeout workers keep their original range .

The naming makes this easy to misread, so I’ll make the distinction explicit and only rewrite the plan for timed campaigns .

mablr: Nice, thanks for clarification.

0xKarl98: Have clarified a bit , ptal again

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -488,39 +519,150 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
...
+            self.prepare_campaign_seed(&invariant_contract, initial_handler_failures)?;
+        let run_in_parallel = worker_plans.len() > 1;
+        let mut runner = self.runner.clone();
+        let config = self.config.clone();
+        let setup_contracts = self.setup_contracts;
+        let project_contracts = self.project_contracts;
+        let base_executor = self.executor.clone();
+        let campaign_state =
+            Arc::new(InvariantCampaignState::new(self.config.timeout, early_exit.clone()));
+
```
mablr, line 571: The timeout timer now starts before worker setup? If setup or corpus replay is slow, the invariant run can time out before doing real fuzz runs. Before this PR, the timer started after setup.

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -1058,30 +1243,38 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
...
         let mut invariant_test =
-            InvariantTest::new(fuzz_state, targeted_contracts, failures, self.runner.clone());
+            InvariantTest::new(fuzz_state, targeted_contracts, failures, runner.clone());
 
         // Seed invariant test with previously persisted optimization state,
-        // but only if the current invariant is in optimization mode.
-        if invariant_contract.is_optimization() {
+        // but only if the current invariant is in optimization mode. Persisted optimization state
+        // is a master-worker artifact loaded with the initial corpus.
+        if is_master_worker && invariant_contract.is_optimization() {
```
mablr: Only worker 0 seems to load the previously saved optimization best value. Other workers start from no best value, so they may treat worse optimization results as "new best" locally and save those sequences into the corpus, even though the full campaign did not improve.

0xKarl98: Have fixed !

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -155,6 +164,19 @@ impl InvariantThroughputMetrics {
     }
 }
 
+fn invariant_worker_count(runs: u32) -> usize {
+    let max_workers = (runs / MIN_RUNS_PER_INVARIANT_WORKER).max(1);
```
grandizzy: **🛑 `FOUNDRY_FUZZ_SEED` reproducibility broken across machines.**

Worker count keys off `rayon::current_num_threads()`, so the same seed produces different partitioning, different per-worker RNG streams (via `invariant_worker_runner` → `runner.new_rng()`), different corpus evolution, and potentially different failures on a 4-core laptop vs. a 32-core CI runner. Users will file repros from CI that don't reproduce locally.

Please add an explicit `invariant.workers` config knob (mirror `fuzz.runs` plumbing), surface it in the campaign banner alongside the seed, and **default to `1`** so existing seeded workflows keep working. Parallelism becomes opt-in until the determinism contract is documented.

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -1058,30 +1246,38 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
...
+            artifact_filters: &campaign_seed.artifact_filters,
+        };
         let worker = WorkerCorpus::new(
-            0,
-            self.config.corpus.clone(),
+            plan.worker_id as usize,
+            config.corpus.clone(),
             strategy.boxed(),
-            Some(&self.executor),
+            is_master_worker.then_some(&*executor),
```
grandizzy: **🛑 (partially fixed) Non-master workers still cold-start their coverage maps.**

Last push lifted the `id == 0` gate on the optimization baseline in `corpus.rs:397` — good. But the broader concern stands: workers 1..N still get `None` for `executor` / `targeted_contracts` / `DynamicTargetCtx`, so they never replay the persisted corpus and start with empty `in_memory_corpus`, `history_map`, `sancov_history_map`, and `edge_indices`. They then treat trivial edges as "new coverage" relative to their local empty maps and return those entries via `corpus_entries`, which `persist_campaign_outputs` writes into `worker0/corpus`. The shared on-disk corpus gets polluted and each worker burns CPU re-discovering edges worker 0 already knows.

Two options:
1. **Simple unblock**: add `&& !self.config.corpus.collect_edge_coverage()` to `can_parallelize` for now.
2. **Correct**: have worker 0 replay corpus + warm its history maps first, build a clonable `WorkerCorpusSeed { in_memory_corpus, history_map, edge_indices, sancov_history_map }`, and clone it into every worker; dedupe returned entries against a campaign-global coverage map before persisting.

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -946,43 +1063,114 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
...
+            // Sharded campaigns must report the original assigned range. Early worker exit changes
+            // the number of executed runs, but it must not shrink `plan.runs`: following workers'
+            // `first_global_run` offsets were computed from the original partition.
+            debug_assert_eq!(reported_runs, plan.runs);
+            plan
+        };
+        Ok(InvariantWorkerOutput { plan: reported_plan, result: worker_result, corpus_entries })
+    }
+
+    fn shrink_handler_failures(
```
grandizzy, line 1135: **🛑 Handler shrink uses an unprepared executor; regresses `workers == 1` too.**

Pre-PR, `prepare_test` mutated `self.executor`: installed the `Fuzzer` inspector, set `mapping_slots`, ran `invariant_preflight_check`. Post-PR, `prepare_worker` mutates only the worker-local executor clones, so the base `self.executor` reaches `shrink_handler_sequence` here without those inspectors/config. This affects every call site, not just parallel mode.

Simplest fix: move handler-failure shrinking back inside `run_invariant_worker` against the worker's prepared executor (accept small dedupe loss, or dedupe up-front before shrinking). If you want to keep aggregator-side shrinking for dedupe, prepare a dedicated shrink executor there.

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -488,39 +520,152 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
...
+                acc.checked_add(runs).ok_or_else(|| eyre!("invariant worker run range overflows"))
+            })?;
+        let mut aggregator =
+            InvariantCampaignAggregator::new(InvariantCampaignSpec::new(total_reported_runs));
+        for worker_output in worker_outputs {
+            aggregator.push(worker_output);
+        }
+        let (mut result, corpus_entries) = aggregator.finish_with_corpus_entries()?;
+        self.shrink_handler_failures(&mut result, progress, early_exit);
+        WorkerCorpus::persist_campaign_outputs(
```
grandizzy: **⚠️ Live persistence lost on Ctrl-C / worker `Err` — regresses single-worker.**

Pre-PR, each interesting corpus input and each optimization improvement flushed to disk immediately. Post-PR the invariant path always uses `process_inputs_for_campaign` (`corpus.rs:544`), buffers in memory, and writes only after merge **and** shrink. A killed process or any worker `Err` (the `try_fold` above propagates) loses every entry the campaign found.

Please preserve live persistence at least for `worker_count == 1` — `process_inputs_inner`'s `persist_now=true` branch is still there, just rewire the single-worker path to use it.

crates/evm/evm/src/executors/invariant/campaign.rs
```diff
@@ -60,6 +64,26 @@ pub struct InvariantWorkerPlan {
     pub runs: u32,
 }
 
+/// Shared state used only to coordinate invariant worker execution.
+pub struct InvariantCampaignState {
```
grandizzy, line 75: **⚠️ No campaign-wide stop on terminal failure.**

A failure in worker 0 doesn't tell workers 1..N to exit. They keep mutating corpus, recording reverts, and bumping `total_runs` after the earliest logical failure. The aggregator hides this for the final `errors` map but side effects drift (extra runs, reverts, gas traces, corpus pollution).

Add a shared `terminal_stop: AtomicBool` here, set it from `run_invariant_worker` when a `break 'stop` happens for terminal reasons (`MaxAssumeRejects`, single-predicate invariant break with `fail_on_revert`), and OR it into the `continue_campaign` check. Multi-predicate campaigns that intentionally continue after a predicate break should not flip it.

grandizzy (commented):
Good direction — sharding invariant campaigns is the right shape, and the recent fixes (`reported_runs` rename, per-worker timer placement, optimization baseline for non-master workers) are nice. Couple of things flagged inline — the handler-shrink one regresses the **single-worker** path so worth fixing independently of the parallel work.

_3 blockers (1 partially fixed), 4 strong concerns._

The optimization-baseline fix shows the pattern that needs to extend to `in_memory_corpus` and the coverage maps. Happy to re-review once any of the three blockers land.

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -68,6 +72,11 @@ pub use shrink::{
     replay_handler_failure_sequence,
 };
 
+/// Minimum number of logical runs assigned to each invariant worker.
+///
+/// Keeps short campaigns single-threaded and avoids producing many small rayon jobs.
+const MIN_RUNS_PER_INVARIANT_WORKER: u32 = 1000;
```
grandizzy: **⚠️ Too aggressive for coverage-guided campaigns.**

On a 16-core box with `runs=10_000`, you get 10 parallel cold-map campaigns of 1,000 runs each — strictly worse than one 10,000-run campaign for bug finding. Bump to `10_000`–`25_000`, or only shard when `runs >= N * MIN` with enough mutation budget per shard to be useful.

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -891,6 +1035,14 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
             }
 
             runs += 1;
+            let total_runs = campaign_state.increment_runs();
+            debug_assert!(
+                is_timed_campaign || total_runs <= config.runs,
+                "worker runs were not distributed correctly"
+            );
+            if let Some(progress) = progress {
+                progress.inc(1);
```
grandizzy: **⚠️ Progress UX races; JSON pulse drops `(N-1)/N` of activity.**

- `set_message` (above) races between workers and shows one worker's local corpus/failure counters as if campaign-global.
- JSON pulse / `DURATION_BETWEEN_METRICS_REPORT` is gated to `plan.worker_id == 0`, so structured progress output silently drops most campaign activity in parallel mode.
- `progress.inc(1)` moved after the metrics block (was before pre-PR) — please double-check the ETA isn't off by one tick.

Either compute global counters in `InvariantCampaignState` and let only worker 0 render them, or prefix messages with `[wN]`. The JSON pulse should reflect global totals.

0xKarl98:
@grandizzy Have addressed your concerns

grandizzy (commented):
@0xKarl98 that's really nice, tysm! Do I understand properly that if we set `corpus_dir` (that is cov guided fuzzing) or timeout (time based campaigns) then the workers are forced to 1 even if we pass `--invariant-workers 16` for example? If so I don't think this is desired?

0xKarl98:
@grandizzy  check it again

0xKarl98:
@grandizzy Am doing another round of simplification , wait a sec to check

0xKarl98:
@grandizzy  can check it now

grandizzy (approved):
ty, makes sense, the local tests I did shows workers flag is honored now for the configs. Going to do a 1h test to compare with prev.

mattsse (approved):
lgtm

mattsse (approved):
LGTM

0xalpharush:
@grandizzy did we bench 1 vs many workers on this?

grandizzy:
> @grandizzy did we bench 1 vs many workers on this?

only compared with master and one worker and seen no regression, let me redo with multiple on master since this was merged and post back

PR #14994, opened 2026-05-31
fix: make forge rerun filtering contract-aware

BROCCOLO1D:
## Summary
- Persist `forge test --rerun` failures as JSON contract/test pairs instead of only a test-name regex.
- Use contract-aware test filtering when rerunning so same-named tests in other contracts are not selected.
- Keep compatibility with existing persisted rerun failure files by falling back to the legacy regex format when JSON parsing fails.

## Why
`forge test --rerun` could rerun too many tests when multiple contracts defined the same test function name and only one of them failed. The previous failure cache only captured test names, so collection could not distinguish the failed contract from passing contracts with the same test name.

## Changes
- Add a contract-aware test filter hook with a default behavior matching existing filtering.
- Thread contract names/identifiers through forge test collection and execution filtering.
- Add versioned JSON persistence for rerun failures while preserving legacy regex loading.
- Add a regression test with two contracts sharing `testSharedName()` where only the failed contract is rerun.

## Validation
- `cargo fmt --check` (passes; stable rustfmt prints existing warnings for nightly-only rustfmt options)
- `cargo test -p forge --test cli rerun_filters_same_named_tests_by_contract -- --nocapture`
- `cargo test -p forge --test cli replays_random_uint_failure -- --nocapture`

## Scope
Focused bugfix for `forge test --rerun` filtering and persisted failure loading. Regular match-test/match-contract filtering behavior is intended to remain unchanged.

### DIFF
--- a/crates/common/src/traits.rs
+++ b/crates/common/src/traits.rs
@@ -15,6 +15,11 @@ pub trait TestFilter: Send + Sync {
 
     /// Returns a contract with the given path should be included.
     fn matches_path(&self, path: &Path) -> bool;
+
+    /// Returns whether the test should be included for the given contract.
+    fn matches_test_function_in_contract(&self, _contract_name: &str, func: &Function) -> bool {
+        func.is_any_test() && self.matches_test(&func.signature())
+    }
 }
 
 impl<'a> dyn TestFilter + 'a {
--- a/crates/common/src/traits.rs
+++ b/crates/common/src/traits.rs
@@ -17,7 +17,9 @@ pub trait TestFilter: Send + Sync {
     fn matches_path(&self, path: &Path) -> bool;
 
     /// Returns whether the test should be included for the given contract.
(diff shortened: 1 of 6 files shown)


stevencartavia (commented):
Pushed a small follow-up on top with two fixes I hit while testing: `--rerun` was returning "No tests found in project!" when a previous run only had non-replayable failures (e.g. `setUp()` reverts), and `get_sources_to_compile` was filtering artifacts with unstripped ids so exact contract-id matching dropped suites living outside `config.src` (most visible on invariant reruns). Also dropped the now-redundant failure-list check in matches_contract and switched `list()` / `matching_test_functions` to pass the full `path:Contract` identifier so `--list --rerun` agrees with what actually runs.

figtracer (approved):

mablr (approved):
Makes sense

PR #14996, opened 2026-06-01
feat(cast): revoke Tempo wallet sessions on-chain

0xKarl98:
Towards OSS-162

Extend `cast wallet session revoke` to revoke provisioned session keys through the AccountKeychain precompile before marking the local session as revoked


crates/cast/src/cmd/wallet/session.rs
```diff
@@ -114,22 +136,116 @@ async fn run_create(
...
+        let info = provider.get_keychain_key(entry.root_account, entry.key_address).await?;
+        if info.isRevoked {
+            return Ok(Some(SessionRevokeStatus::AlreadyRevoked));
+        }
+        if info.keyId == Address::ZERO {
+            return Ok(Some(SessionRevokeStatus::NotProvisioned));
+        }
+
+        let calldata = IAccountKeychain::revokeKeyCall { keyId: entry.key_address }.abi_encode();
+        send_keychain_tx(calldata, tx, &send_tx, Some(entry.root_account)).await?;
```
mablr: `send_keychain_tx` can return `Ok(())` without submitting anything when `--tempo.print-sponsor-hash` is set ? This path then falls through and marks the session `Revoked` locally even though the on-chain key remains active.

0xKarl98:
@mablr  Have explicitly reject this case , ptal again

mablr (dismissed):
sgtm

crates/cast/src/cmd/wallet/session.rs
```diff
@@ -114,22 +138,123 @@ async fn run_create(
...
+        update_session_status(session_id, SessionStatus::Revoked)?;
+        print_revoke_status(session_id, Some(&entry), SessionRevokeStatus::Local)?;
+        return Ok(());
+    }
+
+    if tx.tempo.print_sponsor_hash {
+        eyre::bail!(PRINT_SPONSOR_HASH_REVOKE_ERROR);
+    }
+
+    update_session_status(session_id, SessionStatus::Revoking)?;
```
grandizzy: **Move the `Revoking` transition until after preflight.**

Since this PR makes `Revoking` clear local key material, any preflight error — bad/missing `--rpc-url`, wrong chain id, `get_chain_id` failure, `get_keychain_key` RPC error, missing root signer, or `ensure_root_sender` mismatch — permanently wipes the local key and marks the session `Failed`. Destructive for purely local config mistakes.

Suggested ordering:
1. reject `--tempo.print-sponsor-hash` (already done)
2. build provider, validate `chain_id`
3. `get_keychain_key` → short-circuit `AlreadyRevoked` / `NotProvisioned` directly to `Revoked`
4. resolve + check root signer
5. **only now** `update_session_status(Revoking)`
6. submit

crates/cast/src/cmd/wallet/session.rs
```diff
@@ -114,22 +138,123 @@ async fn run_create(
...
+        let calldata = IAccountKeychain::revokeKeyCall { keyId: entry.key_address }.abi_encode();
+        match send_keychain_tx(calldata, tx, &send_tx, Some(entry.root_account)).await? {
+            KeychainTxOutcome::Submitted => {}
+            KeychainTxOutcome::PrintedSponsorHash => eyre::bail!(PRINT_SPONSOR_HASH_REVOKE_ERROR),
+        }
+        Ok(None)
+    }
+    .await
+    .map_err(|err: eyre::Report| {
+        let _ = update_session_status(session_id, SessionStatus::Failed);
```
grandizzy: **Don't downgrade `Revoked` to `Failed`.**

Three cases produce the wrong final state today:

- **Retry**: session already `Revoked`, user re-runs with a wrong `--rpc-url` → `Revoked → Revoking → Failed`.
- **Concurrent revokes**: A writes `Revoked`; B errors ("already revoked" revert, nonce conflict, RPC timeout) and overwrites with `Failed`.
- **Partial failure**: tx mined but receipt polling times out → local `Failed`, on-chain revoked.

Re-check on-chain state and current status before writing `Failed`:

```rust
.map_err(|err: eyre::Report| {
    let on_chain_revoked = provider
        .get_keychain_key(entry.root_account, entry.key_address)
        .await
        .map(|info| info.isRevoked)
        .unwrap_or(false);

    if on_chain_revoked {
        let _ = update_session_status(session_id, SessionStatus::Revoked);
    } else if !matches!(
        read_session_entry(session_id).ok().flatten().map(|e| e.status),
        Some(SessionStatus::Revoked)
    ) {
        let _ = update_session_status(session_id, SessionStatus::Failed);
    }
    err.wrap_err("failed to revoke Tempo session key on-chain")
})?;
```

A CAS-style `update_session_status_if(current, next)` helper in `crates/common/src/tempo/session.rs` would be cleaner and reusable.

grandizzy (changes requested):
Two blockers before merge:

1. `Revoking` is set before preflight, so any local config error (bad RPC, wrong chain, missing/mismatched signer) permanently wipes the local key.
2. The failure path can overwrite `Revoked` with `Failed` on retry, concurrent runs, or partial failure where the tx actually landed.

Inline comments below.

0xKarl98:
@grandizzy  have addressed your comments , ptal again

grandizzy (changes requested):
looks good, `wallet_session_revoke_failure_marks_failed_and_clears_key` is failing CI, pls check. thanks!

0xKarl98:
@grandizzy  have fixed the failing ci

grandizzy (approved):
lgtm

mablr (approved):

PR #14998, opened 2026-06-01
feat(cast): support Tempo key auth witnesses

figtracer:
## Summary

- add T5 `witness` support to `cast keychain authorize` and keychain display/JSON paths
- add `cast keychain burn-witness` / `is-witness-burned` helpers for TIP-1053 witness state
- add `cast key-auth encode` / `sign` helpers that preserve optional witness and expose RLP length
- add regression coverage for absent witness vs present `bytes32(0)` and signed witness roundtrips

## Testing

- `cargo check -p cast@1.7.2`
- `cargo test -p cast@1.7.2 key_auth --lib`


grandizzy:
@figtracer pls fix merge conflicts, thanks!

figtracer:
> @figtracer pls fix merge conflicts, thanks!

on it

crates/cast/src/cmd/keychain.rs
```diff
@@ -267,6 +303,67 @@ pub enum KeychainSubcommand {
...
+    /// Chain ID for replay protection.
+    #[arg(long)]
+    chain_id: u64,
+
+    /// Key address to authorize.
+    key_address: Address,
+
+    /// Signature type: secp256k1, p256, or webauthn.
+    #[arg(long, default_value = "secp256k1", value_parser = parse_auth_signature_type)]
+    key_type: AuthSignatureType,
```
grandizzy, line 337: **`cast key-auth sign --key-type p256|webauthn` produces a misleading artifact.**

`WalletOpts` only yields a secp256k1 signer, but `run_key_auth_sign` unconditionally wraps the output as `PrimitiveSignature::Secp256k1(signature)` regardless of `--key-type`. If `--key-type` is intended to be the *authorized key* type (matching `KeyAuthorization.key_type`), then this help text is misleading — users will reasonably assume `--key-type p256` produces a P256 signature, but they'll get a secp256k1 root signature over an authorization that names a P256 access key.

Suggested fix:

```rust
/// Type of access key being authorized: secp256k1, p256, or webauthn.
/// The authorization itself is signed with the configured Ethereum (secp256k1) wallet.
```

Ideally also emit both fields in the JSON output so the artifact can't be misread downstream:

```json
{
  "authorized_key_type": "p256",
  "signature_type": "secp256k1"
}
```

crates/cast/src/cmd/keychain.rs
```diff
@@ -2506,6 +2769,40 @@ async fn run_revoke(
...
 }
 
+/// `cast keychain burn-witness` — burn a TIP-1053 key authorization witness.
+async fn run_burn_witness(
+    witness: B256,
+    tx_opts: TransactionOpts,
+    send_tx: SendTxOpts,
+) -> Result<()> {
+    let calldata = IAccountKeychain::burnKeyAuthorizationWitnessCall { witness }.abi_encode();
+    send_keychain_tx(calldata, tx_opts, &send_tx).await
```
grandizzy: **T5 hardfork gating is asymmetric.**

`run_authorize` correctly errors when `--witness` is used pre-T5:

```rust
if witness.is_some() && !is_tempo_hardfork_active(&provider, TempoHardfork::T5).await? {
    eyre::bail!("--witness requires a Tempo T5-capable AccountKeychain RPC");
}
```

But `cast keychain burn-witness` and `cast keychain is-witness-burned` have no such guard, so on pre-T5 chains users will hit raw precompile/RPC errors instead of an actionable message. Please add the same `is_tempo_hardfork_active(T5)` check to both.

grandizzy (changes requested):
A few correctness items:

grandizzy (dismissed):
Both blockers addressed in 491cebd:

- `key-auth sign` help text now clarifies that `--key-type` is the *authorized* access key type and the authorization itself is signed with the wallet's secp256k1 key; JSON output also emits both `authorized_key_type` and `signature_type` for disambiguation.
- `burn-witness` and `is-witness-burned` both gate on T5 with a clear error message.

Bonus: `--expiry 0` is now rejected, and explicit empty `--scopes '[]'` is correctly distinguished from omitting `--scopes`, with a regression test. Nice.

LGTM.

crates/cast/src/cmd/keychain.rs
```diff
@@ -267,6 +303,68 @@ pub enum KeychainSubcommand {
...
+
+    /// Expiry timestamp (unix seconds). Omit for no expiry.
+    #[arg(long)]
+    expiry: Option<u64>,
+
+    /// Enforce spending limits for this key. With no --limit entries, this means no spending.
+    #[arg(long)]
+    enforce_limits: bool,
+
+    /// Spending limit in `TOKEN:AMOUNT[:PERIOD]` format. Can be specified multiple times.
```
mablr, line 347: This doc comment breaks rustdoc as `TOKEN:AMOUNT[:PERIOD]` is parsed as an intra-doc link.

crates/cast/src/cmd/keychain.rs
```diff
@@ -2496,6 +2690,94 @@ async fn run_authorize(
...
+    let mut out = Vec::new();
+    authorization.encode(&mut out);
+    out
+}
+
+impl KeyAuthArgs {
+    fn into_authorization(self) -> Result<KeyAuthorization> {
+        let (scopes, explicit_scopes_json) =
+            if let Some(AuthScopesJson(json_scopes)) = self.scopes_json {
+                (json_scopes, true)
```
mablr, line 2761: It seems that passing `--scopes '[]'` currently produces unrestricted calls, because empty scopes are dropped instead of encoded as `Some([])`.

crates/cast/src/cmd/keychain.rs
figtracer: this is fixed now

crates/cast/src/cmd/keychain.rs
```diff
@@ -2496,6 +2690,94 @@ async fn run_authorize(
...
+    }
+
+    Ok(())
+}
+
+async fn run_key_auth_sign(args: KeyAuthArgs, wallet: WalletOpts) -> Result<()> {
+    let authorization = args.into_authorization()?;
+    let authorized_key_type = auth_signature_type_name(&authorization.key_type);
+    let signature_hash = authorization.signature_hash();
+    let signer = wallet.signer().await?;
```
mablr: `wallet.signer()` can resolve `--from` to a Tempo access-key signer via the wallet fallback, producing an authorization that will not recover to the expected root/admin account. We should use `maybe_signer()` ? reject `tempo_access_key.is_some()`, and ask users to sign with a persistent root signer?

mablr (approved):
lgtm

grandizzy (approved):

PR #15000, opened 2026-06-01
feat(lint): add reentrancy-no-eth detector

figtracer:
## Summary
- add the medium-severity `reentrancy-no-eth` lint for read-before-write reentrancy around external calls that do not transfer ETH
- reuse the existing reentrancy flow analyzer while tagging pending calls as ETH or no-ETH, keeping `reentrancy-eth` behavior separate
- add lint docs and UI coverage for typed external calls, low-level calls, delegatecall, modifiers, internal helper propagation, and ignored safe cases

### DIFF
--- a/crates/lint/src/sol/high/reentrancy.rs
+++ b/crates/lint/src/sol/high/reentrancy.rs
@@ -283,18 +283,18 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
                     self.analyze_expr(lhs, state);
                 }
                 self.analyze_expr(rhs, state);
+                self.analyze_lhs_indices(lhs, state);
                 let written_vars = state_write_lhs_vars(self.hir, lhs);
                 if !written_vars.is_empty() {
                     self.emit_pending_calls(state, &written_vars);
                 }
-                self.analyze_lhs_indices(lhs, state);
             }
             ExprKind::Delete(inner) => {
+                self.analyze_lhs_indices(inner, state);
                 let written_vars = state_write_lhs_vars(self.hir, inner);
                 if !written_vars.is_empty() {
                     self.emit_pending_calls(state, &written_vars);
                 }
-                self.analyze_lhs_indices(inner, state);
             }
             ExprKind::Unary(op, inner)
                 if matches!(
@@ -505,7 +505,7 @@ fn is_uncapped_value_call(
     opts: Option<&[hir::NamedArg<'_>]>,
 ) -> bool {
     let Some(opts) = opts else { return false };
-    let ExprKind::Member(_, member) = &callee.kind else { return false };
+    let ExprKind::Member(_, member) = &callee.peel_parens().kind else { return false };
     if member.name != kw::Call {
         return false;
     }
--- a/crates/lint/src/sol/high/reentrancy.rs
+++ b/crates/lint/src/sol/high/reentrancy.rs
@@ -1,12 +1,14 @@
 use super::ReentrancyEth;
 use crate::{
     linter::{LateLintPass, LintContext},
-    sol::{Severity, SolLint},
+    sol::{Severity, SolLint, analysis::interface::receiver_contract_id},
 };
 use solar::{
     ast::{LitKind, StateMutability, UnOpKind, Visibility},
     interface::{Span, kw, sym},
-    sema::hir::{self, ExprKind, FunctionId, ItemId, Res, StmtKind, VariableId},
+    sema::hir::{
+        self, CallArgs, ExprKind, FunctionId, ItemId, Res, StmtKind, TypeKind, VariableId,
+    },
 };
 use std::collections::{BTreeSet, HashSet};
 
@@ -15,6 +17,13 @@ declare_forge_lint!(
     Severity::High,
     "reentrancy-eth",
     "state read before ETH transfer is written after the transfer"
+);
+
+declare_forge_lint!(
+    REENTRANCY_NO_ETH,
+    Severity::Med,
+    "reentrancy-no-eth",
+    "state read before external call is written after the call"
 );
 
 impl<'hir> LateLintPass<'hir> for ReentrancyEth {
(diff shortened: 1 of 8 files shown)


crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -470,6 +513,74 @@ fn is_uncapped_value_call(
...
+    let Some(contract_id) = receiver_contract_id(hir, receiver) else { return false };
+    hir.contract_item_ids(contract_id).any(|item| {
+        let Some(func_id) = item.as_function() else { return false };
+        let func = hir.function(func_id);
+        func.name.is_some_and(|name| name.name == member)
+            && func.kind.is_function()
+            && func.parameters.len() == arity
+            && is_externally_callable(func)
+            && func.mutates_state()
+    })
```
mablr: As mentionned for other lints, classify member calls by receiver contract, function name, and arity only, may be not enough.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -429,23 +458,37 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
+fn reentrant_call_kind(
+    hir: &hir::Hir<'_>,
+    callee: &hir::Expr<'_>,
+    args: &CallArgs<'_>,
+    opts: Option<&[hir::NamedArg<'_>]>,
+) -> Option<ReentrantCallKind> {
+    if is_uncapped_value_call(hir, callee, opts) {
+        return Some(ReentrantCallKind::Eth);
+    }
+    is_no_eth_reentrant_call(hir, callee, args, opts).then_some(ReentrantCallKind::NoEth)
```
grandizzy: Asymmetric paren handling: `is_no_eth_reentrant_call` peels parens on the callee (line 526), but `is_uncapped_value_call` (line 498) does not. A parenthesized low-level value call falls through both detectors and gets no warning:

```solidity
(target.call){value: amount}("");
```

Suggested fix in `is_uncapped_value_call`:

```rust
let ExprKind::Member(_, member) = &callee.peel_parens().kind else { return false };
```

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -295,8 +316,8 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
                 for func_id in resolved_function_ids(callee) {
                     self.analyze_internal_call(func_id, state);
                 }
-                if is_uncapped_value_call(self.hir, callee, *opts) {
-                    state.push_call(expr.span);
+                if let Some(kind) = reentrant_call_kind(self.hir, callee, args, *opts) {
+                    state.push_call(expr.span, kind);
```
grandizzy, line 329: Not introduced by this PR, but exposed by the new lint: external calls inside an LHS index expression get pushed onto `pending_calls` after `emit_pending_calls` has already run for the surrounding assignment, so this pattern is missed:

```solidity
uint256 amount = balances[msg.sender];
balances[hook.key()] = 0; // hook.key() can reenter before the storage write
```

In `analyze_expr` (`ExprKind::Assign` around line 272 and `ExprKind::Delete` around line 283), move `analyze_lhs_indices` before the `emit_pending_calls` call so reentrant calls in the index are visible at emit time. Worth a test case.

grandizzy (commented):
Thanks, couple of remarks inline.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -470,6 +513,74 @@ fn is_uncapped_value_call(
...
+    })
+}
+
+fn external_member_call_can_reenter(
+    hir: &hir::Hir<'_>,
+    receiver: &hir::Expr<'_>,
+    member: solar::interface::Symbol,
+    arity: usize,
+) -> bool {
+    let Some(contract_id) = receiver_contract_id(hir, receiver) else { return false };
```
grandizzy: `receiver_contract_id` (in `crates/lint/src/sol/analysis/interface.rs`) only matches a contract-typed variable or a direct `IFoo(addr)` wrap. Common external-call receiver shapes fall through to `None` and produce false negatives for the new lint:

- parenthesized receivers — `(hook).notify(...)`
- `this.foo()` / `super.foo()`
- struct/field receivers — `s.hook.notify(...)`
- return-value receivers — `getHook().notify(...)`

Minimum fix is peeling parens in `receiver_contract_id`:

```rust
pub fn receiver_contract_id(hir: &hir::Hir<'_>, recv: &Expr<'_>) -> Option<ContractId> {
    match &recv.peel_parens().kind {
        ...
```

The remaining shapes likely need Solar's resolved expression type; happy to defer those to a follow-up.

crates/lint/src/sol/high/reentrancy.rs
figtracer: Addressed in 8f67a650f. Typed no-ETH member calls now resolve through Solar semantic members and check argument compatibility instead of matching only receiver contract/name/arity. I also added overload coverage where the mutating overload warns and the same-name/same-arity view overload does not.

crates/lint/src/sol/high/reentrancy.rs
figtracer: This was covered by 8f67a650f. The no-ETH detector no longer uses `receiver_contract_id` here; it resolves typed member calls from Solar expression types, so parenthesized, field, and return-value receivers are not limited by the old helper. `super` remains intentionally excluded because it is internal dispatch, not an external reentrant call.

grandizzy (dismissed):

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -470,6 +523,362 @@ fn is_uncapped_value_call(
...
+    }
+}
+
+fn arg_matches_type<'hir>(
+    gcx: Gcx<'hir>,
+    hir: &'hir hir::Hir<'hir>,
+    arg: &'hir hir::Expr<'hir>,
+    param_ty: Ty<'hir>,
+) -> bool {
+    expr_ty(gcx, hir, arg).is_none_or(|arg_ty| arg_ty.convert_implicit_to(param_ty, gcx))
```
mablr: This causes a false positive for overloaded calls when an argument type cannot be inferred. `expr_ty` returns `None` for expressions like `flag && true`, and `is_none_or` then treats that argument as matching every overload.

I confirmed locally that `hook.overloaded(flag && true)` warns even when it resolves to the `external view` bool overload, because the mutating overload is also considered a match.

mablr (commented):
Found a FP.

mablr (approved):
lgtm

grandizzy (approved):

PR #15010, opened 2026-06-01
fix(mutation): reject empty mutation selections

mablr:
Errors when explicit `--mutate-path` or `--mutate-contract` selects zero source files.


grandizzy (approved):
makes sense

PR #15012, opened 2026-06-01
fix(mutation): copy mutation compiler paths

mablr:
Copies in-root `include_paths` and `allow_paths` into mutation workspaces and rejects external compiler paths with a clear error.

### DIFF
--- a/crates/forge/src/mutation/runner.rs
+++ b/crates/forge/src/mutation/runner.rs
@@ -611,6 +611,16 @@ fn temp_config_for_mutation(config: &Config, temp_path: &Path) -> Config {
     temp_config.mutation_dir = rebase_project_path(&config.root, temp_path, &config.mutation_dir);
     temp_config.libs =
         config.libs.iter().map(|lib| rebase_project_path(&config.root, temp_path, lib)).collect();
+    temp_config.include_paths = config
+        .include_paths
+        .iter()
+        .map(|path| rebase_project_path(&config.root, temp_path, path))
+        .collect();
+    temp_config.allow_paths = config
+        .allow_paths
+        .iter()
+        .map(|path| rebase_project_path(&config.root, temp_path, path))
+        .collect();
 
     if let Some(path) = &config.fuzz.failure_persist_dir {
         temp_config.fuzz.failure_persist_dir =
@@ -784,6 +794,8 @@ mod tests {
             broadcast: root.join("custom-broadcast"),
             mutation_dir: root.join("custom-cache/mutation"),
             libs: vec![root.join("vendor")],
+            include_paths: vec![root.join("shared")],
+            allow_paths: vec![root.join("fixtures")],
             dynamic_test_linking: true,
             cache: true,
             ..Default::default()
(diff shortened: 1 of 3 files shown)


grandizzy (changes requested):
## 🚫 Blocker 1: extras-first ordering corrupts `lib/` and `node_modules/`

In `workspace.rs`, `copy_project` now runs in this order:

```text
src/         (copy)
test/        (copy)
include_paths + allow_paths   ← NEW, copied here
script/      (copy)
libs/        (symlink, GUARDED BY: if !target.exists())
node_modules/, dependencies/  (symlink, GUARDED BY: if !target.exists())
```

The lib/dep loops use:

```rust
if !target.exists() {
    if let Some(parent) = target.parent() { fs::create_dir_all(parent)?; }
    if symlink_dir(lib_path, &target).is_err() {
        copy_dir_recursive(lib_path, &target)?;
    }
}
```

So any `include_paths` or `allow_paths` entry that's a subpath of a configured lib will pre-create the lib's directory, causing the later guarded symlink to silently skip:

- **`include_paths = ["lib/foo"]`, `libs = ["lib"]`** → extras copy creates `temp/lib/foo`. Libs loop sees `temp/lib` exists, skips symlinking the rest. `lib/bar`, `lib/baz`, etc. **missing from the mutation workspace** → compile failures masquerading as "Invalid" mutants.
- **`include_paths = ["node_modules"]`** → extras eagerly recursively *copies* the entire `node_modules/` (potentially GB), instead of the existing fast symlink path.

Fix options:
1. Reorder: copy extras **after** libs/node_modules/script.
2. Skip extras whose path is already covered by `src`, `test`, `script`, any `libs` root, or `node_modules`/`dependencies`.

(2) is the cleaner answer — extras that overlap an already-handled root are no-ops semantically. The integration test in this PR uses a top-level `include/` dir that doesn't overlap, so the bug isn't caught.

## 🚫 Blocker 2: dead branch in runner contradicts the workspace-copy policy

In `runner.rs::test_single_mutant_isolated`:

```rust
temp_config.include_paths = config.include_paths.iter()
    .map(|path| {
        let rel = workspace::relative_to_root(&config.root, path);
        if rel.is_absolute() { path.clone() } else { temp_path.join(rel) }
    })
    .collect();
```

The `if rel.is_absolute() { path.clone() }` branch silently keeps external absolute paths pointing at the original project root — but `copy_project` already ran and would have **bailed** on any external path via `ensure_safe_relative_path(...)` in `copy_extra_project_path`. So this branch is unreachable; if it ever were reached, it would defeat workspace isolation by letting the mutant compiler read the original source.

Fix: delete the silent fallback. Either trust `copy_project`'s validation and unconditionally do `temp_path.join(rel)` (which still has the Rust footgun that joining an absolute `rel` returns the absolute path unchanged — so add a `debug_assert!(!rel.is_absolute())`), or extract a checked `rebase_under_root(root, temp, path, label) -> Result<PathBuf>` helper used by both `include_paths` and `allow_paths`.

grandizzy (approved):

PR #15013, opened 2026-06-01
fix(mutation): filter no-op mutation outputs

mablr:
Filters no-op assignment and require mutants, including unchanged `require(true)` and `zero-to-zero` assignment replacements.

### DIFF
--- a/crates/forge/src/mutation/mutators/require_mutator.rs
+++ b/crates/forge/src/mutation/mutators/require_mutator.rs
@@ -15,7 +15,7 @@
 //! - State preconditions (reentrancy guards, paused checks)
 
 use eyre::Result;
-use solar::ast::{CallArgsKind, ExprKind};
+use solar::ast::{CallArgsKind, Expr, ExprKind, UnOpKind};
 
 use super::{MutationContext, Mutator};
 use crate::mutation::mutant::{Mutant, MutationType};
@@ -59,9 +59,7 @@ impl Mutator for RequireMutator {
         let line_number = context.line_number();
         let column_number = context.column_number();
 
-        // Extract condition text from source
         let source = context.source.unwrap_or("");
-        let condition_text = extract_span_text(source, condition_expr.span);
 
         // Build the rest of the call (message and other arguments after the condition,
         // if any) using span-based extraction so that commas inside the condition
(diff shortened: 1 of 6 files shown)


crates/forge/src/mutation/mutators/require_mutator.rs
```diff
@@ -145,3 +134,12 @@ fn extract_span_text(source: &str, span: solar::ast::Span) -> String {
...
 }
+
+fn invert_condition_text(condition: &str) -> String {
+    let condition = condition.trim();
+    if let Some(stripped) = condition.strip_prefix('!') {
+        stripped.trim().to_string()
+    } else {
+        format!("!({condition})")
+    }
+}
```
grandizzy: ## 🚫 Blocker: textual `!` strip is unsafe for composite conditions

This assumes a leading `!` in source text means the whole condition is `!X`. Only true when the entire condition is a top-level unary-not. For composite conditions that merely *start* with `!`, the strip produces something that isn't the logical inverse:

| Original | PR produces | Correct inverse |
|---|---|---|
| `require(!paused)` | `require(paused)` ✅ | `require(paused)` |
| `require(!paused && isOwner)` | `require(paused && isOwner)` ❌ | `require(!(!paused && isOwner))` |
| `require(!a \|\| b)` | `require(a \|\| b)` ❌ | `require(!(!a \|\| b))` |

The new `require_not` test only covers `require(!paused, "paused")` — the safe case — so this lands without coverage. Real-world `require(!X && …)` patterns will get mutants that aren't the inverted condition, producing wrong mutation-testing signal.

### Fix

Either always-wrap:

```rust
fn invert_condition_text(condition: &str) -> String {
    format!("!({})", condition.trim())
}
```

(cosmetic cost: `require(!x) → require(!(!x))` instead of `require(x)`, functionally identical)

Or AST-aware, to keep the pretty `!x → x` form only when safe:

```rust
let inverted_condition = match &condition_expr.kind {
    ExprKind::Unary(op, inner) if op.kind == UnOpKind::Not => {
        extract_span_text(source, inner.span)
    }
    _ => format!("!({})", condition_text.trim()),
};
```

Please add a `require(!a && b)` test case alongside.

grandizzy (changes requested):
One blocker on `invert_condition_text` — see inline.

crates/forge/src/mutation/mutators/require_mutator.rs
mablr: https://github.com/foundry-rs/foundry/pull/15013/commits/a9522a03d70f2f202823502a6999dc25a6ed52bf

grandizzy (approved):

PR #15015, opened 2026-06-01
fix(mutation): surface mutator generation errors

mablr:
Changes `MutatorRegistry::generate_mutations` to return `Result` and threads mutator failures through AST generation instead of dropping them.

### DIFF
--- a/crates/forge/src/mutation/mod.rs
+++ b/crates/forge/src/mutation/mod.rs
@@ -1,7 +1,7 @@
 use std::{
     collections::{BTreeMap, HashSet, hash_map::DefaultHasher},
     hash::{Hash, Hasher},
-    path::PathBuf,
+    path::{Path, PathBuf},
     sync::Arc,
 };
 
@@ -16,6 +16,7 @@ pub use crate::mutation::{
     runner::run_mutations_parallel_with_progress,
 };
 use eyre::eyre;
+use foundry_common::sh_warn;
 use serde::{Deserialize, Serialize};
 use solar::{
     ast::{
@@ -25,6 +26,10 @@ use solar::{
     },
     parse::Parser,
 };
+
+fn failed_to_parse(path: &Path) -> eyre::Report {
+    eyre!("failed to parse {}", path.display())
+}
 
 #[derive(Clone, Copy)]
 enum CacheKind<'a> {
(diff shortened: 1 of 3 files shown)


grandizzy (changes requested):
Main concern is a design inconsistency, plus a few smaller items.

### Design issue: the PR mixes fail-fast and best-effort error models

- **Registry (`mutator_registry.rs`)** is **fail-fast within a context** — `?` discards earlier mutators' successes for that expression when a later one errors.
- **Visitor (`visitor.rs`)** is **best-effort across contexts** — `collect_mutations` swallows errors into `Vec<Report>` and keeps walking.
- **`generate_ast` (`mod.rs`)** then reverts to **fail-fast at the file level**:

  ```rust
  if let Some(err) = mutant_visitor.errors.into_iter().next() {
      return Err(err);
  }
  Ok(mutant_visitor.mutation_to_conduct)
  ```

  A single mutator failure on any expression invalidates every successful mutant collected from the rest of the file, while also dropping `errors[1..]` on the floor.

This is mostly theoretical today — none of the current mutators (`AssemblyMutator`, `BinaryOpMutator`, `RequireMutator`, etc.) can actually return `Err` in normal traversal because their `is_applicable` filters guarantee context shape, and the rest are infallible. But the new `FailingMutator` test proves the path is reachable, and a future mutator (or a buggy `is_applicable`) will hit it.

Pick one model and apply it end-to-end:

- **Fail-fast:** make the visitor break on the first error (`ControlFlow::Break` in `visit_expr` / `visit_variable_definition` / `visit_yul_expr`), swap `Vec<Report>` for `Option<Report>`, and the current `generate_ast` shape is fine.
- **Best-effort (preferable for mutation testing):** `generate_ast` returns `Ok(mutations)` always, emits collected errors via the session emitter or `tracing::warn!`, and reserves `Err` for parser-level fatals.

### Other items

- **`Vec<Report>` only surfaces `.next()`.** Even in fail-fast, this should be `Option<Report>`. In best-effort, all errors should be reported, not just the first.
- **Parser error asymmetry.** `Parser::from_lazy_source_code(...).map_err(|_e| eyre!(...))` drops `_e` silently; the adjacent `parse_file().map_err(|e| { e.emit(); ... })` does emit. Make them consistent — ideally one helper, since both wrap to the same `"failed to parse {path}"` message that's now duplicated in three places.
- **`pub errors: Vec<Report>` leaks impl detail.** Make it private with `take_errors(&mut self) -> Vec<Report>` (or `errors(&self) -> &[Report]`).
- **Test coverage gap.** The registry test is good, but it doesn't exercise the visitor wiring. The existing `MutatorTester` helper in `tests/helper.rs` only reads `mutant_visitor.mutation_to_conduct` and never inspects `mutant_visitor.errors`, so visitor-level failures slip past every existing test. Add a visitor test with one `FailingMutator` and one passing mutator that asserts both the error is surfaced *and* the successful mutations are returned (per whichever model you pick).
- **Error context.** `collect_mutations` wraps with the file path — also include line/column from `MutationContext::line_number()` / `column_number()` so the error pins the offending expression, not just the file.

grandizzy (approved):
lgtm!

PR #15016, opened 2026-06-01
fix(mutation): bound timed-out mutation cleanup

mablr:
Clarifies best-effort timeout behavior and bounds pending timed-out worker cleanup handles by mutation job count.

### DIFF
--- a/crates/config/src/mutation.rs
+++ b/crates/config/src/mutation.rs
@@ -57,7 +57,8 @@ pub struct MutationConfig {
     /// Note: enforcement is best-effort. Background work for a timed-out
     /// mutant may continue briefly until the underlying compile / test loop
     /// reaches a checkpoint, but the worker slot is freed immediately so
-    /// other mutants can proceed.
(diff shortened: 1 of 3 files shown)


grandizzy (changes requested):
### `#[derive(Default)]` on `SharedMutationState` is now a footgun this PR introduced

`SharedMutationState` still derives `Default` (`runner.rs:55`). With the new `max_pending_workers: AtomicUsize` field, `Default::default()` produces `AtomicUsize::new(0)`, which `park_timed_out_worker` then clamps to `1` via `.max(1)`. So any caller using `SharedMutationState::default()` silently gets a backlog cap of **1**, not the unbounded `usize::MAX` that the three explicit constructors set.

No internal caller hits this today, but it's a silent trap for future code (especially tests) and an outright contradiction of the "constructors default to unbounded" pattern this PR establishes.

Fix:

```rust
impl Default for SharedMutationState {
    fn default() -> Self {
        Self::new()
    }
}
```

and drop `#[derive(Default)]` from the struct.

grandizzy (approved):

PR #15024, opened 2026-06-02
fix(symbolic): fail closed on validity-sensitive helpers

figtracer:
## Summary

Fail closed on a few remaining symbolic surfaces that can otherwise fabricate successful/observable behavior:

- reject symbolic BN254 precompile inputs because success depends on curve/pairing validity checks that are not modeled
- reject symbolic or invalid BLAKE2f final flags while keeping symbolic payload bytes supported when the final flag is concrete-valid
- reject `vm.setEvmVersion` / `vm.getEvmVersion` under symbolic execution instead of treating set as a no-op and get as always `"cancun"`
- document larger MVP limitations around explicit CALL gas caps, symbolic account/code query scope, and approximate `allowCheatcodes`


grandizzy (approved):
makes sense

PR #15029, opened 2026-06-02
fix(evm): derive invariant worker RNGs from fuzz seed

0xKarl98:
This is a reproducibility fix for sharded invariant campaigns

Worker 0 keeps the original fuzz seed to preserve the master worker stream. Other workers use `keccak256(seed || worker_id)` and initialize their own ChaCha `TestRng` from that derived seed. When no fuzz seed is configured, the existing stochastic behavior is preserved.

### DIFF
--- a/crates/evm/evm/src/executors/invariant/mod.rs
+++ b/crates/evm/evm/src/executors/invariant/mod.rs
@@ -6,7 +6,9 @@ use crate::{
     inspectors::Fuzzer,
 };
 use alloy_json_abi::Function;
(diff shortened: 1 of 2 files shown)


mablr (commented):
Sgtm overall. Do we also want CLI test in addition to unit tests ?

grandizzy:
> Sgtm overall. Do we also want CLI test in addition to unit tests ?

I think unit tests only are OK to cover this logic

grandizzy (approved):
lgtm, thanks!

mablr (approved):

PR #15030, opened 2026-06-02
fix(mutation): handle cancellation without progress UI

mablr:
## Summary

Fixes mutation testing cancellation when `--show-progress` is not enabled.

## Changes

- Install Ctrl+C handling for every mutation run, not only progress UI runs.
- Return cancellation state from the mutation runner batch result.
- Use that cancellation state in the orchestrator to avoid persisting partial result caches.

### DIFF
--- a/crates/forge/src/mutation/runner.rs
+++ b/crates/forge/src/mutation/runner.rs
@@ -51,6 +51,13 @@ pub struct MutantTestResult {
     pub result: MutationResult,
 }
 
+/// Result of a parallel mutation batch.
+#[derive(Debug, Clone)]
+pub struct MutationBatchResult {
+    pub results: Vec<MutantTestResult>,
+    pub cancelled: bool,
+}
+
 /// Tracks progress and adaptive span skipping across parallel workers.
 pub struct SharedMutationState {
     /// Spans where mutations have survived - shared across workers for adaptive skipping.
@@ -192,10 +199,10 @@ pub fn run_mutations_parallel_with_progress(
     filter_args: FilterArgs,
     selected_sources_relative: Arc<Vec<PathBuf>>,
     isolate: bool,
-) -> Result<Vec<MutantTestResult>> {
+) -> Result<MutationBatchResult> {
     let total = mutants.len();
     if total == 0 {
-        return Ok(vec![]);
+        return Ok(MutationBatchResult { results: vec![], cancelled: false });
     }
 
     // Default to available parallelism if num_workers is 0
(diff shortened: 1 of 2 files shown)


crates/forge/src/mutation/runner.rs
```diff
@@ -337,22 +351,23 @@ pub fn run_mutations_parallel_with_progress(
         let _ = handle.join();
     }
 
+    let cancelled = shared_state.is_cancelled();
```
grandizzy, line 303: **Tail race: read `cancelled` after joining the listener.**

```rust
let cancelled = shared_state.is_cancelled();
...
let _ = shutdown_tx.send(());
let _ = ctrlc_handle.join();
```

A Ctrl+C landing in this window can be taken by the `select!`'s shutdown branch instead of the signal branch, so the batch returns `cancelled: false` and the orchestrator continues to the next file. Suggest reordering:

```rust
let _ = shutdown_tx.send(());
let _ = ctrlc_handle.join();
let cancelled = shared_state.is_cancelled();
```

crates/forge/src/mutation/runner.rs
```diff
@@ -238,29 +245,36 @@ pub fn run_mutations_parallel_with_progress(
...
+        .num_threads(num_workers)
+        .stack_size(16 * 1024 * 1024) // 16MB stack to avoid overflow in deep call chains
+        .build()
+        .map_err(|e| eyre::eyre!("Failed to create thread pool: {}", e))?;
+
+    // Set up Ctrl+C handling for every mutation run, not only the progress UI
+    // path. The shutdown channel lets us join the listener when this batch
+    // finishes instead of leaving a parked thread behind after each run.
+    let (shutdown_tx, shutdown_rx) = tokio::sync::oneshot::channel::<()>();
+    let ctrlc_handle = {
```
grandizzy: **Per-batch signal listener is misleading.**

`tokio::signal::ctrl_c()` installs a *process-global* handler that isn't really uninstalled when this per-batch thread exits. Between batches, and during the post-batch cache/report work after `ctrlc_handle.join()`, there's a blind spot where SIGINT is caught by neither this listener nor the default handler.

Consider installing the Ctrl+C listener **once in `run_mutation_testing`** (orchestrator) and passing an `Arc<AtomicBool>` into each batch's `SharedMutationState`. That removes the per-batch thread + tokio runtime churn and closes the blind-spot windows.

grandizzy (changes requested):
A couple of issues worth addressing before merge. The `MutationBatchResult.cancelled` plumbing and the orchestrator cache guard are the right shape; comments below focus on the runner-side signal handling.

crates/forge/src/mutation/runner.rs
mablr: d89bed69a5f0471d708cc724b306591af3a79c33

crates/forge/src/mutation/runner.rs
mablr: d89bed69a5f0471d708cc724b306591af3a79c33

crates/forge/src/mutation/runner.rs
```diff
@@ -76,11 +83,15 @@ pub struct SharedMutationState {
 
 impl SharedMutationState {
     pub fn new() -> Self {
```
grandizzy: we have now six way to construct the `SharedMutationState`, should we collapse to a single `new(cancelled: Arc<AtomicBool>, silent: bool, progress: Option<MutationProgress>)` one to reduce loc

grandizzy (commented):
looks good, one nit which I think we should address

crates/forge/src/mutation/runner.rs
mablr: Yes, much better! 59f223fbd

grandizzy (approved):

PR #15032, opened 2026-06-02
feat(traces): decode Tempo T5 surfaces

figtracer:
## Summary
- register T5 Tempo channel reserve ABI functions/events in trace decoding
- decode TIP-20 logo URI, factory logo overload, DEX OrderFlipped, and channel reserve surfaces through the shared ABI registry
- annotate top-level Tempo calls with payment/general lane labels using the generated T5 classifiers
- render TIP-1034 packed channel reserve storage as settled/deposit/closeRequestedAt


crates/evm/traces/src/decoder/mod.rs
```diff
@@ -900,6 +909,47 @@ impl CallTraceDecoder {
...
+                    |signature| PrimitiveSignature::from_bytes(signature).is_ok(),
+                ) {
+                    "[payment-lane]"
+                } else {
+                    "[general:channel-reserve-selector]"
+                },
+            );
+        }
+
+        Some("[general:target]")
```
stevencartavia: `t5_lane_annotation` falls through to `[general:target]` for every depth-0 non-create call on Tempo that isn't TIP-20 / channel reserve. That means ordinary test contracts, `setUp`, and unrelated callees all get `[general:target] MyContract` in their label (and unlabeled addresses get one synthesized).

Was the catch-all intended? Feels noisy — could we drop it and only annotate the TIP-20 / channel-reserve cases?

crates/evm/traces/src/decoder/mod.rs
```diff
@@ -900,6 +909,47 @@ impl CallTraceDecoder {
         }
         format_token(value)
     }
+
+    fn annotated_label(&self, trace: &CallTrace, label: Option<String>) -> Option<String> {
+        if let Some(annotation) = self.t5_lane_annotation(trace) {
+            let target = label.unwrap_or_else(|| trace.address.to_string());
+            Some(format!("{annotation} {target}"))
```
stevencartavia: The annotation is prepended into `DecodedCallTrace.label`, which is serialized. So `forge test --json` / `cast run --json` on Tempo now emit `"label":"[payment-lane] PathUSD"` instead of `"PathUSD"` — behavior change for anyone consuming the trace JSON.

Could we keep this as a display-only thing (apply in the writer or a separate field) so `label` stays stable?

stevencartavia (dismissed):

grandizzy (dismissed):

grandizzy:
@figtracer pls fix merge conflicts, ty!

mablr (approved):

grandizzy (approved):

PR #15033, opened 2026-06-02
chore(chisel): use solar expression types

figtracer:
Stacked on #15003.

This updates Chisel for Solar's expression type table from paradigmxyz/solar#833:

- points the Solar dependency at current Solar master (4393d0db) while #15003 is still on the pre-merge rev
- enables Solar typeck for Chisel analysis so Gcx::type_of_expr is populated
- prefers Solar expression types for inspected expressions, keeping the HIR fallback for cases Solar does not type or where Chisel needs literal-kind behavior
- fixes ABI mapping for TyKind::Slice and updates Chisel type expectations for precise literal widths

### DIFF
--- a/crates/chisel/src/executor.rs
+++ b/crates/chisel/src/executor.rs
@@ -438,6 +438,18 @@ fn format_event_definition(gcx: Gcx<'_>, event: &Event<'_>) -> Result<String> {
 ///
 /// `lookup` controls whether user-defined type names are resolved via the HIR.
 fn expr_to_dyn(gcx: Gcx<'_>, expr: &Expr<'_>, lookup: bool) -> Option<DynSolType> {
+    if matches!(expr.kind, ExprKind::Lit(lit) if matches!(lit.kind, LitKind::Str(StrKind::Hex, ..)))
+    {
+        return expr_to_dyn_fallback(gcx, expr, lookup);
+    }
+
+    gcx.type_of_expr(expr.id)
+        .and_then(|ty| solar_ty_to_dyn(gcx, ty))
+        .or_else(|| expr_to_dyn_fallback(gcx, expr, lookup))
+}
+
+/// HIR-based expression type inference for cases not covered by Solar's expression type table.
+fn expr_to_dyn_fallback(gcx: Gcx<'_>, expr: &Expr<'_>, lookup: bool) -> Option<DynSolType> {
     match &expr.kind {
         // Elementary type expression: `uint256`, `address`, etc.
         ExprKind::Type(ty) => hir_ty_to_dyn(gcx, ty),
@@ -1093,10 +1105,11 @@ fn solar_ty_to_dyn<'gcx>(gcx: Gcx<'gcx>, ty: Ty<'gcx>) -> Option<DynSolType> {
             let size: usize = n.try_into().ok()?;
             Some(DynSolType::FixedArray(Box::new(inner), size))
         }
-        TyKind::DynArray(elem) | TyKind::Slice(elem) => {
+        TyKind::DynArray(elem) => {
             let inner = solar_ty_to_dyn(gcx, elem)?;
             Some(DynSolType::Array(Box::new(inner)))
         }
+        TyKind::Slice(array) => solar_ty_to_dyn(gcx, array),
         TyKind::Tuple(tys) => {
             Some(DynSolType::Tuple(tys.iter().filter_map(|t| solar_ty_to_dyn(gcx, *t)).collect()))
         }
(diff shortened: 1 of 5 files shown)


mablr (commented):
I think this doesn't leverage properly expr types exposed by solar. Hacky expr type inference logic should go.

crates/chisel/src/executor.rs
```diff
@@ -1509,6 +842,8 @@ mod tests {
                 return None;
             }
 
+            let _ = c.analysis();
```
DaniPopes: move to the "stage 1" block, including the controlflow

crates/chisel/src/executor.rs
```diff
@@ -429,609 +370,16 @@ fn format_event_definition(gcx: Gcx<'_>, event: &Event<'_>) -> Result<String> {
...
-        };
-    }
-
-    // Members on a resolved DynSolType (`.length`, `.pop`, `.selector`, `.address`).
-    if let Some(lhs_ty) = expr_to_dyn(gcx, lhs, lookup)
-        && let Some(ty) = dyn_member(&lhs_ty, member.as_str())
+fn expr_to_dyn(gcx: Gcx<'_>, expr: &Expr<'_>) -> Option<DynSolType> {
+    // Solar currently lowers hex string literals to `StringLiteral`; ABI inspection should decode
+    // them as bytes, matching Solidity's ABI type for `hex"..."` literals.
+    if matches!(expr.kind, ExprKind::Lit(lit) if matches!(lit.kind, LitKind::Str(StrKind::Hex, ..)))
```
DaniPopes: handle it in solar to dyn?

DaniPopes (approved):
rest lgtm

mablr (approved):
Rebased on `mablr/bump_solar` which is now clean.

LGTM

crates/chisel/src/executor.rs
```diff
@@ -418,6 +411,17 @@ const fn elementary_to_dyn(et: ElementaryType) -> Option<DynSolType> {
 }
 
 /// Maps a solar [`Ty`] to a [`DynSolType`].
+fn solar_expr_ty_to_dyn<'gcx>(gcx: Gcx<'gcx>, ty: Ty<'gcx>, expr: &Expr<'_>) -> Option<DynSolType> {
+    // Solar currently lowers hex string literals to `StringLiteral`; ABI inspection should decode
+    // them as bytes, matching Solidity's ABI type for `hex"..."` literals.
+    if matches!(expr.kind, ExprKind::Lit(lit) if matches!(lit.kind, LitKind::Str(StrKind::Hex, ..)))
```
DaniPopes, line 419: what is the test case that fails for this? i think maybe you need to inspect the type of the expression that is inside the inspector abi.decode call ?

figtracer: https://github.com/foundry-rs/foundry/pull/15033/changes/5a0e99e070c42be34930045e58cd7b9fa11db495 `hex_literal_inspection_type` fails without the override

DaniPopes: why cant you do this in the match? StringLiteral() => ...

crates/chisel/src/executor.rs
```diff
@@ -1084,6 +411,18 @@ const fn elementary_to_dyn(et: ElementaryType) -> Option<DynSolType> {
 }
 
 /// Maps a solar [`Ty`] to a [`DynSolType`].
+fn solar_expr_ty_to_dyn<'gcx>(gcx: Gcx<'gcx>, ty: Ty<'gcx>, expr: &Expr<'_>) -> Option<DynSolType> {
+    // `expr` is the inspected expression inside Chisel's generated `abi.encode(...)` call. Solar
+    // currently reports hex string literals as `StringLiteral`, but solc ABI-encodes
+    // `hex"..."` literals as dynamic bytes in that context.
+    if matches!(expr.kind, ExprKind::Lit(lit) if matches!(lit.kind, LitKind::Str(StrKind::Hex, ..)))
```
stevencartavia, line 419: The override only matches top-level `ExprKind::Lit`, but Solar keeps parens in HIR as `ExprKind::Tuple([Some(inner)])`. So `(hex"6869")` slips through and displays as `string` (with UTF-8 "hi") instead of `dynamic bytes`. Bare `hex"6869"` works fine.

Fix is one line using Solar's helper:

```suggestion
    let expr = expr.peel_parens();
    if matches!(expr.kind, ExprKind::Lit(lit) if matches!(lit.kind, LitKind::Str(StrKind::Hex, ..)))
```

Might be worth extending `hex_literal_inspection_type` to cover the parenthesized form too.

mablr (approved):

DaniPopes (approved):

PR #15034, opened 2026-06-02
feat(tempo): add TIP-20 logo URI tooling

mablr:
## Summary

Adds Tempo TIP-20 T5 logo URI tooling across `cast` and Foundry cheatcodes.

Closes OSS-264

## Changes

- Adds `cast tip20 create --logo-uri` for creating TIP-20 tokens with an initial logo URI, including client-side validation and a compatibility check for T5-capable TIP20Factory deployments.
- Adds `cast tip20 logo-check` for offline validation of logo URI values against T5 constraints.
- Adds `cast tip20 logo-set` for updating a TIP-20 token's logo URI, with support for existing Tempo send flows including browser signing, access keys, sponsor config, sponsor URLs, lane resolution, and sponsor hash printing.
- Introduces shared TIP-20 logo URI validation helpers enforcing the 256-byte limit and allowed schemes: `https`, `http`, `ipfs`, and `data`.
- Adds `vm.setTip20LogoURI` / `vm.setLogoURI` cheatcodes that write the TIP-20 logo URI storage slot directly while preserving the same validation rules.
- Adds `vm.expectTip20LogoURIUpdated` / `vm.expectLogoURIUpdated` helpers for asserting TIP-20 `LogoURIUpdated` events.


crates/cast/src/cmd/tip20/mod.rs
```diff
@@ -109,3 +176,178 @@ impl Tip20Subcommand {
...
     }
 }
+
+pub(super) async fn resolve_tip20_signer(
+    send_tx: &SendTxOpts,
+) -> eyre::Result<(Option<WalletSigner>, Option<TempoAccessKeyConfig>)> {
+    if send_tx.eth.wallet.from.is_some() {
+        send_tx.eth.wallet.maybe_signer().await
+    } else {
+        Ok((None, None))
```
stevencartavia: The `if from.is_some()` gate skips pre-resolution when `--from`/`ETH_FROM` isn't set, so `--private-key`, `--keystore`, etc. don't get pre-resolved on their own. This breaks `--tempo.print-sponsor-hash`, which then errors with "requires a signer" even when one was passed.

Repro (replace `0x...0123` with any valid address, no `ETH_FROM`):
```bash
cast tip20 logo-set 0x0000000000000000000000000000000000000123 "https://x/" \
    --private-key $PK --tempo.print-sponsor-hash --rpc-url $RPC \
    --nonce 0 --gas-limit 100000 --gas-price 1 --priority-gas-price 1
# Error: --tempo.print-sponsor-hash requires a signer (e.g. --private-key)
```
Adding `--from $ADDR` makes it work. The same `cast send` invocation works without `--from`.

Fix: drop the gate and call `maybe_signer().await` unconditionally, like `cast send` does.

crates/cast/src/cmd/tip20/mod.rs
```diff
@@ -109,3 +176,178 @@ impl Tip20Subcommand {
...
+            tx,
+            signer,
+            &ak,
+            send_tx.cast_async,
+            send_tx.confirmations,
+            timeout,
+        )
+        .await?;
+    } else if let Some(sponsor_url) = sponsor_url {
+        let signer = pre_resolved_signer.unwrap_or(send_tx.eth.wallet.signer().await?);
```
stevencartavia:
```rust
let signer = pre_resolved_signer.unwrap_or(send_tx.eth.wallet.signer().await?);
```
`unwrap_or`'s argument is evaluated eagerly, so `signer()` runs even when `pre_resolved_signer` is `Some` — the signer is resolved twice. When that happens (e.g. `--interactive --from`, `--keystore --from`), the user is prompted for their private key twice, or the keystore is decrypted twice (~2× wall-clock time).

Verified via `expect`-driven prompt count (1 for `cast send`, 2 for `cast tip20 logo-set`/`create`) and timing (~0.40s vs ~0.80s with a keystore).

Same issue on line 315.

Fix:
```rust
let signer = match pre_resolved_signer {
    Some(s) => s,
    None => send_tx.eth.wallet.signer().await?,
};
```

mablr: https://github.com/foundry-rs/foundry/pull/15034/commits/ff48d750528547fb502b203167dad9a85e2118a3

crates/cast/src/cmd/tip20/mod.rs
mablr: https://github.com/foundry-rs/foundry/pull/15034/commits/ff48d750528547fb502b203167dad9a85e2118a3

grandizzy (approved):
lgtm

figtracer (approved):

PR #15035, opened 2026-06-02
refactor(lint): use expr types exposed by solar

mablr:
## Description

Stacked on #15003

Use expr types newly exposed by solar (https://github.com/paradigmxyz/solar/pull/833) in `forge-lint`.

- Removed most of type inferences hacks
- fixed a FN in `encode-packed-collision` lint thanks to the solar overload handling

More simplifications are certainly unlocked by this solar update, as well as FN/FP fixes. For now I limited the scope of this PR to obvious/trivial improvements, to keep it contained and "easy" to review.


crates/lint/src/sol/low/return_bomb.rs
```diff
@@ -340,409 +94,17 @@ fn member_is_builtin_address(base: &hir::Expr<'_>, member: Symbol) -> bool {
...
-            .strukt(*id)
-            .fields
-            .iter()
-            .any(|&field| is_dynamic_type(hir, &hir.variable(field).ty)),
+fn is_dynamic_ty(ty: Ty<'_>) -> bool {
+    match ty.peel_refs().kind {
+        TyKind::Elementary(ElementaryType::Bytes | ElementaryType::String)
+        | TyKind::DynArray(_)
+        | TyKind::Slice(_) => true,
+        TyKind::Array(element, _) => is_dynamic_ty(element),
```
stevencartavia: **Regression: `return-bomb` no longer flags struct returns containing dynamic fields**

The new `is_dynamic_ty` doesn't handle `TyKind::Struct`. The old `is_dynamic_type` recursed into struct fields. Solar represents struct returns as `Ref(Struct(id), _)` → `peel_refs()` → `Struct(_)` → no match → returns `false`.

Repro:
```solidity
interface ITarget {
    struct R { bytes data; }
    function fetch() external returns (R memory);
}
contract C {
    function bug(ITarget t, uint256 g) public {
        t.fetch{gas: g}(); // base branch warns; this PR is silent
    }
}
```
Also reproduces with `string` fields and nested structs. Tuples still work (case in the match), so the regression is struct-specific.

Fix: delegate to Solar's [`Ty::is_dynamically_encoded(gcx)`](https://github.com/paradigmxyz/solar/blob/main/crates/sema/src/ty/ty.rs#L545-L554), which already handles structs (with a recursion guard) and arrays. Keep the tuple arm since `is_dynamically_encoded` doesn't cover it. Add the case above to `ReturnBomb.sol`.

mablr: https://github.com/foundry-rs/foundry/pull/15035/commits/7e680867eb492025acb31427aa2cf106d7dd40ae

figtracer (approved):

stevencartavia (approved):

PR #15040, opened 2026-06-03
feat(tempo): add T5 TIP-20 prefix guard and CLI checks

stevencartavia:
Adds T5 TIP-1047 prefix guard.

- CREATE/CREATE2 to a TIP-20-prefixed address reverts before frame setup (no nonce bump, no value transfer, no init-code execution).
- EIP-7702 / Tempo AA entries with prefixed authorities are skipped, not failed (list length preserved).
- Gated on T5 via shared hardfork gate; pre-T5 Tempo and Ethereum behavior unchanged.
- Cast: `cast create2 --avoid-tip20-prefix`, `cast wallet sign-auth --check`, `cast check-auth` (alias `7702-check`).
- Integration tests cover guard side-effect boundaries, auth masking, and pre-T5 compatibility.

Closes OSS-269


crates/cast/src/cmd/create2.rs
mablr, line 161: `--avoid-tip20-prefix` only rejects impossible `--starts-with` filters, but `--matching` may also force the full reserved prefix, e.g. `--matching=20C000000000000000000000[...]`. In that case every candidate is filtered out at line 236 and mine_salt runs indefinitely. Maybe let's validate `--matching` for fixed TIP-20 prefix positions before starting the miner.

stevencartavia: [59b8e50](https://github.com/foundry-rs/foundry/pull/15040/commits/59b8e5052ca8b9964989915876a62bf1424528f2)

mablr (approved):
sgtm

grandizzy:
not in yet

PR #15041, opened 2026-06-03
fix(lint): avoid arbitrary-send-eth modifier blowups

figtracer:
## Summary
- skip expensive modifier caller-restriction analysis unless `arbitrary-send-eth` has found a function-body ETH send candidate
- memoize and cycle-guard state-variable `address(this)` alias analysis used by caller-restriction checks
- add an Ownable-style regression fixture matching the flattened SystemConfig failure shape


grandizzy (approved):

mablr (approved):
nice

PR #15046, opened 2026-06-03
feat(lint): add `controlled-delegatecall`

mablr:
## Motivation

Add `controlled-delegatecall` lint.

Closes OSS-21


crates/lint/src/sol/high/controlled_delegatecall.rs
```diff
@@ -0,0 +1,1167 @@
...
+        | ExprKind::Ident(_)
+        | ExprKind::New(_)
+        | ExprKind::TypeCall(_)
+        | ExprKind::Type(_)
+        | ExprKind::Err(_) => false,
+    }
+}
+
+/// Strips the trailing `if (...) break;` that lowers `do { ... } while (cond);`.
+fn do_while_user_stmts<'a, 'hir>(stmts: &'a [hir::Stmt<'hir>]) -> &'a [hir::Stmt<'hir>] {
```
figtracer, line 1180: this skips side effects in the lowered do-while condition. a condition like `(target = msg.sender)` can make the post-loop delegatecall untrusted but currently emits no warning.

figtracer (commented):
nits

crates/lint/src/sol/high/controlled_delegatecall.rs
```diff
@@ -0,0 +1,1167 @@
...
+                        let arity = args.len();
+                        let contract = hir.contract(cid);
+                        let bases: &[hir::ContractId] = if contract.linearization_failed() {
+                            std::slice::from_ref(&cid)
+                        } else {
+                            contract.linearized_bases
+                        };
+                        let mut fallback: Option<TypeKind<'_>> = None;
+                        for &bid in bases {
+                            for fid in hir.contract(bid).all_functions() {
```
figtracer, line 671: this overload lookup matches only name and arity, so it can use the wrong return type and flag ordinary interface `delegatecall` methods as low-level address delegatecalls.

figtracer (approved):

mattsse (approved):

PR #15047, opened 2026-06-03
feat(forge): derive invariant workers from --jobs

0xKarl98:
Towards OSS-273 

This changes invariant worker selection from a plain numeric count to an explicit mode:

- `auto` derives invariant workers from the active rayon / `--jobs` thread pool.
- positive integers keep the existing fixed-worker override behavior.
- `0` is rejected during config parsing.

Auto worker selection also accounts for the number of matching invariant contracts, so parallel test suites share the available job pool instead of allowing every invariant contract to claim all threads. Short non-timeout campaigns are capped by estimated work (`runs * depth`) to avoid creating many tiny jobs, while timeout campaigns can use the requested auto parallelism because they share a wall-clock deadline.

The actual worker count used by failing invariant campaigns is recorded and printed in the reproduction hint, so users can rerun failures with a fixed `--invariant-workers N`.

### DIFF
--- a/crates/config/src/invariant.rs
+++ b/crates/config/src/invariant.rs
@@ -8,13 +8,18 @@ use serde::{
 use std::{fmt, num::NonZeroUsize, path::PathBuf, str::FromStr};
 
 /// Worker selection mode for invariant campaign sharding.
-#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum InvariantWorkers {
     /// Automatically derive invariant workers from the active `--jobs` / rayon thread pool.
-    #[default]
     Auto,
     /// Explicit user override for invariant campaign sharding.
     Fixed(NonZeroUsize),
+}
+
+impl Default for InvariantWorkers {
+    fn default() -> Self {
+        Self::Fixed(NonZeroUsize::MIN)
+    }
 }
 
 impl Serialize for InvariantWorkers {
@@ -101,8 +106,8 @@ pub struct InvariantConfig {
     pub depth: u32,
     /// Worker selection mode used to shard invariant runs.
     ///
-    /// Defaults to `auto`, which derives the worker count from `--jobs`. Use a positive integer
-    /// to request a fixed worker count independent of `--jobs` for reproducibility.
+    /// Defaults to `1` for reproducible seeded campaigns. Use `auto` to derive the worker count
+    /// from `--jobs`, or a positive integer for an explicit worker count.
     pub workers: InvariantWorkers,
     /// Fails the invariant fuzzing if a revert occurs
     pub fail_on_revert: bool,
@@ -149,7 +154,7 @@ impl Default for InvariantConfig {
         Self {
             runs: 256,
             depth: 500,
-            workers: InvariantWorkers::Auto,
+            workers: InvariantWorkers::default(),
             fail_on_revert: false,
             call_override: false,
             dictionary: FuzzDictionaryConfig { dictionary_weight: 80, ..Default::default() },
@@ -199,6 +204,12 @@ mod tests {
             serde_json::from_str::<InvariantWorkers>(r#""4""#).unwrap(),
             InvariantWorkers::Fixed(NonZeroUsize::new(4).unwrap())
         );
+    }
+
+    #[test]
+    fn invariant_workers_default_to_one() {
+        assert_eq!(InvariantWorkers::default(), InvariantWorkers::Fixed(NonZeroUsize::MIN));
+        assert_eq!(InvariantConfig::default().workers, InvariantWorkers::Fixed(NonZeroUsize::MIN));
     }
 
     #[test]
--- a/crates/config/src/invariant.rs
+++ b/crates/config/src/invariant.rs
@@ -1,8 +1,88 @@
 //! Configuration for invariant testing
 
 use crate::fuzz::{FuzzCorpusConfig, FuzzDictionaryConfig};
-use serde::{Deserialize, Serialize};
-use std::path::PathBuf;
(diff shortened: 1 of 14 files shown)


crates/forge/src/multi_runner.rs
```diff
@@ -197,6 +193,15 @@ impl<FEN: FoundryEvmNetwork> MultiContractRunner<FEN> {
...
         );
+        let num_invariant_contracts = contracts
+            .iter()
+            .filter(|(_, contract)| {
+                contract
+                    .abi
+                    .functions()
+                    .any(|func| func.is_invariant_test() && filter.matches_test_function(func))
+            })
+            .count();
```
mablr: `num_invariant_contracts` may differ from the number of invariant campaigns running in the current pass? Multiple optimization invariants in one contract each run as separate parallel campaigns but still divide by `1`, so `auto` can let each campaign claim the full `--jobs` pool. Multi-network filtering can also overcount contracts that do not run in the pass. Maybe we should count actual runnable campaign anchors instead: merged boolean campaign anchors plus matching optimization invariant functions?

0xKarl98:
@mablr  Have addressed the count , ptal again

crates/config/src/invariant.rs
```diff
@@ -61,7 +149,7 @@ impl Default for InvariantConfig {
         Self {
             runs: 256,
             depth: 500,
-            workers: 1,
+            workers: InvariantWorkers::Auto,
```
mablr: Making the default `Auto` seems to silently change seeded invariant execution when `runs/depth/timeout` allow more than one worker, and the resolved count also depends on `--jobs` and how many invariant campaigns match. That may break reproducibility without an explicit opt-in?

crates/forge/src/result.rs
```diff
@@ -159,6 +156,14 @@ impl TestOutcome {
...
     }
 
+    fn invariant_workers(&self) -> usize {
+        self.failures()
+            .filter(|(_, result)| result.kind.is_invariant())
+            .map(|(_, result)| result.invariant_workers.max(1))
+            .max()
+            .unwrap_or(1)
+    }
+
```
mablr: The reproduction hint collapses all failing invariant campaigns to `max(workers)`. If two failures used different effective worker counts, rerunning with the single printed `--invariant-workers N` can change the worker partitioning for one of them and fail to reproduce. We should probably print per-failure worker counts, or only print the global hint when all failing invariant campaigns used the same count.

0xKarl98:
@grandizzy  wdyt

crates/forge/src/runner.rs
```diff
@@ -92,24 +350,25 @@ impl<'a, FEN: FoundryEvmNetwork> Deref for ContractRunner<'a, FEN> {
...
         mcr: &'a MultiContractRunner<FEN>,
+        num_invariant_campaign_anchors: usize,
     ) -> Self {
         Self {
             name,
             contract,
             executor,
             progress,
-            tokio_handle,
+            tokio_handle: tokio::runtime::Handle::current(),
```
mablr: Dangerous if constructor used outside Tokio runtime

crates/forge/src/cmd/test/mod.rs
```diff
@@ -201,7 +197,7 @@ pub struct TestArgs {
 
     /// Number of workers to use for invariant test campaigns.
     #[arg(long, env = "FOUNDRY_INVARIANT_WORKERS", value_name = "WORKERS")]
-    pub invariant_workers: Option<usize>,
+    pub invariant_workers: Option<InvariantWorkers>,
```
mablr: `--invariant-workers` now accepts `auto`, the help message is outdated.

mablr (changes requested):
Some findings... + some tests seem failed

crates/config/src/invariant.rs
0xKarl98: The tests failed due to run out of time , let me re-run it  in the fix

crates/forge/src/result.rs
0xKarl98: Have chosen the second as the diff would be smaller

0xKarl98:
@mablr  have addressed the test fails and your concerns , plz recheck

mablr (dismissed):
sgtm

crates/lint/src/sol/gas/external_function.rs
```diff
@@ -287,20 +287,16 @@ impl<'hir> hir::Visit<'hir> for ParamEscapeFinder<'_, 'hir> {
...
             ExprKind::Unary(op, inner)
                 if matches!(
                     op.kind,
                     UnOpKind::PreInc | UnOpKind::PreDec | UnOpKind::PostInc | UnOpKind::PostDec
                 ) =>
             {
-                if expr_root_is_param(inner, self.params) {
-                    return ControlFlow::Break(());
-                }
+                return ControlFlow::Break(());
```
stevencartavia: Maybe worth a look — the `Unary` arm refactor here looks different from the `Delete` one: the inner `expr_root_is_param(inner, self.params)` check seems to have been dropped rather than folded into the guard, so any `++`/`--` (e.g. an unrelated local counter) ends up breaking `ParamEscapeFinder` and suppressing the `external-function` suggestion.

Quick repro:

```solidity
function shouldBeExternal(bytes memory data) public {
    uint256 i = 0; i++; stored = data;
}
```

`forge lint --only-lint external-function` emits the note on master but not on this PR. Might be intentional, but if not, mirroring the `Delete` guard (`&& expr_root_is_param(inner, self.params)`) would restore the prior behavior.

0xKarl98: Have restored !

stevencartavia (approved):

mattsse (approved):
LGTM

PR #15048, opened 2026-06-03
fix(debugger): identify forked contracts in debug traces

### LINKED ISSUE #12747: Debug forked contracts at source code level
Hello,

I am not sure that if this feature is done, maybe i have done something wrong. But I have a simple demo project, which deploy two contract to a Anvil chain and i forked the Anvil chain(by `vm.createSelectFork`) and use `forge test --debug` to see if the source mapping works. But no, I still only step into opcode.

I am wondering why this does not work?

https://github.com/sdajklfw389/FoundryDebugForkedContracts

figtracer:
## Summary
- capture runtime bytecode for traced addresses from the execution database
- use captured bytecode during `forge test --debug` trace identification so forked/predeployed local artifacts get debugger source maps
- propagate the bytecode map through unit, fuzz/table, and invariant debug result paths

Closes #12747


mablr (changes requested):
Makes sense, please fix the type mismatch breaking CI :)

crates/forge/src/result.rs
```diff
@@ -881,6 +885,7 @@ macro_rules! extend {
         $a.logs.extend($b.logs);
         $a.labels.extend($b.labels);
         $a.traces.extend($b.traces.map(|traces| ($trace_kind, traces)));
+        $a.debug_bytecodes.extend($b.debug_bytecodes);
```
mablr, line 888: This now seems to merge `debug_bytecodes` for every `extend!()` caller, including `table_result()`, but table fuzz execution never populates `FuzzTestResult.debug_bytecodes` from the selected `raw_call_result` in `run_table_test`. That means table tests with `--debug --dump` still keep the trace without the fork-loaded bytecode map.

Maybe we need to copy `raw_call_result.debug_bytecodes` alongside `raw_call_result.traces` on the table success/failure paths, and add a table-test variant to the fork debug dump test ?

mablr (approved):
sgtm

stevencartavia (approved):

PR #15049, opened 2026-06-03
feat(debugger): add goto program counter

### LINKED ISSUE #7906: [p'<char>]: goto program counter in debugger
### Component

Forge

Screenshot 2024-05-10 at 3 48 56 PM

It would be great, if we could go to a program counter for smart contracts in the foundry debugger. If we type `p` perhaps a little char box pops up for us to enter the program counter to go to.

Otherwise, maybe some other buttons, but this would be a helpful feature. Right now I sort of scroll like crazy to get to breakpoints (when debugging raw assembly, since I can't do `vm.breakpoint` in something like huff)

### Additional context

_No response_

figtracer:
## Summary
- add a debugger TUI `p` prompt for jumping to a program counter
- support hex and explicit decimal PC input with ambiguity handling for bare digits
- search the current trace first, then same-code-context trace segments, without default cross-contract jumps
- show prompt/status feedback in the debugger footer and display the current PC as both hex and decimal

Closes #7906

### DIFF
--- a/crates/debugger/src/tui/context.rs
+++ b/crates/debugger/src/tui/context.rs
@@ -8,6 +8,18 @@ use revm::bytecode::opcode::OpCode;
 use revm_inspectors::tracing::types::{CallKind, CallTraceStep};
 use std::ops::ControlFlow;
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


stevencartavia (approved):

mablr (approved):
sgtm

Issue #12336, opened 2025-10-28
Support rendering `debug_traceCall` traces in `cast`


klkvr:
### Component

Cast

### Describe the feature you would like

Sometimes `cast call --trace` might be too slow and/or not account for chain-specific EVM customizations.

It would be useful if we've had something like `cast call --debug-trace-call` that would be equivalent to `cast call --trace` but would obtain the trace by performing a `debug_traceCall` request to the node

`cast run --debug-trace` could also be useful for fetching `debug_traceTransaction` outputs.

The most complex part here is likely conversion from the RPC trace to the `TraceArena` type

PR #15052, opened 2026-06-03
feat(cast): support rendering debug_traceCall traces

0xMars42:
Converts a `callTracer` response into a `CallTraceArena` so traces fetched over RPC can reuse the existing decoding and rendering, wired to `cast call --debug-trace-call`.

## Scope

A remote call-tree renderer, not a local struct-log replacement. Backed by `callTracer` with `withLog`, it preserves nested calls, value, gas, logs and revert data, but not opcode-level steps. The local-execution-only flags (`--trace`, `--debug`, `--decode-internal`, `--evm-version`) conflict with `--debug-trace-call`.

## Design decisions (after review)

- The `CallFrame -> CallTraceArena` converter stays in `cast` (`rpc_trace.rs`). `cast run` uses the prestateTracer with a local replay, so it is not a consumer of this converter; there is a single call site today. If a real second consumer lands, the converter can move to the shared traces crate then (a file relocation, no API debt), rather than pulling geth RPC types into that crate now.
- Unclassified `callTracer` errors (invalid opcode, provider quirks) keep their original message: when a frame fails with no return data, the raw `error` / `revert_reason` text is surfaced as the trace output instead of collapsing to a coarse `EvmError: Revert`.
- Precompiles are not flagged by the converter; the trace decoder already skips them by address.
- `debug_traceCall` failures map to actionable hints: `-32601` (debug namespace disabled) and missing historical/archive state are distinguished, since they need different fixes.

## Manual acceptance coverage

These callTracer response shapes vary by client and are checked manually against geth / reth / anvil rather than in unit tests:

- normal call tree with logs interleaved between child calls
- CREATE / CREATE2 frame with no `to`
- provider error with a generic code but archive/history text (missing-state hint)
- `-32601` / debug namespace disabled (method-not-found hint)

Addresses the `cast call` part of #12336.

### DIFF
--- a/crates/cast/src/cmd/call.rs
+++ b/crates/cast/src/cmd/call.rs
@@ -1,4 +1,4 @@
-use super::run::fetch_contracts_bytecode_from_trace;
+use super::run::{fetch_contracts_bytecode_from_trace, fetch_contracts_bytecode_via_rpc};
 use crate::{
     Cast,
     debug::handle_traces,
@@ -343,7 +343,21 @@ impl CallArgs {
                 call_options = call_options.with_block_overrides(block_overrides);
             }
 
-            let geth_trace = provider.debug_trace_call(tx, block, call_options).await?;
+            let geth_trace = provider
+                .debug_trace_call(tx, block, call_options)
+                .await
+                .map_err(|err| -> eyre::Report {
+                    // The `debug` namespace is often disabled on public RPCs, which reject the
+                    // call with JSON-RPC error -32601 (method not found). Surface an
+                    // actionable hint instead of the raw transport error.
+                    if err.as_error_resp().is_some_and(|resp| resp.code == -32601) {
+                        eyre::eyre!(
+                            "the RPC endpoint does not support `debug_traceCall` (method not found); use a node with the `debug` namespace enabled (e.g. a local anvil/reth or an archive endpoint), or drop `--debug-trace-call` to run the call locally with `--trace`"
+                        )
+                    } else {
+                        err.into()
+                    }
+                })?;
             let GethTrace::CallTracer(frame) = geth_trace else {
                 eyre::bail!(
                     "`debug_traceCall` did not return a callTracer frame; the RPC endpoint may not \
@@ -363,12 +377,22 @@ impl CallArgs {
                 gas_used,
             };
 
+            // Local-artifact labeling matches deployed runtime bytecode against the
+            // project artifacts. There is no local executor on this path, so fetch the code
+            // over RPC for the addresses in the trace. Skip the extra round-trips unless
+            // local artifacts were requested.
+            let contracts_bytecode = if with_local_artifacts {
+                fetch_contracts_bytecode_via_rpc(&provider, &result, block).await?
+            } else {
+                Default::default()
+            };
+
             let chain = alloy_chains::Chain::from_id(provider.get_chain_id().await?);
             handle_traces(
                 result,
                 &config,
                 chain,
-                &Default::default(),
+                &contracts_bytecode,
                 labels,
                 with_local_artifacts,
                 false,
--- a/crates/cast/src/cmd/call.rs
+++ b/crates/cast/src/cmd/call.rs
@@ -122,7 +122,7 @@ pub struct CallArgs {
     debug_trace_call: bool,
 
     /// Disables the labels in the traces.
-    /// Can only be set with `--trace`.
+    /// Can only be set with `--trace` or `--debug-trace-call`.
     #[arg(long, default_value_t = false, requires = "trace")]
     disable_labels: bool,
 
@@ -141,7 +141,7 @@ pub struct CallArgs {
     decode_internal: bool,
 
     /// Labels to apply to the traces; format: `address:label`.
-    /// Can only be used with `--trace`.
+    /// Can only be used with `--trace` or `--debug-trace-call`.
     #[arg(long, requires = "trace")]
     labels: Vec<String>,
 
@@ -619,11 +619,27 @@ impl CallArgs {
             }
         }).transpose()?;
 
-        // Build eth_call params
-        let call_object = serde_json::json!({
+        // Build eth_call params. `--curl` builds the request offline, so the fields the
+        // RPC-backed builder would resolve against the node (fee style, blob sidecars,
+        // authorization lists) are left to the node's defaults; the scalar fields given on the
+        // command line are forwarded as-is so the printed request runs the same call as the
+        // non-curl command.
+        let mut call_object = serde_json::json!({
             "to": to,
             "data": format!("0x{}", hex::encode(&data)),
         });
+        if let Some(from) = self.wallet.from {
+            call_object["from"] = serde_json::json!(from);
+        }
+        if let Some(value) = self.tx.value {
+            call_object["value"] = serde_json::json!(value);
+        }
+        if let Some(gas_limit) = self.tx.gas_limit {
+            call_object["gas"] = serde_json::json!(gas_limit);
+        }
+        if let Some(nonce) = self.tx.nonce {
+            call_object["nonce"] = serde_json::json!(nonce);
+        }
 
         let block_param = self
             .block
--- a/crates/cast/src/cmd/call.rs
+++ b/crates/cast/src/cmd/call.rs
@@ -106,6 +106,9 @@ pub struct CallArgs {
     /// Fetch the call trace from the node via `debug_traceCall` (callTracer) and render it,
     /// instead of re-executing the call locally like `--trace`.
     ///
+    /// This is a call-tree view: nested calls, value, gas, emitted logs and revert data. It does
+    /// not provide the opcode / struct-log level detail of a local `--trace` / `--debug` run.
+    ///
     /// The local-execution-only trace flags (`--debug`, `--decode-internal`, `--evm-version`) do
     /// not apply, since the trace comes from the node rather than a local run.
     #[arg(
@@ -347,12 +350,35 @@ impl CallArgs {
                 .debug_trace_call(tx, block, call_options)
                 .await
                 .map_err(|err| -> eyre::Report {
-                    // The `debug` namespace is often disabled on public RPCs, which reject the
-                    // call with JSON-RPC error -32601 (method not found). Surface an
-                    // actionable hint instead of the raw transport error.
-                    if err.as_error_resp().is_some_and(|resp| resp.code == -32601) {
+                    // Two RPC rejections deserve an actionable hint instead of the raw transport
+                    // error, and they need different fixes:
+                    // - `debug` namespace disabled: rejected with JSON-RPC -32601 (method not found).
+                    // - missing historical state: an archive-depth error, usually with a generic
+                    //   code (-32000) distinguishable only by message, hit whenever `--block`
+                    //   targets a block a full node has pruned.
+                    let is_method_not_found =
+                        err.as_error_resp().is_some_and(|resp| resp.code == -32601);
+                    let message = err
+                        .as_error_resp()
+                        .map(|resp| resp.message.to_ascii_lowercase())
+                        .unwrap_or_else(|| err.to_string().to_ascii_lowercase());
+                    let is_missing_state = [
+                        "missing trie node",
+                        "required historical state",
+                        "historical state",
+                        "header not found",
+                        "missing state",
+                    ]
+                    .iter()
+                    .any(|needle| message.contains(*needle));
+
+                    if is_method_not_found {
                         eyre::eyre!(
                             "the RPC endpoint does not support `debug_traceCall` (method not found); use a node with the `debug` namespace enabled (e.g. a local anvil/reth or an archive endpoint), or drop `--debug-trace-call` to run the call locally with `--trace`"
+                        )
+                    } else if is_missing_state {
+                        eyre::eyre!(
+                            "the RPC endpoint does not have the historical state for the requested block; use an archive endpoint, or target a more recent block with `--block`"
                         )
                     } else {
                         err.into()
--- a/crates/cast/src/cmd/call.rs
+++ b/crates/cast/src/cmd/call.rs
@@ -2,16 +2,21 @@ use super::run::fetch_contracts_bytecode_from_trace;
 use crate::{
     Cast,
     debug::handle_traces,
+    rpc_trace::call_frame_to_arena,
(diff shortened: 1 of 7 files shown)


crates/cast/src/cmd/call.rs
```diff
@@ -394,6 +412,60 @@ impl CallArgs {
...
+                traces: Some(vec![(TraceKind::Execution, arena)]),
+                gas_used,
+            };
+
+            let chain = alloy_chains::Chain::from_id(provider.get_chain_id().await?);
+            handle_traces(
+                result,
+                &config,
+                chain,
+                &Default::default(),
```
figtracer: this passes an empty bytecode map into `handle_traces`, so `--debug-trace-call --la` cannot label normal call frames from local artifacts. i reproduced local artifact labels with `--trace --la`, but `--debug-trace-call --la` falls back to the raw address. can we fetch bytecode for the trace addresses and pass that map here?

0xMars42: Thanks for the review. Fixed in 47b5e7f: when --with-local-artifacts is set I now collect the trace addresses (the same gather_trace_addresses the --trace path uses), fetch their runtime bytecode over RPC, and pass that map into handle_traces instead of the empty default. --debug-trace-call --la now labels call frames from local artifacts instead of falling back to the raw address. Added a regression test asserting the Counter:: label.

I also mapped the -32601 from nodes without the debug namespace to an actionable hint; that was not in your comment, happy to split it out if you prefer to keep the PR tight.

0xMars42:
Addressed most of this in a new commit; two I'd push back on with reasoning.

1. Docs boundary: the flag help now states this is a call-tree view (nested calls, value, gas, logs, revert data) and that it does not provide the opcode/struct-log detail of a local --trace/--debug run. No user-facing docs live in this repo (that's foundry-book), so the flag doc-comment is the surface.

2. Local-flag separation: --trace/--debug/--decode-internal/--evm-version already conflict with --debug-trace-call via conflicts_with_all, with tests. No change.

3. Preserve unclassified error text: fixed. The converter dropped frame.error when output was empty, so an invalid opcode collapsed to "EvmError: Revert". push_frame now surfaces revert_reason (preferred) or the raw error string as the output when a frame failed with no return data, so the renderer shows the original message. Added a unit test.

4. Client/provider variance: treating this as manual acceptance coverage rather than unit tests, since it needs real geth/reth/anvil responses. On the converter itself: a CREATE/CREATE2 frame with no `to` is handled (address falls back to zero, kind comes from `typ` not `to`, so it still renders as a create), and precompiles are skipped by the trace decoder by address regardless of the converter, so no converter change there.

5. Log/call interleaving: added a test with two children and a log between them, asserting [Call(0), Log(0), Call(1)]. The existing tests only had a single child (before-first/after-last), so this covers the off-by-one.

6. Error hints: added a distinct hint for missing historical/archive state (matched on the message, since those come back with a generic code rather than -32601), pointing at an archive endpoint or a more recent --block, separate from the -32601 debug-namespace hint.

7. Moving the converter to the shared traces crate: I'd defer. cast run doesn't use this converter (it goes through the prestateTracer and replays locally), so there's a single consumer today. Moving now would also pull an alloy-rpc-types/geth dependency into the traces crate, which currently has none. The later move is a file relocation with no API debt, so I'd rather do it when a real second consumer lands.

0xMars42:
Added a Manual acceptance coverage section to the PR description with those four response shapes, and refreshed the rest to reflect the current state (converter stays in cast, raw error text preserved, out of draft).

crates/cast/src/cmd/call.rs
```diff
@@ -98,6 +103,21 @@ pub struct CallArgs {
...
+    /// not provide the opcode / struct-log level detail of a local `--trace` / `--debug` run.
+    ///
+    /// The local-execution-only trace flags (`--debug`, `--decode-internal`, `--evm-version`) do
+    /// not apply, since the trace comes from the node rather than a local run.
+    #[arg(
+        long = "debug-trace-call",
+        default_value_t = false,
+        conflicts_with_all = ["trace", "debug", "decode_internal", "evm_version"]
+    )]
+    debug_trace_call: bool,
```
mablr, line 122: `--debug-trace-call` can still be combined with `--curl`, but `run()` handles curl before this mode and `run_curl()` always emits an `eth_call`. I reproduced `cast call ... --debug-trace-call --curl` printing an `eth_call` curl command, so users get a request for a normal call instead of the promised trace. Consider either generating `debug_traceCall` curl payloads here, or making `--debug-trace-call` conflict with `--curl` and adding a parser test.

crates/cast/src/cmd/call.rs
```diff
@@ -311,6 +331,107 @@ impl CallArgs {
...
+                traces: Some(vec![(TraceKind::Execution, arena)]),
+                gas_used,
+            };
+
+            // Local-artifact labeling matches deployed runtime bytecode against the
+            // project artifacts. There is no local executor on this path, so fetch the code
+            // over RPC for the addresses in the trace. Skip the extra round-trips unless
+            // local artifacts were requested.
+            let contracts_bytecode = if with_local_artifacts {
+                fetch_contracts_bytecode_via_rpc(&provider, &result, block).await?
```
mablr: This bytecode map only comes from `eth_getCode`, so `--debug-trace-call --with-local-artifacts --override-code ...` cannot label a contract that exists only through the state override: the trace runs the override code, but artifact matching sees empty/original chain code. Consider seeding `contracts_bytecode` from parsed `--override-code` entries before falling back to RPC bytecode, with a regression test for `--override-code` plus `--with-local-artifacts`.

crates/cast/tests/cli/main.rs
```diff
@@ -4185,6 +4185,215 @@ Transaction successfully executed.
...
+        "--with-local-artifacts",
+        "--rpc-url",
+        &handle.http_endpoint(),
+    ])
+    .assert_success()
+    .stdout_eq(str![[r#"
+Compiling project to generate artifacts
+[COMPILING_FILES] with [SOLC_VERSION]
+[SOLC_VERSION] [ELAPSED]
+Compiler run successful!
```
mablr, line 4382: This new `--debug-trace-call --with-local-artifacts` snapshot enshrines compile/progress prose on stdout. Since the same path is used under `--json`, stdout can become unparsable before `print_traces()` gets a chance to suppress prose. Consider moving the compile banner/progress to stderr and adding a `--json --debug-trace-call --with-local-artifacts` assertion that stdout parses cleanly.

mablr (changes requested):
Some nits

crates/cast/src/cmd/call.rs
0xMars42: Good catch, fixed in 197e330d: `run_curl()` now targets `debug_traceCall` when the flag is set, with the same tracer options as the non-curl path (`{"tracer": "callTracer", "tracerConfig": {"withLog": true}}`), and `eth_call` otherwise. `curl_call_debug_trace_call` asserts the emitted command contains `debug_traceCall` and `callTracer` and no `eth_call`.

crates/cast/src/cmd/call.rs
0xMars42: Fixed in 197e330d: with `--with-local-artifacts`, the code entries of the parsed state overrides are collected before the overrides are handed to `debug_traceCall`, and they overwrite the `eth_getCode` results in the bytecode map, since the trace ran the override code rather than the on-chain code. `cast_call_debug_trace_call_override_code_local_artifacts` overrides an empty address with Counter's runtime code and asserts the trace is labeled `Counter::number()`.

crates/cast/tests/cli/main.rs
0xMars42: Fixed in 197e330d: the compile banner now goes through `sh_status!` (stderr) instead of stdout. The compiler progress and summary were already kept off stdout under `--json` (`with_compilation_reporter` gates on `quiet || shell::is_json()`, the summary on `!shell::is_json()`), so the banner was the remaining polluter. `cast_call_debug_trace_call_local_artifacts_json_stdout` asserts that stdout of `--json --debug-trace-call --with-local-artifacts` parses as a single JSON document, and the three snapshots that pinned the banner on stdout are updated.

crates/cast/src/cmd/call.rs
mablr: Well `--debug-trace-call --curl` now switches the method to `debug_traceCall`, but the generated third param only includes the tracer config. The non-curl path also forwards `stateOverrides` and `blockOverrides`, so `cast call ... --debug-trace-call --curl --override-code ... --block.number ...` prints a request that would trace different state from the command it represents.

You should probably build the curl options from the same `GethDebugTracingCallOptions` shape and adding coverage for an override flag. This is not very important, so feel free to reject `--curl` support becomes too cursed.

crates/cast/src/rpc_trace.rs
```diff
@@ -0,0 +1,377 @@
...
+        && let Some(text) = frame.revert_reason.as_deref().or(frame.error.as_deref())
+    {
+        output = Bytes::copy_from_slice(text.as_bytes());
+    }
+
+    let trace = CallTrace {
+        depth,
+        success,
+        caller: frame.from,
+        address: frame.to.unwrap_or_default(),
```
mablr: SELFDESTRUCT callTracer frames are currently converted as ordinary calls to `frame.to`, with all `selfdestruct_*` fields left empty. In the local trace representation, selfdestructs are detected through those fields/status, and geth call traces encode them as `from = destructed contract`, `to = refund target`, `value = transferred balance`. As-is, a SELFDESTRUCT frame renders and identifies the beneficiary as a normal callee instead of preserving the destructed contract and refund target. Consider special-casing `frame.typ == "SELFDESTRUCT"` here: set `address`/`selfdestruct_address` from `frame.from`, `selfdestruct_refund_target` from `frame.to`, `selfdestruct_transferred_value` from `frame.value`, and use `InstructionResult::SelfDestruct`.

crates/cast/src/cmd/call.rs
0xMars42: Right, fixed in 36060239: `run_curl()` now builds the third param from the same `GethDebugTracingCallOptions` shape as the non-curl path (tracer options plus `stateOverrides` / `blockOverrides` when set), so the printed request traces the same state. `curl_call_debug_trace_call_forwards_overrides` covers `--override-code` plus `--block.number` and asserts both override objects and the override code appear in the request.

crates/cast/src/rpc_trace.rs
0xMars42: Fixed in 54d8e444. `push_frame` now special-cases `typ == "SELFDESTRUCT"`: `address` and `selfdestruct_address` come from `from` (the destructed contract), `selfdestruct_refund_target` from `to`, `selfdestruct_transferred_value` from `value`, and the status is `InstructionResult::SelfDestruct`, mirroring the local representation, so `is_selfdestruct()` holds and the destructed contract (not the beneficiary) is identified. I confirmed against `geth_selfdestruct_call_trace` that this is the exact inverse of how geth encodes the frame.

Two notes on scope:

- I put the transferred balance only in `selfdestruct_transferred_value` and left the call `value` at zero, since the frame's `value` is the destructed balance rather than a call value; happy to also surface it as the call value if you prefer.
- the frame becomes its own node (as geth sends it), not an annotation on the parent, so the render shows a `[SelfDestruct]` line for the destructed contract. The `TraceWriter` has no dedicated selfdestruct line, so the header still uses the generic call format; a prettier rendering would be a separate `revm-inspectors` writer change.

I kept the test as a unit test on a synthetic frame (like the others in this file) rather than end-to-end: post-Cancun (EIP-6780) a selfdestruct only actually deletes in its creation transaction, so an anvil round-trip would not reliably emit the frame.

crates/cast/src/cmd/call.rs
mablr: It can still print a trace request for a different call than the non-curl command. The normal path builds the full transaction via `CastTxBuilder`, but this curl path only emits `to` and `data`, so flags like `--value 1ether` and `--gas-limit 12345` are dropped from the `debug_traceCall` payload. We would need to serialize the same supported transaction fields here, with coverage for `--debug-trace-call --curl --value ... --gas-limit ...`. I really think it's a lot of effort for a little benefit.

crates/cast/src/cmd/call.rs
```diff
@@ -98,6 +106,21 @@ pub struct CallArgs {
...
+    /// The local-execution-only trace flags (`--debug`, `--decode-internal`, `--evm-version`) do
+    /// not apply, since the trace comes from the node rather than a local run.
+    #[arg(
+        long = "debug-trace-call",
+        default_value_t = false,
+        conflicts_with_all = ["trace", "debug", "decode_internal", "evm_version"]
+    )]
+    debug_trace_call: bool,
+
     /// Disables the labels in the traces.
```
mablr, line 124: These help comments still say `--labels` / `--disable-labels` can only be used with `--trace`, but the debug trace path also passes them into `handle_traces`. Consider updating the wording to say trace rendering, or explicitly mention both `--trace` and `--debug-trace-call`.

crates/cast/src/cmd/call.rs
0xMars42: Agreed on not rebuilding the full transaction here: the fee style, blob sidecars and authorization lists are resolved against the node by the tx builder, and `--curl` is offline by design. The scalar fields are cheap though, so 65e3785b forwards `from`, `value`, `gas` and `nonce` into the call object as given on the command line, with `curl_call_debug_trace_call_forwards_tx_fields` covering `--value 1ether --gas-limit 12345`. The call object is shared with the plain `eth_call` arm, which dropped the same flags, so both requests are fixed.

crates/cast/src/cmd/call.rs
0xMars42: Updated in 65e3785b, both flags now say `--trace` or `--debug-trace-call`. I checked the clap side while at it: `--labels --debug-trace-call` is accepted (the conflict between `--debug-trace-call` and `--trace` lifts the `requires = "trace"` constraint, conflicts take precedence over requires in clap) and `--labels` alone is still rejected, so the wording matches the actual CLI behavior. Added `call_labels_accepted_with_debug_trace_call` and `call_labels_rejected_without_trace_mode` to lock that in.

mablr (approved):
sgtm

mattsse (approved):
lgtm

PR #15053, opened 2026-06-03
feat(mpp): wire T5 precompile escrow

stevencartavia:
Wires Foundry's MPP session transport to the T5 TIP-1034 channel reserve precompile while preserving legacy Solidity escrow compatibility.

`pay_session` discriminates on `is_precompile_escrow(escrow)` per challenge.
Precompile open: single-call `ITIP20ChannelReserve.open` (no `approve`, TIP-1035), `uint96` caps, `expiringNonceHash` derived pre-sign, voucher signed via `sign_precompile_voucher`.
Cached vouchers and channel cache (`channel_key` now includes operator) branch the same way.
Precompile top-up hard-fails (descriptor-based path not implemented); legacy top-up gets `checked_add` overflow guard.
Optional `methodDetails.operator`: missing → ZERO; malformed → error.

https://github.com/tempoxyz/mpp-rs/pull/276

Part of OSS-266

### DIFF
--- a/crates/common/src/provider/mpp/persist.rs
+++ b/crates/common/src/provider/mpp/persist.rs
@@ -7,7 +7,7 @@
 
 use alloy_primitives::{Address, B256};
 use foundry_wallets::{Channel, ChannelDb};
-use mpp::client::channel_ops::ChannelEntry;
+use mpp::client::channel_ops::{ChannelEntry, is_precompile_escrow};
 use std::{
     collections::HashMap,
     sync::OnceLock,
@@ -67,6 +67,12 @@ fn channel_key_from_persisted(ch: &Channel) -> String {
         Address::ZERO,
     )
     .to_lowercase()
+}
+
+/// Precompile channels aren't persisted: the `Channel` schema has no column for
+/// their `operator`, so they'd reload under a wrong `operator=ZERO` key.
+fn is_precompile_channel(ch: &Channel) -> bool {
+    matches!(ch.escrow_contract.parse::<Address>(), Ok(addr) if is_precompile_escrow(addr))
 }
 
 /// Whether a channel can still be used (active and not fully spent).
@@ -147,7 +153,7 @@ pub fn load_channels() -> HashMap<String, Channel> {
 
     let usable: HashMap<String, Channel> = channels
         .into_iter()
-        .filter(is_usable)
+        .filter(|ch| is_usable(ch) && !is_precompile_channel(ch))
         .map(|ch| {
             let key = channel_key_from_persisted(&ch);
             (key, ch)
@@ -164,12 +170,15 @@ pub fn save_channels(channels: &HashMap<String, Channel>) {
         return;
     };
 
-    for ch in channels.values() {
+    let mut saved = 0;
+    for ch in channels.values().filter(|ch| !is_precompile_channel(ch)) {
         if let Err(e) = db.upsert(ch) {
             warn!(%e, channel_id = %ch.channel_id, "failed to save channel");
+        } else {
+            saved += 1;
         }
     }
-    debug!(count = channels.len(), "saved MPP channels");
+    debug!(count = saved, "saved MPP channels");
 }
 
 /// Delete a channel from the database by its channel ID.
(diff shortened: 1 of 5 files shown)


crates/common/src/provider/mpp/session.rs
```diff
@@ -304,12 +313,14 @@ impl SessionProvider {
...
         chain_id: u64,
+        operator: Address,
     ) -> String {
         // Use first 8 bytes of origin hash to scope the key without persisting
         // the full URL (which may contain secrets in query params).
+        // `operator` scopes precompile channels (legacy passes ZERO).
         let origin_hash = &alloy_primitives::keccak256(origin.as_bytes()).to_string()[..18];
         let signer = authorized_signer.unwrap_or(*payer);
-        format!("{origin_hash}:{chain_id}:{payer}:{signer}:{payee}:{currency}:{escrow}")
+        format!("{origin_hash}:{chain_id}:{payer}:{signer}:{payee}:{currency}:{escrow}:{operator}")
```
mablr, line 323: This changes the channel cache key shape by appending operator, but persisted channel reload still reconstructs keys independently in `persist::channel_key_from_persisted()` using the old 7-field format. After a restart, restored channels are inserted under the old key while `pay_session()` looks up the new key, so Foundry will miss persisted channels and open duplicates.

stevencartavia: [b631b8a](https://github.com/foundry-rs/foundry/pull/15053/commits/b631b8afb5bf05bd18f04e39329008c809e9b503)

crates/common/src/provider/mpp/persist.rs
```diff
@@ -50,18 +50,21 @@ fn global_db() -> Option<&'static ChannelDb> {
 
 /// Reconstruct the composite HashMap key from a persisted `Channel`.
 ///
-/// Mirrors `SessionProvider::channel_key()` in session.rs.
+/// Mirrors `SessionProvider::channel_key()` in session.rs. The persisted
+/// schema has no `operator` column, so we always emit `Address::ZERO` as the
+/// operator component — correct for legacy escrow rows.
 fn channel_key_from_persisted(ch: &Channel) -> String {
```
mablr, line 56: I think this still loses non-zero precompile operators across restarts. New precompile channels are persisted through `from_channel_entry()` at `session.rs:910`, but the `Channel` schema has no operator field and reload always reconstructs the key with `Address::ZERO`. For any T5 challenge with `methodDetails.operator != ZERO`, the restored row is inserted under the wrong key, `pay_session()` misses it, and Foundry opens duplicate channels after restart

crates/common/src/provider/mpp/session.rs
```diff
@@ -600,6 +730,21 @@ impl SessionProvider {
             .map_err(|_e| MppError::InvalidConfig("invalid currency address".to_string()))?;
         let amount: u128 = session_req.parse_amount()?;
 
+        // Precompile-only operator from methodDetails; missing → ZERO,
+        // malformed → error. Parsed before the cache lookup since operator
+        // participates in `channel_key`.
+        let operator = match session_req.method_details.as_ref().and_then(|v| v.get("operator")) {
+            None => Address::ZERO,
+            Some(v) => {
```
mablr: `operator` is described as precompile-only, but it is parsed and included in the cache key for every session challenge. If a legacy escrow challenge includes `methodDetails.operator`, the open/voucher logic ignores it while cache lookup and persistence do not, so legacy channels can be fragmented and reloaded under `operator=ZERO`. Consider forcing `operator = Address::ZERO` unless `is_precompile_escrow(escrow_contract)`.

crates/common/src/provider/mpp/persist.rs
stevencartavia: [78b8016](https://github.com/foundry-rs/foundry/pull/15053/commits/78b8016d7bfec3ccc37115c817cc8ec8742de2ae)

crates/common/src/provider/mpp/session.rs
stevencartavia: [78b8016](https://github.com/foundry-rs/foundry/pull/15053/commits/78b8016d7bfec3ccc37115c817cc8ec8742de2ae)

mablr (approved):
sgtm

mattsse (approved):
lgtm

PR #15059, opened 2026-06-04
test(forge): expose invariant workers in json kind

0xKarl98:
Expose the invariant worker count as part of the `TestKind::Invariant` JSON payload:

```json
  "kind": {
    "Invariant": {
      "runs": 4,
      "calls": 8,
      "reverts": 0,
      "workers": 4
    }
}
```
This keeps the field scoped to invariant campaign metadata instead of adding an invariant-specific field to the top-level
TestResult.

The default single-worker case is also emitted as workers: 1, so JSON consumers do not need to treat a missing field as an implicit default. Old serialized results that do not contain workers still deserialize with a default of 1 as fallback .

### DIFF
--- a/crates/forge/tests/cli/test_cmd/invariant/mod.rs
+++ b/crates/forge/tests/cli/test_cmd/invariant/mod.rs
@@ -1211,7 +1211,7 @@ Ran 2 tests for test/OptimizationCorpusTest.t.sol:OptimizationCorpusTest
 
 forgetest_init!(json_reports_invariant_predicate_results, |prj, cmd| {
     prj.update_config(|config| {
-        config.invariant.runs = 1;
+        config.invariant.runs = 4;
         config.invariant.depth = 2;
     });
     prj.add_test(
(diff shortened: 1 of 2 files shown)


crates/forge/tests/cli/test_cmd/invariant/mod.rs
```diff
@@ -1250,6 +1250,24 @@ contract JsonInvariantReportTest is Test {
     let tests = suite["test_results"].as_object().unwrap();
     assert_eq!(tests.len(), 1);
     let result = tests.values().next().unwrap();
+    let kind = &result["kind"]["Invariant"];
```
mablr, line 1253: This covers the new serialized JSON shape, but the PR also promises that old serialized invariant results without workers still deserialize with `workers = 1`. A small `serde_json::from_value::<TestKind>(...)` regression test for an Invariant payload missing workers would lock in that compatibility path.

mattsse (dismissed):

0xKarl98:
@mablr  Have added the test case you mentioned

mablr (approved):

figtracer (approved):

PR #15037, opened 2026-06-02
feat(lint): `arbitrary-send-erc20-permit`

stevencartavia:
Detects `transferFrom` calls with an arbitrary `from` after a `permit(owner, address(this), …)` on the same (`token`, `owner`). 

Closes OSS-40

PR #15060, opened 2026-06-04
refactor(lint): finish solar expression type cleanup

figtracer:
Follow-up to #15003/#15035 after auditing current `master`, including the later `arbitrary-send-erc20-permit` lint from #15037.

This finishes the Solar expression type cleanup in the remaining lints that were still reconstructing expression types from HIR:

- `controlled-delegatecall`: remove the local expression type resolver and use `Gcx::type_of_expr` for delegatecall receiver address typing; thread `Gcx` through modifier-prefix analysis.
- `arbitrary-send-erc20[-permit]`: thread `Gcx` through the analyzer, modifier/constructor seeding, permit/flash-loan/sink matching, and use Solar expression types for contract/address receiver checks. The existing identity/alias tracking and declaration ABI checks remain HIR-based.

### DIFF
--- a/crates/lint/src/sol/high/arbitrary_send_erc20.rs
+++ b/crates/lint/src/sol/high/arbitrary_send_erc20.rs
@@ -6,9 +6,13 @@ use crate::{
 use solar::{
     ast,
     interface::{Span, data_structures::Never, kw, sym},
-    sema::hir::{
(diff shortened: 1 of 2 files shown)


mablr (approved):
Nice, indeed these lints were done before the refactor landed.

stevencartavia (approved):

PR #15062, opened 2026-06-04
fix(forge): propagate fail-fast on setUp / suite-validation failures

grandizzy:
## Summary

When a contract's `setUp()` reverts (or any other early suite-level failure path in `ContractRunner::run` is hit), the suite returns a failing `SuiteResult` without ever calling `early_exit.record_failure()`. Sibling suites already dispatched in parallel via `par_iter().for_each` — most importantly long-running invariant campaigns — therefore never observe `should_stop() == true` and keep running until their configured `invariant.timeout` (potentially hours), even with `--fail-fast`.

## Repro

`forge test --fail-fast` on a workspace where one suite has a reverting `setUp()` and another has a time-based invariant campaign (`invariant.timeout = 18000`). The failure is surfaced within ms but the workflow stays alive for the full 5 h timeout because the in-flight invariant workers never receive the cancellation signal.

## Fix

Trip the existing global `EarlyExit` flag on every early-return failure branch in `ContractRunner::run`:

- multiple `setUp` functions
- multiple `afterInvariant` functions
- invariant function with parameters
- `setUp()` / constructor revert (the real-world case)
- `testFail*` rejection

`EarlyExit::record_failure` is already a no-op when fail-fast is disabled, so this is safe to call unconditionally and changes nothing for non-`--fail-fast` runs.

`InvariantCampaignState::should_stop` already checks the early-exit flag at every run boundary, so in-flight workers exit at their next fuzz sequence (≤ `depth` calls) instead of running to `invariant.timeout`.

## Test

Added `fail_fast_cancels_invariant_after_setup_failure` (mirrors the canary pattern of `test_fail_fast_config`):

- Suite A: `setUp()` reverts immediately
- Suite B: long-running invariant test with `invariant.timeout = 3600`
- Run `forge test --fail-fast --mc 'BrokenSetupTest|LongInvariantTest'`
- Assert wall-clock elapsed `< 60s`

Verified locally:

| | with fix | without fix |
|---|---|---|
| elapsed | ~5 s ✓ | ~60 s+ (asserts) |

Existing fail-fast tests (`test_fail_fast_config`, `exit_code_error_on_fail_fast`, `exit_code_error_on_fail_fast_with_json`, `show_progress_with_fail_fast_exits_early`) still pass.

### DIFF
--- a/crates/forge/src/runner.rs
+++ b/crates/forge/src/runner.rs
@@ -587,6 +587,10 @@ impl<'a, FEN: FoundryEvmNetwork> ContractRunner<'a, FEN> {
 
         // There are multiple setUp function, so we return a single test result for `setUp`
         if setup_fns.len() > 1 {
+            // Trip the global fail-fast flag so sibling parallel suites (notably long-running
+            // invariant campaigns) observe `should_stop()` and exit at their next run boundary
+            // instead of running to their timeout.
+            self.tcfg.early_exit.record_failure();
             return SuiteResult::new(
                 start.elapsed(),
                 [("setUp()".to_string(), TestResult::fail("multiple setUp functions".to_string()))]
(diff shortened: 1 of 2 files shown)


mablr (approved):
lgtm

0xKarl98 (approved):
LGTM

PR #15063, opened 2026-06-04
fix(lint): bound reentrancy helper analysis

figtracer:
## Summary

- Add a bounded shared helper-analysis cache for lints that inline internal helper calls.
- Use it in `reentrancy-events` and `reentrancy-{eth,no-eth}` so shared helper graphs are summarized instead of repeatedly rescanned.
- Keep `reentrancy-events` helper summaries reusable across recursion by replacing cut recursive edges with a conservative cached external-call reachability summary, including modifier arguments.
- Keep high/medium reentrancy helper summaries stack-aware enough to avoid reusing recursive-context summaries from a clean path, while bounding helper-heavy recursion with cached direct-call cutoff keys.
- Keep the `super.transfer(...)` typeless-builtin regression covered through `reentrancy-events` UI fixtures.
- Update Solar to `da6b0c3a8c0edd23f8b1414c61418ba78d6c4670`, which fixes the mapping-to-fixed-array typecheck panic seen before OpenZeppelin v5 linting could run.
- Add focused UI coverage for helper-heavy reentrancy completion, retained diagnostics, and the Solar mapping-array panic shape.

| External repo | Failure before | Fix / result |
|---|---|---|
| Solady | `reentrancy-events` abort; all-lints could not complete | completes; warm all-lints `189.7 ms ± 4.4 ms` |
| tempo forge-std | reentrancy timeout >120s | completes; warm all-lints `91.7 ms ± 0.7 ms` |
| Optimism vendored Solady | reentrancy timeout >120s | completes; warm all-lints `2629.8 ms ± 6.8 ms` |
| Optimism lib-keccak vendored Solady | reentrancy timeout >120s | completes; warm all-lints `2728.4 ms ± 18.4 ms` |
| OpenZeppelin v5 | Solar typeck panic | fixed by Solar `da6b0c3`; warm all-lints `153.6 ms ± 1.7 ms` |

### DIFF
--- a/crates/lint/src/sol/high/reentrancy.rs
+++ b/crates/lint/src/sol/high/reentrancy.rs
@@ -128,6 +128,9 @@ struct Analyzer<'ctx, 's, 'c, 'hir> {
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 struct InlineCallKey {
     func_id: FunctionId,
+    /// Distinguishes top-level helper summaries from summaries computed while a caller is
+    /// on the recursion stack, without keying on the full stack in helper-heavy graphs.
+    caller: Option<FunctionId>,
     state: FlowState,
 }
 
@@ -425,7 +428,11 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
         let func = self.hir.function(func_id);
         let Some(body) = func.body else { return };
 
-        let key = InlineCallKey { func_id, state: state.clone() };
+        let key = InlineCallKey {
+            func_id,
+            caller: self.call_stack.last().copied(),
+            state: state.clone(),
+        };
         if self.inline_cache.is_in_progress(&key) {
             return;
         }
--- a/crates/lint/src/sol/high/reentrancy.rs
+++ b/crates/lint/src/sol/high/reentrancy.rs
@@ -1,7 +1,7 @@
 use super::ReentrancyEth;
 use crate::{
     linter::{LateLintPass, LintContext},
-    sol::{Severity, SolLint},
+    sol::{Severity, SolLint, analysis::helper_cache::HelperAnalysisCache},
 };
 use solar::{
     ast::{
@@ -18,6 +18,8 @@ use solar::{
     },
 };
 use std::collections::{BTreeSet, HashSet};
+
+const INLINE_HELPER_CACHE_LIMIT: usize = 4096;
 
 declare_forge_lint!(
     REENTRANCY_ETH,
@@ -69,20 +71,20 @@ fn is_entry_point(func: &hir::Function<'_>) -> bool {
     func.kind.is_function() && matches!(func.visibility, Visibility::Public | Visibility::External)
 }
 
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash)]
 struct FlowState {
     state_reads: BTreeSet<VariableId>,
     pending_calls: Vec<PendingCall>,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
 struct PendingCall {
     span: Span,
     kind: ReentrantCallKind,
     state_reads: BTreeSet<VariableId>,
 }
 
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 enum ReentrantCallKind {
     Eth,
     NoEth,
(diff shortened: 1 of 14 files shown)


mablr (dismissed):

mablr (dismissed):

crates/lint/src/sol/low/reentrancy_events.rs
```diff
@@ -113,6 +115,13 @@ const fn merge_opt(dst: &mut Option<FlowState>, src: Option<FlowState>) {
     }
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+struct InlineCallKey {
```
stevencartavia, line 120: `InlineCallKey` doesn't include `call_stack`, but the recursion cutoff (`call_stack.contains(&func_id)`) does depend on it. So a cached summary computed with a recursive edge skipped gets reused from a clean stack, dropping the warning.

Repro — `master` warns, this branch is silent:

```solidity
contract R {
    event Tick();
    ICall ext;
    bool once;

    function entry(bool flag) external {
        once = false;
        if (flag) { helperA(); return; }
        noOp();
        emit Tick(); // dropped here
    }
    function noOp() internal { if (!once) { once = true; helperA(); } }
    function helperA() internal { noOp(); ext.poke(); }
}
```

Fix: add `call_stack` to `InlineCallKey` (same for `reentrancy.rs`).

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -118,10 +120,20 @@ struct Analyzer<'ctx, 's, 'c, 'hir> {
...
     reentrancy_eth_enabled: bool,
     reentrancy_no_eth_enabled: bool,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+struct InlineCallKey {
+    func_id: FunctionId,
+    /// Distinguishes top-level helper summaries from summaries computed while a caller is
+    /// on the recursion stack, without keying on the full stack in helper-heavy graphs.
+    caller: Option<FunctionId>,
```
stevencartavia: `caller` (immediate caller only) isn't enough — the cutoff uses the whole stack (`call_stack.contains`), so calls with the same immediate caller but a different deeper recursion target collide and reuse a truncated summary.

Repro — `master` warns, this branch is silent:

```solidity
contract Z {
    uint256 x;
    function entry(bool b) public {
        if (b) { a(1); return; }
        c(1);
        x = 1; // dropped
    }
    function a(uint256 d) internal {
        if (d > 0) c(d - 1);
        uint256 y = x;
        (bool ok,) = msg.sender.call{value: y}(""); require(ok);
    }
    function c(uint256 d) internal { h(d); } // h's caller is always c
    function h(uint256 d) internal { a(d); } // a is the recursion target
}
```

Fix: key on the full `call_stack`.

crates/lint/src/sol/low/reentrancy_events.rs
```diff
@@ -546,6 +590,268 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
+            return (*cached, false);
+        }
+        if !seen.insert(func_id) {
+            return (false, true);
+        }
+
+        let func = self.hir.function(func_id);
+        let mut may_reach = false;
+        let mut cut_recursive_edge = false;
+        for modifier in func.modifiers {
```
stevencartavia, line 614: This walks modifier bodies but not modifier args, while the main analyzer taints from args — so an external call in a modifier arg gets missed and a clean summary is cached.

Repro — `master` warns, this branch is silent:

```solidity
contract M {
    event Tick();
    bool flag;
    function ext() external returns (bool) { flag = true; return true; }
    modifier m(bool ok) { _; }
    function entry(uint256 d, bool b) external {
        if (b) { f(d); return; }
        c(d);
        emit Tick(); // dropped
    }
    function f(uint256 d) internal { if (d > 0) c(d - 1); hh(); }
    function hh() internal m(this.ext()) {} // external call in modifier arg
    function c(uint256 d) internal { h(d); }
    function h(uint256 d) internal { f(d); }
}
```

Moving `this.ext()` into `hh`'s body makes it warn again. Fix: also scan `modifier.args`.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -415,9 +438,219 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
         self.call_stack.push(func_id);
-        self.analyze_callable(func, body, state);
+        self.analyze_callable(func, body, &mut after);
         self.call_stack.pop();
+
+        self.inline_cache.finish(key, after.clone());
+        *state = after;
+    }
+
+    fn first_recursive_cut(&mut self, func_id: FunctionId) -> Option<FunctionId> {
```
stevencartavia, line 464: Could we do this in a follow-up so this PR doesn't get more bloated? wdyt?

`recursive_cut` keys on only the *first* on-stack cut, but a helper reaching two on-stack recursion targets cuts both — so two contexts with different cut sets collide and reuse a truncated summary.

Repro — `master` warns, current branch silent:

```solidity
contract FWarm {
    uint256 x;
    function a(bool flag, uint256 d) internal {
        if (flag) { b(d); return; }   // warm-up: f under stack {a,b}, cuts both
        f(d);
        x = 1;                         // dropped
    }
    function b(uint256 d) internal {
        f(d);
        uint256 y = x;
        (bool ok,) = payable(msg.sender).call{value: y}(""); require(ok);
    }
    function f(uint256 d) internal { if (d > 0) { a(false, d - 1); b(d - 1); } }
    function entry(bool flag, uint256 d) public { a(flag, d); }
}
```

Low likelihood. Fix: key on the full `call_stack` (and drop the `recursive_cut` machinery).

mattsse (approved):

stevencartavia (approved):

PR #15065, opened 2026-06-04
perf(symbolic): improve solver query handling

figtracer:
## Summary

- adds small native symbolic query handling candidates: exact direct-contradiction `unsat` short-circuiting before SMT, cached `unsat` subset reuse, validated hard-arithmetic model fallback/caching, constant `BoolExpr` comparison folding, SMT/cache preallocation, normalized constraint batch flattening/deduplication, and repeated `AND` mask simplification
- preserves solver correctness constraints: hard-arithmetic `is_sat` fallback only runs after solver `unknown`, hard-arithmetic model fallback returns a model validated against the original constraints, direct-contradiction handling only returns `unsat` for normalized `false` or exact `p` / `not p` pairs, and cached `unsat` reuse only generalizes previously proved false constraint subsets

## Benchmark

Local replay workload from `grandizzy/symbolic-bug-suite`, prebuilt forge binaries (`BUILD_FORGE=0`). Expected nonzero exit was retained; all selected-workload runs kept 8 `counterexample:` lines and 31 symbolic paths.

| workload | baseline | this PR | query shape |
| --- | ---: | ---: | --- |
| selected 4-contract replay, controlled `REPS=15` median | 918-922 ms | 609 ms | symbolic queries `50 -> 40`, this PR SMT queries `37` |
| selected 4-contract replay, post-review validation | n/a | 8 counterexamples, 31 paths | symbolic queries `40`, SMT `37`, hard-arith `0` |
| full 22-contract replay, post-review validation | n/a | 22/22 expected counterexamples retained | symbolic queries `365`, SMT `336`, sat-cache hits `55`, hard-arith `7` |

Earlier selected replay passes showed cold run-1 outliers (`4517 ms` baseline, `3897 ms` this PR), but medians stayed stable after reversing run order. The post-review validation retained the selected-workload query shape after removing pre-solver hard-arithmetic `is_sat` fallback. That removal intentionally trades some full-suite SMT calls back so decidable safe proofs are not marked incomplete.

## Notes

- This is workload-specific local measurement, not a broad symbolic-performance claim.
- The PR does not include local autoresearch harness files or experiment logs.

### DIFF
--- a/crates/evm/symbolic/src/runtime/solver.rs
+++ b/crates/evm/symbolic/src/runtime/solver.rs
@@ -526,7 +526,7 @@ impl SmtLibSubprocessSolver {
 
 /// Returns a structural key for normalized solver cache lookups.
 fn constraint_cache_key(constraints: &[BoolExpr]) -> Vec<BoolExpr> {
-    let mut key = Vec::new();
+    let mut key = Vec::with_capacity(constraints.len());
     for constraint in constraints.iter().cloned().map(cache_key_bool) {
         collect_cache_key_conjunct(constraint, &mut key);
     }
--- a/crates/evm/symbolic/src/runtime/solver.rs
+++ b/crates/evm/symbolic/src/runtime/solver.rs
@@ -484,7 +484,7 @@ impl SmtLibSubprocessSolver {
         let commands =
             ordered_commands.iter().map(|(_, command)| command.clone()).collect::<Vec<_>>();
 
-        let mut smt = String::new();
+        let mut smt = String::with_capacity(256 + smt_constraints.len().saturating_mul(192));
         smt.push_str("(set-logic QF_BV)\n");
         if commands.iter().all(|command| command.smt_timeout)
             && let Some(timeout) = self.timeout.filter(|timeout| *timeout > 0)
--- a/crates/evm/symbolic/src/runtime/solver.rs
+++ b/crates/evm/symbolic/src/runtime/solver.rs
@@ -277,6 +277,12 @@ impl SymbolicSolver for SmtLibSubprocessSolver {
             trace!(result, "is_sat: normalized cache hit");
             return Ok(*result);
         }
+        if self.has_cached_unsat_subset(&cache_key) {
+            self.sat_cache_hits += 1;
+            trace!("is_sat: normalized unsat subset cache hit");
+            self.cache_sat_result(cache_key, false);
+            return Ok(false);
+        }
 
         self.reserve_query()?;
         self.record_query();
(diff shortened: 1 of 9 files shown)


mablr (approved):
sgtm, ci failure doesn't seem related to changes.

mattsse (approved):

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -284,17 +294,31 @@ impl SymbolicSolver for SmtLibSubprocessSolver {
...
             return Ok(false);
         }
+        if constraints_prefer_hard_arith_fallback_first(&smt_constraints)
+            && validated_hard_arith_fallback_model(&smt_constraints, constraints).is_some()
+        {
+            self.heuristic_witnesses += 1;
+            trace!("is_sat: validated hard arithmetic fallback model before solver");
+            self.cache_sat_result(cache_key, true);
+            return Ok(true);
+        }
```
stevencartavia: Worth a look: this new pre-solver hard-arith path in `is_sat` turns `checkPacked` from `[PASS]` into a spurious `Incomplete`/FAIL:

```
[FAIL: incomplete symbolic execution (Timeout): hard arithmetic heuristic witness used...]
  checkPacked(...) (queries: 16, smt: 13, sat: 16, hard-arith: 3)
```

(`Timeout` is the `SymbolicStopReason`, not wall-clock — solver was ~140ms.) Removing only this block makes it pass: `(smt: 16, hard-arith: 0)` — z3 decides all 16 queries without fallback.

Likely cause: `constraints_prefer_hard_arith_fallback_first` is syntactic and doesn't check whether z3 can decide the query, so this path records a heuristic witness on decidable queries — and `run_inner` treats any heuristic witness as "can't certify safety." On base, `is_sat` only fell back *after* z3 returned `unknown`.

Simplest fix: drop this `is_sat` block, keep the post-`unknown` fallback. (`model()` already pre-solves on base — leave it.) The `is_sat_uses_validated_hard_arithmetic_fallback_before_solver` test needs updating too.

crates/forge/src/result.rs
```diff
@@ -1455,15 +1459,17 @@ impl fmt::Display for TestKindReport {
...
                 model_queries,
                 sat_cache_hits,
                 model_cache_hits,
+                heuristic_witnesses,
                 solver_time_ms,
             } => {
                 write!(
                     f,
-                    "(paths: {paths}, queries: {solver_queries}, sat: {sat_queries} ({sat_cache_hits} cached), models: {model_queries} ({model_cache_hits} cached), solver: {solver_time_ms}ms)"
+                    "(paths: {paths}, queries: {solver_queries}, smt: {smt_queries}, sat: {sat_queries} ({sat_cache_hits} cached), models: {model_queries} ({model_cache_hits} cached), hard-arith: {heuristic_witnesses}, solver: {solver_time_ms}ms)"
```
stevencartavia, line 1472: Worth a look: this format change adds `smt:`/`hard-arith:`, but the `[METRICS]` regex in `symbolic_helpers.rs` wasn't updated, so the new suffix isn't redacted and the symbolic snapshots fail:

```
- checkZeroIsZero() ([METRICS])
+ checkZeroIsZero() ([METRICS], smt: 0, sat: 0 (0 cached), ..., hard-arith: 0, solver: 0ms)
```

The optional group expects `, sat:` right after `queries:`, but the format now has `, smt:` first. Updating both regex occurrences fixes it (`symbolic_engine_capabilities` 0/12 → 12/12):

```rust
r"paths: \d+, queries: \d+(?:, smt: \d+, sat: \d+ \(\d+ cached\), models: \d+ \(\d+ cached\), hard-arith: \d+, solver: \d+ms)?"
```

Independent of the `is_sat` item. No z3 install step in `.github`, so CI may not be exercising these.

stevencartavia (approved):

PR #15070, opened 2026-06-04
perf(fuzz): reduce invariant campaign memory retention

decofe:
## Summary
- clear per-run invariant input/cmp buffers after the corpus manager has consumed them
- track invariant run/call counts directly instead of retaining every successful fuzz case payload
- shard restored in-memory corpus entries across invariant workers
- persist successful campaign corpus entries without cmp operand payloads
- move out final invariant test data so worker-local fuzz state and targeted-contract state are dropped before returning worker output
- explicitly drop the worker corpus manager once the aggregate worker result has been built

## Context
Tempo saw long invariant campaigns pin memory near the pod limit across completed invariant contracts. This is a follow-up to #15057 that targets worker-local retained campaign state, retained success-case payloads, and duplicated corpus state rather than final corpus filtering clones.

## Validation
- cargo fmt --check
- git diff --check
- cargo test -p foundry-evm --lib executors::invariant --no-default-features

### DIFF
--- a/crates/evm/evm/src/executors/corpus.rs
+++ b/crates/evm/evm/src/executors/corpus.rs
@@ -188,7 +188,6 @@ impl CorpusEntry {
 #[derive(Debug, Clone)]
 pub(crate) struct CampaignCorpusEntry {
     tx_seq: Vec<BasicTxDetails>,
-    cmp_seq: Vec<Vec<CmpOperands>>,
     dedupe_by_coverage: bool,
 }
 
@@ -248,6 +247,21 @@ impl WorkerCorpusSeed {
         if let Some((value, sequence)) = load_optimization_state(config) {
             self.optimization_best_value = Some(value);
             self.optimization_best_sequence = sequence;
+        }
+        self
+    }
+
+    pub(crate) fn for_worker(mut self, worker_id: usize, worker_count: usize) -> Self {
+        if worker_count > 1 {
+            self.in_memory_corpus = self
+                .in_memory_corpus
+                .into_iter()
+                .enumerate()
+                .filter_map(|(idx, entry)| (idx % worker_count == worker_id).then_some(entry))
+                .collect();
+            self.metrics.corpus_count = self.in_memory_corpus.len();
+            self.metrics.favored_items =
+                self.in_memory_corpus.iter().filter(|entry| entry.is_favored).count();
         }
         self
     }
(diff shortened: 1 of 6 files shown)


stevencartavia (dismissed):

figtracer (dismissed):

mablr (commented):
Nice, less memory alloc 👌

mablr (approved):

stevencartavia (approved):

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -252,6 +251,21 @@ impl WorkerCorpusSeed {
         self
     }
 
+    pub(crate) fn for_worker(mut self, worker_id: usize, worker_count: usize) -> Self {
+        if worker_count > 1 {
+            self.in_memory_corpus = self
+                .in_memory_corpus
+                .into_iter()
+                .enumerate()
+                .filter_map(|(idx, entry)| (idx % worker_count == worker_id).then_some(entry))
```
0xalpharush, line 260: This is not the kind of change to make without a benchmark

0xalpharush (changes requested):

crates/evm/evm/src/executors/corpus.rs
mablr: Benchmarked the touched corpus seed fan-out path against the parent commit.

Harness: https://gist.github.com/mablr/b87b9d75f3723b3de8b5796c72331ae2
(focused Rust harness + `hyperfine`, comparing parent full-clone behavior vs PR clone-then-`for_worker` sharding)

Results, 8 workers:

| Entries | Parent | PR | Delta |
|---:|---:|---:|---:|
| 1k | 99.4 ms | 103.6 ms | PR ~4% slower |
| 10k | 255.9 ms | 274.8 ms | PR ~7% slower |
| 50k | 298.7 ms | 354.9 ms | PR ~19% slower |

So this does add measurable startup CPU overhead. It still reduces retained per-worker corpus memory after startup, but clone-then-filter is not free.

mablr: I'll follow that up asap

mablr: edit: mb micro-benchmark is not relevant here, as neglictable

see #15106

PR #15075, opened 2026-06-05
feat(cast): add Tempo wallet session runner

0xKarl98:
Step towards OSS-162 , 

Adds a top-level `cast wallet session` runner mode alongside the existing explicit `create` and `revoke` subcommands.

The new flow:

  - parses session policy flags such as `--root`, `--expires`, `--scope`, `--target`, `--selector`, and `--spend-limit`
  - resolves the session chain id from CLI config or RPC
  - creates and stores a temporary session entry signed by the root account
  - runs the inner command with `TEMPO_SESSION_ID` in its environment
  - strips inherited signer-related environment variables from the child process
  - retires local session key material before revoke preflight
  - revokes the session on-chain after the child command exits
  - handles child failures and wrapper interrupts while still attempting cleanup

This also keeps `Revoking` session entries retryable by preserving key material until a terminal status is reached. That makes explicit revoke attempts remain retryable, while the session runner can still retire its local key material before attempting cleanup.

### DIFF
--- a/crates/cast/src/cmd/wallet/session.rs
+++ b/crates/cast/src/cmd/wallet/session.rs
@@ -111,21 +111,23 @@ impl SessionArgs {
             root_account.ok_or_else(|| eyre::eyre!("cast wallet session requires --root"))?;
         let expires =
             expires.ok_or_else(|| eyre::eyre!("cast wallet session requires --expires"))?;
-        let for_command =
+        let command =
             for_command.ok_or_else(|| eyre::eyre!("cast wallet session requires --for"))?;
+        let command = InnerCommand::parse(command)?;
+        let scope = session_scope(scope, target, selectors)?;
         let send_tx = *send_tx;
         let chain_id = resolve_session_chain_id(&send_tx).await?;
 
-        run_for_command(
+        run_for_command(SessionRunRequest {
             root_account,
             chain_id,
             expires,
-            session_scope(scope, target, selectors)?,
+            scope,
             spend_limits,
-            for_command,
-            *tx,
+            command,
+            tx: *tx,
             send_tx,
-        )
+        })
         .await
     }
 }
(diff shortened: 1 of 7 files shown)


crates/cast/src/cmd/wallet/session.rs
```diff
@@ -96,6 +205,290 @@ impl SessionSubcommands {
...
+            .kill_on_drop(true)
+            .spawn()
+            .wrap_err_with(|| format!("failed to run inner command `{}`", self.raw))?;
+
+        let status = tokio::select! {
+            status = child.wait() => status.wrap_err_with(|| {
+                format!("failed to wait for inner command `{}`", self.raw)
+            })?,
+            interrupt = interrupt => {
+                let _ = child.start_kill();
```
figtracer: this only kills the direct child. if `--for` launches a shell or runner that leaves grandchildren behind, cleanup can revoke the session while work is still running. can we run the child in its own process group and signal that group?

crates/cast/tests/cli/keychain.rs
```diff
@@ -362,3 +362,153 @@ casttest!(wallet_session_revoke_wrong_chain_preserves_local_key, async |_prj, cm
 
     assert_session_file_status_with_key(tempo_home.path(), "active");
 });
+
+casttest!(wallet_session_run_for_wraps_child_and_cleans_key_material, async |_prj, cmd| {
```
figtracer: these tests only assert the child sees the session id, not that a child command can resolve and use it for signing. can we add one live child command that actually signs or broadcasts through the session?

figtracer (commented):
two small comments from local review.

crates/cast/tests/cli/keychain.rs
0xKarl98: Was trying to make it in next pr , hopefully make it tiny but def we can do it in one

mattsse (dismissed):
LGTM

crates/cast/tests/cli/keychain.rs
0xKarl98: Have addressed the first comment , but leave the second in next pr to be reviewer-friendly

mattsse (approved):

mablr (approved):

PR #15076, opened 2026-06-05
feat(cheatcodes): `vm.rpcJson` 

### LINKED ISSUE #15072: `vm.rpc` should return raw JSON string rather than tuple-ification
### Component

Forge

Currently foundry's `vm.rpc` cheatcode will return bytes representing a [tuplification of the JSON data](https://github.com/foundry-rs/foundry/blob/5313c682bab88e730ca133b3995bfc5ffaa22f65/crates/cheatcodes/src/json.rs#L527) returned by the RPC. The tuple doesn't appear to contain any key data, so the receiving solidity code needs to know the exact structure of the JSON in advance.

This is a problem for complex return values for useful functions like `eth_getTransactionByHash`, as the fields returned by the function can vary from RPC provider to RPC provider, or transaction to transaction, so its impossible to predict which data structures will be where.

Foundry already provides excellent [JSON parsing with extraction of value by key](https://www.getfoundry.sh/reference/cheatcodes/parse-json#type-coercion) through its existing cheatcodes, so it would be much easier and more reliable to work with the string directly.

This is a request to either change the existing type of `vm.rpc` to return the string of the raw JSON response *OR* to introduce a new function that returns the raw json.

### Additional context

_No response_

figtracer:
## Summary

Adds `vm.rpcJson` overloads that perform JSON-RPC requests like `vm.rpc`, but return the serialized JSON result as a Solidity string.

This keeps the existing `vm.rpc` bytes/tuple conversion behavior intact for compatibility while making complex provider-specific RPC object responses usable with the existing JSON parsing cheatcodes.

## Details

- Adds `rpcJson(string method, string params)` for the active fork URL.
- Adds `rpcJson(string urlOrAlias, string method, string params)` for explicit endpoints or aliases.
- Shares the underlying RPC request helper with `vm.rpc`.
- Adds fork cheatcode coverage for JSON scalar results and key-preserving transaction object parsing.

Closes #15072.

### DIFF
--- a/crates/cheatcodes/src/evm/fork.rs
+++ b/crates/cheatcodes/src/evm/fork.rs
@@ -457,7 +457,7 @@ fn rpc_result(url: &str, method: &str, params: &str) -> Result<serde_json::Value
     let provider = ProviderBuilder::<AnyNetwork>::new(url).build()?;
     let params_json: serde_json::Value = serde_json::from_str(params)?;
     foundry_common::block_on(provider.raw_request(method.to_string().into(), params_json))
-        .map_err(|err| fmt_err!("{method:?}: {err}").into())
+        .map_err(|err| fmt_err!("{method:?}: {err}"))
 }
 
 /// Convert fixed bytes and address values to bytes in order to prevent encoding issues.
--- a/crates/cheatcodes/src/evm/fork.rs
+++ b/crates/cheatcodes/src/evm/fork.rs
@@ -228,6 +228,23 @@ impl Cheatcode for rpc_1Call {
     }
 }
 
+impl Cheatcode for rpcJson_0Call {
+    fn apply_stateful<FEN: FoundryEvmNetwork>(&self, ccx: &mut CheatsCtxt<'_, '_, FEN>) -> Result {
+        let Self { method, params } = self;
+        let url =
+            ccx.ecx.db().active_fork_url().ok_or_else(|| fmt_err!("no active fork URL found"))?;
+        rpc_json_call(&url, method, params)
+    }
+}
+
+impl Cheatcode for rpcJson_1Call {
+    fn apply<FEN: FoundryEvmNetwork>(&self, state: &mut Cheatcodes<FEN>) -> Result {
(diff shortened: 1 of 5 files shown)


crates/cheatcodes/src/evm/fork.rs
```diff
@@ -434,6 +447,19 @@ fn rpc_call(url: &str, method: &str, params: &str) -> Result {
...
+fn rpc_json_call(url: &str, method: &str, params: &str) -> Result {
+    let result = rpc_result(url, method, params)?;
+    Ok(serde_json::to_string(&result)?.abi_encode())
+}
+
+fn rpc_result(url: &str, method: &str, params: &str) -> Result<serde_json::Value> {
+    let provider = ProviderBuilder::<AnyNetwork>::new(url).build()?;
+    let params_json: serde_json::Value = serde_json::from_str(params)?;
+    foundry_common::block_on(provider.raw_request(method.to_string().into(), params_json))
+        .map_err(|err| fmt_err!("{method:?}: {err}").into())
```
mablr: fmt_err! already returns the target error type, so the trailing .into() is a denied clippy::useless_conversion. Dropping .into() should unblock CI

testdata/default/cheats/Fork2.t.sol
```diff
@@ -360,6 +374,23 @@ contract ForkTest is Test {
         assertEq(txn.to, 0xF04fd9a66DE511BC389D3b830C1F850a4A4A8c61, "tx to mismatch");
         assertEq(txn.blockNumber, hex"588b24", "tx blockNumber mismatch");
     }
+
+    function testRpcJsonTransactionByHashPreservesKeys() public {
+        string memory data = vm.rpcJson(
+            "https://sepolia.drpc.org",
```
grandizzy: Let's do mainnet as the other tests here, that is to avoid flakiness

grandizzy (changes requested):
lgtm, smal test nit

grandizzy (approved):
makes sense, ty!

mablr (approved):

PR #15079, opened 2026-06-05
perf(symbolic): reduce solver work with path-bounded arithmetic guards

figtracer:
## Summary

- Use normalized path upper bounds when proving Solidity checked-mul guards are tautological.
- Keep the rewrite query-local: it only collapses `guard == 0` when the current conjunction proves the multiplication cannot overflow.
- Add regression coverage for direct bounds, normalized `not(a > bound)` bounds, and loose/unbounded cases that must remain solver-visible.

## Scope

This optimizes a class of checked-mul guard queries where the overflow proof needs path context, for example `a <= 1000` plus `a == 0 || ((scale * a) / a == scale)`. In the current symbolic bug suite, the only material hit is `MonoXSameTokenTest`; the other replayed counterexamples keep the same query and SMT counts.

## Benchmarks

All benchmark runs still found the expected counterexamples. The MonoX samples replayed the same counterexample bytes with `args=[1000]`.

| Scenario | Queries | SMT | Solver time |
| --- | ---: | ---: | ---: |
| `MonoXSameTokenTest` merged-base direct sample | 18 | 16 | 47,198 ms |
| `MonoXSameTokenTest` after, direct sample 1 | 17 | 14 | 17,150 ms |
| `MonoXSameTokenTest` after, direct sample 2 | 17 | 14 | 17,246 ms |
| `MonoXSameTokenTest` after, direct sample 3 | 17 | 14 | 16,812 ms |
| Full symbolic bug suite merged-base same-session sample | 360 | 327 | 54,608 ms |
| Full symbolic bug suite after same-session sample | 359 | 325 | 26,393 ms |

### DIFF
--- a/crates/evm/symbolic/src/runtime/solver/normalize.rs
+++ b/crates/evm/symbolic/src/runtime/solver/normalize.rs
@@ -2,8 +2,28 @@ use super::*;
 
 /// Normalizes path constraints into an equivalent, solver-friendlier form.
 pub(crate) fn normalize_constraints_for_solver(constraints: &[BoolExpr]) -> Vec<BoolExpr> {
-    let mut normalized = Vec::with_capacity(constraints.len());
-    for constraint in constraints.iter().cloned().map(normalize_bool_for_solver) {
+    let normalized = normalize_constraint_batch(
+        constraints.iter().cloned().map(normalize_bool_for_solver),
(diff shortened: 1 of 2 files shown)


grandizzy (approved):
🫡

mablr (approved):
noice

PR #15081, opened 2026-06-05
feat(cast): support Tempo sessions in cast send

0xKarl98:
This PR wires Tempo wallet session resolution into `cast send`.

  - Detects `--tempo.session` / `TEMPO_SESSION_ID` before resolving normal wallet signers and routes the command through
  `TempoNetwork`.
  - Resolves the live session signer once the provider chain id is available.
  - Reuses the existing Tempo access-key signing path with the resolved session key and access-key metadata.
  - Rejects incompatible signing modes such as `--unlocked`, `--browser`, and explicit wallet signer options when a session is selected.
  - Adds CLI coverage for `cast wallet session --for` running `cast send`, including a nested child process that inherits `TEMPO_SESSION_ID`.

### DIFF
--- a/crates/cast/src/cmd/send.rs
+++ b/crates/cast/src/cmd/send.rs
@@ -12,7 +12,7 @@ use clap::Parser;
 use eyre::{Result, eyre};
 use foundry_cli::{
     opts::TransactionOpts,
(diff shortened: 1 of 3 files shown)


mattsse (approved):
lgtm

mablr (approved):
cool

PR #15085, opened 2026-06-05
feat(cast): support Tempo sessions in batch-send

0xKarl98:
Towards OSS-162 , 

Add Tempo session support to `cast batch-send` by deferring signer resolution until after the chain is known, so session keys can be validated against the active chain before use. The command now rejects incompatible signing modes such as explicit wallet signers, `--unlocked`, and `--browser` when a Tempo session is active.

This also adds CLI coverage for successful batch-send session usage, inherited session usage through `wallet session --for`, explicit signer rejection, unlocked signer rejection, and wrong-chain session rejection.


mablr (approved):
lg

stevencartavia (approved):

Issue #10451, opened 2025-05-07
Forge test do not show decoded events when emitting events using inline assembly


thepluck:
### Component

Forge

### Have you ensured that all of these are up to date?

- [ ] Foundry
- [x] Foundryup

### What version of Foundry are you on?

forge Version: 1.0.0-v1.0.0

### What version of Foundryup are you on?

foundryup: 1.1.0

### What command(s) is the bug in?

forge test

### Operating System

None

### Describe the bug

I use the below snippet to emit event:
```
  event Exchange(bytes swapResult);

  function emitExchange(
    address pool,
    address tokenIn,
    address tokenOut,
    uint256 amountIn,
    uint256 amountOut
  ) internal {
    assembly ("memory-safe") {
      // Get a pointer to some free memory.
      let fmp := mload(0x40)

      // Write the data into memory in reverse order because we are packing it into 92 bytes.
      // We will use 32 + 20 + 20 + 16 + 16 = 104 bytes of memory.
      mstore(add(fmp, 72), amountOut) // 104 - 32 = 72
      mstore(add(fmp, 56), amountIn) // 72 - 16 = 56
      mstore(add(fmp, 40), tokenOut) // 56 - 16 = 40
      mstore(add(fmp, 20), tokenIn) // 40 - 20 = 20
      mstore(fmp, pool) // 20 - 20 = 0

      // Emit the event.
      // The actual data is starting from fmp + 12
      log1(add(fmp, 12), 92, EXCHANGE_EVENT_SELECTOR)

      // Now clean the memory we used
      mstore(fmp, 0)
      mstore(add(fmp, 0x20), 0)
      mstore(add(fmp, 0x40), 0)
      mstore(add(fmp, 0x60), 0)
    }
  }
```

When running tests, instead of showing `emit Exchange(...)`, it shows the below:

<img width="1630" alt="Image" src="https://github.com/user-attachments/assets/35d01382-0a8c-4a10-bbcc-6eb9fc6402a2" />

And for some places, it shows:

<img width="675" alt="Image" src="https://github.com/user-attachments/assets/3bd341fb-233d-4b6f-a228-81b376472a13" />

grandizzy:
@thepluck this is an internal fn, wonder if you run with dynamic test linking on? could you share all the project settings to repro?

grandizzy:
nm, was able to repro with a simple test, will look into

thepluck:
That is an internal function of a contract

[...]

PR #469, opened 2022-01-15
fix cast autocompletion

clifton:
re: https://github.com/clap-rs/clap/pull/3295

PR #15087, opened 2026-06-05
fix(traces): preserve event name when log data is not ABI-decodable

0xMars42:
## Motivation

When an event is emitted with data that is not ABI-encoded (for example
hand-packed bytes written via inline assembly, as reported in #10451), its
`topic0` still matches a known event but `event.decode_log` fails.
`decode_event` (`crates/evm/traces/src/decoder/mod.rs`) then returned
`DecodedCallLog { name: None, params: None }`, so the matched name was
dropped and the trace fell back to raw `topic 0 / data`.

## Solution

When decoding fails but `topic0` matched a known event, surface the matched
event name with `params: None` instead of dropping it. Unknown events (no
match) and anonymous logs (no topics) are unchanged.

## Scope

This is the decoder half of #10451, not the full fix. With the current
revm-inspectors writer, `name: Some / params: None` renders as `emit Exchange()`,
so the raw topics/data are not yet shown next to the name. The writer change that
renders a named header plus raw topics/data is the follow-up:
paradigmxyz/revm-inspectors#466. Once that lands and the `revm-inspectors`
dependency here is bumped, #10451 is fully addressed.

## Test

`test_decode_event_preserves_name_when_data_is_not_abi` registers an event,
feeds a log whose `topic0` matches but whose data is hand-packed (not ABI),
and asserts the decoded name is `Some("Exchange")` with `params: None`. It
fails before the change (name is `None`) and passes after.

Part of #10451 (decoder half).

### DIFF
--- a/crates/evm/traces/src/decoder/mod.rs
+++ b/crates/evm/traces/src/decoder/mod.rs
@@ -901,7 +901,9 @@ impl CallTraceDecoder {
             }
         }
 
-        DecodedCallLog { name: None, params: None }
+        // The data failed to decode, but if topic0 matched a known event we still surface
+        // its name (without params) instead of dropping the log to raw topics/data.
+        DecodedCallLog { name: events.first().map(|e| e.name.clone()), params: None }
     }
 
     /// Prefetches function and event signatures into the identifier cache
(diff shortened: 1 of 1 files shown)


crates/evm/traces/src/decoder/mod.rs
```diff
@@ -901,7 +901,9 @@ impl CallTraceDecoder {
             }
         }
 
-        DecodedCallLog { name: None, params: None }
+        // The data failed to decode, but if topic0 matched a known event we still surface
+        // its name (without params) instead of dropping the log to raw topics/data.
+        DecodedCallLog { name: events.first().map(|e| e.name.clone()), params: None }
```
mablr, line 906: I think this goes in the right direction but it only fixes half of #10451. A change in `revm-inspectors` writer is certainly needed to render `named-but-undecoded` logs as the event name plus raw topics/data.

You may rescope this PR in the meantime, as you prefer.

mablr (changes requested):

0xMars42:
Thanks @mablr yep, this is the decoder half I flagged in the PR description; happy to pick up the revm-inspectors follow-up I offered there.

I checked `TraceWriter::write_log` on current main: it branches only on `decoded_name()`, so `name: Some / params: None` renders `emit Exchange()` and the raw topics/data `else` branch is skipped — the raw data is dropped. So on its own this PR trades "raw data, no name" for "name, no data"; the writer change is what actually closes #10451.

Plan:
1. PR on paradigmxyz/revm-inspectors adding the missing arm to `write_log`: extract the current raw topics/data block into a helper and match on `(decoded_name(), decoded_params())` — `(Some, Some)` decoded (unchanged), `(Some, None)` name + raw topics/data (new), `(None, _)` raw (unchanged).
2. Once released, update the `revm-inspectors` dependency here (currently 0.41.1) to pick it up.

One format call before I open it: for the `(Some, None)` case, do you prefer the `emit Name( … )` wrapper with the raw data inline (the `emit Exchange(0x..)` I mocked in the description), or an `emit Name` header followed by indented `topic i:` / `data:` lines like the current raw branch? The raw block prints its own `emit topic 0` prefix, so it's a small refactor either way 
just want to match whatever style you'd merge.

mablr:
@0xMars42 Yes, please do the revm-inspectors follow-up; imo a named header plus raw topics/data are good. Until then this PR can be shipped but should be described as the decoder half rather than the complete #10451 fix.

0xMars42:
The writer half is now up as paradigmxyz/revm-inspectors#469: when a log has a known name but no decoded params, it renders the event name followed by the raw topics and data instead of an empty `Name()`.

This PR stays as the decoder half (surfacing the matched name with `params: None`). Once #469 is released and bumped here, the two together render named-but-undecoded logs as the event name plus raw topics and data, which is the full #10451 fix.

This can stay as the decoder half or wait for the revm-inspectors release, as you prefer.

mablr (commented):
Thank you @0xMars42, given the current diff let's wait https://github.com/paradigmxyz/revm-inspectors/pull/469 merge.

Issue #14437, opened 2026-04-24
Invariant Fuzzer: Phased improvement plan to close gap with Echidna/Medusa (SCFuzzBench)


grandizzy:
## Context

[SCFuzzBench](https://scfuzzbench.com/) benchmarks smart contract fuzzers (Echidna, Medusa, Foundry, Recon) against real DeFi protocols (Aave v4, Superform, Liquity). Current results show Foundry finding **0-3 bugs** vs Echidna's **10+** on the same targets with the same time budget.

Analysis of the Foundry invariant testing engine identified several root causes, ranging from architectural blockers to corpus/coverage quality issues. This issue tracks the phased improvement plan.

### Benchmark Baseline (Aave v4, 24h, SCFuzzBench)

| Fuzzer | Bugs Found | AUC (norm) | Coverage Proxy |
|--------|-----------|------------|---------------|
| Echidna 2.3.1 | 10 [9,10] | 0.816 | 456k |
| Medusa 1.4.1 | 10 [9,10] | 0.754 | 5.2k |
| **Foundry v1.6.0-rc1** | **3 [3,3]** | **0.272** | **374** |

---

## Phase 1: Architectural Blockers (prerequisite)

These must land first — without them, the fuzzer uses ~4% of its time budget because trivially-findable assertion failures stop exploration immediately.

- [x] **`fail_on_assert` config** — Gate assertion-failure detection behind a config flag so users can opt in to treating Solidity `assert()` panics in target functions as invariant breaks.
  - Merged: 419f787e5 (`#14275`)

- [ ] **`continuous_run` mode** — Continue running after finding an invariant/assertion failure. Record it and keep searching for more failures. Currently, the first `assert_canary(0)` kills every invariant test in <100 runs.
  - PR: #12587 (open, review required)
  - Impact: 0-3 bugs → 6 bugs (all found in <15s)

- [ ] **Preflight check in `continuous_run` mode** — When `continuous_run = true`, the preflight invariant check (which verifies invariants hold before fuzzing starts) should record failures but not abort the campaign. Currently an always-failing canary invariant causes `"failed to set up invariant testing environment"`.
  - Location: `crates/evm/evm/src/executors/invariant/mod.rs` (`prepare_test` → preflight check)
  - Should be part of #12587

---

## Phase 2: Dictionary Quality

[...]

PR #15088, opened 2026-06-06
feat(fuzz): add Insert/Delete/Swap invariant corpus mutations

0xMars42:
## Motivation

The invariant fuzzer's corpus mutation engine
(`crates/evm/evm/src/executors/corpus.rs`) has Splice / Repeat / Interleave /
Prefix / Suffix / Abi / Cmp mutations but is missing the basic structural
mutations called out in Phase 3 of #14437: inserting, deleting and swapping
calls in a sequence. These let the engine reach shorter and reordered call
sequences that the current set does not produce directly.

## Solution

Adds three mutations alongside the existing ones:

- `Insert`: add a freshly generated call at a random position.
- `Delete`: remove a random call (kept if it would empty the sequence, so the
  sequence stays runnable).
- `Swap`: exchange two distinct calls.

The per-mutation logic is extracted into small helpers (`insert_call` /
`delete_call` / `swap_calls`), mirroring the existing `cmp_mutate` helper, so the
bounds handling is unit-testable.

## Test

Deterministic unit tests for each helper, covering the bounds and the
structural change:

- `swap_calls`: no-op on empty or single-element sequences (the bound that
  prevents an out-of-range index); exchanges the two calls on a length-2
  sequence; on a longer sequence it stays a permutation that changes exactly
  two positions.
- `delete_call`: keeps at least one call, and removes exactly one while keeping
  the survivors in their original order.
- `insert_call`: keeps the original calls and adds the generated one.

## Note on validation

The unit tests prove each mutation's structural behavior (and that they reach
sequence shapes the existing set does not produce directly: within-sequence
reordering, fresh interior insertion, interior deletion). I have not run an
end-to-end bug-finding benchmark here: the SCFuzzBench numbers referenced in
#14437 are best measured with that harness, and a full comparison also depends
on the Phase 1 `continuous_run` work (#12587).

If a coverage or bug-count delta would help before merging, I can run a specific
target from that suite; I wanted to get the implementation reviewed first. I can
also revisit the mutation weighting (these are added at equal weight) if you'd
prefer it tuned.

Addresses #14437 (Phase 3, structural mutations).

### DIFF
--- a/crates/evm/evm/src/executors/corpus.rs
+++ b/crates/evm/evm/src/executors/corpus.rs
@@ -103,6 +103,12 @@ enum MutationType {
     /// Replace input bytes using comparison operands observed for a corpus entry
     /// (input-to-state, LibAFL-style).
     Cmp,
+    /// Insert a freshly generated call at a random position in the sequence.
+    Insert,
+    /// Delete a random call from the sequence (no-op if it would empty the sequence).
+    Delete,
+    /// Swap two calls in the sequence.
+    Swap,
 }
 
 /// Persisted optimization state: the best value found and the sequence that produced it.
@@ -711,6 +717,9 @@ impl WorkerCorpus {
             Just(MutationType::Suffix),
             Just(MutationType::Abi),
             Just(MutationType::Cmp),
+            Just(MutationType::Insert),
+            Just(MutationType::Delete),
+            Just(MutationType::Swap),
         ]
         .boxed();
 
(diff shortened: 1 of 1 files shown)


figtracer (approved):
lgtm

mablr (commented):
Maybe superseded by #15367 @figtracer

mablr:
Sorry @0xMars42, thank you for the effort 🫡

PR #15089, opened 2026-06-06
 refactor(script): isolate Tempo session signer handling

0xKarl98:
Towards OSS-162 , 

Extract the Forge script Tempo session signer policy into `script::session`, including:

- scoped signer identity by `(chain, sender)`
- remaining unsigned transaction signer lookup data
- single-session-sender validation
- insertion of a live session access key only when it covers the remaining transaction set

`broadcast.rs` now uses this module when preparing raw signer maps, and `build.rs` uses the same scope model when checking whether a resumed script has all required signers available.

Also adds a CLI regression test for `cast wallet session --for "<forge> script ... --broadcast"` and updates the test utility helper that builds sibling Foundry binaries so it preserves the current Cargo target directory and profile.

### DIFF
--- a/crates/cast/tests/cli/keychain.rs
+++ b/crates/cast/tests/cli/keychain.rs
@@ -600,7 +600,7 @@ contract SessionForgeScript is Script {{
     );
 
     let for_command = format!(
-        "{} script {} --tc SessionForgeScript --broadcast --rpc-url {} --root {}",
+        "{} script {} --tc SessionForgeScript --broadcast --timeout 1 --rpc-url {} --root {}",
         prj.ensure_foundry_bin("forge").to_slash_lossy(),
         script.to_slash_lossy(),
         rpc,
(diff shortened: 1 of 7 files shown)


mattsse (dismissed):
LGTM

stevencartavia (dismissed):

mablr (dismissed):

crates/cast/tests/cli/keychain.rs
```diff
@@ -580,15 +562,108 @@ test -n "${{TEMPO_SESSION_ID:-}}"
...
+        vm.stopBroadcast();
+    }}
+}}
+"#,
+            recipient = accounts::ADDR3,
+        ),
+    );
+
+    let for_command = format!(
+        "{} script {} --tc SessionForgeScript --broadcast --rpc-url {} --root {}",
```
mablr: `wallet_session_run_for_forge_script_submits_with_session_key` reaches timeout (90s).

The test only needs to prove the session key path submits and writes a broadcast hash, so using the full `forge script` default timeout (120s) is unnecessary. Setting explicitly a lower timeout should fix this test.

mablr (dismissed):
Lgtm, just a smol nit on test :)

crates/cast/tests/cli/keychain.rs
0xKarl98: Have fixed !

PR #14459, opened 2026-04-26
fix: < to <= in validate_uint_mutation

cuiweixie:

PR #15090, opened 2026-06-07
fix: `<=` in `validate_int_mutation`

stevencartavia:
Same off-by-one as #14459 but for signed ints. `validate_int_mutation` used strict comparisons (`<`/`>`) and the wrong negative magnitude (`-(2^(size-1)-1)` instead of `-2^(size-1)`), so it rejected the boundary values `type(intN).max`, `type(intN).min` (and `min + 1`). Fixed to inclusive bounds `[-2^(size-1), 2^(size-1) - 1]` using `mutated <= max` / `mutated >= min`, and added a regression test.


grandizzy (approved):
👍 ty!

DaniPopes (approved):

PR #15091, opened 2026-06-07
fix: redact private keys from Debug

stevencartavia:
Tempo `SessionKeyMaterial` and `KeyEntry` derived `Debug` while holding raw hex private keys as `String`, so `{:?}`/`dbg!`/tracing/panic printed them (and their containers) in plaintext. Replaced the derive with manual redacting `Debug` impls and added regression tests.


grandizzy (approved):
lgtm!

mattsse (approved):

PR #15093, opened 2026-06-07
fix: clean up Tempo sessions after use

0xKarl98:
Step closer to OSS-162 , 

Previously, cleanup could fail during revoke preflight and leave the session entry `active` with local key material still present. This kept the session usable locally after the intended command lifecycle had ended.

Now  treat `revoking` as a cleanup state that clears local session key material while keeping the session metadata available for revoke retries. 
 
 Also:
 - Marks successful `cast wallet session --for` runs as `revoking` before revoke preflight, so local key material is erased even if on-chain revoke must be retried.
  - Keeps failed wrapped commands retiring the local session as `failed`.
  - Preserves existing terminal statuses instead of downgrading them during cleanup.
  - Adds `forge script --session` as a convenience alias for `--tempo.session`.
  - Makes the new alias participate in Tempo session detection, including Tempo network selection and the `--unlocked` conflict check.
  - Updates tests for revoking key cleanup and `forge script --session` parsing/behavior.

### DIFF
--- a/crates/cast/src/cmd/wallet/session.rs
+++ b/crates/cast/src/cmd/wallet/session.rs
@@ -217,17 +217,23 @@ async fn run_for_command(
     upsert_session_entry(entry)?;
 
     let child_result = command.run(session_id).await;
-    let cleanup_result = cleanup_session_run(session_id, tx, send_tx).await;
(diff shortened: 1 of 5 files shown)


mattsse (approved):
LGTM

PR #15096, opened 2026-06-08
feat(cast): support Tempo sessions in transaction commands

0xKarl98:
Towards oss-162
Adds shared Tempo helpers for resolving an explicit session before falling back to wallet signer resolution, and applies them to:

- `cast batch-mktx`
- `cast erc20`
- `cast tip20 create`
- `cast tip20 mine --register`
- `cast vaddr create`

Also rejects unsupported session combinations such as `--browser`, `--ethsign`, and `--raw-unsigned` where applicable,
validates sessions against the active chain, and fills access-key transaction metadata before signing transactions built outside `CastTxBuilder`.

Additionally, `batch-mktx --raw-unsigned` now preserves Tempo access-key metadata while set signer to None.


mattsse (approved):
LGTM

grandizzy (approved):
👍

PR #15099, opened 2026-06-08
fix(lint): improve Slither parity for existing lints

figtracer:
## Changes

- Narrow `cache-array-length` to Slither's comparable detector shape: direct storage dynamic array `.length` reads in simple `for` loop comparisons.
- Narrow `event-fields` to Slither's `unindexed-event-address` detector shape: explicit unindexed `address` and `address payable` event parameters only.
- Teach `arbitrary-send-erc20` a bounded direct-internal-call summary: internal helper parameters are treated as safe only when every direct internal callsite passes `msg.sender` / `address(this)`-safe data for that parameter.
- Stop reporting broader Foundry-only cases for those lints, including memory/calldata/bytes array lengths, id-like `uint256`/`bytes32` event fields, and all-safe internal ERC20 pull helpers.
- Update docs and UI fixtures for the narrowed existing-lint behavior.

## Parity Notes

- `cache-array-length`: removes the OpenZeppelin Foundry-only findings in the five-project corpus; one Slither-only cache-array case remains out of scope.
- `event-fields`: matches Slither's address-type detector semantics. On Solady `src`, both Foundry and Slither report 0 findings after the narrowing. On forge-std `src`, both report 4 findings.
- `arbitrary-send-erc20`: removes the Solady ERC4626 Foundry-only finding without hardcoding ERC4626. On Solady `src`, both Foundry and Slither now report 0 findings for this detector.
- I did not change `write-after-write` here even though the Solady parity table flags it as P0, because Slither is reporting local/Yul temporaries while Foundry's existing lint is storage-write oriented. Matching that needs a broader design/test pass, not a small parity narrowing.

Gist: https://gist.github.com/figtracer/bf37696353bdfc8e422cdf91d8ae74a9


crates/lint/src/sol/info/event_fields.rs
```diff
@@ -51,36 +51,9 @@ fn check_event<'ast>(ctx: &LintContext, event: &'ast ItemEvent<'ast>) {
...
-    }
-    if let Some(prefix) = name.strip_suffix("ID")
-        && let Some(last) = prefix.chars().last()
-    {
-        return last.is_ascii_lowercase();
-    }
-    false
+/// Returns true when the parameter is an `address`.
+const fn is_filterable_field(param: &VariableDefinition<'_>) -> bool {
+    matches!(&param.ty.kind, TypeKind::Elementary(ElementaryType::Address(false)))
```
mablr: This now excludes address payable event parameters because it only matches `ElementaryType::Address(false)`. Slither’s unindexed-event-address detector treats address-type params as `ElementaryType("address")`, so event `PayableAddr(address payable receiver)` should still be reported for parity.

mablr (changes requested):

crates/lint/src/sol/high/arbitrary_send_erc20.rs
```diff
@@ -1284,6 +1311,167 @@ fn seed_immutable_facts<'hir>(
...
+        if call_args.len() != arity || facts.seen.len() != arity {
+            facts.unknown = true;
+            return;
+        }
+
+        let analyzer = Analyzer::new(self.gcx, self.hir, self.has_solady_lib);
+        for (i, arg) in call_args.into_iter().enumerate() {
+            facts.seen[i] = true;
+            facts.all_safe[i] &= analyzer.is_safe(arg);
+            facts.all_self[i] &= analyzer.is_self_expr(arg);
```
mablr, line 1447: After unknown is set on a fid (arity mismatch or unparseable named args), subsequent valid callsites for the same function still create a fresh `Analyzer`, iterate all args, and write into `all_safe`/`all_self`, work silently discarded in `seed_internal_callsite_facts`. Adding if `facts.unknown { return; }` after the `entry(fid)` insert would prevent the wasted work?

mablr (approved):
sgtm

mablr:
I was able to reproduce locally similar benchmarks results (https://gist.github.com/figtracer/bf37696353bdfc8e422cdf91d8ae74a9) -> all lints at least 80x faster than slither counterparts

grandizzy (approved):
lgtm

PR #15100, opened 2026-06-08
feat(debugger): display debugger gas stats

### LINKED ISSUE #8784: Display more context/subcontext info in the debugger
### Component

Forge

'sup
Would it be possible to add more information such as the gas left, the number of subcall and the total gas used in the transaction in the debugger ? 
I am investigating gas issues and not seeing the gas left is making it way harder
Probably quite easy to implement and would be really helpful thxxx

### Additional context

_No response_

figtracer:
## Summary

Closes #8784.

- Display live `gasLeft`, aggregate `totalTraceGasUsed`, and total subcall count in the debugger opcode pane.
- Keep call-frame gas usage and refund visible with shorter lower-camel labels so the requested stats fit in normal TUI layouts.
- Count subcalls from the raw `CallTraceArena` before debugger flattening drops empty-step calls.

### DIFF
--- a/crates/debugger/src/builder.rs
+++ b/crates/debugger/src/builder.rs
@@ -42,11 +42,11 @@ impl DebuggerBuilder {
     #[inline]
     pub fn trace_arena(mut self, arena: CallTraceArena) -> Self {
         if let Some(root) = arena.nodes().first() {
-            self.stats.total_trace_gas_used =
-                self.stats.total_trace_gas_used.saturating_add(root.trace.gas_used);
+            self.stats.session_trace_gas_used =
+                self.stats.session_trace_gas_used.saturating_add(root.trace.gas_used);
         }
-        self.stats.subcalls =
-            self.stats.subcalls.saturating_add(arena.nodes().len().saturating_sub(1));
+        self.stats.session_subcalls =
+            self.stats.session_subcalls.saturating_add(arena.nodes().len().saturating_sub(1));
         flatten_call_trace(arena, &mut self.debug_arena);
         self
     }
@@ -126,28 +126,51 @@ mod tests {
         }
     }
 
+    fn trace_arena(gas_used: u64, subcalls: usize) -> CallTraceArena {
+        let mut arena = CallTraceArena::default();
+
+        {
+            let root = &mut arena.nodes_mut()[0];
+            root.trace.steps.push(step());
+            root.trace.gas_limit = 1;
+            root.trace.gas_used = gas_used;
+            root.ordering.push(TraceMemberOrder::Step(0));
+
+            for idx in 1..=subcalls {
+                root.ordering.push(TraceMemberOrder::Call(idx - 1));
+                root.children.push(idx);
+            }
+        }
+
+        for idx in 1..=subcalls {
+            arena.nodes_mut().push(CallTraceNode {
+                parent: Some(0),
+                idx,
+                trace: CallTrace { depth: 1, kind: CallKind::Call, ..Default::default() },
+                ..Default::default()
+            });
+        }
+
+        arena
+    }
+
     #[test]
     fn trace_arena_accumulates_stats() {
-        let mut arena = CallTraceArena::default();
-        let root = &mut arena.nodes_mut()[0];
-        root.trace.steps.push(step());
-        root.trace.gas_limit = 1;
-        root.trace.gas_used = 100;
-        root.ordering.push(TraceMemberOrder::Step(0));
-        root.ordering.push(TraceMemberOrder::Call(0));
-        root.children.push(1);
+        let builder = DebuggerBuilder::new().trace_arena(trace_arena(100, 1));
 
-        arena.nodes_mut().push(CallTraceNode {
-            parent: Some(0),
-            idx: 1,
-            trace: CallTrace { depth: 1, kind: CallKind::Call, ..Default::default() },
-            ..Default::default()
-        });
-
-        let builder = DebuggerBuilder::new().trace_arena(arena);
-
-        assert_eq!(builder.stats.subcalls, 1);
-        assert_eq!(builder.stats.total_trace_gas_used, 100);
+        assert_eq!(builder.stats.session_subcalls, 1);
+        assert_eq!(builder.stats.session_trace_gas_used, 100);
         assert_eq!(builder.debug_arena.len(), 1);
+    }
+
+    #[test]
+    fn trace_arena_accumulates_session_stats_across_multiple_arenas() {
+        let builder = DebuggerBuilder::new()
+            .trace_arena(trace_arena(100, 1))
+            .trace_arena(trace_arena(200, 2));
+
+        assert_eq!(builder.stats.session_subcalls, 3);
+        assert_eq!(builder.stats.session_trace_gas_used, 300);
+        assert_eq!(builder.debug_arena.len(), 2);
     }
 }
--- a/crates/debugger/src/builder.rs
+++ b/crates/debugger/src/builder.rs
@@ -42,8 +42,8 @@ impl DebuggerBuilder {
     #[inline]
     pub fn trace_arena(mut self, arena: CallTraceArena) -> Self {
         if let Some(root) = arena.nodes().first() {
-            self.stats.total_gas_used =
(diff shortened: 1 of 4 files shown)


crates/debugger/src/debugger.rs
```diff
@@ -32,10 +48,29 @@ impl Debugger {
         identified_contracts: AddressHashMap<String>,
         contracts_sources: ContractSources,
         breakpoints: Breakpoints,
+    ) -> Self {
+        let stats =
+            DebuggerStats { subcalls: debug_arena.len().saturating_sub(1), total_gas_used: 0 };
```
mablr: Silently zeroes `total_gas_used`. It has no access to `CallTraceArena` gas data. Any caller not going through `DebuggerBuilder` will permanently display `totalGasUsed: 0`. Subcall count also diverges (flattened arena length vs. raw node count).

crates/debugger/src/builder.rs
```diff
@@ -39,6 +41,12 @@ impl DebuggerBuilder {
     /// Extends the debug arena.
     #[inline]
     pub fn trace_arena(mut self, arena: CallTraceArena) -> Self {
+        if let Some(root) = arena.nodes().first() {
+            self.stats.total_gas_used =
+                self.stats.total_gas_used.saturating_add(root.trace.gas_used);
```
mablr: `total_gas_used` sums across all arenas (setup + test). `traces()` calls `trace_arena()` for each `TraceKind` (e.g. forge setup and test traces). Each root's `gas_used` is accumulated, so the displayed `totalGasUsed` is session-wide, not per-transaction. The label implies single-transaction scope; a clarifying rename or comment would help.

crates/debugger/src/tui/draw.rs
```diff
@@ -606,6 +610,28 @@ impl TUIContext<'_> {
...
+    gas_remaining: u64,
+    total_gas_used: u64,
+    call_gas_used: u64,
+    gas_refund_counter: u64,
+    subcalls: usize,
+) -> String {
+    let address = short_address(address);
+    format!(
+        "Addr: {address} | PC: 0x{pc:x} ({pc}) | gasleft: {gas_remaining} | totalGasUsed: \
+         {total_gas_used} | subcalls: {subcalls} | callGas: {call_gas_used} | refund: \
```
mablr: Should be all camelCase ?

crates/debugger/src/tui/draw.rs
```diff
@@ -606,6 +610,28 @@ impl TUIContext<'_> {
...
+         {gas_refund_counter}"
+    )
+}
+
+fn short_address(address: &Address) -> String {
+    let address = address.to_string();
+    format!("{}...{}", &address[..6], &address[address.len() - 4..])
```
mablr: uses non-standard truncation and no checksum

crates/debugger/src/builder.rs
```diff
@@ -86,7 +94,60 @@ impl DebuggerBuilder {
...
+            gas_cost: 0,
+            storage_change: None,
+            status: Some(InstructionResult::Stop),
+            immediate_bytes: None,
+            decoded: None,
+        }
+    }
+
+    #[test]
+    fn trace_arena_counts_empty_step_subcalls() {
```
mablr: test name is misleading, `trace_arena_accumulates_stats` would be more accurate?

mablr (commented):
Some comments :)

mablr (dismissed):
sgtm

crates/debugger/src/tui/draw.rs
```diff
@@ -606,6 +609,36 @@ impl TUIContext<'_> {
...
 
+fn op_list_title(
+    address: &Address,
+    pc: usize,
+    gas_remaining: u64,
+    call_gas_used: u64,
+    gas_refund_counter: u64,
+    stats: Option<DebuggerStats>,
+) -> String {
+    let address = short_checksum_address(address);
```
stevencartavia: Is this switching the address to the shortened form (0xd8dA…6045) instead of the full one master showed?

figtracer: yes

mablr (dismissed):

crates/debugger/src/tui/draw.rs
```diff
@@ -606,6 +609,36 @@ impl TUIContext<'_> {
...
+            stats.total_trace_gas_used, stats.subcalls
+        )
+        .unwrap();
+    }
+
+    title
+}
+
+fn short_checksum_address(address: &Address) -> String {
+    format!("{address:#}")
```
grandizzy: Restore full address. `{address:#}` renders the shortened checksummed form (`0xd8dA…6045`), whereas master showed the full address. In a debugger the full address is often needed to copy/compare across proxies, CREATE2, and many local test contracts. Please keep a full checksummed address here. (Addresses @stevencartavia's question.)

crates/debugger/src/tui/draw.rs
```diff
@@ -606,6 +609,36 @@ impl TUIContext<'_> {
...
+    let address = short_checksum_address(address);
+    let mut title = format!(
+        "address: {address} | pc: 0x{pc:x} ({pc}) | gasLeft: {gas_remaining} | \
+         callGasUsed: {call_gas_used} | gasRefund: {gas_refund_counter}"
+    );
+
+    if let Some(stats) = stats {
+        write!(
+            title,
+            " | totalTraceGasUsed: {} | subcalls: {}",
```
grandizzy: Label scope is misleading. These values are session-wide (summed across every `trace_arena()` call, e.g. setup + test), but `totalTraceGasUsed` / `subcalls` read as current-trace. Rename to `sessionTraceGasUsed` / `sessionSubcalls` (or `aggregate…`), or rescope to the current trace.

crates/debugger/src/debugger.rs
```diff
@@ -7,8 +7,17 @@ use foundry_evm_core::Breakpoints;
 use foundry_evm_traces::debug::ContractSources;
 use std::path::Path;
 
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct DebuggerStats {
+    /// Sum of root-call gas used across every trace arena passed to the debugger.
+    pub total_trace_gas_used: u64,
```
grandizzy: Rename fields to match the UI labels and make the aggregate/session scope explicit, e.g. `session_trace_gas_used` / `session_subcalls`. The doc comments already say "across every trace arena" — the public-facing names should too.

crates/debugger/src/builder.rs
```diff
@@ -86,7 +94,60 @@ impl DebuggerBuilder {
...
+            gas_cost: 0,
+            storage_change: None,
+            status: Some(InstructionResult::Stop),
+            immediate_bytes: None,
+            decoded: None,
+        }
+    }
+
+    #[test]
+    fn trace_arena_accumulates_stats() {
```
grandizzy, line 158: Add coverage for the behavior reviewers are worried about: multi-arena accumulation (two arenas → summed gas/subcalls), so the session-wide semantics are locked in.

crates/debugger/src/tui/draw.rs
```diff
@@ -666,6 +699,39 @@ fn hex_digits(n: usize) -> usize {
 
 #[cfg(test)]
 mod tests {
+    use crate::debugger::DebuggerStats;
+    use alloy_primitives::{Address, address};
+
+    #[test]
+    fn op_list_title_includes_gas_and_subcall_stats() {
```
grandizzy, line 706: Add a test asserting the title contains the full address (no `…`), so the truncation regression can't silently return.

grandizzy (dismissed):
Mechanics look good and switching to `saturating_sub` fixes a real underflow panic from master. Two things to fix before merge:

1. The address is now truncated (`{address:#}` renders `0xd8dA…6045`), which is a regression for a debugger where the full address is often needed to copy/compare.
2. `totalTraceGasUsed`/`subcalls` are session-wide (summed across every `trace_arena()` call, e.g. setup + test) but are labeled as if per-trace.

Suggested minimal changes: restore the full address, rename labels/fields to `session*`, and add tests for full-address rendering + multi-arena accumulation.

grandizzy (approved):
lgtm!

mablr (approved):

PR #15104, opened 2026-06-08
fix(forge): hide unreliable mutation score

### LINKED ISSUE #15102: `forge test --mutate`: per-mutant recompile makes mutation impractical on realistic compiler configs (via_ir, high optimizer_runs); timeouts silently produce a meaningless score
### Component

Forge (`forge test --mutate`)

### Describe the issue

Mutation testing recompiles the project for every mutant. On a realistically
configured project — Morpho Blue ships `via_ir = true` and
`optimizer_runs = 999999` — the per-mutant compile dominates and the vast
majority of mutants hit the per-mutant timeout, yielding an unusable run and a
misleading score.

### To Reproduce

```bash
git clone --recurse-submodules https://github.com/morpho-org/morpho-blue
cd morpho-blue

# Project default config: via_ir = true, optimizer_runs = 999999
forge test --mutate src/libraries/MathLib.sol \
  --no-match-contract "Invariant" --mutation-timeout 60
```

### Actual result

`MathLib.sol` (123 mutants):

```
│ Killed    ┆ 0         ┆ 0.0%       │
│ Invalid   ┆ 12        ┆ 9.8%       │
│ Timed out ┆ 111       ┆ 90.2%      │
Mutation Score: 0.0% (0/0 mutants killed); finished in 17m 49s
```

90% of mutants time out (the per-mutant compile alone exceeds 60s), so the score
is meaningless.

Switching to a fast profile makes the *same* mutants fully evaluable:

```bash
FOUNDRY_PROFILE=no_via_ir FOUNDRY_OPTIMIZER_RUNS=200 \
  forge test --mutate src/libraries/MathLib.sol --mc MathLibTest --mutation-timeout 30
# -> Mutation Score: 100.0% (111/111 killed); finished in 12.3s
```

i.e. the slowdown is entirely compile-driven (17m49s -> 12s).

### Expected behavior

- Reuse/cache compilation across mutants where possible (a single-source
  mutation only changes one source unit; full re-optimization at
  `optimizer_runs = 999999` per mutant should be avoidable).
- When timeouts dominate, **don't present a score** — warn that the timeout is
  likely too low for the project's compiler settings and suggest reducing
  `optimizer_runs` / disabling `via_ir` for mutation runs.
- Consider applying a faster compile profile for mutation by default, or
  documenting the recommended config.

### Environment

- `forge 1.7.2-dev` (commit `eba698dcf`)
- macOS (arm64), solc 0.8.19

### Additional context

Each mutant currently goes through a fresh `ProjectCompiler::compile()` in
`crates/forge/src/mutation/runner.rs`.

mablr:
## Summary

Fixes misleading mutation testing reports when most mutants time out before reaching a test verdict.

When timed-out mutants dominate, `forge test --mutate` now reports the mutation score as unavailable instead of showing values like `0.0% (0/0 mutants killed)`, and prints guidance to increase `--mutation-timeout` or use a faster mutation profile.

Closes #15102

### DIFF
--- a/crates/forge/src/mutation/mod.rs
+++ b/crates/forge/src/mutation/mod.rs
@@ -181,6 +181,16 @@ impl MutationsSummary {
         if valid_mutants == 0 { 0.0 } else { self.dead.len() as f64 / valid_mutants as f64 * 100.0 }
     }
 
+    /// Mutants that reached a test verdict and can contribute to the score.
+    pub const fn total_evaluated(&self) -> usize {
+        self.dead.len() + self.survived.len()
+    }
+
+    /// Whether the score is useful enough to present as a coverage signal.
+    pub const fn has_reliable_score(&self) -> bool {
+        self.total_evaluated() > 0 && self.timed_out.len() < self.total_evaluated()
+    }
+
     /// Convert to JSON output format.
     ///
     /// Output is sorted deterministically: files in lexicographic order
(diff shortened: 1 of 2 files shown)


crates/forge/src/mutation/reporter.rs
```diff
@@ -65,31 +65,48 @@ impl MutationReporter {
...
+                summary.total_timed_out(),
+                summary.total_evaluated(),
+                duration_str
+            );
+            let _ = sh_println!(
+                "{}",
+                Paint::yellow(
+                    "Timed-out mutants dominate this run. Increase --mutation-timeout or use a \
+                     faster mutation profile, for example lower optimizer_runs or disable via_ir."
+                )
```
mablr, line 106: A follow-up could to be add specific mutation testing flags/config-keys like `--mutation-optimizer-runs`, `--mutation-via-ir` ?

figtracer (dismissed):

crates/forge/src/mutation/mod.rs
```diff
@@ -181,6 +181,16 @@ impl MutationsSummary {
...
     }
 
+    /// Mutants that reached a test verdict and can contribute to the score.
+    pub const fn total_evaluated(&self) -> usize {
+        self.dead.len() + self.survived.len()
+    }
+
+    /// Whether the score is useful enough to present as a coverage signal.
+    pub const fn has_reliable_score(&self) -> bool {
+        self.total_evaluated() > 0 && self.timed_out.len() <= self.total_evaluated()
```
grandizzy: should we check strict `self.timed_out.len() < self.total_evaluated()` to avoid reporting 50/50 split as reliable (e.g. 1 killed + 1 timed out prints 100.0% - 1/1 killed) while half the scoreable population is unknown

grandizzy (commented):
lgtm overall, one nit, pls check

crates/forge/src/mutation/mod.rs
mablr: Fixed in 63b0493. Sorry for force-push.

grandizzy (approved):

stevencartavia (approved):

PR #15105, opened 2026-06-08
fix(forge): mutation dynamic linking compile set

### LINKED ISSUE #15101: `forge test --mutate`: all mutants silently reported as INVALID (0% score) for certain --match-contract selections; underlying compile error is swallowed
### Component

Forge (`forge test --mutate`)

### Describe the bug

When the test selection passed to mutation testing includes certain test
contracts, **every** mutant is classified as `INVALID` ("mutation produced a
compilation error") and the run reports a meaningless **Mutation Score: 0.0%
(0/0)** — even though the exact same mutations are valid and killable with a
slightly different selection, and the project compiles and tests green normally.

The per-mutant compile is scoped to the file set from
`TestArgs::get_sources_to_compile()` (with `dynamic_test_linking` force-enabled
for mutation). For some selected test contracts this scoped compile returns
`Err`, and `crates/forge/src/mutation/runner.rs` maps that `Err` directly to
`MutationResult::Invalid` **without surfacing the compiler error**. The result
is a silent, misleading 0% score with no way to tell a tooling/compile failure
apart from genuinely non-compiling mutants.

### To Reproduce

```bash
git clone --recurse-submodules https://github.com/morpho-org/morpho-blue
cd morpho-blue

# BROKEN: includes MorphoStorageLibTest -> 92/92 INVALID, score 0.0%
FOUNDRY_PROFILE=no_via_ir FOUNDRY_OPTIMIZER_RUNS=200 \
  forge test --mutate src/libraries/SharesMathLib.sol \
  --mc "SupplyIntegrationTest|WithdrawIntegrationTest|MorphoStorageLibTest" \
  --mutation-timeout 50

# WORKS: drop MorphoStorageLibTest -> mutants killed/survived correctly (~95%)
FOUNDRY_PROFILE=no_via_ir FOUNDRY_OPTIMIZER_RUNS=200 \
  forge test --mutate src/libraries/SharesMathLib.sol \
  --mc "SupplyIntegrationTest|WithdrawIntegrationTest" \
  --mutation-timeout 50
```

The selected contracts run fine as a normal test (no mutation):

```bash
FOUNDRY_PROFILE=no_via_ir forge test \
  --mc "SupplyIntegrationTest|WithdrawIntegrationTest|MorphoStorageLibTest"
# -> all green
```

Other selections also trigger it (e.g. any allowlist including
`MorphoStorageLibTest`, and at least one of the
`Authorization`/`Callbacks`/`OnlyOwner` integration tests); `--no-match-contract`
denylists that leave those contracts in the set hit it too. `MathLibTest`,
`MarketParamsLibTest`, `ExtSloadIntegrationTest`, etc. do **not** trigger it.

### Broken output

```
╭──────────┬───────────┬────────────╮
│ Status   ┆ # Mutants ┆ % of Total │
╞══════════╪═══════════╪════════════╡
│ Survived ┆ 0         ┆ 0.0%       │
│ Killed   ┆ 0         ┆ 0.0%       │
│ Invalid  ┆ 92        ┆ 100.0%     │
│ Skipped  ┆ 0         ┆ 0.0%       │
╰──────────┴───────────┴────────────╯
Mutation Score: 0.0% (0/0 mutants killed)
```

### Expected behavior

- Mutants should be classified identically regardless of which (compiling,
  passing) test contracts are selected; valid mutations must not be reported as
  `INVALID`.
- If a per-mutant compile genuinely fails, the compiler error should be surfaced
  (at least under `-v`/`-vvv`), so an infra/tooling compile failure is
  distinguishable from a non-typechecking mutation.
- A run that produces `0/0` real results should not be presented as a `0.0%`
  score; it should warn that no mutant was actually evaluated.

### Environment

- `forge 1.7.2-dev` (commit `eba698dcf`, master @ the native-symbolic merge)
- macOS (arm64), solc 0.8.19

### Additional context

Relevant code: `TestArgs::get_sources_to_compile()` in
`crates/forge/src/cmd/test/mod.rs` and the `Err(_) => MutationResult::Invalid`
mapping in `crates/forge/src/mutation/runner.rs` (`compile_and_test`).
`config.dynamic_test_linking` is force-enabled for mutation runs, which may be
implicated in the scoped-compile failure.

mablr:
## Summary

Fix mutation testing with dynamic test linking when the selected test set depends on helper artifacts outside the initially selected sources.

Closes #15101 

## Details

- Expands the mutation execution compile set to include graph sources and link-reference artifacts.
- Normalizes project-relative library paths in the linker so absolute/canonical paths still match link references.
- Adds debug logging for mutant compile/test errors that are classified as invalid.
- Adds a regression test covering selected mutation tests that depend on dynamically linked helper artifacts.


crates/forge/src/mutation/orchestrator.rs
```diff
@@ -130,15 +130,33 @@ pub async fn run_mutation_testing(
 ) -> Result<MutationRunResult> {
     let num_workers = mutation_config.effective_workers();
     let json_output = mutation_config.json_output;
+    let mut selected_sources_relative = BTreeSet::new();
+    selected_sources_relative.extend(mutation_config.selected_sources_relative.iter().cloned());
+
+    for source_id in output.graph().files() {
```
grandizzy: this widens input set to the entire project graph, so a narrow `--match-contract` run on a large repo (exactly the #15101 scenario) goes from compiling a handful of files to compiling everything, that is multiplied by mutant count. That largely erases mutation testing's sparse-compile optimization, so we only need the selected sources plus their transitive link-reference closure.
Proposed approach, pls double check effectiveness
```rust
let mut link_edges: BTreeMap<PathBuf, BTreeSet<PathBuf>> = BTreeMap::new();
for (id, artifact) in output.artifact_ids() {
    let Some(src) = project_relative_path(&config.root, &id.source) else { continue };
    let edges = link_edges.entry(src).or_default();
    for file in artifact.all_link_references().into_keys() {
        if let Some(dep) = project_relative_path(&config.root, Path::new(&file)) {
            edges.insert(dep);
        }
    }
}

let mut selected_sources_relative: BTreeSet<PathBuf> =
    mutation_config.selected_sources_relative.iter().cloned().collect();
let mut queue: Vec<PathBuf> = selected_sources_relative.iter().cloned().collect();
while let Some(src) = queue.pop() {
    if let Some(deps) = link_edges.get(&src) {
        for dep in deps {
            if selected_sources_relative.insert(dep.clone()) {
                queue.push(dep.clone());
            }
        }
    }
}
let selected_sources_relative = selected_sources_relative.into_iter().collect::<Vec<_>>();
```

crates/forge/src/mutation/runner.rs
```diff
@@ -419,7 +419,10 @@ fn test_single_mutant_isolated(
             ) {
                 Ok(true) => MutationResult::Dead,
                 Ok(false) => MutationResult::Alive,
-                Err(_) => MutationResult::Invalid,
+                Err(err) => {
+                    tracing::debug!(?err, "mutant compile/test failed");
```
grandizzy: should we surface this to user for awarness?

crates/forge/src/mutation/runner.rs
```diff
@@ -505,7 +508,8 @@ fn run_compile_and_test_with_timeout(
             let _ = handle.join();
             MutationResult::Alive
         }
-        Ok(Err(_)) => {
+        Ok(Err(err)) => {
+            tracing::debug!(?err, "mutant compile/test failed");
```
grandizzy: same as above

grandizzy (changes requested):
left some comments, please check, thanks!

crates/forge/src/mutation/orchestrator.rs
mablr: reproduced and fixed in 19da14344916fefcb748b9a85c5985f4d37d5251 (added a regression test)

crates/forge/src/mutation/runner.rs
mablr: Not necessary, just added this initially while debugging

mablr: 91b67d2

crates/forge/src/mutation/runner.rs
mablr: 91b67d2

figtracer (approved):

grandizzy (approved):

PR #15107, opened 2026-06-08
fix(cast): make cast logs --query-size a working opt-in

stevencartavia:
- Make `--query-size` opt-in (`Option<u64>`): omitted = single request, set = chunk.
- Chunk proactively over the inclusive range and resolve `latest`/`earliest` tags.
- Bisect failed chunks instead of dropping the last block or swallowing errors.

### DIFF
--- a/crates/cast/src/lib.rs
+++ b/crates/cast/src/lib.rs
@@ -40,7 +40,7 @@ use foundry_common::{
 use foundry_config::Chain;
 use foundry_evm::core::bytecode::InstIter;
 use foundry_primitives::FoundryTxEnvelope;
-use futures::{FutureExt, StreamExt, future::Either};
+use futures::{FutureExt, StreamExt, TryStreamExt, future::Either};
 #[cfg(feature = "optimism")]
 use op_alloy_consensus as _;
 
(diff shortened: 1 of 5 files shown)


crates/cast/src/lib.rs
```diff
@@ -654,98 +654,141 @@ impl<P: Provider<N> + Clone + Unpin, N: Network> Cast<P, N> {
...
-            return Ok(logs);
+        // Only chunk a finite block-number range larger than one chunk; `chunk_size == 0`
+        // disables chunking and falls back to a single request.
+        let Some((from, to)) = self.resolve_block_range(filter).await? else {
+            return self.provider.get_logs(filter).await.map_err(Into::into);
+        };
+        // Empty range.
+        if from > to {
+            return Ok(vec![]);
+        }
```
mablr: We may want to warn user for this ? (ie inverted range) maybe at clap args level?

crates/cast/src/lib.rs
```diff
@@ -654,98 +654,141 @@ impl<P: Provider<N> + Clone + Unpin, N: Network> Cast<P, N> {
...
+    {
+        Box::pin(async move {
+            let range_filter = filter.clone().from_block(from).to_block(to);
+            match provider.get_logs(&range_filter).await {
+                Ok(logs) => Ok(logs),
+                Err(e) => {
+                    // A single block already failed; nothing left to split.
+                    if from >= to {
+                        return Err(e.into());
+                    }
```
mablr, line 785: Seems that it treats auth/rate-limit/network/malformed-response errors as if they were range-limit errors and issues recursive follow-up requests before returning. Could you double check?

crates/cast/tests/cli/main.rs
```diff
@@ -2216,22 +2216,36 @@ casttest!(logs_sig_2, |_prj, cmd| {
...
+        "0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B",
+    ];
+
+    // Baseline: single request over the whole range.
+    let unchunked = cmd.args(args).assert_success().get_output().stdout_lossy();
+
+    // Same query split into 10k-block chunks (6 chunks).
+    cmd.cast_fuse();
+    let chunked =
+        cmd.args(args).args(["--query-size", "10000"]).assert_success().get_output().stdout_lossy();
```
figtracer, line 2243: could we cover the bisection path with a mocked-provider test? this happy-path comparison would still pass if failed chunks stopped splitting or singleton failures were swallowed.

stevencartavia: [5468482](https://github.com/foundry-rs/foundry/pull/15107/commits/5468482161b5d10653be5b82668eb479431de82f)

crates/cast/src/lib.rs
stevencartavia: [5468482](https://github.com/foundry-rs/foundry/pull/15107/commits/5468482161b5d10653be5b82668eb479431de82f)

crates/cast/src/lib.rs
stevencartavia: [5468482](https://github.com/foundry-rs/foundry/pull/15107/commits/5468482161b5d10653be5b82668eb479431de82f)

crates/cast/src/lib.rs
```diff
@@ -2427,6 +2508,63 @@ fn explorer_client(
...
+    async fn bisects_failed_range_and_aggregates_in_order() {
+        let asserter = Asserter::new();
+        asserter.push_failure_msg("query returned more than 10000 results");
+        asserter.push_success(&vec![log_at(0)]);
+        asserter.push_success(&vec![log_at(2)]);
+
+        let provider =
+            ProviderBuilder::<_, _, AnyNetwork>::default().connect_mocked_client(asserter);
+
+        let logs = Cast::get_logs_bisecting(&provider, &Filter::new(), 0, 3).await.unwrap();
```
mablr, line 2592: This test proves the queued responses are aggregated in order, but it does not assert which `eth_getLogs` ranges were actually requested. Since the mock transport returns FIFO responses regardless of params, this could still pass if bisection retried the original `[0,3]` range or split at the wrong boundary. Could we assert the requested filter params for `[0,3]`, `[0,1]`, and `[2,3]` so the test locks in the intended split behavior?

mablr (approved):
Sgtm (pending https://github.com/foundry-rs/foundry/pull/15118 merge to fix CI)

mattsse (approved):
LGTM

PR #15112, opened 2026-06-09
  feat(script): expose tempo wallet session wrapper

0xKarl98:
Step closer to OSS-162 :
"Also expose forge script --session ... as a thin wrapper that does this implicitly."

Exposing forge script --session as a thin wrapper around cast wallet session --for . 

The wrapper creates a temporary scoped access key, runs the reconstructed forge script command with it, and lets cast revoke the session on exit.

It also forwards the required session policy, root signer, RPC, and transport options to the outer cast wallet session process,
strips wrapper-only flags from the inner script command, and clears inherited Tempo/access-key signer environment variables to avoid stale signer resolution.


crates/script/src/lib.rs
```diff
@@ -640,6 +640,463 @@ impl ScriptArgs {
...
+    "ETH_KEYSTORE",
+    "ETH_KEYSTORE_ACCOUNT",
+    "ETH_PASSWORD",
+    "TEMPO_ACCESS_KEY",
+    "TEMPO_ROOT_ACCOUNT",
+];
+
+/// Arguments that make `forge script` a thin wrapper around `cast wallet session --for`.
+#[derive(Clone, Debug, Default, Args)]
+pub struct ScriptWalletSessionArgs {
```
mablr: This file (crates/script/src/lib.rs) is already huge and quite messy, let's extract these tempo things ?

mablr (changes requested):

mattsse (dismissed):
LGTM

crates/script/src/lib.rs
0xKarl98: Have extracted into crates/script/src/wallet_session.rs

mablr (approved):

figtracer (approved):

PR #15124, opened 2026-06-10
feat(verify): submit to Etherscan and Sourcify in one go

manuelwedler:
Closes action item 1 of #10593.

## Motivation

See #10593 for motivation. The main point is making it easier for users to verify at multiple places. Every contract verified at Sourcify is also pushed to Verifier Alliance and Blockscout.

## Solution

forge verify-contract and forge create --verify now automatically also submit to Sourcify alongside the user's chosen verifier (Etherscan/Blockscout/etc). This makes contracts being covered on Verifier Alliance by default.

Changes:
- New VerifyArgs::collect_runs plans a Vec<ProviderRun>. For now: primary selected by cli args + secondary Sourcify. Extensible for supporting to verify on more explorers in the future.
- run() iterates the plan sequentially.
- verify-check keeps single-provider semantics.

## Example run

image

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/forge/tests/cli/verify.rs
mablr, line 460: This snapshot still expects no stdout for the successful Blockscout verification path, but the new secondary Sourcify run prints a job ID to stdout.

crates/forge/src/opts.rs
```diff
@@ -73,11 +73,11 @@ pub enum ForgeSubcommand {
...
 
-    /// Verify smart contracts on Etherscan.
+    /// Verify smart contracts on Etherscan and Sourcify.
     #[command(visible_alias = "v")]
     VerifyContract(VerifyArgs),
 
-    /// Check verification status on Etherscan.
+    /// Check verification status on Etherscan and Sourcify.
     #[command(visible_alias = "vc")]
     VerifyCheck(VerifyCheckArgs),
```
mablr: This still checks a single provider via `effective_type().client(...).check(...)`, but the help text now says “Etherscan and Sourcify”. This is misleading given the PR description says verify-check keeps single-provider semantics; consider leaving this as “Check verification status on the selected verifier” or similar.

crates/forge/tests/cli/verify.rs
```diff
@@ -260,6 +260,48 @@ forgetest!(can_verify_random_contract_sepolia, |prj, cmd| {
...
+                info.etherscan.as_str(),
+                "--verifier",
+                info.verifier.as_str(),
+            ])
+            .assert_success()
+            .get_output()
+            .stderr_lossy();
+
+        assert!(output.contains("Verifying on etherscan"), "Etherscan run missing: {output}");
+        assert!(output.contains("Verifying on sourcify"), "Sourcify run missing: {output}");
```
mablr: Both verify tests assert `output.contains("Verifying on sourcify")`, which is the `sh_status!` (`crates/verify/src/verify.rs:560`) emitted before `provider.verify()` is called. Since Sourcify is required: false, its failure is silently logged as a warning and the test passes regardless of whether the actual submission succeeded.

crates/forge/tests/cli/verify.rs
```diff
@@ -260,6 +260,48 @@ forgetest!(can_verify_random_contract_sepolia, |prj, cmd| {
...
+        let address = utils::parse_deployed_address(deploy_output.as_str())
+            .unwrap_or_else(|| panic!("Failed to parse deployer {deploy_output}"));
+
+        let output = cmd
+            .forge_fuse()
+            .arg("verify-contract")
+            .root_arg()
+            .args([
+                "--chain-id",
+                &info.chain.to_string(),
```
mablr:
```suggestion
                info.chain.as_ref(),
```
to make clippy happy :)

mablr (changes requested):
Thank you! I tested locally and it seems to work fine.

I left some comments, please check.

crates/verify/src/verify.rs
```diff
@@ -543,16 +552,70 @@ impl VerifyArgs {
...
+        )?;
+        let primary_is_sourcify = resolved.is_sourcify();
+        runs.push(ProviderRun {
+            label: resolved,
+            args: self.clone(),
+            provider: primary_provider,
+            required: true,
+        });
+
+        if !primary_is_sourcify {
```
grandizzy: Default Sourcify submission matches the intent of #10593, so no objection to the default itself. Two small follow-ups, neither blocking:
- The opt-out issue participants asked for (`[verification] sourcify = false` / a global `verification = false` for testnet/pre-audit) is a separate action item — fine to land without it, just flagging for tracking.
- This adds a **public** Sourcify submission even when the primary is `--verifier custom` / an explicit `--verifier-url` (private/internal verifier) or a local/dev chain. Consider skipping the auxiliary run in those cases, since public submission there is surprising and will just fail/warn.

crates/verify/src/verify.rs
```diff
@@ -543,16 +552,70 @@ impl VerifyArgs {
...
-            )?
-            .verify(self, context)
-            .await
-            .map_err(|err| wrap_verifier_url_error(err, verifier_url.as_deref(), using_etherscan))
+        let runs =
+            self.collect_runs(chain, etherscan_key.as_deref(), resolved, had_user_verifier_url)?;
+
+        let mut required_err = None;
+        for ProviderRun { label, args, mut provider, required } in runs {
+            sh_status!("\nVerifying on {label}...")?;
```
grandizzy, line 562: **`--watch` makes this sequential, not "in one go".** `forge create --verify` sets `watch: true`, and `provider.verify()` both submits *and* blocks on `check()`. So the order becomes: submit Etherscan → poll for minutes → only then submit Sourcify (and the auxiliary upload is delayed, or skipped entirely if the primary watch fails first). Consider setting `watch = false` on the secondary `args` in `collect_runs`, or splitting submit/watch so all providers submit first and only the required primary is watched.

crates/verify/src/provider.rs
```diff
@@ -251,10 +251,6 @@ impl VerificationProviderType {
                 }
                 // Fall through to branch 5 (Sourcify default) below.
             } else {
-                sh_status!(
-                    "ETHERSCAN_API_KEY is set, defaulting to Etherscan verifier. \
-                     Unset it or pass `--verifier sourcify` (or another provider) to override."
-                )?;
                 return Ok(Box::<EtherscanVerificationProvider>::default());
```
grandizzy, line 269: Good to drop this selection-time status message. For consistency, the two `"Attempting to verify on Sourcify..."` `sh_status!` calls elsewhere in `client()` are now redundant too, since `run()` prints `Verifying on {label}...` before each run. Provider *selection* shouldn't emit execution-status lines — consider removing/moving those as well to avoid double messaging.

grandizzy (commented):
A few additional notes on top of the existing review (mostly orchestration/UX; none are hard blockers).

**`crates/verify/src/etherscan/mod.rs:126` — `std::process::exit(1)` bypasses the new orchestration.** (Couldn't anchor inline since the file isn't in this diff.) When the Etherscan primary hits that branch it kills the process immediately, so the auxiliary Sourcify run never executes and the new `required`/warn flow is skipped. For the "submit to both" goal to actually hold, that path should return `Err(...)` and let `run()` handle it. Pre-existing, but this PR makes it load-bearing.

crates/verify/src/verify.rs
```diff
@@ -600,13 +625,17 @@ impl VerifyArgs {
             required: true,
         });
 
-        if !primary_is_sourcify {
+        // Skip the auxiliary Sourcify submission when the user is clearly using a
+        // non-public setup: an explicit `--verifier custom` or a local/dev chain.
+        let is_private_setup = resolved.is_custom() || is_dev_chain(chain);
```
mablr: This seems to add the public Sourcify auxiliary run when the user supplied an explicit `--verifier-url`, unless they also used `--verifier custom`. That leaves private Blockscout/Etherscan-compatible URLs on public chains exposed to a public Sourcify submission. Maybe `had_user_verifier_url` should be included in the private-setup skip, or limit the auxiliary run to known public verifier configurations?

mablr (commented):
Seems good, just a small comment.

crates/verify/src/verify.rs
manuelwedler: I intentionally didn't add the check of the verifier url, because the presence of a `--verifier-url` option doesn't necessarily mean that it is being verified privately. For example, any Blockscout verification requires setting the `--verifier-url`. See:

https://github.com/manuelwedler/foundry/blob/689df5c7fa384605b17be0d8d5a7afb2cda11c13/crates/verify/src/provider.rs#L238-L245

So adding this check would prevent verifying on Blockscout and Sourcify in one go.

What do you think? I can still add it if you think it's better to be more restrictive here.

mablr: Hmm yes I see, `--verifier-url` is ambiguous, but the risk is asymmetric imo: preserving Blockscout + Sourcify convenience is less important than avoiding an unexpected public source submission for private/internal verifier URLs?

manuelwedler: Okay, I see the point. I added the check for the `--verifier-url`.

mablr (approved):
SGTM

stevencartavia (approved):

PR #15127, opened 2026-06-10
refactor(tempo): centralize fee token defaulting

mablr:
## Motivation

Add a shared Tempo fee-token resolver that preserves explicit overrides and returns the canonical default for known Tempo chains.

Use the resolver in forge script preprocessing, forge create, cast erc20 browser sends, and keychain doctor so PathUSD defaulting is not duplicated across call sites.

Keep keychain doctor diagnostics anchored to a concrete fee token while allowing config-chain resolution to refine the default when available.

Towards OSS-165

### DIFF
--- a/crates/cast/src/cmd/keychain.rs
+++ b/crates/cast/src/cmd/keychain.rs
@@ -24,7 +24,7 @@ use foundry_common::{
     sh_warn, shell,
     tempo::{
         self, KeyType, KeysFile, TEMPO_BROWSER_GAS_BUFFER, WalletType, read_tempo_keys_file,
-        tempo_keys_path,
+        resolve_fee_token, tempo_keys_path,
     },
 };
 use foundry_evm::hardfork::TempoHardfork;
@@ -32,9 +32,9 @@ use foundry_wallets::{WalletOpts, WalletSigner, wallet_browser::signer::BrowserS
 use serde::Deserialize;
 use tempo_alloy::{TempoNetwork, provider::TempoProviderExt};
 use tempo_contracts::precompiles::{
-    ACCOUNT_KEYCHAIN_ADDRESS, IAccountKeychain,
+    ACCOUNT_KEYCHAIN_ADDRESS, DEFAULT_FEE_TOKEN,
     IAccountKeychain::{
-        CallScope, KeyInfo, KeyRestrictions, LegacyTokenLimit, SelectorRule, SignatureType,
+        self, CallScope, KeyInfo, KeyRestrictions, LegacyTokenLimit, SelectorRule, SignatureType,
         TokenLimit,
     },
     ITIP20, PATH_USD_ADDRESS,
@@ -1077,8 +1077,7 @@ struct DoctorContext {
     key_address: Option<Address>,
     #[serde(skip_serializing_if = "Option::is_none")]
     chain_id: Option<u64>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    fee_token: Option<Address>,
+    fee_token: Address,
 }
 
 /// Result of resolving a local registry entry for the doctor.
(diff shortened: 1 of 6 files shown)


crates/script/src/lib.rs
```diff
@@ -308,9 +309,12 @@ impl ScriptArgs {
...
 
-        if evm_opts.networks.is_tempo() && tempo.fee_token.is_none() {
-            tempo.fee_token = Some(PATH_USD_ADDRESS);
-        }
+        // Resolve the fee token based on the chain and Tempo settings.
+        // If no chain is set, use the Tempo mainnet chain when the Tempo network is active.
+        let chain = config
+            .chain
+            .or_else(|| evm_opts.networks.is_tempo().then(|| Chain::from_named(NamedChain::Tempo)));
+        tempo.fee_token = resolve_fee_token(chain, tempo.fee_token);
```
grandizzy: This changes the defaulting condition from "the script is executing in Tempo mode" to "`config.chain` is Tempo, falling back to Tempo mode only when `config.chain` is unset." Since `config.chain` short-circuits the `or_else`, it diverges from master (which keyed purely on `evm_opts.networks.is_tempo()`), with no explicit `tempo.fee_token`:

| `config.chain` | `networks.is_tempo()` | master | this PR |
|---|---|---|---|
| non-Tempo | true | `PATH_USD` | `None` |
| Tempo | false | `None` | `DEFAULT` |

`run_script` dispatches to `TempoEvmNetwork` based on `evm_opts.networks.is_tempo()`, **not** `config.chain`, and `config.chain` can be set independently (`--chain`, `foundry.toml`). So e.g. `forge script --network tempo --chain mainnet` would now silently drop the fee-token default in a Tempo run.

Suggest preserving the original predicate:

```suggestion
        // Resolve the fee token: default only when the active EVM network is Tempo.
        let chain = evm_opts
            .networks
            .is_tempo()
            .then(|| Chain::from_named(NamedChain::Tempo));
        tempo.fee_token = resolve_fee_token(chain, tempo.fee_token);
```

crates/cast/src/cmd/keychain.rs
```diff
@@ -1228,6 +1231,8 @@ async fn run_doctor(
             return finalize_doctor(steps, context);
         }
     };
+    context.fee_token =
+        resolve_fee_token(config.chain, requested_fee_token).unwrap_or(DEFAULT_FEE_TOKEN);
```
grandizzy: This reassignment is a no-op: `resolve_fee_token(config.chain, requested_fee_token).unwrap_or(DEFAULT_FEE_TOKEN)` equals `requested_fee_token.unwrap_or(DEFAULT_FEE_TOKEN)` for all `config.chain` values, which is exactly what `context.fee_token` was already initialized to above.

| `requested_fee_token` | `config.chain` | result |
|---|---|---|
| `Some(x)` | any | `x` |
| `None` | Tempo | `DEFAULT` |
| `None` | non-Tempo / None | `DEFAULT` |

There's no intervening mutation of `requested_fee_token` or `context.fee_token`, so this line can be removed with no behavior change (and the `resolve_fee_token` import on line 27 becomes unused — it's the only use in this file). The PR-body claim that config-chain resolution refines the default doesn't hold here.

grandizzy (commented):
Overall looks good, nice consolidation and the `resolve_fee_token` helper is well tested. Left a couple of comments — one behavior change in `forge script` defaulting and one redundant line in `keychain doctor`.

crates/cast/src/cmd/keychain.rs
mablr: https://github.com/foundry-rs/foundry/pull/15127/commits/a2fb7c2358b990cabb73d6e5353094da6c90a6ab

crates/script/src/lib.rs
mablr: https://github.com/foundry-rs/foundry/pull/15127/commits/a2fb7c2358b990cabb73d6e5353094da6c90a6ab

grandizzy (approved):
lgtm!

stevencartavia (approved):

PR #15128, opened 2026-06-10
fix(anvil): return complete eth_feeHistory when the cache lags the chain head

### LINKED ISSUE #13680: bug: Possible race condition in eth_feeHistory. Method returns incomplete results due to async cache population race
### Component

Anvil

- [x] Foundry
- [x] Foundryup

12.14.0

foundryup: 1.5.0

_No response_

### Operating System

MacOS 26.3.1

### Description
[`eth_feeHistory`](https://github.com/foundry-rs/foundry/blob/9f082215ed1203f5c945df7e632a2e4a9abb2556/crates/anvil/src/eth/api.rs#L1671) can return incomplete or empty results when called shortly after blocks are mined. The fee history cache is populated asynchronously by [`FeeHistoryService`](https://github.com/foundry-rs/foundry/blob/9f082215ed1203f5c945df7e632a2e4a9abb2556/crates/anvil/src/eth/fees.rs#L191), which runs as a separate future inside [`NodeService::poll`](https://github.com/foundry-rs/foundry/blob/9f082215ed1203f5c945df7e632a2e4a9abb2556/crates/anvil/src/service.rs#L64). This creates a race window where the cache is stale relative to the actual chain head.
The pipeline has three decoupled stages:

- `Backend::mine_block()`: stores block + receipts, sends a `NewBlockNotification`, and returns. The fee history cache is not updated at this point.
- `FeeHistoryService`: receives notifications and inserts cache entries, but only when the tokio runtime happens to poll the NodeService task.
- `EthApi::fee_history()`: [reads from the cache](https://github.com/foundry-rs/foundry/blob/9f082215ed1203f5c945df7e632a2e4a9abb2556/crates/anvil/src/eth/api.rs#L1731). Missing entries are silently skipped, so the response can have fewer entries than requested with no error. 

This means any RPC call that lands between stage 1 and stage 2 will observe a stale cache.

### Observed behavior
We have a test that mines 100 blocks, then calls `eth_feeHistory` twice requesting the last 10 blocks:

First call: returns fee history for only 1 block (block 100)
Second call: returns fee history for all 11 blocks (90–100)

To confirm the root cause, we added panic statements in a local fork of Anvil inside `EthApi::fee_history `(replacing the silent skip). The panic fired with:

```sh
fee history cache missing entry for block 91
(requested range 91..=100, cache has blocks: [0, 1, 2, ..., 42])
The service had only processed up to block 42 when the first RPC call arrived, despite the chain head being at block 100.
```

### Reproduction difficulty

Due to the nature of this bug, it's very difficult to write a PoC that fails consistently. The race depends entirely on when the runtime happens to poll FeeHistoryService relative to incoming RPC requests.

### Suggested fix

Insert the fee history cache entry synchronously inside `Backend::mine_block()`, immediately after the block and receipts are stored. This guarantees the cache is always consistent with the chain head before any RPC call can observe the new block.

satyakwok:
Fixes #13680.

## Problem

`eth_feeHistory` reads from a cache that `FeeHistoryService` populates asynchronously (it only advances when the node task is polled). An RPC call landing between mining a block and the service processing its `NewBlockNotification` observes a stale cache. The reader loop did `if let Some(block) = cache.get(&n)` with no `else`, so missing blocks were dropped silently — the response could contain fewer entries than requested, with no error.

As reported, a request for blocks `91..=100` returned only block `100` while the service had processed up to block `42`.

## Fix

Compute any missing entry on demand instead of skipping it. The per-block computation in `FeeHistoryService::create_cache_entry` is extracted verbatim into a shared `create_fee_history_cache_item(...)`, which `eth_feeHistory` now calls on a cache miss (and warms the cache for subsequent reads). The async service stays a pure optimization; correctness no longer depends on its timing.

This also fixes a latent inconsistency where `oldest_block` (= `lowest`) could disagree with the first returned entry when a low block was skipped.

## Scope

- Localized to the RPC reader; the mining path is untouched.
- `eth_maxPriorityFeePerGas` also reads this cache; populating it synchronously at mine time would cover that too, but that touches the mining path and is left as a possible follow-up.
- One method-level bound `where N::ReceiptEnvelope: TxReceipt<Log = Log>` is added to `fee_history`, satisfied by all concrete `EthApi` instantiations (the `FeeHistoryService` already requires it).

## Tests

`gas::test_fee_history_complete_when_cache_lags` burst-mines 100 blocks so the cache trails the head, then asserts the response covers the full requested range. Note: the underlying race is timing-dependent (a deterministically-failing PoC is hard, as noted in the issue), so this guards the completeness invariant rather than reproducing exact timing; it passes deterministically with the fix and is not flaky.

Verified locally: `cargo check -p anvil`, `cargo test -p anvil --test it test_fee_history_complete_when_cache_lags` (pass), `cargo fmt --check`, `cargo clippy -p anvil --tests` (clean).


crates/anvil/src/eth/api.rs
```diff
@@ -1143,29 +1149,51 @@ impl<N: Network> EthApi<N> {
...
+                        let header = b.header;
+                        let hash = header.hash_slow();
+                        let (item, block_number) = create_fee_history_cache_item(
+                            hash,
+                            &header,
+                            &storage_info,
+                            blob_params,
+                        );
+                        if block_number.is_some() {
+                            self.fee_history_cache.lock().insert(n, item.clone());
```
mablr: This warms `fee_history_cache` on cache misses but bypasses `FeeHistoryService::insert_cache_entry`, which is the only path that enforces `MAX_FEE_HISTORY_CACHE_SIZE`. I think repeated `anvil_mine` bursts followed by `eth_feeHistory` calls can keep inserting on-demand entries while the async service is still lagging, so the shared cache can grow past its intended bound. Consider reusing the same bounded insertion logic for the RPC fallback.

satyakwok: Yep, you're right. The fallback inserted straight into the map and skipped the `MAX_FEE_HISTORY_CACHE_SIZE` trim, so repeated on-demand inserts could push the cache past its limit while the service is still behind.

Pulled the bounded insert into a shared `insert_fee_history_cache_item` and routed both the service and the `eth_feeHistory` fallback through it, so they trim the same way. Thanks for catching it.

mablr: I think fallback can still return a short/misaligned response. On a cache miss, `let Some(b) = self.backend.get_block(n) else { continue };` (L1166) silently skips the block, the exact bug being fixed. `oldest_block` stays lowest but per-block arrays drop the entry, shortening/misaligning the response (future block numbers, reset/revert races, cross-fork ranges). Better to return `FeeHistoryError::BlockNotFound` or an error ?

crates/anvil/src/eth/fees.rs
```diff
@@ -242,111 +242,140 @@ where
...
+        let mut cache = cache.lock();
+        cache.insert(block_number, item);
+
+        // adhere to cache limit
+        let pop_next = block_number.saturating_sub(fee_history_limit);
+
+        let num_remove = (cache.len() as u64).saturating_sub(fee_history_limit);
+        for num in 0..num_remove {
+            let key = pop_next - num;
+            cache.remove(&key);
```
mablr: Cache trim can underflow & doesn't robustly enforce the limit

`pop_next - num` is non-saturating -> can panic in debug / wrap in release.

The trim also assumes monotonic sequential inserts; out-of-order fallback inserts can leave `len > fee_history_limit`. 

Note: the underflow itself seems pre-existing, but the new fallback makes out-of-order inserts reachable.

crates/anvil/src/eth/api.rs
```diff
@@ -1102,7 +1106,10 @@ impl<N: Network> EthApi<N> {
         block_count: U256,
         newest_block: BlockNumber,
         reward_percentiles: Vec<f64>,
-    ) -> Result<FeeHistory> {
+    ) -> Result<FeeHistory>
+    where
+        N::ReceiptEnvelope: TxReceipt<Log = alloy_primitives::Log>,
```
mablr, line 1178: I think `TxReceipt` bound is enough.

crates/anvil/src/eth/fees.rs
```diff
@@ -242,111 +242,140 @@ where
...
+    // percentile list from 0.0 to 100.0 with a 0.5 resolution.
+    // this will create 200 percentile points
+    let reward_percentiles: Vec<f64> = {
+        let mut percentile: f64 = 0.0;
+        (0..=200)
+            .map(|_| {
+                let val = percentile;
+                percentile += 0.5;
+                val
+            })
```
mablr: The reward_percentiles `Vec` is allocated per block regardless of whether `reward_percentiles` was requested.

crates/anvil/src/eth/api.rs
```diff
@@ -1143,29 +1150,56 @@ impl<N: Network> EthApi<N> {
...
-                            let index = ((p.round() / 2f64) * 2f64) * resolution_per_percentile;
-                            let reward = block.rewards.get(index as usize).map_or(0, |r| *r);
-                            block_rewards.push(reward);
-                        }
-                        rewards.push(block_rewards);
+                // Prefer the cache, but the `FeeHistoryService` populates it asynchronously and can
+                // lag the chain head. Rather than silently dropping a missing block from the
+                // response (which yields fewer entries than requested), compute it on demand with
+                // the same logic and warm the cache for subsequent reads.
+                let block = match self.fee_history_cache.lock().get(&n).cloned() {
```
mablr: This lock in for-loop is probably avoidable

crates/anvil/src/eth/api.rs
satyakwok: Good catch. Fixed in e4b5ba9: the fallback now returns `FeeHistoryError::BlockNotFound` when a block cannot be loaded instead of `continue`, so the range cannot come back short or misaligned.

crates/anvil/src/eth/fees.rs
satyakwok: Fixed in e4b5ba9: replaced the index math with `pop_first` until `len <= limit` — saturating and order-independent, so out-of-order fallback inserts can no longer underflow or leave the cache over the bound.

crates/anvil/src/eth/fees.rs
satyakwok: Fixed in e4b5ba9: hoisted the percentiles to a `LazyLock` static so they are built once instead of per block.

crates/anvil/src/eth/api.rs
satyakwok: Fixed in e4b5ba9: the requested range is now read from the cache under a single lock; misses are computed without holding it and warmed into the cache in one batched insert afterwards.

crates/anvil/src/eth/api.rs
satyakwok: Relaxing only this signature does not compile: `create_fee_history_cache_item` calls `StorageInfo::receipts`, which (through `Backend::mined_receipts`) carries the `TxReceipt<Log = alloy_primitives::Log>` bound. Neither `receipts` nor `mined_receipts` actually uses `Log` — they just clone the receipts — so the constraint can be dropped, but only by also relaxing those two on the backend side (`backend/info.rs` and `backend/mem/mod.rs`). Happy to do that here; just confirming you are OK widening the diff into the backend before I touch it.

crates/anvil/src/eth/api.rs
```diff
@@ -1143,29 +1150,67 @@ impl<N: Network> EthApi<N> {
...
+            // Compute any missing entries on demand (without holding the lock), using the same
+            // logic as the service. A block that can't be found is a hard error rather than a
+            // silently dropped entry, which would otherwise shorten or misalign the response.
+            let mut warmed: Vec<(u64, FeeHistoryCacheItem)> = Vec::new();
+            for (idx, n) in (lowest..=highest).enumerate() {
+                if items[idx].is_none() {
+                    let block = self
+                        .backend
+                        .get_block(n)
+                        .ok_or(FeeHistoryError::BlockNotFound(BlockNumber::Number(n)))?;
```
figtracer: this can now hit any missing block in the range, but the error says "newest block". could we make that wording generic before reusing it here?

figtracer (commented):
one wording issue.

crates/anvil/src/eth/api.rs
satyakwok: Good point — generalized it in c35d9c9: the message is now `could not find block number requested: {0}` (dropped "newest"), since this path can hit any missing block in the range, not just the newest one.

satyakwok:
Heads up — CI hasn't run on this yet; since it's from a fork it needs a maintainer to approve the workflow. Everything from the reviews is addressed (mablr's points and the wording nit), so it should be ready for a CI run and another look whenever someone has time. Thanks!

satyakwok:
The `test all` failures are all pre-existing fork/RPC-secret tests, unrelated to this change:

- `fork::can_deploy_greeter_on_fork`, `fork::can_override_fork_chain_id` — empty `ETH_SEPOLIA_RPC`/fork backend
- `script::can_broadcast_script_skipping_simulation` — empty `ARBITRUM_RPC` (`-32602 could not find block 0x1826e6d`)
- `repl::chisel_can_run_with_live_logs_flag` — flaky pexpect timeout

The fee-history tests added here pass: `test_fee_history_complete_when_cache_lags`, `test_can_use_fee_history`, `test_fork_eth_fee_history`. @mablr would a re-run with secrets or an admin-merge be possible?

mablr:
Triggered CI, I'll try to make a final review pass by the end of the day.

satyakwok:
Thanks @mablr , I’ll watch for any CI/review feedback.

crates/anvil/src/eth/api.rs
mablr: okay we can keep as is

crates/anvil/src/eth/api.rs
```diff
@@ -1143,29 +1150,67 @@ impl<N: Network> EthApi<N> {
...
+                let mut cache = self.fee_history_cache.lock();
+                for (block_number, item) in warmed {
+                    cache.insert(block_number, item);
+                }
+                while cache.len() as u64 > self.fee_history_limit {
+                    cache.pop_first();
+                }
+            }
+
+            for item in items.into_iter().flatten() {
```
mablr: After the miss-fill loop, every entry in `items` should be `Some`: cache hits start as `Some`, misses either return `BlockNotFound` via `?` or assign `items[idx] = Some(item)`. Using `.flatten()` could make a future invariant break silently shorten the response again. Maybe a restructuring is needed so this loop consumes `FeeHistoryCacheItem` directly.

mablr (commented):
Last small nit.

I guess CI failures are not related to your changes 🤔

satyakwok:
Looks unrelated to this change.

the failing fork test needs an RPC secret. I’ll address the `.flatten()` nit now.

crates/anvil/src/eth/api.rs
satyakwok: Fixed in 7d7156e: the loop now consumes `FeeHistoryCacheItem` directly. `items` is built as a `Vec<FeeHistoryCacheItem>` in a single pass, each entry is either the cache hit or the on-demand block, so there is no `Option` left to flatten and the response can't silently shorten if that invariant changes later.

mablr (dismissed):
After having a deeper look to failed tests (`can_deploy_greeter_on_fork`, `can_override_fork_chain_id`, `can_broadcast_script_skipping_simulation`). I see it's related to the fact that PR changed `eth_feeHistory` to compute cache misses locally and hard-error if `self.backend.get_block(n)` is missing. In fork mode, that breaks fee-history requests whose range crosses the fork boundary.

The likely fix direction is to make `eth_feeHistory` fork-range aware:

- If `highest <= fork.block_number()`, keep delegating the whole request to fork.fee_history.
- If `lowest <= fork.block_number() < highest`, either split/merge fork + local fee-history entries, or delegate a compatible request to the fork provider for the fork-origin portion and compute only post-fork local entries.
- Do not call `backend.get_block(n)` for fork-origin blocks.

satyakwok:
Thanks for the review, that makes sense.

I see the issue now. The existing `predates_fork_inclusive` guard only checks the newest block, so ranges that cross the fork boundary can incorrectly enter the local path and hard-error on pre-fork blocks.

I’ll make `eth_feeHistory` fork-range aware: ranges fully at or below the fork block will continue delegating to `fork.fee_history`, crossing ranges will be split between the fork provider and local fee history, and local `backend.get_block(n)` calls will be avoided for fork-origin blocks.

I’ll push a fix for this.

satyakwok:
Pushed in 9124f81 with a regression test that crosses the fork boundary. It fails on the old code (BlockNotFound on the pre-fork block) and passes with the fix.

crates/anvil/src/eth/api.rs
```diff
@@ -1142,30 +1149,107 @@ impl<N: Network> EthApi<N> {
...
+            let count_pre = fork_block - lowest + 1;
+            let pre = fork
+                .fee_history(count_pre, BlockNumber::Number(fork_block), &reward_percentiles)
+                .await
+                .map_err(BlockchainError::AlloyForkProvider)?;
+            // These per-block arrays cover [lowest..=fork_block]; drop the trailing next-block
+            // base-fee entry from each since the first post-fork block is computed locally below.
+            let take = count_pre as usize;
+            response.base_fee_per_gas.extend(pre.base_fee_per_gas.into_iter().take(take));
+            response.base_fee_per_blob_gas.extend(pre.base_fee_per_blob_gas.into_iter().take(take));
```
mablr: When the fork provider omits pre-Cancun blob fee arrays for compatibility, this merge may possibly return partially filled non-empty `base_fee_per_blob_gas` / `blob_gas_used_ratio` arrays ? The pre-fork segment contributes 0 entries, then the local segment appends entries. The solution is probably zero-padding the pre-fork blob arrays to `count_pre` / `count_pre + 1` before appending local data, and extend the crossing test to assert blob array lengths too.

satyakwok: Fixed in `03fb7fd`.
pad the pre-fork blob arrays to `count_pre` so they line up with the gas arrays, plus blob-length asserts in the crossing test.

crates/anvil/src/eth/api.rs
```diff
@@ -1142,30 +1149,113 @@ impl<N: Network> EthApi<N> {
...
 
+        // The early return above only delegates when the *newest* block predates the fork, but a
+        // range can dip below the fork block while its newest block is post-fork. Local storage
+        // has no pre-fork blocks, so resolving them via `backend.get_block` below would fail.
+        // Serve the pre-fork portion from the fork provider and compute only post-fork locally.
+        let local_lowest = if let Some(fork) = self.get_fork()
+            && fork.predates_fork_inclusive(lowest)
+        {
+            let fork_block = fork.block_number();
+            let count_pre = fork_block - lowest + 1;
```
mablr: The straddle path reads the fork block indirectly via `predates_fork_inclusive(lowest)`, then reads it again with `fork.block_number()` before doing `fork_block - lowest + 1`. If `anvil_reset` changes the fork config between those reads, this subtraction can underflow or compute a range inconsistent with the branch condition. It should be: snapshot `fork_block` once and branch on `lowest <= fork_block`.

crates/anvil/src/eth/api.rs
```diff
@@ -1142,30 +1149,113 @@ impl<N: Network> EthApi<N> {
...
+                            hash,
+                            &header,
+                            &storage_info,
+                            blob_params,
+                        );
+                        if let Some(block_number) = block_number {
+                            warmed.push((block_number, item.clone()));
                         }
-                        rewards.push(block_rewards);
+                        item
```
mablr: The miss path confirms `backend.get_block(n)` exists, then `create_fee_history_cache_item` re-fetches the same block and receipts by hash. If that second lookup misses, the helper returns a zero-filled cache item and the RPC returns it. I think a found block should either produce a real row or an error, not synthetic zero data.

crates/anvil/src/eth/api.rs
satyakwok: Makes sense. I'll read `fork_block` once and use that for both the branch and `count_pre`, so we don't call `fork.block_number()` again after the check.

crates/anvil/src/eth/api.rs
satyakwok: Agreed. If we already found the block, the cache item should either be built from real data or return an error. I'll remove the zero-filled fallback for that path.

crates/anvil/src/eth/api.rs
satyakwok: the `fork_block` is now read once at the top of the fork branch, and the range decision and `count_pre` both use that single snapshot (`if lowest <= fork_block { let count_pre = fork_block - lowest + 1; … }`). There's no second `fork.block_number()` call after the check, so an `anvil_reset` moving the fork mid-request can't underflow `count_pre` or pick a branch inconsistent with the subtraction anymore.

crates/anvil/src/eth/api.rs
satyakwok: Done in `7386cd8`. When the on-demand path has already confirmed the block exists, `create_fee_history_cache_item` returning `None` (stored block/receipts missing on the second lookup) now maps to `FeeHistoryError::BlockNotFound(n)` via `?` instead of pushing the zero-filled item. So a confirmed block either yields a real row or errors out, it can't return synthetic zero data.

crates/anvil/src/eth/api.rs
```diff
@@ -1142,30 +1149,123 @@ impl<N: Network> EthApi<N> {
         };
         let mut rewards = Vec::new();
 
+        // The early return above only delegates when the *newest* block predates the fork, but a
+        // range can dip below the fork block while its newest block is post-fork. Local storage
+        // has no pre-fork blocks, so resolving them via `backend.get_block` below would fail.
+        // Serve the pre-fork portion from the fork provider and compute only post-fork locally.
+        let local_lowest = if let Some(fork) = self.get_fork() {
```
figtracer: this still reads the fork block from shared fork config after the earlier fork check. if `anvil_reset` moves the fork between those reads, we can skip the early return but then treat the whole request as pre-fork and append an extra local next-block fee. could we snapshot the fork block once and only split when `lowest <= fork_block && fork_block < highest`?

satyakwok: Fixed in 9837133. The fork block is now read once at the top and reused for both the early return and the pre-fork split, so an `anvil_reset` moving the fork mid-request can't clear the early return and then have the split treat the whole range as pre-fork. The split is only reached when the newest block is post-fork, so `fork_block < highest` already holds and it only splits on a genuine crossing range (`lowest <= fork_block < highest`). Also extended the crossing test to assert `oldest_block`.

crates/anvil/src/eth/api.rs
```diff
@@ -1184,30 +1198,122 @@ impl<N: Network> EthApi<N> {
...
+                for (block_number, item) in warmed {
+                    cache.insert(block_number, item);
+                }
+                while cache.len() as u64 > self.fee_history_limit {
+                    cache.pop_first();
+                }
+            }
+
+            for item in items {
+                response.base_fee_per_gas.push(item.base_fee);
```
grandizzy, line 1312: This fallback now resolves the requested per-block entries for `[lowest..=highest]`, but the trailing fee appended below still always comes from the current chain head. For historical requests like `eth_feeHistory(1, Number(n), [])` where `n < latest`, `self.backend.fees().base_fee()` is the next fee for `best_number + 1`, not for `highest + 1`, so the final `base_fee_per_gas[count]` entry becomes misaligned with the requested range. Same applies to `base_fee_per_blob_gas` below.

Could we only use `self.backend.fees()` when `highest == self.backend.best_number()`, and otherwise append the base/blob fee from block `highest + 1` (or the fork-provider segment when the range is delegated/split)? A regression test with changing base fees and `newest_block = Number(old_block)` would catch this.

crates/anvil/src/eth/api.rs
```diff
@@ -1196,46 +1223,146 @@ impl<N: Network> EthApi<N> {
...
+        let local_lowest = if let (Some(fork), Some(fork_block)) = (fork.as_ref(), fork_block) {
+            if lowest <= fork_block {
+                let count_pre = fork_block - lowest + 1;
+                let pre = fork
+                    .fee_history(count_pre, BlockNumber::Number(fork_block), &reward_percentiles)
+                    .await
+                    .map_err(BlockchainError::AlloyForkProvider)?;
+                // These per-block arrays cover [lowest..=fork_block]; drop the trailing next-block
+                // base-fee entry since the first post-fork block is computed locally below.
+                let take = count_pre as usize;
```
figtracer: upstream may legally return fewer blocks than `count_pre`, but this ignores `pre.oldest_block` and assumes a full segment. i reproduced a `4..10` request where upstream returned `5..10`: the merged response still reported oldest `4` and only 9 gas rows for `block_count = 10`. can we derive the merged start/count from `pre.oldest_block` and cover a shortened response?

mablr: https://github.com/foundry-rs/foundry/pull/15128/commits/8135e1426d24b668f21c75d5fade0784943ab9c2

figtracer (approved):

stevencartavia (approved):

mablr:
Thank you @satyakwok !

PR #15130, opened 2026-06-10
feat(tempo): show resolved fee token + clean-ups

mablr:
## Motivation

Materialize Tempo default fee tokens in cast transaction building and print the resolved fee token before sending, signing, or emitting transactions across cast, forge create, and script broadcast paths.

Add no-RPC labels for well-known Tempo fee tokens and centralize the local AlphaUSD, BetaUSD, and ThetaUSD addresses for reuse in EVM core and Anvil.

Towards OSS-165

### DIFF
--- a/crates/cast/src/cmd/mktx.rs
+++ b/crates/cast/src/cmd/mktx.rs
@@ -15,7 +15,9 @@ use foundry_cli::{
     opts::{EthereumOpts, TransactionOpts},
     utils::{LoadConfig, maybe_print_resolved_lane, resolve_lane},
 };
-use foundry_common::{FoundryTransactionBuilder, provider::ProviderBuilder};
+use foundry_common::{
+    FoundryTransactionBuilder, provider::ProviderBuilder, tempo::print_fee_token_selection,
+};
 use std::{path::PathBuf, str::FromStr};
 use tempo_alloy::TempoNetwork;
 
@@ -178,6 +180,7 @@ impl MakeTxArgs {
             if let Some(sponsor) = &tempo_sponsor {
                 sponsor.attach_and_print::<N>(&mut tx, from).await?;
             }
+            print_fee_token_selection(tx.fee_token())?;
             let raw_tx = hex::encode_prefixed(tx.build_unsigned()?.encoded_for_signing());
 
             print_scalar(raw_tx)?;
@@ -192,6 +195,7 @@ impl MakeTxArgs {
             if let Some(sponsor) = &tempo_sponsor {
                 sponsor.attach_and_print::<N>(&mut tx, config.sender).await?;
             }
+            print_fee_token_selection(tx.fee_token())?;
             let signed_tx = provider.sign_transaction(tx).await?;
 
             print_scalar(signed_tx)?;
@@ -210,6 +214,7 @@ impl MakeTxArgs {
         if let Some(sponsor) = &tempo_sponsor {
             sponsor.attach_and_print::<N>(&mut tx, from).await?;
         }
+        print_fee_token_selection(tx.fee_token())?;
 
         let tx = tx.build(&EthereumWallet::new(signer)).await?;
 
--- a/crates/cast/src/cmd/mktx.rs
+++ b/crates/cast/src/cmd/mktx.rs
@@ -16,7 +16,7 @@ use foundry_cli::{
     utils::{LoadConfig, maybe_print_resolved_lane, resolve_lane},
 };
 use foundry_common::{
-    FoundryTransactionBuilder, provider::ProviderBuilder, tempo::print_fee_token_selection,
+    FoundryTransactionBuilder, provider::ProviderBuilder, tempo::print_resolved_fee_token_selection,
 };
 use std::{path::PathBuf, str::FromStr};
 use tempo_alloy::TempoNetwork;
@@ -138,6 +138,7 @@ impl MakeTxArgs {
             .with_code_sig_and_args(code, sig, args)
             .await?
             .with_blob_data(blob_data)?;
+        let chain = tx_builder.chain();
 
         // If --tempo.print-sponsor-hash was passed, build the tx, print the hash, and exit.
         if print_sponsor_hash {
@@ -180,7 +181,7 @@ impl MakeTxArgs {
             if let Some(sponsor) = &tempo_sponsor {
                 sponsor.attach_and_print::<N>(&mut tx, from).await?;
             }
-            print_fee_token_selection(tx.fee_token())?;
+            print_resolved_fee_token_selection(Some(chain), tx.fee_token())?;
             let raw_tx = hex::encode_prefixed(tx.build_unsigned()?.encoded_for_signing());
 
             print_scalar(raw_tx)?;
@@ -195,7 +196,7 @@ impl MakeTxArgs {
             if let Some(sponsor) = &tempo_sponsor {
                 sponsor.attach_and_print::<N>(&mut tx, config.sender).await?;
             }
-            print_fee_token_selection(tx.fee_token())?;
+            print_resolved_fee_token_selection(Some(chain), tx.fee_token())?;
             let signed_tx = provider.sign_transaction(tx).await?;
 
             print_scalar(signed_tx)?;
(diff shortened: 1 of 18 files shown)


stevencartavia (dismissed):

crates/cast/src/tx.rs
```diff
@@ -434,6 +434,12 @@ where
         // Apply gas, value, fee, and network-specific options.
         tx_opts.apply::<N>(&mut tx, legacy);
 
+        // Default to the chain's canonical fee token unless the user supplied one, so the
+        // resolved selection is materialized (and reported) consistently across cast commands.
+        if let Some(fee_token) = resolve_fee_token(Some(chain), tx.fee_token()) {
+            tx.set_fee_token(fee_token);
+        }
```
grandizzy: This changes the tx, not just the output. Setting `fee_token = PATH_USD` isn't wire-neutral: in `tempo-transaction.rs`, `None` encodes as `EMPTY_STRING_CODE` while `Some(addr)` encodes the address, so for unsponsored txs the **signing hash and tx hash change**.

More importantly, the node's `get_fee_token` precedence is `explicit → setUserToken → stored user preference → TIP20/DEX inference → DEFAULT`. By materializing `PATH_USD`, we short-circuit the user's on-chain preference — which anvil's own genesis sets (`Alice→AlphaUSD`, etc.). So `cast send --from alice` would now pay in PathUSD instead of AlphaUSD.

Could we print the *resolved* token without mutating the request when the user didn't pass `--tempo.fee-token`? Otherwise the "Paying gas in PathUSD" message is self-fulfilling rather than reporting what the node would actually pick.

crates/evm/core/src/tempo.rs
```diff
@@ -40,7 +43,8 @@ pub use tempo_precompiles::{
 };
 
 /// All well-known TIP20 fee token addresses on Tempo networks.
-pub const TEMPO_TIP20_TOKENS: &[Address] = &[PATH_USD_ADDRESS];
+pub const TEMPO_TIP20_TOKENS: &[Address] =
+    &[PATH_USD_ADDRESS, ALPHA_USD_ADDRESS, BETA_USD_ADDRESS, THETA_USD_ADDRESS];
```
grandizzy: Fork-mode warmup fabricates tokens on mainnet. This constant feeds the warmup in `evm/core/src/evm/tempo.rs`, which writes sentinel `0xef` code to each address. Alpha/Beta/Theta don't exist on **mainnet (`presto.json`)** — only on moderato/dev — so forking mainnet would fabricate code at `…0001/0002/0003`, making `is_tip20()` return true and `EXTCODE*` non-zero for tokens that are empty on-chain.

Suggest splitting: keep all four for local genesis/labels, but warm only `PATH_USD` (exists everywhere) in fork mode.

grandizzy (dismissed):
Looks good overall — couple of things to check.

0xKarl98:
Have nothing to against it except for which have been proposed by @grandizzy

crates/evm/core/src/tempo.rs
mablr: mistake reverted 2e1d4a18226f9d9000029549861edc80bd5cb69c

crates/cast/src/tx.rs
mablr: reverted d7703dff99c5efb810f0a468fbdd2114514d4e7f

mablr: Fee token is printed at `crates/cast/src/cmd/send.rs:481` anyway

crates/cast/src/cmd/mktx.rs
```diff
@@ -178,6 +180,7 @@ impl MakeTxArgs {
             if let Some(sponsor) = &tempo_sponsor {
                 sponsor.attach_and_print::<N>(&mut tx, from).await?;
             }
+            print_fee_token_selection(tx.fee_token())?;
```
figtracer: this is still a no-op for the default tempo fee token: after the cast builder defaulting was reverted, fee_token() is only Some when the user passed --tempo.fee-token. on tempo chains without an explicit fee token, this won’t print PathUSD as intended.

figtracer (dismissed):
lgtm

crates/cast/src/cmd/send.rs
```diff
@@ -477,6 +478,7 @@ where
     N::ReceiptResponse: UIfmt + UIfmtReceiptExt,
 {
     let cast = CastTxSender::new(provider);
+    print_fee_token_selection(tx.fee_token())?;
```
grandizzy: This only prints when `--tempo.fee-token` was passed. On the default `cast send` path `tx.fee_token()` is `None` (it's not materialized after the `tx.rs` revert), so this is a no-op — the user never sees "Paying gas in PathUSD" even though the tx pays in PathUSD, which defeats the goal of showing the resolved token here.

Suggest resolving the default for display only, e.g. a `print_resolved_fee_token(Some(chain), tx.fee_token())` helper that prints `resolve_fee_token(chain, fee_token)`. Same applies to the access-key path below (518), `mktx.rs`, `batch_mktx.rs`, `tip20/mod.rs`, and `keychain.rs`.

crates/cast/src/cmd/vaddr/create.rs
```diff
@@ -182,11 +183,19 @@ async fn register(
     tempo::print_expires(expires_at)?;
     tx_opts.apply::<TempoNetwork>(&mut tx, chain.is_legacy());
 
+    // Default to the chain's canonical fee token unless the user supplied one, so the resolved
+    // selection is materialized (and reported) consistently with the other cast send paths.
+    if let Some(fee_token) = resolve_fee_token(Some(chain), tx.fee_token()) {
```
grandizzy: This is the same mutation that was reverted in `cast/src/tx.rs` reintroduced here. Setting `fee_token = PathUSD` isn't wire-neutral — it changes the signing payload and tx hash and flips the tx to the AA type — and it short-circuits the node's fee-token precedence (explicit → setUserToken → stored preference → inference → default), so `--from alice` would pay PathUSD instead of her on-chain AlphaUSD preference.

Please resolve for display only instead of calling `set_fee_token`. The same pre-existing pattern in `cast erc20` and `forge create` should be cleaned up too.

crates/common/src/tempo/mod.rs
```diff
@@ -63,6 +73,33 @@ pub fn resolve_fee_token(
...
+/// Formats a Tempo fee-token selection for command output.
+pub fn format_fee_token_selection(fee_token: Address) -> String {
+    match known_fee_token_symbol(fee_token) {
+        Some(symbol) => format!("Paying gas in {symbol} ({fee_token})"),
+        None => format!("Paying gas in {fee_token}"),
+    }
+}
+
+/// Prints the selected Tempo fee token when one is set.
+pub fn print_fee_token_selection(fee_token: Option<Address>) -> Result<()> {
```
grandizzy, line 96: Consider adding a display-only resolver so all paths are consistent and none mutate the request:

```rust
/// Prints the fee token that will pay for gas, resolving the chain default for display only.
pub fn print_resolved_fee_token(chain: Option<Chain>, fee_token: Option<Address>) -> Result<()> {
    print_fee_token_selection(resolve_fee_token(chain, fee_token))
}
```

Then call this everywhere instead of mixing `print_fee_token_selection` (some paths) with `resolve_fee_token` + `set_fee_token` (others).

grandizzy (dismissed):
Two issues with the fee-token display:

1. The main send paths (`cast send` / `mktx` / `batch-mktx` / `tip20` / `keychain`) never show the default fee token, because it isn't materialized: `tx.fee_token()` is `None` unless the user passed `--tempo.fee-token`, so `print_fee_token_selection` is a no-op on the default path. This defeats the PR's goal of showing the resolved token before sending.

2. `cast vaddr create` materializes the default into the request (and the pre-existing paths in `cast erc20` / `forge create` do the same). Setting `fee_token = PathUSD` is not wire-neutral and overrides the node's on-chain fee-token preference — the same concern raised above on `tx.rs`.

Recommend a single display-only resolver used by all paths, and removing the `set_fee_token` materializations.

crates/common/src/tempo/mod.rs
mablr: Done

`forge script` still uses the pre-existing fee-token materialization path because that value is part of script execution/broadcast config, not just command output. I kept this PR scoped to removing newly introduced/default materialization in cast/forge create send paths and adding display-only printing there. I'll unify fee token resolution separately as a follow-up.

figtracer (approved):

grandizzy (approved):

PR #15133, opened 2026-06-11
fix(`anvil`): commit transactionsRoot over canonical EIP-4844 tx encoding

### LINKED ISSUE #15132: bug(`anvil`): mined blocks commit `transactionsRoot` over pooled (sidecarful) EIP-4844 encoding, not canonical execution payload
### Component

Anvil

master (`c41910099`); the code path is unchanged on current stable releases with Cancun+ hardforks

### Operating System

Linux

### Describe the bug

When Anvil mines a block containing an EIP-4844 blob transaction (submitted with its sidecar, as required), the block's `transactionsRoot` — and therefore the **block hash** — is computed over the **pooled/networking encoding** of the blob transaction (`0x03 || rlp([tx_payload_with_sig, blobs, commitments, proofs])`) instead of the canonical execution-payload encoding (`0x03 || rlp([chain_id, nonce, ..., blob_versioned_hashes, y, r, s])`).

On a real network, the execution block body contains the bare transaction; blobs live in the consensus-layer sidecar. Per [EIP-4844 § Networking](https://eips.ethereum.org/EIPS/eip-4844#networking), the `rlp([tx_payload_body, blobs, commitments, proofs])` wrapper is only used for mempool gossip, and block body retrieval uses the standard [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718) payload (`TransactionType || TransactionPayload`).

Consequences for any block containing a blob transaction:

- `transactionsRoot` cannot be reproduced from the block's transactions by standard tooling — transaction trie inclusion proofs against Anvil headers fail.
- The block hash differs from what a canonical client would compute for the same transactions.
- `eth_getRawTransactionByHash` / `...ByBlockNumberAndIndex` for the **mined** transaction returns the pooled sidecarful bytes (~131 KB per blob) instead of the bare payload. Geth/Reth return the bare payload once the tx is mined.
- `eth_getBlockByNumber(..., true)` / `eth_getTransactionByHash` responses expose non-standard `blobs` / `commitments` / `proofs` fields on the transaction object.

The transaction *hash* itself stays canonical (alloy special-cases the 4844 tx hash to the non-sidecar encoding), which makes the mismatch easy to miss.

### Reproduction

Start `anvil --hardfork cancun`, send one blob transaction (e.g. alloy `with_blob_sidecar_4844`), fetch the mined block and the raw tx, and recompute the trie root both ways:

```text
tx hash:                      0xef20b41939a90b721b2e46272918e4fc8ca3773ccc2a1477b28f5cdc5801e82e
block hash:                   0x5d3f9b146e788cc72196ec781691c44b40000740ad1936281f581e1f32d71c95
header.transactionsRoot:      0x69c88afab56561d1852a47c2918148b3f3674017f8496ac79e7c1f44bb616bb4
raw mined tx length:          131330 bytes        <-- pooled wrapper incl. full blob
root over sidecarful leaves:  0x69c88afab56561d1852a47c2918148b3f3674017f8496ac79e7c1f44bb616bb4   <-- matches header
root over canonical leaves:   0x10c8577754225002f37e255eb24decdec75628a96c56a691f54dd3d1fcd8ce5b   <-- what a real node would commit to
```

The same recomputation against a real network passes — Sepolia block [`11033302`](https://sepolia.etherscan.io/block/11033302) (349 txs, 8 blob txs):

```text
sepolia block:                0x8d29d98de5a21530d22fff3685e96234ee46cc1766bf3624ef82b2fc4229565d
header.transactionsRoot:      0x60fc6ca15847782861f37bcf434d5e7c86e0877f498aedaedef20744f687beaf
root over canonical leaves:   0x60fc6ca15847782861f37bcf434d5e7c86e0877f498aedaedef20744f687beaf   <-- matches
```

Also reproduces on Osaka with EIP-7594 cell-proof sidecars (raw mined tx = 137,556 bytes).

### Root cause

- Mining clones the pool transaction — still the sidecarful `TxEip4844Variant::TxEip4844WithSidecar` — straight into the block body: [`crates/anvil/src/eth/backend/executor.rs#L484`](https://github.com/foundry-rs/foundry/blob/c41910099b1ea8060caf1c456b7bbdd9c4eaf805/crates/anvil/src/eth/backend/executor.rs#L484)
- `create_block` computes the root over those envelopes' `encode_2718` leaves: [`crates/anvil/core/src/eth/block.rs#L36`](https://github.com/foundry-rs/foundry/blob/c41910099b1ea8060caf1c456b7bbdd9c4eaf805/crates/anvil/core/src/eth/block.rs#L36)
- alloy's `TxEip4844WithSidecar::rlp_encode_signed` (correctly, for pooled purposes) encodes `rlp([tx, sidecar])`, so the trie leaves end up being the pooled encoding.

Keeping the sidecar after mining is needed for `anvil_getBlobByHash` / `anvil_getBlobsByBlockId` / `anvil_getBlobsByTransactionHash` (#9211), but storing it inside the block-body transactions leaks it into the consensus encoding of the block.

### Expected behavior

For a mined Anvil block:

- `header.transactionsRoot == ordered_trie_root(canonical encode_2718 leaves)`
- raw transaction bytes of a mined blob tx are the bare `0x03 || rlp([...])` payload
- blobs remain retrievable via the `anvil_getBlobs*` endpoints

I have a fix (strip sidecars at block inclusion; keep them in a dedicated store keyed by tx hash so the blob endpoints and `debug_trace*` replay keep working) and will open a PR shortly.

dghelm:
## Motivation

Closes #15132.

When Anvil mines a block containing an EIP-4844 blob transaction, the pooled (sidecarful) envelope is stored in the block body and the transaction trie root is computed over its `encode_2718` leaves. Since alloy's `TxEip4844WithSidecar` encodes the pooled networking wrapper `rlp([tx, blobs, commitments, proofs])`, the block's `transactionsRoot` — and therefore the block hash — diverges from what a canonical client (geth/reth, mainnet/Sepolia) would produce for the same transactions. Per [EIP-4844 § Networking](https://eips.ethereum.org/EIPS/eip-4844#networking), that wrapper is mempool-gossip-only; block bodies use the standard [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718) payload.

Concretely, on master (repro in the issue): a mined blob tx's raw bytes are ~131 KB and `header.transactionsRoot == root(sidecarful leaves) != root(canonical leaves)`, so trie inclusion proofs and any tooling that recomputes roots/hashes against Anvil headers fail.

## Solution

Strip the sidecar at the single point where a pool transaction is included in a block, and keep sidecars in a dedicated store so the `anvil_getBlobs*` endpoints and trace replay keep working — mirroring how real nodes separate execution data from blob data:

- **`FoundryTxEnvelope::strip_blob_sidecar` / `with_blob_sidecar`** (foundry-primitives): convert between the canonical and pooled forms. Hash and signature are unchanged (the 4844 tx hash is always computed over the non-sidecar encoding), so all hash-keyed lookups stay consistent.
- **Executor** (`execute_pool_transactions`): strips the cloned pool tx before it enters the block-body tx list and returns the sidecars in `ExecutedPoolTransactions.blob_sidecars`. This chokepoint covers all mining paths (automine, interval, `evm_mine`, reorg, pending block).
- **`BlockchainStorage.blob_sidecars`** (`tx hash → Arc<BlobTransactionSidecarVariant>`): populated in `do_mine_block`; entries are removed in `remove_block_transactions` (unwind/reorg/pruning) and on `evm_revert` snapshot rollback.
- **Blob endpoints** (`anvil_getBlobByHash`, `anvil_getBlobsByBlockId`, `anvil_getBlobsByTransactionHash`) read from the store instead of scanning block bodies.
- **`debug_trace*` replay** (`replay_tx_with_inspector`) reattaches stored sidecars when rebuilding pool transactions, since pool validation (`blob_tx.validate`) requires the sidecar.
- **State files**: the on-disk format is unchanged — sidecars are reattached to their transactions on dump and stripped back into the store on load. Old state files load fine (their historical headers and block hashes are preserved as-is), and new dumps remain readable by older versions.

Behavior changes for blocks containing blob transactions:

- `transactionsRoot` and the block hash are now canonical and reproducible from the block's transactions.
- `eth_getRawTransaction*` for **mined** blob txs returns the bare `0x03 || rlp([...])` payload (matching geth/reth); pending pool txs still return the pooled form.
- `eth_getBlockBy*(full)` / `eth_getTransactionByHash` no longer expose non-standard `blobs`/`commitments`/`proofs` fields on mined transactions.

Known limitation (documented in code): **impersonated** blob transactions are not stripped, because their synthetic hash is derived from the encoded transaction, and changing either the hash derivation or the encoding would break loading state files from earlier versions (mined transactions are keyed by the stored synthetic hash). Their sidecar is mirrored into the sidecar store so blob endpoints and replay read from one source. Canonical roots are guaranteed for properly-signed transactions; impersonated blocks are non-canonical regardless due to the bypass signature.

### Tests

- `mined_blob_block_has_canonical_transactions_root` (+ EIP-7594/Osaka variant): raw txs decode as bare envelopes, recomputed `calculate_transaction_root` matches the header, recomputed header hash matches the block hash, blobs stay retrievable, full-block RPC JSON has no sidecar fields. Both fail on master and pass with this change.
- `can_trace_transaction_after_blob_tx`: `debug_traceTransaction` of a tx mined after a blob tx (exercises replay rehydration) and of the blob tx itself.
- `can_load_state_with_blob_txs`: dump → load round trip; block hash unchanged, blobs retrievable from the migrated store, raw tx canonical, tracing works on the loaded chain.
- `reverting_snapshot_removes_blob_sidecars`: `evm_revert` drops sidecars of reverted txs.
- Unit tests: strip/reattach roundtrip for both sidecar variants (foundry-primitives), dump/load/remove cycle for the sidecar store (anvil storage).
- All pre-existing blob tests pass unchanged (`can_get_blobs_by_versioned_hash`, `can_get_blobs_by_tx_hash`, fork/7594/multi-blob/bypass-sidecar, beacon API, state, traces).

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

## AI Disclosure

This PR was developed with AI assistance (Claude Code) for the root-cause analysis, implementation, and tests; I directed and reviewed the work.

### DIFF
--- a/crates/anvil/core/src/eth/transaction/mod.rs
+++ b/crates/anvil/core/src/eth/transaction/mod.rs
@@ -57,13 +57,20 @@ impl MaybeImpersonatedTransaction<FoundryTxEnvelope> {
     /// Removes the blob sidecar from the wrapped transaction so it can be included in a block
     /// body in its canonical (non-pooled) form, returning the sidecar, if any.
     ///
-    /// For impersonated transactions, the synthetic hash is derived from the transaction hash
-    /// rather than the pooled sidecarful encoding (see [`Self::hash`]), so stripping the sidecar
-    /// does not change the hash users received at submission.
+    /// Impersonated transactions are left untouched: their synthetic hash is derived from the
+    /// encoded transaction (see [`Self::hash`]), so stripping the sidecar would change the hash
+    /// users received at submission, and changing the hash derivation would break loading state
+    /// files written by earlier versions (which key mined transactions by the stored synthetic
+    /// hash). Their sidecar is still returned so callers can index it.
     pub fn strip_blob_sidecar(self) -> (Self, Option<BlobTransactionSidecarVariant>) {
         let Self { transaction, impersonated_sender } = self;
-        let (transaction, sidecar) = transaction.strip_blob_sidecar();
-        (Self { transaction, impersonated_sender }, sidecar)
+        if impersonated_sender.is_some() {
+            let sidecar = transaction.sidecar().map(|tx| tx.sidecar.clone());
+            (Self { transaction, impersonated_sender }, sidecar)
+        } else {
+            let (transaction, sidecar) = transaction.strip_blob_sidecar();
+            (Self { transaction, impersonated_sender }, sidecar)
+        }
     }
 
     /// Reattaches a blob sidecar to the wrapped transaction, returning the pooled (sidecarful)
@@ -75,7 +82,7 @@ impl MaybeImpersonatedTransaction<FoundryTxEnvelope> {
     }
 }
 
-impl<T: SignerRecoverable + TxHashRef> MaybeImpersonatedTransaction<T> {
+impl<T: SignerRecoverable + TxHashRef + Encodable> MaybeImpersonatedTransaction<T> {
     /// Recovers the Ethereum address which was used to sign the transaction.
     pub fn recover(&self) -> Result<Address, RecoveryError> {
         if let Some(sender) = self.impersonated_sender {
@@ -87,11 +94,11 @@ impl<T: SignerRecoverable + TxHashRef> MaybeImpersonatedTransaction<T> {
     /// Returns the hash of the transaction.
     ///
     /// If the transaction is impersonated, returns a unique hash derived by appending the
-    /// impersonated sender address to the transaction hash before hashing.
+    /// impersonated sender address to the encoded transaction before hashing.
     pub fn hash(&self) -> B256 {
         if let Some(sender) = self.impersonated_sender {
-            let mut buffer = Vec::with_capacity(52);
-            buffer.extend_from_slice(self.transaction.tx_hash().as_ref());
+            let mut buffer = Vec::new();
+            self.transaction.encode(&mut buffer);
             buffer.extend_from_slice(sender.as_ref());
             return B256::from_slice(alloy_primitives::utils::keccak256(&buffer).as_slice());
         }
@@ -168,7 +175,7 @@ impl<T> PendingTransaction<T> {
     }
 }
 
-impl<T: SignerRecoverable + TxHashRef> PendingTransaction<T> {
+impl<T: SignerRecoverable + TxHashRef + Encodable> PendingTransaction<T> {
     pub fn new(transaction: T) -> Result<Self, RecoveryError> {
         let transaction = MaybeImpersonatedTransaction::new(transaction);
         let sender = transaction.recover()?;
@@ -248,14 +255,30 @@ mod tests {
         (FoundryTxEnvelope::Eip4844(variant.into_signed(signature)), sidecar)
     }
 
+    // Impersonated transactions are not stripped (their synthetic hash is derived from the
+    // encoded transaction), but the sidecar is still returned so callers can index it.
     #[test]
-    fn impersonated_blob_strip_preserves_hash() {
+    fn impersonated_blob_strip_keeps_sidecar_and_hash() {
         let (transaction, sidecar) = sidecarful_4844_envelope();
         let transaction =
             MaybeImpersonatedTransaction::impersonated(transaction, Address::with_last_byte(1));
         let hash = transaction.hash();
 
         assert!(transaction.as_ref().sidecar().is_some());
+
+        let (kept, returned_sidecar) = transaction.strip_blob_sidecar();
+
+        assert_eq!(returned_sidecar, Some(sidecar));
+        assert!(kept.as_ref().sidecar().is_some());
+        assert_eq!(kept.hash(), hash);
+    }
+
+    // Properly-signed transactions are stripped to their canonical form, hash unchanged.
+    #[test]
+    fn signed_blob_strip_removes_sidecar_and_preserves_hash() {
+        let (transaction, sidecar) = sidecarful_4844_envelope();
+        let transaction = MaybeImpersonatedTransaction::new(transaction);
+        let hash = transaction.hash();
 
         let (stripped, stripped_sidecar) = transaction.strip_blob_sidecar();
 
--- a/crates/anvil/core/src/eth/transaction/mod.rs
+++ b/crates/anvil/core/src/eth/transaction/mod.rs
@@ -1,6 +1,6 @@
 //! Transaction related types
 use alloy_consensus::{
-    Transaction, Typed2718,
+    BlobTransactionSidecarVariant, Transaction, Typed2718,
     crypto::RecoveryError,
     transaction::{SignerRecoverable, TxHashRef},
 };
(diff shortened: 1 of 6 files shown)


figtracer (dismissed):

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -3286,12 +3282,23 @@ where
             .position(|tx| tx.hash() == hash)
             .expect("transaction not found in block");
 
+        // Block bodies store blob transactions in their canonical form; reattach the stored
+        // sidecars for replay, since pool transaction validation expects pooled (sidecarful)
+        // blob transactions.
+        let rehydrate = |tx: &MaybeImpersonatedTransaction<FoundryTxEnvelope>| {
+            let tx = tx.clone();
+            match self.blockchain.storage.read().blob_sidecars.get(&tx.hash()) {
```
mablr: read-lock taken inside a loop, could it be made in a more elegant way ?

crates/anvil/core/src/eth/transaction/mod.rs
```diff
@@ -53,6 +53,33 @@ impl<T> MaybeImpersonatedTransaction<T> {
...
+    /// body in its canonical (non-pooled) form, returning the sidecar, if any.
+    ///
+    /// Impersonated transactions are left untouched: their synthetic hash is derived from the
+    /// encoded transaction (see [`Self::hash`]), so stripping the sidecar would change the hash
+    /// users received at submission. Their sidecar is still returned so callers can index it.
+    pub fn strip_blob_sidecar(self) -> (Self, Option<BlobTransactionSidecarVariant>) {
+        let Self { transaction, impersonated_sender } = self;
+        if impersonated_sender.is_some() {
+            let sidecar = transaction.sidecar().map(|tx| tx.sidecar.clone());
+            (Self { transaction, impersonated_sender }, sidecar)
```
mablr, line 69: It seems there is double storage, the impersonated branch returns the sidecar but doesn't strip it from the transaction, so `blob_sidecars` and the block body both hold the  sidecar for impersonated blob txs. Safe (rehydrate is idempotent), but wasteful?

mablr:
^ not blockers but worth taking a look :)

crates/anvil/src/eth/backend/mem/storage.rs
```diff
@@ -409,20 +421,57 @@ impl<N: Network> BlockchainStorage<N> {
...
-            let block: Block = serializable_block.clone().into();
+            let mut block: Block = serializable_block.clone().into();
+            block.body.transactions = block
+                .body
+                .transactions
+                .into_iter()
+                .map(|tx| {
+                    let (tx, sidecar) = tx.strip_blob_sidecar();
+                    if let Some(sidecar) = sidecar {
+                        self.blob_sidecars.insert(tx.hash(), Arc::new(sidecar));
```
mablr, line 474: This breaks old state dumps with impersonated blob txs. Those dumps keep `SerializableTransaction.info.transaction_hash` as the old synthetic hash (`keccak(encoded sidecarful tx || sender)`), but `load_blocks` strips the sidecar and stores it under the new synthetic hash (`keccak(tx_hash || sender)`). After `load_transactions`, the mined tx record is still keyed by the old hash, so `anvil_getBlobByTransactionHash(old_hash)` misses.

Fix for my previous comment is a footgun :/

crates/anvil/src/eth/backend/mem/mod.rs
dghelm: Done in b7cb9a21c — the block's relevant sidecars are now snapshotted into a local map under the same read lock that fetches the block, so the replay loop does not touch the storage lock at all. Also slightly better consistency, since the block and its sidecars come from one lock acquisition.

crates/anvil/src/eth/backend/mem/storage.rs
dghelm: You're right, and thanks for catching it before this went further — the hash-derivation change can't be made compatible: old dumps baked the sidecarful encoding into their stored hashes, so any new derivation drifts from `info.transaction_hash`, and it's worse than blob lookups — `mined_transactions_in_block` would miss on the recomputed hash and `eth_getBlockBy*(full)` would return null for the whole block. Reverted in 4ae1a16bb: impersonated txs keep the original hash derivation and keep their sidecar in the body (mirrored into the store). The double storage for that corner case is deliberate and now documented in `strip_blob_sidecar` with the compat rationale, plus unit tests pinning both behaviors (impersonated: sidecar kept, hash stable; signed: stripped, hash stable).

crates/anvil/core/src/eth/transaction/mod.rs
dghelm: After a detour (see the thread on `load_blocks`) this is now deliberate and documented: impersonated blob txs keep the sidecar in the body because their synthetic hash is derived from the encoded transaction — stripping would change the hash users got at submission, and changing the derivation breaks old state dumps. The store entry is still mirrored so the blob endpoints and replay read from one source. Worst case is one duplicated sidecar per impersonated blob tx, which seemed like the right trade against a hash compatibility break.

mablr (approved):
sgtm

mattsse (changes requested):
these are a lot of changes, unclear what is fix for the txroot issue and what is refactor here.

can we simplify this and focus on the fix?

dghelm:
The current diff makes the scope harder to review than it should.

The actual fix is: mined blocks need to commit to the canonical EIP-4844 transaction encoding, not the pooled sidecar encoding. In Anvil that means normal blob txs have to enter the block body without the sidecar before `create_block` computes `transactions_root`.

> can we simplify this and focus on the fix?

The extra plumbing is there because Anvil currently serves blob data from mined transaction bodies. Once the sidecar leaves the body, we need a sidecar store so `anvil_getBlobs*` still works, and trace replay needs to reattach sidecars because pool validation expects pooled blob txs. That part is not intended as refactor.

dghelm:
@mattsse Any suggestions here?

PR #15136, opened 2026-06-11
feat(anvil): add `eth_resend`

figtracer:
Adds `eth_resend` support to Anvil so pending unlocked-account transactions can be replaced with an updated gas price or gas limit. The method checks that the requested sender and nonce are already pending before signing and submitting the replacement.

ref: https://github.com/paradigmxyz/reth/pull/24339

### DIFF
--- a/crates/anvil/tests/it/transaction.rs
+++ b/crates/anvil/tests/it/transaction.rs
@@ -3,8 +3,12 @@ use crate::{
     utils::{connect_pubsub, http_provider_with_signer},
 };
 use alloy_consensus::Transaction;
-use alloy_network::{EthereumWallet, ReceiptResponse, TransactionBuilder, TransactionResponse};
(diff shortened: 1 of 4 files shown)


crates/anvil/tests/it/transaction.rs
```diff
@@ -195,6 +199,57 @@ async fn can_replace_transaction() {
     );
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn can_resend_transaction() {
```
mablr, line 203: This covers the happy path, but `eth_resend` has new RPC-specific validation that is not exercised: missing nonce, sender/nonce not in the pool, and underpriced replacement through `eth_resend`. Could we add at least one negative test here so regressions in the method-specific checks do not rely only on the generic `eth_sendTransaction` replacement tests?

mablr (commented):
sgtm, just a suggestion on test.

mablr (approved):
sgtm

stevencartavia (approved):

PR #15137, opened 2026-06-11
fix(doc): match inheritdoc overloads with canonical types

mablr:
## Motivation

https://github.com/foundry-rs/foundry/pull/14568#discussion_r3393747789 follow-up.

Match inheritdoc overloads with canonical types

### DIFF
--- a/crates/doc/src/hir_ext.rs
+++ b/crates/doc/src/hir_ext.rs
@@ -263,8 +263,8 @@ pub struct InheritedDoc {
 /// matching function and returns its natspec if found.
 ///
 /// When `param_types` is `Some`, the resolver prefers a function whose parameter
-/// type signature matches exactly; this disambiguates overloads. If no exact
-/// signature match is found, it falls back to the first name match.
+/// type signature matches exactly; this disambiguates overloads. Ambiguous
+/// overloads are never matched by name alone.
 pub fn resolve_inheritdoc(
     gcx: Gcx<'_>,
     contract_id: ContractId,
@@ -315,10 +315,11 @@ pub fn resolve_inheritdoc(
         if let Some(want) = param_types
             && name_matches.len() > 1
         {
-            // Try to find a signature-exact match with docs; fall through to name matches below.
+            // Try to find a signature-exact match with docs; fall through only for
+            // non-overloaded name matches below.
             for &fid in &name_matches {
-                let got = function_param_types(gcx, fid);
-                if got.len() == want.len()
+                if let Some(got) = function_param_types(gcx, fid)
+                    && got.len() == want.len()
                     && got.iter().zip(want).all(|(a, b)| a == b)
                     && let Some(doc) = extract_inherited_doc(gcx, fid)
                     && (!doc.notices.is_empty()
@@ -331,7 +332,7 @@ pub fn resolve_inheritdoc(
             }
         }
 
-        if name_matches.len() > 1 && param_types.is_some() {
+        if name_matches.len() > 1 {
             continue;
         }
 
@@ -394,7 +395,7 @@ pub fn resolve_inheritdoc_var(
                 ItemId::Function(fid) => {
                     let f = gcx.hir.function(fid);
                     if f.name.map(|n| n.as_str() == var_name).unwrap_or(false)
-                        && function_param_types(gcx, fid).is_empty()
+                        && function_param_types(gcx, fid).map(|p| p.is_empty()).unwrap_or(false)
                         && let Some(doc) = extract_inherited_doc(gcx, fid)
                         && (!doc.notices.is_empty()
                             || !doc.devs.is_empty()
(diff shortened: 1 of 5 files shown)


crates/doc/src/hir_ext.rs
```diff
@@ -434,44 +433,18 @@ fn extract_inherited_doc_var(gcx: Gcx<'_>, vid: VariableId) -> Option<InheritedD
...
+    f.parameters
         .iter()
-        .map(|v| {
-            let raw = sm.span_to_snippet(v.ty.span).unwrap_or_default();
-            let t = raw.split_whitespace().collect::<Vec<_>>().join(" ");
-            normalize_sol_type(&t)
+        .map(|&param| {
+            let ty = gcx.type_of_item(param.into());
+            let mut out = String::new();
+            TyAbiPrinter::new(gcx, &mut out, TyAbiPrinterMode::Signature)
```
figtracer: this now panics for valid inherited internal functions with non-abi params, for example `mapping(...) storage`, because the abi printer panics on mapping types. can we keep a source/string fallback for non-abi-printable types?

figtracer (commented):
left one inline review comment.

crates/doc/src/render.rs
```diff
@@ -327,9 +327,18 @@ fn render_contract<'ast, 'gcx>(
...
+                let param_types = hir_id
+                    .and_then(|cid| {
+                        gcx.hir.contract(cid).items.iter().find_map(|&item| match item {
+                            hir::ItemId::Function(id) if gcx.hir.function(id).span == *span => {
+                                Some(id)
+                            }
+                            _ => None,
+                        })
+                    })
+                    .map(|fid| hir_ext::function_param_types(gcx, fid));
```
grandizzy: This span-based `FunctionId` lookup yields `None` on a miss, and `resolve_inheritdoc`'s overload guard is `name_matches.len() > 1 && param_types.is_some()`. So if param types are unavailable for a base with overloads, resolution falls through to "first candidate with docs" and can inherit the wrong overload's NatSpec — a regression vs the old AST path, which always supplied types. Suggest dropping `param_types.is_some()` from that guard (never name-only-match an overloaded base) and adding a `debug_assert!`/`warn!` here on lookup miss.

crates/forge/tests/cli/doc.rs
```diff
@@ -525,6 +525,62 @@ contract Alias is IAlias {
...
+"#,
+    );
+
+    cmd.args(["doc"]).assert_success();
+
+    let content =
+        std::fs::read_to_string(prj.root().join("docs/src/pages/src/contract.C.mdx")).unwrap();
+
+    assert!(
+        content.contains("Sets the account status"),
```
grandizzy: #14568 explicitly requested overload regressions for `uint` vs `uint256` and `uint[]` vs `uint256[]`; only the enum-alias case is covered here. Please add those two (e.g. base `configure(uint)`/`configure(address)`, child `configure(uint256)`; and `batch(uint[] calldata)`/`batch(address[] calldata)`, child `batch(uint256[] calldata)`). Also, `content.contains(...)` over the whole page means a swapped-overload bug could pass as long as both notices appear anywhere — consider asserting the notice lands under the correct function section/anchor.

grandizzy (commented):
Nice cleanup — moving overload matching off `span_to_snippet` onto canonical HIR/ABI types via `TyAbiPrinter` is the right fix and resolves the alias/qualification cases from #14568. Two findings beyond @figtracer's `mapping(...)` panic note:

- **Overloaded base can silently inherit the wrong overload's docs.** `function_param_types` is now derived from a span-based HIR `FunctionId` lookup in `render.rs`, which yields `None` on a miss (the old AST path always produced types). With `param_types == None` and overloads present, `resolve_inheritdoc` no longer `continue`s past the `name_matches.len() > 1 && param_types.is_some()` guard and falls through to "first candidate with docs." Suggest never name-only-matching an overloaded base (drop `param_types.is_some()` from that guard) and/or `debug_assert!` when the lookup misses.
- **Missing the regressions this PR set out to fix.** #14568 explicitly asked for `uint` vs `uint256` and `uint[]` vs `uint256[]` overload tests; only the enum-alias case is covered. Also the `content.contains(...)` assertions are page-wide, so a swapped-overload bug could still pass.

Nit: the `normalize_sol_type` doc comment still says "compare equal during overload matching," but it now only feeds slug/anchor generation — worth updating (anchors and matching can now diverge).

crates/doc/src/hir_ext.rs
mablr: https://github.com/foundry-rs/foundry/pull/15137/commits/40fb5291e687a5a163f853549e64841c83aae10b

crates/doc/src/render.rs
mablr: https://github.com/foundry-rs/foundry/pull/15137/commits/40fb5291e687a5a163f853549e64841c83aae10b

crates/forge/tests/cli/doc.rs
mablr: https://github.com/foundry-rs/foundry/pull/15137/commits/40fb5291e687a5a163f853549e64841c83aae10b

mablr: fixed, and switched to snapbox testing for more robust output matching, lmk if you like it.

grandizzy (approved):
lgtm!

figtracer (approved):
nice

PR #15138, opened 2026-06-11
ci: add `derek bench` trigger

kuyziss:
## Summary

Adds a PR comment trigger that publishes a signed benchmark event to internal
infra over mTLS. Command is `derek bench <subcommand>` (also `decofe bench` /
`@decofe bench`); only `invariant` is live today, `build`/`all` are reserved.
The benchmark runs server-side and reports results separately.

## Changes

- `derek bench invariant [args]` trigger (subcommand namespace; build/all reserved)
- Move authorization into the job-level if; drop unused permission
- Add per-PR concurrency
- Pass secrets via quoted env vars (not inline) and scope them to a shared
  `bench` environment
- Send a minimal payload with only the user-tunable knobs
  (compare-ref, timeout, workers, benchmark-type); everything else comes from
  server defaults
- Reject forks; re-validated server-side

## Usage

```
derek bench invariant                                   # all defaults
derek bench invariant timeout=300 workers=2             # short test run
derek bench invariant compare-ref=v1.0.0 timeout=7200 workers=16 benchmark-type=optimization
```

Knobs: `compare-ref` (default `master`), `timeout` (60-14400s, default `3600`),
`workers` (1-256, default = server), `benchmark-type` (`property`|`optimization`,
default `property`).


.github/workflows/scfuzzbench.yml
```diff
@@ -0,0 +1,270 @@
...
+          set -euo pipefail
+
+          printf '%s' '${{ secrets.EVENTS_KEY }}' > "${RUNNER_TEMP}/key"
+          printf '%s' '${{ secrets.EVENTS_CERT }}' > "${RUNNER_TEMP}/cert"
+          printf '%s' "$PAYLOAD_B64" \
+            | base64 --decode \
+            > "${RUNNER_TEMP}/scfuzzbench-event.json"
+
+          # EVENTS_ARGS may contain additional curl arguments, not just a URL.
+          curl -sf -o /dev/null -X POST ${{ secrets.EVENTS_ARGS }} \
```
mablr: secrets should be quoted? not sure.

mablr (commented):
Sgtm, just a small question.

.github/workflows/bench.yml
```diff
@@ -0,0 +1,291 @@
...
+      github.event.issue.pull_request &&
+      contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association) &&
+      (
+        startsWith(github.event.comment.body, 'derek bench') ||
+        startsWith(github.event.comment.body, 'decofe bench') ||
+        startsWith(github.event.comment.body, '@decofe bench')
+      )
+    runs-on: ubuntu-latest
+    # Scopes the EVENTS_* mTLS secrets; shared across all `bench` subcommands.
+    environment: bench
```
mablr, line 31: The bench environment does not currently exist in foundry-rs/foundry (`gh api repos/foundry-rs/foundry/environments` only lists `github-pages` and `release`). Since this workflow expects `EVENTS_KEY`, `EVENTS_CERT`, and `EVENTS_URL` from that environment, the first real publish will likely run with empty secret values and fail.

.github/workflows/bench.yml
```diff
@@ -0,0 +1,291 @@
...
+  publish:
+    # Gate on association here so unauthorized commenters never start a run.
+    if: >-
+      github.event.issue.pull_request &&
+      contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association) &&
+      (
+        startsWith(github.event.comment.body, 'derek bench') ||
+        startsWith(github.event.comment.body, 'decofe bench') ||
+        startsWith(github.event.comment.body, '@decofe bench')
+      )
```
mablr, line 28: The job-level trigger does not match the parser below: it is case-sensitive and uses `startsWith('derek bench')`, while the parser is case-insensitive and requires bench. That means `Derek bench invariant` is silently skipped, but `derek benchmark` starts the workflow and reports "Unknown bench subcommand derek".

.github/workflows/bench.yml
grandizzy: pls recheck

.github/workflows/bench.yml
grandizzy: ptal https://github.com/foundry-rs/foundry/pull/15138/changes/944ecbf2b0bbe58192b029d5ef3af704067a8c25

mablr (approved):
sgtm

grandizzy (approved):

PR #15139, opened 2026-06-11
feat(symbolic): add json result schema

figtracer:
## Summary

- Adds a v1 JSON Schema artifact for symbolic test result objects.
- Emits a stable `symbolic` object in `forge test --symbolic --json` results with normalized status, incomplete reason, bounds, solver stats, assumptions, trace metadata, replay metadata, and counterexample payloads.
- Classifies replay mismatches/errors/skips as machine-readable `incomplete` results instead of confirmed counterexamples.
- Documents the schema location in the symbolic executor README.

## Why

This is the T0.1 trust-plumbing item from the symbolic testing roadmap: agents and tooling need structured symbolic outcomes rather than parsing display strings or overloaded fuzz counterexample fields.

## Validation

- `cargo fmt`
- `cargo +1.95.0 check -p forge`
- `cargo +1.95.0 test -p forge --test cli symbolic_json_schema -- --nocapture`
- `jq empty crates/evm/symbolic/assets/symbolic-result.schema.json`
- Python `jsonschema.Draft202012Validator` semantic check for valid pass/fail/incomplete samples and contradictory status/replay/call_trace states

### DIFF
--- a/crates/forge/src/result.rs
+++ b/crates/forge/src/result.rs
@@ -7,6 +7,7 @@ use alloy_primitives::{
 };
 use eyre::Report;
 use foundry_common::{ContractsByArtifact, get_contract_name, get_file_name, shell};
+use foundry_config::{SymbolicConfig, SymbolicExplorationOrder, SymbolicStorageLayout};
 use foundry_evm::{
     core::{Breakpoints, evm::FoundryEvmNetwork},
     coverage::HitMaps,
@@ -15,7 +16,7 @@ use foundry_evm::{
     fuzz::{CounterExample, FuzzCase, FuzzFixtures, FuzzTestResult},
     traces::{CallTraceArena, CallTraceDecoder, TraceKind, Traces},
 };
-use foundry_evm_symbolic::{PortfolioDiagnostics, SymbolicStats};
+use foundry_evm_symbolic::{PortfolioDiagnostics, SymbolicStats, SymbolicStopReason};
 use serde::{Deserialize, Serialize};
 use std::{
     borrow::Cow,
@@ -30,6 +31,11 @@ pub(crate) fn invariant_campaign_display_name(contract_name: &str) -> String {
 }
 
 const INVARIANT_CAMPAIGN_FALLBACK_NAME: &str = "Invariant campaign";
+const SYMBOLIC_RESULT_SCHEMA_VERSION: u32 = 1;
+
+const fn symbolic_result_schema_version() -> u32 {
+    SYMBOLIC_RESULT_SCHEMA_VERSION
+}
 
 /// The aggregated result of a test run.
 #[derive(Clone, Debug)]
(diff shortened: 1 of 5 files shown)


crates/forge/src/runner.rs
```diff
@@ -1031,13 +1049,42 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
                     Ok(res) => (res.raw, None),
                     Err(EvmError::Execution(err)) => (err.raw, Some(err.reason)),
                     Err(EvmError::Skip(reason)) => {
-                        self.result.single_skip(reason);
+                        let replay_reason = format!("vm.skip during concrete replay: {reason}");
```
grandizzy, line 1054: Critical: this changes the top-level test status from `Skipped` to `Failure`. Previously this branch called `single_skip(reason)` → `TestStatus::Skipped`; now it routes through `symbolic_result`, which maps every non-`pass` status to `TestStatus::Failure`. A `vm.skip` guard hit during concrete replay will now produce a non-zero exit code / red CI.

The new `symbolic.status="incomplete"` + `replay.status="skipped"` payload is correct, but the legacy status should stay `Skipped`. Suggest giving `symbolic_result` an explicit `TestStatus` argument and passing `Skipped` here. Please also add a regression test for "symbolic finds candidate → replay hits vm.skip" asserting top-level `Skipped` plus `symbolic.replay.status == "skipped"`.

crates/forge/src/result.rs
```diff
@@ -1269,9 +1672,14 @@ impl TestResult {
             heuristic_witnesses: stats.heuristic_witnesses,
             solver_time_ms: stats.solver_time_ms,
         };
-        self.status = if success { TestStatus::Success } else { TestStatus::Failure };
+        self.status = if symbolic.status == SymbolicResultStatus::Pass {
```
grandizzy: Critical (same issue): `self.status = if symbolic.status == Pass { Success } else { Failure }` collapses skip into failure. Take the desired `TestStatus` as a parameter (Success / Failure / Skipped) instead of deriving it solely from `SymbolicResultStatus`.

crates/forge/src/runner.rs
```diff
@@ -1049,22 +1096,44 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
-                    } else {
-                        reason
-                    },
-                    Some(counterexample),
-                    stats,
                 );
+                let symbolic_counterexample = SymbolicCounterexample::from(&base_counterexample);
+                self.result.extend(raw_call_result);
+                if success {
+                    let reason = "symbolic counterexample did not replay".to_string();
```
grandizzy, line 1109: Minor: this passes `None` for the legacy top-level `counterexample`, so the candidate is dropped for non-JSON / legacy consumers (it still lives in `symbolic.counterexample`). Likely intentional since a mismatch isn't a confirmed counterexample — if so, please note it in the README/schema docs; otherwise pass the candidate through to the legacy field too.

crates/forge/src/result.rs
```diff
@@ -601,6 +607,399 @@ pub struct InvariantPredicateResult {
...
+            stats,
+            None,
+            SymbolicReplayMetadata::not_required(),
+            SymbolicCallTrace::none(),
+            None,
+        )
+    }
+
+    /// Creates a symbolic counterexample result that concrete replay confirmed.
+    pub fn fail_counterexample(
```
grandizzy, line 649: Minor: the constructor accepts an arbitrary `SymbolicReplayMetadata`, which can produce schema-invalid output (the schema requires `replay = {required:true, status:"confirmed", reason:null}` for `fail_counterexample`). All call sites currently pass `confirmed()`, but consider setting `SymbolicReplayMetadata::confirmed()` internally to prevent future drift.

crates/forge/src/runner.rs
```diff
@@ -1009,17 +1012,32 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                        stats,
+                        SymbolicReplayMetadata::not_required(),
+                        SymbolicCallTrace::none(),
+                        None,
+                    ),
                 );
             }
             SymbolicRunResult::Counterexample { args, calldata, stats } => {
+                let symbolic_counterexample = SymbolicCounterexample::from(
+                    &BaseCounterExample::from_fuzz_call(calldata.clone(), args.clone(), None),
```
grandizzy, line 1041: Minor: the symbolic counterexample is built here (pre-replay, no traces) and rebuilt after replay with traces. Harmless but drift-prone — consider building the shared `SymbolicCounterexample` once and only constructing the trace-bearing `BaseCounterExample` for the legacy `CounterExample`.

grandizzy (changes requested):
This PR adds a versioned JSON Schema and a stable `symbolic` object to `forge test --symbolic --json`, and correctly reclassifies replay mismatch/error/skip as `incomplete` rather than confirmed counterexamples. Schema↔Rust conformance checks out across all branches (pass / fail_counterexample / incomplete / mismatch / error / skip), including `call_trace`, `replay`, null-ness rules, and `value` (alloy `U256` → `"0x…"` string).

One blocking issue: a `vm.skip` hit during concrete replay now fails the test where it was previously skipped — see inline. The rest are non-blocking.

crates/forge/src/runner.rs
mablr: I think this is worth preserving in the legacy field too. Before this change, the mismatch branch still populated `TestResult.counterexample`, so the normal `[FAIL ...; counterexample: ...]` output showed the solver input that failed to replay. With `None`, non-JSON users only see `symbolic counterexample did not replay` and lose the actual inputs unless they inspect the new symbolic JSON object.

figtracer: took care of these in latest push

grandizzy (approved):
Reevaluated after the force-push (`85f37eb`). All five findings from my previous review are resolved:

- skip-during-replay now preserves top-level `TestStatus::Skipped` (was failing) — `symbolic_result` takes an explicit `TestStatus`.
- mismatch branch repopulates the legacy `counterexample` field.
- `fail_counterexample` sets `replay = confirmed` internally, so it cannot emit schema-invalid output.
- single shared `SymbolicCounterexample` construction.
- bonus: a dedicated `symbolic_json_schema_reports_replay_skip` regression test covering the skip path (asserts top-level `Skipped` + `symbolic.replay.status == "skipped"`).

Schema↔Rust conformance holds across all branches and move semantics are sound. LGTM.

mablr (approved):
sgtm

Issue #410, opened 2022-01-10
forge: show current variable values during debug


crisgarner:
Add some way to show the current value of the variables while debugging

gakonst:
It would be nice if you could be more specific, do you mean like Watch expressions?

onbjerg:
I think the ask is to show the values of variables in the current scope, similar to how Truffle's debugger can output variables in scope (their `v` command).

crisgarner:
> I think the ask is to show the values of variables in the current scope, similar to how Truffle's debugger can output variables in scope (their `v` command).

Exactly, just like truffle debugger. Here is a screenshot from the visual debugger: 

![image](https://user-images.githubusercontent.com/578688/149048736-c1c855fe-c20b-4ab4-886f-898a71be6a87.png)

https://github.com/trufflesuite/truffle-plugin-debugger

mattsse:
The debugger now shows the values of variables in the current scope, added in #15141 with follow-up fixes in #15200. Further debugger improvements are tracked in #927.

PR #15141, opened 2026-06-11
feat(debugger): show scoped variables

figtracer:
## Summary
- add Solidity function-scope metadata for debugger source lookup, including parameters, returns, and visible local declarations
- preserve decoded call data on debugger nodes and run the existing internal-call identifier before flattening traces
- add a Variables pane that decodes external/internal function parameters and completed internal return values when reliable data is available
- list visible locals by name without guessing runtime values when source maps do not expose stable stack/memory slots

Refs #410

## Tests
- cargo +1.95.0 test -p foundry-debugger -p foundry-evm-traces --lib
- cargo +1.95.0 test -p forge --test cli test_cmd::test_debug_with_dump -- --exact

### DIFF
--- a/crates/debugger/src/builder.rs
+++ b/crates/debugger/src/builder.rs
@@ -1,17 +1,20 @@
 //! Debugger builder.
 
-use crate::{DebugNode, Debugger, debugger::DebuggerStats, node::flatten_call_trace};
+use crate::{Debugger, debugger::DebuggerStats, node::flatten_call_trace};
 use alloy_primitives::{Address, map::AddressHashMap};
 use foundry_common::get_contract_name;
 use foundry_evm_core::Breakpoints;
-use foundry_evm_traces::{CallTraceArena, CallTraceDecoder, Traces, debug::ContractSources};
+use foundry_evm_traces::{
+    CallTraceArena, CallTraceDecoder, Traces,
+    debug::{ContractSources, DebugTraceIdentifier},
+};
 
 /// Debugger builder.
 #[derive(Debug, Default)]
 #[must_use = "builders do nothing unless you call `build` on them"]
 pub struct DebuggerBuilder {
     /// Debug traces returned from the EVM execution.
-    debug_arena: Vec<DebugNode>,
+    trace_arenas: Vec<CallTraceArena>,
     /// Aggregate stats for the traces passed to the debugger.
     stats: DebuggerStats,
     /// Identified contracts.
@@ -47,7 +50,7 @@ impl DebuggerBuilder {
         }
         self.stats.session_subcalls =
             self.stats.session_subcalls.saturating_add(arena.nodes().len().saturating_sub(1));
-        flatten_call_trace(arena, &mut self.debug_arena);
+        self.trace_arenas.push(arena);
         self
     }
 
@@ -94,8 +97,33 @@ impl DebuggerBuilder {
     /// Builds the debugger.
     #[inline]
     pub fn build(self) -> Debugger {
-        let Self { debug_arena, stats, identified_contracts, sources, breakpoints } = self;
+        let Self { mut trace_arenas, stats, identified_contracts, sources, breakpoints } = self;
+        identify_internal_calls(&mut trace_arenas, &identified_contracts, &sources);
+        let mut debug_arena = Vec::new();
+        for arena in trace_arenas {
+            flatten_call_trace(arena, &mut debug_arena);
+        }
         Debugger::new_with_stats(debug_arena, stats, identified_contracts, sources, breakpoints)
+    }
+}
+
+fn identify_internal_calls(
+    trace_arenas: &mut [CallTraceArena],
+    identified_contracts: &AddressHashMap<String>,
+    sources: &ContractSources,
+) {
+    if sources.artifacts_by_name.is_empty() {
+        return;
+    }
+
+    let identifier = DebugTraceIdentifier::new(sources.clone());
+    for arena in trace_arenas {
+        for node in arena.nodes_mut() {
+            let Some(contract_name) = identified_contracts.get(&node.trace.address) else {
+                continue;
+            };
+            identifier.identify_node_steps(node, contract_name);
+        }
     }
 }
 
@@ -160,7 +188,7 @@ mod tests {
 
         assert_eq!(builder.stats.session_subcalls, 1);
         assert_eq!(builder.stats.session_trace_gas_used, 100);
-        assert_eq!(builder.debug_arena.len(), 1);
+        assert_eq!(builder.trace_arenas.len(), 1);
     }
 
     #[test]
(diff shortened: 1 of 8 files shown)


crates/evm/traces/src/debug/sources.rs
```diff
@@ -49,12 +83,18 @@ impl SourceData {
...
                 let r = output.parser().solc().compiler().enter(|compiler| -> Option<()> {
                     let (_, source) = compiler.gcx().get_ast_source(root.join(&path))?;
+                    let source_map = compiler.sess().source_map();
                     for item in source.ast.as_ref()?.items.iter() {
                         if let solar::ast::ItemKind::Contract(contract) = &item.kind {
-                            contract_definitions.push((
-                                contract.name.to_string(),
-                                compiler.sess().source_map().span_to_range(item.span).unwrap(),
-                            ));
+                            let contract_range = source_map.span_to_range(item.span).ok()?;
```
mablr: Cascade early-return on span failure, the `?` may abort the entire file's contract processing, all contracts after the first bad span silently lose variable tracking.

crates/debugger/src/tui/draw.rs
```diff
@@ -583,6 +619,330 @@ impl TUIContext<'_> {
...
+        variables
+    }
+
+    fn decode_parameter_values(&self, scope: &DebugSourceScope) -> Option<Vec<String>> {
+        self.decode_internal_parameter_values(scope)
+            .or_else(|| decode_external_parameter_values(scope, &self.debug_call().calldata))
+            .or_else(|| {
+                self.debug_call().decoded.as_ref().and_then(|decoded| {
+                    let call_data = decoded.call_data.as_ref()?;
+                    call_data_signature_name(&call_data.signature)
```
mablr: Matches only on function name, not signature; overloaded functions may show parameters from the wrong variant.

crates/debugger/src/tui/draw.rs
```diff
@@ -583,6 +619,330 @@ impl TUIContext<'_> {
...
+    }
+
+    let inputs = &stack[stack.len() - types.len()..];
+    Some(
+        inputs
+            .iter()
+            .zip(types.iter())
+            .map(|(input, (ty, storage))| {
+                match (ty, storage) {
+                    (DynSolType::Uint(8), Some(Storage::Memory | Storage::Storage)) => None,
```
mablr: Storage-qualified scalar params have their stack slot index decoded as the declared type, displaying the slot number instead of the stored value. Not sure, please double-check.

crates/debugger/src/tui/draw.rs
```diff
@@ -583,6 +619,330 @@ impl TUIContext<'_> {
...
+    if types.is_empty() {
+        return Some(Vec::new());
+    }
+
+    let value = DynSolType::Tuple(types.to_vec()).abi_decode_sequence(data).ok()?;
+    let values = value.as_fixed_seq()?;
+    Some(values.iter().map(format_token).collect())
+}
+
+fn decode_from_memory(ty: &DynSolType, memory: &[u8], location: usize) -> Option<DynSolValue> {
```
mablr: duplicate of https://github.com/figtracer/foundry/blob/c5fd71b82fa294cce6a6edde506ae6bb30075d7c/crates/evm/traces/src/debug/mod.rs#L285 ?

crates/debugger/src/builder.rs
```diff
@@ -94,11 +97,36 @@ impl DebuggerBuilder {
...
+fn identify_internal_calls(
+    trace_arenas: &mut [CallTraceArena],
+    identified_contracts: &AddressHashMap<String>,
+    sources: &ContractSources,
+) {
+    if sources.artifacts_by_name.is_empty() {
+        return;
+    }
+
+    let identifier = DebugTraceIdentifier::new(sources.clone());
```
mablr: Clone not needed ?

crates/debugger/src/tui/draw.rs
```diff
@@ -583,6 +619,330 @@ impl TUIContext<'_> {
...
+        let mut active = None;
+
+        for (node_idx, node) in
+            self.debug_arena().iter().enumerate().take(current_node_idx.saturating_add(1))
+        {
+            if node.trace_node_idx != trace_node_idx {
+                continue;
+            }
+
+            for (step_idx, step) in node.steps.iter().enumerate() {
```
mablr, line 788: O(nodes × steps) scan per frame, could we use caching in `DrawMemory`?

figtracer: nice

crates/debugger/src/tui/draw.rs
```diff
@@ -583,6 +619,330 @@ impl TUIContext<'_> {
...
+
+    fn decode_internal_parameter_values(&self, scope: &DebugSourceScope) -> Option<Vec<String>> {
+        let active = self.active_internal_call()?;
+        if !decoded_internal_name_matches(&active.decoded.func_name, scope) {
+            return None;
+        }
+
+        active.decoded.args.clone().or_else(|| {
+            let step =
+                self.debug_arena().get(active.start_node_idx)?.steps.get(active.start_step)?;
```
grandizzy: This decodes parameters from the wrong step. `active.start_step` is where the `DecodedTraceStep::InternalCall` marker is attached — i.e. the JUMP *into* the internal function. At that point the top of stack still holds jump/return machinery, not the callee-visible arguments.

The existing identifier in `debug/mod.rs` decodes args from `start_idx + 1`:

```rust
try_decode_args_from_step(&t, &self.node.trace.steps[start_idx + 1])
```

so this fallback can show shifted/incorrect values — which is worse than `<unavailable>`.

Because the call may be split across flattened segments, resolve the entry step by absolute index rather than the local segment index, e.g.:

```rust
fn step_by_absolute_index(
    &self,
    trace_node_idx: usize,
    absolute_step: usize,
) -> Option<&CallTraceStep> {
    self.debug_arena()
        .iter()
        .filter(|n| n.trace_node_idx == trace_node_idx)
        .find_map(|n| n.steps.get(absolute_step.checked_sub(n.step_offset)?))
}
```

then:

```rust
let step = self.step_by_absolute_index(
    self.debug_call().trace_node_idx,
    active.start_step.saturating_add(1),
)?;
decode_step_parameters(&scope.parameters_src, step)
```

Note: `active.start_step` here is a node-local step index (set from `step_idx` in `active_internal_call`), so a bare `+1` only works correctly once the lookup is switched to absolute indexing as above — otherwise it could read past the end of a split segment.

grandizzy (dismissed):
Please check finding below.

grandizzy (approved):
Re-reviewed after the latest commits. The merge-blocking internal-args step bug is fixed correctly (absolute `entry_step` + `step_by_absolute_index` handling split segments, with a regression test). Also addressed: per-frame active-call caching in `DrawMemory`, borrowed `ContractSources` (no clone), full canonical signature matching for overloads, storage-location params marked unknown, and the duplicated memory decoder consolidated into the shared `decode_step_parameters`. LGTM.

mablr (approved):
sgtm

PR #15142, opened 2026-06-11
fix(forge): allow verify-bytecode to run without a block explorer

### LINKED ISSUE #13479: forge verify-bytecode command can not verify a contract without external services

PrazwalR:
## Motivation

`forge verify-bytecode` fails unconditionally when no block explorer is configured — three separate failure points: client construction dies on unknown chains, creation-data fetch errors with `InvalidApiKey`, source-code fetch errors with "Contract source code not verified". Private chains, custom L2s, and local anvil nodes are entirely excluded.

This supersedes #14583 (closed as stale — @figtracer asked for a regression test that was never added; approach is the same, credit to @ParthSinghPS).

Closes #13479.

## Changes

**`crates/verify/src/bytecode.rs`**
- Etherscan client construction now `Option<Client>` — unknown chains get `None`, no hard error
- Creation data fetch, source code fetch both wrapped to `None` when explorer unavailable
- EVM version falls back to local `foundry.toml` config when no explorer metadata
- Constructor args fall back to `Bytes::new()` when no source code available
- `if creation_data.is_none()` branch unified: covers both predeploy (genesis) and no-explorer cases — does runtime-only verification at `--block`/latest, emitting a warning instead of erroring

**`crates/verify/src/utils.rs`**
- `print_result` takes `Option<&Metadata>` — mismatch hints only printed when explorer metadata present

**`crates/forge/tests/cli/verify_bytecode.rs`**
- `can_verify_bytecode_without_explorer`: spawns local anvil, deploys Counter, runs `verify-bytecode` with no explorer args, asserts "Runtime code matched" in stdout and "Creation data is unavailable" in stderr — this is the regression test missing from #14583

## Testing

```bash
cargo test -p forge --test cli -- verify_bytecode::can_verify_bytecode_without_explorer

### CODE REVIEW

File: crates/verify/src/bytecode.rs [hunk 1]
```diff
@@ -266,41 +319,57 @@ impl VerifyBytecodeArgs {
...
+                // Deploy at genesis
+                0_u64
+            } else {
+                if !shell::is_json() {
+                    sh_warn!("Creation data is unavailable. Ignoring creation code verification.")?;
+                }
+                match self.block {
+                    Some(BlockId::Number(BlockNumberOrTag::Number(block))) => block,
+                    Some(_) => eyre::bail!("Invalid block number"),
+                    None => provider.get_block_number().await?,
```
figtracer, line 389: this picks a block for the local fork, but the runtime read below still falls back to latest. passing the selected block into `get_runtime_codes` would keep both sides anchored to the same height.

crates/verify/src/bytecode.rs
```diff
@@ -266,41 +319,57 @@ impl VerifyBytecodeArgs {
...
-            if !shell::is_json() {
-                sh_warn!(
-                    "Attempting to verify predeployed contract at {:?}. Ignoring creation code verification.",
-                    self.address
-                )?;
-            }
+        // Without creation data (predeploys, or the explorer being unavailable), the creation
+        // code cannot be verified. Verify the runtime bytecode instead by deploying the local
+        // creation code and comparing the resulting runtime code with the onchain one.
+        if creation_data.is_none() {
```
figtracer, line 361: this early return bypasses `--ignore runtime`, so the fallback still verifies runtime even when the user asked to skip it. consider honoring that flag here too.

crates/verify/src/bytecode.rs
```diff
@@ -223,25 +237,64 @@ impl VerifyBytecodeArgs {
...
         // Obtain Etherscan compilation metadata.
-        let etherscan_metadata = source_code.items.first().unwrap();
+        let etherscan_metadata = source_code.as_ref().map(|source| source.items.first().unwrap());
+
+        // The EVM version to verify against: the explorer-reported version when available,
+        // otherwise the local project configuration.
+        let evm_version = match etherscan_metadata {
+            Some(metadata) => metadata.evm_version()?.unwrap_or_default(),
+            None => config.evm_version,
+        };
```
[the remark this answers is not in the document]
PrazwalR: Fixed in 15334e8 — now and_then(|source| source.items.first()), empty items falls back to config.evm_version

crates/verify/src/bytecode.rs
```diff
@@ -266,41 +319,57 @@ impl VerifyBytecodeArgs {
 
         let mut constructor_args = if let Some(provided) = provided_constructor_args {
             provided.into()
-        } else {
+        } else if let Some(source_code) = &source_code {
             // If no constructor args were provided, try to retrieve them from the explorer.
             check_explorer_args(source_code.clone())?
+        } else {
+            Bytes::new()
         };
```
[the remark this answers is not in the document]
PrazwalR: Fixed in 15334e8 — check_explorer_args now takes &ContractMetadata, no clone.

crates/verify/src/bytecode.rs
```diff
@@ -266,41 +319,57 @@ impl VerifyBytecodeArgs {
...
+                deploy_block,
+                evm_version,
                 evm_opts,
             )
             .await?;
 
-            evm_env.block_env.set_number(U256::ZERO);
-            let genesis_block = provider.get_block(gen_blk_num.into()).full().await?;
+            evm_env.block_env.set_number(U256::from(deploy_block));
+            let genesis_block = provider.get_block(deploy_block.into()).full().await?;
```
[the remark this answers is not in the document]
PrazwalR: Fixed in 15334e8 — renamed to deploy_block_info

crates/verify/src/utils.rs
```diff
@@ -103,7 +103,7 @@ pub fn print_result(
     res: Option<VerificationType>,
     bytecode_type: BytecodeType,
     json_results: &mut Vec<JsonResult>,
-    etherscan_config: &Metadata,
+    etherscan_config: Option<&Metadata>,
```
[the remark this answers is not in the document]
PrazwalR: Fixed in 15334e8 — renamed to etherscan_metadata

crates/verify/src/bytecode.rs
PrazwalR: Fixed in 2eea384 — onchain read now anchored to deploy_block via get_runtime_codes. Predeploys kept at latest (genesis state needs an archive node); noted in a code comment

crates/verify/src/bytecode.rs
PrazwalR: Fixed in 2eea384 — fallback returns early when --ignore runtime set. Regression test extended to cover it

crates/verify/src/bytecode.rs
```diff
@@ -203,9 +203,23 @@ impl VerifyBytecodeArgs {
...
-            EtherscanVerificationProvider.client(&self.etherscan, &self.verifier, &config)?;
+        // Etherscan client. May be unavailable (e.g. unknown chain, missing configuration), in
+        // which case verification proceeds with local data only.
+        let etherscan = match EtherscanVerificationProvider.client(
+            &self.etherscan,
+            &self.verifier,
+            &config,
+        ) {
+            Ok(client) => Some(client),
+            Err(err) => {
```
figtracer, line 222: this catches every explorer client setup error, so an explicit bad `--verifier-url` is silently treated like no explorer. can we only fall back for expected no-explorer cases and still return user-supplied verifier config errors?

figtracer (commented):
another nit

crates/verify/src/bytecode.rs
PrazwalR: Fixed in 822d6243a  client setup errors now only fall back to no-explorer mode when nothing was configured (no --verifier/--verifier-url/--verifier-api-key and no resolved etherscan key). An explicitly configured verifier that fails to resolve surfaces the error. Added a test case asserting a bad --verifier-url fails instead of silently degrading.

mablr (approved):
SGTM

stevencartavia (approved):

PR #15153, opened 2026-06-12
fix(fmt): balance override variable indentation

### LINKED ISSUE #15151: bug(fmt): forge fmt incorrectly indents top-level declarations after override state variable without initializer
### Component

Forge

- [x] Foundry
- [ ] Foundryup

forge Version: 1.7.2-nightly Commit SHA: 1ebe9c26e156de7f3a562895c17c1fced2dd322e Build Timestamp: 2026-06-11T07:39:30.155950000Z (1781163570) Build Profile: dist

NA

forge fmt

### Operating System

None

### Describe the bug

format this code would cause `Info` have extra indent
```solidity
pragma solidity ^0.8.28;

contract BaseStorage {
  uint256 public total;
}

contract ChildStorage is BaseStorage {
  uint256 public override total;
}

struct Info {
  uint256 a;
}
```
after fmt the Info would be like this
```solidity
  struct Info {
    uint256 a;
  }
```

If we set default value for override the bug is gone `uint256 public override total = 0;`

Likely in crates/fmt/src/state/sol.rs around print_var and contract source-unit/box handling (override_ path differs from non-override handling).
https://github.com/foundry-rs/foundry/blob/a0b37374f5bad527749da20bf4550dd51f34e8bc/crates/fmt/src/state/sol.rs#L989-L999

figtracer:
Fixes #15151.

## Summary

- Balance the extra formatter box opened for state variables with an `override` attribute when the variable has no initializer.
- Add regression coverage for uninitialized override variables followed by top-level declarations, a following contract member, override path lists, and the initializer case that already worked.

## Root Cause

`print_var` opens an additional box while printing `override` attributes. The initializer path closes that box before formatting the RHS, but the no-initializer path only closed the regular variable boxes. That left the pretty-printer one indent level deep, so later source-unit items were emitted indented.

### DIFF
--- a/crates/fmt/src/state/sol.rs
+++ b/crates/fmt/src/state/sol.rs
@@ -1024,6 +1024,9 @@ impl<'ast> State<'_, 'ast> {
 
             self.print_assign_rhs(init, pre_init_size, init_space_left, Some(&ty.kind), cache);
         } else {
+            if override_.is_some() {
+                self.end();
+            }
             self.end();
         }
         self.end();
--- a/crates/fmt/tests/formatter.rs
+++ b/crates/fmt/tests/formatter.rs
@@ -250,3 +250,150 @@ contract ProofOfConcept {
 
     assert_eq!(formatted, expected, "Formatting mismatch");
 }
+
+#[test]
+fn test_override_state_variable_without_initializer_does_not_leak_indent() {
+    init_tracing();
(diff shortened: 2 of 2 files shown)


mablr (approved):

grandizzy (approved):
lgtm!

PR #15155, opened 2026-06-12
fix(debugger): auto-scroll memory writes

### LINKED ISSUE #1873: feat(`debugger`) cursor should auto-scroll to modified memory after an `MSTORE`
### Component

Forge

Currently, when an `MSTORE` happens, the debugger will highlight the portion of the memory that has changed with a red colored text.

The problem is that the memory section of the debugger is not auto-scrolling to that location, and you need to move via `CTRL+j/k` to do that manually.
Not knowing that the memory box doesn't auto-scroll, I thought that there was a bug in the `debug` feature because I was not seeing any change in the memory displayed in the UI. 

Would make sense to automatically scroll to that position when a `MSTORE` happen? 

### Additional context

_No response_

figtracer:
## Summary

Auto-scroll the debugger memory pane to the memory word touched by a write operation when navigating through trace steps.

This fixes a debugger UX issue where `MSTORE`/`MSTORE8`/memory-copy writes were highlighted but often stayed off-screen, which made it look like the memory pane was not changing unless the user manually scrolled with `ctrl+j/k`.

Fixes #1873.

### DIFF
--- a/crates/debugger/src/tui/context.rs
+++ b/crates/debugger/src/tui/context.rs
@@ -3,7 +3,7 @@
 use crate::{DebugNode, ExitReason, debugger::DebuggerContext};
 use alloy_primitives::{Address, hex};
 use crossterm::event::{Event, KeyCode, KeyEvent, KeyModifiers, MouseEvent, MouseEventKind};
-use foundry_evm_core::buffer::BufferKind;
+use foundry_evm_core::buffer::{BufferKind, get_buffer_accesses};
 use foundry_tui::TuiApp;
 use ratatui::Frame;
 use revm::bytecode::opcode::OpCode;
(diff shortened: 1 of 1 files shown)


grandizzy (approved):
Nice fix, this addresses #1873 well and the prev-step preference correctly mirrors the GREEN "just written" highlight in `draw_buffer` (the current memory snapshot reflects the previous opcode's effect). Bounding the target line by the current snapshot length is also the right call — it correctly defers scrolling for memory-expanding writes until the memory has actually grown. LGTM overall, just two non-blocking comments:

1. **Jumpiness**: `scroll_memory_to_current_write` always sets `current_buf_startline = line`, which re-pins the write to the top row on every step even when it's already visible (and drops the context above it). Consider only scrolling when `line` falls outside the current viewport `[startline, startline + height)`, ideally centering it. That needs the pane height plumbed out from `draw_buffer`.

2. **Stale scroll on no-target**: when `current_memory_write_line()` returns `None`, a previously-large `current_buf_startline` is left untouched, so stepping into a step with smaller/no memory can leave the pane blank (the `skip()` in `draw_buffer` skips everything). Clamping to `memory_len.div_ceil(32).saturating_sub(1)` on each memory navigation would avoid that.

Minor: the new tests cover the helper math and the "do nothing" branches well, but there's no test with an actual `memory: Some(...)` snapshot exercising a successful scroll (incl. prev-vs-current preference and the expansion-defer-then-scroll sequence). Would be good to add one.

mablr (approved):

PR #15157, opened 2026-06-12
feat(symbolic): model KZG precompile witnesses

figtracer:
## Summary
- model the Cancun KZG point-evaluation precompile (`0x0a`) with replayable symbolic success and failure witnesses
- delegate concrete KZG calls to the revm precompile implementation
- add regressions for concrete calls, symbolic invalid witnesses, symbolic success reachability, and no KZG unsupported false-pass output

## Validation
- `cargo +1.96.0 check -p foundry-evm-symbolic`
- `cargo +1.96.0 test -p foundry-evm-symbolic`
- `cargo +1.96.0 test -p forge --test cli symbolic_precompiles -- --nocapture`
- `cargo +1.96.0 test -p forge --test cli symbolic_false_pass_prevention -- --nocapture`
- `cargo fmt --check`


crates/evm/symbolic/src/runtime/precompiles.rs
```diff
@@ -162,7 +170,7 @@ pub(crate) fn execute_symbolic_precompile(
             }
             Ok(Some(symbolic_fixed_len_precompile_output("blake2f", input, input_len, 64)))
         }
-        Some(10) => Err(SymbolicError::Unsupported("KZG point-evaluation precompile not modeled")),
+        Some(10) => symbolic_kzg_point_evaluation_precompile(input, input_len),
```
grandizzy: **Dead code: this arm (and the function it calls) is now unreachable.**

KZG is intercepted in the executor at `calls.rs:937` (`precompile_number_for_spec(...) == Some(10)` → `execute_kzg_precompile_call`), which returns before `execute_symbolic_precompile` is ever called. So `execute_symbolic_precompile` is only reached for non-KZG numbers, making this `Some(10)` arm, `symbolic_kzg_point_evaluation_precompile`, and its only-caller helper `concrete_precompile_bytes_at` dead.

Since `execute_symbolic_precompile` is `pub(crate)` and called directly from tests, I'd keep an explicit guard here rather than deleting the arm (otherwise an addr-`0x0a` call falls into the `_` branch and gets mishandled as a generic concrete precompile):

```suggestion
        Some(10) => Err(SymbolicError::Unsupported(
            "KZG handled by execute_kzg_precompile_call",
        )),
```

Then also remove the now-orphaned `symbolic_kzg_point_evaluation_precompile` and `concrete_precompile_bytes_at` functions, and drop the `kzg_point_evaluation` import here if nothing else in this file references `VERSIONED_HASH_VERSION_KZG`/`kzg_to_versioned_hash` after removal.

grandizzy (dismissed):
Two non-blocking items:

1. **Spec gating is incomplete.** `precompile_number_for_spec` only gates `0x0a` by Cancun, but the README now says precompiles are "recognized according to the active EVM version." `0x05–0x08` (Byzantium) and `0x09` (Istanbul) stay active on older forks, so e.g. `--evm-version` < Istanbul models `0x09` as BLAKE2f instead of an empty account. Either extend the gating or soften the README wording.
2. **Dead code.** `symbolic_kzg_point_evaluation_precompile` (+ its `concrete_precompile_bytes_at` helper and the `Some(10)` arm) is now unreachable since the executor intercepts KZG earlier. See inline.

Approving — neither blocks merge.

grandizzy (approved):
lgtm!

mablr (approved):

PR #15158, opened 2026-06-12
feat(common): resolve tempo fee token symbols on-chain

mablr:
## Motivation

Towards OSS-165

Resolve fee token symbols on-chain


crates/common/src/tempo/mod.rs
```diff
@@ -66,6 +68,8 @@ pub const BETA_USD_ADDRESS: Address = address!("0x20C000000000000000000000000000
 pub const THETA_USD_ADDRESS: Address = address!("0x20C0000000000000000000000000000000000003");
 
 /// Resolves an explicit Tempo fee token or the canonical default for a known Tempo network.
+///
+/// TODO: fee token resolution is incomplete, must use FeeTokenManager.
```
figtracer, line 72: can we address this now?

mablr: Follow-up coming soon, it's orthogonal to symbol resolution.

crates/common/src/tempo/mod.rs
```diff
@@ -84,29 +88,53 @@ pub const fn known_fee_token_symbol(fee_token: Address) -> Option<&'static str>
...
+    P: Provider<N>,
+{
+    if let Some(symbol) = known_fee_token_symbol(fee_token) {
+        return Some(symbol.to_string());
     }
+
+    let tx = N::TransactionRequest::default()
+        .with_to(fee_token)
+        .with_input(ITIP20::symbolCall.abi_encode());
+    let output = provider.call(tx).await.ok()?;
```
stevencartavia, line 104: `--curl` produces the wrong output for custom fee tokens.

To display a symbol for fee tokens not in the known-token list, `resolve_fee_token_symbol` now performs a provider `eth_call` to `symbol()` from the fee-token print path — which previously did no provider/RPC I/O. In `--curl` mode, `CurlTransport` prints the curl for the *first* JSON-RPC request and then `process::exit(0)`, so in fully-specified flows like the repro below the symbol lookup becomes the first RPC for a custom/unknown token: `--curl` prints the `eth_call symbol()` and exits before the intended `eth_sendRawTransaction` curl is produced.

Repro:
```
cast send 0x0000000000000000000000000000000000001234 --chain tempo \
  --rpc-url https://example.invalid --curl --private-key <key> \
  --tempo.fee-token 0x20C00000000000000000000014f22CA97301EB73 \
  --nonce 0 --gas-limit 100000 --gas-price 1000000000 --priority-gas-price 1000000000 --value 0
```

- master: emits the `eth_sendRawTransaction` curl
- this PR: emits `eth_call` -> `symbol()` (`0x95d89b41`) then exits — send curl never produced

Same with `cast mktx --raw-unsigned --curl`: the `symbol()` curl is printed and the raw tx hex is never produced. Known/default tokens (PathUSD, AlphaUSD, BetaUSD, ThetaUSD) short-circuit without an RPC and are unaffected — only tokens outside the known-token list regress.

mablr: Good catch, `--curl` creates tons of sharp edges like this :/

Fixed in 243d0f1

stevencartavia (approved):

0xKarl98 (approved):

PR #15159, opened 2026-06-12
feat(cli): parse tempo fee token symbols

mablr:
## Summary

  Adds shared parsing for Tempo fee-token symbols in CLI arguments. Towards OSS-165.

  - Accepts `PathUSD`, `AlphaUSD`, `BetaUSD`, and `ThetaUSD` symbols case-insensitively
  - Keeps raw address and numeric TIP-20 token ID parsing unchanged
  - Supersedes former `keychain policy/session` parsing logic (`parse_policy_token`)


crates/cli/src/utils/tempo.rs
```diff
@@ -39,9 +42,33 @@ pub struct ResolvedLane {
...
+                    "invalid fee token '{symbol_or_address}': expected address, numeric TIP-20 token \
+                     id, or one of PathUSD, AlphaUSD, BetaUSD, ThetaUSD: {e}"
+                )
+            })
+        },
+        Ok,
+    )
+}
+
+fn parse_fee_token_symbol(symbol: &str) -> Option<Address> {
```
stevencartavia, line 64: Non-blocking: since `foundry_common::tempo` already owns the known Tempo fee-token addresses and `known_fee_token_symbol`, would it make sense for `parse_fee_token_symbol` to live there too? It doesn't need to be a strict inverse — parser-only aliases (`usd`, `default`, `path-usd`, …) can still be supported — but keeping the canonical symbol/address mapping in one module reduces drift when a new fee token is added. `foundry_cli` already depends on `foundry_common::tempo`, so this would mainly be a locality/maintenance improvement.

mablr: Fair point, but I think it's good as is.

stevencartavia: cool

stevencartavia (approved):

0xKarl98 (approved):
LGTM

PR #15160, opened 2026-06-12
fix: add support for `--jwt-secret` in cast `--curl`

LeTamanoir:
## Motivation

Add support for `--jwt-secret` in cast `--curl` output so users can generate authenticated RPC curl commands.

## Solution

Introduce JWT generation from a hex secret using `JwtSecret` + `Claims::default()` and inject the resulting token as an `Authorization: Bearer ...` header in the curl transport. Wire it through both `call` and `rpc` CLI paths and update the transport to accept a JWT secret instead of a raw token. Add tests validating curl output includes the correct Authorization header.


LeTamanoir:
I am a total Rust noob, so if I did something abominable, I am sorry in advance.

crates/cast/tests/cli/main.rs
```diff
@@ -5619,6 +5619,57 @@ casttest!(curl_rpc, |_prj, cmd| {
...
+        .get_output()
+        .stdout_lossy();
+
+    // Verify curl command structure
+    assert!(output.contains("curl -X POST"));
+    assert!(output.contains("-H 'Content-Type: application/json'"));
+    assert!(output.contains("eth_call"));
+    assert!(output.contains("jsonrpc"));
+    assert!(output.contains(rpc));
+    assert!(output.contains("Authorization: Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9."));
```
stevencartavia: This only checks the constant `eyJ...` (`{"typ":"JWT","alg":"HS256"}`) header prefix, so it passes for any token — even one signed with the wrong secret. Validate the token against the secret with `JwtSecret::validate`. (The `test_curl_with_jwt` unit test also fails CI: it passes a non-hex value — use a real 64-hex secret.)

crates/common/src/provider/curl_transport.rs
```diff
@@ -34,8 +44,10 @@ pub fn generate_curl_command(
     let mut cmd = String::from("curl -X POST");
     cmd.push_str(" -H 'Content-Type: application/json'");
 
-    if let Some(jwt) = jwt {
-        cmd.push_str(&format!(" -H 'Authorization: Bearer {}'", shell_escape(jwt)));
+    if let Some(jwt_secret) = jwt_secret {
```
stevencartavia: Fix the clippy `collapsible_if` warning here (also at L105): `cargo clippy --workspace --all-targets --all-features --locked`

LeTamanoir:
Hey @stevencartavia, thanks a lot for the review, I believe I applied all the fixes you mentioned 👍

crates/common/src/provider/curl_transport.rs
```diff
@@ -11,6 +12,15 @@ fn shell_escape(s: &str) -> String {
...
 }
 
+/// Build a JWT using a secret
+fn build_jwt(jwt_secret: &str) -> eyre::Result<String> {
+    // Decode jwt from hex, then generate claims (iat with current timestamp)
+    let secret = JwtSecret::from_hex(jwt_secret)?;
+    let claims = Claims::default();
+    let token = secret.encode(&claims)?;
+    Ok(token)
+}
```
stevencartavia, line 23: Now that the value is treated as a hex secret here (`from_hex` + `encode`), the existing `test_curl_with_jwt` unit test (L204-210) fails: it passes a non-hex value `"my-jwt-token"` and asserts that exact token, but `build_jwt` returns an error so the header is not added. This is the `test / test all` CI failure: `assertion failed: cmd.contains("Authorization: Bearer my-jwt-token")`. Update that test to use a 64-hex secret and validate the emitted token with `JwtSecret::validate` — the token has a time-based `iat`, so an exact string match won't work.

crates/common/src/provider/curl_transport.rs
```diff
@@ -34,8 +44,10 @@ pub fn generate_curl_command(
     let mut cmd = String::from("curl -X POST");
     cmd.push_str(" -H 'Content-Type: application/json'");
 
-    if let Some(jwt) = jwt {
-        cmd.push_str(&format!(" -H 'Authorization: Bearer {}'", shell_escape(jwt)));
+    if let Some(jwt_secret) = jwt_secret
+        && let Ok(jwt) = build_jwt(jwt_secret)
```
stevencartavia: `if let Ok(jwt) = build_jwt(..)` ignores the error: an invalid `--jwt-secret` produces a curl command with no `Authorization` header and exit 0, while the normal RPC path returns `Invalid JWT: ...`. Since the user passed `--jwt-secret`, this should return an error instead of omitting the header. Same at L105-106 — a shared helper that propagates the error would cover both.

crates/common/src/provider/curl_transport.rs
LeTamanoir: Hey, sorry I didn't understand the 1st time you said it, now I do, will update the tests 👍

crates/common/src/provider/curl_transport.rs
LeTamanoir: I added the error case, not sure about the way I return it to the user at L128-142 though, wdyt?

stevencartavia (commented):
please run `cargo +nightly fmt`

stevencartavia (approved):

mablr (approved):
Makes sense, thank you.

PR #15161, opened 2026-06-12
feat(common): resolve fee token on chain

mablr:
## Summary

Resolve Tempo fee-token selection on-chain before signing or submitting transactions.

This updates cast/forge/script transaction paths so Tempo transactions without an explicit `--tempo.fee-token` query the fee manager for the relevant fee payer's configured token and stamp it onto the transaction before it is signed, sent, or used to compute a sponsor digest.

Implements 1-2-5 steps https://docs.tempo.xyz/protocol/fees/spec-fee#fee-token-preferences (3-4 will be done later)

## Details

- Resolve `userTokens(fee_payer)` on Tempo chains before falling back to the network default fee token.
- Preserve explicit `--tempo.fee-token` behavior.
- Keep `--curl` display paths from performing extra lookup RPCs where possible.
- Resolve local sponsor fee tokens using `sponsor.sponsor()` before `attach_and_print` / sponsor digest computation.
- Leave sponsor-url behavior unchanged, since the local client does not necessarily know the remote sponsor address.
- Add tests for default resolution, explicit token precedence, lookup error handling, curl-mode fallback, and sponsor fee-token resolution.

## Example
By default:
```bash
$ cast send 0x20c0000000000000000000000000000000000000 "transfer(address, uint256)" 0x0000000000000000000000000000000000000001 100 --private-key $PK --rpc-url 'https://rpc.moderato.tempo.xyz'
Paying gas in PathUSD (0x20C0000000000000000000000000000000000000)

...
```

If user has set ThetaUSD as preferred fee-token:
```bash
cast call 0xfeec000000000000000000000000000000000000 "userTokens(address)(address)" $ADDRESS --rpc-url https://rpc.moderato.tempo.xyz

0x20C0000000000000000000000000000000000003
```

The fee token is properly resolved and displayed.
```bash
$ cast send 0x20c0000000000000000000000000000000000000 "transfer(address, uint256)" 0x0000000000000000000000000000000000000001 100 --private-key $PK --rpc-url 'https://rpc.moderato.tempo.xyz'
Paying gas in ThetaUSD (0x20C0000000000000000000000000000000000003)

...
```


crates/common/src/tempo/mod.rs
```diff
@@ -67,18 +69,85 @@ pub const ALPHA_USD_ADDRESS: Address = address!("0x20C00000000000000000000000000
...
+    }
+
+    Ok(Some(DEFAULT_FEE_TOKEN))
+}
+
+/// Resolves and applies the Tempo fee token selected by the network.
 ///
-/// TODO: fee token resolution is incomplete, must use FeeTokenManager.
-pub fn resolve_fee_token(
+/// This must happen before computing a sponsor digest, because Tempo sponsor signatures commit to
```
figtracer, line 72: this contract is not held by the sponsored paths yet. live repro on `anvil --network tempo --chain-id 4329`: `userTokens(0x7099...)` returns betausd, but `cast send --tempo.sponsor 0x7099... --tempo.sponsor-signer private-key://...` prints `tempo fee token: network default`; adding `--tempo.fee-token 0x20c...0002` prints betausd and signs a different digest. `--tempo.print-sponsor-hash --tempo.sponsor 0x7099...` shows the same mismatch: no explicit token prints `0xa221...0e46`, explicit betausd prints `0x0ff4...d668`. can we resolve with `fee_payer = sponsor.sponsor()` before `attach_and_print` / `compute_sponsor_hash` for local sponsor flows? sponsor-url can stay separate since the client may not know that sponsor address.

mablr: Yes i removed sponsored paths for now (see commits history). I'm not sure how to handle them properly while keeping things clean/factorized.

As far as i understand in most cases sponsor-url will be used so we don't know address to resolve, right ? Broadly speaking,fro sponsored tx does the user actually care about fee token ?

mablr: Made an attempt here https://github.com/foundry-rs/foundry/pull/15161/commits/5e4ca70634015a4c709dbe8985079a3e1157731e ptal @figtracer

mattsse (approved):
LGTM

figtracer (approved):
tested more and it lgtm

Issue #14475, opened 2026-04-27
feat(forge verify-contract): expose `licenseType` so Etherscan-verified contracts don't end up with empty `LicenseType`


0xLaz3r:
### Component

Forge

### Describe the feature you would like

`forge verify-contract` does not expose any way to pass `licenseType` to Etherscan's `verifysourcecode` endpoint, so contracts verified via Foundry end up with an empty `LicenseType` field on Etherscan.

The pre-existing direct-API verification flow (e.g. older `verify.py` scripts) used to POST `licenseType=13` for `AGPL-3.0-or-later` (or `5` for `GPL-3.0-or-later`, etc., per Etherscan's [supported license types](https://docs.etherscan.io/etherscan-v2/api-endpoints/contracts#verify-source-code)). After migrating to `forge verify-contract`, `result[0].LicenseType` from the v2 `getsourcecode` API comes back as `""`.

Repro:

```sh
forge verify-contract \
    <ADDR> src/MyContract.sol:MyContract \
    --verifier etherscan \
    --flatten \
    --watch \
    --etherscan-api-key $ETHERSCAN_API_KEY
```

Then:

```sh
curl -s "https://api.etherscan.io/v2/api?chainid=1&module=contract&action=getsourcecode&address=<ADDR>&apikey=$ETHERSCAN_API_KEY" \
  | jq '.result[0].LicenseType'
# → ""
```

Expected: `LicenseType` is set to whatever SPDX license is declared in the source (e.g. `"GNU AGPLv3"`).

Actual: empty string. Etherscan's UI sometimes back-fills it from the `// SPDX-License-Identifier: …` comment via a heuristic, but the API field stays blank, which breaks downstream automation that asserts on `LicenseType` (e.g. license-policy CI checks).

Root cause in the codebase:

- `crates/block-explorers/src/verify.rs` in `foundry-rs/block-explorers` — the `VerifyContract` struct has no `license_type` field. It only exposes `address`, `source`, `code_format`, `contract_name`, `compiler_version`, `optimization_used`, `runs`, `constructor_arguments`, `evm_version`, `via_ir`, plus a `#[serde(flatten)] other: HashMap<String, String>` catch-all.
- `crates/verify/src/etherscan/mod.rs` in `foundry-rs/foundry` — the function that builds `VerifyContract` from `VerifyArgs` populates the typed fields but never writes a `licenseType` key into `other`.

[...]

PR #15162, opened 2026-06-12
feat(verify): accept SPDX identifiers for --license-type

maksim-romanov:
## Summary
- Accept common SPDX identifiers (`MIT`, `Apache-2.0`, `AGPL-3.0-or-later`, …) in
  `--license-type` alongside the existing numeric form.
- Normalize input (lowercase, `_` / whitespace → `-`, collapsed dashes) before matching.
- Validate numeric input against the `1..=14` range.
- Reject invalid input at clap parse time with an actionable error listing example IDs.

#14975 added `--license-type` in its minimal form (numeric only). This layers SPDX
parsing on top so users do not have to look up Etherscan's numeric table.
#14578 took an earlier swing at this but went stale without review.

## Scope
- No auto-detection from `// SPDX-License-Identifier:` in source — kept as
  separate future work.
- No schema change on `block-explorers::VerifyContract`; `licenseType` still goes
  through the existing `other` flatten map.
- Numeric form (`--license-type 13`) is fully backwards-compatible.

## Tests
- `cargo check -p forge-verify --tests --lib`
- `cargo fmt -p forge-verify --check`
- `cargo test -p forge-verify --lib license_type` — 5 passed:
  - `applies_license_type_to_verify_request`
  - `can_parse_verify_contract_license_type_spdx`
  - `errors_on_invalid_verify_contract_license_type`
  - `verify_contract_license_type_accepts_range_bounds`
  - `verify_contract_license_type_is_case_insensitive`

Refs #14475.


figtracer (dismissed):
lgtm

stevencartavia (dismissed):

crates/verify/src/verify.rs
```diff
@@ -423,6 +430,76 @@ pub struct VerifyArgs {
     pub language: Option<ContractLanguage>,
 }
 
+fn parse_etherscan_license_type(value: &str) -> Result<String, String> {
```
DaniPopes, line 434: can you please link to etherscan reference for this mapping? https://etherscan.io/contract-license-types

crates/verify/src/verify.rs
```diff
@@ -423,6 +430,76 @@ pub struct VerifyArgs {
...
+    let value = value.trim();
+    if value.is_empty() {
+        return Err("license type cannot be empty".into());
+    }
+
+    if let Ok(code) = value.parse::<u8>() {
+        if (1..=14).contains(&code) {
```
DaniPopes: I'm guessing they can add more codes so I'm not sure if we should cap at 14

crates/verify/src/verify.rs
maksim-romanov: Added the URL to the field doc comment in a824550 so it surfaces in `--help`

crates/verify/src/verify.rs
maksim-romanov: Agreed dropped the cap in a824550; now accepts any positive integer and Etherscan validates server-side

DaniPopes (approved):

mablr (approved):

PR #15170, opened 2026-06-15
fix(script): honour --code-size-limit CLI flag

shapeshed:
`check_contract_sizes()` ignored the `--code-size-limit CLI` flag, falling back to EIP-170's 24576-byte default. This PR fixes the issue it by reading `evm_opts.env.code_size_limit` before `config.code_size_limit`.

## Motivation

`check_contract_sizes()` built its `size_limits` solely from `config.code_size_limit` (foundry.toml). The `--code-size-limit` CLI flag populates `evm_opts.env.code_size_limit` but was never read here, so the check silently fell back to EIP-170's 24576-byte default even when the flag was explicitly set.

## Solution

Chain `.or(evm_opts.env.code_size_limit)` before the config lookup so the CLI flag takes precedence over foundry.toml. A test asserts the flag is correctly threaded into `evm_opts` (the path the check now reads).

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/script/src/lib.rs
+++ b/crates/script/src/lib.rs
@@ -432,8 +432,10 @@ impl ScriptArgs {
 
             let size_limits = pre_simulation
                 .script_config
-                .config
+                .evm_opts
+                .env
                 .code_size_limit
+                .or(pre_simulation.script_config.config.code_size_limit)
                 .map(ContractSizeLimits::with_runtime_limit)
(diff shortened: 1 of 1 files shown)


mablr (approved):
Makes sense

grandizzy (approved):
lgtm, thank you!

PR #15171, opened 2026-06-15
feat(debugger): add layout selection

### LINKED ISSUE #8256: Debugger UI not split into four quadrants
### Component

Forge

- [X] Foundry
- [X] Foundryup

forge 0.2.0 (ba9fa20 2024-06-24T00:20:52.822818467Z)

forge debug

### Operating System

Linux

### Describe the bug

The debugger layout in the [document](https://book.getfoundry.sh/forge/debugger#debugger-layout) looks like:
image

But mine actually looks like:
image

My display resolution is 1920x1080, the debugger window is maximized.  I tried both Wezterm and Kitty terminal, no difference.

How to split it into four quadrants?

OS: Fedora 40, KDE

figtracer:
## Summary
- add debugger layout selection with auto, horizontal, and vertical modes
- wire `--debug-layout` into `forge test --debug` and `forge script --debug`
- add an in-TUI `l` shortcut to cycle layouts while debugging

Closes #8256

## Validation
- `cargo +nightly fmt --all --check`
- `cargo +1.95.0 clippy -p foundry-debugger --all-targets --all-features --locked -- -D warnings`
- `cargo +1.95.0 check -p foundry-debugger -p forge -p forge-script`
- `cargo +1.95.0 build -p forge --bin forge`
- live PTY smoke at 120 columns: `forge test --debug --debug-layout horizontal --match-test test_Increment` opened the two-column debugger layout and exited cleanly with `q`
- `git diff --check`

### DIFF
--- a/crates/debugger/src/debugger.rs
+++ b/crates/debugger/src/debugger.rs
@@ -7,6 +7,36 @@ use foundry_evm_core::Breakpoints;
 use foundry_evm_traces::debug::ContractSources;
 use std::path::Path;
 
+/// Debugger TUI layout selection.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub enum DebuggerLayout {
+    /// Select horizontal or vertical layout from the terminal size.
+    #[default]
+    Auto,
+    /// Force the two-column debugger layout.
+    Horizontal,
+    /// Force the single-column debugger layout.
+    Vertical,
+}
+
+impl DebuggerLayout {
+    pub(crate) const fn next(self) -> Self {
+        match self {
+            Self::Auto => Self::Horizontal,
+            Self::Horizontal => Self::Vertical,
+            Self::Vertical => Self::Auto,
+        }
+    }
+
+    pub(crate) const fn as_str(self) -> &'static str {
+        match self {
+            Self::Auto => "auto",
+            Self::Horizontal => "horizontal",
+            Self::Vertical => "vertical",
+        }
+    }
+}
+
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
 pub struct DebuggerStats {
     /// Sum of root-call gas used across every trace arena passed to the debugger.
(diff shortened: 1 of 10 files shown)


crates/script/src/lib.rs
```diff
@@ -251,6 +256,24 @@ pub struct ScriptArgs {
...
+
+impl From<DebuggerLayoutArg> for DebuggerLayout {
+    fn from(value: DebuggerLayoutArg) -> Self {
+        match value {
+            DebuggerLayoutArg::Auto => Self::Auto,
+            DebuggerLayoutArg::Horizontal => Self::Horizontal,
+            DebuggerLayoutArg::Vertical => Self::Vertical,
+        }
+    }
+}
```
grandizzy: This `DebuggerLayoutArg` enum and its `From<DebuggerLayoutArg> for DebuggerLayout` impl are copy-pasted verbatim here and in `crates/forge/src/cmd/test/mod.rs`. Let's avoid the duplication since it'll drift over time. Either derive/impl `clap::ValueEnum` directly on `DebuggerLayout` in the debugger crate (drops the wrapper enum + both `From` impls), or define the arg type once in a shared location and reuse it in both commands.

crates/debugger/src/debugger.rs
```diff
@@ -7,6 +7,36 @@ use foundry_evm_core::Breakpoints;
...
+}
+
+impl DebuggerLayout {
+    pub(crate) const fn next(self) -> Self {
+        match self {
+            Self::Auto => Self::Horizontal,
+            Self::Horizontal => Self::Vertical,
+            Self::Vertical => Self::Auto,
+        }
+    }
```
grandizzy, line 29: `next()` cycles back to `Auto`, and `cycle_layout` then prints `"Debugger layout: auto"`. That status is ambiguous to the user since the concrete layout under `Auto` depends on terminal width and is only resolved later in `draw`. Consider dropping `Auto` from the in-TUI cycle, or showing the resolved layout (e.g. `auto (horizontal)`).

grandizzy (changes requested):
Looks good overall, left some comments.

grandizzy (approved):
Thanks for addressing the feedback. Both points are resolved: `ValueEnum` is now derived directly on `DebuggerLayout` (removing the duplicated wrapper enum + `From` impls), and the `l` cycle skips `Auto` so the status is never ambiguous, with a test covering it. LGTM.

mablr (approved):
nice

PR #15172, opened 2026-06-15
feat(symbolic): model ADDMOD and MULMOD

figtracer:
## Summary

- model EVM ADDMOD and MULMOD with exact zero-modulus semantics and full-width intermediates
- emit SMT that zero-extends operands before modular reduction instead of using wrapped 256-bit intermediates
- validate hard-arithmetic fallback witnesses against the exact expression semantics

## Tests

- cargo +1.95.0 test -p foundry-evm-symbolic exact_
- cargo +1.95.0 test -p foundry-evm-symbolic


grandizzy (approved):
lgtm

mablr (approved):

PR #15179, opened 2026-06-15
feat(symbolic): Cancun `SELFDESTRUCT` semantics

mablr:
## Summary

Models fork-aware symbolic `SELFDESTRUCT` semantics for Cancun/EIP-6780.

Closes OSS-365

## Changes

- Pre-Cancun and same-transaction-created contracts still delete.
- Cancun+ existing contracts transfer balance and halt while preserving code/storage.
- Transaction-scoped created accounts are tracked for this decision.

## Tests

- Added Cancun preservation, same-transaction deletion, and false-pass prevention coverage.
- Added unit coverage for created-account tracking across snapshots.

### DIFF
--- a/crates/evm/symbolic/README.md
+++ b/crates/evm/symbolic/README.md
@@ -110,6 +110,11 @@ Current modeling notes:
   incomplete results with an explanatory reason.
 - Symbolic `KECCAK256` supports common Solidity storage patterns; arbitrary
   symbolic hashing may require heuristics and can make a run incomplete.
+- `SELFDESTRUCT` follows the active fork. Before Cancun it deletes the account;
+  from Cancun onward it only deletes contracts created in the same top-level
+  symbolic transaction, otherwise it transfers balance and halts while
+  preserving code and storage. Beneficiaries must resolve to concrete or
+  representative symbolic addresses.
 - Counterexamples are shown only after successful concrete replay.
 
 ## Writing Symbolic Tests
@@ -476,7 +481,7 @@ Known incomplete, bounded, or approximate surfaces include:
 | Area | Current behavior |
 |---|---|
 | Gas-dependent behavior | The engine does not use gas to prove properties. A raw `GAS` / `gasleft()` value is tolerated only as the direct gas operand to a CALL-family opcode and is not used to model gas availability. Explicit CALL-family gas caps are not enforced. Branches, arithmetic, call targets/values, calldata/returndata, memory/log offsets or sizes, `expectCall` gas matching, or solver constraints derived from observed gas report incomplete. Non-observable gas metering helpers are accepted as no-ops; observable gas read/snapshot helpers such as `lastCallGas`, `snapshotGasLastCall`, and `stopSnapshotGas` report incomplete and should not be used as symbolic properties. |
-| Cancun+ `SELFDESTRUCT` | Pre-Cancun deletion is modeled. Cancun/EIP-6780 same-transaction creation/deletion and end-of-transaction finalization are not fully modeled, so Cancun+ `SELFDESTRUCT` reports incomplete. |
+| `SELFDESTRUCT` | Pre-Cancun deletion is modeled. Cancun/EIP-6780 is modeled for concrete and resolved symbolic beneficiaries: contracts created in the current top-level symbolic transaction are deleted, while existing contracts transfer balance and halt without deleting code or storage. |
 | Symbolic account/code queries | `BALANCE`, `EXTCODESIZE`, `EXTCODEHASH`, and `EXTCODECOPY` on symbolic addresses are scoped to the engine's known symbolic/overlay/code-cache candidates plus the documented empty-account fallback. They do not prove quantified properties over every possible fork/backend account. |
 | Symbolic CALL targets | Concrete targets and symbolic targets constrained to known deployed-contract/precompile candidates are supported. By default, a feasible symbolic target outside the known candidate set reports incomplete. With `symbolic_call_targets = true`, the outside-candidate branch is modeled as an empty-account/no-code successful call, including value transfer for `CALL`; it does not model arbitrary unknown external code or custom/future precompiles. Symbolic cheatcode addresses/selectors still report incomplete. |
 | Symbolic CREATE / CREATE2 inputs | Concrete initcode and common bounded symbolic CREATE2 address expressions are supported. Symbolic runtime sizes and unsupported symbolic initcode shapes report incomplete. |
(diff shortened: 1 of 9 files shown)


crates/evm/symbolic/src/runtime/state.rs
```diff
@@ -1357,6 +1374,22 @@ impl SymbolicWorld {
...
 
+    /// Implements Cancun+ `SELFDESTRUCT` semantics for accounts not created in the current tx.
+    pub(crate) fn selfdestruct_cancun_existing<FEN: FoundryEvmNetwork>(
+        &mut self,
+        executor: &Executor<FEN>,
+        address: Address,
+        beneficiary: Address,
+    ) {
+        let balance = self.balance_word_for_address(executor, address);
+        if beneficiary != address && !matches!(balance, SymWord::Concrete(value) if value.is_zero())
```
grandizzy, line 1388: **Potential false PASS for symbolic beneficiaries.** `beneficiary` comes from `pop_address_or_symbolic_slot` → `address_or_symbolic_slot`, which for an unconstrained symbolic word returns a fresh `representative_symbolic_address` that is never equal to `state.address`. So this branch never models the `beneficiary == self` case, where EIP-6780 requires the existing contract to *retain* its balance (no net change). A test asserting balance preservation on a self-targeted selfdestruct with a symbolic beneficiary would PASS incorrectly.

Minimal fix: when the beneficiary is symbolic and may alias `state.address`, return incomplete. Better: fork on `beneficiary_word == state.address_word` — self branch is a no-op, other branch transfers and zeroes the source.

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -601,11 +601,14 @@ impl SymbolicExecutor {
...
-                    return Err(SymbolicError::Unsupported("SELFDESTRUCT/EIP-6780 not modeled"));
-                }
                 let beneficiary = state.pop_address_or_symbolic_slot()?;
-                state.world.selfdestruct(executor, state.address, beneficiary)?;
+                if spec_id < SpecId::CANCUN
+                    || state.world.was_created_in_current_transaction(state.address)
+                {
+                    state.world.selfdestruct_legacy(executor, state.address, beneficiary)?;
+                } else {
+                    state.world.selfdestruct_cancun_existing(executor, state.address, beneficiary);
```
grandizzy, line 615: Since this removes the `Unsupported("SELFDESTRUCT/EIP-6780 not modeled")` path entirely, the symbolic-beneficiary aliasing gap (see state.rs `selfdestruct_cancun_existing`) is no longer caught anywhere. Consider gating the symbolic-beneficiary case here (resolve concreteness before dispatch) so unconstrained symbolic targets still report incomplete rather than silently using a representative address.

crates/evm/symbolic/README.md
```diff
@@ -476,7 +481,7 @@ Known incomplete, bounded, or approximate surfaces include:
...
 | Gas-dependent behavior | The engine does not use gas to prove properties. A raw `GAS` / `gasleft()` value is tolerated only as the direct gas operand to a CALL-family opcode and is not used to model gas availability. Explicit CALL-family gas caps are not enforced. Branches, arithmetic, call targets/values, calldata/returndata, memory/log offsets or sizes, `expectCall` gas matching, or solver constraints derived from observed gas report incomplete. Non-observable gas metering helpers are accepted as no-ops; observable gas read/snapshot helpers such as `lastCallGas`, `snapshotGasLastCall`, and `stopSnapshotGas` report incomplete and should not be used as symbolic properties. |
-| Cancun+ `SELFDESTRUCT` | Pre-Cancun deletion is modeled. Cancun/EIP-6780 same-transaction creation/deletion and end-of-transaction finalization are not fully modeled, so Cancun+ `SELFDESTRUCT` reports incomplete. |
+| `SELFDESTRUCT` | Pre-Cancun deletion is modeled. Cancun/EIP-6780 is modeled for concrete and resolved symbolic beneficiaries: contracts created in the current top-level symbolic transaction are deleted, while existing contracts transfer balance and halt without deleting code or storage. |
```
grandizzy: "resolved symbolic beneficiaries" overstates soundness — representative resolution does not prove the property over all possible beneficiary aliases (notably `beneficiary == self`). Suggest explicitly noting that symbolic beneficiaries are modeled via a single representative address and do not branch over aliasing with the selfdestructing account.

crates/evm/symbolic/src/runtime/state.rs
```diff
@@ -1357,6 +1374,22 @@ impl SymbolicWorld {
...
+        {
+            let beneficiary_balance = self.balance_word_for_address(executor, beneficiary);
+            self.set_balance_word(beneficiary, sym_add(beneficiary_balance, balance));
```
grandizzy, line 1393: The `!matches!(balance, SymWord::Concrete(value) if value.is_zero())` guard treats a symbolic balance as non-zero, so `set_balance_word(beneficiary, …)` marks the beneficiary as `existing` even when the symbolic balance could be 0. Transferring 0 is algebraically harmless, but the existence side-effect can affect account-existence properties. Consistent with existing transfer approximations — worth a one-line comment noting it's intentional/approximate.

grandizzy (changes requested):
One blocking concern: **symbolic beneficiaries can produce a false PASS.** An unconstrained symbolic beneficiary resolves to a synthetic representative address that is never equal to `state.address`, so the `beneficiary != address` guard is always true and the `beneficiary == self` branch is never explored. For an *existing* Cancun contract, real EVM keeps the balance when the target is itself — the model transfers it away and can prove a property that doesn't actually hold. This matters more now that `symbolic_selfdestruct_accepts_symbolic_beneficiary` flips from incomplete → `[PASS]`.

Suggested resolution: either report incomplete when a symbolic beneficiary may alias `state.address`, or fork on `beneficiary_word == state.address_word`. A few extra tests (self-beneficiary no-burn, CREATE2 same-tx, reverted-CREATE, sequence boundary) would also round out coverage. Details inline.

crates/evm/symbolic/src/runtime/state.rs
mablr: minimal fix 858573cc6

crates/evm/symbolic/src/executor/opcodes.rs
mablr: 858573cc6

crates/evm/symbolic/src/runtime/state.rs
mablr: 5c39dd791

crates/evm/symbolic/README.md
mablr: 5c39dd791

grandizzy (approved):
lgtm!

figtracer (approved):
lgtm

PR #15180, opened 2026-06-15
perf(symbolic): prune selectors and model deployCode

figtracer:
## Summary
- fold equality against known-word expressions, avoiding solver work for concrete selector/calldata comparisons
- model `vm.deployCode(string)` and `vm.deployCode(string,bytes)` in symbolic execution by running artifact initcode and installing the returned runtime code
- add artifact path fallback for common Foundry `src/File.sol:Contract` references used by deployment cheatcodes

## Context
These changes improve symbolic testing for regular Foundry test suites. Selector and calldata equality checks can now collapse before reaching the solver when the compared expression is already known, and tests that deploy fixture contracts through `vm.deployCode` can continue through symbolic execution instead of stopping at an unsupported cheatcode.

The local Foundry-vs-Halmos run below validates the changes against an external bug-finding corpus by checking that Foundry finds replay-confirmed counterexamples on the same `check*` targets where Halmos reports valid models.

## Local Validation Benchmark
Suite: `grandizzy/symbolic-bug-suite` at `fdadd49`
Solver: `z3`
Timing: warm artifact median of 3 runs

| Run | Result | Wall time |
| --- | ---: | ---: |
| This PR | 22/22 replay-confirmed counterexamples | 5.90s |
| Halmos | 22/22 valid models | 31.09s |

On this corpus, this PR finds the same 22 counterexample targets while running in 81.0% less wall time than Halmos, or 5.27x faster.

### DIFF
--- a/crates/evm/symbolic/src/executor/cheatcodes.rs
+++ b/crates/evm/symbolic/src/executor/cheatcodes.rs
@@ -93,6 +93,86 @@ impl SymbolicExecutor {
         CheatcodeOutcome::Continue(Vec::new())
     }
 
+    /// Applies the `deploy_code_cheatcode` symbolic executor helper.
+    pub(super) fn deploy_code_cheatcode<FEN: FoundryEvmNetwork>(
+        &mut self,
+        executor: &Executor<FEN>,
+        state: &mut PathState,
+        completed_paths: &mut usize,
+        artifact: String,
+        constructor_args: Vec<u8>,
+    ) -> Result<CheatcodeOutcome, SymbolicError> {
+        if state.is_static {
+            return Ok(CheatcodeOutcome::Failure);
+        }
+
+        let mut initcode = artifact_code(&artifact, false)?;
+        initcode.extend_from_slice(&constructor_args);
+        let initcode = SymCode::concrete(initcode);
+
+        let nonce = state.world.nonce(executor, state.address)?;
+        let created = state.address.create(nonce);
+        let created_word = SymWord::Concrete(address_word(created));
+
+        let mut failure_world = state.world.clone();
+        failure_world.increment_nonce(executor, state.address)?;
+        if failure_world.has_code_or_nonce(executor, created)? {
+            return Ok(CheatcodeOutcome::Failure);
+        }
+
+        let mut frame = CallFrame::new(
+            created,
+            created,
+            created,
+            state.address,
+            SymWord::zero(),
+            false,
+            SymCalldata::new(Vec::new()),
+        );
+        frame.address_word = created_word.clone();
+        frame.caller_word = state.address_word.clone();
+        let mut child = state.child(frame);
+        let pending_expected_creates = std::mem::take(&mut child.expected_creates);
+        child.world = failure_world.clone();
+        child.world.set_nonce(created, 1);
+        child.expected_revert = None;
+        child.assume_no_revert_next_call = None;
+
+        let mut outcomes =
+            self.execute_external_call(executor, child, &initcode, completed_paths)?;
+        if outcomes.len() != 1 {
+            return Err(SymbolicError::Unsupported("symbolic vm.deployCode branching constructor"));
+        }
+        let outcome = outcomes.remove(0);
+        match outcome.status {
+            TopLevelCallStatus::Success => {
+                state.constraints = outcome.state.constraints;
+                state.next_symbol = outcome.state.next_symbol;
+                state.world = outcome.state.world;
+                state.block = outcome.state.block;
+                state.recorded_logs = outcome.state.recorded_logs;
+                state.access_record = outcome.state.access_record;
+                state.expected_emit = outcome.state.expected_emit;
+                state.expected_calls = outcome.state.expected_calls;
+                state.expected_creates = pending_expected_creates;
+                state.call_mocks = outcome.state.call_mocks;
+                state.function_mocks = outcome.state.function_mocks;
+                self.observe_expected_create(
+                    state,
+                    state.address,
+                    CreateKind::Create,
+                    &outcome.return_data,
+                )?;
+                state.world.install_code(created, outcome.return_data.to_code()?);
+                state.world.set_nonce(created, 1);
+                Ok(CheatcodeOutcome::Continue(vec![created_word]))
+            }
+            TopLevelCallStatus::Revert | TopLevelCallStatus::Failure => {
+                Ok(CheatcodeOutcome::Failure)
+            }
+        }
+    }
+
     /// Applies the `observe_expected_create` symbolic executor helper.
     pub(super) fn observe_expected_create(
         &mut self,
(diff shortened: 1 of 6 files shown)


mablr (commented):
Changes sgtm, maybe would be nice to wait Derek's bench udpate ?

crates/evm/symbolic/src/executor/cheatcodes.rs
```diff
@@ -93,6 +93,86 @@ impl SymbolicExecutor {
...
+                    state,
+                    state.address,
+                    CreateKind::Create,
+                    &outcome.return_data,
+                )?;
+                state.world.install_code(created, outcome.return_data.to_code()?);
+                state.world.set_nonce(created, 1);
+                Ok(CheatcodeOutcome::Continue(vec![created_word]))
+            }
+            TopLevelCallStatus::Revert | TopLevelCallStatus::Failure => {
```
grandizzy: 🚫 **Blocker.** A constructor `Revert` is an ordinary revert, not an assertion failure. This maps to `CheatcodeOutcome::Failure` → `StepOutcome::Failure` → `run.rs` reports a **Counterexample**. Compare `create.rs`, where `TopLevelCallStatus::Revert` pushes 0 and continues; only `Failure` fails. Reachable with concrete initcode (`require(false)`, bad args, etc.). These two statuses must be handled separately — likely needs a new `CheatcodeOutcome::Revert`.

crates/evm/symbolic/src/executor/cheatcodes.rs
```diff
@@ -93,6 +93,86 @@ impl SymbolicExecutor {
...
+    /// Applies the `deploy_code_cheatcode` symbolic executor helper.
+    pub(super) fn deploy_code_cheatcode<FEN: FoundryEvmNetwork>(
+        &mut self,
+        executor: &Executor<FEN>,
+        state: &mut PathState,
+        completed_paths: &mut usize,
+        artifact: String,
+        constructor_args: Vec<u8>,
+    ) -> Result<CheatcodeOutcome, SymbolicError> {
+        if state.is_static {
```
grandizzy, line 154: Static-context create should be a revert, not a counterexample. `create.rs` returns `StepOutcome::Revert` here rather than `Failure`.

crates/evm/symbolic/src/executor/cheatcodes.rs
```diff
@@ -93,6 +93,86 @@ impl SymbolicExecutor {
...
+            SymCalldata::new(Vec::new()),
+        );
+        frame.address_word = created_word.clone();
+        frame.caller_word = state.address_word.clone();
+        let mut child = state.child(frame);
+        let pending_expected_creates = std::mem::take(&mut child.expected_creates);
+        child.world = failure_world.clone();
+        child.world.set_nonce(created, 1);
+        child.expected_revert = None;
+        child.assume_no_revert_next_call = None;
```
grandizzy, line 197: The parent's `expected_revert` / `assume_no_revert_next_call` are cleared but never applied to the constructor outcome. So `vm.expectRevert(); vm.deployCode(reverting)` yields a false counterexample instead of satisfying the expectation. `create.rs` consumes both against the child outcome.

crates/evm/symbolic/src/executor/cheatcodes.rs
```diff
@@ -93,6 +93,86 @@ impl SymbolicExecutor {
...
+            state.address,
+            SymWord::zero(),
+            false,
+            SymCalldata::new(Vec::new()),
+        );
+        frame.address_word = created_word.clone();
+        frame.caller_word = state.address_word.clone();
+        let mut child = state.child(frame);
+        let pending_expected_creates = std::mem::take(&mut child.expected_creates);
+        child.world = failure_world.clone();
```
grandizzy, line 193: Missing `child.world.mark_current_transaction_created(created);` (present in `create.rs`). Without it, SELFDESTRUCT / created-in-tx semantics for the deployed account are modeled incorrectly.

crates/evm/symbolic/src/executor/cheatcodes.rs
```diff
@@ -93,6 +93,86 @@ impl SymbolicExecutor {
...
+        child.world.set_nonce(created, 1);
+        child.expected_revert = None;
+        child.assume_no_revert_next_call = None;
+
+        let mut outcomes =
+            self.execute_external_call(executor, child, &initcode, completed_paths)?;
+        if outcomes.len() != 1 {
```
grandizzy: `len == 0` should be `AssumeRejected` (not `Unsupported`), matching `create.rs`. A branching constructor (`len > 1`) also aborts the whole run rather than spilling extra paths to the worklist as `create.rs` does.

crates/evm/symbolic/src/executor/cheatcodes.rs
```diff
@@ -93,6 +93,86 @@ impl SymbolicExecutor {
...
+                state.expected_creates = pending_expected_creates;
+                state.call_mocks = outcome.state.call_mocks;
+                state.function_mocks = outcome.state.function_mocks;
+                self.observe_expected_create(
+                    state,
+                    state.address,
+                    CreateKind::Create,
+                    &outcome.return_data,
+                )?;
+                state.world.install_code(created, outcome.return_data.to_code()?);
```
grandizzy: Unconditional install. `create.rs` guards with `if !world.destroyed_accounts.contains(&created)` so a constructor that selfdestructs doesn't resurrect runtime code.

crates/evm/symbolic/src/runtime/cheatcodes.rs
```diff
@@ -1225,10 +1225,40 @@ pub(crate) fn artifact_json_path(path: &str) -> PathBuf {
...
+pub(crate) fn artifact_json_fallback_paths(path: &str) -> Vec<PathBuf> {
+    if path.ends_with(".json") {
+        return Vec::new();
+    }
+
+    let mut parts = path.split(':');
+    let first = parts.next().unwrap_or_default();
+    let second = parts.next();
+    let file = Path::new(first);
+    let Some(file_name) = file.file_name() else { return Vec::new() };
```
grandizzy: `Path::file_name` is platform-dependent: on Unix, `src\File.sol:Contract` becomes `out/src\File.sol/...`. Use `first.rsplit(['/', '\\']).next()` for the basename. (Non-blocking.)

grandizzy (changes requested):
**Requesting changes.** The selector/calldata equality folding (part 1) and the `src/File.sol:Contract` artifact fallback (part 3) are sound and well-scoped — nice perf win. The `deployCode` model (part 2) has a blocking correctness bug: a constructor **revert** is reported as a test **failure**, so any fixture whose constructor can revert under reachable input (even concrete) produces a spurious counterexample. The benchmark suite has no reverting constructors, so it wouldn't surface this.

The root cause is that `CheatcodeOutcome` has no `Revert` variant, so constructor reverts get collapsed into `Failure`. I'd suggest reusing `create.rs`'s outcome handling (revert vs. failure, `expectRevert`, `assumeNoRevert`, branching, created-in-tx, selfdestruct) rather than reimplementing a subset. Inline comments mark each spot. Parts 1 and 3 are good to merge after the small Windows-path fix.

grandizzy (approved):
lgtm!

mablr (approved):

Issue #10342, opened 2025-04-21
feat(`debugger`): improve debugger experience when tests fail due to event mismatch


smol-ninja:
### Component

Forge, Foundryup

### Problem

When a test fails due to event mismatch, Foundry logs `FAIL: log != expected log; counterexample calldata=0x1234....1234` without specifying which event parameter did not match.

Though this is fine for simple tests, in case of fuzz testing, it becomes very problematic. This is because Foundry prints lengthy `calldata` forcing to scroll extensively to trace down the error.

![Image](https://github.com/user-attachments/assets/b3f53f09-0f8f-4207-bc0e-993c3943f03a)

👾 [Link to warp block](https://app.warp.dev/block/LO8P0XzdGwbeHxToZdRhp7)

### Solution 1

Log the first mismatched event parameter in the failure message: 

```bash
FAIL: log != expected log event_name:event_parameter expectedValue != actualValue; counterexample calldata=0x1234..1234
```

This would make it easier to identify which event parameter did not match with the expected value without having to scroll down so much.

### Solution 2

Add a Foundry cheat code to enable/disable `calldata` logs. It can be called `display_calldata` 

Most of the times, having `calldata` printed is of little value. This is because, use of `bound` can change the value of the input parameters. Thus, having a cheatcode to enable `calldata` logging can be more useful than having it enabled by default.

grandizzy:
Hey @smol-ninja would approach in https://github.com/foundry-rs/foundry/pull/8686 work for your case? Thank you

smol-ninja:
Hi @grandizzy, I looked at the links that you provided.

> If an emitted event is marked as unmatched, it can then be highlighted as an error in red.

If it only highlights the unmatched event, it doesn't work for my case. I suppose `calldata` will still be emitted, so having this new style would be helpful but doesn't eliminate the need to scroll through all the `calldata` before seeing the `unmatched` event.

grandizzy:
Got it, ok, maybe we can accommodate in same pr

topocount:
@grandizzy I'd originally set out to implement @smol-ninja's first proposed solution but realized there wasn't an easy way to decode the event calldata back to english for the error string without accessing the tracing inspector, which is only available at higher verbosity. Is it permissible to have an opaque error without verbosity, and then provide a more detailed error when verbosity is set?

smol-ninja:
@topocount, would `-vvv` provide detailed error, as I have requested? If yes, then it looks good to me.

[...]

PR #15183, opened 2026-06-15
fix(cheatcodes): decode expectEmit mismatches from project ABIs

figtracer:
## Summary

Refs #10342.

This seeds the cheatcode event signature identifier with the current project's compiled ABIs. `expectEmit` mismatch diagnostics can now decode local event names and parameter values without requiring users to run `forge selectors cache` first.

The new regression uses a unique event signature and asserts that a default `forge test` failure reports the mismatched parameter by name.

## Root Cause

`Cheatcodes::signatures_identifier` built an offline `SignaturesIdentifier` from the global selector cache only. If the project event signature was not already cached, event mismatches fell back to generic `log mismatch at param N` output with raw ABI words.

### DIFF
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -774,7 +774,17 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
 
     /// Returns the signatures identifier.
     pub fn signatures_identifier(&self) -> Option<&SignaturesIdentifier> {
-        self.signatures_identifier.get_or_init(|| SignaturesIdentifier::new(true).ok()).as_ref()
+        self.signatures_identifier
+            .get_or_init(|| {
+                if let Some(artifacts) = &self.config.available_artifacts {
+                    return SignaturesIdentifier::new_offline_with_abis(
+                        artifacts.values().map(|contract| &contract.abi),
+                    )
+                    .ok();
+                }
+                SignaturesIdentifier::new(true).ok()
+            })
+            .as_ref()
     }
 
     /// Decodes the input data and applies the cheatcode.
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -778,7 +778,7 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
             .get_or_init(|| {
                 if let Some(artifacts) = &self.config.available_artifacts {
                     return SignaturesIdentifier::new_offline_with_abis(
-                        artifacts.values().map(|contract| &contract.abi),
+                        artifacts.iter().map(|(_, contract)| &contract.abi),
                     )
                     .ok();
                 }
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -778,7 +778,7 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
             .get_or_init(|| {
                 if let Some(artifacts) = &self.config.available_artifacts {
                     return SignaturesIdentifier::new_offline_with_abis(
-                        artifacts.iter().map(|(_, contract)| &contract.abi),
+                        artifacts.values().map(|contract| &contract.abi),
                     )
                     .ok();
                 }
--- a/crates/evm/traces/src/identifier/signatures.rs
+++ b/crates/evm/traces/src/identifier/signatures.rs
@@ -444,7 +444,7 @@ mod tests {
         let mut first_abi = JsonAbi::default();
         first_abi.events.insert(first.name.clone(), vec![first.clone()]);
         let mut second_abi = JsonAbi::default();
-        second_abi.events.insert(second.name.clone(), vec![second.clone()]);
+        second_abi.events.insert(second.name.clone(), vec![second]);
 
         let mut cache = SignaturesCache::default();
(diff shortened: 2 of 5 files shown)


crates/evm/traces/src/identifier/signatures.rs
```diff
@@ -176,19 +176,40 @@ impl SignaturesIdentifier {
...
+    /// Creates a new `SignaturesIdentifier`, seeding the cache with local ABIs.
+    ///
+    /// - `cache_dir` is the cache directory to store the signatures.
+    /// - `offline` disables the OpenChain client.
+    /// - `abis` contains local ABI signatures that should be preferred without persisting them.
+    pub fn new_with_abis<'a>(
+        cache_dir: Option<&Path>,
+        offline: bool,
+        abis: impl IntoIterator<Item = &'a JsonAbi>,
+    ) -> Result<Self> {
```
mablr: This documents the ABI signatures as non-persisted, but when called with `offline = false`, `Drop` will save the whole cache because `client.is_some()`. That can write project-local ABI signatures into the global selector cache. Maybe we should keep ABI-seeded entries in a non-persisted overlay, or make this constructor private/offline-only?

crates/evm/traces/src/identifier/signatures.rs
```diff
@@ -176,19 +176,40 @@ impl SignaturesIdentifier {
...
     pub fn new_with(cache_dir: Option<&Path>, offline: bool) -> Result<Self> {
+        Self::new_with_abis(cache_dir, offline, [])
+    }
+
+    /// Creates a new `SignaturesIdentifier`, seeding the cache with local ABIs.
+    ///
+    /// - `cache_dir` is the cache directory to store the signatures.
+    /// - `offline` disables the OpenChain client.
+    /// - `abis` contains local ABI signatures that should be preferred without persisting them.
+    pub fn new_with_abis<'a>(
```
grandizzy: **Must fix before merge.** This public `new_with_abis` accepts `offline = false`. In that case `client.is_some()`, so on `Drop` -> `save()` the entire cache — including these project-local ABI signatures — gets written into the global selector cache (`~/.foundry/cache/signatures`). That contradicts the doc's "preferred without persisting them" and pollutes the shared cache with local/private ABI entries.

Fix: make ABI seeding offline-only — drop the `offline` param (always offline) or make `new_with_abis` private and expose only `new_offline_with_abis`. (Matches @mablr's comment.)

crates/evm/traces/src/identifier/signatures.rs
```diff
@@ -176,19 +176,40 @@ impl SignaturesIdentifier {
...
-        let (cache, cache_path) = if let Some(cache_dir) = cache_dir {
+        let (mut cache, cache_path) = if let Some(cache_dir) = cache_dir {
             let path = cache_dir.join("signatures");
             let cache = SignaturesCache::load(&path);
             (cache, Some(path))
         } else {
             Default::default()
         };
+        for abi in abis {
+            cache.extend_from_abi(abi);
```
grandizzy: **topic0 collisions are last-write-wins.** Event `topic0 = keccak("Name(types)")` ignores `indexed`, so two project events that differ only in indexed-ness map to the same `SelectorKind::Event` key. `extend_from_abi` -> `extend` overwrites, so the surviving entry depends on iteration order, and decoded indexed/non-indexed param layout can be wrong even though the name resolves. Add a guardrail (don't overwrite a colliding entry, or prefer the variant matching the actual topic count). Same applies to 4-byte function/error selector collisions.

crates/cheatcodes/src/inspector.rs
```diff
@@ -774,7 +774,17 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
 
     /// Returns the signatures identifier.
     pub fn signatures_identifier(&self) -> Option<&SignaturesIdentifier> {
-        self.signatures_identifier.get_or_init(|| SignaturesIdentifier::new(true).ok()).as_ref()
+        self.signatures_identifier
+            .get_or_init(|| {
+                if let Some(artifacts) = &self.config.available_artifacts {
+                    return SignaturesIdentifier::new_offline_with_abis(
+                        artifacts.values().map(|contract| &contract.abi),
```
grandizzy, line 781: Iteration order of `available_artifacts.values()` decides which ABI wins on a selector/topic0 collision (see comment in signatures.rs). If it isn't deterministic, mismatch diagnostics can vary run-to-run. Consider iterating in a stable order (e.g. sorted by `ArtifactId`) so output is reproducible.

crates/forge/tests/cli/test_cmd/repros.rs
```diff
@@ -255,14 +255,14 @@ contract Issue6170Test is Test {
 Compiler run successful!
 
 Ran 1 test for test/Issue6170.t.sol:Issue6170Test
-[FAIL: log != expected log] test() ([GAS])
+[FAIL: log != expected Values] test() ([GAS])
```
grandizzy: This `Issue6170` repro is the topic0-collision case (both events share `Values(uint256,uint256)`). The new `log != expected Values` output proves only that the name resolves — it isn't coverage for correct *parameter* decoding of collided events. Consider adding a dedicated test asserting decoded param values for same-topic0 events with differing indexedness.

grandizzy (changes requested):

grandizzy (approved):
lgtm

mablr (approved):

PR #7058, opened 2024-02-09
fix(forge): debugger should load sourcemaps by file_id

emo-eth:
## Motivation

Summary from #7043:

> [ContractSources](https://github.com/foundry-rs/foundry/blob/725c9a95b1331c67b55facd4da1598cbc8e03112/crates/common/src/compile.rs#L276) is a mapping of name -> id -> (source code, bytecode)
> 
> In src_map, the source for the contract is [looked up by name](https://github.com/foundry-rs/foundry/blob/3344e2ceeb68fd77d6be001ce7990742debc2e9a/crates/debugger/src/tui/draw.rs#L331). Since this may return multiple contracts with the same name, the entries are iterated over and the SourceElement's index is compared to the file_id of the current contract source. [If they match, the source code is returned](https://github.com/foundry-rs/foundry/blob/3344e2ceeb68fd77d6be001ce7990742debc2e9a/crates/debugger/src/tui/draw.rs#L353).
> 
> The problem is that this will never return source code for inherited contracts, because it is only loading source code for (contracts with the same name as) the contract being invoked; if the SourceElement points to a different index (aka file_id), it will never match any of the contract sources loaded from ContractSources.
> 
> ContractSources should maintain an index both of name -> id -> (source, bytecode) and id -> (source, bytecode) for looking up cross-source sourcemaps.

## Solution

`ContractSources` now also maintains a mapping of `file_id/index -> (source code, bytecode)`. When the loaded `file_id` does not match `source_element.index`, it looks up the contract source using the `source_element`'s `index`.

I could use some help testing this – it looks to me like the [`debug.rs` test](https://github.com/foundry-rs/foundry/blob/bacacceb97da35ccd03b955aab2a6642ccfedd37/crates/forge/tests/cli/debug.rs#L6) doesn't actually make any assertions, so I don't have much to go off of. 

Closes #7043

Before:
Screenshot 2024-02-08 at 6 40 40 PM

After: 
Screenshot 2024-02-08 at 6 41 06 PM

Issue #5435, opened 2023-07-19
forge debug doesn't show source code


AmesCornish:
### Component

Forge

### Have you ensured that all of these are up to date?

- [ ] Foundry
- [X] Foundryup

### What version of Foundry are you on?

nightly 0e72b71

### What command(s) is the bug in?

forge script --debug

### Operating System

Linux

### Describe the bug

Source code debugging seems to have recently broken.  In the current nightly (0e72b71), I can do the following:

> foundryup --commit 0e72b71
> forge init
> forge script --debug script/Counter.s.sol
> 

..and where the source should be, it says "Unknown contract at address 0x5b73...0519"

Earlier versions, for example 288ab2e from about 2 week ago, correctly show and step through the source.

Also, I have not been able to get any recent versions to show source code when stepping into on-chain contracts, even though I've set my ETHERSCAN_API_KEY in env and toml.

jameswenzel:
Seeing this on my end as well.

brucexu-eth:
<img width="975" alt="image" src="https://github.com/foundry-rs/foundry/assets/95468177/9e24a4f9-a349-4065-8cf3-19a715e386e5">
<img width="438" alt="image" src="https://github.com/foundry-rs/foundry/assets/95468177/39430a9a-205e-44f7-b477-02febd514e35">

Same here, with the latest version.

iFrostizz:
Closing @Evalir ?

brucexu-eth:
Verified, working on my side.
<img width="432" alt="image" src="https://github.com/foundry-rs/foundry/assets/95468177/2fc6935c-b05f-42ec-9722-74c55c1c6a7a">

block0xhash:
![image](https://github.com/foundry-rs/foundry/assets/16897613/283fdc01-a7d6-4dbc-b778-bc9bcd97a971)
the docs says it should show source code https://book.getfoundry.sh/forge/debugger

mymiracle0118:
I got same issues on my side.
![image](https://github.com/foundry-rs/foundry/assets/146400751/6b22ddf6-d80a-4df5-a2a8-af7969709215)

emo-eth:
cc @iFrostizz @Evalir, this seemed to be mitigated at the time, but I am frequently (inconsistently) not able to view sourcemap for te actual contract(s) I am testing – the test contract sourcemap works fine, but calls to the actual logic contract show no sourcemap. 
Looks like other users are running into this as well.

[...]

PR #15185, opened 2026-06-15
fix(debugger): identify forked script contracts

figtracer:
## Summary

Refs #5435.

Extends the debugger bytecode mapping added for `forge test --debug` to the `forge script --debug` path. Script execution now preserves `RawCallResult::debug_bytecodes` in `ScriptResult` and uses those bytecodes when building the local trace identifier, so debugger dumps can identify forked/deployed contracts that match local project artifacts.

Also moves the debugger dump contract-identification assertion into a shared CLI test helper and adds a `forge script --debug --dump` regression that deploys a local artifact to Anvil, forks that node, calls the deployed address from a script, and asserts the dump identifies the target contract.

## Context

`#7058` fixed source lookup by file id once a contract was already identified. `#15048` fixed forked/predeployed contract identification for `forge test --debug` by collecting runtime bytecode from the execution database. The remaining `#5435` script repro still lost that bytecode map because `ScriptResult` discarded it before trace decoding.

### DIFF
--- a/crates/forge/tests/cli/script.rs
+++ b/crates/forge/tests/cli/script.rs
@@ -64,7 +64,8 @@ contract ScriptForkDebugTarget {
     let rpc = handle.http_endpoint();
     let pk = "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80";
 
-    cmd.forge_fuse()
+    let create_output = cmd
+        .forge_fuse()
         .args([
             "create",
             "./src/ScriptForkDebugTarget.sol:ScriptForkDebugTarget",
@@ -73,10 +74,14 @@ contract ScriptForkDebugTarget {
             "--private-key",
             pk,
             "--broadcast",
+            "--json",
         ])
-        .assert_success();
-
-    let deployed = address!("f39fd6e51aad88f6f4ce6ab8827279cfffb92266").create(0).to_string();
+        .assert_success()
+        .get_output()
+        .stdout
+        .clone();
+    let create_output: Value = serde_json::from_slice(&create_output).unwrap();
+    let deployed = create_output["deployedTo"].as_str().unwrap().to_owned();
 
     prj.add_script(
         "DebugRemote.s.sol",
@@ -88,8 +93,14 @@ interface IScriptForkDebugTarget {{
 }}
 
 contract DebugRemote {{
+    IScriptForkDebugTarget private target = IScriptForkDebugTarget({deployed});
+
+    function setUp() public {{
+        target.set(7);
+        require(target.value() == 7, "setup value");
+    }}
+
     function run() public {{
-        IScriptForkDebugTarget target = IScriptForkDebugTarget({deployed});
         target.set(19);
         require(target.value() == 19, "value");
     }}
--- a/crates/forge/tests/cli/script.rs
+++ b/crates/forge/tests/cli/script.rs
@@ -1,6 +1,9 @@
 //! Contains various tests related to `forge script`.
 
-use crate::{constants::TEMPLATE_CONTRACT, utils::generate_large_runtime_contract};
+use crate::{
+    constants::TEMPLATE_CONTRACT,
+    utils::{assert_debug_dump_identifies_contract, generate_large_runtime_contract},
+};
 use alloy_hardforks::EthereumHardfork;
 use alloy_network::Ethereum;
 use alloy_primitives::{Address, Bytes, address, hex};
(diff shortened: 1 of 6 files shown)


mablr (dismissed):

crates/script/src/execute.rs
```diff
@@ -359,9 +360,13 @@ impl<FEN: FoundryEvmNetwork> ExecutedState<FEN> {
             )
             .build();
 
-        let mut identifier = TraceIdentifiers::new()
-            .with_local(known_contracts)
-            .with_external(&self.script_config.config, chain_id)?;
+        let mut identifier = if self.execution_result.debug_bytecodes.is_empty() {
```
grandizzy: This switches to bytecode-based local identification for *all* script runs, not just `--debug`, because `debug_bytecodes` is essentially always populated (`collect_debug_bytecodes` runs unconditionally in `RawCallResult`). `forge test` gates the equivalent path on `self.debug`:

```rust
if self.debug && !result.debug_bytecodes.is_empty() { /* with_local_and_bytecodes */ }
```

Can we gate this on the actual debug/dump condition too, so non-debug script trace output and external-identifier behavior stay unchanged? e.g.

```rust
let use_bytecodes = self.args.debug /* real flag */ && !self.execution_result.debug_bytecodes.is_empty();
let mut identifier = if use_bytecodes {
    TraceIdentifiers::new().with_local_and_bytecodes(known_contracts, &self.execution_result.debug_bytecodes)
} else {
    TraceIdentifiers::new().with_local(known_contracts)
}.with_external(&self.script_config.config, chain_id)?;
```

crates/script/src/runner.rs
```diff
@@ -65,18 +65,25 @@ impl<FEN: FoundryEvmNetwork> ScriptRunner<FEN> {
 
         let mut library_transactions = VecDeque::new();
         let mut traces = Traces::default();
+        let mut debug_bytecodes: AddressHashMap<Bytes> = Default::default();
```
grandizzy, line 91: Since these are retained into every `ScriptResult` regardless of debug mode, long fork scripts touching many contracts will hold full runtime bytecode for each address even on normal runs. If the use is gated to debug/dump mode (see execute.rs comment), consider only accumulating/merging `debug_bytecodes` when that mode is active to avoid the extra retained memory on the common path.

crates/forge/tests/cli/script.rs
```diff
@@ -43,6 +46,72 @@ contract ContractScript is Script {
...
+            "./src/ScriptForkDebugTarget.sol:ScriptForkDebugTarget",
+            "--rpc-url",
+            rpc.as_str(),
+            "--private-key",
+            pk,
+            "--broadcast",
+        ])
+        .assert_success();
+
+    let deployed = address!("f39fd6e51aad88f6f4ce6ab8827279cfffb92266").create(0).to_string();
```
grandizzy: This hardcodes the default Anvil account and assumes sender nonce is exactly `0` (no prior library/helper tx). It'll silently break if the deployer or nonce changes. Prefer deriving the address from the deployment — e.g. parse `forge create --json` output or read the receipt's `contractAddress` — rather than recomputing from the default key.

crates/forge/tests/cli/script.rs
```diff
@@ -43,6 +46,72 @@ contract ContractScript is Script {
     }
 );
 
+forgetest_async!(script_debug_dump_identifies_contracts_loaded_from_fork, |prj, cmd| {
```
grandizzy, line 49: Optional: a case where the forked contract is touched during `setUp()` would exercise the new `setup_result.debug_bytecodes.extend(...)` merge in execute.rs, which is otherwise uncovered.

grandizzy (changes requested):
One blocking concern: unlike `forge test --debug`, the script path applies the bytecode-based local identifier whenever `debug_bytecodes` is non-empty, with no `--debug` gate. Since `collect_debug_bytecodes` runs on every `RawCallResult` unconditionally, this means *plain* `forge script` runs now switch identification strategy too — which can change user-visible trace output (contract names, ABI decoding, whether the external identifier is consulted) and retains bytecode blobs for the whole run. `forge test` deliberately gates this on `self.debug`; I think script should match that for both correctness-of-behavior and consistency.

Also one nit on the new test: the deployed address is hardcoded via the default Anvil account + nonce-0 assumption, which is brittle. Prefer deriving it from the deployment output/receipt.

Requesting changes for the gating; the test address is non-blocking but worth fixing. Details inline.

crates/script/src/execute.rs
```diff
@@ -168,6 +168,9 @@ impl<FEN: FoundryEvmNetwork> PreExecutionState<FEN> {
             setup_result.logs.extend(script_result.logs);
             setup_result.traces.extend(script_result.traces);
             setup_result.labeled_addresses.extend(script_result.labeled_addresses);
+            if self.args.debug {
+                setup_result.debug_bytecodes.extend(script_result.debug_bytecodes);
+            }
```
mablr, line 173: This guard is redundant, when `args.debug=false` the runner's `maybe_debug_bytecodes` already returns an empty map, making the extend a no-op?

mablr (commented):
sgtm, just a smol nit

crates/script/src/execute.rs
stevencartavia: i did a pass, +1 on this, nothing else to add

grandizzy (approved):
lgtm

mablr (approved):

PR #15186, opened 2026-06-15
feat(evm): share invariant corpus live across workers

0xKarl98:
This is still under experimenting 

Trying to fix the invariant campaign corpus flow in parallel mode . Instead of only sharing corpus at the start and end of a run, workers now exchange newly interesting tx sequences through a campaign-local in-memory broker while the campaign is running.

The following design would possibly change for shared coverage map : 
Each worker still replays pulled candidates locally and keeps its own coverage maps. The broker only moves immutable tx sequences with light deduping, and final corpus persistence stays merged and deferred on the master worker. Single-worker behavior stays unchanged.


0xKarl98:

PR #15187, opened 2026-06-16
feat(script): EIP-1559 fee presets and fee breakdown

stevencartavia:
`forge script` showed the EIP-1559 `maxFeePerGas` (base × 2 + tip) as "Estimated gas price", confusing users who compared it to the base fee on explorers. Now it prints a base/priority/max breakdown and adds an `--eip1559-fee-estimate <low|market|aggressive>` flag + `eip1559_fee_estimate` config key. Default `market` keeps current behavior.


crates/script/src/simulate.rs
```diff
@@ -369,37 +375,73 @@ impl<FEN: FoundryEvmNetwork> FilledTransactionsState<FEN> {
...
+                    raw.trim_end_matches('0').trim_end_matches('.').to_string()
+                };
+
+                let estimated_gas_price = fmt_gwei(per_gas);
+
+                // Surface the EIP-1559 fee breakdown so users aren't surprised that the
+                // "max fee" exceeds the current base fee. Skipped when `--with-gas-price`
+                // pins the max fee directly; otherwise a `--priority-gas-price` override is
+                // reflected so the display matches what will actually be broadcast.
+                let fee_breakdown = if self.args.with_gas_price.is_none() {
```
mablr: The simulation display uses the cached EIP-1559 estimate, but the browser broadcast path can later raise both `maxPriorityFeePerGas` and `maxFeePerGas` from `eth_maxPriorityFeePerGas`. On OP-stack chains this can make the displayed max fee and amount-required lower than the transaction that is actually sent.

crates/script/src/simulate.rs
```diff
@@ -369,37 +375,73 @@ impl<FEN: FoundryEvmNetwork> FilledTransactionsState<FEN> {
...
+                    provider_info.eip1559_fees().map(|fees| {
+                        let priority_fee = self
+                            .args
+                            .priority_gas_price
+                            .map(|p| p.to())
+                            .unwrap_or(fees.max_priority_fee_per_gas);
+                        (fmt_gwei(fees.base_fee_per_gas), fmt_gwei(priority_fee))
+                    })
+                } else {
```
mablr: With `--priority-gas-price`, the printed priority fee is the override but the printed max fee remains the original cap, so the breakdown reads like reconciled fee math even though it is really two caps.

crates/script/src/lib.rs
```diff
@@ -130,6 +130,16 @@ pub struct ScriptArgs {
...
 
+    /// How to estimate EIP1559 fees: `low`, `market` (default), or `aggressive`.
+    ///
+    /// `market` is the default (`base_fee * 2` plus a 20th-percentile priority
+    /// fee). `low` bids a smaller tip and max-fee cap (may stall if the base fee
+    /// rises); `aggressive` bids a larger tip for faster inclusion. Note that the
+    /// base fee paid is set by the network, not this flag. Ignored for `--legacy`
+    /// and overridden by `--with-gas-price` / `--priority-gas-price`.
+    #[arg(long, value_name = "PRESET")]
+    pub eip1559_fee_estimate: Option<Eip1559FeeEstimatePreset>,
```
mablr: The help says the preset is “overridden by `--with-gas-price` / `--priority-gas-price`”, but each flag overrides only one EIP-1559 field?

Could you clarify why `low` preset is especially sensitive.

stevencartavia: [0e070d9](https://github.com/foundry-rs/foundry/pull/15187/commits/0e070d96854b0401ee001fc1a41644ce2eb368e2)

crates/script/src/simulate.rs
stevencartavia: [0e070d9](https://github.com/foundry-rs/foundry/pull/15187/commits/0e070d96854b0401ee001fc1a41644ce2eb368e2)

crates/script/src/simulate.rs
stevencartavia: [0e070d9](https://github.com/foundry-rs/foundry/pull/15187/commits/0e070d96854b0401ee001fc1a41644ce2eb368e2)

mablr (dismissed):
Seems good

figtracer:
lgtm, small readability nit q: can we shorten the flag's name? it's too verbose, maybe something like `--estimate` is fine if we do t have it?

mablr (approved):
sgtm

figtracer (approved):
LGTM from my review pass. No actionable comments.

PR #15189, opened 2026-06-16
feat(fuzz): impl constant folding for typical expressions

### LINKED ISSUE #12044: feat(fuzz): impl constant folding for typical expressions

stevencartavia:
Closes #12044. 

Enhances `LiteralsCollector` to fold typical constant expressions (arithmetic, casts, `keccak256`, bitwise/shift) and seed the results into the fuzz dictionary — e.g. `2 * 2 ether`, `uint(-2)`, and `bytes32(uint256(keccak256('eip1967.proxy.implementation')) - 1)`.

### DIFF
--- a/crates/evm/fuzz/src/strategies/literals.rs
+++ b/crates/evm/fuzz/src/strategies/literals.rs
@@ -418,7 +418,8 @@ impl Num {
     }
 
     /// Negates the value, keeping its width and switching to a signed context. Returns `None` if
-    /// the result would not fit in [`I256`].
+    /// the result doesn't fit [`I256`] or, for a width-carrying value, its `intN` type (negating
+    /// `type(intN).min` reverts under checked arithmetic).
     fn neg(self) -> Option<Self> {
         match self {
             // `-x` fits in `I256` iff `x <= 2**255` (`|I256::MIN|`).
@@ -426,7 +427,13 @@ impl Num {
                 (raw <= I256::MIN.into_raw()).then(|| Self::int(raw.wrapping_neg(), true, width))
             }
             Self::Int { signed: true, width, .. } => {
-                self.to_i256()?.checked_neg().map(|i| Self::int(i.into_raw(), true, width))
+                let r = self.to_i256()?.checked_neg()?;
+                if let Some(bits) = width
+                    && !can_fit_int(r, bits)
+                {
+                    return None;
+                }
+                Some(Self::int(r.into_raw(), true, width))
             }
             Self::Bytes { .. } => None,
         }
@@ -468,35 +475,71 @@ fn apply_bin(op: ast::BinOpKind, a: Num, b: Num) -> Option<Num> {
         if op == Pow {
             return signed_pow(x, y, width);
         }
-        let r = match op {
-            Add => x.wrapping_add(y),
-            Sub => x.wrapping_sub(y),
-            Mul => x.wrapping_mul(y),
-            Div => x.checked_div(y)?,
-            Rem => x.checked_rem(y)?,
-            _ => return None,
+        // Mirror the unsigned path: typed signed operands use checked arithmetic (an overflow of
+        // their `intN` type reverts at runtime), untyped literals keep 256-bit wrapping.
+        let r = match width {
+            Some(bits) => {
+                let r = match op {
+                    Add => x.checked_add(y)?,
+                    Sub => x.checked_sub(y)?,
+                    Mul => x.checked_mul(y)?,
+                    Div => x.checked_div(y)?,
+                    Rem => x.checked_rem(y)?,
+                    _ => return None,
+                };
+                if !can_fit_int(r, bits) {
+                    return None;
+                }
+                r
+            }
+            None => match op {
+                Add => x.wrapping_add(y),
+                Sub => x.wrapping_sub(y),
+                Mul => x.wrapping_mul(y),
+                Div => x.checked_div(y)?,
+                Rem => x.checked_rem(y)?,
+                _ => return None,
+            },
         };
         return Some(Num::int(r.into_raw(), true, width));
     }
 
     let (x, y) = (a.as_u256(), b.as_u256());
     let r = match op {
-        // Wrapping (mod 2**256) matches EVM unchecked arithmetic, so the produced values are ones
-        // that can actually occur at runtime.
-        Add => x.wrapping_add(y),
-        Sub => x.wrapping_sub(y),
-        Mul => x.wrapping_mul(y),
+        // Default-checked arithmetic: a width-carrying op overflowing its type reverts at runtime,
+        // so bail (`None`) instead of seeding the wrapped value. Untyped literals keep 256-bit
+        // wrapping.
+        Add => checked_arith(x.checked_add(y), x.wrapping_add(y), width)?,
+        Sub => checked_arith(x.checked_sub(y), x.wrapping_sub(y), width)?,
+        Mul => checked_arith(x.checked_mul(y), x.wrapping_mul(y), width)?,
         Div => x.checked_div(y)?,
         Rem => x.checked_rem(y)?,
         // `pow` overflowing `U256` means an out-of-range constant, so bail rather than seed a
-        // wrapped value. (`2 ** 256 - 1` therefore doesn't fold; use `type(uint256).max`.)
-        Pow => x.checked_pow(y)?,
+        // wrapped value. (`2 ** 256 - 1` therefore doesn't fold; use `type(uint256).max`.) A
+        // width-carrying result must also fit its type, else checked arithmetic would revert.
+        Pow => {
+            let r = x.checked_pow(y)?;
+            checked_arith(Some(r), r, width)?
+        }
         // Unsigned `>>` is a logical shift.
         Shr => shift_amount(y).map_or(U256::ZERO, |s| x.wrapping_shr(s)),
         // Comparisons and logical operators are not folded into values.
         _ => return None,
     };
     Some(Num::int(r, false, width))
+}
+
+/// Resolves a default-checked unsigned arithmetic result.
+///
+/// For a width-carrying operand the result type is checked: it must neither overflow 256 bits
+/// (`checked` is `None`) nor exceed its `uintN` width, otherwise the expression reverts at runtime
+/// and we return `None` instead of seeding an unreachable value. Untyped literal operands keep the
+/// 256-bit wrapping value.
+fn checked_arith(checked: Option<U256>, wrapping: U256, width: Option<usize>) -> Option<U256> {
+    match width {
+        Some(bits) => checked.filter(|r| can_fit_uint(*r, bits)),
+        None => Some(wrapping),
+    }
 }
 
 /// Folds a signed `base ** exp`, returning `None` on a negative exponent or a magnitude that
@@ -507,13 +550,21 @@ fn signed_pow(base: I256, exp: I256, width: Option<usize>) -> Option<Num> {
     }
     let magnitude = base.unsigned_abs().checked_pow(exp.into_raw())?;
     let negative = base.is_negative() && exp.into_raw().bit(0);
-    if negative {
+    let value = if negative {
         // Negative results must fit `[-2**255, 0)`, i.e. magnitude <= |I256::MIN|.
-        (magnitude <= I256::MIN.into_raw()).then(|| Num::int(magnitude.wrapping_neg(), true, width))
+        (magnitude <= I256::MIN.into_raw()).then(|| I256::from_raw(magnitude.wrapping_neg()))?
     } else {
         // Positive results must fit `[0, 2**255)`.
-        (magnitude < I256::MIN.into_raw()).then(|| Num::int(magnitude, true, width))
+        (magnitude < I256::MIN.into_raw()).then(|| I256::from_raw(magnitude))?
+    };
+    // A width-carrying result that overflows its `intN` type reverts under checked arithmetic, so
+    // don't seed a value the expression can never produce.
+    if let Some(bits) = width
+        && !can_fit_int(value, bits)
+    {
+        return None;
     }
+    Some(Num::int(value.into_raw(), true, width))
 }
 
 /// Combines the widths of two operands: an untyped operand (`None`) inherits the other's width;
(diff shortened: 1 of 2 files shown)


crates/evm/fuzz/src/strategies/literals.rs
```diff
@@ -115,115 +119,365 @@ impl LiteralsCollector {
...
+        let word = B256::from(value);
+        for bits in [8, 16, 32, 64, 128, 256] {
+            if can_fit_uint(value, bits) {
+                self.insert_word(DynSolType::Uint(bits), word);
+            }
+        }
+    }
+
+    /// Seeds a signed value under all `intN` sizes that can represent it.
+    fn seed_int(&mut self, value: I256) {
```
grandizzy, line 163: **`int8` is missing from the size list.** `seed_uint` iterates `[8, 16, 32, 64, 128, 256]` but `seed_int` uses `[16, 32, 64, 128, 256]`. As a result a folded negative like `int8 constant X = -1;` never seeds `DynSolType::Int(8)` (only the explicit-cast path in `seed_cast` does). Add `8` for parity with `seed_uint`.

crates/evm/fuzz/src/strategies/literals.rs
```diff
@@ -115,115 +119,365 @@ impl LiteralsCollector {
...
+            ast::ExprKind::Lit(lit, _) => match &lit.kind {
+                // Sub-denominations (e.g. `ether`, `days`) are already folded into the value.
+                ast::LitKind::Number(n) => Some(Num::U(U256::from(*n))),
+                _ => None,
+            },
+            ast::ExprKind::Unary(op, inner) => {
+                let value = self.eval(inner)?;
+                match op.kind {
+                    ast::UnOpKind::Neg => value.neg(),
+                    ast::UnOpKind::BitNot => {
```
grandizzy: **`~` uses 256-bit semantics regardless of operand width.** Because `Num` doesn't carry the cast width, `~uint8(0)` folds to `!U256::ZERO = 2**256-1` and `seed_uint` only seeds `uint256` — it misses the correct `uint8(255)` and spends a `max_values` slot on a value that can't occur. Same for `~uint16(0)`, etc. Either track width in `Num`, or bail (return `None`) when the operand came from a narrow (<256) cast.

crates/evm/fuzz/src/strategies/literals.rs
```diff
@@ -115,115 +119,365 @@ impl LiteralsCollector {
...
+
+    // Bitwise AND/OR/XOR and left-shift operate purely on the raw bits and are independent of
+    // signedness, so they fold in any context (the signed context is just carried through).
+    if matches!(op, BitAnd | BitOr | BitXor | Shl) {
+        let (x, y) = (a.as_u256(), b.as_u256());
+        let raw = match op {
+            BitAnd => x & y,
+            BitOr => x | y,
+            BitXor => x ^ y,
+            Shl => shift_amount(y).map_or(U256::ZERO, |s| x.wrapping_shl(s)),
```
grandizzy, line 465: **Width lost after casts makes narrow shift/arith results wrong.** `uint8(1) << 8` should truncate to `0` but this folds to `256`; `int8(1) << 7` should be `-128` but folds to `+128`. The result type of a shift is the left operand's type, truncated to it. Without width tracking these seed values that can never appear at runtime. Recommend tracking width in `Num`, or conservatively returning `None` once any operand originates from a narrow cast.

crates/evm/fuzz/src/strategies/literals.rs
```diff
@@ -115,115 +119,365 @@ impl LiteralsCollector {
...
+
+/// Reinterprets a folded value for an elementary type cast used inside a larger expression,
+/// applying the target width so truncation and sign-extension are correct (e.g.
+/// `int256(uint256(type(uint256).max))` -> `-1`).
+fn cast_num(ty: ast::ElementaryType, value: Num) -> Option<Num> {
+    match ty {
+        ast::ElementaryType::UInt(size) => Some(Num::U(cast_uint(value, size.bits() as usize))),
+        ast::ElementaryType::Int(size) => Some(Num::I(cast_int(value, size.bits() as usize))),
+        ast::ElementaryType::Address(_) => Some(Num::U(low_bits(value.as_u256(), 160))),
+        ast::ElementaryType::FixedBytes(_) => Some(Num::U(value.as_u256())),
```
grandizzy: **Nested fixed-bytes truncation keeps the wrong bytes.** `Some(Num::U(value.as_u256()))` discards the `bytesN` length, so `bytes2(bytes4(uint32(0x12345678)))` folds to `0x5678` instead of Solidity's left-truncated `0x1234`. The single-level `seed_cast(FixedBytes(..))` path is fine; only nested fixed-bytes is wrong. Carry the byte length through so truncation keeps the leftmost bytes.

crates/evm/fuzz/src/strategies/literals.rs
```diff
@@ -115,115 +119,365 @@ impl LiteralsCollector {
...
-                                .output
-                                .words
-                                .entry(DynSolType::Uint(bits))
-                                .or_default()
-                                .insert(pos_b256)
+    /// Attempts to constant-fold a compound expression and seed the resulting value(s).
+    ///
+    /// This walks recognized casts (`uintN`/`intN`/`bytesN`/`address`), `keccak256` of literal
+    /// arguments, and arithmetic/bitwise expressions over numeric literals.
+    fn fold_and_seed(&mut self, expr: &ast::Expr<'_>) {
```
grandizzy, line 201: **`type(uintN).max` / `type(intN).min|max` are not folded.** These are member-access expressions, but `fold_and_seed` (and `eval`) only handle `Lit`/`Unary`/`Binary`/`Call`, so the single most common Solidity constant expression (and `type(uint256).max - 1`, etc.) is missed. Given the issue scope is "typical expressions," this should be handled before closing #12044.

crates/evm/fuzz/src/strategies/literals.rs
```diff
@@ -115,115 +119,365 @@ impl LiteralsCollector {
...
+        // Wrapping (mod 2**256) matches EVM unchecked arithmetic, so the produced values are ones
+        // that can actually occur at runtime.
+        Add => x.wrapping_add(y),
+        Sub => x.wrapping_sub(y),
+        Mul => x.wrapping_mul(y),
+        Div => x.checked_div(y)?,
+        Rem => x.checked_rem(y)?,
+        // `pow` overflowing almost always means an out-of-range (compile-error) constant, so bail
+        // rather than seed a wrapped/garbage value.
+        Pow => x.checked_pow(y)?,
```
grandizzy: **`2 ** 256 - 1` doesn't fold.** `checked_pow` bails because `2**256` overflows `U256`, but Solidity keeps literal sub-expressions arbitrary-precision, so this very common max-value idiom is lost. Consider a special-case for the `2**N` pattern (or note it as a known limitation).

crates/evm/fuzz/src/strategies/literals.rs
```diff
@@ -115,115 +119,365 @@ impl LiteralsCollector {
...
+        }
+    }
+
+    /// Returns `true` if the value is in a signed context.
+    const fn is_signed(self) -> bool {
+        matches!(self, Self::I(_))
+    }
+
+    /// Negates the value, if it fits in [`I256`]. The result stays signed.
+    fn neg(self) -> Option<Self> {
```
grandizzy, line 423: **`-2 ** 255` (int256 min) doesn't fold.** `to_i256()` rejects the positive `2**255` intermediate before negation. Minor, but it's the canonical `int256` minimum.

crates/evm/fuzz/src/strategies/literals.rs
```diff
@@ -115,115 +119,365 @@ impl LiteralsCollector {
...
+}
+
+/// Returns the single positional argument of a call, if it has exactly one.
+fn single_arg<'a, 'ast>(args: &'a ast::CallArgs<'ast>) -> Option<&'a ast::Expr<'ast>> {
+    let mut exprs = args.exprs();
+    (exprs.len() == 1).then(|| exprs.next()).flatten()
+}
+
+/// Extracts the raw bytes of a string, unicode, or hex string literal argument.
+fn lit_bytes(expr: &ast::Expr<'_>) -> Option<Vec<u8>> {
```
grandizzy: **Allocates a `Vec<u8>` per call and is re-invoked by every ancestor.** Return `Option<&[u8]>` and pass it straight to `keccak256(..)` to avoid the allocation. Combined with the redundant-eval issue, hash subtrees get re-hashed multiple times.

crates/evm/fuzz/src/strategies/literals.rs
```diff
@@ -115,115 +119,365 @@ impl LiteralsCollector {
...
+                if let Some(value) = self.eval(expr) {
+                    self.seed_num(value);
+                }
+            }
+            _ => {}
+        }
+    }
+
+    /// Recursively evaluates a constant expression to a numeric value, if possible.
+    fn eval(&self, expr: &ast::Expr<'_>) -> Option<Num> {
```
grandizzy, line 241: **Redundant subtree re-evaluation.** `visit_expr` calls `fold_and_seed`→`eval` over the whole subtree, then `walk_expr` descends and re-`eval`s each child's subtree — `O(n^2)` on degenerate operator chains, with keccak re-hashing. Runs once off the hot path so not urgent, but memoizing `eval` by expr span would make it ~`O(n)` and bound the repeated hashing. Deep nesting is also unbounded recursion (stack-overflow risk).

grandizzy (dismissed):
## Summary

My main concerns are that a few narrow-width cases fold to **incorrect** values, and the most common "typical expression" — `type(uint256).max` — isn't covered yet, which matters since this closes #12044.

## Must fix before merge

1. **`seed_int` omits `int8`** (`seed_int`). Inconsistent with `seed_uint`; folded negatives never seed `Int(8)`. One-line fix.
2. **Width is lost after casts → wrong values seeded.** `Num` only tracks signed/unsigned, not bit-width, so 256-bit semantics leak into narrow types:
   - `~uint8(0)` → seeds `2**256-1` under `uint256` instead of `uint8(255)`
   - `uint8(1) << 8` → `256` instead of `0`
   - `int8(1) << 7` → `+128` instead of `-128`

   These aren't harmless noise — they seed unreachable values and consume `max_values` budget. Either carry width in `Num`, or conservatively bail (`None`) once an operand comes from a narrow (<256) cast. The latter keeps all the good 256-bit cases with minimal change.
3. **`type(uintN).max` / `type(intN).min|max` not folded.** Member-access exprs aren't handled by `eval`/`fold_and_seed`. This is the most common constant expression in real Solidity and is squarely in the scope of #12044 ("typical expressions").

## Should fix / consider

4. **Nested fixed-bytes truncation** (`cast_num` `FixedBytes(_)`) keeps low bytes instead of left bytes: `bytes2(bytes4(...))` is wrong. Single-level casts are fine.
5. **Common max idioms** `2 ** 256 - 1` and `-2 ** 255` bail out due to `U256`/`I256` intermediate overflow (Solidity keeps literal sub-exprs arbitrary-precision).

## Performance (non-blocking)

6. `visit_expr`→`fold_and_seed`→`eval` re-evaluates each subtree, then `walk_expr` re-evaluates children — `O(n^2)` on operator chains, with keccak re-hashing and a `Vec` alloc per `lit_bytes` call. Runs once off the hot path so not urgent; memoizing `eval` by span and returning `&[u8]` from `lit_bytes` would clean it up. Deep nesting is also unbounded recursion.

## Nice to have

A short doc comment clarifying this is an **opportunistic 256-bit-oriented heuristic**, not a Solidity-correct constant folder, would set expectations for the width-dependent cases that are intentionally approximate.

figtracer:
derek bench invariant

crates/evm/fuzz/src/strategies/literals.rs
stevencartavia: added parity

crates/evm/fuzz/src/strategies/literals.rs
stevencartavia: [13a6168](https://github.com/foundry-rs/foundry/pull/15189/commits/13a6168bdfd880ed6c508ce630b5386c1146d44c)

crates/evm/fuzz/src/strategies/literals.rs
stevencartavia: [13a6168](https://github.com/foundry-rs/foundry/pull/15189/commits/13a6168bdfd880ed6c508ce630b5386c1146d44c)

crates/evm/fuzz/src/strategies/literals.rs
stevencartavia: [13a6168](https://github.com/foundry-rs/foundry/pull/15189/commits/13a6168bdfd880ed6c508ce630b5386c1146d44c)

crates/evm/fuzz/src/strategies/literals.rs
stevencartavia: [13a6168](https://github.com/foundry-rs/foundry/pull/15189/commits/13a6168bdfd880ed6c508ce630b5386c1146d44c)

stevencartavia:
derek bench invariant

crates/evm/fuzz/src/strategies/literals.rs
```diff
@@ -115,115 +125,496 @@ impl LiteralsCollector {
...
+            _ => return None,
+        };
+        return Some(Num::int(r.into_raw(), true, width));
+    }
+
+    let (x, y) = (a.as_u256(), b.as_u256());
+    let r = match op {
+        // Wrapping (mod 2**256) matches EVM unchecked arithmetic, so the produced values are ones
+        // that can actually occur at runtime.
+        Add => x.wrapping_add(y),
```
figtracer: this folds checked typed arithmetic as wrapping, so it can seed values the solidity expression cannot produce. repro: `uint8 public constant f = uint8(250) + 10;` compiles with solc 0.8.30, but calling `f()` reverts with panic 0x11 instead of returning 4. consider avoiding wrapping for width-carrying operands unless the expression is inside `unchecked`.

stevencartavia: [004bcf9](https://github.com/foundry-rs/foundry/pull/15189/commits/004bcf9a927d572cc46fb5dd0d6b8984018ed71e)

mablr:
derek bench invariant

figtracer (approved):
LGTM from my review pass. No actionable comments.

0xalpharush (approved):

mablr (approved):
I haven't found any blockers.

PR #15190, opened 2026-06-16
bench: PR 15047 invariant workers

0xKarl98:
Do not merge 

Temporary internal PR for benchmarking #15047 against the exact pre-merge baseline.


0xKarl98:
derek bench invariant compare-ref=bench-base-pr-15047

PR #15193, opened 2026-06-16
feat: add T6 admin access key authorization tooling

stevencartavia:
Adds support for T6 admin access keys: preserves/exposes the new `is_admin` and `account` `KeyAuthorization` fields, adds `cast key-authorization encode/sign/inspect` (`--admin`, `--account`, `--witness`) plus `keychain authorize --admin`, `is-admin`, and `verify-admin` with `--json` output, and rejects admin authorizations carrying expiry/limits/scopes or a mismatched account. Non-admin authorizations remain backward compatible.

Closes OSS-341

### DIFF
--- a/crates/common/src/tempo/auth.rs
+++ b/crates/common/src/tempo/auth.rs
@@ -366,6 +366,15 @@ mod tests {
         Address::from_public_key(&vk)
     }
 
+    /// Shape of the `keyAuthorization` the mock `/poll` returns.
+    #[derive(Clone, Copy, Default)]
+    struct MockAuthShape {
+        /// Return a T6 admin key authorization.
+        admin: bool,
+        /// Bind the authorization to this account.
+        account: Option<Address>,
+    }
+
     #[derive(Clone)]
     struct MockState {
         wallet: Arc<Mutex<Option<Address>>>,
@@ -374,6 +383,8 @@ mod tests {
         key_id: Arc<Mutex<Option<Address>>>,
         /// Chain ID the mock `/poll` returns in `keyAuthorization`.
         poll_chain_id: u64,
+        /// Shape of the returned authorization, for exercising rejection paths.
+        shape: MockAuthShape,
     }
 
     async fn create_code_handler(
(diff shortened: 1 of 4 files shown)


crates/common/src/tempo/auth.rs
```diff
@@ -198,6 +198,20 @@ pub async fn ensure_access_key(cfg: EnsureAccessKeyConfig) -> Result<AccessKeyOu
                         signed.authorization.key_type,
                     );
                 }
+                // A 402 access key is a limited key; an admin key is never valid here.
+                if signed.authorization.is_admin() {
+                    eyre::bail!(
+                        "wallet returned an admin key authorization, expected a limited access key"
+                    );
+                }
```
mablr, line 206: This new rejection path doesn't seem to be covered by the mock wallet tests. If a wallet returns an admin authorization, or an account-bound authorization for a different account, we must reject before writing keys.toml?

mattsse (changes requested):
I found a few gaps in the delegated-admin flows. The CLI can create and inspect root-bound artifacts, but the paths that should let an active admin access key manage other keys are still blocked or produce the wrong account binding.

- `crates/cast/src/cmd/keychain.rs:3498`: `resolve_keychain_root_signer` still rejects every Tempo access-key signer before any T6 admin check. That blocks the core admin-key workflow introduced here: after an admin key is authorized, `cast keychain authorize`, `revoke`, and the policy mutators still cannot be submitted by that admin key, even though T6 explicitly allows active admin access keys to call the admin-gated keychain mutators.

  Suggested change: add an access-key signer path for AccountKeychain mutations. It should resolve the root account from the `TempoAccessKeyConfig`, verify the selected key is an active admin key for that account on T6, then build/send the transaction with the access-key flow. Please add a CLI/e2e test where an authorized admin key authorizes or revokes another key.

- `crates/cast/src/cmd/keychain.rs:2930` and `crates/cast/src/cmd/keychain.rs:2944`: `key-authorization sign --admin` cannot create delegated admin-signed key authorizations. It rejects Tempo access-key signers, and if the admin private key is passed directly it binds `account` to the signer address. For an admin key authorizing a child key, the signer is the admin key but the bound account must be the root account being modified. The generated artifact is therefore unusable for delegated admin signing.

  Suggested change: let `key-authorization sign --admin` accept or derive the target account. For access-key signing, derive it from `TempoAccessKeyConfig.wallet_address`; for direct signing, add an explicit `--account` option and only default to `signer_address` for root-signed authorizations. Add a T6 test that signs a child authorization with an admin key and verifies the bound account is the root.

- `crates/cast/src/cmd/keychain.rs:3023`: `key-authorization inspect` rejects valid root-signed admin authorizations that omit `account`. The T6 rules allow root-signed admin authorizations to omit the trailing account field; `account` is only mandatory when the authorization signer is not the target account's root key. Since `inspect` is advertised as a decoder for signed or unsigned artifacts, it should not reject valid artifacts produced by other tooling or lower-level APIs.

  Suggested change: allow `auth.is_admin()` with no `account` unless `--account` was explicitly supplied and requires a binding. Keep rejecting account mismatches and admin authorizations that carry expiry, limits, or call scopes.

crates/common/src/tempo/auth.rs
stevencartavia: new tests in `crates/common/src/tempo/auth.rs`

mattsse (approved):
LGTM

PR #15195, opened 2026-06-16
feat(cast): add Tempo receive policy cast tooling

figtracer:
## Summary

Adds a Foundry-native `cast receive-policy` command surface for Tempo T6 receive policy workflows:

- `set`, `get`, and `validate` for TIP-403 receive policies
- `receipt decode`, `receipt balance`, and `receipt burn` utilities for ReceivePolicyGuard receipts
- `claim <to> <receipt>` that passes the requested target through to the guard and leaves resume vs reroute to onchain policy
- JSON-capable output for read/receipt flows and preview output for `set --preview`
- a preflight warning when originator recovery is used while the configured sender policy blocks Tempo system/precompile sender addresses

The core ReceivePolicyGuard precompile, T6 hardfork gating, fork sentinel warming, Anvil precompile reporting, trace labels, and ABI decoding were already present on `origin/master` through the existing Tempo T6 dependency work. This PR focuses on the missing Cast operator/support workflow.

### DIFF
--- a/crates/cast/src/cmd/receive_policy.rs
+++ b/crates/cast/src/cmd/receive_policy.rs
@@ -18,7 +18,8 @@ use serde_json::{Value, json};
 use std::str::FromStr;
 use tempo_alloy::TempoNetwork;
 use tempo_contracts::precompiles::{
-    IReceivePolicyGuard, ITIP403Registry, RECEIVE_POLICY_GUARD_ADDRESS, TIP403_REGISTRY_ADDRESS,
+    ADDRESS_REGISTRY_ADDRESS, IAddressRegistry, IReceivePolicyGuard, ITIP403Registry,
+    RECEIVE_POLICY_GUARD_ADDRESS, TIP403_REGISTRY_ADDRESS,
 };
 use tempo_primitives::TempoAddressExt;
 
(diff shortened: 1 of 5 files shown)


stevencartavia (dismissed):

crates/cast/src/cmd/receive_policy.rs
```diff
@@ -0,0 +1,664 @@
...
+fn decode_receipt(receipt: Bytes) -> Result<()> {
+    let decoded = decode_claim_receipt(&receipt)?;
+    let payload = receipt_payload(&receipt, &decoded, None);
+    print_payload(payload, |payload| {
+        print_decoded_receipt(payload)?;
+        sh_println!(
+            "\nClaim path: cast receive-policy claim {} {}",
+            payload["recipient"].as_str().unwrap_or_default(),
+            payload["receipt"].as_str().unwrap_or_default(),
+        )
```
mablr: This suggests claiming to `recipient`, but receipt `recipient` can be a virtual address while the guard’s resume path compares to against the resolved receiver/master. For receiver recovery, following this hint can turn a resume into a reroute and trigger extra policy/spending-limit checks.

crates/cast/src/cmd/receive_policy.rs
```diff
@@ -0,0 +1,697 @@
...
+
+fn print_claim_hint(payload: &Value) -> Result<()> {
+    let recipient = payload["recipient"].as_str().unwrap_or_default();
+    let receipt = payload["receipt"].as_str().unwrap_or_default();
+    if payload["recipient_is_virtual"].as_bool().unwrap_or_default() {
+        sh_println!(
+            "\nClaim target: recipient is a virtual address; resolve it first with:\n  cast vaddr resolve {recipient}\nThen claim to the registered master address:\n  cast receive-policy claim <master-address> {receipt}"
+        )
+    } else {
+        sh_println!("\nClaim path: cast receive-policy claim {recipient} {receipt}")
```
mablr, line 562: For originator-recovery receipts (`recovery_authority == 0x0`, the default), `claim(recipient, receipt)` is always a reroute, so sending it back to the same recipient that blocked the inbound transfer will commonly re-run receive-policy checks and fail.

mablr (dismissed):
sgtm

mattsse (changes requested):
Thanks, this fills in the operator workflow well. I found one behavioral mismatch in the receipt hinting that should be fixed before merging.

- `crates/cast/src/cmd/receive_policy.rs:524`: `print_claim_hint` checks `recipient_is_virtual` before checking `recovery_mode == "originator"`. For receipts with both a virtual recipient and originator recovery, the human output tells users to resolve the virtual address and claim to the registered master. That contradicts the originator-recovery flow: originator claims are reroutes and should not default back to the blocked recipient/master, since that can re-run the policy that held the transfer or fail claim-time policy checks.

  Suggested change: handle originator recovery before the virtual-recipient branch, or add a combined virtual+originator message that requires an explicit claim target and does not suggest the master as the default. A small test for `recipient_is_virtual == true` plus `recovery_mode == "originator"` would lock this down.

mablr (dismissed):

0xKarl98:
This still looks mismatched with the real TIP-20 transfer path. cast receive-policy validate resolves the CLI args, but then passes the receiver straight into validateReceivePolicy:
  ```rust
  let receiver = receiver.resolve(&provider).await?;
  let result = registry.validateReceivePolicy(token, sender, receiver).call().await?;
  ```
Real TIP-20 transfers resolve the recipient first:
   ```rust
 let to = Recipient::resolve(to)?;
 self.ensure_transfer_authorized(from, to.target)?;
 ```
 And receive-policy checks use the resolved target:
  ```rust
  check_receive_policy(token, originator, to.target)?
  ```

So if receiver is a TIP-1022 virtual address, the CLI can validate policy against the virtual alias, while an actual transfer validates against the registered master. That can make the CLI report credited even though the real transfer would be held.

0xKarl98 (approved):

mablr (approved):

mattsse (approved):

PR #15199, opened 2026-06-16
fix(anvil): support onlyTopLevelCall tracer config

### LINKED ISSUE #9569: Add `"onlyTopLevelCall"` to `debug_traceTransaction` request data
### Component

Anvil

To get return values from Alchemy, I use this tracer config in debug_traceTransaction:

```
{"tracer": "callTracer", "tracerConfig": {"onlyTopLevelCall": true}}
``

I would like to be able to do the same when chatting up a Foundry node (via RPC).

Unless there is a different way I can the return value for a transaction via its hash?

### Additional context

_No response_

figtracer:
## Summary
- Support Alchemy-style `onlyTopLevelCall` as an alias for callTracer's existing `onlyTopCall` option.
- Add a JSON-RPC `debug_traceTransaction` regression test that verifies the alias suppresses nested calls.

Closes #9569.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -75,9 +75,9 @@ use alloy_rpc_types::{
     trace::{
         filter::TraceFilter,
         geth::{
-            FourByteFrame, GethDebugBuiltInTracerType, GethDebugTracerType,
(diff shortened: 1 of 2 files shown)


stevencartavia (approved):

mablr (approved):
sgtm

PR #15201, opened 2026-06-16
feat(lint): add `incorrect_modifier`

mablr:
## Motivation

Close OSS-381 by replicating Aderyn's `incorrect_modifier` lint.

See also https://github.com/foundry-rs/foundry/issues/14381#issuecomment-4707154414

### DIFF
--- a/crates/lint/src/sol/low/incorrect_modifier.rs
+++ b/crates/lint/src/sol/low/incorrect_modifier.rs
@@ -163,12 +163,17 @@ fn stmt_outcome(stmt: &Stmt<'_>) -> Outcome {
 ///   *without* running the modified function body, which is exactly what this lint flags, so they
 ///   behave like a `return` ([`Outcome::RETURNS`]).
 fn call_outcome(expr: &Expr<'_>) -> Option<Outcome> {
-    let ExprKind::Call(callee, _, _) = &expr.peel_parens().kind else { return None };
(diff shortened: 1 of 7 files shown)


crates/lint/src/sol/low/incorrect_modifier.rs
```diff
@@ -0,0 +1,180 @@
...
+    outcome
+}
+
+fn stmt_outcome(stmt: &Stmt<'_>) -> Outcome {
+    match &stmt.kind {
+        StmtKind::Placeholder => Outcome::COVERED,
+        StmtKind::Return(_) => Outcome::RETURNS,
+        StmtKind::Break => Outcome::BREAKS,
+        StmtKind::Continue => Outcome::CONTINUES,
+        StmtKind::Expr(expr) => call_outcome(expr).unwrap_or(Outcome::FALLTHROUGH),
```
mattsse, line 106: [P2] This expression handling treats every call other than the explicit revert/Yul halt builtins as fallthrough, so `require(false, ...)` and `assert(false)` are still considered reachable. That produces false positives for modifiers like `if (disabled) { require(false, "disabled"); return; } _;`: the branch always reverts before the `return`, so it cannot finish without `_`. Existing lints already model literal-false `require`/`assert` as aborting; please mirror that here and add a regression case.

mablr: f2e5eb85a

mattsse (approved):
LGTM

PR #15202, opened 2026-06-16
feat(lint): add mapping-deletion detector

0xMars42:
## Motivation

Adds the `mapping-deletion` detector from the `forge lint` parity checklist in #14381.

`delete` on a value whose type contains a `mapping` zeroes the non-mapping members but cannot enumerate the mapping's keys, so its entries are silently left in storage. The value looks reset while stale balances or flags remain reachable, which is a common accounting and access-control bug. `solc` does not warn on it.

## Solution

A `LateLintPass` (`med`, `mapping-deletion`) that flags `delete <expr>` when the operand's type reaches a `mapping`, directly or through a nested struct or array. It walks the semantic type (`Mapping` short-circuits; `Array`/`DynArray`/`Slice` recurse into the element; `Struct` recurses into the fields) with a `seen` guard against recursive struct definitions.

This catches `delete s`, `delete nested`, `delete arr`, `delete m[id]`, and `delete arr[i]`, while leaving plain structs, scalar members, and single mapping-entry deletes (`delete plainMap[k]`) untouched.

## Test

`crates/lint/testdata/MappingDeletion.sol` covers the five triggering shapes (direct struct, nested struct, array of structs, mapping value, array element) and three non-triggering ones (plain struct, scalar member, plain mapping entry), with the blessed `.stderr`. `cargo test -p forge --test ui` passes.


crates/lint/src/sol/med/mapping_deletion.rs
```diff
@@ -0,0 +1,67 @@
...
+    Gcx,
+    hir::{self, ExprKind},
+    ty::{Ty, TyKind},
+};
+
+declare_forge_lint!(
+    MAPPING_DELETION,
+    Severity::Med,
+    "mapping-deletion",
+    "`delete` on a struct containing a mapping does not clear the mapping"
```
mablr:
```suggestion
    "`delete` on a value containing a mapping does not clear the mapping"
```
more generic wording imo, pls bless ui test.

mablr (commented):
sgtm, just a smol wording nit

figtracer (dismissed):

crates/lint/src/sol/med/mapping_deletion.rs
0xMars42: Done, generalized to `value` and re-blessed. The lint does fire on arrays too (`delete arr`, `delete arr[i]`), so `value` is the accurate wording.

mablr (commented):
LGTM.

Don't forget to add lint to the Medium Severity list in `crates/lint/README.md`.

mattsse (dismissed):
LGTM

0xMars42:
Added to the Medium severity list in the README.

crates/lint/src/sol/med/mapping_deletion.rs
```diff
@@ -0,0 +1,67 @@
...
+    hir: &'hir hir::Hir<'hir>,
+    ty: Ty<'hir>,
+    seen: &mut Vec<hir::StructId>,
+) -> bool {
+    match ty.peel_refs().kind {
+        TyKind::Mapping(..) => true,
+        TyKind::Array(elem, _) | TyKind::DynArray(elem) | TyKind::Slice(elem) => {
+            ty_contains_mapping(gcx, hir, elem, seen)
+        }
+        TyKind::Struct(id) => {
```
figtracer, line 55: this recursive-struct path works, but the ui fixture does not cover it. i reproduced `struct tree { tree[] children; mapping(address => uint256) balances; }` plus `delete root`; please add that shape so the recursion guard stays covered.

mablr (dismissed):
Conflict resolved, LGTM. Thanks @0xMars42 !

stevencartavia (commented):
it has another conflict :c

stevencartavia (approved):

PR #15203, opened 2026-06-16
feat(lint): add tautological-compare detector

0xMars42:
## Motivation

Adds the `tautological-compare` detector from the `forge lint` parity checklist in #14381.

A comparison whose two sides are the same expression (`x >= x`, `a[i] < a[i]`) has a constant result. It is almost always a typo for a comparison against a different operand, or leftover dead code, and the compiler does not flag it.

## Solution

A `LateLintPass` (`med`, `tautological-compare`) that flags `a <op> a` for the relational and equality operators (`<`, `<=`, `>`, `>=`, `==`, `!=`) when both sides are structurally equal. Structural equality is restricted to side-effect-free reads: identifiers (compared by resolution), member access, and indexing by an equal index. Calls and everything else are treated as unequal, so a comparison whose sides could legitimately differ (e.g. `f() == f()`) is never flagged.

## Test

`crates/lint/testdata/TautologicalCompare.sol` covers six triggering forms (two identifiers, array indexing, mapping indexing, a `msg.sender` member access, and a parenthesized operand) and four non-triggering ones (different identifiers, different index, a call, and a non-self comparison), with the blessed `.stderr`. `cargo test -p forge --test ui` passes.


mablr (dismissed):
Just fixed FN, seems good to me.

mablr:
Pending book update merge https://github.com/foundry-rs/book/pull/1905

crates/lint/src/sol/med/tautological_compare.rs
```diff
@@ -0,0 +1,81 @@
...
+            && matches!(
+                op.kind,
+                BinOpKind::Lt
+                    | BinOpKind::Le
+                    | BinOpKind::Gt
+                    | BinOpKind::Ge
+                    | BinOpKind::Eq
+                    | BinOpKind::Ne
+            )
+            && exprs_equal(left, right)
```
figtracer, line 40: this false-positives on user-defined comparison operators for UDVTs. Solidity allows `using {f as ==} for T global`, and `x == x` then dispatches to `f(x, x)` rather than built-in equality, so it is not guaranteed to be always true/false. Repro: a `Weird` UDVT with `weirdEq(a, b) returns Weird.unwrap(a) + Weird.unwrap(b) > 10` gets warned on `return x == x;`. Consider skipping user-defined operator calls by checking the operand type/operator resolution before applying structural equality. https://www.soliditylang.org/blog/2023/02/22/user-defined-operators/

crates/lint/src/sol/med/tautological_compare.rs
```diff
@@ -0,0 +1,95 @@
...
+fn exprs_equal<'hir>(a: &Expr<'hir>, b: &Expr<'hir>) -> bool {
+    match (&a.peel_parens().kind, &b.peel_parens().kind) {
+        (ExprKind::Ident(ra), ExprKind::Ident(rb)) => ra == rb,
+        (ExprKind::Lit(la), ExprKind::Lit(lb)) => lits_equal(la, lb),
+        (ExprKind::Member(ba, na), ExprKind::Member(bb, nb)) => {
+            na.name == nb.name && exprs_equal(ba, bb)
+        }
+        (ExprKind::Index(ba, ia), ExprKind::Index(bb, ib)) => {
+            exprs_equal(ba, bb) && opt_exprs_equal(*ia, *ib)
+        }
```
mablr, line 75: This treats casts as unequal, which misses side-effect-free self-comparisons like `address(this) == address(this)` and `uint256(x) == uint256(x)`. Consider handling identical cast expressions whose inner expressions are structurally equal.

mablr (commented):
Almost there

crates/lint/src/sol/med/tautological_compare.rs
0xMars42: Added: `exprs_equal` now handles identical elementary-type casts, so `address(this) == address(this)` and `uint256(x) == uint256(x)` are flagged. Casts to different types (`uint256(x) == uint8(x)`) are left alone, since the narrower cast can truncate.

mablr: Nice, you could maybe also expand it to unary (`-x == -x`, `!b == !b`), ternary ? (`(c ? a : d) == (c ? a : d)`), and `payable(x) == payable(x)`.

mattsse (dismissed):
LGTM

crates/lint/src/sol/med/tautological_compare.rs
0xMars42: Done: exprs_equal now covers binary ops (`a + b == a + b`), the pure unary ops (`-`, `!`, `~`), the ternary, and `payable(...)`, all through the same side-effect-free recursion, so `++x == ++x`, side-effecting operands, and casts to different types are left alone. Also added it to the README severity list.

mablr (dismissed):
lgtm

figtracer (dismissed):
sg

mablr (approved):
Fixed typo, now G2G.

figtracer (approved):

PR #15206, opened 2026-06-17
feat(forge): add plateau-based invariant corpus sync

0xKarl98:
Towards OSS-377

This adds plateau-triggered, campaign-local corpus sharing for parallel invariant campaigns.

When invariant corpus sync is enabled (`plateau`, the default), workers publish newly discovered corpus entries into an in-memory exchange. A worker imports entries from sibling workers only after it has gone `invariant.corpus_sync.plateau_seconds` without producing new coverage. Imported entries are replayed against the worker's own coverage state and are kept only if they are useful locally, preserving the existing coverage-deduping behavior.

Implementation notes:
- Adds `invariant.corpus_sync` config with `mode = "plateau" | "off"` and `plateau_seconds`.
- Defaults to plateau mode with a 30-minute no-new-coverage window.
- Adds `--invariant-corpus-sync` / `FOUNDRY_INVARIANT_CORPUS_SYNC` and `--invariant-corpus-sync-seconds` / `FOUNDRY_INVARIANT_CORPUS_SYNC_SECONDS`.
- Uses a campaign-local append-only `InvariantCorpusExchange` instead of filesystem-based worker-to-worker sync.
- Enables sync only for deferred parallel corpus persistence, and skips timed campaigns whose timeout is not longer than the plateau window.
- Shares transaction sequences, cmp operands, and the coverage-dedupe policy so accepted imports can continue participating in mutation normally.


0xKarl98:
derek bench invariant

crates/config/src/invariant.rs
```diff
@@ -97,6 +97,61 @@ fn fixed_workers(workers: usize) -> Result<InvariantWorkers, String> {
...
+    pub plateau_seconds: Option<u32>,
+    /// Maximum candidate entries imported by one worker during a single sync.
+    pub max_imports_per_sync: usize,
+}
+
+impl Default for InvariantCorpusSyncConfig {
+    fn default() -> Self {
+        Self {
+            mode: InvariantCorpusSyncMode::Plateau,
+            plateau_runs: 3_000,
```
figtracer: with the default `invariant.runs = 256`, this default threshold can never trigger in a normal run-count campaign (and each parallel worker gets fewer than 256 runs). repro: `forge test --invariant-workers 2 --invariant-corpus-sync plateau` leaves `runs_since_new_coverage >= 3000` unreachable unless the user also raises runs or sets seconds, so corpus sync is enabled but never imports. consider deriving this from configured runs, lowering it, or defaulting `plateau_seconds`.

0xKarl98: Have lowered it

crates/config/src/invariant.rs
```diff
@@ -97,6 +97,61 @@ fn fixed_workers(workers: usize) -> Result<InvariantWorkers, String> {
...
+    pub plateau_seconds: Option<u32>,
+    /// Maximum candidate entries imported by one worker during a single sync.
+    pub max_imports_per_sync: usize,
+}
+
+impl Default for InvariantCorpusSyncConfig {
+    fn default() -> Self {
+        Self {
+            mode: InvariantCorpusSyncMode::Plateau,
+            plateau_runs: 64,
```
mablr: Lowering the default to 64 fixes the 2-worker default case, but it seems the plateau can still only fire at the end of common parallel campaigns. For example, `--invariant-runs 256 --invariant-workers 4` gives each worker 64 runs, so a worker that reaches `plateau_runs = 64` imports entries after its final run and never mutates them; with more workers the threshold is unreachable. Could you double-check please?

0xKarl98: Have fixed it , plz review again

mablr (commented):
sgtm, should we re-trigger bench before merge ? (unsure it's necessary given the fix scope)

mablr (commented):
I'm guessing (not sure), maybe relscore can be improved as the default plateau may be too late for relscore-style campaigns?
As far as i understand first sync happens after `min(64, worker_run_budget / 2)` runs without new coverage, but imported entries still need time to be selected, mutated, and turn into additional persisted coverage. With many workers, per-worker budgets can be small, so sync may occur around the midpoint and leave too little post-import exploration?

0xKarl98:
can explore that

0xKarl98:
derek bench invariant

grandizzy:

grandizzy:

grandizzy:
derek bench invariant timeout=14400

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -934,17 +983,86 @@ impl WorkerCorpus {
...
+                entry: entry.to_corpus_entry(Uuid::new_v4()),
+                remaining_mutations: mutation_budget,
+            });
+            stats.accepted += 1;
+            stats.shadowed += 1;
+        }
+        stats
+    }
+
+    pub(crate) fn shuffle_corpus(&mut self, rng: &mut impl Rng) {
```
0xalpharush: what is the point of this? don't we randomly select an index uniformly?

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -542,8 +588,10 @@ pub struct WorkerCorpus {
     current_mutated_index: Option<usize>,
     /// Config
     config: Arc<FuzzCorpusConfig>,
-    /// Indices of new entries added to [`WorkerCorpus::in_memory_corpus`] since last sync.
-    new_entry_indices: Vec<usize>,
+    /// New entries added to [`WorkerCorpus::in_memory_corpus`] since last sync.
+    new_entries: Vec<PendingCorpusEntry>,
+    /// Same-coverage imported candidates kept temporarily to escape local coverage shadowing.
+    shadow_candidates: Vec<ShadowCorpusCandidate>,
```
0xalpharush: just don't import them? not following what this is for

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -228,6 +228,52 @@ pub(crate) struct CampaignCorpusEntry {
     dedupe_by_coverage: bool,
 }
 
+/// Campaign-local corpus candidate exchanged between invariant workers.
+#[derive(Debug, Clone)]
+pub(crate) struct SharedCorpusEntry {
```
0xalpharush, line 233: we should just copy?

crates/config/src/invariant.rs
```diff
@@ -123,6 +147,46 @@ impl FromStr for InvariantDepthMode {
...
+    pub shuffle_on_sync: bool,
+}
+
+impl Default for InvariantCorpusSyncConfig {
+    fn default() -> Self {
+        Self {
+            mode: InvariantCorpusSyncMode::Plateau,
+            plateau_runs: 64,
+            plateau_seconds: Some(60),
+            max_imports_per_sync: 64,
```
0xalpharush: we should not put a max

crates/config/src/invariant.rs
```diff
@@ -123,6 +147,46 @@ impl FromStr for InvariantDepthMode {
...
 }
 
+/// Configuration for campaign-local corpus exchange between invariant workers.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(default)]
+pub struct InvariantCorpusSyncConfig {
+    /// Sync strategy.
+    pub mode: InvariantCorpusSyncMode,
+    /// Number of completed runs without new coverage before a plateau sync is attempted.
+    pub plateau_runs: u32,
```
0xalpharush: we should remove run-based plateau and only support time

crates/config/src/invariant.rs
```diff
@@ -123,6 +147,46 @@ impl FromStr for InvariantDepthMode {
...
+    /// Shuffle worker-local corpus order after a plateau sync to perturb stale selection bias.
+    pub shuffle_on_sync: bool,
+}
+
+impl Default for InvariantCorpusSyncConfig {
+    fn default() -> Self {
+        Self {
+            mode: InvariantCorpusSyncMode::Plateau,
+            plateau_runs: 64,
+            plateau_seconds: Some(60),
```
0xalpharush: this should be much higher probably >30min and not done if the campaign is shorter.

crates/evm/evm/src/executors/corpus.rs
0xKarl98: Will remove shadow inputs related logic and put them with restart machanism , according to this :
https://dl.acm.org/doi/10.1145/3712186

crates/config/src/invariant.rs
0xKarl98: This was only for experimenting to improve bench , will remove them

crates/evm/evm/src/executors/corpus.rs
0xKarl98: Have changed Arc to plain copy

0xKarl98:
derek bench invariant

PR #15212, opened 2026-06-17
feat(bench): add Foundry scfuzzbench runner

figtracer:
## Summary
- add a `foundry-scfuzzbench` benches binary for local Foundry scfuzzbench campaigns
- clone/pin scfuzzbench and target repos, select `forge` via `--foundry-bin`, `--foundry-ref`, or PATH, and run with isolated HOME/PATH
- harden campaign validation so failed setup/showmap replay cannot produce a misleading artifact bundle
- invoke scfuzzbench preparation/analysis/report targets and collect deterministic artifacts, including showmap/differential coverage CSVs plus `manifest.json` and `llm_summary.md`
- document the runner, artifact bundle, exact `forge` selection, optimization `--properties-path`, and local macOS/GNU tool shims in `benches/README.md`

OSS-332: https://linear.app/tempoxyz/issue/OSS-332


benches/src/scfuzzbench.rs
```diff
@@ -0,0 +1,1196 @@
...
+    fs::write(&path, lines.join("\n") + "\n")?;
+    Ok(path)
+}
+
+fn run_required(command: &mut Command) -> Result<()> {
+    let display = command_display(command);
+    let status = command.status().wrap_err_with(|| format!("failed to execute {display}"))?;
+    if !status.success() {
+        eyre::bail!("command failed ({status}): {display}");
+    }
```
mablr, line 1177: cancellation does not clean up child processes. When I stopped the hung runner, `make`, `uv`, and Python continued orphaned under PID 1 and had to be killed separately.

benches/src/scfuzzbench.rs
```diff
@@ -0,0 +1,1298 @@
...
+    let _ = sh_println!("📦 Resolving target repository pin");
+    let target_commit = clone_at(&cli.target_repo, &cli.target_ref, &dirs.target_pin)
+        .wrap_err("failed to clone target repository")?;
+
+    let foundry = select_foundry(&cli, &dirs).wrap_err("failed to select Foundry binary")?;
+    let _ = sh_println!("🔨 Foundry: {}", foundry.version_output.trim());
+
+    let run_id = format!("foundry-scfuzzbench-{}", chrono::Utc::now().format("%Y%m%d%H%M%S"));
+    let campaign_status = run_campaign(&cli, &dirs, &foundry, &target_commit, &run_id)
+        .wrap_err("failed to run scfuzzbench campaign")?;
```
mablr, line 255: `run_campaign()` returns the `local-run.sh` exit status, but `main()` never checks or classifies it before proceeding to analysis and reporting success. That means a non-zero campaign from a fuzzer crash/setup failure can still produce a successful artifact bundle if the later file checks pass, with the failure only buried in `manifest.json`.

benches/src/scfuzzbench.rs
```diff
@@ -0,0 +1,1428 @@
...
+    let cli = Cli::parse();
+
+    validate_options(&cli)?;
+    preflight(&cli)?;
+    prepare_output_dir(&cli.output_dir, cli.force)?;
+    let dirs = Dirs::new(cli.output_dir.clone());
+    dirs.create()?;
+    install_date_shim(&dirs.tools_bin)?;
+    install_timeout_shim(&dirs.tools_bin)?;
+
```
mablr, line 240: Optimization mode fails on macOS/BSD because scfuzzbench uses GNU-style `sed -i`; we need a sed shim like the existing date/timeout shims.

benches/src/scfuzzbench.rs
```diff
@@ -0,0 +1,1428 @@
...
+        eyre::bail!("{} has raw_trials=0", manifest_path.display());
+    }
+
+    let combined = manifest
+        .get("campaigns")
+        .and_then(|campaigns| campaigns.get("combined"))
+        .and_then(serde_json::Value::as_object)
+        .ok_or_else(|| {
+            eyre::eyre!("{} does not contain campaigns.combined", manifest_path.display())
+        })?;
```
mablr: Real scfuzzbench showmap manifests put approaches under `campaigns.combined.approaches`, so this validation rejects valid e2e runs with covered edges.

benches/src/scfuzzbench.rs
```diff
@@ -0,0 +1,1428 @@
...
+            manifest_path.display()
+        );
+    }
+
+    for csv in ["differential_coverage_relscores.csv", "differential_coverage_relcov.csv"] {
+        let path = dirs.data.join(csv);
+        ensure_csv_has_data_row(&path)?;
+    }
+    Ok(())
+}
```
mablr: Single-approach campaigns produce header-only `differential_coverage_relcov.csv`; relcov should only require data rows when comparing multiple approaches.

benches/src/scfuzzbench.rs
```diff
@@ -0,0 +1,1428 @@
...
+    if !root.exists() {
+        return Ok(());
+    }
+    for entry in fs::read_dir(root)? {
+        let entry = entry?;
+        let path = entry.path();
+        let name = entry.file_name().to_string_lossy().to_ascii_lowercase();
+        if name.contains("lcov") || name.contains("coverage-diff") || name.contains("coverage_diff")
+        {
+            out.push(path.clone());
```
mablr: This matches `relcov` as if it were `lcov`, causing `differential_coverage_relcov.csv` to be copied into `lcov-diff/`.

mablr (changes requested):
Tested e2e on real-world projects and found some sharp edges.

Please add regression coverage for the problems mentioned inline: nested showmap approaches, single-approach header-only relcov, multi-approach relcov rows, and the relcov/lcov filename false positive.

mablr (approved):
Re-tested locally, all previous concerns seem addressed, I haven't found other blockers.

Pending another review...

0xKarl98 (approved):

PR #15213, opened 2026-06-17
feat(lint): add incorrect-exp detector

0xMars42:
## Motivation

Adds the `incorrect-exp` detector from the `forge lint` parity checklist in #14381.

In Solidity `^` is bitwise xor, not exponentiation, so `10 ^ 18` evaluates to `24`, not `10 ** 18`. Developers coming from languages where `^` means power write it expecting exponentiation; the contract compiles and silently uses the wrong constant.

## Solution

A `LateLintPass` (`high`, `incorrect-exp`) that flags `a ^ b` when the base `a` is `2` or `10` and both operands are decimal integer literals, looking through elementary-type casts (`uint256(10)`) and parentheses. Hex operands are left alone.

This mirrors GCC's and Clang's `-Wxor-used-as-pow` (base restricted to `2` and `10`, decimal literals). Restricting the base to the values people actually write as powers (`2` for bit widths, `10` for decimals) avoids false positives on decimal bitmask xors such as `255 ^ 128` or `170 ^ 85`. For reference, Clippy's `suspicious_xor_used_as_pow`, which omits the base restriction, ships as an allow-by-default `restriction` lint because of that false-positive surface.

## Test

`crates/lint/testdata/IncorrectExp.sol` covers the flagged forms (`10 ^ 18`, `2 ^ 64`, `2 ^ 256`, and the casts `uint256(10) ^ 18`, `2 ^ uint256(64)`, `uint256(uint8(2)) ^ 64`) and the non-triggering ones (decimal bitmasks `255 ^ 128` / `170 ^ 85`, a base outside `2`/`10` with `3 ^ 4`, hex `0xff ^ 0x0f`, a variable operand, an actual `**`, and a non-cast call `id(10) ^ 18`), with the blessed `.stderr`. `cargo test -p forge --test ui` passes.

## Docs

Companion docs PR: foundry-rs/book#1904.

### DIFF
--- a/crates/lint/src/sol/high/incorrect_exp.rs
+++ b/crates/lint/src/sol/high/incorrect_exp.rs
@@ -3,11 +3,12 @@ use crate::{
     linter::{LateLintPass, LintContext},
     sol::{Severity, SolLint},
 };
+use alloy_primitives::U256;
 use solar::{
     ast::{BinOpKind, LitKind},
     sema::{
         Gcx,
-        hir::{self, Expr, ExprKind},
+        hir::{self, Expr, ExprKind, TypeKind},
     },
 };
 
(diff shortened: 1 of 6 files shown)


crates/lint/src/sol/high/incorrect_exp.rs
```diff
@@ -0,0 +1,53 @@
...
+        // `a ^ b` between two decimal integer literals is almost always a mistake for `a ** b`:
+        // `^` is bitwise xor in Solidity, so `10 ^ 18` is `24`, not `10 ** 18`. Hex literals are
+        // excluded since xor of bit patterns is a legitimate operation.
+        if let ExprKind::Binary(lhs, op, rhs) = &expr.kind
+            && matches!(op.kind, BinOpKind::BitXor)
+            && is_decimal_int_lit(lhs)
+            && is_decimal_int_lit(rhs)
+        {
+            ctx.emit(&INCORRECT_EXP, expr.span);
+        }
```
mablr, line 44: False negatives: the lint only accepts raw literal HIR nodes after parens, so obvious mistaken exponentiation through literal casts is missed: `uint256(10) ^ 18` and `10 ^ uint256(18)`.

mablr (changes requested):
Decimal literal xor is not always mistaken exponentiation (false-positives). The lint flags `255 ^ 128` and `170 ^ 85`, even though those are legitimate bit-mask xor expressions written in decimal.

crates/lint/src/sol/high/incorrect_exp.rs
0xMars42: Good catches, thanks. My first cut basically mirrored the reference detectors (flag `^` between decimal literals, skip hex), which felt like a fair starting point, but as you say it over-flags decimal bitmasks and misses casts.

On the false positives: I'd actually landed on "only flag base 2 and 10" myself, but it felt like a bit of a hack, so before committing to it I went and looked at how every tool that has this check handles it. Turns out it isn't a hack at all: GCC and Clang's `-Wxor-used-as-pow` use exactly that (base 2 or 10, decimal literals only), CodeQL and PVS-Studio do something similar, and Clippy, which drops the base restriction, ends up so noisy it ships allow-by-default. Once I tested it against your examples plus a few more, base 2/10 was clearly the best tradeoff: `255 ^ 128` and `170 ^ 85` stop flagging while `10 ^ 18`, `2 ^ 64` and `2 ^ 256` still do.

On the false negatives: the literal check now peels integer casts and parens, so `uint256(10) ^ 18` and `10 ^ uint256(18)` are caught.

Tests updated and re-blessed.

mablr: I think `1e1 ^ 18` is a false positive now because solar 1e1 evaluates to 10 as if it were the plain decimal literal 10, even though the rule is documented to target plain integer-literal power mistakes like 10 ^ 18.

We should prefer false-negatives to false-positives for this lint to avoid annoying devs.

0xMars42: Fixed: the operand check now only accepts a literal whose source text is plain decimal digits, so `1e1` (which solar folds to 10) no longer counts as a base or exponent.

Same spirit on two related cases I hit while there, preferring a false negative to a false positive:
- casts are only peeled through integer types now, so a legit `bytes32(uint256(2)) ^ bytes32(uint256(64))` is left alone;
- denominated literals (`2 wei`, `10 seconds`) are excluded too (the sub-denomination is dropped in the HIR, but the source span still has it).

crates/lint/src/sol/high/incorrect_exp.rs
```diff
@@ -0,0 +1,92 @@
...
+        // xor in Solidity, so `10 ^ 18` is `24`, not `10 ** 18`.
+        //
+        // To stay precise, the base is restricted to `2` and `10` (bit widths and decimals, the
+        // only bases people write as powers) and hex operands are left alone. This mirrors GCC's
+        // and Clang's `-Wxor-used-as-pow`; Clippy's `suspicious_xor_used_as_pow`, which drops the
+        // base restriction, is allow-by-default precisely because of the resulting false positives.
+        if let ExprKind::Binary(lhs, op, rhs) = &expr.kind
+            && matches!(op.kind, BinOpKind::BitXor)
+            && let Some(base) = plain_decimal_int_lit(ctx, lhs)
+            && (base == U256::from(2u64) || base == U256::from(10u64))
```
mablr, line 40: This still false-positives on small decimal bitmask XORs when the left operand happens to be 2 or 10.

mablr (commented):
Could you please check the reference impl in either Slither or Aderyn?

0xMars42:
Thanks, I had a look at both.

They use the same heuristic the first version of this PR did: flag `^` when an operand is a non-hex numeric literal. If anything theirs is a bit broader, a single literal operand is enough and there is no base restriction:

- Slither (`_is_constant_candidate` / `_is_bitwise_xor_on_constant`): https://github.com/crytic/slither/blob/5e78b88c0235885aa0a3f0ac3124d5eace3f1c40/slither/detectors/operations/incorrect_exp.py#L18-L33
- Aderyn (its comment literally says "Copied Heuristic from Slither"): https://github.com/Cyfrin/aderyn/blob/de6a0904a07ec20966b2b8f2ece4e89f0ada7bdb/aderyn_core/src/detect/high/incorrect_caret_operator.rs#L20-L52

So `255 ^ 128`, `1e1 ^ 18` and even `x ^ 1` are flagged there too. If you were pointing me at something specific in how they keep those out, I may have missed it, could you point me to it?

mablr:
Thank you @0xMars42 !

This is fine then, I just wanted to avoid this lint to be noisier than slither/aderyn.

mablr (dismissed):

mattsse (approved):
LGTM

PR #15217, opened 2026-06-18
feat(symbolic): persist replayable counterexample artifacts

mablr:
## Description

Closes OSS-349

Adds symbolic counterexample artifacts that can be replayed independently from the original symbolic run. When `forge test --symbolic` confirms a counterexample, Forge writes a schema-versioned JSON artifact under the project cache and surfaces its path in JSON output and non-JSON test output.

Artifacts can be replayed with:

```bash
forge test --replay-symbolic-artifact <path>
```

Replay supports both single-call symbolic counterexamples and invariant/stateful sequence counterexamples.

**What changed**

- Persist confirmed symbolic counterexamples as `foundry:symbolic.counterexample@v1` artifacts.
- Add `--replay-symbolic-artifact <PATH>` to reconstruct the target contract/test/path and replay the stored calls.
- Add normalized JSON references via `counterexample_artifacts`.
- Surface artifact paths in human output as warnings.
- Validate replay artifacts before execution, including schema header, empty calls, target/function matching, and single-call value constraints.
- Make replay CI-friendly:
  - still-reproducing counterexample fails
  - fixed counterexample passes
  - stale or ambiguous target fails
- Cover single-call, invariant sequence, replay skip, stale target, fixed bug, and schema behavior in tests.
## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/evm/evm/src/executors/invariant/mod.rs
+++ b/crates/evm/evm/src/executors/invariant/mod.rs
@@ -1918,6 +1918,31 @@ pub fn execute_tx<FEN: FoundryEvmNetwork>(
         .map_err(|e| eyre!(format!("Could not make raw evm call: {e}")))
 }
 
+/// Executes a replay call on a validation executor and registers any created targets.
+///
+/// This mirrors sequence replay's non-reverted commit behavior while allowing callers to update
+/// updatable target sets before validating later calls in the same artifact.
+pub fn execute_tx_and_register_created<FEN: FoundryEvmNetwork>(
+    executor: &mut Executor<FEN>,
+    tx: &BasicTxDetails,
+    targeted_contracts: &FuzzRunIdentifiedContracts,
+    dynamic_target_ctx: &DynamicTargetCtx<'_>,
+    created_contracts: &mut Vec<Address>,
+) -> Result<()> {
+    let mut call_result = execute_tx(executor, tx)?;
+    if !call_result.reverted {
+        targeted_contracts.collect_created_contracts(
+            &call_result.state_changeset,
+            dynamic_target_ctx.project_contracts,
+            dynamic_target_ctx.setup_contracts,
+            dynamic_target_ctx.artifact_filters,
+            created_contracts,
+        )?;
+        executor.commit(&mut call_result);
+    }
+    Ok(())
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/crates/evm/evm/src/executors/invariant/mod.rs
+++ b/crates/evm/evm/src/executors/invariant/mod.rs
@@ -1878,7 +1878,7 @@ pub(crate) fn call_invariant_function<FEN: FoundryEvmNetwork>(
 
 /// Executes a fuzz call and returns the result.
 /// Applies any block timestamp (warp) and block number (roll) adjustments before the call.
-pub(crate) fn execute_tx<FEN: FoundryEvmNetwork>(
+pub fn execute_tx<FEN: FoundryEvmNetwork>(
     executor: &mut Executor<FEN>,
     tx: &BasicTxDetails,
 ) -> Result<RawCallResult<FEN>> {
--- a/crates/evm/evm/src/executors/invariant/mod.rs
+++ b/crates/evm/evm/src/executors/invariant/mod.rs
@@ -1934,7 +1934,12 @@ pub(crate) fn call_invariant_function<FEN: FoundryEvmNetwork>(
     Ok((call_result, success))
 }
 
-/// Executes a fuzz call and returns the result.
+/// Executes an invariant replay fuzz call and returns the result.
+///
+/// This applies invariant replay semantics: warp/roll deltas are applied before the call and the
+/// requested value is clamped to the sender balance. It is intended for invariant sequence replay,
+/// shrinking, and artifact validation rather than as a general raw-call helper.
+///
 /// Applies any block timestamp (warp) and block number (roll) adjustments before the call.
 pub fn execute_tx<FEN: FoundryEvmNetwork>(
     executor: &mut Executor<FEN>,
@@ -1976,7 +1981,7 @@ pub fn execute_tx<FEN: FoundryEvmNetwork>(
         .map_err(|e| eyre!(format!("Could not make raw evm call: {e}")))
 }
 
-/// Executes a replay call on a validation executor and registers any created targets.
+/// Executes an invariant replay call on a validation executor and registers created targets.
 ///
 /// This mirrors sequence replay's non-reverted commit behavior while allowing callers to update
 /// updatable target sets before validating later calls in the same artifact.
--- a/crates/evm/evm/src/executors/invariant/shrink.rs
+++ b/crates/evm/evm/src/executors/invariant/shrink.rs
@@ -255,7 +255,7 @@ pub(crate) fn shrink_sequence<FEN: FoundryEvmNetwork>(
         // do not roll back the removal.
         ShrinkErrorPolicy::KeepRemoved,
         |shrinker| {
-            let (success, _, _) = check_sequence(
+            let (success, _, _, _) = check_sequence(
                 executor.clone(),
                 calls,
                 shrinker.current().collect(),
@@ -357,13 +357,17 @@ pub fn check_sequence<FEN: FoundryEvmNetwork>(
     test_address: Address,
     calldata: Bytes,
     options: CheckSequenceOptions<'_>,
-) -> eyre::Result<(bool, bool, Option<String>)> {
+) -> eyre::Result<(bool, bool, Option<String>, usize)> {
+    let mut reverts = 0;
     let early = replay_sequence(
         &mut executor,
         calls,
         &sequence,
         options.accumulate_warp_roll,
         |_idx, call_result| {
+            if call_result.reverted {
+                reverts += 1;
+            }
             // Ignore calls reverted with `MAGIC_ASSUME`. This is needed to handle failed
             // scenarios that are replayed with a modified version of test driver (that use
             // new `vm.assume` cheatcodes).
@@ -375,16 +379,18 @@ pub fn check_sequence<FEN: FoundryEvmNetwork>(
                     false,
                     false,
                     assertion_failure_reason(call_result, options.rd),
+                    reverts,
                 )));
             }
             if call_result.reverted && options.fail_on_revert {
                 if options.expect_assertion_failure {
-                    return Ok(ReplayDecision::Stop((true, false, None)));
+                    return Ok(ReplayDecision::Stop((true, false, None, reverts)));
                 }
                 return Ok(ReplayDecision::Stop((
                     false,
                     false,
                     call_failure_reason(call_result, options.rd),
+                    reverts,
                 )));
             }
             Ok(ReplayDecision::Continue(call_result))
@@ -396,7 +402,9 @@ pub fn check_sequence<FEN: FoundryEvmNetwork>(
 
     // Unlike optimization mode we intentionally do not apply trailing warp/roll before the
     // invariant call: those delays would not be representable in the final shrunk sequence.
-    finish_sequence_check(&executor, test_address, calldata, &options)
+    let (success, replayed_entirely, reason) =
+        finish_sequence_check(&executor, test_address, calldata, &options)?;
+    Ok((success, replayed_entirely, reason, reverts))
 }
 
 fn finish_sequence_check<FEN: FoundryEvmNetwork>(
--- a/crates/evm/evm/src/executors/invariant/shrink.rs
+++ b/crates/evm/evm/src/executors/invariant/shrink.rs
@@ -255,7 +255,7 @@ pub(crate) fn shrink_sequence<FEN: FoundryEvmNetwork>(
         // do not roll back the removal.
         ShrinkErrorPolicy::KeepRemoved,
         |shrinker| {
-            let (success, _, _, _) = check_sequence(
+            let (success, _, _, _, _) = check_sequence(
                 executor.clone(),
                 calls,
                 shrinker.current().collect(),
@@ -357,14 +357,14 @@ pub fn check_sequence<FEN: FoundryEvmNetwork>(
     test_address: Address,
     calldata: Bytes,
     options: CheckSequenceOptions<'_>,
-) -> eyre::Result<(bool, bool, Option<String>, usize)> {
+) -> eyre::Result<(bool, bool, Option<String>, usize, usize)> {
     let mut reverts = 0;
     let early = replay_sequence(
         &mut executor,
         calls,
         &sequence,
         options.accumulate_warp_roll,
-        |_idx, call_result| {
+        |idx, call_result| {
             // Ignore calls reverted with `MAGIC_ASSUME`. This is needed to handle failed
             // scenarios that are replayed with a modified version of test driver (that use
             // new `vm.assume` cheatcodes).
@@ -379,17 +379,19 @@ pub fn check_sequence<FEN: FoundryEvmNetwork>(
                     false,
                     false,
                     assertion_failure_reason(call_result, options.rd),
+                    idx + 1,
                     reverts,
                 )));
             }
             if call_result.reverted && options.fail_on_revert {
                 if options.expect_assertion_failure {
-                    return Ok(ReplayDecision::Stop((true, false, None, reverts)));
+                    return Ok(ReplayDecision::Stop((true, false, None, idx + 1, reverts)));
                 }
                 return Ok(ReplayDecision::Stop((
                     false,
                     false,
                     call_failure_reason(call_result, options.rd),
+                    idx + 1,
                     reverts,
                 )));
             }
@@ -404,7 +406,7 @@ pub fn check_sequence<FEN: FoundryEvmNetwork>(
     // invariant call: those delays would not be representable in the final shrunk sequence.
     let (success, replayed_entirely, reason) =
         finish_sequence_check(&executor, test_address, calldata, &options)?;
-    Ok((success, replayed_entirely, reason, reverts))
+    Ok((success, replayed_entirely, reason, sequence.len(), reverts))
 }
 
 fn finish_sequence_check<FEN: FoundryEvmNetwork>(
--- a/crates/evm/evm/src/executors/invariant/shrink.rs
+++ b/crates/evm/evm/src/executors/invariant/shrink.rs
@@ -365,14 +365,14 @@ pub fn check_sequence<FEN: FoundryEvmNetwork>(
         &sequence,
         options.accumulate_warp_roll,
         |_idx, call_result| {
-            if call_result.reverted {
-                reverts += 1;
-            }
             // Ignore calls reverted with `MAGIC_ASSUME`. This is needed to handle failed
             // scenarios that are replayed with a modified version of test driver (that use
             // new `vm.assume` cheatcodes).
             if call_result.result.as_ref() == MAGIC_ASSUME {
                 return Ok(ReplayDecision::Continue(call_result));
+            }
+            if call_result.reverted {
+                reverts += 1;
             }
             if did_fail_on_assert(&call_result, &call_result.state_changeset) {
                 return Ok(ReplayDecision::Stop((
--- a/crates/evm/evm/src/executors/invariant/shrink.rs
+++ b/crates/evm/evm/src/executors/invariant/shrink.rs
@@ -358,13 +358,15 @@ pub fn check_sequence<FEN: FoundryEvmNetwork>(
     calldata: Bytes,
     options: CheckSequenceOptions<'_>,
 ) -> eyre::Result<(bool, bool, Option<String>, usize, usize)> {
+    let mut calls_executed = 0;
     let mut reverts = 0;
     let early = replay_sequence(
         &mut executor,
         calls,
         &sequence,
         options.accumulate_warp_roll,
-        |idx, call_result| {
+        |_idx, call_result| {
+            calls_executed += 1;
             // Ignore calls reverted with `MAGIC_ASSUME`. This is needed to handle failed
             // scenarios that are replayed with a modified version of test driver (that use
             // new `vm.assume` cheatcodes).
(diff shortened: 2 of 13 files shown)


crates/forge/src/runner.rs
```diff
@@ -1144,6 +1324,133 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                    &txes,
+                    (0..txes.len()).collect(),
+                    self.setup.address,
+                    invariant.selector().to_vec().into(),
+                    CheckSequenceOptions {
+                        // Artifact replay executes every stored call in order, so each call's
+                        // warp/roll delta is applied directly. Accumulation is only needed when a
+                        // shrink candidate skips calls and must fold removed delays forward.
+                        accumulate_warp_roll: false,
+                        fail_on_revert: self.config.invariant.fail_on_revert,
```
figtracer: this makes sequence artifacts depend on the current invariant config, not the config that confirmed the artifact. i reproduced a fail_on_revert=true artifact whose stored call still reverts, then replayed it with fail_on_revert=false and got exit 0. please store the replay semantics in the artifact or reject config drift.

figtracer (commented):
reviewed the replay paths and reproduced one config-drift edge case locally.

crates/forge/src/runner.rs
```diff
@@ -1144,6 +1324,133 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                        expect_assertion_failure: false,
+                        call_after_invariant,
+                        rd: Some(self.revert_decoder()),
+                    },
+                ) {
+                    Ok((success, replayed_entirely, reason)) => {
+                        if success {
+                            self.result.status = TestStatus::Success;
```
figtracer: fixed sequence replays should still be typed as invariant/replay results. right now this leaves the default unit kind, so json reports success as { unit: { gas: 0 } }.

crates/forge/src/runner.rs
```diff
@@ -1144,6 +1328,133 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                        accumulate_warp_roll: false,
+                        fail_on_revert: artifact.replay_semantics.fail_on_revert,
+                        expect_assertion_failure: false,
+                        call_after_invariant,
+                        rd: Some(self.revert_decoder()),
+                    },
+                ) {
+                    Ok((success, replayed_entirely, reason)) => {
+                        if success {
+                            self.result.status = TestStatus::Success;
```
figtracer: this still leaves successful sequence replays as the default `unit { gas: 0 }` result. the failure branch sets invariant metadata via `invariant_replay_fail`, but the fixed-artifact path only updates status/reason, so `--replay-symbolic-artifact --json` reports a fixed invariant replay as a unit test. can this set the invariant/replay kind here and add a json regression for the fixed sequence case?

crates/forge/src/result.rs
```diff
@@ -1563,6 +1938,21 @@ impl TestResult {
         self.counterexample = Some(CounterExample::Sequence(call_sequence.len(), call_sequence));
     }
 
+    /// Returns the success result for a replayed invariant test.
+    pub fn invariant_replay_success(&mut self, call_count: usize) {
+        self.kind = TestKind::Invariant {
+            runs: 1,
+            calls: call_count,
+            reverts: 0,
```
figtracer: this always reports `reverts: 0`, but a fixed sequence replay can still execute reverted calls when stored `fail_on_revert` is false. can we pass the observed revert count through and assert it in the json test?

crates/forge/src/runner.rs
```diff
@@ -1144,6 +1328,132 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                    .calls
+                    .iter()
+                    .map(SymbolicCounterexampleCall::to_base_counterexample)
+                    .collect::<Vec<_>>();
+                let txes = artifact
+                    .calls
+                    .iter()
+                    .map(SymbolicCounterexampleCall::to_basic_tx_details)
+                    .collect::<Vec<_>>();
+                match check_sequence(
```
figtracer, line 1561: sequence replay accepts stale call targets. i could change a stored call to an empty address and get a passing replay; can we validate each stored target/selector before `check_sequence`?

crates/forge/src/runner.rs
```diff
@@ -729,6 +732,87 @@ impl<'a, FEN: FoundryEvmNetwork> ContractRunner<'a, FEN> {
...
 
+        if let Some(replay) = &self.mcr.tcfg.symbolic_artifact_replay {
+            let artifact = &replay.artifact;
+            let replay_functions = functions
+                .iter()
+                .filter(|func| func.signature() == artifact.test.test)
+                .copied()
+                .collect::<Vec<_>>();
+
+            if replay_functions.is_empty() {
```
figtracer, line 772: this can fail replay in multi-network runs: a non-target pass reports the target as missing after another pass already replayed it. can non-matching passes return no result and let the final exactly-one check handle true misses?

crates/forge/src/runner.rs
```diff
@@ -1144,6 +1331,177 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                    &setup_contracts,
+                    &self.cr.mcr.known_contracts,
+                    self.cr.num_invariant_campaign_anchors,
+                );
+                if let Err(err) = evm.select_contract_artifacts(self.address) {
+                    self.result.invariant_setup_fail(err);
+                    return self.result;
+                }
+                let targeted = match evm.select_contracts_and_senders(self.address) {
+                    Ok((_, targeted)) => targeted,
```
stevencartavia: Sequence replay discards the invariant sender filters and only checks target+selector via `can_replay`, so a replayed call can run from a sender that `excludeSender`/`targetSender` forbids — reaching `msg.sender`-gated paths a normal generated sequence wouldn't.

Repro: `Vault.drain()` gated to an owner that's excluded (or a different sender allowlisted). Normal run passes; replaying an artifact with `sender = owner` executes `drain()` and breaks the invariant.

Can we validate each call's `sender` against the computed targeted/excluded senders?

crates/forge/src/runner.rs
```diff
@@ -1144,6 +1331,177 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                    let targets = targeted.targets();
+                    for (idx, tx) in txes.iter().enumerate() {
+                        let Some(selector) = tx.call_details.calldata.get(..4) else {
+                            self.result.single_fail(Some(format!(
+                                "sequence symbolic artifact call {} has calldata shorter than a selector",
+                                idx + 1
+                            )));
+                            return self.result;
+                        };
+                        if !targets.can_replay(tx) {
```
stevencartavia: For updatable invariants (no `targetContract`), the upfront `can_replay` loop validates calls against the initial target set only. If call 1 deploys a contract that call 2 targets, replay fails with `targets unknown function …` before execution — even though the campaign/corpus paths register created contracts between calls and `check_sequence` could execute it fine.

Repro: `Spawner.spawn()` deploys a `Bomb`; `Bomb.explode()` trips the invariant. Normal run replays `[spawn(), explode()]`; the same artifact is rejected.

Could we register mid-sequence created contracts (like `register_replay_created`) or drop the upfront rejection and let `check_sequence` decide?

crates/forge/src/runner.rs
mablr: https://github.com/foundry-rs/foundry/pull/15217/commits/a13d472c294487c9124e2d2a8545a2051c408528

crates/forge/src/runner.rs
mablr: https://github.com/foundry-rs/foundry/pull/15217/commits/eb7fce26c2650104beb25a0ff7e4ae30f1217202

crates/forge/src/runner.rs
```diff
@@ -1144,6 +1332,211 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                    let mut validation_created_contracts = Vec::new();
+                    for (idx, tx) in txes.iter().enumerate() {
+                        let Some(selector) = tx.call_details.calldata.get(..4) else {
+                            self.result.single_fail(Some(format!(
+                                "sequence symbolic artifact call {} has calldata shorter than a selector",
+                                idx + 1
+                            )));
+                            return self.result;
+                        };
+                        if !targeted.targets().has_abi_function(tx) {
```
figtracer: this ignores target/exclude selector filters and accepts any abi selector. can we validate against the fuzzed function set instead?

crates/forge/src/runner.rs
```diff
@@ -1144,6 +1332,211 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                                "sequence symbolic artifact call {} targets unknown function {} on {}",
+                                idx + 1,
+                                hex::encode_prefixed(selector),
+                                tx.call_details.target
+                            )));
+                            return self.result;
+                        }
+                        if (!sender_filters.targeted.is_empty()
+                            && !sender_filters.targeted.contains(&tx.sender))
+                            || (tx.sender != Address::ZERO
```
figtracer: address(0) is excluded by default, but this bypass lets artifacts replay from it. can we reject it here too?

crates/forge/src/runner.rs
```diff
@@ -1144,6 +1332,211 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                        expect_assertion_failure: false,
+                        call_after_invariant,
+                        rd: Some(self.revert_decoder()),
+                    },
+                ) {
+                    Ok((success, replayed_entirely, reason, reverts)) => {
+                        if success {
+                            self.result.invariant_replay_success(txes.len(), reverts);
+                        } else {
+                            self.result.invariant_replay_fail(
```
figtracer, line 1582: this drops the observed replay counts on the failure path, so json still reports calls/reverts as 1/1. can we pass those counts through too?

crates/forge/src/runner.rs
mablr: mb, last commit was a foot-gun, I reverted https://github.com/foundry-rs/foundry/pull/15217/commits/388999a38b9ad1d41e194ccaf681cccf59f980a6

crates/forge/src/runner.rs
mablr: https://github.com/foundry-rs/foundry/pull/15217/commits/efc7e5badfbfb7fd50d67186812b93293bb8661b

crates/forge/src/cmd/test/mod.rs
```diff
@@ -612,7 +706,33 @@ impl TestArgs {
...
+                .rsplit_once(':')
+                .map_or(("", replay.artifact.test.contract.as_str()), |(path, contract)| {
+                    (path, contract)
+                });
+            filter_args.test_pattern =
+                Some(Regex::new(&format!("^{}$", regex::escape(&replay.artifact.test.test)))?);
+            filter_args.contract_pattern =
+                Some(Regex::new(&format!("^{}$", regex::escape(contract)))?);
+            if !path.is_empty() {
+                filter_args.path_pattern = Some(path.parse::<GlobMatcher>()?);
```
figtracer: `--replay-symbolic-artifact` treats the artifact path as an exact target, but this reuses it as a glob. i reproduced with a bracketed test path: artifact creation succeeds, then replay says the target is missing because the brackets are interpreted as a character class. can we escape the path before building the matcher, e.g. `globset::escape(path)`, or use a literal path matcher, with a regression for bracketed paths?

figtracer (dismissed):

crates/evm/evm/src/executors/invariant/shrink.rs
```diff
@@ -396,7 +404,9 @@ pub fn check_sequence<FEN: FoundryEvmNetwork>(
 
     // Unlike optimization mode we intentionally do not apply trailing warp/roll before the
     // invariant call: those delays would not be representable in the final shrunk sequence.
-    finish_sequence_check(&executor, test_address, calldata, &options)
+    let (success, replayed_entirely, reason) =
+        finish_sequence_check(&executor, test_address, calldata, &options)?;
+    Ok((success, replayed_entirely, reason, sequence.len(), reverts))
```
grandizzy: 🔴 **blocker:** the early-stop branches (L382, L388, L394) return `idx + 1`, but `idx` is the index into the original `calls`, not the count of *executed* calls. For a non-contiguous shrink candidate like `[2,4]`, the first executed call reports `calls=3` instead of `1`. Now that `check_sequence` is `pub` and the tuple carries semantic metrics, track an independent `calls_executed` counter (like `reverts` at L361) and return it in every `Stop` branch and here at the final `Ok` instead of `idx + 1` / `sequence.len()`.

crates/forge/src/runner.rs
```diff
@@ -1144,6 +1332,214 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                    self.result.single_fail(Some(err));
+                    return self.result;
+                }
+
+                if self.prepare_test(func).is_err() {
+                    return self.result;
+                }
+
+                let mut executor = self.clone_executor();
+                match execute_tx(&mut executor, &call.to_basic_tx_details()) {
```
grandizzy: 🔴 **blocker:** this doesn't mirror `run_symbolic_test`'s confirmation path (`executor.call(...)` → `into_result`). (1) `vm.skip` is no longer classified as `EvmError::Skip`, so a skipped replay reports a false counterexample; (2) on failure we surface `artifact.replay.reason` instead of decoding the *current* revert reason — output lies if the test changed but fails differently; (3) `warp`/`roll` on a single-call artifact get applied here though the original single-call path never did. Please decode the live reason (fall back to `artifact.replay.reason`), preserve skip handling, reject `warp.is_some() || roll.is_some()` for single-call, and require `replay.status == confirmed`.

crates/forge/src/runner.rs
```diff
@@ -1527,6 +1929,25 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
                 ) {
                     Ok(replayed_call_sequence) if !replayed_call_sequence.is_empty() => {
                         call_sequence = replayed_call_sequence;
+                        let (_txes, replay) = replay_persisted_call_sequence(
```
grandizzy, line 1993: 🔴 **blocker:** `replay_persisted_call_sequence` is called a second time here (first at L1883) only to refresh `replayed_entirely`/`reason`/`calls`/`reverts`. This is a redundant full-sequence EVM execution, and under any time/fork/cheatcode nondeterminism the second run can disagree with the replay that actually established the counterexample. Have the first replay (or `replay_run`) return the final `CheckSequenceResult` instead of re-executing.

crates/forge/src/runner.rs
```diff
@@ -2394,6 +2857,52 @@ fn invariant_contract_name(contract_name: &str) -> &str {
...
+    let sanitized = value
+        .chars()
+        .map(|ch| if ch.is_ascii_alphanumeric() || ch == '-' || ch == '_' { ch } else { '_' })
+        .collect::<String>();
+    if sanitized.is_empty() { "_".to_string() } else { sanitized }
+}
+
+fn symbolic_artifact_file_name(value: &str) -> String {
+    let hash = keccak256(value.as_bytes());
+    let hash = hex::encode(&hash[..4]);
```
grandizzy: 🔴 **blocker:** 4 bytes of hash is a weak durable cache key, and it only hashes the file-name component, not the full `path:Contract::test` + kind identity — different sanitized contracts/names can clobber each other. Hash the full identity and use ≥16 bytes, e.g.:
```rust
let identity = format!("{contract_id}\0{value}\0{kind:?}");
let hash = hex::encode(&keccak256(identity.as_bytes())[..16]);
```

crates/forge/src/cmd/test/mod.rs
```diff
@@ -458,6 +486,72 @@ impl TestArgs {
...
+        })?;
+        if schema != SYMBOLIC_COUNTEREXAMPLE_ARTIFACT_SCHEMA {
+            bail!(
+                "unsupported symbolic counterexample artifact schema `{}` in {}",
+                schema,
+                path.display()
+            );
+        }
+        let artifact = serde_json::from_value::<SymbolicCounterexampleArtifact>(value).wrap_err(
+            format!("failed to parse symbolic counterexample artifact {}", path.display()),
```
grandizzy, line 532: 🟡 non-blocking: `serde_json::from_value` accepts unknown fields even though the schema declares `additionalProperties: false`. Add `#[serde(deny_unknown_fields)]` to the artifact structs (or run real schema validation). Consider also rejecting artifacts whose `replay.status != confirmed` here.

crates/forge/src/cmd/test/mod.rs
```diff
@@ -612,7 +706,33 @@ impl TestArgs {
         // Set up the project.
         let project = config.project()?;
 
-        let filter = self.filter(&config)?;
+        let replay_symbolic_artifact = self.load_symbolic_artifact_replay()?;
+        if replay_symbolic_artifact.is_some() {
+            config.symbolic.enabled = true;
```
grandizzy, line 718: 🟡 non-blocking: sequence replay needs no solver, yet this enables symbolic mode globally for all replays. If enabling symbolic has side effects (solver validation, test discovery), prefer an explicit replay execution mode over flipping this flag.

crates/forge/src/cmd/test/mod.rs
```diff
@@ -630,12 +750,22 @@ impl TestArgs {
...
+            false,
+            replay_symbolic_artifact,
+        )
+        .await
     }
 
     /// Executes all the tests in the project.
     ///
     /// See [`Self::compile_and_run`] for more details.
+    #[allow(clippy::too_many_arguments)]
```
grandizzy, line 775: 🟡 non-blocking: `run_tests` now needs `#[allow(clippy::too_many_arguments)]` (also at L1296/L1335). Since `replay_symbolic_artifact` already lives in `TestRunnerConfig`, consider a `RunTestsOptions` struct instead of threading another positional arg.

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -1936,7 +1936,7 @@ pub(crate) fn call_invariant_function<FEN: FoundryEvmNetwork>(
 
 /// Executes a fuzz call and returns the result.
 /// Applies any block timestamp (warp) and block number (roll) adjustments before the call.
-pub(crate) fn execute_tx<FEN: FoundryEvmNetwork>(
+pub fn execute_tx<FEN: FoundryEvmNetwork>(
```
grandizzy, line 1944: 🟡 non-blocking: `execute_tx` went `pub(crate)`→`pub` and `execute_tx_and_register_created` (L1983) is new `pub`, exposing invariant-specific semantics (warp/roll deltas, value clamping, non-reverted commit, dynamic target registration) across crate boundaries. Prefer a narrower replay-oriented helper, or at least doc-comment these as invariant-replay helpers rather than general EVM call helpers.

crates/forge/src/result.rs
```diff
@@ -1046,6 +1403,14 @@ pub struct TestResult {
     /// Minimal reproduction test case for failing test
     pub counterexample: Option<CounterExample>,
 
+    /// Durable replay artifact for the top-level counterexample, when one was written.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub counterexample_artifact: Option<SymbolicArtifactRef>,
+
+    /// All durable replay artifacts produced for this test result, normalized for JSON consumers.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub counterexample_artifacts: Vec<SymbolicArtifactRef>,
```
grandizzy, line 1454: 🟡 non-blocking: keeping singular `counterexample_artifact` (L1408) + plural `counterexample_artifacts` is fine for compat and `add_counterexample_artifact` keeps them synced, but a manually constructed/deserialized `TestResult` can diverge. Mark the singular field legacy/deprecated and point new consumers at the plural.

crates/evm/symbolic/assets/symbolic-counterexample.schema.json
```diff
@@ -0,0 +1,138 @@
...
+        "calldata",
+        "value",
+        "contract_name",
+        "function_name",
+        "signature",
+        "args",
+        "raw_args"
+      ],
+      "properties": {
+        "warp": {
```
grandizzy, line 96: 🟡 non-blocking: the schema requires compact `0x` hex for `warp`/`roll`/`value`, but the loader doesn't validate against the schema and Alloy's `U256` deserializer may accept other forms — a schema-contract mismatch on input. Add a producer test that `Some(U256::ZERO)` serializes to `"0x0"` (schema allows `0x0`, forbids leading-zero forms). Also note the external HTTPS `$ref`s mean the offline test only checks ref *existence*, not real validation.

grandizzy (dismissed):
Please check findings below

grandizzy (approved):
Re-reviewed after the latest commits. All four earlier blockers are resolved or mitigated:

- **Single-call replay semantics** (`a3ffefe`): now gates on `replay.status == Confirmed`, rejects `warp`/`roll`, uses `call_raw` + `is_raw_call_success` to mirror the original symbolic path, decodes the live revert reason (falling back to the artifact reason only when empty), and routes `EvmError::Skip` to `single_skip`. ✅
- **`check_sequence` executed-call counts** (`96ea2e7`): independent `calls_executed` counter replaces `idx + 1`/`sequence.len()`. ✅
- **Artifact filename identity** (`415d8b5`): 16-byte keccak over the full `contract\0value\0kind` identity, with a collision test. ✅
- **Double `replay_persisted_call_sequence`**: still executes twice, but the second pass now recomputes metrics for the (possibly changed) sequence returned by `replay_error`, so it's a minor efficiency follow-up rather than a correctness issue. ✅ (mitigated)

Most non-blocking items were also addressed: `#[serde(deny_unknown_fields)]` on the artifact structs, loader rejecting non-confirmed artifacts, a `U256::ZERO` compact-hex test, doc-comments scoping `execute_tx`/`execute_tx_and_register_created` to invariant replay, and the legacy singular `counterexample_artifact` field documented in favor of the plural.

Remaining nits are optional follow-ups: `config.symbolic.enabled = true` for sequence replay, the `run_tests` argument count, and true offline schema validation for the external `$ref`s.

LGTM, nice work.

figtracer (approved):

Issue #927, opened 2022-03-14
Meta: Debugger improvements


onbjerg:
**Decoding/display**
- [ ] Guidance on how to debug failed tests in the CLI output
- [ ] Visual clues that a precompile is being called (requires more thought, see #916)
- [ ] Showing current values of variables (https://github.com/gakonst/foundry/issues/410)

**Quality of life**
- [ ] Greying out of zero values in memory (https://github.com/gakonst/foundry/issues/902)

**Control flow**
- [ ] Breakpoints

**Capabilities**
- [ ] Debugging functions with parameters (e.g. fuzz tests)
- [ ] Debugging of constructors

onbjerg:
Idea for an interactive debugger: some `Debugger` struct that handles state/displays UI. Messages are passed between the `Debugger` and the debug inspector. Contracts are identified on the fly when the EVM switches context using any number of identifiers

henry-hz:
@onbjerg supose I want to debug the function in the test below called `symbol()`. Using the `forge test --debug symbol` I have to jump a looootttt until I find it. What would be needed to jump directly to my function using the debugger ? I have the impression that using the function signature `0x95d89b41` as an input we could just search it and jump the counter to there. Makes sense ? Is there anyone writing it ?

(I am using the [foundry-huff](https://github.com/huff-language/foundry-huff))

```solidity
    function testSymbol() public {
        string memory r = token.symbol();
        string memory n = 'JHX';
        assertEq(n,r);
    }
```

```assembly
#define macro SYMBOL() = takes (0) returns (0) {
    0x20 0x00 mstore
    [META_SYMBOL_LENGTH] 0x20 mstore
    [META_SYMBOL] 0x40 mstore
    0x60 0x00 return
}
```

mds1:
@henry-hz You can se [`vm.breakpoint`](https://book.getfoundry.sh/cheatcodes/breakpoint#breakpoint)

Inphi:
To add to this, it'll be awesome to have a vim-styled "command mode" that supercedes navigation shortcuts. Though it'll be nice to have a full-blown lua scripting engine to interface with during debugging, it'll be a huge improvement to simply have basic commands;
- `:mem <pos>`, `mems <slot>` given either a slot position or index, print or snap to and highlight that specific memory location
- `:store <addr>` similar operations to inspect specific storage slots
[...]

PR #15218, opened 2026-06-18
feat(debugger): show storage accesses

figtracer:
## Motivation

Progresses #927. The debugger has memory/stack/source views, but no immediate indication when the current opcode touches contract storage. That makes storage-related debugging require manually interpreting SLOAD/SSTORE steps and stack values.

## Solution

- Record storage diffs/accesses in `TraceMode::Debug` while preserving full debugger memory/stack/returndata snapshots.
- Avoid downgrading `Debug` to the lightweight `RecordStateDiff` mode when high verbosity or explicit state changes are requested.
- Show the current `SLOAD`/`SSTORE` slot/value in the debugger Variables pane.
- Add unit coverage for trace config and storage access line formatting, plus a debugger dump integration test that verifies both SSTORE and SLOAD entries.

### DIFF
--- a/crates/debugger/src/tui/draw.rs
+++ b/crates/debugger/src/tui/draw.rs
@@ -524,11 +524,17 @@ impl TUIContext<'_> {
         }
 
         // `revm-inspectors` records `storage_change` from journal entries, so warm `SLOAD`s do not
-        // get one. Debug traces already include the stack before execution and pushed stack values
-        // after execution, which is enough to display the accessed slot and loaded value.
+        // get one. Debug mode records full stack snapshots before each opcode; the following step's
+        // stack contains the loaded value after this `SLOAD` executes.
         if step.op.get() == opcode::SLOAD {
             let key = step.stack.as_deref()?.last().copied()?;
-            let value = step.push_stack.as_deref()?.last().copied()?;
+            let value = self
+                .debug_steps()
+                .get(self.current_step.checked_add(1)?)?
+                .stack
+                .as_deref()?
+                .last()
+                .copied()?;
             return Some(sload_storage_access_line(key, value));
         }
 
@@ -1479,12 +1485,12 @@ mod tests {
     }
 
     #[test]
-    fn current_storage_access_line_falls_back_for_warm_sload() {
+    fn current_storage_access_line_uses_next_stack_snapshot_for_warm_sload() {
         let mut step = trace_step(vec![U256::from(1)]);
         step.op = OpCode::SLOAD;
-        step.push_stack = Some(vec![U256::from(42)].into_boxed_slice());
         step.storage_change = None;
-        let mut context = context_with_arena(vec![debug_node(0, 0, vec![step])]);
+        let next_step = trace_step(vec![U256::from(42)]);
+        let mut context = context_with_arena(vec![debug_node(0, 0, vec![step, next_step])]);
         let tui = TUIContext::new(&mut context);
 
         assert_eq!(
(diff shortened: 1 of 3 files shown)


crates/debugger/src/tui/draw.rs
```diff
@@ -486,27 +487,39 @@ impl TUIContext<'_> {
...
+        }
 
         let block = Block::default().title(title).borders(Borders::ALL);
         let paragraph = Paragraph::new(text).block(block).wrap(Wrap { trim: true });
         f.render_widget(paragraph, area);
     }
 
+    fn current_storage_access_line(&self) -> Option<Line<'static>> {
+        self.current_step().storage_change.as_deref().map(storage_access_line)
+    }
```
mablr, line 541: This only shows `storage_change`, but `revm-inspectors` populates that from journal entries, so warm `SLOAD`s have no entry and won’t be shown even though the current opcode touches storage. For example, the new test’s `loaded = value` read after `value = 42` can be missed while the assertion still passes via the later cold `untouched` load. Consider deriving `SLOAD` display from the opcode/stack when `storage_change` is absent, and add a repeated/warm SLOAD regression.

crates/debugger/src/tui/draw.rs
```diff
@@ -486,27 +488,53 @@ impl TUIContext<'_> {
...
+            return Some(storage_access_line(change));
+        }
+
+        // `revm-inspectors` records `storage_change` from journal entries, so warm `SLOAD`s do not
+        // get one. Debug traces already include the stack before execution and pushed stack values
+        // after execution, which is enough to display the accessed slot and loaded value.
+        if step.op.get() == opcode::SLOAD {
+            let key = step.stack.as_deref()?.last().copied()?;
+            let value = step.push_stack.as_deref()?.last().copied()?;
+            return Some(sload_storage_access_line(key, value));
```
mablr, line 537: `push_stack` is never populated here. Debug mode's `into_config` sets `StackSnapshotType::Full`, but revm-inspectors only fills `push_stack` for `All`/`Pushes`. So `step.push_stack.as_deref()?` is always `None` in real debug traces, this `?` always short-circuits, and warm SLOADs never render, the feature
added in 976936463 doesn't work outside the unit test (which fakes `push_stack` directly).

crates/forge/tests/cli/test_cmd/mod.rs
```diff
@@ -2757,6 +2781,46 @@ contract DebugVarsTest {
...
+
+    let dump: serde_json::Value =
+        serde_json::from_str(&std::fs::read_to_string(dump_path).unwrap()).unwrap();
+    let mut storage_changes = Vec::new();
+    collect_debug_dump_storage_changes(&dump, &mut storage_changes);
+
+    assert!(
+        storage_changes.iter().any(|change| change["reason"] == "SSTORE"),
+        "debugger dump should include an SSTORE storage change: {storage_changes:#?}"
+    );
```
mablr, line 2871: This assertion was already true before `loadedAgain = value` was added, the cold `coldLoaded = untouched` SLOAD and the `value = 42` SSTORE satisfy it on their own. The warm re-read this test was meant to cover isn't actually checked, and `--dump` never goes through `current_storage_access_line` anyway (TUI-only), so this can't catch finding above.

mablr (dismissed):

stevencartavia (commented):
lgtm, just needs the conflict

stevencartavia (approved):

mablr (approved):

PR #15219, opened 2026-06-18
feat(common): infer Tempo fee token from tx request inspection

mablr:
## Motivation

Close OSS-165 (minimal scope)

Bring full 5-step fee-token preference resolution logic, see: https://docs.tempo.xyz/protocol/fees/spec-fee#fee-token-preferences

## Limitations
- on-chain/tx-req inferred fee token in assumed in USD
- Pre-T7 `ITIP20::distributeReward` is ignored to avoid hardfork resolution (keep it simple)


crates/common/src/tempo/mod.rs
```diff
@@ -125,6 +127,79 @@ where
...
+where
+    N: Network,
+    N::TransactionRequest: FoundryTransactionBuilder<N>,
+{
+    let calls = tx.tempo_calls();
+    if calls.is_empty() || !calls.iter().all(|(_, input)| is_tip20_fee_token_call(input)) {
+        return None;
+    }
+
+    let targets = call_targets(&calls)?;
```
figtracer: this should check for one common target, not one call. a tempo batch with two valid calls to the same tip-20 currently returns `none`; the new `tip20_batch_infers_only_when_calls_match_sender_and_token` test fails locally.

crates/common/src/tempo/mod.rs
```diff
@@ -125,6 +127,79 @@ where
...
+    if !tx.has_tempo_call_list() {
+        return (targets.len() == 1).then_some(targets[0]);
+    }
+
+    if fee_payer != tx.from() || targets.len() != 1 {
```
figtracer: this check also needs to cover single-call sponsored tempo txs. a sponsored `to + input` request becomes aa when signed, and node-side tip-20 inference is disabled when `fee_payer != tx.origin`; stamping the called token here changes the sponsor digest.

crates/common/src/tempo/mod.rs
```diff
@@ -87,19 +89,19 @@ where
...
+    let stored_fee_token = if let (Some(provider), Some(fee_payer)) = (provider, fee_payer) {
         stored_user_fee_token(provider, fee_payer).await?
     } else {
         None
     };
-    let Some(fee_token) = fee_token else {
-        return Ok(None);
-    };
+
+    let fee_token = stored_fee_token
```
figtracer: node resolution handles a non-aa self-paid `FeeManager.setUserToken(token)` call before the stored preference, so the new token pays for that same tx. this lookup currently returns the old stored token instead.

crates/common/src/tempo/mod.rs
```diff
@@ -125,6 +127,79 @@ where
...
+        selector == ITIP20::transferCall::SELECTOR
+            || selector == ITIP20::transferWithMemoCall::SELECTOR
+            || selector == ITIP20::distributeRewardCall::SELECTOR
+    })
+}
+
+fn decode_stablecoin_dex_fee_token(input: &[u8]) -> Option<Address> {
+    let selector = input_selector(input)?;
+    if selector == IStablecoinDEX::swapExactAmountInCall::SELECTOR {
+        IStablecoinDEX::swapExactAmountInCall::abi_decode(input).ok().map(|call| call.tokenIn)
```
figtracer, line 211: before stamping `tokenIn`, this needs to validate it is a usd tip-20. otherwise invalid dex calldata becomes an explicit fee token instead of falling through to the node default.

crates/common/src/tempo/mod.rs
```diff
@@ -125,6 +127,79 @@ where
...
+            TxKind::Create => None,
+        })
+        .collect()
+}
+
+fn is_tip20_fee_token_call(input: &[u8]) -> bool {
+    input_selector(input).is_some_and(|selector| {
+        selector == ITIP20::transferCall::SELECTOR
+            || selector == ITIP20::transferWithMemoCall::SELECTOR
+            || selector == ITIP20::distributeRewardCall::SELECTOR
```
figtracer: the pinned tempo resolver only treats `distributeReward` as a fee-inference selector before t7. this client-side inference needs the same hardfork gate, or it can stamp a token the node would not infer.

crates/common/src/transactions/builder.rs
```diff
@@ -407,6 +417,22 @@ impl FoundryTransactionBuilder<TempoNetwork> for <TempoNetwork as Network>::Tran
         self.fee_token
     }
 
+    fn tempo_calls(&self) -> Vec<(TxKind, &[u8])> {
+        if !self.calls.is_empty() {
```
figtracer: this should mirror `build_aa()`: that builder appends `inner.to` even when `calls` is non-empty. ignoring it here can infer from a different call list than the tx that gets signed.

crates/common/src/tempo/mod.rs
```diff
@@ -87,19 +92,36 @@ where
...
         stored_user_fee_token(provider, fee_payer).await?
     } else {
         None
     };
-    let Some(fee_token) = fee_token else {
-        return Ok(None);
+    let inferred_fee_token =
+        if immediate_user_token.is_none() && stored_fee_token.is_none() && !calls.is_empty() {
+            let hardfork =
+                active_tempo_hardfork(provider).await.unwrap_or_else(latest_active_tempo_hardfork);
```
figtracer: this fallback uses the local wall-clock fork when the node cannot report its active fork. anvil can run tempo at an explicit older fork, so a t7 node could still get pre-t7 distribute reward inference here. can we fall back to node info or skip hardfork-sensitive inference when the fork is unknown?

crates/common/src/tempo/mod.rs
```diff
@@ -125,6 +147,130 @@ where
...
+
+async fn is_usd_tip20_fee_token<N>(provider: Option<&dyn Provider<N>>, fee_token: Address) -> bool
+where
+    N: Network,
+    N::TransactionRequest: Default + NetworkTransactionBuilder<N>,
+{
+    if !fee_token.is_tip20() {
+        return false;
+    }
+    if known_fee_token_symbol(fee_token).is_some() {
```
figtracer: this trusts foundry-local fee-token addresses on every tempo chain. on remote nodes those addresses can be undeployed or non-usd, so we can stamp a fee_token where the node would fall back. can we validate them through the provider instead?

figtracer (commented):
posting the two related findings from the review.

crates/common/src/tempo/mod.rs
mablr: I think it's fine as is, by definition `known_fee_token_symbol` takes this assumption and is used in multiple contexts.

crates/common/src/tempo/mod.rs
mablr: I simply removed the logic https://github.com/foundry-rs/foundry/pull/15219/commits/ab19c8b55c3c447e38b992780b445fe25b074d48

crates/common/src/tempo/mod.rs
mablr: I think we shouldn't over-engineer this, all this fee-token logic should be additive UX sugar, tempo node will return clear error in this case.

figtracer (dismissed):

0xKarl98 (dismissed):

mattsse (dismissed):

crates/common/src/tempo/mod.rs
```diff
@@ -125,6 +139,82 @@ where
...
+) -> Option<Address> {
+    if calls.is_empty() || !calls.iter().all(|(_, input)| is_tip20_fee_token_call(input)) {
+        return None;
+    }
+
+    let target = common_call_target(calls)?;
+    if fee_payer != tx_from {
+        return None;
+    }
+    Some(target)
```
stevencartavia: Missing `is_tip20()` check on the inferred token. The protocol only infers a transfer **callee** / DEX `tokenIn` after `is_valid_fee_token`, whose first gate is the TIP-20 prefix. Here we return it without that and then `set_fee_token`, so a TIP20-transfer-shaped call whose callee is non-TIP20 (or a swap with a non-TIP20 `tokenIn`) gets stamped as `fee_token`, and the node then rejects with `FeeTokenNotTip20`. If `feeToken` were left unset, the protocol would fall back to the stored token or pathUSD instead of rejecting.

Reachable on Tempo tx paths (access key / session / `--tempo.*`) with no stored preference. For transfer inference, self-paid (`fee_payer == from`); the DEX case doesn't even need that. The prefix check is local (no RPC) and we already do it on the `setUserToken` branch.

Repro (currently returns `Some(erc20)`, should be `None`):
```rust
#[tokio::test]
async fn non_tip20_transfer_is_not_inferred() -> eyre::Result<()> {
    let sender = Address::repeat_byte(0x11);
    let erc20 = Address::repeat_byte(0xab); // not a 0x20C0.. TIP20 address
    let mut tx = TempoTransactionRequest {
        inner: TransactionRequest::default()
            .with_from(sender)
            .with_to(erc20)
            .with_input(
                ITIP20::transferCall { to: Address::repeat_byte(0x01), amount: U256::from(1) }
                    .abi_encode(),
            ),
        ..Default::default()
    };
    assert_eq!(
        resolve_and_set_fee_token::<TempoNetwork>(
            None, Some(Chain::from_named(NamedChain::Tempo)), &mut tx, Some(sender),
        ).await?,
        None, // currently returns Some(erc20)
    );
    Ok(())
}
```
Same for `IStablecoinDEX::swapExactAmountIn/Out` with a non-TIP20 `tokenIn`. Checked against the real `tempo-revm::get_fee_token`: the protocol returns pathUSD for both.

crates/common/src/transactions/builder.rs
```diff
@@ -407,6 +417,20 @@ impl FoundryTransactionBuilder<TempoNetwork> for <TempoNetwork as Network>::Tran
...
+            .iter()
+            .map(|call| (call.to, call.input.as_ref()))
+            .chain(self.inner.to.map(|to| {
+                (to, self.inner.input.input().map_or(&[] as &[u8], |input| input.as_ref()))
+            }))
+            .collect()
+    }
+
+    fn has_tempo_call_list(&self) -> bool {
+        !self.calls.is_empty()
```
stevencartavia, line 436: `has_tempo_call_list()` (`!calls.is_empty()`) isn't equivalent to the protocol's `is_aa()`. A request is AA whenever any AA field is set (`nonce_key`, `valid_before/after`, key/auth fields, ...), even with an empty `calls` list, and `output_tx_type()` returns `AA` in those cases.

The protocol only does the immediate `setUserToken` inference for non-AA txs, so a self-paid AA-by-`nonce_key` `setUserToken(X)` send infers `X` here, while the protocol uses the stored token if present, otherwise pathUSD (pathUSD in the repro below). Gating on a real AA predicate (e.g. `output_tx_type() == AA`) would match the protocol.

Repro (currently returns `Some(ALPHA_USD_ADDRESS)`, should be `None`):
```rust
#[tokio::test]
async fn aa_nonce_key_setusertoken_is_not_inferred() -> eyre::Result<()> {
    let sender = Address::repeat_byte(0x11);
    let mut tx = TempoTransactionRequest {
        inner: TransactionRequest::default()
            .with_from(sender)
            .with_to(TIP_FEE_MANAGER_ADDRESS)
            .with_input(IFeeManager::setUserTokenCall { token: ALPHA_USD_ADDRESS }.abi_encode()),
        nonce_key: Some(U256::from(7)), // makes the tx AA, but calls list stays empty
        ..Default::default()
    };
    assert_eq!(
        resolve_and_set_fee_token::<TempoNetwork>(
            None, Some(Chain::from_named(NamedChain::Tempo)), &mut tx, Some(sender),
        ).await?,
        None, // currently returns Some(ALPHA_USD_ADDRESS)
    );
    Ok(())
}
```

PR #15220, opened 2026-06-18
feat(fuzz): seed invariant corpus from observed calls

0xKarl98:
Towards OSS-189 , inspired by [#9](https://github.com/0xalpharush/foundry/pull/9)

The fuzzer now records replayable subcalls, returns them through the raw call result, and hoists them into the invariant corpus when they produce new coverage. It also seeds the corpus from sibling zero-input test traces, so existing hand-written test flows can help guide invariant fuzzing.

A couple of edge cases are handled too: short calldata no longer panics selector checks, and parent warp/roll delay is preserved on the first retained observed call.

### DIFF
--- a/crates/evm/evm/src/executors/corpus.rs
+++ b/crates/evm/evm/src/executors/corpus.rs
@@ -41,14 +41,16 @@ use crate::{
 };
 use alloy_dyn_abi::JsonAbiExt;
 use alloy_json_abi::Function;
-use alloy_primitives::{Address, Bytes, I256};
+use alloy_primitives::{Address, Bytes, I256, U256};
 use eyre::{Result, eyre};
-use foundry_common::{ContractsByAddress, ContractsByArtifact, sh_warn};
+use foundry_common::{ContractsByAddress, ContractsByArtifact, TestFunctionExt, sh_warn};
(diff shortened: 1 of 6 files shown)


crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1124,6 +1146,123 @@ impl WorkerCorpus {
...
+            let raw = match exec.call_raw(CALLER, invariant_contract.address, calldata, U256::ZERO)
+            {
+                Ok(raw) => raw,
+                Err(_) => continue,
+            };
+            if raw.reverted {
+                continue;
+            }
+
+            let observed = exec
```
figtracer: this drains the wrong buffer: `call_raw` runs on a cloned inspector and returns observed calls in `raw.observed_calls`, so draining `exec.inspector_mut().fuzzer` here is empty and test-trace seeding becomes a no-op

0xKarl98: Have fixed it , plz check again

figtracer (approved):
lgtm we dont need bench for this

mattsse (approved):

PR #15225, opened 2026-06-18
fix(traces): decode calldata bytes in internal traces

### LINKED ISSUE #15224: `forge test --decode-internal` incorrectly decodes `bytes32` argument as calldata offset when internal function also takes `bytes calldata`
### Component

Forge

- [x] Foundry
- [x] Foundryup

1.7.2-nightly

1.8.6

forge test --decode-internal

### Operating System

macOS (Apple Silicon)

## Summary
`forge test --decode-internal` appears to incorrectly decode an internal function argument when the internal call includes both a `bytes32` argument and a `bytes calldata` argument.

The contract executes correctly, but the decoded trace displays a calldata offset as the `bytes32` argument.

## Reproduction
Create a Foundry test file with the following contents:

```solidity
pragma solidity ^0.8.15;

import {Test, console2 as console} from "@forge-std/src/Test.sol";

/// @notice Minimal reproduction for a Foundry `--decode-internal` trace issue.
///
/// Suggested command:
///
/// forge test --match-path src/test/poc/FoundryDecodeInternalPoC.t.sol -vvvv --decode-internal
///
/// Expected behavior:
///
/// The trace should print the same bytes32 value for `_internalValidate(...)` that is logged as
/// `expected digest`.
///
/// The trace prints a calldata offset, e.g. `0x...00a4`, as the first argument to
/// `_internalValidate(...)`, even though the function body receives the correct digest.
contract FoundryDecodeInternalPoC is Test {
    function test_decodeInternalShowsCalldataOffsetInsteadOfBytes32() public {
        DecodeInternalTarget target = new DecodeInternalTarget();

        bytes[] memory wrappedSignatures = new bytes[](1);
        wrappedSignatures[0] = hex"11223344556677889900";

        bytes32 expectedDigest = keccak256("digest that should appear in the internal trace");
        console.logBytes32(expectedDigest);

        bool ok = target.outer(wrappedSignatures, expectedDigest);
        assertTrue(ok);
    }
}

contract DecodeInternalTarget {
    function outer(bytes[] calldata signatures, bytes32 digest) external pure returns (bool) {
        bytes calldata signature = signatures[0];

        // This internal call receives `digest` as its first argument. In the failing trace,
        // `--decode-internal` displays a calldata offset instead.
        return _internalValidate(digest, signature);
    }

    function _internalValidate(bytes32 digest, bytes calldata signature)
        internal
        pure
        returns (bool)
    {
        // Force the optimizer to keep both arguments live.
        return digest == keccak256("digest that should appear in the internal trace")
            && signature.length == 10;
    }
}
```

Then run:

```bash
forge test --match-contract FoundryDecodeInternalPoC -vvvv --decode-internal
```

## Expected Behavior
The decoded internal call should display the actual `bytes32` value passed from the external function into the internal function.

For example, the external call is decoded correctly:

```text
DecodeInternalTarget::outer(
  [0x11223344556677889900],   0x9a707a44f6e6038d2b03f3947e82f6403612c5d5b923868f6a759e9c38e16a28
)
```

The internal call should therefore show the same first argument:

```text
DecodeInternalTarget::_internalValidate(
  0x9a707a44f6e6038d2b03f3947e82f6403612c5d5b923868f6a759e9c38e16a28,
  ...
)
```

## Actual Behavior
On the current nightly, the trace displays a calldata offset as the first `bytes32` argument:

```text
DecodeInternalTarget::outer([0x11223344556677889900], 0x9a707a44f6e6038d2b03f3947e82f6403612c5d5b923868f6a759e9c38e16a28)
[staticcall]
  ├─ DecodeInternalTarget::_internalValidate(bytes32,bytes)(
  │   0x00000000000000000000000000000000000000000000000000000000000000a4,
  │   <unknown>
  │ )
  │   └─ ← true
```

The test still passes, proving that the contract logic receives the correct value:

```solidity
return digest == keccak256("digest that should appear in the internal trace")
    && signature.length == 10;
```

So this appears to be a trace-decoding/display issue only.

## Notes
This originally came up while debugging an ERC-1271 verification trace. The decoded trace appeared to show
`isValidSignature(bytes32,bytes)` receiving `0x204` as the digest, but the subsequent `ecrecover` precompile received the correct digest.

The minimal reproduction above removes ERC-1271, delegatecall, proxies, and signature recovery entirely. The issue still reproduces with a simple external function forwarding a `bytes32` and a `bytes calldata` slice into an internal function.

The incorrect value appears to be a calldata offset from the dynamic argument layout, not the actual `bytes32` argument.

figtracer:
## Summary

Fixes #15224.

- account for calldata reference parameters that occupy multiple stack words when decoding internal trace arguments
- decode calldata `bytes`/`string` arguments from the call input when the trace step provides offset and length
- pass frame calldata through the Forge trace and debugger internal-parameter decode paths
- add a Forge CLI regression for `bytes32` followed by `bytes calldata`

## Root Cause

`decode_step_parameters` assumed each Solidity parameter maps to one EVM stack word. A `bytes calldata` internal parameter is represented by offset and length, so the decoder selected too small a stack window and shifted the preceding `bytes32` argument onto the calldata offset.

### DIFF
--- a/crates/debugger/src/tui/draw.rs
+++ b/crates/debugger/src/tui/draw.rs
@@ -1249,6 +1249,37 @@ mod tests {
     }
 
     #[test]
+    fn decode_internal_parameter_values_passes_calldata_to_fallback_decoder() {
+        let digest = U256::from(0x1234);
+        let offset = 0x44;
+        let mut calldata = vec![0; offset];
+        calldata.extend_from_slice(&[0x11, 0x22, 0x33]);
+
+        let mut entry_node =
+            debug_node(0, 1, vec![trace_step(vec![digest, U256::from(offset), U256::from(3)])]);
+        entry_node.calldata = Bytes::from(calldata);
+
+        let mut context = context_with_arena(vec![
+            debug_node(0, 0, vec![internal_call_step_without_args(2)]),
+            debug_node(1, 0, vec![trace_step(Vec::new())]),
+            entry_node,
+        ]);
+        let mut tui = TUIContext::new(&mut context);
+        tui.draw_memory.inner_call_index = 2;
+
+        assert_eq!(
+            tui.decode_internal_parameter_values(&scope(
+                "foo",
+                "(bytes32 digest, bytes calldata signature)"
+            )),
+            Some(vec![
+                "0x0000000000000000000000000000000000000000000000000000000000001234".to_string(),
+                "0x112233".to_string(),
+            ])
+        );
+    }
+
+    #[test]
     fn decode_internal_parameter_values_accepts_matching_overload_args() {
         let mut context = context_with_arena(vec![debug_node(
             0,
--- a/crates/debugger/src/tui/draw.rs
+++ b/crates/debugger/src/tui/draw.rs
@@ -17,9 +17,7 @@ use ratatui::{
     text::{Line, Span, Text},
     widgets::{Block, Borders, List, ListItem, ListState, Paragraph, Wrap},
 };
-use revm_inspectors::tracing::types::{
-    CallKind, CallTraceStep, DecodedInternalCall, DecodedTraceStep,
-};
+use revm_inspectors::tracing::types::{CallKind, DecodedInternalCall, DecodedTraceStep};
 use std::{collections::VecDeque, fmt::Write};
 
 impl TUIContext<'_> {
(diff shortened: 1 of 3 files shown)


mablr (approved):
Reproduced locally w/ issue 15224. SGTM.

stevencartavia (approved):

PR #15226, opened 2026-06-18
feat(config): make isolate the default

mablr:
## Motivation

Close OSS-94 by making isolate mode the default.

Unlocked by #14493.

This is a **breaking** change for gas snapshots and any tests/scripts that assert exact trace output, including gas costs and CREATE-derived deployment addresses. It must be mentioned explicitly in the next release changelog.


.github/workflows/test-flaky.yml
```diff
@@ -84,7 +84,7 @@ jobs:
           ETHERSCAN_KEY: ${{ secrets.ETHERSCAN_API_KEY }}
           ARBITRUM_RPC: ${{ secrets.ARBITRUM_RPC }}
           ETH_SEPOLIA_RPC: ${{ secrets.ETH_SEPOLIA_RPC }}
-        run: cargo nextest run --locked --profile flaky --features=isolate-by-default --no-fail-fast
+        run: cargo nextest run --locked --profile flaky --no-fail-fast
```
stevencartavia: Now that `isolate-by-default` is gone and isolate is the default, the dedicated isolate jobs no longer exercise a distinct mode: `test-isolate.yml` is just the Linux x86/arm subset of the default `ci.yml` test workflow, and the two `test-flaky.yml` jobs run the identical `cargo nextest run --locked --profile flaky --no-fail-fast`. Could we drop the duplicate jobs, or repurpose one to explicitly cover non-isolate behavior (e.g. a `forge test --no-isolate` / non-isolate config override)?

crates/forge/tests/cli/inline_config.rs
```diff
@@ -315,17 +314,11 @@ Ran 2 test suites [ELAPSED]: 3 tests passed, 0 failed, 0 skipped (3 total tests)
...
-    //     test_non_isolated_function: 27722,
-    // }
-
-    // ContractConfig {
-    //     test_isolated_contract: 48926,
-    // }
-
-    assert!(function_config.test_isolated_function > function_config.test_non_isolated_function);
-    assert_eq!(function_config.test_isolated_function, contract_config.test_isolated_contract);
+    assert!(function_config.test_isolated_function > 0);
```
stevencartavia: This test still records both isolated and non-isolated gas, so changing `isolated > non_isolated` to `isolated > 0` drops the actual regression check. Can we keep the isolated-vs-non-isolated comparison?

.github/workflows/benchmarks.yml
```diff
@@ -103,19 +97,6 @@ jobs:
             --output-file forge_test_bench.md \
             --verbose
 
-      - name: Run forge isolate test benchmarks
```
figtracer, line 106: removing the isolate benchmark makes the `forge_test` series compare stable's old non-isolated default with local's new isolated default. can we keep an explicit isolate series, or run this series with `--no-isolate`, so the benchmark mode stays comparable?

mablr: good catch, ty

mablr: https://github.com/foundry-rs/foundry/pull/15226/commits/ee424320799ff0adaf8773538f813eae736931e5

figtracer (approved):

mattsse (approved):
LGTM

PR #15227, opened 2026-06-18
feat(forge): add fuzz run, replay, and show commands

mablr:
## Summary

Part of OSS-297

Adds the first scoped MVP for `forge fuzz`:

- `forge fuzz run`: runs only fuzz and invariant tests
- `forge fuzz replay`: replays persisted fuzz failures, or corpus entries with `--corpus-dir`
- `forge fuzz show`: prints persisted corpus entries

## Follow-ups

This PR intentionally keeps the scope narrow after splitting the original larger implementation. Corpus minimization commands such as `cmin` and `tmin` are deferred to follow-up PRs.

These UX/corpus caveats are also left for follow-up work: `--list` may show non-fuzz tests that execution skips, exact bare-name `--match-test` filters like `^testClamp$` do not match fuzz signatures, hand-crafted corpus files with the same UUID but different timestamps currently dedup to one entry.


crates/forge/src/runner.rs
```diff
@@ -2138,6 +2262,7 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
             trial: showmap.trial.clone(),
```
figtracer, line 3323: `showmap.trial` is used as a raw path component, and the writer later uses `file::create`, so `--showmap-trial ../../victim` overwrote a file outside `--showmap-out` in my local replay. `showmap.approach` has the same issue. please validate these as names, reject separators/absolute components, and avoid truncating existing files.

crates/forge/src/cmd/fuzz.rs
```diff
@@ -0,0 +1,666 @@
...
+    #[arg(value_name = "CORPUS_DIR", value_hint = ValueHint::DirPath)]
+    corpus_dir: PathBuf,
+    /// Output corpus directory.
+    #[arg(long, short, value_name = "DIR", value_hint = ValueHint::DirPath)]
+    out: PathBuf,
+}
+
+impl FuzzCminArgs {
+    async fn run(self) -> Result<()> {
+        if self.out.exists() {
```
figtracer: `cmin` creates `--out` before replay succeeds. when replay later fails, e.g. a bad `--mt`, it leaves the output dir behind and the corrected rerun is blocked by `output corpus directory already exists`. please write to a temp dir and rename on success, or clean the directory created by this invocation on error.

figtracer (commented):
two issues found while replaying the same corpus across temp projects and checking overwrite behavior.

crates/evm/evm/src/executors/showmap.rs
```diff
@@ -211,6 +234,112 @@ pub fn replay_corpus_to_showmap<FEN: FoundryEvmNetwork>(
...
+            let success = if call_result
+                .reverter
+                .is_some_and(|reverter| reverter != target && reverter != CHEATCODE_ADDRESS)
+            {
+                true
+            } else {
+                executor.is_raw_call_mut_success(target, &mut call_result, false)
+            };
+            if !success && observation.failure.is_none() {
+                observation.failure = Some(format!(
```
figtracer: `tmin` can preserve a different fuzz failure here because the identity only includes target, selector, reverter, and exit reason. i minimized a one-call corpus for `testFuzz_reason(2)` where value `2` reverts `"two"`; the minimizer changed calldata to value `0`, which reverts `"zero"`, while still printing `preserved original failure`. please include revert data/reason in the identity, or otherwise compare the actual failing outcome before accepting a candidate.

figtracer (commented):
two more

crates/evm/evm/src/executors/corpus_io.rs
```diff
@@ -80,10 +83,95 @@ pub fn read_corpus_dir(path: &Path) -> impl Iterator<Item = CorpusDirEntry> {
     .into_iter()
 }
 
+/// Reads corpus files from a file, corpus directory, worker corpus directory, or generated corpus
+/// root such as `<root>/<contract>/<test>/worker0/corpus`.
+pub fn read_corpus_tree(path: &Path) -> Result<Vec<CorpusDirEntry>> {
+    if path.is_file() {
+        let name = path.file_name().and_then(|name| name.to_str()).unwrap_or_default();
+        let (uuid, timestamp) = parse_corpus_filename(name)
```
figtracer: `fuzz tmin --out min.json` writes a valid minimized sequence, but `fuzz show min.json` and `fuzz replay --corpus-dir min.json` both fail here because the single-file path must parse as `<uuid>-<ts>.json`. since the cli/tests allow arbitrary `--out` filenames, the minimized file should be consumable by the other corpus subcommands; either relax single-file parsing or require/write corpus-style output names.

crates/evm/evm/src/executors/showmap.rs
```diff
@@ -169,17 +217,65 @@ pub fn replay_corpus_to_showmap<FEN: FoundryEvmNetwork>(
 
             register_replay_created(
                 &call_result.state_changeset,
-                dynamic,
-                fuzzed_contracts,
+                target.dynamic,
+                target.fuzzed_contracts,
```
figtracer, line 261: this misses invariant `fail_on_revert`: plain handler reverts are failures in normal runs, but replay commits and continues. please carry that setting into replay/minimization.

crates/evm/evm/src/executors/showmap.rs
```diff
@@ -169,17 +217,65 @@ pub fn replay_corpus_to_showmap<FEN: FoundryEvmNetwork>(
...
+                            .calldata
+                            .get(..4)
+                            .map(Selector::from_slice)
+                            .unwrap_or_default(),
+                        call_result.reverter,
+                        did_fail_on_assert(&call_result, &call_result.state_changeset),
+                        &call_result,
+                    )
+                {
+                    rollback_replay_created(target.fuzzed_contracts, created);
```
figtracer: this does not match normal invariant replay for `check_interval` or `afterInvariant`. please carry those settings or reuse the existing sequence-check logic.

crates/forge/src/runner.rs
```diff
@@ -2661,17 +2820,11 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
             per_input: showmap.per_input,
             domain,
+            emit_files: showmap.emit_files,
         };
 
         let start = std::time::Instant::now();
```
figtracer, line 3329: `forge fuzz replay --corpus-dir` can pass with `replay: 0 entries` when every input is skipped. please fail or skip when no corpus entries actually ran.

crates/forge/src/cmd/fuzz.rs
```diff
@@ -0,0 +1,696 @@
...
+            CorpusShowFormat::Json => sh_println!("{}", serde_json::to_string_pretty(&entries)?)?,
+        }
+        Ok(())
+    }
+}
+
+/// Minimize a corpus by keeping entries that contribute new coverage.
+#[derive(Clone, Debug, Parser)]
+pub struct FuzzCminArgs {
+    #[command(flatten)]
```
figtracer: `cmin`/`tmin` only accept filters, so users cannot pass replay-critical options like `--fork-url` or compiler overrides. please expose the needed test args here.

figtracer (commented):
some nits

crates/evm/evm/src/executors/showmap.rs
```diff
@@ -144,13 +270,19 @@ pub fn replay_corpus_to_showmap<FEN: FoundryEvmNetwork>(
...
+        let fail_on_revert = target.invariant_fns.iter().any(|(_, fail_on_revert)| *fail_on_revert);
+        // Number of committed (non-`vm.assume`) calls, used to gate invariant checks.
+        let mut accepted = 0usize;
+        let mut entry_failure: Option<ReplayFailure> = None;
         for tx in &tx_seq {
-            if !WorkerCorpus::can_replay_tx(tx, fuzzed_function, fuzzed_contracts) {
+            if !WorkerCorpus::can_replay_tx(tx, target.fuzzed_function, target.fuzzed_contracts) {
                 continue;
             }
             had_replayable = true;
```
figtracer: this counts a matching tx before the magic assume discard, so an all-assume corpus can pass as replayed. can this track accepted txs instead?

crates/forge/tests/cli/test_cmd/fuzz.rs
```diff
@@ -57,6 +104,889 @@ contract FuzzerDictTest is Test {
...
+    function enableChecks() external {
+        checksEnabled = true;
+    }
+
+    function setOk(bool value) external {
+        ok = value;
+    }
+
+    function armAfterInvariant(uint256 value) external {
+        if (value == 0xdeadbeef) {
```
figtracer, line 674: this value can be generated during the seed run above, which fails setup before replay is tested. please seed the run or add the failing selector after corpus generation.

crates/forge/src/runner.rs
```diff
@@ -2821,12 +2862,19 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
                         func.name,
                     );
                 }
-                self.result.replay_result(
-                    stats.corpus_entries,
-                    stats.showmap_files,
-                    stats.skipped_entries,
-                    duration,
-                );
+                if !showmap.emit_files && stats.corpus_entries == 0 {
```
figtracer: when unreadable entries are the only entries, this becomes a successful skip. malformed corpora should fail instead of exiting 0.

crates/evm/evm/src/executors/corpus_io.rs
```diff
@@ -80,10 +88,145 @@ pub fn read_corpus_dir(path: &Path) -> impl Iterator<Item = CorpusDirEntry> {
...
+            }
+        };
+        for child in children {
+            let Ok(child) =
+                child.inspect_err(|err| debug!(%err, "failed to read corpus tree entry"))
+            else {
+                continue;
+            };
+            let child_path = child.path();
+            if child_path.is_dir() {
```
stevencartavia: **`forge fuzz`/showmap corpus reading follows symlinks and escapes the corpus dir**

`read_corpus_tree` walks dirs recursively with `is_dir()`/`is_file()`, which follow symlinks — no symlink check, no depth/visited limit. A symlink inside an untrusted corpus reads (and `cmin` copies) files from outside the corpus root.

Repro:
```bash
mkdir corpus && ln -s /any/outside/dir corpus/link   # outside dir has <uuid>-<ts>.json files
forge fuzz cmin --mc X --mt testFuzz corpus --corpus-out out   # "kept N/N", copies outside files into out/
forge fuzz replay --corpus-dir corpus                          # replays the outside entries
```
A symlink to `/` also makes it traverse the whole tree (DoS).

Fix: skip symlinks via `DirEntry::file_type()`, add a visited-set + depth/entry caps.

crates/evm/evm/src/executors/corpus_io.rs
```diff
@@ -80,10 +88,145 @@ pub fn read_corpus_dir(path: &Path) -> impl Iterator<Item = CorpusDirEntry> {
...
+        return Err(eyre!("corpus path does not exist or is not readable: {}", path.display()));
+    }
+
+    let mut seen_uuids = HashSet::new();
+    let mut entries = Vec::new();
+    let mut stack = vec![path.to_path_buf()];
+    while let Some(dir) = stack.pop() {
+        for replay_dir in canonical_replay_dirs(&dir) {
+            entries
+                .extend(read_corpus_dir(&replay_dir).filter(|entry| seen_uuids.insert(entry.uuid)));
```
stevencartavia: **Corpus dedup by UUID only drops distinct entries**

`read_corpus_tree` dedups on UUID alone, so two files with the same UUID but different timestamp/content collapse to one.

Repro: two files `<sameuuid>-1.json` / `<sameuuid>-2.json` with different calldata → `replay`/`cmin`/`show` see only 1. (Harmless for generated corpora; affects hand-crafted ones.)

Fix: dedup by `(uuid, timestamp)`.

crates/forge/src/cmd/fuzz.rs
```diff
@@ -0,0 +1,779 @@
...
+    #[arg(long = "corpus-out", value_name = "FILE", value_hint = ValueHint::FilePath)]
+    out: PathBuf,
+    /// Maximum candidate replays to attempt.
+    #[arg(long, default_value_t = 5000, value_name = "N")]
+    max_attempts: usize,
+}
+
+impl FuzzTminArgs {
+    async fn run(self) -> Result<()> {
+        if self.out.exists() {
```
stevencartavia: **`forge fuzz tmin --corpus-out` writes through a dangling symlink**

The `out.exists()` guard is `false` for a dangling symlink, then `File::create` follows it — so output lands outside the intended path. (`cmin` is safe because it stages + renames.)

Repro:
```bash
ln -s /tmp/victim.json.gz out.json.gz
forge fuzz tmin --mc X --mt testFuzz entry.json --corpus-out out.json.gz   # writes /tmp/victim.json.gz
```

Fix: write via `File::create_new` / no-clobber (like `cmin`). (A `symlink_metadata` pre-check alone only closes the non-racy case — still TOCTOU.)

crates/forge/src/cmd/fuzz.rs
```diff
@@ -0,0 +1,779 @@
...
+            // existed but did not target the matched test (filter mismatch / `vm.assume`).
+            let mismatched = skipped.saturating_sub(unreadable);
+            if unreadable == total {
+                bail!(
+                    "replayed 0 transactions from {corpus}; all {unreadable} corpus entries could not be read"
+                );
+            }
+            if mismatched > 0 {
+                bail!(
+                    "replayed 0 transactions from {corpus}; {mismatched} transactions did not match \
```
stevencartavia: **`cmin` gives a wrong error for `vm.assume`-rejected entries (and disagrees with `tmin`)**

An entry that produces coverage but is `vm.assume`-rejected makes `cmin` abort with “did not match the test — check that --mc/--mt match” — but the filter *did* match. `tmin` accepts the same entry.

Repro: `testFuzz_g(uint256 x){ sink=x+1; vm.assume(x==999999); }`, corpus entry `x=7` → `cmin` errors, `tmin` succeeds.

Fix: either accept coverage-producing assume-rejects in `cmin` (consistent with `tmin`), or count assume-rejects separately and emit an accurate message.

crates/forge/src/args.rs
```diff
@@ -98,6 +98,10 @@ pub fn run_command(args: Forge) -> Result<()> {
                 outcome.ensure_ok(silent)
             }
         }
+        ForgeSubcommand::Fuzz(cmd) => {
+            let outcome = global.block_on(cmd.run())?;
+            outcome.ensure_ok(shell::is_json())
```
stevencartavia: **`forge fuzz run/replay --junit` corrupts JUnit output on failures**

This arm uses `ensure_ok(shell::is_json())`, unlike the `Test` arm which uses `cmd.junit || shell::is_json()`. With `--junit`, a failing run emits the XML and then appends the human “Failing tests:” footer to stdout, corrupting it for JUnit consumers. (`forge test --junit` stays clean.)

Repro: `forge fuzz run --junit --mc <contract-with-failing-fuzz-test>` → XML followed by human failure summary.

Fix: compute `silent = cmd.is_junit() || shell::is_json()` for `run`/`replay` and pass it to `ensure_ok`. (`show/cmin/tmin` don’t expose `--junit`.)

crates/forge/src/args.rs
```diff
@@ -98,6 +98,10 @@ pub fn run_command(args: Forge) -> Result<()> {
                 outcome.ensure_ok(silent)
             }
         }
+        ForgeSubcommand::Fuzz(cmd) => {
```
stevencartavia, line 65: **`forge fuzz run/replay --watch` is accepted but silently ignored**

`TestArgs` flattens `WatchArgs`, so `--watch` parses, but only the `Test` arm dispatches to `watch::watch_test`; this `Fuzz` arm always runs once and exits.

Repro: `forge fuzz run --watch --mc X` exits immediately, whereas `forge test --watch --mc X` blocks and watches.

Fix: either support it (route through `watch_test` with the fuzz-only config) or explicitly reject `--watch` for `forge fuzz`.

crates/evm/evm/src/executors/corpus_io.rs
mablr: will do as follow-up

crates/forge/src/cmd/fuzz.rs
mablr: will do as follow-up

crates/evm/evm/src/executors/showmap.rs
```diff
@@ -213,6 +415,180 @@ pub fn replay_corpus_to_showmap<FEN: FoundryEvmNetwork>(
...
+    target: Address,
+    selector: Selector,
+    assertion_failure: bool,
+    fail_on_revert: bool,
+    call_result: &crate::executors::RawCallResult<FEN>,
+    fingerprint: Option<B256>,
+) -> Option<ReplayFailure> {
+    let failed = assertion_failure
+        || (fail_on_revert && call_result.reverted && call_result.result.as_ref() != MAGIC_ASSUME);
+    failed.then_some(ReplayFailure::Handler { target, selector, fingerprint })
```
stevencartavia: Under invariant `fail_on_revert`, a plain handler revert is recorded as `ReplayFailure::Handler`, which `is_terminal()` treats as non-terminal — so a later terminal invariant break overwrites it, and `tmin` minimizes toward the wrong failure. In a real campaign that revert is terminal (the run stops there).

Repro: corpus `[revertHandler, breakInvariant]` with `fail_on_revert=true`. `forge fuzz replay` fails on the revert; `forge fuzz tmin` drops the revert tx, keeps `breakInvariant`, prints "preserved original failure", but the minimized file now reproduces `invariant broken` — a different failure.

crates/evm/evm/src/executors/showmap.rs
```diff
@@ -167,21 +298,92 @@ pub fn replay_corpus_to_showmap<FEN: FoundryEvmNetwork>(
...
+                "corpus entry {} failed during replay: {failure}",
+                entry.path.display()
+            ));
+        }
+        rollback_replay_created(target.fuzzed_contracts, created);
 
-        if !had_replayable {
+        if !had_accepted {
             stats.skipped_entries += 1;
             continue;
```
stevencartavia, line 348: `evm_buf`/`san_buf` are shared across corpus entries and aren't cleared when an entry is `vm.assume`-discarded, so a discarded entry's coverage leaks into the next per-input file and into the aggregate file.

Repro: same accepted entry produces a different `--showmap-per-input` file depending on whether a `vm.assume`-rejected entry precedes it (extra edge IDs + doubled hitcounts). A corpus where every entry is rejected (0 accepted) still writes a coverage file containing the rejected branch's edges.

crates/forge/src/runner.rs
```diff
@@ -2849,12 +3029,28 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
                 }
-                self.result.replay_result(
-                    stats.corpus_entries,
-                    stats.showmap_files,
-                    stats.skipped_entries,
-                    duration,
-                );
+                if stats.corpus_entries == 0
+                    && stats.unreadable_entries > 0
+                    && stats.unreadable_entries == stats.skipped_entries
```
stevencartavia: The zero-replay failure only fires when `unreadable_entries == skipped_entries`. Since `skipped_entries` also counts unmatched entries, any unmatched entry makes an unreadable/malformed corpus file exit 0 (successful skip) instead of failing.

Repro: a malformed corpus file alone → `forge fuzz replay` fails "failed to read 1 corpus entries". Add one valid-but-unmatched entry → it passes with "[SKIP: replayed 0 corpus entries]". The malformed file is silently hidden.

crates/evm/evm/src/executors/showmap.rs
```diff
@@ -89,13 +102,132 @@ pub struct ShowmapStats {
...
+/// for code paths within a single function, an edge-coverage fingerprint) rather
+/// than on raw revert bytes, so that:
+/// - the same bug reached with different revert arguments is still considered the same failure
+///   (avoids false negatives), and
+/// - distinct assertion sites that produce identical revert data (e.g. two `Panic(0x01)` paths) are
+///   kept distinct (avoids false positives).
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum ReplayFailure {
+    /// A stateless fuzz test call failed. Keyed by selector and code-path fingerprint.
+    Fuzz { selector: Selector, fingerprint: Option<B256> },
```
stevencartavia, line 135: The stateless-fuzz failure identity is `(test selector, edge fingerprint)` — the revert reason/data isn't included. For code that selects its revert reason branchlessly (same edge path), `tmin`'s calldata simplification can swap to a different revert error while reporting "preserved original failure". Branched reverts are fine (different edges), so real-world likelihood is low; flagging since it's the same class of failure-identity weakness as the `fail_on_revert` issue above.

Repro: branchless contract reverting `B(x)` for `x==0` else `A(x)`; corpus `x=0` (error B) → `tmin` outputs `x=MAX` (error A).

mablr: Will handle this in follow-up.

crates/forge/src/runner.rs
```diff
@@ -2993,7 +3036,54 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                        "persisted fuzz failure selector does not match {}",
+                        func.name
+                    )),
+                    ..Default::default()
+                };
+                self.result.fuzz_result(result);
+                return self.result;
+            }
+
+            fuzz_config.runs = 1;
```
figtracer: if the saved case now hits vm.assume, fuzzed executor consumes it without counting a run, so forge fuzz replay can fall through to fresh random input and report an unrelated failure.

mablr: https://github.com/foundry-rs/foundry/pull/15227/commits/04d910ebaffc20ce75f5be99489e9355335c445c

crates/forge/src/runner.rs
```diff
@@ -2549,6 +2581,17 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
             &replay_ctx,
         );
 
+        // `forge fuzz replay` (without `--corpus-dir`) only replays persisted failures and
+        // must never start a fresh campaign. If handler bugs still reproduce, surface them
+        // through the normal invariant result path below; otherwise report a skip.
+        if self.cr.mcr.tcfg.fuzz_failure_replay && persisted_handler_failures.is_empty() {
+            self.result.single_skip(SkipReason(Some(format!(
+                "no persisted invariant failure reproduced for {}",
```
stevencartavia: `forge fuzz replay` skips persisted failures of non-anchor invariants in a merged suite and exits 0.

Non-anchor persisted files are only used to filter/warn — only the anchor's failure is actually replayed, so a persisted failure for a non-anchor invariant is never reproduced.

Repro: 3 invariants where only the middle one breaks → `forge fuzz replay --mc C` prints `[SKIP: no persisted invariant failure reproduced for <anchor>]`, exit 0. `--mt invariant_b_middle` does reproduce it.

The skip message is also misleading (a persisted failure does exist, and is even named in a warning). Suggest replaying each selected invariant as its own anchor in replay mode.

crates/evm/evm/src/executors/fuzz/mod.rs
```diff
@@ -261,6 +261,101 @@ impl<FEN: FoundryEvmNetwork> FuzzedExecutor<FEN> {
...
+            result.gas_report_traces.extend(call.traces.into_iter().map(|trace| trace.arena));
+        } else {
+            let reason = if call.reverter == Some(CHEATCODE_ADDRESS) {
+                SkipReason::decode(&call.result)
+                    .map(|reason| reason.to_string())
+                    .or_else(|| rd.maybe_decode(&call.result, call.exit_reason))
+            } else {
+                rd.maybe_decode(&call.result, call.exit_reason)
+            };
+            result.reason = reason;
```
stevencartavia, line 345: When a replayed fuzz input hits `vm.skip(true)`, `forge fuzz replay` reports it as a failure instead of a skip.

Normal fuzzing classifies a cheatcode skip as skipped, but both `forge fuzz replay` paths currently treat it as a failure.

Persisted replay: `replay_persisted_failure` decodes the skip reason into `result.reason` but never sets `result.skipped`, so it is reported as `[FAIL: skipped; counterexample: ...]`.

`--corpus-dir` replay: `fuzz_replay_call_succeeded` (in `showmap.rs`) handles `vm.assume` rejects but not the cheatcode skip payload, so the entry becomes `ReplayFailure::Fuzz` (`corpus entry ... failed during replay: fuzz call ... failed`).

Repro (both should be exit 0, but give exit 1):
- Persisted: real campaign persists a counterexample, then the test is disabled with `vm.skip(true)`. `forge test` → `[SKIP]`; `forge fuzz replay` → `[FAIL: skipped; counterexample: ...]`.
- Corpus: `if (v % 2 == 0) { vm.skip(true); return; }`, entry with `v=8`. `forge fuzz run` → `[SKIP]`; `forge fuzz replay --corpus-dir` → `[FAIL: corpus replay failed: ...]`.

Expected: treat a cheatcode skip as skipped / not a replay failure, matching the normal fuzz campaign.

crates/forge/src/runner.rs
mablr: https://github.com/foundry-rs/foundry/pull/15227/commits/2fcfa11e786609cde5b14ddc61f6a84390e729e4

crates/evm/evm/src/executors/fuzz/mod.rs
mablr: https://github.com/foundry-rs/foundry/pull/15227/commits/e2ee23f823ef1e2560c2ea9f9bbef080529af986

crates/evm/evm/src/executors/showmap.rs
```diff
@@ -105,52 +175,60 @@ pub struct ShowmapStats {
...
-    fuzzed_function: Option<&Function>,
-    fuzzed_contracts: Option<&FuzzRunIdentifiedContracts>,
-    dynamic: Option<&DynamicTargetCtx<'_>>,
+    target: ShowmapReplayTarget<'_>,
     opts: &ShowmapOpts,
 ) -> Result<ShowmapStats> {
-    let replay_dirs = canonical_replay_dirs(corpus_dir);
-    if !replay_dirs.iter().any(|d| d.is_dir()) {
-        return Err(eyre!("corpus directory not found: {}", corpus_dir.display()));
+    let entries = read_corpus_tree(corpus_dir)?;
```
stevencartavia, line 184: `forge fuzz replay --corpus-dir <generated-root>` replays other contracts' corpus on selector collisions.

`read_corpus_tree` collects every entry under the given root, and stateless fuzz entries are accepted by selector only (`can_replay_tx` checks the first 4 calldata bytes) — the per-contract/per-test path identity (`<root>/<contract>/<test>/worker0/corpus`) is discarded. Pointing `--corpus-dir` at a generated root is accepted by the current read path, so any two contracts sharing a function selector cross-contaminate.

Repro: contracts `A` and `B` both with `testFuzz_v(uint256)` (same selector), real corpora generated under `fuzz_corpus/{A,B}/...` (4 entries each). `forge fuzz replay --mc B --corpus-dir fuzz_corpus/B` → `replay: 4 entries`; `forge fuzz replay --mc B --corpus-dir fuzz_corpus` (whole root) → `replay: 8 entries` (pulls in A's entries). Beyond inflated counts, contaminating entries can change replay pass/fail behavior or report a misleading PASS.

Expected: when `--corpus-dir` points at a generated corpus tree, resolve the same per-test corpus dir the campaign uses before reading: `<root>/<contract>/<test>` for stateless fuzz, `<root>/<contract>` for boolean invariant suites, and `<root>/<contract>/<invariant>` for optimization invariants; fall back to the supplied dir only for flat/manual corpora.

crates/forge/src/runner.rs
```diff
@@ -1545,11 +1546,24 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
             return self.result;
         }
 
-        // In showmap replay mode, only fuzz/invariant tests are runnable.
-        if self.cr.mcr.tcfg.showmap.is_some()
-            && matches!(kind, TestFunctionKind::UnitTest { .. } | TestFunctionKind::TableTest)
+        // In showmap replay mode and `forge fuzz`, only fuzz/invariant tests are runnable.
```
stevencartavia, line 1564: `forge fuzz run`/`replay` runs `setUp` (and can fail) on suites with no fuzz/invariant tests.

Non-fuzz tests are skipped here, per-function inside `FunctionRunner::run`, not at contract/test selection. So a unit-only suite is still selected, its `setUp` runs first (`run_tests`), and a failing `setUp` fails the command — even though the suite has nothing fuzz-related to run.

Repro: a unit-only suite with a failing `setUp` (e.g. `assertEq(x, 999)` or `revert`). `forge fuzz run --mc UnitTest` → `[FAIL: ...] setUp()`, exit 1 (same for `replay`). Project-wide `forge fuzz run` then fails because of a suite with zero fuzz tests. Contrast: normal `forge test` with a filter that matches no tests excludes the contract before setup, so `setUp` never runs.

Expected: treat suites with no runnable fuzz/invariant tests as empty before setup (don't execute `setUp`), like a filter that excludes everything.

crates/evm/evm/src/executors/showmap.rs
mablr: will do in follow-up

crates/forge/src/runner.rs
```diff
@@ -671,6 +672,39 @@ impl<'a, FEN: FoundryEvmNetwork> ContractRunner<'a, FEN> {
         let start = Instant::now();
         let mut warnings = Vec::new();
 
+        // Filter out functions sequentially since it's very fast and there is no need to do it
+        // in parallel. In fuzz-only mode, use this pre-setup list to drop suites with no
+        // runnable fuzz or invariant tests before executing `setUp`.
+        let find_timer = Instant::now();
+        let symbolic_enabled = symbolic_entrypoints_enabled(
```
stevencartavia: Moving this filter block before the setup phase means `symbolic_enabled` is read before `apply_contract_inline_config()` runs (that only happens inside `_setup()`, called later via `self.setup(call_setup)`). So contract-level `/// forge-config: default.symbolic.enabled = true` no longer enables symbolic discovery in plain `forge test` — the `check_*` functions get filtered out here and silently skipped.

Repro vs the parent commit (clean cache): a `check_x(uint256)` in such a contract ran symbolically and failed before this commit (exit 1), but here it's silently dropped and the suite exits 0. Reproduces with both `--mc` and no filter. `--symbolic` still works since it's global.

This silently turns a failing symbolic test into a pass, so I think it should be fixed in this PR. Minimal fix: keep a separate pre-setup check for the fuzz-only empty-suite case, then recompute `functions` after `self.setup(call_setup)` as before.

mablr: https://github.com/foundry-rs/foundry/pull/15227/commits/8519a95a0e733d2c8f9668b081eec304078409e1

crates/evm/evm/src/executors/showmap.rs
mablr: added todo https://github.com/foundry-rs/foundry/pull/15227/commits/82150f43c5b3c2f8dbacf27880556d366765c628

figtracer (dismissed):
lgtm

figtracer (approved):

figtracer (dismissed):

crates/evm/evm/src/executors/showmap.rs
```diff
@@ -105,52 +178,65 @@ pub struct ShowmapStats {
...
             let mut call_result = execute_tx(&mut executor, tx)?;
+            // Snapshot the edge fingerprint before any coverage merge zeroes the buffer.
+            let fingerprint = snapshot_edge_fingerprint(&call_result);
+            // `vm.assume` rejects and `vm.skip` are discarded by the campaign: the call is not
+            // committed, checked, or counted toward coverage.
+            if call_result.result.as_ref() == MAGIC_ASSUME
+                || SkipReason::decode(&call_result.result).is_some()
+            {
+                continue;
+            }
```
grandizzy: Major: skip payloads are classified without checking the reverter, diverging from the real fuzz path.

The campaign only treats a `SkipReason` payload as a skip when the revert originates from the cheatcode address (`run_worker`). Here any user revert whose bytes happen to decode as a skip payload is silently discarded, so a corpus entry that the real campaign would treat as a failure is reported as a skip. Note also that the invariant campaign discards only `MAGIC_ASSUME` (not `SkipReason`); discarding skips here can hide a `fail_on_revert` handler failure — consider gating the skip-discard on `target.fuzzed_function.is_some()` as a follow-up.

```suggestion
            // `vm.assume` rejects and cheatcode `vm.skip` are discarded by the campaign: the
            // call is not committed, checked, or counted toward coverage. Only skip payloads
            // reverting from the cheatcode address count, mirroring the stateless fuzz path.
            if call_result.result.as_ref() == MAGIC_ASSUME
                || (call_result.reverter == Some(CHEATCODE_ADDRESS)
                    && SkipReason::decode(&call_result.result).is_some())
            {
                continue;
            }
```

crates/evm/evm/src/executors/fuzz/mod.rs
```diff
@@ -261,6 +261,104 @@ impl<FEN: FoundryEvmNetwork> FuzzedExecutor<FEN> {
...
+        if call.result.as_ref() == MAGIC_ASSUME {
+            return Ok(FuzzTestResult {
+                skipped: true,
+                reason: Some("persisted fuzz failure rejected by `vm.assume`".to_string()),
+                ..Default::default()
+            });
+        }
+        if let Some(reason) = SkipReason::decode(&call.result) {
+            return Ok(FuzzTestResult { skipped: true, reason: reason.0, ..Default::default() });
+        }
```
grandizzy: Major (same root cause): unguarded `SkipReason::decode` on the persisted-failure replay.

A persisted counterexample that reverts with skip-decodable bytes from a non-cheatcode address will be reported as `[SKIP]` instead of reproducing the failure. Guard on the cheatcode address, matching `run_worker`.

```suggestion
        if call.reverter == Some(CHEATCODE_ADDRESS)
            && let Some(reason) = SkipReason::decode(&call.result)
        {
            return Ok(FuzzTestResult { skipped: true, reason: reason.0, ..Default::default() });
        }
```

crates/evm/evm/src/executors/showmap.rs
```diff
@@ -105,52 +178,65 @@ pub struct ShowmapStats {
...
+    target: ShowmapReplayTarget<'_>,
     opts: &ShowmapOpts,
 ) -> Result<ShowmapStats> {
-    let replay_dirs = canonical_replay_dirs(corpus_dir);
-    if !replay_dirs.iter().any(|d| d.is_dir()) {
-        return Err(eyre!("corpus directory not found: {}", corpus_dir.display()));
+    // TODO(@mablr): if `corpus_dir` points at a generated corpus root, first narrow it
+    // to the selected contract/test corpus path so selector collisions in sibling contracts do not
+    // contaminate replay. Keep accepting flat/manual corpus directories as-is.
+    let entries = read_corpus_tree(corpus_dir)?;
```
grandizzy, line 184: Major: generated corpus roots can be replayed against the wrong target.

`read_corpus_tree` recursively reads the entire `--corpus-dir`, and stateless `WorkerCorpus::can_replay_tx` filters by selector only, not by originating contract/test. Passing a generated root (e.g. `fuzz_corpus`) means a sibling test sharing a fuzz selector gets replayed against the selected target, producing a false pass/fail. This is exactly the inline TODO above. Suggest narrowing generated layouts before reading:

```diff
- let entries = read_corpus_tree(corpus_dir)?;
+ // Narrow a generated corpus root to the selected target before replaying so sibling
+ // contracts/tests sharing a fuzz selector cannot contaminate this run.
+ let scoped_dir = narrow_generated_corpus_root(corpus_dir, &target);
+ let entries = read_corpus_tree(&scoped_dir)?;
```

where `narrow_generated_corpus_root` resolves `<root>/<contract>/<test>` for stateless fuzz and `<root>/<contract>` for invariants when those paths exist, falling back to `corpus_dir` for flat/manual corpora. This needs the contract/test identity threaded into `ShowmapReplayTarget` (or done in `runner.rs` before `run_showmap`). If deferring, please get explicit sign-off — this is a correctness gap in a newly advertised command, not just UX.

crates/evm/evm/src/executors/showmap.rs
```diff
@@ -167,21 +253,97 @@ pub fn replay_corpus_to_showmap<FEN: FoundryEvmNetwork>(
...
         }
-        rollback_replay_created(fuzzed_contracts, created);
+        if let Some(failure) = entry_failure {
+            rollback_replay_created(target.fuzzed_contracts, created);
+            replay_failures.push(format!(
+                "corpus entry {} failed during replay: {failure}",
+                entry.path.display()
+            ));
+            continue;
+        }
```
grandizzy, line 343: Minor: unbounded failure aggregation.

Replay keeps iterating all (up to `MAX_CORPUS_ENTRIES` = 1M) entries and appends a formatted string per failure even after the result is already decided. Cap the collected detail (or fail fast):

```suggestion
        if let Some(failure) = entry_failure {
            rollback_replay_created(target.fuzzed_contracts, created);
            failed_entries += 1;
            if replay_failures.len() < MAX_REPORTED_REPLAY_FAILURES {
                replay_failures.push(format!(
                    "corpus entry {} failed during replay: {failure}",
                    entry.path.display()
                ));
            }
            continue;
        }
```

with a `const MAX_REPORTED_REPLAY_FAILURES: usize = 20;` and a trailing `… and N more` line appended from `failed_entries` when building the final error.

grandizzy (changes requested):
Please check findings below.

crates/evm/evm/src/executors/showmap.rs
mablr: will do in follow-up

grandizzy (approved):
lgtm

figtracer (approved):

PR #15229, opened 2026-06-19
fix(cast): parse hex call values before calldata

haoshengzhen:
## Motivation

`CallSpec::parse` already supports hex values through `parse_ether_or_wei`, but the outer parser skipped value parsing for fields starting with `0x`. As a result, a spec such as `address:0x0:0x1234` was interpreted as raw calldata `0x0`, and the actual calldata field was ignored.

This matters for batch call specs that combine a hex value with raw calldata or another call payload.

## Solution

- Treat a hex-looking field in the value position as a value when a following sig/data field makes the field order unambiguous.
- Keep `to::0xdata` as the explicit form for raw calldata with no value.
- Add a regression test for `address:0x0:0x1234`.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cast/src/call_spec.rs
+++ b/crates/cast/src/call_spec.rs
@@ -67,10 +67,13 @@ impl CallSpec {
         // Pattern: to:value:sig:args or to::sig:args (empty value) or to:value:0xdata
         let mut idx = 1;
 
-        // Check for value (non-empty, not starting with 0x unless it's a number)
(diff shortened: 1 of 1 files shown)


stevencartavia (approved):

mattsse (approved):
We can keep this as is.

PR #15230, opened 2026-06-19
feat(anvil): allow precompile factories to customize install

mattsse:
## What changed

Adds an `install(&mut PrecompilesMap)` hook to Anvil's `PrecompileFactory`, with the existing `precompiles()` vector remaining the default implementation path.

Anvil now calls the factory install hook when building precompile maps, and the tests cover a factory that installs dynamic lookup behavior instead of returning a static address list.

## Why

Some native precompile integrations need more control than a fixed `(Address, DynPrecompile)` list can provide, such as prefix-based or dynamically resolved precompile dispatch.

## Impact

Existing `PrecompileFactory` implementors remain source-compatible because `install` has a default implementation. Integrations that need custom precompile-map setup can override the new hook.

### DIFF
--- a/crates/anvil/src/evm/mod.rs
+++ b/crates/anvil/src/evm/mod.rs
@@ -1,4 +1,4 @@
-use alloy_evm::precompiles::DynPrecompile;
+use alloy_evm::precompiles::{DynPrecompile, PrecompilesMap};
 use alloy_primitives::Address;
 use std::fmt::Debug;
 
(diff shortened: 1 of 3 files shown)


figtracer (approved):

crates/anvil/src/evm/mod.rs
```diff
@@ -45,27 +50,45 @@ mod tests {
...
+
+    impl PrecompileFactory for DynamicLookupPrecompileFactory {
+        fn precompiles(&self) -> Vec<(Address, DynPrecompile)> {
+            Vec::new()
+        }
+
+        fn install(&self, precompiles: &mut PrecompilesMap) {
+            precompiles.set_precompile_lookup(|address: &Address| {
+                address.as_slice().starts_with(&DYNAMIC_PRECOMPILE_PREFIX).then(echo_precompile)
+            });
```
mablr, line 91: `set_precompile_lookup` replaces any precompile already installed on the map? Which makes this `install` implementation unsafe to compose with other layers. Since it mutates an existing `PrecompilesMap`, maybe using `map_precompile_lookup` and delegating to `previous` for unmatched addresses would be more flexible/robust?

mattsse: > set_precompile_lookup replaces any precompile already installed on the map?

this isnt how it works, it's just extra and since this is a test, it's fine

PR #15231, opened 2026-06-19
fix(cheatcodes): allow vm.load on precompile accounts

mattsse:
## What changed

Removes the precompile-address rejection from `vm.load` while leaving mutation cheatcode guards intact.

Adds regression coverage showing `vm.load` can read a precompile account and `vm.store` / `vm.etch` still reject precompile targets.

## Why

`vm.load` is read-only, so blocking it on precompile accounts prevents tests from inspecting stateful precompile storage without adding a safety guarantee.

## Impact

Tests can inspect storage for stateful native precompiles directly. Mutating precompile accounts through cheatcodes remains blocked.


figtracer (approved):

mablr (approved):

mablr:
flaky CI should be good to go.

stevencartavia (approved):

PR #15234, opened 2026-06-19
feat(fuzz): sample invariant cmp logs in one worker

mablr:
## Motivation

Enable EVM comparison operand logging for only one invariant fuzz worker per parallel campaign.

This keeps cmp-guided mutation available for invariant corpus fuzzing while avoiding the overhead of collecting comparison operands in every worker. Single-worker campaigns keep the existing behavior, and sancov edge coverage still disables EVM cmp logging.

Closes OSS-278

### DIFF
--- a/crates/evm/evm/src/executors/corpus.rs
+++ b/crates/evm/evm/src/executors/corpus.rs
@@ -260,13 +260,24 @@ impl WorkerCorpusSeed {
         self
     }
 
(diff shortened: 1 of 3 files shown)


mablr:
derek bench invariant

mattsse (approved):
LGTM

stevencartavia (approved):

0xKarl98 (approved):

PR #15235, opened 2026-06-19
feat(forge): configure fuzzer ensemble defaults

figtracer:
## Summary

Adds first-class configuration for fuzzer ensemble knobs via TOML, CLI flags, and environment variables:

- dictionary weights and dictionary size caps for fuzz and invariant runs
- invariant `min_depth` and `depth_mode = "fixed" | "random"`
- corpus fresh-input probability and per-mutator weights
- payable `msg.value` generation probability

The runtime wires these through fuzz/invariant call generation and corpus mutation. Corpus mutation now uses precomputed `WeightedIndex` distributions, keeps fuzz and invariant literal dictionaries separate, honors zero ABI/CMP mutator weights as disabled paths, and preserves the all-zero mutator fallback to defaults.

The corpus fresh-input default remains `10%` for normal and two-worker runs. When an invariant campaign uses 3+ workers and the user has not explicitly configured `invariant.corpus_random_sequence_weight`, the last worker uses the exploratory `50%` setting while the other workers keep `10%`. Explicit TOML/CLI/env values are preserved for every worker. The invariant dictionary weight still applies to generated calldata as well as sender selection: default `80%` for both, where invariant calldata previously used a fixed `40%` dictionary branch. Other defaults stay conservative: fixed depth, mutator weights `1/1/1/1/1/1/1`, fuzz payable value `0%`, invariant payable value `15%`.

### DIFF
--- a/crates/config/src/invariant.rs
+++ b/crates/config/src/invariant.rs
@@ -105,6 +105,7 @@ pub enum InvariantDepthMode {
     #[default]
     Fixed,
     /// Sample every invariant run depth uniformly between `min_depth` and `depth`.
+    #[serde(alias = "uniform")]
     Random,
 }
 
@@ -255,6 +256,10 @@ mod tests {
         assert_eq!("uniform".parse::<InvariantDepthMode>().unwrap(), InvariantDepthMode::Random);
         assert_eq!(
             serde_json::from_str::<InvariantDepthMode>(r#""random""#).unwrap(),
+            InvariantDepthMode::Random
+        );
+        assert_eq!(
+            serde_json::from_str::<InvariantDepthMode>(r#""uniform""#).unwrap(),
             InvariantDepthMode::Random
         );
     }
--- a/crates/config/src/invariant.rs
+++ b/crates/config/src/invariant.rs
@@ -124,7 +124,7 @@ impl FromStr for InvariantDepthMode {
 }
 
 /// Contains for invariant testing
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct InvariantConfig {
     /// The number of runs that must execute for each invariant test group.
     pub runs: u32,
@@ -157,6 +157,10 @@ pub struct InvariantConfig {
     /// The fuzz corpus configuration.
     #[serde(flatten)]
     pub corpus: FuzzCorpusConfig,
+    /// Whether `corpus_random_sequence_weight` was supplied by a config provider.
+    #[serde(default, skip_serializing)]
+    #[doc(hidden)]
+    pub corpus_random_sequence_weight_configured: bool,
     /// Path where invariant failures are recorded and replayed.
     pub failure_persist_dir: Option<PathBuf>,
     /// Whether to collect and display fuzzed selectors metrics.
@@ -179,6 +183,32 @@ pub struct InvariantConfig {
     pub check_interval: u32,
 }
 
+impl PartialEq for InvariantConfig {
+    fn eq(&self, other: &Self) -> bool {
+        self.runs == other.runs
+            && self.depth == other.depth
+            && self.min_depth == other.min_depth
+            && self.depth_mode == other.depth_mode
+            && self.workers == other.workers
+            && self.fail_on_revert == other.fail_on_revert
+            && self.call_override == other.call_override
+            && self.dictionary == other.dictionary
+            && self.shrink_run_limit == other.shrink_run_limit
+            && self.max_assume_rejects == other.max_assume_rejects
+            && self.gas_report_samples == other.gas_report_samples
+            && self.corpus == other.corpus
+            && self.failure_persist_dir == other.failure_persist_dir
+            && self.show_metrics == other.show_metrics
+            && self.timeout == other.timeout
+            && self.show_solidity == other.show_solidity
+            && self.max_time_delay == other.max_time_delay
+            && self.max_block_delay == other.max_block_delay
+            && self.check_interval == other.check_interval
+    }
+}
+
+impl Eq for InvariantConfig {}
+
 impl Default for InvariantConfig {
     fn default() -> Self {
         Self {
@@ -194,6 +224,7 @@ impl Default for InvariantConfig {
             max_assume_rejects: 65536,
             gas_report_samples: 256,
             corpus: FuzzCorpusConfig::default(),
+            corpus_random_sequence_weight_configured: false,
             failure_persist_dir: None,
             show_metrics: true,
             timeout: None,
--- a/crates/config/src/invariant.rs
+++ b/crates/config/src/invariant.rs
@@ -97,6 +97,31 @@ fn fixed_workers(workers: usize) -> Result<InvariantWorkers, String> {
         .ok_or_else(|| "invariant workers must be greater than 0".to_string())
 }
 
+/// Per-run invariant depth selection mode.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum InvariantDepthMode {
+    /// Execute every invariant run up to the configured `depth`.
+    #[default]
+    Fixed,
+    /// Sample every invariant run depth uniformly between `min_depth` and `depth`.
+    Random,
+}
+
+impl FromStr for InvariantDepthMode {
+    type Err = String;
+
+    fn from_str(value: &str) -> Result<Self, Self::Err> {
+        match value.trim().to_ascii_lowercase().as_str() {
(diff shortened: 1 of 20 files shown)


crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -177,6 +178,19 @@ fn max_invariant_workers_for_campaign(runs: u32, depth: u32) -> usize {
         .unwrap_or(usize::MAX)
 }
 
+fn invariant_run_depth(config: &InvariantConfig, runner: &mut TestRunner) -> u32 {
+    match config.depth_mode {
+        InvariantDepthMode::Fixed => config.depth,
+        InvariantDepthMode::Random => {
+            if config.depth <= config.min_depth {
+                config.depth
```
mablr: `min_depth` seems not validated, so `depth_mode = "random"` with `min_depth = 0` can sample `run_depth = 0`. That run then skips the handler loop while still counting toward `runs`, silently wasting part of the invariant campaign ?

crates/evm/fuzz/src/strategies/invariants.rs
```diff
@@ -160,19 +173,23 @@ pub fn fuzz_contract_with_calldata<S: FuzzStateReader>(
...
+    let dictionary_weight = dictionary_weight.min(100);
 
     // We need to compose all the strategies generated for each parameter in all possible
     // combinations.
     // `prop_oneof!` / `TupleUnion` `Arc`s for cheap cloning.
     let calldata_strategy = prop_oneof![
-        60 => fuzz_calldata(func.clone(), fuzz_fixtures),
-        40 => fuzz_calldata_from_state(func, fuzz_state),
+        100 - dictionary_weight => fuzz_calldata(func.clone(), fuzz_fixtures),
+        dictionary_weight => fuzz_calldata_from_state(func, fuzz_state),
```
stevencartavia, line 187: On `master`, `fuzz_contract_with_calldata` hardcoded the calldata strategy as `60 => fuzz_calldata` / `40 => fuzz_calldata_from_state` and didn't take `dictionary_weight` — the configured weight was only used for sender selection. Here it's wired into calldata, so with the default invariant `dictionary_weight = 80` the state-dictionary calldata branch goes from 40% to 80% for default invariant runs.

I checked this with a fixed-seed invariant target using a fixed `targetSender` and `poke(address)` (so sender selection can't affect the result): on `master`, `dictionary_weight=0` and `=100` produce identical calldata; here they differ (0 → all-random, 100 → all-dictionary).

The summary lists invariant `dictionary_weight = 80%` among defaults that "stay conservative or preserve existing behavior" — is the calldata change intended? If so, might be worth a note in the summary since it shifts default invariant calldata generation.

figtracer: this is intentional, documented on desc

figtracer:
derek bench invariant

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -703,16 +727,22 @@ impl WorkerCorpus {
...
+        let mutation_distribution = WeightedIndex::new([
+            mutation_weights.mutation_weight_splice,
+            mutation_weights.mutation_weight_repeat,
+            mutation_weights.mutation_weight_interleave,
+            mutation_weights.mutation_weight_prefix,
+            mutation_weights.mutation_weight_suffix,
+            mutation_weights.mutation_weight_abi,
+            mutation_weights.mutation_weight_cmp,
+        ])
+        .expect("effective mutation weights contain at least one non-zero entry");
```
mablr: `WeightedIndex::new` might also fail with `Error::Overflow` when configured `mutation_weight_*` values are individually valid `u32`s but their sum overflows the distribution’s internal total? Since these weights are user-configurable, currently this would panic with a misleading “at least one non-zero entry” message.

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -177,6 +178,20 @@ fn max_invariant_workers_for_campaign(runs: u32, depth: u32) -> usize {
         .unwrap_or(usize::MAX)
 }
 
+fn invariant_run_depth(config: &InvariantConfig, runner: &mut TestRunner) -> u32 {
+    match config.depth_mode {
+        InvariantDepthMode::Fixed => config.depth,
```
mablr: `depth_mode = "fixed"` still allows `invariant.depth = 0`, so `invariant_run_depth` can return `0` and the loop later evaluates `run_depth - 1`. The random-depth path now clamps to at least 1, so fixed mode should probably get the same treatment or config should reject zero depth.

mablr (commented):
Overall sgtm, 2 smol nits.

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -85,6 +85,35 @@ const FAVORABILITY_THRESHOLD: f64 = 0.3;
...
+}
+
+fn assert_supported_mutation_weight_total(mutation_weights: FuzzCorpusMutationWeights) {
+    let total = mutation_weights.total();
+    assert!(
+        total <= u64::from(u32::MAX),
+        "effective mutation weights sum to {total}, which exceeds the maximum supported total {}",
+        u32::MAX
+    );
+}
```
mablr: These weights are user-configurable via TOML/CLI/env, but overflowing totals are only caught by an `assert!` during `WorkerCorpus::from_seed`, which panics mid-run. This should be a normal config/runtime error instead of an assertion?

crates/config/src/invariant.rs
```diff
@@ -149,11 +183,39 @@ pub struct InvariantConfig {
...
+            && self.gas_report_samples == other.gas_report_samples
+            && self.corpus == other.corpus
+            && self.failure_persist_dir == other.failure_persist_dir
+            && self.show_metrics == other.show_metrics
+            && self.timeout == other.timeout
+            && self.show_solidity == other.show_solidity
+            && self.max_time_delay == other.max_time_delay
+            && self.max_block_delay == other.max_block_delay
+            && self.check_interval == other.check_interval
+    }
```
mablr, line 209: `InvariantConfig == InvariantConfig` ignores `corpus_random_sequence_weight_configured`, even though that flag may change worker behavior?

crates/forge/src/multi_runner.rs
```diff
@@ -705,6 +707,11 @@ impl MultiContractRunnerBuilder {
             Some(self.config.project_paths()),
             self.config.fuzz.dictionary.max_fuzz_dictionary_literals,
         );
+        let invariant_literals = LiteralsDictionary::new(
+            Some(analysis.clone()),
+            Some(self.config.project_paths()),
+            self.config.invariant.dictionary.max_fuzz_dictionary_literals,
+        );
```
mablr, line 711: With default/equal `max_fuzz_dictionary_literals` values, this seems now to start two literal collector threads over the same analysis and stores duplicate dictionaries?

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -177,6 +178,20 @@ fn max_invariant_workers_for_campaign(runs: u32, depth: u32) -> usize {
         .unwrap_or(usize::MAX)
 }
 
+fn invariant_run_depth(config: &InvariantConfig, runner: &mut TestRunner) -> u32 {
+    match config.depth_mode {
+        InvariantDepthMode::Fixed => config.depth.max(1),
```
mablr: This changes explicit fixed `invariant.depth = 0` from “run zero calls” into “run one call” because fixed mode now returns `config.depth.max(1)`. The existing `ZeroRuns.t.sol` fixture shows the behavior shift from a zero-call invariant assertion to `calls: 256, reverts: 256`. If zero depth is no longer supported, config loading should reject it with a clear error; otherwise fixed mode should preserve `0` and only random mode should clamp its sampled minimum.

crates/config/src/lib.rs
```diff
@@ -859,7 +868,14 @@ impl Config {
         figment: Figment,
         strict_profile: bool,
     ) -> Result<Self, ExtractConfigError> {
+        let invariant_corpus_random_sequence_weight_configured = figment
+            .extract_inner::<bool>("invariant.corpus_random_sequence_weight_configured")
+            .unwrap_or_else(|_| {
+                figment_value_is_configured(&figment, "invariant.corpus_random_sequence_weight")
+            });
```
mablr, line 875: This provenance check treats values inherited from `[fuzz]` via the standalone fallback as explicit invariant configuration. With only `[fuzz] corpus_random_sequence_weight = 10`, `invariant.corpus_random_sequence_weight_configured` becomes true, so the 3+ worker invariant ensemble would skip the documented exploratory `50%` worker even though `[invariant] corpus_random_sequence_weight` was never set?

mablr (approved):
sgtm

stevencartavia (approved):

PR #15236, opened 2026-06-19
feat(lint): add require-revert-in-loop detector

Berektassuly:
## Motivation

Refs #14381.

Adds the `require-revert-in-loop` detector from the lint parity checklist. A single failing `require` or `revert` inside a loop can revert the entire batch, so surfacing these cases helps users review whether validation should happen before the loop, skip invalid entries, or split work into separate calls.

## Solution

Adds a low-severity Solidity lint for `require` calls and `revert` statements executed in loop context. The detector covers direct loop bodies, internal helper calls reached from loops, modifier placeholder execution inside loops, and inline assembly `revert`.

The shared loop-context visitor now exposes both statements and expressions so this lint can reuse the same internal-call and modifier traversal already used by loop-related lints, while avoiding duplicate reports for helpers that introduce their own loops and are called from outside a loop.

## Validation

Tests pass: `cargo test -p forge --test ui -- RequireRevertInLoop`

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


mablr (approved):
Fixed clippy warning + Added coverage.

Seems good, thank you @Berektassuly!

figtracer (approved):
lgtm

beriktassuly:
Opened the matching Foundry Book docs PR here: https://github.com/foundry-rs/book/pull/1911

This should address the `ensure_lint_rule_docs` failure once the Book PR is merged and deployed

PR #15238, opened 2026-06-20
fix(doc): match inherited params with underscore-wrapped names

### LINKED ISSUE #7331: forge(doc): @inheritdoc does not pickup description
### Component

Forge

`@inheritdoc` should pickup the parameters description even if the parameters name differ.

### Additional context

I have the following Natspec doc in my interface:
```
    /**
     * @notice Mints `amount` token to `recipient`.
     * @param  recipient The address of the account receiving minted token.
     * @param  amount    The amount of token to mint.
     */
    function mint(address recipient, uint256 amount) external;
```

In my contract, I inherit from the Natspec doc the following way:
```j
    /// @inheritdoc IZeroToken
    function mint(address recipient_, uint256 amount_) external onlyStandardGovernor {
        _mint(recipient_, amount_);
    }
```

As you can see, the parameters are suffixed by `_` and in consequence, `forge doc` does not pickup the parameters description:
Screenshot 2024-03-06 at 5 06 27 PM

I think the `try_match_inheritdoc` function can be improved to match parameters that are prefixed or suffixed by `_`:
https://github.com/foundry-rs/foundry/blob/b6e7c8b50a351ec5779c8cf4790854cc0b23ce8c/crates/doc/src/preprocessor/inheritdoc.rs#L81

flag4table:
## Motivation

Fixes #7331.

`forge doc` could resolve `@inheritdoc` from an interface, but inherited
`@param` descriptions were not rendered when the implementation used parameter
names with leading or trailing underscores, such as `recipient_` or `_amount`.

## Solution

When merging inherited NatSpec params, map inherited parameter names to the
current function parameter name if the names match after trimming leading and
trailing underscores.

The matching remains conservative: exact matches are preferred, and underscore
matching is only used when there is a single matching current parameter.

A regression test covers both suffix and prefix underscore cases.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/doc/src/render.rs
```diff
@@ -654,6 +655,34 @@ fn render_function_section(
...
+
+    let normalized_inherited = inherited_name.trim_matches('_');
+    if normalized_inherited.is_empty() {
+        return inherited_name.to_string();
+    }
+
+    let mut candidates = params.iter().filter_map(|param| {
+        let name = param.name?;
+        let name = name.as_str();
+        (name.trim_matches('_') == normalized_inherited).then_some(name)
```
mablr: This does not compile under `cargo +nightly test -p forge --test cli inheritdoc_matches_underscore_wrapped_param_names`: the closure returns a `&str` derived from the local `Ident` binding `(name.as_str())`. Consider returning an owned `String` from the candidate iterator or restructuring so the returned value is not borrowed through the local binding.

mablr (dismissed):

crates/doc/src/render.rs
flag4table: Thanks, fixed by making the candidate iterator return an owned String instead of a &str borrowed through the local Ident binding.

Please review it again.

crates/doc/src/render.rs
```diff
@@ -634,8 +634,9 @@ fn render_function_section(
             );
         }
         for (name, desc) in &inherited.params {
-            if !c.params.iter().any(|(n, _)| n == name) {
-                c.params.push((name.clone(), sanitize(desc)));
+            let name = inherited_param_name_for_current_function(name, &f.header.parameters);
+            if !c.params.iter().any(|(n, _)| n == &name) {
+                c.params.push((name, sanitize(desc)));
             }
```
mablr, line 668: The helper guards against current-side ambiguity (exactly one candidate), but maybe not inherited-side ambiguity. If two documented inherited params normalize to the same trimmed core (e.g. interface has both amount and `_amount`, impl has other and `_amount`), the fuzzy match remaps inherited amount -> `_amount` first, then the dedup at L638 suppresses the exact `_amount` doc, so `_amount` may render the wrong description? Fix: maybe also require the trimmed core to be unique on the inherited side before fuzzy-remapping.

mablr (commented):
Looks good, maybe just a small nit.

crates/doc/src/render.rs
grandizzy: pushed https://github.com/foundry-rs/foundry/pull/15238/changes/7cd45fdf85607d9fcea50d47bd396e9a58517dbc pls check

grandizzy (approved):
thank you, lgtm

mablr (approved):
Thanks

PR #15239, opened 2026-06-20
feat(cast): guard receive-policy T6

stevencartavia:
`cast receive-policy claim` and `receipt burn` send directly to the `ReceivePolicyGuard` precompile, which only has code from T6 onward — pre-T6 those txs silently succeed as no-ops. This adds a T6 preflight (hardfork check, falling back to `eth_getCode`) so both commands fail early with a clear message. Includes a CLI regression test.


mattsse (dismissed):
LGTM

mattsse (approved):
LGTM

PR #15240, opened 2026-06-20
feat(cast): add tip403 command

stevencartavia:
Adds `cast tip403` to create and manage TIP-403 whitelist/blacklist policies (create, whitelist, blacklist, info, check), so the policy IDs that `cast receive-policy set` requires can be created from cast.


mattsse (approved):
LGTM

PR #15241, opened 2026-06-20
chore: fix clippy warning

stevencartavia:
fixed `warning: this function may allocate 516342 bytes on the stack --> crates/cast/src/cmd/tip20/mod.rs:207:23`


mattsse (approved):
LGTM

PR #15242, opened 2026-06-20
feat(forge): inspect contract artifact json

### LINKED ISSUE #15119: Add to `forge inspect` standard output JSON
### Component

Forge

### The problem

I need to extract some contract details for my tools, like the ABI and the bytecode. I have 2 choices:
1. Parsing the standard JSON output from `out/Contract.sol/Contract.json`. It would be just fine, its a simple, stable format, but the hard part is getting the path to the contract, there are so many edge cases to handle. The project may not use the `out` directory, there may be 2 contracts with the same name, the project may not be built yet. I would strongly prefer using `forge inspect`, that's exactly why this tool exists.
2. Using `forge inspect`. It isn't particularly well suited for tools usage, e.g. requesting the ABI formats a human-readable table which is a pain to parse. Getting 2 artifacts needs running `forge inspect` twice and generates 2 outputs that either need to be stored in 2 files or need to be custom-glued into a single JSON.

### The solution

Add to `forge inspect` standard output JSON as the 'field' parameter, e.g. named 'output'. It would just open and print the `out/Contract.sol/Contract.json` file verbatim. It should be the easiest output format to handle by `forge inspect` but it would add a lot of flexibility to 3rd party tooling without the edge cases. The user could either parse the whole file and pick whatever fields they need, or feed it into e.g. jq and slim it down before further handling.

### Additional context

_No response_

Osraka:
## Motivation

Closes #15119.

`forge inspect` already resolves the target contract and handles project-specific artifact lookup, including custom output directories and path-qualified contract identifiers. However, tools that need multiple artifact fields currently either parse the artifact file from `out/...` themselves or call `forge inspect` multiple times and stitch the results together.

This adds a direct way to ask `forge inspect` for the resolved contract artifact JSON, so downstream tooling can reuse Foundry's contract resolution instead of reimplementing artifact path edge cases.

## Solution

Add a new `forge inspect <contract> artifact` field that prints the resolved `ConfigurableContractArtifact` as pretty JSON.

Also adds aliases:

- `artifactJson`
- `artifact-json`
- `artifact_json`
- `output`

`standard-json` is left unchanged and continues to return the compiler standard JSON input. The new `artifact`/`output` field is intentionally separate from human-readable fields such as `abi`, `events`, and `errors`.

Tests cover:

- artifact JSON is valid JSON and includes ABI, creation bytecode, and deployed bytecode
- path-qualified contract resolution with duplicate contract names
- custom `out` directory behavior
- the `output` alias

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

Validated locally with:

```console
cargo +nightly fmt --check
cargo test -p forge contract_output_selection
cargo test -p forge --test cli cmd::can_inspect_artifact_json -- --exact
cargo test -p forge --test cli cmd::can_inspect_artifact_json_with_custom_out_and_duplicate_contract_names -- --exact
```


alirezz0 (approved):

alirezz0 (approved):

figtracer (approved):
lgtm

stevencartavia (approved):

DaniPopes (approved):

PR #15245, opened 2026-06-21
perf(corpus): trim mutation allocations

decofe:
## perf(corpus): trim mutation allocations

Reduces allocations in the hot `WorkerCorpus::new_inputs` mutation paths:

- **Splice**: `reserve` + `extend_from_slice` instead of per-element `push` in a loop.
- **Repeat**: overwrite the target range in place instead of building a temp `Vec` and `splice`-ing it in.
- **Interleave**: `reserve` the output up front.
- **Cmp**: iterate candidate indices lazily instead of collecting them into a `Vec<usize>`.

Behavior is unchanged — same mutations, same RNG draws, same output sequences. Purely an allocation/overhead reduction.


decofe:

mablr (approved):

figtracer (approved):

0xKarl98 (approved):

PR #15246, opened 2026-06-21
perf(fuzz): avoid rebuilding storage layout map

decofe:
`FuzzDictionary::insert_new_state_values` previously received a freshly built `HashMap<Address, Arc<StorageLayout>>` from `targets.get_storage_layouts()`, which iterated **all** targeted contracts and cloned every `Arc` on each call — even when `include_storage` is disabled. This runs once per executed tx during invariant campaigns, so the cost scaled with the full target set regardless of how many accounts a tx actually touched.

This passes `&TargetedContracts` directly and looks up each changed address via `targets.get(address)`:

- No per-call `HashMap` allocation.
- `Arc` clones drop from "all targeted layouts" to only the changed addresses that need a `SlotIdentifier`.
- No storage-layout work at all when `include_storage` is off.

Behavior is unchanged — same lookups resolve to the same storage layouts. Purely an allocation/overhead reduction.

Also removes the now-unused `TargetedContracts::get_storage_layouts` and its obsolete `HashMap` import.

### DIFF
--- a/crates/evm/fuzz/src/invariant/mod.rs
+++ b/crates/evm/fuzz/src/invariant/mod.rs
@@ -1,5 +1,5 @@
 use alloy_json_abi::{Function, JsonAbi};
-use alloy_primitives::{Address, Selector, map::HashMap};
+use alloy_primitives::{Address, Selector};
 use foundry_compilers::artifacts::StorageLayout;
 use itertools::Either;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


decofe:

figtracer (dismissed):

mablr (dismissed):

mablr (approved):

stevencartavia (approved):

PR #15247, opened 2026-06-21
perf(invariant): skip terminal corpus input generation

decofe:
Skips generating the terminal corpus input in the per-run invariant loop. The loop previously generated one input past the final depth that was never executed, wasting an RNG draw and leaking an unexecuted input into corpus processing. Generation is now guarded on `depth < config.depth`, so it only runs when another iteration will execute the input (discarded/popped iterations do not increment `depth`, so the replacement is still produced).

Corpus processing also handles empty input sequences: a run where every call is discarded (magic assume) or popped (reverts without `fail_on_revert`, handler assertions) can leave `inputs` empty while `new_coverage` is still set, since edge coverage is collected before the pop. `process_inputs_inner` now updates mutation/optimization state and skips persisting a corpus entry when no surviving inputs remain, instead of asserting non-empty.

## Tests

- Added `empty_input_sequence_with_new_coverage_does_not_panic_or_insert` covering the deferred and live processing paths.
- `cargo test -p foundry-evm --lib executors::corpus::tests` passes (20).

### DIFF
--- a/crates/evm/evm/src/executors/corpus.rs
+++ b/crates/evm/evm/src/executors/corpus.rs
@@ -834,12 +834,17 @@ impl WorkerCorpus {
         // When the run is interesting only because of optimization (no new coverage),
         // add the best prefix to the corpus instead of the full run — the prefix is
         // the sequence that actually achieved the best value.
-        assert!(!inputs.is_empty());
-        let corpus_inputs = if improved_optimization && !new_coverage {
+        //
+        // `inputs` can be empty when every call was discarded/popped but new coverage was
+        // still recorded; there's nothing to persist, so skip without inserting an entry.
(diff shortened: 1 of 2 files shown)


decofe:

mablr (dismissed):

mablr (commented):
It seems that `invariant_optimization_with_warp` test regressed

grandizzy:
> It seems that `invariant_optimization_with_warp` test regressed

should be fixed by https://github.com/foundry-rs/foundry/pull/15247/changes/505b59c344e9c69a5ec0f53f7e023be7577b83e6

mablr (approved):

figtracer (approved):

PR #15248, opened 2026-06-21
feat(lint): add non-reentrant-not-first detector

Osraka:
## Motivation

Part of #14381.

Solidity applies modifiers in the order they are written. When an OpenZeppelin-style `nonReentrant` modifier appears after another modifier, that modifier's pre-body logic runs before the reentrancy guard is entered.

This adds a focused `forge lint` detector for that ordering issue without trying to infer broader reentrancy behavior.

## Solution

- Add a medium-severity `non-reentrant-not-first` late lint.
- Inspect functions, fallback functions, and receive functions.
- Match HIR-resolved modifiers named `nonReentrant` and report when the first matching modifier is not first in the modifier list.
- Keep the detector intentionally narrow: ordering-only, no CFG requirements, and no attempt to infer custom guard semantics.
- Add lint docs, README registration, and UI coverage for local and inherited `nonReentrant` modifiers plus non-triggering cases.

## Testing

- `git diff --check`
- `cargo test -p forge --test ui -- NonReentrantNotFirst`
- `cargo test -p forge-lint registered_lints_have`

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/lint/src/sol/med/non_reentrant_not_first.rs
+++ b/crates/lint/src/sol/med/non_reentrant_not_first.rs
@@ -1,0 +1,51 @@
+use super::NonReentrantNotFirst;
+use crate::{
+    linter::{LateLintPass, LintContext},
+    sol::{Severity, SolLint},
+};
+use solar::sema::{
+    Gcx,
+    hir::{self, FunctionKind},
+};
+
+declare_forge_lint!(
+    NON_REENTRANT_NOT_FIRST,
+    Severity::Med,
+    "non-reentrant-not-first",
+    "`nonReentrant` should be the first modifier"
+);
+
+impl<'hir> LateLintPass<'hir> for NonReentrantNotFirst {
+    fn check_function(
+        &mut self,
+        ctx: &LintContext,
+        _gcx: Gcx<'hir>,
+        hir: &'hir hir::Hir<'hir>,
+        func: &'hir hir::Function<'hir>,
+    ) {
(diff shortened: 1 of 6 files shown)


crates/lint/src/sol/med/non_reentrant_not_first.rs
```diff
@@ -0,0 +1,51 @@
...
+
+        if let Some((index, modifier)) = func
+            .modifiers
+            .iter()
+            .enumerate()
+            .find(|(_, modifier)| modifier_is_named(hir, modifier, "nonReentrant"))
+            && index != 0
+        {
+            ctx.emit(&NON_REENTRANT_NOT_FIRST, modifier.span);
+        }
```
mablr:
```suggestion
        func.modifiers
            .iter()
            .enumerate()
            .filter(|(index, modifier)| {
                *index != 0 && modifier_is_named(hir, modifier, "nonReentrant")
            })
            .for_each(|(_, modifier)| ctx.emit(&NON_REENTRANT_NOT_FIRST, modifier.span));
```

I think this is more robust to catch also `nonReentrant` duplicates. Coverage should be added for this.

mablr (commented):
Sgtm, thanks, just a small nit (non-blocking).

mattsse (approved):
LGTM

mablr (approved):

PR #15249, opened 2026-06-21
perf(invariant): cache metric selector keys

decofe:
`record_metrics` runs once per executed tx during invariant campaigns and previously recomputed the metric key every time — looking up the target contract, scanning its ABI for the selector, and building a `format!("{identifier}.{function}")` String on every call.

This caches the resolved metric key per `(target, selector)` so the ABI scan and string build happen only once per distinct pair, adding `fuzzed_metric_key_for_selector` and parsing the selector once from calldata.

Since invariant targets are updatable (contracts deployed mid-run are added via `collect_created_contracts` and cleared at run end, and an address can be reused for a different artifact across runs), the cache is kept coherent: only resolved keys are cached, and entries are invalidated when targets are added or removed. Behavior is otherwise unchanged.

### DIFF
--- a/crates/evm/evm/src/executors/invariant/mod.rs
+++ b/crates/evm/evm/src/executors/invariant/mod.rs
@@ -402,8 +402,10 @@ struct InvariantTestData {
     line_coverage: Option<HitMaps>,
     // Metrics for each fuzzed selector.
     metrics: Map<String, InvariantMetrics>,
-    // Cache from fuzzed target selector to the stable metric key.
-    metric_key_cache: Map<(Address, Selector), Option<String>>,
+    // Cache from fuzzed (target, selector) to its metric key. Only resolved keys are cached;
+    // entries are invalidated when targets are added/removed mid-campaign (see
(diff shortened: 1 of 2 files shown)


decofe:

grandizzy:
Pushed a correctness fix on top of the caching change: the `(target, selector) → metric key` cache could go stale because invariant targets are updatable — contracts deployed mid-run are added via `collect_created_contracts` and cleared at run end, and an address can be reused for a different artifact across runs.

Changes:
- No negative caching — unresolved keys aren't cached, so a tx whose target isn't known yet is re-resolved once that target is added.
- Invalidate cached entries when targets change: for newly added targets after `collect_created_contracts`, and for cleared targets in `end_run`.

mablr (approved):
The new cache invalidation approach seems good: unresolved keys are not cached, and resolved entries are dropped when dynamic targets are added or cleared.

LGTM

figtracer (approved):

stevencartavia (approved):

PR #15250, opened 2026-06-21
perf(fuzz): avoid eager observed call calldata

decofe:
The `Fuzzer` inspector's `call` hook materialized observed-call calldata via `inputs.input.bytes(ecx)` on every EVM sub-call, then discarded it unless call recording was active. For `CallInput::SharedBuffer` that copies the calldata into a fresh `Bytes` on each call.

Gate on `should_record_observed_call(scheme)` first and only materialize calldata (and value) when the call will actually be recorded — i.e. when `record_calls` is set (coverage-guided invariant fuzzing), the call is nested, and the scheme is `CALL`. Behavior is unchanged; only the wasted allocation on non-recorded calls is removed.

### DIFF
--- a/crates/evm/fuzz/src/inspector.rs
+++ b/crates/evm/fuzz/src/inspector.rs
@@ -150,7 +150,7 @@ impl Fuzzer {
     }
 
     #[inline]
-    fn should_record_observed_call(&self, scheme: CallScheme) -> bool {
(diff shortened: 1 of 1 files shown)


decofe:

mablr (approved):

figtracer (approved):

stevencartavia (approved):

PR #15251, opened 2026-06-21
perf(invariant): drain fuzzer values in place

decofe:
Drains the fuzzer's collected stack values into the dictionary in place instead of `mem::take`-ing the buffer each call. `mem::take` swapped in a fresh zero-capacity `Vec` and dropped the old allocation, forcing a realloc on the next collection; the new `InvariantFuzzState::collect_fuzzer_values` drains the values while retaining the buffer's capacity for reuse, removing per-call allocator churn in the hot invariant loop.

The now-unused `Fuzzer::drain_collected_values` is removed.


decofe:

mablr (approved):

figtracer (approved):

stevencartavia (approved):

PR #15252, opened 2026-06-21
perf(invariant): recycle observed call buffers

decofe:


decofe:

grandizzy:
## Review: likely perf-neutral under the current `call_raw` clone boundary

The change is semantically sound, but as written the optimization probably doesn't deliver a speedup. Details below.

**What's fine**
- **Ownership is valid.** `observed_calls` comes from `std::mem::take(&mut call_result.observed_calls)`; `hoist_observed_calls` only borrows it, so moving it into `recycle_observed_calls` after the `if` block is safe.
- **No behavior change.** The original buffer was already taken out via `mem::take`, so downstream commit/invariant/shrink paths already saw an empty buffer. Recycling only reuses the allocation.
- **Recycle guard is correct.** The `self.observed_calls.is_empty()` check prevents dropping unprocessed calls, and keeping the larger capacity is reasonable high-water-mark behavior.

**The core concern — the recycled capacity is discarded on the next call**
- `Executor::call_raw` -> `call_with_env` does `let mut stack = self.inspector().clone();` before each call, and `convert_executed_result` drains observed calls from that **cloned** inspector.
- This PR recycles the buffer into the **source** inspector (`current_run.executor.inspector_mut().fuzzer`).
- The next call clones that source again, and `Vec::clone` copies elements, **not spare capacity**. Since the recycled Vec is cleared, the clone gets an empty, zero-capacity buffer — so the hot path still reallocates each call. The retained capacity sits unused on the source inspector.

**Test coverage is thin**
`recycle_observed_calls_reuses_capacity_without_entries` only proves the method stores an empty Vec with capacity. It doesn't exercise the clone boundary that defeats the optimization. A test that clones the fuzzer after recycling and asserts capacity survives would likely fail today and surface the real issue.

**Suggestion**
To make this actually pay off, the buffer needs to be reused/moved into the inspector clone that records calls, or `InspectorStack::clone` needs to preserve capacity. Worth confirming with the benchmark results before merging.

decofe:
Closing during perf PR triage: the optimization appears perf-neutral under the current call_raw/InspectorStack clone boundary, because the recycled spare Vec capacity is retained on the source inspector but not preserved into the cloned inspector used by the next call. This can be reopened or replaced if the reuse path moves across that boundary or benchmarks show a real win.

PR #15253, opened 2026-06-21
perf(corpus): index current mutation

decofe:
`process_inputs` credits the corpus entry that produced the current input on every coverage-guided run, previously locating it with an O(n) `iter().find()` over `in_memory_corpus` by UUID.

This records the entry's index instead, so crediting becomes an O(1) `get_mut(index)`. The index is selected after eviction in `new_inputs`/`new_input` and consumed (via `take`) at the end of the same run, with no removal or reordering of `in_memory_corpus` in between (appends are index-stable), so it always refers to the same entry.


decofe:

figtracer (approved):

mablr (approved):

PR #15254, opened 2026-06-21
perf(fuzz): skip empty call artifact lookup

decofe:
`collect_values_from_call` runs on the hot path for every committed fuzzed call. It always called `fuzzed_artifacts(tx)` to resolve the target ABI and function, even when there were no logs and no return data to decode — and that lookup scans the contract's functions computing a keccak selector per function (alloy's `Function::selector()` is not cached).

Skip the lookup when both `logs` and `result` are empty (common for state-mutating calls that return nothing and emit no events), and only run each `insert_*` decode when its input is non-empty. Both inserts are already no-ops on empty input, so dictionary contents are unchanged.

### DIFF
--- a/crates/evm/fuzz/src/strategies/state.rs
+++ b/crates/evm/fuzz/src/strategies/state.rs
@@ -147,9 +147,17 @@ impl InvariantFuzzState {
     ) {
         let mut dict = self.inner.borrow_mut();
         let targets = fuzzed_contracts.targets();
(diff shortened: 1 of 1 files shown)


decofe:

mablr (approved):

figtracer (approved):

stevencartavia (approved):

PR #15255, opened 2026-06-21
perf: use entry API for edge index map

decofe:
Switch the edge index lookup used during coverage merging to the map entry API and mark it inline.

This keeps stable edge index assignment while avoiding a second map lookup when a new edge is assigned.

### DIFF
--- a/crates/evm/evm/src/inspectors/edge_cov.rs
+++ b/crates/evm/evm/src/inspectors/edge_cov.rs
@@ -1,6 +1,6 @@
 use alloy_primitives::{
     Address, U256,
-    map::{DefaultHashBuilder, HashMap},
+    map::{DefaultHashBuilder, Entry, HashMap},
 };
 use core::{
     fmt,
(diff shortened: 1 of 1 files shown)


decofe:

kuyziss:
derek bench bal=true slack=on-win otlp=false

decofe:

decofe:

grandizzy:
## Review: LGTM — low-risk micro-optimization

Correct and a genuine (if modest) improvement. Details below.

**Correctness — stable index assignment preserved**
The `get`-then-`insert` and the `entry` versions are semantically equivalent for `EdgeIndexMap`: an existing `EdgeKey` returns its existing index; a new one is assigned the current `next_index` which is then incremented and inserted. `next_index` still advances only on the vacant path, so the "first seen gets the next dense index, later hits reuse it" invariant holds. (As before, stability is within the map's lifetime, not deterministic across independent runs if first-seen order differs — unchanged by this PR.)

**Perf claim — accurate, with the right scope**
- **Vacant path (new edges):** old code hashed/probed on the `get` miss and again on `insert`; `entry` does it once and inserts into the located vacant slot. This is the real win.
- **Occupied path (known edges):** `get(&edge)` and `entry(edge)` both hash/probe once — roughly neutral, no regression.

So this helps most early in a campaign or any workload discovering many new edges; once the map is warm and hits are mostly occupied, it's essentially neutral.

**Concerns checked**
- `edge_index` already takes `EdgeKey` by value (`Copy`), and call sites pass `hit.edge` from a drained owned iterator — the entry API introduces no new clone.
- `next_index` is a field disjoint from `self.edge_indices`, so borrowing it while holding the entry is fine under field-sensitive borrow rules.
- `#[inline]` is harmless but likely minor, since the hot caller is in-crate and LLVM can often inline regardless. The meaningful change is the miss-path lookup removal, not the annotation.

**Verdict:** worthwhile cleanup that also removes a duplicate lookup on miss. LGTM.

figtracer (approved):

mablr (approved):

stevencartavia (approved):

PR #15256, opened 2026-06-21
perf: gate mapping slot recording by opcode

decofe:
Add a small opcode predicate for mapping slot recording and pass the loaded opcode into the recorder.

This avoids calling the cold mapping recorder on interpreter steps that cannot observe KECCAK256 or SSTORE while preserving the existing recording behavior for relevant opcodes.


decofe:

decofe:

decofe:

PR #15257, opened 2026-06-21
perf: inline edge coverage hooks

decofe:
Pass the opcode already loaded by the inspector step into the jump and comparison coverage helpers, and mark the small hot coverage helpers inline.

This removes duplicate opcode reads and avoids cold-code treatment on helpers that run frequently during coverage-guided invariant fuzzing.


decofe:

decofe:

decofe:

decofe:

PR #15258, opened 2026-06-21
perf: lazily build invariant inner sequence

decofe:


decofe:

DaniPopes (approved):

mablr (approved):

PR #15260, opened 2026-06-21
perf: preallocate dense edge coverage map

decofe:
Prepare collision-free edge coverage buffers before each raw EVM execution so coverage-guided runs avoid repeated small HashMap growth while collecting dense edge hits.


decofe:

decofe:

PR #15262, opened 2026-06-22
perf: use entry for counted expect calls

decofe:


decofe:

PR #15263, opened 2026-06-22
perf: serialize JSON uints directly

decofe:


decofe:

mablr (commented):
Microbench for the changed conversion path directly:
- `uint64` scalar: 19.2ns old vs 18.3ns new, ~1.05x
- max `U256` scalar: 155.4ns old vs 143.2ns new, ~1.09x
- `uint256[128]`: 2096.1ns old vs 1721.0ns new, ~1.22x

I also tried an intentionally amplified Solidity script workload with 10k scalar `vm.serializeUint` calls and 1k `uint256[128]` calls. That showed ~3% faster wall-clock time.

Given the magnitude of amplification needed to observe a slight gain, I'd consider this PR impact negligible.

decofe:
Closing during perf PR triage: the direct microbench showed only small wins, and the reviewer noted the end-to-end impact is negligible even with an amplified serializeUint workload. This does not look worth carrying as a separate open perf PR.

PR #15266, opened 2026-06-22
perf(invariant): batch throughput counters per run

decofe:


decofe:

decofe:

PR #15267, opened 2026-06-22
perf(invariant): skip zero-value balance lookup

decofe:
`execute_tx` runs once per invariant tx and always read the sender's balance to cap the requested call value, even though most generated calls carry no value (non-payable functions emit `None`, and `fuzz_msg_value` is ~85% `None`).

This skips the balance lookup unless the requested value is `Some(> 0)`, avoiding a per-tx backend/journal read (and remote account fetches on forks) in the common case. When a positive value is requested it's still capped to the sender's balance, so over-budget fuzzed values keep exercising payable paths. The computed value is otherwise unchanged.


decofe:

decofe:

mablr (approved):
nice !

figtracer (approved):
lg

0xKarl98 (approved):

stevencartavia (approved):

PR #15270, opened 2026-06-22
ci: skip rust-cache for no-default-features

decofe:
## Summary
- remove `Swatinem/rust-cache` from the `no-default-features` CI job
- keep `sccache` in place so rustc compile output caching remains enabled
- leave other workflow rust-cache uses unchanged

## Rationale
The `no-default-features` job can fail before `cargo build` starts when the rust-cache restore exhausts runner disk space. This job only needs to build with no default features, and `sccache` already covers the high-value rustc compilation cache.

## Test
- not run locally; workflow-only change

Prompted by: @grandizzy


mattsse (approved):

mablr (approved):

figtracer (approved):

mattsse (approved):
LGTM

PR #15271, opened 2026-06-22
ci(release): skip duplicate immutable nightly releases

figtracer:
## Summary

- Skip nightly release work when the computed immutable nightly release is already published for the current SHA.
- Gate the downstream Docker, binary release, and publish jobs on the new `prepare` skip output.
- Make `create-tag.js` idempotent for existing tags that already point at the current commit, independent of GitHub's 422 message text.

## Root cause

Nightlies are now immutable `nightly-${GITHUB_SHA}` releases. When `master` does not move between two scheduled runs, the second run computes the same tag as the previous successful nightly. The release is already published, so the workflow cannot upload more assets to it.

### DIFF
--- a/.github/scripts/create-tag.js
+++ b/.github/scripts/create-tag.js
@@ -7,9 +7,17 @@ module.exports = async ({ github, context }, tagName) => {
             sha: context.sha,
         });
     } catch (err) {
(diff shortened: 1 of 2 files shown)


grandizzy (approved):

mattsse (approved):
LGTM

mablr (approved):

PR #15272, opened 2026-06-22
feat(debugger): jump to active buffer offsets

figtracer:
## Summary
- Add an `o` shortcut in the debugger TUI to jump the active memory/calldata/returndata pane to a byte offset.
- Keep the existing program-counter parser and terminology intact while parsing buffer offsets separately.
- Compact the debugger footer into three non-wrapping shortcut lines and add coverage for offset parsing, prompt handling, empty buffers, and calldata jumps.

Refs #927.

### DIFF
--- a/crates/debugger/src/tui/context.rs
+++ b/crates/debugger/src/tui/context.rs
@@ -68,6 +68,8 @@ pub(crate) struct TUIContext<'a> {
     pub(crate) key_buffer: String,
     /// Current goto program counter prompt contents, if the prompt is active.
     pub(crate) pc_input: Option<String>,
+    /// Current active-buffer byte offset prompt contents, if the prompt is active.
+    pub(crate) buffer_offset_input: Option<String>,
     /// Current opcode search prompt contents, if the prompt is active.
     pub(crate) opcode_search_input: Option<String>,
     /// Last opcode search term, used by repeat-search shortcuts.
@@ -95,6 +97,7 @@ impl<'a> TUIContext<'a> {
 
             key_buffer: String::with_capacity(64),
             pc_input: None,
+            buffer_offset_input: None,
             opcode_search_input: None,
             last_opcode_search: None,
             status: None,
@@ -164,9 +167,17 @@ impl<'a> TUIContext<'a> {
 
     fn active_buffer(&self) -> &[u8] {
         match self.active_buffer {
-            BufferKind::Memory => self.current_step().memory.as_ref().unwrap().as_bytes(),
+            BufferKind::Memory => self.current_step().memory.as_ref().map_or(&[], |m| m.as_bytes()),
             BufferKind::Calldata => &self.debug_call().calldata,
             BufferKind::Returndata => &self.current_step().returndata,
+        }
+    }
+
+    pub(crate) const fn active_buffer_name(&self) -> &'static str {
+        match self.active_buffer {
+            BufferKind::Memory => "memory",
+            BufferKind::Calldata => "calldata",
+            BufferKind::Returndata => "returndata",
         }
     }
 }
@@ -191,6 +202,11 @@ impl TUIContext<'_> {
 
         if self.pc_input.is_some() {
             self.handle_pc_input_key_event(event);
+            return ControlFlow::Continue(());
+        }
+
+        if self.buffer_offset_input.is_some() {
+            self.handle_buffer_offset_input_key_event(event);
             return ControlFlow::Continue(());
         }
 
@@ -325,6 +341,13 @@ impl TUIContext<'_> {
                 self.pc_input = Some(String::new());
             }
 
+            // Go to byte offset in the active buffer
+            KeyCode::Char('o') => {
+                self.key_buffer.clear();
+                self.status = None;
+                self.buffer_offset_input = Some(String::new());
+            }
+
             // Search opcodes in the current call
             KeyCode::Char('/') => {
                 self.key_buffer.clear();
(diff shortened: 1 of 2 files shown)


crates/debugger/src/tui/context.rs
```diff
@@ -387,6 +410,32 @@ impl TUIContext<'_> {
...
+                }
+            }
+            KeyCode::Char(c)
+                if !event.modifiers.contains(KeyModifiers::CONTROL)
+                    && (c.is_ascii_hexdigit() || matches!(c, 'x' | 'X' | ':')) =>
+            {
+                if let Some(input) = &mut self.buffer_offset_input {
+                    input.push(c);
+                }
+            }
```
mablr: The buffer-offset prompt now duplicates the PC prompt’s numeric input handling. Since both modes intentionally share Esc/Enter/Backspace/numeric-key behavior, maybe consider extracting a small shared helper ?

mablr (dismissed):
LGTM, non-blocking suggestion

crates/debugger/src/tui/context.rs
```diff
@@ -737,6 +822,32 @@ fn parse_pc(input: &str, radix: u32, original: &str) -> Result<usize, String> {
...
+        return Err("Enter a buffer offset".to_string());
+    }
+
+    let (digits, radix) =
+        if let Some(rest) = input.strip_prefix("0x").or_else(|| input.strip_prefix("0X")) {
+            (rest, 16)
+        } else if let Some(rest) = input.strip_prefix("d:").or_else(|| input.strip_prefix("dec:")) {
+            (rest, 10)
+        } else {
+            (input, 10)
```
grandizzy: 🔴 **blocker:** bare input parses as **decimal** here (`(input, 10)`), but the buffer pane displays offsets as bare **hex** (`40|`, `20|`). So `o 40 Enter` jumps to decimal 40 (= 0x28 → line 0x20), not the visible `0x40`. This contradicts the on-screen labels and the PC prompt's bare-hex convention. Compounded by the key filter at L429 (`is_ascii_hexdigit() || x/X/:`), which lets users type hex-looking input like `2a`, `a`, `2x3` that the parser then rejects. Please make bare input **hex** to match the labels, and update the footer/error text and tests (e.g. `hex: 0x20/20 | decimal: d:32`).

crates/debugger/src/tui/context.rs
```diff
@@ -387,6 +410,32 @@ impl TUIContext<'_> {
...
+                self.goto_buffer_offset_from_input(&input);
+            }
+            KeyCode::Backspace => {
+                if let Some(input) = &mut self.buffer_offset_input {
+                    input.pop();
+                }
+            }
+            KeyCode::Char(c)
+                if !event.modifiers.contains(KeyModifiers::CONTROL)
+                    && (c.is_ascii_hexdigit() || matches!(c, 'x' | 'X' | ':')) =>
```
grandizzy: 🟡 non-blocking: this filter accepts `x`/`X`/`:` and hex digits anywhere, so `2x3`, `1:2`, `DEC:32` are typeable but rejected by the parser. Minor; partially resolved once the radix decision above is settled — consider tightening, or just accept that prompts allow invalid-while-typing.

crates/debugger/src/tui/context.rs
```diff
@@ -164,11 +167,19 @@ impl<'a> TUIContext<'a> {
 
     fn active_buffer(&self) -> &[u8] {
         match self.active_buffer {
-            BufferKind::Memory => self.current_step().memory.as_ref().unwrap().as_bytes(),
+            BufferKind::Memory => self.current_step().memory.as_ref().map_or(&[], |m| m.as_bytes()),
```
grandizzy, line 170: 🟡 non-blocking: nice latent-panic fix here (`map_or(&[], ...)`), but it's only partial — `draw_buffer` (crates/debugger/src/tui/draw.rs:566) still does `step.memory.as_ref().unwrap()`. If `memory: None` is reachable during real rendering, the UI panics there before the new "Current memory buffer is empty" error can show. Make it consistent if the PR intends to handle missing memory.

crates/debugger/src/tui/context.rs
```diff
@@ -520,6 +569,39 @@ impl TUIContext<'_> {
...
+            self.set_error(format!(
+                "{buffer_name} offset 0x{offset:x} ({offset}) is outside the {buffer_len}-byte buffer"
+            ));
+            return;
+        }
+
+        self.apply_buffer_offset(offset);
+    }
+
+    fn apply_buffer_offset(&mut self, offset: usize) {
```
grandizzy, line 565: 🟡 non-blocking (pre-existing inconsistency): this jump uses `offset / 32` and can reach the final partial line, but the Ctrl+J scroll clamp at L235 uses `max_buf = (len() / 32).saturating_sub(1)`, which can't reach the last partial line for non-32-multiple buffers. Consider aligning both on `len().div_ceil(32).saturating_sub(1)`. Out of scope but adjacent.

crates/debugger/src/tui/draw.rs
```diff
@@ -197,9 +197,12 @@ impl TUIContext<'_> {
...
     fn footer_height(&self) -> u16 {
         let status_or_input = u16::from(
-            self.pc_input.is_some() || self.opcode_search_input.is_some() || self.status.is_some(),
+            self.pc_input.is_some()
+                || self.buffer_offset_input.is_some()
+                || self.opcode_search_input.is_some()
+                || self.status.is_some(),
         );
-        let shortcuts = if self.show_shortcuts { 2 } else { 0 };
+        let shortcuts = if self.show_shortcuts { 3 } else { 0 };
```
grandizzy, line 205: 🟡 non-blocking: footer shortcuts grow 2→3 lines (status/input mode 3→4 total). Fine at `min_height = 16` but slightly more cramped on the smallest accepted terminals — worth a quick manual check; auto-hiding shortcuts on very small heights would be a reasonable follow-up.

crates/debugger/src/tui/context.rs
```diff
@@ -520,6 +569,39 @@ impl TUIContext<'_> {
         self.set_info(format!("{action} PC 0x{pc:x} ({pc}) in {scope}"));
     }
 
+    fn goto_buffer_offset_from_input(&mut self, input: &str) {
```
grandizzy, line 539: 🟡 non-blocking: re: mablr's shared-helper suggestion — agree it's not worth blocking. A generic Esc/Enter/Backspace/append helper likely ends up more complex than the small readable duplication, since the submit paths call different `&mut self` methods. Fine as a follow-up if desired.

Also suggest adding a partial-last-line test (e.g. 65-byte buffer, `0x40` → startline 2) and a visible-label test once the bare-radix decision is settled.

grandizzy (changes requested):
Please check findings below

grandizzy (approved):
LGTM

mablr (approved):

PR #15273, opened 2026-06-22
test(chisel): relax REPL expect timeout

figtracer:
## Summary
- Relax the chisel REPL test harness expect timeout from 3s to 10s.
- The #15200 post-merge CI failed both Linux test-all jobs in `repl::chisel_can_run_with_live_logs_flag` after timing out while waiting for the prompt/output.
- Keep the change scoped to the test harness so production chisel behavior is unchanged.

Refs #15200.

## Tests
- `cargo +1.96.0 test -p chisel --test it repl::chisel_can_run_with_live_logs_flag --locked`
- Repeated 3 warmed reruns of the same command
- `git diff --check`
- `cargo +1.96.0 fmt --all --check`


grandizzy (approved):
makes sense

mablr (approved):

PR #15276, opened 2026-06-22
fix(corpus): skip unreadable corpus files instead of failing the campaign

mark0-cn:
## Motivation

Invariant corpus entries are persisted non-atomically (`write_json_file` / `write_json_gzip_file`). If a run is killed mid-write — or a corpus directory otherwise ends up with a corrupt/truncated `.json` / `.json.gz` entry — the next campaign aborts on startup: `read_tx_seq()?` propagates the decode error out of both the disk-load loop (`WorkerCorpus::load_from_disk`) and the worker sync-import loop (`load_sync_corpus`). A single bad file in a shared or long-lived corpus directory thus bricks every future run that points at it.

## Solution

Log a warning and skip the unreadable entry instead of propagating the error, so the rest of the corpus still loads. This mirrors how the loops already tolerate empty entries, and how `read_corpus_dir` already tolerates unparseable filenames.

Added a unit test (`corrupt_corpus_file_surfaces_as_error_for_load_to_skip`) asserting that a corrupt file (valid corpus name, garbage content) is surfaced by the directory scan as a per-entry read error alongside a valid entry — the exact condition the load/sync loops now skip.

> The end-to-end skip happens inside `load_from_disk`, which needs a full executor to replay sequences, so the test covers the read-error precondition directly. Happy to add an executor-level regression test if that's preferred.


figtracer (dismissed):
lgtm

mablr (dismissed):
Makes sense, I think doesn't hurt.

mablr:
@mark0-cn could you please fix CI ? (eg. typos...)

mark0-cn:
> @mark0-cn could you please fix CI ? (eg. typos...)

sure

mark0-cn:
@figtracer @mablr I fixed it

mablr (approved):

figtracer (approved):

mark0-cn:
@mablr  can merge?

PR #15281, opened 2026-06-22
fix(invariant): preserve terminal input for optimization sequences

decofe:
Restores terminal input generation for invariant runs so optimization best-sequence reporting keeps the same sequence shape expected by warp-aware shrinking tests. The empty-input corpus guard added in #15247 is left intact, so the original panic case remains covered.

Prompted by: @grandizzy

Verification:
- cargo test -p forge --test cli invariant_optimization_with_warp -- --nocapture
- cargo test -p foundry-evm empty_input_sequence_with_new_coverage_does_not_panic_or_insert -- --nocapture
- cargo fmt --check


decofe:
Closing because the fix was pushed directly to #15247 as requested.

PR #15282, opened 2026-06-22
perf(invariant): index target events for log decoding

figtracer:
## Summary
- Adds a per-target event lookup keyed by `(topic0, indexed_topic_count)` for invariant fuzz dictionary log decoding.
- Preserves the previous `abi.events()` decode priority by merge-walking matching non-anonymous events with anonymous events in flattened ABI order.
- Still inserts raw log topics/data when no ABI event decodes.

## Motivation

`FuzzDictionary::insert_logs_values` decoded each emitted log by trying every event in the target ABI until one succeeded. Event-heavy invariant targets pay `logs * abi_events` failed `decode_log` attempts even though non-anonymous logs already carry the event selector in topic0.

This mirrors the lookup shape used by the trace decoder and makes the common non-anonymous path decode only the relevant candidate events. Anonymous events keep their prior relative priority, so an earlier anonymous event can still win over a later non-anonymous selector match.

## Benchmark

Synthetic invariant fixture:

- 80 ABI events before the emitted event.
- 8 matching logs emitted per handler call.
- `[invariant] runs = 80`, `depth = 700`, fixed seed `1337`, `-j 1`.
- Baseline binary: `67f249d241dfee7264d523e131f82639eb236174`.
- Candidate: this branch.

10 alternating measurement pairs, flipping order each pair:

| Binary | Mean | Median | Min | Max |
| baseline | 7.3823s | 7.3772s | 7.3501s | 7.4111s |
| candidate | 1.0013s | 0.9991s | 0.9857s | 1.0152s |

Mean speedup: 7.37x, 86.4% wall-time reduction.


grandizzy:
this looks similar with https://github.com/foundry-rs/foundry/pull/15246 ?

figtracer:
you’re right

crates/evm/fuzz/src/strategies/state.rs
```diff
@@ -332,20 +334,30 @@ impl FuzzDictionary {
...
-                        samples.extend(decoded_event.body);
-                        log_decoded = true;
-                        break;
-                    }
+            if let Some(contract) = target_contract {
+                if let Some(selector) = log.topics().first()
+                    && let Some(events) =
+                        contract.event_lookup.by_topic(selector, log.topics().len() - 1)
+                {
+                    log_decoded = Self::decode_log_events(events, log, &mut samples);
```
mablr: This seems to change decode priority for anonymous events. Previously `abi.events()` was tried in ABI order, so an anonymous event declared before a non-anonymous event could decode a log whose first indexed topic happens to equal another event’s selector. Now the non-anonymous lookup always wins first, and the anonymous fallback is skipped after any successful selector match?

mablr (commented):
Please double-check the possible behavior change

mablr (dismissed):
sgtm

stevencartavia (dismissed):
just needs the conflicts

mablr (approved):

stevencartavia (approved):

PR #15283, opened 2026-06-22
feat(corpus): add sequence crossover mutators

0xKarl98:
Towards OSS-187 

Adds `CrossoverInsert` and `CrossoverReplace` mutation strategies for invariant corpus sequences.

  - `CrossoverInsert` inserts a bounded contiguous chunk from another corpus entry into the selected base sequence.
  - `CrossoverReplace` replaces part of the selected base sequence with a bounded chunk from another corpus entry.
  - Donors are selected only from other non-empty in-memory corpus entries, keeping mutation on the hot path free of disk I/O and extra corpus state.

This keeps crossover behavior focused on sequence-level corpus recombination and includes unit coverage for donor
selection, insert crossover, replace crossover, and `new_inputs` integration.

### DIFF
--- a/crates/evm/evm/src/executors/corpus.rs
+++ b/crates/evm/evm/src/executors/corpus.rs
@@ -44,7 +44,7 @@ use alloy_json_abi::Function;
 use alloy_primitives::{Address, Bytes, I256, U256};
 use eyre::{Result, eyre};
 use foundry_common::{ContractsByAddress, ContractsByArtifact, TestFunctionExt, sh_warn};
-use foundry_config::FuzzCorpusConfig;
+use foundry_config::{FuzzCorpusConfig, FuzzCorpusMutationWeights};
 use foundry_evm_core::{constants::CALLER, evm::FoundryEvmNetwork, utils::StateChangeset};
 use foundry_evm_fuzz::{
     BasicTxDetails, CallDetails, ObservedCall,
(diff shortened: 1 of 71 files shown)


grandizzy:
derek bench invariant

mattsse (dismissed):
LGTM

mablr (dismissed):
I don’t see any blockers on the code side.

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1101,6 +1181,28 @@ impl WorkerCorpus {
...
+                    );
+
+                    if let Some(donor) = self.select_crossover_donor(corpus.uuid, rng) {
+                        let sequence = if is_insert {
+                            Self::crossover_insert_sequence(&corpus.tx_seq, donor, rng)
+                        } else {
+                            Self::crossover_replace_sequence(&corpus.tx_seq, donor, rng)
+                        };
+                        if let Some(sequence) = sequence {
+                            self.current_mutated = Some(corpus.uuid);
```
stevencartavia: #15253 removed `current_mutated` (now `current_mutated_index: Option<usize>`), but the crossover path still writes the old field:

```rust
self.current_mutated = Some(corpus.uuid);
```

Should this be rebased/updated?

0xKarl98: Have rebased ty

mablr (dismissed):

figtracer (dismissed):

0xKarl98:
derek bench invariant

0xalpharush (changes requested):
I don't think doing in-memory is all the different from splice. These are supposed to use donors from disk like https://github.com/0xalpharush/foundry/pull/11

0xKarl98:
superseded by #15365

PR #15288, opened 2026-06-22
feat(symbolic): minimize counterexample artifacts

figtracer:
## Summary

Deterministic, replay-gated ABI-argument minimization for stateless single-call symbolic counterexamples. Emits original/minimized artifacts when the concrete replay has a stable failure reason; failure-flag-only counterexamples are replayed but not minimized because they lack enough failure identity to distinguish assertion sites.

Also keeps counterexample search moving past undecidable hard-arithmetic branch checks: replay-confirmed failures still win, while proof-only runs remain incomplete if an undecidable branch is the best result.

## Echidna Parity

Compared with `crytic/echidna@6d856e0` and `echidna 2.3.2`.

Matches applicable single-call ABI shrinking: numeric lowering, bool false, address zero/deadbeef, bytes/string/array deletion and contiguous slices, recursive array/tuple/struct shrinking.

Added unit/CLI examples adapted from Echidna fixture shapes that actually shrink: `values/darray.sol`, `abiv2/Dynamic.sol`, `abiv2/MultiTuple.sol`, `addressarrayutils/AddressArrayUtils_withHasDuplicateBug.sol`, and symbolic fixed-array/arithmetic relation fixtures.

Side-by-side rows below use single-call assertion harnesses, `--seq-len 1`, `--workers 1`, `--shrink-limit 5000`, pinned Echidna seed, and a Foundry release build. Speed delta is `(echidna_mean / foundry_mean - 1) * 100`. Rows with `minimized == original` are intentionally omitted.

| Fixture | Foundry minimized | Echidna minimized | Final result | Foundry | Echidna | Foundry faster |
| --- | --- | --- | --- | ---: | ---: | ---: |
| `AddressArrayUtils` duplicate | `[0x00, 0x00]` | `[0x0, 0x0]` | exact match, normalized | `211.7 ms +/- 9.4 ms` | `508.8 ms +/- 19.8 ms` | `2.4x` (`+140%`) |
| `uint256[3]`, `xs[0] == xs[2] + 1` | `[1, 0, 0]` | `[4370001, 112002286318682476054167386428588068701460580937039173544089718020404258470546, 4370000]` | Foundry smaller | `109.9 ms +/- 1.6 ms` | `389.6 ms +/- 16.8 ms` | `3.5x` (`+254%`) |
| `uint256`, `(x >> 5) / 7777 >= 2222` | `552975808` | `553425558` | Foundry smaller | `67.5 ms +/- 0.5 ms` | `373.8 ms +/- 10.6 ms` | `5.5x` (`+454%`) |

The `AddressArrayUtils` row shrinks calldata from 260 bytes to 132 bytes in Foundry and reaches the same final witness as Echidna. The numeric relation rows are not byte-identical parity: Foundry minimizes further under the same shrink limit.

Sequence/no-call shrinking is OSS-351. Call-environment shrinking for symbolic single calls is out of scope until symbolic execution models those values.

### DIFF
--- a/crates/forge/src/runner.rs
+++ b/crates/forge/src/runner.rs
@@ -1102,6 +1102,10 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
         call: &SymbolicCounterexampleCall,
         expected_reason: Option<&str>,
     ) -> Result<(RawCallResult<FEN>, Option<String>), String> {
+        let Some(expected_reason) = expected_reason else {
+            return Err("candidate replay has no stable failure reason to compare".to_string());
+        };
+
         let mut executor = self.clone_executor();
         let raw_call_result = execute_tx(&mut executor, &call.to_basic_tx_details())
             .map_err(|err| err.to_string())?;
@@ -1115,10 +1119,10 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
         }
 
         let reason = self.symbolic_raw_call_failure_reason(&raw_call_result)?;
-        if reason.as_deref() != expected_reason {
+        if reason.as_deref() != Some(expected_reason) {
             return Err(format!(
                 "candidate replay failed with different reason: expected `{}`, got `{}`",
-                expected_reason.unwrap_or(""),
+                expected_reason,
                 reason.as_deref().unwrap_or("")
             ));
         }
--- a/crates/forge/src/runner.rs
+++ b/crates/forge/src/runner.rs
@@ -1398,18 +1398,19 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
                     let mut final_reason = reason;
                     let mut minimization = None;
 
-                    if let Some(candidate) = minimize_single_call_counterexample(
-                        func,
-                        &original_call,
-                        std::slice::from_ref(&self.sender),
-                        self.tcfg.config.invariant.shrink_run_limit as usize,
-                        |candidate| {
-                            self.symbolic_single_call_preserves_failure(
-                                candidate,
-                                final_reason.as_deref(),
-                            )
-                        },
-                    ) {
+                    if final_reason.is_some()
+                        && let Some(candidate) = minimize_single_call_counterexample(
+                            func,
+                            &original_call,
+                            self.tcfg.config.invariant.shrink_run_limit as usize,
+                            |candidate| {
+                                self.symbolic_single_call_preserves_failure(
+                                    candidate,
+                                    final_reason.as_deref(),
+                                )
+                            },
+                        )
+                    {
                         if candidate.changed() {
                             match self.replay_confirmed_symbolic_single_call(
                                 &candidate.minimized_call,
--- a/crates/forge/src/runner.rs
+++ b/crates/forge/src/runner.rs
@@ -1454,7 +1454,7 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
                         &signature,
                         &symbolic_result,
                         SymbolicCounterexampleArtifactKind::SingleCall,
-                        vec![final_call.clone()],
+                        vec![final_call],
                     );
                     if let Some(artifact) = minimized_artifact.clone() {
                         symbolic_result = symbolic_result.with_artifact(artifact);
@@ -3079,7 +3079,7 @@ fn validate_single_call_symbolic_replay(
             call.target, test_address
         ));
     }
-    if !call.calldata.get(..4).is_some_and(|selector| func.selector() == selector) {
+    if call.calldata.get(..4).is_none_or(|selector| func.selector() != selector) {
         return Err(format!(
             "single-call symbolic artifact calldata does not match `{}` selector",
             func.signature()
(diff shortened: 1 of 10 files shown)


crates/forge/src/runner.rs
```diff
@@ -1088,6 +1089,65 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+
+        let reason = self.symbolic_raw_call_failure_reason(&raw_call_result)?;
+        if let Some(expected) = expected_reason
+            && reason.as_deref() != Some(expected)
+        {
+            return Err(format!(
+                "candidate replay failed with different reason: expected `{expected}`, got `{}`",
+                reason.as_deref().unwrap_or("")
+            ));
+        }
```
grandizzy: **blocking** — The reason-equality check is guarded by `if let Some(expected) = expected_reason`, so when the original failure didn't revert (`reason == None`, set at `runner.rs:1310` via `Ok(res) => (res.raw, None)`), this gate is skipped entirely. The minimizer will then accept any candidate that is merely "not successful" — including one that now reverts with a *different* reason — breaking the "same failure" invariant. Reachable for failure-flag style failures (`failed()` / forge-std assertion mode) rather than `assert(false)` reverts.

```suggestion
        if reason.as_deref() != expected_reason {
            return Err(format!(
                "candidate replay failed with different reason: expected `{}`, got `{}`",
                expected_reason.unwrap_or(""),
                reason.as_deref().unwrap_or("")
            ));
        }
```

Please also add a regression test where the original failing call does not revert (`reason == None`) and a shrink candidate reverts — it must be rejected.

crates/forge/src/symbolic_minimizer.rs
```diff
@@ -0,0 +1,1924 @@
...
+    let mut try_args = |candidate_args: &[DynSolValue]| {
+        let Some(candidate_call) = call_with_args(function, call, candidate_args) else {
+            return false;
+        };
+        if candidate_call.calldata == current_call.calldata {
+            return false;
+        }
+        if attempts >= MAX_MINIMIZATION_ATTEMPTS {
+            return false;
+        }
```
grandizzy: **perf** — The attempt-cap check runs *after* `call_with_args` ABI-encodes the candidate. Once the 5000 budget is exhausted you still pay vector clones + ABI encoding for every remaining candidate. Move the cap check first:

```suggestion
        if attempts >= MAX_MINIMIZATION_ATTEMPTS {
            return false;
        }
        let Some(candidate_call) = call_with_args(function, call, candidate_args) else {
            return false;
        };
        if candidate_call.calldata == current_call.calldata {
            return false;
        }
```

crates/forge/src/symbolic_minimizer.rs
```diff
@@ -0,0 +1,1924 @@
+use crate::result::SymbolicCounterexampleCall;
+use alloy_dyn_abi::{DynSolValue, JsonAbiExt};
+use alloy_json_abi::Function;
+use alloy_primitives::{Address, B256, Bytes, Function as SolFunction, I256, U256};
+use itertools::Itertools;
+
+const MAX_MINIMIZATION_ATTEMPTS: usize = 5000;
```
grandizzy: **perf/config** — Each candidate triggers a full `clone_executor()` + `execute_tx`. Worst case is 5000 full EVM replays, which is heavy for fork/large-setup tests. The PR description references `--shrink-limit`, but this is a hard-coded constant — consider wiring it to the existing shrink-limit config instead of a literal.

grandizzy (changes requested):
One blocking soundness bug in the failure-reason gate, plus a couple of perf nits below.

grandizzy (dismissed):
lgtm!

crates/forge/src/runner.rs
```diff
@@ -1088,6 +1089,64 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+
+        let reason = self.symbolic_raw_call_failure_reason(&raw_call_result)?;
+        if reason.as_deref() != expected_reason {
+            return Err(format!(
+                "candidate replay failed with different reason: expected `{}`, got `{}`",
+                expected_reason.unwrap_or(""),
+                reason.as_deref().unwrap_or("")
+            ));
+        }
```
mablr: The minimizer still seems to treat “same failure” as “same decoded revert reason”. For non-reverting DSTest/forge-std failure-flag assertions, `symbolic_raw_call_failure_reason` returns `None`, so two different `fail()`/assertion-flag failures in the same test both satisfy `None == None`. That lets minimization accept a candidate that trips a different non-reverting assertion. Consider rejecting minimization when the original failure has no replay reason, or adding a stronger identity check for failure-flag cases.

crates/forge/src/runner.rs
```diff
@@ -1325,29 +1385,114 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                        func,
+                        &original_call,
+                        std::slice::from_ref(&self.sender),
+                        self.tcfg.config.invariant.shrink_run_limit as usize,
+                        |candidate| {
+                            self.symbolic_single_call_preserves_failure(
+                                candidate,
+                                final_reason.as_deref(),
+                            )
+                        },
```
mablr: The advertised call-environment shrinking doesn't seem reachable in the symbolic single-call path. `BaseCounterExample::from_fuzz_call` creates calls with `sender/value/warp/roll == None`, and this call site passes only `std::slice::from_ref(&self.sender)`, so `minimize_call_sender` always sees `current_idx == 0` and `value`/`warp`/`roll` minimization are no-ops. Consider wiring the actual configured sender set and concrete env fields into the counterexample, or narrowing the PR/test claims to ABI-arg minimization.

mablr (commented):
Seems good, 2 smol nits

mablr (approved):

grandizzy (approved):

PR #15299, opened 2026-06-22
feat: ERC-7201 Storage Layout

wjmelements:
This implements ERC-7201 for `forge inspect storageLayout`.
#### Motivation
I was implementing a library contract that would have delcared its storage using this annotation but the declared storage doesn't show up in the storage layout inspection.
We use this layout inspection to detect harmful regressions in the storage layout when maintaining a delegatecall proxy's implementation. 
For library functionality such as the ERC-1967 upgrade location (not implemented here) and ERC-7201 `custom-storage-location`, we would like to be able to detect and include these slots in the layout description.
Without this feature, we would have to detect all of these locations ad-hoc.
#### Changes
* use solar to collect erc-7201 entries
* add `hir_type_storage_info` to expand entries for the storageLayout inspection output
* add integration tests to verify that entries show up in the storage layou


crates/forge/src/cmd/inspect.rs
```diff
@@ -427,6 +442,267 @@ fn print_table(
...
+            let (elem_label, elem_bytes, elem_slots, _) =
+                hir_type_storage_info(hir, &arr.element.kind);
+            // Try to evaluate a literal array size for precise slot counting.
+            let fixed_len: Option<u64> = arr.size.and_then(|size_expr| {
+                if let ExprKind::Lit(lit) = size_expr.kind {
+                    if let LitKind::Number(n) = lit.kind { n.try_into().ok() } else { None }
+                } else {
+                    None
+                }
+            });
```
mablr, line 610: Fixed-size arrays whose length is a constant expression seem treated as dynamic arrays here because only literal sizes are recognize. Maybe Solar's primitives could be suitable here.

crates/forge/src/cmd/inspect.rs
```diff
@@ -427,6 +442,267 @@ fn print_table(
...
+                });
+
+                types.entry(type_label.clone()).or_insert_with(|| StorageType {
+                    encoding: encoding.to_string(),
+                    key: None,
+                    label: type_label,
+                    number_of_bytes: byte_size.to_string(),
+                    value: None,
+                    other: BTreeMap::new(),
+                });
```
mablr: The synthesized `StorageType` entries drop composite metadata that Solidity includes in `storageLayout.types`: structs should have members, arrays should have base, and mappings should have key/value. Populating these fields recursively seems necessary.

mablr (changes requested):

crates/forge/src/cmd/inspect.rs
mablr: https://github.com/paradigmxyz/solar/pull/877

crates/forge/src/cmd/inspect.rs
```diff
@@ -544,31 +552,126 @@ fn hir_type_storage_info<'hir>(
...
+            }
+        }
+        TypeKind::Custom(ItemId::Struct(id)) => {
+            let hir = &gcx.hir;
+            let s = hir.strukt(*id);
+            let mut members: Vec<Value> = Vec::new();
+            pack_fields(gcx, s.fields, |var_id, field_type, fs, fo| {
+                let var = hir.variable(var_id);
+                let field_name = var.name.map(|n| n.name.as_str().to_string()).unwrap_or_default();
+                register_type_recursive(gcx, &var.ty.kind, types);
```
mablr, line 758: [Blocker] This can recurse indefinitely for valid recursive storage types, e.g. `struct Node { mapping(uint256 => Node) children; uint256 value; }`. `types.contains_key(&label)` only helps after a type is fully inserted, but struct insertion happens after walking all members, so `Node -> mapping -> Node` re-enters forever. Safety check should be added before descending into struct members.

crates/forge/src/cmd/inspect.rs
```diff
@@ -427,6 +448,366 @@ fn print_table(
...
+                    .filter(|&n: &u64| n > 0)
+            });
+            match fixed_len {
+                Some(n) => {
+                    if elem_slots == 0 {
+                        // Packable element: compute tight packing.
+                        // elements_per_slot = floor(32 / elem_bytes), minimum 1.
+                        let per_slot = (32u64 / elem_bytes).max(1);
+                        let slots = n.div_ceil(per_slot);
+                        (n * elem_bytes, slots, "inplace")
```
figtracer: for packed fixed arrays this should report occupied storage bytes. solc reports uint8[2] as 32 bytes here, but this returns 2.

crates/forge/src/cmd/inspect.rs
```diff
@@ -427,6 +448,366 @@ fn print_table(
...
+        TypeKind::Custom(ItemId::Struct(id)) => {
+            let slot_count = pack_fields(gcx, hir.strukt(*id).fields, |_, _, _, _| {});
+            (slot_count * 32, slot_count, "inplace")
+        }
+        TypeKind::Custom(ItemId::Enum(_)) => (1, 0, "inplace"),
+        TypeKind::Custom(ItemId::Udvt(id)) => {
+            let (_, bytes, slots, encoding) = hir_type_storage_info(gcx, &hir.udvt(*id).ty.kind);
+            (bytes, slots, encoding)
+        }
+        TypeKind::Function(_) => (24, 0, "inplace"),
```
figtracer: internal function values are 8 bytes in solc storage layout, so this can shift the next packed field after an internal function.

crates/forge/src/cmd/inspect.rs
```diff
@@ -427,6 +449,376 @@ fn print_table(
...
+            (bytes, slots, encoding)
+        }
+        TypeKind::Function(f) => {
+            let bytes = match f.visibility {
+                Visibility::External | Visibility::Public => 24,
+                _ => 8,
+            };
+            (bytes, 0, "inplace")
+        }
+        TypeKind::Custom(_) | TypeKind::Err(_) => (32, 1, "inplace"),
```
mablr, line 649: [Blocker] Contract/interface-typed fields fall through to `unknown`, so they are treated as 32-byte slot-boundary values instead of 20-byte packable addresses. I reproduced this with `IERC20 token; bool paused;`: the normal solc layout packs `paused` at slot 0 offset 20, but the ERC-7201 synthetic layout emits `token` as `unknown` and moves `paused` to the next slot.

Consider handling `ItemId::Contract(_)` with a solc-compatible contract label and 20-byte inplace storage info.

crates/forge/src/cmd/inspect.rs
```diff
@@ -427,6 +449,376 @@ fn print_table(
...
+        let linearized_bases: HashSet<_> =
+            hir.contract(target_id).linearized_bases.iter().copied().collect();
+        let target_contract_name = hir.contract(target_id).name.as_str().to_string();
+
+        // Walk every struct in the HIR; keep those that belong to a contract in the
+        // linearization chain and carry an @custom:storage-location erc7201:<ns> annotation.
+        for struct_id in hir.strukt_ids() {
+            let strukt = hir.strukt(struct_id);
+
+            let Some(struct_contract_id) = strukt.contract else { continue };
```
mablr: [Suggestion] This skips every `@custom:storage-location erc7201:*` struct declared at file scope, because `strukt.contract` is `None`. Solidity accepts NatSpec on file-level structs, and a library-style accessor using that struct currently compiles but `forge inspect FileStorageLib storageLayout --json` returns empty `storage`/`types`. Consider including file-level annotated structs that are referenced by the target/library path, or otherwise documenting/rejecting that pattern.

mablr (changes requested):

crates/forge/src/cmd/inspect.rs
wjmelements: I haven't ever used this annotation at file scope. Might be hard to identify consumers. I'll think about it.

wjmelements: There were a few cases to consider. I reasoned through some of them more than others. The test cases and their comments should exhibit the decisions I made. If you disagree with some of them I can change it; I don't have a strong opinion on them.

crates/forge/src/cmd/inspect.rs
wjmelements: [0c51e8a](https://github.com/foundry-rs/foundry/pull/15299/commits/0c51e8acc69a392586276833f454d83d544a7018)

crates/forge/src/cmd/inspect.rs
wjmelements: * [8183e02](https://github.com/foundry-rs/foundry/pull/15299/commits/8183e02ad4c6a0d6c73e763af0a2cf7f0463ad6b)
* [57b7abf](https://github.com/foundry-rs/foundry/pull/15299/commits/57b7abf72a8022b370c6b5f37336b5b0c1c8dde9)

crates/forge/src/cmd/inspect.rs
```diff
@@ -441,6 +466,475 @@ fn print_table(
...
+///
+/// Returns the type label used as the key in `types` and as `Storage.storage_type`.
+fn register_type_recursive<'hir>(
+    gcx: Gcx<'hir>,
+    kind: &TypeKind<'hir>,
+    types: &mut BTreeMap<String, StorageType>,
+) -> String {
+    let (label, byte_size, _, encoding) = hir_type_storage_info(gcx, kind);
+
+    if types.contains_key(&label) {
```
mablr: `label` is a display string, but it is also being used as the type ID. Distinct valid types can share it: for example, two imported file-scope structs named `Layout` can have different members, while both become `struct Layout`; function types also discard parameters/returns, and `address payable` becomes `address`. `contains_key` then reuses the first entry, so fields can point at incorrect metadata. Could we generate a unique structural/declaration ID separately from the human-readable label and use that ID for `Storage.type`, array `base`, mapping key/value, and member types? The ID does not need to exactly reproduce solc's spelling, but it must not collapse distinct types.

crates/forge/src/cmd/inspect.rs
```diff
@@ -441,6 +466,475 @@ fn print_table(
...
+/// Returns `true` if any statement in `block` references `struct_id`, recursing into nested
+/// blocks, loops, conditionals, switches, and try/catch clauses.
+fn block_references_struct<'hir>(gcx: Gcx<'hir>, block: &Block<'hir>, struct_id: StructId) -> bool {
+    block.stmts.iter().any(|stmt| stmt_references_struct(gcx, stmt, struct_id))
+}
+
+fn stmt_references_struct<'hir>(gcx: Gcx<'hir>, stmt: &Stmt<'hir>, struct_id: StructId) -> bool {
+    let hir = &gcx.hir;
+    match &stmt.kind {
+        StmtKind::DeclSingle(var_id) => {
```
mablr: This treats every local declaration of the annotated file-scope struct as evidence of namespaced storage, regardless of its data location. A valid memory-only use such as `Layout memory x = Layout({value: 1});` therefore fabricates an ERC-7201 storage entry for the inspected contract. HIR variables expose `data_location`; can we require `DataLocation::Storage` here (and for return declarations used by `function_references_struct`) and add a memory-only regression test? Broader discovery through expressions such as `Lib.layout().value++` can be documented/deferred rather than widening this heuristic.

crates/forge/src/cmd/inspect.rs
mablr: Thanks, the new tests clarify the intended behavior. I think we should keep the rule narrow for this PR: include a file-scope annotated struct when the target or a linearized base references it through a storage return or storage local. The current declaration walk does not check data location, so a memory-only declaration also synthesizes a namespace; I left an inline blocker for that.
Direct expression use such as `Lib.layout().value++` is still missed, but supporting that reliably would require typed expression/call analysis rather than broadening the declaration heuristic. I’m fine documenting or deferring that case for now.

crates/forge/src/cmd/inspect.rs
wjmelements: 6b9e91c79

crates/forge/src/cmd/inspect.rs
wjmelements: 786079a1e

mablr (commented):
@wjmelements I actually think we need a more systematic approach on foundry compiler side (foundry-core) which would make this PR much simpler and robust. Let me think about it in the coming days, I'll prepare smtg you could rebase on.

wjmelements:
> @wjmelements I actually think we need a more systematic approach on foundry compiler side (foundry-core) which would make this PR much simpler and robust. Let me think about it in the coming days, I'll prepare smtg you could rebase on.

@mablr Have you made any progress?

mablr:
I propose a step back from the current PR approach and split the responsibility more sharply:

 - Solar provides one generic primitive: compute a storage layout for any struct rooted at an arbitrary slot.
 - foundry-core owns ERC-7201 semantics: annotation discovery, inheritance traversal, root-slot computation, and conversion to solc-compatible artifacts.
 - Forge only merges the resulting namespaces into `forge inspect storageLayout` and applies duplicate/type-collision policy.

Compared with #15299, this avoids putting ERC-7201 policy and a second layout/type-generation path in Forge. Solar remains responsible for authoritative Solidity packing, while Foundry owns the protocol-specific behavior.

I vibed my idea in draft branches over 3 concerned repos:
 - https://github.com/paradigmxyz/solar/compare/main...mablr:solar:mablr/erc7201-layout
 - https://github.com/foundry-rs/foundry-core/compare/main...mablr/erc7201-layout
 - https://github.com/foundry-rs/foundry/compare/master...mablr/erc7201-layout

@wjmelements Please let me know what do you think, don't hesitate to roast or cherry-pick :)

wjmelements:
This changeset may be superceded by https://github.com/foundry-rs/foundry/pull/16734

PR #15300, opened 2026-06-22
perf(corpus): skip all-empty cmp payloads

decofe:


decofe:
Folded into #15558 for review as a single corpus mutation perf PR.

PR #15301, opened 2026-06-22
perf(invariant): avoid shrinking dropped run payloads

decofe:


DaniPopes:
Incorrect, see doc comment

PR #15302, opened 2026-06-22
perf(evm): insert first dense edge hit directly

decofe:


decofe:

PR #15303, opened 2026-06-22
perf(fuzz): gate push-byte address lookups

decofe:


decofe:
Closing during fold triage: this address guard breaks existing behavior covered by , because a reused address can later have different bytecode that still needs push-byte scanning. The folded fuzz-state PR #15557 keeps the other empty-input skips but intentionally excludes this change.

decofe:
Follow-up detail for the close rationale: the change broke `same_address_with_new_bytecode_is_scanned_again`, because a reused address can later have different bytecode that still needs push-byte scanning. The folded fuzz-state PR #15557 keeps the other empty-input skips but intentionally excludes this change.

PR #15304, opened 2026-06-22
perf(evm): skip empty broadcast clone

decofe:


DaniPopes:
Empty vecdeque clone is a noop

PR #15305, opened 2026-06-22
perf(cheatcodes): skip inactive step hooks

decofe:


DaniPopes:
needs heavy benchmarking

PR #15306, opened 2026-06-23
perf: build invariant settings in one pass

decofe:
Build invariant target contract and selector maps together, and use unstable sorting for sender address filters where sort stability is not observable.


grandizzy:

DaniPopes (approved):

mattsse (approved):
lgtm

PR #15307, opened 2026-06-23
perf: reuse invariant assertion check

decofe:
Pass the assertion-failure result computed in the invariant worker into can_continue so the failure path does not rescan the same call result and state changeset.


DaniPopes (approved):

grandizzy (approved):
makes sense

PR #15308, opened 2026-06-23
perf: move replay payloads without cloning

decofe:
Move owned replay logs, debug bytecodes, and line coverage out of RawCallResult while preserving the existing trace ownership used for reports and counterexamples.


DaniPopes (approved):

mablr (approved):

mattsse (approved):
lgtm

PR #15311, opened 2026-06-23
perf(fuzz): filter log events by selector

decofe:


mablr (commented):
Already addressed by https://github.com/foundry-rs/foundry/pull/15282

PR #15313, opened 2026-06-23
perf(fuzz): gate call override handler lookups

decofe:
Check cheap call override rejection conditions before reading the handler set, then test caller and target membership under one read lock. This trims lock traffic on the invariant call override path without changing override eligibility.


DaniPopes (approved):

mattsse (approved):
lgtm

PR #15314, opened 2026-06-23
perf(fuzz): skip empty call value collection

decofe:
Return before borrowing the invariant fuzz dictionary when a successful call produced no logs, return data, or state changes. This preserves collection semantics while trimming an empty per-call path in invariant campaigns.


decofe:

PR #15315, opened 2026-06-23
perf(fuzz): skip empty fuzzer value drain

decofe:
Return before borrowing the invariant fuzz dictionary when the fuzzer has no staged values to drain. This keeps the per-call invariant worker path out of RefCell borrow machinery for empty collections.


decofe:

PR #15316, opened 2026-06-23
[dnm] feat(invariant): random per-run depth

stevencartavia:
Adds an option to use random depths during invariant fuzz campaigns instead of a fixed depth, so each run samples a depth in `[min_depth, depth]`. 

min_depth default temporarily set to test the benchmark; revert before merge.

Closes OSS-186


stevencartavia:
derek bench invariant

PR #15322, opened 2026-06-23
perf(fuzz): decode push immediates directly

decofe:


grandizzy:
Correctness-equivalent today, but the perf benefit is unmeasurable and it trades away a deliberate panic-safety margin for nothing visible. Not a legit perf improvement

PR #15325, opened 2026-06-23
refactor(forge): isolate symbolic artifact replay options

mablr:
## Description

Addresses follow-up feedback from #15217 around symbolic artifact replay orchestration.

This removes the need to globally enable `config.symbolic.enabled` when replaying symbolic artifacts and instead threads replay-specific execution state through a dedicated `TestExecutionOptions` struct. Single-call artifact replay can still opt into symbolic entrypoint discovery, while sequence replay avoids enabling symbolic mode and its solver-related side effects.

This also reduces the `run_tests` positional argument growth by grouping execution flags into `TestExecutionOptions`, replacing the previous `#[allow(clippy::too_many_arguments)]` workaround.

## What changed

- Add `TestExecutionOptions` for test execution flags.
- Pass replay configuration through execution options instead of standalone positional args.
- Stop forcing `config.symbolic.enabled = true` for all artifact replays.
- Add `symbolic_entrypoints_enabled(...)` to scope symbolic entrypoint discovery to:
  - normal symbolic runs
  - single-call symbolic artifact replay
- Update coverage test execution to use `TestExecutionOptions::coverage()`.
- Remove now-unneeded `too_many_arguments` allowances from the test execution path.

### DIFF
--- a/crates/forge/src/cmd/coverage.rs
+++ b/crates/forge/src/cmd/coverage.rs
@@ -1,4 +1,8 @@
-use super::{install, test::TestArgs, watch::WatchArgs};
+use super::{
+    install,
+    test::{TestArgs, TestExecutionOptions},
+    watch::WatchArgs,
+};
 use crate::coverage::{
     BytecodeReporter, ContractId, CoverageReport, CoverageReporter, CoverageSummaryReporter,
     DebugReporter, ItemAnchor, LcovReporter,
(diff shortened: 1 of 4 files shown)


figtracer (approved):

grandizzy (approved):
lgtm

0xKarl98 (approved):

PR #15326, opened 2026-06-23
perf(fuzz): inline fuzzer call hooks

decofe:


grandizzy:
semantically harmless but an unsubstantiated, likely-no-op perf change

PR #15334, opened 2026-06-23
perf(invariant): avoid duplicate persisted replay

mablr:
Follow-up to #15217.

Removes the second full `replay_persisted_call_sequence` execution when replaying a persisted invariant failure.

`replay_error` now returns the shrunk counterexample plus the `check_sequence` outcome captured during shrinking, so replay metadata (`replayed_entirely`, reason, calls, reverts) can be reused without running the same sequence again.


mablr:
derek bench invariant

0xalpharush:
idt this change affects fuzzing coverage so doesn't need the bench

crates/evm/evm/src/executors/invariant/shrink.rs
```diff
@@ -231,21 +238,37 @@ pub(crate) fn shrink_sequence<FEN: FoundryEvmNetwork>(
...
+        CheckSequenceOptions {
+            accumulate_warp_roll: config.has_delay(),
+            fail_on_revert: config.fail_on_revert,
+            expect_assertion_failure,
+            call_after_invariant: invariant_contract.call_after_invariant,
+            rd,
+        },
+    )?;
+    if !zero_call_result.0 {
+        return Ok(ShrunkSequence { calls: vec![], result: Some(zero_call_result) });
```
stevencartavia: The zero-call check now runs `check_sequence(..., vec![], ...)` (shrink.rs#L251-L266). Since it passes `call_after_invariant`, an empty candidate whose body passes still runs `afterInvariant` — so if `afterInvariant` fails in the initial state, `shrink_sequence` returns `calls: []` before the shrink loop.

That empty sequence yields an empty `counterexample_sequence`, and the runner only records failures inside `!counterexample_sequence.is_empty()` arms — so nothing gets persisted. `forge test --rerun` then has no minimized sequence to replay. Repro'd with a Vault insolvent from `setUp`: master persists + replays a 1-call reproducer, this branch persists nothing.

Suggest restoring the body-only check so only the body can short-circuit to empty (`afterInvariant` then falls through to the shrink loop):

```rust
let (_, success) = call_invariant_function(executor, target_address, calldata.clone())?;
if !success {
    return Ok(ShrunkSequence { calls: vec![], result: None });
}
```

mattsse (approved):
LGTM

stevencartavia (approved):

PR #15335, opened 2026-06-23
feat(symbolic): minimize stateful counterexamples

figtracer:
## Summary

Extends symbolic minimization to stateful invariant counterexample sequences.

- drops redundant calls while preserving replay-confirmed failure identity
- ABI-minimizes remaining calldata with the single-call minimizer
- shrinks eligible senders and call values through replay-accepted candidates
- skips duplicate replay-equivalent minimizer candidates
- emits original/minimized artifacts plus sequence length/calldata metadata
- replays the minimized sequence for final traces and artifact confirmation

## Echidna parity

Echidna values are local Echidna 2.3.2 runs from 2026-06-23. Foundry timings were measured locally on 2026-06-24 with a prebuilt warm `target/release/forge`, compiled fixtures, and cleared counterexample/failure caches before each run, excluding cargo/build setup.

| Fixture | Echidna shrink | Foundry minimized artifact | Echidna | Foundry | vs Echidna |
| --- | --- | --- | ---: | ---: | ---: |
| `basic/flags.sol` | `set0(0); set1(0)` | same, sequence len 2 | 1.90s | 0.59s | -69% |
| `basic/revert.sol` | `f(-1, 0x0, 0xdeadbeef)` | same one-call failure; unused address minimized to zero, guard address remains nonzero | 0.57s | 0.38s | -33% |
| `values/darray.sol` | `[0x123456]` | same address, normalized to 20 bytes | 0.91s | 0.47s | -48% |
| `basic/darray-mutation.sol` | 17-byte `abc...` payload | same minimal length/prefix; zeroes irrelevant suffix bytes | 1.01s | 0.47s | -53% |
| `values/payable.sol` | `payable_function{value: 0x81}()` | same value-bearing call | 0.62s | 0.38s | -39% |


crates/forge/src/runner.rs
```diff
@@ -1179,6 +1209,275 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+            calls,
+            invariant_config,
+            invariant_contract,
+            target_invariant,
+            assertion_failure,
+        )
+        .is_some_and(|actual| {
+            actual.replayed_entirely == expected.replayed_entirely
+                && actual.reason == expected.reason
+        })
```
mablr: This only compares `replayed_entirely` and the decoded reason, so a minimized sequence can be accepted when a different call/assertion site fails with the same reason string. That is especially easy with `fail_on_revert` or generic assertion messages. Maybe it would safer to carry somewhat the failing site identity from `check_sequence` into `SymbolicSequenceFailure` before accepting a candidate?

mablr (dismissed):
sgtm

crates/forge/src/symbolic_minimizer.rs
0xalpharush, line 1: why do we have a separate minimizer than the fuzzer?

mattsse (dismissed):
lgtm

crates/forge/src/symbolic_minimizer.rs
mablr: Maybe the split is because the existing invariant shrinker mostly drops calls from `BasicTxDetails`, while this pass needs to generate symbolic/ABI-aware candidates: calldata reductions, sender/value reductions, duplicate candidate filtering, and artifact metadata.

It does look like the replay/acceptance loop could potentially be factored and shared, with only the symbolic-specific candidate generation staying separate.

mablr: The latest refactor https://github.com/foundry-rs/foundry/pull/15335/commits/088a724a516cae8320aaca8fcafeb6c297d0ea0a reduces the local duplication by centralizing the sequence minimizer state and candidate acceptance in `SequenceMinimizer`. It still keeps symbolic candidate generation separate, since that part is ABI/artifact-specific. I think the remaining possible shared piece would be a future factoring of the generic try/replay/accept loop with the invariant shrinker.

mablr:
@figtracer please fix conflicts 🙏

stevencartavia (approved):

mablr (approved):

PR #15336, opened 2026-06-23
feat: add last frame gas cheatcodes

decofe:
## Summary
- add `lastFrameGas` and `snapshotGasLastFrame` cheatcodes that report the most recent CALL or CREATE frame gas
- mark `lastCallGas` and `snapshotGasLastCall` as deprecated aliases
- record CREATE frame gas in the cheatcode inspector and cover CALL/CREATE/CREATE2 cases

Prompted by: @DaniPopes

## Tests
- `cargo test -p foundry-cheatcodes-spec`
- `cargo test -p foundry-cheatcodes --lib lastFrameGas --no-default-features`
- `cargo build -p forge`
- `/home/agent/branches/foundry-rs/foundry/target/debug/forge test --root /tmp/foundry-framegas-test --match-path paris/cheats/GasSnapshots.t.sol`
- `/home/agent/branches/foundry-rs/foundry/target/debug/forge test --root /tmp/foundry-framegas-test --match-path paris/cheats/LastCallGas.t.sol --match-contract LastCallGasIsolatedTest`

Note: a full `LastCallGas.t.sol` run in the temporary stripped-down project still fails the pre-existing non-isolated gas constants because the temporary root changes project isolation behavior; the isolated suite and new frame-gas coverage pass.


crates/cheatcodes/src/inspector.rs
```diff
@@ -2120,6 +2127,19 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>> for Cheatcode
...
+            // retrieve the gas usage of the last call or create.
+            let gas = outcome.result.gas;
+            self.gas_metering.last_frame_gas = Some(crate::Vm::Gas {
+                gasLimit: gas.limit(),
+                gasTotalUsed: gas.total_gas_spent(),
+                gasMemoryUsed: 0,
+                gasRefunded: gas.refunded(),
+                gasRemaining: gas.remaining(),
+            });
+        }
```
mablr: CREATE records `last_frame_gas` before expected-revert handling, while CALL returns from the expected-revert path before reaching the gas-recording block.

I reproduced this with a temporary fixture: after `vm.expectRevert(); RevertingTarget(target).fail(); vm.lastFrameGas()` still reverts when there was no prior frame, but after `vm.expectRevert(); new RevertingConstructor(); vm.lastFrameGas()` returns the CREATE gas.

mattsse (dismissed):
LGTM

crates/cheatcodes/src/inspector.rs
DaniPopes: [`d255cd4` (this PR)](https://github.com/foundry-rs/foundry/pull/15336/commits/d255cd4a427015a1c328bf0c55a114dd51102d56)

crates/cheatcodes/src/inspector.rs
```diff
@@ -884,6 +884,16 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
...
         executor: &mut dyn CheatcodesExecutor<FEN>,
     ) -> Option<CallOutcome> {
+        // `lastFrameGas` is scoped to the current top-level frame. Clear any cached value when a
+        // new non-cheatcode frame begins so a test/transaction cannot observe gas from a previous
+        // one. Cheatcode and console calls must keep the cache so `vm.lastFrameGas()` can read it.
+        if ecx.journal().depth() <= 1
+            && call.target_address != CHEATCODE_ADDRESS
+            && call.target_address != HARDHAT_CONSOLE_ADDRESS
+        {
+            self.gas_metering.last_frame_gas = None;
```
stevencartavia: **`lastFrameGas`/`snapshotGasLastFrame` return stale gas after an expected revert.**

On `expectRevert`, the frame returns before recording but the cache is never cleared. The only clear is the `depth() <= 1` one in the CALL hook, so it misses CREATE frames and nested CALLs — leaking the previous frame's gas.

```solidity
new Succeeds(1);
vm.expectRevert();
new RevertingConstructor();
vm.lastFrameGas();   // returns stale 107164 instead of reverting
```
```solidity
// nested in a helper
ok.ping();
vm.expectRevert(); bad.fail();
vm.lastFrameGas();   // returns stale 22491 instead of reverting
```

The equivalent top-level CALL correctly reverts, and `snapshotGasLastFrame` writes the stale value into the snapshot. This contradicts `d255cd4` and the guarantee `testExpectedRevertCallDoesNotRecordLastFrameGas` claims (it only passes because it runs top-level with an empty cache).

Minor: deprecated `lastCallGas`/`snapshotGasLastCall` now return CREATE gas (docs/error still say "call") — intentional? And `symbolic/README.md` still lists only the old names.

mablr (approved):

crates/cheatcodes/src/inspector.rs
```diff
@@ -2175,6 +2187,19 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>> for Cheatcode
             }
         }
 
+        if curr_depth > 0 {
+            // Record the gas usage of the create frame, this allows the `lastFrameGas` cheatcode to
+            // retrieve the gas usage of the last call or create.
+            let gas = outcome.result.gas;
+            self.gas_metering.last_frame_gas = Some(crate::Vm::Gas {
```
grandizzy, line 2194: The `last_frame_gas` cache is never cleared when a new top-level test frame begins, so frames created in `setUp()` leak into the test body. A CREATE here (or a CALL recorded in `call_end`) during `setUp` stays cached, and the first `vm.lastFrameGas()` in the test body returns it instead of reverting.

```solidity
function setUp() public { new Target(); }      // CREATE frame cached
function testNoFrameYet() public {
    vm.lastFrameGas();  // succeeds with setUp's gas
                        // — but should revert "no external call or create was made yet"
}
```

`LastCallGas.t.sol` already avoids making calls in `setUp` for exactly this reason. Since this is a new API, can we scope `lastFrameGas` to the current top-level test/script frame? If fixture-scoping is intentional instead, let's document and test it explicitly.

Suggested regression test to add (`testdata/.../cheats/LastCallGas.t.sol`):

```solidity
contract LastFrameGasSetupIsolationTest is DSTest {
    Vm constant vm = Vm(HEVM_ADDRESS);

    // setUp creates a frame; it must NOT leak into the test body.
    function setUp() public {
        new Target();
    }

    function testSetupFrameDoesNotLeak() public {
        // No call or create has happened in the test body yet,
        // so this must revert rather than return setUp's gas.
        vm._expectCheatcodeRevert(
            "no external call or create was made yet"
        );
        vm.lastFrameGas();
    }
}

contract Target {
    uint256 public x;
    function poke() external { x = 1; }
}
```

(Adjust the harness base contract / revert helper to match the file's existing conventions.)

grandizzy (commented):
Pls check finding below

crates/cheatcodes/src/inspector.rs
DaniPopes: pre existing issue, it doesn't matter

figtracer (approved):

PR #15343, opened 2026-06-24
Monad Foundry

haythemsellami:
## Motivation

This PR adds first-class Monad execution support to Foundry while preserving the
existing Ethereum, OP Stack, Celo, and Tempo execution paths.

Monad has protocol-level EVM differences that cannot be modeled only through the
standard Ethereum `SpecId`. The main goal of this PR is to make those differences
available through Foundry's existing network abstraction so `forge`, `anvil`,
`cast`, `chisel`, traces, and cheatcodes can select the correct EVM factory and
hardfork-specific behavior.

The intended user-facing selection model follows the upstream network API:

- `--network monad`
- `--hardfork MonadNine`
- `hardfork = "monad:MonadNine"` in `foundry.toml`
- network-qualified `vm.setEvmVersion(...)` values for tests

### Network and hardfork plumbing

- Adds Monad to the generic network/hardfork layer in `foundry-evm-networks` and
  `foundry-evm-hardforks`.
- Extends `FoundryHardfork` with a Monad variant backed by `MonadHardfork`.
- Supports network-qualified hardfork parsing so Monad hardfork names can coexist
  with standard Ethereum and Tempo hardfork names.
- Wires Monad network selection through shared CLI/config handling instead of a
  tool-specific special case.

### EVM integration

- Adds `MonadEvmNetwork` and `MonadEvmFactory` integration in `foundry-evm-core`.
- Dispatches Monad execution through `alloy-monad-evm` / `monad-revm` while
  leaving the default Ethereum factory path unchanged.
- Threads Monad-aware env, backend, nested EVM, inspector, and trace paths through
  the existing generic EVM interfaces.
- Preserves non-Monad behavior by dispatching only when the active network config
  is Monad.

### Tooling coverage

This PR wires Monad through the main Foundry tools:

- `forge test`, `forge build`, gas reports, and script execution
- `anvil` local node configuration, fork inference, block tags, tx validation,
  and execution backend
- `cast call`, `cast run`, `cast da-estimate`, and bytecode verification paths
- `chisel` REPL network selection
- trace decoding for Monad precompile calls

### Cheatcodes and precompiles

- Adds a Monad cheatcode module and a Monad cheatcode address.
- Adds Monad staking helper cheatcodes and corresponding test fixtures.
- Adds Monad precompile trace decoding so traces are readable instead of showing
  only raw precompile calls.

### Dependency status

This branch uses the published crates.io releases `alloy-monad-evm` 0.5.0 and
`monad-revm` 0.5.1. `alloy-monad-evm` is enabled with `std` and
`memory_limit`; `monad-revm` is enabled with `std`, `serde`, and
`memory_limit`. The lockfile resolves both from crates.io, with no Monad-specific
Git or `[patch.crates-io]` override. These releases contain the Foundry/revm 41
compatibility work from category-labs/monad-revm#34 and
category-labs/alloy-monad-evm#14.

## Testing

Added upstream-style tests in the Foundry repo covering:

- Monad hardfork parsing and network config validation
- `forge test` network/hardfork selection
- `forge build` Monad contract-size handling
- `anvil` Monad config, execution, fork-mode inference, block tags, and tx limits
- `cast call` and `cast run` Monad dispatch paths
- `chisel` Monad network selection
- Monad staking cheatcodes and fixtures
- Monad trace/precompile decoding paths

Local validation performed on this branch:

- `cargo check -p foundry-evm-hardforks --lib`
- `cargo check -p foundry-evm-networks --lib`
- `cargo check -p foundry-evm-core --lib`
- `cargo check -p forge -p cast@1.7.2 -p anvil -p chisel --bins`
- `cargo fmt --all`

E2E integration tests can be found here https://github.com/haythemsellami/monad-foundry-integration/tree/monad-v1.7.1-integration-coverage

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/script/src/lib.rs
```diff
@@ -647,6 +651,20 @@ impl ScriptArgs {
         Ok(())
     }
 
+    fn contract_size_limits(&self) -> ContractSizeLimits {
```
figtracer: this helper is unused by the broadcast check above, so monad scripts still fall back to the ethereum size limit before prompting on valid deployments.

crates/cheatcodes/src/evm.rs
```diff
@@ -1183,16 +1183,16 @@ impl Cheatcode for executeTransactionCall {
         ccx.ecx.tx_mut().set_gas_priority_fee(None);
 
         // Enable nonce checks for realistic simulation.
-        ccx.ecx.cfg_mut().disable_nonce_check = false;
+        ccx.ecx.cfg_env_mut().disable_nonce_check = false;
 
         // EIP-3860: enforce initcode size limit.
-        ccx.ecx.cfg_mut().limit_contract_initcode_size =
+        ccx.ecx.cfg_env_mut().limit_contract_initcode_size =
```
figtracer, line 1344: under monad this reintroduces the ethereum initcode cap, so a raw deploy that the network accepts can be rejected by this cheatcode path.

crates/cast/src/cmd/da_estimate.rs
```diff
@@ -39,6 +39,7 @@ impl DAEstimateArgs {
         match network {
             NetworkVariant::Optimism => da_estimate::<Optimism>(&config, block).await,
             NetworkVariant::Ethereum => da_estimate::<Ethereum>(&config, block).await,
+            NetworkVariant::Monad => da_estimate::<Ethereum>(&config, block).await,
             NetworkVariant::Tempo => Err(eyre::eyre!(
                 "DA estimation is not supported for Tempo: EIP-4844 blob transactions are not available on this network"
```
mablr: Monad docs explicitly say type-3/EIP-4844 blob transactions are unsupported. Consider returning an unsupported-network error, like Tempo.

mablr (changes requested):

crates/evm/traces/src/decoder/mod.rs
```diff
@@ -287,6 +298,10 @@ impl CallTraceDecoder {
                 .chain(ITIP20ChannelReserve::abi::functions().into_values())
                 .chain(ISignatureVerifier::abi::functions().into_values())
                 .chain(IReceivePolicyGuard::abi::functions().into_values())
+                // Monad
+                .chain(IMonadStaking::abi::functions().into_values())
+                .chain(IMonadStakingSyscalls::abi::functions().into_values())
+                .chain(IReserveBalance::abi::functions().into_values())
```
mablr: The Monad trace decoder registers function/event ABIs from local copied `.sol` interfaces, but the regression tests encode staking calls using `monad_revm`’s generated interface. That catches one direction of drift, but probably if `monad_revm` changes the live ABI and the local `.sol` copy stays stale, traces will still decode with the stale local definitions.

A small selector/ABI parity test could be beneficial ?

crates/cast/src/cmd/da_estimate.rs
haythemsellami: Fixed in a378adcc5: cast da-estimate now returns an unsupported-network error for Monad instead of routing Monad to Ethereum blob DA estimation.

crates/evm/traces/src/decoder/mod.rs
haythemsellami: Fixed in 3f0b2af4a: added a Monad ABI parity test comparing the local trace decoder selectors/events against monad_revm typed SolCall/SolEvent selectors, including the staking syscall split.

crates/cheatcodes/src/evm.rs
haythemsellami: Fixed in f004b4026: executeTransactionCall now uses FEN::CONTRACT_INITCODE_SIZE_LIMIT; Monad overrides it to monad_revm::MONAD_MAX_INITCODE_SIZE, with a unit test for the override.

crates/script/src/lib.rs
haythemsellami: Fixed in 674aeb660: prepare_bundled now uses the resolved network contract size limits, so Monad scripts use the Monad size limit in the broadcast pre-simulation path with CLI/config precedence covered by tests.

crates/anvil/tests/it/monad.rs
```diff
@@ -0,0 +1,277 @@
...
+    let accounts = provider.get_accounts().await.unwrap();
+
+    let tx = TransactionRequest::default()
+        .with_from(accounts[0])
+        .with_to(accounts[1])
+        .with_value(U256::from(1))
+        .with_gas_limit(MONAD_TX_GAS_LIMIT_CAP + 1);
+    let err = provider.send_transaction(tx.into()).await.unwrap_err().to_string();
+
+    assert!(err.contains("tx.gas_limit > env.cfg.tx_gas_limit_cap"), "unexpected error: {err}");
```
mablr: This assertion is stale after the implementation changed the gas-limit error detail.

crates/evm/hardforks/src/lib.rs
mablr, line 166: Monad is not handled here, it should be imo.

crates/evm/networks/src/lib.rs
mablr: Is there something to do here for Monad ?

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -4573,11 +4658,11 @@ where
 
             // Check tx gas limit against tx gas limit cap (Osaka hard fork and later).
             if evm_env.cfg_env.tx_gas_limit_cap.is_none()
-                && tx.gas_limit() > evm_env.cfg_env().tx_gas_limit_cap()
+                && tx.gas_limit() > self.tx_gas_limit_cap(evm_env)
```
mablr, line 9234: This path now validates pool transactions with the Monad-aware resolved tx gas cap, but `EthApi::build_tx_request` still auto-fills omitted gas with the plain `evm_env.cfg_env().tx_gas_limit_cap()` fallback. That leaves `eth_sendTransaction` autofill and pool validation using different caps on Monad, so omitted-gas requests can be under-filled with Ethereum/EIP-7825’s cap.

mablr (commented):
An overall concern I have about third-party EVMs/Networks support in Foundry, is the alloy/revm bumps overheads and frictions. We experienced painful OP deps bumps while preparing Alloy 2.0 release, blocking/slowing us on Tempo development/integration. That's why we feature gated OP, and I'm wondering if we should do the same for Monad ? (This is mostly mechanical) If something breaks on Monad it should not impact Tempo's CI hotpath. WDYT?

-> some inline comment

crates/anvil/src/eth/backend/mem/mod.rs
haythemsellami: Fixed in 57943508a: `eth_sendTransaction` gas autofill now uses the backend-resolved fallback tx gas limit, so Monad omitted-gas fallback and pool validation share the same Monad-aware cap. Added a regression test covering the omitted-gas fallback path.

crates/evm/hardforks/src/lib.rs
haythemsellami: Fixed in e5434c640: `FoundryHardfork::from_chain_and_timestamp` now checks Monad after Tempo using `MonadHardfork::from_chain_and_timestamp` from `monad-revm`, and `Cargo.lock` is updated to the `foundry-revm-41-compat` commit that exposes that API. Added coverage for Monad mainnet/testnet timestamps, including pre-MonadNine mapping to MonadEight.

crates/anvil/src/config.rs
```diff
@@ -618,6 +624,9 @@ impl NodeConfig {
         if self.networks.is_tempo() {
             return TempoHardfork::default().into();
         }
+        if self.networks.is_monad() {
```
mablr, line 714: This default bypasses the Monad chain/timestamp mapping added in `FoundryHardfork::from_chain_and_timestamp`. I reproduced `NodeConfig::test_monad().with_chain_id(Some(143)).with_genesis_timestamp(Some(1_763_648_999))` reporting `MonadNine`, while the shared mapping returns `MonadEight`. Please mirror the Tempo inference path here before falling back to `MonadHardfork::default()`.

crates/evm/networks/src/lib.rs
haythemsellami: Fixed in 7fa357cc1. Added Monad precompile metadata to NetworkConfigs and wired Anvil eth_config to pass the active Monad hardfork, so MonadStaking is reported for Monad and MonadReserveBalance is only reported for MonadNine+. Also added NetworkConfigs unit coverage and Anvil eth_config tests for MonadEight/MonadNine.

crates/anvil/src/config.rs
haythemsellami: fixed

haythemsellami:
> An overall concern I have about third-party EVMs/Networks support in Foundry, is the alloy/revm bumps overheads and frictions. We experienced painful OP deps bumps while preparing Alloy 2.0 release, blocking/slowing us on Tempo development/integration. That's why we feature gated OP, and I'm wondering if we should do the same for Monad ? (This is mostly mechanical) If something breaks on Monad it should not impact Tempo's CI hotpath. WDYT?
> 
> -> some inline comment

Yeah agree, I gated Monad Foundry behind a non-default monad feature.
One release/distribution question: how do you expect Foundry releases to handle optional third-party EVM features? E.g, if a Foundry v1.8 binary is released with --features monad, could a later v1.9 official binary be released without that feature enabled?

mablr (commented):
Great! Thank you @haythemsellami.

We have basically two possibilities:
1. Accepting Monad as an optional source-build integration
2. Shipping Monad in official Foundry binaries

1st is obviously easier and require less commitment from our side, 2nd would nicer for Monad team I guess.

We need to discuss it a bit, cc @grandizzy @mattsse @figtracer @stevencartavia.

crates/config/src/lib.rs
```diff
@@ -5402,6 +5402,32 @@ mod tests {
         });
     }
 
+    #[test]
+    fn monad_hardfork_infers_monad_network() {
```
mablr: **Blocker:** This test runs in the default feature configuration, where the Monad hardfork variant is compiled out. It currently fails with `unknown hardfork monad:MonadNine`. Please add a `foundry-config/monad` feature that propagates the hardfork/network features and gate this test with it; the Monad-enabled CI job should then run it explicitly.

crates/chisel/src/dispatcher.rs
```diff
@@ -599,6 +599,19 @@ mod tests {
         );
     }
 
+    #[test]
+    fn ensure_loaded_session_network_matches_rejects_monad_on_default_network() {
```
mablr, line 811: **Blocker:** This is also executed without `chisel/monad`. `config_with_network(Some("monad"))` then panics because the compiled enum only contains Ethereum, Optimism, and Tempo; the focused default-feature test reproduces this. Please gate the test with `#[cfg(feature = "monad")]` and cover it in a Monad-enabled test job.

crates/evm/core/src/env.rs
```diff
@@ -498,11 +513,57 @@ impl<
...
+    }
+
+    fn cfg_env_mut(&mut self) -> &mut CfgEnv<Self::Spec> {
+        self.cfg.inner_mut()
+    }
+
+    fn set_spec_and_gas_params(&mut self, spec: Self::Spec) {
+        let mut cfg = self.cfg.clone().into_inner();
+        cfg.spec = spec;
+        self.cfg = MonadCfgEnv::from(cfg);
```
mablr, line 624: **High:** Rebuilding `MonadCfgEnv` updates the spec and gas parameters, but not the Monad instruction provider created with the EVM. Crossing MonadEight/MonadNine through `vm.setEvmVersion` therefore leaves the old MIP-3 memory opcode handlers installed, producing the wrong memory pricing. The CLZ test does not catch this because CLZ consults the runtime spec dynamically. Please either rebuild the instruction provider through an EVM-level transition hook or reject runtime transitions across this boundary, and add a memory-expansion gas regression test in both directions.

crates/evm/core/src/evm/monad.rs
```diff
@@ -0,0 +1,124 @@
...
+    type FoundryEvm<'db, I: FoundryInspectorExt<Self::FoundryContext<'db>>> =
+        MonadEvm<&'db mut dyn DatabaseExt<Self>, I>;
+
+    fn create_foundry_evm_with_inspector<'db, I: FoundryInspectorExt<Self::FoundryContext<'db>>>(
+        &self,
+        db: &'db mut dyn DatabaseExt<Self>,
+        evm_env: EvmEnv<Self::Spec, Self::BlockEnv>,
+        inspector: I,
+    ) -> Self::FoundryEvm<'db, I> {
+        let mut monad_evm = self.create_evm_with_inspector(db, evm_env, inspector);
```
mablr, line 147: **High:** The `evm_env` reaching this factory contains Foundry’s materialized 128 MiB default memory limit. The pinned `monad-revm` only substitutes MonadNine’s protocol 8 MiB pooled limit when it sees REVM’s untouched default sentinel, so Forge/Cast/script execution can currently exceed the network limit. Please clamp MonadNine+ execution to `min(configured_limit, MONAD_MEMORY_LIMIT)` (including hardfork transitions) and add coverage immediately below/above 8 MiB.

mablr (commented):
We are okay for 1.

I made further AI review pass, catching some blockers due to monad feat propagation.

haythemsellami:
> We are okay for 1.

@mablr Thanks for clarifying. I understand why option 1 is the lower-commitment choice for you guys. I’m trying to understand what value it is intended to provide to external EVM teams compared with maintaining a fork.

Under option 1, an external team would still need to build, publish, and support its own Foundry binaries with the relevant feature enabled. It would also depend on upstream review and merge timelines for changes to its integration. I'm not saying upstreaming add no value in this case, but the distribution burden remains on the external team.

In the case of forking, it gives full control over changes and releases, at the cost of carrying a maintenance burden with newer upstream version(which is not a big burden, especially in the age of LLMs).

My understanding was that the v1.7 network/EVM architecture was intended to make official Foundry binaries capable of supporting multiple EVMs. Was that an intended long-term goal, or was the scope primarily to make it easier to fork Foundry, and just upstream Tempo? 

Knowing wether there is a long-term goal of having other EVMs by default in the official binaries and distribution will help us decide whether upstreaming Monad and maintaining the code is the right path or no.

mablr:
Thanks for laying out the trade-offs. After further discussion, we can commit to option 2 and ship Monad support in official Foundry binaries for the upcoming 1.x releases.

This requires the Monad team to respond quickly when upstream Alloy or Revm changes break the integration, since this puts Monad’s EVM/network implementation on Foundry’s release-critical path.

The immediate goal of the 1.7 architecture work was to eliminate the need for Tempo’s fork as quickly as possible without introducing technical debt into Foundry’s EVM infrastructure. It was not intended merely to make Foundry easier to fork. `FoundryEvmNetwork` ended up providing a clean abstraction and will be the foundation for further composability work.

Longer term, our goal for Foundry 2.0 is to turn parts of Foundry, such as the EVM, scripting, cheatcodes, and Cast tooling, into an SDK with a philosophy similar to Reth. These components would likely live in `foundry-core`. With those composable foundations in place, support for OP, Monad, Base, and other custom EVMs or networks must become lean, and live outside the upstream Foundry repo.

Until that architecture is ready, upstreaming Monad and including it in official 1.x binaries is a reasonable bridge, as long as both teams are aligned on the maintenance commitment. I hope this roadmap clarification will help you make the most convenient decision for your team.

haythemsellami:
> This requires the Monad team to respond quickly when upstream Alloy or Revm changes break the integration, since this puts Monad’s EVM/network implementation on Foundry’s release-critical path.

Yes that makes sense and we can commit to that. feel free to tag me on any issue related to Monad's implementation.

crates/chisel/src/dispatcher.rs
haythemsellami: added #[cfg(feature = "monad")]

crates/evm/core/src/env.rs
haythemsellami: Fixed: updated monad-revm to change the instruction provider when entering a frame and restores the parent frame’s provider when returning. Also added regression tests in Foundry

crates/evm/core/src/evm/monad.rs
haythemsellami: Addressed in monad-revm#37 (https://github.com/category-labs/monad-revm/pull/37):
- MonadCfgEnv::memory_limit() now returns min(configured_limit, MONAD_MEMORY_LIMIT) for MonadNine+, while retaining the original configured value so transitions back to MonadEight restore it. Frame initialization also applies the effective limit to shared memory using the current hardfork.

Added Foundry coverage confirming the materialized 128 MiB limit transitions as 128 MiB → 8 MiB → 128 MiB, plus Forge execution coverage where the final word ending exactly at 8 MiB succeeds and the first word above it fails with MemoryLimitOOG.

crates/cheatcodes/src/inspector.rs
```diff
@@ -954,6 +976,34 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
             };
         }
 
+        #[cfg(feature = "monad")]
+        if is_monad_cheatcode_call::<FEN>(call.target_address) {
+            return match self.apply_monad_cheatcode(ecx, call) {
```
mablr, line 1431: **Blocker:** Monad cheatcodes mutate state before revm creates the callee checkpoint. Returning this synthetic `Revert` therefore leaves prior storage writes, transfers, or logs committed—for example, if `epochSnapshot` succeeds before `epochChange` fails.

Could dispatch create a checkpoint, commit it on success, and revert it on error? `blockReward` also updates `account.info.balance` directly, so that increase may survive an enclosing Solidity revert. Could it use journal-aware `balance_incr` instead? Regression coverage could exercise both partial-handler failure and a successful `blockReward` followed by an outer revert.

crates/evm/core/src/backend/mod.rs
```diff
@@ -66,8 +69,8 @@ pub type LocalForkId = U256;
 type ForkLookupIndex = usize;
 
 /// All accounts that will have persistent storage across fork swaps.
-const DEFAULT_PERSISTENT_ACCOUNTS: [Address; 3] =
-    [CHEATCODE_ADDRESS, DEFAULT_CREATE2_DEPLOYER, CALLER];
+const DEFAULT_PERSISTENT_ACCOUNTS: [Address; 4] =
+    [CHEATCODE_ADDRESS, MONAD_CHEATCODE_ADDRESS, DEFAULT_CREATE2_DEPLOYER, CALLER];
```
mablr: **Blocker:** The optional integration changes default Ethereum behavior. This address is always persisted across forks and is also unconditionally exempted from invariant overrides, omitted from gas reports, and labeled `MonadVM`. A normal contract at this address therefore receives special handling outside Monad.

Could all Monad-only address handling be gated by the feature and active network, using `FEN::EXTRA_CHEATCODE_ADDRESSES` where available? The other affected sites are `evm/fuzz/src/inspector.rs`, `forge/src/gas_report.rs`, `evm/traces/src/decoder/mod.rs`, and the unconditional config-warning suppression. A non-Monad regression test could verify ordinary-account behavior.

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -592,6 +622,11 @@ impl<N: Network> Backend<N> {
         self.networks.is_tempo()
     }
 
+    /// Returns true if Monad network mode is active
+    pub const fn is_monad(&self) -> bool {
+        self.networks.is_monad()
```
mablr, line 1340: **Blocker:** Resetting a plain Anvil instance to a Monad RPC can leave execution on the Ethereum EVM. `setup_fork_db_config` updates `NodeConfig.networks`, but `Backend::is_monad()` reads the copy captured at construction; `reset_block_number` never synchronizes it.

Could cross-family reset either update the backend network atomically or fail explicitly? A regression test could start from `NodeConfig::test()`, reset to chain 143/10143, and execute a Monad-only opcode or precompile. This is separate from the resolved hardfork-inference issue in https://github.com/foundry-rs/foundry/pull/15343#discussion_r3537710286.

crates/evm/traces/src/decoder/mod.rs
```diff
@@ -240,93 +250,122 @@ impl CallTraceDecoder {
...
+            .chain(IValidatorConfig::abi::functions().into_values())
+            .chain(IAccountKeychain::abi::functions().into_values())
+            .chain(IAddressRegistry::abi::functions().into_values())
+            .chain(ITIP20ChannelReserve::abi::functions().into_values())
+            .chain(ISignatureVerifier::abi::functions().into_values())
+            .chain(IReceivePolicyGuard::abi::functions().into_values())
+            .collect();
+        #[cfg(feature = "monad")]
+        {
+            function_groups.extend(IMonadStaking::abi::functions().into_values());
```
mablr: **Blocker:** With `monad` compiled, staking and reserve-balance labels, functions, and events enter global fallback maps. The same binary can trace other networks, so contracts at `0x1000`/`0x1001` or colliding selectors/topics may decode as Monad even though `is_precompile` correctly checks chain ID.

Could Monad metadata be registered by address only for Monad contexts, with reserve-balance entries additionally gated by hardfork? A feature-enabled chain-ID-1 test could cover both addresses and selector/topic collisions. This is distinct from the ABI-parity thread: https://github.com/foundry-rs/foundry/pull/15343#discussion_r3473262614.

crates/evm/core/src/evm/monad.rs
mablr, line 147: **Blocker:** Every Monad EVM starts with an empty `MonadChainContext`, but MonadNine reserve decisions depend on parent/grandparent and current-block sender/authority history plus transaction index. Repeated senders can therefore receive consensus-visible behavior different from Monad.

Could EVM construction receive and update the real chain context before each transaction? Historical replay could either reconstruct the required surrounding-block metadata or fail explicitly when it cannot. Coverage could include repeated same-block senders, parent/grandparent senders, EIP-7702 authorities, and state branching on `dippedIntoReserve()`.

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5565,11 +5709,11 @@ where
 
             // Check tx gas limit against tx gas limit cap (Osaka hard fork and later).
             if evm_env.cfg_env.tx_gas_limit_cap.is_none()
-                && tx.gas_limit() > evm_env.cfg_env().tx_gas_limit_cap()
+                && tx.gas_limit() > self.tx_gas_limit_cap(evm_env)
```
mablr, line 9234: **Blocker:** The later balance check applies Ethereum's `gas_limit * max_fee_per_gas + value` rule to Monad, while the Monad handler requires only `gas_limit * effective_gas_price`. Anvil can therefore reject transactions Monad would admit and mine with a failed receipt, including transactions whose balance covers the effective price but not the fee cap.

Could Monad pool admission use the execution handler's effective-gas-price calculation without adding `tx.value`, while retaining the existing rule for other networks? Boundary tests could cover unaffordable value and a balance below the max-fee cap but above the effective fee.

crates/evm/core/src/env.rs
```diff
@@ -498,11 +513,57 @@ impl<
...
+    }
+
+    fn db_journal_inner_mut(&mut self) -> (&mut Self::Db, &mut JournaledState) {
+        let journal: &mut Journal<DB> = self.journaled_state.deref_mut();
```
mablr, line 628: **Blocker:** `ReserveBalanceTracker` lives beside the wrapped `Journal`, not inside `JournalInner`, but nested-EVM and snapshot plumbing copies only `JournalInner`. Nested execution can therefore lose reserve debits, while `revertToState` restores account state without restoring the tracker's violation set.

Could the context abstraction carry Monad's tracker and auxiliary chain context through nested creation/write-back and snapshot lifecycle operations? Regression tests could cover a reserve-changing nested deployment and a reserve violation followed by `revertToState`.

mablr (commented):
Thank you @haythemsellami for the previous fixes.

I made a further pass focusing on the EVM integration correctness and found some blockers.

crates/cheatcodes/src/inspector.rs
haythemsellami: Fixed in 5a7ab30f3. Monad cheatcode dispatch now creates a journal checkpoint and commits or reverts it with the synthetic call result. blockReward now uses journal-aware balance_incr. Added regressions for partial epochBoundary failure and an outer revert after blockReward.

crates/evm/core/src/backend/mod.rs
haythemsellami: Fixed in 8131ffaa4 and 71533fcd2. Persistence, invariant exclusions, and gas-report filtering now derive from FEN::EXTRA_CHEATCODE_ADDRESSES; config warning suppression is feature-gated, and MonadVM trace metadata is only registered for an active Monad context. Added non-Monad regressions.

crates/anvil/src/eth/backend/mem/mod.rs
haythemsellami: Fixed in 7c742c577. Anvil now detects the target network family before mutating reset state and explicitly rejects Ethereum↔Monad resets while preserving same-family resets. Added coverage for resetting a plain Anvil instance to Monad.

crates/evm/traces/src/decoder/mod.rs
haythemsellami: Fixed in 71533fcd2. Monad labels, functions, and events are now address-scoped and only registered for an active Monad context; reserve-balance metadata additionally requires MonadNine+. Added chain-ID-1 selector/topic collision coverage.

crates/evm/core/src/evm/monad.rs
haythemsellami: Fixed in dc5226b8b with the updated monad-revm pin. Mining and historical replay now provide the exact current, parent, and grandparent participants, transaction index, and EIP-7702 authorities. Replay fails explicitly when full block context cannot be reconstructed. Added repeated-sender, ancestor, authority, and replay regressions.

crates/anvil/src/eth/backend/mem/mod.rs
haythemsellami: Fixed in 919fafd03. Monad pool admission now checks gas_limit × effective_gas_price without adding transaction value, while Ethereum behavior remains unchanged. Added coverage for both the effective-fee boundary and unaffordable-value failed receipts.

crates/evm/core/src/env.rs
haythemsellami: Fixed in dc5226b8b. FoundryContextState now carries the journal, Monad chain context, and ReserveBalanceTracker through nested EVM creation/write-back and snapshot/revert operations. Added nested-deployment and snapshot-revert regressions.

haythemsellami:
> I made a further pass focusing on the EVM integration correctness and found some blockers.

Pushed fixes with the help of codex, made sense for me, lmk.

Pushed few more commits for other findings.

crates/cast/src/cmd/run.rs
```diff
@@ -157,21 +162,26 @@ impl RunArgs {
         evm_opts.infer_network_from_fork().await;
 
         if evm_opts.networks.is_tempo() {
-            return self.run_with_evm::<TempoEvmNetwork>().await;
+            return self.run_with_evm::<TempoEvmNetwork>(evm_opts).await;
+        }
+
+        #[cfg(feature = "monad")]
+        if evm_opts.networks.is_monad() {
+            return self.run_with_evm::<MonadEvmNetwork>(evm_opts).await;
```
mablr: **Blocker:** This dispatches Monad through the generic local replay path below, which derives an Ethereum `SpecId` and requires `parentBeaconBlockRoot` for Cancun+. MonadEight/MonadNine map to Prague/Osaka, but canonical Monad RPC deliberately omits that field, so normal `cast run` fails with `MissingParentBeaconBlockRoot` before replay. Could beacon-root application be gated by the EVM family/capability instead, skipping EIP-4788 for Monad? A MonadEight/Nine fixture with no root would cover the real RPC shape.

crates/evm/core/src/backend/mod.rs
```diff
@@ -942,51 +1017,80 @@ impl<FEN: FoundryEvmNetwork> Backend<FEN> {
...
+        };
+        let Some(target_index) = transactions.iter().position(|tx| tx.tx_hash() == tx_hash) else {
+            return Ok(None);
+        };
+        let target_tx = transactions[target_index].clone();
+        let txs_to_replay = transactions[..target_index]
+            .iter()
+            .enumerate()
+            .filter(|(_, tx)| {
+                !is_known_system_sender(tx.from()) && tx.ty() != SYSTEM_TRANSACTION_TYPE
```
mablr: **Blocker:** Canonical Monad blocks can start with legacy transactions from the reserved system sender for `syscallReward`, `syscallSnapshot`, and `syscallOnEpochChange`. These mutate staking/epoch state, balances, logs, and the system nonce, so filtering them leaves transaction-level forks with stale pre-target state. They also cannot just go through ordinary `transact_commit`: their canonical gas limit is zero and Monad executes them through the dedicated system-call path. Could Monad replay dispatch these via `monad_revm::api::block`/`SystemCallEvm`, while preserving the existing filtering for other networks? A system-prefix + user-tx fixture should assert staking state and system nonce before the target.

crates/cast/src/cmd/run.rs
```diff
@@ -460,25 +481,36 @@ impl RunArgs {
                     }
 
                     let tx_env = TxEnvFor::<FEN>::from_recovered_tx(tx.as_ref(), tx.from());
+                    let context_aux = block_context.as_ref().map_or_else(
```
mablr, line 548: **Blocker:** Although this now supplies the Monad participant context, the replay loop immediately above still skips the leading Monad staking syscalls, so it reconstructs the wrong prestate for the target user transaction. `--replay-system-txes` is not a workaround because canonical system envelopes have zero gas and would enter ordinary transaction validation rather than Monad's system-call executor. Could Cast reuse the shared Monad system-transaction dispatcher needed by the backend replay path?

Cargo.toml
```diff
@@ -530,6 +530,17 @@ tempo-revm = { git = "https://github.com/tempoxyz/tempo", rev = "b314d4822ed6d7a
 tempo-contracts = { git = "https://github.com/tempoxyz/tempo", rev = "b314d4822ed6d7abdfb60cbde25ebad1ae6a98ec" }
 tempo-precompiles = { git = "https://github.com/tempoxyz/tempo", rev = "b314d4822ed6d7abdfb60cbde25ebad1ae6a98ec" }
 
+# Monad
+alloy-monad-evm = { git = "https://github.com/category-labs/alloy-monad-evm.git", branch = "foundry-revm-41-compat", default-features = false, features = [
```
mablr: **Blocker:** both Monad crates still come from mutable compatibility branches, and the corresponding upstream PRs are still open. This will need to be addressed before merge.

.github/workflows/ci.yml
```diff
@@ -68,6 +68,38 @@ jobs:
...
+        with:
+          persist-credentials: false
+      - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # v1
+        with:
+          toolchain: stable
+      - uses: rui314/setup-mold@9c9c13bf4c3f1adef0cc596abc155580bcb04444 # v1
+      - uses: mozilla-actions/sccache-action@9e7fa8a12102821edf02ca5dbea1acd0f89a2696 # v0.0.10
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
+      - name: Check Monad release binaries
+        run: cargo check --locked --bins --no-default-features --features "cli,monad"
```
mablr: **Consider:** release and Docker builds enable `aws-kms,gcp-kms,turnkey,cli,asm-keccak,js-tracer,monad`, while this checks only `cli,monad`. Since Monad is now part of official release binaries, CI should check the exact release feature combination once to catch cross-feature compilation conflicts?

mablr (commented):
Nice follow-up, I've found 3 more after recents changes.

2 remarks related to Ci/release.

We are getting close imo.

crates/evm/core/src/backend/mod.rs
stevencartavia: Monad replay omits preceding Monad system envelopes. Both transaction-level fork replay and default local `cast run` start from parent state and skip preceding transactions based on their system sender/type, so reward/snapshot/epoch transitions can be absent from the target prestate.

I reproduced this on Monad mainnet: a skipped zero-gas `syscallReward` incremented the system sender’s nonce, minted 25 MON into the staking account, changed staking storage, and emitted a log. `--replay-system-txes` is not a workaround because ordinary transaction execution rejects the zero-gas envelope during validation.

Replay needs to apply the full canonical semantics, including reward minting, dedicated system-call execution, and sender nonce handling.

crates/cast/src/cmd/run.rs
haythemsellami: Fixed in 61a9cddfc.

crates/evm/core/src/backend/mod.rs
haythemsellami: Fixed in 6772c20e3.

crates/cast/src/cmd/run.rs
haythemsellami: Fixed in 6772c20e3.

Cargo.toml
haythemsellami: Fixed in a4c2f5787 for now. I think before planning the next release with Monad evm included, I'll publish a new monad-revm and alloy-monad-evm versions.

.github/workflows/ci.yml
haythemsellami: Fixed in 34dec4af2.

crates/evm/core/src/backend/mod.rs
haythemsellami: Fixed in 6772c20e3.

mablr (commented):
Made a final review pass. The previous correctness, feature propagation, and CI/release concerns are all addressed, and I don’t see any remaining blockers beyond replacing the pinned Monad revm/alloy-monad-evm revisions with released versions before merge. 

Great job !🚀

crates/evm/core/src/backend/mod.rs
```diff
@@ -1391,10 +1551,28 @@ impl<FEN: FoundryEvmNetwork> DatabaseExt<FEN::EvmFactory> for Backend<FEN> {
...
+                    block.header().number()
+                )
+            })?;
+            self.block_context_inputs(id, &block)?.transaction(current_tx_index)
+        } else {
+            factory.context_for_transaction(&tx_env)
+        };
+
         let fork = self.inner.get_fork_by_id_mut(id)?;
         commit_transaction::<FEN>(
```
stevencartavia, line 2018: `vm.transact` still routes a canonical Monad staking reward envelope used as the target through ordinary transaction validation instead of the protocol replay path.

I reproduced this at head `8d2141b750` using a real Monad reward transaction, both from a transaction-hash fork and an explicit parent-block fork. Both fail with `gas price is less than basefee`, although the transaction succeeded on-chain with gas limit, gas price, and receipt gas used all zero.

The target path should apply the same full protocol semantics as preceding-envelope replay: validate and increment the system nonce, mint the reward into staking, and execute the dedicated system call.

haythemsellami: Fixed.

crates/evm/core/src/backend/mod.rs
```diff
@@ -1391,16 +1577,55 @@ impl<FEN: FoundryEvmNetwork> DatabaseExt<FEN::EvmFactory> for Backend<FEN> {
...
+            TransactionInputs { evm_env, tx_env, context_aux },
             journaled_state,
             fork,
             &fork_id,
             &persistent_accounts,
             inspector,
-        )
+        )?;
+        if let Some((current_tx_index, transaction_count)) = current_tx_position
+            && fork.position
```
stevencartavia: `vm.transact` commits an index-0 transaction from an explicit parent-block fork but leaves the fork cursor at `AfterBlock(parent)`.

Reproduced at `db61896230` with `rollFork(id, parentBlock)` followed by `transact(id, txHash)`: the sender nonce and recipient balance update, but a subsequent reserve-sensitive transaction treats the replayed transaction’s sender as fresh. Rolling to the transaction first gives the correct context.

When the target is index 0 of the block directly following `parent`, this path should advance to index 1—or `AfterBlock(targetBlock)` when it is the last transaction—like the existing `BeforeTransaction` branch.

haythemsellami:
Tracked the separate Anvil `--fork-transaction-hash` protocol-envelope replay gap in #15878.

The `vm.transact` protocol replay path is addressed in this PR. Anvil’s forced startup block still uses a distinct `PoolTransaction` validation/mining path and needs a dedicated historical replay batch with source block context, source chain ID, protocol-envelope semantics, atomic execution, and isolation from live local pool transactions. That broader Anvil change is therefore deferred to the issue rather than expanding this integration PR into core Anvil replay architecture.

mablr:
For some reason I can't trigger CI 🤔

grandizzy:
> For some reason I can't trigger CI 🤔

you may want to close / reopen PR or have @haythemsellami push a new commit

haythemsellami:
@mablr @stevencartavia @figtracer @grandizzy I pushed few more fixes in this PR.

I think the PR is getting quite big and it is in good state now, and it makes sense to aim to merge it. Let me know if you find any other blockers.

I also published the two new crates for monad-revm and alloy-monad-evm that being used in this PR.

mablr:
Thanks @haythemsellami!

This lgtm, but we need to coordinate with revm 42 bump https://github.com/foundry-rs/foundry/pull/16029 which is about land. It may involve some frictions on your side.

haythemsellami:
> This lgtm, but we need to coordinate with revm 42 bump #16029 which is about land. It may involve some frictions on your side.

Yeah I see the bump PR is merged, working on the new changes to support revm 42.
Btw @mablr I sent you a message on telegram too for better coordination.

mablr (dismissed):
I made an e2e review pass, after the revm 42 bump.

I see no remaining blockers, everything seems well gated behing `monad` feature (no monad build by default). No observed semantic regression for Ethereum, Tempo, or Optimism.

IMO, remaining Monad feats (like MIP-8 support), or correctness fixes can be handled in follow-ups before 1.8.0 release lands.

figtracer (dismissed):
lgtm!

crates/evm/core/src/opts.rs
```diff
@@ -201,49 +488,294 @@ impl EvmOpts {
...
+        );
+    }
+
+    async fn fork_node_info<N: Network, P: Provider<N>>(
+        provider: &P,
+    ) -> eyre::Result<Option<NodeInfo>> {
+        match provider.raw_request::<_, NodeInfo>("anvil_nodeInfo".into(), ()).await {
+            Ok(node_info) => Ok(Some(node_info)),
+            Err(error) if is_rpc_method_not_found(&error) => Ok(None),
+            Err(error) => Err(error).wrap_err("failed to determine network family from endpoint"),
```
grandizzy: This makes the optional `anvil_nodeInfo` capability probe fatal for ordinary fork endpoints unless they return exactly JSON-RPC `-32601`.

RPC gateways commonly reject unknown/custom methods with HTTP 403, vendor-specific `-320xx` errors, or other responses while still supporting every Ethereum method needed for forking. As a result, `forge test --fork-url`, `forge script`, and Cast tracing can now fail before execution, even when Ethereum or Optimism was selected explicitly.

Please treat this probe as unavailable on failure until the endpoint has successfully identified itself as Anvil. Once valid Anvil metadata has been observed, subsequent identity checks can remain strict. The equivalent probe in `anvil/src/config.rs::fork_endpoint_identity` needs the same handling.

crates/evm/networks/src/lib.rs
```diff
@@ -200,6 +348,13 @@ impl From<ChainId> for NetworkVariant {
         if chain.is_tempo() {
             return Self::Tempo;
         }
+        if matches!(chain.named(), Some(NamedChain::Celo | NamedChain::CeloSepolia)) {
+            return Self::Ethereum;
+        }
+        #[cfg(feature = "monad")]
```
grandizzy: In a build without the `monad` feature, known Monad chain IDs skip this branch and are classified as Ethereum by the fallback below. The same issue exists in `NetworkConfigs::with_chain_id`, where `anvil --chain-id 143` and `chisel --chain 143` silently execute Ethereum semantics while advertising a Monad chain ID.

This conflicts with `from_known_chain_id`, which correctly reports that the Monad network family is unavailable. Please make user-facing chain-ID inference fallible and reject known feature-gated networks unless an explicit Ethereum network override was supplied.

grandizzy (commented):
Thank you, looks good, couple of findings inline, please check.

crates/evm/core/src/opts.rs
mablr: Yes, good catch. This is as regression, it should be handled before merge.

crates/evm/networks/src/lib.rs
mablr: Yes I agree, this one is less important, not PR blocking I would say, but could done together with the fix above.

crates/evm/core/src/opts.rs
haythemsellami: Fixed in 7c02e1ef2.

crates/evm/networks/src/lib.rs
haythemsellami: Fixed in c12ac20d6.

grandizzy (dismissed):
thank you, lgtm!

mablr (dismissed):
Let's go🔥

crates/forge/tests/cli/test_cmd/mod.rs
```diff
@@ -4455,7 +4462,7 @@ contract ForkTest is Test {
     cmd.args(["test", "--mt", "test_fork_err_message"]).assert_failure().stdout_eq(str![[r#"
 ...
 Ran 1 test for test/ForkTest.t.sol:ForkTest
-[FAIL: vm.createSelectFork: could not instantiate forked environment with provider eth-mainnet.g.alchemy.com; HTTP error 401 with body: [..]
+[FAIL: vm.createSelectFork: could not instantiate forked environment with provider eth-mainnet.g.alchemy.com; failed to determine network family from endpoint; HTTP error 401 with body: [..]
```
mablr: `7c02e1ef2` makes the initial `anvil_nodeInfo` failure optional, so this now fails at `eth_chainId` with `failed to retrieve chain ID from fork endpoint`. Since this test only checks that the provider hostname is preserved, can we wildcard the remaining context?

```suggestion
[FAIL: vm.createSelectFork: could not instantiate forked environment with provider eth-mainnet.g.alchemy.com; [..]
```

grandizzy (dismissed):

haythemsellami:
Created new issues that were found with codex to keep track of and fix after merging this one.

mablr (dismissed):

mablr (approved):

grandizzy (approved):
thank you!

PR #15344, opened 2026-06-24
fix(tempo): avoid receive-policy recipient collisions

### LINKED ISSUE #15342: bug: tempo nightly workflow failed
The nightly Tempo workflow (mainnet and testnet checks) has failed. This indicates a regression in Foundry's Tempo support or an issue with the Tempo mainnet/testnet endpoints.

Check the [tempo nightly workflow page](https://github.com/foundry-rs/foundry/actions/runs/28072039592
) for details.

This issue was raised by the workflow at `.github/workflows/ci-tempo.yml`.

decofe:
## Summary
- derive the Tempo template recipient from the freshly-created token address instead of the canonical Anvil BOB address
- remove TIP403 probe/retry logic from the starter template
- use a separate bad signer key for the negative signature test now that `BOB` is token-derived

## Why this failed
The scheduled Tempo testnet CI was failing because the generated template used the canonical Anvil `BOB` address, `0x70997970C51812dc3A010C7d01b50e0d17dc79C8`, as the TIP20 recipient.

That address now has a receive policy on public Moderato. The live TIP403 registry rejects direct receipt for the freshly-created template token with `validateReceivePolicy(PATH_USD, ALICE, BOB) -> (false, 2)`, where `2` is `TOKEN_FILTER`. The transfer path succeeds, but the token is routed to the receive-policy guard instead of being credited to `BOB`, so assertions like `balanceOf(BOB) == amount` fail.

## Fix
The template now derives `BOB` during `setUp()` from the generated token address:

```solidity
BOB = address(uint160(uint256(keccak256(abi.encodePacked("tempo-mail-bob", address(token))))));
```

Because the token address is fresh for each template test setup, the recipient is not a reusable public dev address that can be preconfigured on Moderato. This keeps the starter template simple and chain-agnostic while avoiding the public-state collision.

`test_WrongSignerReverts` now signs with an unrelated `BAD_SIGNER_PK` instead of a Bob private key, since Bob is no longer derived from a fixed key.

Closes #15342
Successful run: https://github.com/foundry-rs/foundry/actions/runs/28081824800

## Tests
- `forge fmt crates/forge/assets/tempo/MailTemplate.t.sol`
- generated template direct tests on Moderato: `forge test --fork-url https://rpc.moderato.tempo.xyz --mc MailTest -vv`
- `cargo test -p forge --test cli tempo --no-fail-fast`

Prompted by: @grandizzy

### DIFF
--- a/crates/forge/assets/tempo/MailTemplate.t.sol
+++ b/crates/forge/assets/tempo/MailTemplate.t.sol
@@ -8,13 +8,23 @@ import {StdPrecompiles} from "tempo-std/StdPrecompiles.sol";
 import {StdTokens} from "tempo-std/StdTokens.sol";
 import {Mail} from "../src/Mail.sol";
 
+interface ITIP403Registry {
+    function validateReceivePolicy(address token, address sender, address receiver)
+        external
+        view
+        returns (bool authorized, uint8 blockedReason);
+}
+
 /// @notice Tests for direct mail sending (no signature verification).
 contract MailTest is Test {
+    address internal constant TIP403_REGISTRY = address(0x403c000000000000000000000000000000000000);
+
     ITIP20 public token;
     Mail public mail;
 
     address public constant ALICE = address(0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266);
-    address public constant BOB = address(0x70997970C51812dc3A010C7d01b50e0d17dc79C8);
+    address public BOB;
+    uint256 internal BOB_PK;
 
     function setUp() public virtual {
         address feeToken = vm.envOr("TEMPO_FEE_TOKEN", StdTokens.PATH_USD_ADDRESS);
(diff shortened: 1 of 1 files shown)


crates/forge/assets/tempo/MailTemplate.t.sol
```diff
@@ -8,13 +8,23 @@ import {StdPrecompiles} from "tempo-std/StdPrecompiles.sol";
 import {StdTokens} from "tempo-std/StdTokens.sol";
 import {Mail} from "../src/Mail.sol";
 
+interface ITIP403Registry {
+    function validateReceivePolicy(address token, address sender, address receiver)
+        external
+        view
+        returns (bool authorized, uint8 blockedReason);
+}
```
mablr: This works, but I think we can simply avoid pulling TIP403 probing/retry logic into the starter template. Since new tokens default to always-allow and receive policies only apply when the receiver has one configured, derive `BOB` from the freshly-created token address instead, e.g. `BOB = address(uint160(uint256(keccak256(abi.encodePacked("tempo-mail-bob", address(token))))));`. Then `BOB` has no reusable public key/address to preconfigure on Moderato, and `test_WrongSignerReverts` can sign with an unrelated `BAD_SIGNER_PK` instead of Bob’s key.

wdyt?

mablr (approved):

figtracer (approved):

PR #15348, opened 2026-06-24
feat(fuzz): add invariant selector focus workers

figtracer:
## Summary
- reserve a small tail shard of invariant workers for selector focus mode when campaigns run with more than one worker
- derive focus candidates from each target contract effective fuzzed functions so target/exclude selector configuration remains authoritative
- add unit and forge CLI coverage for focus worker sharding, targeted selector exercise, allowlists, and blocklists

cc @0xalpharush


crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -817,6 +882,8 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
                     let _guard =
                         info_span!("invariant_worker", id = worker_plan.worker_id).entered();
                     let timer = Instant::now();
+                    let worker_campaign_seed =
+                        campaign_seed_for_worker(&campaign_seed, worker_plan, actual_worker_count);
```
mablr: This narrows `worker_campaign_seed` for focus workers, but the `corpus_seed` passed to the same worker is still cloned from the full campaign corpus loaded before focus selection. In coverage-guided runs, corpus-derived sequences can retain calls to non-focused selectors, so focus workers may not actually stay focused.

Maybe consider filtering the worker corpus seed against `worker_campaign_seed.targeted_contracts`, or adding a regression with `corpus_dir` enabled and an off-focus persisted sequence?

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -265,6 +267,69 @@ const fn invariant_worker_collects_evm_cmp_log(
...
+    let Some(focus_index) = invariant_focus_worker_index(plan.worker_id, worker_count) else {
+        return campaign_seed.clone();
+    };
+    let Some(targeted_contracts) =
+        focused_targeted_contracts(&campaign_seed.targeted_contracts, focus_index)
+    else {
+        return campaign_seed.clone();
+    };
+
+    InvariantCampaignSeed { targeted_contracts, ..campaign_seed.clone() }
```
mablr: The focused seed only replaces `targeted_contracts` and preserves `targets_are_updatable`. For auto-updatable campaigns, `collect_created_contracts` can later add newly deployed contracts with their full fuzzable selector set into the focus worker’s `TargetedContracts`, so the worker stops being single-selector focused after a deployment. Is this okay ?

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -265,6 +267,98 @@ const fn invariant_worker_collects_evm_cmp_log(
...
+}
+
+fn invariant_focus_worker_index(worker_id: u32, worker_count: usize) -> Option<usize> {
+    let worker_id = worker_id as usize;
+    let focus_workers = invariant_focus_worker_count(worker_count);
+    if focus_workers == 0 || worker_id >= worker_count {
+        return None;
+    }
+
+    let first_focus_worker = worker_count - focus_workers;
```
stevencartavia: Focus workers can overlap the exploratory worker. When `corpus_random_sequence_weight` is left unconfigured at its default, `invariant_worker_config` makes the last worker (`worker_count-1`, for `worker_count > 2`) the ensemble/fresh-input worker (`ENSEMBLE_CORPUS_RANDOM_SEQUENCE_WEIGHT`). But focus workers are the tail range `[worker_count-focus_count, worker_count-1]`, which includes that last worker. So for `worker_count` 3..=15, if there are at least two effective `(address, selector)` candidates, the only focus worker is also the ensemble worker and gets narrowed to a single selector — leaving no worker with both the ensemble setting and the full target set. Suggest placing focus workers just before the last worker when `worker_count > 2`.

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -265,6 +267,98 @@ const fn invariant_worker_collects_evm_cmp_log(
...
+            if seen.insert((*address, function.selector())) {
+                candidates.push((*address, function.clone()));
+            }
+        }
+    }
+    if candidates.len() <= 1 {
+        return None;
+    }
+
+    let (address, function) = candidates[focus_index % candidates.len()].clone();
```
stevencartavia: For `worker_count` 2..=15, focus always chooses the first effective `(address, selector)` candidate when focus activates. `focused_targeted_contracts` returns `None` for `len <= 1`; otherwise it picks `candidates[focus_index % len]`, and `focus_index` is always 0 in that worker-count range. The candidate order is deterministic and seed-independent, so changing the fuzz seed picks the same candidate for the same target set. Suggest adding a seed-derived offset so focus is spread across candidates.

mablr (commented):
Seems fine, @figtracer please fix `invariant_selector_focus_worker_exercises_targeted_selector` regression.

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -795,6 +893,7 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
         let setup_contracts = self.setup_contracts;
         let project_contracts = self.project_contracts;
         let base_executor = self.executor.clone();
+        let focus_seed = self.fuzz_seed;
```
mablr: This only uses `self.fuzz_seed` for selector-focus rotation, so normal stochastic runs where `fuzz.seed` is unset still pass `None` and always use offset 0. That leaves focus workers pinned to the same first candidate across unseeded reruns even though their `TestRunner` RNG is random. Is this intentional ?

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -265,6 +267,108 @@ const fn invariant_worker_collects_evm_cmp_log(
...
+    let mut seen = HashSet::new();
+    let mut candidates = Vec::new();
+    for (address, contract) in targeted_contracts.iter() {
+        // Build from the effective selector set so user target/exclude config stays authoritative.
+        for function in contract.abi_fuzzed_functions() {
+            if seen.insert((*address, function.selector())) {
+                candidates.push((*address, function.clone()));
+            }
+        }
+    }
```
mablr, line 357: `abi_fuzzed_functions()` skips the excluded_functions filter when `targeted_functions` is non-empty (it returns `targeted_functions.iter()` directly). So if a user calls both `targetSelector` and `excludeSelector` for the same selector, it may end up as the sole focus target here ? The `invariant_focus_skips_excluded_selectors` test only covers the empty-targeted_functions branch.

mablr (commented):
sgtm, please fix new conflicts

mablr (dismissed):

crates/evm/fuzz/src/invariant/mod.rs
```diff
@@ -322,7 +322,11 @@ impl TargetedContract {
                 ) && !self.excluded_functions.contains(func)
             }))
         } else {
-            Either::Left(self.targeted_functions.iter())
+            Either::Left(
+                self.targeted_functions
+                    .iter()
+                    .filter(|func| !self.excluded_functions.contains(func)),
```
stevencartavia, line 388: **Optional / edge case:** now that the targeted branch filters `excluded_functions` (added in `b301601`), a self-contradictory config that both targets and excludes the same selector(s) can leave zero effective fuzzable functions. When that empties the whole set, `fuzzed_functions()` is empty and the invariant strategy hits proptest's `Selector::select` → `.expect("select from empty iterator")`, panicking instead of erroring cleanly.

Repro (this branch panics, master `[PASS]`):
```solidity
contract T { function foo() public {} }
// setUp:
bytes4[] memory s = new bytes4[](1);
s[0] = t.foo.selector;
targetSelector(FuzzSelector({addr: address(t), selectors: s}));
excludeSelector(FuzzSelector({addr: address(t), selectors: s}));
```

This only happens with a nonsensical target+exclude overlap that empties all targets, so it's very unlikely in practice — not a blocker. Could optionally bail with a clear "no functions to fuzz" error instead of panicking.

mablr (dismissed):

stevencartavia (dismissed):

stevencartavia (approved):

mablr (approved):

PR #15349, opened 2026-06-24
fix(invariant): corpus empty sequence input

mablr:
Avoid indexing into an empty initial corpus sequence in `generate_next_input`.

Generate a fresh transaction when the requested depth is outside the available sequence, and cover the empty-sequence case with a regression test.


mattsse (approved):
LGTM

grandizzy (approved):

PR #15354, opened 2026-06-24
feat(symbolic): seed fuzz corpus from symbolic runs

mablr:
## Summary

Closes OSS-370.

Adds `forge test --symbolic-seed-corpus` / `FOUNDRY_SYMBOLIC_SEED_CORPUS`.

When enabled, fuzz tests still run with normal fuzz semantics, but Forge first tries to derive a concrete passing input with the symbolic engine and persist it to the configured fuzz corpus directory using the existing corpus format.

Symbolic seeding is best-effort: incomplete symbolic execution, counterexamples, or concrete replay failures do not fail the fuzz test. Later fuzz runs can load persisted seeds via `--fuzz-corpus-dir`.


0xalpharush:
I think we want the ability to concretize non-failing inputs, right? It's probably pretty rare to have an existing failure to start fuzzing with

crates/config/src/symbolic.rs
```diff
@@ -32,6 +32,8 @@ pub enum SymbolicExplorationOrder {
 pub struct SymbolicConfig {
     /// Whether symbolic tests are enabled.
     pub enabled: bool,
+    /// Whether fuzz tests should be symbolically concretized into fuzz corpus seeds.
+    pub seed_corpus: bool,
```
figtracer, line 36: `seed_corpus` also needs to be added to `SYMBOLIC_KEYS`; otherwise `[symbolic] seed_corpus = true` is reported as an unknown key and `test all` is failing on snapshot output.

crates/forge/tests/cli/test_cmd/symbolic.rs
```diff
@@ -1520,6 +1520,56 @@ invariant_targetNeverBroken()
...
+    cmd.forge_fuse()
+        .args([
+            "test",
+            "--match-test",
+            "testAcceptsMostValues",
+            "--symbolic-seed-corpus",
+            "--fuzz-corpus-dir",
+            "fuzz_corpus",
+        ])
+        .assert_success();
```
figtracer, line 1602: can this assert a second normal fuzz run with `--fuzz-corpus-dir fuzz_corpus` loads the generated seed? right now the test only proves the file was written, so corpus loading could regress without this failing.

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -228,6 +229,19 @@ pub(crate) struct CampaignCorpusEntry {
...
+pub fn persist_corpus_seed(
+    config: &FuzzCorpusConfig,
+    tx_seq: Vec<BasicTxDetails>,
+) -> foundry_common::fs::Result<Option<PathBuf>> {
+    let Some(root) = &config.corpus_dir else {
+        return Ok(None);
+    };
+    let corpus_dir = root.join(format!("{WORKER}0")).join(CORPUS_DIR);
+    foundry_common::fs::create_dir_all(&corpus_dir)?;
+    CorpusEntry::new(tx_seq).write_to_disk_in(&corpus_dir, config.corpus_gzip).map(Some)
```
stevencartavia, line 261: Symex + fuzz beats either alone, but in a narrow zone. Clear win:
```solidity
if (x * 7 != 1) return;   // fuzz can't guess this; the solver gets it instantly
require(y != 13);         // fuzz finds this once it's past the gate
```
fuzz-alone 0/8, symex-alone misses it, hybrid 8/8 — symex opens the gate, fuzz finds the bug behind it. The current `MAGIC = 0xdeadbeef` demo doesn't show this because fuzz already cracks literal gates on its own. Worth swapping in a case like the above.

**Suggestion**: with the flag on, the test randomly fails (~5%) with `EOF while parsing ... workerN/corpus/*.json` — corpus files are written non-atomically, so a worker reads a half-written file. Fix: write to a temp file, then rename. The new test hits this path, so it can flake CI.

Minor: for `assert`-style bugs symex actually finds the counterexample but the code just warns and seeds nothing, so the test can still pass.

mattsse (commented):
I found two things worth tightening before merge:

1. `--symbolic-seed-corpus` can keep appending the same seed forever. `try_seed_fuzz_corpus_symbolically` always calls `persist_corpus_seed`, and `persist_corpus_seed` creates a fresh `CorpusEntry` / UUID every time. Since load-time dedupe only filters duplicate UUIDs, rerunning the same test against the same corpus dir writes duplicate entries, and stateless fuzz then replays those loaded entries verbatim.

   Suggested fix: make seed persistence idempotent for the concrete tx sequence before writing. A simple version would scan the per-test `worker*/corpus` dirs and skip when an existing entry has the same single-call `{sender,target,calldata,value}`. A stronger version would derive a stable content key/hash for seed entries and use that for dedupe, or route symbolic seeds through the same coverage-based replay filter used by campaign output persistence.

2. `--symbolic-seed-corpus` still pays for symbolic execution when there is no configured corpus dir. In that case `persist_corpus_seed` returns `Ok(None)`, so the flag does an expensive no-op with no visible indication to the user.

   Suggested fix: check `fuzz_config.corpus.corpus_dir.is_none()` before constructing/running `SymbolicExecutor` and either warn that `--symbolic-seed-corpus` requires `--fuzz-corpus-dir` / `fuzz.corpus_dir`, or reject the flag combination during CLI/config validation. That keeps the feature from silently doing work it cannot persist.

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -868,6 +913,7 @@ impl WorkerCorpus {
         if let Some(index) = self.current_mutated_index.take() {
             let should_credit = new_coverage || improved_optimization;
             if let Some(corpus) = self.in_memory_corpus.get_mut(index) {
+                corpus.cmp_seq = cmp_seq.iter().take(corpus.tx_seq.len()).cloned().collect();
```
stevencartavia: This line looks unrelated to the symbolic seeding feature. It overwrites the source corpus entry's `cmp_seq` with the cmp log from the mutated run, but leaves that entry's `tx_seq` unchanged — so the stored `tx_seq` and `cmp_seq` can describe different executions. I also don't see a test covering it (removing the line keeps all the new e2e tests and the corpus unit tests green). Do we need that?

crates/forge/src/runner.rs
```diff
@@ -1638,12 +1642,37 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
             ffi_enabled: self.config.ffi,
+            collect_success_input: should_symbolically_seed_fuzz_corpus(&self.config, func),
         });
         let portfolio_diagnostics = symbolic.portfolio_diagnostics();
         let symbolic_diagnostics = symbolic.take_diagnostics();
 
         match result {
-            SymbolicRunResult::Safe(stats) => {
+            SymbolicRunResult::Safe { stats, success_input } => {
+                if let Some(input) = success_input {
```
stevencartavia: I think this block is unreachable: at this call site `collect_success_input` is always false (it's only true for fuzz tests, but `run_symbolic_test` only runs for `check*`/`prove*` entrypoints), so `success_input` is always `None` here. Can we remove the block?

crates/forge/tests/cli/test_cmd/symbolic.rs
```diff
@@ -1520,6 +1520,200 @@ invariant_targetNeverBroken()
...
+}
+"#,
+    );
+
+    cmd.forge_fuse()
+        .args(["test", "--match-test", "testHybridFindsBug", "--threads", "1", "--fuzz-runs", "8"])
+        .assert_success();
+
+    cmd.forge_fuse()
+        .args(["test", "--symbolic", "--match-test", "testHybridFindsBug"])
```
stevencartavia: `testHybridFindsBug` isn't a `check*`/`prove*` entrypoint, so `--symbolic` doesn't select the symbolic executor for it — this is just an ordinary fuzz run. Either drop this step or change it to assert the behavior it's meant to cover.

figtracer (dismissed):

stevencartavia (approved):

figtracer (approved):

PR #15148, opened 2026-06-12
fix(foundryup): resolve latest/stable via release redirect to avoid API limit

stevencartavia:
Resolving `stable`/`latest` hit `api.github.com`, which is rate limited to 60 req/hr per IP. Shared CI IPs exhaust this and fail with `curl: (56) ... 403`.

Resolve via the `github.com/<repo>/releases/latest` redirect instead, which isn't subject to that limit (no `GITHUB_TOKEN` needed). Falls back to the API if the redirect can't be resolved. `nightly` and explicit tags unchanged.

Closes #14836 
Closes OSS-279

Issue #14836, opened 2026-05-06
foundry-toolchain@v1 actions fails with `curl: (56) The requested URL returned error: 403`


heueristik:
The action fails with

```
Run foundry-rs/foundry-toolchain@v1
  with:
    version: stable
    network: ethereum
    cache: true
    cache-key: tests-c492725713a695b0f4ac898b437ce4425b0ff1e9
    cache-restore-keys: tests-
  env:
    FOUNDRY_PROFILE: ci
Installing Foundry (version: stable)
Downloading foundryup installer...
Running foundryup installer...
Running: foundryup 


.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx

 ╔═╗ ╔═╗ ╦ ╦ ╔╗╔ ╔╦╗ ╦═╗ ╦ ╦         Portable and modular toolkit
 ╠╣  ║ ║ ║ ║ ║║║  ║║ ╠╦╝ ╚╦╝    for Ethereum Application Development
 ╚   ╚═╝ ╚═╝ ╝╚╝ ═╩╝ ╩╚═  ╩                 written in Rust.

.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx

Repo       : https://github.com/foundry-rs/foundry
Book       : https://book.getfoundry.sh/
Chat       : https://t.me/foundry_rs/
Support    : https://t.me/foundry_support/
Contribute : https://github.com/foundry-rs/foundry/blob/HEAD/CONTRIBUTING.md

.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx.xOx

foundryup: checking if foundryup is up to date...
foundryup: foundryup is up to date.
foundryup: fetching latest release tag from foundry-rs/foundry...

Error: curl: (56) The requested URL returned error: 403
curl: (56) The requested URL returned error: 403
curl: (56) The requested URL returned error: 403
curl: (56) The requested URL returned error: 403
curl: (56) The requested URL returned error: 403
curl: (56) The requested URL returned error: 403
foundryup: failed to fetch release tags from GitHub API

Error: Error: Command failed: bash "/Users/runner/.foundry/bin/foundryup" 
[...]

PR #15356, opened 2026-06-24
fix(foundryup): resolve nightly via releases Atom feed to avoid API rate limit

grandizzy:
## Motivation

`foundryup --install nightly` resolves the tag by hitting `api.github.com/repos/<repo>/releases`, which is rate limited to **60 req/hr per IP** for unauthenticated requests. Shared CI IPs exhaust this and fail with:

```
foundryup: fetching latest nightly release tags from foundry-rs/foundry...
curl: (22) The requested URL returned error: 403
foundryup: failed to fetch release tags from GitHub API
```

#15148 fixed this for `latest`/`stable` via the `releases/latest` web redirect, but left the `nightly` path on the REST API, so nightly installs still 403 on busy CI runners. (Closes the same class of issue as #14836, for nightly.)

## Solution

Resolve the most recent `nightly-<sha>` release from GitHub's public **releases Atom feed** (`https://github.com/<repo>/releases.atom`):

- lists releases **newest-first**, so the first `nightly-<40-hex-sha>` entry is the latest nightly
- works **unauthenticated** and is **not** subject to the `api.github.com` REST rate limit (same risk profile as the #15148 redirect)
- resolves to the **immutable** `nightly-<sha>` tag, preserving the per-commit version directory the rest of foundryup expects
- falls back to the existing REST API path only when the feed cannot be parsed

The rolling `nightly` tag is intentionally **not** used: it can lag behind the latest nightly (it currently points at an April build while newer nightlies exist), so relying on it would install a stale toolchain.

`latest`/`stable` are deliberately left on the `releases/latest` redirect: that endpoint excludes prereleases (so it can't return a nightly) and honors the maintainer-designated "Latest" release, which is the correct semantic for stable.

## Testing

- `foundryup/test.sh` extended with mocked coverage for `resolve_nightly_tag_via_feed` and the feed→API fallback (no network). All pass; `shellcheck` clean.
- Verified end-to-end with `GITHUB_TOKEN`/`GH_TOKEN` unset: `foundryup --install nightly` resolves `nightly-bc3b7f1e90bb7a30903d7d1ae2c532e4fba5679d` (latest), passes attestation verification, installs into `versions/nightly-<sha>`, and makes **zero** `api.github.com` calls.

### DIFF
--- a/foundryup/foundryup
+++ b/foundryup/foundryup
@@ -7,7 +7,7 @@ set -eo pipefail
 
 # NOTE: if you make modifications to this script, please increment the version number.
 # WARNING: the SemVer pattern: major.minor.patch must be followed as we use it to determine if the script is up to date.
-FOUNDRYUP_INSTALLER_VERSION="1.8.6"
+FOUNDRYUP_INSTALLER_VERSION="1.8.7"
 
 BASE_DIR=${XDG_CONFIG_HOME:-$HOME}
 FOUNDRY_DIR=${FOUNDRY_DIR:-"$BASE_DIR/.foundry"}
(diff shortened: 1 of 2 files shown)


mablr (approved):
Seems good to me ! ^^

stevencartavia (approved):

mattsse (approved):
LGTM

PR #15359, opened 2026-06-24
fix(cast): respect --gas-limit in cast call --trace

### LINKED ISSUE #15357: Incorrect gas limit in `cast call --trace`
### Component

Cast

- [x] Foundry
- [x] Foundryup

forge Version: 1.5.1-stable Commit SHA: b0a9dd9ceda36f63e2326ce530c10e6916f4b8a2 Build Timestamp: 2025-12-22T11:39:01.425730780Z (1766403541) Build Profile: maxperf

foundryup: 1.6.1

cast call --trace

### Operating System

Linux

### Describe the bug

We have transactions we are trying to test with `cast call` that depend on assumptions about the gas limit of the transaction. The overriding done here only apply to `--trace` but cause the transaction to revert.
https://github.com/foundry-rs/foundry/blob/c262f5df4a221681b79108793cb9f1ec64c7fc5e/crates/cast/src/cmd/call.rs#L409

This issue doesn't happen without `--trace` as this override to the transaction is only set when using this flag. In either case, I feel like the result should at least be consistent with or without the flag, but `cast call` does provide a `--gas-limit` flag so I feel like that should be respected if specified. Perhaps set the gas limit to the max if the flag is not specified, and not touch it if it is specified, would be the best behaviour?

Thanks

stevencartavia:
`cast call --trace` ignored `--gas-limit` and always ran with an unbounded gas limit, so gas-dependent calls could revert under `--trace` but not without it.

Now the provided `--gas-limit` is applied to the traced call.

Fixes #15357

### DIFF
--- a/crates/cast/src/cmd/call.rs
+++ b/crates/cast/src/cmd/call.rs
@@ -440,13 +440,18 @@ impl CallArgs {
             let value = tx.value().unwrap_or_default();
             let input = tx.input().cloned().unwrap_or_default();
             let tx_kind = tx.kind().expect("set by builder");
+
+            // Apply a user-provided `--gas-limit` to the executor. `build_test_env` propagates the
+            // executor's gas limit to the executed call/deploy, so setting it here is what takes
+            // effect; writing it onto the tx env directly would be overwritten. When no limit is
+            // given, the executor keeps the block gas limit (`u64::MAX`) set above.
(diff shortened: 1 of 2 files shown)


mattsse (approved):
LGTM

mablr (approved):

PR #15360, opened 2026-06-24
feat(create): pass license type to verification

Osraka:
## Motivation

Follow-up to #14975. `forge verify-contract` can now pass an Etherscan `licenseType`, but the one-step `forge create --verify` path still built `VerifyArgs` with `license_type: None` in both verification call sites.

That means users who deploy and verify in one command cannot include the same license metadata unless they split the flow and call `forge verify-contract` manually afterward.

This intentionally does not overlap with #15162's SPDX parsing and validation work. This PR only forwards the raw Etherscan code through the existing create-and-verify path.

Refs #14475.

## Solution

- Add `--license-type <CODE>` to `forge create`, gated behind `--verify`.
- Pass the value into the preflight verification request.
- Pass the same value into the final verification request after deployment.
- Extend the existing `can_parse_create` parser test so the flag is covered without adding a broader integration fixture.

## Validation

- `git diff --check`
- `git diff --cached --check`
- `cargo test -p forge can_parse_create`
- `cargo check -p forge`

## AI assistance disclosure

I used ChatGPT/Codex to inspect the related issue/PR history, prepare the patch, and run local validation.

### DIFF
--- a/crates/forge/src/cmd/create.rs
+++ b/crates/forge/src/cmd/create.rs
@@ -89,7 +89,8 @@ pub struct CreateArgs {
     #[arg(long, requires = "verify")]
     show_standard_json_input: bool,
 
-    /// The Etherscan license type code or SPDX identifier to include with the verification request.
+    /// The Etherscan license type code or SPDX identifier to include with the verification
+    /// request.
     ///
     /// Accepts either an Etherscan numeric license code or a common SPDX identifier such as `MIT`.
     /// This is only used for Etherscan-style verifiers when `--verify` is enabled.
--- a/crates/forge/src/cmd/create.rs
+++ b/crates/forge/src/cmd/create.rs
@@ -89,6 +89,13 @@ pub struct CreateArgs {
     #[arg(long, requires = "verify")]
     show_standard_json_input: bool,
 
+    /// The Etherscan license type code to include with the verification request.
+    ///
+    /// See Etherscan's supported `licenseType` values. This is only used for Etherscan-style
+    /// verifiers when `--verify` is enabled.
+    #[arg(long, requires = "verify", value_name = "CODE", help_heading = "Verifier options")]
+    license_type: Option<String>,
+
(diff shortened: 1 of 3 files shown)


crates/forge/src/cmd/create.rs
```diff
@@ -89,6 +89,13 @@ pub struct CreateArgs {
     #[arg(long, requires = "verify")]
     show_standard_json_input: bool,
 
+    /// The Etherscan license type code to include with the verification request.
+    ///
+    /// See Etherscan's supported `licenseType` values. This is only used for Etherscan-style
+    /// verifiers when `--verify` is enabled.
+    #[arg(long, requires = "verify", value_name = "CODE", help_heading = "Verifier options")]
+    license_type: Option<String>,
```
mablr: This adds a separate `--license-type` parser for forge create, but it does not reuse the current `VerifyArgs` parser. `forge verify-contract --license-type` MIT is normalized to licenseType=3, while `forge create --verify --license-type MIT` would forward "MIT" directly into `VerifyArgs` and then into the Etherscan licenseType field. Consider sharing/reusing `parse_etherscan_license_type` here and adding create parser coverage for SPDX and invalid values.

mablr (dismissed):

Osraka:
Addressed the parser reuse feedback and the follow-up rustfmt failure.

Changes since the review:
- `forge create --verify --license-type ...` now reuses `parse_etherscan_license_type`, matching `forge verify-contract` normalization.
- Added create parser coverage for SPDX input and invalid values.
- Applied the rustfmt line wrap reported by CI.

Local verification:
- `cargo +nightly fmt --all --check`
- `cargo test -p forge --lib create_license_type`
- `cargo test -p forge-verify --lib license_type`

The latest workflow runs are currently `action_required`, so they should need maintainer approval before rerunning on the new head commit.

grandizzy (approved):
lgtm, ty!

mablr (approved):

PR #15361, opened 2026-06-24
chore: fix `large_stack_frames` clippy warning in `run_script`

stevencartavia:
Box per-branch futures in `run_script` to keep large async state off `run_command`'s frame, fixing `clippy::large_stack_frames`.


mattsse (approved):
LGTM

PR #15364, opened 2026-06-25
fix(cast): reject unclaimable receive-policy recovery authorities

stevencartavia:
`cast receive-policy set` now rejects recovery authorities that can never pass `ReceivePolicyGuard.claim()` (TIP-1022 virtual, TIP-20, and fixed system precompile addresses), so held TIP-1028 funds can't be stranded. Mirrors the protocol's onchain check (tempo v1.9.0). Also adds the T7 `StorageCredits` precompile to Foundry's Tempo precompile list.

Part of OSS-450


mattsse (approved):
LGTM

mablr (approved):

PR #15366, opened 2026-06-25
feat(fuzz): prefer fixed invariant sender pool

decofe:
Prefer the existing fixed default invariant sender pool, retaining rare random/dictionary exploration and explicit sender filters. This refresh moves the proposal into the current transaction generator without changing its selection policy.

Independent of the corpus and mutator PRs. Derek comparison baseline: `bfe6e9a17` (master used for this refresh); performance and fuzzing-effectiveness assessment remain pending.

Originally prompted by @0xalpharush. AI assistance was used to refresh the existing code, resolve API changes, and update this description.


0xalpharush:
derek bench invariant

crates/evm/fuzz/src/strategies/invariants.rs
```diff
@@ -7,13 +7,27 @@ use crate::{
...
+/// The Foundry default deployer is included because owner/deployer-only paths are common in
+/// invariant targets.
+const DEFAULT_INVARIANT_SENDERS: [Address; 3] = [
+    address!("0x0000000000000000000000000000000000010000"),
+    address!("0x0000000000000000000000000000000000020000"),
+    CALLER,
+];
+
+const RANDOM_SENDER_WEIGHT: u32 = 1;
+const DEFAULT_SENDER_WEIGHT: u32 = 99;
```
stevencartavia, line 35: Proposal: split the sender weighting from `--invariant-dictionary-weight`. After this change, when `targetSenders` is empty the sender comes 99% from the fixed pool and only 1% from the random/dictionary strategy, so the knob barely affects which address sends the tx anymore.

To check, I counted how many of 400 sender draws were **not one of the three new fixed-pool addresses**. On `master`, raising the dictionary weight from 0 to 100 changes that count from 400 down to ~203 — so the knob clearly drives sender selection. On this PR it stays at ~2–3 regardless of the weight.

Maybe we can add a dedicated weight for the fixed-pool-vs-random/dictionary split, and leave `dictionary_weight` to govern calldata plus the fallback branch. That keeps the new Echidna-style default while still letting users dial sender randomness back up, and avoids making the existing help text (`invariant calldata/senders`) misleading.

grandizzy:

0xalpharush:

0xalpharush:
derek bench invariant

PR #15368, opened 2026-06-25
feat(fuzz): use minset corpus culling

decofe:
Replace mutation-count eviction with the existing top-rated minset culling proposal, favoring shorter representatives of covered edges. This refresh preserves current campaign execution, immediate persistence from #16065, pending worker synchronization and fan-out, and edge bookkeeping when calls are popped. It does not include new mutations, observed-call dictionaries, or scheduling.

Production generation still samples the in-memory corpus only: cache/disk donor selection remains test-only and is an explicit unresolved review item. This is a refresh of the existing proposal, not a claim that the donor design is complete. Derek comparison baseline: master `bfe6e9a17`; warmed-corpus effectiveness and memory/I/O assessment remain pending, as does the book update for removing corpus_min_mutations.

Originally prompted by @0xalpharush. AI assistance was used for the refresh, conflict resolution, test adaptation, and this description.

### DIFF
--- a/crates/evm/evm/src/executors/corpus.rs
+++ b/crates/evm/evm/src/executors/corpus.rs
@@ -59,7 +59,11 @@ use foundry_evm_fuzz::{
 };
 #[cfg(test)]
 use proptest::prelude::Strategy;
+#[cfg(test)]
+use proptest::test_runner::TestRng;
 use proptest::test_runner::TestRunner;
+#[cfg(test)]
+use rand::Rng;
 use serde::{Deserialize, Serialize};
 use std::{
     collections::{HashMap, HashSet, VecDeque},
@@ -192,7 +196,7 @@ impl CorpusEntry {
 struct CachedDiskCorpus {
     descriptors: Vec<CorpusDirEntry>,
     descriptor_indices: HashMap<Uuid, usize>,
-    cache: VecDeque<CorpusEntry>,
+    cache: VecDeque<Arc<CorpusEntry>>,
     cache_max_len: usize,
 }
 
@@ -217,7 +221,16 @@ impl CachedDiskCorpus {
         }
     }
 
-    fn cache_entry(&mut self, corpus: CorpusEntry) {
+    #[cfg(test)]
+    fn remove_descriptor_at(&mut self, index: usize) {
+        let removed = self.descriptors.swap_remove(index);
+        self.descriptor_indices.remove(&removed.uuid);
+        if let Some(moved) = self.descriptors.get(index) {
+            self.descriptor_indices.insert(moved.uuid, index);
+        }
+    }
+
+    fn cache_entry(&mut self, corpus: Arc<CorpusEntry>) {
         if self.cache_max_len == 0 {
             return;
         }
@@ -227,6 +240,71 @@ impl CachedDiskCorpus {
         self.cache.push_back(corpus);
         while self.cache.len() > self.cache_max_len {
             self.cache.pop_front();
+        }
+    }
+
+    fn retain_replayable(&mut self, is_replayable: impl Fn(&[BasicTxDetails]) -> bool) {
+        let cached_replayability = self
+            .cache
+            .iter()
+            .map(|entry| (entry.uuid, is_replayable(&entry.tx_seq)))
+            .collect::<HashMap<_, _>>();
+        self.cache.retain(|entry| is_replayable(&entry.tx_seq));
+        self.descriptors.retain(|descriptor| {
+            cached_replayability.get(&descriptor.uuid).copied().unwrap_or_else(|| {
+                descriptor
+                    .read_tx_seq()
+                    .is_ok_and(|tx_seq| !tx_seq.is_empty() && is_replayable(&tx_seq))
+            })
+        });
+        self.descriptor_indices.clear();
+        self.descriptor_indices
+            .extend(self.descriptors.iter().enumerate().map(|(index, entry)| (entry.uuid, index)));
+    }
+
+    #[cfg(test)]
+    fn random_entry(
+        &mut self,
+        rng: &mut TestRng,
+    ) -> foundry_common::fs::Result<Option<Arc<CorpusEntry>>> {
+        loop {
+            if self.cache.is_empty() && self.descriptors.is_empty() {
+                return Ok(None);
+            }
+
+            if self.descriptors.is_empty() {
+                let index = rng.random_range(0..self.cache.len());
+                return Ok(self.cache.get(index).cloned());
+            }
+
+            let descriptor_index = rng.random_range(0..self.descriptors.len());
+            let descriptor = self.descriptors[descriptor_index].clone();
+            if let Some(cached) = self.cache.iter().find(|entry| entry.uuid == descriptor.uuid) {
+                return Ok(Some(Arc::clone(cached)));
+            }
+
+            let tx_seq = match descriptor.read_tx_seq() {
+                Ok(tx_seq) if !tx_seq.is_empty() => tx_seq,
+                Ok(_) => {
+                    self.remove_descriptor_at(descriptor_index);
+                    continue;
+                }
+                Err(err) => {
+                    debug!(target: "corpus", path=%descriptor.path.display(), %err, "removing unreadable corpus donor");
+                    self.remove_descriptor_at(descriptor_index);
+                    continue;
+                }
+            };
+            let mut corpus = CorpusEntry::new_with_cmp_and_edges(
+                tx_seq,
+                Vec::new(),
+                Vec::new(),
+                descriptor.uuid,
+            );
+            corpus.timestamp = descriptor.timestamp;
+            let corpus = Arc::new(corpus);
+            self.cache_entry(Arc::clone(&corpus));
+            return Ok(Some(corpus));
         }
     }
 }
@@ -472,9 +550,17 @@ impl WorkerCorpusSeed {
         let is_replayable =
             |tx_seq: &[BasicTxDetails]| tx_seq.iter().all(|tx| targeted_contracts.can_replay(tx));
         self.in_memory_corpus.retain(|entry| is_replayable(&entry.tx_seq));
+        self.disk_corpus.retain_replayable(is_replayable);
+        self.top_rated.clear();
+        for entry in &self.in_memory_corpus {
+            WorkerCorpus::update_top_rated_in(&mut self.top_rated, entry);
+        }
         self.metrics.corpus_count = self.in_memory_corpus.len();
-        self.metrics.favored_items =
-            self.in_memory_corpus.iter().filter(|entry| entry.is_favored).count();
+        WorkerCorpus::recompute_favored_for_entries(
+            &self.top_rated,
+            &mut self.in_memory_corpus,
+            &mut self.metrics,
+        );
 
         if !self.optimization_best_sequence.is_empty()
             && !is_replayable(&self.optimization_best_sequence)
@@ -550,6 +636,7 @@ impl WorkerCorpusSeed {
                 continue;
             }
 
+            seed.disk_corpus.push_descriptor(entry.clone());
             seed.metrics.corpus_count += 1;
             debug!(
                 target: "corpus",
@@ -1384,12 +1471,21 @@ impl WorkerCorpus {
         let mut old_to_new = vec![None; self.in_memory_corpus.len()];
         let mut retained = Vec::with_capacity(self.in_memory_corpus.len());
         let mut evicted_uuids = HashSet::new();
+        let unsynced_uuids = self
+            .new_entry_indices
+            .iter()
+            .filter_map(|&index| self.in_memory_corpus.get(index).map(|entry| entry.uuid))
+            .collect::<HashSet<_>>();
 
         for (old_index, corpus) in self.in_memory_corpus.drain(..).enumerate() {
-            if !corpus.is_favored && remaining_removals > 0 {
+            if !corpus.is_favored
+                && !unsynced_uuids.contains(&corpus.uuid)
+                && remaining_removals > 0
+            {
                 trace!(target: "corpus", corpus=%serde_json::to_string(&corpus).unwrap(), "evict corpus");
-                self.disk_corpus.cache_entry(corpus.clone());
-                evicted_uuids.insert(corpus.uuid);
+                let uuid = corpus.uuid;
+                self.disk_corpus.cache_entry(Arc::new(corpus));
+                evicted_uuids.insert(uuid);
                 remaining_removals -= 1;
             } else {
                 old_to_new[old_index] = Some(retained.len());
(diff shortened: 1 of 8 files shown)


crates/evm/evm/src/executors/corpus.rs
```diff
@@ -79,7 +79,8 @@ const CORPUS_DIR: &str = "corpus";
 const SYNC_DIR: &str = "sync";
 const OPTIMIZATION_BEST_FILE: &str = "optimization_best.json";
 
-const FAVORABILITY_THRESHOLD: f64 = 0.3;
+const SANCOV_EDGE_OFFSET: usize = usize::MAX / 2;
+const CACHED_DISK_CORPUS_MAX_LEN: usize = 64;
```
0xalpharush:
```suggestion
const CACHED_DISK_CORPUS_MAX_LEN: usize = 128;
```

0xalpharush:
derek bench invariant

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -1038,7 +1043,10 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
                     None
                 };
                 // Collect edge coverage and set the flag in the current run.
-                let new_call_coverage = corpus_manager.merge_edge_coverage(&mut call_result);
+                let new_call_coverage = corpus_manager.merge_edge_coverage_with_edges_into(
+                    &mut call_result,
+                    &mut current_run.corpus_edges,
```
stevencartavia, line 1257: A call's coverage edges are added to `current_run.corpus_edges` before the pop decision, but (unlike `cmp_seq`) `corpus_edges` isn't trimmed when the call is popped (`vm.assume` discard / reverted-and-dropped calls).

So a popped call's edges end up on the surviving sequence's `CorpusEntry` and in `top_rated`, which can then favor a sequence that can't actually reproduce that edge.

I reproduced this with a coverage-guided invariant test (call hits a branch then `vm.assume` discards it; another call survives) — no pops → clean. It seems to self-heal on restart, so not a blocker, but should we trim `corpus_edges` on pop to keep minset accuracy correct during a live run?

figtracer (dismissed):
lgtm

mablr (commented):
Should be re-benchmarked (once infra is ready), and de-conflicted.

grandizzy:

grandizzy:

grandizzy:
derek bench invariant

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -445,6 +529,7 @@ impl WorkerCorpusSeed {
         let mut seen_entries =
             seed.in_memory_corpus.iter().map(|entry| entry.uuid).collect::<HashSet<_>>();
         for entry in unique_corpus_entries(&canonical_replay_dirs(corpus_dir), &mut seen_entries) {
+            seed.disk_corpus.push_descriptor(entry.clone());
```
grandizzy, line 621: **Blocking:** this registers the descriptor before the file has passed read and replay validation.

A corrupt file is warned about and skipped below, but its descriptor remains in `disk_corpus`. When `random_entry` eventually selects it, `read_tx_seq()?` propagates the error through `new_input`/`new_inputs`, turning tolerated startup corruption into a campaign failure. Entries rejected by `keep_entry` have the same problem and may later fail ABI decoding.

A focused regression test reproduces the unreadable-file failure. Please add the descriptor only after a successful read and `keep_entry`, and make `random_entry` remove or skip descriptors that subsequently become unreadable rather than returning the error.

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -388,9 +470,11 @@ impl WorkerCorpusSeed {
 
         Self {
             in_memory_corpus,
+            disk_corpus: self.disk_corpus.clone(),
```
grandizzy: **Blocking:** this copies the complete disk donor pool, but `retain_replayable` later filters only `in_memory_corpus`; it does not filter these descriptors, the cache, or `top_rated`.

A selector-focused invariant worker can therefore draw and execute an excluded sequence from disk, while a removed UUID can remain the top-rated owner and prevent valid local entries from becoming favored. A focused regression test reproduces the retained off-target descriptor. Please filter every donor representation by retained UUID and rebuild `top_rated`/favored state afterward.

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1040,6 +1141,30 @@ impl WorkerCorpus {
...
+        rng: &mut TestRng,
+    ) -> foundry_common::fs::Result<Option<(Option<usize>, CorpusEntry)>> {
+        if self.in_memory_corpus.is_empty() {
+            return self
+                .disk_corpus
+                .random_entry(rng)
+                .map(|entry| entry.map(|entry| (None, entry)));
+        }
+
+        if rng.random::<bool>()
```
grandizzy: **Update for `0ba4d59b`:** using `Arc<CorpusEntry>` and returning in-memory indices fixes the deep-clone half of this comment. The thread remains open because the descriptor-first selection still performs fixed-rate disk decoding for warmed corpora.

With `N` descriptors and a 128-entry cache, selecting a descriptor first gives an approximate cache-hit rate of `128/N`; invariant generation therefore approaches one file read/parse per run for large `N`. A concrete cache-first/explore-disk shape is proposed in the follow-up thread: https://github.com/foundry-rs/foundry/pull/15368#discussion_r3655380467.

Please tune that exploration rate with a current-head warmed-corpus benchmark before resolving this thread.

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1122,48 +1314,41 @@ impl WorkerCorpus {
             return Ok(new_seq);
         };
 
-        if !self.in_memory_corpus.is_empty() {
-            self.evict_oldest_corpus()?;
+        if !self.in_memory_corpus.is_empty() || !self.disk_corpus.is_empty() {
+            self.cull_corpus()?;
```
grandizzy: `cull_corpus` already runs whenever corpus membership changes, but this calls it before every generated sequence.

If the favored minset contains more entries than `corpus_min_size`, `cull_corpus` allocates `old_to_new`, drains the corpus, and rebuilds the vector on every run even when every entry is favored and nothing can be removed. Please remove the per-generation cull or add a fast path that avoids allocations unless a removable non-favored entry exists.

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1519,31 +1684,55 @@ impl WorkerCorpus {
...
+        if evicted_uuids.is_empty() {
+            self.in_memory_corpus = retained;
+            return Ok(());
+        }
+
+        self.in_memory_corpus = retained;
+        self.new_entry_indices = self
+            .new_entry_indices
+            .iter()
+            .filter_map(|&i| old_to_new.get(i).copied().flatten())
```
grandizzy: This can remove an unsynced finding from `new_entry_indices`.

`top_rated` is keyed by edge index, so a run whose novelty is a higher hitcount bucket may not become favored when an equal-or-shorter entry already covers that edge. It is then culled immediately, removed here, and never seen by `export_to_master`, which exports only `new_entry_indices`.

Please either retain unsynced entries until after export or export persisted worker entries by timestamp instead of their current in-memory index.

grandizzy (changes requested):
Request changes. The minset selection itself looks reasonable, but the new disk-donor lifecycle introduces two reproduced correctness regressions: unreadable or replay-rejected descriptors remain selectable, and focused-worker filtering does not apply to disk donors or the top-rated map. The input-generation hot path also performs deterministic disk decoding and deep cloning at a frequency that needs to be fixed and benchmarked before merge.

CI and the existing focused corpus tests pass, but targeted regression assertions reproduce both donor-lifecycle failures. The existing inline comment about coverage retained from popped invariant calls is also valid and should be addressed without a duplicate comment here. Please fix these issues and provide a current-head warmed-corpus `foundry-scfuzzbench` comparison.

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1519,31 +1749,64 @@ impl WorkerCorpus {
...
+            .new_entry_indices
+            .iter()
+            .filter_map(|&index| self.in_memory_corpus.get(index).map(|entry| entry.uuid))
+            .collect::<HashSet<_>>();
 
-            // Remove corpus from memory.
-            self.in_memory_corpus.remove(index);
+        for (old_index, corpus) in self.in_memory_corpus.drain(..).enumerate() {
+            if !corpus.is_favored
+                && !unsynced_uuids.contains(&corpus.uuid)
```
grandizzy: **Blocking:** this protection never expires in invariant campaigns. Every insertion pushes an index into `new_entry_indices`, but those indices are only cleared by `WorkerCorpus::sync`, whose only production caller is the stateless fuzz path. Invariant workers never call `sync`.

As a result, every corpus entry discovered during either live or deferred invariant campaigns remains in `unsynced_uuids` forever and cannot be culled, so memory grows for the entire campaign. The new unit test clears the indices manually and therefore misses the production lifecycle.

A concrete fix is to thread an explicit export-tracking flag through insertion instead of treating every entry as sync-pending:

```rust
fn push_corpus_entry(&mut self, corpus: CorpusEntry, track_for_worker_sync: bool) {
    let new_index = self.in_memory_corpus.len();
    if track_for_worker_sync {
        self.new_entry_indices.push(new_index);
    }
    self.metrics.corpus_count += 1;
    self.in_memory_corpus.push(corpus);
    // recompute/cull as today
}
```

Pass `true` only from the stateless path that actually calls `WorkerCorpus::sync`; pass `false` from both invariant persistence modes and observed-call insertion. A regression test should insert repeated equal-cost/same-edge invariant entries without manually clearing indices and assert that the in-memory corpus stays bounded.

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -230,6 +237,139 @@ impl CorpusEntry {
...
+            if self.cache.is_empty() && self.descriptors.is_empty() {
+                return Ok(None);
+            }
+
+            if self.descriptors.is_empty() {
+                let index = rng.random_range(0..self.cache.len());
+                return Ok(self.cache.get(index).cloned());
+            }
+
+            let descriptor_index = rng.random_range(0..self.descriptors.len());
```
grandizzy, line 284: The warmed-corpus I/O concern remains and is worse for sufficiently large corpora. This uniformly chooses one of `N` descriptors and only hits cache when that exact UUID is among the 128 cached entries, so the hit rate is approximately `128/N`.

Because invariant generation takes the disk branch with probability 0.5 for each of two donors, expected file reads/parses per run approach `2 × 0.5 × (1 - 128/N)`, i.e. nearly one disk decode per run for large `N`.

A concrete shape is to separate cache exploitation from occasional descriptor exploration:

```rust
const DISK_REFRESH_DENOMINATOR: u32 = 16; // tune from benchmark

if !self.cache.is_empty()
    && (self.descriptors.is_empty()
        || !rng.random_ratio(1, DISK_REFRESH_DENOMINATOR))
{
    let index = rng.random_range(0..self.cache.len());
    return Ok(self.cache.get(index).cloned());
}

// Otherwise select a descriptor, return its cached UUID if present,
// or read/decode it and populate the cache.
```

This bounds steady-state descriptor reads while still rotating through the full disk corpus. Please tune the ratio using a current-head warmed-corpus throughput/RSS benchmark rather than landing the example constant blindly.

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -230,6 +237,139 @@ impl CorpusEntry {
...
+            let descriptor = self.descriptors[descriptor_index].clone();
+            if let Some(cached) = self.cache.iter().find(|entry| entry.uuid == descriptor.uuid) {
```
grandizzy, line 286: This descriptor-first selection also makes cache-only entries unreachable whenever any descriptor exists. Deferred-mode entries are cached when culled but have no descriptor, so `random_entry` can never select them in a warmed campaign; they only consume cache capacity and evict descriptor-backed entries.

The cache-first branch suggested in the adjacent thread fixes this without another data structure because it samples every cached entry, including descriptor-less evictees. If descriptor-first uniformity must be retained instead, keep cache-only donors in a separate bounded queue and include that queue explicitly in the selection distribution.

Please add a regression test containing one descriptor-backed donor and one cache-only donor; force the cache-selection branch and verify that the cache-only UUID can be returned.

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -230,6 +237,139 @@ impl CorpusEntry {
...
+    fn retain_replayable(&mut self, is_replayable: impl Fn(&[BasicTxDetails]) -> bool) {
+        let cached_replayability = self
+            .cache
+            .iter()
+            .map(|entry| (entry.uuid, is_replayable(&entry.tx_seq)))
+            .collect::<HashMap<_, _>>();
+        self.cache.retain(|entry| is_replayable(&entry.tx_seq));
+        self.descriptors.retain(|descriptor| {
+            cached_replayability.get(&descriptor.uuid).copied().unwrap_or_else(|| {
+                descriptor
```
grandizzy: At worker startup the disk cache is empty, so this rereads and reparses every descriptor for every selector-focused worker after `load_from_disk` already read the same files. With `F` focus workers and `N` warmed entries this adds `F × N` redundant file decodes before fuzzing starts.

The smallest no-I/O approach is to filter the full seed before sharding and retain disk entries by UUID:

```rust
// WorkerCorpusSeed::retain_replayable
self.in_memory_corpus.retain(|entry| is_replayable(&entry.tx_seq));
let allowed = self.in_memory_corpus.iter().map(|entry| entry.uuid).collect::<HashSet<_>>();
self.disk_corpus.retain_uuids(&allowed); // retain cache/descriptors; rebuild descriptor_indices
```

Then, in `campaign_seed_and_corpus_seed_for_worker`, clone/filter the unsharded seed for the focused target set before calling `clone_for_worker`. `load_from_disk` now creates each descriptor together with a validated in-memory entry, so the UUID set is authoritative and no descriptor needs to be decoded again. Add a test that deletes the backing files after seed load and confirms focused-worker filtering still succeeds from seed metadata.

grandizzy (changes requested):
Follow-up review of `0ba4d59b`: the descriptor validation/pruning, focused-worker filtering, deep-clone removal, per-generation cull removal, and stateless pre-export retention changes address their original threads, which are now resolved.

Changes are still required because the new unsynced guard never expires in invariant campaigns and therefore disables in-run culling, while the revised descriptor-first cache selection retains (and for large corpora increases) hot-path disk decoding and makes cache-only deferred donors unreachable. `retain_replayable` also rereads the warmed corpus once per focused worker. The existing unresolved thread about edges retained from popped invariant calls remains valid.

Latest CI and 31 focused corpus tests pass, but they do not exercise the invariant no-sync lifecycle or warmed-corpus I/O. The most recent benchmark request targets prior head `6ebe71c3`, not this commit.

crates/evm/evm/src/executors/invariant/mod.rs
grandizzy: A minimal fix can keep an edge checkpoint parallel to the existing input/cmp bookkeeping:

```rust
let call_edges_start = current_run.corpus_edges.len();
let new_call_coverage = corpus_manager.merge_edge_coverage_with_edges_into(
    &mut call_result,
    &mut current_run.corpus_edges,
);

if discarded {
    current_run.corpus_edges.truncate(call_edges_start);
    current_run.inputs.pop();
    // ...
}
```

Use the same `truncate(call_edges_start)` in the later reverted-and-dropped branch immediately after `current_run.inputs.pop()`. This prevents the retained sequence from claiming the popped call's edges without requiring a second per-call edge vector. A regression test should cover both `MAGIC_ASSUME` and the non-`fail_on_revert` pop path.

mablr:
derek bench invariant

Issue #12435, opened 2025-11-03
`cast run` result is different from the actual transaction


zzh1996:
### Component

Cast

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

1.4.3-stable

### What version of Foundryup are you on?

foundryup: 1.3.0

### What command(s) is the bug in?

cast run

### Operating System

Linux

### Describe the bug

```
$ cast run 0xae290fe8c89c3e83dff20eeb2b8e3261bcdce0d66441c7056918dfb5fafe6d96
Executing previous transactions from the block.
Traces:
  [14185] 0xB731392c0EB5BF2092f9f7B520DA551f70Ea9131::Claim{value: 46698476594582387}()
    ├─ [2204] 0x000F3df6D732807Ef1319fB7B8bB8522d0Beac02::00000000(00000000000000000000000000000000000000000000000069091d4b) [staticcall]
    │   └─ ← [Revert] EvmError: Revert
    ├─ [0] 0x6e6C36B970f8862bA3F148DEdAB8F98f5ed8b426::fallback{value: 46698476594582387}()
    │   └─ ← [Stop]
    ├─ emit NvStuck(param0: 0x6e6C36B970f8862bA3F148DEdAB8F98f5ed8b426, param1: 46698476594582387 [4.669e16], param2: 1762205003 [1.762e9])
    └─ ← [Stop]


Transaction successfully executed.
Gas used: 35249
```

The trace and gas used is different from [the actual transaction](https://app.blocksec.com/explorer/tx/eth/0xae290fe8c89c3e83dff20eeb2b8e3261bcdce0d66441c7056918dfb5fafe6d96), which does not have the revert and the internal value transfer.

mattsse:
oh, I see

idk what https://etherscan.io/address/0x000f3df6d732807ef1319fb7b8bb8522d0beac02/ is exactly but this appears to be blob related, so mostlikely due to missing system EIP-4788 pre block system call this is causing the claim to fail.

integrating the systemcalls properly is currently missing

grandizzy:
> oh, I see
> 
> idk what https://etherscan.io/address/0x000f3df6d732807ef1319fb7b8bb8522d0beac02/ is exactly but this appears to be blob related, so mostlikely due to missing system EIP-4788 pre block system call this is causing the claim to fail.
> 
> integrating the systemcalls properly is currently missing

[...]

PR #15371, opened 2026-06-25
Implement EIP 4788

tskoyo:
Closing https://github.com/foundry-rs/foundry/issues/12435

## Motivation

EIP-4788 implementation

### cast run before fixing bug
```
cast run 0xae290fe8c89c3e83dff20eeb2b8e3261bcdce0d66441c7056918dfb5fafe6d96

Executing previous transactions from the block.
Traces:
  [14185] 0xB731392c0EB5BF2092f9f7B520DA551f70Ea9131::Claim{value: 46698476594582387}()
    ├─ [2204] 0x000F3df6D732807Ef1319fB7B8bB8522d0Beac02::00000000(00000000000000000000000000000000000000000000000069091d4b) [staticcall]
    │   └─ ← [Revert] EvmError: Revert
    ├─ [0] 0x6e6C36B970f8862bA3F148DEdAB8F98f5ed8b426::fallback{value: 46698476594582387}()
    │   └─ ← [Stop]
    ├─ emit NvStuck(param0: 0x6e6C36B970f8862bA3F148DEdAB8F98f5ed8b426, param1: 46698476594582387 [4.669e16], param2: 1762205003 [1.762e9])
    └─ ← [Stop]
```

It would revert because we were not writing beacon root to the storage of `BEACON_ROOT_ADDRESS` (0x000F3df6D732807Ef1319fB7B8bB8522d0Beac02)

### Running tests from this issue https://github.com/foundry-rs/foundry/issues/12435
```
cargo run -p cast -- run 0xae290fe8c89c3e83dff20eeb2b8e3261bcdce0d66441c7056918dfb5fafe6d96 --rpc-url my-rpc-url

Executing previous transactions from the block.
Traces:
  [45054] 0xB731392c0EB5BF2092f9f7B520DA551f70Ea9131::Claim{value: 46698476594582387}()
    ├─ [4320] 0x000F3df6D732807Ef1319fB7B8bB8522d0Beac02::00000000(00000000000000000000000000000000000000000000000069091d4b) [staticcall]
    │   └─ ← [Return] 0x70c7855161ec07af782df915fb3e81702df40f34972da3d740cdfc132ac926f6
    ├─ emit NvStuck(param0: 0x6e6C36B970f8862bA3F148DEdAB8F98f5ed8b426, param1: 46698476594582387 [4.669e16], param2: 1762205003 [1.762e9])
    └─ ← [Stop]
```

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [] Breaking changes


crates/cast/src/cmd/run.rs
```diff
@@ -213,9 +213,16 @@ impl RunArgs {
         evm_env.cfg_env.limit_contract_code_size = None;
         evm_env.block_env.set_number(U256::from(tx_block_number));
 
+        let mut parent_beacon_block_root = None;
         if let Some(block) = &block {
             evm_env.block_env = block_env_from_header(block.header());
 
+            let spec_id = (*evm_env.cfg_env.spec()).into();
+
+            if spec_id.is_enabled_in(SpecId::CANCUN) {
```
figtracer: this uses the config spec before the block hardfork is resolved below, so a project pinned to shanghai can skip the beacon-root write even though cast run later replays the block as cancun.

tskoyo: Thanks, resolved it

DaniPopes (dismissed):

mablr (approved):

figtracer (approved):

mattsse (commented):
Thanks for picking this up. I think there are two correctness issues to address before this fully matches the EIP-4788 pre-block transition semantics.

1. The beacon-root write is currently executed through the normal transaction path here: https://github.com/foundry-rs/foundry/blob/4bab5d4cbae5e28a93af3b4a2587dea492058040/crates/evm/evm/src/executors/mod.rs#L581-L586

   That means revm runs normal tx validation/pre-execution/post-execution and commits the normal transaction changeset. For system calls this is observably different: the dedicated system-call path skips those phases and only keeps the target system contract state, avoiding caller/beneficiary side effects like touching or incrementing `SYSTEM_ADDRESS`.

   Suggestion: route this helper through the EVM system-call API (`transact_system_call` / `system_call_with_caller`, or the alloy system-call helper) and commit only the returned system-call state instead of building a normal `TxEnv` via `transact_raw`.

2. For Cancun+ execution this skips the beacon-root write when the header did not include `parent_beacon_block_root`: https://github.com/foundry-rs/foundry/blob/4bab5d4cbae5e28a93af3b4a2587dea492058040/crates/cast/src/cmd/run.rs#L268-L271

   On a Cancun block that field is mandatory. If the RPC response omits it, silently continuing leaves `cast run` in the same missing-beacon-root state this change is trying to fix, which can produce a misleading replay rather than a clear setup error.

   Suggestion: once the resolved spec is Cancun+, require `parent_beacon_block_root` and return an explicit error when it is `None`, matching the canonical `MissingParentBeaconBlockRoot` behavior.

tskoyo:
@mattsse, wasn't aware that we should use dedicated path for system calls. Thanks for the detailed review!

PR #15372, opened 2026-06-25
perf(invariant): cache dynamic target matches

figtracer:
Avoid repeatedly walking all compiled artifacts for the same touched deployed bytecode during invariant dynamic target discovery.

Updatable invariant runs scan touched accounts and match deployed bytecode back to project artifacts. A factory-created target can be touched many times, so repeatedly calling `find_by_deployed_code` pays the full artifact scan for bytecode that has already been identified at that address.

This caches the deployed-bytecode match by touched address and `Bytecode::hash_slow()` while preserving the existing target refresh and `created_contracts` cleanup behavior.

### What changed

- Cache dynamic target artifact matches by touched address and bytecode hash.
- Share the cached ABI event lookup across refreshed target entries.
- Preserve setup-contract, untouched-account, target refresh, and cleanup semantics.
- Add regression coverage for cache reuse without a second project artifact lookup.

### Benchmarks

Release `forge` binaries were built from the same base commit (`cd5b4cf9a`). The fixture uses a factory-created dynamic target, repeated invariant calls that touch it, and 200 similar artifacts to make repeated rediscovery visible.

| Case | Baseline | This PR | Speedup |
| --- | ---: | ---: | ---: |
| Depth 20,000, `--threads 1` | `364.9ms ± 2.3ms` | `295.2ms ± 2.6ms` | `1.24x ± 0.01` |
| Depth 100,000, `--threads 1` | `1.592s ± 0.018s` | `1.235s ± 0.007s` | `1.29x ± 0.02` |

### DIFF
--- a/crates/evm/fuzz/src/invariant/mod.rs
+++ b/crates/evm/fuzz/src/invariant/mod.rs
@@ -1,5 +1,5 @@
 use alloy_json_abi::{Event, Function, JsonAbi};
-use alloy_primitives::{Address, B256, Selector, map::HashMap};
+use alloy_primitives::{Address, B256, Selector, keccak256, map::HashMap};
 use foundry_compilers::artifacts::StorageLayout;
 use itertools::Either;
 use serde::{Deserialize, Serialize};
@@ -36,12 +36,17 @@ pub struct FuzzRunIdentifiedContracts {
     pub targets: Rc<RefCell<TargetedContracts>>,
     /// Whether target contracts are updatable or not.
     pub is_updatable: bool,
+    artifact_matches: Rc<RefCell<HashMap<B256, Option<CachedTargetContract>>>>,
 }
 
 impl FuzzRunIdentifiedContracts {
     /// Creates a new `FuzzRunIdentifiedContracts` instance.
     pub fn new(targets: TargetedContracts, is_updatable: bool) -> Self {
-        Self { targets: Rc::new(RefCell::new(targets)), is_updatable }
+        Self {
+            targets: Rc::new(RefCell::new(targets)),
+            is_updatable,
+            artifact_matches: Rc::new(RefCell::new(HashMap::default())),
+        }
     }
 
     /// Borrows the current targeted contracts.
@@ -63,7 +68,6 @@ impl FuzzRunIdentifiedContracts {
             return Ok(());
         }
 
-        let mut targets = self.targets.borrow_mut();
         for (address, account) in state_changeset {
             if setup_contracts.contains_key(address) {
                 continue;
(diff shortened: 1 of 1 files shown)


crates/evm/fuzz/src/invariant/mod.rs
```diff
@@ -77,26 +81,56 @@ impl FuzzRunIdentifiedContracts {
...
-            let Some((artifact, contract_data)) =
-                project_contracts.find_by_deployed_code(code.original_byte_slice())
-            else {
-                continue;
+            let code = code.original_byte_slice();
+            let code_hash = if account.info.code_hash == B256::ZERO {
+                keccak256(code)
+            } else {
+                account.info.code_hash
             };
```
grandizzy: Rather than reading `account.info.code_hash` and hand-rolling a `B256::ZERO` fallback (which also misses the `KECCAK_EMPTY` sentinel for unset hashes), use revm's `Bytecode::hash_slow()`. It returns `keccak256(original_byte_slice())` and memoizes it on the `Bytecode` via `OnceCell`, so it's always correct and effectively free on repeat touches — no fallback branch and no `keccak256` import needed.

```suggestion
            let code_hash = code.hash_slow();
            let code = code.original_byte_slice();
```

`code` here is still the `&Bytecode` (`account.info.code`), so `hash_slow()` is called before shadowing it with `original_byte_slice()`. The `keccak256` import added in this PR can then be dropped.

grandizzy (commented):
Two things worth addressing before merge, neither a hard blocker:

1. **Code-hash derivation** — prefer revm's `Bytecode::hash_slow()` over the hand-rolled `code_hash == B256::ZERO` fallback. It's always correct, memoized on the `Bytecode`, and removes the `KECCAK_EMPTY`-sentinel edge case that the current branch silently mishandles for manually-built state (incl. the new test helper).
2. **Cache value** — cache hits still rebuild `event_lookup` (re-parse ABI events) every touch via `TargetedContract::new`; that index is address-independent and can be shared via `Arc`.

Also flagging unbounded cache growth as a non-blocking monitor item for long factory-heavy campaigns.

crates/evm/fuzz/src/invariant/mod.rs
```diff
@@ -108,6 +142,23 @@ impl FuzzRunIdentifiedContracts {
...
+}
+
+impl CachedTargetContract {
+    fn into_targeted_contract(self) -> TargetedContract {
+        let mut contract = TargetedContract::new(self.identifier, self.abi);
+        contract.targeted_functions = self.targeted_functions;
+        contract.storage_layout = self.storage_layout;
+        contract
+    }
+}
```
grandizzy, line 169: Cache hits still rebuild `event_lookup` every touch: `into_targeted_contract` → `TargetedContract::new` re-parses all ABI events into a fresh `Arc<TargetedContractEvents>`, even though that index is purely ABI-derived and address-independent. Build it once at insert time and share the `Arc` so hits only pay an `Arc::clone`.

```suggestion
#[derive(Clone, Debug)]
struct CachedTargetContract {
    identifier: String,
    abi: JsonAbi,
    targeted_functions: Vec<Function>,
    storage_layout: Option<Arc<StorageLayout>>,
    event_lookup: Arc<TargetedContractEvents>,
}

impl CachedTargetContract {
    fn into_targeted_contract(self) -> TargetedContract {
        TargetedContract {
            identifier: self.identifier,
            abi: self.abi,
            targeted_functions: self.targeted_functions,
            excluded_functions: Vec::new(),
            storage_layout: self.storage_layout,
            event_lookup: self.event_lookup,
        }
    }
}
```

And populate it once on the miss path in `target_contract_for_code`:
```rust
    artifact_filters.get_targeted_functions(artifact, &contract_data.abi)?.map(
        |targeted_functions| CachedTargetContract {
            identifier: artifact.name.clone(),
            event_lookup: Arc::new(TargetedContractEvents::new(&contract_data.abi)),
            abi: contract_data.abi.clone(),
            targeted_functions,
            storage_layout: contract_data.storage_layout.as_ref().map(Arc::clone),
        },
    )
```

crates/evm/fuzz/src/invariant/mod.rs
```diff
@@ -36,12 +36,17 @@ pub struct FuzzRunIdentifiedContracts {
     pub targets: Rc<RefCell<TargetedContracts>>,
     /// Whether target contracts are updatable or not.
     pub is_updatable: bool,
+    artifact_matches: Rc<RefCell<HashMap<B256, Option<CachedTargetContract>>>>,
```
grandizzy: The cache persists for the worker lifetime with one entry per distinct deployed code hash and is never evicted. Fine for typical runs, but long campaigns with factories emitting unique runtime bytecode (e.g. differing immutables) will accumulate both positive and negative entries unbounded. Not a blocker — just flagging; a simple size cap on negative entries would suffice if it ever shows up in profiles.

mablr (approved):
sgtm

grandizzy (approved):
lgtm!

0xKarl98 (approved):

PR #15374, opened 2026-06-26
perf(symbolic): reduce execution overhead

DaniPopes:
This PR is a purely mechanical performance pass over the symbolic execution internals. It focuses on reducing allocation, clone, selector-dispatch, visitor/fold, and lookup overhead through data-structure and API cleanup, with very little semantic change intended. The behavior-touching pieces are narrow mechanical cleanups around existing constant evaluation, generated selectors, checked conversions, and helper placement.

<details><summary><strong>Codex goal (steered over time)</strong></summary>
<p>

optimize symbolic execution

/home/doni/github/solady/foundry-symbolic-check-saturating-add-20260625-232936.json.gz this is a samply profile of one test run symbolically in solady, it highlights the major performance issues with the current implementation.

optimize the internal representation and structure of symbolic execution. most time is spent cloning and dropping expressions.

first start by analyzing the entire structure, then make stuff as private as possible, for example internal fields of data structures. if used outside the module, try to see if they can't be converted to methods that only access the specific property needed outside. dont do this if it's just raw data structure like block, thats fine.

then, optimize the representation. start with the primitives expression, symword, memory etc.

aren't SymWord::Concrete and SymWord::Expr(Expr::Const) the same?

why btreemaps and btreesets? use alloy_primitives::map if/where possible.

reduce the size of expr, and use interning to avoid Drop implementation by storing &'a Expr if possible and making it Copyable cheap. same for BoolExpr.

Avoid String, prefer using a string interner (use struct wrapper around alloy primitives IndexSet). pass context around operations.

collect_vars and similar functions should be implemented with a generic `fn visit(&mut impl FnMut)`, to traverse the tree.

then move into the more high level stuff. cheatcodes for example call keccak256 all the time, this is dogshit. use the cheatcode interface generated by the sol macro. look at cheatcodes crates to see how to use it. you can use somethingCall::SELECTOR instead of the selector! macro. remove the selector! macro. use a single match in the implementation of cheatcodes. avoid special casing every single "not handled" cheatcode.

commit as you go in a new branch. make sure that tests pass in between major changes.

</p>
</details>

Before:

After:

image

---

The actual expression trees are still overwhelming and a lot of duplicates still exist even with the cache (x, not(x) queried separately etc), so the entire implementation iss till pretty slow. this will be addressed in a future PR.


mattsse (approved):
LGTM

mablr (approved):

Issue #15338, opened 2026-06-23
Bug: `forge verify-bytecode` reports missing local artifact when Etherscan V1 API returns migration error


amusingaxl:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

forge Version: 1.7.1 Commit SHA: 4072e48705af9d93e3c0f6e29e93b5e9a40caed8 Build Timestamp: 2026-05-08T07:50:55.527285345Z (1778226655) Build Profile: dist

### What version of Foundryup are you on?

foundryup: 1.8.6

### What command(s) is the bug in?

forge verify-bytecode

### Operating System

Linux

### Description

`forge verify-bytecode` fails with:

```text
Error: Build Error: Contract artifact not found locally
```

However, the local artifacts are present and the deployed bytecode matches the local artifacts when checked manually. With debug logging enabled, `forge verify-bytecode` appears to call Etherscan’s deprecated V1 API endpoint, which returns a V2 migration error. Foundry then surfaces this as a misleading local artifact error.

### Foundry Version

```text
forge Version: 1.7.1
Commit SHA: 4072e48705af9d93e3c0f6e29e93b5e9a40caed8
Build Timestamp: 2026-05-08T07:50:55.527285345Z (1778226655)
Build Profile: dist
```

### Command

Example command:

```bash
forge verify-bytecode \
  0xaf7a108b4fb0b2f65e1acc9e1a548abe482559c4 \
  src/VestedRewardsDistribution.sol:VestedRewardsDistribution \
  --chain 1 \
  --rpc-url "$ETH_RPC_URL" \
  --encoded-constructor-args 0x000000000000000000000000cbcfcd450de686894d3c5e7e8975cf23eef077b20000000000000000000000004e41488c19cd35eb4de3083fc3e204854c75c86a \
  -vv
```

### Actual Behavior

```text
Verifying bytecode for contract VestedRewardsDistribution at address 0xAf7a108B4fB0b2F65E1Acc9E1a548abe482559C4
No files changed, compilation skipped
Error: Build Error: Contract artifact not found locally
```

The same issue occurs for another contract in the same repo:

```bash
forge verify-bytecode \
  0x4e41488c19cd35eb4de3083fc3e204854c75c86a \
  src/synthetix/StakingRewards.sol:StakingRewards \
  --chain 1 \
  --rpc-url "$ETH_RPC_URL" \
[...]

PR #15377, opened 2026-06-26
fix(verify): resolve bytecode artifact paths

mablr:
## Description

Discovered while working on #15338.

`forge verify-bytecode` looked up artifacts using the raw relative contract path, while compiled artifacts are keyed by canonical paths, so valid artifacts were missed.

Solution: canonicalize the provided contract path, compile that target, and find the artifact by canonical path plus contract name.


figtracer (approved):
lgtm

mattsse (approved):
LGTM

PR #15379, opened 2026-06-26
feat(debugger): show precompile call clues

figtracer:
## Summary

- Annotate debugger call-op steps when the next child trace is a known precompile call.
- Show the decoded/precompile notice in the Variables pane and add a short `precompile call` clue to the source pane title.
- Cover both decoded precompile traces and debug-dump traces that only have the known precompile address/data/output.

Progresses #927.

### DIFF
--- a/crates/debugger/src/node.rs
+++ b/crates/debugger/src/node.rs
@@ -1,6 +1,7 @@
 use alloy_primitives::{Address, Bytes, hex};
 use foundry_evm_core::precompiles;
 use foundry_evm_traces::{CallKind, CallTrace, CallTraceArena};
+use revm::bytecode::opcode::OpCode;
 use revm_inspectors::tracing::types::{
     CallTraceStep, DecodedCallTrace, DecodedTraceStep, TraceMemberOrder,
 };
@@ -75,16 +76,19 @@ pub fn flatten_call_trace(arena: CallTraceArena, out: &mut Vec<DebugNode>) {
     ) {
         let mut pending = PendingNode { node_idx, steps_count: 0, step_offset: 0 };
         let mut next_step_offset = 0;
+        let mut last_step_idx: Option<usize> = None;
         let node = &arena.nodes()[node_idx];
         for order in &node.ordering {
             match order {
                 TraceMemberOrder::Call(idx) => {
                     let child_idx = node.children[*idx];
-                    if next_step_offset > 0
+                    if let Some(step_idx) = last_step_idx.take()
+                        && let Some(step) = node.trace.steps.get(step_idx)
+                        && is_call_like_op(step.op)
                         && let Some(notice) =
                             precompile_call_notice(&arena.nodes()[child_idx].trace)
                     {
-                        step_notices.push((node_idx, next_step_offset - 1, notice));
+                        step_notices.push((node_idx, step_idx, notice));
                     }
                     out.push(pending);
                     pending =
@@ -97,8 +101,11 @@ pub fn flatten_call_trace(arena: CallTraceArena, out: &mut Vec<DebugNode>) {
                     }
                     pending.steps_count += 1;
                     next_step_offset = step_idx.saturating_add(1);
+                    last_step_idx = Some(*step_idx);
                 }
-                _ => {}
+                _ => {
+                    last_step_idx = None;
+                }
             }
         }
         out.push(pending);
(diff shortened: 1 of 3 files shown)


crates/debugger/src/node.rs
```diff
@@ -89,9 +104,18 @@ pub fn flatten_call_trace(arena: CallTraceArena, out: &mut Vec<DebugNode>) {
...
-    inner(&arena, 0, &mut nodes);
+    let mut step_notices = Vec::new();
+    inner(&arena, 0, &mut nodes, &mut step_notices);
 
     let mut arena_nodes = arena.into_nodes();
+    for (node_idx, step_idx, notice) in step_notices {
+        if let Some(step) =
+            arena_nodes.get_mut(node_idx).and_then(|node| node.trace.steps.get_mut(step_idx))
+        {
+            step.decoded = Some(Box::new(DecodedTraceStep::Line(notice)));
```
grandizzy: **blocker: clobbers existing decoded data.** This overwrites the same `decoded` slot used for `DecodedTraceStep::InternalCall`, which `draw.rs` reads to drive internal-call scope/args. If the target step already has decoded metadata, it's silently lost. Only write when empty:

```rust
match step.decoded.as_deref_mut() {
    None => step.decoded = Some(Box::new(DecodedTraceStep::Line(notice))),
    Some(DecodedTraceStep::Line(line)) if line.is_empty() => *line = notice,
    Some(_) => {} // preserve existing decoded data
}
```

Please add a regression test where the parent step already has `InternalCall`.

crates/debugger/src/node.rs
```diff
@@ -64,17 +67,29 @@ pub fn flatten_call_trace(arena: CallTraceArena, out: &mut Vec<DebugNode>) {
...
         let node = &arena.nodes()[node_idx];
         for order in &node.ordering {
             match order {
                 TraceMemberOrder::Call(idx) => {
+                    let child_idx = node.children[*idx];
+                    if next_step_offset > 0
+                        && let Some(notice) =
+                            precompile_call_notice(&arena.nodes()[child_idx].trace)
+                    {
+                        step_notices.push((node_idx, next_step_offset - 1, notice));
```
grandizzy: **blocker: assumes the previous step is the call op.** `next_step_offset - 1` is only the call-op step under normal full-step ordering (`Step(call-op), Call, Step`). If opcode recording is filtered or ordering differs, this annotates an unrelated step. Track the last `Step(step_idx)` and only annotate when its opcode is call-like (CALL/STATICCALL/DELEGATECALL/CALLCODE/CREATE/CREATE2). Add tests for `Call` with no preceding step and `Call` preceded by a non-call op.

grandizzy (changes requested):
Please check the inline findings — two correctness issues to address before merge.

Non-blocking design notes:
- The decoded path keys off the literal `label == "PRECOMPILES"`, and the known path duplicates an address→name table. Both are brittle/hardcoded — at minimum share the label as a constant with the trace decoder, and note the table is standard-Ethereum-only (omits `CELO_TRANSFER`) and not hardfork/chain-aware.
- The known-path notice embeds full `input=`/`output=` hex — fine for sha256, but modexp/BLS/identity payloads can be huge in the TUI/dump; consider truncating the display.
- Decoded vs known notices render two different formats for the same event; prefer one shape.

**Question — should we support Tempo precompiles here?** The two paths aren't equally portable for a custom chain. Our trace decoder already emits `label == "PRECOMPILES"` for standard precompiles, so the decoded path picks those up for free — but it only labels its static decoder list (standard Ethereum), so Tempo-native precompiles (TIP20 system, keychain, fee-manager) get no label, and the hardcoded `known_precompile_name` table doesn't list their addresses either, so they'd be missed by both paths. If we want Tempo coverage, the clean extension point is the **decoder's precompile list** (which feeds the `"PRECOMPILES"` label and is already chain/hardfork-aware), not this debugger-side address table. Should we (a) keep this PR standard-Ethereum-only and follow up by adding Tempo precompiles to the decoder, or (b) make `known_precompile_name` chain-aware now? Worth being explicit in code that it's a standard-Ethereum fallback, since it silently duplicates recognition our decoder already does at higher fidelity.

grandizzy (approved):
lgtm!

mablr (approved):

PR #15380, opened 2026-06-26
perf(fuzz): reduce inactive inspector overhead

figtracer:
## Summary

Profile-guided fuzzer/debugger performance cleanup focused on removing avoidable hot-path work while preserving existing behavior.

## Changes

- Skip inactive inspector hooks, including `vm.record()` step work on opcodes that cannot affect recorded storage accesses.
- Reuse cached fuzz and invariant inputs, calldata ABI encoders, selector lookups, storage layout metadata, and setup-derived dictionary state where the inputs are stable.
- Cache non-mapping storage slot identification during invariant dictionary updates.
- Avoid trace rendering, trace identification, and rich `expectEmit` diagnostics when the result will not be shown or the run is still passing.
- Cache parsed secp256k1 private-key signers for repeated `vm.addr` and `vm.sign` cheatcodes.

## Benchmarks

Average user CPU over repeated runs.

| Workload | Before | After | Delta |
| --- | ---: | ---: | ---: |
| Solady, `--fuzz-runs 2048` | 90.49s | 78.99s | -11.50s (-12.7%) |
| Storage layout invariant, 500 x 500 | 3.49s | 2.81s | -0.68s (-19.5%) |
| Liquity invariant, 200 x 20 | 1.50s | 1.23s | -0.27s (-17.9%) |
| Optimism L2ToL1MessagePasser, `--fuzz-runs 128` | 10.23s | 8.42s | -1.81s (-17.7%) |
| Aave HubConfig, `--fuzz-runs 1024` (`vm.record()` opcode gate) | 11.51s | 10.36s | -1.15s (-10.0%) |
| Aave HubConfig, `--fuzz-runs 1024` (expectEmit deferral) | 11.90s | 10.95s | -0.95s (-8.0%) |
| Tempo BlockGasLimits invariant (storage slot cache) | 1.84s | 1.58s | -0.26s (-14.2%) |
| Solady SignatureCheckerLib, `--fuzz-runs 2048` | 1.05s | 0.91s | -0.13s (-12.8%) |
| Aave TokenizationSpoke.Permit, `--fuzz-runs 1024` | 0.72s | 0.68s | -0.05s (-6.6%) |
| Tempo TIP20, `--fuzz-runs 512` | 1.41s | 1.36s | -0.04s (-3.1%) |
| Solady FixedPointMathLib, `-s -vvvvv` | 0.352s | 0.294s | -0.058s (-16.5%) |
| Aave HubConfig, `-s -vvvvv` | 5.763s | 5.693s | -0.070s (-1.2%) |
| Suppressed successful traces, 4-workload total | 15.52s | 15.21s | -0.31s (-2.0%) |
| Machine-mode silent traces, 5-workload total (`--machine -vvvvv`) | 14.65s | 14.32s | -0.33s (-2.2%) |
| Calldata ABI encoder, 5-workload total | 18.71s | 18.65s | -0.06s (-0.3%) |


DaniPopes (commented):
makes sense, [evm2](https://github.com/DaniPopes/evm2) will have support for dynamically registering interest in specific opcodes to `step` into, so this kind of config can be carried over

mablr (commented):
SGTM.

Please fix conflicts.

mablr (dismissed):

mablr (dismissed):

crates/evm/fuzz/src/strategies/state.rs
```diff
@@ -440,79 +459,108 @@ impl FuzzDictionary {
...
+        if self.values_full() {
+            return;
+        }
+
+        let Some(code) = &account_info.code else {
+            return;
+        };
+        if self.push_bytecode_hashes.insert(account_info.code_hash) {
+            self.collect_push_bytes(ignore_metadata_hash(code.original_byte_slice()));
         }
```
grandizzy, line 542: **Behavior change: no-code accounts now get marked as "scanned"**

`insert_address` runs *before* the `let Some(code)` guard. Previously the address was inserted into `self.addresses` only when code existed.

Now an account first seen without code gets recorded in `addresses`. If code is later deployed at that same address (pre-funded / CREATE2 target, or an initial DB account that later receives code), the `self.addresses.contains(address)` early-return at the top of the function permanently skips its PUSH-byte scan, shrinking the fuzz dictionary vs. `master`.

The new `push_bytecode_hashes` dedup keys on `code_hash`, so it only rescues this when another address already contributed identical bytecode; a unique contract at a pre-touched address is never scanned.

Suggested fix, move `insert_address` below the code guard:

```suggestion
        let Some(code) = &account_info.code else {
            return;
        };
        self.insert_address(*address);
        if self.values_full() {
            return;
        }
        if self.push_bytecode_hashes.insert(account_info.code_hash) {
            self.collect_push_bytes(ignore_metadata_hash(code.original_byte_slice()));
        }
```

grandizzy (dismissed):
Please check finding inline

mablr (dismissed):

crates/evm/fuzz/src/strategies/invariants.rs
```diff
@@ -27,9 +33,9 @@ pub fn override_call_strat(
     let contracts_ref = contracts.clone();
     proptest::prop_oneof![
         80 => proptest::strategy::LazyJust::new(move || *target.read()),
-        20 => any::<prop::sample::Selector>()
-            .prop_map(move |selector| {
-                let (target, _) = selector.select(contracts_ref.iter());
+        20 => any::<prop::sample::Index>()
```
stevencartavia: `override_call_strat` was changed in `0498a2d4` from `any::<prop::sample::Selector>()` to `any::<prop::sample::Index>()` for the 20% override-target arm. Later, `4bbf0952` ("fix invariant strategy seed stability") changed `invariant_strat` back from `Index` to `Selector`, but left `override_call_strat` on `Index`.

Since `Index` and `Selector` are different proptest strategies and are not seed-equivalent, could this still change reproducibility for `call_override`/reentrancy invariant runs relative to master when that override-target arm is used? Is `override_call_strat` an intentional exception, or should it also use `Selector` for parity with master?

`invariant_reentrancy` would not catch this kind of divergence because it only asserts redacted failure output, whereas `invariant_selectors_weight` pins `--fuzz-seed 119` and exact selector hit counts.

mablr (dismissed):

crates/evm/fuzz/src/strategies/state.rs
```diff
@@ -450,79 +469,107 @@ impl FuzzDictionary {
...
     fn insert_push_bytes_values(&mut self, address: &Address, account_info: &AccountInfo) {
-        if self.config.include_push_bytes
-            && !self.addresses.contains(address)
-            && let Some(code) = &account_info.code
-        {
-            self.insert_address(*address);
-            if !self.values_full() {
-                self.collect_push_bytes(ignore_metadata_hash(code.original_byte_slice()));
-            }
+        if !self.config.include_push_bytes || self.addresses.contains(address) {
```
grandizzy: The `|| self.addresses.contains(address)` short-circuit runs before `push_bytecode_hashes` is checked, so a previously-seen address with *new* code never gets re-scanned — defeating the "once per bytecode" intent. Drop the address guard and let the `push_bytecode_hashes.insert(...)` below be the sole dedup; `insert_address` is an idempotent set insert so calling it again is harmless.

```suggestion
        if !self.config.include_push_bytes {
```

grandizzy (changes requested):
Nice cleanup overall — the cached static dispatch, deferred `expectEmit` formatting, and the various per-run caches (signers, ABI encoders, target strategies, storage layout) all look like solid wins on the hot paths.

One correctness concern in the new PUSH-byte bytecode cache (inline). The function doc now says values are collected "once for a given bytecode" and you added `push_bytecode_hashes` keyed by `code_hash`, but the early `self.addresses.contains(address)` guard still makes the scan address-scoped *before* the hash is ever consulted. So if an address that was already scanned later has different code (e.g. `vm.etch`, same-address redeploy patterns surfaced in the `StateChangeset`), the new bytecode's PUSH literals are silently skipped. Not a crash — just a quietly weaker fuzz dictionary that contradicts the new cache contract. Suggested fix inline; worth a same-address-new-bytecode regression test alongside the existing `duplicate_bytecode_push_bytes_are_collected_once`.

grandizzy (approved):
lgtm

mablr (approved):

PR #15382, opened 2026-06-26
fix(cheatcodes): avoid warming slots while recording state diffs

### LINKED ISSUE #15381: `vm.startStateDiffRecording()` disables `--isolate`'s per-call EIP-2929 access-list reset (gas under-counted)
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.7.1

foundryup: 1.9.0

Forge / cheatcodes/ `--isolate`

### Operating System

macOS (Apple Silicon)

### Summary

Under `--isolate`, each top-level call is executed as its own transaction, so its EIP-2929 access list (warm/cold addresses & storage slots) should start fresh. This works correctly — **unless `vm.startStateDiffRecording()` is active**. With state-diff recording on, the isolated call is metered as if storage slots warmed by a *previous* top-level call are still warm, so storage-heavy calls are under-counted by a large margin.

This makes it impossible to get an accurate, mainnet-equivalent gas measurement for a call while also recording its state diff in the same run.

`vm.recordLogs()` does **not** trigger this — only `startStateDiffRecording()`. Both `gasleft()` deltas and `vm.lastCallGas()` observe the under-counted value, so it is the metering that changes, not the way it is read.

### Environment

Reproduces on the latest stable and the latest nightly:

```
forge 1.7.1         (4072e487, 2026-05-08)   # latest stable
forge 1.7.2-nightly (4b83ca7d, 2026-06-26)   # latest nightly
```

### Minimal reproduction

```solidity
pragma solidity ^0.8.13;

contract Storage {
    uint256[256] private slots;
    function fill() external { for (uint256 i; i < 256; ++i) slots[i] = i + 1; }
    function sum() external view returns (uint256 s) { for (uint256 i; i < 256; ++i) s += slots[i]; }
}

contract IsolateStateDiffGasPOC is Test {
    Storage internal s;
    function setUp() public { s = new Storage(); s.fill(); }

    /// forge-config: default.isolate = true
    function test_isolate_recordingOff() public {
        s.sum();                          // call 1: isolated tx warms the 256 slots, tx ends
        s.sum();                          // call 2: fresh isolated tx -> slots should be COLD
        emit log_named_uint("lastCall recording OFF", vm.lastCallGas().gasTotalUsed);
    }

    /// forge-config: default.isolate = true
    function test_isolate_recordingOn() public {
        s.sum();                          // call 1 (identical)
        vm.startStateDiffRecording();
        s.sum();                          // call 2: identical, recording active
        emit log_named_uint("lastCall recording ON", vm.lastCallGas().gasTotalUsed);
    }
}
```

Run with `forge test -vv` (the inline `forge-config: default.isolate = true` enables isolate per test; `forge test --isolate -vv` behaves the same).

### Actual output

```
[PASS] test_isolate_recordingOff()
  lastCall recording OFF: 607565
[PASS] test_isolate_recordingOn()
  lastCall recording ON: 95565
```

`call 2` is the same call in both tests. Under `--isolate` it should cost the same (cold) in both. With `startStateDiffRecording()` active it is ~512k cheaper — the 256 slots are seen as warm instead of cold, i.e. the per-call access-list reset did not happen.

### Expected

`call 2` should be metered identically whether or not `vm.startStateDiffRecording()` is active — i.e. ~607k (cold) in both tests. State-diff recording should not affect the EVM's gas accounting under `--isolate`.

### Notes

- Only `startStateDiffRecording()` reproduces this; `recordLogs()` does not (gas stays correct).
- Same result whether isolate is enabled via the inline `forge-config` annotation or the `--isolate` CLI flag.

mablr:
## Summary

- Avoid warming storage slots when `vm.startStateDiffRecording()` observes `SLOAD`.
- Add an isolated `lastCallGas` regression covering state-diff recording with cold storage reads

Fixes #15381


figtracer (dismissed):

crates/cheatcodes/src/inspector.rs
```diff
@@ -2578,7 +2578,7 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
                 // Try to load the account and the slot's previous value, otherwise, assume it's
                 // not set (zero value)
                 let previous_value = if ecx.journal_mut().load_account(address).is_ok()
-                    && let Some(previous) = ecx.sload(address, key)
+                    && let Ok(previous) = ecx.sload_skip_cold_load(address, key, true)
```
mattsse: This breaks state-diff correctness for cold writes. On a cold `SSTORE` to a slot that already has a non-zero value, [`sload_skip_cold_load(address, key, true)` returns the skip-cold error and this branch records `previousValue = 0`](https://github.com/foundry-rs/foundry/blob/61441531c0bbbd92eefa9bd2f05f7950a491902f/crates/cheatcodes/src/inspector.rs#L2580-L2585), even though the real `SSTORE` then loads and overwrites the existing value. A repro is: set a target slot to `123`, call `vm.startStateDiffRecording()`, then set it to `456`; `stopAndReturnStateDiff()` reports `previousValue = 0` / `newValue = 456` instead of `previousValue = 123`.

Minimal correctness fix: keep the skip-cold behavior for `SLOAD`, but do not use it for `SSTORE` unless the recorder can get the pre-write value from the actual `SSTORE` execution/journal. The narrow fix is to leave this path on the previous `ecx.sload` behavior; a fuller fix would plumb the real `SStoreResult.present_value` or the post-op journal `storage_changed` previous value into the recorder so gas accounting and diff correctness both hold.

```suggestion
                    && let Some(previous) = ecx.sload(address, key)
```

Test coverage should include a cold-write state diff case, not just read gas accounting: initialize a slot before recording, write a new value while recording, then assert the recorded write has `previousValue == bytes32(uint256(123))` and `newValue == bytes32(uint256(456))`. I would add that next to the existing state-diff storage access tests, and run it together with `testStateDiffRecordingDoesNotWarmStorageReads`.

mattsse (commented):
Found one correctness regression in the state-diff recording path.

crates/cheatcodes/src/inspector.rs
mablr: Reverted the footgun and added a regression test https://github.com/foundry-rs/foundry/pull/15382/commits/ab852c681e66555e11eaa785ff2a8f06739b65a5

crates/cheatcodes/src/inspector.rs
```diff
@@ -2551,7 +2551,7 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
                 // it's not set (zero value)
                 // Try to load the account and the slot's present value
                 let present_value = if ecx.journal_mut().load_account(address).is_ok()
-                    && let Some(previous) = ecx.sload(address, key)
+                    && let Ok(previous) = ecx.sload_skip_cold_load(address, key, true)
```
stevencartavia: This makes cold reads of non-zero slots record `previousValue`/`newValue` as `0`. On a cold slot `sload_skip_cold_load(.., true)` returns `Err` before reading the value, so it falls back to `U256::ZERO`. `getStorageAccesses()`/`stopAndReturnStateDiff()` now return wrong values (master returned the real value).

Repros under `--isolate` and default mode when the slot hasn't already been warmed in the same test transaction (e.g. slot set in `setUp`, first read during recording). Could we read the present value without warming instead of returning 0? Also worth adding a cold-**read** test next to the new cold-write one.

testdata/paris/cheats/LastCallGas.t.sol
```diff
@@ -84,6 +100,20 @@ contract LastCallGasIsolatedTest is LastCallGasFixture {
         _performRefund();
         _assertGas(vm.lastCallGas(), Gas({gasTotalUsed: 21380, gasMemoryUsed: 0, gasRefunded: 4800}));
     }
+
+    function testStateDiffRecordingDoesNotWarmStorageReads() public {
+        StorageGasTarget recordingOff = new StorageGasTarget();
+        recordingOff.fill();
+        recordingOff.sum();
```
stevencartavia, line 235: Heads up that this only partially fixes the issue — recording still warms the slot/account before the opcode for cold `SSTORE` (~2100/slot) and cold `BALANCE`/`EXTCODE*` (~2500/account). Both are pre-existing (same as master), just flagging so it isn't considered a full fix. This new gas test only covers reads, so it won't catch the write case. (Also: the `common.rs` change looks unrelated — split it out?)

mablr: Agree👍, will do a follow-up

grandizzy (approved):
lgtm

figtracer (approved):

PR #15383, opened 2026-06-26
feat(debugger): show chain-aware precompile clues

figtracer:
Follow-up to #15379.

This makes debugger precompile call clues use the decoder's active precompile labels, so chain-specific precompiles such as Tempo FeeManager can be annotated without adding Tempo address tables to the debugger. It also decodes selected debug traces before building the debugger dump so decoded call data is available in the notice.

### DIFF
--- a/crates/debugger/src/node.rs
+++ b/crates/debugger/src/node.rs
@@ -7,6 +7,8 @@ use revm_inspectors::tracing::types::{
 };
 use serde::{Deserialize, Serialize};
 
+const PRECOMPILES_TRACE_LABEL: &str = "PRECOMPILES";
+
 /// Represents a part of the execution frame before the next call or end of the execution.
 #[derive(Clone, Debug, Default, Serialize, Deserialize)]
 pub struct DebugNode {
@@ -187,8 +189,6 @@ const fn is_call_like_op(op: OpCode) -> bool {
             | OpCode::CREATE2
     )
 }
-
-const PRECOMPILES_TRACE_LABEL: &str = "PRECOMPILES";
 
 fn precompile_call_notice(
     trace: &CallTrace,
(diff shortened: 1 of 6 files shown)


grandizzy (commented):
Looks good, one finding pls check.

crates/debugger/src/node.rs
```diff
@@ -175,21 +190,33 @@ const fn is_call_like_op(op: OpCode) -> bool {
...
-fn precompile_call_notice(trace: &CallTrace) -> Option<String> {
-    decoded_precompile_call_notice(&trace.decoded).or_else(|| known_precompile_call_notice(trace))
+fn precompile_call_notice(
+    trace: &CallTrace,
+    precompile_labels: &AddressHashMap<String>,
+) -> Option<String> {
+    decoded_precompile_call_notice(&trace.decoded, trace.maybe_precompile.unwrap_or(false))
+        .or_else(|| known_precompile_call_notice(trace))
+        .or_else(|| labeled_precompile_call_notice(trace, precompile_labels))
+        .or_else(|| marked_precompile_call_notice(trace))
```
grandizzy: **Cleanup: remove the dead `maybe_precompile` fallback.** `trace.maybe_precompile` is always `None` in the debugger (`exclude_precompile_calls: false`), and a `Some(true)` precompile is `PushOnly` in revm-inspectors so it's never an attached child here anyway. Drop the `is_precompile` arg to `decoded_precompile_call_notice` (line 197) and the `.or_else(|| marked_precompile_call_notice(trace))` (line 200) + the function at 248. Suggest reordering to put the chain-aware decoder labels ahead of the hardcoded standard table:

```rust
decoded_precompile_call_notice(&trace.decoded)
    .or_else(|| labeled_precompile_call_notice(trace, precompile_labels))
    .or_else(|| known_precompile_call_notice(trace))
```

Then remove the two unit tests that build `maybe_precompile: Some(true)` children.

grandizzy (approved):
👍

mablr (approved):

PR #15385, opened 2026-06-26
fix(evm): preserve `CreatedLocal` across isolated `SELFDESTRUCT` flows

### LINKED ISSUE #15384: `selfdestruct` in `setUp` doesn't delete code
### Component

Forge

- [x] Foundry
- [x] Foundryup

`1.7.2-nightly`

`1.9.0`

`forge test`

### Operating System

Linux

### Describe the bug

The following test reverts now since Foundry `1.7.1`: https://github.com/pcaversaccio/tornado-cash-exploit/blob/main/test/MetamorphicContract.t.sol

```console
Ran 1 test for test/MetamorphicContract.t.sol:MetamorphicContract
[FAIL: assertion failed: 103 != 0] testMorphingContract() (gas: 7822)
Traces:
  [295103] MetamorphicContract::setUp()
    ├─ [107351] → new Factory@0x4f0574D5Fb21607426AE64606e3dB2FD680F2496
    │   └─ ← [Return] 536 bytes of code
    ├─ [74061] Factory::helloA()
    │   ├─ [20673] → new A@0x12e1ec092B930E495bD65b3017752837a7732e68
    │   │   └─ ← [Return] 103 bytes of code
    │   └─ ← [Return] A: [0x12e1ec092B930E495bD65b3017752837a7732e68]
    ├─ [26166] A::kill()
    │   └─ ← [SelfDestruct]
    ├─ [26188] Factory::kill()
    │   └─ ← [SelfDestruct]
    └─ ← [Stop]

  [7822] MetamorphicContract::testMorphingContract()
    ├─ [0] VM::assertEq(103, 0) [staticcall]
    │   └─ ← [Revert] assertion failed: 103 != 0
    └─ ← [Revert] assertion failed: 103 != 0
```

somehow the `selfdestruct` call in the `setUp` does _not_ delete the code anymore. It has been working until Foundry `1.7.0`.

mablr:
## Motivation

Close #15384

Preserves revm's per-transaction `CreatedLocal` marker across Foundry's isolated/nested execution paths without changing account warmth.

This fixes post-Cancun `SELFDESTRUCT` cleanup for contracts created and destroyed in the same logical test transaction, including when a `STATICCALL` happens between creation and destruction.

### DIFF
--- a/crates/evm/evm/src/inspectors/stack.rs
+++ b/crates/evm/evm/src/inspectors/stack.rs
@@ -1084,17 +1084,14 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>>
         interpreter: &mut Interpreter,
         ecx: &mut FoundryContextFor<'_, FEN>,
     ) {
-        if let Some(inner_context) = &self.inner_context_data {
-            let address = interpreter.input.target_address();
-            if inner_context.locally_created_accounts.contains(&address)
-                && let Some(account) = ecx.journal_mut().evm_state_mut().get_mut(&address)
-            {
-                account.mark_created_locally();
-            }
-        }
-        if self.locally_created_accounts.contains(&interpreter.input.target_address())
-            && let Some(account) =
-                ecx.journal_mut().evm_state_mut().get_mut(&interpreter.input.target_address())
+        let address = interpreter.input.target_address();
+        let should_mark_created_locally = self.locally_created_accounts.contains(&address)
+            || self
+                .inner_context_data
+                .as_ref()
+                .is_some_and(|ctx| ctx.locally_created_accounts.contains(&address));
+        if should_mark_created_locally
+            && let Some(account) = ecx.journal_mut().evm_state_mut().get_mut(&address)
         {
             account.mark_created_locally();
         }
@@ -1441,17 +1438,16 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>>
         call: &CreateInputs,
         outcome: &mut CreateOutcome,
     ) {
-        if self.in_inner_context
-            && outcome.result.result.is_ok()
-            && let Some(address) = outcome.address
-            && let Some(inner_context) = &mut self.inner_context_data
-        {
-            inner_context.locally_created_accounts.insert(address);
-        }
         if outcome.result.result.is_ok()
             && let Some(address) = outcome.address
         {
             self.locally_created_accounts.insert(address);
+
+            if self.in_inner_context
+                && let Some(inner_context) = &mut self.inner_context_data
+            {
+                inner_context.locally_created_accounts.insert(address);
+            }
         }
 
         // We are processing inner context outputs in the outer context, so need to avoid processing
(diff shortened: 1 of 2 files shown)


mattsse (dismissed):
LGTM

grandizzy (commented):
Nice approach on the gas decoupling — re-marking `CreatedLocal` in `initialize_interp` while keeping accounts cold is cleaner than perturbing access warmth, and I confirmed locally that `gas_report_with_fallback` passes unchanged (no 2500 delta). 👍

However, the fix is **incomplete**: it doesn't cover the depth-1 `STATICCALL` path, so EIP-6780 cleanup still regresses if an external `view`/read happens between creation and `selfdestruct` within the same logical `setUp`/test transaction.

**Root cause:** the `initialize_interp` re-marking only runs when `inner_context_data` is set, i.e. inside `transact_inner`. But depth-1 `STATICCALL`s aren't routed through `transact_inner` — they hit the separate cold-marking loop in the `call` hook ([`CallScheme::StaticCall` branch, stack.rs#L1336-L1337](https://github.com/foundry-rs/foundry/blob/e368eb2fb40c42e1d2df70726de2f58332dcc651/crates/evm/evm/src/inspectors/stack.rs#L1336-L1337)). That loop marks the locally-created account `Cold` and cold-loads it inline, so revm runs `unmark_created_locally()` on the **outer** journal, and no re-marking happens (we're not in inner context). By the time the next isolated `a.kill()` call snapshots `locally_created_accounts`, the account is already gone from the set.

**Verified against this exact commit** (`e368eb2`, forge built from source, `--evm-version cancun`):

| Test | Result |
|---|---|
| issue #15384 repro | ✅ PASS |
| `gas_report_with_fallback` | ✅ PASS (gas preserved) |
| STATICCALL variation (below) | ❌ **FAIL: `218 != 0`** (code not deleted) |

**Suggested fix** — mirror the guard on the `STATICCALL` branch:

```rust
// crates/evm/evm/src/inspectors/stack.rs, CallScheme::StaticCall branch
if warm_addresses.is_cold(addr) && !acc_mut.is_created_locally() {
    acc_mut.mark_cold();
}
```

Confirmed: with this one-line guard the STATICCALL regression test below passes and the gas report stays at 29758.

**Suggested regression test** to add to `crates/forge/tests/cli/test_cmd/mod.rs`:

```rust
forgetest_init!(setup_selfdestruct_after_staticcall_deletes_code, |prj, cmd| {
    prj.add_test(
        "SetupStaticcallSelfdestruct.t.sol",
        r#"

contract A {
    function ping() public view returns (uint256) {
        return 1;
    }

    function kill() public {
        selfdestruct(payable(address(0)));
    }
}

contract Factory {
    function helloA() public returns (address) {
        return address(new A());
    }
}

contract SetupStaticcallSelfdestructTest is Test {
    A private a;
    Factory private factory;

    function setUp() public {
        factory = new Factory();
        a = A(factory.helloA());

        // External STATICCALL to `a` between its creation and selfdestruct,
        // both within the same logical setUp transaction.
        a.ping();
        a.kill();
    }

    function testCodeDeletedDespiteStaticcall() public {
        assertEq(address(a).code.length, 0);
    }
}
   "#,
    );

    cmd.args([
        "test",
        "--match-path",
        "test/SetupStaticcallSelfdestruct.t.sol",
        "--evm-version",
        "cancun",
    ])
    .assert_success();
});
```

mablr (commented):
The suggested fix is a footgun, it avoids the clear by not marking locally-created accounts cold. But it re-couples EIP-6780 locality with access warmth.

This PR now keeps accounts cold for gas accounting and separately restores only the `CreatedLocal` semantic marker. IMO that’s the most consistent for gas-accounting.

grandizzy (approved):

PR #15387, opened 2026-06-26
fix(traces): guard Tempo precompile labels by hardfork

figtracer:
## Summary

- move hardfork-gated Tempo precompile labels out of the unconditional decoder label table
- add T3/T5/T6/T7 labels only through the Tempo hardfork-aware decoder builder
- cover both debugger precompile labels and generic decoded trace labels at the activation boundaries

### DIFF
--- a/crates/evm/traces/src/decoder/mod.rs
+++ b/crates/evm/traces/src/decoder/mod.rs
@@ -118,34 +118,7 @@ impl CallTraceDecoderBuilder {
     #[inline]
     pub fn with_tempo_hardfork(mut self, hardfork: Option<TempoHardfork>) -> Self {
         self.decoder.tempo_hardfork = hardfork;
-        if hardfork.is_some_and(|hardfork| hardfork.is_t3()) {
-            self.decoder
-                .labels
-                .entry(ADDRESS_REGISTRY_ADDRESS)
-                .or_insert_with(|| "AddressRegistry".to_string());
-            self.decoder
-                .labels
-                .entry(SIGNATURE_VERIFIER_ADDRESS)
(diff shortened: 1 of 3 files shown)


crates/evm/traces/src/decoder/mod.rs
```diff
@@ -147,6 +136,29 @@ impl CallTraceDecoderBuilder {
     }
 }
 
+fn insert_label(labels: &mut HashMap<Address, String>, address: Address, label: &'static str) {
+    labels.entry(address).or_insert_with(|| label.to_string());
+}
+
+fn insert_tempo_hardfork_labels(
```
stevencartavia, line 143: These hardfork-gated labels now only get added through `with_tempo_hardfork(...)`, but anvil's decoder is built without it, so this PR drops them in anvil traces.

`crates/anvil/src/config.rs` ~L1244 calls `CallTraceDecoderBuilder::new()` and never `.with_tempo_hardfork(...)`. Since these labels are no longer in the default table, `anvil --network tempo --hardfork tempo:t7 --print-traces` now shows raw precompile addresses instead of names (AddressRegistry, SignatureVerifier, TIP20ChannelReserve, ReceivePolicyGuard, StorageCredits).

forge/cast/script/chisel all pass the hardfork; anvil was missed. It has `self.networks.is_tempo()` and `self.get_hardfork()` available, e.g. `self.networks.is_tempo().then(|| TempoHardfork::from(self.get_hardfork()))`.

DaniPopes:
is this necessary? we dont do this even for mainnet which doesn't have all precompiles available always

figtracer:
yeah this def doesn't matter , closing it

PR #15392, opened 2026-06-26
refactor(traces): compose trace requirements

DaniPopes:
Split trace configuration into composed requirements instead of relying on TraceMode ordering. This lets callers request calls, step granularity, snapshots, state diffs, and decode needs independently, so new consumers can combine only the trace data they need without pretending the modes are a linear ladder.


DaniPopes:
Closing in favor of the clean base PR #15393. This branch picked up the previously merged evm-profile squash in its commit history after retargeting.

PR #15393, opened 2026-06-26
refactor(traces): remove trace mode

DaniPopes:
Replace the old TraceMode ordering with TraceRequirements as the tracing API. Callers now compose calls, step granularity, snapshots, state diffs, verbosity, and internal decode needs from TraceRequirements::none().with_* instead of selecting from a linear mode enum.

This keeps independent trace requirements independent: all-step recording can be requested without enabling debugger snapshots, internal function decoding can request only the jump data it needs, and high verbosity can add call or state-diff data without downgrading richer requirements.

### DIFF
--- a/crates/cast/src/cmd/call.rs
+++ b/crates/cast/src/cmd/call.rs
@@ -336,7 +336,7 @@ impl CallArgs {
                 }
             }
 
-            let trace_requirements = TraceRequirements::empty()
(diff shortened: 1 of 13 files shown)


figtracer (approved):

stevencartavia (approved):

mattsse (approved):
LGTM

PR #15394, opened 2026-06-26
feat(forge): visualize traces in speedscope with --evm-profile

DaniPopes:
Stacked on #15393.

Just like `--flamegraph`/`--flamechart`, but opens a [speedscope.app](https://www.speedscope.app) window which has a more friendly UI to navigate the profile.

More info available at [speedscope#usage](https://github.com/jlfwong/speedscope#usage).
In our case, the "CPU time" is gas; `--flamechart` maps to the "Time Order" tab (default), `--flamegraph` maps to "Left heavy".

These screenshots are from a temporary deep-call fixture generated for visual verification (`DeepProfileTest::testDeepProfile`: 56 frames, max open stack depth 14). The Time Order screenshot is from the trace-order profile with 2,904 events.

### Speedscope: Time Order

Speedscope Time Order view for a deep EVM trace

### Speedscope: Left Heavy

Speedscope Left Heavy view for a deep EVM trace

### Speedscope: Sandwich

Speedscope Sandwich view for a deep EVM trace

### Existing `--flamegraph` comparison

Existing Forge flamegraph for the same deep EVM trace


crates/forge/src/cmd/test/mod.rs
```diff
@@ -1090,6 +1129,55 @@ impl TestArgs {
             }
         }
 
+        if profile_format.is_some() {
+            let (profile_json, test_name_trimmed, contract) = {
+                let decoder = outcome.last_run_decoder.clone().unwrap();
```
stevencartavia: This unwraps `None` whenever the run returned before a decoder was set: `--list`, the plain `--json` path, `--junit`, and also a zero-match run (`--match-test` matching nothing). `--evm-profile`'s `conflicts_with_all` is currently only `["flamegraph", "flamechart"]`, so clap allows `--json`/`--junit`/`--list` with it. Add `json`, `junit`, `list` to that conflict list, and replace the unwrap with `ok_or_eyre(...)` so the zero-match (and any future early-return) path can't panic either.

crates/forge/src/cmd/test/mod.rs
```diff
@@ -1090,6 +1129,55 @@ impl TestArgs {
...
+                            (suite_name.as_str(), test_name.as_str(), result)
+                        })
+                    })
+                    .unwrap();
+
+                let (_, arena) = test_result
+                    .traces
+                    .iter_mut()
+                    .find(|(kind, _)| *kind == TraceKind::Execution)
+                    .unwrap();
```
stevencartavia: Panics when the selected result has no `TraceKind::Execution` — e.g. a reverting `setUp`/constructor returns a synthetic `setUp()`/`constructor()` result with only setup/deployment traces (reproduced), and removed `testFail*` functions become `TestResult::fail(...)` without executing. `num_filtered == 1` only guarantees one matching test function, not an execution trace. Replace the unwrap with `ok_or_eyre`. Note `contract`/`test_name_trimmed` are declared just below this `find`, so move them above it (or use the in-scope `suite_name`/`test_name`) to include them in the message, e.g. "cannot generate EVM profile for {contract}::{test}: no execution trace (test may have failed in setUp/constructor or been skipped)".

mattsse (dismissed):
LGTM

stevencartavia (dismissed):

mablr (approved):
Tested thoroughly locally, it works fine!

0xKarl98 (approved):

PR #15395, opened 2026-06-26
perf(symbolic): preserve byte structure

DaniPopes:
Preserves symbolic byte ranges as structured `SymBytes` instead of eagerly lowering everything to `Vec<SymExpr>`. This keeps aligned ABI calldata/memory words as whole `SymExpr` terms through copies, calls, returndata, and PUSH data, then normalizes the remaining packed-word fragments before SMT emission.

The solver path also canonicalizes normalized cache keys, shares repeated SMT subterms with `let`, and reuses cached branch complements: when `base && !X` is already known unsat, a later `base && X` feasibility check can be answered without another solver query.

For the largest Solady deltas, the old runtime was rebuilding symbolic words bytewise before doing ordinary word arithmetic. For example, `check_SaturatingAddEquivalence` emitted about 44k `bvlshr`, 43k `bvshl`, 46k `bvand`, and 44k `bvor` nodes across 8 SMT queries on `master`; this branch emits none of those byte-reassembly shifts for that test.

Benchmarks below used profiling `forge` binaries against Solady check tests with `forge test --symbolic --symbolic-dump-smt --match-path ... --match-test ... -vv --summary`.

`master`: `/tmp/forge-profiling-master-4b83ca7d7`, commit `4b83ca7d`.
This branch: `target/profiling/forge`, commit `dab70da92`.

Process wall time is mostly startup and artifact-cache checking for these single-test invocations, so the table focuses on Forge's reported suite/solver time and emitted SMT body size.

| Solady check | Status | master solver | branch solver | master SMT body | branch SMT body | SMT queries | sat cache |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `FixedPointMathLib.check_SaturatingAddEquivalence` | pass | 563ms | 39ms | 3,947,358 B | 2,038 B | 8 -> 6 | 0 -> 1 |
| `LibSort.check_HasDuplicateHashmapCapacityTrickEquivalence` | pass | 210ms | 84ms | 574,030 B | 4,876 B | 7 -> 4 | 0 -> 3 |
| `ERC20.check_IsNotUint256MaxTrickEquivalence` | pass | 66ms | 59ms | 2,440 B | 2,227 B | 10 -> 9 | 0 -> 1 |
| `ERC20.check_IsPermit2AndValueIsNotInfinityTrickEquivalence` | pass | 79ms | 79ms | 7,598 B | 6,542 B | 12 -> 12 | 1 -> 1 |
| `Timelock.check_DelayRestriction` | pass | 46ms | 46ms | 2,956 B | 2,834 B | 6 -> 6 | 3 -> 3 |
| `ERC721.check_AuxPackEquivalence` | pass | 39ms | 39ms | 3,564 B | 3,468 B | 6 -> 6 | 0 -> 0 |
| `SafeCastLib.check_SafeCastInt256ToIntTrickEquivalence` | pass | 26ms | 26ms | 958 B | 958 B | 2 -> 2 | 0 -> 0 |
| `SignatureCheckerLib.check_EcrecoverTrickEquivalence` | pass | 39ms | 39ms | 2,328 B | 2,440 B | 6 -> 6 | 0 -> 0 |

Some Solady checks still stop as incomplete because of hard-arithmetic fallback paths rather than SMT size. Those are useful controls, but not proof benchmarks:

| Solady check | master | branch | Notes |
| --- | ---: | ---: | --- |
| `FixedPointMathLib.check_SaturatingMulEquivalence` | 8 queries, 5 hard-arith | 7 queries, 4 hard-arith | incomplete on both, no SMT emitted |
| `Timelock.check_OperationStateDifferentialTrick` | 76 queries, 24 SMT | 73 queries, 21 SMT | incomplete on both |
| `Base58.check_CarryBoundsTrick` | 11 queries, 6 SMT | 11 queries, 6 SMT | incomplete on both |


crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -412,6 +416,15 @@ impl SmtLibSubprocessSolver {
...
             return Ok(false);
         }
+        if defer_hard_arith_without_witness
+            && let Some(complement_key) = branch_condition_complement_cache_key(constraints)
+            && self.has_cached_unsat_subset(&complement_key)
+        {
+            self.sat_cache_hits += 1;
+            trace!("is_sat: branch complement unsat cache hit");
+            self.cache_sat_result(cache_key, true);
+            return Ok(true);
```
stevencartavia, line 438: **Branch-complement SAT cache is unsound when the base path can be UNSAT**

In the `is_sat_branch` path, `is_sat_inner` shortcuts `base ∧ cond` to SAT once the complementary branch `base ∧ ¬cond` is cached UNSAT. That only holds if `base` is SAT — if `base` is UNSAT, both branches are UNSAT but the shortcut reports one feasible.

`vm.randomUint(min,max)` creates exactly that unchecked UNSAT base: it just pushes `value >= min` and `value <= max` with no SAT check, so `vm.randomUint(2, 1)` makes the path condition UNSAT.

Repro:
```solidity
function checkUnsatBaseBranchesPruned(uint256 a) public {
    vm.randomUint(2, 1);            // path condition becomes UNSAT
    if (a == 1) { vm.expectSafeMemory(0x80, 0xA0); }
    else        { vm.expectSafeMemory(0x80, 0xA0); }
}
```
- base `4b83ca7d`: `[PASS] ... sat: 2 (0 cached)`
- this PR: `[FAIL: incomplete ... vm.expectSafeMemory not modeled] ... sat: 2 (1 cached)`

It can't cause a false `PASS` or bogus counterexample (`model()` re-validates), but it explores infeasible paths and can turn a sound `PASS` into `INCOMPLETE`. Fix: only take the shortcut when `base` is known SAT.

DaniPopes: Fixed in dac175dc4. The branch-complement shortcut now only fires when the branch base has an exact cached SAT result, so an UNSAT base no longer lets `base && !cond` UNSAT imply `base && cond` SAT.

Added regression coverage for the UNSAT-base case and tightened the positive cache test to prove the base first.

Validation:
- `cargo test -p foundry-evm-symbolic sat_cache --lib`
- `cargo cl -p foundry-evm-symbolic`

crates/evm/symbolic/src/runtime/expressions.rs
```diff
@@ -121,7 +55,7 @@ pub(crate) fn create2_address_word(
             Ok((word, address))
         }
         (_, Err(SymbolicError::Unsupported("symbolic CREATE2 initcode"))) => {
-            let initcode_bytes = initcode.bytes().to_vec();
+            let initcode_bytes = initcode.bytes();
```
mattsse: This still keys symbolic CREATE2 addresses on `format!("{initcode_bytes:?}")`, where `initcode_bytes` is the structural `SymBytes`. That makes the address identity depend on the retained byte representation rather than the byte semantics: for example, `SymCode::from_bytes(init_word.clone().into_bytes())` and `SymCode::from_byte_exprs(init_word.into_byte_exprs())` represent the same 32-byte symbolic initcode, but currently produce different `create2_address_*` vars/representative addresses. That can make repeated symbolic CREATE2 flows fail to alias depending on how the initcode was assembled.

A minimal fix would be to canonicalize this fallback identity through materialized byte terms before calling `symbolic_create2_address_word`, for example:

```rust
let initcode_bytes = initcode.read_byte_exprs(0, initcode.len());
```

Then add a regression test that compares `create2_address_word` for structural-word initcode against the equivalent materialized-byte initcode and expects the same symbolic word/address.

DaniPopes: Canonicalized the symbolic CREATE2 fallback identity through materialized initcode byte expressions and added a regression comparing word-structured initcode against equivalent materialized-byte initcode.

mattsse (approved):
LGTM

mablr (approved):
🚀

PR #15396, opened 2026-06-27
feat: add symbolic benchmarks

DaniPopes:
Adds a focused `forge_symbolic_test` benchmark path that runs symbolic checks on `Vectorized/solady:v0.1.26` and `SorellaLabs/angstrom:73b55b8eca667b9a50fa4d8b6a7f45ec647420f5`, then aggregates symbolic solver counters from `forge test --symbolic --json`.

Solady covers small targeted checks from `FixedPointMathLib`, `LibSort`, `ERC20`, and `Timelock`. Angstrom adds a bounded hard case around `check_matchesSolady_fullMulX128`, which currently reaches an incomplete solver result under a 5s symbolic timeout but still emits useful solver and SMT counters. The benchmark runner now preserves parseable symbolic JSON from non-zero symbolic exits and reports pass, incomplete, and fail counts in the generated table.

The symbolic benchmark commands disable lint-on-build so Solidity lint diagnostics do not dominate solver benchmarks or make otherwise valid benchmark targets fail after compilation succeeds.

| Repo | Mean wall time | Symbolic status | Solver queries | SMT queries | SMT input | Max SMT query |
| --- | ---: | --- | ---: | ---: | ---: | ---: |
| `Vectorized/solady:v0.1.26` | 2.801s | 5 pass | 46 | 43 | 4.53 MiB | 493.5 KiB |
| `SorellaLabs/angstrom:73b55b8eca667b9a50fa4d8b6a7f45ec647420f5` | 6.611s | 1 incomplete | 11 | 2 | 5.6 KiB | 4.8 KiB |

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -34,6 +34,9 @@ env:
     uniswap/v4-core:46c6834698c48bc4a463a86d8420f4eb1d7f3b75 --nmc 'TickMathTestTest',
     sparkdotfi/spark-psm:v1.0.0 --nmc PSMInvariants_TimeBasedRateSetting_WithTransfers_WithPocketSetting
 
+  SYMBOLIC_REPOS: >-
+    Vectorized/solady:v0.1.26
+
   BUILD_REPOS: >-
     ithacaxyz/account:v0.5.7,
     vectorized/solady:v0.1.26,
@@ -116,6 +119,19 @@ jobs:
             --output-file forge_isolate_test_bench.md \
             --verbose
 
+      - name: Run forge symbolic benchmarks
+        env:
+          FOUNDRY_DIR: ${{ github.workspace }}/.foundry
+          VERSIONS: ${{ github.event.inputs.versions || env.DEFAULT_VERSIONS }}
+          REPOS: ${{ github.event.inputs.repos || env.SYMBOLIC_REPOS }}
+        run: |
+          ./target/release/foundry-bench --output-dir ./benches \
+            --versions "$VERSIONS" \
+            --repos "$REPOS" \
+            --benchmarks forge_symbolic_test \
+            --output-file forge_symbolic_bench.md \
+            --verbose
+
       - name: Run forge build benchmarks
         env:
           FOUNDRY_DIR: ${{ github.workspace }}/.foundry
@@ -155,6 +171,7 @@ jobs:
           path: |
             benches/forge_test_bench.md
             benches/forge_isolate_test_bench.md
+            benches/forge_symbolic_bench.md
             benches/forge_build_bench.md
             benches/forge_coverage_bench.md
             benches/LATEST.md
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -126,13 +126,8 @@ jobs:
           VERSIONS: ${{ github.event.inputs.versions || env.DEFAULT_VERSIONS }}
           REPOS: ${{ github.event.inputs.repos || env.SYMBOLIC_REPOS }}
         run: |
-          SYMBOLIC_VERSIONS=$(printf '%s\n' "$VERSIONS" | tr ',' '\n' | sed 's/^[[:space:]]*//;s/[[:space:]]*$//' | grep -E '^(local|nightly)$' | paste -sd, -)
-          if [ -z "$SYMBOLIC_VERSIONS" ]; then
-            echo "Skipping symbolic benchmarks: no symbolic-capable versions in '$VERSIONS'"
-            exit 0
-          fi
           ./target/release/foundry-bench --output-dir ./benches \
-            --versions "$SYMBOLIC_VERSIONS" \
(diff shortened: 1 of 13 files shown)


mattsse (dismissed):
LGTM

crates/forge/src/result.rs
```diff
@@ -1231,6 +1235,18 @@ pub struct SymbolicSolverStats {
...
+    pub smt_input_bytes: u64,
+    /// Largest single SMT-LIB query input sent to a backend solver subprocess, in bytes.
+    #[serde(default)]
+    pub smt_max_query_bytes: u64,
+    /// Wall-clock time spent building SMT-LIB query strings, in milliseconds.
+    #[serde(default)]
+    pub smt_build_time_ms: u64,
+    /// Longest single backend solver subprocess query, in milliseconds.
+    #[serde(default)]
+    pub smt_max_query_time_ms: u64,
```
stevencartavia, line 1275: The 4 new stats fields (`smt_input_bytes`, `smt_max_query_bytes`, `smt_build_time_ms`, `smt_max_query_time_ms`) are serialized in each per-test `symbolic.solver.stats` object, but `crates/evm/symbolic/assets/symbolic-result.schema.json` wasn't updated. Its `solver_stats` def has `additionalProperties: false` and `schema_version` is still 1, so current `--symbolic --json` output no longer validates against the shipped v1 schema. The counterexample artifact schema refs the same `solver` def (which uses `solver_stats`) and artifacts embed `solver.stats`, so they're affected too.

.github/workflows/benchmarks.yml
```diff
@@ -116,6 +119,19 @@ jobs:
...
+          FOUNDRY_DIR: ${{ github.workspace }}/.foundry
+          VERSIONS: ${{ github.event.inputs.versions || env.DEFAULT_VERSIONS }}
+          REPOS: ${{ github.event.inputs.repos || env.SYMBOLIC_REPOS }}
+        run: |
+          ./target/release/foundry-bench --output-dir ./benches \
+            --versions "$VERSIONS" \
+            --repos "$REPOS" \
+            --benchmarks forge_symbolic_test \
+            --output-file forge_symbolic_bench.md \
+            --verbose
```
stevencartavia, line 135: `forge_symbolic_test` runs `forge test --symbolic --json`, which defaults to the `z3` solver, but the apt install in both `benchmarks.yml` and `benchmarks-nightly.yml` only adds `build-essential pkg-config` (hyperfine is installed separately), and z3 isn't on the Ubuntu 24.04 runner image. Without z3, `forge test --symbolic` exits non-zero and the benchmark bails — `benchmarks.yml`'s symbolic step has no `continue-on-error` so the job fails, while nightly continues and the merged JSON just omits symbolic data. Should z3 be added to the apt install in both workflows?

.github/workflows/benchmarks.yml
```diff
@@ -116,6 +119,19 @@ jobs:
             --output-file forge_isolate_test_bench.md \
             --verbose
 
+      - name: Run forge symbolic benchmarks
+        env:
+          FOUNDRY_DIR: ${{ github.workspace }}/.foundry
+          VERSIONS: ${{ github.event.inputs.versions || env.DEFAULT_VERSIONS }}
```
stevencartavia, line 127: The symbolic step in `benchmarks.yml` uses `--versions "$VERSIONS"`, defaulting to `stable,local`, but `stable` (currently v1.7.1) has no `forge test --symbolic` — symbolic landed after that release. The `stable` pass fails with an unknown flag and the benchmark bails; with no `continue-on-error` on the step, the job fails even once z3 is installed. Should the symbolic benchmark target only `local`/`nightly` until a stable release supports `--symbolic`?

.github/workflows/benchmarks.yml
DaniPopes: Added z3 to the apt install in both benchmark workflows.

.github/workflows/benchmarks.yml
DaniPopes: Updated the manual workflow to filter symbolic benchmark versions to local/nightly only, and removed the stable symbolic step from nightly.

crates/forge/src/result.rs
DaniPopes: Added the SMT stats fields to the symbolic result schema and a regression test that checks the schema solver stats keys.

.github/workflows/benchmarks.yml
DaniPopes: Follow-up: moved this into foundry-bench instead of filtering by hard-coded version names. The runner now checks the active `forge test --help` output after switching versions and skips only `forge_symbolic_test` for versions that do not expose `--symbolic`. Restored the stable symbolic workflow step.

mattsse (dismissed):

stevencartavia (dismissed):

figtracer (approved):

0xKarl98 (approved):

PR #15404, opened 2026-06-29
fix(cheatcodes): avoid warming recorded cold accesses

mablr:
## Motivation

Fixes the remaining state-diff recording follow-ups from #15382. Recording now observes cold `SSTORE` slot values and cold `BALANCE`/`EXTCODE*` account metadata through reverted journal checkpoints, so the recorder can keep the same diff metadata without warming those slots or accounts before the real opcode executes.

This also keeps `getStateDiff` from reporting nonce diffs for reverted account accesses, matching the existing reverted-call state-diff behavior, and adds isolated `lastCallGas` regressions for cold storage writes plus `BALANCE`, `EXTCODESIZE`, `EXTCODEHASH`, and `EXTCODECOPY`.


figtracer (approved):

0xKarl98 (approved):

mattsse (approved):
lgtm

PR #15406, opened 2026-06-29
fix(forge): support EIP-2935 history storage

### LINKED ISSUE #15402: forge test: EIP-2935 history storage contract not deployed (BLOCKHASH-from-state unsupported in unit tests)
### Component

Forge (`forge test`)

When the active EVM spec is Prague or later, `forge test` should deploy the EIP-2935 block-hash history storage contract at `HISTORY_STORAGE_ADDRESS` (`0x0000F90827F1C53a10cb7A02335B175320002935`) and populate it via the pre-block `SYSTEM_ADDRESS` system call, the same way real clients (and now Anvil) do.

Today, contracts that read historical block hashes **from state** (via a `staticcall` to the EIP-2935 contract, as opposed to the `BLOCKHASH` opcode) cannot be tested under `forge test` without manually etching a stand-in — even though the same code works on mainnet and under Anvil.

### Background / why this isn't covered already

#10357 requested EIP-2935 support (filed under the **Forge** component). It was closed by #13588, but that PR is **Anvil-only** (`feat(anvil): add EIP-2935 blockhash history storage support`):

- Anvil now deploys `HISTORY_STORAGE_CODE` at genesis when Prague is active and injects the per-block system call. ✅
- The `forge test` in-memory backend (`crates/evm/core`) was not changed. ❌

A grep of `crates/evm`, `crates/forge`, and `crates/cheatcodes` shows no references to `HISTORY_STORAGE_ADDRESS` / `eip2935::` — only `crates/anvil` deploys it. So the Forge half of #10357 is still open in practice.

Note that `vm.setBlockhash` (#7929) does **not** address this: it only sets the `BLOCKHASH` opcode cache (`database.set_blockhash`), not the EIP-2935 contract's code or storage. Code that reads via the contract still sees an empty account.

### Steps to reproduce

```solidity
pragma solidity ^0.8.13;

contract EIP2935Test is Test {
    address constant HISTORY = 0x0000F90827F1C53a10cb7A02335B175320002935;

    function test_historyContractDeployed() public {
        vm.roll(1000);
        // On a real Prague chain / Anvil, this account has code.
        assertGt(HISTORY.code.length, 0, "EIP-2935 history contract should be deployed");
    }

    function test_setBlockhashDoesNotPopulateContract() public {
        vm.roll(1000);
        vm.setBlockhash(900, keccak256("x"));

        // Opcode path works:
        assertEq(blockhash(900), keccak256("x"));

        // But the contract path (what real consumers use) returns empty:
        (bool ok, bytes memory ret) = HISTORY.staticcall(abi.encode(uint256(900)));
        assertTrue(ok && ret.length >= 32 && bytes32(ret) != bytes32(0),
            "EIP-2935 contract returns nothing under forge test");
    }
}
```

Run with a Prague (or later) hardfork:

```bash
forge test --evm-version prague -vv
```

Both assertions fail today.

### Expected behavior

With Prague active, `forge test` should:

1. Set `HISTORY_STORAGE_CODE` at `HISTORY_STORAGE_ADDRESS` (mirroring `crates/anvil/.../mem/mod.rs`, which already does this behind `spec_id() >= SpecId::PRAGUE`).
2. Perform the pre-block `SYSTEM_ADDRESS → HISTORY_STORAGE_ADDRESS` system call so that, as the block number advances (e.g. via `vm.roll`), parent hashes are stored at slot `(number - 1) % 8191` and become readable through the contract.

This would make EIP-2935-from-state behavior consistent across `forge test`, Anvil, and mainnet, removing the need for hand-rolled mocks/etches.

### Additional context

- Real-world consumers: L2/rollup bridges that read anchor block hashes from the EIP-2935 contract (the 8192-block serve window) rather than the 256-block `BLOCKHASH` opcode window.
- Anvil reference implementation: `crates/anvil/src/eth/backend/mem/mod.rs` (`Deploy EIP-2935 blockhash history storage contract if Prague is active`).
- Canonical bytecode/address: [EIP-2935 spec](https://eips.ethereum.org/EIPS/eip-2935).

```
forge Version: 1.7.1-dev
Commit SHA: 0d8cc990864673b71b3e239669d051f1984d6bdf
```

mablr:
## Motivation

Fix #15402.

This adds EIP-2935 history storage support to Forge’s in-memory test executor for Prague and later hardforks. Forge now deploys the canonical history storage contract in non-fork test execution, and the EIP-2935 ring buffer is kept in sync when tests advance block state or set blockhashes through cheatcodes.

The new CLI coverage checks that Prague tests can read hashes through the history storage contract, that cached blockhash updates are reflected in the contract where appropriate, and that the contract remains absent before Prague.

### DIFF
--- a/crates/cheatcodes/src/evm.rs
+++ b/crates/cheatcodes/src/evm.rs
@@ -5,7 +5,6 @@ use crate::{
     Vm::*, inspector::RecordDebugStepInfo,
 };
 use alloy_consensus::transaction::SignerRecoverable;
-use alloy_eips::eip2935::{HISTORY_SERVE_WINDOW, HISTORY_STORAGE_ADDRESS, HISTORY_STORAGE_CODE};
 use alloy_evm::FromRecoveredTx;
 use alloy_genesis::{Genesis, GenesisAccount};
 use alloy_network::eip2718::EIP4844_TX_TYPE_ID;
@@ -28,6 +27,10 @@ use foundry_evm_core::{
     FoundryBlock, FoundryTransaction,
     backend::{DatabaseError, DatabaseExt, RevertStateSnapshotAction},
     constants::{CALLER, CHEATCODE_ADDRESS, HARDHAT_CONSOLE_ADDRESS, TEST_CONTRACT_ADDRESS},
+    eip2935::{
+        HISTORY_SERVE_WINDOW, HISTORY_STORAGE_ADDRESS, HISTORY_STORAGE_CODE, forward_fill_start,
+        history_storage_slot, history_storage_value,
+    },
     env::FoundryContextExt,
     evm::{FoundryEvmNetwork, TxEnvFor, TxEnvelopeFor},
     utils::get_blob_base_fee_update_fraction_by_spec_id,
@@ -610,14 +613,14 @@ impl Cheatcode for rollCall {
     fn apply_stateful<FEN: FoundryEvmNetwork>(&self, ccx: &mut CheatsCtxt<'_, '_, FEN>) -> Result {
         let Self { newHeight } = self;
         let current_height = ccx.ecx.block().number();
-        if (*ccx.ecx.cfg().spec()).into() >= SpecId::PRAGUE
-            && *newHeight > current_height
-            && !newHeight.is_zero()
-        {
-            let parent_number = *newHeight - U256::from(1);
-            let parent_hash =
-                ccx.ecx.db_mut().block_hash(parent_number.saturating_to()).unwrap_or_default();
-            set_eip2935_blockhash(ccx.ecx, parent_number, parent_hash)?;
+        if (*ccx.ecx.cfg().spec()).into() >= SpecId::PRAGUE && *newHeight > current_height {
+            let mut block_number = forward_fill_start(current_height, *newHeight);
+            while block_number < *newHeight {
+                let block_hash =
+                    ccx.ecx.db_mut().block_hash(block_number.saturating_to()).unwrap_or_default();
+                set_eip2935_blockhash(ccx.ecx, block_number, block_hash)?;
+                block_number += U256::from(1);
+            }
         }
         ccx.ecx.block_mut().set_number(*newHeight);
         Ok(Default::default())
@@ -1194,9 +1197,10 @@ impl Cheatcode for setBlockhashCall {
         );
 
         ccx.ecx.db_mut().set_blockhash(blockNumber, blockHash);
+        let current_block = U256::from(ccx.ecx.block().number());
         if (*ccx.ecx.cfg().spec()).into() >= SpecId::PRAGUE
-            && blockNumber + U256::from(HISTORY_SERVE_WINDOW)
-                >= U256::from(ccx.ecx.block().number())
+            && blockNumber < current_block
+            && current_block - blockNumber <= U256::from(HISTORY_SERVE_WINDOW)
         {
             set_eip2935_blockhash(ccx.ecx, blockNumber, blockHash)?;
         }
@@ -1771,8 +1775,8 @@ fn set_eip2935_blockhash<
     ecx.journal_mut()
         .sstore(
             HISTORY_STORAGE_ADDRESS,
-            block_number % U256::from(HISTORY_SERVE_WINDOW),
-            U256::from_be_bytes(block_hash.0),
+            history_storage_slot(block_number),
+            history_storage_value(block_hash),
         )
         .map_err(|e| fmt_err!("failed to store EIP-2935 history slot: {:?}", e))?;
     Ok(())
--- a/crates/cheatcodes/src/evm.rs
+++ b/crates/cheatcodes/src/evm.rs
@@ -5,6 +5,7 @@ use crate::{
     Vm::*, inspector::RecordDebugStepInfo,
 };
 use alloy_consensus::transaction::SignerRecoverable;
+use alloy_eips::eip2935::{HISTORY_SERVE_WINDOW, HISTORY_STORAGE_ADDRESS};
 use alloy_evm::FromRecoveredTx;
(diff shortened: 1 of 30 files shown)


grandizzy (commented):
Minor nit (non-blocking) — nice work on this. I checked the implementation against the [EIP-2935 spec](https://eips.ethereum.org/EIPS/eip-2935) and built the branch to test it. Core slot math is correct: `history_storage_slot` uses `alloy_eips::eip2935::HISTORY_SERVE_WINDOW` (= 8191, matching the canonical bytecode's `0x1fff`), so Forge writes exactly the slot the deployed contract reads. Write semantics, the `vm.roll` forward-fill range, and the `vm.setBlockhash` window all match the spec, and reads return the right hashes.

One small correctness nit worth considering: the cheatcode path warms the history account/storage in the user transaction, which the real pre-block system call must not do (details + reproduction + suggested fix inline).

crates/cheatcodes/src/evm.rs
```diff
@@ -1737,6 +1758,30 @@ pub(super) fn ensure_loaded_account<CTX: ContextTr<Db: Database<Error = Database
     Ok(())
 }
 
+fn set_eip2935_blockhash<
```
grandizzy, line 1761: The history update runs on the live user-tx journal (`ensure_loaded_account` + `journal_mut().sstore`), so it warms `HISTORY_STORAGE_ADDRESS` and the written slot for the rest of the test transaction. EIP-2935 specifies the pre-block update "will not warm the `HISTORY_STORAGE_ADDRESS` account or its storage slots as per EIP-2929 rules," so a test's first access should be cold.

**Repro:**
```solidity
pragma solidity ^0.8.13;

import {Test, console} from "forge-std/Test.sol";

contract Dummy { uint256 public x = 7; }

contract WarmthTest is Test {
    address constant HISTORY = 0x0000F90827F1C53a10cb7A02335B175320002935;

    // First EXTCODEHASH of `a` in this tx. EIP-2929: cold = 2600, warm = 100.
    // No subcall, so no 63/64 gas-forwarding noise -- a clean warmth probe.
    function _firstAccess(address a) internal view returns (uint256 used) {
        assembly {
            let g := gas()
            let h := extcodehash(a)
            used := sub(g, gas())
        }
    }

    function testAccountColdControl() public {
        Dummy d = new Dummy();
        console.log("cold unknown addr access =", _firstAccess(address(0xCAFE)));
        console.log("warm (just created) addr =", _firstAccess(address(d)));
    }

    function testHistoryAccountColdBaseline() public view {
        console.log("HISTORY first access (no roll) =", _firstAccess(HISTORY));
    }

    function testHistoryAccountWarmthAfterRoll() public {
        vm.roll(1000);
        console.log("HISTORY first access after roll =", _firstAccess(HISTORY));
    }

    function testHistoryAccountWarmthAfterSetBlockhash() public {
        vm.roll(2000);
        vm.setBlockhash(1999, keccak256("h"));
        console.log("HISTORY first access after setBlockhash =", _firstAccess(HISTORY));
    }
}
```

`forge test --evm-version prague --block-number 100 -vv`:
```
[PASS] testAccountColdControl()
Logs:
  cold unknown addr access = 2605
  warm (just created) addr = 105

[PASS] testHistoryAccountColdBaseline()
Logs:
  HISTORY first access (no roll) = 2605

[PASS] testHistoryAccountWarmthAfterRoll()
Logs:
  HISTORY first access after roll = 105

[PASS] testHistoryAccountWarmthAfterSetBlockhash()
Logs:
  HISTORY first access after setBlockhash = 105
```

`2605` = cold access, `105` = warm. Without a cheatcode the account is correctly cold (the executor seeds it via the DB at construction), but `vm.roll`/`vm.setBlockhash` make it warm — gas-sensitive tests around the EIP-2935 contract will see incorrect access costs.

**Suggested fix** — capture the prior cold state and restore it after the write (uses the same APIs already in this file: `load_account(...).is_cold`, `Account::mark_cold()`, `EvmStorageSlot.is_cold`):
```rust
fn set_eip2935_blockhash<
    CTX: ContextTr<Db: Database<Error = DatabaseError>, Journal: JournalExt>,
>(
    ecx: &mut CTX,
    block_number: U256,
    block_hash: B256,
) -> Result<()> {
    // EIP-2935: the pre-block system update must NOT warm the history account
    // or its storage slots (EIP-2929). Restore prior access state after writing.
    let account_was_cold = ecx.journal_mut().load_account(HISTORY_STORAGE_ADDRESS)?.is_cold;

    let account =
        ecx.journal_mut().evm_state().get(&HISTORY_STORAGE_ADDRESS).expect("account is loaded");
    if account.info.code_hash != keccak256(&HISTORY_STORAGE_CODE) {
        if account_was_cold
            && let Some(acc) = ecx.journal_mut().evm_state_mut().get_mut(&HISTORY_STORAGE_ADDRESS)
        {
            acc.mark_cold();
        }
        return Ok(());
    }

    let slot = history_storage_slot(block_number);
    let slot_was_cold = ecx
        .journal_mut()
        .sstore(HISTORY_STORAGE_ADDRESS, slot, history_storage_value(block_hash))
        .map_err(|e| fmt_err!("failed to store EIP-2935 history slot: {:?}", e))?
        .is_cold;

    if let Some(acc) = ecx.journal_mut().evm_state_mut().get_mut(&HISTORY_STORAGE_ADDRESS) {
        if account_was_cold {
            acc.mark_cold();
        }
        if slot_was_cold
            && let Some(slot) = acc.storage.get_mut(&slot)
        {
            slot.is_cold = true;
        }
    }
    Ok(())
}
```

mablr: ty https://github.com/foundry-rs/foundry/pull/15406/commits/edbcd14fa35027dc8bc0c3e3f114c4661e2bb9da

grandizzy (approved):
thank you, lgtm

mattsse (approved):
lgtm

PR #15407, opened 2026-06-29
feat(debugger): list tests when debugger needs one

figtracer:
## Motivation

Progresses #927.

When `forge test --debug` matches more than one test, Forge currently reports only the number of matches. That leaves users who run the debugger with broad filters, including no filters, without a direct list of candidates to narrow down.

## Solution

- List up to 12 matching tests in the debugger single-test error, including path, contract, and test name.
- Update the debugger narrowing hint to mention `--match-test` alongside `--match-contract` and `--match-path`.
- Leave the existing flamegraph, flamechart, and EVM profile error text unchanged.

## Testing

- `cargo +1.95.0 test -p forge --test cli requires_single_test -- --nocapture`
- `cargo +1.95.0 clippy -p forge --all-targets -- -D warnings`


mattsse (approved):
LGTM

mablr (approved):

grandizzy (approved):

PR #15408, opened 2026-06-29
feat(invariant): show live shrink sequences

### LINKED ISSUE #12268: feat(invariant): show real-time results while shrinking
### Component

Forge

Seeing shrinking happen in realtime has multiple advantage:
1. Seeing how the sequence is shrunk can help give you intuition of what is going on behind the scenes
2. Knowing when you can cancel shrinking
3. Spotting potential bugs that are missed when fully shrunk (more on that below)

Showing the number of transactions in the sequence while shrinking helps, but ideally you'll want to see the entire sequence in realtime as it is being shrunk. See below for a screenshot of what it looks like in Echidna.

### Why only seeing fully shrunk sequences can miss bugs
Let's use a (slightly oversimplified) example to illustrate this. We have a `deposit` function that can revert in two ways:
1. When depositing a larger amount, it overflows (the real bug)
2. When depositing 0, it reverts because zero deposits are not accepted

For our invariant we are testing "deposit does not revert".

When the fuzzer breaks this invariant with a large number, it will trigger bug 1. While shrinking, it will keep trying smaller amounts and eventually reach 0, which still breaks the invariant but for a different reason (bug 2). If we would only ever see the fully shrunk result, we would miss bug 1.

### Additional context

Image

figtracer:
Closes #12268.

## Summary
- Render the current invariant shrink sequence in the progress message while shrinking.
- Reuse `BaseCounterExample` formatting so live output gets identified contracts and Solidity-style rendering when configured.
- Bound large live sequences with a head/tail preview and skip formatting work when the progress bar is hidden.
- Enable steady ticks for per-test fuzz/invariant progress bars so long shrink runs keep repainting.

### DIFF
--- a/crates/evm/evm/src/executors/invariant/shrink.rs
+++ b/crates/evm/evm/src/executors/invariant/shrink.rs
@@ -10,15 +10,18 @@ use crate::executors::{
 use alloy_json_abi::Function;
 use alloy_primitives::{Address, B256, Bytes, I256, Selector, U256};
 use alloy_sol_types::SolCall;
+use foundry_common::ContractsByAddress;
 use foundry_config::InvariantConfig;
 use foundry_evm_core::{
     FoundryBlock, constants::MAGIC_ASSUME, decode::RevertDecoder, evm::FoundryEvmNetwork,
 };
-use foundry_evm_fuzz::{BasicTxDetails, invariant::InvariantContract};
+use foundry_evm_fuzz::{BaseCounterExample, BasicTxDetails, invariant::InvariantContract};
 use indicatif::ProgressBar;
 use proptest::bits::{BitSetLike, VarBitSet};
 use revm::context::Block;
-use std::{cell::Cell, collections::HashSet, hash::Hash};
+use std::{cell::Cell, collections::HashSet, fmt::Write, hash::Hash};
+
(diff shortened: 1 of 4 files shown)


mablr (dismissed):
Tested locally, seems to work fine.

crates/evm/evm/src/executors/invariant/shrink.rs
```diff
@@ -234,17 +237,126 @@ pub(crate) fn reset_shrink_progress(
...
+
+        let sequence = build_shrunk_sequence(calls, shrinker, accumulate_warp_roll);
+        let message = format_shrink_progress_message(
+            &self.message,
+            &sequence,
+            self.identified_contracts,
+            self.show_solidity,
+        );
         progress.set_message(message);
+        progress.force_draw();
```
stevencartavia: Could we just drop the `force_draw()` here? `set_message()` is already rate-limited and the bar has the 100ms steady tick, so doesn't the message update fine without it? This runs per shrink candidate and bypasses the rate limiter, so it ends up repainting the whole multi-line shrink message every candidate — when I removed just that line (same invariant + seed) repaints dropped from ~5200 to ~150 and stderr from ~33MB to ~1MB, same shrink result. Worth leaning on the normal rate-limited draw path instead?

mablr (approved):

grandizzy (approved):

PR #15410, opened 2026-06-29
fix(fuzz): validate showmap override names

mablr:
## Description

Part of OSS-297. #15227 follow-up.

This tightens showmap name validation so internal `ShowmapConfig` overrides go through the same single-path-component checks as the public `forge test --showmap-*` flags. That keeps callers such as `forge fuzz replay` from bypassing validation for `approach` and `trial` names.

The existing CLI path already rejected path components; this adds shared config validation and a focused unit test for the override path.


mattsse (approved):
LGTM

DaniPopes (approved):

figtracer (approved):

PR #15411, opened 2026-06-29
feat(cheatcodes): add `isIsolateMode`

### LINKED ISSUE #15409: cheatcode to identify `isolation` is active at runtime
### Component

Forge

We want to include a `gasUsed` check in a standard test for onchain proposals, currently `isolation` mode is the only true way to accurately get gas consumed. We think a runtime cheatcode to understand which context test is being run (similar to scripts we're able to identify which broadcast mode is active), would help asserting the test is run isolation mode. 
Current workaround: https://github.com/aave-dao/aave-helpers/pull/743

Two interface ideas:
- getTestMode() returns Enum { Isolation, None }
- startIsolation(), stopIsolation()

### Additional context

_No response_

mablr:
## Motivation

Close #15409.

Add `isIsolateMode` cheatcode to identify if isolation is active at runtime.


mattsse (dismissed):
LGTM

crates/cheatcodes/src/evm.rs
```diff
@@ -824,6 +824,13 @@ impl Cheatcode for coolSlotCall {
     }
 }
 
+impl Cheatcode for isIsolateModeCall {
+    fn apply<FEN: FoundryEvmNetwork>(&self, state: &mut Cheatcodes<FEN>) -> Result {
+        let Self {} = self;
+        Ok(state.config.isolate.abi_encode())
```
stevencartavia, line 854: `isIsolateMode()` reads `config.isolate`, but that isn't always the effective isolation state. `forge test --gas-report` forces isolation on (`evm_opts.isolate = true`) without touching `config.isolate`, so the cheatcode returns `false` while isolation is actually active.

Repro with `isolate = false` in `foundry.toml`:

```
plain:        not isolated, isIsolateMode=false  ✓
--isolate:    isolated,     isIsolateMode=true   ✓
--gas-report: isolated,     isIsolateMode=false  ✗
```

That `--gas-report` is genuinely isolated is confirmed three independent ways: gas matches `--isolate`, EIP-1153 transient storage is cleared between top-level calls, and the caller's nonce increments per top-level call (each runs as its own tx via `transact_inner`).

This breaks the #15409 use case: `require(vm.isIsolateMode())` fails under `--gas-report` even though the run is isolated. Could it read the effective isolation (`TestRunnerConfig.isolation`) instead of `config.isolate`?

mablr: Good catch, thanks.

Addressed here https://github.com/foundry-rs/foundry/pull/15411/commits/d0440fdd35c69fbbab7452a76afd0542f0cd5a01

mattsse (dismissed):

stevencartavia (dismissed):

grandizzy (dismissed):
lgtm

figtracer (dismissed):

mablr (commented):
Pending forge-std release to fix CI

grandizzy (approved):

stevencartavia (approved):

PR #15412, opened 2026-06-29
test(forge): cover invariant corpus root scoping

mablr:
## Description

Part of OSS-297.

This adds regression coverage for generated invariant corpus root replay. `forge fuzz replay --corpus-dir <root>` should narrow a generated invariant corpus root to the selected contract before reading entries, so sibling contracts with matching handler selectors cannot contaminate replay.

The stateless fuzz case already had coverage; this adds the invariant counterpart.


mattsse (approved):
LGTM

stevencartavia (approved):

PR #15413, opened 2026-06-29
fix(evm): use system call for beacon roots

mablr:
## Motivation

This follows up on https://github.com/foundry-rs/foundry/pull/15371#pullrequestreview-4592704265 by routing the EIP-4788 beacon-root write through the EVM system-call path instead of the normal transaction path. The helper now executes the system call against an isolated backend and commits only the beacon roots contract state, avoiding caller or beneficiary side effects from the system address.

`cast run` now also treats a missing `parent_beacon_block_root` as a setup error once the resolved spec is Cancun or later, instead of silently skipping the beacon-root write and replaying with incomplete pre-block state.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5501,6 +5501,7 @@ version = "1.7.2"
 dependencies = [
  "alloy-dyn-abi",
  "alloy-eips",
+ "alloy-evm",
  "alloy-json-abi",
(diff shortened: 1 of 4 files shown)


mattsse (approved):
LGTM

figtracer (approved):

PR #15414, opened 2026-06-29
fix(forge): bound corpus replay failure output

mablr:
## Description

Part of OSS-297.

This bounds `forge fuzz replay --corpus-dir` failure reporting so pathological corpora cannot allocate and print one formatted error per failed entry. Replay now tracks the total number of failed corpus entries separately, reports the first 20 details, and appends a summary for the remaining failures.

Addresses https://github.com/foundry-rs/foundry/pull/15227#discussion_r3490912038.


stevencartavia (approved):

grandizzy (approved):
makes sense

PR #15415, opened 2026-06-29
ci(mpp): use a fresh wallet each run

stevencartavia:
Generate a fresh, faucet-funded wallet each run (`TEMPO_ROTATE_WALLET=1`). A new wallet has no channel history, so the open succeeds.

Also skips cleanly when `MPP_API_KEY` is unset.

Closes OSS-459


mattsse (approved):
LGTM

grandizzy (approved):
makes sense, ty!

PR #15419, opened 2026-06-29
feat(anvil): support eth_baseFee

mattsse:
Adds the missing Reth-compatible `eth_baseFee` RPC method to Anvil. The handler returns the next block base fee when EIP-1559 is active and `null` before London, matching Reth's contract, and wires request deserialization, dispatch, and endpoint coverage for London and pre-London nodes.

Authored with AI assistance.

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -821,6 +821,14 @@ impl<N: Network> EthApi<N> {
         Ok(U256::from(self.gas_price()))
     }
 
+    /// Returns the base fee for the next block, or null before London.
+    ///
+    /// Handler for ETH RPC call: `eth_baseFee`
+    pub fn base_fee(&self) -> Result<Option<U256>> {
+        node_info!("eth_baseFee");
+        Ok(self.backend.fees().is_eip1559().then(|| U256::from(self.backend.base_fee())))
+    }
+
     /// Returns the excess blob gas and current blob gas price
     pub fn excess_blob_gas_and_price(&self) -> Result<Option<BlobExcessGasAndPrice>> {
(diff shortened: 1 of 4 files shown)


crates/anvil/src/eth/api.rs
```diff
@@ -821,6 +821,14 @@ impl<N: Network> EthApi<N> {
         Ok(U256::from(self.gas_price()))
     }
 
+    /// Returns the base fee for the next block, or null before London.
+    ///
+    /// Handler for ETH RPC call: `eth_baseFee`
+    pub fn base_fee(&self) -> Result<Option<U256>> {
+        node_info!("eth_baseFee");
+        Ok(self.backend.fees().is_eip1559().then(|| U256::from(self.backend.base_fee())))
```
stevencartavia: `base_fee()` gates on `self.backend.fees().is_eip1559()`, but `FeeManager.spec_id` is fixed at construction (defaults to post-London) and isn't updated by fork hardfork auto-detection. So forking a pre-London chain without an explicit `--hardfork` returns a base fee instead of `null`.

Repro: fork mainnet at block 12_000_000 (pre-London) and call `eth_baseFee` → returns `0x3b9aca00` (1 gwei), should be `null`.

Fix: use `self.backend.is_eip1559()` (the live spec), like `anvil_setNextBlockBaseFeePerGas` does.

grandizzy: https://github.com/foundry-rs/foundry/pull/15419/changes/ac6564a19591c82a7bf518b46623ef734f3da183

stevencartavia (approved):

grandizzy (approved):

PR #15426, opened 2026-06-29
feat(anvil): support full pending filters

mattsse:
This adds support for the optional boolean parameter on `eth_newPendingTransactionFilter`. The default and `false` behavior continues to return pending transaction hashes, while `true` installs a full-pending-transaction filter backed by Anvil's existing full pending transaction stream.

The parser now accepts omitted, null, false, and true parameter forms, matching the boolean wire shape used by upstream clients.

Authored with AI assistance.

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -1886,8 +1886,8 @@ impl EthApi<FoundryNetwork> {
             EthRequest::EthNewFilter(filter) => self.new_filter(filter).await.to_rpc_result(),
             EthRequest::EthGetFilterChanges(id) => self.get_filter_changes(&id).await,
             EthRequest::EthNewBlockFilter(_) => self.new_block_filter().await.to_rpc_result(),
-            EthRequest::EthNewPendingTransactionFilter(_) => {
-                self.new_pending_transaction_filter().await.to_rpc_result()
+            EthRequest::EthNewPendingTransactionFilter(full) => {
+                self.new_pending_transaction_filter(full.unwrap_or(false)).await.to_rpc_result()
             }
             EthRequest::EthGetFilterLogs(id) => self.get_filter_logs(&id).await.to_rpc_result(),
             EthRequest::EthUninstallFilter(id) => self.uninstall_filter(&id).await.to_rpc_result(),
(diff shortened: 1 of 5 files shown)


stevencartavia (dismissed):

stevencartavia (dismissed):

crates/anvil/src/eth/api.rs
```diff
@@ -2820,9 +2787,13 @@ impl EthApi<FoundryNetwork> {
     /// Creates a filter in the node, to notify when new pending transactions arrive.
     ///
     /// Handler for ETH RPC call: `eth_newPendingTransactionFilter`
-    pub async fn new_pending_transaction_filter(&self) -> Result<String> {
+    pub async fn new_pending_transaction_filter(&self, full: bool) -> Result<String> {
         node_info!("eth_newPendingTransactionFilter");
-        let filter = EthFilter::PendingTransactions(self.new_ready_transactions());
+        let filter = if full {
+            EthFilter::FullPendingTransactions(self.full_pending_transactions())
```
grandizzy: This installs the unbounded subscription stream behind a poll-based filter, so it loses the backpressure the hash filter has and can buffer full tx objects without bound between polls (capped only by the 5-min filter eviction). Back it with a bounded channel instead:

```suggestion
            EthFilter::FullPendingTransactions(self.full_pending_transactions_filter())
```

Add this bounded helper next to `full_pending_transactions` (leaving the existing unbounded one for the WS subscription path, which is push-based and drained continuously):

```rust
/// Returns a bounded stream of full pending transactions for poll-based filters.
///
/// Unlike [`Self::full_pending_transactions`], this applies backpressure via a bounded channel,
/// so an unpolled filter cannot buffer transactions without bound.
pub fn full_pending_transactions_filter(
    &self,
) -> tokio::sync::mpsc::Receiver<AnyRpcTransaction> {
    // Mirror the pool's ready-listener buffer so a full filter is bounded like a hash filter.
    let (tx, rx) = tokio::sync::mpsc::channel(2048);
    let mut hashes = self.new_ready_transactions();
    let this = self.clone();
    tokio::spawn(async move {
        while let Some(hash) = hashes.next().await {
            if let Ok(Some(txn)) = this.transaction_by_hash(hash).await
                && tx.send(txn).await.is_err()
            {
                break;
            }
        }
    });
    rx
}
```

And change the filter variant in `crates/anvil/src/filter.rs` (the `poll_recv` arm stays as-is, since `poll_recv` works on a bounded receiver too):

```rust
// import: use tokio::sync::{Mutex, mpsc};
FullPendingTransactions(mpsc::Receiver<AnyRpcTransaction>),
```

The spawned task already terminates promptly when the filter is dropped/evicted: a full channel unblocks `send().await` with an error on receiver drop, so no extra cancellation handling is needed.

grandizzy (changes requested):
Parser, dispatch, and the full-tx object shape all look correct, and the four wire forms (omitted/null/false/true) are handled cleanly. One change requested before merge.

The full filter reuses `full_pending_transactions()`, which feeds an unbounded channel. That stream was built for WS subscriptions (drained continuously), but a polling filter only drains on `eth_getFilterChanges`, so a full filter can buffer transactions without bound between polls. The hash filter, by contrast, keeps the pool's bounded (2048) `Receiver<TxHash>` and self-limits.

Fix: back the full filter with a bounded channel so it matches the hash filter's bound and lets backpressure reach the pool listener (the pool drops on full via `try_send`, so this can never stall the pool). Inline suggestion + supporting changes below. Verified locally: existing test passes, plus a multi-tx delivery test.

grandizzy (approved):
lgtm!

mablr (approved):

PR #15428, opened 2026-06-29
feat(anvil): support receipt subscriptions

mattsse:
This adds `eth_subscribe` support for `transactionReceipts`. The subscription listens for new Anvil blocks, reuses the existing `eth_getBlockReceipts` receipt shape, and emits receipt vectors for blocks that contain matching transactions.

The optional `transactionHashes` filter is supported as well, so clients can subscribe to receipts for all transactions or only selected transaction hashes.

Authored with AI assistance.

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -53,6 +53,7 @@ use alloy_rpc_types::{
     anvil::{
         ForkedNetwork, Forking, Metadata, MineOptions, NodeEnvironment, NodeForkConfig, NodeInfo,
     },
+    pubsub::TransactionReceiptsParams,
     request::TransactionRequest,
     simulate::{SimulatePayload, SimulatedBlock},
     state::{AccountOverride, EvmOverrides, StateOverridesBuilder},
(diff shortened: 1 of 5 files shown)


crates/anvil/src/eth/api.rs
```diff
@@ -3489,6 +3490,42 @@ impl EthApi<FoundryNetwork> {
...
+                }
+
+                if receipts.is_empty() {
+                    continue;
+                }
+
+                if tx.send(receipts).is_err() {
+                    break;
+                }
+            }
```
grandizzy: The spawned task only terminates when `tx.send(...)` fails, but no-match/empty blocks `continue` before sending — so a subscription whose filter never matches keeps running after `eth_unsubscribe`, draining new-block notifications forever (and it won't notice unsubscribe while parked on `blocks.next().await` either). `select!`-ing on `tx.closed()` makes it exit as soon as the receiver is dropped. This also precomputes the hash filter and logs receipt-build errors:

```suggestion
            // Precompute the hash filter once.
            let hash_filter = filter
                .transaction_hashes
                .filter(|hashes| !hashes.is_empty())
                .map(|hashes| hashes.into_iter().collect::<std::collections::HashSet<_>>());

            loop {
                let block = tokio::select! {
                    biased;
                    // Terminate promptly once the subscriber unsubscribes / disconnects,
                    // even while parked waiting for the next block.
                    _ = tx.closed() => break,
                    maybe_block = blocks.next() => match maybe_block {
                        Some(block) => block,
                        None => break,
                    },
                };

                let receipts = match this.block_receipts(BlockId::Hash(block.hash.into())).await {
                    Ok(Some(mut receipts)) => {
                        if let Some(hashes) = &hash_filter {
                            receipts.retain(|receipt| hashes.contains(&receipt.transaction_hash()));
                        }
                        receipts
                    }
                    Ok(None) => continue,
                    Err(err) => {
                        trace!(target: "node", %err, "failed to build block receipts for subscription");
                        continue;
                    }
                };

                if receipts.is_empty() {
                    continue;
                }

                if tx.send(receipts).is_err() {
                    break;
                }
            }
```

grandizzy (changes requested):
Clean, well-scoped feature — the receipt shape (reuses `eth_getBlockReceipts`), the optional `transactionHashes` filter, and the `eth_subscribe` param dispatch all look correct.

One blocking lifecycle issue before merge (see inline): the spawned producer task is detached and only terminates when `tx.send(...)` fails. Because no-match/empty blocks `continue` before sending, a filtered subscription whose hashes never match never observes the dropped receiver after `eth_unsubscribe` and runs forever — waking on every block and calling `block_receipts` per block. While parked on `blocks.next().await` it also won't notice unsubscribe until the next block arrives. On a long-running node with churning subscriptions this leaks tasks and per-block work.

The inline suggestion uses `tokio::select!` on `tx.closed()` so the task exits promptly when the receiver is dropped on unsubscribe/disconnect (also folds in a precomputed `HashSet` filter and error logging). Dependency-free (`tokio-util` isn't in `anvil`).

Worth adding a regression test that subscribes with a never-matching filter, unsubscribes, and asserts the new-block listener is cleaned up.

crates/anvil/src/eth/api.rs
grandizzy: https://github.com/foundry-rs/foundry/pull/15428/changes/876e7d011d1aebb27815534fe44b44d9ce19c34f

grandizzy (dismissed):
@stevencartavia pls rereview

stevencartavia (dismissed):

grandizzy (approved):

stevencartavia (approved):

PR #15430, opened 2026-06-29
feat(anvil): support debug raw header

mattsse:
Adds Anvil support for geth/reth's `debug_getRawHeader` endpoint. The RPC request now accepts a block id, resolves the matching local block, and returns the RLP-encoded header bytes.

This fills another small compatibility gap for tooling that asks debug APIs for raw chain data instead of expanded JSON block objects. The integration coverage exercises lookup by both block number and block hash and decodes the returned bytes back into an Alloy header.

Authored with AI assistance.

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -1738,6 +1738,7 @@ impl EthApi<FoundryNetwork> {
             EthRequest::DebugGetRawTransactions(block) => {
                 self.raw_transactions(block).await.to_rpc_result()
             }
+            EthRequest::DebugGetRawHeader(block) => self.raw_header(block).await.to_rpc_result(),
             // non eth-standard rpc calls
             EthRequest::DebugClearTxpool(_) => self.debug_clear_txpool().await.to_rpc_result(),
             // non eth-standard rpc calls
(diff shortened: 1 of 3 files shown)


crates/anvil/src/eth/api.rs
```diff
@@ -2827,6 +2828,15 @@ impl EthApi<FoundryNetwork> {
         self.inner_raw_transaction(hash).await
     }
 
+    /// Returns RLP encoded raw block header.
+    ///
+    /// Handler for RPC call: `debug_getRawHeader`.
+    pub async fn raw_header(&self, block: BlockId) -> Result<Bytes> {
+        node_info!("debug_getRawHeader");
+        let block = self.backend.get_block(block).ok_or(BlockchainError::BlockNotFound)?;
```
stevencartavia, line 3103: On a forked node, `debug_getRawHeader` fails for any pre-fork block because `raw_header` uses the local-only `backend.get_block` and doesn't fall back to the fork, unlike `eth_getBlockByNumber`/`eth_getBlockByHash`/`debug_getRawTransaction` (and reth's own `debug_getRawHeader`).

Reproduced against a mainnet fork at block 14608400:

- `eth_getBlockByNumber(14608300)` → returns the block
- `eth_getBlockByHash(<hash>)` → returns the block
- `debug_getRawHeader` by number → `-32001 Resource not found`
- `debug_getRawHeader` by hash → `-32001 Resource not found`
- upstream reth `debug_getRawHeader` for the same block → returns the RLP header
- locally-mined block on the forked node → works

PR #15431, opened 2026-06-29
feat(anvil): support debug raw block

mattsse:
Adds Anvil support for geth/reth's `debug_getRawBlock` endpoint. The RPC request accepts a block id, resolves the local block, and returns the RLP-encoded block bytes.

This fills another raw debug API compatibility gap for tools that need canonical block bytes instead of expanded JSON block responses. The endpoint is covered by request parsing and an integration test that fetches by number and hash, then decodes the returned bytes back into Anvil's block type.

Authored with AI assistance.

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -1739,6 +1739,7 @@ impl EthApi<FoundryNetwork> {
                 self.raw_transactions(block).await.to_rpc_result()
             }
             EthRequest::DebugGetRawHeader(block) => self.raw_header(block).await.to_rpc_result(),
+            EthRequest::DebugGetRawBlock(block) => self.raw_block(block).await.to_rpc_result(),
             // non eth-standard rpc calls
             EthRequest::DebugClearTxpool(_) => self.debug_clear_txpool().await.to_rpc_result(),
             // non eth-standard rpc calls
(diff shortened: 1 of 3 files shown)


crates/anvil/src/eth/api.rs
```diff
@@ -2827,6 +2828,15 @@ impl EthApi<FoundryNetwork> {
         self.inner_raw_transaction(hash).await
     }
 
+    /// Returns RLP encoded raw block.
+    ///
+    /// Handler for RPC call: `debug_getRawBlock`.
+    pub async fn raw_block(&self, block: BlockId) -> Result<Bytes> {
+        node_info!("debug_getRawBlock");
+        let block = self.backend.get_block(block).ok_or(BlockchainError::BlockNotFound)?;
```
stevencartavia, line 3113: `raw_block` uses `backend.get_block`, which only looks in anvil's local block storage — no fork fallthrough, unlike `debug_getRawTransaction` and `eth_getBlockByNumber`/`eth_getBlockByHash`. In fork mode, before any local block is mined, fork/pre-fork blocks fail, including `latest`:

```
eth_getBlockByNumber(1)      -> Some(block)
debug_getRawBlock(1)         -> -32001 Resource not found
debug_getRawBlock("latest")  -> -32001 Resource not found
```

Is local-only intentional here, or should it fetch from the fork to stay consistent with the other block endpoints?

mattsse: I think this is fine, most fork endpoints dont have support for the debug_ endpoint anyway

PR #15433, opened 2026-06-29
feat(anvil): support debug raw receipts

mattsse:
Adds Anvil support for reth's `debug_getRawReceipts` endpoint. The RPC request accepts a block id, resolves the local block, and returns the block's receipts as EIP-2718 encoded bytes.

This fills another raw debug API compatibility gap for tooling that needs canonical receipt bytes by block. The integration coverage mines two transactions into one block, fetches receipts by number and hash, decodes the returned bytes back into Foundry receipt envelopes, and checks the unresolved-block error path.

Authored with AI assistance.

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -1667,6 +1667,9 @@ impl EthApi<FoundryNetwork> {
             EthRequest::DebugGetRawTransaction(hash) => {
                 self.raw_transaction(hash).await.to_rpc_result()
             }
+            EthRequest::DebugGetRawReceipts(block) => {
+                self.raw_receipts(block).await.to_rpc_result()
+            }
             // non eth-standard rpc calls
             EthRequest::DebugTraceTransaction(tx, opts) => {
                 self.debug_trace_transaction(tx, opts).await.to_rpc_result()
@@ -2825,6 +2828,19 @@ impl EthApi<FoundryNetwork> {
     pub async fn raw_transaction(&self, hash: B256) -> Result<Option<Bytes>> {
         node_info!("debug_getRawTransaction");
         self.inner_raw_transaction(hash).await
+    }
+
+    /// Returns EIP-2718 encoded raw receipts for the block.
+    ///
+    /// Handler for RPC call: `debug_getRawReceipts`.
+    pub async fn raw_receipts(&self, block: BlockId) -> Result<Vec<Bytes>> {
+        node_info!("debug_getRawReceipts");
+        let block = self.backend.get_block(block).ok_or(BlockchainError::BlockNotFound)?;
+        let receipts = self
+            .backend
+            .mined_receipts(block.header.hash_slow())
+            .ok_or(BlockchainError::BlockNotFound)?;
+        Ok(receipts.into_iter().map(|receipt| receipt.encoded_2718().into()).collect())
     }
 
     /// Returns EIP-2718 encoded raw transaction by block hash and index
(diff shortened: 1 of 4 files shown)


stevencartavia (dismissed):

crates/anvil/src/eth/api.rs
```diff
@@ -2827,6 +2830,19 @@ impl EthApi<FoundryNetwork> {
...
+    ///
+    /// Handler for RPC call: `debug_getRawReceipts`.
+    pub async fn raw_receipts(&self, block: BlockId) -> Result<Vec<Bytes>> {
+        node_info!("debug_getRawReceipts");
+        let block = self.backend.get_block(block).ok_or(BlockchainError::BlockNotFound)?;
+        let receipts = self
+            .backend
+            .mined_receipts(block.header.hash_slow())
+            .ok_or(BlockchainError::BlockNotFound)?;
+        Ok(receipts.into_iter().map(|receipt| receipt.encoded_2718().into()).collect())
```
grandizzy, line 2890: In fork mode, pre-fork blocks aren't in local storage, so this returns `BlockNotFound`. Fall back to the fork like `balance`/`get_*` do, reusing `eth_getBlockReceipts` (broad support, cached) and re-encoding the consensus envelope. `block_request` resolves both hash and number to a block number, so the by-hash path works too.

Note: bytes are re-encoded from the RPC receipt envelope, not a raw `debug_getRawReceipts` passthrough — byte-identical for standard receipts.

```suggestion
        node_info!("debug_getRawReceipts");

        // In fork mode, serve pre-fork blocks from the upstream provider.
        if let BlockRequest::Number(number) = self.block_request(Some(block)).await?
            && let Some(fork) = self.get_fork()
            && fork.predates_fork_inclusive(number)
        {
            let receipts = fork.block_receipts(number).await?.unwrap_or_default();
            return Ok(receipts
                .into_iter()
                .map(|receipt| {
                    receipt.0.inner.inner.map_logs(|log| log.inner).encoded_2718().into()
                })
                .collect());
        }

        let block = self.backend.get_block(block).ok_or(BlockchainError::BlockNotFound)?;
        let receipts = self
            .backend
            .mined_receipts(block.header.hash_slow())
            .ok_or(BlockchainError::BlockNotFound)?;
        Ok(receipts.into_iter().map(|receipt| receipt.encoded_2718().into()).collect())
```

grandizzy (changes requested):
Solid addition and the tests are good. One change requested before merge.

In fork mode this returns `BlockNotFound` for any pre-fork block, because `raw_receipts` reads only local storage (`get_block` + `mined_receipts`). The sibling `inner_raw_transaction` endpoint already falls back to the fork for pre-fork txs, and a real node would return historical receipts here, so fork tooling will be surprised.

The inline suggestion reuses the fork's `eth_getBlockReceipts` (cached, broadly supported) and re-encodes to canonical 2718 bytes, mirroring the existing `balance`/`get_*` pre-fork pattern. Verified locally with a fork test (passes with the fix, fails without).

crates/anvil/src/eth/api.rs
grandizzy: https://github.com/foundry-rs/foundry/pull/15433/changes/23e23c4d143e987848c5577a80a555185bc03655

grandizzy (approved):
@stevencartavia pls rereview and take https://github.com/foundry-rs/foundry/pull/15433/changes/23e23c4d143e987848c5577a80a555185bc03655 into account. thanks!

0xKarl98 (approved):

PR #15434, opened 2026-06-30
feat(anvil): support debug bad blocks

mattsse:
Adds Anvil support for reth's `debug_getBadBlocks` endpoint. Anvil does not maintain a bad-block store, so the endpoint returns an empty list for the local development node.

This fills a debug API compatibility gap for tooling that probes the bad-block endpoint before deciding which client features are available. The coverage includes request parsing and an HTTP-level integration test that asserts the endpoint exists and returns an empty array.

Authored with AI assistance.


stevencartavia (approved):

decofe:
Consolidated into #15476.

PR #15435, opened 2026-06-30
feat(anvil): support debug trace block

mattsse:
Adds Anvil support for `debug_traceBlock` by accepting an RLP-encoded block, decoding it, and tracing the matching local block through the existing hash-based debug tracer.

Anvil stores executed local blocks rather than exposing a raw block import/replay pipeline, so this supports encoded blocks that are already present in the local chain and otherwise preserves the existing block-not-found behavior. This fills the compatibility path for tooling that retrieves a block from Anvil and then traces it through the raw block endpoint.

Authored with AI assistance.


stevencartavia (dismissed):

stevencartavia (approved):

grandizzy (approved):
lgtm

PR #15436, opened 2026-06-30
feat(anvil): support trace replay transaction

mattsse:
Adds Anvil support for the Parity-style `trace_replayTransaction` endpoint. For local mined transactions, Anvil finds the containing block, reuses the existing block replay inspector path, and returns the requested `TraceResults` for the matching transaction.

Forked Anvil instances forward the request to the upstream provider, matching the existing trace replay block forwarding behavior. This gives trace tooling a single-transaction replay path without changing the broader block replay implementation.

Authored with AI assistance.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -1898,14 +1898,27 @@ impl<N: Network> Backend<N> {
         let block_number =
             self.blockchain.storage.read().transactions.get(&hash).map(|tx| tx.block_number);
 
-        if let Some(block_number) = block_number
-            && let Some(results) =
-                self.mined_parity_trace_replay_block_transactions(block_number, &trace_types)
-            && let Some(result) = results.into_iter().find(|result| result.transaction_hash == hash)
-        {
-            return Ok(result.full_trace);
+        // If the transaction was mined locally, replay it locally. Do not fall
+        // through to the fork when the local replay fails; that would misreport
+        // a local data problem as an upstream transaction lookup.
+        if let Some(block_number) = block_number {
+            let results = self
+                .mined_parity_trace_replay_block_transactions(block_number, &trace_types)
+                .ok_or(BlockchainError::BlockNotFound)?;
+
+            return results
+                .into_iter()
+                .find(|result| result.transaction_hash == hash)
+                .map(|result| result.full_trace)
+                .ok_or_else(|| {
+                    BlockchainError::Internal(format!(
+                        "replayed block {block_number} for local transaction {hash:?}, \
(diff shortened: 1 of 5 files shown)


stevencartavia (dismissed):

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -1806,6 +1808,30 @@ impl<N: Network> Backend<N> {
...
+            && let Some(result) = results.into_iter().find(|result| result.transaction_hash == hash)
+        {
+            return Ok(result.full_trace);
+        }
+
+        if let Some(fork) = self.get_fork() {
+            return Ok(fork.trace_replay_transaction(hash, trace_types).await?);
+        }
+
+        Err(BlockchainError::TransactionNotFound)
```
grandizzy: This chained `if let` collapses three cases (tx found / block replayable / tx in results) into one, so any local failure drops into the fork branch below. That means a tx that *is* mined locally but whose block can't be replayed (pruned/unavailable parent state, load-state inconsistency, race with revert) silently gets forwarded to the upstream provider, misreporting a local data problem as an upstream tx lookup. This diverges from `debug_trace_transaction` / `replay_tx_with_inspector`, which keep found-locally txs on the local path. Once the tx is found locally, commit to the local path:

```suggestion
        // If the transaction was mined locally, replay it locally. Do not fall
        // through to the fork when the local replay fails — that would misreport
        // a local data problem as an upstream transaction lookup.
        if let Some(block_number) = block_number {
            let results = self
                .mined_parity_trace_replay_block_transactions(block_number, &trace_types)
                .ok_or(BlockchainError::BlockNotFound)?;

            return results
                .into_iter()
                .find(|result| result.transaction_hash == hash)
                .map(|result| result.full_trace)
                .ok_or_else(|| {
                    BlockchainError::Internal(format!(
                        "replayed block {block_number} for local transaction {hash:?}, \
                         but its trace was missing"
                    ))
                });
        }

        // Not known locally: forward to the fork if present.
        if let Some(fork) = self.get_fork() {
            return Ok(fork.trace_replay_transaction(hash, trace_types).await?);
        }

        Err(BlockchainError::TransactionNotFound)
```

grandizzy (changes requested):
Looks good, one finding, please see inline.

grandizzy (dismissed):

mablr (dismissed):

grandizzy (dismissed):

grandizzy (dismissed):

mablr (dismissed):

grandizzy (approved):

mablr (approved):

emmajam (approved):

PR #15437, opened 2026-06-30
feat(anvil): support eth call many

mattsse:
Adds Anvil support for reth's `eth_callMany` endpoint. The implementation accepts call bundles, applies the optional state override once, runs calls sequentially with call-style EVM settings, and commits each call into the next call's state before returning `EthCallResponse` values.

This covers the main bundle execution path for local Anvil state and forwards forked historical requests to the upstream provider. Concrete `transactionIndex` state contexts are left for a separate parameter-specific follow-up because they require replaying partial block state.

Authored with AI assistance.


stevencartavia (approved):

grandizzy (approved):
lgtm

PR #15439, opened 2026-06-30
feat(anvil): support eth block access list

mattsse:
Adds Anvil support for the reth-compatible `eth_getBlockAccessList` endpoint. The endpoint resolves the requested block, forwards pre-fork requests to the upstream provider, and returns `null` for local Anvil blocks because Anvil does not currently store EIP-7928 block access list payloads.

This keeps the RPC surface compatible while leaving the by-hash, by-number, raw, and future stored-BAL behavior for separate scoped follow-ups.

Authored with AI assistance.

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -1553,6 +1553,9 @@ impl EthApi<FoundryNetwork> {
                     self.block_by_number(num).await.to_rpc_result()
                 }
             }
+            EthRequest::EthGetBlockAccessList(block_id) => {
+                self.block_access_list(block_id).await.to_rpc_result()
+            }
             EthRequest::EthGetTransactionCount(addr, block) => {
                 self.transaction_count(addr, block).await.to_rpc_result()
             }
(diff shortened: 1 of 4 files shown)


crates/anvil/src/eth/api.rs
```diff
@@ -2087,6 +2090,23 @@ impl EthApi<FoundryNetwork> {
...
+    /// Handler for ETH RPC call: `eth_getBlockAccessList`
+    pub async fn block_access_list(&self, block_id: BlockId) -> Result<Option<serde_json::Value>> {
+        node_info!("eth_getBlockAccessList");
+        let block_request = self.block_request(Some(block_id)).await?;
+        let BlockRequest::Number(number) = block_request else { return Ok(None) };
+
+        if let Some(fork) = self.get_fork()
+            && fork.predates_fork_inclusive(number)
+        {
+            return Ok(fork.block_access_list(block_id).await?);
```
stevencartavia, line 2104: The fork path forwards the original `block_id` instead of the resolved `number`:

```rust
return Ok(fork.block_access_list(block_id).await?);
```

This means relative tags (`latest`/`safe`/`finalized`) get re-resolved by the upstream provider against its current chain view instead of using the block Anvil already resolved locally. On a fresh fork (no blocks mined), `latest` resolves locally to the fork block N, but the literal `"latest"` is sent upstream, whose head has since advanced — so you get a different block's access list.

The sibling handlers (`get_proof`, `get_code`, `storage_at`) all forward the resolved `number`. Suggest doing the same here:

```rust
return Ok(fork.block_access_list(number.into()).await?);
```

decofe:
Consolidated into #15476.

PR #15440, opened 2026-06-30
feat(anvil): support block access list by hash

mattsse:
Adds Anvil support for the reth-compatible `eth_getBlockAccessListByBlockHash` endpoint. The endpoint resolves the supplied block hash, forwards pre-fork hashes to the upstream provider, and returns `null` for local Anvil blocks because Anvil does not currently store EIP-7928 block access list payloads.

This keeps the by-hash RPC surface compatible while leaving the by-number, raw, and future stored-BAL behavior for separate scoped follow-ups.

Authored with AI assistance.

### DIFF
--- a/crates/anvil/core/src/eth/mod.rs
+++ b/crates/anvil/core/src/eth/mod.rs
@@ -109,6 +109,9 @@ pub enum EthRequest {
         bool,
     ),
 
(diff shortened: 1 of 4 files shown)


stevencartavia (approved):

decofe:
Consolidated into #15476.

PR #15441, opened 2026-06-30
feat(anvil): support block access list by number

mattsse:
Adds Anvil support for the reth-compatible `eth_getBlockAccessListByBlockNumber` endpoint. The endpoint accepts the same lenient block number and tag forms as Anvil's other block-number RPCs, forwards pre-fork requests to the upstream provider, and returns `null` for local Anvil blocks because Anvil does not currently store EIP-7928 block access list payloads.

This keeps the by-number RPC surface compatible while leaving the raw and future stored-BAL behavior for separate scoped follow-ups.

Authored with AI assistance.

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -1553,6 +1553,9 @@ impl EthApi<FoundryNetwork> {
                     self.block_by_number(num).await.to_rpc_result()
                 }
             }
+            EthRequest::EthGetBlockAccessListByBlockNumber(block_number) => {
+                self.block_access_list_by_number(block_number).await.to_rpc_result()
+            }
             EthRequest::EthGetTransactionCount(addr, block) => {
                 self.transaction_count(addr, block).await.to_rpc_result()
             }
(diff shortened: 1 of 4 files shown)


crates/anvil/src/eth/api.rs
```diff
@@ -2087,6 +2090,26 @@ impl EthApi<FoundryNetwork> {
...
+        block_number: BlockNumber,
+    ) -> Result<Option<serde_json::Value>> {
+        node_info!("eth_getBlockAccessListByBlockNumber");
+        let block_request = self.block_request(Some(block_number.into())).await?;
+        let BlockRequest::Number(number) = block_request else { return Ok(None) };
+
+        if let Some(fork) = self.get_fork()
+            && fork.predates_fork_inclusive(number)
+        {
+            return Ok(fork.block_access_list_by_number(block_number).await?);
```
stevencartavia, line 2107: **`block_access_list_by_number` forwards the unresolved tag to upstream**

The handler resolves `number` for the `predates_fork_inclusive` check but then forwards the original `block_number`:

```rust
let BlockRequest::Number(number) = block_request else { return Ok(None) };
if let Some(fork) = self.get_fork()
    && fork.predates_fork_inclusive(number)
{
    return Ok(fork.block_access_list_by_number(block_number).await?);
}
```

For a tag like `latest`/`safe`, upstream resolves it against its own current chain view, not the fork block. On a fresh fork `latest` passes the inclusive check and gets forwarded as `"latest"`, returning the BAL of a later block than the fork block. Sibling handlers all forward the resolved number (`get_proof`, `fee_history`, `block_receipts`).

decofe:
Consolidated into #15476.

PR #15442, opened 2026-06-30
feat(anvil): support raw block access list

mattsse:
Adds Anvil support for the reth-compatible `eth_getBlockAccessListRaw` endpoint. The endpoint resolves the requested block, forwards pre-fork requests to the upstream provider, and returns `null` for local Anvil blocks because Anvil does not currently store EIP-7928 block access list payloads.

This completes the block access list RPC surface while leaving future stored-BAL behavior for a separate scoped follow-up.

Authored with AI assistance.

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -1553,6 +1553,9 @@ impl EthApi<FoundryNetwork> {
                     self.block_by_number(num).await.to_rpc_result()
                 }
             }
+            EthRequest::EthGetBlockAccessListRaw(block_id) => {
+                self.block_access_list_raw(block_id).await.to_rpc_result()
+            }
             EthRequest::EthGetTransactionCount(addr, block) => {
                 self.transaction_count(addr, block).await.to_rpc_result()
             }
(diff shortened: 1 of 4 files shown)


crates/anvil/src/eth/api.rs
```diff
@@ -2087,6 +2090,23 @@ impl EthApi<FoundryNetwork> {
...
+    /// Handler for ETH RPC call: `eth_getBlockAccessListRaw`
+    pub async fn block_access_list_raw(&self, block_id: BlockId) -> Result<Option<Bytes>> {
+        node_info!("eth_getBlockAccessListRaw");
+        let block_request = self.block_request(Some(block_id)).await?;
+        let BlockRequest::Number(number) = block_request else { return Ok(None) };
+
+        if let Some(fork) = self.get_fork()
+            && fork.predates_fork_inclusive(number)
+        {
+            return Ok(fork.block_access_list_raw(block_id).await?);
```
stevencartavia, line 2104: For `BlockId::Number`, `block_access_list_raw` resolves the request to `number` for the `predates_fork_inclusive` check, but then forwards the original `block_id` upstream. For mutable tags (`latest`/`safe`/`finalized`) on a forked node this queries the wrong block — e.g. forked at block 5 with upstream at 20, `eth_getBlockAccessListRaw("latest")` forwards `"latest"` and hits upstream block 20, not the fork block 5. The nearby `get_proof` handler does it correctly via `Some(number.into())`.

decofe:
Consolidated into #15476.

PR #15443, opened 2026-06-30
feat(anvil): support trace call

mattsse:
Adds Anvil support for the parity/reth-compatible `trace_call` endpoint. The endpoint accepts a transaction request, requested trace types, and an optional block id, defaults empty trace type input to `trace`, forwards pre-fork historical calls to the upstream provider, and executes local calls through Anvil's parity tracing inspector without mutating state.

The integration coverage exercises `trace`, `vmTrace`, and `stateDiff` together and verifies the simulated call leaves account balances unchanged.

Authored with AI assistance.


crates/anvil/src/eth/api.rs
```diff
@@ -1260,6 +1262,45 @@ impl<N: Network> EthApi<N> {
...
+    ) -> Result<TraceResults>
+    where
+        N: Network<TxEnvelope = FoundryTxEnvelope, ReceiptEnvelope = FoundryReceiptEnvelope>,
+    {
+        node_info!("trace_call");
+        if trace_types.is_empty() {
+            trace_types.insert(TraceType::Trace);
+        }
+
+        let block_id = block_id.unwrap_or_else(BlockId::pending);
```
stevencartavia: `trace_call` defaults an omitted block to `pending`:

```rust
let block_id = block_id.unwrap_or_else(BlockId::pending);
```

This diverges from reth (`block_id.unwrap_or_default()` → `latest`) and from anvil's own `eth_call`/`debug_traceCall`, which default to `latest`. As a result, omitting the block param runs the trace on top of pending txpool state instead of the chain head.

Reproduced with `anvil --no-mining`, one pending (unmined) 5-ETH transfer from account0, then `trace_call` with `stateDiff` and no block param:

```
default (no block):  a0 pre-balance 0x21dd47d12058a376000   (pending tx applied: 5 ETH value + gas deducted)
explicit "latest":   a0 pre-balance 0x21e19e0c9bab2400000   (original)
```

grandizzy (approved):

crates/anvil/src/eth/api.rs
grandizzy: please recheck

mablr (approved):

PR #15445, opened 2026-06-30
feat(anvil): support trace raw transaction

mattsse:
This adds Anvil support for the parity `trace_rawTransaction` RPC endpoint. The handler decodes and validates a raw signed transaction, traces it against the requested block context with the existing parity tracing inspector, and returns `TraceResults` without importing the transaction into the mempool or committing state.

The change also covers request deserialization for the optional block argument and adds an integration test that verifies trace output while confirming nonce and balances remain unchanged.

Authored with AI assistance.


crates/anvil/src/eth/api.rs
```diff
@@ -1256,6 +1258,60 @@ impl<N: Network> EthApi<N> {
         node_info!("trace_replayBlockTransactions");
         self.backend.trace_replay_block_transactions(block, trace_types).await
     }
+
+    /// Traces a raw signed transaction without importing it into the mempool.
+    ///
+    /// Handler for RPC call: `trace_rawTransaction`.
+    pub async fn trace_raw_transaction(
```
stevencartavia: Nit: this could move to `impl EthApi<FoundryNetwork>` and reuse the existing `ensure_typed_transaction_supported` and `block_request` helpers instead of inlining them.

grandizzy: please recheck, thanks!

grandizzy (approved):

emmajam (approved):

PR #15448, opened 2026-06-30
feat(anvil): support block opcode gas

mattsse:
This adds Anvil support for reth's `trace_blockOpcodeGas` RPC endpoint. Local mined blocks are replayed from their parent state with a fresh `OpcodeGasInspector` for each transaction, and forked historical blocks can be delegated upstream through the same raw RPC method.

The request parser accepts reth-style `BlockId` parameters, and integration coverage verifies number and hash lookups for a block containing a storage-writing transaction.

Authored with AI assistance.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -3699,12 +3699,12 @@ where
         &self,
         block_id: BlockId,
     ) -> Result<Option<BlockOpcodeGas>, BlockchainError> {
-        if let Some((block, block_hash)) = self.get_block_with_hash(block_id.clone()) {
+        if let Some((block, block_hash)) = self.get_block_with_hash(block_id) {
             return self.mined_block_opcode_gas(&block, block_hash).map(Some);
         }
 
         if let Some(fork) = self.get_fork() {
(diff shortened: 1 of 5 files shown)


crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -3470,6 +3472,75 @@ impl<N: Network> Backend<N>
...
+
+        Err(BlockchainError::BlockNotFound)
+    }
+
+    fn mined_block_opcode_gas(
+        &self,
+        block: &Block,
+        block_hash: B256,
+    ) -> Result<BlockOpcodeGas, BlockchainError> {
+        let parent_hash = block.header.parent_hash;
```
stevencartavia, line 3729: `trace_blockOpcodeGas` on the genesis block errors with `-32001` instead of returning a `BlockOpcodeGas` with `transactions: []`. Reproduced via `number(0)`, genesis `hash`, and `earliest`. Cause: `mined_block_opcode_gas` looks up genesis's (zero) parent state before handling the empty-tx case; reth returns `Some(BlockOpcodeGas { ..., transactions: [] })`. Fix: early-return empty when `block.body.transactions.is_empty()`.

grandizzy (approved):

mablr (approved):

PR #15449, opened 2026-06-30
feat(anvil): support debug account info at

mattsse:
This adds Anvil support for reth's `debug_accountInfoAt` RPC endpoint. For locally mined blocks, Anvil replays the stored block from its parent state through the requested transaction index, then returns the target account balance, nonce, and code. Historical forked blocks can delegate the same request upstream.

The request parser accepts number and hash `BlockId` forms, and integration coverage verifies both forms after a value transfer.

Authored with AI assistance.


crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -3470,6 +3471,94 @@ impl<N: Network> Backend<N>
...
+        address: Address,
+    ) -> Result<Option<RpcAccountInfo>, BlockchainError> {
+        if let Some((block, _)) = self.get_block_with_hash(block_id) {
+            return self.mined_debug_account_info_at(&block, tx_index, address).map(Some);
+        }
+
+        if let Some(fork) = self.get_fork() {
+            let number = self.ensure_block_number(Some(block_id)).await?;
+            if fork.predates_fork_inclusive(number) {
+                return Ok(fork.debug_account_info_at(block_id, tx_index, address).await?);
```
stevencartavia: `debug_account_info_at` only resolves `block_id` locally for the `predates_fork_inclusive` check, but when it delegates to the fork it forwards the original tag (`latest`/`pending`/`safe`/`finalized`). So upstream resolves the tag against its *own* head, not the fork's.

Repro (anvil forked at block 1, frozen): `debug_accountInfoAt(latest, 0, addr)` returns the fork-point value, but after the upstream mines 3 more blocks the same call returns the upstream's block-4 value — even though the fork never advanced.

Other fork delegations (e.g. `transaction_by_block_number_and_index`) delegate the resolved number, not the tag. Same fix here — delegate `BlockId::number(resolved)` for number/tags, pass hashes through unchanged.

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -3694,81 +3657,92 @@ where
...
+    ) -> Result<Option<RpcAccountInfo>, BlockchainError> {
+        if let Some((block, _)) = self.get_block_with_hash(block_id) {
+            return self.mined_debug_account_info_at(&block, tx_index, address).map(Some);
         }
 
         if let Some(fork) = self.get_fork() {
             let number = self.ensure_block_number(Some(block_id)).await?;
             if fork.predates_fork_inclusive(number) {
-                return Ok(fork.trace_block_opcode_gas(block_id).await?);
+                return Ok(fork.debug_account_info_at(block_id, tx_index, address).await?);
```
grandizzy: This forwards the original `block_id` (tag) to the fork after only using `number` for the predates check, so upstream resolves `latest`/`pending`/`safe`/`finalized` against its own head — the drift @stevencartavia reproduced. Delegate the resolved id instead (hashes pass through unchanged):

```rust
let resolved = match block_id {
    BlockId::Hash(_) => block_id,
    _ => BlockId::number(number),
};
return Ok(fork.debug_account_info_at(resolved, tx_index, address).await?);
```

Worth adding a fork test here: a pre-fork block covers delegation, and `latest` on a frozen fork (advance upstream, re-query) catches the tag-drift.

grandizzy (commented):
Parser, dispatch, and the local `mined_debug_account_info_at` replay all look correct, and the tx-index bounds check + `AccountInfo` shape are right. One blocking correctness issue in the fork path (also flagged by @stevencartavia), plus a missing fork test.

The fork branch resolves the block to a number only for the `predates_fork_inclusive` check, then delegates the **original tag** upstream — so on a frozen fork, `latest`/`pending`/`safe`/`finalized` get resolved against the *upstream's* advancing head instead of the fork's. Other anvil fork delegations (e.g. `transaction_by_block_number_and_index`) forward the resolved number, not the tag.

Note: the just-merged sibling #15448 (`trace_blockOpcodeGas`) shipped the identical pattern, so whatever we settle on here should be applied there as a follow-up too.

Requesting: (1) delegate the resolved id, and (2) a fork-delegation test covering a pre-fork block + a `latest`-on-frozen-fork case.

grandizzy (approved):

stevencartavia (approved):

PR #15450, opened 2026-06-30
feat(anvil): support debug account info at

mattsse:
This adds Anvil support for reth's debug_accountInfoAt RPC endpoint. For locally mined blocks, Anvil replays the stored block from its parent state through the requested transaction index in a concrete in-memory database, then returns the target account balance, nonce, and code. Historical forked blocks can delegate the same request upstream.

The request parser accepts number and hash BlockId forms, and integration coverage verifies local blocks, fork-local blocks, and pre-fork delegation.

Authored with AI assistance.


mablr (dismissed):

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -3470,6 +3470,103 @@ impl<N: Network> Backend<N>
...
+                    requires: vec![],
+                    provides: vec![],
+                    priority: crate::eth::pool::transactions::TransactionPriority(0),
+                })
+            })
+            .collect();
+
+        let replay = |parent_state: &StateDb| -> Result<Option<TrieAccount>, BlockchainError> {
+            let mut replay_db = in_memory_db::MemDb::default();
+            replay_db.init_from_state_snapshot(parent_state.read_as_state_snapshot());
```
stevencartavia: This replays into a detached `MemDb` seeded from `parent_state.read_as_state_snapshot()`, then reads the account via `maybe_as_full_db()` + `db.get(&address)`. `read_as_state_snapshot` copies the cached state maps but not the live `SharedBackend`, so the `MemDb` falls back to an empty DB (no fork fall-through) and uncached fork accounts look absent (and replay can read zero/default state for them).

Repro: fork mainnet, mine a local block with a dev→dev transfer that doesn't touch WETH, then `debug_accountAt(block, 0, WETH)`. It intermittently returns `null` for WETH (~1/10 at a fresh in-range fork block, cache-warmth/race dependent) even though WETH exists — reth falls through to historical state here.

crates/anvil/tests/it/traces.rs
```diff
@@ -68,6 +71,43 @@ async fn test_get_transfer_parity_traces() {
     assert_eq!(traces, block_traces);
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn test_debug_account_at_local() {
```
stevencartavia: The only `debug_accountAt` test added is `test_debug_account_at_local` (no fork). The fork paths are untested: upstream delegation in `ClientFork::debug_account_at`, and the mined-on-fork local-block path through `mined_debug_account_at`. A test that mines a local block after forking and calls `debug_accountAt` for it would cover (and catch) the fall-through issue above; a pre-fork block would cover delegation.

crates/anvil/src/eth/backend/mem/mod.rs
grandizzy: pls recheck

crates/anvil/tests/it/traces.rs
grandizzy: pls recheck

grandizzy (dismissed):

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -3942,6 +3942,103 @@ where
...
+                &|pending, account| self.validate_pool_transaction_for(pending, account, &evm_env),
+            );
+
+            let _ = <AnvilCacheDB<&StateDb> as revm::Database>::basic(&mut replay_db, address)?;
+            let db = replay_db.maybe_as_full_db().ok_or(BlockchainError::DataUnavailable)?;
+            let Some(account) = db.get(&address) else {
+                return Ok(None);
+            };
+
+            let storage_root = storage_root(&account.storage);
```
stevencartavia: `storage_root(&account.storage)` returns `EMPTY_ROOT_HASH` for fork contracts with storage. The `basic(address)` prime loads only account info into the cache, not storage slots, so `account.storage` is empty and the computed root is empty (a partially-touched fork contract would give a root over only the cached/read-or-written slots — also wrong). Balance and code_hash are correct via fall-through.

Repro: fork a chain, mine a local block that doesn't touch a storage-holding contract, then `debug_accountAt(block, 0, contract)`. `storage_root` comes back as `EMPTY_ROOT_HASH` while the real storage root is non-empty (verified via `eth_getProof` upstream and via anvil-forks-anvil against a deployed contract).

grandizzy: pls check fec7920a4

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -3942,6 +3942,119 @@ where
...
+            let db = replay_db.maybe_as_full_db().ok_or(BlockchainError::DataUnavailable)?;
+            let Some(account) = db.get(&address) else {
+                return Ok(None);
+            };
+
+            let storage_root = if let Some(fork_account) = fork_account
+                && fork_account.storage_root != alloy_trie::EMPTY_ROOT_HASH
+            {
+                if account.storage.is_empty() {
+                    fork_account.storage_root
```
stevencartavia: This returns a stale storage root. `fork_account` is fetched at `fork.block_number()` (the fork base), and `account.storage.is_empty()` only proves the queried block's replay didn't cache slots — not that the account's storage is unchanged. If an earlier local block modified the account's storage, that change lives in the queried block's parent state but isn't reflected here, so the fork-base root is returned instead of the current one.

Repro (deterministic): two forks from the same origin (deploys a storage contract "v0"). Fork A: block1 `setValue("v1")`, block2 dev→dev → `getValue()` = "v1". Fork B: block1 dev→dev, block2 dev→dev → `getValue()` = "v0". `debug_accountAt(block2, 0, contract)` returns the **same** fork-base storage root on both, despite A's storage genuinely differing.

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -3942,6 +3942,119 @@ where
...
+                } else {
+                    return Err(BlockchainError::DataUnavailable);
```
stevencartavia: `account.storage.is_empty()` isn't a valid "storage was modified" test: revm's `CacheDB` caches slots on `SLOAD`, so a read-only access populates `account.storage` and trips this error even though nothing changed and the correct root (the fork-base root) is available.

Repro (deterministic): on a fork, one tx = `Multicall.aggregate([(SimpleStorage, getValue())])` — a CALL that only reads the contract's storage. `debug_accountAt(thatBlock, 0, SimpleStorage)` returns `-32603 "Required data unavailable"` even though the storage was never written.

stevencartavia (approved):

grandizzy (approved):

PR #15451, opened 2026-06-30
feat(anvil): support trace call tx index

mattsse:
This adds support for reth's `debug_traceCall` `txIndex` option in Anvil. When the option is present, Anvil resolves the target block, replays that block from its parent state through the requested transaction index, and then executes the traced call against that intra-block state. Calls without `txIndex` continue through the existing post-state path.

The implementation keeps state and block overrides on the traced call after replay, delegates forked historical requests upstream, and adds coverage for parsing `txIndex`. The integration test mines two writes into a single block and verifies `debug_traceCall` sees the different storage values at transaction indexes 0 and 1.

Authored with AI assistance.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -65,14 +65,15 @@ use alloy_primitives::{
 };
 use alloy_rlp::Decodable;
 use alloy_rpc_types::{
-    AccessList, Block as AlloyBlock, BlockId, BlockNumberOrTag as BlockNumber, BlockTransactions,
-    EIP1186AccountProofResponse as AccountProof, EIP1186StorageProof as StorageProof, Filter,
-    Header as AlloyHeader, Index, Log, Transaction, TransactionReceipt,
+    AccessList, Block as AlloyBlock, BlockId, BlockNumberOrTag as BlockNumber, BlockOverrides,
+    BlockTransactions, EIP1186AccountProofResponse as AccountProof,
+    EIP1186StorageProof as StorageProof, Filter, Header as AlloyHeader, Index, Log, Transaction,
+    TransactionReceipt,
     anvil::Forking,
     request::TransactionRequest,
     serde_helpers::JsonStorageKey,
     simulate::{SimBlock, SimCallResult, SimulatePayload, SimulatedBlock},
-    state::EvmOverrides,
+    state::{EvmOverrides, StateOverride},
     trace::{
         filter::TraceFilter,
         geth::{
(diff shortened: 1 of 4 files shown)


crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -2977,147 +2978,317 @@ where
...
+    ) -> Result<GethTrace, BlockchainError> {
+        let transaction_count = block.body.transactions.len();
+        if tx_index >= transaction_count {
+            return Err(BlockchainError::RpcError(RpcError::invalid_params(format!(
+                "tx_index {tx_index} out of bounds for block with {transaction_count} transactions"
+            ))));
+        }
+
+        let pool_txs: Vec<Arc<PoolTransaction<FoundryTxEnvelope>>> = block.body.transactions
+            [..=tx_index]
```
stevencartavia: `mined_trace_call_at_tx_index` replays `block.body.transactions[..=tx_index]` (inclusive), but reth replays `[0, tx_index)` — its `debug_trace_call_at_tx_index` passes `transactions[tx_index].tx_hash()` to `replay_transactions_until`, which breaks *before* committing the target tx. So `txIndex=N` should trace the call as if it were at position N (state after txs with index `< N`, i.e. before tx N), not after tx N.

Repro: block with tx0=`set(1)`, tx1=`set(2)`, tx2=`set(3)`, value before block = 9. This PR returns txIndex 0→1, 1→2, 2→3; reth returns 0→9, 1→1, 2→2. The pre-block state is unreachable, and the last valid index just equals the post-state.

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -2977,147 +2978,317 @@ where
...
+                    provides: vec![],
+                    priority: crate::eth::pool::transactions::TransactionPriority(0),
+                })
+            })
+            .collect();
 
-                            let tracing_inspector = inspector.tracer.expect("tracer disappeared");
+        let trace = |parent_state: &StateDb| -> Result<GethTrace, BlockchainError> {
+            let mut replay_db = in_memory_db::MemDb::default();
+            replay_db.init_from_state_snapshot(parent_state.read_as_state_snapshot());
```
stevencartavia: The parent state is copied into a plain `MemDb` via `read_as_state_snapshot()`, which preserves already-cached state but drops the live `SharedBackend`, so lazy fork fetch is lost. Any account/code/storage not already cached when the parent snapshot was taken is missing, with no way to fetch it from the remote. Since `get_block()` is checked before the `predates_fork_inclusive()` delegation, post-fork local blocks hit this path. (The existing `replay_tx_with_inspector` / `with_database_at` avoid this by wrapping the `StateDb` directly.)

Repro on a mainnet fork: mine a local block with at least one tx, then `debug_traceCall` `USDC.totalSupply()` (never touched before). With `txIndex=0` → empty output (code-less account); without `txIndex` → correct total supply with the proxy delegatecall.

mablr (approved):
LGTM

stevencartavia (approved):

PR #15452, opened 2026-06-30
feat(anvil): support debug seed hash

mattsse:
This adds Anvil support for reth's `debug_seedHash` RPC endpoint. Reth currently accepts a block number and returns the default seed hash, so Anvil now parses the same request and returns the zero hash instead of method-not-found.

The request parser accepts both decimal and hex quantity block numbers, and the integration test verifies the HTTP JSON-RPC endpoint response.

Authored with AI assistance.


stevencartavia (approved):

decofe:
Consolidated into #15476.

PR #15454, opened 2026-06-30
feat(anvil): support debug db ancient

mattsse:
This adds Anvil support for reth's `debug_dbAncient` RPC endpoint. Reth exposes this endpoint with a freezer table kind and block number argument and currently returns success with no result payload, so Anvil now parses the same request shape and returns JSON `null` instead of method-not-found.

Authored with AI assistance.


stevencartavia (approved):

decofe:
Consolidated into #15476.

PR #15460, opened 2026-06-30
feat(anvil): support debug mem stats

mattsse:
This adds Anvil support for reth's `debug_memStats` RPC endpoint. Reth exposes this no-argument endpoint and currently returns success with no result payload, so Anvil now parses the same request and returns JSON `null` instead of method-not-found.

Authored with AI assistance.


stevencartavia (approved):

decofe:
Consolidated into #15476.

PR #15468, opened 2026-06-30
feat(anvil): support debug modified accounts by number

mattsse:
This adds Anvil support for reth's `debug_getModifiedAccountsByNumber` RPC endpoint. Reth exposes this two-number endpoint and currently returns success with no result payload, so Anvil now parses the same request and returns JSON `null` instead of method-not-found.

Authored with AI assistance.


stevencartavia (dismissed):

mablr (dismissed):

mablr (approved):

grandizzy (approved):
approving to be at par with reth but note is a noop for both

PR #15472, opened 2026-06-30
feat(anvil): support debug account range

mattsse:
This adds Anvil support for reth's `debug_accountRange` RPC endpoint. Reth exposes this account-range endpoint and currently returns success with no result payload, so Anvil now parses the same request and returns JSON `null` instead of method-not-found.

Authored with AI assistance.


stevencartavia (dismissed):

mablr (dismissed):

mablr (dismissed):

crates/anvil/tests/it/api.rs
```diff
@@ -698,6 +698,17 @@ async fn can_get_code_by_hash() {
...
+    let provider = handle.http_provider();
+
+    let response: serde_json::Value = provider
+        .raw_request("debug_accountRange".into(), ("latest", "0x", 1u64, false, false, false))
+        .await
+        .unwrap();
+    assert_eq!(response, serde_json::Value::Null);
+}
+
 async fn can_debug_free_os_memory() {
```
grandizzy, line 713: `can_debug_free_os_memory` lost its `#[tokio::test]` attribute when the new test was inserted above it, so it's now an uncalled fn (dropped test + `never used` warning that fails the clippy CI job).

```suggestion
#[tokio::test(flavor = "multi_thread")]
async fn can_debug_free_os_memory() {
```

grandizzy (changes requested):
Implementation looks good and matches reth's `debug_accountRange` no-op stub (reth itself ignores the params and returns `Ok(())`), and the param order matches reth/geth.

One blocker: the new test accidentally consumed the `#[tokio::test]` attribute from `can_debug_free_os_memory`, which is now a dead, uncalled `async fn`. It's dropped from the suite (no longer listed by `--list`) and trips `-Dwarnings` in the clippy CI job (`--all-targets`). Re-adding the attribute fixes it.

grandizzy (approved):

decofe:
Consolidated into #15476.

PR #15473, opened 2026-06-30
feat(anvil): support debug trace chain

mattsse:
This adds Anvil support for reth's `debug_traceChain` RPC endpoint. Reth exposes this block-range endpoint but currently returns an explicit internal `unimplemented` error, so Anvil now parses the same request and returns that error instead of method-not-found.

Authored with AI assistance.


stevencartavia (approved):

decofe:
Consolidated into #15476.

PR #15474, opened 2026-06-30
feat(anvil): support empty tx conditionals

mattsse:
This adds Anvil parsing and dispatch for reth's exported `eth_sendRawTransactionConditional` eth extension endpoint. Empty/default transaction conditions are equivalent to `eth_sendRawTransaction`, so Anvil now accepts those and reuses the existing raw transaction path.

Non-empty preconditions are rejected explicitly instead of being silently ignored, because Anvil does not currently preserve conditional metadata in its transaction pool. That keeps the endpoint compatible for default condition callers while making the remaining semantic gap visible.

Authored with AI assistance.


grandizzy (approved):
lgtm

0xKarl98 (approved):

PR #15477, opened 2026-06-30
feat(symbolic): import fuzz corpus seeds

figtracer:
## Summary
- add `--symbolic-use-fuzz-corpus` and bounded `symbolic.corpus_seed_limit` config
- import matching single-call fuzz corpus entries into symbolic fuzz-test runs as branch-priority hints
- emit imported seed metadata in symbolic JSON output and document the import path

### DIFF
--- a/crates/evm/symbolic/src/executor/opcodes.rs
+++ b/crates/evm/symbolic/src/executor/opcodes.rs
@@ -467,22 +467,15 @@ impl SymbolicExecutor {
                             self.branch_is_sat_or_defer(&false_state.constraints)?;
                         trace!(true_feasible, false_feasible, "JUMPI symbolic branch");
                         match (true_feasible, false_feasible) {
-                            (true, true)
-                                if false_state.corpus_seed_model_count()
-                                    > true_state.corpus_seed_model_count() =>
-                            {
-                                push_preferred_branch(
-                                    worklist,
-                                    false_state,
-                                    true_state,
-                                    self.config.exploration_order,
-                                );
-                            }
                             (true, true) => {
-                                push_preferred_branch(
+                                let true_seed_count = true_state.corpus_seed_model_count();
+                                let false_seed_count = false_state.corpus_seed_model_count();
+                                push_seed_ordered_branches(
                                     worklist,
                                     true_state,
+                                    true_seed_count,
                                     false_state,
+                                    false_seed_count,
                                     self.config.exploration_order,
                                 );
                             }
@@ -778,10 +771,32 @@ impl SymbolicExecutor {
     }
 }
 
-fn push_preferred_branch(
-    worklist: &mut VecDeque<PathState>,
-    preferred: PathState,
-    secondary: PathState,
+fn push_seed_ordered_branches<T>(
+    worklist: &mut VecDeque<T>,
+    true_branch: T,
+    true_seed_count: usize,
+    false_branch: T,
+    false_seed_count: usize,
+    order: SymbolicExplorationOrder,
+) {
+    match false_seed_count.cmp(&true_seed_count) {
+        std::cmp::Ordering::Greater => {
+            push_preferred_branch(worklist, false_branch, true_branch, order);
+        }
+        std::cmp::Ordering::Less => {
+            push_preferred_branch(worklist, true_branch, false_branch, order);
+        }
+        std::cmp::Ordering::Equal => {
+            worklist.push_back(true_branch);
+            worklist.push_back(false_branch);
+        }
+    }
+}
+
+fn push_preferred_branch<T>(
+    worklist: &mut VecDeque<T>,
+    preferred: T,
+    secondary: T,
     order: SymbolicExplorationOrder,
 ) {
     match order {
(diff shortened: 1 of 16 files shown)


mattsse (dismissed):
lgtm

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -450,23 +450,45 @@ impl SymbolicExecutor {
...
+                                    self.config.exploration_order,
+                                );
+                            }
+                            (true, true) => {
+                                push_preferred_branch(
+                                    worklist,
+                                    true_state,
+                                    false_state,
+                                    self.config.exploration_order,
+                                );
```
mablr: This seems to change DFS branch order even when no corpus seeds are active. The old code pushed `true_state` then `false_state`; since DFS pops from the back, it explored the false branch first. In the equal-count case, including zero seed models, `push_preferred_branch(..., true_state, false_state, Dfs)` pushes false then true, so DFS explores true first. That can change bounded symbolic results for existing DFS runs unrelated to corpus import. Is this made on purpose ?

mablr (dismissed):
sgtm

crates/evm/symbolic/src/executor/run.rs
```diff
@@ -155,13 +155,19 @@ impl SymbolicExecutor {
         let code = SymCode::from_bytecode(&bytecode);
         let mut worklist = VecDeque::new();
         for calldata in SymbolicCalldata::variants(input.function, &self.config)? {
+            let corpus_seed_models = input
+                .corpus_seeds
+                .iter()
+                .filter_map(|seed| calldata.seed_model(seed).map(Arc::new))
+                .collect();
```
grandizzy, line 203: **Medium — seeded ABI variants aren't prioritized at the root level.** Roots are enqueued in plain `SymbolicCalldata::variants(...)` order; the corpus-aware ordering only applies *after* a root is popped and reaches a symbolic `JUMPI`. If a corpus seed matches a later dynamic calldata variant (e.g. `bytes` length 2 when symbolic lengths are `[1, 2]`), BFS still explores the length-1 root first. With a tight `--symbolic-width`/`max_paths`, the run can hit the path limit before ever reaching the seeded variant — defeating the "use corpus as path-priority hints" intent.

Suggested fix: collect the roots into a `Vec` first, then enqueue ordered by `corpus_seed_model_count()`, stable when counts are equal:
- BFS (`pop_front`): push higher seed-count roots **earlier**.
- DFS (`pop_back`): push higher seed-count roots **later**.

No-op when all counts are equal or zero, so non-corpus runs are unaffected.

crates/forge/src/runner.rs
```diff
@@ -1004,7 +1022,9 @@ impl<'a, FEN: FoundryEvmNetwork> ContractRunner<'a, FEN> {
                 }
 
                 let sig = func.signature();
```
grandizzy, line 1024: **Medium — `symbolic.use_fuzz_corpus` is evaluated before function-level inline config is applied.** This `should_symbolically_import_fuzz_corpus(&self.config, func)` check runs against the contract-level config, but `apply_function_inline_config(func)` only runs later inside `FunctionRunner::run`. Consequences:
- A function-level inline config that *enables* `symbolic.use_fuzz_corpus` won't upgrade that fuzz test to symbolic.
- A function-level inline config that *disables* it can still leave the test classified as `SymbolicTest`, producing an unguided symbolic run instead of the normal fuzz run.

Suggested fix: drop this clause from the pre-inline `kind` selection and recompute the fuzz→symbolic upgrade *after* `apply_function_inline_config(func)` (before the `fuzz_only`/showmap skip logic), so the effective merged config decides:
```rust
let kind = if should_symbolically_import_fuzz_corpus(&self.config, func) {
    TestFunctionKind::SymbolicTest
} else {
    kind
};
```

crates/forge/src/runner.rs
```diff
@@ -1654,12 +1674,110 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                };
+                let Some(input) = self.symbolic_corpus_seed_input(func, &tx_seq) else {
+                    metadata.skipped += 1;
+                    continue;
+                };
+                metadata.used.push(SymbolicCorpusSeedRef {
+                    path: entry.path,
+                    calldata: input.calldata.clone(),
+                });
+                imported.inputs.push(input);
```
grandizzy: **Low — JSON `used` can over-report.** `metadata.used` is populated right after ABI decoding, but the executor later silently drops seeds it can't model for a given symbolic calldata variant (`.filter_map(|seed| calldata.seed_model(seed)...)` in `run.rs`). Dynamic length mismatch, printable-string constraints, or non-canonical calldata can therefore appear under `corpus_seeds.used` even when no root/branch actually used them as a hint.

Suggested fix: either rename/describe this field as `imported`, or track the seeds that actually produced a model after `seed_model` succeeds and report those as `used`.

grandizzy (changes requested):
Please see findings inline, thanks!

grandizzy (approved):
lgtm

mablr (approved):

PR #15480, opened 2026-06-30
fix(forge): clear history storage when etched

decofe:
## Summary

- clear Foundry's synthetic EIP-2935 history storage when `vm.etch` replaces the canonical history contract bytecode
- add a Prague regression that etches the history address, rolls to a high block whose ring slot was pre-seeded, and verifies the slot remains zero

## Tests

- `cargo test -p forge eip2935_history_storage_in_prague_tests -- --nocapture`

Fixes #15479

Prompted by: @grandizzy

PR #15481, opened 2026-06-30
test(forge): bump snekmate external coverage

mablr:
## Motivation

Bump snekmate after https://github.com/foundry-rs/foundry/pull/15480#pullrequestreview-4599556369

### DIFF
--- a/crates/forge/tests/cli/ext_integration.rs
+++ b/crates/forge/tests/cli/ext_integration.rs
@@ -77,8 +77,9 @@ fn solady() {
 #[test]
 #[cfg_attr(windows, ignore = "Windows cannot find installed programs")]
 fn snekmate() {
-    ExtTester::new("pcaversaccio", "snekmate", "2834e9750b39058f06129c8fde1b12fb832a50d5")
+    ExtTester::new("pcaversaccio", "snekmate", "df2816d6a5ecdb8dcc1257e22089422825fd5e27")
         .fuzz_runs(256)
+        .python_package("git+https://github.com/vyperlang/vyper@v0.5.0a3")
         .install_command(&["pnpm", "install", "--prefer-offline"])
         // Try npm if pnpm fails / is not installed.
         .install_command(&["npm", "install", "--prefer-offline"])
--- a/crates/forge/tests/cli/ext_integration.rs
+++ b/crates/forge/tests/cli/ext_integration.rs
@@ -77,7 +77,8 @@ fn solady() {
 #[test]
 #[cfg_attr(windows, ignore = "Windows cannot find installed programs")]
 fn snekmate() {
-    ExtTester::new("pcaversaccio", "snekmate", "1a54931129f2814cbbd7ddbafb4005707f8a5bf8")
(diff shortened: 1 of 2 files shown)


crates/forge/tests/cli/ext_integration.rs
```diff
@@ -77,7 +77,8 @@ fn solady() {
 #[test]
 #[cfg_attr(windows, ignore = "Windows cannot find installed programs")]
 fn snekmate() {
-    ExtTester::new("pcaversaccio", "snekmate", "1a54931129f2814cbbd7ddbafb4005707f8a5bf8")
+    ExtTester::new("pcaversaccio", "snekmate", "2834e9750b39058f06129c8fde1b12fb832a50d5")
+        .fuzz_runs(10_000)
```
mablr: 10K like on the repo, let's see if it times out.

figtracer (dismissed):

crates/forge/tests/cli/ext_integration.rs
grandizzy: mm, I think will be too much for running this on each pr / commit

figtracer: if it takes much longer we can turn it down a bit, we might not even need this many runs

mattsse (dismissed):
Thanks for bumping the external coverage. I think this needs one more harness update before it can merge.

The new snekmate revision requires Vyper `~=0.5.0a4`, but Foundry's external test harness and CI still provision Vyper `0.4.3` (`crates/test-utils/src/util.rs` downloads `v0.4.3`, and the workflows install `vyper==0.4.3`). With this PR checked out, the focused test fails before running the snekmate suite:

```text
cargo nextest run -p forge --test cli ext_integration::snekmate

Error: Found Vyper sources, but no compiler versions are available for it
```

Suggested fix: make the external test harness able to select the required Vyper version for a fixture, then set snekmate to `0.5.0a4` alongside the new commit. For example, add an `ExtTester::vyper_version(...)`/`vyper_url(...)` style override and use it to place that compiler on `PATH` for the spawned `forge test`; alternatively, if Foundry wants a single Vyper version in CI, bump the shared test Vyper provisioning from `0.4.3` to a version that still keeps the existing Vyper tests passing and supports this snekmate revision. If neither is intended for this PR, pin snekmate to the latest revision that remains compatible with Vyper `0.4.3`.

mablr:
@mattsse mb my local venv contaminated my local testing, pinned to a snekmate commit https://github.com/pcaversaccio/snekmate/commit/df2816d6a5ecdb8dcc1257e22089422825fd5e27 from last week to work with `vyper@0.5.0a3` (as `vyper@0.5.0a4` is not yet tagged). Now it will be really quick to bump.

crates/forge/tests/cli/ext_integration.rs
mablr: 256 seems fine

grandizzy (approved):
nice!

stevencartavia (approved):

PR #15484, opened 2026-06-30
fix(forge): dedup corpus entries by timestamp

mablr:
## Motivation

#15227 follow-up

Minimal fix to preserve same-UUID entries when timestamps differ, while still deduping identical generated worker-shard copies.


mattsse (approved):
LGTM

figtracer (approved):

PR #15485, opened 2026-06-30
feat(debugger): add command prompt

figtracer:
## Summary
- add `:` command mode to the debugger TUI
- support `:continue <pc>` / `:pc <pc>` for PC jumps
- support `:mem <offset>`, `:calldata <offset>`, and `:returndata <offset>` for buffer jumps
- expose the command prompt in the footer

Refs #927

### DIFF
--- a/crates/debugger/src/tui/context.rs
+++ b/crates/debugger/src/tui/context.rs
@@ -70,6 +70,8 @@ pub(crate) struct TUIContext<'a> {
     pub(crate) pc_input: Option<String>,
     /// Current active-buffer byte offset prompt contents, if the prompt is active.
     pub(crate) buffer_offset_input: Option<String>,
+    /// Current debugger command prompt contents, if the prompt is active.
+    pub(crate) command_input: Option<String>,
     /// Current opcode search prompt contents, if the prompt is active.
     pub(crate) opcode_search_input: Option<String>,
     /// Last opcode search term, used by repeat-search shortcuts.
@@ -87,7 +89,6 @@ pub(crate) struct TUIContext<'a> {
     pub(crate) buf_utf: bool,
     pub(crate) show_shortcuts: bool,
     pub(crate) show_source: bool,
-    pub(crate) show_variables: bool,
     /// The currently active buffer (memory, calldata, returndata) to be drawn.
     pub(crate) active_buffer: BufferKind,
 }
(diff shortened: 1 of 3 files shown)


mattsse (dismissed):
LGTM

mablr (dismissed):
Tested locally, sgtm.

figtracer:
@mablr will push something to make this more pretty vim-like

mablr (commented):
@figtracer I improved the UI by spawning an input block for command prompt, ptal.

figtracer:
looks much nicer @mablr

mablr (dismissed):

crates/debugger/src/tui/context.rs
```diff
@@ -733,6 +807,31 @@ const fn toggle_state(enabled: bool) -> &'static str {
...
 
+const fn buffer_name(buffer: &BufferKind) -> &'static str {
+    match buffer {
+        BufferKind::Memory => "memory",
+        BufferKind::Calldata => "calldata",
+        BufferKind::Returndata => "returndata",
+    }
+}
+
+const fn buffer_kind(buffer: &BufferKind) -> BufferKind {
```
grandizzy: `buffer_kind` is a redundant identity helper. `BufferKind` derives `Copy`, so this variant-for-variant match just returns its input. Drop the helper and pass `self.active_buffer` directly at the call site (L583):

```rust
// L583
self.goto_buffer_offset(self.active_buffer, input);
```

No borrow issue — it's copied before the `&mut self` call. Then delete the `buffer_kind` fn.

grandizzy (commented):
Reviewed the command-prompt flow and footer drawing — looks solid and well-tested (modal blocking, named-buffer jumps, optional leading colon, help/usage errors, bordered-block draw). Verified the `Layout::split` with `Length(0)` stays safe, `footer_height` matches `draw_command_prompt` in both shortcut states, and buffer jumps validate the target before switching `active_buffer`. One trivial cleanup below; no blockers.

Optional nits (non-blocking): `:help` omits the aliases (`p`/`cont`/`c`/`cd`/`ret`/`rd`), and long input in the height-3 pane can wrap/clip the cursor — same limitation as the existing single-line prompts.

crates/debugger/src/tui/context.rs
mablr: > `BufferKind` derives `Copy`

not yet but this is doable

mablr: https://github.com/foundry-rs/foundry/pull/15485/commits/cc5d294386fbc61eb325e661862081d882c94a71

mablr (dismissed):

mablr (approved):

grandizzy (approved):

PR #15487, opened 2026-06-30
perf(symbolic): avoid child state clone churn

figtracer:
### Changes
- Removes the executor worklist helper wrappers and keeps the module declarations at the top of `executor/mod.rs`.
- Builds `PathState::child` directly with the supplied call frame instead of cloning the parent state and then replacing the frame/resetting loop-local state.

### Numbers
- `PathState::child(frame.clone())` probe: `319.5ms ± 8.0ms` baseline -> `58.7ms ± 0.5ms` current, `5.44x ± 0.15` faster.
- Samply on the same probe: `1524` benchmark-thread samples -> `294`, about `80.7%` fewer samples.
- Method: temporary ignored state-level probe in both trees, 500,000 child constructions from a parent with populated stack, memory, returndata, and loop-jump map. The probe is not committed.


mablr (dismissed):
SGTM

crates/evm/symbolic/src/executor/calls.rs
```diff
@@ -606,7 +606,7 @@ impl SymbolicExecutor {
                 }
                 let mut constraints = state.constraints.clone();
                 constraints.push(condition);
-                let sat = self.solver.is_sat(&constraints)?;
+                let sat = self.solver.is_sat_branch(&constraints)?;
```
stevencartavia: `constraints_with_condition` switching `is_sat` → `is_sat_branch` isn't behavior-preserving. `is_sat_branch` may defer hard-arith constraints without a local witness and return `SolverUnknown` before making an SMT query, where `is_sat` queries the backend and returns a definitive answer in this case.

Reproduced with z3:
- master: `constraints_with_condition([x/3 == 7] + z==5)` → `Ok((_, true))`
- this PR: same input → `Err(SolverUnknown)`

Since the helper uses `?`, that error aborts the run as `Incomplete` instead of deferring/pruning that branch and continuing with decidable siblings like `branch_is_sat_or_defer` does, so it can miss counterexamples on sibling paths. It's also used by mocks/expected-call matching, expected reverts, expected emits, and assume-no-revert filters, with no guarantee the path is free of hard arithmetic.

stevencartavia (dismissed):

DaniPopes (dismissed):
can you plz hold off on this for a bit, i'll have a pr to refactor the state again to use hashconsing, might overlap

crates/evm/symbolic/src/executor/calls.rs
```diff
@@ -1112,11 +1112,14 @@ impl SymbolicExecutor {
             parents.push_back(parent);
         }
 
-        let Some(first) = pop_batch(&mut parents, self.config.exploration_order) else {
+        let Some(first) = (match self.config.exploration_order {
```
grandizzy, line 1115: can we extract this in a helper and reuse?

grandizzy (approved):
lgtm, left one optional nit

mablr (approved):

PR #15491, opened 2026-06-30
fix(foundryup): sort nightly Atom entries by updated time

decofe:
## Summary
- keep `foundryup -i nightly` on the GitHub releases Atom feed for the normal path to avoid `api.github.com` rate limits
- parse each Atom entry's release link instead of grepping all `nightly-*` occurrences from changelog content
- sort nightly candidates by the entry `<updated>` timestamp before selecting/installability fallback
- preserve the GitHub REST releases fallback only when the Atom feed cannot be parsed/resolved, sorting API results by `published_at`
- bump `FOUNDRYUP_INSTALLER_VERSION` from `1.9.0` to `1.9.1`

## Testing
- `bash foundryup/test.sh`
- `git diff --check`

Prompted by: @grandizzy


foundryup/test.sh
```diff
@@ -148,16 +156,8 @@ check_eq "latest calls API when redirect fails" "1" "$(wc -l < "$api_marker" | t
...
 resolve_version_and_tag
-check_eq "nightly resolves to nightly-<sha> via feed" "nightly-$NIGHTLY_SHA" "$FOUNDRYUP_TAG"
-check_eq "nightly does not call API when feed works" "0" "$(wc -l < "$api_marker" | tr -d ' ')"
-
-: > "$api_marker"
-curl() { return 1; }
-fetch() { echo called >> "$api_marker"; printf '%s' "$NIGHTLY_API_JSON"; }
-FOUNDRYUP_VERSION="nightly"
-resolve_version_and_tag
-check_eq "nightly falls back to API tag" "nightly-deadbeef" "$FOUNDRYUP_TAG"
```
mablr, line 159: This deletes the nightly feed-failure API fallback test while foundryup/foundryup also removes the fallback itself. That regresses `foundryup -i nightly` for cases where `releases.atom` is temporarily unavailable or unparsable but the GitHub releases API would still return a nightly. The case is narrow imo, but it worths mentioning at least.

mablr (commented):
Sgtm, just one non-blocking remark.

foundryup/test.sh
decofe: You were right; I should have fixed the regression instead of just documenting it. Updated the PR to keep Atom as the normal path, but preserve the GitHub REST releases fallback only when the Atom feed cannot be parsed/resolved. The fallback sorts by `published_at`, and the restored test covers the feed-failure path.

stevencartavia (approved):

mablr (approved):

PR #15492, opened 2026-06-30
refactor(config): parse halmos args with shlex

DaniPopes:
Refactors legacy @custom:halmos inline config translation to use shlex for shell-like argument splitting and a data table of supported Halmos flags to map into Foundry symbolic config.

This removes the bespoke split_quoted_args helper and reuses shlex for symbolic solver command splitting as well, so both paths rely on the same argv parsing behavior.

AI assistance was used to prepare this change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5492,6 +5492,7 @@ dependencies = [
  "semver 1.0.28",
  "serde",
  "serde_json",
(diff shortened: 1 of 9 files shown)


mattsse (approved):
LGTM

mablr (approved):

PR #15494, opened 2026-06-30
perf(symbolic): skip overwritten memory byte history

figtracer:
## Summary

- skip memory byte history that is definitely overwritten by a later concrete write
- reuse the optimized concrete-byte lookup from symbolic dynamic memory reads
- add coverage for later symbolic writes still overriding a concrete base write


mattsse (approved):
LGTM

stevencartavia (approved):

mablr (approved):

PR #15495, opened 2026-06-30
perf(symbolic): avoid needless call snapshots

figtracer:
Avoid cloning the full symbolic `PathState` before every concrete-target CALL when no symbolic function mocks, expected calls, or call mocks are installed. The pre-call snapshot is still taken for mock/expect-call paths because those branches need to rewind to the pre-CALL stack state.

```text
master:  228.1 ms ± 6.5 ms wall, 691.3 ms user CPU
this PR:  75.6 ms ± 1.5 ms wall, 112.6 ms user CPU
speedup:  3.02x ± 0.10 wall
```


mablr (approved):
sgtm

DaniPopes (approved):

PR #15496, opened 2026-06-30
fix(ci): benchmarks for 1.8 release

mablr:
## Description

- Skips the known failing Spark invariant in coverage benchmarks to match the existing test benchmark exclusions.
- Temporarily pins benchmark isolate mode explicitly so only `forge_isolate_test` runs isolated and the other benchmarks run non-isolated (as isolate mode becomes default on v1.8.0)
- Temporarily disables symbolic benchmarks because the v1.7.1 baseline does not support native symbolic execution.


mablr (commented):
Do not merge before this is green https://github.com/foundry-rs/foundry/actions/runs/28464593139.

mattsse (approved):
LGTM

figtracer (approved):

stevencartavia (approved):

PR #15497, opened 2026-06-30
fix(debugger): show command aliases in prompt help

mablr:
## Description

This follows up on #15485 by polishing the debugger command prompt after it landed.

The command aliases now live in shared groups used by both command dispatch and help rendering, so aliases like `:cont`, `:c`, `:p`, `:cd`, `:ret`, and `:rd` stay documented with the parser. Long prompt input now renders as a single-line tail view when space is tight, keeping the cursor visible instead of wrapping inside the prompt block.


mattsse (approved):
LGTM

figtracer (approved):

stevencartavia (approved):

PR #15498, opened 2026-06-30
feat: default to Rust foundryup

stevencartavia:
Migrates `foundryup` to the [Rust rewrite](https://github.com/foundry-rs/foundryup). The bash script is now a thin bootstrap: on first run it installs the pinned Rust foundryup (v0.0.6) and replaces itself with the binary, so all subsequent runs use the Rust foundryup directly.


mablr (commented):
With migration now enabled by default, a first-run `foundryup --version` bootstraps the sidecar and writes all of the bootstrap/status output to stdout before the version line, including instructions to add the internal sidecar path to `PATH`. I reproduced this with a fresh `FOUNDRY_DIR` stdout contained the migration/install messages and only then foundryup 0.0.5 .... We should probably redirect the bootstrap/install chatter to stderr.

mattsse (commented):
Thanks for working on the migration. I found one behavior regression that should be fixed before flipping the default.

`migrate_and_exec` now calls `ensure_rust_foundryup` for every non-`--update` invocation before execing the Rust sidecar. That makes previously local/offline commands like `foundryup --help`, `foundryup --version`, `foundryup --list`, and `foundryup --use <installed-version>` require a successful first-time sidecar bootstrap. If the user is offline, GitHub is unavailable, or the bootstrap download fails, those commands fail before they can print local help/version info or inspect already-installed versions. The Rust binary itself treats `--list`/`--use` as offline paths, but the shim prevents reaching those paths unless the sidecar already exists.

A fix would be to keep truly local commands on a local path until the sidecar exists. For example, handle `-h|--help` and `-v|--version` directly in the shim, and either route `-l|--list` / `-u|--use` through the legacy Bash implementation when `rust_foundryup_bin` is missing, or only attempt bootstrap for those commands after confirming the sidecar is already installed. Please also add a top-level shim test that stubs `install_rust_foundryup` to fail and verifies the offline commands still succeed or use the legacy path instead of trying to download.

foundryup/foundryup
```diff
@@ -1002,19 +1069,124 @@ ensure_rust_foundryup() {
...
+shim_action() {
+  while [ "$#" -gt 0 ]; do
+    case "$1" in
+      --) break ;;
+      -U|--update)  printf 'update\n';  return ;;
+      -h|--help)    printf 'help\n';    return ;;
+      -v|--version) printf 'version\n'; return ;;
+      -l|--list|-u|--use) printf 'local\n'; return ;;
+      -r|--repo|-b|--branch|-i|--install|-p|--path|-P|--pr|\
+      -C|--commit|-j|--jobs|-n|--network|--arch|--platform)
```
mablr: `shim_action` keeps scanning after unknown arguments, while `main` errors immediately on the first unknown option. That changes behavior for inputs like `foundryup --bogus --version` and `foundryup --bogus --update`: the shim handles the later terminal flag instead of letting the sidecar or legacy parser reject `--bogus`. Consider adding a catch-all case for unknown args.

foundryup/foundryup
```diff
@@ -1002,19 +1069,124 @@ ensure_rust_foundryup() {
...
+      version
+      ;;
+    local)
+      # --list/--use must not bootstrap the Rust sidecar. If it is already
+      # installed, exec it (the Rust binary owns these); otherwise serve them
+      # from the legacy bash path without attempting to download the sidecar.
+      local _local_bin
+      _local_bin="$(rust_foundryup_bin)"
+      if [ -x "$_local_bin" ]; then
+        exec "$_local_bin" "$@"
```
mablr: The local `--list`/`--use` path execs the sidecar whenever the file has the executable bit set. If that sidecar is corrupt or anything, `exec` fails and the process exits instead of falling back to the legacy bash implementation. Consider probing `sidecar_version "$_local_bin"` here and only execing when it returns a valid version; otherwise call `main "$@"` without bootstrapping.

stevencartavia: [8c59440](https://github.com/foundry-rs/foundry/pull/15498/commits/8c5944011a77b73b76670fab40f6f28649a3086a)

foundryup/foundryup
stevencartavia: [8c59440](https://github.com/foundry-rs/foundry/pull/15498/commits/8c5944011a77b73b76670fab40f6f28649a3086a)

mablr (dismissed):
Invalid options may still trigger a Rust sidecar bootstrap before failing, but it's a narrow corner case. It's good to go as-is.

foundryup/foundryup
```diff
@@ -1002,19 +1073,132 @@ ensure_rust_foundryup() {
...
+      # --list/--use must not bootstrap the Rust sidecar. If a working sidecar is
+      # already installed, exec it (the Rust binary owns these); otherwise serve
+      # them from the legacy bash path without attempting to download the
+      # sidecar. Probe `sidecar_version` rather than the executable bit alone so
+      # a corrupt sidecar (where `exec` would fail and kill the process) falls
+      # back to `main` instead.
+      local _local_bin
+      _local_bin="$(rust_foundryup_bin)"
+      if [ -n "$(sidecar_version "$_local_bin")" ]; then
+        exec "$_local_bin" "$@"
```
figtracer: This still passes trailing args to the sidecar. Since v0.0.5 checks `--update` before `--list`/`--use`, `foundryup --list --update` can update instead of list. Please strip args after the first terminal command or route mixed terminal flags through `main`.

foundryup/foundryup
```diff
@@ -1002,19 +1073,132 @@ ensure_rust_foundryup() {
...
+}
+
+# Bootstraps the Rust foundryup sidecar and execs it. `--update` is handled here
+# (update this launcher, then the sidecar), never by the Rust binary itself.
+migrate_and_exec() {
+  case "$(shim_action "$@")" in
+    update)
+      update_shim "$@"
+      if ensure_rust_foundryup; then
+        say "foundryup is up to date" >&2
```
figtracer: If `update_shim` fails to fetch or parse the launcher but `ensure_rust_foundryup` succeeds, this prints `foundryup is up to date` and exits 0. Please carry the launcher-update failure so `--update` does not report success after a failed launcher refresh.

foundryup/foundryup
```diff
@@ -499,6 +533,16 @@ list() {
   exit 0
 }
 
+# Prints the install dir for version $1, trying the flat layout then the nested
+# foundry-rs/foundry layout. Prints nothing if it is not installed.
+resolve_version_dir() {
+  local _flat _nested
+  _flat="$FOUNDRY_VERSIONS_DIR/$1"
+  if [ -d "$_flat" ]; then printf '%s\n' "$_flat"; return 0; fi
+  _nested="$FOUNDRY_VERSIONS_DIR/foundry-rs/foundry/$1"
```
figtracer: This only resolves `foundry-rs/foundry`, but `list_versions_dir` shows any `versions/<owner>/<repo>/<version>`. In fallback mode a custom nested version can be listed but `--use vX` fails. Please resolve a unique nested match or error on ambiguity.

foundryup/foundryup
```diff
@@ -1002,19 +1073,132 @@ ensure_rust_foundryup() {
...
+    local)
+      # --list/--use must not bootstrap the Rust sidecar. If a working sidecar is
+      # already installed, exec it (the Rust binary owns these); otherwise serve
+      # them from the legacy bash path without attempting to download the
+      # sidecar. Probe `sidecar_version` rather than the executable bit alone so
+      # a corrupt sidecar (where `exec` would fail and kill the process) falls
+      # back to `main` instead.
+      local _local_bin
+      _local_bin="$(rust_foundryup_bin)"
+      if [ -n "$(sidecar_version "$_local_bin")" ]; then
```
figtracer: This sidecar path bypasses the bash `--use` cleanup that removes both bare and `.exe` names. Rust foundryup v0.0.5 removes only the `.exe` path on Windows, so a stale bare `forge` can remain. Please route Windows `--use` through bash or bump the sidecar.

figtracer (changes requested):

DaniPopes:
why doesnt migration just replace `foundryup` script with the rust executable directly? i don't think we want to allow going back to the bash script, it's not that big of a migration

foundryup/foundryup
```diff
@@ -1,1020 +1,135 @@
...
+# corrupts an existing foundryup.
 # ----------------------------------------------------------------------------
 
-# Pinned Rust foundryup installer version to bootstrap from.
+BASE_DIR="${XDG_CONFIG_HOME:-$HOME}"
+FOUNDRY_DIR="${FOUNDRY_DIR:-"$BASE_DIR/.foundry"}"
+FOUNDRY_BIN_DIR="$FOUNDRY_DIR/bin"
+
+# Pinned Rust foundryup version to bootstrap. Override for testing/emergencies.
 FOUNDRYUP_BOOTSTRAP_VERSION="${FOUNDRYUP_BOOTSTRAP_VERSION:-0.0.5}"
```
mablr: This pins the migration to Rust foundryup v0.0.5, which still has the Windows activation cleanup gap: it removes the `.exe` activation target but can leave stale extensionless binaries from prior bash installs. Maybe we should bump the pin to a release that removes both names before defaulting everyone to this path?

mablr (commented):
LGTM overall, just one question.

foundryup/foundryup
stevencartavia: Bumped the pin to v0.0.6, which fixes the Windows activation cleanup gap, done in the latest push

mablr (dismissed):
LGTM

DaniPopes (dismissed):
once we switched the redirects, we should also deprecate the `foundryup/install` script in this repo

mattsse (dismissed):
LGTM

mablr:
@stevencartavia pls resolve conflicts

mablr (dismissed):

foundryup/foundryup
```diff
@@ -1,1102 +1,136 @@
...
+FOUNDRYUP_BOOTSTRAP_VERSION="${FOUNDRYUP_BOOTSTRAP_VERSION:-0.0.6}"
+# `foundryup-init.sh` for the pinned bootstrap version.
+FOUNDRYUP_INIT_URL="${FOUNDRYUP_INIT_URL:-"https://raw.githubusercontent.com/foundry-rs/foundryup/v${FOUNDRYUP_BOOTSTRAP_VERSION}/foundryup-init.sh"}"
 
-say() {
-  printf "foundryup: %s\n" "$1"
-}
+# The Rust installer writes `bin/foundryup` (no `.exe`, even on Windows), so the
+# bootstrap replaces itself at the same extensionless path on every platform.
+FOUNDRY_BIN_PATH="$FOUNDRY_BIN_DIR/foundryup"
```
figtracer: this only replaces the launcher when it lives under the active `$FOUNDRY_DIR`. with `FOUNDRY_DIR=/tmp/custom foundryup --version`, i get a rust binary under `/tmp/custom` but the launcher on `$PATH` stays bash, so every later call bootstraps again and fails offline. could we replace the resolved launcher path while still passing the override to rust?

stevencartavia: [3a2d7e6](https://github.com/foundry-rs/foundry/pull/15498/commits/3a2d7e67739380e5268915c8ce26496f84e616af)

DaniPopes (approved):

figtracer (approved):

PR #15500, opened 2026-06-30
perf(symbolic): centralize test matching

DaniPopes:
Centralizes symbolic test matching so `check*` and `prove*` classification is handled by `TestFunctionKind` and `TestFunctionMatcher` instead of local special cases. Inline config is parsed once during test setup and passed into coverage, mutation, and runner construction, with direct lookup helpers for contract-level symbolic and network overrides.

This improves the matching path by avoiding repeated inline config parsing and by reading the few needed inline values directly from the parsed data instead of round-tripping through Figment serde just to inspect symbolic and network settings. It keeps source filtering, test listing, runner build, and execution on the same function-kind path while preserving default-profile fallback behavior.

Prepared with AI assistance.

Before:
image

After:
image

### DIFF
--- a/crates/common/src/traits.rs
+++ b/crates/common/src/traits.rs
@@ -19,15 +19,32 @@ pub trait TestFilter: Send + Sync {
     /// Returns whether the test should be included for the given contract.
     ///
     /// `contract_id` is the full artifact identifier (`path:Contract`).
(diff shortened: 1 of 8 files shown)


mattsse (approved):
LGTM

figtracer (approved):

PR #15501, opened 2026-06-30
feat(config): support solc --via-ssa-cfg

Amxx:
This mirrors #15177 (commit f4ecc6a46599fc6e6ac23e403b69c21cdd11eea5), which added support for solc's `--experimental` flag, applying the same pattern to expose solc's `viaSSACFG` setting. It adds a `--via-ssa-cfg` CLI flag and a `via_ssa_cfg` config option that turn on SSA CFG-based code generation via the IR (experimental, implies `viaIR`), false by default.

The Foundry-side wiring is in place: the `CompilerOpts` flag, the `Config` field and its default, the `solc_settings()` plumbing, and the corresponding parse/config/snapshot tests. As with #15177, this depends on the setting being exposed on `foundry-compilers`' solc `Settings` struct, which the currently pinned revision does not yet expose, so the config crate does not build on its own. I'm working on the required changes in foundry-core and will follow up here with the dependency bump, the same way #15177 pinned a foundry-core revision. Opening as a draft until that lands.

This change was prepared with AI assistance.

Updated: now ready thanks to https://github.com/foundry-rs/foundry-core/pull/104

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5375,7 +5375,7 @@ dependencies = [
 [[package]]
 name = "foundry-compilers"
 version = "0.21.0"
-source = "git+https://github.com/foundry-rs/foundry-core?rev=39f19b016c2ba793b24255bda83bd3822c8cb525#39f19b016c2ba793b24255bda83bd3822c8cb525"
+source = "git+https://github.com/foundry-rs/foundry-core?rev=3e01874cb1e7f43d2554dc05fb1823687bdc3ab8#3e01874cb1e7f43d2554dc05fb1823687bdc3ab8"
 dependencies = [
  "alloy-json-abi",
  "alloy-primitives",
@@ -5385,7 +5385,7 @@ dependencies = [
  "foundry-compilers-artifacts",
  "foundry-compilers-core",
  "fs_extra",
- "itertools 0.14.0",
+ "itertools 0.13.0",
  "path-slash",
  "rand 0.9.4",
  "rayon",
@@ -5406,7 +5406,7 @@ dependencies = [
 [[package]]
 name = "foundry-compilers-artifacts"
 version = "0.21.0"
-source = "git+https://github.com/foundry-rs/foundry-core?rev=39f19b016c2ba793b24255bda83bd3822c8cb525#39f19b016c2ba793b24255bda83bd3822c8cb525"
+source = "git+https://github.com/foundry-rs/foundry-core?rev=3e01874cb1e7f43d2554dc05fb1823687bdc3ab8#3e01874cb1e7f43d2554dc05fb1823687bdc3ab8"
 dependencies = [
  "foundry-compilers-artifacts-solc",
  "foundry-compilers-artifacts-vyper",
@@ -5415,7 +5415,7 @@ dependencies = [
 [[package]]
 name = "foundry-compilers-artifacts-solc"
 version = "0.21.0"
-source = "git+https://github.com/foundry-rs/foundry-core?rev=39f19b016c2ba793b24255bda83bd3822c8cb525#39f19b016c2ba793b24255bda83bd3822c8cb525"
+source = "git+https://github.com/foundry-rs/foundry-core?rev=3e01874cb1e7f43d2554dc05fb1823687bdc3ab8#3e01874cb1e7f43d2554dc05fb1823687bdc3ab8"
 dependencies = [
  "alloy-json-abi",
  "alloy-primitives",
@@ -5435,7 +5435,7 @@ dependencies = [
 [[package]]
 name = "foundry-compilers-artifacts-vyper"
 version = "0.21.0"
-source = "git+https://github.com/foundry-rs/foundry-core?rev=39f19b016c2ba793b24255bda83bd3822c8cb525#39f19b016c2ba793b24255bda83bd3822c8cb525"
+source = "git+https://github.com/foundry-rs/foundry-core?rev=3e01874cb1e7f43d2554dc05fb1823687bdc3ab8#3e01874cb1e7f43d2554dc05fb1823687bdc3ab8"
 dependencies = [
  "alloy-json-abi",
  "alloy-primitives",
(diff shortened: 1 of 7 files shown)


Cargo.lock
```diff
@@ -5400,7 +5400,7 @@ dependencies = [
  "foundry-compilers-artifacts",
  "foundry-compilers-core",
  "fs_extra",
- "itertools 0.14.0",
+ "itertools 0.13.0",
```
Amxx:
```suggestion
 "itertools 0.14.0",
```

DaniPopes (dismissed):

mablr (dismissed):
LGTM, thank you.

crates/cli/src/opts/build/core.rs
```diff
@@ -284,6 +284,10 @@ impl Provider for BuildOpts {
             dict.insert("experimental".to_string(), true.into());
         }
 
+        if self.compiler.via_ssa_cfg {
```
Amxx, line 287: I see that there is `self.via_ir` and `self.compiler.experimental`. Any reason `via_ir` is not in CompilerOpts ? Should `via_ssa_cfg` be at the same level as `via_ir`, or at the same level as `experimental` (in CompilerOpts) ?

mablr: Good point. I think `via_ir` being in `BuildOpts` is just historical. imo `via_ssa_cfg` is at the right place in `CompilerOpts`. Moving `via_ir` in a follow-up is fine.

crates/config/src/lib.rs
```diff
@@ -1888,6 +1892,7 @@ impl Config {
             model_checker,
             via_ir: Some(self.via_ir),
             experimental: Some(self.experimental),
+            via_ssa_cfg: Some(self.via_ssa_cfg),
```
stevencartavia, line 1896: `--via-ssa-cfg` / `via_ssa_cfg` doesn't work as documented. The doc says it implies `via_ir`, but `solc_settings()` emits an explicit `viaIR: false`, and solc's "viaSSACFG implies viaIR" only applies when `viaIR` is omitted, so it never fires. It also requires `experimental`, which isn't documented.

Built `forge` from this branch and ran `forge build` against a project pinned to `solc = "0.8.35"`:

```
--via-ssa-cfg                          -> Error: "settings.viaSSACFG" requires compilation via IR.
--via-ssa-cfg --via-ir                 -> Error: 'viaSSACFG' setting is experimental and can only be used with the 'settings.experimental' option enabled.
--via-ssa-cfg --experimental           -> Error: "settings.viaSSACFG" requires compilation via IR.
--via-ssa-cfg --via-ir --experimental  -> Compiler run successful!
```

Same via `foundry.toml`: `via_ssa_cfg = true` alone fails; only `via_ssa_cfg + via_ir + experimental` succeeds. Captured the exact standard JSON Foundry pipes to solc and it sends `{"viaIR": false, "viaSSACFG": true, "experimental": false}`.

Fix either by deriving the requirements in `solc_settings()`:

```rust
via_ir: Some(self.via_ir || self.via_ssa_cfg),
experimental: Some(self.experimental || self.via_ssa_cfg),
via_ssa_cfg: Some(self.via_ssa_cfg),
```

(note this makes `via_ssa_cfg` implicitly enable solc's global `experimental` mode — worth documenting), or keep the settings independent, drop the "implies via_ir" claim, document that `via_ir` + `experimental` are required, and validate early instead of forwarding invalid config to solc.

mablr: Yes it worths unifying all `via_ssa_cfg + via_ir + experimental` under `CompilerOpts`, as suggested here https://github.com/foundry-rs/foundry/pull/15501#discussion_r3507074229.

Amxx: [Here is a list](https://docs.soliditylang.org/en/latest/using-the-compiler.html#experimental-mode) of all the flags that are gated by experimental. Should they be handleded differently ? The PR that added "experimental" support to Foundry did not touch that.

mablr: cc @figtracer

Amxx: Updated:
```diff
-            via_ir: Some(self.via_ir),
+            // via_ssa_cfg implies via_ir only when via_ir is omitted, so we need to set both flags to true if via_ssa_cfg is enabled.
+            via_ir: Some(self.via_ir || self.via_ssa_cfg),
```

Amxx: IMO, the experimental tag is designed to protect users against inadvertently enabling an experimental feature in production. A user must "manually" turned it on if he wants to use any of the experimental features.

If foundry turns it on automatically, you lost the protection that its adding. So I would personnaly be against doing
```rust
experimental: Some(self.experimental || self.via_ssa_cfg || ... <all the other experimental flags>),
```

The question is then, is the error provided by the compiler good enough, of should foundry handle that before hitting the compiler?

mablr: I agree we should not derive `experimental` from `via_ssa_cfg`. `via_ir` seems to be a direct mechanical prerequisite of this setting, but `experimental` is a user consent gate, so auto-enabling it would make `--via-ssa-cfg` silently opt the user into global experimental mode.

I think the current behavior is acceptable if the docs/help clearly say `--experimental` is required. An early validation could be nicer UX, but I would not block this on it unless we want consistent pre-validation for all solc experimental features. wdyt @figtracer ?

Amxx: > I think the current behavior is acceptable if the docs/help clearly say --experimental is required.

Do you want me do that change ? If so, can you guide me to where that documentation is located ?

figtracer: i think we don't have a pre-validation layer for these cases so i would just document it properly; the experimental PR was not great in that aspect 👍

mablr: clap args comment should be fine imo

mablr: cool https://github.com/foundry-rs/foundry/pull/15501/commits/3dc7741743c7ccc3a3aec503218aea7ecacc3f46

Amxx:
@mablr what are the remaining blocked/next steps ?

DaniPopes (dismissed):

mablr (dismissed):
@Amxx Looks good, thank you. (just forgot to approve earlier)

mablr (dismissed):

DaniPopes (dismissed):

DaniPopes (approved):

mablr (approved):

PR #15502, opened 2026-06-30
perf(symbolic): avoid kind extraction

DaniPopes:
This refactors symbolic expression constructors to inspect borrowed expression kinds for peephole simplification instead of consuming expressions with `into_kind` and rebuilding the same nodes. `SymExpr::op` now delegates to a borrowed `peephole_op` helper, and the solver normalization/cache-key paths avoid unwrapping expression kinds after already matching on borrowed nodes.

The fallible concrete accessors are renamed to `as_const_or` and `as_usize_or` so callers no longer clone expressions just to surface unsupported symbolic values.

Prepared with AI assistance.


mattsse (approved):
LGTM

figtracer (approved):

PR #15503, opened 2026-06-30
perf(forge): cache symbolic test filter

mablr:
## Description

This avoids recomputing the contract-level symbolic test selection state for every ABI function while pre-filtering tests. The inline config merge is contract-scoped, so `matching_test_functions` and `list` now compute the effective symbolic flag once per matched contract and reuse it for that contract's functions.

This matters for large suites because `forge test` counts matching tests before streaming execution. On Solady, that serial pre-count path was taking roughly 800ms on master before test execution started.

Solady benchmark matrix, `Vectorized/solady@ab96a83`, warm cache, dist binaries, `FOUNDRY_ISOLATE=false FOUNDRY_DYNAMIC_TEST_LINKING=false`:

| Command | stable v1.7.1 | master `21ef7b829` | this PR `30978c0fd` |
| --- | ---: | ---: | ---: |
| `forge test` | `1.360s ± 0.032s` | `1.996s ± 0.078s` | `1.227s ± 0.043s` |
| `forge test --threads 8` | `1.531s ± 0.018s` | `2.131s ± 0.037s` | `1.381s ± 0.034s` |

The PR recovers the master regression introduced by #15227's nit fix https://github.com/foundry-rs/foundry/pull/15227/commits/8519a95a0e733d2c8f9668b081eec304078409e1 (see https://github.com/foundry-rs/foundry/pull/15227#discussion_r3477503220) and is slightly faster than stable in this Solady run.

### DIFF
--- a/crates/forge/src/multi_runner.rs
+++ b/crates/forge/src/multi_runner.rs
@@ -125,9 +125,15 @@ impl<FEN: FoundryEvmNetwork> MultiContractRunner<FEN> {
         let matcher = self.test_function_matcher();
         self.matching_contracts(filter).flat_map(move |(id, c)| {
             let identifier = id.identifier();
-            c.abi
-                .functions()
-                .filter(move |func| matcher.matches_test_function(filter, &identifier, func))
+            let symbolic_tests_enabled = matcher.symbolic_tests_enabled(&identifier);
+            c.abi.functions().filter(move |func| {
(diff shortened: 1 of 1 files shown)


mattsse (approved):
LGTM

mablr (commented):
Original fix 30978c0fd was right but #15500 superseeded it, problem is already solved.

PR #15506, opened 2026-07-01
perf(anvil): filter txpool_contentFrom before conversion

stevencartavia:
`txpool_contentFrom` built the whole txpool then dropped all but the requested sender. This filters by sender first, so only that sender's transactions get converted.


mattsse (approved):
LGTM

mablr (approved):

PR #15507, opened 2026-07-01
fix(script): invalidate fork cache after state-mutating vm.rpc

### LINKED ISSUE #15207: `forge script --broadcast` replay ignores Anvil state changes made via `vm.rpc` earlier in the same script
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.7.1

foundryup: 1.8.5

forge script

### Operating System

macOS (Apple Silicon)

## Summary

When a Forge script mutates Anvil state via `vm.rpc("anvil_setBalance", ...)` before a `vm.startBroadcast()` transaction, the script simulation sees the updated balance, but Forge's later broadcast validation/replay appears to use the RPC state from before the script started.

This causes `forge script --broadcast` to fail with `lack of funds`, even though the account was funded on Anvil earlier in the same script.

A second run can pass, because the first failed run still mutated Anvil state via `anvil_setBalance`, so the updated balance exists before the second script invocation starts.

## Why this matters

This affects Anvil/fork verification scripts that need to prepare local fork state before broadcasting transactions from impersonated/unlocked accounts.

In my real case, a verify script funded an impersonated multisig on Anvil, then broadcasted a transaction from that multisig. The local script simulation succeeded, but Forge's broadcast validation failed because it appeared to validate against the account's original Anvil balance.

### 1. Start Anvil

```bash
anvil --host 127.0.0.1 --port 8557 --balance 10000
```

The default account starts with `10000 ETH`.

### 2. Add this script

`script/StartBroadcastRepro.s.sol`

```solidity
pragma solidity ^0.8.24;

import {Script, console2} from "forge-std/Script.sol";

contract StartBroadcastRepro is Script {
    address constant SENDER = 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266;
    address constant RECIPIENT = 0x000000000000000000000000000000000000dEaD;

    function run() external {
        uint256 fundedAmount = 20_001 ether;
        uint256 sendAmount = 20_000 ether;

        console2.log("sender before:", SENDER.balance);

        vm.deal(SENDER, fundedAmount);

        vm.rpc(
            "anvil_setBalance",
            string.concat(
                "[\"",
                vm.toString(SENDER),
                "\",\"",
                vm.toString(fundedAmount),
                "\"]"
            )
        );

        console2.log("sender after local/rpc funding:", SENDER.balance);

        vm.startBroadcast(SENDER);

        (bool ok,) = RECIPIENT.call{value: sendAmount}("");
        require(ok, "send failed");

        vm.stopBroadcast();

        console2.log("sender after send:", SENDER.balance);
        console2.log("recipient after send:", RECIPIENT.balance);
    }
}
```

### 3. Run with broadcast

```bash
forge script script/StartBroadcastRepro.s.sol:StartBroadcastRepro \
  --rpc-url http://127.0.0.1:8557 \
  --sender 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 \
  --broadcast \
  --unlocked \
  -vv
```

## Actual result

The script logs show the sender balance being updated before the broadcasted call, but Forge later fails during broadcast validation/replay:

```text
## Setting up 1 EVM.
Error: Internal EVM error during simulation

Context:
- transaction validation error: lack of funds (10000000000000000000000) for max fee (20000000000000000000000)
```

The failing balance is the original Anvil balance, `10000 ETH`, not the balance set inside the script.

## Expected result

Forge's broadcast validation/replay should either:

1. Observe the Anvil state changes made earlier in the same script via `vm.rpc`, or
2. Clearly document that broadcast replay is based on a pre-script RPC snapshot and cannot depend on same-script RPC state mutations.

Ideally, if `vm.rpc("anvil_setBalance", ...)` mutates the connected Anvil node before `vm.startBroadcast`, a later broadcast transaction from that account should validate against the updated remote state.

## Additional observations

The following did not fix the issue:

```bash
--slow
--isolate
--no-storage-caching
```

Calling `evm_mine` after `anvil_setBalance` also did not fix it.

Using `vm.broadcast(SENDER)` instead of `vm.startBroadcast(SENDER)` produced the same behavior.

Creating/selecting a fresh fork after `anvil_setBalance` can make the local script simulation see the updated state, but Forge's later broadcast validation still appears to use the pre-script state.

A workaround is to bypass Forge's broadcast queue and send the Anvil transaction directly:

```solidity
vm.rpc("eth_sendTransaction", "[{\"from\":\"0x...\",\"to\":\"0x...\",\"value\":\"0x...\"}]");
```

Then refresh local state with `vm.createSelectFork(...)`.

That works, but it means the transaction is not part of Forge's normal broadcast artifact/replay pipeline.

## Question

Is this expected behavior for `forge script --broadcast`, or should broadcast validation account for RPC state mutations made earlier in the same script invocation?

stevencartavia:
Fixes #15207. 

`vm.rpc("anvil_setBalance", ...)` mutates the forked node directly but not Foundry's fork cache, so `forge script --broadcast` replayed the pre-broadcast simulation against stale state and failed with "lack of funds". Now a state-mutating active-fork `vm.rpc`/`vm.rpcJson` evicts the affected account/slot from the fork cache so the simulation re-fetches the new value.

### DIFF
--- a/crates/cheatcodes/src/evm/fork.rs
+++ b/crates/cheatcodes/src/evm/fork.rs
@@ -5,7 +5,7 @@ use crate::{
 use alloy_dyn_abi::DynSolValue;
 use alloy_evm::EvmEnv;
 use alloy_network::AnyNetwork;
-use alloy_primitives::{B256, U256};
+use alloy_primitives::{Address, B256, U256};
 use alloy_provider::Provider;
 use alloy_rpc_types::Filter;
 use alloy_sol_types::SolValue;
@@ -216,7 +216,9 @@ impl Cheatcode for rpc_0Call {
         let Self { method, params } = self;
         let url =
             ccx.ecx.db().active_fork_url().ok_or_else(|| fmt_err!("no active fork URL found"))?;
-        rpc_call(&url, method, params)
+        let result = rpc_call(&url, method, params)?;
+        invalidate_active_fork_cache(ccx, method, params);
+        Ok(result)
     }
 }
 
@@ -233,7 +235,9 @@ impl Cheatcode for rpcJson_0Call {
         let Self { method, params } = self;
         let url =
             ccx.ecx.db().active_fork_url().ok_or_else(|| fmt_err!("no active fork URL found"))?;
-        rpc_json_call(&url, method, params)
+        let result = rpc_json_call(&url, method, params)?;
+        invalidate_active_fork_cache(ccx, method, params);
+        Ok(result)
     }
 }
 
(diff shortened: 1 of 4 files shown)


mattsse (approved):
LGTM

mablr (approved):

frontier159:
oh thats awesome thanks @mablr @stevencartavia

mablr:
s/o to @stevencartavia 🫡

PR #15510, opened 2026-07-01
refactor: use tempo-hardfork directly

mattsse:
This replaces Foundry's direct dependency on `tempo-chainspec` with the lighter `tempo-hardfork` crate for `TempoHardfork` usage. The direct imports in Anvil, Cast tests, cheatcodes, EVM core, and Foundry's hardfork re-export now use `tempo_hardfork`, while Anvil's base-fee constants are read from `tempo_hardfork::constants::gas`.

This keeps the same pinned Tempo revision and removes the workspace-level direct `tempo-chainspec` dependency. `tempo-chainspec` can still appear transitively through other Tempo crates, but Foundry no longer imports or declares it directly for hardfork handling.

AI assistance was used for this change.


grandizzy (approved):
lgtm!

mablr (approved):
cleaner 👍

PR #15511, opened 2026-07-01
feat(debugger): add storage command

figtracer:
This adds `:storage`, `:store`, and `:slot` debugger commands that jump to the next matching `SLOAD` or `SSTORE` for a storage slot in the current call. The command path reuses the same private storage-access helper as the Variables panel, so recorded storage changes and warm `SLOAD` stack snapshots are interpreted and formatted consistently.

Refs #927

### DIFF
--- a/crates/debugger/src/tui/context.rs
+++ b/crates/debugger/src/tui/context.rs
@@ -1,6 +1,6 @@
 //! Debugger context and event handler implementation.
 
-use super::storage::{find_storage_access, hex_u256};
+use super::storage::{StorageAccess, hex_u256, storage_access_at};
 use crate::{DebugNode, DebuggerLayout, ExitReason, debugger::DebuggerContext};
 use alloy_primitives::{Address, U256, hex};
 use crossterm::event::{Event, KeyCode, KeyEvent, KeyModifiers, MouseEvent, MouseEventKind};
@@ -678,13 +678,23 @@ impl TUIContext<'_> {
             }
         };
 
-        let Some(access) = find_storage_access(self.debug_steps(), self.current_step, slot) else {
+        let Some(target) = find_storage_target(
+            self.debug_arena(),
+            self.draw_memory.inner_call_index,
+            self.current_step,
+            slot,
+        ) else {
             self.set_error(format!("Storage slot {} not accessed in current call", hex_u256(slot)));
             return;
         };
 
+        let access = target.access;
+        self.draw_memory.inner_call_index = target.node_index;
         self.current_step = access.step_index();
+        self.draw_memory.current_buf_startline = 0;
+        self.draw_memory.current_stack_startline = 0;
         self.scroll_memory_to_current_write();
+        self.key_buffer.clear();
         self.set_info(format!(
             "Jumped to {} at PC 0x{:x} ({})",
             access.describe(),
@@ -980,6 +990,12 @@ struct PcTarget {
     scope: PcTargetScope,
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct StorageTarget {
+    node_index: usize,
+    access: StorageAccess,
+}
+
 fn parse_pc_candidates(input: &str) -> Result<Vec<PcCandidate>, String> {
     let input = input.trim();
     if input.is_empty() {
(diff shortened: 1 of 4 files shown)


crates/debugger/src/tui/context.rs
```diff
@@ -656,6 +659,40 @@ impl TUIContext<'_> {
...
+            Err(err) => {
+                self.set_error(err);
+                return;
+            }
+        };
+
+        let Some(access) = find_storage_access(self.debug_steps(), self.current_step, slot) else {
+            self.set_error(format!("Storage slot {} not accessed in current call", hex_u256(slot)));
+            return;
+        };
```
mablr: `self.debug_steps()` only covers the currently selected `DebugNode`, but debugger nodes are flattened segments of an original EVM call split around child calls. That means `:storage <slot>` from the pre-child segment will miss a matching access in the post-child segment even though it is still the same original call (`trace_node_idx`). Consider searching all arena nodes with the current `trace_node_idx`, ordered by `step_offset + local_step_index`, and returning both `node_index` and `step_index` so the jump updates `draw_memory.inner_call_index` too.

crates/debugger/src/tui/storage.rs
```diff
@@ -0,0 +1,107 @@
...
+    if step.op.get() == opcode::SLOAD {
+        return Some(StorageAccess {
+            step_index,
+            pc: step.pc,
+            kind: StorageAccessKind::Sload,
+            slot: step.stack.as_deref()?.last().copied()?,
+            value: steps.get(step_index.checked_add(1)?)?.stack.as_deref()?.last().copied()?,
+            previous: None,
+        });
+    }
```
mablr, line 87: This fallback only reconstructs missing `storage_change` data for `SLOAD`. `revm-inspectors` only attaches `storage_change` when the journal length changes, so a warm/no-op `SSTORE` can execute without a `storage_change` and will be invisible to both the Variables panel and `:storage`. Consider adding an `SSTORE` stack fallback as well, using the pre-op stack’s top item as the slot/key and the next item as the value, with `previous: None`.

mablr (dismissed):
sgtm

crates/debugger/src/tui/storage.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    if step.op.get() == opcode::SSTORE {
```
stevencartavia, line 89: The warm-SSTORE fallback in `storage_access_at` treats an SSTORE with `storage_change == None` as a write by reading `[slot, value]` off the stack, so it misreports SSTOREs that halted without writing (e.g. SSTORE inside a STATICCALL → `StateChangeDuringStaticCall`).

Repro with a real trace: a test doing `address.staticcall(abi.encodeWithSignature("setValue(uint256)", 0x2a))`, dumped via `forge test --debug --mt test_static_sstore --dump dump.json`. The step is `op=SSTORE`, `storage_change=None`, `status=Some(StateChangeDuringStaticCall)`, stack present — and `storage_access_at` returns `Some("storage SSTORE slot 0x0 = 0x2a")`, a write that never happened. Affects both the Variables pane and the `:storage`/`:slot`/`:store` command.

crates/debugger/src/tui/context.rs
```diff
@@ -1017,6 +1072,131 @@ fn invalid_buffer_offset(input: &str) -> String {
     format!("Invalid buffer offset `{input}`; use hex 0x20/20 or decimal d:32")
 }
 
+fn parse_storage_slot(input: &str) -> Result<U256, String> {
```
stevencartavia, line 1075: `parse_storage_slot` accepts underscores (via `U256::from_str_radix`) while `parse_buffer_offset` rejects them (via `usize::from_str_radix`). So `storage 1_0` parses as `0x10`, and worse, `storage _` / `0x_` / `d:_` silently parse as slot `0` and jump, instead of erroring like `storage zz` does.

crates/debugger/src/tui/storage.rs
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+}
+
+pub(super) fn storage_access_at(
+    steps: &[CallTraceStep],
+    step_index: usize,
+) -> Option<StorageAccess> {
+    let step = steps.get(step_index)?;
+    if let Some(change) = step.storage_change.as_deref() {
```
stevencartavia, line 63: The guard added for halted SSTOREs only covers the stack fallback, but the `storage_change` branch runs first and isn't checked, so a halted SSTORE that already journaled still leaks.

Repro: `target.setValue{gas: 15000}(0x2a)` — enough gas to journal the cold SSTORE but not its dynamic gas, so it OOGs after the write is recorded. Dumped via `forge test --debug --dump`, the step is `storage_change=Some(...)`, `status=OutOfGas`, and `storage_access_at` returns `Some("storage SSTORE slot 0x0: 0x0 -> 0x2a")` — a write that was rolled back.

The status check needs to move above the `storage_change` branch:

```rust
let step = steps.get(step_index)?;

if step.op.get() == opcode::SSTORE && !step.status.is_none_or(InstructionResult::is_ok) {
    return None;
}

if let Some(change) = step.storage_change.as_deref() { ... }
```

stevencartavia (approved):

mablr (approved):

PR #15513, opened 2026-07-01
feat(fuzz): capture branch frontier artifacts

figtracer:
This adds opt-in fuzz branch frontier capture so fuzz runs can hand concrete comparison sites to later targeted symbolic execution work. When `fuzz.frontier_dir` is configured, Forge writes a bounded per-test `branch-frontiers.json` artifact under `<frontier_dir>/<contract>/<test>/`, including replay metadata, the concrete one-call sequence, the EVM comparison site, operands, comparison result, operand delta, and whether that run expanded coverage.

The implementation keeps capture off by default and bounded by `fuzz.frontier_limit`, documents the `foundry:fuzz.branch-frontiers@v1` format in the symbolic README, and includes config coverage plus a Forge integration test that models a missed `feeMultiplier < 100` arithmetic bug branch.

### Concrete example

The in-tree regression uses a Velocore-style `feeMultiplier - 100` underflow shape. A short fuzz run passes, but the frontier artifact records an adjacent `LT` comparison at the `feeMultiplier < 100` guard:

```text
opcode: LT
condition: feeMultiplier < 100
observed operands: lhs = 0x64, rhs = 0x64
observed result: false
replay calldata decodes to: amount = 100, feeMultiplier = 100
```

That is the intended handoff: fuzzing reached the boundary of a hard branch, and a later symbolic pass can replay that call and solve for the neighboring `feeMultiplier < 100` path.

I also smoke-tested this against `grandizzy/symbolic-bug-suite` with a temporary Velocore-style fuzz adapter. The short frontier run produced `LT` frontier artifacts, a longer 256-run fuzz pass found a concrete failing side (`args=[15, 0]`), and the existing symbolic Velocore test found a witness with z3 (`checkFeeRebateCannotExceedAmount`, `args=[0, 0]`).

### Review follow-up

This also addresses the first review pass:

- `forge clean` now removes configured `fuzz.frontier_dir` artifacts.
- `--fuzz-frontier-dir` still records EVM comparison frontiers when `sancov_edges = true`; this uses a cmp-only EVM inspector so frontier capture does not add EVM edge hits to sancov-guided campaigns.
- `operand_delta` now uses signed distance for `SLT`/`SGT` and unsigned distance for unsigned comparisons.

Local checks:
- `cargo +nightly fmt --all`
- `git diff --check`
- `cargo +1.95.0 check -p foundry-evm`
- `cargo +1.95.0 build -p forge`
- `cargo +1.95.0 test -p foundry-evm cmp_log_only_does_not_collect_edges -- --nocapture`
- `cargo +1.95.0 test -p foundry-evm operand_delta_uses_signed_distance_for_signed_comparisons -- --nocapture`
- `cargo +1.95.0 test -p foundry-evm signed_operand_delta_handles_full_int256_range -- --nocapture`
- `cargo +1.95.0 test -p forge fuzz_branch_frontiers_capture_comparison_for_symbolic_followup -- --nocapture`
- `cargo +1.95.0 test -p forge can_clean_test_cache -- --nocapture`
- `cargo +nightly clippy -p foundry-evm --all-targets --all-features --locked`
- `cargo +nightly clippy -p forge --all-targets --all-features --locked`
- `grandizzy/symbolic-bug-suite`: short frontier smoke test, 256-run fuzz failure check, `forge test --symbolic --match-test checkFeeRebateCannotExceedAmount`, and full `forge test --symbolic` expected-failure suite exit.

### DIFF
--- a/crates/config/src/fuzz.rs
+++ b/crates/config/src/fuzz.rs
@@ -186,14 +186,15 @@ impl FuzzCorpusConfig {
 
     /// Whether EVM comparison operand capture is enabled.
     ///
-    /// EVM comparison operands are only useful for coverage-guided fuzzing, so they are derived
-    /// from corpus mode. Disabled when sancov edge coverage is active because sancov replaces EVM
-    /// bytecode coverage as the guidance signal.
+    /// EVM comparison operands for corpus mutation are only useful for coverage-guided fuzzing, so
+    /// they are derived from corpus mode and disabled when sancov edge coverage is active.
+    /// Frontier capture still records EVM comparison sites because those artifacts target
+    /// Solidity bytecode branches, not the active coverage guidance source.
     pub fn collect_evm_cmp_log(&self) -> bool {
-        !self.sancov_edges
-            && ((self.corpus_dir.is_some()
+        self.capture_branch_frontiers()
+            || (!self.sancov_edges
+                && self.corpus_dir.is_some()
                 && self.mutation_weights.effective().mutation_weight_cmp > 0)
-                || self.capture_branch_frontiers())
     }
 
     /// Whether fuzz branch frontier artifacts should be captured.
--- a/crates/config/src/fuzz.rs
+++ b/crates/config/src/fuzz.rs
@@ -119,6 +119,10 @@ pub struct FuzzCorpusConfig {
     // Path to corpus directory, enabled coverage guided fuzzing mode.
     // If not set then sequences producing new coverage are not persisted and mutated.
     pub corpus_dir: Option<PathBuf>,
+    // Path to fuzz branch frontier artifacts for symbolic follow-up.
+    pub frontier_dir: Option<PathBuf>,
+    // Maximum number of branch frontier records to write for one fuzz test.
+    pub frontier_limit: usize,
     // Whether corpus to use gzip file compression and decompression.
     pub corpus_gzip: bool,
     // Number of mutations until entry marked as eligible to be flushed from in-memory corpus.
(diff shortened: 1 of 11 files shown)


crates/config/src/fuzz.rs
```diff
@@ -183,8 +191,14 @@ impl FuzzCorpusConfig {
     /// bytecode coverage as the guidance signal.
     pub fn collect_evm_cmp_log(&self) -> bool {
         !self.sancov_edges
-            && self.corpus_dir.is_some()
-            && self.mutation_weights.effective().mutation_weight_cmp > 0
+            && ((self.corpus_dir.is_some()
+                && self.mutation_weights.effective().mutation_weight_cmp > 0)
+                || self.capture_branch_frontiers())
+    }
```
mablr: It seems that `--fuzz-frontier-dir` still silently produces no EVM cmp frontiers when `sancov_edges` is enabled, is this expected behavior ?

crates/evm/evm/src/executors/fuzz/frontier.rs
```diff
@@ -0,0 +1,268 @@
...
+        opcode::SLT => I256::from_raw(cmp.op1) < I256::from_raw(cmp.op2),
+        opcode::SGT => I256::from_raw(cmp.op1) > I256::from_raw(cmp.op2),
+        opcode::ISZERO => cmp.op1.is_zero(),
+        _ => false,
+    }
+}
+
+fn operand_delta(left: U256, right: U256) -> U256 {
+    if left >= right { left - right } else { right - left }
+}
```
mablr: `operand_delta` is unsigned even for SLT/SGT. Fine as a raw field, but could misleading as a priority score for signed branches?

mattsse (commented):
Thanks for adding this. One cleanup lifecycle issue I noticed: `fuzz.corpus.frontier_dir` is now a generated artifact root, but `forge clean` only removes the existing fuzz failure/corpus dirs and invariant dirs. That means a project with `fuzz.frontier_dir` configured will keep stale `branch-frontiers.json` files after `forge clean`.

Suggestion: add `remove_test_dir(&self.fuzz.corpus.frontier_dir)` in `Config::cleanup`, next to `self.fuzz.corpus.corpus_dir`, and extend the existing `can_clean_test_cache` coverage to create/assert removal of the configured frontier dir.

mattsse (approved):
lgtm

mablr (approved):
All blocking findings seem addressed now 👍

Clanker suggested these tests improvements (non-blocking):
- Add a pure unit test for global frontier selection/truncation across multiple workers.
- Add CLI/integration coverage for `--fuzz-frontier-limit 1`; explicitly decide/test limit 0 semantics.
- Add an end-to-end signed comparison frontier test for JSON output, not just signed delta math.

PR #15514, opened 2026-07-01
fix(config): preserve inline config root

### LINKED ISSUE #15512: fs_permissions denies creating a new file under isolation when enabled via inline forge-config with a non-default profile
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.7.1

foundryup: 1.9.0

Forge / cheatcodes / `fs_permissions` / inline `forge-config` isolate

### Operating System

macOS (Apple Silicon) — also reproduces on Linux (GitHub Actions `ubuntu-latest`)

### Summary

Creating a **new** (not-yet-existing) file with `vm.writeJson`/`vm.writeFile`, inside a directory granted `read-write` in `fs_permissions`, is denied:

```
vm.writeJson: the path ./data/new_inlineIsolate.json is not allowed to be accessed for write operations
```

but **only** when both of these hold:

1. isolation is enabled for the test via the inline `/// forge-config: <profile>.isolate = true` annotation, and
2. the active profile is **not** `default` (e.g. `FOUNDRY_PROFILE=test`).

It is specifically the **inline annotation** that triggers it — not isolation itself. Enabling isolation via the `--isolate` CLI flag or via `isolate = true` in the profile config does **not** reproduce it. Overwriting an already-existing file always works.

### Minimal reproduction

`foundry.toml`:

```toml
[profile.default]
src = "src"
out = "out"
libs = ["lib"]
fs_permissions = [{ access = "read-write", path = "./data" }]

# non-default profile, inherits fs_permissions from default
[profile.test]

# non-default profile that enables isolation via config (not the inline annotation, not the CLI flag)
[profile.iso]
isolate = true
```

`test/IsolateFsPOC.t.sol`:

```solidity
pragma solidity ^0.8.13;

contract IsolateFsPOC is Test {
    function setUp() public {
        if (!vm.exists("./data")) vm.createDir("./data", true);
        if (!vm.exists("./data/existing.json")) vm.writeFile("./data/existing.json", "{}");
    }

    // inline isolate annotation + NEW file -> the only failing combination.
    /// forge-config: default.isolate = true
    function test_1_inlineIsolate_newFile() public {
        vm.writeJson('{"a":1}', "./data/new_inlineIsolate.json");
        vm.removeFile("./data/new_inlineIsolate.json");
    }

    // inline isolate annotation + EXISTING file -> always works (overwrite).
    /// forge-config: default.isolate = true
    function test_2_inlineIsolate_existingFile() public {
        vm.writeJson('{"a":2}', "./data/existing.json");
        vm.removeFile("./data/existing.json");
    }

    // no inline annotation + NEW file -> always works (any isolation here comes from the flag/config).
    function test_3_noInlineIsolate_newFile() public {
        vm.writeJson('{"a":1}', "./data/new_noInlineIsolate.json");
        vm.removeFile("./data/new_noInlineIsolate.json");
    }
}
```

Run the whole contract under each environment:

```
[A] FOUNDRY_PROFILE=test forge test --mc IsolateFsPOC -vv            # non-default profile
[B] forge test --mc IsolateFsPOC -vv                                 # default profile
[C] FOUNDRY_PROFILE=test forge test --isolate --mc IsolateFsPOC -vv  # non-default profile + --isolate flag
[D] FOUNDRY_PROFILE=iso  forge test --mc IsolateFsPOC -vv            # non-default profile, isolate=true in config
```

### Actual output

```
test                               inline isolate?  file       [A]    [B]    [C]    [D]
test_1_inlineIsolate_newFile       yes              new        FAIL   PASS   FAIL   FAIL
test_2_inlineIsolate_existingFile  yes              existing   PASS   PASS   PASS   PASS
test_3_noInlineIsolate_newFile     no               new        PASS   PASS   PASS   PASS
```

`[A]` vs `[B]` isolates it to the non-default profile. `[C]`/`[D]` show that isolation enabled by the CLI flag or by profile config does not trigger it — `test_3` is genuinely isolated under `[D]` yet passes — so the cause is the inline annotation, not isolation.

### Expected

`test_1_inlineIsolate_newFile` should PASS in every column: creating a file under a `read-write`-granted directory should be allowed under isolation regardless of how isolation is enabled or which profile is active.

### Notes

- Reproduces on the latest stable (`forge 1.7.1`, `4072e487`, 2026-05-08).
- Only the inline `forge-config` isolate annotation triggers it; the `--isolate` CLI flag and `isolate = true` in profile config do not.
- Only new-file creation is affected; overwriting an existing file works.
- Encountered in the Aave proposal test suite, which enables a per-test gas check under isolation via the inline annotation while running `FOUNDRY_PROFILE=test`; snapshot/report generation writes fresh JSON files under `./reports` and reverts on a clean checkout.

mattsse:
Fixes #15512.

Inline config merges first wrapped the inline provider in a Figment selected on the provider's default profile. That could switch extraction away from the active profile, and Config provider data only restored `root` on the selected profile even though the rest of the Config data is emitted under the default profile. With a non-default `FOUNDRY_PROFILE` and a `default.*` inline annotation, the merged config could fall back to default profile data with `root = "."`. New filesystem paths then normalized relative while permissions canonicalized against the project directory, so creating a new file under an inherited `fs_permissions` entry was denied.

This keeps inline provider merges selected on the active profile and restores `root` in default provider data as well, so profile fallback preserves the project root. The regression test covers inline `isolate` with a non-default profile inheriting default `fs_permissions` and creating a new file.

AI assistance was used to investigate the issue and draft this change.

### DIFF
--- a/crates/config/src/inline/mod.rs
+++ b/crates/config/src/inline/mod.rs
@@ -367,6 +367,50 @@ forge-config: default.symbolic.default_dynamic_length = 4
     }
 
     #[test]
+    fn merge_inline_provider_uses_selected_profile() {
+        let mut inline = InlineConfig::new();
+        inline
+            .insert(&natspec(
+                r#"
+forge-config: default.fuzz.runs = 1
+forge-config: ci.fuzz.runs = 2
+"#,
+            ))
+            .unwrap();
+
+        let profile = Profile::new("ci");
+        let config = Config {
+            profile: profile.clone(),
+            profiles: vec![Profile::Default, profile],
+            ..Default::default()
+        }
+        .merge_inline_provider(inline.provide("test/Symbolic.t.sol:Symbolic", "check"))
+        .unwrap();
+
+        assert_eq!(config.fuzz.runs, 2);
+    }
+
+    #[test]
+    fn merge_inline_provider_preserves_root_for_default_profile_fallback() {
+        let mut inline = InlineConfig::new();
+        inline.insert(&natspec("forge-config: default.isolate = false")).unwrap();
+
+        let profile = Profile::new("ci");
+        let root = std::path::PathBuf::from("project-root");
+        let config = Config {
+            profile: profile.clone(),
+            profiles: vec![Profile::Default, profile],
+            root: root.clone(),
+            ..Default::default()
+        }
+        .merge_inline_provider(inline.provide("test/Symbolic.t.sol:Symbolic", "check"))
+        .unwrap();
+
+        assert_eq!(config.root, root);
+        assert!(!config.isolate);
+    }
+
+    #[test]
     fn contract_symbolic_enabled_reads_contract_inline_config() {
         let mut inline = InlineConfig::new();
         inline
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -830,7 +830,7 @@ impl Config {
     /// Applies an inline provider on top of the current config without reloading external
     /// providers such as `foundry.toml`, env vars, or remappings.
     pub fn merge_inline_provider<T: Provider>(&self, provider: T) -> Result<Self, Error> {
-        let provider = Figment::from(provider);
+        let provider = Figment::from(provider).select(self.profile.clone());
         let invariant_corpus_random_sequence_weight_configured =
             self.invariant.corpus_random_sequence_weight_configured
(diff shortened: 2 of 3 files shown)


mablr (approved):
LGTM

figtracer (approved):
lg

Ari4ka (approved):

PR #15515, opened 2026-07-01
refactor(symbolic): hashcons expressions

DaniPopes:
Refactors symbolic expression construction around `SymCx` so expressions, booleans, and symbolic byte values are always interned through the context. Hash-consing now lives under the expression module, common values are cached on the context, and equality stays cheap by using interned handles instead of structural comparisons.

This contribution was implemented with AI assistance for code generation and refactoring.


mattsse (approved):
LGTM

figtracer (approved):

mablr (approved):
sgtm

PR #15516, opened 2026-07-01
perf(forge): skip unused fixture collection

figtracer:
Avoid collecting fuzz/table fixtures during setup when the selected test functions cannot use them. Forge previously evaluated every `fixture*` function and public fixture array for a matched contract immediately after `setUp`, even when the current filter only selected unit tests.

This keeps fixture collection for selected fuzz, table, and invariant tests, but skips it for unit-only and symbolic-only runs.

Warmed-artifact project with one selected unit test and an unused 5,000-entry fixture array in the same contract, debug Forge build:

- master: `340.5 ms +/- 30.8 ms`
- patched: `141.9 ms +/- 39.6 ms`

Selected fuzz test on the same project is effectively unchanged, as expected:

- master: `335.1 ms +/- 19.8 ms`
- patched: `340.6 ms +/- 16.6 ms`

### DIFF
--- a/crates/forge/tests/cli/test_cmd/table.rs
+++ b/crates/forge/tests/cli/test_cmd/table.rs
@@ -1,5 +1,6 @@
 //! Table tests.
 
+use foundry_config::fs_permissions::PathPermission;
 use foundry_test_utils::{forgetest_init, str};
 
 forgetest_init!(should_run_table_tests, |prj, cmd| {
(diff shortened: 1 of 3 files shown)


crates/forge/tests/cli/test_cmd/table.rs
```diff
@@ -128,6 +129,43 @@ Tip: Run `forge test --debug --match-test <TEST_NAME>` to inspect one failing te
 "#]]);
 });
 
+forgetest_init!(does_not_evaluate_unused_fixtures_for_unit_test_filter, |prj, cmd| {
+    let marker = prj.root().join("fixture-called.txt");
+    prj.update_config(|config| config.fs_permissions.add(PathPermission::write(prj.root())));
+    prj.add_test(
+        "UnusedFixtures.t.sol",
```
mablr, line 136: A similar marker test for fuzz fixtures could be beneficial for better coverage ?

mablr (dismissed):
LGTM, reproduced perf measurements locally.

mattsse (approved):
LGTM

mablr (approved):

PR #15517, opened 2026-07-01
feat(fuzz): add `cmin` corpus minimization subcommand

mablr:
## Description

Part of OSS-297

Adds `forge fuzz cmin`, a corpus minimizer that replays persisted fuzz/invariant corpus entries and keeps only replayable inputs that add new coverage. It preserves corpus layout, stages output safely, rejects existing outputs, and reports actionable diagnostics for stale, empty, unreadable, rejected, or failing entries.

Multi-target runs namespace coverage per target, and replayable coverage preservation is tested via showmap edge-set comparison.

Design decisions (can be discussed): Reverting entries are intentionally skipped, and multi-target zero-replay diagnostics de-duplicate counts for readability.

### DIFF
--- a/crates/forge/src/cmd/fuzz.rs
+++ b/crates/forge/src/cmd/fuzz.rs
@@ -1,13 +1,22 @@
-use crate::{cmd::test::TestArgs, multi_runner::ShowmapConfig, result::TestOutcome};
+use crate::{
+    cmd::test::{FilterArgs, FuzzMinimizeReplaySession, TestArgs},
+    multi_runner::{FuzzMinimizeEdgeIndices, FuzzMinimizeObservation, ShowmapConfig},
+    result::TestOutcome,
+};
 use alloy_dyn_abi::JsonAbiExt;
 use alloy_json_abi::{Function, JsonAbi};
 use alloy_primitives::Selector;
 use clap::{Parser, Subcommand, ValueEnum, ValueHint};
 use eyre::{Context, Result, bail};
-use foundry_common::{fmt::format_tokens_raw, fs, sh_println};
+use foundry_cli::opts::{BuildOpts, EvmArgs, GlobalArgs};
+use foundry_common::{
+    fmt::format_tokens_raw,
+    fs, sh_println, sh_status,
+    shell::{OutputMode, Shell},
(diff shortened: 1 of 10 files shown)


mattsse (dismissed):
LGTM

crates/forge/src/cmd/fuzz.rs
```diff
@@ -318,3 +542,79 @@ fn read_corpus_entries(path: &Path) -> Result<Vec<CorpusDirEntry>> {
...
+        | merge_new_edge_vec(&mut cumulative.sancov_edges, &observation.sancov_edges)
+}
+
+fn merge_new_edge_vec(cumulative: &mut Vec<u8>, candidate: &[u8]) -> bool {
+    if cumulative.len() < candidate.len() {
+        cumulative.resize(candidate.len(), 0);
+    }
+    let mut improved = false;
+    for (cumulative, &candidate) in cumulative.iter_mut().zip(candidate) {
+        if candidate > 0 && *cumulative == 0 {
```
stevencartavia: **cmin drops entries that add coverage.** `merge_new_edge_vec` only treats first-time edges as improvements (`candidate > 0 && *cumulative == 0`), ignoring hit-count bucket increases. But the campaign's coverage metric treats a higher AFL bucket as new coverage (`merge_edge_count`: `*hist < bucket`), and cmin's observations are already AFL-bucketed — so cmin can drop entries the campaign would have kept.

Repro:
```solidity
function testFuzz_reps(uint256 input) public {
    uint256 n = input % 5;
    for (uint256 i = 0; i < n; i++) { value = value + i + 1; }
}
```
Entries `input=1` and `input=4` hit the same edges with different counts (`input=4` lands in higher buckets). Corpus entries are read path-sorted, so with `input=1` named to sort first, cmin keeps 1/2 and drops `input=4`. It succeeds silently — `forge_fuzz_cmin_keeps_coverage_adding_entries` misses it because it compares only edge IDs, not counts.

This affects sancov edges too. If edge-only minimization is intended, let's document it.

mattsse (approved):

grandizzy (approved):
lgtm

PR #15518, opened 2026-07-01
feat(lint): add todo-comment detector

tskoyo:
This adds the info-severity `todo-comment` lint from #14381. It reports unresolved `TODO` and `FIXME` markers in line, block, and NatSpec comments while avoiding prose, dotted names, identifiers, and late bare words that are not marker syntax.

The detector handles case-insensitive punctuation forms, bare markers at the first meaningful token or after a NatSpec tag, physical lines inside multiline block comments, per-comment deduplication, and lint control comments. The companion Book documentation is foundry-rs/book#1952. A fresh Codex context independently reviewed the implementation and corrected multiline block-comment handling and its permanent fixture coverage.

This PR was completed with Codex assistance across code, tests, lint documentation, and review responses. The original implementation and direction are credited to @tskoyo.


crates/lint/src/sol/info/todo.rs
```diff
@@ -0,0 +1,46 @@
...
+
+        for comment in comments.iter() {
+            for marker in MARKERS {
+                if comment.lines.iter().any(|line| line.contains(marker)) {
+                    _ctx.emit_with_msg(
+                        &TODO,
+                        comment.span,
+                        format!("unresolved `{marker}` comment"),
+                    );
+                    break;
```
mablr: The matcher is a raw, case-sensitive substring search, so it misses common marker spellings like `// todo`, `// ToDo`, `// Fix me`, and `// fixme`..., but still flags ordinary words contianing`TODO`. It also reports `TODO` for `// FIXME first, TODO second` because `MARKERS` order wins over source order.

mablr (changes requested):
Thanks for your contribution, please address the inline comment.

crates/lint/src/sol/info/todo.rs
tskoyo: Oh makes sense, wasn't thinking about that.. Thanks for the review

tskoyo: Added more tests in the description

crates/lint/testdata/Todo.stderr
mablr, line 1: This should be blessed for the updated implementation

crates/lint/src/sol/info/todo.rs
```diff
@@ -0,0 +1,47 @@
...
+        // Build comments from the source, same call the crate already uses in mod.rs.
+        let comments = Comments::new(file, _ctx.session().source_map(), false, false, None);
+
+        for comment in comments.iter() {
+            for line in &comment.lines {
+                for word in line.split(|c: char| !c.is_alphanumeric()) {
+                    if MARKERS.iter().any(|m| word.eq_ignore_ascii_case(m)) {
+                        _ctx.emit_with_msg(
+                            &TODO,
+                            comment.span,
```
mablr: Multiple markers in one comment produce duplicate same-span diagnostics, only one warning per span should be emitted.

crates/lint/docs/todo.md
```diff
@@ -0,0 +1,45 @@
...
+**ID**: `todo`
+
+Flags `TODO` and `FIXME` markers left in comments, which signal unfinished work or known
+bugs that have not been resolved before the code reached production.
+
+## What it does
+
+Scans every comment in the source file (single-line `//`, block `/* */`, and NatSpec `///`)
+and reports any comment whose text contains the literal string `TODO` or `FIXME`. Markers
+inside string literals are not flagged.
```
mablr: Docs don’t precisely match implementation, please update once it is stabilized.

crates/lint/testdata/Todo.sol
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+contract Todo {
+    // ToDo: implement access control
+    function unfinished() public {}
```
mablr, line 13: This fixture currently runs with the default lint set and contains many empty function bodies. That passes on this PR alone, but if `empty-block` lands, the default UI run for `Todo.sol` will also emit `empty-block` diagnostics and make `Todo.stderr` fail.

Please isolate this fixture with `//@compile-flags: --only-lint todo`, or make these unrelated function bodies non-empty like the empty-block PR did for other fixtures.

mablr (changes requested):

crates/lint/README.md
```diff
@@ -68,7 +68,11 @@ It helps enforce best practices and improve code quality within Foundry projects
   - `missing-inheritance`: Flags contracts that implement every external function of an interface without explicitly inheriting from it.
   - `low-level-calls`: Direct use of low-level calls should be avoided.
   - `event-fields`: `address` event parameters should be `indexed` for efficient log filtering.
+<<<<<<< HEAD
+  - `todo`: Detects the usage of todo keyword in comments
+=======
   - `unused-error`: Custom error declarations that are never referenced should be removed.
+>>>>>>> master
```
mablr: Please fix these merge conflict markers.

tskoyo: Ah, didn't see it. Fixed now

tskoyo:
Here's the PR to create a new page in the docs for todo lint -> https://github.com/foundry-rs/book/pull/1941

crates/lint/testdata/Todo.stderr
```diff
@@ -0,0 +1,74 @@
+note[todo]: unresolved `todo` comment
```
mablr: This blessed diagnostic is a false positive: the lint is reporting the UI-test `//@compile-flags: --only-lint todo` directive as an unresolved TODO comment. That directive is test harness metadata, not source TODO/FIXME debt. The lint should skip control comments such as `@compile-flags:` and likely `forge-lint:` comments before marker matching, then the fixture should assert the directive stays silent.

crates/lint/src/sol/info/todo.rs
```diff
@@ -0,0 +1,55 @@
...
+        // Build comments from the source, same call the crate already uses in mod.rs.
+        let comments = Comments::new(file, _ctx.session().source_map(), false, false, None);
+
+        for comment in comments.iter() {
+            let mut seen = HashSet::new();
+
+            let found: Vec<&str> = comment
+                .lines
+                .iter()
+                .flat_map(|line| line.split(|c: char| !c.is_alphanumeric()))
```
mablr: This split treats any punctuation as a TODO marker boundary, so ordinary hyphenated/dotted phrases like `todo-list`, `todo.md`, or `todo-lint` are reported as unresolved TODO comments. I verified this by temporarily adding `// todo-list should not be treated as an unresolved marker if whole words are required` to the lint fixture and rerunning `cargo test -p forge --test ui Todo -- --nocapture`; it emitted an extra `note[todo]`. If the intended rule is marker comments, consider matching `TODO`/`FIXME` only when followed by marker punctuation/whitespace such as `:`, `(`, or end-of-token, and add fixture coverage.

crates/lint/src/sol/info/todo.rs
```diff
@@ -0,0 +1,55 @@
...
+                .filter(|&word| MARKERS.iter().any(|m| word.eq_ignore_ascii_case(m)))
+                .filter(|&word| seen.insert(word))
```
mablr: The dedupe is case-sensitive because it stores the original word. As a result, one comment containing `FIXME ... fixme` reports `FIXME, fixme` as two markers even though they are the same logical marker; the new fixture currently blesses `FIXME, TODO, fixme`. Consider deduping by normalized ASCII case and reporting canonical marker names to avoid noisy diagnostics.

tskoyo:
@mablr thanks for your comments. I added a new logic to detect `MARKERS` (todo and fixme). However, this new solution wouldn't detect bare markers.

Quick example here:

```sol
// FixMe: this is broken
function buggy() public {}

uint8 x = 1; //TODO: validate this

// ToDo: implement access control
function unfinished() public {}
```

These are all going to be reported by linter as they have `TRAILING` chars.

However, bare markers with no trailing character are not reported:

```
// TODO
// FIXME first, TODO second, and fixme third
// check this later TODO
```

This is a deliberate tradeoff. Treating `whitespace/EOL` as a valid boundary would catch these, but it reintroduces false positives on ordinary prose like`// see the todo list` or `// check other tickets in todo`. I went with the stricter rule since a missed marker is cheaper than a noisy lint, but happy to revisit if you'd rather bias the other way.

I also introduced TRAILING chars (`[':', '(', ',', ';', '.', ')']`) that are usually used after TODO/FIXME markers, and we can change them if you think some of them are unnecessary.

mablr (dismissed):
Thank you @tskoyo, this looks good enough to me.

Pending another review :)

mablr (dismissed):

crates/lint/src/sol/info/todo.rs
```diff
@@ -0,0 +1,92 @@
...
+
+/// If `token` begins with a marker followed by a valid boundary, return that marker.
+fn marker_at_start(token: &str) -> Option<&str> {
+    MARKERS.iter().find_map(|m| {
+        let prefix = token.get(..m.len())?;
+        if !prefix.eq_ignore_ascii_case(m) {
+            return None;
+        }
+        let mut trailing = token[m.len()..].chars();
+        let after = trailing.next()?;
```
mattsse, line 93: **[P2] Treat marker-position EOL as a valid boundary**

`trailing.next()?` rejects a marker whenever the whitespace-delimited token ends immediately after `TODO`/`FIXME`. For example, all of these currently remain silent:

```solidity
// TODO
// TODO implement access control
// FIXME this check is wrong
```

The first case is already present in `Todo.sol`, but the blessed output confirms that it is not reported. These are canonical marker forms, so this leaves the detector ineffective for a common usage.

Suggested fix: keep the strict punctuation rule for markers appearing later in prose, but track each token's position and allow an empty suffix when the marker is in an intent-signaling position—either the first meaningful token after the comment prefix or immediately after a NatSpec tag such as `@dev`. Concretely, pass an `allow_bare` flag into `marker_at_start`; when `token[m.len()..]` is empty, return the marker only if `allow_bare` is true. This catches the examples above while keeping `// see the todo list`, `// check this later TODO`, `todo.md`, and `TODO_ITEM` silent. Please add these positive and negative cases to the UI fixture and bless the updated diagnostics.

mattsse (dismissed):
Thanks for tightening the matcher and covering the earlier false positives. There is one remaining correctness gap in the core detector behavior: canonical TODO/FIXME comments without punctuation are currently ignored. Please address the inline comment and add the corresponding positive and negative fixture coverage.

mablr (approved):
LGTM

figtracer (approved):

PR #15519, opened 2026-07-01
feat(symbolic): seed fuzz corpus from branch frontiers

figtracer:
### Summary
- Add `symbolic.use_fuzz_frontiers` / `--symbolic-use-fuzz-frontiers` and `symbolic.frontier_limit` / `--symbolic-frontier-limit`.
- Import `branch-frontiers.json` records from the fuzz frontier directory, run symbolic execution against the captured comparison site, and constrain the run to flip the observed branch result.
- Replay every solved candidate concretely before writing it into the fuzz corpus, so frontier-derived seeds only persist when they reproduce the expected pass/fail behavior.

#### Example

```solidity
contract SymbolicFuzzFrontierSeed {
    function testFuzz_frontier(uint64 amount, uint256 feeMultiplier) public pure {
        uint256 credited;
        unchecked {
            credited = uint256(amount) + (feeMultiplier - 100);
        }

        if (feeMultiplier < 100) {
            assert(credited <= amount);
        }
    }
}
```

A short fuzz run can reach the frontier at `feeMultiplier == 100` without crossing the `feeMultiplier < 100` branch:

```sh
forge test --match-test testFuzz_frontier \
  --fuzz-runs 8 \
  --fuzz-seed 0x1234 \
  --threads 1 \
  --fuzz-frontier-dir fuzz_frontiers
```

The new targeted symbolic pass consumes that frontier, flips the captured `LT` comparison, and persists a replay-confirmed corpus seed. In the test, `forge fuzz replay --corpus-dir fuzz_corpus` then fails on the generated seed, proving the seed is useful to the fuzzer rather than just a synthetic symbolic result.


crates/evm/symbolic/src/executor/run.rs
```diff
@@ -273,6 +274,7 @@ impl SymbolicExecutor {
                         });
```
mablr, line 279: Should counterexamples also be gated on `state.satisfies_branch_target()` when `branch_target` is set? It looks like only safe `success_input` collection checks whether the imported frontier was reached, while counterexamples returned here and from the later `Halt`/`Failure` paths may bypass the target. Since `try_seed_fuzz_corpus_from_frontiers` persists any replay-failing counterexample, `--symbolic-use-fuzz-frontiers` could save an unrelated symbolic failure instead of a seed that flipped the imported frontier?

mablr (commented):
Sgtm, just one question

mattsse (dismissed):
LGTM

mattsse (commented):
I found two issues worth fixing before merging:

1. The branch-target constraint is applied on every later visit to the same comparison site, not just the first flipped visit. In [`apply_branch_target_constraint`](https://github.com/foundry-rs/foundry/blob/669b412fa859106b3447d26111bc9ce8f75482f1/crates/evm/symbolic/src/executor/opcodes.rs#L33-L41), once `branch_target_reached` has been set, future executions of the same `(address, pc, opcode)` still push the opposite-result constraint again. That can reject useful seeds for loops or repeated calls where the frontier is flipped once, then the same site is encountered later with the original result.

Suggested fix: make the target a one-shot reach condition. For example, expose a `branch_target_reached()` / `has_reached_branch_target()` helper on `PathState`, and return `Ok(true)` before adding `desired` when the target is already reached. Another option is to have `mark_branch_target_reached()` clear the target, but keeping the target plus a reached flag may be clearer for diagnostics and child-state propagation.

2. `vm.deployCode` drops target progress from constructor/initcode execution. Normal external calls and plain `CREATE` copy the child progress back with [`parent.inherit_branch_target_progress(&outcome.state)`](https://github.com/foundry-rs/foundry/blob/669b412fa859106b3447d26111bc9ce8f75482f1/crates/evm/symbolic/src/executor/calls.rs#L1027-L1030) and [`create.rs`](https://github.com/foundry-rs/foundry/blob/669b412fa859106b3447d26111bc9ce8f75482f1/crates/evm/symbolic/src/executor/create.rs#L123-L126), but the analogous path in [`deploy_code_cheatcode_call`](https://github.com/foundry-rs/foundry/blob/669b412fa859106b3447d26111bc9ce8f75482f1/crates/evm/symbolic/src/executor/cheatcodes.rs#L199-L204) only copies constraints and `next_symbol`. A frontier inside initcode reached through `vm.deployCode` can therefore be solved in the child and then discarded because the parent/root state never satisfies the target.

Suggested fix: add the same `parent.inherit_branch_target_progress(&outcome.state);` immediately after `parent.next_symbol = outcome.state.next_symbol;` in `deploy_code_cheatcode_call`. A focused regression test using `vm.deployCode` with a constructor comparison would cover this path.

mattsse (dismissed):
LGTM

mablr (dismissed):

grandizzy:
@figtracer looks good, though wonder how did we end up with `"frontier_limit": 8`

figtracer:
@grandizzy it's a conservative default since this is quite expensive

edit: changing this to 256, it's considerably slower but still very reasonable

```
8 - 0.09s
256 - 2.93s
```

mablr (approved):

grandizzy (approved):

PR #15521, opened 2026-07-01
fix(fuzz): dedupe branch frontiers across workers

stevencartavia:
The branch-frontier artifact writer just concatenated each worker's records, so multi-threaded runs produced duplicate sites and could keep a worse operand_delta. Apply the existing per-worker dedup (by site, keeping the smallest delta) globally when merging.


mattsse (approved):
LGTM

figtracer (approved):

PR #15522, opened 2026-07-01
fix(fuzz): omit frontier new_coverage when not collected

stevencartavia:
The branch-frontier artifact always wrote `new_coverage=false` in the documented frontier-only workflow, since edge coverage isn't collected there. Make it `Option<bool>` and omit it unless edge coverage is collected (corpus, edge-coverage metrics, or sancov).


mablr (approved):

mattsse (approved):
LGTM

PR #15523, opened 2026-07-01
fix(forge): list only runnable fuzz tests

mablr:
## Description

Follow-up to #15227 as part of OSS-297.

Fix forge fuzz run --list so it only reports tests that fuzz mode can actually execute.

Previously, list output was built from the broader test discovery result, so unit-style and table tests could appear even though forge fuzz skips them at runtime. The list path now applies the same runnable-test classification used by fuzz execution and omits contracts that have no remaining fuzz or invariant tests after filtering.

The CLI coverage adds a mixed contract with unit, fuzz, invariant, and table tests to assert that fuzz list output includes only runnable entries.

### DIFF
--- a/crates/forge/src/multi_runner.rs
+++ b/crates/forge/src/multi_runner.rs
@@ -159,13 +159,24 @@ impl<FEN: FoundryEvmNetwork> MultiContractRunner<FEN> {
                 let tests = c
                     .abi
                     .functions()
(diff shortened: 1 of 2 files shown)


mattsse (approved):
LGTM

stevencartavia (approved):

PR #15524, opened 2026-07-01
perf(symbolic): shrink SMT CSE planning

DaniPopes:
This is stacked on `dani/symbolic-normalization-refactor` and keeps SMT CSE planning linear in the hashconsed expression DAG. It preserves reference counts for binding decisions but avoids walking the same shared node repeatedly, and it avoids eagerly distributing shifts over OR expressions when both sides remain live, which keeps bit-smearing patterns compact.

On Solady `check_HasDuplicateHashmapCapacityTrickEquivalence` with symbolic SMT dumping enabled, this reduced SMT input from 21,908 to 4,961 bytes and max query size from 10,968 to 2,685 bytes. The focused run stayed passing, with wall time moving from 2.40s to 2.38s.

AI assistance was used while preparing this change.

### DIFF
--- a/crates/evm/symbolic/src/runtime/expr/bool.rs
+++ b/crates/evm/symbolic/src/runtime/expr/bool.rs
@@ -68,77 +68,73 @@ impl SymBoolExpr {
         Self::cmp(cx, SymCmpOp::Eq, left, right)
     }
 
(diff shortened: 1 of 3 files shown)


mattsse (approved):
LGTM

PR #15525, opened 2026-07-01
fix(ci): bench harness parity

mablr:
## Description

`foundry-bench` parity fixes:

- local HEAD build with dist profile
- pin dynamic test linking to false (as default 1.7.1 was false and 1.8.0 will be true)


mattsse (approved):
LGTM

figtracer (approved):

PR #15526, opened 2026-07-01
feat(lint): add dangerous-unary-operator detector

0xMars42:
Adds the `dangerous-unary-operator` detector to `forge lint`, one of the open items in #14381 (Slither/Aderyn parity).

## What it does

Flags an assignment whose `=` is fused to a unary operator (`=-`, `=~`), e.g. `x =- 1`, which parses as `x = -1` (a plain assignment of a unary expression) rather than the compound `x -= 1` it resembles. Mirrors Slither's `incorrect-unary` (Dangerous unary expressions).

It is an early AST pass. Because `x =- 1` and the legitimate `x = -1` parse to the same node, it only fires when the source fuses `=` to the leading unary, by checking that the snippet between the LHS and the RHS ends with `=`. The unary may be the whole RHS (`x =- 1`) or lead a larger one (`x =- a + 1`, which parses as `x = (-a) + 1`). Spaced forms (`x = -1`, `x = ~y`), real compound operators (`x -= 1`), plain subtraction (`x = y - 1`), and forms where a comment sits between `=` and the operator are left alone.

`=+` is not reported: unary `+` was removed in Solidity 0.5.0, so `x =+ 1` is a parse error that solar drops without producing a node, and it never reaches the linter.

## Severity

Medium.

## Tests

`crates/lint/testdata/DangerousUnaryOperator.sol` covers the fused forms, including a unary leading a larger RHS and a comment before the `=` (flagged), and the spaced / compound / subtraction / broken-fusion forms (not flagged).

Part of #14381.

### DIFF
--- a/crates/lint/src/sol/med/dangerous_unary_operator.rs
+++ b/crates/lint/src/sol/med/dangerous_unary_operator.rs
@@ -17,18 +17,33 @@ impl<'ast> EarlyLintPass<'ast> for DangerousUnaryOperator {
         // `x =- 1` lexes as `x` `=` `-` `1` and parses as a plain assignment of `-1`, identical to
         // the intentional `x = -1`, yet it reads like the compound `x -= 1` it was probably meant
         // to be. Solidity has no `~=` operator either, so `x =~ y` is the same trap.
-        // Because the parsed node matches the legitimate spaced form, only flag when the
-        // source fuses `=` to the unary operator (`=-` / `=~`), never `= -`. `=+` never
+        // The fused unary can also lead a larger RHS: `x =- a + 1` parses as `x = (-a) + 1`, whose
+        // RHS is a `Binary` (or `Ternary`) with the `-a` unary as its leftmost operand. Follow the
+        // left spine to that leading unary so those forms are caught too, not only `x =- a`.
+        // Because the parsed node matches the legitimate spaced form, only flag when the source
+        // fuses `=` to the leading unary (`=-` / `=~`), never `= -`. The gap between the LHS and
+        // `rhs.span` (which starts at that unary) holds only whitespace, comments and the `=`
+        // token, and no comment can end with `=` (block comments end with `*/`, line comments
+        // with a newline), so the gap ends with `=` exactly when the pair is fused. `=+` never
         // reaches here: unary `+` was removed in Solidity 0.5.0, and solar drops it during
         // parsing without producing a node.
         if let ExprKind::Assign(lhs, None, rhs) = &expr.kind
-            && let ExprKind::Unary(op, _) = &rhs.kind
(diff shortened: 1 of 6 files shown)


crates/lint/src/sol/med/dangerous_unary_operator.rs
```diff
@@ -0,0 +1,34 @@
...
+    fn check_expr(&mut self, ctx: &LintContext, expr: &'ast Expr<'ast>) {
+        // `x =- 1` lexes as `x` `=` `-` `1` and parses as a plain assignment of `-1`, identical to
+        // the intentional `x = -1`, yet it reads like the compound `x -= 1` it was probably meant
+        // to be. Solidity has no `~=` operator either, so `x =~ y` is the same trap.
+        // Because the parsed node matches the legitimate spaced form, only flag when the
+        // source fuses `=` to the unary operator (`=-` / `=~`), never `= -`. `=+` never
+        // reaches here: unary `+` was removed in Solidity 0.5.0, and solar drops it during
+        // parsing without producing a node.
+        if let ExprKind::Assign(lhs, None, rhs) = &expr.kind
+            && let ExprKind::Unary(op, _) = &rhs.kind
```
stevencartavia: The rule only fires when the RHS is a direct unary expression. When the unary is nested inside a larger expression (RHS parses as Binary), nothing is emitted, so these are missed:

```solidity
c =- x + 1;    // typo for `c -= x + 1`
e =- x * 2;    // typo for `e -= x * 2`
```

`c =- x` (simple RHS) is flagged, but `c =- x + 1` is not, even though it's the same fused-operator typo.

Repro: add `a =- x + 1; //~WARN: unary operator fused to` to testdata/DangerousUnaryOperator.sol and the UI test fails with "expected diagnostic not emitted".

0xMars42: Fixed in e600890f. The `rhs.kind == Unary` match only covered a RHS that was itself the unary expression. `leads_with_fusable_unary` now follows the left spine of binary and ternary nodes to the leading operand, so `d =- x + 1` and `e =- x * 2` warn while the spaced `e = -x + 1` stays clean.

The fuse check also moved from the unary's span to `rhs.span`: the gap between the LHS and the RHS holds only whitespace, comments and the `=` token, and no comment can end with `=`, so `gap.ends_with('=')` is exactly the fused condition. This keeps `f /* note */=- 1` (fused, comment before the `=`) flagged and `f =/* note */- 1` (comment breaks the fusion) clean. Your repro cases are both in the testdata now.

mablr (approved):
I haven't more blockers, sgtm.

(I updated branch to land Deny.toml fix)

figtracer (approved):

mattsse (approved):

PR #15527, opened 2026-07-01
perf(symbolic): flatten concrete byte reads

figtracer:
`SymBytes::concrete_bytes` now appends concrete ranges into one output buffer instead of recursively allocating child `Vec`s for each concat/slice segment. The concrete slice append logic is kept inline at the single call site.

Numbers rerun after rebasing onto current `master`:
- hyperfine release probe over 100k materializations of a 512-piece concat: 1.081s ± 0.072s -> 291.2ms ± 9.6ms, 3.71x ± 0.28 faster
- Samply at 4 kHz over 5 iterations: 23,629 -> 5,521 benchmark-thread samples, 76.6% fewer

Validation: `git diff --check`, `cargo +1.96.0 test -p foundry-evm-symbolic --lib`, `cargo +1.96.0 clippy -p foundry-evm-symbolic --lib --all-targets -- -D warnings`.

### DIFF
--- a/crates/evm/symbolic/src/runtime/bytes.rs
+++ b/crates/evm/symbolic/src/runtime/bytes.rs
@@ -207,9 +207,13 @@ impl SymBytes {
         }
         match self.kind() {
             SymBytesKind::Concrete(bytes) => {
-                let out = (0..len)
-                    .map(|idx| bytes.get(offset + idx).copied().unwrap_or_default())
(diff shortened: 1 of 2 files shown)


mattsse (dismissed):
LGTM

stevencartavia (approved):

grandizzy (approved):
lgtm

PR #15528, opened 2026-07-01
perf(symbolic): normalize expression construction

DaniPopes:
This refactors symbolic solver normalization so more simple rewrites happen when expressions are constructed instead of being rebuilt later in the solver path. It also simplifies the solver-side normalization/cache flow, folds comparison equality into the comparison op handling, models addmod/mulmod as ternary word ops, and keeps the SMT CSE planning changes from the stacked PR.

The intent is to reduce repeated symbolic expression rebuilding and make normalized solver queries cheaper and more predictable without changing the external symbolic test surface.


mattsse (approved):
LGTM

figtracer (approved):

PR #15532, opened 2026-07-02
fix(ci): scope bench bins

stevencartavia:
The bench harness builds Foundry with `--bins`, which compiles all 6 workspace binaries under the fat-LTO `dist` profile, but only `forge`, `cast`, `anvil`, and `chisel` are ever activated and benchmarked. Build just those four to skip two unused optimized links on every local benchmark run.


grandizzy (approved):

mattsse (approved):
LGTM

mablr (approved):

PR #15536, opened 2026-07-02
perf(symbolic): use single-var solver witnesses

figtracer:
Symbolic solving now reuses the existing validated single-variable witness search before spawning an SMT subprocess for satisfiable `is_sat` and `model` queries. The fast path only returns after evaluating every original constraint against the candidate model, so unresolved cases still fall through to the solver.

### Results

| --- | ---: | ---: | ---: |
| `grandizzy/symbolic-bug-suite` wall time | 555.6ms ± 15.6ms | 510.2ms ± 39.6ms | 1.09x ± 0.09 faster |
| bug-suite backend SMT queries | 211 | 95 | -55.0% |
| bug-suite reported solver time | 3131ms | 1501ms | -52.1% |
| built-in symbolic benches wall total | 6.137s | 6.183s | +0.7% |
| built-in symbolic benches solver total | 5845ms | 5649ms | -3.4% |
| built-in symbolic benches SMT queries | 140 | 107 | -23.6% |

### DIFF
--- a/crates/evm/symbolic/src/runtime/bytes.rs
+++ b/crates/evm/symbolic/src/runtime/bytes.rs
@@ -430,7 +430,9 @@ impl SymBytes {
     }
 
     pub(crate) fn materialize(&self, cx: &mut SymCx) -> Vec<SymExpr> {
-        (0..self.len()).map(|idx| self.byte(cx, idx)).collect()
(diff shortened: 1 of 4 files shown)


mablr (approved):
Reproduced perf gain locally, seems good

grandizzy (approved):
lgtm

PR #15537, opened 2026-07-02
refactor(symbolic): flatten concrete memory reads

figtracer:
SymMemory::read_concrete now reuses the existing stored-byte range reconstruction before falling back to byte-by-byte lookup for symbolic-offset writes. This keeps the change scoped to the concrete read path and preserves the existing fallback for unresolved symbolic writes.

In a temporary ignored release probe over 2k reads of 512 single-byte concrete writes:

hyperfine: 5.970s ± 0.046s -> 8.4ms ± 1.0ms (155.3x ± 4.1)
samply over five iterations (probe-thread samples): 124,630 -> 807 (99.4% fewer)


mablr (approved):

mattsse (approved):
LGTM

PR #15538, opened 2026-07-02
chore(forge): remove stale fuzz bare-name filter todo

mablr:
## Motivation

This removes a stale TODO about making exact bare-name fuzz filters such as `^testFoo$` match parameterized fuzz test signatures like `testFoo(uint256)`.

`--match-test` currently applies to ABI test signatures, not bare function names. That behavior is consistent across forge test, forge coverage, and forge fuzz run: users can match a test by using an unanchored name regex such as `testFoo`, or by anchoring the full signature shape such as `^testFoo\\(uint256\\)$`. Special-casing forge fuzz run would make fuzz filtering semantics diverge from the rest of Forge and would require overload-specific selection logic for a narrow UX edge case.

### DIFF
--- a/crates/forge/src/cmd/test/filter.rs
+++ b/crates/forge/src/cmd/test/filter.rs
@@ -266,9 +266,6 @@ impl TestFilter for ProjectPathsAwareFilter {
                     && (failure.test == signature || failure.test == name)
             })
         } else {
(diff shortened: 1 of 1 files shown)


mattsse (approved):
LGTM

figtracer (approved):

PR #15539, opened 2026-07-02
feat(lint): add unused-error detector

0xMars42:
Adds the `unused-error` detector to `forge lint`, one of the open items in #14381 (Slither/Aderyn parity; Aderyn ships it as `unused-error`).

## What it does

Flags a custom error declaration that no expression in the compiled sources references.

Any resolved reference counts as a use: `revert Err(...)` (including the qualified `revert Lib.Err(...)`, which the lowering resolves into a single identifier), `require(cond, Err(...))` (Solidity 0.8.26+), and `Err.selector` (including through `abi.encodeWithSelector`). A qualified member access such as `Lib.Err.selector` carries no resolution on the `Err` segment in the HIR, so the name is looked up in the scopes its base can designate: a contract or library, a module alias (including the symbols it re-exports through its own imports), or a member chain leading to one (`NS.Lib.Err.selector`).

It is a project pass: uses are collected across the whole compilation unit while only errors declared in input files are reported, so a shared errors file that is reverted from elsewhere in the project is attributed correctly. Errors declared in interfaces and abstract contracts are not reported: they are ABI surface for implementers and off-chain consumers, which may live outside the compiled sources.

## Severity

Info.

## Tests

`crates/lint/testdata/UnusedError.sol` covers flagged declarations at contract, library, and file level, and the non-flagged uses: plain and qualified reverts, `require` with a custom error, selector accesses (direct, qualified, and chained through a module alias), `abi.encodeWithSelector`, an error declared in an abstract base and reverted by the child, and the interface / abstract contract exemptions.

## Limitations

A selector hardcoded in inline assembly or built with `abi.encodeWithSignature("Err(...)")` does not reference the declaration and is not seen as a use.

Part of #14381.

### DIFF
--- a/crates/lint/src/sol/info/unused_error.rs
+++ b/crates/lint/src/sol/info/unused_error.rs
@@ -1,0 +1,188 @@
+use crate::{
+    linter::{Lint, ProjectLintEmitter, ProjectLintPass, ProjectSource},
+    sol::{Severity, SolLint, info::UnusedError},
+};
+use solar::{
+    ast::ContractKind,
+    interface::{data_structures::Never, source_map::FileName},
+    sema::hir::{self, Visit as _},
+};
+use std::{
+    collections::{HashMap, HashSet},
+    ops::ControlFlow,
+};
+
+declare_forge_lint!(UNUSED_ERROR, Severity::Info, "unused-error", "custom error is never used");
+
+impl<'ast> ProjectLintPass<'ast> for UnusedError {
+    fn check_project(&mut self, ctx: &ProjectLintEmitter<'_, '_>, sources: &[ProjectSource<'ast>]) {
+        if !ctx.is_lint_enabled(UNUSED_ERROR.id()) {
+            return;
+        }
+
+        let gcx = ctx.gcx();
+        let hir = &gcx.hir;
+
+        // Map every input source's HIR `SourceId` to the corresponding `ProjectSource` index, so
+        // only errors declared in user-provided files are reported. Uses are still collected
+        // across the whole unit, so an error declared here and reverted in a dependency (or the
+        // other way around) is attributed correctly.
+        let input_source_idx: HashMap<hir::SourceId, usize> = hir
+            .sources_enumerated()
+            .filter_map(|(sid, src)| {
+                let path = match &src.file.name {
+                    FileName::Real(p) => p,
+                    _ => return None,
+                };
+                let idx = sources.iter().position(|s| &s.path == path)?;
+                Some((sid, idx))
+            })
+            .collect();
+
+        if input_source_idx.is_empty() {
+            return;
+        }
+
+        let used = collect_used_errors(hir);
+
+        // Report input-file error declarations that no expression in the unit references.
+        for error_id in hir.error_ids() {
+            let error = hir.error(error_id);
+            let Some(&src_idx) = input_source_idx.get(&error.source) else { continue };
+            // Errors declared in interfaces and abstract contracts are ABI surface meant for
+            // implementers and off-chain consumers, which may live outside the compiled sources.
+            if let Some(contract_id) = error.contract
+                && matches!(
+                    hir.contract(contract_id).kind,
+                    ContractKind::Interface | ContractKind::AbstractContract
+                )
+            {
+                continue;
+            }
+            if !used.contains(&error_id) {
+                ctx.emit(&sources[src_idx], &UNUSED_ERROR, error.span);
+            }
+        }
+    }
+}
+
+/// Collects every [`hir::ErrorId`] referenced by an expression anywhere in the unit.
+///
+/// Resolved identifiers cover almost every use: the lowering resolves the full path of
+/// `revert Lib.Err(...)` into a single `Ident`, and `require(cond, Err(...))` or `Err.selector`
+/// reference the error through a resolved `Ident` as well. The one exception is a qualified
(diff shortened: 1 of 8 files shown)


crates/lint/src/sol/info/unused_error.rs
```diff
@@ -0,0 +1,188 @@
...
+                        hir::Res::Item(hir::ItemId::Contract(contract_id)) => {
+                            scopes.push(self.hir.contract(*contract_id).items);
+                        }
+                        // A module alias exposes the file's global symbols, including the ones it
+                        // re-exports through its own imports: take the transitive closure.
+                        hir::Res::Namespace(source_id) => {
+                            for sid in self.reachable_sources(*source_id) {
+                                scopes.push(self.hir.source(sid).items);
+                            }
+                        }
```
mablr, line 196: This misses selector-only uses that go through an import alias re-export. For example, `Errors.sol` declares `error Err();`, `Reexport.sol` has `import {Err as E} from "./Errors.sol";`, and `Use.sol` imports `Reexport.sol` as `R` and returns `R.E.selector`; this compiles, but `forge lint --only-lint unused-error` reports `Err` as unused. `reachable_sources()` only exposes each source's declared items, so alias bindings are lost. Consider resolving namespace member chains through Solar's actual import/member resolution, or explicitly accounting for import aliases, and add this case to `UnusedError.sol`.

mablr (commented):
Looks pretty good ty, please have a look to the finding below :)

mattsse (commented):
I think the namespace fallback needs one more adjustment before this lands.

The expansion at [`unused_error.rs#L148-L150`](https://github.com/foundry-rs/foundry/blob/b16c6b06fc2bac79b5cc5189db00844c8f433cc4/crates/lint/src/sol/info/unused_error.rs#L148-L150), together with [`reachable_sources()`](https://github.com/foundry-rs/foundry/blob/b16c6b06fc2bac79b5cc5189db00844c8f433cc4/crates/lint/src/sol/info/unused_error.rs#L175-L186), walks raw transitive source `items`. That does not match the resolved import/export scope Solar builds during import lowering ([Solar `perform_imports`](https://github.com/paradigmxyz/solar/blob/da6b0c3a8c0edd23f8b1414c61418ba78d6c4670/crates/sema/src/ast_lowering/resolve.rs#L46-L119)). Besides missing alias re-export uses, this can also hide real unused errors by treating hidden transitive imports as namespace members.

For example, this compiles:

```solidity
// B.sol
error Lib();
contract Holder {}

// A.sol
import "./B.sol" as B;
library Lib { function f() internal pure returns (uint256) { return 1; } }

// Use.sol
import "./A.sol" as A;
contract C { function f() external pure returns (uint256) { return A.Lib.f(); } }
```

`A.Lib.f()` resolves to the library declared in `A.sol`, but the current walker also includes `B.sol`'s raw items and marks `error Lib()` as used. I checked this branch locally: linting only `B.sol` reports `unused-error`, while linting the whole project emits nothing.

Suggestion: resolve namespace/member chains from Solar's resolved source scope, or model the exact import bindings for plain imports, named aliases, and namespace aliases instead of using a transitive source list. Please also add coverage for both directions: an alias re-export use like `R.E.selector`, and a hidden transitive import whose error name collides with a real local library or contract member.

0xMars42:
Thanks, both directions confirmed and fixed in aca652b3. I reproduced your A/B/Use example on the branch first: linting the whole project was silent while linting B.sol alone reported `error Lib()`, and mablr's `R.E.selector` re-export was reported as unused.

The fallback now resolves module members through Solar's resolved source scope instead of walking raw transitive items: `base_scopes` maps a namespace base to its `SourceId` and looks member names up with `members_of` on the module type, which iterates the resolved scope, so an import alias binds under its local name to the exact declaration and hidden transitive imports are simply not visible. The `reachable_sources` walk is gone. One wrinkle: contract and library members of a module come back wrapped as `Type(Contract(..))`, while errors and nested module aliases come back bare, so the chain resolution unwraps the `Type` before descending, which keeps `NS.Lib.Err.selector` and the fixture's `Self.Errors.UsedViaChainedSelector.selector` working.

Coverage added in both directions, as auxiliary fixtures: `R.Written.selector` (alias re-export of a fixture error through `auxiliary/UnusedErrorReexport.sol`) and `H.Hidden.id()` (a hidden transitive import back to the fixture whose `library Hidden` collides with the fixture's `error Hidden()`, which is now correctly reported). Your A/B/Use example now reports `error Lib()` when linting the whole project, and linting B.sol alone still reports it.

crates/lint/src/sol/info/unused_error.rs
0xMars42: Fixed together with mattsse's point in aca652b3: the member fallback now goes through Solar's resolved source scope (`members_of` on the module type), where `{Err as E}` binds `E` to the exact declaration, so `R.E.selector` marks `Err` used. The alias re-export case is in the testdata as `R.Written.selector` through `auxiliary/UnusedErrorReexport.sol`.

mattsse (approved):
Looks good. I rechecked the namespace fallback after aca652b3, including the alias re-export and hidden transitive import cases, and the implementation now matches Solar's resolved source scope behavior.

Local verification passed:
- `cargo nextest run -p forge --test ui ui_test`

mablr (approved):
LGTM

PR #15540, opened 2026-07-02
feat(lint): add empty-block detector

0xMars42:
## Motivation

`empty-block` is one of the unchecked items of #14381: an empty body on a regular function is dead or unfinished code. This mirrors Aderyn's `empty-block` detector (`aderyn_core/src/detect/low/empty_block.rs`), restricted to function bodies.

## Solution

New `Low` severity early pass: fires on a regular function whose body is `{}` (a comment does not make a body non-empty).

Exempt, because the empty body is the behavior:

- constructors and `receive`/`fallback` (same as Aderyn): base calls, deployability, accepting ether or unknown calls;
- `virtual` functions: an empty body is the intentional default of an extension hook meant to be overridden (deviation from Aderyn, which has no such exemption);
- `payable` functions: an intentional ether sink, e.g. `function deposit() external payable {}` (deviation from Aderyn).

Functions without a body (interfaces, abstract declarations) never fire, and an empty modifier body is a solc compile error (2883), so it never reaches the linter. Empty blocks nested inside a non-empty body (`if (x) {}`) are out of scope: the checklist item is "Empty function body", while Aderyn also flags any nested empty block.

`BooleanCst.sol` had an intentionally empty test helper (`takesBool`) that the new lint flags; it now has a trivial body so the fixture stays single-lint.

## Test

`testdata/EmptyBlock.sol` covers the firing forms (free function, library function, each visibility and mutability, comment-only body, empty non-virtual override, empty body behind a modifier) and the exempt forms (constructor with a parameter, constructor with a base call, receive/fallback, virtual hook, virtual override middle hook, payable sink, bodiless declaration, nested empty `if`).

Running the detector over the repo's own `testdata/` Solidity fixtures yields 16 findings, all empty test helpers in `.t.sol` files (e.g. `doNotRevert()` in `cheats/ExpectRevert.t.sol`), none in non-test sources.

Part of #14381.

### DIFF
--- a/crates/forge/tests/cli/cmd.rs
+++ b/crates/forge/tests/cli/cmd.rs
@@ -1733,7 +1733,11 @@ forgetest!(can_fail_compile_with_warnings, |prj, cmd| {
         r"
 pragma solidity *;
 contract A {
-    function testExample() public {}
+    event Example();
+
+    function testExample() public {
+        emit Example();
+    }
 }
    ",
     );
--- a/crates/lint/src/sol/low/empty_block.rs
+++ b/crates/lint/src/sol/low/empty_block.rs
@@ -11,15 +11,20 @@ impl<'ast> EarlyLintPass<'ast> for EmptyBlock {
     fn check_item_function(&mut self, ctx: &LintContext, func: &'ast ItemFunction<'ast>) {
         // An empty body on a regular function is dead or unfinished code. Bodies whose emptiness
         // is the behavior are exempt: constructors, receive/fallback (the empty body accepts
-        // calls or ether), `virtual` functions (an empty default meant to be overridden) and
-        // `payable` functions (an empty ether sink). Functions without a body (interfaces,
-        // abstract declarations) have nothing to flag, and an empty modifier body is a solc
-        // compile error (2883), so it never reaches the linter.
+        // calls or ether), `virtual` functions (an empty default meant to be overridden),
+        // functions with modifiers (`initialize() external initializer {}`,
+        // `_authorizeUpgrade(address) internal override onlyOwner {}`: the modifier carries the
+        // behavior) and `payable` functions without return values (an empty ether sink; with a
+        // return value the body silently returns the default, an unfinished stub). Functions
+        // without a body (interfaces, abstract declarations) have nothing to flag, and an empty
+        // modifier body is a solc compile error (2883), so it never reaches the linter.
         if let Some(body) = &func.body
             && body.is_empty()
             && matches!(func.kind, FunctionKind::Function)
             && func.header.virtual_.is_none()
-            && func.header.state_mutability() != StateMutability::Payable
+            && func.header.modifiers.is_empty()
+            && (func.header.state_mutability() != StateMutability::Payable
+                || func.header.returns.is_some())
         {
             ctx.emit(&EMPTY_BLOCK, body.span);
         }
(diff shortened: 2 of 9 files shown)


crates/forge/tests/cli/cmd.rs
mablr, line 1: out of scope ?

crates/lint/src/sol/low/empty_block.rs
```diff
@@ -0,0 +1,27 @@
...
+impl<'ast> EarlyLintPass<'ast> for EmptyBlock {
+    fn check_item_function(&mut self, ctx: &LintContext, func: &'ast ItemFunction<'ast>) {
+        // An empty body on a regular function is dead or unfinished code. Bodies whose emptiness
+        // is the behavior are exempt: constructors, receive/fallback (the empty body accepts
+        // calls or ether), `virtual` functions (an empty default meant to be overridden) and
+        // `payable` functions (an empty ether sink). Functions without a body (interfaces,
+        // abstract declarations) have nothing to flag, and an empty modifier body is a solc
+        // compile error (2883), so it never reaches the linter.
+        if let Some(body) = &func.body
+            && body.is_empty()
```
mablr, line 22: This flags modifier-only functions even though modifiers can execute the meaningful behavior. Common Solidity patterns like `function _authorizeUpgrade(address) internal override onlyOwner {}` and `function initialize() external initializer {}` now warn; I reproduced both with `forge lint --only-lint empty-block`. Consider exempting functions with modifiers, or adding fixture coverage for real modifier behavior if this is intentional.

crates/lint/src/sol/low/empty_block.rs
```diff
@@ -0,0 +1,27 @@
...
+            && matches!(func.kind, FunctionKind::Function)
+            && func.header.virtual_.is_none()
+            && func.header.state_mutability() != StateMutability::Payable
```
mablr: The blanket payable exemption also suppresses empty payable functions with return values, e.g. `function payableReturns() external payable returns (uint256) {}`. That is not just an ether sink; it silently returns the default value and looks like an unfinished implementation.

crates/forge/tests/cli/cmd.rs
0xMars42: It is needed by this PR: empty-block runs during forge build, and that fixture's `function testExample() public {}` added a warning to the exact stdout snapshots of `can_fail_compile_with_warnings` (the first CI run failed on it). The body keeps the fixture lint-silent without adding solc warnings. Happy to split it into its own commit message note or a separate PR if you prefer.

crates/lint/src/sol/low/empty_block.rs
0xMars42: Fixed in 1a9be43a: functions with modifiers are exempt now, the modifier carries the behavior. The fixture covers both of your shapes: `withModifier() public noop {}` and the UUPS `_authorizeUpgrade(address) internal override noop {}`.

crates/lint/src/sol/low/empty_block.rs
0xMars42: Fixed in 1a9be43a: the payable exemption only applies without return values, so `payableReturns() external payable returns (uint256) {}` is reported (in the fixture) while `deposit() external payable {}` stays exempt.

crates/forge/tests/cli/cmd.rs
mablr: I see, thank you

mablr (approved):
LGTM

stevencartavia (approved):

PR #15541, opened 2026-07-02
feat(bench): add PR bench helpers for agents

mablr:
This adds local helpers for generating PR performance numbers from matched branch-vs-master benchmark runs. The timing helper builds one foundry-bench harness from the current checkout, points it at isolated base and candidate worktrees, and activates each checked-out Foundry build through an isolated `FOUNDRY_DIR/PATH` so the measured forge matches the intended ref.

The docs keep AGENTS.md at the policy level while keeping the operational command sequences into benches/scripts. The default `foundry-bench` local profile remains release-like for `stable/nightly` comparisons, while the PR helper opts into the `profiling` profile explicitly for branch-vs-master runs.

Authored with AI assistance.

### DIFF
--- a/benches/scripts/pr-bench.sh
+++ b/benches/scripts/pr-bench.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+BASE_REF="${BASE_REF:-origin/master}"
+CANDIDATE_REF="${CANDIDATE_REF:-HEAD}"
+BENCHMARKS="${BENCHMARKS:-forge_test}"
+REPOS="${REPOS:-ithacaxyz/account:v0.5.7}"
+RUN_ID="${RUN_ID:-$(date -u +%Y%m%d%H%M%S)}"
+BENCH_ROOT="${BENCH_ROOT:-/tmp/foundry-pr-bench-${RUN_ID}}"
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/../.." && pwd)"
+
+cd "${REPO_ROOT}"
+
+git fetch origin master
+
+mkdir -p "${BENCH_ROOT}/results"
+git worktree add --detach "${BENCH_ROOT}/master" "${BASE_REF}"
+git worktree add --detach "${BENCH_ROOT}/candidate" "${CANDIDATE_REF}"
+
+cargo build --locked --release --bin foundry-bench
+
+for label in master candidate; do
+  (
+    foundry_dir="${BENCH_ROOT}/${label}/.foundry"
+    FOUNDRY_BENCH_WORKSPACE_ROOT="${BENCH_ROOT}/${label}" \
+      FOUNDRY_BENCH_LOCAL_BUILD_PROFILE=profiling \
+      FOUNDRY_DIR="${foundry_dir}" \
+      PATH="${foundry_dir}/bin:${PATH}" \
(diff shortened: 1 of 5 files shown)


mattsse (dismissed):
LGTM

benches/scripts/pr-bench.sh
```diff
@@ -0,0 +1,42 @@
...
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/../.." && pwd)"
+
+cd "${REPO_ROOT}"
+
+git fetch origin master
+
+mkdir -p "${BENCH_ROOT}/results"
+git worktree add --detach "${BENCH_ROOT}/master" "${BASE_REF}"
+git worktree add --detach "${BENCH_ROOT}/candidate" "${CANDIDATE_REF}"
```
stevencartavia, line 31: Since the scripts default the candidate to `HEAD` and materialize it with `git worktree add --detach ... "${CANDIDATE_REF}"`, the default path benchmarks the committed `HEAD` tree and excludes any staged, unstaged, or untracked changes from the candidate checkout. For the agent flow this targets, someone who hasn't committed yet could get numbers for stale code with no warning and might paste them into the PR body.

Is the intent that these only benchmark committed revisions? If so, it might be worth failing (or warning) when the candidate resolves to the current `HEAD` commit and the tree is dirty.

mablr: Yeah I thought about it at some point; but decided to keep it simple.

> If so, it might be worth failing (or warning) when the candidate resolves to the current HEAD commit and the tree is dirty.

good idea

benches/scripts/pr-bench.sh
```diff
@@ -0,0 +1,42 @@
...
+cargo build --locked --release --bin foundry-bench
+
+for label in master candidate; do
+  (
+    foundry_dir="${BENCH_ROOT}/${label}/.foundry"
+    FOUNDRY_BENCH_WORKSPACE_ROOT="${BENCH_ROOT}/${label}" \
+      FOUNDRY_BENCH_LOCAL_BUILD_PROFILE=profiling \
+      FOUNDRY_DIR="${foundry_dir}" \
+      PATH="${foundry_dir}/bin:${PATH}" \
+      "${REPO_ROOT}/target/release/foundry-bench" \
```
stevencartavia: The scripts build with cargo and then run hardcoded default target paths (`${REPO_ROOT}/target/release/foundry-bench` here, `${REPO_ROOT}/target/release/foundry-scfuzzbench` and `${BENCH_ROOT}/${label}/target/profiling/forge` in pr-scfuzzbench.sh). If `CARGO_TARGET_DIR` is set in the environment, cargo writes the built binary under that dir instead, so the script runs a stale binary from the default path or fails with "No such file or directory".

Since the point is to benchmark exactly the binary that was just built, would it be worth setting an explicit per-script/per-worktree `CARGO_TARGET_DIR` (or deriving the paths from `${CARGO_TARGET_DIR:-...}`) so ambient config can't change which binary gets run?

stevencartavia: Minor follow-up: a relative `BENCH_ROOT` override now lands forge outside where `--foundry-bin` looks (cargo resolves the relative `CARGO_TARGET_DIR` against the worktree cwd). Normalizing `BENCH_ROOT` to absolute once would cover it. Default absolute path is unaffected, so low priority.

stevencartavia (approved):

mattsse (approved):

PR #15542, opened 2026-07-02
feat(lint): add enumerable loop removal

0xMars42:
This adds the remaining High-severity `enumerable-loop-removal` lint from #14381. It reports a semantically resolved `EnumerableSet.remove` when the same storage set is read with `at` at a bare ascending cadence and the loop body matches the conservative flow-free shape: the cadence has no unsupported writes and the body has no branches, jumps, reverts, inline assembly, `try`, or nested loops. Both method and library-qualified calls are recognized, while storage aliases, struct fields, literal mapping keys, and unknown operands are handled conservatively.

Review feedback tightened cadence invalidation across declarations, tuple assignments, resets, decrements, and deletes; restricted `at` to the bare cadence identifier; excluded every unmodeled exit shape; and documented descending, control-flow, expression-arm, and remove-argument precision limits. Permanent positive and negative fixtures cover each corrected case.

The user documentation is in foundry-rs/book#1987, with the review-aligned follow-up in foundry-rs/book#1992.


crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,131 @@
...
+    }
+}
+
+/// Collects the EnumerableSet `remove` call spans of a subtree and whether the subtree also
+/// calls `at` on an EnumerableSet (not necessarily the same instance).
+struct SetOpsCollector<'hir> {
+    gcx: Gcx<'hir>,
+    hir: &'hir Hir<'hir>,
+    removes: Vec<Span>,
+    has_at: bool,
```
mablr: Tracking only `has_at` plus all `remove` calls makes the rule warn on safe or unrelated patterns. For example, the blessed fixture warns on `while (holders.length() > 0) { value = holders.at(0); holders.remove(value); }`, which drains index 0 safely, and on `holders.remove(others.at(i))`, where removing from `holders` cannot corrupt iteration over the distinct `others` set.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn hir(&self) -> &'hir Hir<'hir> {
+        self.hir
+    }
+
+    fn visit_expr(&mut self, expr: &'hir Expr<'hir>) -> ControlFlow<Self::BreakValue> {
+        // Only method-call shapes matter: `set.remove(value)` and `set.at(index)` where the
+        // receiver is a struct declared in a library named `EnumerableSet`.
+        if let ExprKind::Call(callee, _, _) = &expr.kind
+            && let ExprKind::Member(receiver, member) = &callee.peel_parens().kind
+            && receiver_is_enumerable_set(self.gcx, self.hir, receiver)
```
mablr: This only recognizes the `using EnumerableSet for ...; set.remove(...)` method-call form. The equivalent direct library-call form, `EnumerableSet.remove(set, EnumerableSet.at(set, i))`, compiles and has the same swap-and-pop iteration bug, but the receiver here is the library identifier rather than the set struct so `receiver_is_enumerable_set` returns false.

crates/lint/docs/enumerable-loop-removal.md
```diff
@@ -0,0 +1,45 @@
...
+### Good
+
+```solidity
+address[] memory toRemove = new address[](set.length());
+for (uint256 i = 0; i < set.length(); i++) {
+    toRemove[i] = set.at(i);
+}
+for (uint256 i = 0; i < toRemove.length; i++) {
+    set.remove(toRemove[i]);
+}
```
Amxx, line 75: FYI: OpenZeppelin has a `clear` function (that is much more gas efficient has it doesnt constantly update indices like a loop or remove would)
```suggestion
set.clear();
```

0xMars42: Added in 5c5f894b, thanks: the Good example now shows `set.clear()` for the full-drain case and keeps collect-then-remove for selective removal.

crates/lint/src/sol/high/enumerable_loop_removal.rs
0xMars42: Fixed in 581b50a4: `remove` and `at` now carry the identity of the set they operate on (the variable it is stored in) and only the same-set combination reports, with two tolerances. `at` at a literal index does not count as iteration, since the swap refills the read position, so your `while` drain stays clean. An operand too complex to name a single variable (a mapping entry, a call result) conservatively matches either set. Both of your examples are in the fixture, clean (`removeWhileIteratingWhile`, `removeWithAtOnOtherSet`).

crates/lint/src/sol/high/enumerable_loop_removal.rs
0xMars42: Fixed in 581b50a4: calls are resolved through the type checker (`gcx.type_of_expr` on the callee, then `TyKind::Fn.function_id`) and matched on functions declared in a library named `EnumerableSet`, so the receiver shape no longer matters. `EnumerableSet.remove(set, EnumerableSet.at(set, i))` reports (`removeQualifiedForm` in the fixture), mixing the two forms on the same set reports too (`removeMixedForms`), an aliased import resolves the same way (`removeAliasedImport`), and a same-name function from a user library attached to the set type no longer reports (`helperRemoveIsNotEnumerable`).

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,218 @@
...
+                at_identities: Vec::new(),
+            };
+            for inner in body.stmts {
+                let _ = ops.visit_stmt(inner);
+            }
+            // Only removing from a set the same subtree iterates at a varying index corrupts
+            // that iteration: `remove` alone is the recommended collect-then-remove pattern,
+            // `at` alone is a plain read.
+            for (identity, span) in ops.removes {
+                // A removal corrupts the loop when any iterated set can be the removed one.
```
mablr: This reports a remove whenever the loop subtree contains a same-set varying `at`, but it does not account for the loop exiting immediately after the removal. That makes safe find-and-remove-once patterns false positives. I verified this with the lint UI fixture: both cases below produce an unexpected `enumerable-loop-removal` warning, even though iteration cannot continue after mutation:

```solidity
function removeTargetAndBreak(address target) public {
    for (uint256 i = 0; i < holders.length(); i++) {
        if (holders.at(i) == target) {
            holders.remove(target); // unexpected warning
            break;
        }
    }
}

function removeTargetAndReturn(address target) public {
    for (uint256 i = 0; i < holders.length(); i++) {
        if (holders.at(i) == target) {
            holders.remove(target); // unexpected warning
            return;
        }
    }
}
```

Can we avoid reporting removes on paths that immediately `break`/`return`, or otherwise only report when execution can continue the same indexed iteration after the mutation?

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,218 @@
...
+        // Every loop form lowers to `StmtKind::Loop`, so one case covers for/while/do-while.
+        if let StmtKind::Loop(body, _) = &stmt.kind {
+            let mut ops = SetOpsCollector {
+                gcx: self.gcx,
+                hir: self.hir,
+                removes: Vec::new(),
+                at_identities: Vec::new(),
+            };
+            for inner in body.stmts {
+                let _ = ops.visit_stmt(inner);
```
mablr: Because each loop collector walks every descendant statement, an outer loop aggregates independent nested loops and can turn the documented collect-then-remove safe pattern into a warning. I verified this fixture case produces an unexpected warning on the second inner loop’s `remove`:

```solidity
function collectThenRemoveInsideOuterLoop() public {
    for (uint256 round = 0; round < 1; round++) {
        address[] memory toRemove = new address[](holders.length());

        for (uint256 i = 0; i < holders.length(); i++) {
            toRemove[i] = holders.at(i);
        }

        for (uint256 i = 0; i < toRemove.length; i++) {
            holders.remove(toRemove[i]); // unexpected warning
        }
    }
}
```

This is still collect-then-remove; the `at` loop has completed before the `remove` loop starts. Can the analysis avoid treating completed descendant loops as if they were the active loop being mutated?

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,218 @@
...
+}
+
+fn is_named_at(name: solar::ast::Ident) -> bool {
+    name.as_str() == "at"
+}
+
+/// Whether two set identities can refer to the same set: equal variables do, and an
+/// expression too complex to name a single variable (`None`) can alias either set.
+fn identities_can_alias(removed: Option<VariableId>, iterated: Option<VariableId>) -> bool {
+    match (removed, iterated) {
```
mablr, line 677: The identity comparison misses simple Solidity storage aliases, which creates false negatives. A local storage reference and the state variable it aliases have different `VariableId`s, so `Some(local) != Some(state)` even though they are the same set. I verified both of these expected warnings are currently missed by the UI fixture:

```solidity
function removeViaStorageAlias() public {
    EnumerableSet.AddressSet storage set = holders;
    for (uint256 i = 0; i < set.length(); i++) {
        holders.remove(set.at(i)); // missed warning
    }
}

function iterateStateRemoveAlias() public {
    EnumerableSet.AddressSet storage set = holders;
    for (uint256 i = 0; i < holders.length(); i++) {
        set.remove(holders.at(i)); // missed warning
    }
}
```

Could we canonicalize simple `storage` aliases to their underlying storage expression, or at least treat storage-reference locals as unknown/possibly aliasing instead of distinct concrete identities?

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,218 @@
...
+/// The `n`-th argument of a call made with positional arguments.
+fn nth_positional_arg<'hir>(args: &hir::CallArgs<'hir>, n: usize) -> Option<&'hir Expr<'hir>> {
+    match &args.kind {
+        CallArgsKind::Unnamed(exprs) => exprs.get(n),
+        CallArgsKind::Named(_) => None,
+    }
+}
+
+/// The state variable a set expression refers to, the identity telling two sets apart.
+fn set_identity(set_expr: Option<&Expr<'_>>) -> Option<VariableId> {
```
mablr: Only plain identifiers get identities; member/index expressions become `None`, and `None` aliases everything. That causes false positives for definitely different storage locations. I verified these cases produce unexpected warnings:

```solidity
struct Pair {
    EnumerableSet.AddressSet a;
    EnumerableSet.AddressSet b;
}
Pair internal pair;

function removeDifferentStructField() public {
    for (uint256 i = 0; i < pair.a.length(); i++) {
        pair.b.remove(pair.a.at(i)); // unexpected warning
    }
}

mapping(uint256 => EnumerableSet.AddressSet) internal sets;

function removeDifferentMappingKey() public {
    for (uint256 i = 0; i < sets[1].length(); i++) {
        sets[2].remove(sets[1].at(i)); // unexpected warning
    }
}
```

Can we extend the identity to simple storage paths, e.g. member chains and literal mapping keys, while keeping truly dynamic expressions conservative?

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,218 @@
...
+/// expression too complex to name a single variable (`None`) can alias either set.
+fn identities_can_alias(removed: Option<VariableId>, iterated: Option<VariableId>) -> bool {
+    match (removed, iterated) {
+        (Some(removed_id), Some(iterated_id)) => removed_id == iterated_id,
+        _ => true,
+    }
+}
+
+/// The `n`-th argument of a call made with positional arguments.
+fn nth_positional_arg<'hir>(args: &hir::CallArgs<'hir>, n: usize) -> Option<&'hir Expr<'hir>> {
```
mablr: Named arguments are currently treated as having no positional operand, so qualified calls lose the set identity and become `None`/possibly-aliasing. That creates false positives for the same different-set case the fixture already keeps clean in method-call form. I verified this produces an unexpected warning:

```solidity
function namedArgsDifferentSet() public {
    for (uint256 i = 0; i < others.length(); i++) {
        EnumerableSet.remove({
            set: holders,
            value: EnumerableSet.at({set: others, index: i})
        }); // unexpected warning
    }
}
```

Can this helper resolve named args by parameter name for qualified `EnumerableSet.at/remove` instead of returning `None` for all named calls?

mattsse (dismissed):

crates/lint/src/sol/high/enumerable_loop_removal.rs
0xMars42: Fixed in c98c7de9d. A remove is no longer reported when control leaves the loop right after it. `collect_removes` carries, for each statement, whether every path out of it exits the loop (a `break` outside a nested loop, a `return`, a `revert`, or a `try` whose success clause and every `catch` exit), and a remove on such a tail is the last thing the loop does, so it corrupts nothing. `removeAndBreak` and `removeAndReturn` stay clean; `removeAndContinue` still reports (a `continue` comes round again), and a `break` that only leaves an inner loop while the analyzed one keeps iterating still reports (`removeAndBreakInnerLoop`).

crates/lint/src/sol/high/enumerable_loop_removal.rs
0xMars42: Fixed in c98c7de9d. A loop only owns the `at` reads whose index it paces itself. `AtCollector` tracks the variables the nested loops declare or write (`owned_by_inner`); an `at` whose index is one of those belongs to the inner loop, not the outer, so a completed inner `at` loop no longer makes the outer loop look like it iterates the set. `collectThenRemoveInsideOuterLoop` is clean now, while a genuine case where the inner loop reads at the outer loop's index and the outer loop removes still reports (`removeAtOuterIndexFromInnerLoop`).

crates/lint/src/sol/high/enumerable_loop_removal.rs
0xMars42: Fixed in c98c7de9d. A local `storage` reference is canonicalized to the set it was bound to: `set_path` follows a statement-local's initializer, so `EnumerableSet.AddressSet storage set = holders;` resolves `set` to `holders`. Both `removeThroughStorageAlias` and `removeThroughStorageAliasWhileIteratingTheState` report now. A reference the function binds again no longer names a single set, so it is treated conservatively (`removeThroughReassignedStorageAlias`).

crates/lint/src/sol/high/enumerable_loop_removal.rs
0xMars42: Fixed in c98c7de9d. Set identity is a storage path, not just a plain identifier: a base variable plus steps, a struct field (`Step::Field`) or a literal mapping key (`Step::Key`). `pair.first` and `pair.second` share the base but differ in their field step, `keyed[1]` and `keyed[2]` in their key step, so `removeFromAnotherStructField` and `removeAtAnotherMappingKey` are clean. A non-literal mapping key still cannot be read and stays conservative (`removeAtVaryingMappingKey` reports).

crates/lint/src/sol/high/enumerable_loop_removal.rs
0xMars42: Fixed in c98c7de9d. `nth_argument` resolves named arguments by parameter name for the qualified `EnumerableSet.at`/`remove` forms, so a named call keeps the set identity it names instead of collapsing to unknown. `removeNamedArgumentsOnOtherSet` is clean now, matching the method-call form, while a named call on the same set still reports (`removeNamedArgumentsOnSameSet`).

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,766 @@
...
+    /// Whether an `at` read at `index` advances this loop's own iteration. A fixed index does
+    /// not: the swap-and-pop refills that position, so a drain like `remove(at(0))` stays clean.
+    /// An index a nested loop owns is that loop's, not this one's, unless it is this loop's own
+    /// index that the nested loop merely reassigns. An index that cannot be read may be anything,
+    /// so it is assumed to advance.
+    fn index_advances_this_loop(&self, index: &'hir Expr<'hir>) -> bool {
+        if is_fixed_index(self.hir, index) {
+            return false;
+        }
+        !mentions_any(self.hir, index, &self.owned_by_inner)
```
mablr: This treats every nonliteral index as advancing when there is no nested loop (`owned_by_inner` is empty). A safe drain such as `while (length > position) remove(at(position))` is therefore reported, as is the existing `at(0 + 0)` case. Please require the index to depend on this loop’s cadence rather than merely not belong to an inner loop.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,766 @@
...
+                        nested,
+                        out,
+                    );
+                }
+            }
+            // Only leaving the function leaves the analyzed loop from inside a nested one.
+            hir::StmtKind::Loop(block, _) => {
+                collect_removes(gcx, hir, reassigned, block.stmts, false, true, out)
+            }
+            _ => {
```
mablr: The generic visitor applies one aggregate `leaves` value to every `try` clause. The existing `removeAndFallThroughACatch` warning is a false positive: on success the removal is followed by `return`; on failure the catch falls through but no removal occurred. Please analyze each clause with its own trailing exit state.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,766 @@
...
+    }
+
+    for (index, stmt) in stmts.iter().enumerate() {
+        let leaves = after[index] || exits_loop(stmt, nested);
+        match &stmt.kind {
+            hir::StmtKind::Block(block) | hir::StmtKind::UncheckedBlock(block) => {
+                collect_removes(gcx, hir, reassigned, block.stmts, leaves, nested, out);
+            }
+            hir::StmtKind::If(cond, then, else_) => {
+                collect_removes_in_expr(gcx, hir, reassigned, cond, leaves, out);
```
mablr: A removal used as the condition is reported even when mutation implies immediate exit. For `if (holders.remove(value)) break;`, the true path mutates and exits, while the false path continues without mutation. Please preserve this condition/branch relationship.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,766 @@
...
+        }
+        _ => None,
+    }
+}
+
+/// Whether two set operands can name the same set. Two paths that can be read name the same set
+/// exactly when they are equal; a path that cannot be read may be either.
+fn paths_alias(removed: &Option<SetPath>, iterated: &Option<SetPath>) -> bool {
+    match (removed, iterated) {
+        (Some(removed), Some(iterated)) => removed == iterated,
```
mablr, line 678: Distinct storage-reference parameters are not necessarily distinct sets. `_walkAndRemove(holders, holders)` is missed when the helper iterates one storage parameter and removes through another, because their `VariableId` bases compare unequal. This needs alias-aware parameter handling (or a documented interprocedural limitation rather than exact-disjoint treatment).

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,766 @@
...
+            }
+            self.walk_stmt(stmt)
+        }
+        fn visit_expr(&mut self, expr: &'hir Expr<'hir>) -> ControlFlow<Self::BreakValue> {
+            // Only a reassignment that advances the variable from its old value paces the loop:
+            // `i++`, `i += n`, `i = i + 1` read the previous value and progress; `i = 0` resets
+            // it each turn and does not, so a cursor a nested loop reseeds is not this loop's
+            // cadence.
+            let advanced = match &expr.kind {
+                ExprKind::Unary(op, operand)
```
mablr: Direction also matters: the safe reverse drain `for (i = holders.length(); i > 0; i--) holders.remove(holders.at(i - 1));` is reported. Removing the last element preserves the invariant and skips nothing. Please exempt this exact descending-last-element pattern or otherwise track cadence direction.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,766 @@
...
+            })?;
+            if seen.contains(&variable_id) {
+                return None;
+            }
+            seen.push(variable_id);
+            let variable = hir.variable(variable_id);
+            // A local `storage` reference is another name for the set it was bound to, unless the
+            // function binds it again, in which case what it names cannot be read here.
+            match (variable.kind, variable.initializer) {
+                (VarKind::Statement, _) if reassigned.contains(&variable_id) => None,
```
mablr: Alias invalidation is function-wide and flow-insensitive. With `alias_ = others`, a loop removing from `holders` while reading `alias_`, then `alias_ = holders` after the loop, the sets are distinct at the call but still reported as aliasing. The existing reassigned-alias case has the same issue. Please resolve the alias at the analyzed call/loop location.

crates/lint/src/sol/high/enumerable_loop_removal.rs
0xMars42: Fixed in 7fb742865. An `at` now walks a loop only when its index names the loop's own moving cadence, instead of any nonliteral index outside a nested loop. The stationary drain is clean (`removeAtStationaryCursor`), and so is the constant arithmetic case (`removeAtConstantArithmeticIndex`); the same shape with the cursor the loop advances still reports (`removeAtAdvancingStateCursor`). Since the mention is syntactic, straight-line copies carry the cadence too, so `uint256 idx = i;` does not launder the index (`removeAtCopiedIndex`, and `removeAtIndexComputedFromCadence` for a derived one, a conservative report). Requiring the cadence made the inner-loop ownership tracking and the fixed-index check redundant, so both are gone.

crates/lint/src/sol/high/enumerable_loop_removal.rs
0xMars42: Fixed in 7fb742865. The cadence now carries direction: a variable only ever stepped downward (`i--`, `i -= n`, `i = i - n`) is left out, since swap-and-pop moves the tail into the slot just read, which a descending walk never returns to. The reverse drain is clean (`removeAtDescendingIndex`, and `removeAtDescendingAssignedIndex` for the plain-assignment step); one upward step anywhere makes the direction unknown and keeps reporting (`removeAtDescendingIndexAlsoSteppedUp`).

crates/lint/src/sol/high/enumerable_loop_removal.rs
0xMars42: Fixed in 7fb742865. A removal that is itself the whole condition of an exiting branch is read through the condition/branch relationship: `remove` answers true exactly when it took the value out, so `if (set.remove(value)) break` mutates only on the path that leaves (`removeAsExitingCondition`). Only the bare call earns this: another operand riding along can steer execution past the exit after the mutation (`removeConditionCarriesAnotherOperand`), and a branch that does not leave keeps iterating on the shrunk set (`removeAsConditionWithoutExit`).

crates/lint/src/sol/high/enumerable_loop_removal.rs
0xMars42: Fixed in 7fb742865. Each `try` clause is analyzed with its own trailing exit state, and the tried call's arguments are read too. The success clause that removes and returns while the `catch` falls through empty is clean now (`removeAndFallThroughACatch`); a removal in a falling `catch` (`removeInsideFallingCatch`), a success clause that falls through (`removeAndFallThroughSuccessClause`), and one inside the tried call itself (`removeInsideTriedCall`) all report.

crates/lint/src/sol/high/enumerable_loop_removal.rs
0xMars42: Fixed in 7fb742865. Bindings are resolved where the loop runs: the walk carries what each `storage` reference last named on the straight line, so a reference rebound after the loop keeps its binding inside it (`removeThroughAliasReboundAfterTheLoop`), and the existing reassigned-alias pair resolves exactly now, `removeThroughReassignedStorageAlias` reporting because `alias_` is `holders` at the loop and `removeAnotherSetThroughReassignedStorageAlias` clean because `others` is not. A rebinding under a condition or inside the loop leaves the reference without one answer and reports (`removeThroughConditionallyReboundAlias`, `removeThroughAliasReboundInsideTheLoop`), a binding taken from another reference reads what that one named right then, so a later write to the source does not reach back (`removeThroughAliasBoundToEarlierAlias`), and a reference bound no way the walk can read, a tuple-destructured component, may name any set (`removeThroughTupleBoundAlias`).

crates/lint/src/sol/high/enumerable_loop_removal.rs
0xMars42: Documented in 7fb742865, taking the second branch you offered. The two parameters stay distinct: reading them as aliased would report every helper its callers use with two distinct sets, and judging the actual arguments would take reading every call site. The miss is recorded in the docs and pinned by a fixture (`walkOneRemoveOther` handed `holders` twice in `helperHandedTheSameSetTwiceIsMissed`); the same parameter on both sides still reports (`removeThroughSameStorageParameter`).

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,919 @@
...
+                CadenceHint::Descends => {
+                    progressions += 1;
+                    downward += 1;
+                }
+                CadenceHint::Advances => progressions += 1,
+                CadenceHint::Guards => {}
+            }
+        }
+        // No progression read here means the direction is unknown, so the variable is kept.
+        if progressions == 0 || downward < progressions {
```
mablr: This blanket descending-cadence exemption creates a false negative when the loop removes a different value. For `for (i = set.length(); i > 0; i--) { set.at(i - 1); set.remove(target); }`, removing an earlier slot moves the already-read tail into an unvisited slot, but no warning is emitted. Please exempt only a proven reverse-drain relationship rather than every descending walk.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,919 @@
...
+    let mut after = vec![tail_exits; stmts.len()];
+    for index in (0..stmts.len()).rev() {
+        after[index] = match stmts.get(index + 1) {
+            Some(next) => after[index + 1] || exits_loop(next, nested),
+            None => tail_exits,
+        };
+    }
+
+    for (index, stmt) in stmts.iter().enumerate() {
+        let leaves = after[index] || exits_loop(stmt, nested);
```
mablr: An eventual exit is not enough to make the removal safe: `remove(target); at(i); break;` can read a shifted slot or go out of bounds before leaving, yet `leaves` suppresses the warning. Please preserve ordering and only exempt when no same-set cadence read is reachable before the exit.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,919 @@
...
+/// Whether a plain assignment steps its target downward: the right-hand side subtracts from the
+/// target itself, `i = i - 1`. Anything else that reads the target, `i = i + 1` or `i = j - i`,
+/// is not a downward step.
+fn subtracts_from_target<'hir>(
+    hir: &'hir Hir<'hir>,
+    target: &'hir Expr<'hir>,
+    value: &'hir Expr<'hir>,
+) -> bool {
+    let ExprKind::Binary(minuend, op, _) = &value.peel_parens().kind else { return false };
+    op.kind == hir::BinOpKind::Sub && mentions_target(hir, target, minuend)
```
mablr: `mentions_target` is too permissive for direction: `i = (i + 2) - 1` advances by one but is classified as descending, so `remove(at(i))` is missed. The minuend needs to be exactly the target (after parentheses), not merely contain it.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,919 @@
...
+                // A compound assignment (`+=`, `-=`, ...) reads the old value.
+                ExprKind::Assign(lhs, Some(op), _) => Some((*lhs, op.kind == hir::BinOpKind::Sub)),
+                // A plain assignment progresses only when it reads the target itself.
+                ExprKind::Assign(lhs, None, rhs) if mentions_target(self.hir, lhs, rhs) => {
+                    Some((*lhs, subtracts_from_target(self.hir, lhs, rhs)))
+                }
+                _ => None,
+            };
+            if let Some((progressed, descends)) = progressed
+                && let ExprKind::Ident(resolutions) = &progressed.peel_parens().kind
```
mablr: Restricting cadence targets to bare identifiers misses state/member cursors. `while (walk.index < set.length()) { set.remove(set.at(walk.index)); walk.index++; }` corrupts iteration but emits no warning; indexed cursors have the same issue. Cadence identity needs an access path rather than only `VariableId`.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,919 @@
...
+            // before any clause is dispatched.
+            hir::StmtKind::Try(try_) => {
+                collect_removes_in_expr(gcx, hir, bindings, &try_.expr, leaves, out);
+                for clause in try_.clauses {
+                    collect_removes(gcx, hir, bindings, clause.block.stmts, leaves, nested, out);
+                }
+            }
+            // Only leaving the function leaves the analyzed loop from inside a nested one.
+            hir::StmtKind::Loop(block, _) => {
+                collect_removes(gcx, hir, bindings, block.stmts, false, true, out)
```
mablr: Hard-coding `tail_exits = false` reports a safe nested removal when the outer loop exits immediately afterward: `at(i); for (...) remove(value); break;`. The nested loop has no relevant read/backedge and the outer loop never comes around, so this is a false positive.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,919 @@
...
+                ExprKind::Unary(op, operand)
+                    if matches!(
+                        op.kind,
+                        UnOpKind::PreInc | UnOpKind::PreDec | UnOpKind::PostInc | UnOpKind::PostDec
+                    ) =>
+                {
+                    Some((*operand, matches!(op.kind, UnOpKind::PreDec | UnOpKind::PostDec)))
+                }
+                // A compound assignment (`+=`, `-=`, ...) reads the old value.
+                ExprKind::Assign(lhs, Some(op), _) => Some((*lhs, op.kind == hir::BinOpKind::Sub)),
```
mablr: No-op assignments are treated as progression. `while (set.length() > i) { set.remove(set.at(i)); i += 0; }` safely drains one fixed slot but is warned (likewise `i = i`). Please exclude the exact no-op forms from cadence hints.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,919 @@
...
+    init: &'hir [Stmt<'hir>],
+    body: &'hir [Stmt<'hir>],
+    mut carriers: Vec<VariableId>,
+) -> Vec<VariableId> {
+    if carriers.is_empty() {
+        return carriers;
+    }
+    let mut copies = Vec::new();
+    collect_copies(hir, init, &mut copies);
+    collect_copies(hir, body, &mut copies);
```
mablr: Flow-insensitive copy propagation reports fixed-slot drains when a later cadence assignment is killed before every read. In `for (...; ...; i++) { idx = 0; remove(at(idx)); idx = i; }`, every `at` sees zero, but the later assignment globally marks `idx` as moving. This needs statement-order reaching definitions, at least for definite resets.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,919 @@
...
+                collect_removes(gcx, hir, bindings, block.stmts, leaves, nested, out);
+            }
+            hir::StmtKind::If(cond, then, else_) => {
+                // `remove` answers true exactly when it took the value out, so a removal that
+                // is itself the whole condition corrupts nothing when the mutating answer
+                // always leaves the loop: `if (set.remove(value)) break` shrinks the set only
+                // on the exiting path, and the continuing path left it untouched. Any other
+                // operand riding along (`remove(v) && flag`) can steer execution past the exit
+                // after the set shrank, so only the bare call is read this way.
+                let removal_implies_exit = exits_loop(then, nested)
```
mablr: The exit exemption only handles positive polarity. `if (!set.remove(value)) { } else { break; }` mutates only on the exiting path and is safe, but currently warns. Please normalize a unary `!` and associate the successful removal with the corresponding branch.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,1611 @@
...
+        next = flow_before_stmt(context.hir, &stmts[index], next, targets, context.ats);
+    }
+
+    for (index, stmt) in stmts.iter().enumerate() {
+        match &stmt.kind {
+            hir::StmtKind::Block(block) | hir::StmtKind::UncheckedBlock(block) => {
+                collect_removes(context, block.stmts, after[index].clone(), targets, out);
+            }
+            hir::StmtKind::If(cond, then, else_) => {
+                // `remove` answers true exactly when it took the value out, so a removal that
```
mablr: This still warns when `remove` is nested in a short-circuit condition whose successful result necessarily takes the exiting branch. For example, `if (enabled && holders.remove(target)) break;` is safe: if `remove` mutates, the conjunction is true and the loop exits; otherwise the set is unchanged. Please extend the polarity/control-flow handling beyond a bare call without losing the existing unsafe `remove(target) && flag` case, where `flag` can suppress the exit after mutation.

mablr (changes requested):

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,1611 @@
...
+        // control leaves first.
+        let mut removes = Vec::new();
+        let removal_context = RemovalContext {
+            gcx: self.gcx,
+            hir: self.hir,
+            bindings: &self.bindings,
+            ats: &ats.iterated,
+        };
+        let targets = FlowTargets::outer_loop();
+        collect_removes(&removal_context, body, FlowTail::default(), &targets, &mut removes);
```
figtracer: this still reports the canonical safe swap-and-pop walk where `i` advances only when no removal occurs; keeping `i` fixed after removal visits the swapped-in value next. please correlate progression with the removal branch or omit until the flow analysis can distinguish it.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,1611 @@
...
+        for removal in removes {
+            let corrupts = ats.iterated.iter().any(|iterated| {
+                paths_alias(&removal.set, &iterated.set)
+                    && !(iterated.descending && removal.removed_at == Some(iterated.span))
```
mablr: The descending-loop exemption also accepts non-tail removals. For example, `for (uint i = set.length(); i > 1; i--) set.remove(set.at(i - 2));` emits no warning, but swap-and-pop moves the tail into a slot this walk will not revisit. Please require the removed element to be the current tail, not merely a descending `at(...)` expression.

0xMars42 (commented):
Thanks for the thorough passes. Each round I added a layer (cadence direction, copy propagation, an exit-and-branch exemption) to keep a report while dodging the last false positive, and each layer grew the next false negative you then found. Rather than add a fourth, I reworked the detector around a single rule it can judge without a flow analysis, pushed in 9db3decca.

Reported now: an unconditional ascending cadence (`i++`, `i += 1`, `i = i + 1`), an `at` read at that cadence, a `remove` on the same set, and a straight-line body (no `if`, `try`, `break`, `continue`, `return`, or nested loop). The type-checker call resolution and the storage-path set identity from the earlier rounds are kept. The synthetic condition guard the lowering wraps every loop in is peeled so its `break` and the `for`'s next-step read for what they are.

This drops every false positive you raised, by construction:

- swap-and-pop filter (`i++` only in the no-removal branch): the cadence is under a branch, so it is not an unconditional ascending cadence.
- short-circuit exit (`remove(t) && flag` then `break`) and negated exit (`!remove(v)` then `break`): the body has a `break`.
- no-op step (`i += 0`): not a positive step, so no ascending cadence.
- killed copy (`idx = i` reassigned before the read): copy propagation is gone; the index must be the cadence variable.
- nested removal then outer exit: a nested loop is not straight-line.

The false negatives you flagged become documented limitations, in the docs and the fixture: the descending drain of a different value, an eventual exit reading a shifted slot, composite index arithmetic, and a member/opaque cursor are all cases where telling the corruption from a safe lookalike needs the flow analysis this pass does not run, so it stays silent instead of guessing. The canonical ascending remove-as-you-go is still reported.

This is the conservative-but-precise version on the control-flow axis: it never warns on the safe swap-and-pop and drain-and-exit forms it can read from the loop's structure, and in exchange it misses the corruptions that hide behind control flow. It can still over-report on the set-identity axis, unchanged from the earlier rounds: an operand it cannot read statically (a varying mapping key, a conditionally-rebound alias) is matched against every set, reporting a possibly-safe removal rather than missing an unsafe one.

If you would rather have the control-flow corruptions caught and accept the flow analysis it takes (reaching definitions for the cadence, correlating the mutating path with the exiting branch, polarity, short-circuit), I am happy to build that out instead. Your call on which trade you want.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,674 @@
...
+    stmts: &'hir [Stmt<'hir>],
+    out: &mut Vec<VariableId>,
+) {
+    for stmt in stmts {
+        match &stmt.kind {
+            StmtKind::Block(block) | StmtKind::UncheckedBlock(block) => {
+                collect_ascending_steps(hir, block.stmts, out);
+            }
+            StmtKind::Expr(expr) => {
+                if let Some(variable_id) = ascending_step(expr.peel_parens())
```
mablr: A single recognized increment is enough to keep a variable in `cadence`; later writes never disqualify it. The UI fixture reports both `remove(at(i)); i++; i--;` and `remove(at(i)); i++; i = 0`, although each safely drains fixed slot 0. Please reject a cadence when the body also writes it with a decrement/reset (as the contract above says).

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,674 @@
...
+
+    fn hir(&self) -> &'hir Hir<'hir> {
+        self.hir
+    }
+
+    fn visit_expr(&mut self, expr: &'hir Expr<'hir>) -> ControlFlow<Self::BreakValue> {
+        if let Some(call) = enumerable_set_call(self.gcx, self.hir, self.bindings, expr)
+            && call.name == SetOp::At
+            && nth_argument(self.hir, call.function_id, call.args, call.index_arg, INDEX_PARAMETER)
+                .is_none_or(|index| mentions_any(self.hir, index, self.cadence))
```
mablr: `mentions_any` accepts algebraically constant indices, not just an index paced by the loop. I verified `while (length > 0) { remove(at(i % 1)); i++; }` warns even though it safely drains slot 0. Requiring the `at` argument to be the bare cadence identifier would match the documented `at(i)` shape and leave composite arithmetic unreported as promised.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,674 @@
...
+        | StmtKind::Loop(..)
+        | StmtKind::Break
+        | StmtKind::Continue
+        | StmtKind::Return(..) => false,
+        // `revert Foo();`, `revert(...)`, `require(...)`: a revert leaves the loop too. Detected
+        // as an expression statement of a revert or a call that never returns is beyond a plain
+        // structural check, so only the structural jumps above are handled; a body reverting
+        // unconditionally still corrupts nothing, but that is a rare shape and staying silent on
+        // it is safe.
+        _ => true,
```
mablr: This fallback still classifies `revert();` as straight-line fallthrough. The UI fixture warns on `remove(at(i)); revert();`, but the mutation is rolled back and no next iteration is observable. Please make an unconditional revert disqualify the supported body shape; the comment currently says this case stays silent, while `_ => true` does the opposite.

crates/lint/docs/enumerable-loop-removal.md
```diff
@@ -0,0 +1,75 @@
...
+
+These stay clean:
+
+- `remove` in a loop without `at` is the recommended collect-then-remove pattern;
+- `at` in a loop without `remove` is a plain read;
+- `remove` outside a loop is fine;
+- `at` on a different set instance, including two struct fields or two literal mapping keys;
+- `at` at an index the loop never advances (`remove(at(0))`, a `constant`, a stationary cursor, a no-op step like `i += 0`): the swap refills the read position;
+- `at` at a cadence stepped downward (`for (i = set.length(); i > 0; i--) set.remove(set.at(i - 1))`): swap-and-pop moves the tail into the emptied slot, which a descending walk never revisits;
```
mablr: This safety statement only holds when the removed value is the current tail. I verified the detector misses `for (i = length; i > 0; i--) { at(i - 1); remove(target); }`: removing an earlier slot moves the already-read tail into an unvisited slot, so the walk can duplicate/skip values. Since descending cadences are intentionally outside the new rule, please list this as deliberately unreported and narrow the absolute precision/recall claim below.

crates/lint/src/sol/high/enumerable_loop_removal.rs
mattsse: Cadence candidates are now invalidated by every unsupported write, including decrement, reset, declaration, tuple assignment, and delete targets. The fixed-slot drain regressions are covered in the fixture.

crates/lint/src/sol/high/enumerable_loop_removal.rs
mattsse: The at index must now be the bare cadence identifier. Composite and algebraically constant expressions such as i % 1 remain deliberately unreported, with regression coverage.

crates/lint/src/sol/high/enumerable_loop_removal.rs
mattsse: Unconditional Solidity reverts now disqualify the supported straight-line shape. Inline assembly is also excluded so unmodeled Yul exits cannot be mistaken for fallthrough; both cases have regression coverage.

crates/lint/docs/enumerable-loop-removal.md
mattsse: The documentation now treats descending non-tail removals as deliberately unreported and narrows the precision claim to the supported ascending shape. The separate Book follow-up uses the same wording.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,681 @@
...
+            cadence: &cadence,
+            iterated: Vec::new(),
+        };
+        for stmt in body {
+            let _ = ats.visit_stmt(stmt);
+        }
+        if ats.iterated.is_empty() {
+            return;
+        }
+        // Every syntactic removal in the body, including calls nested in short-circuit or ternary
```
mablr: This structural scan reports removals that provably cannot execute. I verified that `for (...; i++) { holders.at(i); false && holders.remove(...); }` produces an unexpected warning. The same applies to a compile-time-false ternary arm. Please at least fold literal reachability before treating every expression arm as a removal; otherwise this high-severity lint flags dead code.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,681 @@
...
+        ExprKind::Assign(lhs, Some(op), rhs)
+            if op.kind == hir::BinOpKind::Add && is_positive_literal(rhs) =>
+        {
+            bare_identifier(lhs)
+        }
+        // `i = i + <positive literal>`.
+        ExprKind::Assign(lhs, None, rhs) => {
+            let target = bare_identifier(lhs)?;
+            let ExprKind::Binary(left, op, right) = &rhs.peel_parens().kind else { return None };
+            (op.kind == hir::BinOpKind::Add
```
mablr, line 358: This recognizes only `i = i + 1`, missing the equivalent unconditional ascending cadence `i = 1 + i`. I verified `for (uint i = 0; i < set.length(); i = 1 + i) set.remove(set.at(i));` emits no warning. Since integer addition is commutative here and the loop has the exact corruption this lint targets, please recognize a positive literal on either side.

mattsse: Fixed in 928f6988e. The cadence matcher now accepts the commuted i = 1 + i form, with an ascending-loop regression and matching documentation. Full UI/clippy/fmt validation and a fresh-context re-review pass.

crates/lint/src/sol/high/enumerable_loop_removal.rs
mattsse: Fixed in 928f6988e. Expression traversal now prunes literal-proven dead &&, ||, and ternary arms before recording at/remove calls, with positive controls for reachable arms. The UI suite, clippy, fmt, and fresh-context re-review pass; the public lint docs were updated to describe the exception.

crates/lint/src/sol/high/enumerable_loop_removal.rs
```diff
@@ -0,0 +1,727 @@
...
+        StmtKind::If(..)
+        | StmtKind::Try(..)
+        | StmtKind::Loop(..)
+        | StmtKind::AssemblyBlock(..)
+        | StmtKind::Break
+        | StmtKind::Continue
+        | StmtKind::Return(..)
+        | StmtKind::Revert(..) => false,
+        StmtKind::Expr(expr) if is_builtin_revert(expr) => false,
+        _ => true,
```
mablr: `require(false)` and `assert(false)` fall through this default, so a loop that removes, increments, then unconditionally fails is warned even though the removal rolls back and no next iteration is reachable. This conflicts with the documented non-continuing-body exemption; please reuse the existing `branch_always_exits` handling for terminal statement forms rather than recognizing only explicit reverts.

mattsse: Fixed in 796597a6f. Enumerable-loop-removal now uses the shared terminal-flow analysis, so require(false), assert(false), revert, and parenthesized exit calls stop impossible next-iteration diagnostics. Added true-form warning controls, updated docs, and preserved parenthesis peeling in the shared helper. Full UI/clippy/fmt validation and final fresh-context re-review pass.

mablr (dismissed):
LGTM

stevencartavia (dismissed):

stevencartavia:
@0xMars42 just needs conflicts

mablr (approved):

stevencartavia (approved):

PR #15544, opened 2026-07-02
feat(symbolic): target fuzz branch frontiers

figtracer:
### Summary

Fuzz frontier artifacts can contain many branch sites, but solver time is limited; these filters let users focus symbolic seeding on specific high-value frontier records by artifact ID, comparison PC, or calldata selector instead of spending the budget strictly by artifact order.

- Add `symbolic.frontier_ids`, `symbolic.frontier_pcs`, and `symbolic.frontier_selectors` plus matching CLI flags for targeted fuzz frontier imports.
- Compose non-empty frontier filters conjunctively, then preserve artifact-order priority under `symbolic.frontier_limit`.
- Emit a frontier selection status line with imported/skipped counts and warn when requested IDs, PCs, or selectors cannot be imported.
- Document how frontier targeting works with the JSON frontier artifact fields.

### DIFF
--- a/crates/config/src/providers/warnings.rs
+++ b/crates/config/src/providers/warnings.rs
@@ -55,6 +55,8 @@ const SYMBOLIC_KEYS: &[&str] = &[
     "use_fuzz_frontiers",
     "frontier_limit",
     "frontier_ids",
+    "frontier_pcs",
+    "frontier_selectors",
     "solver",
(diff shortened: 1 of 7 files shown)


mattsse (approved):
lgtm

mablr (approved):
sgtm

PR #15546, opened 2026-07-02
chore(anvil): remove unused state hook

stevencartavia:
Removes the `state_hook` field and its `set_state_hook` setter from `AnvilBlockExecutor`. The setter had no callers, so the field was always `None` and both `on_state` branches were unreachable. Dead scaffolding left over from the `debug_accountInfoAt` work; anvil reads account/state directly after replay and has no `OnStateHook` consumer.


mattsse (approved):
LGTM

mablr (approved):

PR #15547, opened 2026-07-02
perf(symbolic): skip local unsat facts

figtracer:
Extends the symbolic solver's direct contradiction pass with local unsigned range facts and bounded wrap/product checks so common unsat branches can return before emitting SMT.

Bug suite (`grandizzy/symbolic-bug-suite`, 22/22 expected failures):
- SMT queries: 95 -> 70
- paired solver time avg: 1345.5ms -> 1126.4ms
- hyperfine: 452.0ms +/- 9.1ms -> 438.6ms +/- 8.2ms

`foundry-bench --benchmarks forge_symbolic_test`:
- Solady SMT: 70 -> 60, solver 409ms -> 352ms
- Angstrom unchanged SMT: 2 -> 2
- Farcaster SMT: 35 -> 25, solver 183ms -> 138ms

### DIFF
--- a/crates/evm/symbolic/src/runtime/solver/opt.rs
+++ b/crates/evm/symbolic/src/runtime/solver/opt.rs
@@ -167,14 +167,18 @@ const fn expr_ternop_key(op: SymTernOp) -> u8 {
 }
 
 /// Returns whether normalized conjunctive constraints contain a direct contradiction.
-pub(super) fn constraints_are_directly_unsat(cx: &mut SymCx, constraints: &[SymBoolExpr]) -> bool {
-    if constraints.iter().any(|constraint| match constraint.kind() {
-        SymBoolExprKind::Const(false) => true,
-        SymBoolExprKind::Not(inner) => constraints.contains(inner),
-        _ => {
-            let negated = constraint.clone().not(cx);
-            constraints.contains(&negated)
-        }
+pub(super) fn constraints_are_directly_unsat(constraints: &[SymBoolExpr]) -> bool {
+    if constraints
+        .iter()
+        .any(|constraint| matches!(constraint.kind(), SymBoolExprKind::Const(false)))
+    {
+        return true;
+    }
+
+    let constraint_set: HashSet<_> = constraints.iter().collect();
+    if constraints.iter().any(|constraint| {
+        let SymBoolExprKind::Not(inner) = constraint.kind() else { return false };
+        constraint_set.contains(inner)
     }) {
         return true;
     }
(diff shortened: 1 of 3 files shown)


mattsse (dismissed):
lgtm

crates/evm/symbolic/src/runtime/solver/opt.rs
```diff
@@ -168,14 +168,332 @@ const fn expr_ternop_key(op: SymTernOp) -> u8 {
...
 /// Returns whether normalized conjunctive constraints contain a direct contradiction.
 pub(super) fn constraints_are_directly_unsat(cx: &mut SymCx, constraints: &[SymBoolExpr]) -> bool {
-    constraints.iter().any(|constraint| match constraint.kind() {
+    if constraints.iter().any(|constraint| match constraint.kind() {
         SymBoolExprKind::Const(false) => true,
         SymBoolExprKind::Not(inner) => constraints.contains(inner),
         _ => {
             let negated = constraint.clone().not(cx);
             constraints.contains(&negated)
         }
```
mablr: This first contradiction pass still does linear `constraints.contains(...)` lookups inside the per-constraint loop, so each solver query can pay O(n^2) before reaching the new fact collector. Since this path is specifically a pre-SMT performance fast path, maybe an `HashSet<&SymBoolExpr>` built once and used for membership checks could be relevant?

crates/evm/symbolic/src/runtime/solver/opt.rs
```diff
@@ -168,14 +168,332 @@ const fn expr_ternop_key(op: SymTernOp) -> u8 {
 
 /// Returns whether normalized conjunctive constraints contain a direct contradiction.
 pub(super) fn constraints_are_directly_unsat(cx: &mut SymCx, constraints: &[SymBoolExpr]) -> bool {
-    constraints.iter().any(|constraint| match constraint.kind() {
+    if constraints.iter().any(|constraint| match constraint.kind() {
         SymBoolExprKind::Const(false) => true,
         SymBoolExprKind::Not(inner) => constraints.contains(inner),
         _ => {
             let negated = constraint.clone().not(cx);
```
mablr: The `_` arm allocates `constraint.clone().not(cx)` for every non-`Not` constraint just to check membership. On satisfiable paths these fresh negation nodes are discarded but remain in `SymCx` for the executor lifetime, so repeated solver calls can grow the arena without learning anything. Consider checking only existing `Not(inner)` constraints against a set of positive constraints, or otherwise avoiding synthetic negation allocation for the membership probe.

mablr (commented):
perf questions

Issue #15545, opened 2026-07-02
Panic revert in the code when a revert string is 3 bytes


ManInBlackie:
### Component

Forge

### Have you ensured that all of these are up to date?

- [ ] Foundry
- [ ] Foundryup

### What version of Foundry are you on?

1.5.0-stable

### What version of Foundryup are you on?

foundryup: 1.3.0

### What command(s) is the bug in?

forge test --fork-url "$RPC_URL"

### Operating System

macOS (Apple Silicon)

### Describe the bug

vm.expectRevert(bytes("C38")). "C38" is exactly 3 bytes, and newer Foundry tries to decode the expected value as a 4-byte error selector → slices [0..4] of a length-3 slice → the panic.

Location: /Users/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/alloy-dyn-abi-1.4.1/src/dynamic/error.rs:85

This is a bug. Consider reporting it at https://github.com/foundry-rs/foundry

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                                ⋮ 8 frames hidden ⋮                               
   9: core::slice::index::slice_index_fail::do_panic::runtime::hdb76a588098e04d1
      at <unknown source file>:<unknown line>
  10: core::slice::index::slice_index_fail::hcba6beb3c7e54df7
      at <unknown source file>:<unknown line>
  11: <alloy_json_abi::item::Error as alloy_dyn_abi::ext::error::ErrorExt>::decode_error::h990ce6f007b5cac9
      at <unknown source file>:<unknown line>
  12: foundry_cheatcodes::test::revert_handlers::handle_expect_revert::h4511e073eac6d1a9
      at <unknown source file>:<unknown line>
  13: <foundry_cheatcodes::inspector::Cheatcodes as revm_inspector::inspector::Inspector<revm_context::context::Context<revm_context::block::BlockEnv,revm_context::tx::TxEnv,revm_context::cfg::CfgEnv,&mut dyn foundry_evm_core::backend::DatabaseExt+Error = foundry_fork_db::error::DatabaseError>>>::call_end::h7e63e9056894740a
      at <unknown source file>:<unknown line>
[...]

Issue #13766, opened 2026-03-15
panic in `expectRevert`


bbjubjub2494:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

1.6.0-nightly d055c0999b23e2b045acba0e402f83d1e1e20496

### What version of Foundryup are you on?

1.5.0

### What command(s) is the bug in?

forge test

### Operating System

Linux

### Describe the bug

When `vm.expectRevert(bytes(""))` catches a revert with non-empty calldata, `slice_index_fail` is triggered. Reproduced in nightly.

POC: https://github.com/bbjubjub2494/foundry_crash_poc_20260315

```
The application panicked (crashed).
Message:  range end index 4 out of range for slice of length 0
Location: /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/alloy-dyn-abi-1.5.7/src/dynamic/error.rs:85

This is a bug. Consider reporting it at https://github.com/foundry-rs/foundry

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                                ⋮ 7 frames hidden ⋮
   8: core::slice::index::slice_index_fail::h172ece6e023ae9aa
      at <unknown source file>:<unknown line>
   9: <alloy_json_abi::item::Error as alloy_dyn_abi::ext::error::ErrorExt>::decode_error::h817fc33ad2337c9d
      at <unknown source file>:<unknown line>
  10: foundry_cheatcodes::test::revert_handlers::handle_expect_revert::h75042fca2d07a8d7
      at <unknown source file>:<unknown line>
  11: <foundry_cheatcodes::inspector::Cheatcodes as revm_inspector::inspector::Inspector<CTX>>::call_end::h7076dbe2ed9a9e36
      at <unknown source file>:<unknown line>
  12: <foundry_evm::inspectors::stack::InspectorStackRefMut as revm_inspector::inspector::Inspector<CTX>>::call_end::h9b0703e6fa1e11ab
      at <unknown source file>:<unknown line>
  13: revm_inspector::handler::frame_end::hce535e4ef4a3fe8d
      at <unknown source file>:<unknown line>
[...]

PR #15548, opened 2026-07-02
test(cheatcodes): cover short expectRevert reason

stevencartavia:
Strengthen the #13766 regression test to also cover a non-empty expected reason shorter than an ABI selector (`vm.expectRevert(bytes("C38"))`), the exact case from #15545. Fixed by #13769; this just adds coverage so the short-reason mismatch fails cleanly instead of panicking.


mattsse (approved):
LGTM

mablr (approved):

PR #15549, opened 2026-07-02
feat(forge): add fuzz tmin subcommand

mablr:
Implements `forge fuzz tmin`, closing OSS-297 and following the initial `forge fuzz` subcommand PR #15227 and the `cmin` PR #15517.

`tmin` minimizes replay corpus entries while preserving the original failure identity or, for non-failing inputs, the same hit/no-hit edge set while allowing AFL bucket-count changes. It supports both single files and corpus directories, so the full `cmin -> tmin -> replay/show/showmap` workflow works on real corpus trees.

It removes redundant transactions first, then applies deterministic ABI-aware calldata simplification. Regression coverage includes edge preservation, failure preservation, directory minimization, gzip output, and attempt-budget handling.


mattsse (dismissed):
LGTM

stevencartavia (dismissed):

crates/forge/src/cmd/fuzz.rs
```diff
@@ -390,6 +401,289 @@ struct CminSummary {
...
+        let attempts =
+            minimize_entry(&session, &decoder, &self.input, &mut sequence, self.max_attempts)?;
+        write_sequence_create_new(&self.out, &sequence)?;
+
+        sh_println!(
+            "minimized entry: {before_txs} txs -> {} txs in {}",
+            sequence.len(),
+            self.out.display()
+        )?;
+        sh_println!("attempted {attempts} candidate replays")?;
```
stevencartavia: `attempted {attempts} candidate replays` uses `sh_println!` (stdout). It's diagnostic, so `sh_status!` (stderr) would match `cmin`'s stdout/stderr split.

crates/forge/src/cmd/fuzz.rs
```diff
@@ -390,6 +401,289 @@ struct CminSummary {
...
+        let mut total_entries = 0usize;
+        let mut before_txs = 0usize;
+        let mut after_txs = 0usize;
+        let mut attempts = 0usize;
+
+        for entry in entries {
+            let mut sequence = entry
+                .read_tx_seq()
+                .with_context(|| format!("failed to read corpus entry {}", entry.path.display()))?;
+            if sequence.is_empty() {
```
stevencartavia, line 477: In `run_dir`, one empty/unreadable entry aborts the whole run, while `cmin` skips + counts it. Could mirror `cmin` for raw corpora.

stevencartavia (commented):
a couple optional nits

crates/forge/src/cmd/fuzz.rs
```diff
@@ -390,6 +401,289 @@ struct CminSummary {
...
+        }
+        self.attempts += 1;
+        let observations =
+            replay_candidate(self.session, self.evm_edge_indices.clone(), candidate.to_vec())?;
+        let observations = observations
+            .into_iter()
+            .map(|obs| (obs.target, obs.observation))
+            .collect::<BTreeMap<_, _>>();
+
+        for (target, baseline) in &self.baseline.requirements {
```
grandizzy, line 623: **Acceptance predicate ignores newly-activated targets (blocking).**

This only validates targets present in `baseline.requirements`, but `replay_baseline` drops any target with `replayed==0 && no failure && no coverage`. So a candidate can activate a target that was absent from the baseline and it goes unchecked.

Repro: `tmin --mc '(A|B)' --mt testFuzz_same` on `testFuzz_same(42)` where `B` skips 42 via `vm.assume` (dropped from baseline). The ABI pass tries `0` first; if `B.testFuzz_same(0)` now reverts, tmin accepts it because only `A` is checked → corrupts the identity we claim to preserve.

```suggestion
        for (target, observation) in &observations {
            if !self.baseline.requirements.contains_key(target)
                && (observation.failure.is_some() || observation.replayed > 0 || has_edges(observation))
            {
                return Ok(false);
            }
        }

        for (target, baseline) in &self.baseline.requirements {
```

crates/forge/src/cmd/fuzz.rs
```diff
@@ -401,6 +695,287 @@ fn temporary_cmin_out(out: &Path) -> Result<TempDir> {
...
+}
+
+fn same_edge_set(candidate: &[u8], baseline: &[u8]) -> bool {
+    let len = candidate.len().max(baseline.len());
+    (0..len).all(|idx| {
+        let candidate_hit = candidate.get(idx).copied().unwrap_or_default() != 0;
+        let baseline_hit = baseline.get(idx).copied().unwrap_or_default() != 0;
+        candidate_hit == baseline_hit
+    })
+}
```
grandizzy: **`"exact baseline edge set"` only compares hit/no-hit, not AFL buckets.**

cmin's `merge_new_edges` treats bucket-count increases as new coverage, but this compares nonzero-ness only. tmin can shrink a 4-iteration loop input to 1 iteration (same edge IDs, different buckets) and erase a feature cmin deliberately kept — while the description says it preserves the "exact edge set."

Either compare bucket vectors exactly:

```rust
candidate.get(idx).copied().unwrap_or_default() == baseline.get(idx).copied().unwrap_or_default()
```

…or rename this to hit-set preservation and add a test asserting bucket changes are intentionally allowed.

grandizzy (changes requested):
Thanks for the follow-up to #15227 / #15517 — the `tmin` implementation is well-structured and the test coverage is thorough. Requesting changes on a couple of items; please check findings below.

crates/forge/src/cmd/fuzz.rs
mablr: https://github.com/foundry-rs/foundry/pull/15549/commits/470658f69b1de6ad0ea7ad3aeb172eeeb065bd32

grandizzy (dismissed):
lgtm

mablr (commented):
Sorry @grandizzy, just pushed the optional fixes for Steven's comments

grandizzy (approved):

stevencartavia (approved):

PR #15551, opened 2026-07-02
feat(lint): add unsafe ERC721 mint

0xMars42:
Adds the medium-severity `unsafe-oz-erc721-mint` lint tracked in #14381. The detector resolves calls to canonical OpenZeppelin ERC721 mint implementations, follows transitive user overrides, and exempts a call only when the full delegated path proves the recipient has no code or performs a correlated receiver callback that cannot be bypassed.

The takeover addresses the existing review feedback around recursive identity forwarding, accepted-branch mutation, modifier ordering, transitive assembly exits, selector conversions, path-scoped recursion, and code-less recipient guards. Recipient-only proofs remain valid across computed or remapped token IDs, while callback guards still correlate both recipient and token. Fresh-context review additionally hardened guard-condition mutation, helper parameter remapping, cross-contract caller confusion, opaque assembly changes, internal function-pointer exits, address-only helpers and modifiers, and recursive recipient-versus-token identity, with permanent positive and negative fixture coverage.

The corresponding Book documentation landed in foundry-rs/book#1988, with the final code-less-guard clarification in foundry-rs/book#1991. This change was prepared with AI assistance.


crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    fn visit_expr(&mut self, expr: &'hir Expr<'hir>) -> ControlFlow<Self::BreakValue> {
+        // Flag the call as soon as one resolution the call can dispatch to is the `_mint` of an
+        // ERC721 contract. Overload sets are filtered by arity: a one-argument call cannot reach
+        // the two-parameter base `_mint`, only a same-arity local overload.
+        if let ExprKind::Call(callee, args, _) = &expr.kind
+            && self
+                .resolved_function_ids(callee)
+                .into_iter()
+                .any(|id| self.is_erc721_mint(id, args.len()))
```
stevencartavia: The detector flags a call if any `_mint` candidate in the callee's resolution set has matching arity and is declared on a non-library contract whose name contains `ERC721`. It doesn't account for override resolution, so a contract that safely overrides `_mint` gets a false positive — `_mint(to, id)` below is flagged even though Solidity selects the override, not the inherited base:

```solidity
contract SafeOverride is ERC721 {
    function _mint(address to, uint256 id) internal override { /* safe, no super */ }
    function mint(address to, uint256 id) external { _mint(to, id); }
}
```

Same for `super._mint(...)` — scanning all of `linearized_bases[1..]` can still match an older ERC721 base even when the selected super target is an intermediate override. Before applying the ERC721 check, consider collapsing same-signature candidates to the most-derived declaration (in `linearized_bases` order, comparing parameter types via `gcx.item_parameter_types`), plus a fixture for the safe-override case.

0xMars42: Good catch, fixed in 1505fcec. Rather than judging the resolution set, the detector now takes the single resolved callee from `type_of_expr`, so the type checker's own dispatch decides the target: an override resolves to its own declaration, `super._mint(...)` to the closest base override, and an overloaded `_mint` to the overload the argument types select. The `SafeOverride` case is clean now, and `super._mint(...)` still flags when it reaches the ERC721 base. The fixture gains the safe-override cases (`SafeOverride`, the `MiddleSafe`/`ChildOfMiddle` chain) and a same-arity overload (`DataNft._mint(address, bytes)`) whose `uint256` sibling call still flags while its `bytes` call does not.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,93 @@
...
+    fn check_function(
+        &mut self,
+        ctx: &LintContext,
+        gcx: Gcx<'hir>,
+        hir: &'hir Hir<'hir>,
+        func: &'hir hir::Function<'hir>,
+    ) {
+        // A `_safeMint` implementation is the wrapper itself: it legitimately calls `_mint`
+        // after (or before) its receiver check.
+        if func.name.is_some_and(|name| name.as_str() == "_safeMint") {
```
mablr: This suppresses every function named `_safeMint`, including user overrides that do not perform the ERC721 receiver check. A broken override can call `_mint` directly and avoid the lint entirely:

```solidity
contract BrokenSafeNft is ERC721 {
    function _safeMint(address to, uint256 tokenId) internal override {
        _mint(to, tokenId); //~WARN: `ERC721._mint` does not check
    }
}
```

Please only exempt a known/canonical safe wrapper, verify the body actually performs the receiver check before suppressing `_mint`, or avoid special-casing arbitrary user-defined `_safeMint` functions.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    /// Whether `function_id` is a function named `_mint` declared in a non-library contract whose
+    /// name contains `ERC721` (covers `ERC721`, `ERC721Upgradeable`, `ERC721Enumerable`, ...).
+    /// Libraries are excluded: OpenZeppelin's unchecked `_mint` lives in the `ERC721` contract.
+    fn is_erc721_mint(&self, function_id: FunctionId) -> bool {
+        let function = self.hir.function(function_id);
+        function.name.is_some_and(|name| name.as_str() == "_mint")
+            && function.contract.is_some_and(|id| {
+                let contract = self.hir.contract(id);
+                !contract.kind.is_library() && contract.name.as_str().contains("ERC721")
```
mablr: Using `contains("ERC721")` on the declaring contract name makes safe overrides look unsafe just because the overriding contract name includes that substring. This reports the call below even though resolution targets the local override, matching the already-covered safe override pattern:

```solidity
contract SafeERC721Override is ERC721 {
    function _mint(address to, uint256 tokenId) internal override {
        _owners[tokenId] = to;
    }

    function mint(address to, uint256 id) external {
        _mint(to, id);
    }
}
```

Please avoid substring matching for resolved overrides. Exact known OZ base names such as `ERC721` / `ERC721Upgradeable`, or source/import identification if available, would avoid this false positive.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
0xMars42: Fixed in 2bb63a67: the exemption is now scoped to a `_safeMint` declared in `ERC721`/`ERC721Upgradeable` itself, the canonical wrapper. A user-defined `_safeMint` override is analyzed like any other function, so your `BrokenSafeNft` (which calls `_mint` directly) is flagged; it is in the fixture, where I had previously exempted this shape by mistake.

I went with the simple canonical-wrapper rule rather than inspecting the body to prove a receiver check runs: a user `_safeMint` that does its own check before calling `_mint` would be a residual false positive, but that is rare and suppressible, whereas verifying the check by data flow is a much larger change.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
0xMars42: Fixed in 2bb63a67: `is_erc721_mint` now matches the exact base names `ERC721` / `ERC721Upgradeable` instead of `contains("ERC721")`. Extensions like `ERC721Enumerable` inherit `_mint` rather than redeclare it (checked against current OZ sources), so resolution still lands on the base and they stay covered, while your `SafeERC721Override` resolves to the local override and is no longer flagged. Both are in the fixture now.

mablr (dismissed):

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,103 @@
...
+        hir: &'hir Hir<'hir>,
+        func: &'hir hir::Function<'hir>,
+    ) {
+        // Only the canonical OZ `_safeMint` wrapper is exempt: it legitimately calls `_mint`
+        // next to its receiver check. A user-defined `_safeMint` override stays analyzed, since
+        // it can call `_mint` directly without any check.
+        if func.name.is_some_and(|name| name.as_str() == "_safeMint")
+            && func.contract.is_some_and(|id| is_canonical_erc721(hir.contract(id).name.as_str()))
+        {
+            return;
```
stevencartavia, line 40: This flags the `super._mint` call inside a user `_mint` override, a false positive on the common capped/pausable pattern:

```solidity
contract Capped is ERC721 {
    function _mint(address to, uint256 id) internal override {
        require(total < cap); total++;
        super._mint(to, id); // flagged, but this is just delegation
    }
}
```

The suggested `_safeMint` fix is wrong here: inside a `_mint` override it re-enters via `_safeMint` → virtual `_mint` → back into the override, so every mint reverts (unbounded recursion, or a guard-triggered revert like the `require` above). Meanwhile a direct `_mint(to, id)` resolving to `Capped._mint` (the genuinely unsafe path) is not flagged.

Suggest exempting calls inside a `_mint` override, mirroring the existing `_safeMint` exemption:

```rust
if func.name.is_some_and(|n| n.as_str() == "_mint") && func.override_ {
    return;
}
```

This removes the FP only; flagging the direct calls into a user override would need inheritance-chain analysis and can be a follow-up.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,103 @@
...
+            && function.contract.is_some_and(|id| {
+                let contract = self.hir.contract(id);
+                !contract.kind.is_library() && is_canonical_erc721(contract.name.as_str())
+            })
+    }
+}
+
+/// The OpenZeppelin contracts that declare the unchecked `_mint`.
+fn is_canonical_erc721(name: &str) -> bool {
+    matches!(name, "ERC721" | "ERC721Upgradeable")
```
stevencartavia: `version: 1` JSON mixes range conventions: `line_end = lines.end.saturating_sub(1)` (inclusive) but `byte_end = bytes.end` (exclusive), and `line_start` is 1-based vs `byte_start` 0-based. Repro: `return x*2;` on line 4 reports `line[4..4] byte[123..133]` (byte 133 is the `;`). Undocumented — worth settling before v1 ships: emit both half-open, or document it.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
0xMars42: Fixed in ea43208e with the shape you suggested: a `_mint` override is exempt like the canonical `_safeMint` wrapper (`func.name == "_mint" && func.override_`), since `super._mint` there is delegation and `_safeMint` would re-enter the override through the virtual dispatch. Your capped pattern is in the fixture (`CappedNft`, clean), and `super._mint` from a non-`_mint` function still reports (`SafeOverride.mintSuper`). Agreed on leaving the direct-call-into-an-unsafe-override case to a follow-up, it needs the inheritance-chain analysis you mention.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
0xMars42: On this branch `forge lint --json` emits rustc-style diagnostics straight from solar's JSON emitter (`crates/interface/src/diagnostics/emitter/json.rs` in the solar repo): `$message_type: "diagnostic"` with rustc's `DiagnosticSpan` fields, and I could not find a `version` field on this path. The conventions I observe in the emitted spans are rustc's: `byte_start`/`byte_end` 0-based with an exclusive end, `column_start`/`column_end` 1-based one past the end, `line_end` the line of the span's end (a 13-byte single-line span reports `byte[1671..1684] line[45..45] column[9..22]`). Since the shape is inherited from solar, settling or documenting it would apply to every solar diagnostic, not just the lints, so it seems worth raising there rather than in this PR.

stevencartavia (dismissed):

0xMars42:
9b0bfa37 extends the canonical list to the v4 Consecutive extensions. Stock OZ v4 `ERC721Consecutive` / `ERC721ConsecutiveUpgradeable` override `_mint` with a construction guard and forward to the base through `super._mint`, still without the receiver check, so a consumer inheriting them resolved `_mint` to that override and went unflagged. Verified against v4.9.6 (`ERC721Consecutive.sol`, the `_mint` override forwards after an `Address.isContract` guard); in v5 the Consecutive extension overrides `_update` instead, so the two names match nothing there. Fixture (`ERC721Consecutive` mirror plus a flagged consumer) and docs updated.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,118 @@
...
+        if func.name.is_some_and(|name| name.as_str() == "_safeMint")
+            && func.contract.is_some_and(|id| is_canonical_erc721(hir.contract(id).name.as_str()))
+        {
+            return;
+        }
+        // A user `_mint` override is part of the mint primitive itself: `super._mint` there is
+        // delegation (the capped/pausable pattern), the receiver check belongs to the
+        // override's callers, and `_safeMint` there would re-enter the override through the
+        // virtual dispatch.
+        if func.name.is_some_and(|name| name.as_str() == "_mint") && func.override_ {
```
mablr, line 46: This blanket `_mint` override exemption creates a false negative for direct callers of an override that delegates to OZ `_mint`.

For example, this currently produces no warning on `unsafeMint`, even though the path still reaches unchecked `ERC721._mint` without an `onERC721Received` check:

```solidity
contract DelegatingOverrideNft is ERC721 {
    function _mint(address to, uint256 tokenId) internal virtual override {
        super._mint(to, tokenId);
    }

    function unsafeMint(address to, uint256 id) external {
        _mint(to, id); //~WARN expected
    }

    function safeMint(address to, uint256 id) external {
        _safeMint(to, id);
    }
}
```

I verified this by adding the fixture annotation and running `cargo test -p forge --test ui -- --nocapture`; the UI test failed with “expected diagnostic not found”. Consider treating `_mint` overrides that transitively delegate to canonical unchecked ERC721 `_mint` as unsafe call targets, while still avoiding a warning on the `super._mint` inside the override itself.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,118 @@
...
+/// The OpenZeppelin contracts whose `_mint` skips the receiver check. `ERC721` and
+/// `ERC721Upgradeable` declare the unchecked `_mint`; in the v4 line, `ERC721Consecutive` and
+/// `ERC721ConsecutiveUpgradeable` override it with a construction guard that forwards to the
+/// base through `super._mint`, still without a receiver check, so resolving to them is just
+/// as unsafe. In v5 the Consecutive extension overrides `_update` instead, and the two extra
+/// names match nothing.
+fn is_canonical_erc721(name: &str) -> bool {
+    matches!(
+        name,
+        "ERC721" | "ERC721Upgradeable" | "ERC721Consecutive" | "ERC721ConsecutiveUpgradeable"
```
mablr, line 2056: This exact-name match can false-positive on unrelated contracts named `ERC721Consecutive` / `ERC721ConsecutiveUpgradeable`.

I verified it with this expected-clean fixture:

```solidity
contract ERC721ConsecutiveUpgradeable {
    function _mint(address account, uint256 amount) internal {
        account;
        amount;
    }
}

contract UsesConsecutiveUpgradeableName is ERC721ConsecutiveUpgradeable {
    function mint(address account, uint256 amount) external {
        _mint(account, amount); // should not lint
    }
}
```

`cargo test -p forge --test ui -- --nocapture` produced an unmatched `unsafe-oz-erc721-mint` diagnostic on that call. For the Consecutive names, matching the contract name alone is too broad; please also require that the contract is actually in an ERC721 inheritance chain or that its `_mint` override delegates to canonical unchecked `ERC721._mint` / `ERC721Upgradeable._mint`.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
0xMars42: Fixed in 390d99184 along the lines you suggested: a `_mint` override is judged recursively, and one whose body dispatches to an unsafe `_mint` (canonical or another delegating override) is an unsafe call target itself, so your `unsafeMint` repro reports while `safeMint` and the `super._mint` inside the override stay clean. The existing capped-pattern testdata case now reports at its call site too, and a `seen` set cuts override cycles.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
0xMars42: Fixed in 390d99184 with the provenance approach used on #15579 and #15580: the canonical names must also come from an OpenZeppelin package path, so your local `ERC721ConsecutiveUpgradeable` repro stays clean and is in the testdata. A contract that is genuinely in an ERC721 inheritance chain still reports through resolution or the new delegation check, and the docs state the vendored-copy trade-off.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,198 @@
...
+        }
+        // The canonical unchecked `_mint`: exact OZ name, OZ package provenance. Most
+        // extensions (`ERC721Enumerable`, ...) inherit `_mint` rather than redeclare it, so
+        // resolution still lands here.
+        if is_canonical_erc721(contract.name.as_str())
+            && is_openzeppelin_source(self.hir, function.source)
+        {
+            return true;
+        }
+        // A delegating override: any call in its body dispatching to an unsafe `_mint`.
```
mablr: This treats every `_mint` override that calls an unsafe base `_mint` as an unsafe target, which can still false-positive when the override adds the receiver check itself before delegating. For example, an override can perform an `IERC721Receiver(to).onERC721Received(...)`/code-length check and then call `super._mint(to, id)`; direct callers of that override are safe, but the collector only sees the transitive `super._mint` and reports them. Please either avoid classifying arbitrary user overrides as unsafe solely from delegation, or recognize/suppress overrides that implement the receiver check before forwarding.

0xMars42: Fixed in 5a1cffbef with the recognition option: the override body scan also looks for the check markers, a call whose resolved declaration is named `onERC721Received` or a `code` member read on an address-typed base (so a user field named `code` does not count), and an override carrying either marker is not an unsafe target. Your checked-wrapper repro and a code-length-only variant are in the testdata as clean, and a bare delegating override still reports at its call sites.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,236 @@
...
+                }
+            }
+            // `to.code.length` style inspections: the `code` member of an address value,
+            // judged on the base's type so a user field named `code` does not count.
+            ExprKind::Member(base, member) => {
+                if member.as_str() == "code"
+                    && let Some(ty) = self.gcx.type_of_expr(base.peel_parens().id)
+                    && matches!(ty.peel_refs().kind, TyKind::Elementary(ElementaryType::Address(_)))
+                {
+                    self.has_receiver_check = true;
```
mablr: This treats any `address.code` access in a `_mint` override as proof that the override performs the ERC721 receiver check, but the address may be unrelated to the recipient. For example, an override with `require(address(this).code.length > 0); super._mint(to, tokenId);` still reaches unchecked `ERC721._mint` for arbitrary `to`, yet direct callers are no longer reported because the `address(this).code` construction/status guard sets `has_receiver_check`. Please only suppress delegating overrides when the code/onERC721Received check is tied to the minted recipient (or otherwise avoid treating arbitrary address code inspections as a receiver check).

0xMars42: Fixed in 5d8bb83f9: both markers are now tied to the minted recipient, the override's first address-typed parameter. The `.code` base and the `onERC721Received` call receiver must mention it, so your `address(this).code` construction-guard repro no longer counts as a receiver check and its callers report, while the recipient-checking wrappers stay clean.

mattsse (dismissed):
Found one blocker:

- `crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs:176`: the OpenZeppelin provenance check still accepts any path containing `openzeppelin`. A local or third-party path such as `src/not-openzeppelin/ERC721.sol` can satisfy the check and make an unrelated same-name `_mint` look canonical. Please match actual package/path components such as `@openzeppelin` or `openzeppelin-contracts`, and add a negative fixture for a misleading path containing `openzeppelin`.

0xMars42:
Fixed in `e6b753bb3`. Provenance matches a full path component against the package roots (`@openzeppelin`, `openzeppelin-contracts`, `openzeppelin-contracts-upgradeable`) instead of a substring. Added a negative fixture: a local `ERC721` under `not-openzeppelin/` whose `_mint` stays out of scope.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,300 @@
...
+            // that mentions the recipient, judged on the base's type so a user field named
+            // `code` does not count, and on the mention so `address(this).code` does not
+            // either.
+            ExprKind::Member(base, member) => {
+                if member.as_str() == "code"
+                    && let Some(ty) = self.gcx.type_of_expr(base.peel_parens().id)
+                    && matches!(ty.peel_refs().kind, TyKind::Elementary(ElementaryType::Address(_)))
+                    && self.mentions_recipient(base)
+                {
+                    self.has_receiver_check = true;
```
mablr: This still treats any read of the recipient's `.code` as proof that the override performs the ERC721 receiver check, even if the result is not used to guard the delegated mint. For example, an override can do `uint256 len = to.code.length; len; super._mint(to, tokenId);`; direct callers then skip the lint even though arbitrary contract recipients are still minted through unchecked `ERC721._mint`. Please only suppress delegating overrides when the recipient `.code`/`onERC721Received` check actually gates the mint path, or avoid treating a bare `.code` read as sufficient.

mattsse (dismissed):
The previous OpenZeppelin provenance blocker is fixed, but there is still a false negative in the delegating override handling.

At `crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs:253`, any read of the recipient's `.code` is treated as proof that the override performs a receiver check. An override can read `to.code.length` into a local, ignore it, and then call `super._mint(...)`; direct callers would skip the diagnostic even though the unchecked mint path remains. Please require the recipient code check to actually guard/control the mint or receiver-call path, rather than treating any `.code` read as sufficient.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,300 @@
...
+                    // and the call receiver must involve the recipient.
+                    if self
+                        .hir
+                        .function(function_id)
+                        .name
+                        .is_some_and(|name| name.as_str() == "onERC721Received")
+                        && let ExprKind::Member(receiver, _) = &callee.peel_parens().kind
+                        && self.mentions_recipient(receiver)
+                    {
+                        self.has_receiver_check = true;
```
mablr: This treats any `onERC721Received` call as proof of acceptance without checking its `bytes4` result. A receiver can return the wrong selector without reverting, so an override that ignores the result still suppresses warnings for an unsafe `_mint` path. Please require control flow that enforces `IERC721Receiver.onERC721Received.selector`.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,300 @@
...
+                recipient,
+                callees: Vec::new(),
+                has_receiver_check: false,
+            };
+            // The body's resolved callees and check markers are collected first.
+            for stmt in body.stmts {
+                let _ = scan.visit_stmt(stmt);
+            }
+            // Each callee is judged with the shared `seen` set, so override cycles stop.
+            if !scan.has_receiver_check {
```
mablr: Receiver checks performed by an internal helper never contribute to `has_receiver_check`: recursion here only recognizes unsafe `_mint` targets. Thus `_checkReceiver(to, tokenId); super._mint(...)` is still classified unsafe even when the helper performs the code-length branch and enforces the exact callback selector. Please follow direct helper calls or narrow the documented exemption.

0xMars42:
Fixed in `5b09b9dd4`. Reading `.code` never proved the mint reverts on refusal, and neither does an `onERC721Received` call that merely appears in a condition: `require(to == trusted || hook(to) == selector)` short circuits past the hook, and one riding in the revert message decides nothing. The exemption is now a closed set of guards, each of which reverts exactly when the recipient refuses.

A guard is `require`/`assert` holding `hook(to) == selector` as its whole condition, or the account short circuit `to.code.length == 0 || hook(to) == selector`; an `if` whose reverting branch is the one a refusal takes; either of those under a test that the recipient carries code; or either of those reached through a function or a modifier the recipient itself is handed to, the way `_checkOnERC721Received` is factored out of `_safeMint`. The guard may follow the delegation, since a revert undoes the mint wherever it sits.

Reported now, each with a fixture: a refusal that only returns or escapes through assembly, a guard behind a `virtual` callee whose body an override may replace, a hook asked of an address derived from the recipient, an answer compared against something that is not provably the selector, a second mint handed an address the guard never named, an exiting branch taken on acceptance, and an answer discarded, stored in a local, or wrapped in a `try`. The limits are documented rather than silent: a `try`, a local, and a `bool`-returning helper report, following a value across statements needing a dataflow analysis this detector does not run.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,687 @@
...
+/// answer stored in a local and a helper returning it as a `bool`. Following the value across
+/// statements would take a dataflow analysis this detector does not run.
+fn reverts_on_refusal<'hir>(
+    gcx: Gcx<'hir>,
+    hir: &'hir Hir<'hir>,
+    stmts: &'hir [hir::Stmt<'hir>],
+    recipient: hir::VariableId,
+    seen: &mut Vec<FunctionId>,
+) -> bool {
+    stmts.iter().any(|stmt| stmt_reverts_on_refusal(gcx, hir, stmt, recipient, seen))
```
mablr: This can miss an unsafe path because `any` ignores statement order. For example, `super._mint(to, id); if (id == 0) return; require(hook(to) == selector);` leaves token 0 minted without a callback, but the later guard exempts every caller. An expected-warning UI fixture confirms the diagnostic is missing; the proof needs to account for successful exits before a guard.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,687 @@
...
+fn callee_guards_recipient<'hir>(
+    gcx: Gcx<'hir>,
+    hir: &'hir Hir<'hir>,
+    function_id: FunctionId,
+    args: &'hir hir::CallArgs<'hir>,
+    recipient: hir::VariableId,
+    seen: &mut Vec<FunctionId>,
+) -> bool {
+    let parameters = hir.function(function_id).parameters;
+    for (index, arg) in args.exprs().enumerate() {
```
mablr: Named helper arguments are bound by source position here, not parameter name. `_check({actual: to, checked: checker})` can therefore analyze `to` as the `checked` parameter even though runtime checks `checker`, suppressing the unsafe mint warning. The mint collector already handles named binding correctly; helper calls need the same mapping.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,687 @@
...
+    recipient: hir::VariableId,
+) -> bool {
+    let expr = expr.peel_parens();
+    let ExprKind::Call(callee, ..) = &expr.kind else { return false };
+    let Some(function_id) = resolved_callee(gcx, expr) else { return false };
+    if !is_receiver_hook(hir, function_id) {
+        return false;
+    }
+    let ExprKind::Member(receiver, _) = &callee.peel_parens().kind else { return false };
+    is_exactly_recipient(receiver, recipient)
```
mablr: The callback is tied only to the recipient, not to the delegated token. A wrapper can check `onERC721Received(..., tokenId + 1, ...)` and then `super._mint(to, tokenId)`; a receiver may accept the former and refuse the latter, yet callers are suppressed. The adversarial UI fixture confirms this missing diagnostic, so the callback arguments need correlation with each delegated mint.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,687 @@
...
+            match (operands.next(), operands.next()) {
+                (Some(inner), None) => is_received_selector(hir, inner),
+                _ => false,
+            }
+        }
+        // `IERC721Receiver.onERC721Received.selector`, the answer named rather than spelled.
+        ExprKind::Member(base, member) => {
+            member.as_str() == "selector"
+                && matches!(&base.peel_parens().kind, ExprKind::Member(_, hook)
+                    if hook.as_str() == "onERC721Received")
```
mablr: Any member spelled `onERC721Received.selector` is accepted without resolving its signature or value. Comparing the real hook result with `IWrongSelector.onERC721Received.selector` (for a zero-argument function) therefore suppresses the warning even though that selector is not `0x150b7a02`; the fixture reproduces the false negative. Please resolve/validate the selector rather than matching member names.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,687 @@
...
+        return resolutions.iter().any(
+            |res| matches!(res, hir::Res::Item(hir::ItemId::Variable(vid)) if *vid == recipient),
+        );
+    }
+    // A cast keeps the address: `IERC721Receiver(to)` wraps the callee as an identifier resolved
+    // to the contract, `address(to)` as a type.
+    if let ExprKind::Call(callee, args, _) = &expr.kind
+        && matches!(
+            &callee.peel_parens().kind,
+            ExprKind::Type(..) | ExprKind::Ident([hir::Res::Item(hir::ItemId::Contract(_)), ..])
```
mablr: Unwrapping every one-argument type conversion treats lossy casts as recipient-preserving. `super._mint(address(uint160(uint8(uint160(to)))), id)` is considered a mint to `to`, although it usually targets a different truncated address that was never checked. The expected-warning fixture is missed; only value-preserving address/interface casts should be unwrapped.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,687 @@
...
+fn is_hook_call_on<'hir>(
+    gcx: Gcx<'hir>,
+    hir: &'hir Hir<'hir>,
+    expr: &'hir Expr<'hir>,
+    recipient: hir::VariableId,
+) -> bool {
+    let expr = expr.peel_parens();
+    let ExprKind::Call(callee, ..) = &expr.kind else { return false };
+    let Some(function_id) = resolved_callee(gcx, expr) else { return false };
+    if !is_receiver_hook(hir, function_id) {
```
mablr, line 674: This shape does not prove an external callback was made to the recipient. A `using FakeReceiverCheck for address` library can provide an internal `onERC721Received(address,address,uint256,bytes)` that always returns the selector; `to.onERC721Received(...)` then exempts the wrapper without calling `to` at all. The UI fixture confirms the miss; library/internal extension functions need to be excluded.

mablr (commented):
Please check the following FNs, maybe not all worth fixing.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
0xMars42: Covered by 5b09b9dd4: a bare `.code` read no longer exempts. The read has to gate the mint, either `if (to.code.length > 0)` around a recognized guard or the account short circuit inside `require`; `BareCodeReadNft` pins the bare-read case.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
0xMars42: Covered by 5b09b9dd4: a hook call only counts inside a revert guard comparing its answer to the accepting selector, so a call whose result is ignored reports (`IgnoredAnswerNft`). 7482875c3 tightens what "the answer" means: the callee must resolve to an externally callable declaration and the token asked about must be the delegated one.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
0xMars42: Covered by 5b09b9dd4: a function the recipient is handed to is followed (`callee_guards_recipient`), which is the OZ `_checkOnERC721Received` shape; `CheckedViaHelperNft` pins it. 7482875c3 tightens the follow: arguments bind by parameter name, the token has to be handed too, and the helper's guard has to run before any of its own possible exits.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
0xMars42: Fixed in 7482875c3. The body is read in statement order: a guard covers the delegations before it only when no statement in between may leave the function successfully, so `super._mint(to, id); if (id == 0) return; require(...)` reports (`MintThenExitGuardNft`). Branches are read separately and merged, so a guard behind an unrelated condition still counts only when every branch holds one (`EitherBranchGuardNft` clean, `ConditionalGuardNft` reported), a conditional mint is covered by the unconditional guard after it (`ConditionalMintNft`), and a revert in between is not an escape (`RevertBetweenNft`). Loop bodies stay uncredited as before, and the account branch of a `to.code.length` test satisfies the mints already made as well as the ones to come, so the usual skip stays exempt on both sides of the mint (`MintThenAccountSkipNft`, `MintThenReversedSkipNft`). The same escape existed on the modifier path: a guard after `_;` was credited even when the body can leave through an assembly `return` that skips the modifier tail. `ModifierTailAssemblyNft` reports now, while a tail guard over a body without assembly stays exempt, a plain `return` routing through the tail (`ModifierTailGuardNft`).

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
0xMars42: Fixed in 7482875c3. Helper and hook arguments now bind by parameter name through the same mapping the mint collector used (`argument_bound_to_parameter`); `NamedHelperArgsNft` pins the misdirection. The binding also hardened the using-for case below: named arguments land the token on the right parameter despite the shifted first argument, so only the declaration check stops `NamedAttachedHookNft`.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
0xMars42: Fixed in 7482875c3. The delegated token is the variable the delegation binds to the callee's second parameter, and the hook's third argument must be exactly that variable, through helpers and modifiers (`WrongTokenHookNft`, `HelperWrongTokenNft`). The limits all sit on the reporting side: a token that is not a plain variable (a literal, an expression, a numeric cast) has no name to match a guard against and reports; two delegations minting different tokens report; a mutable state variable reports too, the hook being an external call the recipient can reenter to move it between the ask and the mint (`StateTokenNft`). Identity being by variable and not by value, reassigning the recipient or the token between the guard and the delegation reports as well, including the collection-offset `tokenId = tokenId + 1` and a redirected recipient (`LocalTokenReassignNft`, `RecipientReassignNft`, `ParamTokenRemapNft`, `AccountBranchReassignNft`), whether the reassignment is a bare statement or embedded in an `if` condition that runs before either branch (`CondAssignTokenNft`, `CondAssignRecipientNft`, `CondIncrementTokenNft`); a reassignment before the guard or after the mint it covers stays exempt. Since the check reads variables and not values, a reassignment that changes nothing still reports, a conservative edge.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
0xMars42: Fixed in 7482875c3. The `.selector` member resolves through the type checker and must land on the receiver hook itself, name, `(address, address, uint256, bytes)` shape, externally callable, non-library, so `INotAReceiver.onERC721Received.selector` no longer exempts (`ForeignSelectorNft`).

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
0xMars42: Fixed in 7482875c3. Only conversions that keep the operand peel: `payable(...)`, casts to `address`, casts to contract types. The truncating chain reports (`TruncatedRecipientNft`), and `payable(to)` stops reporting (`PayableRecipientNft`). One conservative edge remains: `address(uint160(to))` is value-preserving in fact but reports, the peel not following numeric intermediates (`RoundTripCastNft`).

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
0xMars42: Fixed in 7482875c3. The resolved callee must be an externally callable declaration of a non-library contract, which rules out attached library and free functions (`AttachedHookNft`). The positional attached form was already dead in practice, using-for shifting the visible arguments off the token check, but named arguments rebind the token correctly and evade everything except the declaration check, which `NamedAttachedHookNft` pins.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,1130 @@
...
+                // to the second, as the canonical `_mint(address to, uint256 tokenId)` orders
+                // them.
+                let handed_the_recipient = recipient.is_some_and(|recipient| {
+                    argument_bound_to_parameter(self.hir, *callee, args, 0)
+                        .is_some_and(|argument| is_exactly_var(argument, recipient))
+                });
+                if !handed_the_recipient {
+                    only_to_recipient = false;
+                }
+                match argument_bound_to_parameter(self.hir, *callee, args, 1)
```
mablr, line 273: This correlates only the arguments passed to the immediate unsafe override, losing remaps deeper in the chain. If an intermediate override forwards `tokenId + 1`, an outer override can guard `tokenId` and pass `tokenId` to it; callers are then suppressed although the canonical mint receives a different token. Please preserve recipient/token identity through recursive targets or conservatively reject the exemption when a target does not forward both unchanged.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,1130 @@
...
+                // does. The branch the guard lets through holds nothing to read: a mint there
+                // runs checked, and an exit there leaves with the recipient's answer in hand.
+                if (is_hook_comparison(gcx, hir, cond, recipient, token, hir::BinOpKind::Ne)
+                    && branch_always_reverts(then))
+                    || (is_hook_comparison(gcx, hir, cond, recipient, token, hir::BinOpKind::Eq)
+                        && else_.is_some_and(branch_always_reverts))
+                {
+                    walk.covered = true;
+                    walk.pending = false;
+                    continue;
```
mablr, line 1245: The early `continue` skips the successful branch of a recognized guard. For `if (hook(to,id) != selector) revert(); else to = redirected; super._mint(to,id);`, coverage remains true even though the accepted branch changes the recipient before minting. Please continue walking the acceptance branch from the covered state instead of skipping both branches.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,1130 @@
...
+        return false;
+    }
+    seen.push(function_id);
+    let function = hir.function(function_id);
+    // A `virtual` callee may be replaced by an override that drops the guard, and the body seen
+    // here is the statically resolved one, not the one dispatched.
+    if function.virtual_ {
+        return false;
+    }
+    let Some(body) = &function.body else { return false };
```
mablr: This proves a helper guard from its body without accounting for the helper’s modifiers. A modifier such as `if (!skip) _;` can return successfully without executing the body, so `_check(to,id); super._mint(to,id)` is still unsafe while callers are suppressed. Until modifier expansion is modeled here, helpers carrying modifiers should not establish coverage.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,1130 @@
...
+/// body holding an assembly block can leave the call frame without ever coming back to the
+/// modifier, so a guard after `_;` is only guaranteed for a body that cannot.
+fn modifiers_revert_on_refusal<'hir>(
+    gcx: Gcx<'hir>,
+    hir: &'hir Hir<'hir>,
+    function: &'hir hir::Function<'hir>,
+    recipient: hir::VariableId,
+    token: hir::VariableId,
+) -> bool {
+    let bypass = function.body.as_ref().is_some_and(|body| contains_assembly(hir, body.stmts));
```
mablr: `bypass` scans only the function body, but another applied modifier can execute `assembly { return(0,0) }` after its placeholder and skip an outer modifier’s tail guard. With `checkedAfter(to,id) escapeAfter`, the body mints, `escapeAfter` returns, and `checkedAfter` never checks the receiver. Include nested/sibling modifier bodies when deciding whether a tail guard can be bypassed.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,1130 @@
...
+/// analysis does not read.
+fn may_return(stmt: &hir::Stmt<'_>) -> bool {
+    match &stmt.kind {
+        hir::StmtKind::DeclSingle(_)
+        | hir::StmtKind::DeclMulti(..)
+        | hir::StmtKind::Emit(_)
+        | hir::StmtKind::Revert(_)
+        | hir::StmtKind::Break
+        | hir::StmtKind::Continue
+        | hir::StmtKind::Expr(_)
```
mablr, line 841: Treating every expression as non-exiting lets `branch_always_reverts` look past an internal helper that executes assembly `return`. In a refusal branch, `_escape(); revert();` can preserve the prior mint because `_escape` exits the call frame, yet the later revert makes the branch look safe. Internal calls need at least a conservative transitive assembly-exit summary here.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,1130 @@
...
+        ExprKind::Lit(lit) => {
+            matches!(&lit.kind, LitKind::Number(value) if *value == U256::from(ERC721_RECEIVED))
+        }
+        // `bytes4(0x150b7a02)` and other conversions of a literal.
+        ExprKind::Call(callee, args, _)
+            if matches!(callee.peel_parens().kind, ExprKind::Type(..)) =>
+        {
+            let mut operands = args.exprs();
+            match (operands.next(), operands.next()) {
+                (Some(inner), None) => is_received_selector(gcx, hir, inner),
```
mablr: Recursively peeling every type conversion accepts lossy casts as the receiver selector. `bytes4(uint32(uint8(uint32(0x150b7a02))))` evaluates to `0x00000002`, but reaches the original literal here and suppresses callers. Please evaluate the constant or only peel conversions proven to preserve the four-byte value.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,1130 @@
...
+    function_id: FunctionId,
+    recipient: hir::VariableId,
+    token: hir::VariableId,
+    bypass: bool,
+    seen: &mut Vec<FunctionId>,
+) -> bool {
+    if seen.contains(&function_id) {
+        return false;
+    }
+    seen.push(function_id);
```
mablr, line 1843: `seen` never pops this helper, so independent later calls are mistaken for recursion. `_check(to,id); id += 1; _check(to,id); super._mint(to,id);` is safe, but the first call leaves `_check` in `seen`, the mutation clears coverage, and the second check is rejected, producing a false positive. Please use stack/path semantics for recursion detection.

mablr (dismissed):
LGTM

crates/lint/testdata/UnsafeOzErc721Mint.sol
```diff
@@ -0,0 +1,2055 @@
...
+// Admitting only code-less recipients is not a receiver check either: a contract minting to
+// itself from its constructor passes it, and never acknowledged the token.
+contract AccountOnlyNft is ERC721 {
+    function _mint(address to, uint256 tokenId) internal virtual override {
+        require(to.code.length == 0, "accounts only");
+        super._mint(to, tokenId);
+    }
+
+    function mint(address to, uint256 id) external {
+        _mint(to, id); //~WARN: `ERC721._mint` does not check
```
figtracer, line 1321: `AccountOnlyNft` is a false positive: after `require(to.code.length == 0)`, `_safeMint` also skips `onERC721Received`, so replacing `_mint` changes no safety behavior. please keep this case clean or omit the detector until its suggested fix is actionable.

mattsse: Addressed in feceaa9b3. A guard that proves the recipient has no code now exempts the delegated mint, including address-only helpers or modifiers and computed or remapped token IDs. Callback guards remain correlated to both recipient and token, and recursive recipient remaps still report. The focused cases and full lint UI suite pass.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,1469 @@
...
+                    self.hir,
+                    body.stmts,
+                    recipient,
+                    recipient,
+                    &unsafe_targets,
+                    false,
+                    &mut Vec::new(),
+                    &mut walk,
+                );
+                if !walk.failed && !walk.pending {
```
mablr, line 328: A code-length proof is not stable across intervening calls. After `require(to.code.length == 0)`, an internal/external call can deploy code at `to` (e.g. via CREATE2), after which `super._mint(to, id)` skips the receiver callback even though `_safeMint` at that point would require it. I verified this by inserting `deployRecipient();` between the guard and delegation in an override; an expected warning at its caller was not emitted by the UI test. Please retire code-less coverage across calls that can change account code, or only credit the proof when it remains valid until the delegation.

mattsse: Fixed in 5ed2fb49e. Code-less recipient proofs are now retired across calls that can actually change account code, including CREATE2-capable helpers, recursive overrides, recognized-guard extra arguments, modifier prefix/tail execution, and modifier-argument recipient changes. Body-inspected storage-only and view helpers retain the proof. Direct regressions cover these paths; full UI/clippy/fmt validation and final fresh-context re-review pass.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,1895 @@
...
+#[derive(Clone, Default)]
+struct GuardWalk {
+    coverage: GuardCoverage,
+    pending: bool,
+    failed: bool,
+    escaped: bool,
+}
+
+/// Reads a body in statement order and judges the delegated mints against the guards. A guard
+/// executed before a delegation covers it, the recipient having accepted the token by the time
```
mablr: A callback before delegation is not equivalent to OZ `_safeMint` and should not suppress a later unchecked mint. OZ establishes ownership before invoking the receiver; a recipient may accept before `super._mint` but reject afterward based on `ownerOf`, balance, or reentrant token operations. Only post-mint callback coverage proves `_safeMint` equivalence (while a code-less proof may still precede the mint).

mattsse: Fixed in 7003908ad. Callback coverage now only discharges mints that have already completed, while code-less-recipient proofs may still cover future mints. The analysis also conservatively rejects mutable state token IDs across external calls. I added post-mint ordering, helper/modifier/branch, and mutable-token regression coverage; the full lint UI suite, strict forge-lint Clippy, formatting, and diff checks pass, and a fresh-context review found no remaining issues.

crates/lint/src/sol/med/unsafe_oz_erc721_mint.rs
```diff
@@ -0,0 +1,1984 @@
...
+        function_id: FunctionId,
+        seen: &mut Vec<FunctionId>,
+    ) -> Option<UnsafeMintTarget> {
+        // A cycle of overrides never reaches the canonical declaration.
+        if seen.contains(&function_id) {
+            return None;
+        }
+        seen.push(function_id);
+        let function = self.hir.function(function_id);
+        if function.name.is_none_or(|name| name.as_str() != "_mint") {
```
mablr: This name gate prevents transitive delegation through ordinary helpers. If an `_mint` override calls a private `mintUnchecked(to, id)` helper that calls `super._mint(to, id)`, an external call to the override is missed; the lint only warns inside the private helper, where `_safeMint` is not a viable call-site fix. Please follow resolved internal/private helpers transitively when determining whether an `_mint` override reaches the canonical unsafe mint.

mattsse: Fixed in afd5b01be. Helper detection now follows override delegation through inherited and overloaded internal helpers, while excluding only the actual _mint override. Added regressions for inherited and overloaded helper callers; the full lint UI suite passes.

mablr (dismissed):
Don't see more false-positives. Some minor false-negatives remain but I suggest to ignore them as this lint is getting bloated.

stevencartavia (commented):
lgtm, missing clippy

mablr (approved):

figtracer (approved):

PR #15552, opened 2026-07-02
fix(anvil): use TransactionConditional

mattsse:
Anvil's `eth_sendRawTransactionConditional` request parser was storing the conditional payload as `serde_json::Value` and then manually checking for a narrow empty shape. This switches the RPC request to Alloy's `TransactionConditional` type so request parsing follows the existing Alloy schema.

Anvil still does not enforce transaction conditionals. Instead of rejecting non-empty conditionals, the endpoint now accepts the typed condition and ignores it before sending the raw transaction, which keeps the endpoint usable for clients that always include condition metadata.

Authored with AI assistance.


stevencartavia (approved):
optional nit,`ignores_raw_transaction_conditional_prestate` doesn't actually test prestate — maybe use a non-empty `knownAccounts` condition or rename it.

grandizzy (approved):
lgtm

PR #15553, opened 2026-07-02
feat(lint): add function-init-state detector

0xMars42:
## Motivation

`function-init-state` is an unchecked item of #14381. State variable initializers run at construction, before the constructor body, in base-to-derived order: an initializer that reads another state variable or calls a non-pure function observes that partial state (default values, surprising ordering), so the computed value is rarely the intended one, and silently so. This mirrors Slither's `function-init-state` detector (`slither/detectors/variables/function_init_state_variables.py`): a state variable is flagged when its initializer references a non-constant state variable or a non-pure function.

## Solution

New `Info` severity late pass. For each state variable of a contract with an inline initializer (constants excluded: their initializers are restricted to compile-time constant expressions), the initializer expression is walked and the variable is flagged when it references:

- a non-constant state variable, or
- a non-pure function, called or referenced, including inside the arguments of a nested call.

Resolution details:

- a public variable's identifier also resolves to its synthesized getter: that resolution is judged as a read of the variable itself, so a reference to a public constant stays clean;
- a qualified or external access (`Base.viewFn()`, `Oracle(addr).price()`) carries no resolution on the member ident, so the base expression is typed: when it is a contract (bare `Contract(..)` for a value, `Type(Contract(..))` for a contract name), its same-name functions and variables are judged;
- references to constants (local, inherited, library, file-level), calls to pure functions, and plain literal expressions never fire, and constructor-body assignments are out of scope.

`new Contract()` in an initializer is not flagged: no state variable or function is referenced, which matches Slither's behavior.

## Test

`testdata/FunctionInitState.sol` covers the firing forms (non-pure call, view call, direct state read, state read hidden in a pure call's argument, inherited call, its qualified `Base.` form, an external getter call, and a non-pure immutable initializer) and the clean forms (literal, file-level and library constants, a public local constant, whose identifier also resolves to its synthesized getter, the qualified pure call, a pure call, and a constructor-body assignment).

Running the detector over the repo's forge integration fixtures (the root `testdata/` directory) yields exactly one finding: `bytes bytecode = vm.getDeployedCode(...)` in `cheats/ExpectCreate.t.sol`, a test contract initializing state through a cheatcode view call, which is faithful to the rule (Slither reports the same shape). Happy to special-case cheatcode calls if that noise is unwanted in test code.

Part of #14381.


crates/lint/src/sol/info/function_init_state.rs
```diff
@@ -0,0 +1,151 @@
...
+    type BreakValue = Infallible;
+
+    fn hir(&self) -> &'hir Hir<'hir> {
+        self.hir
+    }
+
+    fn visit_expr(&mut self, expr: &'hir Expr<'hir>) -> ControlFlow<Self::BreakValue> {
+        match &expr.kind {
+            ExprKind::Ident(resolutions) => {
+                for res in *resolutions {
```
stevencartavia: Calling a pure overload gets flagged when a same-name non-pure overload exists. The `Ident` arm judges every entry in `resolutions` and the `Member` arm judges every same-name item, so neither uses the selected overload.

```solidity
function mix(uint256 x) public pure returns (uint256) { return x + 1; }
function mix(uint256 x, uint256 y) public view returns (uint256) { return x + y + block.timestamp; }

uint256 public a = mix(1); // calls the pure overload, but flagged
```

`mix(1)` calls the pure overload. Removing the view overload makes the warning go away, so the view overload's name match is what triggers it. Same on the member path with `C(addr).mix(1)`.

crates/lint/src/sol/info/function_init_state.rs
```diff
@@ -0,0 +1,151 @@
...
+                    let ty = ty.peel_refs();
+                    match ty.kind {
+                        TyKind::Type(inner) => inner.peel_refs(),
+                        _ => ty,
+                    }
+                });
+                if let Some(ty) = base_ty
+                    && let TyKind::Contract(contract_id) = ty.kind
+                {
+                    for item_id in self.hir.contract(contract_id).items {
```
stevencartavia: Inherited members reached through a cast base aren't flagged. The `Member` arm iterates `hir.contract(id).items`, which is only the contract's own items.

```solidity
contract A { function readIt() public view returns (uint256) { return 1; } }
contract B is A {}

uint256 public direct    = A(address(0x1)).readIt(); // flagged
uint256 public inherited = B(address(0x2)).readIt(); // not flagged
```

`readIt` is inherited by `B`, so it's not in `B`'s own items. Public getters inherited the same way (`B(addr).s()`) are also missed. `linearized_bases` covers the chain (used in `locked_ether.rs` and `incorrect_erc20_interface.rs`).

crates/lint/src/sol/info/function_init_state.rs
```diff
@@ -0,0 +1,151 @@
...
+            ExprKind::Member(base, member) => {
+                let base_ty = self.gcx.type_of_expr(base.peel_parens().id).map(|ty| {
+                    let ty = ty.peel_refs();
+                    match ty.kind {
+                        TyKind::Type(inner) => inner.peel_refs(),
+                        _ => ty,
+                    }
+                });
+                if let Some(ty) = base_ty
+                    && let TyKind::Contract(contract_id) = ty.kind
```
stevencartavia: `using for` calls on a value base aren't flagged. The base type is a value type, not `TyKind::Contract`, so the member arm skips it and the bound library function is never checked.

```solidity
library Clock { function stamp(uint256 x) internal view returns (uint256) { return x + block.timestamp; } }
contract T {
    using Clock for uint256;
    uint256 public direct   = Clock.stamp(5);     // flagged
    uint256 public viaUsing = uint256(5).stamp(); // not flagged
}
```

Same view function, just called through `using for`.

crates/lint/src/sol/info/function_init_state.rs
0xMars42: Right, fixed in fea4cdf1: calls now judge only the overloads they can dispatch to. The visitor handles `Call` explicitly, judges the callee with the argument count (`parameters.len()` must match) and walks the base and arguments manually, so the callee is not re-judged without the count. A plain reference (function passed as a value) keeps the unfiltered behavior. `fromPureOverload = mix(1)` (clean) and `fromViewOverload = mix(1, 2)` (flagged) cover the mixed-purity pair from your example.

crates/lint/src/sol/info/function_init_state.rs
0xMars42: Fixed in fea4cdf1: the member arm walks `linearized_bases` instead of the contract's own items, so inherited functions and getters reached through a cast resolve. `fromInheritedGetter = OracleChild(address(0x1234)).price()` covers your `B(addr)` shape.

crates/lint/src/sol/info/function_init_state.rs
0xMars42: Fixed in fea4cdf1: when the base is not a contract, the member is looked up with `members_of` from the initializer's source and contract, which includes the `using for` attached functions; the receiver counts as the first parameter for the arity check. `fromUsingView = uint256(5).stamp()` (flagged) and `fromUsingPure = uint256(5).pureStamp()` (clean) cover it.

crates/lint/src/sol/info/function_init_state.rs
```diff
@@ -0,0 +1,213 @@
...
+            self.found = true;
+        }
+    }
+
+    /// A non-pure function observes the same partial state. Overload sets are filtered by
+    /// arity when the reference is a call. A variable referenced through its synthesized
+    /// getter is judged as a read of the variable itself, so a public constant stays fine.
+    fn judge_function(&mut self, function_id: FunctionId, arity: Option<usize>) {
+        let function = self.hir.function(function_id);
+        if arity.is_some_and(|count| function.parameters.len() != count) {
```
mablr: This overload filtering is still too broad because it only compares arity. If two overloads have the same number of parameters but different types, a call that resolves to the pure overload can be reported because a same-arity non-pure overload exists.

For example, this should stay clean but currently emits `function-init-state`:

```solidity
uint256 public fromPureSameArityOverload = typed(1);

function typed(uint256 x) internal pure returns (uint256) {
    return x + 1;
}

function typed(string memory) internal view returns (uint256) {
    return seed;
}
```

Please use the type checker's resolved callee for call expressions (`gcx.type_of_expr(callee.peel_parens().id)` / `TyKind::Fn(function_ty).function_id`) instead of re-resolving by name and arity.

crates/lint/src/sol/info/function_init_state.rs
```diff
@@ -0,0 +1,213 @@
...
+        // as `Type(Contract(..))`, while a contract-typed value comes bare.
+        let ty = ty.peel_refs();
+        let ty = match ty.kind {
+            TyKind::Type(inner) => inner.peel_refs(),
+            _ => ty,
+        };
+        if let TyKind::Contract(contract_id) = ty.kind {
+            // Walk the linearization: an inherited function or getter is not among the
+            // contract's own items.
+            for base_id in self.hir.contract(contract_id).linearized_bases {
```
mablr, line 187: Walking every function with this member name in the receiver contract's linearized bases can report the wrong dispatch target. If a child overrides a view/nonpayable base function with a pure implementation, a call to the child override is still reported because this loop also judges the base implementation.

Repro:

```solidity
contract OverrideBase {
    function maybePure() public view virtual returns (uint256) {
        return address(this).balance;
    }
}

contract OverrideChild is OverrideBase {
    function maybePure() public pure override returns (uint256) {
        return 1;
    }
}

uint256 public fromPureOverride = OverrideChild(address(0x1234)).maybePure();
```

This initializer should not warn because the selected override is pure. For call expressions, please use the resolved function from the callee type and reserve linearized/member scanning for non-call member reads where there is no resolved call target.

crates/lint/src/sol/info/function_init_state.rs
0xMars42: Fixed in ec43b134: calls now judge the single function the type checker resolved (`gcx.type_of_expr(callee.peel_parens().id)` then `TyKind::Fn.function_id`), so the arity filter is gone. Your repro is in the fixture as `fromPureSameArityOverload` (clean), with `fromViewSameArityOverload = typed("s")` next to it to check the view twin still reports. A call through a function pointer stored in state keeps judging the pointer variable (the callee's variable resolutions are still judged, `fromPointerCall` in the fixture), and `super` calls resolve like the rest (`fromSuperCall`).

crates/lint/src/sol/info/function_init_state.rs
0xMars42: Fixed in ec43b134: for call expressions the linearization walk is gone, the judged function is the one resolved from the callee type, and the name scan only remains for member reads outside a call position. Your repro is in the fixture as `fromPureOverride` (clean); `fromBaseTypedOverride = OverrideBase(address(0x1234)).maybePure()` still reports, since typing through the base resolves to the base's own view declaration. The `using for` calls also resolve through the callee type, so those cases are unchanged.

mablr: This still reports the wrong dispatch target for member references used as values. For example, adding `function() external view returns (uint256) public ref = OverrideChild(address(0x1234)).maybePure;` emits `function-init-state` even though the selected member is the child’s pure override; this loop walks into OverrideBase and judges its view implementation by name. Consider resolving the member expression’s function type when available, similar to `resolved_callee`, and only falling back to the linearized name scan when there is no resolved target.

0xMars42: Fixed in 00af2ab4: a member reference used as a value now judges the function resolved from the expression type, like a call target, and the linearized name scan only remains for member reads with no resolved function (`Base.stateVar` style). Your repro is in the fixture as `refPureOverride` (clean), with `refBaseTyped` next to it: typed through the base, the reference still reports since the resolved declaration is the base's view one. I also checked the twin shape on the `Ident` side: assigning an overloaded plain name by target type does not compile (solc 0.8.35 rejects it with error 2144), so there is no equivalent path there.

mablr (dismissed):
LGTM

mattsse (dismissed):

mablr (dismissed):
Resolved conflicts, G2G

emmajam (dismissed):

mablr (dismissed):

mablr (approved):
LGTM

figtracer (approved):

PR #15557, opened 2026-07-03
perf(fuzz): fold empty dictionary collection skips

decofe:
Folds small fuzz-state dictionary skip guards into one reviewable PR:

- #15314
- #15315
- #15317
- #15318

This intentionally does not fold #15303: applying that address guard breaks `same_address_with_new_bytecode_is_scanned_again`, because the same address can later have different bytecode that still needs scanning.

The folded changes skip work for empty inputs in fuzz dictionary collection paths:

- empty call value collection
- empty fuzzer value drain
- empty decoded log samples
- empty storage changes before storage-layout lookup

Validation:

- `cargo fmt --all`
- `cargo test -p foundry-evm-fuzz strategies::state --lib`

Prompted by: @grandizzy

### DIFF
--- a/crates/evm/fuzz/src/strategies/state.rs
+++ b/crates/evm/fuzz/src/strategies/state.rs
@@ -478,7 +478,7 @@ impl FuzzDictionary {
             // Insert push bytes.
             self.insert_push_bytes_values(address, &account.info);
             // Insert storage values.
-            if self.config.include_storage {
+            if self.config.include_storage && !account.storage.is_empty() {
                 let slot_identifier_key = targets.get(address).and_then(|contract| {
                     contract.storage_layout.as_ref().map(|layout| {
                         let key = Arc::as_ptr(layout) as usize;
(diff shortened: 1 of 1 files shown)


stevencartavia (approved):

mablr:
I benchmarked the empty dictionary-collection paths against `origin/master` with temporary targeted microbench tests inside `crates/evm/fuzz/src/strategies/state.rs`, built with Foundry's profiling profile. The benchmarks isolate the early-return cases added by this PR rather than relying on a full fuzz campaign, which was too noisy for these small hot-path changes.

Results:

| --- | ---: | ---: | ---: |
| empty fuzzer value drain | 2.185 ns/op | 1.086 ns/op | ~50.3% faster |
| empty `collect_values_from_call` | 2.881 ns/op | 1.841 ns/op | ~36.1% faster |

Recommendation: merge. The gains are targeted to empty/no-op collection cases, so end-to-end impact depends on how frequently those paths are hit, but the change is simple, low-risk early-return logic and clearly improves the isolated hot paths.

mablr (approved):

PR #15558, opened 2026-07-03
perf(evm): fold corpus mutation optimizations

decofe:
Folds the small corpus mutation perf PRs into one reviewable PR:

- #15300
- #15310

The folded changes are limited to corpus mutation bookkeeping:

- avoid storing all-empty cmp payload vectors
- avoid allocating a one-element mutation index vector for single-arg ABI mutation

Validation:

- `cargo fmt --all`
- `cargo test -p foundry-evm corpus --lib`

Prompted by: @grandizzy

### DIFF
--- a/crates/evm/evm/src/executors/corpus.rs
+++ b/crates/evm/evm/src/executors/corpus.rs
@@ -1579,8 +1579,9 @@ impl WorkerCorpus {
         // Mutate calldata.
         let mut arg_mutation_rounds =
             test_runner.rng().random_range(0..=function.inputs.len()).max(1);
-        let round_arg_idx: Vec<usize> = if function.inputs.len() <= 1 {
-            vec![0]
+        let single_input = function.inputs.len() <= 1;
+        let round_arg_idx: Vec<usize> = if single_input {
+            Vec::new()
         } else {
             (0..arg_mutation_rounds)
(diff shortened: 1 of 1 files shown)


stevencartavia (approved):

0xKarl98 (approved):

mablr:
I benchmarked the corpus changes against `origin/master` with temporary targeted microbench tests inside `corpus.rs`, built with Foundry's profiling profile. The benchmarks cover both changed paths: single-argument ABI mutation and all-empty cmp-sequence handling.

Results:

| --- | ---: | ---: | ---: |
| single-arg `abi_mutate` | 446.471 ns/op | 411.623 ns/op | ~7.8% faster |
| all-empty cmp-seq storage, 10k inputs | 48.504 ns/input, stores 10k empty cmp entries | 146.283 ns/input, stores 0 cmp entries | CPU slower, payload much smaller |

Recommendation: merge, but frame the cmp-sequence change as a memory/storage win rather than a CPU win. The single-argument mutation optimization is a clear targeted CPU improvement. The empty cmp-sequence path spends more CPU to detect and trim the all-empty payload, but avoids retaining/persisting thousands of empty cmp vectors, which is likely the more important behavior for corpus size and serialization overhead.

PR #15559, opened 2026-07-03
fix(script): ignore near-empty blocks in priority fee estimate

### LINKED ISSUE #13885: bug(script): inflated gas price estimate on Tempo

stevencartavia:
`forge script` estimated the priority fee as the median of non-zero `eth_feeHistory` rewards, so a one-off tip in an idle block could dominate the estimate. Rewards from near-empty blocks (gas used ratio below 10%) are now dropped before taking the median.

Fixes #13885

### DIFF
--- a/crates/common/src/provider/fee.rs
+++ b/crates/common/src/provider/fee.rs
@@ -17,6 +17,12 @@ const FEE_HISTORY_BLOCKS: u64 = 10;
 /// The minimum priority fee to provide, in wei.
 const MIN_PRIORITY_FEE: u128 = 1;
 
+/// Gas used ratio below which a block is treated as near-empty and its reward is
+/// dropped from sampling, so one-off tips in otherwise idle blocks do not
+/// dominate the cross-block median on low-traffic chains. Local heuristic: there
+/// is no protocol-defined utilization threshold for priority-fee estimation.
+const MIN_GAS_USED_RATIO: f64 = 0.1;
+
 /// The resolved EIP-1559 fees together with the base fee they were derived from.
 ///
 /// The base fee is surfaced so callers can present a fee breakdown (max fee /
@@ -84,8 +90,10 @@ where
             .into(),
     };
 
-    let max_priority_fee_per_gas =
-        estimate_priority_fee(fee_history.reward.as_deref().unwrap_or_default());
+    let max_priority_fee_per_gas = estimate_priority_fee(
+        fee_history.reward.as_deref().unwrap_or_default(),
+        &fee_history.gas_used_ratio,
+    );
 
     let (num, den) = preset.base_fee_multiplier();
     let max_fee_per_gas = base_fee_per_gas
@@ -137,11 +145,17 @@ pub fn resolve_broadcast_eip1559_fees(
     Ok(fees)
 }
 
-/// Estimates the priority fee as the median of the per-block sampled rewards,
-/// ignoring zero rewards and never returning less than [`MIN_PRIORITY_FEE`].
-fn estimate_priority_fee(rewards: &[Vec<u128>]) -> u128 {
-    let mut rewards =
-        rewards.iter().filter_map(|r| r.first()).filter(|r| **r > 0_u128).collect::<Vec<_>>();
+/// Estimates the priority fee as the median of the per-block rewards, ignoring
+/// zero rewards and rewards from blocks below [`MIN_GAS_USED_RATIO`], and never
+/// returning less than [`MIN_PRIORITY_FEE`].
+fn estimate_priority_fee(rewards: &[Vec<u128>], gas_used_ratio: &[f64]) -> u128 {
+    let mut rewards = rewards
+        .iter()
+        .zip(gas_used_ratio)
+        .filter(|(_, ratio)| ratio.is_finite() && ratio.clamp(0.0, 1.0) >= MIN_GAS_USED_RATIO)
+        .filter_map(|(reward, _)| reward.first().copied())
+        .filter(|reward| *reward > 0)
+        .collect::<Vec<_>>();
     if rewards.is_empty() {
         return MIN_PRIORITY_FEE;
     }
@@ -149,8 +163,9 @@ fn estimate_priority_fee(rewards: &[Vec<u128>]) -> u128 {
     rewards.sort_unstable();
 
     let n = rewards.len();
+    // `midpoint` avoids overflow when averaging the two middle values.
     let median =
-        if n % 2 == 0 { (*rewards[n / 2 - 1] + *rewards[n / 2]) / 2 } else { *rewards[n / 2] };
+        if n % 2 == 0 { rewards[n / 2 - 1].midpoint(rewards[n / 2]) } else { rewards[n / 2] };
 
     std::cmp::max(median, MIN_PRIORITY_FEE)
 }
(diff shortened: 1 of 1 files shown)


mattsse (dismissed):
lgtm

crates/common/src/provider/fee.rs
```diff
@@ -137,20 +145,27 @@ pub fn resolve_broadcast_eip1559_fees(
...
-    let mut rewards =
-        rewards.iter().filter_map(|r| r.first()).filter(|r| **r > 0_u128).collect::<Vec<_>>();
+/// Estimates the priority fee as the median of the per-block rewards, ignoring
+/// zero rewards and rewards from blocks below [`MIN_GAS_USED_RATIO`], and never
+/// returning less than [`MIN_PRIORITY_FEE`].
+fn estimate_priority_fee(rewards: &[Vec<u128>], gas_used_ratio: &[f64]) -> u128 {
+    let mut rewards = rewards
+        .iter()
+        .zip(gas_used_ratio)
+        .filter(|(_, ratio)| ratio.is_finite() && ratio.clamp(0.0, 1.0) >= MIN_GAS_USED_RATIO)
```
mablr: Pairing rewards with `gas_used_ratio` via `zip` silently drops any reward whose utilization entry is missing. Alloy deserializes `gasUsedRatio: null` as an empty `Vec`, so a response with usable rewards but null/truncated utilization now falls back to a 1 wei priority fee, where the old estimator still used the reward median. Idk if we need to  treat missing utilization as unknown/keep instead of near-empty/drop.

crates/common/src/provider/fee.rs
```diff
@@ -137,20 +145,27 @@ pub fn resolve_broadcast_eip1559_fees(
...
+        .filter_map(|(reward, _)| reward.first().copied())
+        .filter(|reward| *reward > 0)
+        .collect::<Vec<_>>();
     if rewards.is_empty() {
         return MIN_PRIORITY_FEE;
     }
```
mattsse: Three issues in this sampling chain:

1. **Chains that are chronically below the threshold collapse to a 1-wei tip.** A chain under steady light load (every block ~5% full with consistent, real 1-gwei tips in all 10 samples) has every sample dropped here, so the estimate becomes `MIN_PRIORITY_FEE` = 1 wei where the old code returned the 1-gwei median. Nodes that enforce a minimum tip (geth's `--miner.gasprice` floor on private/PoA chains, Polygon's ~25 gwei policy) will never mine that transaction, so `--broadcast` stalls waiting for receipts. A plain "fall back to the unfiltered median when the filtered set is empty" would reintroduce #13885 (that scenario is exactly all-blocks-idle with a one-off tip), so the fallback below only kicks in when tipping is the norm across the sample (at least half the blocks carry a non-zero reward).

2. **`zip` silently drops samples when `gasUsedRatio` does not line up with `reward`.** Alloy deserializes `gasUsedRatio: null` as an empty vec (`null_as_default` in `alloy-rpc-types-eth`, added precisely because such providers exist — see its `test_fee_hist_null_gas_used_ratio`), while `reward` stays populated. The `zip` then yields nothing and the estimate collapses to 1 wei on those providers. The filter should only apply when the two arrays have equal length.

3. `ratio.clamp(0.0, 1.0)` is dead logic: for every finite ratio the clamp never changes the outcome of the `>= MIN_GAS_USED_RATIO` comparison (>1 clamps to 1, which still passes, <0 clamps to 0, which still fails), and NaN/±inf are already excluded by `is_finite()`.

```suggestion
    // Apply the fullness filter only when `gasUsedRatio` lines up with `reward`:
    // providers may return `gasUsedRatio: null`, which alloy deserializes as an
    // empty vec, and `zip` would then silently drop every sample.
    let apply_filter = gas_used_ratio.len() == rewards.len();
    let non_zero_rewards = |filtered: bool| {
        rewards
            .iter()
            .enumerate()
            .filter(|(i, _)| {
                !filtered
                    || (gas_used_ratio[*i].is_finite()
                        && gas_used_ratio[*i] >= MIN_GAS_USED_RATIO)
            })
            .filter_map(|(_, reward)| reward.first().copied())
            .filter(|reward| *reward > 0)
            .collect::<Vec<_>>()
    };
    let mut rewards = non_zero_rewards(apply_filter);
    // If every sampled block is near-empty but tipping is the norm across the
    // sample (at least half the blocks carry a tip), fall back to the unfiltered
    // non-zero median; a one-off tip on an idle chain (#13885) stays excluded.
    if apply_filter && rewards.is_empty() {
        let unfiltered = non_zero_rewards(false);
        if unfiltered.len() * 2 >= gas_used_ratio.len() {
            rewards = unfiltered;
        }
    }
    if rewards.is_empty() {
        return MIN_PRIORITY_FEE;
    }
```

This changes two of the new tests, intentionally: `priority_fee_tolerates_length_mismatch` (a mismatch now falls back to the unfiltered non-zero median instead of silently truncating — the truncation is the bug, not behavior to pin down) and the NaN/inf case in `priority_fee_ignores_near_empty_blocks` (garbage ratios now fall back to the unfiltered median instead of 1 wei). The half-the-blocks heuristic is one option — outlier rejection relative to the filtered median would also work — but the empty-filter case needs *some* answer other than 1 wei.

crates/common/src/provider/fee.rs
```diff
@@ -159,15 +174,102 @@ fn estimate_priority_fee(rewards: &[Vec<u128>]) -> u128 {
...
+        let rewards = vec![vec![0u128], vec![41_000_000_000_000u128], vec![0u128]];
+        let ratios = vec![BUSY, 0.001, BUSY];
+
+        let priority = estimate_priority_fee(&rewards, &ratios);
+        let (num, den) = Eip1559FeeEstimatePreset::Market.base_fee_multiplier();
+        let max_fee = base_fee.saturating_mul(num) / den + priority;
+
+        assert_eq!(priority, MIN_PRIORITY_FEE);
+        assert_eq!(max_fee, 40_000_000_001);
     }
```
mattsse: This test is redundant: both filter interactions are already covered — busy block with zero reward by `priority_fee_median_of_busy_blocks`, near-empty block with an outlier reward by `priority_fee_ignores_near_empty_blocks` (same rewards) and end-to-end by `estimate_filters_near_empty_block_outliers` (same `40_000_000_001` expectation). The two filters are independent sequential closures, so combining them exercises no new branch, and the max-fee check re-derives `base_fee * num / den + priority` by hand — the same arithmetic `market_preset_max_fee_formula` already asserts.

```suggestion
```

crates/common/src/provider/fee.rs
```diff
@@ -209,39 +311,21 @@ mod tests {
...
-                let alloy = alloy_provider::utils::eip1559_default_estimator(base_fee, &rewards);
-                assert_eq!(ours, alloy, "mismatch for base_fee={base_fee}, rewards={rewards:?}");
-            }
-        }
+    fn priority_fee_preserves_high_tips_on_busy_chains() {
+        // Busy blocks keep the median of non-zero rewards, even when the tip is
+        // far above the base fee (legitimate congestion is not clamped).
+        let rewards = vec![vec![10_000_000_000u128], vec![30_000_000_000u128]];
+        assert_eq!(estimate_priority_fee(&rewards, &[BUSY, BUSY]), 20_000_000_000);
     }
```
mattsse: Redundant with the `estimate_priority_fee(&[vec![2], vec![4]], &[BUSY, BUSY]) == 3` case in `priority_fee_median_of_busy_blocks`: identical path (ratio filter → non-zero filter → even-length `midpoint` median), and there is no magnitude-dependent branch — only the constants differ.

```suggestion
```

mattsse (commented):
The near-empty-block filter is the right signal for #13885, but as implemented it converts the "estimate too high" failure into a silent "estimate too low → stuck transaction" failure on realistic setups: chains whose blocks are legitimately always below 10% utilization, and providers that return `gasUsedRatio: null`. Details and suggestions inline.

One issue that can't be commented inline because the files aren't in the diff: `cast send` (`crates/cast/src/tx.rs`) and `forge create` (`crates/forge/src/cmd/create.rs`) call alloy's `Provider::estimate_eip1559_fees()` and keep the unfiltered estimator, so the same chain state now produces different fees than `forge script`, and the #13885 symptom remains for cast. Worth either routing them through this estimator or stating that the script-only scope is intentional.

crates/common/src/provider/fee.rs
stevencartavia: [17833f8](https://github.com/foundry-rs/foundry/pull/15559/commits/17833f8237ad2da0f68e7a9c1e1e63d4f3c54062)

crates/common/src/provider/fee.rs
stevencartavia: [17833f8](https://github.com/foundry-rs/foundry/pull/15559/commits/17833f8237ad2da0f68e7a9c1e1e63d4f3c54062)

crates/common/src/provider/fee.rs
stevencartavia: [17833f8](https://github.com/foundry-rs/foundry/pull/15559/commits/17833f8237ad2da0f68e7a9c1e1e63d4f3c54062)

crates/common/src/provider/fee.rs
stevencartavia: [17833f8](https://github.com/foundry-rs/foundry/pull/15559/commits/17833f8237ad2da0f68e7a9c1e1e63d4f3c54062)

mablr (approved):
I don’t see any blockers. The rest can be addressed in a follow-up.

mattsse (approved):

PR #15560, opened 2026-07-03
chore(forge/assets): update workflows

StackOverflowExcept1on:


grandizzy:
@StackOverflowExcept1on makes sense but mind pls to pin their specific rev, supply chain best practice. ty!

StackOverflowExcept1on:
@grandizzy Should we be so concerned about this in smart contracts? What's the use case for keeping something secret in CI?

StackOverflowExcept1on:
@grandizzy done

grandizzy (approved):
ty!

grandizzy:
> @grandizzy Should we be so concerned about this in smart contracts? What's the use case for keeping something secret in CI?

was more concerned re mutable action tag that can change the code executed, imo better safe

mablr (approved):

PR #15562, opened 2026-07-03
perf(symbolic): simplify low-mask arithmetic

DaniPopes:
Symbolic arithmetic currently emits some fixed-width decomposition patterns in forms that are cheap to construct but unnecessarily hard for Z3. The worst case I found is SorellaLabs/angstrom at `contracts/test/libraries/X128MathLib.t.sol::check_matchesSolady_fullMulX128`, which compares `X128MathLib.fullMulX128` against Solady's `fullMulDiv(x, y, 1 << 128)`. The library reconstructs the high half with `mulmod(x, y, not(0))`, checks the upper 128 bits with `lt(p1, shl(128, 1))`, and stitches the result with `shl(128, p1)` and `shr(128, z)`. The generated SMT for that test contained a `mulmod(x, y, 2^128)` remainder as a 512-bit `bvurem`, `(product - low128(product)) / 2^128`, and unsigned comparisons between a word and its low 128-bit mask; the second Angstrom SMT query was only 1.8 KiB but timed out.

This normalizes those shapes while constructing symbolic expressions: multiplication by a power of two becomes a shift, `addmod` and `mulmod` by a power-of-two modulus become the corresponding operation masked to the low bits, `(a - (a & (2^n - 1))) / 2^n` becomes `a >> n`, and unsigned comparisons between `a` and `a & low_mask` fold to constants. I also checked the individual SMT dumps from the symbolic benchmark set: Solady and Farcaster queries were already small and fast, while the Angstrom query above was the outlier that needed Foundry-side normalization.

### Results

| Repository | Before | After | Solver before | Solver after |
| Vectorized/solady | 0.493 s | 0.447 s | 655 ms | 605 ms |
| SorellaLabs/angstrom | 5.580 s | 0.500 s | 5030 ms | 8 ms |
| farcasterxyz/contracts | 0.386 s | 0.320 s | 253 ms | 208 ms |

These measurements are from `foundry-bench` `forge_symbolic_test` using a local profiling build. The Angstrom run still reports the existing hard-arithmetic heuristic incompleteness, but the SMT queries now resolve in milliseconds instead of returning unknown at the 5 second solver timeout.

This PR was written with Codex assistance, including code generation and PR text.


mattsse (approved):
lgtm

mablr (approved):

PR #15563, opened 2026-07-03
feat(symbolic): emit solidity regression tests

mablr:
## Description

Closes OSS-352 by adding `--emit-regression` support for symbolic test runs. Confirmed symbolic counterexample artifacts can now be converted into ordinary Solidity regression tests, with configurable output paths and overwrite behavior.

The generated tests import the original contract, replay stateless or stateful counterexample calls through Forge, and are reported back in JSON results via `symbolic_regressions`. This makes minimized symbolic failures reproducible as normal `forge test` cases.

### Usage 
```sh
forge test --symbolic --minimize --emit-regression
forge test --match-path 'test/regressions/*.sol'
```

With a custom output directory:

```sh
forge test --symbolic --emit-regression --regression-out test/custom_regressions
```

### DIFF
--- a/crates/forge/src/multi_runner.rs
+++ b/crates/forge/src/multi_runner.rs
@@ -8,6 +8,7 @@ use crate::{
         ContractRunnerContext, InvariantCampaignScope, LIBRARY_DEPLOYER,
         count_runnable_invariant_campaign_anchors,
     },
+    symbolic_regression::SYMBOLIC_REGRESSION_MARKER,
 };
 use alloy_json_abi::{Function, JsonAbi};
 use alloy_primitives::{Address, Bytes, U256};
@@ -128,9 +129,15 @@ impl<FEN: FoundryEvmNetwork> MultiContractRunner<FEN> {
         let matcher = self.test_function_matcher();
         self.matching_contracts(filter).flat_map(move |(id, c)| {
             let identifier = id.identifier();
-            c.abi
-                .functions()
-                .filter(move |func| matcher.matches_test_function(filter, &identifier, func))
+            let generated_symbolic_regression = is_generated_symbolic_regression_contract(&c.abi);
+            c.abi.functions().filter(move |func| {
+                matcher.matches_test_function(
+                    filter,
+                    &identifier,
+                    func,
+                    generated_symbolic_regression,
+                )
+            })
         })
     }
 
@@ -145,9 +152,13 @@ impl<FEN: FoundryEvmNetwork> MultiContractRunner<FEN> {
             .filter(|(id, _)| filter.matches_path(&id.source) && filter.matches_contract(&id.name))
             .flat_map(move |(id, c)| {
                 let identifier = id.identifier();
-                c.abi
-                    .functions()
-                    .filter(move |func| matcher.test_function_kind(&identifier, func).is_any_test())
+                let generated_symbolic_regression =
+                    is_generated_symbolic_regression_contract(&c.abi);
+                c.abi.functions().filter(move |func| {
+                    matcher
+                        .test_function_kind(&identifier, func, generated_symbolic_regression)
+                        .is_any_test()
+                })
             })
     }
 
@@ -159,11 +170,17 @@ impl<FEN: FoundryEvmNetwork> MultiContractRunner<FEN> {
                 let source = id.source.as_path().display().to_string();
                 let name = id.name.clone();
                 let identifier = id.identifier();
+                let generated_symbolic_regression =
+                    is_generated_symbolic_regression_contract(&c.abi);
                 let tests = c
                     .abi
                     .functions()
                     .filter(|func| {
-                        let kind = matcher.test_function_kind(&identifier, func);
+                        let kind = matcher.test_function_kind(
+                            &identifier,
+                            func,
+                            generated_symbolic_regression,
+                        );
                         (!self.tcfg.fuzz_only
                             || matches!(
                                 kind,
@@ -886,10 +903,9 @@ impl<'a> TestFunctionMatcher<'a> {
         &self,
         contract_id: &str,
         func: &Function,
+        generated_symbolic_regression: bool,
     ) -> TestFunctionKind {
-        if is_symbolic_regression_contract(contract_id)
-            && !func.name.starts_with("test_regression_")
-        {
+        if generated_symbolic_regression && !func.name.starts_with("test_regression_") {
             return TestFunctionKind::Unknown;
         }
 
@@ -905,11 +921,12 @@ impl<'a> TestFunctionMatcher<'a> {
         filter: &dyn TestFilter,
         contract_id: &str,
         func: &Function,
+        generated_symbolic_regression: bool,
     ) -> bool {
         filter.matches_test_function_kind_in_contract(
             contract_id,
             func,
-            self.test_function_kind(contract_id, func),
+            self.test_function_kind(contract_id, func, generated_symbolic_regression),
         )
     }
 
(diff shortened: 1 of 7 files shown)


crates/forge/src/symbolic_regression.rs
```diff
@@ -0,0 +1,382 @@
...
+                write_call(
+                    &mut out,
+                    call,
+                    &format!("address({})", call.target),
+                    artifact.replay_semantics.fail_on_revert,
+                )?;
+            }
+            writeln!(
+                out,
+                "        __foundrySymbolicRegressionCall(address(this), hex\"{}\", 0, true);",
```
figtracer, line 271: this drops the post-invariant hook from sequence regressions. i reproduced a case where artifact replay fails after the invariant body passes, but the generated solidity only calls the invariant selector here and the regression passes.

crates/forge/src/symbolic_regression.rs
```diff
@@ -0,0 +1,382 @@
...
+) -> Result<Vec<SymbolicRegression>> {
+    let mut emitted = Vec::new();
+    let mut seen_tests = HashSet::new();
+    for artifact_ref in results {
+        let artifact = load_artifact(&artifact_ref.path)?;
+        let (_, contract) = artifact_source_and_contract(&artifact_ref.path, &artifact)?;
+        if contract.ends_with("_SymbolicRegression") {
+            continue;
+        }
+        let key = (
```
figtracer, line 52: this dedups by contract/test/kind, so multiple handler assertion artifacts under one invariant collapse to a single emitted file. i reproduced two confirmed handler artifacts for one invariant and only the first regression was written.

mablr: https://github.com/foundry-rs/foundry/pull/15563/commits/4596ffd3261a6209a15ca388d97a7fda6df4b5aa

crates/forge/src/symbolic_regression.rs
mablr: https://github.com/foundry-rs/foundry/pull/15563/commits/69bc1a9b8183f7f3a6fba1b2482df96bc418e095

crates/forge/src/symbolic_regression.rs
```diff
@@ -0,0 +1,429 @@
...
+    writeln!(out, "        }}")?;
+    writeln!(out, "    }}")?;
+    writeln!(out, "}}")?;
+
+    if path.exists() && !regression.overwrite {
+        if std::fs::read_to_string(&path).is_ok_and(|existing| existing == out) {
+            return Ok(SymbolicRegression { artifact: artifact_path.to_path_buf(), path });
+        }
+        bail!(
+            "regression test {} already exists; pass --regression-overwrite to replace it",
```
stevencartavia: If the target regression file already exists with different content and `--regression-overwrite` isn't set, `emit_symbolic_regression` bails and the `?` in `run_tests_inner` aborts the whole run. In `--json`/`--junit` the bail happens before results are printed, so stdout is empty — exit 1, stderr `regression test ... already exists`, and the entire machine-readable payload is dropped. An auxiliary emit step shouldn't discard the primary result of a completed run; suggest downgrading the conflict to `sh_warn!` + skip, keeping hard errors for real I/O/schema failures.

crates/forge/src/symbolic_regression.rs
```diff
@@ -0,0 +1,429 @@
...
+    writeln!(out, "import \"{import_path}\";")?;
+    writeln!(out)?;
+    writeln!(out, "interface {vm_interface} {{")?;
+    writeln!(out, "    function deal(address who, uint256 newBalance) external;")?;
+    writeln!(out, "    function prank(address msgSender) external;")?;
+    writeln!(out, "    function roll(uint256 newHeight) external;")?;
+    writeln!(out, "    function warp(uint256 newTimestamp) external;")?;
+    writeln!(out, "}}")?;
+    writeln!(out)?;
+    writeln!(out, "contract {generated_contract} is {contract} {{")?;
```
stevencartavia: The generated contract is `X_Y_SymbolicRegression is <OriginalTestContract>` under `test/regressions/`, so it inherits the original's `test*`/`invariant*`/`check*` functions. The `ends_with("_SymbolicRegression")` guard only prevents emitting nested regressions, not re-running these: a plain `forge test` runs `invariant_v` twice (original suite plus a second full campaign in the regression contract), and a `--symbolic` run re-symbolic-executes the inherited test and writes a cache dir named after the regression contract. Every emitted regression permanently doubles that contract's campaign in CI and reintroduces nondeterministic fuzzing into files meant to be deterministic. Suggest restricting `*_SymbolicRegression` contracts to `test_regression_*` functions during test collection.

crates/forge/src/symbolic_regression.rs
```diff
@@ -0,0 +1,429 @@
...
+    let contract_ident = sanitize_identifier(contract);
+    let test_ident = sanitize_identifier(test_name);
+    let generated_contract = suffix.map_or_else(
+        || format!("{contract_ident}_{test_ident}_SymbolicRegression"),
+        |suffix| format!("{contract_ident}_{test_ident}_{suffix}_SymbolicRegression"),
+    );
+    let vm_interface = format!("{generated_contract}_Vm");
+    let generated_test = format!("test_regression_{test_ident}_symbolic");
+    let default_path = config.test.join("regressions").join(format!("{generated_contract}.t.sol"));
+    let path = regression.out.as_ref().map_or(default_path, |out| {
```
stevencartavia, line 211: With `--regression-out <file>.sol`, every artifact resolves to the same path, and with `--regression-overwrite` each write clobbers the previous one. `attach_symbolic_regressions` still records a ref per test, so the JSON reports both `checkFirst` and `checkSecond` at `test/onlyone.sol` while the file contains only the last-written `Second_checkSecond_SymbolicRegression`. The default path collides the same way for two same-named contracts in different files with the same test name — the source path is dropped after `rsplit_once(':')`, so the filename is contract+test name only (without `--regression-overwrite` this instead trips the "already exists" bail mid-run). An in-run collision means two distinct regressions want one path; that should be a hard error naming both tests before writing, not a silent overwrite — kept distinct from the stale-file check above.

crates/forge/src/symbolic_regression.rs
```diff
@@ -0,0 +1,479 @@
...
+        writeln!(
+            out,
+            "        __foundrySymbolicVm.warp(block.timestamp + {});",
+            u256_literal(warp)
+        )?;
+    }
+    if let Some(roll) = call.roll {
+        writeln!(out, "        __foundrySymbolicVm.roll(block.number + {});", u256_literal(roll))?;
+    }
+    if let Some(value) = call.value
```
mattsse, line 367: This gives the sender the full requested value before replaying the generated call, so value-bearing regressions do not match the artifact replay semantics. The generated code funds `address(this)` and `call.sender` here: https://github.com/foundry-rs/foundry/blob/ca410d0ca8557d534864cf28f392c194e251a4f1/crates/forge/src/symbolic_regression.rs#L355-L364, but invariant artifact replay clamps `msg.value` to the sender's live balance before `call_raw`: https://github.com/foundry-rs/foundry/blob/ca410d0ca8557d534864cf28f392c194e251a4f1/crates/evm/evm/src/executors/invariant/mod.rs#L2228-L2238. That can make a confirmed payable counterexample execute a different branch, or pass, once emitted as Solidity.

Suggested fix: emit the same clamp into the regression helper instead of pre-funding the sender to the requested value. For sequence calls, compute `uint256 value = requestedValue <= call.sender.balance ? requestedValue : call.sender.balance` before the low-level call, and only `deal(address(this), value)` when the regression contract needs enough balance to forward that clamped value. For single-call artifacts, keep the existing exact value behavior if that path is intentionally not clamped.

mattsse (commented):
One remaining replay-fidelity issue in the generated Solidity regressions.

crates/forge/src/multi_runner.rs
```diff
@@ -887,6 +887,12 @@ impl<'a> TestFunctionMatcher<'a> {
         contract_id: &str,
         func: &Function,
     ) -> TestFunctionKind {
+        if is_symbolic_regression_contract(contract_id)
```
stevencartavia: This gate runs unconditionally during test discovery, not just under `--emit-regression`. Any user contract named `*_SymbolicRegression` has all its non-`test_regression_` functions classified as `Unknown`, so its tests are silently skipped. Reproduced: a `Real_SymbolicRegression is Test` with a failing test is dropped by plain `forge test` and the run exits 0 (green CI with a failing test present). Suggest detecting generated contracts via an ABI marker emitted into the contract instead of matching on the name suffix.

crates/forge/src/symbolic_regression.rs
mablr: https://github.com/foundry-rs/foundry/pull/15563/commits/e17a711d6f1fc0ff5dcb0ddc2e450dc9a1d37110

figtracer (approved):

mattsse (approved):

PR #15564, opened 2026-07-03
refactor(symbolic): defer gasleft solver errors

DaniPopes:
This lets `gasleft()` participate in local symbolic reasoning as a synthetic variable while keeping the raw `GasLeft` expression kind for call-gas handling. Unresolved gasleft expressions still fail before SMT is sent, at assertion emission, so the solver never accidentally models gas consumption.

The concrete-use guards remain in the opcode, call, and log paths where gasleft would need to become an actual EVM value. AI assistance was used in preparing this change.


mattsse (approved):
lgtm

mablr (approved):
Makes sense

grandizzy (approved):
lgtm

PR #15566, opened 2026-07-03
fix(forge): `fuzz show` with ambiguous contract names

mablr:
Fixes `forge fuzz show` so it does not invent a contract name when a corpus entry's calldata can decode against the same function signature in multiple artifacts.

Instead of printing a misleading `Contract.function(...)`, ambiguous entries now show the function call without a contract prefix and include the candidate contract names. JSON output mirrors that with `contract: null` and `ambiguous_contracts`.

Follow-up to #15227.


mattsse (approved):
lgtm

stevencartavia (approved):

PR #15567, opened 2026-07-03
chore(forge): clean up `fuzz cmin` replay diagnostics

mablr:
## Description

#15227 #15517 follow-up cleaning up `forge fuzz cmin` zero-replay diagnostics by removing a redundant successful replay counter that can only be zero in that error path.

The fallback message now reports only the remaining meaningful counters, while the specific unreadable, failed, unmatched, rejected, and empty corpus diagnostics stay unchanged.


stevencartavia (approved):

mattsse (approved):
lgtm

PR #15568, opened 2026-07-03
fix(cast): route send fee estimation through shared estimator

stevencartavia:
`cast send` called alloy's unfiltered EIP-1559 estimator, so it disagreed with `forge script`, which was switched to Foundry's estimator in #15559.

It now uses the same `estimate_eip1559_fees` estimator as `forge script`, so both apply the near-empty-block filter and produce identical fees for the same chain state.

### DIFF
--- a/crates/cast/src/cmd/erc20.rs
+++ b/crates/cast/src/cmd/erc20.rs
@@ -577,6 +577,16 @@ impl Erc20Subcommand {
                                 .await?;
                             sponsor.attach_and_print::<N>(&mut tx, from).await?;
                         }
+                    } else {
+                        // Fill only the fees; the provider fills nonce and gas limit.
+                        fill_transaction_gas_fees(
+                            &$provider,
+                            &mut tx,
+                            chain.is_legacy(),
+                            false,
+                            config.eip1559_fee_estimate,
+                        )
+                        .await?;
                     }
                     cast_send(
                         $provider,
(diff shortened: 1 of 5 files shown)


crates/cast/src/cmd/erc20.rs
```diff
@@ -528,7 +542,15 @@ impl Erc20Subcommand {
...
-                        fill_tx(&$provider, &mut tx, from, chain, false).await?;
+                        fill_tx(
+                            &$provider,
+                            &mut tx,
+                            from,
+                            chain,
+                            false,
+                            config.eip1559_fee_estimate,
+                        )
+                        .await?;
```
mablr, line 553: This only routes ERC20 fee estimation through the shared estimator for the sponsor/hash path. Plain signer ERC20 sends still skip this helper and submit the sol!-built transaction with fees unset, so they continue using the provider filler instead of the shared estimator. Should the normal `cast erc20 transfer/approve/mint/burn` path also call this fill helper before `cast_send` when fees are missing?

stevencartavia: [dd4289e](https://github.com/foundry-rs/foundry/pull/15568/commits/dd4289ec2523ddca9b36ad04cff552b7f586dde2)

mablr (approved):

mattsse (approved):
lgtm

PR #15570, opened 2026-07-03
fix(forge): route create fee estimation through shared estimator

stevencartavia:
`forge create` called alloy's unfiltered EIP-1559 estimator, so it disagreed with `forge script`, which was switched to Foundry's estimator in #15559.

It now uses the same `estimate_eip1559_fees` estimator as `forge script`, so both apply the near-empty-block filter and produce identical fees for the same chain state.


mablr (approved):
makes sense

mattsse (approved):
lgtm

PR #15571, opened 2026-07-03
test(forge): `fuzz tmin` semantics coverage

mablr:
Adds guardrail coverage for `forge fuzz tmin` replay semantics.

The tests lock down that candidate-only active targets are rejected when they produce failures or replayed transactions, and that calldata shrink candidates are not accepted when they turn a successful replay into `vm.assume` or `vm.skip`.

Follow-up to #15227 #15549.


mattsse (approved):
lgtm

PR #15572, opened 2026-07-03
refactor(symbolic): intern symbol names

DaniPopes:
This changes symbolic expressions to store variables and generated symbolic terms as compact symbol ids while keeping display names in the SymCx interner. It also centralizes symbol-bearing expression handling through helper methods and removes the crate-local ptr_eq helpers.

### DIFF
--- a/crates/evm/symbolic/src/runtime/expr/hashcons.rs
+++ b/crates/evm/symbolic/src/runtime/expr/hashcons.rs
@@ -1,6 +1,7 @@
 use alloy_primitives::map::foldhash::fast::FixedState;
 use hashbrown::{HashTable, hash_table::Entry};
 use std::{
+    cmp::Ordering,
     fmt,
     hash::{BuildHasher, Hash, Hasher},
     sync::{Arc, Weak},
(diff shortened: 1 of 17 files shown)


mattsse (dismissed):
lgtm

crates/evm/symbolic/src/runtime/expr/hashcons.rs
```diff
@@ -94,15 +89,23 @@ impl<T> HashConsEntry<T> {
 
 impl<T> HashCons<T> {
     pub(in crate::runtime) fn new() -> Self {
-        Self { table: HashTable::new(), hash_builder: HashBuilder::default() }
+        Self::with_hasher(DefaultHashBuilder::default())
```
stevencartavia: Switching `HashCons` to `foldhash::fast::RandomState` (random seed per instance) can make SMT output non-deterministic across runs, since `ordered_commutative_operands` canonicalizes by `cached_hash()`. On master the fixed-seed hasher kept this stable.

Repro: `x + y` built in 64 fresh `SymCx`:
- master → `(bvadd bbbbbbbb aaaaaaaa)` (always)
- this PR → `(bvadd aaaaaaaa bbbbbbbb)` or `(bvadd bbbbbbbb aaaaaaaa)`

`stable_symbol` names vary too, since it hashes the `Debug` repr of the expressions, which embeds the hash-dependent operand order.

Not a correctness issue, but `dump_smt` is no longer reproducible across runs. `foldhash::fast::FixedState` would fix it if that matters.

DaniPopes: fixed and also now puts constants on the right allowing pattern matching to only have to check one side. documented in agents.md

figtracer (approved):

stevencartavia (approved):

mablr (approved):

PR #15577, opened 2026-07-04
fix(cast): make source output deterministic

### LINKED ISSUE #13577: Allow cast source to be used by diff regardless of the source order
### Component

Cast

cast source and diff is supposed to be used [like so](https://github.com/foundry-rs/foundry/pull/13532#issuecomment-3952864661):
`diff <(cast source ..) <(cast source ..)`

However there is a few instances where that doesn't work, mainly because the source code order is assumed to be consistent, which turns out not to be the case for all contracts. 

Take mainnet 0xDb549616407f8a30799f77F12b6b85AEC936782D and 0xa76645E15c267b876999bf7689E0b2C1EE29BFE6 [for example](https://github.com/foundry-rs/foundry/pull/13532#issuecomment-3953047591):

Idiomatic usage yields a 10k diff:
```
➜  diff <(cast source 0xDb549616407f8a30799f77F12b6b85AEC936782D) <(cast source 0xa76645E15c267b876999bf7689E0b2C1EE29BFE6) | wc -l 
   10776
```

Where a file based diff yields a much more readable 1k diff:
```
➜ cast source -d v1 0xDb549616407f8a30799f77F12b6b85AEC936782D
➜ cast source -d v2 0xa76645E15c267b876999bf7689E0b2C1EE29BFE6
➜ diff -ruN ./v1 ./v2 | wc -l                                                                                                 
    1060
```

### Additional context

I proposed to add a new command in https://github.com/foundry-rs/foundry/pull/13532 . This PR got rejected but I'm not sure maintainer was aware of this issue.
Another solution would be to add a `cast source` option that forces source ordering for diffs (`--force-order` or `--sort-sources` or `--sort`?)

stevencartavia:
Closes #13577 

`cast source` produced nondeterministic output for multi-file contracts. The fix lives upstream in foundry-rs/foundry-core#109, so this PR consumes it via a block-explorers bump and relies on `metadata.source_code()` for the sorted output.

### DIFF
--- a/crates/cast/src/lib.rs
+++ b/crates/cast/src/lib.rs
@@ -31,10 +31,7 @@ use alloy_transport::TransportErrorKind;
 use base::{Base, NumberWithBase, ToBase};
 use chrono::DateTime;
 use eyre::{Context, ContextCompat, OptionExt, Result};
-use foundry_block_explorers::{
-    Client,
-    contract::{ContractMetadata, SourceCodeEntry},
-};
+use foundry_block_explorers::Client;
 use foundry_common::{
     abi::{coerce_value, encode_function_args, encode_function_args_packed, get_event, get_func},
     compile::etherscan_project,
@@ -53,7 +50,6 @@ use rayon::prelude::*;
 use serde::Serialize;
 use std::{
     borrow::Cow,
-    collections::HashMap,
     fmt::Write,
     io,
     marker::PhantomData,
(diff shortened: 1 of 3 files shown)


mattsse (dismissed):
lgtm

crates/cast/src/lib.rs
```diff
@@ -2260,7 +2264,7 @@ impl SimpleCast {
     ) -> Result<String> {
         let client = explorer_client(chain, etherscan_api_key, explorer_api_url, explorer_url)?;
         let metadata = client.contract_source_code(contract_address.parse()?).await?;
-        Ok(metadata.source_code())
+        Ok(format_etherscan_source(&metadata))
```
DaniPopes: please fix in foundry core

DaniPopes (changes requested):

crates/cast/src/lib.rs
stevencartavia: https://github.com/foundry-rs/foundry-core/pull/109

mablr (approved):
given https://github.com/foundry-rs/foundry-core/pull/109 LGTM

figtracer (approved):

DaniPopes (approved):

PR #15579, opened 2026-07-04
feat(lint): add deprecated-oz-function detector

0xMars42:
Adds the `deprecated-oz-function` lint from #14381 (Low).

Reports a reference, called or used as a value, that resolves to a function named `safeApprove` declared in a library named `SafeERC20` / `SafeERC20Upgradeable`, or `_setupRole` declared in a contract named `AccessControl` / `AccessControlUpgradeable`. Both were deprecated in the OpenZeppelin 4.x line and removed in 5.0; the deprecation notes document `safeIncreaseAllowance` / `safeDecreaseAllowance` and `_grantRole` as the replacements.

Resolution goes through the type checker (`gcx.type_of_expr` on the expression, `TyKind::Fn.function_id`), so the `using for` method form, the library-qualified form, import aliases (fixture `UsesAliasedImport` with an auxiliary file), inheritance through extensions that do not redeclare the function (`AccessControlEnumerable` style) and references used as values are all recognized. A same-name function declared outside the canonical OZ names stays clean (`TokenUtils.safeApprove`, a standalone `_setupRole`), and an override of `_setupRole` is the dispatch target of plain calls while its `super._setupRole` delegation still reports, since the deprecation is about using the API at all.

Aderyn's detector matches any identifier named `_setupRole` and any member named `safeApprove` in files importing a path containing "openzeppelin"; resolving instead avoids those false positives and keeps aliased imports covered.

Checked against the real OpenZeppelin 4.9.6 sources: a consumer contract using both deprecated functions and both replacements reports exactly the three deprecated references.


crates/lint/src/sol/low/deprecated_oz_function.rs
```diff
@@ -0,0 +1,101 @@
...
+        let function = self.hir.function(function_id);
+        let Some(name) = function.name else { return false };
+        let Some(contract_id) = function.contract else { return false };
+        let contract = self.hir.contract(contract_id);
+        let contract_name = contract.name.as_str();
+        // `safeApprove` lives in the SafeERC20 library, `_setupRole` in the AccessControl
+        // contract: requiring the matching declaration kind tightens the match.
+        if name.as_str() == "safeApprove" {
+            contract.kind.is_library()
+                && matches!(contract_name, "SafeERC20" | "SafeERC20Upgradeable")
```
mablr, line 86: This currently treats any project-local library named `SafeERC20` as OpenZeppelin. I verified `forge lint --only-lint deprecated-oz-function` reports a warning for a local same-named library with `safeApprove`, so this needs an OZ source/import check (or the docs/tests need to intentionally allow same-name false positives).

crates/lint/src/sol/low/deprecated_oz_function.rs
```diff
@@ -0,0 +1,101 @@
...
+        } else if name.as_str() == "_setupRole" {
+            !contract.kind.is_library()
+                && matches!(contract_name, "AccessControl" | "AccessControlUpgradeable")
```
mablr, line 89: Same issue for `_setupRole`: matching only the declaring contract name flags local contracts named `AccessControl`. I verified a local same-named base contract with `_setupRole` is reported even without any OpenZeppelin import.

crates/lint/src/sol/low/deprecated_oz_function.rs
0xMars42: Fixed in b6afcee0b: the resolved declaration must also come from an OpenZeppelin package path (`lib/openzeppelin-contracts`, `@openzeppelin/...`), so a local `SafeERC20` library stays clean. The testdata now declares one and locks that in; the canonical mirrors moved under an `openzeppelin-contracts` auxiliary path. The trade-off, stated in the docs: a vendored OZ copy under a path that does not name OpenZeppelin is not recognized.

crates/lint/src/sol/low/deprecated_oz_function.rs
0xMars42: Same fix in b6afcee0b: the provenance check applies to `_setupRole` too, so a local `AccessControl` base contract stays clean, with a testdata case covering it.

crates/lint/src/sol/low/deprecated_oz_function.rs
```diff
@@ -0,0 +1,120 @@
...
+        } else {
+            false
+        }
+    }
+
+    /// Whether a source file belongs to an OpenZeppelin package, judged by its path. A
+    /// vendored copy under a path that does not name OpenZeppelin is not recognized.
+    fn is_openzeppelin_source(&self, source_id: hir::SourceId) -> bool {
+        match &self.hir.source(source_id).file.name {
+            FileName::Real(path) => path.to_string_lossy().to_lowercase().contains("openzeppelin"),
```
mablr: This still accepts any path containing `openzeppelin`, so a local or third-party path like `src/notopenzeppelin/LocalSafeERC20.sol` with its own `SafeERC20.safeApprove` is reported as deprecated OZ. The path check needs segment/package matching (for example `@openzeppelin` or `openzeppelin-contracts`) instead of a raw substring.

crates/lint/src/sol/low/deprecated_oz_function.rs
```diff
@@ -0,0 +1,120 @@
...
+
+impl<'hir> LateLintPass<'hir> for DeprecatedOzFunction {
+    fn check_function(
+        &mut self,
+        ctx: &LintContext,
+        gcx: Gcx<'hir>,
+        hir: &'hir Hir<'hir>,
+        func: &'hir hir::Function<'hir>,
+    ) {
+        if let Some(body) = &func.body {
```
figtracer: This only walks function bodies, so value references outside the body are missed even though the lint reports references too. A state variable function-pointer initializer or a modifier argument using `SafeERC20.safeApprove` would not be visited. Consider moving the resolved-function predicate to a normal expression hook so all expression sites are covered.

mattsse (dismissed):
No additional independent findings from my pass, but the existing unresolved review comments still look blocking:

- `crates/lint/src/sol/low/deprecated_oz_function.rs:116`: the OpenZeppelin provenance check still accepts any path containing `openzeppelin`, so misleading local/third-party paths can false-positive.
- `crates/lint/src/sol/low/deprecated_oz_function.rs:31`: the detector only walks function bodies, so non-body expression references such as state variable function-pointer initializers or modifier arguments are missed.

0xMars42:
Both fixed in `b2ee09144`. Provenance matches a full path component against the package roots instead of a substring, with a negative fixture (local `SafeERC20`/`AccessControl` under `not-openzeppelin/`). Non-body references: the detector now judges every expression the linter visits instead of walking function bodies only, so a reference in a header (a modifier or base-constructor argument) is caught. Added a fixture where a deprecated function referenced in a modifier argument reports.

mablr (approved):
LGTM

mattsse (approved):
The prior blockers are fixed: the lint now covers non-body expressions via expression checks, and OpenZeppelin provenance is matched on package/root path components instead of raw substrings. No new actionable findings in this pass.

figtracer (approved):

PR #15580, opened 2026-07-04
feat(lint): add solmate-safe-transfer-lib detector

0xMars42:
Adds the `solmate-safe-transfer-lib` lint from #14381 (Low).

Reports a reference, called or used as a value, that resolves to `safeTransfer`, `safeTransferFrom` or `safeApprove` declared in a library named exactly `SafeTransferLib`. In the released solmate v6, a token call that returns no data is treated as a success without checking that the token has code (`success := 1` on the empty-return path), so a token operation against a token-less address is a silent no-op; the unreleased main branch has since added a code check to that path. `safeTransferETH` involves no token code and stays clean.

Resolution goes through the type checker, so the `using for` method form, the library-qualified form, import aliases (fixture `UsesAliasedImport` with an auxiliary file) and references used as values are recognized, while same-name functions of other libraries stay out (`TransferHelper` style, in the fixture).

Aderyn's detector flags the import directive whose path contains `solmate` and `SafeTransferLib`; resolving the calls anchors the warning where the risk sits, skips files importing the library without using it, and keeps vendored copies under the same name covered.

Checked against the real solmate v6 sources: a consumer using the three token operations and `safeTransferETH` reports exactly the three token operations.


crates/lint/src/sol/low/solmate_safe_transfer_lib.rs
```diff
@@ -0,0 +1,92 @@
...
+        gcx: Gcx<'hir>,
+        hir: &'hir Hir<'hir>,
+        func: &'hir hir::Function<'hir>,
+    ) {
+        if let Some(body) = &func.body {
+            let mut finder = TokenOpFinder { gcx, hir, ctx };
+            for stmt in body.stmts {
+                let _ = finder.visit_stmt(stmt);
+            }
+        }
```
stevencartavia, line 36: This would be simpler as a `check_expr` lint (like `return_bomb.rs`) instead of a custom `hir::Visit` from `check_function`. Since the custom finder only walks `func.body`, it misses reference sites the late visitor already walks for `check_expr` passes — at least state-variable initializers and modifier invocation args:

```solidity
contract StateVarInit {
    function(IToken, address, uint256) internal fn = SafeTransferLib.safeTransfer;
}

contract ModArg {
    modifier withFn(function(IToken, address, uint256) internal f) { _; }
    function run() external withFn(SafeTransferLib.safeTransfer) {}
}
```

Moving the `Ident | Member` / `resolved_function` predicate into `check_expr` fixes these and drops `TokenOpFinder`, while keeping the existing fixture warnings unchanged. Not a blocker for me (Low severity, uncommon positions), just a simplification.

crates/lint/docs/solmate-safe-transfer-lib.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+### Good
+
+```solidity
+using SafeTransferLib for ERC20;
+
+function pay(ERC20 token, address to, uint256 amount) internal {
+    require(address(token).code.length > 0, "token has no code");
+    token.safeTransfer(to, amount);
```
mablr, line 36: This "Good" example still reports today: the lint emits on any resolved `SafeTransferLib.safeTransfer` reference and does not recognize the preceding `address(token).code.length > 0` guard. Either make guarded calls clean and add a fixture, or change this to an example that is actually clean under the lint (for example OZ `SafeERC20`) and document that guarded Solmate calls need suppression.

crates/lint/src/sol/low/solmate_safe_transfer_lib.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// concern does not apply to it. A same-name function of another library (Uniswap's
+    /// `TransferHelper` style) stays out through the resolution.
+    fn is_unchecked_token_op(&self, function_id: FunctionId) -> bool {
+        let function = self.hir.function(function_id);
+        let Some(name) = function.name else { return false };
+        let Some(contract_id) = function.contract else { return false };
+        let contract = self.hir.contract(contract_id);
+        matches!(name.as_str(), "safeTransfer" | "safeTransferFrom" | "safeApprove")
+            && contract.kind.is_library()
+            && contract.name.as_str() == "SafeTransferLib"
```
mablr, line 99: Matching only `library SafeTransferLib` plus these function names will flag implementations that do not have the stated Solmate-v6 bug. For example, Solady-style `SafeTransferLib` uses the same library/function names but checks token code on the empty-return success path. Please either constrain this to Solmate-v6/source/body semantics, or add an explicit fixture documenting that fixed/Solady `SafeTransferLib` copies are intentional false positives.

0xMars42: Fixed in ee0bfeb81 with the provenance option, consistent with the deprecated-oz-function fix: the resolved declaration must come from a solmate package path, so Solady's same-name library stays clean. The testdata covers a Solady-style copy (code check on the empty-return path, path naming solady) as clean, and the docs state the flip side: a vendored solmate copy under a path that does not name solmate is not recognized.

crates/lint/docs/solmate-safe-transfer-lib.md
0xMars42: Fixed in ee0bfeb81 with the second option: the "Good" example now shows OpenZeppelin's SafeERC20, which is actually clean under the lint, and a note explains that a code-length guarded solmate call still reports and needs an explicit `forge-lint: disable-next-line(solmate-safe-transfer-lib)` suppression. Recognizing the guard would take flow analysis that the resolution-based approach does not do.

crates/lint/src/sol/low/solmate_safe_transfer_lib.rs
```diff
@@ -0,0 +1,110 @@
...
+
+declare_forge_lint!(
+    SOLMATE_SAFE_TRANSFER_LIB,
+    Severity::Low,
+    "solmate-safe-transfer-lib",
+    "Solmate's `SafeTransferLib` does not check that the token has code, so a transfer to a token-less address succeeds silently"
+);
+
+impl<'hir> LateLintPass<'hir> for SolmateSafeTransferLib {
+    fn check_function(
```
mablr, line 24: The normal `token.safeTransfer(...)` call reports, but `function(...) internal fn = SafeTransferLib.safeTransfer` and `withFn(SafeTransferLib.safeTransfer)` do not. Since this lint intentionally reports references used as values, please move this predicate to `check_expr` (or otherwise walk all HIR expression sites), and add fixture coverage for these non-body expression positions.

mattsse (changes requested):
Found one blocker:

- `crates/lint/src/sol/low/solmate_safe_transfer_lib.rs:106`: the provenance check still uses `path.to_string_lossy().to_lowercase().contains("solmate")`. That accepts unrelated paths like `src/not-solmate/SafeTransferLib.sol`, `vendor/solmate-fixed/SafeTransferLib.sol`, or project directories containing the word, even when the implementation is not the vulnerable Solmate library. Please match path components/package roots instead, and add a negative fixture for a same-name local/fixed library under a misleading path containing `solmate`.

0xMars42:
Fixed in `3a8ee6324`. Provenance now matches a full path component instead of a substring, so `vendor/solmate-fixed/` and `src/not-solmate/` no longer count as solmate. Added a negative fixture: a local `SafeTransferLib` under `solmate-fixed/` stays out of scope, proven by reverting to the substring check, which makes it fire.

mattsse (approved):
The prior import provenance blocker is fixed by matching full `solmate` path components instead of raw substrings, with negative fixture coverage. No new actionable findings in this pass.

mablr (approved):

PR #15581, opened 2026-07-04
feat(lint): add incorrect-using-for detector

0xMars42:
Adds the `incorrect-using-for` lint from #14381 (Info).

Reports a `using ... for` directive, file-level or contract-level, when no function of the named library accepts the target type as its bound first parameter: the directive attaches nothing, which the compiler accepts silently and is usually a typo in the library or the type.

Attachment is delegated to the type checker: the directive is dead when `members_of` gains no attached function declared in the named library for the target type, so implicit conversions (integer widening, derived to base contracts, storage to memory) follow the compiler. Slither's detector of the same name reimplements the convertibility rules by hand instead. `using L for *` is out of scope, and so is the braced form `using {f} for T`, which the compiler already rejects on a mismatch.

`check_nested_source` gains the `Gcx` parameter, mirroring `check_nested_contract`, since source-level directives need type information; no existing pass overrode it.

Checked against the real OpenZeppelin sources: `using SafeERC20 for IERC20` stays clean and a deliberate `using Strings for IERC20` reports.


crates/lint/src/sol/info/incorrect_using_for.rs
```diff
@@ -0,0 +1,91 @@
...
+        gcx: Gcx<'hir>,
+        hir: &'hir Hir<'hir>,
+        directive: &'hir UsingDirective<'hir>,
+    ) {
+        // `using L for *` attaches every function of the library: nothing to validate.
+        let Some(hir_ty) = &directive.ty else { return };
+        // `members_of` expects reference types wrapped in their data location. Storage
+        // converts implicitly to memory, so the storage form covers library functions
+        // taking either location.
+        let ty = gcx.type_of_hir_ty(hir_ty).with_loc_if_ref(gcx, DataLocation::Storage);
```
stevencartavia: False positive when a `using L for T` directive only attaches through **internal** library functions whose bound first parameter is `calldata`. The type is normalized to `storage` before `members_of`, but `storage → calldata` isn't an implicit conversion (`storage → memory` is), so these calldata-first-param functions are missed and the directive gets flagged even though it's live.

Repro (compiles fine, `b.slice1()` resolves, but lint flags `using BytesLib for bytes`):

```solidity
library BytesLib {
    function slice1(bytes calldata b) internal pure returns (bytes calldata) {
        return b[1:];
    }
}

using BytesLib for bytes;

contract D {
    function f(bytes calldata b) external pure returns (bytes calldata) {
        return b.slice1();
    }
}
```

Removing the `using` line fails with `Member "slice1" not found or not visible after argument-dependent lookup in bytes calldata`, confirming the directive is live. Same happens for `struct T calldata`, `uint256[] calldata`, and `string calldata`. A `memory`-param version is correctly not flagged.

Fix direction: for reference types, check attachment under the relevant data locations (`storage`, `memory`, `calldata`) and only flag when none attach. Value types are unaffected by this data-location issue. Would also want a calldata case added to the testdata.

crates/lint/src/sol/info/incorrect_using_for.rs
```diff
@@ -0,0 +1,91 @@
...
+        ctx: &LintContext,
+        gcx: Gcx<'hir>,
+        hir: &'hir Hir<'hir>,
+        directive: &'hir UsingDirective<'hir>,
+    ) {
+        // `using L for *` attaches every function of the library: nothing to validate.
+        let Some(hir_ty) = &directive.ty else { return };
+        // `members_of` expects reference types wrapped in their data location. Storage
+        // converts implicitly to memory, so the storage form covers library functions
+        // taking either location.
```
stevencartavia: The source comment is incomplete/misleading:

```rust
// `members_of` expects reference types wrapped in their data location. Storage
// converts implicitly to memory, so the storage form covers library functions
// taking either location.
```

That explains storage/memory, but not calldata. The docs (`incorrect-using-for.md`) similarly only mention "a storage reference to a memory one" and omit calldata receiver attachments.

crates/lint/src/sol/info/incorrect_using_for.rs
```diff
@@ -0,0 +1,91 @@
...
+            for member in gcx.members_of(ty, directive.source, directive.contract) {
+                // A member counts when it is an attached function declared in the library.
+                if member.attached
+                    && let TyKind::Fn(function_ty) = member.ty.kind
+                    && let Some(function_id) = function_ty.function_id
+                    && hir.function(function_id).contract == Some(library_id)
+                {
+                    attaches = true;
+                }
+            }
```
stevencartavia: False negative: a no-op `using L for T` can be hidden by a braced directive for the same library/type that references a **private** function. The check asks "is any member from `L` attached on `T`?" instead of "does this `Library(L)` entry attach anything?" Solar's library-form `using L for T` skips private functions, but braced `using {L.g} for T` inside `L` is allowed to attach private `g`, and that attached member shows up in `members_of(T)` — so the dead library-form directive looks useful.

Repro (compiles; lint flags nothing):

```solidity
library SLib {
    using {SLib.g} for St;   // braced; private g
    using SLib for St;       // no-op: only the private g takes St — should flag, doesn't

    function g(St memory s) private pure returns (uint256) { return s.x; }

    function use() internal pure returns (uint256) {
        St memory s = St(1);
        return s.g();
    }
}

struct St { uint256 x; }
```

Removing `using SLib for St;` still compiles, confirming it's a no-op. Dropping the braced directive instead makes the lint correctly flag `using SLib for St`, isolating the cause. (Note the repro must have no non-private `SLib` function taking `St`, or the library-form directive really does attach that one.)

crates/lint/src/sol/info/incorrect_using_for.rs
0xMars42: Fixed in 402fe7667. Attachment is now probed under storage, memory and calldata separately, and the entry only flags when none of them attaches, since storage does not convert to calldata. Your `bytes calldata` repro plus a `string calldata` variant are in the testdata and stay clean.

crates/lint/src/sol/info/incorrect_using_for.rs
0xMars42: Both updated in 402fe7667: the source comment now explains the three probed locations and the docs page covers the calldata attachment and the private-function rule.

crates/lint/src/sol/info/incorrect_using_for.rs
0xMars42: Fixed in 402fe7667. The entry's liveness test now excludes private functions: the library form skips them, so a private member attached by a braced directive in scope no longer makes a dead `using L for T` look useful. Your repro is in the testdata and the library-form directive flags.

crates/lint/testdata/IncorrectUsingFor.sol
```diff
@@ -0,0 +1,132 @@
...
+    }
+}
+
+library StringLib {
+    function shout(string memory s) internal pure returns (string memory) {
+        return s;
+    }
+}
+
+library WideLib {
```
mablr, line 28: Please add a clean fixture case for `public`/`external` library functions. Library-form `using L for T` attaches all non-private applicable functions, not only `internal`, and the current fixture does not lock that in.

crates/lint/testdata/IncorrectUsingFor.sol
```diff
@@ -0,0 +1,132 @@
...
+// The library form skips private functions, but a braced directive is allowed to attach one
+// inside its own library. That attached private member must not make the dead library-form
+// directive look useful: `using PrivateLib for Holder` attaches nothing of its own.
+struct Holder {
+    uint256 x;
+}
+
+library PrivateLib {
+    using {PrivateLib.grab} for Holder;
+    using PrivateLib for Holder; //~NOTE: `using ... for` names a library
```
mablr, line 122: Please add a negative fixture for a library whose only ordinary function has zero parameters. Library-form `using L for T` skips those, so this should produce the lint just like the private-only case.

crates/lint/testdata/IncorrectUsingFor.sol
0xMars42: Added in b7c3d07d0: a `VisibleLib` with one `public` and one `external` function stays clean, locking in that the library form attaches all non-private applicable functions.

crates/lint/testdata/IncorrectUsingFor.sol
0xMars42: Added in b7c3d07d0: a `ZeroParamLib` whose only ordinary function takes no parameter reports, with no code change needed since a zero-parameter function never shows up as an attached member.

crates/lint/docs/incorrect-using-for.md
```diff
@@ -0,0 +1,45 @@
+# Incorrect using-for
+
+**Severity**: `Info`
+**ID**: `incorrect-using-for`
+
+Flags `using L for T` directives whose library has no function applicable to the type: they attach nothing.
+
+## What it does
+
+Reports a `using ... for` directive, file-level or contract-level, when no function of the named library accepts the target type as its bound first parameter. Attachment follows the type checker, implicit conversions included: a `uint8` binds to a `uint256` parameter and a derived contract to a base parameter. Reference types are probed under every data location their bound value may live in (`storage`, `memory` and `calldata`), so a library whose only applicable function takes a `calldata` first parameter counts as attaching. Private library functions never count: the library form skips them, even when a braced directive in scope attaches one. `using L for *` and the braced form `using {f} for T` are out of scope, the latter because the compiler already rejects a function that cannot attach. This mirrors Slither's `incorrect-using-for` detector.
```
mablr: This should not say the lint “mirrors Slither” as-is. The implementation is intentionally more semantic/different in scope in a few places (contract-level directives, Solar member resolution/data locations, and private-function exclusion). Prefer “inspired by Slither’s detector” or drop the comparison.

crates/lint/testdata/IncorrectUsingFor.sol
```diff
@@ -0,0 +1,164 @@
...
+        return v * 2;
+    }
+
+    function extTriple(uint256 v) external pure returns (uint256) {
+        return v * 3;
+    }
+}
+
+contract UsesVisibleAttachments {
+    using VisibleLib for uint256;
```
mablr: This fixture covers non-internal functions, but not external specifically: the public function is enough to keep the directive live, so the test would still pass if external functions stopped counting. Please split public-only and external-only cases, or add an external-only `using ExternalOnlyLib for uint256` case.

crates/lint/docs/incorrect-using-for.md
0xMars42: Reworded in 008ffaf6e to "inspired by Slither's detector", with the scope differences stated: contract-level directives, Solar member resolution across data locations, and the private-function exclusion.

crates/lint/testdata/IncorrectUsingFor.sol
0xMars42: Split in 008ffaf6e into a public-only and an external-only library, each proving its visibility on its own. Verified the external case bites by temporarily excluding external functions from the attachment test: the suite then fails on `using ExternalOnlyLib for uint256`.

mablr (approved):

mattsse (approved):

PR #15582, opened 2026-07-04
feat(lint): add cyclomatic-complexity detector

0xMars42:
Adds the `cyclomatic-complexity` lint from #14381 (Info).

Reports functions whose cyclomatic complexity is strictly above 11, the threshold Slither's detector of the same name uses. The complexity is computed as one plus the number of decision points, the structured-program equivalent of `E - N + 2P` on the control-flow graph: each `if` counts one, loop conditions included, since solar desugars every `for`, `while` and `do while` into a loop branching on its condition, so a condition-less `for (;;)` correctly adds nothing; each ternary, each `catch` clause and each additional assembly `switch` case count one as well. Boolean `&&` / `||` operators add nothing, matching the graph Slither computes on.

The fixture pins the threshold from both sides: ten decision points stay clean, eleven report, a mixed-form function sums loops, ternaries and catch clauses to the same count, and a short-circuit-heavy function stays clean. `TautologicalCompare.sol` had a fixture function above the threshold; it is split in two, with every existing annotation preserved, to keep fixtures mono-lint.

Checked against real OpenZeppelin sources (`Strings`, `AccessControl`, `SafeERC20`, `Math`): no findings, consistent with well-factored code.


crates/lint/src/sol/info/cyclomatic_complexity.rs
```diff
@@ -0,0 +1,89 @@
...
+        match &stmt.kind {
+            // One decision per `if`, the loop conditions included (see above).
+            StmtKind::If(..) => self.decisions += 1,
+            // The first clause is the `returns` one; each `catch` clause is a branch.
+            StmtKind::Try(stmt_try) => {
+                self.decisions += stmt_try.clauses.len().saturating_sub(1);
+            }
+            // Each case of a Yul switch beyond the first is a branch.
+            StmtKind::Switch(switch) => {
+                self.decisions += switch.cases.len().saturating_sub(1);
```
stevencartavia: The Yul `switch` handling counts `switch.cases.len().saturating_sub(1)`, but `cases` includes the `default` clause (an entry with `constant == None`). The formula subtracts one entry regardless, so a switch without a `default` is undercounted by one real `case`, and adding an empty `default {}` bumps the count back up — the score depends on whether a no-op default is present.

Reproduced with a body that is only a Yul switch of 11 real `case` arms: no `default` → not flagged (counted as 10; complexity 11); the identical switch + empty `default {}` → flagged (counted as 11; complexity 12). 11 real branches should be complexity 12 either way. Counting non-default cases fixes it and matches Slither (one `if` per non-default case):

```rust
StmtKind::Switch(switch) => {
    self.decisions += switch.cases.iter().filter(|c| c.constant.is_some()).count();
}
```

crates/lint/src/sol/info/cyclomatic_complexity.rs
```diff
@@ -0,0 +1,89 @@
...
+    ) {
+        let _ = gcx;
+        if let Some(body) = &func.body {
+            let mut counter = DecisionCounter { hir, decisions: 0 };
+            for stmt in body.stmts {
+                let _ = counter.visit_stmt(stmt);
+            }
+            // For a structured program the complexity is one plus the decision points.
+            if counter.decisions + 1 > MAX_COMPLEXITY {
+                ctx.emit(&CYCLOMATIC_COMPLEXITY, func.keyword_span());
```
stevencartavia: `check_function` also fires on modifier definitions (modifiers are function-like in the HIR), so complex modifiers get reported. Reproduced: an identical 12-`if` body flags in both a `modifier` and an `internal function`. Slither's detector (referenced for the threshold) iterates only `functions_declared` + `functions_top_level` and never `modifiers_declared`, so it never reports modifiers. The diagnostic also reads "this function has a cyclomatic complexity…" while pointing at a `modifier`.

Either skip `FunctionKind::Modifier` for parity, or intentionally keep modifiers and document it + reword the message + add a modifier fixture.

crates/lint/src/sol/info/cyclomatic_complexity.rs
```diff
@@ -0,0 +1,89 @@
...
+        &mut self,
+        ctx: &LintContext,
+        gcx: Gcx<'hir>,
+        hir: &'hir Hir<'hir>,
+        func: &'hir hir::Function<'hir>,
+    ) {
+        let _ = gcx;
+        if let Some(body) = &func.body {
+            let mut counter = DecisionCounter { hir, decisions: 0 };
+            for stmt in body.stmts {
```
stevencartavia: `check_function` counts only `body.stmts`, so decision points in modifier-invocation arguments and base-constructor call arguments are never visited.

Reproduced (both stay clean at 10 body ifs + 1 ternary in the call args, when they should be complexity 12): `function f(uint256 x) m(x == 0 ? 1 : 2) { /* 10 ifs */ }` → not flagged; moving the same ternary into the body → flagged. `constructor(uint256 x) Base(x == 0 ? 1 : 2) { /* 10 ifs */ }` → not flagged.

Driving the counter with the full function instead of hand-iterating the body picks up `func.modifiers` too:

```rust
if func.body.is_some() {
    let mut counter = DecisionCounter { hir, decisions: 0 };
    let _ = counter.visit_function(func);
    if counter.decisions + 1 > MAX_COMPLEXITY {
        ctx.emit(&CYCLOMATIC_COMPLEXITY, func.keyword_span());
    }
}
```

crates/lint/src/sol/info/cyclomatic_complexity.rs
stevencartavia: Yul helper functions declared inside `assembly {}` are treated as independent HIR functions, so `check_function` fires on them and this lint can flag them independently.

This has one concrete issue and one scope question: the diagnostic anchor is wrong, and Yul helpers may not be the intended scope. `func.keyword_span()` assumes the span starts at a Solidity `function` keyword, but a Yul function's HIR span starts at the helper name, so the underline starts at the name and extends 8 chars.

Reproduced with a 12-`if` Yul helper: JSON primary span is `column_start=22, column_end=30`, underlining `helper(v` on `function helper(v) -> o {` rather than the `function` keyword (cols 13–20). A plain Solidity function correctly anchors at cols 5–13.

Handle `func.is_yul` explicitly — either skip Yul helpers if the lint is meant to cover only Solidity declarations, or emit on the Yul helper name span instead.

crates/lint/src/sol/info/cyclomatic_complexity.rs
0xMars42: Fixed in adc3e74be. The counter now takes the non-default cases (`constant.is_some()`), one branch each, so the score no longer depends on a no-op `default {}`. The testdata pins the differential pair: 11 real cases without a default flag, 10 real cases plus a default stay at the threshold.

crates/lint/src/sol/info/cyclomatic_complexity.rs
0xMars42: Fixed in adc3e74be, skipping `FunctionKind::Modifier` for parity with Slither's iteration set. A 12-if modifier stays clean in the testdata.

crates/lint/src/sol/info/cyclomatic_complexity.rs
0xMars42: Fixed in adc3e74be with your suggestion: the counter visits the whole function, which picks up modifier-invocation and base-constructor call arguments. Both repros are in the testdata (10 body ifs plus a ternary in the modifier argument, and the same in a base constructor call) and flag at complexity 12.

crates/lint/src/sol/info/cyclomatic_complexity.rs
0xMars42: Fixed in adc3e74be, skipping `func.is_yul`: Yul helpers are independent HIR functions but not Solidity declarations, and that also removes the wrong name-anchored span. One note from the testdata: the helper's own decisions are not attributed to the enclosing function either (its body lives in the helper's HIR function), so a 12-if helper inside an otherwise simple function reports nothing. That felt acceptable for a Solidity-declaration scope, but it deviates from a strict per-node count and is easy to revisit if you prefer.

crates/lint/src/sol/info/cyclomatic_complexity.rs
```diff
@@ -0,0 +1,96 @@
...
+        ctx: &LintContext,
+        gcx: Gcx<'hir>,
+        hir: &'hir Hir<'hir>,
+        func: &'hir hir::Function<'hir>,
+    ) {
+        let _ = gcx;
+        // Slither's detector iterates only declared and top-level functions: modifier
+        // definitions are never reported, and Yul helpers declared inside `assembly {}`
+        // blocks are independent HIR functions but not Solidity declarations.
+        if matches!(func.kind, hir::FunctionKind::Modifier) || func.is_yul {
```
mablr: This creates a false negative versus Slither: Slither does report complex Yul helper functions (the existing fixture's `ComplexityYulHelper.throughHelper.asm_0.helper` is reported with complexity 13), but this skips every `func.is_yul`. Please count/report Yul functions too, or document and test that this lint intentionally diverges from Slither.

0xMars42: Good catch, the Slither run settles it. Fixed in f3c27cbfb: Yul helpers are checked like any function and the diagnostic anchors on the helper name, since their span starts there rather than at a keyword (which also settles the span issue reported above). The testdata helper now reports at complexity 13, and its decisions still do not count toward the enclosing function.

mablr (dismissed):

mattsse (dismissed):

mablr (approved):
Resolved conflicts, G2G

emmajam (approved):

PR #15583, opened 2026-07-04
perf(symbolic): try simple uint shrinks first

figtracer:
Tries `1` and binary lowering before clearing individual set bits when minimizing unsigned symbolic counterexample arguments. The existing bit-clearing fallback stays in place for non-monotonic cases.

Bug suite (`grandizzy/symbolic-bug-suite`, 22/22 expected failures):
- minimization attempts: 817 -> 550
- solver counters unchanged: 212 paths, 283 solver queries, 100 SMT queries
- single-thread full suite: 1.097s +/- 0.003s -> 1.086s +/- 0.002s
- DEI+Rari focused: 654.2ms +/- 5.9ms -> 642.4ms +/- 4.6ms

`foundry-bench` (`forge_symbolic_test`, `pr-bench.sh`, 5 runs):
- solady: 0.359s +/- 0.013s -> 0.358s +/- 0.011s
- SorellaLabs-angstrom: 0.494s +/- 0.009s -> 0.487s +/- 0.004s
- farcasterxyz-contracts: 0.253s +/- 0.005s -> 0.262s +/- 0.006s; targeted rerun 0.252s +/- 0.004s -> 0.254s +/- 0.004s
- symbolic pass/incomplete counts and SMT query counts unchanged

### DIFF
--- a/crates/forge/src/symbolic_minimizer.rs
+++ b/crates/forge/src/symbolic_minimizer.rs
@@ -792,10 +792,6 @@ fn minimize_uint(
         return true;
     }
 
-    if minimize_uint_by_search(value, current, bits, try_value) {
-        return true;
-    }
-
     let bit_limit = bits.min(256);
     for bit in (0..bit_limit).rev() {
         let mask = U256::from(1) << bit;
@@ -808,7 +804,7 @@ fn minimize_uint(
         }
     }
 
-    false
+    minimize_uint_by_search(value, current, bits, try_value)
 }
 
 fn minimize_uint_by_search(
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
I think this needs one adjustment before merging.

The changed order at https://github.com/foundry-rs/foundry/blob/6d94eb0251b140e53b8fd8a7e9a3d47e2eaaba97/crates/forge/src/symbolic_minimizer.rs#L795-L808 lets `minimize_uint_by_search` accept an intermediate value before the bit-clearing candidates are tried. For non-monotonic replay predicates, that can leave the minimizer stuck at a larger counterexample than before. For example, starting from `100`, a replay predicate that still fails only for `{100, 50, 36}` used to reach `36` through the high-bit clear candidate before binary search. With this order it accepts the search midpoint `50` first, then subsequent passes cannot reach `36` from `50`.

A low-risk fix would be to keep the cheap `0`/`1` fast paths if desired, but preserve the previous ordering between bit-clearing and search: run the bit-clearing loop before `minimize_uint_by_search`, and only call search as the fallback. I’d also add a unit test with a non-monotonic accepted set like `{100, 50, 36}` so this does not regress again.

mattsse (approved):
lgtm

PR #15586, opened 2026-07-05
feat(lint): add internal-function-used-once detector

0xMars42:
Adds the `internal-function-used-once` lint from #14381 (Info).

Reports an ordinary internal function, free functions included, that exactly one expression in the unit references, call or value alike. References resolve through the type checker (`gcx.type_of_expr`, `TyKind::Fn.function_id`), so overload selection, the qualified and `using for` forms and import aliases attribute each reference to the right declaration. References are counted across the whole unit, dependencies included, while only functions declared in the project's own sources report, mirroring the unused-error pass.

Exemptions on top of Aderyn's underscore rule: `virtual` functions and overrides, which exist for dynamic dispatch and cannot be inlined, and zero-reference functions, dead code being a different concern. The fixture pins each exemption separately, the function-pointer-value counting, recursion (the self-reference makes two), and a free function used from another contract.

Three existing fixtures had a helper referenced exactly once; each gains a second reference to keep fixtures mono-lint.


crates/lint/src/sol/info/internal_function_used_once.rs
```diff
@@ -0,0 +1,117 @@
...
+                || function.override_
+            {
+                continue;
+            }
+            let Some(name) = function.name else { continue };
+            if name.as_str().starts_with('_') {
+                continue;
+            }
+            // Exactly one reference: zero references is dead code, a different concern.
+            if counts.get(&function_id).copied().unwrap_or(0) == 1 {
```
stevencartavia: A self-recursive internal function with no non-recursive caller gets flagged, because the recursive self-call is counted as its one reference. The "inline into its caller" advice doesn't apply — there's no caller to inline into, and inlining the self-call would just reproduce the recursion.

```solidity
contract Rec {
    // flagged "used only once", but has no non-recursive caller
    function factorial(uint256 n) internal pure returns (uint256) {
        if (n <= 1) return 1;
        return n * factorial(n - 1);
    }
}
```

Mutual recursion (`ping`↔`pong`, no non-recursive caller) flags both too: each function's only counted reference is the other in the cycle. The `recurse` fixture only covers the "self-call + one external caller = 2" case, so this slips through.

crates/lint/src/sol/info/internal_function_used_once.rs
```diff
@@ -0,0 +1,117 @@
...
+impl<'gcx> hir::Visit<'gcx> for ReferenceCounter<'gcx> {
+    type BreakValue = Never;
+
+    fn hir(&self) -> &'gcx hir::Hir<'gcx> {
+        self.hir
+    }
+
+    fn visit_expr(&mut self, expr: &'gcx hir::Expr<'gcx>) -> ControlFlow<Self::BreakValue> {
+        // A name or member expression typed as a function is one resolved reference.
+        if matches!(expr.kind, hir::ExprKind::Ident(..) | hir::ExprKind::Member(..))
```
stevencartavia: Uses through user-defined operators aren't counted (operator expressions aren't `Ident`/`Member`), so a function used once directly and once via its operator is flagged as used-once.

```solidity
type T is uint256;
using {addC as +} for T global;

function addC(T a, T b) pure returns (T) {
    return T.wrap(T.unwrap(a) + T.unwrap(b));
}

contract C {
    // addC is used twice, but reported "used only once":
    function d(T a, T b) internal pure returns (T) { return addC(a, b); } // direct
    function o(T a, T b) external pure returns (T) { return a + b; }      // operator
}
```

mablr: This misses user-defined operator references. I verified by adding a fixture with `using {operatorAdd as +} for U global;` and a single `return a + b;`. Since operator uses do not appear as `Ident`/`Member` function expressions, this can false-negative when the operator is the only use, and false-positive if the function also has one direct call. Please either count resolved operator implementations or conservatively exclude operator implementations, and add fixture coverage.

0xMars42: Fixed in 714b412b1 with the conservative option: functions named by an operator entry of any file-level or contract-level `using` directive are out of scope entirely, collected from the HIR's already-resolved entries. Counting operator uses instead would mean re-resolving user operators by hand, and the suggestion would be wrong for them anyway, since the binding requires a named function. The testdata covers both of your cases: operator-only and operator plus one direct call, the latter verified to report when the exclusion is temporarily disabled.

crates/lint/src/sol/info/internal_function_used_once.rs
```diff
@@ -0,0 +1,148 @@
...
+                || function.override_
+            {
+                continue;
+            }
+            let Some(name) = function.name else { continue };
+            if name.as_str().starts_with('_') {
+                continue;
+            }
+            // Exactly one reference: zero references is dead code, a different concern.
+            if counts.get(&function_id).copied().unwrap_or(0) == 1 {
```
mablr: This still false-positives on recursive functions whose only references are within the recursion cycle. A self-recursive helper with no external caller has one counted reference (the self-call), and mutually-recursive helpers have one counted reference each, so the lint suggests inlining even though there is no non-recursive caller to inline into.

0xMars42: Fixed in ef381ee6d: the counter now tracks the referencing function. A self-reference marks the function recursive instead of counting, which rules it out entirely since a recursive function cannot be inlined anywhere, and a single reference is followed through the chain of single-reference sources, so a cycle with no external caller does not report. Both of your cases, plus a recursive function with one external caller, are in the testdata as clean.

crates/lint/src/sol/info/internal_function_used_once.rs
```diff
@@ -0,0 +1,213 @@
...
+    refs: &HashMap<hir::FunctionId, RefInfo>,
+    start: hir::FunctionId,
+) -> bool {
+    let mut visited = vec![start];
+    let Some(mut current) = refs.get(&start).and_then(|info| info.first_from) else {
+        return false;
+    };
+    // Each hop follows the unique referencing function; a repeat closes the cycle.
+    loop {
+        if visited.contains(&current) {
```
mablr, line 164: This cycle check also suppresses non-recursive helpers that are called by a function which happens to be in a separate single-reference cycle. For example, `tail()` below is used exactly once from `cycleA()` and should be reported/inlineable, but the walk goes `tail <- cycleA <- cycleB <- cycleA` and returns `true` when it sees the `cycleA`/`cycleB` cycle, even though `tail` is not part of it:

```solidity
contract TailOffCycle {
    function tail(uint256 v) internal pure returns (uint256) { return v + 1; }
    function cycleA(uint256 v) internal pure returns (uint256) {
        return v == 0 ? tail(v) : cycleB(v - 1);
    }
    function cycleB(uint256 v) internal pure returns (uint256) { return cycleA(v); }
}
```
The check should only suppress when the cycle includes `start` (or otherwise prove the target itself has no non-recursive caller), not any later cycle in the reference chain.

0xMars42: Fixed in 03e724ffd: the chain is linear, so the cycle contains the candidate exactly when the loop closes on the candidate itself, and the walk now only suppresses in that case. Your tail-off-cycle repro is in the testdata and reports, while the cycle members themselves stay suppressed.

mattsse (dismissed):
Found one blocker:

- `crates/lint/src/sol/info/internal_function_used_once.rs:196`: the reference counter treats every `Ident`/`Member` typed as a function as an inlineable use, including value-position references. A helper assigned to a function pointer, returned, or passed as a callback can be reported even though there is no direct caller to inline into. Please track whether the single reference is a direct call callee, or conservatively exempt value-position function references unless escape/use can be ruled out.

0xMars42:
Fixed in `a8c4658b3`. A reference counts as an inlinable use only when it is the callee of a direct call. A single value-position reference (a function pointer assigned, returned, or passed as a callback) has no call site to inline into and no longer reports. Updated the pointer fixture to stay clean and added the returned and callback positions, alongside a function called once directly that still reports.

mablr (dismissed):
LGTM

mattsse (dismissed):
The prior value-position reference blocker is fixed: function references used as values are no longer treated as direct inlineable calls. No new actionable findings in this pass.

mablr (approved):
Fixed merge conflict, G2G

mattsse (approved):

PR #15587, opened 2026-07-05
feat(lint): add modifier-used-only-once detector

0xMars42:
Adds the `modifier-used-only-once` lint from #14381 (Info).

Reports a modifier that exactly one function invokes, constructors included. Invocations are taken from the resolved modifier lists, so base-constructor calls sitting in the same syntactic position are never confused with modifier calls; they are counted across the whole unit, dependencies included, while only modifiers declared in the project's own sources report, mirroring the unused-error pass.

Exemptions on top of Aderyn's counting: `virtual` modifiers and overrides, which exist for dynamic dispatch and cannot be inlined, and zero-invocation modifiers, dead code being a different concern. The fixture pins the constructor invocation, the base-constructor coexistence and each exemption.


stevencartavia (approved):

grandizzy (approved):
lgtm!

PR #15588, opened 2026-07-05
feat(lint): add literal-instead-of-constant detector

0xMars42:
Adds the `literal-instead-of-constant` lint from #14381 (Info).

Reports every occurrence of a number, address or hex string literal whose value appears more than once in the executable bodies a contract declares: functions, constructors, modifiers, `receive` and `fallback`. Values compare semantically through solar's evaluated literal kinds, so `100`, `0x64` and `1e2` are one value and `1 ether` equals `1e18`; Aderyn's detector of the same name compares the source spellings instead. Matching Aderyn otherwise: `0`, `1` and `2` stay out, so do bare literals indexing an array, string and bool literals and single occurrences, and grouping is per contract. Emissions are sorted by position, since the grouping map iterates in arbitrary order.

Six legacy fixtures exercise repeated numeric patterns as their test subject (arithmetic shapes, shift amounts, digit counts); they now pin their own lints with `--only-lint` compile flags listing every lint their snapshot contains, keeping their existing diagnostics byte for byte.

Measured on OpenZeppelin v4.9.6 passed as direct input: `Math.sol` reports 61 times, its shift cascades repeating the same amounts by design, and `Strings.sol` twice; as dependencies neither reports, only a project's own sources do. If the repetition rule reads too eager on math-heavy code, excluding shift amounts the way index positions are excluded is a small follow-up.


crates/lint/src/sol/info/literal_instead_of_constant.rs
```diff
@@ -0,0 +1,113 @@
...
+                    && !matches!(index.peel_parens().kind, ExprKind::Lit(..))
+                {
+                    let _ = self.visit_expr(index);
+                }
+                return ControlFlow::Continue(());
+            }
+            ExprKind::Lit(lit) => {
+                let key = match &lit.kind {
+                    // `0`, `1` and `2` are structural rather than configuration values.
+                    LitKind::Number(v) if *v > U256::from(2u64) => Some(LiteralValue::Number(*v)),
```
mablr: Negative numbers are grouped with their positive literal value because the collector records only `LitKind::Number(v)` and ignores the surrounding unary minus. A temporary fixture with `return x + 5 + (-5);` reports both `5` and `-5` as repeated, even though those are different constants.

crates/lint/src/sol/info/literal_instead_of_constant.rs
```diff
@@ -0,0 +1,150 @@
...
+                            };
+                            self.groups.entry(key).or_default().push(expr.span);
+                        }
+                        return ControlFlow::Continue(());
+                    }
+                    // Nested unary (`-(-5)`, `~~5`) folds to a value that is neither this
+                    // operator's nor the bare literal's; canonicalizing it is
+                    // not worth it, so skip the whole chain rather than risk
+                    // grouping it with an unrelated constant.
+                    ExprKind::Unary(inner, _)
```
grandizzy: This early-returns on the *whole* subtree whenever the operand is another `-`/`~` unary, which is correct for `-(-5)` / `~~5`, but it also drops unrelated literals inside non-literal operands:

```solidity
return -(-(x + 500));    // 500 never recorded
return ~~(flags & 0xff); // 0xff never recorded
```

Suggest only taking the early return when the inner unary's operand is itself a literal, and otherwise falling through to `walk_expr`:

```rust
ExprKind::Unary(inner, inner_operand)
    if matches!(inner.kind, UnOpKind::Neg | UnOpKind::BitNot)
        && matches!(inner_operand.peel_parens().kind, ExprKind::Lit(..)) =>
{
    return ControlFlow::Continue(());
}
```

crates/lint/src/sol/info/literal_instead_of_constant.rs
```diff
@@ -0,0 +1,150 @@
...
+    type BreakValue = Infallible;
+
+    fn hir(&self) -> &'hir Hir<'hir> {
+        self.hir
+    }
+
+    fn visit_expr(&mut self, expr: &'hir Expr<'hir>) -> ControlFlow<Self::BreakValue> {
+        match &expr.kind {
+            // A bare literal used as an index (`arr[3]`) is positional, not a magic value.
+            ExprKind::Index(base, index) => {
```
grandizzy, line 140: This skips *any* bare literal index, but the comment/scope says "array index." Only syntactic info is available here, so mapping keys — which are genuine magic values — are also silently excluded:

```solidity
m[500] = 1; m[500] = 2;                 // not flagged
balances[0x1111...]; balances[0x1111...]; // not flagged
```

Either make this type-aware (only suppress when the base type is array/slice-like) or update the docs and PR description to say "bare literals in any index expression are ignored," since mappings are currently missed.

grandizzy (dismissed):
A few visitor-scope issues below (all false-negatives / scope inconsistencies rather than wrong values). Note: @mablr's earlier sign-grouping comment is already resolved by the `NegNumber`/`BitNotNumber` fix — finding 1 is the remaining edge in that same area.

crates/lint/src/sol/info/literal_instead_of_constant.rs
stevencartavia: Related: `ExprKind::Slice` isn't special-cased, so this has the complementary issue — bare literals in slice bounds fall through and get flagged as magic values, even though the same positional rationale seems to cover them. `d[555:600]` twice reports both `555` and `600`. Whatever type-aware fix lands here should cover `Slice(base, start, end)` too.

crates/lint/src/sol/info/literal_instead_of_constant.rs
```diff
@@ -0,0 +1,150 @@
...
+}
+
+impl<'hir> Visit<'hir> for LiteralCollector<'hir> {
+    type BreakValue = Infallible;
+
+    fn hir(&self) -> &'hir Hir<'hir> {
+        self.hir
+    }
+
+    fn visit_expr(&mut self, expr: &'hir Expr<'hir>) -> ControlFlow<Self::BreakValue> {
```
stevencartavia, line 136: Yul `switch` case literal labels aren't collected. The collector only overrides `visit_expr` and relies on the default HIR visitor, whose `StmtKind::Switch` arm walks the selector and case bodies but not `case.constant` — so a literal in `case 500 {}` is invisible while `y := 500` in the body is recorded. `case 500 { ... } ... y := 500` reports nothing (control: `let a := 600` + `y := 600` reports both, so ordinary Yul expression literals are handled). Fix: override `visit_stmt` for `StmtKind::Switch`, factor the `ExprKind::Lit` recording into a `record_lit(&Lit)` helper (note `case.constant` is a `Lit`, not an `Expr`), call it for each `case.constant`, then continue with `walk_stmt`; or document Yul switch labels as out of scope.

crates/lint/src/sol/info/literal_instead_of_constant.rs
mablr, line 140: This skips bare literal indices for every `ExprKind::Index`, including mapping lookups. The array-position rationale makes sense for `arr[3]`, but a repeated mapping key like `limits[500] + limits[500]` is still a repeated configuration literal; adding that annotated case to `LiteralInsteadOfConstant.sol` makes the UI test fail with no diagnostic. Please only exempt positional array/bytes indices, or otherwise keep walking mapping keys.

crates/lint/src/sol/info/literal_instead_of_constant.rs
0xMars42: Fixed in 2aef8a24f with your suggestion: the skip now requires the inner operand to be a literal itself, and anything else falls through to the walk. Both repros are in the testdata: `-(-(x + 500))` groups its `500` with a bare one, and the two `0xff` in `~~(f & 0xff) + 0xff` report.

crates/lint/src/sol/info/literal_instead_of_constant.rs
0xMars42: Went with the type-aware option in 2aef8a24f: the base type decides, so mapping keys are collected (both your `m[500]` and address-key repros are in the testdata and report) while array-like indexing stays positional. Docs updated accordingly.

0xMars42: Fixed in 2aef8a24f: `Slice` bounds are excluded with the same positional rationale as indexes, so `d[555:600]` twice stays clean. Covered in the testdata.

crates/lint/src/sol/info/literal_instead_of_constant.rs
0xMars42: Fixed in 2aef8a24f along the lines you suggested: the recording moved into a `record_lit` helper and a `visit_stmt` override records each `case.constant` before walking the statement. The testdata pins a Yul `case 500` grouping with a Solidity `500` in the same function.

crates/lint/src/sol/info/literal_instead_of_constant.rs
0xMars42: Fixed in 2aef8a24f, together with grandizzy's report of the same issue above: the base type now decides, so mapping keys are walked and collected while array/bytes indexing stays positional. The testdata covers a repeated `m[500]` key (equivalent to your `limits[500]` repro) and a repeated address key, both reporting.

mablr (dismissed):
LGTM

mablr (dismissed):

crates/lint/src/sol/info/literal_instead_of_constant.rs
```diff
@@ -0,0 +1,195 @@
...
+                            && matches!(inner_operand.peel_parens().kind, ExprKind::Lit(..)) =>
+                    {
+                        return ControlFlow::Continue(());
+                    }
+                    // Any other operand (`-x`, `-(a + 5)`): walk normally so a literal inside it is
+                    // still recorded with its own value.
+                    _ => {}
+                }
+            }
+            ExprKind::Lit(lit) => self.record_lit(lit),
```
figtracer, line 218: The `Math.sol` result in the PR body seems noisy enough to handle before this lands: repeated shift amounts are usually structural, not named configuration values. Consider excluding bare shift operands, similar to array/slice positions, and adding a fixture for that case.

0xMars42: Fixed in e827823e2: a bare literal shift amount is excluded like array and slice positions, compound assignments included, which is what made the Math.sol measurement noisy. The shifted operand still counts, so a repeated shifted literal value reports, and a computed amount is walked so the literals inside it keep counting. All three cases are in the testdata.

mablr (dismissed):

mattsse (changes requested):
Found one blocker:

- `crates/lint/src/sol/info/literal_instead_of_constant.rs:40`: the collector only visits `body.stmts`, so repeated literals in executable header expressions are missed. For example, `constructor() Base(500) { x = 500; }` or `function f() capped(700) { x = 700; }` only counts the body literal and false-negatives. Please either visit the full HIR function / all relevant expression sites, or explicitly document and test these positions as out of scope.

0xMars42:
Fixed in `93f928f85`. Collection now visits each function whole instead of iterating its body statements, so a literal in a header (a modifier or base-constructor argument) groups with body occurrences. Added a fixture where a value split across `Base(500)` / `capped(700)` and a body now reports. Visiting the whole function also reaches fixed array sizes in parameter types; local sizes were already counted, so this makes the two consistent, covered by a fixture. Rebased on the master merge.

crates/lint/src/sol/info/literal_instead_of_constant.rs
```diff
@@ -0,0 +1,213 @@
...
+        hir: &'hir Hir<'hir>,
+        id: hir::ContractId,
+    ) {
+        // Group the literals of the contract's own functions and modifiers by semantic value,
+        // headers included: a repeated value in a modifier or base-constructor argument counts
+        // like one in a body. Inherited items group with their declaring contract.
+        let mut collector = LiteralCollector { gcx, hir, groups: HashMap::new() };
+        for item_id in hir.contract(id).items {
+            if let hir::ItemId::Function(function_id) = item_id {
+                let _ = collector.visit_function(hir.function(*function_id));
```
mablr: `visit_function` walks the whole function signature, so repeated fixed-array sizes in parameter types are reported even when the value never appears in an executable body. That seems outside this detector's documented scope ("executable bodies") and creates noise for overloads like `function a(uint256[365] calldata)` / `function b(uint256[365] calldata)`: adding that case to `LiteralInsteadOfConstant.sol` produces two new diagnostics on the type annotations. Can we restrict collection to the body plus the intended executable-ish headers (modifier/base-constructor arguments), or otherwise explicitly exclude type nodes such as array lengths?

mattsse (dismissed):
Thanks for the update. The previous blocker around literals in modifiers, base constructors, and header expressions is fixed, but the new `visit_function` call broadens collection too far.

At `crates/lint/src/sol/info/literal_instead_of_constant.rs:39`, walking the whole function also visits type nodes, so repeated fixed-array sizes in function signatures can be reported even though those literals are not executable body literals or the intended header-expression cases. Please keep the broader coverage scoped to function bodies plus the relevant header expressions, or otherwise avoid collecting literals from type nodes.

0xMars42:
Fixed in `e64ae31eb`. Collection now walks the body statements plus the modifier and base-constructor arguments of the header, and nothing else: parameter and return types are no longer visited, so a fixed array size in a signature is out of scope again. Header arguments still group with body occurrences.

The fixture pins both sides: repeated sizes across two signatures stay clean, a size shared between a signature and a body does not group with it, and a return type carrying one is ignored. Reverting to `visit_function` makes five diagnostics reappear on that contract, on the parameter types and on the return type. `crates/lint/docs/literal-instead-of-constant.md` now states the scope.

mablr (dismissed):
LGTM

mablr (dismissed):

crates/lint/src/sol/info/mod.rs
```diff
@@ -54,8 +54,13 @@ use event_fields::EVENT_FIELDS;
 mod unused_error;
 use unused_error::UNUSED_ERROR;
 
+<<<<<<< feat/lint-literal-instead-of-constant
```
stevencartavia: The latest merge committed unresolved conflict markers in `crates/lint/src/sol/info/mod.rs` and `crates/lint/README.md`, preventing `forge-lint` from compiling. Please resolve them while retaining both the `literal-instead-of-constant` and `function-init-state` registrations and documentation entries.

mablr (approved):

stevencartavia (approved):

PR #15591, opened 2026-07-05
fix(symbolic): explore stateful invariant paths

figtracer:
## Description
This improves symbolic invariant execution for stateful Morpho-style paths. It lets bounded symbolic invariant exploration continue through multi-call sequences, replays setup-time arbitrary-storage values for explicit symbolic-storage targets, and completes hard-arithmetic fallback models with the supporting storage values needed by checked ERC20-style arithmetic guards.

- The Morpho canary needs a `supply -> supplyCollateral -> borrow` sequence. On `GalloDaSballo/morpho-foundry-fv@example-concrete-fallback`, the clean harness pins the market/token/oracle addresses, but it does not provide a concrete target path that funds/approves tokens or sets an oracle price. Normal invariant fuzzing confirms that boundary: `invariant_hasBorrowed` passed after 256 runs with all 128000 target calls reverting.
- Plain `forge test --symbolic` on the same repro reaches the relevant shape but stops incomplete after 26 paths with `hard arithmetic heuristic witness used; no replayed counterexample found`.
- With the dependency boundary made explicit by marking `loanToken`, `collateralToken`, and `oracle` as arbitrary storage, this PR finds and replays the 3-call `supply -> supplyCollateral -> borrow` counterexample (`paths: 177`, `models: 1`, `hard-arith: 84` in a fresh run).

Current workaround: explicit dependency storage boundary:
- Use the existing Foundry `vm.setArbitraryStorage(address)` when a symbolic invariant depends on setup-deployed dependency contracts whose storage must be satisfiable, but no target call in the invariant campaign writes that storage. In the Morpho repro, that means the ERC20 token mocks and oracle mock:

```solidity
interface ArbitraryStorageVm {
    function setArbitraryStorage(address target) external;
}

ArbitraryStorageVm(address(vm)).setArbitraryStorage(address(loanToken));
ArbitraryStorageVm(address(vm)).setArbitraryStorage(address(collateralToken));
ArbitraryStorageVm(address(vm)).setArbitraryStorage(address(oracle));
```

- Keep this scoped to external dependencies that model the environment, such as token balances/allowances and oracle prices. Do not blanket-mark the invariant harness or protocol state as arbitrary; that can produce unreachable counterexamples that do not replay.
- This cheatcode annotation is the current workaround for dependency/environment storage. The desired UX is hybrid invariant exploration that learns reachable frontier states without requiring users to add these annotations.

What this intentionally does not do:
- It does not make all setup storage arbitrary. On the same repro, `--symbolic-storage-layout generic` produces a 2-path counterexample that does not replay because it also makes harness/protocol state arbitrary. That is too broad for this class of invariant test.
- It does not add another user-facing modeling cheatcode. The longer-term UX should be an Echidna-style hybrid worker: normal invariant fuzzing builds or loads reachable corpus/frontier states, symbolic execution extends those states, and confirmed counterexamples feed back into the corpus/replay flow.

## Status quo

- Halmos has the same explicit symbolic-storage boundary (`svm.enableSymbolicStorage` / Foundry `vm.setArbitraryStorage`), so it is not a better no-annotation UX for this case.
- Echidna symExec is the product direction to pursue: symbolic execution is an additional exploration worker over the same campaign/tests. A local Echidna run on this repo did not get to the property because HEVM rejected a Foundry deployment cheatcode in the harness, but the model is still the right UX shape.
- Certora Morpho specs also make the boundary explicit: they use a dedicated `MorphoHarness`, `require` assumptions, `extSloads => NONDET`, `_.price() => CONSTANT`, and envfree getters instead of asking the prover to infer arbitrary dependency storage.


crates/forge/src/runner.rs
```diff
@@ -3398,6 +3423,266 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                ffi_enabled: self.config.ffi,
+            }) {
+                SymbolicInvariantRunResult::Safe(stats) => {
+                    self.result.symbolic_result(
+                        TestStatus::Success,
+                        None,
+                        None,
+                        SymbolicResult::pass(&self.config.symbolic, stats),
+                    );
+                    return self.result;
```
mablr: `SymbolicInvariantRunResult::Safe` returns before the normal invariant campaign runs, but this symbolic pass is bounded by `symbolic.invariant_depth` and is not equivalent to fuzzing. I reproduced a fixture where `symbolic.invariant_depth = 1` reports `[PASS]` while the normal invariant campaign with depth 2 finds `first() -> second()` and fails.

crates/evm/symbolic/src/runtime/state.rs
```diff
@@ -146,6 +146,11 @@ impl PathState {
             .and_then(|cheats| cheats.gas_price)
             .unwrap_or_else(|| executor.tx_env().gas_price());
         self.gas_price = SymExpr::constant(cx, U256::from(gas_price));
+        if let Some(cheats) = executor.inspector().cheatcodes.as_ref() {
+            for target in cheats.arbitrary_storage_targets() {
+                self.world.enable_arbitrary_storage(target);
+            }
```
mablr: Should this also account for `vm.copyStorage` targets? Setup can mark from with `vm.setArbitraryStorage(from)` and then copy that storage to `to` concrete execution treats later reads from to as backed by the arbitrary source. Since this import only enables the explicit arbitrary-storage targets, copied targets may stay concrete in the symbolic world and leave some replayable invariant paths unexplored.

mablr (commented):
Looks good, just one question

crates/cheatcodes/src/inspector.rs
```diff
@@ -428,6 +428,28 @@ impl ArbitraryStorage {
...
+    fn copied_targets(&self) -> impl Iterator<Item = Address> + '_ {
+        self.copies.keys().copied()
+    }
+
+    /// Caches a concrete value for a slot on an arbitrary-storage address.
+    fn cache_value(&mut self, address: Address, slot: U256, data: U256) {
+        if let Some(values) = self.values.get_mut(&address) {
+            values.insert(slot, data);
+        }
+    }
```
mablr, line 465: It seems this cache path still ignores copied arbitrary-storage targets because values only contains explicit `vm.setArbitraryStorage` addresses. If symbolic replay records 0 for a copied target slot, `clone_executor_with_symbolic_storage` writes zero to the copied address and then this cache update no-ops; on the first cold SLOAD, `arbitrary_storage_end` takes the `is_arbitrary_storage_copy` branch and seeds a fresh random source value instead. The new copied-storage regression uses 42, so it does not cover that replay mismatch.

crates/forge/src/runner.rs
```diff
@@ -3398,6 +3423,260 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                targets: symbolic_targets,
+                senders: sender_filters.targeted,
+                depth: self.config.symbolic.invariant_depth as usize,
+                fail_on_revert: invariant_contract.invariant_fns[invariant_contract.anchor_idx].1,
+                ffi_enabled: self.config.ffi,
+            }) {
+                SymbolicInvariantRunResult::Safe(stats) => {
+                    self.result.record_symbolic(SymbolicResult::pass(&self.config.symbolic, stats));
+                }
+                SymbolicInvariantRunResult::Incomplete { kind, reason, stats } => {
```
stevencartavia, line 3967: With `--symbolic`, every non-optimization invariant now runs symbolic exploration before fuzzing. On `Safe` it records a pass and continues to fuzz, but on `Incomplete` it sets `TestStatus::Failure` and returns immediately, so the fuzz campaign never runs. Since realistic invariants usually exceed the path/depth/timeout budget, they return `Incomplete` and fail.

Repro: a trivially-true invariant (`assertGe(bank.total(), 0)` over deposit/withdraw) passes on master (`forge test --symbolic`, 256 runs) but on this branch fails with `[FAIL: incomplete symbolic execution (Timeout): symbolic execution timeout exceeded (30s)]`. On master `run_invariant` was never called, so `--symbolic` invariant tests fell through to fuzzing.

Suggestion: on `Incomplete`, record the symbolic metadata and continue into `evm.invariant_fuzz` (only hard-fail in an explicit symbolic-strict mode).

crates/evm/symbolic/src/executor/run.rs
```diff
@@ -29,7 +29,31 @@ impl SymbolicExecutor {
...
+            cx: SymCx::new(),
+            solver: Box::new(solver),
+            deferred_incomplete: None,
+            deadline: None,
+        }
+    }
+
+    fn reset_run_state(&mut self) {
+        self.deferred_incomplete = None;
+        self.deadline = self
```
stevencartavia: On master `symbolic.timeout` (default 30s) was a per-SMT-query budget only (`set-option :timeout` and the subprocess `wait_timeout`). `reset_run_state` now also uses the same value as a whole-run wall-clock deadline, applied to both `run` and `run_invariant`. So one config field is both the per-query limit and the total-run limit, and stateless symbolic runs gain a wall-clock timeout they didn't have before — a slow-but-progressing test can now fail as `Timeout`.

Suggestion: add a separate field (e.g. `symbolic.run_timeout`) or scope the wall-clock deadline to the invariant phase and document it.

crates/forge/src/runner.rs
```diff
@@ -3398,6 +3423,260 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                                TestStatus::Failure,
+                                None,
+                                None,
+                                symbolic_result,
+                            );
+                            return self.result;
+                        }
+                        Ok(_) => {
+                            let reason =
+                                "symbolic invariant counterexample did not replay".to_string();
```
stevencartavia: When an invariant is already false in the post-`setUp` state, `run_invariant_inner` returns a counterexample with an empty sequence. The success arm requires `!replayed.call_sequence.is_empty()`, so this falls into this "did not replay" branch and reports `[FAIL: symbolic invariant counterexample did not replay]` instead of the real reason.

Repro (invariant asserting `x == 1` where `x` stays 0): master prints `[FAIL: assertion failed: 0 != 1]`; this branch prints `[FAIL: symbolic invariant counterexample did not replay]`.

Suggestion: treat an empty sequence as a valid initial-state failure and report the actual predicate reason.

crates/forge/src/runner.rs
```diff
@@ -3398,6 +3423,260 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                    let failure = self
+                        .symbolic_sequence_failure(
+                            &symbolic_calls,
+                            invariant_config,
+                            &invariant_contract,
+                            invariant_contract.anchor(),
+                            false,
+                            &storage,
+                        )
+                        .unwrap_or(SymbolicSequenceFailure {
```
stevencartavia: A symbolic counterexample that doesn't reproduce under concrete replay is still reported as `[FAIL]`. This `unwrap_or(...)` synthesizes a failure when replay can't confirm one, and the `Ok(replayed) if !replayed.call_sequence.is_empty()` arm accepts "some calls replayed" as a violation without checking the invariant actually failed.

Repro on the current head (arbitrary-storage slot read via a mapping key, which isn't cached for replay since only concrete keys are recorded): `useIt(address)` requires `balanceOf(who) == 777`. Same binary, without `--symbolic`: `[PASS]` (256 runs, all calls revert — `balanceOf` is never 777). With `--symbolic`: `[FAIL: symbolic invariant counterexample]`, while the `-vvvvv` trace shows `balanceOf(0x0)` returning a random value, `useIt` reverting, `hit()` = `false`, and the invariant holding. Related to the copyStorage reseed flagged above, but the trigger (uncached/mapping slots) survives the copyStorage fix and the false-`[FAIL]` reporting is the separate issue.

Suggestion: don't `unwrap_or` a missing `symbolic_sequence_failure` — if replay with the model storage doesn't reproduce the failure, return a replay-mismatch/incomplete result instead of recording an invariant failure.

crates/forge/src/runner.rs
```diff
@@ -3496,6 +3521,233 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                                    failure_dir.as_path(),
+                                    primary_failure_file.as_path(),
+                                    &replayed.call_sequence,
+                                    &current_settings,
+                                    false,
+                                );
+
+                                let artifact = replayed.artifact.clone();
+                                let minimization = replayed.minimization.clone();
+                                let invariant_failures = vec![InvariantFailure::Predicate {
```
mablr: This path treats every confirmed symbolic invariant counterexample as an `InvariantFailure::Predicate`, but `run_invariant` can return the same counterexample shape for a target/handler call failure as for an invariant predicate failure. That means a handler `assert(false)` with a passing `invariant_ok()` can be reported and persisted as an `invariant_ok` predicate failure with `assertion_failure = false`. Could we carry the counterexample origin through `SymbolicInvariantRunResult::Counterexample` and route handler/assertion failures through the existing `InvariantFailure::Handler` / handler persistence path?

crates/forge/src/runner.rs
```diff
@@ -3496,6 +3521,233 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                                        ),
+                                    );
+                                if let Some(artifact) = artifact {
+                                    symbolic_result = symbolic_result.with_artifact(artifact);
+                                }
+                                if let Some(minimization) = minimization {
+                                    symbolic_result =
+                                        symbolic_result.with_minimization(minimization);
+                                }
+                                self.result.symbolic_result(
```
mablr: This overwrites the result kind with `TestKind::Symbolic` after we already built the invariant result above. The executed test is still an invariant campaign, so JSON/API consumers lose the `Invariant` fields such as runs/calls/reverts/workers and may stop recognizing this as an invariant failure. Could this attach the confirmed symbolic metadata with `record_symbolic(symbolic_result)` instead, leaving the invariant kind intact?

crates/evm/symbolic/src/runtime/state.rs
```diff
@@ -146,6 +146,14 @@ impl PathState {
...
+            for target in cheats.arbitrary_storage_copied_targets() {
+                self.world.enable_arbitrary_storage(target);
```
mablr: This makes copied arbitrary-storage targets independent symbolic storage. Concrete `copyStorage(A, B)` keeps untouched `B[slot]` tied to `A[slot]`, so symbolic execution can invent impossible `A[slot] != B[slot]` counterexamples. Please model the copy relation and derive copied-target symbols from the source slot instead of enabling them independently.

crates/forge/src/runner.rs
```diff
@@ -1500,7 +1560,7 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
 
         let replay = replay_error(
             replay_config,
-            self.clone_executor(),
+            self.clone_executor_with_symbolic_storage(symbolic_storage)?,
```
mablr, line 1719: This confirms storage-dependent counterexamples only for the current run. The artifact/rerun cache persists the call sequence but not these `SymbolicStorageAssignment`s, so a confirmed `vm.setArbitraryStorage` failure can later fail to reproduce via `--replay-symbolic-artifact` or `--rerun`. Please persist and reapply the storage assignments, with defaulted fields for compatibility.

crates/forge/src/runner.rs
```diff
@@ -3496,6 +3563,342 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                                                            InvariantPredicateResult {
+                                                                name: invariant.name.clone(),
+                                                                status: TestStatus::Failure,
+                                                                reason: failed_predicate_reason
+                                                                    .clone(),
+                                                            }
+                                                        } else {
+                                                            InvariantPredicateResult {
+                                                                name: invariant.name.clone(),
+                                                                status: TestStatus::Success,
```
mablr: This reports non-anchor predicates as passing even though the symbolic run only checked `invariant_contract.anchor()` and returns before fuzzing. In multi-predicate campaigns this can show `[PASS]` for predicates that never ran. Please omit unexecuted predicates, check the confirmed sequence against them, or continue into fuzz before emitting pass rows.

crates/forge/src/runner.rs
```diff
@@ -3496,6 +3563,342 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                            let invariant_signature = invariant_contract.anchor().signature();
+                            let artifact_file_name =
+                                if let Some((reverter, selector, _)) = handler_site {
+                                    format!("handler-{reverter}-{selector}")
+                                } else {
+                                    invariant_signature.clone()
+                                };
+                            let assertion_failure =
+                                matches!(kind, SymbolicInvariantCounterexampleKind::Handler);
+                            let artifact_replay_semantics = SymbolicCounterexampleReplaySemantics {
```
mablr, line 4039: The artifact only carries generic replay semantics here, not whether this was a handler assertion or predicate failure. Initial runs can preserve handler origin, but artifact replay later uses generic invariant replay and loses that attribution. Please persist counterexample origin/handler site in the artifact and use it during replay.

crates/forge/src/runner.rs
figtracer: current code no longer emits synthetic pass rows for unchecked predicates; it records the checked anchor failure plus skipped predicate rows. covered by `symbolic_invariant_omits_unchecked_predicate_pass_rows`.

crates/evm/symbolic/src/runtime/state.rs
figtracer: modeled now via `enable_arbitrary_storage_copy(source, target)`, so copied reads use the source storage symbol instead of independent target storage. covered by `copied_arbitrary_storage_uses_source_symbol_and_replays_both_accounts` and `symbolic_invariant_replays_copied_arbitrary_storage`.

crates/forge/src/runner.rs
figtracer: persisted storage assignments are now written into symbolic artifacts and invariant rerun cache, then reapplied before concrete replay. i added the missing `forge test --rerun` assertion in `test(symbolic): cover invariant storage rerun`; existing coverage also checks `forge fuzz replay` and `--replay-symbolic-artifact`.

crates/forge/src/runner.rs
figtracer: artifact replay now persists `invariant_failure` with predicate/handler origin and uses it during sequence replay, including handler fail-on-revert semantics and handler result routing. covered by `symbolic_counterexample_artifact_serializes_invariant_replay_metadata` and `symbolic_invariant_handler_failure_stays_handler`.

crates/forge/src/runner.rs
```diff
@@ -4334,6 +4894,9 @@ struct InvariantPersistedFailure {
     /// Whether the persisted failure came from a handler assertion instead of the invariant body.
     #[serde(default)]
     assertion_failure: bool,
+    /// Concrete setup-storage assignments required before replaying this failure.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    storage: Vec<SymbolicStorageAssignment>,
```
mablr, line 5127: This new persisted storage is applied for predicate reruns and `--replay-symbolic-artifact`, but the handler replay path drops it: `replay_persisted_handler_failures` reads `persisted_call_sequence(...)`, takes only `persisted.call_sequence`, and then calls `replay_handler_failure_sequence(executor.clone(), ...)` without applying `persisted.storage`. That means a handler-side symbolic failure that depends on `vm.setArbitraryStorage` can be recorded successfully but disappear under `forge test --rerun`. Could we apply the stored assignments in that handler replay path too and add a handler+arbitrary-storage rerun regression?

crates/evm/symbolic/src/runtime/state.rs
```diff
@@ -1663,16 +1709,15 @@ impl SymbolicWorld {
...
         cx: &mut SymCx,
         executor: &Executor<FEN>,
         address: Address,
         key: &SymExpr,
         concrete_key: Option<U256>,
     ) -> Result<SymExpr, SymbolicError> {
-        if self.arbitrary_storage_all || self.arbitrary_storage_accounts.contains(&address) {
-            let name = stable_symbol(cx, "storage", format!("{address:?}:{key:?}").as_bytes());
-            return Ok(SymExpr::get_var(cx, name));
+        if let Some(base) = self.arbitrary_storage_base(cx, address, key, concrete_key) {
```
mablr: Blocking: created accounts need a zero-default storage path after arbitrary-storage overlays. Today a freshly deployed contract is only tracked in `current_transaction_created_accounts`, which is cleared at top-level boundaries, so later unwritten symbolic slots become arbitrary/base storage and BEC-style counterexamples do not replay. Persist created accounts in `SymbolicWorld`/snapshots and return zero from `storage_base` when the address was created.

crates/evm/symbolic/src/runtime/state.rs
```diff
@@ -1690,11 +1735,42 @@ impl SymbolicWorld {
         } else if self.zero_init_symbolic_storage {
             Ok(SymExpr::zero(cx))
         } else {
-            let name = stable_symbol(cx, "storage", format!("{address:?}:{key:?}").as_bytes());
+            let name = symbolic_storage_symbol(cx, address, key);
```
mablr, line 1776: Blocking: concrete mapping writes can still fail to alias symbolic mapping reads because a concrete 64-byte `keccak(key, slot)` folds to a const and loses its preimage/layout. This misses cases like `mapping[0] = 1; if (mapping[root] != 0)` with `root == 0` (NomadZeroRoot), and struct fields add `+ fieldOffset`. Cache 64-byte concrete keccak preimages and teach storage-key equality to reconstruct mapping layouts, including nested mappings and const offsets.

crates/forge/src/runner.rs
```diff
@@ -1526,24 +1610,34 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
         &self,
         test_name: &str,
         artifact_file_name: &str,
         call_sequence: &[BaseCounterExample],
         minimization: Option<MinimizedSequence>,
+        replay_semantics: SymbolicCounterexampleReplaySemantics,
+        storage: &[SymbolicStorageAssignment],
+        invariant_failure: Option<SymbolicInvariantArtifactFailure>,
     ) -> (Option<SymbolicArtifactRef>, Option<SymbolicCounterexampleMinimization>) {
         let Some(minimization) = minimization.filter(MinimizedSequence::changed) else {
```
stevencartavia: These two z3-gated tests pass on master but fail on this branch (they're skipped in CI since the runners have no z3):

- `symbolic_json_reports_sequence_counterexample_artifact`
- `symbolic_json_reports_minimized_sequence_counterexample`

Both fail at `assert_eq!(failure["artifact"], minimization["minimized"])`. For these fixtures the symbolic pre-fuzz phase now returns an already-minimal sequence, so minimization is a no-op and this `minimization.filter(MinimizedSequence::changed)` branch drops the `minimization` field and one artifact. On master these came from the fuzz path, so minimization changed the sequence and populated the metadata (the minimized-sequence test reported `original: 20, shrunk: 2`, the artifact test shrank to length 1).

The counterexample is correct, so it's not a soundness issue — just a JSON output change plus stale tests. Either update the two tests, or keep minimization metadata for already-minimal sequences (like the stateless path does).

crates/evm/symbolic/src/runtime/state.rs
```diff
@@ -1603,6 +1629,28 @@ impl SymbolicWorld {
...
 
+    pub(crate) fn enable_arbitrary_storage_copy(&mut self, source: Address, target: Address) {
+        self.arbitrary_storage_copies.insert(target, source);
+    }
+
+    pub(crate) fn replay_storage_assignments(
+        &self,
+        model: &SymbolicModel,
+    ) -> Vec<SymbolicStorageAssignment> {
+        self.replay_storage_slots
```
mablr: Nice-to-have: consider materializing through a deterministic `(address, slot)` map before returning. That would dedup identical aliases and turn conflicting alias values into an explicit incomplete/error instead of depending on `HashMap` iteration + last-write-wins replay.

crates/evm/symbolic/src/runtime/state.rs
```diff
@@ -1690,11 +1743,42 @@ impl SymbolicWorld {
...
+    fn arbitrary_storage_base(
+        &mut self,
+        cx: &mut SymCx,
+        address: Address,
+        key: &SymExpr,
+        concrete_key: Option<U256>,
+    ) -> Option<SymExpr> {
+        let copied_source = self.arbitrary_storage_copies.get(&address).copied();
+        let has_arbitrary_storage = self.arbitrary_storage_accounts.contains(&address);
+        let symbol_address = copied_source
```
mablr: Nice-to-have: this would be a bit closer to concrete cheatcode behavior if explicit arbitrary storage took precedence over copied storage here. `copyStorage(from, to); setArbitraryStorage(to);` is reachable, and concrete SLOAD treats `to` as its own arbitrary namespace.

crates/evm/symbolic/README.md
```diff
@@ -315,6 +315,40 @@ set, generates symbolic arguments with the same ABI model used for stateless
...
+
+    targetContract(address(this));
+}
+```
+
+If the harness imports a smaller Hevm interface that does not expose this
+cheatcode, declare a local interface with `setArbitraryStorage(address)` and
+cast it to `address(vm)`.
+
+Keep this scoped to external dependencies that model the environment, such as
```
mablr, line 347: Nice-to-have: since the replay handoff currently only materializes slots that are concrete during symbolic execution, it may be worth calling that out here. Mapping entries keyed by symbolic calldata still replay-filter to `Incomplete`/`mismatch` rather than confirmed counterexamples.

mablr (commented):
Looks good overall.

I left 3 non-blocking suggestions, but can be handled later in follow-up as the PR is already big.

crates/evm/symbolic/src/runtime/expr/expr.rs
```diff
@@ -210,33 +214,43 @@ impl SymExpr {
...
     }
 
+    fn storage_mapping_key_bytes(&self, cx: &SymCx) -> Option<Arc<[Self]>> {
+        match self.kind() {
+            SymExprKind::Keccak { len, bytes, .. }
+                if len.as_const() == Some(U256::from(64)) && bytes.len() >= 64 =>
+            {
+                Some(bytes.clone())
+            }
+            SymExprKind::Const(hash) => cx.concrete_keccak_preimage(*hash),
```
mablr, line 247: Seems to regress the Nomad zero-root repro in `grandizzy/symbolic-bug-suite`. `forge test --symbolic --match-test checkOnlyInstalledRootsAreAccepted -vv` now ends as `Incomplete (Timeout): solver returned unknown` after ~30s with `paths: 0, queries: 3`; the `master` build finds `root = 0x00…00` and replay-confirms in ~123ms. Replaying the old cached `root=0` artifact still works on this branch, so this looks isolated to symbolic discovery around concrete constructor-written mapping storage / symbolic mapping keys.

crates/evm/symbolic/src/runtime/state.rs
```diff
@@ -146,6 +146,14 @@ impl PathState {
             .and_then(|cheats| cheats.gas_price)
             .unwrap_or_else(|| executor.tx_env().gas_price());
         self.gas_price = SymExpr::constant(cx, U256::from(gas_price));
+        if let Some(cheats) = executor.inspector().cheatcodes.as_ref() {
+            for target in cheats.arbitrary_storage_targets() {
```
mablr: **Blocker:** this imports the target but drops `setArbitraryStorage`'s `overwrite` mode. A nonzero setup slot marked with the default `overwrite=false` becomes symbolic, and replay then forcibly writes the model value, so an unreachable state can be reported as a confirmed counterexample. Preserve concrete nonzero slots unless overwrite is enabled, including copied-storage aliases, and cover false/true overwrite modes.

crates/evm/symbolic/src/executor/run.rs
```diff
@@ -585,13 +598,16 @@ impl SymbolicExecutor {
...
-                                                    &invariant_outcome.state,
-                                                )?;
+                                                let (sequence, storage) = self
+                                                    .materialize_sequence(
+                                                        &steps,
+                                                        &invariant_outcome.state,
+                                                    )?;
                                                 return Ok(
                                                     SymbolicInvariantRunResult::Counterexample {
+                                                        kind: SymbolicInvariantCounterexampleKind::Predicate,
```
mablr: **Blocker:** successful invariant checks carry `invariant_outcome.state` into the next frontier, including mutations made by the invariant and `afterInvariant`. Concrete checks are noncommitting raw calls, and `afterInvariant` runs at the run boundary rather than after every prefix. Carry the post-handler world forward while retaining only the check's path constraints.

crates/forge/src/runner.rs
```diff
@@ -3498,6 +3698,352 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+
+            let mut symbolic = SymbolicExecutor::new(self.config.symbolic.clone());
+            match symbolic.run_invariant(SymbolicInvariantRunInput {
+                executor: &evm.executor,
+                invariant_address: invariant_contract.address,
+                sender: self.sender,
+                invariant: invariant_contract.anchor(),
+                after_invariant,
+                targets: symbolic_targets,
+                senders: sender_filters.targeted,
```
mablr, line 3944: **Blocker:** an empty targeted-sender list means unrestricted senders to the concrete fuzzer, but symbolic execution turns it into one fixed sender and can still report `Safe`. The same reduced-domain issue exists for payable values, configured warp/roll delays, and updatable/call-override targets. Model these dimensions or return `Incomplete` whenever the campaign permits them.

crates/forge/src/runner.rs
```diff
@@ -3475,7 +3673,9 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
 
         // Replay persisted handler bugs; feed still-reproducing ones into the campaign,
         // delete stale files in place.
-        let persisted_handler_failures = replay_persisted_handler_failures(
+        let mut symbolic_handler_storage =
```
mablr: **Blocker:** this map is populated only for failures discovered in the current symbolic run. `replay_persisted_handler_failures` applies persisted storage but returns only the error, so the first rerun rewrites the failure with `storage: []` and a second rerun can no longer reproduce it. Return the selected failure's storage alongside the error and seed this map from replayed entries.

crates/forge/src/runner.rs
```diff
@@ -3498,6 +3698,352 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                    let symbolic_workers = 1;
+                    let failed_corpus_replays = 0;
+                    let symbolic_calls = symbolic_invariant_counterexample_calls(
+                        &sequence,
+                        identified_contracts,
+                        show_solidity,
+                    );
+                    if let Some(failure) = self.symbolic_sequence_failure(
+                        &symbolic_calls,
+                        invariant_config,
```
mablr: **Blocker:** symbolic exploration uses the anchor's effective inline `fail_on_revert`, but confirmation receives the global `invariant_config` and artifact metadata later reads the global config again. With global false / inline true, a valid exact revert witness is discarded. Resolve this value once per anchor and thread it through confirmation, minimization, and persistence.

crates/forge/src/runner.rs
```diff
@@ -2681,14 +2798,72 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
-                            self.result.invariant_replay_fail(
-                                outcome.replayed_entirely,
-                                &invariant.signature(),
-                                outcome.reason.or_else(|| artifact.replay.reason.clone()),
-                                outcome.calls_count,
-                                outcome.reverts,
-                                calls,
-                            );
+                            let reason = outcome.reason.or_else(|| artifact.replay.reason.clone());
+                            if let Some(SymbolicInvariantArtifactFailure::Handler {
```
mablr, line 2959: **Blocker:** artifact replay trusts this stored handler origin whenever any concrete failure occurs; it never compares `outcome.failure_site` or the stored fingerprint. A stale artifact can therefore be accepted and attributed to the wrong handler or edge. Require the current `CheckSequenceFailureSite` to preserve the stored site before rendering a confirmed handler failure.

crates/evm/symbolic/src/executor/run.rs
```diff
@@ -477,9 +505,12 @@ impl SymbolicExecutor {
             &mut completed_paths,
         )? {
             if outcome.failed {
-                let sequence = self.materialize_sequence(&initial.steps, &outcome.state)?;
+                let (sequence, storage) =
+                    self.materialize_sequence(&initial.steps, &outcome.state)?;
                 return Ok(SymbolicInvariantRunResult::Counterexample {
```
mablr: **Blocker:** this initial check and the checks after every successful call ignore configured `check_interval` semantics (`0` means end-only; `N` means every N calls plus the end). That can turn intentionally unchecked transient violations into replay mismatches and `Incomplete`. Pass the effective interval into symbolic execution and match concrete check scheduling.

crates/forge/src/runner.rs
```diff
@@ -2658,9 +2760,23 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
                         }
                     }
                 }
+                let artifact_failure = artifact.invariant_failure.as_ref();
+                let is_handler_artifact = matches!(
+                    artifact_failure,
+                    Some(SymbolicInvariantArtifactFailure::Handler { .. })
+                );
+                let executor = match self.clone_executor_with_symbolic_storage(&artifact.storage) {
```
mablr: **Blocker:** artifact storage is applied here only after updatable-target validation has already executed the sequence on a plain setup executor. If the stored state controls whether a call succeeds or creates/registers a target, preflight can reject a valid artifact or derive a different target set. Build both validation and replay executors from the same artifact-adjusted setup state.

crates/evm/symbolic/src/executor/run.rs
```diff
@@ -526,79 +562,126 @@ impl SymbolicExecutor {
...
+                                            input.depth,
+                                            input.check_interval,
+                                        ) {
+                                            for invariant_outcome in self.execute_invariant_check(
+                                                input.executor,
+                                                reverted_state.clone(),
+                                                input.invariant_address,
+                                                input.sender,
+                                                input.invariant,
+                                                input.after_invariant,
```
grandizzy: **High:** This runs `afterInvariant` at intermediate `check_interval` boundaries, but the concrete invariant campaign runs that hook only at the end of the complete run.

With depth 2, interval 1, and a target that toggles a flag, the hook can fail after call 1 but pass after call 2. Symbolic execution reports the one-call prefix, and replay ‘confirms’ it because it treats that shortened prefix as the run boundary.

Pass the hook only for the terminal check:

```suggestion
                                                if steps.len() == input.depth {
                                                    input.after_invariant
                                                } else {
                                                    None
                                                },
```

The same change is needed in the successful-call branch below.

crates/evm/symbolic/src/executor/run.rs
```diff
@@ -526,79 +562,126 @@ impl SymbolicExecutor {
...
+                                            input.depth,
+                                            input.check_interval,
+                                        ) {
+                                            for invariant_outcome in self.execute_invariant_check(
+                                                input.executor,
+                                                outcome.state.clone(),
+                                                input.invariant_address,
+                                                input.sender,
+                                                input.invariant,
+                                                input.after_invariant,
```
grandizzy: Apply the same terminal-only handling here:

```suggestion
                                                if steps.len() == input.depth {
                                                    input.after_invariant
                                                } else {
                                                    None
                                                },
```

crates/forge/src/runner.rs
```diff
@@ -97,6 +102,81 @@ fn should_symbolically_use_fuzz_frontiers(config: &Config, func: &Function) -> b
...
+        ) if *reverter == target && *selector == actual_selector && *fingerprint == actual_fingerprint
+    )
+}
+
+const fn symbolic_artifact_predicate_failure_matches(
+    failure: Option<&SymbolicInvariantArtifactFailure>,
+    outcome: &CheckSequenceOutcome,
+    fail_on_revert: bool,
+) -> bool {
+    if !matches!(failure, Some(SymbolicInvariantArtifactFailure::Predicate { .. })) {
```
grandizzy: **High:** This treats invariant, `afterInvariant`, and fail-on-revert sequence failures as equivalent predicate origins. A stale artifact can therefore be accepted after the original predicate stops failing and replay instead fails in a different location.

Predicate artifacts need to persist the exact confirmed site, similar to handler artifacts. For example:

```rust
enum SymbolicInvariantArtifactFailure {
    Predicate {
        name: String,
        site: SymbolicPredicateFailureSite,
    },
    // ...
}

enum SymbolicPredicateFailureSite {
    Invariant { target: Address, selector: Selector, fingerprint: B256 },
    AfterInvariant { target: Address, selector: Selector, fingerprint: B256 },
    SequenceRevert { target: Address, selector: Selector, fingerprint: B256 },
}
```

Then replace this broad matcher with equality against the stored site. Legacy artifacts without site metadata should be rejected as unverifiable.

crates/forge/src/runner.rs
```diff
@@ -3498,6 +3809,376 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+            symbolic_invariant_config.fail_on_revert = anchor_fail_on_revert;
+            let mut symbolic = SymbolicExecutor::new(self.config.symbolic.clone());
+            match symbolic.run_invariant(SymbolicInvariantRunInput {
+                executor: &evm.executor,
+                invariant_address: invariant_contract.address,
+                sender: self.sender,
+                invariant: invariant_contract.anchor(),
+                after_invariant,
+                targets: symbolic_targets,
+                senders: sender_filters.targeted,
```
grandizzy, line 3944: **High:** Only the filtered targeted list is passed here; sender exclusions are dropped. When this list is empty, symbolic execution falls back to `input.sender` even if `excludeSender(input.sender)` was configured.

The unsupported-domain marker only downgrades `Safe`, so an out-of-domain counterexample still proceeds through replay and is reported.

I suggest carrying `excluded_senders: Vec<Address>` in `SymbolicInvariantRunInput`, passing `sender_filters.excluded`, and rejecting an excluded fallback:

```rust
let senders = if input.senders.is_empty() {
    if input.excluded_senders.contains(&input.sender) {
        return Err(SymbolicError::Unsupported(
            "symbolic invariant has no supported sender",
        ));
    }
    vec![input.sender]
} else {
    input.senders.clone()
};
```

Materialized calls should also be validated against both targeted and excluded sender filters before confirmation.

grandizzy (changes requested):
Found three remaining soundness gaps that can report confirmed symbolic invariant failures outside the concrete campaign semantics. Details and suggested changes inline.

crates/evm/symbolic/src/executor/run.rs
```diff
@@ -526,79 +562,126 @@ impl SymbolicExecutor {
...
                                                     stats: self.stats_with_paths(completed_paths),
                                                 },
                                             );
                                         }
+                                        // A reverted top-level call cannot change persistent
+                                        // state, but it still consumes one invariant sequence
+                                        // step. Preserve the pre-call world together with the
+                                        // reverted branch constraints so end-only and periodic
+                                        // invariant checks observe the same call schedule as the
+                                        // concrete campaign.
```
mablr, line 599: This restores the pre-call `PathState`, but concrete top-level reverts retain non-journaled effects such as `vm.mockCall` and `vm.warp`. A reverting handler can therefore affect the next concrete call while symbolic execution drops that effect and may report `Safe`.

crates/evm/symbolic/src/runtime/state.rs
```diff
@@ -1690,11 +1759,91 @@ impl SymbolicWorld {
...
+        } else {
+            copied_source?
+        };
+        let symbol = symbolic_storage_symbol(cx, symbol_address, key);
+        if let Some(slot) = concrete_key.or_else(|| key.as_const()) {
+            if has_arbitrary_storage {
+                self.record_replay_storage_slot(symbol, address, slot);
+            }
+            if let Some(source) = copied_source {
+                self.record_replay_storage_slot(symbol, source, slot);
```
mablr, line 1836: Concrete `ArbitraryStorage::copy` writes a newly generated value to both source and target storage. Here we only record replay assignments, so later symbolic reads can observe a stale source value that concrete replay overwrites.

crates/forge/src/runner.rs
```diff
@@ -3816,18 +4499,13 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
-                    .and_then(|(contract_name, abi)| {
-                        abi.functions()
-                            .find(|f| f.selector() == selector)
-                            .map(|f| format!("{contract_name}::{}", f.name))
-                    })
-                    .unwrap_or_else(|| format!("{reverter}::{selector}"));
+                let resolved_name =
+                    invariant_handler_failure_name(identified_contracts_ro, reverter, selector);
+                let symbolic_storage =
+                    symbolic_handler_storage.get(site).map(Vec::as_slice).unwrap_or(&[]);
```
mablr: This storage is keyed only by handler site, but the following fuzz campaign can replace the seeded symbolic failure with a shorter sequence at the same site. We then attach stale storage to a different sequence and persist it as replay-confirmed.

crates/forge/src/runner.rs
```diff
@@ -4334,9 +5029,15 @@ struct InvariantPersistedFailure {
     /// Whether the persisted failure came from a handler assertion instead of the invariant body.
     #[serde(default)]
     assertion_failure: bool,
+    /// Concrete setup-storage assignments required before replaying this failure.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    storage: Vec<SymbolicStorageAssignment>,
```
mablr, line 5127: Persisting only the sequence and storage loses nested handler origin. Initial replay can identify `nested reverter + outer selector`, while rerun reconstructs `outer target + outer selector`, changing the site and fingerprint.

crates/forge/src/runner.rs
grandizzy: This is partially addressed: fresh symbolic artifacts now persist and compare `SymbolicInvariantFailureSite`. However, `site: None` still falls back to broad origin matching, and the persisted-failure replay path drops `failure_site` and generates new `site: None` artifacts.

Could we reject unverifiable artifacts and preserve/validate the concrete replay site through persisted replay? Suggested change:

```diff
diff --git a/crates/forge/src/runner.rs b/crates/forge/src/runner.rs
@@
 fn symbolic_artifact_predicate_failure_matches(
     failure: Option<&SymbolicInvariantArtifactFailure>,
     outcome: &CheckSequenceOutcome,
-    fail_on_revert: bool,
 ) -> bool {
-    let Some(SymbolicInvariantArtifactFailure::Predicate { site, .. }) = failure else {
+    let Some(SymbolicInvariantArtifactFailure::Predicate {
+        site: Some(expected),
+        ..
+    }) = failure
+    else {
         return false;
     };
-
-    if let Some(site) = *site {
-        return symbolic_invariant_failure_site_matches(site, outcome.failure_site);
-    }
-
-    match outcome.failure_site {
-        Some(
-            CheckSequenceFailureSite::Invariant { .. }
-            | CheckSequenceFailureSite::AfterInvariant { .. },
-        ) => true,
-        Some(CheckSequenceFailureSite::SequenceCall { .. }) => {
-            fail_on_revert && !outcome.sequence_assertion_failure
-        }
-        None => false,
-    }
+    symbolic_invariant_failure_site_matches(*expected, outcome.failure_site)
 }
@@
     ) && !symbolic_artifact_predicate_failure_matches(
         artifact_failure,
         &outcome,
-        artifact.replay_semantics.fail_on_revert,
     ) {
@@
     let Some(InvariantPersistedFailure {
         mut call_sequence,
         assertion_failure,
         storage,
-        ..
+        failure_site,
     }) = persisted_invariant_failure(...)
@@
     if let Ok(replay) = replay
         && !replay.success
     {
+        let Some(confirmed_failure_site) =
+            replay.failure_site.map(symbolic_invariant_failure_site)
+        else {
+            continue;
+        };
+        if failure_site.is_some_and(|expected| expected != confirmed_failure_site) {
+            continue;
+        }
+
         // Existing replay handling...
@@
         if let Some(updated_replay) = replay.check_result {
+            if !symbolic_invariant_failure_site_matches(
+                confirmed_failure_site,
+                updated_replay.failure_site,
+            ) {
+                continue;
+            }
             // Existing result updates...
         }
@@
         record_invariant_failure_with_storage(
             // ...
-            None,
+            Some(confirmed_failure_site),
         );
@@
         Some(SymbolicInvariantArtifactFailure::Predicate {
             name: replay_contract.anchor().name.clone(),
-            site: None,
+            site: Some(confirmed_failure_site),
         }),
```

This still permits legacy persisted failures to be upgraded from a concrete replay, but prevents artifacts from being accepted or republished without an exact confirmed origin.

grandizzy:
One replay-integrity issue remains after the latest fix: fresh symbolic artifacts now record an exact `SymbolicInvariantFailureSite`, but artifacts with `site: None` still use broad matching, and the persisted-failure replay path drops `failure_site` and emits new `site: None` artifacts.

That means a stale artifact can still be accepted when the original predicate no longer fails but the same sequence fails in `afterInvariant`, at another invariant edge, or through `fail_on_revert`.

The persisted replay should validate and preserve the concrete failure site, and artifact replay should reject missing site metadata as unverifiable. Proposed diff: https://github.com/foundry-rs/foundry/pull/15591#discussion_r3570298567

grandizzy (dismissed):

grandizzy (approved):
lgtm

grandizzy (dismissed):

mablr (commented):
I think these previous comments are not fully addressed:

https://github.com/foundry-rs/foundry/pull/15591#discussion_r3568911061
https://github.com/foundry-rs/foundry/pull/15591#discussion_r3568923617
https://github.com/foundry-rs/foundry/pull/15591#discussion_r3568923630

Unsure we need to block on them, but i wanted to pin them at least for follow-up.

mablr (approved):
LGTM

grandizzy (approved):
lgtm

PR #15592, opened 2026-07-06
feat(debugger): toggle variables pane

figtracer:
Adds an uppercase `V` debugger shortcut that hides or restores the Variables pane while keeping the existing lowercase `v` source-pane toggle intact.

When the variables pane is hidden, the debugger reclaims that space for the stack and active buffer views in both horizontal and vertical layouts. The shortcut help now includes `[V] variables`, and the behavior is covered by focused status and draw tests.

This is an incremental per-pane hide/show improvement toward #927.


grandizzy (approved):
lgtm

mablr (approved):

PR #15593, opened 2026-07-06
feat(debugger): add pane toggle commands

figtracer:
Adds command-mode pane visibility controls for source, variables, and stack: `:source`/`:src`, `:variables`/`:vars`, and `:stack`.

This keeps lower-frequency pane configuration behind the existing `:` command prompt instead of adding more global shortcuts. The previously added source and variables pane shortcuts are removed from the key handler and shortcut footer, while the layout splitting now composes the visible variables and stack panes so hidden panes give their space back to the remaining views.


grandizzy (approved):
lgtm

0xKarl98 (approved):

PR #15595, opened 2026-07-06
perf(forge): speed up test --list

figtracer:
## Summary
`forge test --list` only needs contract ABIs to identify matching test functions. Requesting full bytecode output and constructing runners adds avoidable work, especially for generated or large projects where users and agents repeatedly ask for narrow test lists.

- Compile ABI-only for `forge test --list`.
- Build the list output directly from compiler artifacts using the existing test matcher/filter logic.
- Preserve the existing `--mutate` incompatibility behavior and add CLI regression coverage.

## Benchmarks

Generated 5k-contract fixture, fresh first-run `forge test --list --match-test testTarget -q` after the review update:

- master: `3.42s`
- this change: `0.49s`

Plain and JSON list output were also diffed against master during local investigation.

### DIFF
--- a/crates/forge/src/cmd/test/mod.rs
+++ b/crates/forge/src/cmd/test/mod.rs
@@ -29,7 +29,8 @@ use foundry_cli::{
     utils::{self, LoadConfig},
 };
 use foundry_common::{
-    EmptyTestFilter, TestFilter, TestFunctionExt, compile::ProjectCompiler, fs, shell,
+    EmptyTestFilter, TestFilter, TestFunctionExt, TestFunctionKind, compile::ProjectCompiler, fs,
+    shell,
 };
 use foundry_compilers::{
     ProjectCompileOutput,
@@ -1179,7 +1180,7 @@ impl TestArgs {
         }
 
         // Set up the project.
-        let project = config.project()?;
+        let mut project = config.project()?;
         let project_root = project.paths.root.clone();
 
         let replay_symbolic_artifact = self.load_symbolic_artifact_replay()?;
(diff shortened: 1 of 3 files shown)


crates/forge/src/cmd/test/mod.rs
```diff
@@ -2940,19 +2973,70 @@ fn list<FEN: FoundryEvmNetwork>(
...
+) -> Result<TestOutcome> {
+    let matcher = TestFunctionMatcher::new(config, inline_config, symbolic_artifact_replay);
+    let results = output
+        .artifact_ids()
+        .filter_map(|(id, artifact)| {
+            let abi = artifact.abi.as_ref()?;
+            let id = id.with_stripped_file_prefixes(&config.root);
+            if !matcher.matches_contract(filter, &id, abi) {
+                return None;
+            }
```
grandizzy, line 3001: `--list` now includes contracts that `forge test` won't actually run.

The old runner path lists only `deployable_contracts`, which apply two filters beyond `matches_contract` ([`multi_runner.rs#L753-L766`](https://github.com/foundry-rs/foundry/blob/master/crates/forge/src/multi_runner.rs#L753-L766)):
1. constructor has no arguments, and
2. non-empty deployable bytecode.

This new path drops both and iterates every `artifact_id`, so `--list` will now surface **abstract base test contracts** that declare their own `test_*` functions, contracts with constructor args, and interfaces with `test_`-named functions — none of which get deployed/run by `forge test`. That's a visible inconsistency between `--list` and execution.

The empty-bytecode/abstract check can't be done under ABI-only output (that's the whole point of the perf win), but the constructor check is free from the ABI and restores one of the two conditions:

```suggestion
            let abi = artifact.abi.as_ref()?;
            let id = id.with_stripped_file_prefixes(&config.root);
            let deployable =
                abi.constructor.as_ref().map(|c| c.inputs.is_empty()).unwrap_or(true);
            if !deployable || !matcher.matches_contract(filter, &id, abi) {
                return None;
            }
```

Worth adding a test with an `abstract contract Base { function test_x() ... }` to pin the intended behavior either way.

grandizzy (changes requested):
Net: solid, correct for the common case; the one thing I'd raise is the abstract/constructor-arg over-inclusion vs actual execution (see inline).

mablr (approved):
sgtm

grandizzy (approved):

PR #15596, opened 2026-07-06
perf(forge): speed up selectors

figtracer:
## Summary

The selector subcommands only inspect ABIs. The previous setup added ABI output on top of the default compiler outputs, so these read-only commands still paid for bytecode-oriented output they do not consume.

- Use ABI-only compilation for `forge selectors list`, `find`, `cache`, and `upload`.
- Replace repeated additive `extra_output = Abi` build setup with a shared ABI-only project helper.

## Benchmarks

Generated 5k-contract source fixture, fresh first-run `forge selectors list C0000`:

- master: `10.05s`
- this change: `2.13s`

Isolated warmed copies of the same fixture:

- master: `493.6ms ± 58.4ms`
- this change: `366.4ms ± 31.8ms`

The selector table output was byte-for-byte identical against master in the local comparison.

### DIFF
--- a/crates/forge/src/cmd/selectors.rs
+++ b/crates/forge/src/cmd/selectors.rs
@@ -3,7 +3,7 @@ use clap::Parser;
 use comfy_table::{Table, modifiers::UTF8_ROUND_CORNERS, presets::ASCII_MARKDOWN};
 use eyre::Result;
 use foundry_cli::{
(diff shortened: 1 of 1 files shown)


mattsse (approved):
lgtm

mablr (approved):

PR #15600, opened 2026-07-06
feat(debugger): add opcodes pane command

figtracer:
Adds command-mode visibility controls for the debugger opcode/trace pane: `:opcodes`, `:opcode`, and `:ops`. This continues the pane-command pattern from the previous source/variables/stack work without adding another global shortcut.

When the opcode pane is hidden, the vertical layout omits it and the horizontal layout lets the remaining left/right panes reclaim the space. If both source and opcodes are hidden, the variables/stack/buffer panes use the full width.

This branch also carries a tiny current-master compile unblocker in `forge test --list`: the new list-from-compile-output path now derives `generated_symbolic_regression` from the ABI before calling `test_function_kind`, matching the existing `MultiContractRunner::list` path. Without that, `cargo build -p forge --bin forge` fails on current `master` because `test_function_kind` now requires the third argument.

### DIFF
--- a/crates/debugger/src/tui/context.rs
+++ b/crates/debugger/src/tui/context.rs
@@ -89,6 +89,7 @@ pub(crate) struct TUIContext<'a> {
     /// Whether to decode active buffer as utf8 or not.
     pub(crate) buf_utf: bool,
     pub(crate) show_shortcuts: bool,
+    pub(crate) show_opcodes: bool,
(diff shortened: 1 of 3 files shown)


mattsse (approved):
lgtm

mablr (approved):

mattsse (approved):
LGTM

PR #15601, opened 2026-07-06
perf(symbolic): use two-var solver witnesses

figtracer:
Extends the local symbolic solver witness path from one eval var to bounded two-var constraints. Candidate models are still validated against the original path constraints before caching.

This is intentionally narrow: it requires an atomic two-var relation plus concrete one-var bounds for each searched var, and skips hard arithmetic, symbolic hashes, and `gasleft`-shaped constraints so solver/cache/portfolio behavior still exercises SMT where expected.

| --- | ---: | ---: | ---: |
| `grandizzy/symbolic-bug-suite` expected failures | 23/23 | 23/23 | unchanged |
| `grandizzy/symbolic-bug-suite` SMT queries | 81 | 63 | -22.2% |
| `grandizzy/symbolic-bug-suite` solver time | 980ms | 659ms | -32.8% |
| `grandizzy/symbolic-bug-suite` SMT input bytes | 48,307 | 39,483 | -18.3% |

### DIFF
--- a/crates/evm/symbolic/src/runtime/solver/hard_arith_fallback.rs
+++ b/crates/evm/symbolic/src/runtime/solver/hard_arith_fallback.rs
@@ -303,20 +303,23 @@ pub(crate) fn fallback_single_var_model(constraints: &[SymBoolExpr]) -> Option<S
 }
 
 pub(crate) fn fallback_two_var_model(constraints: &[SymBoolExpr]) -> Option<SymbolicModel> {
-    if constraints.iter().any(SymBoolExpr::contains_hard_arith)
-        || constraints.iter().any(SymBoolExpr::contains_symbolic_hash)
-        || constraints.iter().any(SymBoolExpr::contains_gasleft)
-    {
+    if constraints.iter().any(SymBoolExpr::contains_hard_arith) {
         return None;
     }
 
     let mut vars = SymbolicVars::default();
-    let mut constants = HashSet::<U256>::default();
     for constraint in constraints {
         collect_bool_fallback_vars(constraint, &mut vars);
-        collect_bool_constants(constraint, &mut constants);
+        if vars.len() > 2 {
+            return None;
+        }
     }
     if vars.len() != 2 {
+        return None;
+    }
+    if constraints.iter().any(SymBoolExpr::contains_symbolic_hash)
+        || constraints.iter().any(SymBoolExpr::contains_gasleft)
+    {
         return None;
     }
     if !constraints_have_two_var_relation(constraints, &vars)
@@ -325,6 +328,10 @@ pub(crate) fn fallback_two_var_model(constraints: &[SymBoolExpr]) -> Option<Symb
         return None;
     }
 
+    let mut constants = HashSet::<U256>::default();
+    for constraint in constraints {
+        collect_bool_constants(constraint, &mut constants);
+    }
     let mut constants = constants.into_iter().collect::<Vec<_>>();
     constants.sort_unstable();
     let vars = vars.into_iter().collect::<Vec<_>>();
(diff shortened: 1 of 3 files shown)


mattsse (dismissed):
lgtm

crates/evm/symbolic/src/runtime/solver/hard_arith_fallback.rs
```diff
@@ -302,6 +302,120 @@ pub(crate) fn fallback_single_var_model(constraints: &[SymBoolExpr]) -> Option<S
     None
 }
 
+pub(crate) fn fallback_two_var_model(constraints: &[SymBoolExpr]) -> Option<SymbolicModel> {
+    if constraints.iter().any(SymBoolExpr::contains_hard_arith)
+        || constraints.iter().any(SymBoolExpr::contains_symbolic_hash)
+        || constraints.iter().any(SymBoolExpr::contains_gasleft)
```
grandizzy: This runs on every solver call that reaches here, and for the common "not two vars" case it does three full tree walks plus var+constant collection before the cheap `vars.len() != 2` check bails. Suggest reordering so the var count gates first:

```rust
if constraints.iter().any(SymBoolExpr::contains_hard_arith) {
    return None; // keep first so hard-arith routes to its own fallback
}

let mut vars = SymbolicVars::default();
for constraint in constraints {
    collect_bool_fallback_vars(constraint, &mut vars);
    if vars.len() > 2 {
        return None;
    }
}
if vars.len() != 2 {
    return None;
}

if constraints.iter().any(SymBoolExpr::contains_symbolic_hash)
    || constraints.iter().any(SymBoolExpr::contains_gasleft)
{
    return None;
}
```

Then collect `constants` only afterwards, since `fallback_candidates_for_var` is their first user. Keeps the hard-arith fast-bail intact while avoiding the extra scans on the vast majority of non-two-var queries.

crates/evm/symbolic/src/runtime/solver/hard_arith_fallback.rs
```diff
@@ -302,6 +302,120 @@ pub(crate) fn fallback_single_var_model(constraints: &[SymBoolExpr]) -> Option<S
...
+fn constraints_bind_each_search_var(
+    constraints: &[SymBoolExpr],
+    searched_vars: &SymbolicVars,
+) -> bool {
+    searched_vars.iter().all(|var| {
+        constraints.iter().any(|constraint| bool_expr_binds_single_var(constraint, *var))
+    })
+}
+
+fn bool_expr_binds_single_var(expr: &SymBoolExpr, bound_var: Symbol) -> bool {
```
grandizzy: `or` is represented as `Not(And(Not(a), Not(b)))`, and this recurses through both `Not` and `And` unconditionally, so an atom that only appears inside a disjunction can satisfy the gate. E.g. `x < y && (x < 10 || y < 10)` will count `x < 10` and `y < 10` as individual bindings even though the formula only requires one. It's still sound (the candidate gets validated), but it lets this path fire on cases that were meant to stay on SMT. Making the traversal polarity-aware — only descending into `And` in positive polarity, like `MaskHints::apply_bool` does — would keep it as narrow as intended:

```rust
fn bool_expr_binds_single_var(expr: &SymBoolExpr, bound_var: Symbol, inverted: bool) -> bool {
    match expr.kind() {
        SymBoolExprKind::Const(_) => false,
        SymBoolExprKind::Not(inner) => bool_expr_binds_single_var(inner, bound_var, !inverted),
        SymBoolExprKind::And(exprs) if !inverted => {
            exprs.iter().any(|e| bool_expr_binds_single_var(e, bound_var, false))
        }
        SymBoolExprKind::And(_) => false,
        SymBoolExprKind::Cmp(..) => { /* existing atom check */ }
    }
}
```

Same note applies to `bool_expr_has_two_var_relation` just above.

crates/evm/symbolic/src/tests.rs
```diff
@@ -3738,6 +3738,33 @@ fn is_sat_uses_single_var_witness_before_solver() {
     let _ = std::fs::remove_file(&marker);
 }
 
+#[cfg(unix)]
+#[test]
+fn is_sat_uses_two_var_witness_before_solver() {
```
grandizzy, line 3743: These new tests both cover the case where the witness is in the candidate set and no subprocess runs. Could you add a fallthrough test where the gates pass but the local search misses, so the solver subprocess is still invoked and the result cached/returned normally? That's the invariant that actually protects against false negatives silently changing results. A bounded additive relation whose solution isn't near constants±1 or powers of two (e.g. `x + y == K` with ranges chosen so boundary candidates don't satisfy it) should reliably miss the candidate set — assert one counted solver invocation.

grandizzy (commented):
Please check findings inline, no blocker.

mablr (dismissed):

mablr (commented):
lgtm

stevencartavia (approved):

mablr (approved):

PR #15604, opened 2026-07-06
refactor: centralize abi-only compilation

DaniPopes:
Centralizes ABI-only compilation in foundry_common so selectors, test discovery, and verification all share the same narrow helper path instead of constructing ABI-only projects inline. The target ABI helper keeps the current compiler-backed behavior while removing the duplicated setup around output selection, quiet compilation, and ABI extraction.

AI assistance was used to draft this change.


DaniPopes:
Superseded by #15610 with a clean helper-only branch.

PR #15605, opened 2026-07-06
feat(debugger): show external return values

figtracer:
## Summary
The Variables pane could already show external function parameters from calldata and internal function returns from decoded internal-call metadata, but external call return variables stayed `<unavailable>` because debugger nodes did not retain the call output. This adds the missing frame-level return data and gates decoding to frame end so stale EVM returndata is not shown mid-frame.
- carry each call's final output into `DebugNode` as `returndata`
- decode external function return variables in the Variables pane once the selected call frame reaches its final `RETURN`/`STOP` step
- keep internal decoded returns as the higher-priority path and selector-match external scopes before decoding output

Progresses #927.

### DIFF
--- a/crates/debugger/src/node.rs
+++ b/crates/debugger/src/node.rs
@@ -20,6 +20,9 @@ pub struct DebugNode {
     pub kind: CallKind,
     /// Calldata of the call.
     pub calldata: Bytes,
+    /// Return data produced by the call.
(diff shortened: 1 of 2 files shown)


grandizzy (approved):
lgtm

mattsse (approved):
lgtm

PR #15606, opened 2026-07-06
refactor(fuzz): introduce `FuzzRunArgs`

mablr:
## Description

Closes OSS-492

Refactors `forge fuzz run` to use a dedicated argument surface instead of exposing the full `forge test` CLI. The new surface keeps shared fuzz/invariant campaign dials like `--runs`, `--timeout`, `--depth`, and `--workers`, adapts them into `TestArgs`, and continues delegating through the existing test runner.

This also adds invocation-scope warnings when engine-specific flags cannot apply to any matched campaign, defaults invariant workers to auto for `forge fuzz run`, and keeps worker-specific stateless reproduction details out of the public `fuzz run` surface.

### Examples

```bash
forge fuzz run --list
```

```bash
forge fuzz run test/VaultCampaign.t.sol --runs 256 --depth 64
```

```bash
forge fuzz run --match-test '^invariant_' --runs 1000 --depth 128 --workers 4
```

```bash
forge fuzz run --match-contract VaultCampaignTest --runs 512 --depth 32 --dictionary-weight 70 --corpus-dir corpus/fuzz
```

```bash
forge fuzz run --match-test '^testFuzz' --runs 10000 --seed 0x1234
```

```bash
forge fuzz run --match-test '^invariant_' --timeout 300 --depth 64
```

### DIFF
--- a/crates/forge/src/args.rs
+++ b/crates/forge/src/args.rs
@@ -63,7 +63,6 @@ pub fn run_command(args: Forge) -> Result<()> {
             }
         }
         ForgeSubcommand::Fuzz(cmd) => {
-            cmd.reject_unsupported_flags()?;
             let silent = cmd.is_junit() || shell::is_json();
             let outcome = global.block_on(cmd.run())?;
             outcome.ensure_ok(silent)
--- a/crates/forge/src/cmd/fuzz.rs
+++ b/crates/forge/src/cmd/fuzz.rs
@@ -1,5 +1,5 @@
 use crate::{
-    cmd::test::{FilterArgs, FuzzMinimizeReplaySession, TestArgs},
+    cmd::test::{CampaignArgs, FilterArgs, FuzzMinimizeReplaySession, ShowmapDomainArg, TestArgs},
     multi_runner::{FuzzMinimizeEdgeIndices, FuzzMinimizeObservation, ShowmapConfig},
     result::TestOutcome,
 };
@@ -18,7 +18,7 @@ use foundry_common::{
     fs, sh_println, sh_status,
     shell::{OutputMode, Shell},
 };
-use foundry_config::Config;
+use foundry_config::{Config, filter::GlobMatcher};
 use foundry_evm::{
     executors::{CorpusDirEntry, ReplayObservation, ShowmapDomain, read_corpus_tree},
     fuzz::BasicTxDetails,
@@ -43,10 +43,7 @@ pub struct FuzzArgs {
 impl FuzzArgs {
     pub async fn run(self) -> Result<TestOutcome> {
         match self.command {
-            FuzzSubcommands::Run(mut args) => {
-                args.enable_fuzz_only();
-                args.run().await
-            }
+            FuzzSubcommands::Run(args) => TestArgs::from_fuzz_run(args).run().await,
             FuzzSubcommands::Replay(args) => args.run().await,
             FuzzSubcommands::Show(args) => {
                 args.run()?;
@@ -70,24 +67,13 @@ impl FuzzArgs {
             FuzzSubcommands::Show(_) | FuzzSubcommands::Cmin(_) | FuzzSubcommands::Tmin(_) => false,
         }
     }
-
-    pub fn reject_unsupported_flags(&self) -> Result<()> {
-        match &self.command {
-            FuzzSubcommands::Run(args) if args.is_watch() => {
-                bail!("`--watch` is not supported for `forge fuzz run`")
-            }
-            FuzzSubcommands::Replay(args) if args.is_watch() => {
-                bail!("`--watch` is not supported for `forge fuzz replay`")
-            }
-            _ => Ok(()),
-        }
-    }
 }
 
 #[derive(Clone, Debug, Subcommand)]
+#[allow(clippy::large_enum_variant)]
 pub enum FuzzSubcommands {
     /// Run only fuzz and invariant tests.
-    Run(TestArgs),
+    Run(FuzzRunArgs),
     /// Replay persisted fuzz failures, or corpus entries with `--corpus-dir`.
     Replay(FuzzReplayArgs),
     /// Print persisted corpus entries.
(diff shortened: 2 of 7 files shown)


crates/forge/src/cmd/test/mod.rs
```diff
@@ -358,8 +358,153 @@ fn sources_to_compile_from_artifacts(
...
+    #[arg(long, value_name = "DEPTH")]
+    pub min_depth: Option<u32>,
+
+    /// How invariant run depth is selected.
+    #[arg(long, value_name = "fixed|random")]
+    pub depth_mode: Option<InvariantDepthMode>,
+
+    /// Number of workers to use for invariant test campaigns, or `auto` to derive from `--jobs`.
+    #[arg(long, value_name = "WORKERS")]
+    pub workers: Option<InvariantWorkers>,
```
0xalpharush, line 395: why do we need this and jobs? anyway, `auto` shouldn't be required and be default

should set to `thread::available_parallelism()` like elsewhere

crates/forge/src/cmd/fuzz.rs
```diff
@@ -98,6 +84,122 @@ pub enum FuzzSubcommands {
...
+    )]
+    pub(crate) showmap_corpus_dir: Option<PathBuf>,
+
+    /// Run only the fuzz case at the given 1-based run index.
+    #[arg(long, env = "FOUNDRY_FUZZ_RUN", value_name = "RUN")]
+    pub(crate) fuzz_run: Option<u32>,
+
+    /// Run the fuzz case from the given worker. Requires `--fuzz-run`.
+    #[arg(long, env = "FOUNDRY_FUZZ_WORKER", value_name = "WORKER", requires = "fuzz_run")]
+    pub(crate) fuzz_worker: Option<u32>,
```
0xalpharush: why do we need the worker number to run a test case?

crates/forge/src/cmd/test/mod.rs
```diff
@@ -1008,6 +1165,143 @@ impl TestArgs {
...
+            invariant_corpus_random_sequence_weight: args.campaign.corpus_random_sequence_weight,
+            invariant_corpus_dir: args.campaign.corpus_dir,
+            invariant_payable_value_weight: args.campaign.payable_value_weight,
+            invariant_mutation_weight_splice: args.campaign.mutation_weight_splice,
+            invariant_mutation_weight_repeat: args.campaign.mutation_weight_repeat,
+            invariant_mutation_weight_interleave: args.campaign.mutation_weight_interleave,
+            invariant_mutation_weight_prefix: args.campaign.mutation_weight_prefix,
+            invariant_mutation_weight_suffix: args.campaign.mutation_weight_suffix,
+            invariant_mutation_weight_abi: args.campaign.mutation_weight_abi,
+            invariant_mutation_weight_cmp: args.campaign.mutation_weight_cmp,
```
0xalpharush: we should not have fuzz_* and invariant_*? they should just share

crates/forge/src/cmd/test/mod.rs
```diff
@@ -358,8 +358,153 @@ fn sources_to_compile_from_artifacts(
...
+            workers: self.workers.is_some(),
+            ..SuppliedEngineFlags::empty()
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct SuppliedEngineFlags {
+    pub(crate) frontier_dir: bool,
+    pub(crate) frontier_limit: bool,
```
0xalpharush: idk what this does should just remove and pick it up automatically

crates/forge/src/cmd/test/mod.rs
mablr: This was to avoid touching `TestArgs` for now, but if you think it's fine breaking it as well, I agree this would be cleaner.

crates/forge/src/cmd/test/mod.rs
0xalpharush: i think corpus_dir/symbolic plus maybe worker ID would work. follow existing fuzzer storage

mablr: @figtracer could you give more context on these please? or I can remove.

figtracer: > maybe worker ID would work

makes sense to refer to these per worker instead as troy said above

crates/forge/src/cmd/test/mod.rs
```diff
@@ -1008,6 +1163,136 @@ impl TestArgs {
...
+            path: args.path,
+            gas_report: args.gas_report,
+            allow_failure: args.allow_failure,
+            junit: args.junit,
+            fail_fast: args.fail_fast,
+            etherscan_api_key: args.etherscan_api_key,
+            list: args.list,
+            fuzz_seed: args.campaign.seed,
+            fuzz_runs: args.campaign.runs,
+            invariant_workers: Some(args.campaign.workers.unwrap_or(InvariantWorkers::Auto)),
```
stevencartavia: `forge fuzz run` silently ignores `[invariant] workers` from `foundry.toml`. This is `Some` even when `--workers` isn't passed, so the adapter's CLI provider overrides the profile-scoped toml value. With `[invariant] workers = 4`:

- `forge test` → 4 workers (honored)
- `forge fuzz run` → 1 worker (forced to auto)
- `forge fuzz run --workers 4` → 4 workers

`FOUNDRY_INVARIANT_WORKERS` still wins (figment global-profile precedence), so toml is overridden but env isn't — inconsistent. Suggest only emitting `invariant.workers` when `--workers` was supplied, and applying auto after config resolution when unconfigured.

crates/forge/src/args.rs
```diff
@@ -63,7 +63,6 @@ pub fn run_command(args: Forge) -> Result<()> {
             }
         }
         ForgeSubcommand::Fuzz(cmd) => {
```
stevencartavia, line 65: `forge fuzz run --watch` now errors via clap (fine), but `forge fuzz replay --watch` is accepted and does nothing: `FuzzReplayArgs` still flattens `TestArgs`→`WatchArgs`, and this arm never dispatches to watch mode. `forge test --watch` blocks and watches; `forge fuzz replay --watch` exits 0 immediately. Previously this errored with `` `--watch` is not supported for `forge fuzz replay` ``. Suggest restoring a replay-specific rejection (or not flattening `WatchArgs` into `FuzzReplayArgs`) and re-adding the deleted test.

crates/forge/src/cmd/test/mod.rs
```diff
@@ -1008,6 +1163,136 @@ impl TestArgs {
...
+            fuzz_frontier_limit: args.campaign.frontier_limit,
+            fuzz_payable_value_weight: args.campaign.payable_value_weight,
+            fuzz_mutation_weight_splice: args.campaign.mutation_weight_splice,
+            fuzz_mutation_weight_repeat: args.campaign.mutation_weight_repeat,
+            fuzz_mutation_weight_interleave: args.campaign.mutation_weight_interleave,
+            fuzz_mutation_weight_prefix: args.campaign.mutation_weight_prefix,
+            fuzz_mutation_weight_suffix: args.campaign.mutation_weight_suffix,
+            fuzz_mutation_weight_abi: args.campaign.mutation_weight_abi,
+            fuzz_mutation_weight_cmp: args.campaign.mutation_weight_cmp,
+            fuzz_input_file: args.fuzz_input_file,
```
stevencartavia, line 1223: `--fuzz-input-file` is fuzz-only (mapped only into the fuzz config) but is missing from the invariant-only warning path. On an invariant-only match, `--frontier-dir` warns (`...only applies to fuzz tests...`) while `--fuzz-input-file` is silent. Suggest adding `fuzz_input_file` to `SuppliedEngineFlags`, setting it in `from_fuzz_run`, and warning in the `counts.fuzz == 0 && counts.invariant > 0` branch.

crates/forge/src/args.rs
mablr: https://github.com/foundry-rs/foundry/pull/15606/commits/d3527fa77b24a3aebbb830c7980bbc2afade0995

crates/forge/src/cmd/test/mod.rs
mablr: Addressed in https://github.com/foundry-rs/foundry/pull/15606/commits/e8ebb5a4a70da236c8ee96a483ecb39dcb49998a

crates/forge/src/cmd/test/mod.rs
mablr: Addressed in https://github.com/foundry-rs/foundry/pull/15606/commits/fd08596f1353fb4e53e2629b6fb6cc4c3010cdb4

stevencartavia (approved):

figtracer (approved):

PR #15607, opened 2026-07-06
feat(cast): auto-detect network in decode-transaction

stevencartavia:
`cast decode-transaction` now infers the network from the transaction type byte, so Tempo transactions (type `0x76`) decode without passing `--network tempo`. The flag still works as an explicit override.


mablr (approved):
Hacky but handy, sgtm.

mattsse (approved):

PR #15608, opened 2026-07-06
feat(cast): auto-switch browser wallet chain

### LINKED ISSUE #14641: feat(cast): auto-switch chain via `wallet_switchEthereumChain` when `--browser` is used
### Component

Cast

When `cast send --browser` (or any `--browser` flow) is invoked against a chain that doesn't match the connected browser wallet's current chain, the signer errors out with `Transaction `chainId` does not match connected wallet chain ID` and the user has to manually switch their wallet's network and rerun the command.

The non-browser `--unlocked` path already handles this — it calls `wallet_switchEthereumChain` when `config.chain` differs from the RPC's reported chain (introduced in #5077). That logic is currently gated to `unlocked && browser.is_none()` and skipped entirely on the browser path.

#### Where the gap lives

`crates/cast/src/cmd/send.rs` (foundry-rs/foundry, master HEAD):

```rust
// Case 1: --unlocked — switches chain
if unlocked && browser.is_none() {
    if let Some(config_chain) = config.chain {
        let current_chain_id = provider.get_chain_id().await?;
        let config_chain_id = config_chain.id();
        if config_chain_id != current_chain_id {
            sh_warn!("Switching to chain {}", config_chain)?;
            provider.raw_request::<_, ()>(
                "wallet_switchEthereumChain".into(),
                [serde_json::json!({ "chainId": format!("0x{:x}", config_chain_id) })],
            ).await?;
        }
    }
    // ...
}
// Case 2: --browser — does NOT switch chain
} else if let Some(browser) = browser {
    let chain = builder.chain(); // only used for is_tempo() gas buffer, never for switching
    // ...
    let tx_hash = browser.send_transaction_via_browser(tx_request).await?;
}
```

`crates/wallets/src/wallet_browser/signer.rs` (foundry-rs/foundry-core, main HEAD) detects the mismatch but only fails:

```rust
if let Some(chain_id) = tx_request.chain_id()
    && chain_id != self.chain_id
{
    return Err(alloy_signer::Error::other(
        "Transaction `chainId` does not match connected wallet chain ID",
    ));
}
```

#### Motivation

The browser wallet flow is the recommended signing path for agentic workflows where private keys must not touch the terminal. In our case (a Sablier [vesting-withdrawal skill](https://github.com/sablier-labs/sablier-skills/blob/2ac52cc2c566eef3591116108a364d0a9f012c0f/skills/sablier-withdraw-vesting/SKILL.md) that batches `withdrawMultiple` per Lockup contract), the agent already knows the target chain — it resolves the chain from the indexer and passes `--rpc-url` and `--chain` accordingly. The wallet state is the only thing not under the agent's control, so a chain mismatch surfaces as a runtime error mid-batch, requiring a manual extension click and a retry.

The browser wallet bridge already has the connected `chain_id` (POSTed to `/api/connection` during connect — see `foundry-rs/foundry-browser-wallet/src/App.tsx` and the `Connection { address, chain_id }` struct in foundry-core). And EIP-3326 (`wallet_switchEthereumChain`) is universally supported by MetaMask, Rabby, Frame, Coinbase Wallet, etc. The pieces are all there; they're just not wired together on the `--browser` path.

Symmetrically, `cast wallet address --browser` does not surface chainId either, so callers can't even pre-flight-check before launching the bridge.

#### Proposed solution

Mirror Case 1's logic in Case 2 of `crates/cast/src/cmd/send.rs` — before `browser.send_transaction_via_browser(tx_request)`, if `builder.chain()` differs from `browser.chain_id()`, request a switch via the bridge. The bridge would forward the `wallet_switchEthereumChain` RPC to the connected EIP-1193 provider, exactly as a dapp would.

A first-pass sketch:

```rust
} else if let Some(browser) = browser {
    let chain = builder.chain();
    if chain.id() != browser.chain_id() {
        sh_warn!("Switching browser wallet to chain {}", chain)?;
        browser.switch_chain(chain.id()).await?;
    }
    // ...existing code
}
```

This requires a small new method on `BrowserSigner` that pushes a `wallet_switchEthereumChain` request through the same queue that handles `eth_sendTransaction`, plus a bit of UI in `foundry-browser-wallet` to surface the prompt.

Happy to put up the PR if the maintainers are open to it — flagging the issue first so we can agree on the UX (e.g., should it be opt-in via a flag, or always-on with a warn-and-switch like the `--unlocked` path).

### Additional context

- Foundry version: `cast 1.5.1-stable` (Build 2025-12-22)
- OS: macOS Tahoe v26.4.1
- Prior art: #5077 (the `--unlocked` chain-switch logic, merged 2023-06-01)
- Related: #13881 (a separate `cast send --browser` chain-handling bug on Tempo)
- The browser wallet repo (`foundry-rs/foundry-browser-wallet`) currently doesn't surface any switch-chain UI — issues #16, #17, #19, #20 there are unrelated

mablr:
Adds chain switching to `cast send --browser` so browser-wallet sends can target a configured chain even when the connected wallet starts on a different network. After transaction construction, sponsorship, and fee-token resolution succeed, Cast asks the browser wallet to switch to the transaction chain before submitting the transaction through the wallet.

This depends on foundry-rs/foundry-core#111 for the `foundry-wallets` chain-switch request/response protocol and foundry-rs/foundry-browser-wallet#90 for the browser-side `wallet_switchEthereumChain` handling. The `foundry-wallets` dependency is pinned to the pushed `foundry-core` commit containing that protocol.

Closes #14641.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5901,7 +5901,7 @@ dependencies = [
 [[package]]
 name = "foundry-wallets"
 version = "0.1.0"
-source = "git+https://github.com/foundry-rs/foundry-core?rev=904f2f6f12196831fa62cc49aca27bcc78c48f98#904f2f6f12196831fa62cc49aca27bcc78c48f98"
+source = "git+https://github.com/foundry-rs/foundry-core?rev=7808f461e26b2ec980dbd60804c7def861baca7a#7808f461e26b2ec980dbd60804c7def861baca7a"
 dependencies = [
  "alloy-consensus",
  "alloy-dyn-abi",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5901,7 +5901,7 @@ dependencies = [
 [[package]]
 name = "foundry-wallets"
 version = "0.1.0"
-source = "git+https://github.com/foundry-rs/foundry-core?rev=7808f461e26b2ec980dbd60804c7def861baca7a#7808f461e26b2ec980dbd60804c7def861baca7a"
+source = "git+https://github.com/foundry-rs/foundry-core?rev=8cb3bf1f5452ca872fd765c5f51a3074ee511734#8cb3bf1f5452ca872fd765c5f51a3074ee511734"
 dependencies = [
  "alloy-consensus",
  "alloy-dyn-abi",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5901,7 +5901,7 @@ dependencies = [
 [[package]]
 name = "foundry-wallets"
 version = "0.1.0"
-source = "git+https://github.com/foundry-rs/foundry-core?rev=7610c7e6f88429caa400ec35606afd28270ecf9f#7610c7e6f88429caa400ec35606afd28270ecf9f"
+source = "git+https://github.com/foundry-rs/foundry-core?rev=904f2f6f12196831fa62cc49aca27bcc78c48f98#904f2f6f12196831fa62cc49aca27bcc78c48f98"
 dependencies = [
  "alloy-consensus",
  "alloy-dyn-abi",
@@ -5924,7 +5924,9 @@ dependencies = [
  "dirs",
  "eth-keystore",
  "eyre",
+ "hidapi-rusb",
  "rpassword",
+ "rusb",
  "rusqlite",
  "serde",
  "serde_json",
@@ -5934,7 +5936,7 @@ dependencies = [
  "tokio",
  "toml 1.1.2+spec-1.1.0",
  "tower",
- "tower-http 0.6.11",
+ "tower-http 0.7.0",
  "tracing",
  "uuid 1.23.4",
  "webbrowser",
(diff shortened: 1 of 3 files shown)


mablr (commented):
Marked as blocked until https://github.com/foundry-rs/foundry-browser-wallet/pull/90 and  https://github.com/foundry-rs/foundry-core/pull/111 are merged.

mattsse (dismissed):
LGTM

stevencartavia (dismissed):

grandizzy (dismissed):
lgtm

mablr (commented):
Bumped `foundry-wallet` to https://github.com/foundry-rs/foundry-core/commit/8cb3bf1f5452ca872fd765c5f51a3074ee511734, now G2G.

grandizzy (approved):

figtracer (approved):

PR #15609, opened 2026-07-06
refactor(cast): decode-transaction via FoundryTxEnvelope

stevencartavia:
Routes the no-flag `cast decode-transaction` path through `FoundryNetwork`/`FoundryTxEnvelope`, which already dispatches on the EIP-2718 type byte. Removes the duplicate `NetworkVariant::from_tx_type` table added in #15607. Same behavior; `--network` still works as an explicit override.


mattsse (approved):
lgtm

grandizzy (approved):
makes sense

PR #15616, opened 2026-07-07
feat(traces): decode T7 storage credits precompile

stevencartavia:
The trace decoder labeled the T7 Storage Credits precompile but never registered its ABI, so its calls and reverts showed as raw selectors. This registers `IStorageCredits` so they decode like every other Tempo precompile.

Chained ahead of `ITIP20` to avoid the shared `balanceOf(address)` selector clobbering TIP20's `uint256` return.


mattsse (approved):
lgtm

grandizzy (approved):
lgtm

PR #15617, opened 2026-07-07
feat(anvil): apply Tempo base fee rules

stevencartavia:
Anvil computed every chain's next-block base fee with generic EIP-1559, so a Tempo node never reproduced Tempo's own rules: a fixed base fee pre-T7 and the TIP-1067 dynamic controller at T7+.

The fee manager now mirrors Tempo's chainspec — fixed pre-T7, and `tempo_t7_next_block_base_fee` (clamped to `[floor, cap]`) at T7+ — including on forks, `anvil_reset`, and `eth_simulateV1`. On a T7 chain, genesis keeps the 20B seed and block 1 already clamps to the cap, matching a real Tempo node. Non-Tempo chains are unchanged.

### DIFF
--- a/crates/anvil/src/config.rs
+++ b/crates/anvil/src/config.rs
@@ -1197,6 +1197,10 @@ impl NodeConfig {
         let base_fee_params: BaseFeeParams =
             self.networks.base_fee_params(self.get_genesis_timestamp());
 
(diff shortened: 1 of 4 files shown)


grandizzy (approved):
lgtm!

mablr (approved):

PR #15620, opened 2026-07-07
perf(cast): speed up local interface generation

figtracer:
## Summary
- compile local `cast interface` targets with ABI-only output
- avoid writing artifacts for this read-only interface generation path
- add CLI coverage that local interface generation leaves artifact paths untouched

## Results
Benchmarked on a 3000-function/event/error Solidity fixture.

| Scenario | Before | After | Speedup |
| --- | ---: | ---: | ---: |
| cold local source | 1.092s | 326.6ms | 3.34x |
| after full build in prepare | 1.072s | 321.1ms | 3.34x |

### DIFF
--- a/crates/cast/src/cmd/interface.rs
+++ b/crates/cast/src/cmd/interface.rs
@@ -9,7 +9,7 @@ use foundry_cli::{
 };
 use foundry_common::{
     ContractsByArtifact,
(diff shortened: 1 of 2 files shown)


grandizzy (approved):
makes sense

mablr (approved):

PR #15624, opened 2026-07-07
perf(forge): speed up selector analysis

figtracer:
## Summary
- request only method identifiers for `forge selectors collision` and skip artifact writes when build-info is not requested
- skip artifact persistence for ABI-only selector commands loaded through `ProjectPathOpts`
- add regressions for collision/list selector commands not creating artifacts

## Benchmarks
Generated selector collision fixture with 4,000 methods:
- master: 1.499 s ± 0.017 s
- candidate: 301.6 ms ± 10.0 ms
- 4.97x faster

Generated `selectors find` fixture with 20,000 functions:
- master: 609.4 ms ± 78.3 ms, writes 2.8 MB under `out/`
- candidate: 527.0 ms ± 9.9 ms, writes no artifacts
- 1.16x faster

### DIFF
--- a/crates/forge/src/cmd/selectors.rs
+++ b/crates/forge/src/cmd/selectors.rs
@@ -11,7 +11,12 @@ use foundry_common::{
     selectors::{SelectorImportData, import_selectors},
     shell,
 };
-use foundry_compilers::{Project, info::ContractInfo, multi::MultiCompiler};
+use foundry_compilers::{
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

grandizzy (approved):
lgtm

PR #15626, opened 2026-07-07
test(forge): cover showmap worker corpus merge

mablr:
## Description

Adds regression coverage for `forge test --showmap-out` replaying generated stateless fuzz corpus roots with unsynced parallel worker directories. The test builds separate `worker*/corpus` entries, verifies per-input showmap output sees every worker entry, then checks aggregate showmap output equals the summed per-input coverage.

Closes OSS-424.


mattsse (approved):
lgtm

figtracer (approved):
makes sense

PR #15628, opened 2026-07-07
fix(forge): validate mutation compiler profile

mablr:
Mutation testing now validates the unmutated project under the same compiler profile that will be used for mutants. This catches cases where `--mutation-via-ir` or `--mutation-optimizer-runs` makes the baseline fail before any source mutation is applied, and reports a mutation-specific setup error instead of surfacing a raw compiler failure later in the run.

The mutation runner now builds the selected sources with the mutation compiler overrides, runs an unmutated baseline against that output, and only then starts executing mutants. The mutation result cache and mutant execution also use that mutation-profile compile output, keeping setup and execution aligned.


mattsse (dismissed):
lgtm

crates/forge/src/cmd/test/mod.rs
```diff
@@ -2039,12 +2039,85 @@ impl TestArgs {
...
+                        config_for_mutation.clone(),
+                        evm_opts_for_mutation.clone(),
+                        &mutation_output,
+                        filter,
+                        mutation_execution,
+                    )
+                    .await
+                }
+            }
+            .wrap_err("Mutation testing compiler profile failed its unmutated baseline run")?;
```
stevencartavia: For `--mutate`, the baseline test suite runs twice before mutants. Comparing master (`a6620cc7`) and this branch (`b7e6fa7`) with `RUST_LOG=forge::test=trace` on a project where both profiles compile (`forge test --mutate src/Adder.sol --mutation-optimizer-runs 1234 --mutation-jobs 1`): master emits `running all tests` once, this branch emits it twice (two `Suite result: ok`) before `Running mutation tests...`. The pre-existing default-profile `dispatch_network` run still executes; its result is only used for the old failed/zero-passing gate and is then overwritten at the end of the mutate block (`outcome = TestOutcome::empty(...)`), and this adds a new mutation-profile baseline on top of it. Since `run_mutation_testing` now takes `&mutation_output`, the default-profile output no longer feeds mutation execution, artifact selection/build IDs, mutation cache keys, or the new validation. Consider branching `--mutate` before the default-profile compile/run (in `compile_project`), so there's a single baseline under the mutation profile and the extra default compile is avoided — skipping only `dispatch_network` wouldn't avoid the compile, since `compile_project` already compiled by then.

crates/forge/src/cmd/test/mod.rs
```diff
@@ -2039,12 +2039,85 @@ impl TestArgs {
...
+                None,
                 execution.replay_symbolic_artifact.as_ref(),
             )?;
+            let mutation_output = ProjectCompiler::new()
+                .dynamic_test_linking(config_for_mutation.dynamic_test_linking)
+                .quiet(json_output || self.junit)
+                .files(selected_sources.clone())
+                .compile(&config_for_mutation.project()?)
+                .wrap_err(
+                    "Mutation testing compiler profile failed to compile before applying mutations",
```
stevencartavia: `--mutation-via-ir` / `--mutation-optimizer-runs` can't rescue a project that fails to compile under the default profile. On a project with `[profile.default] via_ir=false` and a stack-too-deep contract: `forge build --via-ir` succeeds (exit 0), but `forge test --mutate src/StackTooDeep.sol --mutation-via-ir true` fails with a generic `Compiler run failed: Stack too deep` (not this new `failed to compile before applying mutations` wrap) — on both master and this branch. `compile_project` compiles under the default compiler settings (aside from mutation's dynamic-linking/cache plumbing) and bails before `run_tests`; it never applies `apply_mutation_compiler_overrides`, which happens later in the mutate block. So this handles the "mutation profile makes it worse" case but not the "mutation profile makes it compile" case. Both this and the double-baseline resolve with the same change: route `--mutate` before the default-profile compile/normal dispatch, apply the mutation overrides first, and make the mutation-profile compile the only front-door baseline before entering the orchestrator.

stevencartavia (dismissed):

figtracer (dismissed):

stevencartavia (dismissed):

grandizzy (approved):
lgtm

figtracer (approved):

PR #15629, opened 2026-07-07
fix(forge): keep mutation config after reload

mablr:
Mutation mode requires dynamic test linking and cache usage before the test project is built. If missing dependency installation triggers a config reload for auto-detected remappings, applying those settings before the reload can lose them.

This applies the mutation-required config overrides after the dependency-install reload path has completed, keeping fresh dependency installs consistent with normal mutation setup without duplicating the override logic.


mattsse (approved):
lgtm

figtracer (approved):

stevencartavia (approved):

PR #15630, opened 2026-07-07
fix(forge): preserve mutation workspace paths

mablr:
Mutation workspaces now preserve external read-only compiler paths instead of trying to copy them into each temporary project. This lets monorepo-style projects with remappings or allow paths pointing at sibling Solidity directories run mutation testing without every mutant becoming invalid during workspace setup.

The workspace rebasing logic now resolves project-relative paths against the original root, rebases project-local paths into the temp workspace, and keeps external paths external. Project-local fuzz and invariant corpus/frontier directories are also copied and rebased into each mutant workspace so mutation runs do not share those mutable paths with the original project. Paths already covered by handled dependency roots are skipped to avoid interfering with dependency symlinks.


crates/forge/src/workspace.rs
```diff
@@ -415,6 +480,36 @@ fn create_project_local_permission_dir(
...
+fn copy_project_local_optional_path(
+    root: &Path,
+    temp_dir: &Path,
+    path: &Option<PathBuf>,
+    handled_roots: &[PathBuf],
+) -> Result<()> {
+    let Some(path) = path else { return Ok(()) };
+    let resolved = resolve_against_root(root, path);
+    let rel = relative_to_root(root, &resolved);
+    if rel.is_absolute() || rel.as_os_str().is_empty() || !resolved.exists() {
```
stevencartavia: External `corpus_dir`/`frontier_dir` (paths outside the project root, e.g. `corpus_dir = "../shared-corpus"`) aren't isolated per mutant. `rebase_project_path` keeps them external and `copy_project_local_optional_path` skips them (`rel.is_absolute()`), but these paths are mutable during fuzzing: corpus seeds/dirs are created under `corpus_dir` via `persist_corpus_seed`/`WorkerCorpus`, and frontiers via `write_branch_frontiers`/`write_frontier_artifact`. Result: every mutant reads and writes the same external corpus, polluting the user's real corpus and sharing mutable state across concurrent mutants — contradicting the isolation rationale for mutable corpus/frontier paths.

Repro: `[invariant] corpus_dir = "../shared-corpus"`, pre-seed a file, run `forge test --mutate --mutation-jobs 2` → the pre-seeded file survives (no isolated copy) and a new `../shared-corpus/.../worker0/corpus/<uuid>.json` is written by a mutant run. With a project-local `corpus_dir`, the original is left unchanged, so only the external case leaks. Consider copying/rebasing external corpus/frontier to a per-mutant temp dir, or rejecting them under `--mutate`.

crates/forge/src/workspace.rs
```diff
@@ -415,6 +480,36 @@ fn create_project_local_permission_dir(
...
+        return Ok(());
+    }
+    if is_covered_by_handled_root(&rel, handled_roots) {
```
stevencartavia: A corpus/frontier path under a symlinked handled root leaks writes back to the original project. `copy_project_local_optional_path` never materializes an isolated temp copy for such a path — if it already exists it's skipped as covered by a handled root (`libs`/`node_modules`/`dependencies`), and if absent it's skipped via `!resolved.exists()`. Those roots are then symlinked into the temp workspace (when `symlink_dir` succeeds), so `corpus_dir = "lib/mycorpus"` rebases to `temp/lib/mycorpus`, which resolves through the symlink to the original `lib`.

Repro: `[invariant] corpus_dir = "lib/mycorpus"` (original `lib/mycorpus` absent), run `forge test --mutate` → original `lib/mycorpus/.../worker0/corpus/<uuid>.json` gets created. Contrived config, low severity — same root cause as the external-corpus case (mutable paths not fully isolated).

crates/forge/tests/cli/test_cmd/mutation.rs
```diff
@@ -970,6 +971,79 @@ contract UsesSharedTest {
     );
 });
 
+forgetest_init!(mutation_workspace_preserves_external_read_only_remappings, |prj, cmd| {
```
stevencartavia, line 1231: Project-local remapping targets that aren't otherwise copied by `copy_project` — i.e. not under `src`/`test`/`script`, configured libs, `node_modules`/`dependencies`, `include_paths`/`allow_paths`, or readable fs-permission paths — are rebased into the temp workspace but never copied. `rebase_remapping` rewrites the target to `temp/...`, but `copy_project` doesn't copy arbitrary remapping targets, so any mutant whose compilation needs that target becomes invalid.

Repro: `remappings = ["@shared/=packages/shared/src/"]` with `packages/shared/src/Math.sol` inside root. Baseline `forge test` passes; `forge test --mutate` → all mutants invalid with `.../forge_mutation_XXX/packages/shared/src/Math.sol: No such file or directory`. Adding `include_paths = ["packages/shared/src"]` fixes it. This is pre-existing (predates this PR), but it's in the same monorepo/path-preservation area and this new test only covers the external-remapping variant.

crates/forge/src/workspace.rs
```diff
@@ -336,8 +398,11 @@ fn copy_extra_project_path(
     handled_roots: &[PathBuf],
     label: &str,
 ) -> Result<()> {
-    let resolved = if path.is_absolute() { path.to_path_buf() } else { root.join(path) };
+    let resolved = resolve_against_root(root, path);
     let rel = relative_to_root(root, &resolved);
+    if rel.is_absolute() {
```
stevencartavia, line 467: Related gap on this same external-path handling: external `libs` (e.g. `libs = ["lib", "../shared-lib"]` pointing at a sibling dir in a monorepo) are hard-rejected during mutation-workspace setup, even though this PR now preserves external `include_paths`/`allow_paths`/`remappings` (here in `copy_extra_project_path`). `handled_project_roots` → `push_handled_project_root` runs every existing lib through `ensure_safe_relative_path`, which bails for any outside-root path (the later lib copy loop does the same). So `copy_project` fails outright and every mutant is invalid — the exact monorepo scenario this PR targets.

Repro: `libs = ["lib", "../shared-lib"]` with an imported dependency in the sibling `../shared-lib`. Baseline `forge test` passes; `forge test --mutate` → all mutants invalid with `Failed to copy project: requires lib directory under project root, got: .../shared-lib`. Fix: classify libs the same way as include/allow/remappings — leave external libs external (don't copy/symlink them, keep the rebased config pointing at the original path), reject/copy only project-local ones.

crates/forge/src/workspace.rs
```diff
@@ -415,6 +535,45 @@ fn create_project_local_permission_dir(
...
+    handled_roots: &[PathBuf],
+) -> Result<()> {
+    let Some(path) = path else { return Ok(()) };
+    let resolved = resolve_against_root(root, path);
+    let rel = relative_to_root(root, &resolved);
+    if rel.as_os_str().is_empty() || !resolved.exists() {
+        return Ok(());
+    }
+    let is_external = rel.is_absolute();
+    let target = if is_external || is_covered_by_handled_root(&rel, handled_roots) {
```
mablr, line 560: This uses the broader handled-roots set, so a corpus under `src`/`test`/`script` can get copied into `.foundry_mutable` even though `rebase_mutable_project_path` still points at the normal temp copy. Correctness looks fine; I'll handle the classifier cleanup separately as a follow-up.

mattsse (commented):
One more mutable workspace path looks like it should be isolated before this lands.

`failure_persist_dir` is mutable too, so preserving external values can still let isolated mutation/brutalize workspaces write into the user's real external failure cache. Fuzz writes counterexamples under `test_paths(... failure_persist_dir ...)`, and invariant writes under `invariant_failure_dir` / `record_invariant_failure`, so a config like `failure_persist_dir = "../shared-failures"` can escape the temp workspace.

I think these two fields should use the same mutable-path helper as corpus/frontier:

```rust
if let Some(path) = &config.fuzz.failure_persist_dir {
    temp_config.fuzz.failure_persist_dir =
        Some(rebase_mutable_project_path(config, temp_path, path));
}

if let Some(path) = &config.invariant.failure_persist_dir {
    temp_config.invariant.failure_persist_dir =
        Some(rebase_mutable_project_path(config, temp_path, path));
}
```

Test suggestion: add a workspace unit test with `fuzz.failure_persist_dir = Some(PathBuf::from("../shared-fuzz-failures"))` and `invariant.failure_persist_dir = Some(PathBuf::from("../shared-invariant-failures"))`, then assert both rebased paths start with `workspace.join(".foundry_mutable")`. I would keep this as rebase-only for the minimal fix; project-local failure dirs are not copied today, so this closes the write escape without changing persisted-failure replay semantics more broadly.

crates/forge/src/workspace.rs
```diff
@@ -33,6 +33,26 @@ pub fn relative_to_root(root: &Path, path: &Path) -> PathBuf {
...
+    if path.is_absolute() { normalize_path(path) } else { normalize_path(&root.join(path)) }
+}
+
+fn normalize_path(path: &Path) -> PathBuf {
+    let mut normalized = PathBuf::new();
+    for component in path.components() {
+        match component {
+            Component::CurDir => {}
+            Component::ParentDir => {
+                normalized.pop();
```
grandizzy, line 50: **High:** Lexically removing `..` changes filesystem semantics when the preceding component is a symlink. For example, with `project/link -> /tmp/shared/pkg`, `link/../include` resolves to `/tmp/shared/include`, but this normalizes it to `project/include`. The rebased config then points at a nonexistent temp path, making mutant or brutalize compilation fail.

A similar issue affects mutable paths under externally targeted dependency symlinks: they are classified as project-local and subsequently rejected by the canonical containment check.

Please classify paths using the canonical root and canonicalized deepest existing ancestor rather than removing `..` lexically.

crates/forge/src/workspace.rs
```diff
@@ -415,6 +535,45 @@ fn create_project_local_permission_dir(
...
+    }
+    let is_external = rel.is_absolute();
+    let target = if is_external || is_covered_by_handled_root(&rel, handled_roots) {
+        temp_dir.join(isolated_mutable_path_rel(&resolved))
+    } else {
+        ensure_safe_relative_path(&rel, "corpus/frontier", path)?;
+        ensure_within_root(root, &resolved, "corpus/frontier", path)?;
+        temp_dir.join(rel)
+    };
+    if target.exists() {
```
grandizzy: **Medium:** Returning whenever `target` exists can leave nested mutable paths only partially copied. For example, with `corpus_dir = "state/corpus"` and `frontier_dir = "state"`, copying the corpus creates `temp/state`; the frontier copy then returns here, so `state/frontier.json` is missing even though the rebased frontier points at `temp/state`.

Existing destination directories should be merged recursively rather than skipped wholesale.

crates/forge/src/workspace.rs
```diff
@@ -108,8 +144,36 @@ pub fn rebase_config_paths(config: &Config, temp_path: &Path) -> Config {
...
+    if rel.is_absolute() || is_covered_by_symlinked_project_root(config, &rel) {
+        return temp_path.join(isolated_mutable_path_rel(&resolved));
+    }
+    temp_path.join(rel)
+}
+
+fn isolated_mutable_path_rel(resolved: &Path) -> PathBuf {
+    let mut rel = PathBuf::from(".foundry_mutable");
+    for component in resolved.components() {
+        if let Component::Normal(component) = component {
```
grandizzy: **Medium, Windows-only:** Dropping `Component::Prefix` makes paths on different drives or UNC shares collide. For example, `C:\shared\state` and `D:\shared\state` both become `.foundry_mutable/shared/state`; the second tree is then skipped because the destination exists.

Please include an encoded drive/UNC prefix or a stable hash of the complete source path in the isolated destination.

grandizzy (changes requested):
Found three path-handling edge cases that can cause invalid mutation/brutalize runs or mixed mutable state. Details inline.

grandizzy (dismissed):
lgtm

crates/forge/src/workspace.rs
```diff
@@ -84,11 +108,27 @@ pub fn rebase_config_paths(config: &Config, temp_path: &Path) -> Config {
 
     if let Some(path) = &config.fuzz.failure_persist_dir {
         temp_config.fuzz.failure_persist_dir =
-            Some(rebase_project_path(&config.root, temp_path, path));
+            Some(rebase_mutable_project_path(config, temp_path, path));
```
stevencartavia, line 111: `failure_persist_dir` is rebased into `.foundry_mutable`, but its existing counterexamples are not copied there. This can produce false survivors.

I reproduced this with an absolute external fuzz failure directory: before this PR, the persisted failure killed 10 mutants; on this PR, the same run killed 0 and behaved exactly as if the failure file were absent.

Could we copy existing persisted failures into the isolated directory before running each mutant?

figtracer (approved):

grandizzy (approved):

PR #15631, opened 2026-07-07
feat(fuzz): reuse coverage-winning whole calls

figtracer:
This retargets #15631 away from automatic symbolic assistance. The solver
prototype did generate cold-start seeds on selected linear fixtures, but its
automatic A/B campaigns did not retain a final coverage advantage: Nerite and
Aave controls caught up within 15 minutes, and Liquity ended a one-hour run with
the same 2,755 PC offsets despite one useful seed. The later long-plateau policy
ran seven assists after roughly 7.6 million concrete runs across Tomb, Spark
PSM, and Aave KeyValueList but produced no candidate, useful corpus entry, or
counterexample: four attempts ended with no input, two reverted on every input,
and one timed out. That evidence, along with the solver lifecycle and resource
cost, did not justify making it part of ordinary fuzz campaigns.

Stacked on #15816, this instead follows Echidna's direct generation primitive: a [concretely executed top-level call is retained only when its execution grows coverage](https://github.com/crytic/echidna/blob/58d7fc5860de5a7a6e5a3f46384f3e9d9a486596/lib/Echidna/Campaign.hs#L569-L584), in a [signature-keyed whole-call dictionary](https://github.com/crytic/echidna/blob/58d7fc5860de5a7a6e5a3f46384f3e9d9a486596/lib/Echidna/ABI.hs#L109-L137), and a reused call has [one ABI argument mutated](https://github.com/crytic/echidna/blob/58d7fc5860de5a7a6e5a3f46384f3e9d9a486596/lib/Echidna/ABI.hs#L349-L356). The #15816 stack inherits #15220, which hoists recorded nested calls into the invariant corpus and seeds it from sibling zero-input test traces; #15817 moves coverage-winning observed calls to a separate generation pool. This PR intentionally removes both seed paths, the nested-call recorder and depth bookkeeping, and the public `ObservedCall`/`RawCallResult.observed_calls` surface.

Although the inherited recorder retains caller and value, replaying a nested frame as a top-level transaction cannot reproduce the surrounding call frame, reentrancy context, or preceding sequence state. This design instead preserves only ABI-argument correlation from calls with unambiguous top-level coverage attribution. Sender, value, warp, and roll are generated normally, and a donor can still depend on earlier state in its original invariant sequence; reuse is a heuristic, not full-call reproduction.

This PR instead retains argument-bearing top-level ABI calldata whose complete concrete execution first raises the worker's edge or hit-count map. Donors are indexed by the full canonical-signature hash. For each freshly generated invariant call, Foundry still selects the destination, function, sender, and value normally; it may then retarget compatible donor calldata to that destination and mutate exactly one ABI argument. The call executes normally, and only sequence-level inputs satisfying the existing coverage or optimization admission rules are persisted.

Absent explicit corpus paths, an ordinary `forge fuzz run` creates and reuses separate `cache/fuzz/corpus` and `cache/invariant/corpus` roots automatically; a configured `failure_persist_dir` derives `<dir>/corpus`. Ordinary `forge test` remains opt-in through configured corpus paths. There is no `forge fuzz seed` subcommand. This donor mechanism adds no solver, hybrid engine, extra worker, or new mode flag; solver pre-seeding remains explicit through `forge test --symbolic-seed-corpus` or the two-step `--fuzz-frontier-dir`/`--symbolic-use-fuzz-frontiers` workflow.

Startup replay reconstructs donors from top-level calls that win coverage during replay. Parallel workers clone that seed and then learn locally. Each worker and the retained startup seed are separately bounded to 1,024 donors and 4 MiB of calldata, so the conservative campaign-level calldata bound is `(workers + 1) × 4 MiB` plus container overhead. `dictionary_weight = 0` disables donor selection; an effective `mutation_weight_abi = 0` disables collection and reuse. The dictionary weight is shared with Foundry's existing value dictionary rather than introducing another knob. Keeping dictionaries worker-local avoids shared hot-path synchronization, but discoveries are not live-shared between parallel workers; that is an explicit parallel-efficiency limitation.

### Results

The Sky rows used the earlier donor-only build `def3ea2f4dc2` against #15816 so
both arms still had the inherited nested-call recorder; they isolate the donor
mechanism but are not a benchmark of the final composite diff. The deterministic
fixture and Superform campaign use benchmarked implementation commit
`1303932c9b4c`; the later `a377bf2c8fc9` commit changes documentation only.
The cold real-world A/B arms use baseline #15816 commit
`926d4e36fd160aaaf5f34b2e085a992e3e933c9a`, pinned targets, seed
`0x5f3759df`, depth 100, the default dictionary/ABI weights (40/4), separate
initially empty corpora, and replay-only showmaps. The primary metric is the
exact set of emitted `(8-byte deployed bytecode-hash prefix, PC)` IDs; corpus
size, collision-free branch-edge and hit-count telemetry, throughput, and RSS
are diagnostics.

The Superform candidate completed in 3,600 seconds wall time. The host
suspended repeatedly during the baseline: Forge still accumulated 3,569
seconds of user CPU and 14.50 million calls before its one-hour active timeout,
but `/usr/bin/time` recorded 10,493 seconds wall time. The coverage comparison
is retained; arm-order and wall-throughput differences are not treated as
causal.

| Campaign | Baseline | Candidate | Interpretation |
| --- | --- | --- | --- |
| Deterministic two-process corpus-reconstruction fixture | No whole-call reconstruction | Reconstructs exact persisted ABI calldata against a hash-only oracle in a second process; one-argument donor mutation reaches a new branch requiring exactly three of four correlated preimages | Establishes cross-process corpus reconstruction and causally attributes the branch to a replayed whole-call donor |
| [Sky `dss-conduits` `invariant_A_B_C_D`](https://github.com/sky-ecosystem/dss-conduits/blob/a09e16356374b3ec4f25292d3ae5b4682790263f/test/arranger-conduit/invariants/BoundedInvariants.t.sol), 1 hour, 2 workers | #15816: 20,096,800 calls; 43 corpus files; 11,540 bytecode/PC IDs | Donor-only `def3ea2f4dc2`: 19,366,600 calls; 51 corpus files; the same 11,540 IDs; summed worker high-water marks of 3,333,985 donor mutations, 654,028 cross-target mutations, 61 hit-count/edge feature wins, and 96 entries | Heavy activation, but no failure and zero new retained bytecode/PC locations. LCOV reported four getter-only lines and zero branches; both arms warned that `--ir-minimum` source mappings can be inaccurate, so this is not usefulness evidence. The baseline-first throughput comparison is not causal because unrelated host contention occurred during the candidate arm. |
| Sky warm-corpus continuation, another hour, 2 workers | — | Donor-only `def3ea2f4dc2`: 18,040,300 additional calls; 51 to 54 corpus files; summed worker high-water mark of 9,333,467 donor mutations; one hit-count win and no new edge; the same 11,540 bytecode/PC IDs | The three files arrived in the first 64 seconds, then the campaign plateaued for the rest of the hour. This raises that exploratory candidate build to 37,406,900 calls without expanding retained PC coverage or finding a failure. |
| [Superform v2 `ERC7540OracleInvariantTest.invariant_INV2_noOverAttribution`](https://github.com/superform-xyz/v2-core/blob/9920d580d636d8d89a2954c9909989df477b3ade/test/invariant/ERC7540OracleInvariant.t.sol#L425-L519), 1 active hour, 1 worker | 14,497,500 calls; 29 corpus files; 40 new-edge and 21 hit-count feature events; 11,194 emitted PC IDs; cmin kept 11/29 entries spanning 1,368 collision-free EVM edges; 129.5 MiB max RSS | 14,207,100 calls; 33 corpus files; 42 new-edge and 26 hit-count feature events; 1,373,633 donor mutations, including 804,289 cross-target and 14 feature-winning mutations; the same 11,194 PC IDs; cmin kept 13/33 entries spanning the same 1,368 edges; 135.0 MiB max RSS | No failure and zero candidate-only PC or collision-free edge identities. The exact sorted edge-key sets have the same SHA-256 (`82f767da…`) and no set difference. The extra live events and two extra minimized entries are consistent with corpus-composition, hit-count, and discovery-order effects, not expanded control-flow reach. |

The Superform repository was pinned at `9920d580d636d8d89a2954c9909989df477b3ade`
and remained clean. Because unrelated suites in that checkout did not compile,
the harness imported the exact upstream invariant and its dependency closure
through a minimal Foundry root; it did not edit the benchmark source. INV2
asserts the vanilla and Centrifuge handlers, while the targeted degraded handler
still contributes cross-target generation traffic but is not checked by that
property.

The deterministic fixture establishes that the mechanism works, but neither
real-world campaign expanded retained PC or collision-free edge coverage or
found a failure. Heavy donor activity and live feature counters are not enough
to justify merging this as default behavior. This therefore remains a draft
and should be treated as experimental until a pinned real target shows a
repeatable candidate-only edge or failure across multiple seeds. It is a
bounded alternative to #15817, not a broad DeFi or time-to-bug claim and not a
claim that nested observed calls belong in the corpus.

Developed with extensive Codex assistance across architecture analysis, implementation, tests, documentation, benchmark harnesses, result analysis, and this PR description.

### DIFF
--- a/.changelog/forge-whole-call-donors.md
+++ b/.changelog/forge-whole-call-donors.md
@@ -1,0 +1,5 @@
+---
+forge: minor
+---
+
+Added bounded reuse of coverage-winning top-level ABI calls during coverage-guided invariant fuzzing.
--- a/crates/config/src/fuzz.rs
+++ b/crates/config/src/fuzz.rs
@@ -66,7 +66,11 @@ impl FuzzConfig {
 /// Contains for fuzz testing
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct FuzzDictionaryConfig {
-    /// The weight of the dictionary
+    /// Percentage of generated values to draw from the fuzz dictionary.
+    ///
+    /// For invariant campaigns with a corpus, this also controls how often compatible whole calls
+    /// whose concrete execution won an edge or hit-count coverage feature are used instead of
+    /// fully synthesized calldata.
     #[serde(deserialize_with = "crate::deserialize_stringified_percent")]
     pub dictionary_weight: u32,
     /// The flag indicating whether to include values from storage
@@ -116,8 +120,12 @@ impl Default for FuzzDictionaryConfig {
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct FuzzCorpusConfig {
-    // Path to corpus directory, enabled coverage guided fuzzing mode.
-    // If not set then sequences producing new coverage are not persisted and mutated.
+    /// Path used to persist coverage-winning fuzz and invariant sequences.
+    ///
+    /// When set, coverage-guided corpus mutation is enabled. In invariant campaigns,
+    /// argument-bearing top-level calls whose concrete execution wins an edge or hit-count
+    /// coverage feature also populate a bounded, worker-local ABI donor dictionary for later call
+    /// generation.
     pub corpus_dir: Option<PathBuf>,
     // Path to fuzz branch frontier artifacts for symbolic follow-up.
     pub frontier_dir: Option<PathBuf>,
@@ -263,6 +271,9 @@ pub struct FuzzCorpusMutationWeights {
     /// Weight for replacing a corpus sequence suffix with generated calls.
     pub mutation_weight_suffix: u32,
     /// Weight for ABI-aware argument mutation.
+    ///
+    /// An effective value of zero also disables collection and reuse of coverage-winning
+    /// whole-call donors.
     pub mutation_weight_abi: u32,
     /// Weight for comparison-operand guided argument mutation.
     pub mutation_weight_cmp: u32,
--- a/crates/evm/evm/src/executors/corpus.rs
+++ b/crates/evm/evm/src/executors/corpus.rs
@@ -41,16 +41,14 @@ use crate::{
 };
 use alloy_dyn_abi::JsonAbiExt;
 use alloy_json_abi::Function;
-use alloy_primitives::{Address, Bytes, I256, U256};
+use alloy_primitives::{Address, B256, Bytes, I256};
 use eyre::{Result, eyre};
-use foundry_common::{ContractsByAddress, ContractsByArtifact, TestFunctionExt, sh_warn};
+use foundry_common::{ContractsByAddress, ContractsByArtifact, sh_warn};
 use foundry_config::{FuzzCorpusConfig, FuzzCorpusMutationWeights};
-use foundry_evm_core::{constants::CALLER, evm::FoundryEvmNetwork, utils::StateChangeset};
+use foundry_evm_core::{constants::MAGIC_ASSUME, evm::FoundryEvmNetwork, utils::StateChangeset};
 use foundry_evm_fuzz::{
-    BasicTxDetails, CallDetails, ObservedCall,
-    invariant::{
-        ArtifactFilters, FuzzRunIdentifiedContracts, InvariantContract, TargetedContracts,
-    },
+    BasicTxDetails,
+    invariant::{ArtifactFilters, FuzzRunIdentifiedContracts, TargetedContracts},
     strategies::{
         EvmFuzzState, FuzzStateReader, InvariantFuzzState, generate_msg_value, mutate_param_value,
     },
@@ -84,9 +82,136 @@ const CACHED_DISK_CORPUS_MAX_LEN: usize = 128;
 // Prefer decoded donors while periodically refreshing the bounded cache from disk.
 const DISK_CORPUS_REFRESH_DENOMINATOR: u32 = 16;
 
+/// Keep whole-call donors worker-local and small enough to scan during deduplication.
+const WHOLE_CALL_DICTIONARY_MAX_LEN: usize = 1024;
+/// Bound retained calldata independently from the number of whole-call donors.
+const WHOLE_CALL_DICTIONARY_MAX_BYTES: usize = 4 * 1024 * 1024;
 /// Threshold for compressing corpus entries.
 /// 4KiB is usually the minimum file size on popular file systems.
 const GZIP_THRESHOLD: usize = 4 * 1024;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+struct WholeCallSignature(B256);
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct WholeCall {
+    /// The unique target that produced this call, or `None` if multiple targets produced it.
+    source_target: Option<Address>,
+    calldata: Bytes,
+}
+
+/// Worker-local ABI donors learned from argument-bearing top-level calls whose complete concrete
+/// execution wins an edge or hit-count coverage feature.
+///
+/// Donors are mutation inputs, not corpus entries. Reusing one never bypasses normal concrete
+/// execution or the existing sequence-level corpus admission rules.
+#[derive(Clone, Default)]
+struct WholeCallDictionary {
+    by_signature: HashMap<WholeCallSignature, VecDeque<WholeCall>>,
+    insertion_order: VecDeque<(WholeCallSignature, Bytes)>,
+    len: usize,
+    calldata_bytes: usize,
+}
+
+impl WholeCallDictionary {
+    fn insert(
+        &mut self,
+        signature: WholeCallSignature,
+        source_target: Address,
+        calldata: Bytes,
+    ) -> bool {
+        let calldata_len = calldata.len();
+        if calldata_len > WHOLE_CALL_DICTIONARY_MAX_BYTES {
+            return false;
+        }
+        if let Some(existing) = self
+            .by_signature
+            .get_mut(&signature)
+            .and_then(|calls| calls.iter_mut().find(|call| call.calldata == calldata))
+        {
+            if existing.source_target != Some(source_target) {
+                existing.source_target = None;
+            }
+            return false;
+        }
+
+        while self.len >= WHOLE_CALL_DICTIONARY_MAX_LEN
+            || self.calldata_bytes + calldata_len > WHOLE_CALL_DICTIONARY_MAX_BYTES
+        {
+            self.evict_oldest();
+        }
+
+        self.by_signature.entry(signature).or_default().push_back(WholeCall {
+            source_target: Some(source_target),
+            calldata: calldata.clone(),
+        });
+        self.insertion_order.push_back((signature, calldata));
+        self.len += 1;
+        self.calldata_bytes += calldata_len;
+        true
+    }
+
+    fn evict_oldest(&mut self) {
+        let Some((signature, calldata)) = self.insertion_order.pop_front() else {
+            return;
+        };
+        let remove_bucket = if let Some(calls) = self.by_signature.get_mut(&signature) {
+            if let Some(index) = calls.iter().position(|call| call.calldata == calldata) {
+                let call = calls.remove(index).expect("whole-call donor index exists");
+                self.len -= 1;
+                self.calldata_bytes = self.calldata_bytes.saturating_sub(call.calldata.len());
+            }
+            calls.is_empty()
+        } else {
+            false
+        };
+        if remove_bucket {
+            self.by_signature.remove(&signature);
+        }
+    }
+
+    fn sample(&self, signature: WholeCallSignature, rng: &mut impl Rng) -> Option<WholeCall> {
+        let calls = self.by_signature.get(&signature)?;
+        calls.get(rng.random_range(0..calls.len())).cloned()
+    }
+
+    const fn is_empty(&self) -> bool {
+        self.len == 0
+    }
+
+    fn insert_tx(&mut self, tx: &BasicTxDetails, targets: &TargetedContracts) -> bool {
+        let Some((function, signature)) = fuzzed_function_for_tx(tx, targets) else {
+            return false;
+        };
+        if function.inputs.is_empty()
+            || function.abi_decode_input(&tx.call_details.calldata[4..]).is_err()
+        {
+            return false;
+        }
+        self.insert(signature, tx.call_details.target, tx.call_details.calldata.clone())
+    }
+}
+
+fn fuzzed_function_for_tx<'a>(
+    tx: &BasicTxDetails,
+    targets: &'a TargetedContracts,
+) -> Option<(&'a Function, WholeCallSignature)> {
+    let selector = tx
+        .call_details
+        .calldata
+        .get(..4)
+        .and_then(|selector| <[u8; 4]>::try_from(selector).ok())?;
+    let selector = selector.into();
+    let contract = targets.get(&tx.call_details.target)?;
+    Some((
+        contract.fuzzed_function_by_selector(selector)?,
+        WholeCallSignature(contract.fuzzed_signature_by_selector(selector)?),
+    ))
+}
+
+pub(crate) const fn is_whole_call_coverage_winner(new_coverage: bool, discarded: bool) -> bool {
+    new_coverage && !discarded
+}
 
 fn weighted_arg_mutation(
     rng: &mut impl Rng,
@@ -503,7 +628,6 @@ fn rewrite_campaign_corpus(config: &FuzzCorpusConfig, gzip: bool) -> Result<()> 
 pub(crate) enum CorpusInsertionMode {
     Live,
     Deferred,
-    MemoryOnly,
     WorkerSync,
 }
 
@@ -546,6 +670,7 @@ struct ReplayCoverage<'a> {
     edge_indices: &'a mut EdgeIndexMap,
     sancov_history_map: &'a mut Vec<u8>,
     metrics: Option<&'a mut CorpusMetrics>,
+    whole_calls: Option<&'a mut WholeCallDictionary>,
 }
 
 /// Campaign-level corpus state produced by replaying persisted corpus entries once.
@@ -562,6 +687,7 @@ pub(crate) struct WorkerCorpusSeed {
     sancov_history_map: Vec<u8>,
     top_rated: HashMap<usize, (Uuid, usize)>,
     metrics: CorpusMetrics,
+    whole_calls: WholeCallDictionary,
     failed_replays: usize,
     optimization_best_value: Option<I256>,
     optimization_best_sequence: Vec<BasicTxDetails>,
@@ -640,6 +766,7 @@ impl WorkerCorpusSeed {
             sancov_history_map: self.sancov_history_map.clone(),
             top_rated: HashMap::new(),
             metrics,
+            whole_calls: self.whole_calls.clone(),
             failed_replays: self.failed_replays,
             optimization_best_value: self.optimization_best_value,
             optimization_best_sequence: self.optimization_best_sequence.clone(),
@@ -693,6 +820,7 @@ impl WorkerCorpusSeed {
         let Some(executor) = executor else {
             return Ok(seed);
         };
+        let collect_whole_calls = config.mutation_weights.effective().mutation_weight_abi > 0;
         let mut seen_entries =
             seed.in_memory_corpus.iter().map(|entry| entry.uuid).collect::<HashSet<_>>();
         for entry in unique_corpus_entries(&canonical_replay_dirs(corpus_dir), &mut seen_entries) {
@@ -716,6 +844,7 @@ impl WorkerCorpusSeed {
                 edge_indices: &mut seed.edge_indices,
                 sancov_history_map: &mut seed.sancov_history_map,
                 metrics: Some(&mut seed.metrics),
+                whole_calls: collect_whole_calls.then_some(&mut seed.whole_calls),
             };
             let ReplayOutcome {
                 keep_entry, new_edge, edges_covered, cmp_seq, failed_replays, ..
@@ -777,6 +906,7 @@ impl WorkerCorpusSeed {
                     edge_indices: &mut edge_indices,
                     sancov_history_map: &mut sancov_history_map,
                     metrics: None,
+                    whole_calls: None,
                 };
                 let ReplayOutcome { keep_entry, new_coverage, .. } =
                     replay_corpus_sequence(&entry.tx_seq, executor, target, coverage)?;
@@ -822,8 +952,13 @@ impl GlobalCorpusMetrics {
             cumulative_features_seen: self.cumulative_features_seen.load(Ordering::Relaxed),
             corpus_count: self.corpus_count.load(Ordering::Relaxed),
             favored_items: self.favored_items.load(Ordering::Relaxed),
+            ..Default::default()
         }
     }
+}
+
+const fn usize_is_zero(value: &usize) -> bool {
+    *value == 0
 }
 
 #[derive(Serialize, Default, Clone)]
@@ -836,6 +971,21 @@ pub(crate) struct CorpusMetrics {
     corpus_count: usize,
     // Number of corpus entries that are favored.
     favored_items: usize,
+    // Worker-local number of compatible whole-call donors selected during fresh generation.
+    #[serde(skip_serializing_if = "usize_is_zero")]
+    whole_call_attempts: usize,
+    // Worker-local number of selected donors that produced changed calldata.
+    #[serde(skip_serializing_if = "usize_is_zero")]
+    whole_call_mutations: usize,
+    // Worker-local subset of mutations whose donor came from a different target.
+    #[serde(skip_serializing_if = "usize_is_zero")]
+    whole_call_cross_target_mutations: usize,
+    // Worker-local number of emitted donor mutations that themselves won coverage.
+    #[serde(skip_serializing_if = "usize_is_zero")]
+    whole_call_coverage_wins: usize,
+    // Current number of retained worker-local donors.
+    #[serde(skip_serializing_if = "usize_is_zero")]
+    whole_call_dictionary_entries: usize,
 }
 
 impl fmt::Display for CorpusMetrics {
@@ -846,6 +996,28 @@ impl fmt::Display for CorpusMetrics {
         writeln!(f, "        - cumulative features seen: {}", self.cumulative_features_seen)?;
         writeln!(f, "        - corpus count: {}", self.corpus_count)?;
         write!(f, "        - favored items: {}", self.favored_items)?;
+        if self.whole_call_attempts == 0
+            && self.whole_call_mutations == 0
+            && self.whole_call_cross_target_mutations == 0
+            && self.whole_call_coverage_wins == 0
+            && self.whole_call_dictionary_entries == 0
+        {
+            return Ok(());
+        }
+        writeln!(f)?;
+        writeln!(f, "        - whole-call donor attempts: {}", self.whole_call_attempts)?;
+        writeln!(f, "        - whole-call donor mutations: {}", self.whole_call_mutations)?;
+        writeln!(
+            f,
+            "        - cross-target whole-call mutations: {}",
+            self.whole_call_cross_target_mutations
+        )?;
+        writeln!(f, "        - whole-call coverage wins: {}", self.whole_call_coverage_wins)?;
+        write!(
+            f,
+            "        - whole-call dictionary entries: {}",
+            self.whole_call_dictionary_entries
+        )?;
         Ok(())
     }
 }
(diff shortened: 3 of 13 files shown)


grandizzy:
derek bench symex

0xalpharush:
@grandizzy what is the point in triggering symex bench on fuzzer worker that isn't a default config? think it would need to be `derek bench invariant` and pass the flag

crates/forge/src/runner.rs
```diff
@@ -4091,22 +4201,59 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                state,
+                self.address,
+                revert_decoder,
+                progress.as_ref(),
+                &self.tcfg.early_exit,
+                symbolic_worker_stop.clone(),
+                &self.cr.tokio_handle,
+            );
+            let symbolic_worker_result =
+                symbolic_worker_handle.and_then(|handle| match handle.join() {
```
grandizzy: **blocking — the race is one-way; fuzz-finishes-first blocks on the symbolic solve**

The stop flag only flows symbolic → fuzz. After `fuzz(...)` returns the main thread never signals the worker before `join()`, and `SymbolicExecutor::run` takes no cancellation input. Since `std::thread::scope` always waits at scope exit, if fuzz finishes first Forge still blocks on the z3 solve — potentially hanging with large/disabled symbolic bounds.

```suggestion
            );
            // Fuzzing is done; signal the symbolic worker to stop before joining.
            if let Some(stop) = symbolic_worker_stop.as_ref() {
                stop.store(true, Ordering::Relaxed);
            }
            let symbolic_worker_result =
                symbolic_worker_handle.and_then(|handle| match handle.join() {
```

This is only half the fix: `SymbolicExecutor::run` must also poll the flag (it already has an internal `cancel: &AtomicBool` for portfolio cancellation — thread the same token in and check it in the path/solver loops), otherwise the store won't interrupt an in-flight solve.

crates/forge/src/runner.rs
```diff
@@ -4091,22 +4201,59 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
             Err(e) => {
                 self.result.fuzz_setup_fail(e);
                 return self.result;
             }
         };
+        if result.success
+            && let Some(symbolic_worker_result) = symbolic_worker_result
+        {
+            result = symbolic_worker_result;
+        }
```
grandizzy: **blocking — merge reports a symbolic result even if it "lost" the race**

Because the main thread waits for the worker, `symbolic_worker_result` can be produced *after* fuzz already passed. Gating on `result.success` still reports a late symbolic counterexample. Combined with an atomic claim in the worker (see the comment on `run_symbolic_fuzz_worker`), the worker returns `Some` only when it actually won, so the merge becomes unconditional:

```suggestion
        if let Some(symbolic_worker_result) = symbolic_worker_result {
            result = symbolic_worker_result;
        }
```

crates/forge/src/runner.rs
```diff
@@ -2912,6 +2920,99 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+            return None;
+        };
+
+        let Ok(raw_call_result) = executor.call_raw(sender, address, calldata.clone(), U256::ZERO)
+        else {
+            return None;
+        };
+        if raw_call_result.result.as_ref() == MAGIC_ASSUME {
+            return None;
+        }
```
grandizzy: **major — skip-handling diverges from the normal replay path**

The normal fuzz / `replay_persisted_failure` path treats cheatcode skips (`reverter == CHEATCODE_ADDRESS` with a decodable `SkipReason`) as skipped, not failures. Only `MAGIC_ASSUME` is filtered here, so a candidate replaying into `vm.skip` is wrongly reported as a failure.

```suggestion
        if raw_call_result.result.as_ref() == MAGIC_ASSUME {
            return None;
        }
        if raw_call_result.reverter == Some(CHEATCODE_ADDRESS)
            && SkipReason::decode(&raw_call_result.result).is_some()
        {
            return None;
        }
```

Better long-term: extract one classify-replay helper shared by normal replay, persisted replay, and this worker.

crates/forge/src/runner.rs
```diff
@@ -4091,22 +4201,59 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
             Ok(x) => x,
             Err(e) => {
                 self.result.fuzz_setup_fail(e);
                 return self.result;
             }
         };
+        if result.success
+            && let Some(symbolic_worker_result) = symbolic_worker_result
+        {
+            result = symbolic_worker_result;
```
grandizzy: **major — wholesale substitution drops normal fuzz telemetry**

Replacing the whole `FuzzTestResult` discards `first_case`, aggregated coverage, `gas_by_case` history, and `failed_corpus_replays` from work fuzz already did. Since this is surfaced as a regular fuzz failure, overlay the failure fields onto the existing result instead of replacing it:

```rust
result.success = false;
result.reason = sym.reason;
result.counterexample = sym.counterexample;
result.traces = sym.traces;
result.gas_report_traces = sym.gas_report_traces;
result.breakpoints = sym.breakpoints;
result.debug_bytecodes = sym.debug_bytecodes;
result.deprecated_cheatcodes = sym.deprecated_cheatcodes;
result.logs.extend(sym.logs);
result.labels.extend(sym.labels);
result.line_coverage = merge_coverage(result.line_coverage, sym.line_coverage);
result.gas_by_case.extend(sym.gas_by_case);
```

(Sketch rather than a suggestion block — the exact coverage-merge call needs confirming against surrounding code.)

crates/forge/src/runner.rs
```diff
@@ -2912,6 +2920,99 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+            logs,
+            labels,
+            traces,
+            gas_report_traces,
+            line_coverage,
+            breakpoints: Some(breakpoints),
+            debug_bytecodes,
+            deprecated_cheatcodes,
+            ..Default::default()
+        })
```
grandizzy: **nit + race-claim — build the result first, then claim the race atomically**

Setting `stop` before building the result means a panic in between logs the panic, returns `None`, yet fuzz may already have halted early via `external_stop` → a truncated "success". Building first and claiming with `compare_exchange` closes the panic window and gives the "who won" guarantee that the unconditional merge above relies on:

```suggestion
        let (breakpoints, deprecated_cheatcodes) =
            raw_call_result.cheatcodes.as_ref().map_or_else(Default::default, |cheats| {
                (cheats.breakpoints.clone(), cheats.deprecated.clone())
            });
        let reason = if raw_call_result.reverter == Some(CHEATCODE_ADDRESS) {
            SkipReason::decode(&raw_call_result.result)
                .map(|reason| reason.to_string())
                .or_else(|| rd.maybe_decode(&raw_call_result.result, raw_call_result.exit_reason))
        } else {
            rd.maybe_decode(&raw_call_result.result, raw_call_result.exit_reason)
        };
        let traces = raw_call_result.traces.clone();
        let debug_bytecodes = raw_call_result.debug_bytecodes.clone();
        let logs = raw_call_result.logs.clone();
        let labels = raw_call_result.labels.clone();
        let line_coverage = raw_call_result.line_coverage.clone();
        let gas_report_traces = traces.clone().into_iter().map(|trace| trace.arena).collect();
        let counterexample =
            BaseCounterExample::from_fuzz_call(calldata, args, raw_call_result.traces);

        let test_result = FuzzTestResult {
            success: false,
            gas_by_case: vec![(raw_call_result.gas_used, raw_call_result.stipend)],
            reason,
            counterexample: Some(CounterExample::Single(counterexample)),
            logs,
            labels,
            traces,
            gas_report_traces,
            line_coverage,
            breakpoints: Some(breakpoints),
            debug_bytecodes,
            deprecated_cheatcodes,
            ..Default::default()
        };

        // Claim the race: only report if fuzzing hasn't already signalled completion.
        if stop.compare_exchange(false, true, Ordering::AcqRel, Ordering::Relaxed).is_err() {
            return None;
        }
        Some(test_result)
```

grandizzy (dismissed):
Please check findings inline

figtracer:
derek bench invariant timeout=3600

crates/forge/tests/cli/test_cmd/symbolic.rs
```diff
@@ -2550,6 +2550,130 @@ Ran 1 test for test/SymbolicFuzzCorpusSeed.t.sol:SymbolicFuzzCorpusSeed
...
+            "testFuzz_symbolicWorker",
+            "--fuzz-runs",
+            "1",
+            "--fuzz-seed",
+            "0x1",
+            "--threads",
+            "1",
+            "--symbolic-fuzz-worker",
+        ])
+        .assert_success();
```
mablr: **should fix — this assertion is scheduler-dependent**

This assumes one fuzz run always finishes before the newly spawned symbolic worker, but the scheduler may run symbolic first, making a failure valid behavior. The later 100,000-run assertion makes the inverse timing assumption and can legitimately pass when the solver is slow or contended. The skip test similarly passes when symbolic never reaches replay, so it does not prove skip classification.

Please avoid using workload size as synchronization. Prefer controlled synchronization for each winner, and test skip replay classification directly (or otherwise prove symbolic replay occurred). At minimum, remove this one-run worker assertion until the race has a deterministic harness.

crates/evm/evm/src/executors/fuzz/mod.rs
```diff
@@ -237,9 +246,11 @@ impl<FEN: FoundryEvmNetwork> FuzzedExecutor<FEN> {
         rd: &RevertDecoder,
         progress: Option<&ProgressBar>,
         early_exit: &EarlyExit,
+        external_stop: Option<Arc<AtomicBool>>,
         tokio_handle: &tokio::runtime::Handle,
     ) -> Result<FuzzTestResult> {
-        let shared_state = SharedFuzzState::new(state, self.config.timeout, early_exit.clone());
+        let shared_state =
+            SharedFuzzState::new(state, self.config.timeout, early_exit.clone(), external_stop);
```
mablr: **blocking — fuzz completion is not linearized with the symbolic claim**

`fuzz()` constructs `shared_state` with the race flag here, but returns the aggregated normal result without claiming that flag. The caller only stores it after this function returns, so it can be descheduled in that interval and allow the symbolic worker to successfully `compare_exchange(false, true, ...)` even though fuzz already finished first.

Please claim completion inside this function, immediately after aggregation and before returning:

```rust
let result = self.aggregate_results(workers, func, &shared_state);
if let Some(stop) = &shared_state.external_stop {
    let _ = stop.compare_exchange(false, true, Ordering::AcqRel, Ordering::Acquire);
}
Ok(result)
```

The caller-side store can remain as error-path/fallback cancellation. Moving that store closer to the call does not close the preemption window.

mablr (changes requested):
Looks good, but found one blocker.

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -1479,6 +1513,7 @@ fn run_solver_process(
     smt: &str,
     timeout: Option<u32>,
     cancel: &AtomicBool,
+    external_cancel: Option<&AtomicBool>,
```
stevencartavia: `stdin.write_all` runs before timeout or cancellation polling. If the pipe fills and the solver stops reading, Forge cannot cancel the symbolic worker and waits when joining it. Should solver input delivery also be cancellable and time-bounded?

stevencartavia (dismissed):

crates/forge/src/runner.rs
```diff
@@ -2912,6 +2961,107 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+            traces,
+            gas_report_traces,
+            line_coverage,
+            breakpoints: Some(breakpoints),
+            debug_bytecodes,
+            deprecated_cheatcodes,
+            ..Default::default()
+        };
+
+        if stop.compare_exchange(false, true, Ordering::AcqRel, Ordering::Relaxed).is_err() {
```
mablr: The fuzz worker only claims `failed_worker_id`; it does not set `stop` until after aggregation. A later symbolic failure can therefore win this CAS and overwrite the earlier fuzz failure. Can we claim the shared winner state when the fuzz failure is first recorded?

crates/forge/src/runner.rs
```diff
@@ -4091,22 +4250,60 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                progress.as_ref(),
+                &self.tcfg.early_exit,
+                symbolic_worker_stop.clone(),
+                &self.cr.tokio_handle,
+            );
+            if let Some(stop) = symbolic_worker_stop.as_ref() {
+                stop.store(true, Ordering::Release);
+            }
+            let symbolic_worker_result =
+                symbolic_worker_handle.and_then(|handle| match handle.join() {
```
mablr: This join can hang indefinitely: the symbolic worker's solver availability probe uses uncancellable `Command::output()` without a timeout. Can that probe use the timeout/cancellation-aware child-process path before we synchronously join here?

crates/forge/src/runner.rs
```diff
@@ -4091,22 +4250,60 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                .zip(symbolic_worker_executor)
+                .map(|(stop, executor)| {
+                    let stop = Arc::clone(stop);
+                    scope.spawn(move || {
+                        Self::run_symbolic_fuzz_worker(
+                            executor,
+                            symbolic_worker_address,
+                            symbolic_worker_sender,
+                            symbolic_worker_config,
+                            symbolic_worker_ffi,
```
mablr: Passing FFI into this worker makes cancellation non-cooperative: symbolic `vm.ffi` uses blocking `Command::output()`, while the fuzz winner waits on the join below. A hanging FFI command can therefore prevent the normal fuzz result from returning.

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -1201,13 +1222,25 @@ fn run_solver_commands(
...
         let mut saw_unsat = false;
+        let mut saw_cancelled = false;
         let mut saw_invalid_sat_model = false;
         let mut errors = Vec::new();
         let mut decisive = None;
         let mut summaries = Vec::new();
 
         while running > 0 || !pending.is_empty() {
-            if decisive.is_none() {
+            if solver_cancelled(&cancel, external_cancel) {
```
grandizzy: **Major:** After external cancellation is observed, queued solver responses are still processed normally. A valid queued `sat` result can set `decisive`, and the final result prioritizes it over `saw_cancelled`. This allows symbolic execution and concrete replay to continue after normal fuzzing has already won.

Internal portfolio cancellation and external race cancellation need separate state. For example:

```rust
let mut externally_cancelled = false;

// ...

if external_cancel.is_some_and(|cancel| cancel.load(Ordering::SeqCst)) {
    externally_cancelled = true;
    cancel.store(true, Ordering::SeqCst);
    while let Some(solver) = pending.pop_front() {
        summaries.push(summary_for_unstarted_solver(solver));
    }
}
```

Then discard every subsequently received result once external cancellation has occurred:

```rust
if externally_cancelled {
    summaries.push(summary_for_cancelled_solver_result(
        index, display, scheduled_after, started_after, elapsed, outcome,
    ));
    continue;
}
```

Give external cancellation precedence in the final result as well. A controlled test should cover `unknown → external cancellation → queued valid sat`.

crates/forge/src/runner.rs
```diff
@@ -2912,6 +2961,107 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+            corpus_seeds: Vec::new(),
+            branch_target: None,
+            cancel: Some(Arc::clone(&stop)),
+        });
+        let SymbolicRunResult::Counterexample { args, calldata, .. } = result else {
+            return None;
+        };
+
+        let Ok(raw_call_result) = executor.call_raw(sender, address, calldata.clone(), U256::ZERO)
+        else {
```
grandizzy: **Major:** Please also recheck the race token immediately before concrete replay. Even with solver-side cancellation fixed, fuzzing can claim completion between `SymbolicExecutor::run` returning and replay starting.

```suggestion
        if stop.load(Ordering::Acquire) {
            return None;
        }
        let Ok(raw_call_result) = executor.call_raw(sender, address, calldata.clone(), U256::ZERO)
```

This prevents a losing symbolic worker from starting a potentially slow fork request or blocking cheatcode execution.

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -1545,6 +1587,62 @@ fn run_solver_process(
...
+    }
+}
+
+fn start_solver_input_writer(
+    mut stdin: ChildStdin,
+    smt: &str,
+) -> mpsc::Receiver<std::io::Result<()>> {
+    let (tx, rx) = mpsc::channel();
+    let smt = smt.as_bytes().to_vec();
+    thread::spawn(move || {
```
grandizzy: **Moderate:** This detached writer owns both solver stdin and a full copy of the SMT query. Cancellation kills only the direct solver child and does not join this thread.

If a custom solver forks a descendant that inherits stdin, killing the direct child leaves the pipe open and this writer blocked indefinitely. Repeated cancellations can leak threads, query-sized allocations, and descendants.

The writer and process lifetime should be owned together. Sketch:

```rust
struct SolverProcess {
    child: SolverChild,
    input_writer: Option<JoinHandle<std::io::Result<()>>>,
}

impl SolverProcess {
    fn cancel_and_join(&mut self) {
        self.child.kill_process_group();
        if let Some(writer) = self.input_writer.take() {
            let _ = writer.join();
        }
    }
}
```

The important part is terminating the entire process group/job object before joining the writer; joining after killing only the direct child can still hang when a descendant inherited the pipe.

crates/forge/src/runner.rs
```diff
@@ -4091,22 +4250,60 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                            symbolic_worker_sender,
+                            symbolic_worker_config,
+                            symbolic_worker_ffi,
+                            symbolic_worker_fuzz_config,
+                            func,
+                            revert_decoder,
+                            stop,
+                        )
+                    })
+                });
```
grandizzy: **Moderate:** If a Rayon fuzz worker panics, control never reaches the explicit stop store below. `thread::scope` then waits for the symbolic worker before resuming the unwind, while that worker still sees `stop == false`.

Please install a cancellation guard before entering fuzzing so unwind paths also signal the worker:

```rust
struct CancelSymbolicWorkerOnDrop(Option<Arc<AtomicBool>>);

impl Drop for CancelSymbolicWorkerOnDrop {
    fn drop(&mut self) {
        if let Some(stop) = &self.0 {
            stop.store(true, Ordering::Release);
        }
    }
}
```

Then inside the scope:

```rust
let _cancel_on_drop =
    CancelSymbolicWorkerOnDrop(symbolic_worker_stop.clone());

let result = fuzzed_executor.fuzz(
    // ...
);
```

The existing explicit store before `join()` can remain; the guard closes the error and panic paths.

grandizzy (changes requested):
Cancellation is much improved, but three remaining paths can still continue or leak symbolic work after fuzzing has won. Details and suggested fixes inline.

crates/forge/src/runner.rs
grandizzy: The guard is present, but it currently lives outside the `thread::scope` closure. If `fuzz()` panics, `thread::scope` waits for its scoped children before returning and unwinding the outer frame, so this guard cannot signal cancellation until after the symbolic worker has already been joined.

Moving it inside the closure ensures it is dropped while the closure unwinds, before `thread::scope` joins the worker:

```diff
diff --git a/crates/forge/src/runner.rs b/crates/forge/src/runner.rs
index b4f8b950edf10..xxxxxxxxxxxxx 100644
--- a/crates/forge/src/runner.rs
+++ b/crates/forge/src/runner.rs
@@ -4272,9 +4272,9 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
             self.result.fuzz_result(result);
             return self.result;
         }
-        let _symbolic_worker_cancel_guard =
-            CancelSymbolicWorkerOnDrop::new(symbolic_worker_stop.clone());
         let (result, symbolic_worker_result) = std::thread::scope(|scope| {
+            let _symbolic_worker_cancel_guard =
+                CancelSymbolicWorkerOnDrop::new(symbolic_worker_stop.clone());
             let symbolic_worker_handle = symbolic_worker_stop
                 .as_ref()
                 .zip(symbolic_worker_executor)
```

The explicit store before `join()` can remain for the normal return path; the inner guard closes the panic/unwind path.

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -1497,17 +1576,11 @@ fn run_solver_process(
...
-    {
-        return SolverProcessOutcome::Error(format!("failed to write solver query: {err}"));
-    }
-
     let started_at = Instant::now();
     let timeout =
         timeout.filter(|seconds| *seconds > 0).map(|seconds| Duration::from_secs(seconds.into()));
     loop {
-        if cancel.load(Ordering::SeqCst) {
+        if solver_cancelled(cancel, external_cancel) {
```
mablr: **Blocking:** this polling only bounds the direct child lifetime. Once that child exits, the unchanged `wait_with_output()` below can still block indefinitely if a solver wrapper forked a descendant inheriting stdout/stderr; the descendant keeps the pipes open, and `SolverChild::drop` only kills the direct PID. Because fuzz synchronously joins this worker, cancellation can still hang. Please own the solver process tree (process group/job object), terminate it on cancellation/timeout, and keep output collection bounded by the same deadline.

crates/forge/src/runner.rs
```diff
@@ -4091,22 +4272,61 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                    Err(_) => {
+                        warn!(test = %func.signature(), "symbolic fuzz worker panicked");
+                        None
+                    }
+                });
+            (result, symbolic_worker_result)
+        });
+        let mut result = match result {
             Ok(x) => x,
             Err(e) => {
```
mablr, line 5012: **Blocking:** a `Some(symbolic_worker_result)` means symbolic already won the shared CAS, but this arm returns `fuzz_setup_fail` before consuming it. `fuzz()` still has fallible paths (for example corpus sync/worker collection) that can return after the symbolic winner is established, so a later fuzz error can overwrite a replay-confirmed counterexample. Resolve the pair together: preserve/merge the symbolic winner when present, and report the fuzz error only when the worker returned `None`.

crates/forge/tests/cli/test_cmd/symbolic.rs
```diff
@@ -2588,6 +2588,59 @@ contract SymbolicFuzzCorpusBestEffort {
...
+            "--symbolic-fuzz-worker",
+            "--ffi",
+            "--fuzz-runs",
+            "1",
+            "--fuzz-seed",
+            "0x1234",
+            "--threads",
+            "1",
+        ])
+        .assert_success()
```
mablr: **Blocking test gap:** this can pass solely because the one normal fuzz run claims completion before the symbolic thread executes, so it does not prove that the symbolic worker reached and rejected `vm.ffi`. More broadly, there is still no deterministic positive test showing a symbolic worker winning, concrete replay succeeding, and its failure being merged, nor a controlled fuzz-win/panic-cancellation test. Please use synchronization or a unit-level race harness rather than fuzz-run counts/scheduler timing.

grandizzy:
One cancellation issue remains after the latest fix: `CancelSymbolicWorkerOnDrop` is created outside the `std::thread::scope` closure. If `fuzz()` panics, `thread::scope` waits for the symbolic worker before returning and unwinding the outer frame, so the guard cannot signal cancellation until after that join has already completed.

Moving the guard inside the scope closure makes it drop during closure unwinding, before scoped children are joined. Proposed diff: https://github.com/foundry-rs/foundry/pull/15631#discussion_r3570306136

crates/forge/src/runner.rs
```diff
@@ -2912,6 +3144,114 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+            cancel: Some(Arc::clone(&stop)),
+        });
+        let SymbolicRunResult::Counterexample { args, calldata, .. } = result else {
+            return None;
+        };
+
+        if stop.load(Ordering::Acquire) {
+            return None;
+        }
+        let Ok(raw_call_result) = executor.call_raw(sender, address, calldata.clone(), U256::ZERO)
```
mablr: Concrete replay is still non-cancellable after it enters `call_raw`. Symbolic execution currently treats `vm.sleep` as a no-op, but replay executes the real blocking `std::thread::sleep`; for example, a narrow symbolic branch containing `vm.sleep(type(uint256).max)` followed by a failure can let symbolic reach replay while normal fuzzing misses the branch and finishes. Fuzz then sets `stop`, but the scoped thread waits indefinitely in `join()` because the replay never polls it.

Maybe we need to reject `vm.sleep` in symbolic-fuzz-worker/replay-safe mode, analogous to disabling FFI for this worker, and add a focused test proving such a path is not concretely replayed.

crates/evm/evm/src/executors/fuzz/hybrid.rs
```diff
@@ -0,0 +1,206 @@
...
+    pub(super) max_attempts: u32,
+}
+
+impl Default for HybridFuzzPolicy {
+    fn default() -> Self {
+        Self {
+            // Wait for the concrete engine to build a useful corpus before spending solver work.
+            warmup_runs: 1_024,
+            // Long enough to avoid reacting to normal short corpus synchronization pauses.
+            plateau: Duration::from_secs(2),
```
0xalpharush: I think this should be much higher likely millions of runs and 15/30 minute

0xalpharush (changes requested):

PR #15638, opened 2026-07-08
ci: add github actions scan

decofe:
Adds the Tempo GitHub Actions scan reusable workflow to CI and includes it in the `ci-success` aggregate check so workflow security and linting are required alongside the existing jobs.

This also fixes the findings that would fail the new scan: pins the remaining checkout tag refs, moves dynamic shell expressions through environment variables, updates stale version comments on pinned actions, and configures actionlint for the repo's Depot runner labels.

Prompted by: @grandizzy

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -193,7 +193,8 @@ jobs:
     steps:
       - name: Checkout repository # zizmor: ignore[artipacked] persists credentials to push benchmark data
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
-        # persist-credentials defaults to true so we can push.
+        with:
+          persist-credentials: true
 
       - name: Download benchmark results
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -70,7 +70,7 @@ jobs:
           sudo apt-get install -y build-essential pkg-config z3
 
       - name: Setup Rust toolchain
-        uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # master
+        uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # v1
         with:
           toolchain: stable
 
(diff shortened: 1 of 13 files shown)


.github/workflows/test.yml
```diff
@@ -121,4 +121,11 @@ jobs:
...
           ETH_SEPOLIA_RPC: ${{ secrets.ETH_SEPOLIA_RPC }}
-        run: cargo nextest run --locked ${{ matrix.flags }}
+          NEXTEST_FILTER: ${{ matrix.filter }}
+          NEXTEST_PARTITION: ${{ matrix.partition_arg }}
+        run: |
+          args=(-E "$NEXTEST_FILTER" --no-fail-fast)
+          if [[ -n "$NEXTEST_PARTITION" ]]; then
+            args+=(--partition "$NEXTEST_PARTITION")
+          fi
+          cargo nextest run --locked "${args[@]}"
```
grandizzy: nit (correctness): The old `matrices.py` emitted `-E '<filter>' --partition count:x/y --no-fail-fast`; this array produces `--no-fail-fast` *before* `--partition`. Equivalent for nextest and moot today (both cases use `n_partitions=1`), but if you want byte-for-byte parity for future partitioned suites, append `--no-fail-fast` last:

```suggestion
          args=(-E "$NEXTEST_FILTER")
          if [[ -n "$NEXTEST_PARTITION" ]]; then
            args+=(--partition "$NEXTEST_PARTITION")
          fi
          args+=(--no-fail-fast)
          cargo nextest run --locked "${args[@]}"
```

.github/workflows/benchmarks.yml
```diff
@@ -191,7 +191,7 @@ jobs:
       pull-requests: write
 
     steps:
-      - name: Checkout repository
+      - name: Checkout repository # zizmor: ignore[artipacked] persists credentials to push benchmark data
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         # persist-credentials defaults to true so we can push.
```
grandizzy: nit: The `zizmor: ignore[artipacked]` is justified and the syntax is correct. Make the intent explicit by setting the default rather than relying on it implicitly:

```suggestion
      - name: Checkout repository # zizmor: ignore[artipacked] persists credentials to push benchmark data
        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
        with:
          persist-credentials: true
```

mattsse (approved):
lgtm

figtracer (approved):

PR #15639, opened 2026-07-08
ci: run nextest step with bash

decofe:
Fixes the post-merge Windows CI failure from the GitHub Actions scan follow-up. The nextest step now builds its arguments with a Bash array, but the default shell on Windows runners is PowerShell, which fails to parse the Bash syntax before tests run. Set the step shell to Bash so the same command works across the push matrix.

Prompted by: @grandizzy


grandizzy (approved):

mattsse (approved):
lgtm

PR #15641, opened 2026-07-08
refactor(anvil): deduplicate Tempo boilerplate

0xMars42:
## Motivation

The Tempo integration in anvil repeats a few blocks of boilerplate that have to be edited in several places whenever a signature or an error mapping changes.

## Solution

Four mechanical extractions, no behavior change:

- `Backend::with_tempo_storage` owns the storage-context preamble (hardfork, chain id, timestamp, block number, db write lock) that the three fee setters repeated (the snapshot is now taken under a single `evm_env` read instead of three), and `tempo_db_err` owns the eyre-to-`DatabaseError` mapping they each spelled out on every fallible call.
- `Backend::build_tempo_evm_env` builds the Tempo `EvmEnv` (spec plus gas params plus `TempoBlockEnv`) that the transact path and the block-executor path both assembled inline.
- `nonce_markers` folds the 2D-nonce fallback around the existing `tempo_parallel_nonce_markers` at the three call sites that repeated it verbatim. The fourth site (`send_raw_transaction`) computes its on-chain nonce lazily inside the fallback and intentionally stays as is, folding it would add an await on the Tempo path.
- `EthApi::ensure_tempo_mode` replaces the `is_tempo` guard the three Tempo-only RPC setters repeated.

`cargo test -p anvil` passes (449 tests; the only failure, `fork::flaky_test_arb_fork_mining`, fails identically on a pristine master checkout from this environment, so it is unrelated), and `cargo clippy -p anvil` is clean.

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -2539,13 +2539,7 @@ impl EthApi<FoundryNetwork> {
         // pre-validate
         self.backend.validate_pool_transaction(&pending_transaction).await?;
 
-        let (requires, provides) = if let Some((requires, provides)) =
-            tempo_parallel_nonce_markers(&pending_transaction)
-        {
-            (requires, provides)
-        } else {
-            (required_marker(nonce, on_chain_nonce, from), vec![to_marker(nonce, from)])
-        };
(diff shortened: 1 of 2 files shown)


stevencartavia (approved):

mablr (approved):
Thanks!

PR #15636, opened 2026-07-08
fix(verify): keep create --verify stdout clean

stevencartavia:
Fixes #1976 

`forge create --verify` and `forge script --verify` leaked the verification submission id/URL to stdout, breaking `--json` output. That line is only meaningful for standalone `forge verify-contract`, so it's now gated behind an opt-in flag and suppressed for embedded verification (the same info is still shown on stderr).

PR #15643, opened 2026-07-08
fix(script): keep verify json stdout clean

mattsse:
Follow-up to #15636. `forge script --broadcast --verify --json` still emitted the script verification summary banner on stdout, so the command could produce non-JSON lines even after the verification submission GUID itself was suppressed. This moves that summary text to stderr, matching the CLI stdout/stderr contract, and adds coverage for the script verification path to ensure stdout remains JSON Lines while the submission GUID is still reported on stderr.

AI assistance was used to prepare this change.


grandizzy (approved):
lgtm

DaniPopes (approved):

PR #15644, opened 2026-07-08
feat(anvil): add anvil_setNextBlockPrevRandao

ritzdorf:
## Motivation

Anvil lets users pin most header fields of the next block — coinbase, timestamp, base fee, and so on — but there was no way to control `prevrandao`, the post-merge replacement for `difficulty` that surfaces in the header as `mixHash` and is exposed to the EVM through the `PREVRANDAO` (`0x44`) opcode. Contracts that read `block.prevrandao` as a randomness source could therefore only be exercised against whatever value anvil derives internally, which makes deterministic tests and the reproduction of a specific on-chain scenario awkward.

## Solution

Add `anvil_setNextBlockPrevRandao`, which sets the `prevrandao` value used for the next mined block. Mirroring the semantics of `evm_setNextBlockTimestamp`, the override is one-shot: it applies to the next block only and is consumed as that block is mined, after which anvil resumes its default per-block derivation from the parent hash and block number. The value is held on the existing cheats state and taken in `do_mine_block`, so it flows into both the block header's `mixHash` and the EVM environment and is observable via the `PREVRANDAO` opcode within the overridden block.

Coverage includes a serde round-trip for the new method plus two integration tests: one asserting the header's `mixHash` is overridden and reverts to the default derivation on the following block, and one asserting the value is visible to the EVM through `Multicall.getCurrentBlockDifficulty`.

*This PR was written primarily by Claude Code.*

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

(similar functions also had no documentation at the moment)

### DIFF
--- a/crates/anvil/src/eth/backend/cheats.rs
+++ b/crates/anvil/src/eth/backend/cheats.rs
@@ -2,7 +2,7 @@
 
 use alloy_evm::precompiles::{Precompile, PrecompileInput};
 use alloy_primitives::{
-    Address, Bytes,
+    Address, B256, Bytes,
     map::{AddressHashSet, foldhash::HashMap},
 };
 use parking_lot::RwLock;
@@ -83,6 +83,22 @@ impl CheatsManager {
     pub fn has_recover_overrides(&self) -> bool {
         !self.state.read().signature_overrides.is_empty()
     }
+
+    /// Sets the `prevrandao` value to use for the next mined block.
+    ///
+    /// This is a one-shot override that is consumed by the next block and applies to that block
+    /// only.
+    pub fn set_next_block_prevrandao(&self, prevrandao: B256) {
+        trace!(target: "cheats", %prevrandao, "set next block prevrandao");
+        self.state.write().next_block_prevrandao.replace(prevrandao);
+    }
+
+    /// Takes the manually set `prevrandao` value for the next block, if any.
+    ///
+    /// This consumes the override so it only applies to a single block.
+    pub fn take_next_block_prevrandao(&self) -> Option<B256> {
+        self.state.write().next_block_prevrandao.take()
+    }
 }
 
 /// Container type for all the state variables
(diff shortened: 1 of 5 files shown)


crates/anvil/src/eth/backend/cheats.rs
```diff
@@ -83,6 +83,22 @@ impl CheatsManager {
...
+        trace!(target: "cheats", %prevrandao, "set next block prevrandao");
+        self.state.write().next_block_prevrandao.replace(prevrandao);
+    }
+
+    /// Takes the manually set `prevrandao` value for the next block, if any.
+    ///
+    /// This consumes the override so it only applies to a single block.
+    pub fn take_next_block_prevrandao(&self) -> Option<B256> {
+        self.state.write().next_block_prevrandao.take()
+    }
```
stevencartavia, line 101: `anvil_setNextBlockPrevRandao` mirrors `evm_setNextBlockTimestamp`, but the override isn't cleared on reset/revert like the timestamp is, so a set-but-unmined value leaks into a later block.

Reproduced:
- set override → `anvil_reset(None)` → mine → block uses the override
- snapshot → set override → `evm_revert` → mine → block still uses the override (timestamp override *is* cleared here)

Cause: `time.reset()` clears the timestamp override, but nothing clears `next_block_prevrandao` in `CheatsManager`.

Fix: clear the override alongside every `time.reset(...)` call (`reset_to_in_mem`, `reset_fork`, `revert_state_snapshot`, `rollback`).

ritzdorf: Thanks @stevencartavia 
Addressed in d8a941708f53af78532a55a0a5938792ad120fc1

stevencartavia (approved):

mablr (approved):

mattsse (approved):

PR #15649, opened 2026-07-09
fix(forge): filter internal libraries from --sizes

### LINKED ISSUE #1356: bug(forge): contract size incorrect for libraries
### Component

Forge

- [X] Foundry
- [X] Foundryup

forge 0.2.0 (b3fe652 2022-04-18T00:11:21.447876194+00:00)

forge build --sizes

### Operating System

_No response_

### Describe the bug

Library sizes aren't reported accurately (at least libraries with no public functions). Have to have a nonabstract/library contract utilize the library to get the size. Not sure if there is an easy fix as the compiler doesnt actually output the true bytecode it seems.

image

stevencartavia:
`forge build --sizes` listed internal libraries (no external/public functions) with a misleading size, even though they're inlined into consumers and never deployed. Now with Solar, it parses sources to detect libraries and skips those with no functions in their ABI. 

Closes #1356.

### DIFF
--- a/crates/common/src/compile.rs
+++ b/crates/common/src/compile.rs
@@ -289,28 +289,39 @@ impl ProjectCompiler {
             }
 
             // Internal libraries are inlined into consumers and never deployed; skip them.
+            // Only artifacts whose ABI has no functions can be internal libraries, so restrict the
+            // solar parse to those sources to avoid a second full parse pass.
             let abs_source = |path: &Path| -> PathBuf {
                 if path.is_absolute() { path.to_path_buf() } else { self.project_root.join(path) }
             };
             let source_paths = artifacts
                 .values()
                 .flatten()
+                .filter(|(_, artifact)| {
+                    artifact.abi.as_ref().is_some_and(|abi| abi.functions().next().is_none())
+                })
                 .map(|(path, _)| abs_source(path))
                 .collect::<BTreeSet<_>>();
             let libraries = collect_libraries(&source_paths);
 
             for (name, artifact_list) in artifacts {
-                for (path, artifact) in &artifact_list {
-                    // A library with no functions in its ABI is internal-only; fail open if the
-                    // ABI is missing.
-                    let is_library =
-                        libraries.get(&abs_source(path)).is_some_and(|libs| libs.contains(&name));
-                    let has_no_abi_functions =
-                        artifact.abi.as_ref().is_some_and(|abi| abi.functions().next().is_none());
-                    if is_library && has_no_abi_functions {
-                        continue;
-                    }
+                // A library with no functions in its ABI is internal-only; fail open if the ABI is
+                // missing. Filter first so the duplicate-name suffix below reflects kept contracts.
+                let kept = artifact_list
+                    .iter()
+                    .filter(|(path, artifact)| {
+                        let is_library = libraries
+                            .get(&abs_source(path))
+                            .is_some_and(|libs| libs.contains(&name));
+                        let has_no_abi_functions = artifact
+                            .abi
+                            .as_ref()
+                            .is_some_and(|abi| abi.functions().next().is_none());
+                        !(is_library && has_no_abi_functions)
+                    })
+                    .collect::<Vec<_>>();
 
+                for (path, artifact) in &kept {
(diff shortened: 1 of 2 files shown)


mattsse (dismissed):
lgtm

crates/common/src/compile.rs
```diff
@@ -283,8 +288,29 @@ impl ProjectCompiler {
...
 
+            // Internal libraries are inlined into consumers and never deployed; skip them.
+            let abs_source = |path: &Path| -> PathBuf {
+                if path.is_absolute() { path.to_path_buf() } else { self.project_root.join(path) }
+            };
+            let source_paths = artifacts
+                .values()
+                .flatten()
+                .map(|(path, _)| abs_source(path))
+                .collect::<BTreeSet<_>>();
```
grandizzy, line 304: Only parse candidate sources. `collect_libraries` reparses every source with solar, but the skip below can only fire for artifacts that have an ABI with zero functions. Restricting the set to those keeps behavior identical while avoiding a full second parse pass on large/cached builds.

```suggestion
            let source_paths = artifacts
                .values()
                .flatten()
                .filter(|(_, artifact)| {
                    artifact.abi.as_ref().is_some_and(|abi| abi.functions().next().is_none())
                })
                .map(|(path, _)| abs_source(path))
                .collect::<BTreeSet<_>>();
```

crates/common/src/compile.rs
```diff
@@ -283,8 +288,29 @@ impl ProjectCompiler {
...
                 for (path, artifact) in &artifact_list {
+                    // A library with no functions in its ABI is internal-only; fail open if the
+                    // ABI is missing.
+                    let is_library =
+                        libraries.get(&abs_source(path)).is_some_and(|libs| libs.contains(&name));
+                    let has_no_abi_functions =
+                        artifact.abi.as_ref().is_some_and(|abi| abi.functions().next().is_none());
+                    if is_library && has_no_abi_functions {
+                        continue;
+                    }
```
grandizzy: `artifact_list.len()` further down (in the `unique_name` branch) is computed pre-filter, so the duplicate-name path suffix can be wrong: when a skipped internal library shares a name with a kept contract, the survivor is now unique in the report but still prints as `Name (src/…)`. Filter the list first, then base both the skip and the suffix decision on the kept entries:

```diff
-                for (path, artifact) in &artifact_list {
-                    // A library with no functions in its ABI is internal-only; fail open if the
-                    // ABI is missing.
-                    let is_library =
-                        libraries.get(&abs_source(path)).is_some_and(|libs| libs.contains(&name));
-                    let has_no_abi_functions =
-                        artifact.abi.as_ref().is_some_and(|abi| abi.functions().next().is_none());
-                    if is_library && has_no_abi_functions {
-                        continue;
-                    }
-
+                // Internal libraries (a `library` with no functions in its ABI) are inlined into
+                // consumers and never deployed; skip them. Fail open if the ABI is missing.
+                let kept = artifact_list
+                    .iter()
+                    .filter(|(path, artifact)| {
+                        let is_library = libraries
+                            .get(&abs_source(path))
+                            .is_some_and(|libs| libs.contains(&name));
+                        let has_no_abi_functions = artifact
+                            .abi
+                            .as_ref()
+                            .is_some_and(|abi| abi.functions().next().is_none());
+                        !(is_library && has_no_abi_functions)
+                    })
+                    .collect::<Vec<_>>();
+
+                for (path, artifact) in &kept {
                     let runtime_size = contract_size(*artifact, false).unwrap_or_default();
                     let init_size = contract_size(*artifact, true).unwrap_or_default();
@@
-                    let unique_name = if artifact_list.len() > 1 {
+                    let unique_name = if kept.len() > 1 {
```

Minor/cosmetic — only affects duplicate-name display.

grandizzy (commented):
Looks good, couple of perf related findings, please check

crates/common/src/compile.rs
stevencartavia: [0009e68](https://github.com/foundry-rs/foundry/pull/15649/commits/0009e68079a88b61b4fa6c43eccf8f68c5dd4e36)

grandizzy (approved):
lgtm, resolves long standing issue, thanks!

mablr (approved):
nice

PR #15651, opened 2026-07-09
chore: replace isolate bench with no-isolate bench

decofe:
## Summary
- replace the explicit isolate forge benchmark with a no-isolate benchmark
- update the benchmark id, dispatch, display label, and command/setup env

## Tests
- cargo +nightly fmt --check
- git diff --check
- cargo check -p foundry-bench (blocked: Cargo cannot fetch foundry-rs/optimism dependency at a305f5a..., HTTP 401/revision not found)

Prompted by: @grandizzy

### DIFF
--- a/benches/src/lib.rs
+++ b/benches/src/lib.rs
@@ -557,8 +557,8 @@ impl BenchmarkProject {
         )
     }
 
-    /// Benchmark forge test with --isolate flag
-    pub fn bench_forge_isolate_test(
+    /// Benchmark forge test with --no-isolate flag
+    pub fn bench_forge_no_isolate_test(
         &self,
         version: &str,
         runs: u32,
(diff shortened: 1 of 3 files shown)


stevencartavia (approved):

grandizzy (approved):

benches/src/lib.rs
```diff
@@ -557,22 +557,22 @@ impl BenchmarkProject {
...
         verbose: bool,
     ) -> Result<HyperfineResult> {
         // Build before running tests
         self.hyperfine(
-            "forge_isolate_test",
+            "forge_no_isolate_test",
             version,
             &self.cmd(
-                "FOUNDRY_DYNAMIC_TEST_LINKING=false FOUNDRY_ISOLATE=true forge test --isolate",
+                "FOUNDRY_DYNAMIC_TEST_LINKING=false FOUNDRY_ISOLATE=false forge test --no-isolate",
```
mablr, line 572: @grandizzy this will not work to bench vs v1.7.1 as there was no `--no-isolate`.

Moreover `bench_forge_no_isolate_test` is now the same as `bench_forge_test`.

grandizzy: Ah, I see, but we can pass isolate false / true as env var `FOUNDRY_` , why the new arg?

mablr: Yeah this is duplicate, env vars should be enough.

mablr: I'll revert this and clean-up

mablr: https://github.com/foundry-rs/foundry/pull/15656

PR #15653, opened 2026-07-09
feat(forge): enable default corpus for forge fuzz

figtracer:
## Summary
Coverage-guided fuzzing already persists and replays generated fuzz corpora when a fuzz corpus dir is configured. This change makes `forge fuzz run` give stateless fuzz tests a default corpus dir so users get corpus persistence and seeding without passing `--fuzz-corpus-dir`.

- Makes `forge fuzz run` default `fuzz.corpus_dir` when the user has not configured it.
- Defaults to the existing fuzz failure-persistence layout: `cache/fuzz/corpus`, or under a custom `fuzz.failure_persist_dir`.
- Preserves explicit config/env/CLI corpus dirs.
- Leaves invariant corpus dirs opt-in because invariant corpus mode can seed from sibling zero-input unit-test traces; users can still opt in via `--corpus-dir`, `--invariant-corpus-dir`, config, or env.
- Leaves `forge fuzz replay`, `cmin`, and `tmin` explicit.

### DIFF
--- a/crates/forge/src/cmd/fuzz.rs
+++ b/crates/forge/src/cmd/fuzz.rs
@@ -48,7 +48,7 @@ impl FuzzArgs {
         match self.command {
             FuzzSubcommands::Run(args) => {
                 let mut test = TestArgs::from_fuzz_run(args);
-                test.enable_fuzz_only_with_auto_corpus();
+                test.enable_fuzz_only_with_auto_fuzz_corpus();
                 Box::pin(test.run())
             }
             FuzzSubcommands::Replay(args) => Box::pin(args.run()),
--- a/crates/forge/src/cmd/fuzz.rs
+++ b/crates/forge/src/cmd/fuzz.rs
@@ -29,9 +29,12 @@ use std::{
     fs::OpenOptions,
     io::{BufWriter, Write},
     path::{Path, PathBuf},
+    pin::Pin,
     time::{SystemTime, UNIX_EPOCH},
 };
 use tempfile::{Builder as TempDirBuilder, TempDir};
+
+type FuzzOutcomeFuture = Pin<Box<dyn Future<Output = Result<TestOutcome>>>>;
 
 /// Run and manage Forge fuzzing corpora.
 #[derive(Clone, Debug, Parser)]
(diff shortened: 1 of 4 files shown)


crates/forge/src/cmd/test/mod.rs
```diff
@@ -1115,7 +1136,35 @@ impl TestArgs {
...
+            return;
+        }
+
+        if config.fuzz.corpus.corpus_dir.is_none() {
+            config.fuzz.corpus.corpus_dir = Some(match &config.fuzz.failure_persist_dir {
+                Some(root) => root.join(AUTO_CORPUS_DIR),
+                None => config.cache_path.join(AUTO_FUZZ_FAILURE_DIR).join(AUTO_CORPUS_DIR),
+            });
+        }
+        if config.invariant.corpus.corpus_dir.is_none() {
```
mablr: Defaulting `invariant.corpus_dir` here also enables invariant corpus seeding from zero-arg unit-test traces. In `forge fuzz run` those tests are otherwise reported as skipped, so this can still execute side effects (`writeFile`, `ffi`, RPC). Can we keep invariant trace seeding opt-in, or separate persistence from unit-test seeding?

crates/forge/src/cmd/fuzz.rs
```diff
@@ -41,22 +44,26 @@ pub struct FuzzArgs {
...
 impl FuzzArgs {
-    pub async fn run(self) -> Result<TestOutcome> {
+    pub fn run(self) -> FuzzOutcomeFuture {
         match self.command {
-            FuzzSubcommands::Run(args) => TestArgs::from_fuzz_run(args).run().await,
-            FuzzSubcommands::Replay(args) => args.run().await,
-            FuzzSubcommands::Show(args) => {
+            FuzzSubcommands::Run(args) => {
+                let mut test = TestArgs::from_fuzz_run(args);
+                test.enable_fuzz_only_with_auto_corpus();
```
mablr: This only affects `forge fuzz run`; scfuzzbench still drives `forge test --mc CryticToFoundry`, so the branch doesn't actually enable corpus mode for scfuzzbench. Please either wire work-local `FOUNDRY_*_CORPUS_DIR` envs into the bench runner or drop that claim.

figtracer: i dropped this accidentally will get it back in

mablr (approved):

stevencartavia (approved):

PR #15655, opened 2026-07-09
fix(config): warn on deprecated keys in all profiles

Osraka:
## Motivation

Deprecated config key warnings were only collected from profile names and then from the currently selected profile. As a result, a deprecated key inside an inactive profile, for example `[profile.ci] deny_warnings = true`, was not reported when loading the default profile.

Refs #15654.

## Solution

Walk the profile tables under `[profile]` and collect their root keys before applying the deprecation map. This keeps the warning path consistent with unknown-key validation, which already checks every profile, while preserving the existing warning shape.

A regression test covers a deprecated key in an inactive profile.

## AI assistance

I used Codex to inspect this code path and run local tests. The final patch is scoped to the config warning traversal and regression test.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

## Tests

- `cargo test -p foundry-config warns_on_deprecated_keys_in_inactive_profiles --lib`
- `cargo test -p foundry-config --lib`
- `cargo +nightly fmt -- --check`
- `cargo check -p foundry-config`

### DIFF
--- a/crates/config/src/providers/warnings.rs
+++ b/crates/config/src/providers/warnings.rs
@@ -150,15 +150,13 @@ impl<P: Provider> WarningsProvider<P> {
             .filter(|(profile, _)| **profile == Config::PROFILE_SECTION)
             .map(|(_, dict)| dict);
 
-        out.extend(profiles.clone().flat_map(BTreeMap::keys).filter_map(deprecated_key_warning));
-        out.extend(
-            profiles
-                .clone()
-                .filter_map(|dict| dict.get(self.profile.as_str().as_str()))
-                .filter_map(Value::as_dict)
-                .flat_map(BTreeMap::keys)
-                .filter_map(deprecated_key_warning),
-        );
+        let deprecated_profile_keys = profiles
+            .clone()
+            .flat_map(BTreeMap::values)
+            .filter_map(Value::as_dict)
+            .flat_map(BTreeMap::keys)
+            .collect::<BTreeSet<_>>();
+        out.extend(deprecated_profile_keys.into_iter().filter_map(deprecated_key_warning));
 
         // Add warning for unknown keys within profiles (root keys only here).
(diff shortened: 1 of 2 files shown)


stevencartavia (dismissed):

crates/config/src/providers/warnings.rs
```diff
@@ -150,15 +150,13 @@ impl<P: Provider> WarningsProvider<P> {
...
-                .flat_map(BTreeMap::keys)
-                .filter_map(deprecated_key_warning),
-        );
+        let deprecated_profile_keys = profiles
+            .clone()
+            .flat_map(BTreeMap::values)
+            .filter_map(Value::as_dict)
+            .flat_map(BTreeMap::keys)
+            .collect::<BTreeSet<_>>();
+        out.extend(deprecated_profile_keys.into_iter().filter_map(deprecated_key_warning));
```
grandizzy, line 159: This fixes deprecated keys inside inactive profiles, but it drops the previous profile-name deprecation path. Before this change, `profiles.clone().flat_map(BTreeMap::keys)` warned for `[profile.cancun]` because `DEPRECATIONS` includes `("cancun", "evm_version = Cancun")`; now only keys inside profile dicts are checked, so `[profile.cancun]` silently loses its deprecation warning. Please keep collecting profile names and union them with keys inside every profile table.

```suggestion
        let deprecated_profile_keys = profiles
            .clone()
            .flat_map(|dict| {
                dict.keys().chain(
                    dict.values().filter_map(Value::as_dict).flat_map(BTreeMap::keys),
                )
            })
            .collect::<BTreeSet<_>>();
        out.extend(deprecated_profile_keys.into_iter().filter_map(deprecated_key_warning));
```

grandizzy (dismissed):
Please check finding inline.

grandizzy (approved):
thank you, lgtm

mablr (approved):
makes sense

PR #15659, opened 2026-07-09
fix(forge): preflight mutate mode conflicts

mablr:
## Motivation

This preflights `--mutate` mode conflicts before compilation so invalid project sources no longer mask clearer CLI incompatibility errors. The shared compatibility check is now reused by `forge coverage`, which rejects `forge coverage --mutate` before the coverage-specific build/analysis path, while `run_tests` keeps the same defensive check for nonstandard callers.


figtracer (approved):

mattsse (approved):
lgtm

PR #15661, opened 2026-07-09
fix(config): support standalone symbolic section

mablr:
Support top-level `[symbolic]` config sections by treating `symbolic` like the existing standalone `fuzz` and `invariant` sections during config normalization. This keeps `[profile.<name>.symbolic]` working as before while allowing shared symbolic defaults to be written as `[symbolic]`, including env var normalization through `FOUNDRY_SYMBOLIC_*` and section-aware unknown-key warnings.


crates/config/src/lib.rs
```diff
@@ -744,6 +744,7 @@ impl Config {
         "doc",
         "fuzz",
         "invariant",
+        "symbolic",
```
figtracer, line 747: this enables top-level `[symbolic]`, but standalone warning validation still uses serialized defaults. valid keys omitted from defaults (`depth`, `width`, `solver_portfolio`, `dynamic_lengths`, frontier filters) parse successfully but now warn as unknown. can you use the same symbolic allowlist here and cover one skipped key in the warning test?

figtracer (changes requested):

crates/config/src/lib.rs
mablr: https://github.com/foundry-rs/foundry/pull/15661/commits/2dcac27901ff31e3807be106937a20086593ec51

stevencartavia, line 747: When an inherited base config uses `[profile.default.symbolic]` and the extending config uses `[symbolic]`, overlapping fields target the same effective default-profile config, but the base value wins despite the documented local-over-base precedence. `no-collision` also does not report this effective overlap. Could we preserve source precedence and collision detection when standalone and profile-qualified `symbolic` entries converge on the same effective key?

stevencartavia, line 747: Adding `symbolic` here also matches the unwrapped `[profile.symbolic]` profile, causing key normalization to be skipped and existing hyphenated keys such as `eth-rpc-url` to be silently ignored. Could we distinguish standalone sections from selected profiles and add coverage for `FOUNDRY_PROFILE=symbolic`?

mattsse (dismissed):
The standalone happy path and warning coverage look good, but this currently changes behavior in two unsupported edge cases that need to be addressed before merging:

- In inherited configs, `[symbolic]` and `[profile.<name>.symbolic]` converge only after the raw providers have been merged. This can make a base value override a local value and lets `no-collision` miss an effective collision: https://github.com/foundry-rs/foundry/pull/15661#discussion_r3563249242
- Adding `symbolic` to `STANDALONE_SECTIONS` also makes an unwrapped profile named `symbolic` bypass key normalization, so valid profile keys such as `eth-rpc-url` can be silently ignored: https://github.com/foundry-rs/foundry/pull/15661#discussion_r3563296039

Please separate standalone-section identity from selected-profile identity during normalization, and normalize standalone/profile-qualified symbolic entries to the same effective shape before applying inheritance precedence and collision detection. Regression coverage should include mixed inherited forms and `FOUNDRY_PROFILE=symbolic`.

grandizzy (approved):
lgtm

figtracer (approved):

PR #15665, opened 2026-07-09
feat(tracing): consolidate trace options

### LINKED ISSUE #3381: Hide address when label is available
### Component

Forge, Cast, Anvil

Hi, I like the label features for traces! However, I think it would make things even easier to read if there was a flag to hide labelled addresses and just show their names.

### Additional context

_No response_

DaniPopes:
Consolidates trace rendering options in shared `TracingArgs` for cast, forge, and script, with defaults under `[tracing]`. CLI overrides resolve into a cloned effective `TracingConfig`, which is passed through trace collection, decoder construction, and rendering. Legacy `[labels]` and profile-local `labels` remain supported with deprecation warnings that point to `[tracing.labels]`. This subsumes #15662 by including `--compact-labels` in the shared trace options and threading it through decoding. Closes #3381. AI assistance was used for this change, per CONTRIBUTING.md.


mattsse (changes requested):
Thanks, this is much cleaner overall. I found two compatibility regressions that should be addressed before merge.

1. `crates/config/src/lib.rs`: the configured checks treat the presence of any `[tracing]` table as if `tracing.verbosity` was explicitly configured. With a config like top-level `verbosity = 4` plus `[tracing] disable_labels = true`, `normalize_tracing_settings` copies the default tracing verbosity (`0`) back into `self.verbosity`, suppressing verbose trace output.

Suggested fix: make the verbosity checks leaf-key only in both `from_figment_inner` and `merge_inline_provider`, e.g. `figment_value_is_configured(&figment, "tracing.verbosity")` / `provider.contains("tracing.verbosity")`, and add a regression test where `[tracing]` contains a non-verbosity option while top-level `verbosity` remains set.

2. `cast run -l <addr:label>` no longer parses. Flattening `TracingArgs` keeps `--labels` and `--label`, but drops the previous `#[arg(long, short)]` `label` field from `RunArgs`.

Suggested fix: avoid adding `short = 'l'` to shared `TracingArgs` globally because `forge test -l` is already `--list`. Instead, keep a command-specific hidden `#[arg(short = 'l', value_name = "ADDRESS:LABEL")]` Vec on `RunArgs`, merge it into `self.tracing.labels` before `resolve`, and add a parse test for `cast run -l ...`.

crates/script/src/lib.rs
```diff
@@ -220,9 +220,8 @@ pub struct ScriptArgs {
     #[arg(long)]
     pub disable_code_size_limit: bool,
 
-    /// Disables the labels in the traces.
-    #[arg(long)]
-    pub disable_labels: bool,
+    #[command(flatten)]
```
mablr, line 231: Adding `--decode-internal` here also needs to update the script runner trace requirements. `_get_runner` still only calls `with_debug(debug)`, so `forge script --decode-internal` enables the decoder later without collecting the jump/stack/memory snapshots it needs.

crates/forge/tests/cli/config.rs
```diff
@@ -287,7 +287,11 @@ skip_files = []
 include_operators = []
 exclude_operators = []
 
-[labels]
+[tracing]
```
mablr, line 290: Since the default config now serializes this `tracing` section, the `forge config --json` snapshot below also needs the matching `tracing` object. `cargo nextest run -p forge --test cli config::test_default_config` currently fails on that missing JSON entry.

mablr (commented):
I confirm Matt's findings, and have two more.

crates/forge/tests/cli/config.rs
stevencartavia: +1

crates/script/src/lib.rs
stevencartavia: +1

crates/config/src/lib.rs
```diff
@@ -939,12 +953,49 @@ impl Config {
         Ok(config)
     }
 
+    fn normalize_tracing_figment(mut figment: Figment) -> Figment {
+        if figment.contains("tracing.verbosity")
+            && let Ok(verbosity) = figment.extract_inner::<u8>("tracing.verbosity")
+        {
+            figment = figment.merge(("verbosity", verbosity));
```
mablr: Alias normalization happens only after all providers have been merged, so using different spellings can reverse provider precedence. A global `[tracing] verbosity = 5` overrides both project-local `verbosity = 1` and `FOUNDRY_VERBOSITY=1`; a global `tracing.labels` entry likewise overrides a colliding project-local legacy `labels` entry.

crates/config/src/lib.rs
```diff
@@ -550,6 +555,7 @@ pub struct Config {
     pub enable_tx_gas_limit: bool,
 
     /// Address labels
+    #[serde(default, skip_serializing)]
```
mablr: `skip_serializing` silently discards programmatic updates to the public legacy `Config::labels` field. `Config::default()` with `verbosity = 4` and an inserted legacy label serializes without that label and with `[tracing] verbosity = 0`; reloading the emitted settings resolves both verbosity fields to `0`.

crates/cli/src/opts/tracing.rs
```diff
@@ -0,0 +1,110 @@
...
+    )]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub labels: Vec<String>,
+}
+
+impl TracingArgs {
+    /// Resolves CLI overrides against the configured trace rendering settings.
+    pub fn resolve(&self, config: &TracingConfig, verbosity: u8) -> TracingConfig {
+        let mut tracing = config.clone();
+        tracing.verbosity = tracing.verbosity.max(verbosity);
```
mablr, line 57: Using `max` gives Cast different CLI precedence from Forge. With `[tracing] verbosity = 5`, `forge ... -v` resolves to `1` through `EvmArgs`, while Cast keeps `5` here. Explicit verbosity flags historically override configured verbosity.

crates/config/src/lib.rs
stevencartavia: +1 from my pass

crates/script/src/simulate.rs
```diff
@@ -180,7 +180,7 @@ impl<FEN: FoundryEvmNetwork> PreSimulationState<FEN> {
             let (tx, is_noop_tx, mut traces) = res?;
 
             // Transaction will be `None`, if execution didn't pass.
-            if tx.is_none() || self.script_config.evm_opts.verbosity > 3 {
+            if tx.is_none() || self.script_config.config.tracing.verbosity > 3 {
                 for (_, trace) in &mut traces {
```
mablr, line 185: `--trace-depth` is applied to the initial script execution traces, but not to the simulated on-chain transaction traces rendered here. As a result, `forge script --trace-depth N` may still print traces deeper than `N` during simulation.

crates/script/src/simulate.rs
```diff
@@ -180,9 +181,12 @@ impl<FEN: FoundryEvmNetwork> PreSimulationState<FEN> {
             let (tx, is_noop_tx, mut traces) = res?;
 
             // Transaction will be `None`, if execution didn't pass.
-            if tx.is_none() || self.script_config.evm_opts.verbosity > 3 {
+            if tx.is_none() || tracing.verbosity > 3 {
```
mablr: This prints trace trees to stdout under `--json` when configured verbosity is >3, corrupting machine-readable output. `!shell::is_json()` should be included in this guard.

crates/script/src/execute.rs
```diff
@@ -469,6 +476,9 @@ impl<FEN: FoundryEvmNetwork> PreSimulationState<FEN> {
                 if should_include {
                     let mut trace = trace.clone();
                     decode_trace_arena(&mut trace, decoder).await;
+                    if let Some(trace_depth) = tracing.trace_depth {
+                        prune_trace_depth(&mut trace, trace_depth);
```
mablr, line 484: Should the same pruning be applied in `show_json`? Currently `forge script --json --trace-depth 0` still serializes nested calls, unlike text execution and simulation output.

mattsse (dismissed):
The requested compatibility regressions are addressed: tracing verbosity is no longer inferred from the presence of the parent table, and `cast run -l` is preserved with a command-specific compatibility alias. Both paths have regression coverage, and the latest CI run is green.

crates/evm/traces/src/decoder/mod.rs
```diff
@@ -1193,6 +1221,9 @@ impl CallTraceDecoder {
         if let DynSolValue::Address(addr) = value
             && let Some(label) = self.labels.get(addr)
         {
+            if self.compact_labels {
```
stevencartavia, line 1225: `--compact-labels --disable-labels` currently renders labeled address parameters as only the label, for example `takes(Alice)`, hiding the raw address even though labels are disabled. Could `disable_labels` take precedence in `format_value` so this combination renders the raw address instead?

crates/config/src/lib.rs
```diff
@@ -940,12 +951,33 @@ impl Config {
...
+
     fn normalize_hardfork_settings(&mut self) -> Result<(), Error> {
         let Some(hardfork) = self.hardfork else { return Ok(()) };
         self.networks = self.networks.normalize_for_hardfork(hardfork).map_err(Error::from)?;
         Ok(())
     }
 
+    fn normalize_tracing_settings(&mut self, tracing_labels_configured: bool) {
+        if tracing_labels_configured {
+            self.labels.extend(self.tracing.labels.clone());
```
stevencartavia: With only `[tracing.labels]` configured, this also populates deprecated `Config::labels`, so `forge config` emits both tables and reloading its output warns about `[labels]`. Could it emit only the canonical table while preserving compatibility for callers that populate `Config::labels` directly?

crates/script/src/lib.rs
```diff
@@ -896,7 +896,14 @@ impl<FEN: FoundryEvmNetwork> ScriptConfig<FEN> {
...
                     .trace_requirements(
-                        TraceRequirements::none().with_calls(true).with_debug(debug),
+                        TraceRequirements::none()
+                            .with_calls(true)
+                            .with_debug(debug)
+                            .with_decode_internal(if self.config.tracing.decode_internal {
+                                InternalTraceMode::Full
+                            } else {
+                                InternalTraceMode::None
+                            }),
```
mattsse: This is a pre-existing issue, not a regression introduced by this PR: the base branch’s script runner also did not apply verbosity to `TraceRequirements`. The new shared tracing configuration inherits that gap, so adding `.with_verbosity(self.config.tracing.verbosity)` and a script regression test can be handled as follow-up work.

mattsse (commented):
One remaining issue.

mattsse (dismissed):
LGTM

crates/cast/src/cmd/call.rs
```diff
@@ -388,12 +391,9 @@ impl CallArgs {
                 &config,
                 chain,
                 &contracts_bytecode,
-                labels,
+                &tracing,
```
mablr, line 401: Configured `decode_internal` leaks into this RPC-only path. `TracingArgs::resolve` preserves `[tracing] decode_internal = true`, even though Clap only prevents passing `--decode-internal` alongside `--debug-trace-call`. A callTracer arena has no opcode/jump snapshots, so `handle_traces` cannot decode internals here; instead it may download and compile externally identified sources, fail an otherwise renderable RPC trace, and print `Compiling: ...` into JSON stdout.

crates/cast/src/cmd/call.rs
```diff
@@ -257,14 +261,13 @@ impl CallArgs {
...
-            decode_internal,
-            labels,
             data,
             with_local_artifacts,
-            disable_labels,
+            tracing,
             wallet,
             ..
         } = self;
+        let tracing = tracing.resolve(&config.tracing, shell::verbosity());
```
mablr: The config used here was loaded directly with `Config::from_provider`, bypassing the warning emission performed by `LoadConfig::load_config`. Consequently, `cast call --trace` silently consumes deprecated `[labels]` / profile-local `labels` instead of showing the migration warning to `[tracing.labels]`; `cast run` has the same issue.

mattsse (approved):

mablr (approved):

PR #15666, opened 2026-07-09
fix: support ArbSys block number on Arbitrum forks

### LINKED ISSUE #5085: Arbitrum fork can't read l2 block number
### Component

Forge, Anvil

- [X] Foundry
- [X] Foundryup

forge 0.2.0 (1caa0fa 2023-05-30T00:11:10.062347000Z)

forge test --mt testGetArbL2BlockNumber -vvvv

### Operating System

macOS (Apple Silicon)

### Describe the bug

Since `block.number` on arbitrum is block number on l1, arbitrum has a precompiled address (0x64) to reflect l2 address. Call address(0x64) `arbBlockNumber` to get l2 block number.

```
interface ArbSys {
    /**
     * @notice Get Arbitrum block number (distinct from L1 block number; Arbitrum genesis block has block number 0)
     * @return block number as int
     */
    function arbBlockNumber() external view returns (uint);
}
```

Create arbitrum fork and call 0x64 with `arbBlockNumber()`, it wil revert with `"EvmError: InvalidFEOpcode"`. 
I think it's just because the foundry evm doesn't fulfill the precompile contract function, but there should be a way to get l2 block number on arbitrum fork. Fulfill it to compatible with arbitrum official implemtation would be great.

A minimal test to reproduce it

```
interface ArbSys {
    /**
     * @notice Get Arbitrum block number (distinct from L1 block number; Arbitrum genesis block has block number 0)
     * @return block number as int
     */
    function arbBlockNumber() external view returns (uint);
}

contract ArbL2NumberTest is Test {
    function testGetArbL2BlockNumber() public {
        // <https://arbiscan.io/block/75219831>
        vm.createSelectFork("https://rpc.ankr.com/arbitrum", 75219831);

        ArbSys(address(0x64)).arbBlockNumber();
    }
}
```

stevencartavia:
Adds narrow Arbitrum fork compatibility for `ArbSys.arbBlockNumber()` at address `0x64`.

This keeps `eth_getCode` behavior unchanged while making local Forge and Anvil fork execution return the current Arbitrum L2 block number, matching live RPC behavior.

Fixes #5085

### DIFF
--- a/crates/evm/networks/src/arbitrum.rs
+++ b/crates/evm/networks/src/arbitrum.rs
@@ -1,0 +1,45 @@
+//! Minimal Arbitrum system contract compatibility helpers.
+
+use std::borrow::Cow;
+
+use alloy_chains::Chain;
+use alloy_evm::precompiles::{DynPrecompile, PrecompileInput};
+use alloy_primitives::{Address, Bytes, U256, address, hex};
+use revm::precompile::{PrecompileHalt, PrecompileId, PrecompileOutput, PrecompileResult};
+
+/// ArbSys system contract address.
+pub const ARB_SYS_ADDRESS: Address = address!("0000000000000000000000000000000000000064");
+
+/// `ArbSys.arbBlockNumber()` selector.
+pub const ARB_BLOCK_NUMBER_SELECTOR: [u8; 4] = hex!("a3b1b31d");
+
+/// ID for the ArbSys precompile.
+pub static PRECOMPILE_ID_ARB_SYS: PrecompileId = PrecompileId::Custom(Cow::Borrowed("ArbSys"));
+
+/// Returns whether `chain_id` is an Arbitrum chain.
+pub fn is_arbitrum_chain(chain_id: u64) -> bool {
+    Chain::from_id(chain_id).is_arbitrum()
+}
+
+/// Returns the ABI-encoded result for `ArbSys.arbBlockNumber()`.
+pub fn arb_block_number_output(block_number: u64) -> Bytes {
+    Bytes::copy_from_slice(&U256::from(block_number).to_be_bytes::<32>())
+}
+
+/// Returns an ArbSys precompile for the provided L2 block number.
+pub fn arb_sys_precompile(block_number: u64) -> DynPrecompile {
+    DynPrecompile::new_stateful(PRECOMPILE_ID_ARB_SYS.clone(), move |input| {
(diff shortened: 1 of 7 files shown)


mattsse (dismissed):
lgtm

crates/evm/networks/src/arbitrum.rs
```diff
@@ -0,0 +1,45 @@
...
+
+fn arb_sys_precompile_call(input: PrecompileInput<'_>, block_number: u64) -> PrecompileResult {
+    if input.data.get(..4) != Some(&ARB_BLOCK_NUMBER_SELECTOR) {
+        return Ok(PrecompileOutput::halt(
+            PrecompileHalt::Other("unsupported ArbSys selector".into()),
+            input.reservoir,
+        ));
+    }
+
+    Ok(PrecompileOutput::new(0, arb_block_number_output(block_number), input.reservoir))
```
grandizzy: This makes `ArbSys.arbBlockNumber()` free in the Anvil precompile path. The Forge inspector path also returns `Gas::new(call.gas_limit)` without recording any cost, so gas-limited calls to `0x64` can succeed locally when they should fail/OOG on Arbitrum. Since this PR is specifically trying to match live Arbitrum fork execution, the supported ArbSys method should charge a shared gas cost in both paths.

A good way to avoid the Anvil/Forge behavior split would be to keep the backend-specific part limited to computing the current L2 block number, then move the shared ArbSys precompile registration/call/gas/error behavior into `foundry_evm_networks::arbitrum` and call it through the existing network precompile injection path.

```suggestion
    let gas_cost = arb_block_number_gas_cost();
    if input.gas < gas_cost {
        return Ok(PrecompileOutput::halt(PrecompileHalt::OutOfGas, input.reservoir));
    }

    Ok(PrecompileOutput::new(gas_cost, arb_block_number_output(block_number), input.reservoir))
```

And apply the same shared `gas_cost` in the Forge inspector `CallOutcome` path instead of returning `Gas::new(call.gas_limit)` unchanged.

grandizzy (changes requested):
The ArbSys block number source is backend-specific, but the ArbSys precompile behavior itself should be centralized in the networks crate and wired through the existing precompile injection path. That would avoid splitting selector/output/gas/error behavior across Anvil and Forge paths. Please check the inline finding as well.

grandizzy (approved):
lgtm, we can do the networks cleanup when needed

mablr (approved):

PR #15668, opened 2026-07-10
feat(cheatcodes): allow mockCall without code injection

### LINKED ISSUE #10703: mockCall could lead to issues

mattsse:
`vm.mockCall` etches a single byte onto an empty target account so that Solidity's `extcodesize` check passes. As reported in #10703, this has the side effect that once any mock is registered for an empty address, every unmocked call to that address silently succeeds instead of reverting, which can hide bugs in tests. Following the approach agreed on in the issue, this adds a `mockCall(address callee, bytes calldata data, bytes calldata returnData, bool injectCode)` overload that lets users opt out of the code injection with `injectCode = false`, leaving the account codeless so unmocked calls to it revert in the caller as they would for any address without code.

One nuance, documented on the cheatcode: without the injected code, mocked calls that return data still succeed because solc checks `returndatasize()` instead of `extcodesize()` when return data is expected, but mocked calls to functions without return values may still revert in the caller due to the `extcodesize` check, so the opt-out is primarily useful when the mocked functions return data. The new overload is placed after the existing `mockCall` overloads so the generated call struct names remain stable.

Closes #10703

AI assistance was used to implement this change.

### DIFF
--- a/crates/cheatcodes/assets/cheatcodes.json
+++ b/crates/cheatcodes/assets/cheatcodes.json
@@ -7370,6 +7370,26 @@
     },
     {
       "func": {
+        "id": "mockCall_4",
+        "description": "Mocks a call to an address, returning specified data.\nCalldata can either be strict or a partial match, e.g. if you only\npass a Solidity selector to the expected calldata, then the entire Solidity\nfunction will be mocked.\nOverload to control whether code is injected into `callee`. The other overloads etch a\nsingle byte into an empty account to circumvent Solidity's `extcodesize` check, with the\nside effect that unmocked calls to it no longer revert; `injectCode = false` leaves the\naccount codeless, so unmocked calls to it revert in the caller. Mocked calls that return\ndata still succeed, as Solidity checks `returndatasize()` instead of `extcodesize()` when\nreturn data is expected, but mocked calls to functions without return values may still\nrevert in the caller due to the `extcodesize` check.",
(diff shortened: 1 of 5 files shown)


grandizzy (approved):
lgtm

figtracer (approved):

PR #15669, opened 2026-07-10
fix(lint): emit safe suggestions in unwrapped-modifier-logic

### LINKED ISSUE #12724: bug(lint): unwrap modifier doesn't handle var assignments properly
### Component

Forge

- [ ] Foundry
- [ ] Foundryup

_No response_

_No response_

_No response_

### Operating System

None

### Describe the bug

repro:
```sh
           - modifier payMeSubsidizedGas(uint256 amount) {
           -     uint256 pre = gasLeft();
           -     _;
           -     _payMeSubsidizedGasAfter(pre, amount);
           - }
           + modifier payMeSubsidizedGas(uint256 amount) {
           +     _payMeSubsidizedGas(amount);
           +     _;
           + }
           + 
           + function _payMeSubsidizedGas(uint256 amount) internal {
           +     uint256 pre = gasLeft();
           + }
```

mattsse:
The machine-applicable suggestion emitted by `unwrapped-modifier-logic` could corrupt user code in three ways. First, when only one side of the placeholder required wrapping, the rewritten modifier body consisted solely of the helper call and `_;`, silently dropping the other side's statements from the modifier. Second, a local variable declared before `_;` was hoisted into the extracted helper along with the rest of the segment, so any reference to it after the placeholder no longer resolved, since helpers only receive the modifier's parameters. Third, extracting a write to a modifier parameter into the pre-placeholder helper could change later reads because value-type parameters are passed to the helper by value.

The rewrite now preserves the non-wrapped side's statements verbatim in the new modifier body and bails out entirely when a local declared before the placeholder is referenced after it. It also suppresses the suggestion when extracting the pre-placeholder segment would separate a modifier parameter write from a later reference. The detection walks the relevant HIR, including assembly blocks, and recognizes assignments, deletion, increment and decrement operations, tuple targets, and indexed or member targets. Snippet extraction failures now also abort the suggestion instead of producing partial code.

Closes #12724

### DIFF
--- a/crates/lint/src/sol/codesize/unwrapped_modifier_logic.rs
+++ b/crates/lint/src/sol/codesize/unwrapped_modifier_logic.rs
@@ -5,8 +5,9 @@ use crate::{
 };
 use solar::{
     ast,
-    sema::hir::{self, Res},
+    sema::hir::{self, Res, Visit as _},
 };
+use std::ops::ControlFlow;
 
 declare_forge_lint!(
     UNWRAPPED_MODIFIER_LOGIC,
@@ -113,18 +114,25 @@ impl UnwrappedModifierLogic {
         res
     }
 
-    fn get_snippet<'a>(
+    fn get_snippet<'hir>(
         &self,
         ctx: &LintContext,
-        hir: &hir::Hir<'_>,
+        hir: &'hir hir::Hir<'hir>,
         func: &hir::Function<'_>,
-        before: &'a [hir::Stmt<'a>],
-        after: &'a [hir::Stmt<'a>],
+        before: &'hir [hir::Stmt<'hir>],
+        after: &'hir [hir::Stmt<'hir>],
     ) -> Option<Suggestion> {
         let wrap_before = !before.is_empty() && self.stmts_require_wrapping(hir, before);
         let wrap_after = !after.is_empty() && self.stmts_require_wrapping(hir, after);
 
         if !(wrap_before || wrap_after) {
+            return None;
+        }
+
+        // A local variable declared before the placeholder and referenced after it makes any
+        // rewrite unsafe: extracted helpers only receive the modifier's parameters, so moving
+        // either side out of the modifier separates the declaration from its use.
+        if has_shared_locals(hir, before, after) {
             return None;
         }
 
@@ -152,18 +160,27 @@ impl UnwrappedModifierLogic {
         let body_indent = " ".repeat(ctx.get_span_indentation(
             before.first().or(after.first()).map(|stmt| stmt.span).unwrap_or(func.span),
         ));
-        let body = match (wrap_before, wrap_after) {
-            (true, true) => format!(
-                "{body_indent}_{modifier_name}Before({param_list});\n{body_indent}_;\n{body_indent}_{modifier_name}After({param_list});"
-            ),
-            (true, false) => {
-                format!("{body_indent}_{modifier_name}({param_list});\n{body_indent}_;")
+        // Statements on a side that doesn't require wrapping are preserved verbatim in the new
+        // modifier body, so the rewrite never drops them.
+        let mut body_lines = Vec::new();
+        if wrap_before {
+            let suffix = if wrap_after { "Before" } else { "" };
+            body_lines.push(format!("{body_indent}_{modifier_name}{suffix}({param_list});"));
+        } else {
+            for stmt in before {
+                body_lines.push(format!("{body_indent}{}", ctx.span_to_snippet(stmt.span)?));
             }
-            (false, true) => {
-                format!("{body_indent}_;\n{body_indent}_{modifier_name}({param_list});")
+        }
+        body_lines.push(format!("{body_indent}_;"));
+        if wrap_after {
+            let suffix = if wrap_before { "After" } else { "" };
+            body_lines.push(format!("{body_indent}_{modifier_name}{suffix}({param_list});"));
+        } else {
+            for stmt in after {
+                body_lines.push(format!("{body_indent}{}", ctx.span_to_snippet(stmt.span)?));
             }
-            _ => unreachable!(),
-        };
+        }
+        let body = body_lines.join("\n");
 
         let mod_indent = " ".repeat(ctx.get_span_indentation(func.span));
         let mut replacement =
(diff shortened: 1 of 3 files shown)


crates/lint/src/sol/codesize/unwrapped_modifier_logic.rs
```diff
@@ -128,6 +129,13 @@ impl UnwrappedModifierLogic {
             return None;
         }
 
+        // A local variable declared before the placeholder and referenced after it makes any
+        // rewrite unsafe: extracted helpers only receive the modifier's parameters, so moving
+        // either side out of the modifier separates the declaration from its use.
+        if has_shared_locals(hir, before, after) {
```
grandizzy: This still seems unsafe to mark as `MachineApplicable`. The new guard covers locals declared before `_` and read after it, but not modifier parameters that are mutated before `_` and read after `_`.

Because the generated helpers receive parameters by value, auto-applying the fix can silently change behavior. For example:

```solidity
modifier bump(uint256 x) {
    x += 1;
    _;
    check(x);
}
```

can become:

```solidity
modifier bump(uint256 x) {
    _bump(x);
    _;
    check(x); // now observes the original x, not x + 1
}

function _bump(uint256 x) internal {
    x += 1;
}
```

Since this suggestion is emitted as `Applicability::MachineApplicable`, tools may apply it without review, so I think this needs to be guarded before this PR can safely keep machine applicability.

Suggested fix: also suppress the machine-applicable suggestion when a modifier parameter is written before `_` and then referenced after `_`:

```diff
diff --git a/crates/lint/src/sol/codesize/unwrapped_modifier_logic.rs b/crates/lint/src/sol/codesize/unwrapped_modifier_logic.rs
@@
-        if has_shared_locals(hir, before, after) {
+        if has_shared_locals(hir, before, after)
+            || has_written_params_used_after(hir, func.parameters, before, after)
+        {
             return None;
         }
@@
 fn has_shared_locals<'hir>(
     hir: &'hir hir::Hir<'hir>,
     before: &'hir [hir::Stmt<'hir>],
     after: &'hir [hir::Stmt<'hir>],
 ) -> bool {
@@
     let mut finder = SharedLocalFinder { hir, locals: &declared };
     after.iter().any(|stmt| finder.visit_stmt(stmt).is_break())
 }
+
+fn has_written_params_used_after<'hir>(
+    hir: &'hir hir::Hir<'hir>,
+    params: &'hir [hir::VariableId],
+    before: &'hir [hir::Stmt<'hir>],
+    after: &'hir [hir::Stmt<'hir>],
+) -> bool {
+    let mut written = Vec::new();
+    let mut writer = ParamWriteFinder { hir, params, written: &mut written };
+    for stmt in before {
+        let _ = writer.visit_stmt(stmt);
+    }
+
+    if written.is_empty() {
+        return false;
+    }
+
+    let mut finder = SharedLocalFinder { hir, locals: &written };
+    after.iter().any(|stmt| finder.visit_stmt(stmt).is_break())
+}
+
+struct ParamWriteFinder<'a, 'hir> {
+    hir: &'hir hir::Hir<'hir>,
+    params: &'a [hir::VariableId],
+    written: &'a mut Vec<hir::VariableId>,
+}
+
+impl<'hir> hir::Visit<'hir> for ParamWriteFinder<'_, 'hir> {
+    type BreakValue = ();
+
+    fn hir(&self) -> &'hir hir::Hir<'hir> {
+        self.hir
+    }
+
+    fn visit_expr(&mut self, expr: &'hir hir::Expr<'hir>) -> ControlFlow<Self::BreakValue> {
+        match &expr.kind {
+            // `x = ...`, `x += ...`, `(x, y) = ...`
+            hir::ExprKind::Assign(lhs, _, _) => {
+                collect_written_params(lhs, self.params, self.written);
+            }
+            // `delete x`
+            hir::ExprKind::Delete(inner) => {
+                collect_written_params(inner, self.params, self.written);
+            }
+            // `++x`, `x++`, `--x`, `x--`
+            hir::ExprKind::Unary(op, inner)
+                if matches!(
+                    op.kind,
+                    ast::UnOpKind::PreInc
+                        | ast::UnOpKind::PreDec
+                        | ast::UnOpKind::PostInc
+                        | ast::UnOpKind::PostDec
+                ) =>
+            {
+                collect_written_params(inner, self.params, self.written);
+            }
+            _ => {}
+        }
+
+        self.walk_expr(expr)
+    }
+}
+
+fn collect_written_params<'hir>(
+    expr: &'hir hir::Expr<'hir>,
+    params: &[hir::VariableId],
+    written: &mut Vec<hir::VariableId>,
+) {
+    match &expr.kind {
+        hir::ExprKind::Ident(resolutions) => {
+            for res in *resolutions {
+                if let Res::Item(hir::ItemId::Variable(id)) = res
+                    && params.contains(id)
+                    && !written.contains(id)
+                {
+                    written.push(*id);
+                }
+            }
+        }
+        hir::ExprKind::Tuple(items) => {
+            for item in *items {
+                if let Some(item) = item {
+                    collect_written_params(item, params, written);
+                }
+            }
+        }
+        hir::ExprKind::Index(base, _)
+        | hir::ExprKind::Slice(base, _, _)
+        | hir::ExprKind::Member(base, _)
+        | hir::ExprKind::YulMember(base, _) => {
+            collect_written_params(base, params, written);
+        }
+        _ => {}
+    }
+}
```

A regression test should also assert no machine-applicable suggestion is emitted here:

```diff
diff --git a/crates/lint/testdata/UnwrappedModifierLogic.sol b/crates/lint/testdata/UnwrappedModifierLogic.sol
@@
     modifier restoreFreeMemory() {
         uint256 m = gasLeft();
         checkPublic(msg.sender);
         _;
         assembly ("memory-safe") {
             mstore(0x40, m)
         }
     }
+
+    // No lint: `amount` is mutated before the placeholder and read after it.
+    // Extracting the pre-placeholder side into a helper would mutate only the helper's
+    // by-value parameter copy, while the post-placeholder code would still see the original value.
+    modifier mutatedParamUsedAfter(uint256 amount) {
+        amount += 1;
+        _;
+        _payMeSubsidizedGasAfter(amount, amount);
+    }
 }
```

grandizzy (commented):
Please check findings inline

figtracer (approved):

grandizzy (approved):
lgtm

PR #15670, opened 2026-07-10
fix(anvil): mark logs of reorged blocks as removed

### LINKED ISSUE #9933: bug(`anvil`): log's `remove` field is always `false`
### Component

Anvil

- [x] Foundry
- [x] Foundryup

forge Version: 1.0.0-stable

foundryup: 1.0.1

_No response_

### Operating System

macOS (Apple Silicon)

### Describe the bug

We've been wanting to setup reorg testing for our system based on the `removed` field of logs.

However, it turns out that [`anvil` always returns `false`](https://github.com/foundry-rs/foundry/blob/c609884bdb13b9846fe9ddc5f08d99cf30c53695/crates/anvil/src/pubsub.rs#L175) for this field.

mattsse:
Anvil always returned logs with `removed: false`, even for logs from blocks that were reorged out via `anvil_reorg`. This makes it impossible to build reorg handling or reorg testing on top of the `removed` field the way geth and reth support it: when blocks are removed from the canonical chain, active log subscriptions and installed log filters should receive the removed blocks' logs again with `removed: true`, followed by the logs of the replacement blocks.

The rollback path now collects the logs of the blocks that are about to be unwound, while their transactions and receipts are still in storage, and delivers them through the existing new block notification channel as a new `ChainNotification::RemovedLogs` variant. Because removed logs and the subsequent new block notifications flow through the same channel, subscribers are guaranteed to see the removed logs first, then the new chain's logs. Log subscriptions and `eth_getFilterChanges` match the removed logs against their filter params and yield them marked as removed, while consumers that only care about imported blocks (newHeads, block filters, fee history) skip the new variant. `eth_getLogs` is unaffected and only returns canonical logs since the reorged blocks are dropped from storage.

Closes #9933

This change was implemented with AI assistance.

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -9,7 +9,7 @@ use crate::{
             self,
             db::SerializableState,
             mem::{MIN_CREATE_GAS, MIN_TRANSACTION_GAS},
-            notifications::NewBlockNotifications,
+            notifications::ChainNotifications,
             validate::TransactionValidator,
         },
         error::{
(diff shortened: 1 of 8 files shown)


grandizzy (approved):
lgtm

mablr (approved):

PR #15671, opened 2026-07-10
fix(cast): decode custom errors via local signatures cache

### LINKED ISSUE #11085: `cast call` can't decode custom errors with local artifacts cache
### Component

Cast

- [x] Foundry
- [x] Foundryup

forge Version: 1.3.0-nightly Commit SHA: 9c3feff Build Timestamp: 2025-07-24T06:25:20.174689000Z (1753338320) Build Profile: maxperf

foundryup: 1.3.0

cast call

### Operating System

macOS (Apple Silicon)

### Describe the bug

```solidity
pragma solidity ^0.8.13;

contract WTF {
    event WTF34578(uint256 a, uint256 b);

    error WTF273987(uint256 a, uint256 b);

    constructor() {}

    function wtf2890230(uint256 a, uint128 b) external {
        emit WTF34578(a, b);
        revert WTF273987(a, b);
    }
}

contract Counter {
    uint256 public number;

    constructor() {}

        number = newNumber;
        WTF wtf = new WTF();
        wtf.wtf2890230(42, 69);
    }

    function increment() public {
        number++;
    }
}
```

```
$ forge cache clean

$ forge build

...

$ cast decode-error 0x991623c600000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000002
WTF273987(uint256,uint256)
1
2

$ cast call 0x5FbDB2315678afecb367f032d93F642f64180aa3 "setNumber(uint256)" 100 --trace
Traces:
  [167180] 0x5FbDB2315678afecb367f032d93F642f64180aa3::setNumber(100)
    ├─ [108951] → new <unknown>@0xa16E02E87b7454126E5E10d957A927A7F5B5d2be
    │   └─ ← [Return] 544 bytes of code
    ├─ [2642] 0xa16E02E87b7454126E5E10d957A927A7F5B5d2be::wtf2890230(42, 69)
    │   ├─ emit WTF34578(a: 42, b: 69)
    │   └─ ← [Revert] custom error 0x991623c6: *E
    └─ ← [Revert] custom error 0x991623c6: *E


Error: Transaction failed.
Gas used: 188384
```

The custom error is not correctly decoded. (`--with-local-artifacts` can decode it correctly)

mattsse:
`cast call --trace` rendered custom errors as `custom error 0x…` even when `forge build` had already cached the error signature in the local signatures cache, while the very same trace decoded the unknown function and event signatures of that contract through this cache. The reported workaround was `--with-local-artifacts`, which compiles the project and collects its ABI errors into the revert decoder.

The root cause is that `CallTraceDecoder` falls back to the `SignaturesIdentifier` for unknown function selectors and event topics, but never for revert data: reverts were decoded exclusively by the `RevertDecoder`, whose custom error table is only filled from known contract ABIs. `SignaturesIdentifier::identify_error` existed but was only used by `cast decode-error`, which is why that command decoded the same selector fine. #11090 did not help here because it only switched the config figment to include remappings when compiling the project for `--with-local-artifacts`; the path without the flag was untouched.

Revert data now goes through the same fallback as functions and events: `RevertDecoder::maybe_decode` is split so the trace decoder can first try all known decodings (`Error(string)`, `Panic(uint256)`, cheatcode errors, registered ABI errors, plain strings) without the generic `custom error <selector>` fallback, then ask the signatures identifier for the unknown error selector, and only fall back to the generic representation if that fails as well. Error selectors of reverted traces are also included in the signature prefetching batch. With this, `cast call --trace` and `cast run` decode custom errors from the signatures cache populated by `forge build` without requiring `--with-local-artifacts`, and from OpenChain when online, consistent with how unknown functions and events are already handled.

Closes #11085

AI assistance was used to diagnose and implement this change.

### DIFF
--- a/crates/evm/traces/src/decoder/mod.rs
+++ b/crates/evm/traces/src/decoder/mod.rs
@@ -26,7 +26,7 @@ use foundry_evm_core::{
 };
 use foundry_evm_hardforks::TempoHardfork;
 use itertools::Itertools;
-use revm::bytecode::opcode::OpCode;
+use revm::{bytecode::opcode::OpCode, interpreter::InstructionResult};
 use revm_inspectors::tracing::types::{DecodedCallLog, DecodedCallTrace};
 
 use std::{collections::BTreeMap, sync::OnceLock};
@@ -610,7 +610,7 @@ impl CallTraceDecoder {
             return DecodedCallTrace {
                 label,
                 call_data: Some(DecodedCallData { signature: "create2".to_string(), args: vec![] }),
-                return_data: self.default_return_data(trace),
+                return_data: self.default_return_data(trace).await,
             };
         }
 
@@ -638,7 +638,7 @@ impl CallTraceDecoder {
                 let return_data = if trace.success {
                     None
                 } else {
-                    let revert_msg = self.revert_decoder.decode(&trace.output, trace.status);
+                    let revert_msg = self.decode_revert(&trace.output, trace.status).await;
 
                     if trace.output.is_empty() || revert_msg.contains("EvmError: Revert") {
                         Some(format!(
@@ -670,7 +670,7 @@ impl CallTraceDecoder {
                 return DecodedCallTrace {
                     label,
                     call_data: self.fallback_call_data(trace),
-                    return_data: self.default_return_data(trace),
+                    return_data: self.default_return_data(trace).await,
                 };
             };
 
@@ -687,13 +687,13 @@ impl CallTraceDecoder {
             DecodedCallTrace {
                 label,
                 call_data: Some(call_data),
-                return_data: self.decode_function_output(trace, contract_functions),
+                return_data: self.decode_function_output(trace, contract_functions).await,
             }
         } else {
             DecodedCallTrace {
                 label,
                 call_data: self.fallback_call_data(trace),
-                return_data: self.default_return_data(trace),
+                return_data: self.default_return_data(trace).await,
             }
         }
     }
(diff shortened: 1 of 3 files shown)


grandizzy (dismissed):
lgtm

crates/evm/traces/src/decoder/mod.rs
```diff
@@ -888,15 +893,35 @@ impl CallTraceDecoder {
...
+    }
+
+    /// Decodes revert data into a human-readable representation.
+    ///
+    /// If the revert decoder does not know the custom error, tries identifying the error selector
+    /// with the signatures identifier, like unknown function and event signatures. This resolves
+    /// errors from the local signatures cache populated by `forge build`, or from remote signature
+    /// databases.
+    async fn decode_revert(&self, output: &[u8], status: Option<InstructionResult>) -> String {
+        if let Some(reason) = self.revert_decoder.maybe_decode_known(output) {
```
stevencartavia, line 959: `maybe_decode_known` still includes the raw ASCII fallback, so it can return before `decode_revert` tries `identify_error`.

I reproduced this with `PrintableError18()` (`0x2e2c426f`, ASCII `.,Bo`). After `forge build`, the cache and `cast decode-error 0x2e2c426f` resolve it, but `cast call --trace` and `cast run` render:

```text
└─ ← [Revert] .,Bo
```

With `--with-local-artifacts`, the same trace renders `PrintableError18()`.

Could we move the raw string fallback after the signature-cache lookup?

decofe: Fixed in eb5a77560b01f1bf02c2448416c6149430620f92. Plain-string decoding now runs after known-error decoding, so the trace decoder can query the signatures cache first. I also added a regression using `PrintableError18()` / `0x2e2c426f` to cover the ASCII-selector case. `cargo +nightly fmt --all -- --check` passes; focused tests could not run in this sandbox because Cargo Git dependency fetches return 401.

grandizzy (approved):

stevencartavia (approved):

PR #15672, opened 2026-07-10
fix(doc): resolve same-contract references to anchor links

### LINKED ISSUE #11677: forge doc: links generated from contract/function references are incorrect
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.3.5-stable Commit SHA: 9979a41b5daa5da1572d973d7ac5a3dd2afc0221 Build Timestamp: 2025-09-09T04:49:44.505104000Z (1757393384) Build Profile: maxperf

foundryup: 1.3.0

forge doc

### Operating System

macOS (Apple Silicon)

## Function reference, same file

See the following line:
```
     * recover to arbitrary addresses for non-hashed data. A safe way to ensure
     * this is by receiving a hash of the original message (which may otherwise
     * be too long), and then calling {toEthSignedMessageHash} on it.
     */
    function recover(bytes32 hash, bytes memory signature) internal pure returns (address)
```

Source: https://github.com/OlympusDAO/olympus-v3/blob/b214bbf24fd3cf5d2d9c92dfcdc682d8721bf8db/src/external/OlympusERC20.sol#L125

`forge doc` generates this in the `src/external/OlympusERC20.sol/library.ECDSA.md` file:

```
recover to arbitrary addresses for non-hashed data. A safe way to ensure
this is by receiving a hash of the original message (which may otherwise
be too long), and then calling [toEthSignedMessageHash](/src/external/OlympusERC20.sol/library.ECDSA.md#toethsignedmessagehash) on it.*
```

The problem is the link is an absolute path, even though the heading is in the same file.

## Function reference (different format), same file

See the following line:
```
    /**
     * @dev Overload of {ECDSA-tryRecover} that receives the `r` and `vs` short-signature fields separately.
 ```

Source: https://github.com/OlympusDAO/olympus-v3/blob/b214bbf24fd3cf5d2d9c92dfcdc682d8721bf8db/src/external/OlympusERC20.sol#L188

`forge doc` generates this in the `src/external/OlympusERC20.sol/library.ECDSA.md` file:

```
*Overload of [ECDSA-tryRecover](/dependencies/chainlink-ccip-1.6.0/contracts/src/v0.8/vendor/openzeppelin-solidity/v4.8.3/contracts/utils/cryptography/ECDSA.sol/library.ECDSA.md#tryrecover) that receives the `r` and `vs` short-signature fields separately.
```

The format of `<library>-<function>` confuses it, and it links to a completely unrelated file (it's not imported in the original file). Also using an absolute reference, and a file that doesn't exist (since the docs were generated for `/src` and not `/dependencies`.

## Inheriting from contract in same file

```
interface IOHM is IERC20 {
```

Source: https://github.com/OlympusDAO/olympus-v3/blob/b214bbf24fd3cf5d2d9c92dfcdc682d8721bf8db/src/external/OlympusERC20.sol#L547

`IERC20` is defined in the same file (it's not imported). Yet `forge doc` links to a completely different file:

```
# IOHM
[Git Source](https://github.com/OlympusDAO/olympus-v3/blob/b214bbf24fd3cf5d2d9c92dfcdc682d8721bf8db/src/external/OlympusERC20.sol)

**Inherits:**
[IERC20](/src/test/lib/zuniswapv2/ZuniswapV2Pair.sol/interface.IERC20.md)
```

## Inheriting from contract in a different file

```
contract CoolerTreasuryBorrower is ICoolerTreasuryBorrower, Policy, PolicyEnabler 
```

Source: https://github.com/OlympusDAO/olympus-v3/blob/b214bbf24fd3cf5d2d9c92dfcdc682d8721bf8db/src/policies/cooler/CoolerTreasuryBorrower.sol

`forge doc` generates the following:

```
**Inherits:**
[ICoolerTreasuryBorrower](/src/policies/interfaces/cooler/ICoolerTreasuryBorrower.sol/interface.ICoolerTreasuryBorrower.md), [Policy](/src/Kernel.sol/abstract.Policy.md), [PolicyEnabler](/src/policies/utils/PolicyEnabler.sol/abstract.PolicyEnabler.md)
```

While the contracts are correct (compared to earlier examples), absolute paths are used in the links, which is different to what is used in the README.md file trees

### PR DESCRIPTION
**Pull request by mattsse**
#11677 reported four broken link patterns in `forge doc` output against the old mdbook-based pipeline. Rechecking them on current master after the solar/vocs rewrite (#14568), three no longer reproduce: `{Contract-member}` references such as `{ECDSA-tryRecover}` resolve to the same-file contract page instead of an unrelated dependency copy, since cross-reference candidates are restricted to documented sources and disambiguated by path proximity; same-file inheritance like `interface IOHM is IERC20` links to the same-file base through its exact HIR contract id; and inheritance links use the same root-relative vocs URLs as the generated sidebar, so the reported inconsistency with the summary tree no longer exists.

The remaining gap was the first reported case: a reference naming a member of the contract being rendered, like `{toEthSignedMessageHash}` inside `library ECDSA`, produced no link at all and degraded to inline code, because the global name index only contains top-level items. `{Contract-member}` self-references also went through the ambiguous name index even though the target page is by definition the one being rendered. Reference resolution now prefers lexical proximity: the members of the current contract are indexed while rendering its page, and both `{member}` and `{Contract-member}` self-references resolve to anchor-only links (`[toEthSignedMessageHash](#toethsignedmessagehash)`, `[ECDSA.tryRecover](#tryrecover)`) before any global name lookup, so a same-named contract elsewhere in the project can never win over the local definition. Unresolvable references keep rendering as inline code rather than guessing a link.

Closes #11677

### DIFF
--- a/crates/doc/src/hir_ext.rs
+++ b/crates/doc/src/hir_ext.rs
@@ -673,6 +673,13 @@ impl LocalMembers {
     fn member_anchor(&self, member: &str) -> Option<String> {
         self.members.contains(member).then(|| slug_anchor_segment(member))
     }
+
+    /// Anchor for a qualified `{Contract-member[-params...]}` reference, if `member` is
+    /// documented on this page.
+    fn xref_member_anchor(&self, part: &str) -> Option<String> {
+        let member = part.split('-').find(|piece| !piece.is_empty())?;
+        self.members.contains(member).then(|| xref_part_anchor(part))
+    }
 }
 
 /// Escape a string for use as a markdown link label.
@@ -720,7 +727,9 @@ pub fn replace_inline_links(
                 if let Some(local) = local {
                     let anchor = match part {
                         None => local.member_anchor(lookup_name),
-                        Some(member) if lookup_name == local.name => Some(xref_part_anchor(member)),
+                        Some(member) if lookup_name == local.name => {
+                            local.xref_member_anchor(member)
+                        }
                         Some(_) => None,
                     };
                     if let Some(anchor) = anchor
@@ -953,6 +962,15 @@ mod tests {
             Some(&local),
         );
         assert_eq!(out, "See `unknownMember`.");
+
+        // Unknown qualified self-reference should not create a broken same-page anchor.
+        let out = replace_inline_links(
+            "See {ECDSA-doesNotExist}.",
+            &name_to_page,
+            Path::new("src/library.ECDSA.mdx"),
+            Some(&local),
+        );
+        assert_eq!(out, "See `ECDSA`.");
     }
 
     #[test]
--- a/crates/doc/src/hir_ext.rs
+++ b/crates/doc/src/hir_ext.rs
@@ -642,6 +642,39 @@ pub(crate) fn clean_block_doc_content(raw: &str) -> String {
 
 // ── inline link replacement ───────────────────────────────────────────────────
 
+/// Members of the contract page currently being rendered.
+///
+/// Used to resolve `{member}` and `{Contract-member}` references lexically:
+/// a name that belongs to the current contract links to its heading anchor on
+/// the same page instead of going through the global name index (which only
+/// contains top-level items and could otherwise resolve to an unrelated page).
+#[derive(Debug)]
+pub struct LocalMembers {
+    /// The current contract's name.
+    name: String,
+    /// Member names with a heading (and thus an anchor) on the current page.
+    members: HashSet<String>,
+}
+
+impl LocalMembers {
+    /// Create an empty member set for the contract `name`.
+    pub fn new(name: &str) -> Self {
+        Self { name: name.to_string(), members: HashSet::new() }
+    }
+
+    /// Record a member that is rendered as a `### member` heading on the page.
+    pub fn insert(&mut self, member: &str) {
+        self.members.insert(member.to_string());
+    }
+
+    /// Anchor for a bare `{member}` reference, if `member` is documented on this page.
+    ///
+    /// Overloads share the base heading slug; the first heading owns it.
+    fn member_anchor(&self, member: &str) -> Option<String> {
+        self.members.contains(member).then(|| slug_anchor_segment(member))
+    }
+}
+
 /// Escape a string for use as a markdown link label.
 ///
 /// Prevents MDX from treating user-controlled NatSpec label text as JSX or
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: crates/doc/src/hir_ext.rs [hunk 1]
```diff
@@ -671,6 +714,29 @@ pub fn replace_inline_links(text: &str, name_to_page: &NameToPage, current_page:
                     lookup_name
                 };
 
+                // Same-contract references resolve to anchor-only links: a bare
+                // `{member}` documented on this page, or `{Contract-member}` where
+                // `Contract` is the contract being rendered.
+                if let Some(local) = local {
+                    let anchor = match part {
+                        None => local.member_anchor(lookup_name),
+                        Some(member) if lookup_name == local.name => Some(xref_part_anchor(member)),
```
grandizzy: This branch creates same-page anchors for any `{CurrentContract-anything}` reference without checking whether `anything` is actually rendered on the current page.

Bare local references are guarded by `local.member_anchor(lookup_name)`, which checks `self.members.contains(member)`, but qualified self-references skip that membership check:

```rust
Some(member) if lookup_name == local.name => Some(xref_part_anchor(member)),
```

So `{ECDSA-doesNotExist}` now becomes `[ECDSA.doesNotExist](#doesnotexist)` instead of staying unresolved/inline-code, silently producing a broken anchor.

Suggested fix: validate the first member segment before generating the anchor.

```diff
diff --git a/crates/doc/src/hir_ext.rs b/crates/doc/src/hir_ext.rs
@@
 impl LocalMembers {
@@
     fn member_anchor(&self, member: &str) -> Option<String> {
         self.members.contains(member).then(|| slug_anchor_segment(member))
     }
+
+    fn xref_member_anchor(&self, part: &str) -> Option<String> {
+        let member = part.split('-').find(|piece| !piece.is_empty())?;
+        self.members.contains(member).then(|| xref_part_anchor(part))
+    }
 }
@@
                 if let Some(local) = local {
                     let anchor = match part {
                         None => local.member_anchor(lookup_name),
-                        Some(member) if lookup_name == local.name => Some(xref_part_anchor(member)),
+                        Some(member) if lookup_name == local.name => {
+                            local.xref_member_anchor(member)
+                        }
                         Some(_) => None,
                     };
```

Regression test:

```diff
diff --git a/crates/doc/src/hir_ext.rs b/crates/doc/src/hir_ext.rs
@@
     fn same_contract_member_links_anchor_only() {
         let name_to_page = NameToPage::new();
         let mut local = LocalMembers::new("ECDSA");
         local.insert("toEthSignedMessageHash");
         local.insert("tryRecover");
@@
         let out = replace_inline_links(
             "See {unknownMember}.",
             &name_to_page,
             Path::new("src/library.ECDSA.mdx"),
             Some(&local),
         );
         assert_eq!(out, "See `unknownMember`.");
+
+        // Unknown qualified self-reference should not create a broken same-page anchor.
+        let out = replace_inline_links(
+            "See {ECDSA-doesNotExist}.",
+            &name_to_page,
+            Path::new("src/library.ECDSA.mdx"),
+            Some(&local),
+        );
+        assert_eq!(out, "See `ECDSA`.");
     }
```

If you want the unresolved qualified fallback to preserve the full `ECDSA-doesNotExist` text instead of just `ECDSA`, that fallback will need a small separate adjustment because the existing unresolved path formats `lookup_name`, not the original inner reference.

grandizzy (commented):
Please check finding inline

crates/doc/src/hir_ext.rs
decofe: Confirmed and fixed in 22440cc4: qualified same-contract references now validate the first member segment against the members rendered on the page before generating an anchor. Added the suggested unknown-qualified-reference regression case.

grandizzy (approved):

mablr (approved):

PR #15673, opened 2026-07-10
feat(cast): fetch cast run traces via debug_traceTransaction

### LINKED ISSUE #12336: Support rendering `debug_traceCall` traces in `cast`

mattsse:
`cast run` re-executes a transaction locally by replaying every prior transaction in its block, which is slow and cannot account for chain-specific EVM customizations the local executor does not implement. #15052 solved the same problem for `cast call` with `--debug-trace-call`; this adds the symmetric `cast run --debug-trace-transaction`, completing the second half of #12336.

The new flag fetches the transaction's trace from the node via `debug_traceTransaction` with the `callTracer` and renders it through the same arena conversion, decoding and rendering path as `--debug-trace-call`, skipping local execution entirely. Since the node already holds the transaction's exact pre-state and EVM rules, this needs no block replay and also handles system transactions. Label and rendering options (`--label`, `--disable-labels`, `--trace-depth`, `--with-local-artifacts`, `--json`) keep working; for `--with-local-artifacts` the runtime bytecode used for artifact matching is fetched over RPC at the transaction's block, as there is no local executor on this path. The local-execution-only flags (`--debug`, `--decode-internal`, `--trace-printer`, `--quick`, `--prestate-tracer`, `--evm-version`) conflict with the new flag, mirroring the interactions of `--debug-trace-call`. The classification of the two RPC rejections that deserve actionable hints (disabled `debug` namespace, pruned historical state) is now shared between `cast call` and `cast run` so both surface the same messages.

Closes #12336

This PR was implemented with AI assistance (code and tests generated by Claude, reviewed before submission).

### DIFF
--- a/crates/cast/src/cmd/run.rs
+++ b/crates/cast/src/cmd/run.rs
@@ -1,6 +1,8 @@
 use crate::{
     debug::handle_traces,
-    rpc_trace::{call_frame_to_arena, is_method_not_found_error, is_missing_state_error},
+    rpc_trace::{
+        call_frame_to_arena_with_root_address, is_method_not_found_error, is_missing_state_error,
+    },
     traces::TraceKind,
     utils::{apply_chain_and_block_specific_env_changes, block_env_from_header},
 };
@@ -239,8 +241,11 @@ impl RunArgs {
 
             let success = receipt.status();
             let gas_used = receipt.gas_used();
+            let root_create_address = Transaction::to(&tx).is_none().then(|| {
+                receipt.contract_address().unwrap_or_else(|| tx.from().create(tx.nonce()))
+            });
             let arena = SparsedTraceArena {
-                arena: call_frame_to_arena(&frame),
+                arena: call_frame_to_arena_with_root_address(&frame, root_create_address),
                 ignored: Default::default(),
             };
             let result = TraceResult {
@@ -254,7 +259,13 @@ impl RunArgs {
             // for the addresses in the trace, at the transaction's block. Skip the extra
             // round-trips unless local artifacts were requested.
             let contracts_bytecode = if with_local_artifacts {
-                fetch_contracts_bytecode_via_rpc(&provider, &result, tx_block_number.into()).await?
+                fetch_transaction_contracts_bytecode_via_rpc(
+                    &provider,
+                    &result,
+                    tx_hash,
+                    tx_block_number.into(),
+                )
+                .await?
             } else {
                 Default::default()
             };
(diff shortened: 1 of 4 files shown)


crates/cast/src/cmd/run.rs
```diff
@@ -174,6 +193,86 @@ impl RunArgs {
...
+                })?;
+            let GethTrace::CallTracer(frame) = geth_trace else {
+                eyre::bail!(
+                    "`debug_traceTransaction` did not return a callTracer frame; the RPC endpoint \
+                     may not support the `callTracer`"
+                );
+            };
+
+            let success = frame.error.is_none() && frame.revert_reason.is_none();
+            let gas_used = frame.gas_used.saturating_to();
```
grandizzy: `debug_traceTransaction` is tracing a mined transaction, so the final `cast run` transaction summary should use the receipt as the source of truth. The root `callTracer` frame’s `gasUsed` is call-frame/tracer output and can differ from canonical transaction gas used, e.g. around intrinsic gas/refunds or client/provider-specific tracer behavior. Keeping `frame.gas_used` in the converted trace nodes is fine, but `TraceResult.success`/`gas_used` should come from the receipt so `Transaction successfully executed` and `Gas used:` match the actual mined transaction.

```suggestion
            let receipt = provider
                .get_transaction_receipt(tx_hash)
                .await?
                .ok_or_else(|| eyre::eyre!("tx receipt not found: {:?}", tx_hash))?;

            let success = receipt.status();
            let gas_used = receipt.gas_used();
```

grandizzy (commented):
Please check finding inline

crates/cast/src/cmd/run.rs
decofe: Confirmed and fixed in 64de4687: the final success/gas summary now comes from the canonical transaction receipt, while the callTracer frame remains the source for rendered trace nodes. Added an assertion that the reported gas matches the receipt.

grandizzy (dismissed):

crates/cast/src/cmd/run.rs
```diff
@@ -174,6 +193,91 @@ impl RunArgs {
...
+                traces: Some(vec![(TraceKind::Execution, arena)]),
+                gas_used,
+            };
+
+            // Local-artifact labeling matches deployed runtime bytecode against the project
+            // artifacts. There is no local executor on this path, so fetch the code over RPC
+            // for the addresses in the trace, at the transaction's block. Skip the extra
+            // round-trips unless local artifacts were requested.
+            let contracts_bytecode = if with_local_artifacts {
+                fetch_contracts_bytecode_via_rpc(&provider, &result, tx_block_number.into()).await?
```
stevencartavia: `--debug-trace-transaction --with-local-artifacts` fetches bytecode with `eth_getCode(addr, tx.blockNumber)`, which reads the end-of-block state rather than the state at this transaction’s index.

That means artifact matching can miss labels when code exists during the traced tx but is removed/changed later in the same block. I reproduced this by mining `ping()` and then `die()`/`selfdestruct` in the same Shanghai block: local replay with `--la` rendered `PingVictim::ping()`, but this new path rendered the raw address.

The trace itself is fine; this only affects local artifact matching/labeling.

crates/cast/src/cmd/run.rs
```diff
@@ -174,6 +193,91 @@ impl RunArgs {
...
+
+            let receipt = provider
+                .get_transaction_receipt(tx_hash)
+                .await?
+                .ok_or_else(|| eyre::eyre!("tx receipt not found: {:?}", tx_hash))?;
+
+            let success = receipt.status();
+            let gas_used = receipt.gas_used();
+            let arena = SparsedTraceArena {
+                arena: call_frame_to_arena(&frame),
```
stevencartavia: Failed top-level contract creations can render as `new <unknown>@0x0000000000000000000000000000000000000000`.

I reproduced this with a reverting create tx (`0x60006000fd`). `debug_traceTransaction` produced a create frame without `to`, so `call_frame_to_arena` defaulted it to zero. Local replay rendered the derived create address instead.

Since `cast run` has the tx/receipt context here, the root create node could be patched from `receipt.contract_address()` or derived from sender + nonce before rendering.

crates/cast/src/cmd/run.rs
decofe: Confirmed and fixed in 0994f210. The transaction trace path now fetches code from a prestate trace (with storage disabled), which reflects state immediately before this transaction rather than end-of-block state. Addresses absent from prestate, such as newly created contracts, still fall back to eth_getCode.

crates/cast/src/cmd/run.rs
decofe: Confirmed and fixed in 0994f210. For root create transactions, the converted arena now fills a missing tracer address from the receipt when available, otherwise from sender.create(nonce), covering failed creations. Added a focused conversion regression test.

grandizzy (approved):

mablr (approved):

PR #15678, opened 2026-07-10
feat(debugger): show constructor arguments

figtracer:
Creation traces already carry the full initcode input, and local trace identification already strips ABI-decodable constructor arguments before matching creation bytecode. The debugger did not surface those argument values because create-frame decoding returned an empty call-data payload, while debugger nodes clear raw calldata for create frames.

This threads the constructor-argument offset discovered during local identification into the trace decoder, stores the address-scoped constructor ABI, and decodes create traces as `constructor(...)` calls. That gives the existing debugger variable/decoded-parameter path a value source for constructor parameters without changing the TUI command surface.

Progresses #927.


mattsse (approved):
lgtm

mablr (approved):

PR #15679, opened 2026-07-10
fix(anvil): preserve active block task

mablr:
`BlockListener` previously replaced its in-flight task whenever another block became available, cancelling the active task. For fork-following listeners, this could interrupt `anvil_reset` after it had begun updating the fork state.

This waits for the active task to finish before coalescing queued notifications and processing the latest block. It also waits for the final task when the stream closes and adds regression coverage for the cancellation and coalescing behavior.

### DIFF
--- a/crates/anvil/src/tasks/block_listener.rs
+++ b/crates/anvil/src/tasks/block_listener.rs
@@ -7,11 +7,15 @@ use std::{
     task::{Context, Poll},
 };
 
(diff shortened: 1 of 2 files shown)


DaniPopes (approved):

figtracer (approved):

mattsse (approved):
lgtm

PR #15680, opened 2026-07-10
feat(anvil): add TIP-20 funding RPCs

mattsse:
Anvil's Tempo mode pre-funds only its genesis accounts, while `anvil_dealERC20` cannot modify precompile-backed TIP-20 balances. This adds `tempo_fundAddress` compatibility and an `anvil_dealTip20` alias, returning one real transaction hash per requested token. Calls without an explicit token list fund PathUSD, AlphaUSD, BetaUSD, and ThetaUSD; callers can instead select any deployed TIP-20.

The helper validates the full token list before changing state, installs the local faucet admin as an issuer for custom or fork-backed tokens, tops up only missing PathUSD fee funds, and submits scoped unsigned Tempo AA mint transactions on independent random 2D nonce lanes. This avoids global impersonation state, supports concurrent calls, preserves caller token order, and makes an empty token list a true no-op.

This PR was written with Codex, including code generation, tests, and review assistance.

### DIFF
--- a/crates/anvil/core/src/eth/mod.rs
+++ b/crates/anvil/core/src/eth/mod.rs
@@ -575,6 +575,10 @@ pub enum EthRequest {
     )]
     DealERC20(Address, Address, #[serde(deserialize_with = "deserialize_number")] U256),
 
+    /// Mints TIP-20 tokens to an account using Anvil's Tempo faucet issuer.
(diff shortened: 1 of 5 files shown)


mablr (commented):
superseded by #15683

Let's postpone `tempo_fundAddress` for now.

PR #15681, opened 2026-07-10
fix(anvil): preserve Tempo header fields

mattsse:
Anvil injects Tempo timestamp and gas metadata into the RPC block wrapper's extra fields, but `eth_getHeaderByNumber` and `eth_getHeaderByHash` previously cloned only the inner header and discarded that metadata. Tempo clients decoding `TempoHeaderResponse` therefore failed because required fields such as `timestampMillis` were missing.

This preserves the wrapper fields when extracting mined and pending headers and adds integration coverage that decodes Tempo header responses through the HTTP RPC. Standard Ethereum header responses remain unchanged because their extra-field maps are empty.

Authored with AI assistance.

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -920,9 +920,15 @@ impl<N: Network> EthApi<N> {
     /// Returns block header with given hash.
     ///
     /// Handler for ETH RPC call: `eth_getHeaderByHash`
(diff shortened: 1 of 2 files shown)


0xrusowsky (approved):
makes sense

mablr (approved):
Makes sense.

Would it be a good idea to extend Alloy's `AnyNetwork` header? (breaking by wrapping Header in `WithOtherFields` :/)

figtracer (approved):

PR #15683, opened 2026-07-10
feat(anvil): add TIP-20 deal RPC

mablr:
## Motivation

Adds `anvil_dealTIP20` for directly setting the balance of a deployed TIP-20 token. The endpoint validates tokens through the TIP-20 factory and preserves deal
semantics by updating only the balance slot without changing total supply or protocol roles.

On Tempo nodes, `anvil_dealERC20` now detects TIP-20 addresses and uses the same native path, allowing existing ERC-20 tooling to work with TIP-20 tokens.


figtracer (approved):

stevencartavia (approved):

mattsse (approved):
LGTM

PR #15685, opened 2026-07-10
fix(anvil): use canonical Tempo block hashes

mattsse:
Anvil previously represented every local block with an Ethereum header and added Tempo-specific consensus fields only while serializing RPC responses. Because those fields participate in `TempoHeader` RLP hashing, the reported hash could not be reproduced from the returned header and child blocks linked to the Ethereum-only value.

This introduces a network-aware `FoundryHeader` that preserves either an Ethereum or Tempo consensus header and makes it the header type for `FoundryNetwork` and Anvil's stored blocks. Tempo fields are selected once when genesis or a new block is built; hashing, RLP, state dumps, raw block responses, and RPC conversion then derive from the stored header without propagating Tempo mode checks through downstream call sites. The regression coverage reconstructs canonical Tempo hashes from RPC and verifies reported hashes, parent links, raw RLP, and state dump persistence.

State dumps created before this change contain plain Ethereum headers. Reloading one in Tempo mode preserves those historical Ethereum-style hashes so existing transaction references remain consistent, while blocks mined after reload use canonical Tempo hashes. This intentionally permits a mixed-hash boundary for legacy dumps. The pre-existing Ethereum-shaped simulated headers returned by `eth_simulateV1` remain out of scope.

This pull request was prepared with assistance from OpenAI Codex.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -243,6 +243,11 @@ fn call_config_from_tracer_config(
 
 pub type State = foundry_evm::utils::StateChangeset;
 
+struct BlockTransactionData {
+    transactions: Vec<MaybeImpersonatedTransaction<FoundryTxEnvelope>>,
+    transaction_infos: Vec<TransactionInfo>,
+}
+
 /// A block request, which includes the Pool Transactions if it's Pending
 pub enum BlockRequest<T> {
     Pending(Vec<Arc<PoolTransaction<T>>>),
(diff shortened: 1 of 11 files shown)


crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -1052,11 +1052,20 @@ impl<N: Network> Backend<N> {
         let transactions = block.body.transactions;
 
         let hash = known_hash.unwrap_or_else(|| header.hash_slow());
-        let Header { number, withdrawals_root, .. } = header;
+        let number = header.number();
+        let withdrawals_root = header.withdrawals_root();
+        let tempo_fields = header.as_tempo().map(|header| {
```
stevencartavia: Legacy-loaded headers in Tempo mode omit the four required Tempo RPC fields because `tempo_fields` is gated on `header.as_tempo()`, causing `TempoHeaderResponse` deserialization to fail with `missing field timestampMillis`. Could we retain native values for Tempo headers and, when `self.is_tempo()`, emit the previous compatibility values for legacy Ethereum-shaped headers? This only changes the RPC projection, so the stored header and hash remain unchanged.

stevencartavia (approved):

PR #15689, opened 2026-07-11
fix: handle invariant interrupts and assertions

### LINKED ISSUE #15684: invariant testing: SIGINT ignored after long campaigns on some targets — killed process loses the end-of-run summary, silently dropping assertion failures
**Version:** `forge 1.7.2-nightly` (`907ba0817e8f`), Linux x86_64 (Ubuntu on EC2 and WSL2)

### Bug

We benchmark forge with a fixed wall-clock budget: `forge test --mc CryticToFoundry --show-progress`, then SIGINT at budget expiry to trigger the graceful early exit. On some targets, once the campaign has run ~30+ min, SIGINT (repeated ones too) is never acted on — the process fuzzes at full CPU forever and must be SIGKILLed. Same harness, 2h budgets: three targets exit gracefully, two wedge reliably (7/7 cloud instances + deterministic WSL2 repro). Reproduces with `FOUNDRY_INVARIANT_SHRINK_RUN_LIMIT=0`, so inline shrinking is ruled out.

A SIGKILLed run has no end-of-run summary in the log, which is the **only** place handler assertion failures and pre-flight-failing invariants are reported (mid-run JSON `failure` events cover only invariant violations found during the campaign) — so those findings are silently lost and the run looks clean.

Ending the campaign from the inside via `[invariant] timeout` works on the same targets: full summary printed. The wind-down is fine; the SIGINT path is the problem.

### Repro

```bash
git clone https://github.com/scfuzzbench-org/superform-v2-periphery-scfuzzbench
cd superform-v2-periphery-scfuzzbench && git checkout 4d56ecd
export FOUNDRY_DYNAMIC_TEST_LINKING=false FOUNDRY_INVARIANT_SHRINK_RUN_LIMIT=0
forge test --mc CryticToFoundry --show-progress &
sleep 2400 && kill -INT %1   # no effect, ever
```

### Asks

1. Fix the SIGINT/early-exit path for long invariant campaigns (happy to provide a gdb thread dump of a wedged process — the repro is deterministic).
2. Also emit handler assertion failures as mid-run `failure` events, so a killed process doesn't lose already-discovered findings.

stevencartavia:
Allows `SIGINT` and other early-exit requests to interrupt an invariant campaign while a transaction is still executing.

Also emits handler assertion failures as deduplicated mid-run events so discovered findings are preserved if the final summary is not written.

Fixes #15684.

### DIFF
--- a/crates/evm/evm/src/executors/invariant/mod.rs
+++ b/crates/evm/evm/src/executors/invariant/mod.rs
@@ -49,7 +49,7 @@ pub(crate) use result::did_fail_on_assert;
 use result::{assert_after_invariant, can_continue, invariant_preflight_check};
 use revm::{context::Block, state::Account};
 use serde::{Deserialize, Serialize};
-use serde_json::json;
+use serde_json::{Value, json};
 use std::{
     collections::{HashMap as Map, HashSet, btree_map::Entry},
     sync::Arc,
@@ -478,6 +478,32 @@ impl InvariantFailureMetrics {
         });
         let _ = sh_eprintln!("{}", serde_json::to_string(&event).unwrap_or_default());
     }
+
+    /// Records a handler assertion and emits a structured JSON `"failure"` event.
+    fn record_handler_failure(&mut self, target: Address, selector: Selector, reason: &str) {
+        self.broken_handlers += 1;
+
+        let timestamp =
+            SystemTime::now().duration_since(UNIX_EPOCH).map(|d| d.as_secs()).unwrap_or(0);
+        let event = build_handler_failure_event(timestamp, target, selector, reason);
+        let _ = sh_eprintln!("{}", serde_json::to_string(&event).unwrap_or_default());
+    }
+}
+
+fn build_handler_failure_event(
+    timestamp_secs: u64,
+    target: Address,
+    selector: Selector,
+    reason: &str,
+) -> Value {
+    json!({
+        "timestamp": timestamp_secs,
+        "event": "failure",
+        "failure_type": "handler_assertion",
+        "target": target,
+        "selector": selector,
+        "reason": reason,
+    })
 }
 
 /// Bridges newly-recorded invariant breaks from `failures.errors` into the pulse
@@ -1073,6 +1099,7 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
         // suite runner so parameterized `invariant_*` functions are rejected with a per-test
         // failure entry before any campaign runs.
         let config = invariant_worker_config(config, plan.worker_id, worker_count);
+        executor.inspector_mut().set_early_exit(campaign_state.early_exit().clone());
 
         let (mut invariant_test, mut corpus_manager) = Self::prepare_worker(
             &mut executor,
@@ -1154,6 +1181,9 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
                     &mut current_run.executor,
                     current_run.inputs.last().expect("checked above"),
                 )?;
+                if campaign_state.early_exit().should_stop() {
+                    break 'stop;
+                }
                 if let Some(fuzzer) = current_run.executor.inspector_mut().fuzzer.as_mut() {
                     invariant_test.fuzz_state.collect_fuzzer_values(fuzzer);
                 }
@@ -1285,6 +1315,7 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
                         };
 
                     let errors_before_check = invariant_test.test_data.failures.invariant_count();
+                    let handlers_before_check = invariant_test.test_data.failures.handler_count();
                     let (continues, broken) = if should_check_invariant {
                         let outcome = can_continue(
                             &invariant_contract,
@@ -1355,6 +1386,10 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
                             (true, None)
                         }
                     };
+
+                    if invariant_test.test_data.failures.handler_count() > handlers_before_check {
+                        campaign_state.sync_handler_failures(&invariant_test.test_data.failures);
+                    }
 
                     // Keep `cmp_seq` parallel to `inputs`: only push when the input survived the
                     // pop branch above.
@@ -2931,6 +2966,24 @@ mod tests {
         assert_eq!(payload["metrics"]["broken_invariants"], json!(2));
         assert_eq!(payload["metrics"]["broken_assertions"], json!(7));
         assert!(payload["metrics"].get("broken_handlers").is_none());
+    }
+
+    #[test]
+    fn handler_assertion_failure_event_includes_site_and_reason() {
+        let target = Address::repeat_byte(0x11);
+        let selector = Selector::from([0xde, 0xad, 0xbe, 0xef]);
+
+        assert_eq!(
+            build_handler_failure_event(123, target, selector, "assertion failed"),
+            json!({
+                "timestamp": 123,
+                "event": "failure",
+                "failure_type": "handler_assertion",
+                "target": target,
+                "selector": "0xdeadbeef",
+                "reason": "assertion failed",
+            })
+        );
     }
 
     #[test]
--- a/crates/evm/evm/src/executors/invariant/mod.rs
+++ b/crates/evm/evm/src/executors/invariant/mod.rs
@@ -609,13 +609,6 @@ struct InvariantTestData {
     optimization_best_sequence: Vec<BasicTxDetails>,
 }
 
-struct InvariantCallMetric {
-    target: Address,
-    selector: Selector,
-    reverted: bool,
-    discarded: bool,
-}
-
 /// Contains invariant test data.
 struct InvariantTest {
     // Fuzz state of invariant test.
@@ -673,28 +666,17 @@ impl InvariantTest {
     /// Update metrics for a fuzzed selector, extracted from tx details.
     /// Always increments number of calls; discarded runs (through assume cheatcodes) are tracked
     /// separated from reverts.
-    fn call_metric(
-        tx_details: &BasicTxDetails,
-        reverted: bool,
-        discarded: bool,
-    ) -> Option<InvariantCallMetric> {
-        let selector = tx_details
+    fn record_metrics(&mut self, tx_details: &BasicTxDetails, reverted: bool, discarded: bool) {
+        let Some(selector) = tx_details
             .call_details
             .calldata
             .get(..4)
             .and_then(|selector| <[u8; 4]>::try_from(selector).ok())
-            .map(Selector::from)?;
-        Some(InvariantCallMetric {
-            target: tx_details.call_details.target,
-            selector,
-            reverted,
-            discarded,
-        })
-    }
-
-    fn record_metric(&mut self, metric: InvariantCallMetric) {
-        let InvariantCallMetric { target, selector, reverted, discarded } = metric;
-        let cache_key = (target, selector);
+            .map(Selector::from)
+        else {
+            return;
+        };
+        let cache_key = (tx_details.call_details.target, selector);
 
         if let Some(metric_key) = self.test_data.metric_key_cache.get(&cache_key) {
             if let Some(invariant_metrics) = self.test_data.metrics.get_mut(metric_key) {
@@ -711,8 +693,10 @@ impl InvariantTest {
 
         // Not cached: resolve from the current target set. Unresolved keys aren't cached so a tx
         // whose target isn't known yet is re-resolved once that target is added.
-        let Some(metric_key) =
-            self.targeted_contracts.targets().fuzzed_metric_key_for_selector(target, selector)
+        let Some(metric_key) = self
+            .targeted_contracts
+            .targets()
+            .fuzzed_metric_key_for_selector(tx_details.call_details.target, selector)
         else {
             return;
         };
@@ -778,8 +762,6 @@ struct InvariantTestRun<FEN: FoundryEvmNetwork> {
     rejects: u32,
     // Whether new coverage was discovered during this run.
     new_coverage: bool,
-    // Selector metrics staged until the run is accepted.
-    metrics: Vec<InvariantCallMetric>,
     // Line coverage staged until the run is accepted.
     line_coverage: Option<HitMaps>,
     // Whether this run's inputs should become the reported last run.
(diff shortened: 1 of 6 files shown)


crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -1073,6 +1099,7 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
         // suite runner so parameterized `invariant_*` functions are rejected with a per-test
```
mattsse, line 1105: Please add a regression test that starts an invariant handler which remains in active EVM execution, triggers the shared `EarlyExit`, and asserts that the campaign terminates promptly. The current tests only cover handler-event serialization and deduplication, so the primary behavior fixed here could regress and leave `forge test` hanging indefinitely without any test failing.

mattsse (dismissed):
The interrupt path needs regression coverage before this can merge.

crates/evm/evm/src/executors/invariant/mod.rs
stevencartavia: [b6a1600](https://github.com/foundry-rs/foundry/pull/15689/commits/b6a16000b9749308b685218a193c3ba58767a796)

crates/evm/evm/src/inspectors/stack.rs
```diff
@@ -1055,6 +1065,11 @@ impl<FEN: FoundryEvmNetwork> InspectorStackRefMut<'_, FEN> {
         interpreter: &mut Interpreter,
         ecx: &mut FoundryContextFor<'_, FEN>,
     ) {
+        if self.inner.early_exit.as_ref().is_some_and(EarlyExit::should_stop) {
+            interpreter.halt(InstructionResult::Stop);
```
decofe: The executor-level active-EVM interruption is already covered by `b6a1600`; the remaining gap is only the invariant-campaign integration. Because that test manually calls `set_early_exit`, it would still pass if the campaign stopped wiring its shared `EarlyExit` into worker executors or stopped checking it after `execute_tx`. If we want that extra coverage without slowing CI, a one-worker campaign (`runs = 1`, `depth = 1`) with an infinite-loop handler and a test-only first-opcode channel/barrier would make the timing deterministic; trigger Ctrl-C once the first opcode is reached and require the campaign result within one second. On timeout, set the exit flag before joining so cleanup remains bounded.

grandizzy: test nit, we could live without it

grandizzy (approved):
makes sense

mablr (dismissed):

crates/evm/evm/src/inspectors/stack.rs
```diff
@@ -1055,6 +1065,11 @@ impl<FEN: FoundryEvmNetwork> InspectorStackRefMut<'_, FEN> {
         interpreter: &mut Interpreter,
         ecx: &mut FoundryContextFor<'_, FEN>,
     ) {
+        if self.inner.early_exit.as_ref().is_some_and(EarlyExit::should_stop) {
```
grandizzy: **[P2] Check cancellation before classifying invariant-call results**

`STOP` still runs `call_end` hooks, so an interrupted invariant predicate with a pending `vm.expectRevert()` or similar expectation can have its `Stop` rewritten into a revert. `assert_invariants` would then record a synthetic `BrokenInvariant`.

We can avoid threading `EarlyExit` through the invariant APIs by exposing a read-only query on the inspector and checking it before interpreting predicate/hook results. This also covers preflight because it uses `assert_invariants` after the flag has been installed.

```diff
diff --git a/crates/evm/evm/src/inspectors/stack.rs b/crates/evm/evm/src/inspectors/stack.rs
@@
     pub(crate) fn set_early_exit(&mut self, early_exit: EarlyExit) {
         self.early_exit = Some(early_exit);
     }

+    /// Returns whether cancellation has been requested for this execution.
+    #[inline]
+    pub(crate) fn early_exit_requested(&self) -> bool {
+        self.early_exit.as_ref().is_some_and(EarlyExit::should_stop)
+    }
+
     /// Sets the block for the relevant inspectors.

diff --git a/crates/evm/evm/src/executors/invariant/result.rs b/crates/evm/evm/src/executors/invariant/result.rs
@@
         let (call_result, success) = call_invariant_function(
             executor,
             invariant_contract.address,
             invariant.abi_encode_input(&[])?.into(),
         )?;
+        if executor.inspector().early_exit_requested() {
+            break;
+        }
         if !success {
@@
             let (inv_result, success) = call_invariant_function(
                 &invariant_run.executor,
                 invariant_contract.address,
                 invariant_contract.anchor().abi_encode_input(&[])?.into(),
             )?;
+            if invariant_run.executor.inspector().early_exit_requested() {
+                return Ok(ContinueOutcome { continues: true, broken: None });
+            }
             if success
@@
     let (call_result, success) =
         call_after_invariant_function(&invariant_run.executor, invariant_contract.address)?;
+    if invariant_run.executor.inspector().early_exit_requested() {
+        return Ok(None);
+    }
     // Fail the test case if `afterInvariant` doesn't succeed.
```

grandizzy (changes requested):
Please check edge case below

crates/evm/evm/src/inspectors/stack.rs
stevencartavia: [acfbc87](https://github.com/foundry-rs/foundry/pull/15689/commits/acfbc87a8725554e30301702bfa386f0d853bd91)

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -1073,6 +1099,7 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
...
         // failure entry before any campaign runs.
         let config = invariant_worker_config(config, plan.worker_id, worker_count);
+        executor.inspector_mut().set_early_exit(campaign_state.early_exit().clone());
```
grandizzy: **[P1] Include campaign timeout and terminal stop in active-EVM cancellation**

This installs only the global `EarlyExit` token in the inspector, while `InvariantCampaignState::should_stop()` also includes the campaign timeout and `terminal_stop`. An active transaction with a very large gas limit therefore still cannot be interrupted by `[invariant] timeout`; similarly, when one worker requests a terminal stop, another worker can remain inside EVM execution while Rayon waits for it. This leaves the timeout case described in #15684 unresolved.

Please expose a campaign-level cancellation handle to the inspector whose stop condition covers the global early exit, terminal stop, and deadline, and use that same condition for the post-`execute_tx` check.

crates/evm/evm/src/executors/invariant/result.rs
```diff
@@ -207,6 +207,9 @@ pub(crate) fn assert_invariants<'a, FEN: FoundryEvmNetwork>(
             invariant_contract.address,
             invariant.abi_encode_input(&[])?.into(),
         )?;
+        if executor.inspector().early_exit_requested() {
```
grandizzy: **[P2] Propagate cancellation instead of returning a normal invariant result**

Breaking here suppresses the synthetic invariant failure, but makes cancellation indistinguishable from a successful check. The caller can consequently persist corpus data and increment the completed-run counters even though the predicate never completed. The optimization and `afterInvariant` paths have the same behavior.

Please propagate an explicit cancelled outcome, or re-check the campaign stop condition immediately after `can_continue` and `assert_after_invariant` and exit before corpus/end-of-run accounting.

grandizzy (changes requested):
The SIGINT path is improved, but campaign-local stop conditions still cannot interrupt active EVM execution, and interrupted predicate/hook calls are accounted as completed runs.

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -1154,6 +1181,9 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
                     &mut current_run.executor,
                     current_run.inputs.last().expect("checked above"),
                 )?;
+                if campaign_state.early_exit().should_stop() {
```
mablr: This checks the campaign-global flag after `execute_tx`, so it cannot distinguish a call that was halted by the inspector from one that completed normally just before SIGINT arrived. In the latter case, a genuine handler assertion/revert is dropped before `can_continue` can record or emit it, which loses the finding this PR is intended to preserve.

crates/evm/evm/src/executors/invariant/mod.rs
stevencartavia: [343ce4b](https://github.com/foundry-rs/foundry/pull/15689/commits/343ce4b8d6747181b836ce5941eab59d6656d3a5)

crates/evm/evm/src/executors/invariant/result.rs
stevencartavia: [343ce4b](https://github.com/foundry-rs/foundry/pull/15689/commits/343ce4b8d6747181b836ce5941eab59d6656d3a5)

crates/evm/evm/src/executors/invariant/mod.rs
stevencartavia: [3bc6d82](https://github.com/foundry-rs/foundry/pull/15689/commits/3bc6d82b9e495863a04551fdb717c1c76b97152e)

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -1154,6 +1209,10 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
                     &mut current_run.executor,
                     current_run.inputs.last().expect("checked above"),
                 )?;
+                if call_result.execution_cancelled {
+                    invariant_test.test_data.failures.clone_from(&failures_checkpoint);
```
grandizzy: **[P1] Preserve completed findings when a later call is cancelled**

This unconditional checkpoint restore can erase a genuine predicate failure found earlier in the same run. For example, predicate A can fail after call 1, the multi-predicate campaign continues, and call 2 is interrupted; restoring the pre-run checkpoint then deletes predicate A.

Restore the checkpoint only if no new invariant or handler finding completed. Otherwise retain and publish those findings while still aborting the partial run.

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -1356,25 +1414,26 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
                         }
                     };
 
+                    // A predicate or optimization call can be interrupted after the handler
+                    // completed. Do not accept the partial run as a successful one.
+                    if predicate_cancelled {
+                        if completed_predicate_finding {
+                            current_run.save_last_run_inputs = true;
+                            stop_after_run = true;
```
grandizzy: **[P1] Do not accept an interrupted run just to preserve its finding**

`stop_after_run = true` preserves the finding, but falls through to corpus persistence, `end_run`, completed-call accounting, and `increment_runs`. Since a later predicate was interrupted, this reintroduces the partial-run accounting issue.

Cancellation with a completed finding needs a separate finalization path: retain and emit the finding and its reproducer, then exit before corpus/run accounting. The cancelled `afterInvariant` and generic campaign-stop branches need the same treatment.

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -1211,7 +1267,8 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
                             InvariantFuzzError::MaxAssumeRejects(config.max_assume_rejects),
                         );
                         campaign_state.request_terminal_stop();
-                        break 'stop;
+                        stop_after_run = true;
```
grandizzy: **[P1] Do not count max-assume rejection as a completed run**

Breaking only the inner loop reaches `end_run` and increments `runs`, although the configured sequence and final invariant evaluation never completed. This changes the established result from `runs: 0` to `runs: 1` and can persist the rejected prefix; it is the behavior currently failing `should_not_fail_replay_assume`.

Restore the outer-loop exit here and preserve discard-selector metrics independently rather than accepting the run.

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -1162,16 +1221,21 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
...
-                    invariant_test.record_metrics(
+                if config.show_metrics
+                    && let Some(metric) = InvariantTest::call_metric(
                         current_run.inputs.last().expect("checked above"),
                         call_result.reverted,
                         discarded,
-                    );
+                    )
+                {
+                    current_run.metrics.push(metric);
```
grandizzy: **[P2] Retain metrics for calls completed before timeout**

Selector metrics are staged until the entire run is accepted, so timeout during a later call or predicate discards metrics for earlier handler calls that completed normally. This is why `should_apply_configured_timeout` now loses its `TimeoutHandler.increment` row.

Record selector metrics after confirming that the individual handler execution was not cancelled; keep run-owned corpus, coverage, and completed-run counters staged.

grandizzy (changes requested):
The campaign cancellation signal is wired correctly now, but partial-run finalization still either drops completed findings/metrics or accepts interrupted runs as complete. The two failing invariant CLI tests expose these accounting regressions.

crates/evm/evm/src/executors/invariant/mod.rs
stevencartavia: [4e7c84e](https://github.com/foundry-rs/foundry/pull/15689/commits/4e7c84e4d18b4f7d488fe3260777a5e3fb78bf41)

crates/evm/evm/src/executors/invariant/mod.rs
stevencartavia: [4e7c84e](https://github.com/foundry-rs/foundry/pull/15689/commits/4e7c84e4d18b4f7d488fe3260777a5e3fb78bf41)

crates/evm/evm/src/executors/invariant/mod.rs
stevencartavia: [4e7c84e](https://github.com/foundry-rs/foundry/pull/15689/commits/4e7c84e4d18b4f7d488fe3260777a5e3fb78bf41)

crates/evm/evm/src/executors/invariant/mod.rs
stevencartavia: [4e7c84e](https://github.com/foundry-rs/foundry/pull/15689/commits/4e7c84e4d18b4f7d488fe3260777a5e3fb78bf41)

grandizzy (approved):
lgtm, ty!

mablr (approved):

PR #15691, opened 2026-07-11
feat(anvil): add eth_callBundle endpoint

mattsse:
Adds `eth_callBundle` support to Anvil using Alloy's MEV RPC types. Signed EIP-2718 transactions are decoded and recovered, simulated sequentially against the requested state, and returned with Flashbots-compatible per-transaction and aggregate gas, coinbase, output, and bundle-hash fields. The simulation applies the request's block context overrides without persisting state, validates required inputs and aggregate blob gas, and forwards historical fork-state requests to the upstream provider.

The execution and response behavior follows reth's implementation while reusing Anvil's network-aware transaction path so Ethereum, Optimism, and Tempo transaction handling remains consistent.

AI disclosure: This PR was written primarily with OpenAI Codex, including code generation and tests.

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -69,6 +69,7 @@ use alloy_rpc_types::{
     txpool::{TxpoolContent, TxpoolContentFrom, TxpoolInspect, TxpoolInspectSummary, TxpoolStatus},
 };
 use alloy_rpc_types_eth::{AccountInfo, Bundle, EthCallResponse, FillTransaction, StateContext};
+use alloy_rpc_types_mev::{EthCallBundle, EthCallBundleResponse};
 use alloy_serde::WithOtherFields;
 use alloy_sol_types::{SolCall, SolValue, sol};
 use alloy_transport::TransportErrorKind;
@@ -1820,6 +1821,7 @@ impl EthApi<FoundryNetwork> {
             EthRequest::EthCallMany(bundles, state_context, state_override) => {
                 self.call_many(bundles, state_context, state_override).await.to_rpc_result()
             }
+            EthRequest::EthCallBundle(bundle) => self.call_bundle(bundle).await.to_rpc_result(),
             EthRequest::EthSimulateV1(simulation, block) => {
                 self.simulate_v1(simulation, block).await.to_rpc_result()
             }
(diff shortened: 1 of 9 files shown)


figtracer (dismissed):

crates/anvil/src/eth/api.rs
```diff
@@ -2871,6 +2873,51 @@ impl EthApi<FoundryNetwork> {
...
+                "bundle missing blockNumber".to_string(),
+            )));
+        }
+
+        let block_request = self.block_request(Some(bundle.state_block_number.into())).await?;
+        if let BlockRequest::Number(number) = block_request
+            && let Some(fork) = self.get_fork()
+            && fork.predates_fork(number)
+        {
+            return Ok(fork.call_bundle(bundle).await?);
```
stevencartavia, line 2897: For a pre-fork request, `stateBlockNumber` is resolved locally for routing, but the original `safe`/`finalized` tag is forwarded upstream. If that tag has advanced upstream, the bundle can be simulated against different state. Should we forward the resolved block number here, as `eth_callMany` does?

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -4814,6 +4815,137 @@ impl<N: Network<ReceiptEnvelope = FoundryReceiptEnvelope>> Backend<N> {
...
+                let (ResultAndState { result, state }, _) = self
+                    .transact_envelope_with_inspector_ref(
+                        &cache_db,
+                        &evm_env,
+                        &mut inspector,
+                        &tx,
+                        sender,
+                    )?;
+
+                let gas_price = tx.effective_tip_per_gas(base_fee).unwrap_or_default();
```
stevencartavia, line 4896: Tempo fee-token transactions are accepted here, but `gasFees` uses the Ethereum effective tip while `coinbaseDiff` and `bundleGasPrice` use the native coinbase balance. These units do not represent the TIP20 fee payment. Should Tempo envelopes be rejected unless token-aware response semantics are supported?

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -4814,6 +4815,137 @@ impl<N: Network<ReceiptEnvelope = FoundryReceiptEnvelope>> Backend<N> {
...
+        block_request: Option<BlockRequest<FoundryTxEnvelope>>,
+    ) -> Result<EthCallBundleResponse, BlockchainError> {
+        let EthCallBundle {
+            block_number,
+            coinbase,
+            timestamp,
+            gas_limit,
+            difficulty,
+            base_fee,
+            ..
```
stevencartavia, line 4832: `timeout` is currently ignored, so the full bundle executes regardless of the requested deadline. Should we enforce it between transactions or reject unsupported timeout values?

PR #15692, opened 2026-07-11
feat(anvil): support eth_sendBundle

### LINKED ISSUE #8332: `eth_sendBundle` for MEV-related integration tests
### Component

Anvil

`eth_sendBundle` has become a very used primitive for MEV-aware application such as Private Mempool, block builders, snipers, etc. 

Being able to test this feature using Anvil would close the loop on a bunch of our integration tests.

Specs: https://docs.flashbots.net/flashbots-auction/advanced/rpc-endpoint

```
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "eth_sendBundle",
  "params": [
    {
      txs,               // Array[String], A list of signed transactions to execute in an atomic bundle
      blockNumber,       // String, a hex encoded block number for which this bundle is valid on
      minTimestamp,      // (Optional) Number, the minimum timestamp for which this bundle is valid, in seconds since the unix epoch
      maxTimestamp,      // (Optional) Number, the maximum timestamp for which this bundle is valid, in seconds since the unix epoch
      revertingTxHashes, // (Optional) Array[String], A list of tx hashes that are allowed to revert
      replacementUuid,   // (Optional) String, UUID that can be used to cancel/replace this bundle
    }
  ]
}
```

### Additional context

https://docs.flashbots.net/flashbots-auction/advanced/rpc-endpoint

mattsse:
Adds Flashbots-compatible `eth_sendBundle` request and response using Alloy's existing MEV RPC types. Signed transactions are decoded and validated into a dedicated private bundle queue with exact block and timestamp eligibility, allowed reverting transaction hashes, and replacement UUID semantics.

Bundles remain outside the public transaction pool, so their transactions do not appear in pending-transaction RPCs, txpool views, or pending subscriptions. Eligible bundles flow through automine, interval, manual, and pending-block construction as ordered atomic units; the executor checkpoints database, receipt, gas, and blob-gas state and restores the checkpoint whenever any non-allowed transaction fails or cannot be included. Consumed bundles follow transaction-pool lifecycle behavior across reset and snapshot reverts without being replayed.

Closes #8332.

AI assistance disclosure: This PR's implementation, tests, and description were produced with OpenAI Codex assistance.

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -1305,7 +1305,10 @@ impl<N: Network> EthApi<N> {
         let block_request = match &block_id {
             BlockId::Number(BlockNumber::Pending) => {
                 let pending_txs = self.pool.ready_transactions().collect();
-                BlockRequest::Pending(pending_txs)
+                BlockRequest::Pending(MiningTransactions::new(
+                    pending_txs,
+                    self.pool.private_bundles(),
+                ))
             }
             _ => {
                 let number = self.backend.ensure_block_number(Some(block_id)).await?;
@@ -1428,7 +1431,10 @@ impl EthApi<FoundryNetwork> {
         let block_request = match block_number {
             Some(BlockId::Number(BlockNumber::Pending)) => {
                 let pending_txs = self.pool.ready_transactions().collect();
-                BlockRequest::Pending(pending_txs)
+                BlockRequest::Pending(MiningTransactions::new(
+                    pending_txs,
+                    self.pool.private_bundles(),
+                ))
             }
             _ => {
                 let number = self.backend.ensure_block_number(block_number).await?;
@@ -2468,14 +2474,8 @@ impl EthApi<FoundryNetwork> {
     ) -> Result<Option<U256>> {
         node_info!("eth_getBlockTransactionCountByNumber");
         let block_request = self.block_request(Some(block_number.into())).await?;
-        if let BlockRequest::Pending(txs) = block_request {
-            let block = self
-                .backend
-                .pending_block_with_bundles(MiningTransactions::new(
-                    txs,
-                    self.pool.private_bundles(),
-                ))
-                .await;
+        if let BlockRequest::Pending(mining_transactions) = block_request {
+            let block = self.backend.pending_block_with_bundles(mining_transactions).await;
             return Ok(Some(U256::from(block.block.body.transactions.len())));
         }
         let block = self.backend.block_by_number(block_number).await?;
@@ -2752,22 +2752,22 @@ impl EthApi<FoundryNetwork> {
     pub async fn send_bundle(&self, bundle: EthSendBundle) -> Result<EthBundleHash> {
         node_info!("eth_sendBundle");
         if bundle.txs.is_empty() {
-            return Err(BlockchainError::InvalidTransactionRequest(
-                "bundle must contain at least one transaction".to_string(),
-            ));
+            return Err(BlockchainError::RpcError(RpcError::invalid_params(
+                "bundle must contain at least one transaction",
+            )));
         }
         if bundle.block_number <= self.backend.best_number() {
-            return Err(BlockchainError::InvalidTransactionRequest(format!(
+            return Err(BlockchainError::RpcError(RpcError::invalid_params(format!(
                 "bundle target block {} has already been mined",
-                bundle.block_number
-            )));
+                bundle.block_number,
+            ))));
         }
         if let (Some(minimum), Some(maximum)) = (bundle.min_timestamp, bundle.max_timestamp)
             && minimum > maximum
         {
-            return Err(BlockchainError::InvalidTransactionRequest(
-                "bundle minTimestamp exceeds maxTimestamp".to_string(),
-            ));
+            return Err(BlockchainError::RpcError(RpcError::invalid_params(
+                "bundle minTimestamp exceeds maxTimestamp",
+            )));
         }
 
         let bundle_hash = bundle.bundle_hash();
@@ -2779,9 +2779,12 @@ impl EthApi<FoundryNetwork> {
             let mut data = raw.as_ref();
             let transaction = FoundryTxEnvelope::decode_2718(&mut data)
                 .map_err(|_| BlockchainError::FailedToDecodeSignedTransaction)?;
+            if !data.is_empty() {
+                return Err(BlockchainError::FailedToDecodeSignedTransaction);
+            }
             self.ensure_typed_transaction_supported(&transaction)?;
             let pending_transaction = PendingTransaction::new(transaction)?;
-            self.backend.validate_pool_transaction(&pending_transaction).await?;
+            self.backend.validate_transaction_chain_id(&pending_transaction.transaction)?;
             let nonce = pending_transaction.nonce();
             let sender = *pending_transaction.sender();
             transactions.push(Arc::new(PoolTransaction {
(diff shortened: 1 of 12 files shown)


figtracer (dismissed):

crates/anvil/src/eth/api.rs
```diff
@@ -2739,6 +2746,65 @@ impl EthApi<FoundryNetwork> {
...
+        for raw in &bundle.txs {
+            if raw.is_empty() {
+                return Err(BlockchainError::EmptyRawTransactionData);
+            }
+            let mut data = raw.as_ref();
+            let transaction = FoundryTxEnvelope::decode_2718(&mut data)
+                .map_err(|_| BlockchainError::FailedToDecodeSignedTransaction)?;
+            self.ensure_typed_transaction_supported(&transaction)?;
+            let pending_transaction = PendingTransaction::new(transaction)?;
+            self.backend.validate_pool_transaction(&pending_transaction).await?;
```
stevencartavia, line 2784: Validating each bundle transaction against current state here can reject valid intra-bundle dependencies. I reproduced a bundle that first funded a fresh account and then spent from it; `eth_sendBundle` returned insufficient funds. Could we keep only state-independent checks at submission and perform state-dependent validation during sequential bundle execution?

crates/anvil/src/eth/pool/mod.rs
```diff
@@ -90,6 +99,41 @@ impl<T> Pool<T> {
...
+    /// Adds a private bundle, replacing an existing bundle with the same replacement UUID.
+    pub fn add_private_bundle(&self, bundle: PrivateBundle<T>) {
+        let hash = bundle.hash;
+        let mut bundles = self.private_bundles.write();
+        if let Some(replacement_uuid) = &bundle.replacement_uuid {
+            bundles.retain(|queued| queued.replacement_uuid.as_ref() != Some(replacement_uuid));
+        }
+        bundles.push(Arc::new(bundle));
+        drop(bundles);
+        Self::notify_listeners(&self.mining_listener, hash);
```
stevencartavia, line 129: A bundle targeting beyond the next block wakes the instant miner before it is eligible. At height 0, submitting one for block 10 immediately mined an empty block 1. Could we trigger instant mining only when a bundle can be included in the next block?

crates/anvil/src/eth/api.rs
```diff
@@ -4206,15 +4278,26 @@ impl EthApi<FoundryNetwork> {
     /// Returns the pending block with tx hashes
     async fn pending_block(&self) -> AnyRpcBlock {
         let transactions = self.pool.ready_transactions().collect::<Vec<_>>();
-        let info = self.backend.pending_block(transactions).await;
+        let info = self
+            .backend
+            .pending_block_with_bundles(MiningTransactions::new(
```
stevencartavia, line 4283: `eth_getBlockByNumber("pending")` includes eligible bundle transactions, but pending state is still built from public pool transactions only. I reproduced a transfer present in the pending block while `eth_getBalance(recipient, "pending")` remained equal to `latest`. Could pending block and pending state use the same bundle-aware transaction set?

crates/anvil/src/eth/backend/executor.rs
```diff
@@ -292,6 +313,36 @@ where
...
+            state: self.evm.db().read_as_state_snapshot(),
+            receipts_len: self.receipts.len(),
+            gas_used: self.gas_used,
+            blob_gas_used: self.blob_gas_used,
+        }
+    }
+
+    fn revert_bundle(&mut self, checkpoint: BundleCheckpoint) {
+        let db = self.evm.db_mut();
+        db.clear();
```
stevencartavia, line 334: Bundle rollback can leak state in fork mode. Using this commit’s binary, I submitted `[valid nonce-0 transfer, second nonce-0 transfer]` in both in-memory and fork modes. Both mined an empty block, but only the fork advanced the sender’s `latest` nonce to 1 and changed its balance. `ForkedDatabase::read_as_state_snapshot`/`init_from_state_snapshot` preserve `inner().db()`, while transaction commits land in `cache_db`, so `revert_bundle` leaves the first transaction’s cached state behind. Could the bundle checkpoint restore the active `cache_db` as well, with a fork-mode regression test asserting the nonce and balances remain unchanged?

crates/anvil/src/eth/backend/executor.rs
```diff
@@ -324,8 +397,72 @@ pub struct PoolTxGasConfig {
...
+            Transaction = FoundryTxEnvelope,
+            Evm: Evm<DB: Database + Debug, Inspector = AnvilInspector>,
+        >,
+    B::Receipt: TxReceipt,
+    <B::Result as TxResult>::HaltReason: Clone + IntoInstructionResult,
+    <B::Evm as Evm>::Tx: FromTxWithEncoded<B::Transaction> + FoundryTransaction,
+{
+    let mut executed = ExecutedPoolTransactions::default();
+
+    for bundle in private_bundles {
```
stevencartavia, line 426: In Optimism mode, an identical deposit can execute twice. I submitted the same encoded deposit in two separately accepted bundles targeting block 1; the mined block contained its transaction hash twice, the recipient balance increased by 2× the deposit value, and `eth_getTransactionReceipt(hash)` reported `transactionIndex: 0x1` because the first occurrence’s hash-keyed metadata was overwritten. Could block construction enforce transaction-hash uniqueness across the entire candidate—including private bundles and public transactions—and reject a conflicting bundle atomically?

crates/anvil/src/eth/api.rs
```diff
@@ -2739,6 +2746,65 @@ impl EthApi<FoundryNetwork> {
...
+            }));
+        }
+
+        self.pool.add_private_bundle(PrivateBundle {
+            hash: bundle_hash,
+            transactions,
+            block_number: bundle.block_number,
+            min_timestamp: bundle.min_timestamp,
+            max_timestamp: bundle.max_timestamp,
+            reverting_tx_hashes: bundle.reverting_tx_hashes.into_iter().collect(),
```
mablr, line 2801: `EthSendBundle` also deserializes `droppingTxHashes`, but that field is discarded here. For example, `[valid nonce-0, invalid nonce-0 listed as droppable, valid nonce-1]` should mine the first and third transactions; this implementation instead sees only 2/3 included and rolls the entire bundle back. Could we carry these hashes into `PrivateBundle` and checkpoint/drop listed transactions independently (or reject the unsupported field rather than accepting the request)?

crates/anvil/src/eth/api.rs
```diff
@@ -2739,6 +2746,65 @@ impl EthApi<FoundryNetwork> {
...
+        }
+
+        let bundle_hash = bundle.bundle_hash();
+        let mut transactions = Vec::with_capacity(bundle.txs.len());
+        for raw in &bundle.txs {
+            if raw.is_empty() {
+                return Err(BlockchainError::EmptyRawTransactionData);
+            }
+            let mut data = raw.as_ref();
+            let transaction = FoundryTxEnvelope::decode_2718(&mut data)
```
mablr, line 2780: This accepts trailing bytes because `decode_2718` advances `data` without requiring it to be empty. Appending one byte to a valid envelope therefore returns a bundle hash computed from the suffixed bytes, while the queued/mined transaction is the decoded canonical envelope with a different transaction hash. Could we reject the request when `!data.is_empty()` so the returned bundle hash identifies the submitted transaction bytes?

crates/anvil/src/eth/pool/mod.rs
```diff
@@ -194,6 +245,8 @@ impl<T: Transaction> Pool<T> {
     pub fn on_mined_block(self: &Arc<Self>, outcome: MinedBlockOutcome<T>) -> PruneResult<T> {
         let MinedBlockOutcome { block_number, included, invalid, not_yet_valid } = outcome;
 
+        self.prune_private_bundles(block_number);
+
         // remove invalid transactions from the pool
         self.remove_invalid(invalid.into_iter().map(|tx| tx.hash()).collect());
```
mablr, line 251: Removing invalid public transactions before pruning markers from included private transactions can delete valid dependents. If the pool has public nonce-0/nonce-1 transactions, a private replacement for nonce 0 is mined, the public nonce 0 becomes invalid, and nonce 1 is skipped only because its gas limit no longer fits this block, `remove_invalid` recursively removes both public transactions. Pruning the private nonce-0 marker afterward cannot promote nonce 1, so it disappears instead of remaining ready for the next block. Could we apply included markers first, then remove any remaining invalid hashes?

crates/anvil/src/eth/backend/executor.rs
```diff
@@ -324,8 +397,72 @@ pub struct PoolTxGasConfig {
...
+        if !bundle.is_eligible(block_number, block_timestamp) {
+            continue;
+        }
+
+        let checkpoint = executor.bundle_checkpoint();
```
mablr, line 431: This clones the entire cached database before every eligible bundle, while the new private-bundle queue has no count or aggregate-size bound. An RPC client can queue thousands of individually valid same-nonce bundles for one target block; after the first executes, each remaining bundle still clones all cached accounts/storage before failing nonce validation, making mining O(bundle count × cached state) in CPU/allocation. Could we use a transactional overlay/journal for bundle rollback, or at least bound queued bundles/transactions as an immediate mitigation?

PR #15693, opened 2026-07-11
feat(foundryup): install solar binary

DaniPopes:
Foundry release archives now ship the Solar compiler binary, but the legacy foundryup installer does not manage or activate it. This adds Solar to downloads, verification, local builds, version listing, and activation. Solar remains optional for releases created before it was added to the archives, while matching the Rust foundryup rules that reject an optional binary when it is present without an attested hash and clear stale optional files before version replacement.

This PR was written with Codex, including code generation and testing.


grandizzy (dismissed):

figtracer (dismissed):

foundryup/foundryup
```diff
@@ -260,11 +271,12 @@ main() {
     fi
 
     # Download and extract the binaries archive
-    say "downloading forge, cast, anvil, and chisel for $FOUNDRYUP_TAG version"
+    say "downloading forge, cast, anvil, chisel, and solar for $FOUNDRYUP_TAG version"
     if [ "$PLATFORM" = "win32" ]; then
       tmp="$(mktemp -d 2>/dev/null)" || err "failed to create temp dir"
       tmp="$tmp/foundry.zip"
       ensure download "$BIN_ARCHIVE_URL" "$tmp"
+      clear_version_bins "$FOUNDRY_VERSIONS_DIR/$FOUNDRYUP_TAG"
```
stevencartavia: `clear_version_bins` removes the managed files from the stored version before extraction; the active copies in `$FOUNDRY_BIN_DIR` remain, but a failed `unzip` or `tar` extraction leaves the stored version incomplete and may leave partially extracted files. Could we extract and perform any enabled verification in a temporary directory, then replace the stored binaries only after staging succeeds?

foundryup/foundryup
```diff
@@ -387,6 +402,9 @@ main() {
     # Create foundry custom version directory.
     ensure mkdir -p "$FOUNDRY_VERSIONS_DIR/$FOUNDRYUP_VERSION"
     for bin in "${BINS[@]}"; do
+      rm -f \
```
stevencartavia, line 408: `cargo build --bins` leaves artifacts for binary targets absent from the current checkout. Remote branch/PR/commit builds reuse one clone’s `target` directory, so a Solar-less revision can move a stale `target/release/solar` or `target/release/solar.exe` into its version and activate it; `--path` can likewise symlink a stale artifact. Could we remove both paths immediately before `cargo build` in both source-build paths, so Solar is present only if that build recreates it?

figtracer (approved):

emmajam (approved):

PR #15694, opened 2026-07-11
feat(debugger): jump to source lines

figtracer:
## Motivation

The debugger can jump to a program counter, but that requires users to already know the bytecode location. #927 explicitly calls out the inverse workflow: start from a Solidity source line and find the opcode mapped to it.

## Changes

- add `:line <line>` and `:ln <line>` to the existing debugger command prompt
- resolve the line against the source file currently shown, using the same source map as the source pane
- search the current trace segment first, then other segments executing the same contract code
- report invalid, out-of-range, and unmapped lines without moving the debugger
- reuse the existing PC navigation target search for source-line navigation

Progresses #927.

## Validation

- `cargo +1.95.0 test -p foundry-debugger` (108 passed)
- live TUI smoke test: `:line 8` jumped to `test/SourceLine.t.sol:8` at PC `0x3d (61)` and highlighted the mapped expression


mattsse (approved):
lgtm

stevencartavia (approved):

Ari4ka (approved):

PR #15695, opened 2026-07-11
perf(forge): speed up eip712

figtracer:
## Summary

- Resolve the target source graph directly for `forge eip712` and reuse its Solar parser, avoiding an ABI-only `solc` build that the command never consumes.
- Treat EIP-712 generation as analysis-only and leave compiler artifacts and cache untouched, including when a custom output directory is configured.
- Cover both the default path and custom `--out` behavior in the EIP-712 CLI test.

## Benchmarks

Measured release builds with `hyperfine --warmup 3 --runs 30`, removing `out` and `cache` before every sample:

```console
$ forge eip712 testdata/utils/Vm.sol --json
before  34.3 ms ± 1.9 ms
after    9.2 ms ± 0.3 ms
```

This is a **3.73x speedup** (73.2% less time). JSON output is identical. Disk output drops from 148 KiB of artifacts plus the compiler cache to no generated compiler output.

### DIFF
--- a/crates/forge/src/cmd/eip712.rs
+++ b/crates/forge/src/cmd/eip712.rs
@@ -3,6 +3,7 @@ use clap::{Parser, ValueHint};
 use eyre::Result;
 use foundry_cli::{opts::BuildOpts, utils::LoadConfig};
 use foundry_common::{compile::ProjectCompiler, shell};
+use foundry_compilers::{Graph, artifacts::Source, multi::MultiCompilerParser};
 use serde::Serialize;
 use solar::sema::{
     Gcx, Hir,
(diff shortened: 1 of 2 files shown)


crates/forge/src/cmd/eip712.rs
```diff
@@ -48,9 +49,29 @@ impl Display for Eip712Output {
 impl Eip712Args {
     pub fn run(self) -> Result<()> {
         let config = self.build.load_config()?;
-        let project = config.solar_project()?;
-        let mut output = ProjectCompiler::new().files([self.target_path]).compile(&project)?;
-        let compiler = output.parser_mut().solc_mut().compiler_mut();
+        let artifacts_requested = self.build.out_path.is_some()
```
DaniPopes: hacky, this is a random special case to have only for this command

DaniPopes:
Not really necessary for 30ms

PR #15697, opened 2026-07-11
perf(forge): speed up linting

figtracer:
## Summary

- Configure Solar directly from the project source graph for `forge lint`, avoiding the ABI-only compiler subprocess and subsequent source extraction.
- Use an ephemeral project so linting no longer writes artifacts or compiler cache entries.
- Preserve build skip filters, multi-version source selection, and the existing behavior for unsupported Solidity versions.

## Benchmarks

Measured on Solady with clean output and cache directories before every sample:

| | Mean | Range |
|---|---:|---:|
| Before | 1.340 s ± 0.014 s | 1.316–1.361 s |
| After | 68.8 ms ± 0.6 ms | 67.5–69.5 ms |

### DIFF
--- a/crates/cli/src/opts/build/utils.rs
+++ b/crates/cli/src/opts/build/utils.rs
@@ -26,6 +26,28 @@ pub fn configure_pcx(
     project: Option<&Project>,
     target_paths: Option<&[PathBuf]>,
 ) -> Result<()> {
+    configure_pcx_with_sources(pcx, config, project, target_paths, false).map(drop)
+}
+
+/// Configures a Solar parsing context with all Solar-compatible project sources.
+///
+/// Returns `true` if any compatible sources were configured.
+pub fn configure_pcx_all_sources(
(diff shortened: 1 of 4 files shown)


crates/cli/src/utils/mod.rs
```diff
@@ -627,13 +627,29 @@ ignore them in the `.gitignore` file."
         I: IntoIterator<Item = S>,
         S: AsRef<OsStr>,
     {
+        if !self.has_gitmodules_file() {
```
stevencartavia: A sparse checkout can exclude `.gitmodules` while `git submodule status` still reports uninitialized submodules. In this case, the base version restores the dependency and builds successfully, but this early return skips installation and the build fails with a missing import.

Could we avoid relying solely on the working-tree `.gitmodules` file and retain the Git status fallback?

mattsse (approved):
lgtm

mablr (approved):

PR #15699, opened 2026-07-12
fix(anvil): canonicalize blob tx roots

### LINKED ISSUE #15132: bug(`anvil`): mined blocks commit `transactionsRoot` over pooled (sidecarful) EIP-4844 encoding, not canonical execution payload

mattsse:
Closes #15132. This is a smaller alternative to #15133.

Anvil used the pooled EIP-4844 encoding both when building transaction trie leaves and when materializing mined RPC transactions. This keeps the pooled transaction in internal storage, where existing blob lookup and replay code already depend on its sidecar, but introduces canonical block-body encoding for trie calculation and drops the sidecar only from mined RPC views. The resulting transaction root and block hash are canonical, mined raw and full transaction responses no longer expose the sidecar, and the `anvil_getBlobs*` behavior remains intact.

This avoids a separate sidecar store and changes to execution, snapshots, reorgs, state serialization, and trace replay. Focused coverage checks both EIP-4844 and EIP-7594 roots, then exercises the mined RPC representation and blob retrieval end to end.

Developed with AI assistance under my direction and review.

### DIFF
--- a/crates/anvil/core/src/eth/block.rs
+++ b/crates/anvil/core/src/eth/block.rs
@@ -2,8 +2,8 @@ use super::transaction::TransactionInfo;
 #[cfg(test)]
 use alloy_consensus::Header;
 use alloy_consensus::{
-    BlockBody, EMPTY_OMMER_ROOT_HASH, Typed2718, proofs::ordered_trie_root_with_encoder,
-    transaction::RlpEcdsaEncodableTx,
+    BlockBody, EMPTY_OMMER_ROOT_HASH, TxEip4844Variant, Typed2718,
+    proofs::ordered_trie_root_with_encoder, transaction::RlpEcdsaEncodableTx,
 };
 use alloy_eips::eip2718::Encodable2718;
 use alloy_network::Network;
@@ -38,6 +38,23 @@ impl EncodableBlockTransaction for FoundryTxEnvelope {
             self.encode_2718(out);
         }
     }
+}
+
+/// Drops pooled sidecars so a transaction uses its canonical block-body representation.
+pub fn canonical_block_transaction(tx: FoundryTxEnvelope) -> FoundryTxEnvelope {
+    match tx {
+        FoundryTxEnvelope::Eip4844(tx) => {
+            FoundryTxEnvelope::Eip4844(tx.map(TxEip4844Variant::drop_sidecar))
+        }
+        tx => tx,
+    }
+}
+
+/// Returns a block whose transactions use canonical block-body representations.
+pub fn canonical_block(mut block: Block) -> Block {
+    block.body.transactions =
+        block.body.transactions.into_iter().map(|tx| tx.map(canonical_block_transaction)).collect();
+    block
 }
 
 /// Helper function to create a new block with Header and Anvil transactions, generic over the
--- a/crates/anvil/core/src/eth/block.rs
+++ b/crates/anvil/core/src/eth/block.rs
@@ -1,7 +1,10 @@
 use super::transaction::TransactionInfo;
 #[cfg(test)]
 use alloy_consensus::Header;
-use alloy_consensus::{BlockBody, EMPTY_OMMER_ROOT_HASH, proofs::calculate_transaction_root};
+use alloy_consensus::{
+    BlockBody, EMPTY_OMMER_ROOT_HASH, Typed2718, proofs::ordered_trie_root_with_encoder,
+    transaction::RlpEcdsaEncodableTx,
+};
 use alloy_eips::eip2718::Encodable2718;
 use alloy_network::Network;
 use foundry_primitives::{FoundryHeader, FoundryTxEnvelope};
@@ -20,6 +23,23 @@ pub struct BlockInfo<N: Network> {
     pub receipts: Vec<N::ReceiptEnvelope>,
 }
 
+/// A transaction that can be encoded for inclusion in a block body.
+pub trait EncodableBlockTransaction: Encodable2718 {
+    /// Encodes the transaction in its canonical block-body form.
+    fn encode_2718_for_block(&self, out: &mut dyn bytes::BufMut);
+}
+
+impl EncodableBlockTransaction for FoundryTxEnvelope {
+    fn encode_2718_for_block(&self, out: &mut dyn bytes::BufMut) {
+        if let Self::Eip4844(tx) = self {
+            out.put_u8(self.ty());
+            tx.tx().tx().rlp_encode_signed(tx.signature(), out);
+        } else {
+            self.encode_2718(out);
+        }
+    }
+}
+
 /// Helper function to create a new block with Header and Anvil transactions, generic over the
 /// transaction envelope with a default of [`FoundryTxEnvelope`].
 ///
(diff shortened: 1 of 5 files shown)


crates/anvil/core/src/eth/block.rs
```diff
@@ -30,11 +50,13 @@ pub fn create_block<T, Tx>(
...
 ) -> Block<Tx>
 where
-    Tx: Encodable2718,
+    Tx: EncodableBlockTransaction,
     T: Into<MaybeImpersonatedTransaction<Tx>>,
 {
     let transactions: Vec<_> = transactions.into_iter().map(Into::into).collect();
-    let transactions_root = calculate_transaction_root(&transactions);
+    let transactions_root = ordered_trie_root_with_encoder(&transactions, |tx, out| {
+        tx.as_ref().encode_2718_for_block(out)
```
mablr, line 75: This hashes the canonical sidecar-free encoding, but the block body still stores pooled transactions. `debug_getRawBlock` therefore serializes sidecars, so recomputing the root from the returned body does not match the header.

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5771,7 +5774,10 @@ pub fn transaction_build(
         TxEnvelope::Legacy(s) => AnyTxEnvelope::Ethereum(TxEnvelope::Legacy(rehash(s, hash))),
         TxEnvelope::Eip1559(s) => AnyTxEnvelope::Ethereum(TxEnvelope::Eip1559(rehash(s, hash))),
         TxEnvelope::Eip2930(s) => AnyTxEnvelope::Ethereum(TxEnvelope::Eip2930(rehash(s, hash))),
-        TxEnvelope::Eip4844(s) => AnyTxEnvelope::Ethereum(TxEnvelope::Eip4844(rehash(s, hash))),
+        TxEnvelope::Eip4844(s) => {
+            let s = if block.is_some() { s.map(TxEip4844Variant::drop_sidecar) } else { s };
```
mablr, line 5778: This only canonicalizes mined JSON transactions. Raw transactions/raw blocks still expose pooled sidecars, and block `size` is still measured from that pooled encoding.

crates/anvil/core/src/eth/block.rs
mattsse: i think thats fine if this includes the sidecar in the txs

grandizzy (dismissed):
lgtm

grandizzy (approved):

mablr (approved):
LGTM

Issue #6440, opened 2023-11-27
Better Hardfork -> evm SpecId handling for L2s


mattsse:
### Component

Anvil

### Describe the feature you would like

currently the Hardfork to specid conversion is only handled properly for mainnet
https://github.com/foundry-rs/foundry/blob/64320313edd81795751b8ec61d3cdcb35e419ce9/crates/anvil/src/config.rs#L805-L805

if no hardfork argument is provided:

https://github.com/foundry-rs/foundry/blob/64320313edd81795751b8ec61d3cdcb35e419ce9/crates/anvil/src/config.rs#L940-L941

we can make this smarter if we have hardfork/specid heights for L2s

## TODO
gather hardfork to evm specid blockheights for commonly used L2s/networks
* polygon
* OP
* Arbitrum
* ...

MuhtasimTanmoy:
Is this active?

zerosnacks:
I think this issue has not been resolved yet

https://github.com/foundry-rs/foundry/blob/642f13f5b0cdac4c0b83b1fecade8393febe1c33/crates/anvil/src/config.rs#L1047-L1051

Users are however, as stated, able to specify the hardfork

https://github.com/foundry-rs/foundry/blob/642f13f5b0cdac4c0b83b1fecade8393febe1c33/crates/anvil/src/config.rs#L592-L597

zerosnacks:
@mattsse

Optimism recently contributed the specifying of their hardforks here: https://github.com/foundry-rs/foundry/pull/8749

Is this sufficient or should we consider a more generalized solution

chiscookeke11:
Could I try solving this?

PR #15700, opened 2026-07-12
fix(anvil): resolve fork hardforks

mattsse:
Anvil fork startup now resolves the active hardfork through a reusable helper while preserving an explicit `--hardfork` as the highest-priority choice. Known Ethereum and OP Stack schedules are resolved from existing dependencies, while Arbitrum Nitro forks query `ArbSys.arbOSVersion()` at the pinned fork block and fall back cleanly when historical state or pre-Nitro support is unavailable. Unknown chains retain the configured default EVM version.

This keeps the integration limited to Anvil for now and addresses https://github.com/foundry-rs/foundry/issues/6440 without adding a new OP Stack or chain-registry dependency.

This PR was implemented with OpenAI Codex assisting with code generation, tests, and PR preparation.


crates/anvil/src/config.rs
```diff
@@ -1419,14 +1420,17 @@ latest block number: {latest_block}"
...
-        if self.hardfork.is_none()
-            && let Some(hardfork) =
-                FoundryHardfork::from_chain_and_timestamp(chain_id, block.header.timestamp())
-        {
-            evm_env.cfg_env.spec = SpecId::from(hardfork);
-            self.hardfork = Some(hardfork);
-        }
+        let hardfork = resolve_fork_hardfork(
+            provider.as_ref(),
+            self.hardfork,
```
mablr, line 1421: This passes the previously auto-resolved hardfork back as an explicit override. After `anvil_reset` crosses a hardfork boundary (for example ArbOS 11 -> 20), resolution returns the old hardfork without inspecting the new block. Can we keep the user-configured value separate from the active resolved value and add a reset regression test?

crates/evm/core/src/fork/hardfork.rs
```diff
@@ -0,0 +1,211 @@
...
+where
+    N: Network,
+    P: Provider<N>,
+{
+    if let Some(hardfork) = configured {
+        return hardfork;
+    }
+
+    let chain = Chain::from_id(chain_id);
+    if chain.is_ethereum()
```
mablr: This guard excludes Hoodi in the locked `alloy-chains`, even though `EthereumHardfork::from_chain_and_timestamp` has a Hoodi schedule. Pre-Prague Hoodi therefore falls back to the latest hardfork instead of Cancun. Can we apply this lookup to non-Arbitrum chains and add a Hoodi boundary test?

crates/evm/core/src/fork/hardfork.rs
```diff
@@ -0,0 +1,211 @@
...
+        && let Some(hardfork) = arbitrum_hardfork(U256::from_be_slice(&result))
+    {
+        return hardfork.into();
+    }
+
+    fallback
+}
+
+fn arbitrum_hardfork(arb_os_version: U256) -> Option<EthereumHardfork> {
+    if arb_os_version >= U256::from(50) {
```
stevencartavia, line 84: `ArbSys.arbOSVersion()` returns `55 + internal ArbOS version`, but these comparisons use internal thresholds directly. Arbitrum One block 185,000,000 returns `66` (internal version 11) and is Shanghai, yet this resolves it to Osaka. Could we checked-subtract 55, retaining the fallback on underflow, and test the ABI boundaries `66/75/95/105`?

crates/anvil/src/config.rs
```diff
@@ -1422,19 +1423,21 @@ latest block number: {latest_block}"
...
+            self.get_hardfork(),
+            chain_id,
```
mablr: Could we resolve against the remote chain ID rather than the locally overridden `chain_id`? With `--chain-id 31337` while forking mainnet, this misses the mainnet schedule and falls back to the latest spec. The local EVM chain ID and upstream chain identity need to remain separate.

crates/anvil/src/config.rs
```diff
@@ -1422,19 +1423,21 @@ latest block number: {latest_block}"
...
+            block.header.timestamp(),
+            fork_block_number,
+        )
+        .await;
+        evm_env.cfg_env.spec = SpecId::from(hardfork);
```
mablr, line 1428: Keeping the resolved fork separate from `self.hardfork` makes historical OP mining use the stale default: the OP executor still builds its env from `self.hardfork` instead of the fork-aware `self.hardfork()`.

crates/evm/core/src/fork/hardfork.rs
```diff
@@ -0,0 +1,238 @@
...
+        Some(EthereumHardfork::Osaka)
+    } else if arb_os_version >= U256::from(40) {
+        Some(EthereumHardfork::Prague)
+    } else if arb_os_version >= U256::from(20) {
+        Some(EthereumHardfork::Cancun)
+    } else if arb_os_version >= U256::from(11) {
+        Some(EthereumHardfork::Shanghai)
+    } else {
+        None
```
mablr, line 95: Pre-Shanghai Nitro versions (raw ArbSys values 56–65) reach this branch and then fall back to the current default, potentially Osaka/BPO. Should we map supported pre-11 Nitro versions or fall back to Alloy's Arbitrum schedule before using the generic latest default?

mablr (approved):
The follow-up commits address all prior review comments. The configured and resolved hardforks remain separate across resets, resolution uses the upstream chain identity, OP execution reads the active fork hardfork, and Hoodi/ArbSys boundaries are covered by focused regressions.

crates/anvil/src/config.rs
```diff
@@ -1406,35 +1396,41 @@ latest block number: {latest_block}"
...
-                FoundryHardfork::from_chain_and_timestamp(chain_id, block.header.timestamp())
-        {
-            evm_env.cfg_env.spec = SpecId::from(hardfork);
-            self.hardfork = Some(hardfork);
-        }
+        let hardfork = resolve_fork_hardfork(
+            provider.as_ref(),
+            self.hardfork,
+            self.get_hardfork(),
+            fork_chain_id,
```
stevencartavia, line 1423: The hardfork is resolved with `fork_chain_id`, but the blob parameters below are still derived from the local `chain_id`. On a Prague mainnet fork with local chain ID `31337`, the EVM resolves to Prague while `FeeManager` gets Bpo1 blob parameters (target/max 10/15 instead of 6/9), so blob fee and excess-gas calculations disagree with the active spec.

crates/anvil/src/config.rs
```diff
@@ -1537,7 +1533,7 @@ latest block number: {latest_block}"
             provider,
             chain_id,
             override_chain_id,
-            hardfork: self.hardfork,
+            hardfork: Some(hardfork),
             timestamp: block.header.timestamp(),
```
stevencartavia, line 1537: `anvil_reset(None)` clears this fork state but does not restore `evm_env.cfg_env.spec`. After resetting a Paris-era fork, `anvil_nodeInfo` reports Bpo1 while `PUSH0` still fails with `NotActivated`, showing that execution remains on Paris.

crates/anvil/src/config.rs
```diff
@@ -1406,35 +1396,41 @@ latest block number: {latest_block}"
...
+        let hardfork = resolve_fork_hardfork(
+            provider.as_ref(),
+            self.hardfork,
+            self.get_hardfork(),
+            fork_chain_id,
+            block.header.timestamp(),
+            fork_block_number,
+        )
+        .await;
+        evm_env.cfg_env.spec = SpecId::from(hardfork);
```
stevencartavia, line 1428: `FeeHistoryService` captures the initial blob parameters by value and is not updated when this resolved hardfork changes on reset. After a Cancun → Prague reset, a one-blob block reports `blobGasUsedRatio` as 1/6 instead of Prague’s 1/9.

PR #15702, opened 2026-07-13
perf(forge): avoid compiling for doc generation

figtracer:
## Summary

- Configure the bundled, in-process Solar parser directly for `forge doc`, avoiding the ABI-only compiler process when every Solidity source is Solar-compatible.
- Preserve existing behavior for pre-0.8, mixed-version, and Solidity-free projects through the compiler-backed fallback, and honor configured `skip` filters on the fast path.
- Add regression coverage proving the fast path neither invokes `solc` nor documents skipped sources, while empty and mixed-version projects retain their fallback behavior.

## Benchmarks

Benchmarked release builds on Solady with 277 Solidity files. Each sample removed `out`, `cache`, and both generated doc directories:

```console
hyperfine --warmup 1 --runs 10 \
  --prepare 'rm -rf out cache /tmp/solady-doc-before /tmp/solady-doc-after' \
  'forge-before doc --out /tmp/solady-doc-before --quiet' \
  'forge-after doc --out /tmp/solady-doc-after --quiet'
```

```text
Before: 1.412 s ± 0.032 s
After:  135.8 ms ± 8.9 ms
Speedup: 10.40x
```

The generated documentation is identical according to `diff -qr`.

### DIFF
--- a/crates/cli/src/opts/build/utils.rs
+++ b/crates/cli/src/opts/build/utils.rs
@@ -37,9 +37,16 @@ pub fn configure_pcx(
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
 pub struct SolarSourceStatus {
     /// Whether any Solar-compatible sources were configured.
-    pub has_compatible_sources: bool,
+    has_compatible_sources: bool,
     /// Whether any sources require a Solidity version unsupported by Solar.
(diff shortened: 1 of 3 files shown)


mattsse (approved):
lgtm

mablr (approved):
Cool, big gain on big projects like aave-v4.

PR #15703, opened 2026-07-13
perf(forge): speed up filtered tests

figtracer:
## Summary

- Preserve dynamic test linking while ABI-compiling the source set for filtered test runs.
- Reuse the existing preprocessed cache instead of recompiling the full project before `--match-test` and `--match-contract` runs.
- Add a regression test that ensures a cached filtered run does not invoke Solc.

## Benchmarks

Aave v4 with a warm dynamic-test-linking cache (`hyperfine --warmup 1 --runs 5`):

| Command | Before | After | Speedup |
| --- | ---: | ---: | ---: |
| `forge test --match-contract __NoSuchContract__ -q` | 4.757 s | 407.4 ms | 11.68x |

### DIFF
--- a/crates/forge/tests/cli/test_optimizer.rs
+++ b/crates/forge/tests/cli/test_optimizer.rs
@@ -1,5 +1,39 @@
 //! Tests for the `forge test` with preprocessed cache.
 
+#[cfg(unix)]
+forgetest_init!(filtered_tests_reuse_preprocessed_cache, |prj, cmd| {
+    use std::{fs, os::unix::fs::PermissionsExt};
+
+    prj.initialize_default_contracts();
+    prj.update_config(|config| config.dynamic_test_linking = true);
+    cmd.arg("build").assert_success();
+
+    let solc = prj.root().join("fake-solc");
+    let invoked = prj.root().join("fake-solc.invoked");
+    fs::write(
+        &solc,
+        r#"#!/bin/sh
+if [ "$1" = "--version" ]; then
+    echo "solc, the solidity compiler commandline interface"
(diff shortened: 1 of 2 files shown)


mattsse (dismissed):
lgtm

crates/cast/src/cmd/storage.rs
DaniPopes, line 1: not a fan of these changes, like I'd rather we not add 200 lines everywhere for every minimal edge case even if it does speed it up. also, we can switch to solar for most outputs that don't require codegen, including storage layout

figtracer: we would need a new solar release for storage layout, 0.2.0 doesn't have `Gcx::storage_layout` yet (https://github.com/paradigmxyz/solar/pull/933)

crates/forge/tests/cli/test_optimizer.rs
```diff
@@ -1,5 +1,39 @@
...
+    )
+    .unwrap();
+    let mut permissions = fs::metadata(&solc).unwrap().permissions();
+    permissions.set_mode(0o755);
+    fs::set_permissions(&solc, permissions).unwrap();
+    prj.update_config(|config| {
+        config.solc = Some(foundry_config::SolcReq::Local(solc));
+    });
+
+    cmd.forge_fuse().args(["test", "--match-contract", "CounterTest"]).assert_success();
```
mablr: Could we assert that `CounterTest` actually ran? A filtered run with zero discovered tests also exits successfully, so this can pass without proving the cached ABI produced the expected test selection.

mablr (approved):
sgtm

DaniPopes (approved):

PR #15706, opened 2026-07-13
fix(anvil): use latest active Tempo hardfork

mattsse:
## Summary

- update Foundry's Tempo dependencies to the latest pinned revision
- default local Tempo Anvil chains to the latest active Tempo hardfork
- exercise automatic hardfork selection in Tempo CI while preserving explicit historical overrides

## Context

`anvil --network tempo` uses local chain ID 31337, which is not present in Tempo's chain-aware activation schedules. The unresolved schedule previously fell back to `TempoHardfork::default()`, or T0, even though the currently active fork is T7. This disabled hardfork-gated precompile selectors, including the T2+ TIP-403 authorization methods used by ZonePortal deposits.

Tempo CI always supplied `--hardfork T7`, which masked the default local path. The fallback now reuses Foundry's existing `latest_active_tempo_hardfork()` resolver, and CI omits the override unless `TEMPO_HARDFORK` is explicitly set. Future Tempo activations therefore flow through the shared schedule without another Anvil-specific default update.

## Impact

Developers can run `anvil --network tempo` without an additional hardfork flag and get the latest valid Tempo behavior. Explicit hardfork selection remains available for historical compatibility testing.

### DIFF
--- a/.github/scripts/tempo-check.sh
+++ b/.github/scripts/tempo-check.sh
@@ -6,8 +6,8 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 # Non-verification tempo checks: local tests, fork tests, cast commands, DEX operations
 
-# Hardfork version, defaults to T7.
-HARDFORK="${TEMPO_HARDFORK:-T7}"
+# Optional hardfork override. When unset, tools resolve the latest active Tempo hardfork.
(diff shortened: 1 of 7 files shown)


mablr (commented):
nice

mablr (dismissed):

figtracer (dismissed):

grandizzy (dismissed):
makes sense

mablr (approved):

grandizzy (approved):

PR #15707, opened 2026-07-13
fix(lint): handle masked casts and denials

mablr:
## Description
Treat unsigned casts as safe when a literal bitmask bounds the result to the destination width, eliminating false positives such as `uint8(value & 0xff)`.

This also distinguishes diagnostics rejected by `deny` from genuine lint engine failures, so `forge build` no longer displays the internal-error bug-report notice for expected lint failures. The regression coverage ensures type-dependent late lints still run after recoverable Solar frontend diagnostics.

Fixes part of regression https://github.com/tempoxyz/tempo-std/actions/runs/29244890442/job/86799327897?pr=128
The remaining warnings are not safely fixable as generic Foundry false positives.
- `uint64(block.chainid)` is a real narrowing conversion. Foundry cannot assume every chain ID fits in 64 bits. Tempo must validate that invariant or suppress the lint with an explanation.
- `abi.encodePacked` warnings are intentionally conservative. Foundry sees multiple dynamic values and cannot generally prove that RLP framing or WebAuthn’s surrounding structure prevents ambiguity. Teaching the lint about those application-specific protocols would risk false negatives elsewhere.

Ref: #15611 & https://github.com/paradigmxyz/solar/pull/838

### DIFF
--- a/crates/forge/src/cmd/build.rs
+++ b/crates/forge/src/cmd/build.rs
@@ -1,7 +1,10 @@
 use super::{install, watch::WatchArgs};
 use clap::Parser;
 use eyre::Result;
-use forge_lint::{linter::Linter, sol::SolidityLinter};
+use forge_lint::{
+    linter::Linter,
+    sol::{DeniedLintDiagnostics, SolidityLinter},
+};
 use foundry_cli::{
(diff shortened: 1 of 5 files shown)


mattsse (approved):
lgtm

figtracer (approved):

PR #15709, opened 2026-07-13
fix(symbolic): allow predicate sites

stevencartavia:
Allow predicate failure-site metadata in the symbolic counterexample schema so the schema matches the artifacts Forge emits. Add coverage for all supported failure-site kinds.


mablr (approved):

mattsse (approved):
lgtm

PR #15710, opened 2026-07-13
fix(symbolic): retain original artifact

stevencartavia:
Retain the confirmed original replay artifact when invariant minimization produces an empty sequence that schema v1 cannot represent. Ensure the reported artifact can reproduce the failure.


mablr (approved):

mattsse (approved):
lgtm

grandizzy (approved):
lgtm

PR #15711, opened 2026-07-13
fix(lint): recognize guarded downcasts

decofe:
Teach `unsafe-typecast` to carry unsigned upper-bound facts across terminating conditional branches, so an explicit overflow guard proves a following downcast safe. The analysis also refines bounds inside guarded branches and invalidates them after reassignment, preserving warnings when safety is not proven.

This contribution was implemented with Codex assistance.

Prompted by: @0xalpharush

### DIFF
--- a/crates/lint/src/sol/med/unsafe_typecast.rs
+++ b/crates/lint/src/sol/med/unsafe_typecast.rs
@@ -7,7 +7,8 @@ use solar::{
     ast::{BinOpKind, LitKind, StrKind},
     sema::{
         Gcx,
-        hir::{self, ElementaryType, ExprKind, TypeKind},
+        eval::ConstantEvaluator,
+        hir::{self, ElementaryType, ExprKind, Res, StmtKind, TypeKind},
         ty::TyKind,
     },
 };
@@ -33,6 +34,7 @@ impl<'hir> LateLintPass<'hir> for UnsafeTypecast {
             && args.len() == 1
             && let Some(call_arg) = args.exprs().next()
             && is_unsafe_typecast_hir(gcx, call_arg, ty)
+            && !is_bounded_by_dominating_check(gcx, _hir, expr, call_arg, ty)
         {
             ctx.emit_with_suggestion(
                 &UNSAFE_TYPECAST,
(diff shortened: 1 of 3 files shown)


crates/lint/src/sol/med/unsafe_typecast.rs
```diff
@@ -46,6 +48,206 @@ impl<'hir> LateLintPass<'hir> for UnsafeTypecast {
...
+) -> bool {
+    let ElementaryType::UInt(size) = target else { return false };
+    let Some(var) = resolved_variable(source) else { return false };
+    let max = if size.bits() == 256 {
+        alloy_primitives::U256::MAX
+    } else {
+        (alloy_primitives::U256::from(1) << size.bits()) - alloy_primitives::U256::from(1)
+    };
+
+    hir.all_functions().any(|id| {
```
figtracer: This does not compile against Solar 0.2.0: `Hir` has no `all_functions()`, and `resolved_variable` also returns `&VariableId` at line 211 where `VariableId` is required. The compile-compatible form here is to iterate `hir.functions()` as `&Function`, and line 211 needs `Some(*id)`.

crates/lint/src/sol/med/unsafe_typecast.rs
```diff
@@ -46,6 +48,206 @@ impl<'hir> LateLintPass<'hir> for UnsafeTypecast {
...
+            BinOpKind::Ge => BinOpKind::Le,
+            _ => return false,
+        };
+        (*rhs, *lhs, reversed)
+    } else {
+        return false;
+    };
+    debug_assert_eq!(resolved_variable(value), Some(var));
+
+    let Ok(limit) = ConstantEvaluator::new(gcx).try_eval(limit) else { return false };
```
figtracer: `ConstantEvaluator` does not support `ExprKind::Member` or `ExprKind::TypeCall`, so this returns `false` for `type(uint128).max`. After repairing the compile errors locally, both new guarded-downcast fixtures still emit `unsafe-typecast`, and the UI test fails. Please recognize `type(T).min/max` explicitly or extend/reuse Solar's constant evaluation support.

crates/lint/src/sol/med/unsafe_typecast.rs
```diff
@@ -46,6 +48,206 @@ impl<'hir> LateLintPass<'hir> for UnsafeTypecast {
...
+/// out-of-range edge terminates contributes an upper-bound fact to the surviving edge. Assigning
+/// to the tracked variable invalidates that fact.
+fn is_bounded_by_dominating_check<'hir>(
+    gcx: Gcx<'hir>,
+    hir: &'hir hir::Hir<'hir>,
+    cast: &'hir hir::Expr<'hir>,
+    source: &'hir hir::Expr<'hir>,
+    target: &ElementaryType,
+) -> bool {
+    let ElementaryType::UInt(size) = target else { return false };
```
figtracer: An upper-bound proof alone is unsound for signed sources. For example, `int256 value; if (value <= 100) return uint128(value);` is considered bounded here even though negative values are not representable as `uint128`; I reproduced the warning being suppressed. Please either restrict this proof to unsigned source variables or track a lower bound and require it to be at least zero.

crates/lint/src/sol/med/unsafe_typecast.rs
```diff
@@ -46,6 +48,206 @@ impl<'hir> LateLintPass<'hir> for UnsafeTypecast {
...
+        StmtKind::AssemblyBlock(_) | StmtKind::Switch(_) | StmtKind::Try(_) => true,
+        StmtKind::If(_, then_stmt, else_stmt) => {
+            stmt_assigns_var(then_stmt, var)
+                || else_stmt.is_some_and(|stmt| stmt_assigns_var(stmt, var))
+        }
+        _ => false,
+    }
+}
+
+fn expr_assigns_var(expr: &hir::Expr<'_>, var: hir::VariableId) -> bool {
```
figtracer: Write invalidation only recognizes a root `ExprKind::Assign`. It misses `++value`, `--value`, and assignments nested inside larger expressions or conditions. With a literal `uint128`-max guard followed by `++value`, the max input becomes `2**128` and the cast truncates, but this analysis suppresses the warning. Please walk the full expression and recognize every mutation of the tracked variable.

crates/lint/src/sol/med/unsafe_typecast.rs
```diff
@@ -46,6 +48,206 @@ impl<'hir> LateLintPass<'hir> for UnsafeTypecast {
...
+}
+
+fn resolved_variable(expr: &hir::Expr<'_>) -> Option<hir::VariableId> {
+    let ExprKind::Ident([Res::Item(hir::ItemId::Variable(id))]) = expr.peel_parens().kind else {
+        return None;
+    };
+    Some(id)
+}
+
+fn stmt_always_terminates(stmt: &hir::Stmt<'_>) -> bool {
```
figtracer: This duplicates the existing shared `sol::analysis::primitives::branch_always_exits` while recognizing fewer exits: it misses builtin `revert()`, `require(false)`, and `assert(false)`. Please reuse the shared primitive so termination semantics do not diverge between lints.

figtracer (changes requested):
Requesting changes because the current patch does not compile, does not recognize the motivating `type(uint128).max` guard after the compile errors are repaired, and can suppress warnings for genuinely unsafe casts.

Architecturally, this should also avoid scanning the whole HIR once per candidate cast. Please analyze each function once and record/report the resulting facts, ideally through shared structured-HIR dataflow support under `sol::analysis`. Solar has MIR-level CFG/dominator/range machinery, but it is not currently a drop-in query API for the HIR lint stage.

Validation performed:
- `cargo check -p forge-lint` fails on the PR head.
- After minimal compile-only repairs, the UnsafeTypecast UI test still fails because the two new guarded cases continue to warn.
- Added review-only regressions confirm that signed negative values and `++value` incorrectly lose the warning.

mattsse (changes requested):
The guarded-downcast dataflow is still unsound when a cast and a mutation of the tracked variable are sibling subexpressions. For example:

    if (value > type(uint128).max) revert();
    consume(uint128(value), value = type(uint256).max);

walk_expr visits the cast before the assignment and suppresses the warning, but Solidity does not specify sibling expression evaluation order, so the assignment may execute first and the cast may truncate. I reproduced this with a UI annotation expecting unsafe-typecast; the diagnostic was missing.

One conservative fix is to pre-scan the complete expression for writes before checking or traversing any of its children, invalidate every affected bound up front, and keep those bounds invalidated after the expression. A small side-effect-only visitor or helper would avoid coupling that scan to diagnostic emission. Please add the example above as a regression test, using a non-empty helper body or disabling unrelated fixture lints.

PR #15712, opened 2026-07-13
chore: normalize benchmark definitions

stevencartavia:
Centralizes benchmark suite definitions so CI and local runs use the same repositories, options, and outputs.

Closes OSS-473


mattsse (approved):
lgtm

DaniPopes (approved):

PR #15713, opened 2026-07-13
fix(anvil): preserve fork cache on reset

### LINKED ISSUE #7966: Contract call after anvil_reset on a fork hangs indefinitely
### Component

Anvil

- [X] Foundry
- [X] Foundryup

forge 0.2.0 (a4d79ac 2024-05-10T00:18:07.559888000Z)

cast rpc

### Operating System

macOS (Apple Silicon)

### Describe the bug

After resetting a forked network to a given block, the next call to a contract fails (but after this it will succeed).

Steps to Repro:
1. Run foundry, I'm running it in docker-compose like so:
```
  anvil-eth:
    image: ghcr.io/foundry-rs/foundry:latest
    container_name: anvil-eth
    ports: [ "8550:8545" ]
    platform: linux/amd64/v8
    entrypoint: [ "anvil", "--host", "0.0.0.0", "--fork-url", $MAINNET_RPC_URL, "--steps-tracing"]
```
2. Call `anvil_reset`:
```
cast rpc --rpc-url http://localhost:8550/ anvil_reset '{"jsonRpcUrl": $MAINNET_RPC_URL,"blockNumber":19916015}'
```

3. Make a call to a contract, in this example we're calling the OneInch OffchainOracle:
```
cast call 0x0AdDd25a91563696D8567Df78D5A01C9a991F9B8 --rpc-url http://localhost:8550/ "getRateWithThreshold(address srcToken,address dstToken,bool useWrappers,uint256 thresholdFilter)(uint256)" 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48 0x6b175474e89094c44da98b954eedeac495271d0f false 50
```

4. Notice that this hangs indefinitely.
5. Notice that if you stop this call and fire the same one again, it works.

mattsse:
Fixes #7966.

`anvil_reset` rebuilt the fork database with the correct per-block cache, but then cleared the newly loaded cache. It also cleared the outgoing database before its fetched remote state could be persisted. This change separates fork metadata preparation from destructive database reset, flushes remote state before rebuilding, and retains the block-specific cache in the replacement database while continuing to discard local overlay mutations. A local-origin regression covers repeated resets to the same fork block.

This pull request was developed with AI assistance.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -2486,7 +2486,10 @@ impl<N: Network> Backend<N> {
             // reset the fork entirely and reapply the genesis config
             let reset_urls =
                 forking.json_rpc_url.as_ref().map(|url| vec![url.clone()]).unwrap_or_default();
-            fork.reset(reset_urls, block_number).await?;
+            // Persist fetched remote state before rebuilding the fork database so the new
+            // block-specific database can load it from disk.
+            fork.database.read().await.maybe_flush_cache().map_err(BlockchainError::Internal)?;
+            fork.prepare_reset(reset_urls, block_number.into()).await?;
             let fork_block_number = fork.block_number();
             let fork_block = fork
                 .block_by_number(fork_block_number)
@@ -2549,7 +2552,6 @@ impl<N: Network> Backend<N> {
                 fork.total_difficulty(),
             );
             self.states.write().clear();
-            self.db.write().await.clear();
 
             self.apply_genesis().await?;
 
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -2552,7 +2552,6 @@ impl<N: Network> Backend<N> {
                 fork.total_difficulty(),
             );
             self.states.write().clear();
-
             self.apply_genesis().await?;
 
             trace!(target: "backend", "reset fork");
--- a/crates/anvil/tests/it/fork.rs
+++ b/crates/anvil/tests/it/fork.rs
@@ -1779,6 +1779,40 @@ async fn test_reset_updates_cache_path_when_rpc_url_not_provided() {
 }
 
 #[tokio::test(flavor = "multi_thread")]
+async fn test_fork_reset_reuses_cached_remote_state() {
+    let address = Address::random();
+    let balance = U256::from(1337u64);
+    let chain_id =
+        u64::from_be_bytes(address.as_slice()[12..].try_into().unwrap()) % 1_000_000 + 1_000_000;
+    let cache_dir = Config::foundry_chain_cache_dir(chain_id).unwrap();
+    let _ = std::fs::remove_dir_all(&cache_dir);
+
+    let origin_config = NodeConfig::test()
+        .with_chain_id(Some(chain_id))
+        .with_funded_accounts([(address, balance)].into_iter().collect());
+    let (_origin_api, origin_handle) = spawn(origin_config).await;
+    let fork_config = NodeConfig::test()
+        .with_chain_id(Some(chain_id))
+        .with_eth_rpc_url(Some(origin_handle.http_endpoint()));
+    let (api, handle) = spawn(fork_config).await;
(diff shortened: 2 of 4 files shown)


stevencartavia (dismissed):

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -2486,7 +2486,10 @@ impl<N: Network> Backend<N> {
             // reset the fork entirely and reapply the genesis config
             let reset_urls =
                 forking.json_rpc_url.as_ref().map(|url| vec![url.clone()]).unwrap_or_default();
-            fork.reset(reset_urls, block_number).await?;
+            // Persist fetched remote state before rebuilding the fork database so the new
+            // block-specific database can load it from disk.
+            fork.database.read().await.maybe_flush_cache().map_err(BlockchainError::Internal)?;
+            fork.prepare_reset(reset_urls, block_number.into()).await?;
```
grandizzy, line 2492: This reload only works while the current `evm_env.block_env` still matches the original fork block. After mining, the local base fee differs from the cached metadata; `setup_fork_db_config` preserves it, so `BlockchainDb::new` rejects the flushed cache as non-matching. Could cache metadata be constructed from the canonical fork header separately from local execution overrides?

crates/anvil/tests/it/fork.rs
```diff
@@ -1778,6 +1778,40 @@ async fn test_reset_updates_cache_path_when_rpc_url_not_provided() {
...
+    let fork_config = NodeConfig::test()
+        .with_chain_id(Some(chain_id))
+        .with_eth_rpc_url(Some(origin_handle.http_endpoint()));
+    let (api, handle) = spawn(fork_config).await;
+    let provider = handle.http_provider();
+    let fork_block_number = api.anvil_node_info().await.unwrap().fork_config.fork_block_number;
+
+    assert_eq!(provider.get_balance(address).await.unwrap(), balance);
+
+    for _ in 0..2 {
```
grandizzy, line 1804: Please mine a block before this reset loop. The current test leaves `BlockEnv` unchanged, so it misses the common workflow where local state is modified before `anvil_reset`. Adding `api.mine_one().await` here should expose the cache-metadata mismatch.

grandizzy (dismissed):
**Request changes:** The ordering fix looks correct, but cache reuse still fails after mining because reset preserves the locally updated `BlockEnv`, causing the rebuilt DB to reject the original fork cache. Please derive cache metadata from the canonical fork block and add coverage that mines before resetting.

decofe:
The requested change is valid: cache metadata was rebuilt from the locally advanced `BlockEnv`, so mining could make the flushed cache fail metadata comparison. I pushed the fix and mine-before-reset coverage in 18fbcf624378e8a23298a04bf903b3cb27fa258b and opened mattsse/foundry#2 against this PR branch. `cargo +nightly fmt --all -- --check` and `git diff --check` pass; the targeted test could not compile locally because Cargo dependency fetches currently return HTTP 401.

grandizzy (approved):
lgtm

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -2486,7 +2486,10 @@ impl<N: Network> Backend<N> {
             // reset the fork entirely and reapply the genesis config
             let reset_urls =
                 forking.json_rpc_url.as_ref().map(|url| vec![url.clone()]).unwrap_or_default();
-            fork.reset(reset_urls, block_number).await?;
+            // Persist fetched remote state before rebuilding the fork database so the new
+            // block-specific database can load it from disk.
+            fork.database.read().await.maybe_flush_cache().map_err(BlockchainError::Internal)?;
```
mablr, line 2491: Follow-up needed.

This exposes a pre-existing cache-identity issue: hosts are ignored and `BlockEnv` omits the block hash/state root. Because this now flushes origin A before switching URLs, resetting to origin B at the same chain ID/height can reload A’s state.

Persisted-cache reuse should be bypassed when `json_rpc_url` changes. Canonical block-hash validation can be addressed separately.

mablr (approved):
LGTM, but a follow-up is needed.

PR #15715, opened 2026-07-13
perf(anvil): stabilize long-chain mining

mattsse:
Anvil rebuilt and sorted the complete account and storage tries for every mined block, while historical state retention cloned and eventually serialized the entire accumulated state; EIP-2935 made both paths progressively more expensive as its history contract filled. This change maintains the state root from dirty paths with a local mutable Merkle Patricia trie built on Alloy's canonical node encoders, without adding a Reth dependency; it also stores historical states as structurally shared snapshots and bounds the EVM block-hash cache to the current hash plus its 256-block window, preserving historical calls, dump/load, and reorg behavior without chain-height-dependent work. Codex was used for investigation, implementation, testing, and this PR description; AI-generated responses may also be used during review.

### Results

| 2,500 transaction-bearing blocks | master | this PR | change |
| --- | ---: | ---: | ---: |
| Total wall time | 3.857s | 0.826s | 4.67x faster |
| First 100 throughput | 2,479 blocks/s | 2,340 blocks/s | comparable |
| Last 100 throughput | 385 blocks/s | 3,153 blocks/s | 8.19x faster |

### DIFF
--- a/crates/anvil/src/eth/backend/db.rs
+++ b/crates/anvil/src/eth/backend/db.rs
@@ -13,7 +13,7 @@ use alloy_eips::eip4895::Withdrawals;
 use alloy_network::Network;
 use alloy_primitives::{
     Address, B256, Bytes, U256, keccak256,
-    map::{AddressMap, HashMap},
+    map::{AddressMap, HashMap, U256Map},
 };
 use alloy_rpc_types::BlockId;
 use anvil_core::eth::{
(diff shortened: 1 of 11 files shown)


stevencartavia (dismissed):

crates/anvil/src/eth/backend/db.rs
```diff
@@ -42,12 +42,37 @@ use serde_json::Value;
...
+/// addition to the 256 preceding hashes is necessary for calls against the latest block.
+pub(crate) fn expired_block_hash(number: U256) -> Option<U256> {
+    let cache_size = U256::from(BLOCKHASH_HISTORY + 1);
+    (number >= cache_size).then(|| number - cache_size)
+}
+
+/// Inserts a block hash while discarding the entry that left the EVM-visible cache.
+pub(crate) fn cache_block_hash(block_hashes: &mut U256Map<B256>, number: U256, hash: B256) {
+    if let Some(expired) = expired_block_hash(number) {
+        block_hashes.remove(&expired);
```
grandizzy: **Keep the cache bounded across block-number jumps**

Removing only `number - 257` assumes hashes are inserted contiguously. Anvil supports block-number jumps and state/head replacement, so sparse inserts such as `0, 258, 516, ...` never remove an existing key and the cache continues growing.

Please prune all entries outside the active `[head - 256, head]` window, or clear and reseed the cache when the head is replaced. The corresponding logic in `HistoricalStateCache::record_block_hash` should follow the same rule. A sparse/out-of-order insertion test would cover this case.

```suggestion
    let min_number = number.saturating_sub(U256::from(BLOCKHASH_HISTORY));
    block_hashes.retain(|cached, _| *cached >= min_number && *cached <= number);
    block_hashes.insert(number, hash);
```

The exact implementation may need to distinguish a forward head advance from historical/out-of-order insertion, but it should not rely on every intermediate block number being present.

grandizzy (changes requested):
The incremental trie and structurally shared history approach looks sound. The block-hash eviction logic should handle non-contiguous block numbers before merging.

grandizzy (dismissed):
lgtm

crates/anvil/src/eth/backend/db.rs
```diff
@@ -42,12 +42,30 @@ use serde_json::Value;
 
 use crate::mem::storage::MinedTransaction;
 
+/// Number of preceding block hashes available to the EVM's `BLOCKHASH` opcode.
+pub(crate) const BLOCKHASH_HISTORY: u64 = 256;
+
+/// Inserts a block hash while discarding the entry that left the EVM-visible cache.
+pub(crate) fn cache_block_hash(block_hashes: &mut U256Map<B256>, number: U256, hash: B256) {
+    let head = block_hashes.keys().copied().max().map_or(number, |head| head.max(number));
```
mablr, line 50: This seems to make the largest cached key the permanent head, which breaks loading an older dump into a nonempty node. For example, after mining to 1,000 and loading a dump at 500, the existing head keeps `min_number` at 744, so the loaded `244..=500` hashes are rejected and `BLOCKHASH(499)` returns zero.

mablr (approved):

grandizzy (approved):

PR #15716, opened 2026-07-13
fix(anvil): align eth_simulateV1 behavior

### LINKED ISSUE #13823: eth_simulateV1: spec conformance fixes (86/91 tests passing)
## Summary

`eth_simulateV1` (added in #10227, #11253) has several spec conformance issues when tested against the [official ethereum/execution-apis test suite](https://github.com/ethereum/execution-apis/tree/main/tests/eth_simulateV1). Before fixes: **46/91 passed**. After fixes: **86/91 passed** (94.5% conformance).

### 1. Wrong error codes (fixed 14 tests)
The spec defines custom error codes for `eth_simulateV1` that anvil was not using:
- `-38012` BaseFee too low → was returning `-32003`
- `-38013` Intrinsic gas too low → was returning `-32000`
- `-38014` Insufficient funds → was returning `-32003`
- `-38020` Block numbers not in order → not validated
- `-38021` Block timestamps not in order → not validated
- `-38022` movePrecompileToAddress self-reference → not validated
- `-38026` Too many blocks → was returning `-32603`

### 2. Transaction build failure on missing `to` field (fixed 9 tests)
When a call has no `to` field (empty `{}` or create-style tx), EVM execution succeeds but post-execution `build_unsigned()` fails because it requires `to`. Fixed by setting `TxKind::Create` when `to` is missing.

### 3. Missing block number/timestamp ordering validation (fixed 8 tests)
The spec requires:
- Block numbers must be strictly increasing (error `-38020`)
- Block timestamps must be strictly increasing (error `-38021`)
- Intermediate empty blocks generated when block number gaps exist

### 4. `validation: true` mode broken (fixed in prior commit)
`build_call_env()` unconditionally sets `disable_nonce_check = true`, `disable_base_fee = true`, `disable_block_gas_limit = true`. When `validation: true`, these should be re-enabled per the spec.

### 5. Missing movePrecompileToAddress validation (fixed 3 tests)
- Source must be a precompile (0x01-0x09), else error `-32000`
- Source ≠ destination, else error `-38022`

### 6. Future block returns wrong error code (fixed 2 tests)
When block parameter references a block beyond head, should return `-32000` "header not found", was returning `-32602`.

### 7. parent_hash always zero in multi-block simulation (fixed)
Multi-block simulation had `parent_hash: Default::default()` for all blocks instead of chaining from previous block.

### 8. Revert error missing output data (fixed)
`SimulateError` was missing the `data` field with revert bytes.

## Remaining 5 Failures

| Test | Root Cause |
|------|-----------|
| `blocknumber-increment` | Timestamp env-dependent: spec expects genesis-based timestamps (~462) but anvil uses wallclock |
| `use-as-many-features-as-possible` | Same timestamp issue |
| `eth-send-should-produce-more-logs-on-forward` | CALL+value forwarding behavior difference |
| `overflow-nonce-validation` | revm returns NonceTooLow instead of NonceMaxValue on overflow |
| `run-out-of-gas-in-block-38015` | Per-block gas accounting not enforced in simulation mode |

## Spec Test Runner

Included a Python runner script that downloads and runs all 91 official spec tests against a running anvil instance. Can be used as CI regression gate.

## How to reproduce

```bash
# Start anvil and mine 45 blocks to match spec test expectations
anvil --port 18547 &
cast rpc anvil_mine 0x2d --rpc-url http://localhost:18547
# Run spec tests
python3 crates/anvil/tests/it/spec_eth_simulateV1/run_spec_tests.py http://localhost:18547
```

mattsse:
Fixes #13823.

`eth_simulateV1` diverged from Geth and the execution-apis fixtures because simulated blocks inherited ordinary transaction defaults and reused validation, gas, and log behavior that does not apply to multi-block simulation. This aligns block numbering, timestamps, parent hashes, base-fee defaults, per-block and request-wide gas accounting, nonce handling, validation errors, transfer tracing, revert results, contract creation, and moved-precompile warming with the endpoint semantics.

The change also adds endpoint-level HTTP JSON-RPC coverage for block sequencing, gas exhaustion, nonce overflow, validation failures, creation and reverts, forwarded and SELFDESTRUCT transfers, reverted log-index gaps, request gas caps, and precompile relocation.

AI assistance was used to investigate the execution-apis fixtures and Geth behavior, implement the code and tests, and draft this description. The changes were reviewed and verified locally.


crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5039,7 +5128,15 @@ impl Backend<FoundryNetwork> {
                 let mut log_index = 0;
                 let mut gas_used = 0;
                 let mut transactions = Vec::with_capacity(calls.len());
-                let mut logs= Vec::new();
+                let mut logs = Vec::new();
+                if !validation {
```
stevencartavia, line 5269: `block_env` is reused across simulated blocks, so a `prevRandao` override remains active when the next block omits it. Execution-apis and Geth default it to zero per block. Could we reset `block_env.prevrandao` before applying each block's overrides?

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5148,19 +5349,19 @@ impl Backend<FoundryNetwork> {
...
-                let transactions_envelopes: Vec<AnyTxEnvelope> = transactions
-                .iter()
-                .map(|tx| AnyTxEnvelope::from(tx.clone()))
-                .collect();
+                let transactions_envelopes: Vec<AnyTxEnvelope> =
+                    transactions.iter().map(|tx| AnyTxEnvelope::from(tx.clone())).collect();
+                let blob_gas_used =
+                    transactions_envelopes.iter().filter_map(|tx| tx.blob_gas_used()).sum();
                 let header = Header {
                     logs_bloom: logs_bloom(logs.iter()),
```
stevencartavia: This includes synthetic `0xeeee…` ETH transfer logs in `logsBloom`. The official fixture expects those logs in `calls[].logs` but excludes them from the bloom. Could we calculate it from the EVM/receipt logs before synthetic logs are appended?

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5097,14 +5266,21 @@ impl Backend<FoundryNetwork> {
                     // commit the transaction
                     cache_db.commit(state);
                     gas_used += result.tx_gas_used();
+                    rpc_gas_budget = rpc_gas_budget.saturating_sub(result.tx_gas_used());
 
                     // create the transaction from a request
                     let from = request.from.unwrap_or_default();
 
-                    let mut request = Into::<FoundryTransactionRequest>::into(WithOtherFields::new(request));
+                    let mut request =
```
stevencartavia, line 5427: With `returnFullTransactions: true`, the constructed transactions have null `blockHash`, `blockNumber`, and `transactionIndex`, while `blockTimestamp` is absent. Could we populate them once the simulated block hash is known?

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5059,17 +5201,25 @@ impl Backend<FoundryNetwork> {
...
 
-                    if !validation {
-                        evm_env.cfg_env.disable_base_fee = !validation;
+                    if validation {
+                        evm_env.cfg_env.disable_nonce_check = false;
+                        evm_env.cfg_env.disable_base_fee = false;
+                        evm_env.cfg_env.disable_block_gas_limit = false;
+                    } else {
+                        evm_env.cfg_env.disable_base_fee = true;
                         evm_env.block_env.basefee = 0;
```
stevencartavia: With `validation: false`, an explicit `baseFeePerGas` override is preserved in the returned header, but `BASEFEE` still returns zero. Could we preserve the override in the EVM context while keeping transaction base-fee validation disabled?

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5173,8 +5374,13 @@ impl Backend<FoundryNetwork> {
                     nonce: Default::default(),
                     base_fee_per_gas: Some(block_env.basefee),
                     withdrawals_root: None,
-                    blob_gas_used: None,
-                    excess_blob_gas: None,
+                    blob_gas_used: Some(blob_gas_used),
+                    excess_blob_gas: Some(
```
stevencartavia: A `blobBaseFee` override remains active in the next simulated block when that block omits it. I reproduced block 1 overriding it to `21` and block 2 still observing `21` through `BLOBBASEFEE`. Could we resolve the blob base fee independently for each block instead of carrying over the previous override?

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5147,20 +5354,19 @@ impl Backend<FoundryNetwork> {
...
-                .map(|tx| AnyTxEnvelope::from(tx.clone()))
-                .collect();
+                let transactions_envelopes: Vec<AnyTxEnvelope> =
+                    transactions.iter().map(|tx| AnyTxEnvelope::from(tx.clone())).collect();
+                let blob_gas_used =
+                    transactions_envelopes.iter().filter_map(|tx| tx.blob_gas_used()).sum();
                 let header = Header {
                     logs_bloom: logs_bloom(logs.iter()),
                     transactions_root: calculate_transaction_root(&transactions_envelopes),
                     receipts_root: calculate_receipt_root(&transactions_envelopes),
```
mablr: `calculate_receipt_root` is being fed transaction envelopes, so every non-empty simulated block reports a receipts trie made from transactions rather than from status/cumulative-gas/log receipts. The same header also leaves `state_root` and `mix_hash` at zero even after state changes or a `prevRandao` override, then hashes that header for full transaction metadata, the next `parentHash`, and `BLOCKHASH`. Geth creates a receipt for every call and finalizes the post-state before assembling/hashing the block (`internal/ethapi/simulate.go`, `processBlock`, lines 296-360 and 390-425). Could we accumulate real receipts/canonical logs and finalize all execution-dependent header fields before computing this hash? The blob fields should also remain absent before Cancun instead of being populated unconditionally.

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5117,21 +5296,49 @@ impl Backend<FoundryNetwork> {
...
+                                code: alloy_rpc_types::simulate::SimulateError::VM_EXECUTION_ERROR_CODE,
+                                message,
+                                data: None,
+                            })
+                        }
+                    };
+                    let result_logs =
+                        if result.is_success() { result.clone().into_logs() } else { Vec::new() };
+                    logs.extend(
+                        result_logs.iter().filter(|log| !is_simulated_transfer_log(log)).cloned(),
```
mablr: This distinguishes synthetic transfer logs by address/topic content, which can also match a legitimate EVM log. A state override can put code at `0xeeee…` that emits the `Transfer` topic; that canonical log would then be incorrectly removed from `logsBloom` (and from a corrected receipt root). Geth keeps canonical receipt logs separate from the tracer’s response-only transfer logs. Could we preserve provenance/separate streams instead of filtering by log contents?

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -635,6 +643,47 @@ impl<N: Network> Backend<N> {
...
     }
 
+    fn simulation_precompile_moves(
+        &self,
+        state_overrides: &StateOverride,
+    ) -> Result<Vec<(Address, Address)>, BlockchainError> {
+        let moves = state_overrides
+            .iter()
+            .filter_map(|(source, account)| {
+                account.move_precompile_to.map(|destination| (*source, destination))
```
mablr, line 662: Considering only entries with `movePrecompileToAddress` misses two Geth state-override semantics: overriding an active precompile with ordinary code must remove the precompile even when no move is requested, and a move destination that is also present in the state override map must be rejected. As written, code installed at (for example) `0x04` is ignored by the still-active precompile, while a destination override is silently shadowed by the moved precompile. Also, the dispatch at lines 5236-5253 silently bypasses non-empty moves on OP/Tempo. Could we derive the precompile map from the full override set, reject destination collisions, and either apply or explicitly reject moves on unsupported network paths?

mablr (changes requested):
The five earlier review threads are addressed by the latest commit, so I resolved them. I found three remaining compatibility issues in block finalization, transfer-log provenance, and precompile overrides.

crates/anvil/src/eth/backend/mem/mod.rs
mattsse: Addressed in e7db2a9f3. Simulation now builds typed receipts from canonical logs and cumulative gas, derives the post-state root from the recursively merged cache, fills execution/fork-dependent header fields, and hashes only the finalized header before assigning transaction metadata and chaining parent/BLOCKHASH values. The HTTP regression independently checks the receipt root and bloom, compares stateRoot with an equivalently mined block, recomputes each returned hash, verifies prevRandao/full-transaction/parent/BLOCKHASH metadata, and checks that blob fields are absent before Cancun.

crates/anvil/src/eth/backend/mem/mod.rs
mattsse: Addressed in e7db2a9f3. Canonical EVM logs now remain in ExecutionResult/receipts, while response-only transfer logs are recorded in a separate frame-aware inspector stream with execution-order indices. No address/topic content filtering remains. The HTTP regression installs code at 0xeeee… that emits the transfer topic and independently verifies that this canonical log remains in logsBloom and the receipt root; the nested-revert regression also verifies rollback and attempted-log ordering.

crates/anvil/src/eth/backend/mem/mod.rs
mattsse: Addressed in e7db2a9f3. Precompile handling now derives from the full state-override address set: move destinations that also have overrides are rejected, every overridden active precompile is removed (including dynamic lookups) so ordinary override code executes, and the same move/remove/warm-address logic is applied on Ethereum, OP, and Tempo EVM paths. HTTP regressions cover ordinary code replacing identity, destination collision rejection, and successful moved identity calls on Tempo and OP.

mablr:
Superseded by #15770, #15784, #15792, #15828, #15841, #15844, #15852, #15869, #15870, #15934, #15938, #15975, #16027.

PR #15720, opened 2026-07-14
feat(bench): add regression thresholds

stevencartavia:
Adds calibrated per-benchmark thresholds to nightly benchmarks and reports meaningful regressions without failing the benchmark run. Coverage remains advisory and uncalibrated benchmarks do not alert.

Closes OSS-479

### DIFF
--- a/.github/scripts/compare-benchmark-thresholds.py
+++ b/.github/scripts/compare-benchmark-thresholds.py
@@ -84,7 +84,12 @@ def compare(base, candidate, rules):
         if not has_base or not has_candidate:
             present = duration(mean_of(candidate[key] if not has_base else base[key], key))
             left, right = ("N/A", present) if not has_base else (present, "N/A")
-            print(f"| `{key}` | {left} | {right} | ⚠️ Inconclusive (missing side) |")
+            if has_base and rules.get(key, {}).get("alert", False):
+                has_regression = True
+                verdict = "❌ Missing nightly result"
+            else:
+                verdict = "⚠️ Inconclusive (missing side)"
+            print(f"| `{key}` | {left} | {right} | {verdict} |")
             continue
 
         baseline = mean_of(base[key], key)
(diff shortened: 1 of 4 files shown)


.github/scripts/compare-benchmark-thresholds.py
```diff
@@ -0,0 +1,136 @@
...
+    print("| Benchmark | Stable | Nightly | Change |")
+    print("|-----------|--------:|---------:|--------|")
+    has_regression = False
+    for key in sorted(base.keys() | candidate.keys()):
+        has_base, has_candidate = key in base, key in candidate
+        if not has_base or not has_candidate:
+            present = duration(mean_of(candidate[key] if not has_base else base[key], key))
+            left, right = ("N/A", present) if not has_base else (present, "N/A")
+            print(f"| `{key}` | {left} | {right} | ⚠️ Inconclusive (missing side) |")
+            continue
```
grandizzy, line 93: A missing nightly result needs to alert. Every benchmark step uses `continue-on-error`, so if a nightly benchmark crashes while stable succeeds, this branch reports it as inconclusive and returns success. Please distinguish the two cases: a missing baseline may remain inconclusive, but a missing nightly result for an `alert: true` benchmark should set `has_regression`.

```suggestion
        if not has_base or not has_candidate:
            present = duration(mean_of(candidate[key] if not has_base else base[key], key))
            left, right = ("N/A", present) if not has_base else (present, "N/A")
            if has_base and rules.get(key, {}).get("alert", False):
                has_regression = True
                verdict = "❌ Missing nightly result"
            else:
                verdict = "⚠️ Inconclusive (missing side)"
            print(f"| `{key}` | {left} | {right} | {verdict} |")
            continue
```

grandizzy (changes requested):
The per-benchmark threshold implementation looks good overall, but missing nightly results can currently be silently accepted. Since benchmark steps use `continue-on-error`, this needs to be fixed before merging.

.github/scripts/compare-benchmark-thresholds.py
stevencartavia: [4a72e24](https://github.com/foundry-rs/foundry/pull/15720/commits/4a72e24c352f208c97ab37475e0d9e85598c9892)

grandizzy (approved):
lgtm

mablr (approved):

mattsse (commented):
The original implementation treated a missing nightly result as non-alerting even though the benchmark steps use `continue-on-error`, which could silently hide a broken nightly benchmark. This was addressed in 4a72e24c by alerting when an alert-enabled benchmark has a stable result but no nightly result.

PR #9190, opened 2024-10-24
fix(forge): fix stack overflow when the lib path is absolute.

l1nxy:
## Motivation
Close #9127
The crash is caused by the recursive call of `get_mapping` in the config crate:
https://github.com/foundry-rs/foundry/blob/c6d59b32fad4b78453354b92acfef5a95013b17f/crates/config/src/providers/remappings.rs#L101-L161
The `get_mapping` function calls the `lib_foundry_toml_remappings` function:
https://github.com/foundry-rs/foundry/blob/c6d59b32fad4b78453354b92acfef5a95013b17f/crates/config/src/providers/remappings.rs#L196-L211
In the `lib_foundry_toml_remappings`, it reads config from a nested lib folder. However, when the lib path in `self.lib_paths` is absolute, it uses `self.root.join(lib_path)`, it will produce wrong path if lib_path is absolute.
The join function is described as follows:
> pub fn [join](https://doc.rust-lang.org/std/path/struct.Path.html#method.join)<P: [AsRef](https://doc.rust-lang.org/std/convert/trait.AsRef.html)<[Path](https://doc.rust-lang.org/std/path/struct.Path.html)>>(&self, path: P) -> [PathBuf](https://doc.rust-lang.org/std/path/struct.PathBuf.html)
> Creates an owned [PathBuf](https://doc.rust-lang.org/std/path/struct.PathBuf.html) with path adjoined to self.
> **If the path is absolute, it replaces the current path.**

Therefore, if the lib paths are absolute, they will only return the setting path and recursively search for nested libs.

## Solution
I propose adding a check to determine whether the `lib` path is absolute or not. If it is absolute, we should return the root folder joined by `lib`. Otherwise, we should return just the new path based on lib's relative path.
By the way, I am not sure if setting an absolute path for the `lib` option in config is allowed. If we need to disable it, we can do so and print a warning message to inform users.

PR #15723, opened 2026-07-14
fix(config): resolve absolute lib remappings

mablr:
## Motivation

Absolute library paths were skipped when discovering remappings from nested foundry.toml files, causing shared or external dependencies to lose their configured remappings. This scans the configured absolute path while preventing inherited FOUNDRY_LIBS and DAPP_LIBS values from recursively re-entering the same library tree.

Addresses the absolute-path issue reported in #15721.

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -808,7 +808,9 @@ impl Config {
     /// that may not define all profiles the main project uses.
     #[track_caller]
     pub fn load_with_root_and_fallback(root: impl AsRef<Path>) -> Result<Self, ExtractConfigError> {
-        let figment = Self::figment_with_root(root.as_ref());
+        // Nested configs must use their own library paths. Inheriting an absolute `FOUNDRY_LIBS`
+        // or `DAPP_LIBS` value would make remapping discovery recurse into the same directory.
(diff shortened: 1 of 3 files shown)


grandizzy:
this needs careful review as there are many variations not covered by unit tests and can break project builds

grandizzy (changes requested):
this reintroduce https://github.com/foundry-rs/foundry/pull/9190

mattsse (changes requested):
[P1] Guard against cycles when scanning absolute libs

Ignoring inherited `FOUNDRY_LIBS`/`DAPP_LIBS` only prevents the direct environment-variable recursion. A nested dependency can still specify the same absolute library root in its own `foundry.toml`, causing `find_nested_foundry_remappings` to reload that dependency indefinitely.

I reproduced this with top-level `FOUNDRY_LIBS=["/absolute/external"]` and `/absolute/external/nested/foundry.toml` containing `libs = ["/absolute/external"]`. Running `forge config` aborts with a stack overflow and exit status 134.

Please add cycle detection (or otherwise prevent recursive re-entry) and cover a self-referential or cyclic absolute library configuration with a regression test.

mablr (commented):
Closing, as the design is a bit flawed.

PR #15717, opened 2026-07-13
fix(ci): make Tempo devnet checks toggleable

decofe:
## Summary
- keep Tempo devnet CI enabled by default
- allow maintainers to disable devnet-dependent checks by setting the `TEMPO_DEVNET_CI` repository variable to `false`
- preserve manual and automatic devnet test coverage whenever the toggle is enabled

## Motivation
The Tempo devnet RPC began returning `-32002: no healthy upstreams available` on July 12 at 19:05 UTC, making unrelated Foundry PRs fail. The previous successful run started at 13:58 UTC. This provides an incident toggle without requiring another code change or permanently disabling coverage.

## Validation
- `git diff --check`
- verified the final PR diff contains only the three workflow condition changes

Prompted by: @DaniPopes

PR #15725, opened 2026-07-14
revert(ci): restore unconditional Tempo devnet checks

decofe:
## Summary

- revert #15717
- restore unconditional Tempo devnet CI checks
- restore direct use of `TEMPO_DEVNET_RPC_URL` in fork schedule tests

## Motivation

The devnet outage was caused by network policy and devnet is back online, so the temporary repository-variable toggle is no longer needed.

## Validation

- `git diff HEAD^ --check`
- verified the diff is the exact inverse of #15717

Prompted by: @grandizzy


mablr (approved):

grandizzy:
not needed, will set `TEMPO_DEVNET_CI = true` var

PR #15726, opened 2026-07-14
fix(forge): default invariant workers to auto

mablr:
Defaults invariant campaign workers to `auto` across the shared configuration so `forge test` uses the active Rayon pool consistently with `forge fuzz run`. Explicit numeric worker settings remain authoritative.

This removes the command-specific fallback and adds regression coverage proving the default behavior is isolated from environment overrides.

I don't think this is a breaking change but we need to be careful.

### DIFF
--- a/crates/forge/tests/cli/test_cmd/invariant/mod.rs
+++ b/crates/forge/tests/cli/test_cmd/invariant/mod.rs
@@ -1436,6 +1436,46 @@ contract JsonInvariantReportTest is Test {
     assert!(safe.get("reason").is_none());
 });
 
+forgetest_init!(forge_test_defaults_invariant_workers_to_auto, |prj, cmd| {
+    prj.add_test(
+        "AutoInvariantWorkers.t.sol",
+        r#"
+import {Test} from "forge-std/Test.sol";
+
+contract AutoInvariantWorkersHandler {
+    uint256 public counter;
+
+    function inc() external {
+        counter++;
+    }
+}
+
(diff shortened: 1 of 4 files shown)


crates/forge/tests/cli/test_cmd/invariant/mod.rs
```diff
@@ -1436,6 +1436,46 @@ contract JsonInvariantReportTest is Test {
...
+    }
+
+    function invariant_break() public view {
+        require(handler.counter() == 0, "broken");
+    }
+}
+   "#,
+    );
+
+    cmd.unset_env("FOUNDRY_INVARIANT_WORKERS");
```
mattsse, line 1471: This only clears the environment override. A developer’s cached `foundry-forge-test-template` may have been generated before this change and still contain `workers = 1`; that explicit value remains authoritative, so this test fails locally with `"workers": 1` even though it passes on fresh CI. Reset the project value to the code default before invoking Forge (the unit test above still verifies that the default itself is `Auto`).

```suggestion
    prj.update_config(|config| {
        config.invariant.workers = foundry_config::InvariantWorkers::default();
    });
    cmd.unset_env("FOUNDRY_INVARIANT_WORKERS");
```

mattsse (commented):
The implementation looks sound, but the new regression test needs to be isolated from cached test-template configuration.

mattsse (approved):
LGTM

figtracer (approved):

PR #15727, opened 2026-07-14
feat(debugger): select tests interactively

figtracer:
## Motivation

`forge test --debug` currently requires exactly one matching test. The candidate list added in #15407 helps identify matches, but interactive users still have to quit and rerun Forge with a narrower filter.

This progresses the debugger quality-of-life tracking in #927.

## Changes

- prompt interactive users to select a test when `--debug` matches multiple tests
- preserve exact test signatures so overloaded tests remain distinct
- reuse the existing `RerunFailure` filter path to run only the selected test
- keep the existing bounded candidate list and error for CI and other non-interactive environments
- treat Escape or cancellation as an explicit cancellation instead of selecting implicitly


mablr (dismissed):

crates/forge/src/cmd/test/mod.rs
```diff
@@ -2604,6 +2606,38 @@ impl TestArgs {
...
+            };
+
+            let mut selected_filter = filter.clone();
+            selected_filter.set_rerun_failures(vec![matching_debug_tests.swap_remove(selected)]);
+            num_filtered = 1;
+            Some(selected_filter)
+        } else {
+            None
+        };
+        let filter = selected_filter.as_ref().unwrap_or(filter);
```
grandizzy: This selection is pass-local, but `run_tests_inner` is invoked once for the default network and once per inline network override. Candidate discovery is global while network routing happens later, so users receive the same prompt repeatedly and can select different tests across passes. The merged outcome can then contain multiple tests, while the debugger silently takes the first.

One minimal fix is to clone the filter once in `run_tests`, pass it mutably through `dispatch_network` → `build_and_run_tests` → `run_tests_inner`, and store the selected exact test directly on that shared filter:

```diff
 pub(crate) async fn run_tests(
     &mut self,
     ...
     filter: &ProjectPathsAwareFilter,
     ...
 ) -> Result<TestOutcome> {
+    let mut filter = filter.clone();
     ...
-    self.dispatch_network(..., filter, ...).await?
+    self.dispatch_network(..., &mut filter, ...).await?
 }

 async fn dispatch_network(
     &self,
     ...
-    filter: &ProjectPathsAwareFilter,
+    filter: &mut ProjectPathsAwareFilter,
     ...
 )

 async fn build_and_run_tests<FEN: FoundryEvmNetwork>(
     &self,
     ...
-    filter: &ProjectPathsAwareFilter,
+    filter: &mut ProjectPathsAwareFilter,
     ...
 )

 async fn run_tests_inner<FEN: FoundryEvmNetwork>(
     &self,
     ...
-    filter: &ProjectPathsAwareFilter,
+    filter: &mut ProjectPathsAwareFilter,
     ...
 ) -> eyre::Result<TestOutcome> {
     ...
-    let selected_filter = if interactive_debug_selection {
+    if interactive_debug_selection {
         let Some(selected) = Select::new()
             ...
         else {
             bail!("Debugger test selection cancelled");
         };

-        let mut selected_filter = filter.clone();
-        selected_filter.set_rerun_failures(vec![
+        filter.set_rerun_failures(vec![
             matching_debug_tests.swap_remove(selected),
-        ]);
+        ]);
         num_filtered = 1;
-        Some(selected_filter)
-    } else {
-        None
-    };
-    let filter = selected_filter.as_ref().unwrap_or(filter);
+    }
```

The same mutable filter must be reused for every default/override dispatch. That makes the prompt run once, while only the network pass owning the selected test executes it.

grandizzy (changes requested):
The exact test selection works for single-network runs, but multi-network projects currently prompt once per pass and can execute multiple selected tests. Selection should happen once for the entire run.

crates/forge/src/cmd/test/mod.rs
```diff
@@ -2604,6 +2606,38 @@ impl TestArgs {
...
+        let selected_filter = if interactive_debug_selection {
+            let Some(selected) = Select::new()
+                .with_prompt("Select a test to debug")
+                .items(
+                    matching_debug_tests
+                        .iter()
+                        .map(|test| format!("{}.{}", test.contract, test.test)),
+                )
+                .max_length(DEBUGGER_MATCHING_TESTS_DISPLAY_LIMIT)
+                .interact_on_opt(&Term::stdout())?
```
mattsse: The selection prompt is diagnostic UI, so it should follow the CLI channel contract and render on stderr.

```suggestion
                .interact_on_opt(&Term::stderr())?
```

mattsse (changes requested):
Selection currently happens inside `run_tests_inner`, which executes once per multi-network pass. This can prompt repeatedly and allow different tests to be selected in one invocation. Please select once for the whole run and reuse the exact selected filter across all dispatches. A PTY integration test covering multi-network selection would prevent this regression.

grandizzy (dismissed):
lgtm

mablr (dismissed):

crates/forge/src/cmd/test/mod.rs
```diff
@@ -2607,6 +2611,33 @@ impl TestArgs {
...
+        if interactive_debug_selection {
+            let Some(selected) = Select::new()
+                .with_prompt("Select a test to debug")
+                .items(
+                    matching_debug_tests
+                        .iter()
+                        .map(|test| format!("{}.{}", test.contract, test.test)),
+                )
+                .max_length(DEBUGGER_MATCHING_TESTS_DISPLAY_LIMIT)
+                .interact_on_opt(&Term::stderr())?
```
stevencartavia: This picker uses `dialoguer::Select`, which hides the cursor while waiting for input, but pressing Ctrl-C terminates Forge without restoring it. I reproduced this with the shell prompt returning while the terminal cursor remained hidden. Could we restore cursor visibility on interruption while preserving the expected SIGINT exit behavior?

crates/forge/src/cmd/test/mod.rs
```diff
@@ -2607,6 +2611,33 @@ impl TestArgs {
             return Ok(TestOutcome::empty(Some(runner.known_contracts.clone()), false));
         }
 
+        let interactive_debug_selection =
+            self.debug && num_filtered != 1 && tui_mode().is_interactive();
```
stevencartavia: `tui_mode()` checks CI/stdin/stdout, but this picker uses `Term::stderr()`. With stdin/stdout attached to a terminal and only stderr redirected, Forge exits with `Error: IO error: not a terminal` instead of showing the existing bounded matching-tests fallback. Could we also require stderr to be a terminal before opening the picker and otherwise use the non-interactive fallback?

mattsse (approved):
lgtm

PR #15728, opened 2026-07-14
fix(anvil): isolate fork cache on RPC reset

mablr:
## Motivation

https://github.com/foundry-rs/foundry/pull/15713#discussion_r3578296667 follow-up

Clear persisted remote state only after a replacement RPC endpoint has been validated. This prevents resets between origins with colliding block environments from reusing the previous origin's accounts while preserving same-origin cache reuse.

### DIFF
--- a/crates/anvil/src/eth/backend/fork.rs
+++ b/crates/anvil/src/eth/backend/fork.rs
@@ -59,12 +59,20 @@ pub struct ClientFork<N: Network = AnyNetwork> {
     pub config: Arc<RwLock<ClientForkConfig<N>>>,
     /// This also holds a handle to the underlying database
     pub database: Arc<AsyncRwLock<Box<dyn Db>>>,
+    /// The RPC URL associated with the state in the underlying database.
+    database_rpc_url: Arc<RwLock<Option<String>>>,
 }
 
 impl<N: Network> ClientFork<N> {
     /// Creates a new instance of the fork
     pub fn new(config: ClientForkConfig<N>, database: Arc<AsyncRwLock<Box<dyn Db>>>) -> Self {
-        Self { storage: Default::default(), config: Arc::new(RwLock::new(config)), database }
+        let database_rpc_url = config.eth_rpc_url().map(ToOwned::to_owned);
+        Self {
+            storage: Default::default(),
+            config: Arc::new(RwLock::new(config)),
+            database,
+            database_rpc_url: Arc::new(RwLock::new(database_rpc_url)),
+        }
     }
 
     /// Removes all data cached from previous responses
@@ -109,6 +117,14 @@ impl<N: Network> ClientFork<N> {
 
     pub fn eth_rpc_url(&self) -> Option<String> {
         self.config.read().eth_rpc_url().map(|s| s.to_string())
+    }
+
+    pub(crate) fn database_rpc_url(&self) -> Option<String> {
+        self.database_rpc_url.read().clone()
+    }
+
+    pub(crate) fn set_database_rpc_url(&self, url: Option<String>) {
+        *self.database_rpc_url.write() = url;
     }
 
     pub fn chain_id(&self) -> u64 {
@@ -421,6 +437,7 @@ impl<N: Network> ClientFork<N> {
         let number = self.block_number();
         let block_hash = self.block_hash();
         self.database.write().await.insert_block_hash(U256::from(number), block_hash);
+        self.set_database_rpc_url(self.eth_rpc_url());
 
         Ok(())
     }
--- a/crates/anvil/src/eth/backend/fork.rs
+++ b/crates/anvil/src/eth/backend/fork.rs
@@ -437,7 +437,6 @@ impl<N: Network> ClientFork<N> {
         let number = self.block_number();
         let block_hash = self.block_hash();
         self.database.write().await.insert_block_hash(U256::from(number), block_hash);
-        self.set_database_rpc_url(self.eth_rpc_url());
 
         Ok(())
     }
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -2486,14 +2486,12 @@ impl<N: Network> Backend<N> {
             // reset the fork entirely and reapply the genesis config
             let reset_urls =
                 forking.json_rpc_url.as_ref().map(|url| vec![url.clone()]).unwrap_or_default();
-            let rpc_url_changed = forking
-                .json_rpc_url
-                .as_ref()
-                .is_some_and(|url| fork.eth_rpc_url().as_ref() != Some(url));
+            let target_rpc_url = forking.json_rpc_url.clone().or_else(|| fork.eth_rpc_url());
+            let rpc_url_changed = target_rpc_url != fork.database_rpc_url();
             fork.prepare_reset(reset_urls, block_number.into()).await?;
             if rpc_url_changed {
                 // Clear state fetched from the previous RPC URL before persisting the cache.
-                fork.database.write().await.clear();
+                fork.database.write().await.clear_into_state_snapshot();
             }
             // Persist fetched remote state before rebuilding the fork database so the new
             // block-specific database can load it from disk.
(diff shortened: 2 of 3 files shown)


figtracer (dismissed):

mattsse (dismissed):
lgtm

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -2486,10 +2486,18 @@ impl<N: Network> Backend<N> {
             // reset the fork entirely and reapply the genesis config
             let reset_urls =
                 forking.json_rpc_url.as_ref().map(|url| vec![url.clone()]).unwrap_or_default();
+            let rpc_url_changed = forking
+                .json_rpc_url
+                .as_ref()
+                .is_some_and(|url| fork.eth_rpc_url().as_ref() != Some(url));
```
grandizzy: `fork.eth_rpc_url()` reflects mutable `ClientForkConfig`. After `anvil_setRpcUrl(B)`, it already returns B even though the active database still contains state fetched from A. Consequently, `anvil_reset(None)` treats the origin as unchanged and can persist/reload A’s state for B.

Please track the URL associated with the active database separately, then compare the effective reset URL against it:

```suggestion
            let target_rpc_url =
                forking.json_rpc_url.clone().or_else(|| fork.eth_rpc_url());
            let rpc_url_changed = target_rpc_url != fork.database_rpc_url();
```

For example, `ClientFork` could capture that immutable value when its database is constructed:

```diff
 pub struct ClientFork<N: Network = AnyNetwork> {
     pub storage: Arc<RwLock<ForkedStorage<N>>>,
     pub config: Arc<RwLock<ClientForkConfig<N>>>,
     pub database: Arc<AsyncRwLock<Box<dyn Db>>>,
+    database_rpc_url: Option<String>,
 }

 impl<N: Network> ClientFork<N> {
     pub fn new(config: ClientForkConfig<N>, database: Arc<AsyncRwLock<Box<dyn Db>>>) -> Self {
-        Self { storage: Default::default(), config: Arc::new(RwLock::new(config)), database }
+        let database_rpc_url = config.eth_rpc_url().map(ToOwned::to_owned);
+        Self {
+            storage: Default::default(),
+            config: Arc::new(RwLock::new(config)),
+            database,
+            database_rpc_url,
+        }
     }
+
+    pub fn database_rpc_url(&self) -> Option<String> {
+        self.database_rpc_url.clone()
+    }
 }
```

Please also add a regression case using `A → anvil_setRpcUrl(B) → anvil_reset(None)`; the current test only covers passing B explicitly to `anvil_reset`.

grandizzy (changes requested):
The explicit `anvil_reset` URL-change path is improved, but changes made through `anvil_setRpcUrl` can still bypass cache invalidation because the comparison uses mutable fork configuration rather than the RPC origin backing the active database. Please track the database’s origin separately and cover `A → anvil_setRpcUrl(B) → anvil_reset(None)`.

figtracer (dismissed):

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -2486,10 +2486,16 @@ impl<N: Network> Backend<N> {
             // reset the fork entirely and reapply the genesis config
             let reset_urls =
                 forking.json_rpc_url.as_ref().map(|url| vec![url.clone()]).unwrap_or_default();
+            let target_rpc_url = forking.json_rpc_url.clone().or_else(|| fork.eth_rpc_url());
+            let rpc_url_changed = target_rpc_url != fork.database_rpc_url();
+            fork.prepare_reset(reset_urls, block_number.into()).await?;
+            if rpc_url_changed {
+                // Clear state fetched from the previous RPC URL before persisting the cache.
+                fork.database.write().await.clear_into_state_snapshot();
```
stevencartavia, line 2494: This clears the current database’s remote state, but resetting to an earlier block can still load that block’s persisted cache from the previous RPC endpoint. I reproduced this by forking the first endpoint at block 1, resetting to block 2, switching to a second endpoint, and resetting back to block 1. The result used the first endpoint’s cached balance (`1337`) instead of the second endpoint’s balance (`42`). Deleting only the persisted block-1 cache made the same test pass. Could we prevent caches created for one RPC endpoint from being loaded after switching endpoints and add this sequence as a regression test?

crates/anvil/src/eth/backend/fork.rs
```diff
@@ -421,6 +437,7 @@ impl<N: Network> ClientFork<N> {
         let number = self.block_number();
         let block_hash = self.block_hash();
         self.database.write().await.insert_block_hash(U256::from(number), block_hash);
+        self.set_database_rpc_url(self.eth_rpc_url());
```
stevencartavia: `ForkedDatabase::reset` ignores the new URL and continues using its existing provider, but this assignment records the new URL as the database’s origin. I reproduced this by forking the first endpoint, calling `ClientFork::reset` with the second endpoint, and then performing a normal reset. The result still used the first endpoint’s balance (`1337`) instead of the second endpoint’s (`42`). Removing only this assignment made the test pass. Could we leave the tracked database URL unchanged unless the database provider is actually replaced?

mattsse (dismissed):
Two cache-origin mismatches remain before this is safe to merge.

First, when the RPC origin changes, invalidate the persisted cache for the **target** block before `setup_fork_db_config` can load it. Clearing and flushing only the currently active database handles same-block resets, but not `A@1 → A@2 → B@1`; that sequence still loads A's block-1 cache and returns `1337` instead of B's `42`. A contained fix would remove or bypass the target block's cache after the replacement endpoint has been validated and before rebuilding the database. A more structural fix would include a stable chain/block identity such as the fork block hash in cache validation, so equivalent providers for the same chain can still share valid entries. Please add the two-block sequence as a regression test.

Second, treat `database_rpc_url` as the origin of the provider that actually backs the database. `ForkedDatabase::reset` currently ignores the supplied URLs and retains its existing `SharedBackend`, so `ClientFork::reset` must not record the new URL unless it also replaces that provider. Either rebuild the fork database/backend with the new provider inside that path, or leave `database_rpc_url` unchanged so the next backend reset detects the mismatch and performs the rebuild. Please cover `ClientFork::reset(A → B)` followed by a normal backend reset.

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -2486,11 +2486,29 @@ impl<N: Network> Backend<N> {
...
+                let cache_path = self.node_config.read().await.block_cache_path(fork_block_number);
+                if let Some(cache_path) = cache_path
+                    && let Err(err) = std::fs::remove_file(&cache_path)
+                    && err.kind() != std::io::ErrorKind::NotFound
+                {
+                    return Err(BlockchainError::Internal(format!(
+                        "failed to invalidate fork cache at {}: {err}",
+                        cache_path.display()
+                    )));
+                }
```
grandizzy: Persisted fork caches are keyed by chain ID and block, not RPC origin. Removing only the immediate target block leaves other blocks from the previous origin behind. For example, `A@1 → A@2 → B@2 → B@1` reloads A's block-1 cache during the final same-origin reset. Please invalidate the complete chain cache namespace when the origin changes, or include the origin/canonical block identity in cache validation, and extend the regression test with the final reset from B@2 back to B@1. The smallest contained fix is:

```suggestion
                let cache_dir = {
                    let config = self.node_config.read().await;
                    if config.no_storage_caching || config.fork_urls.is_empty() {
                        None
                    } else {
                        foundry_config::Config::foundry_chain_cache_dir(config.get_chain_id())
                    }
                };
                if let Some(cache_dir) = cache_dir
                    && let Err(err) = std::fs::remove_dir_all(&cache_dir)
                    && err.kind() != std::io::ErrorKind::NotFound
                {
                    return Err(BlockchainError::Internal(format!(
                        "failed to invalidate fork cache at {}: {err}",
                        cache_dir.display()
                    )));
                }
```

grandizzy (changes requested):
The immediate URL-switch paths are fixed, but cache invalidation remains incomplete across block heights. A later same-origin reset can still reload state persisted by the previous origin.

grandizzy (approved):
lgtm

figtracer (approved):

PR #15730, opened 2026-07-14
fix(anvil): skip history tracking when state history is pruned

mattsse:
Follow-up to #15715. `HistoricalStateCache` is only ever drained by `current_state()`, which `mine_block` calls behind the `is_state_history_supported()` gate. With `--prune-history` that call never happens, so the dirty sets recorded on every account/storage access grow with the touched key-space and are never released — with EIP-2935 the history contract alone adds one slot per block toward the 8191 ring size. That is pure bookkeeping overhead in exactly the memory-conscious mode.

History recording is now disabled at construction when state history is unsupported: `NodeConfig::setup` builds the db via `StateRootDb::new(self.prune_history.is_state_history_supported())`, and a disabled cache turns all `record_*` calls into no-ops. If `current_state()` is ever called anyway, it still returns a correct snapshot built fresh from the live db, without caching it, so a stale incremental state can never be served. The incremental state-root cache is unaffected since it is drained on every mined block.

Claude was used to implement and test this change; AI-generated responses may also be used during review.

### DIFF
--- a/crates/anvil/src/config.rs
+++ b/crates/anvil/src/config.rs
@@ -1217,7 +1217,8 @@ impl NodeConfig {
             if let Some(eth_rpc_url) = self.fork_urls.first().cloned() {
                 self.setup_fork_db(eth_rpc_url, &mut evm_env, &fees).await?
             } else {
-                (Arc::new(TokioRwLock::new(Box::<StateRootDb>::default())), None)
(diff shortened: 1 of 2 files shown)


mablr (approved):
LGTM

stevencartavia (approved):

Issue #10863, opened 2025-06-27
Cast run does not work sometimes with weird errors (lack of funds, nonce too low, etc.)


0xDEnYO:
### Component

Forge, Cast

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

forge Version: 1.2.3-stable

### What version of Foundryup are you on?

foundryup: 0.3.3

### What command(s) is the bug in?

Cast run

### Operating System

macOS (Apple Silicon)

### Describe the bug

Sometimes when I run cast run I do not get any stack trace. 
Here is one example: 

Case 1:
`cast run 0xe5dff9c3f5ec5b1af5f641f1c1f1829b38713641b4401b2affeaf794d7b48e0b --rpc-url https://zksync2-mainnet.zksync.io -vvv`

Explorer URL 1: 
https://era.zksync.network/tx/0xe5dff9c3f5ec5b1af5f641f1c1f1829b38713641b4401b2affeaf794d7b48e0b
Explorer URL 2: 
https://explorer.zksync.io/tx/0xe5dff9c3f5ec5b1af5f641f1c1f1829b38713641b4401b2affeaf794d7b48e0b
(in this explorer we get some data back that helps to extract the revert reason)
Also Phalcon is able to produce a full stack trace: 
https://app.blocksec.com/explorer/tx/zksync-era/0xe5dff9c3f5ec5b1af5f641f1c1f1829b38713641b4401b2affeaf794d7b48e0b

This transaction uses a precompile which in Telegram was pointed out could be an issue. Is there a way this could be improved so that also those stack traces are shown?

![Image](https://github.com/user-attachments/assets/75636c25-307b-4681-9e65-2f3c55954968)

---------
Case 2:
Here is a second transaction that does not work: 

`cast run 0x715e633fc1ae98bfda07a52b11758acb9bd5bf73cd4c386708e694ec42da5924 --rpc-url https://rpc.hyperliquid.xyz/evm -vvv`

Explorer URL:
https://www.hyperscan.com/tx/0x715e633fc1ae98bfda07a52b11758acb9bd5bf73cd4c386708e694ec42da5924

This one is more difficult to debug since HyperEVM is not supported by Tenderly or Phalcon which is why I would love to see a full stack trace for it. 
But using cast run we get a nonce too low error (that i've seen a few times before):

[...]

PR #15731, opened 2026-07-14
fix(evm): enable BSC P256 at Haber

mattsse:
BSC activated the RIP-7212 secp256r1 precompile at Haber, before the corresponding Ethereum hardfork. Foundry selected precompiles only from the Ethereum spec ID, so historical BSC execution treated address `0x100` as an empty account and diverged from on-chain behavior.

This models the BSC P256 schedule on mainnet and testnet: `0x100` is absent before Haber, uses the 3,450-gas variant from Haber, and switches to the 6,900-gas variant at BSC Osaka. The overlay is selected from the chain ID and timestamp whenever the Forge or Anvil EVM is constructed, including active-fork transaction-prefix replay. Anvil applies it to calls, mining, and `eth_config`; local genesis setup also keeps the initial EVM timestamp aligned with block 0.

Precompile selection intentionally follows the existing EVM hardfork lifecycle. `vm.warp` changes the block timestamp but does not reselect the spec or precompile schedule. The pre-existing non-active cross-chain `rollFork` replay-environment issue is outside this focused change.

Related to #10863.

This PR was written with Codex, including investigation, code generation, and test generation.


crates/evm/networks/src/lib.rs
```diff
@@ -328,6 +337,28 @@ impl NetworkConfigs {
...
+        chain_id: ChainId,
+        timestamp: u64,
+    ) {
+        let is_haber = match chain_id {
+            BSC_MAINNET_CHAIN_ID => timestamp >= BSC_MAINNET_HABER_TIMESTAMP,
+            BSC_TESTNET_CHAIN_ID => timestamp >= BSC_TESTNET_HABER_TIMESTAMP,
+            _ => false,
+        };
+        if is_haber {
+            let p256verify = P256VERIFY;
```
figtracer: this needs all three bsc ranges. before haber, an osaka map already has `0x100` and this leaves it active; after bsc osaka, this replaces the 6,900-gas variant with the 3,450-gas one. please remove it pre-haber and select the osaka variant from `1777343400` on mainnet / `1774319400` on testnet, with boundary tests.

crates/evm/core/src/evm/eth.rs
```diff
@@ -45,9 +45,13 @@ impl FoundryEvmFactory for EthEvmFactory {
...
         inspector: I,
     ) -> Self::FoundryEvm<'db, I> {
+        let chain_id = evm_env.cfg_env.chain_id;
+        let timestamp = evm_env.block_env.timestamp.saturating_to();
         let mut eth_evm = Self::default().create_evm_with_inspector(db, evm_env, inspector);
         eth_evm.cfg.tx_chain_id_check = true;
-        eth_evm.inspector().get_networks().inject_precompiles(eth_evm.precompiles_mut());
+        let networks = eth_evm.inspector().get_networks();
+        networks.inject_precompiles(eth_evm.precompiles_mut());
+        networks.inject_chain_precompiles(eth_evm.precompiles_mut(), chain_id, timestamp);
```
figtracer, line 54: anvil bypasses this method: its call and mining paths construct the eth evm directly and only run the generic injection. an exact-haber bsc node with `--hardfork cancun` still returns empty from `0x100` on this head. please apply the chain/timestamp injection in those paths too.

figtracer (changes requested):

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -1224,12 +1224,18 @@ impl<N: Network> Backend<N> {
...
     /// 1. Network-specific precompiles (e.g. Tempo, OP)
-    /// 2. User-provided precompiles via [`PrecompileFactory`]
-    /// 3. Cheatcode ecrecover overrides (if active)
-    fn inject_precompiles(&self, precompiles: &mut PrecompilesMap) {
+    /// 2. Chain- and timestamp-specific precompiles
+    /// 3. User-provided precompiles via [`PrecompileFactory`]
+    /// 4. Cheatcode ecrecover overrides (if active)
+    fn inject_precompiles(&self, precompiles: &mut PrecompilesMap, evm_env: &EvmEnv) {
         self.networks.inject_precompiles(precompiles);
+        self.networks.inject_chain_precompiles(
```
mablr, line 1241: Could we apply this same chain/timestamp overlay when building `Backend::precompiles()`? `eth_config` uses that method to report the currently available precompiles, but it still derives them only from the Ethereum spec and configured network. As a result, Cancun-at/after-Haber executes P256 while omitting it from `eth_config`, and Osaka-before-Haber reports P256 even though execution removes it. The overlay should run before custom factory entries so reporting keeps the same precedence as execution.

crates/evm/networks/src/lib.rs
```diff
@@ -328,6 +360,23 @@ impl NetworkConfigs {
...
 
+    /// Injects chain-specific precompiles active at the given timestamp.
+    pub fn inject_chain_precompiles(
+        self,
+        precompiles: &mut PrecompilesMap,
+        chain_id: ChainId,
+        timestamp: u64,
+    ) {
+        let Some(p256verify) = bsc_p256_precompile(chain_id, timestamp) else { return };
+        precompiles.apply_precompile(&BSC_P256_ADDRESS, move |_| {
```
stevencartavia, line 371: Could fork selection and rolling rebuild both this map and its warm-address set from the target fork? Selecting pre-Haber BSC block 37,000,000 left P256 active, and rolling to 37,000,001 kept `0x100` warm, undercharging its first access by 2,500 gas.

crates/anvil/tests/it/anvil.rs
```diff
@@ -97,6 +109,42 @@ async fn test_can_set_genesis_timestamp() {
...
+        NodeConfig::test()
+            .with_chain_id(Some(BSC_MAINNET_CHAIN_ID))
+            .with_hardfork(Some(EthereumHardfork::Cancun.into()))
+            .with_genesis_timestamp(Some(BSC_MAINNET_HABER_TIMESTAMP)),
+    )
+    .await;
+    let provider = handle.http_provider();
+
+    let tx = TransactionRequest::default().with_to(P256_VERIFY).with_input(P256_INPUT);
+    let output = provider.call(tx.clone().into()).await.unwrap();
```
stevencartavia: This call defaults to `pending`, which hides a genesis `latest` issue. With BSC configured at the Haber timestamp, explicit `latest` and block `0x0` returned empty while `pending` returned `1`; `latest` worked after mining. Could the current EVM environment also be initialized with the configured genesis timestamp?

crates/evm/core/src/evm/eth.rs
stevencartavia, line 54: `replay_until` uses a bare EVM, so transaction-target forks bypass this overlay. On BSC block 83,275,039 this replayed the preceding P256 transaction with the 6,900-gas variant instead of 3,450, leaving the sender balance 22,477,903,500 wei too low. Could we apply the same chain/timestamp selection to the replay EVM?

crates/anvil/src/eth/backend/mem/mod.rs
mattsse: Addressed in e09b20ba0. Backend::precompiles() now builds the effective spec map, applies the active BSC chain/timestamp overlay, then adds configured-network and custom-factory entries so factory precedence is preserved. Exact-Haber and pre-Haber eth_config regressions cover both reporting directions.

crates/evm/networks/src/lib.rs
mattsse: Addressed in e09b20ba0. Foundry EVMs now use a chain/timestamp-backed dynamic P256 selector, and Backend/CowBackend refresh it together with the journal warm-precompile set on fork selection, rolling, and transaction-target rolling. Boundary tests cover the pre-Haber/Haber map transition and warm-address removal/addition.

crates/anvil/tests/it/anvil.rs
mattsse: Addressed in e09b20ba0. Non-fork startup now copies both the configured genesis number and timestamp into the active EVM block environment. The exact-Haber integration test calls P256 at explicit latest, block 0, and pending and expects the same successful result.

crates/evm/core/src/evm/eth.rs
mattsse: Addressed in e09b20ba0. replay_until now overlays the bare replay EVM with the target block chain ID and timestamp before executing preceding transactions, and transaction-target rolling refreshes the active selector and warm set again after replay.

crates/evm/networks/src/lib.rs
```diff
@@ -328,6 +411,88 @@ impl NetworkConfigs {
...
+        if install_static {
+            let precompile = dynamic_p256_precompile(has_fallback, Some(state));
+            precompiles.apply_precompile(&BSC_P256_ADDRESS, |_| Some(precompile));
+        } else {
+            precompiles.set_precompile_lookup(move |address: &Address| {
+                if address != &BSC_P256_ADDRESS {
+                    return None;
+                }
+
+                let (chain_id, timestamp) = state.get();
```
stevencartavia: This lookup uses `ChainPrecompileState`, but `vm.warp` changes the current EVM timestamp without updating that state. With BSC/Cancun initialized at `1718863499`, warping to `1718863500` leaves a valid call to `0x100` returning empty, while direct initialization at `1718863500` returns `1`. Could we refresh the chain-precompile state and address classification after `vm.warp` and add a regression test?

crates/evm/networks/src/lib.rs
```diff
@@ -328,6 +411,88 @@ impl NetworkConfigs {
...
+    ) {
+        let (chain_id, timestamp) = state.get();
+        let mut has_fallback = false;
+        precompiles.apply_precompile(&BSC_P256_ADDRESS, |existing| {
+            has_fallback = existing.is_some();
+            None
+        });
+
+        let p256verify = bsc_p256_precompile(chain_id, timestamp);
+        let install_static = match &p256verify {
```
stevencartavia: `install_static` is fixed when the EVM is created, so a Haber-to-pre-Haber warp cannot fully deactivate `0x100`; the journal’s precompile-address set is also left unchanged. With BSC/Cancun initialized at `1718863500`, warping to `1718863499` makes `vm.etch(0x100, ...)` fail with `cannot use precompile`, while direct initialization at `1718863499` allows the etch and the installed code returns `42`. Could we refresh both precompile-map membership and the journal’s precompile-address classification on this transition and add a regression test?

crates/evm/core/src/backend/mod.rs
```diff
@@ -968,8 +1005,15 @@ impl<FEN: FoundryEvmNetwork> Backend<FEN> {
 
             // Clone the fork's CacheDB once. The underlying SharedBackend is Arc-backed,
             // so only the local cache layer is actually duplicated.
+            let chain_id = evm_env.cfg_env.chain_id;
```
stevencartavia, line 972: When `id` targets a non-active fork, `roll_fork` leaves `evm_env.cfg_env` on the active fork, so `replay_until` replays the target fork’s preceding transactions with the active chain ID. On a cross-chain roll, prefix transactions fail with `transaction validation error: invalid chain ID`. Could we build the replay environment from the target fork and leave the active environment and journal unchanged unless the target fork is active?

mattsse (commented):
lgtm

crates/evm/core/src/evm/eth.rs
mablr, line 54: This selects the BSC P256 variant only when the outer test EVM is created. `vm.createSelectFork`/`selectFork`/`rollFork` later replace the live chain ID and timestamp without refreshing this map, so a common `vm.createSelectFork(BSC_RPC, postHaberBlock)` call still has no P256 and rolls across Haber/Osaka retain the old variant.

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -615,11 +615,18 @@ impl<N: Network> Backend<N> {
...
     pub fn precompiles(&self) -> BTreeMap<String, Address> {
         let spec_id = self.spec_id();
-        let precompiles = Precompiles::new(PrecompileSpecId::from_spec_id(spec_id));
+        let mut precompiles =
+            PrecompilesMap::from_static(Precompiles::new(PrecompileSpecId::from_spec_id(spec_id)));
+        let (chain_id, timestamp) = {
+            let evm_env = self.evm_env.read();
+            (evm_env.cfg_env.chain_id, evm_env.block_env.timestamp.saturating_to())
+        };
+        self.networks.inject_chain_precompiles(&mut precompiles, chain_id, timestamp);
```
mablr, line 624: `eth_config` enumerates this map before applying `PrecompileFactory`, while execution applies the factory afterward and lets it replace the chain precompile. A factory override at `0x100` therefore reports both P256 and the custom implementation even though only the latter executes. Please apply the factory to this map before enumerating it.

mattsse: I think we can ignore this for bsc

crates/evm/core/src/evm/eth.rs
mattsse: I think we can ignore this for bsc

mablr (approved):
sgtm

figtracer (approved):

PR #15733, opened 2026-07-14
fix(forge): fix JSON compiler errors

DaniPopes:
`forge lint --json` configured its initial Solar session with a human stderr emitter, so parse and compiler failures produced prose instead of machine-readable diagnostics. Route both Solar phases through the rustc-like JSON emitter on stdout and exit unsuccessfully after emitted compiler errors without invoking Forge's top-level prose error printer.

`forge build --json` now likewise exits unsuccessfully after emitting structured compiler errors, giving both commands a clean JSON stdout contract and a meaningful failure status. Human output and build-time lint diagnostic routing remain unchanged.

This PR was written primarily by OpenAI Codex, including code generation and tests.

### DIFF
--- a/crates/forge/src/cmd/build.rs
+++ b/crates/forge/src/cmd/build.rs
@@ -139,6 +139,9 @@ impl BuildArgs {
         if format_json && !self.names && !self.sizes {
             sh_println!("{}", serde_json::to_string_pretty(&output.output())?)?;
         }
+        if format_json && output.has_compiler_errors() {
+            std::process::exit(1);
+        }
 
         // Only run the `SolidityLinter` if lint on build and no compilation errors.
         if !self.no_lint
--- a/crates/forge/src/cmd/build.rs
+++ b/crates/forge/src/cmd/build.rs
@@ -136,7 +136,7 @@ impl BuildArgs {
         // Cache project selectors.
         cache_local_signatures(&output)?;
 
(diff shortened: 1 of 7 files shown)


crates/forge/src/cmd/build.rs
```diff
@@ -139,6 +139,9 @@ impl BuildArgs {
         if format_json && !self.names && !self.sizes {
             sh_println!("{}", serde_json::to_string_pretty(&output.output())?)?;
         }
+        if format_json && output.has_compiler_errors() {
+            std::process::exit(1);
```
mablr, line 143: This seems to exit without emitting the compiler diagnostics when `--names` or `--sizes` is also set. `ProjectCompiler::handle_output` has already printed the names/size JSON, so a broken `forge build --json --sizes` currently returns status 1 with `{}` on stdout and no diagnostic.

DaniPopes: Fixed in bec76a9f6. `ProjectCompiler` now suppresses auxiliary names/size JSON when compilation fails, and `BuildArgs` emits the full compiler diagnostic output regardless of those flags before exiting 1. I added coverage for `forge build --json --sizes` with a compiler error. This response was generated with OpenAI Codex.

mablr (approved):

stevencartavia (approved):

PR #15734, opened 2026-07-14
fix(debugger): guard previous call boundary

figtracer:
## Motivation

The `c` shortcut moves to the previous call, but at the root call it used `saturating_sub(1)` and then unconditionally moved to the final opcode of the unchanged call. As a result, pressing `c` with no previous call unexpectedly jumped from the user's current opcode to the end of the root call.

This is a small debugger navigation fix tracked under #927.

## Changes

- make previous-call navigation a no-op at the root boundary, symmetric with the existing next-call boundary guard
- retain the existing behavior of landing on the previous call's final opcode when a previous call exists
- add focused regression coverage for both paths

## Validation

- live TUI test: opened a single-call Solidity test at root PC `0x0`, pressed `c`, and confirmed the debugger remained at PC `0x0`

### DIFF
--- a/crates/debugger/src/tui/context.rs
+++ b/crates/debugger/src/tui/context.rs
@@ -318,9 +318,8 @@ impl TUIContext<'_> {
             }
 
             // Go to previous call
-            KeyCode::Char('c') => {
-                self.draw_memory.inner_call_index =
-                    self.draw_memory.inner_call_index.saturating_sub(1);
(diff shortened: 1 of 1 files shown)


mablr (approved):

stevencartavia (approved):

mattsse (approved):
lgtm

PR #15735, opened 2026-07-14
fix(forge): skip libraries during test discovery

### LINKED ISSUE #10418: forge: test/coverage failures for libs with `invariant*` functions
### Component

Forge

- [ ] Foundry
- [ ] Foundryup

_No response_

_No response_

_No response_

### Operating System

None

### Describe the bug

for an InternalLib
```Solidity
library InternalLib {   
    function invariantProtocol() public pure returns (uint256 code) {
        return 1;
    }  
}
```

with test
```Solidity
contract InternalLibTest is Test {
    function testInternalLibInvariantProtocol() public {
        assertEq(InternalLib.invariantProtocol(), 1);
    }
}
```

`forge test` fails with `[FAIL: failed to set up invariant testing environment: No contracts to fuzz.]`

If `invariantProtocol` is made internal then it works but there's no coverage reported when `forge coverage`

If renaming `invariantProtocol` to `_invariantProtocol` then both works. happens because `invariant*` fns are reserved and they conflict when inlined in test contract

figtracer:
Library artifacts are now excluded from Forge's deployable test-contract map using Solar's actual contract kind, while remaining available for linking, deployment, revert decoding, and coverage attribution. This prevents public library functions named `invariant*` or `test*` from being executed as test suites. Compile-output source paths are resolved from the project root so the same classification works after coverage strips artifact prefixes.

Closes #10418.

### DIFF
--- a/crates/forge/src/multi_runner.rs
+++ b/crates/forge/src/multi_runner.rs
@@ -11,7 +11,10 @@ use crate::{
     symbolic_regression::SYMBOLIC_REGRESSION_MARKER,
 };
 use alloy_json_abi::{Function, JsonAbi};
-use alloy_primitives::{Address, Bytes, U256};
+use alloy_primitives::{
+    Address, Bytes, U256,
+    map::{HashMap, HashSet},
+};
 use eyre::Result;
 use foundry_cli::opts::configure_pcx_from_compile_output;
 use foundry_common::{
@@ -41,6 +44,7 @@ use foundry_evm_networks::NetworkVariant;
 
 use foundry_linking::{LinkOutput, Linker};
 use rayon::prelude::*;
+use solar::sema::interface::source_map::FileName;
 use std::{
     borrow::Borrow,
     collections::BTreeMap,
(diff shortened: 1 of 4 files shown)


crates/forge/src/multi_runner.rs
```diff
@@ -755,6 +759,50 @@ impl MultiContractRunnerBuilder {
...
+        ));
+        dcx.set_flags_mut(|f| f.track_diagnostics = false);
+
+        // Populate solar's global context by parsing and lowering the sources.
+        let files: Vec<_> = output.output().sources.as_ref().keys().cloned().collect();
+        let mut library_contracts = HashMap::<PathBuf, HashSet<String>>::default();
+
+        analysis.enter_mut(|compiler| -> Result<()> {
+            let mut pcx = compiler.parse();
+            configure_pcx_from_compile_output(
```
mattsse, line 803: Solar cannot classify libraries compiled with Solidity versions before 0.8 because `get_solar_sources_from_compile_output` skips every artifact below `MIN_SOLIDITY_VERSION`. I reproduced this on the PR head with `solc = "0.7.6"`: Forge still discovered `InternalLib.invariantProtocol()` as an invariant suite and failed with `No contracts to fuzz`. Please derive the library kind from compiler output or AST data that covers every Solidity version supported by Forge, or add a non-Solar fallback for unsupported versions. A 0.7.x variant of this regression test would prevent the gap from recurring.

mattsse (changes requested):
The fix is incomplete for Solidity versions that Solar does not support. Please address the inline finding and add legacy-version coverage before merging.

DaniPopes:
is this worth it? i would close this as wontfix rather than add this

figtracer:
> is this worth it? i would close this as wontfix rather than add this

hm change might be too invasive for what we need will retarget

PR #15051, opened 2026-06-03
feat(tempo): expose T5 payment lane classifier

figtracer:
## Summary

- add a reusable Foundry payment-lane classification adapter around Tempo's T5 `is_payment_v2()` classifier
- expose `anvil_classifyTransaction(rawTx)` and log classification during `eth_sendRawTransaction` without affecting admission, ordering, pricing, or execution
- add local Cast surfaces via `cast classify <rawTx>` and `cast tx --lane`

## Notes
- Cheatcodes such as `vm.assumePayment()` / `vm.assumeGeneral(...)` are intentionally deferred for now because the cheatcode API doesn't fit for this since it doesn't carry the signed raw transaction envelope needed for full top-level classification (access lists, authorization lists, AA calls, tempo authorization list, key authorization RLP size). Maybe we should extend it?

PR #15736, opened 2026-07-14
refactor(tempo): move payment-lane classification out of foundry-primitives

0xMars42:
## Motivation

#15051 added the T5 payment-lane classifier to `foundry-primitives`: a `clone_into_tempo_envelope` helper that clones every transaction variant just to classify it, a classification wrapper, and three serde output types. The classification itself only needs `tempo_primitives::TempoTxEnvelope::is_payment_v2`, and the `PaymentLane*` types are output shapes for the anvil RPC and cast JSON, so `foundry-primitives` is not their natural home.

## Solution

Move the payment-lane types and classification into `foundry_common::tempo` (`crates/common/src/tempo/lane.rs`), a module both anvil and cast already depend on, so there are no dependency changes. The classifier takes the raw EIP-2718 bytes and decodes them with `TempoTxEnvelope::decode_2718` directly, which removes the need for the cloning conversion. `foundry-primitives` returns to its pre-#15051 surface.

Behavior is unchanged:
- every call site (`anvil_classifyTransaction`, `cast classify`, `cast tx --lane`) keeps its signed transaction validation and fails on undecodable input with the same errors;
- transaction types without a Tempo envelope variant (EIP-4844, Optimism deposits) do not decode as a Tempo envelope and still classify as `unsupported_transaction_type`.

Also adds unit tests for the classifier and its serialized JSON shape, which had no coverage.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mablr (approved):
Definitely cleaner ! Thanks

figtracer (approved):

PR #15739, opened 2026-07-14
perf(anvil): speed up block mining

mattsse:
This removes three recurring costs from Anvil block mining: prevrandao input now uses fixed stack storage, sequential block-hash eviction uses a tracked head instead of rescanning the 257-entry EVM window, and incremental trie encoding reuses RLP scratch storage while keeping branch children on the stack. Out-of-order block-hash writes retain the existing full-scan fallback, so snapshot and revert behavior remains unchanged. This PR was developed with AI assistance.

### Results

| 2,500 single-transfer blocks | `master` | This PR | Improvement |
| --- | ---: | ---: | ---: |
| Mean wall time | 163.2 ms | 142.9 ms | 12.4% |
| Median wall time | 161.9 ms | 142.0 ms | 12.3% |


stevencartavia (dismissed):

mablr (approved):

grandizzy (approved):
lgtm

PR #15741, opened 2026-07-14
feat(script): include full signatures in broadcast logs

### LINKED ISSUE #3887: In broadcast log, include arg names and use standard syntax
### Component

Forge

## Description
For now the `function` field of a broadcast log's transaction looks like:

```solidity
"function": "functionName(type1,type2):(returnType)",
```

More useful and more standard (ie parseable by ethers.js' human-readable ABI parser) would be :
```solidity
"function": "functionName(type1,type2 name2) returns (returnType)",
```

Differences:
* The name of the argument is there when available (useful when the broadcat log is an input to a UI) (keeps more info)
* `:` becomes `returns` (helps with parsing)

## Alternative
Stay in full json and have something like
```json
{
  "name":"functionName",
  "payable":false,
  "inputs":[{"type":"type1"},{"type":"type2", "name":"name2"}],
  "outputs":[{"type":"returnType"}]
}
```

### Additional context

_No response_

figtracer:
## Motivation

Broadcast logs expose selector-oriented function metadata in the `function` field, which drops argument and return names and cannot be reused as a complete human-readable ABI fragment. Changing that existing field would break consumers that expect the canonical `name(types)` format.

## Solution

Keep `function` unchanged and add an optional `functionAbi` field containing Alloy's full human-readable declaration, including parameter names, mutability, and standard `returns` syntax. Terminal display metadata remains separate so decoded calls render as calls rather than declaration/call hybrids.

Closes #3887.

> This PR was written with Codex.

### DIFF
--- a/crates/script/src/transaction.rs
+++ b/crates/script/src/transaction.rs
@@ -67,7 +67,8 @@ impl<N: Network> ScriptTransactionBuilder<N> {
                 };
 
                 if let Some(function) = function {
-                    self.transaction.function = Some(function.full_signature());
+                    self.transaction.function = Some(function.signature());
+                    self.transaction.function_abi = Some(function.full_signature());
                     self.transaction.display_function = Some(function.name.clone());
 
                     let values = function.abi_decode_input(data).inspect_err(|_| {
--- a/crates/script/src/transaction.rs
+++ b/crates/script/src/transaction.rs
@@ -68,6 +68,7 @@ impl<N: Network> ScriptTransactionBuilder<N> {
 
                 if let Some(function) = function {
                     self.transaction.function = Some(function.full_signature());
+                    self.transaction.display_function = Some(function.name.clone());
 
                     let values = function.abi_decode_input(data).inspect_err(|_| {
                         error!(
--- a/crates/script/src/transaction.rs
+++ b/crates/script/src/transaction.rs
@@ -67,7 +67,7 @@ impl<N: Network> ScriptTransactionBuilder<N> {
                 };
 
                 if let Some(function) = function {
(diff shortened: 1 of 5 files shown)


mattsse (commented):
The full signature is also reused by format_transaction, which appends a second argument list. This changes decoded dry-run output into an invalid declaration/call hybrid such as `data (decoded): function configure(...) returns (bool accepted)(`; the updated dry-run snapshot currently codifies the regression.

Please decouple the persisted ABI fragment from the terminal display representation. For example, keep full_signature() in the broadcast artifact, but retain separate function metadata (or a call-style name/signature) for format_transaction and render the decoded values as `configure(1, address(...))`. A display test using a function with return values would cover the problematic shape.

mattsse (dismissed):
lgtm

crates/script/src/transaction.rs
```diff
@@ -67,7 +67,8 @@ impl<N: Network> ScriptTransactionBuilder<N> {
                 };
 
                 if let Some(function) = function {
-                    self.transaction.function = Some(function.signature());
+                    self.transaction.function = Some(function.full_signature());
```
stevencartavia: Maybe we should keep `function` unchanged and add an optional `functionAbi` field for `full_signature()` instead, preserving the existing field’s semantics for consumers that expect the canonical `name(types)` format?

figtracer: yeah we can do this instead of making it breaking

stevencartavia: it just needs a test update

stevencartavia (approved):

grandizzy (approved):

PR #15744, opened 2026-07-14
feat(cast): support balance call overrides

mattsse:
Adds a shared `CallOverrideOpts` for Cast `eth_call` commands and uses it for both `cast erc20-token balance` and the deprecated `cast balance --erc20` path. State and block overrides now compose with the selected block, while native balance queries reject these flags because `eth_getBalance` cannot apply them. This makes balance queries capable of simulating account state and block context without changing their existing output behavior.

The implementation and tests were written with OpenAI Codex assistance.


stevencartavia (approved):

grandizzy (approved):
lgtm

PR #15745, opened 2026-07-14
feat(cast): extend call block overrides

mattsse:
Adds block gas limit, fee recipient, base fee, and blob base fee options to `cast call`, applying them consistently to regular calls, remote debug traces, local traces, and generated curl requests. Outbound block overrides use the execution-client field names, and ordinary curl requests now retain supplied state and block overrides.

### DIFF
--- a/crates/cast/src/cmd/call.rs
+++ b/crates/cast/src/cmd/call.rs
@@ -1,6 +1,6 @@
 use super::run::{fetch_contracts_bytecode_from_trace, fetch_contracts_bytecode_via_rpc};
 use crate::{
-    Cast,
+    Cast, RpcBlockOverrides,
     debug::handle_traces,
     rpc_trace::{call_frame_to_arena, is_method_not_found_error, is_missing_state_error},
     traces::TraceKind,
@@ -12,7 +12,7 @@ use alloy_primitives::{
     Address, B256, Bytes, TxKind, U256, hex,
     map::{AddressHashMap, HashMap},
 };
-use alloy_provider::{Provider, ext::DebugApi};
+use alloy_provider::Provider;
 use alloy_rpc_types::{
     BlockId, BlockNumberOrTag, BlockOverrides,
     state::{StateOverride, StateOverridesBuilder},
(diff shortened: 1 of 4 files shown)


crates/cast/src/cmd/call.rs
```diff
@@ -523,12 +593,36 @@ impl CallArgs {
...
             }
 
-            let trace = match tx_kind {
-                TxKind::Create => {
+            // The test executor normally zeros the block base fee together with the transaction
+            // gas price. An explicit call override must remain visible to the BASEFEE opcode, so
+            // execute with the configured environments in that case.
+            let call_env = preserve_block_base_fee.then(|| {
+                let mut evm_env = executor.evm_env().clone();
+                evm_env.cfg_env.disable_balance_check = true;
```
stevencartavia, line 601: Could this call environment also disable REVM’s base-fee check? With `--block.base-fee 100 --gas-price 2`, the regular call succeeds and returns `100`, while the same call with `--trace` fails with `gas price is less than basefee`. A regression test for this case would be useful.

crates/evm/core/src/env.rs
```diff
@@ -52,6 +53,9 @@ pub trait FoundryBlock: Block {
         _base_fee_update_fraction: u64,
     );
 
+    /// Sets the blob gas price directly.
+    fn set_blob_gasprice(&mut self, _blob_gasprice: u128) {}
```
stevencartavia, line 57: Could `TempoBlockEnv` implement the new blob-gas-price setter instead of using the default no-op? With a Tempo test node, `cast call --block.blob-base-fee 4660` returns `4660`, while the same call with `--trace` returns `1`. A Tempo parity test would catch this.

crates/cast/src/cmd/call.rs
```diff
@@ -523,12 +593,36 @@ impl CallArgs {
...
+                tx_env.set_data(input.clone());
+                tx_env.set_value(value);
+                tx_env.set_gas_limit(executor.gas_limit());
+                tx_env.set_chain_id(Some(evm_env.cfg_env.chain_id));
+                (evm_env, tx_env)
+            });
+
+            let trace = match (tx_kind, call_env) {
+                (TxKind::Create, Some((evm_env, tx_env))) => {
+                    let deploy_result = executor.deploy_with_env(evm_env, tx_env, None)?;
```
stevencartavia, line 614: Could this branch preserve traces from reverting constructors like the non-override path does? `cast call --trace --create 0x60006000fd` renders the CREATE/Revert trace, while adding `--block.base-fee 0` exits with `execution reverted` without rendering it. Using `TraceResult::try_from` around `deploy_with_env` should retain the failed deployment trace, and a regression test for this case would be useful.

crates/cast/src/cmd/call.rs
```diff
@@ -435,6 +491,8 @@ impl CallArgs {
         }
 
         if trace {
+            let preserve_block_base_fee =
+                block_overrides.as_ref().is_some_and(|overrides| overrides.base_fee.is_some());
```
grandizzy, line 495: Could this env-preserving path also cover block gas-limit overrides? Otherwise `transact_raw` copies `Executor.gas_limit` into the block environment, so `--gas-limit` overwrites `--block.gas-limit`.

```suggestion
            let preserve_block_base_fee = block_overrides.as_ref().is_some_and(|overrides| {
                overrides.base_fee.is_some() || overrides.gas_limit.is_some()
            });
```

Please also add a regression test combining `--block.gas-limit 100000` with `--gas-limit 5000000` and asserting that `GASLIMIT` returns `100000`.

grandizzy (changes requested):
**Request changes:** The RPC paths look good, but local trace parity still has unresolved block-environment issues. In addition to the existing comments, `--gas-limit` overwrites `--block.gas-limit`; please preserve the block environment independently and add coverage for this combination.

PR #15748, opened 2026-07-15
refactor(cast): deduplicate Tempo boilerplate

0xMars42:
## Motivation

Cast-side counterpart of #15641. The Tempo commands in cast repeat the same blocks: the payload query commands each branch the same way between JSON and human output, the keychain hardfork gates repeat the same check-and-bail block seven times, and `send_tip20_transaction` copies its sponsor attachment and signer resolution across its send paths.

## Solution

Three mechanical extractions into the shared cast tempo module, behavior unchanged:

- `print_payload` owns the JSON-or-human output branch shared by the payload query commands (tip403 info/check, receive-policy get/validate/receipt, storage-credits balance/mode/budget).
- `ensure_tempo_precompile_active` merges the near-identical T6/T7 precompile guards of the receive-policy and storage-credits commands, code-presence fallback included.
- `attach_sponsor` and `resolve_send_signer` own the sponsor fee-token-and-attach block (copied in three of the four send paths) and the signer resolution (copied in two). The four send paths themselves stay separate, they dispatch differently.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cast/src/tempo.rs
+++ b/crates/cast/src/tempo.rs
@@ -4,13 +4,36 @@ use crate::tx::fill_transaction_gas_fees;
 use alloy_network::{Network, TransactionBuilder};
 use alloy_provider::Provider;
 use eyre::Result;
-use foundry_cli::opts::TempoOpts;
-use foundry_common::FoundryTransactionBuilder;
-use foundry_config::{Chain, Eip1559FeeEstimatePreset};
+use foundry_cli::{json::print_json_success, opts::TempoOpts};
+use foundry_common::{
+    FoundryTransactionBuilder,
+    provider::{ProviderBuilder, RetryProvider},
+    shell,
+};
+use foundry_config::{Chain, Config, Eip1559FeeEstimatePreset};
 use foundry_wallets::{TempoAccessKeyConfig, WalletOpts, WalletSigner};
+use serde_json::Value;
 use tempo_alloy::TempoNetwork;
 
 pub use foundry_common::tempo::{TempoSponsor, TempoSponsorPreview, resolve_tempo_sponsor_signer};
+
+/// Builds the Tempo network provider from a loaded config.
+pub(crate) fn tempo_provider(config: &Config) -> Result<RetryProvider<TempoNetwork>> {
+    ProviderBuilder::<TempoNetwork>::from_config(config)?.build()
+}
+
+/// Prints a command result: the raw payload in JSON mode, the human rendering otherwise.
+pub(crate) fn print_payload<F>(payload: Value, human: F) -> Result<()>
+where
+    F: FnOnce(&Value) -> Result<()>,
+{
+    if shell::is_json() {
+        print_json_success(payload)?;
+    } else {
+        human(&payload)?;
+    }
+    Ok(())
+}
 
 pub(crate) fn print_expires(expires_at: Option<u64>) -> Result<()> {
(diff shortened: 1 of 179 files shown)


stevencartavia (dismissed):

crates/cast/src/tempo.rs
```diff
@@ -4,14 +4,37 @@ use crate::tx::fill_transaction_gas_fees;
...
+
+/// Prints a command result: the raw payload in JSON mode, the human rendering otherwise.
+pub(crate) fn print_payload<F>(payload: Value, human: F) -> Result<()>
+where
+    F: FnOnce(&Value) -> Result<()>,
+{
+    if shell::is_json() {
+        print_json_success(payload)?;
+    } else {
+        human(&payload)?;
```
figtracer, line 24: i'm not sure we even want this anymore since we should be getting rid of this machine/human distinction

crates/cast/src/tempo.rs
```diff
@@ -4,14 +4,37 @@ use crate::tx::fill_transaction_gas_fees;
...
+};
+use foundry_config::{Chain, Config, Eip1559FeeEstimatePreset};
 use foundry_wallets::{TempoAccessKeyConfig, WalletOpts, WalletSigner};
+use serde_json::Value;
 use tempo_alloy::TempoNetwork;
 
 pub use foundry_common::tempo::{TempoSponsor, TempoSponsorPreview, resolve_tempo_sponsor_signer};
 
+/// Builds the Tempo network provider from a loaded config.
+pub(crate) fn tempo_provider(config: &Config) -> Result<RetryProvider<TempoNetwork>> {
```
figtracer: i think we don need an abstraction for this

mablr: I agree

0xMars42: Dropped, the call sites are back to the inline construction.

crates/cast/src/tempo.rs
0xMars42: This one only consolidates the three existing copies of the branching, so when the machine/human distinction goes away it is a single place to delete. I can also strip the human output of these three commands here if you prefer.

mablr: There is no more machine/human distinction ig

0xMars42: Removed the human rendering, these commands now always print the JSON payload. The set --preview branch in receive-policy predates this PR so I left it alone.

mablr: There is misunderstanding on https://github.com/foundry-rs/foundry/pull/15748/commits/5d227d0f87ee94f2f40fe87c96076c7c2d154187, we should not change behavior.

0xMars42: Reverted 5d227d0f8. The payload commands render as they do on master again (JSON under `--json`, human-readable otherwise), so the PR no longer changes behavior and is back to the three mechanical extractions.

That does bring back `print_payload`, which now consolidates the shared JSON-or-human branch in a single place. Happy to inline that branch back into the three commands and drop the helper if you would rather not keep it, still without any behavior change.

mablr (approved):
Makes sense

stevencartavia (approved):

mattsse (approved):

PR #15750, opened 2026-07-15
fix(debugger): clamp pane scroll positions

figtracer:
## Motivation

Pane scroll offsets persist while navigating between debugger steps and calls. When the current stack or active buffer is shorter than the previous one, the stale offset can skip every available row and make a non-empty pane appear blank.

## Solution

- centralize post-navigation scroll position updates
- clamp stack and active-buffer offsets to shorter non-empty data
- preserve memory-write autoscroll and empty-snapshot behavior

Progresses #927.

### DIFF
--- a/crates/debugger/src/tui/context.rs
+++ b/crates/debugger/src/tui/context.rs
@@ -307,21 +307,21 @@ impl TUIContext<'_> {
             KeyCode::Char('g') => {
                 self.draw_memory.inner_call_index = 0;
                 self.current_step = 0;
-                self.scroll_memory_to_current_write();
+                self.update_scroll_positions();
             }
 
             // Go to bottom of file
             KeyCode::Char('G') => {
                 self.draw_memory.inner_call_index = self.debug_arena().len() - 1;
                 self.current_step = self.n_steps() - 1;
-                self.scroll_memory_to_current_write();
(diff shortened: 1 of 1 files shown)


grandizzy (dismissed):
lgtm

crates/debugger/src/tui/context.rs
```diff
@@ -938,17 +938,27 @@ impl TUIContext<'_> {
...
         }
-        self.scroll_memory_to_current_write();
+        self.update_scroll_positions();
     }
 
-    fn scroll_memory_to_current_write(&mut self) {
-        if self.active_buffer != BufferKind::Memory {
-            return;
+    fn update_scroll_positions(&mut self) {
+        if let Some(stack) = &self.current_step().stack {
```
stevencartavia: Could we skip clamping when the captured stack is empty? Real debugger traces contain empty `stack: []` snapshots, and navigating through one resets the saved stack offset to `0`. This loses the user’s scroll position when navigating directly to a later non-empty stack. Adding `&& !stack.is_empty()` preserves the offset across empty snapshots while still clamping shorter non-empty stacks.

mablr (approved):

grandizzy (approved):

Ari4ka (approved):

PR #15751, opened 2026-07-15
perf(forge): skip redundant submodule status

figtracer:
## Summary

Skip the expensive `git submodule status` process when `.gitmodules` and the matching `.git` worktree markers already prove that the requested dependencies are initialized. Missing, unreadable, and ambiguous cases still fall back to the existing status check, preserving automatic dependency installation. Codex assisted with profiling, implementation, benchmarking, and test iteration.

## Benchmarks

Profiling builds on `Recon-Fuzz/aave-v4-scfuzzbench@v0.5.6-recon`, measured by `foundry-bench`/hyperfine over five cached `forge build` runs:

| --- | ---: | ---: | ---: |
| Cached `forge build` | 170.8 ± 3.0 ms | 97.5 ± 5.2 ms | -42.9% (1.75x faster) |

### DIFF
--- a/crates/cli/src/utils/mod.rs
+++ b/crates/cli/src/utils/mod.rs
@@ -644,6 +644,13 @@ ignore them in the `.gitignore` file."
         let Some(root) = self.root.ancestors().find(|root| root.join(".git").exists()) else {
             return Ok(false);
         };
+        if paths.iter().any(|path| {
+            Path::new(path)
+                .components()
+                .any(|component| matches!(component, std::path::Component::ParentDir))
+        }) {
+            return Ok(false);
+        }
         let relative_root = self.root.strip_prefix(root).unwrap_or_else(|_| Path::new(""));
         let gitmodules = root.join(".gitmodules");
         if !gitmodules.is_file() {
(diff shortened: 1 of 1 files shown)


mattsse (dismissed):
lgtm

mablr (dismissed):

crates/cli/src/utils/mod.rs
```diff
@@ -627,13 +627,56 @@ ignore them in the `.gitignore` file."
...
             .get_stdout_lossy()
             .map(|stdout| stdout.lines().any(|line| line.starts_with('-')))
     }
 
+    /// Returns true if all submodules matching `paths` have initialized worktrees.
+    fn submodules_initialized(self, paths: &[OsString]) -> Result<bool> {
+        let Some(root) = self.root.ancestors().find(|root| root.join(".git").exists()) else {
+            return Ok(false);
+        };
+        let relative_root = self.root.strip_prefix(root).unwrap_or_else(|_| Path::new(""));
```
grandizzy, line 654: This comparison is lexical, so a nested project with `libs = ["../../lib"]` does not match a submodule at `lib/forge-std`. The helper then returns `true` and skips `git submodule status`, even when that submodule is uninitialized. Git’s existing check correctly handles `../../lib`.

Conservatively fall back whenever a requested path requires parent normalization:

```suggestion
        let relative_root = self.root.strip_prefix(root).unwrap_or_else(|_| Path::new(""));
        if paths.iter().any(|path| {
            Path::new(path)
                .components()
                .any(|component| matches!(component, std::path::Component::ParentDir))
        }) {
            return Ok(false);
        }
```

grandizzy (changes requested):
The fast path can skip automatic dependency installation for parent-relative library paths in monorepos.

grandizzy (approved):
lgtm

mablr (approved):

PR #15752, opened 2026-07-15
fix(chisel): display raw dynamic bytes

### LINKED ISSUE #5253: chisel: ABI-encoded values incorrectly displayed when dumped

figtracer:
Dynamic `bytes` inspection currently ABI-encodes the decoded value a second time before formatting it, so Chisel shows a synthetic outer memory and tuple layout instead of the actual bytes. This renders dynamic bytes directly as hex data, matching fixed bytes and Forge's `console.logBytes`, while retaining the existing rich display for strings. The regression uses the exact `abi.encode`, `abi.decode`, and `bytes(...)` case from #5253.

Closes #5253.

### DIFF
--- a/crates/chisel/src/executor.rs
+++ b/crates/chisel/src/executor.rs
@@ -279,17 +279,20 @@ fn format_token(token: DynSolValue) -> String {
         DynSolValue::Bool(b) => {
             format!("Type: {}\n└ Value: {}", "bool".red(), b.cyan())
         }
-        DynSolValue::String(_) | DynSolValue::Bytes(_) => {
-            let hex = hex::encode(token.abi_encode());
-            let s = token.as_str();
+        DynSolValue::Bytes(bytes) => {
             format!(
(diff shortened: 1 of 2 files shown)


grandizzy (approved):
lgtm

stevencartavia (approved):

PR #15753, opened 2026-07-15
feat(debugger): add storage explorer

figtracer:
## Motivation

#927 asks for a storage pane that can inspect larger contract state, including transient storage. Today the debugger can describe only the `SLOAD`/`SSTORE`/`TLOAD`/`TSTORE` at the current step and jump to an access by slot, so users cannot scan the slots already observed in the current call.

## Solution

- turn the existing data pane into a persistent or transient storage explorer through `:storage` and `:transient`
- list slots accessed in the current logical call up to the current replay step, preserving first-access order while showing each slot's latest observed value
- keep `:storage <slot>` and `:transient <slot>` navigation, then focus the selected slot in the explorer
- reuse data-pane scrolling, `b` to return to the previous byte buffer, and `o` to prefill a slot query

The explorer intentionally reflects recorded accesses, not an exhaustive query of unaccessed EVM state.

## Testing

- live `forge test --debug` smoke test covering persistent and transient slot selection, scrolling, command prefill, and return to memory

Progresses #927.

### DIFF
--- a/crates/debugger/src/tui/context.rs
+++ b/crates/debugger/src/tui/context.rs
@@ -1,8 +1,10 @@
 //! Debugger context and event handler implementation.
 
-use super::storage::{StorageAccess, StorageSpace, hex_u256, storage_access_at};
+use super::storage::{
+    StorageAccess, StorageSpace, hex_u256, storage_access_at, storage_accesses_until,
+};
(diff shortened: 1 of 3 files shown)


mattsse (dismissed):
lgtm

mablr (dismissed):

mablr (approved):

mattsse (approved):

PR #15754, opened 2026-07-15
perf(cast): speed up read-only commands

figtracer:
## Summary

Read-only `CastTxBuilder` calls currently fetch an account nonce even when no authorization needs it, then discard the result because raw transactions do not fill nonce fields. This skips that unused `eth_getTransactionCount` request for `cast call`, `cast estimate`, and `cast access-list`; transaction sends and every authorization path retain the existing nonce resolution. A mocked-provider regression test queues no responses, so any RPC during a raw build fails the test. This PR was developed with OpenAI Codex assistance.

## Benchmarks

Measured exact `master` (`5b9ce7810`) and PR profiling binaries against a deterministic local JSON-RPC endpoint that delays every response by 100ms. Hyperfine used 3 warmups and 20 measured runs; command output was identical before and after. The only removed request was `eth_getTransactionCount`.

| Command | RPC requests | `master` | This PR | Speedup |
| `cast call` | 3 → 2 | 336.18ms ± 6.51ms | 231.93ms ± 3.69ms | 1.45× |
| `cast estimate` | 2 → 1 | 229.85ms ± 5.21ms | 123.11ms ± 0.62ms | 1.87× |
| `cast access-list` | 2 → 1 | 230.02ms ± 4.15ms | 122.54ms ± 2.61ms | 1.88× |

### DIFF
--- a/crates/cast/src/cmd/storage.rs
+++ b/crates/cast/src/cmd/storage.rs
@@ -1,7 +1,7 @@
 use crate::{Cast, opts::parse_slot};
 use alloy_ens::NameOrAddress;
 use alloy_network::AnyNetwork;
-use alloy_primitives::{Address, B256, U256};
+use alloy_primitives::{Address, B256, Bytes, U256};
 use alloy_provider::Provider;
 use alloy_rpc_types::BlockId;
 use clap::Parser;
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm

grandizzy (approved):
lgtm

PR #15756, opened 2026-07-15
fix(cast): compute correct topic for indexed string/bytes in abi-encode-event

Kropiunig:
## Motivation

`cast abi-encode-event` computes an incorrect event topic for indexed `string` and `bytes` parameters.

In `SimpleCast::abi_encode_event`, the indexed branch hashes `token.abi_encode()` for every dynamic type (`String`, `Bytes`, `Array`, `Tuple`). For `string`/`bytes` this hashes the standard ABI encoding (a 32-byte offset word + 32-byte length word + right-padded data). But the EVM stores the topic of an indexed `string`/`bytes` as `keccak256` of the **raw, unpadded contents**, with no offset/length words.

### Failing scenario

```
cast abi-encode-event "Log(string indexed message, uint256 data)" hello 42
```

- Current output: `topic1 = 0x984002fcc0ca639f96622add24c2edd2fe72c65e71ca3faa243e091e0bc7cdab` (= `keccak256(abi.encode("hello"))`)
- Actual EVM topic for `emit Log("hello", 42)`: `topic1 = keccak256("hello") = 0x1c8aff950685c2ed4bc3174f3472287b56d9517b9c948127319a09a7a36deac8`

So any log filter or event-topic verification built from this command's output silently never matches real logs. Indexed `bytes` is wrong the same way.

## Solution

For indexed `string`/`bytes`, hash `token.abi_encode_packed()` (the raw bytes, no offset/length padding), which yields the correct topic. Dynamic arrays/tuples keep hashing the ABI encoding and fixed-size types keep their 32-byte encoding, so those paths are unchanged.

Adds a regression test: updates the existing `abi_encode_event_dynamic_indexed` snapshot to the correct `keccak256("hello")` value and adds an `abi_encode_event_dynamic_indexed_bytes` case asserting `topic1 == keccak256(0xdeadbeef)`.


mablr (dismissed):
Makes sense

figtracer:
fully superseded by merged [#16049](https://github.com/foundry-rs/foundry/pull/16049)

PR #15757, opened 2026-07-15
fix(bind): support large serde arrays

### LINKED ISSUE #10911: Generated Rust bindings for large fixed-size arrays causes Serde Deserialize error
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.2.3-nightly Commit SHA: ab0768ae494fa4aa902596c03ef226e38b4771a0 Build Timestamp: 2025-07-03T07:17:37.071084000Z (1751527057) Build Profile: maxperf

foundryup: 1.1.0

forge bind --alloy

### Operating System

None

### Describe the bug

Hi, provided some solidity code like this:

```solidity
    function getFoobar() external view returns (uint256[48] memory) {
      // ...
    }
```

the following bindings code will be generated:
```rust
    ///Container type for the return parameters of the [`getFoobar()`](getFoobarCall) function.
    #[allow(non_camel_case_types, non_snake_case, clippy::pub_underscore_fields)]
    #[derive(Clone)]
    pub struct getFoobarReturn {
        #[allow(missing_docs)]
        pub _0: [alloy::sol_types::private::primitives::aliases::U256; 48usize],
    }
```

which will fail to compile:

```shell
5346 |     #[derive(serde::Serialize, serde::Deserialize)]
     |                                ^^^^^^^^^^^^^^^^^^ the trait `Deserialize<'_>` is not implemented for `[alloy::alloy_primitives::Uint<256, 4>; 48]`
```

I believe this is because [serde doesn't support arrays larger than 32](https://stackoverflow.com/questions/62665558/how-can-i-implement-serdedeserialize-for-arrays-larger-than-32)

the solution would be to use `serde-big-array`, or add a custom deserialization implementation

figtracer:
Generated Rust bindings derive Serde traits by default, but Serde only implements fixed arrays through 32 elements. A Solidity `uint256[48]` therefore produces a generated crate that cannot compile.

This teaches Foundry's binding crate writer to add ordinary Serde `with` adapters only to expanded struct fields whose fixed-array or `Vec` shape needs them, and to add `serde_with` only for affected outputs. Keeping the integration in Foundry is deliberate: Foundry opts into the extra Serde derives and owns the generated `Cargo.toml`, while Alloy's expansion API has no dependency hook and drops source field attributes for some unnamed function and error fields.

The regression covers errors, events, function inputs, direct and dynamically nested outputs, and struct-wrapped arrays. It also uses an interface named `serde_with` to guard against dependency-name shadowing and exercises the existing binding consistency check.

Closes #10911.

### DIFF
--- a/crates/forge/tests/cli/bind.rs
+++ b/crates/forge/tests/cli/bind.rs
@@ -163,3 +163,33 @@ Bindings have been generated to [..]
     let bindings_path = prj.root().join("out/bindings");
     assert_bindings_compile(&bindings_path);
 });
+
+// <https://github.com/foundry-rs/foundry/issues/10911>
+forgetest!(bind_large_fixed_arrays, |prj, cmd| {
+    prj.add_source(
+        "BigArrays.sol",
+        r#"
+interface BigArrays {
+    function fixedArray() external returns (uint256[48] memory);
+    function nestedArray() external returns (uint256[48][] memory);
+}
+"#,
+    );
+
+    cmd.args(["bind", "--select", "^BigArrays$"])
+        .assert_success()
+        .stdout_eq(str![[r#"
+[COMPILING_FILES] with [SOLC_VERSION]
+[SOLC_VERSION] [ELAPSED]
+Compiler run successful!
+
+"#]])
+        .stderr_eq(str![[r#"
+Generating bindings for 1 contracts
+Bindings have been generated to [..]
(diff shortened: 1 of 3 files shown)


mattsse (approved):
lgtm

mablr (approved):
sgtm

PR #15758, opened 2026-07-15
fix(fmt): indent chained named call args

### LINKED ISSUE #15755: forge fmt weird identation
### Component

Forge

- [x] Foundry
- [x] Foundryup

_No response_

_No response_

_No response_

### Operating System

macOS (Apple Silicon)

### Describe the bug

When a call using named arguments is long enough that the formatter breaks the member-access chain (Iface(addr) / newline / .call({...})), the named-argument block loses its indentation: arguments are indented relative to the statement instead of the call, and the closing }) dedents past the block it closes.

This is the call-arguments sibling of #13362, which was fixed for struct literals nested in chained calls by #13367 (merged May 2026). Named call arguments still reproduce on the current nightly with an all-default [fmt] config.

Actual output of forge fmt (function body):

```solidity
try IVault(vault)
    .updatePosition({
    positionId_: positionId,
    batchId_: batchId,
    account_: account,
    operator_: operator,
    amount_: amount,
    deadline_: deadline,
    data_: data
}) {}
    catch {}
```

Expected output (mirroring the post-#13367 behavior for struct literals):

```solidity
try IVault(vault).updatePosition({
    positionId_: positionId,
    batchId_: batchId,
    account_: account,
    operator_: operator,
    amount_: amount,
    deadline_: deadline,
    data_: data
}) {} catch {}
```

mablr:
## Motivation

Close #15755.

Keep fitting chained callees together so their multiline named arguments retain their own indentation. Preserve inline empty catch clauses for the same call shape and add formatter regression coverage.

### DIFF
--- a/crates/fmt/src/state/sol.rs
+++ b/crates/fmt/src/state/sol.rs
@@ -1315,7 +1315,7 @@ impl<'ast> State<'_, 'ast> {
                 .then(|| ChainedNamedCall {
                     callee: call_expr.span,
                     keep_inline: !call_chain_contains_options(call_expr)
-                        && !self.call_chain_contains_arg_comments(call_expr)
+                        && !self.has_comment_between(call_expr.span.lo(), call_expr.span.hi())
                         && self.estimate_size(call_expr.span)
                             + if call_args.is_empty() { 4 } else { 2 }
                             <= self.space_left(),
@@ -1448,19 +1448,6 @@ impl<'ast> State<'_, 'ast> {
             ast::ExprKind::Err(_) => self.print_span(span),
         }
         self.cursor.advance_to(span.hi(), true);
-    }
-
-    fn call_chain_contains_arg_comments(&self, expr: &'ast ast::Expr<'ast>) -> bool {
-        match &expr.peel_parens().kind {
-            ast::ExprKind::Call(expr, args) => {
-                self.has_comments_between_elements(args.span, args.exprs())
-                    || self.call_chain_contains_arg_comments(expr)
-            }
-            ast::ExprKind::CallOptions(expr, ..)
-            | ast::ExprKind::Index(expr, ..)
-            | ast::ExprKind::Member(expr, ..) => self.call_chain_contains_arg_comments(expr),
-            _ => false,
-        }
     }
 
     /// Prints a simple assignment expression of the form `lhs = rhs`.
--- a/crates/fmt/src/state/sol.rs
+++ b/crates/fmt/src/state/sol.rs
@@ -1314,9 +1314,13 @@ impl<'ast> State<'_, 'ast> {
                     && is_call_chain(&call_expr.kind, true))
                 .then(|| ChainedNamedCall {
                     callee: call_expr.span,
-                    keep_inline: self.estimate_size(call_expr.span)
-                        + if call_args.is_empty() { 4 } else { 2 }
-                        <= self.space_left(),
+                    keep_inline: !call_chain_contains_options(call_expr)
+                        && self.estimate_size(call_expr.span)
+                            + if call_args.is_empty() { 4 } else { 2 }
+                            <= self.space_left(),
+                })
+                .or_else(|| {
+                    chained_named_call_cache.filter(|call| call.callee.contains(expr.span))
                 });
                 let list_format = if keep_inline {
                     ListFormat::inline()
@@ -1376,7 +1380,7 @@ impl<'ast> State<'_, 'ast> {
                             ast::ExprKind::Ident(_) | ast::ExprKind::Type(_) => (),
                             ast::ExprKind::Index(..) if s.skip_index_break => (),
                             _ if s.chained_named_call.is_some_and(|call| {
-                                call.keep_inline && call.callee == expr.span
+                                call.keep_inline && call.callee.contains(expr.span)
                             }) => {}
                             // Don't add break when accessing a field after a call with named args.
                             // e.g., `_lzSend({_dstEid: x, ...}).guid` should keep `.guid`
@@ -3055,6 +3059,16 @@ fn is_call_chain(expr_kind: &ast::ExprKind<'_>, must_have_child: bool) -> bool {
         is_call_chain(&child.kind, false)
     } else {
         !must_have_child && is_call(expr_kind)
+    }
+}
+
+fn call_chain_contains_options(expr: &ast::Expr<'_>) -> bool {
+    match &expr.peel_parens().kind {
+        ast::ExprKind::CallOptions(..) => true,
+        ast::ExprKind::Call(expr, ..)
+        | ast::ExprKind::Index(expr, ..)
+        | ast::ExprKind::Member(expr, ..) => call_chain_contains_options(expr),
+        _ => false,
     }
 }
 
--- a/crates/fmt/src/state/sol.rs
+++ b/crates/fmt/src/state/sol.rs
@@ -1314,7 +1314,9 @@ impl<'ast> State<'_, 'ast> {
                     && is_call_chain(&call_expr.kind, true))
                 .then(|| ChainedNamedCall {
                     callee: call_expr.span,
-                    keep_inline: self.estimate_size(call_expr.span) <= self.space_left(),
+                    keep_inline: self.estimate_size(call_expr.span)
+                        + if call_args.is_empty() { 4 } else { 2 }
+                        <= self.space_left(),
                 });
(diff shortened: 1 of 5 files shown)


crates/fmt/src/state/sol.rs
```diff
@@ -1303,7 +1303,24 @@ impl<'ast> State<'_, 'ast> {
...
+                let keep_inline = chained_named_call_cache
+                    .is_some_and(|call| call.keep_inline && call.callee.contains(expr.span))
+                    && !self.has_comments_between_elements(call_args.span, call_args.exprs());
                 self.call_with_opts_and_args = is_call_with_opts_and_args(&expr.kind);
+                self.chained_named_call = (matches!(call_args.kind, ast::CallArgsKind::Named(_))
+                    && is_call_chain(&call_expr.kind, true))
+                .then(|| ChainedNamedCall {
+                    callee: call_expr.span,
+                    keep_inline: self.estimate_size(call_expr.span) <= self.space_left(),
+                });
```
mattsse, line 1330: An unnamed call encountered while recursively printing the tracked callee currently clears `chained_named_call`, so intermediate links are formatted without the outer context. Preserve the cached state only while the nested expression is contained by that callee:

```suggestion
                self.chained_named_call = (matches!(call_args.kind, ast::CallArgsKind::Named(_))
                    && is_call_chain(&call_expr.kind, true))
                .then(|| ChainedNamedCall {
                    callee: call_expr.span,
                    keep_inline: self.estimate_size(call_expr.span) <= self.space_left(),
                })
                .or_else(|| {
                    chained_named_call_cache.filter(|call| call.callee.contains(expr.span))
                });
```

crates/fmt/src/state/sol.rs
```diff
@@ -1357,6 +1373,9 @@ impl<'ast> State<'_, 'ast> {
                         match member_expr.kind {
                             ast::ExprKind::Ident(_) | ast::ExprKind::Type(_) => (),
                             ast::ExprKind::Index(..) if s.skip_index_break => (),
+                            _ if s.chained_named_call.is_some_and(|call| {
+                                call.keep_inline && call.callee == expr.span
```
mattsse: Once the context is preserved, suppress the break for every member inside the tracked callee, rather than only its outermost member:

```suggestion
                                call.keep_inline && call.callee.contains(expr.span)
```

mattsse (changes requested):
The single-link fixture passes, but longer callees still reproduce the dedented named-argument block. For example, `factory().foo(bar(a)).baz({firstExtremelyLongArgumentName: firstExtremelyLongValueName, ...})` formats with the arguments aligned with `.foo` and the closing `});` dedented. The two inline suggestions below fix that reproducer together; please add a multi-link regression case. A callee containing call options (`foo{value: 1}().bar({...})`) also remains affected, so that shape should either be kept inline or excluded from `keep_inline` so the named arguments retain their own indentation.

crates/fmt/src/state/sol.rs
```diff
@@ -1303,7 +1303,30 @@ impl<'ast> State<'_, 'ast> {
...
+                    && is_call_chain(&call_expr.kind, true))
+                .then(|| ChainedNamedCall {
+                    callee: call_expr.span,
+                    keep_inline: !call_chain_contains_options(call_expr)
+                        && self.estimate_size(call_expr.span)
+                            + if call_args.is_empty() { 4 } else { 2 }
+                            <= self.space_left(),
+                })
+                .or_else(|| {
+                    chained_named_call_cache.filter(|call| call.callee.contains(expr.span))
```
figtracer, line 1329: when a nested call has an argument comment, its local `keep_inline` becomes false but this preserves the outer `keep_inline: true` state. `factory().item(/* preserve */ value).update({long: ...})` now keeps `.update` attached and dedents the named args and closing `});`; master formats this case correctly. can we invalidate the preserved context when the nested call must break and add a fixture?

crates/fmt/src/state/sol.rs
```diff
@@ -1303,7 +1303,30 @@ impl<'ast> State<'_, 'ast> {
...
                 let cache = self.call_with_opts_and_args;
+                let chained_named_call_cache = self.chained_named_call;
+                // Keep calls within a chained callee inline when they fit, so a multiline named
+                // argument list does not force an earlier break inside the callee.
+                let keep_inline = chained_named_call_cache
+                    .is_some_and(|call| call.keep_inline && call.callee.contains(expr.span))
+                    && !self.has_comments_between_elements(call_args.span, call_args.exprs());
                 self.call_with_opts_and_args = is_call_with_opts_and_args(&expr.kind);
+                self.chained_named_call = (matches!(call_args.kind, ast::CallArgsKind::Named(_))
+                    && is_call_chain(&call_expr.kind, true))
```
figtracer, line 1319: `is_call_chain` still excludes indexed and parenthesized callees. `factory().items()[i].update({...})` leaves the args and closing `});` at statement indentation, while `(factory().item()).update({...})` also dedents `.update`. can we cover these chain shapes here?

mattsse (changes requested):
The original multi-link and call-options cases are addressed, but one formatter regression remains. With `line_length = 120`, a chain containing a comment between links such as `factory() // preserve` followed by `.item().update({...});` now formats `.item().update({` at the function-body indentation (8 spaces) instead of the chain indentation (12 spaces); the base branch keeps `.item()` under the chain box. `call_chain_contains_arg_comments` only excludes comments inside argument lists, so `keep_chain_inline` still suppresses the chain box when a between-link comment forces a break. Please account for comments anywhere in the tracked callee and add this shape as a regression case.

crates/fmt/src/state/sol.rs
```diff
@@ -1303,7 +1303,31 @@ impl<'ast> State<'_, 'ast> {
...
+                .then(|| ChainedNamedCall {
+                    callee: call_expr.span,
+                    keep_inline: !call_chain_contains_options(call_expr)
+                        && !self.has_comment_between(call_expr.span.lo(), call_expr.span.hi())
+                        && self.estimate_size(call_expr.span)
```
stevencartavia: `estimate_size(call_expr.span)` derives the chained callee’s width from the unformatted source, so the callee containing `a,b,c` fits on the first pass, while the formatted `a, b, c` makes it no longer fit on the second. With `line_length = 40`, this panics with `formatter is not idempotent`. Could we make this decision independent of input token spacing and add a regression test?

crates/fmt/src/state/sol.rs
```diff
@@ -2246,7 +2279,9 @@ impl<'ast> State<'_, 'ast> {
             {
                 let current_block_multiline = self.is_multiline_block(block, false, true);
                 if !pos.is_first || !skip_ind {
-                    if prev_block_multiline && (current_block_multiline || pos.is_last) {
+                    if (pos.is_first && block.is_empty() && is_call_with_named_args(&expr.kind))
```
mattsse, line 2287: `nbsp()` removes the only break even when the catch clause does not fit. With `line_length = 80`, `try factory().item().update({firstExtremelyLongArgumentName: 1, secondExtremelyLongArgumentName: 2}) {} catch Error(string memory thisErrorReasonNameAlmostFillsTheLine) {} catch {}` emits an 81-character line, whereas the base formatter breaks before `catch` and keeps that line at 79 characters. Please compute a canonical formatted width for the empty catch and use `nbsp()` only when `1 + catch_width <= self.space_left()`; otherwise keep the existing `space()`/offset branch. Avoid measuring the raw `catch_span` unless that estimate normalizes spacing, or this can recreate the same two-pass instability as the callee-size issue.

mattsse (changes requested):
The between-link comment regression is fixed and its fixture passes, but two issues remain:

1. The existing open thread at `estimate_size(call_expr.span)` is reproducible: with `line_length = 40`, `factory().foo(a,b,c).barMethod({first: value, second: value});` panics as non-idempotent because comma normalization changes the second-pass width. Please make this decision use a canonical formatted width—either measure normalized tokens/AST (including required comma spaces) or render the callee into a scratch printer and measure that output. A conservative fallback is to disable `keep_inline` for callees whose formatting can change their width. Please add this exact shape as a regression fixture; the formatter’s built-in second pass will verify idempotence.

2. The named-call `try` special case still unconditionally removes the break before the first empty `catch`; see the inline comment for a fit-aware fix direction.

mattsse (approved):
lgtm

grandizzy (approved):
lgtm

PR #15759, opened 2026-07-15
perf: reuse inferred fork chain ID

figtracer:
## Summary

Cache the chain ID fetched while inferring a fork's network and reuse it during fork environment setup. `cast call` carries the inferred ID through its second config extraction, removing a redundant sequential `eth_chainId`; Forge tests, scripts, and Chisel reuse it instead of sending two additional chain ID lookups during fork setup. Explicit chain overrides are preserved, and Anvil network detection is unchanged.

## Benchmarks

Compared profiling builds of exact `master` (`14b8607ed`) and this branch against a deterministic local JSON-RPC endpoint with 200 ms response latency. Hyperfine used 3 warmups and 20 measured runs.

| Version | RPC sequence | Time (mean ± σ) | Range |
| --- | --- | ---: | ---: |
| `master` | `eth_chainId ×2 + eth_call` | 642.4 ms ± 8.0 ms | 624.5–656.2 ms |
| This PR | `eth_chainId + eth_call` | 434.8 ms ± 5.5 ms | 426.7–448.4 ms |

This PR is **1.48× faster** in this benchmark.

### DIFF
--- a/crates/cast/src/cmd/call.rs
+++ b/crates/cast/src/cmd/call.rs
@@ -205,7 +205,7 @@ pub enum CallSubcommands {
 }
 
 impl CallArgs {
-    pub async fn run(self) -> Result<()> {
+    pub async fn run(mut self) -> Result<()> {
         // Handle --curl mode early, before any provider interaction
         if self.rpc.curl {
             return self.run_curl().await;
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm

mablr (approved):
Makes sense

Issue #9539, opened 2024-12-11
--state seems to not work with --fork-block-number


naps62:
### Component

Anvil

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### What version of Foundry are you on?

anvil 0.2.0 (59f354c 2024-12-11T00:25:55.121360682Z)

### What command(s) is the bug in?

anvil --fork-url ... --fork-block-number ... --state

### Operating System

Linux

### Describe the bug

I've seen some recent work (https://github.com/foundry-rs/foundry/pull/9215) that fixed issues when `--state` and `--fork-url` are used together.

However, some issues still seem to persist if `--fork-block-number` is added to the mix.

to reproduce:
- `anvil --fork-url https://eth.llamarpc.com --fork-block-number 21381199 --state foo`
- in a separate terminal: `cast rpc anvil_mine`
- kill anvil, and restart with the same command. current block number will have reverted back to the original

Use case:
I have a docker-compose setup where anvil is spawned in fork mode, and with persistence enabled to make the whole setup resilient.
In some cases, I want to ensure this setup targets a specific block number of the forked chain).
The docker image runs `anvil --fork-url ... --fork-block-number ... --state /data/anvil`

When restarting the docker-compose setup, the same command will be triggered.

I suppose a quick hack to fix this is to patch the docker entrypoint so that, depending on whether the state file exists, `--state` or `--fork-block-number` is added, but not both. This seems like it shouldn't be necessary though

PR #9215, opened 2024-10-28
fix(`anvil`): set `storage.best_number` correctly

yash-atreya:
## Motivation

Closes #9053 + Closes https://github.com/foundry-rs/foundry/issues/8493

This arises when `--fork-url` and `--load-state` are used together.

## Solution

- ~Introduce `account_at` in `ForkedStorage` which stores the state of accounts at specific block numbers.~
- ~Populate `account_at` using the provided state file.~ 

Solution is a lot simpler.

- We are correctly loading the data to the db: https://github.com/foundry-rs/foundry/blob/4012adefd376bd618d1348398c1da07224d2dace/crates/anvil/src/eth/backend/mem/mod.rs#L925
- However, the `storage.best_number` i.e head of the chain is not correctly set when `--load-state` is used with `--fork-url`. We continue to set `storage.best_number = state.block.number` which is incorrect as the `best_number` should be `fork_block_number`. 
- This issue is observed when `fork_block_number` > `state.block_number`. In this case while all state data has been correctly loaded to the db, we fetch via the provider due to this check. https://github.com/foundry-rs/foundry/blob/4012adefd376bd618d1348398c1da07224d2dace/crates/anvil/src/eth/api.rs#L683
- Setting `storage.best_number = fork_block_number` fixes this

- [x] Test

PR #15760, opened 2026-07-15
fix(anvil): continue the saved timeline after loading state

### LINKED ISSUE #10331: bug(`anvil`): persistent state seems to ignore timestamp changes
### Component

Anvil

- [x] Foundry
- [x] Foundryup

forge 0.3.0 (a5c5be5 2025-01-06T00:23:40.588668993Z)

_No response_

anvil --fork-url https://mainnet.swell-rpc.com --state a --state-interval 10 --chain-id 7566690 --auto-impersonate --fork-block-number 5672511

### Operating System

None

### Describe the bug

Newer blocks after restarting anvil ignore the previous block timestamp.

Steps to reproduce:

```bash
# anvil --fork-url https://mainnet.swell-rpc.com --state a --state-interval 10 --chain-id 7566690 --auto-impersonate --fork-block-number 5672511

❯ cast age latest
Mon Apr  7 16:02:05 2025

❯ cast rpc evm_setNextBlockTimestamp $(date +%s)
null

❯ cast rpc anvil_mine
null

❯ cast age latest
Thu Apr 17 19:33:45 2025

❯ cast block

hash                 0xdf0ac0c25c0fb9887b1d4297fcc9137e08e80851ad062524b9b848ad7ffde33b
number               5672512
parentHash           0x232db3a3b42961dbada3599779554acd5c0514a03c17310b91461dca9a82457d
timestamp            1744918425 (Thu, 17 Apr 2025 19:33:45 +0000)

# (anvil restart) anvil --fork-url https://mainnet.swell-rpc.com --state a --state-interval 10 --chain-id 7566690 --auto-impersonate --fork-block-number 5672511

❯ cast block

hash                 0xdf0ac0c25c0fb9887b1d4297fcc9137e08e80851ad062524b9b848ad7ffde33b
number               5672512
parentHash           0x232db3a3b42961dbada3599779554acd5c0514a03c17310b91461dca9a82457d
timestamp            1744918425 (Thu, 17 Apr 2025 19:33:45 +0000)

❯ cast age latest
Thu Apr 17 19:33:45 2025

❯ cast rpc anvil_mine
null

❯ cast age latest
Mon Apr  7 16:02:34 2025

❯ cast block

hash                 0x211e24b15071c2e1a36d26287e05ed890fdf0138e8b99d1fd5140a5bbb39ae43
number               5672513
parentHash           0xdf0ac0c25c0fb9887b1d4297fcc9137e08e80851ad062524b9b848ad7ffde33b
timestamp            1744041754 (Mon, 7 Apr 2025 16:02:34 +0000)

```

0xMars42:
## Motivation

Fixes #10331. `Backend::load_state` restores the block env, base fee and blob params from the saved head, but never re-anchors the `TimeManager`. The first block mined after loading a state file therefore takes its timestamp from the node's startup anchor (genesis or fork block) instead of the saved chain head: the loaded chain sits on its saved timeline while the next block jumps back to wall-clock time, which can produce a non-monotonic chain.

## Solution

Re-anchor block time to the canonical head `load_state` selects. When the loaded head stays canonical (plain load, or a state file ahead of the fork block, see #9539), the anchor is its header timestamp read from storage. When the state file is at or below the fork block (see #9215), the canonical head rolls back to the fork block, whose header is not in local storage, so the anchor takes the fork timestamp, the same source `reset_fork` uses. Resolving the head by identity also keeps the timeline of stale blocks out of the anchor: a state file can carry blocks above its own best block (`load_blocks` never prunes, so dumping after loading an older state keeps the discarded blocks), and the highest loaded header is then not the canonical head.

This also applies to the `anvil_loadState` RPC, which goes through the same path: loading a state at runtime re-anchors block time the same way, consistent with `--load-state` at startup.

Two deliberate semantic notes: a pending `evm_setNextBlockTimestamp` and a previously applied `evm_increaseTime` offset do not survive a load, matching the other re-anchor paths (`anvil_reset`, snapshot revert); and `--timestamp` combined with `--load-state` now resolves to the loaded head's timeline (previously the flag survived the load, which could produce a non-monotonic chain against the loaded head).

Regression tests: the original one (chain moved a year ahead of wall-clock, saved, reloaded on a fresh node: the next block continues the saved timeline; fails on master with `block after load_state went back in time: 1784120095 < 1815656095`), the equal-height fork case (fork anchor kept), the fork rollback (future-dated local block, load a height-0 state onto the fork: `latest` returns the exact fork head and the next block leaves the discarded timeline), and a dump carrying stale blocks above its best block (the next block continues the canonical timeline). The last two fail on the previous revision of this branch.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -4758,6 +4758,13 @@ impl<N: Network<ReceiptEnvelope = FoundryReceiptEnvelope>> Backend<N> {
                     header.timestamp,
                 ),
             ));
+
+            // Re-anchor block time to the loaded head so subsequent blocks continue the saved
+            // timeline instead of the node's startup anchor (genesis or fork block). Skip when
+            // the canonical head is not the loaded one (state file older than the fork block).
+            if header.number == self.blockchain.storage.read().best_number {
```
mattsse: When the loaded state head and fork head have the same block number, the selection above keeps the fork head (`best_number > number` is false), but this height-only check still treats the loaded header as canonical and resets `TimeManager` from its timestamp. I reproduced this with a future-dated local block at height N loaded onto a fork also at N; the next block followed the stale state timeline. Comparing the actual canonical hash fixes it:

```suggestion
            if header.hash_slow() == self.blockchain.storage.read().best_hash {
```

Please also add an equal-height fork regression test.

mattsse (changes requested):
The equal-height fork case can still replace the canonical fork time anchor with the loaded state's timestamp. Please use canonical head identity for this guard and add coverage for a loaded state and fork head at the same height.

0xMars42 (commented):
Good catch, the equal-height case did fire the reset: reproduced with a test forking a node at the same height as the loaded state's head, which failed on the previous guard and passes now. The guard compares the loaded head's hash with the canonical best hash instead of the numbers.

mattsse (dismissed):
The follow-up now guards the time re-anchor by canonical head identity and adds the requested equal-height fork regression. The focused regression test passes locally.

mablr (dismissed):

mablr (dismissed):

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5588,6 +5588,14 @@ impl<N: Network<ReceiptEnvelope = FoundryReceiptEnvelope>> Backend<N> {
                     header.timestamp,
                 ),
             ));
+
+            // Re-anchor block time to the loaded head so subsequent blocks continue the saved
+            // timeline instead of the node's startup anchor (genesis or fork block). Skip when
+            // the canonical head is not the loaded one (state file at or below the fork block),
+            // comparing by identity since a fork head can share the loaded head's height.
+            if header.hash_slow() == self.blockchain.storage.read().best_hash {
```
stevencartavia: When `anvil_loadState` rolls an already-advanced fork back to its fork head using state older than that head, `best_hash` becomes the fork hash while `header` is the loaded state’s latest header. Their hashes differ, so this guard leaves `TimeManager` on the discarded local timeline. This is reproducible by mining a future-dated local block, loading height-0 state, and confirming that `latest` returns to the exact fork head while the next block reuses the discarded block’s future timestamp.

Could we re-anchor `TimeManager` from the selected fork head’s timestamp when the loaded header does not match it, and add a runtime regression test for this case?

0xMars42: Reproduced, thanks. And the guard actually hides a second case: a state file can carry stale blocks above its own best block (dump at height N, keep mining, load that dump back: `load_blocks` never prunes, so the next dump contains the discarded blocks). Loading such a file skipped the re-anchor too, for the same reason: the highest loaded header is not the canonical head.

So instead of special-casing the fork head I replaced the guard with the invariant it was approximating: after `load_state` selects the canonical head, block time always re-anchors to that head. When the loaded head stays canonical, the anchor is its header timestamp read from storage. When the state file is at or below the fork block, the head rolls back to the fork block, whose header is not in local storage, so the anchor takes the fork timestamp, the same source `reset_fork` uses.

Added the runtime regression test you asked for (future-dated local block, load a height-0 state, assert `latest` returns the exact fork head and the next block leaves the discarded timeline) plus one for the stale-blocks dump. Both fail on the previous revision of this branch.

Two behavior notes to be explicit about: re-anchoring is now active, so a load that keeps the fork head canonical no longer preserves a previously applied `evm_increaseTime` offset, and a pending `evm_setNextBlockTimestamp` does not survive a load; both match the other re-anchor paths (`anvil_reset`, snapshot revert). And `BlockEnv.timestamp` still comes from the loaded file until the next block is mined, on master as well; aligning it with the selected head felt like a separate change, happy to fold it in here if preferred.

figtracer (dismissed):

mablr (approved):
LGTM
Thank you @0xMars42 !

figtracer (approved):

PR #15761, opened 2026-07-15
feat(lint): detect inconsistent type names

mattsse:
Adds the low-severity `inconsistent-type-names` detector from #14381. It reports shorthand `uint` or `int` declarations when the same directly declared contract also uses `uint256` or `int256`, including nested array and mapping types, while keeping inheritance and other contracts in separate consistency scopes. The implementation uses resolved Solar HIR ownership and exact typed source spans so normalized integer types do not turn into name-based guesses; focused UI coverage, a blessed snapshot, and the lint reference page are included.

The companion documentation is in [foundry-rs/book#1974](https://github.com/foundry-rs/book/pull/1974).

This PR was implemented with Codex assistance across code, tests, and documentation, then independently reviewed by a fresh Codex context. PR responses may also be drafted with Codex assistance.


mablr (approved):
LGTM

figtracer (approved):

PR #15762, opened 2026-07-15
feat(lint): detect selector collisions

mattsse:
This adds the high-severity `function-selector-collision` lint from #14381. It identifies proxy and implementation pairs when a concrete fallback forwards full calldata through Solar's resolved `address.delegatecall` builtin on an address converted from a statically typed contract or interface value, then compares the effective external ABIs for different signatures sharing a four-byte selector.

Forwarding is tracked per reachable path at the call site. The collector evaluates callee, call options, and arguments before recording a target; follows direct `msg.sig ==` and `!= Contract.function.selector` constraints; and joins input and selector state across branches, try clauses, while/for/do-while loops, break/continue exits, short-circuit expressions, and ternaries. Loop states are canonicalized and conservatively widened at a fixed cap. Unrelated contracts, erased address targets, assembly proxies, helper-hidden delegatecalls, and user-defined same-name methods remain outside scope. Fresh-context reviews drove permanent positive and negative coverage for every corrected path.

The user documentation is in foundry-rs/book#1980, with the review-aligned forwarding-path follow-up in foundry-rs/book#1997.


crates/lint/src/sol/high/function_selector_collision.rs
```diff
@@ -0,0 +1,246 @@
...
+            return;
+        }
+        let Some(fallback_id) = proxy.fallback else { return };
+        let fallback = hir.function(fallback_id);
+        let Some(body) = fallback.body else { return };
+
+        let mut collector = DelegateTargetCollector {
+            gcx,
+            hir,
+            fallback_input: unmodified_fallback_input(hir, fallback, body.stmts),
```
mablr: This whole-body mutation check creates a false negative when the fallback input is reassigned only after it has already been passed to `delegatecall`. For example, `delegatecall(input); input = input[4:];` forwards the full calldata, but the lint emits nothing. Please track whether the argument is unmodified at the call site rather than whether it is ever modified anywhere in the fallback. I reproduced this with the UI fixture.

mablr (changes requested):

crates/lint/src/sol/high/function_selector_collision.rs
```diff
@@ -0,0 +1,246 @@
...
+                continue;
+            }
+            let implementation = hir.contract(implementation_id);
+            if implementation.kind == ContractKind::Library || implementation.linearization_failed()
+            {
+                continue;
+            }
+
+            for proxy_function in proxy_functions.all() {
+                for implementation_function in gcx.interface_functions(implementation_id).all() {
```
figtracer, line 84: this treats a typed target as the full fallback api even when the delegatecall is selector-gated. i reproduced `if (msg.sig == IImplementation.forwarded.selector) delegatecall(msg.data)` still reporting an unrelated `gsf`/`tgeo` collision; please account for the guard or omit unreachable pairs.

crates/lint/src/sol/high/function_selector_collision.rs
mattsse: The collector now evaluates the call receiver, options, and arguments before recording the delegatecall, and keeps fallback-input mutation state per reachable path at that exact call site. Mutations after a forwarding call no longer suppress it; mutations before or within evaluated call children still do. The fixture covers both orders plus branches, try clauses, loops, short-circuit, and ternary paths.

crates/lint/src/sol/high/function_selector_collision.rs
mattsse: Delegate targets now retain per-path msg.sig == and != Contract.function.selector constraints, including compound conditions and loop paths. Collision comparison is restricted to selectors allowed by at least one path reaching that delegatecall, so the forwarded-only gate no longer reports the unrelated gsf/tgeo pair. Exact positive and negative regressions are included.

crates/lint/src/sol/high/function_selector_collision.rs
```diff
@@ -0,0 +1,679 @@
...
+        gcx: Gcx<'hir>,
+        hir: &'hir hir::Hir<'hir>,
+        proxy_id: ContractId,
+    ) {
+        let proxy = hir.contract(proxy_id);
+        if proxy.kind != ContractKind::Contract || proxy.linearization_failed() {
+            return;
+        }
+        let Some(fallback_id) = proxy.fallback else { return };
+        let fallback = hir.function(fallback_id);
```
mablr, line 50: Fallback modifiers are not included in this traversal, which causes both false positives and false negatives. I reproduced a false positive with `modifier onlyForwarded { if (msg.sig == IImplementation.forwarded.selector) _; }` on a fallback whose body delegates `msg.data`: the unrelated `tgeo`/`gsf` collision is still reported. Conversely, moving the typed `delegatecall(msg.data)` into a modifier and leaving the fallback body empty emits nothing for the reachable collision. Please account for the fallback's applied modifier control flow and delegatecalls (or conservatively avoid diagnosing modified fallbacks).

mattsse: Fixed in f96754dd6. Fallback analysis now expands applied modifier chains in execution order, binds full-calldata provenance through modifier parameters, resumes outer suffixes after inner returns, scopes assembly invalidation to active lexical calldata values, and caches continuation states to bound repeated placeholders. Direct regressions cover each case; full UI/clippy/fmt validation and final fresh-context re-review pass.

crates/lint/src/sol/high/function_selector_collision.rs
```diff
@@ -0,0 +1,898 @@
...
+            && matches!(op.kind, BinOpKind::And | BinOpKind::Or)
+        {
+            let (mut true_paths, false_paths) = self.visit_condition(expr);
+            Self::extend_unique(&mut true_paths, false_paths);
+            self.paths = true_paths;
+            self.dedup_paths();
+            return ControlFlow::Continue(());
+        }
+
+        if let ExprKind::Call(callee, args, opts) = &expr.kind {
```
mablr, line 643: The successful continuation of `require(msg.sig == IImplementation.forwarded.selector)` does not inherit the selector constraint. A fallback gated this way still reports collisions with unrelated implementation functions that cannot reach the delegatecall. Please apply direct selector guards from `require`/`assert` to the continuing path.

mattsse: Fixed in c1d9f8873. Successful require/assert paths now inherit the selector constraint; remaining builtin arguments are evaluated on both outcomes, with only successful descendants continuing. Added require/assert suppression regressions plus a reachable-collision control. Full UI/clippy/fmt validation and final fresh-context re-review pass.

mablr (approved):
LGTM

stevencartavia (approved):

wyattloveslife-create (commented):
Thank you

wyattloveslife-create (commented):
Thank you

PR #15764, opened 2026-07-15
fix(forge): respect Amsterdam code size limits

figtracer:
## Summary

- derive default contract size-report limits from the active EVM specification
- use the EIP-170/EIP-3860 limits before Amsterdam and the finalized EIP-7954 64 KiB runtime / 128 KiB initcode limits for Amsterdam
- apply the same spec-aware fallback to script deployment warnings while preserving explicit `code_size_limit` overrides

## Motivation

Selecting `evm_version = "amsterdam"` still made `forge build --sizes` and script checks use the pre-Amsterdam 24 KiB runtime / 48 KiB initcode limits. Setting `code_size_limit` works around the reporting issue, but it is an explicit EVM override rather than the protocol default implied by the selected EVM version.

With this change, selecting Amsterdam automatically uses its finalized 64 KiB runtime and 128 KiB initcode limits. Explicit `code_size_limit` values continue to override the default for build and script checks.

The workspace currently uses revm 41, whose EIP-7954 constants still contain the earlier 32/64 KiB draft values, so the finalized limits are defined from the EIP directly. The regression tests cover a 50 KiB runtime contract and the finalized size-report margins.

Addresses https://github.com/foundry-rs/foundry/pull/14761#issuecomment-4981425557

### DIFF
--- a/crates/script/src/lib.rs
+++ b/crates/script/src/lib.rs
@@ -477,7 +477,11 @@ impl ScriptArgs {
                 .code_size_limit
                 .or(pre_simulation.script_config.config.code_size_limit)
                 .map(ContractSizeLimits::with_runtime_limit)
-                .unwrap_or_default();
+                .unwrap_or_else(|| {
+                    ContractSizeLimits::for_spec_id(
+                        pre_simulation.script_config.config.evm_spec_id(),
+                    )
+                });
             pre_simulation.args.check_contract_sizes(
                 size_limits,
                 &pre_simulation.execution_result,
@@ -592,8 +596,8 @@ impl ScriptArgs {
         Ok((func.clone(), data.into()))
     }
 
-    /// Checks if the transaction is a deployment with either a size above the default contract size
-    /// limit or specified `code_size_limit`.
+    /// Checks if the transaction is a deployment with a size above the active EVM specification's
+    /// contract size limit or the specified `code_size_limit`.
     ///
     /// If `self.broadcast` is enabled, it asks confirmation of the user. Otherwise, it just warns
     /// the user.
--- a/crates/script/src/lib.rs
+++ b/crates/script/src/lib.rs
@@ -52,7 +52,7 @@ use foundry_evm::{
     backend::Backend,
     core::{
         Breakpoints, FoundryTransaction,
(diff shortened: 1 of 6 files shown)


mablr (dismissed):
Makes sense

sakulstra:
> With this change, selecting Amsterdam automatically uses its 32 KiB runtime and 64 KiB initcode limits, so users do not need to change the test executor limit just to get accurate size checks.

Should be `Raise the maximum contract code size from 24KiB to 64KiB and initcode size from 48KiB to 128KiB.`, no?
https://forkcast.org/eips/7954/

Thanks for patching this <3

mablr:
@sakulstra Yes, you’re right. But the revm 41.0.0 constants used here still reflect the earlier 32/64 KiB draft values; so we'll need to bump revm.

mablr:
Marking as "blocked" until next revm bump.

sakulstra:
Should `code_size_limit` be ignored on tests/scripts anyways though?

mablr:
@figtracer could you recheck please?

mablr:
LGTM. We'll get revm 42 in Foundry next week, let's wait for it ?

mablr (dismissed):

crates/script/src/lib.rs
```diff
@@ -477,7 +477,11 @@ impl ScriptArgs {
                 .code_size_limit
                 .or(pre_simulation.script_config.config.code_size_limit)
                 .map(ContractSizeLimits::with_runtime_limit)
-                .unwrap_or_default();
+                .unwrap_or_else(|| {
+                    ContractSizeLimits::for_spec_id(
+                        pre_simulation.script_config.config.evm_spec_id(),
```
stevencartavia: This constrains `config.evm_spec_id()` to Ethereum `SpecId` instead of the `SpecFor<FEN>` used by the executor. Reproduced with both non-Ethereum networks: inferred Amsterdam emits no warning, while the equivalent explicit `optimism:karst` / `tempo:T8` hardfork warns at 24,576 bytes despite using the same effective execution spec. Could we resolve `SpecFor<FEN>` first, then convert it to `SpecId` for the fallback limits?

mablr (approved):

DaniPopes (approved):

PR #15765, opened 2026-07-15
feat(lint): detect unsafe ecrecover

mattsse:
This adds the medium-severity `ecrecover` lint from #14381. It reports direct uses of Solidity's builtin unless the signature's `s` value is proven to be in the canonical lower half of the secp256k1 order, with path-sensitive handling for guards, aliases, branches, loops, assignments, mutable calls, and Solidity constant expressions.

A locally stored recovery is deferred until its first observable use, so a dominating post-call low-s guard can validate the exact signature value before the signer affects behavior. Pending recoveries survive joins when any path remains unguarded, while pre-guard reads, state writes, reassigned signature values, and implicit named-return uses still report. The UI fixture also covers comparison polarity, control-flow exits, tuple and ternary assignments, loop-carried writes, mutable-state invalidation, builtin resolution, and safe counterexamples.

The companion Book documentation is in foundry-rs/book#1981, with the post-call guard follow-up in foundry-rs/book#1994. This PR was implemented with Codex assistance across code, tests, and lint documentation, then independently reviewed and corrected by a fresh Codex context.

### DIFF
--- a/crates/lint/docs/ecrecover.md
+++ b/crates/lint/docs/ecrecover.md
@@ -18,10 +18,10 @@ retained only when they hold on every path reaching the call.
 Checks on `v`, the recovered address, nonces, domain separators, or the top bit of an EIP-2098
 signature do not prove that `s` is canonical and do not suppress the warning.
 
-The analysis is intentionally local and requires the low-`s` guard to be established before the
-call. It does not summarize internal helpers or modifiers, prove post-call guards, recognize bound
-signature schemes that fix one recovery ID, or inspect Yul and low-level calls to precompile
-address `0x01`.
+The analysis is intentionally local. A low-`s` guard may be established before the call or after
+assigning the recovered address to a local, provided it dominates the address's first observable
+use. The analysis does not summarize internal helpers or modifiers, recognize bound signature
+schemes that fix one recovery ID, or inspect Yul and low-level calls to precompile address `0x01`.
 
 ## Why is this bad?
 
--- a/crates/lint/docs/ecrecover.md
+++ b/crates/lint/docs/ecrecover.md
@@ -12,7 +12,8 @@ The lint follows simple local aliases and recognizes low-`s` bounds established 
 `assert`, conditionals, and early-return or revert branches. It accepts the canonical maximum
 `0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF5D576E7357A4501DDFE92F46681B20A0`, stricter bounds, reversed
 comparisons, and the equivalent strict comparison against that value plus one. Facts are
-invalidated by reassignment and retained only when they hold on every path reaching the call.
+invalidated by reassignment, loop-carried writes, and calls that may change mutable state. They are
+retained only when they hold on every path reaching the call.
 
 Checks on `v`, the recovered address, nonces, domain separators, or the top bit of an EIP-2098
 signature do not prove that `s` is canonical and do not suppress the warning.
--- a/crates/lint/docs/ecrecover.md
+++ b/crates/lint/docs/ecrecover.md
@@ -1,0 +1,53 @@
+# Unsafe ecrecover
+
+**Severity**: `Med`
+**ID**: `ecrecover`
+
+Flags direct calls to Solidity's `ecrecover` builtin when the signature's `s` value is not proven
+to be in the canonical lower half of the secp256k1 curve order.
+
+## What it does
+
+The lint follows simple local aliases and recognizes low-`s` bounds established by `require`,
+`assert`, conditionals, and early-return or revert branches. It accepts the canonical maximum
+`0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF5D576E7357A4501DDFE92F46681B20A0`, stricter bounds, reversed
+comparisons, and the equivalent strict comparison against that value plus one. Facts are
+invalidated by reassignment and retained only when they hold on every path reaching the call.
+
+Checks on `v`, the recovered address, nonces, domain separators, or the top bit of an EIP-2098
+signature do not prove that `s` is canonical and do not suppress the warning.
+
+The analysis is intentionally local and requires the low-`s` guard to be established before the
+call. It does not summarize internal helpers or modifiers, prove post-call guards, recognize bound
+signature schemes that fix one recovery ID, or inspect Yul and low-level calls to precompile
+address `0x01`.
+
+## Why is this bad?
+
+For each valid high-`s` ECDSA signature, an attacker can derive a second signature for the same
+message and signer by replacing `s` with its complement in the curve order and flipping `v`.
+Contracts that use the signature bytes as a unique identifier can therefore have replay or
+double-use protections bypassed. The `ecrecover` precompile does not enforce the low-`s` rule that
+Ethereum transactions enforce.
+
+## Example
+
+### Bad
+
+```solidity
+function recover(bytes32 hash, uint8 v, bytes32 r, bytes32 s) pure returns (address) {
+    return ecrecover(hash, v, r, s);
+}
+```
+
+### Good
+
+```solidity
+uint256 constant HALF_ORDER =
+    0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF5D576E7357A4501DDFE92F46681B20A0;
+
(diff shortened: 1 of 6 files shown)


mablr (commented):
LGTM

mablr (dismissed):

crates/lint/testdata/Ecrecover.sol
```diff
@@ -0,0 +1,357 @@
...
+        return ecrecover(hash, v, r, s); //~WARN: ecrecover should reject malleable signatures
+    }
+
+    function topBitMask(bytes32 hash, uint8 v, bytes32 r, bytes32 vs) external pure returns (address) {
+        bytes32 s = bytes32(uint256(vs) & TOP_BIT_MASK);
+        return ecrecover(hash, v, r, s); //~WARN: ecrecover should reject malleable signatures
+    }
+
+    function guardAfterCall(bytes32 hash, uint8 v, bytes32 r, bytes32 s) external pure returns (address) {
+        address signer = ecrecover(hash, v, r, s); //~WARN: ecrecover should reject malleable signatures
```
figtracer, line 382: this case is safe: `ecrecover` is pure and `signer` is not used before the low-s `require`, so the revert validates it before any observable use. please keep post-call guards clean or omit the detector until use tracking exists.

mattsse: Addressed in 022ebd4c1. Locally stored recoveries are now deferred until their first observable use, and a dominating low-s fact for the exact signature value clears the pending diagnostic first. Branch joins retain any unguarded path; pre-guard reads, state writes, named-return fallthrough, and reassigned signature values remain covered by regressions.

crates/lint/src/sol/med/ecrecover.rs
```diff
@@ -0,0 +1,1246 @@
...
+            StmtKind::Block(block) | StmtKind::UncheckedBlock(block) => {
+                for stmt in block.stmts {
+                    let _ = self.visit_stmt(stmt);
+                    if branch_always_exits(stmt) {
+                        break;
+                    }
+                }
+                return ControlFlow::Continue(());
+            }
+            StmtKind::If(condition, then_stmt, else_stmt) => {
```
mablr, line 1008: Constant conditions are analyzed on both paths, which produces a false positive for unreachable calls. I verified that `if (false) return ecrecover(h, v, r, s);` is warned. Please use `const_bool(condition)` here (as the loop-effects analysis already does) to skip the impossible branch.

crates/lint/docs/ecrecover.md
```diff
@@ -0,0 +1,54 @@
...
+`0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF5D576E7357A4501DDFE92F46681B20A0`, stricter bounds, reversed
+comparisons, and the equivalent strict comparison against that value plus one. Facts are
+invalidated by reassignment, loop-carried writes, and calls that may change mutable state. They are
+retained only when they hold on every path reaching the call.
+
+Checks on `v`, the recovered address, nonces, domain separators, or the top bit of an EIP-2098
+signature do not prove that `s` is canonical and do not suppress the warning.
+
+The analysis is intentionally local and requires the low-`s` guard to be established before the
+call. It does not summarize internal helpers or modifiers, prove post-call guards, recognize bound
```
mablr: This is stale after 022ebd4c1: the implementation and fixture now explicitly support a dominating post-call guard before the recovered value’s first observable use. Please document that behavior rather than saying post-call guards are not proven.

crates/lint/src/sol/med/ecrecover.rs
mattsse: Fixed in `48f375ce5`. Constant `if` conditions now follow only the feasible arm in ordinary and loop traversal, and constant-selected exits stop enclosing traversal. Added regressions for the exact `if (false) return ecrecover(...)` case, an unreachable call after `if (true) return`, and a loop-nested constant branch. Focused/full UI, `forge-lint` clippy with warnings denied, nightly fmt, and diff checks pass; a fresh-context re-review found no remaining issues.

crates/lint/docs/ecrecover.md
mattsse: Updated in `48f375ce5`. The docs now state that a low-`s` guard may occur after assigning the recovered address when it dominates that local’s first observable use, matching the current implementation and fixture.

crates/lint/src/sol/med/ecrecover.rs
```diff
@@ -0,0 +1,1284 @@
...
+        self.walk_stmt(stmt)
+    }
+
+    fn visit_expr(&mut self, expr: &'hir hir::Expr<'hir>) -> ControlFlow<Self::BreakValue> {
+        if matches!(expr.kind, ExprKind::Ident(_))
+            && let Some(value) = self.current_value(expr)
+        {
+            self.use_value(value);
+        }
+        if let ExprKind::Binary(lhs, op, rhs) = &expr.kind
```
mablr, line 1148: Constant feasibility is not applied to expression arms. Both `false && ecrecover(...) != address(0)` and `false ? ecrecover(...) : address(0)` warn even though the call cannot execute; an equivalent low-s guard such as `require(uint256(s) <= HALF_ORDER || false)` is also rejected. Please apply `const_bool` before visiting and joining these arms.

crates/lint/src/sol/med/ecrecover.rs
```diff
@@ -0,0 +1,1284 @@
...
+                self.restore(joined);
+                return ControlFlow::Continue(());
+            }
+            StmtKind::DeclSingle(var) => {
+                let init = self.hir.variable(*var).initializer;
+                if let Some(init) = init {
+                    self.visit_stored_expr(init, true);
+                }
+                self.assign_var(*var, init);
+                if let Some(recovery) = init.and_then(|init| self.pending_recovery(init)) {
```
mablr: Deferred recovery provenance is limited to direct initializers/assignments. `(signer,) = (ecrecover(...), 0); require(uint256(s) <= HALF_ORDER);` and `address signer = useRecovery ? ecrecover(...) : address(0);` followed by the same guard both warn before any observable use. Please retain pending recovery through composite local storage.

crates/lint/src/sol/med/ecrecover.rs
mattsse: Fixed in cc8619fea. Constant feasibility now prunes unreachable short-circuit and ternary arms during call traversal, loop effects, low-s fact extraction, side-effect checks, and ternary proof. Added safe/unsafe controls, including unreachable-arm side effects. Full validation and final fresh-context re-review pass.

crates/lint/src/sol/med/ecrecover.rs
mattsse: Fixed in cc8619fea. Deferred recovery provenance now survives tuple destructuring, ternary storage, and nested local assignments, while nested observable/state writes warn immediately. Added direct tuple, nested-local, and mixed local/state regressions. Full validation and final fresh-context re-review pass.

mablr (dismissed):
LGTM

crates/lint/src/sol/med/ecrecover.rs
```diff
@@ -0,0 +1,1445 @@
...
+                    self.record_pending(var, recovery);
+                }
+                return ControlFlow::Continue(());
+            }
+            StmtKind::Expr(expr) => {
+                self.visit_discarded_expr(expr);
+                return ControlFlow::Continue(());
+            }
+            StmtKind::Err(_) | StmtKind::AssemblyBlock(_) => {
+                self.state = FlowState::default();
```
stevencartavia, line 1132: Resetting `FlowState` here also drops deferred `pending` recoveries, allowing an assembly block to suppress an expected `ecrecover` warning. For example, `address signer = ecrecover(...); assembly { pop(0) } return signer;` currently produces no warning. I reproduced this with unrelated, empty, and memory-safe assembly blocks; moving the same assembly before `ecrecover` warns correctly. One option would be to treat `AssemblyBlock` as an opaque boundary and preserve or conservatively report unresolved pending recoveries instead of clearing them, while keeping overwritten recoveries clean.

mablr (approved):
LGTM

figtracer (approved):

PR #15766, opened 2026-07-15
feat(lint): add protected-vars

mattsse:
This adds the high-severity `protected-vars` lint from #14381. It enforces Solidity's `@custom:security write-protection` annotations by tracing writes from externally callable leaf-contract entry points through reachable modifier continuations, resolved internal and library calls, storage aliases, returned storage locations, collection mutations, and resolvable assembly slots.

Protection is tracked as must-flow state at each write, so a guard must dominate that occurrence across branches, short-circuit expressions, loops, returns, and modifier expansion. The analysis also retargets Yul storage-pointer `.slot` assignments, distinguishes persistent `sstore` from transient `tstore`, and uses resolved Yul return summaries without contaminating results with unrelated argument roots. Exact overload signatures and virtual dispatch remain respected, while malformed or unresolved annotations stay unsatisfied.

The companion Book documentation is in foundry-rs/book#1982, with the dominance and Yul follow-up in foundry-rs/book#1995. This PR was implemented with Codex assistance across code, tests, and lint documentation, then independently reviewed and corrected by a fresh Codex context.

### DIFF
--- a/crates/lint/src/sol/high/protected_vars.rs
+++ b/crates/lint/src/sol/high/protected_vars.rs
@@ -1,0 +1,588 @@
+use super::ProtectedVars;
+use crate::{
+    linter::{LateLintPass, LintContext},
+    sol::{Severity, SolLint},
+};
+use solar::{
+    ast::{ContractKind, DataLocation, ElementaryType, FunctionKind, Visibility},
+    interface::sym,
+    sema::{
+        Gcx,
+        hir::{
+            self, ContractId, ExprKind, FunctionId, ItemId, NatSpecKind, Res, StmtKind, VariableId,
+        },
+        ty::TyKind,
+    },
+};
+use std::collections::{HashMap, HashSet};
+
+declare_forge_lint!(
+    PROTECTED_VARS,
+    Severity::High,
+    "protected-vars",
+    "protected variable is written without its required protection"
+);
+
+impl<'hir> LateLintPass<'hir> for ProtectedVars {
+    fn check_nested_contract(
+        &mut self,
+        ctx: &LintContext,
+        gcx: Gcx<'hir>,
+        hir: &'hir hir::Hir<'hir>,
+        contract_id: ContractId,
+    ) {
+        let contract = hir.contract(contract_id);
+        if !matches!(contract.kind, ContractKind::Contract | ContractKind::AbstractContract)
+            || contract.linearization_failed()
+            || !is_most_derived_contract(hir, contract_id)
+        {
+            return;
+        }
+
+        let protected = protected_variables(gcx, hir, contract.linearized_bases);
+        if protected.is_empty() {
+            return;
+        }
+
+        let targets = ProtectionTargets::new(gcx, hir, contract.linearized_bases);
+        for entry_id in effective_entry_points(gcx, hir, contract.linearized_bases) {
+            let mut analyzer = EntryAnalyzer::new(gcx, hir, contract.linearized_bases);
+            let (writes, calls) = analyzer.analyze(entry_id);
+
+            for var_id in writes {
+                let Some(requirements) = protected.get(&var_id) else { continue };
+                for requirement in requirements {
+                    let Some(target_id) = targets.resolve(requirement) else { continue };
+                    if calls.contains(&target_id) {
+                        continue;
+                    }
+
+                    let entry = hir.function(entry_id);
+                    let span = entry.name.map_or(entry.keyword_span(), |name| name.span);
+                    let variable = hir
+                        .variable(var_id)
+                        .name
+                        .map_or_else(|| "<unnamed>".to_string(), |name| name.as_str().to_string());
+                    ctx.emit_with_msg(
+                        &PROTECTED_VARS,
+                        span,
+                        format!(
+                            "protected variable `{variable}` is written without `{requirement}`"
+                        ),
+                    );
+                }
+            }
+        }
+    }
+}
+
+/// Slither analyzes the effective entry points of leaf contracts so inherited declarations are
+/// interpreted in the context in which they are ultimately deployed.
+fn is_most_derived_contract(hir: &hir::Hir<'_>, contract_id: ContractId) -> bool {
+    !hir.contract_ids().any(|candidate_id| {
+        candidate_id != contract_id
+            && hir
+                .contract(candidate_id)
+                .linearized_bases
+                .get(1..)
+                .is_some_and(|bases| bases.contains(&contract_id))
+    })
+}
+
+fn protected_variables(
+    gcx: Gcx<'_>,
+    hir: &hir::Hir<'_>,
+    bases: &[ContractId],
+) -> HashMap<VariableId, Vec<String>> {
+    let mut protected = HashMap::new();
+
+    for &contract_id in bases {
+        for var_id in hir.contract(contract_id).variables() {
+            let var = hir.variable(var_id);
+            if !var.kind.is_state() {
+                continue;
+            }
+
+            let requirements: Vec<_> = gcx
+                .natspec_doc_comments(var.doc)
+                .iter()
+                .filter_map(|item| {
+                    let NatSpecKind::Custom { name } = item.kind else { return None };
+                    if name.as_str() != "security" {
+                        return None;
+                    }
+                    parse_write_protection(item.content()).map(str::to_owned)
+                })
+                .collect();
+            if !requirements.is_empty() {
+                protected.insert(var_id, requirements);
+            }
+        }
+    }
+
+    protected
+}
+
+fn parse_write_protection(content: &str) -> Option<&str> {
+    let value = content.split_once("write-protection=\"")?.1;
+    let (signature, _) = value.split_once('"')?;
+    (!signature.is_empty()).then_some(signature)
+}
+
+struct ProtectionTargets {
+    functions: HashMap<String, FunctionId>,
+    modifiers: HashMap<String, FunctionId>,
+}
+
+impl ProtectionTargets {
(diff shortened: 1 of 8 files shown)


crates/lint/src/sol/high/protected_vars.rs
```diff
@@ -0,0 +1,1239 @@
...
+
+    fn analyze_function(&mut self, function_id: FunctionId) -> Vec<StorageRoots> {
+        let function = self.hir.function(function_id);
+        let Some(body) = function.body else {
+            return function.returns.iter().map(|_| StorageRoots::new()).collect();
+        };
+        self.stack.push(function_id);
+        self.return_stack.push(function.returns.iter().map(|_| StorageRoots::new()).collect());
+        self.analyze_modifiers(function);
+        self.analyze_block(body);
```
mablr: `analyze_modifiers` analyzes each modifier body and then this unconditionally analyzes the function body, even when the effective modifier never reaches `_`. For example, `modifier blocked() { revert(); _; }` on an external function whose body writes a protected variable will be reported although that write is unreachable.

mablr (changes requested):

crates/lint/src/sol/high/protected_vars.rs
```diff
@@ -0,0 +1,1239 @@
...
+                    };
+                    let variable = hir
+                        .variable(var_id)
+                        .name
+                        .map_or_else(|| "<unnamed>".to_string(), |name| name.as_str().to_string());
+                    match requirement {
+                        ProtectionRequirement::Signature(signature) => {
+                            if targets
+                                .resolve(signature)
+                                .is_some_and(|target_id| calls.contains(&target_id))
```
figtracer: this makes the high-severity lint accept a guard on only one branch or after the write; both unsafe cases are currently blessed in `ReachabilitySemantics`. please require the protection to dominate each write, or omit the detector until that analysis exists.

crates/lint/src/sol/high/protected_vars.rs
```diff
@@ -0,0 +1,1239 @@
...
+
+        if !compound && let Some(local) = lhs_local_var(self.hir, lhs) {
+            self.set_storage_alias(local, rhs);
+            if self.assembly_depth > 0 {
+                self.set_slot_alias(local, rhs);
+            }
+            return;
+        }
+
+        self.record_write(lhs);
```
stevencartavia, line 1165: `pointer.slot := rhs.slot` reaches this path because the LHS is a `YulMember`, so it records the pointer's old root as written and never updates the alias. This causes a false positive when retargeting away from protected storage and a false negative for `pointer.slot := protectedArray.slot` followed by `pointer.push(1)`. Could local storage-pointer `.slot` assignments retarget the base local's storage alias instead?

crates/lint/src/sol/high/protected_vars.rs
```diff
@@ -0,0 +1,1239 @@
...
+        merged.entry(variable_id).or_default().extend(roots);
+    }
+    merged
+}
+
+fn is_storage_write_builtin(callee: &hir::Expr<'_>) -> bool {
+    let ExprKind::Ident(resolutions) = &callee.peel_parens().kind else { return false };
+    resolutions.iter().any(|resolution| {
+        matches!(
+            resolution,
```
stevencartavia: `tstore` should not be treated as a persistent state-variable write. Transient and persistent storage are separate namespaces, so `tstore(protectedValue.slot, 1)` currently warns even though `protectedValue` is unchanged. Could transient roots be tracked separately from `sstore` roots?

crates/lint/src/sol/high/protected_vars.rs
```diff
@@ -0,0 +1,1239 @@
...
+        }
+        ExprKind::Call(callee, args, options) => {
+            recurse(callee);
+            if let Some(options) = options {
+                for option in options.args {
+                    recurse(&option.value);
+                }
+            }
+            for argument in args.exprs() {
+                recurse(argument);
```
stevencartavia, line 1474: Resolved Yul call results inherit the roots of every argument here, even when the return summary selects only one. For example, a `pickSecond(protectedValue.slot, ordinary.slot)` helper that returns only its second argument still warns for `protectedValue`. Could `slot_roots` use the resolved call-return summary without recursively adding all call arguments?

crates/lint/src/sol/high/protected_vars.rs
mattsse: Addressed in b5aeb2f51. Modifier expansion now follows only reachable placeholder continuations, including nested modifier aborts, while function returns still execute reachable modifier postludes. Regressions cover blocked bodies and nested return/revert ordering.

crates/lint/src/sol/high/protected_vars.rs
mattsse: Addressed in b5aeb2f51. Protection is now must-flow state: branch, short-circuit, ternary, loop, and return joins intersect guards, and each write is checked against the guards that dominate that occurrence. The one-branch and post-write cases now warn, with additional fresh-review regressions.

crates/lint/src/sol/high/protected_vars.rs
mattsse: Addressed in b5aeb2f51. Local Yul storage-pointer slot assignments now retarget the base local alias instead of recording a write through the old root. Tests cover retargeting both away from and toward protected storage.

crates/lint/src/sol/high/protected_vars.rs
mattsse: Addressed in b5aeb2f51. Only the persistent sstore builtin records a state-variable write; tstore is excluded because it uses the transient namespace. A regression keeps tstore on a protected slot clean.

crates/lint/src/sol/high/protected_vars.rs
mattsse: Addressed in b5aeb2f51. Resolved Yul calls now use the per-output return summary directly and do not union every argument root. The pick-second regression confirms that an unused protected argument no longer contaminates the ordinary returned slot.

crates/lint/src/sol/high/protected_vars.rs
```diff
@@ -0,0 +1,1407 @@
...
+                    (true, false) => then_state,
+                    _ => merge_flow_states(&then_state, &else_state),
+                };
+                self.set_flow_state(merged);
+                then_continues || else_continues
+            }
+            StmtKind::Try(try_statement) => {
+                self.analyze_expr(&try_statement.expr);
+                let before = self.flow_state();
+                let mut merged = before.clone();
```
mablr: `try` always selects one success/catch clause, but seeding `merged` with the pre-try state models an extra path that skips every clause. As a result, a required guard called in every success/catch clause does not dominate a following write and produces a false positive. I reproduced this with `try this.externalCall() { guard(); } catch { guard(); } protectedValue = 1;`: the UI test emitted `protected-vars`. Please merge only continuing clause states (and similarly avoid the synthetic no-case path for an exhaustive Yul `switch` with `default`).

mattsse: Fixed in `cb77406d1`. `try` now merges only continuing success/catch clause states, and exhaustive Yul switches no longer retain the incoming synthetic no-case path when a `default` exists. Added clean regressions for guards in every try clause and storage-pointer retargeting in every exhaustive switch arm. Full Forge UI, focused `forge-lint` clippy with warnings denied, nightly fmt, and diff checks pass; a fresh-context review found no issues.

crates/lint/src/sol/high/protected_vars.rs
```diff
@@ -0,0 +1,1418 @@
...
+            self.aliases = saved_aliases;
+            self.guards = summary.guards;
+            return summary.returns;
+        }
+        if !self.seen_calls.insert(context.clone()) {
+            self.aliases = saved_aliases;
+            return Vec::new();
+        }
+
+        let returns = self.analyze_function(function_id);
```
mablr: `analyze_function` determines whether the callee completes, but `analyze_call` only retains its storage returns and guards. An internal helper containing only `revert()` therefore appears to return normally: with `alwaysReverts(); protectedValue = 1;`, the unreachable assignment emits `protected-vars`. Please propagate callee completion state so analysis stops after a non-completing internal call.

mattsse: Fixed in 9d7c0873f. Internal-call completion now propagates through recursive fixed-point summaries, loop exits, LHS evaluation, and modifier postludes, so unreachable protected writes are not reported. Added focused regressions; the full lint UI suite passes.

crates/lint/src/sol/high/protected_vars.rs
```diff
@@ -0,0 +1,1582 @@
...
+            let roots = self.storage_roots_for_output(expression, index, variables.len());
+            self.set_storage_alias_roots(variable_id, roots);
+            if self.assembly_depth > 0 {
+                self.set_slot_alias(variable_id, expression);
+            }
+        }
+    }
+
+    fn set_return_aliases(&mut self, expression: &'hir hir::Expr<'hir>) {
+        let Some(&function_id) = self.stack.last() else { return };
```
mablr, line 1184: Per-output precision is discarded here: `storage_roots_for_output` selects the Yul return, then `set_slot_alias` recomputes roots from the whole call and unions all returns. Thus `let a, b := pair(protected.slot, ordinary.slot); sstore(b, 1)` falsely reports a protected write. Please preserve the corresponding output here and in tuple assignment.

mablr (approved):
LGTM

figtracer (approved):

PR #15768, opened 2026-07-15
feat(lint): add reentrancy-unlimited-gas

mattsse:
This adds the info-severity `reentrancy-unlimited-gas` lint from #14381. It reports Solidity's built-in `transfer` and `send` when a persistent state write or event emission remains reachable after a successful interaction, without treating the fixed stipend as a reentrancy boundary.

The flow analysis resolves builtins semantically and follows helpers, modifiers, inheritance, branches, loops, storage aliases, and Yul effects. Stored `send` results are tracked as must-facts across assignments and joins, invalidated by overwrites and repeated static spans, and used to prune failed-only paths. Rollback-only siblings, non-returning paths, and failed sends do not report; do-while `continue` transitions and Yul `sstore`, `tstore`, logs, switches, helpers, and loop posts are modeled explicitly. Fresh-context reviews added regressions for every corrected path and found no remaining issues.

The user documentation is in foundry-rs/book#1983, with the review-aligned flow follow-up in foundry-rs/book#1996.

### DIFF
--- a/crates/lint/src/sol/high/reentrancy.rs
+++ b/crates/lint/src/sol/high/reentrancy.rs
@@ -85,6 +85,7 @@ fn is_entry_point(func: &hir::Function<'_>) -> bool {
 struct FlowState {
     state_reads: BTreeSet<VariableId>,
     pending_calls: Vec<PendingCall>,
+    stored_send_results: Vec<StoredSendResult>,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
@@ -92,13 +93,32 @@ struct PendingCall {
     span: Span,
     kind: ReentrantCallKind,
     state_reads: BTreeSet<VariableId>,
+    result_correlatable: bool,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+struct SendResultSource {
+    call_span: Span,
+    succeeds_when_true: bool,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+struct StoredSendResult {
+    variable: VariableId,
+    source: SendResultSource,
 }
 
 #[derive(Default)]
 struct LoopJumps<'hir> {
     breaks: Option<FlowState>,
     continues: Option<FlowState>,
-    continue_epilogue: Option<&'hir hir::Expr<'hir>>,
+    continue_epilogue: Option<ContinueEpilogue<'hir>>,
+}
+
+#[derive(Clone, Copy)]
+enum ContinueEpilogue<'hir> {
+    Expr(&'hir hir::Expr<'hir>),
+    Block(hir::Block<'hir>),
 }
 
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
@@ -165,17 +185,37 @@ impl FlowState {
             self.pending_calls.iter_mut().find(|call| call.span == span && call.kind == kind)
         {
             existing.state_reads.extend(self.state_reads.iter().copied());
+            existing.result_correlatable = false;
+            self.stored_send_results.retain(|result| result.source.call_span != span);
         } else {
             self.pending_calls.push(PendingCall {
                 span,
                 kind,
                 state_reads: self.state_reads.clone(),
+                result_correlatable: true,
             });
         }
     }
 
     fn remove_call(&mut self, span: Span, kind: ReentrantCallKind) {
         self.pending_calls.retain(|call| call.span != span || call.kind != kind);
+        self.stored_send_results.retain(|result| result.source.call_span != span);
+    }
+
+    fn set_stored_send_results(&mut self, variable: VariableId, sources: &[SendResultSource]) {
+        self.stored_send_results.retain(|result| result.variable != variable);
+        for &source in sources {
+            if self.pending_calls.iter().any(|call| {
+                call.span == source.call_span
+                    && call.kind == ReentrantCallKind::Stipend
+                    && call.result_correlatable
+            }) {
+                let result = StoredSendResult { variable, source };
+                if !self.stored_send_results.contains(&result) {
+                    self.stored_send_results.push(result);
+                }
+            }
+        }
     }
 }
 
@@ -187,14 +227,14 @@ fn merge_optional_state(target: &mut Option<FlowState>, state: &FlowState) {
     }
 }
 
-/// Returns the update expression that Solar lowers after a `for` loop body.
+/// Returns the Solidity update expression or Yul post block lowered after a `for` loop body.
 ///
-/// A source-level `continue` skips this synthetic statement in HIR, even though Solidity executes
-/// the update before the next condition check.
+/// A source-level `continue` skips this synthetic statement in HIR, even though the source
+/// language executes it before the next condition check.
 fn for_loop_continue_epilogue<'hir>(
     block: hir::Block<'hir>,
     source: LoopSource,
-) -> Option<&'hir hir::Expr<'hir>> {
+) -> Option<ContinueEpilogue<'hir>> {
     if source != LoopSource::For || block.stmts.len() != 1 {
         return None;
     }
@@ -213,8 +253,28 @@ fn for_loop_continue_epilogue<'hir>(
         return None;
     }
 
-    let StmtKind::Expr(epilogue) = body.stmts.last()?.kind else { return None };
-    Some(epilogue)
+    match body.stmts.last()?.kind {
+        StmtKind::Expr(epilogue) => Some(ContinueEpilogue::Expr(epilogue)),
+        StmtKind::Block(epilogue) => Some(ContinueEpilogue::Block(epilogue)),
+        _ => None,
+    }
+}
+
+/// Returns the user statements and condition from Solar's lowered `do-while` loop.
+fn do_while_parts<'hir>(
+    block: hir::Block<'hir>,
+    source: LoopSource,
+) -> Option<(&'hir [hir::Stmt<'hir>], &'hir hir::Expr<'hir>)> {
+    if source != LoopSource::DoWhile {
+        return None;
+    }
+
+    let (check, body) = block.stmts.split_last()?;
+    let StmtKind::If(condition, then_stmt, Some(else_stmt)) = check.kind else { return None };
+    if !matches!(then_stmt.kind, StmtKind::Continue) || !matches!(else_stmt.kind, StmtKind::Break) {
+        return None;
+    }
+    Some((body, condition))
 }
 
 fn is_state_mutating_array_call(gcx: Gcx<'_>, callee: &hir::Expr<'_>) -> bool {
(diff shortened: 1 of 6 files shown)


crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -425,40 +713,262 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
+    ) {
+        if !self.reentrancy_unlimited_gas_enabled {
             return;
         }
 
+        let calls = state
+            .pending_calls
+            .iter()
+            .filter(|call| {
+                call.kind == ReentrantCallKind::Stipend && condition.span.contains(call.span)
```
mablr: The failed-send pruning only recognizes a send nested directly inside the current condition. If the result is stored first (for example, `bool success = recipient.send(1); if (!success) counter++;`), the pending call survives and the failed-only write is reported.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -36,6 +38,13 @@ declare_forge_lint!(
     "state read before external call is written after the call"
 );
 
+declare_forge_lint!(
```
figtracer, line 41: [solc 0.8.31](https://github.com/argotorg/solidity/releases/tag/v0.8.31) already warns on every built-in `send`/`transfer` and recommends `call`, so this duplicates a recent compiler diagnostic while adding flow-analysis false positives. per #14381, can we omit this detector?

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -229,48 +368,101 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
+            StmtKind::Break => {
+                if let Some(jumps) = self.loop_jumps.last_mut() {
+                    merge_optional_state(&mut jumps.breaks, state);
                 }
                 false
             }
-            StmtKind::Break | StmtKind::Continue => false,
-            StmtKind::Loop(block, _) => {
+            StmtKind::Continue => {
+                let epilogue = self.loop_jumps.last().and_then(|jumps| jumps.continue_epilogue);
```
stevencartavia, line 615: A source-level `continue` in a `do-while` must evaluate the condition, but this records a direct backedge and skips Solar's synthesized `if (condition) continue; else break;`. For example:

```solidity
do {
    recipient.transfer(1 wei);
    continue;
} while (again);

counter = 1;
```

produces no warning, although `counter = 1` is reached when `again` is false. Could we add a `do-while`-specific condition transition and regression tests for effects after the loop and in the condition?

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -425,40 +713,262 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
+        let mut all_complete = true;
+        for &expr in exprs {
+            let prior_calls = state
+                .pending_calls
+                .iter()
+                .filter(|call| call.kind == ReentrantCallKind::Stipend)
+                .map(|call| call.span)
+                .collect::<HashSet<_>>();
+            let prior_effects = self.state_effects;
+            all_complete &= self.analyze_expr(expr, state);
```
stevencartavia, line 1045: `analyze_unordered_exprs` can report a stipend call when another argument cannot return. For example:

```solidity
consume(recipient.send(1 wei), writeThenRevert());
```

reports the `send` when `writeThenRevert()` writes and then reverts, even though the enclosing call cannot complete in either argument order and no effect survives. Could we suppress diagnostics from non-completing paths and add this as a negative test? Guarding only this summary is insufficient because writes emit pending calls eagerly and helper summaries track `has_state_effect` separately from `may_return`.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -316,107 +523,188 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
         }
     }
 
-    fn analyze_expr(&mut self, expr: &'hir hir::Expr<'hir>, state: &mut FlowState) {
+    fn analyze_expr(&mut self, expr: &'hir hir::Expr<'hir>, state: &mut FlowState) -> bool {
```
mablr, line 794: The new flow analysis still treats every assembly block as effect-free. A fixture with `recipient.transfer(1 wei); assembly { sstore(counter.slot, 1) }` emits no warning even though the write follows the interaction. Please inspect Yul storage writes or conservatively treat assembly reached with a pending stipend call as a state effect.

crates/lint/src/sol/high/reentrancy.rs
mattsse: Stored send results are now tracked as must-facts through declarations, assignments, tuples, branches, and delete. Failed-only paths prune only the correlated send, while a repeated static send span disables correlation so a later failure cannot erase an earlier successful interaction. Both cases have regressions.

crates/lint/src/sol/high/reentrancy.rs
mattsse: Source-level continue in a do-while now transitions through the synthesized condition before either the backedge or loop exit. The fixture covers effects in the condition and effects reached after the loop when the condition becomes false.

crates/lint/src/sol/high/reentrancy.rs
mattsse: Unordered siblings now commit diagnostics only from paths where the enclosing expression can complete. Writes and events on rollback-only paths are discarded, including the writeThenRevert case, with negative regression coverage.

crates/lint/src/sol/high/reentrancy.rs
mattsse: The analysis now follows Yul sstore, tstore, and log effects through blocks, switch clauses, helper calls, and for-loop post blocks, including control flow after continue. Direct and indirect assembly effects after a stipend call are covered by regressions.

crates/lint/src/sol/high/reentrancy.rs
mattsse: I kept the detector because #14381 explicitly tracks it and Foundry linting also serves projects on older compiler versions or with compiler diagnostics configured differently. The reviewed implementation is path-aware and now excludes failed sends, rollback-only effects, and other false-positive shapes. I am leaving this policy thread open for maintainer preference.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -229,48 +477,129 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
+                } else {
+                    true
                 }
-                true
-            }
-            StmtKind::DeclMulti(_, expr) | StmtKind::Expr(expr) => {
-                self.analyze_expr(expr, state);
-                true
             }
+            StmtKind::DeclMulti(_, expr) | StmtKind::Expr(expr) => self.analyze_expr(expr, state),
```
mablr: `DeclMulti` never records which declared variable received a `send` result, so failed-only continuations become false positives. For example, `(bool success, uint256 value) = (recipient.send(1 wei), 1); if (!success) counter = value;` retains the pending call and reports even though the write is reachable only when the callback failed. Please propagate each tuple initializer's send source to its corresponding declaration (and add a negative regression).

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -425,40 +900,392 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
+                let invert = (op.kind == BinOpKind::Eq) != literal;
+                self.send_result_sources(source_expr, state)
+                    .into_iter()
+                    .map(|source| SendResultSource {
+                        succeeds_when_true: source.succeeds_when_true != invert,
+                        ..source
+                    })
+                    .collect()
+            }
+            _ => Vec::new(),
```
mablr, line 1505: Send-result correlation is lost across internal helper returns. For example, `bool success = sendInHelper(recipient); if (!success) counter = 1;`, where `sendInHelper` returns `recipient.send(1 wei)`, reports the helper's send even though the write is on the failed-only path. Since the analyzer follows helper effects, it also needs to summarize/propagate returned send-result sources (or otherwise prune this path); please add a negative regression.

crates/lint/src/sol/high/reentrancy.rs
mattsse: Fixed in `3458d1146`. Multi-declarations now correlate each declared variable with its matching tuple or call-result slot, so failed-only continuations discard the pending send. Added the exact tuple regression plus success/sibling positive controls.

crates/lint/src/sol/high/reentrancy.rs
mattsse: Fixed in `3458d1146`. Internal call summaries now propagate indexed send-result sources through explicit returns, named fallthrough returns, cache hits, direct helper conditions, and multi-result helpers, while unknown/overridden paths deliberately drop correlation so warnings remain. Added failed-only and positive regressions, including nested modifiers and a nonzero tuple return slot. Full UI/bless, `forge-lint` clippy with warnings denied, nightly fmt, and diff checks pass; fresh review findings were addressed and re-review is clean.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -307,116 +733,274 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
                 }
-                self.analyze_expr(rhs, state);
-                self.analyze_lhs_indices(lhs, state);
-                let written_vars = state_write_lhs_vars(self.hir, lhs);
-                if !written_vars.is_empty() {
-                    self.emit_pending_calls(state, &written_vars);
+                completes &= self.analyze_expr(rhs, state);
+                completes &= self.analyze_lhs_indices(lhs, state);
+                if completes {
+                    let direct_variable = assigned_variable(lhs);
```
mablr: Tuple assignments drop the stored `send` result correlation because `assigned_variable(lhs)` is `None`. `(success, value) = (recipient.send(1 wei), 1); if (!success) counter = value;` is therefore warned even though the write is reachable only after a failed send. Please propagate result sources per tuple slot as the declaration path does.

mattsse: Fixed in 9bfb76537. Tuple assignments now align LHS/RHS structures recursively and apply source writes in Solidity order, preserving holes, sibling isolation, stale clearing, and duplicate-LHS overwrite behavior. Added flat, nonzero-slot, discriminating sibling, and nested-tuple regressions. Full UI/clippy/fmt validation and final fresh-context re-review pass; the policy thread remains untouched.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -425,40 +1006,581 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
+        }
+    }
+
+    fn send_result_sources(
+        &self,
+        expr: &'hir hir::Expr<'hir>,
+        state: &FlowState,
+    ) -> Vec<SendResultSource> {
+        let expr = expr.peel_parens();
+        match &expr.kind {
```
mablr, line 1443: Send-result correlation is dropped through ternaries and short-circuit expressions. Both `bool ok = attempt ? recipient.send(1 wei) : false; if (!ok) counter = 1;` and `bool ok = attempt && recipient.send(1 wei); if (!ok) counter = 1;` warn, although every path reaching the write either skipped the send or observed failure. Please preserve path-sensitive result correlation through `?:`, `&&`, and `||`.

mattsse: Fixed in a32a45e5c. Send-result success and failure remain correlated through ternary, &&, and || expressions, including repeated aliases, so failure-only branches do not warn. Added focused regressions; the full lint UI suite passes.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -307,116 +733,271 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
+                        let mut returned_state = FlowState::default();
+                        let mut any_returns = false;
+                        for func_id in func_ids {
+                            let mut candidate_state = before_call.clone();
+                            if let Some(return_sources) =
+                                self.analyze_internal_call(func_id, &mut candidate_state)
+                            {
+                                candidate_state.set_stored_call_results(expr.span, &return_sources);
+                                returned_state.merge(&candidate_state);
+                                any_returns = true;
```
mablr, line 893: Send-result correlation is lost when passed into an internal helper: `bool ok = recipient.send(1); failedOnly(ok)`, with `if (!ok) counter = 1`, warns even though the write is failure-only. Please bind actual send-result sources to the callee parameters and retain a success-path control.

mablr:
Closing because solc 0.8.31 already diagnoses built-in send and transfer usage, and Foundry avoids duplicating diagnostics covered by recent compiler versions.

PR #15769, opened 2026-07-15
feat(lint): detect balance reentrancy

mattsse:
This adds the high-severity `reentrancy-balance` lint from #14381. It reports reentrant external calls between saving `address(this).balance` and comparing a fresh contract balance against that stale local value on the opposite comparison operand.

The analysis follows casts, tuples, derived locals, helper parameters and returns, modifiers, inherited dispatch, loops, and both operand orders while excluding token balances, mutable storage baselines, view/static calls, overwritten baselines, and mutually exclusive checks. Explicit gas caps include the value-transfer stipend. Strict standard state-lock modifiers suppress this detector only when the same lock protects every effective mutable re-entry point, and bounded recursion remains available to the existing reentrancy detectors. Fresh-context review additionally corrected modifier ordering around the lock model.

The companion Book documentation is in foundry-rs/book#1984, with the lock and stipend follow-up in foundry-rs/book#1993. This PR was implemented with Codex assistance across code, tests, and lint documentation, then independently reviewed and corrected by fresh Codex contexts.


crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -715,6 +939,299 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
+                        | BinOpKind::Eq
+                        | BinOpKind::Ne
+                );
+                let current_locals = state
+                    .balance_locals
+                    .difference(&call.stale_locals)
+                    .copied()
+                    .collect::<BTreeSet<_>>();
+                (is_comparison
+                    && self.expr_depends_on_balance(
```
mablr: This checks both dependencies against the entire comparison, so it also reports when the fresh and stale balances occur on the same operand, e.g. `require(address(this).balance + balanceBefore > 0)`. That guard does not compare the current balance against the stale baseline and cannot establish the payment condition this lint targets.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -352,24 +473,55 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
                 }
                 if !state.state_reads.is_empty()
                     && let Some(kind) = self.reentrant_call_kind(callee, args, *opts)
                 {
                     state.push_call(expr.span, kind);
                 }
+                if self.reentrancy_balance_enabled
+                    && is_balance_reentrant_call(self.gcx, self.hir, callee, args, *opts)
+                {
+                    state.push_balance_call(expr.span);
```
figtracer, line 678: this reproduces [the known slither `nonReentrant` false positive](https://github.com/crytic/slither/issues/735): even with the only external entry point wrapped by a standard lock, this still warns. please model re-entry reachability or omit this high-severity detector.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -430,35 +582,107 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
-        if self.call_stack.contains(&func_id) {
-            return;
+    fn analyze_internal_call(
+        &mut self,
+        func_id: FunctionId,
+        args: &CallArgs<'hir>,
+        state: &mut FlowState,
+    ) -> Vec<BalanceValue> {
+        if self.call_stack.contains(&func_id)
+            || self.hir.function(func_id).span == self.entry_function_span
```
stevencartavia: The new `entry_function_span` guard skips the first recursive invocation of the current public entry point, regressing the existing `reentrancy-no-eth` lint.

```solidity
contract RecursiveRegression {
    uint256 x;

    function f(uint256 depth, address target) public {
        if (depth != 0) {
            uint256 snapshot = x;
            f(depth - 1, target);
            x = snapshot;
        } else {
            (bool ok,) = target.call("");
            if (!ok) revert();
        }
    }
}
```

This warns at `target.call("")` on the base revision but not on this PR. Is that regression intentional? The existing call-stack check would still bound later recursive calls.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -752,6 +1271,144 @@ impl FlowState {
...
+}
+
+fn call_options_allow_reentrancy(hir: &hir::Hir<'_>, opts: Option<&hir::CallOptions<'_>>) -> bool {
+    let Some(gas) = opts.and_then(|opts| opts.args.iter().find(|opt| opt.name.name == kw::Gas))
+    else {
+        return true;
+    };
+    let mut seen = BTreeSet::new();
+    concrete_gas_cap(hir, &gas.value, &mut seen)
+        .is_none_or(|gas| gas > U256::from(REENTRANCY_GAS_STIPEND))
```
stevencartavia: `call_options_allow_reentrancy` compares the explicit `gas` value to 2,300 as if it were the callee’s entire gas budget, but a value-bearing `CALL` receives the capped requested gas plus the 2,300-gas stipend.

```solidity
interface Callback {
    function pay() external payable;
}

contract BalanceRegression {
    function f(Callback callback, uint256 amount) external payable {
        uint256 beforeBalance = address(this).balance;
        callback.pay{value: 1, gas: 2300}();
        require(address(this).balance >= beforeBalance + amount);
    }
}
```

This produces no `reentrancy-balance` warning, while the no-value `{gas: 2301}` control does. An equivalent low-level call observed 4,573 gas at `gasleft()` inside the fallback. The cap check should account for a `value` option that is nonzero or not provably zero.

crates/lint/src/sol/high/reentrancy.rs
mattsse: Addressed in 65bf0620d. The balance detector now requires fresh and stale dependencies on opposite comparison operands; same-operand arithmetic remains clean, with both polarities covered by regressions.

crates/lint/src/sol/high/reentrancy.rs
mattsse: Addressed in 65bf0620d. The analysis recognizes strict standard state-lock modifiers, verifies the same lock protects every effective mutable re-entry point across deployed interfaces, and suppresses only reentrancy-balance while preserving the other detectors. Helper-based guards, ordering, delegate-context calls, invalidation, and malformed controls are covered; a fresh review also fixed later-modifier ordering.

crates/lint/src/sol/high/reentrancy.rs
mattsse: Addressed in 65bf0620d. Bounded recursive entry analysis is restored for the existing reentrancy detectors, and the supplied reentrancy-no-eth shape is pinned as a regression without reopening unbounded recursion.

crates/lint/src/sol/high/reentrancy.rs
mattsse: Addressed in 65bf0620d. Explicit gas caps now include the 2,300-gas value-transfer stipend whenever value may be nonzero, while provably zero-value calls retain the cap-only behavior. The value-bearing 2,300 and zero-value 2,301 controls are covered.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -715,6 +980,306 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
+            ExprKind::Call(callee, args, opts)
+                if opts.is_none()
+                    && matches!(
+                        callee.peel_parens().kind,
+                        ExprKind::Type(_) | ExprKind::TypeCall(_)
+                    ) =>
+            {
+                args.exprs().any(|arg| self.guard_has_stale_balance_comparison(arg, call, state))
+            }
+            _ => false,
```
mablr, line 1296: This misses stale-balance comparisons returned by internal boolean helpers. I verified no warning for `beforeBalance = address(this).balance; callback.pay(); require(isPaid(address(this).balance, beforeBalance, amount));` where `isPaid(current, beforeBalance, amount)` returns `current >= beforeBalance + amount`. Internal-call return tracking currently collapses dependencies into one `BalanceValue`, so the comparison relationship is unavailable here; please propagate/check the helper comparison shape.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -752,6 +1339,450 @@ impl FlowState {
...
+        Self {
+            balance_locals: self.balance_locals.clone(),
+            pending_balance_calls: self.pending_balance_calls.clone(),
+            invalidated_balance_guards: self.invalidated_balance_guards.clone(),
+            ..Self::default()
+        }
+    }
+
+    fn merge_balance(&mut self, other: &Self) {
+        self.balance_locals.extend(other.balance_locals.iter().copied());
```
mablr, line 1889: The branch merge loses predicate correlation and produces a false positive across sequential complementary conditions. I verified that `if (takeBaseline) beforeBalance = address(this).balance; if (!takeBaseline) callback.pay(); require(address(this).balance >= beforeBalance + amount);` warns at `pay()`, although no execution can both save that baseline and make the call. Please preserve enough path information to keep mutually exclusive paths clean beyond a single `if`/`else`.

mattsse: Fixed in e14c09e7e. Balance locals now retain boolean path provenance across complementary conditions, while branch, ternary, loop, and try joins remove facts invalidated by mutation. Added complementary-path and reachable-warning controls plus the review-found mutation cases; full UI/clippy/fmt and fresh-context re-review pass.

crates/lint/src/sol/high/reentrancy.rs
mattsse: Fixed in e14c09e7e. Internal helper return summaries now preserve stale-balance comparison shape through direct, named, and compound-assigned boolean returns, including parameter-to-argument predicate remapping and cache-key separation. Added focused positive and negative regressions; the full lint UI suite, clippy, fmt, and a fresh-context re-review pass.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -752,173 +1547,660 @@ impl FlowState {
...
+    gcx.type_of_expr(callee.peel_parens().id).is_some_and(
+        |ty| matches!(ty.kind, TyKind::Fn(function) if function.kind == TyFnKind::DelegateCall),
+    )
+}
+
+fn is_self_balance(expr: &hir::Expr<'_>) -> bool {
+    matches!(
+        &expr.peel_parens().kind,
+        ExprKind::Member(base, member)
+            if member.as_str() == "balance" && is_self_address(base)
```
mablr: An unmodified local alias of `address(this)` loses self-address provenance here. For example, `address payable self = payable(address(this)); uint256 before = self.balance; callback.pay(); require(self.balance >= before + amount);` produces no warning, although it is the same stale-balance flow as the direct form. Please preserve self-address identity through local aliases.

mattsse: Fixed in ba7bed00f. address(this) provenance now flows through locals, casts, tuple slots, helper/modifier parameters, and path-correlated pending calls, and is invalidated on reassignment, deletion, compound writes, or assembly. Added direct aliases, conditional retargeting, tuple, helper/modifier, and assembly regressions. Full validation and final fresh-context re-review pass.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -346,30 +568,75 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
+                {
+                    state.push_balance_call(expr.span);
+                }
+                if uses_delegate_context {
+                    state
+                        .invalidated_balance_guards
+                        .extend(self.active_balance_guards.iter().copied());
+                }
             }
             ExprKind::Binary(lhs, _, rhs) => {
```
mablr, line 720: Short-circuit operands are analyzed unconditionally here, which loses path exclusivity. For example, `takeBaseline && (balanceBefore = address(this).balance) > 0; if (!takeBaseline) callback.pay(); require(address(this).balance >= balanceBefore + amount);` warns even though no execution can both save the balance and make the call. Please analyze `&&`/`||` RHS execution under the corresponding path constraint.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -715,210 +1093,1261 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
+                _ => None,
+            }) {
+                if let Some(local_paths) = state.self_address_local_paths.get(var_id) {
+                    paths.extend(constrain_paths(local_paths, &state.path_predicates));
+                }
+            }
+            paths
+        }
+        ExprKind::Payable(inner) => self_address_paths(inner, state),
+        ExprKind::Call(callee, args, opts)
```
mablr, line 2531: Self-address provenance is lost when `address(this)` is returned by an internal helper. Assigning `address payable self = currentSelf();`, saving `self.balance`, making the external call, then checking `self.balance` produces no warning. Please propagate self-address provenance through internal helper return summaries, as is already done for balance values.

crates/lint/src/sol/high/reentrancy.rs
mattsse: Fixed in 8654028dc. Short-circuit and ternary path constraints now propagate recursively, including multi-term side effects and complementary paths. Added focused regressions; the full lint UI suite passes.

crates/lint/src/sol/high/reentrancy.rs
mattsse: Fixed in 8654028dc. Self-address and self-balance provenance now follows helper returns, tuple components, arguments, and member expressions while preserving feasible path constraints. Added focused regressions; the full lint UI suite passes.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -430,35 +756,123 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
+
+    fn resolved_internal_function_id(&self, callee: &'hir hir::Expr<'hir>) -> Option<FunctionId> {
+        match &callee.peel_parens().kind {
+            ExprKind::Ident(_) => {}
+            ExprKind::Member(base, _) if is_super(base) => {}
+            _ => return None,
+        }
+        let ty = self.gcx.type_of_expr(callee.peel_parens().id)?;
+        let TyKind::Fn(function) = ty.kind else { return None };
+        if function.is_internal() { function.function_id } else { None }
```
mablr: Internal helper calls through function-pointer locals are missed because their function type has no `function_id`. A stale-balance check around `function(Callback) internal fn = invoke; fn(callback);` never reaches `invoke`, so the external call produces no diagnostic. Please track possible pointer targets, including assignments and joins, or otherwise account for them.

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -715,210 +1129,1342 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
+    }
+    let body = modifier.body?;
+    if body.stmts.iter().map(count_modifier_placeholders).sum::<usize>() != 1 {
+        return None;
+    }
+    let mut placeholders = body
+        .stmts
+        .iter()
+        .enumerate()
+        .filter(|(_, stmt)| matches!(stmt.kind, StmtKind::Placeholder));
```
mablr: Placeholder counting is recursive, but this lookup only accepts a top-level `_`. A standard guard with `{ _; }` is therefore rejected and warns even though the lock covers the body. Please make the lookup consistent with the recursive count while preserving unconditional-path semantics.

mablr (dismissed):
LGTM

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -715,210 +1166,1361 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
+    modifier: &hir::Function<'_>,
+) -> Option<VariableId> {
+    if !matches!(modifier.kind, hir::FunctionKind::Modifier) || !modifier.modifiers.is_empty() {
+        return None;
+    }
+    let body = modifier.body?;
+    if body.stmts.iter().map(count_modifier_placeholders).sum::<usize>() != 1 {
+        return None;
+    }
+    let placeholder_index = body.stmts.iter().position(contains_unconditional_placeholder)?;
```
figtracer: this treats a nested block as if it were only `_`. i reproduced no warning with `{ locked = 0; _; }`, so the modifier is classified as non-reentrant even though the body runs unlocked. can we include statements before the nested placeholder in the activation check?

mablr: https://github.com/foundry-rs/foundry/pull/15769/commits/2111d025190884c32df10a0919f5b2faf9d28cb3

mablr (dismissed):

crates/lint/src/sol/high/reentrancy.rs
```diff
@@ -715,210 +1166,1388 @@ impl<'ctx, 's, 'c, 'hir> Analyzer<'ctx, 's, 'c, 'hir> {
...
+                    Some(*var_id)
+                }
+                _ => None,
+            }))?;
+            Some((var_id, true))
+        }
+        ExprKind::Unary(op, inner) if op.kind == UnOpKind::Not => {
+            path_predicate(hir, inner).map(|(var_id, value)| (var_id, !value))
+        }
+        _ => None,
```
stevencartavia, line 1944: `path_predicate` does not recognize complementary comparisons such as `mode == 0` and `mode != 0`, causing a false positive here:

```solidity
uint256 beforeBalance = address(this).balance;
if (mode == 0) callback.pay();
if (mode != 0) require(address(this).balance >= beforeBalance + amount);
```

No path can execute both the call and comparison. Could we track direct complementary `==`/`!=` conditions so these mutually exclusive paths remain separate?

stevencartavia (approved):

mablr (approved):

PR #15770, opened 2026-07-15
fix(anvil): enforce simulate block gas limit

mablr:
## Motivation

This addresses the block gas accounting portion of #13823 as the first targeted step toward full `eth_simulateV1` conformance. Calls without an explicit gas limit now receive the remaining block gas, calls requesting more than the remainder return `-38015`, and calls exhausting that budget report an execution failure with `-32015` without allowing reported block gas to exceed its limit.

The remaining parts of #13823 will be handled through separate focused PRs (superseding #15716). Closes OSS-563

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -5049,13 +5049,19 @@ impl Backend<FoundryNetwork> {
             } = request;
             let mut cache_db = CacheDB::new(state);
             let mut block_res = Vec::with_capacity(block_state_calls.len());
+            let (is_amsterdam, tx_gas_limit_cap) = {
+                let cfg_env = &self.evm_env.read().cfg_env;
+                (cfg_env.spec >= SpecId::AMSTERDAM, cfg_env.tx_gas_limit_cap())
+            };
 
             // execute the blocks
             for block in block_state_calls {
                 let SimBlock { block_overrides, state_overrides, calls } = block;
                 let mut call_res = Vec::with_capacity(calls.len());
                 let mut log_index = 0;
-                let mut gas_used = 0;
+                let mut cumulative_gas_used = 0;
+                let mut block_regular_gas_used = 0;
+                let mut block_state_gas_used = 0;
                 let mut transactions = Vec::with_capacity(calls.len());
                 let mut logs= Vec::new();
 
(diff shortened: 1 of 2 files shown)


crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5068,7 +5068,21 @@ impl Backend<FoundryNetwork> {
                 }
 
                 // execute all calls in that block
-                for (req_idx, request) in calls.into_iter().enumerate() {
+                for (req_idx, mut request) in calls.into_iter().enumerate() {
+                    let remaining_gas = block_env.gas_limit.saturating_sub(gas_used);
```
mattsse: Could we preserve Amsterdam’s EIP-8037 accounting here? Foundry supports `EthereumHardfork::Amsterdam`, where regular gas and state gas have independent block budgets. Subtracting `result.tx_gas_used()` from this single accumulator can prematurely cap or reject a valid later call with `-38015`. Please track `block_regular_gas_used()` and `block_state_gas_used()` separately when Amsterdam is active (as reth’s simulation path does), and add an Amsterdam regression test.

mattsse (changes requested):
The block gas-limit fix needs to account for Amsterdam’s split gas dimensions before merging.

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5068,7 +5068,21 @@ impl Backend<FoundryNetwork> {
...
+                    let requested_gas = request.gas.unwrap_or(remaining_gas);
```
stevencartavia, line 5103: When the remaining block gas is below the next call’s intrinsic cost, an omitted `gas` returns the wrong RPC error. With a 41,999 gas block and two plain transfers, Anvil returns `-32000`, while Geth returns the expected `eth_simulateV1` error `-38013`. At 42,000 gas, both calls succeed. Could we map this failure to `-38013` and add a regression test?

mablr: Already tracked in OSS-561 👍

crates/anvil/src/eth/backend/mem/mod.rs
mablr: https://github.com/foundry-rs/foundry/pull/15770/commits/505d256ad57598c684112fa4a78ae32580ee529a

mattsse (approved):
lgtm

stevencartavia (approved):

PR #15771, opened 2026-07-15
perf(debugger): avoid repeated storage scans

stevencartavia:
Avoids rebuilding the storage access map for every repeated scroll step in the debugger’s storage explorer. The pane length is now calculated once before applying the requested scroll count, keeping large jumps responsive on traces with many accessed slots.


mattsse (approved):
lgtm

figtracer (approved):

PR #15772, opened 2026-07-15
fix(forge): clean up failed installs

### LINKED ISSUE #6790: failed "forge install" should leave workspace untouched.
### Component

Forge

- [ ] Foundry
- [ ] Foundryup

_No response_

forge 0.2.0 (8343e7d 2024-01-07T00:19:11.564229000Z)

### Operating System

macOS (Apple Silicon)

### Describe the bug

A failed attempt at `forge install` (e.g, with broken repo link, like unknown branch) currently leaves the local storage in a "dirty" state, requiring "git reset".
Just like "git" itself, forge should clean up any incomplete work data

stevencartavia:
Rolls back new submodules when `forge install` fails, leaving the repository clean. 

Fixes #6790.

### DIFF
--- a/crates/cli/src/utils/mod.rs
+++ b/crates/cli/src/utils/mod.rs
@@ -724,7 +724,7 @@ ignore them in the `.gitignore` file."
         path: impl AsRef<OsStr>,
     ) -> Result<()> {
         self.cmd()
-            .env("GIT_LITERAL_PATHSPECS", "1")
+            .arg("--literal-pathspecs")
             .stderr(self.stderr())
             .args(["submodule", "add"])
             .args(self.shallow.then_some("--depth=1"))
--- a/crates/cli/src/utils/mod.rs
+++ b/crates/cli/src/utils/mod.rs
@@ -475,7 +475,7 @@ impl<'a> Git<'a> {
         I: IntoIterator<Item = S>,
         S: AsRef<OsStr>,
     {
-        self.cmd().arg("rm").args(force.then_some("--force")).arg("--").args(paths).exec().map(drop)
+        self.cmd().arg("rm").args(force.then_some("--force")).args(paths).exec().map(drop)
     }
 
     pub fn remove_index_path(self, path: &Path) -> Result<()> {
@@ -527,19 +527,6 @@ impl<'a> Git<'a> {
             .arg(path)
             .exec()
             .map(|out| out.stdout.is_empty())
-    }
-
-    /// Returns whether a path has no unstaged changes relative to the index.
-    pub fn is_path_worktree_clean(self, path: &Path) -> Result<bool> {
-        let output = self.cmd().args(["diff", "--quiet", "--"]).arg(path).output()?;
-        match output.status.code() {
-            Some(0) => Ok(true),
-            Some(1) => Ok(false),
-            _ => Err(eyre::eyre!(
-                "failed to inspect path: {}",
-                String::from_utf8_lossy(&output.stderr).trim()
-            )),
-        }
     }
 
     pub fn has_branch(self, branch: impl AsRef<OsStr>, at: &Path) -> Result<bool> {
@@ -737,6 +724,7 @@ ignore them in the `.gitignore` file."
         path: impl AsRef<OsStr>,
     ) -> Result<()> {
         self.cmd()
+            .env("GIT_LITERAL_PATHSPECS", "1")
             .stderr(self.stderr())
             .args(["submodule", "add"])
             .args(self.shallow.then_some("--depth=1"))
@@ -761,13 +749,12 @@ ignore them in the `.gitignore` file."
     {
         self.cmd()
             .stderr(self.stderr())
-            .args(["--literal-pathspecs", "submodule", "update", "--progress", "--init"])
+            .args(["submodule", "update", "--progress", "--init"])
             .args(self.shallow.then_some("--depth=1"))
             .args(force.then_some("--force"))
             .args(remote.then_some("--remote"))
             .args(no_fetch.then_some("--no-fetch"))
             .args(recursive.then_some("--recursive"))
-            .arg("--")
             .args(paths)
             .exec()
             .map(drop)
@@ -797,17 +784,6 @@ ignore them in the `.gitignore` file."
         self.cmd().stderr(self.stderr()).args(["submodule", "init"]).exec().map(drop)
     }
 
-    pub fn submodule_deinit(self, force: bool, path: &Path) -> Result<()> {
-        self.cmd()
-            .stderr(self.stderr())
-            .args(["--literal-pathspecs", "submodule", "deinit"])
-            .args(force.then_some("--force"))
-            .arg("--")
-            .arg(path)
-            .exec()
-            .map(drop)
-    }
-
     pub fn submodules(&self) -> Result<Submodules> {
         self.cmd().args(["submodule", "status"]).get_stdout_lossy().map(|stdout| stdout.parse())?
     }
@@ -824,14 +800,11 @@ ignore them in the `.gitignore` file."
             .map(|url| Some(url.trim().to_string()))
     }
 
-    /// Returns whether the default section name conflicts and the mapping for an exact path.
-    pub fn submodule_mapping_for_path(
-        self,
-        path: &Path,
-    ) -> Result<(bool, Option<SubmoduleMapping>)> {
+    /// Returns whether `.gitmodules` contains the default section name or an exact path mapping.
+    pub fn has_submodule_mapping(self, path: &Path) -> Result<bool> {
         let gitmodules = self.root.join(".gitmodules");
         if !gitmodules.exists() {
-            return Ok((false, None));
+            return Ok(false);
         }
 
         let output = self
@@ -843,7 +816,6 @@ ignore them in the `.gitignore` file."
         match output.status.code() {
             Some(0) => {
                 let expected_path = path.to_slash_lossy();
-                let mut sections = HashMap::<String, (Option<String>, Option<String>)>::default();
                 for entry in
                     output.stdout.split(|byte| *byte == 0).filter(|entry| !entry.is_empty())
                 {
@@ -854,27 +826,13 @@ ignore them in the `.gitignore` file."
                     let value = std::str::from_utf8(&entry[separator + 1..])?;
                     let Some(key) = key.strip_prefix("submodule.") else { continue };
                     let Some((name, field)) = key.rsplit_once('.') else { continue };
-                    let section = sections.entry(name.to_string()).or_default();
-                    match field {
-                        "path" if section.0.is_none() => section.0 = Some(value.to_string()),
-                        "url" if section.1.is_none() => section.1 = Some(value.to_string()),
-                        _ => {}
+                    if name == expected_path || field == "path" && value == expected_path {
+                        return Ok(true);
                     }
                 }
-
-                let default_name_exists = sections.contains_key(expected_path.as_ref());
-                let mut mappings = sections
-                    .into_iter()
-                    .filter(|(_, (mapped_path, _))| mapped_path.as_deref() == Some(&expected_path))
-                    .map(|(name, (_, url))| SubmoduleMapping { name, url });
-                let mapping = mappings.next();
-                let multiple_mappings = mappings.next().is_some();
-                let name_conflict = multiple_mappings
-                    || default_name_exists
-                        && mapping.as_ref().is_none_or(|mapping| mapping.name != expected_path);
-                Ok((name_conflict, (!multiple_mappings).then_some(mapping).flatten()))
+                Ok(false)
             }
-            Some(1) => Ok((false, None)),
+            Some(1) => Ok(false),
             _ => Err(eyre::eyre!(
                 "failed to inspect .gitmodules: {}",
                 String::from_utf8_lossy(&output.stderr).trim()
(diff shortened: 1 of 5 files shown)


crates/forge/src/cmd/install.rs
```diff
@@ -402,35 +407,99 @@ impl Installer<'_> {
...
     fn install_as_submodule(self, dep: &Dependency, path: &Path) -> Result<Option<String>> {
+        let root = Git::root_of(self.git.root)?;
+        let relative_path = path.strip_prefix(&root)?;
+        let git = self.git.root(&root);
+        let gitmodules = root.join(".gitmodules");
+        let gitmodules_existed = gitmodules.exists();
+        let module_dir = git.absolute_git_dir()?.join("modules").join(relative_path);
+        let can_rollback = !path.exists()
+            && !module_dir.exists()
+            && !git.has_submodule_config(relative_path)?
```
mattsse: Any pre-existing `submodule.<path>.*` local config disables rollback entirely. I reproduced this from a Git-clean repository with `submodule.lib/solady.custom=keep`: an invalid-tag install leaves `.gitmodules`, `lib/solady`, `.git/modules/lib/solady`, and the newly added `url`/`active` config entries behind.

I suggest snapshotting the matching local config entries before `submodule add`, allowing rollback when the filesystem/index ownership checks pass, then restoring the snapshot after `submodule deinit`. The regression test should use a harmless custom key, assert the repository is clean and the worktree/module directory are gone, and verify that only the original config remains. The current `update=none` fixture causes `git submodule add` to fail before this rollback path is exercised.

mattsse (commented):
Found one cleanup gap when the target already has local submodule config.

crates/forge/src/cmd/install.rs
stevencartavia: [8896e64](https://github.com/foundry-rs/foundry/pull/15772/commits/8896e646ed8f7ced1bfc77d1eea8d60ccd4d5559)

crates/forge/src/cmd/install.rs
```diff
@@ -402,35 +407,103 @@ impl Installer<'_> {
...
+        let gitmodules_existed = gitmodules.exists();
+        let module_dir = git.absolute_git_dir()?.join("modules").join(relative_path);
+        let submodule_config = git.submodule_config(relative_path)?;
+        let can_rollback = !path.exists()
+            && !module_dir.exists()
+            && git.is_path_clean(relative_path)?
+            && git.is_path_clean(Path::new(".gitmodules"))?;
+
         // install the dep
         self.git_submodule(dep, path)?;
```
mablr, line 519: `git submodule add` itself needs to be inside the rollback boundary. It is not atomic: with an existing `.git/index.lock`, Git clones the dependency and creates `.git/modules/<path>` before failing to update the index. This `?` then returns immediately, leaving both directories behind.

crates/forge/src/cmd/install.rs
```diff
@@ -402,35 +407,103 @@ impl Installer<'_> {
...
+            if self.commit {
+                self.git.add(Some(path))?;
+            }
 
-        if self.commit {
-            self.git.add(Some(path))?;
+            Ok(dep.tag)
+        })();
+
+        if result.is_err() && can_rollback && actual_module_dir == module_dir {
```
mablr: A false `can_rollback` should prevent the mutation rather than only suppress cleanup afterward. I reproduced this with a tracked `.gitmodules` containing an unstaged edit: the invalid-tag install left `.gitmodules` staged/modified, `lib/solady` staged, the module gitdir, and new local config entries. We should either fail before `submodule add` when rollback is unsafe or snapshot enough state to restore it exactly; staged and unstaged `.gitmodules` cases would be useful regression tests.

crates/forge/src/cmd/install.rs
```diff
@@ -402,35 +407,103 @@ impl Installer<'_> {
     /// This will add the git submodule to the given dir, initialize it and checkout the tag if
     /// provided or try to find the latest semver, release tag.
     fn install_as_submodule(self, dep: &Dependency, path: &Path) -> Result<Option<String>> {
+        let root = Git::root_of(self.git.root)?;
+        let relative_path = path.strip_prefix(&root)?;
+        let git = self.git.root(&root);
+        let gitmodules = root.join(".gitmodules");
+        let gitmodules_existed = gitmodules.exists();
```
mablr: Existence alone is not enough to restore `.gitmodules`. A clean file can already contain an orphan section for this exact path. `git submodule add --force` reuses/rewrites that section, and the later `git rm` removes it entirely, deleting pre-existing user configuration. Maybe reject an existing mapping for the requested name/path or snapshot and restore the worktree and index state verbatim. The current test only covers an unrelated section, so a same-path fixture would catch this.

stevencartavia: [4940c19](https://github.com/foundry-rs/foundry/pull/15772/commits/4940c19a391cf75e0f7173886e0b6b21574740ed)

crates/forge/src/cmd/install.rs
stevencartavia: [4940c19](https://github.com/foundry-rs/foundry/pull/15772/commits/4940c19a391cf75e0f7173886e0b6b21574740ed)

crates/forge/src/cmd/install.rs
stevencartavia: [4940c19](https://github.com/foundry-rs/foundry/pull/15772/commits/4940c19a391cf75e0f7173886e0b6b21574740ed)

crates/forge/src/cmd/install.rs
```diff
@@ -128,10 +128,19 @@ impl DependencyInstallOpts {
         if !no_git {
             lockfile = lockfile.with_git(&git);
 
-            // Check if submodules are uninitialized, if so, we need to fetch all submodules
-            // This is to ensure that foundry.lock syncs successfully and doesn't error out, when
-            // looking for commits/tags in submodules
-            if git.submodules_uninitialized()? {
+            for dep in &dependencies {
+                let path = libs.join(dep.name());
+                if path.is_symlink() || (path.exists() && !path.join(".git").exists()) {
```
grandizzy: **[P1] Allow registered, uninitialized submodule worktrees**

A non-recursive clone creates each registered submodule path as an empty directory without a `.git` entry. Because initialization now only runs when `dependencies.is_empty()`, an explicit command such as `forge install vectorized/solady@<tag>` is rejected here before the existing submodule can be initialized.

Please defer this check until after resolving the `.gitmodules` mapping and gitlink. For a registered submodule, distinguish an empty uninitialized worktree from a symlink or non-empty foreign directory, then run `submodule update` for the former:

```rust
let uninitialized_worktree = existing_submodule
    && !path.is_symlink()
    && path.is_dir()
    && std::fs::read_dir(path)?.next().is_none();

if existing_submodule && path.exists() && !uninitialized_worktree {
    self.ensure_submodule_worktree(path, &module_dir, relative_path)?;
}

let setup = if existing_submodule {
    if !path.exists() || uninitialized_worktree {
        git.submodule_update(false, false, false, false, Some(relative_path))
    } else {
        Ok(())
    }
} else {
    self.git_submodule(dep, path)
};
```

An integration test that installs a dependency, deinitializes it, and then explicitly installs it again would cover both the fresh-clone and `git submodule deinit` cases.

crates/forge/src/cmd/install.rs
```diff
@@ -402,35 +418,183 @@ impl Installer<'_> {
...
-            dep.tag = self.last_tag(path);
+        let root = Git::root_of(self.git.root)?;
+        let relative_path = path.strip_prefix(&root)?;
+        let git = self.git.root(&root);
+        let gitmodules = root.join(".gitmodules");
+        let gitmodules_existed = gitmodules.exists();
+        let gitmodules_contents = gitmodules_existed.then(|| fs::read(&gitmodules)).transpose()?;
+        let module_dir = git.absolute_git_dir()?.join("modules").join(relative_path);
+        let submodule_config = git.submodule_config(relative_path)?;
+        let (submodule_name_collision, has_default_mapping, has_submodule_path) =
```
grandizzy: **[P2] Resolve existing submodules by path rather than assuming name equals path**

Valid Git submodules can use a section name different from their worktree path:

```ini
[submodule "solady"]
    path = lib/solady
    url = https://github.com/vectorized/solady
```

This currently produces `(false, false, true)` and is rejected as a collision, even when the mapping, gitlink, URL, and worktree are all valid. The corresponding gitdir is also `.git/modules/solady`, not `.git/modules/lib/solady`.

Please resolve the section whose `.path` equals `relative_path`, then use that section name for the URL, local config, and module gitdir checks. For example:

```rust
let mapping = git.submodule_mapping_for_path(relative_path)?;
let module_name = mapping.as_ref().map_or(relative_path, |mapping| mapping.name.as_ref());
let module_dir = git.absolute_git_dir()?.join("modules").join(module_name);
let registered_url = mapping.as_ref().and_then(|mapping| mapping.url.as_deref());
let existing_submodule = mapping.is_some() && git.is_gitlink(relative_path)?;
```

grandizzy (dismissed):
The rollback handling now covers the cleanup gaps raised in the earlier reviews, including partial `git submodule add` failures and preservation of staged/config state.

I’m requesting changes because the new safety checks regress explicit installs of registered but uninitialized submodules—a common state after a non-recursive clone or `git submodule deinit`. I also left a non-blocking comment about resolving valid custom-named submodules by path rather than assuming the section name equals the path.

crates/forge/src/cmd/install.rs
stevencartavia: Fixed. Existing submodules are now resolved by path and use their section name.

crates/forge/src/cmd/install.rs
stevencartavia: Fixed. Registered empty worktrees are now initialized, with regression coverage.

crates/forge/src/cmd/install.rs
```diff
@@ -402,35 +408,188 @@ impl Installer<'_> {
...
+            mapping.as_ref().map_or(relative_path, |mapping| Path::new(&mapping.name));
+        let module_dir = git.absolute_git_dir()?.join("modules").join(module_name);
+        let submodule_config = git.submodule_config(module_name)?;
+        if gitmodules_existed && !git.is_tracked(Path::new(".gitmodules"))? {
+            eyre::bail!(
+                "cannot safely install dependency at {} because the target or .gitmodules has existing changes",
+                relative_path.display()
+            );
         }
+        let existing_submodule = mapping.is_some() && git.is_gitlink(relative_path)?;
```
mablr: **Preserve pre-existing index entries.** `existing_submodule` requires both a mapping and a gitlink, so an indexed gitlink without a `.gitmodules` mapping is classified as newly owned. `is_path_clean` is not sufficient to establish index absence (for example, an `assume-unchanged` gitlink is reported clean). If checkout then fails, rollback reaches `git rm --force` and stages deletion of that pre-existing gitlink. I reproduced this locally. Please inspect the index directly before enabling rollback: reject an existing gitlink without a valid mapping, and reject any tracked entries at/below the target.

crates/forge/src/cmd/install.rs
```diff
@@ -204,6 +203,8 @@ impl DependencyInstallOpts {
                     {
                         // always work with relative paths when directly modifying submodules
                         git.set_submodule_branch(rel_path, tag_or_branch)?;
+                        let root = Git::root_of(git.root)?;
+                        git.root(&root).add(Some(".gitmodules"))?;
```
mablr: **Keep rollback ownership until the install command has succeeded.** The rollback transaction ends when `install_as_submodule` returns above, but this mutation and the subsequent dependency resolution, lockfile write, staging, and `git commit` can still fail. A failing commit hook under `--commit`, for example, returns an error while leaving the newly added submodule and metadata behind. Please return/retain a rollback guard and only commit it after the per-dependency metadata, lockfile, and optional commit operations succeed.

crates/forge/src/cmd/install.rs
```diff
@@ -402,35 +408,188 @@ impl Installer<'_> {
...
+
+        let actual_module_dir = if path.exists() {
+            self.git.root(path).absolute_git_dir().unwrap_or_else(|_| module_dir.clone())
+        } else {
+            module_dir.clone()
+        };
 
-        if self.commit {
-            self.git.add(Some(path))?;
+        if result.is_err() && !existing_submodule && actual_module_dir == module_dir {
```
mablr: **Restore the initialization state of existing submodules on failure.** An existing deinitialized submodule is initialized above before checkout. If the requested ref is invalid (or recursive update/sync fails), this condition excludes it from rollback, leaving a previously deinitialized submodule initialized with new local config. Please track whether this invocation initialized the worktree and deinitialize it on failure; failures after checkout may also need the original HEAD restored.

stevencartavia: [04c2fc9](https://github.com/foundry-rs/foundry/pull/15772/commits/04c2fc9f35c87b4b26093f18d2359741e8c0151f)

crates/forge/src/cmd/install.rs
stevencartavia: [04c2fc9](https://github.com/foundry-rs/foundry/pull/15772/commits/04c2fc9f35c87b4b26093f18d2359741e8c0151f)

crates/forge/src/cmd/install.rs
stevencartavia: [04c2fc9](https://github.com/foundry-rs/foundry/pull/15772/commits/04c2fc9f35c87b4b26093f18d2359741e8c0151f)

mablr (changes requested):
This has grown disproportionately broad for #6790. The PR now implements partial transaction handling for existing/deinitialized/custom-named submodules, lockfiles, commit hooks, config, and --no-git, but those expanded guarantees are still incomplete, for example, an existing submodule modified before a failed commit hook is not restored, and restoring only its SHA can detach a previously attached branch.

Could we narrow this to failed setup of a newly added submodule?

- Preflight that the target path, index entry, and module directory are absent and .gitmodules is safe to restore.
- Keep a guard around submodule add, checkout, recursive update, and sync.
- On failure, remove only the worktree/gitlink/module directory/config created by that attempt and restore .gitmodules.
- Retain the invalid-tag, collision, .gitmodules, and literal-pathspec regression tests.
- Split existing-submodule reinstall, commit-hook, custom mapping, and --no-git transaction behavior into follow-ups.

This would fix #6790 with a much smaller and clearer ownership boundary instead of continuing to patch edge cases in a partial general transaction system.

mablr (dismissed):
Much better like this, sgtm

crates/cli/src/utils/mod.rs
```diff
@@ -704,6 +724,7 @@ ignore them in the `.gitignore` file."
         path: impl AsRef<OsStr>,
     ) -> Result<()> {
         self.cmd()
+            .env("GIT_LITERAL_PATHSPECS", "1")
```
DaniPopes: why sometimes env and sometimes flag

crates/cli/src/utils/mod.rs
```diff
@@ -449,6 +449,10 @@ impl<'a> Git<'a> {
         self.cmd().arg("add").args(paths).exec().map(drop)
     }
 
+    pub fn add_literal(self, path: &Path) -> Result<()> {
+        self.cmd().args(["--literal-pathspecs", "add", "--"]).arg(path).exec().map(drop)
```
DaniPopes, line 453: can you explain why use this flag

stevencartavia: Dependency aliases can contain pathspec characters like `*`, so this prevents Git from treating the dependency path as a pattern and matching sibling entries.

crates/cli/src/utils/mod.rs
stevencartavia: [a501a22](https://github.com/foundry-rs/foundry/pull/15772/commits/a501a22276f258e45b68f217d440ef6db47d4fe6) updated for consistency

DaniPopes (approved):

mablr (approved):

Ari4ka (approved):

PR #15773, opened 2026-07-15
feat(bench): ingest results into ClickHouse

stevencartavia:
Adds trusted ClickHouse ingestion for branch-versus-master benchmark runs while preserving existing reports and common JSON outputs. Provisioning the destination table and credentials pending.

Closes OSS-477

### DIFF
--- a/.github/scripts/benchmark_clickhouse.py
+++ b/.github/scripts/benchmark_clickhouse.py
@@ -218,8 +218,15 @@ def build_rows(artifact_dir, trusted=None):
         raise ValueError("artifact came from an unexpected workflow")
     if run.get("event") != "workflow_dispatch":
         raise ValueError("artifact came from an unexpected event")
-    if run.get("run_id") != trusted["run_id"] or run.get("run_attempt") != trusted["run_attempt"]:
+    if run.get("run_id") != trusted["run_id"]:
         raise ValueError("artifact run identity does not match triggering workflow")
+    artifact_attempt = run.get("run_attempt")
+    if (
+        isinstance(artifact_attempt, bool)
+        or not isinstance(artifact_attempt, int)
+        or not 1 <= artifact_attempt <= trusted["run_attempt"]
+    ):
+        raise ValueError("artifact run attempt does not match triggering workflow")
 
     baseline_commit = _require_sha(comparison.get("baseline_commit"), "baseline commit")
     candidate_commit = _require_sha(comparison.get("candidate_commit"), "candidate commit")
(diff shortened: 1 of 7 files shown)


.github/scripts/benchmark_clickhouse.py
```diff
@@ -0,0 +1,470 @@
...
+        raise ValueError("unsupported benchmark run manifest")
+    run = manifest.get("run", {})
+    comparison = manifest.get("comparison", {})
+    if run.get("repository") != trusted["repository"]:
+        raise ValueError("artifact repository does not match triggering workflow")
+    if run.get("workflow_file") != trusted["workflow_file"]:
+        raise ValueError("artifact came from an unexpected workflow")
+    if run.get("event") != "workflow_dispatch":
+        raise ValueError("artifact came from an unexpected event")
+    if run.get("run_id") != trusted["run_id"] or run.get("run_attempt") != trusted["run_attempt"]:
```
grandizzy: Strict attempt equality breaks GitHub’s “Re-run failed jobs” flow. If `run-benchmarks` uploads the artifact in attempt 1 but `publish-results` fails, rerunning only the failed job completes as attempt 2 while retaining the attempt-1 artifact, so this rejects a valid result.

Could we require `1 <= artifact_attempt <= trusted_attempt`, then use the artifact attempt for `workflow_run_attempt` and `result_id`? A test with manifest attempt 1 and trusted attempt 2 would cover this recovery path.

grandizzy (changes requested):
The privileged `workflow_run` trust boundary looks sound, and I found no path for branch-controlled artifacts to access or redirect the ClickHouse credentials. I found one operational issue with partial workflow reruns: a valid artifact produced by an earlier attempt is rejected after only the failed downstream job is rerun. I’d fix the attempt handling and add the suggested regression test before merging.

.github/scripts/benchmark_clickhouse.py
stevencartavia: Addressed in f55b44804. Artifact attempts now require `1 <= artifact_attempt <= trusted_attempt` while the workflow run ID still matches exactly. Rows and `result_id` use the artifact attempt, so a failed-jobs rerun reusing an earlier artifact stays idempotent; a genuinely rerun benchmark remains a distinct attempt. The regression test covers both the attempt-1 artifact/attempt-2 workflow recovery path and rejection of an artifact from a future attempt.

grandizzy (approved):
lgtm

mablr (approved):

PR #15774, opened 2026-07-15
feat(doc): inherit natspec implicitly from base members

### LINKED ISSUE #4070: feat(docs): Support implicit inheritance for Natspec
### Component

Forge

For repositories that declare all Natspec at the interface level and inherit and override all functions, it would be nice to be able to parse the Natspec that corresponds to those functions without `@inheritdoc` syntax since the vast majority of the time the inheritance relationship is one-to-one. 

In the case of a function that overrides multiple inherited functions, `@inheritdoc` syntax can be enforced, or a contract can be selected dynamically.

### Additional context

_No response_

0xMars42:
## Motivation

Closes #4070. Repositories that declare all natspec at the interface level render with empty descriptions unless every override carries an explicit `@inheritdoc` tag: an undocumented `function doThing(uint256 value) external override` renders its parameter and return descriptions empty even when `IThing.doThing` documents everything.

## Solution

When a function or public state variable has no `@inheritdoc` tag, resolve its natspec from the nearest documented matching member in the contract's linearized bases, following Solidity's linearization order. Explicit `@inheritdoc` keeps taking precedence, and the overload disambiguation by parameter types applies unchanged: the implicit path reuses the same walker as the explicit resolver, extracted rather than duplicated.

One design choice worth calling out: the issue suggests requiring `@inheritdoc` when a function overrides members from several bases. This implementation instead inherits from the nearest documented one in linearization order, which is the order Solidity itself resolves inheritance in, and `@inheritdoc` stays available to pick a different base explicitly. Happy to make the ambiguous case stricter if preferred.

The regression test documents `deposit` on the interface only and asserts the override's page renders the notice, parameter and return descriptions. The pre-existing doc tests, including the explicit `@inheritdoc` ones, pass unchanged.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/doc/src/render.rs
```diff
@@ -347,29 +356,46 @@ fn render_contract<'ast, 'gcx>(
...
+                            hir_ext::resolve_inheritdoc(
+                                gcx,
+                                cid,
+                                fname,
+                                &base,
+                                param_types.as_deref(),
+                            )
+                        })
+                    })
+                    .or_else(|| {
```
mablr: An explicit `@inheritdoc` that fails to resolve currently falls through to implicit resolution, which can select documentation from a different base. For example, `@inheritdoc A` where `A.f` is undocumented and `B.f` is documented will render B's docs. Please branch on the presence of the tag so implicit lookup only runs when `@inheritdoc` is absent.

crates/doc/src/render.rs
```diff
@@ -347,29 +356,46 @@ fn render_contract<'ast, 'gcx>(
...
+                        hir_id.and_then(|cid| {
+                            hir_ext::resolve_implicit_inheritdoc(
```
mablr: Implicit inheritance needs to be gated on the declaration having no local NatSpec. As written, a local `@notice` still pulls missing `@dev`, `@param`, and `@return` tags from the base. Solidity only auto-inherits for functions without NatSpec (and also excludes renamed parameters); partial merging should remain exclusive to explicit `@inheritdoc`.

crates/doc/src/hir_ext.rs
```diff
@@ -290,7 +290,36 @@ pub fn resolve_inheritdoc(
...
+        contract.linearized_bases.iter().copied().filter(|&id| id != contract_id).collect();
+    if search_contracts.is_empty() {
+        return None;
+    }
+    search_function_doc(gcx, &search_contracts, fn_name, param_types)
+}
+
+/// Walks `search_contracts` in order and returns the first documented function matching
+/// `fn_name`, preferring an exact parameter signature match when overloads exist.
+fn search_function_doc(
```
mablr: Exact signature matching should apply whenever `param_types` is available, not only when this particular base declares multiple same-name functions. If a nearer base has only `f(address)` and a farther base documents `f(uint256)`, resolving an override of `f(uint256)` currently stops at the nearer wrong overload. Please require an exact signature at every level and continue upward when it does not match.

crates/doc/src/hir_ext.rs
```diff
@@ -374,7 +403,34 @@ pub fn resolve_inheritdoc_var(
...
+        contract.linearized_bases.iter().copied().filter(|&id| id != contract_id).collect();
+    if search_contracts.is_empty() {
+        return None;
+    }
+    search_variable_doc(gcx, &search_contracts, var_name)
+}
+
+/// Walks `search_contracts` in order and returns the first documented state variable or
+/// zero-parameter getter function matching `var_name`.
+fn search_variable_doc(
```
mablr: This limits public-variable inheritance to scalar zero-argument getters. A common implementation such as `mapping(address => uint256) public override balanceOf` has getter signature `balanceOf(address)` and cannot inherit the interface getter's NatSpec. Since the PR promises public state variables generally, please match through the generated HIR getter signature and render its input/return types, or narrow the stated scope to zero-argument scalar getters.

0xMars42 (commented):
Thanks, all four are addressed in 1448b76b6.

- Fall-through and partial-merge gating: implicit inheritance now runs only when the member has no `@inheritdoc` tag and no NatSpec of its own. A local `@notice` no longer pulls the base `@param`/`@return`, and an explicit `@inheritdoc` that fails to resolve no longer falls through to a different base. Partial merging stays exclusive to explicit `@inheritdoc`.
- Overload signature: a function is now matched by its exact parameter signature at every base level when the signature is known, continuing upward past a nearer base that declares a different same-name overload instead of stopping at it.
- Public variable getter: the variable is matched through its compiler-generated getter signature, so `mapping(address => uint256) public balanceOf` inherits the interface getter `balanceOf(address)`, instead of only scalar zero-argument getters.

Regression tests cover each: a documented override that inherits nothing, an override picking the far base's matching overload, and a mapping getter inheriting the interface docs.

crates/doc/src/hir_ext.rs
```diff
@@ -311,48 +344,42 @@ pub fn resolve_inheritdoc(
...
-                    && (!doc.notices.is_empty()
-                        || !doc.devs.is_empty()
-                        || !doc.params.is_empty()
-                        || !doc.returns.is_empty())
+        match param_types {
+            // With a known signature, accept only an exact match, at this level and every one
+            // above: continue upward when this base has no signature-exact documented match.
+            Some(want) => {
+                for &fid in &name_matches {
+                    if function_param_types(gcx, fid).as_deref() == Some(want)
```
stevencartavia: Routing explicit `@inheritdoc` through the new exact-signature search regresses non-ABI parameter types because their fallback signatures are compared as source strings. For example, a base using `mapping(uint => uint)` and an override using the equivalent `mapping(uint=>uint)` no longer inherit, while the base revision does. Since explicit `@inheritdoc` is intended to remain unchanged, could we compare semantic types or preserve the prior single-candidate behavior for the explicit path?

0xMars42: Good catch, thanks. Fixed by dropping the string comparison entirely: the resolver now compares compiler-resolved parameter types (`type_of_item` normalized through `as_externally_callable_function`, which also folds calldata into memory), so equivalent spellings, aliases and qualified names match, and overloads with non-ABI types are disambiguated by type instead of source text. The public variable getter path goes through the same comparison. Added your exact case as a regression test, an implicit-path twin, and a negative test asserting that a non-matching overload does not inherit.

crates/doc/src/hir_ext.rs
```diff
@@ -290,7 +290,40 @@ pub fn resolve_inheritdoc(
...
+    fn_name: &str,
+    target: Option<FunctionId>,
+) -> Option<InheritedDoc> {
+    let contract = gcx.hir.contract(contract_id);
+    let search_contracts: Vec<ContractId> =
+        contract.linearized_bases.iter().copied().filter(|&id| id != contract_id).collect();
+    if search_contracts.is_empty() {
+        return None;
+    }
+    search_function_doc(gcx, &search_contracts, fn_name, target)
```
mablr: Implicit inheritance must stop when more than one base function matches; Solidity requires explicit `@inheritdoc` for this ambiguous case. This currently picks the first documented base in linearization order.

crates/doc/src/hir_ext.rs
```diff
@@ -311,48 +344,46 @@ pub fn resolve_inheritdoc(
...
-                        || !doc.devs.is_empty()
-                        || !doc.params.is_empty()
-                        || !doc.returns.is_empty())
+        match target.and_then(|target| resolved_param_types(gcx, target)) {
+            // With a resolved signature, accept only a semantically equal match, at this level
+            // and every one above: continue upward when this base has no signature-exact
+            // documented match.
+            Some(want) => {
+                for &fid in &name_matches {
+                    if resolved_param_types(gcx, fid) == Some(want)
```
mablr: Resolved types are not sufficient for implicit inheritance: Solidity excludes overrides whose parameter names differ. Please require matching names on the implicit path; explicit `@inheritdoc` can keep the existing normalization.

crates/doc/src/hir_ext.rs
```diff
@@ -374,7 +405,56 @@ pub fn resolve_inheritdoc_var(
...
+    contract_id: ContractId,
+    var_name: &str,
+) -> Option<InheritedDoc> {
+    let contract = gcx.hir.contract(contract_id);
+    let search_contracts: Vec<ContractId> =
+        contract.linearized_bases.iter().copied().filter(|&id| id != contract_id).collect();
+    if search_contracts.is_empty() {
+        return None;
+    }
+    let getter = variable_getter(gcx, contract_id, var_name);
```
mablr: Please require a generated getter here. As written, private/internal variables can implicitly inherit docs from same-name base variables/functions; this feature should apply only to public state-variable getters.

crates/doc/src/hir_ext.rs
```diff
@@ -387,15 +467,26 @@ pub fn resolve_inheritdoc_var(
...
+                        Some(getter) => {
+                            resolved_param_types(gcx, fid) == resolved_param_types(gcx, getter)
+                        }
+                        // No getter (the variable is not public): only a parameterless base
+                        // function can document it.
+                        None => resolved_param_types(gcx, fid).is_some_and(|p| p.is_empty()),
+                    };
                     if f.name.map(|n| n.as_str() == var_name).unwrap_or(false)
-                        && function_param_types(gcx, fid).map(|p| p.is_empty()).unwrap_or(false)
+                        && signature_matches
```
mablr: Matching the getter inputs is only half the fix: rendering still drops inherited `@param` docs and reports the variable declaration type (for example, the mapping) instead of the getter return type. Please carry the getter signature through to rendering and snapshot `@param`/`@return` output.

crates/doc/src/render.rs
```diff
@@ -937,6 +951,24 @@ fn inheritdoc_base(docs: &DocComments<'_>) -> Option<String> {
...
+/// Whether the declaration carries any documentation tag of its own (`@notice`, `@dev`,
+/// `@param`, `@return`). Solidity only auto-inherits NatSpec for a member with none, so implicit
+/// inheritance is gated on this being false; explicit `@inheritdoc` still merges into a partial
+/// local doc.
+fn has_local_natspec(docs: &DocComments<'_>) -> bool {
+    docs.iter().any(|doc| {
+        doc.natspec.iter().any(|item| {
+            matches!(
+                item.kind,
+                NatSpecKind::Notice
```
mablr: A local `@custom:*` tag is NatSpec too, so it must suppress automatic inheritance. Please treat every local NatSpec item except `@inheritdoc` as local documentation.

crates/doc/src/hir_ext.rs
0xMars42: Fixed. Implicit resolution now takes the nearest matching declaration from each direct-base branch. Distinct surviving declarations make the inheritance ambiguous, while the same declaration reached through both sides of a pure diamond is counted once. The same rule is applied at each implicit level, so a leaf cannot bypass an ambiguity hidden behind an intermediate override.

Covered by `implicit_inheritance_resolves_base_ambiguity_per_branch`, which checks two distinct direct bases, an asymmetric diamond, and a shared declaration through a pure diamond.

crates/doc/src/hir_ext.rs
0xMars42: Fixed. Parameter names are compared against the single direct base at every implicit inheritance step, so an intermediate rename stops propagation even if a later override restores the original name. The explicit `@inheritdoc` path keeps its existing normalization.

Covered by `implicit_inheritance_requires_matching_param_names` with `Base(amount) -> Mid(shares) -> Leaf(amount)`.

crates/doc/src/hir_ext.rs
0xMars42: Fixed. The target variable must have a compiler-generated getter, and the automatic source must be an external declared base function matching that getter. A same-name base variable is no longer considered an automatic documentation source.

Covered by `implicit_inheritance_requires_public_getter_and_function_source`, which checks both a non-public target and a same-name base variable.

crates/doc/src/hir_ext.rs
0xMars42: Fixed. An inherited public variable now renders the ABI signature of its generated getter: one row per input and output, with the getter's names and types, and descriptions taken from the corresponding source function entries. Undocumented entries keep their own empty row instead of borrowing another one.

Covered by `inherited_getter_renders_param_and_return` with a two-key nested mapping and a two-field struct return, asserted row by row.

crates/doc/src/render.rs
0xMars42: Fixed. Any local NatSpec item is terminal for automatic inheritance at every level, including tags forge-doc does not render such as `@custom:*`. An intermediate custom-only override therefore cannot be skipped to reach older documentation.

Covered by `implicit_inheritance_skips_custom_tagged_members` on both the tagged intermediate contract and its leaf.

crates/doc/src/hir_ext.rs
```diff
@@ -290,67 +301,302 @@ pub fn resolve_inheritdoc(
...
+        })
+}
+
+/// Resolve automatic documentation level by level.
+///
+/// Any NatSpec on the current declaration is terminal. Renderable tags are returned as-is;
+/// tags that forge-doc does not render still stop the walk. Explicit inheritdoc resolution
+/// remains on the separate legacy path.
+fn implicit_function_doc(gcx: Gcx<'_>, fid: FunctionId) -> Option<InheritedDoc> {
+    if function_has_natspec(gcx, fid) {
```
stevencartavia: This treats the intermediate override’s `@inheritdoc` as terminal without resolving it. With documented `A.f`, `B.f` using `@inheritdoc A`, and undocumented `C.f`, C receives no documentation; Solidity propagates A’s resolved docs through B to C. Please resolve B’s merged documentation before returning here, and add a three-level regression test.

crates/doc/src/render.rs
```diff
@@ -347,29 +394,37 @@ fn render_contract<'ast, 'gcx>(
...
-                    .and_then(|fid| hir_ext::function_param_types(gcx, fid));
-                hir_id.and_then(|cid| {
-                    hir_ext::resolve_inheritdoc(gcx, cid, fname, &base, param_types.as_deref())
-                })
+                        hir_id.and_then(|cid| {
+                            hir_ext::resolve_inheritdoc(gcx, cid, fname, &base, hir_fn)
+                        })
+                    }
+                    None if !has_local_natspec(docs) => hir_id.and_then(|cid| {
+                        hir_ext::resolve_implicit_inheritdoc(gcx, cid, fname, hir_fn)
```
stevencartavia: Inherited return documentation is not remapped when the overriding function renames a return variable. If A documents `oldName` and B returns `newName`, Forge renders `oldName` as part of `newName`’s description, while Solidity maps the description positionally. Please remap inherited returns to the overriding function’s return slots and cover renamed returns with a regression test.

crates/doc/src/hir_ext.rs
0xMars42: Fixed. The implicit resolver now resolves an intermediate override's `@inheritdoc` and merges it into the result, so documentation propagates through it level by level. Covered by `implicit_inheritance_resolves_intermediate_inheritdoc` (A documented, B `@inheritdoc A`, C undocumented, and C inherits A's documentation through B).

crates/doc/src/render.rs
0xMars42: Fixed. Each inherited `@return` is resolved to the source function's declared return name at its position, so it maps positionally onto the overriding function's return slots and a renamed return receives the description on its own row instead of the base name being glued into the text. Covered by `implicit_inheritance_remaps_renamed_returns`.

crates/doc/src/render.rs
```diff
@@ -318,7 +323,14 @@ fn render_contract<'ast, 'gcx>(
                 }
                 write_comment_block(out, &c);
                 write_code_block(out, &ctx.dedented_snippet(*span));
-                if !c.returns.is_empty() {
+                // When the documentation is inherited from the variable's generated getter,
+                // render the getter's parameter and return signature instead of the declared
+                // (possibly mapping) type.
+                let getter_doc = inherited.as_ref().filter(|d| !d.getter_returns.is_empty());
+                if let Some(base_doc) = getter_doc {
+                    write_getter_table(out, "Parameters", &base_doc.getter_params);
```
mablr: The getter tables bypass the NatSpec sanitization applied just above. An inherited description containing `<amount>` or `{Reference}` is emitted as MDX syntax, and an unnamed getter field emits raw `<none>`, which MDX interprets as JSX. Please pass these descriptions through the same `replace_inline_links` sanitizer and render the absent-name placeholder as `&lt;none&gt;`. A full generated-output assertion with an unnamed return and MDX-sensitive text would cover both cases.

crates/doc/src/hir_ext.rs
```diff
@@ -413,6 +434,340 @@ pub fn resolve_inheritdoc_var(
...
+    out: &mut Vec<FunctionId>,
+) {
+    if !visited.insert(contract) {
+        return;
+    }
+    let mut declared = false;
+    for &item in gcx.hir.contract(contract).items {
+        if let ItemId::Function(fid) = item {
+            let f = gcx.hir.function(fid);
+            let eligible = !external_declared_only
```
mablr: This reconstructed override frontier currently accepts declarations that cannot be Solidity override bases. In particular, a same-signature private base function can donate docs to an unrelated child function, and a same-name Yul function in an intermediate contract can shadow the real Solidity ancestor. Please exclude Yul and private functions, exclude generated getters on the ordinary function/modifier path, and require the candidate to be in the same function-kind domain as the target.

crates/doc/src/render.rs
```diff
@@ -355,29 +367,32 @@ fn render_contract<'ast, 'gcx>(
         writeln!(out).unwrap();
         for (span, f, docs) in &functions {
             let fn_name = f.header.name.map(|n| n.as_str().to_string());
```
mablr: Fallback and receive declarations have no AST header name, so `fn_name` is `None` and this skips both explicit and implicit inheritance for them. The HIR resolver already represents these as `fallback`/`receive`; please derive the member name from `FunctionKind` here (while continuing to exclude constructors) and add fallback/receive coverage.

crates/doc/src/hir_ext.rs
```diff
@@ -413,6 +434,340 @@ pub fn resolve_inheritdoc_var(
...
+        }
+        return Some(local).filter(InheritedDoc::is_renderable);
+    }
+
+    let fn_name = function_member_name(gcx, fid)?;
+    let base = single_direct_base(gcx, fid, &fn_name, false)?;
+    if !param_names_match(gcx, fid, base) {
+        return None;
+    }
+    implicit_function_doc(gcx, base)
```
mablr: The recursive result is returned without remapping inherited return names to this intermediate callable. That still loses docs for a multi-hop getter chain where `IRoot.pair` returns `originalFirst`, an undocumented override renames it to `middleFirst`, and the final public-variable getter exposes `getterFirst`: `attach_getter_signature` searches for `middleFirst` in a doc still keyed by `originalFirst` and renders an empty description. Please remap return entries positionally at every inheritance hop, including after an explicit-`@inheritdoc` merge, rather than only when the current function has local NatSpec.

crates/doc/src/hir_ext.rs
```diff
@@ -413,6 +444,350 @@ pub fn resolve_inheritdoc_var(
...
+    if let Some(docs) = function_doc_comments(gcx, fid)
+        && docs.iter().any(|doc| !doc.natspec.is_empty())
+    {
+        let mut local = collect_inherited_doc(docs);
+        resolve_return_names(gcx, fid, &mut local);
+        if let Some(base_name) = inheritdoc_base_name(docs)
+            && let Some(cid) = gcx.hir.function(fid).contract
+            && let Some(fn_name) = function_member_name(gcx, fid)
+        {
+            let base_doc = resolve_inheritdoc_source(
```
mablr: This reuses a documentation-first resolver for an effective inheritance hop. That resolver only checks the signature when a level has multiple same-name functions and skips an exact declaration when it has no raw docs, so a failed `@inheritdoc` can select a wrong-signature or farther ancestor and then propagate here. Could we resolve the exact overridden callable first and treat its empty effective docs as terminal? A regression with a parameter rename at the intermediate base would pin this down.

crates/doc/src/hir_ext.rs
```diff
@@ -413,6 +444,350 @@ pub fn resolve_inheritdoc_var(
...
+                gcx,
+                cid,
+                &fn_name,
+                &base_name,
+                function_param_types(gcx, fid).as_deref(),
+            );
+            return match base_doc {
+                Some((source, mut base_doc)) => {
```
mablr: `base_doc` is the selected declaration’s raw comments, not its effective documentation. Across two partial `@inheritdoc` relays, fields inherited by the first relay are therefore lost at the second. Could this recursively obtain the source member’s effective doc before `merge_missing`, with a test preserving notice/dev/param/return across two hops?

crates/doc/src/hir_ext.rs
```diff
@@ -413,6 +444,350 @@ pub fn resolve_inheritdoc_var(
...
+/// The compiler-generated getter of the public variable `var_name` in `contract_id`, or
+/// `None` when it has none (not public).
+fn variable_getter(gcx: Gcx<'_>, contract_id: ContractId, var_name: &str) -> Option<FunctionId> {
+    for &item_id in gcx.hir.contract(contract_id).items {
+        if let ItemId::Variable(vid) = item_id {
+            let v = gcx.hir.variable(vid);
+            if v.name.map(|n| n.as_str() == var_name).unwrap_or(false) {
+                return v.getter;
+            }
+        }
```
mablr: The generated getter signature is used only for implicit inheritance. The explicit variable resolver still accepts only zero-argument functions, so adding `@inheritdoc` to a mapping or array getter disables otherwise-working inheritance. Could explicit and implicit variable resolution share this exact getter lookup?

crates/doc/src/hir_ext.rs
```diff
@@ -413,6 +444,350 @@ pub fn resolve_inheritdoc_var(
...
+    let mut declared = false;
+    for &item in gcx.hir.contract(contract).items {
+        if let ItemId::Function(fid) = item {
+            let f = gcx.hir.function(fid);
+            let eligible = !f.is_yul
+                && f.visibility != Visibility::Private
+                && f.kind == domain.kind
+                && f.gettee.is_none()
+                && (!domain.external_only || f.visibility == Visibility::External);
+            if eligible
```
mablr: This reconstructs an override edge from kind plus input signature, but does not check visibility, mutability, return compatibility, or the resolved override relationship. Since `forge doc` can process parseable sources without full semantic validation, incompatible declarations can inherit docs here. Can we reuse Solar’s base-function relation, or otherwise complete the compatibility predicate?

crates/doc/src/hir_ext.rs
```diff
@@ -413,6 +444,350 @@ pub fn resolve_inheritdoc_var(
...
+        .enumerate()
+        .map(|(index, &target)| {
+            let source_name = source.parameters.get(index).and_then(|&vid| var_name(vid));
+            let description = source_name
+                .as_ref()
+                .and_then(|name| doc.params.iter().find(|(candidate, _)| candidate == name))
+                .map(|(_, description)| description.clone())
+                .unwrap_or_default();
+            GetterField {
+                name: var_name(target).or(source_name),
```
mablr: Falling back to the interface parameter name means this table no longer reflects the generated getter ABI: an unnamed mapping key is rendered as named. Is that intentional presentation behavior? If so, please document/test the fallback explicitly; otherwise keep `None` so the renderer emits `&lt;none&gt;`. The same applies to returns below.

mablr (dismissed):

crates/doc/src/hir_ext.rs
```diff
@@ -252,219 +252,272 @@ pub fn inheritance_links(
...
+            NatSpecKind::Dev => doc.devs.push(content),
+            _ => {}
         }
     }
-
-    None
+    if let Some(fid) = callable_function(gcx, item) {
+        let function = gcx.hir.function(fid);
+        for index in 0..function.parameters.len() {
+            doc.params[index] = positional_description(gcx, view.items(), view.parameter(index));
```
stevencartavia, line 409: This PR regresses explicit `@inheritdoc` when a multiline `/// @param` on the inherited declaration is combined with a local `@notice` on the override: the generated docs drop the parameter’s continuation line. The same integration fixture preserves both lines on `master`, and an equivalent case retains the second line in solc’s NatSpec output.

mablr (approved):

mattsse (approved):

wyattloveslife-create (commented):
🍒

PR #15775, opened 2026-07-15
feat(lint): detect boundary tautology compositions

Osraka:
Related to #14381

## Summary

Extends the existing `type-based-tautology` lint to catch boolean `||` compositions that cover an entire integer type range, even when neither comparison is tautological in isolation.

The check is intentionally narrow:
- both comparisons must resolve to the same variable
- the compared constant must be the integer type minimum or maximum
- only boundary compositions such as `x > 0 || x == 0` are reported
- no new detector or CFG/data-flow analysis is introduced

## Tests

- `cargo test -p forge --test ui`
- `cargo test -p forge --test ui -- --bless`
- `cargo check -p forge-lint --all-targets`

The fixture covers unsigned and signed lower/upper boundaries, different-variable negatives, and signed interior values.


crates/lint/src/sol/med/tautology.rs
```diff
@@ -61,6 +71,100 @@ impl<'hir> LateLintPass<'hir> for TypeBasedTautology {
...
+            | BinOpKind::Gt
+            | BinOpKind::Ge
+            | BinOpKind::Eq
+            | BinOpKind::Ne
+    ) {
+        return None;
+    }
+
+    if let (Some((variable, ty)), Some((val_neg, val_mag))) =
+        (variable_type_of(hir, left), lit_value_of(right))
```
mablr: Boundary compositions over repeated explicit casts are missed: `uint8(x) > 0 || uint8(x) == 0` emits no warning, although it is always true. `comparison_of` only accepts `variable_type_of`, while the existing lint supports explicit casts. Please preserve a stable operand identity while using the cast type as the effective range.

crates/lint/src/sol/med/tautology.rs
```diff
@@ -61,6 +71,100 @@ impl<'hir> LateLintPass<'hir> for TypeBasedTautology {
...
+            && is_lower_point(left, lower))
+        // Values below the maximum plus the maximum itself cover the whole range.
+        || (matches_comparison(left, BinOpKind::Lt, upper)
+            && is_upper_point(right, upper))
+        || (matches_comparison(right, BinOpKind::Lt, upper)
+            && is_upper_point(left, upper))
+}
+
+fn matches_comparison(comparison: Comparison, op: BinOpKind, value: (bool, U256)) -> bool {
+    comparison.op == op && comparison.val_neg == value.0 && comparison.val_mag == value.1
```
mablr, line 156: Please normalize zero's sign before matching boundaries. For unsigned `x`, `x > -0 || x == -0` is always true but emits no warning because `-0` is represented as `(true, 0)` while the lower bound is `(false, 0)`.

Osraka:
Addressed in commit 5d7f917.

- Boundary composition matching now preserves the stable variable identity through repeated integer casts while using the cast type as the effective range.
- Signed zero is normalized before boundary matching, so `-0` is treated like `0`.
- Comparisons with different effective ranges are excluded to avoid false positives.
- Added UI regressions for explicit casts, `-0`, and mismatched ranges.

The full lint UI suite and `forge-lint` tests pass locally.

crates/lint/src/sol/med/tautology.rs
```diff
@@ -144,24 +248,67 @@ fn elem_type_of<'hir>(
...
+            if !matches!(ty, ElementaryType::Int(_) | ElementaryType::UInt(_)) {
+                return None;
+            }
+
+            let mut exprs = args.exprs();
+            let inner = exprs.next()?;
+            if exprs.next().is_some() {
+                return None;
+            }
+            let (variable, _) = comparison_operand_of(hir, inner)?;
```
mablr: Discarding the inner cast path causes false positives. For `uint256(uint8(raw)) > 0 || raw == 0`, `raw = 256` makes both branches false, but both operands normalize to `(raw, uint256)` and emit a warning. Please preserve/compare value-changing cast paths (or reject nested casts) and add this as an `// ok` fixture.

Osraka:
Addressed the latest review feedback in commit bc1437d.

The comparison operand path now rejects nested casts instead of treating the outer cast as a comparison over the original variable. This avoids false positives such as:

```solidity
uint256(uint8(raw)) > 0 || raw == 0
```

where the inner value-changing cast can make the two branches differ.

Direct explicit-cast comparisons remain supported, and the regression fixture now covers this nested-cast case as an `// ok` example.

Validation:
- `cargo test -p forge --test ui -- TypeBasedTautology --bless`
- `cargo test -p forge --test ui`
- `cargo test -p forge-lint`
- `cargo fmt --all --check`
- `git diff --check`

All passed locally.

crates/lint/src/sol/med/tautology.rs
```diff
@@ -61,6 +71,100 @@ impl<'hir> LateLintPass<'hir> for TypeBasedTautology {
...
+
+/// Returns true for boundary comparisons whose union covers the complete integer type range.
+fn is_boundary_composition(left: Comparison, right: Comparison) -> bool {
+    if left.variable != right.variable || left.ty != right.ty {
+        return false;
+    }
+
+    let Some((lower, upper)) = integer_bounds(left.ty) else { return false };
+
+    // Values greater than the minimum plus the minimum itself cover the whole range.
```
mablr, line 142: `is_boundary_composition` only handles a strict boundary comparison paired with the point at that same boundary. It misses opposite-bound unions: for `uint8 x`, `x > 0 || x < 255` is true for every value. Please cover these cross-boundary compositions and retain this case as a regression.

crates/lint/src/sol/med/tautology.rs
```diff
@@ -144,24 +248,82 @@ fn elem_type_of<'hir>(
...
+
+            let mut exprs = args.exprs();
+            let inner = exprs.next()?;
+            if exprs.next().is_some() {
+                return None;
+            }
+            // Reject nested casts: an inner value-changing cast can make this expression
+            // differ from a comparison over the original variable.
+            let (variable, _) = plain_variable_type_of(hir, inner)?;
+            return Some((variable, *ty));
```
mablr: Using the cast target as the effective range misses widening casts whose reachable range remains bounded by the source. For `uint8 x`, `uint16(x) < 255 || uint16(x) == 255` is always true. Please preserve the narrower source range for value-preserving widening casts and add this regression.

Osraka:
Addressed the two latest review comments in commit 5d44cbd.

- `is_boundary_composition` now recognizes opposite-bound strict unions such as `x > 0 || x < 255`.
- Comparisons now carry their reachable integer range rather than only the cast target type. For value-preserving same-signed widening casts, the narrower source range is retained, so `uint16(x) < 255 || uint16(x) == 255` is recognized when `x` is `uint8`.
- The existing range-equality guard remains in place, so mixed expressions such as `uint8(raw) > 0 || raw == 0` are not reported.
- Nested value-changing casts continue to be rejected as requested in the previous review.

Added regression coverage for both reported cases and kept the change scoped to the type-based tautology detector.

Validation:
- `cargo test -p forge --test ui -- TypeBasedTautology --bless`
- `cargo test -p forge --test ui` (`88 passed`)
- `cargo test -p forge-lint` (`11 passed`)
- `cargo fmt --all --check`
- `git diff --check`

All available checks pass locally. `cargo clippy` could not be run because the `cargo-clippy` component is not installed in the local toolchain.

crates/lint/src/sol/med/tautology.rs
```diff
@@ -144,24 +256,105 @@ fn elem_type_of<'hir>(
...
+            };
+            if !matches!(ty, ElementaryType::Int(_) | ElementaryType::UInt(_)) {
+                return None;
+            }
+
+            let mut exprs = args.exprs();
+            let inner = exprs.next()?;
+            if exprs.next().is_some() {
+                return None;
+            }
```
mablr, line 301: Rejecting every nested cast also creates a false negative when both branches use the same cast path. For example, `uint256(uint8(raw)) > 0 || uint256(uint8(raw)) == 0` is always true, but the UI fixture confirms that no diagnostic is emitted. Please retain a comparable cast-path identity so identical nested conversions can still be matched while different paths remain excluded.

Osraka:
Addressed the latest nested-cast review in commit 4f352bc.

Boundary comparisons now carry the resolved variable, ordered explicit-cast path, and effective reachable range. Identical nested paths such as `uint256(uint8(raw))` are matched, while different paths such as `uint256(uint8(raw))` vs `raw` remain excluded.

Added a regression fixture for the identical nested path and retained the negative different-path case.

Validation:
- cargo test -p forge --test ui -- TypeBasedTautology --bless (88 passed)
- cargo test -p forge-lint (11 passed)
- cargo fmt --all -- --check
- git diff --check

mablr (dismissed):
Thanks, LGTM.

crates/lint/src/sol/med/tautology.rs
```diff
@@ -61,6 +71,119 @@ impl<'hir> LateLintPass<'hir> for TypeBasedTautology {
...
+        });
+    }
+
+    None
+}
+
+/// Returns true for boundary comparisons whose union covers the complete integer type range.
+fn is_boundary_composition(left: &Comparison, right: &Comparison) -> bool {
+    if left.variable != right.variable
+        || left.cast_path != right.cast_path
```
stevencartavia, line 133: Requiring identical `cast_path`s misses mixed cast/uncast comparisons when the differing cast step is value-preserving. For example, with `uint8 amount`, `uint256(amount) > 0 || amount == 0` is always true, but its paths differ, so it emits no warning; casting both operands does emit one. Could we omit only individually proven value-preserving widening steps from operand identity and add a regression test? Value-changing steps should remain so `uint256(uint8(raw)) > 0 || raw == 0` stays unflagged.

Osraka:
Follow-up for the post-approval CI failure: clippy reported `clippy::missing-const-for-fn` for `is_value_preserving_widening` in `crates/lint/src/sol/med/tautology.rs`.

Fixed in commit `0b4c65c` by marking that pure helper `const`; no detector semantics or fixtures were changed. The maintainer merge commit was preserved before pushing.

Validation:
- `cargo test -p forge --test ui -- TypeBasedTautology`: 90 passed
- `cargo test -p forge-lint`: 11 passed; doctest ignored as expected
- `cargo fmt --all -- --check`
- `git diff --check`

Osraka:
Addressed the mixed cast/uncast operand-identity case in commit 0f4f83a.

Same-signed value-preserving widening casts are now omitted from the operand identity, so equivalent expressions such as uint256(amount) > 0 || amount == 0 are matched. Value-changing casts remain part of the identity, so cases such as uint256(uint8(raw)) > 0 || raw == 0 stay unflagged when the effective ranges differ.

Added a regression fixture for the mixed-path case.

Validation:
- cargo test -p forge --test ui -- TypeBasedTautology
- cargo test -p forge-lint
- cargo fmt --all -- --check
- git diff --check

stevencartavia (approved):

mablr (approved):

stevencartavia (approved):

PR #15778, opened 2026-07-16
fix(coverage): anchor else branches

### LINKED ISSUE #10792: bug(forge): coverage report shows `if/else` branch as uncovered when reading `bytes` from storage
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.0.0-stable

foundryup: 1.0.1

forge coverage

### Operating System

macOS (Apple Silicon)

### Describe the bug

Hey  👋

I noticed that there are a couple of improvements for `if/else` branch handling in coverage, but I’ve encountered a different edge case that still behaves unexpectedly.

The issue seems to be, when reading `bytes` from **storage** and assigning it to a `memory` variable inside an `if/else` block, **one of the branches is incorrectly marked as uncovered**, even though tests execute both branches.
If I replace the storage read with a constant or inline literal, both branches are marked as covered correctly.

For example, this code : 

```solidity
contract Test {
    bytes public optA = bytes("optA");
    bytes public optB = bytes("optB");

    bytes public constant OPT_A = bytes("optA");
    bytes public constant OPT_B = bytes("optB");

    function fun(uint16 _t) public view {
        bytes memory element;

        if (_t == 4) {
            // ✅ works
            // element = bytes("asset opt");
            // element = OPT_A;

            // ❌ doesn't work 
            element = optA;
        } else {
            // ✅ works
            // element = bytes("opt");
            // element = OPT_B;

            // ❌ doesn't work
            element = optB;
        }
    }
}
```

Let me know if you need a full test contract or repo. Happy to provide one or isolate this further if needed!

stevencartavia:
Fixes #10792 by anchoring non-empty `else` branches to their first executable opcode. This prevents compiler-generated jumps from storage `bytes` reads from causing incorrect branch coverage while preserving existing reporting locations and empty-branch behavior.

### DIFF
--- a/crates/evm/coverage/src/analysis.rs
+++ b/crates/evm/coverage/src/analysis.rs
@@ -128,6 +128,7 @@ impl<'gcx> SourceVisitor<'gcx> {
                 lines.push(CoverageItem {
                     kind: CoverageItemKind::Line,
                     loc: reference_item.loc.clone(),
+                    anchor_loc: None,
                     hits: 0,
                 });
             }
@@ -141,13 +142,15 @@ impl<'gcx> SourceVisitor<'gcx> {
 
     /// Creates a coverage item for a given kind and source location. Pushes item to the internal
     /// collection (plus additional coverage line if item is a statement).
-    fn push_item_kind(&mut self, kind: CoverageItemKind, span: Span) {
-        let item = CoverageItem { kind, loc: self.source_location_for(span), hits: 0 };
+    fn push_item_kind(&mut self, kind: CoverageItemKind, span: Span) -> &mut CoverageItem {
+        let item =
+            CoverageItem { kind, loc: self.source_location_for(span), anchor_loc: None, hits: 0 };
 
         debug_assert!(!matches!(item.kind, CoverageItemKind::Line));
         self.all_lines.push(item.loc.lines.start);
 
         self.items.push(item);
+        self.items.last_mut().unwrap()
     }
 
     fn source_location_for(&self, mut span: Span) -> SourceLocation {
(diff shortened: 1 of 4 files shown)


crates/evm/coverage/src/analysis.rs
```diff
@@ -259,18 +262,15 @@ impl<'ast> ast::Visit<'ast> for SourceVisitor<'_> {
                         CoverageItemKind::Branch { branch_id, path_id: 0, is_first_opcode: true },
                         then_stmt.span,
                     );
-                    if else_stmt.is_some() {
-                        // We use `stmt.span`, which includes `else_stmt.span`, since we need to
-                        // include the condition so that this can be marked as covered.
-                        // Initially implemented in https://github.com/foundry-rs/foundry/pull/3094.
+                    if let Some(else_stmt) = else_stmt {
+                        let is_first_opcode = stmt_has_statements(else_stmt);
```
grandizzy, line 266: `stmt_has_statements` does not guarantee that the body emits an opcode, so `find_anchor_simple` can drop the false-path anchor entirely. Could we test a syntactically non-empty but code-free `else` and fall back to `find_anchor_branch` when no body anchor is found?

```solidity
function execute(bool condition) external {
    if (condition) {
        value = 1;
    } else {
        uint256 unused;
    }
}
```

grandizzy (changes requested):
The reporting/anchor span separation looks like the right approach and fixes the storage `bytes` case. I found one regression for syntactically non-empty `else` bodies that emit no opcode: the new simple-anchor lookup fails, making full branch coverage unattainable. I’d preserve the existing branch-target lookup as a fallback and add the suggested regression test.

crates/evm/coverage/src/analysis.rs
stevencartavia: [a52b4c3](https://github.com/foundry-rs/foundry/pull/15778/commits/a52b4c38a8c248ba099645140af33c8d478d993c)

grandizzy (approved):
lgtm

mablr (approved):

PR #15781, opened 2026-07-16
fix(script): reject mismatched sequence caches

pengqima:
## Motivation

Script sequence restoration assumed that the public deployment file and sensitive cache always contained the same number of transactions. Multi-chain restoration made the same assumption about deployments. Partially written, corrupted, or manually edited cache files could therefore trigger an out-of-bounds panic.

## Solution

This validates both counts before restoring sensitive RPC data and uses zipped iteration after validation. Mismatched cache files now produce a descriptive error without partially updating the sequence.

Partially addresses #15779.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/script-sequence/src/sequence.rs
+++ b/crates/script-sequence/src/sequence.rs
@@ -104,7 +104,7 @@ impl<N: Network> ScriptSequence<N> {
         )
         .wrap_err(format!("Deployment's sensitive details not found for chain `{chain_id}`."))?;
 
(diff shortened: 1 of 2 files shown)


mablr (commented):
Thank you! But as mentioned in https://github.com/foundry-rs/foundry/issues/15779#issuecomment-4990256496, I think we can keep as-is.

PR #15782, opened 2026-07-16
feat(cast): accept mktx signatures

### LINKED ISSUE #10167: Feature request: Add field in `cast mktx --raw-unsigned` to accept signatures to build an ethereum transaction
### Component

Cast

for reference, this is what i wish to achieve using foundry tools
https://medium.com/@LucasJennings/a-step-by-step-guide-to-generating-raw-ethereum-transactions-c3292ad36ab4

Thanks to the newly introduced `--raw-unsigned` flag, I am able to create a raw RLP transaction, hash it and sign it using `cast wallet sign`

what is left is to be able to use the signature to create a new RLP transaction, which now includes the created signature using `cast mktx` to create a raw ethereum transaction

As per the medium article reference, Step 4 is to be tackled.

Hope this was enough information, let me know if this is feasable
Thanks

### Additional context

_No response_

mattsse:
Adds `--signature` to `cast mktx` so an externally produced 65-byte signature can be attached to a fully resolved transaction and encoded as a signed EIP-2718 envelope. The command recovers the signer from the final signing hash and rejects an explicit `--from` that does not match, while preserving existing fee-token and sponsor resolution.

Closes #10167.

This PR was written with Codex, which generated the implementation and tests under my direction.

### DIFF
--- a/crates/cast/src/cmd/mktx.rs
+++ b/crates/cast/src/cmd/mktx.rs
@@ -69,7 +69,12 @@ pub struct MakeTxArgs {
     ethsign: bool,
 
     /// Generate a raw signed transaction using the provided 65-byte signature.
-    #[arg(long, value_name = "SIGNATURE", conflicts_with_all = ["raw_unsigned", "ethsign"])]
+    #[arg(
+        long,
+        value_name = "SIGNATURE",
+        requires = "from",
+        conflicts_with_all = ["raw_unsigned", "ethsign"]
+    )]
     signature: Option<Signature>,
 }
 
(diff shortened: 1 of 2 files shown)


mablr (dismissed):
LGTM

crates/cast/src/cmd/mktx.rs
```diff
@@ -67,6 +67,10 @@ pub struct MakeTxArgs {
     /// Call `eth_signTransaction` using the `--from` argument or $ETH_FROM as sender
     #[arg(long, requires = "from", conflicts_with = "raw_unsigned")]
     ethsign: bool,
+
+    /// Generate a raw signed transaction using the provided 65-byte signature.
+    #[arg(long, value_name = "SIGNATURE", conflicts_with_all = ["raw_unsigned", "ethsign"])]
```
grandizzy: `--from` needs to be required here. ECDSA recovery against a changed digest generally succeeds by recovering a different address, so without an expected sender, RPC-resolved fee/gas drift—or a changed transaction argument—silently produces an envelope from the wrong account. I reproduced this by changing only `--value`: the command succeeded but the recovered signer changed. Requiring `--from` makes the existing mismatch check meaningful for every invocation. The now-unreachable nonce/sponsor guards below can then be removed.

```suggestion
    #[arg(
        long,
        value_name = "SIGNATURE",
        requires = "from",
        conflicts_with_all = ["raw_unsigned", "ethsign"]
    )]
```

crates/cast/src/cmd/mktx.rs
```diff
@@ -216,6 +230,57 @@ impl MakeTxArgs {
             return Ok(());
         }
 
+        if let Some(signature) = signature {
```
grandizzy, line 238: Recovery normalizes high-s signatures internally, but the original high-s signature is later passed to `into_signed`, so the signer check passes while the emitted envelope remains non-canonical under EIP-2. Normalize once before both recovery and encoding, and please add a high-s regression test asserting the canonical output.

```suggestion
        if let Some(signature) = signature {
            let signature = signature.normalized_s();
```

grandizzy (dismissed):
Requesting changes for two correctness issues in the external-signature path. Without `--from`, transaction-field drift can silently change the recovered sender; high-s signatures are also validated after normalization but encoded unchanged, producing a non-canonical envelope. Please see the inline suggestions.

mablr (approved):
Should be G2G now

stevencartavia (approved):

PR #15783, opened 2026-07-16
feat(script): add sender nonce override

### LINKED ISSUE #7248: Add `--contract-sender` flag to forge script for correct nonce handling w/ linked libraries
### Component

Forge

## Context

Forge scripts provide a convenient way to collect transaction data that can be executed through a multisig wallet like a Gnosis Safe. We collect transactions in this manner at [Sphinx](https://github.com/sphinx-labs/sphinx). Specifically, here's how the (non-standard) way that we use Forge scripts:
1. The user defines their Forge script. In the script, the user's Gnosis Safe is impersonated, i.e. the sender of the user's transactions is the Safe.
2. We dry run the script to collect the user's transactions. It's possible to collect the transactions through the dry run file written to the broadcast folder, or by using the state diff cheatcode.
3. After the multisig owners have approved the transaction data, an EOA executes the transactions on-chain. Specifically, the EOA calls the Gnosis Safe, which executes the user's transactions. Crucially, the sender of all the collected transactions is the Gnosis Safe. (I.e. the `msg.sender` in the user's transactions is the Gnosis Safe, not the EOA that initiated the transaction).

If you're wondering why we execute transactions through a Gnosis Safe at all, it's to provide features like gas abstraction and secure deployments/upgrades.

## Problem

Collecting transactions in this manner works well for contract deployments and standard transactions that are defined in the Forge script. However, if the user's multisig wallet hasn't been deployed yet, Foundry uses an incorrect sender nonce to deploy linked libraries, which causes the addresses of the libraries to be incorrect.

Specifically, here's why the issue happens:
1. Foundry fetches the nonce of the Gnosis Safe (i.e. the sender) when `forge script` is called. Since the Gnosis Safe hasn't been deployed yet, the initial nonce at its address is 0. This means the address of the first library deployed in the script will be calculated with a sender nonce of 0.
2. When the Gnosis Safe is deployed on-chain, its initial nonce is 1. (The EVM requires that all contracts have an initial nonce of 1 as of [EIP-161](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-161.md)). This means the address of the first library deployed on-chain will be calculated with a sender nonce of 1 instead of 0, resulting in a different library address. This is dangerous because the incorrect library address (i.e. the address calculated w/ a sender nonce of 0) is already injected into the initcode of any contract that depends on the linked library.

This issue doesn't occur if the user's multisig is already deployed because Foundry fetches the correct Gnosis Safe nonce, resulting in consistent library addresses.

### Preferred Solution

@RPate97 and I think the most straightforward solution is to add a new `--contract-sender` flag to `forge script`. This flag would simply set the initial nonce of the sender to 1 if its nonce on-chain is currently 0. This flag should also be available in `forge test` so that the user can get the same library addresses in their tests.

We acknowledge that this flag only makes sense in the context of collecting transactions to be executed later through a multisig wallet. This flag doesn't make sense when broadcasting transactions onto a live network via `forge script` because a smart contract can't sign a transaction.

We see this flag as a patch that satisfies our use case with minimal effort. In the future, it may make more sense to have some sort of `forge collect` command built specifically for this purpose.

### Alternative Solution 1

An alternative solution is to calculate the correct library addresses outside of the Forge script, then deploy these libraries manually inside of the script. However, this would require us to re-implement Foundry's [logic to resolve library dependencies](https://github.com/foundry-rs/foundry/pull/7027), which is non-trivial and would lead to quite a bit of redundancy between our systems. It makes a lot more sense to leverage Foundry's existing logic instead of re-implementing it.

### Alternative Solution 2

Another solution is to deploy the multisig in advance so that Foundry recognizes that its nonce is 1. This isn't an acceptable solution because the user should be able to fully simulate their script before any transactions are sent on-chain, which is one of Forge script's most compelling features.

### Additional context

_No response_

mattsse:
Adds `--sender-nonce` to `forge script` so callers can override the sender nonce used for library linking, script execution, and generated transactions. The override remains authoritative if Forge discovers a different broadcast sender and reruns linking, preventing RPC nonce lookup from silently changing the requested deployment addresses.

Closes #7248.

This PR was written with Codex, which generated the implementation and tests under my direction.

### DIFF
--- a/crates/script/src/lib.rs
+++ b/crates/script/src/lib.rs
@@ -166,6 +166,10 @@ pub struct ScriptArgs {
     #[arg(long, short, default_value = "130")]
     pub gas_estimate_multiplier: u64,
 
+    /// Override the sender's initial nonce for script execution and transaction generation.
+    #[arg(long, value_name = "NONCE")]
+    pub sender_nonce: Option<u64>,
+
     /// Send via `eth_sendTransaction` using the `--sender` argument as sender.
     #[arg(
         long,
@@ -321,7 +325,8 @@ impl ScriptArgs {
 
         tempo.resolve_expires();
 
-        let script_config = ScriptConfig::new(config, evm_opts, args.batch, tempo).await?;
+        let script_config =
+            ScriptConfig::new(config, evm_opts, args.batch, tempo, args.sender_nonce).await?;
         Ok(PreprocessedState { args, script_config, script_wallets, browser_wallet })
     }
 
@@ -778,6 +783,7 @@ pub struct ScriptConfig<FEN: FoundryEvmNetwork> {
     pub config: Config,
     pub evm_opts: EvmOpts,
     pub sender_nonce: u64,
+    sender_nonce_override: Option<u64>,
     /// Maps a rpc url to a backend
     pub backends: HashMap<String, Backend<FEN>>,
     /// Whether to batch all broadcast transactions into a single Tempo batch transaction.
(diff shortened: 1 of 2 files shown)


crates/script/src/lib.rs
```diff
@@ -778,6 +783,7 @@ pub struct ScriptConfig<FEN: FoundryEvmNetwork> {
     pub config: Config,
     pub evm_opts: EvmOpts,
     pub sender_nonce: u64,
+    sender_nonce_override: Option<u64>,
```
grandizzy, line 786: **[P1] Apply the override to on-chain simulation runners**

`PreSimulationState::build_runners` creates fresh runners from RPC state, so the sender starts at the RPC nonce instead of this override. With RPC nonce `0` and `--sender-nonce 7`, linking generates addresses from nonces `7+`, but simulation deploys at nonces `0+`; a dependent deployment then fails against the expected library address.

Please seed every runner with the override. The smallest fix is in `_get_runner`:

```rust
let mut runner = ScriptRunner::new(
    builder.build(evm_env, tx_env, db),
    self.evm_opts.clone(),
)
.with_debug_bytecodes(debug);

if self.sender_nonce_override.is_some() {
    runner.executor.set_nonce(self.evm_opts.sender, self.sender_nonce)?;
}

Ok(runner)
```

Please also make the regression test generate and simulate transactions using a linked library with the default CREATE2 deployer disabled; the current view-only test never reaches this path.

grandizzy (changes requested):
Request changes: the override is preserved through linking and script execution, but it is not applied to the EVM used for on-chain simulation. This breaks the linked-library deployment case when the RPC nonce differs from `--sender-nonce`. See inline finding.

mablr (approved):
sgtm

grandizzy (approved):
lgtm

crates/script/src/lib.rs
```diff
@@ -166,6 +166,10 @@ pub struct ScriptArgs {
     #[arg(long, short, default_value = "130")]
     pub gas_estimate_multiplier: u64,
 
+    /// Override the sender's initial nonce for script execution and transaction generation.
+    #[arg(long, value_name = "NONCE")]
```
figtracer, line 170: `u64::MAX` is accepted here, but a linked script panics at `runner.rs:147` with `nonce overflow in transaction`. can we reject that value during parsing so invalid input returns a normal error?

PR #15784, opened 2026-07-16
fix(anvil): handle `eth_simulateV1` calls without `to` as contract creation

mablr:
## Motivation

Close OSS-562.

This treats `eth_simulateV1` calls without `to` as contract creation, allowing empty and from-only calls to produce returned transactions. Successful creation output is preserved, while reverts expose their bytes through the simulation error response.

Omitted-nonce progression remains tracked by OSS-561. Canonical returned transaction metadata, including chain ID, and exact revert-message conformance remain tracked by OSS-565.


mattsse (approved):
lgtm

figtracer (approved):

PR #15785, opened 2026-07-16
fix(script): reject max sender nonce

figtracer:
The `--sender-nonce` flag currently accepts `u64::MAX`, but every transaction increments the sender nonce before execution. As a result, linked-library scripts can reach the script runner and panic with `nonce overflow in transaction` rather than reporting invalid CLI input. This was identified in [review feedback on #15783](https://github.com/foundry-rs/foundry/pull/15783#discussion_r3594919492).

Restrict the parser to nonces below `u64::MAX` and cover the upper boundary, preserving `u64::MAX - 1` while returning a Clap value-validation error for `u64::MAX`.

This PR was written with Codex, which analyzed the failure path and generated the implementation and tests under my direction.

### DIFF
--- a/crates/script/src/lib.rs
+++ b/crates/script/src/lib.rs
@@ -22,7 +22,7 @@ use alloy_primitives::{
 use alloy_signer::Signer;
 use broadcast::next_nonce;
 use build::PreprocessedState;
-use clap::{Parser, ValueHint};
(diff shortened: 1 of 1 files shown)


mattsse (approved):
lgtm

mablr (approved):

PR #15788, opened 2026-07-16
feat(debugger): toggle data pane

figtracer:
#927 tracks pane visibility controls, but the debugger's shared data view was still always rendered even when the current memory, calldata, returndata, or storage view was not useful. This adds a `:data` command through the existing pane-command path and lets both layouts reclaim the hidden pane's space, including source-only horizontal layouts.

Progresses #927.

### DIFF
--- a/crates/debugger/src/tui/draw.rs
+++ b/crates/debugger/src/tui/draw.rs
@@ -25,9 +25,9 @@ use std::{collections::VecDeque, fmt::Write};
 
 impl TUIContext<'_> {
     pub(crate) fn draw_layout(&mut self, f: &mut Frame<'_>) {
-        // We need 100 columns to display a 32 byte word in the memory and stack panes.
+        // We need 100 columns to display a 32 byte word in the data and stack panes.
         let area = f.area();
-        let min_width = 100;
+        let min_width = if self.show_data || self.show_stack { 100 } else { 1 };
         let min_height = 16;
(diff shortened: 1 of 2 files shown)


crates/debugger/src/tui/draw.rs
```diff
@@ -112,9 +112,9 @@ impl TUIContext<'_> {
 
         let opcodes_weight = if self.show_opcodes { 1 } else { 0 };
         let source_weight = if self.show_source { 3 } else { 0 };
-        let memory_weight = if self.show_source { 1 } else { 2 };
+        let data_weight = if self.show_data { if self.show_source { 1 } else { 2 } } else { 0 };
```
mattsse, line 115: This new hidden-data state makes the existing fixed width guard stale. `draw_layout` still returns early below 100 columns because that width was required for the data and stack panes, so after `:data` and `:stack` are both hidden an 80–99-column terminal still shows ‘Terminal size too small’ instead of the source-only layout.

Suggested change in `draw_layout`:

```rust
let min_width = if self.show_data || self.show_stack { 100 } else { 1 };
```

Please also cover this with a narrow source-only render test (for example, an 80-column `TestBackend`) so the reclaimed layout is exercised below the old limit.

mattsse (commented):
One suggestion for the hidden-data layout.

stevencartavia (approved):

mablr (approved):

PR #15792, opened 2026-07-16
fix(anvil): honor `eth_simulateV1` validation

mablr:
## Motivation

Closes OSS-561.

This aligns `eth_simulateV1` with its validation semantics: validation defaults to disabled, permissive simulations ignore nonce and base-fee checks, validating simulations enforce transaction checks and return the expected `-38010` through `-38014` errors, and omitted nonces follow simulated state across calls.

First-child block derivation and base-fee progression remain tracked by OSS-564. Fork-correct blob fee and request behavior, including the zero-cap validation case, remains tracked by OSS-565.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -5124,12 +5124,11 @@ impl Backend<FoundryNetwork> {
                     request.gas = Some(requested_gas);
 
                     let caller = request.from.unwrap_or_default();
+                    let caller_nonce = RevmDatabase::basic(&mut cache_db, caller)?
+                        .map(|account| account.nonce)
+                        .unwrap_or_default();
                     if request.nonce.is_none() {
-                        request.nonce = Some(
-                            RevmDatabase::basic(&mut cache_db, caller)?
-                                .map(|account| account.nonce)
-                                .unwrap_or_default(),
-                        );
+                        request.nonce = Some(caller_nonce);
                     }
 
                     let fee_details = FeeDetails::new(
@@ -5185,8 +5184,8 @@ impl Backend<FoundryNetwork> {
                         }
                         result => result?,
                     };
-                    if !validation && let Some(account) = state.get_mut(&caller) &&
-                        account.info.nonce == u64::MAX
+                    if !validation && caller_nonce == u64::MAX &&
+                        let Some(account) = state.get_mut(&caller)
                     {
                         account.info.nonce = 0;
                     }
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -168,7 +168,7 @@ impl<DB: Database, T> BackendInspector<DB> for T where
 }
 use parking_lot::{Mutex, RwLock, RwLockUpgradableReadGuard};
 use revm::{
-    DatabaseCommit, Inspector,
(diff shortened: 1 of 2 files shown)


mattsse (dismissed):
lgtm

crates/anvil/tests/it/simulate.rs
```diff
@@ -230,6 +230,160 @@ async fn test_simulate_create_calls_rpc() {
...
+            .unwrap()
+            .iter()
+            .all(|call| call["status"] == "0x1")
+    );
+    assert!(
+        response["result"][0]["transactions"]
+            .as_array()
+            .unwrap()
+            .iter()
+            .all(|transaction| transaction["nonce"] == "0xffffffffffffffff")
```
stevencartavia: Nit, and since this adds a test for the overflow case: with validation disabled, Geth and the official `ethSimulate-overflow-nonce.io` fixture wrap the nonce from `u64::MAX` to `0` for the second call. Would it be worth matching that behavior and updating the test expectation?

mablr: https://github.com/foundry-rs/foundry/pull/15792/commits/e617ec2c030c5235e45c37d03718488a8844142a

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5155,13 +5172,24 @@ impl Backend<FoundryNetwork> {
...
-                    )?;
+                    );
+                    let ResultAndState { result, mut state } = match execution_result {
+                        Err(BlockchainError::InvalidTransaction(error)) => {
+                            return Err(simulate_transaction_error(error));
+                        }
+                        result => result?,
+                    };
+                    if !validation && let Some(account) = state.get_mut(&caller) &&
+                        account.info.nonce == u64::MAX
```
stevencartavia: I think this wraps one call too early when the starting nonce is `u64::MAX - 1`. I reproduced `MAX-1, 0, 1` here, while Geth returns `MAX-1, MAX, 0`. Could we check whether the pre-execution state nonce was already `MAX` instead?

mablr: https://github.com/foundry-rs/foundry/pull/15792/commits/c23125bb4ba76c8b85395bf1c0b00c140ddae8eb

DaniPopes (dismissed):

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5155,13 +5171,24 @@ impl Backend<FoundryNetwork> {
...
                         op_deposit,
-                    )?;
+                    );
+                    let ResultAndState { result, mut state } = match execution_result {
+                        Err(BlockchainError::InvalidTransaction(error)) => {
+                            return Err(simulate_transaction_error(error));
+                        }
+                        result => result?,
+                    };
+                    if !validation && caller_nonce == u64::MAX &&
```
stevencartavia, line 5187: This also applies the manual wrap to contract creation. With a caller nonce of `MAX`, a create followed by a call returns `MAX, 0` here, while Geth leaves the nonce at `MAX` because the create fails with nonce overflow. Could we restrict this workaround to `TxKind::Call`?

mattsse (dismissed):

stevencartavia (approved):

wyattloveslife-create (commented):
Pardon my confusion

PR #15798, opened 2026-07-16
fix(doc): resolve inherited member links

### LINKED ISSUE #9506: forge doc: links to functions or objects in other files don't work
### Component

Forge

- [X] Foundry
- [X] Foundryup

forge 0.2.0

forge doc --build --serve

### Operating System

Linux

### Describe the bug

Hello, I need help with the forge doc please.
When I build the doc, some links to functions/objects in the parent contract (when your contract is inherited from the library) don't work. I have created a repo to view a simple example of code: https://github.com/lilyanB/testFoundry/blob/master/src/libraries/lib.sol#L6

The result when I run `forge doc --build --serve`:
image

Thanks

mablr:
## Motivation

Close #9506.

Index members from the linearized base contracts so bare NatSpec cross-references link to the declaring page.

### DIFF
--- a/crates/doc/src/hir_ext.rs
+++ b/crates/doc/src/hir_ext.rs
@@ -1002,7 +1002,7 @@ fn slug_anchor_segment(s: &str) -> String {
         if ch.is_ascii_alphanumeric() || ch == '_' {
             out.push(ch);
             last_was_dash = false;
-        } else if !last_was_dash && !out.is_empty() {
+        } else if ch != '$' && !last_was_dash && !out.is_empty() {
             out.push('-');
             last_was_dash = true;
         }
--- a/crates/doc/src/hir_ext.rs
+++ b/crates/doc/src/hir_ext.rs
@@ -8,7 +8,9 @@
 
 use path_slash::PathBufExt;
 use solar::{
-    ast::{CommentKind, ContractKind, DocComments, FunctionKind, ItemKind, NatSpecKind},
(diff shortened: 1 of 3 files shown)


mattsse (dismissed):
lgtm

crates/doc/src/hir_ext.rs
```diff
@@ -677,8 +791,39 @@ impl LocalMembers {
...
+    }
+
+    /// Page and anchor for a bare inherited-member reference.
+    ///
+    /// The outer option indicates whether the name is inherited; the inner option is absent when
+    /// the effective declaration has no rendered page.
+    fn inherited_member_link(&self, member: &str, current_page: &Path) -> Option<Option<String>> {
+        let page = self.inherited.get(member)?;
+        Some(page.as_ref().map(|page| {
+            format!("{}#{}", page_link(page, current_page), slug_anchor_segment(member))
```
stevencartavia, line 805: This bare inherited-member link uses `slug_anchor_segment`, but `$` is valid in Solidity identifiers and Vocs removes it rather than replacing it with `-`. For a base variable named `balance$raw`, `{balance$raw}` links to `#balance-raw`, while Vocs renders the heading with `id="balanceraw"`, so the link is broken. Could we use Vocs-compatible slugging here or emit a matching explicit anchor?

stevencartavia (approved):

figtracer (approved):

Issue #9146, opened 2024-10-19
security: automatic remappings can be dangerous


QGarchery:
### Component

Forge

### Have you ensured that all of these are up to date?

- [X] Foundry
- [X] Foundryup

### What version of Foundry are you on?

forge 0.2.0 (a8c3e9c 2024-10-19T00:21:12.472031180Z)

### What command(s) is the bug in?

forge build

### Operating System

Linux

### Describe the bug

Forge infers remappings by taking into account remappings of sub-projects. When there are conflicting remappings, it seems like the longest / most specified has the priority. This can be dangerous, as adding sub-projects can now completely change the code that is executed, even if the remappings of the root project are not changed.

Reproduced in [this test](https://github.com/QGarchery/forge-automatic-remappings/blob/9d10257bdcec2aff92c5b344b5fcf54eb150d27e/test/ERC20Test.sol#L17-L19). Notice that removing the `lib/interesting-project` makes the test pass again. Also notice how `interesting-project` can be very deep in the sub-projects, so it can be difficult to detect that the executed bytecode has changed.

Proposed solution: make remappings of the root project have priority over remappings of sub-projects.

grandizzy:
reopen per https://github.com/foundry-rs/foundry/pull/9274

PR #15800, opened 2026-07-17
fix(config): prioritize nested remappings

### LINKED ISSUE #6431: Forge remappings output wrong with multiple remappings.txt on dependencies.
### Component

Forge

- [ ] Foundry
- [X] Foundryup

forge 0.2.0 (33f3fee 2023-05-27T00:03:28.386134290Z)

forge flatten -C src -o src/flatten/1.0/MyContract.sol src/1.0/MyContract.sol

### Operating System

Linux

### Describe the bug

Here's what I run:
```
$ forge install ProjectOpenSea/seadrop
```

then I put remappings.txt as shown on the pic (right side), the other one is from opensea's drop repo (I create another [issue](https://github.com/ProjectOpenSea/seadrop/issues/101) see the left side of the pic).

image

stevencartavia:
Nested dependency remappings can point to a source directory such as `contracts/` or `src/`, but filesystem detection may incorrectly replace them with the package root.

Preserve configured nested remappings when they refine the detected package path, while keeping root and direct dependency precedence unchanged.

Fixes #6431.


decofe:
Potential regression to cover: this change can narrow a package-root remapping globally and break imports outside the configured source directory.

Example fixture:

```text
lib/outer/
├── foundry.toml
├── remappings.txt       # inner/=lib/inner/contracts/
├── src/Outer.sol        # imports "inner/Root.sol"
└── lib/inner/
    ├── Root.sol
    └── contracts/
```

Before this change, auto-detection yields `inner/=lib/outer/lib/inner/`, so `inner/Root.sol` resolves. This change refines it to `inner/=lib/outer/lib/inner/contracts/`, so the same import resolves to the nonexistent `contracts/Root.sol`.

Could we add coverage for mixed package layouts where Solidity files are imported from the package root, `interfaces/`, or `utils/`, and clarify whether globally narrowing the mapping is intended? I prepared the package-root regression test locally, but could not execute it because Cargo's `foundry-rs/optimism` dependency fetch returned 401.

grandizzy (changes requested):
pls consider https://github.com/foundry-rs/foundry/pull/15800#issuecomment-4999021710 Projects are using remappings in many ways and our unit tests doesn't cover them so likely a fix introduce regressions. Some other scenarios to take into account for this fix:

- Cross-dependency leakage: two parents defining the same alias differently remain uncontextualized; the globally “closest” path wins, potentially compiling one parent against the other’s dependency.
- Dependency-controlled hijack: a transitive dependency can narrow an auto-detected alias to a deeper, unexpected directory—similar to the security concern in [#9146](https://github.com/foundry-rs/foundry/issues/9146).
- Scoped/versioned packages: filtering self-remappings by directory basename fails when folder and import names differ, e.g. foo-1.2.0 declaring `foo/` or `@scope/foo/`.
- Symlinks/nonexistent targets: `Path::starts_with()` is lexical when canonicalization fails, so `..` paths may incorrectly qualify as descendants.
- Performance: every nested `foundry.toml` is loaded twice, adding configuration and filesystem work across large dependency trees.

decofe:
Remaining contextual-remapping regressions to cover:

**1. Root file-level context**

```text
project/
├── foundry.toml
├── src/A.sol             # imports "@x/X.sol"
├── src/B.sol
└── lib/x/X.sol
```

```toml
[profile.default]
remappings = ["src/A.sol:@x/=lib/x/"]
```

The context becomes `src/A.sol/`. Solar checks `parent.starts_with(context)`, so parent `src/A.sol` no longer matches and the import fails.

**2. Dependency file-level context**

```text
project/
└── lib/outer/
    ├── foundry.toml      # remappings=["src/A.sol:@x/=lib/x/"]
    ├── src/A.sol         # imports "@x/X.sol"
    └── lib/x/X.sol
```

The context becomes `lib/outer/src/A.sol/`, producing the same mismatch inside the dependency.

**3. Source-prefix context without a directory**

```text
project/
├── foundry.toml
├── src/generatedA.sol    # imports "@x/X.sol"
└── lib/x/X.sol
```

```toml
[profile.default]
remappings = ["src/generated:@x/=lib/x/"]
```

Contexts use source-unit prefix matching, so this currently applies to `src/generatedA.sol`. The change rewrites it as `src/generated/`, which no longer matches. If declared by a dependency, it is dropped earlier because `src/generated` cannot be canonicalized.

These cases come from unconditionally treating contexts as directories. Could we preserve exact user-supplied contexts and only add directory boundaries to synthesized dependency contexts?

grandizzy (changes requested):
ty! Please check variations in https://github.com/foundry-rs/foundry/pull/15800#issuecomment-5002438031 🙏

crates/config/src/providers/remappings.rs
```diff
@@ -239,62 +264,151 @@ impl RemappingsProvider<'_> {
...
+
+        // Import resolution uses the first applicable remapping. Prefer deeper dependency contexts
+        // and longer import prefixes before their broader fallbacks.
+        remappings.sort_by(|a, b| {
+            let context_depth = |r: &Remapping| {
+                r.context.as_deref().map(Path::new).map_or(0, |p| p.components().count())
+            };
+            context_depth(b)
+                .cmp(&context_depth(a))
+                .then_with(|| b.name.len().cmp(&a.name.len()))
```
mablr, line 495: This new contextual ordering exposes a collision in `into_inner`: its key is `context + name`, so `(lib/outer/, lib/inner/c/)` and `(lib/outer/lib/inner/, c/)` collapse to the same string and one valid mapping is dropped. Please deduplicate by `(context, name)` instead.

crates/config/src/providers/remappings.rs
```diff
@@ -239,62 +264,151 @@ impl RemappingsProvider<'_> {
...
-        // Eventually, we could set context for remappings at this location,
-        // taking into account the OS platform. We'll need to be able to handle nested
-        // contexts depending on dependencies for this to work.
-        // For now, we just leave the default context (none).
         let mut remappings =
             config.remappings.into_iter().map(Remapping::from).collect::<Vec<Remapping>>();
 
+        remappings = remappings
+            .into_iter()
+            .filter_map(|remapping| Self::with_dependency_context(remapping, lib))
```
mablr: This scopes every dependency-declared alias, while the only global replacement above uses the checkout directory name. A direct dep in `lib/foo-1.2.0` declaring `foo/=src/` therefore loses its root-visible `foo/` alias. Please preserve configured self-remappings globally for direct dependencies.

crates/forge/tests/cli/config.rs
```diff
@@ -961,37 +962,277 @@ forgetest_init!(can_detect_lib_foundry_toml, |prj, cmd| {
...
+        pretty_err(&lib, fs::create_dir_all(&lib));
+        pretty_err(
+            &lib,
+            fs::write(lib.join(format!("{contract}.sol")), format!("contract {contract} {{}}\n")),
+        );
+    }
+    prj.add_source("A.sol", "import {X} from \"@x/X.sol\"; contract A is X {}\n");
+    prj.add_source("generatedA.sol", "import {Y} from \"@y/Y.sol\"; contract GeneratedA is Y {}\n");
+
+    cmd.arg("build").assert_success();
```
mablr, line 1269: The graph resolver matches contexts against the importer’s parent directory, so `src/A.sol` is checked against `src` and cannot match; Solar-specific normalization does not fix that earlier step. Please update `foundry-compilers` for source-unit contexts or keep this PR to directory contexts.

stevencartavia: [fde6311](https://github.com/foundry-rs/foundry/pull/15800/commits/fde6311f499cc8b2c1596390de45a1df672fd65d)

crates/config/src/providers/remappings.rs
stevencartavia: [fde6311](https://github.com/foundry-rs/foundry/pull/15800/commits/fde6311f499cc8b2c1596390de45a1df672fd65d)

crates/config/src/providers/remappings.rs
stevencartavia: [fde6311](https://github.com/foundry-rs/foundry/pull/15800/commits/fde6311f499cc8b2c1596390de45a1df672fd65d)

crates/config/src/providers/remappings.rs
```diff
@@ -206,18 +224,21 @@ impl RemappingsProvider<'_> {
...
         // scan all library dirs and autodetect remappings
-        // TODO: if a lib specifies contexts for remappings manually, we need to figure out how to
-        // resolve that
         if self.auto_detect_remappings {
             let (nested_foundry_remappings, auto_detected_remappings) = rayon::join(
                 || self.find_nested_foundry_remappings(),
                 || self.auto_detect_remappings(),
             );
 
+            // Root remappings remain first, followed by dependency-scoped remappings and global
```
mablr: This still drops valid overlapping explicit mappings. For `@scope/pkg/` plus `@scope/` in the same dependency, the sort visits the specific alias first and `push` rejects the broader fallback, breaking `@scope/other/...`. Explicit mappings should not use autodetection's prefix-conflict filtering.

crates/config/src/providers/remappings.rs
```diff
@@ -206,18 +224,21 @@ impl RemappingsProvider<'_> {
...
+            // filesystem fallbacks. The compiler resolves the first applicable remapping, so a
+            // dependency's configuration applies within that dependency without affecting root or
+            // sibling imports.
+            let (dependency_remappings, package_remappings): (Vec<_>, Vec<_>) =
```
mablr: These contextual mappings are later passed through `Remappings::with_figment`, whose project-path check rejects `src/`, `test/`, and `script/` regardless of context. A dependency-scoped `src/` alias therefore disappears in the CLI path; that exclusion should apply only to global mappings.

crates/config/src/providers/remappings.rs
```diff
@@ -239,62 +260,165 @@ impl RemappingsProvider<'_> {
...
+        // broader filesystem fallback.
+        let package_name =
+            lib.file_name().and_then(|name| name.to_str()).map(|name| format!("{name}/"));
+        let configured_package_remappings = if direct {
+            remappings
+                .iter()
+                .filter(|remapping| {
+                    remapping.context.is_none()
+                        && Some(&remapping.name) != package_name.as_ref()
+                        && Path::new(&remapping.path) == config.src
```
mablr: Why must a direct dependency alias target `config.src` to remain global? `lib/foo-1.2.0` with `foo/=contracts/` previously supported root imports, but now `foo/` is scoped and only `foo-1.2.0/` is global. This is a direct-dependency compatibility regression.

crates/config/src/providers/remappings.rs
```diff
@@ -239,62 +260,165 @@ impl RemappingsProvider<'_> {
...
             })
-            .collect::<Vec<_>>()
-            .into_iter()
+            .flat_map(foundry_toml_dirs)
+            .map(|lib| (lib, true))
+            .collect::<Vec<_>>();
+        let mut seen = HashSet::new();
+        let mut remappings = Vec::new();
+
+        while let Some((lib, direct)) = pending.pop() {
```
mablr: `pop()` reverses configured `libs` order. For equal-name, equal-depth aliases, `insert_closest` keeps the first visited candidate, so the last library now wins. Please preserve configured library precedence or encode an explicit deterministic tie-break.

crates/config/src/providers/remappings.rs
```diff
@@ -239,62 +260,165 @@ impl RemappingsProvider<'_> {
...
+                || context.components().any(|component| component == Component::ParentDir)
+            {
+                trace!(?context, ?lib, "skipping dependency remapping with escaping context");
+                return None;
+            }
+            lib.join(context)
+        } else {
+            lib.to_path_buf()
+        };
+        let Ok(context) = dunce::canonicalize(context) else { return None };
```
mablr: This silently removes valid explicit contexts whose directories are generated after config loading. Contexts should be normalized and escape-checked lexically; existence at load time should not be required.

crates/config/src/providers/remappings.rs
```diff
@@ -239,62 +260,165 @@ impl RemappingsProvider<'_> {
...
+            .flat_map(|p| {
+                if p.is_absolute() {
+                    vec![p.clone(), self.root.join("lib")]
+                } else {
+                    vec![self.root.join(p)]
+                }
             })
-            .collect::<Vec<_>>()
-            .into_iter()
+            .flat_map(foundry_toml_dirs)
```
mablr, line 446: `foundry_toml_dirs` only scans to depth 1, so `node_modules/@scope/pkg/foundry.toml` is never reached here. Scoped packages therefore miss the nested-config behavior this change introduces.

crates/config/src/providers/remappings.rs
stevencartavia: [981c8e4](https://github.com/foundry-rs/foundry/pull/15800/commits/981c8e4945d86d3d600d03e049eae6d24c2baa06)

crates/config/src/providers/remappings.rs
stevencartavia: [981c8e4](https://github.com/foundry-rs/foundry/pull/15800/commits/981c8e4945d86d3d600d03e049eae6d24c2baa06)

crates/config/src/providers/remappings.rs
stevencartavia: [99ce9f7](https://github.com/foundry-rs/foundry/pull/15800/commits/99ce9f7bee6ff4192336fbcf46fa55dee9147a71)

crates/config/src/providers/remappings.rs
stevencartavia: [99ce9f7](https://github.com/foundry-rs/foundry/pull/15800/commits/99ce9f7bee6ff4192336fbcf46fa55dee9147a71)

crates/config/src/providers/remappings.rs
stevencartavia: [c7c217d](https://github.com/foundry-rs/foundry/pull/15800/commits/c7c217d10ac4cd49186a2027e84942b170a0c009)

crates/config/src/providers/remappings.rs
stevencartavia: leaving this one for a follow-up

crates/config/src/providers/remappings.rs
```diff
@@ -239,62 +288,173 @@ impl RemappingsProvider<'_> {
...
+                remappings.append(&mut nested_remappings);
+                pending.extend(
+                    nested_libs.into_iter().flat_map(foundry_toml_dirs).map(|lib| (lib, false)),
+                );
+            }
+        }
+
+        // Import resolution uses the first applicable remapping. Prefer deeper dependency contexts
+        // and longer import prefixes before their broader fallbacks.
+        remappings.sort_by(|a, b| {
```
mablr: This ordering only defines foundry-compilers' first-match behavior. Solar selects by longest context/prefix, so an explicit ancestor context and generated deeper context can resolve different files across build vs lint/doc. We need one documented precedence model that both consumers implement, plus cross-resolver coverage.

stevencartavia: [c0fafde](https://github.com/foundry-rs/foundry/pull/15800/commits/c0fafde461f922a97d914049be9ed43715088e3c)

crates/config/src/providers/remappings.rs
```diff
@@ -331,7 +542,17 @@ impl Provider for RemappingsProvider<'_> {
...
+            .map(|r| {
+                let has_boundary = r.context.as_ref().is_some_and(|context| context.ends_with('/'));
+                let mut r = RelativeRemapping::new(r, self.root);
+                if has_boundary
+                    && let Some(context) = &mut r.context
+                    && !context.ends_with('/')
+                {
+                    context.push('/');
+                }
+                r.to_string()
```
mablr, line 694: On Windows, an external absolute context serializes as `C:/deps/pkg/:alias/=...`; the parser splits the left side at the first `:`, so this reparses with context `C` and a corrupted alias. This PR now generates absolute contexts for absolute libs. Please avoid this string round-trip or make parsing drive-aware, with a Windows round-trip test.

crates/forge/tests/cli/config.rs
```diff
@@ -1059,23 +1408,43 @@ forgetest_init!(can_prioritise_project_remappings, |prj, cmd| {
...
 Global:
 
+Context: lib/dep1/
+
 
 "#]]);
+    cmd.forge_fuse().arg("build").assert_success();
 });
 
 // Verifies the contract invariant: `forge remappings` and `forge remappings --pretty` emit
```
mablr, line 1898: This only tests already-grouped input. `--pretty` groups through a `BTreeMap`, so interleaved input such as global, contextual, global is reordered and no longer matches plain stdout. Please interleave this fixture and preserve original stdout order.

stevencartavia: [27e741c](https://github.com/foundry-rs/foundry/pull/15800/commits/27e741cfb3ff4c654f3ce702b10d0e66f805f4f1)

crates/config/src/providers/remappings.rs
stevencartavia: [27e741c](https://github.com/foundry-rs/foundry/pull/15800/commits/27e741cfb3ff4c654f3ce702b10d0e66f805f4f1)

crates/config/src/providers/remappings.rs
```diff
@@ -226,7 +370,22 @@ impl RemappingsProvider<'_> {
...
+            // Preserve legacy global aliases from direct dependencies while retaining closest-path
+            // arbitration with auto-detected package mappings and stable output grouping.
+            let mut package_lib_remappings = BTreeMap::new();
+            for r in package_remappings {
+                let path = PathBuf::from(r.path);
+                if let Some(existing) = lib_remappings
+                    .get_mut(&r.context)
+                    .and_then(|remappings| remappings.get_mut(&r.name))
+                {
+                    if existing.components().count() > path.components().count() {
```
mablr, line 398: **Blocker:** This path-depth comparison still breaks the PR's primary case when the dependency directory and configured alias have the same name. With `node_modules/foo/foundry.toml` declaring `foo/=contracts/`, autodetection contributes `foo/=node_modules/foo/`; because that path is shorter, it remains the global winner while the configured mapping is only retained as `node_modules/foo/:foo/=node_modules/foo/contracts/`. A root `import "foo/Foo.sol"` then looks for `node_modules/foo/Foo.sol` and fails. The current `foo-1.2.0` test avoids this collision because autodetection produces a different alias. Please add the exact `node_modules/foo` regression case and give a direct dependency's configured global mapping precedence over its synthesized/filesystem fallback rather than arbitrating these two sources by path depth.

crates/config/src/providers/remappings.rs
```diff
@@ -119,6 +233,27 @@ impl Remappings {
...
+    }
+
+    /// Extend with lower-precedence explicit remappings while preserving overlaps within the
+    /// incoming batch.
+    pub fn extend_explicit_with_precedence(&mut self, remappings: Vec<Remapping>, root: &Path) {
+        let authoritative = self.remappings.clone();
+        for remapping in remappings {
+            if !authoritative.iter().any(|existing| {
+                remapping_context_covers(root, existing, &remapping)
+                    && remapping_names_overlap(&existing.name, &remapping.name)
```
mablr: **High:** The symmetric overlap check removes valid broader dependency fallbacks. For example, a root `pkg/sub/=src/local/` plus a dependency-scoped `pkg/=src/pkg/` drops the latter, so that dependency can no longer resolve `pkg/Other.sol`; only imports under `pkg/sub/` are actually covered by the root mapping. This reproduces as a build failure. Precedence needs to be directional: an authoritative broad `pkg/` may suppress an incoming `pkg/sub/`, but authoritative `pkg/sub/` must not suppress incoming broader `pkg/` outside that subtree. Please cover both overlap directions.

crates/config/src/utils.rs
```diff
@@ -140,14 +140,16 @@ pub fn to_array_value(val: &str) -> Result<Value, figment::Error> {
 ///     ├── foundry.toml
 /// ```
 pub fn foundry_toml_dirs(root: impl AsRef<Path>) -> Vec<PathBuf> {
-    walkdir::WalkDir::new(root)
+    let mut dirs = walkdir::WalkDir::new(root)
         .max_depth(1)
```
mablr: **High:** Nested remapping discovery cannot load scoped npm packages because `max_depth(1)` stops at `node_modules/@scope` and never reaches `node_modules/@scope/pkg/foundry.toml`. A package there declaring `@scope/pkg/=contracts/` is reduced to the auto-detected `@scope/=node_modules/@scope/`, and `import "@scope/pkg/Foo.sol"` fails. Although this traversal limit predates the PR, the new nested-config precedence relies on this helper, so scoped dependencies remain an uncovered form of the bug being fixed. Please add namespace-aware discovery and an integration test for `node_modules/@scope/pkg`.

crates/config/src/utils.rs
```diff
@@ -140,14 +140,16 @@ pub fn to_array_value(val: &str) -> Result<Value, figment::Error> {
...
         .into_iter()
         .filter_map(Result::ok)
         .filter(|e| e.file_type().is_dir())
```
mablr: **High:** This filters direct dependency symlinks out before canonicalization: `WalkDir` reports `lib/foo -> /external/foo` as a symlink, so `file_type().is_dir()` is false and the target's `foundry.toml` is never loaded despite this function's symlink contract. Reproduced with a symlinked dependency declaring a custom `zzalias/=vendor/`; `forge remappings` emits only `foo/=lib/foo/src/`, and the dependency import fails. Discovery needs to retain the lexical path for aliases/output while using the canonical target for containment, deduplication, and cycle handling. Please add a symlinked-dependency regression test.

crates/config/src/providers/remappings.rs
stevencartavia: [32ccfda](https://github.com/foundry-rs/foundry/pull/15800/commits/32ccfdab7ea97ca25011d2bf503464bc5021a3b4)

crates/config/src/providers/remappings.rs
stevencartavia: [32ccfda](https://github.com/foundry-rs/foundry/pull/15800/commits/32ccfdab7ea97ca25011d2bf503464bc5021a3b4)

crates/config/src/utils.rs
stevencartavia: [32ccfda](https://github.com/foundry-rs/foundry/pull/15800/commits/32ccfdab7ea97ca25011d2bf503464bc5021a3b4)

crates/config/src/utils.rs
stevencartavia: [32ccfda](https://github.com/foundry-rs/foundry/pull/15800/commits/32ccfdab7ea97ca25011d2bf503464bc5021a3b4)

crates/config/src/providers/remappings.rs
```diff
@@ -119,6 +234,34 @@ impl Remappings {
...
+    }
+
+    /// Extend with lower-precedence explicit remappings while preserving overlaps within the
+    /// incoming batch.
+    pub fn extend_explicit_with_precedence(&mut self, remappings: Vec<Remapping>, root: &Path) {
+        let authoritative = self.remappings.clone();
+        for remapping in remappings {
+            if !authoritative.iter().any(|existing| {
+                remapping_context_covers(root, existing, &remapping)
+                    && remapping_name_covers(&existing.name, &remapping.name)
```
mablr, line 263: **Blocker:** This directional check keeps the inverse overlap (`pkg/sub/` global + `lib/dep/:pkg/` scoped). `foundry-compilers` then resolves a dependency import `pkg/sub/X.sol` with the first/global mapping, while solc/Solar choose the longer dependency context. Extending `narrow_root_remapping_preserves_broad_dependency_fallback` with that import makes `forge build` fail because the graph and compiler load different files. Please cover this overlap and make both resolver paths select the same mapping.

crates/forge/src/cmd/remappings.rs
```diff
@@ -24,20 +24,17 @@ impl RemappingArgs {
...
-            let mut groups = BTreeMap::<_, Vec<_>>::new();
-            for remapping in config.remappings {
-                groups.entry(remapping.context.clone()).or_default().push(remapping);
+            let mut groups = BTreeSet::new();
+            for remapping in &config.remappings {
+                sh_println!("{remapping}")?;
+                groups.insert(remapping.context.clone());
             }
-            for (group, remappings) in groups {
+            for group in groups {
```
mablr: `--pretty` no longer groups the output: the first loop prints every mapping, then this loop emits bare context headers after them. This contradicts the flag help. Could we emit each header while iterating remappings in original order so the mappings remain attached to their group?

crates/config/src/providers/remappings.rs
stevencartavia: 4a5868e9987f943880cce853ec85ac10c6a06920

crates/forge/src/cmd/remappings.rs
stevencartavia: 4a5868e9987f943880cce853ec85ac10c6a06920

crates/config/src/providers/remappings.rs
```diff
@@ -239,62 +403,227 @@ impl RemappingsProvider<'_> {
...
+                Ok(_) => break,
+                Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
+                    ancestor = ancestor.parent()?;
+                }
+                Err(_) => return None,
+            }
         }
-        remappings
+        let Ok(canonical_ancestor) = dunce::canonicalize(ancestor) else { return None };
+        if !canonical_ancestor.is_dir() || !canonical_ancestor.starts_with(canonical_lib) {
```
mablr: Could we preserve explicit file and source-prefix contexts here? `src/A.sol:...` is dropped because its canonical ancestor is not a directory, while `src/generated:...` later gains `/` and changes its matching semantics. Only synthesized dependency contexts should receive a directory boundary.

stevencartavia: b81836764e9c40bae2fce613dd7bdbaaf8b7011e

crates/config/src/providers/remappings.rs
```diff
@@ -236,62 +405,232 @@ impl RemappingsProvider<'_> {
...
             .into_iter()
+            .flat_map(foundry_toml_dirs)
+            .map(|lib| (lib, true))
+            .collect::<VecDeque<_>>();
+        let mut seen = HashSet::new();
+        let mut remappings = NestedRemappings::default();
+
+        while let Some((lib, direct)) = pending.pop_front() {
+            let Ok(canonical_lib) = dunce::canonicalize(&lib) else { continue };
+            if !seen.insert(canonical_lib.clone()) {
```
mattsse: **Blocker:** Canonical-path deduplication drops valid lexical contexts when the same transitive dependency is reached through more than one symlink. I reproduced this with two parent dependencies whose `child-left` and `child-right` entries point to the same canonical package; that package declares `zzalias/=vendor/`. Only the first contextual remapping is emitted, and the second parent's `Child.sol` cannot resolve `zzalias/Z.sol` during `forge build`.

The config can be loaded/cached once by canonical path, but its remappings still need to be materialized for every lexical dependency path. Please separate canonical cycle/config tracking from lexical emission (for example, cache parsed config by canonical path and track emitted contexts by `(canonical, lexical)`), then add a two-parent/two-symlink build regression. This is the remaining case not covered by the single-symlink test.

crates/forge/tests/cli/config.rs
```diff
@@ -982,37 +988,592 @@ forgetest_init!(can_detect_lib_foundry_toml, |prj, cmd| {
...
+    );
+    pretty_err(&dep, fs::write(dep.join("contracts/Foo.sol"), "contract Foo {}\n"));
+    prj.add_source(
+        "UsesFoo.sol",
+        "import {Foo} from \"@scope/pkg/Foo.sol\"; contract UsesFoo is Foo {}\n",
+    );
+
+    cmd.args(["remappings"]).assert_success().stdout_eq(str![[r#"
+node_modules/@scope/pkg/:@scope/pkg/=node_modules/@scope/pkg/contracts/
+@scope/pkg/=node_modules/@scope/pkg/contracts/
```
mattsse, line 1501: This snapshot fails on the current head: `forge remappings` also emits `@scope/=node_modules/@scope/`. If that broad fallback from the merged master behavior is intended, please add that exact line after `@scope/pkg/=node_modules/@scope/pkg/contracts/`; otherwise the precedence logic needs to suppress it intentionally. As written, `cargo nextest run -p forge --test cli -E 'test(/config::scoped_dependency_config_remappings_are_discovered/)'` fails on all retries.

mattsse (changes requested):
Two merge-blocking issues remain on `b81836764e9c40bae2fce613dd7bdbaaf8b7011e`:

1. Canonical-only traversal still loses configured remappings for a second lexical symlink path to the same transitive dependency. I reproduced the resulting unresolved import and build failure before the latest explicit-context-only commit; the current traversal is unchanged, so the symlink portion of the earlier changes-requested feedback is not fully addressed.
2. The scoped-package integration test currently fails because the latest merged base behavior adds the broad `@scope/=node_modules/@scope/` fallback, but the inline snapshot was not updated.

The inline comments include concrete implementation and regression-test suggestions.

crates/forge/tests/cli/config.rs
stevencartavia: 74bffa62b8b1d19dcea92bcb9fce562d447d00bb

crates/config/src/providers/remappings.rs
stevencartavia: 74bffa62b8b1d19dcea92bcb9fce562d447d00bb

mattsse (dismissed):
lgtm

mablr (dismissed):

crates/config/src/providers/remappings.rs
```diff
@@ -236,62 +405,239 @@ impl RemappingsProvider<'_> {
...
+                || !emitted.insert((canonical_lib.clone(), lib.clone()))
+            {
+                continue;
+            }
+
+            trace!(?lib, ?canonical_lib, "find all remappings of nested foundry.toml");
+            let config = configs.entry(canonical_lib.clone()).or_insert_with(|| {
+                Config::load_with_root_and_fallback_without_auto_detected_remappings(&canonical_lib)
+                    .ok()
+                    .map(Config::sanitized)
```
mablr: Could we preserve the declared `libs` order here? `sanitized()` sorts `config.libs`, but equal-depth remapping ties keep the first path encountered, so a nested `libs = ["vendor-b", "vendor-a"]` can incorrectly prefer `vendor-a`. Please canonicalize and sanitize remappings without sorting `libs`, stable-deduplicate them, and cover the nested tie case.

mablr (commented):
another one :/

crates/config/src/providers/remappings.rs
stevencartavia: ad0aa3ad3b232b44db06adade69c4ce5ecd391de

mablr (approved):

figtracer (approved):

figtracer (approved):

PR #15801, opened 2026-07-17
fix: trace pranked caller

### LINKED ISSUE #10050: `forge script` trace does not respect `vm.startPrank` for caller.
### Component

Forge

- [x] Foundry
- [x] Foundryup

1.0.0-stable

1.0.0

forge script

### Operating System

macOS (Apple Silicon)

### Describe the bug

See the following script:

```
 uint256 public counter;
    function increment() public {
        counter += 1;
    }
}

contract DeployAndTest is Script {
    function run() external {
        vm.startBroadcast();

        // Deploy EmptyContract
        EmptyContract empty = new EmptyContract();

        vm.stopBroadcast();

        // Arbitrary address A
        address A = address(0x1234);

        // Prank as A
        vm.startPrank(A);
        empty.increment();
        vm.stopPrank();
    }
}
```

Should the trace for `empty.increment` show the caller as `A`, or the deployed script?

IMO, it should be A. This is confusing and counter-intuitive, and is breaking a tool that I'm building. Am I missing something?

stevencartavia:
Updates call traces after cheatcode processing so the caller reflects `prank` and `broadcast` overrides while preserving trace ordering for nested cheatcodes. Closes #10050.

### DIFF
--- a/crates/evm/evm/src/inspectors/stack.rs
+++ b/crates/evm/evm/src/inspectors/stack.rs
@@ -1417,6 +1417,11 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>>
             },
         );
 
+        // The tracer records call inputs before cheatcodes apply caller overrides such as pranks
+        // and broadcasts. Keep the trace lifecycle ordering, but remember the node so its caller
+        // can be synchronized with the inputs that are actually executed.
+        let trace_idx = self.tracer.as_ref().map(|tracer| tracer.traces().nodes().len() - 1);
+        let mut cheatcode_outcome = None;
         if let Some(cheatcodes) = self.cheatcodes.as_deref_mut() {
             // Handle mocked functions, replace bytecode address with mock if matched.
             if let Some(mocks) = cheatcodes.mocked_functions.get(&call.bytecode_address) {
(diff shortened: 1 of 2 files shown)


crates/evm/evm/src/inspectors/stack.rs
```diff
@@ -1417,6 +1417,11 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>>
             },
         );
 
+        // The tracer records call inputs before cheatcodes apply caller overrides such as pranks
```
grandizzy, line 1420: This only corrects `Inspector::call`. `Inspector::create` has the same tracer-before-cheatcodes ordering at line 1554, so a direct `new Target()` under `startPrank` still records the script contract as the CREATE caller.

I reproduced this by replacing the test’s `vm.deployCode(...)` with `new Target()`; the expected `0x1234` caller remained `0x9f7c…`. Please mirror this synchronization in `create` and test direct CREATE/CREATE2 rather than only `vm.deployCode`.

grandizzy (changes requested):
Direct contract creation under a prank or broadcast still records the original caller. The call path is fixed, but the equivalent create path needs the same synchronization.

crates/evm/evm/src/inspectors/stack.rs
stevencartavia: [adb6dc9](https://github.com/foundry-rs/foundry/pull/15801/commits/adb6dc91472b8860d4b417f8cd3686b81eb2ee4c)

grandizzy (approved):
lgtm

mattsse (approved):
lgtm

Ari4ka (approved):

PR #15805, opened 2026-07-17
test(cast): normalize executable suffix in mktx snapshot

decofe:
Fix the `mktx_signature_requires_from` snapshot on Windows, where clap renders the executable name as `cast.exe`. The existing snapbox `[..]` wildcard keeps the assertion portable while still checking the complete usage line.

Validation:
- `git diff --check`
- Focused test could not run locally because the pinned `foundry-rs/optimism` dependency returned HTTP 401 during fetch.

Prompted by: @grandizzy


grandizzy (approved):

mattsse (approved):
lgtm

PR #15804, opened 2026-07-17
fix(anvil): preserve tempo calls in tx requests

jxom:
`anvil --network tempo` silently drops `calls`.

Minimal repro against `anvil --network tempo`:

```bash
cast rpc eth_sendTransaction '{
  "from": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
  "type": "0x76",
  "calls": [
    {"to": "0x70997970C51812dc3A010C7d01b50e0d17dc79C8", "data": "0xdead"},
    {"to": "0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC", "data": "0xbeef"}
  ]
}'
```

PR #15806, opened 2026-07-17
fix(anvil): preserve custom request fields

mablr:
## Description

This supersedes #15804 by preserving the complete set of Tempo request extensions, including batched calls, through Anvil’s transaction-building path; a follow-up PR extending those semantics across local simulation and execution paths will come later.

`FoundryTransactionRequest` now decodes all known Tempo extensions fallibly and preserves consensus fields during typed transaction round-trips. The Tempo-aware transaction-building path rejects conflicting explicit transaction types and only infers Tempo extensions when Anvil is running with the Tempo network enabled. Calls-only Tempo batches are built without a synthetic creation call, while requests without an explicit gas limit use a conservative fallback rather than an extension-free estimate.

This PR is limited to fallible request decoding, typed transaction round-trips, and Anvil’s send/build path. Tempo-aware execution, simulation, tracing, response construction, and historical-fork forwarding are deferred to the follow-up PR.

The implementation and review were completed with AI assistance.

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -2918,7 +2918,7 @@ impl EthApi<FoundryNetwork> {
 
     pub async fn simulate_v1(
         &self,
-        request: SimulatePayload,
+        request: SimulatePayload<WithOtherFields<TransactionRequest>>,
         block_number: Option<BlockId>,
     ) -> Result<Vec<SimulatedBlock<AnyRpcBlock>>> {
         node_info!("eth_simulateV1");
@@ -4249,7 +4249,9 @@ impl EthApi<FoundryNetwork> {
         request: WithOtherFields<TransactionRequest>,
         nonce: u64,
     ) -> Result<FoundryTypedTx> {
-        let mut request = Into::<FoundryTransactionRequest>::into(request);
+        let mut request =
+            FoundryTransactionRequest::try_from_rpc_request(request, self.backend.is_tempo())
+                .map_err(|_| BlockchainError::FailedToDecodeTransaction)?;
         let from = request.from().or(self.accounts()?.first().copied());
         if let Some(from) = from {
             request.set_from(from);
@@ -4258,7 +4260,11 @@ impl EthApi<FoundryNetwork> {
         // Fill common fields for all tx types
         request.chain_id().is_none().then(|| request.set_chain_id(self.chain_id()));
         request.nonce().is_none().then(|| request.set_nonce(nonce));
-        request.kind().is_none().then(|| request.set_kind(TxKind::default()));
+        let is_tempo_batch =
+            matches!(&request, FoundryTransactionRequest::Tempo(tx) if !tx.calls.is_empty());
+        if request.kind().is_none() && !is_tempo_batch {
+            request.set_kind(TxKind::default());
+        }
         if request.gas_limit().is_none() {
             let fallback_gas_limit = {
                 let evm_env = self.backend.evm_env().read();
@@ -4269,17 +4275,20 @@ impl EthApi<FoundryNetwork> {
                     block_gas_limit
                 }
             };
-            let estimated_gas = self
-                .do_estimate_gas(request.as_ref().clone().into(), None, EvmOverrides::default())
-                .await
-                .map(|v| v as u64)
-                .unwrap_or_else(|_| {
-                    if is_simple_transfer_request(request.as_ref()) {
-                        MIN_TRANSACTION_GAS as u64
-                    } else {
-                        fallback_gas_limit
-                    }
-                });
+            let estimated_gas = if matches!(&request, FoundryTransactionRequest::Tempo(_)) {
+                fallback_gas_limit
+            } else {
+                self.do_estimate_gas(request.as_ref().clone().into(), None, EvmOverrides::default())
+                    .await
+                    .map(|v| v as u64)
+                    .unwrap_or_else(|_| {
+                        if is_simple_transfer_request(request.as_ref()) {
+                            MIN_TRANSACTION_GAS as u64
+                        } else {
+                            fallback_gas_limit
+                        }
+                    })
+            };
             request.set_gas_limit(estimated_gas);
         }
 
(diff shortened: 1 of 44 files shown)


crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -1316,6 +1316,45 @@ impl<N: Network> Backend<N> {
...
+                .and_then(serde_json::from_value::<TempoTransactionRequest>)
+                .map_err(|err| BlockchainError::InvalidTransactionRequest(err.to_string()))?;
+            Ok::<_, BlockchainError>((request, is_explicit))
+        });
+
+        let (evm_env, tx_env, op_deposit) = self.build_call_env(request, fee_details, block_env);
+        if let Some(request) = tempo_request {
+            let (request, is_explicit) = request?;
+            let tempo_tx = request
+                .into_tempo_tx_env(tx_env.clone(), self.tempo_hardfork().is_t1c(), is_explicit)
```
mattsse: [P1] Preserve top-level CREATE requests before canonical conversion

into_tempo_tx_env rejects an explicit Tempo request when calls is empty and request.inner.to is absent. A contract creation is represented exactly that way (to: null with non-empty initcode), even though the base tx_env.kind built above is TxKind::Create. This now breaks deployment gas estimation: the Tempo CI check fails with Invalid transaction request: empty calls list, and the full test job reports related Forge/Cast failures.

Suggested direction: normalize a non-empty top-level creation into the Tempo call list before invoking the converter, or update the canonical converter to derive the top-level call from the supplied base TxEnv, while retaining rejection for a truly empty request.

A focused regression test alongside the new top-level-call coverage could be:

~~~rust
async fn test_tempo_top_level_create_supported_by_simulation_endpoints() {
    let (api, handle) = spawn(NodeConfig::test_tempo()).await;
    let from = handle.dev_accounts().next().unwrap();
    // PUSH1 0; PUSH1 0; RETURN: valid empty-runtime initcode.
    let initcode = Bytes::from_static(&[0x60, 0x00, 0x60, 0x00, 0xf3]);
    let mut request = WithOtherFields::new(
        TransactionRequest::default().from(from).with_input(initcode),
    );
    request.inner.transaction_type = Some(0x76);

    api.call(request.clone(), None, EvmOverrides::default()).await.unwrap();
    api.estimate_gas(request.clone(), None, EvmOverrides::default()).await.unwrap();
    let access_list = api.create_access_list(request, None, None).await.unwrap();
    assert!(access_list.error.is_none());
}
~~~

mattsse (changes requested):
The canonical request conversion is the right direction, but this currently regresses top-level Tempo contract creation and leaves eth_simulateV1 outside the preservation path.

For eth_simulateV1, the RPC still deserializes into SimulatePayload<TransactionRequest>, so Tempo extension fields are gone before execution. I reproduced a calls-only type-0x76 request returning -32603: blocking task panicked; the request log showed calls had been discarded, and block construction then panicked while encoding the unknown transaction type. Please use an extensible request payload there, route it through the same canonical Tempo conversion, and add a raw-RPC regression test that submits a Tempo batch and asserts the batch survives without an error.

mattsse (changes requested):
The RPC boundary now preserves extension fields and historical-fork forwarding is fixed, but the local simulation path still drops them.

[P1] `crates/anvil/src/eth/backend/mem/mod.rs:5100-5101` immediately replaces each `WithOtherFields<TransactionRequest>` with `request.inner`. For a local `eth_simulateV1` request this discards `calls` and every other Tempo extension before both execution and response construction. A calls-only type-0x76 batch can therefore be simulated as an empty top-level creation, or reconstructed later as a malformed Tempo transaction. This leaves the original failure mode in place whenever the requested block is handled by local state rather than forwarded to a historical fork.

Suggested fix: keep the `WithOtherFields` value intact and parse it once with `FoundryTransactionRequest::try_from_rpc_request(request, self.is_tempo())`. For the Tempo variant, construct the native `TempoTxEnv` from the complete request (including `calls`, authorization, nonce/validity, and fee fields), execute it through the Tempo transaction path, and reuse that same parsed request when building the simulated transaction response. The Ethereum path can continue using `build_call_env`. If local Tempo execution must remain out of scope, reject these extensions explicitly instead of silently dropping them, although that would not provide the preservation behavior this PR is intended to add.

Please add an end-to-end raw-RPC regression under `crates/anvil/tests/it/tempo.rs`: spawn `NodeConfig::test_tempo()`, call `eth_simulateV1` without a historical block using a calls-only `type: "0x76"` request containing two distinguishable calls, and set `returnFullTransactions: true`. Assert that the RPC succeeds, the simulated call status is successful, observable return data or logs prove that both calls executed, and the returned transaction has type `0x76` with `calls` exactly equal to the submitted array. This exercises the local branch and catches both execution-time and response-time field loss.

mablr (commented):
@mattsse I outscoped the eth_simulateV1 changes from this PR; complete Tempo simulation handling will be included in the follow-up execution PR.

crates/anvil/src/eth/api.rs
```diff
@@ -4301,9 +4302,15 @@ impl EthApi<FoundryNetwork> {
         request: WithOtherFields<TransactionRequest>,
         nonce: u64,
     ) -> Result<FoundryTypedTx> {
-        let mut request =
-            FoundryTransactionRequest::try_from_rpc_request(request, self.backend.is_tempo())
-                .map_err(|_| BlockchainError::FailedToDecodeTransaction)?;
+        let transaction_type = request.transaction_type;
+        let mut request: FoundryTransactionRequest =
+            request.try_into().map_err(|_| BlockchainError::FailedToDecodeTransaction)?;
+        if matches!(&request, FoundryTransactionRequest::Tempo(_))
```
mattsse, line 4308: can we add fn is_ to all the network enums, this would be nicer than using matches!

mattsse (approved):
lgtm, smol suggestion for followup

figtracer (approved):

crates/anvil/src/eth/api.rs
mablr: already in my follow-up branch ^^ pushing asap

PR #10137, opened 2025-03-20
chore: remove experimental from cheatcodes

grandizzy:
## Motivation
- due to forge-std script issue, the cheatcodes marked as experimental are not included (so `vm.cool` wasn't ever exposed)
- remove experimental from `vm.cool` and the newly added cheatcodes and follow up with a better way to handle `forge-std` experimental cheatcodes

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #15808, opened 2026-07-17
fix(cheatcodes): warn on internal, reject removed

### LINKED ISSUE #7323: Non-stable cheatcodes should emit warnings and errors
### Component

Forge

- [X] Foundry
- [X] Foundryup

forge 0.2.0 (ce22450 2024-03-06T00:18:18.671530895Z)

_No response_

### Operating System

None

### Describe the bug

Cheatcodes can be tagged with a `status`: https://github.com/foundry-rs/foundry/blob/ce22450e4d625d12ff88fae347a68e3d3d9d2b61/crates/cheatcodes/spec/src/cheatcode.rs#L40-L57

Use of these cheatcodes should result in a warning or hard error when appropriate, which currently does not appear to have been implemented.

Related: https://github.com/foundry-rs/foundry/pull/7317#discussion_r1512788652

0xMars42:
Closes #7323. Builds on the prototype in `master...mablr/cheatcode-status-handling`.

Only `Status::Deprecated` is currently acted on at dispatch: internal and removed cheatcodes are silently treated as stable, contrary to what the `Status` docs promise. Internal cheatcode usage is now recorded before dispatch and reported once per test suite as a warning. The record set is shared by all inspectors cloned from the same suite executor, because internal usage can happen outside of any test result (contract constructors, invariant `targetContracts()` or `afterInvariant()`, corpus replays) where the per-test plumbing used by deprecation warnings never sees it. Removed cheatcodes are rejected before dispatch with `vm.<name>: cheatcode has been removed`; no cheatcode currently carries this status, so that path is covered by a unit test that fabricates one. Experimental cheatcodes keep dispatching normally: the status has been unused since #10137 and its runtime behavior has not been decided, so its doc no longer promises a warning.

The symbolic executor reimplements cheatcodes from the spec, so status handling is mirrored there: internal usage is recorded into the same per-suite set, and removed cheatcodes surface as an unsupported-operation error, consistent with how the engine already treats unknown cheatcodes.

Compared to the prototype branch, the status-to-action match in the spec is exhaustive so a future status variant cannot be silently ignored again, multi-network passes dedupe repeated suite warnings without reordering existing ones, and the deprecated warning now lists cheatcodes in sorted order (previously hash-map iteration order), which also makes that dedupe reliable.

### DIFF
--- a/crates/cheatcodes/spec/src/cheatcode.rs
+++ b/crates/cheatcodes/spec/src/cheatcode.rs
@@ -40,7 +40,7 @@ pub enum Status<'a> {
     /// The cheatcode is unstable, meaning it may contain bugs and may break its API on any
     /// release.
     ///
-    /// Use of experimental cheatcodes will result in a warning.
+    /// This status currently has no runtime effect.
     Experimental,
     /// The cheatcode has been deprecated, meaning it will be removed in a future release.
     ///
@@ -57,6 +57,31 @@ pub enum Status<'a> {
     ///
     /// Use of internal cheatcodes is discouraged and will result in a warning.
     Internal,
+}
+
+/// The runtime action associated with a cheatcode [`Status`].
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum StatusAction {
+    /// Continue with normal cheatcode dispatch.
+    Continue,
+    /// Record use of an internal cheatcode before dispatching it.
+    RecordInternal,
+    /// Reject the cheatcode before dispatching it.
+    RejectRemoved,
+}
+
+impl Status<'_> {
+    /// Returns the runtime action for this status.
+    pub const fn action(&self) -> StatusAction {
+        match self {
+            // Experimental cheatcodes dispatch normally: the status is currently unused and its
+            // runtime behavior has not been decided. Deprecated cheatcodes are recorded
+            // separately, with their replacement.
+            Self::Stable | Self::Experimental | Self::Deprecated(_) => StatusAction::Continue,
+            Self::Internal => StatusAction::RecordInternal,
+            Self::Removed => StatusAction::RejectRemoved,
+        }
+    }
 }
 
 /// Cheatcode groups.
--- a/crates/cheatcodes/src/inspector.rs
+++ b/crates/cheatcodes/src/inspector.rs
@@ -49,6 +49,7 @@ use foundry_evm_traces::{
 };
 use foundry_wallets::wallet_multi::MultiWallet;
 use itertools::Itertools;
+use parking_lot::Mutex;
 use proptest::test_runner::{RngAlgorithm, TestRng, TestRunner};
 use rand::Rng;
 use revm::{
@@ -80,6 +81,23 @@ mod utils;
 
 pub mod analysis;
 pub use analysis::CheatcodeAnalysis;
+
+/// Internal cheatcodes observed by all inspectors cloned from the same suite executor.
+#[derive(Clone, Debug, Default)]
+struct InternalCheatcodeUsage(Arc<Mutex<HashSet<&'static str>>>);
+
+impl InternalCheatcodeUsage {
+    #[inline]
+    fn record(&self, signature: &'static str) {
+        self.0.lock().insert(signature);
+    }
+
+    fn signatures(&self) -> Vec<&'static str> {
+        let mut signatures = self.0.lock().iter().copied().collect::<Vec<_>>();
+        signatures.sort_unstable();
+        signatures
+    }
+}
 
 /// Helper trait for running nested EVM operations from inside cheatcode implementations.
 pub trait CheatcodesExecutor<FEN: FoundryEvmNetwork> {
@@ -675,6 +693,10 @@ pub struct Cheatcodes<FEN: FoundryEvmNetwork = EthEvmNetwork> {
 
     /// Deprecated cheatcodes mapped to the reason. Used to report warnings on test results.
     pub deprecated: HashMap<&'static str, Option<&'static str>>,
+    /// Internal cheatcodes observed across the current test suite. Shared by all inspectors
+    /// cloned from the same suite executor, because usage can occur outside of any test result
+    /// (e.g. in constructors or corpus replays).
+    internal_cheatcodes: InternalCheatcodeUsage,
     /// Unlocked wallets used in scripts and testing of scripts.
     pub wallets: Option<Wallets>,
     /// Parsed secp256k1 private-key signers for repeated `vm.addr` / `vm.sign` calls.
@@ -772,6 +794,7 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
             ignored_traces: Default::default(),
             arbitrary_storage: Default::default(),
             deprecated: Default::default(),
+            internal_cheatcodes: Default::default(),
             wallets: Default::default(),
             private_key_signers: Default::default(),
             signatures_identifier: Default::default(),
@@ -781,6 +804,16 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
             env_overrides_snapshots: Default::default(),
             in_isolation_context: false,
         }
+    }
+
+    /// Returns the sorted internal cheatcode signatures observed by this suite.
+    pub fn internal_cheatcodes(&self) -> Vec<&'static str> {
+        self.internal_cheatcodes.signatures()
+    }
+
+    /// Records use of an internal cheatcode in this suite.
+    pub fn record_internal_cheatcode(&self, signature: &'static str) {
+        self.internal_cheatcodes.record(signature);
     }
 
     /// Enables cheatcode analysis capabilities by providing a solar compiler instance.
@@ -3140,10 +3173,6 @@ fn apply_dispatch<FEN: FoundryEvmNetwork>(
     let _guard = debug_span!(target: "cheatcodes", "apply", id = %cheatcode_id(cheat)).entered();
     trace!(target: "cheatcodes", cheat = %cheatcode_signature(cheat), "applying");
 
-    if let spec::Status::Deprecated(replacement) = cheat.status {
-        ccx.state.deprecated.insert(cheatcode_signature(cheat), replacement);
-    }
-
     // Monomorphized dispatch: calls apply_full directly, no trait objects.
     macro_rules! dispatch {
         ($($variant:ident),*) => {
@@ -3152,7 +3181,10 @@ fn apply_dispatch<FEN: FoundryEvmNetwork>(
             }
         };
     }
-    let mut result = vm_calls!(dispatch);
+    let mut result = match record_cheatcode_status(cheat, ccx.state) {
+        Ok(()) => vm_calls!(dispatch),
+        Err(err) => Err(err),
+    };
 
     // Format the error message to include the cheatcode name.
     if let Err(e) = &mut result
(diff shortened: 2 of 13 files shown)


crates/forge/src/cmd/test/mod.rs
```diff
@@ -3732,6 +3732,10 @@ fn merge_outcomes(base: &mut TestOutcome, other: TestOutcome) {
                 let base_suite = e.get_mut();
                 base_suite.test_results.extend(other_suite.test_results);
                 base_suite.warnings.extend(other_suite.warnings);
+                // Suite-level warnings (deprecated or internal cheatcodes) are re-emitted by
+                // every pass; drop exact duplicates while preserving the original order.
+                let mut seen = std::collections::HashSet::new();
+                base_suite.warnings.retain(|warning| seen.insert(warning.clone()));
```
stevencartavia, line 3760: Multi-network runs can emit multiple internal-cheatcode warnings for the same suite because each pass prints its warnings before `merge_outcomes` deduplicates them. I reproduced this with passing default and Tempo tests: shared `setUp()` usage prints the identical warning twice, while using different internal overloads in each pass prints two separate warning blocks. Could we merge the signatures across passes and emit one suite-level warning after merging?

mablr: I reproduced this on the current branch HEAD (https://github.com/foundry-rs/foundry/pull/15808/commits/a7399375c689e813c5a9ae97f06cab8548d9a25b). The warning needs to be deduplicated before each network pass emits it, rather than later in `merge_outcomes`.

0xMars42: Fixed. Suite-level cheatcode warnings (internal and deprecated) are now deduplicated before each pass prints them, rather than in `merge_outcomes`: `run_tests` threads a set of already-printed `(suite, warning)` pairs through the per-pass emission, so a warning shared across passes is printed once. `merge_outcomes` keeps deduplicating the stored outcome for the serialized/JSON output. Covered by `internal_cheatcode_warning_not_duplicated_across_network_passes`: a helper using an internal cheatcode is called from an unannotated default-network test and a Tempo-annotated test, and the warning is asserted to appear once.

crates/evm/symbolic/src/executor/calls.rs
```diff
@@ -776,6 +777,22 @@ impl SymbolicExecutor {
                 }
             }
 
+            if to == CHEATCODE_ADDRESS
```
mablr, line 786: This status check only runs for concrete targets. A symbolic target constrained to `address(vm)` stays on `call_symbolic_target`, bypassing internal recording and removed-cheatcode rejection. Could target resolution consult `state.constrained_word(...)` before dispatch?

crates/forge/src/cmd/test/mod.rs
mablr: The identical-warning case is fixed, but the different-overload case from this thread remains: each pass renders a different warning string, so both blocks are emitted and shared signatures can repeat. We should union structured signatures across passes before formatting the suite warning.

crates/evm/symbolic/src/executor/calls.rs
0xMars42: Done, and the gap was wider than the status check: on the symbolic path a cheatcode never executes at all. `symbolic_call_targets` excludes the cheatcode addresses from the candidate set, so with the default config a target constrained to `address(vm)` dies with the generic unsupported-target failure, and with `symbolic_call_targets = true` it becomes a silent no-op that "succeeds" even for a removed cheatcode.

Target resolution now consults `state.constrained_word(...)` first, before the alias table, the same way `address_or_symbolic_slot` and the call value and in/out sizes already concretize: path constraints are ground truth, so a target they force to a single address takes the concrete path (status policy, handlers, mocks and expectations included) even when a stale alias exists for it. This deliberately covers any constrained target, not just the cheatcode address: a forced target is concrete by definition, and routing it concretely is both more faithful than the modeled outside branch and cheaper than the candidate enumeration. Two regression tests: the internal warning fires for `vm.assume(who == address(vm))` followed by a call, and a stable cheatcode (`deal`) actually executes through a constrained target.

Two limits worth stating: the extraction is the syntactic one `constrained_word` already implements, so a target forced to a single value by arithmetic rather than by direct equality still stays symbolic; and a plain interface call still trips solc's `extcodesize` check before the CALL (the constrained address has no code in the symbolic world), so the tests use a low-level call. The full symbolic suite is unchanged apart from the intended cases.

crates/forge/src/cmd/test/mod.rs
0xMars42: Done: the signatures now live on `SuiteResult` as structured data, `merge_outcomes` unions them and re-renders the suite-level cheatcode warnings from the merged data (deprecated ones re-derive from the merged test results the same way), and the per-pass emission is gone in multi-pass mode: the suite warnings are emitted once from the merged outcome, after the last pass. The printed-strings set from the previous revision is removed; single-pass emission is unchanged, byte for byte.

Regression test: a suite whose default pass observes the shared overload plus a second one, and whose tempo pass observes only the shared one, now emits a single block with the union, each signature exactly once.

Trade-offs of deferring that seem acceptable but worth stating: in multi-pass runs the warnings (including non-cheatcode ones) print after the passes rather than in each suite header, and they are lost if a later pass aborts the run with a hard error, where streaming used to show the earlier passes' warnings before dying. Per-pass emission cannot produce the union, so this looks inherent to formatting after merging. The per-pass JSON documents keep their per-pass strings as before (they are emitted before any merge).

crates/evm/symbolic/src/executor/calls.rs
```diff
@@ -20,7 +21,13 @@ impl SymbolicExecutor {
         }
         let target = state.stack.pop()?;
         ensure_expr_not_gasleft(&target)?;
-        let target_address = state.world.resolve_address(&target);
+        // Path constraints are ground truth for the target: a word they force to a single
+        // value is concrete even when unresolved (or stale) in the alias table, mirroring
+        // address_or_symbolic_slot. Constants short-circuit inside constrained_word.
+        let target_address = state
+            .constrained_word(&mut self.cx, &target)
```
mablr, line 28: **Blocker:** preferring the constrained literal here changes target resolution for every symbolic external call, not just cheatcode calls. CREATE/CREATE2 can install code and storage under a synthetic world alias; if that expression is later constrained to a concrete address, this ordering looks up the concrete address instead and can lose the created account's code/state. Please preserve alias-first resolution and use `constrained_word` as the fallback. If an existing alias later constrained to `address(vm)` must dispatch as a cheatcode, special-case known cheatcode addresses rather than globally overriding aliases.

crates/cheatcodes/spec/src/cheatcode.rs
```diff
@@ -59,6 +59,31 @@ pub enum Status<'a> {
...
+}
+
+impl Status<'_> {
+    /// Returns the runtime action for this status.
+    pub const fn action(&self) -> StatusAction {
+        match self {
+            // Experimental cheatcodes dispatch normally: the status is currently unused and its
+            // runtime behavior has not been decided. Deprecated cheatcodes are recorded
+            // separately, with their replacement.
+            Self::Stable | Self::Experimental | Self::Deprecated(_) => StatusAction::Continue,
```
mablr, line 80: **Blocker:** this makes the supposedly exhaustive runtime-action mapping non-authoritative. `Deprecated` is `Continue` here, concrete dispatch records it in a separate branch, and symbolic dispatch trusts this mapping and silently loses the warning. Please either represent deprecation explicitly (including its replacement) and handle every action in both dispatchers, or narrow/rename this abstraction so it is clearly only the pre-dispatch gate. An exhaustive match over `StatusAction` should not be able to omit a documented status consequence.

crates/forge/src/result.rs
```diff
@@ -887,6 +880,56 @@ impl SuiteResult {
...
 
+pub(crate) const INTERNAL_CHEATCODES_WARNING_HEADER: &str =
+    "the following cheatcode(s) are intended for internal use and may change or be removed:";
+
+pub(crate) const DEPRECATED_CHEATCODES_WARNING_HEADER: &str =
+    "the following cheatcode(s) are deprecated and will be removed in future versions:";
+
+/// Returns `true` for the suite-level cheatcode warnings, which are re-rendered from the
+/// structured data when multi-network passes are merged.
+pub(crate) fn is_suite_cheatcode_warning(warning: &str) -> bool {
```
mablr, line 891: Can this merge structured warning state rather than use rendered English strings as type tags? The current flow renders status warnings in `SuiteResult::new`, recognizes/removes them by prefix during merge, then renders them again. That couples merge correctness to presentation text and duplicates ownership between `result.rs` and the command. Keeping structured status usage through `SuiteResult` merging and rendering once afterward would be simpler and harder to break.

crates/forge/src/cmd/test/mod.rs
```diff
@@ -2046,6 +2046,17 @@ impl TestArgs {
                 merge_outcomes(&mut outcome, pass_outcome);
             }
 
+            // Emit each suite's warnings once from the merged outcome: after the passes
+            // they carry the union of every pass's signatures. JUnit never emitted the
+            // per-pass warnings, keep it silent here too.
+            if !shell::is_json() && !self.junit {
+                for suite_result in outcome.results.values() {
```
mablr, line 2053: These warnings are no longer adjacent to a suite header. If multiple suites emit the same internal-cheatcode warning, the final output contains indistinguishable bare warning blocks and users cannot tell which suite caused either one. Please retain suite attribution when deferring output, or print each merged suite's warnings together with its suite identifier.

crates/evm/symbolic/src/executor/calls.rs
```diff
@@ -776,6 +783,22 @@ impl SymbolicExecutor {
...
+            {
+                match cheatcode.status.action() {
+                    StatusAction::Continue => {}
+                    StatusAction::RecordInternal => {
+                        if let Some(cheatcodes) = &executor.inspector().cheatcodes {
+                            cheatcodes.record_internal_cheatcode(cheatcode.func.signature);
+                        }
+                    }
+                    StatusAction::RejectRemoved => {
+                        return Err(SymbolicError::Unsupported("removed cheatcode"));
```
mablr, line 797: This is materially different from concrete Removed semantics: concrete dispatch returns a cheatcode revert, while `Unsupported` terminates symbolic exploration, so a low-level call or `expectRevert` cannot catch it symbolically. If that distinction is intentional for a "hard error", please document it and test it directly; otherwise this should model the concrete revert path rather than report an unsupported operation.

crates/cheatcodes/src/inspector.rs
```diff
@@ -3192,6 +3242,36 @@ const fn will_exit(action: &InterpreterAction) -> bool {
...
+        });
+        for worker in workers {
+            worker.join().unwrap();
+        }
+
+        assert_eq!(usage.signatures(), ["first()", "second()"]);
+    }
+
+    #[test]
+    fn records_internal_and_rejects_removed_cheatcodes() {
```
mablr, line 3260: The helper unit test does not cover the risky integration boundaries introduced here. Please add focused coverage for: (1) symbolic CREATE/CREATE2 alias preservation after constraining the address, then calling/reading its state; (2) a deprecated cheatcode on a successful symbolic-only path; (3) Removed through the real dispatch gate, including expected error formatting and symbolic semantics; (4) cross-suite isolation in one command; and (5) constructor/setup internal usage across multi-network passes with one attributed merged warning. The alias-preservation case is required before the target-resolution change is safe to merge.

PR #15809, opened 2026-07-17
test(anvil): skip invalid Base fork height

mattsse:
The public Base RPC intermittently reports block zero as its latest height. Anvil then forks the genesis block and the gas-limit assertion fails against its 30 million gas limit even though the current Base block limit is valid.

This returns early when the resolved Base fork height is zero, avoiding the false failure while preserving the gas-limit assertions for valid fork heads.

This change was developed with assistance from OpenAI Codex.


wyattloveslife-create (commented):
Thank you

PR #15810, opened 2026-07-17
fix(script): show storage changes at verbosity five

mattsse:
`forge script -vvvvv` advertised storage changes but script execution and simulation always rendered traces with storage changes disabled, while the runner only requested call traces. This propagates the resolved EVM verbosity into trace requirements, renders state diffs at verbosity five for both phases, and adds CLI snapshot coverage for the behavior.

Codex assisted with diagnosis, code changes, tests, and PR drafting.

### DIFF
--- a/crates/forge/tests/cli/script.rs
+++ b/crates/forge/tests/cli/script.rs
@@ -150,6 +150,49 @@ Script ran successfully.
 "#]]);
 });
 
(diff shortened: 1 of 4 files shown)


stevencartavia (approved):

grandizzy (approved):
lgtm

PR #15812, opened 2026-07-18
fix(anvil): serve pre-fork debug_getRawTransactions

stevencartavia:
In fork mode, `debug_getRawTransactions` only read anvil's in-memory blocks, so pre-fork blocks returned an empty array despite having transactions. It now falls back to the upstream provider like `debug_getRawReceipts` and the trace-block endpoints, while still returning empty for genuinely unknown blocks.


mattsse (approved):
lgtm

grandizzy (approved):
lgtm

PR #15814, opened 2026-07-18
fix(fuzz): preserve stateless transactions

stevencartavia:
Stateless fuzzing generates complete transactions but previously discarded everything except calldata. This meant values such as `msg.value` were not preserved consistently across execution, corpus storage, counterexamples, and replay.

This PR carries a complete `BasicTxDetails` through the stateless fuzzing path. It also keeps existing failure artifacts replayable.

This is PR 1 because a shared transaction representation is the foundation for unifying stateless and invariant fuzzing around sequences. Follow-up PRs will unify transaction generation and dictionaries, sequence generation and mutation, and finally the execution campaign. Those architectural changes are intentionally out of scope here.

Part of OSS-252

### DIFF
--- a/crates/evm/evm/src/executors/corpus.rs
+++ b/crates/evm/evm/src/executors/corpus.rs
@@ -1314,10 +1314,10 @@ impl WorkerCorpus {
         test_runner: &mut TestRunner,
         fuzz_state: &EvmFuzzState,
         function: &Function,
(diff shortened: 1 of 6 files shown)


mablr (approved):
LGTM

mattsse (approved):
lgtm

PR #15763, opened 2026-07-15
feat(fuzz): stack corpus mutator improvements

decofe:
Stacks the changes from:

- #15365
- #15366
- #15367

On
- #15368

Validation:
- `cargo +nightly fmt --all -- --check`
- `git diff --check`

Prompted by: @0xalpharush

PR #15816, opened 2026-07-18
feat(fuzz): persist worker corpus immediately

decofe:
## Summary
- remove live/deferred corpus insertion modes and write interesting entries directly to each worker corpus
- stop carrying corpus artifacts through end-of-run aggregation
- expand configured gzip corpus files before invariant campaigns and recompress eligible entries after workers finish
- keep the aggregated optimization best as the final persisted optimization state

## Validation
- `git diff --check`
- `cargo metadata --no-deps --format-version 1`
- `cargo check -p foundry-evm --lib` could not start because fetching `foundry-rs/optimism` returned HTTP 401

Stacked on #15763. Closes OSS-574

Prompted by: @0xalpharush

### DIFF
--- a/crates/evm/evm/src/executors/corpus.rs
+++ b/crates/evm/evm/src/executors/corpus.rs
@@ -2248,6 +2248,223 @@ mod tests {
         assert_eq!((ok, err), (1, 1), "the corrupt file must read as Err, the valid one as Ok");
     }
 
+    fn empty_targeted_contracts() -> FuzzRunIdentifiedContracts {
+        FuzzRunIdentifiedContracts::new(TargetedContracts::new(), false)
+    }
+
+    #[test]
+    fn invariant_crossover_insert_loads_tx_from_persisted_corpus() {
+        let corpus_root = temp_corpus_dir();
+        let mut config = corpus_config(corpus_root.clone());
+        config.mutation_weights = FuzzCorpusMutationWeights {
+            mutation_weight_splice: 0,
+            mutation_weight_repeat: 0,
+            mutation_weight_interleave: 0,
+            mutation_weight_prefix: 0,
+            mutation_weight_suffix: 0,
+            mutation_weight_abi: 0,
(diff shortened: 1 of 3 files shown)


0xalpharush:
derek bench invariant

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -367,11 +360,63 @@ fn same_tx_sequence(left: &[BasicTxDetails], right: &[BasicTxDetails]) -> bool {
...
+        for entry in read_corpus_dir(&dir) {
+            let is_gzip = entry
+                .path
+                .extension()
+                .and_then(|ext| ext.to_str())
+                .is_some_and(|ext| ext.eq_ignore_ascii_case("gz"));
+            if is_gzip == gzip {
+                continue;
+            }
+            let tx_seq = entry.read_tx_seq()?;
```
mablr, line 449: This makes one corrupt/truncated corpus entry abort campaign preparation or finalization. The normal loader warns and skips unreadable entries; this rewrite should preserve that tolerance (or preferably avoid rewriting the whole corpus at campaign boundaries).

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1078,18 +1057,16 @@ impl WorkerCorpus {
         {
             self.optimization_best_value = Some(value);
             self.optimization_best_sequence = best_seq;
-            if persist_now {
-                self.persist_optimization_state();
-            }
+            self.persist_optimization_state();
```
mablr: All workers can write the same `optimization_best.json` concurrently via truncate-and-write. Overlapping writes or a crash can leave malformed JSON or a lower best. Please keep this coordinator-owned, or use an atomic locked max update.

mablr:
Superseded by #16065, which landed immediate parallel invariant corpus persistence. Closing this obsolete stack layer as part of the corpus PR refresh. AI-assisted maintenance.

PR #15817, opened 2026-07-18
feat(fuzz): keep observed calls out of corpus

decofe:
Keep observed calls in a generation dictionary instead of promoting them directly into the coverage corpus. Sibling-test seed coverage stays separate from campaign coverage so a later fuzz execution can earn normal corpus promotion. This refresh ports the existing proposal directly onto master and retains the persistence behavior from #16065.

The existing unbounded dictionary, per-plan cloning, fixed sampling rate, and dynamic-target lifetime handling remain review questions; this refresh does not redesign them. Derek comparison baseline: master `bfe6e9a17`.

Originally prompted by @0xalpharush. AI assistance was used to refresh the existing code, resolve conflicts, and update this description.

### DIFF
--- a/crates/evm/evm/src/executors/corpus.rs
+++ b/crates/evm/evm/src/executors/corpus.rs
@@ -1365,7 +1365,6 @@ impl WorkerCorpus {
             let targets = targeted_contracts.targets();
             sequence_from_observed(observed, &targets, depth, None)
         };
-
         calls.into_iter().filter(|call| self.sequence_generator.observe_call(call.clone())).count()
     }
 
@@ -1383,6 +1382,12 @@ impl WorkerCorpus {
         }
 
         let mut added = 0;
+        // Test traces may nominate useful observed-call shapes, but they are not campaign
+        // executions. Keep their coverage accounting isolated so a later fuzzed execution can
+        // still discover and persist the same coverage.
+        let mut test_history_map = self.history_map.clone();
+        let mut test_edge_indices = self.edge_indices.clone();
+        let mut test_sancov_history_map = self.sancov_history_map.clone();
 
         for func in invariant_contract.abi.functions() {
             if !func.is_unit_test() {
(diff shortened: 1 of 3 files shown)


0xalpharush:
TBD if this is desired https://github.com/foundry-rs/foundry/pull/15817/commits/46180e7603de9c709ab1771b50476b6371657522

I think also mutating immediately upon selecting an observed call remains

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1651,37 +1664,38 @@ impl WorkerCorpus {
...
-                ObservedCallDepth::All,
-                Some((parent_tx.warp, parent_tx.roll)),
-            )
+            sequence_from_observed(observed, &targets, depth, None)
         };
-
-        self.push_observed_sequence(tx_seq)
+        let mut added = 0;
+        for call in calls {
+            if !self.observed_call_pool.iter().any(|existing| {
```
mablr: This pool is unbounded and each insertion scans/com­pares every retained call, making distinct observed calls O(N^2) time plus unbounded calldata memory. We could use a dedupe set with a bounded reservoir/LRU.

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1614,6 +1620,13 @@ impl WorkerCorpus {
 
     /// Generates single call from corpus strategy.
     pub fn new_tx(&self, test_runner: &mut TestRunner) -> Result<BasicTxDetails> {
+        // Keep synthesizing fresh calls, but regularly use coverage-winning observed shapes as
+        // generation seeds. They are deliberately separate from corpus selection and persistence.
+        if !self.observed_call_pool.is_empty() && test_runner.rng().random_ratio(1, 2) {
```
mablr: Calls to run-local created contracts remain in this pool after those targets are cleared, so this fixed 50% path can increasingly select dead addresses. We could prune/tag dynamic-target entries at run end (and consider making this rate weighted rather than fixed).

PR #15818, opened 2026-07-18
feat(fuzz): prioritize favored corpus entries

decofe:
Prioritize mutation donors using the existing favored-entry, edge-rarity, and recent-yield schedule. This refresh places scheduling directly on #15368, without the observed-call dictionary or additional mutation operators. Applicable corpus and synchronization tests remain enabled.

Compare Derek results against the exact #15368 head `e6845bc79` to isolate scheduling. The existing expanded weight vector can grow quadratically, the recent-yield bonus retains a floor, and selection remains limited to the in-memory donor pool. Those design questions are deliberately unchanged by this refresh and remain open for review.

Originally prompted by @0xalpharush. AI assistance was used for the refresh, conflict resolution, test adaptation, and this description.

### DIFF
--- a/crates/evm/evm/src/executors/corpus.rs
+++ b/crates/evm/evm/src/executors/corpus.rs
@@ -64,6 +64,7 @@ use proptest::test_runner::TestRng;
 use proptest::test_runner::TestRunner;
 #[cfg(test)]
 use rand::Rng;
+use rustc_hash::FxHashMap;
 use serde::{Deserialize, Serialize};
 use std::{
     collections::{HashMap, HashSet, VecDeque},
@@ -1259,7 +1260,7 @@ impl WorkerCorpus {
     /// Builds an AFL-style weighted schedule favoring minset entries, rare edges, and recently
     /// productive inputs. Every entry retains base energy so the rest of the corpus is not starved.
     fn mutation_schedule(&self) -> Vec<usize> {
-        let mut edge_frequency = HashMap::<usize, usize>::new();
+        let mut edge_frequency = FxHashMap::<usize, usize>::default();
         for corpus in &self.in_memory_corpus {
             for &edge in &corpus.unique_edges_covered {
                 *edge_frequency.entry(edge).or_default() += 1;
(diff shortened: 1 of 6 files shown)


mablr (commented):
We should probably build the mutation schedule once per `new_inputs` call and reuse it for both donor picks, rather than rebuilding edge frequencies and weights twice.

The edge-frequency map also uses trusted integer keys, so switching it to `FxHashMap` (or avoiding hashing entirely) appears worthwhile.

In the same fixed-seed Nerite local run, snapshot reuse improved throughput from 604 to 988 tx/s, and snapshot reuse plus FxHashMap reached 2,867 tx/s.

mablr:
derek bench fuzz compare-ref=master timeout=600

grandizzy:
derek bench invariant timeout=10800

mablr:

mablr:

mablr:
derek bench invariant timeout=600 trials=6 paired-seeds=true

grandizzy:
derek bench invariant timeout=600 trials=6 paired-seeds=true

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1238,6 +1257,44 @@ impl WorkerCorpus {
...
+        self.in_memory_corpus
+            .iter()
+            .enumerate()
+            .flat_map(|(index, corpus)| {
+                std::iter::repeat_n(index, self.mutation_energy(corpus, &edge_frequency) as usize)
+            })
+            .collect()
+    }
+
+    fn mutation_energy(
```
0xalpharush, line 1279: maybe worth looking at https://github.com/AFLplusplus/LibAFL/blob/de0b34d2c561256e8d241af1a42851ea8803ab22/crates/libafl/src/schedulers/testcase_score.rs#L53-L267

and the metadata tracking

mablr: Would it make sense to start narrowly with favored status and per-entry selection/productivity counts? These directly inform donor selection using metadata Foundry can already track, while execution-time calibration and full power scheduling would require broader fuzzing-lifecycle changes.

mablr:
derek bench invariant

PR #15825, opened 2026-07-20
test(fmt): cover nested named call chains

decofe:
#15911 now provides the generalized method-chain layout that also covers this case, so this branch no longer carries a separate width estimator. The focused fixture keeps regression coverage for nested named calls with overlong identifier, boolean, and binary values, the 120-column boundary, a short inline case, and a trailing comment.

Updated with AI assistance.


DaniPopes (dismissed):

crates/fmt/src/state/sol.rs
```diff
@@ -1394,6 +1394,9 @@ impl<'ast> State<'_, 'ast> {
                             _ if s.chained_named_call.is_some_and(|call| {
                                 call.keep_inline && call.callee.contains(expr.span)
                             }) => {}
+                            _ if s.call_has_overlong_named_call_arg(&member_expr.kind) => {
+                                s.hardbreak()
```
mablr: `print_trailing_comment` may already emit a hard break for a trailing `//` comment. Calling `hardbreak()` again here can introduce a blank line before the next member. Could this use `hardbreak_if_not_bol()` and add a trailing-comment regression case?

crates/fmt/src/state/sol.rs
```diff
@@ -2851,6 +2854,28 @@ impl<'ast> State<'_, 'ast> {
...
+        let ast::ExprKind::Call(callee, args) = &expr.kind else { return None };
+        let ast::CallArgsKind::Named(args) = &args.kind else { return None };
+        let args_size = args.iter().try_fold(0usize, |size, arg| {
+            Some(size + arg.name.as_str().len() + 2 + self.estimate_call_chain_size(arg.value)?)
+        })?;
+        Some(
+            self.estimate_call_chain_size(callee)?
+                + args_size
+                + args.len().saturating_sub(1) * 2
+                + 2,
```
mablr: This seems to undercount the flat width by two characters: `callee({args})` has four delimiters (`(`, `{`, `}`, `)`), while `+ 2` accounts for only two. That misclassifies calls whose rendered width is one or two columns over `line_length`.

crates/fmt/src/state/sol.rs
decofe: Fixed in aa6a82f: this now uses `hardbreak_if_not_bol()`, with a trailing `//` comment regression case to ensure no blank line is introduced.

crates/fmt/src/state/sol.rs
decofe: Fixed in aa6a82f: the estimate now counts all four `({})` delimiters, with a 121-column boundary fixture that the previous `+ 2` misclassified.

crates/fmt/src/state/sol.rs
```diff
@@ -2851,6 +2854,28 @@ impl<'ast> State<'_, 'ast> {
         }
     }
 
+    fn estimate_named_call_size(&self, expr: &ast::Expr<'_>) -> Option<usize> {
+        let ast::ExprKind::Call(callee, args) = &expr.kind else { return None };
+        let ast::CallArgsKind::Named(args) = &args.kind else { return None };
+        let args_size = args.iter().try_fold(0usize, |size, arg| {
+            Some(size + arg.name.as_str().len() + 2 + self.estimate_call_chain_size(arg.value)?)
```
stevencartavia: At `line_length = 120`, overlong named calls with boolean or arithmetic values leave the following chain link attached as `).update()`, while calls with the same callee and named-argument names but identifier values place `.update()` on a new line. `estimate_call_chain_size` returns `None` for boolean literals and binary expressions, so `estimate_named_call_size` cannot detect that these calls are overlong. Could these value kinds participate in the width check as well?

decofe: Fixed in 9bf59ea: boolean literals and binary expressions now participate in the flat-width estimate, with regression cases for both argument value kinds.

crates/fmt/testdata/NestedNamedCallArgumentChain/fmt.sol
```diff
@@ -0,0 +1,43 @@
...
+                })
+            ) // keep this comment
+            .update();
+        factory().item(innerFactory().method({value: firstExtremelyLongValueName})).update();
+        factory()
+            .item(
+                innerFactory().anExtremelyLongMethodNameThatForcesTheNestedCalleeToWrap({
+                    firstExtremelyLongArgumentName: true
+                })
+            )
```
mablr, line 33: This snapshot currently fails: the formatter emits `).update();`. This nested call is only 111 characters, so it does not trigger the `> 120` rule.

crates/fmt/src/state/sol.rs
```diff
@@ -2881,6 +2890,17 @@ impl<'ast> State<'_, 'ast> {
...
+            ast::ExprKind::Lit(lit, None) => match lit.kind {
+                ast::LitKind::Bool(true) => Some(4),
+                ast::LitKind::Bool(false) => Some(5),
+                _ => None,
+            },
+            ast::ExprKind::Binary(lhs, op, rhs) => Some(
+                self.estimate_call_chain_size(lhs)?
+                    + op.to_string().len()
+                    + self.estimate_call_chain_size(rhs)?
+                    + 2,
```
mablr: This always counts two operator spaces, but `pow_no_space = true` emits `**` without spaces.

crates/fmt/src/state/sol.rs
```diff
@@ -2891,6 +2911,28 @@ impl<'ast> State<'_, 'ast> {
...
+        let ast::CallArgsKind::Named(args) = &args.kind else { return None };
+        let args_size = args.iter().try_fold(0usize, |size, arg| {
+            Some(size + arg.name.as_str().len() + 2 + self.estimate_call_chain_size(arg.value)?)
+        })?;
+        Some(
+            self.estimate_call_chain_size(callee)?
+                + args_size
+                + args.len().saturating_sub(1) * 2
+                + 4,
+        )
```
mablr: Should we include configured brace spacing in the estimate?

crates/fmt/src/state/sol.rs
figtracer: Superseded in f8aafe8c: after merging #15911, this behavior comes from the generalized consistent-chain layout. The boolean and binary regression cases remain, while the width-estimator extensions have been removed.

crates/fmt/testdata/NestedNamedCallArgumentChain/fmt.sol
figtracer: Addressed in f8aafe8c: merging master picks up #15911’s consistent-chain layout, and the expected `.update()` break now matches formatter output. The focused fixture passes.

crates/fmt/src/state/sol.rs
figtracer: Addressed in f8aafe8c: #15911 now owns the chain-break behavior, so the PR-specific named-call width estimator was removed. There is no remaining brace-spacing estimate to drift from configuration.

crates/fmt/src/state/sol.rs
figtracer: Addressed in f8aafe8c: #15911 supersedes the PR-specific overlong-width branch, so the named-call estimator and its boolean/binary extensions were removed rather than duplicating layout logic. `pow_no_space` is therefore no longer approximated here.

mattsse (approved):

stevencartavia (approved):

PR #15828, opened 2026-07-20
fix(anvil): enforce `eth_simulateV1` request limits

mablr:
## Motivation

Close OSS-566.

This is the next focused `eth_simulateV1` conformance step after #15770, #15784, and #15792, continuing the split of the broader work from #15716. It rejects empty requests, enforces the 256-block limit with `-38026`, and returns `-32000 header not found` for unavailable base blocks.

The simulation now also applies a 50M request-wide gas budget across blocks. Each call is capped by the remaining budget, which is reduced by the gas consumed by every execution, including reverted calls. Block sequencing and derived block context remain scoped to the follow-up work.


mattsse (approved):
lgtm

PR #15830, opened 2026-07-20
fix(debugger): cycle breakpoint hits

figtracer:
Repeated executions of the same `vm.breakpoint` PC are recorded as separate trace hits, particularly in loops and across parent segments split around calls. The debugger currently scans from the start of the arena for every breakpoint jump, so invoking the shortcut again remains at the first hit and later iterations are unreachable through the breakpoint.

This searches forward from the current trace position before wrapping to the earliest match. Repeated `'<key>` shortcuts can therefore visit every recorded hit while preserving the existing address/PC matching, missing-target errors, and single-hit behavior. The focused regression covers later hits in the current segment, a later split segment, and wraparound.

Progresses #927.

### DIFF
--- a/crates/debugger/src/tui/context.rs
+++ b/crates/debugger/src/tui/context.rs
@@ -945,15 +945,12 @@ impl TUIContext<'_> {
             return;
         };
 
-        // Find the location of the called breakpoint in the whole debug arena (at this address with
-        // this pc)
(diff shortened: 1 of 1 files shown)


mattsse (approved):
lgtm

mablr (approved):

stevencartavia (approved):

PR #9427, opened 2024-11-28
feat(`forge`): `--profile` to set the profile via CLI

yash-atreya:
## Motivation

Closes #5479

## Solution

Introduces the `--profile` flag in `CoreBuildArgs`

PR #15831, opened 2026-07-20
feat(cli): add --profile flag

### LINKED ISSUE #5479: feature request: allow specifying profile through command-line argument
### Component

Forge

Allow specifying profile via `forge --profile profile_name` 

Right now profiles are useful for dividing different compiler settings for different development processes. Sometimes development process may require frequent switching between profiles, and it's not very convinient to edit .env every time. Also, passing profile through command-line arg would allow always keeping it in focus and will help to not forget switch profile when needed

### Additional context

_No response_

figtracer:
Adds a global `--profile` option that initializes Foundry’s selected configuration profile before command-specific configuration is loaded. The command-line value overrides `FOUNDRY_PROFILE`, so `forge --profile ci ...` and the same selector on other Foundry tools consistently use the requested profile. This follows the global-argument direction discussed in #9427 and closes #5479.

OpenAI Codex was used to investigate the issue, generate the implementation and test, and prepare this pull request.

### DIFF
--- a/crates/cli/src/opts/global.rs
+++ b/crates/cli/src/opts/global.rs
@@ -3,6 +3,7 @@ use foundry_common::{
     shell::{ColorChoice, OutputFormat, OutputMode, Shell, Verbosity},
     version::{IS_NIGHTLY_VERSION, NIGHTLY_VERSION_WARNING_MESSAGE},
 };
+use foundry_config::{Config, figment::Profile};
 use serde::{Deserialize, Serialize};
 
 /// Global arguments for the CLI.
@@ -48,6 +49,10 @@ pub struct GlobalArgs {
     /// Number of threads to use. Specifying 0 defaults to the number of logical cores.
     #[arg(global = true, long, short = 'j', visible_alias = "jobs")]
     threads: Option<usize>,
+
+    /// The configuration profile to use.
+    #[arg(global = true, long, value_name = "PROFILE")]
+    profile: Option<Profile>,
 }
 
 impl GlobalArgs {
@@ -65,6 +70,14 @@ impl GlobalArgs {
 
     /// Initialize the global options.
     pub fn init(&self) -> eyre::Result<()> {
+        if let Some(profile) = &self.profile {
+            // SAFETY: Global options are initialized before any worker threads are spawned.
+            unsafe {
+                std::env::set_var("FOUNDRY_PROFILE", profile.to_string());
+            }
+        }
+        let _ = Config::selected_profile();
+
         // Set the global shell.
         let shell = self.shell();
         // Argument takes precedence over the env var global color choice.
--- a/crates/cli/src/opts/global.rs
+++ b/crates/cli/src/opts/global.rs
@@ -70,11 +70,13 @@ impl GlobalArgs {
 
     /// Initialize the global options.
     pub fn init(&self) -> eyre::Result<()> {
-        if let Some(profile) = &self.profile {
-            // SAFETY: Global options are initialized before any worker threads are spawned.
-            unsafe {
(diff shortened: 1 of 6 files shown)


crates/cli/src/opts/global.rs
```diff
@@ -65,6 +70,14 @@ impl GlobalArgs {
 
     /// Initialize the global options.
     pub fn init(&self) -> eyre::Result<()> {
+        if let Some(profile) = &self.profile {
+            // SAFETY: Global options are initialized before any worker threads are spawned.
+            unsafe {
+                std::env::set_var("FOUNDRY_PROFILE", profile.to_string());
+            }
```
mattsse: `Profile::from_env_or` matches environment variable names case-insensitively, so on Unix an existing casing variant (for example `foundry_profile=default`) can coexist with the uppercase key and win the lookup. I reproduced `foundry_profile=default forge --profile ci config` loading `default`. Remove all accepted aliases before setting the CLI override, and add a regression test with a differently cased key.

```suggestion
        if let Some(profile) = &self.profile {
            let profile_keys = std::env::vars_os()
                .filter_map(|(key, _)| {
                    key.to_string_lossy()
                        .eq_ignore_ascii_case("FOUNDRY_PROFILE")
                        .then_some(key)
                })
                .collect::<Vec<_>>();
            // SAFETY: Global options are initialized before any worker threads are spawned.
            unsafe {
                for key in profile_keys {
                    std::env::remove_var(key);
                }
                std::env::set_var("FOUNDRY_PROFILE", profile.to_string());
            }
        }
```

mattsse (commented):
Found one profile-precedence edge case.

crates/cli/src/opts/global.rs
```diff
@@ -65,6 +70,14 @@ impl GlobalArgs {
 
     /// Initialize the global options.
     pub fn init(&self) -> eyre::Result<()> {
+        if let Some(profile) = &self.profile {
+            // SAFETY: Global options are initialized before any worker threads are spawned.
+            unsafe {
+                std::env::set_var("FOUNDRY_PROFILE", profile.to_string());
```
mablr: This `set_var` safety invariant is not upheld for every supported build. With the `tracy` feature, `setup()` constructs `TracyLayer::default()` before `global.init()`, which starts the Tracy client; `tracy-allocator` can start it on the first allocation even earlier. The process may therefore already be multithreaded here, making environment mutation unsound on Unix. The suggested case-insensitive-key cleanup would add more unsafe mutations rather than fix this. Shoule be initialized directly in `foundry-config`?

crates/cli/src/opts/global.rs
stevencartavia: +1 from my pass

stevencartavia (dismissed):

crates/cli/src/opts/global.rs
```diff
@@ -65,6 +70,16 @@ impl GlobalArgs {
 
     /// Initialize the global options.
     pub fn init(&self) -> eyre::Result<()> {
+        if let Some(profile) = &self.profile
+            && let Err(selected) = Config::try_set_selected_profile(profile.clone())
```
stevencartavia, line 74: The CLI-selected `--profile` is now process-local, so the outer `cast wallet session` process does not inherit it. I reproduced `FOUNDRY_PROFILE=default forge --profile ci script ... --session` sending the `default` profile’s RPC headers from Cast, while the inner Forge command retained `--profile ci`. Please pass `--profile Config::selected_profile()` to the outer Cast argv so both processes use the same profile.

mablr (commented):
SGTM

mablr (approved):

stevencartavia (approved):

PR #15833, opened 2026-07-20
feat(forge): clone proxy implementations

### LINKED ISSUE #15829: feat: Add implementation support for forge clone
### Component

Forge

`forge clone 0xC02aB1A5eaA8d1B114EF786D9bde108cD4364359 sparklend-usds` will pull just the proxy code, but would be much more useful to be able to pull the implementation code as well, maybe with a flag?

### Additional context

_No response_

mattsse:
Closes #15829.

`forge clone --implementation` now follows Etherscan proxy metadata recursively and uses the resolved implementation for downloaded sources, creation data, storage layout, and `.clone.meta`, while preserving the existing default behavior. SparkLend’s verified `AToken` sources live entirely under `lib`, so cloning also creates a minimal `src/Clone.sol` import entrypoint when necessary to keep library-only source trees buildable. Regression coverage includes proxy-chain and cycle cases plus a live end-to-end clone of the SparkLend USDS proxy.

This PR was written primarily with Codex.

### DIFF
--- a/crates/forge/src/cmd/clone.rs
+++ b/crates/forge/src/cmd/clone.rs
@@ -29,7 +29,7 @@ use path_slash::PathExt;
 use reqwest::StatusCode;
 use serde::Deserialize;
 use std::{
-    collections::{BTreeMap, HashMap, HashSet},
+    collections::{BTreeMap, HashMap},
     fs::read_dir,
     path::{Path, PathBuf},
     time::Duration,
@@ -119,8 +119,6 @@ pub struct CloneArgs {
     pub sourcify_url: Option<String>,
 
     /// Clone the implementation contract if the address is a proxy.
-    ///
-    /// If the implementation is also a proxy, its implementation is resolved recursively.
     #[arg(long)]
     pub implementation: bool,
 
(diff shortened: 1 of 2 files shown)


crates/forge/src/cmd/clone.rs
```diff
@@ -635,6 +671,32 @@ fn dump_sources(meta: &Metadata, root: &PathBuf, no_reorg: bool) -> Result<Vec<R
...
+        .filter(|path| !path.starts_with(&src) && !path.starts_with(&forge_std))
+        .collect::<Vec<_>>();
+    eyre::ensure!(!sources.is_empty(), "no Solidity sources found in cloned contract");
+    sources.sort_unstable();
+
+    let imports = sources
+        .into_iter()
+        .map(|path| {
+            let path = path.strip_prefix(root)?;
+            Ok(format!("import \"../{}\";", path.to_slash_lossy()))
```
stevencartavia: Could these use namespace imports with a deterministic unique alias per source? Plain imports introduce every imported source unit’s global symbols into `Clone.sol`’s scope, so otherwise-independent files with duplicate top-level names can fail with `Identifier already declared`.

```solidity
import * as CloneSource0 from "../lib/a/Token.sol";
```

stevencartavia:
I tested this against several real proxies. Resolving only the direct implementation keeps the PR goal intact and correctly clones SparkLend, USDC, Lido, Aave V3, and Compound III; recursive resolution could incorrectly follow unrelated delegatecall targets.

crates/forge/src/cmd/clone.rs
```diff
@@ -224,12 +236,30 @@ impl CloneArgs {
...
-    ) -> Result<Metadata> {
+        implementation: bool,
+    ) -> Result<(Address, Metadata)> {
         let mut meta = client.contract_source_code(address).await?;
         eyre::ensure!(meta.items.len() == 1, "contract not found or ill-formed");
         let meta = meta.items.remove(0);
         eyre::ensure!(!meta.is_vyper(), "Vyper contracts are not supported");
-        Ok(meta)
+
+        if !implementation || meta.proxy == 0 {
```
mablr, line 245: Could we defer the Vyper check above until this branch determines which metadata will actually be cloned? With `--implementation`, a verified Vyper proxy that points to a Solidity implementation currently fails before we fetch the supported implementation, even though the wrapper source is never compiled. Moving the initial `is_vyper()` check into this return branch, while retaining the check on the fetched implementation below, handles both targets correctly. A mock case with a Vyper proxy and Solidity implementation would cover it.

stevencartavia: [143bfb5](https://github.com/foundry-rs/foundry/pull/15833/commits/143bfb5462e0c78f304ef5a183b4b38bc3bf402e)

crates/forge/src/cmd/clone.rs
```diff
@@ -635,6 +666,33 @@ fn dump_sources(meta: &Metadata, root: &PathBuf, no_reorg: bool) -> Result<Vec<R
     Ok(remappings.into_iter().map(|r| r.into_relative(root)).collect())
 }
 
+/// Ensure sources moved entirely into `lib` remain buildable from the project's `src` directory.
+fn ensure_source_entrypoint(root: &Path) -> Result<()> {
+    let src = root.join("src");
+    if fs::files_with_ext(&src, "sol").next().is_some() {
```
mablr: Could we avoid returning just because *some* Solidity source exists under `src`? `dump_sources` can split one verified compiler input between `src` and `lib` (for example, `src/Unrelated.sol` plus a target at `lib/dependency/Implementation.sol`). Foundry seeds compilation from `src`/`test`/`script` and only discovers library files through imports, so the target can remain uncompiled and `find_main_contract` then fails with `contract not found`. Please generate an entrypoint whenever cloned sources outside `src` need importing, and add a mixed-tree test where the target exists only under `lib`.

crates/forge/src/cmd/clone.rs
```diff
@@ -635,6 +666,33 @@ fn dump_sources(meta: &Metadata, root: &PathBuf, no_reorg: bool) -> Result<Vec<R
...
+        .collect::<Vec<_>>();
+    eyre::ensure!(!sources.is_empty(), "no Solidity sources found in cloned contract");
+    sources.sort_unstable();
+
+    let imports = sources
+        .into_iter()
+        .enumerate()
+        .map(|(index, path)| {
+            let path = path.strip_prefix(root)?;
+            Ok(format!("import * as CloneSource{index} from \"../{}\";", path.to_slash_lossy()))
```
mablr: Can we validate or Solidity-escape this path before interpolating it into an import string? Explorer source-unit names are untrusted and can contain quotes, backslashes, or control characters that are valid in Unix filenames but malformed or differently interpreted in a Solidity string literal. `to_slash_lossy()` only normalizes separators. Conservatively rejecting those characters with an actionable error would be sufficient; a focused path test would cover it.

crates/forge/src/cmd/clone.rs
stevencartavia: [df66237](https://github.com/foundry-rs/foundry/pull/15833/commits/df6623789dbafb900b95fe593a9df6b2d7664efa)

crates/forge/src/cmd/clone.rs
stevencartavia: [df66237](https://github.com/foundry-rs/foundry/pull/15833/commits/df6623789dbafb900b95fe593a9df6b2d7664efa)

mablr (approved):

Issue #15823, opened 2026-07-20
Weird indentation on chained functions with parameters


sakulstra:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What command(s) is the bug in?

forge fmt

### Operating System

macOS (Apple Silicon)

### Describe the bug

I "think", [this pr](https://github.com/foundry-rs/foundry/pull/15758) changed some other things unintend.
Having now an issue similar to what was reported here: https://github.com/foundry-rs/foundry/issues/7549 last year (did did not exist in last weeks nightly).

```
    RedactedRootConfiguration.getRedactedEngine({
      redactedId: redactedId
    }).initializeSetRedacted({
      setId_: setId, redactedId_: redactedId, redactedEngineData_: redactedEngineData
    });
```
became
```
    RedactedRootConfiguration.getRedactedEngine({
        redactedId: redactedId
      }).initializeSetRedacted({
        setId_: setId, redactedId_: redactedId, redactedEngineData_: redactedEngineData
      });
```

DaniPopes:
https://github.com/foundry-rs/foundry/pull/15824

PR #15834, opened 2026-07-20
fix(fmt): preserve nested call indentation

mablr:
## Motivation

#15824 fixed the double indentation reported in #15823 by tracking whether a chained named call was nested. That ancestry-based classification is too coarse: it can suppress required indentation when the immediate callee suffix wraps, while retaining duplicate indentation for multiline outer named calls.

This follow-up bases the decision on indentation ownership instead. Named arguments suppress their own indentation only when the parent chain fully owns it and the immediate callee suffix cannot introduce another break. The handling covers member and index suffixes, transparent parentheses, and breakable comments (including comments adjacent to span boundaries) while preserving inline field access after named calls.

### DIFF
--- a/crates/fmt/src/state/mod.rs
+++ b/crates/fmt/src/state/mod.rs
@@ -100,7 +100,6 @@ impl CallStack {
 struct ChainedNamedCall {
     callee: Span,
     keep_inline: bool,
-    nested: bool,
 }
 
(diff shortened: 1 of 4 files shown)


mattsse (approved):
lgtm

DaniPopes (approved):

PR #15835, opened 2026-07-20
refactor(fuzz): unify fuzz input pipeline

stevencartavia:
Stacked on #15814.

This PR moves transaction and sequence generation into `foundry-evm-fuzz` behind concrete `TxGenerator`, `FuzzState`, `SequenceGenerator`, and `SequencePlan` types. Stateless fuzzing uses a fixed depth-one plan, while invariant fuzzing retains isolated worker feedback, lazy target discovery, and lazy sequence continuation.

The existing execution loops remain separate. Corpus persistence and replay stay in the executor crate, while sequence and ABI/CMP mutation move into the fuzz crate. There is no implicit funding behavior or on-disk format change.

This is the second staged PR because the shared input pipeline must exist before execution campaigns can be unified. Commit, check, replay, shrinking, scheduling, and result semantics remain deferred to PR 3.

Part of OSS-252


mattsse (dismissed):
lgtm

mablr (dismissed):
LGTM.

Persistent CMP dictionary values now correctly survive between runs, maybe it would worth to benchmark.

mablr:
Should we run bench here finally ?

stevencartavia:
derek bench fuzz compare-ref=master timeout=600

mattsse (dismissed):
lgtm

mablr (dismissed):

mablr (approved):

mattsse (approved):
lgtm

figtracer (approved):

PR #15838, opened 2026-07-21
fix(forge): preserve revert data when tracing

### LINKED ISSUE #11344: bug(forge test): different revertData depending on verbosity
### Component

Forge

- [x] Foundry
- [x] Foundryup

1.3.1-stable

0.3.1

"forge test -v" versus "forge test -vvvv"

### Operating System

macOS (Apple Silicon)

### Describe the bug
Issue was introduced in https://github.com/foundry-rs/foundry/pull/10446

To reproduce:

```
import { Test } from "forge-std/Test.sol";

contract EOA {
    function foo() external { }
}

contract Mock {
    function callEOA(address eoa) external {
        EOA(eoa).foo();
    }
}

contract Issue is Test {
    function testIssue(address eoa) public {
        Mock mock = new Mock();
        vm.assume(eoa.code.length == 0);
        assumeNotPrecompile(eoa);

        vm.expectRevert(abi.encodePacked("call to non-contract address ", vm.toString(eoa)));
        mock.callEOA(eoa);
    }
}
```

Image

stevencartavia:
Keeps inferred non-contract call diagnostics as trace-only metadata instead of injecting them into EVM revert data. This makes `vm.expectRevert` behavior consistent across verbosity levels while preserving helpful diagnostics in verbose traces.

Fixes #11344.

### DIFF
--- a/crates/evm/traces/src/lib.rs
+++ b/crates/evm/traces/src/lib.rs
@@ -21,10 +21,11 @@ use serde::{Deserialize, Serialize};
 use std::{
     borrow::Cow,
     collections::{BTreeMap, BTreeSet},
+    fmt,
     ops::{Deref, DerefMut},
 };
 
-use alloy_primitives::{U256, map::HashMap};
+use alloy_primitives::{Address, U256, map::HashMap};
 use tempo_contracts::precompiles::TIP20_CHANNEL_RESERVE_ADDRESS;
 
 pub use revm_inspectors::tracing::{
@@ -55,6 +56,27 @@ pub mod speedscope;
 
 pub type Traces = Vec<(TraceKind, SparsedTraceArena)>;
 
+/// Presentation-only detail for an otherwise empty EVM revert.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub enum RevertDiagnostic {
+    /// A call targeted an address without code.
+    CallToNonContract(Address),
+    /// A delegate call targeted an address without code.
+    DelegateCallToNonContract(Address),
+}
+
+impl fmt::Display for RevertDiagnostic {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::CallToNonContract(addr) => write!(f, "call to non-contract address {addr}"),
+            Self::DelegateCallToNonContract(addr) => write!(
+                f,
+                "delegatecall to non-contract address {addr} (usually an unliked library)"
+            ),
+        }
+    }
+}
+
 /// Trace arena keeping track of ignored trace items.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct SparsedTraceArena {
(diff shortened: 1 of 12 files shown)


crates/evm/traces/src/lib.rs
```diff
@@ -65,12 +87,15 @@ pub struct SparsedTraceArena {
     /// See `foundry_cheatcodes::utils::IgnoredTraces` for more information.
     #[serde(default, skip_serializing_if = "HashMap::is_empty")]
     pub ignored: HashMap<(usize, usize), (usize, usize)>,
+    /// Presentation-only revert diagnostics, keyed by trace node index.
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub diagnostics: HashMap<usize, RevertDiagnostic>,
```
mablr, line 92: Could we keep this presentation-only sidecar out of direct serialization and resolve it for the JSON presentation path instead? `forge test --json` and `forge script --json` serialize `SparsedTraceArena` directly, bypassing `resolve_arena()`. As written, their output gains an internal `diagnostics` property while `decoded.return_data` remains unresolved, so JSON disagrees with the text renderer and changes the machine-readable schema. A regression test for a non-contract call under `forge test --json` would catch both issues.

stevencartavia: [9ceebb4](https://github.com/foundry-rs/foundry/pull/15838/commits/9ceebb49a8abe802c12baa3fb37e52d8513ca0ec)

crates/evm/traces/src/lib.rs
```diff
@@ -65,12 +87,32 @@ pub struct SparsedTraceArena {
...
+    {
+        #[derive(Serialize)]
+        struct ResolvedArena<'a> {
+            #[serde(flatten)]
+            arena: &'a CallTraceArena,
+            #[serde(skip_serializing_if = "HashMap::is_empty")]
+            ignored: &'a HashMap<(usize, usize), (usize, usize)>,
+        }
+
+        ResolvedArena { arena: &self.resolve_arena(), ignored: &self.ignored }.serialize(serializer)
```
mablr: Could we avoid serializing the fully resolved arena together with the original `ignored` ranges? `resolve_arena()` has already removed the ignored ordering entries, but the serialized `ignored` offsets still refer to the raw arena. After deserializing, rendering or serializing again applies those stale ranges a second time and can hide unrelated entries. This also changes the previous wire representation for every `vm.pauseTracing()` trace, even when `diagnostics` is empty.

The smallest compatibility-preserving fix seems to be cloning the raw arena and applying only `diagnostics` for serialization, while retaining `ignored` for consumers to resolve once. Alternatively, if this is intentionally a fully resolved presentation representation, `ignored` should be omitted. Could we also cover serialize -> deserialize -> render with both an ignored range and a diagnostic?

stevencartavia: [afbf6aa](https://github.com/foundry-rs/foundry/pull/15838/commits/afbf6aa32989693a3615eb2bd47d438a1505f3d3)

mablr (approved):

figtracer (approved):

PR #15839, opened 2026-07-21
fix(forge): align test and script library linking

### LINKED ISSUE #10448: External function visibility inside a library produces different creation code during testing
### Component

Forge

- [x] Foundry
- [x] Foundryup

forge Version: 1.1.0-stable Commit SHA: d484a00089d789a19e2e43e63bbb3f1500eb2cbf Build Timestamp: 2025-04-30T13:50:49.971365000Z (1746021049) Build Profile: maxperf

foundryup: 1.0.1

forge test

### Operating System

macOS (Apple Silicon)

### Describe the bug

We discovered that the creation code produced inside forge tests is different than in forge scripts for some of the contracts despite following the standard recommended guidelines to ensure matching bytecodes in such scenarios. 

We have narrowed it down to a compiler interpretation of function visibility inside a library that is called from within the main contract.

If a library function is either external or public, the creationCode generated will differ between forge script and forge test. Changing it to internal fixes the issue, probably because the compiler would embed it inside the main contract unlike using delegatecall for external functions.

This is important because different bytecode will produce different create2 addresses and some of the tests will yield incorrect results while others might not be possible to test at all.

We created a simple example to reproduce the issue using the hello-foundry starter.

Link: [https://github.com/chrisciszak/foundry_compiler_issue/tree/main](url)

### Steps to reproduce
1. Clone [https://github.com/chrisciszak/foundry_compiler_issue/tree/main](url)
2. Run `forge test test/Bytecode.t.sol`
3. The test fails
4. Run `forge script test/Bytecode.t.sol`
5. The script passes

Any prompt help or workaround would be highly appreciated as we run out of ideas how to ensure compatibility in this scenario.

stevencartavia:
Aligns Forge tests with scripts by using the same configured CREATE2 library deployer, salt, and linked initcode when available, while retaining nonce-based fallback. This ensures creation-code hashes and downstream CREATE2 addresses match between `forge test` and `forge script`.

This intentionally changes default test-linked library addresses and creation-code hashes. Closes #10448.


crates/linking/src/lib.rs
```diff
@@ -240,14 +240,16 @@ impl<'a> Linker<'a> {
...
-        target: &'a ArtifactId,
+        targets: impl IntoIterator<Item = &'a ArtifactId>,
     ) -> Result<LinkOutput, LinkerError> {
         // Library paths in `link_references` keys are always stripped, so we have to strip
         // user-provided paths to be able to match them correctly.
         let mut libraries = libraries.with_stripped_file_prefixes(self.root.as_path());
 
         let mut needed_libraries = BTreeSet::new();
-        self.collect_dependencies(target, &mut needed_libraries)?;
+        for target in targets {
```
mablr, line 327: Collecting dependencies across every target can add multiple compiler-version-specific artifacts for the same source/name, but the eventual `Libraries` key drops the artifact version (`convert_artifact_id_to_lib_path` keeps only source/name). Both versions can then be deployed at different CREATE2 addresses and the final map insertion makes the last one win; `get_linked_artifacts_cow` applies that address to every version-specific artifact. Could we detect distinct artifact IDs that collapse to the same library key and return an error so the builder falls back to nonce linking? Please add a multi-version regression test as well.

crates/forge/src/runner.rs
```diff
@@ -733,27 +733,81 @@ impl<'a, FEN: FoundryEvmNetwork> ContractRunner<'a, FEN> {
...
+        }
 
-            let (raw, reason) = RawCallResult::from_evm_result(deploy_result.map(Into::into))?;
-            result.extend(raw, TraceKind::Deployment);
-            if reason.is_some() {
-                debug!(?reason, "deployment of library failed");
-                result.reason = reason;
-                return Ok(result);
+        // Configured libraries may already exist and are not present in `libs_to_deploy`.
+        for address in self.mcr.libraries.libs.values().flat_map(|libs| libs.values()) {
```
mablr: `self.mcr.libraries` includes every configured library, not only libraries referenced by this suite. Consequently, any unrelated configured address that has code is added to `deployed_libs` and excluded from fuzzed address inputs; this also performs configured-libraries × suites backend lookups (potentially against fork state). Can we carry the set of library addresses actually required by linked test artifacts and inspect/add only those here?

crates/linking/src/lib.rs
stevencartavia: [92f9a61](https://github.com/foundry-rs/foundry/pull/15839/commits/92f9a6101fc1cc5ad168ce54f4f00f6249e8ace7)

crates/forge/src/runner.rs
stevencartavia: [92f9a61](https://github.com/foundry-rs/foundry/pull/15839/commits/92f9a6101fc1cc5ad168ce54f4f00f6249e8ace7)

crates/forge/src/multi_runner.rs
```diff
@@ -760,12 +787,36 @@ impl MultiContractRunnerBuilder {
...
+        let create2 = create2_deployer_available.then(|| {
+            linker.link_with_create2(
+                configured_libraries.clone(),
+                evm_opts.create2_deployer,
+                self.config.create2_library_salt,
+                linker.contracts.keys(),
+            )
+        });
+        let (LinkOutput { libraries, library_addresses, libs_to_deploy }, library_deployment) =
+            if let Some(Ok(output)) = create2 {
```
mablr: This fallback still breaks test/script parity for the collision this PR now detects. If two compiler versions both reference the same `file:name`, `forge test` passes every artifact here, gets `ConflictingLibraryArtifacts`, and falls back to nonce addresses; `forge script` links only its selected target and can still use CREATE2. Their linked creation code therefore differs. The nonce path is also lossy here because its `Libraries` map collapses both artifacts to one `file:name` entry. Could we propagate `ConflictingLibraryArtifacts` instead of silently falling back (while retaining fallback for unavailable CREATE2 and, if intended, cycles), and cover this at the builder/CLI level? The current linker test only proves that the low-level CREATE2 call returns the collision error.

stevencartavia: [199ebfb](https://github.com/foundry-rs/foundry/pull/15839/commits/199ebfbf67a8f004caa1f8a446d4ff1a20dbce5a)

crates/forge/src/multi_runner.rs
```diff
@@ -760,12 +787,42 @@ impl MultiContractRunnerBuilder {
...
+            ) {
+                Ok(output) => Some(output),
+                Err(LinkerError::CyclicDependency) => None,
+                Err(err) => return Err(err.into()),
+            }
+        } else {
+            None
+        };
+        let (LinkOutput { libraries, library_addresses, libs_to_deploy }, library_deployment) =
+            if let Some(output) = create2 {
```
mablr, line 808: If CREATE2 linking returns no libraries to deploy, this still selects `Create2`, so the canonical factory is installed before the test constructor. That changes observable constructor state even when CREATE2 setup is unnecessary. Could we retain the resolved libraries but use `Nonce` when `libs_to_deploy.is_empty()`?

crates/linking/src/lib.rs
```diff
@@ -183,6 +187,46 @@ impl<'a> Linker<'a> {
         Ok(())
     }
 
+    fn collect_library_keys(
+        &self,
+        needed_libraries: &BTreeSet<&ArtifactId>,
+        libraries: &Libraries,
+    ) -> Result<BTreeSet<(PathBuf, String)>, LinkerError> {
+        let mut library_keys = BTreeSet::new();
+        for id in needed_libraries {
```
mablr, line 203: This collision check only sees artifacts already selected by `find_artifact_id_by_library_path`, which returns the first path/name/version match. Same-version artifacts from different build IDs/profiles can therefore still select arbitrary initcode without reaching this check. Please resolve within the target build/profile or reject multiple compatible candidates.

crates/script/src/lib.rs
```diff
@@ -799,11 +799,14 @@ pub struct ScriptConfig<FEN: FoundryEvmNetwork> {
...
         config: Config,
-        evm_opts: EvmOpts,
+        mut evm_opts: EvmOpts,
         batch: bool,
         tempo: TempoOpts,
         sender_nonce_override: Option<u64>,
     ) -> Result<Self> {
+        // Linking happens before runner construction, so pin now to ensure its CREATE2 factory
+        // lookup and all later environment/backend construction use the same fork block.
+        evm_opts.pin_fork_block().await?;
```
mablr, line 809: `update_sender` still queries its replacement sender nonce at `latest` (`next_nonce(..., None)`) before relinking. Once this fork is pinned, that can derive nonce-linked addresses from a different block than the execution backend. Please pass `self.evm_opts.fork_block_number` there too.

stevencartavia: [9add5c4](https://github.com/foundry-rs/foundry/pull/15839/commits/9add5c473af5625b31a972374f6292eaa94fb849)

crates/linking/src/lib.rs
stevencartavia: [9add5c4](https://github.com/foundry-rs/foundry/pull/15839/commits/9add5c473af5625b31a972374f6292eaa94fb849)

crates/forge/src/multi_runner.rs
stevencartavia: [9add5c4](https://github.com/foundry-rs/foundry/pull/15839/commits/9add5c473af5625b31a972374f6292eaa94fb849)

mablr (approved):

figtracer (approved):

PR #15841, opened 2026-07-21
fix(anvil): `eth_simulateV1` normalize simulated blocks

mablr:
## Motivation

Closes OSS-564.

This is the next focused `eth_simulateV1` conformance step after #15770, #15784, #15792, and #15828. Local simulations now derive block numbers and timestamps from the selected base block, expand number gaps with empty blocks, reject invalid ordering with `-38020` and `-38021`, chain parent hashes for `BLOCKHASH`, scope block overrides to their target block, and derive subsequent base fees from the simulated parent. Timestamp defaults honor Anvil’s configured block interval.

Normalization of delegated pre-fork simulations remains tracked by OSS-572, while canonical response finalization remains tracked by OSS-565.


mattsse (approved):
lgtm

PR #15842, opened 2026-07-21
fix(doc): neutralize ESM statement lines in generated NatSpec

0xMars42:
`forge doc` copies NatSpec prose into generated MDX pages. MDX treats a line whose first token is `import` or `export` as ESM syntax instead of displaying it as prose. The text may also reach a page through `@inheritdoc`.

This change neutralizes those lines after the complete rendered NatSpec comment block is assembled in `write_comment_block`. The keyword's first letter is replaced with an HTML entity, preserving the rendered text while preventing MDX from treating the line as ESM.

Markdown code ranges are identified with `pulldown-cmark`, so fenced blocks and inline code spans remain unchanged. Because MDX and CommonMark differ on indented fence closers, `mdx_fence_end` trims each parsed fenced range at the point where MDX closes it.

Regression coverage includes local and inherited NatSpec, complete comment blocks, matching fence markers and lengths, suffixed markers, indented fence closers, and multi-line inline code spans.


crates/doc/src/hir_ext.rs
```diff
@@ -800,7 +800,29 @@ pub fn replace_inline_links(
         i += ch.len_utf8();
     }
 
-    out
+    neutralize_esm(&out)
```
mablr: This also rewrites `import`/`export` inside fenced NatSpec code blocks, where character references render literally and corrupt the example. Please neutralize only outside code fences after continuation lines are assembled, and add coverage preserving fenced imports/exports.

DaniPopes:
don't we have a sanitization function somewhere? this should apply everywhere where we display any text

crates/doc/src/hir_ext.rs
0xMars42: Good catch, thanks. I moved the neutralization out of `replace_inline_links` into `write_comment_block`, so it runs after the continuation lines are assembled, and made it fence-aware: lines inside a ``` or ~~~ block are left untouched. Added `natspec_preserves_esm_inside_code_fences`, which keeps a fenced `import` verbatim while still neutralizing a top-level `export` after the fence closes.

0xMars42:
@DaniPopes the text escaping we have is `replace_inline_links`, which escapes `<` and `{` per tag. The ESM case is different: it is line-initial and must respect code fences, so handling it there (before continuation lines and fences are assembled) corrupted fenced examples. I moved it to `write_comment_block`, the single function every rendered NatSpec description goes through, and made it fence-aware, so it now applies wherever prose is displayed while leaving fenced code intact. Other surfaces (table cells, prefixed list items) are never line-initial, so MDX does not parse them as ESM.

crates/doc/src/hir_ext.rs
stevencartavia: +1 from my pass

mablr:
> don't we have a sanitization function somewhere? this should apply everywhere where we display any text

+1 let's do follow-up 🫡

crates/doc/src/render.rs
```diff
@@ -995,6 +995,42 @@ fn italicize_dev(content: &str) -> String {
...
+/// via `@inheritdoc`, so this must run wherever displayed prose is assembled. Lines inside a
+/// fenced code block are left untouched, where the entity would render literally and corrupt
+/// the example; a line legitimately starting with `import`/`export` outside a fence would
+/// break the MDX build anyway.
+fn neutralize_esm(text: &str) -> String {
+    let mut in_fence = false;
+    text.lines()
+        .map(|line| {
+            let trimmed = line.trim_start();
+            if trimmed.starts_with("```") || trimmed.starts_with("~~~") {
```
stevencartavia: Non-blocking follow-up: the fence-aware fix still toggles state on an opposite-marker line inside an open fence. CommonMark only closes a fence with the marker character that opened it, so `import`/`export` in a literal `~~~javascript` example inside a backtick fence are still rewritten. Could we remember the opening marker and only close on the same marker?

crates/doc/src/render.rs
```diff
@@ -1010,9 +1046,13 @@ fn write_comment_block(out: &mut String, data: &CommentData) {
     // `@dev` paragraphs are wrapped in `_..._` per paragraph so each multi-line block renders
     // as a single italic span (markdown emphasis cannot cross blank lines).
     for desc in &data.descriptions {
+        // Neutralize a line MDX would run as an ESM statement (a top-level `import`/`export`),
+        // but never inside a fenced code block, where the character reference would render
+        // literally and corrupt the example.
+        let content = neutralize_esm(&desc.content);
```
stevencartavia: Neutralization does not cover the complete rendered comment block. Multiline `@author` content can emit an unchanged top-level import, and calling `neutralize_esm` separately for each description resets fence state, allowing inherited notices split across fence boundaries to do the same. Both cases reproduce in generated MDX. Could we assemble the complete displayed comment block and neutralize it once, with regressions for both cases?

crates/doc/src/render.rs
0xMars42: Fixed. `neutralize_esm` now remembers the marker that opened the fence and closes only on that same marker, so a `~~~` line inside a backtick fence stays literal and the `import`/`export` around it are untouched. Covered by `natspec_fence_closes_only_on_matching_marker`.

crates/doc/src/render.rs
0xMars42: Fixed. `write_comment_block` now assembles the entire displayed block (titles, authors, notices, custom notes) and runs `neutralize_esm` once over it, so fence state stays continuous across descriptions and non-description lines are covered too. A multiline `@author` whose continuation is a top-level `import` is now neutralized. Covered by `natspec_neutralizes_esm_across_whole_comment_block`.

crates/doc/src/render.rs
```diff
@@ -995,36 +995,91 @@ fn italicize_dev(content: &str) -> String {
...
+/// break the MDX build anyway.
+fn neutralize_esm(text: &str) -> String {
+    // The marker that opened the current fence (```` ``` ```` or `~~~`), or `None` outside a
+    // fence. CommonMark closes a fence only with the marker character that opened it, so an
+    // opposite marker inside an open fence is literal content and must not toggle.
+    let mut fence: Option<&str> = None;
+    text.split('\n')
+        .map(|line| {
+            let trimmed = line.trim_start();
+            let marker = if trimmed.starts_with("```") {
```
mablr: This can desynchronize from CommonMark fence parsing and leave ESM executable. For example, a 4-backtick opener, a 3-backtick content line, then a 4-backtick closer makes this scanner close/reopen at the wrong points, so a following `export` is skipped. A suffixed marker such as ```not-a-close is also treated as a closer. Could we use parsed Markdown ranges, or at least track the full marker length and require a matching-or-longer, whitespace-only closer? Please add adversarial coverage for these cases.

crates/doc/src/render.rs
```diff
@@ -995,36 +995,91 @@ fn italicize_dev(content: &str) -> String {
...
+                    None => fence = Some(marker),
+                    Some(open) if open == marker => fence = None,
+                    Some(_) => {}
+                }
+                return line.to_string();
+            }
+            if fence.is_some() {
+                return line.to_string();
+            }
+            for keyword in ["import", "export"] {
```
mablr: This also rewrites lines inside multiline inline-code spans. MDX cannot parse those as ESM, and entities render literally in code spans, so users would see `&#105;mport`. Could we preserve inline-code ranges (ideally via the same Markdown-aware parsing) and add a rendered-output regression test?

crates/doc/src/render.rs
0xMars42: Fixed. Code ranges are now derived from a Markdown parse (pulldown-cmark) instead of a line scanner, so a fence is closed only by a matching-or-longer marker that is blank after the backticks. On top of that, because MDX has no indented code blocks and closes a fence at such a marker at any indentation (unlike CommonMark, which requires it to be indented at most three spaces), each fenced range is trimmed to MDX's close, so an `import`/`export` after an indented closer is still neutralized instead of hidden inside a range MDX has already left. Covered by `natspec_neutralizes_esm_after_longer_fence`, `natspec_fence_does_not_close_on_suffixed_marker`, and `natspec_neutralizes_esm_after_indented_fence_close`.

crates/doc/src/render.rs
0xMars42: Fixed. Inline code spans, including multi-line ones, are part of the parsed code ranges now, so an `import`/`export` inside them stays verbatim. Covered by `natspec_preserves_esm_inside_multiline_inline_code`. Indented code blocks are deliberately kept out of the code ranges, since MDX does not treat indentation as code, so an indented statement is still neutralized.

crates/doc/src/render.rs
```diff
@@ -995,36 +996,132 @@ fn italicize_dev(content: &str) -> String {
...
+/// Neutralize any line MDX would parse as an ESM statement (`import`/`export` at the start of
+/// a line): the keyword's first letter becomes an HTML entity, so the line renders the same but
+/// no longer begins with an ESM token. NatSpec text can be inherited from a dependency via
+/// `@inheritdoc`, so this must run wherever displayed prose is assembled. A keyword that falls
+/// inside a Markdown code span or fenced code block is left untouched (see `code_regions`): the
+/// entity would render literally and corrupt the example, and MDX would not execute it there.
+fn neutralize_esm(text: &str) -> String {
+    let regions = code_regions(text);
+    let mut offset = 0;
+    text.split('\n')
```
mablr: `split('\n')` misses lone `\r` line endings, but MDX treats a lone CR as a line ending. A block comment such as `Intro.\r\rexport const injected = 1` can therefore leave a top-level `export` untouched (including through `@inheritdoc`). Could we normalize CRLF and lone CR to LF before both `code_regions` and these offset calculations, and add a CR-only regression test?

crates/doc/src/render.rs
```diff
@@ -995,36 +996,132 @@ fn italicize_dev(content: &str) -> String {
...
+/// spans, including multi-line ones (fenced blocks are trimmed to MDX's close, see
+/// `mdx_fence_end`). MDX does not execute an `import`/`export` inside them, and
+/// an HTML entity would render literally there, so neutralization skips these ranges. Indented
+/// code blocks are deliberately excluded: MDX does not treat indentation as code, so an indented
+/// `import`/`export` is not executed and renders as ordinary text, where neutralizing it is
+/// harmless rather than corrupting.
+fn code_regions(text: &str) -> Vec<std::ops::Range<usize>> {
+    let mut regions = Vec::new();
+    let mut fenced_start: Option<usize> = None;
+    for (event, range) in Parser::new_ext(text, Options::empty()).into_offset_iter() {
```
mablr: This only protects fences recognized by CommonMark. Because MDX disables indented code, it can recognize an opening fence indented by 4+ spaces that `pulldown-cmark` does not report as `Fenced`. In that case an `import` inside the MDX fence is neutralized and the entity renders literally. Could we also detect MDX-only indented opening fences (through their matching closer or EOF) and cover a 4+-indented opener in a test?

crates/doc/src/render.rs
```diff
@@ -995,36 +996,184 @@ fn italicize_dev(content: &str) -> String {
...
+}
+
+/// Neutralize any line MDX would parse as an ESM statement (`import`/`export` at the start of
+/// a line): the keyword's first letter becomes an HTML entity, so the line renders the same but
+/// no longer begins with an ESM token. NatSpec text can be inherited from a dependency via
+/// `@inheritdoc`, so this must run wherever displayed prose is assembled. A keyword that falls
+/// inside a Markdown code span or fenced code block is left untouched (see `code_regions`): the
+/// entity would render literally and corrupt the example, and MDX would not execute it there.
+fn neutralize_esm(text: &str) -> String {
+    let normalized = text.contains('\r').then(|| text.replace("\r\n", "\n").replace('\r', "\n"));
```
mablr: This normalizes every line ending before finding code regions, so bare CR/CRLF bytes inside fenced and inline code are changed even though those regions are intended to remain intact. For example, `neutralize_esm("```\rimport inside\r```\rexport outside")` safely rewrites the final `export`, but also changes the fence body's CR separators to LF. Could we scan `\n`, `\r\n`, and bare `\r` as logical separators while splicing replacements into the original string, preserving all unrelated bytes? The CR tests should assert exact separator preservation inside code regions.

crates/doc/src/render.rs
```diff
@@ -995,36 +996,184 @@ fn italicize_dev(content: &str) -> String {
...
+    let text = normalized.as_deref().unwrap_or(text);
+    let regions = code_regions(text);
+    let mut offset = 0;
+    text.split('\n')
+        .map(|line| {
+            let line_start = offset;
+            offset += line.len() + 1;
+            let trimmed = line.trim_start();
+            let keyword_pos = line_start + (line.len() - trimmed.len());
+            if regions.iter().any(|region| region.contains(&keyword_pos)) {
```
mablr: This scans every code region for every line, making sanitization `O(lines × regions)`. A large local or inherited NatSpec comment with alternating inline spans and ESM-looking lines can trigger billions of range comparisons and effectively hang `forge doc`. Could we sort/merge the regions once and advance a single cursor as line positions increase (and only consult it after finding an `import`/`export` candidate)?

mablr (dismissed):
Simplified logic and made tests cleaner and more robust.

Now G2G imo

mablr (dismissed):

mablr (dismissed):

stevencartavia (dismissed):

crates/doc/src/render.rs
```diff
@@ -995,36 +997,132 @@ fn italicize_dev(content: &str) -> String {
...
+/// literally and corrupt the example, and MDX would not execute it there.
+fn neutralize_esm(text: &str) -> String {
+    let regions = code_regions(text);
+    let mut region_cursor = 0;
+    let mut copied = 0;
+    let mut out = String::with_capacity(text.len());
+
+    for (line_start, line) in logical_lines(text) {
+        let replacement = if line.starts_with("import ") {
+            Some(("&#105;&#109;", 2))
```
figtracer, line 1072: `natspec_neutralizes_esm_statement_lines` now fails because the checked-in markdown snapshot still expects `&#105;mport`. can we update that snapshot with this output change?

mablr (approved):

stevencartavia (approved):

PR #15843, opened 2026-07-21
fix(anvil): propagate simulated prevRandao

stevencartavia:
`eth_simulateV1` applied `prevRandao` during execution but omitted it from the synthetic block header, causing different `prevRandao` values to produce the same block hash and subsequent `BLOCKHASH` result.

Populate `mixHash` from the simulated block environment so chained hashes reflect the context used during execution.


mattsse (approved):
lgtm

mablr (approved):

PR #15844, opened 2026-07-21
fix(anvil): `eth_simulateV1` normalize delegated simulations

mablr:
## Motivation

Closes OSS-572.

This is the next focused `eth_simulateV1` conformance step after #15770, #15784, #15792, #15828, and #15841.

Pre-fork requests delegated upstream now fetch the selected base header, normalize block numbers and timestamps using Anvil's configured interval, and pin execution to that header's hash. Gap expansion and ordering and block-count errors therefore match local simulation without allowing a reorganization or noncanonical selector to normalize and execute against different same-height blocks.

> This PR was written and tested with the assistance of Codex.

### DIFF
--- a/crates/anvil/src/eth/api.rs
+++ b/crates/anvil/src/eth/api.rs
@@ -1,5 +1,5 @@
 use super::{
-    backend::mem::{BlockRequest, DatabaseRef, State},
+    backend::mem::{BlockRequest, DatabaseRef, State, sanitize_simulation_blocks},
     sign::build_impersonated,
 };
 use crate::{
@@ -2922,7 +2922,7 @@ impl EthApi<FoundryNetwork> {
 
     pub async fn simulate_v1(
         &self,
-        request: SimulatePayload,
+        mut request: SimulatePayload,
         block_number: Option<BlockId>,
     ) -> Result<Vec<SimulatedBlock<AnyRpcBlock>>> {
         const DEFAULT_BLOCK_INTERVAL_SECS: u64 = 12;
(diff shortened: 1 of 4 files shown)


mattsse (dismissed):
lgtm

crates/anvil/src/eth/api.rs
```diff
@@ -2970,6 +2960,30 @@ impl EthApi<FoundryNetwork> {
                 })
                 .unwrap_or(DEFAULT_BLOCK_INTERVAL_SECS)
         });
+
+        // check if the number predates the fork, if in fork mode
+        if let BlockRequest::Number(number) = block_request
+            && let Some(fork) = self.get_fork()
+            && fork.predates_fork(number)
+        {
+            let base_block = fork.block_by_number(number).await?.ok_or_else(|| {
```
stevencartavia: In the delegated path, `block_request` reduces the original `BlockId` to a number, so this lookup can return whichever same-height hash is currently mapped in the fork cache. I reproduced both directions: a hash-selected simulation can use another cached block, and a numeric simulation can use a cached noncanonical block instead of the upstream canonical one. Could we resolve the base block directly from upstream using the original `BlockId`, then normalize and forward that block’s hash?

mablr: https://github.com/foundry-rs/foundry/pull/15844/commits/0f2979bc0ed2f9edca6db2885ae6d5ecb991fbd5

figtracer (approved):

crates/anvil/src/eth/api.rs
```diff
@@ -2970,6 +2961,31 @@ impl EthApi<FoundryNetwork> {
                 })
                 .unwrap_or(DEFAULT_BLOCK_INTERVAL_SECS)
         });
+
+        // check if the number predates the fork, if in fork mode
+        if let BlockRequest::Number(number) = block_request
+            && let Some(fork) = self.get_fork()
+            && fork.predates_fork(number)
+        {
+            let block_id = block_id.unwrap_or_else(|| number.into());
```
stevencartavia, line 2970: Could we preserve the original `BlockId` only for hash selectors and use the already-resolved `number` for number/tag selectors? Reusing `safe` or `finalized` here re-resolves the tag against the live upstream, so if it advances after the fork, simulation can use a different—or even post-fork—base block. I reproduced this with a fork at block 40: local `safe` resolved to block 8, but after the upstream advanced, the simulation used block 40 and returned block 41 instead of 9.

mattsse (approved):
lgtm

PR #15848, opened 2026-07-22
fix(forge): include linked library deployments in state diffs

### LINKED ISSUE #6636: Include deployed libraries in `vm.stopAndReturnStateDiff()` results
### Component

Forge

Currently, if you deploy a contract that requires an externally linked library while recording with `vm.startStateDiffRecording()`, the results returned will not include an `AccountAccess` struct for deploying the library. Can we include library deployments in the state diff results?

### Additional context

_No response_

stevencartavia:
Records Forge’s automatic linked-library deployments in the first state-diff recording session for each test. This ensures `startStateDiffRecording` reports the actual library `CREATE` access while preserving test isolation and normal recording reset behavior. Fixes #6636.

### DIFF
--- a/crates/forge/src/runner.rs
+++ b/crates/forge/src/runner.rs
@@ -60,6 +60,7 @@ use foundry_evm::{
         },
         strategies::EvmFuzzState,
     },
+    inspectors::cheatcodes::Vm::AccountAccess,
     revm::{bytecode::opcode, primitives::hardfork::SpecId},
     traces::{TraceKind, TraceRequirements, load_contracts},
 };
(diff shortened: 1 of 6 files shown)


crates/forge/src/runner.rs
```diff
@@ -733,17 +734,33 @@ impl<'a, FEN: FoundryEvmNetwork> ContractRunner<'a, FEN> {
         self.executor.set_balance(LIBRARY_DEPLOYER, U256::MAX)?;
 
         let mut result = TestSetup::default();
+        let mut pending_account_diffs = Vec::new();
         for code in &self.mcr.libs_to_deploy {
+            let recording_library_deployment = !self.contract.library_addresses.is_empty()
```
mablr, line 739: Non-blocking: Could we only enable state-diff recording when the library about to be deployed belongs to this contract? `libs_to_deploy` contains all automatically deployed libraries in the project, so once this contract references any library, this currently runs the opcode-level state-diff hooks for every library deployment and discards unrelated results afterward. In projects with many libraries, that extra setup work repeats for each linked suite. We can derive the expected address from `LIBRARY_DEPLOYER` and the deployment nonce/index, check it against `self.contract.library_addresses` before starting the recording, and retain the post-deployment address check as a safeguard.

mattsse (approved):
lgtm

mablr (approved):
g2g, https://github.com/foundry-rs/foundry/pull/15848#discussion_r3628386361 can be done as follow-up.

figtracer (approved):

crates/forge/src/runner.rs
stevencartavia: https://github.com/foundry-rs/foundry/pull/15851

PR #15850, opened 2026-07-22
fix(invariant): gate selector-focus workers on target surface size

mablr:
Selector-focus mode (#15348) collapses one invariant worker's entire
mutation surface to a single selector for the whole campaign. On a
target with few candidate functions relative to worker count, that
trade sacrifices a large share of parallel compute: an 8-worker
campaign against the pinned scfuzzbench Liquity V2 Governance target
(16 candidate functions registered on one `targetContract`) consistently
stalled its focus worker (deterministically worker index 6 at 8
workers) at near-zero coverage while its peers reached thousands of
edges in the same wall-clock budget.

This gates focus-worker allocation on the ratio of candidate target
functions to worker count, so small-surface targets keep every worker
on the full handler set instead of sacrificing one to a single
selector. The threshold (4 candidate functions per worker) was picked
empirically against the pinned Liquity target's actual candidate count.

## Results

Matched master-vs-patched 8-worker run, same seed (`0x1`) and 330s
budget, pinned Liquity V2 Governance scfuzzbench target
(`42212075645e4fc7bc46714e3f14582a06181560`):

### Results

| | Master | Patched |
|---|---:|---:|
| Worker 6 (previously focus-assigned) edges | 4 | 5126 |
| Mean edges across 8 workers | 3768 | 4967 (+32%) |
| Worst-worker edges | 4 (worker 6) | 3420 (worker 7) |

This is a single-seed diagnostic result on one target, not yet a
multi-seed statistical claim. Full write-up, including a false start on
a target where the candidate-count pool wasn't actually the operative
variable (Nerite's 159-candidate handler surface still exhibits the same
worker collapse, pointing to a distinct, unaddressed mechanism there),
is in the accompanying fuzzing-evaluation research notes.

### DIFF
--- a/crates/evm/evm/src/executors/invariant/mod.rs
+++ b/crates/evm/evm/src/executors/invariant/mod.rs
@@ -95,6 +95,9 @@ const MIN_ESTIMATED_CALLS_PER_INVARIANT_WORKER: u64 =
     MIN_RUNS_PER_INVARIANT_WORKER as u64 * DEFAULT_DEPTH_FOR_INVARIANT_WORKER_CAP as u64;
 /// Share of parallel workers reserved for selector focus mode.
 const INVARIANT_FOCUS_WORKER_DIVISOR: usize = 8;
+/// Minimum number of distinct candidate target functions required per worker before any worker
+/// is carved out for selector-focus mode.
+const INVARIANT_FOCUS_MIN_CANDIDATES_PER_WORKER: usize = 4;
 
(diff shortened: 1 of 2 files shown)


mablr:
Closing — this fix was tuned against a single target/seed (threshold calibrated directly against Liquity's candidate count) and doesn't generalize (Nerite still shows the same worker collapse). Needs broader validation before this is worth pursuing again.

PR #15851, opened 2026-07-22
perf(forge): record only required library deployments

stevencartavia:
Enables state-diff recording only for linked libraries required by the current test contract, avoiding recording and discarding unrelated library deployments.
https://github.com/foundry-rs/foundry/pull/15848#discussion_r3628386361


mattsse (approved):
lgtm

mablr (approved):

PR #15852, opened 2026-07-22
fix(anvil): pin delegated simulation tags

mablr:
## Description

Follow-up to https://github.com/foundry-rs/foundry/pull/15844#discussion_r3631355194.

Delegated pre-fork simulations now preserve the original selector only for block hashes. Number and tag selectors use the height already resolved against Anvil’s fork, preventing moving upstream `safe` or `finalized` tags from selecting a different or post-fork base block.


mattsse (approved):
lgtm

stevencartavia (approved):

PR #15853, opened 2026-07-22
fix(fuzz): warn on bool invariant returns

mablr:
Boolean-returning invariant functions currently compile and run, but Forge ignores their return value, allowing `return false` to appear to pass silently. This adds a warning directing users to assertions or reverts while preserving existing pass/fail behavior and `int256` optimization invariants.

Closes OSS-577

This change was implemented with AI assistance.


mattsse (approved):
lgtm

stevencartavia (approved):

PR #15854, opened 2026-07-22
fix(fuzz): minimize broken invariant corpus

mablr:
## Summary

`forge fuzz cmin` treated invariant predicate failures reproduced during replay as execution errors, making it unable to minimize corpora collected after a continuous invariant campaign found a bug. This change treats reproduced invariant and `afterInvariant` failures as valid coverage observations while continuing to reject stateless fuzz and invariant handler failures.

The regression coverage exercises a canary-style broken invariant and confirms that cmin retains the distinct coverage paths instead of rejecting the corpus.

Closes OSS-575

### DIFF
--- a/crates/forge/src/cmd/fuzz.rs
+++ b/crates/forge/src/cmd/fuzz.rs
@@ -397,7 +397,7 @@ impl FuzzCminArgs {
             let mut entry_unmatched_txs = 0usize;
             let mut entry_rejected_txs = 0usize;
             for FuzzMinimizeObservation { target, observation } in observations {
-                if observation.failure.is_some() {
+                if observation.failure.as_ref().is_some_and(|failure| !failure.is_predicate()) {
                     entry_failed = true;
                     entry_failed_replays += observation.replayed;
                     continue;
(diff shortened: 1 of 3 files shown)


mattsse (dismissed):
lgtm

crates/forge/src/cmd/fuzz.rs
```diff
@@ -397,7 +397,7 @@ impl FuzzCminArgs {
             let mut entry_unmatched_txs = 0usize;
             let mut entry_rejected_txs = 0usize;
             for FuzzMinimizeObservation { target, observation } in observations {
-                if observation.failure.is_some() {
+                if observation.failure.as_ref().is_some_and(|failure| !failure.is_predicate()) {
```
stevencartavia: This can merge coverage from a truncated replay. With `[breakA(), laterBranch(1)]` and `[breakA(), laterBranch(2)]`, replay stopped after detecting the predicate failure and cmin kept only 1/2 entries. Comparing showmap edge sets found seven original edge IDs missing afterward. If predicate failures are accepted, could replay continue collecting coverage from the remaining transactions?

stevencartavia: The representative `failure` can hide an earlier handler failure. With `check_interval = 0`, `forge fuzz replay` rejected `[breakInvariant(), assertHandler()]` because the handler failed, but `forge fuzz cmin` accepted the same corpus and kept 1/1 entries because the later predicate failure replaced it. Could we track non-predicate failures separately and reject the observation when any occurred?

stevencartavia (approved):

mattsse (approved):
lgtm

PR #15857, opened 2026-07-22
fix(anvil): properly execute Tempo tx requests

mablr:
## Description

#15806 follow-up

Route parsed transaction requests through network-specific call environments so Tempo AA requests retain their complete call batches across eth_call, estimation, access-list generation, tracing, eth_simulateV1, and node-signed transaction flows.

Preserve Tempo nonce, fee-payer, signature, and transaction identity semantics during probing and sequential simulation. Construct synthetic simulation envelopes from the original request while applying gas limits and other execution-only changes to the prepared environment.

Keep Ethereum and Optimism request classification intact and build simulated receipts with the appropriate network-specific fields.


mattsse (changes requested):
Requesting changes for two regressions.

The new typed gas-estimation path executes Tempo keychain/session requests with the probe gas limit installed in the full AA environment. The first probe commonly uses the block gas limit, and Tempo's keychain fee validation treats that candidate limit as spend. As a result, otherwise valid session-key transactions fail with `SpendingLimitExceeded` before the binary search can find their actual gas requirement. This currently breaks multiple supported Cast flows; I reproduced `keychain::batch_send_uses_tempo_session_id_env` locally on all three retries, and the full CI run reports ten related keychain failures.

Please preserve the request's keychain identity and call batch without letting an execution-only probe limit invalidate the authorization. Depending on the intended Tempo semantics, that likely means either separating the estimation gas cap from the gas value used by keychain fee-policy validation, or bounding the initial probe by the remaining keychain fee allowance. The important property is that a valid transaction must reach the binary search rather than fail solely because the first candidate is the block gas limit.

Please add Anvil-level regression coverage for `eth_estimateGas` with both same-transaction key authorization and an already-provisioned session key. The tests should cover a valid constrained key that estimates successfully and a genuinely over-limit request that remains rejected. Keeping the existing Cast session tests green will also protect the end-to-end workflow.

Separately, `crates/anvil/src/eth/backend/mem/optimism.rs:20` still links to the removed `Backend::transact_with_inspector_ref` method. The repository denies broken rustdoc links, so this deterministically fails the docs job. Please update the link to the replacement method or remove it.

crates/anvil/src/eth/backend/executor.rs
```diff
@@ -38,43 +38,92 @@ use revm::{
...
+        cumulative_gas_used: u64,
+        trace_transfers: bool,
+        deposit_nonce: Option<u64>,
+        deposit_receipt_version: Option<u64>,
+    ) -> FoundryReceiptEnvelope {
+        let logs = result
+            .clone()
+            .into_logs()
+            .into_iter()
+            .filter(|log| {
```
stevencartavia: With `traceTransfers` enabled, this filter also removes genuine EVM logs emitted by `0xeeee…eeee` whose first topic is `Transfer(address,address,uint256)`. Such a log remains in `calls[].logs` but is omitted from the simulated receipt, making the block’s `logsBloom` and `receiptsRoot` incorrect.

Could we distinguish synthetic transfer logs by provenance instead of filtering receipt logs by address and topic?

mattsse (changes requested):
Requesting changes for a deterministic `eth_simulateV1` panic. A Tempo batch can emit canonical logs in an early subcall and then revert later; the final `ExecutionResult` has no logs, but the simulation collector assumes every previously observed canonical log survived. The current `test_tempo_fork_forwards_request_extensions` fails on all three CI attempts with `collected canonical log exists in result`, returning JSON-RPC `-32603` instead of the simulated failure response.

Suggested fix: pass the final execution status into the collector and, on top-level failure, truncate all collected response logs while preserving `next_index` so reverted attempts still leave the expected log-index gap. Only reconcile surviving canonical logs on success. The failing fork test should then return status `0x0`; a direct non-fork batch-revert regression would also make this invariant explicit.

mattsse (dismissed):
lgtm

figtracer (dismissed):

figtracer (approved):

stevencartavia (approved):

PR #15860, opened 2026-07-23
fix(forge): exclude unrelated re-export imports

### LINKED ISSUE #7239: bug: `forge flatten` should ignore intermediary files for imports
### Component

Forge

- [X] Foundry
- [X] Foundryup

0.2.0

forge flatten

### Operating System

macOS (Apple Silicon)

### Describe the bug

In more complex codebases, you might want to import objects from files other than the file in which an object is originally defined, in order to make the codebase easier to read. A simplified example:

File `A`:
```solidity
import {Test} from "./Intermediary";
```

File `Intermediary`:
```solidity
import {Test} from "./Test";
import {Test2} from "./Test2";
```

File `Test`:
```solidity
contract Test {}
```

However, when running `forge flatten` (and when using `--verify`), all files are treated equally, even when they are only used as "intermediary files". This means that the flattened version of file A will also include `Test2`, which is never imported into A.

#### Proposed solution
The flattener should determine whether a file is used only as an intermediary and ignore it if it is, jumping to the file from which the used import is exported from.

stevencartavia:
Updates `foundry-core` to `fa7fd22`, bringing semantic source-unit selection into `forge flatten`. Named imports through intermediary re-export files now include only declaration-owning sources, while plain and namespace imports remain conservative.

Fixes #7239.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5271,7 +5271,7 @@ dependencies = [
 [[package]]
 name = "foundry-block-explorers"
 version = "0.24.0"
-source = "git+https://github.com/foundry-rs/foundry-core?rev=9c4a397f42d254e95a1ee8354b81b17805e147ef#9c4a397f42d254e95a1ee8354b81b17805e147ef"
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm

PR #15862, opened 2026-07-23
fix(script): skip unused library deployments

### LINKED ISSUE #6215: Library linking
### Component

Forge

Right now unused libraries may be deployed even if there is no usage of them outside of broadcasting:

- https://github.com/foundry-rs/foundry/issues/3295
- https://github.com/foundry-rs/foundry/issues/5375

This happens because from perspective of Solidity compiler, script contract depends on all libraries which are used in it regardeless of part of code they are used in.

One possible solution is as follows:
1. After script simulation, go through all `CREATE`'s and match bytecodes to artifacts.
2. Determine which libraries are actually going to be used in newly created contracts. This can be done via [existing code for dependency resolution
](https://github.com/foundry-rs/foundry/blob/master/crates/utils/src/lib.rs)
3. If any unnecessary libraries are found, recompile the code and rerun the script, predeploying only the necessary libraries. This step is necessary because modifying the pre-deployed libraries list affects the sender nonce. All other libraries can be injected by some way similar to what `vm.etch` does

This approach has several drawbacks:
1. Current code matching bytecode to artifacts uses [diff_score](https://github.com/foundry-rs/foundry/blob/master/crates/common/src/contracts.rs#L25) which may result in false positives/negatives in some cases. Right now such fails are accepted because they are only affecting verification and trace decoding. However, in case of library linking it becomes much more dangerous as it may lead to contracts being deployed without needed libraries
2. Recompiling code and re-running script with new libraries is pretty expensive.

Would be happy to get other opinions on this

### Additional context

_No response_

stevencartavia:
Avoid deploying unused libraries when `forge script` broadcasts direct contract creations. Required libraries are determined from the created artifacts, while omitted libraries remain available during local execution. Unsupported transaction shapes retain the existing deploy-all behavior.

Closes #6215

### DIFF
--- a/crates/script/src/library_deployments.rs
+++ b/crates/script/src/library_deployments.rs
@@ -287,6 +287,10 @@ impl<FEN: FoundryEvmNetwork> PreSimulationState<FEN> {
     fn used_rerun_unsafe_cheatcode(&self) -> bool {
         const SIGNATURES: &[&str] = &[
             "setEnv(string,string)",
+            "sign(bytes32)",
+            "sign(address,bytes32)",
+            "signCompact(bytes32)",
+            "signCompact(address,bytes32)",
             "rpc(string,string)",
             "rpc(string,string,string)",
             "rpcJson(string,string)",
--- a/crates/script/src/library_deployments.rs
+++ b/crates/script/src/library_deployments.rs
@@ -1,0 +1,349 @@
+use crate::{
+    build::{LinkedBuildData, ScriptPredeployLibraries},
+    execute::LinkedState,
+    simulate::PreSimulationState,
+};
+use alloy_network::TransactionBuilder;
+use alloy_primitives::{Address, keccak256};
+use eyre::Result;
+use foundry_common::{LIBRARY_DEPLOYER, matches_contract_creation, shell};
+use foundry_compilers::ArtifactId;
+use foundry_evm::{constants::CHEATCODE_ADDRESS, core::evm::FoundryEvmNetwork};
+use std::collections::BTreeSet;
+
+struct EligibleTransactions {
+    required: BTreeSet<ArtifactId>,
+    artifacts: Vec<ArtifactId>,
+    rpc: String,
+    baseline_count: usize,
+}
+
+impl<FEN: FoundryEvmNetwork> PreSimulationState<FEN> {
+    /// Reruns narrowly eligible scripts with libraries unused by direct broadcasts deployed only
+    /// in the local EVM.
+    pub async fn optimize_library_deployments(self) -> Result<Self> {
+        let Ok(Some(EligibleTransactions { required, artifacts, rpc, baseline_count })) =
+            self.eligible()
+        else {
+            return Ok(self);
+        };
+        let onchain = match &self.build_data.predeploy_libraries {
+            ScriptPredeployLibraries::Default { onchain, .. }
+            | ScriptPredeployLibraries::Create2 { onchain, .. } => onchain,
+        };
+        if required.len() == onchain.len() {
+            return Ok(self);
+        }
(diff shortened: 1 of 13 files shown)


figtracer (dismissed):

crates/script/src/library_deployments.rs
```diff
@@ -0,0 +1,349 @@
...
+            remaps.push((
+                old.from().unwrap().create(old_nonce),
+                new.from().unwrap().create(new_nonce),
+            ));
+        }
+        true
+    }
+
+    fn used_rerun_unsafe_cheatcode(&self) -> bool {
+        const SIGNATURES: &[&str] = &[
```
mablr: Could the wallet-backed signing overloads also be treated as rerun-unsafe? The speculative second execution can prompt a hardware/remote signer twice, and falling back afterward cannot undo that interaction. Please include `sign(bytes32)`, `sign(address,bytes32)`, `signCompact(bytes32)`, and `signCompact(address,bytes32)` here.

mattsse (commented):
the candidate [reruns the script with cloned wallets](https://github.com/foundry-rs/foundry/blob/b3daaa71cfe4febc4c8e9679cb8f21c45f94ac4d/crates/script/src/library_deployments.rs#L81-L92), but the [rerun-unsafe selector list](https://github.com/foundry-rs/foundry/blob/b3daaa71cfe4febc4c8e9679cb8f21c45f94ac4d/crates/script/src/library_deployments.rs#L287-L314) doesn't include wallet-backed `vm.sign`/`vm.signCompact`. this can prompt hardware, browser, or remote signers twice. can we add `sign(bytes32)`, `sign(address,bytes32)`, `signCompact(bytes32)`, and `signCompact(address,bytes32)` to `SIGNATURES` and cover that they skip the optimization?

crates/script/src/library_deployments.rs
stevencartavia: [228a5c5](https://github.com/foundry-rs/foundry/pull/15862/commits/228a5c5367326605f7de093f8ae04e371d6000f0)

mablr (approved):
LGTM

crates/script/src/library_deployments.rs
```diff
@@ -0,0 +1,353 @@
...
+            || self.execution_artifacts.rpc_data.is_multi_chain()
+            || self.script_config.evm_opts.sender == LIBRARY_DEPLOYER
+            || self.script_config.config.fs_permissions.permissions.iter().any(|permission| {
+                matches!(
+                    permission.access,
+                    foundry_config::fs_permissions::FsAccessPermission::Write
+                        | foundry_config::fs_permissions::FsAccessPermission::ReadWrite
+                )
+            })
+            || self.used_rerun_unsafe_cheatcode()
```
mablr, line 171: This gate only sees unsafe cheatcodes reached by the baseline run. Nondeterministic control flow can reach `vm.rpc`, prompting/signing, or another side effect only in the speculative rerun, before equivalence rejects that candidate. I reproduced this with a candidate-only `vm.rpc("anvil_setBalance", ...)`: the RPC mutation escaped even though the deploy-all baseline was retained. Please hard-disable side-effecting cheatcodes during the candidate execution (an error can safely fall back to the baseline).

mablr (changes requested):

crates/script/src/library_deployments.rs
mablr: Make sure #15207 is not reintroduced.

stevencartavia: d06e78d

mablr (approved):

mattsse (approved):
lgtm

PR #15863, opened 2026-07-23
feat(forge): verify external factory deployments

### LINKED ISSUE #10164: verification on external code
### Component

Forge

When deploying a script that interacts with an external factory, foundry does not verify the created contracts.
E.g. [here](https://sepolia.basescan.org/tx/0x17080c36a867b4a0ef6d93333636fa380d2d0e1e9dc91875bd8c9eac7f4272d1) we call a proxy factory that creates on oz proxy - foundry did not verify the proxy(it's verified because i did it manually).

By my understanding, the reason for that is that foundry does not have the local bytecode of that contract in that case (given the factory was deployed from another repo potentially with other compiler settings).

---

Now, as i understand foundry scripts, for tracing all touched addresses are downloaded from explorers already - so in fact found should have the code, even if it's not part of the repo. While this is a bit of an edge case, over the last couple years we faced this shortcoming quite a few times.

### Additional context

_No response_

stevencartavia:
Adds opt-in verification of contracts created by already-deployed external factories through `forge script --verify --verify-external`.

Forge records creator provenance from execution traces, retrieves verified Standard JSON from Sourcify or Etherscan, recompiles it with the original compiler settings, matches the observed `CREATE`/`CREATE2` init code and constructor arguments, and submits unique matches through the existing verification providers. Local verification remains first, and broadcast logs remain backward-compatible.

Closes #10164

### DIFF
--- a/crates/script/src/lib.rs
+++ b/crates/script/src/lib.rs
@@ -239,6 +239,10 @@ pub struct ScriptArgs {
     #[arg(long, requires = "broadcast")]
     pub verify: bool,
 
+    /// Opt-in verification of contracts deployed by external factories using explorer source.
+    #[arg(long, requires = "verify")]
+    pub verify_external: bool,
+
     /// Gas price for legacy transactions, or max fee per gas for EIP1559 transactions, either
     /// specified in wei, or as a string with a unit type.
     ///
@@ -753,20 +757,38 @@ impl<N: Network> ScriptResult<N> {
         self.traces
             .iter()
             .flat_map(|(_, traces)| {
-                traces.nodes().iter().filter_map(|node| {
-                    if node.trace.kind.is_any_create() {
-                        let init_code = node.trace.data.clone();
-                        let contract_name = known_contracts
-                            .find_by_creation_code(init_code.as_ref())
-                            .map(|artifact| artifact.0.name.clone());
-                        return Some(AdditionalContract {
-                            call_kind: node.trace.kind,
-                            address: node.trace.address,
-                            contract_name,
-                            init_code,
-                        });
+                let nodes = traces.nodes();
+                nodes.iter().filter_map(|node| {
+                    if !node.trace.kind.is_any_create() || !node.trace.success {
+                        return None;
                     }
-                    None
+
+                    let mut creator_code_addresses = Vec::new();
+                    let mut parent = node.parent;
+                    while let Some(parent_idx) = parent {
+                        let ancestor = &nodes[parent_idx];
+                        if !ancestor.trace.success {
+                            return None;
+                        }
+                        if !ancestor.trace.kind.is_any_create()
+                            && !creator_code_addresses.contains(&ancestor.trace.address)
+                        {
+                            creator_code_addresses.push(ancestor.trace.address);
+                        }
+                        parent = ancestor.parent;
+                    }
+
+                    let init_code = node.trace.data.clone();
+                    let contract_name = known_contracts
+                        .find_by_creation_code(init_code.as_ref())
+                        .map(|artifact| artifact.0.name.clone());
+                    Some(AdditionalContract {
+                        call_kind: node.trace.kind,
+                        address: node.trace.address,
+                        contract_name,
+                        init_code,
+                        creator_code_addresses,
+                    })
                 })
             })
             .collect()
@@ -962,6 +984,9 @@ mod tests {
         upsert_session_entry,
     };
     use foundry_config::UnresolvedEnvVarError;
+    use foundry_evm::traces::{
+        CallKind, CallTrace, CallTraceArena, CallTraceNode, SparsedTraceArena, TraceKind,
+    };
     use std::{fs, sync::LazyLock};
     use tempfile::tempdir;
     use tokio::sync::{Mutex, MutexGuard};
@@ -980,6 +1005,101 @@ mod tests {
 
         let tracing = script_trace_requirements(&config, false).into_config().unwrap();
         assert!(tracing.record_state_diff);
+    }
+
+    #[test]
+    fn created_contracts_include_successful_creator_code_ancestry() {
+        let root = Address::repeat_byte(0x11);
+        let implementation = Address::repeat_byte(0x22);
+        let helper = Address::repeat_byte(0x33);
+        let created = Address::repeat_byte(0x44);
+        let init_code = Bytes::from_static(&[0x60, 0x00]);
+        let mut arena = CallTraceArena::default();
+        arena.nodes_mut()[0].trace =
+            CallTrace { success: true, address: root, kind: CallKind::Call, ..Default::default() };
+        arena.nodes_mut().extend([
+            CallTraceNode {
+                parent: Some(0),
+                idx: 1,
+                trace: CallTrace {
+                    success: true,
+                    address: implementation,
+                    kind: CallKind::DelegateCall,
+                    ..Default::default()
+                },
+                ..Default::default()
+            },
+            CallTraceNode {
+                parent: Some(1),
+                idx: 2,
+                trace: CallTrace {
+                    success: true,
+                    address: helper,
+                    kind: CallKind::Call,
+                    ..Default::default()
+                },
+                ..Default::default()
+            },
+            CallTraceNode {
+                parent: Some(2),
+                idx: 3,
+                trace: CallTrace {
+                    success: true,
+                    address: created,
+                    kind: CallKind::Create2,
+                    data: init_code.clone(),
+                    ..Default::default()
+                },
+                ..Default::default()
+            },
+        ]);
+        let result = ScriptResult::<Ethereum> {
+            traces: vec![(
+                TraceKind::Execution,
+                SparsedTraceArena { arena, ignored: Default::default() },
+            )],
+            ..Default::default()
+        };
+
+        let contracts = result.get_created_contracts(&ContractsByArtifact::default());
+        assert_eq!(contracts.len(), 1);
+        assert_eq!(contracts[0].address, created);
+        assert_eq!(contracts[0].init_code, init_code);
+        assert_eq!(contracts[0].creator_code_addresses, [helper, implementation, root]);
+    }
+
+    #[test]
+    fn created_contracts_exclude_creations_rolled_back_by_an_ancestor() {
+        let mut arena = CallTraceArena::default();
+        arena.nodes_mut()[0].trace = CallTrace { success: true, ..Default::default() };
+        arena.nodes_mut().extend([
+            CallTraceNode {
+                parent: Some(0),
+                idx: 1,
+                trace: CallTrace { success: false, kind: CallKind::Call, ..Default::default() },
+                ..Default::default()
+            },
+            CallTraceNode {
+                parent: Some(1),
+                idx: 2,
+                trace: CallTrace {
+                    success: true,
+                    kind: CallKind::Create,
+                    data: Bytes::from_static(&[0x60, 0x00]),
+                    ..Default::default()
+                },
+                ..Default::default()
+            },
+        ]);
+        let result = ScriptResult::<Ethereum> {
+            traces: vec![(
+                TraceKind::Execution,
+                SparsedTraceArena { arena, ignored: Default::default() },
+            )],
+            ..Default::default()
+        };
+
+        assert!(result.get_created_contracts(&ContractsByArtifact::default()).is_empty());
     }
 
     fn active_session_entry(
@@ -1040,6 +1160,23 @@ mod tests {
         let sig = "0x522bb704000000000000000000000000f39fd6e51aad88f6f4ce6ab8827279cfFFb92266";
         let args = ScriptArgs::parse_from(["foundry-cli", "Contract.sol", "--sig", sig]);
         assert_eq!(args.sig, sig);
+    }
+
+    #[test]
+    fn verify_external_requires_verify() {
+        let err = ScriptArgs::try_parse_from(["foundry-cli", "Contract.sol", "--verify-external"])
+            .unwrap_err();
+        assert_eq!(err.kind(), clap::error::ErrorKind::MissingRequiredArgument);
+
+        let args = ScriptArgs::try_parse_from([
+            "foundry-cli",
+            "Contract.sol",
+            "--broadcast",
+            "--verify",
+            "--verify-external",
+        ])
+        .unwrap();
+        assert!(args.verify_external);
     }
 
     #[test]
--- a/crates/script/src/lib.rs
+++ b/crates/script/src/lib.rs
@@ -240,7 +240,7 @@ pub struct ScriptArgs {
     pub verify: bool,
 
     /// Opt-in verification of contracts deployed by external factories using explorer source.
-    #[arg(long, requires = "verify")]
+    #[arg(long, requires = "verify", conflicts_with_all = ["skip_simulation", "offline"])]
     pub verify_external: bool,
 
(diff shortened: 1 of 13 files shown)


figtracer (dismissed):

crates/script/src/verify/external.rs
```diff
@@ -0,0 +1,975 @@
...
+
+    pub(super) async fn resolve_etherscan(
+        &mut self,
+        config: &Config,
+        chain: Chain,
+        address: Address,
+        api_key: Option<String>,
+    ) -> Result<Option<ExternalSource>, String> {
+        let resolved = config
+            .get_etherscan_config_with_chain(Some(chain))
```
mablr: Could this use the selected verifier endpoint? As written, external discovery ignores `--verifier-url`, so custom Blockscout/Etherscan-compatible explorers can submit verification but cannot provide sources. The test currently masks this by duplicating the mock URL under `[etherscan]`.

crates/verify/src/etherscan/mod.rs
```diff
@@ -296,7 +340,14 @@ impl EtherscanVerificationProvider {
             builder.chain(chain)?
         };
 
+        let http = reqwest::Client::builder()
+            .timeout(Duration::from_secs(30))
+            .connect_timeout(Duration::from_secs(10))
+            .redirect(reqwest::redirect::Policy::none())
```
mablr: Can we keep normal redirect handling for trusted verification requests and restrict this policy to external source discovery? Preflight still follows redirects, so a 307/308 provider can pass preflight and then fail verification; this also changes existing non-external flows.

mattsse (changes requested):
Couple blockers before this is ready:

- External discovery only consults the configured Etherscan entry, ignoring the selected `--verifier-url` ([code](https://github.com/foundry-rs/foundry/blob/508d45cc64fa4355a0f7dc4be4700918f4c8bc2d/crates/script/src/verify/external.rs#L130-L140)). Please pass the selected endpoint/key into the resolver and update the integration test to rely on the CLI endpoint without duplicating it under `[etherscan]`.
- Redirects are disabled for all trusted verification requests ([Etherscan](https://github.com/foundry-rs/foundry/blob/508d45cc64fa4355a0f7dc4be4700918f4c8bc2d/crates/verify/src/etherscan/mod.rs#L343-L347), [Sourcify](https://github.com/foundry-rs/foundry/blob/508d45cc64fa4355a0f7dc4be4700918f4c8bc2d/crates/verify/src/sourcify.rs#L132-L138)), while preflight still follows them. Please keep normal redirect handling for submission/status requests, scope the no-redirect policy to external source discovery, and add a redirect regression test.

crates/script/src/verify/external.rs
stevencartavia: [9f37d12](https://github.com/foundry-rs/foundry/pull/15863/commits/9f37d1299f6af324e1a6481858f41c326be5a69c)

crates/verify/src/etherscan/mod.rs
stevencartavia: [9f37d12](https://github.com/foundry-rs/foundry/pull/15863/commits/9f37d1299f6af324e1a6481858f41c326be5a69c)

mattsse (dismissed):
lgtm

crates/script/src/verify/external.rs
```diff
@@ -0,0 +1,1008 @@
...
+        let key = FetchKey { chain: chain.id(), provider: provider.clone(), address };
+        if let Some(cached) = self.fetch_cache.get(&key) {
+            return cached.clone();
+        }
+        if self.fetch_cache.len() >= MAX_FETCHES {
+            return Err("external source fetch limit exceeded".to_string());
+        }
+        let result = async {
+            let url = format!(
+                "https://sourcify.dev/server/v2/contract/{}/{address}?fields=stdJsonInput,compilation.compilerVersion",
```
mablr: This hard-codes canonical Sourcify, so discovery ignores a selected custom/private Sourcify endpoint (including chain-specific endpoints such as Tempo) and may leak queried addresses publicly. Please carry the resolved Sourcify endpoint into the resolver and cache key.

crates/script/src/verify/external.rs
```diff
@@ -0,0 +1,1008 @@
...
+                    return Err(err)
+                }
+                Err(_) => {
+                    let _ = child.kill().await;
+                    let _ = child.wait().await;
+                    return Err("subprocess output task failed".to_string())
+                }
+                Ok(Ok(output)) => {
+                    completed_output = Some(output);
+                    child.wait().await.map_err(|_| "failed to wait for subprocess".to_string())?
```
mablr: The deadline no longer participates once this branch wins, so this `child.wait()` can hang past the timeout. The post-`select!` stdin/output awaits have the same issue. Please apply one absolute deadline to the full subprocess lifecycle, then kill/reap and abort I/O tasks on expiry.

crates/script/src/verify/external.rs
```diff
@@ -0,0 +1,1008 @@
...
+        .get("result")
+        .and_then(Value::as_array)
+        .and_then(|items| items.first())
+        .and_then(Value::as_object)
+        .ok_or_else(|| "empty Etherscan response".to_string())?;
+    let compiler_version = item
+        .get("CompilerVersion")
+        .and_then(Value::as_str)
+        .ok_or_else(|| "Etherscan compiler version is missing".to_string())?
+        .to_string();
```
mablr: This retains an unbounded remote compiler-version string; cached error strings are similarly uncharged, and the version is cloned into candidates. That bypasses the cumulative memory budgets. Please cap remote versions/errors and include retained strings in accounting (or share the parsed version).

crates/script/src/verify.rs
```diff
@@ -223,35 +428,77 @@ async fn verify_contracts<FEN: FoundryEvmNetwork>(
...
+                            init_code,
+                            creator_code_addresses,
+                            receipt.transaction_hash(),
+                        )
+                        .await
+                        {
+                            Ok(job) => verification_jobs.push(job),
+                            Err(reason) => {
+                                let _ = sh_warn!(
+                                    "Skipping external verification for {address}: {reason}"
```
mablr, line 542: Follow-up: the submitted-job-only `All (0)` summary predates this PR, but this new path makes it routine for requested external verification to be skipped yet reported as complete success. We may certainly track/report skipped external attempts and explicitly choose their exit-status semantics.

stevencartavia: [ac32763](https://github.com/foundry-rs/foundry/pull/15863/commits/ac32763ad7c41bd6d72804308fd9bfdd3f5c6de0)

crates/script/src/verify/external.rs
stevencartavia: [ac32763](https://github.com/foundry-rs/foundry/pull/15863/commits/ac32763ad7c41bd6d72804308fd9bfdd3f5c6de0)

crates/script/src/verify/external.rs
stevencartavia: [ac32763](https://github.com/foundry-rs/foundry/pull/15863/commits/ac32763ad7c41bd6d72804308fd9bfdd3f5c6de0)

crates/script/src/verify/external.rs
stevencartavia: [ac32763](https://github.com/foundry-rs/foundry/pull/15863/commits/ac32763ad7c41bd6d72804308fd9bfdd3f5c6de0)

crates/script/src/verify/external.rs
```diff
@@ -0,0 +1,1284 @@
...
+            Some(constructor) if constructor.inputs.is_empty() => suffix.is_empty(),
+            Some(constructor) => constructor
+                .abi_decode_input(suffix)
+                .ok()
+                .and_then(|values| constructor.abi_encode_input(&values).ok())
+                .is_some_and(|encoded| encoded == suffix),
+        };
+        if valid
+            && !matches.iter().any(|item: &ExternalMatch| {
+                item.fingerprint == candidate.fingerprint
```
mablr: This makes equivalent provider results ambiguous because `fingerprint` hashes the original Standard JSON input. Sourcify and Etherscan can return inputs that differ only in `outputSelection` or provider-specific fields while compiling to the same deployment. Since both candidates match the observed init code, this causes an explicit `--verify-external` run to skip verification as ambiguous. Could we deduplicate by deployment identity instead (FQN, compiler identity, matched creation bytecode, and constructor suffix)?

crates/script/src/verify/external.rs
```diff
@@ -0,0 +1,1284 @@
...
+        return Err("solc compilation failed".to_string());
+    }
+    let fingerprint = fingerprint(&source.input).map_err(|e| e.to_string())?;
+    let mut candidates = Vec::new();
+    let mut creation_bytecode_bytes = 0usize;
+    for (path, contracts) in output.contracts {
+        for (name, contract) in contracts {
+            let Some(abi) = contract.abi else { continue };
+            let Some(bytecode) = contract.evm.and_then(|evm| evm.bytecode) else { continue };
+            if !bytecode.link_references.is_empty() || bytecode.object.is_unlinked() {
```
mablr, line 594: This silently drops Standard JSON inputs with unresolved/manual library links. Sourcify can verify these using bytecode link transformations even when `settings.libraries` is absent, so an otherwise reproducible deployment becomes a generic no-match. Can we either recover the addresses from the observed init code using `link_references`, patch `settings.libraries`, and recompile for an exact match, or report this case explicitly as unsupported?

crates/script/src/verify/external.rs
stevencartavia: [3dbceb9](https://github.com/foundry-rs/foundry/pull/15863/commits/3dbceb938fd9b4b430b29383bbb69826c580c54b)

crates/script/src/verify/external.rs
stevencartavia: [3dbceb9](https://github.com/foundry-rs/foundry/pull/15863/commits/3dbceb938fd9b4b430b29383bbb69826c580c54b)

crates/script/src/verify.rs
```diff
@@ -91,16 +108,55 @@ impl VerifyBundle {
...
+            // An explicit non-Sourcify endpoint may be private. Do not disclose provenance to the
+            // public Sourcify service as an implicit fallback.
+            self.source_sourcify_url = None;
+            self.source_etherscan_url = Some(url.clone());
+            self.source_etherscan_key = resolved_key.clone();
+        } else {
+            self.source_sourcify_url =
+                Some(sourcify_api_url(chain).unwrap_or_else(|| SOURCIFY_URL.to_string()));
+            self.source_etherscan_url =
+                config.get_etherscan_config_with_chain(Some(chain))?.map(|config| config.api_url);
```
mablr: Could we derive the chain endpoint when the selected key came from `--verifier-api-key`? Without `--verifier-url` or `ETHERSCAN_API_KEY`, this returns `None` even on a supported chain, while `resolved_key` selects Etherscan. External discovery then cannot query the selected provider although submission succeeds via `builder.chain(chain)`. Falling back to `chain.etherscan_urls()` (or resolving this config with `resolved_key`) plus a focused test would keep discovery and submission aligned.

stevencartavia: [20b6298](https://github.com/foundry-rs/foundry/commit/20b6298926a597dce60680cf8da5a053fc2097d3)

crates/script/src/lib.rs
```diff
@@ -239,6 +239,10 @@ pub struct ScriptArgs {
     #[arg(long, requires = "broadcast")]
     pub verify: bool,
 
+    /// Opt-in verification of contracts deployed by external factories using explorer source.
+    #[arg(long, requires = "verify")]
```
mablr: `--skip-simulation` leaves `additional_contracts` empty, so this combination can report `All (0) contracts were verified!` without discovering external deployments. Please make `verify_external` conflict with `skip_simulation` and add a parser test.

crates/script/src/verify.rs
```diff
@@ -223,35 +466,82 @@ async fn verify_contracts<FEN: FoundryEvmNetwork>(
...
                     0,
                     init_code.as_ref(),
                     &sequence.libraries,
                     config.evm_version,
                 ) {
-                    Some(verify) => future_verifications.push(verify.run()),
-                    None => unverifiable_contracts.push(*address),
+                    Some(args) => verification_jobs.push(VerificationJob::Local(args)),
+                    None if !verify.verify_external => unverifiable_contracts.push(*address),
+                    None if config.offline => {
```
mablr, line 503: Can we reject `verify_external && config.offline` before broadcasting? At this point deployment has already happened even though external verification is guaranteed to be skipped.

crates/script/src/verify.rs
```diff
@@ -91,16 +108,63 @@ impl VerifyBundle {
...
+                    .unwrap_or_else(|| SOURCIFY_URL.to_string()),
+            );
+            // Etherscan is optional when Sourcify is selected, so an invalid fallback must not
+            // prevent verification through the selected provider.
+            self.source_etherscan_url = config
+                .get_etherscan_config_with_chain(Some(chain))
+                .ok()
+                .flatten()
+                .map(|config| config.api_url);
+            self.source_etherscan_key = config_key;
```
mablr, line 150: If this Etherscan URL resolves to the selected Sourcify URL (for example on Tempo), the fallback sends the ambient Etherscan key to that endpoint. Please suppress the fallback and clear its key when the URLs match.

crates/script/src/verify/external.rs
```diff
@@ -0,0 +1,1356 @@
...
+) -> MatchResult {
+    let mut matches = Vec::new();
+    for candidate in candidates {
+        let bytecode = candidate.creation_bytecode.as_ref();
+        if bytecode.is_empty() {
+            continue;
+        }
+        let Some(suffix) = observed.strip_prefix(bytecode) else { continue };
+        let valid = match &candidate.constructor {
+            None => suffix.is_empty(),
```
mablr, line 776: This rejects valid init code with trailing data. Please accept arbitrary suffixes for no-arg constructors and, for typed constructors, require canonical ABI encoding to be a prefix while preserving the full suffix.

mablr (commented):
Some findings, maybe pre-existing, maybe addressable later.

crates/script/src/verify.rs
stevencartavia: [f3715ecd5](https://github.com/foundry-rs/foundry/commit/f3715ecd5d33dd244b2c245bcfdc4a261b6160fa)

crates/script/src/lib.rs
stevencartavia: [f3715ecd5](https://github.com/foundry-rs/foundry/commit/f3715ecd5d33dd244b2c245bcfdc4a261b6160fa)

crates/script/src/verify/external.rs
stevencartavia: deferring this for a follow-up

crates/script/src/verify.rs
stevencartavia: [f3715ecd5](https://github.com/foundry-rs/foundry/commit/f3715ecd5d33dd244b2c245bcfdc4a261b6160fa)

mablr (approved):

mattsse (approved):

crates/script/src/verify/external.rs
stevencartavia: https://github.com/foundry-rs/foundry/pull/15915

PR #15867, opened 2026-07-23
fix(fuzz): gate minimization predicates

mablr:
Invariant minimization replay evaluated invariant predicates after handler outcomes that the campaign handler-success gate would reject. This mirrors the campaign gate for periodic and final invariant checks, so tmin preserves the campaign-visible failure identity without changing which observations cmin treats as execution errors. `afterInvariant` remains evaluated under its separate campaign semantics, including after a handler failure.

Addresses OSS-579.

This PR was written primarily by OpenAI Codex.


mattsse (approved):
lgtm

stevencartavia (approved):

PR #15868, opened 2026-07-23
fix(fuzz): preserve tmin replay failures

mablr:
Invariant minimization replay retained only one representative failure, allowing tmin candidates to discard secondary handler, predicate, or `afterInvariant` failures. This records every unique replay failure, derives cmin classification from that set, and requires exact baseline/candidate failure-set equality. Replay skips invariant predicates already observed during the sequence and evaluates `afterInvariant` after final predicate checks only when no predicate or hook failure was recorded.

Addresses OSS-578.

This PR was written with OpenAI Codex assistance.


mattsse (dismissed):
lgtm

crates/evm/evm/src/executors/showmap.rs
```diff
@@ -503,7 +506,7 @@ pub fn replay_sequence_for_minimization<FEN: FoundryEvmNetwork>(
                 fingerprint,
             ) {
                 let terminal = failure.is_terminal();
-                record_replay_failure(&mut observation, failure);
+                observation.failures.insert(failure);
```
stevencartavia, line 558: `ReplayFailure::Handler` equality includes the fingerprint, so the same `(target, selector)` reached through different paths becomes multiple failures. The live campaign deduplicates handler assertions by `(target, selector)` and uses the fingerprint only for path-specific shrinking/replay. Reproduced with two branches of one asserting handler: base minimizes 2 → 1 tx, while this PR keeps 2 → 2. Could the failure-set identity use `(target, selector)` while retaining the fingerprint separately?

crates/evm/evm/src/executors/showmap.rs
```diff
@@ -516,12 +519,16 @@ pub fn replay_sequence_for_minimization<FEN: FoundryEvmNetwork>(
...
-                        broken_invariant(&executor, address, target.invariant_fns)?
                 {
-                    record_replay_failure(&mut observation, failure);
+                    for failure in newly_broken_invariants(
+                        &executor,
+                        address,
+                        target.invariant_fns,
+                        &observation.failures,
+                    )? {
+                        observation.failures.insert(failure);
```
stevencartavia, line 582: After all selected predicates fail, the live campaign stops the current run, but replay continues into later transactions here. That can make failures from an unreachable suffix mandatory for `tmin`. Reproduced with `breakBothPredicates() → assertHandler()`: base minimizes 2 → 1 tx, while this PR keeps 2 → 2. Could replay break after this check once every selected invariant name has been recorded?

mablr: https://github.com/foundry-rs/foundry/pull/15868/commits/43c55d8c3b34ee0b80852f374ba32f0f5c2255c4

crates/evm/evm/src/executors/showmap.rs
mablr: https://github.com/foundry-rs/foundry/pull/15868/commits/43c55d8c3b34ee0b80852f374ba32f0f5c2255c4

stevencartavia (commented):
g2g once tests are fixed

mattsse (dismissed):
lgtm

mablr:
@mattsse I made smol naming fix sorry

stevencartavia (approved):

figtracer (approved):

PR #15869, opened 2026-07-23
feat(anvil): `eth_simulateV1` support moved precompiles

mablr:
## Motivation

Adds support for `movePrecompileToAddress` in `eth_simulateV1` state overrides. Each simulated block validates relocations against the active Ethereum precompile set, rejects self-relocations and duplicate destinations with the specified error codes, and applies valid moves only for that block.

Relocating a precompile frees its original address for a bytecode override while preserving protocol-address warming under EIP-2929. Optimism and Tempo retain their existing execution paths.

Closes OSS-560


mattsse (approved):
lgtm

stevencartavia (approved):

PR #15870, opened 2026-07-23
fix(anvil): `eth_simulateV1` trace ETH transfers

mablr:
## Motivation

Corrects `eth_simulateV1` transfer tracing by collecting synthetic ETH transfer logs separately from EVM state. Transfers from top-level and internal calls, `CREATE`, `CREATE2`, and `SELFDESTRUCT` are interleaved with canonical logs in execution order.

Logs from reverted frames are discarded while attempted log indices remain observable as gaps. Synthetic transfer logs are excluded from the block bloom, and simulations without `traceTransfers` retain their existing behavior.

Closes OSS-567

### DIFF
--- a/crates/anvil/src/eth/backend/mem/inspector.rs
+++ b/crates/anvil/src/eth/backend/mem/inspector.rs
@@ -1,7 +1,8 @@
 //! Anvil specific [`revm::Inspector`] implementation
 
 use crate::eth::macros::node_info;
-use alloy_primitives::{Address, Log, U256};
+use alloy_primitives::{Address, B256, Log, LogData, U256};
+use alloy_sol_types::SolValue;
 use foundry_evm::{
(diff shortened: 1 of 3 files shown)


mattsse (changes requested):
`test_simulate_preserves_amsterdam_transfer_logs_rpc` fails at [this assertion](https://github.com/foundry-rs/foundry/blob/6c2a97015ceb6d75dacd4884423252d438c91b74/crates/anvil/tests/it/simulate.rs#L893) because Amsterdam’s canonical EIP-7708 emitter is `0xfffffffffffffffffffffffffffffffffffffffe`, not the synthetic trace-transfer address. Please update the expected address.

mattsse (approved):
lgtm

stevencartavia (approved):

PR #15871, opened 2026-07-23
fix(config): load symlinked dependency configs

mablr:
Symlinked dependencies can have configuration and remappings that are missed when discovery canonicalizes their paths.

This change preserves each dependency’s lexical install path for emitted remappings while using canonical paths only for config loading, deduplication, and cycle detection. It loads direct dependency `foundry.toml` and `remappings.txt` entries, supports custom source directories, preserves existing physical dependency behavior, and avoids recursively traversing symlinked dependency graphs.

Added coverage for builds and remapping output for both physical and symlinked dependencies, including custom source layouts and standard `src` directories.

Closes OSS-582


mattsse (changes requested):
The standard dependency source remapping is dropped by [this condition](https://github.com/foundry-rs/foundry/blob/ae29e578673afd73dc8a780ad0c7c57b3344206a/crates/config/src/providers/remappings.rs#L311-L327). This regresses `can_detect_lib_foundry_toml`, `can_prioritise_closer_lib_remappings`, and `can_prioritise_project_remappings`: dependencies using `src`, `contracts`, or `lib` no longer get a package remapping when the directory does not exist yet, and closest-remapping precedence can change. Please synthesize the declared source remapping for every nested config, not only nonstandard source directories. Removing this condition makes all three tests pass locally.

crates/config/src/providers/remappings.rs
```diff
@@ -246,55 +256,74 @@ impl RemappingsProvider<'_> {
...
+            } else {
+                if failed.contains(&entry.canonical) {
+                    continue;
+                }
+
+                trace!(lib = ?entry.canonical, "find all remappings of nested foundry.toml");
+                // Nested configs are traversed here, so load their declared remappings without
+                // recursively installing another remappings provider.
+                let figment =
+                    Config::with_root(&entry.canonical).to_figment(FigmentProviders::Cast);
```
mattsse: Using `FigmentProviders::Cast` skips `RemappingsProvider` entirely, so nested `remappings.txt` files are dropped. This is why `snekmate` loses the `@openzeppelin/contracts/` mapping from `lib/properties/remappings.txt`. Keep traversal non-recursive but still merge each dependency’s `remappings.txt`, for example by installing a provider with auto-detection disabled and no library paths, or by parsing and merging that file explicitly. Please add focused coverage for this case.

crates/config/src/providers/remappings.rs
```diff
@@ -246,55 +256,74 @@ impl RemappingsProvider<'_> {
...
+                        .into_iter()
+                        .map(|entry| (entry, ancestors.clone())),
+                );
+            }
 
-        if let Some(r) = src_remapping {
-            remappings.push(r);
+            // Synthesize the declared source remapping for every lexical alias. The source
+            // directory may not exist yet, so `Remapping::find_many` cannot always detect it.
+            if let Some(name) = entry.path.file_name().and_then(|name| name.to_str()) {
```
mattsse: Always pointing the package alias at its configured source directory breaks dependencies whose imports intentionally include that directory, such as `forge-std/src/...`; `prb_math` now resolves these as `node_modules/forge-std/src/src/...`. Only synthesize `src`, `contracts`, or `lib` when the declared directory is missing, and otherwise let `Remapping::find_many` preserve the package-root mapping. Custom source directories should continue to be synthesized.

mattsse (dismissed):
The external suite still has two deterministic remapping regressions; details and fix suggestions inline.

crates/config/src/providers/remappings.rs
stevencartavia: +1 from my pass

mablr:
Reworked the PR https://github.com/foundry-rs/foundry/pull/15871/commits/9a596f5567bf88b84ffaf4a1bd0a88e5f0295f88, see updated description.

crates/config/src/providers/remappings.rs
```diff
@@ -339,61 +355,92 @@ impl RemappingsProvider<'_> {
...
-    }
+            .iter()
+            .map(|path| if path.is_absolute() { path.clone() } else { self.root.join(path) })
+            .flat_map(foundry_toml_dir_entries)
+            .collect::<BTreeSet<_>>();
+        let mut seen = HashSet::from([root]);
+        let mut remappings = Vec::new();
+
+        while let Some(entry) = pending.pop_first() {
+            if !seen.insert(entry.canonical.clone()) {
```
figtracer: this drops every lexical alias after the first one that resolves to the same dependency. with `lib/a-alias` and `lib/z-alias` symlinked to a package using `src = "custom-source"`, `forge remappings` emits `z-alias/=lib/z-alias/` and `forge build` cannot resolve `z-alias/Dep.sol`. can we cache config loading by canonical path but emit and rebase remappings for each direct lexical entry?

figtracer (dismissed):

crates/config/src/providers/remappings.rs
```diff
@@ -339,61 +364,84 @@ impl RemappingsProvider<'_> {
...
+            }));
+            remappings.extend(config.file_remappings.iter().cloned().map(|remapping| {
+                (entry.path.clone(), RelativeRemapping::new(remapping, &entry.path).into(), false)
+            }));
+
+            // Preserve existing physical nested-config discovery. Symlink discovery is limited to
+            // direct configured-library entries; nested symlink graphs remain out of scope.
+            if seen.insert(entry.canonical.clone())
+                && !fs::symlink_metadata(&entry.path)
+                    .is_ok_and(|metadata| metadata.file_type().is_symlink())
```
grandizzy: **[P2] Don’t mark symlink aliases as traversed**

`seen.insert()` runs before the symlink check. If a symlink alias sorts before a physical entry resolving to the same canonical dependency, it consumes the canonical identity without traversing nested libraries. The later physical entry then skips traversal.

Reorder the conditions so only physical entries mutate `seen`:

```suggestion
            if !fs::symlink_metadata(&entry.path)
                .is_ok_and(|metadata| metadata.file_type().is_symlink())
                && seen.insert(entry.canonical.clone())
```

Please also add coverage where a symlink alias sorts before a physical installation of the same dependency.

grandizzy (dismissed):
Previous review comments are addressed, but there is one remaining ordering-dependent correctness issue in nested dependency traversal.

crates/config/src/providers/remappings.rs
stevencartavia: +1

mablr: https://github.com/foundry-rs/foundry/pull/15871/commits/a169d52d312adb5d510796d99e97df2829fe17b3

grandizzy (approved):
lgtm!

figtracer (approved):

PR #15872, opened 2026-07-23
fix(config): preserve broad dependency aliases

mablr:
## Motivation

Close OSS-581

This preserves broader auto-detected dependency remappings when the root project defines a narrower alias in the same namespace. The narrower root mapping remains authoritative through longest-prefix resolution, while the broader dependency mapping provides a fallback for imports outside that subtree. Root mappings still suppress equal or more specific dependency aliases, preserving the existing protection against dependencies overriding project code. Carefully assessed it will not reopen #9146.

For example, `pkg/sub/=src/local/` can now coexist with `pkg/=lib/pkg/src/`: imports under `pkg/sub/` resolve locally, while other `pkg/` imports resolve from the dependency.

### DIFF
--- a/crates/config/src/providers/remappings.rs
+++ b/crates/config/src/providers/remappings.rs
@@ -79,22 +79,19 @@ impl Remappings {
                     && existing.path == remapping.path;
             }
 
-            // What we're doing here is filtering for ambiguous paths. For example, if we have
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm

mattsse (approved):
lgtm

stevencartavia (approved):

PR #15875, opened 2026-07-24
fix: preserve dumpState deployment order

### LINKED ISSUE #8503: `vm.dumpState` inconsistent address ordering w.r.t. deployment order
### Component

Forge

- [X] Foundry
- [X] Foundryup

forge 0.2.0 (62cdea8 2024-07-23T00:19:34.618350138Z)

forge script

### Operating System

Linux

### Describe the bug

The order of the address entries in the JSON produced by `vm.dumpState` is inconsistent regarding the order in which the contracts are deployed.
In the example below, we orderly deploy `counter0`, `counter1`, and `counter2`, setting `counterN.number()` to `N`. However, the order of the accounts in the produced JSON is `counter0`, `counter2`, `counter1`.

Note that ordering the entries w.r.t. deployment order might be one of many suitable orderings (it could be ordered w.r.t. storage updates, for instance). However, keeping the deployment order in the JSON might maximize its readability.

### Reproduction steps

1. `forge init dumpState`
2. Append the following function to `test/Counter.t.sol`:
    ```solidity
    function multiple_counters() public {                                                                                               
        for (uint256 i; i <= 2; ++i) {                                                                                                  
            counter = new Counter();                                                                                                    
            counter.setNumber(i);                                                                                                       
        }                                                                                                                               
        vm.dumpState(string.concat(vm.projectRoot(), "/dumpState.json"));                                                               
    }
    ```
3. `forge script ./test/Counter.t.sol:CounterTest --sig multiple_counters ; cat dumpState.json | jq .`

Executing step 3 produces the following JSON containing the accounts in a different order than deployed (0, 2, 1 instead of 0, 1, 2):

```json
{
  "0x2e234dae75c793f67a35089c9d99245e1c58470b": {
    "nonce": "0x1",
    "balance": "0x0",
    "code": "0x6080604052348015600f57600080fd5b5060043610603c5760003560e01c80633fb5c1cb1460415780638381f58a146053578063d09de08a14606d575b600080fd5b6051604c3660046083565b600055565b005b605b60005481565b60405190815260200160405180910390f35b6051600080549080607c83609b565b9190505550565b600060208284031215609457600080fd5b5035919050565b60006001820160ba57634e487b7160e01b600052601160045260246000fd5b506001019056fea2646970667358221220091e48831e9eee32d4571d6291233a4fdaaa34b7dced8770f36f5368be825c5264736f6c63430008190033",
    "storage": {
      "0x0000000000000000000000000000000000000000000000000000000000000000": "0x0000000000000000000000000000000000000000000000000000000000000000"
    }
  },
  "0x5991a2df15a8f6a256d3ec51e99254cd3fb576a9": {
    "nonce": "0x1",
    "balance": "0x0",
    "code": "0x6080604052348015600f57600080fd5b5060043610603c5760003560e01c80633fb5c1cb1460415780638381f58a146053578063d09de08a14606d575b600080fd5b6051604c3660046083565b600055565b005b605b60005481565b60405190815260200160405180910390f35b6051600080549080607c83609b565b9190505550565b600060208284031215609457600080fd5b5035919050565b60006001820160ba57634e487b7160e01b600052601160045260246000fd5b506001019056fea2646970667358221220091e48831e9eee32d4571d6291233a4fdaaa34b7dced8770f36f5368be825c5264736f6c63430008190033",
    "storage": {
      "0x0000000000000000000000000000000000000000000000000000000000000000": "0x0000000000000000000000000000000000000000000000000000000000000002"
    }
  },
  "0xf62849f9a0b5bf2913b396098f7c7019b51a820a": {
    "nonce": "0x1",
    "balance": "0x0",
    "code": "0x6080604052348015600f57600080fd5b5060043610603c5760003560e01c80633fb5c1cb1460415780638381f58a146053578063d09de08a14606d575b600080fd5b6051604c3660046083565b600055565b005b605b60005481565b60405190815260200160405180910390f35b6051600080549080607c83609b565b9190505550565b600060208284031215609457600080fd5b5035919050565b60006001820160ba57634e487b7160e01b600052601160045260246000fd5b506001019056fea2646970667358221220091e48831e9eee32d4571d6291233a4fdaaa34b7dced8770f36f5368be825c5264736f6c63430008190033",
    "storage": {
      "0x0000000000000000000000000000000000000000000000000000000000000000": "0x0000000000000000000000000000000000000000000000000000000000000001"
    }
  }
}
```

stevencartavia:
Preserves deployment order for accounts created in the current transaction when `vm.dumpState` writes genesis allocations, while retaining deterministic address ordering for all remaining accounts. Creation tracking follows reverts, snapshots, and active forks so the output reflects the surviving state.

Closes #8503.

### DIFF
--- a/crates/cheatcodes/src/evm.rs
+++ b/crates/cheatcodes/src/evm.rs
@@ -57,7 +57,7 @@ mod record_debug_step;
 use foundry_common::fmt::format_token_raw;
 use foundry_config::{ExecutionSpec, evm_spec_id_from_str};
 use record_debug_step::{convert_call_trace_ctx_to_debug_step, flatten_call_trace};
-use serde::Serialize;
+use serde::{Serialize, Serializer, ser::SerializeMap};
 
 mod fork;
 pub(crate) mod mapping;
@@ -74,6 +74,21 @@ struct LogJson {
     data: String,
     /// The address of the log's emitter.
     emitter: String,
+}
+
+struct StateDump<'a>(&'a [(Address, GenesisAccount)]);
+
+impl Serialize for StateDump<'_> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut map = serializer.serialize_map(Some(self.0.len()))?;
+        for (address, account) in self.0 {
+            map.serialize_entry(address, account)?;
+        }
+        map.end()
+    }
 }
 
 /// Records storage slots reads and writes.
@@ -361,6 +376,9 @@ impl Cheatcode for dumpStateCall {
         let Self { pathToStateJson } = self;
         let path = Path::new(pathToStateJson);
 
+        let fork_id = ccx.ecx.db().active_fork_id();
+        let created_accounts = ccx.state.created_accounts(fork_id).copied().collect::<Vec<_>>();
+
         // Do not include system account or empty accounts in the dump.
         let skip = |key: &Address, val: &Account| {
             key == &CHEATCODE_ADDRESS
@@ -372,16 +390,24 @@ impl Cheatcode for dumpStateCall {
                 || val.is_empty()
         };
 
-        let alloc = ccx
+        let mut alloc = ccx
             .ecx
             .journal_mut()
             .evm_state_mut()
             .iter_mut()
             .filter(|(key, val)| !skip(key, val))
-            .map(|(key, val)| (key, genesis_account(val)))
+            .map(|(key, val)| (*key, genesis_account(val)))
             .collect::<BTreeMap<_, _>>();
 
-        write_json_file(path, &alloc)?;
+        let mut ordered_alloc = Vec::with_capacity(alloc.len());
+        for address in created_accounts {
+            if let Some(account) = alloc.remove(&address) {
+                ordered_alloc.push((address, account));
+            }
+        }
+        ordered_alloc.extend(alloc);
+
+        write_json_file(path, &StateDump(&ordered_alloc))?;
         Ok(Default::default())
     }
 }
@@ -1016,6 +1042,7 @@ impl Cheatcode for deleteSnapshotCall {
         let Self { snapshotId } = self;
         let result = ccx.ecx.db_mut().delete_state_snapshot(*snapshotId);
         ccx.state.env_overrides_snapshots.remove(snapshotId);
+        ccx.state.delete_created_accounts_snapshot(*snapshotId);
         Ok(result.abi_encode())
     }
 }
@@ -1025,6 +1052,7 @@ impl Cheatcode for deleteStateSnapshotCall {
         let Self { snapshotId } = self;
         let result = ccx.ecx.db_mut().delete_state_snapshot(*snapshotId);
         ccx.state.env_overrides_snapshots.remove(snapshotId);
+        ccx.state.delete_created_accounts_snapshot(*snapshotId);
         Ok(result.abi_encode())
     }
 }
@@ -1035,6 +1063,7 @@ impl Cheatcode for deleteSnapshotsCall {
         let Self {} = self;
         ccx.ecx.db_mut().delete_state_snapshots();
         ccx.state.env_overrides_snapshots.clear();
+        ccx.state.clear_created_accounts_snapshots();
         Ok(Default::default())
     }
 }
@@ -1044,6 +1073,7 @@ impl Cheatcode for deleteStateSnapshotsCall {
         let Self {} = self;
         ccx.ecx.db_mut().delete_state_snapshots();
         ccx.state.env_overrides_snapshots.clear();
+        ccx.state.clear_created_accounts_snapshots();
         Ok(Default::default())
     }
 }
@@ -1251,6 +1281,7 @@ impl Cheatcode for executeTransactionCall {
         let sender =
             tx.recover_signer().map_err(|err| fmt_err!("failed to recover signer: {err}"))?;
         let tx_env = TxEnvFor::<FEN>::from_recovered_tx(&tx, sender);
+        let created_address = tx_env.kind().is_create().then(|| sender.create(tx_env.nonce()));
 
         // Save current env for restoration after execution.
         let cached_evm_env = ccx.ecx.evm_clone();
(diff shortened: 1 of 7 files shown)


crates/cheatcodes/src/inspector.rs
```diff
@@ -839,6 +864,94 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
             .unwrap_or_default()
     }
 
+    /// Returns successfully created accounts for the active fork in execution order.
+    pub(crate) fn created_accounts(
+        &self,
+        fork_id: Option<LocalForkId>,
+    ) -> impl Iterator<Item = &Address> {
+        self.created_accounts.iter().filter_map(move |(created_fork_id, address)| {
+            (*created_fork_id == fork_id).then_some(address)
```
mablr: Filtering only by the fork where an account was created loses deployment order for persistent accounts. Repro: create A on fork 1, `makePersistent(A)`, switch to fork 2, create B, then `dumpState`. A survives in fork 2's allocation but is filtered out here, so B is emitted first and A falls into the address-sorted suffix. Please preserve creation records for migrated persistent accounts and add a focused fork regression test.

stevencartavia: f0f7387b85216b44b8488b3a09d7ae395b13af85

crates/cheatcodes/src/evm.rs
```diff
@@ -361,6 +376,16 @@ impl Cheatcode for dumpStateCall {
         let Self { pathToStateJson } = self;
         let path = Path::new(pathToStateJson);
 
+        let fork_id = ccx.ecx.db().active_fork_id();
+        let created_accounts = ccx
+            .state
+            .created_accounts()
+            .filter_map(|(created_fork_id, address)| {
+                (*created_fork_id == fork_id || ccx.ecx.db().is_persistent(address))
```
mablr: Could we track whether this creation was propagated to the active fork instead of checking current persistence? After `makePersistent(A) -> selectFork -> revokePersistent(A)`, `A` remains in the active fork state but fails this filter; deploying `B` then calling `dumpState` emits `B` before `A`. Please add a regression test for this sequence.

stevencartavia: 127bc3a3d3a79c1ce1227529a5a143400c57e755

crates/cheatcodes/src/evm/fork.rs
```diff
@@ -410,6 +428,31 @@ fn fork_env_op<CTX: FoundryContextExt, T: SolValue>(
...
+
+fn record_fork_switch<FEN: FoundryEvmNetwork>(
+    ccx: &mut CheatsCtxt<'_, '_, FEN>,
+    source_fork_id: Option<LocalForkId>,
+    initial: AddressHashMap<usize>,
+    propagated: Vec<(Address, usize)>,
+) {
+    let target_fork_id = ccx.ecx.db().active_fork_id();
+    if source_fork_id != target_fork_id {
+        ccx.state.commit_created_account_changes(source_fork_id);
```
mablr, line 458: Could we also commit creation changes after a successful active `rollFork`? The backend preserves created accounts across the roll, but an enclosing revert drops their tracker bindings, so `dumpState` can emit an older deployment after a newer one.

stevencartavia: 845f318fc4be74b016ae0f67470c7101a709002e

figtracer (approved):

crates/cheatcodes/src/evm.rs
```diff
@@ -1281,6 +1319,10 @@ impl Cheatcode for executeTransactionCall {
         // Mark as inner context so isolation mode doesn't trigger a nested transact_inner
         // when the inner EVM executes calls at depth == 1.
         executor.set_in_inner_context(true, Some(sender));
+        if let Some(address) = created_address {
+            let fork_id = ccx.ecx.db().active_fork_id();
+            ccx.state.record_created_account(fork_id, address);
```
mablr, line 1324: Follow-up (non-blocking): could we verify whether this explicit record is needed for EOF or another top-level creation variant? The nested EVM appears to use the same cheatcode inspector, so a normal legacy create transaction may also hit the regular `create` hook and record this address a second time. The active binding keeps the output correct, but if all supported transaction formats reach that hook, removing this would leave a single source of truth.

testdata/default/cheats/dumpState.t.sol
```diff
@@ -119,6 +137,100 @@ contract DumpStateTest is Test {
         vm.removeFile(path);
     }
 
+    function testDumpStateDeploymentOrder() public {
+        string memory path = string.concat(vm.projectRoot(), "/fixtures/Json/test_dump_state_deployment_order.json");
+
+        SimpleContract first = new SimpleContract();
+        SimpleContract second = new SimpleContract();
+        SimpleContract third = new SimpleContract();
+        vm.dumpState(path);
```
mablr, line 146: Follow-up (non-blocking): could we add focused coverage for the complete ordering contract here? In addition to created accounts retaining CREATE invocation order, the test could include arbitrary low-address accounts before/between deployments to prove the non-created suffix remains address-sorted, plus a parent whose constructor deploys a child to make the intended parent-before-child semantics explicit.

testdata/default/cheats/Fork2.t.sol
```diff
@@ -146,6 +147,174 @@ contract ForkTest is Test {
         assertEq(dummy.val(), expectedValue);
     }
 
+    function testForkDumpStatePreservesPersistentDeploymentOrder() public {
```
mablr, line 150: Follow-up (non-blocking): could the remaining fork-tracker transitions be locked down either with focused integration cases (`createSelectFork`, `createSelectFork(..., txHash)`, explicit-ID active rolls, and rolling a non-active fork before selection) or with a compact Rust transition test over the underlying bindings? The goal is to cover active/non-active bookkeeping and commit/revert invariants without necessarily testing both layers.

mablr (commented):
Non-blocking follow-ups suggestions.

mablr (approved):

mattsse (approved):
lgtm

crates/cheatcodes/src/evm.rs
stevencartavia: https://github.com/foundry-rs/foundry/pull/15926

testdata/default/cheats/dumpState.t.sol
stevencartavia: https://github.com/foundry-rs/foundry/pull/15926

testdata/default/cheats/Fork2.t.sol
stevencartavia: https://github.com/foundry-rs/foundry/pull/15927

PR #15883, opened 2026-07-25
fix(traces): bound external identification

### LINKED ISSUE #12714: perf: printing traces is too slow

stevencartavia:
Bounds external trace identification with a cumulative timeout so stalled Sourcify or Etherscan requests cannot indefinitely block trace rendering. Setting the timeout to zero or passing `--disable-external-identification` disables external lookups without preventing compiler downloads.

Fixes #12714.

### DIFF
--- a/crates/cast/src/cmd/call.rs
+++ b/crates/cast/src/cmd/call.rs
@@ -258,7 +258,8 @@ impl CallArgs {
         let mut config = load_config_from_provider(figment)?;
         let state_overrides = self.get_state_overrides()?;
         let block_overrides = self.get_block_overrides()?;
-        let tracing = self.resolve_tracing(&config.tracing, shell::verbosity());
+        config.tracing = self.resolve_tracing(&config.tracing, shell::verbosity());
+        let tracing = config.tracing.clone();
 
         let Self {
             to,
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -5931,6 +5931,7 @@ mod tests {
                 compact_labels = true
                 trace_depth = 3
                 decode_internal = true
+                external_identification_timeout = 9
 
                 [tracing.labels]
                 0x0000000000000000000000000000000000000002 = "Bob"
@@ -5944,6 +5945,7 @@ mod tests {
             assert_eq!(config.tracing.trace_depth, Some(3));
             assert!(config.tracing.decode_internal);
             assert!(config.tracing.compact_labels);
+            assert_eq!(config.tracing.external_identification_timeout, 9);
             let labels = AddressHashMap::from_iter(vec![(
                 address!("0x0000000000000000000000000000000000000002"),
                 "Bob".to_string(),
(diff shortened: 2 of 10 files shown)


mablr (approved):
sgtm

mablr:
maybe book update is needed

figtracer (approved):

crates/config/src/lib.rs
```diff
@@ -5966,6 +5968,18 @@ mod tests {
...
+    fn test_external_identification_timeout_env() {
+        figment::Jail::expect_with(|jail| {
+            jail.set_env("FOUNDRY_TRACING_EXTERNAL_IDENTIFICATION_TIMEOUT", "0");
+
+            let config = Config::load().unwrap();
+
+            assert_eq!(config.tracing.external_identification_timeout, 0);
+            Ok(())
+        });
+    }
```
mablr, line 5981: Non-blocking: this would make backward compatibility for existing partial `[tracing]` sections explicit rather than incidental.

```suggestion
    }

    #[test]
    fn test_external_identification_timeout_defaults_for_partial_tracing_config() {
        figment::Jail::expect_with(|jail| {
            jail.create_file("foundry.toml", "[tracing]\ndisable_labels = true")?;

            let config = Config::load().unwrap();

            assert_eq!(config.tracing.external_identification_timeout, 5);
            Ok(())
        });
    }
```

crates/cast/src/cmd/call.rs
```diff
@@ -258,7 +258,8 @@ impl CallArgs {
         let mut config = load_config_from_provider(figment)?;
         let state_overrides = self.get_state_overrides()?;
         let block_overrides = self.get_block_overrides()?;
-        let tracing = self.resolve_tracing(&config.tracing, shell::verbosity());
+        config.tracing = self.resolve_tracing(&config.tracing, shell::verbosity());
```
mablr, line 261: Non-blocking follow-up: could we add one command-level regression test for `--disable-external-identification` (here or on `forge test`)? The shared resolver tests protect the flag-to-timeout mapping, but not the command-specific propagation into the config consumed by trace identification.

mablr (commented):
Non-blocking follow-up suggesions

crates/config/src/lib.rs
stevencartavia: https://github.com/foundry-rs/foundry/pull/15914

crates/cast/src/cmd/call.rs
stevencartavia: https://github.com/foundry-rs/foundry/pull/15914

PR #353, opened 2022-01-01
fix: fuzzer handles int/uint size generation correctly

mds1:
Closes https://github.com/gakonst/foundry/issues/341 by always generating a U256 and wrapping to the desired size

Question—How can I verify the values being output by the strategy are in the expected range? I did some spot checks with solidity tests but not sure how to get more insight

Also, one thing that's a bit annoying (but I think can be handled in a separate issue) is if you write a test like the below:

```solidity
function testLimits(int40 a) public {
    assertGe(a, 0);
  }
```

you get a failure case like this, where I'd expect the int to be shown as a negative value instead of a uint as currently shown:

`[FAIL. Reason: . Counterexample: calldata=0x4db19789ffffffffffffffffffffffffffffffffffffffffffffffffffffff8000000000, args=[Int(115792089237316195423570985008687907853269984665640564039457584007363373826048)]]`

PR #15885, opened 2026-07-26
refactor(tempo): use Alloy Accounts wallet

gakonst:
Foundry currently duplicates Tempo Accounts persistence, key selection, authorization reconstruction, keychain signatures, sponsor ordering, and paid-RPC state across Cast, Forge, scripts, and transports. This change makes `tempo-alloy`’s concrete `TempoAccountsWallet` and `$TEMPO_HOME/wallet/store.json` the single implementation and persistence boundary, while preserving the exact prepared wallet through estimation, sponsorship, and signing.

The branch removes the parallel `keys.toml` and local MPP session implementations, routes automatic paid RPC through mpp-rs’s Charge provider, and keeps both HTTPS and WebSocket Charge handshakes. Payment handling now serializes each challenge-to-acknowledgement lifecycle, commits only on explicit acceptance, rolls back only before send or after definite rejection, and checks the complete call scope before launching Accounts authorization. It also fixes Tempo AA CREATE detection and exercises the resulting flow across Cast, Forge, scripts, forks, Anvil, Chisel, and NanoUSD MPP.

Dependency stack: tempoxyz/tempo#6965 at `9081cc659`, mpp-rs #353/#354/#355 at `5caad442`, and merged foundry-rs/foundry-core#129 at `1d3c2590`. The lockfile contains one Tempo SDK revision and drops the duplicate node dependency path. Overall this deletes roughly 3,000 net lines.


gakonst:
cyclops audit

PR #15888, opened 2026-07-26
docs(cli): add usage examples to command help

mattsse:
Only a handful of commands carry usage examples in their help text, so both local --help output and the generated CLI reference on getfoundry.sh show flags without a single concrete invocation. For the long tail of cast commands that reference is the only documentation page, which makes the first working command line unnecessarily hard to find — for people and for coding agents alike.

This adds short Examples blocks to the doc comments of the highest-traffic commands: 27 cast commands (including the wallet family), 11 forge commands, and the anvil top level — 39 blocks total, 2-4 lines each, following the format the existing to-unit/to-rlp examples established. Since clap collapses doc-comment lines in help output (the pre-existing example lists rendered as a single run-on paragraph), the touched commands get verbatim_doc_comment, which also fixes the four existing lists to render line by line; their first-line summaries lose their source periods to keep the Commands lists uniformly period-free, since verbatim also disables clap's trailing-period stripping.

Every example was verified against the current argument definitions and nearly all were executed end-to-end (offline commands verbatim; chain commands against a local anvil; project commands in a scratch forge project). That process caught two orderings the examples must not teach: parent flags have to precede the --create subcommand of cast send, and --constructor-args has to come last in forge create.

The examples reach the book automatically through the weekly generated-reference update; no book changes are needed.

Written with Claude Code, including an adversarial review pass that executed the documented commands.

### DIFF
--- a/crates/anvil/src/opts.rs
+++ b/crates/anvil/src/opts.rs
@@ -3,9 +3,15 @@ use clap::{Parser, Subcommand};
 use foundry_cli::opts::GlobalArgs;
 use foundry_common::version::{LONG_VERSION, SHORT_VERSION};
 
-/// A fast local Ethereum development node.
+/// A fast local Ethereum development node
+///
+/// Examples:
+/// - anvil (start a local node on 127.0.0.1:8545)
+/// - anvil --fork-url $RPC_URL (fork the latest state of a live network)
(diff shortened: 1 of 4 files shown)


mablr (approved):
cool

grandizzy (approved):
👍

PR #15898, opened 2026-07-27
fix(anvil): disable pre-London simulation fees

### LINKED ISSUE #15895: bug(anvil): pre-London simulation calculates base fees
Component: Anvil

`eth_simulateV1` uses the raw next-block base-fee calculator. That path bypasses the configured zero-fee sentinel as intended, but it also bypasses hardfork activation and calculates an EIP-1559 base fee for pre-London execution.

When the active hardfork is Berlin or earlier, simulated blocks should keep `baseFeePerGas` at zero. The raw calculator should gate EIP-1559 logic on London activation while preserving the existing London, post-London, and network-specific fee behavior.

haythemsellami:
The raw next-block fee calculator used by `eth_simulateV1` bypasses the normal zero-fee sentinel, but it also calculated EIP-1559 fees before London. This change gates that calculation on hardfork activation so Berlin simulations keep a zero base fee while London and network-specific fee behavior remain unchanged.

The regression simulates multiple pre-London blocks through the RPC and checks that each reported base fee remains zero. Closes #15895.

AI assistance was used for implementation and code review.

### DIFF
--- a/crates/anvil/src/eth/fees.rs
+++ b/crates/anvil/src/eth/fees.rs
@@ -185,6 +185,9 @@ impl FeeManager {
         gas_limit: u64,
         last_fee_per_gas: u64,
     ) -> u64 {
+        if !self.is_eip1559() {
+            return 0;
+        }
         // Tempo replaces EIP-1559 with its own hardfork-specific base fee rules.
         if let Some(hardfork) = self.tempo_hardfork() {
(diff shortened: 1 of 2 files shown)


grandizzy (approved):
makes sense, ty!

mablr (approved):

PR #15899, opened 2026-07-27
fix(anvil): prepare local block replays

### LINKED ISSUE #15894: bug(anvil): local block replay skips pre-execution
Component: Anvil

Parity block replay and `trace_blockOpcodeGas` reconstruct the parent state and then execute the block's transactions directly. They skip the network-aware block pre-execution phase used by mining, so protocol state changes such as the EIP-2935 parent-hash system call are absent from replayed execution.

Locally mined block replay should prepare its database through the same block executor pre-execution path as mining before replaying transactions. This also keeps future network-specific pre-execution behavior consistent across tracing APIs.

haythemsellami:
Local parity block replay and opcode-gas replay start from the parent state but currently skip the protocol pre-execution performed during mining. This change prepares both replay databases through the network-aware block executor before applying transactions, keeping EIP-2935 and future pre-execution behavior aligned with mined execution. Executor preparation is fallible so real preparation and execution failures propagate as JSON-RPC errors instead of panicking or being treated as missing state.

The regressions cover the EIP-2935 parent-hash state change and deterministic pre-execution failures across block replay, transaction replay, and opcode-gas replay. Closes #15894.


crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -831,6 +831,36 @@ impl<N: Network> Backend<N> {
...
+        block: &Block,
+        parent_state: &'a StateDb,
+    ) -> (CacheDB<&'a StateDb>, EvmEnv) {
+        let mut cache_db = AnvilCacheDB::new(parent_state);
+        let evm_env = self.tx_replay_evm_env(&block.header);
+        let spec_id = *evm_env.spec_id();
+        let inspector_tx_config = self.inspector_tx_config();
+        let gas_config = self.pool_tx_gas_config(&evm_env);
+
+        let _ = self.execute_with_block_executor(
```
mablr: This can panic because `execute_with_block_executor` uses `expect` for pre-execution changes, which may fail on fork DB reads. Replay RPCs should propagate that error; please make this preparation fallible and thread the `Result` through both replay paths.

stevencartavia: +1 from my pass

haythemsellami:
Updated in `72f8344c1`: block-executor setup is now fallible, and replay callers propagate preparation and finalization failures as JSON-RPC internal errors. The parity helpers distinguish genuine absence from execution failures, while opcode-gas and the other fallible replay paths propagate the same error. HTTP coverage now exercises block replay, transaction replay, and opcode-gas replay with a deterministic EIP-2935 pre-execution failure and confirms the node remains responsive.

mablr (dismissed):

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -4249,11 +4292,9 @@ where
         let parent_hash = block.header.parent_hash;
 
         let trace = |parent_state: &StateDb| -> Result<Vec<TransactionOpcodeGas>, BlockchainError> {
-            let mut cache_db = CacheDB::new(Box::new(parent_state));
+            let (mut cache_db, evm_env) = self.prepare_block_replay(block, parent_state)?;
```
grandizzy, line 4902: The early return above bypasses `prepare_block_replay`, so `trace_blockOpcodeGas` does not run or propagate pre-execution failures for empty blocks. For example, replaying an empty Prague block with a failing history precompile succeeds with `transactions: []`, while `trace_replayBlockTransactions` returns the expected preparation error. Could we remove this shortcut and let the normal preparation path run with an empty transaction loop?

grandizzy (commented):
Looks good, please check finding inline

crates/anvil/src/eth/backend/mem/mod.rs
haythemsellami: Removed the post-genesis empty-block shortcut, so opcode-gas replay now performs block preparation before iterating transactions. The EIP-2935 regression covers an empty Prague block and verifies the pre-execution error is propagated, while preserving genesis replay because genesis has no parent state.

mablr (approved):

stevencartavia (approved):

PR #15900, opened 2026-07-27
fix(evm): publish fork transactions atomically

### LINKED ISSUE #15892: bug(evm): fork transactions publish state non-atomically
Component: Forge / EVM backend

Fork transaction application commits the transaction changeset to the fork database before refreshing the active and fork journaled states. If either refresh needs an uncached remote account or storage slot and that lookup fails, the caller receives an error after part of the transaction state has already been published.

The operation should be atomic: stage the changeset in an overlay, refresh cloned journals against that overlay, and publish the database and both journals only after every required read succeeds. A regression test can close the shared backend, force a missing storage lookup during refresh, and verify that balances, nonces, storage, and journals remain unchanged.

This was found while reviewing #15343, but it is pre-existing network-neutral Foundry behavior and should be fixed independently.

haythemsellami:
Fork transaction application currently commits its changeset before both journaled states have been refreshed. A failed remote read can therefore return an error after partial state has escaped. This change stages the commit against a clone of the active fork cache, refreshes cloned journals against that exact database, and publishes the database and journals only after every read succeeds.

The regressions close the shared backend during required storage refreshes and cover a locally nonexistent account becoming touched, ensuring staged reads preserve the same account-state semantics as the published database. Closes #15892.


crates/evm/core/src/backend/mod.rs
```diff
@@ -2249,9 +2249,25 @@ fn apply_state_changeset<N: Network, B: ForkBlockEnv>(
     fork: &mut Fork<N, B>,
     persistent_accounts: &HashSet<Address>,
 ) -> Result<(), BackendError> {
-    // commit the state and update the loaded accounts
+    // Refresh cloned journals against an overlay so a failed database read cannot publish only
+    // part of the transaction state.
+    let mut staged_journaled_state = journaled_state.clone();
+    let mut staged_fork_journaled_state = fork.journaled_state.clone();
+    {
+        let mut staged_db = CacheDB::new(&mut fork.db);
```
mablr: A fresh outer `CacheDB` does not preserve the inner cache’s `AccountState`, so journal refresh can observe different storage semantics than the final `fork.db.commit` (for example, a locally `NotExisting` account becoming touched). Please stage against a clone of `fork.db` and publish that clone only after both refreshes succeed.

stevencartavia: +1 from my pass

haythemsellami:
Updated in `64a0d94a8`: staging now clones `fork.db`, applies the changeset, refreshes both cloned journals against that exact cache, and publishes all three only after both refreshes succeed. I also added a regression for a locally `NotExisting` account becoming `Touched` with an uncached slot, which catches the account-state mismatch in the previous outer-cache approach.

mablr (approved):

grandizzy (approved):
ty, lgtm!

PR #15901, opened 2026-07-27
fix(anvil): report per-transaction gas

### LINKED ISSUE #15893: bug(anvil): mined traces use cumulative transaction gas
Component: Anvil

Mined receipts correctly store cumulative gas for consensus accounting, but Anvil also uses `receipt.cumulative_gas_used()` for transaction-local output and debug traces. For any transaction after index zero, the reported gas therefore includes gas consumed by preceding transactions in the same block.

Mining output, the default `debug_traceTransaction` frame, and the call tracer should use the transaction's own `TransactionInfo::gas_used`. A regression should mine at least two transactions in one block and compare the second transaction's default and call-tracer gas with its receipt `gasUsed`.

haythemsellami:
Anvil receipts correctly contain cumulative gas, but transaction-local mining output and debug trace frames were reading that cumulative value. Transactions after index zero therefore reported gas consumed by earlier transactions in the same block. This change uses the transaction's own `TransactionInfo::gas_used` for those outputs while leaving receipt accounting unchanged.

The regression mines two transactions together and compares the second transaction's default and call-tracer gas with its receipt `gasUsed`. Closes #15893.

AI assistance was used for implementation and code review.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -3282,7 +3282,7 @@ where
                 if let Some(contract) = &info.contract_address {
                     node_info!("    Contract created: {contract}");
                 }
-                node_info!("    Gas used: {}", receipt.cumulative_gas_used());
+                node_info!("    Gas used: {}", info.gas_used);
                 if !info.exit.is_ok() {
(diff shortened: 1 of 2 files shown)


mablr (approved):
LGTM

grandizzy (approved):
ty!

Issue #15897, opened 2026-07-27
bug(fork-db): exact forks mix canonical ancestry


haythemsellami:
Component: Forge / fork database

An exact transaction-hash or block-hash fork can pin account and storage reads to a block hash, but the fork database still resolves `BLOCKHASH(number)` ancestry through canonical number-based RPC lookups and number-keyed cache entries. After a same-height reorg, execution can therefore combine orphan block A's state with replacement chain B's ancestry.

A complete fix needs source-aware fork identity throughout `foundry-fork-db`: cache provenance must include the endpoint and exact block lineage, and historical block-hash lookups must follow the pinned block's ancestors rather than the endpoint's current canonical block at each height. This should be designed as an architectural change rather than folded into a narrow preflight fix.

This was found while reviewing #15343. It is pre-existing network-neutral fork infrastructure behavior and should be tracked independently.

PR #15902, opened 2026-07-27
fix(forge): pin fork preflight reads

### LINKED ISSUE #15896: bug(forge): CREATE2 preflight ignores fork hash
Component: Forge

Fork-backed CREATE2 availability and script nonce preflight currently discard the block hash returned with the fork environment and query by block number. A same-height reorg can therefore make linking or nonce selection observe replacement block B after the environment was resolved from block A. Script multi-RPC execution can also accidentally carry a materialized `latest` height to another endpoint, and cached script backends are keyed only by URL rather than the resolved fork context.

Preflight reads should carry an endpoint-bound `{number, hash}` identity, use the hash for code and nonce queries, preserve whether the user requested `latest` or an explicit height when changing endpoints, and invalidate a cached script backend when its resolved context changes. Existing public number-based APIs can remain for compatibility, while internal consumers use the exact path.

This issue is limited to exact preflight reads and script context lifecycle. Making the complete fork database and `BLOCKHASH` ancestry exact is tracked separately in #15897.

This was found while reviewing #15343, but it is pre-existing network-neutral Foundry behavior and should be fixed independently.

haythemsellami:
Fork preparation previously materialized only a block number, allowing preflight reads and later block-environment reconstruction to observe different blocks after a same-height reorg. This introduces a source-, header-, and selector-bound resolved fork context and reuses it for CREATE2 and nonce reads plus script and normal Forge block-environment setup. Script backends are keyed by that complete context, so endpoint, header, selector, or hash changes cannot reuse a stale backend. The fork database remains number-pinned; exact state and ancestry are tracked separately in #15897.

Closes #15896.


crates/script/src/lib.rs
```diff
@@ -872,6 +894,66 @@ impl<FEN: FoundryEvmNetwork> ScriptConfig<FEN> {
...
+    /// Changes the script fork endpoint and selector atomically.
+    pub fn set_fork_endpoint(&mut self, fork_url: String, fork_block_number: Option<u64>) {
+        self.evm_opts.fork_url = Some(fork_url);
+        self.evm_opts.fork_block_number = fork_block_number;
+        self.requested_fork_block_number = fork_block_number;
+        self.fork_block = None;
+    }
+
+    fn set_fork_url(&mut self, fork_url: String) {
+        self.set_fork_endpoint(fork_url, self.requested_fork_block_number);
```
mablr: For an unchanged RPC URL, this discards the pinned hash and re-resolves the selector during simulation. A same-height reorg can therefore switch hashes between execution and simulation. Please preserve the exact fork identity when the URL is unchanged and add a same-endpoint reorg test.

crates/script/src/lib.rs
```diff
@@ -872,6 +894,66 @@ impl<FEN: FoundryEvmNetwork> ScriptConfig<FEN> {
...
+            return Ok(Some(fork_block));
+        }
+
+        if let Some(context) = &self.fork_block
+            && self.evm_opts.fork_url.as_deref() == Some(context.url.as_str())
+        {
+            // A changed selector on the same endpoint is an explicit new fork request.
+            self.requested_fork_block_number = self.evm_opts.fork_block_number;
+        } else if let Some(context) = &self.fork_block
+            && self.evm_opts.fork_block_number == Some(context.block.number)
```
mablr: This infers whether `fork_block_number` is user intent or a materialized height, but that is ambiguous when a new endpoint explicitly requests the old resolved height. Can we keep the selector immutable and store the resolved `{number, hash}` separately instead?

crates/script/src/lib.rs
```diff
@@ -820,14 +820,23 @@ pub struct ScriptConfig<FEN: FoundryEvmNetwork> {
     pub evm_opts: EvmOpts,
     pub sender_nonce: u64,
     sender_nonce_override: Option<u64>,
+    fork_block: Option<ScriptForkBlock>,
+    requested_fork_block_number: Option<u64>,
     /// Maps a rpc url to a backend
     pub backends: HashMap<String, Backend<FEN>>,
+    backend_blocks: HashMap<String, ForkBlock>,
```
mablr: Could this be one map from endpoint/resolved identity to `(ForkBlock, Backend)`? Parallel maps make the block/backend invariant manual and allow the entries to drift.

crates/evm/core/src/opts.rs
```diff
@@ -88,6 +88,46 @@ pub struct EvmOpts {
...
+        Self { number, hash, source: None }
+    }
+
+    /// Creates an exact block identity resolved from a fork endpoint.
+    pub fn from_endpoint(number: BlockNumber, hash: B256, fork_url: &str) -> Self {
+        Self::new(number, hash).resolved_from(fork_url)
+    }
+
+    fn resolved_from(mut self, fork_url: &str) -> Self {
+        self.source = Some(keccak256(fork_url.as_bytes()));
```
mablr: `source` fingerprints only the raw URL, while provider identity also depends on headers, and the public constructor can manufacture arbitrary resolved values. Would a `ResolvedFork` context be cleaner than an optionally bound `ForkBlock`?

crates/script/src/lib.rs
```diff
@@ -1021,6 +1113,207 @@ mod tests {
...
+        assert_eq!(runner.executor.get_balance(recipient).unwrap(), U256::from(1));
+
+        config.set_fork_endpoint(url_a.clone(), Some(1));
+        let runner = config.get_runner().await.unwrap();
+        assert_eq!(config.backend_blocks[&url_a].number, 1);
+        assert_eq!(runner.executor.get_balance(recipient).unwrap(), U256::ZERO);
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn script_link_preflight_and_runner_use_pinned_fork_hash() {
```
mablr: This name overstates the guarantee: the environment is revalidated by hash, but `get_fork` still creates a number-pinned database. Could we rename this and mention #15897 so it does not imply exact runner state?

haythemsellami:
Reworked the fix around an immutable, source/header-bound `ResolvedFork`: same-endpoint script preparation retains the original hash, while source, header, or selector changes re-resolve and invalidate backend reuse. Forge and mutation setup now reuse that context, mutation caches include number/hash, and the number-pinned DB limitation remains isolated to #15897.

crates/evm/core/src/opts.rs
```diff
@@ -148,6 +163,45 @@ impl EvmOpts {
...
+    ) -> eyre::Result<bool> {
+        let Some(_) = &self.fork_url else {
+            eyre::ensure!(fork.is_none(), "resolved fork provided without a configured fork");
+            return Ok(self.create2_deployer == DEFAULT_CREATE2_DEPLOYER);
+        };
+        let fork = fork.ok_or_else(|| eyre::eyre!("fork must be resolved"))?;
+        let provider = self.provider_for_resolved_fork::<AnyNetwork>(fork)?;
+        Ok(!provider
+            .get_code_at(self.create2_deployer)
+            .block_id(BlockId::hash(fork.hash()))
```
mablr: **Blocker:** `BlockId::hash` leaves `require_canonical` unset, so Alloy serializes this as a bare hash string rather than the EIP-1898 object form. Anvil accepts that extension, but strict RPCs may reject it or interpret it as a quantity. Can we centralize an exact selector on `ResolvedFork` (for example `BlockId::from((fork.hash(), Some(false)))`) and use it for both code and nonce reads? Please also add a serialization or mocked-request test asserting `{ blockHash, requireCanonical: false }`, since the Anvil reorg test cannot catch the wire shape.

crates/script/src/broadcast.rs
```diff
@@ -72,16 +74,14 @@ where
...
+    evm_opts: &EvmOpts,
+    fork: &ResolvedFork,
 ) -> eyre::Result<u64> {
-    let provider = try_get_http_provider(provider_url)
-        .wrap_err_with(|| format!("bad fork_url provider: {provider_url}"))?;
-
-    let block_id = block_number.map_or(BlockId::latest(), BlockId::number);
-    Ok(provider.get_transaction_count(caller).block_id(block_id).await?)
+    let provider = evm_opts.provider_for_resolved_fork::<AnyNetwork>(fork)?;
+    Ok(provider.get_transaction_count(caller).block_id(BlockId::hash(fork.hash())).await?)
```
mablr: **Blocker:** this nonce preflight has the same bare-hash serialization issue as the CREATE2 lookup. Please use the centralized EIP-1898 selector here as well so `eth_getTransactionCount` is unambiguously hash-bound across RPC implementations.

crates/forge/src/mutation/orchestrator.rs
```diff
@@ -131,11 +144,12 @@ pub struct MutationRunResult {
 pub async fn run_mutation_testing(
     config: Arc<Config>,
     output: &ProjectCompileOutput<MultiCompiler>,
-    mut evm_opts: EvmOpts,
+    evm_opts: EvmOpts,
     mutation_config: MutationRunConfig,
 ) -> Result<MutationRunResult> {
-    let fork_block = evm_opts.pin_fork_block().await?;
-    let create2_deployer_available = evm_opts.can_use_create2_deployer(fork_block).await?;
+    let resolved_fork = evm_opts.resolve_fork().await?;
```
mablr: **Blocker:** this resolves `latest` only after the unmutated baseline has completed, while the baseline independently resolves its context in `build_and_run_tests`. If the chain advances or reorgs between phases, the baseline can run at A and every mutant at B, invalidating kill/alive comparisons. Please resolve once before the baseline and carry that `ResolvedFork` through baseline setup, cache identity, and workers. If mutation pinning is intentionally outside #15896, I would instead remove the mutation expansion from this PR and handle it separately.

crates/evm/core/src/opts.rs
```diff
@@ -128,6 +129,20 @@ impl EvmOpts {
         Ok(self.fork_block_number)
     }
 
+    /// Resolves the configured fork selector without changing it.
+    pub async fn resolve_fork(&self) -> eyre::Result<Option<ResolvedFork>> {
+        let Some(fork_url) = &self.fork_url else { return Ok(None) };
+        let provider = self.fork_provider_with_url::<AnyNetwork>(fork_url)?;
+        let (_, block) =
+            self.fork_evm_env_resolved::<SpecId, revm::context::BlockEnv, _, _>(&provider).await?;
```
mablr: Non-blocking simplification: `resolve_fork` currently constructs and discards a generic EVM environment just to obtain the selected block identity. A small shared helper that fetches the selected header / `BlockNumHash`, reused by this method and `fork_evm_env_resolved`, would make the ownership clearer and avoid unnecessary environment construction.

crates/script/src/lib.rs
```diff
@@ -1024,6 +1072,287 @@ mod tests {
...
+                    .to(recipient)
+                    .value(U256::from(1))
+                    .into(),
+            )
+            .await
+            .unwrap()
+            .get_receipt()
+            .await
+            .unwrap();
+        config.set_fork_url(url_a.clone());
```
mablr, line 1256: Test improvement: this proves backend replacement when endpoint A advances from block 1 to block 2, but the reported bug is specifically a same-height replacement. Could this switch away, `anvil_reorg` A at the same height, then switch back and assert the changed hash causes a fresh context/backend? That would directly cover hash-based invalidation rather than number advancement.

crates/evm/core/src/opts.rs
haythemsellami: Fixed in 94132ec19: ResolvedFork now owns a centralized exact_block_id() selector that serializes as {blockHash, requireCanonical:false}, with an exact JSON regression test. CREATE2 code reads now use that selector.

crates/script/src/lib.rs
haythemsellami: Updated in 08187aeaf: the script backend test now switches from A to B, performs a same-height anvil_reorg on A, switches back, and asserts the changed hash, cache eviction, and rebuilt A backend and environment.

crates/evm/core/src/opts.rs
haythemsellami: Addressed in 08187aeaf: selected-block fetching and BlockNumHash derivation now live in a shared helper used by resolve_fork and fork_evm_env_resolved. The environment path keeps chain-ID and block requests concurrent and preserves the existing diagnostics.

crates/script/src/broadcast.rs
haythemsellami: Fixed in 94132ec19: the nonce preflight now uses the same ResolvedFork EIP-1898 selector, so eth_getTransactionCount is unambiguously hash-bound.

crates/forge/src/mutation/orchestrator.rs
haythemsellami: Fixed in 60579cb93: build_and_run_tests now returns the exact context used by the baseline, and the CLI passes that same owned context through CREATE2 preflight, cache fingerprinting, and every mutation worker. An end-to-end regression advances Anvil during the baseline and verifies mutants remain on the original block.

crates/forge/src/mutation/runner.rs
```diff
@@ -726,10 +856,40 @@ fn compile_and_test_inner<FEN: FoundryEvmNetwork>(
...
+        }
+        MutationForkContext::Resolved(resolved_fork) => {
+            let (evm_env, tx_env) = context
+                .evm_opts
+                .env_with_resolved_fork::<SpecFor<FEN>, BlockEnvFor<FEN>, TxEnvFor<FEN>>(
+                    resolved_fork.as_ref(),
+                )
+                .await
+                .map_err(|error| eyre::Report::new(MutationInfrastructureError(error)))?;
+            Ok((evm_env, tx_env, resolved_fork.as_ref().map(ResolvedFork::number)))
```
mablr: **Blocker:** this preserves the hash for environment reconstruction, but then reduces the fork DB identity back to the block number. After a same-height reorg, `evm_env` can describe A while `get_fork` serves account/storage/code from replacement B, so baseline and mutants are not necessarily comparable. This is the #15897 boundary; could we split the mutation expansion from this PR, or narrow it so it does not imply exact mutant state?

crates/script/src/lib.rs
```diff
@@ -903,21 +944,24 @@ impl<FEN: FoundryEvmNetwork> ScriptConfig<FEN> {
...
-                    backend
-                }
+        let resolved = self.ensure_resolved_fork().await?;
+        let (evm_env, mut tx_env) =
+            self.evm_opts.env_with_resolved_fork::<_, _, TxEnvFor<FEN>>(resolved.as_ref()).await?;
+        let fork_block_number = resolved.as_ref().map(ResolvedFork::number);
+
+        let db = if let Some(fork_url) = self.evm_opts.fork_url.clone() {
+            resolved.context("fork must be resolved")?;
+            if let Some(backend) = self.backends.get(&fork_url) {
```
mablr: The issue calls out URL-only backend reuse, but the cache is still keyed only by URL and correctness depends on every mutation path clearing the whole map first. Could we encode the invariant in the cache itself, e.g. a private `Option<(ResolvedFork, Backend)>` or a full-context key, so a stale backend cannot be reused independently of its resolved context?

crates/forge/src/mutation/runner.rs
```diff
@@ -748,6 +908,100 @@ mod tests {
...
+            .await
+            .unwrap();
+        let replacement = evm_opts.resolve_fork().await.unwrap().unwrap();
+        assert_eq!(resolved.number(), replacement.number());
+        assert_ne!(resolved.hash(), replacement.hash());
+        let replacement_block =
+            handle.http_provider().get_block_by_number(1.into()).await.unwrap().unwrap();
+        assert_ne!(replacement_block.header.mix_hash, Some(prevrandao));
+
+        match mutation_worker_env::<EthEvmNetwork>(&context).await {
```
mablr: This regression can pass when reconstruction fails for any reason, and the success branch checks only the hash-derived block environment—not state read through the number-pinned backend. Could this be deterministic and assert observable backend state, or be removed with the mutation scope? As written it does not prove baseline/mutant state consistency across a same-height reorg.

mablr (changes requested):
The `ResolvedFork`/EIP-1898 direction looks sound and the script lifecycle changes appear to address #15896. My remaining concern is scope and guarantee: mutation now carries an exact header/preflight identity, but its fork DB is still number-pinned, so a same-height reorg can mix block A’s environment with block B’s state. I’d split the mutation expansion from this PR (or narrow it to preflight/cache identity without claiming exact mutant state), keep the core/script fix, and make backend reuse structurally context-bound rather than URL-bound.

crates/forge/src/mutation/runner.rs
haythemsellami: Addressed by removing the mutation-specific resolved-context propagation from this PR. Mutation execution and cache behavior are back to upstream; exact mutant state remains deferred to #15897.

crates/script/src/lib.rs
haythemsellami: Addressed: the script backend cache is now private and keyed by the complete ResolvedFork (source, ordered headers, selector, number, and hash), so reuse no longer depends on clearing a URL-keyed map.

crates/forge/src/mutation/runner.rs
haythemsellami: Addressed by removing this regression together with the mutation scope. The remaining same-height tests cover exact preflight/environment reconstruction and structural script-cache identity without claiming hash-bound backend state.

haythemsellami:
> The ResolvedFork/EIP-1898 direction looks sound and the script lifecycle changes appear to address https://github.com/foundry-rs/foundry/issues/15896. My remaining concern is scope and guarantee: mutation now carries an exact header/preflight identity, but its fork DB is still number-pinned, so a same-height reorg can mix block A’s environment with block B’s state. I’d split the mutation expansion from this PR (or narrow it to preflight/cache identity without claiming exact mutant state), keep the core/script fix, and make backend reuse structurally context-bound rather than URL-bound.

Makes sense, I did cut the mutation expansion from this PR and applied the rest of the fixes.

mablr (dismissed):
LGTM, thank you @haythemsellami !

mablr (dismissed):

mablr (dismissed):

crates/forge/src/mutation/runner.rs
```diff
@@ -711,7 +727,7 @@ fn compile_and_test_inner<FEN: FoundryEvmNetwork>(
             .set_debug(false)
             .initial_balance(evm_opts.initial_balance)
             .sender(evm_opts.sender)
-            .with_fork(evm_opts.get_fork(config, evm_env.cfg_env.chain_id, fork_block))
+            .with_fork(evm_opts.get_fork(config, evm_env.cfg_env.chain_id, fork_block_number))
```
figtracer: this reintroduces the same-height reorg gap from the earlier mutation thread: the env is rebuilt from block a by hash, but the fork db is opened at number n, so it can serve replacement b's state. please keep mutation out of scope until the db accepts the resolved hash, or pass the exact selector through.

mablr: mb, rolled back https://github.com/foundry-rs/foundry/pull/15902/commits/c7d20b54973326ff9d0d22c9a44f3b657ca8eeb1

mablr (approved):

stevencartavia (approved):

PR #15907, opened 2026-07-27
fix(symbolic): debit pranked call value

### LINKED ISSUE #15903: bug(symex): debit pranked caller for value transfers
### Component

Forge symbolic execution

### Version

Current `main` at `c874e16b7cb77b0f6c77b55373ef974a5361c039`.

### Command

`forge test --symbolic`

### Operating system

Linux

### Description

originally reported by @clement-ux 

A value-bearing external call made after `vm.prank(alice)` does not debit the pranked caller in the symbolic model. This produces a spurious counterexample that concrete replay rejects.

Minimal reproducer:

```solidity
function check_deposit_withdraw_roundtrip(uint96 amount) public {
    vm.deal(alice, amount);
    uint256 preEthBalance = alice.balance;

    vm.prank(alice);
    weth.deposit{value: amount}();

    vm.prank(alice);
    weth.withdraw(amount);

    assertEq(alice.balance, preEthBalance);
}
```

Result:

```text
FAIL: symbolic counterexample did not replay
```

Expected: symbolic execution should model the same prank/value-transfer semantics as concrete Forge execution. The deposit should debit `alice`, and the withdrawal should restore Alice's original ETH balance, so the property passes.

Actual: symbolic execution appears to debit the executing test contract rather than the pranked caller. Alice remains funded after the deposit, then receives `amount` again during withdrawal. The solver reports an assertion failure, but concrete replay correctly debits and credits Alice, so the candidate cannot replay.

On current `main`, `prepare_value_transfer` reads the balance of `state.address`, while prank handling is applied later when constructing the callee frame:

- https://github.com/foundry-rs/foundry/blob/c874e16b7cb77b0f6c77b55373ef974a5361c039/crates/evm/symbolic/src/executor/calls.rs#L1248
- https://github.com/foundry-rs/foundry/blob/c874e16b7cb77b0f6c77b55373ef974a5361c039/crates/evm/symbolic/src/executor/calls.rs#L938

The individual capabilities are covered—symbolic value transfers and `vm.prank`—but their composition does not appear to preserve concrete Forge semantics.

mablr:
## Description

Symbolic execution previously applied `vm.prank` when constructing the callee frame, after checking and transferring call value from the executing contract. Resolve the effective caller before those operations and use it consistently across deployed contracts, empty accounts, precompiles, and symbolic call targets. This aligns symbolic value transfers with concrete Forge execution and prevents replay-rejected counterexamples.

Adds regression coverage for a pranked WETH deposit and withdrawal round trip.

Fixes #15903.

### DIFF
--- a/crates/evm/symbolic/src/executor/calls.rs
+++ b/crates/evm/symbolic/src/executor/calls.rs
@@ -857,11 +857,16 @@ impl SymbolicExecutor {
             }
         }
 
+        if matches!(kind, CallKind::DelegateCall) && state.prank.has_active() {
+            return Err(SymbolicError::Unsupported("symbolic prank delegatecall"));
+        }
+        let (call_caller, call_caller_word, pranked_origin) = state.prank_for_next_call();
         if matches!(kind, CallKind::Call)
             && !self.prepare_value_transfer(
                 executor,
                 state,
                 worklist,
+                call_caller,
                 value.clone(),
                 out_offset.clone(),
                 &out_size,
@@ -882,6 +887,7 @@ impl SymbolicExecutor {
                     worklist,
                     kind,
                     to,
+                    call_caller,
                     value,
                     out_offset,
                     &out_size,
@@ -899,7 +905,7 @@ impl SymbolicExecutor {
                 Some(return_data) => {
                     state.return_data = return_data;
                     if matches!(kind, CallKind::Call) {
-                        state.world.transfer(&mut self.cx, executor, state.address, to, value);
+                        state.world.transfer(&mut self.cx, executor, call_caller, to, value);
                     }
                     state.copy_call_output_offset(&mut self.cx, out_offset, &out_size)?;
                     state.stack.push(SymExpr::one(&mut self.cx))?;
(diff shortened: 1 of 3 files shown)


crates/evm/symbolic/src/executor/calls.rs
```diff
@@ -916,7 +922,7 @@ impl SymbolicExecutor {
         let child_code = state.world.extcode(&mut self.cx, executor, code_address)?;
         if child_code.is_empty() {
             if matches!(kind, CallKind::Call) {
-                state.world.transfer(&mut self.cx, executor, state.address, to, value);
+                state.world.transfer(&mut self.cx, executor, call_caller, to, value);
```
grandizzy, line 925: This exposes incorrect accounting when the pranked caller is also the target:

```solidity
vm.deal(ALICE, amount);
vm.prank(ALICE);
(bool ok,) = ALICE.call{value: amount}("");
assertEq(ALICE.balance, amount);
```

Here this becomes `transfer(ALICE, ALICE, amount)`. `SymbolicWorld::transfer` snapshots both balances before writing them, so its credit overwrites the debit and leaves `ALICE` with `balance + amount`. On this PR, the example produces `symbolic counterexample did not replay` for `amount = 1`.

Could we make self-transfers a no-op in `SymbolicWorld::transfer` and add the example as regression coverage?

```diff
-if value.as_const().is_some_and(|value| value.is_zero()) {
+if from == to || value.as_const().is_some_and(|value| value.is_zero()) {
     return;
 }
```

The helper issue predates this PR, but using the effective prank caller here newly exposes it in the behavior this change is fixing.

grandizzy (changes requested):
The overall approach looks correct: resolving the effective caller before affordability checks aligns the symbolic model with concrete prank semantics, and the added deposit/withdraw regression passes.

One remaining correctness issue exists in the newly enabled path: when the pranked caller is also the call target, symbolic value transfer mints ETH and produces another replay-rejected counterexample. An inline comment includes a minimal reproduction and suggested fix. With that addressed, this looks good to merge.

grandizzy (approved):
lgtm

mattsse (approved):
lgtm

PR #15908, opened 2026-07-27
fix(anvil): sync auto-detected fee spec

mablr:
## Motivation

#15898 gates raw base-fee calculation on the fee manager’s active spec, but historical fork auto-detection happens after the fee manager is constructed. Forks without an explicit hardfork could therefore retain the default post-London spec and continue deriving EIP-1559 base fees. Fork-derived values could also become persistent configuration overrides, preventing later resets from refreshing them from the target fork.

Built on #15919, this keeps configured overrides separate from resolved fork state. Fork setup synchronizes the EVM and fee-manager hardfork with the target block and refreshes inferred base fee, gas price, and gas limit on every reset, while explicit overrides remain pinned. Clearing a fork restores the configured non-fork fallback, and fork reporting uses the active resolved values.

### DIFF
--- a/crates/anvil/src/config.rs
+++ b/crates/anvil/src/config.rs
@@ -120,7 +120,7 @@ pub struct NodeConfig {
     pub disable_min_priority_fee: bool,
     /// Default blob excess gas and price
     pub blob_excess_gas_and_price: Option<BlobExcessGasAndPrice>,
-    /// The hardfork to use
+    /// The hardfork to force, or `None` to infer it from chain activation data.
     pub hardfork: Option<FoundryHardfork>,
     /// Signer accounts that will be initialised with `genesis_balance` in the genesis block
     pub genesis_accounts: Vec<PrivateKeySigner>,
@@ -310,7 +310,14 @@ Chain ID
             );
         }
 
-        if (SpecId::from(self.get_hardfork()) as u8) < (SpecId::LONDON as u8) {
+        let active_hardfork = fork
+            .and_then(|fork| fork.config.read().hardfork)
+            .unwrap_or_else(|| self.get_hardfork());
+        let active_base_fee =
+            fork.and_then(ClientFork::base_fee).unwrap_or_else(|| self.get_base_fee() as u128);
+        let active_gas_price =
+            fork.map(ClientFork::gas_price).unwrap_or_else(|| self.get_gas_price());
+        if (SpecId::from(active_hardfork) as u8) < (SpecId::LONDON as u8) {
             let _ = write!(
                 s,
                 r#"
@@ -319,7 +326,7 @@ Gas Price
 
 {}
 "#,
-                self.get_gas_price().green()
+                active_gas_price.green()
             );
         } else {
             let _ = write!(
@@ -330,7 +337,7 @@ Base Fee
 
 {}
 "#,
-                self.get_base_fee().green()
+                active_base_fee.green()
             );
         }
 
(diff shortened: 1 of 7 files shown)


grandizzy (dismissed):
lgtm

mattsse (changes requested):
The fee spec remains stale across `anvil_reset`. Counterexample: start an auto-detected Berlin fork, reset it to a London block, then simulate two blocks with a `0x3e8` base-fee override on the first. The second block returns `0x0` instead of deriving `0x36b`, because initial detection sets `self.hardfork = Some(...)` and later fork setup skips re-detection.

Please add this Berlin→London reset regression and distinguish explicitly configured hardforks from auto-detected ones, or otherwise refresh both the EVM and fee-manager specs on reset.

mablr (commented):
@mattsse

Addressed your finding in https://github.com/foundry-rs/foundry/pull/15908/commits/455fca556640bee6552c4289c8e6436e1f974b88.

mattsse (dismissed):
lgtm

crates/anvil/src/config.rs
```diff
@@ -122,6 +122,8 @@ pub struct NodeConfig {
     pub blob_excess_gas_and_price: Option<BlobExcessGasAndPrice>,
     /// The hardfork to use
     pub hardfork: Option<FoundryHardfork>,
+    /// The hardfork auto-detected from the active fork.
+    detected_hardfork: Option<FoundryHardfork>,
```
figtracer: this is the first private field on `NodeConfig`, so downstream `NodeConfig { ..Default::default() }` literals no longer compile with e0451. could this stay public, perhaps `#[doc(hidden)]`, to preserve the existing api?

crates/anvil/src/config.rs
```diff
@@ -1439,19 +1443,24 @@ latest block number: {latest_block}"
...
+        // Resolve the fork block's hardfork without materializing it into `self.hardfork`.
+        // That field represents the user's explicit override; keeping inference on the fork
+        // config lets a later reset re-resolve timestamp-based activations.
+        let fork_hardfork = self.hardfork.or_else(|| {
+            FoundryHardfork::from_chain_and_timestamp(chain_id, block.header.timestamp())
+        });
+        if let Some(hardfork) = fork_hardfork {
+            let spec_id = SpecId::from(hardfork);
+            evm_env.cfg_env.set_spec_and_mainnet_gas_params(spec_id);
+            fees.set_spec_id(spec_id);
```
stevencartavia: `anvil_reset` with JSON-RPC `params: []` clears the fork without restoring the node’s resolved non-fork fallback spec. Reproduction: after resetting an auto-detected Berlin fork, an EIP-1559 transaction was still rejected as pre-London, while a fresh default non-fork node accepted the same transaction. Could `reset_to_in_mem` reapply the non-fork fallback spec to `EvmEnv` (including gas parameters) and `FeeManager`? The fork-to-fork path is refreshed, but the fork-to-in-memory transition remains stale.

crates/anvil/src/config.rs
```diff
@@ -1439,19 +1443,24 @@ latest block number: {latest_block}"
...
-        if self.hardfork.is_none()
-            && let Some(hardfork) =
-                FoundryHardfork::from_chain_and_timestamp(chain_id, block.header.timestamp())
-        {
-            evm_env.cfg_env.spec = SpecId::from(hardfork);
-            self.hardfork = Some(hardfork);
+        // Resolve the fork block's hardfork without materializing it into `self.hardfork`.
+        // That field represents the user's explicit override; keeping inference on the fork
+        // config lets a later reset re-resolve timestamp-based activations.
+        let fork_hardfork = self.hardfork.or_else(|| {
```
stevencartavia, line 1463: Auto-detected Berlin forks now leave `NodeConfig.hardfork` unset, but the startup banner still chooses between “Gas Price” and “Base Fee” using `self.get_hardfork()`. Reproduction: the CLI printed a `Base Fee` section with `1000000000`, while the matching chain-ID-1 fork block had no `baseFeePerGas`. Could the banner use the passed `ClientFork`’s resolved hardfork instead of the default fallback?

PR #15910, opened 2026-07-27
fix(debugger): search split call segments

figtracer:
Opcode search currently inspects only the active flattened trace segment even though its scope is presented as the current call. Parent calls are split around nested calls, so opcodes before or after a child can be invisible to `/`, `n`, and `N`.

Search now follows the stable trace node ID and absolute step offsets across every segment of the logical call. Forward and backward searches retain their wraparound behavior while skipping opcodes that belong to child calls.

Progresses #927.


mattsse (approved):
lgtm

grandizzy (approved):
lgtm

PR #15911, opened 2026-07-27
feat(fmt): wrap method chains consistently

### LINKED ISSUE #12344: feat(fmt): option for supporting method chaining
### Component

Forge

Current behavior:

```solidity
stdstore.target(address(eigenDAServiceManager)).sig("batchIdToBatchMetadataHash(uint32)")
    .with_key(defaultBatchId).checked_write(CertV1Lib.hashBatchMetadata(batchMetadata));
```

Preferred optional behavior: 

```solidity
stdstore.target(address(eigenDAServiceManager))
    .sig("batchIdToBatchMetadataHash(uint32)")
    .with_key(defaultBatchId)
    .checked_write(CertV1Lib.hashBatchMetadata(batchMetadata));
```

### Additional context

_No response_

figtracer:
Closes #12344.


crates/fmt/src/state/sol.rs
```diff
@@ -1405,15 +1405,17 @@ impl<'ast> State<'_, 'ast> {
...
                                 ident.span.lo(),
                                 CommentConfig::skip_ws().mixed_no_break().mixed_prev_space(),
                             );
+                            true
                         } else {
-                            s.print_trailing_comment(member_expr.span.hi(), Some(ident.span.lo()));
-                        }
-                        if has_mixed_comment || s.member_suffix_emits_break(expr, member_expr) {
+                            !s.print_trailing_comment(member_expr.span.hi(), Some(ident.span.lo()))
+                                && s.member_suffix_emits_break(expr, member_expr)
```
mattsse, line 1421: A standalone comment can break the consistent box after this breakpoint, leaving `.` on its own line (`factory()` / `.` / `// preserve` / `first()`). Skip the breakpoint while an unconsumed comment remains between the receiver and member; a snapshot for that case would guard the interaction.

```suggestion
                            !s.print_trailing_comment(member_expr.span.hi(), Some(ident.span.lo()))
                                && s
                                    .peek_comment_between(member_expr.span.hi(), ident.span.lo())
                                    .is_none()
                                && s.member_suffix_emits_break(expr, member_expr)
```

mattsse (commented):
One formatting edge case.

crates/fmt/src/state/sol.rs
mablr: +1

mablr (dismissed):

crates/fmt/src/state/sol.rs
```diff
@@ -1797,7 +1801,7 @@ impl<'ast> State<'_, 'ast> {
             }
 
             if chain_has_indent {
-                self.s.ibox(self.ind);
+                self.s.cbox(self.ind);
```
stevencartavia, line 1807: When the chain wraps, this also wraps an otherwise-fitting argument list after call options. At `line_length = 60`, `target{value: amount}(shortArg).firstLongMethod().secondLongMethod()` puts `shortArg` two indentation levels below `target`. Could the argument-list boundary breaks be isolated from the chain’s consistent breaking?

mattsse, line 1807: Counterexample at `line_length = 60`:

```solidity
target{value: amount}(shortArg).firstLongMethod().secondLongMethod();
```

This currently expands the fitting argument to `target{value: amount}(\n        shortArg\n    )` because the delimiter breaks from `.with_delimiters(!s.call_with_opts_and_args)` are exposed to this outer `cbox`.

A concrete fix is to keep those delimiters inside the call list's own box when the current context is an indented chain:

```rust
.with_delimiters(
    !s.call_with_opts_and_args
        || s.call_stack
            .last()
            .is_some_and(|call| call.is_chained() && call.has_indent),
)
```

That keeps `(shortArg)` inline while this box consistently wraps only the chain suffixes. A `line_length = 60` snapshot for the counterexample would cover the boundary.

mattsse (commented):
One more layout interaction to isolate.

crates/fmt/src/state/sol.rs
```diff
@@ -1362,9 +1362,12 @@ impl<'ast> State<'_, 'ast> {
                             };
                         s.print_call_args(
                             call_args,
-                            list_format
-                                .without_ind(s.return_bin_expr)
-                                .with_delimiters(!s.call_with_opts_and_args),
+                            list_format.without_ind(s.return_bin_expr).with_delimiters(
```
mablr, line 1365: Could we add a fixture with call options and overflowing positional arguments followed by a chain? The current `shortArg` case covers fitting arguments, but an overflowing case would verify that this condition still allows the argument list to wrap normally.

mablr (dismissed):
LGTM, I left an optional suggestion.

mattsse (dismissed):

mablr (approved):

.github/scripts/validate_changelog.py
mablr, line 1: seems unrelated to pr

mablr (changes requested):

mablr (approved):

stevencartavia (approved):

PR #15912, opened 2026-07-27
perf(cast): speed up transaction building

figtracer:
## Summary

Fetch nonce and fee data concurrently when Cast fills ordinary Ethereum transactions. Gas estimation still runs after both fields are populated, while EIP-7702 authorization and Tempo transaction paths retain their existing ordering. This removes one network round trip from `cast send` and `cast mktx`.

A barrier-based mocked transport test requires exactly one nonce request and one fee request to be in flight together. Validated with all 226 Cast library tests, `cargo +nightly fmt --all -- --check`, `git diff --check`, `cargo +1.95.0 check -p cast`, and Clippy with warnings denied. Implemented with Codex.

## Benchmarks

Measured exact `master` (`1a9cf57c8`) against this PR using profiling builds and Anvil behind a deterministic HTTP proxy that adds 200 ms to every JSON-RPC request. Each result is 20 `hyperfine` runs after 3 warmups.

| Command | `master` | This PR | Speedup |
| --- | ---: | ---: | ---: |
| `cast mktx --chain 31337 --legacy` | `632.3ms ± 1.6ms` | `428.3ms ± 1.8ms` | `1.48x` |
| `cast mktx --chain 31337` | `631.1ms ± 2.6ms` | `427.9ms ± 2.7ms` | `1.47x` |
| `cast mktx` with chain discovery | `834.6ms ± 2.6ms` | `632.4ms ± 3.1ms` | `1.32x` |
| `cast send --async --chain 31337 --legacy` | `837.3ms ± 2.1ms` | `634.0ms ± 2.1ms` | `1.32x` |

Every case saves approximately 203–204 ms: one full delayed RPC round trip.

### DIFF
--- a/crates/cast/src/tx.rs
+++ b/crates/cast/src/tx.rs
@@ -595,15 +595,46 @@ where
         // resolve
         // Read-only calls do not need a nonce unless it is required to sign an authorization.
         // Avoid an otherwise unused `eth_getTransactionCount` request for raw transactions.
-        if fill || !self.auth.is_empty() {
-            let tx_nonce = self.resolve_nonce(sender.address(), fill).await?;
+        let resolve_in_parallel =
+            fill && self.auth.is_empty() && tempo_wallet.is_none() && !self.chain.is_tempo();
+        let tx_nonce = if resolve_in_parallel {
+            let nonce = self.tx.nonce();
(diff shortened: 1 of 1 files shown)


mablr (approved):

mattsse (approved):
lgtm

PR #15914, opened 2026-07-27
test(traces): cover external identification config

stevencartavia:
Adds regression coverage for the external identification timeout. Partial tracing configurations retain the five-second default, and `cast call --disable-external-identification` is verified to suppress external requests through the complete command-to-identifier path.

https://github.com/foundry-rs/foundry/pull/15883#pullrequestreview-4788961503


mablr (approved):

mattsse (approved):
lgtm

PR #15915, opened 2026-07-27
fix(script): match trailing init data

stevencartavia:
Allows external contract verification to match valid creation code containing data after canonical constructor arguments. The complete suffix is preserved for verification submission while malformed ABI encodings and ambiguous candidates remain rejected.

https://github.com/foundry-rs/foundry/pull/15863#discussion_r3647103068


mattsse (approved):
lgtm

mablr (approved):

Issue #15913, opened 2026-07-27
Recommended `forge lint` CI usage should include the `--deny` flag


wjmelements:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

forge Version: 1.7.1 Commit SHA: 4072e48705af9d93e3c0f6e29e93b5e9a40caed8 Build Timestamp: 2026-05-08T07:54:16.980521000Z (1778226856) Build Profile: dist

### What version of Foundryup are you on?

foundryup: 1.9.2

### What command(s) is the bug in?

forge lint

### Operating System

Linux

### Describe the bug

The [recommended](https://www.getfoundry.sh/forge/linting#ci-integration) CI integration is:
```
- name: Run linter
  run: forge lint
```
This is unsatisfactory because lint will always succeed, exiting with status 0.

Therefore, the recommended CI integration should be:
```
- name: Run linter
  run: forge lint --deny warnings --quiet
```
With this usage, the linter step will actually fail if the linter flags any issues. With `--quiet` the unrelated build output is also hushed.

wjmelements:
I also noticed the `forge lint` build will not install missing dependencies in the same way that `forge build` does.

PR #15916, opened 2026-07-27
fix(forge): install dependencies for lint

stevencartavia:
Installs missing Git-submodule dependencies before `forge lint` resolves project sources, matching existing Forge command behavior. Configuration is reloaded after installation so auto-detected remappings include restored dependencies, with the same behavior preserved for the `geiger` alias.

Follow-up to #15913.


mattsse (approved):
lgtm

mablr (approved):

PR #15919, opened 2026-07-27
fix(anvil): make fork resets atomic

mablr:
Anvil currently mutates live fork metadata and fee state while `anvil_reset` is still performing asynchronous, fallible setup. A failed reset can therefore expose partially updated node state, and reset can race active mining.

This prepares the candidate configuration, database, EVM environment, genesis allocation, and fee state independently, then publishes them only after setup succeeds. It also preserves chain-ID override and database RPC provenance, keeps local signers aligned with the active chain after a reset, delays instance-ID changes until the backend commits successfully, and snapshots active blob parameters for each fee-history entry.

This includes #15908, which adds target-fork hardfork and fee inference on top of the atomic reset path. It is stacked on #15931, which isolates persisted fork caches by endpoint so publishing a replacement cannot later be contaminated by the retired database's final cache flush.

This PR was written with Amp, which assisted with code generation, tests, and review. PR responses may also be generated with Amp.


crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -2597,161 +2596,140 @@ impl<N: Network> Backend<N> {
...
-            fork.prepare_reset(reset_urls, block_number.into()).await?;
-            if rpc_url_changed {
-                // Clear state fetched from the previous RPC URL before persisting the cache.
-                fork.database.write().await.clear_into_state_snapshot();
-            }
-            // Persist fetched remote state before rebuilding the fork database so the new
-            // block-specific database can load it from disk.
+            })?;
+        if let Some(fork) = &current_fork {
             fork.database.read().await.maybe_flush_cache().map_err(BlockchainError::Internal)?;
```
mattsse: `BlockchainDbMeta` intentionally ignores the RPC host, so flushing here and rebuilding at a matching `BlockEnv` can load the previous endpoint's accounts/storage. I reproduced this at block 0: after resetting between origins with balances 1337 and 42, `eth_getBalance` still returned 1337.

Please preserve the database RPC URL provenance from #15728, compare it with the target URL, and invalidate the chain cache after validating the replacement endpoint but before constructing the detached DB. A block-0 regression test with identical environments and different allocations would cover this.

mattsse (changes requested):
One fork cache isolation regression.

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -2597,161 +2596,140 @@ impl<N: Network> Backend<N> {
...
+        node_config.fork_choice =
+            forking.block_number.map(|number| ForkChoice::Block(number as i128));
 
-                // reset the time to the timestamp of the forked block
-                self.time.reset(fork_block.header.timestamp());
-                // drop any pending next-block prevrandao override so it does not leak into a block
-                self.cheats.clear_next_block_prevrandao();
+        let mut evm_env = self.evm_env.read().clone();
+        node_config.apply_tempo_fork_beneficiary_default(&mut evm_env);
+        let candidate_fees = self.fees.detached();
```
stevencartavia: `FeeHistoryService` captures `backend.blob_params()` by value at startup, but this reset path can now change the inferred hardfork and blob schedule. With storage caching disabled, I reset from Cancun to Prague and mined a real one-blob EIP-4844 transaction. `eth_config` reported Prague’s nine-blob maximum, and both the receipt and header reported `blobGasUsed = 0x20000`, but `eth_feeHistory` returned `blobGasUsedRatio = 1/6` instead of `1/9`. After a Prague-to-Cancun reset, it returned `1/9` instead of `1/6`, confirming that fee history retains the startup parameters.

Could the service use the active shared blob parameters, snapshot them once per cache entry, and use that snapshot for both `header.blob_fee(...)` and the ratio denominator? A reset regression covering a real blob transaction would also prevent this from recurring.

mattsse (dismissed):
lgtm

crates/anvil/src/eth/backend/mem/mod.rs
mablr: Addressed across the stacked changes. #15931 keys persisted fork caches by RPC endpoint and adds the block-zero regression, including a second fork started while the replacement remains alive. This PR preserves immutable database RPC provenance and compares it with the reset target. Endpoint namespacing replaces directory deletion so the retired backend’s final flush cannot recreate or overwrite the target endpoint’s cache.

crates/anvil/src/eth/backend/mem/mod.rs
mablr: Addressed in `fe2723125`. `FeeHistoryService` now holds the shared `FeeManager`, snapshots `blob_params` once in `create_cache_entry`, and uses that snapshot for both `header.blob_fee(...)` and the ratio denominator. The regression resets Cancun→Prague→Cancun, submits a real one-blob transaction in each target fork, and checks both the blob gas ratio and blob base fee.

PR #15920, opened 2026-07-27
fix(cast): allow remote sponsors with Tempo access keys

gakonst:
## What changed

- allow Tempo Accounts access keys to be combined with `--sponsor-url`
- route access-key-signed transactions through Alloy Tempo's sign-only relay transport
- preserve the original configured RPC transport for the final broadcast, including MPP handling
- apply the behavior consistently to `cast send`, `cast erc20`, and TIP-20 transaction commands
- add an end-to-end Accounts Store + remote sponsor integration test

## Why

The CLI rejected `--sponsor-url` whenever signer resolution selected a Tempo access key, even though the two signatures authorize different parts of a Tempo transaction: the access key signs the user authorization and the sponsor signs the fee-payer authorization.

## Validation

- `cargo +nightly clippy --locked -p cast@1.7.2 --all-targets --all-features -- -D warnings`
- `cargo test --locked -p cast@1.7.2 --test cli 'keychain::send_' -- --nocapture`
- production `cast send` with the existing Tempo Accounts Store, `https://sponsor.tempo.xyz`, and `https://rpc.tempo.xyz`
- production `cast send` with the same Store/sponsor through `https://rpc.mpp.tempo.xyz`
- production `cast erc20 transfer` with the existing Store and remote sponsor

The production receipts were successful type-`0x76` transactions and reported the remote sponsor, rather than the user account, as `feePayer`.

### DIFF
--- a/crates/cast/src/cmd/send.rs
+++ b/crates/cast/src/cmd/send.rs
@@ -275,8 +275,8 @@ impl SendTxArgs {
 
         let timeout = send_tx.timeout.unwrap_or(config.transaction_timeout);
 
-        // --sponsor-url is only valid with a local signer (Case 4). Bail early with a clear
-        // error rather than silently ignoring it in the other signing paths.
+        // --sponsor-url is valid with local signers and Tempo access keys. Bail early rather than
+        // silently ignoring it in signing paths that cannot produce a raw transaction locally.
         if let Some(ref url) = sponsor_url {
             validate_sponsor_url(url)?;
             if unlocked {
@@ -284,9 +284,6 @@ impl SendTxArgs {
             }
             if send_tx.browser.browser {
                 eyre::bail!("--sponsor-url cannot be combined with --browser");
-            }
-            if access_key.is_some() {
-                eyre::bail!("--sponsor-url cannot be combined with a Tempo access key");
             }
         }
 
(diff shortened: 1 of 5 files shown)


crates/cast/src/cmd/send.rs
```diff
@@ -592,6 +602,39 @@ where
...
+) -> Result<B256>
+where
+    N::TransactionRequest: Default + FoundryTransactionBuilder<N>,
+    N::ReceiptResponse: UIfmt + UIfmtReceiptExt,
+{
+    tx.set_fee_payer_signature(FEE_PAYER_SIGNATURE_MARKER);
+    let connector = tempo::sponsor_relay_connector(provider, sponsor_url)?;
+    let sponsor_provider =
+        AlloyProviderBuilder::<_, _, N>::default().connect_with(&connector).await?;
+    cast_send_with_tempo_wallet(
```
stevencartavia, line 624: `--sync` is silently ignored in all three new Tempo access-key + `--sponsor-url` paths (`cast send`, ERC-20, and TIP-20). They do not forward `send_tx.sync`, and the helper always uses `send_raw_transaction`. Could we forward `sync` and use `send_raw_transaction_sync` when requested, matching the existing sync path?

PR #15921, opened 2026-07-27
fix(traces): redact private keys in all signing cheatcodes

mattsse:
The cheatcode input redaction in the trace decoder only matched `sign` and `signP256`, so the remaining signing cheatcodes taking a raw private key printed it in verbose (`-vvvvv`) traces: `signCompact` (both the `uint256` and `Wallet` overloads), `signWithNonceUnsafe`, `signKeychain`, `signKeychainAdmin`, and `signEd25519`, which takes the key as its third parameter. These now redact the key argument to `<pk>` like `sign` does. The digest-only and signer-address overloads (`signCompact(bytes32)` and `signCompact(address,bytes32)`) carry no key material and are unchanged.

Verified end-to-end with a `-vvvvv` trace exercising all of the above that neither the hex nor the decimal rendering of the key appears in the output.

### DIFF
--- a/crates/evm/traces/src/decoder/mod.rs
+++ b/crates/evm/traces/src/decoder/mod.rs
@@ -894,7 +894,7 @@ impl CallTraceDecoder {
                     decoded[0] = DynSolValue::String("<pk>".to_string());
                 }
 
-                Some(decoded.iter().map(format_token).collect())
+                Some(decoded.iter().map(|value| self.format_value(value)).collect())
             }
             "signEd25519" => {
                 let mut decoded = func.abi_decode_input(&data[SELECTOR_LEN..]).ok()?;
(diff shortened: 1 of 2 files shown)


crates/evm/traces/src/decoder/mod.rs
```diff
@@ -894,6 +896,13 @@ impl CallTraceDecoder {
 
                 Some(decoded.iter().map(format_token).collect())
```
stevencartavia: `format_token` bypasses configured and `vm.label` labels for the `signer`/`account` address arguments in `signCompact(address,bytes32)`, `signKeychain`, and `signKeychainAdmin`. Could this iterator use `.map(|value| self.format_value(value))` instead? That restores address labels while leaving the redacted `"<pk>"` rendering unchanged.

mattsse: Good catch — for `signCompact(address,bytes32)`, `signKeychain`, and `signKeychainAdmin` this was a regression from this PR: they previously fell through to the generic decode path, which formats through `format_value` and applies labels. Switched the iterator to `self.format_value` and added a labeled-account case to `test_should_redact` (973c451).

`sign(address,bytes32)` shares this iterator, so it now picks up labels too (it bypassed them even before this PR). Left `signDelegation`/`signAndAttachDelegation` as-is: same pre-existing `format_token` rendering, but that predates this PR and existing script snapshots assert the raw addresses, so changing it is out of scope here.

stevencartavia (approved):

mablr (approved):

PR #15922, opened 2026-07-27
fix(traces): redact private keys in key derivation cheatcodes

mattsse:
Verbose traces redact cheatcode arguments and return values that contain private keys, but the key derivation cheatcodes were missed: `publicKeyP256` and `publicKeyEd25519` printed their raw private-key input, and `createEd25519Key` printed the generated private key in its return data. This adds the two inputs to the existing `<pk>` redaction arm in `decode_cheatcode_inputs` and redacts the `createEd25519Key` output in `decode_cheatcode_outputs`, mirroring `createWallet`, with the new cases covered in `test_should_redact`.

### DIFF
--- a/crates/evm/traces/src/decoder/mod.rs
+++ b/crates/evm/traces/src/decoder/mod.rs
@@ -867,7 +867,8 @@ impl CallTraceDecoder {
                         .collect(),
                 )
             }
-            "addr" | "createWallet" | "deriveKey" | "rememberKey" => {
+            "addr" | "createWallet" | "deriveKey" | "publicKeyEd25519" | "publicKeyP256" |
+            "rememberKey" => {
                 // Redact private key in all cases
                 Some(vec!["<pk>".to_string()])
             }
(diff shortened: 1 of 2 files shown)


crates/evm/traces/src/decoder/mod.rs
```diff
@@ -1021,7 +1022,7 @@ impl CallTraceDecoder {
     fn decode_cheatcode_outputs(&self, func: &Function) -> Option<String> {
         match func.name.as_str() {
             s if s.starts_with("env") => Some("<env var value>"),
-            "createWallet" | "deriveKey" => Some("<pk>"),
+            "createEd25519Key" | "createWallet" | "deriveKey" => Some("<pk>"),
```
stevencartavia, line 1025: `createEd25519Key` returns `salt` unchanged as `privateKey`, so the key remains visible in the call even though the output is redacted:

```text
VM::createEd25519Key(0x...)
  └─ ← [Return] <pk>
```

Would it make sense to also add `"createEd25519Key"` to the unconditional input-redaction arm and cover `createEd25519Key(bytes32)` in the input cases of `test_should_redact`?

mattsse: Good catch — `create_ed25519_key` uses the salt directly as the signing key and returns it unchanged as `privateKey`, so the input is the key itself. Added `createEd25519Key` to the unconditional input-redaction arm and covered `createEd25519Key(bytes32)` in the input cases of `test_should_redact` in 8cfe3f2.

(Response written by Claude Code.)

stevencartavia (approved):

PR #15923, opened 2026-07-27
fix(traces): harden signing cheatcode redaction

mattsse:
Follow-up hardening from the review of the signing cheatcode redaction, whose original scope landed on master via #15921.

`signWithNonceUnsafe` printed its nonce argument in traces. The nonce is the raw ECDSA k value, so together with the visible digest and the returned `(r, s)` it allows recovering the private key, making the `<pk>` redaction cosmetic; the nonce is now redacted as `<nonce>`.

Malformed calldata could bypass redaction entirely: calldata with a known cheatcode selector but a non-canonical ABI shape (for example a nonzero trailing byte) skipped function resolution and was rendered as raw calldata hex by the trace writer, private key included. Calls to the cheatcode address with a known selector are now routed through normal input decoding regardless of shape, and signing cheatcode arguments that cannot be ABI-decoded render as `<redacted>`. An end-to-end test exercises real selector dispatch with generated `Vm` call encodings, covering well-formed, trailing-byte, and truncated calldata.


crates/evm/traces/src/decoder/mod.rs
```diff
@@ -881,11 +881,13 @@ impl CallTraceDecoder {
                 // getNonce(Wallet)
                 (!func.inputs.is_empty() && func.inputs[0].ty == "tuple").then(|| vec!["<pk>".to_string()])
             }
-            "sign" | "signP256" => {
+            "sign" | "signP256" | "signCompact" | "signWithNonceUnsafe" | "signKeychain" |
```
stevencartavia: `signWithNonceUnsafe` still exposes enough information to recover the private key. The rendered trace includes the digest, nonce, and returned `(r, s)`. Trying `s′ = s` and `s′ = n - s` to account for low-`s` normalization gives `d = (s′ * k - z) * r⁻¹ mod n`. Using different inputs recovered the exact input key. Could we redact the nonce as well?

crates/evm/traces/src/decoder/mod.rs
```diff
@@ -894,6 +896,13 @@ impl CallTraceDecoder {
 
                 Some(decoded.iter().map(format_token).collect())
```
stevencartavia: `format_token` in this custom decoding path drops address labels. This reproduces with `signCompact(address,bytes32)`: the base branch prints `compact-signer: [0x…]`, while this PR prints only the address. Could we use `self.format_value` for the decoded arguments after redaction so labels are preserved?

crates/evm/traces/src/decoder/mod.rs
mattsse: Good catch — the nonce here is the raw ECDSA k value, so with the digest and returned (r, s) visible the key is indeed recoverable. Moved signWithNonceUnsafe into its own arm in e1c2378349 that redacts both the key and the nonce (`<nonce>`).

crates/evm/traces/src/decoder/mod.rs
mattsse: Fixed in e1c2378349: the sign/signEd25519/signWithNonceUnsafe arms now format decoded arguments with self.format_value, so labels render as before, and added a labeled signCompact(address,bytes32) test case. Note the same format_token quirk pre-exists on master in the signDelegation/signAndAttachDelegation arm; left that for a follow-up since it's not touched by this PR.

crates/evm/traces/src/decoder/mod.rs
```diff
@@ -881,18 +881,36 @@ impl CallTraceDecoder {
                 // getNonce(Wallet)
                 (!func.inputs.is_empty() && func.inputs[0].ty == "tuple").then(|| vec!["<pk>".to_string()])
             }
-            "sign" | "signP256" => {
+            "sign" | "signP256" | "signCompact" | "signKeychain" | "signKeychainAdmin" => {
```
mablr: Malformed sensitive calldata can bypass this redaction: a known signing selector plus a complete key and a nonzero trailing byte fails the outer ABI-shape check, then fallback rendering prints the raw calldata. Maybe we should recognize known `Vm` selectors before that heuristic and fail closed when sensitive arguments cannot decode.

crates/evm/traces/src/decoder/mod.rs
```diff
@@ -1643,6 +1665,145 @@ mod tests {
                         .to_string(),
                 ]),
             ),
```
mablr, line 1691: These fixtures call the helper with an independently parsed `Function`, so they do not verify selector lookup and would pass if real dispatch stopped reaching redaction. Let's add an end-to-end decode/render regression using generated `Vm` call encoding, including malformed trailing data and asserting the key sentinel is absent.

crates/evm/traces/src/decoder/mod.rs
mattsse: Confirmed the full chain: a nonzero trailing byte fails `is_abi_call_data` (it only tolerates all-zero remainders), `fallback_call_data` returns None for the Vm address, and the writer then prints raw calldata hex. Fixed in f84ef39c37: cheatcode-address calls with a known selector are now routed through normal input decoding before that heuristic, and the signing arms render `<redacted>` when the arguments cannot be ABI-decoded, so known-selector calldata can no longer reach the raw fallback. Unknown selectors still render raw, since there is nothing to recognize them by.

crates/evm/traces/src/decoder/mod.rs
mattsse: Added test_should_redact_end_to_end in f84ef39c37: it drives decode_function with generated Vm::sign_1Call / Vm::signEd25519Call encodings through real selector dispatch and asserts the key hex never appears in the rendered args — covering the well-formed calls, a nonzero trailing byte (signature still resolves, no raw fallback), and truncated calldata (renders `<redacted>`).

mablr (commented):
Once the conflicts are resolved, LGTM.

mablr (approved):

crates/evm/traces/src/decoder/mod.rs
```diff
@@ -776,6 +776,19 @@ impl CallTraceDecoder {
                 return_data: self.decode_function_output(trace, contract_functions).await,
             }
         } else {
+            // Calls to the cheatcode address with a known selector must not fall back to raw
+            // calldata rendering: even malformed calldata can contain sensitive inputs like
+            // private keys that the custom cheatcode decoding redacts.
+            if trace.address == CHEATCODE_ADDRESS
+                && let Some(selector) = cdata.first_chunk().map(Selector::from)
+                && let Some([func, ..]) = self.functions_for_selector(trace.address, &selector)
```
stevencartavia, line 784: `functions_for_selector` falls back to the decoder-wide function map, so selectors outside the `Vm` ABI can be rendered as named `VM` calls here. Malformed `transfer(address,uint256)` calldata is rendered as `VM::transfer(...)`, even though execution then reverts with `unknown cheatcode with selector 0xa9059cbb`. Could this branch look up the selector directly in the `Vm` ABI instead?

mablr (approved):

PR #15926, opened 2026-07-28
test: cover dump state ordering

stevencartavia:
Adds regression coverage for dumpState’s complete ordering contract and contract creation through executeTransaction. This verifies that created accounts remain in deployment order while non-created accounts stay address-sorted.


mablr (approved):

mattsse (approved):
lgtm

PR #15927, opened 2026-07-28
test: cover dump state fork transitions

stevencartavia:
Adds regression coverage for dumpState ordering across block- and transaction-based createSelectFork calls and explicit active or inactive fork rolls.


mablr (approved):

mattsse (approved):
lgtm

PR #15929, opened 2026-07-28
fix(config): preserve nested remappings

### LINKED ISSUE #6431: Forge remappings output wrong with multiple remappings.txt on dependencies.

stevencartavia:
Preserves nested dependency remappings when they refine an auto-detected package root, while retaining broad package fallbacks and root/CLI precedence. 

Fixes #6431.


crates/config/src/providers/remappings.rs
```diff
@@ -116,6 +116,39 @@ impl Remappings {
...
+            })
+        };
+        let suppressed = remappings
+            .iter()
+            .enumerate()
+            .filter(|(_, remapping)| {
+                is_generated_refinement(remapping)
+                    && self.remappings.iter().any(|existing| {
+                        existing.context.is_none()
+                            && remapping_names_overlap(&existing.name, &remapping.name)
```
mablr: This reconstructs whether a remapping was generated from its shape, but an explicitly configured contextual/global pair can satisfy the same predicate. For example, `lib/dep/:pkg/=lib/dep/pkg/contracts/` plus `pkg/=lib/dep/pkg/` is classified as generated and the contextual mapping is dropped when a CLI `pkg/sub/=` override is present, even with auto-detection disabled. Could we preserve provenance through this merge (or merge CLI aliases before contextual generation) instead of inferring it from `Remapping` values? Please add a regression for an explicit contextual/global pair with `auto_detect_remappings = false`.

crates/config/src/providers/remappings.rs
```diff
@@ -198,6 +231,8 @@ impl RemappingsProvider<'_> {
         }
 
         user_remappings.extend(remappings);
+        let user_remapping_names =
+            user_remappings.iter().map(|r| r.name.clone()).collect::<Vec<_>>();
```
mablr: This drops each user remapping context and later treats any overlapping name as global. A root mapping such as `test/:inner/=...` would therefore suppress the refinement needed under `lib/outer/`, despite applying to an unrelated context. Please retain the context here and only let an authoritative mapping suppress/override refinements in contexts where it actually applies; a regression with two unrelated contexts would make this clear.

crates/config/src/providers/remappings.rs
```diff
@@ -210,9 +245,29 @@ impl RemappingsProvider<'_> {
...
-            for r in nested_foundry_remappings {
+            let mut contextual_remappings = Vec::new();
+            for (lib, r) in nested_foundry_remappings {
+                // A dependency can intentionally refine an auto-detected package root to its
+                // source directory. Scope that refinement to the dependency so root imports keep
+                // the broader package mapping.
+                if r.context.is_none()
+                    && !user_remapping_names
+                        .iter()
+                        .any(|root| remapping_names_overlap(root, &r.name))
```
mablr: The symmetric overlap check suppresses the entire broad dependency refinement when the root only defines a narrower alias. With root `inner/sub/=src/local/`, dependency `inner/=lib/inner/contracts/`, and auto `inner/=lib/outer/lib/inner/`, `inner/sub/Selected.sol` resolves correctly but an `inner/I.sol` import from the dependency falls back to `lib/outer/lib/inner/I.sol` instead of `contracts/I.sol`. This also conflicts with the directional fallback rule in `Remappings::push`. Please preserve both behaviors—likely by overlaying authoritative aliases into the generated dependency context so the narrow alias and broad refinement coexist—and extend `root_remapping_precedes_nested_refinement` with an unaffected broad import for both config and CLI cases.

crates/config/src/providers/remappings.rs
```diff
@@ -210,9 +245,29 @@ impl RemappingsProvider<'_> {
...
+                    && auto_detected_remappings.iter().any(|auto| {
+                        auto.context == r.context
+                            && auto.name == r.name
+                            && auto.path != r.path
+                            && Path::new(&r.path).starts_with(&auto.path)
```
mablr, line 281: These paths can describe the same dependency using different identities: `foundry_toml_dirs` canonicalizes nested roots, while `Remapping::find_many` scans the lexical install path. For `lib/outer -> /external/outer`, this compares `/external/outer/lib/inner/contracts/` with `/project/lib/outer/lib/inner/`, so `starts_with` is false and no refinement is generated. Please retain/translate back to the lexical installation path for comparison and emitted contexts, and add a Unix symlink regression.

crates/config/src/providers/remappings.rs
```diff
@@ -242,14 +298,14 @@ impl RemappingsProvider<'_> {
...
         self.lib_paths
             .par_iter()
             .map(|p| if p.is_absolute() { self.root.join("lib") } else { self.root.join(p) })
             .flat_map(foundry_toml_dirs)
             .flat_map_iter(|lib| {
                 trace!(?lib, "find all remappings of nested foundry.toml");
-                self.nested_foundry_remappings(&lib)
+                self.nested_foundry_remappings(&lib).into_iter().map(move |r| (lib.clone(), r))
             })
             .collect::<Vec<_>>()
```
mablr: The newly emitted `contextual_remappings` preserve this collection order, but `foundry_toml_dirs` does not sort `WalkDir` results and this pipeline is parallel. Unlike `lib_remappings`, these entries bypass the `BTreeMap` normalization, so `forge remappings` output—and first-wins behavior for duplicate context/name candidates—can vary by filesystem or Rayon scheduling. Could we sort dependency/context groups deterministically while preserving declared remapping order within each dependency? A two-dependency regression under different `RAYON_NUM_THREADS` values would cover this.

crates/config/src/providers/remappings.rs
stevencartavia: [d5e62bd](https://github.com/foundry-rs/foundry/pull/15929/commits/d5e62bd295e440aeb944d66c0bd5802dc1a1588a)

crates/config/src/providers/remappings.rs
stevencartavia: [d5e62bd](https://github.com/foundry-rs/foundry/pull/15929/commits/d5e62bd295e440aeb944d66c0bd5802dc1a1588a)

crates/config/src/providers/remappings.rs
stevencartavia: [d5e62bd](https://github.com/foundry-rs/foundry/pull/15929/commits/d5e62bd295e440aeb944d66c0bd5802dc1a1588a)

crates/config/src/providers/remappings.rs
stevencartavia: follow-up so this doesn't get bloated

crates/config/src/providers/remappings.rs
stevencartavia: follow-up

mablr: okay

crates/config/src/providers/remappings.rs
mablr: The PR explicitly claims it retains “broad package fallbacks and root/CLI precedence,” but the implementation loses the broad fallback. This should be in imo.

stevencartavia: [4b75109](https://github.com/foundry-rs/foundry/pull/15929/commits/4b75109d79c0fd18b12a7d70d4d6b47c6919cefa)

figtracer (dismissed):

mablr (dismissed):

crates/config/src/providers/remappings.rs
mattsse, line 1: this needs new docs

crates/config/src/providers/remappings.rs
```diff
@@ -116,6 +120,45 @@ impl Remappings {
             self.push(remapping);
         }
     }
+
+    /// Extract generated contextual refinements from a Figment.
+    pub fn generated_from_figment(figment: &Figment) -> Vec<Remapping> {
+        figment.extract_inner(GENERATED_REMAPPINGS_KEY).unwrap_or_default()
+    }
+
+    /// Extend with lower-precedence remappings, preserving global alias precedence over generated
```
mattsse: unclea what lower precendce is or what this does

crates/config/src/providers/remappings.rs
```diff
@@ -149,7 +192,10 @@ impl RemappingsProvider<'_> {
     /// - `remappings.txt`
     /// - Environment variables
     /// - CLI parameters
-    fn get_remappings(&self, remappings: Vec<Remapping>) -> Result<Vec<Remapping>, Error> {
+    fn get_remappings(
+        &self,
+        remappings: Vec<Remapping>,
+    ) -> Result<(Vec<Remapping>, Vec<Remapping>), Error> {
```
mattsse: why does this return two vecs

mattsse (commented):
not super clear to me what this does

crates/config/src/providers/remappings.rs
stevencartavia: [b66c4b1](https://github.com/foundry-rs/foundry/pull/15929/commits/b66c4b10c75c1f65b308c7b3b5bffa5804c5ec36) - Replaced the tuple with named fields separating all remappings from generated refinements needed for the later CLI merge.

crates/config/src/providers/remappings.rs
stevencartavia: [b66c4b1](https://github.com/foundry-rs/foundry/pull/15929/commits/b66c4b10c75c1f65b308c7b3b5bffa5804c5ec36) - Renamed it to `extend_with_config_remappings` and documented the exact CLI/config precedence rules.

crates/config/src/providers/remappings.rs
stevencartavia: https://github.com/foundry-rs/book/pull/2025

mablr (approved):
LGTM, pending @mattsse

mattsse (approved):
lgtm

PR #15930, opened 2026-07-28
fix(cast): honor sync for Tempo access keys

mablr:
## Motivation

Follow-up to https://github.com/foundry-rs/foundry/pull/15920#discussion_r3661811521, this ensures `--sync` is honored when Cast signs Tempo transactions with an Accounts access key. Sponsored and unsponsored access-key paths now forward the send mode and use `eth_sendRawTransactionSync` instead of silently falling back to `eth_sendRawTransaction` followed by receipt polling.

Raw transaction submission now shares one sync-aware path across cast send, ERC-20, TIP-20, batch sends, and virtual-address registration. This also fixes `cast vaddr create --sync --json`, which previously changed submission behavior in JSON mode while still preserving its single composite JSON response.

### DIFF
--- a/.changelog/tempo-access-key-sync.md
+++ b/.changelog/tempo-access-key-sync.md
@@ -1,0 +1,5 @@
+---
+cast: patch
+---
+
+Honor `--sync` when sending transactions with Tempo Accounts access keys.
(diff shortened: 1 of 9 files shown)


mattsse (approved):
lgtm

grandizzy (approved):
lgtm

PR #15931, opened 2026-07-28
fix(anvil): isolate fork caches by endpoint

mablr:
Persisted fork caches are currently keyed only by chain ID and block number, allowing databases backed by different RPC endpoints to share a writer. A retired database can consequently recreate stale endpoint state when its final cache guard is dropped.

This keys Anvil cache files by a hash of the primary RPC URL, keeps legacy and endpoint-specific files visible to cache reporting, and adds block-zero coverage that starts another fork while the replacement remains alive. Existing legacy files remain manageable but are intentionally not reused by Anvil because their endpoint provenance is ambiguous.

This is a prerequisite for #15919: atomic fork publication can otherwise drop the retired database after constructing its replacement and recreate the stale shared cache.

This PR was written with Amp, which assisted with code generation, tests, and review. PR responses may also be generated with Amp.

### DIFF
--- a/.changelog/anvil-fork-cache-isolation.md
+++ b/.changelog/anvil-fork-cache-isolation.md
@@ -1,0 +1,5 @@
+---
+anvil: patch
+---
+
+Isolate persisted fork caches by RPC endpoint.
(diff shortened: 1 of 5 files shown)


mattsse (approved):
lgtm

grandizzy (approved):
lgtm

PR #15933, opened 2026-07-28
fix(debugger): advance repeated PC hits

figtracer:
`:continue <pc>` currently uses the same target selection as `:pc <pc>`. When execution is already at a repeated program counter, it selects the current step again and reports `Already at`, making later loop iterations unreachable through the continue command.

Give `:continue` next-hit selection while preserving the existing direct-goto behavior of `:pc`. The implementation reuses the debugger's trace traversal and current-contract matching, so repeated commands advance through loop iterations and split trace segments without duplicating PC parsing or diagnostics.

Progresses #927.

### DIFF
--- a/.changelog/debugger-continue-pc.md
+++ b/.changelog/debugger-continue-pc.md
@@ -1,0 +1,5 @@
+---
+foundry-debugger: patch
+---
+
+Fixed `:continue <pc>` to advance through repeated program counter hits.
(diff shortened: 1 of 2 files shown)


mablr (approved):

grandizzy (approved):
lgtm

mattsse (approved):
lgtm

PR #15934, opened 2026-07-28
fix(anvil): canonicalize `eth_simulateV1` state roots

mablr:
## Description

Anvil constructed simulated state roots from incomplete candidate overlays and retained zero-valued storage or deleted accounts in trie inputs, producing non-canonical headers. This recursively merges candidate and base state, applies canonical absence rules across full, incremental, and historical paths, and returns the resulting root from local `eth_simulateV1` responses. Coverage includes storage updates and deletions plus London `SELFDESTRUCT` parity between simulated and mined roots.

Closes OSS-598. Fork-backed root policy and broader response assembly remain in OSS-565; block transitions, blob handling, and mining finalization are unchanged.


mattsse (dismissed):
lgtm

mablr (commented):
Sorry @mattsse I forgot the changelog.

crates/anvil/src/eth/backend/mem/in_memory_db.rs
```diff
@@ -152,7 +152,13 @@ impl HistoricalStateCache {
         };
 
         for (address, dirty) in std::mem::take(&mut self.dirty) {
-            let Some(account) = db.inner.cache.accounts.get(&address) else {
+            let Some(account) = db
+                .inner
+                .cache
+                .accounts
+                .get(&address)
+                .filter(|account| account.account_state != AccountState::NotExisting)
```
stevencartavia: Filtering `NotExisting` in `from_mem_db()` and `HistoricalStateCache::snapshot()` discards tombstones from the internal historical state. `PersistentStateDb::basic_ref()` returns `Some(AccountInfo::default())` for an absent entry but `None` for a tombstone, so historical simulation treats a deleted account as existing-empty. This affects pre-EIP-161 execution.

Reproduced on Frontier: starting from historical and latest bases with identical roots, the historical `eth_simulateV1` root differed from mining, while the latest simulation matched. Can we retain tombstones internally and filter them only in `state_snapshot()` and `full_db()`?

crates/anvil/src/eth/backend/db.rs
```diff
@@ -342,16 +388,41 @@ impl<T: DatabaseRef<Error = DatabaseError> + Send + Sync + Clone + fmt::Debug> D
...
+            }
+            if overlay.account_state == AccountState::StorageCleared {
+                accounts.insert(*address, overlay.clone());
+                continue;
+            }
+
+            let mut account = accounts.remove(address).unwrap_or_default();
+            account.info = overlay.info.clone();
+            account.account_state = overlay.account_state.clone();
+            account.storage.extend(overlay.storage.clone());
```
stevencartavia, line 420: This merge can restore storage deleted by a previous simulated block. On London, after one block self-destructs a contract and a later block transfers value to its address, REVM changes the overlay from `NotExisting` to `Touched` while its storage remains empty. This path then treats it as a partial overlay and retains the original base storage.

Reproduced with equivalent two-block simulation and mining: mined storage was zero, but the second simulated root differed from the second mined root. Please preserve the storage-cleared provenance across this transition, for example as `NotExisting` → `StorageCleared`.

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5858,8 +5858,28 @@ impl Backend<FoundryNetwork> {
...
+                        .copied()
+                        .collect::<Vec<_>>();
+
                     // commit the transaction
                     cache_db.commit(state);
+                    for address in previously_deleted {
+                        if let Some(account) = cache_db.cache.accounts.get_mut(&address)
+                            && account.account_state != AccountState::NotExisting
+                        {
+                            account.account_state = AccountState::StorageCleared;
```
stevencartavia: The normal selfdestruct-then-transfer path is fixed, but state overrides still bypass the new guard because `apply_state_overrides` commits directly to the cache.

I reproduced this after a simulated `SELFDESTRUCT`: a code-only override that reads slot 0 returns the deleted value `42`, while the same override with `state: {}` returns `0`; their roots also differ. Could we preserve `NotExisting → StorageCleared` around state-override application too and add this case as a regression?

grandizzy:
cyclops audit fast

grandizzy:
cyclops audit fast

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5696,7 +5696,12 @@ impl Backend<FoundryNetwork> {
                 // Apply state overrides after validating precompile moves against this block's
                 // active precompile set.
                 if let Some(state_overrides) = state_overrides {
+                    let previously_deleted = previously_deleted_accounts(
+                        &cache_db.cache.accounts,
+                        state_overrides.keys().copied(),
+                    );
                     apply_state_overrides(state_overrides, &mut cache_db)?;
+                    preserve_deleted_storage(&mut cache_db.cache.accounts, previously_deleted);
```
figtracer, line 5711: an empty override still changes the returned root. reproduced live with `blockStateCalls: [{}]` vs `blockStateCalls: [{"stateOverrides":{"0x...42":{}}}]`: the roots differ even though no field is overridden. can we skip no-op entries before applying them and add this regression?

figtracer (approved):

mattsse (approved):
lgtm

PR #15935, opened 2026-07-28
fix(cli): estimate Tempo WebAuthn gas

figtracer:
Tempo browser wallets can sign transactions with WebAuthn. Foundry previously estimated the unsigned request and added a fixed 7,000-gas buffer, but Tempo also charges calldata gas for the variable-length WebAuthn payload. A live 169-byte signature used 41,656 gas while the old estimate plus buffer provided only 41,358.

This replaces the fixed buffer with an estimation-only Tempo request that supplies `keyType: webAuthn` and the conservative 1,400-byte `keyData` hint used by Viem. The hint is used by `cast send` and its token/keychain flows, `forge create`, and `forge script`; it is never forwarded to the browser wallet. User-supplied gas limits continue to take precedence.

For CREATE transactions, only the estimation clone is converted to a Tempo AA call, so the RPC receives a valid request without changing what Foundry submits.

Codex was used to investigate the issue and prepare the implementation.

### DIFF
--- a/.changelog/gas-estimate-multiplier.md
+++ b/.changelog/gas-estimate-multiplier.md
@@ -1,6 +1,0 @@
----
-forge: minor
-cast: minor
----
-
-Added `--gas-estimate-multiplier` to `forge create` and `cast send`.
--- a/.changelog/gas-estimate-multiplier.md
+++ b/.changelog/gas-estimate-multiplier.md
@@ -1,0 +1,6 @@
+---
+forge: minor
+cast: minor
+---
+
+Added `--gas-estimate-multiplier` to `forge create` and `cast send`.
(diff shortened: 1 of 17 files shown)


mablr (commented):
I'm not sure we want this, are there still cases where there is slight underestimation like mentioned in the initial issue ? (maybe w tempo wallet signature?)

mattsse (dismissed):
lgtm

grandizzy (dismissed):
lgtm! before merge @figtracer pls check @mablr comment https://github.com/foundry-rs/foundry/pull/15935#pullrequestreview-4798171398

stevencartavia (dismissed):

mablr (commented):
sgtm

mablr (dismissed):

mablr (approved):

mattsse (approved):
lgtm

PR #15936, opened 2026-07-28
perf(forge): speed up path-filtered coverage

figtracer:
Path-filtered `forge coverage` now compiles only conventional `.t.sol` roots selected by `--match-path` or `--no-match-path`, while retaining all project sources, scripts, non-test fixtures, and transitive imports. This avoids code generation for tests that cannot execute under the filter and intentionally omits unmatched test roots from the report; the regression covers an imported unmatched helper, an invalid unrelated test, omission of an unselected valid test, and script reporting. Developed with OpenAI Codex assistance.

### Results

Profiling builds from exact master `c3285d584` on Solady `5dc5fc8`, using an identical controlled 20-test source set, `forge clean` before every sample, and 10 runs after one warmup:

| Version | Time (mean ± σ) | Range | Speedup |
| --- | ---: | ---: | ---: |
| master | 1.833s ± 0.036s | 1.768–1.877s | — |
| this PR | 1.127s ± 0.014s | 1.096–1.141s | 1.63× |

### DIFF
--- a/.changelog/forge-path-filtered-coverage.md
+++ b/.changelog/forge-path-filtered-coverage.md
@@ -1,0 +1,5 @@
+---
+forge: patch
+---
+
+Sped up path-filtered coverage by compiling only matching test roots.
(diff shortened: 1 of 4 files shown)


mablr (approved):

mattsse (approved):
lgtm

stevencartavia (approved):

PR #15937, opened 2026-07-28
fix(config): tolerate cache listing races

mablr:
`forge cache ls` can race with other Foundry processes that remove or replace entries in the shared global cache. An entry may disappear after directory enumeration but before its metadata is read, causing the entire command to fail with `NotFound`. The endpoint-specific fork cache layout introduced in #15931 exposed this pre-existing enumeration race.

This change treats vanished entries as absent while continuing to propagate other filesystem errors. It refreshes entry metadata from the current path and avoids following symlinks when sizing enumerated block and file entries, applying the same race handling to RPC and block-explorer caches. Regression coverage exercises removed and replaced entries, symlinks, preservation of non-`NotFound` errors, and both cache-listing CLI modes.

AI assistance was used to investigate the failure and prepare an initial implementation; I manually revised and validated the final code and tests.


crates/config/src/lib.rs
```diff
@@ -2418,19 +2418,24 @@ impl Config {
...
-                for entry in entries.flatten().filter(|x| x.path().is_dir()) {
-                    if let Ok(chain) = Chain::from_str(&entry.file_name().to_string_lossy()) {
-                        cache.chains.push(Self::list_foundry_chain_cache(chain)?);
-                    }
+            };
+            for entry in entries {
+                let Some(entry) = Self::ignore_not_found(entry)? else {
+                    continue;
+                };
+                let Some(metadata) = Self::ignore_not_found(fs::metadata(entry.path()))? else {
```
stevencartavia, line 2431: Symlinked chain directories are still followed for both RPC and explorer caches. This behavior predates this PR, but it makes the description’s unqualified “avoids following symlinks” claim too broad. Could we either skip symlinked chain directories or narrow the claim to the entries this change now checks?

crates/config/src/lib.rs
```diff
@@ -5690,6 +5745,99 @@ mod tests {
...
+    #[test]
+    fn list_cached_blocks_ignores_symlinks() -> eyre::Result<()> {
+        let chain_dir = tempdir()?;
+        let target_dir = tempdir()?;
+        fs::write(target_dir.path().join("storage.json"), [0; 10])?;
+        std::os::unix::fs::symlink(target_dir.path(), chain_dir.path().join("1"))?;
+
+        let block_path = chain_dir.path().join("2");
+        fs::create_dir(&block_path)?;
+        let target_file = tempfile::NamedTempFile::new()?;
```
stevencartavia, line 5821: `target_file` is empty, so this assertion also passes if the symlink is followed: its reported size is zero and the block is omitted anyway. Could we write nonzero content to `target_file` before creating the symlink so this test catches that regression?

stevencartavia (dismissed):
The cache-listing race fix looks good.

figtracer (approved):

stevencartavia (approved):

PR #15938, opened 2026-07-28
fix(anvil): canonicalize `eth_simulateV1`  blob txs

mablr:
## Description

`eth_simulateV1` previously treated pooled EIP-4844 requests differently across execution, returned transactions, and block construction. In particular, sidecar commitments were not consistently converted into versioned hashes before execution, pooled sidecars could reach the returned representation, and simulated headers did not carry canonical blob gas accounting across blocks.

This change canonicalizes Ethereum requests before simulation while preserving the typed Tempo and OP request paths. It derives versioned hashes from sidecars, resolves one transaction type and fee representation for both execution and output, strips sidecars before constructing transaction envelopes and roots, and enforces the active Cancun, Prague, Osaka, BPO1, or BPO2 cumulative block limit. Consecutive simulated blocks now derive `blobGasUsed` and `excessBlobGas` from their parent inputs and the child block’s active blob schedule, including parameter-only BPO timestamp transitions.

The change remains isolated to simulation behavior and does not alter state-root calculation, system transitions, mining finalization, shared pool limits, or malformed-sidecar/KZG validation.

Closes OSS-600

> Assisted by Amp


mattsse (dismissed):
lgtm

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5701,6 +5779,36 @@ impl Backend<FoundryNetwork> {
...
+                    if is_ethereum_request {
+                        request.populate_blob_hashes();
+                        let preferred_type = request.preferred_type();
+                        request.transaction_type = Some(preferred_type as u8);
+                        request.trim_conflicting_keys();
+                        request.populate_blob_hashes();
+                        if preferred_type == TxType::Eip4844
+                            && request.max_fee_per_blob_gas.is_none()
+                        {
+                            request.max_fee_per_blob_gas = block_env.blob_gasprice();
```
stevencartavia: For `eth_simulateV1`, omitted `maxFeePerBlobGas` defaults to `0`; it should not be copied from the block. In the reproduced Cancun case, this fallback returns `0x1` in both modes and lets `validation: true` pass instead of rejecting cap `0` below blob base fee `1`.

Please let `prep_for_submission()` materialize the zero cap. With `validation: false`, default the call-local blob price to zero only when no `blockOverrides.blobBaseFee` is provided; explicit overrides should remain effective. Test that validation enabled rejects, while disabled succeeds and returns `maxFeePerBlobGas: 0x0`.

figtracer (approved):

mattsse (approved):
lgtm

PR #15924, opened 2026-07-27
refactor(tempo): reuse canonical MPP transport and Accounts store

gakonst:
## Summary

- replace Foundry’s custom MPP HTTP and WebSocket execution with `alloy-transport-mpp`’s canonical transports
- delete the separate `wallet/sessions.toml` registry entirely
- persist Foundry-managed temporary access keys in the canonical Accounts `wallet/store.json`
- resolve `--tempo.session` by its signed authorization witness and pin the exact Accounts key
- persist MPP session channels only in Accounts SDK `wallet/channels.db`, including under an overridden `TEMPO_HOME`
- compose the Accounts signer, autoswap, batched channel funding, durable cross-process delivery leases, stale-channel recovery, and cancellation cleanup from merged mpp-rs
- make paid MPP CI select the Accounts store’s active wallet and reject `MPP_API_KEY`, so it cannot silently bypass HTTP 402
- preserve unrelated Accounts keys and retain authorization metadata needed for revoke retries

The combined change removes 3,832 lines and adds 1,219 (net -2,613).

## Validation

- `cargo +nightly fmt --all -- --check`
- warnings-denied all-target/all-feature Clippy for Foundry common and Cast
- full `foundry-common` library tests: 106 passed, 1 ignored
- 68 focused Cast keychain, session, and sponsorship tests passed, including Accounts-store signing, remote sponsorship, run-for cleanup, child/grandchild inheritance, Forge script submission, and on-chain revoke
- mpp-rs #363 and #364: gate, integration, lint, tests, conformance, cross-SDK, deny, changelog, and security checks all passed
- real `cast keychain doctor` against the current `~/.tempo/wallet/store.json` through `https://rpc.mpp.tempo.xyz`: signing, registration, expiry, scopes, limits, and paid RPC balance checks all passed without `MPP_API_KEY`
- real mainnet paid-RPC matrix with the current active Accounts wallet: two Cast payments, Forge script, Forge `createSelectFork`, Anvil fork, and Chisel fork all passed on chain 4217
- that matrix reused one durable channel and advanced its cumulative voucher from 368,000 to 481,000 with batched top-ups
- prior 20-process Cast stress: 20/20 succeeded with one durable channel and no SQLite or voucher races
- real `cast send` and `cast erc20 transfer` through the paid RPC, signed by the Accounts-store access key and remotely sponsored by `https://sponsor.tempo.xyz`; both landed as type-`0x76` transactions
[...]

PR #15940, opened 2026-07-28
fix(tempo): lock Accounts store mutations

mablr:
This serializes Foundry's whole-file Tempo Accounts store mutations behind a shared in-process mutex and cross-process file lock. Session creation, retirement, expiry cleanup, interactive login, and access-key imports now use the same critical section, preventing concurrent operations from dropping persistent keys or restoring retired key material. A subprocess regression test exercises the lost-update boundary.

Follow-up to https://github.com/foundry-rs/foundry/pull/15924#discussion_r3665734691.

> This PR was written with Codex, including code generation and test implementation. Reviewed manually.

### DIFF
--- a/.changelog/tempo-accounts-store-lock.md
+++ b/.changelog/tempo-accounts-store-lock.md
@@ -1,0 +1,6 @@
+---
+cast: patch
+foundry-common: patch
+---
+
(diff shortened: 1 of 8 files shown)


mattsse (approved):
lgtm

stevencartavia (approved):

PR #15942, opened 2026-07-29
fix(mutation): prune equivalent bounds

gretzke:
Mutation testing currently generates every comparison replacement from the untyped Solar AST. This leaves provably equivalent mutants such as `uint256 value == 0` to `value <= 0`, which survive even with complete tests and lower the reported mutation score.

This change runs Solar lowering and type analysis once for the compiled project, maps inferred integer and address ranges back to per-file AST spans, and removes only replacements whose predicates match the original for every value in the type range. Tautological but behavior-changing replacements remain. In particular, `value == 0` still mutates to `value < 0`.

The analysis covers signed and unsigned bounds, address bounds, reversed operands, explicit casts, `type(T).min` and `type(T).max`, and inferred member or expression types. If the analysis cannot prove equivalence, the mutant remains. The mutation cache salt is updated because the generated set changes. Unit and CLI coverage includes the retained false-boundary mutation and the adjusted existing mutation summary.

Codex was used to inspect the mutation pipeline, implement the change, and generate tests.

### DIFF
--- a/.changelog/forge-mutation-equivalent-bounds.md
+++ b/.changelog/forge-mutation-equivalent-bounds.md
@@ -1,0 +1,5 @@
+---
+forge: patch
+---
+
+Stopped Forge mutation testing from generating provably equivalent integer and address boundary comparisons.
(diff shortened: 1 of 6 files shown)


stevencartavia (dismissed):

stevencartavia (approved):

grandizzy (approved):
lgtm, ty!

PR #15882, opened 2026-07-25
ci: validate PR changelogs

stevencartavia:
Adds deterministic validation for pull request changelog entries and documents the `L-ignore` exemption.

Trusted same-repository PRs receive an Amp-prefilled “Add changelog” link, while fork PRs receive generic guidance without exposing secrets. AI remains advisory and cannot affect the validation result.

Closes OSS-588

Stacked on #15881.

PR #15946, opened 2026-07-29
fix(ci): validate changelogs on stale forks

mablr:
Pull requests opened before #15882 retain a base SHA from before the trusted changelog validator existed. The changelog job checked the validator out at that historical SHA and then correctly refused to execute the external contributor's copy, causing valid changelog entries to fail without being inspected.

This loads the validator from the current trusted target branch while continuing to validate the recorded base and head SHAs. It also removes the bootstrap path entirely, so contributor-controlled validator code is never executed. The check remains blocking and the existing `L-ignore` exemption is unchanged.

This PR was written with Codex assistance.


figtracer (approved):

mattsse (approved):
lgtm

PR #15947, opened 2026-07-29
feat(script): wait for confirmations

### LINKED ISSUE #4748: Script Wait for confirmations
### Component

Forge

Right now, the forge script verifier gets a little confused if it doesn't find any bytecode, and just exists. An easy solution I use with Brownie, Hardhat, Truffle, and Ape is to just wait X confirmations before attempting a verification.

Having a flag to wait x confirmations would be great.

```
forge script script/MyScript.sol --wait-confirmations 6
```

And this will result in:

```
##
Waiting for receipts.
⠉ [00:00:12] [######################################################################################################################################################################################] 6/6 receipts (0.0s)
```

### Additional context

_No response_

figtracer:
Adds a `--confirmations <N>` option to `forge script`, defaulting to the existing one-confirmation behavior and rejecting zero. Fresh and resumed broadcasts now pass the requested depth through the shared receipt watcher, while Tempo batch broadcasts apply the same rule before recording their receipt. Contract verification therefore starts only after the requested confirmation depth has been reached.

Closes #4748.


mablr (approved):

mattsse (changes requested):
Two correctness issues need to be addressed before this can merge:

[P1] Preserve the Tempo batch checkpoint when the transaction is mined but still awaiting confirmations (`crates/script/src/broadcast.rs:919`). `wait_for_batch_receipt` now keeps running after a receipt exists until the requested depth is reached, but the resume timeout branch treats any timeout as a dead hash and clears every stamped transaction hash. If a batch is included and the chain does not produce enough additional blocks before `transaction_timeout`, the next `--resume` can rebroadcast calls that already executed. Please distinguish “no receipt” from “receipt awaiting confirmations,” retain the mined batch hash in the latter case, and add a regression test.

[P2] Apply an explicit timeout around the complete generic confirmation wait (`crates/script/src/receipts.rs:46`). In Alloy 2.2.0, when `watch_pending_transaction` finds an existing receipt with more than one required confirmation, it registers the watcher with `received_at_block`; the heartbeat then returns before adding the configured timeout. Consequently a resumed or quickly mined transaction can wait forever if the chain stops before reaching the requested depth. The new test always mines the remaining blocks, so it does not cover this case. Please enforce the deadline around the full wait and add a stalled-chain regression test.

mattsse (approved):
lgtm

mablr (approved):

PR #15948, opened 2026-07-29
perf(forge): speed up forked scripts

figtracer:
Forked `forge script` runs now retain successful chain IDs fetched by the existing post-execution Shanghai compatibility check and reuse them for trace decoding and simulation provider setup, instead of sending two more `eth_chainId` requests per RPC URL. The cache is scoped to exact URLs and does not cross script execution, preserving explicit `--chain`, failed-lookup fallbacks, silent optimization candidates, and scripts that mutate RPC state while running. Developed with OpenAI Codex assistance.

### Results

Local debug builds from exact master `0c6ed5d09`, using `ScriptAdditionalContracts` through a deterministic proxy that delays only `eth_chainId` by 200ms. Hyperfine used 5 warmups and 20 measured runs, removing generated `out`, `cache`, and `broadcast` output before each sample:

| Version | `eth_chainId` requests | Time (mean ± σ) | Range | Speedup |
| master | 4 | 1.264s ± 0.042s | 1.237–1.439s | — |
| this PR | 2 | 841.6ms ± 9.1ms | 826.1–866.5ms | 1.50× |

### DIFF
--- a/.changelog/forge-script-chain-id-reuse.md
+++ b/.changelog/forge-script-chain-id-reuse.md
@@ -1,0 +1,5 @@
+---
+forge: patch
+---
+
+Reduced redundant chain ID requests when running scripts against a fork.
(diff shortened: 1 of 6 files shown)


mablr (approved):

mattsse (approved):
lgtm

Issue #15878, opened 2026-07-24
bug(anvil): transaction-hash forks mishandle protocol system transactions


haythemsellami:
### Component

Anvil

### Describe the bug

`anvil --fork-transaction-hash` reconstructs the selected block prefix by converting source envelopes into `PoolTransaction`s and feeding them through ordinary pool validation and mining. That model is incorrect for networks with protocol-level system envelopes.

On Monad, canonical staking reward, snapshot, and epoch envelopes have protocol semantics that ordinary pool transactions do not: they bypass normal fee admission, validate and increment the system nonce, mint reward value when applicable, execute a dedicated staking system call, retain logs, and account zero receipt/block gas. Today these envelopes can be rejected before execution with errors such as `GasTooLow` or a base-fee validation failure.

The same forced-transaction path validates historical typed transactions against the configured local execution chain ID. Consequently, a source transaction from Monad chain ID 143 can be rejected when the fork is run with a different local `--chain-id`. Forced source transactions can also be coalesced with live local pool transactions even though they require different execution context and provenance.

### Reproduction

Choose a Monad transaction whose preceding block prefix contains a canonical protocol envelope, then start Anvil with that transaction hash:

```console
anvil --fork-url <MONAD_RPC_URL> --fork-transaction-hash <TRANSACTION_HASH>
```

The protocol envelope is routed through ordinary transaction validation instead of being replayed with its source-chain semantics, so the reconstructed prefix is rejected or produces incorrect state.

### Expected behavior

The source block prefix should be replayed as a dedicated historical batch using the exact target block environment, source chain ID, and active source hardfork. Canonical protocol envelopes should use the network factory protocol replay semantics and produce zero-gas receipts where required. The batch should commit atomically, remain isolated from live local pool transactions, and restore the configured local chain ID afterward.

### Suggested direction

Preserve explicit historical provenance rather than converting every source envelope to `PoolTransaction`; isolate the replay batch from the live pool; and stage, commit, or revert the entire prefix atomically. Ordinary source transactions should execute under source context, while protocol envelopes should use the same network replay hook already used by `vm.transact`.

This was exposed while integrating Monad in #15343. Related historical reports include #8940 and #13630.

mablr:
What if we treat transaction-hash forks on Monad as a synchronous historical replay step performed before the pool and miner exist?

figtracer:
retested on current master: the monad protocol-system-prefix transaction-hash replay regression passes. closing as completed.

PR #15952, opened 2026-07-29
fix(anvil): isolate transaction hash replay

mablr:
Transaction-hash forks currently convert the selected source prefix into forced pool transactions. This lets historical replay share miner control flow with live transactions and makes completion depend on the configured mining mode.

This resolves the target and its parent against hash-consistent source blocks, prepares the complete inclusive prefix, and executes it against an isolated backend overlay during startup. The prefix is published as one local Anvil block only after the full batch succeeds; reverted transactions remain included, while fatal failures publish nothing. The live pool and miner are constructed afterward, removing forced replay state from `ClientForkConfig` and `Miner`.

This establishes the network-neutral replay boundary needed by #15878 for #15343. Source-chain-specific protocol replay semantics remain follow-up work.

AI assistance disclosure: Codex was used for codebase analysis and an initial implementation; the final implementation was largely reworked and completed by the author.

### DIFF
--- a/.changelog/anvil-transaction-hash-replay.md
+++ b/.changelog/anvil-transaction-hash-replay.md
@@ -1,0 +1,5 @@
+---
+anvil: patch
+---
+
+Replayed transaction-hash fork prefixes atomically during startup before local mining begins.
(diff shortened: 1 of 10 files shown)


mablr:
@haythemsellami please let me know if this approach is convenient for you :)

mattsse (approved):
lgtm

figtracer (approved):

Issue #15950, opened 2026-07-29
feat(symex): add SLOAD/SSTORE hook cheatcodes


decofe:
## Description

Expose storage hooks as Forge cheatcodes that register Solidity callbacks:

```solidity
vm.registerSloadHook(address(target), this.onLoad.selector);
vm.registerSstoreHook(address(target), this.onStore.selector);
```

Callbacks receive:

```solidity
onLoad(address account, bytes32 slot, bytes32 value);
onStore(address account, bytes32 slot, bytes32 oldValue, bytes32 newValue);
```

The cheatcodes are the user-facing registration API. Internally, symex should instrument `SLOAD` and `SSTORE`, invoke registered callbacks on every feasible symbolic path, and treat callback state as revert-aware ghost state.

## Example

```solidity
uint256 ghostTotal;

function setUp() public {
    vm.registerSstoreHook(address(vault), this.onStore.selector);
}

function onStore(
    address,
    bytes32,
    bytes32 oldValue,
    bytes32 newValue
) external {
    ghostTotal = ghostTotal - uint256(oldValue) + uint256(newValue);
}

function testSymex_totalTracksStorage(address user, uint256 amount) public {
    vault.deposit(user, amount);
    assertEq(ghostTotal, vault.totalAssets());
}
```

## Implementation scope

- Add hook-registration cheatcodes.
- Instrument raw `(address, slot)` accesses; source-level storage-layout matching can follow.
- Support concrete and symbolic slots and values.
- Preserve the correct storage context through calls and `delegatecall`.
- Roll back hook effects when the corresponding EVM context reverts.
- Prevent hook bookkeeping from recursively triggering hooks.

## Tests

- Correct `SLOAD` address, slot, and value.
- Correct `SSTORE` old and new values.
- Symbolic mapping slots and values.
- Multiple writes, nested calls, and `delegatecall`.
- Revert rollback and contract filtering.
- No-hook performance baseline.

Inspired by [Certora storage hooks](https://docs.certora.com/en/latest/docs/cvl/hooks.html).

Suggested by: @clement-ux

clement-ux:
## Motivation: ERC20 Balance Accounting
  A concrete motivating use case is ERC20 balance accounting:

```solidity
      totalSupply() == sum(balanceOf(account)) for every address
```

  Without storage hooks, our Foundry symbolic invariant has to use a closed, finite holder set:

```solidity
      balanceOf(alice) + balanceOf(bob) == totalSupply()
```

[...]

PR #15954, opened 2026-07-29
feat(symex): add storage hook cheatcodes

mablr:
## Description

Implements the raw storage-hook foundation of #15950. The stacked mapping-aware extension in #15976 completes the mapping-specific ERC20 accounting use case described in the issue discussion.

This adds `vm.registerSloadHook` and `vm.registerSstoreHook`, allowing callers to register callbacks for storage reads and writes associated with a target storage address. The callbacks receive the effective storage address, computed slot, and observed values, and are supported by both concrete and symbolic execution.

Storage-hook callbacks execute as instrumentation frames after the matching opcode. Their stack, gas, returndata, access warmth, and guidance effects do not leak into the parent frame. Callback failures revert the instrumented execution, successful callback writes roll back with the enclosing EVM context, registrations survive EVM reverts, and hooks are suppressed throughout callback subtrees. Because callbacks only exist during simulation and may mutate state or cause later calls to revert, both registration cheatcodes are classified as unsafe for scripts.

This PR intentionally operates on raw effective storage addresses and computed slots. It does not expose mapping roots or decoded keys; #15976 adds that behavior as a separate reviewable layer built on this callback lifecycle.

A warmed no-hook `forge test` benchmark on `ithacaxyz/account:v0.5.7` and `vectorized/solady:v0.1.26`, using profiling builds, found no statistically meaningful regression.

Mablr developed this PR with assistance from Amp.


crates/cheatcodes/src/inspector.rs
```diff
@@ -2921,6 +3056,144 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
...
+    ///
+    /// Returns whether a failed callback was propagated to the parent frame.
+    #[inline]
+    pub fn finish_storage_hook_callback(&mut self, interpreter: &mut Interpreter) -> bool {
+        let Some(active) = self.active_storage_hook.as_ref() else { return false };
+        let Some((result, output)) = active.outcome.clone() else { return false };
+
+        let active = self.active_storage_hook.take().expect("active storage hook exists");
+        let _ = interpreter.stack.pop();
+        interpreter.gas = active.saved_gas;
```
mattsse, line 3138: Callback cleanup restores the parent Gas value, but a successful callback leaves its account and storage accesses warm in the shared transaction journal. A callback that reads an otherwise-cold account or slot therefore makes a later user access cheaper and can change gasleft() branches or out-of-gas behavior. A practical fix is to snapshot transaction access state before installing the synthetic frame—warm_addresses plus the warm marker and transaction ID for loaded accounts and storage slots—and restore only that access metadata when the callback finishes, while retaining callback world-state writes. Accounts or slots first loaded by the callback should become cold again unless the pre-callback snapshot, access list, or precompile rules already made them warm. Please add a regression where the callback reads an unrelated cold slot and assert that a later target access has the same gas delta with and without the hook.

mattsse (commented):
One correctness issue to address: storage-hook callbacks currently leak EIP-2929 access warmth into the instrumented execution. I left an inline suggestion with a concrete restoration strategy.

crates/cheatcodes/src/inspector.rs
```diff
@@ -2921,6 +3116,187 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
...
+        &mut self,
+        interpreter: &mut Interpreter,
+        ecx: &mut FoundryContextFor<'_, FEN>,
+    ) -> bool {
+        let Some(active) = self.active_storage_hook.as_ref() else { return false };
+        let Some((result, output)) = active.outcome.clone() else { return false };
+
+        let active = self.active_storage_hook.take().expect("active storage hook exists");
+        active.access_snapshot.restore::<FEN>(ecx);
+        let _ = interpreter.stack.pop();
```
figtracer, line 3133: [P1] This pop assumes the synthetic CALL successfully pushed a return word. A valid `SLOAD` can leave all 1,024 stack slots occupied; revm then cannot push the callback success word, and this pop removes the actual SLOAD value. On this head I reproduced `1024 × PUSH0; SLOAD; JUMP`: the baseline succeeds, while registering a no-op hook changes the same execution to `InvalidJump`, both normally and with `--isolate`. Please reserve/restore the SLOAD result or use a continuation that does not insert a normal CALL result, and add this bytecode as a regression.

crates/evm/evm/src/inspectors/stack.rs
```diff
@@ -1406,17 +1416,24 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>>
...
-            |inspector| {
-                let mut out = None;
-                if let Some(output) = inspector.call(ecx, call) {
-                    out = Some(Some(output));
+        let storage_hook_callback = self
+            .cheatcodes
+            .as_deref()
+            .is_some_and(|cheatcodes| cheatcodes.is_storage_hook_callback(ecx, call));
+
+        if !storage_hook_callback {
```
figtracer: [P1] This suppresses only `Fuzzer::call` for the exact callback root. The callback opcode stream still runs through `edge_coverage.step` and `fuzzer.step`, and calls made by the callback no longer match this root predicate. With an identical seed/run count, a no-hook target emitted 16 frontiers while the hooked target emitted 20—the extra sites came from the callback branch; a longer artifact likewise contained callback selector/body sites attributed to the parent test sequence. This can steer corpus retention, comparison mutation, and `--symbolic-use-fuzz-frontiers` into instrumentation, while nested callback calls can still be overridden or recorded. Please track the active hook subtree and suppress edge/comparison/fuzzer `step`, `call`, and `call_end` for the whole subtree, with frontier/showmap and nested-helper regressions.

crates/evm/symbolic/README.md
```diff
@@ -217,6 +217,36 @@ Current modeling notes:
...
+Storage hooks can maintain revert-aware ghost state for raw storage accesses.
+Register a callback during `setUp` for each target and access kind:
+
+```solidity
+function setUp() public {
+    vm.registerSloadHook(address(vault), this.onLoad.selector);
+    vm.registerSstoreHook(address(vault), this.onStore.selector);
+}
+
+function onLoad(address account, bytes32 slot, bytes32 value) external {
```
figtracer, line 229: [P1] The documented callback is an ordinary unguarded external function. I reproduced a target directly calling `onStore` with a fabricated `newValue = 999`; the ghost changed even though no `SSTORE` occurred. Concrete and symbolic synthetic frames already use `address(vm)` as `msg.sender`, so the example and tests should require `msg.sender == address(vm)` (or the framework should enforce equivalent isolation). Please also ensure registered callback selectors cannot become default invariant targets when callbacks live on a handler.

crates/cheatcodes/src/inspector.rs
```diff
@@ -2921,6 +3116,187 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
...
+            ));
+            true
+        }
+    }
+
+    fn snapshot_storage_hook_access(
+        ecx: &mut FoundryContextFor<'_, FEN>,
+    ) -> StorageHookAccessSnapshot {
+        let (_, journal) = ecx.db_journal_inner_mut();
+        let accounts = journal
```
figtracer: [P2] This copies access metadata for every journal-loaded account and storage slot on every hooked access. A warm distinct-slot loop with a no-op SLOAD hook measured 0.22s at 1k slots, 0.50s at 2k, 1.54s at 4k, and 5.49s at 8k, versus 0.15s for 8k unhooked reads; the scaling is approaching quadratic as loaded slots accumulate. The PR's no-hook benchmark cannot detect this active-path cost. Please use an access-delta strategy or otherwise bound this work, and add a hook-enabled scaling benchmark before treating the API as stable.

crates/forge/tests/cli/test_cmd/symbolic_cheatcodes.rs
```diff
@@ -4234,3 +4234,481 @@ contract SymbolicDeployCodeCheatcode is Test {
...
+        recursiveHookCalls++;
+    }
+
+    function branchingStoreHook(address, bytes32, bytes32, bytes32 newValue) external pure {
+        if (uint256(newValue) == 7) {
+            revert("seven");
+        }
+    }
+
+    function checkSymbolicMapping(address account, uint256 newValue) public {
```
figtracer, line 4735: [P2] This covers only one fresh mapping write, and the checked callback masks unknown-old paths by reverting on underflow. An unchecked two-write mapping probe with symbolic keys constrained equal and `ghost = ghost - oldValue + newValue` reports a replay-rejected candidate under the default `storage_layout = "solidity"`, while the same property passes with `zero_init`; separate checks confirm the aliased second `oldValue` itself is correct. Please add repeated-key aliasing plus pre-existing/nonzero base-state coverage, and document or configure the initialization/storage model required by this ghost pattern. Otherwise the advertised mapping-ghost workflow fails red by default even though replay prevents a false confirmed counterexample.

crates/cheatcodes/spec/src/vm.rs
```diff
@@ -416,6 +416,20 @@ interface Vm {
     #[cheatcode(group = Evm, safety = Safe)]
     function accesses(address target) external view returns (bytes32[] memory readSlots, bytes32[] memory writeSlots);
 
+    /// Registers a callback invoked after every SLOAD in the target's storage context.
+    ///
+    /// The callback must have the signature `function(address,bytes32,bytes32) external`.
+    /// Registering another callback for the same target replaces the previous callback.
+    #[cheatcode(group = Evm, safety = Unsafe)]
```
figtracer: [P2] Omitting `status` makes these generated cheatcodes `Stable`, but the current API exposes only the opaque final slot. It cannot identify `_balances[user]` versus `_allowances[owner][spender]` or recover mapping keys, so the motivating ERC20 aggregate cannot yet be expressed safely. Please mark the raw hooks `Experimental` while their public semantics settle, and keep a mapping-provenance follow-up with a real balances/allowances/mint/burn/transfer/approve acceptance test.

figtracer (changes requested):
Re-reviewed current head `cdfccd53e74a6511044fface614b5e788097ca86` after the access-warmth fix. That original leak is resolved: the callback and baseline both retain the 2,610-gas cold access cost, and the new upstream regression passes.

Requesting changes because a no-op SLOAD hook still changes valid EVM execution, and callback execution still leaks into fuzz guidance and remains externally spoofable. I also left bounded performance/model comments that matter before stabilizing this API.

Fresh validation on this head included a Forge rebuild, the upstream concrete+symbolic hook CLI test, normal and isolated full-stack repros, cold-access parity, fresh branch-frontier comparison, direct callback spoofing, and default-versus-zero-init symbolic alias probes. The red MPP check is unrelated to this feature.

crates/cheatcodes/src/inspector.rs
stevencartavia: +1 from my pass

crates/evm/evm/src/inspectors/stack.rs
stevencartavia: +1 from my pass

stevencartavia (dismissed):

crates/cheatcodes/src/inspector.rs
```diff
@@ -2957,6 +3100,184 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
...
+    }
+
+    fn restore_storage_hook_access(ecx: &mut FoundryContextFor<'_, FEN>, journal_start: usize) {
+        let (_, journal) = ecx.db_journal_inner_mut();
+        let entries =
+            journal.journal.drain(journal_start.min(journal.journal.len())..).collect_vec();
+        for entry in entries {
+            match entry {
+                JournalEntry::AccountWarmed { address } => {
+                    journal.state.get_mut(&address).expect("warmed account exists").mark_cold();
```
grandizzy, line 3188: ### 🔴 Blocker — warmth restoration corrupts EIP-2200 `original_value` with locked `revm-state 41.0.0`

The drain-and-`mark_cold()` approach is conceptually right, but REVM 41.0.0 resets `original_value = present_value` whenever a slot is marked cold — even mid-transaction. See [`mark_cold` / `mark_warm_with_transaction_id` in 41.0.0](https://docs.rs/revm-state/41.0.0/src/revm_state/lib.rs.html#603-629).

**Failing sequence:**
1. Callback writes a previously cold ghost slot from `0` → `1`.
2. Cleanup marks that slot cold while retaining the write.
3. Parent later accesses the ghost slot.
4. REVM 41.0.0 rewarms it and incorrectly changes `original_value` from `0` → `1`.
5. Clearing it back to zero is priced as a clean nonzero clear (~7000 gas, ~4800 refund) rather than a dirty reset (~2200 gas, ~19900 refund).

Under a bounded gas limit this changes success into OOG and rolls back an otherwise-valid execution.

Fixed upstream in [`revm` commit 84fe60c](https://github.com/bluealloy/revm/commit/84fe60ca382aa903180d6e49ecd156d2aa6a1957), released in `revm-state 41.1.0`.

**Fix:** Update the lockfile to `revm-state >= 41.1.0`. Add a regression where a callback writes a cold ghost slot and the parent subsequently clears/reaccesses it, asserting gas/refund parity.

crates/evm/symbolic/src/runtime/state.rs
```diff
@@ -366,6 +409,8 @@ impl PathState {
         self.expected_creates = reverted.expected_creates.clone();
         self.call_mocks = reverted.call_mocks.clone();
         self.function_mocks = reverted.function_mocks.clone();
+        self.storage_load_hooks = reverted.storage_load_hooks.clone();
```
grandizzy, line 421: ### 🟠 High — symbolic registrations made by successful invariant/`afterInvariant` checks are discarded

`merge_reverted_top_level_effects` correctly copies both hook maps here (lines 412-413). However, `merge_noncommitting_check_constraints` (line 394, just above this hunk) only merges constraints, symbol allocation, and replay metadata — it does **not** merge hook maps.

`merge_noncommitting_check_constraints` is used after successful invariant and `afterInvariant` checks in both the reverted-handler and successful-handler flows. So a registration made inside a successful invariant check:
```
invariant/afterInvariant registers or replaces hook
→ check succeeds
→ world changes are intentionally discarded
→ registration is accidentally discarded too
→ later handler does not invoke the hook
```

Concrete execution retains the registration (it's inspector state), so concrete and symbolic campaigns diverge.

**Fix:** Copy both `storage_load_hooks` and `storage_store_hooks` in `merge_noncommitting_check_constraints` as well. Add tests for registration from a successful invariant, `afterInvariant`, and registration replacement.

crates/cheatcodes/spec/src/vm.rs
```diff
@@ -416,6 +416,28 @@ interface Vm {
     #[cheatcode(group = Evm, safety = Safe)]
     function accesses(address target) external view returns (bytes32[] memory readSlots, bytes32[] memory writeSlots);
 
+    /// Registers a callback invoked after each SLOAD against `target`'s effective storage account,
+    /// including when its code runs by delegatecall.
+    ///
+    /// The callback must have the signature `function(address,bytes32,bytes32) external`.
```
grandizzy, line 422: ### 🟡 Medium — callback authentication is user-enforced but not documented as mandatory

The internal synthetic-frame identification is sound (`is_storage_hook_callback` checks `caller == CHEATCODE_ADDRESS`, exact parent depth, exact target/input). An ordinary external caller cannot spoof that.

However, the callback is still a public/external Solidity function callable by anyone unless the user adds:
```solidity
require(msg.sender == address(vm), "only storage hook");
```

The PR's tests and README example use this defense, but the **interface documentation** here does not state it as a mandatory requirement. A user who copies the signature without the auth check has a spoofable callback.

**Fix:** Add a sentence to both doc comments: "The callback **must** authenticate `msg.sender == address(vm)` to prevent external spoofing." Consider a reusable modifier in forge-std.

crates/cheatcodes/src/inspector.rs
```diff
@@ -2957,6 +3100,184 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
...
+            parent_depth,
+            callback_target: hook.callback_target,
+            callback_input: input.clone(),
+            saved_gas,
+            saved_return_data,
+            saved_stack_item,
+            journal_start,
+            outcome: None,
+        });
+        interpreter.bytecode.set_action(InterpreterAction::NewFrame(FrameInput::Call(Box::new(
```
grandizzy, line 3312: ### 🟡 Medium — synthetic frame consumes protocol call depth

The callback is scheduled as an ordinary `InterpreterAction::NewFrame`, so it passes through REVM's normal call-depth limit (1024). At maximum legal parent depth:
```
without hook: SLOAD succeeds
with hook:    SLOAD succeeds
              callback frame is rejected as CallTooDeep
              callback failure propagates as parent revert
```

This is a rare boundary case, but instrumentation changes an otherwise-valid execution. Symbolic callbacks likely consume a corresponding call-depth level through `storage_hook_child`.

**Fix options:**
1. Use a synthetic frame that does not consume one of the instrumented program's call-depth slots (may require REVM support).
2. Document the limitation and add a depth-boundary test before stabilization.

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -1,6 +1,89 @@
...
+            parent.next_symbol = outcome.state.next_symbol;
+            parent.storage_load_hooks = outcome.state.storage_load_hooks.clone();
+            parent.storage_store_hooks = outcome.state.storage_store_hooks.clone();
+            parent.storage_hook_active = false;
+
+            match outcome.status {
+                TopLevelCallStatus::Success => {
+                    parent.world = outcome.state.world.clone();
+                    parent.block = outcome.state.block.clone();
+                    parent.recorded_logs = outcome.state.recorded_logs.clone();
```
grandizzy: ### 🟡 Medium — expectations, mocks, logs, and access recording are not isolated

The PR clearly isolates gas, returndata, access warmth, and fuzz/coverage guidance. But on callback success, the symbolic path merges `recorded_logs`, `access_record`, `expected_emit`, `expected_calls`, `expected_creates`, `call_mocks`, and `function_mocks` from the child into the parent.

Consequences:
- A callback helper could satisfy the parent's `expectCall`.
- A callback event could satisfy `expectEmit` or appear in `recordLogs`.
- A parent mock could replace callback helper behavior.
- Callback ghost accesses could appear in `vm.accesses`.

Concrete and symbolic appear consistent here, so this may be intentional. But "synthetic instrumentation frame" suggests invisibility, while the current behavior is "ordinary external function."

**Fix:** Choose and document one policy:
1. **Instrumentation-visible:** explicitly document that these inspector fields include callback subtrees, and add parity tests.
2. **Instrumentation-hidden:** snapshot/filter these fields while retaining journaled EVM writes and registration changes.

Option 2 is less surprising for hidden instrumentation.

grandizzy (commented):
## Oracle-Assisted Review

Ran an automated deep-dive review of this PR. The concrete lifecycle (capture → invoke → restore) is mostly sound, and the P1 issues from earlier review rounds appear fixed. Below are additional findings the current head may still need to address before stabilization.

### Findings Summary

| # | Severity | Finding | Status |
|---|----------|---------|--------|
| 1 | 🔴 Blocker | Warmth restoration corrupts EIP-2200 `original_value` with locked `revm-state 41.0.0` | Needs fix |
| 2 | 🟠 High | Symbolic registrations made by successful invariant/`afterInvariant` checks are discarded | Needs fix |
| 3 | 🟡 Medium | Callback authentication is user-enforced but not documented as mandatory | Needs docs |
| 4 | 🟡 Medium | Synthetic frame consumes protocol call depth | Needs decision/docs |
| 5 | 🟡 Medium | Expectations, mocks, logs, and access recording are not isolated | Needs policy decision |
| 6 | 🟢 Low | Non-static callbacks beneath STATICCALL are intentional but under-documented | Needs docs |

Inline comments are attached to each finding. Happy to discuss any of these.

---
*Review generated with assistance from Amp's Oracle (GPT-5.6 Sol). Posted by @grandizzy.*

crates/cheatcodes/src/inspector.rs
```diff
@@ -2957,6 +3100,184 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
...
+            CallInputs {
+                input: CallInput::Bytes(input),
+                return_memory_offset: 0..0,
+                gas_limit,
+                reservoir: 0,
+                bytecode_address: hook.callback_target,
+                known_bytecode,
+                target_address: hook.callback_target,
+                caller: CHEATCODE_ADDRESS,
+                value: CallValue::Transfer(U256::ZERO),
```
grandizzy, line 3322: ### 🟢 Low — non-static callbacks beneath STATICCALL are intentional but under-documented

Both concrete and symbolic engines create callbacks with `is_static = false`, even when the observed SLOAD/SSTORE occurs beneath a `STATICCALL`. This lets an SLOAD callback update ghost state while instrumenting view getters — a reasonable instrumentation exception.

But it means:
```
STATICCALL → target SLOAD → callback SSTORE (succeeds, mutates ghost state)
```
...can successfully mutate ghost state even though ordinary descendants of a static frame cannot.

**Fix:** Document explicitly that callbacks do not inherit staticness. Add concrete and symbolic tests proving:
- The target still cannot perform static-forbidden operations.
- The callback can update ghost state under STATICCALL.
- The ghost update rolls back if an enclosing context later reverts.

crates/cheatcodes/src/inspector.rs
grandizzy, line 3188: ### 🔴 Blocker — warmth restoration corrupts EIP-2200 `original_value` with locked `revm-state 41.0.0`

The drain-and-`mark_cold()` approach is conceptually right, but REVM 41.0.0 resets `original_value = present_value` whenever a slot is marked cold — even mid-transaction. See [`mark_cold` / `mark_warm_with_transaction_id` in 41.0.0](https://docs.rs/revm-state/41.0.0/src/revm_state/lib.rs.html#603-629).

**Failing sequence:**
1. Callback writes a previously cold ghost slot from `0` → `1`.
2. Cleanup marks that slot cold while retaining the write.
3. Parent later accesses the ghost slot.
4. REVM 41.0.0 rewarms it and incorrectly changes `original_value` from `0` → `1`.
5. Clearing it back to zero is priced as a clean nonzero clear (~7000 gas, ~4800 refund) rather than a dirty reset (~2200 gas, ~19900 refund).

Under a bounded gas limit this changes success into OOG and rolls back an otherwise-valid execution.

Fixed upstream in [`revm` commit 84fe60c](https://github.com/bluealloy/revm/commit/84fe60ca382aa903180d6e49ecd156d2aa6a1957), released in `revm-state 41.1.0`.

**Fix:** Update the lockfile to `revm-state >= 41.1.0`. Add a regression where a callback writes a cold ghost slot and the parent subsequently clears/reaccesses it, asserting gas/refund parity.

crates/evm/symbolic/src/runtime/state.rs
grandizzy, line 421: ### 🟠 High — symbolic registrations made by successful invariant/`afterInvariant` checks are discarded

`merge_reverted_top_level_effects` correctly copies both hook maps here (lines 412-413). However, `merge_noncommitting_check_constraints` (line 394, just above this hunk) only merges constraints, symbol allocation, and replay metadata — it does **not** merge hook maps.

`merge_noncommitting_check_constraints` is used after successful invariant and `afterInvariant` checks in both the reverted-handler and successful-handler flows. So a registration made inside a successful invariant check:
```
invariant/afterInvariant registers or replaces hook
→ check succeeds
→ world changes are intentionally discarded
→ registration is accidentally discarded too
→ later handler does not invoke the hook
```

Concrete execution retains the registration (it's inspector state), so concrete and symbolic campaigns diverge.

**Fix:** Copy both `storage_load_hooks` and `storage_store_hooks` in `merge_noncommitting_check_constraints` as well. Add tests for registration from a successful invariant, `afterInvariant`, and registration replacement.

crates/cheatcodes/spec/src/vm.rs
grandizzy, line 422: ### 🟡 Medium — callback authentication is user-enforced but not documented as mandatory

The internal synthetic-frame identification is sound (`is_storage_hook_callback` checks `caller == CHEATCODE_ADDRESS`, exact parent depth, exact target/input). An ordinary external caller cannot spoof that.

However, the callback is still a public/external Solidity function callable by anyone unless the user adds:
```solidity
require(msg.sender == address(vm), "only storage hook");
```

The PR's tests and README example use this defense, but the **interface documentation** here does not state it as a mandatory requirement. A user who copies the signature without the auth check has a spoofable callback.

**Fix:** Add a sentence to both doc comments: "The callback **must** authenticate `msg.sender == address(vm)` to prevent external spoofing." Consider a reusable modifier in forge-std.

crates/cheatcodes/src/inspector.rs
grandizzy, line 3312: ### 🟡 Medium — synthetic frame consumes protocol call depth

The callback is scheduled as an ordinary `InterpreterAction::NewFrame`, so it passes through REVM's normal call-depth limit (1024). At maximum legal parent depth:
```
without hook: SLOAD succeeds
with hook:    SLOAD succeeds
              callback frame is rejected as CallTooDeep
              callback failure propagates as parent revert
```

This is a rare boundary case, but instrumentation changes an otherwise-valid execution. Symbolic callbacks likely consume a corresponding call-depth level through `storage_hook_child`.

**Fix options:**
1. Use a synthetic frame that does not consume one of the instrumented program's call-depth slots (may require REVM support).
2. Document the limitation and add a depth-boundary test before stabilization.

crates/evm/symbolic/src/executor/opcodes.rs
grandizzy: ### 🟡 Medium — expectations, mocks, logs, and access recording are not isolated

The PR clearly isolates gas, returndata, access warmth, and fuzz/coverage guidance. But on callback success, the symbolic path merges `recorded_logs`, `access_record`, `expected_emit`, `expected_calls`, `expected_creates`, `call_mocks`, and `function_mocks` from the child into the parent.

Consequences:
- A callback helper could satisfy the parent's `expectCall`.
- A callback event could satisfy `expectEmit` or appear in `recordLogs`.
- A parent mock could replace callback helper behavior.
- Callback ghost accesses could appear in `vm.accesses`.

Concrete and symbolic appear consistent here, so this may be intentional. But "synthetic instrumentation frame" suggests invisibility, while the current behavior is "ordinary external function."

**Fix:** Choose and document one policy:
1. **Instrumentation-visible:** explicitly document that these inspector fields include callback subtrees, and add parity tests.
2. **Instrumentation-hidden:** snapshot/filter these fields while retaining journaled EVM writes and registration changes.

Option 2 is less surprising for hidden instrumentation.

crates/cheatcodes/src/inspector.rs
grandizzy, line 3322: ### 🟢 Low — non-static callbacks beneath STATICCALL are intentional but under-documented

Both concrete and symbolic engines create callbacks with `is_static = false`, even when the observed SLOAD/SSTORE occurs beneath a `STATICCALL`. This lets an SLOAD callback update ghost state while instrumenting view getters — a reasonable instrumentation exception.

But it means:
```
STATICCALL → target SLOAD → callback SSTORE (succeeds, mutates ghost state)
```
...can successfully mutate ghost state even though ordinary descendants of a static frame cannot.

**Fix:** Document explicitly that callbacks do not inherit staticness. Add concrete and symbolic tests proving:
- The target still cannot perform static-forbidden operations.
- The callback can update ghost state under STATICCALL.
- The ghost update rolls back if an enclosing context later reverts.

grandizzy (changes requested):
## Oracle-Assisted Review

Ran an automated deep-dive review of this PR. The concrete lifecycle (capture → invoke → restore) is mostly sound, and the P1 issues from earlier review rounds appear fixed. Below are additional findings the current head may still need to address before stabilization.

### Findings Summary

| # | Severity | Finding | Status |
|---|----------|---------|--------|
| 1 | 🔴 Blocker | Warmth restoration corrupts EIP-2200 `original_value` with locked `revm-state 41.0.0` | Needs fix |
| 2 | 🟠 High | Symbolic registrations made by successful invariant/`afterInvariant` checks are discarded | Needs fix |
| 3 | 🟡 Medium | Callback authentication is user-enforced but not documented as mandatory | Needs docs |
| 4 | 🟡 Medium | Synthetic frame consumes protocol call depth | Needs decision/docs |
| 5 | 🟡 Medium | Expectations, mocks, logs, and access recording are not isolated | Needs policy decision |
| 6 | 🟢 Low | Non-static callbacks beneath STATICCALL are intentional but under-documented | Needs docs |

Inline comments are attached to each finding. Happy to discuss any of these.

---
*Review generated with assistance from Amp's Oracle (GPT-5.6 Sol). Posted by @grandizzy.*

crates/cheatcodes/src/inspector.rs
mablr: Revm 42 bump can't be done in the scope of this PR, I will track this separately. 57e3074597b3503d443acbb9ef4d8151b4942088

crates/cheatcodes/src/inspector.rs
mablr: Documented https://github.com/foundry-rs/foundry/pull/15954/commits/5b32b3c531ee1f27545f1c01c847ddeabd5cf3bb

grandizzy (approved):
lgtm, ty!

figtracer (approved):

PR #15956, opened 2026-07-29
feat(forge): check storage compatibility

mattsse:
Adds `forge storage-layout`, a semantic compatibility check for compiler-reported storage entries that compares slots, offsets, recursive type shape, and declaring contracts while allowing non-overlapping appends. The command reads raw layout or artifact JSON, defaults to `.clone.meta` for clone workflows, emits text or structured JSON, and returns a failing exit status when compatibility cannot be established.

The result is deliberately scoped to the compiler `storage` array; EIP-7201 and manually computed slots absent from that array, enum member changes, and contract behavior are not treated as checked.

This PR was written primarily with Codex.

### DIFF
--- a/crates/forge/src/cmd/storage_layout.rs
+++ b/crates/forge/src/cmd/storage_layout.rs
@@ -1,0 +1,866 @@
+use alloy_primitives::U256;
+use clap::Parser;
+use eyre::{Context, Result};
+use foundry_cli::opts::{BuildOpts, CompilerOpts};
+use foundry_common::{
+    compile::{PathOrContractInfo, ProjectCompiler},
+    find_matching_contract_artifact, find_target_path, shell,
+};
+use foundry_compilers::artifacts::{
+    Storage, StorageLayout, StorageType,
+    output_selection::{ContractOutputSelection, OutputSelection},
+};
+use serde::Serialize;
+use serde_json::Value;
+use std::{
+    collections::HashSet,
+    fs,
+    path::{Path, PathBuf},
+    str::FromStr,
+};
+
+const LAYOUT_SCOPE: &str = "compilerStorageLayout";
+const LAYOUT_LIMITATION: &str = "Only entries in the compiler-reported `storage` array are \
+    compared. Namespaced (including EIP-7201) and manually computed slots are outside this check \
+    unless represented in that array; enum member changes and state-variable behavior are not \
+    checked.";
+
+/// CLI arguments for `forge storage-layout`.
+#[derive(Clone, Debug, Parser)]
+pub struct StorageLayoutArgs {
+    /// The contract whose current storage layout should be checked.
+    #[arg(value_parser = PathOrContractInfo::from_str)]
+    pub contract: Option<PathOrContractInfo>,
+
+    /// A previous compiler storage layout, artifact, or `.clone.meta` file.
+    ///
+    /// Defaults to `.clone.meta` in the project root.
+    #[arg(long, short, value_hint = clap::ValueHint::FilePath, value_name = "PATH")]
+    pub reference: Option<PathBuf>,
+
+    /// Treat variable and struct-member renames as compatible.
+    ///
+    /// Without this flag, label changes fail because compiler metadata cannot distinguish a safe
+    /// rename from reusing an existing slot for different state.
+    #[arg(long)]
+    pub allow_renames: bool,
+
+    /// All build arguments are supported.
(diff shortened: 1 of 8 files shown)


crates/forge/src/cmd/storage_layout.rs
```diff
@@ -0,0 +1,866 @@
...
+) -> bool {
+    if !visited.insert((previous_id.to_owned(), current_id.to_owned())) {
+        return true;
+    }
+    let (Some(previous_type), Some(current_type)) =
+        (previous.types.get(previous_id), current.types.get(current_id))
+    else {
+        return false;
+    };
+    if previous_type.encoding != current_type.encoding
```
mablr, line 395: **Blocker:** `solc` omits a UDVT’s underlying type here, so same-name/same-width changes such as `type Amount is uint256` → `bytes32` compare compatible. Mark reachable UDVTs unverifiable and add a compiler-generated regression test.

crates/forge/src/cmd/storage_layout.rs
```diff
@@ -0,0 +1,866 @@
...
+                message: format!(
+                    "State variable `{}` changed type from `{}` to `{}`.",
+                    previous_storage.label,
+                    storage_type_label(previous, previous_storage),
+                    storage_type_label(current, current_storage)
+                ),
+            });
+        }
+
+        let previous_contract = declaring_contract(&previous_storage.contract);
```
mablr, line 303: `Storage.contract` is the most-derived output contract, not declaration provenance. This can reject harmless contract renames and cannot prove inheritance movement. Please remove this check and rely on order, slot, offset, and type shape.

crates/forge/src/cmd/storage_layout.rs
```diff
@@ -0,0 +1,866 @@
...
+        eyre::ensure!(report.compatible, "compiler storage layout is incompatible");
+        Ok(())
+    }
+}
+
+/// A machine-readable semantic storage layout compatibility report.
+#[derive(Clone, Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+struct CompatibilityReport {
+    compatible: bool,
```
mablr, line 115: Please distinguish `compatible`, `incompatible`, and `unverifiable` in text and JSON. Missing semantic metadata should not be collapsed into ordinary incompatibility.

crates/forge/src/cmd/storage_layout.rs
```diff
@@ -0,0 +1,866 @@
...
+        if !user_extra_output && !project.build_info {
+            project.no_artifacts = true;
+            project.update_output_selection(|selection| {
+                *selection = OutputSelection::common_output_selection([
+                    ContractOutputSelection::StorageLayout.to_string(),
+                ]);
+            });
+        }
+
+        let reference = reference.unwrap_or_else(|| project.root().join(".clone.meta"));
```
mablr, line 77: Explicit relative references remain CWD-relative, unlike the default. Please resolve them against `project.root()` for consistent `--root` behavior.

crates/forge/tests/cli/storage_layout.rs
```diff
@@ -0,0 +1,201 @@
+use serde_json::json;
+
+const LIMITATION: &str = "Only entries in the compiler-reported `storage` array are compared. \
+    Namespaced (including EIP-7201) and manually computed slots are outside this check unless \
+    represented in that array; enum member changes and state-variable behavior are not checked.";
+
+forgetest_init!(accepts_semantic_append_with_json_report, |prj, cmd| {
```
mablr, line 7: Please add compiler-generated coverage for nested structs, mappings/arrays, and `--allow-renames`; the current references are mostly handcrafted and may not match real `solc` metadata shapes.

crates/forge/src/cmd/storage_layout.rs
```diff
@@ -0,0 +1,866 @@
...
+                ) {
+                    return false;
+                }
+            }
+            (None, None) => {}
+            _ => return false,
+        }
+    }
+
+    let has_structure = has_members
```
stevencartavia, line 457: Fixed-array length is ignored when rounded allocation is unchanged. Both `uint8[33]` → `uint8[34]` and `uint8[34]` → `uint8[33]` return `compatible: true` on solc 0.8.20 and 0.8.36 because both occupy 64 bytes. The shrinking case makes an existing element inaccessible. Should the exact extent be compared at every recursive fixed-array node?

PR #15957, opened 2026-07-29
feat(cast): add erc20 amount units

mattsse:
Adds an opt-in `--units <DECIMALS|auto>` interface to amount-bearing `cast erc20` reads and writes while retaining raw smallest-unit integers as the default. Explicit decimal counts avoid metadata calls; automatic mode validates the ABI response from `decimals()` and fails with actionable guidance when metadata is missing, reverting, incompatible, or unsupported. Conversions reuse Cast's existing unit primitives and reject precision loss before a transaction is constructed.

This PR was implemented with Codex assistance across investigation, code generation, tests, documentation, and PR drafting.

### DIFF
--- a/crates/cast/src/cmd/erc20.rs
+++ b/crates/cast/src/cmd/erc20.rs
@@ -36,7 +36,6 @@ use foundry_common::{
     shell,
     tempo::{TEMPO_BROWSER_GAS_BUFFER, maybe_print_fee_token, resolve_and_set_fee_token},
 };
-use foundry_config::Config;
 use foundry_wallets::{TempoAccountsWallet, WalletSigner};
 use tempo_alloy::TempoNetwork;
 use tempo_primitives::transaction::FEE_PAYER_SIGNATURE_MARKER;
@@ -118,10 +117,21 @@ fn decode_erc20_decimals(output: &[u8]) -> eyre::Result<u8> {
     Ok(output[31])
 }
 
+fn validate_erc20_decimals(decimals: u8) -> eyre::Result<u8> {
+    if Unit::new(decimals).is_none() {
+        eyre::bail!(
+            "ERC-20 decimals() returned {decimals}, but at most 77 decimals are supported; omit \
+             `--units` to use raw amounts"
+        );
+    }
+    Ok(decimals)
+}
+
 fn normalize_decimal_amount(value: &str) -> Option<String> {
     let (integer, fractional) = value.split_once('.').unwrap_or((value, ""));
-    if integer.chars().any(|character| !character.is_ascii_digit())
-        || fractional.chars().any(|character| !character.is_ascii_digit())
+    if (integer.is_empty() && fractional.is_empty())
+        || integer.bytes().any(|byte| !byte.is_ascii_digit())
+        || fractional.bytes().any(|byte| !byte.is_ascii_digit())
     {
         return None;
     }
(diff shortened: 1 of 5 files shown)


crates/cast/src/cmd/erc20.rs
```diff
@@ -75,6 +79,174 @@ where
...
+    ) -> eyre::Result<U256>
+    where
+        N: Network + RecommendedFillers,
+    {
+        let Some(units) = self.units else {
+            return U256::from_str(amount).wrap_err("invalid raw ERC-20 amount");
+        };
+        let decimals = match units {
+            Erc20Units::Auto => {
+                let provider = ProviderBuilder::<N>::from_config(config)?.build()?;
```
mablr: Auto mode needs one consistent resolved target and cannot currently work with `--curl`. Curl exits after the first provider request, so writes print `decimals()` instead of the requested transaction and reads can skip auto formatting; additionally, this metadata path resolves an ENS token before `erc20_send!` resolves it again, allowing the decimals source and transaction target to diverge. Please reject `--units auto --curl` before any request, and query metadata from the same resolved `IERC20` instance later used to construct the transaction.

crates/cast/src/cmd/erc20.rs
```diff
@@ -75,6 +79,174 @@ where
...
+                self.decimals(&token, BlockId::default(), None)
+                    .await?
+                    .expect("auto units always resolve decimals")
+            }
+            Erc20Units::Decimals(decimals) => decimals,
+        };
+
+        if let Some((_, fractional)) = amount.split_once('.')
+            && fractional.len() > decimals as usize
+            && fractional[decimals as usize..].chars().any(|character| character != '0')
```
mablr: This mixes a UTF-8 byte length with string indexing, so input such as `1.💥` with `--units 1` slices inside a character and panics. Please validate a strict ASCII decimal grammar first (also rejecting empty and dot-only input), then inspect `fractional.as_bytes()[precision..]` so malformed CLI input returns an error rather than crashing.

docs/dev/output-channels.md
```diff
@@ -106,7 +106,9 @@ Each row's status is one of:
 | `cast wallet new`      | One record per wallet: `address` (keystore) or `address\tprivate_key` (no keystore) | JSON array of `{ address, public_key, path }` (keystore) or `{ address, public_key, private_key }` (no keystore) | migrated |
 | `cast wallet sign`     | Signature                                            | JSON                                                           | migrated |
 | `cast wallet sign-auth`| Signed authorization RLP                             | JSON                                                           | migrated |
-| `cast erc20 balance`   | Balance (decimal)                                    | JSON string                                                    | migrated |
+| `cast erc20 balance`   | Smallest-unit integer, or decimal amount with `--units` | JSON string                                                 | migrated |
+| `cast erc20 allowance` | Smallest-unit integer, or decimal amount with `--units` | JSON string                                                 | migrated |
+| `cast erc20 total-supply` | Smallest-unit integer, or decimal amount with `--units` | JSON string                                              | migrated |
```
mablr: These rows are marked `migrated` and promise a smallest-unit integer, but raw allowance and total-supply still use `format_uint_exp`, which appends an exponent hint for values >= 10,000. Please either document that existing output or mark these rows `todo`; changing the established text output seems outside this feature PR.

crates/cast/src/cmd/erc20.rs
```diff
@@ -75,6 +79,174 @@ where
...
+                 precision"
+            );
+        }
+
+        let parsed = SimpleCast::parse_units(amount, decimals).wrap_err_with(|| {
+            format!("invalid ERC-20 amount `{amount}` for {decimals} decimals")
+        })?;
+        let parsed = U256::from_str(&parsed).wrap_err("ERC-20 amounts must be unsigned")?;
+        let formatted = SimpleCast::format_units(&parsed.to_string(), decimals)?;
+        if normalize_decimal_amount(amount) != normalize_decimal_amount(&formatted) {
```
mablr: Please add focused boundaries around this exactness guarantee: units 0/77/78, metadata returning 78, excess trailing zeroes, `U256::MAX`, and scaled overflow. Also include malformed ABI outputs (empty/short, trailing words, and nonzero high-byte padding). For metadata above 77, the error should recommend raw amounts only, since explicit units have the same limit.

crates/cast/src/cmd/erc20.rs
```diff
@@ -330,6 +551,13 @@ impl Erc20Subcommand {
     }
 
     pub async fn run(self) -> eyre::Result<()> {
```
stevencartavia, line 553: The new guard handles the literal `--curl` flag, but curl enabled through `eth_rpc_curl = true` or `FOUNDRY_ETH_RPC_CURL=true` still bypasses it. In those cases, the command exits successfully after printing only the primary request without querying `decimals()`. Could we check the resolved `config.eth_rpc_curl` here and add config/env coverage?

PR #15958, opened 2026-07-29
feat(forge): check storage layout compatibility

mattsse:
Adds `forge inspect <contract> storageLayout --check [<path>]` to compare compiler-reported storage layouts semantically instead of diffing serialized JSON. The checker matches existing variables, recursively compares storage types, reports slot, offset, type, and inheritance-order changes, and accepts only trailing non-overlapping additions. It reads either raw storage-layout JSON or the `storageLayout` saved by `forge clone` in `.clone.meta`, emits a structured report under `--json`, and exits nonzero for incompatible or unverifiable layouts.

EIP-7201 declarations currently produce an unverifiable result because compiler storage-layout output does not expose namespace roots; this avoids claiming safety while #15299 remains unresolved. Compiler metadata that cannot establish semantics, such as enum or user-defined value type definitions, is handled the same way.

This PR was written primarily with OpenAI Codex.

### DIFF
--- a/crates/forge/src/cmd/storage_layout.rs
+++ b/crates/forge/src/cmd/storage_layout.rs
@@ -1,0 +1,915 @@
+//! Semantic storage-layout compatibility checks.
+
+use alloy_primitives::U256;
+use eyre::{Context, Result, eyre};
+use foundry_common::shell;
+use foundry_compilers::{
+    ProjectCompileOutput,
+    artifacts::{Storage, StorageLayout, StorageType},
+};
+use serde::Serialize;
+use serde_json::Value;
+use std::{
+    collections::{BTreeMap, HashSet},
+    fs,
+    path::Path,
+    str::FromStr,
+};
+
+/// Result of comparing a reference storage layout with a current compiler layout.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct StorageLayoutCheck {
+    /// Whether compatibility was proven.
+    pub compatible: bool,
+    /// Overall outcome of the comparison.
+    pub status: CheckStatus,
+    /// Reference file used for the comparison.
+    pub reference: String,
+    /// Number of existing storage variables checked.
+    pub checked: usize,
+    /// Compatible variables appended after the reference layout.
+    pub appended: Vec<AppendedStorage>,
+    /// Compatibility errors or limitations discovered by the check.
+    pub issues: Vec<StorageLayoutIssue>,
+}
+
+impl StorageLayoutCheck {
+    /// Returns whether the current layout is compatible with the reference.
+    pub const fn is_compatible(&self) -> bool {
+        self.compatible
+    }
+}
+
+/// Overall storage-layout check outcome.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum CheckStatus {
+    Compatible,
+    Incompatible,
+    Unverifiable,
+}
+
+/// A compatible variable appended to the reference layout.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct AppendedStorage {
+    pub label: String,
+    pub slot: String,
+    pub offset: i64,
+    #[serde(rename = "type")]
+    pub storage_type: String,
+}
+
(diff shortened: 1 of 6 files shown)


crates/forge/src/cmd/storage_layout.rs
```diff
@@ -0,0 +1,915 @@
...
+
+/// Reads either a raw compiler storage layout or the `storageLayout` field of `.clone.meta`.
+pub fn read_storage_layout(path: &Path) -> Result<StorageLayout> {
+    let input = fs::read_to_string(path).wrap_err_with(|| {
+        format!("failed to read storage layout reference `{}`", path.display())
+    })?;
+    let value: Value = serde_json::from_str(&input).wrap_err_with(|| {
+        format!("failed to parse storage layout reference `{}`", path.display())
+    })?;
+    let value = value.get("storageLayout").unwrap_or(&value).clone();
```
mablr, line 135: **Blocker:** This discards whether the reference contained EIP-7201 storage, while the later namespace check only inspects the current sources. If a reference namespace is removed, both ordinary `storageLayout`s can still match and this reports `compatible: true`. Please retain reference-side namespace provenance (`present` / `absent` / `unknown`) and fail closed for `present` or `unknown`, or explicitly scope the command to compiler-reported storage rather than claiming namespaced storage is conservatively rejected. A regression test should remove a namespace that existed in the reference.

crates/forge/src/cmd/storage_layout.rs
```diff
@@ -0,0 +1,915 @@
...
+    })
+}
+
+/// Returns whether the target source or one of its imports declares EIP-7201 storage.
+pub fn contains_namespaced_storage(
+    output: &ProjectCompileOutput,
+    target_path: &Path,
+) -> Result<bool> {
+    static MARKER: &str = "@custom:storage-location";
+    let paths = std::iter::once(target_path).chain(output.graph().imports(target_path));
```
mablr, line 150: This scans every transitive import, so an unrelated imported contract/library with an ERC-7201 annotation makes the selected target unverifiable. Please scope detection to the selected contract and its linearized bases (with a conservative fallback if needed), rather than treating any marker in the import graph as target storage. The current test’s file-level struct plus empty contract actually codifies this false positive.

crates/forge/tests/cli/cmd.rs
```diff
@@ -1483,6 +1483,131 @@ forgetest!(cannot_inspect_linearization_non_solidity_target, |prj, cmd| {
...
+        .args(["inspect", "Layout", "storageLayout", "--check"])
+        .assert_success()
+        .stdout_eq(str![[r#"
+Storage layout is compatible.
+Checked 1 existing variable(s); found 1 compatible append(s).
+
+"#]]);
+});
+
+forgetest!(storage_layout_check_reports_semantic_type_changes_as_json, |prj, cmd| {
```
mablr, line 1526: The recursive comparator has substantial mapping/array/struct logic, but the CLI coverage only exercises packed scalar append and a scalar type change. Please add at least one compiler-generated nested layout covering mappings, fixed/dynamic arrays, and structs, then mutate a key/value/base/member or array length. That would verify the code against real solc metadata shapes rather than only handcrafted unit values.

crates/forge/src/cmd/storage_layout.rs
stevencartavia, line 135: The parsed reference retains no indication that it used EIP-7201 storage, while namespace detection scans only the current sources. In a reproduction, removing the annotation and changing a namespaced member’s type left both compiler layouts empty, so `--check` returned `compatible: true`. Could we retain reference-side namespace provenance and treat references without it as unknown/unverifiable?

crates/forge/src/cmd/storage_layout.rs
```diff
@@ -0,0 +1,915 @@
...
+        if should_compare_label(expected_kind, expected) && expected.label != actual.label {
+            self.type_changed(
+                path,
+                "storage type changed",
+                expected.label.as_str(),
+                actual.label.as_str(),
+            );
+        }
+
+        if matches!(expected_kind, "enum" | "user_defined_value_type") {
```
stevencartavia, line 463: Stored internal function pointers have bytecode-specific targets that can change while their storage metadata remains identical. A reproduction had V1 storing `0x100` and V2 using `0x129`; the checker returned compatible, but the persisted V1 pointer reverted with `InvalidJump` after the upgrade. Would it make sense to classify `t_function_internal_*` as unverifiable?

PR #15960, opened 2026-07-29
feat(cast): add ERC-20 amount units

mattsse:
`cast erc20` currently treats every token amount as a raw integer, which makes human-scale input and output awkward across tokens with different decimals. This adds opt-in `--units raw|auto|<DECIMALS>` handling to balance, allowance, total-supply, transfer, approve, mint, and burn while preserving raw behavior by default. Automatic mode queries `decimals()` at the selected block, accepts ABI-compatible integer metadata in the ERC-20 range, and fails clearly for missing, reverting, malformed, or out-of-range metadata; explicit counts bypass metadata. Amounts that would lose nonzero precision are rejected, and JSON output remains a clean string payload.

This PR was implemented with Codex, including code, tests, and documentation.

### DIFF
--- a/.changelog/cast-erc20-decimal-units.md
+++ b/.changelog/cast-erc20-decimal-units.md
@@ -1,0 +1,5 @@
+---
+cast: minor
+---
+
+Added opt-in decimal-aware amounts to `cast erc20` with explicit and automatic unit modes.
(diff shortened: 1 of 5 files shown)


mablr (commented):
Closing as duplicate of #15957

PR #15961, opened 2026-07-29
test(anvil): cover Tempo request execution

mablr:
Adds focused regression coverage following #15857 for two Tempo request-execution boundaries that were not previously exercised end to end.

The fork coverage creates observable state divergence between a Tempo source node and its local fork, then verifies that `eth_call`, `eth_estimateGas`, `eth_createAccessList`, and `eth_simulateV1` execute batched Tempo requests against local fork state at `latest`.

The keychain coverage provisions a constrained access key, uses the returned gas estimate as the exact gas limit of a real V2 keychain-signed transaction, verifies successful mining, and confirms that over-limit requests remain rejected.

This is test-only and has no release-note impact, so a maintainer should apply `L-ignore`.

### DIFF
--- a/crates/anvil/tests/it/tempo.rs
+++ b/crates/anvil/tests/it/tempo.rs
@@ -64,7 +64,9 @@ use tempo_precompiles::{
 };
 use tempo_primitives::{
     AASigned, TempoHeader, TempoSignature, TempoTransaction,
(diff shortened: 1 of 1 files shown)


mattsse (approved):
lgtm

stevencartavia (approved):

PR #15962, opened 2026-07-29
fix(mutation): filter unsigned negation

mablr:
Mutation testing currently generates unary-negation replacements for unsigned operands even though Solidity rejects them during compilation. This uses resolved HIR operand types to omit those mutants while retaining negation for signed values, unresolved types, and user-defined value types that may overload the operator.

Amp was used to review the implementation and strengthen regression coverage for overloaded user-defined value types.

### DIFF
--- a/crates/forge/src/mutation/type_analysis.rs
+++ b/crates/forge/src/mutation/type_analysis.rs
@@ -1,9 +1,10 @@
-//! Type-aware filtering for binary operator mutations.
+//! Type-aware filtering for operator mutations.
 //!
-//! This module records replacements that have the same result as the original comparison for
-//! every value in an operand's type range. It deliberately does not filter replacements merely
-//! because they are tautological. For example, `uint256 x == 0` and `x <= 0` are equivalent, but
-//! `x < 0` is retained because changing the original condition to `false` can expose missing tests.
+//! This module records replacements that are invalid or have the same result as the original
+//! comparison for every value in an operand's type range. It deliberately does not filter
+//! replacements merely because they are tautological. For example, `uint256 x == 0` and `x <= 0`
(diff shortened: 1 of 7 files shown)


mattsse (dismissed):
lgtm

crates/forge/src/mutation/type_analysis.rs
```diff
@@ -105,9 +117,30 @@ impl<'hir> Visit<'hir> for EquivalentMutationCollector<'hir> {
...
+        let ExprKind::Unary(op, _) = &expr.kind else { return };
+        let replacement = match op.kind {
+            UnOpKind::PreInc => UnOpKind::PostInc,
+            UnOpKind::PostInc => UnOpKind::PreInc,
+            UnOpKind::PreDec => UnOpKind::PostDec,
+            UnOpKind::PostDec => UnOpKind::PreDec,
+            _ => return,
+        };
+        let Some(span) = self.local_span(expr.span) else { return };
+        self.mutations.insert(EquivalentMutation::unary(span, replacement));
```
stevencartavia: Standalone `x++` and `++x` do not always compile to equivalent EVM behavior. With solc 0.8.26 and optimization enabled, otherwise identical contracts used 80 versus 75 gas, with a supplied-gas boundary where `x++` ran out of gas while `++x` succeeded. Applying this exact mutation made both distinguishing tests fail, but the PR-head mutation run omitted `++x`. Could we retain these swaps unless their compiled behavior is proven equivalent?

mablr: Good catch, I retargeted the PR completely

figtracer (approved):

stevencartavia (approved):

PR #15963, opened 2026-07-29
fix(anvil): ignore empty state diffs

stevencartavia:
An empty `stateDiff` override performs no storage writes, but `eth_simulateV1` currently materializes an absent account and changes the simulated state root and derived block hash. Ignore otherwise-empty state diffs while preserving full storage replacements, non-empty diffs, and override conflict validation, and extend the existing root-parity regression.


mablr (approved):

mattsse (approved):
lgtm

PR #15964, opened 2026-07-29
fix(ci): harden stable version preparation

stevencartavia:
Rejects changelog release plans containing warnings. The pinned changelogs CLI can report an unknown package while still returning a complete plan, then delete that package’s fragment without including its note in the generated changelog.

This also reruns stable PR reconciliation when its workflow or preparation script changes and preserves captured subprocess diagnostics when preparation fails.

Part of OSS-592


mattsse (approved):
lgtm

mablr (approved):

figtracer (approved):

PR #15965, opened 2026-07-29
fix(anvil): preserve simulated blob fee caps

stevencartavia:
This preserves the requested `maxFeePerBlobGas` during `eth_simulateV1` execution. Non-validating zero-cap blob calls now skip only the blob fee-cap check while retaining the real block blob fee for `BLOBBASEFEE` and fee accounting. Explicit caps continue to use normal validation and affordability checks.


mattsse (approved):
lgtm

mablr (approved):

figtracer (approved):

Issue #7076, opened 2024-02-10
feat: add cast create2 initcodehash


0xClandestine:
### Component

Cast

### Describe the feature you would like

Something like this would be nice for computing CREATE2 initcode hashes on the fly:

```sh
cast create2 initcodehash <contract> <optional_args_selector> <args>
```

Output should equal:

```solidity
keccak256(abi.encodePacked(type(c).creationCode, abi.encode(params))
```

### Additional context

I've currently been using this helper in my `package.json`:

```sh
cast keccak $(cast concat-hex \"$(forge inspect --silent Witness bytecode)\" \"$(cast abi-encode 'abi.encode(address)' \"$OWNER_ADDRESS\")\")
```

figtracer:
retested on current master: `cast create2 init-code-hash` accepts constructor arguments and returns the hash. closing as completed.

PR #15969, opened 2026-07-30
feat(cast): add CREATE2 init code hash

figtracer:
## Motivation

`cast create2` accepts raw init code or an existing hash, but computing that hash from a local contract and constructor arguments currently requires a multi-command shell pipeline. This addresses #7076.

## Solution

This adds `cast create2 init-code-hash <contract> [args]...`, which compiles the target using the project configuration, parses constructor arguments against its ABI, appends their canonical encoding to the creation bytecode, and prints the resulting Keccak-256 hash. Constructor parsing is shared with `forge create` so both commands retain identical validation and encoding behavior. The existing `cast create2` interface remains unchanged, and `initcodehash` is available as an alias.

This PR was written with Codex, which assisted with issue investigation, implementation, tests, live validation, and PR text.

### DIFF
--- a/crates/cast/src/cmd/create2.rs
+++ b/crates/cast/src/cmd/create2.rs
@@ -3,6 +3,7 @@ use alloy_primitives::{Address, B256, U256, hex, keccak256};
 use clap::{Args, Parser, Subcommand};
 use eyre::{Result, WrapErr};
 use foundry_cli::{
+    json::print_scalar,
     opts::BuildOpts,
     utils::{LoadConfig, find_contract_artifacts, parse_constructor_args},
 };
@@ -30,7 +31,7 @@ struct InitCodeHashArgs {
     contract: ContractInfo,
 
     /// The constructor arguments.
-    #[arg(value_name = "ARGS", allow_hyphen_values = true)]
+    #[arg(value_name = "ARGS", allow_negative_numbers = true)]
     constructor_args: Vec<String>,
 
     #[command(flatten)]
@@ -60,6 +61,9 @@ impl InitCodeHashArgs {
         let Some(bytecode) = bin.object.into_bytes() else {
             eyre::bail!("contract contains unlinked libraries");
         };
+        if bytecode.is_empty() {
+            eyre::bail!("no bytecode found in bin object for {}", self.contract.name);
+        }
 
         let mut init_code = bytecode.to_vec();
         if let Some(constructor) = &abi.constructor {
@@ -69,7 +73,7 @@ impl InitCodeHashArgs {
             eyre::bail!("contract does not have a constructor");
         }
 
-        sh_println!("{}", keccak256(init_code))?;
+        print_scalar(keccak256(init_code))?;
         Ok(())
     }
 }
--- a/crates/cast/src/cmd/create2.rs
+++ b/crates/cast/src/cmd/create2.rs
@@ -58,7 +58,7 @@ impl InitCodeHashArgs {
         let output = compile::compile_target(&target_path, &project, true)?;
         let (abi, bin, _) = find_contract_artifacts(output, &target_path, &self.contract.name)?;
         let Some(bytecode) = bin.object.into_bytes() else {
-            eyre::bail!("contract contains unlinked libraries")
+            eyre::bail!("contract contains unlinked libraries");
         };
 
         let mut init_code = bytecode.to_vec();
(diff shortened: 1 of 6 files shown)


crates/cast/src/cmd/create2.rs
```diff
@@ -1,16 +1,86 @@
...
+        let project = config.project()?;
+        let target_path = if let Some(path) = &self.contract.path {
+            canonicalize(project.root().join(path))?
+        } else {
+            project.find_contract_path(&self.contract.name)?
+        };
+
+        let output = compile::compile_target(&target_path, &project, true)?;
+        let (abi, bin, _) = find_contract_artifacts(output, &target_path, &self.contract.name)?;
+        let Some(bytecode) = bin.object.into_bytes() else {
```
mablr, line 61: This still accepts empty bytecode from abstract contracts and returns `keccak256("")`. Please reject `bytecode.is_empty()` before appending constructor args, matching `forge create`.

crates/cast/src/cmd/create2.rs
```diff
@@ -1,16 +1,86 @@
...
+
+foundry_config::impl_figment_convert!(InitCodeHashArgs, build);
+
+#[derive(Clone, Debug, Args)]
+struct InitCodeHashArgs {
+    /// The contract identifier in the form `<path>:<contractname>`.
+    contract: ContractInfo,
+
+    /// The constructor arguments.
+    #[arg(value_name = "ARGS", allow_hyphen_values = true)]
```
mablr: `allow_hyphen_values` makes constructor args swallow later options (for example, `-5 --root ...`). Can this use `allow_negative_numbers = true` so build flags remain parseable?

crates/cast/src/cmd/create2.rs
```diff
@@ -1,16 +1,86 @@
...
+
+        let mut init_code = bytecode.to_vec();
+        if let Some(constructor) = &abi.constructor {
+            let params = parse_constructor_args(constructor, &self.constructor_args)?;
+            init_code.extend(constructor.abi_encode_input(&params)?);
+        } else if !self.constructor_args.is_empty() {
+            eyre::bail!("contract does not have a constructor");
+        }
+
+        sh_println!("{}", keccak256(init_code))?;
```
mablr: Please use `foundry_cli::json::print_scalar` here so global `--json` returns Cast's standard JSON envelope instead of a bare hash.

crates/cast/src/cmd/create2.rs
```diff
@@ -89,8 +163,16 @@ pub struct Create2Output {
...
+    pub fn execute(self) -> Result<()> {
+        if let Some(command) = &self.command {
+            return command.run();
+        }
+        self.run().map(drop)
+    }
+
     pub fn run(self) -> Result<Create2Output> {
         let Self {
+            command: _,
```
mablr, line 181: Could we avoid silently discarding `command` here? `subcommand_negates_reqs` makes `Create2Args::parse_from(["cast", "init-code-hash", ...])` valid, but a direct caller of this still-public `run()` then reaches the `unreachable!()` below because neither legacy init-code argument is set. Making the legacy runner private, or returning an actionable error when `command.is_some()`, would keep a valid parsed state from panicking. A small regression test for this path would also be useful.

mablr (commented):
LGTM, just one suggestion.

crates/cast/src/cmd/create2.rs
stevencartavia: +1 from my pass

mablr (dismissed):

crates/cast/src/cmd/create2.rs
```diff
@@ -89,8 +163,16 @@ pub struct Create2Output {
...
 impl Create2Args {
-    pub fn run(self) -> Result<Create2Output> {
+    pub fn execute(self) -> Result<()> {
+        if let Some(command) = &self.command {
+            return command.run();
+        }
+        self.run().map(drop)
+    }
+
+    fn run(self) -> Result<Create2Output> {
```
stevencartavia: Would it make sense to use the other option suggested in the earlier thread: keep `run()` public and return an actionable error when `command.is_some()`? The private alternative in the latest commit appears to break existing Cast library callers—the same external consumer compiles against the base revision but fails here because `run()` is private—and `execute()` returns `()` rather than `Create2Output`.

stevencartavia (approved):

mablr (approved):

PR #15971, opened 2026-07-30
fix(mutation): prune invalid comparisons

mablr:
## Description
This is stacked on [#15962](https://github.com/foundry-rs/foundry/pull/15962) and builds on its HIR type-analysis pass. Binary mutation currently treats comparison and logical operators as one replacement pool, producing mutants that Solidity cannot type-check, such as replacing a numeric comparison with `&&` or `||`. This change prunes replacements proven invalid for numeric, boolean, and function operands before scheduling them, while retaining candidates when type information is unavailable.

### Results

Measured on Solady’s `ERC20Votes` with only binary mutations enabled, eight mutation workers, and three runs per version. Mutation-result caches were cleared between runs.

| Metric | `master` | `current` | Delta |
|---|---:|---:|---:|
| Total mutants | 130 | 110 | -15.4% |
| Invalid | 24 | 6 | -75.0% |
| Killed / survived | 86 / 15 | 86 / 15 | unchanged |
| Wall time | 42.882 ± 1.154s | 39.212 ± 1.854s | -8.5% |

The valid candidate reduction is deterministic.

### DIFF
--- a/crates/forge/src/mutation/type_analysis.rs
+++ b/crates/forge/src/mutation/type_analysis.rs
@@ -1,7 +1,7 @@
 //! Type-aware filtering for operator mutations.
 //!
-//! This module records replacements that are invalid or have the same result as the original
-//! comparison for every value in an operand's type range. It deliberately does not filter
+//! This module records replacements that are either type-invalid or have the same result as the
+//! original comparison for every value in an operand's type range. It deliberately does not filter
 //! replacements merely because they are tautological. For example, `uint256 x == 0` and `x <= 0`
 //! are equivalent, but `x < 0` is retained because changing the original condition to `false` can
 //! expose missing tests.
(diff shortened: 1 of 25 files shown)


mattsse (dismissed):
lgtm

mablr (commented):
Don't merge before #15962 to preserve history.

figtracer (approved):

crates/forge/src/mutation/type_analysis.rs
```diff
@@ -132,6 +132,29 @@ impl MutationExclusionCollector<'_> {
         original: BinOpKind,
         right: &hir::Expr<'_>,
     ) {
+        let Some(span) = self.local_span(expr.span) else { return };
+
+        for candidate in [
```
stevencartavia, line 137: Non-blocking: This PR changes the generated mutant set, but the cache salt remains `mutant-set-v4`, so the head can load and execute invalid `&&`/`||` mutants from stale v4 caches. Could it be bumped to `mutant-set-v5`?

stevencartavia (approved):

crates/forge/src/mutation/type_analysis.rs
mablr: Already bumped in base branch (just merged)

PR #15972, opened 2026-07-30
fix(traces): decode P256VERIFY precompile calls

HwanHeo:
Widen the zero-prefix check in `is_known_precompile` from 19 to 18 bytes. The 19-byte check only admits addresses up to 0xff, so P256VERIFY at 0x...0100 never matched it and the decoder added in #13094 was unreachable. Traces rendered the call as an unrelated 4-byte selector, as captured in the blessed snapshots. The exact-address match still rejects every other address, which is pinned by a new unit test. Only trace rendering changes. Execution is unaffected. Update both snapshots accordingly.

This PR was written primarily by Claude Code. I reviewed and verified it locally: ran the forge and cast tests against the updated snapshots (including replaying the real Base transaction), and re-introduced the bug to confirm the tests catch the regression.


mablr (dismissed):
Makes sense

grandizzy (commented):
Looks good, ty! Left one nit inline, please check.

crates/evm/traces/src/decoder/precompiles.rs
```diff
@@ -61,7 +61,8 @@ pub(crate) fn is_known_precompile(
     tempo_hardfork: Option<TempoHardfork>,
 ) -> bool {
     // Standard EVM precompiles (all chains).
-    let is_standard = address[..19].iter().all(|&x| x == 0)
+    // An 18-byte zero prefix, as `P256_VERIFY` (0x..0100) occupies the two lowest bytes.
+    let is_standard = address[..18].iter().all(|&x| x == 0)
```
grandizzy, line 76: This now recognizes `0x100` as P256 regardless of whether the precompile is active. Before Osaka—or before chain-specific activation such as BSC Haber—`0x100` can execute ordinary bytecode, but `decode_function` calls this address-based decoder before contract ABI decoding and does not consult `trace.maybe_precompile`. For example, a Cancun test that etches code at `address(0x100)` will execute that code but render the call as `PRECOMPILES::p256Verify(...)`.

Could we at least skip precompile decoding when `trace.maybe_precompile == Some(false)`?

HwanHeo: Good catch, thanks! Done — `decode` now returns early when `trace.maybe_precompile == Some(false)`, pinned with a unit test.

crates/evm/traces/src/decoder/precompiles.rs
```diff
@@ -72,7 +72,8 @@ pub(crate) fn is_known_precompile(
     let _ = monad_hardfork;
 
     // Standard EVM precompiles (all chains).
-    let is_standard = address[..19].iter().all(|&x| x == 0)
+    // An 18-byte zero prefix, as `P256_VERIFY` (0x..0100) occupies the two lowest bytes.
```
figtracer, line 75: this still skips abi identification for ordinary code at `0x100` before p256 is active: `identify_addresses` calls `is_known_precompile` without checking `node.is_precompile()`/`maybe_precompile`, so the new decode guard returns `None` but no contract abi was registered either. can we make that path account for whether the call actually executed as a precompile?

crates/evm/traces/src/decoder/precompiles.rs
```diff
@@ -165,6 +166,12 @@ pub(super) fn decode(
     tempo_hardfork: Option<TempoHardfork>,
     monad_hardfork: Option<MonadHardfork>,
 ) -> Option<DecodedCallTrace> {
+    // The tracer determined the call did not execute as a precompile, e.g. code etched at the
+    // address of a precompile that is not active on the current hardfork.
+    if trace.maybe_precompile == Some(false) {
```
stevencartavia: Suggestion: could we add a production-path Cancun test with ordinary code at `0x100`? Real Forge/Cast traces still decode that call as P256 because this guard does not fire there. This is niche and trace-only, but produces misleading output in `vm.etch`/state-override workflows.

mablr (approved):
LGTM

figtracer (approved):

stevencartavia (approved):

PR #15973, opened 2026-07-30
fix(config): handle Windows remapping separators

figtracer:
#15929 compared nested remapping paths as raw strings for inequality but used `Path` semantics for ancestry. On Windows, nested-config sanitization changes backslashes to forward slashes while auto-detected paths retain native separators, so the same directory was misclassified as a refinement and emitted as a duplicate contextual remapping.

This change compares both paths semantically and canonicalizes preserved context boundaries to the platform separator before slash-normalized rendering. The focused config and Forge regressions retain coverage for the serialized remapping behavior. Together, these changes fix the 20 failures in the [Windows test job](https://github.com/foundry-rs/foundry/actions/runs/30529082800/job/90826907468).

> This PR's implementation, tests, description, and review follow-up were written primarily with Codex.

### DIFF
--- a/crates/config/src/providers/remappings.rs
+++ b/crates/config/src/providers/remappings.rs
@@ -277,7 +277,8 @@ impl RemappingsProvider<'_> {
                     && auto_detected_remappings.iter().any(|auto| {
                         auto.context == r.context
                             && auto.name == r.name
-                            && is_strict_descendant(Path::new(&r.path), Path::new(&auto.path))
+                            && Path::new(&r.path) != Path::new(&auto.path)
+                            && Path::new(&r.path).starts_with(&auto.path)
                     })
                 {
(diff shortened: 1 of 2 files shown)


crates/config/src/providers/remappings.rs
```diff
@@ -410,6 +409,10 @@ impl RemappingsProvider<'_> {
     }
 }
 
+fn is_strict_descendant(path: &Path, parent: &Path) -> bool {
+    path != parent && path.starts_with(parent)
+}
```
mablr: Can be inlined, associated unit test is not very useful btw.

mablr (commented):
sgtm

grandizzy (approved):
lgmt, ty!

mablr (approved):

PR #15975, opened 2026-07-30
fix(anvil): `eth_simulateV1` apply canonical block transitions

mablr:
## Description

Closes OSS-599.

Applies canonical Cancun and Prague block transitions to complete Ethereum blocks produced by `eth_simulateV1`, pending-block construction, and mining. This includes ordered EIP-2935/EIP-4788 pre-block calls and EIP-6110/EIP-7002/EIP-7251 request collection, while preserving fork-replay inputs and skipping post-block transitions for transaction-prefix replay.

Tempo and Optimism remain explicitly excluded. Mining failure recovery remains scoped to OSS-601.


crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -3778,21 +3892,30 @@ where
...
+        parent_beacon_block_root: Option<B256>,
         transactions: &[HistoricalReplayTransaction],
         inspector_tx_config: &InspectorTxConfig,
     ) -> Result<ExecutedHistoricalReplay>
     where
         DB: StateDB<Error = DatabaseError>,
     {
         let inspector = self.build_mining_inspector();
+        let ethereum_transitions = self.ethereum_block_transitions(
+            parent_beacon_block_root,
```
grandizzy, line 3930: The transition hardfork needs to come from the source block rather than `self.hardfork()`. Transaction-hash forks normally infer the backend hardfork from the source block’s parent, so the first Cancun block skips EIP-4788 and the first Prague block skips EIP-2935. An explicit newer `--hardfork` can cause the inverse problem; replaying a pre-Cancun source as Prague now fails because the source header has no `parentBeaconBlockRoot`. Could we resolve the source hardfork from its chain ID and timestamp, carry it in `PreparedForkTransactionReplay`, and use it here? A fork-boundary regression test would cover this.

grandizzy (changes requested):
Looking good overall—please check the finding inline.

crates/anvil/src/eth/backend/mem/mod.rs
mablr: https://github.com/foundry-rs/foundry/pull/15975/commits/d6c252b9d24a57fd42c2cf125d8496241f15fff9

grandizzy (approved):
lgtm

emmajam (approved):

PR #15976, opened 2026-07-30
feat(symex): add mapping SSTORE hooks

### LINKED ISSUE #15950: feat(symex): add SLOAD/SSTORE hook cheatcodes

mablr:
## Description

Builds on #15954 and closes #15950, including the [ERC20 balance-accounting use case](https://github.com/foundry-rs/foundry/issues/15950#issuecomment-5117020127).

This adds `vm.registerMappingSstoreHook`, allowing concrete and symbolic Forge tests to register post-store callbacks for one mapping root in a target's effective storage context. Callbacks receive the computed slot, terminal mapping root, root-to-leaf keys, and old and new values, making it possible to maintain mapping-specific ghost state without reacting to unrelated mappings such as ERC20 allowances.

Mapping callbacks inherit the raw hooks' revert-aware lifecycle, delegatecall storage attribution, callback isolation, and recursion suppression. Raw and mapping SSTORE hooks are mutually exclusive for the same target, while multiple mapping roots can coexist and re-registering the same root replaces its callback.

Mapping provenance is intentionally exact and execution-local. A callback requires a complete 64-byte Keccak chain observed after the target's latest mapping-hook registration in the current top-level execution. Resolution follows the chain to its terminal root; offsets, incomplete chains, previously precomputed slots, and source-level storage layouts are outside this API's scope.

The symbolic coverage includes the motivating ERC20 property by maintaining a ghost sum across mint, transfer, and burn operations, excluding allowance writes, and checking rollback when token writes later revert.

Developed with assistance from Amp.


grandizzy (dismissed):
lgtm

emmajam (dismissed):

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -584,7 +613,24 @@ impl SymbolicExecutor {
                 let hook = (!state.storage_hook_active)
                     .then(|| state.storage_store_hooks.get(&state.storage_address).copied())
                     .flatten();
-                let old_value = if hook.is_some() {
+                let mapping = (!state.storage_hook_active)
+                    .then(|| {
+                        key.storage_mapping_provenance_observed_with(&mut self.cx, |hash| {
```
figtracer: **[p1] do not silently skip constraint-equivalent mapping slots**

this lookup only follows the exact symbolic expression used as the hashmap key. for example, after computing `h = keccak256(abi.encode(key, 0))`, `sstore(sub(add(h, delta), delta), value)` writes exactly `h` for every evm input.

concrete execution resolves the numeric slot and invokes the hook, but symbolic execution skips it. i reproduced a check asserting `calls == 0` being reported as a symbolic pass even though the same concrete test fails for every input; asserting `calls == 1` instead produces a counterexample that does not replay.

please normalize or prove equivalent keys, or conservatively mark an unresolved hash-derived store incomplete rather than treating it as unhooked, and add this concrete/symbolic parity regression.

testdata/utils/Vm.sol
```diff
@@ -450,6 +450,7 @@ interface Vm {
     function readLink(string calldata linkPath) external view returns (string memory targetPath);
     function record() external;
     function recordLogs() external;
+    function registerMappingSstoreHook(address target, bytes32 rootSlot, bytes4 callback) external;
```
figtracer, line 453: **[p1] update the snapshot changed by this fixture interface**

adding this declaration changes the metadata in the imported pause-tracing fixture. the required ci job failed all three retries, and the following reproduces it locally:

```console
cargo +1.96.0 test -p forge --test cli pause_tracing -- --nocapture
```

the snapshot expects `558957` while the new output is `558945`. please update the dependent snapshot in `crates/forge/tests/cli/test_cmd/mod.rs`.

crates/evm/evm/src/inspectors/stack.rs
```diff
@@ -1070,6 +1070,9 @@ impl<FEN: FoundryEvmNetwork> InspectorStackRefMut<'_, FEN> {
     /// Invoked at the beginning of a new top-level (0 depth) frame.
     fn top_level_frame_start(&mut self, ecx: &mut FoundryContextFor<'_, FEN>) {
         self.locally_created_accounts.clear();
+        if let Some(cheatcodes) = &mut self.cheatcodes {
+            cheatcodes.clear_storage_hook_mapping_slots();
```
figtracer, line 1074: **[p2] clear provenance before it is cloned into the next run**

this start-of-frame clear occurs only after the executor has cloned the persisted inspector. a setup that registers a hook and then performs many mapping hashes therefore leaves the full provenance map in the fuzz baseline, and every worker clones it before discarding it here; committing invariant steps have the same problem.

with 20,000 retained hashes, a 10,000-run no-op fuzz suite grew from 45 ms to 359 ms.

nothing consumes this top-level-local data after the root finishes, so please also clear it in `top_level_frame_end` on both success and revert, keeping this clear as a defensive fallback.

crates/cheatcodes/src/inspector.rs
```diff
@@ -1983,11 +2045,24 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>> for Cheatcode
             );
         }
 
-        // `startMappingRecording`: record SSTORE and KECCAK256.
+        // `startMappingRecording`: record SSTORE.
```
figtracer: **[p2] isolate mapping recording from synthetic callbacks**

`startmappingrecording` remains active throughout the synthetic callback subtree, so a callback that updates a ghost mapping adds those instrumentation-only slots to the user-visible mapping recording.

i reproduced a target mapping write whose callback writes `ghost[1]`; `getmappinglength` on the callback contract then returns `1`.

`mapping_slots` should be saved, disabled, and restored with the other storage-hook inspector state. the raw hook path already has this isolation hole, and the new mapping callback api inherits it.

crates/forge/tests/cli/test_cmd/symbolic_cheatcodes.rs
```diff
@@ -5172,3 +5172,516 @@ contract StorageHookFuzzGuidanceTest is Test {
     assert!(showmap_files > 0, "no showmap files were produced");
     assert!(saw_target_coverage, "showmap did not contain target coverage");
 });
+
+forgetest_init!(symbolic_mapping_storage_hooks, |prj, cmd| {
+    skip_unless_z3!("symbolic_mapping_storage_hooks");
```
figtracer, line 5177: **[p2] make the new symbolic behavior execute in required ci**

this guard returns before the added fixture runs because the required ci workflow does not install `z3`.

the ci entry completed in 0.162 s, while the real test took 48.164 s locally, so the new symbolic dispatch and erc20 accounting paths currently receive no runtime coverage in required ci.

please add a focused `z3`-backed job or solver-independent executor regressions for mapping dispatch and provenance parity.

crates/cheatcodes/spec/src/vm.rs
```diff
@@ -448,6 +448,23 @@ interface Vm {
...
 
+    /// Registers a callback after exact mapping-element SSTOREs rooted at `rootSlot` in `target`'s effective storage account.
+    ///
+    /// The callback signature is `function(address account, bytes32 computedSlot, bytes32 rootSlot,
+    /// bytes32[] keys, bytes32 oldValue, bytes32 newValue) external`; keys are raw words in
+    /// root-to-leaf order. Only complete 64-byte Keccak chains observed after the latest mapping
+    /// hook registration for `target` in the current top-level execution match; provenance is
+    /// cleared between top-level executions. Resolution follows the complete chain to its terminal
+    /// root and ignores registered intermediate hashes. Offsets, incomplete or unknown chains,
+    /// earlier precomputation, and source layouts do not match. Registration persists across
```
figtracer: **[p2] qualify which precomputation is rejected**

same-top-level precomputation after registration is intentionally accepted: `checkordinarycallprovenance` computes the hash in one call and stores it in a later call, and both concrete and symbolic hooks dispatch.

only hashes computed before the latest registration or in an earlier top-level execution are rejected. saying that precomputation does not match without that qualifier contradicts the implementation and the added test.

crates/cheatcodes/spec/src/vm.rs
```diff
@@ -448,6 +448,23 @@ interface Vm {
...
+    /// reverts and replaces the same target/root callback; callback state rolls back with its
+    /// enclosing context, callback reverts propagate, and hooks are suppressed throughout callback
+    /// subtrees. The callback must authenticate `msg.sender == address(vm)` to prevent external
+    /// spoofing. Raw and mapping SSTORE hooks conflict per target, while multiple mapping roots may
+    /// be registered.
+    #[cheatcode(group = Evm, safety = Unsafe, status = Experimental)]
+    function registerMappingSstoreHook(address target, bytes32 rootSlot, bytes4 callback) external;
```
figtracer, line 467: **[p2] document which contract receives the callback**

the implementation stores the caller of `registermappingsstorehook` as `callback_target`, so a helper that performs registration receives the selector rather than the test contract or the storage target.

because this api accepts only a selector and no callback address, this receiver rule is important public behavior and should be stated here.

crates/forge/tests/cli/test_cmd/symbolic_cheatcodes.rs
mablr: Agreed that required CI currently skips this test. This is also true for the existing Z3-dependent symbolic suite: there is no required solver-backed CI lane today. Adding a mapping-hooks-only workflow would duplicate CI infrastructure for one feature, so I’d prefer handling a general curated symbolic/Z3 test lane separately, wdyt? The regression remains covered in solver-enabled environments.

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -1,6 +1,46 @@
...
+                    |candidate| {
+                        state
+                            .mapping_hook_keccak_preimages
+                            .get(&(account, candidate.clone()))
+                            .cloned()
+                    },
+                ));
+            }
+        }
+        Ok(None)
```
stevencartavia: The previous fix handles slots equal to an observed hash in every model allowed by the constraints at SSTORE, but misses slots that equal it in only some models. With a mapping hook registered, `sstore(add(hash, and(control, 1)), value)` must invoke the callback for even `control` and skip it for odd `control`.

I reproduced both concrete outcomes. Symbolic execution works when parity is constrained before SSTORE, but when the branch occurs afterward, the inverted property asserting no callback for even `control` incorrectly passes.

Could we fork into an equality-constrained callback path and an inequality-constrained non-callback path when `slot == hash` and `slot != hash` are both feasible?

crates/forge/tests/cli/test_cmd/symbolic_cheatcodes.rs
figtracer: makes sense, let's keep it this way

crates/evm/symbolic/src/executor/opcodes.rs
mablr: https://github.com/foundry-rs/foundry/pull/15976/commits/48e364f7ab61311e5d157ce73c9a1c92fad83077

figtracer (dismissed):

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -265,7 +342,18 @@ impl SymbolicExecutor {
...
+                        let hash = keccak_word(&mut self.cx, bytes.clone());
+                        let has_mapping_hook = state
+                            .mapping_storage_store_hooks
+                            .keys()
+                            .any(|(address, _)| *address == state.storage_address);
+                        if has_mapping_hook && !state.storage_hook_active && size == 64 {
+                            state
+                                .mapping_hook_keccak_preimages
+                                .entry((state.storage_address, hash.clone()))
+                                .or_insert_with(|| bytes.into());
```
stevencartavia, line 446: The symbolic-size `KECCAK256` arm does not populate `mapping_hook_keccak_preimages` when `size == 64` is feasible, unlike the concrete-size arm. A later `SSTORE` can therefore miss the mapping provenance and skip a registered mapping callback, allowing the symbolic property to report `PASS` even though the concrete 64-byte case fails.

Could we conservatively report `Incomplete` when `size == 64` is feasible, a mapping SSTORE hook is registered for the current storage address, and `!storage_hook_active`? Alternatively, split on `size == 64` and use the concrete 64-byte hash/provenance path on that branch.

mablr: https://github.com/foundry-rs/foundry/pull/15976/commits/1137a4a2bedc0d1fd87eb677dbbd938c596dc302

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -1,6 +1,65 @@
...
+                hash.storage_mapping_provenance_observed_with(&mut self.cx, |candidate| {
+                    state.mapping_hook_keccak_preimages.get(&(account, candidate.clone())).cloned()
+                })
+            else {
+                continue;
+            };
+            if !state.mapping_storage_store_hooks.contains_key(&(account, provenance.root_slot)) {
+                continue;
+            }
+            if key.as_const().is_some()
```
figtracer: this filter skips constant slots and differently shaped `KECCAK` expressions before checking constraint equivalence. i reproduced false symbolic passes for both: concrete dispatches the mapping hook, while symbolic proves `calls == 0` with replay not required. please query feasibility first or mark the path incomplete.

crates/evm/symbolic/src/runtime/expr/expr.rs
```diff
@@ -224,6 +224,37 @@ impl SymExpr {
...
+            }
+            visited.push(current.clone());
+            let bytes = observed_preimage(&current)?;
+            if bytes.len() != 64 {
+                return None;
+            }
+            let key = Self::from_bytes(cx, bytes[..32].iter().cloned());
+            keys.push(key);
+            current = Self::from_bytes(cx, bytes[32..64].iter().cloned());
+            match current.kind() {
```
figtracer, line 247: this only accepts a syntactic constant mapping root. with `require(root == 0)` followed by `sstore(keccak256(key, root), value)`, concrete dispatches the root-0 hook but symbolic falsely proves zero callbacks. please match or fork roots under current constraints, including nested parents, or return incomplete.

crates/common/src/mapping_slots.rs
```diff
@@ -44,28 +76,86 @@ impl MappingSlots {
...
+    pending: PendingMappingHash,
+) {
+    let Ok(result) = interpreter.stack.peek(0) else { return };
+    let data = interpreter.memory.slice_len(pending.offset, 0x40);
+    let key = B256::from_slice(&data[..0x20]);
+    let parent = B256::from_slice(&data[0x20..]);
+    mapping_slots.entry(pending.address).or_default().record_hash(result.into(), key, parent);
+}
+
+/// Function to be used in `Inspector::step` to record mapping slots.
```
figtracer, line 114: `step` is now sstore-only, but `Fuzzer` still calls only this helper and has no `step_end` capture/record path. `seen_sha3` therefore stays empty, disabling mapping-aware storage-layout dictionary sampling in invariant campaigns. please track pending hashes outside the one-shot `collect` gate and add a regression.

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -574,6 +678,7 @@ impl SymbolicExecutor {
                 }
             }
             opcode::SSTORE => {
+                let pre_store_state = state.clone();
```
figtracer: this clones the entire `PathState` before every `SSTORE`, even with no hooks. an exact 5,000-store benchmark regressed from 0.60–0.61s to 0.90–0.91s, about 50%. please inspect the stack first and clone only when a fork is needed.

crates/cheatcodes/src/inspector.rs
```diff
@@ -1983,11 +2046,24 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>> for Cheatcode
...
         if let Some(mapping_slots) = &mut self.mapping_slots {
             mapping_step(mapping_slots, interpreter);
         }
 
+        let account = interpreter.input.target_address;
```
figtracer: this mapping-hook lookup now runs on every opcode whenever any raw storage hook is active. a raw-hook compute-only benchmark regressed about 9.4%, from a 4.14s to 4.53s median. please short-circuit when `mapping_storage_store_hooks` is empty so raw hooks keep their prior hot path.

crates/evm/fuzz/src/inspector.rs
```diff
@@ -46,15 +51,18 @@ pub struct Fuzzer {
...
             self.collect_data(interp);
-            if let Some(mapping_slots) = &mut self.mapping_slots {
-                mapping_step(mapping_slots, interp);
-            }
         }
     }
 
+    #[inline]
+    fn step_end(&mut self, interp: &mut Interpreter, _context: &mut CTX) {
+        self.record_mapping_hash(interp);
```
figtracer, line 63: can we just inline this instead

crates/evm/symbolic/src/runtime/expr/expr.rs
```diff
@@ -224,6 +224,10 @@ impl SymExpr {
         Some(StorageMappingKey { key, key_bytes: preserve_key_bytes, slot })
     }
 
+    pub(crate) fn is_storage_mapping_key(&self, cx: &mut SymCx) -> bool {
```
figtracer: also this

mablr: yes

crates/evm/fuzz/src/inspector.rs
mablr: this is trait method impl

figtracer: lmao glanced over it mb

crates/evm/symbolic/src/executor/opcodes.rs
mablr: Addressed in 9878f80be. SSTORE now peeks at and validates the operands, computes hook/provenance state first, and clones PathState only when an equality/inequality fork is actually required. An exact-head 5,000-store symbolic benchmark (10 runs, 2 warmups) improved from 1.262 ± 0.007 s to 807.1 ± 7.2 ms, or 1.56x.

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -1,6 +1,150 @@
...
+                let equality = SymBoolExpr::eq(&mut self.cx, key.clone(), hash.clone());
+                if let Some(provenance) =
+                    self.classify_mapping_match(state, equality, provenance)?
+                {
+                    return Ok(provenance);
+                }
+                continue;
+            }
+            if (key_is_const || key_contains_keccak) && !key_is_storage_mapping_key {
+                continue;
```
figtracer, line 122: a compiler-folded mapping slot is skipped here before solver equivalence is checked. 

with `require(keccak256(abi.encode(key, 0)) == SLOT_42)` followed by `sstore(SLOT_42, value)`, current head reports `assert(callbacks == 0)` as a symbolic pass, while concrete execution fires the callback. 

the added test computes the slot at runtime, so its preimage is recorded and misses this case. please preserve the observed hash provenance through the equality or mark the path incomplete.

crates/cheatcodes/src/inspector.rs
```diff
@@ -1983,11 +2046,27 @@ impl<FEN: FoundryEvmNetwork> Inspector<FoundryContextFor<'_, FEN>> for Cheatcode
...
+                .get(&account)
+                .is_some_and(|hooks| !hooks.is_empty());
+        if mapping_hook_active {
+            mapping_step(&mut self.storage_hook_mapping_slots, interpreter);
+        }
+        self.pending_mapping_hash = if self.mapping_slots.is_some() || mapping_hook_active {
+            capture_mapping_hash(interpreter)
+        } else {
+            None
+        };
```
figtracer: this mapping bookkeeping still executes on every opcode when only a raw SSTORE hook is registered. 

an optimized 1,000,000-iteration compute-only benchmark takes 322.2 ms on this head versus 284.5 ms on the PR base across ten runs, a stable 13% regression. 

please bypass the entire mapping block when mapping recording is disabled and no mapping hooks exist.

crates/evm/symbolic/src/executor/opcodes.rs
mablr: Addressed in e6c7c9459. Constant SSTORE slots now test whether the inequality with each observed mapping hash is feasible; when current constraints prove equality, the original observed provenance dispatches exactly. Unconstrained constant/Keccak collisions are not forked. Added checkConstraintEquivalentFoldedSlot, which constrains a runtime observed hash to compiler-folded SLOT_42 and verifies callback key/slot parity.

figtracer (approved):

mattsse (approved):
lgtm

PR #15978, opened 2026-07-30
fix(cast): reject malformed vanity wallet files

grandpig:
## Motivation

Existing non-empty files passed via `cast wallet vanity --save-path` were treated as empty whenever deserialization failed because `unwrap_or_default()` discarded the parse error. The subsequent write could therefore replace previously saved wallets, including their private keys, with only the newly generated wallet.

This change keeps the existing empty-file initialization behavior, but returns a path-aware error for malformed non-empty files before any write occurs. The regression test also verifies that the original file contents remain unchanged.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/.changelog/reject-malformed-vanity-wallet-files.md
+++ b/.changelog/reject-malformed-vanity-wallet-files.md
@@ -1,0 +1,5 @@
+---
+cast: patch
+---
+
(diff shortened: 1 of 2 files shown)


stevencartavia (approved):

mablr (approved):
Makes sense

PR #15984, opened 2026-07-31
fix(forge): resolve symlinked library artifacts

### LINKED ISSUE #11593: Unable to run "forge test" in a package that depends upon another package in a monorepo because a library artifact is not found
### Component

Forge

- [x] Foundry
- [x] Foundryup

v1.3.5-nightly

v1.3.0

forge test

### Operating System

macOS (Apple Silicon)

## Problem

`forge test` fails to find library artifacts for workspace dependencies, while `forge build` works fine.

**Error:**
```
Error: wasn't able to find artifact for library Helpers at ../node_modules/@sablier/lockup/src/libraries/Helpers.sol
```

## Structure
```
monorepo/
├── foundry.toml              # Base config: "@sablier/=../node_modules/@sablier/"
├── node_modules/@sablier/
│   └── lockup -> ../../lockup # Bun workspace symlink
├── lockup/                   # Workspace package with Helpers library
│   └── out/Helpers.sol/Helpers.json ✅ Artifact exists
└── airdrops/                 # Workspace package depending on lockup
    ├── foundry.toml          # Inherits base config
    └── tests/                # Tests fail to link Helpers library
```

## Dependency

The `airdrops` package depends upon the `lockup` package like so:

```json
"@sablier/lockup": "workspace:*"
```

> [!NOTE]
> Switching to a remote version of the package, e.g. `github:sablier-labs/lockup#staging`, doesn't solve the issue.

## Root Cause

I think the root cause is that the test runner's library linking can't resolve artifacts through Bun workspace symlinks, even though compilation works.

Possible related issue: https://github.com/foundry-rs/foundry/issues/11587

## Repro

1. Clone https://github.com/sablier-labs/evm-monorepo
2. `bun install`
3. `cd airdrops`
4. `forge test`

The tests in all other packages work.

## Environment

- bun v1.2.21 with [workspaces](https://bun.com/docs/install/workspaces) enabled
- forge v1.3.5-nightly

stevencartavia:
Fixes #11593 by resolving library artifact paths through workspace symlinks while preserving exact configured paths and existing nonce and CREATE2 linking behavior.

### DIFF
--- a/crates/linking/src/lib.rs
+++ b/crates/linking/src/lib.rs
@@ -196,7 +196,6 @@ impl<'a> Linker<'a> {
         target: &ArtifactId,
     ) -> Result<Option<&'a ArtifactId>, LinkerError> {
         let library_path = self.project_relative_path(Path::new(file));
-        let canonical_library_path = self.canonical_path(&library_path);
         let candidates = self
             .contracts
             .keys()
@@ -205,14 +204,29 @@ impl<'a> Linker<'a> {
                     return false;
                 }
                 let (artifact_path, artifact_name) = self.convert_artifact_id_to_lib_path(id);
-                artifact_name == *name
-                    && self.path_matches(
-                        &artifact_path,
-                        &library_path,
-                        canonical_library_path.as_deref(),
-                    )
+                artifact_name == *name && artifact_path == library_path
             })
             .collect::<Vec<_>>();
+        let candidates = if candidates.is_empty() {
+            let canonical_library_path = self.canonical_path(&library_path);
+            self.contracts
+                .keys()
+                .filter(|id| {
+                    if id.version != target.version {
+                        return false;
+                    }
+                    let (artifact_path, artifact_name) = self.convert_artifact_id_to_lib_path(id);
+                    artifact_name == *name
+                        && self.path_matches(
+                            &artifact_path,
+                            &library_path,
+                            canonical_library_path.as_deref(),
+                        )
+                })
+                .collect()
+        } else {
+            candidates
+        };
         let same_build_and_profile = candidates
             .iter()
             .copied()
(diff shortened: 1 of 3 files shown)


crates/linking/src/lib.rs
```diff
@@ -199,20 +254,63 @@ impl<'a> Linker<'a> {
...
+        &self,
+        references: &BTreeMap<&ArtifactId, BTreeSet<PathBuf>>,
+        libraries: &mut Libraries,
+    ) -> Result<(), LinkerError> {
+        for (id, references) in references {
+            let (file, name) = self.convert_artifact_id_to_lib_path(id);
+            let mut configured = references.iter().filter_map(|reference| {
+                libraries
+                    .libs
+                    .get(reference)
```
grandizzy: **Configured addresses under a symlink-equivalent path are ignored here.**

Dependency resolution above allows a link reference and artifact path to resolve to the same canonical file, but this lookup only checks the compiler reference’s exact lexical spelling.

For example, if the link reference is:

```text
../node_modules/@workspace/library/src/Helpers.sol
```

while `foundry.toml` configures:

```text
../library/src/Helpers.sol:Helpers:0x...
```

dependency discovery succeeds, but this function does not propagate the configured address to the artifact key. Nonce/CREATE2 linking can then generate and deploy a different address, silently ignoring the configuration.

Could we first preserve exact-key priority, then fall back to a canonical-equivalent configured key? If multiple equivalent keys specify different addresses, they should produce a conflict. It would also be helpful to cover this with the workspace-symlink test using a configured address under the real path.

crates/linking/src/lib.rs
```diff
@@ -149,14 +196,21 @@ impl<'a> Linker<'a> {
...
                 let (artifact_path, artifact_name) = self.convert_artifact_id_to_lib_path(id);
-                id.version == target.version
-                    && artifact_name == *name
-                    && artifact_path == library_path
+                artifact_name == *name
+                    && self.path_matches(
+                        &artifact_path,
+                        &library_path,
+                        canonical_library_path.as_deref(),
+                    )
```
grandizzy: **Please prefer exact artifact-path matches before considering canonical fallbacks.**

This currently puts exact and canonical-equivalent artifacts into the same candidate set. If the same physical library is compiled under both a direct path and a workspace-symlink path, an exact compiler reference can therefore match both artifact IDs and produce `ConflictingLibraryArtifacts`, even though the exact path identifies the intended artifact.

Could candidate resolution use this order?

1. Collect exact path matches.
2. Only if none exist, collect canonical-equivalent matches.
3. Apply the existing build/profile narrowing.
4. Report ambiguity if multiple candidates remain.

A regression test where the same library is compiled through both direct and symlink import paths would cover this case.

grandizzy (changes requested):
Thanks for tackling this—the canonical-path fallback is the right direction and the workspace integration test captures the original failure well.

Two cases should be addressed before merging:

1. A configured library address under the real path can be ignored when the compiler reference uses a symlink-equivalent path, causing Forge to generate and deploy a different address.
2. Canonical matching is considered alongside exact matching, which can turn an otherwise unambiguous exact artifact match into `ConflictingLibraryArtifacts`.

Both should be fixable by treating canonical resolution as a fallback after exact matching and applying the same equivalence handling to configured library keys.

crates/linking/src/lib.rs
stevencartavia: [6178f73](https://github.com/foundry-rs/foundry/pull/15984/commits/6178f731eda416cbdb3891c0fca5a9a1f60d8652)

crates/linking/src/lib.rs
stevencartavia: [6178f73](https://github.com/foundry-rs/foundry/pull/15984/commits/6178f731eda416cbdb3891c0fca5a9a1f60d8652)

grandizzy (approved):
ty, lgtm!

mablr (approved):

figtracer (approved):

mattsse (approved):
lgtm

PR #15988, opened 2026-07-31
perf(forge): speed up selector commands

figtracer:
## Summary

Reuses the project's dynamic test-linking mode for ABI-based `forge selectors` commands. This lets a warm normal build satisfy the narrower ABI request from its existing artifacts instead of invoking a fresh compiler job. `selectors collision` is intentionally unchanged because it requests method identifiers rather than ABI-only output.

AI assistance: Codex was used to investigate, benchmark, and implement this change.

## Benchmarks

Aave v4, warm normal cache, profiling builds, 10 runs:

| Command | master | PR | Speedup |
| --- | ---: | ---: | ---: |
| `forge selectors list` | `2.034s ± 0.161s` | `74.0ms ± 0.8ms` | `27.49x` |

Peak RSS averaged over three runs decreased from `1350.4 MiB` to `211.0 MiB` (`6.40x` lower).

### DIFF
--- a/.changelog/forge-selectors-cache.md
+++ b/.changelog/forge-selectors-cache.md
@@ -1,0 +1,5 @@
+---
+forge: patch
+---
+
+Reused the preprocessed compiler cache for ABI-based `forge selectors` commands.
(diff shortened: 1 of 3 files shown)


mattsse (approved):
lgtm

grandizzy (approved):
lgtm!

mablr (approved):

PR #15990, opened 2026-07-31
test(forge): batch binding compile checks

decofe:
The generated-binding regression tests each ran a separate nested `cargo check`, causing three concurrent cold builds to exceed nextest's 90-second limit on macOS ARM. This keeps the individual binding-generation cases while compiling all three fixtures together in one generated crate, preserving compile-time coverage without increasing the timeout.

This PR was written primarily by Codex.

Prompted by: @grandizzy

### DIFF
--- a/.changelog/batch-bind-compile-tests.md
+++ b/.changelog/batch-bind-compile-tests.md
@@ -1,0 +1,5 @@
+---
+forge: patch
+---
+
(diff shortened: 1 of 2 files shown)


grandizzy (approved):

mattsse (approved):
lgtm

PR #15991, opened 2026-07-31
chore(ci): seed root changelog

decofe:
Seeds the root changelog before its first generated release so the stable-release validator observes `CHANGELOG.md` as modified rather than newly added. This is repository maintenance and requires the `L-ignore` label instead of a changelog fragment.

This PR was written primarily by Codex.

Prompted by: @grandizzy


mattsse (approved):
lgtm

grandizzy (approved):

PR #15992, opened 2026-07-31
fix(config): preserve Windows remapping contexts

figtracer:
#15973 preserved trailing separators when contextual remappings left the config provider, but Windows path normalization crosses several more serialization boundaries. Config and CLI merging could still turn a directory context such as `lib/outer/` into the lexical prefix `lib/outer`, and `ProjectBuilder` converted the repaired native separator back to `/`. Solar then compared that context directly with source names containing native or mixed separators, so solc compilation and Solar linting could resolve the same import differently.

This preserves contextual directory boundaries when config structs and merged CLI remappings are serialized, then restores the native boundary after project paths are slashed. At the Solar boundary, source names and remapping contexts pass through one `PathExt::to_slash_lossy()` compatibility adapter so lexical matching uses consistent separators without rewriting UNC or verbatim prefixes or dropping a trailing directory boundary. The durable ownership remains upstream: contextual-boundary preservation belongs in `foundry-compilers`, while separator-equivalent remapping matching belongs in Solar.

The [focused Windows job](https://github.com/foundry-rs/foundry/actions/runs/30639853898/job/91186758455) ran the three reported regressions plus the UNC/verbatim-path regression and passed all four.

> This PR's implementation, tests, description, and review follow-up were written primarily with Codex.


mablr (dismissed):
sgtm

crates/cli/src/opts/build/utils.rs
```diff
@@ -283,7 +283,11 @@ pub fn configure_pcx_from_solc(
             .sources
             .par_iter()
             .filter_map(|(path, source)| {
-                pcx.sess.source_map().new_source_file(path.clone(), source.content.as_str()).ok()
+                #[cfg(windows)]
+                let path = PathBuf::from(path.to_string_lossy().replace('\\', "/"));
```
grandizzy: Could we avoid the blind `replace(\, "/")` and use one prefix-aware normalization helper for both source names and remapping contexts? The CLI already depends on `path-slash`, and `PathExt::to_slash_lossy()` matches the convention used by `foundry-compilers` while preserving UNC and verbatim Windows prefixes. Blind replacement turns `\\server\share` into `//server/share` and `\\?\C:` into `//?/C:`, which can change `PathBuf` prefix and absolute-path semantics inside Solar.

The helper should explicitly retain an existing trailing `/` or `\`, because `path-slash` can otherwise drop a trailing non-native `/` on Windows. Please add a focused UNC/verbatim regression if this adapter remains local.

Longer term, the trailing remapping-context boundary should be fixed in `foundry-compilers`, while separator-equivalent remapping matching belongs in Solar’s existing `sanitize_path` TODO.

grandizzy (commented):
Looking good overall, and the reported Windows remapping cases are well covered. Please check the path-normalization finding inline.

The durable fix should be split upstream: contextual-boundary preservation belongs in `foundry-compilers`, while source/context separator normalization belongs in Solar. The local compatibility adapter is reasonable in the meantime, but it should use one prefix-aware helper rather than blind string replacement.

crates/cli/src/opts/build/utils.rs
figtracer: Addressed in cc592041b: both Solar source names and remapping contexts now use one `PathExt::to_slash_lossy()` helper, which preserves Windows UNC/verbatim prefixes and explicitly retains either trailing separator. I added Windows coverage for native, UNC, verbatim drive, and verbatim UNC paths; the focused Windows job passed all four relevant tests: https://github.com/foundry-rs/foundry/actions/runs/30639853898/job/91186758455

mablr (dismissed):

crates/config/src/lib.rs
```diff
@@ -3122,6 +3142,31 @@ impl BasicConfig {
...
+        S: Serializer,
+    {
+        remappings
+            .iter()
+            .map(|remapping| {
+                let Some(context) = &remapping.context else { return remapping.to_string() };
+                let mut remapping = remapping.clone();
+                remapping.context = None;
+                #[cfg(windows)]
+                let context = context.replace('\\', "/");
```
stevencartavia: On Windows, replacing every `\` turns a verbatim UNC context such as `\\?\UNC\server\share\project\` into `//?/UNC/server/share/project/`. `std::path` parses the first as `Prefix::VerbatimUNC("server", "share")`, but the second as `Prefix::UNC("?", "UNC")`, so a config serialization round trip changes which path the remapping context denotes. Could we preserve verbatim UNC prefixes while slash-normalizing the remainder?

crates/cli/src/opts/build/utils.rs
```diff
@@ -298,11 +322,33 @@ fn configure_pcx_from_solc_cli(
     pcx.file_resolver
         .set_current_dir(cli_settings.base_path.as_ref().unwrap_or(&project_paths.root));
     for remapping in &project_paths.remappings {
+        let context = remapping.context.clone().unwrap_or_default();
+        // Solar compares the context directly with the parent source path. Match the slash form
+        // used above instead of allowing mixed Windows separators to change prefix semantics.
+        #[cfg(windows)]
+        let context = solar_slash_path(Path::new(&context));
```
stevencartavia, line 329: On Windows, `bind-json` calls `configure_pcx_from_solc(..., false)` and then manually adds the original native source paths. After Solar strips the project root, it compares a source like `src\A.sol` lexically against the normalized `src/` context, so a CLI-provided contextual remapping such as `src/:pkg/=vendor/pkg/` no longer applies. Could we slash-normalize these manually added source paths too?

grandizzy:
maybe we should upstream sanitize path to solar https://github.com/paradigmxyz/solar/blob/9593f29765e302eb3c21cfd409239acc16781d60/crates/interface/src/source_map/file_resolver.rs#L423-L426 and use that one, https://github.com/paradigmxyz/solar/pull/758 attempted smth similar

mablr (approved):

grandizzy (approved):

PR #15993, opened 2026-07-31
test(cast): cover fallback selectors

stevencartavia:
Adds an offline regression test ensuring `cast selectors` includes ABI-dispatched selectors while excluding selectors dispatched through fallback logic.

Ref #15985


mablr (approved):

mattsse (approved):
lgtm

PR #15994, opened 2026-07-31
fix(cast): accept 0x-prefixed vanity patterns

racequite:
## Motivation

`cast wallet vanity` treats patterns as exact hex when decoding succeeds and otherwise falls back to regex matching against an unprefixed hexadecimal address. As a result, a value such as `--starts-with 0xdead` became the impossible regex `^0xdead`, causing the infinite wallet generator to keep consuming CPU without ever finding a match.

## Solution

Strip an optional `0x` or `0X` prefix before parsing, and reject patterns that are empty after stripping. This preserves odd-nibble patterns such as `0x9` through the existing regex path and applies equally to `--starts-with` and `--ends-with`. Unit tests cover prefixed exact hex, uppercase prefixes, odd-nibble matching, and empty-pattern rejection.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/.changelog/accept-prefixed-vanity-patterns.md
+++ b/.changelog/accept-prefixed-vanity-patterns.md
@@ -1,0 +1,5 @@
+---
+cast: patch
+---
+
+Accepted `0x`-prefixed patterns in `cast wallet vanity`.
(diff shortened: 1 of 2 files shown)


stevencartavia (dismissed):

racequite:
@stevencartavia Thank you for your approval. If there's anything need I to do, please feel free to let me know.

mablr (commented):
@racequite could you please resolve conflict? Otherwise seems fine.

racequite:
> @racequite could you please resolve conflict? Otherwise seems fine.

@mablr @stevencartavia Resolved. Please review it again. Thnaks!

mablr (approved):

stevencartavia (approved):

PR #15995, opened 2026-07-31
feat(cast): add Touch ID enrollment

### LINKED ISSUE #15815: feat(cast): Touch ID-protected keystores on macOS (Secure Enclave-wrapped)
### Component

Cast

Using an encrypted keystore with `cast` today means either retyping the password on every command or leaking it through `--password` in argv, `ETH_PASSWORD`, `--unsafe-password`, or a plaintext password file (#2794 added file passwords for exactly this friction). On macOS, `cast` could offer the UX users already know from `sudo` with `pam_tid`, 1Password, and `age-plugin-se`: unlock a key with Touch ID.

Proposed UX:

```console
$ cast wallet import deployer --interactive --touch-id
Enter private key: ****
[Touch ID prompt]
`deployer` keystore saved; unlock protected by Touch ID on this Mac.

$ cast send --account deployer 0x... "transfer(...)"
[Touch ID prompt]
✅ Sent transaction ...
```

### Proposed design: Secure Enclave key-wrapping (the `age-plugin-se` pattern)

The obvious approaches do not work for a CLI, so it is worth being explicit about the design that does:

1. On `--touch-id` enrollment, generate a P-256 key **inside the Secure Enclave** with `SecAccessControl(.privateKeyUsage | .userPresence)` (or `.biometryCurrentSet` for strict mode).
2. Use it to ECIES-wrap the keystore password (or the secp256k1 key itself), and store the wrapped blob plus the enclave key's encrypted `dataRepresentation` as a sidecar next to the normal keystore JSON in `~/.foundry/keystores`.
3. At signing time, decryption of the blob requires the Secure Enclave to evaluate the access control — the Touch ID prompt is **hardware-enforced by the SEP**. The decrypted secret then unlocks the keystore in memory as usual.

Key properties, all verified against Apple docs (TN3137, Secure Enclave docs) and empirically with ad-hoc-signed Swift probes on an Apple Silicon Mac:

- **Works for an unbundled, foundryup-distributed CLI.** No entitlements, provisioning profile, app bundle, or Developer ID signing needed — only the ad-hoc linker signature every Apple Silicon binary already has. `age-plugin-se` ships exactly this via Homebrew.
- **No Keychain items at all**, which deliberately sidesteps two traps: (a) biometry-protected keychain items (`kSecAttrAccessControl`) require the data-protection keychain, which requires provisioning-profile-backed entitlements in an app-like bundle — an ad-hoc CLI gets `errSecMissingEntitlement (-34018)`; (b) file-based keychain "Always Allow" ACLs bind to the binary's code signature, so every `foundryup` upgrade of an unsigned binary would re-prompt (aws-vault code-signs its releases solely to avoid this; `gh` execs `/usr/bin/security` instead, which lets any local process read the item).
- **The keystore JSON stays canonical.** Touch ID is a convenience unlock, not custody: the encrypted keystore remains the portable, recoverable artifact, so losing the Mac, wiping biometrics, or moving machines degrades gracefully to the password path. Security is not reduced for existing users; nothing changes unless you opt in.

### Constraints (inherent to the platform, worth recording as non-goals)

- **Signing can never happen inside the Secure Enclave.** The SE supports only NIST P-256 and cannot import pre-existing keys (Apple-documented), so the secp256k1 key is always reconstructed in process memory to sign — same exposure window as today's keystore decrypt. (SE-*resident* signing is only usable on-chain via contract-verified P-256 flows — EIP-7951 is live on L1 since Fusaka, RIP-7212 on major L2s — i.e. smart accounts / EIP-7702, out of scope here.)
- **iCloud Keychain sync is off the table.** Per-item biometry ACLs and `kSecAttrSynchronizable` are mutually exclusive (`SecItemAdd` returns `errSecParam` when combined), sync requires the entitlement-gated data-protection keychain anyway, and SE blobs are device-bound by design.
- **GUI login session only.** Touch ID prompts cannot be raised over SSH, in headless CI, or from daemons; the flow must detect this and fall back to the password prompt with a clear message.
- With `.biometryCurrentSet`, re-enrolling fingerprints permanently invalidates the wrap key — another reason the keystore JSON stays the source of truth.

### Additional context

**Implementation home.** Signer/credential resolution lives in `foundry-wallets` (foundry-core), not this repo. The natural seam is the keystore-password resolution in `create_keystore_signer` / `PendingSigner::unlock` (currently: `--password` > `--password-file` > interactive `rpassword` prompt) — a Touch ID unwrap slots in before the interactive prompt, keyed by the keystore address already carried in `PendingSigner::Keystore`. That makes the change a password *source*, far smaller than a new `WalletSigner` backend, and every consumer (`cast send`/`mktx`/`call`/`estimate`, `forge create`, `forge script` broadcast and cheatcode wallets) benefits with no per-command changes. The cast side is enrollment UX on `cast wallet import`/`new`, plus `list`/`remove` awareness. The Swift shim is unavoidable but tiny: the SE key's `dataRepresentation` is a CryptoKit-only API with no C surface, so a small compiled-in Swift helper (or `swift-rs`) is needed, feature-gated and `target_os = "macos"`-only, following the existing `hidapi`/`aws-kms` optional-dependency patterns.

**Naming.** `cast keychain` is taken by Tempo's on-chain AccountKeychain, and "keyring" collides with GCP KMS "key ring" terminology in existing PRs — hence `--touch-id` (or `--secure-enclave`) rather than anything keychain-flavored. The macOS Keychain is genuinely not involved in this design, so the naming is also more accurate.

**Duplicate check.** No existing issue or PR requests OS-keychain/biometric/Secure Enclave key storage; all ~30 "keychain" hits are Tempo's on-chain keychain, and "keyring" hits are GCP KMS.

**Prior art.** `age-plugin-se` (this exact pattern, Homebrew-distributed, no signing infra); Secretive (SE-resident SSH agent, Touch ID per signature); aws-vault (keychain items + signed releases); 1Password `op` (signed desktop-app agent over XPC); `dfx` (Rust `keyring` crate, plain keychain item, no biometrics); geth `clef` (own encrypted vault, no OS integration).

**Rust deps.** `security-framework` ≥3.7 exposes `SecAccessControl` with the biometry flags; `objc2-local-authentication` covers `LAContext`; the CryptoKit bridge is the only Swift piece. A future cross-platform password-source seam (Windows Hello, Secret Service) could reuse the same hook, but macOS-first keeps the scope honest.

dadadave80:
Adds opt-in Touch ID enrollment to `cast wallet new` and `cast wallet import`, backed by the Secure Enclave support now provided by `foundry-wallets`. Existing keystore password behavior remains the fallback, while wallet listing, removal, and password changes account for the Touch ID sidecar lifecycle. The `touch-id` feature is propagated through Cast, Forge, release builds, and Docker builds, with macOS CI link checks for both Apple Silicon and Intel targets.

Closes #15815.

This PR's implementation, tests, workflow changes, and description were written primarily with Codex.


crates/cast/src/cmd/wallet/mod.rs
```diff
@@ -998,10 +1078,38 @@ flag to set your key via:
...
+                    &new_password,
                     Some(&account_name),
                 )?;
 
+                #[cfg(all(target_os = "macos", feature = "touch-id"))]
+                if touch_id_enrolled
+                    && let Err(enrollment_error) = foundry_wallets::touch_id::enroll(
+                        &keypath,
+                        &new_password,
+                        foundry_wallets::touch_id::Policy::default(),
```
stevencartavia: `wallet change-password` re-enrolls an existing Touch ID enrollment with `Policy::default()`, silently changing `CurrentBiometry` to `UserPresence`.

Reproduced: the command succeeded, the old password stopped working, the new password worked, and the persisted policy changed from `current-biometry` to `user-presence`.

Could we preserve the existing policy when re-enrolling?

crates/cast/src/cmd/wallet/mod.rs
```diff
@@ -438,22 +459,46 @@ impl WalletSubcommands {
...
-                                password.clone(),
+                                &password,
                                 account_name_ref.as_deref(),
                             )?;
                             let identifier = account_name_ref.as_deref().unwrap_or(&uuid);
+                            let keystore_path = path.join(identifier);
+
+                            #[cfg(all(target_os = "macos", feature = "touch-id"))]
+                            if touch_id {
+                                ensure_touch_id_sidecar_available(&keystore_path)?;
```
stevencartavia: The sidecar collision check runs after `new_keystore` writes each target.

Reproduced with a valid keystore at `batch_2.touchid`: `wallet new --number 2 --touch-id` created `batch_1`, `batch_1.touchid`, and `batch_2`, then exited with an error on the collision.

Could we preflight all named sidecar paths before creating the batch?

crates/cast/src/cmd/wallet/mod.rs
mablr: +1

crates/cast/src/cmd/wallet/mod.rs
mablr: +1

crates/cast/src/cmd/wallet/mod.rs
```diff
@@ -1086,12 +1232,197 @@ flag to set your key via:
...
+fn is_touch_id_sidecar(path: &Path) -> Result<bool> {
+    let Some(name) = path.file_name().and_then(|name| name.to_str()) else {
+        return Ok(false);
+    };
+    if !name.ends_with(TOUCH_ID_SIDECAR_SUFFIX) {
+        return Ok(false);
+    }
+
+    match is_keystore(path) {
+        Ok(is_keystore) => Ok(!is_keystore),
```
mablr: **Blocker:** This classifies every parseable non-keystore JSON file as a Touch ID sidecar, so unrelated `account.touchid` JSON can be hidden, overwritten, or deleted. Please strictly parse the sidecar schema and preserve/refuse unknown JSON; add `{}`, array, unrelated-object, and unknown-version tests.

crates/cast/src/cmd/wallet/mod.rs
```diff
@@ -435,31 +461,69 @@ impl WalletSubcommands {
...
+                                }
+                                json.push(result);
                             } else {
                                 sh_status!(
                                     "Created new encrypted keystore file: {}",
-                                    path.join(identifier).display()
+                                    keystore_path.display()
                                 )?;
+                                if touch_id {
+                                    sh_status!("Touch ID unlock enabled.")?;
```
mablr: The default `UserPresence` policy allows macOS login-password fallback, and explicit password sources bypass Touch ID. This message overstates the guarantee. Please clarify the help/success text and changelog, or use a biometric-only policy if Touch ID must be enforced.

.github/workflows/release.yml
```diff
@@ -321,11 +321,12 @@ jobs:
...
 
-      - name: Apple M1 setup
-        if: matrix.target == 'aarch64-apple-darwin'
+      - name: Apple setup
+        if: contains(matrix.target, 'apple-darwin')
         run: |
           printf 'SDKROOT=%s\n' "$(xcrun -sdk macosx --show-sdk-path)" >> "$GITHUB_ENV"
-          printf 'MACOSX_DEPLOYMENT_TARGET=%s\n' "$(xcrun -sdk macosx --show-sdk-platform-version)" >> "$GITHUB_ENV"
+          # Touch ID APIs require macOS 11 or newer.
+          printf 'MACOSX_DEPLOYMENT_TARGET=11.0\n' >> "$GITHUB_ENV"
```
mablr, line 330: This also raises the minimum OS for the Intel artifact to macOS 11, including users who never use Touch ID. Can we preserve the prior x86_64 deployment target (after checking the required APIs), or explicitly document this compatibility break?

crates/cast/src/cmd/wallet/mod.rs
```diff
@@ -72,6 +76,10 @@ pub enum WalletSubcommands {
         /// Overwrite existing keystore files without prompting.
         #[arg(long)]
         force: bool,
+
+        /// Also enroll the keystore for Touch ID unlock (macOS only).
```
mablr: What is the recovery path for existing or invalidated enrollments? Existing keystores cannot enroll, disable, inspect, or re-enroll even though library errors instruct users to re-enroll. Please add a small lifecycle command surface (at least enroll/remove/status).

crates/cast/src/cmd/wallet/mod.rs
dadadave80: Fixed in https://github.com/dadadave80/foundry/pull/5, which has now been merged into the PR branch.

The implementation now strictly parses the supported v1 Touch ID sidecar schema, explicitly distinguishes missing, recognized, keystore, and unknown files, hides or removes only recognized sidecars, and refuses to overwrite, delete, or rewrite around unknown or future JSON.

Regression coverage was added for `{}`, arrays, unrelated objects, unknown versions, unknown fields and policies, malformed JSON, and keystore collisions.

crates/cast/src/cmd/wallet/mod.rs
dadadave80: Addressed on  [fix/touch-id-auth-wording](https://github.com/dadadave80/foundry/pull/6)  by clarifying the user-facing wording without changing the existing authentication behavior.

The  --touch-id  help text now describes the feature as Touch ID-assisted macOS authentication and explicitly states that the macOS login password and explicit keystore passwords remain available. The enrollment success messages and changelog were updated with the same clarification, with regression coverage added for both command help outputs.

The existing  UserPresence  policy and password fallback behavior remain unchanged.

crates/cast/src/cmd/wallet/mod.rs
```diff
@@ -1086,12 +1269,567 @@ flag to set your key via:
...
+    if value.get("version").is_some()
+        && (value.get("crypto").is_some() || value.get("Crypto").is_some())
+    {
+        return Ok(TouchIdSidecarState::Keystore);
+    }
+
+    // Attempt strict sidecar parse.  Any missing/extra field or unsupported
+    // version/policy maps to `Unknown` rather than `Recognized`.
+    match serde_json::from_value::<TouchIdSidecarWire>(value) {
+        Ok(wire) if wire.version == TOUCH_ID_SIDECAR_VERSION => Ok(TouchIdSidecarState::Recognized),
```
mablr: Could we reject invalid payloads here? Empty, non-hex, or truncated strings currently become `Recognized`, making them eligible to be hidden, overwritten, or deleted. Decode both fields, require a non-empty `se_key` and a `sealed_password` of at least 93 bytes beginning with `0x04`; otherwise return `Unknown`. Please update the fixtures and add preservation tests.

mablr (commented):
Tested E2E, it looks awesome !

Just one last nit, see inline comment.

.github/workflows/release.yml
dadadave80: Good point. I’m concerned about raising the packaged Intel minimum to macOS 11 just for Touch ID, since that would cut older Intel Macs off from current Foundry releases even though Foundry itself doesn’t otherwise require macOS 11.

I considered shipping separate legacy and Touch ID-enabled Intel artifacts, but that would also require `foundryup` to become macOS-version-aware so it can choose the right artifact, plus extra release/attestation complexity.

Given the scope, I’m leaning toward keeping the existing Intel deployment target and excluding `touch-id` from the packaged `x86_64-apple-darwin` build for now, while keeping it enabled on Apple Silicon. Intel Touch ID support could be revisited separately with proper installer-aware artifact selection.

Does that sound like the better tradeoff?

mablr: > Given the scope, I’m leaning toward keeping the existing Intel deployment target and excluding touch-id from the packaged x86_64-apple-darwin build for now, while keeping it enabled on Apple Silicon. Intel Touch ID support could be revisited separately with proper installer-aware artifact selection.

Seems fair to me 👍

crates/cast/src/cmd/wallet/mod.rs
dadadave80: Addressed in 215d5b4ec. The sidecar classifier now decodes both payload fields, requires a non-empty se_key, and requires sealed_password to be at least 93 bytes and begin with 0x04. Empty, non-hex, truncated, and invalid-prefix payloads are classified as Unknown, so they cannot be hidden, overwritten, or deleted. I also updated the fixtures and added preservation tests for these invalid cases. Could you confirm this resolves the concern?

crates/cast/src/cmd/wallet/mod.rs
dadadave80: I’ve added the Touch ID lifecycle commands in e3cf0aab6f57dccda84b19c80258ed91f1583ae8 and use foundry_wallets::touch_id::policy() during re-enrollment to preserve the existing persisted policy. This should avoid duplicating sidecar parsing in Foundry. Could you please confirm this addresses your earlier concern?

.github/workflows/release.yml
dadadave80: Addressed in 10ea82e2bb26f850e5fc93c81ca575d871a2256b.
The packaged x86_64-apple-darwin build now keeps its previous deployment behavior and no longer enables touch-id. touch-id is added only to the aarch64-apple-darwin release build, which remains at a macOS 11 deployment target.
I kept the Intel Touch ID CI leg at macOS 11 so the feature remains explicitly buildable/tested for compatible Intel Macs; it is just not included in the standard Intel release artifact for now.

mablr (commented):
Nice, LGTM on Foundry side.

I just found 2 nits to harden on `foundry-wallets` side https://github.com/foundry-rs/foundry-core/pull/153, please re-pin once merged.

mablr (dismissed):
LGTM, thank you very much @dadadave80 !

mablr (approved):

figtracer (approved):

PR #15996, opened 2026-07-31
fix(forge): preserve fuzz replay runs

stevencartavia:
Prevent persisted failure replays from consuming generated fuzz runs, preserving campaign coverage and reproduction metadata.


mattsse (approved):
lgtm

mablr (approved):

PR #15997, opened 2026-07-31
fix(cast): handle odd-length vanity hex

stevencartavia:
Treat odd-length plain-hex vanity patterns case-insensitively instead of allowing uppercase nibbles to become impossible regexes that consume CPU indefinitely. Reject plain-hex patterns longer than an address before starting wallet generation.


mattsse (approved):
lgtm

mablr (approved):

PR #15999, opened 2026-08-01
fix(fuzz): preserve enum bounds in mutations

stevencartavia:
Preserve Solidity enum bounds during ABI and comparison-guided corpus mutations. This prevents invalid enum values from reverting in the ABI decoder before the fuzz test body executes.


grandizzy (approved):
lgtm

mablr (approved):
sgtm

PR #16000, opened 2026-08-02
fix(forge): link multi-profile libraries

mablr:
## Description

#15839 correctly stopped silently choosing between conflicting library artifacts, but it also exposed that Forge flattened libraries to `(path, name)` before linking. When the same library is compiled under multiple compiler profiles, as in Aave v4’s `LiquidationLogic` tests, those distinct artifacts collide even though their build and profile identities are available.

This change preserves artifact identity throughout nonce and CREATE2 linking and gives each consumer the complete transitive address map for its selected variants. Detailed output exposes both per-consumer library mappings and every `ArtifactId -> Address` association, while CREATE2 may still deduplicate byte-identical variants into one physical deployment. The legacy flattened APIs continue rejecting genuinely lossy mappings instead of silently selecting an artifact.

Tested against aave/aave-v4 @ af1f0f2b, now forge test works correctly.

### DIFF
--- a/.changelog/multi-profile-library-linking.md
+++ b/.changelog/multi-profile-library-linking.md
@@ -1,0 +1,5 @@
+---
+forge: patch
+---
+
+Fixed test linking when multiple compiler profiles produce different artifacts for the same library.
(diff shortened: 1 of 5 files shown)


mattsse (approved):
lgtm

stevencartavia (approved):

PR #16003, opened 2026-08-03
fix(mutation): retain gas-distinct bounds

mablr:
## Motivation

#15942 correctly identified comparison mutants that produce the same Boolean result across an operand’s valid type range. However, equal truth tables do not guarantee equivalent EVM behavior: with the optimizer disabled, `uint256 value == 0` compiles using `EQ`, while `value <= 0` uses `GT; ISZERO`.

### Results

| solc | `== 0` gas | `<= 0` gas |
|---|---:|---:|
| 0.8.13 | 592 | 595 |
| 0.8.20 | 583 | 586 |
| 0.8.36 | 582 | 585 |

The three-gas difference creates a supplied-gas boundary where the original succeeds and the mutant runs out of gas. Optimized builds currently compile both expressions identically, but pruning cannot rely on optimizer-dependent equivalence when mutation testing uses the project’s actual compiler configuration.

This removes the source-level boundary-equivalence pruning while preserving the unsigned-negation filtering from #15962 and type-invalid comparison filtering from #15971. It also advances the mutation cache version for the restored mutant set.

### DIFF
--- a/.changelog/retain-gas-distinct-bounds.md
+++ b/.changelog/retain-gas-distinct-bounds.md
@@ -1,0 +1,5 @@
+---
+forge: patch
+---
+
+Retained comparison mutants that can compile to gas-distinct bytecode.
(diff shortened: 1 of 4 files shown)


figtracer (approved):

mattsse (approved):
lgtm

PR #16004, opened 2026-08-03
feat(forge): add test JSON file output

### LINKED ISSUE #2210: `forge test --json`: Add an optional `--json-file` argument
### Component

Forge

When running tests in a CI, it is useful to both show the human-readable output on stdout *and* write a JSON file to be processed at later stages by the CI. 

Right now, the only way to output both is either 1) to run `forge test` twice, or 2) to run `forge test --json | myProcessingTool`, where `myProcessingTool` writes a JSON file _and_ format the JSON back to human-readable output.

With `forge test --json-file path/to/file` you would get both outputs.

### Additional context

_No response_

figtracer:
Adds `--json-file <PATH>` to `forge test`, allowing CI to retain the normal human-readable console output while persisting structured test results for later processing. The file uses the same result normalization as `--json`, including log verbosity and trace-depth handling, and is emitted for both passing and failing test runs. Mutation and coverage modes reject the flag because they produce different result schemas.

Closes #2210.

This PR was written with Codex assistance.

### DIFF
--- a/crates/forge/src/cmd/test/mod.rs
+++ b/crates/forge/src/cmd/test/mod.rs
@@ -3290,8 +3290,18 @@ fn prepare_results_for_json(
             } else {
                 test_result.logs = Vec::new();
             }
-            if let Some(trace_depth) = trace_depth {
-                for (_, arena) in &mut test_result.traces {
+            for (_, arena) in &mut test_result.traces {
+                // Discard presentation-only decoding populated by the streaming renderer.
+                for node in arena.nodes_mut() {
+                    node.trace.decoded = None;
+                    for log in &mut node.logs {
+                        log.decoded = None;
+                    }
+                    for step in &mut node.trace.steps {
+                        step.decoded = None;
+                    }
+                }
+                if let Some(trace_depth) = trace_depth {
                     *arena = trace_arena_at_depth(arena, trace_depth);
                 }
             }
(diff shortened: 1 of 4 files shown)


mattsse (changes requested):
P2: `crates/forge/src/cmd/test/mod.rs:2088` clones the outcome after the human-output path has decoded rendered traces. As a result, `forge test -vvvv --json-file out.json` serializes populated `trace.decoded` fields, while the equivalent `forge test -vvvv --json` collector path leaves them null. A focused probe produced three decoded entries in the file and zero in stdout JSON, so the new file format depends on whether human rendering is active instead of matching `--json`. The current parity test misses this because combining `--json` and `--json-file` forces both through the collector, and the standalone case uses default verbosity. Please normalize the streaming result before writing and cover a trace-rendering verbosity.

mablr (dismissed):
lgtm

crates/forge/src/cmd/test/mod.rs
```diff
@@ -2067,6 +2084,12 @@ impl TestArgs {
             }
         }
 
+        if let Some(path) = &self.json_file {
+            let mut results = outcome.results.clone();
```
stevencartavia: **P2: Preserve completed in-flight suites in `--json-file`**

With `--fail-fast`, a concurrently running suite can complete but be omitted from the JSON file because the streaming receiver stops after the first failed suite. Adding `--json` retains that result through `test_collect`. Could we retain completed suite results for the file while preserving the current console behavior?

stevencartavia (approved):

mattsse (approved):
lgtm

PR #16007, opened 2026-08-03
perf(forge): speed up bytecode verification by name

figtracer:
## Summary

Resolve uniquely named `forge verify-bytecode` contracts to their source path and compile only that target and its imports instead of compiling the whole project. The fast path is limited to Solidity-only projects and contracts that are project inputs; missing or ambiguous names, Vyper projects, and dependency artifacts retain the existing whole-project compilation context.

The local-verification regression covers the fast path as well as dependency and Vyper fallbacks. Implemented with Codex.

## Benchmarks

Measured against an Aave v4 source tree with WETH9 deployed to local Anvil. Each cold sample ran `forge clean` before `forge verify-bytecode 0x5fbdb2315678afecb367f032d93f642f64180aa3 WETH9 --rpc-url http://127.0.0.1:18545 --ignore creation`.

| | Mean | Range |
| --- | ---: | ---: |
| master (`22931fe90`) | `1.566s ± 0.019s` | `1.553s–1.588s` |
| this PR | `69.2ms ± 2.4ms` | `67.1ms–75.6ms` |

This is about **22.6x faster**.

### DIFF
--- a/crates/verify/src/utils.rs
+++ b/crates/verify/src/utils.rs
@@ -97,8 +97,11 @@ pub fn build_project(
     let project = config.project()?;
     let compiler = ProjectCompiler::new().quiet(true);
 
-    if let Some(path) = args.contract.path() {
-        let target_path = canonicalize(project.root().join(path))?;
+    let target_path = match args.contract.path() {
+        Some(path) => Some(canonicalize(project.root().join(path))?),
+        None => project.find_contract_path(&args.contract.name).ok(),
+    };
+    if let Some(target_path) = target_path {
         let mut output = compiler.files([target_path.clone()]).compile(&project)?;
         let artifact =
             find_matching_contract_artifact(&mut output, &target_path, Some(&args.contract.name))?;
--- a/crates/verify/src/utils.rs
+++ b/crates/verify/src/utils.rs
@@ -16,7 +16,10 @@ use foundry_common::{
     ignore_metadata_hash, shell,
 };
 use foundry_compilers::{
+    Graph,
     artifacts::{BytecodeHash, CompactContractBytecode, EvmVersion},
+    compilers::ParsedSource,
+    multi::{MultiCompilerLanguage, MultiCompilerParser},
     utils::canonicalize,
 };
 use foundry_config::Config;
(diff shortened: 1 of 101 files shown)


crates/verify/src/utils.rs
```diff
@@ -97,8 +97,11 @@ pub fn build_project(
     let project = config.project()?;
     let compiler = ProjectCompiler::new().quiet(true);
 
-    if let Some(path) = args.contract.path() {
-        let target_path = canonicalize(project.root().join(path))?;
+    let target_path = match args.contract.path() {
+        Some(path) => Some(canonicalize(project.root().join(path))?),
+        None => project.find_contract_path(&args.contract.name).ok(),
```
mablr: Could we preserve the fallback when Vyper inputs are present? `find_contract_path` does not discover Vyper artifact names, so `Counter.sol` + `Counter.vy` is treated as unique and silently targets Solidity. Please gate this optimization to Solidity-only projects or fix Vyper name discovery.

mattsse (changes requested):
[P2] Preserve importer compiler constraints when targeting dependencies

Resolving a bare name and compiling the resulting file as a root input loses compiler constraints inherited from its importers. I reproduced this with `lib/dep/Target.sol` using `pragma solidity >=0.8.20 <0.9.0`, imported by project roots pinned to 0.8.20 and 0.8.25. The full project emits Target only under those two versions, and a deployed 0.8.25 artifact verifies as a full match through the pre-PR path. This change compiles Target independently with the newest compatible installed compiler (0.8.36) and reports `Runtime code did not match`.

Please preserve the importing project compilation context, or otherwise select one of the compiler versions actually resolved for the dependency. The added test deploys and verifies through target-only compilation, so it does not cover this case.

mablr (approved):

mattsse (approved):
lgtm

crates/verify/src/utils.rs
```diff
@@ -97,8 +100,26 @@ pub fn build_project(
...
+                node.data.contract_names().iter().any(|name| name == &args.contract.name)
+            });
+            let target = matches.next()?;
+            (matches.next().is_none()
+                && graph.input_nodes().any(|input| input.path() == target.path()))
+            .then(|| target.path().to_path_buf())
+        }),
+    };
+    if let Some(target_path) = target_path {
         let mut output = compiler.files([target_path.clone()]).compile(&project)?;
```
stevencartavia, line 123: The new bare-name fast path can change solc auto-detection. With `offline = true` and solc `0.8.20` and `0.8.30` installed, this reproduces using `Target.sol` with `pragma solidity >=0.8.20 <0.8.31` and a sibling input pinned to `0.8.20`: full-project compilation assigns `Target` to `0.8.20`, while `files([target_path])` assigns it to `0.8.30`. After deploying the full-project artifact, `master` reports a full runtime match, while this PR reports `Runtime code did not match`.

Could we retain full-project compilation for bare names when solc is not explicitly configured, or preserve the full graph’s compiler assignment for the target?

PR #16013, opened 2026-08-03
perf(cast): parallelize explicit-fee gas fill

figtracer:
## Summary

Estimate gas concurrently with nonce resolution when a Cast transaction already has a complete, valid set of explicit fee fields. Automatic and partial fee configurations are filled before gas estimation, so the estimate observes the same fee values as the final transaction. Automatic access-list generation, browser-wallet requests, Tempo transactions, and custom revert decoding retain their existing behavior.

AI assistance: Codex was used to investigate, benchmark, and implement this change.

## Benchmarks

Profiling builds against current `master` (`2f54078aa`) and a local Anvil RPC with a deterministic 200ms delay per request, using 2 warmups and 10 runs:

`cast mktx <TO> --legacy --gas-price 1 --rpc-url <RPC> --private-key <KEY>`

| | Mean | Range |
| --- | ---: | ---: |
| master | `651.8ms ± 7.1ms` | `642.1ms–661.8ms` |
| this PR | `441.8ms ± 6.3ms` | `429.8ms–447.6ms` |

This is about **1.48x faster**.

### DIFF
--- a/crates/anvil/src/config.rs
+++ b/crates/anvil/src/config.rs
@@ -201,7 +201,7 @@ pub struct NodeConfig {
     pub prune_history: PruneStateHistoryConfig,
     /// Max number of states cached on disk.
     pub max_persisted_states: Option<usize>,
-    /// The file where to load the state from
+    /// The initial state to apply and consume during startup.
     pub init_state: Option<SerializableState>,
     /// max number of blocks with transactions in memory
     pub transaction_block_keeper: Option<usize>,
--- a/crates/anvil/src/config.rs
+++ b/crates/anvil/src/config.rs
@@ -1401,48 +1401,36 @@ impl NodeConfig {
                 .wrap_err("failed to establish provider to fork url")?,
         );
 
-        // Fetch independent metadata while resolving the fork block.
-        let fetch_chain_id =
-            self.chain_id.is_none() && (self.fork_choice.is_none() || self.fork_chain_id.is_none());
-        let chain_id_task = tokio::spawn({
-            let provider = Arc::clone(&provider);
-            async move {
-                if fetch_chain_id { provider.get_chain_id().await.map(Some) } else { Ok(None) }
-            }
-        });
-        let fetch_gas_price = self.gas_price.is_none();
-        let gas_price_task = tokio::spawn({
-            let provider = Arc::clone(&provider);
-            async move {
-                if fetch_gas_price { provider.get_gas_price().await.map(Some) } else { Ok(None) }
-            }
-        });
-
-        let (fork_block_number, fork_transaction_replay, block) =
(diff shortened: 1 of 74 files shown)


mablr (dismissed):

crates/anvil/src/config.rs
```diff
@@ -1401,36 +1401,48 @@ impl NodeConfig {
...
-                .wrap_err("failed to get fork block number")?;
-            (bn, None, None)
+                .wrap_err("failed to get fork block")?
         };
-
-        let block = provider
-            .get_block(BlockNumberOrTag::Number(fork_block_number).into())
-            .await
-            .wrap_err("failed to get fork block")?;
+        let fork_chain_id = chain_id_task.await?.wrap_err("failed to fetch network chain ID")?;
```
stevencartavia: Could we check `block` for `None` before awaiting the metadata tasks? With a missing pinned block, the PR waited for a slow gas-price response, and with an explicit hardfork a chain-ID failure replaced the existing missing/non-archive diagnostic. Moving the block check above these awaits keeps the successful path concurrent while preserving the actionable error.

figtracer:
Superseded by #16007, which has been retargeted to this optimization.

crates/cast/src/tx.rs
```diff
@@ -613,7 +613,9 @@ where
             fill && self.auth.is_empty() && tempo_wallet.is_none() && !self.chain.is_tempo();
         let tx_nonce = if resolve_in_parallel {
             let nonce = self.tx.nonce();
-            let (tx_nonce, ()) = tokio::try_join!(
+            let gas_request = (self.access_list.is_none() && self.tx.gas_limit().is_none())
+                .then(|| self.tx.clone());
```
mablr: This snapshots the request before `fill_fees`. With only `--priority-gas-price`, `eth_estimateGas` receives a positive priority fee but no max fee; Anvil treats the missing max fee as zero and rejects an otherwise valid transaction. Gas estimation needs the completed fee pair.

crates/cast/src/tx.rs
```diff
@@ -623,7 +625,18 @@ where
                     self.browser,
                     self.eip1559_fee_estimate,
                 ),
+                async {
+                    match gas_request {
+                        Some(request) => {
+                            Self::estimate_gas(&self.provider, request).await.map(Some)
```
mablr, line 640: Estimating before automatic fees are applied changes execution semantics: fee-sensitive contracts see a different `tx.gasprice` here than in the final transaction, which can revert or produce an unsafe gas limit. Please estimate from the fee-filled request.

stevencartavia (approved):

mablr (approved):

PR #16012, opened 2026-08-03
fix(cast): restore browser wallet signing

mablr:
## Motivation
Restore `cast wallet sign --browser` for personal messages and dynamic EIP-712 typed data. Browser message signing was removed with the unused generic signer implementations in #13613 + #13657.

## Solution

This uses the new explicit `BrowserSigner` APIs instead of restoring the generic `Signer`, `SignerSync`, or `TxSigner` implementations. Raw hash signing remains unsupported for browser wallets and is rejected before starting the browser server. Existing quiet, verbose, and JSON output modes use the connected browser-wallet address.

## Examples

Personal-message signing:

```sh
cast wallet sign "hello from cast" --browser
```

EIP-712 signing:

```sh
cast wallet sign --browser --data \
  '{"types":{"EIP712Domain":[{"name":"name","type":"string"},{"name":"version","type":"string"},{"name":"chainId","type":"uint256"}],"Message":[{"name":"contents","type":"string"}]},"primaryType":"Message","domain":{"name":"Cast browser test","version":"1","chainId":1},"message":{"contents":"hello from EIP-712"}}'
```

PR #16015, opened 2026-08-03
fix(cast): complete browser wallet restoration

mablr:
Follow-up to #16012, completing the restoration of browser-wallet support across Cast. This restores browser-wallet address resolution for `cast wallet address`, `cast call`, `cast access-list`, and `cast estimate`.

These commands unintentionally lost `--browser` when #13613 extracted `BrowserSigner` from `WalletSigner` and moved browser options out of the shared `WalletOpts`. They do not require generic message or transaction signing; they only need the connected wallet address when constructing their RPC requests.

This also pins `foundry-wallets` to foundry-rs/foundry-core#146. That dependency update recovers the signer from personal-message and EIP-712 signatures returned by the browser and rejects signatures that do not match the connected wallet address, instead of accepting them based only on their encoding.

This PR was written with Codex assistance, including implementation, tests, and dependency integration.

### DIFF
--- a/.changelog/restore-cast-browser-wallet-readonly-commands.md
+++ b/.changelog/restore-cast-browser-wallet-readonly-commands.md
@@ -1,0 +1,5 @@
+---
+cast: patch
+---
+
+Restored browser wallet address resolution for `cast wallet address`, `cast call`, `cast access-list`, and `cast estimate`.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6066,7 +6066,7 @@ dependencies = [
 [[package]]
 name = "foundry-wallets"
 version = "0.1.0"
-source = "git+https://github.com/foundry-rs/foundry-core?rev=886503a096f15c26e13dfddf2b6330508c97b3bd#886503a096f15c26e13dfddf2b6330508c97b3bd"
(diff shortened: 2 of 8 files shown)


figtracer (dismissed):

grandizzy (dismissed):

mattsse (dismissed):
Two runtime paths still need handling before browser support is complete.

[P2] crates/cast/src/cmd/estimate.rs:104-105 — Preserve the browser context for Tempo gas estimation. With cast estimate --tempo --browser --create, this branch keeps only the address, and the later provider.estimate_gas(tx) receives the ordinary request instead of browser_wallet_gas_estimation_request(). That omits the conservative WebAuthn signer-size hints and Tempo CREATE-to-call conversion, so the reported gas and --cost can model a different transaction or underestimate a passkey wallet gas. The new help test only checks that the flag is exposed.

[P2] crates/cast/src/cmd/call.rs:149-150 — Handle or reject --browser together with --curl. run() returns through run_curl() before the browser option is resolved, and that path only populates from from self.wallet.from. Consequently, cast call --browser --curl succeeds but emits a request without the connected wallet address, so contracts that inspect msg.sender behave differently from the corresponding non-curl browser call.

stevencartavia (dismissed):

stevencartavia (dismissed):

mablr:
Sorry @stevencartavia I had to re-pin after foundry-core pr merge 😅

grandizzy (approved):
lgtm

stevencartavia (approved):

PR #16017, opened 2026-08-03
fix(fuzz): reliably synchronize worker corpus

stevencartavia:
Removes timestamp-based corpus synchronization so delayed worker findings are not permanently skipped. Uses inbox presence and explicit pending-entry tracking while preserving startup distribution and corpus file compatibility.

### DIFF
--- a/crates/evm/evm/src/executors/corpus.rs
+++ b/crates/evm/evm/src/executors/corpus.rs
@@ -224,6 +224,41 @@ fn same_tx_sequence(left: &[BasicTxDetails], right: &[BasicTxDetails]) -> bool {
         })
 }
 
+fn accept_synced_corpus_file(
+    entry: &CorpusDirEntry,
+    tx_seq: &[BasicTxDetails],
+    corpus_path: &Path,
+) -> bool {
+    if corpus_path.is_file() {
+        let existing = CorpusDirEntry {
+            path: corpus_path.to_path_buf(),
+            uuid: entry.uuid,
+            timestamp: entry.timestamp,
+        };
+        if existing.read_tx_seq().is_ok_and(|existing| same_tx_sequence(&existing, tx_seq)) {
+            if let Err(err) = std::fs::remove_file(&entry.path) {
+                debug!(target: "corpus", %err, "failed to remove synced corpus link {}", entry.path.display());
+                return false;
+            }
+            return true;
+        }
+
+        warn!(target: "corpus", "not overwriting conflicting corpus file {}", corpus_path.display());
+        let quarantine_path =
+            entry.path.with_file_name(format!("{}.{}.invalid", entry.name(), Uuid::new_v4()));
+        if let Err(err) = std::fs::rename(&entry.path, &quarantine_path) {
+            debug!(target: "corpus", %err, "failed to quarantine conflicting corpus file {}", entry.path.display());
+        }
+        return false;
+    }
+
+    if let Err(err) = std::fs::rename(&entry.path, corpus_path) {
+        debug!(target: "corpus", %err, "failed to move synced corpus from {:?} to {corpus_path:?} dir", entry.path);
+        return false;
+    }
+    true
+}
+
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub(crate) enum CorpusInsertionMode {
     Live,
@@ -283,6 +318,7 @@ pub(crate) struct WorkerCorpusSeed {
     edge_indices: EdgeIndexMap,
     sancov_history_map: Vec<u8>,
     metrics: CorpusMetrics,
+    replay_dirs: Option<Vec<PathBuf>>,
     failed_replays: usize,
     optimization_best_value: Option<I256>,
     optimization_best_sequence: Vec<BasicTxDetails>,
@@ -343,6 +379,7 @@ impl WorkerCorpusSeed {
             edge_indices: self.edge_indices.clone(),
             sancov_history_map: self.sancov_history_map.clone(),
             metrics,
+            replay_dirs: self.replay_dirs.clone(),
             failed_replays: self.failed_replays,
             optimization_best_value: self.optimization_best_value,
             optimization_best_sequence: self.optimization_best_sequence.clone(),
@@ -375,6 +412,8 @@ impl WorkerCorpusSeed {
         let Some(corpus_dir) = &config.corpus_dir else {
             return Ok(seed);
         };
+        let replay_dirs = canonical_replay_dirs(corpus_dir);
+        seed.replay_dirs = Some(replay_dirs.clone());
 
         // Seed in-memory corpus with the persisted optimization best sequence so the mutation
         // engine can build on it in future runs.
@@ -395,7 +434,7 @@ impl WorkerCorpusSeed {
         };
         let mut seen_entries =
             seed.in_memory_corpus.iter().map(|entry| entry.uuid).collect::<HashSet<_>>();
-        for entry in unique_corpus_entries(&canonical_replay_dirs(corpus_dir), &mut seen_entries) {
+        for entry in unique_corpus_entries(&replay_dirs, &mut seen_entries) {
             // A corrupt or truncated corpus file (e.g. a process killed mid-write, since entries
             // are persisted non-atomically) must not abort the whole campaign startup: skip it
             // and keep loading the rest of the corpus.
@@ -584,8 +623,8 @@ pub struct WorkerCorpus {
     config: Arc<FuzzCorpusConfig>,
     /// Indices of new entries added to [`WorkerCorpus::in_memory_corpus`] since last sync.
     new_entry_indices: Vec<usize>,
-    /// Whether the master still needs to distribute corpus entries loaded at startup.
-    initial_export_pending: bool,
+    /// Corpus directories the master loaded at startup and still needs to distribute.
+    initial_export_dirs: Option<Vec<PathBuf>>,
     /// Worker Dir
     /// corpus_dir/worker1/
     worker_dir: Option<PathBuf>,
@@ -812,8 +851,9 @@ impl WorkerCorpus {
         id: usize,
         config: FuzzCorpusConfig,
         sequence_generator: SequenceGenerator,
-        seed: WorkerCorpusSeed,
+        mut seed: WorkerCorpusSeed,
     ) -> Result<Self> {
+        let initial_export_dirs = if id == 0 { seed.replay_dirs.take() } else { None };
         let worker_dir = config.corpus_dir.as_ref().map(|corpus_dir| {
             let worker_dir = corpus_dir.join(format!("{WORKER}{id}"));
             let worker_corpus = worker_dir.join(CORPUS_DIR);
@@ -838,7 +878,7 @@ impl WorkerCorpus {
             current_mutated_index: None,
             config: config.into(),
             new_entry_indices: Default::default(),
-            initial_export_pending: id == 0,
+            initial_export_dirs,
             worker_dir,
             last_sync_metrics: Default::default(),
             optimization_best_value: seed.optimization_best_value,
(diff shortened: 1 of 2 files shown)


crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1343,9 +1369,15 @@ impl WorkerCorpus {
 
         let worker_dir = self.worker_dir.as_ref().unwrap();
         let master_corpus_dir = worker_dir.join(CORPUS_DIR);
-        let filtered_master_corpus = read_corpus_dir(&master_corpus_dir)
-            .filter(|entry| entry.timestamp > self.last_sync_timestamp)
-            .collect::<Vec<_>>();
+        let entries = if self.initial_export_pending {
+            read_corpus_dir(&master_corpus_dir).map(|entry| entry.path).collect::<Vec<_>>()
```
mablr: **Blocker:** Startup replay loads persisted entries from every `worker*/corpus`, but this initial fan-out scans only `worker0/corpus`. A finding persisted by worker 1 after its last sync is restored into worker 0 on the next run, but is never distributed to the other workers because restored entries are not queued in `new_entry_indices`. Could initial export use the same deduplicated canonical corpus set as startup replay?

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1750,6 +1780,107 @@ mod tests {
...
+        assert!(corpus_root.join("worker2").join(SYNC_DIR).join(name).is_file());
+    }
+
+    #[test]
+    fn master_distributes_startup_corpus_only_once() {
+        let corpus_root = temp_corpus_dir();
+        let mut master = empty_worker_corpus(0, corpus_root.clone());
+        let corpus = CorpusEntry::new(vec![basic_tx()]);
+        let name = corpus.file_name(false);
+        corpus.write_to_disk_in(&corpus_root.join("worker0").join(CORPUS_DIR), false).unwrap();
```
mablr, line 2017: Could we cover the synchronization lifecycle in one grouped regression test (or test group): startup data under a non-master corpus directory, delayed old-timestamp arrival relayed through master to another worker, partial fan-out failure/retry semantics, and a finding persisted after the final periodic sync? The current tests exercise the helpers but miss these interleavings.

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1360,18 +1392,19 @@ impl WorkerCorpus {
...
+                    debug!(target: "corpus", %err, from=?path, to=?sync_path, "failed to distribute corpus");
                     continue;
                 }
                 any_distributed = true;
-                trace!(target: "corpus", %name, ?target_dir, "distributed corpus");
+                trace!(target: "corpus", name=%name.to_string_lossy(), ?target_dir, "distributed corpus");
             }
         }
 
+        self.initial_export_pending = false;
```
mablr: **Non-blocking:** This marks initial export complete even if one or more hard links failed, so a transient partial failure is never retried. Consider retaining pending state for failed destinations (`AlreadyExists` can count as delivered).

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1214,6 +1218,9 @@ impl WorkerCorpus {
             };
             if tx_seq.is_empty() {
                 warn!(target: "corpus", "skipping empty corpus entry: {}", entry.path.display());
+                if let Err(err) = std::fs::remove_file(&entry.path) {
```
mablr, line 1295: **Non-blocking:** With the timestamp watermark removed, an unreadable file remains in the inbox and is retried (and warned about) on every sync. Should malformed entries be removed or quarantined like empty entries?

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1450,12 +1483,9 @@ impl WorkerCorpus {
             self.export_to_master()?;
         }
 
-        let last_sync = SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs();
-        self.last_sync_timestamp = last_sync;
-
         self.new_entry_indices.clear();
 
-        debug!(target: "corpus", last_sync, "synced");
+        debug!(target: "corpus", "synced");
```
mablr, line 1624: **Non-blocking:** There is still no final synchronization before workers return, so findings created after the last periodic sync cannot help peers in the current campaign. This is worth addressing if the intended guarantee is synchronization before campaign completion.

crates/evm/evm/src/executors/corpus.rs
stevencartavia: [3169e14](https://github.com/foundry-rs/foundry/pull/16017/commits/3169e14b97476b7553f4ad7a3184cd28760e0336)

crates/evm/evm/src/executors/corpus.rs
stevencartavia: [3169e14](https://github.com/foundry-rs/foundry/pull/16017/commits/3169e14b97476b7553f4ad7a3184cd28760e0336)

crates/evm/evm/src/executors/corpus.rs
stevencartavia: follow-up

crates/evm/evm/src/executors/corpus.rs
stevencartavia: follow-up

crates/evm/evm/src/executors/corpus.rs
stevencartavia: partially addressed, tied to the follow-ups comments 3 and 5

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1450,12 +1533,9 @@ impl WorkerCorpus {
             self.export_to_master()?;
         }
 
-        let last_sync = SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs();
-        self.last_sync_timestamp = last_sync;
-
         self.new_entry_indices.clear();
```
grandizzy: **Blocking:** This clears every pending entry even when an individual `hard_link` failed, because both export methods log link errors but still return `Ok(())`. A transient worker→master failure therefore permanently drops that entry from synchronization; master fan-out can likewise miss only one target worker. Please let each export remove only entries delivered to every intended destination. On retries, `AlreadyExists` should count as delivered only after verifying that the destination contains the same sequence.

Suggested shape:
```rust
let mut delivered = HashSet::new();
for &index in &self.new_entry_indices {
    // Link the entry to every intended destination.
    // Record it only if every link succeeds or already has matching content.
    if delivered_to_all_targets(index)? {
        delivered.insert(index);
    }
}
self.new_entry_indices.retain(|index| !delivered.contains(index));
```

The unconditional clear here can then be removed.

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1266,8 +1336,7 @@ impl WorkerCorpus {
             if keep_entry && new_coverage {
                 // Move file from sync/ to corpus/ directory.
                 let corpus_path = corpus_dir.join(sync_path.components().next_back().unwrap());
-                if let Err(err) = std::fs::rename(sync_path, &corpus_path) {
-                    debug!(target: "corpus", %err, "failed to move synced corpus from {sync_path:?} to {corpus_path:?} dir");
+                if !accept_synced_corpus_file(&entry, &tx_seq, &corpus_path) {
```
grandizzy, line 1364: **Blocking:** Replay has already committed new coverage to `history_map` before this file operation. If acceptance fails, the inbox file remains, but its next replay reports no new coverage and deletes it, permanently losing the sequence that produced the edge. Could we accept/claim the file before mutating coverage, or replay against temporary coverage state and commit only after acceptance succeeds?

For example, replay into cloned state and assign it only after acceptance:
```rust
let mut history_map = self.history_map.clone();
let mut edge_indices = self.edge_indices.clone();
let mut sancov_history_map = self.sancov_history_map.clone();
let mut metrics = self.metrics.clone();

let coverage = ReplayCoverage {
    history_map: &mut history_map,
    edge_indices: &mut edge_indices,
    sancov_history_map: &mut sancov_history_map,
    metrics: Some(&mut metrics),
};
// replay using `coverage`

if !accept_synced_corpus_file(&entry, &tx_seq, &corpus_path) {
    continue; // real coverage state remains unchanged, so this can retry
}
self.history_map = history_map;
self.edge_indices = edge_indices;
self.sancov_history_map = sancov_history_map;
self.metrics = metrics;
```
A claim/staging-file design could avoid the cloning cost.

crates/evm/evm/src/executors/corpus.rs
```diff
@@ -1343,9 +1416,18 @@ impl WorkerCorpus {
 
         let worker_dir = self.worker_dir.as_ref().unwrap();
         let master_corpus_dir = worker_dir.join(CORPUS_DIR);
-        let filtered_master_corpus = read_corpus_dir(&master_corpus_dir)
-            .filter(|entry| entry.timestamp > self.last_sync_timestamp)
-            .collect::<Vec<_>>();
+        let entries = if let Some(replay_dirs) = &self.initial_export_dirs {
```
grandizzy: **Blocking:** The first sync exports either startup paths or `new_entry_indices`, not both. With a flat legacy corpus, `initial_export_dirs` is `[root]`; a new finding created under the newly initialized `worker0/corpus` before the first sync is omitted, then its pending index is cleared. Please export a deduplicated union of startup and pending paths.

Suggested replacement shape:
```rust
let mut entries = Vec::new();
if let Some(replay_dirs) = &self.initial_export_dirs {
    let mut seen_entries = HashSet::new();
    entries.extend(
        unique_corpus_entries(replay_dirs, &mut seen_entries).map(|entry| entry.path),
    );
}
entries.extend(
    self.new_entry_indices
        .iter()
        .filter_map(|&index| self.in_memory_corpus.get(index))
        .map(|corpus| master_corpus_dir.join(corpus.file_name(self.config.corpus_gzip))),
);
let mut seen_names = HashSet::new();
entries.retain(|path| {
    path.file_name().is_some_and(|name| seen_names.insert(name.to_owned()))
});
```
This should be combined with retaining startup/pending state until fan-out completes.

grandizzy (dismissed):
The timestamp race is addressed, but pending work can still be lost through export failures, failed file acceptance, and the first-sync startup branch. These interleavings should be fixed before merging.

crates/evm/evm/src/executors/corpus.rs
stevencartavia: [df26031](https://github.com/foundry-rs/foundry/pull/16017/commits/df26031d34fc9a57b50b3235af1f169e7588f5c3)

crates/evm/evm/src/executors/corpus.rs
stevencartavia: [df26031](https://github.com/foundry-rs/foundry/pull/16017/commits/df26031d34fc9a57b50b3235af1f169e7588f5c3)

crates/evm/evm/src/executors/corpus.rs
stevencartavia: [df26031](https://github.com/foundry-rs/foundry/pull/16017/commits/df26031d34fc9a57b50b3235af1f169e7588f5c3)

mablr (approved):
Ack for the follow-ups, lgtm.

grandizzy (approved):
lgtm

figtracer (approved):

crates/evm/evm/src/executors/corpus.rs
stevencartavia: #16058

crates/evm/evm/src/executors/corpus.rs
stevencartavia: #16058

crates/evm/evm/src/executors/corpus.rs
stevencartavia: #16058

PR #16020, opened 2026-08-04
fix(fuzz): report only failure logs

stevencartavia:
Ensures failed fuzz tests report logs from the counterexample-producing input and worker, without including unrelated successful-case logs.


mattsse (approved):
lgtm

grandizzy (approved):
lgtm

PR #16023, opened 2026-08-04
fix(anvil): reduce state load memory

### LINKED ISSUE #9562: Anvil OOM-killed on startup with `--state`
### Component

Anvil

- [X] Foundry
- [X] Foundryup

nightly-75fc63be4fc9241a1981a55c12b6e300fd82a51b

_No response_

### Operating System

Linux

### Describe the bug

We've run anvil for a few weeks with interval mining, such that the state is a few GB. 
On node updates that restart anvil, it gets OOM-killed while trying to restore the current state. 
In principle it should be possible to stream from the state file with backpressure so that it doesn't fall over - not sure if this constitutes a bug or a feature request, but the program getting nuked is unexpected

stevencartavia:
Anvil currently reads the complete state file into memory and retains or clones the deserialized state several times during startup. This switches state-file loading to buffered deserialization and moves each state component into its final backend destination, reducing avoidable peak memory while preserving the state format and loading order.

Closes #9562.

### DIFF
--- a/.changelog/reduce-anvil-state-load-memory.md
+++ b/.changelog/reduce-anvil-state-load-memory.md
@@ -1,0 +1,5 @@
+---
+anvil: patch
+---
+
+Reduced peak memory usage when Anvil loads state files.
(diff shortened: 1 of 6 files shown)


grandizzy (approved):
lgtm

mattsse (approved):
lgtm

tomekmb-glitch:
- [ ] Cr4dQ9tFBSsNzMXFSrmNwszba2WpNpHYF1z7sQgConsU

Issue #9782, opened 2025-01-29
fuzz testing missing important case with 2 struct inputs


moodysalem:
### Component

Forge

### Have you ensured that all of these are up to date?

- [x] Foundry
- [x] Foundryup

### What version of Foundry are you on?

0.3.0 (5a8bd89 2024-12-20T08:45:53.204298000Z)

### What version of Foundryup are you on?

0.3.1

### What command(s) is the bug in?

forge test

### Operating System

macOS (Apple Silicon)

### Describe the bug

fuzzer seems to be missing very obvious test cases when given 2 of the same struct input, seems like it's passing 2 similar structs far too often

with 1 million test runs this test passes that checks that no 2 structs have the same hash, even when the hash function is intentionally broken to leave out 1 of the fields

code (intentionally broken to fail the test by missing extension from the hash function)

```solidity
struct PoolKey {
    address token0;
    address token1;
    uint128 fee;
    uint32 tickSpacing;
    address extension;
}

function toPoolId(PoolKey memory key) pure returns (bytes32 result) {
    result = keccak256(abi.encode(key.token0, key.token1, key.fee, key.tickSpacing));
}
```

test:
```solidity
    function test_toPoolId_aligns_with_eq(PoolKey memory pk0, PoolKey memory pk1) public pure {
        bytes32 pk0Id = pk0.toPoolId();
        bytes32 pk1Id = pk1.toPoolId();

        assertEq(
            pk0.token0 == pk1.token0 && pk0.token1 == pk1.token1 && pk0.fee == pk1.fee
                && pk0.tickSpacing == pk1.tickSpacing && pk0.extension == pk1.extension,
            pk0Id == pk1Id
        );
    }
```

passes with 1 million runs!

```
Ran 1 test for test/types/keys.t.sol:KeysTest
[PASS] test_toPoolId_aligns_with_eq((address,address,uint128,uint32,address),(address,address,uint128,uint32,address)) (runs: 1000006, μ: 5131, ~: 5131)
```

grandizzy:
@moodysalem your test can be written as

```solidity
        bool cond1 = pk0.token0 == pk1.token0 && pk0.token1 == pk1.token1 && pk0.fee == pk1.fee
                && pk0.tickSpacing == pk1.tickSpacing && pk0.extension == pk1.extension;
        bool cond2 = pk0Id == pk1Id;

        assertEq(cond1, cond2);
```

[...]

PR #16024, opened 2026-08-04
test(forge): cover omitted struct hash field w/ symbex

mablr:
## Motivation

Adds a symbolic execution regression covering omitted struct fields in derived hashes. The test reproduces a case random fuzzing is unlikely to reach and confirms symbolic execution finds the mismatched keys directly.

See #9782


grandizzy (approved):
ty!

mattsse (approved):
lgtm

PR #16026, opened 2026-08-04
feat(cheatcodes): resolve artifact remappings

### LINKED ISSUE #11446: Support remappings in deployCode / deployCodeTo
### Component

Forge

Hey team, right now, the `string memory what` parameter for the Std cheats `deployCode` and `deployCodeTo` does not apply remappings, which I think may be useful. 

I’d like to be able to pass the exact same paths I already use for imports in my contracts.

Example, suppose I'm building a library and I have the following remapping:

`@organization/library/=/src/`

Such that when my library contracts imports content from it's own, I can consistently do:

`import "@organization/library/Something.sol";`

I’d like to be able to use that same form in tests:

`deployCodeTo("@organization/library/Something.sol", addr);`

instead of having to fall back to raw path:

`deployCodeTo("src/Something.sol", addr);`

This would make tests more consistent with contract code and reduce friction when working across projects with heavy remapping usage.

Thanks!

### Additional context

_No response_

figtracer:
Artifact path lookup currently matches Solidity paths directly against compiled artifact sources, so import-style paths such as `@organization/library/Something.sol` cannot be used with `vm.getCode`, native `vm.deployCode`, or the forge-std deployment helpers built on `vm.getCode`.

This resolves artifact paths through the project's existing remapping resolver before matching them, using the running test or script as context so contextual remappings work as well. The artifact lookup unit coverage and Solidity `DeployCode` fixture cover the remapped path end to end.

Fixes #11446.

AI assistance disclosure: Codex investigated, implemented, and tested this change, and may generate responses to review feedback.

### DIFF
--- a/.changelog/deploy-code-remappings.md
+++ b/.changelog/deploy-code-remappings.md
@@ -1,0 +1,5 @@
+---
+forge: minor
+---
+
+Added support for project remappings in artifact paths passed to `vm.getCode` and `vm.deployCode`.
--- a/crates/cheatcodes/src/fs.rs
+++ b/crates/cheatcodes/src/fs.rs
@@ -579,9 +579,19 @@ fn get_artifact_source<'a, FEN: FoundryEvmNetwork>(
             .as_ref()
             .and_then(|artifact| artifact.source.parent())
             .unwrap_or(&state.config.paths.root);
-        state.config.paths.resolve_library_import(cwd, &file).map_or(file, |resolved| {
-            resolved.strip_prefix(&state.config.paths.root).unwrap_or(&resolved).to_path_buf()
-        })
+        let relative_cwd = cwd.strip_prefix(&state.config.paths.root).unwrap_or(cwd);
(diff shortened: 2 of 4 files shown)


mablr (dismissed):

mattsse (dismissed):
P2 — Preserve non-remapped artifact paths (crates/cheatcodes/src/fs.rs:582-584)

resolve_library_import also falls back to searching configured library directories when no remapping matches. With a project artifact src/Thing.sol:RootThing and a configured library artifact lib/src/Thing.sol:LibThing, vm.getCode("src/Thing.sol:RootThing") previously selects the project artifact; this change rewrites the file to lib/src/Thing.sol, excludes the project artifact, and fails with "no matching artifact found." The same regression affects getDeployedCode, getSelectors, and deployCode through the shared resolver.

Please limit this resolution to explicit remapping matches (or otherwise preserve existing project-relative artifact paths) and add coverage for a colliding library path.

mablr (approved):

stevencartavia (approved):

PR #16027, opened 2026-08-04
fix(anvil): make mining finalization atomic

mablr:
## Description

Closes OSS-601

Anvil previously executed ordinary mining directly against live state and assumed block finalization could not fail. A failing Prague system transition or malformed deposit request could therefore partially apply state while producing no valid block, consume one-shot timestamp or prevrandao controls, and cause manual mining RPCs to report success.

This executes complete candidates against an isolated database overlay and publishes state, block metadata, history, fees, notifications, pool changes, and time controls only after finalization succeeds. Mining errors now propagate through the manual RPC paths. Background mining retains the live pool while discarding stale candidate selections, and generation checks prevent a failed candidate from modifying newer miner modes, timestamp overrides, prevrandao overrides, or concurrent time resets.

The regression coverage exercises both post-block failure modes, verifies state and metadata rollback against a clean node, checks all manual mining RPCs and one-shot controls, and covers fresh pool reselection plus fixed and mixed mining recovery.

### DIFF
--- a/.changelog/anvil-atomic-mining-finalization.md
+++ b/.changelog/anvil-atomic-mining-finalization.md
@@ -1,0 +1,5 @@
+---
+anvil: patch
+---
+
+Made block mining atomic when finalization fails and return errors from manual mining RPCs.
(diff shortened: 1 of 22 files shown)


mattsse (approved):
lgtm

figtracer (approved):

PR #16033, opened 2026-08-05
fix(fuzz): replay runs after rejected inputs

stevencartavia:
Track generated fuzz inputs separately from accepted executions so persisted `fuzz_run` coordinates remain aligned with the deterministic input stream after `vm.assume` rejections. This allows `--fuzz-run` to reproduce the original failing calldata.


mablr (approved):

mattsse (approved):
lgtm

PR #16034, opened 2026-08-05
fix(fuzz): disable zero invariant delays

stevencartavia:
Treat zero invariant time and block delay limits as disabled. This prevents modulo-by-zero panics and keeps delay detection consistent with transaction generation.


mablr (approved):

mattsse (approved):
lgtm

PR #16036, opened 2026-08-05
fix(anvil): return sync timeout data

### LINKED ISSUE #16035: bug: Anvil does not return correct error code when timeouting for eth_sendRawTransactionSync
### Component

Anvil

- [x] Foundry
- [x] Foundryup

1.8.0-nightly

1.9.2

_No response_

### Operating System

Linux

### Describe the bug

While preparing a pull request for [alloy-rs](https://github.com/alloy-rs/alloy/) implementing the possibility of eth_sendRawTransactionSync with timeout (see https://github.com/alloy-rs/alloy/issues/4041) I found out that Anvil's error response to the call does not conform to [EIP-7966](https://eips.ethereum.org/EIPS/eip-7966#error-codes-and-response-structure).

Example as a test for alloy:
```rust
#[tokio::test]
async fn test_send_raw_transaction_sync_with_timeout_fail() {
    let provider = ProviderBuilder::new().connect_anvil_with_wallet();

    // Create a transaction
    let tx = TransactionRequest {
        nonce: Some(0),
        value: Some(U256::from(100)),
        to: Some(address!("d8dA6BF26964aF9D7eEd9e03E53415D37aA96045").into()),
        gas_price: Some(20e9 as u128),
        gas: Some(21000),
        ..Default::default()
    };

    // Build and sign the transaction to get the envelope
    let tx_envelope = tx.build(&provider.wallet()).await.expect("failed to build tx");

    // Encode the transaction
    let encoded = tx_envelope.encoded_2718();

    // Send using the sync method - this has to return ErrorResp
    let rpc_error = provider
        .send_raw_transaction_sync_with_timeout(&encoded, 1)
        .await
        .expect_err("timeout should fire");

    let error_payload = rpc_error
        .as_error_resp()
        .expect("error should be ErrorResp");

    // Verify error code
    assert_eq!(error_payload.code, 4);
    // Data has to contain transaction hash
    let data = error_payload.data.as_ref().expect("data should contain tx hash");
    let data_str = data.get().trim_matches('"').trim();
    let tx_hash: B256 = data_str.parse().expect("data should contain tx hash");
    assert!(tx_hash != B256::ZERO, "error payload data should have valid tx hash");
}
```

The response is expected to look something like:
```json
{ 
  "code": 4,
  "message": "Transaction confirmation timeout",
  "data": "0x123....."
}
```

But instead it looks like:
```json
{
  "code": -32603,
  "message": "Transaction confirmation timeout"
}
```

stevencartavia:
Fixes #16035 by returning EIP-7966 error code `4` and the transaction hash when synchronous transaction confirmation times out. Adds regression coverage for the JSON-RPC response.

### DIFF
--- a/.changelog/anvil-sync-timeout-error.md
+++ b/.changelog/anvil-sync-timeout-error.md
@@ -1,0 +1,5 @@
+---
+anvil: patch
+---
+
+Return error code 4 and the transaction hash when synchronous transaction confirmation times out.
--- a/crates/anvil/src/eth/error/mod.rs
+++ b/crates/anvil/src/eth/error/mod.rs
@@ -451,9 +451,11 @@ impl<T: Serialize> ToRpcResponseResult for Result<T> {
                 BlockchainError::ChainIdNotAvailable => {
                     RpcError::invalid_params("Chain Id not available")
                 }
-                BlockchainError::TransactionConfirmationTimeout { .. } => {
-                    RpcError::internal_error_with("Transaction confirmation timeout")
-                }
+                BlockchainError::TransactionConfirmationTimeout { hash, .. } => RpcError {
+                    code: ErrorCode::ServerError(4),
+                    message: "Transaction confirmation timeout".into(),
+                    data: Some(serde_json::Value::String(hash.to_string())),
+                },
                 BlockchainError::InvalidTransaction(err) => match err {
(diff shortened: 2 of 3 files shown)


mattsse (approved):

mattsse (approved):
lgtm

PR #16041, opened 2026-08-06
perf(forge): speed up coverage line analysis

figtracer:
## Summary

Build coverage line entries by walking the already-sorted coverage items once instead of rescanning every item for every executable line. This changes line construction from quadratic to linear in the number of coverage items while preserving the first matching source location for each line.

The stress fixture's 2.3 MB LCOV report is byte-for-byte identical before and after the change. AI assistance: Amp was used to investigate, benchmark, and implement this change.

## Benchmarks

Release builds from the same `master` commit, measured with Hyperfine (one warmup and three runs) on a 1.9 MiB generated Solidity source containing 200,000 executable statements plus one smoke test:

```text
forge coverage --report summary --no-match-coverage test/
```

| | Mean | Range |
| --- | ---: | ---: |
| master | `12.491s ± 0.067s` | `12.447s–12.568s` |
| this PR | `752.9ms ± 4.2ms` | `749.3ms–757.6ms` |

This is **16.59x faster**.

### DIFF
--- a/.changelog/forge-coverage-line-analysis.md
+++ b/.changelog/forge-coverage-line-analysis.md
@@ -1,0 +1,5 @@
+---
+forge: patch
+---
+
+Speed up `forge coverage` analysis for sources with many executable lines.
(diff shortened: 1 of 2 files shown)


mablr (approved):

mattsse (approved):
lgtm

PR #16048, opened 2026-08-06
feat(forge): add locked build mode

### LINKED ISSUE #16044: feat: foundry.lock divergence checker
### Component

Forge

There's currently no way to enforce in CI that a committed foundry.lock matches the effective submodule commits. forge build warns on a mismatch but exits 0, and forge install / forge update treat the lock as the source of truth and move submodules to match it, so neither can be used as a read-only verify.

Could Foundry add a non-mutating check (e.g. forge lock --check, or a --locked / --frozen flag) that exits non-zero on drift and reports which dependencies differ, like npm ci or cargo build --locked?

or arguably forge build etc should even fail by default when they diverge.

mablr:
## Motivation

Closes #16044.

Adds `forge build --locked`, which fails when `foundry.lock` diverges from the project’s direct Git dependency submodules.

Strict validation runs before automatic dependency installation or compilation, making it suitable as a deterministic CI gate without changing the lockfile or dependency checkout state.

Ordinary forge build continues to report inconsistencies as warnings, while `--locked` exits unsuccessfully for missing or stale lock entries, revision drift, uninitialized submodules, missing `.gitmodules` mappings, and conflicted gitlinks. Conflict diagnostics avoid reporting an arbitrary index-stage revision.


mattsse (dismissed):
lgtm

decofe:
There is existing logic that can be reused here:

- `crates/forge/src/cmd/build.rs::BuildArgs::check_foundry_lock_consistency` already reads `foundry.lock`, iterates its entries, resolves the checked-out revision with `Git::get_rev("HEAD", ...)`, and compares it with the locked revision.
- `crates/forge/src/lockfile.rs::Lockfile` already owns lockfile parsing and `DepIdentifier::rev()` exposes the expected revision.

Could the common comparison/discovery logic move into `Lockfile` (returning structured mismatches), with `forge build` formatting those as warnings and `forge lock --check` returning an error? That would preserve the broader missing/stale/uninitialized/conflicted coverage from this PR without duplicating the existing revision-check path. `Lockfile::sync()` is also related, but should not be called by the new check because it can update/write state.

grandizzy:
@mablr pls check https://github.com/foundry-rs/foundry/pull/16048#issuecomment-5207749768
Also, why not just enforcing on forge build as proposed in ticket?

crates/forge/src/lockfile.rs
```diff
@@ -115,6 +175,77 @@ impl<'a> Lockfile<'a> {
...
+                    mismatches.push(LockfileMismatch::Revision {
+                        path: path.clone(),
+                        expected: expected.to_string(),
+                        actual: submodule.rev().to_string(),
+                    });
+                }
+                SubmoduleCheckoutStatus::Current | SubmoduleCheckoutStatus::Modified => {}
+            }
+        }
+        mismatches.extend(submodules.into_iter().map(|(path, submodule)| {
```
stevencartavia: `Lockfile::check` hides `Uninitialized` and `Conflicted` when the submodule is also missing from `foundry.lock`. Reproduced this with both a fresh `git clone --no-recurse` and a genuine submodule merge conflict. In the conflict case, the reported “found” SHA is index stage 3, even though the path has no resolved stage-0 gitlink.

Could we preserve the checkout-state diagnostic when the lock entry is missing, or report both mismatches?

crates/forge/src/lockfile.rs
```diff
@@ -115,6 +181,107 @@ impl<'a> Lockfile<'a> {
...
+            {
+                return Ok(Vec::new());
+            }
+            Err(err) => return Err(err),
+        };
+        let project_prefix = project_root.strip_prefix(&git_root).map_err(|_| {
+            eyre::eyre!("Project root is not contained in Git root {}", git_root.display())
+        })?;
+
+        let git = git.root(&git_root);
```
stevencartavia: `Lockfile::check` re-roots Git commands from the project directory to the repository root. This breaks valid relative `GIT_DIR` and `GIT_WORK_TREE` values because Git resolves them against the command’s working directory.

Reproduced with both a separate Git directory and a linked worktree: direct Git succeeds, Forge with absolute paths succeeds, but `forge build --locked` with equivalent relative paths fails with `not a git repository`.

Could we preserve the original CWD or resolve these environment paths before re-rooting?

mablr: https://github.com/foundry-rs/foundry/pull/16048/commits/0a9924e3c706a010c8c315eabaf981c41d2a5a1e

crates/forge/src/lockfile.rs
```diff
@@ -115,6 +181,107 @@ impl<'a> Lockfile<'a> {
...
+            {
+                return Ok(Vec::new());
+            }
+            Err(err) => return Err(err),
+        };
+        let project_prefix = project_root.strip_prefix(&git_root).map_err(|_| {
+            eyre::eyre!("Project root is not contained in Git root {}", git_root.display())
+        })?;
+
+        let git = if project_prefix.as_os_str().is_empty() { *git } else { git.root(&git_root) };
```
stevencartavia: The relative Git environment issue still occurs when the Foundry project is nested below the Git worktree root. In that case, `project_prefix` is non-empty, so Git is still re-rooted and relative `GIT_DIR` / `GIT_WORK_TREE` values are resolved from the wrong directory.

Reproduced at `0a9924e3c` with both a separate Git directory and a linked worktree: direct Git and Forge with equivalent absolute paths succeed, while Forge with relative paths fails with `not a git repository`.

The updated test only covers the project-is-Git-root case. Could we also preserve the original environment paths for nested projects?

stevencartavia (approved):

mattsse (approved):

MathisGD:
thanks for tackling this. at the end is it enforced optional?

mablr:
@MathisGD Yes, strict enforcement is opt-in via forge build `--locked`, similar to Cargo. A regular forge build remains non-breaking and should only warn about inconsistencies.

PR #16049, opened 2026-08-06
fix(cast): encode tuple event parameters

mattsse:
`cast abi-encode-event` parsed each event input from its JSON ABI `type` string directly. Tuple inputs are represented by the placeholder `tuple`, so coercion failed without resolving the component types. This change resolves complete event parameter types before coercion, emits word-sized indexed values directly, and ABI-encodes non-indexed values as one event body sequence.

This allows tuple events such as [`ActionLogged((uint256,function))`](https://eth-sepolia.blockscout.com/tx/0x62d8f0260cfb7b5687183f3ec67b878e7b7300c9c8e55ea7b22f23de5a9da5be) to round-trip through Cast, including structs containing 24-byte external function pointers. Regression coverage also preserves explorer-style JSON ABI interface generation and Forge trace decoding for the original Solidity syntax.

This PR was authored with OpenAI Codex assistance, including investigation, implementation, tests, and PR text.


crates/cast/src/lib.rs
```diff
@@ -2017,44 +2017,42 @@ impl SimpleCast {
...
-                    // For fixed-size types, encode directly to 32 bytes
-                    let mut encoded = [0u8; 32];
-                    let token_encoded = token.abi_encode();
-                    if token_encoded.len() <= 32 {
-                        let start = 32 - token_encoded.len();
-                        encoded[start..].copy_from_slice(&token_encoded);
-                    }
-                    topics.push(B256::from(encoded));
+                    // Complex indexed values are represented by their encoded hash.
+                    topics.push(keccak256(token.abi_encode()));
```
mattsse: This predates the PR, so this is non-blocking, but this fallback still hashes regular ABI encoding rather than the Solidity [special indexed-event preimage](https://docs.soliditylang.org/en/latest/abi-spec.html#encoding-of-indexed-event-parameters). That leaves indexed `string`, `bytes`, and dynamic or nested arrays/tuples with topics that cannot match emitted logs; for example indexed `"hello"` should be `0x1c8aff950685c2ed4bc3174f3472287b56d9517b9c948127319a09a7a36deac8`, while the current test locks in `0x984002…`.

I recommend routing indexed values through a helper along these lines:

```rust
fn encode_event_topic(value: &DynSolValue) -> B256 {
    if let Some(word) = value.as_word() {
        return word;
    }

    let mut preimage = Vec::new();
    encode_event_topic_preimage(value, false, &mut preimage);
    keccak256(preimage)
}
```

The recursive helper should emit word values directly; use raw contents for top-level `string`/`bytes`; and recursively concatenate `Array`, `FixedArray`, and `Tuple` members, right-padding nested `string`/`bytes` to 32 bytes without offsets or length words. Please update the existing indexed-string expectation and add `(uint256,string) indexed` plus `uint256[] indexed` cases—the latter two distinguish the special preimage from `abi_encode()`.

mattsse (commented):
Non-blocking follow-up noted inline.

mablr (approved):

figtracer (approved):

stevencartavia (approved):

PR #16050, opened 2026-08-06
fix(mutation): filter invalid assignments

mablr:
Assignment mutations currently replace values with `0` or their negation without considering the destination type. This produces mutants that cannot compile, such as assigning `0` or a negated value to an address, boolean, enum, array, or struct.

This change uses Solar’s resolved destination types to exclude replacements that are invalid under Solidity 0.8 semantics. It handles both variable initializers and plain assignments while preserving valid replacements for numeric and fixed-bytes destinations. User-defined value types, unresolved types, compound assignments, and incomplete semantic analysis fail open so mutation testing does not discard replacements unless their invalidity is established.

AI assistance disclosure: Amp was used to review the implementation, identify edge cases, and assist with code and test generation.

### DIFF
--- a/crates/forge/src/mutation/mod.rs
+++ b/crates/forge/src/mutation/mod.rs
@@ -430,7 +430,7 @@ impl MutationHandler {
         let mut mutant_cfg_hasher = DefaultHasher::new();
         // Version salt for this mutant-set cache schema. Bump this if the
         // inputs that define generated mutants change.
-        "mutant-set-v5".hash(&mut mutant_cfg_hasher);
+        "mutant-set-v6".hash(&mut mutant_cfg_hasher);
         for op in self.config.mutation.enabled_operators() {
             op.to_string().hash(&mut mutant_cfg_hasher);
         }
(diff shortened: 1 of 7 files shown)


mattsse (dismissed):
lgtm

crates/forge/src/mutation/mod.rs
```diff
@@ -430,7 +430,7 @@ impl MutationHandler {
         let mut mutant_cfg_hasher = DefaultHasher::new();
         // Version salt for this mutant-set cache schema. Bump this if the
         // inputs that define generated mutants change.
-        "mutant-set-v5".hash(&mut mutant_cfg_hasher);
+        "mutant-set-v6".hash(&mut mutant_cfg_hasher);
```
stevencartavia, line 433: The mutant cache key doesn’t include the current type-analysis exclusions. Reproduced a stale cache: fixing an unrelated Solar-only diagnostic left the target’s build ID unchanged, so the cache-retained run still reported two invalid assignment mutants; moving `cache/mutation` aside reduced the invalid count to zero. Could we include the exclusions in the cache fingerprint or reapply them to cached mutants?

figtracer (approved):

stevencartavia (approved):

PR #16051, opened 2026-08-06
fix(mutation): filter non-lvalue increments

mablr:
This PR is stacked on #16050.

Unary mutation currently generates prefix and postfix increment or decrement operators for operands that cannot be assigned to, such as literals, binary expressions, casts, and ordinary function results. Those mutants are guaranteed to fail compilation and add noise to mutation runs.

This change filters increment and decrement replacements only when the operand is structurally known not to be an lvalue. The classification fails open for unknown or erroneous expression forms and preserves Solidity’s assignable zero-argument dynamic storage-array `push()` result, including parenthesized and indexed call expressions. This keeps valid mutants while avoiding compilation work for definitively invalid replacements.

AI assistance disclosure: Amp was used to review the implementation, identify edge cases, and assist with code and test generation.

### DIFF
--- a/crates/forge/src/mutation/mutators/unary_op_mutator.rs
+++ b/crates/forge/src/mutation/mutators/unary_op_mutator.rs
@@ -112,11 +112,12 @@ impl Mutator for UnaryOpMutator {
 }
 
 fn is_definitely_non_lvalue(expr: &solar::ast::Expr<'_>) -> bool {
-    if let ExprKind::Call(callee, _) = &expr.peel_parens().kind {
-        return !matches!(
-            &callee.peel_parens().kind,
-            ExprKind::Member(_, member) if member.as_str() == "push"
-        );
+    if let ExprKind::Call(callee, args) = &expr.peel_parens().kind {
+        return !(args.is_empty()
+            && matches!(
+                &callee.peel_parens().kind,
+                ExprKind::Member(_, member) if member.as_str() == "push"
+            ));
     }
 
     matches!(
--- a/crates/forge/src/mutation/mutators/unary_op_mutator.rs
+++ b/crates/forge/src/mutation/mutators/unary_op_mutator.rs
@@ -19,12 +19,12 @@ impl Mutator for UnaryOpMutator {
 
         let expr = context.expr.unwrap();
 
-        let (op, target_span) = match &expr.kind {
-            ExprKind::Unary(un_op, target) => (un_op.kind, target.span),
+        let (op, target) = match &expr.kind {
+            ExprKind::Unary(un_op, target) => (un_op.kind, &**target),
             _ => unreachable!(),
         };
 
-        let target_content = extract_span_text(context.source.unwrap_or(""), target_span);
+        let target_content = extract_span_text(context.source.unwrap_or(""), target.span);
         if target_content.is_empty() {
             return Ok(vec![]);
         }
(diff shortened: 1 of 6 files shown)


crates/forge/src/mutation/mutators/unary_op_mutator.rs
```diff
@@ -104,6 +111,32 @@ impl Mutator for UnaryOpMutator {
     }
 }
 
+fn is_definitely_non_lvalue(expr: &solar::ast::Expr<'_>) -> bool {
+    if let ExprKind::Call(callee, _) = &expr.peel_parens().kind {
+        return !matches!(
+            &callee.peel_parens().kind,
+            ExprKind::Member(_, member) if member.as_str() == "push"
+        );
+    }
```
mattsse: Only the zero-argument storage-array built-in can return an lvalue, but this exemption applies to every member call named `push`. For example, `-producer.push(1)` is valid when an ordinary `push(int256)` method returns `int256`, yet this path retains `++producer.push(1)` and the other increment/decrement mutants, all of which fail compilation. Please require an empty argument list before exempting the call; the current tests only cover zero-argument storage `push()`.

```suggestion
if let ExprKind::Call(callee, args) = &expr.peel_parens().kind {
    return !(args.is_empty()
        && matches!(
            &callee.peel_parens().kind,
            ExprKind::Member(_, member) if member.as_str() == "push"
        ));
}
```

mattsse (changes requested):
One invalid-mutant case remains; details inline.

crates/forge/src/mutation/mutators/unary_op_mutator.rs
mablr: https://github.com/foundry-rs/foundry/pull/16051/commits/01c0325830709eb755f410dbb3e9c437e647e918

crates/forge/src/mutation/mutators/unary_op_mutator.rs
```diff
@@ -55,6 +55,10 @@ impl Mutator for UnaryOpMutator {
         mutations = operations
             .into_iter()
             .filter(|&kind| kind != op)
+            .filter(|&kind| {
```
figtracer, line 58: please bump the mutant-set cache salt with this generation change: a cache populated by #16050 is reused here and retains the four invalid `-foo()` increment/decrement mutants.

crates/forge/src/mutation/mutators/unary_op_mutator.rs
```diff
@@ -104,6 +111,33 @@ impl Mutator for UnaryOpMutator {
     }
 }
 
+fn is_definitely_non_lvalue(expr: &solar::ast::Expr<'_>) -> bool {
+    if let ExprKind::Call(callee, args) = &expr.peel_parens().kind {
+        return !(args.is_empty()
+            && matches!(
+                &callee.peel_parens().kind,
+                ExprKind::Member(_, member) if member.as_str() == "push"
```
figtracer, line 119: `-this.push()` still hits this exemption for an ordinary zero-argument method, retaining four increment/decrement mutants that fail with `expression has to be an lvalue`.

crates/forge/src/mutation/mutators/unary_op_mutator.rs
stevencartavia: +1

mattsse (changes requested):
The mutant-set cache salt still needs to be bumped before approval. With `mutant-set-v6` unchanged, a cache populated by #16050 is loaded before regeneration and retains the invalid unary mutants this PR is intended to remove.

crates/forge/src/mutation/mutators/unary_op_mutator.rs
mablr: Clankers are a bit nervous about that, as this is merged together with #16050, no risk.

mattsse (dismissed):
lgtm

figtracer (dismissed):

stevencartavia (dismissed):

stevencartavia (approved):

mattsse (approved):
lgtm

PR #16053, opened 2026-08-06
fix(cast): reject trailing call fields

stevencartavia:
Rejects trailing fields after raw calldata in Cast batch call specifications instead of silently ignoring them. This prevents malformed inputs from being signed or broadcast as shortened calls.

### DIFF
--- a/crates/cast/src/call_spec.rs
+++ b/crates/cast/src/call_spec.rs
@@ -67,10 +67,14 @@ impl CallSpec {
         // Pattern: to:value:sig:args or to::sig:args (empty value) or to:value:0xdata
         let mut idx = 1;
 
-        // Check for value (non-empty, not starting with 0x unless it's a number)
+        // Check for value (non-empty and not a signature). A terminal lowercase hex field is
+        // treated as raw calldata, while one followed by another field is a value.
         if idx < parts.len() {
             let part = parts[idx];
-            if !part.is_empty() && !part.starts_with("0x") && !part.contains('(') {
+            if !part.is_empty()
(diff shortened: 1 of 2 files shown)


mattsse (dismissed):
lgtm

crates/cast/src/call_spec.rs
```diff
@@ -85,9 +85,12 @@ impl CallSpec {
             let part = parts[idx];
             if part.starts_with("0x") {
                 // Raw calldata
-                data = Some(Bytes::from(
-                    hex::decode(part).map_err(|e| eyre!("Invalid hex data '{}': {}", part, e))?,
-                ));
+                let decoded =
+                    hex::decode(part).map_err(|e| eyre!("Invalid hex data '{}': {}", part, e))?;
+                if idx + 1 != parts.len() {
```
mablr, line 94: This also rejects lowercase hex values in an otherwise valid value-plus-signature form. For example, `to:0x10:deposit()` is unambiguous and `parse_ether_or_wei` supports `0x10`, but the value-dispatch condition above leaves it to this calldata branch, so `deposit()` becomes an unexpected trailing field. Could the dispatch treat lowercase hex as a value when another field follows, while preserving terminal `to:0x10` as raw calldata? A parser-level regression test for `to:0x10:deposit()` would cover the path that `test_parse_hex_value` currently misses.

stevencartavia: [386495f](https://github.com/foundry-rs/foundry/pull/16053/commits/386495f0fa37beb85ae759a0db45b810efb11221)

mablr (approved):

figtracer (approved):

PR #9689, opened 2025-01-16
fix(config): normalize optimizer settings

grandizzy:
## Motivation
Closes #9688 

## Solution
- move optimizer settings normalization in helper fn and reuse when generating configs
- add tests for default configs and optimizer options

PR #16055, opened 2026-08-06
fix(verify-bytecode): report the default optimizer runs, not "unknown"

mkzung:
## Motivation

When `forge verify-bytecode` fails to match, `find_mismatch_in_settings` prints hints about which compiler settings differ. For `optimizer_runs` it prints the local value as `unknown` and reports a mismatch whenever the on-chain value is non-zero:

```rust
if local_settings.optimizer_runs.is_some_and(|runs| etherscan_settings.runs != runs as u64)
    || (local_settings.optimizer_runs.is_none() && etherscan_settings.runs > 0)
```

A project that never sets `optimizer_runs` compiles with 200. `Config::normalize_optimizer_settings` is what turns the unset value into `Some(200)`, and the verify path never calls it, so the field is still `None` here. A contract compiled with the default 200 therefore gets:

```
Optimizer runs mismatch: local=unknown, onchain=200
```

There is no mismatch. The message appears only after verification has already failed, which is exactly when a false lead is most expensive: the reader goes looking for a settings problem that does not exist.

## Solution

Normalize a copy of the config before comparing, rather than hardcoding 200 in a second place where it could drift from `normalize_optimizer_settings`.

## Testing

Two unit tests in `crates/verify/src/utils.rs`, no network needed. Both fail on `master`:

```
left:  ["Optimizer runs mismatch: local=unknown, onchain=999"]
right: ["Optimizer runs mismatch: local=200, onchain=999"]

assertion failed: find_mismatch_in_settings(&onchain, &local).is_empty()
```

`cargo clippy -p forge-verify` is clean.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

## AI disclosure

This PR was written with Claude Code, including the code, tests and this description. I reviewed and verified every line, ran the tests against both the old and new behaviour, and am responsible for the change.

### DIFF
--- a/crates/verify/src/utils.rs
+++ b/crates/verify/src/utils.rs
@@ -206,6 +206,12 @@ fn find_mismatch_in_settings(
     etherscan_settings: &Metadata,
     local_settings: &Config,
 ) -> Vec<String> {
+    // The verify path never calls `normalize_optimizer_settings`, so an unset `optimizer_runs`
+    // arrives here as `None` and used to be reported as "unknown", making every contract built
+    // with the default look like a mismatch. Normalize a copy rather than hardcoding the default,
+    // so the two cannot drift.
+    let local_settings = &local_settings.clone().normalized_optimizer_settings();
+
     let mut mismatches: Vec<String> = vec![];
     if etherscan_settings.evm_version != local_settings.evm_version.to_string().to_lowercase() {
         let str = format!(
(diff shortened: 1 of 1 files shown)


mattsse:
Thanks for the PR, but the premise doesn't hold on master: `VerifyBytecodeArgs::run` loads its config via [`self.load_config()`](https://github.com/foundry-rs/foundry/blob/50dd4b8accae878391b0825885f0632d1e0a1e4c/crates/verify/src/bytecode.rs#L160) → [`load_config_from_provider`](https://github.com/foundry-rs/foundry/blob/50dd4b8accae878391b0825885f0632d1e0a1e4c/crates/cli/src/utils/cmd.rs#L177-L179) → `Config::from_provider` → `from_figment_inner`, which has called [`normalize_optimizer_settings()`](https://github.com/foundry-rs/foundry/blob/50dd4b8accae878391b0825885f0632d1e0a1e4c/crates/config/src/lib.rs#L937) since #9689. By the time `find_mismatch_in_settings` runs, an unset `optimizer_runs` is already `Some(200)`, so `local=unknown` can't be printed by `forge verify-bytecode` — the [`is_none()` branch](https://github.com/foundry-rs/foundry/blob/50dd4b8accae878391b0825885f0632d1e0a1e4c/crates/verify/src/utils.rs#L226-L227) this removes is dead code in the CLI path. The unit tests fail on master only because they construct `Config` by hand, bypassing config loading.

If you can show an actual CLI repro on master, happy to reopen. As a pure dead-code cleanup this would also be acceptable, but then the title, description, and the added comment need to say that rather than describing a user-facing bug.

PR #16056, opened 2026-08-07
fix(verify-bytecode): fail when provided constructor args do not match the deployment

### LINKED ISSUE #11102: forge verify-bytecode should verify full runtime code including immutables

mattsse:
## Motivation

`forge verify-bytecode` reports a match for a contract even when the supplied constructor args are not the ones used at deployment. In #11102, five zero addresses verify clean against a contract whose immutables are all non-zero.

Two defects combine to make the args unfalsifiable. Supplied args that are not the tail of the onchain creation code are silently replaced with the real tail, which also feeds the real args into the runtime simulation, so even the full runtime comparison (immutables included) passes. And `is_partial_match` strips `constructor_args.len()` bytes from both sides before comparing, without ever checking that the onchain code actually ends with those args, so wrong args of the right length can never fail the creation comparison either.

## Solution

Fix it at the comparison layer: `is_partial_match` now requires the onchain creation code to end with the constructor args before stripping them. In `verify-bytecode`, user-supplied args are kept instead of substituted (with a warning), while explorer-derived args retain the existing substitution workaround for Etherscan returning incorrect args. With the wrong args kept, the full match fails on the differing tail and the partial match fails on the new check, so the mismatch surfaces through the normal reporting paths, including `--json`.

This supersedes #16054, which fixes the same bug by special-casing the `verify-bytecode` call site and forcing the match result to `None`. Checking the invariant inside the comparison covers all creation-path callers and needs no extra state. Credit to @mkzung for the diagnosis and the regression test; the commit lists them as co-author.

Behaviour changes for anyone passing constructor args that do not match the deployment: that now reports a mismatch instead of a partial match.

A unit test covers the comparison change directly, and `flaky_verify_bytecode_warns_on_wrong_constructor_args` (adapted from #16054) exercises the CLI path against the deployed `StrategyManager`.

Fixes #11102

### DIFF
--- a/crates/verify/src/bytecode.rs
+++ b/crates/verify/src/bytecode.rs
@@ -697,17 +697,18 @@ impl VerifyBytecodeArgs {
             false,
             config.bytecode_hash,
         );
-        if args_from_user && creation_match_type.is_none() {
+        if args_from_user
+            && creation_match_type.is_none()
+            && self.ignore.is_none_or(|b| !b.is_creation())
+        {
             let message =
                 "Provided constructor args could not be validated against deployment creation code";
             if shell::is_json() {
-                if self.ignore.is_none_or(|b| !b.is_creation()) {
-                    json_results.push(JsonResult {
-                        bytecode_type: BytecodeType::Creation,
-                        match_type: None,
-                        message: Some(message.to_string()),
-                    });
-                }
+                json_results.push(JsonResult {
+                    bytecode_type: BytecodeType::Creation,
+                    match_type: None,
+                    message: Some(message.to_string()),
+                });
                 if self.ignore.is_none_or(|b| !b.is_runtime()) {
                     json_results.push(JsonResult {
                         bytecode_type: BytecodeType::Runtime,
@@ -718,15 +719,13 @@ impl VerifyBytecodeArgs {
                 sh_println!("{}", serde_json::to_string(&json_results)?)?;
             } else {
                 sh_warn!("{message}")?;
-                if self.ignore.is_none_or(|b| !b.is_creation()) {
-                    crate::utils::print_result(
-                        None,
-                        BytecodeType::Creation,
-                        &mut json_results,
-                        etherscan_metadata,
-                        &config,
-                    );
-                }
+                crate::utils::print_result(
+                    None,
+                    BytecodeType::Creation,
+                    &mut json_results,
+                    etherscan_metadata,
+                    &config,
+                );
                 if self.ignore.is_none_or(|b| !b.is_runtime()) {
                     crate::utils::print_result(
                         None,
--- a/crates/verify/src/bytecode.rs
+++ b/crates/verify/src/bytecode.rs
@@ -4,6 +4,7 @@ use crate::{
     utils::{
         BytecodeType, JsonResult, check_and_encode_args, check_explorer_args,
         load_fork_config_and_evm_opts, maybe_predeploy_contract, synthetic_deployment_context,
+        validate_encoded_constructor_args,
     },
     verify::VerifierArgs,
 };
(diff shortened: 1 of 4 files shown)


crates/verify/src/utils.rs
```diff
@@ -187,6 +187,13 @@ fn is_partial_match(
         return try_extract_and_compare_bytecode(local_bytecode, bytecode);
     }
 
+    // The constructor args are part of what is being verified: the onchain creation code must
+    // actually end with them before they can be stripped, otherwise args of the right length
+    // could never fail the comparison.
+    if !bytecode.ends_with(constructor_args) {
```
stevencartavia, line 194: `--encoded-constructor-args` can still produce a false match if the contract’s metadata is prepended to the supplied arguments.

The suffix check passes because that metadata already appears immediately before the real constructor arguments in the deployment transaction. After the suffix and remaining metadata are stripped, the bytecodes compare equal even though different constructor data was supplied.

Reproduced using solc 0.8.26 bytecode and the PR’s actual matcher. Deploying both inputs on Anvil also produced identical runtime bytecode but different constructor-initialized storage.

Could encoded arguments be ABI-validated or could the matcher prevent them from overlapping the contract metadata?

crates/verify/src/bytecode.rs
```diff
@@ -551,8 +552,15 @@ impl VerifyBytecodeArgs {
         // happens, try extracting arguments ourselves.
         if !maybe_creation_code.ends_with(&constructor_args) {
             trace!("mismatch of constructor args with etherscan");
-            // If local bytecode is longer than on-chain one, this is probably not a match.
-            if maybe_creation_code.len() >= local_bytecode.len() {
+            if args_from_user {
+                // Keep user-supplied args: substituting them with the onchain tail would make
+                // the creation comparison vacuous and let a deployment with different
+                // constructor args (and immutables) verify clean.
+                if !shell::is_json() {
```
mablr: Could we make a user-supplied mismatch terminal here? With `--ignore creation`, a constructor that only initializes storage can still report `Runtime code matched`, and JSON suppresses this warning entirely.

marco-jardim:
@mattsse I reproduced #11102 and reviewed the two outstanding comments on this PR. I have a focused supplemental commit based directly on `9d5526295` that adds the required changelog, makes mismatched user-supplied arguments terminal (including with `--ignore creation` and `--json`), canonical-validates `--encoded-constructor-args`, and replaces the network-dependent regression test with deterministic local coverage.

Would you be comfortable with me opening a draft PR targeting your `mattsse/verify-bytecode-user-ctor-args` branch, keeping your commit and authorship intact? If you prefer, I can instead share the commit for you to cherry-pick. The supplemental patch and this message were prepared with LLM assistance; the patch was validated locally.

crates/verify/src/bytecode.rs
stevencartavia: [e9fed9c](https://github.com/foundry-rs/foundry/pull/16056/commits/e9fed9c2d8e819a735b4014cdeae66f1210f4834)

crates/verify/src/utils.rs
stevencartavia: [e9fed9c](https://github.com/foundry-rs/foundry/pull/16056/commits/e9fed9c2d8e819a735b4014cdeae66f1210f4834)

mablr (dismissed):
Addressed the remaining review findings in two commits. User-provided constructor arguments are now validated against the complete deployment creation code, even with --ignore creation, preventing dynamic ABI suffixes from producing false matches. Typed and file-based arguments are also rejected when the contract has no constructor, consistent with encoded-argument validation.

SGTM now

stevencartavia (dismissed):

crates/verify/src/bytecode.rs
```diff
@@ -682,28 +687,73 @@ impl VerifyBytecodeArgs {
...
+        // be a suffix of a different valid encoding. Always compare the complete creation code
+        // when arguments came from the user, even if creation output is ignored.
+        let creation_match_type = crate::utils::match_bytecodes(
+            local_bytecode_vec.as_slice(),
+            &maybe_creation_code,
+            &constructor_args,
+            false,
+            config.bytecode_hash,
+        );
+        if args_from_user && creation_match_type.is_none() {
```
figtracer: could we avoid making this terminal when `--ignore creation` is set? a constructor that only initializes storage can use different args while producing identical runtime bytecode, so this branch reports a runtime mismatch without comparing runtime. keeping the supplied args and running the runtime simulation still catches the immutable mismatch from #11102 while preserving the flag's semantics.

figtracer: fixed in [`c55b722`](https://github.com/foundry-rs/foundry/commit/c55b722477955d8fc2cdb75747aaff62f1bb2948): `--ignore creation` now falls through to the existing runtime comparison with the supplied args, while normal creation verification remains terminal. focused verifier tests and the live mainnet regression pass.

mablr (approved):

stevencartavia (approved):

PR #16058, opened 2026-08-07
fix(forge): finalize fuzz corpus sync

stevencartavia:
Follow-up to #16017, this adds an ordered final corpus drain for parallel fuzz campaigns. After mutation stops, non-master workers export pending findings, the master calibrates them and completes fan-out, and peers calibrate their final inboxes before the campaign returns. This prevents findings created after the last periodic sync from remaining stranded on one worker.

The final drain preserves partial fan-out retry semantics and fails rather than reporting successful completion with pending entries. A grouped lifecycle regression covers non-master startup data, old-timestamp arrivals, partial fan-out failure and retry, and a late finding relayed through the master to another worker.

### DIFF
--- a/.changelog/fuzz-final-corpus-sync.md
+++ b/.changelog/fuzz-final-corpus-sync.md
@@ -1,0 +1,6 @@
+---
+forge: patch
+foundry-evm: patch
+---
+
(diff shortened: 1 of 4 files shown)


mattsse (approved):
lgtm

mablr (approved):

PR #16059, opened 2026-08-07
fix(anvil): add transaction status RPC

### LINKED ISSUE #13352: bug(anvil): improve UX on Anvil's side and user's side when a transaction is dropped from the mempool
### Component

Anvil

We currently get stuck into a transactionReceipt fetching loop on the user side and the user has no feedback that their transaction has been dropped.

### Additional context

_No response_

stevencartavia:
Anvil now exposes `anvil_getTransactionStatus` for querying whether a transaction is pending, mined, recently dropped, or unknown while preserving the standard receipt-or-null behavior of `eth_getTransactionReceipt`. The bounded dropped-transaction index covers explicit drops, invalidation, clearing, and ready or queued replacements without introducing an additional canonical source of truth.

Fixes #13352.

This PR was developed with AI assistance.

### DIFF
--- a/.changelog/anvil-dropped-transaction-receipts.md
+++ b/.changelog/anvil-dropped-transaction-receipts.md
@@ -2,4 +2,4 @@
 anvil: patch
 ---
 
-Anvil now reports when an accepted transaction was dropped from the mempool instead of leaving receipt consumers polling indefinitely.
+Added `anvil_getTransactionStatus` to report whether a transaction is pending, mined, or recently dropped without changing standard receipt behavior.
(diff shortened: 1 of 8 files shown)


mattsse (commented):
this deviates from what regular clients do and I think we should not do this

mablr (changes requested):
Yes I agree this is an architectural footgun because it makes ephemeral mempool state influence the canonical `eth_getTransactionReceipt` endpoint, creating competing sources of truth, race conditions with in-flight mining, and history-dependent behavior when dropped-hash tombstones expire. We sould keep receipt lookup canonical (receipt or null) and expose dropped/replaced status through an Anvil-specific RPC or subscription instead, wdyt?

mattsse:
imo i dont think this is a useful feature so maybe we just close this pr and the issue as wont fix

PR #16063, opened 2026-08-07
perf(anvil): speed up block receipt queries

figtracer:
## Summary

- build local block receipts from one coherent storage snapshot instead of rerunning the single-transaction receipt path for every transaction
- assign cumulative log indices in one pass, removing the repeated receipt-prefix scans that made `eth_getBlockReceipts` quadratic
- preserve the legacy lookup path for inconsistent block/transaction mappings loaded through `anvil_loadState`

## Performance

Benchmarked release builds with a loaded Anvil state containing one block with 10,000 transactions and one log per receipt. The measured request was `eth_getBlockReceipts("latest")`, with the 13 MiB response redirected to `/dev/null`.

| | Mean |
|---|---:|
| `master` | 17.098 s ± 0.182 s |
| this PR | 47.7 ms ± 0.2 ms |

This is about **358× faster**. The full JSON responses were byte-for-byte identical (SHA-256 `9968ce09d038a05739f0e5a9e097b9cbcf0ec9f654866dcbb414f473212bb58f`).

## Validation

- `cargo nextest run --package anvil --test it 'logs::'`
- `cargo check --package anvil`
- `cargo clippy --package anvil --all-targets -- -D warnings`
- `python3 .github/scripts/validate_changelog.py --base origin/master --head HEAD`

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -5689,25 +5689,20 @@ where
 
     /// Returns all transaction receipts of the block
     pub fn mined_block_receipts(&self, id: impl Into<BlockId>) -> Option<Vec<FoundryTxReceipt>> {
-        let (block, transactions, hash) = {
-            let storage = self.blockchain.storage.read();
-            let hash = match id.into() {
-                BlockId::Hash(hash) => hash.block_hash,
-                BlockId::Number(number) => storage.hash(number, self.slots_in_an_epoch)?,
-            };
-            let block = storage.blocks.get(&hash)?.clone();
-            let transactions = block
-                .body
-                .transactions
-                .iter()
-                .map(|transaction| storage.transactions.get(&transaction.hash()).cloned())
-                .collect::<Option<Vec<_>>>()?;
-            (block, transactions, hash)
+        let storage = self.blockchain.storage.read();
+        let hash = match id.into() {
+            BlockId::Hash(hash) => hash.block_hash,
+            BlockId::Number(number) => storage.hash(number, self.slots_in_an_epoch)?,
         };
+        let block = storage.blocks.get(&hash)?.clone();
 
-        if transactions.iter().enumerate().any(|(index, transaction)| {
-            transaction.block_hash != hash || transaction.info.transaction_index as usize != index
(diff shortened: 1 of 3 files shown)


mattsse (dismissed):
lgtm

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5689,12 +5689,43 @@ where
...
+            let hash = match id.into() {
+                BlockId::Hash(hash) => hash.block_hash,
+                BlockId::Number(number) => storage.hash(number, self.slots_in_an_epoch)?,
+            };
+            let block = storage.blocks.get(&hash)?.clone();
+            let transactions = block
+                .body
+                .transactions
+                .iter()
+                .map(|transaction| storage.transactions.get(&transaction.hash()).cloned())
```
mablr: This clones every `MinedTransaction` up front, including each transaction’s complete trace arena, so peak memory scales with all traces in the block and a public RPC call could cause a large spike or OOM. Maybe let's borrow the transactions while building the fast-path receipts, or snapshot only the receipt fields needed.

mablr (commented):
The performance improvement reproduces locally: 24.94 s on the merge-base versus 54.5 ms on this branch, with identical 14 MB responses. But it seems there is a memory footgun.

crates/anvil/src/eth/backend/mem/mod.rs
figtracer: Agreed — this was cloning the complete trace arena for every transaction. I changed the fast path to keep the coherent storage read snapshot and borrow each `MinedTransaction`, cloning only its receipt immediately before conversion.

I also checked RSS over three fresh-process runs with a stress block containing 2,000 transactions and 101 trace nodes per transaction (202,000 trace nodes total):

- before: 147.3 MB idle / 241.1 MB peak / 93.8 MB request delta
- after: 147.4 MB idle / 170.8 MB peak / 23.4 MB request delta

That reduces the request-time RSS increase by about 75%. The receipt response remained byte-identical, and the original 10,000-receipt fixture now runs in 42.4 ms ± 0.6 ms. Pushed in 713de96dd.

mablr (approved):

mattsse (approved):

PR #16064, opened 2026-08-07
feat(fmt): use nearest project configs

### LINKED ISSUE #5686: `forge fmt` with nearest `foundry.toml`
### Component

Forge

We have a big monorepo with Solidity scattered in various packages, example dirs, templates, etc. We'd ideally like to run `forge fmt` across the entire repo (`forge fmt **/*.sol`), but use whatever the nearest `foundry.toml` is for formatting (so certain paths get ignored, etc.). Is this possible to expose via CLI option?

In the short term, we can work around this by having a root `foundry.toml` that is aligned with the same `[fmt]` config as the rest of them, but a little trickier to maintain context-specific ignore lists that way.

### Additional context

_No response_

figtracer:
## Motivation

`forge fmt` applies one config to every input, making mixed-config Solidity monorepos require separate invocations.

Closes #5686.

## Solution

Add `forge fmt --nearest` to resolve each file's nearest `foundry.toml`, including its formatter settings and ignore rules. Existing behavior is unchanged without the flag.

## Validation

- CLI coverage for per-project formatting, nested ignore ownership, and config conflicts.
- `sablier-labs/evm-monorepo`: 645 files across 6 projects matched independent `--root` runs byte-for-byte; its ignored deployed source remained untouched.
- `latticexyz/mud`: 495 files across 25 projects matched independent `--root` runs byte-for-byte, including 487 reformatted files.

### DIFF
--- a/crates/forge/src/cmd/fmt.rs
+++ b/crates/forge/src/cmd/fmt.rs
@@ -4,11 +4,15 @@ use eyre::Result;
 use foundry_cli::utils::{FoundryPathExt, LoadConfig};
 use foundry_common::{errors::convert_solar_errors, fs};
 use foundry_compilers::{compilers::solc::SolcLanguage, solc::SOLC_EXTENSIONS};
-use foundry_config::{filter::expand_globs, impl_figment_convert_basic};
+use foundry_config::{
+    Config, filter::expand_globs, find_project_root, fmt::FormatterConfig,
+    impl_figment_convert_basic,
+};
 use rayon::prelude::*;
 use similar::{ChangeTag, TextDiff};
 use solar::sema::Compiler;
 use std::{
+    collections::{HashMap, hash_map::Entry},
     fmt::{self, Write},
     io,
     io::Write as _,
@@ -31,6 +35,10 @@ pub struct FmtArgs {
     #[arg(long, value_hint = ValueHint::DirPath, value_name = "PATH")]
     root: Option<PathBuf>,
 
+    /// Use each input file's nearest `foundry.toml` for formatter settings.
+    #[arg(long, conflicts_with = "root")]
+    use_nearest_config: bool,
+
     /// Run in 'check' mode.
     ///
     /// Exits with 0 if input is formatted correctly.
@@ -50,6 +58,13 @@ impl_figment_convert_basic!(FmtArgs);
 
 impl FmtArgs {
     pub fn run(self) -> Result<()> {
+        if self.use_nearest_config {
+            for var in ["FOUNDRY_CONFIG", "FOUNDRY_ROOT", "DAPP_ROOT"] {
+                if std::env::var_os(var).is_some() {
+                    eyre::bail!("`--use-nearest-config` cannot be used when `{var}` is set");
+                }
+            }
+        }
         let config = self.load_config()?;
         let cwd = std::env::current_dir()?;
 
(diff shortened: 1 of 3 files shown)


crates/forge/src/cmd/fmt.rs
```diff
@@ -65,28 +80,29 @@ impl FmtArgs {
...
             [] => {
                 // Retrieve the project paths, and filter out the ignored ones and libs.
                 let project_paths: Vec<PathBuf> = config
                     .project_paths::<SolcLanguage>()
                     .input_files_iter()
                     .filter(|p| {
-                        !(ignored.contains(p)
-                            || ignored.contains(&cwd.join(p))
-                            || is_under_ignored_dir(p, true))
+                        !((!self.nearest
```
mablr, line 106: Could we skip expanding the invocation config's `fmt.ignore` under `--nearest`? These ignores are unused in this mode, but an invalid outer glob currently aborts a file owned by a valid nested config.

crates/forge/src/cmd/fmt.rs
```diff
@@ -125,6 +144,42 @@ impl FmtArgs {
...
+            let mut path_configs = HashMap::new();
+            let mut filtered_paths = Vec::with_capacity(paths.len());
+
+            for path in std::mem::take(paths) {
+                let path = fs::canonicalize_path(path)?;
+                let root = find_project_root(path.parent())?;
+                let (fmt_config, ignored) = match root_configs.entry(root) {
+                    Entry::Occupied(entry) => entry.into_mut(),
+                    Entry::Vacant(entry) => {
+                        let nearest_config = Config::load_with_root(entry.key())?.sanitized();
```
mablr, line 167: Could we preserve warnings when loading nested configs? This direct load drops deprecation/config warnings that `forge fmt --root <project>` would report; the root cache can keep them once per project.

crates/forge/tests/cli/fmt.rs
```diff
@@ -139,6 +139,57 @@ Diff in src/FmtTest.sol:
...
+    );
+    prj.create_file(
+        "second/foundry.toml",
+        r#"[fmt]
+multiline_func_header = "attributes_first"
+tab_width = 6
+"#,
+    );
+
+    cmd.args(["fmt", "--nearest", "."]).assert_success();
```
mablr: Could we add a no-`--nearest` compatibility control and a multi-root `--check` case? These would protect default config ownership plus per-root exit/output behavior.

mablr (dismissed):
LGTM, some non-blocking suggestions...

crates/forge/tests/cli/fmt.rs
figtracer: Added both: a no-`--nearest` ownership control and a multi-root `--check` case that verifies each project’s diff output and failing exit status.

crates/forge/src/cmd/fmt.rs
figtracer: Done — `--nearest` now skips the invocation config’s ignore expansion. The regression test includes an invalid outer glob while formatting valid nested projects.

crates/forge/src/cmd/fmt.rs
figtracer: Done — nested config warnings are emitted once per cached project root, without duplicating warnings from the invocation root.

mablr (approved):

stevencartavia (approved):

PR #16065, opened 2026-08-07
fix(fuzz): persist parallel invariant corpus

mablr:
## Motivation

Fixes OSS-574 by atomically persisting interesting invariant corpus entries as each parallel worker discovers them. This isolates the persistence fix previously included in #15816, whose performance impact could not be measured conclusively within the broader PR stack. Corpus setup failures now stop the campaign, while runtime persistence failures emit one warning per worker rather than silently losing discoveries.

Immediate worker-local persistence intentionally trades campaign-wide coverage deduplication for durability: workers may retain separate entries representing equivalent coverage, increasing corpus size and replay work. This prevents external termination from losing the entire campaign corpus; compaction can be considered separately if longer campaigns demonstrate unacceptable growth.

## Results
Quick scfuzzbench to control throughput, to be confirmed with long derek run.
| 8-worker Aave v4 campaign | `master` | This PR |
| --- | ---: | ---: |
| 180s throughput | 5,087 tx/s | 5,999 tx/s |
| Corpus after termination | Missing | 1,593 entries / 56 MiB |
| Corpus replay | Failed | Passed in 36s |

Assisted by AI.

### DIFF
--- a/crates/evm/evm/src/executors/invariant/mod.rs
+++ b/crates/evm/evm/src/executors/invariant/mod.rs
@@ -992,6 +992,9 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
                         actual_worker_count,
                         gas_report_samples,
                     );
+                    if output.is_err() {
+                        campaign_state.request_terminal_stop();
+                    }
                     debug!("finished in {:?}", timer.elapsed());
                     output
                 })
--- a/crates/evm/evm/src/executors/invariant/mod.rs
+++ b/crates/evm/evm/src/executors/invariant/mod.rs
@@ -3,6 +3,7 @@ use crate::{
         DURATION_BETWEEN_METRICS_REPORT, EarlyExit, EvmError, Executor, RawCallResult,
         corpus::{
             CorpusInsertionMode, DynamicTargetCtx, ReplayTarget, WorkerCorpus, WorkerCorpusSeed,
+            persist_campaign_optimization,
         },
     },
     inspectors::Fuzzer,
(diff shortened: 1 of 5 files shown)


mablr:
derek bench invariant timeout=600

mablr:
derek bench invariant timeout=600 trials=6 paired-seeds=true

mablr:
derek bench invariant

crates/evm/evm/src/executors/invariant/mod.rs
```diff
@@ -1007,7 +989,6 @@ impl<'a, FEN: FoundryEvmNetwork> InvariantExecutor<'a, FEN> {
                         &campaign_state,
                         worker_campaign_seed,
                         worker_corpus_seed,
-                        corpus_persistence,
                         actual_worker_count,
```
stevencartavia, line 992: This corpus setup error returns from `run_invariant_worker` without signaling the shared campaign stop. Since Rayon cannot interrupt an already-running worker, a sibling can continue through its assigned runs before the error is returned.

Reproduced by making only worker 1’s corpus directory inaccessible: worker 0 continued running and persisted corpus entries. Could the parallel closure call `campaign_state.request_terminal_stop()` when `run_invariant_worker` returns an error?

mattsse (dismissed):
lgtm

crates/evm/evm/src/executors/invariant/mod.rs
mablr: https://github.com/foundry-rs/foundry/pull/16065/commits/70f85573a143bdd68706c8129e6cb3706a229669

mablr:
derek bench invariant timeout=600 workers=8 trials=20 paired-seeds=true

mablr:
derek bench invariant timeout=600 workers=8 trials=20 paired-seeds=true

figtracer:
derek bench invariant compare-ref=4b6038a83bfcf64119b0737e6a8958bed3329fac timeout=600 workers=8 trials=20 paired-seeds=true

figtracer (approved):

0xKarl98 (approved):

PR #16066, opened 2026-08-07
refactor(verify): drop the unreachable unknown optimizer runs branch

mkzung:
Follow-up to #16055, which @mattsse closed because the premise was wrong: the CLI path
normalizes `optimizer_runs`, so nothing user-facing was broken. This is the other option
offered there: the branch removed, and called dead code rather than a bug.

`find_mismatch_in_settings` runs only from `print_result`, whose four callers all sit in
`bytecode.rs` and pass the `Config` that comes from `load_config`. That normalizes
`optimizer_runs`, so it is always set by then and the `None` arm cannot print.

No behaviour change, so no test. The unit tests in the closed PR failed only because they
built `Config` by hand and skipped config loading, which is what made the premise look true.

Claude Code was used for implementation assistance, covering the code change and this
description. All changes have been reviewed, understood and manually tested by me.


mablr (dismissed):
sgtm

stevencartavia (dismissed):

mkzung:
Rebased onto master and added a `.changelog` entry, so the two red checks are gone. The changelog gate landed after this PR opened, and `cargo deny` was flagging RUSTSEC-2026-0258 because the branch still pinned h2 0.4.15 while master is on 0.4.16.

Sorry about the dismissed approvals, the rebase took them with it. The diff is unchanged apart from the new fragment.

mablr (approved):

figtracer (approved):

PR #16067, opened 2026-08-07
fix(fuzz): preserve pending corpus entries

mablr:
This follows up on #16058, which tracks corpus entries pending worker synchronization by index. Corpus eviction could remove a pending entry and its tracked index before it was delivered, allowing final synchronization to report success while silently losing the entry.

This prevents entries pending synchronization from being evicted. The protection is enabled only for stateless fuzz workers, since invariant campaigns do not use worker synchronization and must retain normal eviction behavior. Pending indices remain sorted, allowing eviction checks to use binary search without adding a linear lookup to the hot path.

The regressions verify that pending entries survive until worker fan-out while entries in non-synchronizing corpora remain evictable.

> Amp was used to review the implementation and draft this PR description.

### DIFF
--- a/.changelog/fuzz-pending-corpus-eviction.md
+++ b/.changelog/fuzz-pending-corpus-eviction.md
@@ -1,0 +1,6 @@
+---
+forge: patch
+foundry-evm: patch
+---
+
+Fixed pending fuzz corpus entries being evicted before worker synchronization completed.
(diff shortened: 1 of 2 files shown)


stevencartavia (approved):

mattsse (approved):
lgtm

PR #16068, opened 2026-08-07
feat(fmt): make disable-line contextual

### LINKED ISSUE #3831: `forgefmt: disable-line` match prettier's `prettier-ignore` behavior
### Component

Forge

It would be nice if `forgefmt: disable-line` behaved similarly to prettier's `prettier-ignore`. This would make switching formatters easier.

```solidity
contract TestContract is Test {
    function test() public {
        assembly { sstore(0, 0) }// prettier-ignore

// prettier-ignore
        assembly { sstore(0, 0) }

        // prettier-ignore
        assembly { sstore(0, 0) 

        sstore(0, 0) 
        }

        this;      // prettier-ignore
    }
}
```

gets turned into

```solidity
contract TestContract is Test {
    function test() public {
        assembly { sstore(0, 0) } // prettier-ignore

        // prettier-ignore
        assembly { sstore(0, 0) }

        // prettier-ignore
        assembly { sstore(0, 0) 

        sstore(0, 0) 
        }

        this; // prettier-ignore
    }
}
```

while `forge fmt` transforms the same into
```solidity
contract TestContract is Test {
    function test() public {
        assembly { sstore(0, 0) }// forgefmt: disable-line

    // forgefmt: disable-line
        assembly {
            sstore(0, 0)
        }// forgefmt: disable-line
        assembly {
            sstore(0, 0)

            sstore(0, 0)
        }

        this;      // forgefmt: disable-line
    }
}
```
I'd first note that some of the whitespace formatting doesn't look good.

I know there is the `disable-next-line`, but I'd argue that I've never had a need to distinguish those two cases.
What's more is that `disable-next-line` doesn't work on scope, e.g.
```solidity
        // forgefmt: disable-next-line
        assembly { sstore(0, 0) 
        }
```
is transformed to
```solidity
        // forgefmt: disable-next-line
        assembly {
            sstore(0, 0)
        }
```

Also writing `// prettier-ignore` is more succinct, than `forgefmt: disable-line`. It's slightly confusing that there is no dash "-" between *forge* and *fmt* while there is one for *disable-line* and having both, a semicolon and a dash makes it harder to type (or remember).

I'm not sure what other kind of comments could be needed (I don't think prettier has many more), but perhaps it could be changed to `// forge-fmt-ignore` or `forgefmt-ignore`.

### Additional context

_No response_

figtracer:
## Motivation

A comment-only `forgefmt: disable-line` does not preserve the following Solidity item, unlike `prettier-ignore`.

Closes #3831.

## Solution

Treat isolated `disable-line` directives as `disable-next-item`. Trailing and mixed directives keep their existing current-line behavior, and `disable-next-line` remains supported.

### DIFF
--- a/crates/fmt/src/lib.rs
+++ b/crates/fmt/src/lib.rs
@@ -255,6 +255,9 @@ fn parse_inline_config<'ast>(
         }
         let item = item.trim_start().strip_prefix("forgefmt:")?.trim();
         match item.parse::<InlineConfigItem<()>>() {
+            Ok(InlineConfigItem::DisableLine(())) if cmnt.style.is_isolated() => {
+                Some((cmnt.span, InlineConfigItem::DisableNextItem(())))
+            }
             Ok(item) => Some((cmnt.span, item)),
             Err(e) => {
                 sess.dcx.warn(e.to_string()).span(cmnt.span).emit();
--- a/.changelog/forgefmt-contextual-disable-line.md
+++ b/.changelog/forgefmt-contextual-disable-line.md
@@ -1,0 +1,5 @@
+---
+forge-fmt: minor
+---
+
+Made isolated `forgefmt: disable-line` directives preserve the following Solidity item.
--- a/crates/fmt/tests/formatter.rs
+++ b/crates/fmt/tests/formatter.rs
@@ -61,6 +61,37 @@ fn chained_named_call_layout_ignores_source_spacing() {
     }
 }
 
+// <https://github.com/foundry-rs/foundry/issues/3831>
+#[test]
+fn disable_line_uses_comment_context() {
+    let source = r#"contract  C {
+    function f() public {
+        // forgefmt: disable-line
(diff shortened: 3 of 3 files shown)


mablr (approved):
Makes sense

crates/fmt/src/lib.rs
```diff
@@ -255,6 +255,9 @@ fn parse_inline_config<'ast>(
         }
         let item = item.trim_start().strip_prefix("forgefmt:")?.trim();
         match item.parse::<InlineConfigItem<()>>() {
+            Ok(InlineConfigItem::DisableLine(())) if cmnt.style.is_isolated() => {
```
stevencartavia, line 258: Non-blocking: with `wrap_comments = true`, an isolated `// forgefmt: disable-line` immediately after a trailing comment is grouped with it and retains the `Trailing` style, so the following item is reformatted. Fine to address separately if needed.

stevencartavia (approved):

PR #16069, opened 2026-08-07
perf(linking): index artifacts by library path

figtracer:
## Summary

- lazily index artifact IDs by project-relative source path and unversioned library name instead of scanning every artifact for each dependency lookup
- reuse a `Resolver` across Forge's per-test library-address lookups while preserving the existing `Linker` fields and public method signatures
- preserve artifact iteration order and the existing build/profile disambiguation within each indexed candidate bucket
- retain the full canonical-path fallback for symlinked library aliases

## Performance

Benchmarked release builds with cached `forge test --quiet` on a generated large project containing 2,000 libraries, 1,000 test contracts, and 40 external library references per test contract (3,000 artifacts total). Hyperfine ran five measured iterations after one warmup.

| | End-to-end mean | Peak RSS (3-run average) |
|---|---:|---:|
| `master` | 9.631 s ± 0.106 s | 347.5 MiB |
| this PR | 2.555 s ± 0.003 s | 349.6 MiB |

This is **3.77× faster**, saving **7.08 seconds**. Peak RSS increased by **2.1 MiB (0.60%)**.

The index is initialized only on the first dependency lookup. Forge reuses one resolver across test contracts, while linker operations without library references do not pay the construction or memory cost.

### DIFF
--- a/.changelog/anvil-trace-filter-pagination.md
+++ b/.changelog/anvil-trace-filter-pagination.md
@@ -1,5 +1,0 @@
----
-anvil: patch
----
-
-Avoid materializing unneeded traces for paginated local `trace_filter` requests.
--- a/.changelog/anvil-trace-filter-pagination.md
+++ b/.changelog/anvil-trace-filter-pagination.md
@@ -1,0 +1,5 @@
+---
+anvil: patch
+---
+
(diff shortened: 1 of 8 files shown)


mablr (approved):
noice

stevencartavia (approved):

PR #16071, opened 2026-08-07
fix(cast): reject unknown scope fields

stevencartavia:
Reject unknown fields in Cast call scope JSON instead of silently ignoring them. This prevents misspelled selector restrictions from becoming target-wide key authorizations.

### DIFF
--- a/crates/cast/src/cmd/keychain.rs
+++ b/crates/cast/src/cmd/keychain.rs
@@ -653,6 +653,7 @@ fn abi_scope_to_auth_scope(scope: CallScope) -> AuthCallScope {
 }
 /// Represents a single scope entry in JSON format for `--scopes`.
 #[derive(serde::Deserialize)]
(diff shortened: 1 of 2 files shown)


mablr (approved):

mattsse (approved):

crates/cast/src/cmd/keychain.rs
```diff
@@ -4256,6 +4257,13 @@ mod tests {
         assert_eq!(result[0].selectorRules[0].recipients.len(), 1);
     }
 
+    #[test]
+    fn test_parse_scopes_json_deny_unknown_scope_fields() {
+        let json =
+            r#"[{"target":"0x20c0000000000000000000000000000000000001","selector":["transfer"]}]"#;
```
wyattloveslife-create, line 4263: Why do this?

PR #16073, opened 2026-08-08
fix(doc): preserve docs for unnamed params

stevencartavia:
Preserves inherited NatSpec parameter descriptions when an overriding function leaves its parameters unnamed. Forge now uses the parameter position as a fallback while retaining existing behavior for named and locally documented parameters.


mattsse (approved):
lgtm

mablr (approved):

PR #16074, opened 2026-08-08
fix(doc): preserve unnamed return docs

stevencartavia:
Preserves complete NatSpec descriptions for unnamed function return values. Solar already separates optional return names from their descriptions, so Forge no longer removes the first word from unnamed return prose.


mattsse (approved):
lgtm

mablr (approved):

PR #16077, opened 2026-08-08
fix(forge): revalidate cached invariants

stevencartavia:
Persisted failures for secondary invariants previously caused those predicates to be omitted from later grouped campaigns. Because cache compatibility does not prove that a counterexample still fails against the current code, this could allow a deterministically failing invariant suite to exit successfully.

Keep all selected secondary invariants in the live campaign so current execution determines their result and rediscovered failures replace stale artifacts.


crates/forge/src/runner.rs
```diff
@@ -3651,58 +3651,10 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
         let replay_invariant_fns = live_invariants.clone();
-        let invariant_fns: Vec<(&Function, bool)> = live_invariants
-            .into_iter()
-            .filter(|(invariant_fn, _)| {
-                *invariant_fn == campaign_anchor
-                    || (!is_optimization
-                        && !persisted_invariants.contains(invariant_fn.name.as_str()))
-            })
-            .collect();
+        let invariant_fns = live_invariants;
```
mattsse, line 3666: Keeping the predicate live does not revalidate its cached sequence: ordinary runs still put only the anchor in `replay_candidates`. If the fresh budget does not rediscover a still-valid secondary—deterministically with `runs = 0`—the grouped run reports it as passing, while `forge fuzz replay --mt invariant_secondary` immediately proves the same cache still fails. The added tests always rediscover their `inc()` paths, so they do not cover this case. Please replay compatible secondary caches once and seed/report reproduced failures before continuing the campaign.

mattsse (dismissed):
One correctness issue remains.

crates/forge/src/runner.rs
stevencartavia: [e1afd9d](https://github.com/foundry-rs/foundry/pull/16077/commits/e1afd9d509df4697914ec8510913eeba3c35be30)

crates/forge/src/runner.rs
```diff
@@ -3878,6 +3841,38 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                            replay_contract.address,
+                            replay_invariant,
+                            txes,
+                            replay.reason,
+                            invariant_config,
+                            fail_on_revert,
+                            assertion_failure,
+                            is_revert,
+                        ),
+                    ));
```
mablr, line 3883: Persisted symbolic failures can require non-empty `storage`. This reconstruction drops it; the seeded error is later replayed with `&[]` and can overwrite a valid cache with a non-reproducing artifact. Please retain the replay metadata/storage unless the campaign discovers a fresh failure.

crates/forge/tests/cli/test_cmd/invariant/mod.rs
```diff
@@ -2452,6 +2452,213 @@ Tip: Run `forge test --rerun` to retry only the 1 failed test
...
+}
+   "#;
+    prj.add_test("PersistedSecondaryReplayTest.t.sol", test);
+
+    cmd.args(["test", "--mt", "invariant_"]).assert_failure();
+
+    prj.update_config(|config| {
+        config.invariant.runs = 0;
+    });
+    cmd.forge_fuse().args(["test", "--mt", "invariant_"]).assert_failure().stdout_eq(str![[r#"
```
mablr, line 2570: Please extend this case with non-empty symbolic `storage`, run the grouped `runs = 0` campaign twice, and assert the storage survives. This should catch cache corruption after a seeded replay.

crates/forge/src/runner.rs
```diff
@@ -3878,6 +3841,38 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
+                            if target == replay_contract.address
+                                && selector == replay_invariant.selector() =>
+                        {
+                            false
+                        }
+                        SymbolicInvariantFailureSite::SequenceCall { .. }
+                            if fail_on_revert && !replay.sequence_assertion_failure =>
+                        {
+                            true
+                        }
```
mablr, line 3866: Could we add focused coverage for cached secondary replay through this `fail_on_revert` branch? The current tests only exercise predicate failures.

crates/forge/src/runner.rs
stevencartavia: [e7ff08a](https://github.com/foundry-rs/foundry/pull/16077/commits/e7ff08a2001b25a6e881294974c660dac7e70d0f)

crates/forge/tests/cli/test_cmd/invariant/mod.rs
stevencartavia: [e7ff08a](https://github.com/foundry-rs/foundry/pull/16077/commits/e7ff08a2001b25a6e881294974c660dac7e70d0f)

crates/forge/src/runner.rs
stevencartavia: [e7ff08a](https://github.com/foundry-rs/foundry/pull/16077/commits/e7ff08a2001b25a6e881294974c660dac7e70d0f)

mablr (commented):
clippy 🙏 otherwise lgtm

crates/forge/src/runner.rs
```diff
@@ -3878,6 +3842,39 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
                 if failure_site.is_some_and(|expected| expected != confirmed_failure_site) {
                     continue;
                 }
+                if replay_invariant != invariant_contract.anchor()
+                    && !self.cr.mcr.tcfg.fuzz_failure_replay
+                {
+                    let is_revert = match confirmed_failure_site {
+                        SymbolicInvariantFailureSite::Invariant { target, selector, .. }
+                            if target == replay_contract.address
```
mablr: `Invariant.target` can be the nested callee recorded by `call_result.reverter`, not the test contract. This rejects a still-valid cached secondary and lets `runs = 0` pass. Please classify this terminal failure by the replayed invariant selector (the exact stored-site check above still applies) and cover a predicate that bubbles a nested revert.

crates/forge/src/runner.rs
```diff
@@ -4610,18 +4611,22 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
                             &current_settings,
                             &invariant.signature(),
                             &invariant.signature(),
-                            &[],
+                            replayed_storage.map(Vec::as_slice).unwrap_or_default(),
```
mablr: This re-shrinks an already site-validated cached failure with the generic shrinker, which accepts any failing outcome. A candidate can switch to a handler assertion, plain revert, or `afterInvariant` failure and still be reported as this predicate. Please either skip shrinking seeded failures or require shrink candidates to preserve the confirmed site/classification, with a nonzero-shrink regression.

mablr (commented):
Clanker found 2 more ^^

crates/forge/src/runner.rs
stevencartavia: [2919527](https://github.com/foundry-rs/foundry/pull/16077/commits/2919527fe15263872db29c67edd583e557783c62)

crates/forge/src/runner.rs
stevencartavia: [2919527](https://github.com/foundry-rs/foundry/pull/16077/commits/2919527fe15263872db29c67edd583e557783c62)

figtracer (dismissed):

crates/forge/src/runner.rs
```diff
@@ -4610,18 +4618,22 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
...
```
mablr: This replay writes the same deterministic symbolic artifact path as the original failure, but `replay_invariant_error_sequence` supplies no `invariant_failure` metadata. With symbolic mode enabled, a grouped rerun can therefore overwrite a valid secondary artifact and drop its validated failure site, causing later `--replay-symbolic-artifact` to reject it. Please preserve the confirmed site in this path (or avoid rewriting the artifact) and extend the storage regression to assert the artifact metadata survives a grouped rerun.

stevencartavia: [185662d](https://github.com/foundry-rs/foundry/pull/16077/commits/185662d0620d5bd16f50484551bf855262b938c8)

mablr (dismissed):
SGTM
Does it have an impact on scfuzzbench ?

stevencartavia:
> SGTM Does it have an impact on scfuzzbench ?

Cache-clean scfuzzbench runs are unaffected, but reused target workdirs may change because persisted secondary failures are now revalidated.

crates/forge/src/runner.rs
```diff
@@ -4409,6 +4407,13 @@ impl<'a, FEN: FoundryEvmNetwork> FunctionRunner<'a, FEN> {
                 return self.result;
             }
         };
+        let mut replayed_secondary_metadata = BTreeMap::new();
+        for (name, failure, storage, failure_site) in replayed_secondary_failures {
```
figtracer, line 4466: `replayed_secondary_failures` is only merged here, after the symbolic anchor path can return early. i reproduced a grouped `--symbolic --fuzz-runs 0` run reporting only the anchor and dropping a still-valid cached secondary. can we merge the replayed failures before returning?

stevencartavia: [3761c3a](https://github.com/foundry-rs/foundry/pull/16077/commits/3761c3a9681dcf5449dc2e34de670008e65fcb89)

mablr (dismissed):

mablr (approved):

figtracer (approved):

PR #16082, opened 2026-08-08
perf(bind): generate bindings in parallel

figtracer:
## Summary

- generate and format independent Solidity bindings in parallel
- retain finalized Rust source instead of non-`Send` token streams, enabling Rayon while reducing retained memory
- preserve contract ordering, generated output, all bind modes, and `--threads 1` behavior

## Performance

Benchmarked optimized builds with cached `forge bind --skip-build --select-all` on a generated project containing 2,000 libraries and 1,000 test contracts (3,000 artifacts total). Hyperfine ran five measured iterations after one warmup.

| | End-to-end mean | Peak RSS |
|---|---:|---:|
| `master` | 3.774 s ± 0.013 s | 603.4 MiB |
| this PR | 438.5 ms ± 14.3 ms | 189.2 MiB |

This is **8.61× faster**, saving **3.34 seconds**, while peak RSS decreases by **414.2 MiB (68.6%)**. With `--threads 1`, runtime remains unchanged (3.743 s vs. 3.658 s).

Generated output was byte-identical for all 3,000 contracts and across crate, module, single-file, module single-file, and no-extra-derives modes.

## Validation

- `cargo test -p forge --test cli bind_` (12 passed)
- `cargo clippy -p forge-sol-macro-gen --all-targets --all-features -- -D warnings`

### DIFF
--- a/.changelog/cast-wallet-mnemonic-derivation.md
+++ b/.changelog/cast-wallet-mnemonic-derivation.md
@@ -1,0 +1,5 @@
+---
+cast: patch
+---
+
+Speed up deriving multiple accounts from a mnemonic.
--- a/.changelog/cast-wallet-mnemonic-derivation.md
+++ b/.changelog/cast-wallet-mnemonic-derivation.md
@@ -1,5 +1,0 @@
----
-cast: patch
----
-
-Speed up deriving multiple accounts from a mnemonic.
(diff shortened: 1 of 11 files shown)


stevencartavia (approved):

mattsse (approved):
lgtm

PR #16083, opened 2026-08-08
fix(verify): report null-contract errors

figtracer:
The Tempo nightly failure in #16075 exposed a valid Sourcify-compatible error response where a completed job carries `contract: null` alongside an `error` object. Foundry required `contract` to be an object, so it retried a permanent deserialization failure and hid the verifier's actual RPC rate-limit error.

This models the contract result as optional, allowing the existing terminal-error path to report the verifier code and message immediately. Regression coverage exercises the full status-polling path with a null-contract error response.

This PR was prepared with AI assistance.

### DIFF
--- a/.changelog/sourcify-null-contract.md
+++ b/.changelog/sourcify-null-contract.md
@@ -1,0 +1,5 @@
+---
+forge-verify: patch
+---
+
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm

stevencartavia (approved):

PR #16087, opened 2026-08-08
feat(symbolic): prove bounded arithmetic identities

figtracer:
## Motivation

`forge test --symbolic` reports some valid EVM arithmetic properties as incomplete when compiler-generated nonlinear expressions reach the hard-arithmetic fallback. A pinned Solady property provides a real example: `check_CarryBoundsTrick(uint248,uint8)` proves that a base-58 carry remains bounded after accumulating a masked 248-bit limb, but Foundry previously stopped with heuristic witnesses instead of proving it.

## Solution

This PR adds two deliberately bounded proof rules. First, it canonicalizes small polynomial identities over the EVM word ring. Second, it proves the compiler-generated `(x * d) / d == x` obligation and propagates unsigned bounds through a concrete logical right shift, but only when the path proves `d != 0` and `x * d` cannot wrap.

The optimization is transactional. Unsupported expressions keep their original solver shape, and wrapping multiplication or a zero divisor is never simplified. Unit tests cover both unsafe cases and a shifted product whose multiplication can wrap.

## Results

The comparison uses the same optimized Solidity source and command on an exact `master` binary and this branch, with Solady pinned to v0.1.26.

| pinned target | master | this branch |
| --- | --- | --- |
| Solady `check_CarryBoundsTrick(uint248,uint8)` | incomplete; 8 solver queries; 2 heuristic witnesses | pass; 4 solver queries; 0 heuristic witnesses |
| Other 12 Solady `check_` properties | unchanged statuses and metrics | unchanged statuses and metrics |
| Wall time, 30 runs | 1.028 ± 0.018 s | 1.007 ± 0.015 s |

The official `derek bench symex timeout=1800` run independently reproduced the completeness change on the proof change:

| suite | master | this branch |
| --- | --- | --- |
| Solady | 9 pass, 4 incomplete, 172 solver queries, 30 heuristic witnesses | 10 pass, 3 incomplete, 168 solver queries, 28 heuristic witnesses |
| Angstrom | 1 incomplete, 12 solver queries | unchanged |
| Farcaster | 2 pass, 14 solver queries | unchanged |
| symbolic bug suite | 22 expected failures, 273 solver queries | unchanged |

Solady wall time remained 695 ms, while the other suite means ranged from neutral to 3.9% faster. Derek labels the Solady row a regression because its generic verdict treats every symbolic correctness-status change as a regression; the underlying change is the intended improvement from incomplete to proven. The claim here is completeness for a sound, concrete class of bounded arithmetic obligations, not a broad symbolic-performance improvement.

Validation includes 255 symbolic engine unit tests, a Solidity CLI regression compiling with the optimizer enabled, exhaustive wrapping-model checks for the polynomial normalizer, expansion-budget tests, `cargo check`, rustfmt, and nightly clippy. A broad symbolic CLI sweep completed 289 tests successfully; its 18 existing snapshot, minimization, modeling-expectation, and timeout failures are outside the changed solver rules.

AI assistance disclosure: Codex was used for implementation, tests, benchmark design and execution, and drafting this description. I reviewed the resulting code and evidence.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


figtracer:
derek bench symex timeout=1800

figtracer:
derek bench symex timeout=1800

mattsse (dismissed):
lgtm

crates/evm/symbolic/src/runtime/solver/opt.rs
```diff
@@ -977,10 +1021,223 @@ fn nonzero_bound<'a>(expr: &'a SymExpr, value: &'a SymExpr) -> Option<(&'a SymEx
...
+        SymExprKind::BinOp(SymBinOp::Add | SymBinOp::Sub, left, right) => {
+            is_product(left) || is_product(right)
+        }
+        _ => false,
+    };
+    if !crosses_sum_product_boundary {
+        return false;
+    }
+
+    fn ring_shape(expr: &SymExpr) -> Option<(usize, usize)> {
```
mablr: **Footgun:** This recursively revisits both children without memoizing hash-consed nodes. A compact shared DAG such as `p(n+1) = p(n) + p(n)` can take exponential work while producing only a few polynomial terms, so the output limits never fire. Please memoize by `SymExpr` or add an input-work budget.

crates/evm/symbolic/src/tests.rs
```diff
@@ -3143,6 +3192,150 @@ fn solver_does_not_invert_guarded_zero_self_division() {
...
+            }
+        }
+    }
+}
+
+#[test]
+fn solver_polynomial_normalization_stops_at_factor_limit() {
+    let mut cx = SymCx::new();
+    let mut expression = SymExpr::one(&mut cx);
+    for index in 0..9 {
```
mablr: This test never reaches `MAX_MONOMIAL_FACTORS`: a nested product does not cross the sum/product boundary, so normalization exits before parsing it. Wrap the product in a sum with another product so deleting the factor cap would make this test fail.

crates/evm/symbolic/src/tests.rs
```diff
@@ -3059,6 +3059,55 @@ fn solver_normalizes_checked_mul_guard_for_bounded_operands() {
...
+        normalize_constraints_for_solver(&mut cx, std::slice::from_ref(&identity)),
+        vec![identity]
+    );
+}
+
+#[test]
+fn solver_does_not_normalize_zero_divisor_mul_div_identity() {
+    let mut cx = SymCx::new();
+    let value = SymExpr::var(&mut cx, "value");
+    let zero = SymExpr::zero(&mut cx);
```
mablr: This does not exercise the divisor-nonzero guard because `value * 0` and then `/ 0` fold to zero during expression construction. Use a symbolic divisor without a `d != 0` constraint; `d = 0, value = 1` is the concrete counterexample.

crates/forge/tests/cli/test_cmd/symbolic.rs
```diff
@@ -540,6 +540,51 @@ contract SymbolicPass {
...
+        );
+        return;
+    }
+
+    prj.add_test(
+        "SymbolicBoundedCarry.t.sol",
+        r#"
+contract SymbolicBoundedCarry {
+    function checkCarryBounds(uint248 limb, uint8 carry) public pure {
+        if (carry < 58) {
```
mablr, line 556: Please add the immediate failure boundary (`carry <= 58`; `limb = 2^248 - 1`, `carry = 58`) and the exact `MAX / factor` versus `+1` multiplication boundary. These protect the no-wrap condition rather than only its happy path.

crates/evm/symbolic/src/tests.rs
```diff
@@ -3143,6 +3230,173 @@ fn solver_does_not_invert_guarded_zero_self_division() {
...
+    for index in 0..9 {
+        let factor = SymExpr::var(&mut cx, &format!("x_{index}"));
+        oversized_product = SymExpr::binop(&mut cx, SymBinOp::Mul, oversized_product, factor);
+    }
+    let left = SymExpr::var(&mut cx, "left");
+    let right = SymExpr::var(&mut cx, "right");
+    let second_product = SymExpr::binop(&mut cx, SymBinOp::Mul, left, right);
+    let expression = SymExpr::binop(&mut cx, SymBinOp::Add, oversized_product, second_product);
+
+    assert_eq!(normalize_polynomial_for_solver(&mut cx, expression.clone()), expression);
```
mablr: This now reaches the factor cap, but the assertion remains mutation-insensitive: allowing nine factors can rebuild the same canonical product and still pass. Use an eight-factor product multiplied by `(a + b)` so weakening the cap visibly distributes it into two monomials.

crates/forge/tests/cli/test_cmd/symbolic.rs
```diff
@@ -540,6 +540,115 @@ contract SymbolicPass {
...
+    }
+
+    function checkCarryBoundary(uint248 limb, uint8 carry) public pure {
+        if (limb == type(uint248).max && carry == 58) {
+            uint256 accumulator = uint256(limb) * 58 + uint256(carry);
+            assert((accumulator >> 248) < 58);
+        }
+    }
+
+    function checkMulDivBoundary() public pure {
```
mablr: These boundary functions are input-free and compiled with `--optimize`, so Solc may fold them before symbolic execution. Consider accepting a symbolic value and constraining it to `boundary` / `boundary + 1` so this integration test exercises the mul/div rewrite.

crates/evm/symbolic/src/runtime/solver/opt.rs
```diff
@@ -1355,7 +1637,10 @@ impl ConstraintContext {
     }
 
     pub(super) fn mul_cannot_overflow_256(&self, left: &SymExpr, right: &SymExpr) -> bool {
-        self.unsigned_bits(left).saturating_add(self.unsigned_bits(right)) <= 256
+        self.interval(left)
+            .zip(self.interval(right))
+            .is_some_and(|(left, right)| left.max.checked_mul(right.max).is_some())
+            || self.unsigned_bits(left).saturating_add(self.unsigned_bits(right)) <= 256
```
mablr: **Important:** This new interval path exposes uncached recursive traversal of shared arithmetic DAGs. For `p0 = x & 1; p(n+1) = p(n) + p(n)`, proving a no-wrap identity over `p64` causes roughly `2^64` visits despite only 65 distinct nodes. Please memoize `Option<WordInterval>` by `SymExpr` or impose a work budget, with a construction-level regression.

crates/evm/symbolic/src/runtime/solver/opt.rs
```diff
@@ -593,6 +593,13 @@ fn normalize_bool_node_for_solver(cx: &mut SymCx, expr: SymBoolExpr) -> SymBoolE
         SymBoolExprKind::Cmp(op, left, right) => {
             let left = normalize_expr_for_solver(cx, left.clone());
             let right = normalize_expr_for_solver(cx, right.clone());
+            if *op == SymCmpOp::Eq {
+                let polynomial_left = normalize_polynomial_for_solver(cx, left.clone());
```
mablr: **Important:** For non-identity equalities, this rebuilds canonical polynomial `SymExpr` DAGs and immediately drops them. `HashCons` retains dead weak entries, so repeated normalization before SAT-cache lookup can accumulate O(Q) dead entries per shape and O(Q^2) bucket scans. Compare the bounded `Polynomial` term maps directly here instead of materializing temporary expressions.

crates/evm/symbolic/src/runtime/solver/opt.rs
```diff
@@ -1355,37 +1675,236 @@ impl ConstraintContext {
...
+        for index in 0..8 {
+            let factor = SymExpr::var(&mut cx, &format!("x_{index}"));
+            product_at_limit = SymExpr::binop(&mut cx, SymBinOp::Mul, product_at_limit, factor);
+        }
+        let left = SymExpr::var(&mut cx, "left");
+        let right = SymExpr::var(&mut cx, "right");
+        let sum = SymExpr::binop(&mut cx, SymBinOp::Add, left, right);
+        let expression = SymExpr::binop(&mut cx, SymBinOp::Mul, product_at_limit, sum);
+
+        assert!(!polynomial_identity(&expression, &expression));
```
mablr: This exercises nine-factor monomials, so it only covers rejection above the limit. Please also assert a factored/expanded identity with exactly `MAX_MONOMIAL_FACTORS` factors; otherwise changing `>` to `>=` remains undetected.

crates/evm/symbolic/src/tests.rs
mablr: Is this fully addressed now ? @figtracer

mablr (dismissed):

crates/evm/symbolic/src/runtime/solver/opt.rs
```diff
@@ -684,6 +707,21 @@ impl ConstraintContext {
             SymBoolExprKind::Not(value) if self.unsigned_bool_always_true(value) => {
                 SymBoolExpr::constant(cx, false)
             }
+            SymBoolExprKind::Cmp(SymCmpOp::Eq, left, right)
```
stevencartavia, line 747: **Blocking:** `mul_div_identity` can use bounds derived from the equality it then removes.

For `d = 2^254 + 1`, `q = (x * d) / d`, and:

```text
q == x
q == y
y == 0
x == 4
```

At `x = 4`, EVM multiplication wraps: `4 * d mod 2^256 = 4`, so `q = 4 / d = 0`. Therefore the original conjunction is unsatisfiable.

However, `ConstraintContext` derives `q <= 0` from `q == y` and `y == 0`, then propagates it through the candidate `q == x` to derive `x <= 0`. That candidate-derived bound causes `mul_cannot_overflow_256` to classify `x * d` as non-wrapping, so `mul_div_identity` removes `q == x`. The returned constraints are then satisfiable with `x = 4, y = 0`.

Please prevent a conjunct from contributing to the proof used to eliminate itself, or disable this contextual rewrite. A regression test should verify that both the original and normalized conjunction remain unsatisfiable for this case.

crates/evm/symbolic/src/runtime/solver/opt.rs
```diff
@@ -16,7 +16,17 @@ pub(crate) fn normalize_constraints_for_solver(
...
+        normalized.iter().cloned().enumerate().map(|(index, constraint)| {
+            let rewritten = context.normalize_bool(cx, constraint.clone());
+            if rewritten == constraint {
+                return constraint;
+            }
+
+            // Contextual rewrites may remove this conjunct, so their proof cannot depend on
+            // bounds learned from the conjunct itself. Rebuild only after a rewrite succeeds;
+            // ordinary constraints retain the single shared context above.
+            ConstraintContext::excluding(&normalized, index).normalize_bool(cx, constraint)
```
mablr: **Blocking:** Per-conjunct exclusion still permits mutual support between distinct contextual candidates. Two mul/div identities connected by `a == b`, with each quotient independently bounded to zero, can each prove the other operand bounded; this parallel map then removes both even though the remainder admits the wrapping `a = b = 4` model. Please prove candidates sequentially from the retained set, or build one context excluding all rewrite candidates. The latter also avoids rebuilding an O(n^2) propagation context for each candidate (O(n^3) total).

crates/evm/symbolic/src/runtime/solver/opt.rs
```diff
@@ -14,6 +14,19 @@ pub(crate) fn normalize_constraints_for_solver(
     }
 
     let context = ConstraintContext::new(&normalized);
+    let mut contextual_candidates = Vec::with_capacity(normalized.len());
+    for constraint in &normalized {
+        contextual_candidates.push(context.normalize_bool(cx, constraint.clone()) != *constraint);
+    }
+    // A contextual rewrite must not use itself or another removable conjunct as proof. Removing
+    // facts cannot enable a new rewrite, so the full context identifies every candidate and one
```
mablr: **Blocking:** “Removing facts cannot enable a new rewrite” is false because inconsistent lower/upper bounds make `interval()` return `None`. Let `B = MAX / 58`, `K = B + 1`, `q58 = (x * 58) / 58`, and `q3 = (x * 3) / 3`. With `q58 == x`, `q58 <= B`, `q3 == x`, and `q3 == K`, the full context retains the first identity but marks the second as a candidate. Excluding the second removes the conflicting lower bound, allowing the first to prove itself via `x <= B`; both are then deleted, although `x = K` satisfies the remainder and falsifies `q58 == x` due to wrapping. Please exclude a syntactic superset of all context-rewriteable conjuncts when building the proof context rather than discovering candidates only through successful full-context rewrites.

mablr (approved):

stevencartavia (approved):

PR #16090, opened 2026-08-08
fix(forge): hide auth from create help

figtracer:
`forge create` inherits `--auth` from the shared transaction options even though EIP-7702 transactions require a destination and cannot deploy contracts. This hides the unsupported option from this command's help without changing its availability for Cast, and adds a regression test for the generated help output.

This pull request was prepared with AI assistance.


stevencartavia (approved):

mattsse (approved):
lgtm

PR #16091, opened 2026-08-08
perf(cast): skip prefix transaction traces

mattsse:
`cast run` now replays transactions before the selected transaction with tracing disabled, then installs the requested trace requirements immediately before executing the target. Prefix transactions still commit the same state transitions and revert behavior, while the target retains call traces, storage changes, internal decoding, and debugger data. This PR was prepared with Codex assistance for implementation, tests, and benchmarks.

### Results

Profiling builds at `origin/master` (`7c271517a3`) and this branch replayed targets at indices 0, 99, and 199 in the same 200-transaction local Anvil block running `work(200)`. Values are mean ± standard deviation over 10 runs for `--decode-internal` and 5 runs for `--debug`; debugger wall time includes fixed pseudo-terminal startup and exit overhead.

| Mode | Target | master | This PR | Speedup |
| `--decode-internal` | 0 | 0.301 ± 0.073 s | 0.339 ± 0.078 s | 0.89× |
| `--decode-internal` | 99 | 1.383 ± 0.183 s | 0.299 ± 0.050 s | 4.63× |
| `--decode-internal` | 199 | 1.775 ± 0.313 s | 0.246 ± 0.015 s | 7.23× |
| `--debug` | 0 | 2.272 ± 0.041 s | 2.265 ± 0.053 s | 1.00× |
| `--debug` | 99 | 2.593 ± 0.065 s | 2.382 ± 0.043 s | 1.09× |
| `--debug` | 199 | 3.536 ± 0.090 s | 2.344 ± 0.059 s | 1.51× |

For the late target, retired instructions fell from 5.592B to 0.765B in `--decode-internal` and from 6.003B to 0.761B in `--debug` (86.3% and 87.3% reductions). Peak RSS was effectively flat at 85.8 MB to 84.8 MB and 86.0 MB to 85.4 MB, respectively.

### DIFF
--- a/crates/cast/tests/cli/main.rs
+++ b/crates/cast/tests/cli/main.rs
@@ -4206,6 +4206,146 @@ async fn deploy_counter_and_set_number(
         .tx_hash()
 }
 
+// `cast run` must replay a transaction's block prefix without changing the trace requested for the
+// selected transaction. A single block covers a deployment in the first position, a revert in the
+// middle, and an internally traced state change in the last position.
+forgetest_async!(cast_run_fork_traces_only_target_transaction, |prj, cmd| {
+    let (api, handle) = anvil::spawn(NodeConfig::test()).await;
+    let endpoint = handle.http_endpoint();
+    let provider = handle.http_provider();
+    let sender = handle.dev_wallets().next().unwrap().address();
+
+    foundry_test_utils::util::initialize(prj.root());
+    prj.add_source(
+        "ReplayTarget",
+        r#"
+contract ReplayTarget {
+    uint256 public number;
+
+    constructor() {
+        number = 1;
+    }
+
+    function revertingIncrement() external {
+        number++;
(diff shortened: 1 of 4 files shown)


DaniPopes (approved):

stevencartavia (approved):

crates/cast/tests/cli/main.rs
```diff
@@ -4291,6 +4291,146 @@ async fn deploy_counter_and_set_number(
...
+    }
+
+    function _increment() internal {
+        number++;
+    }
+}
+"#,
+    );
+    let bytecode = cmd
+        .forge_fuse()
```
stevencartavia, line 4330: Suggestion: could this use `prj.ensure_foundry_bin("forge")` instead of `forge_fuse()` so the test can run independently from a clean, focused Cast test build?

mablr (approved):

PR #16093, opened 2026-08-08
feat(symbolic): prove saturating multiplication

figtracer:
Stack: #16087 → #16093 → #16094 → #16095.

This is the first follow-up to #16087. It teaches the symbolic engine the exact expression shapes used by Solidity's checked-multiply guard and branchless saturating multiplication, while keeping unrecognized nonlinear arithmetic fail-closed. Concrete witnesses from the recognized guard are still replayed against the original path constraints before they are accepted. Constructive support propagation now shares a strict 256-visit budget across every guard and seed order; exhaustion returns to the existing solver path instead of allowing repeated quadratic scans.

On the pinned Solady v0.1.26 benchmark, `check_SaturatingMulEquivalence` changes from incomplete to proven: paths remain 4, solver queries remain 7, SMT queries remain 0, and heuristic witnesses fall from 4 to 0. Across the full 13-property Solady slice, the result changes from 10 pass / 3 incomplete / 28 heuristic witnesses to 11 pass / 2 incomplete / 24 heuristic witnesses with the same 168 solver queries. A 20-run focused comparison was neutral at 1.214s baseline versus 1.210s here.

Validated with the complete 266-test `foundry-evm-symbolic` suite, a minimized Forge CLI regression, the full pinned Solady slice, `cargo check -p foundry-evm-symbolic -p forge`, formatting, and `git diff --check`. Nightly clippy for `foundry-evm-symbolic` passes with all targets and warnings denied.

This PR was developed with AI assistance. The implementation and benchmark results were manually reviewed and validated.


mattsse (dismissed):
lgtm

crates/evm/symbolic/src/runtime/solver/hard_arith_fallback.rs
```diff
@@ -110,6 +110,77 @@ pub(crate) fn hard_arith_fallback_model(
     search.model(0, &mut model, &mut assignments)
 }
 
+/// Returns a concrete SAT witness for constraints containing Solidity's checked-multiply guard.
+///
+/// Unlike the generic hard-arithmetic fallback, this is limited to an exact expression shape.
+/// A returned assignment is still validated by the caller against the original constraints.
+pub(crate) fn checked_mul_guard_model(
```
mablr: **Important:** This is not an exact witness path; it delegates to the same bounded candidate search as the generic fallback. Omitting the `heuristic_witnesses` increment therefore misreports heuristic search as exact. Remove this special path or classify its witness source honestly.

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -398,6 +398,14 @@ impl SymbolicSolver for SmtLibSubprocessSolver {
             self.cache_model_result(cache_key, model.clone());
             return Ok(model);
         }
+        if let Some(model) = checked_mul_guard_model(cx, &smt_constraints)
```
mablr: **Footgun:** When this search finds no witness, control can immediately invoke `validated_hard_arith_fallback_model`, repeating essentially the same search. With the 50k assignment cap this can double the work. Invoke the fallback once and retain its source/category if diagnostics need it.

crates/forge/tests/cli/test_cmd/symbolic.rs
```diff
@@ -585,6 +585,52 @@ contract SymbolicBoundedCarry {
...
+
+contract SymbolicSaturatingMul {
+    function mul(uint256 x, uint256 y) external pure returns (uint256) {
+        return x * y;
+    }
+
+    function checkSaturatingMul(uint256 x, uint256 y) public view {
+        (bool success,) = address(this).staticcall(abi.encodeCall(this.mul, (x, y)));
+        uint256 expected = success ? x * y : type(uint256).max;
+        assert(SaturatingMath.saturatingMul(x, y) == expected);
```
mablr, line 682: Please add boundary models for `(0, MAX)`, `(MAX, 0)`, `(MAX, 1)`, `(MAX, 2)`, `(2^255, 2)`, and swapped operands, plus a deliberately wrong overflow result that must yield a replay-confirmed counterexample. The current true-property test cannot detect every over-aggressive local contradiction rewrite.

crates/evm/symbolic/src/runtime/solver/hard_arith_fallback.rs
```diff
@@ -110,6 +110,110 @@ pub(crate) fn hard_arith_fallback_model(
...
+/// represent its semantic cases directly: the zero disjunct, a nonzero exact product, and a
+/// wrapping product. This does not perform the generic bounded candidate search, and a model is
+/// returned only when it satisfies every original constraint.
+pub(crate) fn checked_mul_guard_branch_model(constraints: &[SymBoolExpr]) -> Option<SymbolicModel> {
+    for constraint in constraints {
+        let Some((zero_operand, expected, guard_is_true)) = checked_mul_guard_branch(constraint)
+        else {
+            continue;
+        };
+        let assignments = [
```
mablr: **Footgun:** These fixed pairs miss simple constrained witnesses, e.g. `guard_true && y == 7` has `x = 0, y = 7`, and `guard_false && y == 3` has `x = MAX, y = 3`; both fall back to the bounded heuristic search. Consider completing simple operand equalities before validation, or add tests documenting this performance boundary.

crates/forge/tests/cli/test_cmd/symbolic.rs
```diff
@@ -649,6 +649,105 @@ contract SymbolicBoundedCarry {
...
+    }
+
+    function checkSaturatingMulBoundaries() public pure {
```
mablr: This boundary test is input-free and compiled with `--optimize`, so Solc may fold these assertions before symbolic execution. Parameterizing `x`/`y` and guarding them to the listed pairs would ensure the checked-multiply construction is exercised.

crates/evm/symbolic/src/runtime/solver/hard_arith_fallback.rs
```diff
@@ -110,6 +110,126 @@ pub(crate) fn hard_arith_fallback_model(
...
+            continue;
+        };
+        let assignments = if guard_is_true {
+            [(U256::ZERO, U256::ZERO), (U256::ONE, U256::ONE)]
+        } else {
+            [(U256::MAX, U256::from(2)), (U256::from(2), U256::MAX)]
+        };
+        for (zero_default, expected_default) in assignments {
+            let mut model = SymbolicModel::default();
+            for support in constraints {
```
mablr: **Footgun:** Support propagation runs before semantic seeds, so simple relations with neither side initially known still miss constructive models: `guard_false && x == y` can use `x=y=MAX`, and `guard_true && y == x+7` can use `x=0,y=7`, but both fall back to generic search. Consider rerunning support propagation after assigning one guard operand.

crates/evm/symbolic/src/tests.rs
```diff
@@ -2646,6 +2646,58 @@ fn bool_word_equality_simplifies_to_condition() {
     assert_eq!(actual, expected);
 }
 
+#[test]
+fn expression_simplifies_saturating_select_idiom() {
```
mablr, line 2656: The constant-only boundary table was removed, avoiding optimizer false confidence, but explicit `(0,MAX)`, `(MAX,1)`, `(MAX,2)`, and `(2^255,2)` coverage is now absent. Consider evaluating this original/simplified expression under those concrete models, or add symbolically constrained Solidity inputs.

crates/evm/symbolic/src/runtime/solver/hard_arith_fallback.rs
```diff
@@ -110,6 +110,132 @@ pub(crate) fn hard_arith_fallback_model(
...
+            [(U256::MAX, U256::from(2)), (U256::from(2), U256::MAX)]
+        };
+        for (zero_default, expected_default) in assignments {
+            let seed_orders = [
+                [(&zero_operand, zero_default), (&expected, expected_default)],
+                [(&expected, expected_default), (&zero_operand, zero_default)],
+            ];
+            for seeds in seed_orders {
+                let mut model = SymbolicModel::default();
+                if !propagate_fallback_support_constraints(constraints, &mut model) {
```
mablr: **Footgun:** Please share a total support-visit budget across guards and seed orders. A miss can currently perform O(g·n²) work before the bounded fallback; on exhaustion, returning `None` preserves the existing heuristic/unknown diagnostics.

mablr (commented):
No correctness bugs, maybe a complexity footgun to eval.

crates/evm/symbolic/src/runtime/solver/hard_arith_fallback.rs
```diff
@@ -756,6 +956,105 @@ fn zero_mask_equality(var: &Symbol, masked: &SymExpr, zero: &SymExpr) -> Option<
...
+    #[test]
+    fn checked_mul_guard_branch_model_stops_at_shared_support_budget() {
+        let mut cx = SymCx::new();
+        let x = SymExpr::var(&mut cx, "x");
+        let y = SymExpr::var(&mut cx, "y");
+        let guard = checked_mul_guard_word(&mut cx, &x, &y);
+        let zero = SymExpr::zero(&mut cx);
+        let mut constraints = vec![SymBoolExpr::eq(&mut cx, guard, zero).not(&mut cx)];
+
+        let support = (0..MAX_CHECKED_MUL_SUPPORT_VISITS)
```
mablr: The regression exhausts all 256 visits in its first scan, so it still passes if the budget is reset for every guard or seed attempt. Consider using smaller failed attempts whose individual work stays below the cap but whose cumulative work exceeds it, so the test specifically pins budget sharing.

mablr (dismissed):
LGTM, just a smol coverage suggestion.

crates/evm/symbolic/src/runtime/solver/opt.rs
```diff
@@ -168,16 +168,46 @@ const fn expr_ternop_key(op: SymTernOp) -> u8 {
 
 /// Returns whether normalized conjunctive constraints contain a direct contradiction.
 pub(super) fn constraints_are_directly_unsat(cx: &mut SymCx, constraints: &[SymBoolExpr]) -> bool {
-    constraints.iter().any(|constraint| match constraint.kind() {
+    let derived = constraints
+        .iter()
+        .filter_map(|constraint| bitwise_bool_word_fact(cx, constraint))
+        .collect::<Vec<_>>();
```
stevencartavia: Non-blocking: For `(bool_word(A) | bool_word(B)) == 0`, this derives `And(!A, !B)`, but `contains` only checks exact top-level membership. A separate `A` constraint therefore does not match the nested `!A`, so the impossible branch can reach hard-arithmetic deferral and return `SolverUnknown`. Consider retaining the aggregate while also adding the top-level conjuncts of a derived positive `And`. This should not flatten `Not(And(...))`, whose members are not individually implied.

stevencartavia (dismissed):

crates/evm/symbolic/src/runtime/solver.rs
```diff
@@ -396,6 +398,14 @@ impl SymbolicSolver for SmtLibSubprocessSolver {
             self.cache_model_result(cache_key, model.clone());
             return Ok(model);
         }
+        if let Some(model) = checked_mul_guard_branch_model(&smt_constraints)
+            && model_satisfies_constraints(&model, constraints)
+        {
+            trace!("model: validated constructive checked-multiply guard model");
+            self.cache_sat_result(cache_key.clone(), true);
+            self.cache_model_result(cache_key, model.clone());
+            return Ok(model);
```
mablr, line 477: **Important:** Please explicitly complete this local model before validating it. Missing symbols evaluate as zero here, but replay storage only writes symbols present in the map, so a model can validate under an implicit zero yet replay with the existing nonzero slot. Collect evaluation variables from the original constraints, insert zero for missing entries, then validate/cache the completed model.

crates/evm/symbolic/src/runtime/solver/hard_arith_fallback.rs
```diff
@@ -295,6 +460,41 @@ fn complete_fallback_support_model(constraints: &[SymBoolExpr], model: &mut Symb
...
+    model: &mut SymbolicModel,
+    remaining_support_visits: &mut usize,
+) -> Option<bool> {
+    let mut changed = false;
+    for constraint in constraints {
+        if *remaining_support_visits == 0 {
+            return None;
+        }
+        *remaining_support_visits -= 1;
+        match constraint.eval_model_if_complete(model) {
```
mablr, line 602: **Footgun:** This counts top-level constraint scans, not traversal work: `eval_model_if_complete` and support completion can recursively revisit a shared expression DAG exponentially within one charged visit. Please memoize/budget the expression traversal (or decline oversized support constraints), with a shared-DAG regression that survives normalization.

crates/evm/symbolic/src/runtime/solver/hard_arith_fallback.rs
```diff
@@ -110,6 +110,176 @@ pub(crate) fn hard_arith_fallback_model(
...
+/// semantic cases directly: the zero disjunct, a nonzero exact product, and wrapping products in
+/// either operand order. Simple support constraints are completed first so an exact operand value
+/// from the path is preserved instead of being overwritten by the semantic default. This does not
+/// perform the generic bounded candidate search, and a model is returned only when it satisfies
+/// every original constraint.
+pub(super) fn checked_mul_guard_branch_model(
+    constraints: &[SymBoolExpr],
+    original_constraints: &[SymBoolExpr],
+) -> Option<SymbolicModel> {
+    let mut remaining_support_visits = MAX_CHECKED_MUL_SUPPORT_VISITS;
```
mablr, line 142: **Blocking:** Reject non-replayable symbolic `Hash` and `GasLeft` leaves before constructing this model. For example, a true checked-multiply guard plus `b == 0` and `sha256([b]) == 0` validates after zero-completing the hash symbol, but replay recomputes SHA-256 and cannot apply that assignment. Likewise, `guard(x, y) && gasleft() == 0` can validate with `gasleft_0 = 0`, while replay cannot control remaining gas. Please return `None` when the original constraints contain a symbolic hash or `gasleft()`, matching the existing hard-arithmetic fallback policy.

crates/evm/symbolic/src/runtime/solver/hard_arith_fallback.rs
```diff
@@ -110,6 +110,182 @@ pub(crate) fn hard_arith_fallback_model(
...
+/// semantic cases directly: the zero disjunct, a nonzero exact product, and wrapping products in
+/// either operand order. Simple support constraints are completed first so an exact operand value
+/// from the path is preserved instead of being overwritten by the semantic default. This does not
+/// perform the generic bounded candidate search, and a model is returned only when it satisfies
+/// every original constraint.
+pub(super) fn checked_mul_guard_branch_model(
+    constraints: &[SymBoolExpr],
+    original_constraints: &[SymBoolExpr],
+) -> Option<SymbolicModel> {
+    if original_constraints.iter().any(SymBoolExpr::contains_symbolic_hash)
```
mablr: **Blocking:** This guard still treats every ordinary `Var` as replay-controllable, but several derived values share that representation. For example, combine a successful checked-multiply guard with `vm.computeCreateAddress(address(0), nonce) == address(0)`: the symbolic create address is an opaque `Var`, so support completion can validate `nonce = 0, create_address = 0`, while concrete replay computes a nonzero CREATE address. Random/SVM-derived variables have the same provenance issue. Please decline this constructive model when original constraints contain symbols that cannot be materialized through ABI/sequence inputs or registered replay storage.

crates/evm/symbolic/src/runtime/expr/expr.rs
```diff
@@ -1273,6 +1318,19 @@ impl SymExpr {
...
 
+    pub(in crate::runtime) fn bitwise_bool_word_condition(
+        &self,
+        cx: &mut SymCx,
+    ) -> Option<SymBoolExpr> {
+        if let Some(condition) = self.bool_word_condition() {
+            return Some(condition);
+        }
+        let SymExprKind::BinOp(SymBinOp::Or, left, right) = self.kind() else { return None };
+        let left = left.bitwise_bool_word_condition(cx)?;
```
mablr: **Blocking:** This recursively revisits shared `Or` DAG nodes per occurrence, with no memoization or budget. A linear Yul construction (`t := or(s, eq(b, i)); s := or(s, t)` repeated, then `sub(s, 1)`) doubles traversal work each iteration; about 26 iterations causes hundreds of millions of visits over roughly 53 unique OR nodes before executor limits can intervene. Please traverse unique expression identities with a shared budget and decline the optional rewrite when exhausted.

mablr (dismissed):

stevencartavia (commented):
test failures look related

mablr:
Re-triggered, seems fine now

stevencartavia (dismissed):

mablr (dismissed):

mattsse (dismissed):

mablr (approved):

mattsse (approved):
lgtm

PR #16094, opened 2026-08-08
feat(symbolic): normalize branchless word selection

figtracer:
Stack: #16087 → #16093 → #16094 → #16095.

This is the second PR in the symbolic-completeness stack and is based on #16093. It recognizes the exact branchless XOR-selection identity `a ^ ((a ^ b) * bool_word(c))` as `c ? b : a`, which is the form used by Solady's P256 signature normalization and several other assembly libraries.

On pinned Solady v0.1.26, `check_P256Normalized` changes from incomplete (`solver returned unknown`) to proven. Paths remain 2, solver queries fall from 6 to 4, and both SMT queries and heuristic witnesses remain 0. Across the full 13-property slice, the stacked result is now 12 pass / 1 incomplete, 166 solver queries, and 24 heuristic witnesses, compared with 10 / 3 / 168 / 28 at #16087. A 20-run focused comparison was neutral at 1.278s baseline versus 1.273s here.

Selector recovery is iterative, capped at 256 unique word nodes, and deduplicates recovered leaves before constructing one Boolean OR. On an adversarial overlapping `w(n) = w(n-1) | w(n-2)` DAG, six alternating warmed direct-binary samples reduced median runtime from 55 ms to 10 ms and median peak RSS from 42.6 MB to 11.2 MB (about 74%).

Validated with the complete 272-test `foundry-evm-symbolic` suite, a minimized Forge CLI regression, the full pinned Solady slice, `cargo check -p foundry-evm-symbolic -p forge`, formatting, and `git diff --check`. Nightly clippy for `foundry-evm-symbolic` passes with all targets and warnings denied.

This PR was developed with AI assistance. The implementation and benchmark results were manually reviewed and validated.

PR #16095, opened 2026-08-08
feat(symbolic): prove branchless operation states

figtracer:
Stack: #16087 → #16093 → #16094 → #16095.

This is the third and final PR in the symbolic-completeness stack, based on #16094. It turns multiplication and wrapping offset arithmetic over exact EVM 0/1 words into conditionals, simplifies the associated XOR forms, and expands equality only for bounded conditional trees with constant leaves. These are exact structural rewrites; unsupported or over-budget shapes retain their original expression and continue through the normal solver path.

The review hardening keeps every new traversal local, memoized for shared DAGs, and capped at 256 visited nodes. The implementation adds no public configuration or data fields: helpers are private except for narrow sibling-module access where the existing module split requires it.

On pinned Solady v0.1.26, `check_OperationStateDifferentialTrick` changes from incomplete to proven with the same 6 paths. Solver queries fall from 73 to 59 and heuristic witnesses fall from 24 to 0; SMT queries rise from 13 to 37 because branches that previously stopped at heuristic witnesses are now certified.

An exact-master run with the repository's symbolic benchmark harness produced:

| suite | master | complete stack |
| --- | --- | --- |
| Solady status | 9 pass / 4 incomplete | 13 pass / 0 incomplete |
| Solady paths | 35 | 35 |
| Solady solver queries | 172 | 148 |
| Solady SMT queries | 56 | 76 |
| Solady heuristic witnesses | 30 | 0 |
| Solady wall mean | 331.65 ms | 332.93 ms (+0.39%) |
| Farcaster | 2 pass; 14 paths; 14 solver / 9 SMT queries | unchanged; wall -1.33% |

The extra SMT work is intentional: all four formerly incomplete Solady properties are now solver-certified. No path count or failure status changed. A final rerun at `bacc130a2` reproduced all 13 proofs, 35 paths, 148 solver queries, 76 SMT queries, and 0 heuristic witnesses. An additional exact-profile comparison was neutral: the 20-run median was 1.233 s on master versus 1.230 s on the final stack, while six alternating warmed samples measured median peak RSS of 676.6 MB versus 677.0 MB (+0.05%).

Validation includes all 286 `foundry-evm-symbolic` unit tests, four optimized Forge CLI proof/replay regressions, `cargo check -p foundry-evm-symbolic -p forge`, rustfmt, `git diff --check`, and nightly clippy for `foundry-evm-symbolic` with all targets and warnings denied.

This PR was developed with AI assistance. The implementation and benchmark results were manually reviewed and validated.


figtracer:
derek bench symex timeout=1800

crates/evm/symbolic/src/runtime/expr/expr.rs
```diff
@@ -637,6 +664,16 @@ impl SymExpr {
                 (SymExprKind::Const(value), _) if *value == U256::ONE => right,
                 // `a * 1 => a`.
                 (_, SymExprKind::Const(value)) if *value == U256::ONE => left,
+                // `bool_word(c) * a => ite(c, a, 0)`.
```
mablr: **Important:** When both operands match, this left-first dispatch makes commutative construction noncanonical: `mul(bool_word(c), bool_word(d))` and its reverse produce different hash-consed ITEs. The new dual-match `Add` and `Xor` helpers have the same issue, causing missed simplifications/cache hits. Deterministically order operands before symmetric rewrites.

crates/evm/symbolic/src/runtime/expr/bool.rs
```diff
@@ -315,9 +315,31 @@ impl SymBoolExpr {
             let condition = Self::not_bool(cx, condition.clone());
             return Some(Self::or(cx, vec![condition, then_matches]));
         }
+        if expected.as_const().is_some()
```
mablr: **Footgun:** Expanding every constant-leaf ITE recursively can take exponential work on a compact shared DAG, and `ite_has_only_constant_leaves` repeats the same traversal. Memoize by `(SymExpr, expected_constant)` or enforce a node/leaf work budget and retain the original equality when exceeded.

crates/evm/symbolic/src/tests.rs
```diff
@@ -2688,6 +2688,59 @@ fn expression_simplifies_xor_bool_select_idiom() {
     assert_eq!(actual, expected);
 }
 
+#[test]
+fn expression_simplifies_bool_word_multiplication() {
```
mablr, line 2801: Please assert identical results for reversed operands across dual-match `Mul`, `Add`, and `Xor`. Also cover `k = 0/MAX` wrapping offsets, direct and inverted boolean words, two boolean-word operands, and the distinction between 0/1 words and 0/MAX masks.

crates/evm/symbolic/src/runtime/expr/expr.rs
```diff
@@ -1335,13 +1434,43 @@ impl SymExpr {
...
-        let SymExprKind::BinOp(SymBinOp::Or, left, right) = self.kind() else { return None };
-        let left = left.bitwise_bool_word_condition(cx)?;
-        let right = right.bitwise_bool_word_condition(cx)?;
-        Some(SymBoolExpr::or(cx, vec![left, right]))
+        if let SymExprKind::BinOp(SymBinOp::Or, left, right) = self.kind() {
+            let left = left.bitwise_bool_word_condition(cx)?;
+            let right = right.bitwise_bool_word_condition(cx)?;
+            return Some(SymBoolExpr::or(cx, vec![left, right]));
+        }
+        if self.unsigned_bits() != 1 {
```
mablr: **Important:** This calls uncached `unsigned_bits()` during `Add`/`Sub` construction. A compact DAG `p0 = x & 1; p(n+1) = p(n) + p(n); p64 + 2` has 65 distinct nodes but triggers roughly `2^64` visits. Please memoize sizing per match, add a work budget, or restrict construction-time matching to cheap structural forms; test the actual construction boundary.

crates/evm/symbolic/src/runtime/expr/bool.rs
```diff
@@ -315,9 +315,44 @@ impl SymBoolExpr {
...
+        }
+        let result = match expr.kind() {
+            SymExprKind::Const(value) => Some(Self::constant(cx, *value == expected)),
+            SymExprKind::Ite(condition, then_expr, else_expr) => {
+                let condition = condition.clone();
+                let then_matches = Self::constant_ite_eq(cx, then_expr, expected, cache);
+                let else_matches = Self::constant_ite_eq(cx, else_expr, expected, cache);
+                match (then_matches, else_matches) {
+                    (Some(then_matches), Some(else_matches)) => {
+                        let then_selected = Self::and(cx, vec![condition.clone(), then_matches]);
```
mablr: **Footgun:** Memoization fixes repeated DAG traversal, but flattening and caching each intermediate conjunction in a linear ITE chain retains lengths `1..n`, yielding quadratic handle construction and linear recursion depth. Consider a modest work budget or a representation that avoids flattening every cached intermediate.

crates/evm/symbolic/src/runtime/expr/expr.rs
```diff
@@ -600,7 +600,33 @@ impl SymExpr {
                     let max = Self::constant(cx, U256::MAX);
                     Self::ite(cx, condition, zero, max)
                 }
-                _ => Self::commutative_binop(cx, binop, left, right),
+                // `one_bit_word(c) + k => ite(c, k + 1, k)`.
+                (SymExprKind::Const(value), _)
+                    if right.bool_word_condition().is_none()
```
mablr: Equivalent 0/1 encodings normalize asymmetrically here: `ite(c,1,0) + 2` is excluded and remains `Add`, while `ite(!c,0,1) + 2` rewrites to an ITE. This is sound but makes proof/cache shape representation-dependent. Normalize both consistently or document the exclusion, with direct/inverted Add/Sub tests in both orders.

crates/evm/symbolic/src/runtime/expr/bool.rs
```diff
@@ -315,9 +340,52 @@ impl SymBoolExpr {
...
+
+        let result = match expr.kind() {
+            SymExprKind::Const(value) => Some(Self::constant(cx, *value == expected)),
+            SymExprKind::Ite(condition, then_expr, else_expr) => {
+                let condition = condition.clone();
+                let then_matches = Self::constant_ite_eq(cx, then_expr, expected, cache, remaining);
+                let else_matches = Self::constant_ite_eq(cx, else_expr, expected, cache, remaining);
+                match (then_matches, else_matches) {
+                    (Some(then_matches), Some(else_matches)) => {
+                        let then_selected = Self::and(cx, vec![condition.clone(), then_matches]);
```
mablr: **Important:** The per-call budget bounds this expansion, but repeated equalities rebuild flattened intermediate conjunctions that die after return while leaving weak hash-cons entries behind. Across Q calls this can cause O(n·Q²) bucket scanning and O(n·Q) retained weak entries. Consider a bounded cross-call `(SymExpr, constant)` cache or avoid hash-consing every flattened intermediate.

crates/evm/symbolic/src/runtime/expr/expr.rs
```diff
@@ -1335,13 +1432,59 @@ impl SymExpr {
...
+        bit_widths: &mut HashMap<Self, usize>,
+    ) -> Option<SymBoolExpr> {
+        if let Some(condition) = conditions.get(self) {
+            return condition.clone();
+        }
+
+        let condition = if let Some(condition) = self.direct_bool_word_condition(cx) {
+            Some(condition)
+        } else if let SymExprKind::BinOp(SymBinOp::Or, left, right) = self.kind() {
+            let left = left.bitwise_bool_word_condition_cached(cx, conditions, bit_widths);
```
mablr: **Footgun:** Memoization removes exponential DAG revisits, but this analysis remains recursively unbudgeted; a unique chain near the opcode limit can still cause deep Rust recursion and repeated linear scans. A modest node/depth budget or iterative traversal would harden it. Also consider short-circuiting when the left OR child is unrecognized instead of eagerly traversing both sides.

crates/evm/symbolic/src/tests.rs
```diff
@@ -2773,6 +2963,61 @@ fn ite_equality_with_selected_arm_exposes_branch_condition() {
...
+    let mut shared = SymExpr::ite(&mut cx, first.clone(), zero.clone(), one.clone());
+
+    for _ in 0..32 {
+        let left = SymExpr::ite(&mut cx, first.clone(), shared.clone(), zero.clone());
+        let right = SymExpr::ite(&mut cx, second.clone(), shared.clone(), one.clone());
+        shared = SymExpr::ite(&mut cx, third.clone(), left, right);
+    }
+
+    let matches = SymBoolExpr::eq(&mut cx, shared, one);
+    assert!(matches.as_const().is_none());
```
mablr: This assertion does not prove memoization succeeded: without memoization, the 128-node budget can return the original nonconstant equality and this still passes. Assert that the result differs from the raw comparison fallback or matches the expected expanded condition shape.

figtracer:
derek bench symex timeout=1800

crates/evm/symbolic/src/tests.rs
```diff
@@ -2746,6 +2746,196 @@ fn expression_does_not_treat_max_mask_as_bool_select() {
...
+    let one = SymExpr::one(&mut cx);
+    let two = SymExpr::constant(&mut cx, U256::from(2));
+    let three = SymExpr::constant(&mut cx, U256::from(3));
+    let four = SymExpr::constant(&mut cx, U256::from(4));
+    let first_offset = SymExpr::ite(&mut cx, first_condition.clone(), one, two);
+    let second_offset = SymExpr::ite(&mut cx, second_condition.clone(), three, four);
+    let add_forward =
+        SymExpr::binop(&mut cx, SymBinOp::Add, first_offset.clone(), second_offset.clone());
+    let add_reverse = SymExpr::binop(&mut cx, SymBinOp::Add, second_offset, first_offset);
+    assert_eq!(add_forward, add_reverse);
```
mablr: Please also assert the expected rewritten `Ite` shape for each dual-match case. Forward/reverse equality alone still passes if a helper is removed and both constructions fall back to the canonical raw operator.

mablr (commented):
LGTM, just a coverage suggestion.

mablr (dismissed):

stevencartavia (dismissed):

crates/evm/symbolic/src/runtime/expr/bool.rs
```diff
@@ -315,9 +340,68 @@ impl SymBoolExpr {
...
+            cache.insert(expr.clone(), None);
+            return None;
+        }
+        *remaining -= 1;
+
+        let result = match expr.kind() {
+            SymExprKind::Const(value) => Some(Self::constant(cx, *value == expected)),
+            SymExprKind::Ite(condition, then_expr, else_expr) => {
+                let condition = condition.clone();
+                let then_matches = Self::constant_ite_eq(cx, then_expr, expected, cache, remaining);
```
mablr, line 441: **Blocking:** Memoization keeps construction linear, but does not bound downstream work. A shared ITE DAG can produce a Boolean DAG where each level references the prior result twice; the solver’s non-memoized `fold` then unfolds it exponentially before SMT CSE runs. Please preflight/cap unfolded occurrence cost and fall back to the raw comparison when exceeded. Preflighting also avoids leaving dead weak hash-cons entries when a later unsupported arm makes expansion fail.

crates/evm/symbolic/src/runtime/expr/expr.rs
```diff
@@ -1346,27 +1443,63 @@ impl SymExpr {
...
+
+            if let SymExprKind::BinOp(SymBinOp::Or, left, right) = word.kind() {
+                pending.push(right.clone());
+                pending.push(left.clone());
+                continue;
+            }
+
+            if word.unsigned_bits_cached(&mut bit_widths, &mut remaining) == Some(1) {
+                let zero = Self::zero(cx);
+                leaf_conditions.insert(SymBoolExpr::eq(cx, word, zero).not(cx));
```
mablr: **Footgun:** The 256-node budget does not include this simplifying equality: every one-bit leaf can start a fresh 128-node ITE expansion. Overlapping leaves therefore multiply work hidden behind one charged node. Please use a raw canonical zero comparison here or share/charge the equality-expansion budget.

crates/evm/symbolic/src/runtime/expr/expr.rs
```diff
@@ -911,6 +967,21 @@ impl SymExpr {
...
+        }
+        let SymExprKind::Ite(condition, then_expr, else_expr) = conditional.kind() else {
+            return None;
+        };
+        if then_expr.as_const().is_none() || else_expr.as_const().is_none() {
+            return None;
+        }
+        let then_expr = Self::binop(cx, SymBinOp::Add, other.clone(), then_expr.clone());
+        let else_expr = Self::binop(cx, SymBinOp::Add, other.clone(), else_expr.clone());
+        Some(Self::ite(cx, condition.clone(), then_expr, else_expr))
```
mablr, line 989: **Blocking:** This distribution duplicates `other` into both ITE arms without bounding unfolded occurrences. With `acc₀ = a`, `offsetᵢ = ite(cᵢ, 2, 3)`, and `accᵢ₊₁ = accᵢ + offsetᵢ`, each rewrite produces `ite(cᵢ, accᵢ + 2, accᵢ + 3)`. The interned DAG is linear, but `normalize_expr_for_solver`’s non-memoized fold visits `accᵢ` twice per level, becoming exponential before SMT CSE. Please preflight/cap the resulting unfolded cost and retain the raw canonical `Add` when exceeded.

crates/evm/symbolic/src/runtime/expr/expr.rs
```diff
@@ -938,6 +1009,32 @@ impl SymExpr {
...
+        }
+    }
+
+    fn xor_with_zero_ite(cx: &mut SymCx, base: &Self, conditional: &Self) -> Option<Self> {
+        let SymExprKind::Ite(condition, then_expr, else_expr) = conditional.kind() else {
+            return None;
+        };
+        if then_expr.as_const().is_some_and(|value| value.is_zero()) {
+            let selected = Self::binop(cx, SymBinOp::Xor, base.clone(), else_expr.clone());
+            return Some(Self::ite(cx, condition.clone(), base.clone(), selected));
```
mablr, line 1039: **Blocking:** This rewrite duplicates `base` into the selected XOR and the unselected arm. Repeating `accᵢ₊₁ = accᵢ ^ ite(cᵢ, sᵢ, 0)` creates a linear hash-consed DAG but an unfolded tree with `2^n` occurrences; solver normalization recursively folds that tree before CSE. Please apply an unfolded-occurrence budget and fall back to the raw canonical `Xor` once duplication would exceed it.

mablr (dismissed):

stevencartavia (dismissed):

mablr (approved):

stevencartavia (approved):

PR #16096, opened 2026-08-08
fix(wallet): reject BIP32 paths that overflow the harden bit

SashaMIT:
## Summary
`coins-bip32` (via `MnemonicBuilder::derivation_path`) parses hardened segments with `harden_index(v) = v + 0x80000000`. For `2147483648'`:

- `v = 0x80000000`
- harden wraps to `0`
- Foundry silently derives the **unhardened** account-0 key instead of failing

Raw `2147483648` (no `'`) already equals the harden bit and selects `0'` without a harden marker.

geth already rejects this (`accounts/hd.go` + `"m/2147483648'"` test). This adds an equivalent guard at the Foundry boundary before mnemonic derivation (sibling of [alloy-rs/alloy#4124](https://github.com/alloy-rs/alloy/pull/4124)):

- `foundry_common::wallet::{validate_bip32_path, derive_key_path}`
- cheatcodes `derive_wallets` / `deriveKey` path (via common helpers)
- anvil `AccountGenerator::generate`

Optional follow-up: fix upstream in `summa-tx/coins` `try_parse_index` and bump the pin.

## Test plan
- [x] `cargo test -p foundry-common wallet::tests --lib` → 4/4
- [x] `cargo check -p foundry-cheatcodes -p anvil --lib`

Made with [Cursor](https://cursor.com)


crates/common/src/wallet.rs
```diff
@@ -7,14 +7,55 @@ use alloy_signer_local::{
...
+/// silently selecting the wrong (unhardened) child. Raw `2147483648` is also rejected
+/// because it already equals the harden bit (`0'`).
+pub fn validate_bip32_path(path: &str) -> Result<(), String> {
+    for (i, raw) in path.split('/').enumerate() {
+        let c = raw.trim();
+        if c.is_empty() || (i == 0 && (c == "m" || c == "M")) {
+            continue;
+        }
+
+        let num = c.trim_end_matches(['\'', 'h', 'H']);
```
mablr: Could we keep this as a narrow overflow guard rather than a second path parser? Trimming whitespace, accepting `M`/`H`, skipping empty segments, and rejecting every parse error changes the grammar and errors from `coins-bip32`. Please strip one terminal `'`/`h`, reject only parsed values >= `BIP32_HARDEN`, and leave malformed syntax to `MnemonicBuilder`.

crates/common/src/wallet.rs
```diff
@@ -7,14 +7,55 @@ use alloy_signer_local::{
...
+    }
+    Ok(())
+}
+
 /// Appends `index` to `path`, inserting a `/` separator when needed.
-pub fn derive_key_path(path: &str, index: u32) -> String {
+///
+/// Rejects indices that already set the harden bit (would silently select hardened
+/// children when concatenated as a raw decimal segment).
+pub fn derive_key_path(path: &str, index: u32) -> Result<String, String> {
```
mablr: Can we preserve `derive_key_path(path, index) -> String` and add a checked helper for these consumers? Changing this public helper’s return type is unnecessary for the boundary fix.

crates/common/src/wallet.rs
```diff
@@ -66,3 +107,50 @@ pub fn private_key_from_u256(private_key: U256) -> Result<PrivateKeySigner, Stri
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // anvil default test mnemonic
+    const MNEMONIC: &str =
+        "test test test test test test test test test test test junk";
+
+    #[test]
+    fn validate_rejects_raw_harden_bit_index() {
```
mablr: Please add regressions at the exposed boundaries (`deriveKey`/`rememberKeys` and Anvil), not only helper tests. It would also be useful to cover `h`, `+2147483648`, leading zeros, `u32::MAX`, and the largest valid index.

SashaMIT:
Thanks, that is the right cut.

Overflow guard only now: strip one trailing ' or h, reject parsed values >= BIP32_HARDEN, leave malformed syntax to MnemonicBuilder.

derive_key_path is String again. Consumers use derive_key_path_checked.

Regressions added on deriveKey / rememberKeys and Anvil generate, plus h, +2147483648, leading zeros, u32::MAX, and the largest valid index.

Changelog fragment is in .changelog/pr-16096.md.

stevencartavia (commented):
code-wise looks good, needs conflicts and `cargo +nightly fmt --all`

SashaMIT:
Rebased on current master (kept the new monad hardfork test) and ran `cargo +nightly fmt --all`.

stevencartavia (commented):
also `cargo +nightly clippy --workspace --all-targets --all-features`

SashaMIT:
Ran `cargo +nightly clippy --workspace --all-targets --all-features`. One warning in our helper (`collapsible_if`); collapsed it in b4c5426.

mablr (approved):
SGTM, thanks!

stevencartavia (approved):

SashaMIT:
Thanks for landing this. Glad the overflow-only cut was the one that shipped.

wyattloveslife-create (commented):
Thanks

PR #16098, opened 2026-08-09
fix(coverage): report file-level (free) functions

### LINKED ISSUE #16085: forge coverage ignores file-level free functions; free-function-only files get no LCOV SF record
This is unrelated to #16084, which concerns source-ID collisions between compilation jobs. The two look superficially similar (files missing from coverage) but have different causes and would be fixed in different places.

`forge coverage` never creates coverage items for file-level (free) functions. A file containing only free functions gets no `SF:` record at all; a file mixing free functions with a contract gets a record that silently omits the free functions' lines, and can then be reported as 100% covered.

### Version

Reproduced on the current nightly, in a clean container:

```
forge Version: 1.8.0-nightly
Commit SHA: 7c271517a32620b5b181d25abfd86db970ad4d7d
Build Timestamp: 2026-08-08T06:14:06.447311930Z (1786169646)
Build Profile: dist
```

Originally found on 1.7.2-nightly (`160b60260db63ce6204f2ee15764aca3e9ef04fe`).

### Mechanism

`SourceAnalysis::new` walks only contract definitions:

```rust
for item in ast.items.iter() {
    // Visit only top-level contracts.
    let ItemKind::Contract(contract) = &item.kind else { continue };
```

Every other top-level item — including a free-function AST item — is skipped, before anchors, hit counting or any reporter runs.

https://github.com/foundry-rs/foundry/blob/7c271517a32620b5b181d25abfd86db970ad4d7d/crates/evm/coverage/src/analysis.rs#L493-L507

### Repro A — free-function-only file gets no record

`foundry.toml`:

```toml
[profile.default]
src = "src"
test = "test"
```

`src/Free.sol`:

```solidity
pragma solidity =0.8.35;
function double(uint256 x) pure returns (uint256) { return x * 2; }
```

`src/User.sol`:

```solidity
pragma solidity =0.8.35;
import {double} from "./Free.sol";
contract User { function run(uint256 x) external pure returns (uint256) { return double(x); } }
```

`test/UserTest.t.sol`:

```solidity
pragma solidity =0.8.35;
import {User} from "../src/User.sol";
contract UserTest { function testRun() public { require(new User().run(2) == 4); } }
```

`forge coverage --report lcov` — the test passes, `double` executes, and the only `SF:` entry is:

```
SF:src/User.sol
```

### Repro B — mixed file reports 100% while ignoring the free function

Replace `src/Free.sol` and `src/User.sol` with a single `src/Mixed.sol`, and change the test's import to `import {User} from "../src/Mixed.sol";`:

```solidity
pragma solidity =0.8.35;
function double(uint256 x) pure returns (uint256) { return x * 2; }
contract User { function run(uint256 x) external pure returns (uint256) { return double(x); } }
```

The record is emitted, but covers only line 3 (relevant entries):

```
SF:src/Mixed.sol
DA:3,1
FN:3,User.run
FNDA:1,User.run
LF:1
LH:1
```

`double` on line 2 produces no coverage item. `--report summary` then prints:

```
| src/Mixed.sol | 100.00% (1/1) | 100.00% (2/2) | N/A (0/0)  | 100.00% (1/1) |
```

100% for a file one of whose two functions is never measured. Both reporters agree, so this is not an LCOV serialization problem — the items were never created.

Repro B also separates the two effects: file filtering is not the cause, since the file *is* reported. It is coverage-item discovery.

### Expected

Coverage items for free functions, so their bodies are measured like any other executable code — and so a file containing them cannot report 100% while part of it goes unmeasured.

### Impact

Solidity requires user-defined operator implementations to be free functions, so this particularly affects arithmetic and UDVT utility modules.

A public real-world example is [flare-foundation/flare-smart-contracts-v2](https://github.com/flare-foundation/flare-smart-contracts-v2), where [`FixedPointArithmetic.sol`](https://github.com/flare-foundation/flare-smart-contracts-v2/blob/b69873e1e1a0785e2450d811f35c7927a625716b/contracts/fastUpdates/lib/FixedPointArithmetic.sol) declares 27 file-level functions and [`Sortition.sol`](https://github.com/flare-foundation/flare-smart-contracts-v2/blob/b69873e1e1a0785e2450d811f35c7927a625716b/contracts/fastUpdates/lib/Sortition.sol) declares 3. On the affected branch both have dedicated, passing Forge test suites; neither file appears in any `forge coverage` report under any Foundry configuration we tried. For free-function-only files, execution and non-execution are indistinguishable in forge's output: both produce no record at all, where unexecuted contract code would normally produce zero-hit entries.

zexoverz:
## Motivation

`forge coverage` never reports file-level (free) functions. `SourceAnalysis::new` only walked `ItemKind::Contract`, so a source file whose executable code lives in free functions is analyzed as if it had no code and shows no coverage.

Closes #16085

## Solution

Match on `ItemKind` in the top-level loop and also visit `ItemKind::Function` via `ast::Visit::visit_item`, so file-level functions get coverage items the same way contract functions do. Contract handling (interface skip, test-file checkpointing) is unchanged.

## Tests

Adds `coverage_reports_free_functions`, which asserts a source file whose only code is a free function appears in the coverage report. It fails on `master` and passes with the fix.

### DIFF
--- a/crates/evm/coverage/src/analysis.rs
+++ b/crates/evm/coverage/src/analysis.rs
@@ -5,7 +5,7 @@ use foundry_compilers::ProjectCompileOutput;
 use rayon::prelude::*;
 use solar::{
     ast::{self, ExprKind, ItemKind, StmtKind, yul},
-    data_structures::{Never, map::FxHashSet},
+    data_structures::{Never, map::FxHashMap},
     interface::{BytePos, Span},
     sema::{Gcx, hir},
 };
@@ -33,8 +33,10 @@ struct SourceVisitor<'gcx> {
     items: Vec<CoverageItem>,
 
     all_lines: Vec<u32>,
-    function_calls: Vec<Span>,
-    function_calls_set: FxHashSet<Span>,
+    /// Deferred function-call spans, each paired with the contract scope active where the call
+    /// was collected, so scope travels with the span to the delayed HIR resolution pass.
+    function_calls: Vec<(Span, Arc<str>)>,
+    function_call_scopes: FxHashMap<Span, Arc<str>>,
 }
 
 struct SourceVisitorCheckpoint {
@@ -52,7 +54,7 @@ impl<'gcx> SourceVisitor<'gcx> {
             branch_id: 0,
             all_lines: Default::default(),
             function_calls: Default::default(),
-            function_calls_set: Default::default(),
+            function_call_scopes: Default::default(),
             items: Default::default(),
         }
     }
@@ -109,7 +111,7 @@ impl<'gcx> SourceVisitor<'gcx> {
     }
 
     fn resolve_function_calls(&mut self, hir_source_id: hir::SourceId) {
-        self.function_calls_set = self.function_calls.iter().copied().collect();
+        self.function_call_scopes = self.function_calls.iter().cloned().collect();
         let _ = hir::Visit::visit_nested_source(self, hir_source_id);
     }
 
@@ -325,8 +327,8 @@ impl<'ast> ast::Visit<'ast> for SourceVisitor<'_> {
                 }
             }
             ExprKind::Call(callee, _args) => {
-                // Resolve later.
-                self.function_calls.push(expr.span);
+                // Resolve later. Capture the current contract scope so it travels with the span.
+                self.function_calls.push((expr.span, self.contract_name.clone()));
 
                 if let ExprKind::Ident(ident) = &callee.kind {
                     // Might be a require call, add branch coverage.
@@ -421,10 +423,14 @@ impl<'gcx> hir::Visit<'gcx> for SourceVisitor<'gcx> {
 
     fn visit_expr(&mut self, expr: &'gcx hir::Expr<'gcx>) -> ControlFlow<Self::BreakValue> {
         if let hir::ExprKind::Call(lhs, ..) = &expr.kind
-            && self.function_calls_set.contains(&expr.span)
             && is_regular_call(lhs)
+            && let Some(scope) = self.function_call_scopes.get(&expr.span).cloned()
         {
+            // Attribute the call with the scope captured where it was collected, not the
+            // visitor's final scope (which leaks across free functions and contracts).
+            let prev = std::mem::replace(&mut self.contract_name, scope);
             self.push_stmt(expr.span);
+            self.contract_name = prev;
         }
         self.walk_expr(expr)
     }
@@ -510,6 +516,10 @@ impl SourceAnalysis {
                             // defined inside a contract. Without this, a file of only free
                             // functions gets no coverage record at all.
                             ItemKind::Function(_) => {
+                                // A free function is not scoped to any contract. Clear any scope
+                                // left by a previously visited contract so it is attributed at
+                                // file level, not as `Contract.freeFn`.
+                                visitor.contract_name = Arc::default();
                                 let _ = ast::Visit::visit_item(&mut visitor, item);
                             }
                             _ => {}
--- a/crates/evm/coverage/src/analysis.rs
+++ b/crates/evm/coverage/src/analysis.rs
@@ -464,11 +464,12 @@ pub struct SourceAnalysis {
 }
 
 impl SourceAnalysis {
-    /// Analyzes contracts in the sources held by the source analyzer.
+    /// Analyzes the sources held by the source analyzer.
     ///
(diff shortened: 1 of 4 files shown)


crates/evm/coverage/src/analysis.rs
```diff
@@ -491,18 +492,27 @@ impl SourceAnalysis {
...
+                                visitor.visit_contract(contract);
+                                if visitor.has_tests(&checkpoint) {
+                                    visitor.restore_checkpoint(checkpoint);
+                                }
+                            }
+                            // File-level (free) functions are covered too, not only functions
+                            // defined inside a contract. Without this, a file of only free
+                            // functions gets no coverage record at all.
+                            ItemKind::Function(_) => {
+                                let _ = ast::Visit::visit_item(&mut visitor, item);
```
mablr, line 524: Please preserve explicit scope here. `visit_contract` leaves `contract_name` set, so a free function declared after a contract is attributed as `Contract.freeFn`; a free-only function is emitted by LCOV as `.freeFn`. Delayed call resolution also uses the visitor’s final scope, so scope should travel with collected call spans.

crates/forge/tests/cli/coverage.rs
```diff
@@ -2975,6 +2975,40 @@ contract ReportScript {
...
+        require(double(3) == 6);
+    }
+}
+"#,
+    );
+
+    let output = cmd.arg("coverage").assert_success().get_output().stdout_lossy();
+    // Before the fix, a file of only free functions produced no coverage record at all.
+    assert!(
+        output.contains("src/Free.sol"),
```
mablr: Could this assert exact LCOV output and cover the issue’s mixed-file case? Checking only the path misses malformed function names, incorrect hits/denominators, and order-dependent attribution. Please assert `SF`, function names/hits, and both functions in a contract-before-free-function fixture.

mablr (commented):
Thanks for your contribution, please have a look to inline comments.

zexoverz:
Thanks for the review, addressed all three. Free functions now clear the contract scope before being visited, so they land at file level instead of Contract.freeFn. Deferred call spans carry the scope they were collected in, so resolution no longer uses the visitor final scope. And the LCOV reporter emits the bare name when there is no contract scope, which fixes the leading-dot .freeFn. The regression test is now a contract-before-free-function fixture that asserts SF, both function names, and nonzero hits.

crates/evm/coverage/src/analysis.rs
```diff
@@ -491,18 +498,31 @@ impl SourceAnalysis {
...
+                                let checkpoint = visitor.checkpoint();
+                                visitor.visit_contract(contract);
+                                if visitor.has_tests(&checkpoint) {
+                                    visitor.restore_checkpoint(checkpoint);
+                                }
+                            }
+                            // File-level (free) functions are covered too, not only functions
+                            // defined inside a contract. Without this, a file of only free
+                            // functions gets no coverage record at all.
+                            ItemKind::Function(_) => {
```
mablr, line 519: Adding free functions exposes scope-insensitive disambiguation: a contract method `C.f` and file-level `f` are renamed to `C.f.0` and `f.1`, although neither is overloaded. Could the disambiguation key include `item.loc.contract_name` and add a mixed-scope same-name regression?

crates/forge/tests/cli/coverage.rs
```diff
@@ -2975,6 +2975,80 @@ contract ReportScript {
...
+        require(c.bump(2) == 6);
+    }
+}
+"#,
+    );
+
+    cmd.args(["coverage", "--report=lcov", "--lcov-version=2"]).assert_success();
+    let lcov = std::fs::read_to_string(prj.root().join("lcov.info")).unwrap();
+
+    // The file is reported at all (before the fix a free-function-only file was dropped).
```
mablr: This mixed-file `SF` assertion does not cover the issue’s free-function-only failure mode; the original free-only fixture was replaced. Could we keep a free-only case too, ideally with an exact LCOV record?

zexoverz:
Thanks, addressed both. Disambiguation now keys on the function scope, so a contract method and a same-named free function are no longer renamed as overloads of each other, and there is a regression test for that. I also kept a free-function-only fixture alongside the mixed one and assert its exact LCOV record. Merged in master too, which resolves the conflict with the build-id source keying that just landed.

mablr (approved):
LGTM, thanks.

stevencartavia (approved):

PR #16099, opened 2026-08-09
feat(cast): add cast memory subcommand

### LINKED ISSUE #16088: feat(cast) : Adding `cast memory` (alias `cast mem`) helper for bytes memory layout
### Component
Cast

### Describe the feature you'd like

A convenience command that produces the full in-memory representation of a `bytes` value (the 32-byte length prefix + the data) in one step.

#### Proposed CLI
```bash
cast memory <type> [args...]
cast mem <type> [args...]    # short alias
```

#### Example
```bash
cast memory "transfer(address,uint256)" 0xabc123.... 200 -> 0x[byte memory length in 32 bytes] [4 bytes selector] [memory data payload in 32 bytes]
```

### Additional context

_No response_

zexoverz:
Adds a cast memory subcommand (alias cast mem) that ABI-encodes a function signature and args like cast calldata, then lays the result out as a Solidity bytes memory value: a 32-byte big-endian length prefix followed by the data right-padded to a whole number of 32-byte words. The layout is factored into a bytes_memory_layout helper with unit tests and a doctest. Closes #16088.

### DIFF
--- a/.changelog/cast-memory.md
+++ b/.changelog/cast-memory.md
@@ -1,0 +1,5 @@
+---
+cast: patch
+---
+
+Add `cast memory` (alias `cast mem`) to lay ABI-encoded calldata out as a `bytes` memory value.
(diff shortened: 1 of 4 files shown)


mablr (commented):
Thanks for your contribution @zexoverz. Unfortunately it was superseded by #16103.

PR #16100, opened 2026-08-09
fix(coverage): key sources by build ID

### LINKED ISSUE #16084: forge coverage silently drops and misattributes sources when one solc version has multiple compilation jobs
solc assigns source IDs per compilation job, starting at 0. `CoverageReport` keys source identity by `(Version, source_id)` and holds one `SourceAnalysis` per version:

```rust
pub source_paths: HashMap<(Version, usize), PathBuf>,
pub source_paths_to_ids: HashMap<(Version, PathBuf), usize>,
pub analyses: HashMap<Version, SourceAnalysis>,
```

So when a project produces two compilation jobs on the **same** solc version — which `additional_compiler_profiles` + `compilation_restrictions` do, and which is the documented way to give some contracts `via_ir` and others not — their source IDs collide. One path wins each slot. The other is dropped before hit-counting, and its hits can be credited to the winner.

### Version

Reproduced on the current nightly, in a clean container:

```
forge Version: 1.8.0-nightly
Commit SHA: 7c271517a32620b5b181d25abfd86db970ad4d7d
Build Timestamp: 2026-08-08T06:14:06.447311930Z (1786169646)
Build Profile: dist
```

Originally found on 1.7.2-nightly (`160b60260db63ce6204f2ee15764aca3e9ef04fe`).

The identity maps are unchanged at that commit: https://github.com/foundry-rs/foundry/blob/7c271517a32620b5b181d25abfd86db970ad4d7d/crates/evm/coverage/src/lib.rs#L34-L79

### Minimal repro (no forge-std needed)

`foundry.toml`:

```toml
[profile.default]
src = "src"
test = "test"
optimizer = false
via_ir = false

[[profile.default.additional_compiler_profiles]]
name = "via-ir"
via_ir = true

[[profile.default.compilation_restrictions]]
paths = "src/A.sol"
via_ir = true
```

`src/A.sol`:

```solidity
pragma solidity =0.8.35;
contract A { function value() external pure returns (uint256) { return 1; } }
```

`src/B.sol`:

```solidity
pragma solidity =0.8.35;
contract B { function value() external pure returns (uint256) { return 2; } }
```

`test/ATest.t.sol`:

```solidity
pragma solidity =0.8.35;
import {A} from "../src/A.sol";
contract ATest { function testA() public { require(new A().value() == 1); } }
```

`test/BTest.t.sol`:

```solidity
pragma solidity =0.8.35;
import {B} from "../src/B.sol";
contract BTest { function testB() public { require(new B().value() == 2); } }
```

`forge coverage --report lcov` — both tests pass, and the entire report is:

```
TN:
SF:src/B.sol
DA:2,2
FN:2,B.value
FNDA:2,B.value
FNF:1
FNH:1
LF:1
LH:1
BRF:0
BRH:0
end_of_record
```

Two defects in this tiny report:

1. **`src/A.sol` is missing entirely** — not zero-covered, absent.
2. **`FNDA:2,B.value` and `DA:2,2`** — `B.value` is called exactly once, by `BTest`. It is credited with two hits, because `A.value`'s execution is attributed to `B`. Coverage is not merely incomplete here; it is wrong for the file that survived.

### Controls

**Split the solc versions.** Change `src/B.sol` *and* `test/BTest.t.sol` to `pragma solidity =0.8.34`, so the two jobs no longer share a version:

Relevant LCOV entries:

```
SF:src/A.sol
DA:2,1
FNDA:1,A.value
SF:src/B.sol
DA:2,1
FNDA:1,B.value
```

Both files appear, each with the correct single hit. Same contract and test bodies; only the version overlap is gone.

**Rename the restricted file.** Renaming `A.sol` to `Z.sol` (updating the restriction and the import) flips which file survives — `src/Z.sol` is reported and `B.sol` disappears. In this repro the winner tracks path ordering; I have not established the general iteration order.

### Impact

In the real-world invocation, raw LCOV contained 130 `SF:` records, only 65 of them under `contracts/`. Whole modules were missing while their suites ran and passed in the same invocation — one contract with a 55-test suite, deployed with plain `new`, had no record at all.

The failure is silent in both directions: dropped files leave the denominator as well as the numerator, so the percentage stays plausible rather than collapsing. `forge coverage` emits no warning that sources were dropped or hits misattributed. We only noticed because a module we knew was well covered reported nothing.

That project is not public in the affected configuration, so the minimal repro above is the reproducible reference.

### Prior art

PR #9418 attempted to address the same identity problem by introducing `(build_id, source_id)`. It was merged into the feature branch `yash/fix-8840`, not master; its parent #9366 was later closed unmerged, and [review identified correctness problems in that implementation](https://github.com/foundry-rs/foundry/pull/9366#discussion_r1860512126). At `7c271517`, master still uses `(Version, source_id)`. `compilation_restrictions` arrived in #8668 without a corresponding coverage-identity change.

### Workaround

No reliable single-invocation config workaround when two non-empty compilation jobs must use the same exact solc version. Pinning each job family to a different exact solc version avoids the collision, since `Version` is part of the key — but that is impossible when contracts must share a version and differ only in `via_ir`. We run two coverage passes, each configured to produce exactly one job per solc version, and merge the LCOV outputs.

figtracer:
Solc assigns source IDs per compilation job, but coverage keyed source paths and analyses by solc version. Multiple jobs using the same version could therefore overwrite one another, dropping sources and crediting hits to the wrong file.

This keys report-level source identity, analyses, anchors, and hit resolution by compiler build ID instead. File reporters merge the same source item when a shared dependency appears in multiple builds, preserving one denominator while accumulating hits. Fixes #16084.

AI disclosure: This PR was written and tested with Codex.

### DIFF
--- a/.changelog/forge-coverage-build-ids.md
+++ b/.changelog/forge-coverage-build-ids.md
@@ -1,0 +1,6 @@
+---
+forge: patch
+foundry-evm-coverage: patch
+---
+
+Fixed coverage dropping or misattributing sources compiled in separate jobs with the same solc version.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5830,6 +5830,7 @@ dependencies = [
  "foundry-evm-core",
  "rayon",
  "revm",
+ "semver 1.0.28",
  "solar-compiler",
  "tracing",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5830,7 +5830,6 @@ dependencies = [
  "foundry-evm-core",
  "rayon",
  "revm",
- "semver 1.0.28",
  "solar-compiler",
  "tracing",
 ]
--- a/crates/evm/coverage/Cargo.toml
+++ b/crates/evm/coverage/Cargo.toml
@@ -21,7 +21,6 @@ foundry-evm-core.workspace = true
 alloy-primitives.workspace = true
 eyre.workspace = true
 revm.workspace = true
-semver.workspace = true
 tracing.workspace = true
 rayon.workspace = true
 solar.workspace = true
--- a/crates/evm/coverage/Cargo.toml
+++ b/crates/evm/coverage/Cargo.toml
@@ -21,6 +21,7 @@ foundry-evm-core.workspace = true
 alloy-primitives.workspace = true
 eyre.workspace = true
 revm.workspace = true
+semver.workspace = true
 tracing.workspace = true
 rayon.workspace = true
 solar.workspace = true
--- a/crates/evm/coverage/src/analysis.rs
+++ b/crates/evm/coverage/src/analysis.rs
@@ -474,9 +474,8 @@ impl SourceAnalysis {
     /// source ID the item is in, the source code range of the item, and the contract name the item
     /// is in.
     ///
-    /// Note: Source IDs are only unique per compilation job; that is, a code base compiled with
-    /// two different solc versions will produce overlapping source IDs if the compiler version is
-    /// not taken into account.
+    /// Note: Source IDs are only unique per compilation job, so report-level source identity must
+    /// also include the compiler build ID.
     #[instrument(name = "SourceAnalysis::new", skip_all)]
     pub fn new(data: &SourceFiles, output: &ProjectCompileOutput) -> eyre::Result<Self> {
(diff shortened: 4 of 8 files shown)


mattsse (approved):
lgtm

mattsse (approved):
lgtm

mablr (approved):

PR #16108, opened 2026-08-10
fix(ci): avoid restoring Cargo binaries

mablr:
The macOS test jobs restored a Rust cache that replaced the active Cargo and Rustup proxies, causing `cargo nextest` to invoke `rustup-init` and fail before running any tests.

This disables caching of `CARGO_HOME/bin` and installs Nextest after restoring the Rust cache. Cargo registries and build targets remain cached, while the test runner is always installed from a clean binary.

AI assistance was used to diagnose the CI failure and prepare this change.


grandizzy (approved):

mattsse (approved):
lgtm

PR #16111, opened 2026-08-10
fix(cast): confirm EIP-7702 auth disclosure

rappie:
## Motivation

Cast can send signed EIP-7702 authorizations to an RPC endpoint while simulating or preparing a transaction. This happens through calls such as `eth_call`, `debug_traceCall`, `eth_estimateGas`, and `eth_createAccessList`.

These authorizations remain valid outside the original request and can be submitted on-chain by anyone once their nonce is valid. A user invoking a simulation or transaction-building command may therefore unintentionally disclose a replayable authorization to an untrusted RPC endpoint.

## Solution

Warn and require confirmation before Cast sends a signed EIP-7702 authorization to an RPC endpoint. Add `--force` for users who intentionally want to skip the confirmation, and fail closed under `--quiet` rather than creating an invisible prompt.

The disclosure check is shared across Cast commands and uses transaction-builder state to distinguish RPC-backed simulation or estimation from local-only execution. This covers direct calls, transaction estimation and preparation, sending, batch commands, and the shared keychain and wallet-session transaction path without warning when the authorization remains local.

```text
Warning: This command will send a signed EIP-7702 authorization to the RPC endpoint. The authorization can be submitted on-chain by anyone once its nonce is valid.

Continue anyway? [y/N]
```

Codex was used to help understand the existing code and assist with the implementation. I fully understand and manually reviewed the resulting changes.

### DIFF
--- a/crates/cast/src/cmd/keychain.rs
+++ b/crates/cast/src/cmd/keychain.rs
@@ -57,7 +57,7 @@ use crate::cmd::tempo_policy_args::{
 
 use crate::{
     cmd::{auth::confirm_auth_rpc_disclosure_during_build, send::cast_send},
-    tx::{CastTxBuilder, CastTxSender, SendTxOpts},
+    tx::{CastTxBuilder, CastTxSender, SendTxOpts, SenderKind},
 };
 
 /// Tempo keychain management commands.
@@ -3506,6 +3506,13 @@ impl KeychainRootSigner {
             Self::Wallet(signer) => signer.address(),
         }
     }
+
+    fn sender(&self) -> SenderKind<'_> {
+        match self {
+            Self::Browser(browser) => browser.address().into(),
+            Self::Wallet(signer) => signer.as_ref().into(),
+        }
+    }
 }
 
 /// Resolve the root-authorized signer used for AccountKeychain policy changes.
@@ -3605,14 +3612,14 @@ pub(crate) async fn send_keychain_tx_with_root_signer(
         .with_code_sig_and_args(None, Some(hex::encode_prefixed(&calldata)), vec![])
         .await?;
 
-    if !confirm_auth_rpc_disclosure_during_build(&builder, root_signer.address(), force)? {
+    if !confirm_auth_rpc_disclosure_during_build(&builder, root_signer.sender(), force)? {
         return Ok(KeychainTxOutcome::Aborted);
     }
 
     if print_sponsor_hash {
         let from = root_signer.address();
         let chain = builder.chain();
-        let (mut tx, _) = builder.build(from).await?;
+        let (mut tx, _) = builder.build(root_signer.sender()).await?;
         if let Some(fee_payer) = sponsor_fee_payer {
             resolve_and_set_fee_token(
                 (!config.eth_rpc_curl).then_some(&provider),
@@ -3669,7 +3676,7 @@ pub(crate) async fn send_keychain_tx_with_root_signer(
         KeychainRootSigner::Wallet(signer) => {
             let from = signer.address();
             let chain = builder.chain();
-            let (mut tx, _) = builder.build(from).await?;
+            let (mut tx, _) = builder.build(signer.as_ref()).await?;
             maybe_print_resolved_lane(resolved_lane.as_ref(), tx.nonce().unwrap_or_default())?;
             if let Some(sponsor) = &tempo_sponsor {
                 sponsor
--- a/crates/cast/src/cmd/keychain.rs
+++ b/crates/cast/src/cmd/keychain.rs
@@ -56,7 +56,7 @@ use crate::cmd::tempo_policy_args::{
 };
 
 use crate::{
-    cmd::send::cast_send,
+    cmd::{auth::confirm_auth_rpc_disclosure_during_build, send::cast_send},
     tx::{CastTxBuilder, CastTxSender, SendTxOpts},
 };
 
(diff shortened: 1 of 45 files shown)


crates/cast/src/cmd/keychain.rs
```diff
@@ -3588,6 +3605,10 @@ pub(crate) async fn send_keychain_tx_with_root_signer(
         .with_code_sig_and_args(None, Some(hex::encode_prefixed(&calldata)), vec![])
         .await?;
 
+    if !confirm_auth_rpc_disclosure_during_build(&builder, root_signer.address(), force)? {
```
mablr: Passing only the address discards a resolved wallet signer, so `--auth <delegate-address> --private-key ...` fails validation with “No signer available.” Could we pass the signer through confirmation and `build` for `KeychainRootSigner::Wallet` (keeping browser address-only) and cover the address-form auth path?

crates/cast/src/opts.rs
```diff
@@ -1390,6 +1390,10 @@ pub enum CastSubcommand {
     /// Tempo keychain (access key) management.
     #[command(visible_alias = "kc")]
     Keychain {
+        /// Skip the EIP-7702 authorization disclosure confirmation.
+        #[arg(long, global = true)]
```
mablr: `global = true` exposes a no-op `--force` on read-only commands such as `keychain list`, `show`, and `doctor`.

crates/cast/src/cmd/wallet/session.rs
```diff
@@ -64,6 +64,10 @@ pub struct SessionArgs {
     #[command(subcommand)]
     pub command: Option<SessionSubcommands>,
 
+    /// Skip the EIP-7702 authorization disclosure confirmation.
+    #[arg(long, global = true)]
```
mablr: Making this global also exposes `--force` on `wallet session create`, where no authorization can be disclosed.

crates/cast/tests/cli/main.rs
```diff
@@ -3853,6 +4112,316 @@ casttest!(send_eip7702, async |_prj, cmd| {
...
+        .stderr_eq(str![[r#"
+Warning: This command will send a signed EIP-7702 authorization to the RPC endpoint. The authorization can be submitted on-chain by anyone once its nonce is valid.
+
+Continue anyway? [y/N] Aborted.
+
+"#]]);
+
+    cmd.cast_fuse()
+        .args(base_args)
+        .arg("--quiet")
```
mablr, line 4553: Could we also cover `--quiet --force` succeeding with the primary result on stdout and empty stderr? That is the intended non-interactive escape hatch, so it is worth locking down explicitly.

crates/cast/src/cmd/keychain.rs
stevencartavia: Reproduced with address-form `--auth` plus `--private-key`: it fails with “No signer available,” while the equivalent pre-signed authorization reaches the confirmation prompt.

crates/cast/src/cmd/mktx.rs
```diff
@@ -314,6 +336,9 @@ impl MakeTxArgs {
         if ethsign {
             // Use "eth_signTransaction" to sign the transaction only works if the node/RPC has
             // unlocked accounts.
+            if !confirm_auth_rpc_disclosure_during_build(&tx_builder, config.sender, force)? {
```
stevencartavia: The build-only predicate in the `--ethsign` path does not cover the subsequent `eth_signTransaction`. With a pre-signed `--auth` and explicit `--gas-limit`, no confirmation is shown; `provider.sign_transaction(tx)` then sends `authorizationList` to the RPC.

Reproduced for both `mktx` and `batch-mktx`: an RPC capture received `eth_signTransaction` with `authorizationList` even with `n\n` on stdin. Both `--ethsign` branches should use the unconditional disclosure confirmation when the builder has auth.

crates/cast/src/cmd/keychain.rs
rappie: Should be fixed now

rappie:
All comments should be fixed now

crates/cast/src/cmd/mktx.rs
```diff
@@ -203,6 +216,9 @@ impl MakeTxArgs {
                     None => eth.wallet.signer().await?,
                 };
                 let from = signer.address();
+                if !confirm_auth_rpc_disclosure_during_build(&tx_builder, &signer, force)? {
+                    return Ok(());
+                }
                 let (tx, _) = tx_builder.build(from).await?;
```
mablr: We validate with the wallet signer, but `build(from)` immediately discards it, so address-form `--auth` fails with `No signer available`. Please pass `&signer` into `build` and keep `from` only for the sponsor hash. The same issue exists in `send`'s sponsor-hash branch.

crates/cast/tests/cli/main.rs
```diff
@@ -1869,6 +1871,99 @@ casttest!(estimate_contract_deploy_gas, |_prj, cmd| {
...
+
+    let output = cmd
+        .cast_fuse()
+        .args(args)
+        .arg("--force")
+        .assert_success()
+        .stderr_eq(str![""])
+        .get_output()
+        .stdout_lossy();
+    assert!(output.trim().parse::<u64>().unwrap() > 21_000);
```
mablr, line 1964: Could we also cover `estimate --quiet --force`? Unlike the new `call` path, `print_scalar` still goes through quiet-suppressed shell output, so the command succeeds after disclosure but returns no primary result. Representative coverage here would prevent that automation footgun.

mablr (approved):
sgtm, ty

stevencartavia (approved):

PR #16112, opened 2026-08-10
feat(forge): rerun watched tests with `a`

### LINKED ISSUE #4734: Watch mode: Add key command to re-run tests
### Component

Forge

From time to time, I'd like to re-run tests in `--watch` mode without changing and saving a file. This is especially useful for fuzz tests, where I might want to see a few example failures in a row to understand why a particular test is failing. It would be nice to be able to press a hotkey to re-run all tests.

Jest's watch mode is a good example: pressing `a` at any time re-runs all tests:
40672937-25dab91a-6325-11e8-965d-4e55ef23e135

### Additional context

_No response_

figtracer:
`forge test --watch` previously only reacted to file changes, which made it impossible to rerun fuzz tests without touching the project. This enables Watchexec keyboard events while the watcher is idle and maps an unmodified `a` to a rerun without the automatic last-file filter, while preserving explicit CLI filters. Keyboard monitoring is suspended for the full child lifetime so prompts and the debugger retain deterministic ownership of terminal input, then resumes after the child exits.

Closes #4734.

AI disclosure: This PR was written and tested with Codex.

### DIFF
--- a/crates/forge/src/cmd/watch.rs
+++ b/crates/forge/src/cmd/watch.rs
@@ -12,7 +12,7 @@ use std::{
     io::IsTerminal,
     path::PathBuf,
     sync::{
-        Arc,
+        Arc, OnceLock, Weak,
         atomic::{AtomicU8, Ordering},
     },
     time::Duration,
@@ -33,6 +33,7 @@ use watchexec_signals::Signal;
 use yansi::{Color, Paint};
 
 type SpawnHook = Arc<dyn Fn(&[Event], &mut TokioCommand) + Send + Sync + 'static>;
+type KeyboardConfig = Arc<OnceLock<Weak<watchexec::Config>>>;
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 enum KeyboardAction {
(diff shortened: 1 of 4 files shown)


mablr (dismissed):

crates/forge/src/cmd/watch.rs
```diff
@@ -365,6 +401,12 @@ pub async fn watch_test(args: TestArgs) -> Result<()> {
             }
         },
     )?;
+
+    if std::io::stdin().is_terminal() {
+        config.keyboard_events(true);
```
stevencartavia: Enabling Watchexec’s stdin reader introduces a race with watched children that also read terminal input. In PTY tests, the base completed a watched `vm.prompt` test 3/3 times, while this PR completed only 1/3; the other runs hung. With `--watch --debug`, pressing `a` sometimes restarted the child instead of reaching the debugger.

Is this stdin ownership conflict intended? Could we avoid leaving interactive watched tests with nondeterministic input behavior?

stevencartavia (approved):

grandizzy (approved):
nice, lgtm!

PR #16113, opened 2026-08-10
feat(cli): require approval for local compiler executables

figtracer:
## Motivation

A project can point `solc` at an arbitrary executable in `foundry.toml`, for example `solc = "./payload"`. Foundry probes and runs that path as part of ordinary project commands, so building an untrusted repository can execute code before the user has approved it. This is particularly risky for non-interactive and agent-driven workflows.

## Solution

- require interactive confirmation before running project-configured local Solc or Vyper executables
- remember the decision per canonical executable path for the lifetime of the process
- fail closed in non-interactive environments
- add a global `--allow-local-compiler` flag for explicit CI and automation opt-in
- cover the refusal path with a regression test that verifies the configured executable is not invoked

The opt-in is CLI-only, so a project cannot grant itself permission through `foundry.toml`.


crates/config/src/lib.rs
```diff
@@ -152,6 +157,64 @@ pub use semver;
...
+}
+
+fn ensure_local_compiler_approved(path: &Path) -> Result<(), SolcError> {
+    use std::io::Write;
+
+    if ALLOW_LOCAL_COMPILERS.load(Ordering::Relaxed) {
+        return Ok(());
+    }
+
+    let path = dunce::canonicalize(path).unwrap_or_else(|_| path.to_path_buf());
```
mablr: The canonical path is only used as the cache key; callers still pass the original path to `Solc::new` / `Vyper::new`, and later compilation spawns that original path. A symlink can therefore be retargeted after approval, and bare Vyper names may resolve to a different PATH entry. Could this fail closed on resolution/canonicalization, return the canonical path, and use that exact path for every probe and spawn?

crates/config/src/lib.rs
```diff
@@ -152,6 +157,64 @@ pub use semver;
...
+    if !io::stdin().is_terminal() {
+        return Err(local_compiler_not_approved(&path));
+    }
+
+    let mut stderr = io::stderr().lock();
+    writeln!(
+        stderr,
+        "Warning: this project is configured to use a local compiler executable:\n  {}\n\
+         Running this executable may execute arbitrary code.",
+        path.display()
```
mablr: This renders a project-controlled filename directly in a security prompt. Unix paths can contain newlines and terminal escape sequences, allowing the displayed approval target to be spoofed. Please use a quoted/escaped representation here and in the refusal error, with a control-character regression test.

mablr (commented):
The feature itself feels well scoped, but I’m not sure foundry-config should be the final enforcement boundary. Config is the right place to identify project-supplied compiler paths, while the CLI should own prompting and `--allow-local-compiler`. Since foundry-core owns both version probes and compilation spawns, the final approval policy should ideally be enforced there against the exact resolved executable path.

crates/cli/src/opts/global.rs
```diff
@@ -3,8 +3,55 @@ use foundry_common::{
...
+
+static LOCAL_COMPILER_APPROVALS: Mutex<Vec<(PathBuf, bool)>> = Mutex::new(Vec::new());
+
+fn ensure_local_compiler_approved(path: &Path) -> CompilerResult<()> {
+    let mut approvals = LOCAL_COMPILER_APPROVALS.lock().unwrap_or_else(|err| err.into_inner());
+    if let Some((_, approved)) = approvals.iter().find(|(approved_path, _)| approved_path == path) {
+        return if *approved { Ok(()) } else { Err(local_compiler_not_approved(path)) };
+    }
+
+    if !io::stdin().is_terminal() {
```
mablr: Could we require stderr to be a terminal too? With TTY stdin but redirected stderr, this blocks on an invisible approval prompt instead of failing closed.

stevencartavia (dismissed):

mablr (dismissed):

mablr (dismissed):

crates/cli/src/utils/mod.rs
```diff
@@ -191,35 +192,89 @@ pub fn now() -> Duration {
 }
 
 /// Common setup for all CLI tools. Does not include [tracing subscriber](subscriber).
-pub fn common_setup() {
+pub fn common_setup() -> Result<()> {
     install_crypto_provider();
     crate::handler::install();
-    load_dotenv();
+    let allow_project_env =
+        std::env::args_os().take_while(|arg| arg != "--").any(|arg| arg == "--allow-project-env");
```
mablr: This raw argv scan can mistake positional data for approval. For example, `forge create ... --constructor-args --allow-project-env` consumes the token as a constructor argument, but this still loads `.env`. Please derive approval from an unambiguous parsed option instead.

mablr (commented):
Resolved conflicts, sgtm

mablr (dismissed):

mablr (approved):

mattsse (approved):

PR #16114, opened 2026-08-10
perf(forge): batch selector uploads

figtracer:
Batch all non-empty project ABIs into one deduplicated selector-registry request instead of paying one network round trip per contract. This preserves ABI-only compilation and per-contract progress output while removing request amplification and repeated inherited signatures. Codex assisted with investigation, implementation, benchmarking, and testing.

### Results

Profiling builds came from the same checkout. Hyperfine ran one warmup and three measured runs against the live registry on a 50-contract project with one distinct function per contract.

| Command | Before | After | Improvement |
|---|---:|---:|---:|
| `forge selectors upload --all` | `6.283s ± 0.075s` | `163.4ms ± 9.1ms` | `38.45x` |

A duplicate-heavy 3,000-contract stress project also completed successfully in `7.48s` total, including `4.84s` of ABI compilation, using one registry request instead of 3,000.

### DIFF
--- a/.changelog/forge-batch-selector-uploads.md
+++ b/.changelog/forge-batch-selector-uploads.md
@@ -1,3 +1,6 @@
-### `forge`
+---
+forge: patch
+---
 
-- Batch project selector uploads into one deduplicated registry request instead of sending one request per contract. ([#16114](https://github.com/foundry-rs/foundry/pull/16114))
+Batch project selector uploads into deduplicated, bounded registry requests instead of sending one
+request per contract.
--- a/.changelog/forge-batch-selector-uploads.md
+++ b/.changelog/forge-batch-selector-uploads.md
@@ -1,0 +1,3 @@
+### `forge`
+
+- Batch project selector uploads into one deduplicated registry request instead of sending one request per contract. ([#16114](https://github.com/foundry-rs/foundry/pull/16114))
--- a/crates/forge/src/cmd/selectors.rs
+++ b/crates/forge/src/cmd/selectors.rs
@@ -148,21 +148,18 @@ impl SelectorsSubcommands {
                     vec![(contract, artifact)]
                 };
 
-                let mut artifacts = artifacts.into_iter().peekable();
-                while let Some((contract, artifact)) = artifacts.next() {
+                let mut abis = Vec::with_capacity(artifacts.len());
+                for (contract, artifact) in artifacts {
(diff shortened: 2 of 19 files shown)


mablr (dismissed):

crates/forge/src/cmd/selectors.rs
```diff
@@ -148,21 +148,18 @@ impl SelectorsSubcommands {
...
-                    // upload abi to selector database
-                    import_selectors(SelectorImportData::Abi(vec![abi])).await?.describe();
-
-                    if artifacts.peek().is_some() {
-                        sh_println!()?
-                    }
+                    abis.push(abi);
+                }
+                if !abis.is_empty() {
+                    import_selectors(SelectorImportData::Abi(abis)).await?.describe();
```
mattsse, line 162: Could we keep the request body bounded here? [Sourcify's server](https://github.com/argotorg/sourcify/blob/ab64e6d3cd9e8368f155f5023a8a9ba72960b999/services/4byte/src/FourByteServer.ts#L44-L45) parses this endpoint with `express.json()` at its 100 KB default. For example, 3,000 distinct signatures shaped like `functionName0000(address,uint256)` serialize to 108,025 bytes, so `--all` now gets a 413 where the previous per-contract requests succeeded. Deduplication only protects duplicate-heavy projects; please chunk the deduplicated signatures and add coverage above the limit rather than placing every ABI in one request.

.changelog/forge-batch-selector-uploads.md
```diff
@@ -0,0 +1,3 @@
+### `forge`
+
+- Batch project selector uploads into one deduplicated registry request instead of sending one request per contract. ([#16114](https://github.com/foundry-rs/foundry/pull/16114))
```
mattsse: The current validator rejects this fragment because it requires YAML frontmatter with a package bump; CI is failing with `frontmatter must start with '---'`. Please replace the legacy heading/list with:

```md
---
forge: patch
---

Batch project selector uploads into one deduplicated registry request instead of sending one request per contract.
```

mattsse (changes requested):
Two blocking issues need addressing before this can merge.

mattsse (approved):
lgtm

stevencartavia (approved):

DaniPopes (approved):

PR #16116, opened 2026-08-10
fix(doc): preserve multiline notice metadata

stevencartavia:
Preserves complete multiline NatSpec notices in generated page metadata. Forge now derives the frontmatter description from the joined notice used by the rendered body instead of dropping continuation lines.


mablr (approved):

mattsse (approved):
lgtm

PR #16117, opened 2026-08-10
fix(doc): preserve notices with multiline params

stevencartavia:
Preserves inherited NatSpec notices when an override provides a multiline parameter description. Forge now recognizes synthetic notice entries as parameter continuations instead of treating them as local notices that suppress inheritance.


mattsse (approved):
lgtm

mablr (approved):

PR #16126, opened 2026-08-11
fix(anvil): execute JSON-RPC notifications

stevencartavia:
Anvil previously ignored JSON-RPC notifications entirely. This change executes notifications through the normal RPC handler while continuing to suppress their responses, as required by JSON-RPC 2.0. It applies consistently across HTTP, WebSocket, and IPC transports.


mattsse (approved):
lgtm

grandizzy (approved):
lgtm

PR #16129, opened 2026-08-11
perf(forge): speed up clone compilation

figtracer:
## Summary

- compile cloned projects with storage-layout-only output selection
- keep compiler output in memory because `forge clone` only reads the target artifact's storage layout
- avoid bytecode generation and writing unused build artifacts while preserving the cloned project's compiler settings
- cover storage layout retention, omitted bytecode, and the absence of artifact writes in the existing clone regression test

## Benchmarks

Aave v4 compilation phase, cold `out` and cache, 2 measured runs:

| Compilation | Before | After | Speedup |
| --- | ---: | ---: | ---: |
| Full cloned project | 196.379 s ± 1.662 s | 4.648 s ± 0.103 s | 42.25x |

Artifact output drops from 272 MB to none.

Measured with `hyperfine --runs 2 --warmup 0`, removing the benchmark `out` and cache directories before each sample. The baseline uses the equivalent full-project `forge build --offline --no-cache --no-lint`; the candidate uses a temporary driver around `clone::compile_project` so explorer and dependency download latency is excluded from both sides.

### DIFF
--- a/crates/cli/src/utils/mod.rs
+++ b/crates/cli/src/utils/mod.rs
@@ -810,6 +810,7 @@ ignore them in the `.gitignore` file."
             .exec()?;
         let (_, mappings) = self.submodule_mappings_at(worktree_root)?;
         let mut gitlinks = BTreeMap::new();
+        let mut status_paths = Vec::new();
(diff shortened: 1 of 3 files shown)


mablr (approved):

mattsse (approved):
lgtm

PR #16130, opened 2026-08-11
feat(forge): cover empty special functions

### LINKED ISSUE #9458: coverage: identify hits for empty fallbacks
### Component

Forge

Currently we can correctly identify hits for empty functions, but not constructors, fallbacks, or receives.

### Additional context

- https://github.com/foundry-rs/foundry/issues/9270
- https://github.com/foundry-rs/foundry/issues/9444

figtracer:
Empty constructors, receive functions, and fallbacks had no reliable source-mapped opcode, so coverage either omitted them or could not tell receive and fallback apart. This records call routing alongside instruction hits, then uses artifact ABI selectors and Solar's resolved inheritance data to attribute those calls without conflating receive, fallback, or named functions.

Closes #9458.

AI assistance was used for implementation and verification.

### DIFF
--- a/crates/evm/coverage/src/analysis.rs
+++ b/crates/evm/coverage/src/analysis.rs
@@ -45,6 +45,7 @@ struct SourceVisitorCheckpoint {
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub(crate) enum EmptySpecialFunctionKind {
+    Constructor,
     Receive,
     Fallback,
 }
@@ -74,11 +75,19 @@ fn resolve_empty_special_functions(
     for contract_id in gcx.hir.contract_ids() {
         let contract = gcx.hir.contract(contract_id);
         let Some(&contract_source_id) = hir_source_ids.get(&contract.source) else { continue };
-        for (function_id, kind) in [
+        let constructors = contract.linearized_bases.iter().filter_map(|&base_id| {
+            gcx.hir
+                .contract(base_id)
+                .ctor
+                .map(|function_id| (function_id, EmptySpecialFunctionKind::Constructor))
+        });
+        let runtime_functions = [
             (contract.receive, EmptySpecialFunctionKind::Receive),
             (contract.fallback, EmptySpecialFunctionKind::Fallback),
-        ] {
-            let Some(function_id) = function_id else { continue };
+        ]
+        .into_iter()
+        .filter_map(|(function_id, kind)| function_id.map(|function_id| (function_id, kind)));
+        for (function_id, kind) in constructors.chain(runtime_functions) {
             let function = gcx.hir.function(function_id);
             if !function.body.is_some_and(|body| body.is_empty()) {
                 continue;
--- a/crates/evm/coverage/src/analysis.rs
+++ b/crates/evm/coverage/src/analysis.rs
@@ -43,6 +43,63 @@ struct SourceVisitorCheckpoint {
     function_calls: usize,
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum EmptySpecialFunctionKind {
+    Receive,
+    Fallback,
+}
+
+struct ResolvedEmptySpecialFunction {
+    contract_source_id: u32,
+    contract_name: Arc<str>,
+    function_source_id: u32,
+    function_contract_name: Arc<str>,
+    function_bytes: Range<u32>,
+    kind: EmptySpecialFunctionKind,
+}
+
+fn resolve_empty_special_functions(
+    gcx: Gcx<'_>,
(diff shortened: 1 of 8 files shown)


crates/evm/coverage/src/lib.rs
```diff
@@ -353,6 +438,50 @@ impl fmt::Display for ItemAnchor {
...
+    has_receive: bool,
+}
+
+impl ContractCallAnchors {
+    fn hits(&self, hit_map: &HitMap, kind: CallAnchorKind) -> u32 {
+        match kind {
+            CallAnchorKind::Receive => hit_map.empty_calls,
+            CallAnchorKind::Fallback => {
+                let empty_calls = if self.has_receive { 0 } else { hit_map.empty_calls };
+                empty_calls
```
mattsse, line 515: `ContractCallAnchors::hits` classifies routed calls from calldata alone, so it also credits calls that the compiler rejects before fallback entry. For example, with `fallback() external {}`, `address(a).call{value: 1}(hex"deadbeef")` reverts at the generated nonpayable guard, yet LCOV reports the fallback as hit; the same call against a non-empty fallback reports zero. Please account for fallback payability and call value, or otherwise confirm entry before incrementing this anchor. The current tests only exercise zero-value fallback calls.

crates/evm/coverage/src/analysis.rs
```diff
@@ -209,9 +266,9 @@ impl<'ast> ast::Visit<'ast> for SourceVisitor<'_> {
     fn visit_item(&mut self, item: &'ast ast::Item<'ast>) -> ControlFlow<Self::BreakValue> {
         match &item.kind {
             ItemKind::Function(func) => {
-                // TODO: We currently can only detect empty bodies in normal functions, not any of
-                // the other kinds: https://github.com/foundry-rs/foundry/issues/9458
-                if func.kind != ast::FunctionKind::Function && !has_statements(func.body.as_ref()) {
+                // Empty modifiers have no bytecode of their own to anchor. Empty constructors,
+                // receive functions, and fallbacks are handled separately.
+                if func.kind == ast::FunctionKind::Modifier && !has_statements(func.body.as_ref()) {
```
mattsse, line 280: Unconditionally including an empty constructor still depends on ordinary opcode anchoring. With `forge coverage --ir-minimum`, `contract A { constructor() {} }` and a test that executes `new A()` produce no matching constructor anchor, and LCOV reports `FNDA:0`/`DA:0` because the via-IR creation source map only contains whole-contract spans. Please add constructor-specific attribution or avoid reporting the item when no reliable anchor exists. The new constructor test only covers non-IR compilation.

mattsse (changes requested):
Please address the two inline coverage-attribution issues.

mattsse (dismissed):
lgtm

mablr (dismissed):

crates/evm/coverage/src/inspector.rs
```diff
@@ -34,7 +34,14 @@ impl Default for LineCoverageCollector {
 
 impl<CTX> Inspector<CTX> for LineCoverageCollector {
     fn initialize_interp(&mut self, interpreter: &mut Interpreter, _context: &mut CTX) {
+        let call = interpreter.input.bytecode_address.is_some().then(|| {
+            let input = interpreter.input.input.as_bytes_memory(&interpreter.memory);
```
stevencartavia: A successful nested call with empty calldata crashes `forge coverage` at this read. Reproduced with `(bool success,) = target.call{gas: 100_000}("")` to `receive() external payable {}`. The test passes normally, but coverage exits 101 with `slice OOB: 18446744073709551615..18446744073709551615; len: 0`. This reproduces with normal coverage and `--ir-minimum`. Could this empty `CallInput::SharedBuffer` be handled without slicing the invalid shared-memory range?

crates/evm/coverage/src/inspector.rs
```diff
@@ -34,7 +34,14 @@ impl Default for LineCoverageCollector {
...
+            (CallData::new(&input), !interpreter.input.call_value.is_zero())
+        });
         let map = self.get_or_insert_map(interpreter);
+        if let Some((call, with_value)) = call {
+            map.call(call, with_value);
```
stevencartavia, line 48: Non-blocking: recording the calldata route during `initialize_interp` treats child-frame initialization as entry into an empty `receive`. Reproduced with gas 0 and gas 1 calls to empty and non-empty receives: all four calls return `false` with `OutOfGas`, but LCOV reports `FNDA:2` for the empty receive and `FNDA:0` for the non-empty receive, whose body statement also remains unhit.

mablr (approved):

stevencartavia (approved):

PR #16131, opened 2026-08-11
fix(evm): preserve exact fork ancestry

### LINKED ISSUE #15897: bug(fork-db): exact forks mix canonical ancestry

mablr:
## Motivation

Close #15897.

Foundry can mix an initially resolved fork block with state or `BLOCKHASH` ancestry from its canonical replacement after a same-height reorg. foundry-rs/foundry-core#157 adds the hash-anchored database support; this PR completes the fix by carrying that exact block and RPC-source identity through Foundry’s fork consumers.

Fork creation, rolling, transaction replay, Anvil resets, cached Chisel sessions, and fork reuse now remain tied to the resolved ancestry instead of identifying forks by block number alone. The regression reproduces a same-height reorg and verifies that both state and ancestry remain consistent with the original fork.

The hash-anchored database support from foundry-rs/foundry-core#157 is now part of Foundry's dependency baseline; this PR adds the Foundry-side wiring and regressions on top of that bump.


figtracer (dismissed):

crates/evm/core/src/fork/multi.rs
```diff
@@ -609,9 +681,76 @@ async fn create_fork<
 
     let provider = fork.evm_opts.fork_provider_with_url::<N>(&fork.url)?;
     let db = BlockchainDb::new(meta, cache_path);
-    let (backend, handler) = SharedBackend::new(provider, db, Some(number.into()));
-    let fork_id = ForkId::new(&fork.url, Some(number));
+    let anchor = ForkBlock::with_rpc_number(
```
stevencartavia, line 767: This makes `blockhash(block.number - 1)` fail in Forge tests on Arbitrum forks. Foundry replaces the EVM-visible block number with `l1BlockNumber`, while the exact backend rejects uncached lookups below its anchor when that number differs from the RPC block number.

Reproduced against Arbitrum block 493511363 (`l1BlockNumber` 25733601):

```text
failed to get block hash for 25733600: exact block ancestry is unsupported when the EVM-visible block number (25733601) differs from the RPC block number (493511363)
```

Could remapped-number forks retain hash-pinned state reads while using a working, non-anchored `BLOCKHASH` lookup path?

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -3361,48 +3361,18 @@ impl<N: Network> Backend<N> {
                     )));
                 }
             }
+            let fork_url = target_rpc_url
+                .clone()
+                .ok_or_else(|| BlockchainError::InvalidUrl("fork URL is missing".to_string()))?;
+            let fork = self
+                .reset_block_number(fork_url, fork_block_number, forking.json_rpc_url.is_some())
+                .await?;
             let fork_block = fork
```
stevencartavia: `reset_block_number` publishes the new database, `node_config`, `ClientFork`, and EVM environment before `reset_fork` performs the fallible `block_by_hash` lookup.

Reproduced with `eth_getBlockByHash` returning `null`: `anvil_reset` returned `Resource not found`, but the node then reported chain ID 11155111 while `latest` was still the old chain’s head. Could the hash-selected block be fetched and validated before publishing those replacements, so this lookup failure leaves the current fork state unchanged?

stevencartavia (dismissed):

crates/evm/core/src/fork/multi.rs
```diff
@@ -70,6 +71,13 @@ impl ForkId {
         Self(id)
     }
 
+    /// Returns the identifier for an exactly resolved fork.
+    fn resolved(url: &str, fork: &ResolvedFork) -> Self {
+        let mut id = Self::new_with_context(url, Some(fork.number()), Some(&fork.context())).0;
+        write!(id, "#{}:{}", fork.hash(), fork.source_id()).unwrap();
```
stevencartavia, line 77: This regresses `ArbSys.arbBlockNumber()` on Arbitrum forks. Reproduced at the current head: it returns the remapped L1 block (`16939475`) instead of the L2 fork block (`75219831`).

`ForkId::resolved()` appends `#<hash>:<source-id>` after the block number, but `fork_block_number()` parses the entire suffix as hex. It returns `None`, causing a fallback to the L1-valued `env_block`. Could we handle the resolved suffix and add a resolved-ID parser test?

crates/forge/src/mutation/runner.rs
```diff
@@ -714,9 +714,9 @@ fn compile_and_test_inner<FEN: FoundryEvmNetwork>(
 
     // Use block_on to run within the runtime context
     let results: BTreeMap<String, SuiteResult> = rt.block_on(async {
-        let (evm_env, tx_env, fork_context) = evm_opts
-            .env_with_fork_context::<SpecFor<FEN>, BlockEnvFor<FEN>, TxEnvFor<FEN>>()
-            .await?;
+        let (evm_env, tx_env, fork) =
+            evm_opts.env_resolved::<SpecFor<FEN>, BlockEnvFor<FEN>, TxEnvFor<FEN>>().await?;
```
figtracer: `run_mutation_testing` still checks `can_use_create2_deployer` by block number before each worker resolves this exact fork. a same-height reorg between those calls can derive linker behavior from block a while this runner executes block b. please resolve once in the orchestrator, use `can_use_create2_deployer_resolved`, and pass that identity into each worker.

Cargo.toml
```diff
@@ -515,30 +515,30 @@ flate2 = "1.1"
 ethereum_ssz = "0.10"
 
 # Tempo
-alloy-transport-mpp = { git = "https://github.com/tempoxyz/mpp-rs", rev = "63171c2", default-features = false, features = [
+alloy-transport-mpp = { git = "https://github.com/tempoxyz/mpp-rs", rev = "85c62c59205789d6506ad0c77f543a8680f02612", default-features = false, features = [
```
grandizzy, line 518: This merge resolution unexpectedly upgrades Tempo to `4d248169…` and MPP to `85c62c…`, although neither merge parent contained those revisions. It also changes Tempo hardfork behavior from T9 to T10 and overlaps the closed dependency PR #16301.

The exact-fork `foundry-fork-db` change itself has no Tempo or MPP dependency. Could we pin a compatible foundry-core revision containing foundry-core#157 and restore the base Tempo/MPP revisions, leaving those upgrades to their dedicated PR?

grandizzy (commented):
The exact-ancestry plumbing looks directionally sound, and CI is green. I found one unrelated dependency bump and one missing ancestry validation that should be addressed before merging.

crates/evm/core/src/backend/mod.rs
grandizzy, line 1: At `block_context_inputs_from_backend` (around lines 1134–1155), `SharedBackend::get_full_block` does not verify that a block-by-hash response matches the requested hash or expected height. Consequently, these parent and grandparent responses can introduce unrelated transaction history into Monad’s supposedly exact `BlockContext`.

Could we validate each returned header against the requested hash and expected number, as the transaction block is validated above at lines 1073–1080?

Cargo.toml
mablr: Yes, I pinned to HEAD, I'll fix that

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5043,15 +5047,14 @@ where
...
         if evm_env.block_env.basefee == 0 {
             evm_env.cfg_env.disable_base_fee = true;
         }
 
         let best_number = self.blockchain.storage.read().best_number;
         let block_number = best_number.saturating_add(1);
-        if is_arbitrum(source_chain_id) {
-            evm_env.block_env.number = U256::from(block_number);
+        if let Some(block_number) = arbitrum_block_number {
+            evm_env.block_env.number = block_number;
```
stevencartavia: In a local chain-ID-42161 repro at RPC block `102` with proxy-injected `l1BlockNumber=10`, the source trace and pre-PR replay return ArbSys `102`, while this PR returns `101`; immediately afterward, `eth_blockNumber` is `102` but an ArbSys call returns `101`. ArbSys is taking `max(parent RPC block 101, EVM block 10)`, so please retain the target RPC block separately while using `l1BlockNumber` as EVM-visible `block.number`. This used a decodable local prefix; the recent public-mainnet blocks tested failed before replay on unsupported transaction type `0x6a`.

crates/evm/core/src/fork/multi.rs
0xsarawut-eth: > This makes `blockhash(block.number - 1)` fail in Forge tests on Arbitrum forks. Foundry replaces the EVM-visible block number with `l1BlockNumber`, while the exact backend rejects uncached lookups below its anchor when that number differs from the RPC block number.
> 
> Reproduced against Arbitrum block 493511363 (`l1BlockNumber` 25733601):
> 
> ```text
> failed to get block hash for 25733600: exact block ancestry is unsupported when the EVM-visible block number (25733601) differs from the RPC block number (493511363)
> ```
> 
> Could remapped-number forks retain hash-pinned state reads while using a working, non-anchored `BLOCKHASH` lookup path?

figtracer (dismissed):

stevencartavia (dismissed):

figtracer (approved):

stevencartavia (approved):

PR #16133, opened 2026-08-11
test(anvil): cover Tempo signed tx identity

mablr:
Follow-up to #15857, which preserved complete Tempo transaction requests through Anvil’s node-signing path. The existing regression covers Tempo’s special call-batch handling with a minimally specified `eth_sendTransaction` request; this adds coverage for the signed transaction identity produced by a fully specified sponsored request on a nonzero nonce lane.

The new test decodes the returned Tempo transaction and jointly verifies its calls, lane and nonce, node-generated sender signature and recovery, sponsor signature and recovery, and both the returned and independently recomputed transaction hash. This protects fields that collectively determine the signed payload even when transaction execution itself still succeeds.


mattsse (approved):
lgtm

stevencartavia (approved):

PR #16134, opened 2026-08-11
test(anvil): strengthen replay coverage

mablr:
This follows up on PR #15899, which made local block replay apply protocol pre-execution changes and propagate preparation failures. Its EIP-2935 regression only converted any nonzero history result into a sentinel storage write, so it could not distinguish the expected parent block hash from another nonzero value. The probe now stores the returned word and asserts the exact zero-to-parent-hash transition.

The deterministic failing-precompile case now also exercises `trace_transactionOpcodeGas`. Unlike `trace_replayTransaction`, that RPC uses the separate `replay_tx_with_inspector` path, so this verifies that preparation errors are propagated consistently through that caller as well.

### DIFF
--- a/crates/anvil/tests/it/eip2935.rs
+++ b/crates/anvil/tests/it/eip2935.rs
@@ -12,8 +12,8 @@ use alloy_rpc_types::{
     TransactionRequest,
     simulate::{SimBlock, SimulatePayload},
     trace::{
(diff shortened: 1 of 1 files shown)


mattsse (approved):
lgtm

stevencartavia (approved):

PR #16136, opened 2026-08-11
fix(mutation): filter fixed-bytes unary ops

mablr:
The unary mutator generated negation and increment or decrement replacements for fixed-bytes operands, even though fixed bytes support bitwise complement but not those replacement operators. These candidates always failed compilation.

This classifies resolved elementary operand types and records replacement-specific exclusions for fixed bytes. It preserves bitwise complement, valid unsigned increment and decrement mutations, storage `push()` behavior, and the existing fail-open handling for unresolved or overloaded user-defined value types.

Implementation was assisted by Codex

### DIFF
--- a/.changelog/mutation-fixed-bytes-unary.md
+++ b/.changelog/mutation-fixed-bytes-unary.md
@@ -1,0 +1,5 @@
+---
+forge: patch
+---
+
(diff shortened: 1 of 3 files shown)


figtracer (approved):

stevencartavia (approved):

mattsse (approved):
lgtm

PR #16137, opened 2026-08-11
fix(mutation): filter compound negation

mablr:
Assignment mutation also applies to compound assignments, but semantic filtering previously covered only plain assignments. This allowed mutation testing to negate unsigned or fixed-bytes right-hand operands and shift amounts, producing mutants that cannot compile.

This extends the HIR analysis to compound assignments and excludes only type-invalid negations based on the operator and resolved operand types. Signed non-shift operands and valid numeric-literal replacements remain available, while unresolved types and user-defined value types continue to fail open.

Implementation was assisted by Codex

### DIFF
--- a/crates/forge/tests/cli/test_cmd/mutation.rs
+++ b/crates/forge/tests/cli/test_cmd/mutation.rs
@@ -1947,16 +1947,16 @@ MUTATION TESTING RESULTS
 ╭──────────┬───────────┬────────────╮
 │ Status   ┆ # Mutants ┆ % of Total │
 ╞══════════╪═══════════╪════════════╡
-│ Survived ┆ 3         ┆ 5.6%       │
+│ Survived ┆ 3         ┆ 5.7%       │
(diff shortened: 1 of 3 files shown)


crates/forge/tests/cli/test_cmd/mutation.rs
```diff
@@ -235,6 +235,95 @@ exclude_operators = [
...
+"#,
+    )
+    .unwrap();
+
+    let output = cmd
+        .args(["test", "--mutate", "src/CompoundAssignments.sol", "--mutation-jobs", "1", "--json"])
+        .assert_success();
+    let summary = mutation_summary(&output.get_output().stdout_lossy());
+
+    assert_eq!(summary["total"], 6);
```
figtracer, line 320: please update `mutation_testing_require_mutator`'s snapshot too; this filter changes it from 52 to 51 mutants, and `test / test all` currently fails.

figtracer (approved):

stevencartavia (approved):

PR #16138, opened 2026-08-11
fix(anvil): close IPC connections on shutdown

stevencartavia:
Dropping `NodeHandle` stopped the IPC listener, but connections that had already been accepted were running as detached tasks and continued serving RPC requests. This allowed existing IPC clients to retain access to the node backend after Anvil was expected to have shut down.

This change tracks accepted connections under the listener task so shutting down the node also closes its active IPC connections. Completed connection tasks are cleaned up during normal operation, and the regression test accounts for Alloy’s automatic IPC reconnection behavior.

### DIFF
--- a/.changelog/close-ipc-connections-on-shutdown.md
+++ b/.changelog/close-ipc-connections-on-shutdown.md
@@ -1,0 +1,5 @@
+---
+anvil: patch
+---
+
+Close established IPC connections when their Anvil node shuts down.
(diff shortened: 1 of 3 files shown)


mablr (approved):

Elenore23:
Ok

Elenore23 (approved):

figtracer (approved):

PR #16141, opened 2026-08-12
fix(config): preserve custom source defaults

### LINKED ISSUE #10300: leaky interface: _with_root

stevencartavia:
Preserve customized compiled `src` defaults instead of replacing them during `src`/`contracts` directory auto-detection. Canonical Foundry defaults continue to support Hardhat directory detection.

Closes #10300.

### DIFF
--- a/.changelog/config-custom-src-default.md
+++ b/.changelog/config-custom-src-default.md
@@ -1,0 +1,6 @@
+---
+forge: patch
+foundry-config: patch
+---
+
+Preserved customized default source directories when loading Forge projects.
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -774,6 +774,8 @@ impl Config {
     /// File name of config toml file
     pub const FILE_NAME: &'static str = "foundry.toml";
 
+    const DEFAULT_SRC: &'static str = "src";
+
     /// The name of the directory foundry reserves for itself under the user's home directory: `~`
     pub const FOUNDRY_DIR_NAME: &'static str = ".foundry";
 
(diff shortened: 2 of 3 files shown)


grandizzy (approved):
lgtm

mablr (approved):

PR #16148, opened 2026-08-12
fix(networks): make Monad-only lint clean

### LINKED ISSUE #16146: fix(networks): make Monad-only clippy clean
### Component

CI / EVM networks

### Affected version

The Monad network tests proposed in #15343 at `40f293806e535762cabe33c70aa465c8e7b47918`.

### Description

The isolated Monad feature configuration fails strict Clippy because `fork_sources_only_isolate_monad` declares a mutable vector whose only mutation is compiled behind the `optimism` feature.

This is hidden by the workspace all-features Clippy job and by Monad jobs that retain the default features.

### Reproduction

```console
RUSTFLAGS=-Dwarnings cargo +nightly clippy \
  -p foundry-evm-networks \
  --no-default-features \
  --features monad \
  --tests \
  --locked
```

Clippy reports that `non_monad` does not need to be mutable.

### Expected behavior

Every supported feature combination should compile and lint cleanly with warnings denied.

### Suggested fix

Construct the test inputs in a cfg-safe way so mutability exists only when the Optimism entry is present. Avoid a lint suppression.

Consider adding a small no-default-features Monad check for `foundry-evm-networks` so this isolated feature configuration remains covered without duplicating the full workspace matrix.

haythemsellami:
Fixes #16146.

The Monad-only feature combination compiled a test fixture with a mutable vector whose only mutation was gated by Optimism, so strict Clippy failed when default features were disabled. This builds the fixture directly from cfg-gated array elements and adds the isolated feature combination to CI so supported profiles remain warning-free.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -235,6 +235,10 @@ jobs:
       - run: cargo clippy --workspace --all-targets --all-features --locked
         env:
           RUSTFLAGS: -Dwarnings
+      - name: Lint Monad-only network support
+        run: cargo clippy -p foundry-evm-networks --no-default-features --features monad --tests --locked
+        env:
+          RUSTFLAGS: -Dwarnings
 
   rustfmt:
     runs-on: depot-ubuntu-latest
--- a/.changelog/monad-only-network-clippy.md
+++ b/.changelog/monad-only-network-clippy.md
@@ -1,0 +1,5 @@
+---
+foundry-evm-networks: patch
+---
+
+Fixed linting for Monad-only builds of Foundry's EVM network package.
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -235,6 +235,10 @@ jobs:
       - run: cargo clippy --workspace --all-targets --all-features --locked
         env:
           RUSTFLAGS: -Dwarnings
+      - name: Lint Monad-only network support
+        run: cargo clippy -p foundry-evm-networks --no-default-features --features monad --tests --locked
+        env:
+          RUSTFLAGS: -Dwarnings
```
mablr, line 241: This is not already covered above ?

haythemsellami: `--all-features` enables Optimism. the lint described in the issue was caught only by running Monad and Optimism disabled.

haythemsellami: Definitely not a high priority...

mablr: Hmm yes I think we def can keep as is, let's not bloat CI.

PR #16149, opened 2026-08-12
fix(cast): reject unavailable DA networks

### LINKED ISSUE #16145: bug(cast): da-estimate misclassifies disabled networks
### Component

Cast

### Affected version

The generic network inference changes in #15343 at `40f293806e535762cabe33c70aa465c8e7b47918`.

### Description

`cast da-estimate` uses the infallible `From<ChainId>` compatibility conversion when it infers a network from an RPC endpoint. That conversion intentionally defaults to Ethereum when classification fails.

In a Cast build with Optimism support but without the `monad` feature, an endpoint reporting known Monad chain ID 143 therefore resolves to Ethereum. Cast proceeds with Ethereum DA estimation instead of reporting that the known network family is unavailable, or reporting that DA estimation is unsupported for Monad.

Other user-facing network inference paths introduced by #15343 use the fallible known-chain classifier specifically to prevent a feature-disabled known family from silently executing with Ethereum semantics.

### Reproduction

1. Build Cast with the `optimism` feature and without the `monad` feature.
2. Point an RPC endpoint at a chain reporting chain ID 143.
3. Run `cast da-estimate latest --rpc-url <endpoint>` without an explicit network.

The command enters the Ethereum DA-estimation path.

### Expected behavior

A known but feature-disabled network should produce an actionable feature-unavailable error. Only genuinely unknown chain IDs should use the existing Ethereum fallback. An explicit network selection should remain authoritative.

### Suggested fix

Resolve the RPC chain ID through `NetworkVariant::from_known_chain_id`, propagate its unavailable-family error, and default to Ethereum only when the classifier returns `Ok(None)`.

Add tests for Monad mainnet and testnet IDs with and without the feature, an unknown chain ID, and explicit-network precedence.

haythemsellami:
Fixes #16145.

`cast da-estimate` used the compatibility conversion that maps failed chain-family classification to Ethereum. It now uses the fallible known-chain resolver, reports feature-disabled network families directly, and retains the Ethereum fallback only for genuinely unknown chain IDs. Explicit network selection continues to take precedence without probing the RPC endpoint.

### DIFF
--- a/crates/cast/src/cmd/da_estimate.rs
+++ b/crates/cast/src/cmd/da_estimate.rs
@@ -49,10 +49,7 @@ async fn resolve_da_network(
     }
 
     let provider = ProviderBuilder::<AnyNetwork>::from_config(config)?.build()?;
-    infer_da_network_from_chain_id(provider.get_chain_id().await?)
-}
-
-fn infer_da_network_from_chain_id(chain_id: u64) -> Result<NetworkVariant> {
+    let chain_id = provider.get_chain_id().await?;
     NetworkVariant::from_known_chain_id(chain_id)
         .map(|network| network.unwrap_or(NetworkVariant::Ethereum))
         .map_err(|error| {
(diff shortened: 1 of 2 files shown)


crates/cast/src/cmd/da_estimate.rs
```diff
@@ -46,6 +40,26 @@ impl DAEstimateArgs {
...
+    infer_da_network_from_chain_id(provider.get_chain_id().await?)
+}
+
+fn infer_da_network_from_chain_id(chain_id: u64) -> Result<NetworkVariant> {
+    NetworkVariant::from_known_chain_id(chain_id)
+        .map(|network| network.unwrap_or(NetworkVariant::Ethereum))
+        .map_err(|error| {
+            eyre::eyre!("cannot infer execution network from chain ID {chain_id}: {error}")
+        })
+}
```
mablr: Can we simply inline this function in `resolve_da_network` ?

mablr (commented):
I actually think whole PR may be a footgun. Previously it was always failling back to Eth, now it triggers an error.

This is not a big deal, the only corner case i see is calling monad rpc with bin compiled without `monad` feat, actually the error message would be misleading, so we can keep this as-is imo.

PR #16153, opened 2026-08-12
fix(monad): refresh tracker after state splices

### LINKED ISSUE #16142: bug(monad): state-splice cheatcodes leave reserve tracking stale
### Component

Forge / Cheatcodes

### Affected version

The Monad integration proposed in #15343 at `40f293806e535762cabe33c70aa465c8e7b47918`.

### Description

`vm.deal`, `vm.cloneAccount`, and `vm.loadAllocs` can leave Monad's live reserve-balance tracker inconsistent with the journaled account state.

Monad reserve tracking keeps derived violation state outside REVM's raw `JournaledState`. Normal balance and code mutations notify the Monad journal hooks, while transaction state replacement explicitly calls `refresh_context_after_state_change`. These three cheatcodes instead write balances or code directly into `JournaledState` and do not refresh the auxiliary Monad context.

The manual rollback of recorded `vm.deal` operations has the same problem because it restores `AccountInfo::balance` directly.

### Reproduction

1. Run a Forge test with `network = "monad"` and `hardfork = "monad:MonadNine"`.
2. Establish a 12 MON reserve for a restricted EOA and transfer 3 MON from it so the reserve precompile reports a violation.
3. Call `vm.deal(eoa, 12 ether)` to restore its balance.
4. Query `dippedIntoReserve()` at `0x0000000000000000000000000000000000001001`.

The balance is restored, but the precompile still returns `true`. The equivalent stale state can be produced by replacing a tracked account through `cloneAccount` or `loadAllocs`.

### Expected behavior

Reserve tracking should reflect the account state immediately after each state-mutating cheatcode and after a recorded deal is reverted. Restoring the balance in the example above should clear the violation.

Accounts that were not already relevant to reserve tracking should not become spuriously tracked merely because the context was refreshed.

### Suggested fix

Refresh the network-specific auxiliary context after successful `deal`, `cloneAccount`, and `loadAllocs` mutations, and after direct deal rollback. The existing `refresh_context_after_state_change::<FEN>` path used after other state replacement operations should provide the appropriate boundary.

Regression coverage should include:

- violation followed by `deal` restoration;
- restoration through `cloneAccount` and `loadAllocs`;
- `deal` followed by an outer-frame revert or expected revert;
- an untracked-account mutation that remains untracked.

haythemsellami:
`vm.deal`, `vm.cloneAccount`, and `vm.loadAllocs` mutate journal state outside the normal EVM transfer hooks. On Monad, the reserve-balance tracker could therefore retain stale thresholds and violation state, including when `DealRecord` cleanup restored balances after a root revert.

This rebases network auxiliary context after each successful state splice and after rollback restores recorded balances. The existing network hook preserves tracked-account membership while recomputing violations from the resulting state, with regressions covering each mutation path, unrelated accounts, expected reverts, and direct rollback.

Closes #16142


mablr (dismissed):

mablr (dismissed):

crates/cheatcodes/src/inspector.rs
```diff
@@ -1345,11 +1345,16 @@ impl<FEN: FoundryEvmNetwork> Cheatcodes<FEN> {
...
         // This will prevent overflow issues in revm's [`JournaledState::journal_revert`] routine
         // which rolls back any transfers.
+        let mut restored = false;
         while let Some(record) = self.eth_deals.pop() {
             if let Some(acc) = ecx.journal_mut().evm_state_mut().get_mut(&record.address) {
                 acc.info.balance = record.old_balance;
+                restored = true;
             }
         }
+        if restored {
```
figtracer: this misses `cloneAccount`/`loadAllocs` rollbacks: after either clears a violation, a later frame revert restores the 9 ether balance but leaves `dippedIntoReserve()` false because there is no deal record. can we rebase on every top-level revert?

haythemsellami: fixed and added regression for cloneAccount, loadAllocs, reverting constructors, and INVALID halts.

mablr (dismissed):

mablr:
just missing `forge fmt`

mablr (approved):

stevencartavia (approved):

PR #16155, opened 2026-08-12
fix(forge): use project root for installs

### LINKED ISSUE #7205: `forge install <lib>` installs at path of command, not root of project dir
### Component

Forge

- [X] Foundry
- [X] Foundryup

forge 0.2.0 (a436a0d 2024-02-20T00:16:37.237839814Z)

forge install <somelib> (command called outside of project root)

### Operating System

Linux

### Describe the bug

`forge install <somelib>` will install <somelib> outside the project root, causing confusing fs gymnastics to fix the project structure.
Example: pwd is myProject/lib/solmate and I call `forge install github.com/openzeppelin/openzeppelin-contracts`. This installs a `lib` dir at myProject/lib/solmate/lib/openzeppelin-contracts.
Forge install should be "aware" of project root where it is invoked.
Cheers.

figtracer:
## Motivation

Closes #7205.

Running `forge install` from inside an unconfigured nested Git checkout causes project discovery to stop at that checkout. This creates another dependency directory inside the nested repository instead of respecting the enclosing Foundry project.

## Solution

When `--root` is absent, `forge install` now prefers the nearest ancestor containing `foundry.toml` before using the existing Git-root fallback. An explicit root remains authoritative, and a nested repository with its own Foundry configuration remains its own project. Regression coverage exercises both nested-repository cases.

This PR was written with Codex, which assisted with issue investigation, code generation, and testing.

### DIFF
--- a/.changelog/install-project-root.md
+++ b/.changelog/install-project-root.md
@@ -1,0 +1,5 @@
+---
+forge: patch
+---
+
+Made `forge install` use an enclosing Foundry project when invoked from a nested Git repository without its own `foundry.toml`.
--- a/crates/forge/src/cmd/install.rs
+++ b/crates/forge/src/cmd/install.rs
@@ -67,7 +67,13 @@ pub struct InstallArgs {
 impl_figment_convert_basic!(InstallArgs);
 
 impl InstallArgs {
-    pub async fn run(self) -> Result<()> {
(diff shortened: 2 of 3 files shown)


mattsse (approved):
lgtm

mablr (approved):

grandizzy (approved):
lgtm

PR #16158, opened 2026-08-12
perf(forge): speed up gas reports

figtracer:
Profiling fuzz and invariant gas reports showed that every retained sample re-registered its identified ABI's functions, events, and errors globally, even though the gas-report decoder already loads all local ABIs up front. This keeps the per-sample address metadata needed for selector collisions and fallback handling, while reusing the existing global ABI metadata.

### Results

Cached full-suite `forge test --gas-report --fuzz-seed 0x1234` runs on Solady using profiling builds:

| Metric | master | this PR | Change |
| --- | ---: | ---: | ---: |
| Wall time | 23.01 s | 18.60 s | -19% |
| Peak RSS | 3.08 GB | 1.37 GB | -55% |

This PR was written with Amp, including code generation and analysis.

### DIFF
--- a/.changelog/faster-fuzz-gas-reports.md
+++ b/.changelog/faster-fuzz-gas-reports.md
@@ -1,0 +1,5 @@
+---
+forge: patch
+---
+
+Reduced gas report runtime and peak memory use for fuzz and invariant tests.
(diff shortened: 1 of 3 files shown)


mattsse (approved):
lgtm

mablr (approved):

PR #16159, opened 2026-08-12
fix(foundryup): clarify install status

mablr:
Foundryup previously used the same message for a normal first install and an installed binary that failed hash verification, making routine downloads look like integrity failures. This change reports the specific cache state and reserves SHA-256 mismatch language for an actual mismatch, while aborting clearly if a hash cannot be calculated.

### DIFF
--- a/foundryup/foundryup
+++ b/foundryup/foundryup
@@ -7,7 +7,7 @@ set -eo pipefail
 
 # NOTE: if you make modifications to this script, please increment the version number.
 # WARNING: the SemVer pattern: major.minor.patch must be followed as we use it to determine if the script is up to date.
-FOUNDRYUP_INSTALLER_VERSION="1.9.2"
+FOUNDRYUP_INSTALLER_VERSION="1.9.3"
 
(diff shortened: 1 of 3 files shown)


foundryup/foundryup
```diff
@@ -608,6 +586,77 @@ compute_sha256() {
...
+    fi
+    if [ ! -f "$path" ] || [ ! -x "$path" ]; then
+      printf 'not-executable:%s\n' "$bin"
+      return
+    fi
+    if [ -z "$expected" ]; then
+      printf 'unverifiable:%s\n' "$bin"
+      return
+    fi
+    if ! actual=$(compute_sha256 "$path" 2>/dev/null); then
```
stevencartavia, line 624: This handles cached binaries, but the post-download `actual=$(compute_sha256 "$path")` still exits immediately without a clear Foundryup error when hashing fails. Could we guard that assignment too?

stevencartavia (approved):

figtracer (approved):

Issue #10868, opened 2025-06-27
feat(`cheatcodes`): add `vm.ec{Add, Mul?, AddP256, MulP256?}` (elliptic-curve cryptography)


StackOverflowExcept1on:
### Component

Forge

### Describe the feature you would like

It would be great to have such methods in `vm` since in rollups we could see more complex smart contracts that use secp256k1 / secp256r1 elliptic-curves. Intended use case - multi-signature schemes, ECDH on-chain, etc.

### Additional context
`Mul` is marked with a question mark because we already have `vm.createWallet(uint256 privateKey)` and `vm.publicKeyP256(uint256 privateKey)`, but we don't have any generic method like `add/mul` that returns `struct PublicKey { uint256 x, y; }`.

0xrusowsky:
@StackOverflowExcept1on aren't you able to use solidity libraries for those? if not, we could discuss further

StackOverflowExcept1on:
@0xrusowsky this is literally what I'm doing now (https://github.com/StackOverflowExcept1on/frost-secp256k1-evm/blob/master/src/utils/cryptography/Secp256k1Arithmetic.sol), but it would be useful to have built-in cheatcodes for this since the same secp256k1 curve is the standard for Ethereum.

grandizzy:
@StackOverflowExcept1on @0xrusowsky I think would be OK to have a built-in cheatcode for such too, @zerosnacks wdyt?

zerosnacks:
I wouldn't be against this only note is that we have to be absolutely sure of the implementation of the cryptography. An implementation using k256 (a dependency we already use) seems straightforward. Regarding the naming, it needs to be specific enough to be future proof so perhaps .secp256k1 rather than .ec?

StackOverflowExcept1on:
Ok, I'd like us to design API for this feature.

zerosnacks:
@StackOverflowExcept1on any concrete suggestions for the API?

StackOverflowExcept1on:
@zerosnacks for example:
- https://github.com/StackOverflowExcept1on/frost-secp256k1-evm/blob/master/src/utils/cryptography/Secp256k1Arithmetic.sol
- https://github.com/verklegarden/crysol/blob/main/src/arithmetic/Points.sol
[...]

PR #16160, opened 2026-08-12
feat(crates/cheatcodes): add secp256k1 EC arithmetic

StackOverflowExcept1on:
## Motivation

Ref #10868

Foundry currently exposes secp256k1 key generation but does not provide generic elliptic-curve arithmetic. This makes testing protocols that use Schnorr signatures, threshold cryptography, ECDH, or stealth addresses unnecessarily difficult.

## Solution

Add `vm.ecAddAffine`, `vm.ecAddProjective`, `vm.ecMulAffine`, and `vm.ecMulProjective` cheatcodes backed by RustCrypto’s `k256` implementation.

The cheatcodes support affine and homogeneous projective coordinates, including canonical point-at-infinity representations. Inputs are validated using `k256` field and curve types.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/cheatcodes/spec/src/vm.rs
+++ b/crates/cheatcodes/spec/src/vm.rs
@@ -2987,6 +2987,49 @@ interface Vm {
     #[cheatcode(group = Crypto)]
     function publicKeyP256(uint256 privateKey) external pure returns (uint256 publicKeyX, uint256 publicKeyY);
 
+    /// Adds the secp256k1 affine points `point1 = (pointX1, pointY1)` and
+    /// `point2 = (pointX2, pointY2)`.
+    /// The point at infinity is represented as `(0, 0)`.
+    #[cheatcode(group = Crypto)]
+    function ecAddAffine(uint256 pointX1, uint256 pointY1, uint256 pointX2, uint256 pointY2)
+        external
+        pure
+        returns (uint256 resultX, uint256 resultY);
+
+    /// Adds the secp256k1 projective points `point1 = (pointX1, pointY1, pointZ1)` and
+    /// `point2 = (pointX2, pointY2, pointZ2)`.
+    /// The point at infinity is accepted as `(0, y, 0)` for any non-zero `y` and returned as
+    /// `(0, 1, 0)`. Any other result is normalized to `(x, y, 1)`.
+    #[cheatcode(group = Crypto)]
+    function ecAddProjective(
+        uint256 pointX1,
+        uint256 pointY1,
+        uint256 pointZ1,
+        uint256 pointX2,
+        uint256 pointY2,
+        uint256 pointZ2
+    )
+        external
+        pure
+        returns (uint256 resultX, uint256 resultY, uint256 resultZ);
+
+    /// Multiplies the secp256k1 affine point `(pointX, pointY)` by `scalar`.
+    /// The point at infinity is represented as `(0, 0)`.
+    #[cheatcode(group = Crypto)]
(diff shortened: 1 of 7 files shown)


mattsse (dismissed):
all of these look good, and imo doesnt hurt if we add.

wdyt @grandizzy ?

stevencartavia (dismissed):

crates/cheatcodes/spec/src/vm.rs
```diff
@@ -2987,6 +2987,49 @@ interface Vm {
     #[cheatcode(group = Crypto)]
     function publicKeyP256(uint256 privateKey) external pure returns (uint256 publicKeyX, uint256 publicKeyY);
 
+    /// Adds the secp256k1 affine points `point1 = (pointX1, pointY1)` and
+    /// `point2 = (pointX2, pointY2)`.
+    /// The point at infinity is represented as `(0, 0)`.
+    #[cheatcode(group = Crypto)]
+    function ecAddAffine(uint256 pointX1, uint256 pointY1, uint256 pointX2, uint256 pointY2)
```
mablr, line 2994: Could these names identify the curve, e.g. `secp256k1AddAffine`? Generic `ec*` names make future P-256 support ambiguous.

crates/cheatcodes/src/crypto.rs
```diff
@@ -463,6 +504,77 @@ fn sign_p256(private_key: &U256, digest: &B256) -> Result {
...
+}
+
+fn parse_field_element(value: &U256, name: &str) -> Result<FieldElement> {
+    FieldElement::from_bytes(&FieldBytes::from(value.to_be_bytes()))
+        .into_option()
+        .ok_or_else(|| fmt_err!("invalid secp256k1 {name}"))
+}
+
+fn parse_ec_scalar(scalar: &U256) -> Result<Scalar> {
+    Scalar::from_repr(scalar.to_be_bytes().into())
```
mablr: This rejects scalars ≥ the group order. Scalar multiplication should reduce the full `uint256` input modulo the order, so `n * P` returns infinity rather than reverting.

testdata/default/cheats/Ec.t.sol
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    function testEcAddProjective() public {
+        Vm.Wallet memory wallet1 = vm.createWallet(1);
+        Vm.Wallet memory wallet2 = vm.createWallet(2);
+
+        (uint256 x, uint256 y, uint256 z) = vm.ecAddProjective(0, 1, 0, GX, GY, 1);
+        _assertProjectivePoint(x, y, z, wallet1);
+
+        (x, y, z) = vm.ecAddProjective(GX, GY, 1, GX, GY, 1);
```
mablr: Please cover a non-unit homogeneous `z` and assert normalized `z == 1`. Boundary cases for invalid points, infinity results, and scalars `0`, `n`, and `n + 1` would also exercise the new contracts.

crates/cheatcodes/spec/src/vm.rs
StackOverflowExcept1on: Please take a look at `Vm.sol`. All functions there are named without the `secp256k1` prefix and use the `P256` suffix instead.

mablr: okay, fair

stevencartavia (approved):

mablr (approved):
ty!

grandizzy (approved):
thanks!

PR #16161, opened 2026-08-12
refactor(evm): clean-up FoundryEvmNetwork trait

mablr:
This restores `FoundryEvmNetwork` to a lean compatibility interface that only associates an Alloy network with its EVM factory. Execution-specific capabilities now live on `FoundryEvmFactory`, while policy such as contract size limits and beacon-root handling is derived from the resolved network configuration.

It also removes the redundant `supports_network` and cheatcode-address helpers. Script fork transitions retain the necessary Monad compatibility check using the active and requested network profiles directly.


crates/cast/src/cmd/call.rs
```diff
@@ -335,13 +335,6 @@ impl CallArgs {
             return Ok(());
         };
         evm_opts.infer_network_from_fork().await?;
-        let network = evm_opts.networks.execution_network();
-        if !FEN::supports_network(network) {
-            eyre::bail!(
-                "the selected EVM network cannot execute `{network}`; use the matching network \
-                 implementation"
-            );
-        }
```
mablr, line 344: Unsure if we need these blocks

crates/verify/src/bytecode.rs
```diff
@@ -298,13 +298,7 @@ impl VerifyBytecodeArgs {
         }
         let network_was_inferred = Self::configured_network(None, &config).is_none();
         let endpoint_identity = Self::endpoint_identity(&config).await?;
-        let network = Self::materialize_execution_network(&mut config, endpoint_identity.as_ref());
-        if !FEN::supports_network(network) {
-            eyre::bail!(
-                "the selected EVM network cannot execute `{network}`; use the matching network \
-                 implementation"
-            );
-        }
```
mablr, line 307: same

crates/verify/src/bytecode.rs
```diff
@@ -318,16 +312,6 @@ impl VerifyBytecodeArgs {
...
-        let network = Self::effective_network(
-            Self::configured_network(None, &config),
-            endpoint_identity.as_ref(),
-        );
-        if !FEN::supports_network(network) {
-            eyre::bail!(
-                "the selected EVM network cannot execute `{network}`; use the matching network \
-                 implementation"
-            );
-        }
```
mablr, line 330: same

crates/cast/src/cmd/call.rs
figtracer: i don't recall this support_networks function

crates/cast/src/cmd/call.rs
```diff
@@ -335,13 +335,6 @@ impl CallArgs {
             return Ok(());
         };
         evm_opts.infer_network_from_fork().await?;
```
stevencartavia: Could the network compatibility check be preserved in the public generic APIs? On base, using `EthEvmNetwork` with a Monad endpoint returns an error; on this PR it executes successfully, and `get_fork_material` returns an `OSAKA` environment with `MonadNine` context. CLI dispatch is unaffected, but library callers can now silently use mismatched semantics.

crates/cast/src/cmd/call.rs
mablr: It was added in Monad PR, I don't really see the purpose.

figtracer (dismissed):

crates/script/src/lib.rs
```diff
@@ -871,21 +871,25 @@ pub struct ScriptConfig<FEN: FoundryEvmNetwork> {
...
-        "fork network `{}` is incompatible with the active EVM",
-        evm_opts.networks.execution_network()
-    );
+    if let Some(active_networks) = active_networks
+        && !active_networks.supports_fork_source(&evm_opts.networks)
+    {
+        eyre::bail!(
+            "fork network `{}` is incompatible with the active EVM",
+            evm_opts.networks.execution_network()
+        );
```
mablr, line 891: We just check we haven't changed network, it was a footgun to check against early dispatched `FEN` which core purpose is actually to associate Network with its EVM.

figtracer (approved):

crates/evm/evm/src/executors/trace.rs
```diff
@@ -158,13 +158,6 @@ impl<FEN: FoundryEvmNetwork> TracingExecutor<FEN> {
         evm_opts.fork_block_number = config.fork_block_number;
         evm_opts.infer_network_from_fork().await?;
         let networks = evm_opts.networks;
```
stevencartavia, line 160: `TracingExecutor::<EthEvmNetwork>::get_fork_material` remains publicly callable and returns an `OSAKA` environment with Monad/`MonadNine` context for a Monad endpoint. This boundary should validate compatibility or be made non-public.

mablr: Agree, but this is not called and I do think it’s fine. I’ll continue the refactor/clean-up as follow-up.

stevencartavia (approved):

PR #16164, opened 2026-08-12
fix(cast): reject disabled networks in da-estimate

### LINKED ISSUE #16145: bug(cast): da-estimate misclassifies disabled networks

pucedoteth:
`cast da-estimate` inferred its network family from the RPC chain ID through the infallible `From<ChainId>` conversion, which deliberately swallows the classifier's error and defaults to Ethereum. A known chain belonging to a family the build does not support therefore resolved to Ethereum instead of being rejected: a stock build (which enables `optimism` but not `monad`) pointed at chain ID 143 silently produced an Ethereum DA estimate for a Monad endpoint. The other network-inference paths added in #15343 use the fallible classifier precisely to prevent this.

This resolves the chain ID through `NetworkVariant::from_known_chain_id`, propagates its unavailable-family error, and falls back to Ethereum only when the classifier returns `Ok(None)` for a genuinely unknown chain. An explicit `--network` selection still short-circuits before any RPC call, so it remains authoritative.

The test module was gated behind the non-default `monad` feature and so never compiled in a default build. It is now ungated, with only the Monad-specific cases feature-gated, and covers an unknown chain ID, Ethereum mainnet, OP mainnet, and the Monad IDs both with and without the feature.

Fixes #16145

Written with assistance from Claude Code, per CONTRIBUTING.md. Note that I was unable to compile locally — the dependency fetch did not complete in my environment — so this change is unverified beyond review; CI is the first real check.

### DIFF
--- a/.changelog/cast-da-estimate-known-networks.md
+++ b/.changelog/cast-da-estimate-known-networks.md
@@ -1,0 +1,5 @@
+---
+cast: patch
+---
+
+Fixed `cast da-estimate` silently using Ethereum data availability semantics for a known chain whose network family is not enabled in the build.
--- a/crates/cast/src/cmd/da_estimate.rs
+++ b/crates/cast/src/cmd/da_estimate.rs
@@ -33,7 +33,7 @@ impl DAEstimateArgs {
             Some(n) => n,
             None => {
                 let provider = ProviderBuilder::<AnyNetwork>::from_config(&config)?.build()?;
-                provider.get_chain_id().await?.into()
+                infer_network(provider.get_chain_id().await?)?
             }
         };
         match network {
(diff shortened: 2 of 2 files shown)


mablr (commented):
Thanks for your contribution, but I think we can keep things as is.

PR #16169, opened 2026-08-13
fix(symbolic): preserve call semantics

stevencartavia:
Corrects divergences between symbolic and concrete EVM execution. Static calls now preserve both feasible zero-value and reverting nonzero-value paths, while unknown symbolic targets retain insufficient-funds failures instead of silently dropping them.

`CALLCODE` now checks and transfers value from the effective caller—including pranked callers—to the current execution context, with proper rollback on revert. Reverted contract creation also exposes constructor revert data through `RETURNDATA`, matching revm behavior.

### DIFF
--- a/crates/evm/symbolic/src/executor/calls.rs
+++ b/crates/evm/symbolic/src/executor/calls.rs
@@ -879,40 +879,45 @@ impl SymbolicExecutor {
             }
         }
         let code_address = self.function_mock_target(state, to, &call_input)?.unwrap_or(to);
+        let call_context =
+            (!matches!(kind, CallKind::DelegateCall)).then(|| state.prank_for_next_call());
+        let transfer_to = if matches!(kind, CallKind::Call) { to } else { state.address };
+        if matches!(kind, CallKind::Call | CallKind::CallCode) {
+            let call_caller = call_context.as_ref().expect("value calls have a call context").0;
+            if !self.prepare_value_transfer(
+                executor,
+                state,
+                worklist,
+                call_caller,
+                transfer_to,
+                value.clone(),
+                out_offset.clone(),
+                &out_size,
+            )? {
+                return Ok(StepOutcome::Continue);
+            }
+        }
         if !state.call_mocks.is_empty() {
             let concrete_value = state.constrained_word(&mut self.cx, &value);
             if let Some(mock) =
                 self.take_call_mock(state, code_address, concrete_value, &call_input)?
             {
-                if !matches!(kind, CallKind::DelegateCall) {
-                    let _ = state.prank_for_next_call();
-                }
                 let (return_data, reverts) = mock.into_parts();
                 state.return_data = return_data;
+                if !reverts && let Some((call_caller, _, _)) = call_context {
+                    self.apply_call_value_transfer(executor, state, kind, to, call_caller, value);
+                }
                 state.copy_call_output_offset(&mut self.cx, out_offset, &out_size)?;
                 let success = SymExpr::constant(&mut self.cx, U256::from(!reverts));
                 state.stack.push(success)?;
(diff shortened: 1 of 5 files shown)


mattsse (approved):
lgtm

crates/evm/symbolic/src/executor/calls.rs
```diff
@@ -867,7 +900,7 @@ impl SymbolicExecutor {
             return Err(SymbolicError::Unsupported("symbolic prank delegatecall"));
         }
         let (call_caller, call_caller_word, pranked_origin) = state.prank_for_next_call();
-        if matches!(kind, CallKind::Call)
+        if matches!(kind, CallKind::Call | CallKind::CallCode)
```
mablr: Could we run this balance check before `take_call_mock` and apply the transfer on mocked success? Currently mocked `CALLCODE` can succeed with insufficient funds and never updates either balance.

crates/forge/tests/cli/test_cmd/symbolic_creates.rs
```diff
@@ -813,6 +813,47 @@ contract SymbolicCreateValue is Test {
...
+    function checkCreateRevertData() public {
+        bytes memory initcode = hex"61123460005260206000fd";
+        address created;
+        uint256 size;
+        assembly {
+            created := create(0, add(initcode, 0x20), mload(initcode))
+            size := returndatasize()
+        }
+        assert(created == address(0));
+        assert(size == 32);
```
mablr, line 840: Could this also assert the returned `0x1234` payload? Checking only the size would not catch incorrect revert contents.

crates/evm/symbolic/src/executor/calls.rs
```diff
@@ -1278,6 +1303,23 @@ impl SymbolicExecutor {
...
+        to: Address,
+        from: Address,
+        value: SymExpr,
+    ) {
+        let to = match kind {
+            CallKind::Call => to,
+            CallKind::CallCode => state.address,
+            CallKind::DelegateCall | CallKind::StaticCall => return,
+        };
+        state.world.transfer(&mut self.cx, executor, from, to, value);
```
mattsse, line 1325: Please model recipient-balance overflow before applying this transfer. With `address(this).balance == type(uint256).max`, a pranked caller holding 1 wei, and a `CALLCODE` value of 1, revm returns failure (`OverflowPayment`) and leaves the recipient at `uint256.max`. This path instead wraps the recipient balance to zero and treats the call as successful, so a property matching the concrete result ends as `RevertAll`. The added tests only start the recipient at zero and do not cover this case.

mattsse (changes requested):
One concrete symbolic/concrete divergence remains.

crates/evm/symbolic/src/executor/calls.rs
stevencartavia: [11a8de9](https://github.com/foundry-rs/foundry/pull/16169/commits/11a8de98f63029ddfe708e067838d4d2488eff33)

crates/forge/tests/cli/test_cmd/symbolic_creates.rs
stevencartavia: [11a8de9](https://github.com/foundry-rs/foundry/pull/16169/commits/11a8de98f63029ddfe708e067838d4d2488eff33)

crates/evm/symbolic/src/executor/calls.rs
stevencartavia: [11a8de9](https://github.com/foundry-rs/foundry/pull/16169/commits/11a8de98f63029ddfe708e067838d4d2488eff33)

mablr (approved):

mattsse (approved):
lgtm

PR #16173, opened 2026-08-13
feat(cli): add gas estimate multiplier

### LINKED ISSUE #1803: feat(`forge create / cast send`): add `--gas-estimate-multiplier` flag
### Component

Forge

For certain networks, gas(price) may be incorrectly estimated, resulting in transactions that are underpriced: 

https://github.com/foundry-rs/foundry/issues/1703

in `forge create` we use the default `Provider::{estimate_gas,get_gas_price}` functions to fill the transactions, if they're not provided via cli arguments

it would be useful to scale these estimates percentage-wise, perhaps `--scale-gas-limit 20` to use a `1.2` multiplier for the gas limit.

### Additional context

_No response_

figtracer:
## Motivation

`forge script` supports padding provider gas estimates, but `forge create` and `cast send` do not. Closes #1803.

## Solution

Add an opt-in `--gas-estimate-multiplier <PERCENT>` to both commands. Omitting it preserves existing estimates, and explicit `--gas-limit` values remain unchanged. This change was developed with AI assistance.

### DIFF
--- a/crates/cli/src/utils/mod.rs
+++ b/crates/cli/src/utils/mod.rs
@@ -46,6 +46,14 @@ pub const STATIC_FUZZ_SEED: [u8; 32] = [
     0x5d, 0x64, 0x0b, 0x19, 0xad, 0xf0, 0xe3, 0x57, 0xb8, 0xd4, 0xbe, 0x7d, 0x49, 0xee, 0x70, 0xe6,
 ];
 
+/// Applies an optional percentage multiplier to a gas estimate.
+pub const fn apply_gas_estimate_multiplier(estimate: u64, multiplier: Option<u64>) -> u64 {
+    match multiplier {
+        Some(multiplier) => estimate * multiplier / 100,
+        None => estimate,
+    }
+}
+
 /// Regex used to parse `.gitmodules` file and capture the submodule path and branch.
 pub static SUBMODULE_BRANCH_REGEX: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r#"\[submodule "([^"]+)"\](?:[^\[]*?branch = ([^\s]+))"#).unwrap());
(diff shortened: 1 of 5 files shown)


crates/cli/src/utils/mod.rs
```diff
@@ -46,6 +46,14 @@ pub const STATIC_FUZZ_SEED: [u8; 32] = [
     0x5d, 0x64, 0x0b, 0x19, 0xad, 0xf0, 0xe3, 0x57, 0xb8, 0xd4, 0xbe, 0x7d, 0x49, 0xee, 0x70, 0xe6,
 ];
 
+/// Applies an optional percentage multiplier to a gas estimate.
+pub const fn apply_gas_estimate_multiplier(estimate: u64, multiplier: Option<u64>) -> u64 {
+    match multiplier {
+        Some(multiplier) => estimate * multiplier / 100,
```
mattsse: `estimate * multiplier` can overflow before the division. Both new flags accept any `u64`; for example, an estimate of `21_000` with multiplier `1_000_000_000_000_000` should produce `210_000_000_000_000_000` (which still fits in `u64`), but release builds wrap to `25_532_559_262_904_483` and debug builds panic. Please use widened or checked arithmetic (or reject overflowing inputs) and add a boundary test.

mattsse (changes requested):
One correctness issue needs addressing before merge.

crates/cli/src/utils/mod.rs
figtracer: Fixed in 7ba2530d8: the multiplication now uses a `u128` intermediate, checked conversion rejects results above `u64::MAX`, and the regression test covers both the reported large multiplier and a final-result overflow.

mattsse (approved):
lgtm

mablr (approved):

PR #16175, opened 2026-08-13
refactor(evm): isolate Monad context state

mablr:
This follows #15343 and #16161 by moving Monad execution state behind the factory boundary established by the network cleanup. Monad needs exact chain-position metadata for reserve rules and transaction-scoped reserve-balance tracking across nested EVM and snapshot boundaries, but those concerns should not be exposed through generic Foundry context APIs.

This replaces the generic auxiliary context wrapper with a factory-owned `ChainContext`, using `MonadChainContext` for Monad and `()` for other EVM families. Reserve tracking remains separate transaction-lifecycle state with explicit capture and restore hooks, while snapshots store the journal and chain context as one atomic entry so they cannot drift. The result keeps `FoundryContextExt` structural, preserves nested and isolated execution semantics, and confines Monad-specific behavior and dependencies to the `monad` feature.

### DIFF
--- a/.changelog/isolate-monad-context-state.md
+++ b/.changelog/isolate-monad-context-state.md
@@ -1,0 +1,10 @@
+---
+anvil: patch
+cast: patch
+forge-verify: patch
+foundry-cheatcodes: patch
+foundry-evm: patch
+foundry-evm-core: patch
+---
+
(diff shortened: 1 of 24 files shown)


figtracer (approved):

mattsse (approved):
lgtm

PR #16178, opened 2026-08-13
fix(ci): retry transient MPP failures

decofe:
Moderato's MPP gateway intermittently rejects valid vouchers or replaces a reusable channel, which makes the external end-to-end check flaky. Retry only those known gateway failure signatures up to three times, using a fresh Tempo home on every attempt so a failed wallet or channel store cannot affect the next run. Other failures still stop immediately.

This CI-only change needs the `L-ignore` label instead of a release changelog entry.

This PR was written with Codex.

Prompted by: @DaniPopes


DaniPopes (approved):

stevencartavia (approved):

mattsse (approved):
lgtm

mablr (approved):

PR #7606, opened 2024-04-09
fix(forge): selectively enable Etherscan trace resolving when a test in ran in a forked environment and return block number in trace on a failed test

zerosnacks:
## Motivation

Closes https://github.com/foundry-rs/foundry/issues/7555 (bug)
Closes https://github.com/foundry-rs/foundry/issues/7312 (bug)
Closes https://github.com/foundry-rs/foundry/issues/7574 (feature)

In short, when an `ETHERSCAN_API_KEY` is defined in a users' `.env` it would trigger Etherscan trace resolving for local tests which would result in a massive slowdown after running into rate limiting.

## Solution

Implements a way to mark test results as whether having ran in a forked or non-forked environment. This is then used to selectively enable or disable Etherscan tracing (without re-initializing). 

As requested in #7574, upon a encountering a test failure in a forked test it will now display the fork block number. On succesful runs it is not displayed.

**Known limitations:**
- Generally unable to differentiate remote forks and local forks
- When using `vm.roll(n)` local forks are created

Given that trace caching is now resolved these limitations are not that big of a deal, at worst the experience is as bad as it is now but only for the initial run.

The chosen approach was to add high-level context tracking which can easily be expanded to track additional call context related information.

PR #16179, opened 2026-08-13
feat(forge): show fork block on failures

### LINKED ISSUE #7574: Display block number used on fork tests
### Component

Forge

There are times that fork tests will fail on very specific blocks. Displaying the block number on which the fork test failed would be helpful in repro-ing the issue.

### Additional context

_No response_

figtracer:
## Motivation

Closes #7574.

Failed fork tests currently omit the fork block at normal verbosity, which makes failures tied to a specific chain state harder to reproduce. The issue discussion suggested carrying fork information in the test result, while the earlier #7606 explored a broader execution-context collector that was ultimately closed.

## Solution

Capture the backend's canonical active fork block after EVM execution and propagate it through raw calls, fuzz failures, setup, and Forge test results. Failed results now render `(block: N)` at normal verbosity; successful tests and failures outside a fork keep their existing output.

This uses the existing backend fork-block API, including its handling of pinned and environment block numbers, instead of reviving the separate context-collection machinery. A full CLI regression runs against a local Anvil and covers failed and successful fork and non-fork tests.

This PR was written with Codex, which assisted with issue investigation, implementation, and testing.


mablr (commented):
cool

mablr (dismissed):

crates/forge/src/result.rs
```diff
@@ -2271,6 +2278,7 @@ impl TestResult {
             traces: setup.traces.clone(),
             debug_bytecodes: setup.debug_bytecodes.clone(),
             line_coverage: setup.coverage.clone(),
+            fork_block_number: setup.fork_block_number,
```
mattsse, line 2281: `TestResult::new` initializes this from `setUp`, but `invariant_result` never replaces it with the failing worker or replay's fork state. In the updated `invariant_roll_fork` fixture, the predicate cannot fail until `vm.rollFork` advances to at least block 19812634, yet the blessed output reports 19812632. That gives users a stale block that does not reproduce the reported failure. Please propagate the failure-time fork block through the invariant result path, and assert the rolled block rather than the setup block.

mattsse (changes requested):
One correctness issue in the invariant result path.

crates/evm/evm/src/executors/mod.rs
```diff
@@ -1535,7 +1545,19 @@ fn convert_executed_result<FEN: FoundryEvmNetwork>(
...
 ) -> eyre::Result<RawCallResult<FEN>> {
+    // The backend carries the canonical fork pin, but a call may advance the execution block
+    // without publishing its copy-on-write backend (notably invariant `rollFork` calls). Use the
+    // post-call EVM block on regular chains so a later failing invariant observes that advancement.
+    // Arbitrum is the exception: its EVM block can be the lower L1 block while the fork pin is L2.
+    let fork_block_number = fork_block_number.map(|block| {
+        if arbitrum::is_arbitrum_chain(evm_env.cfg_env.chain_id) {
+            block
+        } else {
+            evm_env.block_env.number().saturating_to::<u64>()
```
stevencartavia, line 1558: `vm.roll` changes `block.number` without rolling the backing fork, but this reports the rolled value as the fork block. For example, a fork pinned at block 7 followed by `vm.roll(99)` reports `(block: 99)`, although the backing fork remains pinned at block 7. Could we preserve the canonical fork block here and propagate `vm.rollFork` separately?

mattsse (approved):
lgtm

mablr (approved):

Issue #16181, opened 2026-08-13
release channels inconsistent: docker `latest` tracks nightly, `stable` frozen at v1.5.1, foundryup `nightly` stale


ericvelazquez:
### Component

Other (release / distribution channels)

### Describe the bug

After the release-pipeline rework (finalize-release.py, #16022, Aug 5), the distribution channel pointers are inconsistent with each other and with the promotion logic in the script itself:

**1. Docker `latest` now tracks nightly builds.** As of 2026-08-13:

```
latest  -> sha256:ede4c35d...   (identical digest to `nightly`)
nightly -> sha256:ede4c35d...
```

OCI labels on `latest`: `version: nightly`, `revision: ae4490e2...`, `created: 2026-08-13T06:25Z`. Until roughly Aug 5, `latest` pointed at the v1.7.1 digest (`sha256:8347b7...`) and had been stable for ~3 months. `finalize-release.py`'s own `stable_aliases()` says only the highest stable release should receive `latest`, so this looks unintended. Practical impact: every unpinned `foundry:latest` user silently switched from v1.7.1 to daily nightlies last week, and today's nightly cannot fork OP-stack chains (see #16180), which broke our CI fleet.

**2. Docker `stable` is frozen at v1.5.1.** `stable` points at a build from 2025-12-22 (`revision b0a9dd9c...`, i.e. v1.5.1), predating v1.6.x and v1.7.x entirely.

**3. The `nightly` GitHub release / foundryup nightly channel is stale.** `foundryup -i nightly` installs `1.6.0-nightly` built 2026-04-28 (commit `5e88010a`), while actual nightlies are published daily as `nightly-<sha>` releases.

Verification commands:

```bash
docker buildx imagetools inspect ghcr.io/foundry-rs/foundry:latest
docker buildx imagetools inspect ghcr.io/foundry-rs/foundry:nightly
docker buildx imagetools inspect ghcr.io/foundry-rs/foundry:stable
foundryup -i nightly && anvil --version
```

Expected: `latest` -> newest stable release (v1.7.1), `stable` -> newest stable release, `nightly` (docker tag, GitHub release, foundryup channel) -> newest nightly.

mablr:
Stable is deprecated since v1.7.0 and the introduction of immutable releases.

ericvelazquez:
Thanks, that explains the `stable` tag - happy to drop that point.

The two remaining pointers are the ones that bit us though:

[...]

PR #16182, opened 2026-08-13
fix(ci): stop nightlies overwriting latest

stevencartavia:
Depot normalizes an untagged multi-platform image output to `:latest`, causing nightly builds to overwrite the stable Docker alias before promotion. Publish the merged image under the internal `:build-staging` tag instead, while retaining the existing digest-based signing, attestation, and promotion flow.

This addresses the Docker regression reported in #16181


DaniPopes (approved):

mattsse (approved):
lgtm

PR #16183, opened 2026-08-14
fix(symbolic): reject oversized memory offsets

stevencartavia:
Rejects unrepresentable offsets for symbolic `MLOAD`, `MSTORE`, and `MSTORE8`, matching concrete EVM failure behavior and preventing false passes.


mattsse (dismissed):
lgtm

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -290,6 +290,56 @@ impl SymbolicExecutor {
...
+        let representable =
+            SymBoolExpr::cmp(&mut self.cx, SymCmpOp::Ule, offset.clone(), max_offset);
+        let (valid_constraints, valid_sat) =
+            self.constraints_with_condition(state, representable.clone())?;
+        let invalid = representable.not(&mut self.cx);
+        let (invalid_constraints, invalid_sat) = self.constraints_with_condition(state, invalid)?;
+        match (valid_sat, invalid_sat) {
+            (true, true) => {
+                let mut valid = state.clone();
+                valid.pc = valid.pc.saturating_sub(1);
```
mablr, line 359: Blocker: also restore `valid.depth` before re-queueing. The execution loop increments depth before `step`, so replaying this opcode currently counts it twice and can hit the depth limit one instruction early.

crates/forge/tests/cli/test_cmd/symbolic_memory.rs
```diff
@@ -50,6 +50,72 @@ contract SymbolicMload {
...
+        uint256 offset = type(uint256).max;
+        (bool loadOk,) = address(this).call(abi.encodeCall(this.load, (offset)));
+        (bool storeOk,) = address(this).call(abi.encodeCall(this.store, (offset)));
+        (bool store8Ok,) = address(this).call(abi.encodeCall(this.store8, (offset)));
+        assertFalse(loadOk);
+        assertFalse(storeOk);
+        assertFalse(store8Ok);
+    }
+
+    function checkConstrainedOversizedMemoryAccess(uint256 offset) public {
```
mablr, line 95: Worth adding an unconstrained-offset case where both sides are satisfiable. These cases only exercise wholly invalid offsets, so they do not cover the rewind/re-queue path or prove the valid sibling remains explored.

stevencartavia: [31d45fc](https://github.com/foundry-rs/foundry/pull/16183/commits/31d45fc9304dbe9b8b659386af17fb7a351a4b53)

crates/evm/symbolic/src/executor/opcodes.rs
stevencartavia: [31d45fc](https://github.com/foundry-rs/foundry/pull/16183/commits/31d45fc9304dbe9b8b659386af17fb7a351a4b53)

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -290,6 +290,57 @@ impl SymbolicExecutor {
...
     }
 
+    fn guard_fixed_memory_access(
+        &mut self,
+        state: &mut PathState,
+        worklist: &mut VecDeque<PathState>,
+        offset: &SymExpr,
+        size: usize,
+    ) -> Result<Option<StepOutcome>, SymbolicError> {
+        let max_offset = usize::MAX - size;
```
mablr: Blocker: account for 32-byte memory rounding here. `usize::MAX - size` still allows 31 offsets whose rounded memory size is unrepresentable and silently saturates in `SymMemory::size_after_access`. Use `(usize::MAX & !31usize) - size` and cover the first invalid boundary.

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -290,6 +290,57 @@ impl SymbolicExecutor {
...
+        let (valid_constraints, valid_sat) =
+            self.constraints_with_condition(state, representable.clone())?;
+        let invalid = representable.not(&mut self.cx);
+        let (invalid_constraints, invalid_sat) = self.constraints_with_condition(state, invalid)?;
+        match (valid_sat, invalid_sat) {
+            (true, true) => {
+                let mut valid = state.clone();
+                valid.pc = valid.pc.saturating_sub(1);
+                valid.depth = valid.depth.saturating_sub(1);
+                valid.constraints = valid_constraints;
```
mablr, line 361: Please partition corpus seed models on `representable` before requeueing. Keeping invalid-only seeds on `valid` can distort BFS/DFS priority under path or timeout limits; `split_corpus_seed_models` already provides the needed split.

mablr (changes requested):

crates/evm/symbolic/src/executor/opcodes.rs
stevencartavia: [f418ffb](https://github.com/foundry-rs/foundry/pull/16183/commits/f418ffb8fe5dcaa7cfb3f6bc6a248fbad664efd4)

crates/evm/symbolic/src/executor/opcodes.rs
stevencartavia: [f418ffb](https://github.com/foundry-rs/foundry/pull/16183/commits/f418ffb8fe5dcaa7cfb3f6bc6a248fbad664efd4)

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -290,6 +290,61 @@ impl SymbolicExecutor {
...
     }
 
+    fn guard_fixed_memory_access(
+        &mut self,
+        state: &mut PathState,
+        worklist: &mut VecDeque<PathState>,
+        offset: &SymExpr,
+        size: usize,
+    ) -> Result<Option<StepOutcome>, SymbolicError> {
+        let max_offset = (usize::MAX & !31usize) - size;
```
figtracer: blocker: this also needs to respect the configured `memory_limit`, not just `usize::MAX`. with `memory_limit = 4096`, `mstore(4096, 1)` succeeds symbolically but replays as `MemoryLimitOOG`; please bound the rounded end by the configured limit and add a regression.

stevencartavia: [082cbc4](https://github.com/foundry-rs/foundry/pull/16183/commits/082cbc4c806702f31ebfe660406d95c597ddbcf5)

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -290,6 +290,70 @@ impl SymbolicExecutor {
...
 
+    fn guard_fixed_memory_access<FEN: FoundryEvmNetwork>(
+        &mut self,
+        executor: &Executor<FEN>,
+        state: &mut PathState,
+        worklist: &mut VecDeque<PathState>,
+        offset: &SymExpr,
+        size: usize,
+    ) -> Result<Option<StepOutcome>, SymbolicError> {
+        let max_memory = usize::try_from(executor.evm_env().cfg_env.memory_limit())
```
mablr: Blocker: revm's memory limit is cumulative across call frames (`ancestor checkpoint + rounded local size`). This gives each child the full limit, so nested calls can pass symbolically but fail replay. Track inherited memory usage, bound against the remaining capacity, and add a nested-call regression.

crates/forge/tests/cli/test_cmd/symbolic_memory.rs
```diff
@@ -50,6 +50,158 @@ contract SymbolicMload {
...
+        vm.assume(offset == type(uint256).max);
+        (bool ok,) = address(this).call(abi.encodeCall(this.store, (offset)));
+        assertFalse(ok);
+    }
+
+    function checkOversizedRoundedMemoryAccess() public {
+        uint256 offset = (uint256(type(uint64).max) & ~uint256(31)) - 31;
```
mablr: This no longer isolates rounding: the offset already exceeds the default 128 MiB limit. Use a small configured limit and cover the exact-valid and first-invalid boundaries, including `MSTORE8`.

stevencartavia: [07f0052](https://github.com/foundry-rs/foundry/pull/16183/commits/07f0052fb17c9fa3a49e17696bd4d91d8f994c7e)

crates/evm/symbolic/src/executor/opcodes.rs
stevencartavia: [07f0052](https://github.com/foundry-rs/foundry/pull/16183/commits/07f0052fb17c9fa3a49e17696bd4d91d8f994c7e)

crates/evm/symbolic/src/runtime/state.rs
```diff
@@ -203,7 +203,8 @@ impl PathState {
         }
     }
 
-    pub(crate) fn child(&self, frame: CallFrame) -> Self {
+    pub(crate) fn child(&self, cx: &mut SymCx, mut frame: CallFrame) -> Self {
+        frame.memory_checkpoint = self.total_memory_size(cx);
```
mablr: This misses CALL input/output memory expansion: those ranges are not added to `SymMemory` before creating the child. With `memory_limit = 4096`, `CALL(..., 2048, 2048, 0, 0)` followed by child `MSTORE8(0)` succeeds symbolically but fails in revm. Please track CALL range expansion before child creation and cover this case.

crates/evm/symbolic/src/runtime/state.rs
```diff
@@ -1470,10 +1472,16 @@ impl CallFrame {
...
             calldata,
             stack: SymStack::default(),
+            memory_checkpoint: SymExpr::zero(cx),
             memory: SymMemory::default(),
             return_data: SymReturnData::empty(cx),
         }
     }
+
+    pub(crate) fn total_memory_size(&self, cx: &mut SymCx) -> SymExpr {
+        let size = self.memory.size_word(cx);
```
mablr: `size_word()` can overstate symbolic-size copies because the byte model materializes `max_size` bytes even when the path size is zero. Using that as a child checkpoint can reject feasible siblings. Please track logical expansion separately and add a zero/nonzero symbolic-size regression.

stevencartavia: [b855ae7](https://github.com/foundry-rs/foundry/pull/16183/commits/b855ae748b7e1aa0a41c44d3362f49b30a3b2ada)

crates/evm/symbolic/src/runtime/state.rs
stevencartavia: [b855ae7](https://github.com/foundry-rs/foundry/pull/16183/commits/b855ae748b7e1aa0a41c44d3362f49b30a3b2ada)

crates/evm/symbolic/src/runtime/memory.rs
```diff
@@ -79,6 +72,22 @@ impl SymbolicMemoryWrite {
 }
 
 impl SymMemory {
+    pub(crate) fn size_after_access_word(cx: &mut SymCx, offset: SymExpr, len: usize) -> SymExpr {
+        let size = SymExpr::constant(cx, U256::from(len));
+        Self::size_after_range_word(cx, offset, size)
+    }
+
+    fn size_after_range_word(cx: &mut SymCx, offset: SymExpr, size: SymExpr) -> SymExpr {
+        let end = SymExpr::binop(cx, SymBinOp::Add, offset, size.clone());
```
mablr: This host-memory extent uses wrapping EVM arithmetic. An unchecked CALL range such as `(type(uint256).max, 1)` wraps to a zero checkpoint, so a callee `MSTORE8(0, 1)` succeeds symbolically even though revm rejects the CALL offset before entering the callee. Please validate nonzero ranges for representability, rounding/addition overflow, and `memory_limit` before recording them, and make cumulative checkpoint addition non-wrapping; add direct and nested wrap regressions.

stevencartavia: [92f6018](https://github.com/foundry-rs/foundry/pull/16183/commits/92f6018e40b6e0b8448dc3014a4b8faccf619b35)

mablr (changes requested):

mablr (dismissed):

crates/evm/symbolic/src/runtime/state.rs
figtracer: blocker: this inherited checkpoint does not match forge's concrete executor. on this head with `memory_limit = 4096`, a parent `mstore(2048, 1)` followed by a self-call whose fallback also executes `mstore(2048, 1)` succeeds concretely, but symbolic makes the call fail, so `assert(!ok)` is incorrectly proven. `checkMemoryLimitNestedCall` only runs under `--symbolic`, which is why it misses the divergence. please remove parent memory from the symbolic limit unless the concrete executor is changed too, and run this regression in both lanes.

crates/evm/symbolic/src/executor/create.rs
```diff
@@ -44,6 +45,19 @@ impl SymbolicExecutor {
         let salt =
             if matches!(kind, CreateKind::Create2) { Some(state.stack.pop()?) } else { None };
 
+        let size_word = size.size_word(&mut self.cx);
```
figtracer: blocker: this early return still bypasses the new range guard. for `create(0, 0, not(0))`, concrete execution raises `InvalidOperandOOG` in the current frame, while this branch pushes zero and continues. i confirmed a low-level self-call followed by `assert(ok)` passes symbolically and fails concretely on this head. treat the unrepresentable nonzero size as the frame failure and cover both `CREATE` and `CREATE2`.

crates/evm/symbolic/src/executor/calls.rs
```diff
@@ -9,6 +9,7 @@ impl SymbolicExecutor {
         completed_paths: &mut usize,
         kind: CallKind,
     ) -> Result<StepOutcome, SymbolicError> {
+        let memory_rewind_state = state.clone();
```
figtracer: perf: this clones the entire `PathState` on every call, including the common concrete-range path where neither guard can fork. that copies the world, memory histories, constraints, hooks, mocks, and other collections on one of the hottest opcodes. please guard before mutating the stack, or snapshot only the operands/stack, so a full state clone is paid only when both range branches are satisfiable. the same applies to `create`.

crates/evm/symbolic/src/runtime/memory.rs
```diff
@@ -44,7 +44,9 @@ pub(crate) enum BoundedCopySize {
 #[derive(Clone, Debug, Default)]
 pub(crate) struct SymMemory {
     symbolic_writes: Vec<SymbolicMemoryWrite>,
-    size: usize,
+    symbolic_access_sizes: Vec<SymExpr>,
```
figtracer: perf: retaining every symbolic access makes each state clone copy an ever-growing vector, and every `size_word()` refolds the full history into the same max expression. a single running symbolic logical-size expression, updated with `max` at each access, represents the same value while making state clones and size reads constant-time. this is particularly important in the symbolic hot path where expression clone/drop traffic already dominates.

figtracer (commented):
found two concrete/symbolic parity blockers and two hot-path representation costs on the current head. the full symbolic crate suite and the focused cli tests pass, but one added regression currently codifies a false proof.

crates/evm/symbolic/src/runtime/memory.rs
stevencartavia: [1c635cf](https://github.com/foundry-rs/foundry/pull/16183/commits/1c635cfe516648d12e56a1937bb5c5d7368740de)

crates/evm/symbolic/src/executor/calls.rs
stevencartavia: [1c635cf](https://github.com/foundry-rs/foundry/pull/16183/commits/1c635cfe516648d12e56a1937bb5c5d7368740de)

crates/evm/symbolic/src/executor/create.rs
stevencartavia: [1c635cf](https://github.com/foundry-rs/foundry/pull/16183/commits/1c635cfe516648d12e56a1937bb5c5d7368740de)

crates/evm/symbolic/src/runtime/state.rs
stevencartavia: [1c635cf](https://github.com/foundry-rs/foundry/pull/16183/commits/1c635cfe516648d12e56a1937bb5c5d7368740de)

mablr (dismissed):
I don't see other blockers, sgtm

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -290,6 +290,135 @@ impl SymbolicExecutor {
...
+            (false, true) => {
+                state.constraints = invalid_constraints;
+                state.return_data = SymReturnData::empty(&mut self.cx);
+                Ok(Some(StepOutcome::Revert))
+            }
+            (false, false) => Ok(Some(StepOutcome::AssumeRejected)),
+        }
+    }
+
+    pub(super) fn guard_memory_range<FEN: FoundryEvmNetwork>(
```
figtracer, line 382: blocker: the new range guard is still not applied to copy opcodes. on this head, `calldatacopy(not(0), 0, 1)` fails in the concrete executor but succeeds symbolically, so `assert(ok)` is falsely proven. please guard the destination/size range here and the sibling memory-consuming paths, with concrete/symbolic parity coverage.

stevencartavia: [c2d9cc9](https://github.com/foundry-rs/foundry/pull/16183/commits/c2d9cc9faffaf4e1995cafaa4cfd52698ed22985)

mablr, line 382: `guard_memory_range` only validates that the access fits the memory limit; it never grows `logical_size`. Read-only ops that route through it (KECCAK256, LOGn, MCOPY's `src`) leave memory size stale, so a later `MSIZE` won't reflect the expansion a real EVM would have performed for these reads.

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -290,6 +290,135 @@ impl SymbolicExecutor {
         Ok(true)
     }
 
+    fn guard_fixed_memory_access<FEN: FoundryEvmNetwork>(
```
mablr, line 294: This and `guard_memory_range` below duplicate ~25 lines of near-identical fork/clone/split-corpus-seed/push-to-worklist logic on the same `(valid_sat, invalid_sat)` match. Worth factoring into a shared helper so a future fix to the split idiom doesn't need to be applied twice.

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -290,6 +290,135 @@ impl SymbolicExecutor {
...
+            }
+            (false, true) => {
+                state.constraints = invalid_constraints;
+                state.return_data = SymReturnData::empty(&mut self.cx);
+                Ok(Some(StepOutcome::Revert))
+            }
+            (false, false) => Ok(Some(StepOutcome::AssumeRejected)),
+        }
+    }
+
```
mablr, line 445: Unlike `guard_fixed_memory_access`'s `local_size.as_const()` fast path, this always issues two `constraints_with_condition` solver calls, even for fully concrete offset/size that are trivially in-bounds. Once `size_word()` becomes non-constant (any earlier symbolic-length write in the path), every guarded opcode here pays two SMT queries unconditionally.

stevencartavia: [19e1895](https://github.com/foundry-rs/foundry/pull/16183/commits/19e189533daf051c6122a761697038b33f846a36)

crates/evm/symbolic/src/executor/opcodes.rs
stevencartavia: [19e1895](https://github.com/foundry-rs/foundry/pull/16183/commits/19e189533daf051c6122a761697038b33f846a36)

crates/evm/symbolic/src/executor/opcodes.rs
stevencartavia: [19e1895](https://github.com/foundry-rs/foundry/pull/16183/commits/19e189533daf051c6122a761697038b33f846a36)

mablr (dismissed):

crates/evm/symbolic/src/runtime/memory.rs
```diff
@@ -200,14 +220,20 @@ impl SymMemory {
...
     pub(crate) fn load_word_offset(
-        &self,
+        &mut self,
         cx: &mut SymCx,
         offset: SymExpr,
     ) -> Result<SymExpr, SymbolicError> {
         if let Some(offset) = offset.as_const() {
             let Ok(offset) = usize::try_from(offset) else { return Ok(SymExpr::zero(cx)) };
+            let size = Self::size_after_access(offset, 32);
+            self.materialized_size = self.materialized_size.max(size);
```
figtracer: `mload` creates no stored bytes, so this should update only `logical_size`. `materialized_size` bounds `load_word_dynamic`’s byte-by-byte scan: a concrete `mload(0x400000)` followed by a symbolic `mload` took 7.31s versus 43ms at a small offset, and removing this assignment restored 43ms while preserving `msize`.

crates/evm/symbolic/src/executor/opcodes.rs
```diff
@@ -290,6 +290,136 @@ impl SymbolicExecutor {
...
+            offset.clone(),
+            size.clone(),
+        );
+        let memory_limit = SymExpr::constant(&mut self.cx, U256::from(memory_limit));
+        let local_fits = SymBoolExpr::cmp(&mut self.cx, SymCmpOp::Ule, local_size, memory_limit);
+        let nonzero_valid =
+            SymBoolExpr::and(&mut self.cx, vec![size_fits, offset_fits, local_fits]);
+        let valid_access = SymBoolExpr::or(&mut self.cx, vec![zero_size, nonzero_valid]);
+
+        let outcome = self.apply_memory_access_guard(state, worklist, valid_access)?;
```
figtracer, line 439: the fast path added for the prior solver-query thread still misses locally bounded symbolic ranges. the existing `symbolic_log_accepts_symbolic_offset` test (`uint16` offset, size 32) exceeds nextest’s 90s timeout on this head but passes base in 6.29s. please short-circuit when local upper bounds prove the rounded range is within `memory_limit` before these solver queries.

stevencartavia: [2b13e8f](https://github.com/foundry-rs/foundry/pull/16183/commits/2b13e8f4d8e9bad15e81f42848c8a1ddbb5f8772)

crates/evm/symbolic/src/runtime/memory.rs
stevencartavia: [2b13e8f](https://github.com/foundry-rs/foundry/pull/16183/commits/2b13e8f4d8e9bad15e81f42848c8a1ddbb5f8772)

mablr (approved):

figtracer (approved):

PR #16185, opened 2026-08-14
feat(anvil): add debug_executionWitness endpoint

mattsse:
Adds support for reth's `debug_executionWitness` endpoint to anvil, as a best-effort implementation.

Anvil does not track which state a block's execution actually touched, so instead of the minimal witness reth produces, the endpoint returns a witness for the entire parent state: the RLP encoding of every node of the parent state trie and all storage tries (rebuilt via the incremental trie from the parent's flat state snapshot), all contract codes, the preimages of all account addresses and storage slots, and the locally known ancestor headers within the 256 block `BLOCKHASH` window. Since this is a strict superset of the touched nodes, stateless re-execution of the block against the witness works, at the cost of witness size growing with total state rather than accessed state, which is acceptable for the small states of a dev node.

Limitations, also documented on the backend method: the endpoint errors while forking, because only remotely accessed accounts are known locally and anvil's computed state roots do not match the remote chain's; the parent block's state must still be available in the state history (`--prune-history` discards it); and the genesis block has no witness since it has no parent state.

The trie change in `mem/state.rs` splits the cached `TrieNode::rlp` into a cache wrapper plus an unconditional `encode`, and adds a walk that collects every node's encoding, so witness collection stays O(state) and reuses the existing incremental trie types.

AI assisted.

### DIFF
--- a/.changelog/anvil-debug-execution-witness.md
+++ b/.changelog/anvil-debug-execution-witness.md
@@ -1,0 +1,5 @@
+---
+anvil: minor
+---
+
+Added support for the `debug_executionWitness` RPC endpoint, returning a best-effort witness built from the full parent state of the requested block. Not supported while forking.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -635,6 +635,7 @@ checksum = "88813dc63f261e25d158f07063a71e926f36b3ee8076c1e6d8396348452c5f69"
 dependencies = [
  "alloy-primitives",
  "alloy-rpc-types-anvil",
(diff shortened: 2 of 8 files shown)


stevencartavia (approved):
Non-blocking: current reth exposes this as `(BlockId, Option<ExecutionWitnessMode>)`, but this accepts only a single number/tag parameter. Could we also support hash/EIP-1898 selectors and `null`/`"legacy"` as the optional second parameter?

mablr (approved):

PR #16189, opened 2026-08-14
fix(anvil): stabilize state dump ordering

### LINKED ISSUE #16187: bug(anvil): --dump-state collection ordering is still non-deterministic (blocks, transactions, historical_states)
### Component

Anvil

- [x] Foundry
- [x] Foundryup

forge Version: 1.5.1-v1.5.1 Commit SHA: b0a9dd9ceda36f63e2326ce530c10e6916f4b8a2 Build Timestamp: 2025-12-19T14:07:55.455914129Z (1766153275) Build Profile: maxperf

foundryup: 1.9.3

anvil --dump-state

### Operating System

Linux

### Describe the bug

Two `--dump-state` outputs produced from identical chain state differ byte-for-byte because `blocks`, `transactions`, and `historical_states` are serialized in hash-map iteration order, which is randomized.

This is the remaining half of #9570. That issue was closed by #11415, which made `prevrandao` deterministic — a real fix, and it's in Foundry 1.5.1. But #11415 didn't touch collection ordering, and #9570 was closed with it, so the ordering half went unaddressed. It's the same defect first reported in #3891 in December 2022, where the diagnosis was already correct: sort before serializing, or use a deterministic hasher.

#### What's deterministic and what isn't

`SerializableState` (`crates/anvil/src/eth/backend/db.rs:709`) mixes both kinds of collection:

```rust
pub accounts: BTreeMap<Address, SerializableAccountRecord>,   // ordered
pub blocks: Vec<SerializableBlock>,                            // not
pub transactions: Vec<SerializableTransaction>,                // not
pub historical_states: Option<SerializableHistoricalStates>,   // not
```

The `BTreeMap` fields are stable. The three `Vec`s are built by iterating hash maps in `crates/anvil/src/eth/backend/mem/storage.rs`:

```rust
// :469
pub fn serialized_blocks(&self) -> Vec<SerializableBlock> {
    self.blocks.values().map(|block| block.clone().into()).collect()
}

// :525
pub fn serialized_transactions(&self) -> Vec<SerializableTransaction> {
    self.transactions.values().map(|tx: &MinedTransaction<N>| tx.clone().into()).collect()
}
```

`serialized_states` (`:252`) does the same over `states` and `on_disk_states`. The backing fields are `B256HashMap` (`:305`, `:318`, `:57`).

#### Why the order is random rather than merely unspecified

`B256HashMap` uses alloy's `FbBuildHasher`, which wraps `DefaultHashBuilder`. Foundry enables the `map-foldhash` feature on `alloy-primitives` in the root `Cargo.toml`, and that selects `foldhash::fast::RandomState`. Its seed is derived from the stack pointer, a function pointer, a static address, an allocator address, and the current time.

So the ordering is randomized **per process**, not per machine. Two runs on the same machine with byte-identical inputs produce differently-ordered dumps. Note that this also means back-to-back `anvil_dumpState` calls against a single running instance will agree — the seed is fixed for that process — so a repro has to span two processes.

#### Reproduction

The cleanest repro is the round-trip test requested in #9570, since loading mines no new blocks and so cannot change contents — only ordering can differ:

```bash
# produce an initial state with a few blocks
anvil --dump-state s0.json &
sleep 1
cast rpc anvil_mine 0x5 >/dev/null
kill -INT %1; wait

# round-trip it twice, in separate processes
anvil --load-state s0.json --dump-state s1.json & sleep 1; kill -INT %1; wait
anvil --load-state s1.json --dump-state s2.json & sleep 1; kill -INT %1; wait

sha256sum s1.json s2.json
jq -c '[.blocks[].header.number]' s1.json s2.json
```

Expected: `s1.json` and `s2.json` are identical.
Actual: the hashes differ, and the block-number sequences come back in different orders.

#### Observed in the wild

Comparing two dumps of the same 18-transaction deployment, produced by identical source on two machines:

- `accounts` — all 30 entries byte-identical, same key order (`BTreeMap`).
- `transactions` — same 18 transaction hashes, permuted. Matched by hash, every field agrees: nonce, gas used, deployed address, calldata.
- `blocks` — same 19 block numbers, permuted. Order in one run: `0x6 0x5 0x4 0xd 0x8 0x7 0xa 0x10 0x3 0x1 ...`
- `historical_states` — same keys, permuted.

#### Impact

`--dump-state` output can't be checksummed, which is the use case #9570 was originally filed about: verifying in CI that what's under test matches what was deployed. It also can't be committed to a repository without producing spurious diffs, and it can't be included in a reproducible release artifact.

Sorting is also strictly better than a fixed hasher seed here, because a fixed seed would still reshuffle whenever key values change, and the keys are block hashes.

### Suggested fix

Sort in the three serialization functions. Load is unaffected — `load_blocks`, `load_transactions`, and `load_states` iterate and insert into maps, so file order has never mattered on the way in.

```rust
pub fn serialized_blocks(&self) -> Vec<SerializableBlock> {
    let mut blocks: Vec<SerializableBlock> =
        self.blocks.values().map(|block| block.clone().into()).collect();
    blocks.sort_by_key(|b| b.header.number);
    blocks
}
```

`serialized_transactions` sorts by `(block_number, transaction_index)`; `serialized_states` by hash.

### Regression tests

#9570 asked for these and they still don't exist — `crates/anvil/tests/it/state.rs` on `master` has no assertion that two dumps are equal:

- a blank anvil dumps identically across two processes
- a deployment dumps identically across two processes
- round trip: `dump` → `load` → `dump` produces identical output

The round-trip test is the one that would have caught this, and it's cheap: no wall-clock dependence, since loading mines no blocks.

### Note on the remaining wall-clock dependence

Sorting fixes ordering but not timestamps. Block timestamps still come from the wall clock, so two independent runs of the same deployment script produce different block hashes even with `--timestamp` pinning genesis (the gap from genesis to the first mined block varies with how long the script takes to start broadcasting — I measured 6s versus 11s across two runs).

That's arguably a separate feature request rather than a bug, and the round-trip test above sidesteps it entirely. Flagging it only so the ordering fix isn't mistaken for making `--dump-state` fully reproducible across independent runs. Happy to open that separately if it's wanted.

stevencartavia:
Fixes #16187 by deterministically ordering blocks, transactions, and historical state during serialization while preserving canonical block semantics. This makes equivalent state dumps byte-identical without changing runtime map types or the JSON schema.


mattsse (dismissed):
lgtm

crates/anvil/tests/it/state.rs
```diff
@@ -669,6 +669,24 @@ async fn can_preserve_historical_states_between_dump_and_load() {
...
+    greeter.setGreeting("World!".to_string()).send().await.unwrap().watch().await.unwrap();
+    api.mine_one().await.unwrap();
+
+    let dump = api.anvil_dump_state(Some(true)).await.unwrap();
+    assert_eq!(dump, api.anvil_dump_state(Some(true)).await.unwrap());
+
+    let (loaded_api, _handle) =
+        spawn(NodeConfig::test().with_genesis_timestamp(timestamp.into())).await;
+    assert!(loaded_api.anvil_load_state(dump.clone()).await.unwrap());
+    assert_eq!(dump, loaded_api.anvil_dump_state(Some(true)).await.unwrap());
```
mablr, line 699: Could we add exact unit coverage with deliberately shuffled insertion order for transactions, historical states, and nested snapshot maps? Fresh maps here usually get different seeds, but this same-process round trip is probabilistic.

crates/anvil/src/eth/backend/mem/storage.rs
```diff
@@ -914,6 +930,32 @@ mod tests {
...
+        };
+
+        let mut storage = BlockchainStorage::<FoundryNetwork>::empty();
+        let stale_hash = storage.insert_block(stale);
+        let canonical_hash = storage.insert_block(canonical);
+        assert!(canonical_hash < stale_hash);
+
+        let mut loaded = BlockchainStorage::<FoundryNetwork>::empty();
+        loaded.load_blocks(storage.serialized_blocks());
+        assert_eq!(loaded.hashes.get(&1), Some(&canonical_hash));
```
mablr, line 956: Could this also cover the loaded backend head and the next mined block’s parent? That would exercise the full head-selection path, not only `hashes[number]`.

mablr (commented):
Sgtm, just 2 nits

crates/anvil/src/eth/backend/mem/storage.rs
stevencartavia: [0f59652](https://github.com/foundry-rs/foundry/pull/16189/commits/0f59652245277567127ac56156ef00dd06cb3081)

crates/anvil/tests/it/state.rs
stevencartavia: [0f59652](https://github.com/foundry-rs/foundry/pull/16189/commits/0f59652245277567127ac56156ef00dd06cb3081)

.changelog/anvil-deterministic-state-dumps.md
```diff
@@ -0,0 +1,5 @@
+---
+anvil: patch
```
mablr, line 2: Since this PR changes `StateSnapshot` serialization in the publishable `foundry-evm-core` crate, include it in the package mapping.

```suggestion
anvil: patch
foundry-evm-core: patch
```

mablr (approved):
LGTM

figtracer (approved):

PR #16190, opened 2026-08-14
chore(deps): use released Monad crates

haythemsellami:
Replaces the temporary git patches for `monad-revm` and `alloy-monad-evm` with their published 0.6.0 crates.io releases.


mablr (approved):
Noice, I removed monad from deny whitelist to reduce vulnerability surface.

stevencartavia (approved):

PR #16192, opened 2026-08-14
feat(cheatcodes): add expectDelegateCall

### LINKED ISSUE #8301: Feature Request: Improved delegatecall detection in forge-std
## Current Situation

Currently, there doesn't seem to be a straightforward way to detect or expect `delegatecall` operations using `vm.expectCall` in forge-std. The only method I've found to identify a `delegatecall` is by using state diff recording, which is not as convenient or explicit as `vm.expectCall`. Here's an example of the current workaround:

```solidity
vm.startStateDiffRecording();
address(1).delegatecall("");
VmSafe.AccountAccess[] memory accountAccesses = vm.stopAndReturnStateDiff();
console2.log(accountAccesses[0].kind == VmSafe.AccountAccessKind.DelegateCall);
```

## Feature Request

To improve the testing capabilities for `delegatecall` operations, I propose one or both of the following enhancements:

1. Introduce a new function: `vm.expectDelegateCall`
   This would work similarly to `vm.expectCall` but specifically for `delegatecall` operations.

2. Extend `vm.expectCall` with an `AccountAccessKind` option
   This would allow users to specify the type of call they're expecting, including `delegatecall`.

## Rationale

These additions would make it easier and more intuitive to write tests involving `delegatecall` operations. It would align with the existing patterns in forge-std, making the testing process more consistent across different types of contract interactions.

## Potential Implementation

For the `vm.expectDelegateCall` option:

```solidity
function expectDelegateCall(address target, bytes memory data) external;
```

For the extended `vm.expectCall`:

```solidity
function expectCall(
    address target,
    bytes memory data,
    VmSafe.AccountAccessKind kind
) external;
```

figtracer:
Adds `vm.expectDelegateCall(address,bytes)` so tests can assert that matching calldata is executed specifically through `DELEGATECALL`. The existing expectation tracker now keys matching calldata by an optional call scheme, which preserves `expectCall`’s current any-scheme behavior and lets both expectations coexist for the same target and calldata.

Closes #8301.

This PR was written with Codex, which helped investigate the existing expectation machinery, implement the change, and test it.

### DIFF
--- a/.changelog/expect-delegate-call.md
+++ b/.changelog/expect-delegate-call.md
@@ -1,0 +1,7 @@
+---
+forge: minor
+foundry-cheatcodes: minor
+foundry-cheatcodes-spec: minor
+---
+
+Added `vm.expectDelegateCall` for asserting delegate calls separately from ordinary calls.
--- a/crates/cheatcodes/assets/cheatcodes.json
+++ b/crates/cheatcodes/assets/cheatcodes.json
@@ -5466,6 +5466,26 @@
     },
     {
       "func": {
+        "id": "expectDelegateCall",
+        "description": "Expects a delegate call to an address with the specified calldata.\nCalldata can either be a strict or a partial match.",
+        "declaration": "function expectDelegateCall(address callee, bytes calldata data) external;",
+        "visibility": "external",
+        "mutability": "",
+        "signature": "expectDelegateCall(address,bytes)",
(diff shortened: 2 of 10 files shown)


grandizzy (dismissed):
makes sense, ty!

mablr (dismissed):

mattsse (approved):
lgtm

mablr (approved):

PR #16193, opened 2026-08-14
perf(bind): stream single-file output

figtracer:
## Motivation

`forge bind --single-file` currently joins all generated bindings into one large source string, then parses and pretty-prints that combined file again. On projects with hundreds of contracts, this redundant whole-file pass dominates runtime and peak memory.

## Solution

Stream the already parsed and formatted binding expansions directly through a buffered writer for single-file crate and module output. The multi-file path is unchanged.

The generated Aave v4 single-file output is byte-identical to `master`; regression coverage also checks that crate and module output match and that the generated crate compiles.

### Results

Profiling builds from the same checkout, 595 Aave v4 bindings, three measured Hyperfine runs:

| Metric | `master` | This PR | Improvement |
| --- | ---: | ---: | ---: |
| Wall time | 7.413 s ± 0.043 s | 0.941 s ± 0.072 s | 7.88x |
| Peak RSS | 8.85 GB | 1.18 GB | 7.5x lower |

Command:

```sh
forge bind --skip-build --overwrite --bindings-path <path> --select-all --single-file
```

### Validation

- `cargo test -p forge-sol-macro-gen`
- `cargo test -p forge --test cli bind::bind_single_file_crate_and_module_match -- --exact`
- `cargo clippy -p forge-sol-macro-gen --all-targets --all-features -- -D warnings`
- `cargo clippy -p forge --test cli --all-features -- -D warnings`

Amp was used for investigation, implementation, benchmarking, tests, and drafting this description. I reviewed the resulting code and evidence.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/sol-macro-gen/src/sol_macro_gen.rs
```diff
@@ -233,34 +231,23 @@ edition = "2021"
 
         fs::create_dir_all(bindings_path)?;
 
-        let mut mod_contents =
```
figtracer, line 236: better to have this as a constant at the top i think

mattsse (approved):
lgtm

mablr (approved):
sgtm

stevencartavia (approved):

PR #16199, opened 2026-08-14
fix(symbolic): preserve reverted recordings

stevencartavia:
Preserves log and storage-access recordings from reverted child calls, creates, and `deployCode` executions, matching concrete Forge inspector behavior while still rolling back transactional state.


mattsse (approved):
lgtm

mablr (approved):

PR #16200, opened 2026-08-14
fix(forge): preserve skip after caught revert

### LINKED ISSUE #16197: bug(forge): vm.skip in setUp reported as FAIL instead of SKIP when setUp does substantial work (regression)
Yes, this is an AI agent written issue, but I've reviewed and edited for accuracy.

### Component

Forge

```
forge Version: 1.8.0-nightly
Commit SHA: 50dd4b8accae878391b0825885f0632d1e0a1e4c
Build Timestamp: 2026-08-07T06:26:42.249460000Z
Build Profile: dist
```

`forge test`

### Operating System

macOS (Darwin 25.6.0, arm64)

### Describe the bug

`vm.skip(true, …)` called inside `setUp()` is reported as a **failing test** rather than a
skipped suite, with the internal `FOUNDRY::SKIP` sentinel leaking into the failure message:

```
Ran 1 test for test/SkipProbe.t.sol:SkipProbe_AfterSuper_Test
[FAIL: FOUNDRY::SKIPprobe after super] setUp() (gas: 0)
Suite result: FAILED. 0 passed; 1 failed; 0 skipped
```

Expected (and what older versions do):

```
[SKIP: skipped: probe after super] setUp() (gas: 0)
Suite result: ok. 0 passed; 0 failed; 1 skipped
```

This is a regression of the behaviour added for #9531 by #9532. Note the missing separator
in `FOUNDRY::SKIPprobe after super` — the sentinel prefix is being surfaced verbatim as a
revert string instead of being recognised and stripped.

**The trigger is conditional on what `setUp()` does before the skip.** A `setUp` that only
calls `vm.skip` works fine. A `setUp` that first runs a large amount of setup work (in our
case a full multi-contract protocol deployment via an inherited base `setUp`) fails.

### Minimal-ish reproduction

We could not reduce this to a standalone snippet — see "What we ruled out" below. It does
reproduce reliably in a public repo with a small test file.

Human note: Sorry to make you download our massive monorepo, but it is in fact the case that we cannot reproduce this in a nice clean stand-alone repo.

```bash
git clone https://github.com/ethereum-optimism/optimism
cd optimism/packages/contracts-bedrock
# any recent develop; we saw it at 4c3d0d3
forge install
```

Add `test/SkipProbe.t.sol`:

```solidity
pragma solidity 0.8.15;

import { CommonTest } from "test/setup/CommonTest.sol";

contract SkipProbe_AfterSuper_Test is CommonTest {
    function setUp() public virtual override {
        super.setUp();
        vm.skip(true, "probe after super");
    }

    function test_probe_succeeds() public pure { }
}

contract SkipProbe_BeforeSuper_Test is CommonTest {
    function setUp() public virtual override {
        vm.skip(true, "probe before super");
        super.setUp();
    }

    function test_probe_succeeds() public pure { }
}

contract SkipProbe_NoSuper_Test is CommonTest {
    function setUp() public virtual override {
        vm.skip(true, "probe without super");
    }

    function test_probe_succeeds() public pure { }
}
```

Run:

```bash
forge test --deny never --match-path test/SkipProbe.t.sol
```

`CommonTest.setUp()` deploys a full L1 + L2 protocol stack — it is the "substantial work"
that appears to matter.

### Actual output

```
Ran 1 test for test/SkipProbe.t.sol:SkipProbe_BeforeSuper_Test
[SKIP: skipped: probe before super] setUp() (gas: 0)
Suite result: ok. 0 passed; 0 failed; 1 skipped; finished in 987.54µs

Ran 1 test for test/SkipProbe.t.sol:SkipProbe_NoSuper_Test
[SKIP: skipped: probe without super] setUp() (gas: 0)
Suite result: ok. 0 passed; 0 failed; 1 skipped; finished in 990.62µs

Ran 1 test for test/SkipProbe.t.sol:SkipProbe_AfterSuper_Test
[FAIL: FOUNDRY::SKIPprobe after super] setUp() (gas: 0)
Suite result: FAILED. 0 passed; 1 failed; 0 skipped; finished in 1.81s
```

The only difference between the passing and failing case is whether `super.setUp()` runs
before `vm.skip`.

### Regression range

Same repo, same commit, same test file, artifacts unchanged — only the forge binary varies:

| forge | Result |
|---|---|
| 1.2.3 | `[SKIP]`, suite ok |
| 1.5.1-dev | `[SKIP]`, suite ok |
| 1.6.0-nightly (`5e88010a83`) | inconclusive (unrelated RPC 429 in our env) |
| 1.8.0-nightly (`50dd4b8`) | **`[FAIL: FOUNDRY::SKIP…]`** |

So it broke somewhere in 1.6 → 1.8. We did not narrow further.

### What we ruled out

None of these reproduce it in a standalone project — all still report `[SKIP]` correctly on
1.8.0-nightly:

- The two-argument `vm.skip(bool, string)` form vs the bare `vm.skip(bool)` form (both
  affected in the real repo, both fine standalone).
- Simple state changes before the skip (`new`, an external call, `vm.deal`).
- Heavy gas consumption before the skip (60 sequential contract deployments).
- `vm.etch` + `vm.label` + `vm.warp` + `vm.roll` + `vm.deal` before the skip.
- Deep nested external calls (12 levels) before the skip.
- `isolate` mode — reproduces identically with `--isolate` and `--no-isolate`. (Noting it
  because `isolate` now defaults to `true`, cf. #9912.)
- State snapshots — the affected `setUp` chain uses none.

### Workarounds

- `--allow-failure` makes the process exit 0, but the suite is still reported as FAILED, so
  it does not help you read results.
- Moving the `vm.skip` call before the heavy setup work restores correct reporting, but is
  often not viable: the condition being tested frequently depends on state the setup work
  initialises.
- Filtering the output. What we do:
  ```bash
  forge test … 2>&1 | awk '/FOUNDRY::SKIP/ { held=0; next } held { print buf } { buf=$0; held=1 } END { if (held) print buf }'
  ```
  (the buffer is needed because the `Encountered N failing test in …` header precedes the
  `[FAIL: FOUNDRY::SKIP…]` line, so a plain `grep -v` orphans the header)

mablr:
## Motivation

Fix #16197.

When `setUp` caught a nested revert before calling `vm.skip`, Foundry retained the earlier call as the revert origin. The valid skip payload then failed the cheatcode-origin check and was reported as a test failure. This lets genuine skip reverts supersede earlier caught reverts while preserving protection against forged `FOUNDRY::SKIP` payloads. Regression coverage exercises both cases.

AI assistance was used to investigate the root cause.

### DIFF
--- a/crates/evm/evm/src/inspectors/stack.rs
+++ b/crates/evm/evm/src/inspectors/stack.rs
@@ -44,7 +44,7 @@ use revm::{
     state::{Account, AccountStatus},
 };
 use std::{
-    ops::{Deref, DerefMut},
+    ops::{Deref, DerefMut, Range},
     sync::Arc,
 };
 
@@ -300,6 +300,18 @@ macro_rules! call_inspectors {
             }
         )+
     };
+    (#[stop] [$($inspector:expr),+ $(,)?], |$id:ident $(,)?| $body:expr $(,)?) => {{
+        let mut stop = false;
+        $(
+            if !stop {
+                if let Some($id) = $inspector {
+                    $crate::utils::cold_path();
+                    stop = $body;
+                }
+            }
+        )+
+        stop
+    }};
     (#[ret] [$($inspector:expr),+ $(,)?], |$id:ident $(,)?| $body:expr $(,)?) => {{
         $(
             if let Some($id) = $inspector {
@@ -336,6 +348,68 @@ pub struct InnerContextData {
     original_origin: Address,
     /// Accounts that were created locally before entering the nested EVM context.
     locally_created_accounts: AddressHashSet,
+}
+
+#[derive(Clone, Debug, Default)]
+struct SkipFrameProvenance {
+    return_data_len: Option<usize>,
+    memory: Vec<Range<usize>>,
+    reverting_skip: bool,
+}
+
+#[derive(Clone, Debug, Default)]
+struct SkipProvenance {
+    frames: Vec<SkipFrameProvenance>,
+}
+
+impl SkipProvenance {
+    fn is_active(&self) -> bool {
+        self.frames.iter().any(|frame| frame.return_data_len.is_some() || !frame.memory.is_empty())
+    }
+
+    fn ensure_depth(&mut self, depth: usize) {
+        let len = depth + 1;
+        if self.frames.len() < len {
+            self.frames.resize_with(len, SkipFrameProvenance::default);
+        }
+    }
+
+    fn clear_frame(&mut self, depth: usize) {
+        if let Some(frame) = self.frames.get_mut(depth) {
+            *frame = SkipFrameProvenance::default();
+        }
+    }
+
+    fn take_reverting_skip(&mut self, depth: usize) -> bool {
+        self.frames.get_mut(depth).is_some_and(|frame| std::mem::take(&mut frame.reverting_skip))
+    }
+
+    fn set_return_data(&mut self, depth: usize, len: Option<usize>) {
+        self.ensure_depth(depth);
+        self.frames[depth].return_data_len = len;
+    }
+
+    fn invalidate_memory(&mut self, depth: usize, range: &Range<usize>) {
+        if range.is_empty() {
+            return;
+        }
+        if let Some(frame) = self.frames.get_mut(depth) {
+            frame.memory.retain(|tagged| tagged.start >= range.end || range.start >= tagged.end);
+        }
+    }
+
+    fn tag_memory(&mut self, depth: usize, range: Range<usize>) {
+        if range.is_empty() {
+            return;
+        }
+        self.ensure_depth(depth);
+        self.invalidate_memory(depth, &range);
+        self.frames[depth].memory.push(range);
+    }
+
+    fn is_tagged(&self, depth: usize, range: &Range<usize>) -> bool {
+        self.frames.get(depth).is_some_and(|frame| frame.memory.contains(range))
+    }
 }
 
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
@@ -418,8 +492,8 @@ pub struct InspectorStackInner {
     top_level_frame_failed_before_rewrite: bool,
     /// Address that reverted the call, if any.
     pub reverter: Option<Address>,
-    /// Payload of the skip that most recently set the reverter to the cheatcode address.
-    skip_revert: Option<Bytes>,
+    /// Tracks genuine skip data as it propagates through return data and memory.
+    skip_provenance: SkipProvenance,
     /// LIFO stack tracking CREATE2 frames that were redirected to the CREATE2 factory.
     pending_create2_redirects: Vec<PendingCreate2Redirect>,
     /// Pending CREATE2 deployer validation error, deferred from `frame_start` to `create` so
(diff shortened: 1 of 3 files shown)


crates/evm/evm/src/inspectors/stack.rs
```diff
@@ -902,8 +902,11 @@ impl<FEN: FoundryEvmNetwork> InspectorStackRefMut<'_, FEN> {
             },
         );
 
-        // Record first address that reverted the call.
-        if result.is_revert() && self.reverter.is_none() {
+        // Record the first address that reverted the call, but let an actual skip take precedence
+        // over earlier caught reverts so its cheatcode origin is preserved.
+        let is_skip =
+            inputs.target_address == CHEATCODE_ADDRESS && outcome.output().starts_with(MAGIC_SKIP);
+        if result.is_revert() && (self.reverter.is_none() || is_skip) {
```
mattsse: This can leave stale cheatcode origin after a caught skip. For example, catch an ordinary revert, catch a low-level `vm.skip(...)`, then have `setUp` revert with a user-created `FOUNDRY::SKIP...` payload. The caught skip overwrites `reverter` here; the later top-level revert does not replace it, so `RawCallResult::into_evm_error` reports the suite as skipped instead of failed. The forged-payload regression does not cover the intervening caught genuine skip. Please only retain the skip origin when it belongs to the terminal revert.

mattsse (changes requested):
One correctness issue remains.

crates/evm/evm/src/inspectors/stack.rs
mablr: https://github.com/foundry-rs/foundry/pull/16200/commits/61165b241cb564eb7eee2c41d7156e8df852db3b

crates/evm/evm/src/inspectors/stack.rs
```diff
@@ -902,9 +904,21 @@ impl<FEN: FoundryEvmNetwork> InspectorStackRefMut<'_, FEN> {
...
+        // over earlier caught reverts while its payload propagates to the terminal revert.
+        let is_skip =
+            inputs.target_address == CHEATCODE_ADDRESS && outcome.output().starts_with(MAGIC_SKIP);
+        if result.is_revert() {
+            if is_skip {
+                self.reverter = Some(inputs.target_address);
+                self.skip_revert = Some(outcome.output().clone());
+            } else if self.reverter.is_none()
+                || (self.reverter == Some(CHEATCODE_ADDRESS)
+                    && self.skip_revert.as_ref() != Some(outcome.output()))
```
mattsse: This payload comparison is not a reliable provenance boundary. For a non-skip cheatcode revert such as `vm.assertEq(1, 2)`, `skip_revert` is `None`; when the same revert unwinds through the handler, `None != Some(outcome.output())` clears the `CHEATCODE_ADDRESS` origin. `is_cheatcode_assert_revert` then misses it, so invariant/fuzz commands incorrectly succeed—the current `test / test all` job has four deterministic failures from this path. Conversely, after catching `vm.skip(true, "caught skip")`, a terminal forged `FOUNDRY::SKIPcaught skip` has identical bytes, so this condition preserves the stale cheatcode origin and still classifies the forged revert as a skip. The provenance needs to be tied to the actual propagating call/frame rather than inferred from user-controlled payload equality.

mattsse (changes requested):
The prior sequence now has regression coverage, but the payload-based provenance tracking still has correctness holes.

crates/evm/evm/src/inspectors/stack.rs
mablr: https://github.com/foundry-rs/foundry/pull/16200/commits/e94187ce2cd2e103c8de41a8be4ae4e0d72d3574

mattsse (approved):
lgtm

PR #16201, opened 2026-08-14
fix(forge): report canonical fork blocks

stevencartavia:
Failed fork tests can report the execution block after `vm.roll` rather than the backing fork block, producing misleading reproduction information.

This preserves the canonical fork block while separately tracking real fork transitions such as `rollFork` and `selectFork`, including invariant campaigns using copy-on-write backends. Snapshot restoration and inactive fork rolls also preserve the correct active fork block.

### DIFF
--- a/.changelog/canonical-fork-block-reporting.md
+++ b/.changelog/canonical-fork-block-reporting.md
@@ -1,0 +1,5 @@
+---
+forge: patch
+---
+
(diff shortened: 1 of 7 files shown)


mattsse (approved):
lgtm

mablr (approved):

PR #16203, opened 2026-08-15
fix(anvil): support block IDs in witnesses

mattsse:
Follow-up to #16185. `debug_executionWitness` now accepts reth-compatible `BlockId` selectors, including block hashes and EIP-1898 objects. The original request type narrowed the selector to `BlockNumberOrTag`.

Hashes are resolved to the exact block they name instead of being reduced to a number: the witness is built from the parent state looked up by the parent hash and the ancestor headers are walked by parent hash. A block that was replaced at the same height (e.g. via `anvil_loadState`) therefore still resolves to its own witness rather than the canonical block's, and `requireCanonical: true` rejects hashes that are no longer canonical.

AI assisted with the implementation and tests.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -179,7 +179,7 @@ use op_alloy_consensus::{DEPOSIT_TX_TYPE_ID, POST_EXEC_TX_TYPE_ID};
 use op_revm::{OpTransaction, transaction::deposit::DepositTransactionParts};
 use parking_lot::{Mutex, RwLock, RwLockUpgradableReadGuard};
 use revm::{
-    Database as RevmDatabase, DatabaseCommit, Inspector,
+    Database as RevmDatabase, DatabaseCommit, DatabaseRef as RevmDatabaseRef, Inspector,
     context::{Block as RevmBlock, BlockEnv, Cfg, CfgEnv, ContextSetters, ContextTr, TxEnv},
     context_interface::{
         JournalTr,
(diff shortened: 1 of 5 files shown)


crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -7062,13 +7062,9 @@ where
...
         &self,
-        block: BlockNumber,
+        block: BlockId,
     ) -> Result<ExecutionWitness, BlockchainError> {
-        let number = self.convert_block_number(Some(block));
-        let best = self.best_number();
-        if number > best {
-            return Err(BlockchainError::BlockOutOfRange(best, number));
-        }
+        let number = self.ensure_block_number(Some(block)).await?;
```
stevencartavia: Could we preserve the selected block hash here instead of reducing it to a number? After `anvil_loadState` replaces a block at the same height, the old block remains addressable by hash. Requesting its witness then returns the canonical block’s witness, and `requireCanonical: true` incorrectly succeeds.

mablr (approved):

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -7062,26 +7062,47 @@ where
...
-        let Some(parent) = number.checked_sub(1) else {
+        // Resolve hashes to the exact block they name instead of reducing them to a number: a
+        // hash can refer to a block that was since replaced at the same height (e.g. via
+        // `anvil_loadState`) and must not be conflated with the canonical block at that number.
+        let block = match block {
+            BlockId::Hash(id) => {
+                let block =
+                    self.get_block_by_hash(id.block_hash).ok_or(BlockchainError::BlockNotFound)?;
+                if id.require_canonical == Some(true)
+                    && self.block_hash_by_number(block.header.number()) != Some(id.block_hash)
```
figtracer, line 7075: `anvil_loadState` can rewind `best_number` without removing higher `hashes` entries, so this still accepts an old block above the new head as canonical. reproduced by dumping genesis, mining block 2, reloading genesis, then requesting block 2 with `requireCanonical: true`: it returned a witness. please also require `block.header.number() <= self.best_number()` and test before remine.

mattsse:
honestly, dont wanna touch this again, this is kinda a mess

PR #16202, opened 2026-08-15
Revert "fix(forge): preserve skip after caught revert"

mattsse:
Reverts foundry-rs/foundry#16200

@mablr maybe merging this pr was a bit pre-mature, i dont actually find these changes reasonable, the last commit in that pr doesnt feel right

PR #16204, opened 2026-08-15
fix(forge): classify skips by minted cheatcode payload

### LINKED ISSUE #16197: bug(forge): vm.skip in setUp reported as FAIL instead of SKIP when setUp does substantial work (regression)

mattsse:
Fixes #16197. The executor only classified a top-level revert as a skip when the skip cheatcode was the first recorded reverter, so any revert caught in `setUp` before `vm.skip` claimed that slot and the genuine skip was reported as `[FAIL: FOUNDRY::SKIP...]`. The previous fix (#16200) addressed this with opcode-level payload provenance tracking in the inspector stack and was reverted in #16202.

This replaces the reverter-identity check with payload identity: the skip cheatcode records each payload it mints in the cheatcode state, result conversion copies the payloads onto `RawCallResult`, and a revert is classified as a skip iff its data byte-equals a recorded payload. The new `RawCallResult::skip_reason` helper is shared by the executor error path, fuzz, showmap, and symbolic replay, which previously each paired the reverter check with `SkipReason::decode` by hand. Reverter attribution itself is untouched, so first-reverter semantics for `should_ignore_revert` and invariant failure attribution do not change, and no per-opcode tracking is involved. User-crafted `FOUNDRY::SKIP` revert data still fails the test because no matching payload was minted, including forging different bytes after catching a genuine skip. One semantic is pinned deliberately: a caught genuine skip re-raised byte-identically counts as a skip, since distinguishing manual re-raises from compiler-generated revert bubbling requires exactly the byte-flow tracking #16200 was reverted for, and either way the payload provably originated from `vm.skip(true)` in the same call.

The issue's reproduction is converted into the `issue_16197` repro test (caught revert inside an inherited base `setUp` before the skip), with further tests covering the minimal caught-revert trigger under `--isolate`, forged payloads with and without a preceding caught genuine skip, and the identical re-raise.

AI assistance was used for the investigation and implementation.

### DIFF
--- a/crates/evm/evm/src/executors/mod.rs
+++ b/crates/evm/evm/src/executors/mod.rs
@@ -898,9 +898,6 @@ impl<FEN: FoundryEvmNetwork> Executor<FEN> {
             // Clear broadcastable transactions
             cheats.broadcastable_transactions.clear();
             cheats.ignored_traces.ignored.clear();
-            // Skip payloads are scoped to the call they were minted in.
-            cheats.skip_payloads.clear();
-
             // if tracing was paused but never unpaused, we should begin next frame with tracing
             // still paused
             if let Some(last_pause_call) = cheats.ignored_traces.last_pause_call.as_mut() {
@@ -1300,7 +1297,7 @@ pub struct RawCallResult<FEN: FoundryEvmNetwork = EthEvmNetwork> {
     pub reverter: Option<Address>,
     /// Revert payloads minted by the `skip` cheatcode during this call.
     ///
-    /// Copied out of the cheatcode state on conversion since `commit` moves that state back into
+    /// Moved out of the cheatcode state on conversion since `commit` moves that state back into
     /// the executor before results are classified.
     pub skip_payloads: Vec<Bytes>,
 }
@@ -1603,7 +1600,7 @@ fn convert_executed_result<FEN: FoundryEvmNetwork>(
         line_coverage,
         edge_coverage,
         evm_cmp_values,
-        cheatcodes,
+        mut cheatcodes,
         chisel_state,
         reverter,
     } = inspector.collect();
@@ -1621,7 +1618,8 @@ fn convert_executed_result<FEN: FoundryEvmNetwork>(
         .as_ref()
         .map(|c| c.broadcastable_transactions.clone())
         .filter(|txs| !txs.is_empty());
-    let skip_payloads = cheatcodes.as_ref().map(|c| c.skip_payloads.clone()).unwrap_or_default();
+    let skip_payloads =
+        cheatcodes.as_mut().map(|c| std::mem::take(&mut c.skip_payloads)).unwrap_or_default();
 
     Ok(RawCallResult {
         exit_reason: Some(exit_reason),
--- a/crates/evm/evm/src/executors/mod.rs
+++ b/crates/evm/evm/src/executors/mod.rs
@@ -898,6 +898,8 @@ impl<FEN: FoundryEvmNetwork> Executor<FEN> {
             // Clear broadcastable transactions
             cheats.broadcastable_transactions.clear();
             cheats.ignored_traces.ignored.clear();
+            // Skip payloads are scoped to the call they were minted in.
+            cheats.skip_payloads.clear();
 
             // if tracing was paused but never unpaused, we should begin next frame with tracing
             // still paused
@@ -1296,6 +1298,11 @@ pub struct RawCallResult<FEN: FoundryEvmNetwork = EthEvmNetwork> {
     /// The chisel state
     pub chisel_state: Option<(Vec<U256>, Vec<u8>)>,
     pub reverter: Option<Address>,
+    /// Revert payloads minted by the `skip` cheatcode during this call.
+    ///
+    /// Copied out of the cheatcode state on conversion since `commit` moves that state back into
+    /// the executor before results are classified.
+    pub skip_payloads: Vec<Bytes>,
 }
 
 impl<FEN: FoundryEvmNetwork> Default for RawCallResult<FEN> {
@@ -1328,6 +1335,7 @@ impl<FEN: FoundryEvmNetwork> Default for RawCallResult<FEN> {
             fork_block_number: None,
             chisel_state: None,
             reverter: None,
+            skip_payloads: Vec::new(),
         }
     }
 }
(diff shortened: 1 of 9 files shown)


stevencartavia (dismissed):

crates/evm/evm/src/executors/mod.rs
```diff
@@ -898,6 +898,8 @@ impl<FEN: FoundryEvmNetwork> Executor<FEN> {
             // Clear broadcastable transactions
             cheats.broadcastable_transactions.clear();
             cheats.ignored_traces.ignored.clear();
+            // Skip payloads are scoped to the call they were minted in.
+            cheats.skip_payloads.clear();
```
mablr: Consider pinning this lifetime boundary with an integration test: catch `vm.skip(..., "x")` in `setUp` and return successfully, then have the test revert with the exact `FOUNDRY::SKIPx` bytes. It should fail, proving minted payloads cannot authenticate a later executor call.

mablr (dismissed):
SGTM, just an optional suggestion

maurelian:
I tested this branch at f7a1468. The bug is fixed on that commit.

crates/evm/evm/src/executors/mod.rs
```diff
@@ -1604,6 +1621,7 @@ fn convert_executed_result<FEN: FoundryEvmNetwork>(
         .as_ref()
         .map(|c| c.broadcastable_transactions.clone())
         .filter(|txs| !txs.is_empty());
+    let skip_payloads = cheatcodes.as_ref().map(|c| c.skip_payloads.clone()).unwrap_or_default();
```
figtracer: `cheatcodes` is owned here, so can we move `skip_payloads` out with `std::mem::take` instead of cloning every user-controlled payload? this also removes the separate `skip_payloads.clear()` in `commit`.

mablr: https://github.com/foundry-rs/foundry/pull/16204/commits/f5766e4574e84a8b9710e13d1c6dd9c61d7e8423

mablr (approved):

figtracer (approved):

figtracer (approved):

PR #16206, opened 2026-08-15
fix(cast): save vanity wallets atomically

stevencartavia:
`cast wallet vanity --save-path` currently truncates the existing wallet file before writing its replacement. If that write fails due to an I/O error, quota, or insufficient space, all previously saved private keys can be lost.

This change writes the updated wallet into a secured staging file and atomically replaces the destination on Linux and macOS. It preserves existing symlink behavior and `0600` permissions, while preventing inherited macOS ACLs from weakening the replacement file. Other platforms retain the existing write behavior.

AI assistance was used to investigate filesystem failure modes, develop the implementation and tests, and review cross-platform permission and ACL behavior. I reviewed and validated the resulting changes.

### DIFF
--- a/.changelog/atomically-save-vanity-wallets.md
+++ b/.changelog/atomically-save-vanity-wallets.md
@@ -1,0 +1,5 @@
+---
+cast: patch
+---
+
+Preserved existing vanity wallets on Linux and macOS when saving an additional wallet fails.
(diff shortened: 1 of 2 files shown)


mattsse (commented):
i dont think we need this

PR #16207, opened 2026-08-15
feat(anvil): add getLastBlockWallTime RPC

### LINKED ISSUE #8217: feat(`anvil`): add `getLastBlockWallTime` debug RPC method
### Component

Anvil

To sync clocks effectively for real-time games, it would be really helpful to be able to get the unix wall time in ms when the last block was created. What I've been doing up until now is start anvil with a script like this:

```
function checkTimestamp() {
  const timestamp = Date.now();
  if ((timestamp + 25) % 1000 === 0) {
    console.log(`Timestamp ${timestamp} is divisible by 1000.`);
    process.exit(0); // Stop the program
  } else {
    setImmediate(checkTimestamp);
  }
}

console.log("Waiting for a timestamp divisible by 1000...");
checkTimestamp();

// startup anvil
```

But the interval between anvil's blocks sometimes drifts a little, resulting in timestamps drifting backwards or ahead of real time. Having an rpc method to get the exact offset would allow clients to adjust to any minor timestamp offset without the need for anvil to robustly sync to wall time.

### Additional context

_No response_

stevencartavia:
Adds `anvil_getLastBlockWallTime`, returning the UNIX wall time in milliseconds when the current head was published. The marker updates only after successful mining, remains unchanged by `evm_setTime`, and resets when another head is installed.

This enables precise clock synchronization for real-time clients where second-resolution block timestamps are insufficient.

Closes #8217.

### DIFF
--- a/crates/anvil/src/eth/backend/mem/mod.rs
+++ b/crates/anvil/src/eth/backend/mem/mod.rs
@@ -5087,7 +5087,9 @@ impl<N: Network> Backend<N> {
         let Some((num, hash)) = self.active_state_snapshots.lock().remove(&id) else {
             return Ok(false);
         };
-        let best_block_hash = {
+        let block = self.block_by_hash(hash).await?.ok_or(BlockchainError::BlockNotFound)?;
+
+        {
             // revert the storage that's newer than the snapshot
             let current_height = self.best_number();
             let mut storage = self.blockchain.storage.write();
@@ -5108,10 +5110,7 @@ impl<N: Network> Backend<N> {
 
             storage.best_number = num;
             storage.best_hash = hash;
-            hash
-        };
-        let block =
-            self.block_by_hash(best_block_hash).await?.ok_or(BlockchainError::BlockNotFound)?;
+        }
 
         let reset_time = block.header.timestamp();
         self.time.reset(reset_time);
@@ -9798,6 +9797,25 @@ mod tests {
             source_fork_block_number: None,
             source_fork_block_hash: None,
         }
+    }
+
+    #[tokio::test]
+    async fn missing_snapshot_block_does_not_change_head() {
+        let (api, _handle) = spawn(NodeConfig::test()).await;
+        let snapshot_hash = api.backend.best_hash();
+        let snapshot = api.backend.create_state_snapshot().await;
+        api.mine_one().await.unwrap();
+        let best_number = api.backend.best_number();
+        let best_hash = api.backend.best_hash();
+        let head_wall_time = api.backend.time().last_block_wall_time();
+
+        api.backend.blockchain.storage.write().blocks.remove(&snapshot_hash);
+        let err = api.backend.revert_state_snapshot(snapshot).await.unwrap_err();
+
+        assert!(matches!(err, super::BlockchainError::BlockNotFound));
+        assert_eq!(api.backend.best_number(), best_number);
+        assert_eq!(api.backend.best_hash(), best_hash);
+        assert_eq!(api.backend.time().last_block_wall_time(), head_wall_time);
     }
 
     struct CacheFlushingDb(BlockchainDb);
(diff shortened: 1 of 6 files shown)


crates/anvil/src/eth/backend/time.rs
```diff
@@ -60,10 +61,22 @@ impl TimeManager {
     /// Resets the current time manager to the given timestamp, resetting the offsets and
     /// next block timestamp option
     pub fn reset(&self, start_timestamp: u64) {
-        let current = duration_since_unix_epoch().as_secs() as i128;
+        self.reset_timestamp(start_timestamp, true);
```
mablr, line 64: `reset` is also used when restoring an existing head, so this records installation time rather than creation time. Could we align the RPC docs and changelog with that behavior?

crates/anvil/src/eth/backend/mem/mod.rs
```diff
@@ -5926,6 +5926,7 @@ where
 
             storage.blocks.insert(block_hash, block);
             storage.hashes.insert(block_number, block_hash);
+            self.time.mark_block_created();
```
mablr: Could this be moved to the end of the storage update? It currently becomes observable before transaction indexing and pruning finish.

crates/anvil/tests/it/anvil_api.rs
```diff
@@ -390,6 +390,30 @@ async fn can_mine_manually() {
     }
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn get_last_block_wall_time_updates_after_mining() {
+    let before_spawn =
+        SystemTime::now().duration_since(SystemTime::UNIX_EPOCH).unwrap().as_millis() as u64;
+    let (api, _handle) = spawn(NodeConfig::test()).await;
+    let genesis_wall_time = api.anvil_get_last_block_wall_time().unwrap();
```
mablr: Could this call the method through JSON-RPC? The direct API call does not verify routing or response serialization.

crates/anvil/src/eth/backend/time.rs
stevencartavia: [f490507](https://github.com/foundry-rs/foundry/pull/16207/commits/f49050764039f02258c6664b63f7f418681688c9)

crates/anvil/src/eth/backend/mem/mod.rs
stevencartavia: [f490507](https://github.com/foundry-rs/foundry/pull/16207/commits/f49050764039f02258c6664b63f7f418681688c9)

crates/anvil/tests/it/anvil_api.rs
stevencartavia: [f490507](https://github.com/foundry-rs/foundry/pull/16207/commits/f49050764039f02258c6664b63f7f418681688c9)

crates/anvil/src/eth/backend/time.rs
```diff
@@ -60,10 +61,22 @@ impl TimeManager {
...
+        self.reset_timestamp(timestamp, false);
+    }
+
+    fn reset_timestamp(&self, start_timestamp: u64, mark_new_head: bool) {
+        let current = duration_since_unix_epoch();
         let mut state = self.state.write();
         state.last_timestamp = start_timestamp;
-        state.offset = (start_timestamp as i128) - current;
+        if mark_new_head {
+            state.last_block_wall_time = current.as_millis().try_into().unwrap_or(u64::MAX);
```
mablr, line 77: `revert_state_snapshot` updates `best_*` before a fallible block lookup and calls `reset` only afterward. If that lookup fails, the head changes while this marker remains stale. Could we resolve the block before mutating storage?

stevencartavia: [c58e1f8](https://github.com/foundry-rs/foundry/pull/16207/commits/c58e1f8be2d1baa279f73f7fcd14251bb6abe0e3)

mablr (approved):

figtracer (approved):

